PR #2217, opened 2019-03-13
Fix a couple of issues in README.md under the Windows section

remyabel:
The chocolately command was not copy/paste friendly. This fixes that. Note that
the leading space is intentional since we are doing a line continuation. Also,
the other parts of the guide mention installing git, so we add that as well to
be consistent.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2217)

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,8 +337,8 @@ Studio 2017". If you have not, install [Chocolatey](https://www.chocolatey.com)
 using this command as the administrator:
 
 ```console
-@"%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -ExecutionPolicy Bypass -Command "iex (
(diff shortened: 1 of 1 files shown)


remyabel:
Flaky build?

```
Submodule 'googleapis' (https://github.com/googleapis/googleapis.git) registered for path 'third_party/googleapis'
Cloning into 'T:/src/github/google-cloud-cpp/third_party/googleapis'...
Submodule path 'third_party/googleapis': checked out '6a3277c0656219174ff7c345f31fb20a90b30b97'
ERROR: Aborting VM command due to timeout of 7200 seconds
```

coryan:
Usually "flaky" means "fails only sometimes", this one is failing, but for unrelated reasons (the test infrastructure seems broken). I pushed some debugging information to another PR.

PR #2220, opened 2019-03-13
Modify the scripts to update the ABI snapshot.

coryan:
Developers would be able to run:

$ BUILD_NAME=update-abi ./ci/kokoro/docker/build.sh

to update the ABI snapshots. The new snapshots were created using that
technique.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2220)


coryan:
For the record, this is the effect of the change.

## Before:

https://source.cloud.google.com/results/invocations/c668ffec-25f0-4489-bd3b-946cc3bb3467/targets/cloud-cpp%2Fgithub%2Fdocker%2Fcheck-abi/log

the interesting stuff is around line 4181 (sorry, no direct link).

## After:

https://source.cloud.google.com/results/invocations/4a5c1a40-43cf-4fdf-a307-9cde95e819c8/targets/cloud-cpp%2Fgithub%2Fdocker%2Fcheck-abi-presubmit/log

the interesting stuff is around line 4638.

PR #2223, opened 2019-03-13
Changed to use the latest version of bazel

devjgm:
Updated the install-bazel.sh script to test using a newer version of bazel. Also changed the script to compute the `PLATFORM` value from `uname` rather than require the user to pass it as an argument.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2223)

### DIFF
--- a/ci/install-bazel.sh
+++ b/ci/install-bazel.sh
@@ -18,12 +18,12 @@ set -eu
 
 readonly PLATFORM=$(uname -s | tr '[:upper:]' '[:lower:]')
 # Ensures that the platform is one that we expect
-grep -P "linux|darwin" <<<"${PLATFORM}" > /dev/null
+test "${PLATFORM}" = "linux" -o "${PLATFORM}" = "darwin"
 
 # Gets the latest version number of bazel from the GitHub JSON API.
 readonly BAZEL_VERSION=$(
   curl -sL "https://api.github.com/repos/bazelbuild/bazel/releases/latest" \
-    | grep -P '"tag_name":' \
+    | grep '"tag_name":' \
     | cut -f4 -d\"
 )
 test -n "${BAZEL_VERSION}"
--- a/ci/install-bazel.sh
+++ b/ci/install-bazel.sh
@@ -29,8 +29,8 @@ test -n "${BAZEL_VERSION}"
 
 readonly GITHUB_DL="https://github.com/bazelbuild/bazel/releases/download"
 readonly SCRIPT_NAME="bazel-${BAZEL_VERSION}-installer-${PLATFORM}.sh"
-wget -q "${GITHUB_DL}/${BAZEL_VERSION}/${SCRIPT_NAME}"
-wget -q "${GITHUB_DL}/${BAZEL_VERSION}/${SCRIPT_NAME}.sha256"
+curl -sLO "${GITHUB_DL}/${BAZEL_VERSION}/${SCRIPT_NAME}"
+curl -sLO "${GITHUB_DL}/${BAZEL_VERSION}/${SCRIPT_NAME}.sha256"
 
 # We want to protect against accidents (i.e., we don't want to download and
 # execute a 404 page), not malice, so downloading the checksum and the file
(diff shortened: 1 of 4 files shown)


ci/install-bazel.sh
```diff
@@ -16,24 +16,26 @@
...
+  curl -sL "https://api.github.com/repos/bazelbuild/bazel/releases/latest" \
+    | grep -P '"tag_name":' \
+    | cut -f4 -d\"
+)
+test -n "${BAZEL_VERSION}"
 
-readonly BAZEL_VERSION=0.20.0
 readonly GITHUB_DL="https://github.com/bazelbuild/bazel/releases/download"
-readonly SCRIPT_NAME="bazel-${BAZEL_VERSION}-installer-${PLATFORM}.sh"
+readonly SCRIPT_NAME="bazel-${BAZEL_VERSION}-installer-${PLATFORM}-x86_64.sh"
```
devbww: $(uname -m)?

devjgm: I think that'd work. Although, bazel only has [x86_64 builds available](https://github.com/bazelbuild/bazel/releases), so do you think it's worth machine the machine hardware more dynamic like that?

devbww: I don't know, but I'm assuming that if you're not on an x86_64 that you don't want to run the x86_64 installer.

devjgm: Actually, I think it is better w/ `uname -m`. Done.

ci/install-bazel.sh
```diff
@@ -16,24 +16,25 @@
...
-if [[ "$1" = "linux" ]]; then
-  readonly PLATFORM="linux-x86_64"
-elif [[ "$1" = "macos" ]] || [[ "$1" = "osx" ]]; then
-  readonly PLATFORM="darwin-x86_64"
-fi
+# Gets the latest version number of bazel from the GitHub JSON API.
+readonly BAZEL_VERSION=$(
+  curl -sL "https://api.github.com/repos/bazelbuild/bazel/releases/latest" \
+    | grep '"tag_name":' \
+    | cut -f4 -d\"
```
remyabel: Why the API? Can this result in rate-limiting? If it does, an alternative may be to parse `https://github.com/bazelbuild/bazel/releases.atom` instead.

remyabel: Alternatively, you could get the redirect URL: `URL=$(curl -Ls -o /dev/null -w '%{url_effective}' https://github.com/bazelbuild/bazel/releases/latest/)`, then use parameter substitution to get the latest part of the URL (which will be the version number: `${URL##*/}`

devjgm: Good points. While I think the rate limit of 60 requests per hour may be enough for us, it may not be and that would lead to annoying flakes in our CI build. I like your suggestions about looking at the redirect URL, however, that also seems potentially flaky to me because GitHub's web pages are for humans not scripts, so I could imagine the way they do redirects (or not) to change, and again, that may lead to flakes. 

So, for now, I think the practical thing to do is go back to using a hard-coded bazel version in our CI tests like we were before. However, I updated the version to 0.23.2. I also left in the change that compute the PLATFORM from uname.

devjgm:
> > Reviewed 1 of 1 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2223)_ status: all files reviewed, 4 unresolved discussions (waiting on @devbww, @devjgm, and @remyabel)
> 
> _[ci/install-bazel.sh, line 19 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2223#-L_x0uiz0fDcu1PzT5aP:-L_x0uiz0fDcu1PzT5aQ:b-lyusg6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/477059964846c982c339209e065a6d82e92a9674/ci/install-bazel.sh#L19)):_
> 
> > ```shell
> > set -eu
> > 
> > if [[ $# -lt 1 ]]; then
> > ```
> 
> nit: do you think we should cleanup the scripts that call `ci/install-bazel.sh` to _not_ pass the OS name in?

yes, I do. I meant to, but forgot. Thanks for mentioning this. Done.

PR #2224, opened 2019-03-13
Implement simple asynchronous RPCs with futures.

coryan:
In this PR we introduce functions to create simple (unary) asynchronous
RPCs that return a future<StatusOr<T>> instead of taking a callback
parameter. Because the futures can also accept callbacks via `.then()`
these new functions will (soon) replace the callback-based functions,
without loss of functionality.

The implementation does not use any locking because I could not
reproduce the synchronization problems. If we find out there is a
problem we can introduce them again.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2224)

### DIFF
--- a/google/cloud/bigtable/completion_queue.h
+++ b/google/cloud/bigtable/completion_queue.h
@@ -17,6 +17,7 @@
 
 #include "google/cloud/bigtable/internal/completion_queue_impl.h"
 #include "google/cloud/future.h"
+#include "google/cloud/status_or.h"
 
 namespace google {
 namespace cloud {
@@ -74,6 +75,35 @@ class CompletionQueue {
   future<std::chrono::system_clock::time_point> MakeRelativeTimer(
       std::chrono::duration<Rep, Period> duration) {
     return MakeDeadlineTimer(std::chrono::system_clock::now() + duration);
+  }
+
+  /**
+   * Make an asynchronous unary RPC.
+   *
+   * @param async_call a callable to start the asynchronous RPC.
+   * @param request the contents of the request.
+   * @param context an initialized request context to make the call.
+   *
+   * @tparam AsyncCallType the type of @a async_call. It must meet be invocable
+   *     with `(grpc::ClientContext*,
+   *     requirements for `internal::CheckAsyncUnaryRpcSignature<>`.
+   * @tparam Request the type of the request parameter in the gRPC.
+   *
+   * @return a future that becomes satisfied when the operation completes.
+   */
+  template <
+      typename AsyncCallType, typename Request,
+      typename Sig = internal::AsyncCallResponseType<AsyncCallType, Request>,
+      typename Response = typename Sig::type,
+      typename std::enable_if<Sig::value, int>::type = 0>
+  future<StatusOr<Response>> MakeUnaryRpc(
+      AsyncCallType async_call, Request const& request,
+      std::unique_ptr<grpc::ClientContext> context) {
+    auto op =
+        std::make_shared<internal::AsyncUnaryRpcFuture<Request, Response>>();
+    void* tag = impl_->RegisterOperation(op);
+    op->Start(async_call, std::move(context), request, &impl_->cq(), tag);
+    return op->GetFuture();
   }
 
   //@{
--- a/google/cloud/bigtable/completion_queue.h
+++ b/google/cloud/bigtable/completion_queue.h
@@ -84,9 +84,14 @@ class CompletionQueue {
    * @param request the contents of the request.
    * @param context an initialized request context to make the call.
    *
-   * @tparam AsyncCallType the type of @a async_call. It must meet be invocable
-   *     with `(grpc::ClientContext*,
-   *     requirements for `internal::CheckAsyncUnaryRpcSignature<>`.
+   * @tparam AsyncCallType the type of @a async_call. It must be invocable with
+   *     `(grpc::ClientContext*, RequestType const&, grpc::CompletionQueue*)`.
+   *     Furthermore, it should return a
+   *     `std::unique_ptr<grpc::ClientAsyncResponseReaderInterface<Response>>>`.
+   *     These requirements are verified by
+   *     `internal::CheckAsyncUnaryRpcSignature<>`, and this function is
+   *     excluded from overload resolution if the parameters do not meet these
+   *     requirements.
    * @tparam Request the type of the request parameter in the gRPC.
    *
    * @return a future that becomes satisfied when the operation completes.
--- a/google/cloud/bigtable/completion_queue_test.cc
+++ b/google/cloud/bigtable/completion_queue_test.cc
@@ -137,6 +137,8 @@ TEST(CompletionQueueTest, AyncRpcSimple) {
   auto reader = google::cloud::internal::make_unique<ReaderType>();
   EXPECT_CALL(*reader, Finish(_, _, _))
       .WillOnce(Invoke([](btadmin::Table* table, grpc::Status* status, void*) {
+        // Initialize a value to make sure it is carried all the way back to
+        // the caller.
         table->set_name("fake/table/name/response");
         *status = grpc::Status(grpc::StatusCode::OK, "mocked-status");
       }));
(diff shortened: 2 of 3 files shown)


coryan:
@devbww this is the resurrected PR for CompletionQueue.  As we discussed offline, I tried to reproduce the synchronization problems, was unable to, and therefore removed the weird synchronization constructs.

google/cloud/bigtable/completion_queue_test.cc
```diff
@@ -166,6 +176,104 @@ TEST(CompletionQueueTest, AyncRpcSimple) {
...
+  MockClient client;
+
+  using ReaderType =
+      ::google::cloud::bigtable::testing::MockAsyncResponseReader<
+          btadmin::Table>;
+  auto reader = google::cloud::internal::make_unique<ReaderType>();
+  EXPECT_CALL(*reader, Finish(_, _, _))
+      .WillOnce(Invoke([](btadmin::Table* table, grpc::Status* status, void*) {
+        table->set_name("fake/table/name/response");
+        *status = grpc::Status(grpc::StatusCode::OK, "");
```
devbww: Do they have a special OK factory?

google/cloud/bigtable/completion_queue_test.cc
```diff
@@ -121,42 +126,47 @@ class MockClientAsyncReaderInterface
...
+            btadmin::Table>>(reader.get());
       }));
 
   auto impl = std::make_shared<testing::MockCompletionQueue>();
   bigtable::CompletionQueue cq(impl);
 
   // In this unit test we do not need to initialize the request parameter.
-  btproto::MutateRowRequest request;
+  btadmin::GetTableRequest request;
+  request.set_name("fake/table/name/request");
```
devbww: Is the comment on line 157 (and 206) now misleading?

google/cloud/bigtable/completion_queue_test.cc
coryan: Yup, used `grpc::Status::OK`.

google/cloud/bigtable/completion_queue_test.cc
coryan: Fixed.

coryan:
PTAL.

google/cloud/bigtable/completion_queue_test.cc
devbww: I think I see the 206 update but not the 157 one.

google/cloud/bigtable/completion_queue.h
```diff
@@ -76,6 +77,35 @@ class CompletionQueue {
...
+   *     requirements for `internal::CheckAsyncUnaryRpcSignature<>`.
+   * @tparam Request the type of the request parameter in the gRPC.
+   *
+   * @return a future that becomes satisfied when the operation completes.
+   */
+  template <
+      typename AsyncCallType, typename Request,
+      typename Sig = internal::AsyncCallResponseType<AsyncCallType, Request>,
+      typename Response = typename Sig::type,
+      typename std::enable_if<Sig::value, int>::type = 0>
```
devjgm: I was wondering what this enabler was doing. If `Sig` is an `AsyncCallResponseType`, it was not clear to me what `Sig::value` meant. It looks like the value member is inherited from `false_type` or `true_type`. So the `AsyncCallResponseType` struct is both a meta-"function" (returning the response type as `::type`) and also a meta-"predicate" setting `::value` to true or false depending on something.

I don't know if I've seen this style of metafunction that is also a predicate; I usually see one or the other. Are there advantages to doing this over relying on sfinae? For example, this template signature could be written as:

```cc
template <
    typename AsyncCallType, typename Request,
    typename Response = typename internal::AsyncCallResponseType<AsyncCallType, Request>::type>
```

if the metafunctions were written such that the primary template did not define `::type` when the signature is incorrect and therefore there is no type. Doing this kind of makes sense, since there really is no async response type when the wrong signature is used. For example:

```cc
template <typename ResponseType>
struct AsyncCallResponseTypeUnwrap {};  // No ::type member alias

template <typename ResponseType>
struct AsyncCallResponseTypeUnwrap<
    std::unique_ptr<grpc::ClientAsyncResponseReaderInterface<ResponseType>>> {
  using type = ResponseType;
};
```

No change is required, and I'm not arguing *for* this. Just curious to hear your thoughts about the pros-n-cons of each.

coryan: > Just curious to hear your thoughts about the pros-n-cons of each.

Just using SFINAE is simpler to write, of course.  With the predicate and the `std::enable_if<>` Clang produces nice error messages (or would, if I was good with names and AsyncCallResponseTypeUnwrap had a decent name):

```
  auto future = cq.MakeUnaryRpc(
                ~~~^~~~~~~~~~~~
/v/google/cloud/bigtable/completion_queue.h:104:30: note: candidate template ignored: requirement 'AsyncCallResponseTypeUnwrap<void>::value' was not satisfied [with AsyncCallType = (lambda at /v/google/cloud/bigtable/completion_queue_test.cc:218:7), Request = google::bigtable::admin::v2::GetTableRequest, Sig = google::cloud::bigtable::v0::internal::AsyncCallResponseTypeUnwrap<void>, Response = void]
  future<StatusOr<Response>> MakeUnaryRpc(
```

google/cloud/bigtable/completion_queue.h
```diff
@@ -76,6 +77,35 @@ class CompletionQueue {
...
   }
 
+  /**
+   * Make an asynchronous unary RPC.
+   *
+   * @param async_call a callable to start the asynchronous RPC.
+   * @param request the contents of the request.
+   * @param context an initialized request context to make the call.
+   *
+   * @tparam AsyncCallType the type of @a async_call. It must meet be invocable
```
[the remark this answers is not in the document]
coryan: Fixed.

google/cloud/bigtable/completion_queue.h
```diff
@@ -76,6 +77,35 @@ class CompletionQueue {
...
+   *     with `(grpc::ClientContext*,
```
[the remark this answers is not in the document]
coryan: Fixed.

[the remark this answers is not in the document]
coryan: Fixed.

coryan:
PTAL.

google/cloud/bigtable/completion_queue.h
devjgm: Gotcha. Thanks. There are already a bunch of other meta functions following this pattern, so I do not think you should change anything.

But for the sake of discussion and my own clarification, I'll reply w/ a followup comment/question.

I'm totally in favor of using std::enable_if to produce better error messages. I've seen this have the most improvement when the use of enable_if moves the sfinae error from the *body* of the template instantiation and into the template *parameter list*. In that case, using std::enable_if greatly improves the compiler error messages.

However, in this case, we're trading one sfinae error within the template parameter list for another sfinae error within the template parameter list. This does change the generated error slightly, but it's not clear to me whether it's an improvement or not.

Experimenting with https://godbolt.org/z/i-uBob, I see an error message difference like:

**Error generated using `std::enable_if<>`**
```
<source>:34:11: error: no matching function for call to 'Call'
  return Call<decltype(Bad)>();
         ^~~~~~~~~~~~~~~~~~~
<source>:28:5: note: candidate template ignored: requirement 'AsyncCallResponseTypeUnwrap<int>::value' was not satisfied [with F = int (), Sig = AsyncCallResponseTypeUnwrap<int>, Response = void]
int Call() { return 2; }
   ^
1 error generated.
Compiler returned: 1
```

**Error generated with vanilla SFINAE in the template parameter list**
```
<source>:34:11: error: no matching function for call to 'Call'
  return Call<decltype(Bad)>();
         ^~~~~~~~~~~~~~~~~~~
<source>:28:5: note: candidate template ignored: substitution failure [with F = int ()]: implicit instantiation of undefined template 'AsyncCallResponseTypeUnwrap<int>'
int Call() { return 2; }
   ^
1 error generated.
Compiler returned: 1
```

The error from the second is shorter, but it's hard to tell if either one is clearer.

Issue #2225, opened 2019-03-13
Disabling signals from curl


nickolasrossi:
### What component of `google-cloud-cpp` is this related to?
- Cloud Storage (under development).

### What version of `google-cloud-cpp` are you using? 
(master) abecda143b305f57853eb5d7b35154af5be1a5fa

### What compiler and version are you using?
clang++ 3.8.0-2ubuntu4

### What operating system and version are you using?
ubuntu 16.04

### What were you trying to do?

I have a server performing multithreaded downloads from GCS. (Each thread uses its own GCS Client instance.)

Often, when the server is idle for a couple minutes after performing some downloads, the process will crash with a segfault. The stack trace captured in gdb (pasted below) indicates the curl DNS resolver.

This is a longstanding problem with libcurl, due to its use of signals to time out DNS resolution. The recommended workaround is to either build libcurl with "enable-threaded-resolver" or to set CURLOPT_NOSIGNAL=1:

https://curl.haxx.se/libcurl/c/CURLOPT_NOSIGNAL.html

However, I don't see a clean way to do this without mucking around in curl_handle.cc...

Any guidance?

#0  __strcpy_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strcpy-sse2-unaligned.S:313
#1  0x000000000178744f in Curl_failf (data=0x7ffff06749bf, fmt=0x2c0bbf "name lookup timed out")
   at external/com_github_curl_curl/lib/sendf.c:262
#2  0x000000000176348d in Curl_resolv_timeout (conn=0x20,
   hostname=0x26d928f <constant_time_select_8+47> "\210\301\017\266\301H\203\304\020]\303f\017\037D", port=32765, entry=0x0,
   timeoutms=97) at external/com_github_curl_curl/lib/hostip.c:708
#3  0x00000000026d8a8e in conv_bin2ascii (a=32 ' ') at external/boringssl/src/crypto/base64/base64.c:97
#4  0x00000000026d8871 in EVP_EncodeBlock (dst=0x7ffd1c2e9768 "", src=0x7ffd247edacb "", src_len=16)
   at external/boringssl/src/crypto/base64/base64.c:229
#5  0x0000000001681650 in std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::resize (
   this=0x7ffd247eda90, __n=24) at external/llvm_toolchain/bin/../include/c++/v1/string:949
[...]

PR #2226, opened 2019-03-13
Disable signals in libcurl.

coryan:
TIL: libcurl uses signals to perform timeouts by default. This change
disables the signals for all the handles we create, per the
documentation on:

https://curl.haxx.se/libcurl/c/CURLOPT_NOSIGNAL.html

I will try to write a regression test later, though it might be hard
because it requires "waiting for a couple of minutes" to get the
signal.

In any case, given the documentation, and the fact that there is a bug
report (#2225) associated with this, I believe we should incorporate
these fixes.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2226)

### DIFF
--- a/google/cloud/storage/tests/curl_nosignal_integration_test.cc
+++ b/google/cloud/storage/tests/curl_nosignal_integration_test.cc
@@ -12,33 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "google/cloud/internal/random.h"
 #include "google/cloud/storage/client.h"
 #include "google/cloud/storage/list_objects_reader.h"
-#include "google/cloud/storage/testing/storage_integration_test.h"
-#include "google/cloud/testing_util/assert_ok.h"
-#include "google/cloud/testing_util/init_google_mock.h"
-#include <gmock/gmock.h>
 #include <chrono>
-#include <future>
 #include <fstream>
+#include <future>
 
-namespace google {
-namespace cloud {
-namespace storage {
-inline namespace STORAGE_CLIENT_NS {
 namespace {
 
-using ::google::cloud::storage::testing::TestPermanentFailure;
-using ::testing::ElementsAreArray;
-using ::testing::HasSubstr;
 using ms = std::chrono::milliseconds;
 
-/// Store the options captured from the command-line arguments.
-std::string FLAG_bucket_name;
-std::chrono::seconds FLAG_idle_duration;
-
-class CurlNoSignalIntegrationTest
-    : public google::cloud::storage::testing::StorageIntegrationTest {};
 
 void ConfigureWorkingResolver() {
   std::ofstream("/etc/resolv.conf") << R"""(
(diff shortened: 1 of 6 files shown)


coryan:
@nickolasrossi if you have a few minutes to spare: I added an integration test to this PR that I hoped would reproduce the problem. Alas! It does not.  Any suggestions on what to change would be great.

nickolasrossi:
Yeah, it’s tough to repro an intermittent problem... since it comes from a DNS timeout, I think you want to download from a host name that causes DNS to take a while, and set a small CURLOPT_TIMEOUT to make it time out faster.

google/cloud/storage/tests/curl_nosignal_integration_test.cc
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  auto bucket_name = FLAG_bucket_name;
+  StatusOr<Client> client = Client::CreateDefaultClient();
+  if (!client) {
+    return client.status();
+  }
+
+  std::random_device rd;
+  std::mt19937_64 gen(rd());
```
remyabel: It is generally considered better practice to not recreate `gen` multiple times due to being an expensive object with a lot of state, so that may inadvertedly affect your results. I would make these global.

remyabel:
> Yeah, it’s tough to repro an intermittent problem... since it comes from a DNS timeout, I think you want to download from a host name that causes DNS to take a while, and set a small CURLOPT_TIMEOUT to make it time out faster.

That's a good suggestion. What do you think about something similar to this? 

https://github.com/google/google-api-cpp-client/blob/d0bbe169d81a50936ec5fcea4e6dbcfb97303f13/src/googleapis/client/transport/curl_http_transport.cc#L241-L254

google/cloud/storage/tests/curl_nosignal_integration_test.cc
coryan: Yup, they are big, but they are also not thread safe.  I am trading off expensive initialization once per thread vs. extra mutexes N times on each thread.  I am happy with that trade off.

The real downside is the draining of the entropy on `std::random_device`.  But that in itself is not as bad as you might think on modern processors (whatever the ones with the `RND` instruction are called).  Though it is a problem on VMs, which have terribly low entropy for more or less obvious reasons.

In any case, this is a test that will be executed manually, most likely on my workstation, so maybe: ¯\_(ツ)_/¯

remyabel: > I am trading off expensive initialization once per thread vs. extra mutexes N times on each thread.

Should `thread_local` not suffice here? See https://github.com/lichray/randint/blob/master/random.h#L49. But since this is a one-off test it's not that big of a problem.

coryan:
I created a clean PR with just the fixes, closing this for now.  If/When I can create a regression test I will submit a separate PR.

PR #2230, opened 2019-03-14
Initial draft of "flags" for integration tests.

coryan:
The integration tests have been using this weird pattern where we store
the command-line flags as static member variables of the test
"Environment", and then we initialize these member variables in the
environment constructor.

This is a first draft of how we could stop doing that. Ideally we would
use a real command-line parsing library (absl::Flag is not a thing,
Boost.ProgramOptions is a depedency hassle).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2230)


coryan:
Note that this is an **intentionally** incomplete fix.  I want to gather opinions before proceeding to change all the existing integration tests.

google/cloud/storage/tests/bucket_integration_test.cc
```diff
@@ -29,41 +30,21 @@ using ::google::cloud::storage::testing::TestPermanentFailure;
...
+testing_util::SimpleFlag<std::string> FLAG_bucket_name;
+testing_util::SimpleFlag<std::string> FLAG_topic;
 
 class BucketIntegrationTest
     : public google::cloud::storage::testing::StorageIntegrationTest {
  protected:
   std::string MakeEntityName() {
     // We always use the viewers for the project because it is known to exist.
-    return "project-viewers-" + BucketTestEnvironment::project_id();
+    return "project-viewers-" + *FLAG_project_id;
```
remyabel: Is it safe to deference this without checking it first?

coryan:
___
*[google/cloud/storage/tests/bucket_integration_test.cc, line 42 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2230#-L_wcS_7llgqKJSbpdca-r1-42:-L_wza530JG1Ies5sayt:be9rbdi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a513b6cd8a4c99ac7b716ff61176e3f101d7f017/google/cloud/storage/tests/bucket_integration_test.cc#L42)):*
<details><summary><i>Previously, remyabel wrote…</i></summary><blockquote>

Is it safe to deference this without checking it first? 
</blockquote></details>

Nope.  We could make `.value()` raise and `operator*` be UB.  This is why I am looking for early feedback. Thanks!

google/cloud/storage/tests/bucket_integration_test.cc
```diff
@@ -29,41 +30,21 @@ using ::google::cloud::storage::testing::TestPermanentFailure;
...
-  static std::string bucket_name_;
-  static std::string topic_;
-};
-
-std::string BucketTestEnvironment::project_id_;
-std::string BucketTestEnvironment::bucket_name_;
-std::string BucketTestEnvironment::topic_;
+testing_util::SimpleFlag<std::string> FLAG_project_id;
+testing_util::SimpleFlag<std::string> FLAG_bucket_name;
+testing_util::SimpleFlag<std::string> FLAG_topic;
```
devjgm: This seems like it may be a tad more than we need. One day if/when an `absl::Flag` exists, and if/when we are depending on absl, we could use that if we want. Until then, it seems like we maybe could just use some namespace-scope pointers. Something like:

```cc
namespace {
// These are initialized in main()
std::string const* kProjectid;
std::string const* kBucketName;
std::string const* kTopic;
}  // namespace
...
// uses pointers in tests
int main(int argc, char* argv[]) {
  ...
  kProjectId = new std::string(argv[1]);
  kBucketName = new std::string(argv[2])
  kTopic = new std::string(argv[3]);
  return RUN_ALL_TESTS();
}
```

Pros:
- Fairly vanilla code; easy to follow understand
- Nothing new to be written

Cons:
- Uses naked `new`; and no delete (though it's fine in this case)
- Uses "constant" pointers that are not really constant, so it's a little funny that way

Call site usage would look pretty much like this code in this PR: `*kProjectId`.

Just proposing this as an alternative.

Thoughts?

coryan: I think you are right that it is overkill.  What do you think of this new version?

coryan:
I got to run do an interview, just leaving this code here...

google/cloud/storage/tests/bucket_integration_test.cc
devjgm: My only concern with namespace-scope `optional<std::string>` is that it's not trivially destructible since `std::string` is not trivially destructible. And the [style guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables) says that namespace scope objects are forbidden unless they are trivially destructible. That's the (only) reason I originally suggested pointers, which are trivially destructible.

In general, I think the style guide's advice here is good. However, maybe it's the case that we  know enough about our test environment to know that destruction is irrelevant. Like, maybe if our tests never have any unjoined/detached threads, then it may be fine (?). But if we're deciding to eschew the style guide's ruling here, we can consider simply using a namespace scope `std::string` directly (no optional). That would make the call sites simpler by removing the need to dereference.

Another option is to avoid the `new` and `std::string`s altogether, and just have namespace-scope `char const* kProjectid;` variables. They would refer directly into the `argv` c-strings, and would/should be implicitly converted to `std::string` when needed in test code. This would also require no dereferencing at the call sites.

coryan: I think it is either `std::string` or `char const*`.  Using `std::string const*` or `std::string*` would obviously leak resources, which is terrible habit to get into: you need to learn to ignore some leaks but not others when using any dynamic analysis tool.

I *prefer* `std::string`, because I think the style-guide advice is not applicable in this case. Yes we know enough about the environment to assert that there are no detached threads running around like (or we can treat any such as bugs).  We also know that initialization order is a non-issue (these variables are in an anonymous namespace in the same translation unit as `main()`).  In other words, none of the "cons" in the style guide are applicable.

If I did not convince you, then `char const*` is the way to go.

devbww: My choice would be for `char const*` (embedded NULs can't exist, it looks like we only create `std::string`s later with them anyway, and style guide without "we know"s), but I'm happy either way.

devjgm: > would obviously leak resource

Well, yes and no. Yes, if you mean that we never call delete. But there will always be a pointer to the allocated object, so it's not really a leak and leak detectors and sanitizers will not flag it as a leak. It's just an object that we haven't yet deleted. *shrug* :-)

Using `const char*` is fine with me. It has the nice property that it doesn't require any dereference at the call site, so that's somewhat nice.

remyabel: > But there will always be a pointer to the allocated object, so it's not really a leak and leak detectors and sanitizers will not flag it as a leak.

If there's no longer a pointer to the object, isn't that dangling? The Address Sanitizer *does* complain about this leak. I agree it's a harmless leak since it's a trivial amount of memory and leaks are more of a problem in long running programs, not short ones. But @coryan's point about having to ignore this stands.

devjgm: Yes, **if** there's no longer a pointer the object, then that is a leak. But that is not the situation here. In this case, there *is* always a valid pointer to the object, so it's not a leak. And since there's a valid pointer, sanitizers should not flag it as a leak. 

```console
$ cat -n noleak.cc 
     1	#include <string>
     2	std::string const* x;
     3	int main() {
     4	  x = new std::string("No leak");
     5	}
$ g++ -o noleak noleak.cc -fsanitize=address
$ ./noleak  # No leak detected
$ echo $?
0 
```

This is a not uncommon way to avoid destruction problems in C++. So, I don't personally interpret this situation as ignoring leaks or similar, rather I just view this as a weird pattern that's sometimes needed in C++ (which it's sadly full of these corner cases).

remyabel: Thank you that's good to know.

google/cloud/storage/tests/bucket_integration_test.cc
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "google/cloud/optional.h"
```
devbww: Don't need this addition any longer?

coryan: Nope, will fix in #2235.

PR #2238, opened 2019-03-15
Update gRPC in Dockerfiles and install instructions

remyabel:
There were places where `1.17.x` was still being used. Note that it's important we say `1.19.1` and not just `1.19.x` because `1.19.1` contains a vital fix.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2238)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -30,7 +30,7 @@ platforms.
 
 | Library | Minimum version | Description |
 | ------- | --------------: | ----------- |
-| gRPC    | 1.19.1 | gRPC++ for Cloud Bigtable |
+| gRPC    | 1.16.x | gRPC++ for Cloud Bigtable |
 | libcurl | 7.47.0  | HTTP client library for the Google Cloud Storage client |
 | crc32c  | 1.0.6 | Hardware-accelerated CRC32C implementation |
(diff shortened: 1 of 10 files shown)


coryan:
Can you tell me what is the critical fix in 1.19.1?  Are you referring to the Bazel fixes?

In general, I am in favor of updating all the Dockerfiles to say 1.19.1.

I am not sure we should say that we require 1.19.1.  The code still compiles with gRPC-1.16.x:

https://github.com/googleapis/google-cloud-cpp/blob/273ecad7131c2bd1b8a7f21601568e506eb223aa/ci/test-readme/Dockerfile.opensuse#L47

and

https://build.opensuse.org/package/show/openSUSE:Factory/grpc

Maybe it won't compile from source with the latest Bazel, but does work, and compiles, with CMake or the previous versions of Bazel.

I would rather keep the requirements to 1.17.x (actually it should probably say 1.16.x) and update the build instructions to 1.19.1.  I know this is complicated, but I am not sure how to simplify the messaging here.

coryan:
On a separate topic, should we fix this as part of this PR on a separate one?

https://github.com/googleapis/google-cloud-cpp/blob/273ecad7131c2bd1b8a7f21601568e506eb223aa/ci/travis/Dockerfile.ubuntu-install#L143

PR #2250, opened 2019-03-18
Create signed URL V4 conformance tests

### LINKED ISSUE #2173: Create signed URL V4 conformance tests.
We want all the libraries for Google Cloud Storage to implement the same unit tests.  Essentially this is complicated code and we want some certainty that all our implementations are correct.

There is a series of "tests" expressed as a JSON array here:

https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json

All these tests use the same service account (which is inactive):

https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestAccount.json

The goal of this task is to read the JSON file, pass the parameters to `Client::CreateV4SignedUrl()` and verify that the signature matches the expected value in the JSON structure.

### LINKED ISSUE #2350: client.CreateV4SignedUrl() does not work when object_name == ""
In storage/tests/client_sign_url_integration_test.cc
    client.CreateV4SignedUrl() does not work when object_name == ""

The actual url is different than the expected url.
```
EXPECT_EQ(expected, *actual); 
```
fails.
Related to #2250

rsimion:
Fixes #2173 and fixes #2350.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2250)


rsimion:
@coryan 

**Is there any reason to only test one verb per-function? Seems like the code below does not care about the verb at all.**
If it finds the method, then it breaks out of the for loop via return. We should probably use a different way to differentiate between all those objects


**You are ignoring the headers field on jObj. Do the tests even pass? In any case, we should use the headers via AddExtensionHeader... which I just realized might be hard.**
V4SignPost uses headers



**Why is this called V4SignPut?**
I meant V4SignPost. On my local I also added a V4SignPut but only on my local.


**It would be useful to print the number or even the contents of jObj on failure.**
Something like 
if(expected != *actual){
	std::cout << j_obj.dump(4) << '\n';
	}
?

or is there any example to follow?

rsimion:
@coryan 

**Note:**
There are 2 issues with this json library

1. it does not parse comments
https://github.com/nlohmann/json/issues/294

2. it does not parse the last comma(if any)
https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json

I removed all the comments and the last comma from those 2 files.
Also for the timestamp I used 
"timestamp": "2019-02-01T09:00:00Z"

instead of 
"timestamp": "20190201T090000Z"

coryan:
Can you script the work to remove the comments, and trailing commas?  Or you can load the file into a string and do it all in C++, both are fine, but manually changing the file gets boring pretty fast.

Same thing with the date format. Can you either script the change or modify the program to convert the format to something we know how to parse?

coryan:
___
*[google/cloud/storage/tests/client_sign_url_integration_test.cc, line 87 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2250#-LaHPSKk5jOafr5xFIxh:-LaHPSKk5jOafr5xFIxi:bxi9pff) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a66e66f79ebc39a4dfe3ede0cbe80a5eb8987fc8/google/cloud/storage/tests/client_sign_url_integration_test.cc#L87)):*
> ```cpp
>       std::string const expected = j_obj["expectedUrl"];
>       EXPECT_EQ(expected, *actual);
>       return;
> ```

Oh, I just noticed that.  We want to run *all* the tests in the file, not just the first one. Please remove this line.

rsimion:
I will try to load the file into a string and do it all in C++. 
Same for the date format.

rsimion:
@coryan 
Is that json file reliable?
https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json

 because for  the first two GET with only the expiration and description  being different, the expectedUrl is different, that is why I added the description as a filter for now.

Also some of the Get have headers so we have to account for them.

rsimion:
@coryan 

I rushed. My mistake.

rsimion:
@coryan

I broke it up into 3 functions . I can combine them later.
Functions with headers
Functions without headers
Special function with headers : "Customer-supplied encryption key"

I put comments all over the places.
There is also a function that does not work and is commented out :  "List Objects"
but there is an explanation for that.

There are 2 functions there are not supposed to work, or I do not know how to pass the arguments 
 "Multi-value headers"
 "Trimming of multiple header values"
 
 https://github.com/googleapis/google-cloud-dotnet/blob/e918df5a988f53ed71cebf708a0dd06bed8bef43/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json#L42
 
I left a std::cout in each function to make it easy to understand when checking the json file.
Now I am working on loading the file into a string then turn it into a json object and change the date format.

coryan:
___
*[google/cloud/storage/tests/client_sign_url_integration_test.cc, line 321 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2250#-LaMflVa07VCZSuqMgsE:-LaMflVa07VCZSuqMgsF:b9vraxy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/838663c19557e9400f18519ee44c526afcfc1dfb/google/cloud/storage/tests/client_sign_url_integration_test.cc#L321)):*
> ```cpp
> 
> /*
> //This does not work the object is empty
> ```

What exactly does not work?  Does our code return an error?  The generated URL does not match?

rsimion:
yes the generated URL does not match the expected one for that object

coryan:
Okay, let's consolidate the code to a single loop, and skip the `List Objects` case for now.

rsimion:
@coryan
Left std::cout in for now.

rsimion:
@coryan
Changed the approach. I wrote the function to extract the headers from each json obj. 

There are 2 types of headers:
with  key : single value
 and
with key : multi value  

Another problem with this library is that the keys in any object are being outputted as if coming out from a map(in alphabetical order not the order they are in the file)

https://github.com/nlohmann/json/issues/106
https://github.com/nlohmann/json/issues/262

About the python json5 parser,  I am looking into it.

rsimion:
@coryan 

**I think we should consider doing the python script approach, download the file and apply a JSON5 to JSON**

What file?

I am not sure I understand what you mean.
I ran the one-liner in cli and in a bash script but how do I pass it to c++?

Do you mean a python script that moves a clean json into another file and then read it in C++?

coryan:
> @coryan
> 
> **I think we should consider doing the python script approach, download the file and apply a JSON5 to JSON**
> 
> What file?

https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json

> 
> I am not sure I understand what you mean.
> I ran the one-liner in cli and in a bash script but how do I pass it to c++?

Save the one-liner CLI into a program much like the one below:

https://github.com/googleapis/google-cloud-cpp/blob/f7f7bc509b7f3f68c980f78d70ccf4aad6da4b5b/google/cloud/bigtable/tools/convert_acceptance_tests.py#L28

include in the program the instructions to fetch the `UrlSignerV4TestData.json` file, and how to use the one-liner to transform the file.

> 
> Do you mean a python script that moves a clean json into another file and then read it in C++?

See above.

> One suggestion, why not use a clean json file, do the C++ testing and change the issue's title
> and then open another issue to strip off the comments and trailling commas from the json file...

Not a bad idea. Let me summon @frankyn: can we cleanup the [UrlSignerV4TestData.json](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json) to be standard JSON (aka [RFC 7159](https://tools.ietf.org/html/rfc7159)) as opposed to JSON5 (https://json5.org/)? Our JSON parser is strict, and rejects comments or trailing commas, we are hacking stuff that maybe is simpler to fix in the source file.

frankyn:
Thanks for raising the question @coryan. I'll send a PR to address comments raised in this PR along with another related change.

coryan:
Actually it was @rsimion who raised this question.  I just happen to know a tiny bit more about where to redirect the question.

frankyn:
@rsimion will the change in https://github.com/googleapis/google-cloud-dotnet/pull/2974 help unblock you? 

I do have a follow-up question: I was not able to find a trailing comma. Could you tell which line you're seeing it on?

coryan:
I think the trailing comma was removed here:

https://github.com/googleapis/google-cloud-dotnet/commit/9532321f52d5e0d7e7b16b27c13cb2da7961c06d#diff-02dbf7042353c4ea8f71a1065791b2da

frankyn:
Thanks!

rsimion:
@frankyn 
sorry,  are we all good?

coryan:
@rsimion there is a pending PR to remove the comments: googleapis/google-cloud-dotnet#2974, and newer versions of the UrlSignerV4TestData.json do not have trailing commas.  Feel free to update this PR with a newer version of that file, and manually strip the comments if you have to.

There are still pending changes from my review on the test itself, right?

rsimion:
@coryan 
ok, I will update the json file.

No, I have merged headers with non-headers into one big function, and  extracted every json obj like you asked.
 The problem is that  the json library outputs the data(keys in headers) in a different order than the order in the file meaning, 

In the new file  client_sign_url_integration_test.cc
Look at the lines 275, first is 1 then 0 but in the json file is the other way 0 then 1
```
            AddExtensionHeader(headers.at(1).first, headers.at(1).second.at(0)),
            AddExtensionHeader(headers.at(0).first,
                               headers.at(0).second.at(0)));
```

 and 296
1,2,0 vs json file 0,1,2
```
            AddExtensionHeader(headers.at(1).first, headers.at(1).second.at(0)),
            AddExtensionHeader(headers.at(2).first, headers.at(2).second.at(0)),
            AddExtensionHeader(headers.at(0).first,
                               headers.at(0).second.at(0)));
```

 in the new file  client_sign_url_integration_test.cc.

rsimion:
@coryan
Maybe on Wednesday we can talk for about 5 minutes after the meeting is over.

I replaced the json file with

https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json

Right now everything works(at least on my local) for headers like 
```
"Headers should be trimmed"
        "headers": {
            "leading": [ "    xyz" ],
            "trailing": [ "abc    " ],
            "collapsed": [ "abc    def" ]
        },
```		
with only one value per key. I added another function to check for headers only, using a different approach.	

There are several objects for example that have more than one value per key:
```
"Multi-value headers"

        "headers": {
            "foo": [ "foo-value1", "foo-value2" ],
            "bar": [ "bar-value1", "bar-value2" ]
        },
```		
and 
```
"Trimming of multiple header values"
      "headers": {
            "foo": [ "  abc  ", "  def  ", "  ghi  jkl  " ]
        },
```

I am not sure how they are supposed to work for us. I tried concatenating their values and it did not work.

That is why I used
```
std::vector<std::pair<std::string, std::vector<std::string>>>
```
 instead of 
```
std::vector<std::pair<std::string, std::string>>
```
because 
headers.front().second.size() == 1 allows me to filter out the all the headers with more than one value per key(see above)

Maybe on Wednesday we can talk for about 5 minutes after the meeting is over.

frankyn:
@rsimion short update:
We (me and @jskeet) discussed the multi-valued headers and @jskeet will send out a PR tomorrow to update them to simple key/value pairs (String/String). 

Thank you for your patience.

frankyn:
@rsimion changes were made to conformance tests and are now merged into master. Please move forward.

Thank you for your patience.

rsimion:
@frankyn 
thanks

rsimion:
@coryan 
you were right the order does not matter, my mistake.

rsimion:
@coryan 
Created a new issue #2350 
to address the bug when the object_name in  CreateV4SignedURL() is empty.

rsimion:
@coryan 
After you review it, please take a look at what @remyabel  found.

remyabel:
@rsimion #2350 should be fixed now. Please use the updated test data: https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json

rsimion:
@coryan
#2350 is fixed now thanks to @remyabel 
I have updated both files. I am testing now and I will push my changes soon.

PR #2255, opened 2019-03-19
Remove third_party/googleapis submodule.

### LINKED ISSUE #1248: Simplify CMake dependency management.
### Overview

The configuration of dependencies for CMake is getting out of control, we need to simply it.  I am going to propose that we support only two configurations:

- `external`: all the dependencies are compiled from source, obtained using `ExternalProject`. This would be the preferred build configuration, we would use this most of the time ourselves.  It is the same approach we use for the Bazel builds.
- `package`: all the dependencies are assumed to be installed and discovered using `find_package()`.  This would work for both `vcpkg`-based builds and builds that install `google-cloud-cpp`.

#### New external dependencies

We would need to introduce `external` packages for some of our current dependencies:

- `libcurl` is currently supported only as a `find_package()` module.
- `OpenSSL` (or more likely `BoringSSL`) is needed by `libcurl`, `gRPC` and `google/cloud/storage`.
- `zlib` is needed by `libcurl` and `protobuf`.

#### Simplify configuration

We should introduce a single configuration option `GOOGLE_CLOUD_CPP_DEPENDENCY_PROVIDER` that sets the source for all the modules.

#### Handling Differences

Some dependencies create different targets when compiled using `ExternalProject` vs. discovered via `find_package`.  For example, `googletest` creates `GTest::gtest_main` when compiled from source, but the CMake module creates `GTest::Main`.  We will create aliases for such cases and prefer the names created by the `ExternalProject`.

Sometimes the results of using `vcpkg + find_package()` vs. `find_package()` on a Linux installation are slightly different.  In those cases we will normalize the targets created by `vcpkg + find_package()` to be the same results as `ExternalProject`.  Again, googletest is the main culprit on this area.

#### Exceptions

OpenSSL is very hard to manage and an `ExternalProject`, it does not have a CMake file, and in fact requires Perl to compile.  BoringSSL cannot be installed, so it is not usable as an external project.  We propose to keep using OpenSSL as an external dependency.

### Alternatives Considered

**Do nothing:** we already have too many configurations we do not test, this is not viable.

**Only Support one of the configurations:** this is not tenable, both `package` and `external` are needed because:
* Supporting `find_package()` is good hygiene, it is a necessary configuration to support any package manager, and it is expected that some of our users will require it.
* Supporting `ExternalProject` makes it easier to start using the library (nothing to install before you try it), it makes it easier to develop, and makes the CMake and Bazel build more homogeneous.

**What happened to pkg-config:** sometimes the OS does not install the CMake configuration files and the only way to discover the dependencies is to use `pkg-config(1)`.  We have "supported" pkg-config builds in the past, but they are not tested and are very brittle.

### LINKED ISSUE #1989: Consider removing third_party/googleapis submodule
Sorry if this issue is already filed and I couldn't find it...

We should consider removing the third_party/googleapis git submodule. It's the only submodule that we currently use. It's only used by the cmake build, since (I think) bazel will fetch its deps on its own.

It may simplify the build strategy for this project if we cmake and bazel were a bit more similar, and if their of them required a git submodule.

If we decide not to do this, we can record our decision and rationale here.

### LINKED ISSUE #2246: Tarballs are missing submodule contents
Build failure on Fedora 29:

```
wget https://github.com/googleapis/google-cloud-cpp/archive/v0.7.0.tar.gz
tar zxvf v0.7.0.tar.gz
cd google-cloud-cpp-0.7.0/
cmake -H. -Bbuild-output \
    -DGOOGLE_CLOUD_CPP_DEPENDENCY_PROVIDER=package \
    -DGOOGLE_CLOUD_CPP_PROTOBUF_PROVIDER=pkg-config \
    -DGOOGLE_CLOUD_CPP_GRPC_PROVIDER=pkg-config \
    -DGOOGLE_CLOUD_CPP_GMOCK_PROVIDER=external
cmake --build build-output -- -j $(nproc)
```

Then:

```
gmake[2]: *** No rule to make target '../third_party/googleapis/google/longrunning/operations.proto', needed by 'google/cloud/bigtable/google/longrunning/operations.grpc.pb.cc'.  Stop.
gmake[1]: *** [CMakeFiles/Makefile2:3682: google/cloud/bigtable/CMakeFiles/bigtable_protos.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
Scanning dependencies of target nlohmann_json_project
Scanning dependencies of target googletest_project
Scanning dependencies of target google_cloud_cpp_common
[  0%] Creating directories for 'nlohmann_json_project'
[  0%] Creating directories for 'googletest_project'
[  0%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/iam_policy.cc.o
[  0%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/iam_bindings.cc.o
[  0%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/internal/backoff_policy.cc.o
[  0%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/internal/build_info.cc.o
[  1%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/internal/filesystem.cc.o
[  1%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/internal/future_impl.cc.o
[  1%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/internal/getenv.cc.o
[  1%] Performing download step for 'nlohmann_json_project'
[  1%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/internal/setenv.cc.o
[  1%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/internal/random.cc.o
[  1%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/internal/throw_delegate.cc.o
[  2%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/log.cc.o
[  2%] Performing download step (download, verify and extract) for 'googletest_project'
[  2%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/status.cc.o
[  2%] Building CXX object google/cloud/CMakeFiles/google_cloud_cpp_common.dir/terminate_handler.cc.o
[  2%] Linking CXX static library libgoogle_cloud_cpp_common.a
[  2%] Built target google_cloud_cpp_common
-- googletest_project download command succeeded.  See also /tmp/google-cloud-cpp-0.7.0/build-output/external/googletest/src/googletest_project-stamp/googletest_project-download-*.log
[  2%] No patch step for 'googletest_project'
[  2%] No update step for 'googletest_project'
[  2%] Performing configure step for 'googletest_project'
-- nlohmann_json_project download command succeeded.  See also /tmp/google-cloud-cpp-0.7.0/build-output/external/nlohmann_json/src/nlohmann_json_project-stamp/nlohmann_json_project-download-*.log
[  3%] No update step for 'nlohmann_json_project'
[  3%] No patch step for 'nlohmann_json_project'
[  3%] No configure step for 'nlohmann_json_project'
[  3%] Performing build step for 'nlohmann_json_project'
[  3%] Performing install step for 'nlohmann_json_project'
-- nlohmann_json_project install command succeeded.  See also /tmp/google-cloud-cpp-0.7.0/build-output/external/nlohmann_json/src/nlohmann_json_project-stamp/nlohmann_json_project-install-*.log
[  3%] Completed 'nlohmann_json_project'
[  3%] Built target nlohmann_json_project
-- googletest_project configure command succeeded.  See also /tmp/google-cloud-cpp-0.7.0/build-output/external/googletest/src/googletest_project-stamp/googletest_project-configure-*.log
[  4%] Performing build step for 'googletest_project'
-- googletest_project build command succeeded.  See also /tmp/google-cloud-cpp-0.7.0/build-output/external/googletest/src/googletest_project-stamp/googletest_project-build-*.log
[  5%] Performing install step for 'googletest_project'
-- googletest_project install command succeeded.  See also /tmp/google-cloud-cpp-0.7.0/build-output/external/googletest/src/googletest_project-stamp/googletest_project-install-*.log
[  5%] Completed 'googletest_project'
[  5%] Built target googletest_project
gmake: *** [Makefile:141: all] Error 2
```

This would be useful for anyone building from the tarball (i.e, mock users).

coryan:
Remove the third_party/googleapis submodule, replace it with an external
project. The external project is compiled, more or less, as usual:

- CMake downloads the code.
- CMake calls CMake to configure the external project.
- CMake calls CMake to build the external project.
- CMake calls CMake to install the external project.

The interesting differences are:

- After downloading the code we copy a `CMakeLists.txt` file and some support modules because googleapis does not include build scripts (other than Bazel `BUILD` files).
- The googleapis external project creates (right now) a single `googleapis-c++::bigtable_protos` target, leaving room for future targets such as `googleapis-c++::spanner_protos`.
- The installed pkg-config files manage dependencies a little bit better, but fully fixing that is for a future PR (this is is too big already).

This fixes #1989, fixes #2246, and (I think) fixes #1248.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2255)


coryan:
Building with vcpkg (or any other package manager) won't work.  Back to the drawing board.

coryan:
The tests did pass. Ship it!

PR #2258, opened 2019-03-19
Implement TableAdmin::AsyncDeleteTable().

coryan:
This implements `TableAdmin::AsyncDeleteTable()`, the integration test,
and an example. No unit test is needed because that would just duplicate
the existing unit tests for `internal::AsyncRetryUnaryRpc`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2258)

### DIFF
--- a/google/cloud/bigtable/examples/table_admin_async_snippets.cc
+++ b/google/cloud/bigtable/examples/table_admin_async_snippets.cc
@@ -109,6 +109,33 @@ void AsyncGetTable(cbt::TableAdmin admin, cbt::CompletionQueue cq,
   (std::move(admin), std::move(cq), argv[1]);
 }
 
+void AsyncDeleteTable(cbt::TableAdmin admin, cbt::CompletionQueue cq,
+                      std::vector<std::string> argv) {
+  if (argv.size() != 2U) {
+    throw Usage{"async-delete-table: <project-id> <instance-id> <table-id>"};
+  }
+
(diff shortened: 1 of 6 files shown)


google/cloud/bigtable/examples/table_admin_async_snippets.cc
```diff
@@ -109,6 +109,33 @@ void AsyncGetTable(cbt::TableAdmin admin, cbt::CompletionQueue cq,
...
+  //! [async delete table]
+  namespace cbt = google::cloud::bigtable;
+  [](cbt::TableAdmin admin, cbt::CompletionQueue cq, std::string table_id) {
+    google::cloud::future<google::cloud::Status> future =
+        admin.AsyncDeleteTable(cq, table_id);
+
+    auto final =
+        future.then([table_id](google::cloud::future<google::cloud::Status> f) {
+          auto status = f.get();
+          if (!status.ok()) {
```
devbww: s/status.ok()/status/

coryan: I do not believe we have a `operator bool()` for `google::cloud::Status` which would be the type here, right?

devbww: Right you are ... I was thinking of `StatusOr`.

PR #2259, opened 2019-03-19
Do not construct an OK grpc::Status with an error message

devbww:
grpc::Status says, "It is an error to construct an OK status
with [a] non-empty error_message," so use grpc::Status::OK instead.
Similarly, don't ask for status.error_message() when status.ok().

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2259)


devbww:
Fixed.  Thanks.  PTAL.

PR #2261, opened 2019-03-20
Break from stream loop early if we encounter an error

### LINKED ISSUE #2144: Consider not checking for eof when reading from streams
There are several places where this happens, i.e https://github.com/googleapis/google-cloud-cpp/blob/51e147b7c10dab5839fb65a3486ae0438ff8e4d1/google/cloud/storage/benchmarks/storage_latency_benchmark.cc#L306 and https://github.com/googleapis/google-cloud-cpp/blob/5244e73ccf104837e21bd52c419b33b45ee64a92/google/cloud/storage/client_options.cc#L105. This is considered [bad practice](https://stackoverflow.com/questions/5605125/why-is-iostreameof-inside-a-loop-condition-considered-wrong) as it is preferred to instead check the state of the stream while looping, i.e:

```c++
int data;
while(inStream >> data){
  // when we land here, we can be sure that the read was successful.
  // if it wasn't, the returned stream from operator>> would be converted to false
  // and the loop wouldn't even be entered
  // do stuff with correctly initialized data (hopefully)
}
```

instead of:

```c++
while(!inStream.eof()){
  int data;
  // yay, not end of stream yet, now read ...
  inStream >> data;
  // oh crap, now we read the end and *only* now the eof bit will be set (as well as the fail bit)
  // do stuff with (now uninitialized) data
}
```

If the above is acceptable, then `ObjectReadStream` will need to be updated accordingly.

remyabel:
Fixes #2144

Consider the following. Let's say we are looping over a stream and checking for `eof()`:

```c++
std::string message("hi there i am a message whoo hoo\nhi there i am a message whoo hoo\nhi there i am a message whoo hoo\nhi there i am a message whoo hoo\n");
std::vector<std::string> log_lines;
std::istringstream is(message);

while (!is.eof()) {
  std::string line;
  std::getline(is, line);
  log_lines.emplace_back(std::move(line));
}
```

This will output:

```
hi there i am a message whoo hoo|hi there i am a message whoo hoo|hi there i am a message whoo hoo|hi there i am a message whoo hoo||
```

Notice the extra blank entry at the end. It is incorrect to loop on `eof()` because if an error in the stream occurred, we will do an extra read on an invalid stream before the loop ends. One approach is to refactor the loops but this can result in [performance degradation](http://quick-bench.com/hY0cwQzDf7F3cFc-2IlA0JUDf9I). Instead, we just check the stream state inside the loop and break early.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2261)

### DIFF
--- a/google/cloud/examples/gcs2cbt.cc
+++ b/google/cloud/examples/gcs2cbt.cc
@@ -178,12 +178,9 @@ int main(int argc, char* argv[]) try {
           std::cerr << "Apply failed: " << status;
         }
       };
-  while (!is.eof()) {
+  while (std::getline(is, line, '\n')) {
     ++lineno;
-    std::getline(is, line, '\n');
-    if (!is) {
-      break;
-    }
+
     if (line.empty()) {
       break;
     }
@@ -252,13 +249,9 @@ std::vector<std::string> ParseLine(long lineno, std::string const& line,
 
   // Extract the fields one at a time using a std::istringstream.
   std::istringstream tokens(line);
-  while (!tokens.eof()) {
-    std::string tk;
-    std::getline(tokens, tk, separator);
-    if (!tokens) {
-      break;
-    }
-    result.emplace_back(std::move(tk));
+  std::string tk;
+  while (std::getline(tokens, tk, separator)) {
+    result.emplace_back(tk);
   }
   return result;
 } catch (std::exception const& ex) {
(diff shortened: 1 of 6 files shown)


remyabel:
FYI, there is one more instance I didn't tackle in this PR because it is a little bit more complicated to refactor, so I would like to leave it to someone else: https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/storage/client.cc#L125.

PR #2269, opened 2019-03-20
Write a custom module for find protobuf.

coryan:
Finding protobuf is a mess (see the documentation in the module for
details), but with this module we can just type:

find_package(ProtobufTargets REQUIRED)

and it works. In the future we will use this module to remove the
googleapis submodule, and we will install it so application developers
can benefit from a simpler way to discover protos.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2269)


cmake/FindProtobufTargets.cmake
```diff
@@ -0,0 +1,206 @@
...
+A module to use `Protobuf` with less complications.
+
+Using ``find_package(Protobuf)`` should be simple, but it is not.
+
+CMake provides a ``FindProtobuf`` module. Unfortunately it does not generate
+`protobuf::*` targets until CMake-3.9, and `protobuf::protoc` does not
+appear until CMake-3.10.
+
+The CMake-config files generated by `protobuf` always create these targets,
+but on some Linux distributions (e.g. Fedora>=29, and openSUSE-Tumbleweed) the
```
devbww: s/the/there/

cmake/FindProtobufTargets.cmake
```diff
@@ -0,0 +1,206 @@
...
+are system packages for protobuf, but these packages are installed without the
+CMake-config files. One must either use the ``FindProtobuf`` module, find the
+libraries via `pkg-config`, or find the libraries manually.
+
+When the CMake-config files installed they produce the same targets as recent
```
devbww: s/installed/are installed/

cmake/FindProtobufTargets.cmake
coryan: Fixed.

cmake/FindProtobufTargets.cmake
coryan: Fixed.

coryan:
PTAL.

PR #2279, opened 2019-03-21
Pass CMake flags if we disable testing

remyabel:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2279)


coryan:
+kokoro:run

Issue #2015, opened 2019-02-16
Flaky test: DataIntegrationTest.TableReadMultipleCellsBigValue


coryan:
One of the tests takes about 10min to run.  It is only enabled against the production environment, so it does not affect our regular builds, but this is clearly too slow.

```
[ RUN      ] DataIntegrationTest.TableReadMultipleCellsBigValue
unknown file: Failure
C++ exception with description "Deadline Exceeded" thrown in the test body.
[  FAILED  ] DataIntegrationTest.TableReadMultipleCellsBigValue (661966 ms)
```

coryan:
Actually in most builds it only takes 10 seconds to run, e.g.:

https://source.cloud.google.com/results/invocations/1ee5991a-1488-4d79-8b07-eb2d39f863ce/targets

```
[ RUN      ] DataIntegrationTest.TableCellValueInt64Test
[       OK ] DataIntegrationTest.TableCellValueInt64Test (1563 ms)
[ RUN      ] DataIntegrationTest.TableSampleRowKeysTest
[       OK ] DataIntegrationTest.TableSampleRowKeysTest (10003 ms)
[ RUN      ] DataIntegrationTest.TableReadMultipleCellsBigValue
[       OK ] DataIntegrationTest.TableReadMultipleCellsBigValue (7862 ms)
[----------] 17 tests from DataIntegrationTest (42797 ms total)
```

coryan:
I thought the problem could be where the build gets scheduled, one build may get "close" to the location for our test instance, while other builds may get scheduled very far.

I tested from home (I live on the US East Coast). It takes ~1min wherever the instance is (US/East, US/West, and EU/West).  That (I believe) eliminates the build cluster selection as a possible source of trouble (unless we have build clusters on Mars or something :grin:)

coryan:
I upgraded the instances to "production" and does not seem to make a difference either.

coryan:
Another repeat, cannot see anything obvious in the log:

https://source.cloud.google.com/results/invocations/1266d6a1-177f-4dd1-b5f2-06aeadbb04de/targets

coryan:
Separately I ran the test 200 times (with some changes) and got no failures, the longest execution times were:

```
[==========] 1 test from 1 test suite ran. (93061 ms total)
[==========] 1 test from 1 test suite ran. (90195 ms total)
[==========] 1 test from 1 test suite ran. (89327 ms total)
[...]

PR #2291, opened 2019-03-22
Replaced (date +%s) with {RANDOM}

### LINKED ISSUE #1460: Test driver scripts may miss errors.
The driver scripts for tests and examples use expressions like this:

```sh
local foo="blah-$(date +%s)"
```

These are a problem because errors in the shell expansion `$(...)` are not reported in this case.  We can use:

```sh
local foo
foo="blah-$(date +%s)"
```

but those do not work with `local -r`.  We can replace the "date as a source of random numbers" with `${RANDOM}` in some cases.

Regardless, we should search for such expressions and fix them appropriately.

Thanks to @dopiera for pointing this out.

rsimion:
Fixes #1460

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2291)

### DIFF
--- a/google/cloud/bigtable/examples/run_examples_emulator.sh
+++ b/google/cloud/bigtable/examples/run_examples_emulator.sh
@@ -26,8 +26,8 @@ echo "Running Bigtable Example programs"
 start_emulators
 
 # Use a (likely unique) project id for the emulator.
-readonly PROJECT_ID="project-$(date +%s)"
-readonly INSTANCE_ID="in-$(date +%s)-${RANDOM}"
+readonly PROJECT_ID="project-${RANDOM}-${RANDOM}"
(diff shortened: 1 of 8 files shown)


rsimion:
That error(Kokoro Ubuntu) was not triggered by my changes.

Just checked on on my local

[ RUN      ] DataIntegrationTest.TableReadMultipleCellsBigValue
[       OK ] DataIntegrationTest.TableReadMultipleCellsBigValue (41753 ms)
[----------] 17 tests from DataIntegrationTest (94933 ms total)

remyabel:
It looks like a regression of this: https://github.com/googleapis/google-cloud-cpp/issues/2015. Another example where it's occurred: https://source.cloud.google.com/results/invocations/a992d644-75d3-468c-bdbb-fee6eb59e9b1/log

Issue #2097, opened 2019-02-27
Remove `noex::Table`


coryan:
This is a placeholder, we will need to break it down into smaller tasks, now that `bigtable::Table` uses `StatusOr<>` we do not need `noex::Table`. We should remove the class.

If we do it in a single PR, then it is "easy", but it would be such a large PR that nobody would volunteer to review it.

I think we need to:

- Expose the `client()` from `noex::Table`.
- Move the implementation of each member function from `noex::Table` to `bigtable::Table`.
- Then remove `noex::Table`.

rsimion:
@coryan

What do you think about adding these accessor functions( or something similar)  to /bigtable/internal/table.h in order to make a smooth transition when replacing functions
```
  // Accessor functions - temporary until we remove this class
  std::shared_ptr<DataClient> get_client() const { return client_; }
  bigtable::AppProfileId get_app_profile_id() const { return app_profile_id_; }
  bigtable::TableId get_table_name() const { return table_name_; }
  std::shared_ptr<RPCRetryPolicy> get_rpc_retry_policy() const {
    return rpc_retry_policy_;
  }
  std::shared_ptr<RPCBackoffPolicy> get_rpc_backoff_policy() const {
    return rpc_backoff_policy_;
  }
  MetadataUpdatePolicy get_metadata_update_policy() const {
    return metadata_update_policy_;
  }
  std::shared_ptr<IdempotentMutationPolicy> get_idempotent_mutation_policy()
      const {
    return idempotent_mutation_policy_;
  }
```

 and for example in /bigtable/table.cc
 
 ```
 StatusOr<bool> Table::CheckAndMutateRow(std::string row_key, Filter filter,
                                        std::vector<Mutation> true_mutations,
                                        std::vector<Mutation> false_mutations) {
  grpc::Status status;
  bool value = impl_.CheckAndMutateRow(std::move(row_key), std::move(filter),
                                       std::move(true_mutations),
                                       std::move(false_mutations), status);
  if (!status.ok()) {
    return bigtable::internal::MakeStatusFromRpcError(status);
  }
  return value;
}
```
becomes something like
```
StatusOr<bool> Table::CheckAndMutateRow(std::string row_key, Filter filter,
                                        std::vector<Mutation> true_mutations,
                                        std::vector<Mutation> false_mutations) {
  grpc::Status status;
  // bool value = impl_.CheckAndMutateRow(std::move(row_key), std::move(filter),
[...]

PR #2292, opened 2019-03-22
Do not depend on noex::Table::AsyncApply().

rsimion:
Part of the work for #2097 
Fourth request in a series of pull requests to move functions from noex:;Table to Table.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2292)

### DIFF
--- a/google/cloud/bigtable/table.cc
+++ b/google/cloud/bigtable/table.cc
@@ -14,13 +14,12 @@
 
 #include "google/cloud/bigtable/table.h"
 #include "google/cloud/bigtable/internal/async_future_from_callback.h"
+#include "google/cloud/bigtable/internal/async_retry_unary_rpc.h"
 #include "google/cloud/bigtable/internal/bulk_mutator.h"
 #include "google/cloud/bigtable/internal/grpc_error_delegate.h"
 #include "google/cloud/bigtable/internal/unary_client_utils.h"
 #include <thread>
 #include <type_traits>
-
-#include "google/cloud/bigtable/internal/async_retry_unary_rpc.h"
 
 namespace btproto = ::google::bigtable::v2;
 namespace google {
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/table.cc
```diff
@@ -74,24 +76,38 @@ Status Table::Apply(SingleRowMutation mut) {
...
+             internal::ConstantIdempotencyPolicy(is_idempotent),
+             clone_metadata_update_policy(),
+             [client](grpc::ClientContext* context,
+                      google::bigtable::v2::MutateRowRequest const& request,
+                      grpc::CompletionQueue* cq) {
+               return client->AsyncMutateRow(context, request, cq);
+             },
+             std::move(request), cq)
+      .then([](future<StatusOr<google::bigtable::v2::MutateRowResponse>> r) {
+        return r.get().status();
```
remyabel: Should you check `if (r.get())` before getting `status()`?

coryan: I do not think we need to do that:

- If `r.get().ok()` is false, then we want to return `.status()`.
- If `r.get().ok()` is true, then `.status() == Status()` and that is exactly what we want to return.
- Our code is not supposed to throw, so the future should not contain an exception, but if it did, `r.get()` throws and the exception is captured in the future returned by `.then()`.

google/cloud/bigtable/table.cc
```diff
@@ -20,6 +20,8 @@
 #include <thread>
 #include <type_traits>
 
+#include "google/cloud/bigtable/internal/async_retry_unary_rpc.h"
```
coryan: Please move this `#include` to the block of includes, note that the block will be sorted by `clang-format`.

google/cloud/bigtable/table.cc
```diff
@@ -74,24 +76,38 @@ Status Table::Apply(SingleRowMutation mut) {
...
+  internal::SetCommonTableOperationRequest<
+      google::bigtable::v2::MutateRowRequest>(
+      request, impl_.app_profile_id_.get(), impl_.table_name_.get());
+  mut.MoveTo(request);
+  auto context = google::cloud::internal::make_unique<grpc::ClientContext>();
+
+  // Determine if all the mutations are idempotent. The idempotency of the
+  // mutations won't change as the retry loop executes, so we can just compute
+  // it once and use a constant value for the loop.
+
```
coryan: nit: remove this blank line please.

PR #2311, opened 2019-03-23
Workaround build problems on openSUSE:Leap.

coryan:
This weekend openSUSE:Leap released the 1.1.0i version of the OpenSSL
runtime libraries, but they did *not* release the development libraries
at the same time. Effectively this means that installing the development
libraries requires a downgrade.

I just changed both openSUSE:Leap and openSUSE:Tumbleweed to always
allow a downgrade when installing the development tools.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2311)

### DIFF
--- a/ci/test-readme/Dockerfile.opensuse
+++ b/ci/test-readme/Dockerfile.opensuse
@@ -23,7 +23,8 @@ FROM opensuse/tumbleweed:${DISTRO_VERSION}
 
 # ```bash
 RUN zypper refresh && \
-    zypper install -y cmake gcc gcc-c++ git libcurl-devel libopenssl-devel make
+    zypper install  --allow-downgrade -y cmake gcc gcc-c++ git gzip \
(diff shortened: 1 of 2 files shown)


ci/test-readme/Dockerfile.opensuse
```diff
@@ -23,7 +23,8 @@ FROM opensuse/tumbleweed:${DISTRO_VERSION}
 
 # ```bash
 RUN zypper refresh && \
-    zypper install -y cmake gcc gcc-c++ git libcurl-devel libopenssl-devel make
+    zypper install  --allow-downgrade -y cmake gcc gcc-c++ git gzip \
```
devbww: Nit: It looks like there are two spaces there.

coryan: Fixed.

Issue #2256, opened 2019-03-19
Create breakdown for additional Storage examples.


coryan:
For each region tag (and description) below we need to either:

- Check if the example exists, if it does, create a bug (or simply a PR) to fix the region tag. I think there are several in that category, I am blinding copying the region tag descriptions here.
- If the example does not exist, create a bug to implement that example. If multiple examples can be grouped in a single bug that is Okay.

We can close this bug once all the child bugs are created.

- [x] storage_create_bucket_class_location: Create a GCS bucket with storage class and location parameters
- [x] storage_add_bucket_owner: Add owner access control to a GCS bucket
- [x] storage_remove_bucket_owner: Remove owner access control to a GCS bucket
- [x] storage_add_bucket_default_owner: Add default access control to a GCS bucket
- [x] storage_add_file_owner: Add owner access control to an object in a GCS bucket
- [x] storage_remove_file_owner: Remove owner access to an object in a GCS bucket
- [x] storage_print_file_acl_for_user: View the access control list for an object in a GCS bucket, filtered by user
- [x] storage_change_file_storage_class: Rewrite and update the object storage class of a GCS object
- [x] storage_view_lifecycle_management_configuration: View lifecycle configuration of a GCS bucket
- [x] storage_define_bucket_website_configuration: Set static website configuration for a GCS bucket
- [x] storage_enable_versioning: Enable object versioning for a GCS bucket
- [x] storage_disable_versioning: Disable object versioning for a GCS bucket
- [x] storage_view_versioning_status: Check if object versioning status is enabled for a GCS bucket
- [x] storage_copy_file_archived_generation: Copy an archived generation of an object in a GCS bucket
- [x] storage_delete_file_archived_generation: Delete an archived generation of an object in a GCS bucket
- [x] storage_download_public_file: Download an object from a public data set object
- [x] storage_cors_configuration: Add CORS configuration to a GCS storage bucket

coryan:
I just submitted a PR to fix several region tags. I clicked their checkboxes already.

remyabel:
What should the region tags be for disabling or viewing the bucket website configuration?

coryan:
> What should the region tags be for disabling or viewing the bucket website configuration?

@remyabel I looked again, and there is not one defined. If we add such an example we can use `storage_print_website_configuration` (there are 5 `storage_print_*` region tags and 3 `storage_view_*` region tags, so I am going with `print`).

[...]

PR #2318, opened 2019-03-24
Implement remove/add object owner samples.

remyabel:
Ref: #2256

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2318)

### DIFF
--- a/google/cloud/storage/examples/storage_object_acl_samples.cc
+++ b/google/cloud/storage/examples/storage_object_acl_samples.cc
@@ -260,6 +260,83 @@ void PatchObjectAclNoRead(gcs::Client client, int& argc, char* argv[]) {
   (std::move(client), bucket_name, object_name, entity, role);
 }
 
+void AddObjectOwner(gcs::Client client, int& argc, char* argv[]) {
+  if (argc != 4) {
+    throw Usage{"add-object-owner <bucket-name> <object-name> <entity>"};
+  }
+  auto bucket_name = ConsumeArg(argc, argv);
+  auto object_name = ConsumeArg(argc, argv);
+  auto entity = ConsumeArg(argc, argv);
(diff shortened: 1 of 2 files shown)


google/cloud/storage/examples/storage_object_acl_samples.cc
```diff
@@ -260,6 +260,83 @@ void PatchObjectAclNoRead(gcs::Client client, int& argc, char* argv[]) {
...
+
+    std::cout << "ACL entry for " << patched_acl->entity() << " in object "
+              << patched_acl->object() << " in bucket " << patched_acl->bucket()
+              << " is now " << *patched_acl << "\n";
+  }
+  //! [add file owner] [END storage_add_file_owner]
+  (std::move(client), bucket_name, object_name, entity);
+}
+
+void RemoveObjectOwner(gcs::Client client, int& argc, char* argv[]) {
```
remyabel: When I ran this sample manually, it had a noticeable delay. Performance improvements welcome.

coryan:
___
*[google/cloud/storage/examples/storage_object_acl_samples.cc, line 291 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2318#-Laik5KylVAyH2EfUSdR-r1-291:-LapTON2C4EUCNAs9ZXg:b-7v97wl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/350ca63b008431f2a3f8293f4464b95f13b022ff/google/cloud/storage/examples/storage_object_acl_samples.cc#L291)):*
> The code looks about optimal for this purpose.

I should have added some detail to that.  Using `DeleteObjectAcl()` should be faster, because it saves a roundtrip, but we want to explicitly demonstrate how to remove the ACL via the metadata operations (I think). So we need two roundtrips, and that can get slow.

PR #2323, opened 2019-03-25
Change CMake-config files to work without `pkg-config` dependency.

### LINKED ISSUE #2234: Install only installs google_cloud_cpp_common
I followed the install.md to install the lates version on Ubuntu 18.04 and it looks like google_cloud_cpp_common got installed but storage_client not. Am I missing something? When I tried to build a small test project, CMake can't fing storage_client but it has no trouble finding google_cloud_cpp_common.

### LINKED ISSUE #2275: Fallback on find_library() to find gRPC libraries.
I am about to submit a PR that introduces the `FindgRPC` module. Currently the module supports finding gRPC using either CMake-config files or using `pkg-config`. That is enough for us, but would be nice to support `find_library()` and `find_path()` too in case none of those work.

coryan:
In this PR we fix `FindgRPC.cmake` to do not depend on `pkg-config`, and change the
generated config files for `googleapis` and `bigtable_client` to take advantage of this
new file.  This fixes #2234 and fixes #2275.

To test these changes I modified the `ci/test-readme/Dockerfile.*` files to also build
sample programs using CMake after installing `google-cloud-cpp`. Separately we enabled
these builds as presubmit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2323)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,17 +159,3 @@ add_custom_target(doxygen-docs)
 
 # Add subprojects here.
 add_subdirectory(google/cloud)
-
-include(GNUInstallDirs)
-install(FILES
-        "cmake/FindProtobufTargets.cmake"
-        DESTINATION
-        "${CMAKE_INSTALL_LIBDIR}/cmake/ProtobufTargets")
-install(FILES
-        "cmake/FindgRPC.cmake"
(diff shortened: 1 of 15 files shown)


coryan:
PTAL.

PR #2327, opened 2019-03-25
Remove pkg config references from the documentation.

### LINKED ISSUE #1610: Define support level for pkg-config based builds.
The current CMake configuration supports using [`pkg-config`](https://www.freedesktop.org/wiki/Software/pkg-config/) to discover the flags for our dependencies.  Basically the CMake file calls `pkg-config(1)` to discover what compiler and linker flags are needed to find `libcurl`, or `OpenSSL`, or `protobuf`.

In some Linux distribution it seems like this is the preferred way to configure dependencies (e.g. Gentoo [packages google-cloud-cpp](https://github.com/gentoo/gentoo/tree/master/net-libs/google-cloud-cpp) using `pkg-config(1)`.

I think we have basically three options:

1. Drop the support for `pkg-config(1)` completely, we remove the CMake code that uses `pkg-config(1)` to discover the dependencies.

1. Support `pkg-config(1)` as "contributed", meaning we accept patches to improve it, but do not create a build that runs it or fix the configuration ourselves (this is basically the current support).

1. Support `pkg-config(1)` for "realsies", meaning we create at least one build with it, and keep it working.

coryan:
Fix the documentation, where we still had a couple of places saying that
`pkg-config` was an optional mechanism to configure dependencies. We
have eliminated the need for this, and the code is cleaner. Also fix the
description of how we support dependencies via CMake.

This fixes #1610.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2327)

### DIFF
--- a/doc/managing-dependencies-in-cmake.md
+++ b/doc/managing-dependencies-in-cmake.md
@@ -1,4 +1,4 @@
-# Dependency Management for google-cloud-cpp
+# CMake Dependency Management for google-cloud-cpp
 
 `google-cloud-cpp` must have a minimal set of dependencies, every dependency is
 viewed as a blocker for adoption by our customers.  Only the following
@@ -17,6 +17,8 @@ libraries are allowed as dependencies:
   library to parse and generate JSON objects.
 * [OpenSSL](https://www.openssl.org/source/) is used by the storage library to
   perform some Base64 encoding and JWT signing.
+* [crc32c](https://github.com/google/crc32c) is used by the storage library to
+  generate CRC32C checksums.
 
 Some of these libraries have dependencies themselves:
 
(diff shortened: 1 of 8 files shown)


doc/managing-dependencies-in-cmake.md
```diff
@@ -34,19 +36,19 @@ included in CMake files.
...
 
 Bazel builds are basically external project builds.
 
-When compiline with external projects we will create
-[targets][cmake-doc-targets] that have the same names (and roles) as the
-installed for that dependency.
-For example, when protobuf is installed and discovered via `find_package()` the
-module for protobuf introduces a `protobuf::libprotobuf` target. We will
-construct our external projects to exhibit the same behavior.
+When compiline with external projects we create [targets][cmake-doc-targets]
```
devbww: s/compiline/compiling/

PR #2328, opened 2019-03-25
Update HMAC functions to catchup with protocol.

### LINKED ISSUE #2293: Update resource name for HmacKeyMetadata
The resource name and some of the fields changed from `HmacKey` to `HmacKeyMetadata`, we need to update the code.

coryan:
The JSON protocol for HMAC keys is (as advertised) not stable. Changed
the implementation to catchup with the latest updates in the spec.

This fixes #2293.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2328)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -2342,7 +2342,7 @@ class Client {
     if (!result) {
       return result.status();
     }
-    return std::make_pair(std::move(result->resource),
+    return std::make_pair(std::move(result->metadata),
                           std::move(result->secret));
   }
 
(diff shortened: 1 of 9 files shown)


coryan:
@devbww 

> It is a bit of a shame that it is `timeCreated` but just `updated` (instead of `timeUpdated`).

Indeed.  But that is what the spec says, and follows the "pattern" for other JSON resources.

In general, we prefer not to rename or otherwise make the fields "better".  If the JSON (or gRPC) protocol calls them "fooBar" then we just translate that to "foo_bar" even if "foo_baz" would have been a an improvement.

devbww:
Absolutely.  I was just lamenting the upstream choice.

PR #2334, opened 2019-03-26
Implement ObjectMetadata::set_storage_class().

### LINKED ISSUE #2322: Implement ObjectMetadata::set_storage_class
Correct me if I'm wrong, I don't think there's currently any implementation for this. We can either model it after this:

https://github.com/googleapis/google-cloud-cpp/blob/7b26e02f7ce8b47580cd6c242edff8e503272e87/google/cloud/storage/bucket_metadata.h#L780

Or create a `WellKnownParameter`. The reason being is that the [docs](https://cloud.google.com/storage/docs/changing-storage-classes) state changing the object's storage class is a rewrite operation.

coryan:
The field is writeable, though not via Update or Patch operations.

This fixes #2322.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2334)

### DIFF
--- a/google/cloud/storage/object_metadata_test.cc
+++ b/google/cloud/storage/object_metadata_test.cc
@@ -442,6 +442,15 @@ TEST(ObjectMetadataTest, SetEventBasedHold) {
   EXPECT_NE(expected, copy);
 }
 
+/// @test Verify we can change the storageClass field.
+TEST(ObjectMetadataTest, SetStorageClass) {
+  auto expected = CreateObjectMetadataForTest();
+  auto copy = expected;
+  copy.set_content_type("NEARLINE");
+  EXPECT_EQ("NEARLINE", copy.content_type());
+  EXPECT_NE(expected, copy);
+}
+
 /// @test Verify we can delete metadata fields.
 TEST(ObjectMetadataTest, DeleteMetadata) {
   auto expected = CreateObjectMetadataForTest();
--- a/google/cloud/storage/object_metadata_test.cc
+++ b/google/cloud/storage/object_metadata_test.cc
@@ -446,8 +446,8 @@ TEST(ObjectMetadataTest, SetEventBasedHold) {
 TEST(ObjectMetadataTest, SetStorageClass) {
   auto expected = CreateObjectMetadataForTest();
   auto copy = expected;
-  copy.set_content_type("NEARLINE");
-  EXPECT_EQ("NEARLINE", copy.content_type());
+  copy.set_storage_class("NEARLINE");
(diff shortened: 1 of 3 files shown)


google/cloud/storage/testbench/gcs_object.py
```diff
@@ -406,10 +406,11 @@ def _remove_non_writable_keys(cls, metadata):
         writeable_keys = {
             'acl', 'cacheControl', 'contentDisposition', 'contentEncoding',
             'contentLanguage', 'contentType', 'eventBasedHold', 'metadata',
-            'temporaryHold'
+            'temporaryHold', 'storageClass'
         }
         for key in metadata.keys():
             if key not in writeable_keys:
+                print("\n\nDEBUG DO NOT MERGE removing %s\n\n" % key)
```
devbww: Remove?

google/cloud/storage/object_metadata_test.cc
```diff
@@ -442,6 +442,15 @@ TEST(ObjectMetadataTest, SetEventBasedHold) {
   EXPECT_NE(expected, copy);
 }
 
+/// @test Verify we can change the storageClass field.
+TEST(ObjectMetadataTest, SetStorageClass) {
+  auto expected = CreateObjectMetadataForTest();
+  auto copy = expected;
+  copy.set_content_type("NEARLINE");
```
remyabel: Shouldn't this be `set_storage_class()` and `copy.storage_class()`?

remyabel:
Is this PR complete? It is my understanding that we should add a `WellKnownParameter` so that the following is possible:

```
client.RewriteObjectBlocking(source_bucket_name, source_object_name,
                                     destination_bucket_name,
                                     destination_object_name,
                                     gcs::ObjectStorageClass{"NEARLINE"});
```

coryan:
@remyabel I do not think we need a `WellKnownParameter`, we should be able to say:

```C++
client.RewriteObjectBlocking(
    source_bucket_name, source_object_name, destination_bucket_name,
    destination_object_name, gcs::WithObjectMetadata(
        gcs::ObjectMetadata().set_storage_class("NEARLINE")));
```

There are a few attributes that can be set either via well-known parameters or via the metadata resource, we give users a choice because the underlying API does.

google/cloud/storage/object_metadata_test.cc
coryan: Indeed. Fixed.

google/cloud/storage/testbench/gcs_object.py
coryan: Done.

coryan:
PTAL.

PR #2337, opened 2019-03-26
Disable some MSVC warnings because Protobuf.

coryan:
Protobuf headers generate too many warnings on MSVC. Specifically, they
use `std::iterator<>` which will be deprecated in C++17, but MSVC warns
about it now, and they convert from `int64` to `int` without a cast,
which MSVC also dislikes.

Oh, and we are getting these warnings now because in the changes to improve
the installation process I removed a bunch of code to turn off warnings. Good thing
too because only 2 out of like 10 was relevant.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2337)

### DIFF
--- a/google/cloud/bigtable/CMakeLists.txt
+++ b/google/cloud/bigtable/CMakeLists.txt
@@ -55,6 +55,16 @@ endif ()
 # users to add more common options without having to touch the code.
 add_library(bigtable_common_options INTERFACE)
 google_cloud_cpp_add_common_options(bigtable_common_options)
+if (MSVC)
+    #
+    # We turn off 4996 because some protobuf types derive from std::iterator,
+    # turn off the warnings about this being deprecated in C++17.
+    #
+    # We turn off 4244 because protobuf headers convert from `int64` to
+    # `RepeatedPtrField< T>::size_type`, which is really `int` and that may
+    # cause loss of data. But we only run on 64-bit platforms.
+    target_compile_options(bigtable_common_options INTERFACE /wd4996 /wd4244)
+endif ()
 
 # Configure the location of proto files, particulary the googleapis protos.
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/CMakeLists.txt
```diff
@@ -55,6 +55,16 @@ endif ()
 # users to add more common options without having to touch the code.
 add_library(bigtable_common_options INTERFACE)
 google_cloud_cpp_add_common_options(bigtable_common_options)
+if (MSVC)
+    #
```
devbww: Delete this empty line perhaps?

google/cloud/bigtable/CMakeLists.txt
```diff
@@ -55,6 +55,16 @@ endif ()
...
+    # We turn off 4996 because some protobuf types derive from std::iterator,
+    # turn off the warnings about this being deprecated in C++17.
+    #
+    # We turn off 4244 because protobuf headers convert from `int64` to
+    # `RepeatedPtrField< T>::size_type`, which is really `int` and that may
+    # cause loss of data. But we only run on 64-bit platforms.
```
devjgm: The last sentence is a little confusing to me because even on a 64-bit platform, assigning int64->int could cause data loss, unless it's an ILP64 data model, which I don't think msvc uses.

coryan: Good point.  I have changed the comment.  We could leave the warning on, but then almost every file would get it.  Protobuf code uses `std::distance()` and sticks the value in what amounts to an `int`:

https://github.com/protocolbuffers/protobuf/blob/3.7.x/src/google/protobuf/repeated_field.h#L2106-L2113

I am guessing that is Okay because the protobuf repeated fields cannot hold anything bigger than `MAX_INT` elements `¯\_(ツ)_/¯`

coryan:
PTAL.

PR #2341, opened 2019-03-26
Fix std::size_t to int warnings.

coryan:
These are easy because the values are small enough to fit into the
destination type, so a `static_cast<>` does the trick.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2341)

### DIFF
--- a/google/cloud/bigtable/testing/table_integration_test.cc
+++ b/google/cloud/bigtable/testing/table_integration_test.cc
@@ -77,18 +77,20 @@ std::string TableTestEnvironment::RandomTableId() {
   // This value was discovered by trial and error, it is not documented in the
   // proto files.
   constexpr int kMaxTableIdLength = 50;
-  static std::string const prefix = "table-";
-  return CreateRandomId(prefix,
-                        static_cast<int>(kMaxTableIdLength - prefix.length()));
+  static char const prefix[] = "table-";
+  static_assert(kMaxTableIdLength > sizeof(prefix), "prefix is too long");
+  constexpr int sample_count = kMaxTableIdLength - sizeof(prefix) + 1;
+  return CreateRandomId(prefix, sample_count);
 }
 
 std::string TableTestEnvironment::RandomInstanceId() {
   // This value was discovered by trial and error, it is not documented in the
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/testing/table_integration_test.cc
```diff
@@ -78,15 +78,17 @@ std::string TableTestEnvironment::RandomTableId() {
   // proto files.
   constexpr int kMaxTableIdLength = 50;
   static std::string const prefix = "table-";
-  return CreateRandomId(prefix, kMaxTableIdLength - prefix.length());
+  return CreateRandomId(prefix,
```
devbww: Suggestion: Perhaps consider an alternative where `CreateRandomId()` takes a `std::size_t max_length`?

Then it can be the one place that has to deal with `prefix` being too long, and we can probably avoid the reverse `std::size_t(n)` cast in `Sample()`.

google/cloud/bigtable/testing/table_integration_test.cc
```diff
@@ -77,18 +77,20 @@ std::string TableTestEnvironment::RandomTableId() {
   // This value was discovered by trial and error, it is not documented in the
   // proto files.
   constexpr int kMaxTableIdLength = 50;
-  static std::string const prefix = "table-";
-  return CreateRandomId(prefix,
-                        static_cast<int>(kMaxTableIdLength - prefix.length()));
+  static char const prefix[] = "table-";
+  static_assert(kMaxTableIdLength > sizeof(prefix), "prefix is too long");
+  constexpr int sample_count = kMaxTableIdLength - sizeof(prefix) + 1;
```
devbww: Do you not get the same `size_t` to `int` warning on these, or do we avoid it now that it is at `constexpr` time?

coryan: I was checking that (on two platforms, thus the delay) before saying PTAL, but yes, it works.

devbww: Cool.  Thanks.

PR #2343, opened 2019-03-26
Update README shields.

coryan:
Update the shields in the README file.  You can see them here:

https://github.com/coryan/google-cloud-cpp/blob/update-README-shields/README.md

We were missing a few and it was getting unwieldy, so I grouped them.  I am not sure I like the grouping or the ragged formatting, if you have suggestions please speak up.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2343)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,13 @@
 
 C++ Idiomatic Clients for [Google Cloud Platform][cloud-platform] services.
 
-<!-- Core Results -->
+**Core Results:**
 [![Kokoro CI status][kokoro-ubuntu-shield]][kokoro-ubuntu-link]
 [![Kokoro CI status][kokoro-macos-shield]][kokoro-macos-link]
 [![Kokoro CI status][kokoro-windows-shield]][kokoro-windows-link]
 [![Codecov Coverage status][codecov-shield]][codecov-link]
-<!-- Test Install Instructions -->
+
+**Test Install Instructions:**
 [![Kokoro install centos status][kokoro-install-centos-shield]][kokoro-install-centos-link]
 [![Kokoro install debian status][kokoro-install-debian-shield]][kokoro-install-debian-link]
 [![Kokoro install fedora status][kokoro-install-fedora-shield]][kokoro-install-fedora-link]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,15 +2,31 @@
 
 C++ Idiomatic Clients for [Google Cloud Platform][cloud-platform] services.
 
-[![Travis CI status][travis-shield]][travis-link]
+<!-- Core Results -->
```
devbww: Would these be nice as subheadings instead of comments?

README.md
```diff
@@ -31,6 +47,17 @@ C++ Idiomatic Clients for [Google Cloud Platform][cloud-platform] services.
 [kokoro-noex-link]: https://storage.googleapis.com/cloud-cpp-kokoro-status/kokoro-noex-link.html
 [kokoro-centos-7-shield]: https://storage.googleapis.com/cloud-cpp-kokoro-status/kokoro-centos-7.svg
 [kokoro-centos-7-link]: https://storage.googleapis.com/cloud-cpp-kokoro-status/kokoro-centos-7-link.html
+[kokoro-check-abi-shield]: https://storage.googleapis.com/cloud-cpp-kokoro-status/kokoro-check-abi.svg
```
devbww: I don't see these producing any output when I click "View file".  Am I missing something?

README.md
coryan: Using actual headers (the markdown equivalent of h2, h3, etc) looked bad (to me), I used bolded words:

https://github.com/coryan/google-cloud-cpp/blob/update-README-shields/README.md

README.md
coryan: This is what I see:

image

Do you see something different?

README.md
devbww: Optional: Perhaps stick them in a little table like on the internal team page?

README.md
devbww: I see that, but I don't see anything between the "Client Library Documentation" link and the "This library supports the following ..." text, where these sit.  Are they supposed to render something?

coryan: Ah, those are link definitions (or reference-style links):

https://www.markdownguide.org/basic-syntax/#reference-style-links

basically something like `[foo]: https://example.com` defines `[foo]` as an alias for `https://example.com`. It leaves the inline links a little more readable.

devbww: Gotcha.  Thanks.

README.md
coryan: I did try tables:

https://github.com/coryan/google-cloud-cpp/blob/5257e43b1e9afc18305350552b5e18b09ffee183/README.md

Did not like how much horizontal space they took.  But if we can figure out a way to make them more compact I would be all for them.  Keep in mind that the README file is the first file folks go to for information about "what is this?", "how do I compile it?", etc.

devbww: Ack.  Let's leave any possibility of compact tables for another day.  Thanks.

Issue #1811, opened 2019-01-08
Change remaining std::pair<Status, T> to StatusOr<T>.


coryan:
Evidently I missed at least one internal function (`SignString`) that returns `std::pair<Status, T>` and should return `StatusOr<T>`.

coryan:
Fixed in #1810.

remyabel:
Is this actually fixed? It looks like `SignString` still returns `std::pair<Status, std::string>`:

https://github.com/googleapis/google-cloud-cpp/blob/0160ebc1c1caa1b10b9e2f679eaae2fb78cd0e43/google/cloud/storage/oauth2/service_account_credentials.h#L124

coryan:
Evidently not.

PR #2370, opened 2019-03-30
Support policy documents

### LINKED ISSUE #1932: Support policy documents with V2 Signed URLs.
Figure out how to create signed URLs that support this:

https://cloud.google.com/storage/docs/xml-api/post-object#usage_and_examples

remyabel:
Fixes #1932

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2370)


google/cloud/storage/client.cc
```diff
@@ -268,6 +268,36 @@ credentials, as described in:
...
+Please configure your google::cloud::storage::Client to use service account
+credentials, as described in:
+https://cloud.google.com/storage/docs/authentication
+)""");
+  }
+
+  auto base64_policy = internal::Base64Encode(request.StringToSign());
+
+  auto result = credentials->SignString(base64_policy);
+  if (!result.first.ok()) {
```
remyabel: I copied this line from this example:

https://github.com/googleapis/google-cloud-cpp/blob/63fcd93928a6aaab1bf045bccb610938ba246b21/google/cloud/storage/client.cc#L217

But I am unable to see in what scenario this would never equal `ok()`. To add to that, the line is red in the code coverage report.

remyabel: I commented in https://github.com/googleapis/google-cloud-cpp/issues/1811 that `SignString` doesn't return `StatusOr`. But I attempted to fix it myself and realized that the call chain is actually pretty large, so I gave up.

coryan:
@remyabel Argh, I did not notice the problem with `join()` vs. `Join()`, good thing we have tests, can you fix it?

Issue #2374, opened 2019-03-30
Client::WriteObject() still raises errors in some cases.


coryan:
In most cases `Client::WriteObject()` signals errors by setting the iostream status bits and the `.status()` accessor in the returned `ObjectWriteStream`.  But if the error condition is detected before we even create a valid `ObjectWriteStreambuf` then the function raises an exception.

That is not the expected behavior.

PR #2375, opened 2019-03-30
Do not raise exceptions in ReadObject().

### LINKED ISSUE #2373: Client::ReadObject() still raises errors in some cases.
In most cases `Client::ReadObject()` signals errors by setting the iostream status bits and the `.status()` accessor in the returned `ReadObjectStream`.  But if the error condition is detected before we even create a valid `ReadObjectStreambuf` then the function raises an exception.

That is not the expected behavior.

coryan:
The plan was to return the status in the ReadObjectStream, that was not
working when there were errors (such as authentication problems)
preventing us from even creating a valid ReadObjectStreambuf. The
"solution" proposed here is to create a streambuf that is permanently
set to "this is an error stream, everything is terrible".

I am not sure this is the best fix, and welcome other ideas. It does not
solve all the known problems with error handling in `ReadObject()`, but
I think it is a pre-requisite to solve them. There is a separate bug to track
similar problems on `WriteObject` (#2374), and I think it should have a
similar solution.

This fixes #2373.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2375)

### DIFF
--- a/google/cloud/storage/internal/object_streambuf.h
+++ b/google/cloud/storage/internal/object_streambuf.h
@@ -51,6 +51,33 @@ class ObjectReadStreambuf : public std::basic_streambuf<char> {
 };
 
 /**
+ * A read stream in a permanent error status.
+ */
+class ObjectReadErrorStreambuf : public ObjectReadStreambuf {
+ public:
+  explicit ObjectReadErrorStreambuf(Status status)
+      : status_(std::move(status)) {}
+
+  void Close() override {}
+  bool IsOpen() const override { return false; }
+  Status const& status() const override { return status_; }
+  std::string const& received_hash() const override { return received_hash_; }
+  std::string const& computed_hash() const override { return computed_hash_; }
+  std::multimap<std::string, std::string> const& headers() const override {
+    return headers_;
+  }
+
+ protected:
+  int_type underflow() override { return traits_type::eof(); }
+
+ private:
+  Status status_;
+  std::string received_hash_;
+  std::string computed_hash_;
+  std::multimap<std::string, std::string> headers_;
+};
+
+/**
  * Defines a compilation barrier for libcurl.
  *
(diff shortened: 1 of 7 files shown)


google/cloud/storage/internal/object_streambuf.h
```diff
@@ -50,6 +50,33 @@ class ObjectReadStreambuf : public std::basic_streambuf<char> {
   virtual std::multimap<std::string, std::string> const& headers() const = 0;
 };
 
+/**
+ * A read stream in a permanent error status.
```
devbww: s/status/state/ perhaps?

google/cloud/storage/object_test.cc
```diff
@@ -391,6 +392,54 @@ TEST_F(ObjectTest, PatchObjectPermanentFailure) {
...
+
+  Status status =
+      client.ReadObject("test-bucket-name", "test-object-name").status();
+  EXPECT_EQ(TransientError().code(), status.code());
+  EXPECT_THAT(status.message(), HasSubstr("Retry policy exhausted"));
+  EXPECT_THAT(status.message(), HasSubstr("ReadObject"));
+}
+
+TEST_F(ObjectTest, ReadObjectPermanentFailure) {
+  // We cannot use google::cloud::storage::testing::TooManyFailuresStatusTest,
```
devbww: s/TooManyFailuresStatusTest/PermanentFailureStatusTest/?

google/cloud/storage/internal/object_streambuf.h
coryan: Fixed, much better.

google/cloud/storage/object_test.cc
coryan: Fixed. Thanks!

coryan:
PTAL. I added a unit test because I needed a fix of code coverage improvements.

PR #2377, opened 2019-03-31
BulkMutator is overusing UNKNOWN error.

### LINKED ISSUE #2048: BulkMutator is overusing UNKNOWN error.
Currently, if an attempt to execute `BulkApply` fails because of a broken or cancelled stream, `UNKNOWN` is reported for all mutations which we didn't get responses for. This also happens on hitting the retry deadline.

This is suboptimal - we should report the status of the stream instead.

rsimion:
Fixes #2048

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2377)


rsimion:
@coryan, @dopiera 
This is what I was talking about during our video meeting on Wednesday. Without that change, the status code in the table_bulk_apply_test.cc was coming out as UNKNOWN in both cases.

google/cloud/bigtable/table_bulk_apply_test.cc
```diff
@@ -223,6 +223,8 @@ TEST_F(TableBulkApplyTest, TooManyFailures) {
       bt::SingleRowMutation("foo", {bt::SetCell("fam", "col", 0_ms, "baz")}),
       bt::SingleRowMutation("bar", {bt::SetCell("fam", "col", 0_ms, "qux")})));
   EXPECT_FALSE(failures.empty());
+  EXPECT_EQ(google::cloud::v0::StatusCode::kAborted,
```
remyabel: Why `google::cloud::v0::StatusCode`? Can't this just be `google::cloud::StatusCode`?

rsimion:
@dopiera 
I agree.
 Also, AsyncBulkApply has not yet been re -written( like AsyncApply) , so it can be changed at that time.

Waiting for @coryan suggestions (on how to proceed).

google/cloud/bigtable/table_bulk_apply_test.cc
rsimion: sure

dopiera:
> @dopiera
> I agree.
> Also, AsyncBulkApply has not yet been re -written( like AsyncApply) , so it can be changed at that time.
> 
> Waiting for @coryan suggestions (on how to proceed).

Here are is my suggestion on how to proceed:
- have a `protected` member like `last_status`
- add a `status` parameter to `BulkMutator::FinishRequst()` and fill it accordingly
- inside `BulkMutator::FinishRequest()`, store the `status` in `last_status`
- in `BulkMutator::ExtractFinalFailures()` use `last_status`.

Though, Carlos might have a better idea.

rsimion:
@dopiera 
sounds good to me

google/cloud/bigtable/internal/bulk_mutator_test.cc
```diff
@@ -417,7 +417,7 @@ TEST(MultipleRowsMutatorTest, RetryOnlyIdempotent) {
 
   EXPECT_EQ(2, failures[1].original_index());
   // EXPECT_EQ("baz", failures[1].mutation().row_key());
-  EXPECT_EQ(google::cloud::StatusCode::kUnknown, failures[1].status().code());
+  EXPECT_EQ(google::cloud::StatusCode::kInternal, failures[1].status().code());
```
dopiera: This tests only the first of the changed errors.

Could you please also test the second one? You can easily achieve it by adding another `bt::SingleRowMutation("bar", {bt::SetCell("fam", "col", 0_ms, "qux")}` to the `BulkMutation` in this test and checking that there is one more failure, also with `kInternal`.

PR #2382, opened 2019-04-01
Implement bigtable::Table::AsyncCheckAndMutateRow

### LINKED ISSUE #1489: Implement bigtable::Table::AsyncCheckAndMutateRow
This function will implement an asynchronous version of `Table::ReadRow`. The function will have the following signature:

```C++
future<bool> AsyncCheckAndMutateRow(
    std::string row_key, Filter filter,
    std::vector<Mutation> true_mutations,
    std::vector<Mutation> false_mutations,
    CompletionQueue& cq);
};
```

- [ ] Implement the function.

rsimion:
Fixes #1489 

Part of #2097

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2382)

### DIFF
--- a/google/cloud/bigtable/examples/data_async_snippets.cc
+++ b/google/cloud/bigtable/examples/data_async_snippets.cc
@@ -22,6 +22,7 @@
 
 namespace {
 namespace cbt = google::cloud::bigtable;
+const std::string MAGIC_ROW_KEY = "key-000005";
 
 struct Usage {
   std::string msg;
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/examples/data_async_snippets.cc
```diff
@@ -22,6 +22,7 @@
 
 namespace {
 namespace cbt = google::cloud::bigtable;
+const std::string MAGIC_ROW_KEY = "key-000005";
```
remyabel: What's the agreed upon convention for constant naming here? Should this be `kMagicRowKey`?

rsimion: I used the same format as in data_snippets.cc
@coryan would know better.

google/cloud/bigtable/examples/data_async_snippets.cc
```diff
@@ -124,6 +125,47 @@ void AsyncBulkApply(cbt::Table table, cbt::CompletionQueue cq,
...
+          if (!row) {
+            throw std::runtime_error(row.status().message());
+          }
+
+          return google::cloud::Status();
+        });
+    final.get();
+  }
+  //! [async check and mutate]
+  (std::move(table), std::move(cq), argv[1]);
```
remyabel: What do you think about using the [`ConsumeArg`](https://github.com/googleapis/google-cloud-cpp/blob/e8e461a3416d339cde9948fb2f785ce7dc22e885/google/cloud/bigtable/examples/data_snippets.cc#L39) function present in the other sample files for consistency?

rsimion: I agree, but I do not want to start changing things around unless approved (@coryan).

google/cloud/bigtable/examples/data_async_snippets.cc
coryan: We have been terribly inconsistent with naming constants, we need to make a pass cleaning them up.  Unfortunately the clang-tidy checks break if you change some of these, so I have been leaving them alone until we figure out the correct incantations for clang-tidy.

google/cloud/bigtable/examples/data_async_snippets.cc
coryan: Another case where we should probably make a pass and remove all the `ConsumeArg` things.

google/cloud/bigtable/examples/data_async_snippets.cc
remyabel: I did attempt it a while ago. Something like the following should work in theory:

```
- { key: readability-identifier-naming.GlobalConstantPrefix,    value: k }
- { key: readability-identifier-naming.GlobalConstantCase,     value: CamelCase }
```

If I recall, after I applied this pattern to the existing check options, it started complaining about about the include guards.

PR #2384, opened 2019-04-01
Make `StartAsyncPollOp` accept future<Operation>.

dopiera:
Per @coryan 's suggestion, `StartAsyncPollOp` is always used in then, se we
might make it friendlier to use in such a case.

This change allows calling `StartAsyncPollOp` before the operation is
ready, which allows us to not capture arguments required for its
creating in yet another entity - that was cumbersome.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2384)


dopiera:
This is how it simplifies #2357:
https://github.com/dopiera/google-cloud-cpp/commit/cce4cbe9734e1a49eb2719180980ae0c15e628a5

PR #2387, opened 2019-04-02
Add instructions for generating compile_commands.json [skip ci]

### LINKED ISSUE #2021: Consider mentioning CMAKE_EXPORT_COMPILE_COMMANDS in the docs
[CMAKE_EXPORT_COMPILE_COMMANDS](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html) is a useful flag that when enabled outputs a JSON file that can be used with code completion engines like [YouCompleteMe](https://github.com/Valloric/YouCompleteMe), allowing them to have their code statically analyzed by libclang on the fly while inside a text editor. Perhaps this can go into `CONTRIBUTING.md`.

remyabel:
Fixes #2021

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2387)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ github-io-staging/
 
 google/cloud/storage/testbench/__pycache__/
 google/cloud/storage/testbench/*.pyc
+
+compile_commands.json
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -276,3 +276,29 @@ which combinations are tested regularly.
  * `CREATE_GRAPHVIZ`: if set to `yes`, use `CMake` to generate a dependency
    graph of each target. This is useful when troubleshooting dependencies, or
    simply when trying to document them.
+
+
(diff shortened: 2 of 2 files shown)


remyabel:
> Is this just to compile the external projects and get the compilation database for them too? There is a target (google-cloud-cpp-dependencies) that compiles just the dependencies. Do you think it would be better to document that as a target?

The compilation database will already be generated for the external projects. The problem is some files (like `nlohmann_json.hpp`) don't get installed unless you actually build the target, and it will show up as missing (see discussion in the linked issue). Building the entire target is a nuclear issue, not sure if there's a better way. But I'm game with only building the dependencies target.

remyabel:
I'm closing this for now. It's getting a little bikesheddy and the amount of people it will impact will probably be little to none.

PR #2389, opened 2019-04-02
Fixed small StatusOr::op bool, usage bug

devjgm:
`WaitForConsistency` returns a `future<StatusOr<bool>>`, so checking `.get()` is only checking that the future's `StatusOr` is OK, but it's not checking the boolean value of the `StatusOr`'s contained bool.

This may be an easy mistake to make with `StatusOr<bool>` in general. This is the only bug I saw in our codebase, but we might want to avoid `StatusOr<bool>` in general. I wonder if it's an antipattern. I'll file a separate issue where we can discuss that notion in general.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2389)


coryan:
@devjgm seems like this is all Okay, should we merge it?

PR #2397, opened 2019-04-03
Always use resumable uploads in ObjectWrite() for better error detection and retries.

coryan:
Currently we have 3 ways to upload object data:

- If you give the library a string, we upload using a single POST request.
- If you create a `ObjectWriteStream` this might be backed by a resumable upload, in which case a "chunk" is uploaded at a time (using PUT), and we apply the full retry loop for each chunk.
- If you create a `ObjectWriteStream` without resumable uploads the library just tries to upload your data, if there is an error the library cannot recover.

Instead of improving the error handling and detection for the second approach this PR proposes that we simply remove it.  There is some downside: for small objects the second approach is faster, but for small objects the first approach is even faster than the second.

This is a big change and we can discuss how to break it into smaller pieces if you want. Note that the public APIs remain the same, we just changed one of the implementations.

If you are wondering why we implemented all three, it is because we implemented the first and second approaches first because they are easier to implement.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2397)


devjgm:
> There is some downside: for small objects the second approach is faster, but for small objects the first approach is even faster than the second.

I'm not seeing how this is a downside of removing the third approach.

coryan:
> > There is some downside: for small objects the second approach is faster, but for small objects the first approach is even faster than the second.
> 
> I'm not seeing how this is a downside of removing the third approach.

Argh, because we are removing the *second* approach.  I just fixed the description.

PR #2399, opened 2019-04-04
Standardize documentation on "cmake-out" for cmake's binary output.

### LINKED ISSUE #2398: Standardize documentation for CMake build directory
We have docs that use `cmake -Bbuild-output` and `cmake -B.build`. AFAIK, this is a matter of preference, so I think we should pick one and use it consistently in our docs. Though I don't have much preference here, I've heard preference for `.build`.

devjgm:
Fixes #2398

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2399)


devjgm:
Yep, I think I did the first change all wrong. I've now read the ci/test-readme/README.md file and I think this may be right this time. PTAL

frankyn:
Question @devjgm based on context provided in issue #2398.
Is there more background on the preference for `.build`? 

My only concern with using `.build` is it's also a hidden path. I'm not convinced with using a hidden path `.build` over `build-output` for developer readmes.

remyabel:
If we're going to change it, shouldn't we change it to consistently be `build-output` so that users are potentially not affected by the switch.

devjgm:
First off, this change is pretty easy for me to generate, so I'm totally happy to change .build -> something-else. That's no problem. I don't personally have much of a preference here, other than for consistency in our docs.

.build = SGTM
build-outpt = SGTM

I think @coryan was in favor of .build. I'd be curious to hear his thoughts.

coryan:
To be clear, I *personally* prefer `.build` because (a) it is less typing, and (b) I really do not need to see those directories in my `ls` output.

I cannot claim which is better for documentation purposes, the point from @frankyn that using `build-output/` for new developers might be a good one.  Unfortunately we have two audiences: (1) customers (or developers of the library) that are fairly familiar with CMake, know what these directories are, and basically want to ignore them, (2) customers that might be unfamiliar with CMake and where using visible directories, with an obvious name triggers less "What is this?" moments.

I think the documentation should cater for group (2), and group (1) are clever enough to say: "Why such a long name? Forget it, I will use `.bu` for my builds".

Also, Bazel creates visible directories, so it might be less surprising to use visible directories everywhere.

Sorry if I seem to be flip-flopping on this.

frankyn:
Thanks for the background context for `.build` @coryan. 

I think keeping the directory in non-hidden folder would be helpful here in documentation `build-output`. 
I understand the case for shorter and hidden build paths, and think it's up to the developer to decide based on the example provided in documentation.

@remyabel and @devjgm agree on `build-output` and the goal of this PR is to standardize documentation for binary output. 

I vote to move forward with `build-output` for now. We can always revisit this in the future based on developer feedback.

devbww:
If we're going to change/unify, and we're going to vote, I would cast mine for "cmake-out", to match "bazel-out".  [Would also require a .gitignore tweak.]

devjgm:
FTR, I was going to suggest the same thing as @devbww. I think `cmake-out` would be nice since it is clearly associated with cmake and mirrors the `bazel-out` directory. Any objections/support for this idea?

coryan:
`cmake-out` also SGTM.

frankyn:
Nope! Sgtm!

devjgm:
Changed to use `cmake-out` everywhere. I believe all comments addressed. PTAL

devjgm:
> Add `cmake-out` to `.gitignore`.

It is.

PR #2402, opened 2019-04-04
Implement bigtable::Table::AsyncReadModifyWriteRow

### LINKED ISSUE #1491: Implement bigtable::Table::ReadModifyWriteRow
This function will implement an asynchronous version of `Table::ReadModifyWriteRow`. The function will have the following signature:

```C++
future<Row> AsyncReadModifyWriteRow(
    CompletionQueue& cq, std::string row_key,
    bigtable::ReadModifyWriteRule rule, Args&&... rules);
};
```

- [ ] Implement the function.

rsimion:
Fixes #1491 

Part of #2097

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2402)


google/cloud/bigtable/examples/data_async_snippets.cc
```diff
@@ -166,6 +166,42 @@ void AsyncCheckAndMutate(cbt::Table table, cbt::CompletionQueue cq,
...
 }
+
+void AsyncReadModifyWrite(cbt::Table table, cbt::CompletionQueue cq,
+                          std::vector<std::string> argv) {
+  if (argv.size() != 2U) {
+    throw Usage{
+        "async-read-modify-write: <project-id> <instance-id> <table-id>"};
+  }
+
+  //! [async read modify write]
```
coryan: Please add a `// TODO(#2404) - remove hard-coded key values` just before this. I just created that bug btw, no way for you to know about it.

PR #2407, opened 2019-04-05
CMake: Use CURL::libcurl instead of CURL::CURL

coryan:
This matches what libcurl exports and also what FindCURL defines as of CMake 3.12+.

See: https://cmake.org/cmake/help/v3.12/module/FindCURL.html

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2407)


remyabel:
@past-due I think you need to write "I signed it!" here to make googlebot happy.

coryan:
For the record, I checked that @past-due has signed the CLA using our internal tool and then manually set the `cla: yes` label.

coryan:
Adding new folks so they can get more familiar with the workflows.

PR #2410, opened 2019-04-05
Create HMAC Key samples.

coryan:
We want separate samples to enable and disable a HMAC Key for
consistency with the samples for other languages.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2410)


coryan:
/cc: @jkwlui @frankyn in case you want to review the tags or something.

PR #2411, opened 2019-04-05
Enable HMAC Key integration tests in production.

coryan:
The HMAC Key support arrived (for whitelisted projects) to production,
so we can enable this for our integration tests.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2411)


coryan:
I am closing this because there are too many things to fix before we can enable these tests in production.

PR #2412, opened 2019-04-05
Fix region tags for a couple of storage samples.

coryan:
In one case the tag was rather long, so I removed the (unused) Doxygen
tags.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2412)

### DIFF
--- a/google/cloud/storage/examples/storage_bucket_samples.cc
+++ b/google/cloud/storage/examples/storage_bucket_samples.cc
@@ -640,7 +640,7 @@ void GetBucketLifecycleManagement(google::cloud::storage::Client client,
     }
     std::cout << "\n";
   }
-  //! [END storage_view_lifecycle_management_configuration]
+  // [END storage_view_lifecycle_management_configuration]
   (std::move(client), bucket_name);
 }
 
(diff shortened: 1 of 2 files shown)


google/cloud/storage/examples/storage_bucket_samples.cc
```diff
@@ -615,7 +615,7 @@ void GetBucketLifecycleManagement(google::cloud::storage::Client client,
     throw Usage{"get-bucket-lifecycle-management <bucket-name>"};
   }
   auto bucket_name = ConsumeArg(argc, argv);
-  //! [view_lifecycle_management] [START storage_view_lifecycle_management]
+  // [START storage_view_lifecycle_management_configuration]
```
devbww: I don't yet grok when we do/don't need each variety of tag, if substrings matter, etc.

google/cloud/storage/examples/storage_object_acl_samples.cc
```diff
@@ -52,7 +52,7 @@ void ListObjectAcl(gcs::Client client, int& argc, char* argv[]) {
   }
   auto bucket_name = ConsumeArg(argc, argv);
   auto object_name = ConsumeArg(argc, argv);
-  //! [list object acl]
+  //! [list object acl] [storage_print_file_acl]
```
devbww: Should we have START/END somewhere here, for example?

google/cloud/storage/examples/storage_bucket_samples.cc
coryan: I sent a doc about this earlier this evening, happy to answer specific questions there.

devbww: Thanks.

PR #2419, opened 2019-04-05
 Clean up relational operators

mr-salty:
Clean up relational operators.  …
* Convert all member operators to free functions.
* Fix a bug in operator== for CommonMetadata (it wasn't checking owner)
* Standardize on using std::rel_ops
* Move 'friend' declarations to class private sections.
* changed operator<< for RowRange to not require friendship

In cases where I was able to implement the operators using the existing
public API I did so, otherwise I preferred making them friends over adding
to the API.

I left PaginationIterator operators defined inline in the class because
all the verbiage necessary (forward declaration, friend declaration,
then definition) to make them friends makes it less readable IMO.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2419)

### DIFF
--- a/google/cloud/bigtable/row_range.h
+++ b/google/cloud/bigtable/row_range.h
@@ -18,6 +18,7 @@
 #include "google/cloud/bigtable/internal/prefix_range_end.h"
 #include <google/bigtable/v2/data.pb.h>
 #include <chrono>
+#include <utility>
 
 namespace google {
 namespace cloud {
@@ -144,9 +145,6 @@ class RowRange {
    */
   std::pair<bool, RowRange> Intersect(RowRange const& range) const;
 
-  bool operator==(RowRange const& rhs) const;
-  bool operator!=(RowRange const& rhs) const { return !(*this == rhs); }
-
   /// Return the filter expression as a protobuf.
   ::google::bigtable::v2::RowRange const& as_proto() const& {
     return row_range_;
@@ -156,8 +154,6 @@ class RowRange {
   ::google::bigtable::v2::RowRange&& as_proto() && {
     return std::move(row_range_);
   }
-
-  friend std::ostream& operator<<(std::ostream& os, RowRange const& x);
 
  private:
   /// Private to avoid mistaken creation of uninitialized ranges.
@@ -172,6 +168,12 @@ class RowRange {
  private:
   ::google::bigtable::v2::RowRange row_range_;
 };
+
+bool operator==(RowRange const& lhs, RowRange const& rhs);
+
+inline bool operator!=(RowRange const& lhs, RowRange const& rhs) {
+  return std::rel_ops::operator!=(lhs, rhs);
+}
 
 /// Print a human-readable representation of the range, mostly for testing.
 std::ostream& operator<<(std::ostream& os, RowRange const& x);
--- a/google/cloud/firestore/field_path.h
+++ b/google/cloud/firestore/field_path.h
@@ -18,6 +18,7 @@
 #include <iostream>
 #include <regex>
 #include <string>
+#include <utility>
 #include <vector>
 
 namespace google {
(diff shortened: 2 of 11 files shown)


google/cloud/bigtable/row_range.h
```diff
@@ -173,6 +169,12 @@ class RowRange {
   ::google::bigtable::v2::RowRange row_range_;
 };
 
+bool operator==(RowRange const& lhs, RowRange const& rhs);
```
remyabel: Any reason why this implementation shouldn't be moved to the header file? I know it was already in the `.cc` file, but I'm wondering.

google/cloud/firestore/field_path.h
```diff
@@ -183,6 +139,26 @@ class FieldPath {
    */
   bool valid_;
 };
+
+bool operator==(FieldPath const& lhs, FieldPath const& rhs);
+bool operator<(FieldPath const& lhs, FieldPath const& rhs);
```
remyabel: It looks like you forgot to move the comment blocks as well.

google/cloud/storage/internal/access_control_common.h
```diff
@@ -151,6 +140,25 @@ class AccessControlCommon {
...
+inline bool operator==(AccessControlCommon const& lhs,
+                       AccessControlCommon const& rhs) {
+  // Start with id, bucket, etag because they should fail early, then
+  // alphabetical for readability.
+  return lhs.id() == rhs.id() && lhs.bucket() == rhs.bucket() &&
+         lhs.etag() == rhs.etag() && lhs.domain() == rhs.domain() &&
+         lhs.email() == rhs.email() && lhs.entity() == rhs.entity() &&
+         lhs.entity_id() == rhs.entity_id() && lhs.kind() == rhs.kind() &&
+         lhs.has_project_team() == rhs.has_project_team() &&
+         (!lhs.has_project_team() ||
```
remyabel: I am confused by these two lines. Would `lhs.has_project_team() && rhs.has_project_team()` instead of `lhs.has_project_team() == rhs.has_project_team()` be more readable for the intent here? Also, I think this will evaluate to true if `lhs.has_project_team` is `false`. I'm not sure if that's what you want.

google/cloud/bigtable/row_range.h
mr-salty: The implementation is 37 lines which is pretty far above the 10 line threshold the style guide recommends for inlining.

google/cloud/firestore/field_path.h
mr-salty: I dropped them intentionally (although I'm willing to listen to reasons for keeping them).  My rationale is that they fall under the "simple and obvious" clause of the style guide for function comments, since the meaning of the operators is well-known and the comments were just reinforcing that. Actually, the comments make me wonder "is this doing something unexpected?", which is forbidden by the style guide: "Define overloaded operators only if their meaning is obvious, unsurprising, and consistent with the corresponding built-in operators".

google/cloud/storage/internal/access_control_common.h
mr-salty: Maybe it's a little clever but it also seems like a pretty well-established idiom for writing this sort of thing without using ifs (although [optional's operator==](https://github.com/googleapis/google-cloud-cpp/blob/440439dcd72f362ff5eccfea34790a36b32a8523/google/cloud/optional.h#L235) takes that approach). The logic is the same, for them to be equal we need both:

1. lhs and rhs must both either: have a value, or not have a value (hence == rather than &&)
2. if they both have a value, those values must be equal

clause 2 can be restated as "either they both **don't** have values, or those values are equal" - except since clause 1 already established equality of has_XXX() we only need to check one.

remyabel: Thanks for the explanation. 

> lhs and rhs must both either: have a value, or not have a value (hence == rather than &&)

If they don't have a value, then we **don't** want to run the following check (otherwise we are deferencing a NULL pointer). Correct me if I'm wrong.

> if they both have a value, those values must be equal

What I'm concerned about is `if (!flag || a == b)`. As I understand it, if flag is `false`, the entire statement is `true`.

remyabel: OK disregard, your comment. I took a look at the implementation for optional and your approach LGTM.

remyabel:
It looks like you pulled in some unnecessary commits. Can you rebase your changes?

Issue #1595, opened 2018-12-05
Support blob signature via project service account.


coryan:
This would be handy when trying to create signed URLs for GCE or GAE credentials:

https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts/signBlob

More on:

https://github.com/GoogleCloudPlatform/google-cloud-cpp/pull/1591#discussion_r239198179

frankyn:
Use this instead: https://cloud.google.com/iam/credentials/reference/rest/v1/projects.serviceAccounts/signBlob

PR #2424, opened 2019-04-08
Implement SignBlobRequest and SignBlobResponse.

coryan:
This is part of the changes to use the APIs to sign blobs. That will
allow us to sign URLs and policy docs, even when the credentials are
not service account credentials.

Look at #1595 for further motivation.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2424)


google/cloud/storage/internal/sign_blob_requests.cc
```diff
@@ -0,0 +1,62 @@
...
+#include "google/cloud/storage/internal/nljson.h"
+
+namespace google {
+namespace cloud {
+namespace storage {
+inline namespace STORAGE_CLIENT_NS {
+namespace internal {
+
+std::ostream& operator<<(std::ostream& os, SignBlobRequest const& r) {
+  auto join = [](std::vector<std::string> const& v) -> std::string {
```
mr-salty: are we not ready to use absl StrJoin() here?

google/cloud/storage/internal/sign_blob_requests_test.cc
```diff
@@ -0,0 +1,79 @@
...
+  std::string const text = R"""({
+      "keyId": "test-key-id",
+      "signedBlob": "test-signed-blob"
+})""";
+
+  auto parsed = SignBlobResponse::FromHttpResponse(text).value();
+
+  std::ostringstream os;
+  os << parsed;
+  std::string actual = std::move(os).str();
```
mr-salty: the same thing is written above as 'auto str = os.str();' - make them consistent (aside: does this move actually help?)

google/cloud/storage/internal/sign_blob_requests.cc
coryan: Because we don't have Abseil at all in the project?

google/cloud/storage/internal/sign_blob_requests_test.cc
coryan: Let me create a new PR to fix the inconsistency.
The move supposedly would help in the future (allegedly C++ 20 or so, which btw our customers might use to compile the code, even if we do not).

PR #2426, opened 2019-04-08
Implement get-bucket-class-and-location example.

### LINKED ISSUE #2414: Create the  storage_get_bucket_class_and_location sample.
Create a sample that "Retrieve the storage class and location for a GCS bucket".  The tag should be `storage_get_bucket_class_and_location`.  Details (such as they exist) are available inside Google at go/gcs-sample-tracker

Other bucket samples are in:

https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/storage/examples/storage_bucket_samples.cc

coryan:
This is one of the standard examples for GCS client libraries, it should
show how to get the storage class and location of a bucket.

This fixes #2414.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2426)

### DIFF
--- a/google/cloud/storage/examples/storage_bucket_samples.cc
+++ b/google/cloud/storage/examples/storage_bucket_samples.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/storage/client.h"
+
 #include <functional>
 #include <iostream>
 #include <map>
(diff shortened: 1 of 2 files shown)


google/cloud/storage/examples/storage_bucket_samples.cc
```diff
@@ -325,6 +326,33 @@ void PatchBucketStorageClassWithBuilder(google::cloud::storage::Client client,
...
+        bucket_name, gcs::Fields("name,location,storageClass"));
+
+    if (!bucket_metadata) {
+      throw std::runtime_error(bucket_metadata.status().message());
+    }
+
+    std::cout << "Bucket " << bucket_metadata->name()
+              << " default storage class is "
+              << bucket_metadata->storage_class()
+              << ", and the default location is " << bucket_metadata->location()
```
frankyn: "bucket location" instead of "default location"

google/cloud/storage/examples/storage_bucket_samples.cc
```diff
@@ -325,6 +326,33 @@ void PatchBucketStorageClassWithBuilder(google::cloud::storage::Client client,
...
+  if (argc != 2) {
+    throw Usage{"get-bucket-class-and-location <bucket-name>"};
+  }
+  auto bucket_name = ConsumeArg(argc, argv);
+  // [START storage_get_bucket_class_and_location]
+  namespace gcs = google::cloud::storage;
+  using ::google::cloud::StatusOr;
+  [](gcs::Client client, std::string bucket_name) {
+    StatusOr<gcs::BucketMetadata> bucket_metadata = client.GetBucketMetadata(
+        bucket_name, gcs::Fields("name,location,storageClass"));
```
frankyn: I'm questioning if Fields should be selected in samples. 
It has additional overhead on the user knowing field names. I understand it's also a best practice to only select the fields you need. The sample should show a user the "happy path".

I'm open to discussing this more. If no objections, please remove the field selection.

google/cloud/storage/examples/storage_bucket_samples.cc
coryan: Fixed.

google/cloud/storage/examples/storage_bucket_samples.cc
coryan: No objection, removed.

PR #2437, opened 2019-04-09
Fix region tags for bigtable_hello_world.

coryan:
I checked this against the internal tracker, and fixed the region tags.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2437)

### DIFF
--- a/google/cloud/bigtable/examples/bigtable_hello_world.cc
+++ b/google/cloud/bigtable/examples/bigtable_hello_world.cc
@@ -17,7 +17,7 @@
 //! [bigtable includes] [START bigtable_hw_imports]
 #include "google/cloud/bigtable/table.h"
 #include "google/cloud/bigtable/table_admin.h"
-//! [bigtable includes] [START bigtable_hw_imports]
+//! [bigtable includes] [END bigtable_hw_imports]
 
 int main(int argc, char* argv[]) try {
   if (argc != 4) {
--- a/google/cloud/bigtable/examples/bigtable_hello_world.cc
+++ b/google/cloud/bigtable/examples/bigtable_hello_world.cc
@@ -14,10 +14,10 @@
 
 //! [all code]
 
-//! [bigtable includes] [START dependencies]
+//! [bigtable includes] [START bigtable_hw_imports]
 #include "google/cloud/bigtable/table.h"
 #include "google/cloud/bigtable/table_admin.h"
-//! [bigtable includes] [END dependencies]
+//! [bigtable includes] [START bigtable_hw_imports]
 
 int main(int argc, char* argv[]) try {
   if (argc != 4) {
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/examples/bigtable_hello_world.cc
```diff
@@ -14,10 +14,10 @@
 
 //! [all code]
 
-//! [bigtable includes] [START dependencies]
+//! [bigtable includes] [START bigtable_hw_imports]
 #include "google/cloud/bigtable/table.h"
 #include "google/cloud/bigtable/table_admin.h"
-//! [bigtable includes] [END dependencies]
+//! [bigtable includes] [START bigtable_hw_imports]
```
devbww: s/START/END/

google/cloud/bigtable/examples/bigtable_hello_world.cc
```diff
@@ -33,32 +33,32 @@ int main(int argc, char* argv[]) try {
...
-  //! [create table] [START creating_a_table]
+  //! [create table] [START bigtable_hw_create_table]
   // Define the desired schema for the Table.
   auto gc_rule = google::cloud::bigtable::GcRule::MaxNumVersions(1);
   google::cloud::bigtable::TableConfig schema({{"family", gc_rule}}, {});
 
   // Create a table.
   auto returned_schema = table_admin.CreateTable(table_id, schema);
-  //! [create table] [END creating_a_table]
+  //! [create table]
```
devbww: "END bigtable_hw_create_table" ?

google/cloud/bigtable/examples/bigtable_hello_world.cc
```diff
@@ -33,32 +33,32 @@ int main(int argc, char* argv[]) try {
...
 
   // Create an object to access the Cloud Bigtable Data API.
-  //! [connect data] [START connecting_to_bigtable]
+  //! [connect data]
```
devbww: "START bigtable_hw_create_table" ?

coryan: This is the same as above, I am effectively merging the `connecting_to_bigtable` and `creating_a_table` region tags into a single `bigtable_hw_create_table` region.

google/cloud/bigtable/examples/bigtable_hello_world.cc
coryan: I intentionally made the scope of `creating_a_table` larger.  The examples in other languages (see following links) both create the Table (as in the "the thing inside the service that holds data") and the Table *handle* (as in "the thing in the application memory that gives you access to that thing in the service").

https://github.com/GoogleCloudPlatform/php-docs-samples/blob/5dfa97f1b8694cc22f4d633e1234ddb6865f56fe/bigtable/src/helloworld/create_table.php#L34-L80

https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/4a9929abc4c03ed97d7964dbecf1c9d48b1354d8/bigtable/hello_world.rb#L18-L30

https://github.com/googleapis/nodejs-bigtable/blob/0d8412140ceb32bf70837c6d18d5d2ce47ffba78/samples/hello-world/index.js#L38-L53

google/cloud/bigtable/examples/bigtable_hello_world.cc
coryan: Ooops, fixed, thanks!

coryan:
PTAL.

google/cloud/bigtable/examples/bigtable_hello_world.cc
devbww: I see.  Thanks.

google/cloud/bigtable/examples/bigtable_hello_world.cc
devbww: Ack.

PR #2438, opened 2019-04-09
Output content-length-range parameters as numbers, not strings.

remyabel:
Previously it looked like:

```
["content-length-range", "0", "1000000"]
```

We want it to actually be like:

```
["content-length-range", 0, 1000000]
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2438)

### DIFF
--- a/google/cloud/storage/client_sign_policy_document_test.cc
+++ b/google/cloud/storage/client_sign_policy_document_test.cc
@@ -75,20 +75,19 @@ TEST(SignedPolicyDocumentIntegrationTest, Sign) {
             internal::FormatRfc3339(actual->expiration));
 
   EXPECT_EQ(
-      "eyJjb25kaXRpb25zIjpbWyJzdGFydHMtd2l0aCIsIiRrZXkiLCIiXSx7ImFjbCI6ImJ1Y2tl"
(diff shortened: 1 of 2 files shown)


coryan:
___
*[google/cloud/storage/internal/policy_document_request.cc, line 48 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2438#-Lc2L3tn-_lKuEM_iYeB:-Lc2L3tn-_lKuEM_iYeC:b8ee4zd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d18e2e4194877bb4766ed93c763500e522267bba/google/cloud/storage/internal/policy_document_request.cc#L48)):*
> ```cpp
>     } else {
>       if (elements.at(0) == "content-length-range") {
>         j["conditions"].push_back({elements.at(0), std::stol(elements.at(1)),
> ```

BTW, `std::stol()` might throw exceptions. Is that the behavior we want here? We need to support applications that disable exceptions altogether (don't get me started).  Or maybe there is a reason to believe this would never happen?  If so, can you add a comment.

remyabel:
Both `at` and `stol` have the potential to throw exceptions. I think we should fix #2440 first to return an error early and ensure that exceptions can never be thrown at this point.

PR #2443, opened 2019-04-09
Moved Status operator== and operator!= from member to free functions

### LINKED ISSUE #2391: Change Status member op== and op!= to free functions
When binary operators are implemented as member functions, they are subtly asymmetric since the LHS does not participate in any implicit conversions, while the RHS does. For example:

```cc
struct A {
  bool operator==(A const& a);
};

struct Foo {
  operator A() const;  // convertible to A;
};

Foo foo;
A a;
assert(a == foo); // OK
assert(foo == a);  // Won't compile
```

To ensure that binary operators are symmetric wrt implicit conversions, it's best to implement them as non-member functions. See also http://go/gh/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c161-use-nonmember-functions-for-symmetric-operators

Please change [op== and op!=](https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/status.h#L72) in `Status` to be non-members.

scotthart:
Fixes #2391.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2443)

### DIFF
--- a/google/cloud/status.h
+++ b/google/cloud/status.h
@@ -69,11 +69,6 @@ class Status {
 
   bool ok() const { return code_ == StatusCode::kOk; }
 
-  bool operator==(Status const& rhs) const {
-    return code() == rhs.code() && message() == rhs.message();
-  }
-  bool operator!=(Status const& rhs) const { return !(*this == rhs); }
-
   StatusCode code() const { return code_; }
   std::string const& message() const { return message_; }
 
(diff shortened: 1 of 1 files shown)


mr-salty:
___
*[google/cloud/status.h, line 89 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2443#-Lc3HGAd3UIG5bc2eQKw:-Lc3HwoSDgZu-UNbq-uW:bg3qc39) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e22bfa5978c149822a6df6901e3944e3c7877e5a/google/cloud/status.h#L89)):*

> ```
>  return !(lhs == rhs);
> ```

FWIW, when I was doing the other operators I changed all occurrences to this kind of thing to "return std::rel_ops::operator!=(lhs, rhs);" for consistency (previously we had some rel_ops and some "hand crafted").  You have to include <utility>

</blockquote></details>

oops, markdown ate the angle brackets so make that ```include <utility>```

coryan:
@scotthart this is ready to merge, do you want me to push the button or would you rather do it yourself?

PR #2449, opened 2019-04-10
Use SignBlob API if needed.

### LINKED ISSUE #1595: Support blob signature via project service account.

coryan:
When the local Credentials cannot sign a blob, either because the
credentials type does not support this, or because the credentials do
not match the signing service account we need to fallback on the
SignBlob API.

This fixes #1595 (though there is some cleanup that I want to do after this).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2449)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -2593,8 +2593,8 @@ class Client {
    * which is a POST request to GCS.
    *
    * @param document the policy document.
-   * @param options a list of optional parameters, this include:
-   *      `SigningAccount`, and `SigningAccountDeletegates`.
+   * @param options a list of optional parameters, this includes:
+   *      `SigningAccount`, and `SigningAccountDelegates`.
    *
    * @par Example
    * @snippet storage_bucket_samples.cc create signed policy document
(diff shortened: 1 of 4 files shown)


google/cloud/storage/client.h
```diff
@@ -2591,12 +2593,8 @@ class Client {
    * which is a POST request to GCS.
    *
    * @param document the policy document.
-   *
-   * @par Helper Functions
-   *
-   * The following functions create a `PolicyDocumentCondition` with less
-   * opportunities for typos: `StartsWith()`, `ExactMatchObject()`,
-   * `ExactMatch()`, `ContentLengthRange()`.
```
remyabel: Why is this comment deleted?

coryan: Argh... I thought these were the options, my bad, I will send a PR to restore them.

PR #2458, opened 2019-04-11
Include what you use: google/cloud/storage/version.h

devbww:
Ensure that every header that uses "inline namespace STORAGE_CLIENT_NS"
first includes google/cloud/storage/version.h.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2458)


devbww:
PTAL

PR #2459, opened 2019-04-11
Include what you use: google/cloud/bigtable/version.h

devbww:
Ensure that every header that uses "inline namespace BIGTABLE_CLIENT_NS"
first includes google/cloud/bigtable/version.h.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2459)


devbww:
PTAL

PR #2461, opened 2019-04-11
Use SFINAE to constrain applicability of the BulkMutation(M&&...) ctor

devbww:
The unrestricted BulkMutation perfect-forwarding constructor was a
better match than the (generated) copy constructor for a non-const
lvalue, so its embedded static_assert requiring convertibility to a
SingleRowMutation meant that non-const BulkMutation lvalues could
not be copied (so they could not be used as pass-by-value function
arguments, for example).

So, make the "convertibility to a SingleRowMutation" requirement a
SFINAE constraint instead, meaning the perfect-forwarding constructor
will be disabled when it does not apply, and that allows the copy
constructor to be found instead.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2461)

### DIFF
--- a/google/cloud/bigtable/mutations.h
+++ b/google/cloud/bigtable/mutations.h
@@ -416,13 +416,10 @@ class BulkMutation {
   }
 
   /// Create a muti-row mutation from a variadic list.
-  template <typename... M>
+  template <typename... M,
+            typename = typename std::enable_if<internal::conjunction<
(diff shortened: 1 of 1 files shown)


coryan:
The GitHub settings on the branch are set to "very strict", you need to rebase after I merged #2460. In general, each merge will require you to rebase the other PRs.

PR #2462, opened 2019-04-11
Add test for using google-cloud-cpp as a submodule.

### LINKED ISSUE #2395: Verify google-cloud-cpp can be used as sub-directory.
Some of out users want to compile `google-cloud-cpp` as a submodule, we should verify this works too.

remyabel:
Fixes #2395 

This depends on #2455

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2462)

### DIFF
--- a/ci/travis/build-docker.sh
+++ b/ci/travis/build-docker.sh
@@ -184,10 +184,16 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
   make -C "${TEST_INSTALL_MAKE_OUTPUT_DIR}" -f"${TEST_INSTALL_DIR}/Makefile" VPATH="${TEST_INSTALL_DIR}" CXX=${CXX}
   echo
   echo "${COLOR_YELLOW}Test installed libraries when used as a submodule.${COLOR_RESET}"
-  mkdir -p "${TEST_INSTALL_SUBMODULE_OUTPUT_DIR}"
+  mkdir -p "${TEST_INSTALL_SUBMODULE_OUTPUT_DIR}/submodule/google-cloud-cpp"
   cp -r "${TEST_INSTALL_DIR}"/* "${TEST_INSTALL_SUBMODULE_OUTPUT_DIR}"
-  git -C "${TEST_INSTALL_SUBMODULE_OUTPUT_DIR}/submodule" init
-  git -C "${TEST_INSTALL_SUBMODULE_OUTPUT_DIR}/submodule" submodule add https://github.com/googleapis/google-cloud-cpp
+  find . -mindepth 1 \( -path ./.git \
+      -o -path ./third_party \
+      -o -path './cmake-build-*' \
+      -o -path ./build-output \
+      -o -path ./.build \
+      -o -path ./_build \
+      -o -path ./cmake-out \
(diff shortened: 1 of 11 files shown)


coryan:
Quick question: did you rebase to pick the changes from #2455?

ci/travis/build-docker.sh
```diff
@@ -188,6 +188,13 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
   cp -r "${TEST_INSTALL_DIR}"/* "${TEST_INSTALL_SUBMODULE_OUTPUT_DIR}"
   git -C "${TEST_INSTALL_SUBMODULE_OUTPUT_DIR}/submodule" init
   git -C "${TEST_INSTALL_SUBMODULE_OUTPUT_DIR}/submodule" submodule add https://github.com/googleapis/google-cloud-cpp
+  find "${PROJECT_ROOT}" \( -path ./.git \
+          -o -path ./third_party \
+          -o -path './cmake-build-*' \
+          -o -path ./build-output \
+          -o -path ./.build \
+          -o -path ./_build \
+       \) -prune -exec cp '{}' "${TEST_INSTALL_SUBMODULE_OUTPUT_DIR}/submodule/google-cloud-cpp"'{}' \;
```
remyabel: This is actually broken. Let me attempt to fix it.

remyabel:
> Quick question: did you rebase to pick the changes from #2455?

I will after I fix my find command.

> Here and elsewhere: Do we need the git commands anymore?

Perhaps I'm misunderstanding something, but unless we actually run the `git submodule` commands, we're not really testing it, we're just building the project like normal. For example when I was testing this earlier, I ran into errors that were caught only by these commands in the tests.

> Do we need these git commands? In fact, I think this test works for both submodules and CMake's FetchContent right?

Not sure, I've never used `FetchContent` before. Let's save that for a future PR.

coryan:
> > Quick question: did you rebase to pick the changes from #2455?
> 
> I will after I fix my find command.

Ack. Ping me, I will kick off the Kokoro builds at that point.

> 
> > Here and elsewhere: Do we need the git commands anymore?
> 
> Perhaps I'm misunderstanding something, but unless we actually run the `git submodule` commands, we're not really testing it,

You are correct.  But the "submodule" aspect is not the problem, we do not need to test that `git submodule` works, we know it does (or if it does not, well, then it is not our problem). The real question is whether our code compiles as a sub-directory of a larger project.

Maybe the original bug should have said "Verify google-cloud-cpp can be used as a *sub-directory*", whether the code is put into that sub-directory by `git submodule`, `tar`, `cp`, or `FetchContent` does not really matter that much to us (I think).

> we're just building the project like normal. For example when I was testing this earlier, I ran into errors that were caught only by these commands in the tests.
> 
> > Do we need these git commands? In fact, I think this test works for both submodules and CMake's FetchContent right?
> 
> Not sure, I've never used `FetchContent` before. Let's save that for a future PR.

Ack, though I think your tests already check the general case of "Somehow the code gets into a sub-directory, does it compile?"

remyabel:
Problems should be fixed now. PTAL.

PR #2464, opened 2019-04-11
Implementation of `TableAdmin::AsyncDropRowsByPrefix`

### LINKED ISSUE #1498: Implement bigtable::TableAdmin::AsyncDropRowsByPrefix
We want to implement an asynchronous version of `DropRowsByPrefix` for the interfaces that use exceptions. The signature of this function should look like so:

```C++
 future<void> AsyncDropRowsByPrefix(
      std::string const& table_id,
      std::string row_key_prefix,
      CompletionQueue& cq);
};
```

- [ ] Implement the function.

manish-qlogic:
This commit includes

1. Implement the function.
2. Implement an integration test for this function.
3. Implement an example for this function.
4. Execute the example as part of the CI build.

This closes #1498.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2464)

### DIFF
--- a/google/cloud/bigtable/examples/run_examples_utils.sh
+++ b/google/cloud/bigtable/examples/run_examples_utils.sh
@@ -251,6 +251,9 @@ function run_all_table_admin_async_examples {
   # Use the same table in all the tests.
   local -r TABLE="sample-table-for-admin-${RANDOM}"
 
+  # Use sample row key wherever needed.
+  local -r ROW_KEY="sample-row-key-${RANDOM}"
+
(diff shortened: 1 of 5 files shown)


manish-qlogic:
@coryan ,

Reviewable comments are repetitive and little confusing.  I removed the hard coding for row-key and done corresponding changes in sh file.

Let me know if further changes are required.

coryan:
Sorry about the confusing reviewable comments, I got a bunch of 500 errors and I guess posted more than once.

PR #2467, opened 2019-04-11
Use the unparsed args in export_variables_to_bazel

mr-salty:
This allows properly passing arguments (i.e. YEAR) without them ending
up in the bazel file as (likely empty) variables.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2467)


coryan:
I just realized that the other CMake functions in this file have the same problem.  Do you want to fix them in this PR (which I am fine with) or a future one?

Issue #2463, opened 2019-04-11
Consider fixing warnings emitted by ShellCheck


remyabel:
[ShellCheck](https://github.com/koalaman/shellcheck) is a static analysis tool for shell scripts. I ran it on this codebase and here's the list of warnings emitted run through `uniq`. We can decide individually what warnings we care about and don't.

[SC2007](https://github.com/koalaman/shellcheck/wiki/SC2007): Use $((..)) instead of deprecated $[..]
[SC2016](https://github.com/koalaman/shellcheck/wiki/SC2016): Expressions don't expand in single quotes, use double quotes for that.
[SC2034](https://github.com/koalaman/shellcheck/wiki/SC2034): attempt appears unused. Verify use (or export if used externally).
[SC2044](https://github.com/koalaman/shellcheck/wiki/SC2044): For loops over find output are fragile. Use find -exec or a while read loop.
[SC2046](https://github.com/koalaman/shellcheck/wiki/SC2046): Quote this to prevent word splitting.
[SC2048](https://github.com/koalaman/shellcheck/wiki/SC2048): Use "$@" (with quotes) to prevent whitespace problems.
[SC2078](https://github.com/koalaman/shellcheck/wiki/SC2078): This expression is constant. Did you forget a $ somewhere?
[SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086): Double quote to prevent globbing and word splitting.
[SC2146](https://github.com/koalaman/shellcheck/wiki/SC2146): This action ignores everything before the -o. Use \( \) to group.
[SC2153](https://github.com/koalaman/shellcheck/wiki/SC2153): Possible misspelling: BUCKET_NAME may not be assigned, but bucket_name is.
[SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155): Declare and assign separately to avoid masking return values.
[SC2156](https://github.com/koalaman/shellcheck/wiki/SC2156): Injecting filenames is fragile and insecure. Use parameters.
[SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
[SC2166](https://github.com/koalaman/shellcheck/wiki/SC2166): Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
[SC2181](https://github.com/koalaman/shellcheck/wiki/SC2181): Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.
[...]

PR #2474, opened 2019-04-12
Fix portability/logical errors in shell scripts.

remyabel:
Part 3 of #2463

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2474)

### DIFF
--- a/ci/check-style.sh
+++ b/ci/check-style.sh
@@ -54,7 +54,7 @@ find google/cloud \( -name '*.cc' -o -name '*.h' \) -print0 \
 #       are obsoleted by the gRPC team, so we should not use them in our code.
 #     - Replace grpc::<BLAH> with grpc::StatusCode::<BLAH>, the aliases in the
 #       `grpc::` namespace do not exist inside google.
-for file in $(find google/cloud -name '*.h' -o -name '*.cc' -print); do
+while IFS= read -r -d '' file; do
   # We used to run run `sed -i` to apply these changes, but that touches the
   # files even if there are no changes applied, forcing a rebuild each time.
   # So we first apply the change to a temporary file, and replace the original
@@ -68,7 +68,7 @@ for file in $(find google/cloud -name '*.h' -o -name '*.cc' -print); do
   else
       mv -f "${file}.tmp" "${file}"
   fi
-done
+done < <(find google/cloud \( -name '*.h' -o -name '*.cc' \) -print0)
 
 # Apply buildifier to fix the BUILD and .bzl formatting rules.
 #    https://github.com/bazelbuild/buildtools/tree/master/buildifier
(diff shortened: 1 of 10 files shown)


ci/check-style.sh
```diff
@@ -54,7 +54,7 @@ find google/cloud \( -name '*.cc' -o -name '*.h' \) -print0 \
 #       are obsoleted by the gRPC team, so we should not use them in our code.
 #     - Replace grpc::<BLAH> with grpc::StatusCode::<BLAH>, the aliases in the
 #       `grpc::` namespace do not exist inside google.
-for file in $(find google/cloud -name '*.h' -o -name '*.cc' -print); do
+while IFS= read -r -d '' file; do
```
devbww: Perhaps it is clearer to say `-d $'\0'` to more directly match the `-print0`.

ci/check-style.sh
```diff
@@ -68,7 +68,7 @@ for file in $(find google/cloud -name '*.h' -o -name '*.cc' -print); do
   else
       mv -f "${file}.tmp" "${file}"
   fi
-done
+done < <(find google/cloud \( -name '*.h' -o -name '*.cc' \) -print0)
```
devbww: I think I would just use a pipe rather than process substitution.  That is, ...
```
find google/cloud \( -name '*.h' -o -name '*.cc' \) -print0 |
  while IFS= read -r -d $'\0' file; do
    ...
  done
```
Then the `find` is closer to the `read`.

ci/travis/dump-reports.sh
```diff
@@ -44,9 +44,9 @@ export -f dump_report
 # are copied into the scan-cmake-out directory by the build-docker.sh script.
 
 find "${BUILD_OUTPUT}" -name 'compat_report.html' \
-    -exec bash -c "dump_report '{}'" \; 2>/dev/null || \
+    -exec bash -c "dump_report '$1'" _ {} \; 2>/dev/null || \
```
devbww: Don't the quotes have to be around the other way here (i.e., `bash -c 'dump_report "$1"' _ {}`) lest the `$1` gets expanded too early?

ci/travis/dump-reports.sh
```diff
@@ -44,9 +44,9 @@ export -f dump_report
...
   echo "No ABI compability reports found."
 
 find scan-cmake-out/ -name '*.html' \
-    -exec bash -c "dump_report '{}'" \; 2>/dev/null || \
+    -exec bash -c "dump_report '$1'" _ {} \; 2>/dev/null || \
```
devbww: Ditto.

ci/travis/dump-reports.sh
remyabel: You are correct. Good catch.

remyabel:
PTAL.

PR #2475, opened 2019-04-12
Declare and assign local variables separately.

remyabel:
Part 4 of #2463

ShellCheck warns that `local name=$(...)` masks the return value of the subcommand. There are three options:

- Use `-r` if the variable should be read-only
- Use `|| true` to explicitly mask the return value
- Close this PR if we just don't care about the return value at all

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2475)

### DIFF
--- a/ci/define-example-runner.sh
+++ b/ci/define-example-runner.sh
@@ -49,7 +49,8 @@ run_example() {
   local example=$2
   shift 2
   local arguments=$*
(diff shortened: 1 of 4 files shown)


coryan:
Sigh, @remyabel do you mind rebasing so we can merge?

PR #2476, opened 2019-04-12
Put quotes around variables where necessary.

remyabel:
Part 5 of #2463

I intentionally ignore variables that are meant to be arguments or lists because when enclosing them in quotes, things break. They are false positives and we need an inline comment to suppress the warning.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2476)

### DIFF
--- a/ci/define-example-runner.sh
+++ b/ci/define-example-runner.sh
@@ -48,7 +48,7 @@ run_example() {
   local program_path=$1
   local example=$2
   shift 2
-  local arguments=$*
+  local -a arguments=( "$@" )
   local program_name
   program_name="$(basename "${program_path}")"
 
@@ -60,13 +60,9 @@ run_example() {
   log="$(mktemp -t "run_example.XXXXXX")"
   echo    "${COLOR_GREEN}[ RUN      ]${COLOR_RESET}" \
       " ${program_name} ${example} running"
-  # We want ${arguments} to expand as separate parameters.
-  # shellcheck disable=SC2086
-  echo "${program_path}" "${example}" ${arguments} >"${log}"
+  echo "${program_path}" "${example}" "${arguments[@]}" >"${log}"
   set +e
-  # Ditto about ${arguments}.
-  # shellcheck disable=SC2086
-  ${program_path} "${example}" ${arguments} >>"${log}" 2>&1 </dev/null
+  "${program_path}" "${example}" "${arguments[@]}" >>"${log}" 2>&1 </dev/null
   if [ $? = 0 ]; then
     echo  "${COLOR_GREEN}[       OK ]${COLOR_RESET}" \
         " ${program_name} ${example}"
--- a/ci/define-example-runner.sh
+++ b/ci/define-example-runner.sh
@@ -50,9 +50,9 @@ run_example() {
   shift 2
   local arguments=$*
   local program_name
-  program_name=$(basename ${program_path})
+  program_name="$(basename "${program_path}")"
 
-  if [ ! -x ${program_path} ]; then
+  if [ ! -x "${program_path}" ]; then
     echo "${COLOR_YELLOW}[  SKIPPED ]${COLOR_RESET}" \
         " ${program_name} is not compiled"
     return
@@ -60,9 +60,11 @@ run_example() {
   log="$(mktemp -t "run_example.XXXXXX")"
   echo    "${COLOR_GREEN}[ RUN      ]${COLOR_RESET}" \
       " ${program_name} ${example} running"
-  echo ${program_path} ${example} ${arguments} >"${log}"
+  # shellcheck disable=SC2086
+  echo "${program_path}" "${example}" ${arguments} >"${log}"
   set +e
-  ${program_path} ${example} ${arguments} >>"${log}" 2>&1 </dev/null
+  # shellcheck disable=SC2086
+  ${program_path} "${example}" ${arguments} >>"${log}" 2>&1 </dev/null
   if [ $? = 0 ]; then
     echo  "${COLOR_GREEN}[       OK ]${COLOR_RESET}" \
         " ${program_name} ${example}"
@@ -97,12 +99,12 @@ run_example_usage() {
     exit 1
   fi
 
-  local program_path=$1
+  local program_path="$1"
   shift 1
   local program_name
-  program_name=$(basename ${program_path})
+  program_name="$(basename "${program_path}")"
 
-  if [ ! -x ${program_path} ]; then
+  if [ ! -x "${program_path}" ]; then
     echo "${COLOR_YELLOW}[  SKIPPED ]${COLOR_RESET}" \
         " ${program_name} is not compiled"
     return
@@ -110,7 +112,7 @@ run_example_usage() {
   log="$(mktemp -t "run_example.XXXXXX")"
   echo    "${COLOR_GREEN}[ RUN      ]${COLOR_RESET}" \
       " ${program_name} running"
-  echo ${program_path} >"${log}"
+  echo "${program_path}" >"${log}"
   set +e
   ${program_path} >>"${log}" 2>&1 </dev/null
   if [ $? != 0 ] && grep -q 'Usage' "${log}"; then
@@ -121,7 +123,7 @@ run_example_usage() {
     echo    "${COLOR_RED}[    ERROR ]${COLOR_RESET}" \
         " ${program_name}"
     echo
-    dump_log ${log}
+    dump_log "${log}"
     if [ -f "${EMULATOR_LOG}" ]; then
       dump_log "${EMULATOR_LOG}"
     fi
--- a/ci/define-example-runner.sh
+++ b/ci/define-example-runner.sh
@@ -60,9 +60,13 @@ run_example() {
   log="$(mktemp -t "run_example.XXXXXX")"
   echo    "${COLOR_GREEN}[ RUN      ]${COLOR_RESET}" \
       " ${program_name} ${example} running"
-  echo "${program_path}" "${example}" "${arguments[@]}" >"${log}"
+  # We use parameter expansion for ${arguments} because set -u doesn't like
+  # empty arrays on older versions of Bash (which some of our builds use). The
+  # expression ${parameter+word} will expand word only if parameter is not
+  # unset.
+  echo "${program_path}" "${example}" "${arguments[@]+"${arguments[@]}"}" >"${log}"
   set +e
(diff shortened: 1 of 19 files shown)


coryan:
@remyabel can you rebase when you get a chance?

ci/define-example-runner.sh
```diff
@@ -49,19 +49,23 @@ run_example() {
...
+  if [ ! -x "${program_path}" ]; then
     echo "${COLOR_YELLOW}[  SKIPPED ]${COLOR_RESET}" \
         " ${program_name} is not compiled"
     return
   fi
   log="$(mktemp -t "run_example.XXXXXX")"
   echo    "${COLOR_GREEN}[ RUN      ]${COLOR_RESET}" \
       " ${program_name} ${example} running"
-  echo ${program_path} ${example} ${arguments} >"${log}"
+  # We want ${arguments} to expand as separate parameters.
```
devbww: I would think we don't want `${arguments}` to expand as separate parameters, but rather that we want to pass along exactly the arguments we were given.  That is, something more along the lines of ...
```
local -a arguments=( "$@" )
${program_path} "${example}" "${arguments[@]}"
```

ci/define-example-runner.sh
```diff
@@ -49,19 +49,23 @@ run_example() {
...
+  # shellcheck disable=SC2086
+  echo "${program_path}" "${example}" ${arguments} >"${log}"
   set +e
-  ${program_path} ${example} ${arguments} >>"${log}" 2>&1 </dev/null
+  # Ditto about ${arguments}.
+  # shellcheck disable=SC2086
+  ${program_path} "${example}" ${arguments} >>"${log}" 2>&1 </dev/null
```
devbww: I too am unsure why `${program_path}` does not deserve quoting in such contexts.

ci/travis/build-docker.sh
```diff
@@ -84,6 +84,9 @@ if [ "${USE_LIBCXX:-}" = "yes" ]; then
   cmake_extra_flags="${cmake_extra_flags} -DGOOGLE_CLOUD_CPP_USE_LIBCXX=ON"
 fi
 
+# We want ${cmake_extra_flags} and ${CMAKE_FLAGS} to expand as separate
```
devbww: Similar comment to before.  We should probably be using arrays.

google/cloud/bigtable/tools/run_emulator_utils.sh
```diff
@@ -94,6 +94,8 @@ function wait_until_emulator_connects {
   connected=no
   local -r attempts=$(seq 1 8)
   for attempt in ${attempts}; do
+    # We want ${CBT_ARGS} to expand as separate arguments.
```
devbww: Here is doesn't matter too much because we know the exact contents of `${CBT_ARGS}`, but I still think I'd use arrays and avoid the special casing.

devbww:
> Should that be "${BINDIR}/check-include-guards.gawk" in case BINDIR ends with a space?

It seems to me that it doesn't matter.

Something can be said for just quoting the unknown part, and something can be said for simply quoting the whole thing.  I guess just being consistent is the best path.

ci/travis/build-docker.sh
remyabel: I agree for `${cmake_extra_flags}`. But `CMAKE_FLAGS` is a user-passed environment variable.

remyabel:
I implemented your suggestions and now one of the tests is failing:

```
./table_admin_snippets check-consistency project-20171-1366 in-19042-32397 sample-table-for-admin-1620 
TokenFor-projects/project-20171-1366/instances/in-19042-32397/tables/sample-table-for-admin-1620
Standard C++ exception raised: CheckConsistency(name=projects/project-20171-1366/instances/in-19042-32397/tables/sample-table-for-admin-1620) token "\nTokenFor-projects/project-20171-1366/instances/in-19042-32397/tables/sample-table-for-admin-1620" not valid
```

Either I made a typo somewhere or we've exposed a bug. For some reason the token has a newline in it.

ci/define-example-runner.sh
remyabel: I was wrong. There's no reason not to quote it.

remyabel:
For now I've trimmed the newline. I (or someone else) can figure out why the generated token has a newline in it later. PTAL.

ci/travis/build-docker.sh
devbww: So should we do `${cmake_extra_flags}` then?  I agree that it isn't an issue as we set the contents and there are no special characters, but not needing to call it out is nice.

google/cloud/bigtable/examples/run_examples_utils.sh
```diff
@@ -221,7 +221,7 @@ function run_all_table_admin_examples {
   run_example ./table_admin_snippets modify-table "${project_id}" "${INSTANCE}" "${TABLE}"
   run_example ./table_admin_snippets wait-for-consistency-check "${project_id}" "${INSTANCE}" "${TABLE}"
   run_example ./table_admin_snippets generate-consistency-token "${project_id}" "${INSTANCE}" "${TABLE}"
-  local token="$(./table_admin_snippets generate-consistency-token ${project_id} ${INSTANCE} ${TABLE} | awk '{print $5}')"
+  local token="$(./table_admin_snippets generate-consistency-token "${project_id}" "${INSTANCE}" "${TABLE}" | awk '{print $5}' | tr -d '\n')"
```
devbww: It looks like it is probably because google/cloud/bigtable/examples/table_admin_snippets.cc first emits a (useless) empty line (line 286).  Lets get rid of that and see how things go.

coryan:
@remyabel when you have a chance, please rebase and remove the conflicts.

ci/travis/build-docker.sh
remyabel: I forgot to address this. Let me fix it.

remyabel:
Sigh. We hit a edge case. According to [StackOverflow](https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u/39687362#39687362), `set -u` doesn't like empty arrays on older versions of Bash. Unless there's any objections, I'll add the workaround `"${cmake_extra_flags[@]+"${cmake_extra_flags[@]}"}"` with a comment explaining what it means.

coryan:
Ugh... @devbww ?

google/cloud/bigtable/examples/run_examples_utils.sh
devbww: Did we find any joy here?

remyabel: i did take a look but this is line 286:

https://github.com/googleapis/google-cloud-cpp/blob/8dcf3bded1598df71603b8ff6038a2a09e68a110/google/cloud/bigtable/examples/table_admin_snippets.cc#L286

Did you mean to point me to a different line?

coryan: That file has a bit of new code in `master`, I am pretty sure @devbww meant this:

https://github.com/googleapis/google-cloud-cpp/blob/8dcf3bded1598df71603b8ff6038a2a09e68a110/google/cloud/bigtable/examples/table_admin_snippets.cc#L439

devbww:
Ugh ... yes, but I'm happy with the workaround.

Alternatively, can we use bash 4.4?

google/cloud/bigtable/examples/run_examples_utils.sh
devbww: Yeah ... that one.

coryan:
> Alternatively, can we use bash 4.4?

I would prefer if we did not. Upgrading stuff on the supported distributions to get the code to compile feels like cheating.  And the users may want to run the tests on their preferred platform.

google/cloud/bigtable/examples/run_examples_utils.sh
remyabel: My bad. Fixed now.

ci/travis/build-docker.sh
```diff
@@ -83,11 +83,15 @@ if [ "${USE_LIBCXX:-}" = "yes" ]; then
   cmake_extra_flags+=( "-DGOOGLE_CLOUD_CPP_USE_LIBCXX=ON" )
 fi
 
-# We want ${CMAKE_FLAGS} to expand as separate arguments.
+# We add ${cmake_extra_flags[@]} to itself is because set -u doesn't like empty
```
devbww: I'm not sure I understand "add to itself" here.  Is that a remnant of a different workaround?

remyabel: No, I just worded it poorly. Fixed.

devjgm:
Re: upgrading bash... fwiw, bash 4.4 was released in early in 2016 (https://ftp.gnu.org/gnu/bash/), so it's pretty old. bash 5.0 is current. What version of bash are our tests running with, and where are they picking it up?

coryan:
> Re: upgrading bash... fwiw, bash 4.4 was released in early in 2016 (https://ftp.gnu.org/gnu/bash/), so it's pretty old.

In the "Real World":tm: 2016 is not "pretty old", it is "we are just starting to look at it".

CentOS 7 which is supported by GCE, was released in 2014, it goes EOL in 2020 (if we are lucky).  Ubuntu:16.04 was released in 2016, it goes EOL in 20**24**. 

The timescales to support enterprise are *loooong*.

> bash 5.0 is current. What version of bash are our tests running with, and where are they picking it up?

We compile and run the tests on a lot of distros, and the the builds pick up bash from whatever comes with the distro. So the short answer is "it depends", but a sample:

| Distro | Bash Version |
| ------- | ------------------ |
| Centos:7 | 4.2.46 |
| openSUSE:Leap | 4.4.23 |
| Ubuntu:14.04 | 4.3.11 |
| Ubuntu:16.04 | 4.3.48 |

BTW, you can just use commands like this to figure out what comes with the OS:

```console
docker run centos:7 /bin/bash --version
```

remyabel:
> CentOS 7 which is supported by GCE, was released in 2014, it goes EOL in 2020 (if we are lucky). 

It's actually CentOS 6 that goes EOL in 2020. CentOS 7 (what we're targetting) will go EOL in 2024. The rest of your explanation is still valid, of course.

> Re: upgrading bash... fwiw, bash 4.4 was released in early in 2016 (https://ftp.gnu.org/gnu/bash/), so it's pretty old.

These older versions of bash will receive security/maintenance patches and the like which a custom-install of Bash would not receive.

devjgm:
Ok, I'm sold on not upgrading bash. Thanks

coryan:
@remyabel thanks for these fixes.

PR #2477, opened 2019-04-12
Adds spanner_tool for easy experimenting

devjgm:
This PR adds a `spanner_tool` binary that we can use to easily experiment with the spanner API in C++. It currently only calls spanner's `ListDatabases` RPC, which is kinda like a hello-world program. It can be extended to support others.

It only works with bazel at this point because I don't really know how to make it work with cmake yet :-) But that can be done in a later PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2477)

### DIFF
--- a/bazel/googleapis.BUILD
+++ b/bazel/googleapis.BUILD
@@ -207,7 +207,7 @@ cc_grpc_library(
     srcs = ["google/spanner/v1/mutation.proto"],
     deps = [
         ":google_api_annotations",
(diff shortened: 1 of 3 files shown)


devjgm:
OK, thanks. I'm disconnecting now for a flight. Feel free to merge this PR if the builds are good and someone wants to use this `spanner_tool`. Otherwise, I'll merge later today or Monday.

PR #2479, opened 2019-04-12
Purge some newly-introduced trailing whitespace

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2479)


coryan:
Can we change `ci/check-style.sh` to enforce this? I really don't like us to spend time doing formatting that the robots can do for us.

devbww:
> Can we change ci/check-style.sh to enforce this?

I'll take a look.

For those playing along at home, we only really notice this when importing the code into environments with stricter formatting rules.

PR #2480, opened 2019-04-12
Support P12 key files.

### LINKED ISSUE #1193: Decide whether we should support loading service account credentials from a P12 file
(Breaking this off from the more general issue in #579)

Other client libraries support loading credentials from .p12 files, but we *strongly* encourage users to use .json keyfiles instead.  However, if we plan to continue to support .p12 files in other languages' Cloud Storage client libraries, we should be consistent with those.

coryan:
Application developers can now use a P12 file to initialize the
credentials used to access GCS. This PR includes an integration test to
verify these credentials are usable for both accessing GCS as well as
signing URLs.

This fixes #1193.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2480)

### DIFF
--- a/ci/kokoro/ubuntu/build.sh
+++ b/ci/kokoro/ubuntu/build.sh
@@ -95,6 +95,8 @@ echo "================================================================"
 echo "Download dependencies for integration tests $(date)."
 echo "================================================================"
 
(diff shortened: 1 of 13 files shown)


coryan:
Sorry for the large binary blobs, but that is what a .p12 file is, and I needed both valid files, files with errors, etc.

PR #2484, opened 2019-04-15
Implementation of `TableAdmin::AsyncDropAllRows`

### LINKED ISSUE #1501: Implement bigtable::TableAdmin::AsyncDropAllRows
We want to implement an asynchronous version of `DropAllRows` for the interfaces that use exceptions. The signature of this function should look like so:

```C++
  future<void> AsyncDropAllRows(
};
```

- [ ] Implement the function.

manish-qlogic:
This change includes
1. Implement the function.
2. Implement an integration test for this function.
3. Implement an example for this function.
4. Execute the example as part of the CI build.

This closes #1501

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2484)


manish-qlogic:
@coryan 

I need your help as 2 checks are failing and both are not related to code change;
1.  "Kokoro Windows" : Log suggests BSOD.
2.  "Kokoro install ubuntu:Bionic" : logs are insufficient, it seems that build didn't start.

coryan:
For the record, both failures are known bugs in Kokoro.

Issue #2283, opened 2019-03-21
Refactor common protos out of bigtable_protos library.


coryan:
The bigtable_protos library (soon to be refactored out of `google/cloud/bigtable/CMakeLists.txt`) includes a number of common protos that would be shared with any future library. They should be refactored out to a smaller library, maybe even one per-proto.

coryan:
Argh, this was fixed by #2486.

Issue #2286, opened 2019-03-21
Fix bigtable_protos RPATH


remyabel:
This could very well be user error on my part, but I'm getting this when building an RPM:

```
*******************************************************************************
*
* WARNING: 'check-rpaths' detected a broken RPATH and will cause 'rpmbuild'
*          to fail. To ignore these errors, you can set the '$QA_RPATHS'
*          environment variable which is a bitmask allowing the values
*          below. The current value of QA_RPATHS is 0x0000.
*
*    0x0001 ... standard RPATHs (e.g. /usr/lib); such RPATHs are a minor
*               issue but are introducing redundant searchpaths without
*               providing a benefit. They can also cause errors in multilib
*               environments.
*    0x0002 ... invalid RPATHs; these are RPATHs which are neither absolute
*               nor relative filenames and can therefore be a SECURITY risk
*    0x0004 ... insecure RPATHs; these are relative RPATHs which are a
*               SECURITY risk
*    0x0008 ... the special '$ORIGIN' RPATHs are appearing after other
*               RPATHs; this is just a minor issue but usually unwanted
*    0x0010 ... the RPATH is empty; there is no reason for such RPATHs
*               and they cause unneeded work while loading libraries
*    0x0020 ... an RPATH references '..' of an absolute path; this will break
*               the functionality when the path before '..' is a symlink
*          
*
* Examples:
* - to ignore standard and empty RPATHs, execute 'rpmbuild' like
*   $ QA_RPATHS=$(( 0x0001|0x0010 )) rpmbuild my-package.src.rpm
* - to check existing files, set $RPM_BUILD_ROOT and execute check-rpaths like
*   $ RPM_BUILD_ROOT=<top-dir> /usr/lib/rpm/check-rpaths
*  
*******************************************************************************
ERROR   0002: file '/usr/lib64/libgoogleapis_cpp_bigtable_protos.so.0.1.0' contains an invalid rpath '/tmp/RhgfwdoDBe/BUILD/google-cloud-cpp-20190321gitfda670cd/external/lib' in [/tmp/RhgfwdoDBe/BUILD/google-cloud-cpp-20190321gitfda670cd/external/lib:/tmp/RhgfwdoDBe/BUILD/google-cloud-cpp-20190321gitfda670cd/external/lib64]
[...]

PR #2486, opened 2019-04-15
Support compiling spanner_tool with CMake.

### LINKED ISSUE #2451: import spanner protos
For the initial (non-GAPIC) Spanner work we need to import and compile the spanner protos from googleapis, similar to what we're already doing for bigtable.

coryan:
Compile the spanner protos in googleapis. I had to split the
bigtable_protos into a number of smaller libraries because there are
shared protos between bigtable and spanner.

This fixes #2451 and #2283.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2486)

### DIFF
--- a/cmake/external/googleapis.cmake
+++ b/cmake/external/googleapis.cmake
@@ -56,12 +56,13 @@ if (NOT TARGET googleapis_project)
 
     create_external_project_library_byproduct_list(
         googleapis_byproducts
-        "googleapis_cpp_longrunning_operations_protos"
         "googleapis_cpp_api_http_protos"
         "googleapis_cpp_api_annotations_protos"
         "googleapis_cpp_iam_v1_policy_protos"
         "googleapis_cpp_iam_v1_iam_policy_protos"
+        "googleapis_cpp_rpc_error_details_protos"
         "googleapis_cpp_rpc_status_protos"
+        "googleapis_cpp_longrunning_operations_protos"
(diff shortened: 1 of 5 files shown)


coryan:
Sorry for the large PR, the only excuse is that splitting the libraries was hard, and I had to test the final result to discover what to do.

cmake/external/googleapis.cmake
```diff
@@ -103,7 +113,6 @@ if (NOT TARGET googleapis_project)
                    -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
                    -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
                    -DCMAKE_INSTALL_RPATH=${GOOGLE_CLOUD_CPP_INSTALL_RPATH}
-                   -DCMAKE_SKIP_INSTALL_RPATH=ON
```
remyabel: Is there any particular reason why this is removed?

coryan: I cannot recall. Should we add it back?

remyabel: Yes. It was added because of https://github.com/googleapis/google-cloud-cpp/issues/2286. rpmbuild doesn't like an rpath of "external/lib" because it's "insecure", but it doesn't seem worth the effort to fix the RPATH unless somebody runs into a weird corner case. So the simplest solution is just to skip it.

PR #2488, opened 2019-04-16
The Google Cloud Storage C++ client library is GA.

coryan:
In this change:

- We update the version number for the Google Cloud Storage C++ client
  to 1.0.0.
- We mark the library as GA in the README file.
- We enable the API checks for the library, one of the CI builds will
  break if we break the API. Unfortunately the build also breaks with
  ABI changes, but in that case we can just update the baseline.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2488)


google/cloud/storage/doc/storage-main.dox
```diff
@@ -271,8 +271,8 @@ The library automatically retries requests that fail with transient errors, and
 follows the
 [recommended practice](https://cloud.google.com/storage/docs/exponential-backoff)
 to backoff between retries. Application developers can override the default
-[retry](@ref google::cloud::storage::v0::RetryPolicy) and
-[backoff](@ref google::cloud::storage::v0::BackoffPolicy) policies:
+[retry](@ref google::cloud::storage::v1::RetryPolicy) and
```
frankyn: Question: After this is merged, I'm assuming the client library API reference documentation will be updated from `v0` to `v1`. If so, please ignore and merge.

coryan: Yes, it will.

PR #2518, opened 2019-04-17
Refactor Table::*ReadModifyWriteRow().

coryan:
The new implementations do not depend on the code in `noex::Table`.

Part of the work for #2097.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2518)

### DIFF
--- a/google/cloud/bigtable/table.cc
+++ b/google/cloud/bigtable/table.cc
@@ -13,11 +13,13 @@
 // limitations under the License.
 
 #include "google/cloud/bigtable/table.h"
+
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/table.cc
```diff
@@ -13,11 +13,13 @@
 // limitations under the License.
 
 #include "google/cloud/bigtable/table.h"
+
```
devbww: I thought we weren't doing this (yet)?

coryan: Somehow I picked up a clang-format from the future. Fixed.

PR #2522, opened 2019-04-18
Check for (and fix) whitespace formatting in non-C++ files

devbww:
Add a phase in ci/check-style.sh that fixes whitespace formatting
in files not handled by clang-format.  For now this simply removes
trailing blanks.

Also build paths to ignore in style checks by reading .gitignore.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2522)


ci/check-style.sh
```diff
@@ -25,36 +25,39 @@ fi
...
 
-find google/cloud -name '*.h' -print0 \
-  | xargs -0 awk -f "${BINDIR}/check-include-guards.gawk"
+# Build paths to ignore in find(1) commands by reading .gitignore.
+declare -a ignore=( -path ./.git )
+if [ -f .gitignore ]; then
+  while read -r line; do
+    case "${line}" in
+    [^#]*/*) ignore+=( -o -path "./$(expr "${line}" : '\(.*\)/')" );;
+    [^#]*)   ignore+=( -o -name "${line}" );;
```
remyabel: Should this be `"./${line}"`?

ci/travis/build-docker.sh
```diff
@@ -65,7 +65,7 @@ fi
```
remyabel: Accidental permission change?

remyabel:
I am not 100% sure, but I think that for `find . \( "${ignore[@]}" \)`, the array will expand to something like ` -name bazel-* -o -name *~`. Notice the lack of quotes. We probably want to quote them to prevent shell globbing.

ci/check-style.sh
devbww: A slash in a `-name` argument will cause it to never match.  That is, `find -name` is only looking at the basename of the candidate path.

In the case of `bazel-*` from our `.gitignore`, we should probably be saying `bazel-*/'.

All that said, `find` can't implement all the `.gitignore` functionality (`**` is a simple example), so this mapping is certainly based on us only using simple "ignore" specifications.

remyabel:
Actually, disregard my last comment. It seems that `.gitignore` uses `*` for globbing, not regex, so leaving the expressions unquoted will be consistent.

ci/travis/build-docker.sh
devbww: Thanks.  Fixed.

devbww:
> I am not 100% sure, but I think that for find . \( "${ignore[@]}" \), the array will expand to something like -name bazel-* -o -name *~. Notice the lack of quotes. We probably want to quote them to prevent shell globbing.

I believe it will expand to the equivalent of (`"-name" "bazel-*" "-o" "-name" "*~"`), so there is no risk of those arguments being re-expanded.  That is, they are quoted.  For example, ...

```
$ echo C*
CMakeLists.txt CODE_OF_CONDUCT.md CONTRIBUTING.md
$ declare -a A=( 'C*' )
$ echo "${A[@]}"
C*
$ echo ${A[@]}
CMakeLists.txt CODE_OF_CONDUCT.md CONTRIBUTING.md
```

coryan:
FWIW, the UBSAN build seems to have finished, but Travis may have lost the notification.  Feel free to merge if you want.

devbww:
I'll wait to see if @remyabel has anything to add.

remyabel:
Nope it LGTM. Sorry for the mistaken comments.

PR #2526, opened 2019-04-19
Restore code coverage uploads.

coryan:
ETOOMANYQUOTES.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2526)

### DIFF
--- a/ci/travis/upload-codecov.sh
+++ b/ci/travis/upload-codecov.sh
@@ -22,7 +22,7 @@ fi
 source "${PROJECT_ROOT}/ci/travis/linux-config.sh"
 source "${PROJECT_ROOT}/ci/define-dump-log.sh"
 
-if [ "${BUILD_TYPE:-Release}" != "Coverage" ]; then
+if [[ "${BUILD_TYPE:-Release}" != "Coverage" ]]; then
     echo "Skipping code coverage as it is disabled for this build."
     exit 0
 fi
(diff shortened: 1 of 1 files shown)


ci/travis/upload-codecov.sh
```diff
@@ -22,16 +22,19 @@ fi
...
 # Upload the results using the script from codecov.io
 # Save the log to a file because it exceeds the 4MB limit in Travis.
 echo -n "Uploading code coverage to codecov.io..."
-readonly CI_ENV=$(bash <(curl -s https://codecov.io/env))
-sudo docker run "${CI_ENV}" \
+readonly -a CI_ENV=( $(bash <(curl -s https://codecov.io/env)) )
+# We use parameter expansion for ${CI_ENV} because set -u doesn't like empty
+# arrays on older versions of Bash (which some of our builds use). The
+# expression ${parameter+word} will expand word only if parameter is not unset.
+sudo docker run "${CI_ENV[@]+"${CI_ENV[@]}"}" \
```
remyabel: I think that we'll need to remember everytime to write this comment/syntax and that can be error-prone. What do you think about changing this to something like `${TEST[@]:-}` (note: lack of quotes). It is less typing and we don't need a comment everywhere describing what it means.

remyabel: Actually, I retract my statement. The disadvantage of `"${TEST[@]:-}"` is that it expands to an array with one single empty value (instead of an empty array).

PR #2529, opened 2019-04-22
Support move-only callables in future<T>.

coryan:
I noticed this while trying to use `future<T>` in a larger change,
separating the change to its own bug seems to make sense.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2529)

### DIFF
--- a/google/cloud/future_generic_then_test.cc
+++ b/google/cloud/future_generic_then_test.cc
@@ -111,6 +111,70 @@ TEST(FutureTestInt, ThenUnwrap) {
   EXPECT_FALSE(next.valid());
 }
 
+TEST(FutureTestInt, ThenMoveOnlyCallable) {
+  class MoveOnlyCallable {
+   public:
+    explicit MoveOnlyCallable(bool& called) : called_(&called) {}
+
+    MoveOnlyCallable(MoveOnlyCallable&&) = default;
+    MoveOnlyCallable& operator=(MoveOnlyCallable&&) = default;
+
+    MoveOnlyCallable(MoveOnlyCallable const&) = delete;
+    MoveOnlyCallable& operator=(MoveOnlyCallable const&) = delete;
+
+    int operator()(future<int> f) {
+      *called_ = true;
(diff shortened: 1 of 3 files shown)


google/cloud/future_generic_then_test.cc
```diff
@@ -111,6 +111,70 @@ TEST(FutureTestInt, ThenUnwrap) {
...
+
+    MoveOnlyCallable(MoveOnlyCallable&&) = default;
+    MoveOnlyCallable& operator=(MoveOnlyCallable&&) = default;
+
+    MoveOnlyCallable(MoveOnlyCallable const&) = delete;
+    MoveOnlyCallable& operator=(MoveOnlyCallable const&) = delete;
+
+    int operator()(future<int> f) {
+      *called_ = true;
+      return 2 * f.get();
```
devbww: Opt: I might put some variant of "Doubler" in the class name.

google/cloud/future_generic_then_test.cc
```diff
@@ -111,6 +111,70 @@ TEST(FutureTestInt, ThenUnwrap) {
...
+  EXPECT_FALSE(fut.valid());
+  EXPECT_TRUE(next.valid());
+  EXPECT_FALSE(called);
+
+  p.set_value(42);
+  EXPECT_TRUE(called);
+  EXPECT_TRUE(next.valid());
+  EXPECT_EQ(std::future_status::ready, next.wait_for(0_ms));
+
+  EXPECT_EQ(84, next.get());
```
devbww: Opt: Is `2 * 42` clearer?

google/cloud/future_generic_then_test.cc
```diff
@@ -111,6 +111,70 @@ TEST(FutureTestInt, ThenUnwrap) {
...
```
devbww: Opt: So perhaps `auto doubler = ...` and `2 * 42` here too then?

PR #2530, opened 2019-04-23
Adding `AsyncGenerateConsistencyToken` and `AsyncCheckConsistency` to `TableAdmin`.

### LINKED ISSUE #1499: Implement bigtable::TableAdmin::AsyncGenerateConsistencyToken
We want to implement an asynchronous version of `GenerateConsistencyToken` for the interfaces that use exceptions. The signature of this function should look like so:

```C++
  future<std::string> AsyncGenerateConsistencyToken(
      std::string const& table_id,
      CompletionQueue& cq);
};
```

- [ ] Implement the function.

manish-qlogic:
This change includes following
1. Implement the function `TableAdmin::AsyncGenerateConsistencyToken` and `TableAdmin::AsyncCheckConsistency`.
2. Implement integration tests for these functions.
3. Implement examples for these functions.
4. Execute the examples as part of the CI build.

This closes #1499

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2530)

### DIFF
--- a/google/cloud/bigtable/examples/run_examples_utils.sh
+++ b/google/cloud/bigtable/examples/run_examples_utils.sh
@@ -293,6 +293,13 @@ function run_all_table_admin_async_examples {
       "${project_id}" "${INSTANCE}" "${TABLE}"
   run_example ./table_admin_async_snippets async-modify-table \
       "${project_id}" "${INSTANCE}" "${TABLE}"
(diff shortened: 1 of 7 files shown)


manish-qlogic:
@coryan ,
Logs are not available for the failed builds, please help

PR #2539, opened 2019-04-25
Fix noex build and improve build logs.

coryan:
We cannot install `clang-format-7` on all versions of Ubuntu. We already install it on the
section of the Dockerfile for Ubuntu:18.04, so remove it from the generic section, and things
are happy.

I also changed the CI build scripts to actually print any errors while creating the docker images.
I could not figure out why the build was broken until I did that.

----

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2539)

### DIFF
--- a/ci/install-retry.sh
+++ b/ci/install-retry.sh
@@ -36,8 +36,8 @@ min_wait=120
 # Do not exit on failures for this loop.
 set +e
 for i in 1 2 3; do
-  ${PROGRAM} "$@"
-  if [ $? -eq 0 ]; then
+  "${PROGRAM}" "$@"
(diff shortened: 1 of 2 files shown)


ci/install-retry.sh
```diff
@@ -36,8 +36,8 @@ min_wait=120
 # Do not exit on failures for this loop.
 set +e
 for i in 1 2 3; do
-  ${PROGRAM} "$@"
-  if [ $? -eq 0 ]; then
+  "${PROGRAM}" "$@"
```
devbww: Just a suggestion: Perhaps `"${PROGRAM}" "$@" && exit 0`.

PR #2548, opened 2019-04-26
Fix race condition in async_read_stream_test.

coryan:
The test for "read 3 messages and then an error" was a bit optimistic:
sometimes all three messages and the error are coalesced and then the
client just sees the error. Add some synchronization to make sure this
does not happen.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2548)

### DIFF
--- a/google/cloud/bigtable/async_read_stream_test.cc
+++ b/google/cloud/bigtable/async_read_stream_test.cc
@@ -304,14 +304,24 @@ TEST_F(AsyncReadStreamTest, Return3) {
 
 /// @test Verify that the AsyncReadStream detect errors reported by the server.
 TEST_F(AsyncReadStreamTest, Return3ThenFail) {
+  // Very rarely (in the CI builds, with high load), all 3 responses and the
+  // error message are coalesced into a single message from the server, and then
+  // the OnRead() calls do not happen. We need to explicitly synchronize the
+  // client and server threads t
+  SimpleBarrier server_barrier;
   impl_.SetCallback(
-      [this](grpc::ServerContext*,
-             google::bigtable::v2::MutateRowsRequest const*,
-             grpc::ServerWriter<google::bigtable::v2::MutateRowsResponse>*
-                 writer) {
+      [this, &server_barrier](
+          grpc::ServerContext*, google::bigtable::v2::MutateRowsRequest const*,
+          grpc::ServerWriter<google::bigtable::v2::MutateRowsResponse>*
+              writer) {
         WriteOne(writer, 0);
         WriteOne(writer, 1);
-        WriteLast(writer, 2);
+        // Cannot use WritLast() because that blocks until the status is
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/async_read_stream_test.cc
```diff
@@ -304,14 +304,24 @@ TEST_F(AsyncReadStreamTest, Return3) {
 
 /// @test Verify that the AsyncReadStream detect errors reported by the server.
 TEST_F(AsyncReadStreamTest, Return3ThenFail) {
+  // Very rarely (in the CI builds, with high load), all 3 responses and the
+  // error message are coalesced into a single message from the server, and then
+  // the OnRead() calls do not happen. We need to explicitly synchronize the
+  // client and server threads t
```
devbww: s/ t$/./

google/cloud/bigtable/async_read_stream_test.cc
```diff
@@ -304,14 +304,24 @@ TEST_F(AsyncReadStreamTest, Return3) {
...
+          grpc::ServerWriter<google::bigtable::v2::MutateRowsResponse>*
+              writer) {
         WriteOne(writer, 0);
         WriteOne(writer, 1);
-        WriteLast(writer, 2);
+        // Cannot use WritLast() because that blocks until the status is
+        // returned, and we want to pause in `server_barrier` to ensure all
+        // messages are received.
+        WriteOne(writer, 2);
+        // block until the client has received the responses.
```
devbww: s/block/Block/

google/cloud/bigtable/async_read_stream_test.cc
```diff
@@ -304,14 +304,24 @@ TEST_F(AsyncReadStreamTest, Return3) {
...
-             grpc::ServerWriter<google::bigtable::v2::MutateRowsResponse>*
-                 writer) {
+      [this, &server_barrier](
+          grpc::ServerContext*, google::bigtable::v2::MutateRowsRequest const*,
+          grpc::ServerWriter<google::bigtable::v2::MutateRowsResponse>*
+              writer) {
         WriteOne(writer, 0);
         WriteOne(writer, 1);
-        WriteLast(writer, 2);
+        // Cannot use WritLast() because that blocks until the status is
```
devbww: s/WritLast/WriteLast/

coryan: Fixed.

google/cloud/bigtable/async_read_stream_test.cc
coryan: Fixed.

google/cloud/bigtable/async_read_stream_test.cc
coryan: Fixed.

PR #2549, opened 2019-04-26
Fix exit status in examples runner.

coryan:
Sometimes we were failing to report the exit status, and even when we
did it was not obvious there was a problem, because it may have scrolled
up past the terminal size.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2549)

### DIFF
--- a/ci/define-example-runner.sh
+++ b/ci/define-example-runner.sh
@@ -191,3 +191,28 @@ run_example_usage() {
   set -e
   /bin/rm -f "${log}"
 }
+
+################################################
+# Report the result of running many examples.
+# Globals:
+#   COLOR_*: colorize output messages, defined in colors.sh
+#   EXIT_STATUS: control the final exit status for the program.
+#   EMULATOR_LOG: the name of the emulator logfile.
+# Arguments:
+#   None
+# Returns:
+#   None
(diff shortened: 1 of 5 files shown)


ci/define-example-runner.sh
```diff
@@ -191,3 +191,28 @@ run_example_usage() {
...
+    echo "${COLOR_GREEN}All the examples finished successfully.${COLOR_RESET}"
+    echo
+  else
+    echo
+    echo "${COLOR_RED}Some of the examples failed.${COLOR_RESET}"
+    echo
+  fi
+
+  exit "${EXIT_STATUS}"
+}
```
mr-salty: I think you can do I think "trap exit_example_runner EXIT" here to have it auto-execute on exit?

coryan: Unfortunately there is already a trap present when using the testbench or the emulator.  And traps do not stack.

mr-salty: ok... you can use 'trap -p' to get the current trap and chain them together, but of course that assumes everyone cooperates or it's subject to ordering...  so let's leave this for now.

PR #2552, opened 2019-04-26
Implement InstanceAdmin::AsyncGetAppProfile.

### LINKED ISSUE #1531: Implement bigtable::InstanceAdmin::AsyncCreateAppProfile
We want to implement an asynchronous version of `CreateAppProfile` for the interfaces that use exceptions. The signature of this function should look like so:

```C++
class google::cloud::bigtable::InstanceAdmin { public:
  future<google::bigtable::admin::v2::AppProfile> CreateAppProfile(
      bigtable::InstanceId const& instance_id, AppProfileConfig config,
      CompletionQueue& cq);
};
```

- [ ] Implement the function.

dopiera:
This fixes #1531.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2552)

### DIFF
--- a/google/cloud/bigtable/examples/instance_admin_async_snippets.cc
+++ b/google/cloud/bigtable/examples/instance_admin_async_snippets.cc
@@ -277,28 +277,28 @@ void AsyncGetAppProfile(cbt::InstanceAdmin instance_admin,
   }
 
   //! [async get app profile]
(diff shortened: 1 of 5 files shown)


dopiera:
The Windows build failure seems unrelated.

PR #2565, opened 2019-04-27
Implement MutationBatcher without using noex::Table.

### LINKED ISSUE #2112: Create `make_ready_future<>`
There are a couple of places in the code where we create futures which are satisfied since creation. 

Currently, in order to do so, you need to create a promise, satisfy it and finally extract a future from it. `make_ready_future<>` would allow you to avoid that.

@coryan already implemented it, but it needs to be tested:

https://github.com/coryan/google-cloud-cpp/blob/63bbc6667bacc887a949a0ef2e731391b7416959/google/cloud/future_generic.h#L128

https://github.com/coryan/google-cloud-cpp/blob/63bbc6667bacc887a949a0ef2e731391b7416959/google/cloud/future_void.h#L125

### LINKED ISSUE #2562: Move BulkMutator to use new Table::AsyncBulkApply
What it says on the tin. We are removing dependencies on `noex::*` everywhere.

coryan:
This version intentionally drops some functionality, specifically the
previous version was able to handle partial results from a BulkApply()
to admit more mutations early. Unfortunately the server does not send
partial results yet, and the code is more difficult to maintain.

This fixes #2562, and incidentally fixes #2112.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2565)

### DIFF
--- a/google/cloud/bigtable/mutation_batcher.cc
+++ b/google/cloud/bigtable/mutation_batcher.cc
@@ -172,11 +172,12 @@ void MutationBatcher::OnBulkApplyDone(CompletionQueue cq,
       data.done = true;
     }
   }
+  auto const num_mutations = batch.mutation_data.size();
   batch.mutation_data.clear();
 
   std::unique_lock<std::mutex> lk(mu_);
   outstanding_size_ -= batch.requests_size;
-  num_requests_pending_--;
+  num_requests_pending_ -= num_mutations;
   num_outstanding_batches_--;
   SatisfyPromises(TryAdmit(cq), lk);  // unlocks the lock
 }
--- a/google/cloud/bigtable/mutation_batcher.cc
+++ b/google/cloud/bigtable/mutation_batcher.cc
@@ -159,8 +159,6 @@ void MutationBatcher::OnBulkApplyDone(CompletionQueue& cq,
     // Release resources as early as possible.
     batch.mutation_data.erase(it);
   }
-  failed.clear();
-  failed.shrink_to_fit();
   // Any remaining mutations were successful then.
   for (auto& kv : batch.mutation_data) {
     MutationData& data = kv.second;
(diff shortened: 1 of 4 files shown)


coryan:
I understand the argument, but I think we should ignore it for now, two reasons:

1. We will not be changing the API when we add that functionality, and I would rather get the right APIs out ASAP.
2. I do not think the performance impact would be that bad, or rather, the improvement would not be that big: if we have one hot tablet server, then on average, each batch of a list L (no matter how we split L) will contain one element on the hot tablet server.  The overall time to commit the list is the same, on average.

dopiera:
> I understand the argument, but I think we should ignore it for now, two reasons:
> 
> 1. We will not be changing the API when we add that functionality, and I would rather get the right APIs out ASAP.
> 2. I do not think the performance impact would be that bad, or rather, the improvement would not be that big: if we have one hot tablet server, then on average, each batch of a list L (no matter how we split L) will contain one element on the hot tablet server.  The overall time to commit the list is the same, on average.

I'm afraid you're right about the improvement this brings. I had a nasty plan in my head to consider a batch done as soon as e.g. 90% of requests in it were - that would have made it better provided that the service was actually streaming results.

In short, I'm convinced that there is little reason to keep this functionality.

That said, the code can be made simpler by also reverting all the prep work which I did for streaming:
#2055, #2058 and #2063.

google/cloud/bigtable/mutation_batcher.cc
```diff
@@ -131,92 +128,50 @@ bool MutationBatcher::HasSpaceFor(PendingSingleRowMutation const& mut) const {
...
+    if (it == batch.mutation_data.end()) {
+      continue;
+    }
+    MutationData& data = it->second;
+    data.completion_promise.set_value(f.status());
     // Release resources as early as possible.
     batch.mutation_data.erase(it);
   }
-  // TODO(#2093): remove once `FailedMutations` are small.
   failed.clear();
```
dopiera: You can actually remove them now :)

PR #2568, opened 2019-04-27
Preserve timestamps for unchanged files.

coryan:
Adding `IncludeBlocks: Merge` apparently triggers a bug in
clang-format, where it changes the timestamp of the file even if
there are no formatting changes. I think we want that directive, so I
changed the check-style.sh script to workaround the issue.

While testing this I discovered that cmake-format has the same problem,
so I worked around that bug too.

Finally, I made some cosmetic changes to make my IDE happy inside this
script.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2568)

### DIFF
--- a/ci/check-style.sh
+++ b/ci/check-style.sh
@@ -39,6 +39,22 @@ fi
 find google/cloud -name '*.h' -print0 |
   xargs -0 awk -f "${BINDIR}/check-include-guards.gawk"
 
+replace_original_if_changed() {
+  if [[ $# != 1 ]]; then
+    return 1
+  fi
+
+  local file="$1"
+  shift
+
+  chmod --reference="${file}" "${file}.tmp"
+  if cmp -s "${file}" "${file}.tmp"; then
+    rm -f "${file}.tmp"
+  else
+    mv -f "${file}.tmp" "${file}"
+  fi
+}
+
 # Apply cmake_format to all the CMake list files.
 #     https://github.com/cheshirekow/cmake_format
 find . \( "${ignore[@]}" \) -prune -o \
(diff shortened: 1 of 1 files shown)


coryan:
@mr-salty you said LGTM, but looks like you accidentally forgot to click the approve button.

devbww:
It looks like it might have already been fixed.

```
$ ls -l sign_blob_requests.cc
-rw-r--r-- 1 bww eng 2218 Apr  8 23:33 sign_blob_requests.cc
$ clang-format -i sign_blob_requests.cc
$ ls -l sign_blob_requests.cc
-rw-r--r-- 1 bww eng 2218 Apr  8 23:33 sign_blob_requests.cc
$ clang-format-7 -i sign_blob_requests.cc
$ ls -l sign_blob_requests.cc
-rw-r--r-- 1 bww eng 2218 Apr 27 23:27 sign_blob_requests.cc
$ clang-format -version
clang-format version google3-trunk (trunk r358772)
$ clang-format-7 -version  
clang-format version 7.0.1-6 (tags/RELEASE_701/final)
```

And given that we didn't notice it before, perhaps it was a regression in version 7.

ci/check-style.sh
```diff
@@ -27,30 +27,52 @@ readonly BINDIR="$(dirname "$0")"
...
+
 # Apply cmake_format to all the CMake list files.
 #     https://github.com/cheshirekow/cmake_format
 find . \( "${ignore[@]}" \) -prune -o \
        \( -name 'CMakeLists.txt' -o -name '*.cmake' \) \
        -print0 |
-  xargs -0 cmake-format -i
+  while IFS= read -r -d $'\0' file; do
+    cmake-format "${file}" >"${file}.tmp"
+    replace_original_if_changed "${file}"
```
devbww: It's a little confusing that `"${file}.tmp"` isn't mentioned here.  I think I'd just make the function take two names (i.e., not assume `.tmp`).

coryan:
PTAL.

ci/check-style.sh
```diff
@@ -40,18 +40,18 @@ find google/cloud -name '*.h' -print0 |
...
 
-  local file="$1"
-  shift
+  local original="$1"
+  local reformatted="$2"
 
-  chmod --reference="${file}" "${file}.tmp"
-  if cmp -s "${file}" "${file}.tmp"; then
-    rm -f "${file}.tmp"
+  chmod --reference="${original}" "${reformatted}"
```
devbww: I'm not sure why I didn't originally put the "chmod" in the "mv" block, but we can do just that if we get back here.

PR #2571, opened 2019-04-28
Implement InstanceAdmin::AsyncSetIamPolicy.

### LINKED ISSUE #1537: Implement bigtable::InstanceAdmin::AsyncSetIamPolicy
We want to implement an asynchronous version of `SetIamPolicy` for the interfaces that use exceptions. The signature of this function should look like so:

```C++
class google::cloud::bigtable::InstanceAdmin { public:
  future<google::cloud::IamPolicy> AsyncSetIamPolicy(
      std::string const& instance_id,
      google::cloud::IamBindings const& iam_bindings,
      std::string const& etag,
      CompletionQueue& cq);
  future<google::cloud::IamPolicy> AsyncSetIamPolicy(
      std::string const& instance_id,
      google::cloud::IamBindings const& iam_bindings,
      CompletionQueue& cq);
};
```

- [ ] Implement the function.

dopiera:
This fixes #1537.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2571)

### DIFF
--- a/google/cloud/bigtable/examples/instance_admin_async_snippets.cc
+++ b/google/cloud/bigtable/examples/instance_admin_async_snippets.cc
@@ -705,6 +705,69 @@ void AsyncDeleteAppProfile(
   (std::move(instance_admin), std::move(cq), argv[1], argv[2]);
 }
 
+void AsyncSetIamPolicy(google::cloud::bigtable::InstanceAdmin instance_admin,
+                       google::cloud::bigtable::CompletionQueue cq,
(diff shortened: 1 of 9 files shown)


coryan:
@dopiera you need to rebase and resolve the conflicts.

dopiera:
> Not for this PR, but we should rethink these loops, they make the examples really long and hard to read. Do you mind creating a bug in the GA milestone so we do not forget?

Done

PR #2572, opened 2019-04-28
Remove Ubuntu trusty build/references.

### LINKED ISSUE #2408: Ubuntu 14.04 will be EOL soon
This bug is to track that Ubuntu 14.04 will be EOL [April 17th](https://askubuntu.com/questions/1126909/when-is-the-exact-date-for-eol-of-ubuntu-14-04-lts/1126933#1126933).

remyabel:
Fixes #2408 

Technically, it stops receiving updates April 30th. Opening this early because it is easy to review. I don't know if GCE plans on removing the Ubuntu 14.04 image. If it doesn't, maybe it will be worth it to close this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2572)


coryan:
I think this is going to have to wait until we have some telemetry of usage over Ubuntu:14.04.  There are no immediate plans to remove the images from Google Compute Engine, and without telemetry we should assume that customers are still using it.

I want to remove support for this platform ASAP, but we need some more data before proceeding.

PR #2573, opened 2019-04-29
Implemented `TableAdmin::AsyncWaitForConsistency`

### LINKED ISSUE #1500: Implement bigtable::TableAdmin::AsyncWaitForConsistency
We want to implement an asynchronous version of `WaitForConsistencyCheck` for the interfaces that use exceptions. The signature of this function should look like so:

```C++
  future<bool> AsyncAwaitConsistency(
};
```

Note that this function will need to query the completion token and then poll until the operation is successful (or the polling policy expires).  This is probably blocked by #1499.

- [ ] Implement the function.

manish-qlogic:
Implemented `TableAdmin::AsyncWaitForConsistency`.

This PR closes #1500

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2573)


manish-qlogic:
Closing this PR as this implementation is not purely "async" and it behave synchronously in do while loop.

PR #2574, opened 2019-04-29
Return google::cloud::future from InstanceAdmin functions.

### LINKED ISSUE #2564: Return google::cloud::future from InstanceAdmin functions.
A few TableAdmin functions return a `std::future<>`, we should return a `google::cloud::future<>` for consistency with the Async APIs.

coryan:
A few functions were returning `std::future<>` because they were implemented
before we had `google::cloud::future<>`, changed the return type to
be consistent with the `Async*()` variants.

Changed the implementation to use `Async*()`, that seems simpler than having
an implementation using blocking calls in a separate thread.

Removed a bunch of redudant unit tests once the implementation
is refactored. We already have unit and integration tests for
the `Async*()` versions of these functions.

This fixes #2564.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2574)


google/cloud/bigtable/instance_admin.cc
```diff
@@ -124,10 +124,16 @@ future<StatusOr<InstanceList>> InstanceAdmin::AsyncListInstances(
       });
 }
 
-std::future<StatusOr<google::bigtable::admin::v2::Instance>>
-InstanceAdmin::CreateInstance(InstanceConfig instance_config) {
-  return std::async(std::launch::async, &InstanceAdmin::CreateInstanceImpl,
-                    this, std::move(instance_config));
+future<StatusOr<btadmin::Instance>> InstanceAdmin::CreateInstance(
+    InstanceConfig instance_config) {
+  CompletionQueue cq;
```
devbww: Aside: The `shared_ptr`-ness of `CompletionQueue` is too hidden methinks.

coryan:
Rebased to resolve conflicts.

PR #2578, opened 2019-04-29
Prevent transient problems from breaking production test.

coryan:
One of the test runs against production failed in this test:

https://source.cloud.google.com/results/invocations/7e17585e-596f-40c3-b237-e4745b54dfc1

The log is not detailed enough to diagnose the problem, so first I
fixed that. My guess is that the test failed because the mutations
used to prepare data for the test are non-idempotent, so any transient
failure will stop the test. I fixed that too.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2578)

### DIFF
--- a/google/cloud/bigtable/tests/data_integration_test.cc
+++ b/google/cloud/bigtable/tests/data_integration_test.cc
@@ -518,14 +518,23 @@ TEST_F(DataIntegrationTest, TableSampleRowKeysTest) {
       for (int col = 0; col != COLUMN_COUNT; ++col) {
         std::string colid = "c" + std::to_string(col);
         std::string value = colid + "#" + os.str();
-        mutation.emplace_back(
-            bigtable::SetCell(family1, std::move(colid), std::move(value)));
+        mutation.emplace_back(bigtable::SetCell(family1, std::move(colid),
+                                                std::chrono::milliseconds(0),
+                                                std::move(value)));
       }
       bulk.emplace_back(std::move(mutation));
       ++rowid;
     }
     auto failures = table.BulkApply(std::move(bulk));
-    ASSERT_TRUE(failures.empty());
+    ASSERT_TRUE(failures.empty()) << "failures=" << [&failures]() {
+      std::ostringstream os;
+      os << "[";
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/tests/data_integration_test.cc
```diff
@@ -518,14 +518,23 @@ TEST_F(DataIntegrationTest, TableSampleRowKeysTest) {
...
       bulk.emplace_back(std::move(mutation));
       ++rowid;
     }
     auto failures = table.BulkApply(std::move(bulk));
-    ASSERT_TRUE(failures.empty());
+    ASSERT_TRUE(failures.empty()) << "failures=" << [&failures]() {
+      std::ostringstream os;
+      os << "[";
+      for (auto&& f : failures) {
+        os << "failed[" << f.original_index() << "]=" << f.status() << ", ";
```
devbww: Do you care about the trailing `", "`?

google/cloud/bigtable/tests/data_integration_test.cc
```diff
@@ -518,14 +518,23 @@ TEST_F(DataIntegrationTest, TableSampleRowKeysTest) {
...
       }
       bulk.emplace_back(std::move(mutation));
       ++rowid;
     }
     auto failures = table.BulkApply(std::move(bulk));
-    ASSERT_TRUE(failures.empty());
+    ASSERT_TRUE(failures.empty()) << "failures=" << [&failures]() {
+      std::ostringstream os;
+      os << "[";
+      for (auto&& f : failures) {
```
devbww: What does the `&&` buy you here?

google/cloud/bigtable/tests/data_integration_test.cc
coryan: Don't care, but it is easy to fix, done.

google/cloud/bigtable/tests/data_integration_test.cc
coryan: Lazyness.  Fixed.

PR #2579, opened 2019-04-29
Implement Table::AsyncReadRows.

### LINKED ISSUE #1487: Implement bigtable::Table::AsyncReadRows
This function will implement an asynchronous version of `Table::ReadRows`. The function will have the following signature:

```C++
future<void> AsyncReadRows(
    RowSet row_set, Filter filter, CompletionQueue& cq, ReadRowFunctor&&);
future<void> AsyncReadRows(
    RowSet row_set, std::int64_t rows_limit, Filter filter, CompletionQueue& cq, ReadRowFunctor&&);
};
```

The `ReadRowFunctor` parameter will be called for each row received. The future becomes satisfied when all the rows have been read.

- [ ] Implement the function.

dopiera:
This fixes #1487.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2579)

### DIFF
--- a/google/cloud/bigtable/CMakeLists.txt
+++ b/google/cloud/bigtable/CMakeLists.txt
@@ -101,6 +101,8 @@ add_library(bigtable_client
             app_profile_config.h
             app_profile_config.cc
             async_operation.h
+            async_row_reader.h
+            async_row_reader.cc
             bigtable_strong_types.h
             ${CMAKE_CURRENT_BINARY_DIR}/version_info.h
             cell.h
@@ -256,6 +258,7 @@ if (BUILD_TESTING)
         async_list_clusters_test.cc
         async_list_instances_test.cc
         async_read_stream_test.cc
+        async_row_reader_test.cc
         bigtable_version_test.cc
         cell_test.cc
         client_options_test.cc
--- a/google/cloud/bigtable/async_row_reader.cc
+++ b/google/cloud/bigtable/async_row_reader.cc
@@ -40,9 +40,19 @@ AsyncRowReader::AsyncRowReader(
       rows_count_(0) {}
 
 AsyncRowReader::~AsyncRowReader() {
-  assert(whole_op_finished_);
+  // At this point no one should be doing anything to the object, but acquire
+  // the mutex to make sure that modifications made in other threads are visible
+  // in this one.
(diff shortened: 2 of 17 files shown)


coryan:
I think we should implement this will callbacks for now. This is too easy to use wrong, and too many layers of type wrapping.

coryan:
Thanks, I will review when I get back to work tomorrow.

To be clear: I think we will want something *like* futures for asynchronous streaming reads, it is just that we are not ready for them yet.  What I am thinking about is something like RxCpp (which I believe you introduced me to), so application would be able to say things like:

```C++
Stream<Row> rows = table.AsyncReadRows(...);
// Call the lambda for each result and return a new stream, note that this is done asynchronously ...
Stream<Bar> bars = rows | transform([](Row r) -> Bar { ... });
// Aggregate the Bars into some kind of total
future<Total> final = bars | accumulate(Total{}, [](Total acc, Bar b) -> Total { ... });
final.then([](Total t) { std::cout << t << std::endl; });
```

But I do not think we have the time or the enough understanding of this stuff to design it correctly, so I am going with callbacks for now.

Somewhat unrelated, are you sure about that parallel version? "[Read] ..  should not be called concurrently with other streaming APIs on the same stream. It is not meaningful to call it concurrently with another AsyncReaderInterface::Read on the same stream since reads on the same stream are delivered in order."

https://grpc.github.io/grpc/cpp/classgrpc_1_1internal_1_1_async_reader_interface.html#adf87c602036d69158950b7299d0aae70

dopiera:
> Somewhat unrelated, are you sure about that parallel version? "[Read] .. should not be called concurrently with other streaming APIs on the same stream. It is not meaningful to call it concurrently with another AsyncReaderInterface::Read on the same stream since reads on the same stream are delivered in order."
> 
This is only parallel row processing - not reading. The first two examples instruct gRPC to read more data only after the `future<>` returned by `Process()` is satisfied. The parallel example allows for `N` simultaneous `Process()` to run, but gRPC still gets only one `AsyncReaderInterface::Read()` at a time - this is implemented in `AsyncRowReader`.

coryan:
Should we close this PR?

coryan:
Closing in favor of #2701.

PR #2605, opened 2019-05-04
Cleanup the Doxygen documentation.

coryan:
I made a pass through the Doxygen documentation for Cloud Bigtable,
making the following changes:

- I documented the `google::cloud::bigtable` namespace (oops).
- I removed any symbol in the `internal` namespace from the generated
  documentation. Made the same change for the Cloud Storage and common
  libraries.
- I moved a few symbols into the `google::cloud::bigtable::internal`
  namespace, as they were not intended for general use.
- I removed a forward declaration for AsyncRetryAndPollUnaryRpc, a
  symbol that does not exist, but the forward declaration made it appear
  in the docs.
- I filled in some missing documentation, and fixed the formatting and
  wording for some Doxygen comments.
- I fixed some code samples that rendered poorly in the HTML output.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2605)

### DIFF
--- a/google/cloud/bigtable/examples/table_admin_async_snippets.cc
+++ b/google/cloud/bigtable/examples/table_admin_async_snippets.cc
@@ -289,30 +289,31 @@ void AsyncCheckConsistency(google::cloud::bigtable::TableAdmin admin,
 
   //! [async check consistency]
   namespace cbt = google::cloud::bigtable;
+  using google::cloud::future;
+  using google::cloud::StatusOr;
   [](cbt::TableAdmin admin, cbt::CompletionQueue cq, std::string table_id_param,
      std::string consistency_token_param) {
     cbt::TableId table_id(table_id_param);
     cbt::ConsistencyToken consistency_token(consistency_token_param);
 
-    google::cloud::future<google::cloud::StatusOr<cbt::Consistency>> future =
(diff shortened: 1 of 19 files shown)


google/cloud/bigtable/examples/table_admin_async_snippets.cc
```diff
@@ -289,30 +289,31 @@ void AsyncCheckConsistency(google::cloud::bigtable::TableAdmin admin,
...
+              auto consistency = f.get();
+              if (!consistency) {
+                throw std::runtime_error(consistency.status().message());
+              }
+              if (consistency.value() == cbt::Consistency::kConsistent) {
+                std::cout << "Table is consistent\n";
+              } else {
+                std::cout << "Table is not yet consistent, Please try again"
+                          << " later with the same token ("
+                          << consistency_token_param << ")";
```
devbww: What's the idea here behind no `"\n"` but a `flush`?

coryan: No idea. I think I only "touched" that code because reformatting.  But I removed the explicit flush (here and elsewhere in the examples) and just used `\n` as usual.

PR #2615, opened 2019-05-06
Create aliases for table schema view constants.

### LINKED ISSUE #2599: Create aliases for ::google::bigtable::admin::v2::Table::SCHEMA_VIEW.
We have aliases for other magic constants, but not for `::google::bigtable::admin::v2::Table::SCHEMA_VIEW` and `::google::bigtable::admin::v2::Table:FULL`, those would be both convenient and consistent to have.

coryan:
Define aliases for the google.bigtable.admin.v2.Table.View constants.
We do this for other constants and it seems more consistent to have them
for these too.

In some of the examples we were using `VIEW_UNSPECIFIED`, but only
needed `NAME_ONLY`. Change the examples to use the more accurate
table view.

This fixes #2599.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2615)

### DIFF
--- a/google/cloud/bigtable/examples/bigtable_samples.cc
+++ b/google/cloud/bigtable/examples/bigtable_samples.cc
@@ -40,7 +40,7 @@ void RunTableOperations(google::cloud::bigtable::TableAdmin admin,
 
   std::cout << "Listing tables:\n";
   auto tables =
-      admin.ListTables(google::bigtable::admin::v2::Table::VIEW_UNSPECIFIED);
+      admin.ListTables(google::cloud::bigtable::TableAdmin::NAME_ONLY);
 
   if (!tables) {
     throw std::runtime_error(tables.status().message());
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/examples/bigtable_samples.cc
```diff
@@ -40,7 +40,7 @@ void RunTableOperations(google::cloud::bigtable::TableAdmin admin,
 
   std::cout << "Listing tables:\n";
   auto tables =
-      admin.ListTables(google::bigtable::admin::v2::Table::VIEW_UNSPECIFIED);
+      admin.ListTables(google::cloud::bigtable::TableAdmin::NAME_ONLY);
```
devbww: Mention the value changes in the PR description?

coryan: Done.

Issue #2604, opened 2019-05-04
Use explicit return types in code samples.


coryan:
Though using `auto result = ...` is perfectly idiomatic, the purpose of the examples is to show how the functions are used, we should be explicit in them, e.g.:

```C++
namespace cbt = google::cloud::bigtable;
using google::cloud::StatusOr;
[](Table table) {
  StatusOr<std::pair<bool, cbt::Row>> row = table.ReadRow("foo");
```

That makes the example more readable to our customers. This gets even more important for more complex return types, such as:

```C++
namespace cbt = google::cloud::bigtable;
using google::cloud::StatusOr;
[](Table table) {
  future<StatusOr<std::pair<bool, cbt::Row>>> row = table.AsyncReadRow("foo");
```

coryan:
Assigning to @remyabel so nobody else will start on it (they already did).  @remyabel please let me know if you want me to un-assign this issue.

remyabel:
You can unassign me. It was fairly tedious doing the PR (even though I was using YCM to get the type of each expression) so others may prefer to tackle individual files.

coryan:
Create a checklist for myself:

- [ ] bigtable_hello_app_profile.cc 
- [ ] bigtable_hello_world.cc 
- [ ] bigtable_instance_admin_snippets.cc 
- [ ] bigtable_quickstart.cc 
- [ ] data_async_snippets.cc 
- [ ] data_snippets.cc 
- [ ] instance_admin_async_snippets.cc: in #2610.
- [ ] table_admin_async_snippets.cc 
- [ ] table_admin_snippets.cc

PR #2622, opened 2019-05-06
Use explicit return types in table_admin_snippets.cc

coryan:
Part of the work for #2604.

I also found a hard-coded key in one of the examples, fixed that and that required some cleaning up in the driver script.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2622)

### DIFF
--- a/google/cloud/bigtable/examples/run_examples_utils.sh
+++ b/google/cloud/bigtable/examples/run_examples_utils.sh
@@ -234,12 +234,16 @@ function run_all_table_admin_examples {
       "${project_id}" "${INSTANCE}" "${zone_id}"
 
   run_example ./bigtable_samples run "${project_id}" "${INSTANCE}" "${TABLE}"
-  run_example ./table_admin_snippets create-table "${project_id}" "${INSTANCE}" "${TABLE}"
-  run_example ./table_admin_snippets list-tables "${project_id}" "${INSTANCE}"
+
+  run_example ./table_admin_snippets create-table \
+      "${project_id}" "${INSTANCE}" "${TABLE}"
+  run_example ./table_admin_snippets list-tables \
+      "${project_id}" "${INSTANCE}"
   run_example ./table_admin_snippets get-table \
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/examples/run_examples_utils.sh
```diff
@@ -274,15 +284,27 @@ function run_all_table_admin_examples {
...
+  run_example ./table_admin_snippets generate-consistency-token \
+      "${project_id}" "${INSTANCE}" "${TABLE}"
   local token
-  token="$(./table_admin_snippets generate-consistency-token "${project_id}" "${INSTANCE}" "${TABLE}" | awk '{print $5}')"
-  run_example ./table_admin_snippets check-consistency "${project_id}" "${INSTANCE}" "${TABLE}" "${token}"
-  run_example ./table_admin_snippets drop-rows-by-prefix "${project_id}" "${INSTANCE}" "${TABLE}"
-  run_example ./data_snippets read-rows "${project_id}" "${INSTANCE}" "${TABLE}"
-  run_example ./table_admin_snippets drop-all-rows "${project_id}" "${INSTANCE}" "${TABLE}"
-  run_example ./table_admin_snippets delete-table "${project_id}" "${INSTANCE}" "${TABLE}"
+  token="$(./table_admin_snippets generate-consistency-token \
```
devbww: Need we call `generate-consistency-token` twice?

coryan: Need?  No.  I do like to see this output (when I look at the output that is):

```
[ RUN      ]  table_admin_snippets wait-for-consistency-check running
[       OK ]  table_admin_snippets wait-for-consistency-check
```

Which does not appear if all we do is capture the return in `token="$(...)"`.

That is, I would prefer to keep it.

devbww: Ack.

PR #2628, opened 2019-05-07
Consolidate InstanceAdmin snippets to one file.

coryan:
It just does not make sense for some code snippets to look different and be on a different file.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2628)


coryan:
> I don't really understand the organization of this yet, but can the remaining 'samples' function get folded into the 'snippets' file as well?

I think it should, I created #2826 to track that.

PR #2632, opened 2019-05-07
Use correct user-agent for internal analytics.

### LINKED ISSUE #2203: Add C++ version information to the user-agent string.
We need to know how many customers enable C++14, C++17 or any other variants.  The information does not need to be perfect, some compilers are iffy in the C++ version reporting, that is Okay.

tmatsuo:
This change will use the x-goog-api-client headers as following examples:
- for bigtable: "gl-cpp/GNU-7.3.0-ex-2011 gccl/v0.9.0+306647f4"
- for storage: "gl-cpp/GNU-7.3.0-ex-2011 gccl/v1.0.0+306647f4"

and use user-agent headers as follows:

- for bigtable "gcloud-cpp/v0.9.0+306647f4"
- for storage: "gcloud-cpp/v1.0.0+306647f4 libcurl/7.60.0 OpenSSL/1.1.1b zlib/1.2.11 c-ares/1.14.0"

These user-agent and x-goog-api-client headers allow our internal analytics pipeline to pick up requests from this library.

also fix #2203 

w.r.t. how to store static string values, now I'm having static std::string, following to the existing code.

Caveat: Some versions of MSVC report the incorrect version in __cplusplus, so the version of the c++ standard is not perfect.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2632)

### DIFF
--- a/google/cloud/bigtable/client_options.cc
+++ b/google/cloud/bigtable/client_options.cc
@@ -82,8 +82,7 @@ ClientOptions& ClientOptions::set_connection_pool_size(std::size_t size) {
 }
 
 std::string ClientOptions::UserAgentPrefix() {
-  std::string agent = "gl-cpp/" + google::cloud::internal::language_version();
-  agent += " gccl/" + version_string();
+  std::string agent = "gcloud-cpp/" + version_string();
   return agent;
 }
 
--- a/google/cloud/bigtable/client_options.cc
+++ b/google/cloud/bigtable/client_options.cc
@@ -82,14 +82,8 @@ ClientOptions& ClientOptions::set_connection_pool_size(std::size_t size) {
 }
 
 std::string ClientOptions::UserAgentPrefix() {
-  std::string agent = "cbt-c++/" + version_string();
-#if GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
-  agent += " ex";
-#else
-  agent += " noex";
-#endif  // GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
-  agent += ' ';
-  agent += google::cloud::internal::compiler();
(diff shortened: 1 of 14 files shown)


tmatsuo:
Please hold off merging this. I want to make sure that our internal pipeline correctly picks up the requests before merging.

coryan:
No worries, in any case, you probably want to fix the CI builds first.

tmatsuo:
@coryan PTAL

I updated the PR and the PR description for also using `x-goog-api-client` headers. This is consistent with most of the client libraries for other languages which are sending both `x-goog-api-client` and `user-agent` headers. 

I think we can remove the `do not merge` label now.

tmatsuo:
The integration test is failing. I assume I'm adding the `x-goog-api-client` header in a wrong place in the code, I'll fix this.

tmatsuo:
Oh, If we add a custom header like `x-goog-api-client` when using the V2 signed URL, we should also add the same header to the signed blob. The SignedUrlV2 integration test passed locally with the last change.

tmatsuo:
@coryan Kokoro Ubuntu test failed, but according to the log, I assume it's just a sporadic error. Is it possible to just re-run Kokoro Ubuntu test?

The log says:
```
[    ERROR ]  storage_bucket_samples get-bucket-policy-only
================ [begin run_example.6O9dj0] ================
./storage_bucket_samples get-bucket-policy-only cloud-cpp-test-bucket-15660-8595-27602
Standard C++ exception raised: Permanent error in GetBucketMetadata: EasyPerform() - CURL error [35]=SSL connect error [UNKNOWN]
```

I think we should fix the flake too though.

coryan:
I restarted the flaky build and created #2660 to track the flakiness.

PR #2637, opened 2019-05-08
Fix doxygen page for Hello World example.

coryan:
Make sure any reference to functions or classes are actual links to the
documentation. Also cross-reference other functions or classes that the
user might be interested in, for example, if we show one type of
mutation, link the other types too.

The output can be reviewed at:

https://coryan.github.io/google-cloud-cpp/0.2637.0/bigtable/bigtable-hello-world.html

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2637)


coryan:
/cc: @billyjacobson in case you are interested.

PR #2638, opened 2019-05-08
Fix doxygen page for Hello Table Admin example.

coryan:
First, the name of the doxygen file did not match the name of the
example, so renamed it. Then make sure all types and functions mentioned
in the document are links to the Doxygen docs. And cross-reference each
portion of the example with other related functions or types.

https://coryan.github.io/google-cloud-cpp/0.2638.0/bigtable/bigtable-hello-table-admin.html

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2638)


coryan:
/cc: @billyjacobson in case you are interested.

PR #2650, opened 2019-05-09
fix!: Inconsistent return type in AsyncCheckAndMutateRow.

### LINKED ISSUE #2602: Return consistent types from {Async,}CheckAndMutateRow.
The blocking version returns a `bool`, the async version returns a raw proto. Probably both of them are wrong.

coryan:
BREAKING CHANGE(bigtable.async): `Table::CheckAndMutateRow` returns
`StatusOr<bool>` to indicate which branch of the predicate was taken
in the atomic change. Meanwhile, `AsyncCheckAndMutateRow()` returned a
`future<StatusOr<proto-with-long-name>>`. Changed the async version to
return `future<StatusOr<bool>>`, note that the Async* functions are
still experimental and subject to change.

This fixes #2602.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2650)

### DIFF
--- a/google/cloud/bigtable/table.h
+++ b/google/cloud/bigtable/table.h
@@ -482,7 +482,7 @@ class Table {
    * @par Example
    * @snippet data_async_snippets.cc async check and mutate
    */
-  future<StatusOr<google::bigtable::v2::CheckAndMutateRowResponse>>
+  future<StatusOr<bool>>
   AsyncCheckAndMutateRow(std::string row_key, Filter filter,
                          std::vector<Mutation> true_mutations,
                          std::vector<Mutation> false_mutations,
--- a/google/cloud/bigtable/table.h
+++ b/google/cloud/bigtable/table.h
@@ -482,11 +482,9 @@ class Table {
    * @par Example
    * @snippet data_async_snippets.cc async check and mutate
    */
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/table.h
```diff
@@ -482,7 +482,7 @@ class Table {
    * @par Example
    * @snippet data_async_snippets.cc async check and mutate
    */
-  future<StatusOr<google::bigtable::v2::CheckAndMutateRowResponse>>
+  future<StatusOr<bool>>
```
devbww: `StatusOr<bool>` is easy to use incorrectly (mistaking `if (b)` for `if (b && *b)`), so perhaps we should avoid it.

coryan: Yeah.... That would require changing the non-Async version too, maybe with a `kPredicateNotMatched` and `kPredicateMatched` enum.  We can either fix both in this PR, or I can create a separate PR to fix both after this one.

I slightly prefer the second option for future software archaeologists.

/cc: @devjgm because he was interested in this.

devbww: > I slightly prefer the second option for future software archaeologists.

Sounds good to me.

PR #2658, opened 2019-05-14
Avoid SIGPIPE signals when connections break.

coryan:
Make the use of CURLOPT_NOSIGNAL optional. Some applications need this
to avoid the signals generated by timeouts in DNS lookups, but
CURLOPT_NOSIGNAL does more than just disabling timeout signals, it also
enables SIGPIPE for closed connections (or more accurately: it disables
the automatic disabling of SIGPIPE), which can make the application
crash on broken connections.

I considered using the approach described here:

https://curl.haxx.se/mail/lib-2010-11/0008.html

but that is problematic on libraries supporting multi-threaded
applications.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2658)

### DIFF
--- a/google/cloud/storage/client_options.h
+++ b/google/cloud/storage/client_options.h
@@ -141,6 +141,12 @@ class ClientOptions {
     return *this;
   }
 
+  bool enable_libcurl_nosignal() const { return enable_libcurl_nosignal_; }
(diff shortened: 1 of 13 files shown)


coryan:
After further thought, this is not the right fix. Closing for now.

PR #2661, opened 2019-05-15
Refactor common code in storage/benchmarks.

coryan:
I am about to create additional benchmarks for the GCS library, so it is
time to refactor common code across these benchmarks. That requires a
new library, some unit tests for the library functions, and a lot of
additional complexity in the BUILD and CMakeLists.txt files to support
both.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2661)


google/cloud/storage/benchmarks/BUILD
```diff
@@ -16,14 +16,60 @@ package(default_visibility = ["//visibility:public"])
...
+    ],
 )
 
-cc_binary(
-    name = "storage_throughput_benchmark",
-    srcs = ["storage_throughput_benchmark.cc"],
-    deps = ["//google/cloud/storage:storage_client"],
+load(
+    ":storage_benchmark_programs.bzl",
+    "storage_benchmark_programs",
```
devbww: Remove trailing comma so it formats on one line like the others?

google/cloud/storage/benchmarks/benchmark_utils.cc
```diff
@@ -0,0 +1,120 @@
...
+// purposes.
+std::chrono::seconds ParseDuration(std::string const& val) {
+  long s = std::stol(val);
+  if (EndsWith(val, "h")) {
+    return s * std::chrono::seconds(std::chrono::hours(1));
+  }
+  if (EndsWith(val, "m")) {
+    return s * std::chrono::seconds(std::chrono::minutes(1));
+  }
+  return std::chrono::seconds(s);
```
devbww: Perhaps at least allowing EndsWith(val, "s") might make uses more readable?

google/cloud/storage/benchmarks/benchmark_utils.h
```diff
@@ -0,0 +1,63 @@
...
+
+constexpr std::int64_t kKB = 1000;
+constexpr std::int64_t kMB = 1000 * kKB;
+constexpr std::int64_t kGB = 1000 * kMB;
+constexpr std::int64_t kTB = 1000 * kGB;
+
+/// Parse a string as a byte size, with support for unit suffixes.
+std::int64_t ParseSize(std::string const& val);
+
+/// Parse a string as a duration in seconds, with support for a unit suffixes.
```
devbww: "in seconds" seems to contradict "unit suffixes".

google/cloud/storage/benchmarks/BUILD
coryan: Done.

google/cloud/storage/benchmarks/benchmark_utils.cc
coryan: Done.

google/cloud/storage/benchmarks/benchmark_utils.h
coryan: Yeah... Since the return resolution is always seconds I changed it to just say what inputs it supports. Let me know what you think.

coryan:
PTAL.

PR #2665, opened 2019-05-16
Remove noex::Table.

### LINKED ISSUE #2097: Remove `noex::Table`

coryan:
Obviously I removed any tests for it too, which is the bulk of this
change. I also made the last adjustments in `bigtable::Table` to not
depend on the `noex::` version, and removed many (but probably not all)
supporting classes for Async* functions. There are separate bugs open to
remove these supporting classes.

This fixes #2097.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2665)


coryan:
/cc: @dopiera FYI, in case it affects your changes for `AsyncReadRows()`.

PR #2668, opened 2019-05-18
Send larger chunks is CurlResumableStreambuf.

### LINKED ISSUE #2662: Streaming uploads send very small chunks.
In CurlResumableStreambuf we send chunks of size `max_buffer_size_` (which really should be renamed to `chunk_size_`) even when the application provides much bigger amounts of data per `.write()` call (aka `xsputn()`).

That leads to poor performance and accumulation of data in the buffers, eventually to crashes due to memory exhaustion.

coryan:
The class was sending chunks of size 256KiB (the default configuration parameter),
even when the application provides a multi-megabyte buffer, and then it would save
the buffer to send it later, on the small 256KiB chunks. Sigh...

This version still performs a lot of data copying, but that is a matter for a future
change (there is already an open bug for it).

Fixes #2662.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2668)

### DIFF
--- a/google/cloud/storage/internal/curl_resumable_streambuf_test.cc
+++ b/google/cloud/storage/internal/curl_resumable_streambuf_test.cc
@@ -184,6 +184,46 @@ TEST(CurlResumableStreambufTest, FlushAfterFullQuantum) {
   EXPECT_STATUS_OK(response);
 }
 
+/// @test Verify that a stream flushes when adding one character at a time.
+TEST(CurlResumableStreambufTest, OverflowFlushAtFullQuantum) {
+  auto mock = google::cloud::internal::make_unique<
+      testing::MockResumableUploadSession>();
+
+  auto const quantum = UploadChunkRequest::kChunkSizeQuantum;
+  std::string const payload(quantum, '*');
+
+  int count = 0;
+  EXPECT_CALL(*mock, UploadChunk(_, _))
+      .WillOnce(Invoke([&](std::string const& p, std::uint64_t s) {
+        ++count;
+        EXPECT_EQ(1, count);
+        auto expected = payload;
+        EXPECT_EQ(expected, p);
+        EXPECT_EQ(0, s);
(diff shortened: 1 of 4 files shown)


coryan:
Sorry, I found some problems for this code, needs more changes before it is ready for review.

coryan:
Okay, ready for review now.

google/cloud/storage/internal/curl_resumable_streambuf_test.cc
```diff
@@ -0,0 +1,276 @@
...
+namespace {
+
+using ::google::cloud::storage::testing::canonical_errors::PermanentError;
+using ::google::cloud::storage::testing::canonical_errors::TransientError;
+using ::testing::_;
+using ::testing::HasSubstr;
+using ::testing::Invoke;
+using ::testing::Return;
+
+/// @test Verify that uploading a small creates a single chunk.
```
devbww: "a small \<something\>"?

google/cloud/storage/internal/curl_resumable_streambuf_test.cc
```diff
@@ -0,0 +1,276 @@
...
+      google::cloud::internal::make_unique<NullHashValidator>());
+
+  for (auto const& c : payload) {
+    streambuf.sputc(c);
+  }
+  auto response = streambuf.Close();
+  EXPECT_STATUS_OK(response);
+}
+
+/// @test Verify that a stream flushes when adding one character at a time.
```
devbww: Same comment as previous test?

coryan: Fixed.

google/cloud/storage/internal/curl_resumable_streambuf_test.cc
coryan: Fixed.

PR #2670, opened 2019-05-19
`future<>::then()` doesn't destroy l-value refs.

dopiera:
Before this PR, if an l-value reference was passed to
`future<>::then()`, it was moved from. This was wrong - they should have
been copied from.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2670)


coryan:
No tests?  FYI, I just merged the fixes for the openSuSE build, you will need to rebase.

PR #2671, opened 2019-05-19
Refactor command-line parsing for benchmarks.

coryan:
Wrote helper function to parse command-line, and use it in both benchmarks.
I changed how the throughput benchmark specifies object sizes, because now
we have something better than "N chunks" where the user has to guess (or
read the code) to figure out what is a chunk.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2671)

### DIFF
--- a/google/cloud/storage/benchmarks/benchmark_parser_test.cc
+++ b/google/cloud/storage/benchmarks/benchmark_parser_test.cc
@@ -51,7 +51,7 @@ TEST(StorageBenchmarksUtilsTest, ParseBoolean) {
   EXPECT_EQ(true, ParseBoolean("True", false));
 
   EXPECT_EQ(false, ParseBoolean("false", true));
-  EXPECT_EQ(false, ParseBoolean("False", false));
+  EXPECT_EQ(false, ParseBoolean("False", true));
 }
 
 }  // namespace
--- a/google/cloud/storage/benchmarks/benchmark_parser_test.cc
+++ b/google/cloud/storage/benchmarks/benchmark_parser_test.cc
@@ -43,6 +43,17 @@ TEST(StorageBenchmarksUtilsTest, ParseDuration) {
   EXPECT_EQ(s(1800).count(), ParseDuration("1800s").count());
 }
 
+TEST(StorageBenchmarksUtilsTest, ParseBoolean) {
+  EXPECT_EQ(true, ParseBoolean("", true));
+  EXPECT_EQ(false, ParseBoolean("", false));
+
+  EXPECT_EQ(true, ParseBoolean("true", false));
+  EXPECT_EQ(true, ParseBoolean("True", false));
+
+  EXPECT_EQ(false, ParseBoolean("false", true));
+  EXPECT_EQ(false, ParseBoolean("False", false));
+}
+
 }  // namespace
 }  // namespace storage_benchmarks
 }  // namespace cloud
--- a/google/cloud/storage/benchmarks/benchmark_utils.cc
+++ b/google/cloud/storage/benchmarks/benchmark_utils.cc
@@ -186,13 +186,10 @@ std::vector<std::string> OptionsParse(std::vector<OptionDescriptor> const& desc,
       d.parser(val);
       // This is a match, consume the argument and stop the search.
       matched = true;
-      next_arg = argv.erase(next_arg);
       break;
     }
-    // If it was not matched just skip it,
-    if (!matched) {
-      ++next_arg;
-    }
+    // If next_arg is matched against any option erase it, otherwise skip it.
+    next_arg = matched ? argv.erase(next_arg) : next_arg + 1;
   }
   return argv;
 }
--- a/google/cloud/storage/benchmarks/benchmark_utils.cc
+++ b/google/cloud/storage/benchmarks/benchmark_utils.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/storage/benchmarks/benchmark_utils.h"
+#include <cctype>
 #include <sstream>
 #include <stdexcept>
 
@@ -124,7 +125,7 @@ bool ParseBoolean(std::string const& val, bool default_value) {
   }
   auto lower = val;
   std::transform(lower.begin(), lower.end(), lower.begin(),
-                 [](char x) { return std::tolower(x); });
+                 [](char x) { return static_cast<char>(std::tolower(x)); });
   if (lower == "true") {
     return true;
   } else if (lower == "false") {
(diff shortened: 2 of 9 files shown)


google/cloud/storage/benchmarks/benchmark_parser_test.cc
```diff
@@ -43,6 +43,17 @@ TEST(StorageBenchmarksUtilsTest, ParseDuration) {
...
 
+TEST(StorageBenchmarksUtilsTest, ParseBoolean) {
+  EXPECT_EQ(true, ParseBoolean("", true));
+  EXPECT_EQ(false, ParseBoolean("", false));
+
+  EXPECT_EQ(true, ParseBoolean("true", false));
+  EXPECT_EQ(true, ParseBoolean("True", false));
+
+  EXPECT_EQ(false, ParseBoolean("false", true));
+  EXPECT_EQ(false, ParseBoolean("False", false));
```
devbww: Did you mean that last `false` to be `true`?

google/cloud/storage/benchmarks/benchmark_utils.cc
```diff
@@ -118,6 +119,78 @@ std::chrono::seconds ParseDuration(std::string const& val) {
...
+  }
+  auto lower = val;
+  std::transform(lower.begin(), lower.end(), lower.begin(),
+                 [](char x) { return static_cast<char>(std::tolower(x)); });
+  if (lower == "true") {
+    return true;
+  } else if (lower == "false") {
+    return false;
+  }
+  throw std::runtime_error("Cannot parse " + val + " as a boolean");
```
devbww: Can we `throw` unconditionally?

google/cloud/storage/benchmarks/benchmark_utils.cc
```diff
@@ -118,6 +119,78 @@ std::chrono::seconds ParseDuration(std::string const& val) {
...
+}
+
+std::string Basename(std::string const& path) {
+  // With C++17 we would use std::filesytem::path, until then do the poor's
+  // person version.
+#if _WIN32
+  return path.substr(path.find_last_of('\\') + 1);
```
devbww: Windows should probably `find_last_of("\\/")`.

google/cloud/storage/benchmarks/benchmark_utils.cc
```diff
@@ -118,6 +119,78 @@ std::chrono::seconds ParseDuration(std::string const& val) {
...
+  if (argv.empty()) {
+    return argv;
+  }
+  std::string const usage = BuildUsage(desc, argv[0]);
+
+  auto next_arg = argv.begin() + 1;
+  while (next_arg != argv.end()) {
+    std::string const& argument = *next_arg;
+
+    // Try to match argv[next_arg] against the options in `desc`
```
devbww: s/argv[next_arg]/`argument`/

google/cloud/storage/benchmarks/benchmark_utils.cc
```diff
@@ -118,6 +119,78 @@ std::chrono::seconds ParseDuration(std::string const& val) {
...
+    bool matched = false;
+    for (auto const& d : desc) {
+      if (argument.rfind(d.option, 0) != 0) {
+        // Not a match, keep searching
+        continue;
+      }
+      std::string val;
+      if (d.option.size() + 1 < argument.size()) {
```
devbww: Should we check `argument[d.option.size()] == '='` too, and continue if not?

google/cloud/storage/benchmarks/benchmark_utils.cc
```diff
@@ -118,6 +119,78 @@ std::chrono::seconds ParseDuration(std::string const& val) {
...
+        val = argument.substr(d.option.size() + 1);
+      }
+      d.parser(val);
+      // This is a match, consume the argument and stop the search.
+      matched = true;
+      argv.erase(next_arg);
+      next_arg = argv.begin() + 1;
```
devbww: `next_arg = argv.erase(next_arg);`?

Although I would be tempted to move that below into a single `next_arg` update.

`next_arg = matched ? argv.erase(next_arg) : next_arg + 1;`

coryan: I am not sure I follow. I made the first change, but `next_arg = next_arg + 1` should only happen if the `for (d : desc)` loop terminates without finding a match, so it needs to be outside the for loop?  I just probably misread what you meant.

google/cloud/storage/benchmarks/benchmark_utils.cc
coryan: Fixed.

google/cloud/storage/benchmarks/benchmark_utils.cc
coryan: Done.

google/cloud/storage/benchmarks/benchmark_utils.cc
coryan: We can because the whole directory is not compiled if exceptions are disabled. I fixed this instance, if we want to enable this directory for `-fno-exceptions` I would prefer to do that in a future PR.

google/cloud/storage/benchmarks/benchmark_parser_test.cc
coryan: Yes, thanks for noticing. Fixed.

coryan:
PTAL.

google/cloud/storage/benchmarks/benchmark_utils.cc
devbww: > should only happen if the for (d : desc) loop terminates without finding a match

Right ... that's why the suggestion tests `matched`.  Unless I'm not following of course.

devbww: The whole suggestion is supposed to be outside the loop, yes. That's what I meant by "move that below", but sorry for not being clear.

google/cloud/storage/benchmarks/benchmark_utils.cc
```diff
@@ -131,14 +132,15 @@ bool ParseBoolean(std::string const& val, bool default_value) {
...
+  google::cloud::internal::ThrowRuntimeError("Cannot parse " + val +
+                                             " as a boolean");
 }
 
 std::string Basename(std::string const& path) {
   // With C++17 we would use std::filesytem::path, until then do the poor's
   // person version.
 #if _WIN32
-  return path.substr(path.find_last_of('\\') + 1);
+  return path.substr(path.find_last_of('\\/') + 1);
```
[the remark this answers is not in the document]
coryan: Fixed.

google/cloud/storage/benchmarks/benchmark_utils.cc
coryan: I think I got it this time.

Issue #2673, opened 2019-05-21
Cmake project doesn't propagate setup for protobuf or gRPC


mbrucher:
I'm currently trying to build the latest 0.9.x version of the branch with a custom protobuf and gRPC (for compatibility reasons, we already use specific versions and cannot upgrade them right now).

### Does this issue affect the `google-cloud-cpp` project?

Yes, this is a problem with building the project.

### What version of `google-cloud-cpp` are you using?

64bf0e32fbd234a709c3aa910878723da9378f2a

### What compiler and version are you using?

```
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/opt/rh/devtoolset-3/root/usr/libexec/gcc/x86_64-redhat-linux/4.9.2/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/opt/rh/devtoolset-3/root/usr --mandir=/opt/rh/devtoolset-3/root/usr/share/man --infodir=/opt/rh/devtoolset-3/root/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --enable-multilib --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --enable-languages=c,c++,fortran,lto --enable-plugin --with-linker-hash-style=gnu --enable-initfini-array --disable-libgcj --with-isl=/builddir/build/BUILD/gcc-4.9.2-20150212/obj-x86_64-redhat-linux/isl-install --with-cloog=/builddir/build/BUILD/gcc-4.9.2-20150212/obj-x86_64-redhat-linux/cloog-install --enable-gnu-indirect-function --with-tune=generic --with-arch_32=i686 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.9.2 20150212 (Red Hat 4.9.2-6) (GCC)

```
### What operating system and version are you using?

CentOS Linux release 7.5

### What were you trying to do?

When building the project with protobuf and gRPC set to package, the built google-apis doesn't get the protobug or gRPC that were (correctly) detected in the parent CMakeCache.txt.

I set these up by calling:

     export CMAKE_PREFIX_PATH=/path/to/protobuf/:/path/to/grpc/
[...]

Issue #2674, opened 2019-05-21
Storage object sample cannot process an object


mbrucher:
After building https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/storage/examples/storage_object_samples.cc as an executable against 0.9.x, I used it to ensure the project was working as expected.

### Does this issue affect the `google-cloud-cpp` project?

I think it does (I'm happy to be proven wrong).

### What component of `google-cloud-cpp` is this related to?

- Cloud Storage (under development).

### What version of `google-cloud-cpp` are you using?

64bf0e32fbd234a709c3aa910878723da9378f2a

### What compiler and version are you using?

```
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/opt/rh/devtoolset-3/root/usr/libexec/gcc/x86_64-redhat-linux/4.9.2/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/opt/rh/devtoolset-3/root/usr --mandir=/opt/rh/devtoolset-3/root/usr/share/man --infodir=/opt/rh/devtoolset-3/root/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --enable-multilib --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --enable-languages=c,c++,fortran,lto --enable-plugin --with-linker-hash-style=gnu --enable-initfini-array --disable-libgcj --with-isl=/builddir/build/BUILD/gcc-4.9.2-20150212/obj-x86_64-redhat-linux/isl-install --with-cloog=/builddir/build/BUILD/gcc-4.9.2-20150212/obj-x86_64-redhat-linux/cloog-install --enable-gnu-indirect-function --with-tune=generic --with-arch_32=i686 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.9.2 20150212 (Red Hat 4.9.2-6) (GCC)

```
### What operating system and version are you using?

CentOS Linux release 7.5

### What were you trying to do?

I was trying to list objects in a bucket, this worked fine. I also checked that I needed the right credentials to do so (without the credentials, I had an error indicating wrong credentials, as expected).
I then tried to download a file through this example and then use read-object on the same file. This failed.

### What did you expect to see?

[...]

PR #2676, opened 2019-05-21
Pass CMAKE_PREFIX_PATH to external projects.

coryan:
This is motivated by #2673 (probably fixes it, but want to wait until the build finishes before declaring success).


mbrucher:
thanks a lot for the fast fix! Will it make it to 0.9.x or should I manually merge this in my repo?

coryan:
> thanks a lot for the fast fix! Will it make it to 0.9.x or should I manually merge this in my repo?

@mbrucher: we probably won't backport this to 0.9.x, the 0.10.x release will be out in about a week though, so it might be easier to just wait?  It would be really helpful if you could test this fix in your environment, do you think you would have a chance to do so?

mbrucher:
Oh, if there is a new release in a week form master, I can probably wait, yes. Especially since I have to solve my issues in #2674 before deploying the package, so probably lots of time for this.
Thanks a lot!

mbrucher:
Oddly, it still could not populate the variables.
Same protobuf warnings, but identical to the main project, the CMAKE_PREFIX_PATH is set properly inside CMakeCache, but doesn't seem to be used?

mbrucher:
Oh, I see, I have a ":" instead of the CMake list.

mbrucher:
OK, odd thing, the main prefix path has both paths I have (one for protobuf and one for gRPC), but the command is split in two:
    '-DCMAKE_PREFIX_PATH=/path/to/protobuf/' '/path/to/grpc/'

instead of being one argument like `CMAKE_INSTALL_RPATH` is.

coryan:
I fear we may have to do the same trick we did for `CMAKE_INSTALL_RPATH` and replace the list separator character in `CMAKE_PREFIX_PATH`.

mbrucher:
Indeed, that may solve the issue. I'm hitting another issue, but that may be due to my protobuf version?

    google/iam/v1/policy.protogoogle/iam/v1/policy.proto::2727::88: : Option "php_namespace" unknown.Option "php_namespace" unknown.

If that's the case, I'll get a newer protobuf, as this would be my problem!

coryan:
Yes, that `php_namespace` error means you are trying to use a version of protobuf that is too old for gRPC.

mbrucher:
Great, thanks for the confirmation. I added the same workaround on CMAKE_PREFIX_PATH and it managed to configure the build, so it should do the trick.
Maybe the old gRPC and protobuf may also explain the issue I faced. Thanks very much for the help.

coryan:
@mbrucher can you take a look and see if the fixes for the separator look right to you?

mbrucher:
Yes, it works as expected.

PR #2678, opened 2019-05-22
Implement TableAdmin::AsyncWaitForConsistency.

### LINKED ISSUE #1500: Implement bigtable::TableAdmin::AsyncWaitForConsistency

coryan:
Implements the function, the examples, an integration test. Because the function
uses an ad-hoc retry loop, I also implemented a new unit test.

Fixes #1500.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2678)


coryan:
/cc: @manish-qlogic because you mentioned you wanted to see the code.

PR #2679, opened 2019-05-22
Remove substring after space in language_version.

tmatsuo:
I found out that the internal analytics pipeline doesn't work well with parentheses; `(` and `)`. I decided to cut the compiler version string after the first space, which is enough for our telemetry purpose.

Examples:

Good x-goog-api-client header: GNU-6.3.0-ex-2011
Bad x-goog-api-client header: compiler-gcc-4.9_(ubuntu_blah_blah)_-ex-2011

This change will cut off the `_(ubuntu_blah_blah)_` part above, from the `x-goog-api-client` header. Just in case some interesting info there, I also add `google::cloud::internal::compiler` to `user-agent` headers so that we won't lose them if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2679)

### DIFF
--- a/google/cloud/CMakeLists.txt
+++ b/google/cloud/CMakeLists.txt
@@ -191,6 +191,7 @@ if (BUILD_TESTING)
         iam_bindings_test.cc
         internal/backoff_policy_test.cc
         internal/big_endian_test.cc
+        internal/build_info_test.cc
         internal/filesystem_test.cc
         internal/future_impl_test.cc
         internal/invoke_result_test.cc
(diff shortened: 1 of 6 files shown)


coryan:
What information are we losing with this change?  Meaning, what goes into that `ubuntu blah` section?  Do any of the platforms in the CI build has interesting bits of information in there?

tmatsuo:
@coryan Thanks for the comment.

Yeah I think the information is not always there and I assume it's some kind of os and packaging information. Because the internal pipeline uses the regex which is now in the test, so in any case we can not utilize the information if it contains parens.  I suggest that we add that information (the full return value from `compiler()`) to user-agent header instead. I'll do that tomorrow, alongside with fixing the test on windows (the regex doesn't seem to work there).

tmatsuo:
@coryan Thanks!

PR #2682, opened 2019-05-23
Enable API/ABI checks for bigtable.

### LINKED ISSUE #694: Enable build breakage when bigtable ABI is declared stable.
Once we decide that the ABI is frozen for Bigtable the CI build should break if a change introduces changes to the ABI.  With #652 we are introducing a build to warn when there is a breakage, but it will not stop the merges yet (nor should it).

coryan:
There are still a couple of changes we may want to make to the API, but
I would prefer to notice those and not miss other breakage.

This fixes #694.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2682)


ci/check-abi.sh
```diff
@@ -70,12 +70,8 @@ check_library() {
 exit_status=0
 # We are keeping the library list alphabetical for now, there is no preferred
```
devbww: This convention is broken.

coryan: Ooops, fixed.

coryan:
PTAL.

Issue #2655, opened 2019-05-10
Automatically restart downloads when server closes connection.


coryan:
If a download is idle for a few minutes the GCS service closes the connection. The `storage::ObjectReadStream` should reconnect, starting after the last byte fully received.

Note that the application may be downloading only a portion of the file, so we should respect any range that the application has set too.

coryan:
Also, the download should resume from the same version of the object, in case "the last version" changed after the download started. Hopefully one of the headers received when the download starts has this information.

PR #2690, opened 2019-05-25
feature: Implement option to read from an offset.

coryan:
Allow application developers to read an object starting at a given
offset. This will also be used to restart downloads from the last
received byte. Fix the code to properly combine this option and the
gcs::ReadRange() option: the highest of the range beginning and the
offset is used as the starting point.

Refactored some code to avoid duplication.

Part of the work for #2655.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2690)

### DIFF
--- a/google/cloud/storage/tests/object_media_integration_test.cc
+++ b/google/cloud/storage/tests/object_media_integration_test.cc
@@ -686,6 +686,89 @@ TEST_F(ObjectMediaIntegrationTest, ReadRangeXml) {
   EXPECT_STATUS_OK(status);
 }
 
+/// @test Read a portion of a relatively large object using the JSON API.
+TEST_F(ObjectMediaIntegrationTest, ReadFromOffsetJSON) {
+  // The testbench always requires multiple iterations to copy this object.
+  StatusOr<Client> client = Client::CreateDefaultClient();
+  ASSERT_STATUS_OK(client);
+
+  std::string bucket_name = flag_bucket_name;
+  auto object_name = MakeRandomObjectName();
+
+  // This produces a 64 KiB text object, normally applications should download
+  // much larger chunks from GCS, but it is really hard to figure out what is
+  // broken when the error messages are in the MiB ranges.
+  long const chunk = 16 * 1024L;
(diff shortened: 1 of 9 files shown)


google/cloud/storage/tests/object_media_integration_test.cc
```diff
@@ -686,6 +686,89 @@ TEST_F(ObjectMediaIntegrationTest, ReadRangeXml) {
...
+/// @test Read a portion of a relatively large object using the JSON API.
+TEST_F(ObjectMediaIntegrationTest, ReadFromOffsetJSON) {
+  // The testbench always requires multiple iterations to copy this object.
+  StatusOr<Client> client = Client::CreateDefaultClient();
+  ASSERT_STATUS_OK(client);
+
+  std::string bucket_name = flag_bucket_name;
+  auto object_name = MakeRandomObjectName();
+
+  // This produces a 64 KiB text object, normally applications should download
```
devbww: s/, normally/. Normally/?

google/cloud/storage/tests/object_media_integration_test.cc
```diff
@@ -686,6 +686,89 @@ TEST_F(ObjectMediaIntegrationTest, ReadRangeXml) {
...
+/// @test Read a portion of a relatively large object using the XML API.
+TEST_F(ObjectMediaIntegrationTest, ReadFromOffsetXml) {
+  // The testbench always requires multiple iterations to copy this object.
+  StatusOr<Client> client = Client::CreateDefaultClient();
+  ASSERT_STATUS_OK(client);
+
+  std::string bucket_name = flag_bucket_name;
+  auto object_name = MakeRandomObjectName();
+
+  // This produces a 64 KiB text object, normally applications should download
```
devbww: Ditto.

coryan: Here too.

google/cloud/storage/tests/object_media_integration_test.cc
coryan: Done.

PR #2691, opened 2019-05-25
bug: Service account credentials not refreshing properly.

### LINKED ISSUE #2689: Service account credentials fail to refresh

yes

Remove the ones that do not apply.

- Cloud Storage.

specifically google/cloud/storage/oauth2

v0.9, but the issue appear to also be in head.

### What compiler and version are you using?

clang 8.0

### What operating system and version are you using?

Ubuntu 16

Use a service account to access GCS

That it would work longer than an hour

### What did you see instead?

Permanent error in ReadObject: {
  "error": "invalid_grant",
  "error_description": "Invalid JWT: Token must be a short-lived token (60 minutes) and in a reasonable timeframe. Check your iat and exp values and use a clock with skew to account for clock differences between systems."
}

I believe this is an issue with the way the ServiceAccountCredentials is implemented. Specifically, "iat" and "exp" are built into the ServiceAccountCredentials::payload_ on construction, and not updated on refresh.

They are created here:https://github.com/googleapis/google-cloud-cpp/blob/c41513a6ec8f6c37294310771a7c06d8931c9867/google/cloud/storage/oauth2/service_account_credentials.h#L192

Then built into the payload_ via the constructor here: https://github.com/googleapis/google-cloud-cpp/blob/c41513a6ec8f6c37294310771a7c06d8931c9867/google/cloud/storage/oauth2/service_account_credentials.h#L113

Then reused without being updated in the Refresh function here: https://github.com/googleapis/google-cloud-cpp/blob/c41513a6ec8f6c37294310771a7c06d8931c9867/google/cloud/storage/oauth2/service_account_credentials.h#L220

note that both "iat" and "exp" are absolute times, and not durations.

coryan:
The code was not updating the expiration time when creating a new credential,
effectively "refreshing" by creating an immediately expired credential. Most of the
changes are (as usual) needed to make the code testable, in particular,
`oauth2::ServiceAccountCredentials` was not using the mockable clock everywhere,
and we needed functions to parse url-safe-base64-encoded strings.

I updated the API baseline because there are a number of changes to private
implementation classes, but the API/ABI checks do not know what is public vs.
what is private.

This fixes #2689.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2691)


google/cloud/storage/internal/openssl_util_test.cc
```diff
@@ -0,0 +1,76 @@
+// Copyright 2018 Google LLC
```
devbww: s/2018/2019/

google/cloud/storage/oauth2/service_account_credentials_test.cc
```diff
@@ -352,6 +359,102 @@ TEST_F(ServiceAccountCredentialsTest, ParseMissingFieldFails) {
...
+/// @test Verify that refreshing a credential updates the timestamps.
+TEST_F(ServiceAccountCredentialsTest, RefreshingUpdatesTimestamps) {
+  auto info = ParseServiceAccountCredentials(kJsonKeyfileContents, "test");
+  ASSERT_STATUS_OK(info);
+
+  auto make_request_assertion = [&info](long timestamp) {
+    return [timestamp, &info](std::string const& p) {
+      std::string const prefix =
+          std::string("grant_type=") + kGrantParamEscaped;
+      EXPECT_THAT(p, HasSubstr(prefix));
```
devbww: It looks like below we're depending on this to "begin with prefix" rather than "has substr prefix".

google/cloud/storage/oauth2/service_account_credentials_test.cc
```diff
@@ -352,6 +359,102 @@ TEST_F(ServiceAccountCredentialsTest, ParseMissingFieldFails) {
...
+      EXPECT_THAT(p, HasSubstr("&assertion="));
```
devbww: And similarly for the next part beginning with "&assertion=".

google/cloud/storage/oauth2/service_account_credentials_test.cc
coryan: Fixed. I tried to use `ASSERT_THAT()`, but that makes the lambda return `void` and we want a different return type.

google/cloud/storage/oauth2/service_account_credentials_test.cc
coryan: See above.

coryan:
PTAL.

google/cloud/storage/oauth2/service_account_credentials_test.cc
```diff
@@ -39,6 +39,7 @@ using ::google::cloud::storage::testing::MockHttpRequestBuilder;
 using ::testing::_;
 using ::testing::An;
 using ::testing::HasSubstr;
+using ::testing::StartsWith;
```
devbww: Re-sort.

coryan:
For the record, I rebased to resolved the conflicts.

PR #2695, opened 2019-05-28
A throughput vs. CPU benchmark.

coryan:
Introduce a new benchmark to measure the throughput when uploading and
downloading objects, as well as the CPU cost (think "cpu cycles per
uploaded byte") to achieve that throughput. We will use this benchmark
to evaluate some performance improvements I have in mind.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2695)

### DIFF
--- a/google/cloud/storage/benchmarks/benchmark_utils.h
+++ b/google/cloud/storage/benchmarks/benchmark_utils.h
@@ -86,12 +86,22 @@ class SimpleTimer {
  public:
   SimpleTimer() = default;
 
+  /// Start the timer, call before the code being measured.
   void Start();
+
+  /// Stop the timer, call after the code being measured.
   void Stop();
 
+  //@{
+  /**
+   * @name Measurement results.
+   *
+   * @note The values are only valid after calling Start() and Stop().
+   */
   std::chrono::microseconds elapsed_time() const { return elapsed_time_; }
   std::chrono::microseconds cpu_time() const { return cpu_time_; }
   std::string const& annotations() const { return annotations_; }
+  //@}
 
  private:
   std::chrono::steady_clock::time_point start_;
--- a/google/cloud/storage/benchmarks/benchmark_utils.h
+++ b/google/cloud/storage/benchmarks/benchmark_utils.h
@@ -19,6 +19,9 @@
 #include <chrono>
 #include <functional>
 #include <string>
+#if GOOGLE_CLOUD_CPP_HAS_GETRUSAGE
+#include <sys/resource.h>
+#endif  // GOOGLE_CLOUD_CPP_HAS_GETRUSAGE
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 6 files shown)


google/cloud/storage/benchmarks/benchmark_utils.h
```diff
@@ -79,6 +82,27 @@ std::string BuildUsage(std::vector<OptionDescriptor> const& desc,
...
                                       std::vector<std::string> argv);
 
+class SimpleTimer {
+ public:
+  SimpleTimer() = default;
+
+  void Start();
+  void Stop();
+
+  std::chrono::microseconds elapsed_time() const { return elapsed_time_; }
```
devbww: Perhaps a note that these are only valid after `Stop()`?

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -0,0 +1,410 @@
...
+
+  return 0;
+} catch (std::exception const& ex) {
+  std::cerr << "Standard exception raised: " << ex.what() << "\n";
+  return 1;
+}
+
+namespace {
+char const* ToString(OpType type) {
+  static char const* kOpTypeNames[] = {"UPLOAD", "DOWNLOAD", "LAST"};
```
devbww: Is just `switch (type) ...` easier?

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -0,0 +1,410 @@
...
+
+  std::uniform_int_distribution<std::uint64_t> size_generator(
+      options.minimum_object_size, options.maximum_object_size);
+  std::uniform_int_distribution<std::uint64_t> chunk_generator(
+      options.minimum_chunk_size, options.maximum_chunk_size);
+
+  auto deadline = std::chrono::steady_clock::now() + options.duration;
+
+  gcs_bm::SimpleTimer timer;
+  // This obviously depends on the size of the objects, but a good estimate is
```
devbww: s/ is$//

google/cloud/storage/benchmarks/benchmark_utils.h
coryan: Done.

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
coryan: Done.

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
coryan: Fixed.

PR #2699, opened 2019-05-29
Return an enum from Table::CheckAndMutateRow.

### LINKED ISSUE #2651: Return StatusOr<SomeEnum> from {Async,}CheckAndMutateRow.
See this discussion:

https://github.com/googleapis/google-cloud-cpp/pull/2650#pullrequestreview-235748888

and #2390.  We are returning `StatusOr<bool>` which are somewhat finicky to use.

coryan:
Returning a bool via `StatusOr<>` is prone to errors, too easy to say
`!status_or` when you mean `!*status_or`.

This fixes #2651.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2699)

### DIFF
--- a/google/cloud/bigtable/table.h
+++ b/google/cloud/bigtable/table.h
@@ -40,12 +40,12 @@ namespace bigtable {
 inline namespace BIGTABLE_CLIENT_NS {
 /// The branch taken by a Table::CheckAndMutateRow operation.
 enum class MutationBranch {
+  /// The predicate provided to CheckAndMutateRow did not match and the
+  /// corresponding mutations (if any) were applied.
   kPredicateNotMatched,
-  ///< The predicate provided to CheckAndMutateRow did not match and the
-  ///< corresponding mutations (if any) were applied.
+  /// The predicate provided to CheckAndMutateRow matched and the corresponding
+  /// mutations (if any) were applied.
   kPredicateMatched,
-  ///< The predicate provided to CheckAndMutateRow matched and the corresponding
-  ///< mutations (if any) were applied.
 };
 
 class MutationBatcher;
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/table.h
```diff
@@ -38,6 +38,16 @@ namespace google {
 namespace cloud {
 namespace bigtable {
 inline namespace BIGTABLE_CLIENT_NS {
+/// The branch taken by a Table::CheckAndMutateRow operation.
+enum class MutationBranch {
+  kPredicateNotMatched,
+  ///< The predicate provided to CheckAndMutateRow did not match and the
```
devbww: Q: Do we like the "after the fact" comments?  I couldn't see any others.

coryan: Thank you for asking. Really, because I picked this habit a long time ago: the formatting for `enum` comments was all wonky with "before the fact" comments, but it is fine now. Fixed.

google/cloud/bigtable/table.h
```diff
@@ -40,12 +40,12 @@ namespace bigtable {
 inline namespace BIGTABLE_CLIENT_NS {
 /// The branch taken by a Table::CheckAndMutateRow operation.
 enum class MutationBranch {
+  /// The predicate provided to CheckAndMutateRow did not match and the
+  /// corresponding mutations (if any) were applied.
```
devbww: Should that be "not applied"?

coryan: There are two sets of mutations, one applied if the predicate matches, another applied when the predicate does not match. So "corresponding mutations" is technically correct.  But maybe I should be more explicit?

devbww: I see.  I'm just naive, so "corresponding" was lost on me.  So yes, perhaps you could say `true_mutations` and `false_mutations` explicitly instead.

coryan:
PTAL.

PR #2701, opened 2019-05-29
Implement Table::AsyncReadRows.

dopiera:
It follows the same principles as `CompletionQueue::MakeStreamingReadRpc`, i.e. accepts a callback for rows and a callback for overall stream status. The row callback is expected to return a `future<bool>`. Before the `future<>` is satisfied, the client will not be receiving more data to allow for flow control.

TBH, I'm genuinely unsure which API I prefer: this or #2579.
What I don't like about #2579:
 - more boilerplate
 - using lambdas is tedious because they have to be recursive

What I don't like about this API:
 - propagating errors from processing a row is tedious - you have to have an external structure to store it (something like `stream_res_override_` in this PR)
 - it's unclear what to do when the provided callbacks throw; even if we wanted to gracefully handle it, I don't know how we should do it (convert an arbitrary exception into a `Status`?); in #2579 lambdas make it well defined

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2701)


dopiera:
I don't think there is much I can do about the ABI incompatibility. It complains that a memfn in `DataClient` was pure virtual and now it isn't. I need that memfn implemented.

coryan:
> I don't think there is much I can do about the ABI incompatibility. It complains that a memfn in `DataClient` was pure virtual and now it isn't. I need that memfn implemented.

I agree, just run this to update the ABI baseline:

```
BUILD_NAME=update-abi ./ci/kokoro/docker/build.sh
```

Then commit `ci/test-abi/bigtable_client.expected.abi.dump.gz`

Issue #2711, opened 2019-05-31
Deadlock in storage::internal::CurlClient::LockShared()


coryan:
This was reported by one of our customers, I am opening the bug for tracking.

The customer reports that an application deadlocks, with a single thread calling a function similar to:

```C++
void F(char const* buf, std::size_t len) {
  auto client = gcs::Client::CreatedDefaultClient().value();
  auto stream = client.WriteObject("my-bucket", "my-object");
  stream.write(buf, len);
  stream.Close();
}
```

The stack trace shows the thread blocked on the mutex here:

https://github.com/googleapis/google-cloud-cpp/blob/bcefc9f6322860b546ee5727cb63cdb191592a7c/google/cloud/storage/internal/curl_client.cc#L1148

No other threads are active in the application.

Several instances of the application deadlocked around the same time (different processes, and different machines).

coryan:
Deadlocks with a single thread are possible if the same mutex is used twice. That could happen if we fail to release the mutex.

Alfus:
I just got a deadlock as well:

I0602 19:37:02.407344     1 database_update.cc:738] Adding last route segment.
*** Aborted at 1559628577 (unix time) try "date -d @1559628577" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGSEGV (@0xea2) received by PID 1 (TID 0x7fc3bb81aa00) from PID 3746; stack trace: ***
    @     0x7fc3bb1fd390 (unknown)
    @     0x7fc3bb1fc26d __lll_lock_wait
    @     0x7fc3bb1f5dbd __GI___pthread_mutex_lock
    @     0x56550f9c392d CurlShareLockCallback
    @     0x56550fa9e27b Curl_share_lock
    @     0x56550fa8d58c Curl_hostcache_prune
    @     0x56550faa1a9a Curl_disconnect
    @     0x56550faa5810 create_conn
    @     0x56550faa26dc Curl_connect
    @     0x56550fa94ae9 multi_runsingle
    @     0x56550fa9410e curl_multi_perform
    @     0x56550fa8c2f0 curl_easy_perform
    @     0x56550f9c1329 google::cloud::storage::v1::internal::CurlRequest::MakeRequest()
    @     0x56550f9f8dbb google::cloud::storage::v1::internal::CurlClient::CreateResumableSessionGeneric<>()
    @     0x56550f9d768c google::cloud::storage::v1::internal::CurlClient::WriteObjectResumable()
[...]

PR #2713, opened 2019-05-31
Add option to disable CURL shared handle.

coryan:
In some versions of libcurl (we are still trying to narrow down which ones) there
are locking bugs for the CURLSH* handles. Users may need to disable the
use of CURLSH* handles altogether (at a performance cost) to workaround
these bugs.

This is related to #2711.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2713)

### DIFF
--- a/google/cloud/storage/client_options.h
+++ b/google/cloud/storage/client_options.h
@@ -147,6 +147,17 @@ class ClientOptions {
     return *this;
   }
 
+  /**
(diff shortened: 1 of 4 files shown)


coryan:
I have now convinced myself this is the wrong fix.  I also have evidence:

https://curl.haxx.se/mail/lib-2019-06/0011.html

PR #2722, opened 2019-06-02
Update googleapis to a recent version.

### LINKED ISSUE #2537: Update our version of "googleapis" protos
Our cmake build is fetching a version of github.com/googleapis/googleapis from [June 2018](https://github.com/googleapis/googleapis/commit/6a3277c0656219174ff7c345f31fb20a90b30b97) ([our cmake config](https://github.com/googleapis/google-cloud-cpp/blob/master/cmake/external/googleapis.cmake#L29)). This old version is missing some recent features from Spanner, such as the batch DML stuff, which was added around [March 2019](https://github.com/googleapis/googleapis/commit/8610b13d3da5e3230cf99c503558961626186249#diff-d3a56721144d7d2ae9f2707f0dc16547). It may be missing other proto features that were added.

This is the same in our [bazel config](https://github.com/googleapis/google-cloud-cpp/blob/master/bazel/google_cloud_cpp_deps.bzl#L45) also.

I'm going to prepare a PR to update our version of the googleapis protos, but this raises a couple questions:

1. Could this break anything else? In general proto and gRPC changes are supposed to be backward compatible, so I want to say "no, nothing will break here", but ya know, maybe? Is there anything for us to do here other than our normal CI testing?

2. Should we do anything more to keep current w/ the rpc protos we're using? Maybe check for and update them after each monthly release? Or just do a manual update when we discover that we need a new version for some reason?

coryan:
Newer versions of googleapis include Bazel BUILD files to compile (most)
of the protos, and in many cases include Bazel rules to generate C++
libraries for these protos. Unfortunately we still need a custom
googleapis.BUILD file because the bigtable protos do not have C++
support (yet).

I had to update gRPC at the same time, because it includes better
`cc_grpc_library()` rules too. And that requires updating protobuf.

One has to be careful to disable the support for other languages in the
WORKSPACE file, or the googleapis Bazel rules require code generators
for Java, Go, and possibly other languages.

I was also able to simplify the dependency on OpenCensus, as gRPC now
pulls the right dependencies for it.

Note that I removed the support for the Spanner protos in the
googleapis.BUILD, the Spanner project knows how to build these.

This is all based on work that @scotthart did for the Spanner library.

Fixes #2537.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2722)


coryan:
Rebased to resolve the conflicts.

coryan:
PTAL, this is ready to merge now that the release branch was created.

coryan:
Ping?

PR #2723, opened 2019-06-02
Adding install rule to install storage/testing/mock_client.h

### LINKED ISSUE #2441: Consider installing storage/testing/mock_client.h
Some of our users may want run unit tests and mock the behavior of `storage::Client`. Realistically, mocking the `storage::internal::RawClient` class is very hard, so maybe we should offer a way to make this easier?

We could just install `google/cloud/storage/testing/mock_client.h` and the users can include it, but there are several considerations that we should think about first:

- Implicitly this is making the `storage::internal::RawClient` class more public than we want it to be, this is a very strong downside.
- Linking against googletest or googlemock is supposedly a bad idea, so we should consider making the "testing library" a pure header library.

manish-qlogic:
This closes #2441

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2723)


manish-qlogic:
@coryan 
PTAL

PR #2725, opened 2019-06-03
gitignore CMakeCache.txt and CMakeFiles/

mr-salty:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2725)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,10 @@ bazel-*
 # Backup files for Emacs
 *~
 
+# Ignore cmake output
+CMakeCache.txt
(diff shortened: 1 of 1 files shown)


coryan:
Hmmm... Why not generate theses files in a separate build directory using `-Bcmake-out` or `-B.build` which are already ignored?  And if you accidentally leave the `-B` option out, wouldn't you want to know?

mr-salty:
yeah, I guess I must have accidentally run cmake without -B at some point. argh.

PR #2729, opened 2019-06-04
Bump version numbers

scotthart:
version numbers increased to one beyond to-be-released numbers

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2729)


google/cloud/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ set(GOOGLE_CLOUD_CPP_IS_RELEASE "")
 
 # This should be included from the top level CMakeLists file
 set(GOOGLE_CLOUD_CPP_VERSION_MAJOR 0)
-set(GOOGLE_CLOUD_CPP_VERSION_MINOR 8)
+set(GOOGLE_CLOUD_CPP_VERSION_MINOR 11)
```
devbww: s/11/9/?

scotthart: Seeing as how I'm working on the v.0.10.x release of google-cloud-cpp, it seemed appropriate that version should be bumped to 0.11.0. Perhaps I incorrectly inferred that adjustment had been overlooked in prior releases. If the value should have gone from 8->9, I'd like to understand what "component" version that's based on.

coryan: This is the version number for the "common" components. Not the greatest name for those variable names in hindsight.

devbww: I'd say you're working on the 0.10.0 release of google-cloud-cpp-bigtable, the 1.1.0 release of google-cloud-cpp-storage, and the 0.8.0 release of google-cloud-cpp(-common).  But that's just my understanding, so I'll defer to someone who knows.

scotthart: changed to 9

google/cloud/google_cloud_cpp_common_version.bzl
```diff
@@ -18,5 +18,5 @@
 
 GOOGLE_CLOUD_CPP_IS_RELEASE = ""
 GOOGLE_CLOUD_CPP_VERSION_MAJOR = "0"
-GOOGLE_CLOUD_CPP_VERSION_MINOR = "8"
+GOOGLE_CLOUD_CPP_VERSION_MINOR = "11"
```
coryan: You probably did not run CMake after changing this number from 11 to 9.

scotthart: re-ran cmake and committed the new version

PR #2730, opened 2019-06-04
 bug: Use one mutex per data kind in CURLSH.

### LINKED ISSUE #2711: Deadlock in storage::internal::CurlClient::LockShared()

coryan:
I misread the documentation for libcurl and created a single mutex for
all the "kinds" of data in a CURLSH handle. This is incorrect:

https://curl.haxx.se/mail/lib-2019-06/0011.html

I have been unable to reproduce the deadlocks that result from this, so
no additional tests.

I wrote a program to reproduce the bug, without success, but I think the
code and the documentation in that program is useful: it shows how to run
a test on multiple machines use Kubernetes.

This fixes #2711.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2730)


google/cloud/storage/internal/curl_client.h
```diff
@@ -224,8 +224,13 @@ class CurlClient : public RawClient,
   std::string xml_download_endpoint_;
   std::string iam_endpoint_;
 
+  std::mutex mu_share_;
+  std::mutex mu_dns_;
+  std::mutex mu_ssl_session_;
+  std::mutex mu_connect_;
+  CurlShare share_;
+
   std::mutex mu_;
```
devbww: Now that this looks like it only protects `generator_`, perhaps a new `GUARDED_BY` comment on that member would be illuminating.

coryan: Done.

coryan:
PTAL.

coryan:
PTAL.

PR #2741, opened 2019-06-06
Support running scan-build manually.

coryan:
Too many false positives to enable automatically, but it does produce
interesting messages. I want to be able to run it manually without too
much effort.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2741)

### DIFF
--- a/google/cloud/bigtable/internal/rowreaderiterator.cc
+++ b/google/cloud/bigtable/internal/rowreaderiterator.cc
@@ -29,21 +29,19 @@ RowReaderIterator::RowReaderIterator() : owner_() {}
 
 // Defined here because it needs to see the definition of RowReader
 RowReaderIterator& RowReaderIterator::operator++() {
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/internal/rowreaderiterator.cc
```diff
@@ -29,19 +29,21 @@ RowReaderIterator::RowReaderIterator() : owner_() {}
 
 // Defined here because it needs to see the definition of RowReader
 RowReaderIterator& RowReaderIterator::operator++() {
-  if (owner_ != nullptr && !row_) {
+  if (owner_ == nullptr) {
```
devbww: Should this be separated?

coryan: Yes. Done.

PR #2744, opened 2019-06-06
Remove unnecessary 'unsigned' literals.

### LINKED ISSUE #2742: Remove unnecessary unsigned literals from the examples.
A lot of the examples have code like `argv.size() != 2U` and that `U` is not needed. Remove them, if some compiler complains, fine tune the warnings for that compiler.

coryan:
Probably because I am lazy, I started using unsigned literals when
comparing to an unsigned. This is not needed, warnings are produced when
comparing signed vs. unsigned *variables*, but the literals are promoted
to the larger (unsigned) type.

This fixes #2742.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2744)


coryan:
> I really didn't intend for you to go change these. I was just wondering if there was a warning somewhere about these comparisons. But anyway, I hope it wasn't too much of a hassle. Thanks.

It was not a hassle at all. And (good news!) we had no such warning enabled.

PR #2749, opened 2019-06-10
Add options to control number of samples.

coryan:
To perform statistical analysis on the benchmark results we need to
control the actual number of iterations. In many cases we want a minimum
number of iterations. I think we still want the options to control the
maximum test duration, it is useful when running quick manual tests.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2749)

### DIFF
--- a/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
+++ b/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
@@ -70,7 +70,7 @@ this loop until a prescribed *time* has elapsed:
 The loop stops when any of the following conditions are met:
 
 - The test has obtained more than a prescribed "maximum number of samples"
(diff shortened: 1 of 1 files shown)


google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -67,9 +67,14 @@ this loop until a prescribed *time* has elapsed:
   used during the download, and the status code for the download.
 - The program then deletes this object and starts another iteration.
 
-The loop runs for a prescribed number of seconds. Once the threads finish
-running their loops the program prints the captured performance data. The bucket
-is deleted after the program terminates.
+The loop stops when any of the following conditions are met:
+
+- The test has obtained more than a prescribed "maximum number of samples"
+- The test has obtained at east a prescribed "minimum number of samples" *and*
```
devbww: s/east/least/

PR #2752, opened 2019-06-10
Use check_cxx_symbol_exists() to search for RUSAGE_THREAD.

coryan:
We were using check_symbol_exists() which uses the C compiler. On Linux, the C++ compiler defines `_GNU_SOURCE` by default, that enables `RUSAGE_THREAD`, and we can use it in our code. The CMake test based on the C compiler could not find the symbol because the C compiler does not define `_GNU_SOURCE`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2752)

### DIFF
--- a/google/cloud/storage/benchmarks/CMakeLists.txt
+++ b/google/cloud/storage/benchmarks/CMakeLists.txt
@@ -22,6 +22,7 @@ target_link_libraries(storage_benchmarks
 
 include(CheckSymbolExists)
 check_symbol_exists(getrusage sys/resource.h GOOGLE_CLOUD_CPP_HAVE_GETRUSAGE)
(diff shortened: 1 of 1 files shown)


coryan:
> Did you mean to say RUSAGE_THREAD instead of RUSAGE_SELF in the PR description?
> 

Fixed.

> [Aside: How come we get this in a "normal" build, but not here?]

Thanks for asking that.  Turns out that we get `_GNU_SOURCE` from the C++ compiler, and I was using the C compiler to test for the symbol.

devbww:
OK, cool.  So now the PR description is obsolete, yes?

coryan:
Yup. Fixed.

PR #2753, opened 2019-06-11
Make build metadata configurable.

coryan:
On development builds we annotate the build with the short git commit
id, but in some cases this is not available (Google Cloud Build does not
send the .git directory to the server), and in other cases we want to
override it. Allow the developer (or script) calling `cmake` to set the
build metadata.

Update API/ABI baseline as this change removes a function in
`google::cloud::internal::`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2753)


google/cloud/CMakeLists.txt
```diff
@@ -72,29 +72,57 @@ include(GoogleCloudCppCommon)
...
-    if (NOT IS_DIRECTORY ${PROJECT_SOURCE_DIR}/.git)
-        set(GOOGLE_CLOUD_CPP_GIT_HEAD "unknown-commit")
-    else()
-        find_program(GIT_PROGRAM NAMES git)
-        if ("${GIT_PROGRAM}" STREQUAL "GIT_PROGRAM-NOTFOUND")
-            set(GOOGLE_CLOUD_CPP_GIT_HEAD "unknown-commit")
-        else()
-            execute_process(COMMAND "${GIT_PROGRAM}" rev-parse --short HEAD
+# Define a function to fetch the current git revision. Using a function creates
+# a new scope, so the variables
```
devbww: "so the variables ..."?

coryan: Fixed, thanks.

coryan:
PTAL.

Issue #2748, opened 2019-06-08
Reduce data copying in ObjectReadStream.


coryan:
There is excessive data copying in ObjectReadStream:

- We should take advantage of the [std::basic_streambuf<char>::xsgetn](https://en.cppreference.com/w/cpp/io/basic_streambuf/sgetn) function.
- Most of the time we should have the libcurl write callbacks write directly into the user buffer provided by `xsgetn`.
- When that is not possible, we should have the callback write directly to a the get area in the `ObjectReadStreambuf`.

Issue #2664, opened 2019-05-16
Eliminate data copying in CurlResumableStreambuf.


coryan:
In `xputsn()` we copy the data into the internal buffer, we should pass the data to libcurl without any additional copying.

coryan:
@dopiera I think I mentioned this in passing today, just FYI.

coryan:
With the latest PR (in flight as I write this):

## Summary for Upload (MiB/s)

| ApiName | Version | count | mean | std | min | 25% | 50%  | 75%  | max |
| ------------ | ---------- | -------- | ------- | ---- | ----- | ------ | ------- | ------- | ----- |
| XML        | baseline |  986.0 |  44.626597 |   4.436110 |  20.593110 |  42.264098 |  45.017343 |  47.479515 |   55.648292 |
|                | develop  | 1222.0 | 71.027427 | 13.893069 | 13.195084 | 62.854907 | 72.098442 | 80.994735 | 101.056221 |

![image](https://user-images.githubusercontent.com/6241635/87179791-50dccd80-c2ad-11ea-99d6-89fca2c48ec3.png)


## Summary for CPU Time per Byte (nanoseconds)

| ApiName | Version | count | mean | std | min | 25% | 50%  | 75%  | max |
| ------------ | ---------- | -------- | ------- | ---- | ----- | ------ | ------- | ------- | ----- |
|  XML       |   baseline |  986.0 | 1.715984 | 0.316655 | 1.332797 | 1.520781 | 1.595095 | 1.731812 |  2.987575 |
|                 |   develop |  1222.0 | 0.897990 | 0.072550 | 0.740600 | 0.848260 | 0.886833 | 0.935264 | 1.302610 |

![image](https://user-images.githubusercontent.com/6241635/87179935-91d4e200-c2ad-11ea-85cc-2e05a3fbb116.png)

PR #2754, opened 2019-06-11
Change hash validators to operate on raw buffers.

coryan:
To minimize data copies we need to operate on `char*` and `char const*`
buffers provided by the application without storing them in a
`std::string`. This PR changes the hash validators to work with such
buffers.

Update the API/ABI baseline because we changed the API for classes in
`google::cloud::storage::internal::`.

Part of the work for #2748 and #2664.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2754)


google/cloud/storage/internal/hash_validator_test.cc
```diff
@@ -224,7 +228,7 @@ TEST(CreateHashValidator, Read_Null) {
       CreateHashValidator(ReadObjectRangeRequest("test-bucket", "test-object")
                               .set_multiple_options(DisableCrc32cChecksum(true),
                                                     DisableMD5Hash(true)));
-  validator->Update("The quick brown fox jumps over the lazy dog");
+  validator->Update("The quick brown fox jumps over the lazy dog", 43);
```
devbww: It looks like there are still 8 instances of the explicit-count variety.

coryan: Thanks! Fixed.

PR #2760, opened 2019-06-12
Remove storage location to match other code samples

grant:
All other languages do not set a location for the bucket metadata. We shouldn't do so here.

See other languages:
https://cloud.google.com/storage/docs/reference/libraries

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2760)

### DIFF
--- a/google/cloud/storage/examples/storage_quickstart.cc
+++ b/google/cloud/storage/examples/storage_quickstart.cc
@@ -39,8 +39,8 @@ int main(int argc, char* argv[]) {
   }
 
   google::cloud::StatusOr<gcs::BucketMetadata> bucket_metadata =
-      client->CreateBucketForProject(bucket_name, project_id,
-          gcs::BucketMetadata());
+      client->CreateBucketForProject(
+          bucket_name, project_id, gcs::BucketMetadata());
 
   if (!bucket_metadata) {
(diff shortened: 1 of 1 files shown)


coryan:
Thanks for submitting this patch. As you probably noticed, the build fails for two reasons:

- With your patch the example is not executable (it compiles, but fails at runtime). If you set the storage class to "regional" then the default location ("US") does not work, do you mind also removing the `set_storage_class()` call?

- You need to reformat the code, I would use `BUILD_NAME=shared ./ci/kokoro/docker/build.sh` on a Linux machine that has Docker installed to do the reformatting.  Or you could run clang-format, though if you do not use the same version the build does you *might* get different output from what we require.

grant:
@coryan Yes, sorry for the back-and-forth. I was not able to run the command on my OSX.

```
BUILD_NAME=shared ./ci/kokoro/docker/build.sh
================================================================
./ci/kokoro/docker/build.sh: line 118: nproc: command not found
```

Will ssh to my gubuntu, clone my fork, and try applying changes there.

grant:
@coryan I guess you'll have to approve on the other tool too?

coryan:
Thanks a lot for working through the build issues!

PR #2761, opened 2019-06-12
Retry network commands for less flake

### LINKED ISSUE #2594: [Flake]: "Install" tests can fail if package repos are temporarily not working.
The `cloud-cpp/install/*` Kokoro builds can fail if the package repositories for the particular distribution are not working for a while. For example see:

https://source.cloud.google.com/results/invocations/c59bdbb0-b8d7-47ad-967d-04cf895869fc/targets

One fix would be to retry on these failures. That might be hard because it is all in one large Dockerfile so we would need to split it.

Another fix would be to cache the Docker image with the development tools in GCR (the Google Cloud Container Registry), docker would not need to actually run the "apt install" if that portion of the Dockerfile has not changed.  That is both potentially faster and should be less flaky.

```
http://mirror.sjc02.svwh.net/fedora-epel/7/x86_64/repodata/b8a254680843940cc0fd372acbdad5fd3fe094ce9893a66023e7235c83d5ea63-other.sqlite.bz2: [Errno 14] HTTP Error 404 - Not Found
Trying other mirror.
 One of the configured repositories failed (Unknown),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:
     1. Contact the upstream for the repository and get them to fix the problem.
     2. Reconfigure the baseurl/etc. for the repository, to point to a working
        upstream. This is most often useful if you are using a newer
        distribution release than is supported by the repository (and the
        packages for the previous distribution release still work).
     3. Run the command with the repository temporarily disabled
            yum --disablerepo=<repoid> ...
     4. Disable the repository permanently, so yum won't use it by default. Yum
        will then just ignore the repository until you permanently enable it
        again or use --enablerepo for temporary usage:
            yum-config-manager --disable <repoid>
        or
            subscription-manager repos --disable=<repoid>
     5. Configure the failing repository to be skipped, if it is unavailable.
        Note that yum will try to contact the repo. when it runs most commands,
        so will have to try and fail each time (and thus. yum will be be much
        slower). If it is a very temporary problem though, this is often a nice
        compromise:
            yum-config-manager --save --setopt=<repoid>.skip_if_unavailable=true
epel: Check uncompressed DB failed
The command '/bin/sh -c yum makecache &&     yum install -y automake cmake3 curl-devel gcc gcc-c++ git libtool make         openssl-devel pkgconfig tar wget which zlib-devel' returned a non-zero code: 1
```

tmatsuo:
Also I updated the README.md and INSTALL.md.

Fixes #2594

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2761)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -155,7 +155,8 @@ Install the minimal development tools:
 
 ```bash
 sudo dnf makecache && \
-sudo dnf install -y cmake gcc-c++ git make openssl-devel pkgconfig zlib-devel
+sudo dnf install -y cmake gcc-c++ git make openssl-devel pkgconfig \
+        zlib-devel
 ```
 
 Fedora includes packages for gRPC, libcurl, and OpenSSL that are recent enough
@@ -572,7 +573,8 @@ We use the `ubuntu-toolchain-r` PPA to get a modern version of CMake:
 sudo apt update && sudo apt install -y software-properties-common
 sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
 sudo apt update && \
-sudo apt install -y cmake3 git gcc g++ make pkg-config tar wget zlib1g-dev
+sudo apt install -y cmake3 git gcc g++ make pkg-config tar wget \
+        zlib1g-dev
 ```
 
 Ubuntu:14.04 ships with a very old version of OpenSSL, this version is not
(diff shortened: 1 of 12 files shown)


tmatsuo:
@coryan Thanks for the comments, PTAL

Issue #894, opened 2018-07-16
clang-tidy should enforce correct naming style for constants


houglum:
Per the GSG [1], constants should be named like` kSomeConstantThing`, not `SOME_CONSTANT_THING`.  clang-tidy should enforce this so that we don't end up with a mixture of the two styles.

PR #2762, opened 2019-06-12
Allow specifying non-default scopes and subject in GoogleDefaultCredentials(). (#1866)

### LINKED ISSUE #1866: Support alternative scope(s) in storage::oauth2::Credentials
GoogleDefaultCredentials() uses the hardcoded "https://www.googleapis.com/auth/cloud-platform" scope when it produces credentials based on Service Account JSON credentials. A caller should be able to optionally pass a set of alternate scopes to use.

timford:
This is to enable callers who don't know ahead of time what type of
credential they'll create (hence calling GoogleDefaultCredentials and
not one of the more specific functions), but who need to specify the
scopes and/or subject to be used if a service account file is used.

This augments the fix for the already-closed #1866.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2762)

### DIFF
--- a/google/cloud/storage/oauth2/google_credentials.cc
+++ b/google/cloud/storage/oauth2/google_credentials.cc
@@ -145,8 +145,9 @@ StatusOr<std::unique_ptr<Credentials>> MaybeLoadCredsFromAdcPaths(
 
   // If the path was specified, try to load that file; explicitly fail if it
   // doesn't exist or can't be read and parsed.
-  return LoadCredsFromPath(path, non_service_account_ok, service_account_scopes,
-                           service_account_subject);
+  return LoadCredsFromPath(path, non_service_account_ok,
+                           std::move(service_account_scopes),
+                           std::move(service_account_subject));
 }
 
 StatusOr<std::shared_ptr<Credentials>> GoogleDefaultCredentials() {
@@ -278,7 +279,8 @@ StatusOr<std::shared_ptr<Credentials>>
 CreateServiceAccountCredentialsFromDefaultPaths(
     google::cloud::optional<std::set<std::string>> scopes,
     google::cloud::optional<std::string> subject) {
-  auto creds = MaybeLoadCredsFromAdcPaths(false, scopes, subject);
+  auto creds =
+      MaybeLoadCredsFromAdcPaths(false, std::move(scopes), std::move(subject));
   if (!creds) {
     return StatusOr<std::shared_ptr<Credentials>>(creds.status());
   }
--- a/google/cloud/storage/oauth2/google_credentials.cc
+++ b/google/cloud/storage/oauth2/google_credentials.cc
@@ -36,13 +36,13 @@ constexpr char kAdcLink[] =
     "https://developers.google.com/identity/protocols/"
     "application-default-credentials";
 
-/// Parses the JSON or P12 file at `path` and creates the appropriate
-/// Credentials type.
-///
-/// If `service_account_scopes` or `service_account_subject` are specified, the
-/// file at `path` must be a P12 service account or a JSON service account. If
-/// a different type of credential file is found, this function returns
-/// nullptr to indicate a service account file wasn't found.
+// Parses the JSON or P12 file at `path` and creates the appropriate
+// Credentials type.
+//
+// If `service_account_scopes` or `service_account_subject` are specified, the
+// file at `path` must be a P12 service account or a JSON service account. If
+// a different type of credential file is found, this function returns
+// nullptr to indicate a service account file wasn't found.
 StatusOr<std::unique_ptr<Credentials>> LoadCredsFromPath(
     std::string const& path, bool non_service_account_ok,
     google::cloud::optional<std::set<std::string>> service_account_scopes,
@@ -110,17 +110,17 @@ StatusOr<std::unique_ptr<Credentials>> LoadCredsFromPath(
                  path + "."));
 }
 
-/// Tries to load the file at the path specified by the value of the Application
-/// Default %Credentials environment variable and to create the appropriate
-/// Credentials type.
-///
-/// Returns nullptr if the environment variable is not set or the path does not
-/// exist.
-///
-/// If `service_account_scopes` or `service_account_subject` are specified, the
-/// found file must be a P12 service account or a JSON service account. If a
-/// different type of credential file is found, this function returns nullptr
-/// to indicate a service account file wasn't found.
+// Tries to load the file at the path specified by the value of the Application
+// Default %Credentials environment variable and to create the appropriate
+// Credentials type.
+//
+// Returns nullptr if the environment variable is not set or the path does not
+// exist.
+//
+// If `service_account_scopes` or `service_account_subject` are specified, the
+// found file must be a P12 service account or a JSON service account. If a
+// different type of credential file is found, this function returns nullptr
+// to indicate a service account file wasn't found.
 StatusOr<std::unique_ptr<Credentials>> MaybeLoadCredsFromAdcPaths(
     bool non_service_account_ok,
     google::cloud::optional<std::set<std::string>> service_account_scopes,
--- a/google/cloud/storage/oauth2/google_credentials.cc
+++ b/google/cloud/storage/oauth2/google_credentials.cc
@@ -32,7 +32,7 @@ namespace storage {
 inline namespace STORAGE_CLIENT_NS {
 namespace oauth2 {
 
-constexpr char ADC_LINK[] =
+constexpr char kAdcLink[] =
     "https://developers.google.com/identity/protocols/"
     "application-default-credentials";
 
@@ -176,7 +176,7 @@ StatusOr<std::shared_ptr<Credentials>> GoogleDefaultCredentials() {
       Status(StatusCode::kUnknown,
              "Could not automatically determine credentials. For more "
              "information, please see " +
-                 std::string(ADC_LINK)));
+                 std::string(kAdcLink)));
 }
 
 std::shared_ptr<Credentials> CreateAnonymousCredentials() {
@@ -293,7 +293,8 @@ CreateServiceAccountCredentialsFromDefaultPaths(
       Status(StatusCode::kUnknown,
              "Could not create service account credentials using Application"
              "Default Credentials paths. For more information, please see " +
-                 std::string(ADC_LINK)));
+                 std::string(kAdcLink
+                            )));
 }
 
 StatusOr<std::shared_ptr<Credentials>>
--- a/google/cloud/storage/oauth2/google_credentials.cc
+++ b/google/cloud/storage/oauth2/google_credentials.cc
@@ -34,7 +34,9 @@ namespace oauth2 {
 
 /// Parses the JSON file at `path` and creates the appropriate Credentials type.
 StatusOr<std::unique_ptr<Credentials>> LoadCredsFromPath(
-    std::string const& path) {
+    std::string const& path,
+    google::cloud::optional<std::set<std::string>> scopes,
+    google::cloud::optional<std::string> subject) {
   namespace nl = google::cloud::storage::internal::nl;
 
   std::ifstream ifs(path);
@@ -55,6 +57,8 @@ StatusOr<std::unique_ptr<Credentials>> LoadCredsFromPath(
       return Status(StatusCode::kInvalidArgument,
                     "Invalid credentials file " + path);
     }
+    info->subject = std::move(subject);
+    info->scopes = std::move(scopes);
     auto credentials =
         google::cloud::internal::make_unique<ServiceAccountCredentials<>>(
             *info);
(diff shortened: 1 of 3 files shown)


timford:
Forgot to mention that this pattern (passing in scopes which will be used only if applicable to the chosen credential type) is common. See
* https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java#L211
* https://godoc.org/golang.org/x/oauth2/google#DefaultClient

timford:
Hi Carlos, thanks for the review. Responding here, since I'm not sure about the access request reviewable.io is making when I try to comment (seems like it's asking for access to all repos I can write to). Hope that's okay.

> Shouldn't we create some kind of error if the scopes and subject are set, but the credential type does not support them?

I don't think so. That would make this library hard to use, because you'd have to know ahead of time what type of credential would be chosen in order to know whether or not to pass the scopes. Or you'd have to write something like the following, which seems silly:

```
  auto status_or_credentials = GoogleDefaultCredentials(scopes);
  if (!status_or_credentials.ok() && status_or_credentials.status() == <some documented status>) {
    status_or_credentials = GoogleDefaultCredentials();
  }
```

Note that the Google-provided Go and Java implementations I pointed to (https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java#L211, https://godoc.org/golang.org/x/oauth2/google#DefaultClient) do not return errors in these situations. 

The semantics I'm looking for are, "Use the scopes argument if you need to create a credential type that requires scopes. If the credential type doesn't take scopes as an argument (because the credential type is hardcoded or previously configured somehow), ignore the scopes argument."

My use-case is a client library that needs credentials carrying certain scopes. I will document that my client library uses Google Application Default Credentials, and that if the GCE default service account is used then the scopes will need to have been set on the VM instance at instance creation time. If the user tries to use my client library with ADC and the instance doesn't have the right scopes, my client library calls will fail, which is expected.

If the google-cloud-cpp library went as far as probing the environment to find out if the given scopes are supported (e.g., was the instance created with compatible scopes?), so that I could provide an earlier error to my users, that would actually be great. But I don't expect this library to go that far.

By the way, note that the Go implementation recently gained the ability to apply the given scopes argument to GCE credentials (https://github.com/golang/oauth2/commit/9f3314589c9a9136388751d9adae6b0ed400978a).  So maybe ComputeEngineCredentials should be updated to be able to take scopes. Similarly, the Go library seems to support scopes to be applied for the "Authorized User" type of credential. I'd like to consider those questions out of scope for the purposes of my pull request, if that's okay.

coryan:
> The semantics I'm looking for are, "Use the scopes argument if you need to create a credential type that requires scopes. If the credential type doesn't take scopes as an argument (because the credential type is hardcoded or previously configured somehow), ignore the scopes argument."

Okay, how can we add that to the function documentation, so the purpose of the optional parameters are clear?

> My use-case is a client library that needs credentials carrying certain scopes. I will document that my client library uses Google Application Default Credentials, and that if the GCE default service account is used then the scopes will need to have been set on the VM instance at instance creation time. If the user tries to use my client library with ADC and the instance doesn't have the right scopes, my client library calls will fail,

That sort of worries me. Essentially if I have an application that **requires** scopes and/or subject, I call this function to create a credential with those properties, and I won't know until much later that the application "helpfully" created the wrong type of credentials for me.

> which is expected.

Maybe... the subtlety of how the different credential types and scopes are used could be easily missed. And since we are adding the code to `google-cloud-cpp` it would be doubly easy to miss.

What if we created a new pair of functions say `CreateServiceAccountCredentialsFromDefaultPaths(scope, subject)` and `CreateServiceAccountCredentialsFromDefaultPaths()` which *only* created a service account credential, failed if the defaults pointed to anything else.  Then, when I call this function I know I either got the valid credentials I need, or I got an error (because the created credential type is not the right type or any of the existing errors).


> By the way, note that the Go implementation recently gained the ability to apply the given scopes argument to GCE credentials (golang/oauth2@9f33145). So maybe ComputeEngineCredentials should be updated to be able to take scopes. Similarly, the Go library seems to support scopes to be applied for the "Authorized User" type of credential. 

That seems like a good improvement for the future, at which point I would be happy to have the function that you are proposing added.

> I'd like to consider those questions out of scope for the purposes of my pull request, if that's okay.

I absolutely agree, you should not be taking on more than needed in this PR.

timford:
> > The semantics I'm looking for are, "Use the scopes argument if you need to create a credential type that requires scopes. If the credential type doesn't take scopes as an argument (because the credential type is hardcoded or previously configured somehow), ignore the scopes argument."
> 
> Okay, how can we add that to the function documentation, so the purpose of the optional parameters are clear?

I felt I already had :) Tried to at least. Note the use of the following language: "scopes the scopes to request during the authorization grant, if the GOOGLE_APPLICATION_CREDENTIALS environment variable points to a service account"

> 
> > My use-case is a client library that needs credentials carrying certain scopes. I will document that my client library uses Google Application Default Credentials, and that if the GCE default service account is used then the scopes will need to have been set on the VM instance at instance creation time. If the user tries to use my client library with ADC and the instance doesn't have the right scopes, my client library calls will fail,
> 
> That sort of worries me. Essentially if I have an application that **requires** scopes and/or subject, I call this function to create a credential with those properties, and I won't know until much later that the application "helpfully" created the wrong type of credentials for me.

I was going for consistency with what I see in other Google-published libraries. I mentioned the Go and Java ones already. Here's the C# version, with similar semantics: https://github.com/googleapis/google-api-dotnet-client/blob/master/Src/Support/Google.Apis.Auth/OAuth2/GoogleCredential.cs#L209-L211

I think there's something to be said for consistency across languages, but of course that's your decision.

That C# library also appears to include the ability to apply the scope afterward. So you can check `credential.IsScopeRequired()` before calling `CreateScoped(scopes)`. That's kind of nice. But note that `CreateScoped(scopes)` just ignores the scopes if they're not required, rather than failing.

I'd be happy with something like `credential.IsScopeRequired()` and `credential.CreateScoped(scopes)`, but I don't know how that would work given the fact that this library adds a default scope today. Maybe `credential.ReplaceScopes(scopes)` or something?

I also haven't investigated how amenable the code is to changing the scope out after credential-creation. I assume it would take some refactoring.

> What if we created a new pair of functions say `CreateServiceAccountCredentialsFromDefaultPaths(scope, subject)` and `CreateServiceAccountCredentialsFromDefaultPaths()` which _only_ created a service account credential, failed if the defaults pointed to anything else. Then, when I call this function I know I either got the valid credentials I need, or I got an error (because the created credential type is not the right type or any of the existing errors).

I'm fine going that way if that's your preference. IMO it's a worse user experience, but it's certainly no biggie from my POV to use this library that way from my code.

Thanks for your thoughtful attention to this PR!

coryan:
> > That sort of worries me. Essentially if I have an application that **requires** scopes and/or subject, I call this function to create a credential with those properties, and I won't know until much later that the ~application~ librar "helpfully" created the wrong type of credentials for me.
> 
> I was going for consistency with what I see in other Google-published libraries. I mentioned the Go and Java ones already. Here's the C# version, with similar semantics: https://github.com/googleapis/google-api-dotnet-client/blob/master/Src/Support/Google.Apis.Auth/OAuth2/GoogleCredential.cs#L209-L211
> I think there's something to be said for consistency across languages, but of course that's your decision.

I do like consistency too, but I like correctness better.

> 
> That C# library also appears to include the ability to apply the scope afterward. So you can check `credential.IsScopeRequired()` before calling `CreateScoped(scopes)`. That's kind of nice. But note that `CreateScoped(scopes)` just ignores the scopes if they're not required, rather than failing.

I like that approach, but sounds like a much larger PR.  

> 
> I'd be happy with something like `credential.IsScopeRequired()` and `credential.CreateScoped(scopes)`, but I don't know how that would work given the fact that this library adds a default scope today. Maybe `credential.ReplaceScopes(scopes)` or something?
> 
> I also haven't investigated how amenable the code is to changing the scope out after credential-creation. I assume it would take some refactoring.

Yes, I believe I would.  I think a function that *only* works for service account credentials is a good compromise until we can do the larger changes, as described below:

> 
> > What if we created a new pair of functions say `CreateServiceAccountCredentialsFromDefaultPaths(scope, subject)` and `CreateServiceAccountCredentialsFromDefaultPaths()` which _only_ created a service account credential, failed if the defaults pointed to anything else. Then, when I call this function I know I either got the valid credentials I need, or I got an error (because the created credential type is not the right type or any of the existing errors).
> 
> I'm fine going that way if that's your preference. IMO it's a worse user experience,

In the case when things just magically work, `GoogleDefaultCredentials(scope, subject)` is a nicer experience, yes.

In the case when the mysteriously fail, `CreateServiceAccountCredentialsFromDefaultPaths(scope, subject)` is nicer because you immediately knows what fails.

In the case when (in a future PR) we support scopes and subjects across all credentials types (or allow changing them ala C#) my proposal will look sad...

> but it's certainly no biggie from my POV to use this library that way from my code.

Let's go with  `CreateServiceAccountCredentialsFromDefaultPaths(scope, subject)`  then.  We know there are downsides, but the obscure failures really worry me.

> Thanks for your thoughtful attention to this PR!

Thanks for spending the time to send it and sticking through the review.

> > Okay, how can we add that to the function documentation, so the purpose of the optional parameters are clear?
> 
> I felt I already had :) Tried to at least. Note the use of the following language: "scopes the scopes to request during the authorization grant, if the GOOGLE_APPLICATION_CREDENTIALS environment variable points to a service account"

I missed that, sorry.  I am sort of average at reading documentation, so probably others would miss it too.  In any case, this is not a problem if we go with the `CreateServiceAccountCredentailsFromDefaultPaths()` approach, right?

coryan:
Ping. Any thoughts on the proposed changes?

timford:
Sorry, busy EOQ! I have added CreateServiceAccountCredentialsFromDefaultPaths, with a version that takes scopes. PTAL

google/cloud/storage/oauth2/google_credentials.cc
```diff
@@ -32,9 +32,21 @@ namespace storage {
 inline namespace STORAGE_CLIENT_NS {
 namespace oauth2 {
 
-/// Parses the JSON file at `path` and creates the appropriate Credentials type.
+constexpr char kAdcLink[] =
+    "https://developers.google.com/identity/protocols/"
+    "application-default-credentials";
+
+/// Parses the JSON or P12 file at `path` and creates the appropriate
```
coryan: nit: Please use just `//` for this comment.  Doxygen comments with `///` are not useful in `.cc` files, I do realize that the previous comment was wrong already....

timford:
Fixed the comments, and fixed the clang-tidy errors about unnecessary copies. The clang-tidy error about the `kAdcLink` link looks wrong to me, but let me know what you think.

coryan:
> Fixed the comments, and fixed the clang-tidy errors about unnecessary copies.

Thanks!

> The clang-tidy error about the `kAdcLink` link looks wrong to me, but let me know what you think.

I think our clang-tidy configuration file is broken, but fixing that is beyond the scope of this PR. As a workaround, just use a `UPPPER_CASE_NAME` and I will fix this in a future PR (#894).

timford:
Done, changed to ADC_LINK

coryan:
I am not sure this helps, but we fixed the `.clang-tidy` settings, you need to rebase to pick them up.

google/cloud/storage/oauth2/google_credentials.cc
```diff
@@ -32,9 +32,21 @@ namespace storage {
...
+// Parses the JSON or P12 file at `path` and creates the appropriate
+// Credentials type.
+//
+// If `service_account_scopes` or `service_account_subject` are specified, the
+// file at `path` must be a P12 service account or a JSON service account. If
+// a different type of credential file is found, this function returns
+// nullptr to indicate a service account file wasn't found.
 StatusOr<std::unique_ptr<Credentials>> LoadCredsFromPath(
-    std::string const& path) {
+    std::string const& path, bool non_service_account_ok,
```
devjgm: No change necessary, but I think we should be careful about using bool parameters, *unless* they are the only argument and their meaning is obvious from the function's name. The reason is that they usually lead to unclear call sites because the sense of their meaning is unclear. Consider a caller calling this function:

```cc
auto x = LoadCredsFromPath(path, false, scopes, subjet);
```
It's unclear what the `false` argument means. It probably means "don't *something*", but don't what?

Callers might notice and try to fix this ambiguity by adding a comment to the argument
```cc
auto x = LoadCredsFromPath(path, /* require service accounts */false, scopes, subjet);
```
But this comment doesn't really help because now it's unclear if false means that we should require service accounts, or if we're saying *do not* require service accounts.

All of these problems can usually be fixed using an enum instead of bool. Consider this callsite instead:

```cc
auto x = LoadCredsFromPath(path, AccountType::kServiceAccount, scopes, subject);
```

Now no comment is needed because the code is perfectly clear on its own.

There's more on this topic at https://abseil.io/tips/94

PR #2763, opened 2019-06-13
Add a kokoro job for generating and uploading refdocs to googleapis.dev

tmatsuo:
This job only works with master branch yet. However, I'd like to start configuring this job and add the github hook so that this job will regularly run for every commit.

It is little bit tricky for this job to work with older branches because the older branches doesn't have the CMake options for changing the relative link URL for Doxygen tagfile.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2763)


ci/kokoro/refdocs.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# work from the git root directory
+cd "$(dirname "$0")/../../"
+
+# Try to determine the branch
+if [[ -z "${KOKORO_GITHUB_COMMIT:-}" ]]; then
+  echo "KOKORO_GITHUB_COMMIT not found, aborting..."
+  exit 1
+else
+  BRANCH=`git branch --no-color --contains ${KOKORO_GITHUB_COMMIT} \
```
remyabel: Here and elsewhere, you should use `$(...)` instead of backticks. Also, enclose `${KOKORO_GITHUB_COMMIT}` in quotes. The build will fail for any shellcheck warnings, you can run `ci/check-style.sh` inside Docker locally.

tmatsuo: Done

tmatsuo:
@coryan PTAL

Side note about opensuse test, when I tried the docker build in question on my workstation last night, it was tremendously slow fetching the zypper repo files. Today it seems OK, so I'm expecting those tests will pass this time, let's see how they go.

If those tests failed just because the repo server was super slow, maybe we can extend the timeout?

remyabel:
> If those tests failed just because the repo server was super slow, maybe we can extend the timeout?

As I stated in https://github.com/googleapis/google-cloud-cpp/issues/2780 based on my research for that particular problem, zypper hangs indefinitely. Carlos' new approach will most likely take care of the issue.

PR #2764, opened 2019-06-13
Reject IAM Binding protos with unknown fields.

dopiera:
The key motivation is the `condition` field, which breaks the assumption
that the there is one member list per-role. However, to make it a tiny
bit more robust we reject all Binding`s unknown fields.

As a byproduct, parsing JSON was changed to return Status instead of
throwing (it had been doing both).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2764)

### DIFF
--- a/google/cloud/bigtable/instance_admin.h
+++ b/google/cloud/bigtable/instance_admin.h
@@ -1023,12 +1023,23 @@ class InstanceAdmin {
       std::vector<std::string> const& permissions);
 
  private:
-  static inline google::cloud::IamPolicy ProtoToWrapper(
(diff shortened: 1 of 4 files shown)


coryan:
FYI, there are some builds stuck in Kokoro.

dopiera:
Looks like I missed something - when I clicked merged, it was saying "all checks passed" :/

coryan:
Yes, but GitHub does not know about a Kokoro build until the Kokoro build starts, and there seems to be a backlog. No big deal, if master breaks I will let you know.

dopiera:
Sorry

PR #2773, opened 2019-06-14
chore: Update all copyright headers.

remyabel:
We may as well do this instead of only updating it everytime we change a file.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2773)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-# Copyright 2017 Google Inc.
+# Copyright 2019 Google LLC
 #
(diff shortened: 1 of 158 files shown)


coryan:
Hey, I am very sorry you spent time on this. But the internal rules are only *new* files get an updated Copyright notice with the *LLC* and the new year. Existing files are supposed to stay the same.

I do not understand it either, but I trust the lawyers to know what they are doing.

remyabel:
I'm guessing it has to do with licensing new/old code.

PR #2778, opened 2019-06-14
cleanup: Fix integration tests for failures.

coryan:
Some of the integration tests for failures used "localhost:0" as a
destination address that was known to fail. Recent versions of libcurl
generate a different error for those addresses. Use "localhost:1" which
should work unless the build is running on a machine running ancient
Unix.

I am planning to update our version of libcurl for external projects, and this
is a pre-requisite for that change.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2778)


coryan:
The openSuSE builds are broken across the board. The arguments for caching the Docker images are getting stronger and stronger.

coryan:
BTW, I am merging with failed tests for openSuSE because the failures are due to external problems. I am knowingly taking the (small) risk that I broke that platform .

PR #2781, opened 2019-06-15
chore: Minimize impact of package server outages.

coryan:
When the package servers for a distro go bad (and they do from time to time),
our builds break because we cannot rebuild the images with the latest and
greatest packages. In this PR we use Google Container Registry (GCR) to cache
the Docker images with all the base packages we need to build the code.
We do this only for the *install* tests, because those use the more obscure
distributions, and they tend to fail more. We will use the same techniques
for all builds in a future PR.

The expectation is that GCR has a better observed SLO that the combination
of all the package servers out there. It is based on GCS, so there is good
reason to believe it should work well.

To prevent our base images from getting stale, the builds on the `master`
branch:
- Try to build the base image, without using the cache.
- If successful, they push the new base image up, and use that to run
  the build.
- If not successful, they still use the cached image to run the build.

This technique does not work well if the Dockerfile installs additional packages
after the initial base image. I plan to refactor the builds in a separate
PR to eliminate such cases.

The cache is also invalidated when the contents of the base image changes,
that is, we add a new tool or depedency to the image. That is relatively
rare. We expect that rolling out such changes when the package servers
are having an outage does not happen often, and if it does we will just
need to wait the outage out.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2781)


ci/kokoro/install/build.sh
```diff
@@ -17,15 +17,71 @@
 set -eu
 
 echo "================================================================"
-echo "Change working directory to project root."
+echo "Change working directory to project root $(date)."
 cd "$(dirname "$0")/../../.."
+
+
```
devbww: Delete extra blank line?

ci/kokoro/install/build.sh
```diff
@@ -17,15 +17,71 @@
...
+echo "================================================================"
+echo "Load Google Container Registry configuration parameters $(date)."
+
+if [[ -f "${KOKORO_GFILE_DIR:-}/gcr-configuration.sh" ]]; then
+  source "${KOKORO_GFILE_DIR:-}/gcr-configuration.sh"
+fi
+
 echo "================================================================"
+echo "Setup Google Container Registry access  $(date)."
```
devbww: Nit: Usually not two spaces before `$(date)`.

ci/kokoro/install/build.sh
```diff
@@ -17,15 +17,71 @@
...
+  gcloud auth activate-service-account --key-file \
+    "${KOKORO_GFILE_DIR}/gcr-service-account.json"
+fi
+gcloud auth configure-docker
 
+readonly DEV_IMAGE="gcr.io/${PROJECT_ID}/google-cloud-cpp/test-install-${DISTRO}"
 echo "================================================================"
-printenv
+echo "Download existing image (if available) for ${DISTRO} $(date)."
+has_cache="/bin/false"
```
devbww: Why "/bin/"?

ci/kokoro/README.md
```diff
@@ -0,0 +1,127 @@
...
+In brief, the Kokoro configuration is split in two:
+
+1. A series of configuration files inside Google define what builds exist, and
+   what resources these builds have access to.
+   * These files are in a hierarchy that mirrors the ci/kokoro directory in this
+     repo.
+   * The `common.cfg` files are parsed first.
+   * The `common.cfg` files are applied according to the directory hierarchy.
+   * Finally any settings in `foobar.cfg` are applied.
+1. A series of configuration files in the `ki/kokoro` directory further define
```
devbww: s/ki/ci/?

ci/kokoro/install/build.sh
coryan: Done.

ci/kokoro/install/build.sh
coryan: Done.

ci/kokoro/install/build.sh
coryan: No good reason, removed here and elsewhere in the script.

ci/kokoro/README.md
coryan: Done.

Issue #2772, opened 2019-06-14
Do not include nlohmann/json.hpp from any public header.


coryan:
We do not want to pollute the namespace for our users with macros or global functions in the nljson.h.

remyabel:
The straight-forward solution is to move the member function definitions of `AuthorizedUserCredentials` into its `.cc` then explicitly instantiate:

```c++
// in authorized_user_credentials.cc
template class AuthorizedUserCredentials<storage::internal::CurlRequestBuilder>;
// in authorized_user_credentials_test.cc
template class AuthorizedUserCredentials<testing::MockHttpRequestBuilder>;
```

The problem is now linking errors exist where we're using `make_unique`:

```c++
google::cloud::internal::make_unique<AuthorizedUserCredentials<>>
```

No idea how to fix this.

Putting `extern template class AuthorizedUserCredentials` in `authorized_user_credentials.h` might work (?) but is undesirable since we'd be including a testing header.

coryan:
Is it possible to refactor the code that uses nl::json to standalone (non-template) functions?

Consider this code:

https://github.com/googleapis/google-cloud-cpp/blob/f6c3c0871ceb7c9df93be6c760c245a7695df535/google/cloud/storage/oauth2/authorized_user_credentials.h#L92-L123

Lines 100 to 122 do not really depend on the template parameters, so we could write a function:

```C++
namespace google::cloud::storage::internal {
StatusOr<RefreshingCredentialsWrapper::TemporaryToken> ParseRefreshResponse(std::string const& payload) {
  // ... basically the code from above
}
}
```

and change the code to:

```C++
  StatusOr<RefreshingCredentialsWrapper::TemporaryToken> Refresh() {
    auto response = request_.MakeRequest(payload_);
    if (!response) {
      return std::move(response).status();
    }
    if (response->status_code >= 300) {
      return AsStatus(*response);
    }
    return ParseRefreshResponse(response->payload);
  }
```

remyabel:
That's a good solution I had not considered, thanks.

remyabel:
Factoring out `AssertionComponentsFromInfo` and `ServiceAccountCredentials::Refresh` is a little trickier since they both depend on `ClockType`. Should the factored out functions be passed `std::chrono::system_clock::now()` or something else?

coryan:
[...]

PR #2784, opened 2019-06-15
cleanup!: Don't include nljson.h in some public headers.

remyabel:
BREAKING CHANGE: this removes two functions from the public API that converted strings and `nljson` objects to a `google::cloud::storage::ServiceAccount`. These functions should have been cleaned up before 1.0 and were never intended for external use.

Ref: #2772 

The only files I do not touch are in the `oauth2` directory as I am unsure how to handle those. I will leave that for someone else.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2784)


coryan:
Thank for you these changes. The code looks great. There are some documentation changes we need to make at the same time though:

- We need to update the API/ABI baseline as part of this change, as we are removing functions from the public APIs.  Arguably those APIs should never have been there, but we should still treat this as a breaking change.
- We need to update the description of the PR to say that this is a breaking change and why (we build the release notes from these). I will be doing that in a couple of minutes.

@remyabel can you run `BUILD_NAME=update-abi ./ci/kokoro/docker.sh` and commit the `storage_client` baseline?

PR #2785, opened 2019-06-17
Use correct docs version

tmatsuo:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2785)

### DIFF
--- a/ci/kokoro/refdocs.sh
+++ b/ci/kokoro/refdocs.sh
@@ -42,7 +42,7 @@ upload_docs() {
   # Create docs metadata
   python3 -m docuploader create-metadata \
     --name "${docs_name}" \
-    --version "${docs_dir}" \
+    --version "${docs_version}" \
     --language cpp
 
   # Upload the docs


tmatsuo:
Travis doesn't pass. Travis cache doesn't seem to work well. The log for the cache section says:
```
$ export CASHER_DIR=${TRAVIS_HOME}/.casher
$ Installing caching utilities
attempting to download cache archive
fetching PR.2785/cache-linux-trusty-81a8cc38b01797cc13e777e2f0ee446d2fcc65e20c58b9774904c8bcd825926f--compiler-gcc.tgz
found cache
adding /home/travis/.cache/bazel to cache
creating directory /home/travis/.cache/bazel
adding /home/travis/build/googleapis/google-cloud-cpp/cmake-out/ccache to cache
creating directory /home/travis/build/googleapis/google-cloud-cpp/cmake-out/ccache
/home/travis/.cache/bazel is not yet cached
/home/travis/build/googleapis/google-cloud-cpp/cmake-out/ccache is not yet cached
```

The cache file seems much smaller than the previous ones.
Has anyone seen similar failure?

remyabel:
In this case, the failure is intentional. If there is no existing cache, the build is aborted prematurely to "warm up the cache" (see https://github.com/googleapis/google-cloud-cpp/blob/151cf0f467405773272bd321712ea3396c436bc5/ci/travis/build-docker.sh#L123). So you can safely restart the build.

Carlos has stated that he plans on moving all of the builds to Kokoro, so this will no longer be an issue.

coryan:
The cache is not warming up though. I have created a bug to track that down.

tmatsuo:
Yeah sorry I was not clear. I already restarted the build few times. The cache file was created, but it is mush smaller than usual cache file, and a consecutive build picks up that small cache file, then thinks there's no cache.

With a hypothesis of hitting total cache size limit or sth, I deleted some old cache files, as well as the cache file for this PR, and restarted the build.

tmatsuo:
Ok, this time, the cache seems working.

I presume they've just started compress the cache file, and failed to correctly decompress the file with the past few builds. With the latest build, the cache script seems like correctly decompressing the cache files, and the build is progressing.

coryan:
Oh cool, I was about to blame Travis, but could not nail it, thanks for following up.

PR #2787, opened 2019-06-18
Automatically build image for benchmarks.

coryan:
In this PR we describe how to automate the steps to create the Docker
image for the benchmarks. Once this is merged to `master`, we can enable
the benchmarks and save their logs, but that will come in separate PRs.

Also include the bigtable benchmarks in the image, as we will enable
those in a future PR too.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2787)


coryan:
I think the Ubuntu install failure and the cache failure (on Travis) are separate issues. The first one was a transient (downloading the googleapis code failed). The second I am debugging elsewhere.

PR #2788, opened 2019-06-19
Initial config for Bigtable benchmarks.

coryan:
Create GKE configuration file (templates) to launch the Bigtable bencharks.
The benchmarks run every hour, using the latest version of the Docker image available
at that time. This Docker image is automatically created on each commit to `master`.

We run the benchmarks periodically because this is the easiest way to guarantee that
only one instance of each benchmark is running.

The raw logs from the benchmarks are captured in BigQuery tables, the analysis and
dashboards will come in a future PR.

This is part of the work to remove the Jenkins builds for google-cloud-cpp.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2788)

### DIFF
--- a/ci/benchmarks/bigtable/README.md
+++ b/ci/benchmarks/bigtable/README.md
@@ -1,0 +1,177 @@
+# Setting up GKE Deployments for the Bigtable Benchmarks.
+
+This document describes how to setup GKE deployments to continuously run the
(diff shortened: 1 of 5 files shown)


coryan:
Rebased to resolve conflicts.

PR #2793, opened 2019-06-20
Enable CMAKE_SKIP_INSTALL_RPATH for googleapis external project

remyabel:
This restores the fix for https://github.com/googleapis/google-cloud-cpp/issues/2286

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2793)


coryan:
And now I remember, yes, I was getting the same failures for shared library buillds.

remyabel:
OK, I will close this for now and come up with a better fix later. Since this test shows what would happen if somebody attempted to link against the installed libraries, then that means we can't just skip the RPATH. But my question remains if somebody is ever going to be directly linking against the googleapi* shared libraries or if we *only* care about the RPATH for our own purposes. If so:

1. If the installed googleapi* shared libraries are meant to be linked against by users we should fix the RPATH (maybe using something like `$ORIGIN` or whatever, I need to do more research)

2. If they're *not* we should be using either be setting the appropriate cmake RPATH flags or refactor the tests (again I need to think about this)

PR #2797, opened 2019-06-21
cleanup: Move openssl_util from gcs::internal to google::cloud::internal

### LINKED ISSUE #2796: Move the Base64 encode/decode functions from g/c/storage/internal -> g/c/internal
We need to use the `Base64Encode()` and `Base64Decode()` functions in google-cloud-cpp-spanner. It would be best to use these from google/cloud/common/ rather than reaching into google/cloud/storage/.

remyabel:
Fixes #2796

This moves openssl_util to google_cloud_cpp_common. Since JwtSigningAlgorithms
doesn't seem to be used anywhere else I also moved that. Furthermore, OpenSSL
is added as a dependency to google_cloud_cpp_common; ~storage should pull it in
transitively now.~

~I am open to better ways of handling the dependency move if there are any.~

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2797)


remyabel:
I suspect the reason for the build failure is `find_pa

PR #2800, opened 2019-06-21
cleanup: Another small amount of import friction.

coryan:
I missed these problems in the last PR to reduce friction.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2800)

### DIFF
--- a/google/cloud/bigtable/tests/filters_integration_test.cc
+++ b/google/cloud/bigtable/tests/filters_integration_test.cc
@@ -306,13 +306,12 @@ TEST_F(FilterIntegrationTest, CellsRowLimit) {
     auto inserted = actual.emplace(cell.row_key(), 0);
     inserted.first->second++;
   }
-  using bigtable::RowKeyType;
   std::map<bigtable::RowKeyType, int> expected{
-      {RowKeyType(prefix + "/one-cell"), 1},
-      {RowKeyType(prefix + "/two-cells"), 2},
-      {RowKeyType(prefix + "/many"), 3},
-      {RowKeyType(prefix + "/many-columns"), 3},
-      {RowKeyType(prefix + "/complex"), 3}};
+      {bigtable::RowKeyType(prefix + "/one-cell"), 1},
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/tests/filters_integration_test.cc
```diff
@@ -306,11 +306,13 @@ TEST_F(FilterIntegrationTest, CellsRowLimit) {
     auto inserted = actual.emplace(cell.row_key(), 0);
     inserted.first->second++;
   }
-  std::map<bigtable::RowKeyType, int> expected{{prefix + "/one-cell", 1},
-                                               {prefix + "/two-cells", 2},
-                                               {prefix + "/many", 3},
-                                               {prefix + "/many-columns", 3},
-                                               {prefix + "/complex", 3}};
+  using bigtable::RowKeyType;
```
devbww: If we're going to use this to shorten the initializers (which I'm ambivalent on), should we also use it for the other instances of RowKeyType in this function?

google/cloud/bigtable/tests/filters_integration_test.cc
```diff
@@ -329,9 +331,11 @@ TEST_F(FilterIntegrationTest, CellsRowOffset) {
     auto inserted = actual.emplace(cell.row_key(), 0);
     inserted.first->second++;
   }
-  std::map<bigtable::RowKeyType, int> expected{{prefix + "/many", 2},
-                                               {prefix + "/many-columns", 2},
-                                               {prefix + "/complex", 78}};
+  using bigtable::RowKeyType;
```
devbww: Ditto.

google/cloud/bigtable/tests/filters_integration_test.cc
coryan: On second thought (thanks for asking) the longer name is fine. Fixed.

google/cloud/bigtable/tests/filters_integration_test.cc
coryan: Done.

Issue #2750, opened 2019-06-10
Resuming an upload session should query the upload state.


coryan:
When we resume a resumable upload session, e.g.:

```C++
auto stream = client.WriteObject(
    bucket_name, object_name, gcs::RestoreResumableUploadSession(upload_id));
```

the returned stream is not initialized with the current state of the upload. This upload may have been completed, and in that case the final state should be returned, and the stream should be immediately "closed". In any case, `next_expected_byte()` should have the right value immediately after the stream is created.

coryan:
Assigning to myself, even though it is @jseph who is doing the work, I just can't seem to assign this issue to them.

coryan:
Marking this closed, I agree with @Jseph  that the remaining bit (`next_expected_byte()`) does not make sense for a closed stream.

PR #2804, opened 2019-06-21
Resuming a Finished Upload Returns Closed Stream

Jseph:
This partially solves issue #2750. Note that this doesn't fix the second half of the issue where next_expected_byte() returns 0.  IMO it's more important to represent that the stream is closed and allow the returned metadata to be correctly populated.  next_expected_byte() doesn't really mean anything for a closed stream.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2804)


google/cloud/storage/internal/logging_resumable_upload_session_test.cc
```diff
@@ -133,6 +135,53 @@ TEST_F(LoggingResumableUploadSessionTest, NextExpectedByte) {
...
+  EXPECT_EQ(1, CountLines("upload url"));
+  EXPECT_EQ(1, CountLines("payload bytes"));
+}
+
+TEST_F(LoggingResumableUploadSessionTest, LastResponseBadStatus) {
+  auto mock = google::cloud::internal::make_unique<
+      testing::MockResumableUploadSession>();
+
+  const StatusOr<ResumableUploadResponse> last_response(
+      Status(StatusCode::kUnavailable, "something bad"));
```
remyabel: Is this supposed to be `StatusCode::kFailedPrecondition`?

google/cloud/storage/internal/object_requests.h
```diff
@@ -473,9 +473,12 @@ struct ResumableUploadResponse {
   static StatusOr<ResumableUploadResponse> FromHttpResponse(
       HttpResponse&& response);
 
+  bool operator==(const ResumableUploadResponse& other) const;
+
   std::string upload_session_url;
   std::uint64_t last_committed_byte;
   std::string payload;
+  bool done;
```
remyabel: Per the GCSG, I would either change this to not be a bool *or* add a comment. For example:

```C++
ResumableUploadResponse{"", quantum - 1, {}, /*done=*/false}
```

Jseph:
___
*[google/cloud/storage/internal/object_requests.h, line 481 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2804#-LhvzfkfGExtoTjxNVaN-r1-481:-LhwGNyu7jmVRvLJUn_9:b-tjwxls) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7c4cd860d31784858d957dfdbb2d64353f05cc17/google/cloud/storage/internal/object_requests.h#L481)):*
<details><summary><i>Previously, remyabel wrote…</i></summary><blockquote>

Per the GCSG, I would either change this to not be a bool *or* add a comment. For example:

```C++
ResumableUploadResponse{"", quantum - 1, {}, /*done=*/false}
```
</blockquote></details>

Ah good point.  Is there currently a preference in this codebase between the two?   My preference is to nest an old-style enum (not enum class) in ResumableUploadResponse with 2 values (something like kInProgress and kDone).  Represents a decent amount of refactoring so I'd like to get consensus before I get started on it.

coryan:
If you have Docker installed in your workstation, you can run one of the CI builds using:

```
$ BUILD_NAME=shared ./ci/kokoro/docker/build.sh
```

Without docker is a bit more complicated: you need to install the Python dependencies, I typically create a python virtual environment:

```
$ cd $HOME
$ virtualenv testbench
```

activate the environment and install the dependencies:

```
$ source testbench/bin/activate
$ pip install flask httpbin gevent gunicorn crc32c
```

Then in a directory where you have compiled the code (with CMake):

```
cd $HOME/google-cloud-cpp
env -C ${MY_BINARY_DIR} $PWD/google/cloud/storage/ci/run_integration_tests.sh
```

I typically use `cmake-build-debug` for `MY_BINARY_DIR` but could be `cmake-out/blah`.  I do not believe the driver scripts work if you compiled with Bazel. To run a single test look at what the scripts do.

```
$ BUCKET_NAME=fake-bucket gunicorn --worker-class gevent --access-logfile - --bind 0.0.0.0:8080 --pythonpath google/cloud/storage/testbench --reload testbench:application
```

and then in another shell (for example):

```
$ env CLOUD_STORAGE_ENABLE_CLOG=yes CLOUD_STORAGE_TESTBENCH_ENDPOINT=http://localhost:8080 ./cmake-build-debug/google/cloud/storage/tests/object_media_integration_test  fake-project fake-bucket --gtest_filter=ObjectMediaIntegrationTest.UploadFileBinary
```

Jseph:
Thanks, just pushed what I think is the fix.  Will test it out locally and iterate as needed.

coryan:
Regarding the `CheckABI` build, this is just there to let us know when we break the API/ABI, but it is overly sensitive: it complains (as in this case) about changes in `google::cloud::storage::internal` which is not part of the public API.

Could you run `BUILD_NAME=update-abi ./ci/kokoro/docker/build.sh` in your workstation? And commit the new baseline for the `storage_client` library?  That will fix that build.

Regarding the ClangTidy build, I think you just need to fix that one.

remyabel:
Your build failed because of clang-format. Try running `env BUILD_NAME=shared ./ci/kokoro/docker/build.sh`.

Jseph:
Awesome! Thanks the help and patience!

PR #2809, opened 2019-06-24
Update the docs link to googleapis.dev

tmatsuo:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2809)


tmatsuo:
@coryan 
That is because we're still using the top page at https://googleapis.github.io/google-cloud-cpp/index.html. I'll migrate the library top page once we migrate it to googleapis.dev.

Issue #2807, opened 2019-06-22
Refactor generic gRPC wrappers out of bigtable library.


coryan:
Things like `google::cloud::bigtable::CompletionQueue` or `google::cloud::bigtable::internal::MakeStatusFromGrpcError()` should be moved to a separate library.

That will make it easier to use them from the spanner client project, and eventually it will be easier to refactor them to a separate repository.

coryan:
Fixed by @remyabel in #2814

PR #2814, opened 2019-06-25
Move grpc_error_delegate to new grpc_wrappers library.

remyabel:
This is part of the work for https://github.com/googleapis/google-cloud-cpp/issues/2807

I have moved `MakeStatusFromRpcError` to a new `grpc_wrappers` library. `bigtable` now depends upon this new library. There is still work to be done for future PR's: adding documentation, README, examples, test install, move the rest of the gRPC wrappers etc.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2814)


devjgm:
Just some thoughts. Nothing binding here.

FWIW, I lean against using the term "gax" here. IIUC, "gax" is a specific library with a specific set of things produced by some other team/project. While they don't currently make a "gax" C++ library, they might one day, and it may be good if we didn't take the name "gax" from them. Additionally, it may be strange if our C++ "gax" library looked quite different from "gax" libraries in other languages.

As for proposing a name? Uhh. What exactly do we plan to put in this library? All the common code that we share between google-cloud-cpp and google-cloud-cpp-spanner? So might it contain more than just grpc-related helpers? If we're looking for library for all our common code I might suggest "common". If it's just grpc helpers, then maybe including "grpc" in the name would be

coryan:
> FWIW, I lean against using the term "gax" here. IIUC, "gax" is a specific library with a specific set of things produced by some other team/project. While they don't currently make a "gax" C++ library, they might one day, and it may be good if we didn't take the name "gax" from them. Additionally, it may be strange if our C++ "gax" library looked quite different from "gax" libraries in other languages.

Ack, though what we had in the gax prototype is exactly the kind of thing we would be putting here:

https://github.com/googleapis/gapic-generator-cpp/tree/master/gax

> 
> As for proposing a name? Uhh. What exactly do we plan to put in this library? All the common code that we share between google-cloud-cpp and google-cloud-cpp-spanner?

Cannot be that. There are things we share with GCS, Bigtable, and Spanner and those cannot depend on gRPC (because GCS does not).

> So might it contain more than just grpc-related helpers? If we're looking for library for all our common code I might suggest "common".

Already taken by `google_cloud_cpp_common` which is used by the `storage_client`.

> If it's just grpc helpers, then maybe including "grpc" in the name would be ok?

What about `google_cloud_cpp_common_grpc`, it is consistent with the `google_cloud_cpp_common` library. We can use:

* Name: `google_cloud_cpp_common_grpc`.
* Artifacts: `libgoogle_cloud_cpp_common_grpc.{a,so}` / `google_cloud_cpp_common.{lib,dll}`
* Headers: `google/cloud/grpc/*.h`
* Namespace: `google::cloud::grpc::`

It can depend on `google_cloud_cpp_common`, but only the libraries that already depend on gRPC (everything but GCS basically) can depend on it.

devjgm:
> > FWIW, I lean against using the term "gax" here. IIUC, "gax" is a specific library with a specific set of things produced by some other team/project. While they don't currently make a "gax" C++ library, they might one day, and it may be good if we didn't take the name "gax" from them. Additionally, it may be strange if our C++ "gax" library looked quite different from "gax" libraries in other languages.
> 
> Ack, though what we had in the gax prototype is exactly the kind of thing we would be putting here:
> 
> https://github.com/googleapis/gapic-generator-cpp/tree/master/gax

OK, thanks for clarifying that. Feel free to push back if you think my slight opposition to using the name "gax" is misguided and we should go that direction.

> 
> > As for proposing a name? Uhh. What exactly do we plan to put in this library? All the common code that we share between google-cloud-cpp and google-cloud-cpp-spanner?
> 
> Cannot be that. There are things we share with GCS, Bigtable, and Spanner and those cannot depend on gRPC (because GCS does not).
> 
> > So might it contain more than just grpc-related helpers? If we're looking for library for all our common code I might suggest "common".
> 
> Already taken by `google_cloud_cpp_common` which is used by the `storage_client`.
> 
> > If it's just grpc helpers, then maybe including "grpc" in the name would be ok?
> 
> What about `google_cloud_cpp_common_grpc`, it is consistent with the `google_cloud_cpp_common` library. We can use:
> 
> * Name: `google_cloud_cpp_common_grpc`.
> * Artifacts: `libgoogle_cloud_cpp_common_grpc.{a,so}` / `google_cloud_cpp_common.{lib,dll}`
> * Headers: `google/cloud/grpc/*.h`
> * Namespace: `google::cloud::grpc::`

All the naming you suggest here, SGTM. 

> It can depend on `google_cloud_cpp_common`, but only the libraries that already depend on gRPC (everything but GCS basically) can depend on it.

coryan:
@remyabel we have agreed on `google_cloud_cpp_common_grpc`. Do you mind changing this PR to reflect my previous [comment](https://github.com/googleapis/google-cloud-cpp/pull/2814#issuecomment-506098300).

remyabel:
If I follow the convention of `bigtable` -> `gax` and `bigtable_client` -> `gax_cpp`, I get: `gax` -> `grpc` and `gax_cpp` -> `google_cloud_cpp_common_grpc`. Obviously `grpc` is a poor name here for the CMake target. What should it be instead?

remyabel:
I decided to rename `gax` and `gax_cpp` both to `google_cloud_cpp_common_grpc.`. Furthermore, due to conflicts, I did a mass sed-replace to change `grpc::Status*` to `::grpc::Status*` which ended up affecting a lot of files. I apologize for the inconvenience.

remyabel:
The code should be ready to review now.

remyabel:
Is the ultimate goal to move `google_cloud_cpp_common` to a separate repository or just `grpc_utils`? If the former, I think this PR is kicking the can down the road. A better approach would be to modularize the components and export targets to explicitly be linked against, say `common::grpc` or `common::foo` so that there aren't any dependency issues.

coryan:
I think eventually we would move both `google_cloud_cpp_common` *and* `grpc_utils` to separate repositories, in principle, a separate repository for each.

The benefit of this PR is that we can start moving other things (`google::cloud::bigtable::CompletionQueue` and friends) to this new library, and we can use these things from `google::cloud::spanner` without also having to link `google::cloud::bigtable`).

remyabel:
It is getting unwieldly to maintain this branch with the rebase. I will close it and continue on another one.

PR #2819, opened 2019-06-26
Implement NativeIamBinding/Policy and ops.

dopiera:
These are an alternative to `Iam{Binding,Bindings,Policy}`, which allows
for manipulating directly on protobufs and hence unblock features like
binding conditions.

The `NativeIam{Binding,Policy}` objects are wrappers around protos, so
they keep unknown fields which allows handling future protocol changes
gracefully.

For now, they are implemented in `google::cloud::bigtable` because adding
`googleapis to google::cloud` is a bigger task. Once that's done, they
should be moved there.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2819)


coryan:
Ping. Is there a new version I should review?

PR #2823, opened 2019-06-27
feat: Add ability to optionally install google/cloud/testing_util

### LINKED ISSUE #2810: Consider optionally installing google/cloud/testing_util.
We may want to uses the symbols in that directory for `google-cloud-cpp-spanner`, for example: `ASSERT_STATUS_OK()` would be handy.

remyabel:
Fixes #2810

This PR adds an option to install the "testing_util" files. The flag is independent of `BUILD_TESTING`. If `BUILD_TESTING` is enabled, it will build the files (since it is needed by other files) but not install them. If the install flag is enabled, it will build and install the files, but not the rest of the tests. This is needed by google-cloud-cpp-spanner.

Furthermore, I refactored the code into a separate `CMakeLists.txt` and added a CMake support file just for the GTest aliases. It is possible the other code/support files is extraneous but I will defer that judgment to review.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2823)


remyabel:
I will rework this PR from the ground up. I apparently had a misunderstanding about how the builds in google-cloud-cpp-spanner were structured.

remyabel:
> Isn't there an old version of this file that now needs to get removed? And probably a BUILD file that loads the old version?

Indeed, thanks for pointing that out. Although after moving said code to `google/cloud/testing_util/BUILD`, I suspect it's now failing because `google_cloud_cpp_common_unit_tests` depends upon it. 

```
load("//google/cloud/testing_util:google_cloud_cpp_testing.bzl", "google_cloud_cpp_testing_hdrs", "google_cloud_cpp_testing_srcs")
```

doesn't seem to be the right idea either. How should I fix it?

coryan:
> 
> doesn't seem to be the right idea either. How should I fix it?

I *think* (but you need to check my math here) you will need to move this code to `testing_util/BUILD` too:

https://github.com/googleapis/google-cloud-cpp/blob/0ac7b896ee447569993717ee6a09bb48573130fc/google/cloud/BUILD#L53-L63


and then change all the references to the testing library, such as this one:

https://github.com/googleapis/google-cloud-cpp/blob/0ac7b896ee447569993717ee6a09bb48573130fc/google/cloud/BUILD#L77

to have the new path:

```
  "//google/cloud/testing_util:google_cloud_cpp_testing",
```

remyabel:
Yep, I should've been more clear that I already tried that and got the `crosses boundary of subpackage` error, which is why I mentioned `load` (since that's where the error appears to be originating). It mentioned using `export_files` as a solution but I don't think that would be a good idea.

coryan:
I get that error when there is no `BUILD` file in a directory, or the `BUILD` file has the wrong `default_visibility` errr.. command/rule/directive/whatever-obscure-name-Bazel-uses. Do you mind sharing the branch name and/or delta with what you tried?

remyabel:
I'm sure I just made a typo somewhere but I've pushed the commit.

remyabel:
If it helps the error message is:

```
/v/google/cloud/BUILD:60:12: Label '//google/cloud:testing_util/assert_ok_test.cc' crosses boundary of subpackage 'google/cloud/testing_util' (perhaps you meant to put the colon here: '//google/cloud/testing_util:assert_ok_test.cc'?)
```

remyabel:
The error should be fixed now. I didn't understand what the error was talking about originally, but now I understand it was complaining about `testing_util/assert_ok_test.cc`.

PR #2830, opened 2019-06-28
Add $ORIGIN to CMAKE_INSTALL_RPATH only on Linux.

remyabel:
Supposedly there exists similar mechanisms on other platforms (`@rpath` on MacOS), however given that this will only affect a small amount of people (i.e, Linux maintainers building RPMs), there's no point in fixing what isn't broken. Similarly, I didn't change the other files because `googleapis` was the only library `check-rpaths` was complaining about.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2830)

### DIFF
--- a/cmake/ExternalProjectHelper.cmake
+++ b/cmake/ExternalProjectHelper.cmake
@@ -134,3 +134,30 @@ function (set_external_project_build_parallel_level var_name)
         set(${var_name} "" PARENT_SCOPE)
     endif ()
 endfunction ()
+
+function (set_external_project_install_rpath)
+    set(GOOGLE_CLOUD_CPP_INSTALL_RPATH "<INSTALL_DIR>/lib;<INSTALL_DIR>/lib64"
+        PARENT_SCOPE)
+
+    # On Linux, using an RPATH that is neither an absolute or relative path is
+    # considered a security risk and will cause package building to fail. We use
+    # the Linux-specific variable $ORIGIN to resolve this.
+    if (UNIX AND NOT APPLE)
+        set(GOOGLE_CLOUD_CPP_INSTALL_RPATH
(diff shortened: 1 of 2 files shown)


remyabel:
> instead of setting both directories all the time.

I thought we needed to set both? IIRC there was an issue where some of the libraries (particularly, the ones where we're setting both) didn't link properly, unless that's been resolved already.

Another hitch with `CMAKE_INSTALL_LIBDIR` is that on Redhat in particular, it appears to be an absolute path (see this [bug](https://bugzilla.redhat.com/show_bug.cgi?id=795542), their logic being it's unspecified if it's relative or absolute. This isn't an issue in the contexts we already use it, but it may result in a surprising future bug.

coryan:
> > instead of setting both directories all the time.
> 
> I thought we needed to set both?

I think we cheat. On some platforms you are supposed to install in `lib` and thefefore need `../lib`, on other platforms you are supposed to install in `lib64, and therefore need `../lib64`.  But some libraries always install themselves in `../lib`, using both makes those libraries when `../lib64` would be the "correct" directory.

Oh, I remember now, I was updating `libcurl` and it no longer needs `ALWAYS_LIB`, but there are other libraries that have the `ALWAYS_LIB`. Never mind...

> IIRC there was an issue where some of the libraries (particularly, the ones where we're setting both) didn't link properly, unless that's been resolved already.
> 
> Another hitch with `CMAKE_INSTALL_LIBDIR` is that on Redhat in particular, it appears to be an absolute path (see this [bug](https://bugzilla.redhat.com/show_bug.cgi?id=795542), their logic being it's unspecified if it's relative or absolute. This isn't an issue in the contexts we already use it, but it may result in a surprising future bug.

Ugh.

PR #2833, opened 2019-06-29
Move grpc_error_delegate to new grpc_utils library.

remyabel:
This is a continuation of https://github.com/googleapis/google-cloud-cpp/pull/2814

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2833)


remyabel:
Hm, I think for the Bazel errors it's because I forgot to add it as a dep to the respective BUILD files. Let me try that.

remyabel:
> We recently changed the clang-tidy config, I am sure you noticed this has to be spelled kVersion now.

Done. PTAL.

PR #2841, opened 2019-06-30
cleanup: Refactor time parsing functions.

### LINKED ISSUE #530: Handle dynamic timezone offset changes in the RFC 3339 parser.
The parser in `storage::internal::ParseRfc3339()` assumes the timezone offset does not change during the lifetime of the program.

coryan:
This PR copies the time parsing functions from the Spanner project to
the common library. It also replaces the existing functions to parse
RFC3339 timestamps with the spanner functions. A future PR will remove
the existing functions in the Cloud Spanner repository in favor of
these.

As a side benefit, this fixes #530

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2841)


coryan:
I may have declared success too early... Seems like the new parser is less strict in some platforms.

devbww:
> I may have declared success too early... Seems like the new parser is less strict in some platforms.

Yes.  All of the new failures (that is, new successes) are of the "short" variety.  `std::get_time()` does not demand that the MDhms fields have width 2.

```
Expected: ParseRfc3339("2018-1-18T14:42:03Z") throws an exception of type std::invalid_argument.
  Actual: it throws nothing.
[  FAILED  ] ParseRfc3339Test.DetectShortMonth (0 ms)

google/cloud/internal/parse_rfc3339_test.cc:204: Failure
Expected: ParseRfc3339("2018-05-1T14:42:03Z") throws an exception of type std::invalid_argument.
  Actual: it throws nothing.
[  FAILED  ] ParseRfc3339Test.DetectShortMDay (0 ms)

google/cloud/internal/parse_rfc3339_test.cc:231: Failure
Expected: ParseRfc3339("2018-05-18T1:42:03Z") throws an exception of type std::invalid_argument.
  Actual: it throws nothing.
[  FAILED  ] ParseRfc3339Test.DetectShortHour (0 ms)

google/cloud/internal/parse_rfc3339_test.cc:258: Failure
Expected: ParseRfc3339("2018-05-18T14:2:03Z") throws an exception of type std::invalid_argument.
  Actual: it throws nothing.
[  FAILED  ] ParseRfc3339Test.DetectShortMinute (0 ms)

google/cloud/internal/parse_rfc3339_test.cc:285: Failure
Expected: ParseRfc3339("2018-05-18T14:42:3Z") throws an exception of type std::invalid_argument.
  Actual: it throws nothing.
[  FAILED  ] ParseRfc3339Test.DetectShortSecond (1 ms)
```

coryan:
> > I may have declared success too early... Seems like the new parser is less strict in some platforms.
> 
> Yes. All of the new failures (that is, new successes) are of the "short" variety.

There is also accepting `t` as a separator for the date vs. time parts:

```
[ RUN      ] ParseRfc3339Test.ParseAlternativeSeparators
unknown file: Failure
C++ exception with description "2018-05-18t14:42:03z: Failed to match RFC3339 date-time" thrown in the test body.
```

RFC3339 requires that both forms be accepted:

https://tools.ietf.org/html/rfc3339#section-5.6

> `std::get_time()` does not demand that the MDhms fields have width 2.

But RFC3339 does, we could take the position that we should be "generous in what we accept".

I had also removed the checks for "month-day must be in this range":

https://tools.ietf.org/html/rfc3339#section-5.7

But now I wonder.

coryan:
PTAL.

Note that `TimestampFromString()` accepts arbitrary timezone offsets, because the existing `ParseRfc3339()` function did too. I renamed the function moving from spanner `TimestampFromStringZ()` in case we want the more string 'Z-only' parsing.

I am not using `std::get_time()` to parse RFC3339 timestamps as it looks that old versions of libc++ and stdlibc++ do not (a) support `t` as a separator between date and time, and (b) do not check the length of the fields, and (c) accept out of range month days.  I left the generic `ParseTime()` function alone because it seems like it could be useful for other purposes.

Still used `std::put_time()` for formatting, and the `TimeZ`, `ZTime` functions to convert between `std::time_t` and `std::tm`.

devbww:
> There is also accepting t as a separator for the date vs. time parts

That should already be handled.  I think the problem with the case you quoted was 'z' -v- 'Z', which we can address.

[Late News: Apparently `strptime()` doesn't do the case-insensitive comparison for literals that `std::get_time()` does.  Sigh.]

I'm not sure what "**say**" means in "_Applications using this syntax may choose, for the sake of readability, to specify a full-date and full-time separated by (say) a space character_", but with `strptime()`/`std::get_time()` we would need to introduce an alternate format specifier to handle that.  I'm not sure that implies building our own parser.  Perhaps we could just do some simple transformations on the input string, that could also handle the case-insensitivity question.

> But RFC3339 does [demand MDhms fields have width 2], we could take the position that we should be "generous in what we accept".

Yes, that is my thought.

> I had also removed the checks for "month-day must be in this range"

We can certainly add a simple month-length check after converting to the `std::tm` in `ParseTime()`.

> Note that TimestampFromString() accepts arbitrary timezone offsets, because the existing ParseRfc3339() function did too. I renamed the function moving from spanner TimestampFromStringZ() in case we want the more string 'Z-only' parsing.

The spanner `TypeCode` proto says, "The time zone must be present, and must be `"Z"`."  If storage/bigtable require arbitrary offsets then I agree we would need to write parsing code (`std::get_time()` doesn't support `tm_gmtoff`, and it is a glibc extension for `strptime()`.)

coryan:
Closing for now. After talking to @devbww we need some more pre-work.

PR #2854, opened 2019-07-04
Implement support for IAM conditions.

### LINKED ISSUE #2732: Support IAM conditions in Cloud Bigtable.
Once we merge #2722 (or something similar) we will get the new IAM protos:

At that time we should change the IAM classes as outlined in #1833. This is the tracking bug for Cloud Bigtable of that work. We need to:

- [ ] Change `bigtable::InstanceAdmin::GetIamPolicy()` and `bigtable::InstanceAdmin::SetIamPolicy()` to return an error if the protos received from the wire are not representable using a `google::cloud::IamPolicy`.

- [ ] Create a new class in `google::cloud::bigtable` (for now) to represent `google.iam.v1.Binding` protos, this class should be a faithful representation of the proto, including the new fields.

- [ ] Create a new class in `google::cloud::bigtable` (for now) to represent `google.iam.v1.Policy` protos. This class should be a faithful representation of the proto, no attempts to reindex the data in any way. Helper functions to add, remove, and change bindings are acceptable, but there should be no "index by role" or "index by role, condition, etc."

- [ ] Create new functions in `google::cloud::bigtable::InstanceAdmin` to return the new classes represented above. Naming is important:
  * `InstanceAdmin::SetIamPolicy` can simply be an overload with the new types.
  * `InstanceAdmin::SetIamPolicyUninterpreted` is not a great name, but start exploring from there.

The first step in this bug is to create child issues for each step. We should expect the usual unit, integration, and examples for these new functions.

dopiera:
This fixes #2732

This is basically a couple of helpers for constructing and printing
relevant protos.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2854)


dopiera:
I've updated the ABI expectations, but TBH, I don't understand what it's complaining about.

It claims that one symbol is removed in the "Source compatibility" category (either I'm reading it wrong or it doesn't specify which one). When locally generating these reports, I get 100% compatibility and the PR doesn't remove anything, so I'll claim it a false positive.

Issue #2442, opened 2019-04-09
Replace ad-hoc Join with Abseil StrJoin


ghost:
This has been brought up in a couple of PRs, so we might as well open an issue to track it. When we add Abseil as a dep, we should be using `StrJoin` in these places:

https://github.com/googleapis/google-cloud-cpp/blob/4bfb249145767fd2740a339403206ac6ed64060a/google/cloud/storage/policy_document.cc#L26

https://github.com/googleapis/google-cloud-cpp/blob/60117041083c3fe9925fab2eb650807eb96066cf/google/cloud/storage/internal/sign_blob_requests.cc#L25

https://github.com/googleapis/google-cloud-cpp/blob/51e147b7c10dab5839fb65a3486ae0438ff8e4d1/google/cloud/storage/bucket_metadata.cc#L30

coryan:
Note that this is blocked by #2309.

Issue #2309, opened 2019-03-23
Add Abseil dep in cmake and bazel


devjgm:
Now that Abseil is close to (or done?) adding a cmake install target (https://github.com/abseil/abseil-cpp/issues/111), we may want to consider adding Abseil as a dep and using it where appropriate. I think a good first step for this particular issue would be to:

1. Add Abseil as a dep to our cmake and bazel builds.

2. Change some small piece of our *internal* code to use Abseil. This will demonstrate that the build deps adding in step 1 work in both cmake and build.

It's important that we don't initially change anything that we export in a public interface until we explicitly decide to use Abseil in interfaces. And example of a change for step 2 that might make sense is changing calls to our internal make_unique -> absl::make_unique. That touches about 275 locations, so that may be a bit big. Another idea might be changing the implementation of `internal::ParseRfc3339` and `internal::FormatRfc3339` to be implemented in terms of the absl/time types and functions. This would be a slightly more involved change, but it would only touch a couple files. 

Maybe there are smaller things we could change to satisfy step 2 above also. Feel free to suggest something. The goal is to test that the build deps work correctly.

coryan:
We should try to find some tests or integration tests that could benefit from using Abseil.  That way we can introduce the dependency without complicating the installation instructions for existing users.

devjgm:
SGTM.

remyabel:
I have this partially (?) working ([WIP branch](https://github.com/remyabel/google-cloud-cpp-1/tree/add-abseil-dep)). The easiest solution is to just use `add_subdirectory` but then we have to vendor abseil and lose all the benefits of using an external project, so this is out.

The current solution requires manually listing all the targets (and I've seen this used in other projects like TensorFlow for example), like so:

```cmake
    # Create imported target absl::config
    add_library(absl::config INTERFACE IMPORTED)
    set_library_properties_for_external_project(absl::config absl_config
                                                HEADER_ONLY)
    add_dependencies(absl::config abseil_project)
    # ...and so on...
```

I have managed to get `absl::StrJoin` working but `absl::FromChrono` is not working. For some reason, despite me specifying linking against `absl::time`, the bigtable benchmarks are instead linking against `absl::dynamic_annotations` and `absl::civil_time`. So either I made a typo or there is a flaw in my approach somewhere.

[...]

PR #2856, opened 2019-07-04
WIP: add abseil cmake dependency

remyabel:
Fixes https://github.com/googleapis/google-cloud-cpp/issues/2442

Ref: https://github.com/googleapis/google-cloud-cpp/issues/2309

This begins the work of reintroducing Abseil. For this PR, I have added abseil as an external project. There are some things to be ironed out (see the TODO's) which is why I'm opening this PR early. For future PRs, we will need to add the Bazel files/test installs.

Some other discussion points:

- For the external project, I have explicitly disabled the tests/googletest. Although it may be possible that a user may use an Abseil with googletest embedded, since we define our own aliases I don't think it will conflict.

- I haven't looked into the CCTZ dependency yet. AFAIK, CCTZ is now embedded into Abseil but LMK if I'm wrong.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2856)

### DIFF
--- a/cmake/IncludeAbseil.cmake
+++ b/cmake/IncludeAbseil.cmake
@@ -1,0 +1,29 @@
+# Copyright 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


remyabel:
Abandoning this approach. Currently I'm defining the import targets, but the problem is the libraries have a lot of dependencies, often on internal libraries, for example `absl::time` depends upon:

```
    absl::base
    absl::core_headers
    absl::int128
    absl::strings
    absl::civil_time
    absl::time_zone
```

PR #2858, opened 2019-07-05
cleanup: Refactor ObjectWriteStreambuf.

coryan:
There `internal::CurlObjectWriteStreambuf` was not adding value, there
was nothing "curl" about it (not anymore), it just made the inheritance
tree deeper.

Updated the API/ABI baseline, all the changes are in the `internal::`
namespace and therefore not part of the public API.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2858)


google/cloud/storage/internal/object_streambuf.h
```diff
@@ -89,7 +91,13 @@ class ObjectReadStreambuf : public std::basic_streambuf<char> {
  */
 class ObjectWriteStreambuf : public std::basic_streambuf<char> {
  public:
-  ObjectWriteStreambuf() : std::basic_streambuf<char>() {}
+  ObjectWriteStreambuf() = default;
+
+  explicit ObjectWriteStreambuf(
```
devbww: Do we bother marking multi-argument constructors as `explicit`?

google/cloud/storage/testbench/gcs_bucket.py
```diff
@@ -601,6 +601,7 @@ def receive_upload_chunk(self, gcs_url, request):
         total = end
         final_chunk = False
         content_range = request.headers.get('content-range')
+        print("\n\n%s" % content_range)
```
devbww: Testing remnant?

google/cloud/storage/internal/object_streambuf.h
coryan: Removed. We have some multi-argument constructors marked explicit and some not `¯\_(ツ)_/¯`

google/cloud/storage/testbench/gcs_bucket.py
coryan: Yes. Removed.

coryan:
PTAL.

PR #2864, opened 2019-07-09
bug: Fix ReadObject() when reading the last chunk.

### LINKED ISSUE #2860: Incorrect results in ReadObject().
Create an object with size 3 MiB + 129 KiB. Then use `ReadObject(..., ReadRange(a, b))` to:

* Read the range [0, 1MiB)
* Read the range [1MiB, 2MiB)
* Read the range [2MiB, 3MiB)
* Try to read the range [3MiB, 4MiB)

The expected result is that the last read succeeds, reads 129KiB and sets the eof(), and fail() flags. The actual result is that the last read succeed, but reads only 12**8**KiB.

coryan:
When reading the last chunk, if the chunk is between 128KiB and
128KiB+16KiB, the library would return a short chunk (sometimes, it is
complicated). Basically opening the stream would "peek" the data stream,
with 128KiB. When unlucky, the stream would download all the data (not
just the first 128KiB), mark the stream as closed, and save the extra bytes
in the spill buffer.

Future reads would not retrieve the data from the spill buffer because the
stream was already closed.

This fixes #2860

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2864)

### DIFF
--- a/google/cloud/storage/tests/object_media_integration_test.cc
+++ b/google/cloud/storage/tests/object_media_integration_test.cc
@@ -327,19 +327,21 @@ TEST_F(ObjectMediaIntegrationTest, ReadLastChunk) {
 
   // This produces an object larger than 3MiB, but with a size that is not a
   // multiple of 128KiB.
-  auto const kKiB = 1024L;
-  auto const kMiB = 1024L * kKiB;
-  auto const object_size = 3 * kMiB + 129 * kKiB;
-  auto const lines = object_size / 128;
+  auto constexpr kKiB = 1024L;
+  auto constexpr kMiB = 1024L * kKiB;
+  auto constexpr object_size = 3 * kMiB + 129 * kKiB;
+  auto constexpr line_size = 128;
+  auto constexpr lines = object_size / line_size;
   std::string large_text;
   for (long i = 0; i != lines; ++i) {
-    auto line = google::cloud::internal::Sample(generator_, 127,
+    auto line = google::cloud::internal::Sample(generator_, line_size - 1,
                                                 "abcdefghijklmnopqrstuvwxyz"
                                                 "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
                                                 "012456789");
     large_text += line + "\n";
   }
-  static_assert(object_size % 128 == 0, "Object must be multiple of line size");
+  static_assert(object_size % line_size == 0,
+                "Object must be multiple of line size");
   EXPECT_EQ(object_size, large_text.size());
 
   StatusOr<ObjectMetadata> source_meta = client->InsertObject(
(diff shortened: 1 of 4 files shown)


google/cloud/storage/tests/object_media_integration_test.cc
```diff
@@ -317,6 +317,123 @@ TEST_F(ObjectMediaIntegrationTest, ReadMixedChunks) {
...
+  auto const lines = object_size / 128;
+  std::string large_text;
+  for (long i = 0; i != lines; ++i) {
+    auto line = google::cloud::internal::Sample(generator_, 127,
+                                                "abcdefghijklmnopqrstuvwxyz"
+                                                "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                                                "012456789");
+    large_text += line + "\n";
+  }
+  static_assert(object_size % 128 == 0, "Object must be multiple of line size");
```
devbww: Doesn't this require that object_size is (explicitly) constexpr?

google/cloud/storage/tests/object_media_integration_test.cc
```diff
@@ -317,6 +317,123 @@ TEST_F(ObjectMediaIntegrationTest, ReadMixedChunks) {
...
+
+  stream.read(buffer.data(), buffer.size());
+  EXPECT_TRUE(stream.eof());
+  EXPECT_TRUE(stream.fail());
+  EXPECT_FALSE(stream.bad());
+  EXPECT_EQ(object_size - 3 * kMiB, stream.gcount());
+  std::string actual(buffer.data(), stream.gcount());
+  auto expected = large_text.substr(3 * kMiB);
+  EXPECT_EQ(expected.size(), actual.size());
+  // EXPECT_EQ(expected, actual);
```
devbww: Delete or uncomment?

google/cloud/storage/tests/object_media_integration_test.cc
coryan: I do not think it *requires* it. `object_size` is a constant expression [1], its value is known at compile time, regardless of whether it is declared `constexpr` or not.  This is how we did `const int n = 10; int array[n];` in C++98 and C++03.  But I think it is probably more readable if we *say* that this is indeed a `constexpr`, and therefore changed the code.  I think I will adopt this as a practice.


[1]: I am, of course, referring to the expression that evaluates `object_size`, which is not the same thing as the object `object_size`, which is an object of type `long const` (or thereabouts, I think at least one of the literals has an `L` and propagates to all the constants). But this is me trying to be pedantic, which is bound to blow up in my face, I am terrible at these things :grinning:

google/cloud/storage/tests/object_media_integration_test.cc
coryan: Fixed, thanks!

google/cloud/storage/tests/object_media_integration_test.cc
devbww: Ack.

PR #2865, opened 2019-07-10
Implement native IAM operations for GCS.

dopiera:
native representation.

The native operations use raw json in this implementation.

The alternatives were to either use strings or some wrappers for either
strings or json. The biggest problem with json is exposing an internal
in the API. It seems to be the least bad option, because:
- strings are hard to manipulate on (think iterating over bindings or
    members)
- wrapper objects introduce the problem which we're trying to circumvent
    by this reimplemenatation.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2865)


coryan:
I think we should close this PR until the new version we discussed over IM is ready, thoughts?

dopiera:
Closing, because this will be changed not to expose raw json.

PR #2867, opened 2019-07-11
cleanup: Move bigtable::CompletionQueue to grpc_utils.

coryan:
This will make `CompletionQueue` available for other libraries. I would
like to use it in the Spanner library to perform background operations
without creating an excessive number of threads.

Updated the API/ABI baselines. Because we still have a
`google::cloud::bigtable::CompletionQueue` alias the API does not
change, in fact, none of the examples or tests had to change. But the
tool does not realize the alias is there, so it claims we removed about
10% of the symbols.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2867)


coryan:
Ping.

PR #2868, opened 2019-07-12
Set docker_uid to root if TEST_INSTALL=yes

remyabel:
This conditional existed in the original PR but for some reason was removed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2868)


coryan:
Do we need this for any of the existing builds?  And if we do, can we get the same coverage from the ci/kokoro/readme and/or the ci/kokoro/install builds?

remyabel:
> Do we need this for any of the existing builds?

Not sure, I was doing it to be consistent. Based on your other comment it may be better to go for the non-root approach.

Issue #2874, opened 2019-07-14
nlohmann json is available as a package, no need to download it


yurivict:
See cmake/DownloadNlohmannJson.cmake

Repology shows that nlohmann-json is available on virtually all distros: https://repology.org/project/nlohmann-json/versions
There is no need to download it.

Downloads aren't allowed in the modern packaging systems due to security concerns, and such downloads only clutter the packaging process.

remyabel:
It's suggested in https://github.com/googleapis/google-cloud-cpp/issues/2802#issuecomment-510881426 that we use the same version of nlohmann_json that the customer has installed. In other words, it's planned that we will use `find_package` to discover it rather than downloading it.

amensel:
Hoping to draw some attention back to this one, especially in relation to #4094. 

GCC 10 has improved improved detection of incomplete types. This revealed a bug in `nlohmann::json`: https://github.com/nlohmann/json/issues/1920

Unfortunately, versions later than 3.4.0 (the version vendored into google-cloud-cpp) are no longer safe to use alongside 3.4.0. The header guards have changed and as has the template for the `basic_json` type. This creates a nasty little fight for which I still haven't found a clean resolution.

It would be great if `nlohmann::json` could be provided as a package like other dependencies in e.g. the Protobuf or gRPC builds.

If not that, it would at least be polite to not expose the touched up and customized version that this library uses in its public includes, despite whatever claims are made in the comments of `nljson.h` about it being "safe".

coryan:
Thanks for letting us know about the additional problems.  I have moved up this issue in our backlog, I do not think we will be able to fix it for the next release (due in less than a week).  I am hoping for the 2020-09 release.

amensel:
Awesome! Thanks for the speedy response.

coryan:
As you may have noticed we now have a PR that will fix this problem. But note that we have not upgraded to a newer version of nlohmann_json with #4747.  We are staying on v3.4.0 to keep the PR manageable.  The good news is that it seems to work, some of our builds use vcpkg and they are are using v3.8.0 of the nlohmann_json library, but I would like to run all builds with that version before declaring success.

amensel:
Ah, I had missed that; I'll pull it into our build in the interim while we wait for the official release.

Thanks!

coryan:
No problem, please let us know if you run into any trouble.

Issue #2802, opened 2019-06-21
Simplify CMake files to always use `find_package()`.


coryan:
This is a large task, probably should be broken down into smaller tasks.

I would like us to consider simplifying the CMake files to **always** use `find_package()` to find the dependencies. Then CI builds can pre-install all the dependencies. The users can do the same, or (at their choice), use a super build that installs all the dependencies for them.

We did this in googleapis/google-cloud-cpp-spanner#39 and it has been much easier to use, and I think it is easier to maintain.

remyabel:
What does this mean for the `readme/Dockerfile*`? Since they're testing grabbing the dependencies via external projects, does this mean we should remove the test readmes?

coryan:
We will add a super build (make in `ci/super/`, or simply `super/`), and change the README.md instructions (and the corresponding Dockerfiles) to use the super build.

I believe (but have not thought much about it, nor discussed it with my colleagues) the README structure will change to:

- Do you just want to build this quickly, run the examples and see if it works?
  - With Bazel: --> Link to the instructions on how to do that with Bazel.
  - With CMake --> Link to the instructions with a super build.
- Do you want to install the library in /usr/local and use them from there? --> link to the INSTALL.md file.
- Do you want to use the library as part of a larger Bazel project? --> link to a new document or README section.
- Do you want to use the library as part of a larger CMake project? --> link to a new document or README section on how to do that (currently this is hidden in the landing page for the reference docs).

remyabel:
Do we also want to change how we obtain nlohmann_json (build, install, `find_package`) or leave it alone? The logical place to install it would be `/usr/local/include`, but that might also be where a user has installed it. It wouldn't make much sense to `#include <nlohmann/json.hpp>` in `nljson_use_after_third_party_test.cc`.

coryan:
> Do we also want to change how we obtain nlohmann_json (build, install, `find_package`) or leave it alone?

Good question. I think we should aim to finding it via `find_package()` too.  We can take an intermediate step where we leave it alone if that is simpler to implement.

remyabel:
It should be easy to implement, I'm just asking what is the right thing to do here. We currently support the user including a potentially different version of the library, we also patch it to remove the user-defined literals. If we use `find_package` for it I think we need to install it to a non-standard location then set the search paths to that location.

coryan:
> If we use `find_package` for it I think we need to install it to a non-standard location then set the search paths to that location.

[...]

PR #2880, opened 2019-07-16
WIP: Use find_package for nlohmann_json

remyabel:
This begins the work of https://github.com/googleapis/google-cloud-cpp/issues/2802#issuecomment-510881426

Fixes https://github.com/googleapis/google-cloud-cpp/issues/2874

This PR, whenever possible, attempts to install nlohmann_json via the package manager. If it is unavailable, then we install it from source. Note that since different distros package different versions, then it might be a good idea to install a newer version from source. However, the disadvantage of this is that many package maintainers would prefer the packaged version, even if it is old, because it will contain patches and packaging standards relevant to that distro. On the other side of the argument, supporting a version as old as 2.1.1 would be absurd.

I have only tested one Dockerfile manually on my system; testing all of it would be time consuming. I'm relying on the first round of Kokoro builds to spot any errors. I apologize in advance.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2880)


coryan:
> I have only tested one Dockerfile manually on my system; testing all of it would be time consuming. I'm relying on the first round of Kokoro builds to spot any errors. I apologize in advance.

No need to apologize, that is why we have robots to do this work.  If you do get errors please test your fixes locally first though.

remyabel:
Hmm, this will be less straight forward than I thought. I'm running into strange compilation errors:

```
In file included from /home/build/google-cloud-cpp/google/cloud/storage/internal/nljson_use_after_third_party_test.cc:23:
/home/build/google-cloud-cpp/google/cloud/storage/internal/nljson.h:70:26: error: variable or field 'PrintTo' declared void
   70 | inline void PrintTo(json const& j, std::ostream* os) { *os << j.dump(); }
      |                          ^~~~~
/home/build/google-cloud-cpp/google/cloud/storage/internal/nljson.h:70:21: error: 'json' was not declared in this scope
   70 | inline void PrintTo(json const& j, std::ostream* os) { *os << j.dump(); }
      |                     ^~~~
/home/build/google-cloud-cpp/google/cloud/storage/internal/nljson.h:70:21: note: suggested alternatives:
In file included from /home/build/google-cloud-cpp/google/cloud/storage/internal/nljson_use_after_third_party_test.cc:20:
/usr/include/json.hpp:871:7: note:   'nlohmann::json'
  871 | using json = basic_json<>;
      |       ^~~~
In file included from /home/build/google-cloud-cpp/google/cloud/storage/internal/nljson_use_after_third_party_test.cc:20:
/usr/include/json.hpp:2259:29: note:   'nlohmann::detail::input_format_t::json'
 2259 | enum class input_format_t { json, cbor, msgpack, ubjson, bson };
      |                             ^~~~
In file included from /home/build/google-cloud-cpp/google/cloud/storage/internal/nljson_use_after_third_party_test.cc:23:
/home/build/google-cloud-cpp/google/cloud/storage/internal/nljson.h:70:48: error: expected primary-expression before '*' token
   70 | inline void PrintTo(json const& j, std::ostream* os) { *os << j.dump(); }
      |                                                ^
/home/build/google-cloud-cpp/google/cloud/storage/internal/nljson.h:70:50: error: 'os' was not declared in this scope; did you mean 'cos'?
   70 | inline void PrintTo(json const& j, std::ostream* os) { *os << j.dump(); }
      |                                                  ^~
      |                                                  cos
```

I think this might be a compatibility issue with 3.6.1.

remyabel:
I am closing this for now. There are a surprising amount of obstacles here:

```
Step 55/81 : RUN cmake -H. -Bcmake-out
 ---> Running in 55baf1f8a701
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.8 or higher is required.  You are running version 3.5.1
```

Issue #2877, opened 2019-07-15
Change bigtable/benchmarks to compile with exceptions disabled.


coryan:
For mostly historical reasons the benchmarks only compile with exceptions enabled.

remyabel:
For removing the exception from `BenchmarkSetup`, what is your preferred course of action? The most straight forward approach is to use a boolean and have a `is_setup()` flag. Something like:

```C++
  auto usage = [this, argv](char const* msg) {
    # ... snip ...
    std::cerr << msg;
    this->is_setup_ = false;
  };

  if (argc < 4) {
    usage("too few arguments for program.");
    return; # return instead of throwing an exception
  }

  # ... snip ...

  is_setup_ = true;
```

Then we would need to explicitly check if `benchmark_setup.is_setup()` before using it (otherwise we'll call it "undefined behavior", but I am open to more rigorous ways of handling it), something like:

```C++
BenchmarkSetup blah(...);
ASSERT_TRUE(blah.is_setup());
```

Unfortunately, this is causing the tests to fail, so I am open to other ideas.

coryan:
The common practice is to have a factory function that returns `StatusOr<BenchmarkSetup>` (or std::unique_ptr<BenchmarkSetup>`) and make the constructor private. Alternatively return `StatusOr<BenchmarkSetupData>` from the factory function, where `BenchmarkSetupData` is a plain struct consumed by the (public) `BenchmarkSetup` constructor.

Neither are awesome, for obvious reasons.

PR #2881, opened 2019-07-17
Partially change benchmarks to use StatusOr instead of exceptions.

remyabel:
Part of the work for #2877 

The only thing I haven't done yet is change `BenchmarkSetup` (see above issue for discussion). I prefer someone with more experience to tackle that.

The gist of this PR is that all try/catch blocks are replaced with reporting the error then returning. Instead of a `successful` flag `OperationResult` now contains a `google::cloud::Status`. This also means that the `op` functors return a `Status` as well.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2881)


remyabel:
I think I fixed all the cases where I was returning early instead of printing the error and continuing. PTAL.

PR #2882, opened 2019-07-17
Implement native IAM operations for GCS.

dopiera:
native representation.

The native operations use raw json underneath, so that even if the
protocol is exptended, fields aren't silently dropped.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2882)


dopiera:
This will be split.

Issue #1833, opened 2019-01-11
Implement IAM conditions for Google Cloud Storage.


coryan:
Check with @coryan or @frankyn for details.

coryan:
## Background

GCS and Cloud Bigtable will support IAM conditions:

https://cloud.google.com/iam/docs/conditions-overview

This will change the protos (and JSON objects) used to represent a IAM binding:

https://github.com/googleapis/googleapis/blob/0026f4b890ed9e2388fb0573c0727defa6f5b82e/google/iam/v1/policy.proto#L133

https://github.com/googleapis/googleapis/blob/master/google/type/expr.proto

The current API will not be able to support these.

## The Problem

The current API creates a map for all the bindings returned in a `SetIamPolicy` or `GetIamPolicy` request. This map is indexed by the `role` field in the binding. With the addition of the `condition` field the `role` is no longer a valid key: the same `role` may have different bindings with different conditions, they should not be merged together. This creates a series of problems:

- The old APIs, when receiving a new proto or JSON object would return an invalid representation of the data to the application.
- Existing applications, will perform data destructive changes if they perform the usual read-modify-update cycle using the new data.

## Proposal

We should remove the wrappers around a `Policy` and `Binding`. This will not be the last change to this type, we should design a C++ API that will grow well as these types change. We should implement new APIs that use simple representations for the Policy and Binding types. These should make no attempt to index the data by any key. They should return the sequence of values to the application, as they are received from the wire.

### Existing Functions and Types

The existing functions and types will signal an error if they receive data they cannot process accurately. That is a request for `GetIamPolicy` or `SetIamPolicy` returns a proto or JSON object that the old functions cannot represent accurately we will return an `StatusOr<>` with an error (`kInvalidArgument`).

The old APIs will continue to work if the application is only dealing with bindings without any conditions.

coryan:
/cc: @frankyn @jkwlui @crwilcox please see my proposal above for the IAM changes.

crwilcox:
@coryan , this seems a reasonable idea. We can either make it an array or a set with the value being a list.
ie:
```
[ 
(a, b),
(a,c),
(d, e),
(d, f),
]
```
- OR - 
```
[ 
(a, {b,c}),
(d, {e,f}),
]
```

I think I am going to side with the flat list though. The user could always elect to turn that to the map.

jkwlui:
+1! We are planning to implement the same fix across other languages as well.

PR #2886, opened 2019-07-17
Implement native IAM data structures for GCS.

dopiera:
native representation.

The native operations use raw json underneath, so that even if the
protocol is extended, fields aren't silently dropped.

This is needed for #1833.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2886)


google/cloud/storage/iam_policy.cc
```diff
@@ -219,7 +219,8 @@ struct NativeIamBinding::Impl {
         if (!status.ok()) {
           return status;
         }
-        members.emplace_back(member);
+        std::string member_str(member);
+        members.emplace_back(std::move(member_str));
```
remyabel: Can you do something like `member.get<std::string>()`?

PR #2887, opened 2019-07-17
cleanup: Fix shellcheck warning in upload-docs.sh

remyabel:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2887)


coryan:
Thanks! I lost count of how many times I told myself: "I should fix that".

PR #2897, opened 2019-07-20
Convert BenchmarkSetup to use StatusOr.

remyabel:
Fixes https://github.com/googleapis/google-cloud-cpp/issues/2877

As we do not want to use exceptions, we create a "factory" function named `MakeBenchmarkSetup` that does all of the actual work and returns a `StatusOr`. As suggested by @coryan, I moved all of the fields from `BenchmarkSetup` into `BenchmarkSetupData`, which is accepted by the former's public constructor.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2897)

### DIFF
--- a/google/cloud/bigtable/CMakeLists.txt
+++ b/google/cloud/bigtable/CMakeLists.txt
@@ -349,9 +349,9 @@ if (BUILD_TESTING)
     add_subdirectory(tests)
 endif ()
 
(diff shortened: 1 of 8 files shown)


remyabel:
> Why not just return setup_data?

That doesn't compile, even if I remove explicit. Did you have other changes in mind that would make this work?

PR #2898, opened 2019-07-20
feat: Add CMake super build. 

remyabel:
Thanks to @tmatsuo's work on cpp-cmakefiles this is now possible. Similar to google-cloud-cpp-spanner, this PR adds a super build. It piggybacks on the existing external project files.

Note about some specific changes here that differ from spanner:
- Unlike spanner where `RUN_INTEGRATION_TESTS` is set to "yes" for each build that should run those tests, we set it to "no" for builds that skip the tests. This is because (I think) the integration tests run by default for all of the builds.

- I changed the `INSTALL_DIR` for the external projects to `GOOGLE_CLOUD_CPP_EXTERNAL_PREFIX`. This allows us to reuse those files with minimal fuss.

TODO: Documentation still needs to be added.

Another TODO: some of the Dockerfiles do `COPY . /var/tmp/ci` which means modifying `ci/super` triggers a rebuild.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2898)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -137,11 +137,21 @@ _EOF_
 fi
 
 if [[ "${BUILD_TESTING:-}" = "yes" ]]; then
-  # Run the tests and output any failures.
-  echo
-  echo "${COLOR_YELLOW}Running unit tests $(date)${COLOR_RESET}"
-  echo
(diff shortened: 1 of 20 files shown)


remyabel:
I am adding the `DO NOT MERGE` because I am suffering from the same issue in https://github.com/googleapis/google-cloud-cpp-spanner/issues/249, even though `CMAKE_PREFIX_PATH` is already there. I am scratching my head trying to figure out what the issue is.

PR #2901, opened 2019-07-22
Disable ABI compatibility checks.

dopiera:
We're only interested in source compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2901)


dopiera:
FWIW, I tested it on #2900 and it reported only source compatibility problems.

PR #2916, opened 2019-07-26
feat: Change storage/benchmarks to compile with exceptions disabled.

### LINKED ISSUE #2878: Change storage/benchmarks to compile with exceptions disabled.
For mostly historical reasons the benchmarks only compile with exceptions enabled.

remyabel:
Fixes #2878

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2916)

### DIFF
--- a/google/cloud/storage/benchmarks/benchmark_utils.cc
+++ b/google/cloud/storage/benchmarks/benchmark_utils.cc
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #include "google/cloud/storage/benchmarks/benchmark_utils.h"
-#include "google/cloud/internal/throw_delegate.h"
 #include <cctype>
 #include <sstream>
 #include <stdexcept>
@@ -128,7 +127,8 @@ std::chrono::seconds ParseDuration(std::string const& val) {
   return std::chrono::seconds(s);
 }
 
-bool ParseBoolean(std::string const& val, bool default_value) {
+google::cloud::StatusOr<bool> ParseBoolean(std::string const& val,
+                                           bool default_value) {
   if (val.empty()) {
     return default_value;
   }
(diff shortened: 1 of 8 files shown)


google/cloud/storage/benchmarks/storage_latency_benchmark.cc
```diff
@@ -419,11 +428,11 @@ Options ParseArgs(int argc, char* argv[]) {
        }},
       {"--enable-connection-pool", "enable the client library connection pool",
        [&options](std::string const& val) {
-         options.enable_connection_pool = gcs_bm::ParseBoolean(val, true);
+         options.enable_connection_pool = *gcs_bm::ParseBoolean(val, true);
```
devjgm: Dereferencing a `StatusOr` that contains an error is technically UB. AFAICT, there's no guarantee that the parse will succeed here.

One alternative would be to have ParseBoolean() return an `optional<bool>`, and not take a default argument. That is
`optional<bool> ParseBoolean(std::string const&)`.

That way, a caller could invoke it like this:

```cc
options.enable_connection_pool = gcs_bm::ParseBoolean(val).value_or(true);
```

What do you think?

remyabel: It will require a little more work but I think the best solution here is to propagate the `Status`/`StatusOr`, regardless of how we to choose to represent the default argument.

remyabel: This is what I got so far and it has a few problems:

```C++
       [&options](std::string const& val) -> google::cloud::Status {
         auto status = gcs_bm::ParseBoolean(val, true);
         if (status && *status) {
           options.enable_xml_api = *status;
         }
         return status.status();
       }},
```

We're using `StatusOr<bool>` here, which you don't recommend obviously. Furthermore, there is a lot of boilerplate here.

I will go for your suggestion instead.

google/cloud/storage/benchmarks/benchmark_utils.cc
```diff
@@ -127,10 +127,9 @@ std::chrono::seconds ParseDuration(std::string const& val) {
   return std::chrono::seconds(s);
 }
 
-google::cloud::StatusOr<bool> ParseBoolean(std::string const& val,
-                                           bool default_value) {
+google::cloud::optional<bool> ParseBoolean(std::string const& val) {
   if (val.empty()) {
-    return default_value;
+    return google::cloud::optional<bool>{};
```
devjgm: Does `return {};` work? If so, it may be nice to use that here and at the end of the function.

PR #2917, opened 2019-07-26
cleanup: copy install scripts explicitly.

### LINKED ISSUE #2905: Avoid `COPY . /var/tmp/ci` or move the super build from `ci/super` to `super`
Currently anytime changes are made to `ci/super`, a rebuild will be triggered for any Dockerfile that has `COPY . /var/tmp/ci` in it. The simplest solution would be to move the directory.

Otherwise, if we copy the files explicitly (`install-bazel.sh`, etc) we'll have to make sure that nothing breaks due to build context changes.

remyabel:
Fixes #2905

This should minimize rebuilds when files in `ci` are modified.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2917)

### DIFF
--- a/ci/kokoro/docker/Dockerfile.centos
+++ b/ci/kokoro/docker/Dockerfile.centos
@@ -45,9 +45,12 @@ RUN pip install httpbin flask gevent gunicorn crc32c
 RUN ln -sf /usr/bin/cmake3 /usr/bin/cmake && \
     ln -sf /usr/bin/ctest3 /usr/bin/ctest
 
+# Copy over scripts explicitly to avoid cache busting.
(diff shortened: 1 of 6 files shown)


coryan:
I am not sure this necessarily helps.  I believe Docker tars the complete "path" argument (to create a context), creates a hash of the context, and rebuilds if the hash changes, even if you do not use most of the files in the context.

I think if we really care about this we will need to add the directories we want to ignore into the `.dockerignore` or change the directory hierarchy.

remyabel:
> and rebuilds if the hash changes

That's a good point. I confused the distinction between "minimize build time" and "rebuilds".

PR #2919, opened 2019-07-28
making partial errors/last_status available to `ObjectWriteStream`

### LINKED ISSUE #2827: Make partial errors accessible in ObjectWriteStream.
When uploading objects it would be nice to be able to check for errors immediately, with some details about what failed, for example:

```C++
auto stream = client.WriteObject(...);

stream.write(...);
if (stream.bad()) {
    std::cerr << "Error writing to stream... status=" << stream.last_status();
}
```

The `.last_status()` function does not exist today, of course. We need something like it.

manish-qlogic:
`ObjectWriteStreamBuf` now offers `last_status` and this status can be used by `ObjectWriteStream` to determine partial errors. 

This fixes #2827.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2919)

### DIFF
--- a/google/cloud/storage/internal/object_streambuf.cc
+++ b/google/cloud/storage/internal/object_streambuf.cc
@@ -201,7 +201,8 @@ ObjectWriteStreambuf::ObjectWriteStreambuf(
     : upload_session_(std::move(upload_session)),
       max_buffer_size_(UploadChunkRequest::RoundUpToQuantum(max_buffer_size)),
       hash_validator_(std::move(hash_validator)),
-      last_response_{HttpResponse{400, {}, {}}} {
+      last_response_{HttpResponse{400, {}, {}}},
+      last_status_(Status()) {
   current_ios_buffer_.reserve(max_buffer_size_);
   auto pbeg = &current_ios_buffer_[0];
   auto pend = pbeg + current_ios_buffer_.size();
@@ -296,7 +297,9 @@ StatusOr<HttpResponse> ObjectWriteStreambuf::FlushFinal() {
   StatusOr<ResumableUploadResponse> result =
       upload_session_->UploadFinalChunk(current_ios_buffer_, upload_size);
   if (!result) {
-    // This was an unrecoverable error, time to signal an error.
+    // This was an unrecoverable error, time to store status and signal an
+    // error.
+    last_status_ = result.status();
     return std::move(result).status();
   }
   // Reset the iostream put area with valid pointers, but empty.
@@ -309,6 +312,7 @@ StatusOr<HttpResponse> ObjectWriteStreambuf::FlushFinal() {
   // If `result.ok() == false` we never get to this point, so the last response
   // was actually successful. Represent that by a HTTP 200 status code.
   last_response_ = HttpResponse{200, std::move(result).value().payload, {}};
+  last_status_ = Status();
   return last_response_;
 }
 
@@ -332,7 +336,9 @@ StatusOr<HttpResponse> ObjectWriteStreambuf::Flush() {
   StatusOr<ResumableUploadResponse> result =
       upload_session_->UploadChunk(current_ios_buffer_);
   if (!result) {
-    // This was an unrecoverable error, time to signal an error.
+    // This was an unrecoverable error, time to store status and signal an
+    // error.
+    last_status_ = result.status();
     return std::move(result).status();
   }
   // Reset the put area, preserve any data not setn.
@@ -347,6 +353,7 @@ StatusOr<HttpResponse> ObjectWriteStreambuf::Flush() {
   // If `result.ok() == false` we never get to this point, so the last response
   // was actually successful. Represent that by a HTTP 200 status code.
   last_response_ = HttpResponse{200, std::move(result).value().payload, {}};
+  last_status_ = Status();
   return last_response_;
 }
 
--- a/google/cloud/storage/internal/object_streambuf.cc
+++ b/google/cloud/storage/internal/object_streambuf.cc
@@ -201,8 +201,7 @@ ObjectWriteStreambuf::ObjectWriteStreambuf(
     : upload_session_(std::move(upload_session)),
       max_buffer_size_(UploadChunkRequest::RoundUpToQuantum(max_buffer_size)),
       hash_validator_(std::move(hash_validator)),
-      last_response_{HttpResponse{400, {}, {}}},
-      last_status_(Status()) {
+      last_response_{HttpResponse{400, {}, {}}} {
   current_ios_buffer_.reserve(max_buffer_size_);
   auto pbeg = &current_ios_buffer_[0];
   auto pend = pbeg + current_ios_buffer_.size();
@@ -299,7 +298,7 @@ StatusOr<HttpResponse> ObjectWriteStreambuf::FlushFinal() {
   if (!result) {
     // This was an unrecoverable error, time to store status and signal an
     // error.
-    last_status_ = result.status();
+    last_response_ = result.status();
     return std::move(result).status();
   }
   // Reset the iostream put area with valid pointers, but empty.
@@ -312,7 +311,6 @@ StatusOr<HttpResponse> ObjectWriteStreambuf::FlushFinal() {
   // If `result.ok() == false` we never get to this point, so the last response
   // was actually successful. Represent that by a HTTP 200 status code.
   last_response_ = HttpResponse{200, std::move(result).value().payload, {}};
-  last_status_ = Status();
   return last_response_;
 }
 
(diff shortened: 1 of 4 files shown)


google/cloud/storage/internal/object_streambuf.h
```diff
@@ -146,6 +148,7 @@ class ObjectWriteStreambuf : public std::basic_streambuf<char> {
   HashValidator::Result hash_validator_result_;
 
   StatusOr<HttpResponse> last_response_;
+  Status last_status_;
```
devjgm: Is there some relationship between last_response_ and last_statsus_? Like, if last_status_ is OK, then last_response_ is some 200 status? Or if last_response_ is not a 2xx HTTP status, then last_status_ is not OK?

It looks like that might be the case, other than in the constructor where last_response_ is initialized to a 400 HTTP status code, and last_status_ is initialized with OK.

I'm not suggesting any change here, just trying to understand this better since I'm not too familiar with this code.

google/cloud/storage/internal/object_streambuf_test.cc
```diff
@@ -341,6 +341,95 @@ TEST(ObjectWriteStreambufTest, CreatedForFinalizedUpload) {
   EXPECT_EQ(close_result.value().payload, payload);
 }
 
+/// @test Verify that last error status is accessible for small payload.
+TEST(ObjectWriteStreambufTest, erroneousStream) {
```
devjgm: I think the casing of this test name should be `ErroneousStream`.

google/cloud/storage/internal/object_streambuf_test.cc
```diff
@@ -341,6 +341,95 @@ TEST(ObjectWriteStreambufTest, CreatedForFinalizedUpload) {
...
+  auto mock = google::cloud::internal::make_unique<
+      testing::MockResumableUploadSession>();
+  EXPECT_CALL(*mock, done).WillRepeatedly(Return(false));
+
+  auto const quantum = UploadChunkRequest::kChunkSizeQuantum;
+  std::string const payload = "small test payload";
+
+  int count = 0;
+  EXPECT_CALL(*mock, UploadFinalChunk(_, _))
+      .WillOnce(Invoke([&](std::string const& p, std::uint64_t s) {
```
devjgm: `s` as the name for an integer may mislead people into thinking it's a string. How about `n`?

google/cloud/storage/internal/object_streambuf_test.cc
```diff
@@ -341,6 +341,95 @@ TEST(ObjectWriteStreambufTest, CreatedForFinalizedUpload) {
...
+  auto response = streambuf.Close();
+
+  EXPECT_EQ(StatusCode::kInvalidArgument, response.status().code())
+      << ", status=" << response.status();
+  EXPECT_EQ(StatusCode::kInvalidArgument, streambuf.last_status().code())
+      << ", status=" << streambuf.last_status();
+}
+
+/// @test Verify that last error status is accessible for large payloads.
+TEST(ObjectWriteStreambufTest, errorInLargePayload) {
```
devjgm: Casing on the test name here too.

google/cloud/storage/object_stream.h
```diff
@@ -343,6 +343,15 @@ class ObjectWriteStream : public std::basic_ostream<char> {
...
   void Suspend() &&;
 
+  /**
+   * Returns the status of partial errors.
+   *
+   * Application may write multiple times before closing the stream, this
+   * function gives the capability to find out status even before stream
+   * closure.
+   */
+  Status const& last_status() const { return buf_->last_status(); }
```
remyabel: Since this is part of the public API, `last_status()` may be confused to have a similar role to `ObjectReadStream`'s `status()`. Maybe this should document that `last_status()` has a relationship to `last_response` and signals different errors not already covered by `metadata()`.

google/cloud/storage/internal/object_streambuf.cc
```diff
@@ -201,7 +201,8 @@ ObjectWriteStreambuf::ObjectWriteStreambuf(
     : upload_session_(std::move(upload_session)),
       max_buffer_size_(UploadChunkRequest::RoundUpToQuantum(max_buffer_size)),
       hash_validator_(std::move(hash_validator)),
-      last_response_{HttpResponse{400, {}, {}}} {
+      last_response_{HttpResponse{400, {}, {}}},
+      last_status_(Status()) {
```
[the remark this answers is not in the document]
manish-qlogic: fixed.

google/cloud/storage/internal/object_streambuf_test.cc
manish-qlogic: fixed.

google/cloud/storage/object_stream.h
remyabel: Alternatively, if `last_status()` overlaps with `metadata().status()` maybe this can just be renamed to `status()`?

google/cloud/storage/internal/object_streambuf_test.cc
manish-qlogic: Changed to n.

google/cloud/storage/internal/object_streambuf_test.cc
manish-qlogic: fixed.

google/cloud/storage/object_stream.h
manish-qlogic: AFAIK, `metadata()` and `last_response` shows the final status where as `last_status()` can be used to find intermediate `UploadChunk` status or partial errors. Let me update comments for `last_status()`

google/cloud/storage/internal/object_streambuf.cc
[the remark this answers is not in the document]
manish-qlogic: `Kokoro ClangTidy` did not like this change, reverting it.

google/cloud/storage/object_stream.h
remyabel: Thanks, makes sense. What do you think about renaming it to `latest_status` or something along those lines? But I'll defer to @coryan for that.

Nit: s/then/than/

google/cloud/storage/internal/object_streambuf.h
coryan: I think this is the key question.

For background, the application usually uses a `gcs::ObjectWriteStream`, and it performs error handling using IOStream errors, for example:

```C++
  gcs::ObjectWriteStream stream = client.WriteObject(....);
  ...
  ..
  ..
  .
  if (!stream.good()) {
   std::cerr << "Error, the last status was = " << stream.last_status(); // TODO: this is the point of this PR.
  }
```

I think we want that error checking code to always work, regardless of the previous operations (`.write()`, `.Close()`, `operator<<`, whatever). I think what we should do is change the code to be something like:

```
Status last_status() const {
  if (last_response_) {
    return Status();
  }
  return last_response_.status();
}
```

And always set `last_reponse_` when there is an error.

google/cloud/storage/internal/object_streambuf.cc
```diff
@@ -332,7 +336,9 @@ StatusOr<HttpResponse> ObjectWriteStreambuf::Flush() {
   StatusOr<ResumableUploadResponse> result =
       upload_session_->UploadChunk(current_ios_buffer_);
   if (!result) {
-    // This was an unrecoverable error, time to signal an error.
+    // This was an unrecoverable error, time to store status and signal an
+    // error.
+    last_status_ = result.status();
```
coryan: See below, I think this should be `last_response_ = result.status()`.

google/cloud/storage/internal/object_streambuf.cc
```diff
@@ -347,6 +353,7 @@ StatusOr<HttpResponse> ObjectWriteStreambuf::Flush() {
   // If `result.ok() == false` we never get to this point, so the last response
   // was actually successful. Represent that by a HTTP 200 status code.
   last_response_ = HttpResponse{200, std::move(result).value().payload, {}};
+  last_status_ = Status();
```
coryan: and then we can remove this.

google/cloud/storage/object_stream.h
```diff
@@ -343,6 +343,18 @@ class ObjectWriteStream : public std::basic_ostream<char> {
...
+   * Returns the status of partial errors.
+   *
+   * Application may write multiple times before closing the stream, this
+   * function gives the capability to find out status even before stream
+   * closure.
+   *
+   * This function is different than `metadata()` as calling `metadata()`
+   * before Close() is undefined.
+   */
+  Status const& last_status() const { return buf_->last_status(); }
```
coryan: Probably will need to return by value here.

manish-qlogic: fixed.

google/cloud/storage/internal/object_streambuf.cc
manish-qlogic: removed.

google/cloud/storage/internal/object_streambuf.cc
manish-qlogic: done.

google/cloud/storage/internal/object_streambuf.h
manish-qlogic: done.

Issue #2928, opened 2019-07-31
Re-evaluate where we define our Dockerfile build-args


remyabel:
Intuitively, defining our build-args like this:

```Dockerfile
ARG DISTRO_VERSION=7
ARG NCPU=4
FROM centos:${DISTRO_VERSION} AS devtools
```

in a multi-stage build should work as expected. But it doesn't. `NCPU` will be empty in the second stage:

```Dockerfile
Step 14/14 : RUN cmake --build cmake-out -- -j ${NCPU}
 ---> Running in 1d5b2ee2b116
/usr/bin/ninja-build: option requires an argument -- 'j'
usage: ninja-build [options] [targets...]

if targets are unspecified, builds the 'default' target (see manual)
```

The following does **not** work:

```Dockerfile
ARG NCPU=4
FROM devtools AS readme
```

It needs to be after:

```Dockerfile
FROM devtools AS readme
ARG NCPU=4
```

See https://github.com/moby/moby/issues/34129 for discussion.

remyabel:
FTR, using Ninja caught this bug. Using `make -j` without an argument puts no limits on the amount of parallel jobs, whereas it appears that Ninja will complain. I'll do a more thorough sweep to see where else NCPU might be misused.

coryan:
Fixed by #2929.

remyabel:
It's not fixed yet. I haven't made the PR with the NCPU changes yet.

PR #2929, opened 2019-07-31
feat: Enable Ninja for the README builds

### LINKED ISSUE #2891: Create a Linux build with CMake+Ninja.
Ninja is popular for CI builds, and for Windows builds.  In some corner cases a CMake file will work with Make, but will fail with Ninja. Mostly this is because Ninja is strict about byproducts of `ExternalProject_Add()` and some other custom rules.

Having a build with CMake+Ninja on Linux will detect these problems, and it would be easier to run and reproduce for most of the developers in the project.

remyabel:
Fixes #2891 and #2928

I also changed where NCPU is defined to ensure that it stays in scope for each
stage.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2929)

### DIFF
--- a/ci/kokoro/readme/Dockerfile.centos
+++ b/ci/kokoro/readme/Dockerfile.centos
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 ARG DISTRO_VERSION=7
-ARG NCPU=4
 FROM centos:${DISTRO_VERSION} AS devtools
+ARG NCPU=4
 
 # Please keep the formatting in these commands, it is optimized to cut & paste
 # into the README.md file.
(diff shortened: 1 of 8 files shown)


coryan:
Hmm... I did not realize that changing the super builds would change the `ci/kokoro/readme/*` builds.  Those are used to generate the instructions in the README.md file, and I would prefer to keep those minimal, and avoid introducing more tools than necessary.

I think we should do the CMake+Ninja build separately.

remyabel:
I can move installing `ninja-build` to the `START IGNORED` section.

remyabel:
To keep things simple I'll make it a separate build and add the NCPU changes in a different PR.

PR #2932, opened 2019-08-03
testing: Fix build on windows 2019

tmatsuo:
I need to use the full path of cmd in order to run 7-zip in both images. Probably because some difference between cygwin on windows 2016 image and msys2 on windows 2019 image.

Do we want to keep builds on windows 2016? If not, I'll eventually remove files for windows 2016 builds.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2932)


tmatsuo:
Using full path didn't work on windows 2019. Closing for now.

tmatsuo:
It works when setting Path env var in the powershell script. Reopening

Issue #666, opened 2018-06-06
Create one Bazel build on Windows


coryan:

tmatsuo:
I'm testing Bazel build with change in #2937, but it still fails with:

```
ERROR: C:/users/kbuilder/_bazel_kbuilder/vxekc77j/external/com_github_curl_curl/BUILD.bazel:52:1: C++ compilation of rul
e '@com_github_curl_curl//:curl' failed (Exit 2)
external/com_github_curl_curl/lib/vtls/schannel.c(41): fatal error C1083: Cannot open include file: 'curl_setup.h': No s
uch file or directory
```

tmatsuo:
Probably this line:
https://github.com/googleapis/google-cloud-cpp/blob/d37faac72bbff1d5842366953c6c320b2e88d3d2/bazel/curl.BUILD#L29

The path is different from the non-windows case:
https://github.com/googleapis/google-cloud-cpp/blob/d37faac72bbff1d5842366953c6c320b2e88d3d2/bazel/curl.BUILD#L280

tmatsuo:
Almost there, but it still fails because BoringSSL header file conflicts with windows.h. The fix is on the way.

tmatsuo:
There is one last thing `-lpthread`.

I'll try installing [winpthread pacman package](https://packages.msys2.org/package/mingw-w64-x86_64-winpthreads) and link against it.

remyabel:
That's for MinGW isn't it? Why not install pthreads from vcpkg?

coryan:
I think we should not add `-lpthread` when linking on Windows. We need one of those `select()` statements in Bazel to only use that option on Linux and similar operating systems.

tmatsuo:
Ah ok, sure, I'll just make it conditional.

tmatsuo:
Still seeing few more errors for storage tests.

```
curl.lib(schannel.obj) : error LNK2019: unresolved external symbol verify_certificate referenced in function schannel_connect_step2
curl.lib(schannel.obj) : error LNK2019: unresolved external symbol Curl_parseX509 referenced in function pkp_pin_peer_pubkey
curl.lib(schannel.obj) : error LNK2019: unresolved external symbol Curl_extract_certinfo referenced in function schannel_connect_step3
```

A typical link command:
```
[...]

PR #2940, opened 2019-08-06
fix: Fix bazel build on windows.

tmatsuo:
feat: Add config files for windows bazel builds.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2940)


tmatsuo:
Needs more work. Closing for now

tmatsuo:
Probably it will build on Windows with bazel. I'll manually build it tomorrow.

google/cloud/BUILD
```diff
@@ -55,8 +55,11 @@ load(
 [cc_test(
     name = "google_cloud_cpp_common_" + test.replace("/", "_").replace(".cc", ""),
     srcs = [test],
-    # TODO(#664 / #666) - use the right condition when porting Bazel builds
-    linkopts = ["-lpthread"],
+    # TODO(#664) - use the right condition when porting Bazel builds
+    linkopts = select({
+        "@bazel_tools//src/conditions:linux_x86_64": ["-lpthread"],
```
remyabel: What do you think about inverting the conditions, since `pthread` is used on more than just Linux? Using platform (untested), something like:

```
":windows": [""],
"//conditions:default": ["-lpthread"],
```

Note I suggest using the platform constraint here since it is more generic.

tmatsuo: I inverted the condition.

"@bazel_tools//src/conditions:windows" is defined at:
https://github.com/bazelbuild/bazel/blob/ec7084d91ae2ac20102729e13d6f1e2ad442f81d/src/conditions/BUILD.tools#L31

If I use `":windows":` in select, I needed to have the definition of it in the same file. I think we can just use "@bazel_tools//src/conditions:". WDYT?

remyabel: Sorry, I didn't mean `:windows` directly. I meant something like `@platforms//os:windows` per the [docs](https://docs.bazel.build/versions/master/platforms.html). I am not familiar enough with Bazel to know if this is more generic than `@bazel_tools//src/conditions:windows`.

tmatsuo: Probably we can just use `@bazel_tools//src/conditions`. Anyways thanks for the suggestion. With the inverted condition, it's slightly better now.

tmatsuo:
Finally bazel build passes on Windows 2019 :)
Hooray!

```
Executed 0 out of 155 tests: 155 tests pass.
INFO: Build completed successfully, 1 total action
```

bazel/curl.BUILD
```diff
@@ -46,7 +46,9 @@ CURL_WIN_SRCS = [
     "lib/inet_ntop.c",
     "lib/system_win32.c",
     "lib/vtls/schannel.c",
+    "lib/vtls/schannel_verify.c",
     "lib/idn_win32.c",
+    "lib/x509asn1.c",
```
tmatsuo: @gunan JFYI, I needed to add 2 files to our curl.BUILD which is originally copied from tensorflow repo.
We had errors described in: https://github.com/googleapis/google-cloud-cpp/issues/666#issuecomment-518859170

tmatsuo:
I think it's ready now. PTAL

PR #2942, opened 2019-08-07
Favor small sizes in storage throughput benchmark.

dopiera:
When picking a random size for a buffer from the range 128K and 64M a
uniform distribution is prohibitively wasteful.   It's unlikely that one
will need to decide whether 48M or 48.2M is better, but quite likely
need to decide between 128K and 256K. This PR introduces a distribution
which will produce twice as many samples between 0 and 1 than between 1
and 2 and so on.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2942)


coryan:
Per our IM discussion, I am going to close this PR. We can reopen it if we change our minds.

Siyler1 (approved):

PR #2943, opened 2019-08-07
Random CRC and MD5 in storage throughput benchmark

dopiera:
This is in order to be able to compare apples to apples. Results are
very unstable and the way to circumvent it is to reuse the same client.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2943)


dopiera:
> It looks good to me. Can you remind me if the checksum and hash settings are dumped in the benchmark results?
This PR introduces this.

> You need to resolve the conflicts.
Done

PR #2954, opened 2019-08-09
cleanup: Don't include nljson.h from public oauth2 headers.

### LINKED ISSUE #2772: Do not include nlohmann/json.hpp from any public header.

remyabel:
Fixes #2772

Per @coryan's suggestion, this approach factors out any logic w.r.t JSON into
separate functions. For the functions dealing with JWT assertions, since they
were not only private functions but also essentially standalone, they were
moved out of the `ServiceAccountCredentials` class and made to accept a
`time_point` parameter. Functions which would normally take just a payload take
a `HttpResponse` to preserve the status. Finally, a `ServiceAccountMetadata`
struct was added; `ComputeEngineCredentials` was not modified to accept this to
prevent API breakage.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2954)

### DIFF
--- a/google/cloud/storage/CMakeLists.txt
+++ b/google/cloud/storage/CMakeLists.txt
@@ -189,6 +189,7 @@ add_library(storage_client
             oauth2/authorized_user_credentials.h
             oauth2/authorized_user_credentials.cc
             oauth2/compute_engine_credentials.h
+            oauth2/compute_engine_credentials.cc
             oauth2/credential_constants.h
             oauth2/credentials.h
             oauth2/credentials.cc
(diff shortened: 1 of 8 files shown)


remyabel:
> Does this function need to be in the header? It is only used in one place, and that is a private member function of the class. We could move the declaration to the .cc file (and I think save the .dump() calls).

I agree. But then we have to factor this out into a separate function too:

https://github.com/googleapis/google-cloud-cpp/blob/825c3a6a985b55a0b5c33914f1fa4ed784332ea0/google/cloud/storage/oauth2/service_account_credentials.h#L215-L219

Which would also be a one-use function and probably not worth it I think. Happy to change it if you disagree.

PR #2958, opened 2019-08-09
Tidy cleanup

devjgm:
This change copies the .clang-tidy file from https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/.clang-tidy

The only difference is that for this project we disable the `google-runtime-int` check, which suggests avoiding types like `long` and `long long` in favor of explicitly sized types like `std::int64_t`. However, it appears that the curl library requires that are specifically `long` or `long long`, so keeping this warning would require lots of `NOLINT`, which simply doesn't seem worth it.

FWIW, I've always found this particular warning to be of dubious value anyway, so I don't mind turning it off.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2958)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,27 +2,27 @@
 # Configure clang-tidy for this project.
 
 # Disabled:
-#  -google-readability-namespace-comments the BIGTABLE_CLIENT_NS is a macro, and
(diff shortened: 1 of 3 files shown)


devjgm:
PTAL. I disabled the `google-runtime-int` check, and reverted all those changes.

PR #2959, opened 2019-08-10
Revert "bug: Fix nlohmann json macro leaks. (#2774)"

remyabel:
This reverts commit ef9d062721d7a6522c5f6e00fe1f93ad8b700c12.

These changes should no longer be necessary since `nljson.h` is no longer
included in any public header.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2959)

### DIFF
--- a/google/cloud/storage/internal/nljson.h
+++ b/google/cloud/storage/internal/nljson.h
@@ -46,14 +46,8 @@
 // nlohmann::json. This is safe because google/cloud/storage always includes
 // the nlohmann::json through this header, so after the first time our own
 // include guards are enough.
-#undef NLOHMANN_BASIC_JSON_TPL
-#undef NLOHMANN_BASIC_JSON_TPL_DECLARATION
 #undef NLOHMANN_JSON_HPP
 #undef NLOHMANN_JSON_FWD_HPP
-#undef NLOHMANN_JSON_SERIALIZE_ENUM
-#undef NLOHMANN_JSON_VERSION_MAJOR
-#undef NLOHMANN_JSON_VERSION_MINOR
-#undef NLOHMANN_JSON_VERSION_PATCH
 
 namespace nlohmann {
 //


remyabel:
> If we think nljson.h is no longer pulled by the public headers I think we should change google/cloud/storage/storage_include_test.cpp to test for this. Thoughts?

Thanks for the suggestion. It looks like `google/cloud/storage/iam_policy.h`, which includes `nljson.h`, was added in a commit after my initial comb-through, which is why I didn't catch it. I believe the following test should not fail:

```
// Note: pick a macro that isn't undefined by macro_unscope
#ifdef NLOHMANN_JSON_SERIALIZE_ENUM
#error "NLOHMANN_JSON shold not be included by storage public headers"
#endif // NLOHMANN_JSON_SERIALIZE_ENUM
```

remyabel:
Note: I am probably missing something obvious, but I'm unsure of how to change this test. The test only includes public headers, but those can include internal headers which in turn can include `nljson.h` for example:

```
# 1 "/v/google/cloud/storage/internal/nljson.h" 1
In file included from /v/google/cloud/storage/internal/metadata_parser.h:18:0,
                 from /v/google/cloud/storage/internal/common_metadata.h:20,
                 from /v/google/cloud/storage/internal/access_control_common.h:20,
                 from /v/google/cloud/storage/bucket_access_control.h:19,
                 from /v/google/cloud/storage/bucket_metadata.h:19,
                 from /v/google/cloud/storage/internal/raw_client.h:20,
                 from /v/google/cloud/storage/internal/logging_client.h:18,
                 from /v/google/cloud/storage/client.h:23,
                 from google/cloud/storage/tests/storage_include_test.cc:15:
 #include "google/cloud/storage/internal/nlohmann_json.hpp"
```

coryan:
> The test only includes public headers,

That seems right.

> but those can include internal headers which in turn can include `nljson.h` for example:

Which is a bug, or at least something we want to avoid.  We need to find each one of these indirect includes and clean them up too.

PR #2963, opened 2019-08-12
chore: Keep release tags in master branch.

coryan:
Change the code to not require changes for a new release branch, and
update the instructions to create tags in the `master` branch.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2963)


google/cloud/CMakeLists.txt
```diff
@@ -14,11 +14,6 @@
 # limitations under the License.
 # ~~~
 
-# Set this variable to a true value for branches that are releases. That will
-# remove the "+dev" suffix (technically "build metadata" according to
-# semver.org). We only want that suffix on development branches.
-set(GOOGLE_CLOUD_CPP_IS_RELEASE "")
```
devjgm: Should similar changes be made to the spanner project? If so, would you mind filing an issue about this and linking to this PR so we have something to copy?

coryan: Done in googleapis/google-cloud-cpp-spanner#339.

PR #2967, opened 2019-08-13
cleanup: Use only find_package to find dependencies.

### LINKED ISSUE #2802: Simplify CMake files to always use `find_package()`.

coryan:
This PR changes the CMake files to use `find_package()` exclusively when
looking for dependencies. To automatically download the dependencies the
project can be embedded in a larger "super build".

There are still some places where we use `find_package()` through a helper
file, because different versions of the CMake module (or config file) behave
differently and we want to manage those differences in one place.

This fixes #2802

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2967)


coryan:
/cc: @tmatsuo @remyabel because you are interested in this type of change.

INSTALL.md
```diff
@@ -412,15 +443,31 @@ sudo cmake --build cmake-out --target install -- -j ${NCPU:-4}
 sudo ldconfig
 ```
 
+#### googletest
+
+We need a revent version of GoogleTest to compile the unit and integration
```
devjgm: Typo: recent

INSTALL.md
```diff
@@ -510,15 +557,31 @@ sudo cmake --build cmake-out --target install -- -j ${NCPU:-4}
...
```
devjgm: typo: recent

INSTALL.md
```diff
@@ -632,15 +695,31 @@ sudo cmake --build cmake-out --target install -- -j ${NCPU:-4}
...
```
devjgm: typo: recent, and elsewhere...

cmake/FindProtobufWithTargets.cmake
```diff
@@ -0,0 +1,202 @@
...
+``protobuf::libprotobuf-lite``
+  The protobuf lite library.
+``protobuf::libprotoc``
+  The protoc library.
+``protobuf::protoc``
+  The protoc compiler.
+
+Example:
+
+.. code-block:: cmake
```
devjgm: I'm not sure I understand the two dots here, but maybe just take a look at this to make sure it's how you want it.

INSTALL.md
coryan: Fixed here and elsewhere.

cmake/FindProtobufWithTargets.cmake
coryan: This matches (as far as I can tell) the standard CMake documentation style for modules:

https://gitlab.kitware.com/cmake/cmake/blob/master/Help/dev/documentation.rst

We don't test or generate documentation from these, but I prefer to follow their style instead of inventing a new one.

remyabel: That's restructured text syntax. `.. code-block:: cmake` gets translated to syntax highlighted HTML. While it's fine to be consistent, it's worth pointing out that it doesn't mean anything for this codebase.

coryan:
@remyabel @googleapis/yoshi-cpp if you are doing development for `google-cloud-cpp` note that the build instructions changed. Please check doc/setup-cmake-environment.md.

cmake/FindProtobufWithTargets.cmake
```diff
@@ -0,0 +1,202 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ~~~
+
+#[=======================================================================[.rst:
+FindProtobufWithTargets
+-------------------
+
+A module to use `Protobuf` with less complications.
```
remyabel: Nit: since this is following RST syntax, it should use two backticks.

cmake/FindProtobufWithTargets.cmake
```diff
@@ -0,0 +1,202 @@
...
+module that always produces the same outputs:
+
+- It always generates the ``protobuf::*`` targets.
+- It always defines ``ProtobufWithTargets_FOUND`` and
+  ``ProtobufWithTargets_VERSION``.
+- It *prefers* using the CMake config files if they are available.
+- It fallsback on the ``FindProtobuf`` module if the config files are not found.
+- It populates any missing targets and their properties.
+
+The following :prop_tgt:`IMPORTED` targets are defined:
```
remyabel: Nit: `:prop_tgt` is a Sphinx directive. We probably don't care about it, we can just do IMPORTED with two backticks.

google/cloud/storage/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ if (TARGET storage-docs AND TARGET cloud-docs)
 endif ()
 
 include(IncludeNlohmannJson)
```
flokli: @coryan Can we use the cmake files provided by the nlohmann json package, instead of fetching it manually via cmake?

```
$out/include
$out/include/nlohmann
$out/include/nlohmann/json.hpp
$out/lib
$out/lib/cmake
$out/lib/cmake/nlohmann_json
$out/lib/cmake/nlohmann_json/nlohmann_jsonConfig.cmake
$out/lib/cmake/nlohmann_json/nlohmann_jsonConfigVersion.cmake
$out/lib/cmake/nlohmann_json/nlohmann_jsonTargets.cmake

```

coryan: See #2874.  We have not had time to do that, but yes, it makes sense.

flokli: Thanks, subscribed to that issue.

PR #2968, opened 2019-08-13
fix: use MetadataUpdatePolicy::FromClusterId.

### LINKED ISSUE #2704: bigtable::MetadataUpdatePolicy for ClusterId only used in tests?
This is just a code smell, but could be a bug in that we are passing the wrong resource header to the service.

dopiera:
This is a part of #2704.

These are the only place where these ctors can be used (I checked by making the `  MetadataUpdatePolicy(std::string const& resource_name, MetadataParamTypes const& metadata_param_type)` ctor private temporarily.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2968)

### DIFF
--- a/google/cloud/bigtable/instance_admin.cc
+++ b/google/cloud/bigtable/instance_admin.cc
@@ -508,8 +508,8 @@ Status InstanceAdmin::DeleteCluster(std::string const& instance_id,
   btadmin::DeleteClusterRequest request;
   request.set_name(ClusterName(instance_id, cluster_id));
 
-  MetadataUpdatePolicy metadata_update_policy(
-      ClusterName(instance_id, cluster_id), MetadataParamTypes::NAME);
+  auto metadata_update_policy = MetadataUpdatePolicy::FromClusterId(
+      instance_id, MetadataParamTypes::NAME, cluster_id);
 
   // This API is not idempotent, lets call it without retry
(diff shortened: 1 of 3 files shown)


coryan:
While this improves things, I believe it does not fix all the problems (though maybe we should open separate bugs).  My reading is that we should be setting the [routing headers](https://aip.dev/client-libraries/4222) with the cluster id whenever `clusters/*` is mentioned in the proto file, for example here:

https://github.com/googleapis/googleapis/blob/3406d1e899f1f41123b3fa9210ad4bef25c9a720/google/bigtable/admin/v2/bigtable_instance_admin.proto#L113

or here:

https://github.com/googleapis/googleapis/blob/3406d1e899f1f41123b3fa9210ad4bef25c9a720/google/bigtable/admin/v2/bigtable_instance_admin.proto#L99

which this change does not cover.  Can we fix those too?

coryan:
@dopiera do you want to merge this as-is, with more changes to follow, update this PR, or create a new PR with all the ClusterId changes?

dopiera:
I'm really close to fixing all the invalid headers, but it's over 500 loc,
so maybe a separate PR is a good idea.

pon., 19 sie 2019, 6:05 PM użytkownik coryan <notifications@github.com>
napisał:

PR #2979, opened 2019-08-16
bug: Account for when CreateHashValidator's options are set to false

### LINKED ISSUE #2978: DisableMD5Hash(false) Yields Surprising Results
`DisableMD5Hash` and `DisableCrc32cChecksum` are both ComplexOptions containing bools.  This would seem to imply the could be set to true or false.  Right now only their existence is checked, not their values.  I'm pretty sure this means both values result in the same behavior or disabling the hash checks.  This is surprising and I suspect it's a bug.

https://github.com/googleapis/google-cloud-cpp/blob/de6d1b5ddc5472966fe82f54db8a15ca5cde4798/google/cloud/storage/internal/hash_validator.cc#L82

yes

- Cloud Storage.

0.11.0

remyabel:
Fixes #2978

Previously, the `CreateHashValidator` functions did not check the value of the
options, only the existence of the options. As a consequence, whether the user
specifies `true` or `false` doesn't matter. This PR checks the values of both
options and adds the logic to both overloads. One hitch (to preserve existing
behavior) is that the options are mutually exclusive IFF only one option is
specified.  It's possible that this approach can be simplified, but boolean
logic is hard.

cc: @Jseph

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2979)

### DIFF
--- a/google/cloud/storage/internal/hash_validator.cc
+++ b/google/cloud/storage/internal/hash_validator.cc
@@ -70,17 +70,43 @@ std::unique_ptr<HashValidator> CreateHashValidator(bool disable_md5,
 
 std::unique_ptr<HashValidator> CreateHashValidator(
     ReadObjectRangeRequest const& request) {
+  auto disable_md5 = false;
+  auto disable_crc32c = false;
   if (request.RequiresRangeHeader()) {
     return google::cloud::internal::make_unique<NullHashValidator>();
   }
-  return CreateHashValidator(request.HasOption<DisableMD5Hash>(),
-                             request.HasOption<DisableCrc32cChecksum>());
+  if (request.HasOption<DisableMD5Hash>()) {
+    disable_md5 = request.GetOption<DisableMD5Hash>().value();
+    if (!request.HasOption<DisableCrc32cChecksum>()) {
+      disable_crc32c = !disable_md5;
+    }
+  }
+  if (request.HasOption<DisableCrc32cChecksum>()) {
+    disable_crc32c = request.GetOption<DisableCrc32cChecksum>().value();
+    if (!request.HasOption<DisableMD5Hash>()) {
+      disable_md5 = !disable_crc32c;
+    }
+  }
+  return CreateHashValidator(disable_md5, disable_crc32c);
 }
 
 std::unique_ptr<HashValidator> CreateHashValidator(
     ResumableUploadRequest const& request) {
-  return CreateHashValidator(request.HasOption<DisableMD5Hash>(),
-                             request.HasOption<DisableCrc32cChecksum>());
+  auto disable_md5 = false;
+  auto disable_crc32c = false;
+  if (request.HasOption<DisableMD5Hash>()) {
+    disable_md5 = request.GetOption<DisableMD5Hash>().value();
+    if (!request.HasOption<DisableCrc32cChecksum>()) {
+      disable_crc32c = !disable_md5;
+    }
+  }
+  if (request.HasOption<DisableCrc32cChecksum>()) {
+    disable_crc32c = request.GetOption<DisableCrc32cChecksum>().value();
+    if (!request.HasOption<DisableMD5Hash>()) {
+      disable_md5 = !disable_crc32c;
+    }
+  }
+  return CreateHashValidator(disable_md5, disable_crc32c);
 }
 
 }  // namespace internal
--- a/google/cloud/storage/internal/hash_validator.h
+++ b/google/cloud/storage/internal/hash_validator.h
@@ -112,10 +112,10 @@ class ResumableUploadRequest;
  * - If neither `DisableMD5Hash(true)` nor `DisableCrc32cChecksum(true)` are
  *   provided, then both MD5Hash and Crc32Checksum are used.
  * - If only `DisableMD5Hash(true)` is provided, then only Crc32cChecksum is
- * used.
+ *   used.
  * - If only `DisableCrc32c(true)` is provided, then only MD5Hash is used.
  * - If both `DisableMD5Hash(true)` and `DisableCrc32cChecksum(true)` are
- * provided, then neither are used.
+ *   provided, then neither are used.
  *
  * Specifying the option with `false` or no argument (default constructor) has
  * the same effect as not passing the option at all.
--- a/google/cloud/storage/internal/hash_validator.h
+++ b/google/cloud/storage/internal/hash_validator.h
@@ -102,6 +102,24 @@ class CompositeValidator : public HashValidator {
 class ReadObjectRangeRequest;
 class ResumableUploadRequest;
 
+/**
+ * @{
+ * The requests accepted by `CreateHashValidator` can be configured with the
+ * `DisableMD5Hash` and `DisableCrc32Checksum` options. You must explicitly
+ * pass these options to disable each respectively.
+ *
+ * The valid combinations are:
+ * - If neither `DisableMD5Hash(true)` nor `DisableCrc32cChecksum(true)` are
+ *   provided, then both MD5Hash and Crc32Checksum are used.
+ * - If only `DisableMD5Hash(true)` is provided, then only Crc32cChecksum is
+ * used.
+ * - If only `DisableCrc32c(true)` is provided, then only MD5Hash is used.
+ * - If both `DisableMD5Hash(true)` and `DisableCrc32cChecksum(true)` are
+ * provided, then neither are used.
+ *
+ * Specifying the option with `false` or no argument (default constructor) has
+ * the same effect as not passing the option at all.
+ */
 /// Create a hash validator configured by @p request.
 std::unique_ptr<HashValidator> CreateHashValidator(
     ReadObjectRangeRequest const& request);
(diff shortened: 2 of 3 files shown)


remyabel:
> Can you say why do we need this? Why do you think the options were mutually exclusive before and/or need to be now?

There are a couple of reasons. In `CreateHashValidator(bool disable_md5, bool disable_crc32c)` if both of them are not set to true, then only one option can be true. For example, disable_md5 -> use crc32chashvalidator. Then these lines:

https://github.com/googleapis/google-cloud-cpp/blob/959f84ca7fdee4ae1540a66ad9a4a3aca61e41e7/google/cloud/storage/internal/hash_validator.cc#L71-L74

Is where the problem begins. In order to preserve existing behavior, these both need to be set to `false`. Therefore without the mutual exclusitivity checks, then it's possible for both values to be false at the same time (when we've only set one option) which is not what we want.

Jseph:
___
*[google/cloud/storage/internal/hash_validator.cc, line 85 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2979#-LmRIUP4AmQccv3L9gRN:-LmRIUP4AmQccv3L9gRO:b-h2mjjz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/959f84ca7fdee4ae1540a66ad9a4a3aca61e41e7/google/cloud/storage/internal/hash_validator.cc#L85)):*
> ```
> request.GetOption<DisableCrc32cChecksum>().value();
> ```
I'm a little concerned this cold introduce check fails if people were to use the default constructor of DisableCrc32cChecksum, which used to work.  It sort of looks like this approach is recommended here:

https://github.com/googleapis/google-cloud-cpp/blob/dbd614681fe1b35de04396e25c59fe0c419c0f4a/google/cloud/storage/client.cc#L135

So anyone taking that comment literally could end up experiencing runtime failures as a result of this change.

remyabel:
> I'm a little concerned this cold introduce check fails if people were to use the default constructor of DisableCrc32cChecksum, which used to work. 

No existing behavior should've changed in this PR, as I didn't change any of the original tests, only added to them. If you see any differently LMK.

> So anyone taking that comment literally could end up experiencing runtime failures as a result of this change.

It is my understanding that `DisableCrc32cChecksum()` defaults to false, which is no different than not specifying it at all.

remyabel:
```
auto disable_md5 = request.GetOption<DisableMD5Hash>().value_or(false);
```

I like this suggestion but this doesn't compile as it doesn't have `value_or`, just `has_value` or `value`. Alternative? 

> I don't follow, if both are set to true the code returns a NullValidator. If only one is true then it returns the validator that is not disabled. If none is true then it returns a CompositeValidator. All cases are allowed I think?

Maybe my logic is flawed, but I am talking about when you pass one option (that is, the other is unset). This means if one is unset then it will be set to false in `CreateHashValidator`. If both are false it will create a `CompositeValidator`. Note that if this **is** desirable then the existing tests will need to be changed unless I'm missing something.

> I do not think that is correct. It defaults to not set:

Perhaps I misspoke. I meant passing `DisableMD5Hash()` to `CreateHashValidator` should **not** result in a crash, but if it's undefined behavior, then maybe it ran fine on my system by coincidence.

coryan:
>> ```
>> auto disable_md5 = request.GetOption<DisableMD5Hash>().value_or(false);
>> ```
> 
> I like this suggestion but this doesn't compile as it doesn't have `value_or`, just `has_value` or `value`. Alternative?

Doh!  I was thinking of `optional<T>`.  One alternative is:

```
auto disable_md5 = request.HasOption<DisableMD5Hash>() && request.GetOption<DisableMD5Hash>().value();
```

The other alternative is to add the `value_or()` function, should be easy as `optional<T>` already has it.

> Perhaps I misspoke. I meant passing `DisableMD5Hash()` to `CreateHashValidator` should **not** result in a crash, but if it's undefined behavior, then maybe it ran fine on my system by coincidence.

Calling `.value()` on an `optional<T>` that does not have a value is UB.  For a `bool` crashes would be rare I think.  But that is the thing about UB, you cannot trust that what works once will work in the future.

remyabel:
> Doh! I was thinking of optional<T>. One alternative is:

With this simplification the tests fail:

```
/v/google/cloud/storage/internal/hash_validator_test.cc:249: Failure
  QUICK_FOX_CRC32C_CHECKSUM
    Which is: "ImIEBA=="
  result.computed
    Which is: "crc32c=ImIEBA==,md5=nhB9nTcrtoJr2B01QqQZ1g=="
[  FAILED  ] CreateHashValidator.Read_OnlyCrc32c (0 ms)
[ RUN      ] CreateHashValidator.Read_OnlyMD5
# .. snip ..
```

My previous point stands. Did you want me to modify the original tests per your previous comment?

> Calling .value() on an optional<T> that does not have a value is UB. For a bool crashes would be rare I think. But that is the thing about UB, you cannot trust that what works once will work in the future.

Yes I know that. But I still don't see where a crash can occur. For:

```
  if (request.HasOption<DisableMD5Hash>()) {
    disable_md5 = request.GetOption<DisableMD5Hash>().value();
    if (!request.HasOption<DisableCrc32cChecksum>()) {
      disable_crc32c = !disable_md5;
    }
  }
```

If `DisableMD5Hash` is set, then we call `GetOption` (no UB there). If `DisableCrc32Checksum` is set we don't call `GetOption`. Is there a code path where that can happen?

coryan:
> > Doh! I was thinking of optional. One alternative is:
> 
> With this simplification the tests fail:
> 
> ```
> /v/google/cloud/storage/internal/hash_validator_test.cc:249: Failure
> Expected equality of these values:
>   QUICK_FOX_CRC32C_CHECKSUM
>     Which is: "ImIEBA=="
>   result.computed
>     Which is: "crc32c=ImIEBA==,md5=nhB9nTcrtoJr2B01QqQZ1g=="
> [  FAILED  ] CreateHashValidator.Read_OnlyCrc32c (0 ms)
> [ RUN      ] CreateHashValidator.Read_OnlyMD5
> # .. snip ..
> ```

Maybe the in your branch has drifted in other ways?  I just did this:

https://github.com/googleapis/google-cloud-cpp/compare/master...coryan:fix-hash-validator-false?expand=1

and the tests pass.




> 
> My previous point stands. Did you want me to modify the original tests per your previous comment?
> 

No.  I thought (and still think) that the tests should pass without changes (I did add *new* tests, of course).

> > Calling .value() on an optional that does not have a value is UB. For a bool crashes would be rare I think. But that is the thing about UB, you cannot trust that what works once will work in the future.
> 
> Yes I know that. But I still don't see where a crash can occur. For:
> 
> ```
>   if (request.HasOption<DisableMD5Hash>()) {
>     disable_md5 = request.GetOption<DisableMD5Hash>().value();
>     if (!request.HasOption<DisableCrc32cChecksum>()) {
>       disable_crc32c = !disable_md5;
>     }
>   }
> ```
> 
> If `DisableMD5Hash` is set, then we call `GetOption` (no UB there). If `DisableCrc32Checksum` is unset we don't call `GetOption`. Is there a code path where that can happen?

You are correct. I misread the code.  @Jseph, to answer your question that started this whole thread about UB: if the application says `DisableMD5Hash()` with the default constructor then `HasOption<>()` returns `false`.

remyabel:
> Maybe the in your branch has drifted in other ways?

Our tests make different assumptions. For example:

```c++
@@ -246,7 +246,8 @@ TEST(CreateHashValidator, Read_OnlyCrc32c) {
           .set_multiple_options(DisableCrc32cChecksum(false)));
   UpdateValidator(*validator, "The quick brown fox jumps over the lazy dog");
   result = std::move(*validator).Finish();
-  EXPECT_EQ(QUICK_FOX_CRC32C_CHECKSUM, result.computed);
+  EXPECT_THAT(result.computed, HasSubstr(QUICK_FOX_MD5_HASH));
+  EXPECT_THAT(result.computed, HasSubstr(QUICK_FOX_CRC32C_CHECKSUM));
```

Which means we are probably not on the same page about the expected behavior. I will be happy to incorporate your changes, but just FYI incase you can correct my understanding.

coryan:
I think the behavior should be:

- If neither `DisableMD5Hash(true)` nor `DisableCrc32cChecksum(true)` are provided, then both hashes should be present.
- If only  `DisableMD5Hash(true)` is provided, then the Crc32c checksum should be present by itself.
- If only  `DisableCrc32c(true)` is provided, then the MD5 hash should be present by itself.
- If both `DisableMD5Hash(true)` nor `DisableCrc32cChecksum(true)` are provided, then no hashes should be present.

Beyond that: passing `DisableMD5Hash(false)` or `DisableMD5Hash()` should have the same effect as *not* passing the option at all.  Same thing for `DisableCrc32c(false)` and `DisableCrc32c()`, they should be equivalent to *not* passing the options.

In summary: the checksums/hashes are enabled by default, and you can have to explicitly turn off each one of them by providing the `Disable*(true)` option.

remyabel:
Thanks for the help. I've incorporated your branch and added some comments to `CreateHashValidator` and made the test suite slightly more comprehensive. Maybe overkill, but also maybe worth it to prevent this type of bug from occurring again.

PR #2986, opened 2019-08-19
chore: reduce docker timeout from 120 minutes to 60 minutes

### LINKED ISSUE #2984: [Flake]: Timeout while building docker image.
Two builds failed.  They timed out while building the docker image:

https://source.cloud.google.com/results/invocations/df4ae4d5-a5a4-44f4-a3dd-b0b830c51a54
https://source.cloud.google.com/results/invocations/cf482fc7-d267-47f5-b639-89f9dcce27f7

The log is empty because we do not print the log of this phase unless it fails, and the timeout is managed outside our script.  We should change the script to timeout the docker image creation within 60 minutes (that is a rough guess, it should be more than 30 minutes, but less than the 120 minutes for the full build).

scotthart:
fixes #2984

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2986)

### DIFF
--- a/ci/kokoro/docker/build.sh
+++ b/ci/kokoro/docker/build.sh
@@ -225,7 +225,7 @@ echo "Logging to ${BUILD_OUTPUT}/create-build-docker-image.log"
 # is an error, so disabling from this point on is the right choice.
 set +e
 mkdir -p "${BUILD_OUTPUT}"
-if docker build "${docker_build_flags[@]}" ci \
+if timeout 3600s docker build "${docker_build_flags[@]}" ci \
     >"${BUILD_OUTPUT}/create-build-docker-image.log" 2>&1 </dev/null; then
    update_cache="true"
 fi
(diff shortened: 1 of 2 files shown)


ci/kokoro/docker/common.cfg
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 build_file: "google-cloud-cpp/ci/kokoro/docker/build.sh"
-timeout_mins: 120
+timeout_mins: 60
```
coryan: This does not have the desired effect. This reduces the timeout for the full build to 60 minutes.

ci/kokoro/docker/build.sh
```diff
@@ -225,7 +225,7 @@ echo "Logging to ${BUILD_OUTPUT}/create-build-docker-image.log"
 # is an error, so disabling from this point on is the right choice.
 set +e
 mkdir -p "${BUILD_OUTPUT}"
-if docker build "${docker_build_flags[@]}" ci \
+if timeout 3600s docker build "${docker_build_flags[@]}" ci \
```
coryan: nit: Apparently you can also write this as `timeout 60m`, or `timeout 1h`...

PR #2988, opened 2019-08-20
fix: correct invalid routing headers.

### LINKED ISSUE #2704: bigtable::MetadataUpdatePolicy for ClusterId only used in tests?

dopiera:
This fixes #2704.

Also, add unit tests for all API calls to make sure we catch mismatches
in the headers.

Sorry for the huge size, but I don't know how to meaningfully split it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2988)


dopiera:
___
*[google/cloud/bigtable/testing/validate_metadata.cc, line 27 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2988#-LmkFvce6IdXP6FmhJCG:-LmtOri7BfePTdHD4L_o:b-otfs82) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bdda4594055b9fd2e90ea03a5d9d15b539425873/google/cloud/bigtable/testing/validate_metadata.cc#L27)):*

At least it does not rely on implementation details.  The one thing that is dangerous is that a `ClientContext` can only be used once by gRPC, but presumably the tests that call this would never use it?  Do you think we should document that in `IsMDContextValid()` as: `// @note this function destroys the ClientContext`  or other suitably scary comment?
</blockquote></details>

Added a `@warning`.

BTW, I wonder why they didn't make the context movable - an r-value reference would be self descriptive.

PR #2989, opened 2019-08-20
Improve efficency of ObjectWriteStreambuf by replacing O(n^2) code.

Jseph:
Changes ObjectWriteStreambuf to always use the same buffer.

Right now `ObjectWriteStreambuf` reallocates `current_ios_buffer_` every time `xsputn()` is called.  This has O(n^2) efficency.  Using this library with a 2MiB buffer size with 1KiB writes to the buffer causes major CPU usage (image attached).  This change makes uploads O(n) for bytes uploaded regardless of the ratio of stream write size to buffer size.

TaskEfficiency

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2989)

### DIFF
--- a/google/cloud/storage/internal/object_streambuf.cc
+++ b/google/cloud/storage/internal/object_streambuf.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/storage/internal/object_requests.h"
 #include "google/cloud/storage/object_stream.h"
 #include <cstring>
+#include <stdint.h>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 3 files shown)


Jseph:
___
*[google/cloud/storage/internal/object_streambuf_test.cc, line 372 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2989#-Lmk-D8d8CsKqiNIF13F:-Lmk-D8d8CsKqiNIF13G:b-jfhflr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91b0a82eaa54230f5638112d41ddf8f305004c7d/google/cloud/storage/internal/object_streambuf_test.cc#L372)):*
> ```cpp
> 
> /// @test Verify that last error status is accessible for large payloads.
> TEST(ObjectWriteStreambufTest, ErrorInLargePayload) {
> ```

Note: I'm pretty sure this test was actually verifying bad behavior.  As noted above in a comment I'm pretty sure all upload failures are irrecoverable.

PR #2996, opened 2019-08-22
cleanup: Add missing TODO comment.

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2996)


dopiera:
FWIW, I think we actually need another test once it works. I.e. we need a test that 2 retries are not enough to exhaust the retry policy and a test to see if the exhausted retry policy returns the right error.

PR #2998, opened 2019-08-22
cleanup: Remove instances of redundant std::move

remyabel:
There were instances where `std::move` was called on `const` or trivial types.
These were spotted by enabling `clang-tidy` on the headers. Enabling it
wholesale has too many false positives, I only used it to spot these errors.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/2998)

### DIFF
--- a/google/cloud/bigtable/instance_update_config.h
+++ b/google/cloud/bigtable/instance_update_config.h
@@ -58,13 +58,13 @@ class InstanceUpdateConfig {
   //@}
 
   InstanceUpdateConfig& set_type(InstanceType type) {
-    proto_.mutable_instance()->set_type(std::move(type));
+    proto_.mutable_instance()->set_type(type);
     AddPathIfNotPresent("type");
     return *this;
   }
 
   InstanceUpdateConfig& set_state(StateType state) {
-    proto_.mutable_instance()->set_state(std::move(state));
+    proto_.mutable_instance()->set_state(state);
     AddPathIfNotPresent("state");
     return *this;
   }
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/instance_update_config.h
```diff
@@ -58,13 +58,13 @@ class InstanceUpdateConfig {
   //@}
 
   InstanceUpdateConfig& set_type(InstanceType type) {
-    proto_.mutable_instance()->set_type(std::move(type));
```
devjgm: What's the warning it gives here? My guess is that it complains because we're moving an object to a param that accepts by const-ref. But I'll say, so what? There's nothing wrong or inefficient about it. That seems like a clang-tidy bug to me -- I don't think it should complain about that. I say this because it would be nice if calling code could always use local reasoning to know when to move or not to move. The rule would simply be:

* If you're finished with the object, `std::move` is always fine.
* If you're not finished with the object, `std::move` is never fine.

Neither of those points have anything to do with how the called function accepts the argument. AFAICT, that's irrelevant. 

To be clear, @remyabel, I'm not complaining about you or your change. I'm complaining about clang-tidy warning about this :-)

remyabel: Correct, the check is `performance-move-const-arg`, which warns about `std::move`'s that have no effect. It also warns about trivial types that do not have a move constructor, which I think `InstanceType` is. Despite the name I don't think this check is about performance.

> If you're finished with the object, std::move is always fine.

There are cases where `std::move` can be pessimizing (i.e, prevent copy elision) but since in this case the `std::moves` simply have no effect and you do not like this check, I'll close this PR.

PR #3005, opened 2019-08-23
fix: return proper error code from upload metadata

### LINKED ISSUE #3002: Closing a failed upload stream yields invalid error code
Long story short, the last line of the following snippet fails., because metadata() return kUnknown.

```c++
auto os = client.WriteObject(...);
os << buf.data();  // assume this fails with kUnavailable
EXPECT_EQ(StatusCode::kUnavailable, os.last_status().code());
os.Close();
EXPECT_FALSE(os.metadata().ok());
EXPECT_EQ(StatusCode::kUnavailable, os.metadata().status().code());
```

dopiera:
This fixes #3002.

Before this fix, if a fatal error occurred while uploading a stream,
`stream.metadata()` would return `StatusCode::kUnknown` irrespectively
of the underlying cause. This PR fixes it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3005)

### DIFF
--- a/google/cloud/storage/client_write_object_test.cc
+++ b/google/cloud/storage/client_write_object_test.cc
@@ -156,37 +156,6 @@ TEST_F(WriteObjectTest, WriteObjectPermanentSessionFailurePropagates) {
       << ", status=" << stream.metadata().status();
 }
 
-TEST_F(WriteObjectTest, WriteObjectUnderhandedStreamClose) {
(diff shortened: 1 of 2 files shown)


dopiera:
Looks like the reason for failure under Windows is #3010.

coryan:
Can you rebase?  Code looks good but I would like all the tests to pass.

PR #3007, opened 2019-08-23
bug: Fix the initial backoff interval.

coryan:
The recommendation for backing off is to start with 1 second.
That makes sense because some operations have quotas of 1 per second.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3007)


coryan:
https://cloud.google.com/storage/docs/exponential-backoff

Issue #3003, opened 2019-08-23
[Flake]: Timeout in integration tests.


coryan:
The integration tests are flaking with:

```
Running storage::internal::CurlRequestDownload integration test.
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from CurlDownloadRequestTest
[ RUN      ] CurlDownloadRequestTest.SimpleStream
/v/google/cloud/storage/tests/curl_download_request_integration_test.cc:56: Failure
Value of: result
Expected: is OK
Actual: PerformWork() - CURL error [28]=Timeout was reached [DEADLINE_EXCEEDED]
/v/google/cloud/storage/tests/run_integration_tests_testbench.sh: line 48:  8710 Segmentation fault      (core dumped) ./curl_download_request_integration_test
```

I suspect this is my changes in #2993.  The default timeouts are probably too low for the CI environment.

https://source.cloud.google.com/results/invocations/8ed88241-7810-4a66-aa56-c4ea1e7300a7/targets/cloud-cpp%2Fgithub%2Fdocker%2Fcoverage-presubmit/log

dopiera:
FWIW, I'm moderately confident that the reason for the SIGSEGV after the timeout is #2955.

dopiera:
> FWIW, I'm moderately confident that the reason for the SIGSEGV after the timeout is #2955.
 I am now very confident that the reason for the SIGSEGV is different - #2955 was fixed, but the segmentation fault still happens: https://source.cloud.google.com/results/invocations/af99f437-a717-48b2-bb96-4546b2077aae/targets/cloud-cpp%2Fgithub%2Fdocker%2Fcoverage-presubmit/log

coryan:
Ack.  Shoulnd't the changes in #3028 have solved this problem?  A timeout should have resulted in a retry and the test should have (at least) taken much longer to fail?

dopiera:
I looked into it and:
* the SIGSEGV is most likely an easy oversight in the test, I sent #3031 
* the piece of code which is tested is deep enough that it doesn't really have retries (it's raw curl); I'm thinking of removing the test - it's bound to timeout on occasion and the logic must be tested in higher level integration tests too

coryan:
> * the SIGSEGV is most likely an easy oversight in the test, I sent #3031

I oversight when I changed the code to include timeouts.  The fix will eliminate the crashes, but not the problem...

[...]

PR #3028, opened 2019-08-30
fix: retry harder when reading

### LINKED ISSUE #2890: storage::Client::ReadObject() may not retry hard enough.
A customer is reporting that `storage::Client::ReadObject()` returns a stream with `.bad()` set and a status of `[UNAVAILABLE]`.  Maybe this sequence of events is taking place, if so, it is unclear that retrying on `Read()` is a good strategy (it would work poorly in the middle of a download), maybe we need to get the error status on `ReadObjectNotWrapped()` to start the retry loop earlier.

The only thing would be this sequence:
Client::ReadObjectImpl calls peek():
https://github.com/googleapis/google-cloud-cpp/blob/7011dd82004d89b1f2c77dfd47b3938b6a42cdbe/google/cloud/storage/client.cc#L62

That calls (after some intermediate code) RetryObjectReadSource::Read(), which then calls Read() on its child source:
https://github.com/googleapis/google-cloud-cpp/blob/7011dd82004d89b1f2c77dfd47b3938b6a42cdbe/google/cloud/storage/internal/retry_object_read_source.cc#L37

if that fails, then it will try to create a new child:
https://github.com/googleapis/google-cloud-cpp/blob/7011dd82004d89b1f2c77dfd47b3938b6a42cdbe/google/cloud/storage/internal/retry_object_read_source.cc#L48

That uses the MakeCall() function:
https://github.com/googleapis/google-cloud-cpp/blob/7011dd82004d89b1f2c77dfd47b3938b6a42cdbe/google/cloud/storage/internal/retry_client.cc#L246

So any error messages there would be familiar ones, assuming it succeeds, then it tries to read from the new child:
https://github.com/googleapis/google-cloud-cpp/blob/7011dd82004d89b1f2c77dfd47b3938b6a42cdbe/google/cloud/storage/internal/retry_object_read_source.cc#L55

but if that fails it gives up.

dopiera:
This fixes #2890.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3028)


dopiera:
The builds are failing because of unrelated bug: #3032

coryan:
Please rebase to pick up the openSUSE / CMake 3.15 fixes.

dopiera:
#3003 hits again. I sent #3034 to remove it.

dopiera:
@coryan  would you mind restarting the coverage test for me? I believe it's #3003 .

Issue #3027, opened 2019-08-30
internal::SetEnv inconsistent behavior on Windows and other systems


tmatsuo:
I found this on https://github.com/googleapis/google-cloud-cpp-spanner/pull/475

This code behaves differently on windows and other systems:
```
google::cloud::internal::SetEnv("WHATEVER", "");
google::cloud::internal::GetEnv("WHATEVER").has_value();
// false on Windows and true on other systems.
```
I think it's because of the spec of `putenv_s`.

[putenv_s](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=vs-2019):
> You can remove a variable from the environment by specifying an empty string (that is, "") for value_string.

remyabel:
I think on Windows there's no way to create an empty environment variable without doing something like setting it to a space. Should we have:

```
google::cloud::internal::SetEnv("WHATEVER", "");
```

do `putenv_s(var, " ")` on Windows? Or do we change the behavior of `SetEnv` so that an empty string deletes the environment variable? I lean towards the ladder since doing `SetEnv("WHATEVER", nullptr)` or `UnsetEnv` is more explicit when deleting a variable.

tmatsuo:
@remyabel 
Thanks!
How about the third option of doing nothing other than adding a documentation?

remyabel:
Fine by me. But I'll leave the tests in since they didn't exist before and it complements the documentation.

Issue #2949, opened 2019-08-07
Investigate errors upon file cleanup in windows bazel build


tmatsuo:
The windows bazel build passes, but I see some errors in the cleanup code:

```
Get-ChildItem : The directory name is invalid.
At line:1 char:4
+ & {Get-ChildItem -Recurse -File -Exclude test.xml,sponge_log.xml,buil ...
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ReadError: (T:\src\github\g...thon_wrapper.sh:String) [Get-ChildItem], IOException
    + FullyQualifiedErrorId : DirIOError,Microsoft.PowerShell.Commands.GetChildItemCommand
Get-ChildItem : The directory name is invalid.
At line:1 char:4
+ & {Get-ChildItem -Recurse -File -Exclude test.xml,sponge_log.xml,buil ...
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ReadError: (T:\src\github\g...arty\zlib.BUILD:String) [Get-ChildItem], IOException
    + FullyQualifiedErrorId : DirIOError,Microsoft.PowerShell.Commands.GetChildItemCommand
```

coryan:
We could simply remove that code if it does not help.

devjgm:
Have we seen anymore of these errors? If not, maybe we should just close this issue.

coryan:
These are just harmless. The build scripts try to remove any artifacts created by the build for reasons [1], it just happens that no such artifacts are present in these builds (they are in the `C:\b` directory).

[1]: We remove the artifacts because Kokoro performs an rsync after the build which is extremely slow on Windows, we are implementing the recommended workaround b/79892746

PR #3030, opened 2019-08-31
feat: Document/test the behavior of passing an empty string to SetEnv on Windows

remyabel:
Fixes https://github.com/googleapis/google-cloud-cpp/issues/3027

~This is a quick PR to implement my suggestion. Note that I'm not 100% behind this idea because it may be surprising that the env var isn't actually empty when using `SetEnv(var, "")` on Windows.~

This documents the Windows behavior and adds some tests.

cc @tmatsuo

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3030)

### DIFF
--- a/google/cloud/internal/setenv.h
+++ b/google/cloud/internal/setenv.h
@@ -29,6 +29,9 @@ void UnsetEnv(char const* variable);
  * Set the @p variable environment variable to @p value.
  *
  * If @value is the null pointer then the variable is unset.
+ *
+ * @note On Windows, due to the underlying API function, an empty @value unsets
+ * the variable, while on Linux an empty environment variable is created.
  */
 void SetEnv(char const* variable, char const* value);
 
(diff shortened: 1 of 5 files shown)


tmatsuo:
@remyabel Thanks, LGTM, but I'd like to hear from @coryan too.

tmatsuo:
Maybe you can tweak the title, because it just adds some comments?

google/cloud/internal/setenv.h
```diff
@@ -29,6 +29,9 @@ void UnsetEnv(char const* variable);
  * Set the @p variable environment variable to @p value.
  *
  * If @value is the null pointer then the variable is unset.
+ *
+ * @note On Windows, due to the underlying API function, an empty @value unsets
+ * the variable, while on Linux an empty environment variable is created.
```
coryan: nit: can we link to some Windows documentation here?

coryan:
@remyabel @tmatsuo I definitely do not understand something here, because the tests are failing.

remyabel:
The tests were failing because I forgot to modify the other tests. I think in the second log this [error](https://github.com/googleapis/google-cloud-cpp/issues/2949) came up again.

PR #3031, opened 2019-08-31
fix: eliminate a SIGSEGV from test

dopiera:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3031)


dopiera:
The builds are failing because of an unrelated bug: #3032

Issue #3032, opened 2019-08-31
Installation integration test fails on OpenSuse


dopiera:
When trying to compile `test-install/bigtable` after installing on OpenSuse, cmake fails with:
```
CMake Error at /usr/local/lib64/cmake/googleapis/googleapis-config.cmake:43 (add_library):
  add_library cannot create imported target
  "googleapis-c++::api_resource_protos" because another target with the same
  name already exists.

```
This seems to boil down to including the file file (`googleapis-config.cmake`) twice:
* via
```
  /usr/share/cmake/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /usr/local/lib64/cmake/bigtable_client/bigtable_client-config.cmake:17 (find_dependency)
  CMakeLists.txt:30 (find_package)
```
* and via
```
Call Stack (most recent call first):
  /usr/share/cmake/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /usr/local/lib64/cmake/google_cloud_cpp_grpc_utils/google_cloud_cpp_grpc_utils-config.cmake:16 (find_dependency)
  /usr/share/cmake/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /usr/local/lib64/cmake/bigtable_client/bigtable_client-config.cmake:19 (find_dependency)
  CMakeLists.txt:30 (find_package)
```

I haven't figured out why, yet. I'll get back to it on Monday.

remyabel:
It appears that in the failed PR's the images are being rebuilt, probably from a base image update. That suggests to me that a package update (maybe CMake) is causing the failure. I don't have the proper permissions to pull from gcr.io, can you try using the cached image and see if it works?

```diff
diff --git a/ci/kokoro/install/Dockerfile.opensuse b/ci/kokoro/install/Dockerfile.opensuse
index b9ad60021..b33e906b4 100644
--- a/ci/kokoro/install/Dockerfile.opensuse
+++ b/ci/kokoro/install/Dockerfile.opensuse
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 ARG DISTRO_VERSION=latest
-FROM opensuse/tumbleweed:${DISTRO_VERSION} AS devtools
+FROM gcr.io/cloud-cpp-testing-resources/google-cloud-cpp/test-install-opensuse:latest AS devtools
 ARG NCPU=4
 
 ## [START INSTALL.md]
```

```shell
[...]

PR #3033, opened 2019-09-01
bug: include FindgRPC directly.

coryan:
Update to the latest cpp-cmakefiles release which supports CMake 3.15.

We also include `FindgRPC.cmake` directly because with CMake 3.16
`find_dependency()` can no longer find it.

Part of the work for #3032.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3033)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -205,9 +205,9 @@ There is no Fedora package for this library. To install it, use:
 
 ```bash
 cd $HOME/Downloads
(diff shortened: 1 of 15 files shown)


coryan:
I think this is ready for review now.

coryan:
> Thanks. Do you think we should open a new issue to determine if this was a CMake bug and potentially report it? Even if our code was incorrect, updating shouldn't break compatibility, at least not without a policy warning.

It would be nice to report the issue back, I am not sure how to repro this in a small enough scale.

PR #3048, opened 2019-09-13
test: Add a test for telemetry header

tmatsuo:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3048)

### DIFF
--- a/google/cloud/storage/tests/curl_request_integration_test.cc
+++ b/google/cloud/storage/tests/curl_request_integration_test.cc
@@ -173,12 +173,13 @@ TEST(CurlRequestTest, CheckResponseHeaders) {
   EXPECT_EQ("bar", response->headers.find("x-test-foo")->second);
 }
 
-/// @test Verify the user agent prefix affects the request.
-TEST(CurlRequestTest, UserAgentPrefix) {
+/// @test Verify the user agent header.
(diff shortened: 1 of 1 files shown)


google/cloud/storage/tests/curl_request_integration_test.cc
```diff
@@ -199,6 +199,30 @@ TEST(CurlRequestTest, UserAgentPrefix) {
...
+  builder.AddHeader("Accept: application/json");
+  builder.AddHeader("charsets: utf-8");
+
+  auto response = builder.BuildRequest().MakeRequest(std::string{});
+  ASSERT_STATUS_OK(response);
+  EXPECT_EQ(200, response->status_code);
+  auto payload = nl::json::parse(response->payload);
+  ASSERT_EQ(1U, payload.count("headers"));
+  auto headers = payload["headers"];
+  EXPECT_THAT(headers.value("User-Agent", ""), HasSubstr("gcloud-cpp/"));
```
coryan: So the only difference with the previous test is that we check the `gcloud-cpp/` value is automatically inserted?  Should we simply merge the two tests?

tmatsuo: Agreed, done!

PR #3058, opened 2019-09-14
feat: added ReadLast option for reading object from end

### LINKED ISSUE #3006: Implement option to read the last N bytes from an object.
Apparently sending `Range: bytes=-10` gives you the last 10 bytes, who knew.

We can create an option like `ReadFromOffset`, maybe call it `ReadTail` or `ReadLast`.

manish-qlogic:
This PR tries to fix #3006.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3058)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -35,6 +35,7 @@
 #include "google/cloud/storage/retry_policy.h"
 #include "google/cloud/storage/upload_options.h"
 #include "google/cloud/storage/version.h"
+#include <type_traits>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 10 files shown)


coryan:
___
*[google/cloud/storage/internal/object_requests.cc, line 262 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3058#-Lop--yUBi1Zv9oduqTI:-LozWBqBDZFG5sKbbmKJ:br6u0zf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d9e16a20c3062aaacc03288787bdf425ce30c051/google/cloud/storage/internal/object_requests.cc#L262)):*
<details><summary><i>Previously, manish-qlogic wrote…</i></summary><blockquote>

apparently,

ReadLast(0) returns stream.status() as [OUT_OF_RANGE]

and ReadLast(-1) a.k.a ("Range: bytes=--1") or any other negative value returns the entire object. 
 
</blockquote></details>

Ack, thanks for looking this up.  I think we want the no-cache option turn on for any value in ReadLast.  The errors sent by the server (or the handling of negative values) are not a concern.

PR #3062, opened 2019-09-16
bug: calls to resumable_session_id() result in segfaults

### LINKED ISSUE #3047: Calls to resumable_session_id() Result in Segfaults

yes

- Cloud Storage.

1.4

call resumable_session_id() after encountering a retriable error in the resumable upload stream.

The resumable session id.

### What did you see instead?
A segfault.

I introduced this behavior in #2989 where I reset the upload session when calls to Flush() fail.  What do we want the behavior here to be?

Jseph:
This fixes the problem as discussed in the issue.

Fixes #3047 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3062)

### DIFF
--- a/google/cloud/storage/client_write_object_test.cc
+++ b/google/cloud/storage/client_write_object_test.cc
@@ -145,10 +145,12 @@ TEST_F(WriteObjectTest, WriteObjectPermanentSessionFailurePropagates) {
     return StatusOr<std::unique_ptr<internal::ResumableUploadSession>>(
         std::unique_ptr<internal::ResumableUploadSession>(mock_session));
   };
+  std::string const empty;
   EXPECT_CALL(*mock, CreateResumableSession(_)).WillOnce(Invoke(returner));
   EXPECT_CALL(*mock_session, UploadChunk(_))
       .WillRepeatedly(Return(PermanentError()));
   EXPECT_CALL(*mock_session, done()).WillRepeatedly(Return(false));
+  EXPECT_CALL(*mock_session, session_id()).WillRepeatedly(ReturnRef(empty));
   auto stream = client->WriteObject("test-bucket-name", "test-object-name");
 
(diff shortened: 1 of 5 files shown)


Jseph:
___
*[google/cloud/storage/client_write_object_test.cc, line 153 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3062#-LovmMxqBJyhDU7RmYKA:-LovmKBP9bzs2AMcpByG:b26kiwt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b0f7af178aaedd06be042398239a0008d9ebd100/google/cloud/storage/client_write_object_test.cc#L153)):*
> ```
> EXPECT_CALL(*mock_session, session_id()).WillRepeatedly(ReturnRef(empty));
> ```

This test was really hard for me to debug because exceptions are "silently" handled by the C++ streams library, so gmock wasn't able to report that this function had no default behavior.   Do you know of any way we could disable that behavior for tests so future contributors don't run into this?

Jseph:
___
*[google/cloud/storage/internal/resumable_upload_session.h, line 153 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3062#-LoxQ7i7373d5xtt0FmU:-Lp4z1xV3wSSiQZPjwxf:bpvti4w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b0f7af178aaedd06be042398239a0008d9ebd100/google/cloud/storage/internal/resumable_upload_session.h#L153)):*

nit: I think we do not have any other member variables that are `const` in the library. Is it important that these be, or would you be Okay with making them non-const?
</blockquote></details>

Ok, don't mind at all. Removed the const markers.

coryan:
I made some changes to the  description and title, I hope they are Okay.

Issue #3029, opened 2019-08-30
GCS upload timeout / spinning in poll


andir:
### Does this issue affect the `google-cloud-cpp` project?

Yes, we are using the `google-cloud-cpp` library to upload files.

### What component of `google-cloud-cpp` is this related to?

- Cloud Storage specifically uploading new files.

### What version of `google-cloud-cpp` are you using?

0.12.0 & 0.11.0

### What compiler and version are you using?

gcc 7.4.0

### What operating system and version are you using?

NixOS with based on commit https://github.com/nixos/nixpkgs/commits/3f4144c30a6

(curl version 7.65.3, manually bumped google-cloud-cpp to 0.12.0 from 0.11.0 but same issue persists)

### What were you trying to do?

Upload files to a GCS bucket using [`nix`](https://nixos.org/nix). Specifically with the changes from my PR https://github.com/NixOS/nix/pull/3021.

We tried with big and small files. Both `InsertObject` and `WriteObject` and in all cases we get the same result (randomly). The relevant code is are at https://github.com/NixOS/nix/pull/3021/files#diff-b1e603a1ea8f9c5a0d38f7caf7079cdaR97-R122.

Command that we used (after building nix from that PR with a recent nixpkgs checkout):
```
$ nix copy --to gs://my-bucket-name?secret-key=/my/secret/signing/key /nix/store/…-something
```

### What did you expect to see?

To successfully upload the binary without wasting a lot of time in `poll(2)` and without seeing the error message in the end.

### What did you see instead?

`nix` did waste a lot of CPU time (minutes) in `poll(2)` called by `Curl_poll(2)`. See backtrace and perf data below.

Eventually the google-cloud-cpp library did throw an exception showing a `Request Timeout` error.

It seems like it never tells google that the upload is finished but waits for a response to the request.

### Anything else you would like us to know?

I am currently working on integrating GCS support into `nix`. While I can not see any obvious errors with my implementation this error occurs randomly for some people.

We captured a backtrace while the process was spinning in `poll(2)`:

```
Thread 1 (Thread 0x7f0418b7fa00 (LWP 15471)):
#0  0x00007f0419ba8501 in poll () from target:/nix/store/iykxb0bmfjmi7s53kfg6pjbfpd8jmza6-glibc-2.27/lib/libc.so.6
[...]

PR #3066, opened 2019-09-18
feat: implement a regression test for large uploads

coryan:
This is a regression test, but so far does not reproduce a problem. But
it might help us diagnose the problem reported in #3029 

The test runs multiple threads, all of them uploading a mid-sized
object (128MiB). The test uses `InsertObject()` and `WriteObject()`
to perform the upload, and in the case of `WriteObject()` it both
uploads the object using a single `operator<<` call, and multiple
`.write()` calls.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3066)

### DIFF
--- a/google/cloud/storage/tests/parallel_upload_regression/Dockerfile
+++ b/google/cloud/storage/tests/parallel_upload_regression/Dockerfile
@@ -1,0 +1,54 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM ubuntu:bionic AS devtools
+
+# Install the typical development tools and some additions:
+RUN apt update && \
+    apt install -y build-essential cmake git gcc g++ cmake \
+        libc-ares-dev libc-ares2 libcurl4-openssl-dev libssl-dev make \
+        pkg-config tar wget zlib1g-dev
+
+FROM devtools AS build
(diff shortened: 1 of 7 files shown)


google/cloud/storage/tests/parallel_upload_regression/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+RUN apt update && \
+    apt install -y libstdc++6 libc-ares2 libcurl4 zlib1g binutils \
+        ca-certificates tcpdump coreutils strace procps
+RUN /usr/sbin/update-ca-certificates
+
+COPY --from=build \
+    /w/.build/build/google-cloud-cpp/src/google_cloud_cpp_project-build/google/cloud/storage/tests/parallel_upload_regression/parallel_upload_test \
+    /r/
+
+CMD ["/bin/false"]
```
devbww: What does this do?

google/cloud/storage/tests/parallel_upload_regression/parallel_upload_test.cc
```diff
@@ -0,0 +1,204 @@
...
+      auto writer = client.WriteObject(flag_dst_bucket_name, name);
+      for (std::size_t offset = 0; offset < upload_contents_.size();
+           offset += chunk_size) {
+        auto count = (std::min)(chunk_size, upload_contents_.size() - offset);
+        writer.write(upload_contents_.data() + offset, count);
+        EXPECT_FALSE(writer.bad());
+        EXPECT_STATUS_OK(writer.last_status());
+      }
+      writer.Close();
+      std::cout << '.' << std::flush;
```
devbww: When do these lines get terminated?

coryan: Thanks, fixed.

google/cloud/storage/tests/parallel_upload_regression/Dockerfile
coryan: Set the default command for `docker run` to `/bin/false`.  Harder to misuse the resulting docker image that way (I think).

devbww: Okie dokie.  Thanks.

PR #3072, opened 2019-09-22
bug: restart downloads with retryable HTTP errors.

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3072)


coryan:
PTAL.

PR #3073, opened 2019-09-23
feat: add long delays to read stall regression test

coryan:
Verify the library can handle applications that, from time to
time, sleep for more than 5 minutes. This seems to be a critical
value to disconnect idle transfers on the server-side, and thus
they are important to test.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3073)

### DIFF
--- a/google/cloud/storage/tests/read_object_stall_regression/read_object_stall_test.cc
+++ b/google/cloud/storage/tests/read_object_stall_regression/read_object_stall_test.cc
@@ -343,8 +343,7 @@ class ReadObjectStallTest
     std::uniform_real_distribution<double> use_long_delay(0.0, 1.0);
     for (int i = 0; i != max_chunks; ++i) {
       auto sleep_seconds = Delay();
-      std::cout << "Sleeping for " << sleep_seconds.count() << "s\n"
-                << std::flush;
+      std::cout << "Sleeping for " << sleep_seconds.count() << "s\n";
       std::this_thread::sleep_for(sleep_seconds);
       WriteSummary(client, summary_name, read_summaries);
     }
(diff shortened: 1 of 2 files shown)


google/cloud/storage/tests/read_object_stall_regression/read_object_stall_test.cc
```diff
@@ -336,17 +340,22 @@ class ReadObjectStallTest
...
     std::geometric_distribution<int> sleep_period(1.0 / flag_average_delay);
+    std::uniform_real_distribution<double> use_long_delay(0.0, 1.0);
     for (int i = 0; i != max_chunks; ++i) {
-      int sleep_seconds = sleep_period(generator_);
-      std::cout << "Sleeping for " << sleep_seconds << "s\n" << std::flush;
-      std::this_thread::sleep_for(std::chrono::seconds(sleep_seconds));
-
+      auto sleep_seconds = Delay();
+      std::cout << "Sleeping for " << sleep_seconds.count() << "s\n"
+                << std::flush;
```
devbww: Do we need the flush after a '\n'?

coryan: Probably, the next line is a sleep. Without the flush one may wonder why is the program making no progress.  :man_shrugging:  typically I just read the logs after the program runs for a couple of hours.

devbww: If we're trying to read block-buffered logs while they are still open, I agree there is an issue.  Perhaps we should try to line buffer cout/clog no matter where they are directed?

Issue #3074, opened 2019-09-24
Crashes when downloading to files in multiple threads.


coryan:
I got a report via email. Calling `client.DownloadToFile()` from multiple threads, sharing the same `client` can result in crashes.

This is on Debian 10 (Buster), with:

curl libs : curl 7.64.0 (x86_64-pc-linux-gnu) libcurl/7.64.0 OpenSSL/1.1.1c zlib/1.2.11 libidn2/2.0.5 libpsl/0.20.2 (+libidn2/2.0.5) libssh2/1.8.0 nghttp2/1.36.0 librtmp/2.3

compiler : g++ (Debian 8.3.0-6) 8.3.0

google-cloud-cpp: 0.13.0
crc32c: 1.1.0

To repro, create about 1,000 small objects (about 16KiB) in a single bucket, download them using 8 threads.

coryan:
FWIW, I was able to reproduce this on Debian 10, but cannot reproduce on Ubuntu 16:04.  I think the main different is the libcurl version, will try this will different versions and see if I can repro that way.

coryan:
Just tested on Ubuntu 16:04 with libcurl-7.64 and I got a crash pretty reliably.  Same platform with libcurl-7.66 no crash so far (32 threads 10,000 objects, running again as I write this).

I am inclined to believe this is a libcurl bug.  Time for me to read the change log for libcurl.

coryan:
For what is worth, 7.65 seems fine too.  An anonymous contributor clued me to curl/curl#3904 which seems very likely to be the issue.  I will follow up with a PR to merge the test I used in these experiments.

coryan:
I think with #3079 and #3080 we can close this bug.

PR #3079, opened 2019-09-26
feat: a new test for multithreaded downloads

coryan:
I wrote this test to reproduce #3074. I believe this is a libcurl bug, but the test and the Dockerfile to build with Debian 10 are useful.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3079)

### DIFF
--- a/google/cloud/storage/tests/object_file_multi_threaded_test.cc
+++ b/google/cloud/storage/tests/object_file_multi_threaded_test.cc
@@ -80,14 +80,14 @@ class ObjectFileMultiThreadedTest
 
   void CreateObjects(Client const& client,
                      std::vector<std::string> const& object_names) {
-    int const thread_count = ThreadCount();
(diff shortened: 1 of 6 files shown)


google/cloud/storage/tests/object_file_multi_threaded_test.cc
```diff
@@ -0,0 +1,211 @@
...
+    }
+    return Status();
+  }
+
+  void CreateObjects(Client const& client,
+                     std::vector<std::string> const& object_names) {
+    int const thread_count = ThreadCount();
+    std::vector<std::future<Status>> tasks(thread_count);
+    // Parallelize the object creation too because it can be slow.
+    int modulo = 0;
```
devbww: Nit: I'd move the local `modulo` declarations down to the loops that use them.

PR #3025, opened 2019-08-29
feat: Add ability to disable building libraries

remyabel:
Fixes #3024

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3025)

PR #3177, opened 2019-10-02
update release notes [skip ci]

mr-salty:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3177)

### DIFF
--- a/google/cloud/bigtable/README.md
+++ b/google/cloud/bigtable/README.md
@@ -46,7 +46,7 @@ Apache 2.0; see [`LICENSE`](../../../LICENSE) for details.
 
 ### v1.3.x - 2019-10
 
-* *no bigtable-specific changes; see [common](../README.md) v0.12.x*
+* bug: fix runtime install directory (#3063)
 
 ### v1.2.x - 2019-09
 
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/README.md
```diff
@@ -42,7 +42,11 @@ Apache 2.0; see [`LICENSE`](../../../LICENSE) for details.
 
 ## Release Notes
 
-### v1.3.x - TBD
+### v1.4.x - TBD
+
+### v1.3.x - 2019-10
```
coryan: nit: I would just duplicate the other release note, but I also do no think it is that important.

mr-salty:
___
*[google/cloud/bigtable/README.md, line 47 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3177#-LqDYfnzY_y0r_cY9JId-r1-47:-LqDawjU4ku3C3clCqaC:b-73d2lo) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7e3aa29f8ff1923bb6088ab8ba0119421c89d3ec/google/cloud/bigtable/README.md#L47)):*

I can go either way on that, I noted that bww "pulled up" some common changes in the prior version like #3025 so I did the same for consistency. I suppose it will not matter once the repos are split.
</blockquote></details>

Actually I guess the splitting makes a good argument for duplicating, so I'll do that.

mr-salty:
Can I get a LG or do we have edits to the notes?

PR #3197, opened 2019-10-13
feat:new helper function for 'archive' storage class

### LINKED ISSUE #3075: Create new helper function for "archive" storage class.
We have a number of helper functions for storage class:

https://github.com/googleapis/google-cloud-cpp/blob/e9ccd39d9f12066f43d77f621fdb560a6a0a178b/google/cloud/storage/storage_class.h#L26-L54

We need to add another one of these for `ARCHIVE`.

manish-qlogic:
This fixes #3075

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3197)

### DIFF
--- a/google/cloud/storage/lifecycle_rule.cc
+++ b/google/cloud/storage/lifecycle_rule.cc
@@ -54,6 +54,10 @@ LifecycleRuleAction LifecycleRule::SetStorageClassDurableReducedAvailability() {
   return SetStorageClass(storage_class::DurableReducedAvailability());
 }
 
+LifecycleRuleAction LifecycleRule::SetStorageClassArchive() {
+  return SetStorageClass(storage_class::Archive());
+}
+
 LifecycleRuleAction LifecycleRule::SetStorageClass(std::string storage_class) {
   return LifecycleRuleAction{"SetStorageClass", std::move(storage_class)};
 }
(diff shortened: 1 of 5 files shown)


coryan:
> * Would we want to add Archive to the integration test as well? See https://github.com/googleapis/google-cloud-cpp/blob/150aaf8296a5dba47dd28e2a448e4bdea46f9037/google/cloud/storage/tests/bucket_integration_test.cc#L83

Maybe, the goal of that integration test is to verify we can change the storage class field value, any valid value would be acceptable in the test.  Feel free to open a separate issue if you think it is important.

> * In the future we should probably wait to merge until the feature is no longer behind a flag (though not a big deal this time).

I disagree. In general, we merge as soon as possible, and will continue to do so. Gating on everything to be "just so" is a sure way to kill velocity.

tritone:
1. Makes sense, no need to add it to the integration test in that case.
2. I definitely understand that you don't want to leave PRs sitting unresolved, but it does seem like it would be disorienting to users that we allow them to make calls  which will fail because the feature is not ready yet. Would it be preferable to just assign the work later in a situation like this? Or do you think it's not a big deal from a user perspective? I guess it's only a matter of weeks until release...

Issue #3016, opened 2019-08-26
Implement helper function to recursive compose objects.


coryan:
The JSON API is limited to composing 32 objects at a time:

https://cloud.google.com/storage/docs/json_api/v1/objects/compose

However there is no limit to the number of components in an object:

https://cloud.google.com/storage/docs/composite-objects#_Count

The library could have a helper function to efficiently compose "N" objects where N is much larger than 32.  The library should divide the source objects is groups of 32 or less objects, then compose those objects (in parallel?) once those succeed it can recursively compose the resulting objects again until there is only one object left to compose.

acastruc:
Not trying to press my luck, but it would be awesome if this function had an option to delete the parts?

coryan:
I think this will have to remove the temporary objects it creates, so adding an option to remove the initial files does not sound that out of scope.

I do worry about the return type for this.  The compose operations may fail (even after retries) and we should stop merging, delete *only* the temporary objects in that case.  But those deletes may also fail, and we need to indicate which ones so the application developers can delete those themselves (and/or they should be created with a lifecycle policy to be automatically deleted by the service).

Only once the composition has succeeded we can proceed to delete the source objects. But that *also* fail, so we should report any failures in that step.  The return value is becoming something like:

```C++
namespace google::cloud::storage {

struct CompositionResult {
  StatusOr<ObjectMetadata> composed_object;
  // Object names and reason for delete failures
  std::vector<std::pair<std::string, Status>> delete_failures;
};

// Standalone function to compose multiple objects.
template <....>
CompositionResult ComposeMany(
    std::string bucket_name,
    std::vector<ComposeSourceObject> source_objects,
    std::string destination_object_name,
    Options&&... options);
}  // namespace 
```

Because this operation may take "a while" to run, we should consider returning this wrapped in a `google::cloud::future<>`.

coryan:
@dopiera we should discuss this in the next meeting.

dopiera:
I think we should take a different approach.

The client process might simply die in the middle of the operation. With this proposal we won't have an easy way to communicate to the user which temporary objects can be deleted. Similarly, if the client dies after executing the final composition but before removing the source files, a manual user intervention is still necessary.

[...]

PR #3198, opened 2019-10-14
feat: implement std::apply.

dopiera:
This is a part of #3016.

Composing GCS objects consists of multiple operations under the hood.
Different subsets of the options passed to the recursive compose
operation have to be passed on to the underlying operations.

By GCS C++ client convention, options are passed as parameter packs,
hence in the general case the filtering has to be done in compile time.

An easy way to filter these parameters is to pack them in a
`std::tuple`, process the tuple and apply it via `std::apply`.
Unfortunately, `std::apply` is only available in C++14, so this PR
reimplements it in a API-compatible way.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3198)

### DIFF
--- a/google/cloud/CMakeLists.txt
+++ b/google/cloud/CMakeLists.txt
@@ -161,6 +161,8 @@ add_library(google_cloud_cpp_common
             internal/setenv.cc
             internal/throw_delegate.h
             internal/throw_delegate.cc
+            internal/tuple.h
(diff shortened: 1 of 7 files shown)


dopiera:
Closing in favor of https://github.com/googleapis/google-cloud-cpp-common/pull/59

PR #3202, opened 2019-10-15
cleanup: update MultiRegional and Regional to Standard

### LINKED ISSUE #3044: Update MultiRegional and Regional to Standard
Previously, we implemented the concept of "Location Type" to move the concept of bucket location away from storage class. It is no longer considered correct to use "MultiRegional" or "Regional" as a storage class, instead those should be location types.

As such, we need to update every instance of the "MultiRegional" and "Regional" storage in samples, library documentation, and tests to use "Standard" instead--Standard is a correct storage class.

Note that "MultiRegional" and "Regional" should only be replaced with "Standard" in cases where they are used a storage class. If they're used as a location type, that's fine and doesn't need to be replaced.

https://github.com/googleapis/google-cloud-cpp/search?utf8=%E2%9C%93&q=regional&type=

manish-qlogic:
This fixes #3044

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3202)


manish-qlogic:
___
*[google/cloud/storage/benchmarks/storage_file_transfer_benchmark.cc, line 93 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3202#-LrF956U2ir1fKxPejKn:-LrGpgRt0vbWt4qIstks:bymirpq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/fda2219798752d88f70ca07d2d13358d82cc6c9b/google/cloud/storage/benchmarks/storage_file_transfer_benchmark.cc#L93)):*

Instead, change this to say `.set_location_type(gcs::location_type::Regional())` hmmm... if we do not have the `location_type` helpers we should probably add them.
</blockquote></details>

I don't find a setter for location_type, is this intentional?

manish-qlogic:
___
*[google/cloud/storage/benchmarks/storage_file_transfer_benchmark.cc, line 93 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3202#-LrF956U2ir1fKxPejKn:-LrGvXVB9-IFsF9n8HpW:b-6bqill) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/fda2219798752d88f70ca07d2d13358d82cc6c9b/google/cloud/storage/benchmarks/storage_file_transfer_benchmark.cc#L93)):*
<details><summary><i>Previously, manish-qlogic wrote…</i></summary><blockquote>

I don't find a setter for location_type, is this intentional?
</blockquote></details>

and current values for location_type are `region`, `dual-region` and `multi-region` as shown in following documentation

https://cloud.google.com/storage/docs/storage-classes

I presume, we should have helper functions with the same names.

manish-qlogic:
___
*[google/cloud/storage/benchmarks/storage_latency_benchmark_go.go, line 100 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3202#-LrF9Oeb5kFDOMYkdGUs:-LrHAl2r06Y3ico8Uhla:bqoq3i5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/fda2219798752d88f70ca07d2d13358d82cc6c9b/google/cloud/storage/benchmarks/storage_latency_benchmark_go.go#L100)):*

Same here, though I am not sure how this is represented in the golang library.
</blockquote></details>

`LocationType` as mentioned here
https://github.com/googleapis/google-cloud-go/blob/b9197a057a6cb1129f29cd21ca11d1bb043666d1/storage/bucket_test.go#L407

Issue #3056, opened 2019-09-14
Migrate storage testbench to Python 3


coryan:
Currently it only runs with Python 2.7 and that is going EOL soon.

coryan:
The builds are failing, I think because some of the modules assume Python 3, staying on Python 2.7 has become untenable I think.

PR #3204, opened 2019-10-15
fix(testbench): improve Python 3 compatibility

crwilcox:
Addresses some incompatibilities with Python 3. With these changes I was able to use Python 3 to upload a file to the test bench.

Related bug: #3056

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3204)


ci/kokoro/docker/Dockerfile.fedora
```diff
@@ -36,15 +36,15 @@ RUN dnf makecache && dnf install -y \
     ncurses-term \
     openssl-devel \
     pkgconfig \
-    python \
+    python38 \
```
remyabel: Why python38? Why not just install both python and python3? Most people will not have python38 installed. The current version of `python3` is `3.7.5`, which is also the default in Fedora 31.

ci/kokoro/docker/Dockerfile.fedora-install
```diff
@@ -40,15 +40,15 @@ RUN dnf makecache && dnf install -y \
...
+    python38 \
     shtool \
     unzip \
     wget \
     which \
     zlib-devel
 
-RUN pip install --upgrade pip
-RUN pip install httpbin flask gevent gunicorn crc32c
+RUN python3.8 -m pip install --upgrade pip
```
remyabel: Genuine question: is it better to do `python3 -m pip` or `pip3` here?

ci/kokoro/docker/Dockerfile.fedora
crwilcox: I am happy to rely on 3.7.5. I didn't count on there being one. happy to change this :)

ci/kokoro/docker/Dockerfile.fedora-install
crwilcox: Not all OS platforms have `pip3`. Calling `python3 -m pip` ensures the pip we are using is the one connected to that particular interpreter. So the more consistent experience across platforms and double-sureness is why I tend to do this. In theory, they are equivalent.

crwilcox:
@remyabel @coryan since opening this I figured out running the tests locally. If having this hang around is irritating, I can close it and reopen it once it is ready to merge. Just let me know.

coryan:
Let's close it for now. Thanks for making these changes btw.

PR #3221, opened 2019-10-24
feat: implement tuple filtering

dopiera:
This is a part of #3016.

Composing GCS objects consists of multiple operations under the hood.
Different subsets of the options passed to the recursive compose
operation have to be passed on to the underlying operations.

By GCS C++ client convention, options are passed as parameter packs,
hence in the general case the filtering has to be done in compile time.

An easy way to filter these parameters is to pack them in a
std::tuple, filter it and apply it via std::apply.

The intended use is available at https://github.com/dopiera/google-cloud-cpp/commit/60fa4402bb317c56a6c8cea6e558119c7024bcb7 and https://github.com/dopiera/google-cloud-cpp/commit/bd75ad53f05cdbfd3f32557357b4e9bb3cc08b38

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3221)

### DIFF
--- a/google/cloud/storage/internal/tuple_filter.h
+++ b/google/cloud/storage/internal/tuple_filter.h
@@ -30,6 +30,16 @@ namespace storage {
 inline namespace STORAGE_CLIENT_NS {
 namespace internal {
 
+/// Prepend a type to tuple's type list - unmatched case.
+template <typename TL, typename T>
+struct TupleTypePrepend {};
+
+/// Prepend a type to tuple's type list - std::tuple specialization
+template <typename T, typename... Args>
+struct TupleTypePrepend<std::tuple<Args...>, T> {
+  using Result = std::tuple<T, Args...>;
+};
+
 /**
  * A helper class to filter a single element from a tuple.
  *
@@ -81,13 +91,9 @@ template <template <class> class TPred, typename Head, typename... Tail>
 struct FilteredTupleReturnType<TPred, std::tuple<Head, Tail...>> {
   using Result = typename std::conditional<
       TPred<Head>::value,
-      typename google::cloud::internal::invoke_result<
-          decltype(std::tuple_cat<std::tuple<Head>,
-                                  typename FilteredTupleReturnType<
-                                      TPred, std::tuple<Tail...>>::Result>),
-          std::tuple<Head>,
-          typename FilteredTupleReturnType<TPred,
-                                           std::tuple<Tail...>>::Result>::type,
+      typename TupleTypePrepend<
+          typename FilteredTupleReturnType<TPred, std::tuple<Tail...>>::Result,
+          Head>::Result,
       typename FilteredTupleReturnType<TPred,
                                        std::tuple<Tail...>>::Result>::type;
 };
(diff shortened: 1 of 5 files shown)


dopiera:
I managed to make g++ segfault on this PR. I shouldn't be proud of it, but I sort of am :)

google/cloud/storage/internal/tuple_filter.h
```diff
@@ -0,0 +1,146 @@
...
+template <typename T, bool Filter>
+struct TupleFilterItem {};
+
+/**
+ * Implementation of TupleFilterItem - true branch.
+ */
+template <typename T>
+struct TupleFilterItem<T, true> {
+  using Result = std::tuple<T>;
+  Result operator()(T&& t) const { return Result(std::forward<T>(t)); }
```
devjgm: FWIW, since `T` is class-level template parameter, it will be already bound when this function is called, and so the parameter type `T&&` here is not a forwarding reference, it's just a regular rvalue. And since this is just an rvalue, there's no need to use `std::forward`, we should just use `std::move` instead.

However, if instead we want this `op()` to use a forwarding reference, I think we'll need to make this member function itself a template and have it accept a parameter of type, say, `U&&`. That would be a forwarding reference so we could then use `std::forward`.

In general, I think we should only use `std::forward` with actual forwarding references, and use `std::move` otherwise. I think there are a other cases below that use `std::forward` with regular rvalues that we might want to look at again also.

google/cloud/storage/internal/tuple_filter.h
```diff
@@ -0,0 +1,146 @@
...
+ * provided type predicate.
+ *
+ * @tparam TPred a type predicate telling if an element stays or is filtered out
+ * @tparam Args the type of the tuple's elements
+ *
+ * @param tuple the tuple to filter elements from
+ */
+template <template <class> class TPred, typename... Args>
+typename FilteredTupleReturnType<TPred, std::tuple<Args...>>::Result
+StaticTupleFilter(std::tuple<Args...> t) {
```
devjgm: This parameter, `t`, is being passed by value. is that intended?

google/cloud/storage/internal/tuple_filter.h
```diff
@@ -0,0 +1,146 @@
...
+namespace google {
+namespace cloud {
+namespace storage {
+inline namespace STORAGE_CLIENT_NS {
+namespace internal {
+
+/**
+ * A helper class to filter a single element from a tuple.
+ *
+ * Depending on whether to filter the element of not, appropriate tuple type is
```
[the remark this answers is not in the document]
dopiera: Done.

google/cloud/storage/internal/tuple_filter.h
dopiera: Ooops. Fixed now.

google/cloud/storage/internal/tuple_filter.h
dopiera: It wasn't, now it should be better.

dopiera:
Compilation will keep failing until `https://github.com/googleapis/google-cloud-cpp-common/pull/85` is picked up.

Issue #3224, opened 2019-10-25
Test Dockerfiles create too many layers


coryan:
Some of the Dockerfiles in ci/kokoro/install create too many layers, and fail to complete with some Docker configurations.

One easy solution is to collapse some `RUN` commands into a single line with `&&`, that creates fewer layers.

A more complicated solution is to compile all the dependencies, make them install in `/usr/local`, and then create a new Dockefile target that just copies the contents of `/usr/local`.

The latter would scale better, but would require more changes to cache correctly.

PR #3226, opened 2019-10-25
doc: fix CentOS 7 build instructions

coryan:
The CentOS instructions did not quite work out of the box. The main
problem was using `LD_LIBRARY_PATH` and `sudo`, the latter unsets the
environment variables. The gRPC and googleapis build need tools
installed in `/usr/local/` with shared libraries in that directory. One
fix for that is to make ld.so search `/usr/local/lib{,64}/` by default.
These directories will be needed anyway, our libraries (and our
dependencies), install .so's there.

There were other minor problems, like missing `sudo` in some of the
commands that need root privileges.

Finally, I changed the Dockerfile to create fewer layers (see #3224),
otherwise I could not iterate quickly on my workstation. I just worked
around the issue by issuing multiple commands in each `RUN` line.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3226)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -437,7 +437,7 @@ directories, there are multiple ways to solve this problem, the following
 steps are one solution:
 
 ```bash
-(echo "/usr/local/lib" ; "/usr/local/lib64") | \
+(echo "/usr/local/lib" ; echo "/usr/local/lib64") | \
 sudo tee /etc/ld.so.conf.d/usrlocal.conf
 ```
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -387,7 +387,7 @@ does not search for shared libraries in these directories by default, there
 are multiple ways to solve this problem, the following steps are one solution:
 
 ```bash
-(echo "/usr/local/lib" ; "/usr/local/lib64") | \
+(echo "/usr/local/lib" ; echo "/usr/local/lib64") | \
 sudo tee /etc/ld.so.conf.d/usrlocal.conf
 ```
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -382,6 +382,15 @@ sudo zypper install --allow-downgrade -y cmake gcc gcc-c++ git gzip \
         libcurl-devel libopenssl-devel make tar wget
 ```
 
+The following steps will install libraries and tools in `/usr/local`. CentOS
+does not search for shared libraries in these directories by default, there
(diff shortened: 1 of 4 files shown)


INSTALL.md
```diff
@@ -382,6 +382,15 @@ sudo zypper install --allow-downgrade -y cmake gcc gcc-c++ git gzip \
         libcurl-devel libopenssl-devel make tar wget
 ```
 
+The following steps will install libraries and tools in `/usr/local`. CentOS
+does not search for shared libraries in these directories by default, there
+are multiple ways to solve this problem, the following steps are one solution:
+
+```bash
+(echo "/usr/local/lib" ; "/usr/local/lib64") | \
```
devjgm: Does this command work? It looks like it's trying to run `"/usr/local/lib64"` as a command. Will that work?

ci/kokoro/install/Dockerfile.opensuse-leap
```diff
@@ -66,6 +75,16 @@ RUN cmake --build cmake-out --target install -- -j ${NCPU:-4}
 RUN ldconfig
 # ```
 
+# The following steps will install libraries and tools in `/usr/local`.
+# By default openSUSE/Leap does not search for shared libraries in these
+# directories, there are multiple ways to solve this problem, the following
+# steps are one solution:
+
+# ```bash
+RUN (echo "/usr/local/lib" ; "/usr/local/lib64") | \
```
devjgm: Here too, looks like it's missing an `echo`.

INSTALL.md
```diff
@@ -422,6 +431,16 @@ sudo cmake --build cmake-out --target install -- -j ${NCPU:-4}
 sudo ldconfig
 ```
 
+The following steps will install libraries and tools in `/usr/local`.
+By default openSUSE/Leap does not search for shared libraries in these
+directories, there are multiple ways to solve this problem, the following
+steps are one solution:
+
+```bash
+(echo "/usr/local/lib" ; "/usr/local/lib64") | \
```
devjgm: Here too.

ci/kokoro/install/Dockerfile.opensuse-leap
coryan: Argh. Fixed.

INSTALL.md
coryan: Fixed too.

PR #3233, opened 2019-10-29
fix: make sure ARGs are visible in Dockerfiles

dopiera:
Whenever a `FROM` directive appears in a Dockerfile, all `ARG`s need to
be redeclared. Otherwise, they won't be propagated to the environment of
the executed commands.

For example, this problem affected `NCPU` in
`Dockerfile.ubuntu-install`, which yielded running `make -j`, which
means unbound concurrency, which made 32G not enough to compile gRPC.

I believe this is why this image is not being built and its old, cached
incarnation is being used in #3221, which leads to compilation errors.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3233)


dopiera:
Looks like grpc update changed some symbol names :/

coryan:
Ah, that explains the API breakage.  Thoughts on how to fix it? I think adding the API baseline update to this PR is Okay.

dopiera:
> Ah, that explains the API breakage. Thoughts on how to fix it? I think adding the API baseline update to this PR is Okay.

I don't think we have much choice. Without this PR, the docker image fails to build and uses the cache instead (otherwise we would have seen this error before), so we wouldn't be able to regenerate the ABI.

Perhaps for the future, we should fail if building the docker image fails.

PR #3236, opened 2019-10-30
ci: configuration files for CI generators

coryan:
These files configure the generators for `ci/kokoro/readme` and
`ci/kokoro/install`. The files on those directories are not automatically
generated just yet, but this is (I hope) the last PR before we can do
so.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3236)

### DIFF
--- a/ci/etc/publish-refdocs.sh
+++ b/ci/etc/publish-refdocs.sh
@@ -18,7 +18,7 @@ set -eu
 # This script is meant to source from ci/kokoro/docker/publish_refdocs.sh.
 
 upload_docs "google-cloud-bigtable" "${BUILD_OUTPUT}/google/cloud/bigtable/html" \
-  "cloud.tag" "${BRANCH}" "${CREDENTIALS_FILE}" "${STAGING_BUCKET}"
+  "bigtable.tag" "${BRANCH}" "${CREDENTIALS_FILE}" "${STAGING_BUCKET}"
 
 upload_docs "google-cloud-storage" "${BUILD_OUTPUT}/google/cloud/storage/html" \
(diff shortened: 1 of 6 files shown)


ci/etc/publish-refdocs.sh
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+# This script is meant to source from ci/kokoro/docker/publish_refdocs.sh.
+
+upload_docs "google-cloud-bigtable" "${BUILD_OUTPUT}/google/cloud/bigtable/html" \
+  "cloud.tag" "${BRANCH}" "${CREDENTIALS_FILE}" "${STAGING_BUCKET}"
```
tmatsuo: It should be "bigtable.tag" instead of "cloud.tag"

ci/etc/publish-refdocs.sh
```diff
@@ -0,0 +1,24 @@
...
+
+upload_docs "google-cloud-storage" "${BUILD_OUTPUT}/google/cloud/storage/html" \
+  "cloud.tag" "${BRANCH}" "${CREDENTIALS_FILE}" "${STAGING_BUCKET}"
```
tmatsuo: It should be "storage.tag" instead of "cloud.tag"

ci/etc/publish-refdocs.sh
coryan: Thanks, fixed.

ci/etc/publish-refdocs.sh
coryan: Fixed too.

PR #3246, opened 2019-10-31
ci: remove submodule builds

### LINKED ISSUE #3229: Proposal: do not test building as a submodule.
This is a proposal to remove any builds that test building `google-cloud-cpp` as a submodule of a larger CMake project. I think that use case is rare, and fraught with problems (as target names are global in a CMake project).

Building the project as part of a larger super build is a cleaner solution for both us and our users.

Thoughts?

coryan:
Submodules are difficult to support, and probably not a best practice,
as the target names must remain different from all the targets that the
user might define (CMake has a global namespace for targets).

Fixes #3229

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3246)


coryan:
Rebased to resolve conflicts. PTAL.

PR #3250, opened 2019-11-05
chore: reformat for 80 columns; fix spacing

mr-salty:
shell style guide specifies 80 columns

since `echo` adds spaces between its arguments the line that was already
split resulted in an extra space:

`Verify installed headers created only  expected directories.`

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3250)


coryan:
Consider hooking something like `shfmt` to the build.  We should have the robots do the drudgery of placing whitespace: https://github.com/mvdan/sh

PR #3254, opened 2019-11-07
cleanup: make Policy protoypes `const` and name them _prototype_

mr-salty:
This makes it clearer when *Policy objects are meant to be cloned and
not modified (`const` prevents inadvertent modification).

I had to make a couple methods `const` to make this work.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3254)

### DIFF
--- a/google/cloud/bigtable/idempotent_mutation_policy.cc
+++ b/google/cloud/bigtable/idempotent_mutation_policy.cc
@@ -30,7 +30,7 @@ std::unique_ptr<IdempotentMutationPolicy> SafeIdempotentMutationPolicy::clone()
 }
 
 bool SafeIdempotentMutationPolicy::is_idempotent(
-    google::bigtable::v2::Mutation const& m) const {
+    google::bigtable::v2::Mutation const& m) {
   if (!m.has_set_cell()) {
     return true;
   }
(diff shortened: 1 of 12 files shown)


mr-salty:
Carlos requested this on https://github.com/googleapis/google-cloud-cpp-spanner/pull/1019 (I'm also going to apply the `_prototype_` naming in that repo, I saw that was done here in some places and thought it was a good idea).

mr-salty:
___
*[google/cloud/bigtable/idempotent_mutation_policy.h, line 38 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3254#-Lt2oAnT8p7oD3ySl1KF:-Lt2oAnT8p7oD3ySl1KG:bw53fr7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/fdc6aa21ff2e7a56cbcd8ac4de1dc08009eeec5a/google/cloud/bigtable/idempotent_mutation_policy.h#L38)):*
> ```objc
> 
>   /// Return true if the mutation is idempotent.
>   virtual bool is_idempotent(google::bigtable::v2::Mutation const&) const = 0;
> ```

FYI I am a bit on the fence about this part of the change - is this actually required to be `const`?  We are treating it that way in `retry_policy.cc` since we use the policy without cloning it, so we should either:

* make `is_idempotent` const
* leave `is_idempotent` non-const and change uses of `IdempotentMutationPolicy` so we `clone()` prior to using it.

Issue #3258, opened 2019-11-11
Match build matrix in ci/kokoro/{install,readme} with g-c-cpp-common


coryan:
There are a number of builds in g-c-cpp-common that we do not have in this repo, such as the CentOS-8 build, or the Debian-Buster build.

As the build configuration in ci/kokoro/{install,readme} are automatically generated, it should be easy to make them match.

PR #3259, opened 2019-11-11
ci: fix install build for Debian-10 (Buster)

coryan:
Part of the work for #3258

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3259)


coryan:
If this looks similar to googleapis/google-cloud-cpp-common#105 it is because this is automatically generated from those changes.

PR #3262, opened 2019-11-12
fix!: avoid duplicate cluster ids in bigtable::InstanceConfig

### LINKED ISSUE #2589: Detect duplicate cluster ids in bigtable::InstanceConfig.
It is possible to pass the same cluster id more than once to the constructor to `bigtable::InstanceConfig`, which is probably almost always a bug.

We should throw an exception (only kidding), or have a separate factory function to return `StatusOr<bigtable::InstanceConfig>`.

manish-qlogic:
BREAKING CHANGE: replacing the constructor to stop allowing duplicate cluster

Fixes #2589 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3262)

### DIFF
--- a/google/cloud/bigtable/instance_config.h
+++ b/google/cloud/bigtable/instance_config.h
@@ -30,15 +30,13 @@ inline namespace BIGTABLE_CLIENT_NS {
 class InstanceConfig {
  public:
   InstanceConfig(std::string instance_id, std::string display_name,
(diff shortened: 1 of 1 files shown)


coryan:
I think you should update the API baseline, the CheckABI build is almost certainly going to break.

PR #3263, opened 2019-11-12
bug: fix error messages in resumable sessions

coryan:
The error messages reported on a resumable session failure were
incorrect or incomplete. This improves the details in the error messages
so we can better diagnose problems.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3263)


coryan:
I am expecting the Tumbleweed failures, but the code should be in good enough shape to review.

PR #3270, opened 2019-11-16
cleanup: don't set CURL_STATICLIB in libcurl super build file

### LINKED ISSUE #3269: Fix static vs. shared support in libcurl super build file.
The libcurl CMake files do not use the standard configuration parameters for static vs. shared, we probably need to set `CURL_STATICLIB` and `CURL_SHAREDLIB` based on `BUILD_SHARED_LIBS`.

remyabel:
Fixes #3269

curl now uses `BUILD_SHARED_LIBS` to set `CURL_STATICLIB`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3270)

### DIFF
--- a/super/external/curl.cmake
+++ b/super/external/curl.cmake
@@ -55,7 +55,6 @@ if (NOT TARGET curl-project)
             -DHTTP_ONLY=ON
             -DCMAKE_ENABLE_OPENSSL=ON
             -DENABLE_ARES=ON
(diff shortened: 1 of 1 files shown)


remyabel:
That macOS error looks like a [gRPC bug](https://github.com/grpc/grpc/issues/20311).

coryan:
Seems like we need to update the default (and maybe the recommended) gRPC version.

PR #3271, opened 2019-11-17
fix: remove storage class 'ARCHIVE' as this feature is still not supported.

### LINKED ISSUE #3253: Warn users that the `ARCHIVE` storage class is not GA
We should add comments to the helpers related to the `ARCHIVE` storage class saying that these are not generally available.  I think `storage_class::Archive()` and `SetStorageClassArchive()` are the main ones.

manish-qlogic:
This closes #3253

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3271)


coryan:
The changes look good, but I need to check with @frankyn because I do not recall if we wanted to wait or what.

frankyn:
We'd prefer we add comments that this feature is under a whitelist instead.

cc: @tritone is on point for Archive Storage class.

PR #3275, opened 2019-11-20
ci: use google-cloud-cpp-common from vcpkg

coryan:
Now that vcpkg has a good version of google-cloud-cpp-common, we do
not need to compile our own version.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3275)


coryan:
This won't work. As the builds stand today we need the testing libraries and those are not installed by vcpkg (nor should they).

Issue #2951, opened 2019-08-08
Use parallel streams when uploading large files.


coryan:
When uploading large files to GCS one can get better performance by running multiple uploads in parallel.  The file is divided in "parts", uploaded to a temporary object, and then the parts are reassembled using `Client::Compose`.  Then the parts are deleted using `Client::DeleteObject()` (with an specific version!).

We need to think about the names for these parts, maybe something like `<object-name-sha256>-tmp-parts/partNN`.

We should consider how these parts get deleted (object lifecycle rules?)

We should consider how we save the state of these uploads so the application can resume them. Though I presume "not resumable" is an option that would work for files that are not excessively large.

In this context "large" file is of course relative, but anything below 20MiB probably makes not sense to split.  And something like 1GiB is probably large enough that it deserves to be split.

Note that GCS cannot compose more than 1024 parts, and a single compose operation cannot compose more than 32 parts (so composing 32 parts each composed by 32 parts is the absolute limit).

Since the rest of the library is single threaded, the obvious implementation needs to create N threads to do the work.  Later we could implement a primitive that "uploads N things" at the same time using the Curl primitives for simultaneous transfers.

Also note that more than each upload uses 50MiB/s (that is about 400Mbps) and there are only 2Gpbs per core, so more than 5 uploads per core is not going to help.

dopiera:
# Requirements
Some additional (nice-to-have) requirements I thought of:
* allow for separate processes writing to streams
* a process restart/crash shouldn't imply starting from scratch

# API
For the first step, I propose the following:
```C++

// Additional options:
// Let the user chose how many uploads stream to generate
struct NumShards {};
// Indicate the upload session id to restart
struct RestoreParallelUploadSession {};

struct ParallelUploadState {
  // Obtain a session id to be used with `RestoreParallelUploadSession`
  std::string resumable_session_id() const;
  // Return the number of shards of this upload session.
  std::size_t NumShards() const;
  // Get the writer for a shard of the upload.
  StatusOr<GcsObjectWriteStream> Shard(std::size_t idx);
  // Poll for the operation to complete.
  template< class Rep, class Period >
  StatusOr<ObjectMetadata> PollForCompletion(
      const std::chrono::duration<Rep, Period>& rel_time);
  // Cleanup all the temporary files
  Status Cleanup();
};

template <typename... Options>
StatusOr<ParallelUploadState> PrepareParallelUpload(
    Client client, std::string const &bucket, std::string const &object_name,
[...]

PR #3279, opened 2019-11-25
feat: implement non-resumable parallel uploads

dopiera:
This is a part of #2951.

This is the raw API on which uploading files will be built.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3279)


coryan:
I dismissed my own review because I should not be blocking things right now.

PR #3283, opened 2019-12-03
fix: add logic to ObjectWriteStreambuf for handling jumps in upload ranges to fix #3280

### LINKED ISSUE #3280: Parallel resumable upload corrupting uploads to GCS
Yes

- Cloud Storage.

When resumable upload is used from 2 different hosts, client library is inconsistent in sending bytes upstream. Currently, next_expected_bytes is returned from the server as a response to content-range upload on resumable stream. If there are 2 clients uploading same object from 2 hosts, lagging stream gets the updated next_expected_byte. The client library does not support jump in the range bytes in the buffer, and sends incorrect byte range to the updated next_expected_byte. 

This results in - 
1. possible data corruption due to incorrect byte range sent from client. 
2. one of the stream sending more bytes than needed (since it is a jump in the next_expected_bytes, but not in the cached buffer data) resulting in inconsistent_data error from GCS. 

example scenario- for object with 1000 bytes:

1. stream 1 starts sending bytes 0-99, 100-199,.. 
2. stream 2 starts resumable uploads 500-599, 
3. stream 1 receives 503 bigtable_stale_version on 500-599
4. stream 1 uploads 500-599,600-699,700-799. 
5. from response to point 2, stream 2 receives next_expected byte 800 rather than 600. 
6. stream 2 sends (600-699) range data from its cached buffer to 800-899,
7. stream 1 is closed. 
8. stream 2 gets error inconsistent_data_size due to sending more bytes from cached buffer to GCS.

Jseph:
Fix #3280 by putting the upload stream in a bad state when there is an unexpected jump in the next expected byte from GCS.  Also prevents corruption of uploads if not all of the bytes written in Flush get committed to GCS.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3283)

### DIFF
--- a/google/cloud/storage/internal/object_streambuf.cc
+++ b/google/cloud/storage/internal/object_streambuf.cc
@@ -368,6 +368,10 @@ StatusOr<ResumableUploadResponse> ObjectWriteStreambuf::Flush() {
 
   auto chunk_count = actual_size / UploadChunkRequest::kChunkSizeQuantum;
   auto chunk_size = chunk_count * UploadChunkRequest::kChunkSizeQuantum;
+  // GCS upload returns an updated range header that sets the next expected
+  // byte. Check to make sure it remains consistent with the bytes stored in the
+  // buffer.
+  auto expected_next_byte = upload_session_->next_expected_byte() + chunk_size;
 
   hash_validator_->Update(pbase(), chunk_size);
   StatusOr<ResumableUploadResponse> result;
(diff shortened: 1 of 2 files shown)


Jseph:
I can't remember how to run tests against the test bench.  I'm not sure if I've broken them or not, depends on if the test bench is returning a correct range header for resumable uploads or not.  If the tests are failing please remind me how to run those tests on my computer.

scotthart:
```
./ci/kokoro/docker/build.sh clang-tidy
```
Is a good place to start. I'm working on verifying if that config also builds/runs the integration tests as well.

scotthart:
I patched your PR into my own branch; made the two changes I commented on and ran the clang-tidy.cfg, which does run the integration tests. Everything looks good.

Jseph:
I'm having some issues with this conventionalcommit thing.  I've never seen it before and IMO the documentation it links to isn't sufficient.  It doesn't seem to explain why the first commit needs to be a sentence while the second one isn't allowed to be a sentence.  Anyway sorry for the noise while I try and make this thing happy.

Jseph:
Fixed the narrowing conversions problems and possible loss of data warnings.

PR #3287, opened 2019-12-10
fix: GenericRequest::set_option() accepts crefs

### LINKED ISSUE #3285: GenericRequestBase::set_option() should not take rvalue (I think)
https://github.com/googleapis/google-cloud-cpp/blob/53acefa8c1730b603faa533cbd6664550d95c927/google/cloud/storage/internal/generic_request.h#L66

That call takes its argument as an rvalue, not a forwarding reference. This requires callers to pass rvalues, even though the caller may have a copyable lvalue.

An alternative would be to have `set_option` take a forwarding reference, but since we know that a copy is needed (to copy into the class's data member), taking by value is probably ideal.

If I've misunderstood why this function takes an rvalue, feel free to comment and close this issue.

dopiera:
This fixes #3285.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3287)

### DIFF
--- a/google/cloud/storage/internal/generic_request.h
+++ b/google/cloud/storage/internal/generic_request.h
@@ -63,7 +63,7 @@ class GenericRequestBase;
 template <typename Derived, typename Option>
 class GenericRequestBase<Derived, Option> {
  public:
-  Derived& set_option(Option&& p) {
+  Derived& set_option(Option p) {
     option_ = std::move(p);
     return *static_cast<Derived*>(this);
   }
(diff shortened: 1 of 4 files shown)


google/cloud/storage/internal/generic_request.h
```diff
@@ -63,11 +63,19 @@ class GenericRequestBase;
...
  public:
+  // Can't use capture by unversal references because if it were a template
+  // guarded by `std::enable_if<>` it would shadow the exact same superclass'
+  // member function.
   Derived& set_option(Option&& p) {
     option_ = std::move(p);
     return *static_cast<Derived*>(this);
   }
 
+  Derived& set_option(Option const& p) {
```
devjgm: Rather than adding an overload, how about just having a single `set_option(Option)` that accepts by value? That would be simpler and it should also fixed the linked issue. Thoughts?

PR #3299, opened 2019-12-24
Fix an error in FindProtobufWithTargets.cmake

snnn:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3299)


snnn:
At least, in the same file, the usage of Protobuf_FOUND/protobuf_FOUND should be consistent. Either always the upper case(Protobuf_FOUND), or always the lowercase(protobuf_FOUND).

coryan:
Thanks for your contribution!

PR #3303, opened 2020-01-03
feat: build Chain filter from a range

### LINKED ISSUE #3291: Add function to create chain filters from a collection
Some customers want to create filters dynamically, we should provide a function to create a `bigtable::Filter` from a range of filters, such as:

```C++
class Filter { public:
  template <typename Iterator>
  static Filter ChainFromRange(Iterator begin, Iterator end) {...}
...
};
```

coryan:
Allow application developers to build chain filters dynamically, from a
collection (or range) of filters.

Fixes #3291

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3303)

### DIFF
--- a/google/cloud/bigtable/filters_test.cc
+++ b/google/cloud/bigtable/filters_test.cc
@@ -252,6 +252,48 @@ TEST(FiltersTest, ChainOneArg) {
   EXPECT_EQ(2, chain.filters(0).cells_per_column_limit_filter());
 }
 
+/// @test Verify that `bigtable::Filter::ChainFromRange` works as expected.
+TEST(FiltersTest, ChainFromRangeMany) {
+  using F = bigtable::Filter;
+  std::vector<F> filter_collection{F::FamilyRegex("fam"), F::ColumnRegex("col"),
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/filters_test.cc
```diff
@@ -252,6 +252,48 @@ TEST(FiltersTest, ChainOneArg) {
...
+/// @test Verify that `bigtable::Filter::ChainFromRange` works as expected.
+TEST(FiltersTest, ChainFromRangeEmpty) {
+  using F = bigtable::Filter;
+  std::vector<F> filter_collection{};
+  auto filter =
+      F::ChainFromRange(filter_collection.begin(), filter_collection.end());
+  auto proto = filter.as_proto();
+  ASSERT_TRUE(proto.has_chain());
+  auto const& chain = proto.chain();
+  ASSERT_EQ(0, chain.filters_size());
```
devbww: It looks like this one only need be an `EXPECT_EQ`.

PR #3306, opened 2020-01-07
chore: update readmes for release

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3306)

### DIFF
--- a/google/cloud/bigtable/README.md
+++ b/google/cloud/bigtable/README.md
@@ -42,7 +42,12 @@ Apache 2.0; see [`LICENSE`](../../../LICENSE) for details.
 
 ## Release Notes
 
-### v1.6.x - TBD
+### v1.7.x - TBD
+
+### v1.6.x - 2020-01
+
+* feat: build Chain filter from a range (#3303)
+* feat: implement bigtable::Filter::InterleaveFromRange
 
 ### v1.5.x - 2019-12
 * feat: implement Bigtable sync vs. async benchmark (#3276)
--- a/google/cloud/bigtable/README.md
+++ b/google/cloud/bigtable/README.md
@@ -46,8 +46,7 @@ Apache 2.0; see [`LICENSE`](../../../LICENSE) for details.
 
 ### v1.6.x - 2020-01
 
-* feat: build Chain filter from a range (#3303)
-* feat: implement bigtable::Filter::InterleaveFromRange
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/README.md
```diff
@@ -42,7 +42,12 @@ Apache 2.0; see [`LICENSE`](../../../LICENSE) for details.
 
 ## Release Notes
 
-### v1.6.x - TBD
+### v1.7.x - TBD
+
+### v1.6.x - 2020-01
+
+* feat: build Chain filter from a range (#3303)
```
coryan: This two could be merged into a single comment, they are basically related, maybe "* feat: new functions to create Chain and Interleave filters from ranges of Filter objects"

google/cloud/storage/README.md
```diff
@@ -44,7 +44,14 @@ Apache 2.0; see [`LICENSE`](../../../LICENSE) for details.
 
 ## Release Notes
 
-### v1.8.x - TBD
+### v1.9.x - TBD
+
+### v1.8.x - 2020-01
+
+* feat: implement non-resumable parallel uploads (#3279)
```
coryan: I think we should advertise a feature that improves performance from ~50MiB/s to about 1300MiB/s a bit more strongly...  #3279  Maybe:

* feat: implement parallel uploads. For large files this can improve the upload bandwidth by a factor of 20. Note that in this release such uploads cannot be resumed after a program restart.

google/cloud/storage/README.md
```diff
@@ -44,7 +44,14 @@ Apache 2.0; see [`LICENSE`](../../../LICENSE) for details.
...
+* feat: support optionsRequestedPolicyVersion query parameter in GetBucketIamPolicy
+* feat: move creating prefix marker to ComposeMany (#3296)
```
coryan: This is actually a breaking change, but the merged message does not say so, see #3296 for details.

google/cloud/bigtable/README.md
devjgm: Done.

google/cloud/storage/README.md
devjgm: Done.

google/cloud/storage/README.md
devjgm: Done.

PR #3316, opened 2020-01-14
fix: handle CURLE_GOT_NOTHING as retryable

### LINKED ISSUE #3315: InsertObject() api is returning error: EasyPerform() - CURL error [52]
Affect it how? I'm seeing an error, not sure if it's user error or something else.

If the problem is with the service exposed by the `google-cloud-cpp` APIs
instead of the client libraries you may consider opening a support request
instead.  The `google-cloud-cpp` developers cannot help you troubleshoot
problems with the service itself.
Ok. But I cannot tell at this time whether or not this is the case

- Cloud Storage.

6430b38
0.11

### What compiler and version are you using?
Apple LLVM version 10.0.0 (clang-1000.11.45.5)

### What operating system and version are you using?
Mac 10.13.6

I work within a business space that's concerned with transferring large files quickly and securely to destinations such as cloud storage. We maximize upload throughput by multi-part concurrent uploads which are then assembled into the original object. 
The specific scenario that we are investigating here was a 5Gb file broken into 100Mb part files.
(So about 50 parts)

We are using client::InsertObject(...) api for each part.

Once all part files are uploaded we compose them into the original file. 
Unfortunately we never got that far.

### What did you see instead?
Near the end of the 50 part upload we are seeing
InsertObject returns error
Failed to upload part 46 of files/test/file-5G.dat: Error in non-idempotent operation InsertObjectMedia: EasyPerform() - CURL error [52]=Server returned nothing (no headers, no data) [UNKNOWN]
We see this error for 4 or 5 of the last few parts in flight.

coryan:
Getting nothing is approximately the same as getting a timeout, so we
should treat that as a kUnavailable (one of the retryable errors). The
application can use the strict idempotency policy and some preconditions
if they worry about the operation being too destructive when sent twice
to the server.

Changed the code to make it easier to review what error codes are handled
already. Fixes #3315.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3316)

### DIFF
--- a/google/cloud/storage/internal/curl_handle.cc
+++ b/google/cloud/storage/internal/curl_handle.cc
@@ -218,6 +218,10 @@ Status CurlHandle::AsStatus(CURLcode e, char const* where) {
   os << where << "() - CURL error [" << e << "]=" << curl_easy_strerror(e);
   // Map the CURLE* errors using the documentation on:
   //   https://curl.haxx.se/libcurl/c/libcurl-errors.html
+  // the error codes are listed in the same order as shown on that page, so
+  // one can quickly find out how an error code is handled. All the error codes
+  // are listed, but those that do not appear in old libcurl versions are
+  // commented out and handled by the `default:` case.
   StatusCode code;
   switch (e) {
     case CURLE_UNSUPPORTED_PROTOCOL:
@@ -233,12 +237,7 @@ Status CurlHandle::AsStatus(CURLcode e, char const* where) {
       code = StatusCode::kUnavailable;
       break;
 
-#if LIBCURL_VERSION_NUM >= 0x075100
-    case CURLE_WEIRD_SERVER_REPLY:
-      code = StatusCode::kUnknown;
-      break;
-#endif  // CURL >= 7.51.0
-
+    // missing in some older libcurl versions:   CURLE_WEIRD_SERVER_REPLY
     case CURLE_REMOTE_ACCESS_DENIED:
       code = StatusCode::kPermissionDenied;
       break;
@@ -247,7 +246,7 @@ Status CurlHandle::AsStatus(CURLcode e, char const* where) {
     case CURLE_FTP_WEIRD_PASS_REPLY:
     case CURLE_FTP_WEIRD_227_FORMAT:
     case CURLE_FTP_CANT_GET_HOST:
-    // case CURLE_HTTP2:  unavailable in old versions of libcurl
+    // missing in some older libcurl versions:   CURLE_HTTP2
     case CURLE_FTP_COULDNT_SET_TYPE:
       code = StatusCode::kUnknown;
       break;
@@ -354,10 +353,18 @@ Status CurlHandle::AsStatus(CURLcode e, char const* where) {
 
     case CURLE_SSH:
     case CURLE_SSL_SHUTDOWN_FAILED:
+      code = StatusCode::kUnknown;
+      break;
+
     case CURLE_AGAIN:
-      // Only returned by curl_easy_{recv,send}, but should not with the
-      // configuration we use for libcurl. And the recovery action is to call
-      // curl_easy_{recv,send} again, but it is too late to do that here.
+      // This looks like a good candidate for kUnavailable, but it is only
+      // returned by curl_easy_{recv,send}, and should not with the
+      // configuration we use for libcurl, and the recovery action is to call
+      // curl_easy_{recv,send} again, which is not how this return value is used
+      // (we restart the whole transfer).
+      code = StatusCode::kUnknown;
+      break;
+
     case CURLE_SSL_CRL_BADFILE:
     case CURLE_SSL_ISSUER_ERROR:
     case CURLE_FTP_PRET_FAILED:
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/curl_handle.cc
```diff
@@ -226,26 +226,36 @@ Status CurlHandle::AsStatus(CURLcode e, char const* where) {
...
     case CURLE_COULDNT_CONNECT:
-    case CURLE_WEIRD_SERVER_REPLY:
       code = StatusCode::kUnavailable;
       break;
+
+#if LIBCURL_VERSION_NUM >= 0x075100
+    case CURLE_WEIRD_SERVER_REPLY:
+      code = StatusCode::kUnknown;
+      break;
+#endif  // CURL >= 7.51.0
```
devbww: It seems that 7.51.0 would be represented as 0x073300.

But then again, why shouldn't this just be handled by ...
```
      // There are about 100 error codes, some are explicitly marked as
      // obsolete, some are not available in all libcurl versions. Use this
      // `default:` case to treat all such errors as `kUnknown`.
```

google/cloud/storage/internal/curl_handle.cc
```diff
@@ -226,26 +226,36 @@ Status CurlHandle::AsStatus(CURLcode e, char const* where) {
...
     case CURLE_REMOTE_ACCESS_DENIED:
       code = StatusCode::kPermissionDenied;
       break;
+
     case CURLE_FTP_ACCEPT_FAILED:
     case CURLE_FTP_WEIRD_PASS_REPLY:
     case CURLE_FTP_WEIRD_227_FORMAT:
     case CURLE_FTP_CANT_GET_HOST:
-    case CURLE_HTTP2:
+    // case CURLE_HTTP2:  unavailable in old versions of libcurl
```
devbww: Q: Should be conditionalize on LIBCURL_VERSION_NUM, like above, then?  Or, similarly handled by the default case?

google/cloud/storage/internal/curl_handle.cc
```diff
@@ -220,41 +220,167 @@ Status CurlHandle::AsStatus(CURLcode e, char const* where) {
...
     case CURLE_REMOTE_FILE_NOT_FOUND:
       code = StatusCode::kNotFound;
       break;
+
+    case CURLE_SSH:
+    case CURLE_SSL_SHUTDOWN_FAILED:
+    case CURLE_AGAIN:
+      // Only returned by curl_easy_{recv,send}, but should not with the
+      // configuration we use for libcurl. And the recovery action is to call
+      // curl_easy_{recv,send} again, but it is too late to do that here.
```
devbww: This comment is difficult to associate with the relevant cases cases.  Maybe the case should just be split, like some others?

google/cloud/storage/internal/curl_handle.cc
```diff
@@ -220,41 +220,167 @@ Status CurlHandle::AsStatus(CURLcode e, char const* where) {
...
+
+    /*case CURLE_HTTP_RETURNED_ERROR:*/
+    /*case CURLE_NO_CONNECTION_AVAILABLE:*/
+    /*case CURLE_SSL_PINNEDPUBKEYNOTMATCH:*/
+    /*case CURLE_SSL_INVALIDCERTSTATUS:*/
+    /*case CURLE_HTTP2_STREAM:*/
+    /*case CURLE_RECURSIVE_API_CALL:*/
+    /*case CURLE_AUTH_ERROR:*/
+    /*case CURLE_HTTP3:*/
+    /*case CURLE_QUIC_CONNECT_ERROR:*/
```
devbww: Q. Why include these?

google/cloud/storage/internal/curl_handle.cc
coryan: Good point. Let me give it another shot.

I find it hard to discover which error codes I thought about before and which ones I did not, so I listed all of them (they add new ones from time to time, more than 10 since I last looked), so I wanted to list all of them explicitly. But cannot really list "all" because some do not show up with older version.

Let me know what you think of the new approach.

google/cloud/storage/internal/curl_handle.cc
coryan: See above.

google/cloud/storage/internal/curl_handle.cc
coryan: I agree, thank you, fixed.

google/cloud/storage/internal/curl_handle.cc
coryan: See above.

coryan:
PTAL.

google/cloud/storage/internal/curl_handle.cc
```diff
@@ -218,6 +218,10 @@ Status CurlHandle::AsStatus(CURLcode e, char const* where) {
   os << where << "() - CURL error [" << e << "]=" << curl_easy_strerror(e);
   // Map the CURLE* errors using the documentation on:
   //   https://curl.haxx.se/libcurl/c/libcurl-errors.html
+  // the error codes are listed in the same order as shown on that page, so
```
devbww: s/the/The/

coryan: Fixed.

PR #3320, opened 2020-01-15
feat: implement Delimiter option for ListObjects

### LINKED ISSUE #3319: Delimiter support for ListObjects?
Objects: list API (https://cloud.google.com/storage/docs/json_api/v1/objects/list) supports the "delimiter" arg to list prefixes ending with that prefix, but I don't see this supported in the C++ SDK. Is there a way to do this?

coryan:
Somehow we completely missed this option in the initial implementation.

Fixes #3319

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3320)

### DIFF
--- a/google/cloud/storage/well_known_parameters.h
+++ b/google/cloud/storage/well_known_parameters.h
@@ -432,7 +432,14 @@ struct Prefix : public internal::WellKnownParameter<Prefix, std::string> {
 };
 
 /**
+ * Returns results in a directory-like mode.
  *
+ * Used in `Client::ListObjects` to return only those objects that do **not**
+ * contain the delimiter, unless said delimiter appears in the `Prefix`
+ * parameter (if any).
+ *
+ * @see https://cloud.google.com/storage/docs/json_api/v1/objects/list for more
+ *   information.
  */
 struct Delimiter : public internal::WellKnownParameter<Delimiter, std::string> {
   using WellKnownParameter<Delimiter, std::string>::WellKnownParameter;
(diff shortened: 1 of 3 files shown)


google/cloud/storage/well_known_parameters.h
```diff
@@ -431,6 +431,14 @@ struct Prefix : public internal::WellKnownParameter<Prefix, std::string> {
   static char const* well_known_parameter_name() { return "prefix"; }
 };
 
+/**
+ *
```
devbww: Missing documentation?

coryan: Doh! Fixed.

yasushi-saito: Can you document how a prefix distinguished from a leaf object in ObjectMetadata?

coryan: There is no difference.  GCS really does not have directories, as the documentation says:

https://cloud.google.com/storage/docs/cloud-console#_creatingfolders

> Because the Cloud Storage system **has no notion of folders**, folders created in the Cloud Console are a convenience to help you organize objects in a bucket. 

You can create an object called "foo/bar/baz" and there may or may not be any objects called  called "foo/" and/or "foo/bar/".  Or you can create an object *with text content* called "foo/" and another with text content called "foo/bar".

gsutil uses some conventions that may be relevant for your use-case:

https://cloud.google.com/storage/docs/gsutil/addlhelp/HowSubdirectoriesWork

PR #3321, opened 2020-01-15
docs: document parallel uploads design

dopiera:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3321)


dopiera:
> There are limits as to how fast you can change an object (once a second or so) in GCS. Could that be a problem?

I experimented with it. 100 updates happen without a problem with an average latency of 130ms (median 78ms), so this shouldn't be a problem.

PR #3325, opened 2020-01-16
cleanup: remove tmpnam

### LINKED ISSUE #3318: Remove the use of tmpnam()
This function yields linker warnings and simpler (namely `MakeRandomObjectName()`) alternatives can be used instead.

dopiera:
This fixes #3318.

tmpnam generates linker warnings.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3325)

### DIFF
--- a/google/cloud/storage/testing/temp_file.cc
+++ b/google/cloud/storage/testing/temp_file.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/storage/testing/temp_file.h"
+#include "google/cloud/storage/testing/random.h"
 #include "google/cloud/storage/testing/storage_integration_test.h"
 #include "google/cloud/terminate_handler.h"
 #include <cassert>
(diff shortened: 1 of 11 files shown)


google/cloud/storage/testing/temp_file.cc
```diff
@@ -27,16 +28,14 @@ namespace testing {
...
   // This is obviously racy, but there is no portable way to create a
   // uniquely-named temporary file and know its name.
-  char tmpfile_name[L_tmpnam];
-  if (!std::tmpnam(tmpfile_name)) {
-    Terminate("Failed to create a temporary file name");
-  }
-  std::ofstream f(tmpfile_name, std::ios::binary | std::ios::trunc);
+  auto generator = google::cloud::internal::MakeDefaultPRNG();
+  auto name = MakeRandomFileName(generator);
+  std::ofstream f(name, std::ios::binary | std::ios::trunc);
```
devjgm: The process's CWD may not be writable, so we should probably place `name` in an appropriate temp directory for the platform. We may be able to do this with a gtest function `testing::TempDir() + "/" + name` or something.

PR #3330, opened 2020-01-17
feat: improve storage_throughput_benchmark

coryan:
A number of small (but important) improvements

- Randomize which API is used for each iteration,
  to better compare XML vs. JSON effects.
- Make chunk-size configurable at run-time.
- Flush the log more often for quicker iterations
  during development.
- Rewrite script to generate plots from R to Python,
  which is more familiar to other developers in the
  team.
- Omit measurements that are not relevant, such
  as the time to delete objects.
- Use the Mann-Whitney U-Test to decide if there is
  a performance difference between XML and JSON.

You would typically run this with:

```
.../storage_throughput_benchmark \
    --project-id=${GOOGLE_CLOUD_PROJECT} \
    --region=${REGION_WHERE_YOUR_VM_IS_RUNNING} \
    --chunk-size=16MiB --object-size=256MiB \
    --object-count=8 --thread-count=1 --duration=60m | \
    tee ../storage.tp.$(date +%s).txt
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3330)


google/cloud/storage/benchmarks/storage_throughput_benchmark.cc
```diff
@@ -240,62 +238,63 @@ TestResult CreateOnce(gcs::Client client, std::string const& bucket_name,
...
 
+std::string CreateChunk(google::cloud::internal::DefaultPRNG& generator,
+                        std::size_t desired_size) {
+  std::string const random_data =
+      gcs_bm::MakeRandomData(generator, gcs_bm::kMiB);
+  std::string chunk;
+  while (chunk.size() < desired_size) {
+    chunk.append(random_data);
+  }
+  return chunk;
```
devbww: Should we limit to `desired size`?

google/cloud/storage/benchmarks/storage_throughput_benchmark.cc
```diff
@@ -176,45 +177,42 @@ int main(int argc, char* argv[]) {
 
 namespace {
 char const* ToString(OpType type) {
-  static char const* kOpTypeNames[] = {"READ", "WRITE", "CREATE", "DELETE",
-                                       "LAST"};
+  static char const* kOpTypeNames[] = {"READ", "WRITE", "CREATE", "LAST"};
```
devbww: Should anyone be calling these with `XXX_LAST`?

coryan: No. Removed the `_LAST` enum values and used `switch` to be able to detect any missing cases.

google/cloud/storage/benchmarks/storage_throughput_benchmark.cc
coryan: Done.

google/cloud/storage/benchmarks/storage_throughput_benchmark.cc
```diff
@@ -177,17 +177,25 @@ int main(int argc, char* argv[]) {
...
+  return "";
 }
 
 char const* ToString(ApiName api) {
-  static char const* kApiNames[] = {"JSON", "XML", "API_LAST"};
-  static_assert(API_LAST + 1 == (sizeof(kApiNames) / sizeof(kApiNames[0])),
-                "Mismatched size for OpType names array");
-  return kApiNames[api];
+  switch (api) {
+    case ApiName ::API_JSON:
```
devbww: nit: Whitespace.

google/cloud/storage/benchmarks/storage_throughput_benchmark.cc
```diff
@@ -280,9 +288,11 @@ std::string CreateChunk(google::cloud::internal::DefaultPRNG& generator,
   std::string const random_data =
       gcs_bm::MakeRandomData(generator, gcs_bm::kMiB);
   std::string chunk;
+  chunk.reserve(desired_size);
```
devbww: The `reserve()` doesn't really help if the `append()` exceeds it.

coryan: I think it still "helps", it reduces the number of allocations (and copies) from `log(desired_size / kMiB)` to **2** (the worst case you mention) or **1** (if `desired_size % kMiB == 0`, which I think it is most of the time) .

devbww: > reduces the number of allocations (and copies)

The number of copy calls, yes ... but not the amount of data copied.  The "2" case still copies O(n) data, which, I'd say, dominates the O(log n) allocations.

Not a big deal either way of course, and particularly so if "1" is the expected case.

PR #3348, opened 2020-01-24
fix: handle completed uploads in UploadChunk

### LINKED ISSUE #3342: GCS Resumable Uploads erroneously ABORTED

Yes.

- Cloud Storage.

v1.8.0

Upload objects to GCS via Resumable Upload.

Objects are uploaded successfully.

### What did you see instead?

Uploads of objects that are exact multiples of the upload chunk size are erroneously `ABORTED` and must be retried.

When uploading objects of sizes 2MiB, 3MiB, and 4MiB with an upload chunk size of 2MiB, the final chunk upload of only the 4MiB object results in the client library returning:
+ `ABORTED`
+ "Could not continue upload stream. GCS requested byte 0 which has already been uploaded."

The actual response received by the client library is a HTTP 200, which indicates that [the resumable upload is complete](https://cloud.google.com/storage/docs/performing-resumable-uploads#upload-file).

This error message originates from [this PR](https://github.com/googleapis/google-cloud-cpp/pull/3283). Perhaps there is a boundary edge-case which causes this error to be erroneously returned?

coryan:
Applications can provide the `X-Upload-Content-Length` header in their
resumable uploads. This let's GCS know how many bytes to expect, and GCS
completes an upload as soon as that many bytes arrive. In this case the
response does not have a `Range:` header indicating the last committed
byte, and these responses must be handled differently.

Fixes #3342

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3348)

### DIFF
--- a/google/cloud/storage/internal/logging_resumable_upload_session.cc
+++ b/google/cloud/storage/internal/logging_resumable_upload_session.cc
@@ -24,12 +24,12 @@ namespace internal {
 
 StatusOr<ResumableUploadResponse> LoggingResumableUploadSession::UploadChunk(
     std::string const& buffer) {
-  GCP_LOG(INFO) << __func__ << "(), buffer.size=" << buffer.size();
+  GCP_LOG(INFO) << __func__ << "() << {buffer.size=" << buffer.size() << "}";
   auto response = session_->UploadChunk(buffer);
   if (response.ok()) {
-    GCP_LOG(INFO) << __func__ << " >> payload={" << response.value() << "}";
+    GCP_LOG(INFO) << __func__ << "() >> payload={" << response.value() << "}";
   } else {
-    GCP_LOG(INFO) << __func__ << " >> status={" << response.status() << "}";
+    GCP_LOG(INFO) << __func__ << "() >> status={" << response.status() << "}";
   }
   return response;
 }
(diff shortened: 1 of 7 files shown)


google/cloud/storage/internal/logging_resumable_upload_session.cc
```diff
@@ -41,47 +41,47 @@ LoggingResumableUploadSession::UploadFinalChunk(std::string const& buffer,
...
   auto const& response = session_->session_id();
-  GCP_LOG(INFO) << __func__ << " >> " << response;
+  GCP_LOG(INFO) << __func__ << "() >> " << response;
   return response;
 }
 
 StatusOr<ResumableUploadResponse> const&
 LoggingResumableUploadSession::last_response() const {
-  GCP_LOG(INFO) << __func__ << " << ()";
+  GCP_LOG(INFO) << __func__ << "() << {}}";
```
devbww: s/}}/}/?

google/cloud/storage/tests/object_resumable_write_integration_test.cc
```diff
@@ -252,6 +252,116 @@ TEST_F(ObjectResumableWriteIntegrationTest, StreamingWriteSlow) {
...
+  ASSERT_STATUS_OK(options);
+  Client client(options->SetUploadBufferSize(chunk_size));
+
+  std::string bucket_name = flag_bucket_name;
+
+  auto const chunk = MakeRandomData(chunk_size);
+
+  for (auto const desired_size : {2 * MiB, 3 * MiB, 4 * MiB}) {
+    auto object_name = MakeRandomObjectName();
+    SCOPED_TRACE("Testing with desired_size = " + std::to_string(desired_size) +
```
devbww: s/ = /=/ perhaps?

google/cloud/storage/tests/object_resumable_write_integration_test.cc
```diff
@@ -252,6 +252,116 @@ TEST_F(ObjectResumableWriteIntegrationTest, StreamingWriteSlow) {
...
+        CustomHeader("X-Upload-Content-Length", std::to_string(desired_size)));
+    auto offset = 0L;
+    while (offset < desired_size) {
+      auto const n = (std::min)(desired_size - offset, chunk_size);
+      os.write(chunk.data(), n);
+      ASSERT_FALSE(os.bad());
+      offset += n;
+    }
+
+    // This operation should fail because the object already exists.
```
devbww: Q. Which part of the following indicates the failure?  Indeed, why should the random object exist?

coryan: That was a cut&paste error, fixed here and elsewhere.

google/cloud/storage/tests/object_resumable_write_integration_test.cc
coryan: Better, thanks.

google/cloud/storage/internal/logging_resumable_upload_session.cc
coryan: Fixed.

PR #3349, opened 2020-01-24
fix!: insert logging layer only if requested

coryan:
**BREAKING CHANGE**: the `google::cloud::storage::Client` constructor
now calls `RawClient::client_options()`, which may need changes in how
you initialize any Mocks for this class.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3349)


google/cloud/storage/client_write_object_test.cc
```diff
@@ -149,6 +149,7 @@ TEST_F(WriteObjectTest, WriteObjectPermanentSessionFailurePropagates) {
   EXPECT_CALL(*mock, CreateResumableSession(_)).WillOnce(Invoke(returner));
   EXPECT_CALL(*mock_session, UploadChunk(_))
       .WillRepeatedly(Return(PermanentError()));
+  EXPECT_CALL(*mock_session, next_expected_byte()).WillRepeatedly(Return(0));
```
devbww: Unrelated?

coryan: Yes, will fix separately.

PR #3359, opened 2020-01-30
chore: simpler include guard style

devjgm:
Changes:
1. Updated `check-include-guards.gawk` to look for `_H$` in include guards
2. Updated the style guide to mention this difference from the GSG
3. Updated all headers to use new include guard naming. Also updated file templates in `*.h.in` files.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3359)


devjgm:
Reverted the enum change. Will do that in a separate PR. Re-running builds.

PR #3366, opened 2020-02-04
chore: upgrade to google-cloud-cpp-common v0.18.0

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3366)


devbww:
As expected, the Windows build is failing.

coryan:
As to the `Kokoro Ninja` breakage. Looks like the changes in `ci/kokoro/docker/Dockerfile.ubuntu-install` did not take effect.  That can happen if the build fails to refresh the Docker image, it then falls back on the previous version of the Docker image which in this case is not good.

A rebuild may work, if it does not, we should fix the `ci/kokoro/docker/build.sh` because I think it fails to dump the log on error as it should

coryan:
I created #3367 to fix the logging for any future problems like this.

PR #3377, opened 2020-02-09
ci: pin python modules used in integration tests

coryan:
The builds broke when they picked up a new version of the Wekrzeug
module. In particular the changes around DispatcherMiddleware listed
here:

https://werkzeug.palletsprojects.com/en/master/changes/#version-0-15-0

I updated the testbench to use the new version, and then pinned the
builds to use that version, which should keep us from breaking in a
similar way.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3377)


coryan:
I used the additional tracing that @devbww added in #3376 and was able to fix the build problem. PTAL.

PR #3402, opened 2020-02-22
chore: upgrade to Python 3

### LINKED ISSUE #3056: Migrate storage testbench to Python 3

coryan:
Python 2.7 went EOL on 2020-01-01. The builds are starting to fail
because some of our dependencies assume (with good reason I think)
they are working with Python 3.

Fixes #3056

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3402)

### DIFF
--- a/ci/kokoro/docker/Dockerfile.fedora
+++ b/ci/kokoro/docker/Dockerfile.fedora
@@ -36,15 +36,16 @@ RUN dnf makecache && dnf install -y \
     ncurses-term \
     openssl-devel \
     pkgconfig \
-    python \
+    python3 \
+    python3-devel \
     shtool \
     unzip \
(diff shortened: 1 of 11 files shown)


coryan:
Incidentally I fixed the nightly build, it was failing because our CMEK was in the `global` namespace, and it seems that CMEK keys now have to match the location of the bucket:

https://cloud.google.com/storage/docs/encryption/customer-managed-keys#restrictions

coryan:
/cc: @crwilcox because you were interested in this.

crwilcox:
Thanks for picking this up and driving this in!

PR #3407, opened 2020-02-24
cleanup: simplify Curl*Request implementation

### LINKED ISSUE #3393: Simplify storage::internal::{CurlRequest,CurlDownloadRequest}
Both these classes have complicated move and assignment operators because they "need" to call `ResetOptions()` after each move. This is needed because a `CurlHandle()` passes the address of a `std::function<...>` to the `extern "C"` callbacks required by libcurl.

But it turns out this is not really needed. The callbacks are unused until `MakeRequest` is called, so we could delay setting these callbacks until that point, and therefore avoid the need to call `ResetOptions`. That would simplify the code (I think).

coryan:
We do not need to setup the libcurl callbacks until the request is made, and
therefore we can avoid having custom move and assignment operators. As an
additional (but probably trivial) improvement, the calls through function
pointers are more efficient than through `std::function<>`.

Fixes #3393 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3407)


google/cloud/storage/internal/curl_download_request.cc
```diff
@@ -28,6 +28,20 @@ namespace storage {
 inline namespace STORAGE_CLIENT_NS {
 namespace internal {
 
+extern "C" std::size_t CurlDownloadRequestWrite(char* ptr, size_t size,
```
devbww: Nit: It is a little strange that a function callable from C is marked as returning a C++ standard type.

google/cloud/storage/internal/curl_request.h
```diff
@@ -25,73 +25,50 @@ namespace cloud {
...
-        user_agent_(std::move(rhs.user_agent_)),
-        response_payload_(std::move(rhs.response_payload_)),
-        received_headers_(std::move(rhs.received_headers_)),
-        logging_enabled_(rhs.logging_enabled_),
-        socket_options_(rhs.socket_options_),
-        handle_(std::move(rhs.handle_)),
-        factory_(std::move(rhs.factory_)) {
-    ResetOptions();
+    if (!factory_) return;
+    factory_->CleanupHandle(std::move(handle_));
```
devbww: Nit: Perhaps `if (factory_) factory_->CleanupHandle(std::move(handle_));` might be clearer?

google/cloud/storage/internal/curl_download_request.cc
coryan: Fixed.

google/cloud/storage/internal/curl_request.h
coryan: Fixed.

PR #3412, opened 2020-02-27
ci: use the same Bazel program for all invocations

coryan:
A few places used `${BAZEL_BIN}` while others used a naked `bazel`. When those two versions are incompatible the build can break. Also cleaned up some unused code and removed an IAM policy binding that was leaked.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3412)


coryan:
The java.log and invocation_id code was an attempt to get per-test reports in Kokoro, I never got it to work consistently, when we try again these would be distractions at best.

PR #3419, opened 2020-03-03
test: snapshot of V4 signature conformance tests

dopiera:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3419)


manish-qlogic:
FWIW,

`expectedCanonicalRequest` key for test case with description "Vary expiration and timestamp" is missing an extra `\n` before the last occurrence of `host`. This point is raised with @frankyn.

May be this will lead to change in `v4_signatures.json` file.

dopiera:
> `expectedCanonicalRequest` key for test case with description "Vary expiration and timestamp" is missing an extra `\n` before the last occurrence of `host`. This point is raised with @frankyn.
> 
> May be this will lead to change in `v4_signatures.json` file.

Good to know, thanks! The plan was to sync it to what is exactly in the conformance tests repo. I think I should update it here once it gets updated there.

PR #3428, opened 2020-03-04
fix: use BUILD_ALWAYS in super build

coryan:
Sometimes we want to use a super build for development (or users might
use a super build for exploration), without BUILD_ALWAYS the code is not
compiled after a change.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3428)

### DIFF
--- a/super/CMakeLists.txt
+++ b/super/CMakeLists.txt
@@ -42,7 +42,10 @@ include(ExternalProject)
 ExternalProject_Add(
     google_cloud_cpp_project
     DEPENDS ${GOOGLE_CLOUD_CPP_DEPENDENCIES_LIST}
-    EXCLUDE_FROM_ALL OFF
+    EXCLUDE_FROM_ALL
+        OFF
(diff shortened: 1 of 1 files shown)


super/CMakeLists.txt
```diff
@@ -42,7 +42,10 @@ include(ExternalProject)
 ExternalProject_Add(
     google_cloud_cpp_project
     DEPENDS ${GOOGLE_CLOUD_CPP_DEPENDENCIES_LIST}
-    EXCLUDE_FROM_ALL OFF
+    EXCLUDE_FROM_ALL
+        OFF
+        BUILD_ALWAYS
+        1
```
devbww: Nit: Is that the desired formatting?

coryan: No, but it is the formatting that cmake-format creates. I try to create a separate PR that teaches the robots to do better.

PR #3456, opened 2020-03-09
chore: update google-cloud-cpp-common to v0.22.0

coryan:
This updates the version of `g-c-cpp-common` used for the CI builds, for
Bazel, and for the CMake super build. It also updates the README and
INSTALL files (which required some fixes to the markdown generator).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3456)


coryan:
This won't work, need to re-release `-common`.

Issue #3329, opened 2020-01-17
V4 Signed Urls formatting fixes


coryan:
To increase compatibility with S3 we (maybe) need to make some changes to the code. For each of these either write a test to show that the code is already, or create a separate bug to fix it:

- [x] query parameters should be sorted by Unicode code point, *not* by case insensitive ordering.  Note that C++ typically uses UTF-8, so sorting by bytes should work (see *sorting order* in [UTF-8](https://en.wikipedia.org/wiki/UTF-8))
- [x] Consecutive whitespace characters should be replaced by a single space 
- [x] Query parameters should be URL encoded only once.
- [x] The signed header list must be sorted
- [x] Require that any header starting with `x-goog-` (or `x-amz-`) *is* signed (aka sensitive headers must always be signed).
- [x] If the user provides a `x-goog-content-sha256` header (or `x-amz-content-sha256`) then use that value instead of `UNSIGNED-PAYLOAD` in the signature algorithm.

manish-qlogic:
Conformance test at following location needs to be corrected

https://github.com/googleapis/conformance-tests/blob/1efa0ca5bffd4373c56a69bf73cd3161f0eeae89/storage/v1/v4_signatures.json#L12

The value of `expectedStringToSign` and `expectedCanonicalRequest` are swapped.

Following PR will track this point
https://github.com/googleapis/conformance-tests/pull/23

manish-qlogic:
@coryan 

>Require that any header starting with x-goog- (or x-amz-) is signed (aka sensitive headers must always be signed).

is it fine to interpret  "_sensitive headers must always be signed_" as "_sensitive headers must be part of signed header list_".?

GCS C++ client always keep extension headers in signed header list. So if above interpretation is correct then we do not have any additional work on this task.

coryan:
> @coryan
> 
> > Require that any header starting with x-goog- (or x-amz-) is signed (aka sensitive headers must always be signed).
> 
> is it fine to interpret "_sensitive headers must always be signed_" as "_sensitive headers must be part of signed header list_".?
> 
> GCS C++ client always keep extension headers in signed header list. So if above interpretation is correct then we do not have any additional work on this task.

I think you are correct. I would phrase it slightly different: in the C++ client library it is impossible to have a header that is *not* signed. Therefore this is trivially true for the client library and we can close this bug.

PR #3460, opened 2020-03-11
feat: implement virtual hostname V4 signatures

### LINKED ISSUE #3327: V4 Signed URLs support virtual hosted-style URLs
We want `Client::SignUrlV4()` to (**optionally**) create URLs of the form `https://[BUCKET_NAME].storage.googleapis.com/...` as opposed to (the only option today) `https://storage.googleapis.com/[BUCKET_NAME]/...`.

dopiera:
This fixes #3327.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3460)

### DIFF
--- a/google/cloud/storage/internal/signed_url_requests.cc
+++ b/google/cloud/storage/internal/signed_url_requests.cc
@@ -139,8 +139,12 @@ std::string TrimHeaderValue(std::string const& value) {
 void V4SignUrlRequest::AddMissingRequiredHeaders() {
   auto const& headers = common_request_.extension_headers();
   if (headers.find("host") == headers.end()) {
-    SetOption(AddExtensionHeader("host", "storage.googleapis.com"));
+    SetOption(AddExtensionHeader("host", Hostname()));
   }
+}
+
+void V4SignUrlRequest::SetOption(VirtualHostname const& hostname) {
+  virtual_host_name_ = hostname.has_value() && hostname.value();
 }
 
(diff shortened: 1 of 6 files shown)


google/cloud/storage/internal/signed_url_requests.cc
```diff
@@ -185,6 +191,35 @@ std::string V4SignUrlRequest::StringToSign(std::string const& client_id) const {
...
+             : Status(StatusCode::kInvalidArgument,
+                      "specified 'host' (" + host_it->second +
+                          ") header stands in conflict with "
+                          "'VirtualHostname' option.");
+}
+
+std::string V4SignUrlRequest::Hostname() {
+  return virtual_host_name_
+             ? (common_request_.bucket_name() + ".storage.googleapis.com")
+             : std::string("storage.googleapis.com");
```
manish-qlogic: issue #3329 emphasizes on compatibility with S3. 
I think, setting up host name like this will cause interoperability issues.
What do you think?

coryan: > I think, setting up host name like this will cause interoperability issues.

Can you clarify?  What incompatibility are you thinking about?

manish-qlogic: as we do not allow any other `hostname`, we cannot set it something like `http://s3.amazonaws.com/bucket/`.

Although, I am not sure if this comes under interoperability.

PR #3462, opened 2020-03-11
chore: update cpp-cmakefile version

tmatsuo:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3462)


tmatsuo:
@coryan Thanks, changed the bazel files too. PTAL

PR #3463, opened 2020-03-11
feat: allow domain named buckets in signed URLs v4

### LINKED ISSUE #3328: V4 Signed URLs support bucket-bound hostnames
GCS supports bucket-bound hostnames:

https://cloud.google.com/storage/docs/request-endpoints#cname

That is, the owners of a domain, say `gcs-subdomain.my.domain.tld`, can create a bucket with the name of this domain (`gcs-subdomain.my.domain.tld`) and they can have GCS serve data from that bucket.

We want `Client::SignUrlV4()` to (**optionally**) create URLs of the form `http://[BUCKET_NAME]/` **or** `https://[BUCKET_NAME]` when the caller explicitly requests such names.

dopiera:
This fixes #6874.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3463)

### DIFF
--- a/google/cloud/storage/internal/signed_url_requests.cc
+++ b/google/cloud/storage/internal/signed_url_requests.cc
@@ -147,6 +147,21 @@ void V4SignUrlRequest::SetOption(VirtualHostname const& hostname) {
   virtual_host_name_ = hostname.has_value() && hostname.value();
 }
 
+void V4SignUrlRequest::SetOption(BucketBoundHostname const& o) {
+  if (!o.has_value()) {
+    domain_named_bucket_.reset();
+    return;
+  }
+  domain_named_bucket_ = o.value();
+}
+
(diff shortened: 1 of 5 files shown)


google/cloud/storage/internal/signed_url_requests.cc
```diff
@@ -147,6 +147,14 @@ void V4SignUrlRequest::SetOption(VirtualHostname const& hostname) {
   virtual_host_name_ = hostname.has_value() && hostname.value();
 }
 
+void V4SignUrlRequest::SetOption(DomainNamedBucket const& o) {
```
frankyn: can this follow the expected name `BucketBoundHostname`

coryan: Both mean nothing to me, so 🤷‍♂ .  @dopiera can you modify this PR to include that change or would a future PR be better?

Issue #3489, opened 2020-03-22
Use build system to drive integration tests and samples


coryan:
In `g-c-pp-spanner` and `g-c-cpp-pubsub` we use the build system (either CMake/CTest and/or Bazel) to drive the integration tests and examples. This makes it easier to add new integration tests, as one does not need to remember to add the test to some shell script.  Likewise, the examples are easier to implement, as only one `.cc` file is modified.

As an additional benefit, the tests execution can be parallelized, which reduces the overall CI cycle time.

I think we would do this in stages:

- [x] Change the integration test configuration to support better named environment variables.
- [x] Cleanup the Bigtable integration tests to require no parameters.
- [x] Cleanup the Storage integration tests to require no parameters.
- [x] ~For each example program, create a new `auto` example that runs all the examples in the required order.~ This is now tracked in #3523 and #3524
- [x] ~Change the Bigtable scripts to run this example instead of each individual example.~ This is now tracked in #3528
- [x] ~Change the Storage scripts to run this example instead of each individual example.~ This is now tracked in #3528 
- [x] Change the build scripts to:
  - [x] Exclude any test labeled `integration-test` from the unit test builds.
  - [x] Change all the integration tests and examples to be labeled `integration-test`.
- [x] The CMake build scripts run the integration tests against emulators. We need to change them to:
  - [x] Start the emulators and leave them running.
  - [x] Use CTest to run the tests labeled `integration-tests`
  - [x] Shutdown the emulators.
- [x] The Bazel build scripts only run the integration tests for one build, and only against production.
       We need to change them to use `Bazel` to run the tests labeled `integration-tests` passing all necessary environment variables.
- [ ] Remove the old scripts to run tests.
- [ ] Cleanup the old environment variables from the integration test configuration.

coryan:
Planning to cleanup the storage integration tests command-line.

coryan:
I had forgotten about #441. It might still be relevant.

coryan:
I think the bugs in https://github.com/googleapis/google-cloud-cpp/issues/3523 capture all the issues here.

Closing this issue as it has become redundant.

PR #3495, opened 2020-03-24
cleanup: refactor samples to C++ driver

coryan:
Part of the work for #3489

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3495)

### DIFF
--- a/google/cloud/bigtable/examples/bigtable_grpc_credentials.cc
+++ b/google/cloud/bigtable/examples/bigtable_grpc_credentials.cc
@@ -135,8 +135,8 @@ void GCECredentials(std::vector<std::string> argv) {
 void RunAll(std::vector<std::string> argv) {
   if (!argv.empty()) throw Usage{"auto"};
   for (auto const& var :
-      {"GOOGLE_CLOUD_PROJECT", "GOOGLE_CLOUD_CPP_BIGTABLE_TEST_INSTANCE_ID",
-       "ACCESS_TOKEN", "GOOGLE_APPLICATION_CREDENTIALS"}) {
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/examples/bigtable_grpc_credentials.cc
```diff
@@ -23,34 +22,20 @@ struct Usage {
...
-  argc--;
-  return result;
-}
-
 std::string command_usage;
 
-void PrintUsage(int, char* argv[], std::string const& msg) {
-  std::string const cmd = argv[0];
+void PrintUsage(std::string const& cmd, std::string const& msg) {
   auto last_slash = std::string(cmd).find_last_of('/');
```
devbww: Opt: The `std::string()` around `cmd` looks like a confusing remnant.

google/cloud/bigtable/examples/bigtable_grpc_credentials.cc
```diff
@@ -152,18 +129,50 @@ void GCECredentials(int argc, char* argv[]) {
...
   }
   //! [test gce credentials]
-  (std::move(project_id), std::move(instance_id));
+  (argv[0], argv[1]);
+}
+
+void RunAll(std::vector<std::string> argv) {
+  if (!argv.empty()) throw Usage{"auto"};
+
+  auto const project_id =
```
devbww: Opt Nit: Might it be clearer to run the checking `for` loop first, and then make these just `*GetEnv()` calls?

coryan: Done.

google/cloud/bigtable/examples/bigtable_grpc_credentials.cc
coryan: Fixed.

Issue #441, opened 2018-04-16
Troubleshoot problems with SampleRowKeys()


coryan:
On Fedora (and only on Fedora) the regression tests freeze from time to time.  No responses are recorded in the test.

coryan:
Well, I just saw it freeze on Ubuntu, which I guess is good news because it means this is not platform specific?  Note that this could be in gRPC or in the emulator or in the client library.

coryan:
Here is a example of a frozen build:

https://travis-ci.org/coryan/google-cloud-cpp/jobs/368372069

hemant-qlogic:
I also got this problem while running tests on ubuntu-16.04
Here is the link
https://travis-ci.org/GoogleCloudPlatform/google-cloud-cpp/jobs/372332188

DanielMahu:
Some signs point to an emulator bug. I've run this locally with some added arguments:
```
  --gtest_filter='*SampleRow*' \
    --gtest_repeat=1000 --gtest_break_on_failure
```
and after 206 iterations of the test, output stopped and the emulator window had a crash with the message `fatal error: concurrent map iteration and map write`. I can take a look at this.

@hemant-qlogic please file a different bug for the unit test, it's a separate issue.

coryan:
@garye who has implemented most of the emulator and might want to hear about it.

garye:
Can you point me to a test that is failing?

hemant-qlogic:
Added new issue for table_admin_test no response case in #470

coryan:
@garye the specific test that breaks is here:

https://github.com/GoogleCloudPlatform/google-cloud-cpp/blob/master/bigtable/tests/data_integration_test.cc#L549

The test is running against the latest gcloud emulator, we download the latest and greatest version before running the integration tests.  I will let the folks actually working on this bug to speak about results with a local build.

DanielMahu:
The older emulator version on my laptop produced a crash dump with a stack trace inside hashmap code, which is pasted at the end of this comment, since it's no longer relevant: the implementation was changed meanwhile to btree in https://github.com/GoogleCloudPlatform/google-cloud-go/commit/781cba7caa1b7524afce83ae96e255d4d7a8ee69, which seems to have fixed the concurrency bug.

[...]

PR #3502, opened 2020-03-25
ci: use CTest to drive Bigtable integration tests

coryan:
For the CMake-based builds use CTest to run the Bigtable integration
tests. This allows us to parallelize the tests. This PR is a really
small improvement, we also need to parallelize the examples, and the
benchmark smoke test. But it serves as a good prototype.

I removed the workarounds for #441: I ran the test 100 times, with 4 cores
and the emulator no longer seems to crash. I will mark #441 as fixed separately.

Part of the work for #3489 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3502)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -64,7 +64,7 @@ echo "================================================================"
 echo "$(date -u): Compiling and running unit tests"
 echo "================================================================"
 "${BAZEL_BIN}" test \
-    "${bazel_args[@]}" \
+    "${bazel_args[@]}" "--test_tag_filters=-integration-tests" \
     -- //google/cloud/...:all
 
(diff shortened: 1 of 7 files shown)


mr-salty:
___
*[google/cloud/bigtable/ci/run_integration_tests.sh, line 40 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3502#-M3I-H7k-yCoK1rbqC0t:-M3I10Na36NBee_mufti:bf7ysw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b15f8c804cbb88efe4399cb40dd716ce14e825b2/google/cloud/bigtable/ci/run_integration_tests.sh#L40)):*

we don't need this loop or the TODO(#441) anymore
</blockquote></details>

meant this to have a '?' after it...

PR #3509, opened 2020-03-26
ci: speed up the PR builds for Windows+CMake

coryan:
Use a Debug build for faster PR builds, a future change will enable
a Release build for post-submit builds. Also parallelize the unit
tests it saves maybe 30 seconds, but I will take it. Prepare the
ground work for running the integration tests for this build too.

Part of the work for #3489

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3509)


coryan:
I think this takes us from about 45m (sometimes 50m, sometimes 70m!) to about 35m. Not enough, but I will take whatever I can get.

PR #3515, opened 2020-03-27
ci: use Bazel cache for Windows builds

coryan:
Makes the build faster, from about 30m to below 15m

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3515)


coryan:
Ping.

PR #3599, opened 2020-03-28
ci: enable caching for macOS builds

### LINKED ISSUE #3520: Enable CMake Cache for ci/kokoro/macos

coryan:
To make ccache usable in the CMake build we need to enable it through
the super build config files. The support was almost there, but needed
multiple tweaks.

I also had to look at the output from the builds multiple times so I
changed all the messages to print the UTC time at the beginning and to
be color coded.

Finally, avoid downloading files to $HOME and $PROJECT_ROOT, and silence
some of the more noisy commands (say installing Bazel)

Fixes #3520 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3599)


ci/kokoro/macos/upload-cache.sh
```diff
@@ -0,0 +1,64 @@
...
+  echo "Service account for cache access is not configured."
+  echo "No attempt will be made to download the cache, exit with success."
+  exit 0
+fi
+
+if [[ "${KOKORO_JOB_TYPE:-}" == "PRESUBMIT_GERRIT_ON_BORG" ]] || \
+   [[ "${KOKORO_JOB_TYPE:-}" == "PRESUBMIT_GITHUB" ]]; then
+  echo "================================================================"
+  echo "This is a presubmit build, cache will not be updated, exist with success."
+# TODO(coryan) - warm up the cache on first PR  exit 0
```
devjgm: Is that exit supposed to be commented out like it it's?

coryan: No. The first time I send a PR for a cached build I need to warm up the cache *event on PRs*, then I remove this comment before merging so future PRs do not dirty up the cache.

devjgm: Woops, I missed that this was a draft earlier.

Issue #3523, opened 2020-03-27
Cleanup Examples for Bigtable


devjgm:
- [x] Refactor command-line parsing and auto-running to a common library.
- [x] Refactor bigtable_grpc_credentials.cc
- [x] Refactor data_snippets.cc
- [x] Refactor data_async_snippets.cc
- [x] Refactor bigtable_hello_app_profile.cc
- [x] Refactor bigtable_hello_instance_admin.cc
- [x] Refactor bigtable_hello_table_admin.cc
- [x] Refactor bigtable_hello_world.cc
- [x] Refactor bigtable_instance_admin_snippets.cc
- [x] Refactor bigtable_quickstart.cc
- [x] Refactor instance_admin_async_snippets.cc
- [x] Refactor table_admin_async_snippets.cc
- [x] Refactor table_admin_snippets.cc

coryan:
Broke this down into smaller issues, closing.

PR #3617, opened 2020-03-31
refactor: common code for bigtable examples

coryan:
We have too much code duplication between the examples, and as I am
refactoring all the code it seemed like a good idea to clean this up.

Part of the work for #3523

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3617)


google/cloud/bigtable/examples/bigtable_examples_common.h
```diff
@@ -0,0 +1,55 @@
...
+#include <vector>
+
+namespace google {
+namespace cloud {
+namespace bigtable {
+namespace examples {
+
+class Usage : public std::runtime_error {
+ public:
+  explicit Usage(std::string const& msg) : std::runtime_error(msg.c_str()) {}
```
devbww: There is a `std::runtime_error(std::string const&)` ctor.

google/cloud/bigtable/examples/bigtable_examples_common.cc
```diff
@@ -0,0 +1,93 @@
...
+namespace google {
+namespace cloud {
+namespace bigtable {
+namespace examples {
+
+Example::Example(std::map<std::string, CommandType> commands)
+    : commands_(std::move(commands)) {
+  std::string full_usage;
+  // Force each command to generate its Usage string, so we can provide a good
+  // usage string for the whole program. We need to create an Table
```
devbww: s/an Table/a Table/

google/cloud/bigtable/examples/bigtable_examples_common.cc
```diff
@@ -0,0 +1,93 @@
...
+  // object to do this, but that object is never used, it is passed to the
```
devbww: s/never used, it is/never used. It is/

google/cloud/bigtable/examples/bigtable_examples_common.cc
```diff
@@ -0,0 +1,93 @@
...
+      kv.second({});
+    } catch (Usage const& u) {
+      full_usage += "    ";
+      full_usage += u.what();
+      full_usage += "\n";
+    } catch (...) {
+      // ignore other exceptions.
+    }
+  }
+  full_usage_ = std::move(full_usage);
```
devbww: Q. Why not just build within `full_usage_` in the first place?

google/cloud/bigtable/examples/bigtable_examples_common_test.cc
```diff
@@ -0,0 +1,143 @@
...
+      {"test",
+       [&](std::vector<std::string> const& args) {
+         ++test_calls;
+         if (args.empty()) throw Usage("test-usage");
+         ASSERT_EQ(2, args.size());
+         EXPECT_EQ("a0", args[0]);
+         EXPECT_EQ("a1", args[1]);
+       }},
+  });
+  char argv0[] = "argv0";
```
devbww: Perhaps `Run()` should take a `char const* argv[]`?

coryan: I am planning to pass the arguments from `main()` directly into this function, so that would not work, right?

google/cloud/bigtable/examples/bigtable_examples_common.cc
coryan: Done.

google/cloud/bigtable/examples/bigtable_examples_common.cc
coryan: Ditto.

google/cloud/bigtable/examples/bigtable_examples_common.cc
coryan: Remove the whole comment, it was a cut&paste error.

google/cloud/bigtable/examples/bigtable_examples_common.h
coryan: Doh! Fixed.

google/cloud/bigtable/examples/bigtable_examples_common_test.cc
devbww: I believe that if you change `Run(int argc, char* argv[])` to `Run(int argc, char const* const* argv)` that you be able to ...
```
  char const* args[] = {"1", "2", "3"};
  Run(3, args);
```
and
```
int main(int argc, char* argv[]) {
  Run(argc, argv);
}
```

google/cloud/bigtable/examples/bigtable_examples_common.cc
```diff
@@ -22,24 +22,19 @@ namespace examples {
 
 Example::Example(std::map<std::string, CommandType> commands)
     : commands_(std::move(commands)) {
-  std::string full_usage;
+  full_usage_.clear();
```
devbww: Typically we would choose not to do this redundant step in the constructor, I think.

google/cloud/bigtable/examples/bigtable_examples_common_test.cc
coryan: I will send a follow up PR to fix that.

google/cloud/bigtable/examples/bigtable_examples_common.cc
coryan: Will fix too.

PR #3648, opened 2020-04-02
ci: coverage build runs some tests against production

### LINKED ISSUE #3542: Make sure the coverage build can run against production
Make sure the coverage build can run against production because that is what we do for -pubsub and -spanner

### LINKED ISSUE #3547: Make sure we have all the configuration parameters in environment variables
Make sure we have all the configuration parameters in environment variables and there are no conflicts in env var names

coryan:
The main goal in this change is to run some tests that can *only*
run in production during the code coverage builds. This is important
because we have no visibility into their coverage, and it will be
even more important when we import `-spanner` which only
runs integration tests against production.

Move the configuration parameters for integration tests to git.
Easier to track and change them if they are part of the code, and
there are no secrets in this configuration.

More importantly, soon we will want to run all kinds of tests against
production in this build, e.g., the Cloud Spanner and Cloud Pub/Sub
tests.

To make it easier to identify which tests are skipping production 
and/or the emulator I used the GTEST_SKIP() annotations more
consistently.

Fixes #3547 and fixes #3542.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3648)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -186,16 +186,17 @@ if [[ "${BUILD_TESTING:-}" = "yes" ]]; then
   readonly GOOGLE_CLOUD_CPP_STORAGE_TEST_KEY_FILE_JSON="/c/service-account.json"
   readonly GOOGLE_CLOUD_CPP_STORAGE_TEST_KEY_FILE_P12="/c/service-account.p12"
   readonly GOOGLE_APPLICATION_CREDENTIALS="/c/service-account.json"
+
   if [[ # yes: always try to run the integration tests
-        "${RUN_INTEGRATION_TESTS}" == "yes" || \
-        # auto: only try to run integration tests if the config files are present
-        ( "${RUN_INTEGRATION_TESTS}" == "auto" && \
-          -r "${INTEGRATION_TESTS_CONFIG}" && \
-          -r "${GOOGLE_APPLICATION_CREDENTIALS}" && \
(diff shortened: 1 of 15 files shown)


ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -182,6 +182,43 @@ if [[ "${BUILD_TESTING:-}" = "yes" ]]; then
...
+  if [[ # yes: always try to run the integration tests
+        "${RUN_INTEGRATION_TESTS}" == "yes" || \
+        # auto: only try to run integration tests if the config files are present
+        ( "${RUN_INTEGRATION_TESTS}" == "auto" && \
+          -r "${INTEGRATION_TESTS_CONFIG}" && \
+          -r "${GOOGLE_APPLICATION_CREDENTIALS}" && \
+          -r "${GOOGLE_CLOUD_CPP_STORAGE_TEST_KEY_FILE_JSON}" && \
+          -r "${GOOGLE_CLOUD_CPP_STORAGE_TEST_KEY_FILE_P12}" ) ]] && \
+      # super builds cannot run the integration tests
+      [[ "${SOURCE_DIR}" != "super" ]]; then
```
devbww, line 198: Nit: It is a little strange to say ...
```
if [[ A || B ]] && [[ C ]]; then
```
instead of ...
```
if [[ ( A || B ) && C ]]; then
```

coryan: Fixed (I hope).

Issue #3635, opened 2020-04-01
Implement escaping in V4 POST policies


dopiera:
Policy should be encoded as defined in
https://docs.google.com/document/d/1XYkqJ_pPazkg6NuhRUEl_cReUtzX1DbgvF_jbPzgzog/edit#heading=h.rze9ynrxgv37

frankyn:
@dopiera I believe this can be closed. Is that correct?

coryan:
Yes.

PR #3687, opened 2020-04-03
feat: implement POST policy escaping

### LINKED ISSUE #3326: V4 Signed URLs for POST requests
This is also known as "V4 signed policy documents".  We need a version of `Client::CreatedSignedPolicyDocument` that works for V4 signatures.

The details of the algorithm to follow.

dopiera:
This fixes #3326 and #3635.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3687)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -49,7 +49,7 @@ build:ubsan --copt=-fsanitize=undefined
 build:ubsan --copt=-fno-omit-frame-pointer
 build:ubsan --linkopt=-fsanitize=undefined
 build:ubsan --linkopt=-lubsan
-build:ubsan --action_env=UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1
+build:ubsan --action_env=UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1:suppressions=/v/bazel/ubsan_suppressions.txt
 
 # --config msan: Memory Sanitizer
 build:msan --strip=never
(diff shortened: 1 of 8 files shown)


dopiera:
> The code LGTM, but the `ubsan` failure seems worrisome, thoughts?

I still haven't resolved it. It's weird. It says that a pointer does not point to `std::__codecvt_abstract_base<char32_t, char, __mbstate_t>` but to `std::codecvt_utf8<char32_t, 1114111ul, (std::codecvt_mode)0>` instead. That doesn't seem to be a problem to me because the former is the base of the latter.

What makes it even weirder, the same code isolated in a test doesn't yield these errors. I'm tempted to call it a false-positive, but either way I need to find a good workaround :/

frankyn:
Hi @dopiera, looks like this got stuck. Are you still trying to debug?

dopiera:
> Hi @dopiera, looks like this got stuck. Are you still trying to debug?

It became extreme yak shaving:
* either standard library or clang ubsan is broken on ubuntu-18.04, so following @coryan's  advice, I tried running ubsan on something newer (namely Fedora 31)
* clang's ubsan doesn't complain about the piece of code anymore (when compiled by hand), but fails to compile through bazel, because bazel uses a wrong linker (I filed
[a bug to address it](https://github.com/bazelbuild/bazel/issues/11122) )

I don't think it's possible to easily override the linker in bazel, so I'll try suppressing this failure tomorrow.

frankyn:
Thanks for the update @dopiera! Suppressing is a reasonable path froward while the bug is addressed.

coryan:
@dopiera soon you will need to rebase this PR. I am making changing to the build configuration that require a more recent baseline than this is based on.

PR #3637, opened 2020-04-01
fix: cleanup KOKORO_ARTIFACTS_DIR

devjgm:
Fixes #3618

I suspect this will shave a few minutes off most of our docker builds.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3637)

PR #3694, opened 2020-04-06
feat: use -prune flag to find

devjgm:
When cleaning up the artifacts dir, we want to remove the top-level
`cmake-out` directory, but once we find it, we can stop descending into
it since we're going to delete the whole thing. This change is in
response to a comment that @devbww made to the original PR
https://github.com/googleapis/google-cloud-cpp/pull/3637

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3694)


devjgm:
No idea why the codecov build is reporting that it failed. I don't think it *should* have failed, and clicking the link indicates that it seemed to work (i.e., I see no indication of error on this page: https://codecov.io/gh/googleapis/google-cloud-cpp/pull/3694). Also, I have no idea how to get more info, nor how to trigger a rebuild.

PR #3695, opened 2020-04-06
refactor: no scripts for table_admin_snippets

### LINKED ISSUE #3677: Refactor run_all_table_admin_examples to table_admin_snippets.cc
The title references a shell function in `google/cloud/bigtable/run_examples_utils.sh` this function needs to be converted to a C++ function similar to `RunAll()` in `data_snippets.cc`:

https://github.com/googleapis/google-cloud-cpp/blob/277b4a5bc7a308ad8b85337c8fe0f56371f23e04/google/cloud/bigtable/examples/data_snippets.cc#L1117

This change might require converting the `.cc` program referenced above to use the `google::cloud::bigtable::Example` as the main driver, so it can be automatically executed.

coryan:
Fixes #3677

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3695)


google/cloud/bigtable/examples/table_admin_snippets.cc
```diff
@@ -70,14 +42,11 @@ void CreateTable(google::cloud::bigtable::TableAdmin admin, int argc,
               << "\n";
   }
   //! [create table] [END bigtable_create_table]
-  (std::move(admin), table_id);
+  (std::move(admin), argv.at(0));
```
devbww, line 45: Do we want `vector:at` in these cases?  That is, any run-time error generated here is really a static error in the "auto" call or the `MakeCommandEntry()` call, and a simple subscript would make that clearer.  (I don't recall seeing these in other instances.)

coryan: I learned the hard way that it was very hard to debug mistakes in `RunAll()` with the number of arguments, the crash was very obscure without them.  I think we should change the existing examples to use `std::vector::at`, but if you think they are distracting then I can just use brackets for now. Just say the word.

devbww: > I think we should change the existing examples to use `std::vector::at`

Sounds good to me.  Thanks.

Issue #3532, opened 2020-03-27
Add top-level BUILD file with user-facing targets


devjgm:
Add a top-level BUILD file with top-level targets for customers to use, similar to [-spanner/BUILD](https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/BUILD)

PR #3702, opened 2020-04-07
feat: add top-level BUILD file

devjgm:
Fixes: #3532

The goal is to get bazel users to *only* use these top-level BUILD
targets. This target is shorter and easier to type, and customers need
not rely on the directory structure that we choose to use.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3702)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -19,3 +19,17 @@ licenses(["notice"])  # Apache 2.0
 exports_files([
     "LICENSE",
 ])
+
+cc_library(
+    name = "bigtable_client",
+    deps = [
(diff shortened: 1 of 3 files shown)


devjgm:
> Should we change the `google/cloud/**/quickstart/BUILD` files to use these new packages as part of this PR?

We can't, because the quickstarts use the latest Release of google-cloud-cpp. We need to make that change after this BUILD file is included in a release. I suppose I should have made this change before last weeks release, but I forgot about this issue. 

I filed https://github.com/googleapis/google-cloud-cpp/issues/3703 to track this.

PR #3704, opened 2020-04-07
refactor: ci scripts now use log_<color> functions

devjgm:
This shortens the log statements, gives them all the same timestamp
prefix, and removes the need to remember to reset the terminal.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3704)

### DIFF
--- a/ci/colors.sh
+++ b/ci/colors.sh
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Prefer the log_* functions, such as log_yellow, instead of directly using
+# these variables. The functions don't require the caller to remember to reset
+# the terminal.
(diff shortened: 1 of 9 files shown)


devjgm:
> This is really nice, can we get (now or in a future PR) a `log_normal` that prints the timestamp but leaves the color alone?
> 
> > Reviewed 9 of 9 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3704#-:-M4KYm4BEY2EVHEXyL9P:b-hcgoua)_ status: :shipit: complete! all files reviewed, all discussions resolved

Good idea. Done. This replaces almost all of the remaining uses of `date -u`

Issue #3536, opened 2020-03-27
Change to use a single top-level CHANGELOG.md


devjgm:
Change -cpp to use a single top-level CHANGELOG.md file rather than per-library CHANGELOG.md files.

PR #3706, opened 2020-04-07
feat: merged changelogs into a top-level CHANGELOG.md

devjgm:
Fixes: #3536

Note: in this PR I remove the old per-library CHANGELOG.md files. Alternatively, if we think users will expect those files in that location, we can leave them there with a 1-liner directing them to the single top-level `CHANGELOG.md` file.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3706)


devjgm:
Done. The sections should now be (mostly) sorted alphabetically. I agree that we should sort them going forward too.

PR #3707, opened 2020-04-07
cleanup: remove repetitive code in storage examples

coryan:
The examples get their configuration from environment variables and
need to validate that these are set and have some value. That code was
getting repetitive, so copied the code from the Bigtable examples.
Eventually this should all make it to the `-common` library.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3707)

### DIFF
--- a/google/cloud/storage/examples/storage_bucket_acl_samples.cc
+++ b/google/cloud/storage/examples/storage_bucket_acl_samples.cc
@@ -231,20 +231,16 @@ void RunAll(std::vector<std::string> const& argv) {
   namespace examples = ::google::cloud::storage::examples;
   namespace gcs = ::google::cloud::storage;
 
+  examples::CheckEnvironmentVariablesAreSet({
+      "GOOGLE_CLOUD_PROJECT",
+      "GOOGLE_CLOUD_CPP_STORAGE_TEST_HMAC_SERVICE_ACCOUNT",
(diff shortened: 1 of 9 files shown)


google/cloud/storage/examples/storage_bucket_acl_samples.cc
```diff
@@ -231,20 +231,16 @@ void RunAll(std::vector<std::string> const& argv) {
   namespace examples = ::google::cloud::storage::examples;
   namespace gcs = ::google::cloud::storage;
 
+  examples::CheckEnvironmentVariablesAreSet({
+      "GOOGLE_CLOUD_PROJECT",
+      "GOOGLE_CLOUD_CPP_STORAGE_TEST_HMAC_SERVICE_ACCOUNT",
+  });
   auto const project_id =
       google::cloud::internal::GetEnv("GOOGLE_CLOUD_PROJECT").value_or("");
```
devbww, line 239: Q.  Should we still use `value_or()` after `CheckEnvironmentVariablesAreSet()`?

coryan: Good question, I think not.  I made the changes in this PR, but will send a separate PR to fix `bigtable/examples/`.

PR #3714, opened 2020-04-08
refactor: run storage CMEK examples without scripts

### LINKED ISSUE #3666: Refactor run_all_cmek_examples shell function to storage_cmek_samples.cc

coryan:
Split the object and bucket CMEK examples to their own files, which
are automatically executed by CMake/Bazel.

Fixes #3666

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3714)


google/cloud/storage/examples/storage_bucket_default_kms_key_samples.cc
```diff
@@ -0,0 +1,161 @@
...
+
+    if (!updated->has_encryption()) {
+      std::cerr << "The change to set the encryption attribute on bucket "
+                << updated->name()
+                << " was sucessful, but the encryption is not set."
+                << "This is unexpected, maybe a concurrent change?\n";
+      return;
+    }
+
+    std::cout << "Successfully set default KMS key on bucket  "
```
devjgm, line 47: Nit: looks like two spaces in this string. Not sure if that's on purpose.

google/cloud/storage/examples/storage_bucket_default_kms_key_samples.cc
```diff
@@ -0,0 +1,161 @@
...
+  auto const cmek_key =
+      google::cloud::internal::GetEnv("GOOGLE_CLOUD_CPP_STORAGE_TEST_CMEK_KEY")
+          .value();
+  auto generator = google::cloud::internal::DefaultPRNG(std::random_device{}());
+  auto const bucket_name =
+      examples::MakeRandomBucketName(generator, "cloud-cpp-test-examples-");
+  auto client = gcs::Client::CreateDefaultClient().value();
+
+  std::cout << "\nCreating bucket to run the example (" << bucket_name << ")"
+            << std::endl;
```
devjgm, line 119: These `std::endl`s look a little out of place. Consider using "\n" at the end of the preceding string, which I think is our normal style. And if flushing is needed, consider using `std::flush` to make that clear.

google/cloud/storage/examples/storage_bucket_default_kms_key_samples.cc
coryan: Thanks, fixed.

google/cloud/storage/examples/storage_bucket_default_kms_key_samples.cc
coryan: We have been using `std::endl` in all the examples. Yes, flushing is needed because when an example crashes you want to know which one was it.

If it is all the same, I am going to stick to `std::endl`, I think it is clear enough.

devjgm: If that's what we've been doing, then OK.

PR #3717, opened 2020-04-08
cleanup: cleanup the firestore documentation and deps

devjgm:
Cleans up the README.md to make it clearer that this library is not GA
and has no specific support guarantees. Also cleans up the build a bit
since it doesn't actually depend on -common.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3717)

### DIFF
--- a/google/cloud/firestore/BUILD
+++ b/google/cloud/firestore/BUILD
@@ -22,9 +22,7 @@ cc_library(
     name = "firestore_client",
     srcs = firestore_client_srcs,
     hdrs = firestore_client_hdrs,
-    deps = [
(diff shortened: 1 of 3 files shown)


devjgm:
Oh, awesome. Bazel problem fixed. It appears that in in Bazel, the spelling of running "gmock main" [is `:gtest_main`](https://github.com/google/googletest/blob/e3f0319d89f4cbf32993de595d984183b1a9fc57/BUILD.bazel#L114-L115)

That's different in CMake, where it's actually "gmock_main"

PR #3718, opened 2020-04-08
fix: incorrect name for signature field

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3718)

### DIFF
--- a/google/cloud/storage/examples/storage_policy_doc_samples.cc
+++ b/google/cloud/storage/examples/storage_policy_doc_samples.cc
@@ -110,7 +110,7 @@ void CreatePolicyDocumentFormV4(google::cloud::storage::Client client,
        << "' type='hidden' />\n"
        << "  <input name='x-goog-date' value='"
        << gcs::FormatDateForForm(*document) << "' type='hidden' />\n"
-       << "  <input name='signature' value='" << document->signature
+       << "  <input name='x-goog-signature' value='" << document->signature
        << "' type='hidden' />\n"
        << "  <input type='submit' value='Upload File' name='submit' /><br />\n"
        << "  <input type='file' name='file' /><br />\n"


google/cloud/storage/examples/storage_policy_doc_samples.cc
```diff
@@ -110,7 +110,7 @@ void CreatePolicyDocumentFormV4(google::cloud::storage::Client client,
        << "' type='hidden' />\n"
        << "  <input name='x-goog-date' value='"
        << gcs::FormatDateForForm(*document) << "' type='hidden' />\n"
-       << "  <input name='signature' value='" << document->signature
+       << "  <input name='x-goog-signature' value='" << document->signature
```
frankyn, line 113: @coryan for other  languages we are iterating over the resulting document object to get the fields to reduce the need to explicitly get the value for each individually. Is this not possible with the C++ surface?

Example: https://github.com/googleapis/nodejs-storage/blob/master/samples/generateV4SignedPolicy.js#L55-L58

I want developers to rely on the library to tell them which fields need to be added to the form.

coryan: No, the result is a C++ structure, there is no iteration over the fields of a struct.

coryan: I guess the obvious followup is: why not return something that we can iterate over?  That is not very idiomatic, C++ programmers expect things to have a well-known type. They expect "dates" and "expiration times" to be actual objects of the right type, not just a string.

And why not a `std::map<std::string, std::any>`?  Because `std::any` is a C++17 type, and we support C++11 (and also because discovering the type at run-time is not very idiomatic).

frankyn: Thanks that was going to be my follow-up. Approving.

PR #3721, opened 2020-04-09
feat: helper functions for bigtable async examples

coryan:
The Bigtable instance admin and table admin async examples will
benefit from a function to adapt an example of the form:

```C++
void F(
  bigtable::TableAdmin, CompletionQueue cq,
  std::vector<std::string> argv)
```

to something that just takes `std::vector<std::string>` as its only
argument. This PR creates two helper functions to create such adapters.

Part of the work for #3676

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3721)

### DIFF
--- a/google/cloud/bigtable/examples/bigtable_examples_common_test.cc
+++ b/google/cloud/bigtable/examples/bigtable_examples_common_test.cc
@@ -409,6 +409,67 @@ TEST(BigtableExamplesCommon, MakeTableAsyncCommandEntry) {
   EXPECT_EQ(1, call_count);
 }
 
+TEST(BigtableExamplesCommon, MakeInstanceAdminAsyncCommandEntry) {
+  // Pretend we are using the emulator to avoid loading the default
+  // credentials from $HOME, which do not exist when running with Bazel.
+  google::cloud::testing_util::ScopedEnvironment emulator(
+      "BIGTABLE_EMULATOR_HOST", "localhost:9090");
+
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/examples/bigtable_examples_common_test.cc
```diff
@@ -409,6 +409,67 @@ TEST(BigtableExamplesCommon, MakeTableAsyncCommandEntry) {
...
+                               CompletionQueue const&,
+                               std::vector<std::string> const& argv) {
+    ++call_count;
+    ASSERT_EQ(2, argv.size());
+    EXPECT_EQ("a", argv[0]);
+    EXPECT_EQ("b", argv[1]);
+  };
+  auto const actual = MakeCommandEntry("command-name", {"foo", "bar"}, command);
+  EXPECT_EQ("command-name", actual.first);
+  EXPECT_THROW(
```
devjgm, line 459: Do we need to be wrapping these tests that throw in `GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS` so that they can be run in environments with exceptions disabled?

coryan: No. The examples do not compile without exceptions (we actually throw on error to keep the code simple).  I don't see the point of complicating the code to support examples to work in environments without exceptions.

Issue #3539, opened 2020-03-27
Move release/release.sh into -cpp


devjgm:
Move release/release.sh from -common to -cpp.
https://github.com/googleapis/google-cloud-cpp-common/tree/master/release

PR #3728, opened 2020-04-09
feat: moves release.sh (with history) to -cpp

devjgm:
This is part of: https://github.com/googleapis/google-cloud-cpp/issues/3539

This copies `releases/` from https://github.com/googleapis/google-cloud-cpp-common/tree/master/release into this directory, *preserving its git history*. I preserved the git history using `git filter-repo`, which is not part of the base git release, but it is *recommended* by the base git release (`git help filter-branch` recommends this). https://github.com/newren/git-filter-repo

I have not made any edits to this code, because I want this to be a clean copy from `-common`. I'll do a followup PR if there are edits that need to be made.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3728)

### DIFF
--- a/release/README.md
+++ b/release/README.md
@@ -1,0 +1,2 @@
+This directory contains documents and tools for managing releases of the Google
+Cloud C++ client libraries.
--- a/release/release.sh
+++ b/release/release.sh
@@ -117,7 +117,6 @@ banner "Creating and pushing branch ${NEW_BRANCH}"
 run git checkout -b "${NEW_BRANCH}" "${NEW_TAG}"
 run git push --set-upstream origin "${NEW_BRANCH}"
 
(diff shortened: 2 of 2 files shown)


devjgm:
Fixed. PTAL.

For the record, I was able to fix this using the following command in `-common`:

```shell
cd common2  # A new clone of google-cloud-cpp-common
git filter-repo --path release --force --message-callback 'return re.sub(b"\(#(\d+)\)", b"(googleapis/google-cloud-cpp-common#\\\\1)", message)'
```

Then to merge that change into `-cpp`, I did the following

```shell
cd google-cloud-cpp
git remote add common2 ../common2
git fetch common2
git merge common2/master --allow-unrelated-histories  # Here I used common2/master, but any branch would work
```

PR #3743, opened 2020-04-10
feat: taught release.sh how to insert release notes

devjgm:
Fixes: #3539

Cleaned up the release.sh script to do the following:

1. Quote command-line args correctly when printing them in the `run`
function.
2. Automatically extract the release notes for the new version from
`CHANGELOG.md` and insert them into the new release that we create on
the GitHub UI (using the hub command).
3. Only clone the previous 90 days worth of commits, which should be
more than enough to get the recently released tags. This saves a lot of
time when cloning `-cpp`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3743)

### DIFF
--- a/release/release.sh
+++ b/release/release.sh
@@ -99,10 +99,41 @@ function banner() {
 }
 
 function run() {
-  echo "# $*" | paste -d' ' -s -
+  printf "# "
+  for arg in "$@"; do
+    printf "%s " "'$arg'"
+  done
+  printf "\n"
   if [[ "${FORCE_FLAG}" == "yes" ]]; then
     "$@"
   fi
+}
+
+# Outputs the release notes for the given tag. Looks for the release notes in
+# the CHANGELOG.md file starting at a heading that looks like "## <tag>". For
+# example `get_release notes v0.6.0` would look for a line like "## v0.6.0". If
+# no release notes could be found, a default string is output telling the
+# operator to insert the release notes manually.
+function get_release_notes() {
+  local tag="$1"
+  local found=false
+  local notes=()
+  while IFS= read -r line; do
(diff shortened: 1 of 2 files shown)


devjgm:
You've written almost all of our exiting shell scripts, so if this is OK w/ you, then that's good enough fo rme ;-) Thanks.

release/release.sh
```diff
@@ -164,25 +198,18 @@ banner "Creating and pushing branch ${NEW_BRANCH}"
...
-readonly release_fmt="
-   date: %cI
-    url: %U
-tarball: %uT
-zipball: %uZ
-  state: %S
-  title: %t
-   body: %b
-"
+readonly release_fmt="%n date: %cI %n  url: %U %nstate: %S %ntitle: %t %n"
```
devbww, line 212: Do you like the line-ending whitespace?

release/release.sh
```diff
@@ -99,12 +99,43 @@ function banner() {
 }
 
 function run() {
-  echo "# $*" | paste -d' ' -s -
+  printf "# "
+  for arg in "$@"; do
+    printf "%s " "'$arg'"
+  done
+  printf "\n"
```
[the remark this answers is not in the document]
devjgm: Cool. I didn't know about `%q`. Thanks! Fixed in https://github.com/googleapis/google-cloud-cpp/pull/3760

release/release.sh
devjgm: Yes, I did like them because I think they made the format string easier to read. But removed them in https://github.com/googleapis/google-cloud-cpp/pull/3760

PR #3779, opened 2020-04-14
feat: update to new version of googleapis protos

devjgm:
With this version, we can now use the googleapis-provided BUILD files
for the Bigtable Admin V2 protos, so we don't need to define our own. I
also move the unnecessary proto deps out of the `googleapis.BUILD` file
to make the targets in there look more consistent -- they only exist so
that we can explicitly sort the `grpc++` target *after* the grpc proto
rules that depend on it (this is working around a grpc issue).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3779)

### DIFF
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -64,10 +64,10 @@ def google_cloud_cpp_deps():
         http_archive(
             name = "com_google_googleapis",
             urls = [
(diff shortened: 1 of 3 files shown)


devjgm:
> FYI, I like it better when the CMake and Bazel versions of these protos are the same (see the cpp-cmakefiles repo), but that can wait until a future PR.

They were already different, but I'll bring them into sync in a separate PR.

PR #3780, opened 2020-04-15
test: move ubsan tests to cmake

dopiera:
Issue #3687 has shown that ubsan needs an update. The failure generated
by it couldn't be suppressed. bazel couldn't be used with newer versions
of clang  ecause of https://github.com/bazelbuild/bazel/issues/11122, so
the only option was to move ubsan to cmake and workaround the bazel bug.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3780)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -41,16 +41,6 @@ build:tsan --copt=-fno-omit-frame-pointer
 build:tsan --linkopt=-fsanitize=thread
 build:tsan --action_env=TSAN_OPTIONS=halt_on_error=1:second_deadlock_stack=1
 
-# --config ubsan: Undefined Behavior Sanitizer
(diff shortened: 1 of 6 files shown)


devjgm:
See if you can fix the ubsan problem by adding this line to `.bazelrc`

```
build:ubsan --linkopt=-fsanitize-link-c++-runtime
```

dopiera:
> See if you can fix the ubsan problem by adding this line to `.bazelrc`
> 
> ```
> build:ubsan --linkopt=-fsanitize-link-c++-runtime
> ```
@devjgm That does solve the problem indeed! Thank you. It looks like we can now migrate to Fedora 31 with UBSan without moving to cmake.

PR #3783, opened 2020-04-15
refactor: run storage_client_mock_samples without scripts

### LINKED ISSUE #3673: Refactor run_mocking_client_examples shell function to storage_client_mock_samples.cc

coryan:
Fixes #3673 

----
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3783)

### DIFF
--- a/google/cloud/storage/examples/CMakeLists.txt
+++ b/google/cloud/storage/examples/CMakeLists.txt
@@ -62,11 +62,6 @@ create_bazel_config(storage_examples_common)
 if (BUILD_TESTING)
     find_package(google_cloud_cpp_testing CONFIG REQUIRED)
 
-    add_executable(storage_client_mock_samples storage_client_mock_samples.cc)
-    target_link_libraries(
-        storage_client_mock_samples storage_client storage_common_options
-        GTest::gmock_main GTest::gmock GTest::gtest)
-
     foreach (fname ${storage_examples_unit_tests})
         string(REPLACE "/" "_" target ${fname})
         string(REPLACE ".cc" "" target ${target})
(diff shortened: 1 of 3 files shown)


coryan:
___
*[google/cloud/storage/examples/storage_client_mock_samples.cc, line 28 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3783#-M4yLpX76V94iG4Azu_-:-M4yMX8M0dJfrmyUVBY2:bpclb5g) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/88cd864aada41dc123e8483415a1ea019f741d0c/google/cloud/storage/examples/storage_client_mock_samples.cc#L28)):*

nit: we've been avoiding the `using` since this is only used in one place.
</blockquote></details>

Ack, but this file uses it 5 times.  The functions have a different form because they are supposed to show you how to initialize `gcs::Client`

mr-salty:
___
*[google/cloud/storage/examples/storage_client_mock_samples.cc, line 28 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3783#-M4yLpX76V94iG4Azu_-:-M4yQRQ39rXdkQ7kZNwO:b-wtbfxr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/88cd864aada41dc123e8483415a1ea019f741d0c/google/cloud/storage/examples/storage_client_mock_samples.cc#L28)):*

Ack, but this file uses it 5 times.  The functions have a different form because they are supposed to show you how to initialize `gcs::Client`
</blockquote></details>

ha, i shouldn't do reviews when I first get up I guess ;)

in that case, I think we should fully qualify it (`::google...`)

PR #3795, opened 2020-04-16
cleanup: add varargs support to `CreateCommandEntry`

mr-salty:
It saves a little bit of boilerplate and avoids a special case where the
commands are defined.

I also made `storage_bucket_iam_samples` use it (more uses are in other
to-be-converted samples).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3795)

### DIFF
--- a/google/cloud/storage/examples/storage_examples_common.cc
+++ b/google/cloud/storage/examples/storage_examples_common.cc
@@ -132,9 +132,13 @@ std::string MakeRandomObjectName(google::cloud::internal::DefaultPRNG& gen,
 Commands::value_type CreateCommandEntry(
     std::string const& name, std::vector<std::string> const& arg_names,
     ClientCommand const& command) {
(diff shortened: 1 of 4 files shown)


google/cloud/storage/examples/storage_examples_common.cc
```diff
@@ -132,9 +132,13 @@ std::string MakeRandomObjectName(google::cloud::internal::DefaultPRNG& gen,
 Commands::value_type CreateCommandEntry(
     std::string const& name, std::vector<std::string> const& arg_names,
     ClientCommand const& command) {
+  bool allow_varargs =
+      !arg_names.empty() &&
+      arg_names[arg_names.size() - 1].find("...") != std::string::npos;
```
devbww, line 137: `arg_names.back()` perhaps?

Issue #3794, opened 2020-04-16
speed up `Verify Formatting` step


mr-salty:
When I run `./ci/kokoro/docker/build.sh clang-tidy`) on my workstation, the `Verify formatting` step takes ~45 seconds even with only a few files modified. Investigate whether we can speed this up (does it/could it format changed files only?)

coryan:
There are also opportunities to parallelize this, e.g., running clang-format in different threads for `google/cloud/{storage,bigtable,spanner,.}`

Likewise, the cmake-format, buildifier and python formatting can run in parallel.

devjgm:
Here's timing info from my laptop:

```
## Checking include guards START:  Thu Apr 16 13:13:36 UTC 2020
## Checking include guards   END:  Thu Apr 16 13:13:36 UTC 2020
Delta: <1s

## Checking cmake format START:  Thu Apr 16 13:13:36 UTC 2020
## Checking cmake format   END:  Thu Apr 16 13:13:47 UTC 2020
Delta: 11s

## Checking clang-format START:  Thu Apr 16 13:13:47 UTC 2020
## Checking clang-format   END:  Thu Apr 16 13:14:05 UTC 2020
Delta: 18s

## Checking sed xform START:  Thu Apr 16 13:14:05 UTC 2020
## Checking sed xform   END:  Thu Apr 16 13:14:06 UTC 2020
Delta: 1s
 
## Checking buildifier START:  Thu Apr 16 13:14:06 UTC 2020
## Checking buildifier   END:  Thu Apr 16 13:14:06 UTC 2020
Delta: <1s

## Checking python formatting START:  Thu Apr 16 13:14:06 UTC 2020
## Checking python formatting   END:  Thu Apr 16 13:14:07 UTC 2020
Delta: 1s

## Checking shell check START:  Thu Apr 16 13:14:07 UTC 2020
## Checking shell check   END:  Thu Apr 16 13:14:08 UTC 2020
Delta: 1s

## Checking fixing whitespace START:  Thu Apr 16 13:14:08 UTC 2020
## Checking fixing whitespace   END:  Thu Apr 16 13:14:10 UTC 2020
Delta: 2s
```

This is quite unscientific, and it's just one snapshot. But it tells us roughly which ones steps are the big ones and which are small. Note: that I've seen the clang-format step take upwards of 40s.

I'd say that focusing on the clang-format and cmake-format steps will be the biggest bang.

devjgm:
[...]

PR #3802, opened 2020-04-16
fix: speed up clang-format in check-style

devjgm:
Changes this loop back to using `clang-format -i`. This used to cause a
problem when we were using clang-format-7, in that it would update every
file's timestamp even if no edits were made. However, since #3458
(2020-03-10), we're now using fedora with clang-format-9, and this
problem no longer exists. So we can go back to the faster `clang-format
-i`.

This PR reduces the "clang-format" step in the check-style.sh script
from about 30 seconds to about 4.

I also removed the clang-format-7 packages from the Dockerfiles that
don't use it, because it was just confusing to see a version being
installed that we don't use.

Part of https://github.com/googleapis/google-cloud-cpp/issues/3794

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3802)

### DIFF
--- a/ci/check-style.sh
+++ b/ci/check-style.sh
@@ -68,11 +68,8 @@ find . \( "${ignore[@]}" \) -prune -o \
 # Apply clang-format(1) to fix whitespace and other formatting rules.
 # The version of clang-format is important, different versions have slightly
 # different formatting output (sigh).
-find google/cloud \( -name '*.cc' -o -name '*.h' \) -print0 |
-  while IFS= read -r -d $'\0' file; do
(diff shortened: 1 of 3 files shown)


coryan:
This is probably unrelated, but do we understand why the cmake-dll build failed? @devjgm if rerunning fixes the problem please file a `[Flake]` bug.

devjgm:
> This is probably unrelated, but do we understand why the cmake-dll build failed? @devjgm if rerunning fixes the problem please file a `[Flake]` bug.

https://github.com/googleapis/google-cloud-cpp/issues/3803

PR #3806, opened 2020-04-16
ci: another iteration on quickstart builds for macOS

### LINKED ISSUE #3798: Find a safer way to handle quickstart arguments
We are currently storing the arguments for the different quickstart programs in an associate array of strings, with each argument separated by a space, for example:

```bash
declare -A args=(
  ["lib1"]="A B C"
  ["lib2"]="C D E"
)
```

That is `lib1`'s quickstart will receive "A", "B", and "C" as arguments. If we wanted to pass arguments that contained spaces this would not work. It would be nice to have a more robust way to represent this, like an associate array of lists, but those do not seem to work with bash.

For now this works.  This bug is left as a reference in case it becomes a problem or somebody is looking for something to fix.

coryan:
Made the quickstart builds for Linux and macOS more similar, fixed a
number of shellcheck warnings, incidentally fixing #3798. Also made the
output more legible (I think). Also made sure all quickstarts are built,
even if some of them fail.

Fixes #3798 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3806)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-quickstart-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-quickstart-bazel.sh
@@ -62,7 +62,8 @@ fi
 build_quickstart() {
   local -r library="$1"
 
-  cd "${PROJECT_ROOT}/google/cloud/${library}/quickstart"
+  pushd "${PROJECT_ROOT}/google/cloud/${library}/quickstart" >/dev/null
+  trap "popd >/dev/null" RETURN
   log_normal "capture bazel version"
   ${BAZEL_BIN} version
   log_normal "fetch dependencies for ${library}'s quickstart"
--- a/ci/kokoro/docker/build-in-docker-quickstart-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-quickstart-bazel.sh
@@ -59,43 +59,52 @@ if [[ -r "/c/kokoro-run-key.json" ]]; then
   )
 fi
 
-build_service() {
-  local -r service="$1"
+build_quickstart() {
+  local -r library="$1"
 
-  echo "================================================================"
-  log_yellow "building ${service}"
-  ( cd "google/cloud/${service}/quickstart";
-    log_normal "capture bazel version"
-    ${BAZEL_BIN} version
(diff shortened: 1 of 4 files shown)


ci/kokoro/docker/build-in-docker-quickstart-bazel.sh
```diff
@@ -59,43 +59,52 @@ if [[ -r "/c/kokoro-run-key.json" ]]; then
...
 
-  echo "================================================================"
-  log_yellow "building ${service}"
-  ( cd "google/cloud/${service}/quickstart";
-    log_normal "capture bazel version"
-    ${BAZEL_BIN} version
-    log_normal "fetch dependencies for ${service}'s quickstart"
-    "${PROJECT_ROOT}/ci/retry-command.sh" \
-        "${BAZEL_BIN}" fetch -- ...
+  cd "${PROJECT_ROOT}/google/cloud/${library}/quickstart"
```
devbww, line 65: Perhaps a function should never have a `cd` side effect?  Given that we're in `bash`, `pushd` and `popd` may be appropriate.

ci/kokoro/macos/build-quickstart-bazel.sh
```diff
@@ -74,10 +74,14 @@ readonly run_quickstart
 echo "================================================================"
 cd "${PROJECT_ROOT}"
 
-for library in $(quickstart_libraries); do
-  echo "================================================================"
+build_quickstart() {
+  local -r library="$1"
+
   cd "${PROJECT_ROOT}/google/cloud/${library}/quickstart"
```
devbww, line 80: Ditto.

ci/kokoro/macos/build-quickstart-cmake.sh
```diff
@@ -68,25 +71,53 @@ cmake_flags=(
...
 )
 
-for library in $(quickstart_libraries); do
-  echo "================================================================"
+build_quickstart() {
+  local -r library="$1"
+
+  echo
   log_yellow "Configure CMake for ${library}'s quickstart."
   cd "${PROJECT_ROOT}"
```
devbww, line 79: Ditto.

PR #3813, opened 2020-04-16
refactor: run storage rename object example without scripts

### LINKED ISSUE #3660: Refactor run_rename_object_example shell function to storage_rename_samples.cc

coryan:
Fixes #3660

----
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3813)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -1512,7 +1512,7 @@ class Client {
    * @snippet storage_object_samples.cc rotate encryption key
    *
    * @par Example: using rewrite object to rename an object
-   * @snippet storage_object_samples.cc rename object
+   * @snippet storage_object_rewrite_samples.cc rename object
    */
   template <typename... Options>
   StatusOr<ObjectMetadata> RewriteObjectBlocking(
--- a/google/cloud/storage/examples/run_examples_utils.sh
+++ b/google/cloud/storage/examples/run_examples_utils.sh
@@ -237,27 +237,6 @@ run_rename_object_example() {
 }
 
 ################################################
-# Run the examples showing how to rewrite objects.
-# Globals:
-#   COLOR_*: colorize output messages, defined in colors.sh
-#   EXIT_STATUS: control the final exit status for the program.
-# Arguments:
-#   source_bucket_name: an existing bucket where the source object will be
-#     created.
-#   target_bucket_name: an existing bucket where the target object will be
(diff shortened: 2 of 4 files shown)


coryan:
___
*[google/cloud/storage/examples/storage_object_rewrite_samples.cc, line 190 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3813#-M53jK593Vulpdw2Gvdt:-M53kD-L9v23Mu8il7aR:b-3axzz5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/237261481b7aba206f56b13c5110f3f7bdf51c45/google/cloud/storage/examples/storage_object_rewrite_samples.cc#L190)):*

> ```
>   (void)client.InsertObject(bucket_name, old_object_name, text).value();
> 
>   std::cout << "\nRunning the RenameObject() example" << std::endl;
>   RenameObject(client, {bucket_name, old_object_name, new_object_name});
> 
>   std::cout << "\nCleanup" << std::endl;
>   (void)client.DeleteObject(bucket_name, new_object_name);
> ```

FWIW, when I had cases like this in `storage_object_examples` I tended to only output one message (`Running the foo() example`) followed by a series of operations (look at `ComposeObject` as one example).  Do you think I should change those to follow this pattern?
</blockquote></details>

If you have lots of operations in a row then the failures may be hard to troubleshoot, with "lots" anything bigger than 5 maybe?  Otherwise I would leave it alone.

mr-salty:
___
*[google/cloud/storage/examples/storage_object_rewrite_samples.cc, line 190 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3813#-M53jK593Vulpdw2Gvdt:-M53knrmE0UlPDv3izlI:b4vmzxn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/237261481b7aba206f56b13c5110f3f7bdf51c45/google/cloud/storage/examples/storage_object_rewrite_samples.cc#L190)):*

If you have lots of operations in a row then the failures may be hard to troubleshoot, with "lots" anything bigger than 5 maybe?  Otherwise I would leave it alone.
</blockquote></details>

Yeah, "lots" is less than 5, but I guess the key point being `Insert` and `Delete` are really just prep work for the actual example. If it's hard to tell what failed when something fails, though, then it's probably better to output more.

PR #3819, opened 2020-04-16
fix: cleanup after moving the bigquery code

devjgm:
These are changes that *should have* been part of
https://github.com/googleapis/google-cloud-cpp/pull/3808 but jgm forgot
to run the `clang-tidy` build locally.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3819)


devjgm:
> Here and in the other `.bzl` files, I think it would be nicer to change the source `CMakeLists.txt` file to override the copyright year and leave this file was it was. Thoughts?

SGTM. Done.

PR #3842, opened 2020-04-17
refactor: remove unneeded targets from overlay BUILD

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3842)

### DIFF
--- a/bazel/googleapis.BUILD
+++ b/bazel/googleapis.BUILD
@@ -38,21 +38,3 @@ cc_library(
     ],
 )
 
-cc_library(
-    name = "bigquery_protos",
-    # Do not sort: grpc++ must come last
-    deps = [
-        "@com_google_googleapis//google/cloud/bigquery/storage/v1beta1:storage_cc_grpc",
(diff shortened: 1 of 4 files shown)


devjgm:
> I think the release notes should note this as a breaking change, and therefore this should be marked as `refactor!` as it removes public targets for the library. All the more reason to have more limited public targets as you want them, but nevertheless.
> 
> We should consider leaving targets with the right dependencies and a `deprecation` attribute saying something like "this will be removed on or about 2020-07-01" or some date like that.
> 
> Thoughts?

I don't think of this as a breaking change, because I don't think any users would or should think that they can depend on us providing a build target for something in some other repo, like `@com_google_googleapis`. If you think that expectation may not have been clear to customers and that they might have thought these build targets were part of our public API, then I can leave them and comment about their deprecation.

We should also clearly document our support guarantees so users aren't surprised when we break them for doing something that they shouldn't do. Filed https://github.com/googleapis/google-cloud-cpp/issues/3846 is track this.

PR #3848, opened 2020-04-18
fix: moved .dockerignore to ci/

devjgm:
During the docker build phase, the `.dockerignore` file is searched for
in the "context", which in the case of our builds is the `ci/`
directory. So, I believe, previously the `.dockerignore` file was not
being used. Moving it into `ci/` seems to cause it to be used now in my
tests.

The result of this is that our docker images should be rebuilt less
often.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3848)


devjgm:
> This is a great find, thanks.
> 
> We still have `ci/benchmarks/Dockerfile` which builds all the code. We are current not using this file, so arguably we should remove it (and the whole directory for that matter). If we want to keep it (I vote we do not) then we should keep the old top-level `.dockerignore` too.
> 
> Thoughts?

OK, so as you said, it looks like `ci/benchmarks/` is not used, so I removed that whole directory. That SGTM, but let me know if I misunderstood what you were suggesting.

Issue #3540, opened 2020-03-27
Enable integration tests on Windows and Mac


devjgm:
Make sure test coverage is similar (e.g. -cpp does not run integration tests for Windows or macOS, while -spanner does).

devjgm:
The test coverage for all repos is currently similar, between 93% (`-cpp) - 97% (`-pubsub`). So merging these repos we'll likely end up somewhere around 95%.

![coverage](https://user-images.githubusercontent.com/9447643/78923405-2852f680-7a66-11ea-985e-41d3824e67ae.png)

Is that all we wanted to confirm with this issue, or is there more? I'm wondering if this issue is now done?

coryan:
No, see the description.

* We do not run any of the integration tests on Windows, they "should" work, but  how do we know?
* We do not run any integration tests on macOS either.  How do we know the code works?
* We do not run the integration tests with any of the sanitizers, there may be bugs we are missing because the unit tests do not provide nearly the same coverage.

In `-spanner` we do run the integration tests and samples for Windows and macOS and with the sanitizers. I do not think we want to lose that coverage as we move the code to `-cpp`.

devjgm:
I see. OK. So I think i'll re-title this issue "Enable integration tests on Windows and Mac"... done. If I've misunderstood (again) that's a bad title, let me know and I'll fix it.

PR #3866, opened 2020-04-20
ci: enable integration tests for macOS+Bazel

coryan:
Part of the work for #3540

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3866)

### DIFF
--- a/ci/kokoro/macos/build-bazel.sh
+++ b/ci/kokoro/macos/build-bazel.sh
@@ -81,3 +81,62 @@ echo "================================================================"
 log_yellow "build all targets."
 "${BAZEL_BIN}" build \
   "${bazel_args[@]}" -- //google/cloud/...:all
+
+readonly CONFIG_DIR="${KOKORO_GFILE_DIR:-/private/var/tmp}"
+readonly INTEGRATION_TESTS_CONFIG="${PROJECT_ROOT}/ci/etc/integration-tests-config.sh"
+readonly TEST_KEY_FILE_JSON="${CONFIG_DIR}/kokoro-run-key.json"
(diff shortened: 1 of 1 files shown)


ci/kokoro/macos/build-bazel.sh
```diff
@@ -81,3 +81,62 @@ echo "================================================================"
...
+    "--test_env=GOOGLE_CLOUD_CPP_STORAGE_TEST_SIGNING_CONFORMANCE_FILENAME=${PROJECT_ROOT}/google/cloud/storage/tests/v4_signatures.json"
+    "--test_env=GOOGLE_CLOUD_CPP_STORAGE_TEST_KEY_FILE_JSON=${TEST_KEY_FILE_JSON}"
+    "--test_env=GOOGLE_CLOUD_CPP_STORAGE_TEST_KEY_FILE_P12=${TEST_KEY_FILE_P12}"
+  )
+
+  "${BAZEL_BIN}" test \
+    "${bazel_args[@]}" \
+    "--test_tag_filters=bigtable-integration-tests,storage-integration-tests" \
+    -- //google/cloud/...:all \
+    -//google/cloud/bigtable/examples:bigtable_grpc_credentials \
```
devjgm, line 138: (no change necessary) It would be nice if we could minimize the product-specific info in the CI scripts, and in this case maybe we could use a generic tag that would correctly filter out these tests.

coryan: I agree, but I have no good ideas yet, I will create a GitHub issue with more questions than answers.

coryan: Done: #3868 ideas on how to organize this are welcome. I am just making it up as I go.

Issue #3872, opened 2020-04-21
ensure `cloud-cpp-testing-bucket` gets cleaned up regularly


mr-salty:
Currently, our testing bucket `cloud-cpp-testing-bucket` has 43156 objects with date stamps from 2020-03-21 to 2020-04-21.  Having that many objects in the bucket drastically increases the runtime of some of our integration tests (see #3854) - when running against an empty bucket the runtimes of the two longest tests drop from 115 to 5 seconds and 72 to 2 seconds.

We should clean these objects up much more aggressively, or find another way to preserve performance even if the bucket becomes heavily populated.

mr-salty:
is there any reason not to change this to 1 day from 30 days?

```
gsutil lifecycle  get gs://cloud-cpp-testing-bucket
{"rule": [{"action": {"type": "Delete"}, "condition": {"age": 30}}]}
```

coryan:
> is there any reason not to change this to 1 day from 30 days?

I cannot think of any, please make the change.

mr-salty:
thanks - done.  I'll call this closed for now, probably even a day is longer than we need but GCS only supports days and I don't think it's worth writing some custom job to do that.

n.b.: "Updates to your lifecycle configuration may take up to 24 hours to go into effect."  Maybe I'll clean some of them up manually.

```
$ gsutil lifecycle get gs://cloud-cpp-testing-bucket
{"rule": [{"action": {"type": "Delete"}, "condition": {"age": 1}}]}
```

mr-salty:
> Maybe I'll clean some of them up manually.

this is easy enough to do with xargs so I'm doing this now.

mr-salty:
I cleaned up everything older than 24h but this still seems like a lot, are they "leaking" from failed tests?

$ `gsutil ls gs://cloud-cpp-testing-bucket  | wc -l`
6980

coryan:
> I cleaned up everything older than 24h but this still seems like a lot, are they "leaking" from failed tests?

As long as they don't leak from successful tests and we have the lifecycle rule I am mostly happy.  Making the tests more complicated to delete things on a failure seems like the wrong place to put effort.

devbww:
> As long as they don't leak from successful tests ...

I don't grok the API, so forgive a naive question, but who deletes those `.src-<N>` objects?

mr-salty:
oh yeah, I meant to check my own bucket and forgot... but indeed it was left with a bunch of files:

[...]

Issue #3854, opened 2020-04-19
Split storage/tests/object_integration_test


coryan:
This test takes 230 seconds to run on my workstation when running against production, and a similar time for those builds that run against production.

There are several expensive test cases that should be moved to their own `.cc` files (and therefore can run in paralel), or their parameters should be reduced to make the testing more efficient (save coverage for smaller cost), including:

ObjectIntegrationTest.BasicCRUD (27674 ms)
ObjectIntegrationTest.ListObjectsVersions (49172 ms)
ObjectIntegrationTest.PlentyClientsSimultaneously (20931 ms)
ObjectIntegrationTest.PlentyClientsSerially (20161 ms)
ObjectIntegrationTest.ComposeMany (32518 ms)
ObjectIntegrationTest.ParallelUpload (36683 ms)
ObjectIntegrationTest.ResumeParallelUploadFile (12907 ms)

```
[==========] Running 33 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 33 tests from ObjectIntegrationTest
[ RUN      ] ObjectIntegrationTest.BasicCRUD
[       OK ] ObjectIntegrationTest.BasicCRUD (27674 ms)
[ RUN      ] ObjectIntegrationTest.FullPatch
[       OK ] ObjectIntegrationTest.FullPatch (932 ms)
[ RUN      ] ObjectIntegrationTest.ListObjectsVersions
[       OK ] ObjectIntegrationTest.ListObjectsVersions (49172 ms)
[ RUN      ] ObjectIntegrationTest.ListObjectsDelimiter
[       OK ] ObjectIntegrationTest.ListObjectsDelimiter (2595 ms)
[ RUN      ] ObjectIntegrationTest.BasicReadWrite
[       OK ] ObjectIntegrationTest.BasicReadWrite (852 ms)
[ RUN      ] ObjectIntegrationTest.PlentyClientsSimultaneously
[       OK ] ObjectIntegrationTest.PlentyClientsSimultaneously (20931 ms)
[ RUN      ] ObjectIntegrationTest.PlentyClientsSerially
[       OK ] ObjectIntegrationTest.PlentyClientsSerially (20161 ms)
[ RUN      ] ObjectIntegrationTest.EncryptedReadWrite
[       OK ] ObjectIntegrationTest.EncryptedReadWrite (906 ms)
[ RUN      ] ObjectIntegrationTest.ReadNotFound
[...]

PR #3873, opened 2020-04-21
refactor: parallelize `object_integration_test`

mr-salty:
Split the longer-running tests of `object_integration_test.cc` into
their own files, so we can run them in parallel.

The longest component test executes for only half the time of what
Carlos reported (115s vs 230s), although my timings were generally
slower than his to begin with. Also note #3872 has a drastic effect
on the two longest running tests, if we address that the runtime
should drop to ~47 seconds which is a 5x speedup.

Part of #3854

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3873)


mr-salty:
*Note there are no actual code changes here, it's a straight refactoring.*

The common bits got pulled to `testing/` and the rest are just tests from `storage_object_test.cc` split out.

I tried to fix up the #includes but we could really use IWYU (I think there were both missing and extra headers in the original).

google/cloud/storage/testing/object_integration_test.cc
```diff
@@ -0,0 +1,86 @@
...
+  if (!res) {
+    return res;
+  }
+  if (*res < 3) {
+    return Status(StatusCode::kInternal,
+                  "Expected at least three entries in /proc/self/fd: ., .., "
+                  "and the directory itself, found " +
+                      std::to_string(*res));
+  }
+  return *res - 3;
```
devbww, line 65: I'm not sure why we bother with the `3` business given that we only deal in differentials, but it should probably just be `2` and forget about "the directory itself" bit.  It is, after all, an open file.

google/cloud/storage/testing/object_integration_test.cc
```diff
@@ -0,0 +1,86 @@
...
+  dirent** namelist;
+  int res = scandir(path.c_str(), &namelist, nullptr, alphasort);
+  if (res < 0) {
+    return Status(StatusCode::kInternal, "Failed to open directory \"" + path +
+                                             "\": " + strerror(errno));
+  }
+  for (int i = 0; i < res; ++i) {
+    free(namelist[i]);
+  }
+  free(namelist);
```
devbww, line 43: It looks like a `readdir()` loop would be simpler here, and avoid extra allocations and unnecessary sorting.

mr-salty:
since this is all existing code, my preference would be to check this in and address the comments in follow-up PRs - SG?

mr-salty:
___
*[google/cloud/storage/tests/object_compose_many_integration_test.cc, line 48 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3873#-M5SALLZBplDhjWeoa5t:-M5SaD9y42CIdMrg9P-K:bxq3w51) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/344ce74d941ff2f91a356d711c90ed97cf3f3584/google/cloud/storage/tests/object_compose_many_integration_test.cc#L48)):*

FYI: I think this could be smaller. I think something like 33 (or 65) is Okay. Basically we want something that:

- Is larger than 32 because <= 32 results in a single `Compose()` call.
- It is not a multiple of 32 so we verify that we can handle remainders, when the size does not fit in exactly N `Compose()` calls.
- We may want to test with more than one compose, suggesting >= 64
- We want to test the case where of 32*N + 1

Hmmm, maybe we want to test with 33 and 34 to cover the cases where the remainder both requires and does not require a compose call.  OTOH, the unit tests cover some of this I believe.


</blockquote></details>

Addressed in #3875 (although I'm not clear if you actually landed on 33 so LMK)

mr-salty:
___
*[google/cloud/storage/testing/object_integration_test.cc, line 43 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3873#-M5PqvZ0bHdI2S98Zfxg-r1-43:-M5T95rF2TclemnFHmuA:bept859) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/344ce74d941ff2f91a356d711c90ed97cf3f3584/google/cloud/storage/testing/object_integration_test.cc#L43)):*

It looks like a `readdir()` loop would be simpler here, and avoid extra allocations and unnecessary sorting.
</blockquote></details>

Fixed in #3876

mr-salty:
___
*[google/cloud/storage/testing/object_integration_test.cc, line 65 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3873#-M5PqvZ0bHdI2S98Zfxg-r1-65:-M5T97pC6k1xG8k4FL2W:baiz7qt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/344ce74d941ff2f91a356d711c90ed97cf3f3584/google/cloud/storage/testing/object_integration_test.cc#L65)):*

I'm not sure why we bother with the `3` business given that we only deal in differentials, but it should probably just be `2` and forget about "the directory itself" bit.  It is, after all, an open file.
</blockquote></details>

I think 3 is correct, see my comment in #3876

mr-salty:
___
*[google/cloud/storage/tests/object_plenty_clients_serially_integration_test.cc, line 60 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3873#-M5PqvZ0bHdI2S98Zfxm-r1-60:-M5T9H8_01NcfHmUHJik:bpl7pw7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/344ce74d941ff2f91a356d711c90ed97cf3f3584/google/cloud/storage/tests/object_plenty_clients_serially_integration_test.cc#L60)):*

Should this be inside the conditional?
</blockquote></details>

yes, I also changed this around a bit in #3876 - LMK what you think

PR #3876, opened 2020-04-21
test: rework some logic for counting open files

mr-salty:
addresses devbww's comments on #3873

* use `opendir()` instead of `scandir()` to read the directory
* I do think eliding 3 entries makes sense since opening `/proc/self/fd`
  is a side effect of taking the measurement. I updated the comment.
* contained the #ifdef to one place, which I think makes it easier to
  read. The test still fails if we can't get the count on linux, and it
  will also fail as a reminder to update it if we find a way to get
  the count on non-linux.

part of #3854

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3876)


mr-salty:
@devbww ptal

PR #3879, opened 2020-04-21
test: ensure sure tests clean up objects they create

### LINKED ISSUE #3872: ensure `cloud-cpp-testing-bucket` gets cleaned up regularly

mr-salty:
In the "Compose Many" test I use `TearDown` in case there are any
`ASSERT`s that terminate the test early.  We don't have that issue
in the "List" test but it might be worth considering a more generic
cleanup mechanism for all tests.

I ran all of the integration tests against my own bucket and verified
no files were left over.

Fixes #3872

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3879)

### DIFF
--- a/google/cloud/storage/tests/object_compose_many_integration_test.cc
+++ b/google/cloud/storage/tests/object_compose_many_integration_test.cc
@@ -33,42 +33,56 @@ namespace storage {
 inline namespace STORAGE_CLIENT_NS {
 namespace {
 
-using ObjectComposeManyIntegrationTest =
-    ::google::cloud::storage::testing::ObjectIntegrationTest;
+class ObjectComposeManyIntegrationTest
+    : public google::cloud::storage::testing::ObjectIntegrationTest {
+ protected:
+  void SetUp() override {
+    google::cloud::storage::testing::ObjectIntegrationTest::SetUp();
+    client_ = MakeIntegrationTestClient();
+    ASSERT_STATUS_OK(client_);
+  }
+
+  void TearDown() override {
+    for (auto const& source_object : source_objects_) {
+      EXPECT_STATUS_OK(
+          client_->DeleteObject(bucket_name_, source_object.object_name));
+    }
+    google::cloud::storage::testing::ObjectIntegrationTest::TearDown();
+  }
+
+  StatusOr<Client> client_;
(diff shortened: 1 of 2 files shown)


mr-salty:
___
*[google/cloud/storage/tests/object_list_objects_versions_integration_test.cc, line 107 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3879#-M5TiyCAEOi-XNduoTSK:-M5Tpgc5E_-2kx-9c8ur:b-dj7h52) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/62e9863b8a5205681f734e72dd1aedbc04b79440/google/cloud/storage/tests/object_list_objects_versions_integration_test.cc#L107)):*

This will delete the "current" version of the object, but not the "archived" versions (and I think the latest one becomes "current", but not sure).

I think you want to delete all 3?  You should do something like:

```C++
std::vector<ObjectMetadata> objects_to_delete; // populate with created objects...


....
...
for (auto const& o : objects_to_delete) {
  EXPECT_STATUS_OK(client->DeleteObject(o.bucket(), o.name(), gcs::Generation(o.generation())));
}
```


</blockquote></details>

ah, I think you may be getting at a much larger issue here - I was not aware of how the versioning works.

when I run the integration tests with these changes, there are no live objects when we're done (`ls` shows nothing).  But, you are correct that there are archived versions, which `ls -a` shows. These do not become automatically live when the live object is deleted though.

this must also afflict the `cloud-cpp-testing` bucket - `ls -a` there shows 230k objects (previously there were 40k live objects, but I deleted all but 6-7k of those).  So, our buckets are way more full than I thought.

mr-salty:
closing this PR for now to work on a version that ensures both live and prior versions of all objects (not limited to these two tests) are cleaned up.

PR #3880, opened 2020-04-22
refactor: use git-ls-files rather than parse .gitignore

devjgm:
This simplifies the our check-style.sh script by not parsing the `.gitignore` file. 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3880)

### DIFF
--- a/ci/check-style.sh
+++ b/ci/check-style.sh
@@ -87,7 +87,7 @@ fi
 #     - Replace grpc::<BLAH> with grpc::StatusCode::<BLAH>, the aliases in the
 #       `grpc::` namespace do not exist inside google.
 git ls-files -z | grep -zE '\.(cc|h)$' |
-  while IFS= read -r file; do
+  while IFS= read -r -d $'\0' file; do
     # We used to run run `sed -i` to apply these changes, but that touches the
     # files even if there are no changes applied, forcing a rebuild each time.
     # So we first apply the change to a temporary file, and replace the original
@@ -104,7 +104,7 @@ git ls-files -z | grep -zE '\.(cc|h)$' |
 # we do not expand TABs (they currently only appear in Makefiles and Makefile
 # snippets).
 git ls-files -z | grep -zv '\.gz$' |
-  while IFS= read -r file; do
+  while IFS= read -r -d $'\0' file; do
     sed -e 's/[[:blank:]][[:blank:]]*$//' \
       "${file}" >"${file}.tmp"
     replace_original_if_changed "${file}" "${file}.tmp"
--- a/ci/check-style.sh
+++ b/ci/check-style.sh
@@ -50,27 +50,30 @@ replace_original_if_changed() {
 
 # Apply cmake_format to all the CMake list files.
 #     https://github.com/cheshirekow/cmake_format
-git ls-files | grep -P '(CMakeLists\.txt|\.cmake)$' | xargs cmake-format -i
+git ls-files -z | grep -zE '((^|/)CMakeLists\.txt|\.cmake)$' |
+  xargs -0 cmake-format -i
 
 # Apply clang-format(1) to fix whitespace and other formatting rules.
 # The version of clang-format is important, different versions have slightly
 # different formatting output (sigh).
-git ls-files | grep -P '\.(cc|h)$' | xargs clang-format -i
+git ls-files -z | grep -zE '\.(cc|h)$' | xargs -0 clang-format -i
 
 # Apply buildifier to fix the BUILD and .bzl formatting rules.
 #    https://github.com/bazelbuild/buildtools/tree/master/buildifier
-git ls-files | grep -P '\.(BUILD|bzl)$' | xargs buildifier -mode=fix
+git ls-files -z | grep -zE '\.(BUILD|bzl)$' | xargs -0 buildifier -mode=fix
+git ls-files -z | grep -zE '((^|/)BUILD|WORKSPACE)$' |
+  xargs -0 buildifier -mode=fix
 
 # Apply psf/black to format Python files.
 #    https://github.com/bazelbuild/buildtools/tree/master/buildifier
-git ls-files | grep '\.py$' | xargs python3 -m black
+git ls-files -z | grep -z '\.py$' | xargs -0 python3 -m black
 
 # Apply shfmt to format all shell scripts
-git ls-files | grep '\.sh$' | xargs shfmt -w -i 2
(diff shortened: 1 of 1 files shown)


ci/check-style.sh
```diff
@@ -61,24 +50,41 @@ replace_original_if_changed() {
 
 # Apply cmake_format to all the CMake list files.
 #     https://github.com/cheshirekow/cmake_format
-find . \( "${ignore[@]}" \) -prune -o \
-  \( -name 'CMakeLists.txt' -o -name '*.cmake' \) \
-  -print0 |
-  xargs -0 cmake-format -i
+git ls-files | grep -P '(CMakeLists\.txt|\.cmake)$' | xargs cmake-format -i
```
devbww, line 53: Precise translation of `-name 'CMakeLists.txt'` would be to match `(^|/)CMakeLists\.txt$` (i.e., not just a suffix).  Similarly for other full names below.

ci/check-style.sh
```diff
@@ -61,24 +50,41 @@ replace_original_if_changed() {
...
 # Apply clang-format(1) to fix whitespace and other formatting rules.
 # The version of clang-format is important, different versions have slightly
 # different formatting output (sigh).
-find google/cloud \( -name '*.cc' -o -name '*.h' \) -print0 |
-  xargs -0 clang-format -i
+git ls-files | grep -P '\.(cc|h)$' | xargs clang-format -i
+
+# Apply buildifier to fix the BUILD and .bzl formatting rules.
+#    https://github.com/bazelbuild/buildtools/tree/master/buildifier
+git ls-files | grep -P '\.(BUILD|bzl)$' | xargs buildifier -mode=fix
```
devbww, line 62: Previously we did `BUILD` and `WORKSPACE` files too.

ci/check-style.sh
devjgm: Done.

ci/check-style.sh
devjgm: Oops. Accidentally lost those. Added back. Thanks.

devjgm:
> > I did drop the feature of passing nul-terminated strings
> 
> `xargs -d '\n'` will get you back the space-in-name files (but not the newline-in-name files).
> 
> But why does it get "a little weirder with `grep -z`"?

Because the man page documents that `-z` doesn't work right with `-P`, which I was using. But I changed to EREs, so it should be fine now. Done.

ci/check-style.sh
```diff
@@ -61,57 +50,29 @@ replace_original_if_changed() {
...
+git ls-files -z | grep -zE '\.(cc|h)$' | xargs -0 clang-format -i
 
 # Apply buildifier to fix the BUILD and .bzl formatting rules.
 #    https://github.com/bazelbuild/buildtools/tree/master/buildifier
-find . \( "${ignore[@]}" \) -prune -o \
-  \( -name WORKSPACE -o -name BUILD \
-  -o -name '*.BUILD' -o -name '*.bzl' \) \
-  -print0 |
+git ls-files -z | grep -zE '\.(BUILD|bzl)$' | xargs -0 buildifier -mode=fix
+git ls-files -z | grep -zE '((^|/)BUILD|WORKSPACE)$' |
```
devbww, line 64: Why two passes when we do all the others with `grep` alternations?

ci/check-style.sh
```diff
@@ -120,18 +81,30 @@ if ! find . \( "${ignore[@]}" \) -prune -o \
...
-# Apply shfmt to format all shell scripts
-find . \( "${ignore[@]}" \) -prune -o \
-  -iname '*.sh' -print0 | xargs -0 shfmt -w -i 2
+# Apply several transformations that cannot be enforced by clang-format:
+#     - Replace any #include for grpc++/* with grpcpp/*. The paths with grpc++
+#       are obsoleted by the gRPC team, so we should not use them in our code.
+#     - Replace grpc::<BLAH> with grpc::StatusCode::<BLAH>, the aliases in the
+#       `grpc::` namespace do not exist inside google.
+git ls-files -z | grep -zE '\.(cc|h)$' |
+  while IFS= read -r file; do
```
devbww, line 90: We need to restore the `-d $'\0'` here.

ci/check-style.sh
```diff
@@ -120,18 +81,30 @@ if ! find . \( "${ignore[@]}" \) -prune -o \
...
 # Apply transformations to fix whitespace formatting in files not handled by
 # clang-format(1) above.  For now we simply remove trailing blanks.  Note that
 # we do not expand TABs (they currently only appear in Makefiles and Makefile
 # snippets).
-find . \( "${ignore[@]}" \) -prune -o \
-  -type f ! -name '*.gz' \
-  -print0 |
-  while IFS= read -r -d $'\0' file; do
+git ls-files -z | grep -zv '\.gz$' |
+  while IFS= read -r file; do
```
devbww, line 107: We need to restore the `-d $'\0'` here.

ci/check-style.sh
```diff
@@ -120,18 +81,30 @@ if ! find . \( "${ignore[@]}" \) -prune -o \
...
 fi
 
-# Apply shfmt to format all shell scripts
-find . \( "${ignore[@]}" \) -prune -o \
-  -iname '*.sh' -print0 | xargs -0 shfmt -w -i 2
+# Apply several transformations that cannot be enforced by clang-format:
+#     - Replace any #include for grpc++/* with grpcpp/*. The paths with grpc++
+#       are obsoleted by the gRPC team, so we should not use them in our code.
+#     - Replace grpc::<BLAH> with grpc::StatusCode::<BLAH>, the aliases in the
+#       `grpc::` namespace do not exist inside google.
```
devbww, line 88: Aside: We should probably justify this transformation more generally.

devjgm: ack

ci/check-style.sh
devjgm: Yep done.

ci/check-style.sh
devjgm: Yep. done.

ci/check-style.sh
devjgm: It seems more readable this way rather than merging them into one bigger regex.

devbww, line 64: It looks like the `(^|/)` should be outside the parentheses.

devjgm: Done

Issue #3549, opened 2020-03-27
Copy code from google-cloud-cpp-pubsub


devjgm:
# Moving code into `-cpp` (a single PR)

## Move the code into `google/cloud/$project`

Think of this as simply copying the files. But ideally, we'll copy the files over *and* preserve their git history. See https://github.com/googleapis/google-cloud-cpp/pull/3728#issuecomment-611804509 for details about how this was done for the `release/` directory. Once the files are in `-cpp/google/cloud/$project`, we'll need to make further changes, otherwise the `-cpp` build will break.

GOAL: Try make preparations in the original repo so that when the code moves into `-cpp`, you can make _zero_ changes to the newly added files in `google/cloud/$project`. 

## Update top-level `CMakeLists.txt`

Add an option/flag to make building the new dir optional.

## Update files in `ci/...`

grep in `ci/...` for places that mention `bigtable` or `storage` to see where you might need to add an additional configuration for the new `$project` being added. In particular:

* `ci/kokoro/docker/build-in-docker-cmake.sh` has a list of expected directories that get installed, which will likely need to be updated.
* `ci/kokoro/docker/publish-refdocs.sh` has the list of directories that produce documentation.
* `ci/etc/integration-tests-config.*` list the environment variables necessary to run the integration tests and examples.
* `ci/etc/quickstart-config.*` lists the directories with a `quickstart` example and now to invoke that example.

## Update top-level README.md

This file should contain a link/reference to the newly added code. This file is probably autogenerated, so be sure to edit the right file(s).

## Quick/local tests

1. Run `build.sh asan`, which will use bazel to build.
1. Run `build.sh clang-tidy`, which will use cmake
1. Run `build.sh integration`

In both cases, try creating a build break in the newly imported code to make sure the build fails. Also try creating a runtime test failure, to make sure the test fails.

Also, create a DRAFT PR and see what the presubmit CI builds say.

NOTE: All these steps should go in a single PR.

devjgm:
Note: here's the script I used to move the code from `-bigquery` while preserving history. It should be easy to modify for other cases:

```sh
#!/bin/bash
#
# Creates a `NEW_REPO` that contains all code in `google/cloud` from the old
# repo and its git history. Uses the "filter-repo" command from
# https://github.com/newren/git-filter-repo. This could also work with the
# built-in "filter-branch", but the docs for that command say "don't use it;
# use filter-repo instead".

set -eu

[...]

PR #3887, opened 2020-04-22
feat: move google-cloud-cpp-pubsub (with history) into this repo

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/3549

This PR moves the code in http://github.com/googleapis/google-cloud-cpp-pubsub into this repo at  `google/cloud/pubsub`. All git history from the previous repo has been preserved according to [this script](https://github.com/googleapis/google-cloud-cpp/issues/3549#issuecomment-614963681). 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3887)


coryan:
FYI, I think we need to add the `integration-tests-no-emulator` label to all the integration tests for them to be picked up by the CI scripts. Right now they are not running because they are neither `storage-integration-tests`, nor `bigtable-integration-tests` to run under the respective emulator.

We probably need to change `ci/etc/integration-tests-config-*` to add any Pub/Sub specific configurations too, maybe.

devjgm:
> FYI, I think we need to add the `integration-tests-no-emulator` label to all the integration tests for them to be picked up by the CI scripts. Right now they are not running because they are neither `storage-integration-tests`, nor `bigtable-integration-tests` to run under the respective emulator.
> 
> We probably need to change `ci/etc/integration-tests-config-*` to add any Pub/Sub specific configurations too, maybe.

Done.

PR #3889, opened 2020-04-22
ci: disable cache upload/download for non-CI builds

coryan:
These scripts are used to run the builds locally, on a developer
workstations. With these changes the cache is not downloaded or
downloaded unless KOKORO_JOB_TYPE is configured, which hopefully
developers only will set when they mean to simulate the CI environment
and they understand the consequences.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3889)

### DIFF
--- a/ci/kokoro/docker/download-cache.sh
+++ b/ci/kokoro/docker/download-cache.sh
@@ -46,7 +46,7 @@ if [[ ! -f "${KEYFILE}" ]]; then
   exit 0
 fi
 
-if [[ -z "${KOKORO_JOB_TYPE:-}" ]] ||
+if [[ "${RUNNING_CI:-}" != "yes" ]] ||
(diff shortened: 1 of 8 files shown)


ci/kokoro/docker/download-cache.sh
```diff
@@ -46,7 +46,8 @@ if [[ ! -f "${KEYFILE}" ]]; then
   exit 0
 fi
 
-if [[ "${KOKORO_JOB_TYPE:-}" != "PRESUBMIT_GERRIT_ON_BORG" ]] &&
+if [[ -z "${KOKORO_JOB_TYPE:-}" ]] ||
```
devbww, line 49: Use a single `[[ ... ]]` command?

coryan: Done.

PR #3894, opened 2020-04-23
ci: use ccache in ci/kokoro/install builds

### LINKED ISSUE #3521: Enable Cache for ci/kokoro/install

coryan:
On Pull-Request builds, and if available, download the
ccache tarball and store them in `ci/kokoro/install/ccache-contents`.
The Dockerfile then extracts this tarball in `/h/.ccache` and configures
ccache to use that directory.

At the end of a successful build update the tarball. If the build is a CI
build then upload the tarball to the GCS folder.

Fixes #3521 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3894)

### DIFF
--- a/ci/kokoro/cache-functions.sh
+++ b/ci/kokoro/cache-functions.sh
@@ -63,7 +63,7 @@ cache_upload_enabled() {
     return 1
   fi
 
-  if [[ "${RUNNING_CI:-}" != "TODO(coryan)" || "${KOKORO_JOB_TYPE:-}" != "CONTINUOUS_INTEGRATION" ]]; then
+  if [[ "${RUNNING_CI:-}" != "yes" || "${KOKORO_JOB_TYPE:-}" != "CONTINUOUS_INTEGRATION" ]]; then
     echo "================================================================"
     log_normal "Cache not updated as this is not a CI build or it is a PR build."
     return 1
--- a/ci/kokoro/cache-functions.sh
+++ b/ci/kokoro/cache-functions.sh
@@ -35,7 +35,7 @@ cache_download_enabled() {
 
 cache_gcloud_cleanup() {
   revoke_service_account_keyfile "${CACHE_KEYFILE}" || true
-  delete_gcloud_config
+  delete_gcloud_config || true
 }
 
 cache_download_tarball() {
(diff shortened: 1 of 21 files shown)


coryan:
Note that the builds are super slow in this PR because (a) we need to change the Dockerfiles, and that invalidates the docker cache, and (b) the ccache is not built yet. Which is to say we may need another PR to see any benefits, and probably some troubleshooting :disappointed:

ci/kokoro/cache-functions.sh
```diff
@@ -0,0 +1,87 @@
...
+  revoke_service_account_keyfile "${CACHE_KEYFILE}" || true
+  delete_gcloud_config
+}
+
+cache_download_tarball() {
+  local -r GCS_FOLDER="$1"
+  local -r DESTINATION="$2"
+  local -r FILENAME="$3"
+
+  trap cache_gcloud_cleanup EXIT
```
devjgm, line 46: I believe traps aren't local to a function, so this trap will still be set after this function exits. Therefore, I wonder if it'd be clearer to list this trap immediately after declaring the `cache_gcloud_cleanup()` function? WDYT? Is that a bad idea?

ci/kokoro/install/build.sh
```diff
@@ -49,46 +49,59 @@ else
...
   has_cache="true"
 fi
 
+readonly CACHE_BUCKET="${GOOGLE_CLOUD_CPP_KOKORO_RESULTS:-cloud-cpp-kokoro-results}"
+readonly CACHE_FOLDER="${CACHE_BUCKET}/build-cache/google-cloud-cpp/master/install"
+readonly CACHE_NAME="${DISTRO}.tar.gz"
+
+if cache_download_enabled; then
+  cache_download_tarball \
+    "${CACHE_FOLDER}" "ci/kokoro/install/ccache-contents" "${CACHE_NAME}" || true
```
devjgm, line 100: It seems a little strange that we're downloading ephemeral, working files and stuff to a directory inside our repo. Is there a reason why this needs to be the case? Can this not be somewhere like /tmp or something? It's fine to keep, but I'm just curious the reason.

coryan:
PTAL

coryan:
I made too many changes, so I re-requested a review.

ci/kokoro/install/build.sh
devjgm: FTR, @coryan and I discussed this offline. It would be nice to avoid having a cache directory in our repo, but neither of us can think of a reasonable way to avoid it at this point.

coryan: Created #3915

PR #3905, opened 2020-04-23
fix: crash for bigtable's scan_throughput_benchmark

coryan:
On Windows+CMake+Debug builds the C++ library has a few more assertions,
and it crashed when we passed invalid ranges to the random number
generators.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3905)


coryan:
@scotthart still needs an actual LGTM

PR #3910, opened 2020-04-23
refactor: move some setup into the fixture class

mr-salty:
`StorageIntegrationTest::SetUp()` now handles client creation and
reading environment variables, so we don't have to repeat that in
every test.

I merged `ObjectIntegrationTest` into `StorageIntegrationTest` because
there wasn't enough difference to justify two classes.

My goal here is to have `StorageIntegrationTest::TearDown()` ensure all
objects created by the test get cleaned up (see #3872) but I will do that
in another PR since this one is large enough already.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3910)


mr-salty:
___
*[google/cloud/storage/testing/storage_integration_test.cc, line 72 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3910#-M5dJyei-LLfMnD71R41:-M5wK_CE5AfGTwG3NI33:bv22in6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5e5fa679fc111ddff74be656a863b57fe744f5d5/google/cloud/storage/testing/storage_integration_test.cc#L72)):*

Many of our tests needed only one environment variable, and before this change I could run them from the command line (or from a debugger, or an IDE or with valgrind) using something like:

```sh
GOOGLE_CLOUD_CPP_STORAGE_TEST_BUCKET_NAME=foo ./the-one-test
```

Now I have to set 4 environment variables, some of which are unrelated, and some which have long names and values (look at the value of `GOOGLE_CLOUD_CPP_STORAGE_TEST_SERVICE_ACCOUNT`).  That seems like less than ideal to me.

What if you refactored the function to validate a list of environment variables to `StorageIntegrationTest` and let each test define its own list (maybe as part of their `setup` maybe with some default list).  Or what if we post-pone the validation until the first call to `bucket_name()` or something like that?

</blockquote></details>

I really wanted to postpone the validation until access but you can't use `ASSERT` in a function unless it returns `void`.  I could throw an exception or just `CHECK`, neither of which seemed great to me (or change those methods to use an out parameter - yuck)

Initially I only had the environment variables that were truly ubiquitous in the base class, but I moved as many as possible there to avoid needing to derive and implement `SetUp` (and remember to call the base class `SetUp`, which is a pain).  Deriving and implementing a constructor isn't as bad, but you can't use `ASSERT` in a constructor either per the [FAQ](https://github.com/google/googletest/blob/master/googletest/docs/faq.md#should-i-use-the-constructordestructor-of-the-test-fixture-or-setupteardown-ctorvssetup)

Maybe there's another option, let me give it a bit of thought.

devjgm:
@mr-salty FYI, you can close this PR without losing the work on your branch if you'd like to think more about which direction to go.

mr-salty:
will reopen once I rework this

PR #3912, opened 2020-04-24
perf: parallelize slow formatting steps

devjgm:
With the more C++ and CMake files being added, the check-style step is
getting slower. The cmake-format step is the slowest, and was taking
about 45 seconds on my laptop. clang-format was taking about 23 seconds.

This change parallelizes the executions of these steps with `xargs -P`.
The parallelism and number of arguments values were determined
experimentally on my machine.

With this change, the cmake-format step is about 11 seconds, and the
clang-format step is about 4 seconds.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3912)

### DIFF
--- a/ci/check-style.sh
+++ b/ci/check-style.sh
@@ -51,12 +51,13 @@ replace_original_if_changed() {
 # Apply cmake_format to all the CMake list files.
 #     https://github.com/cheshirekow/cmake_format
 git ls-files -z | grep -zE '((^|/)CMakeLists\.txt|\.cmake)$' |
-  xargs -0 cmake-format -i
+  xargs -P "${NCPU}" -n 1 -0 cmake-format -i
 
 # Apply clang-format(1) to fix whitespace and other formatting rules.
 # The version of clang-format is important, different versions have slightly
 # different formatting output (sigh).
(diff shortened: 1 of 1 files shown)


ci/check-style.sh
```diff
@@ -51,12 +51,13 @@ replace_original_if_changed() {
 # Apply cmake_format to all the CMake list files.
 #     https://github.com/cheshirekow/cmake_format
 git ls-files -z | grep -zE '((^|/)CMakeLists\.txt|\.cmake)$' |
-  xargs -0 cmake-format -i
+  xargs -P "${NCPU}" -n 1 -0 cmake-format -i
```
devbww, line 54: Did `-n 1` really win?  There are 44 files that currently match that pattern, so, with `NCPU` ~= 8, I've got to imagine that `-n 6` should be better, no?

For what it's worth, I time `-n 1` at about 4.2s and `-n 6` at about 1.4s.

devjgm: I believe it did in my experiments yesterday. But if you find a better combinations, SGTM.

devbww: I was unable to produce such a big change inside a docker run.  `-n 3` seemed to be the best, but the differentials were fractional seconds, so I'm happy to stick with `-n 1`.

PR #3918, opened 2020-04-24
cleanup: import google/cloud from g-c-cpp-common

### LINKED ISSUE #3582: Copy code from google-cloud-cpp-common
UPDATED: based on @coryan 's comment below.

... to `google/cloud/*.{h,cc}` so that it lives at the same path as it does in its current repo. Otherwise, callers includes would break.

We should investigate how to preserve history as well when doing this copy.

coryan:
Fixes #3582 

----
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3918)


coryan:
The manual changes are in:

https://github.com/googleapis/google-cloud-cpp/pull/3918/files/2d0bd51711b36fcbe1ecc4158cfccea433059116..0db1df4d6113389ffd9d1d36b57ffdbb6a992b8b

PR #3921, opened 2020-04-24
fix: actually use the ccache tarball

coryan:
I made so many mistakes in these changes. In hindsight I should
have broken this down in a change to create the ccache tarball
(without using it), and then a change to use it. As it was my local
testing failed because:

- I expected the docker cache to not work as the devtools changed
  to include ccache, but then it did not work for other reasons too.
- I expected the local tests to fail (expect when I manually did the
  steps) because the cache tarballs had not been populated.

So all the testing was ad-hoc and did not catch the problems.

This time it will work, for realsies, really :fearful:

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3921)


coryan:
This was a comedy of errors, but finally, the `kokoro/install` builds in this PR took about 5m, from around 15m.

coryan:
Ping

PR #3948, opened 2020-04-26
cleanup: enable clang-tidy in pubsub headers

coryan:
clang-tidy does not scan headers by default, you have to enable a
regular expression for it. As we import projects where the headers do
pass the clang-tidy checks, we should enable the checks for that
particular project.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3948)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -49,7 +49,7 @@ Checks: >
 WarningsAsErrors: "*"
 
 # TODO(#3920) - Enable clang-tidy checks in our headers.
-HeaderFilterRegex: "google/cloud/pubbsub/.*"
+HeaderFilterRegex: "google/cloud/pubsub/.*"
 
 CheckOptions:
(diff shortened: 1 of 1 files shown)


.clang-tidy
```diff
@@ -48,8 +48,8 @@ Checks: >
 # Turn all the warnings from the checks above into errors.
 WarningsAsErrors: "*"
 
-# TODO(#205) - Enable clang-tidy checks in our headers.
-#    HeaderFilterRegex: "google/cloud/.*"
+# TODO(#3920) - Enable clang-tidy checks in our headers.
+HeaderFilterRegex: "google/cloud/pubbsub/.*"
```
devbww, line 52: s/pubbsub/pubsub/

coryan: Thanks, fixed.

Issue #3958, opened 2020-04-27
Enable clang-tidy for all targets


devjgm:
Currently I think clang-tidy is not running for all of our targets. I only see the following CMake files that are calling the `google_cloud_cpp_add_clang_tidy` CMake function:

```shell
cmake/GoogleCloudCppCommon.cmake
148:function (google_cloud_cpp_add_clang_tidy target)

google/cloud/storage/CMakeLists.txt
273:google_cloud_cpp_add_clang_tidy(storage_client)

google/cloud/bigtable/CMakeLists.txt
197:google_cloud_cpp_add_clang_tidy(bigtable_client)

google/cloud/firestore/CMakeLists.txt
66:google_cloud_cpp_add_clang_tidy(firestore_client)

google/cloud/CMakeLists.txt
200:google_cloud_cpp_add_clang_tidy(google_cloud_cpp_common)
353:    google_cloud_cpp_add_clang_tidy(google_cloud_cpp_grpc_utils)
```

Specifically there is nothing for "common", "pubsub", or "bigquery". This is probably a step we missed when migrating code over.

The *extra credit* solution would do something generic that automatically applies to *all* of our CMake targets today and in the future, so we don't have to remember to do this.

devjgm:
Here is where our build script enables clang tidy:

https://github.com/googleapis/google-cloud-cpp/blob/141d2c8d375c0b9c580a6bd78de6c3e8e3d63ae4/ci/kokoro/docker/build-in-docker-cmake.sh#L75-L77

I'm wondering if we just change to this line:

```shell
cmake_extra_flags+=("-DCMAKE_CXX_CLANG_TIDY=clang-tidy")
```
would that work for everything that's built by this cmake invocation? It may also let us avoid the need to explicitly enable clang-tidy for individual targets.

devjgm:
A quick experiment doing this shows a lot of new clang-tidy issues that were hiding out.

coryan:
I spent some time today looking into this.  The tests and examples have never had clang-tidy enabled, it might be nice to do that, but it would take significant effort.

I will send a PR (need to do some testing) to make it more obvious that the libraries have this enabled.

The fix using `CMAKE_CXX_CLANG_TIDY` seems like a good idea. I believe the path to implement should be:

- Manually add `google_cloud_cpp_add_clang_tidy(${target})` to the targets in one directory, and fix any warnings / errors generated by clang-tidy.
[...]

PR #3960, opened 2020-04-28
fix: run clang-tidy on the code for library targes

coryan:
That is, the code that gets compiled into `.a` files that our customers would
link, and including test support libraries. The code for tests and examples
will need additional work (and lots of it).

Fixed the existing (but unreported) clang-tidy warnings.

Part of the work for #3958

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3960)

### DIFF
--- a/google/cloud/bigtable/testing/table_integration_test.cc
+++ b/google/cloud/bigtable/testing/table_integration_test.cc
@@ -20,8 +20,6 @@
 #include <algorithm>
 #include <cctype>
 
-namespace btadmin = ::google::bigtable::admin::v2;
-
 namespace google {
 namespace cloud {
 namespace bigtable {
@@ -69,12 +67,16 @@ void TableTestEnvironment::SetUp() {
   std::string const family2 = "family2";
   std::string const family3 = "family3";
   std::string const family4 = "family4";
-  bigtable::TableConfig table_config =
-      bigtable::TableConfig({{family1, bigtable::GcRule::MaxNumVersions(10)},
-                             {family2, bigtable::GcRule::MaxNumVersions(10)},
-                             {family3, bigtable::GcRule::MaxNumVersions(10)},
-                             {family4, bigtable::GcRule::MaxNumVersions(10)}},
-                            {});
+  auto constexpr kTestMaxVersions = 10;
+  auto const test_gc_rule = bigtable::GcRule::MaxNumVersions(kTestMaxVersions);
+  bigtable::TableConfig table_config = bigtable::TableConfig(
+      {
+          {family1, test_gc_rule},
+          {family2, test_gc_rule},
+          {family3, test_gc_rule},
+          {family4, test_gc_rule},
+      },
+      {});
 
   table_id_ = RandomTableId();
   ASSERT_STATUS_OK(table_admin.CreateTable(table_id_, table_config));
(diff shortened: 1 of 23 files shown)


google/cloud/storage/testing/storage_integration_test.cc
```diff
@@ -50,13 +51,16 @@ StorageIntegrationTest::MakeIntegrationTestClient(
...
+    initial_delay = kShortDelayForTestbench;
   }
 
+  auto constexpr kMaximumBackoffDelay = std::chrono::minutes(5);
+  auto constexpr kBackoffScalingFactor = 2.0;
   return ExponentialBackoffPolicy(initial_delay,
-                                  /*maximum_delay=*/std::chrono::minutes(5),
-                                  /*scaling=*/2.0)
+                                  /*maximum_delay=*/kMaximumBackoffDelay,
+                                  /*scaling=*/kBackoffScalingFactor)
```
devbww, line 63: The `/*argname=*/` comments now seem less useful.

google/cloud/bigtable/testing/table_integration_test.cc
```diff
@@ -69,12 +67,16 @@ void TableTestEnvironment::SetUp() {
   std::string const family2 = "family2";
   std::string const family3 = "family3";
   std::string const family4 = "family4";
-  bigtable::TableConfig table_config =
-      bigtable::TableConfig({{family1, bigtable::GcRule::MaxNumVersions(10)},
-                             {family2, bigtable::GcRule::MaxNumVersions(10)},
-                             {family3, bigtable::GcRule::MaxNumVersions(10)},
-                             {family4, bigtable::GcRule::MaxNumVersions(10)}},
-                            {});
+  auto constexpr kTestMaxVersions = 10;
```
devjgm, line 70: This is fine to leave, as-is. But if this change was made because of the readability-magic-numbers check, we should disable it. I think that check goes way too far and more often than not makes code less clear. Of course, we can still disable it later even with this code, so no change necessary.

google/cloud/bigtable/testing/table_integration_test.cc
```diff
@@ -99,11 +101,11 @@ std::string TableTestEnvironment::CreateRandomId(std::string const& prefix,
 std::string TableTestEnvironment::RandomTableId() {
   // This value was discovered by trial and error, it is not documented in the
   // proto files.
-  constexpr int kMaxTableIdLength = 50;
-  static char const prefix[] = "table-";
-  static_assert(kMaxTableIdLength > sizeof(prefix), "prefix is too long");
-  constexpr int sample_count = kMaxTableIdLength - sizeof(prefix) + 1;
-  return CreateRandomId(prefix, sample_count);
+  auto constexpr kMaxTableIdLength = 50;
+  static char const kPrefix[] = "table-";  // NOLINT(modernize-avoid-c-arrays)
```
devjgm, line 105: (No change needed) I hate this clang-tidy check (modernize-avoid-c-arrays). I'd be happy to disable it.

Alternatively, a good way to avoid it is to do something like: `static const auto kPrefix = "foo"`, which, IIRC, fools it.

google/cloud/storage/testing/storage_integration_test.cc
coryan: Fixed.

google/cloud/bigtable/testing/table_integration_test.cc
coryan: There was some repetition too. In any case, I think the warning is pretty annoying in tests, but useful in the code itself.  Others can remove it later.

google/cloud/bigtable/testing/table_integration_test.cc
coryan: I think I tried that, and it makes `kPrefix` a `char const*`, which does not work if you use `sizeof()` or any static test on the length of the string (as we do here).

coryan:
> The PR title/description suggests that this fixes the warnings for all the library code and tests. Does that mean we can go all the way to enabling `CMAKE_CXX_CLANG_TIDY`, or is there still something that's not tidy clean?

Most of the tests and examples would require work, for a taste look at:

https://github.com/coryan/google-cloud-cpp/compare/enable-clang-tidy-for-libraries...coryan:enable-clang-tidy-everywhere?expand=1
> 


> This was a lot of work. Thanks for doing this!

No problem. I have a much larger branch which partially fixes the tests, but I gave up because I think the changes for the examples are actually harmful. For example, changing `std::string` to `std::string const&` makes the examples less readable, and I don't care about the performance implications in that case, nor do I want the `// NOLINT` comments.

PR #3970, opened 2020-04-29
chore: disable modernize-avoid-c-arrays

devjgm:
Justification is included in the changed file.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3970)


devjgm:
> you should also delete the corresponding NOLINTs: https://github.com/googleapis/google-cloud-cpp/search?q=modernize-avoid-c-arrays

Good point. Done.

PR #3979, opened 2020-04-30
cleanup: simplify quickstart CMakeLists.txt files

### LINKED ISSUE #3729: Consider requiring CMake>=3.15 to simplify quickstarts
I am about to add some CMake nastiness to handle static vs. dynamic MSVC runtimes in the quickstart CMake files.

If we were to require CMake >= 3.15 for the quickstart guides (which maybe it is reasonable), we can simplify all of this using:

https://cmake.org/cmake/help/v3.15/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#variable:CMAKE_MSVC_RUNTIME_LIBRARY

coryan:
Require CMake >= 3.15 for the quickstart examples, with that version we
can write slightly cleaner code to pick the MSVC runtime.

Fixes #3729

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3979)


coryan:
> This looks good. You hinted that there was a downside to this change: what was/is it?

We moved complexity to the READMEs, of course. If the users miss that they will get a bad experience (but the default is to use DLLs, so maybe won't affect that many folks).

PR #3984, opened 2020-05-01
cleanup: enable clang-tidy for `storage/examples`

mr-salty:
part of #3958 (also see my comment there about whether we actually want
to do this).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/3984)

### DIFF
--- a/google/cloud/storage/examples/storage_object_samples.cc
+++ b/google/cloud/storage/examples/storage_object_samples.cc
@@ -31,7 +31,7 @@ void ListObjects(google::cloud::storage::Client client,
                  std::vector<std::string> const& argv) {
   //! [list objects] [START storage_list_files]
   namespace gcs = google::cloud::storage;
-  [](gcs::Client client, std::string bucket_name) {
+  [](gcs::Client client, std::string const& bucket_name) {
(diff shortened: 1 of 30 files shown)


google/cloud/storage/examples/storage_object_samples.cc
```diff
@@ -135,12 +138,14 @@ void InsertObjectStrictIdempotency(google::cloud::storage::Client,
   (argv.at(0), argv.at(1), argv.at(2));
 }
 
+// NOLINTNEXTLINE(performance-unnecessary-value-param)
```
devjgm, line 141: The NOLINT is fine here, but I'm curious if you tried just declaring the first arg as a `const&`?

devjgm:
> It does work, although it seems a little odd since these are all supposed to conform to this:

I think that should be fine. The signature of the function template within a `std::function` does not need to exactly match that of the target functions assigned to it. The requirement is that each target function needs to be "Callable" with the std::function's argument types. So the following is totally fine and intentionally works:

```cc
int Blah(int const&);
std::function<int(int)> f = Blah;
```

PR #4011, opened 2020-05-05
feat: outputs timing for check-style blocks

devjgm:
The "Verifying formatting" step is sometimes slow-ish, and when we run
clang-tidy from the command line, it would be nice to see what it's
doing. This PR adds timing output to each section of the check-style
script so the user can see what it's doing, and we'll know if a section
all of a sudden gets slow and can be optimized. The output looks like:

```
2020-05-05T18:41:32Z: Verify formatting
Running check-include-guards: ... 0.045 seconds
Running cmake-format:         ... 3.607 seconds
Running clang-format:         ... 1.981 seconds
Running buildifier:           ... 0.041 seconds
Running black:                ... 0.148 seconds
Running shfmt:                ... 0.024 seconds
Running shellcheck:           ... 1.237 seconds
Running include fixes:        ... 1.876 seconds
Running whitespace fixes:     ... 2.632 seconds
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4011)

### DIFF
--- a/ci/check-style.sh
+++ b/ci/check-style.sh
@@ -26,11 +26,18 @@ fi
 readonly BINDIR="$(dirname "$0")"
 source "${BINDIR}/colors.sh"
 
-problems=""
-if ! find google/cloud -name '*.h' -print0 |
-  xargs -0 awk -f "${BINDIR}/check-include-guards.gawk"; then
-  problems="${problems} include-guards"
-fi
+# This controls the output format from bash's `time` command, which we use
+# below to time blocks of the script. A newline is automatically included.
+readonly TIMEFORMAT="... %R seconds"
+
+printf "%-30s" "Running check-include-guards:"
+time {
+  problems=""
+  if ! find google/cloud -name '*.h' -print0 |
+    xargs -0 awk -f "${BINDIR}/check-include-guards.gawk"; then
+    problems="${problems} include-guards"
+  fi
+}
 
 replace_original_if_changed() {
   if [[ $# != 2 ]]; then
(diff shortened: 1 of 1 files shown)


devjgm:
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4011)_ status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)
> 
> _[ci/check-style.sh, line 35 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4011#-M6aFe73D3QFjn8hj1sE:-M6aFe73D3QFjn8hj1sF:bdjvqcy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8982814a8fe65461f2fa02a1a9f94b34b33e3bcd/ci/check-style.sh#L35)):_
> 
> > ```shell
> > printf "%-30s" "Running check-include-guards:"
> > time {
> >   problems=""
> > ```
> 
> The initialization of `problems` probably belongs outside the `time` block.

Fixed, though I don't think it matters in bash since curly braces don't create a new scope or subshell.

> 
> _[ci/check-style.sh, line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4011#-M6a9zqg0miJD8-7DXwD:-M6aFqrRD95bDwAm-Fxb:bf8nem7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8982814a8fe65461f2fa02a1a9f94b34b33e3bcd/ci/check-style.sh#L63)):_
> 
> Previously, devjgm (Greg Miller) wrote…
> In a followup to that previous comment I said, _"I was unable to produce such a big change inside a docker run. -n 3 seemed to be the best, but the differentials were fractional seconds, so I'm happy to stick with -n 1."_ Take that as you see fit.

Oh, I never saw that followup. Not sure how I missed it. But thanks for checking. 1, 3, 42, whatever is fine w/ me.

Issue #3712, opened 2020-04-08
Refactor the different [Vv]ersion{Major,Minor,Patch,}() functions


coryan:
We probably want a single implementation to all these functions, this needs `-common` merged into the repository.

I do not think we should remove the existing functions, they are all in the public API and that would be a needless breaking change.

PR #4062, opened 2020-05-08
cleanup: refactor version() and version_string()

coryan:
We had multiple implementations of the version() and version_string()
functions, one on each library. This refactors the implementations to
use only the one from `google::cloud::`. I do not think we should remove
these functions, while rarely used they are part of the public API, and
we should avoid breaking changes when we can.

In the case of bigquery/ and pubsub/ we probably can remove the
functions, and I propose doing so in a followup PR.

Part of the work for #3712

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4062)

### DIFF
--- a/google/cloud/version.h
+++ b/google/cloud/version.h
@@ -64,9 +64,20 @@ int constexpr version_minor() { return GOOGLE_CLOUD_CPP_VERSION_MINOR; }
  */
 int constexpr version_patch() { return GOOGLE_CLOUD_CPP_VERSION_PATCH; }
 
+namespace internal {
+auto constexpr kMaxMinorVersions = 100;
+auto constexpr kMaxPatchVersions = 100;
+}  // namespace internal
+
 /// A single integer representing the Major/Minor/Patch version.
 int constexpr version() {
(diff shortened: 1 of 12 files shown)


google/cloud/version.h
```diff
@@ -64,9 +64,20 @@ int constexpr version_minor() { return GOOGLE_CLOUD_CPP_VERSION_MINOR; }
...
+auto constexpr kMaxPatchVersions = 100;
+}  // namespace internal
+
 /// A single integer representing the Major/Minor/Patch version.
 int constexpr version() {
-  return 100 * (100 * version_major() + version_minor()) + version_patch();
+  static_assert(version_minor() < internal::kMaxMinorVersions,
+                "version_minor() should be < kMaxMinorVersions");
+  static_assert(version_patch() < internal::kMaxPatchVersions,
+                "version_minor() should be < kMaxPatchVersions");
```
devjgm, line 77: version_minor -> version_patch

coryan: Thanks, fixed.

PR #4065, opened 2020-05-08
refactor: use version functions from common

### LINKED ISSUE #3712: Refactor the different [Vv]ersion{Major,Minor,Patch,}() functions

coryan:
For new libraries (where we have no backward compatibility constraints)
we use the `google::cloud::version*()` functions (via using
declarations). I think having these functions in the right namespace is
useful, but if somebody feels strongly we can just remove them
altogether.

This fixes #3712

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4065)


google/cloud/pubsub/version.h
```diff
@@ -42,40 +42,12 @@ namespace pubsub {
...
-  return ::google::cloud::version();
-}
 
-/// The version as a string, in MAJOR.MINOR.PATCH+gitrev format.
-std::string VersionString();
+using ::google::cloud::version;
+using ::google::cloud::version_major;
+using ::google::cloud::version_minor;
+using ::google::cloud::version_patch;
+using ::google::cloud::version_string;
```
devjgm, line 50: Given the choice, I slightly lean toward not importing these into the pubsub namespace. They're not *needed* in this namespace. Having "version" functions in separate namespaces suggests that the version numbers might be different, when they won't be. Having them in two places makes people have to decide which to use. And we could safely add them to the pubsub  namespace later if we needed to, as opposed to never being able to remove them.

But then there is the argument for consistency w/ the other libraries. Either is fine with me. Your call.

coryan: Removed. We can add them later, but when these libraries become GA there is no way to remove them.

PR #4074, opened 2020-05-10
cleanup: enable clang-tidy for `bigtable`

devbww:
Part of #3958.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4074)


google/cloud/bigtable/async_row_reader_test.cc
```diff
@@ -32,10 +32,15 @@ inline namespace BIGTABLE_CLIENT_NS {
 namespace {
 
 namespace btproto = google::bigtable::v2;
-using namespace ::testing;
-using namespace google::cloud::testing_util::chrono_literals;
+
+using ::google::cloud::testing_util::chrono_literals::operator"" _ms;
 using bigtable::testing::MockClientAsyncReaderInterface;
```
devjgm, line 37: This MockClientAsyncReaderInterface is the only using directive that's not fully qualified. It might be nice if it were also fully qualified.

Issue #3911, opened 2020-04-23
Cleanup how functions are sourced in CI scripts


coryan:
We have several `.sh` files that define functions used by other scripts, and these are getting more complex as the functions in one `.sh` file assume the functions in another `.sh` file have been loaded.

It would be nice to make these dependencies explicit.

It is also possible that this is a smell and that shell is the wrong tool for this.

devjgm:
I don't (yet) think bash scripts are the wrong tool. But we probably could improve things with a little more structure around our scripts. For example, we might consider

* Move the bash scripts that are actually _libraries_ into a `lib` directory to make it clearer what's a library and what's an executable.
* Sourcing bash libraries should probably happen near the top of the executable, like how includes work in C++
* Bash _libraries_ should use the include guard trick so they can be sourced multiple times safely
```shell
if (( CI_LIB_GCLOUD_LIB_SH__++ == 0 )); then
...
fi  # include guard
```
* Libraries could use "namespacing", like
```shell
# File: ci/lib/gcloud_lib.sh
function gcloud::create_config() {
...
}
```

PR #4076, opened 2020-05-10
feat: switch to bash library design for CI scripts

devjgm:
Part of https://github.com/googleapis/google-cloud-cpp/issues/3911

This PR is just the first step toward more refactoring of our CI scripts
to make them cleaner and more easily able to use bash "libraries". There
is a lot more cleanup that I'm eager to do following this, but this PR
is intentionally cut short so that we can discuss the design without
digging in too far.

The new design for our scripts is as follows.

* All bash scripts will be executable. All bash libraries will *not* be
executable.
* Bash libraries will live in a "lib" directory, with the primary lib
directory being `ci/lib`.
* All bash libraries will have an "include guard" so they can be sourced
potentially multiple times from any script that needs them. Note: Our
include guards return early rather than wrap the contents of the file. I
did this to prevent shfmt form indenting everything in the file.
* All bash scripts that want to use these common bash libraries will
should start by sourcing `ci/lib/init.sh`, which will set up the basic
environment for all scripts. Typically, this will be sourced as some
path relative to `dirname $0`.
* After `init.sh` has been sourced, other modules ("libraries") can be
sourced using `source module lib/foo.sh`. The `module` in that command
will know how to find the specified `lib/foo.sh`.
* All module sourcing should be at the top of the file, similar to `#include`s.

NOTE: I copied many (good?) ideas from our internal "gbash" library,
which I think is quite well designed and very nice to use.

This PR demonstrates these designs by converting the `colors.sh` library
to `ci/lib/io.sh`. I updated all the places that use colors.sh to
instead use the new library. I added all the necessary sourcing to make
this work. There is a lot more that can/should be done to simplify
things, but I needed to stop at some point to prevent the PR from
getting too big before we could even discuss the ideas in it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4076)


devjgm:
> FYI, in the past we have used the idiom described here:
> 
> https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u

Oh, nice. I'll switch to that to be consistent (and possibly more correct... wow, that SO answer was thorough).

PR #4078, opened 2020-05-10
ci: fix clang-tidy PR builds for storage/*

coryan:
When compiling PR builds we run `clang-tidy` before compiling the code,
this skips the step to download and patch the lohmann json header.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4078)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -129,6 +129,7 @@ if [[ "${CLANG_TIDY:-}" == "yes" && (\
   # w.r.t. the target branch.
   echo
   log_yellow "Running clang-tidy on presubmit build, only changed files are tested."
+  ${CMAKE_COMMAND} --build "${BINARY_DIR}" --target nlohmann_json_project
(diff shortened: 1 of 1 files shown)


ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -129,6 +129,7 @@ if [[ "${CLANG_TIDY:-}" == "yes" && (\
   # w.r.t. the target branch.
   echo
   log_yellow "Running clang-tidy on presubmit build, only changed files are tested."
+  ${CMAKE_COMMAND} --build "${BINARY_DIR}" --target nlohmann_json_project
```
devjgm, line 132: Q: Rather than running a separate cmake build here, could this block be moved down about 10 lines until after we finish the normal build? It looks like maybe that'd work? Just wondering.

coryan: I wanted the `clang-tidy` results early, I find it frustrating to wait for a full build and then discover that the formatting / clang-tidy or something like that is broken.

But I will gladly approve a PR if you or anybody else in the team feels strongly enough to make the changes.

devjgm: That reason makes sense to me. Thanks.

PR #4085, opened 2020-05-11
refactor: all docker install builds now use fedora-install

devjgm:
Note: the builds that are changing in this PR are the builds that previously used the `ubuntu-install` image. Those builds are:

```
distro=ubuntu-install name=publish-refdocs
distro=ubuntu-install name=ninja
distro=ubuntu-install name=noex
distro=ubuntu-install name=no-tests
distro=ubuntu-install name=shared
distro=ubuntu-install name=check-abi
distro=ubuntu-install name=update-abi
```
So of course, any breakage matters, but these are the builds that we're particularly paying attention to.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4085)


coryan:
I do not think `publish-refdocs` runs automatically as part of the build. Can you test it manually? And/or trigger a build in the continuous build pointing to your branch?  There is stuff in that build that might be Ubuntu specific (I doubt it, but we should check):

https://github.com/googleapis/google-cloud-cpp/blob/d9d61ea0bdbb6ddc2fbe278467e2f0496ea47716/ci/kokoro/docker/publish-refdocs.sh#L35

https://github.com/googleapis/google-cloud-cpp/blob/d9d61ea0bdbb6ddc2fbe278467e2f0496ea47716/ci/kokoro/docker/publish-refdocs.sh#L56

devjgm:
The `publish-refdocs` test failed: https://source.cloud.google.com/results/invocations/11f54664-5e56-4c3d-bc21-a4de9b010a26/targets

I'm looking into it.

devjgm:
> The `publish-refdocs` test failed: https://source.cloud.google.com/results/invocations/11f54664-5e56-4c3d-bc21-a4de9b010a26/targets
> 
> I'm looking into it.

Testing the fix now. Spanner was using the wrong variable name for the version number in its CMakeFiles.txt. The edit is now part of this PR. Note: this was broken before this PR, but we just noticed due to the explicit testing of publish-refdocs with a non-master branch.

PR #4087, opened 2020-05-12
feat!: add a dep on Abseil

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/2309

Adds a dep on Abseil. Uses `absl::make_unique` in one test in Spanner just to show that the Abseil dep is working in all the builds.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4087)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -11,11 +11,11 @@ IncludeBlocks: Merge
 IncludeCategories:
 # Matches common headers first, but sorts them after project includes
 - Regex: '^\"google/cloud/(internal/|grpc_utils/|testing_util/|[^/]+\.h)'
-  Priority: 1500
+  Priority: 1000
 - Regex: '^\"google/cloud/'  # project includes should sort first
   Priority: 500
 - Regex: '^\"'
-  Priority: 1000
+  Priority: 1500
 - Regex: '^<grpc/'
   Priority: 2000
 - Regex: '^<google/*'
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -37,6 +37,17 @@ def google_cloud_cpp_deps():
             sha256 = "d21d38c4b8e81eed8fa95ede48dd69aba01a3b938be6ac03d2b9dc61886a7183",
         )
 
+    # Load Abseil
+    if "com_google_absl" not in native.existing_rules():
+        http_archive(
+            name = "com_google_absl",
+            strip_prefix = "abseil-cpp-20200225.2",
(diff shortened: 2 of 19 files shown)


devjgm:
> > Reviewed 16 of 17 files at r1, 2 of 2 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4087)_ status: 18 of 19 files reviewed, all discussions resolved
> 
> _[.clang-format, line 17 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4087#-M77fwVp2LARlYZcea3f:-M77fwVp2LARlYZcea3g:b-49joyp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3f50b82830525e134d5594bd105454cd8515259a/.clang-format#L17)):_
> 
> > ```yaml
> > - Regex: '^\"google/cloud/'  # project includes should sort first
> >   Priority: 500
> > - Regex: '^\"'
> > ```
> 
> FYI, by including abseil using `#include "abseil/...."` we are missing out on this:
> 
> https://github.com/googleapis/google-cloud-cpp-pubsub/blob/6db6162cb6bbe9bedb04933ef8ec073386dbda7f/cmake/EnableWerror.cmake#L29-L39
> 
> That allows us to increase the warning level in our code, even if our dependencies are not ready.

Including Abseil headers with angle brackets breaks the bazel builds. I know we talked about making this work with them, but it was either forgotten or it was decided not to do it.

If our build breaks because of warnings in Abseil headers, we could file issues and escalate to them. They'll likely either need to fix the breakages quickly or perhaps enable angle-bracket includes. 

Or we could avoid the dep until they enable angle-bracket includes of their headers (Note: we currently don't support angle-bracket includes of our headers in bazel builds either).

WDYT?

> 
> _[super/CMakeLists.txt, line 55 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4087#-M77farm0opHmRtlHmMI:-M77farn3peF-LNXfVGs:bum5rhp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3f50b82830525e134d5594bd105454cd8515259a/super/CMakeLists.txt#L55)):_
> 
> > ```cmake
> > set(GOOGLE_CLOUD_CPP_DEPENDENCIES_LIST
> >     abseil-cpp-project
> > ```
> 
> FYI, when we upgrade `grpc-project` we need to change its deps to add `abseil-cpp-project` to them.

Gotcha. Thanks.

PR #4088, opened 2020-05-12
cleanup: enable clang-tidy for `bigtable/examples`

mr-salty:
part of #3958

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4088)

### DIFF
--- a/google/cloud/bigtable/examples/bigtable_instance_admin_snippets.cc
+++ b/google/cloud/bigtable/examples/bigtable_instance_admin_snippets.cc
@@ -26,8 +26,8 @@ void CreateInstance(google::cloud::bigtable::InstanceAdmin instance_admin,
   namespace cbt = google::cloud::bigtable;
   using google::cloud::future;
   using google::cloud::StatusOr;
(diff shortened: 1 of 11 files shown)


google/cloud/bigtable/examples/bigtable_instance_admin_snippets.cc
```diff
@@ -690,24 +690,26 @@ void SetNativeIamPolicy(google::cloud::bigtable::InstanceAdmin instance_admin,
...
   if (argv.size() < 3) {
     throw Usage{
         "test-iam-permissions <project-id> <resource-id>"
         " <permission> [permission ...]"};
   }
-  std::string resource = argv[1];
+  auto it = argv.cbegin();
+  auto const project_id = *it++;
+  auto const resource = *it++;
+  std::vector<std::string> const permissions(it, argv.cend());
```
devbww, line 702: I assume that we don't care too much about copying in the cases like this.  Otherwise, consider something that starts with:
```
std::rotate(argv.begin(), argv.begin() + 2, argv.end());
```

PR #4096, opened 2020-05-12
cleanup: use absl::make_unique instead of our own

devjgm:
This is used only internally. Well, it may be used in a header, but it doesn't affect any of our public APIs -- it's just a function call.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4096)


coryan:
I just sent #4097 to fix the unrelated `clang-tidy` problems.

coryan:
I think the build failures are because our CMake configuration files are missing a `find_dependency(absl)` or something similar, probably around here:

https://github.com/googleapis/google-cloud-cpp/blob/694a426ccb6e7380fe20c73cb12e6b0ad8393fc9/google/cloud/config.cmake.in#L15

devjgm:
> I think the build failures are because our CMake configuration files are missing a `find_dependency(absl)` or something similar, probably around here:
> 
> https://github.com/googleapis/google-cloud-cpp/blob/694a426ccb6e7380fe20c73cb12e6b0ad8393fc9/google/cloud/config.cmake.in#L15

Thank you. That was exactly the bug I was trying to track down -- very helpful!

PR #4097, opened 2020-05-12
ci: only clang-tidy headers in the .clang-tidy regexp

coryan:
Many of our headers are not clang-tidy clean, so we need to limit the
regular expression to whatever is on the `.clang-tidy` file.

There is some cleanup which can happen after #3958

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4097)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -129,11 +129,13 @@ if [[ "${CLANG_TIDY:-}" == "yes" && (\
   io::log "Running clang-tidy for: "
   # TODO(#3958) - use a simple regular expression like '\.(h|cc)$' when all
   # targets are clang-tidy clean.
+  RE=$(grep -o '^HeaderFilterRegex.*' "${PROJECT_ROOT}/.clang-tidy" |
+    sed -e 's/HeaderFilterRegex: "//' -e 's/"//')
+  RE="(\.cc|${RE}\.h)$"
   git diff --name-only "${KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH:-${BRANCH}}" |
-    grep -E '(\.cc|/(bigquery|firestore|pubsub|spanner)/.*\.h)$' |
-    xargs -r echo
+    grep -E "${RE}" | xargs -r echo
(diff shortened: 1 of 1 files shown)


ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -126,8 +126,14 @@ if [[ "${CLANG_TIDY:-}" == "yes" && (\
...
+  io::log "Running clang-tidy for: "
+  # TODO(#3958) - use a simple regular expression like '\.(h|cc)$' when all
+  # targets are clang-tidy clean.
   git diff --name-only "${KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH:-${BRANCH}}" |
-    grep -E '\.(cc|h)$' |
+    grep -E '(\.cc|/(bigquery|firestore|pubsub|spanner)/.*\.h)$' |
+    xargs -r echo
+  git diff --name-only "${KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH:-${BRANCH}}" |
+    grep -E '(\.cc|/(bigquery|firestore|pubsub|spanner)/.*\.h)$' |
     xargs -d '\n' -r -n 1 -P "${NCPU}" clang-tidy -p="${BINARY_DIR}"
```
devjgm, line 137: I assumed that the invocation of `clang-tidy` would have it read the `.clang-tidy` file at the root of our repo and ignore all files and headers that it's supposed to ignore. That's not happening? Boo clang-tidy.

You might be able to use the regex at `cat "${PROJECT_ROOT}/.clang-tidy" | grep HeaderFilterRegex`

devjgm: Actually, consider this optional.

coryan: > I assumed that the invocation of `clang-tidy` would have it read the `.clang-tidy` file at the root of our repo and ignore all files and headers that it's supposed to ignore. That's not happening?

AFAICT, no, it does not happen.

> Boo clang-tidy.

Well, this was always going to be messy.


> You might be able to use the regex at `cat "${PROJECT_ROOT}/.clang-tidy" | grep HeaderFilterRegex`

Done, but please check my math here, regexps in variables scare me.

PR #4098, opened 2020-05-12
test: update rules about running integration tests

### LINKED ISSUE #3869: Make integration test running work more like spanner
spanner's `ci/kokoro/docker/build.sh` has a couple small differences I think we should incorporate here:

* If the build name is explicitly `integration` (and perhaps `integration-nightly`), `RUN_INTEGRATION_TESTS` should be set to `yes` instead of `auto`.  Doing that makes the integration test fail instead of being skipped if there is something wrong with the environment. I think that's preferable given the user wanted the tests to run.
* If the user explicitly sets RUN_INTEGRATION_TESTS in the environment, continue to honor that unconditionally (currently the environment variable is overridden).

For reference, see https://github.com/googleapis/google-cloud-cpp-spanner/pull/619 (IIRC I intended to "port" that to the other repos but must have never finished it)

mr-salty:
* Honor `RUN_INTEGRATION_TESTS` if set in the environment.
* Always run integration tests for `integration*` builds.
* Default to not running integration tests for `clang-tidy`,
  `abi-check`, and `abi-update`.
* Default to running integration tests for `clang-3.8` and `gcc-4.8`.

Fixes #3869

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4098)


mr-salty:
Moved this back out of draft, PTAL.  I plan to check in some retries for #4113 before I check this in, to avoid test flakiness.  I spot-checked 5 build logs to make sure we are or are not running integration tests as expected.

ci/kokoro/docker/build.sh
```diff
@@ -226,16 +220,29 @@ elif [[ "${BUILD_NAME}" = "cxx17" ]]; then
...
   in_docker_script="ci/kokoro/docker/build-in-docker-quickstart-cmake.sh"
 else
   echo "Unknown BUILD_NAME (${BUILD_NAME}). Fix the Kokoro .cfg file."
   exit 1
 fi
 
+# If RUN_INTEGRATION_TESTS wasn't set in the environment or by one of the
+# builds above, default it to "yes" for kokoro builds and "auto" otherwise.
+if [[ -z "${RUN_INTEGRATION_TESTS:-}" ]]; then
+  if [[ -n "${KOKORO_JOB_NAME:-}" ]]; then
```
devjgm, line 238: Rather than consulting `KOKORO_JOB_NAME` as a proxy for whether this is running as a CI build, you might use `RUNNING_CI`

PR #4099, opened 2020-05-12
fix: proper routing headers for longrunning ops

### LINKED ISSUE #2990: Calls to `GetOperation` set wrong `x-goog-request-params`.
Currently, they're using the same headers as the operation which led to creating the longrunning operation. Instead, to should be `name=operation_name`.

dopiera:
- The polling loop was refactoring some code, and that made it hard to set the metadata attributes
- Changing the polling loop to take the `longrunning::Operation` proto allows you to set the metadata attributes
- After changing the metadata attributes, I wanted to add a test that it's properly set and noticed that there are no tests anymore
- I added the tests back
- After adding the tests, I could now check if the newly changed metadata behavior is proper
- The `IsContextMDValid()` turned out to have not worked for `ClientContexts` whose deadline is in the past (it took me a while to figure it out, BTW), so I fixed that too.

This fixes #2990.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4099)


dopiera:
FWIW, I've split the reverts and the changes into multiple commits for easier review.

PR #4111, opened 2020-05-13
cleanup: enable clang-tidy for `bigtable` headers

devbww:
Note that some "performance-unnecessary-value-param" NOLINT
directives are only due to an apparent clang-tidy bug, and
are tagged as "TODO(#4112)" to make their later removal easy.

Part of #3958.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4111)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -56,7 +56,7 @@ Checks: >
 WarningsAsErrors: "*"
 
 # TODO(#3920) - Enable clang-tidy checks in our headers.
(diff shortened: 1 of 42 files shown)


coryan:
I think `TODO(#...)` or even `#...` (instead of BOGUS) pointing to a GitHub issue explaining what the bug is, and what the cleanup needs to be would be easier to understand by our future selves. Is that possible?

devbww:
Yes.  Good idea.  Done.  Thanks.

PR #4133, opened 2020-05-15
fix: enable grpc_utils if any gRPC-based library is

### LINKED ISSUE #3926: Compile only the common components that are needed
With CMake, customers may disable compiling specific libraries, say because they
only want to compile `google/cloud/storage` and do not want to compile the
other dependencies (gRPC, Protobuf, etc.)

After we import `google-cloud-cpp-common` we will always compile all the
common components, even if they are not needed by the libraries that
the customer selected. It would be nice to disable the common
components that are not needed.

This may be as simple as adding the right `EXCLUDE_FROM_ALL` options
into the CMake files, or may require some more fine tuning.

### LINKED ISSUE #4130: Pub/Sub can't build without Bigtable
(First, I am aware that the Pub/Sub is still experimental, thanks for working on it)

The Pub/Sub library seems to depend on the gRPC utils library but the only way to enable it is via the `GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS` which depends on `GOOGLE_CLOUD_CPP_ENABLE_BIGTABLE` being enabled.

**To Reproduce**
1. Enable `GOOGLE_CLOUD_CPP_ENABLE_PUBSUB`
2. Disable `GOOGLE_CLOUD_CPP_ENABLE_BIGTABLE`
3. Build
4. Link error:
```
[ 50%][1/2] Linking CXX shared library google/cloud/pubsub/libpubsub_client.so.
FAILED: google/cloud/pubsub/libpubsub_client.so. 
: && /usr/lib/ccache/clang++-10 -fPIC -fcolor-diagnostics -O3 -DNDEBUG   -shared -Wl,-soname,libpubsub_client.so.1 -o google/cloud/pubsub/libpubsub_client.so. google/cloud/pubsub/CMakeFiles/pubsub_client.dir/connection_options.cc.o google/cloud/pubsub/CMakeFiles/pubsub_client.dir/internal/publisher_stub.cc.o google/cloud/pubsub/CMakeFiles/pubsub_client.dir/internal/subscriber_stub.cc.o google/cloud/pubsub/CMakeFiles/pubsub_client.dir/internal/user_agent_prefix.cc.o google/cloud/pubsub/CMakeFiles/pubsub_client.dir/publisher_client.cc.o google/cloud/pubsub/CMakeFiles/pubsub_client.dir/publisher_connection.cc.o google/cloud/pubsub/CMakeFiles/pubsub_client.dir/subscriber_client.cc.o google/cloud/pubsub/CMakeFiles/pubsub_client.dir/subscriber_connection.cc.o google/cloud/pubsub/CMakeFiles/pubsub_client.dir/subscription.cc.o google/cloud/pubsub/CMakeFiles/pubsub_client.dir/topic.cc.o  -Wl,-rpath,/home/sylvain/prog/google-cloud-cpp/build/google/cloud:  -lgoogle_cloud_cpp_grpc_utils  google/cloud/libgoogle_cloud_cpp_common.so.1.14.0  /usr/lib/x86_64-linux-gnu/libgoogleapis_cpp_pubsub_protos.so.0.10.0  /usr/lib/x86_64-linux-gnu/libgoogleapis_cpp_api_annotations_protos.so.0.10.0  /usr/lib/x86_64-linux-gnu/libgoogleapis_cpp_api_http_protos.so.0.10.0  /usr/lib/x86_64-linux-gnu/libgoogleapis_cpp_api_client_protos.so.0.10.0  /usr/lib/x86_64-linux-gnu/libgoogleapis_cpp_api_field_behavior_protos.so.0.10.0  /usr/lib/x86_64-linux-gnu/libgoogleapis_cpp_api_resource_protos.so.0.10.0  /usr/lib/x86_64-linux-gnu/libgrpc++.so  /usr/lib/x86_64-linux-gnu/libgrpc.so  /usr/lib/x86_64-linux-gnu/libprotobuf.so  -lpthread && :
/usr/bin/ld : ne peut trouver -lgoogle_cloud_cpp_grpc_utils
clang: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

**Expected behavior**
Being able to build the Pub/Sub library with Bigtable disabled. Specifying `GOOGLE_CLOUD_CPP_ENABLE_PUBSUB` in the dependent option condition should do the trick.

**Operating system:**
* Linux Debian testing/bullseye
* Commit 72a6a0ab0e8f2e410af9bf39d47d0443c7f36d14
* CMake 3.17.1

coryan:
The `cmake_dependent_option` expression was incomplete.

Fixes #4130 and fixes #3926

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4133)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,11 +138,25 @@ option(GOOGLE_CLOUD_CPP_ENABLE_FIRESTORE
        "Enable building the Firestore library." ON)
 option(GOOGLE_CLOUD_CPP_ENABLE_PUBSUB "Enable building the Pub/Sub library." ON)
 
+# Compute the default value for GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS, use a
+# function to create a scope for the temporary variables.
+function(google_cloud_cpp_grpc_utils_expression)
+    set(expression "GOOGLE_CLOUD_CPP_ENABLE_BIGQUERY")
+    foreach (lib BIGTABLE SPANNER PUBSUB)
+        string(APPEND expression " OR GOOGLE_CLOUD_CPP_ENABLE_${lib}")
+    endforeach()
+    set(GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS_EXPRESSION "${expression}" PARENT_SCOPE)
+endfunction()
+google_cloud_cpp_grpc_utils_expression()
+
 cmake_dependent_option(
-    GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS
-    "Enable building the gRPC utilities library." ON
-    "GOOGLE_CLOUD_CPP_ENABLE_BIGTABLE" OFF)
+        GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS
+        "Enable building the gRPC utilities library." ON
+        "${GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS_EXPRESSION}" OFF)
 mark_as_advanced(GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS)
+
+message(STATUS "GRPC_UTILS = ${GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS}")
+message(STATUS "GRPC_UTILS_DEFAULT = ${GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS_EXPRESSION}")
 
 # Enable testing in this directory so we can do a top-level `make test`. This
 # also includes the BUILD_TESTING option, which is on by default.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,23 +140,18 @@ option(GOOGLE_CLOUD_CPP_ENABLE_PUBSUB "Enable building the Pub/Sub library." ON)
 
 # Compute the default value for GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS, use a
 # function to create a scope for the temporary variables.
-function(google_cloud_cpp_grpc_utils_expression)
-    set(expression "GOOGLE_CLOUD_CPP_ENABLE_BIGQUERY")
-    foreach (lib BIGTABLE SPANNER PUBSUB)
-        string(APPEND expression " OR GOOGLE_CLOUD_CPP_ENABLE_${lib}")
-    endforeach()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -138,12 +138,26 @@ option(GOOGLE_CLOUD_CPP_ENABLE_FIRESTORE
...
+# Compute the default value for GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS, use a
+# function to create a scope for the temporary variables.
+function(google_cloud_cpp_grpc_utils_expression)
+    set(expression "GOOGLE_CLOUD_CPP_ENABLE_BIGQUERY")
+    foreach (lib BIGTABLE SPANNER PUBSUB)
+        string(APPEND expression " OR GOOGLE_CLOUD_CPP_ENABLE_${lib}")
+    endforeach()
+    set(GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS_EXPRESSION "${expression}" PARENT_SCOPE)
+endfunction()
+google_cloud_cpp_grpc_utils_expression()
```
devjgm, line 150: One thought: This function might be more complicated than something like

```cmake
string(CONCAT GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS_EXPRESSION
    "GOOGLE_CLOUD_CPP_ENABLE_BIGQUERY OR "
    "GOOGLE_CLOUD_CPP_ENABLE_BIGTABLE OR "
    "GOOGLE_CLOUD_CPP_ENABLE_SPANNER OR "
    "GOOGLE_CLOUD_CPP_ENABLE_PUBSUB"
)
```
and this version is more grep-able so we hopefully might not miss it the next time we're adding things.

Totally optional. Your call about which you prefer.

coryan: I like, thanks!

BTW, `string(JOIN...)` would have been nice, but does not exist in CMake 3.5 ☹️

CMakeLists.txt
```diff
@@ -138,10 +138,19 @@ option(GOOGLE_CLOUD_CPP_ENABLE_FIRESTORE
        "Enable building the Firestore library." ON)
 option(GOOGLE_CLOUD_CPP_ENABLE_PUBSUB "Enable building the Pub/Sub library." ON)
 
+# Compute the default value for GOOGLE_CLOUD_CPP_ENABLE_GRPC_UTILS, use a
+# function to create a scope for the temporary variables.
```
devjgm, line 142: nit: this comment about using a function no longer applies.

coryan: Argh, will send a new PR to fix that.

PR #4137, opened 2020-05-15
ci: log the clang-tidy config

devbww:
This helps make tracking of clang-tidy warnings a little easier.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4137)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -118,6 +118,11 @@ ${CMAKE_COMMAND} \
 echo
 io::log_yellow "Finished CMake config"
 
+if [[ "${CLANG_TIDY:-}" = "yes" ]]; then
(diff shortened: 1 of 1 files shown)


ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -118,6 +118,11 @@ ${CMAKE_COMMAND} \
 echo
 io::log_yellow "Finished CMake config"
 
+if [[ "${CLANG_TIDY:-}" = "yes" ]]; then
+  io::log_yellow "clang-tidy config:"
+  clang-tidy -dump-config
```
devjgm, line 123: fwiw, this dumps over 300 lines of config output on all `clang-tidy` builds. An alternative might be to run `ci/kokoro/docker/build.sh clang-tidy clang-tidy --dump-config` when we need to see the output exactly as the build would see it.

Issue #4119, opened 2020-05-13
[Flake] Standard C++ exception raised: Socket closed


devjgm:
Here's a flake from the `macos/cmake-super-presubmit` build ([full log](https://source.cloud.google.com/results/invocations/43cad785-afe5-4c78-b662-1dd4f2b39d4c/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fmacos%2Fcmake-super-presubmit/log)).

```
 89/105 Test #268: bigtable_examples_data_snippets .................................***Failed   19.78 sec
Running MutateInsertUpdateRows() example [1]
Row successfully updated
Running MutateInsertUpdateRows() example [2]
Row successfully updated
Running WriteSimple() example
Successfully wrote rowphone#4c410523#20190501
Running WriteBatch() example
Successfully wrote 2 rows.
Running WriteIncrement() example
Successfully updated rowphone#4c410523#20190501
Running WriteConditionally() example
Successfully updated row
Creating table tbl-data-1589403548-tdnncews

Preparing data for MutateDeleteColumns()
Row successfully updated
Running MutateDeleteColumns() example
Columns successfully deleted from row
Running MutateDeleteRows() example [1]
All rows successfully deleted

Preparing data for MutateDeleteRows()
Row successfully updated
Row successfully updated
Running MutateDeleteRows() example [2]
All rows successfully deleted

Preparing data for RenameColumn()
Row successfully updated
Row successfully updated
Row successfully updated
Running RenameColumn() example
Row successfully updated

Preparing data for multiple examples
Running Apply() example
Running Apply() with relaxed idempotency example
Running Apply() with custom retry example
Running BulkApply() example
All mutations applied successfully

Populate data for prefix and row set examples
Standard C++ exception raised: Socket closed

        Start 338: spanner_database_admin_integration_test
```

I seem to recall @devbww seeing similar issues recently also. I think they may have been on his Linux laptop, so perhaps not related to macOS.

devjgm:
Another one, this time from the macOS bazel build.
https://source.cloud.google.com/results/invocations/510e852f-b29d-4b19-a788-8c42c15eb892/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fmacos%2Fbazel-presubmit/log

coryan:
And here is a FD shutdown....

[...]

Issue #3990, opened 2020-05-01
Use Abseil stacktrace reporting in tests


devjgm:
If/when we explicitly take a dep on Abseil (even just for unit tests), I think we can configure Abseil+googletest to handle dumping stacktraces when segfaults and other crashes happen. Some info here: https://github.com/abseil/abseil-cpp/blob/master/absl/debugging/failure_signal_handler.h

This would be useful for us on all platforms.

devjgm:
We have a dep on Abseil now, so it should be possible and useful to enable good stack trace reporting in all of our builds on all platforms.

devjgm:
Doing this might help us catch https://github.com/googleapis/google-cloud-cpp/issues/4119, so I'm going to take a stab at this issue.

devjgm:
We now have https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/testing_util/crash_handler.h, which we can optionally call from `main()` when running a unit test in order to install Abseil's crash handlers. I don't know of any good, portable way to do this all the time for all unit tests (whether or not we provide a custom main). So we can call this function if/when we need it. I think this wraps up the spirit of this issue.

PR #4150, opened 2020-05-17
feat: add absl crash handler support for bigtable examples

devjgm:
Part of https://github.com/googleapis/google-cloud-cpp/issues/3990
Might help diagnose https://github.com/googleapis/google-cloud-cpp/issues/4119

This PR adds a crash handler function that must be called early in `main()` to be used. This function will print a stack trace for any segfault, etc, and uncaught stack trace. This means that caught stack traces will not result in stack traces. I instrumented all of bigtable's example's main functions to install this crash handler. 

Note: I also needed to disable the clang-tidy naming check for macros, because this code relies on a clang macro name that violates our style guide.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4150)


devjgm:
> One suggestion to make this easier to use in examples. I will approve in case you want to do that in a future PR.
> 
> > Reviewed 27 of 27 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4150#-:-M7hYkXk1nT-ZU486U3H:b-o4e6at)_ status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)
> 
> _[google/cloud/bigtable/examples/bigtable_hello_table_admin.cc, line 151 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4150#-M7hTwIPBXvl7lbpKlAc:-M7hTwIPBXvl7lbpKlAd:b-8dj4qw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5011bf0fbd9957a2c5460a21232aa85358dea8e5/google/cloud/bigtable/examples/bigtable_hello_table_admin.cc#L151)):_
> 
> > ```c++
> > int main(int argc, char* argv[]) {
> >   google::cloud::testing_util::InstallCrashHandler(argv[0]);
> > ```
> 
> If we are going to do this for each example, why not do it in `Example::Run()` or even in the constructor?

I thought about that, but favored this approach for now because I think it's more explicit and possibly safer, if a bit more verbose. Installing a process-wide crash handler from a normal function (either the c'tor or Run()) seems strange to me. The crash handler is essentially unrelated to the ExampleDriver, so it's not clear why _that_ class would be designated as the one to install the crash handler. What if the example driver functions are called multiple times? Abseil says [don't install their failure handlers multiple times](https://github.com/abseil/abseil-cpp/blob/3f347c46272886a099852a4cd303ecf37a054de8/absl/debugging/failure_signal_handler.h#L39). And even without that warning from Abseil, it seemed weird because the crash handler isn't specific to the Example runner; we just so happen to be installing it in the binaries that run the examples currently. What would happen if we wanted to run the Examples in some other binary (maybe a unit test) that also wanted to install the crash handler?

There are likely solutions to most/all of what I mention above. But until we answer those questions, the simplest and safest thing seemed to simply install the crash handler as the first line in main().

WDYT?

> 
> _[super/external/abseil.cmake, line 46 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4150#-M7hWZFp0w8pI5Vu5cd2:-M7hWZFp0w8pI5Vu5cd3:b-1dwneh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5011bf0fbd9957a2c5460a21232aa85358dea8e5/super/external/abseil.cmake#L46)):_
> 
> > ```cmake
> >                    -DBUILD_TESTING=OFF
> >                    -DCMAKE_PREFIX_PATH=${GOOGLE_CLOUD_CPP_PREFIX_PATH}
> >                    -DCMAKE_INSTALL_RPATH=${GOOGLE_CLOUD_CPP_INSTALL_RPATH}
> > ```
> 
> FYI: I assume you cut&pasted `abseil.cmake` from some other source? That means the other source also lacks RPATH?

I copied it from here: https://github.com/coryan/google-cloud-cpp-common/blob/prototype-abseil/super/external/abseil.cmake

devjgm:
> Maybe create a standalone function `int testing_util::RunWithCrashHandler(Example& ex, int argc, char* argv[])` and call that from each example? Completely optional.

Unless we think this needs to be fixed in this PR, I filed https://github.com/googleapis/google-cloud-cpp/issues/4236 so we can discuss this, track it, and hopefully fix it in a future PR. SG?

Issue #4151, opened 2020-05-17
Should "examples" be able to use "testing_util"?


devjgm:
https://github.com/googleapis/google-cloud-cpp/pull/4150 is a draft PR adding a `testing_util/crash_handler` library. This is a library that our production code does not need, but we would like to use in our tests and example, i.e., the binaries that we produce and run for our own testing. So "testing_util" seemed like a reasonable spot.

However, the `bigtable/examples` binaries cannot depend on `google_cloud_cpp_testing` because the examples are always built, but the latter is only built if `BUILD_TESTING=ON` (there's also a `GOOGLE_CLOUD_CPP_TESTING_UTIL_ENABLE_INSTALL` option, but it's not clear that we want to install the testing utils just for the examples).

So some questions:

1. Should our examples (bigtable/examples, storage/examples, etc.) all be wrapped in a `if (BUILD_TESTING)`? This may seem reasonable, because we'd typically only build with `BUILD_TESTING=OFF` if we're building release libraries to be installed, and in this case, I don't think the examples need to be there. Do we ever _need_ the examples to be built when we're building with `BUILD_TESTING=OFF`?

2. Should we move the `common/testing_utils/crash_handler` library into `common/internal/crash_handler`? This would allow the examples to use them because `google_cloud_cpp_common` is always built and installed. The downside here is that our crash_reporter library would always be built and linked into our customer's binaries, when it will (probably) never be used. We might decide to do this at some point in the future, but if we do, we should decide to do so because there's a customer-benefit, not just a jgm-is-trying-to-get-the-examples-to-compile reason.

3. Should I somehow use the `GOOGLE_CLOUD_CPP_TESTING_UTIL_ENABLE_INSTALL` option to cause the testing_util library to be installed? If we do this, it seems like it'd always be installed, and we probably don't want it to be.

Maybe there are more options. I'm thinking option 1 seems the best, but I may not understand the current situation and all its implications well enough at this point. Option 2 doesn't sound terrible either.

coryan:
> (there's also a `GOOGLE_CLOUD_CPP_TESTING_UTIL_ENABLE_INSTALL` option, but it's not clear that we want to install the testing utils just for the examples).

Ugh, that was so `-common` could be compiled without its own tests, but with the testing_util library for its dependencies.

coryan:
I vote for option (1) too.

coryan:
[...]

PR #4223, opened 2020-05-18
cleanup: move example_driver to testing_util/

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/4151

Moved the `example_driver.*` code from `google/cloud/internal` -> `google/cloud/testing_util/`. This makes it clearer that it's intended only for our "tests", and not for production code. This also avoids th need to compile and install the example driver.

I also changed bigtable and storage "examples" to only compile `if (BUILD_TESTING)`. This avoids the need to compile them when a user is trying to compile a release build and install it. 

I also cleaned up the old `GOOGLE_CLOUD_CPP_TESTING_UTIL_ENABLE_INSTALL` option, which is no longer needed since `-common` merged into the monorepo.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4223)


devjgm:
Rebased.

PR #4225, opened 2020-05-18
fix: clang-tidy HeaderFilterRegex was not working

devjgm:
It's not always safe to append "\.h$" to a regex, so this PR changes our
clang-tidy filtering to two invocations. One to handle all .cc files,
where it can ignore the HeaderFitlerRegex. And a second invocation that
applies to all .h files that also match HeaderFitlerRegex.

Related: This should fix issues like [this one](https://source.cloud.google.com/results/invocations/bae42914-9b04-463a-9b8e-3156ce046a40/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fdocker%2Fclang-tidy-presubmit/log) where we were running clang-tidy on BUILD and CMakeLists.txt files.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4225)


devjgm:
Actually, this doesn't quite work. Need to try a few more things.

devjgm:
Nm. I believe it is working.

PR #4226, opened 2020-05-18
cleanup: enable clang-tidy for `storage` headers

devbww:
Adding `\\.h$` to `HeaderFilterRegex` eliminates checking of `nlohmann_json.hpp`.

Part of #3958.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4226)

### DIFF
--- a/google/cloud/storage/parallel_upload.h
+++ b/google/cloud/storage/parallel_upload.h
@@ -42,6 +42,7 @@ inline namespace STORAGE_CLIENT_NS {
  */
 class MaxStreams {
  public:
+  // NOLINTNEXTLINE(google-explicit-constructor)
   MaxStreams(std::size_t value) : value_(value) {}
   std::size_t value() const { return value_; }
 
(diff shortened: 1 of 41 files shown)


google/cloud/storage/parallel_upload.h
```diff
@@ -953,16 +962,16 @@ std::vector<std::uintmax_t> ComputeParallelFileUploadSplitPoints(
   };
   // These defaults were obtained by experiments summarized in
   // https://github.com/googleapis/google-cloud-cpp/issues/2951#issuecomment-566237128
-  MaxStreams const kDefaultMaxStreams(64);
-  MinStreamSize const kDefaultMinStreamSize(32 * 1024 * 1024);
+  MaxStreams const k_default_max_streams(64);
```
devjgm, line 965: This looks like a style that I"m not familiar with. Could these just be named like `default_max_streams`?

devbww: Yes.  clang-tidy suggested the `k_` form, but it is fine without it too.

PR #4229, opened 2020-05-18
feat: implement ReadLast option for GCS+gRPC plugin

coryan:
For applications that want to read the last N bytes of a GCS object.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4229)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -1011,6 +1011,10 @@ google::storage::v1::GetObjectMediaRequest GrpcClient::ToProto(
     r.set_read_offset(range.begin);
     r.set_read_limit(range.end - range.begin);
   }
+  if (request.HasOption<ReadLast>()) {
+    auto const offset = request.GetOption<ReadLast>().value();
+    r.set_read_offset(-(offset + 1));
+  }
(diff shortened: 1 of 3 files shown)


google/cloud/storage/internal/grpc_client.cc
```diff
@@ -1011,6 +1011,10 @@ google::storage::v1::GetObjectMediaRequest GrpcClient::ToProto(
     r.set_read_offset(range.begin);
     r.set_read_limit(range.end - range.begin);
   }
+  if (request.HasOption<ReadLast>()) {
+    auto const offset = request.GetOption<ReadLast>().value();
+    r.set_read_offset(-(offset + 1));
```
devbww, line 1016: I don't know enough to have any question about its correctness, but the `1` is surprising.  Does it deserve a note?

coryan: Thanks for asking, I agree. I have followed up with the GCS+gRPC folks and created a bug and TODO to track this down.  I am planning to merge as-is, because it is good to have the functionality, even if the implementation for it is quirky.

devbww: SGTM.  Thanks.

PR #4231, opened 2020-05-18
cleanup: `bigtable/benchmarks` is clang-clean after #4138

devbww:
The presubmit tests won't verify that, but I'll assert it,
and we'll learn whether I'm wrong as future edits are made.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4231)


devjgm:
Brad you can merge first. FTR, I also think benchmarks is clean. I think you're right.

PR #4237, opened 2020-05-19
fix: GCS+gRPC use the documented behavior for ReadLast

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4237)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -1025,7 +1025,7 @@ google::storage::v1::GetObjectMediaRequest GrpcClient::ToProto(
   }
   if (request.HasOption<ReadLast>()) {
     auto const offset = request.GetOption<ReadLast>().value();
-    r.set_read_offset(-(offset + 1));
+    r.set_read_offset(-offset);
   }
   if (request.HasOption<ReadFromOffset>()) {
     auto const offset = request.GetOption<ReadFromOffset>().value();
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -229,6 +229,17 @@ StatusOr<ObjectMetadata> GrpcClient::GetObjectMetadata(
 
 StatusOr<std::unique_ptr<ObjectReadSource>> GrpcClient::ReadObject(
     ReadObjectRangeRequest const& request) {
+  // With the REST API this condition was detected by the server as an error,
+  // generally we prefer the server to detect errors because its answers are
+  // authoritative. In this case, the server cannot: with gRPC '0' is the same
+  // as "not set" and the server would send back the full file, which was
+  // unlikely to be the customer's intent.
+  if (request.HasOption<ReadLast>() &&
+      request.GetOption<ReadLast>().value() == 0) {
+    return Status(
+        StatusCode::kOutOfRange,
+        "ReadLast(0) is invalid in REST and produces incorrect output in gRPC");
+  }
   auto const proto_request = ToProto(request);
   auto create_stream = [&proto_request, this](grpc::ClientContext& context) {
     return stub_->GetObjectMedia(&context, proto_request);
--- a/google/cloud/storage/CMakeLists.txt
+++ b/google/cloud/storage/CMakeLists.txt
@@ -479,11 +479,11 @@ if (GOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC)
     if (BUILD_TESTING)
         set(storage_client_grpc_unit_tests
             # cmake-format: sort
-            internal/grpc_client_failures_test.cc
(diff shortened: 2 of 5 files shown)


coryan:
Agree.

google/cloud/storage/internal/grpc_client.cc
```diff
@@ -1013,7 +1013,7 @@ google::storage::v1::GetObjectMediaRequest GrpcClient::ToProto(
   }
   if (request.HasOption<ReadLast>()) {
     auto const offset = request.GetOption<ReadLast>().value();
-    r.set_read_offset(-(offset + 1));
+    r.set_read_offset(-offset);
```
devbww, line 1016: As an example of what might now be a problem, consider "ReadLast 0".

I should have thought a little harder in the first place.  Sorry.

coryan: If you mean that passing `-0` to `set_read_offset()` will do something weird unexpected, then I agree, we should returns an error.

If you mean: there should be a way to read "0 bytes from the end" then I do not. That is useless. If you want to read nothing there are cheaper ways to do that.

Or do a `set_read_offset(0); set_read_limit(0)`. That is *also* 0 bytes and all empty strings, no matter where they come from, look the same to me.

devbww: > If you mean that passing -0 to set_read_offset() will do something weird unexpected, then I agree, we should returns an error.

Yes, that is mostly what I mean.  "ReadLast N" used to do something related to its name for non-negative N, but now it only does that for positive N.  So, we probably should catch `offset == 0` somehow.

> If you mean: there should be a way to read "0 bytes from the end" then I do not. That is useless. If you want to read nothing there are cheaper ways to do that.

I'm not so much concerned about someone explicitly saying "read 0 bytes from the end", but rather someone who has computed the offset and in some particular case it (legitimately) turns out to be zero.  They probably should read from the end instead of from the beginning.

In the bigger picture, I'm not sure how they can just change the meaning of their API.  Is this an unreleased feature?

Anyway, as usual, these are thoughts from 10,000 feet.  Your context/experience trumps that.  Thanks.

coryan: > > If you mean that passing -0 to set_read_offset() will do something weird unexpected, then I agree, we should returns an error.
> 
> Yes, that is mostly what I mean. "ReadLast N" used to do something related to its name for non-negative N, but now it only does that for positive N. So, we probably should catch `offset == 0` somehow.

Ack.  Generally we have not validated inputs in the client-side (with very few exceptions). The server knows best and it can tell us if we did something wrong.

> 
> > If you mean: there should be a way to read "0 bytes from the end" then I do not. That is useless. If you want to read nothing there are cheaper ways to do that.
> 
> I'm not so much concerned about someone explicitly saying "read 0 bytes from the end", but rather someone who has computed the offset and in some particular case it (legitimately) turns out to be zero. 

Yeah, that would be a bit shocking, you would read the whole file.

>They probably should read from the end instead of from the beginning.

I would argue that reading 0 bytes at the beginning or 0 bytes at the end is the same. OTOH, reading the whole file when you said `ReadLast(my_variable_expected_to_be_small)` is quite unexpected.

> 
> In the bigger picture, I'm not sure how they can just change the meaning of their API. Is this an unreleased feature?

gRPC is not released. The REST version has been out for years, and has the same semantics (-1 means read 1 bytes from the end), it cannot be changed:

```
curl -H "Authorization: Bearer $(gcloud auth print-access-token)" -H "Range: bytes=-4"   'https://storage.googleapis.com/coryan-test-bucket/fox.txt'
dog
```

(the newline at EOF is the fourth character).

Now, because REST is text based you can I least write "Range: bytes=-0", but does not work:

```
curl -H "Authorization: Bearer $(gcloud auth print-access-token)" -H "Range: bytes=-0"   'https://storage.googleapis.com/coryan-test-bucket/fox.txt'
```

```
<?xml version='1.0' encoding='UTF-8'?><Error><Code>InvalidRange</Code><Message>The requested range cannot be satisfied.</Message><Details>bytes=-0</Details></Error>
```

That is, there is no way in the existing APIs to request "0 bytes at the end".

> Anyway, as usual, these are thoughts from 10,000 feet. Your context/experience trumps that.

I need to think about error reporting with `ReadLast(0)`... it seems really odd to blindly return the whole file (potentially 5TiB worth!)

devbww: > there is no way in the existing APIs to request "0 bytes at the end"

Yes, I think that is the crux of it.  Thanks.  Indeed, there is no way to request "0 bytes" anywhere ... the first and last positions are inclusive, and the last must be at least the first.

So, I agree that `offset == 0` is the only error case that need to be handled on the client side.

coryan:
Rebased to resolve conflicts.

PR #4239, opened 2020-05-20
feat: add rvalue reference overloads to `Row`

### LINKED ISSUE #4107: Provide rvalue-ref overloads for `Row` accessors
This is needed in order to allow for passing the underlying data by move rather than by cref, which usually implies copying.

dopiera:
This fixes #4107.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4239)

### DIFF
--- a/google/cloud/bigtable/row.h
+++ b/google/cloud/bigtable/row.h
@@ -43,7 +43,9 @@ class Row {
   RowKeyType const& row_key() const { return row_key_; }
 
   /// Return all cells.
-  std::vector<Cell> const& cells() const { return cells_; }
+  std::vector<Cell> const& cells() const& { return cells_; }
+  /// Return all cells.
+  std::vector<Cell>&& cells() && { return std::move(cells_); }
 
  private:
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/row.h
```diff
@@ -40,10 +40,14 @@ class Row {
 
   /// Return the row key. The returned value is not valid
   /// after this object is deleted.
-  RowKeyType const& row_key() const { return row_key_; }
+  RowKeyType const& row_key() const& { return row_key_; }
+  /// Return the row key.
+  RowKeyType row_key() && { return std::move(row_key_); }
```
devjgm, line 45: This is returning a value, but I think rvalue overloaded methods should return `T&&` not `T`. The difference is that as written, it will return a move-constructed `T`, but if the signature returns a `T&&`, it will not even move-construct a `T`, it'll simply return the rvalue reference.

dopiera: I learn everyday.

PR #4245, opened 2020-05-20
feat: remove dep on cpp-cmakefiles, integrating the CMake rules into this repo

devjgm:
Removes our dep on the `cpp-cmakefiles` repo, moving all the CMake files and logic into this repo. The CMake files in this PR were copied from the `cpp-cmakefiles` repo, and they do the same thing (a few variable names were changed/removed because they already exist).

Fixes: https://github.com/googleapis/google-cloud-cpp/issues/4063

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4245)


devjgm:
> I really like this, maybe we can simplify it even further? The double installation for `FindgRPC` should be dealt with in this PR, the other stuff can wait for a followup PR too.
> 
> > Reviewed 28 of 28 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4245#-:-M7nhYwl-xBc3QN5ac_e:b-8fsbd2)_ status: all files reviewed, 2 unresolved discussions (waiting on @devjgm)
> 
> _[cmake/GoogleApisProtos.cmake, line 26 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4245#-M7nfuqN5FGf3IHzToIS:-M7nfuqO3Gh1X7AoiVql:b-11o6z3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dba688e18a770d4905b5f955489862a3ae7ef084/cmake/GoogleApisProtos.cmake#L26)):_
> 
> > ```cmake
> > # actual proto files.
> > 
> > file(COPY "${CMAKE_CURRENT_LIST_DIR}/googleapis/CMakeLists.txt"
> > ```
> 
> So, I absolutely do not know if this will work: why copy these files? Why not have `googleapis` as a directory at the top-level and just use `add_subdirectory(googleapis)`?

Great idea. I _think_ that will work. The builds are super slow on my laptop, but it appears like it's working, so I made the change and kicked off a presbumit build. We'll see. Anyway, this is a much simpler arrangement.

> 
> _[cmake/googleapis/CMakeLists.txt, line 459 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4245#-M7nga910zWqBrsRlBEs:-M7nga910zWqBrsRlBEt:b-brspy1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dba688e18a770d4905b5f955489862a3ae7ef084/cmake/googleapis/CMakeLists.txt#L459)):_
> 
> > ```cmake
> >     FILES "${CMAKE_CURRENT_BINARY_DIR}/googleapis-config.cmake"
> >           "${CMAKE_CURRENT_BINARY_DIR}/googleapis-config-version.cmake"
> >           "${PROJECT_SOURCE_DIR}/cmake/FindgRPC.cmake"
> > ```
> 
> I think we might be installing another copy of this file from elsewhere in the code. ...
> 
> https://github.com/googleapis/google-cloud-cpp/blob/563555b5fdccfe2888c0b3f3eef455aef006266d/google/cloud/CMakeLists.txt#L436-L437
> 
> We probably just want to do this once?

Good catch. I removed the install of these files from google/cloud/, and they're only installed now from `external/googleapis`.

devjgm:
> > So, I absolutely do not know if this will work: why copy these files? Why not have `googleapis` as a directory at the top-level and just use `add_subdirectory(googleapis)`?
> 
> Great idea. I _think_ that will work. The builds are super slow on my laptop, but it appears like it's working, so I made the change and kicked off a presbumit build. We'll see. Anyway, this is a much simpler arrangement.

It works.

PR #4250, opened 2020-05-20
fix: fix off-by-one in uploading streams to GCS

### LINKED ISSUE #3432: Consider fixing next expected byte calculation in storage Client::UploadStreamResumable
It seems next expected byte calculation in `Client::UploadStreamResumable` :
`auto expected = session->next_expected_byte() + gcount - 1;`
should be:
`auto expected = session->next_expected_byte() + gcount;`

That would match with `CurlResumableUploadSession::Update()` next expected byte calculation:
`next_expected_ = result->last_committed_byte + 1;`

Please note that wrong next expected byte calculation doesn't affect what the library uploads because later stream pointer is adjusted (actually to the same location):
```
 if (session->next_expected_byte() != expected) {
      GCP_LOG(WARNING) << "unexpected last committed byte "
                       << " expected=" << expected
                       << " got=" << session->next_expected_byte();
      source.seekg(session->next_expected_byte(), std::ios::beg);
    }
```
So expected that fixing calculation will avoid `source.seekg()` on every file chunk.

dopiera:
This fixes #3432.

Before this change, `seekg` was called for proper behavior for every
`UploadChink`. Because it was `expected`, which was broken, the `seekg`
was effectively a noop, so everything worked, except for generating
warnings.
    
This PR also removes the `seekg` from that execution path altogether.
Playing with tests proved that unless there is a bug in the code, there
is no way to end up in a situation when `seekg` was called and even if
this situation occurred, it would be impossible to properly recover from
it. In order to not increase the likelihood of a corruption, I replaced
this logic with returning an error.
    
In order to make sure that resumable upload sessions are properly
handled, a `seekg` is added at the beginning to handle resuming a
stream.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4250)

### DIFF
--- a/google/cloud/storage/client.cc
+++ b/google/cloud/storage/client.cc
@@ -153,6 +153,7 @@ StatusOr<ObjectMetadata> Client::UploadStreamResumable(
   }
 
   auto session = std::move(*session_status);
+  source.seekg(session->next_expected_byte(), std::ios::beg);
 
   // GCS requires chunks to be a multiple of 256KiB.
   auto chunk_size = internal::UploadChunkRequest::RoundUpToQuantum(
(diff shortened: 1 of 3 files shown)


dopiera:
Failures seem to be caused by #4264

PR #4251, opened 2020-05-21
ci: increase Windows build timeout to 180m

coryan:
The builds are taking (routinely) about 110m, we need to increase
the timeout to make room for new code, such as `cpp-cmakefiles`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4251)


coryan:
Filed #4252 with an idea to improve the time.

Issue #3868, opened 2020-04-20
Cleanup test tags and their use in the CI scripts


coryan:
We use build tags to split the tests across several dimensions:

- Unit tests basically have no tags, they run by default.
- Integration tests of any kind are labeled `integration-tests`, we mostly use that tag to exclude integration tests when running the unit test only.
- Some integration tests need to have the `bigtable` emulator running **or** the environment variables necessary to contact production, these are labeled `bigtable-integration-tests`
- Some integration tests need to have the `storage` emulator (aka tetsbench) running **or** the environment variables necessary to contact production, these are labeled `storage-integration-tests`
- Some integration tests only run against production, these are labeled `integration-tests-no-emulator`

Some integration tests need environment variables that change on each run **and** they typically need a lot of support from the CI scripts to setup the environment to run the test.

Because passing these variables to all the other integration tests would disable caching, we exclude them *by name* in the Bazel runs, and then run them *by name* in the CI scripts **if** we have setup the environment configuration they need.

This is all too complicated, and passing names of tests in the scripts is hard to maintain. we need to clean all of this up.

devjgm:
Hmm, I don't really see a way to do this that works in both bazel and cmake. Basically, we need a way to say something like "run all the integration tests for Bigtable". There are two important pieces of information: "integration tests" and "bigtable". With bazel we can specify both of these like `bazel test --test_tag_fitlers=integration-tests google/cloud/bigtable/...`. That is, we have the build target path and the tag filter to provide those two pieces of information.

With CMake/CTest, I don't believe we have the target path option. We only have labels. And AFAICT, there's no way with ctest to specify run the targets that have *both* labels, i.e., the following doesn't work `ctest -L bigtable -L integration-tests`. The second `-L` overwrites the first. Therefore, I think with ctest we need to combine both of those two pieces of information into a single label, which is exactly what we've done and where we're at now.

We could simplify these labels for bazel, but I'd say trying to keep the labels in sync between cmake and bazel is probably a good idea.

So, I'm thinking there's not much we can do to simplify our use of test tags/labels. Are there other tasks or items being tracked by this issue? If not, we might be able to close this.

coryan:
There is also the crazy `*no-emulator` any thoughts on that?  Maybe we can treat this as the "documentation" of how these work and close it?  Or move it to ci/README.md?  WDYT?

devjgm:
[...]

PR #4257, opened 2020-05-21
cleanup: make the test labels/tags more consistent

devjgm:
Fixes https://github.com/googleapis/google-cloud-cpp/issues/3868

This PR attempts to make our use of labels (cmake) and tags (bazel) a bit simpler and more consistent. This might be a small step in that direction:

* Use the singular term "integration-test" everywhere
* Removed all `$product-integration-test` labels. These are no longer needed. Products are identified in ctest invocations using `-R $product`, and in bazel invocations using `bazel test google/cloud/$product/...`.
* All integration tests in bazel and cmake have an `integration-test` label. This is most helpful when *excluding* all integration tests from a test run.
* [CMAKE ONLY] All integration tests also have a label of either `integration-test-emulator` or `integration-test-production`, to identify whether or not they should run w/ an emulator. (there's one documented exception here about a bigquery integration test that just doesn't fully work today because it needs a script to invoke it w/ some arguments, and that's more than we need to do today).

That last bullet point suggests something odd: The bazel and cmake labels are not in sync. In bazel we have no indication of whether integration tests use an emulator or not. Instead, when we run integration tests in bazel, we manually pick and choose which tests do what (see [build-in-docker-bazel.sh](https://github.com/googleapis/google-cloud-cpp/blob/master/ci/kokoro/docker/build-in-docker-bazel.sh)). It would be nice to clean that up in another PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4257)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -183,7 +183,7 @@ if should_run_integration_tests; then
   # below to avoid invalidating the cached test results for all the other tests.
   "${BAZEL_BIN}" test \
     "${bazel_args[@]}" \
-    "--test_tag_filters=storage-integration-tests,pubsub-integration-tests,spanner-integration-tests" \
+    "--test_tag_filters=integration-test" \
     -- //google/cloud/...:all "${excluded_targets[@]}"
 
   # Changing the PATH disables the Bazel cache, so use an absolute path.
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -173,7 +173,13 @@ if should_run_integration_tests; then
   access_token_targets=(
     "//google/cloud/bigtable/examples:bigtable_grpc_credentials"
   )
-  excluded_targets=()
+  excluded_targets=(
+    # The Bigtable integrations that use the emulator *and* production were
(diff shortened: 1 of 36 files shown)


devjgm:
> I like it, much better than before, so :lgtm:
> 
> Can we summarize the situation in #3868, I think basically it is:
> 
> * Both Bazel and CMake: unit tests have no labels.
> * CMake+Bazel: test programs (or programs that need testing) that are **not** unit tests (integration tests, examples, benchmarks, etc) have the `integration-test` label.
> * CMake: test programs that can run against only against production have the `integration-test-production` label.
> * CMake: test programs that can run against production or the corresponding emulator(s) have the `integration-test-emulator` label.
> * CMake: to run the tests against emulators filter by the `integration-test-emulator` and by the program prefix (e.g. `bigtable_`(
> * Bazel: there are no labels for distinguish what programs can only run in production, instead the CI scripts have hard-coded lists of targets.
> 
> If I got something wrong, just write the right version, no need to convince me that I am wrong 
> 
> > Reviewed 36 of 36 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4257#-:-M7sYxM9B_p9Y0hKF3yN:bs7sl0)_ status: :shipit: complete! all files reviewed, all discussions resolved

Your summary was right. Added to the issue. Thanks.

PR #4262, opened 2020-05-22
cleanup: set CMAKE_CXX_CLANG_TIDY rather than per-target CXX_CLANG_TIDY

### LINKED ISSUE #3958: Enable clang-tidy for all targets

devbww:
Remove the `google_cloud_cpp_add_clang_tidy` cmake function.  Combine
the "header" and "source" cases of the "clang-tidy-only-changed-files"
used in presubmit builds.

Fixes #3958.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4262)


cmake/EnableClangTidy.cmake
```diff
@@ -34,5 +34,8 @@ else ()
     else ()
         message(
             STATUS "clang-tidy found: ${GOOGLE_CLOUD_CPP_CLANG_TIDY_PROGRAM}")
+        if (GOOGLE_CLOUD_CPP_CLANG_TIDY)
```
devjgm, line 37: I'm thinking/hoping that we can get rid of this whole `EnableClangTidy.cmake` file. Instead of using this file and setting the `CMAKE_CXX_CLANG_TIDY` option programatically, I think we can remove this file and then in `build-in-docker-cmake.sh:85` we could change

```sh
# OLD
cmake_extra_flags+=("-DGOOGLE_CLOUD_CPP_CLANG_TIDY=ON")
```
to
```sh
# NEW 
cmake_extra_flags+=("-DCMAKE_CXX_CLANG_TIDY=clang-tidy")
```

I think that would be simpler because we wouldn't need to detect and act on the CMAKE_VERSION to figure out if it works with clang-tidy. Instead, that's simply up to the user who installed a particular version of cmake and, if they have cmake >= 3.6, they can enable the clang-tidy checks using the standard instructions [described here](https://blog.kitware.com/static-checks-with-cmake-cdash-iwyu-clang-tidy-lwyu-cpplint-and-cppcheck/). Additionally, there's no need for us to provide the `GOOGLE_CLOUD_CPP_CLANG_TIDY` option, which is just another option for people to know about.

We could certainly consider/do this in a future PR, so it need not block you.

Thoughts?

devbww: I'm certainly happy to do whatever is "right", but I can tell you that the above quickly dies on clang-tidy errors in (at least) generated code.  For example, ...
```
/v/cmake-out/ci-fedora-install-31-clang-tidy/external/googleapis/google/type/expr.pb.cc:28:3: error: invalid case style for variable '_Expr_default_instance_' [readability-identifier-naming,-warnings-as-errors]
} _Expr_default_instance_;
  ^~~~~~~~~~~~~~~~~~~~~~~
  expr_default_instance
```

devjgm: I see. Well, I'm happy to move this to a separate issue for us to consider (or drop) so as to not block your PR.

PR #4268, opened 2020-05-26
fix: cleanup how shared libs get a VERSION

### LINKED ISSUE #4265: Invalid VERSION value for pubsub library
The `VERSION` property value for the pubsub libraries seems invalid:
```
$ ls build/google/cloud/pubsub/libpubsub_client.so*
google/cloud/pubsub/libpubsub_client.so -> libpubsub_client.so.1
google/cloud/pubsub/libpubsub_client.so.
google/cloud/pubsub/libpubsub_client.so.1 -> libpubsub_client.so.
```
The `GOOGLE_CLOUD_CPP_VERSION` variable used to fill that property is not set in the context it is used:
https://github.com/googleapis/google-cloud-cpp/blob/d915c7f57fa407cf58e6c78ad5401657345da990/google/cloud/pubsub/CMakeLists.txt#L61-L63

Other libraries use the separate version variables `${GOOGLE_CLOUD_CPP_VERSION_MAJOR}.${GOOGLE_CLOUD_CPP_VERSION_MINOR}.${GOOGLE_CLOUD_CPP_VERSION_PATCH}`. I believe bigquery and spanner may have the same issue.

This is probably another case for #4132

**Context**
* Linux Debian testing/bullseye
* google-cloud-cpp d915c7f57fa407cf58e6c78ad5401657345da990
* CMake 3.17.1

### LINKED ISSUE #4266: Broken CMake config file for pubsub
The CMake config file generated and installed along the Pub/Sub library is broken resulting in a CMake error when trying to `find_package` it.

I believe the issue is here:
https://github.com/googleapis/google-cloud-cpp/blob/d915c7f57fa407cf58e6c78ad5401657345da990/google/cloud/pubsub/config-version.cmake.in#L19
as the `GOOGLE_CLOUD_CPP_PUBSUB_CLIENT_VERSION` variable is not defined in the CMakeLists.txt generating the config file.

**To Reproduce**
1. Build and intall pubsub client library
2. Attempt a `find_package(pubsub_client REQUIRED)` from a consumer CMake project
3. Error:
```
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/pubsub_client/pubsub_client-config-version.cmake:19 (if):
  if given arguments:

    "EQUAL" "0"

  Unknown arguments specified
```

**Context**
* Linux Debian testing/bullseye
* google-cloud-cpp d915c7f57fa407cf58e6c78ad5401657345da990
* CMake 3.17.1

coryan:
Always use GOOGLE_CLOUD_CPP_VERSION which is now set in the top-level
CMake file.

Also fixed the pubsub/config-version.cmake.in file which was referencing
GOOGLE_CLOUD_CPP_**PUBSUB**_VERSION, a variable that no longer exists.

Fixes #4265 and fixes #4266

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4268)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,9 @@ project(google-cloud-cpp CXX C)
 set(GOOGLE_CLOUD_CPP_VERSION_MAJOR 1)
 set(GOOGLE_CLOUD_CPP_VERSION_MINOR 14)
 set(GOOGLE_CLOUD_CPP_VERSION_PATCH 0)
+string(CONCAT GOOGLE_CLOUD_CPP_VERSION "${GOOGLE_CLOUD_CPP_VERSION_MAJOR}"
+              ".${GOOGLE_CLOUD_CPP_VERSION_MINOR}"
+              ".${GOOGLE_CLOUD_CPP_VERSION_PATCH}")
 
 # Configure the Compiler options, we use C++11 features by default.
 set(GOOGLE_CLOUD_CPP_CXX_STANDARD
--- a/external/googleapis/CMakeLists.txt
+++ b/external/googleapis/CMakeLists.txt
@@ -16,11 +16,6 @@
 
 cmake_minimum_required(VERSION 3.5)
 
-string(
-    CONCAT GOOGLE_APIS_CPP_PROTOS_VERSION "${GOOGLE_CLOUD_CPP_VERSION_MAJOR}"
-           "." "${GOOGLE_CLOUD_CPP_VERSION_MINOR}" "."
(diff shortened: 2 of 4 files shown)


coryan:
> https://github.com/googleapis/google-cloud-cpp/blob/7c52b89b44357c1fb9beaaa16e1991b8b1fbfe17/external/googleapis/CMakeLists.txt#L19-L22
> 
> is a similar looking construct. Should this one remain separate or should it use `GOOGLE_CLOUD_CPP_VERSION` too?

I am removing that construct in this PR, right?  I think that answers the question 😁

PR #4271, opened 2020-05-26
cleanup: fix struct/class mismatch for bigtable::Mutation

coryan:
It seems the warning is only generated in Bazel builds, which do not
have `-Werror` set.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4271)


coryan:
Duplicated, closing.

Issue #4273, opened 2020-05-26
Change storage/benchmarks to also use GrpcClient


coryan:
We want to compare the performance of the GrpcClient against the existing REST-based client.

I think we only need to change the `storage_throughput_vs_cpu_benchmark` program.

PR #4278, opened 2020-05-27
feat: DefaultGrpcClient(Options) cannot fail

coryan:
This overload cannot fail, its return type should indicate that.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4278)


coryan:

PR #4282, opened 2020-05-27
fix: super builds when compiling static libraries

coryan:
By default gRPC compiles a C# extension shared library (`.so`) even if
all the dependencies (like Abseil) are compiled statically. In some
platforms that just breaks (mixing PIC code and non-PIC code
results in an error). Since we do not use this extension I think the
simplest thing is to disable it for super builds.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4282)


coryan:

PR #4283, opened 2020-05-27
cleanup: remove ProgressReporter in benchmarks

coryan:
I thought it would be useful to report throughput stats after N bytes,
but turns out we did not need that data. Removing the code because I
would rather not carry unused code around.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4283)


coryan:

PR #4284, opened 2020-05-27
feat: support X-Upload-Content-Length header

### LINKED ISSUE #3162: X-Upload-Content-Length not Well Supported
Issue #857 states all standard parameters in google/cloud/storage are supported in the client libraries.  This doesn't appear to be the case for X-Upload-Content-Length.  We are setting it as a custom header but I would expect it to be in the [well_supported_headers](https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/storage/well_known_headers.h) code with all the other x- headers.

dopiera:
This fixes #3162.

Unfortunately, there does not seem to be a way to set the counterpart of
this header for gRPC version, so only cURL is supported.

Emulator is updated to support this too.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4284)

### DIFF
--- a/google/cloud/storage/client.cc
+++ b/google/cloud/storage/client.cc
@@ -117,7 +117,7 @@ StatusOr<ObjectMetadata> Client::UploadFileSimple(
 
 StatusOr<ObjectMetadata> Client::UploadFileResumable(
     std::string const& file_name,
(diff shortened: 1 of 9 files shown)


coryan:
Hm... Did you look into the clang-tidy error?

PR #4285, opened 2020-05-28
feat: allow docker to use host network

devjgm:
This fixes some users who were running the integration tests on GCE
instances.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4285)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -197,12 +197,12 @@ if should_run_integration_tests; then
   source "${PROJECT_ROOT}/ci/kokoro/gcloud-functions.sh"
 
   trap delete_gcloud_config EXIT
+  activate_service_account_keyfile "${KOKORO_SETUP_KEY}"
   create_gcloud_config
 
   echo
   echo "================================================================"
   io::log "Delete any stale service account used in HMAC key tests."
-  activate_service_account_keyfile "${KOKORO_SETUP_KEY}"
   cleanup_stale_hmac_service_accounts
 
   echo
--- a/ci/kokoro/docker/build.sh
+++ b/ci/kokoro/docker/build.sh
@@ -506,6 +506,13 @@ docker_flags=(
   "--rm"
 )
 
+# Some machines have complex host network configurations that we want to shared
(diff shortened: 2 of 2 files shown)


ci/kokoro/docker/build.sh
```diff
@@ -506,6 +506,13 @@ docker_flags=(
   "--rm"
 )
 
+# Some machines have complex host network configurations that we want to shared
+# with the running docker container. Setting this environment variable tells
+# the docker container to use the host network rather than a bridge network.
+if [[ -n "${DOCKER_USE_HOST_NETWORK:-}" ]]; then
+  docker_flags+=(--net host)
```
devbww, line 513: It doesn't seem to do it for me.  That is, the diagnosis test case I had set up still fails.

If I wait for a while (presumably to clear the throttling cache), the first error occurs at the same point.  There is a difference after that however, as it seems to timeout earlier, which leads to additional (faster) failures.

ci/kokoro/docker/build-in-docker-bazel.sh
```diff
@@ -197,12 +197,12 @@ if should_run_integration_tests; then
   source "${PROJECT_ROOT}/ci/kokoro/gcloud-functions.sh"
 
   trap delete_gcloud_config EXIT
+  activate_service_account_keyfile "${KOKORO_SETUP_KEY}"
```
devbww, line 200: I don't know what this is related to, but I see the same error as in some of your presubmits.

devjgm: This "fix" was wrong. I think I fixed it correctly in my new PR: https://github.com/googleapis/google-cloud-cpp/pull/4288

ci/kokoro/docker/build.sh
devjgm: Hmm. I don't understand why you'd still be seeing DNS flakiness since using the host network should mean the docker container is now using your host's DNS cache. Anyway, I dropped this PR in favor of https://github.com/googleapis/google-cloud-cpp/pull/4288, which seems to address the issues I was seeing (which I thought were the same as the issues you were seeing). Maybe try that new PR and then let me know exactly what errors you're seeing.

PR #4288, opened 2020-05-28
feat: optionally use host network in docker containers

devjgm:
When enabled with the `DOCKER_USE_HOST_NETWORK` environment variable, our docker containers will share the host's networking. This can be useful when running on a host with a non-trivial networking configuration, such as DNS caches, etc, that you want to share with the docker container. 

This change also exposed an issue where some of our gcloud commands used
incidental gcloud accounts that existed on the system, other than the
service acount that we intended. This is fixed by passing `--account=""`
to some gcloud commands where we know we don't want to use any existing
system account.

I believe this should fix the issue that @devbww and I were having when running the integration tests on our GCE instances.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4288)

### DIFF
--- a/ci/kokoro/docker/build.sh
+++ b/ci/kokoro/docker/build.sh
@@ -373,6 +373,13 @@ mkdir -p "${BUILD_HOME}"
 
 # We use an array for the flags so they are easier to document.
 docker_flags=(
+  # Our docker containers don't need network-level isolation from the host,
+  # because we can run our builds directly on the host anyway. Additionally,
+  # some hosts have non-trivial networking configurations that the container
+  # should also use (DNS caches, etc). Therefore, our containers will directly
+  # use the host network.
+  "--net" "host"
+
   # Enable ptrace as it is needed by s
   "--cap-add" "SYS_PTRACE"
 
--- a/ci/kokoro/docker/build.sh
+++ b/ci/kokoro/docker/build.sh
@@ -373,13 +373,6 @@ mkdir -p "${BUILD_HOME}"
 
 # We use an array for the flags so they are easier to document.
 docker_flags=(
-  # Our docker containers don't need network-level isolation from the host,
-  # because we can run our builds directly on the host anyway. Additionally,
-  # some hosts have non-trivial networking configurations that the container
-  # should also use (DNS caches, etc). Therefore, our containers will directly
-  # use the host network.
-  "--net" "host"
-
   # Enable ptrace as it is needed by s
   "--cap-add" "SYS_PTRACE"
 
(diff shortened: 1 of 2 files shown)


ci/kokoro/docker/build.sh
```diff
@@ -514,6 +514,13 @@ if [[ -n "${KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH:-}" ]]; then
   docker_flags+=("--env" "KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH=${KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH:-}")
 fi
 
+# Optionally allow the docker container to use the host's networking in order
+# to take advantage of potentially non-trivial network configurations (e.g.,
+# DNS cache) that might be a PITA to duplicate within the container itself.
+if [[ -n "${DOCKER_USE_HOST_NETWORK:-}" ]]; then
+  docker_flags+=("--net" "host")
```
devbww, line 521: I would use the non-abbreviated flag name here, so `s/--net/--network/`.

ci/kokoro/docker/build.sh
```diff
@@ -514,6 +514,13 @@ if [[ -n "${KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH:-}" ]]; then
   docker_flags+=("--env" "KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH=${KOKORO_GITHUB_PULL_REQUEST_TARGET_BRANCH:-}")
 fi
 
+# Optionally allow the docker container to use the host's networking in order
+# to take advantage of potentially non-trivial network configurations (e.g.,
+# DNS cache) that might be a PITA to duplicate within the container itself.
```
devbww, line 519: If this doesn't deserve to be the default, we should mention why (e.g., the [current] use of a fixed emulator port).

Aside: There is also the possibility of making the choice automatically if we can predict when the container will use a rate-limited DNS server.  But in the end it might be nicer if we can indeed figure out how to run a caching domain server in the container, or how to forward just the DNS traffic to the host caching server.

Aside2: We should probably also keep in mind that customers not using a DNS cache may see the same issue.  And even if they do have a cache, they may see rate limiting when actually connecting to the target service (as we saw when running many tests quickly).

devjgm: > If this doesn't deserve to be the default, we should mention why (e.g., the [current] use of a fixed emulator port).
> 
> Aside: There is also the possibility of making the choice automatically if we can predict when the container will use a rate-limited DNS server.

Not sure how to do that generically.

> But in the end it might be nicer if we can indeed figure out how to run a caching domain server in the container,

That'd work.

> or how to forward just the DNS traffic to the host caching server.

Done. That was my first "hacky" solution.

```shell
sudo socat UDP-LISTEN:53,bind=192.168.9.1,reuseaddr,fork UDP:127.0.0.1:53
```

Which binds my `docker0`'s IP (192.168.9.1) and forwards everything to localhost:53. Well, actually, this only does UDP; we should forward TCP also to handle DNS more robustly. But any forwarding solution like this needs to be set up before we start the docker container, which can get a little complicated (e.g., the above solution requires that the local machine has `socat` installed).

> 
> Aside2: We should probably also keep in mind that customers not using a DNS cache may see the same issue. And even if they do have a cache, they may see rate limiting when actually connecting to the target service (as we saw when running many tests quickly).

These sound mostly like application-level issues.

Issue #4228, opened 2020-05-18
Audit spanner tests and RUN_SLOW_INTEGRATION_TESTS


coryan:
In `-spanner` we used to `RUN_SLOW_INTEGRATION_TESTS` environment variable to turn on specific tests. I think the nightly builds set that to `backup,instance`, while the coverage build only set it to `backup`, and the other builds did not set it at all.

I believe we lost this when we moved to `-cpp`, and most (all?) the builds do not set this at all.

We need to audit how the variable is used in the code, and make sure the CI scripts are covering these cases. Note that `instance` turns on tests that have very limited quota and are *very* slow, so it cannot be just turned on everywhere (OTOH, we have an emulator now, so maybe we can turn it on for more builds).

devjgm:
NOTE: We already use a `RUN_SLOW_INTEGRATION_TESTS` variable, but only in our macos/build.sh script. It's set to either "yes" or "no".

https://github.com/googleapis/google-cloud-cpp/blob/07cd8ee3118cd3ec549c4b45f1484088e0393287/ci/kokoro/macos/build.sh#L27-L31 

Our `google/cloud/spanner/...` code does check an environment variable named `RUN_SLOW_INTEGRATION_TESTS`, but it expects a value like "instance", or "backup", or "instance,backup". Not yes/no.

So I think nothing is using our current `RUN_SLOW_INTEGRATION_TESTS` varaible.

coryan:
Thanks for looking into this, I know it is hard to untangle all the scripts sometimes.

> So I think nothing is using our current `RUN_SLOW_INTEGRATION_TESTS` varaible.

You probably know this: that is an oopsie. Effectively there are tests / examples that we wrote and are not running, that should make us sad.  We probably want to see how the variable was used in `-spanner` and copy that usage to `-cpp`.

nit: s/var**ai**ble/variable/

devjgm:
Yep, my plan is to get these running in `-cpp`. I'm using this issue as my place for notes as I untangle a little bit of stuff.

devjgm:
How did this work in `-spanner`? Back in the `-spanner` repo, `RUN_SLOW_INTEGRATION_TESTS` was only enabled (set to "instance,backup") in a `coverage` build.

https://github.com/devjgm/google-cloud-cpp-spanner/blob/e9f73518069f69f79ab8e74e69ea68babcc20a58/ci/kokoro/docker/build.sh#L86-L88

Other than that, these slow builds were not run.

PR #4295, opened 2020-05-28
fix: enable spanner slow integration tests 

devjgm:
These "slow" integration tests are only enabled for certain builds at
this point. For now, only the "integration" build (both CI and
presubmit), and builds that use the Spanner emulator.

In the previous repo, google-cloud-cpp-spanner, these tests were
controlled by the variable `RUN_SLOW_INTEGRATION_TESTS`, but in this
monorepo, that variable was too generic, so it was renamed to be
Spanner specific.

If these builds aren't too slow or cause too much contention, we can
enable them in more builds. Possibly even include them in all
builds that run integration tests. We'll see.

Fixes: #4228

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4295)


devjgm:
The timing for the newly enabled tests in this from the "integration" build is:

```
//google/cloud/spanner/integration_tests:spanner_client_backup_integration_test PASSED in 742.8s
...
//google/cloud/spanner/integration_tests:spanner_client_instance_admin_integration_test PASSED in 25.7s
```

The overall time for the build was 20 min, 24 sec: https://source.cloud.google.com/results/invocations/6fa52828-e9d1-49e5-8e2d-6a4071e3616e/targets

coryan:
I just realized: spanner now runs against the emulator, are these tests/examples disabled when running against the emulator?  I recall that backups are, maybe?

devjgm:
> I just realized: spanner now runs against the emulator, are these tests/examples disabled when running against the emulator? I recall that backups are, maybe?

I don't think Spanner uses the emulator in the "integration" build that uses bazel. I filed an issue to track that (https://github.com/googleapis/google-cloud-cpp/issues/4297).

There are the Spanner+emulator+cmake tests (https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/spanner/ci/run_integration_tests_emulator_cmake.sh), and in this PR I added "instance,backup".... though I'm not totally sure which of our builds hits this code path.

devjgm:
> Ah, so the code is running in the `integration` build, but not in the `coverage` build. I missed that. LGTM then. We are slightly blind as to coverage, but otherwise Okay.
> 
> > Reviewed 7 of 7 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4295#-:-M8Rw8fO78g0-AUqat1_:bw16ysc)_ status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)
> 
> _[google/cloud/spanner/integration_tests/instance_admin_integration_test.cc, line 58 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4295#-M8RvpTV0IgK8bnQQKcX:-M8RvpTV0IgK8bnQQKcY:b-ilfco4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/178d428816309e3e532f03683f2e24e4bbd92e92/google/cloud/spanner/integration_tests/instance_admin_integration_test.cc#L58)):_
> 
> > ```c++
> >             .value_or("");
> >     auto const run_slow_integration_tests =
> >         google::cloud::internal::GetEnv("ENABLE_SPANNER_SLOW_INTEGRATION_TESTS")
> > ```
> 
> nit: we have both `GOOGLE_CLOUD_CPP_SPANNER_TEST_*` and `ENABLE_SPANNER_*` variables... not sure if we want that.

True. I added it next to [`ENABLE_BIGTABLE_ADMIN_INTEGRATION_TESTS`](https://github.com/googleapis/google-cloud-cpp/blob/e01df8e3eb7e7d3c61c71dd7c59c1b713f8ff3e1/ci/kokoro/docker/build.sh#L444) and copied that naming style. But I think you're right that the GOOGLE_CLOUD_CPP prefix is more common and probably what we should following going forward.

Fixed.

PR #4307, opened 2020-05-29
feat(storage/benchmarks): quantize app buffers

### LINKED ISSUE #4303: Benchmark using only chunks that are multiples of upload quantum
In `storage_throughput_vs_cpu_benchmark` we upload data in "chunks", that is, we break the object in small blocks of equal size, and call `.write()` for each block.

The library uploads data also in chunks but these are different from the chunks described above. 
 A resumable upload must break the data in chunks, and each chunk size must be a multiple of 256KiB (which we call the quantum).

- [ ] I think we should rename the benchmark "chunks" to "blocks" or "segments" or something, to avoid confusion.
- [ ] We should also have an option to ensure the blocks sizes are always a multiple of 256KiB, as this is the recommended behavior for applications.

coryan:
The benchmark uses randomly sized application-level buffers to upload
and download objects. There are some sizes that are known to be much
better. For example, for uploads, anything that is a multiple of 256KiB.
Sometimes it is interesting to test with "bad" buffer sizes, but most
often it is better to use the right values.

There is also no reason to upload and download with the same buffer
sizes, this change uses different buffer sizes for each.

Furthermore, the benchmark called application-level buffers "chunks",
this was a poorly chosen name. There are other "chunks" in a resumable
upload, and this made things confusing.

Finally, changed the default values for these application level buffers
to something that is more reasonable.

Fixes #4303

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4307)

### DIFF
--- a/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
+++ b/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
@@ -54,8 +54,9 @@ this loop (see below for the conditions to stop the loop):
 
 - Select a random size, between two values configured in the command line of the
   object to upload.
-- Select a random chunk size, between two values configured in the command line,
-  the data is uploaded in chunks of this size.
+- The application buffer sizes for `read()` and `write()` calls are also
+  selected at random. These sizes are quantized, and the quantum can be
+  configured in the command-line.
 - Select, at random, the protocol / API used to perform the test, could be XML,
   JSON, or gRPC.
 - Select, at random, if the client library will perform CRC32C and/or MD5 hashes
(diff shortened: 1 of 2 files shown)


google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -506,12 +547,37 @@ google::cloud::StatusOr<Options> ParseArgsDefault(
...
-       << options.maximum_chunk_size << "]";
+    os << "Invalid range for write size [" << options.minimum_write_size << ','
+       << options.maximum_write_size << "]";
     return make_status(os);
   }
+  if (options.write_quantum <= 0 ||
+      options.write_quantum > options.minimum_write_size) {
+    std::ostringstream os;
+    os << "Invalid value for --write-quantum (" << options.write_quantum
+       << "), it should be in the [0," << options.minimum_write_size
```
devbww, line 561: s/0/1/?

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -506,12 +547,37 @@ google::cloud::StatusOr<Options> ParseArgsDefault(
...
+    std::ostringstream os;
+    os << "Invalid range for read size [" << options.minimum_read_size << ','
+       << options.maximum_read_size << "]";
+    return make_status(os);
+  }
+  if (options.read_quantum <= 0 ||
+      options.read_quantum > options.minimum_read_size) {
+    std::ostringstream os;
+    os << "Invalid value for --read-quantum (" << options.read_quantum
+       << "), it should be in the [0," << options.minimum_read_size
```
devbww, line 576: Ditto.

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
coryan: Yup, thanks!

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
coryan: fixed too.

Issue #4306, opened 2020-05-29
Spanner backup integration test timeout


devjgm:
https://source.cloud.google.com/results/invocations/a6f0c51a-80ee-4690-886a-71476ad4b87f/targets

We should probably fix this by running the "integration" tests (which use bazel) w/ the Spanner emulator if we can.

devjgm:
Oh, wait this suggests that Spanner backups are not supported by the emulator.

https://github.com/googleapis/google-cloud-cpp/blob/080825e2387de8688fd0a7e5c5970f04d2d118c0/google/cloud/spanner/integration_tests/backup_integration_test.cc#L92-L94

Need a new plan.

devjgm:
I'm reverting the PR that "fixed" this. Revert PR: https://github.com/googleapis/google-cloud-cpp/pull/4316

Increasing the bazel timeout did not work because the policy still had a 15 minute timeout:

devjgm:
Before re-enabling the Spanner backup tests, we should:

1. Understand why/where all the time is going.
2. See if we can adjust the timeouts used in the policy objects to reduce the total time or at least remove the flakiness
3. Re-enable the tests only once (we think) the flakiness (i.e., timeouts) are gone or very rare.

devjgm:
I added some timing and logging to the https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/spanner/integration_tests/backup_integration_test.cc to see which functions are taking the longest.

**Raw output:**
```
[ RUN      ] BackupTestWithCleanup.BackupTestSuite
2020-05-29T22:09:10.997560838+00:00 ### RUNNING BACKUP TESTS
2020-05-29T22:09:11.075771304+00:00 ### CreateInstance()...
2020-05-29T22:09:12.511216361+00:00 ### DONE
2020-05-29T22:09:16.768812067+00:00 ### CreateBackup()...
2020-05-29T22:09:21.514117866+00:00 ### got future
2020-05-29T22:09:22.835259606+00:00 ### cancelling...
2020-05-29T22:09:23.42027205+00:00 ### DONE
2020-05-29T22:09:23.420340957+00:00 ### CreateBackup() (again)...
2020-05-29T22:09:25.980628454+00:00 ### ListBackupOperationrs()...
2020-05-29T22:09:26.150394747+00:00 ### DONE
2020-05-29T22:09:26.150484425+00:00 ### Getting backup_future.get()...
[...]

PR #4313, opened 2020-05-29
fix: re-enable spanner backup tests w/ increased timeout

devjgm:
Fixes: #4306

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4313)


devjgm:
If the "integration" build gets flaky again, we can rollback this PR.

PR #4324, opened 2020-06-02
feat(source<T>): a function to capture all values

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4324)


google/cloud/testing_util/mock_sources.h
```diff
@@ -0,0 +1,79 @@
...
+namespace testing_util {
+
+/**
+ * A type meeting the requirements for source<T, E> used in testing.
+ *
+ * @tparam T the value type for the source
+ * @tparam E the error type for the source
+ */
+template <class T, typename E>
+class MockSource {
```
devjgm, line 36: Naming nit: The word "mock" here may make readers think that this object is doing some verification (it's not, is it?). Especially with our use of Googletest, readers may expect it to respond to ON_CALL() and friends.

I think "Fake" may be a more commonly used term for this type of class. WDYT?

google/cloud/internal/source_accumulators.h
```diff
@@ -0,0 +1,120 @@
...
+namespace google {
+namespace cloud {
+inline namespace GOOGLE_CLOUD_CPP_NS {
+namespace internal {
+
+template <typename Source>
+struct AccumulateAllEventsType {
+  using type = absl::variant<std::vector<typename Source::value_type>,
+                             typename Source::error_type>;
+};
```
devjgm, line 33: I'm wondering if you actually need both the struct and the alias. Could both of these be replaced with something like the following?

```
template <typename Source>
using AccumulateAllEventsType =
    absl::variant<std::vector<typename Source::value_type>,
                              typename Source::error_type>;
```

google/cloud/internal/source_accumulators.h
```diff
@@ -0,0 +1,120 @@
...
+ * `E` are the value type and error type in the original source. The returned
+ * future is satisfied with an error if the original source is closed with an
+ * error.
+ *
+ * @tparam Source
+ * @param source
+ * @return
+ */
+template <typename Source>
+future<accumulate_all_events_t<absl::decay_t<Source>>> accumulate_all(
```
devjgm, line 54: `accumulate_all()` doesn't match the function naming convention of our style guide. Is it intentionally different to match the style of something else?

Related question for `start()`, `on_next()` below.

google/cloud/testing_util/mock_sources.h
coryan: Fixed.

google/cloud/internal/source_accumulators.h
coryan: Fixed.

google/cloud/internal/source_accumulators.h
coryan: Fixed

> CamelCase delenda est

google/cloud/testing_util/mock_sources.h
```diff
@@ -0,0 +1,79 @@
...
+namespace testing_util {
+
+/**
+ * A type meeting the requirements for source<T, E> used in testing.
+ *
+ * @tparam T the value type for the source
+ * @tparam E the error type for the source
+ */
+template <class T, typename E>
+class FakeSource {
```
devjgm, line 36: Note: this file is still named `mock_sources.h`.

coryan: Renamed. I expect we will have more mocks (for real mocks) eventually.

PR #4326, opened 2020-06-02
feat(storage/benchmarks): control CRC32C/MD5 options

coryan:
Sometimes we want to run a benchmark with fixed values for CRC32C and/or
MD5 hashing. This adds command line options to run the benchmarks with
either these options fixed to some value from the command-line, or to
chose a random value on each iteration (the default, as before).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4326)

### DIFF
--- a/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
+++ b/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
@@ -465,9 +465,11 @@ google::cloud::StatusOr<Options> ParseArgsDefault(
   auto parse_checksums = [](std::string const& val) -> std::vector<bool> {
     if (val == "enabled") {
       return {true};
-    } else if (val == "disabled") {
+    }
+    if (val == "disabled") {
       return {false};
-    } else if (val == "random") {
+    }
+    if (val == "random") {
       return {false, true};
     }
     return {};
--- a/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
+++ b/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
@@ -110,8 +110,8 @@ struct Options {
       gcs_bm::ApiName::kApiGrpc,
 #endif  // GOOGLE_CLOUD_CPP_STORAGE_HAVE_GRPC
   };
-  std::vector<bool> enable_crc32c = {false, true};
-  std::vector<bool> enable_md5 = {false, true};
+  std::vector<bool> enabled_crc32c = {false, true};
+  std::vector<bool> enabled_md5 = {false, true};
 };
 
 enum OpType { kOpWrite, kOpInsert, kOpRead0, kOpRead1, kOpRead2 };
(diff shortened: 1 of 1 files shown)


google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -162,6 +165,28 @@ int main(int argc, char* argv[]) {
   std::transform(notes.begin(), notes.end(), notes.begin(),
                  [](char c) { return c == '\n' ? ';' : c; });
 
+  auto join_apis = [](std::vector<ApiName> const& apis) {
```
devjgm, line 168: JFYI, you could use `absl::StrJoin()` here ([link](https://github.com/abseil/abseil-cpp/blob/master/absl/strings/str_join.h)). Since you already have a `gcs_bm::ToString` function and you're formatting bools not as 0 and 1, you'd want a custom formatter. That would look something like this:

```
struct MyFormatter {
  void operator()(std::string* out, ApiName api) const {
    out->append(gcs_bm::ToString(api));
  }
  void operator()(std::string* out, bool b) const {
    out->append(b ? "true" : "false");
  }
};
...

std::cout << absl::StrJoin(options->enabled_apis, ',', MyFormatter{})
          << absl::StrJoin(options->enabled_crc32, ',', MyFormatter{})
          << absl::StrJoin(options->enabled_md5, ',', MyFormatter{})
```

Your call whether or not that's an improvement... JFYI.

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -109,6 +110,8 @@ struct Options {
       gcs_bm::ApiName::kApiGrpc,
 #endif  // GOOGLE_CLOUD_CPP_STORAGE_HAVE_GRPC
   };
+  std::vector<bool> enable_crc32c = {false, true};
+  std::vector<bool> enable_md5 = {false, true};
```
devjgm, line 114: The existing vector field is named `enabled_apis` with the `d`. These don't have the `d` in the name. Is that difference intentional, or do you want these to match? Note: the flag for `enable_crc32c` is named `enabled-crc32c`... could be confusing.

mr-salty:
___
*[google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc, line 558 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4326#-M8qx3qK3mTMGm9gIaS9:-M8qx3qK3mTMGm9gIaSA:b8b5mgi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1ebde8523eaa92704ad27290cb400e8ce97fb64c/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc#L558)):*
> ```cpp
>          options.enable_crc32c.clear();
>          std::set<bool> crc32c;
>          for (auto& token : absl::StrSplit(val, ',')) {
> ```

IIUC the argument is a list of bool, and the number of times "true" appears controls the probability that it will be enabled?

If we want to be able to control the probabilities, I'd have the flag take a number directly.
If we don't, limit the list size to 1 or 2 values, or maybe use "enabled, disabled, random" (so you don't have to spell the last one like "false,true")

mr-salty:
running out for a bit but LG on my comment, I'll let greg LG the whole thing

PR #4333, opened 2020-06-03
ci: add basic github actions for PR and CI builds

devjgm:
This PR is basically a "hello world" using GitHub Actions in our repo. This PR adds two workflows, the PR workflow (defined in pr.yml) runs for all pull requests targeting the master branch. The CI workflow (defined in ci.yml) runs on all pushes/merges to the master branch.

These workflows essentially just compile the code and run the unit tests (not integration tests) using bazel. The `~/.cache/bazel` directory is cached between runs, so the builds should be relatively efficient.

The goal here is to give us some experience with GitHub actions.

You can see the PR results for this build by looking at the checks that were run for this build. You can also see the status of all github actions (including the ones for this build) by clicking on https://github.com/googleapis/google-cloud-cpp/actions.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4333)


devjgm:
> I love the build times, the simplicity of this initial `.yml` files, and keeping **all** the CI configuration under one SCM. I love that it manages caches as a real CI system should. At least the documentation sounds like it would support matrix builds better than Kokoro, so yay!
> 
> I am not sure how it would display multiple builds for the same PR / branch / commit? I am hopeful that it does the right thing.
> 
> I do not love keeping a separate file for PR and CI builds, can we merge them? (not necessarily in this PR)
> 
> Do we have a long term plan? Do we want to eventually migrate all CI to GitHub Actions? There is some complexity in keeping two, so either:
> 
> * we plan move all the builds (which requires figuring out the "secrets problem"), or
> * we plan to keep both indefinitely, in which case what is the benefit and does it pay off?
> 
> These long term discussions can wait if you wan to commit this to gain experience, but we should be willing to rollback fast if there are problems. The stuff below is more or less optional. Let me know what you thinkg.

As discussed offline, the goal w/ this PR (and likely some tweaks afterward), is to learn how GH Actions work so we can make an informed decision about whether they add value for our use cases or not. I'm willing to roll back *anything* that doesn't provide net value, including this (rolling back is easy, adding them is easy, tweaking them is easy, so this is a fairly easy experiment).

> 
> > Reviewed 2 of 2 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4333#-:-M8wRgSqC0leep5JLBEB:b-wsgve1)_ status: all files reviewed, 5 unresolved discussions (waiting on @devjgm)
> 
> _[.github/workflows/ci.yml, line 1 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4333#-M8wU5Sp0Yshu-oytLWh:-M8wU5Sp0Yshu-oytLWi:b-dgnflp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8ded822f73c8e4441c35a19b526330ca9c7bd756/.github/workflows/ci.yml#L1)):_
> 
> > ```yaml
> > # For an overview of the syntax of this file, see
> > ```
> 
> I think we are supposed to use the Copyright boilerplate, even for these files.

You're right. Added.

> 
> _[.github/workflows/ci.yml, line 11 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4333#-M8wUvRoEjTopFO3d543:-M8wUvRoEjTopFO3d544:b-3q0kih) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8ded822f73c8e4441c35a19b526330ca9c7bd756/.github/workflows/ci.yml#L11)):_
> 
> > ```yaml
> >   push:
> >     branches:
> >       - master
> > ```
> 
> We would probably want to run these builds for all `vX.Y.Z` branches too?

OK. I tweaked the rules to have no branch restrictions. The build(s) will run on any/all of our branches (maybe I should exclude gh-pages since that's a totally different world?).

> 
> _[.github/workflows/pr.yml, line 8 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4333#-M8wUK2G9kTDQ5-9G8fC:-M8wUK2G9kTDQ5-9G8fD:b-xtivdt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8ded822f73c8e4441c35a19b526330ca9c7bd756/.github/workflows/pr.yml#L8)):_
> 
> > ```yaml
> > # This workflow triggers only on Pull Requests to the master branch
> > on:
> >   pull_request:
> > ```
> 
> This looks identical other than the `on: ` section, can we merge them? If there are small tweaks we can use the `if` syntax:
> 
> https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions
> 
> Merging can wait for a future PR.

They files weren't identical. They differed in their caching steps. On PR builds we also look up (restore-keys) keys for the `github.base_ref` branch, which in our case is essentially always master. I tried to merge them into a single file but couldn't get the `if:` stuff working correctly. I now merged them into a single file w/ a single Cache step where I now hardcode the word "master" instead of using `github.base_ref`, which is not defined on "push" events.

FWIW, having two workflow files, one for PRs and one for CI, didn't bother me particularly. It might ultimately be simpler than cramming control-flow conditions in a yaml file. But we'll see. This is all part of the experiment! :-)

> 
> _[.github/workflows/pr.yml, line 10 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4333#-M8wV3EeE2C4BC7A-cYb:-M8wV3EeE2C4BC7A-cYc:b7qbhnh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8ded822f73c8e4441c35a19b526330ca9c7bd756/.github/workflows/pr.yml#L10)):_
> 
> > ```yaml
> >   pull_request:
> >     branches:
> >       - master
> > ```
> 
> Same comment about branches, can also wait.

Done, I think?

> 
> _[.github/workflows/pr.yml, line 14 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4333#-M8wV5xcEBufIGDODhny:-M8wV5xcEBufIGDODhnz:bokdywy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8ded822f73c8e4441c35a19b526330ca9c7bd756/.github/workflows/pr.yml#L14)):_
> 
> > ```yaml
> > jobs:
> >   build:
> >     runs-on: ubuntu-latest
> > ```
> 
> "latest" can change unexpectedly on us and break our builds, consider pinning to a specific version.

Changed to `ubuntu-18.04`

PR #4334, opened 2020-06-03
feat(source<T>): flow control building blocks

coryan:
For asynchronous streaming RPCs we need to limit the number of Read()
calls to just one. In other contexts having a class to limit the number
of outstanding operations is useful too. This change introduces some
building blocks to implement flow control. They are not intended for
general use, they will always be part of the implementation of a
user-facing class.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4334)


google/cloud/internal/source_ready_token.h
```diff
@@ -0,0 +1,126 @@
...
+      : max_outstanding_(max_outstanding) {}
+
+  ReadyTokenFlowControl(ReadyTokenFlowControl&& rhs) noexcept
+      : current_outstanding_(rhs.current_outstanding_),
+        max_outstanding_(rhs.max_outstanding_),
+        pending_(std::move(rhs.pending_)) {}
+  ReadyTokenFlowControl& operator=(ReadyTokenFlowControl&& rhs) noexcept {
+    ReadyTokenFlowControl tmp(std::move(rhs));
+    using std::swap;
+    swap(current_outstanding_, tmp.current_outstanding_);
```
devjgm, line 94: Do you intentionally want to not move `tmp.mu_`?

coryan: I wish I could move it, but:

"std::mutex is neither copyable nor movable."

https://en.cppreference.com/w/cpp/thread/mutex

PR #4345, opened 2020-06-05
refactor(storage/benchmarks): options parsing

coryan:
Move functions to parse the throughput_vs_cpu_benchmark options to a
common file in the benchmarks library. If nothing else it makes testing
easier, but I may create separate benchmarks soonish.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4345)

### DIFF
--- a/google/cloud/storage/benchmarks/throughput_options.cc
+++ b/google/cloud/storage/benchmarks/throughput_options.cc
@@ -1,0 +1,240 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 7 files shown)


google/cloud/storage/benchmarks/throughput_options.cc
```diff
@@ -0,0 +1,240 @@
...
+    if (val == "disabled") {
+      return {false};
+    }
+    if (val == "random") {
+      return {false, true};
+    }
+    return {};
+  };
+
+  std::vector<OptionDescriptor> desc{
```
devjgm, line 41: (no change necessary... just an FYI) Now that we have a dep on Abseil, we may want to consider using Abseil's flag parsing library, otherwise, we end up re-inventing a lot of standard flag parsing stuff. I know we already have a number of places where we parse flags ourselves, so I see no problem with adding more of it here in this PR (again, no change necessary). But going forward, we may want to consider Abseil flags.

https://abseil.io/docs/cpp/guides/flags

Issue #4045, opened 2019-08-05
[Flake]: Bazel builds failed while downloading http archive.


coryan:
Two (otherwise unrelated) builds failed while downloading the BoringSSL archive:

https://sponge.corp.google.com/target?id=936c747b-c24b-4938-bb80-108e46618629&target=cloud-cpp%2Fspanner%2Fdocker%2Fubsan-presubmit

https://source.cloud.google.com/results/invocations/2340826a-9f4e-4277-a244-0bbacbbb84aa/targets/cloud-cpp%2Fspanner%2Fmacos%2Fbazel/log

We should consider caching the most important tarballs in GCS so the download has an alternative location. I believe the TensorFlow folks do this:

https://github.com/tensorflow/tensorflow/blob/76aa7f81ab726eb9f862c6c691ef2c76ea4e6af3/WORKSPACE#L10

coryan:
These are still happening, but maybe just on indirect dependencies, the latest example is:

https://source.cloud.google.com/results/invocations/82c4e4a9-4cde-4f1c-b637-28d8f81feb42/targets

coryan:
https://source.cloud.google.com/results/invocations/d77fd19f-24ea-46c0-a3c1-da34656f63be

devjgm:
That last one looks like general networking issues to me. bazel failed to fetch everything. I filed https://github.com/bazelbuild/bazel/issues/11520 for the heck of it.

devjgm:
The error in https://github.com/googleapis/google-cloud-cpp/issues/4045#issuecomment-625820538 is from a Windows build. There are a bunch of errors from bazel, but I suspect they all stem from the same issue. I suspect the problem was

```
2020-04-15T13:17:05.7816072+00:00  extracting build cache failed with exit code 2.
```
So perhaps we left the build cache directory in some weird state, that bazel then tried to use, and the cached files were corrupt somehow.

I suspect the fix for this is to clear out the build cache directory if extracting the cache fails. I'll see if I can figure out how to do "rm -rf <dir>" in powershell :-)

coryan:
> The error in [#4045 (comment)](https://github.com/googleapis/google-cloud-cpp/issues/4045#issuecomment-625820538) is from a Windows build. There are a bunch of errors from bazel, but I suspect they all stem from the same issue. I suspect the problem was
> 
> ```
> 2020-04-15T13:17:05.7816072+00:00  extracting build cache failed with exit code 2.
> ```

[...]

PR #4355, opened 2020-06-05
fix: don't preserve symlinks in 7z archive

devjgm:
This is an attempt to fix the flakes in #4045

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4355)


Rone90122 (commented):
> ``

PR #4362, opened 2020-06-08
feat(storage/benchmarks): compare to raw downloads

coryan:
Implement experiments to download using raw gRPC and raw libcurl calls,
to make it easier to compare the results with other teams and measure
the "overhead" of the library. As usual, this ignores that the client
library adds functionality, such as retries.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4362)

### DIFF
--- a/google/cloud/storage/benchmarks/throughput_experiment.cc
+++ b/google/cloud/storage/benchmarks/throughput_experiment.cc
@@ -18,9 +18,7 @@
 #include "google/cloud/storage/grpc_plugin.h"
 #include "google/cloud/grpc_error_delegate.h"
 #include "absl/memory/memory.h"
-#if GOOGLE_CLOUD_CPP_STORAGE_HAVE_GRPC
 #include <google/storage/v1/storage.grpc.pb.h>
-#endif  // GOOGLE_CLOUD_CPP_STORAGE_HAVE_GRPC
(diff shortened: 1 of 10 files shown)


google/cloud/storage/benchmarks/throughput_experiment.cc
```diff
@@ -165,27 +295,29 @@ std::vector<std::unique_ptr<ThroughputExperiment>> CreateUploadExperiments(
...
+      case ApiName::kApiRawGrpc: {
+        gcs::Client grpc_client =
+            google::cloud::storage_experimental::DefaultGrpcClient(
+                client_options);
         result.push_back(
             absl::make_unique<UploadObject>(grpc_client, a, contents, false));
         result.push_back(
             absl::make_unique<UploadObject>(grpc_client, a, contents, true));
-        break;
+      } break;
```
devjgm, line 311: Nit: This `break` looks strange sitting on the same line as a closing brace. I'd suggest moving it inside the brace'd block, or at least on its own line following the braced block.

coryan: meh. Fixed. FWIW, clang-format made me do it :man_shrugging:

Issue #4070, opened 2020-05-09
[Flake] Resource exhausted in spanner integration tests


coryan:
We just got two failures, caused by the same issue:

```
Creating database and table google/cloud/spanner/testing/database_integration_test.cc:81: Failure
Value of: database
Expected: is OK
Actual: Error in non-idempotent operation CreateDatabase: The instance projects/cloud-cpp-testing-resources/instances/test-instance-5 has too many database splits to complete the operation.  Please add more nodes and try again. [RESOURCE_EXHAUSTED] [RESOURCE_EXHAUSTED]
```

https://source.cloud.google.com/results/invocations/90b7291b-013c-4abe-8e15-98785f399637
https://source.cloud.google.com/results/invocations/bde9c3a9-59a8-45a0-a478-0054bfd43ab9

coryan:
Used this command to cleanup some databases:

```
for db in $(gcloud spanner databases list --project cloud-cpp-testing-resources --instance test-instance-5 --filter='name:db-' --format='csv(name)[no-heading]' --limit 5); do sleep 1; gcloud spanner databases delete --project cloud-cpp-testing-resources --instance test-instance-5 --quiet "${db}"; echo ${db}; done
```

coryan:
This is (I think) related to #4033

devjgm:
These errors say `Please add more nodes and try again`. It's a resource exhausted error. Our test instances only have a single node. I think we could probably safely make them all 2-node test instances. Perhaps we should just do that?

coryan:
We may need to move some instances to a different region because we are close to exhausting our node quota for `us-central1` (as I recall).

If that is the path we take we would need to:

- Add more instances in a new region with 2 nodes each.
- Remove some of the existing instances (ideally with no builds running at the time, to prevent more flakes).
- Add nodes to the instances not removed in the previous step.

I suspect that will probably just buy time, eventually we will saturate two node instances too. That is a guess though.

coryan:
Doing another cleanup (I found about 180 stale databases):

```console
for instance in $(gcloud spanner instances list --project cloud-cpp-testing-resources --filter="name:test-instance" --format='csv(name)[no-heading]'); do
  for db in $(gcloud spanner databases list --project cloud-cpp-testing-resources --instance ${instance} --filter='name:db-' --format='csv(name)[no-heading]' | grep -v 'db-2020-'); do
     echo $instance $db; sleep 1; 
[...]

PR #4364, opened 2020-06-08
ci(spanner): auto cleanup old databases

### LINKED ISSUE #4033: Automatically clean up the old leaked databases
We automatically delete old leaked instances at:
https://github.com/googleapis/google-cloud-cpp-spanner/blob/6e82ef3f1f2c87c5894f851d344e04bf2bfc1fd0/google/cloud/spanner/integration_tests/instance_admin_integration_test.cc#L63

We can do similar thing for leaked databases too.

coryan:
Our CI builds create randomly named databases (for isolation). Normally
these are cleaned up by the tests that create them, from time to time
these may crash or get interrupted and the cleanup does not happen. From
now on the instances are prefixed by the date (in `db-YYYY-MM-DD`
format). Any database with an ID matching that format and older than 48
hours is automatically removed.

This is part of the work for #4070 and fixes #4033 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4364)

### DIFF
--- a/google/cloud/spanner/testing/random_database_name.cc
+++ b/google/cloud/spanner/testing/random_database_name.cc
@@ -14,6 +14,7 @@
 
 #include "google/cloud/spanner/testing/random_database_name.h"
 #include "google/cloud/internal/format_time_point.h"
+#include <ctime>
 
 namespace google {
 namespace cloud {
--- a/google/cloud/spanner/testing/random_database_name.cc
+++ b/google/cloud/spanner/testing/random_database_name.cc
@@ -22,7 +22,7 @@ namespace spanner_testing {
 inline namespace SPANNER_CLIENT_NS {
 
 std::string RandomDatabasePrefixRegex() {
-  return R"re("^db-\d{4}-\d{2}-\d{2})re";
+  return R"re("^db-\d{4}-\d{2}-\d{2}-)re";
 }
 
 std::string RandomDatabasePrefix(std::chrono::system_clock::time_point tp) {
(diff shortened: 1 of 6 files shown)


google/cloud/spanner/testing/random_database_name.cc
```diff
@@ -13,17 +13,31 @@
...
+#include "google/cloud/internal/format_time_point.h"
+#include <ctime>
 
 namespace google {
 namespace cloud {
 namespace spanner_testing {
 inline namespace SPANNER_CLIENT_NS {
+
+std::string RandomDatabasePrefixRegex() {
+  return R"re("^db-\d{4}-\d{2}-\d{2})re";
```
devbww, line 25: Perhaps this should have the dash on the end too (to match `RandomDatabasePrefix()`).

coryan: Thanks, done.

google/cloud/spanner/testing/random_database_name.cc
```diff
@@ -13,17 +13,31 @@
...
 namespace google {
 namespace cloud {
 namespace spanner_testing {
 inline namespace SPANNER_CLIENT_NS {
+
+std::string RandomDatabasePrefixRegex() {
+  return R"re("^db-\d{4}-\d{2}-\d{2}-)re";
+}
+
+std::string RandomDatabasePrefix(std::chrono::system_clock::time_point tp) {
```
devjgm, line 28: FWIW, if we wanted to use absl/time/time.h, this would look like:
```cc
std::string RandomDatabasePrefix(std::chrono::system_clock::time_point tp) {
  return absl::FormatTime("db-%Y-%m-%d-", absl::FromChrono(tp), absl::UTCTimeZone());
}
```
IMO, as long as we're not talking about public interface types, I think Abseil Time adds a lot of value and simplicity (but I realize I'm biased).

PR #4374, opened 2020-06-09
feat(source<T>): add source builder

### LINKED ISSUE #4341: Implement SourceBuilder<Source>
In my prototype I found that using TransformedSource<Source, Transform> directly was awkward, a helper class to apply transformations turns out to be very useful. In outline this class would look like so:

```C++
template <typename Source>
class SourceBuilder { public:
  SourceBuilder(Source source); // holds `source` and applies transformations.

  // Return the current source, note that it "cannot" be used afterwards.
  Source Build() &&;

  future<E> Background() &&; // run current source in background, discard values

  template <typename Functor>
  SourceBuilder<...> Transform(Functor&& fun) && ; // apply `fun` to current source.

  template <typename Predicate>
  SourceBuilder<...> Filter(Predicate&& p) &&; // apply predicate p to current source.

  template <typename Composition>
  SourceBuilder<...> Compose() &&; // apply a complex composition (e.g. trailing averages)

  template <typename Accumulator>
  future<...> Accumulate(...) &&; // run an accumulator in the background
};
```

coryan:
Any class meeting the `source<T, E>` interface can be modified with
transforms, accumulators, filters (soon), and arbitrary composition
(soon). The `SourceBuilder<Source>` template class can perform these
transformations with a good user-facing API.

Fixes #4341

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4374)


google/cloud/internal/source_builder.h
```diff
@@ -0,0 +1,109 @@
...
+
+ private:
+  Source source_;
+};
+
+/**
+ * Factory function to create a `SourceBuilder` for @p s
+ */
+template <typename Source>
+auto make_source_builder(Source&& s) -> SourceBuilder<absl::decay_t<Source>> {
```
devjgm, line 100: The naming convention in here seems to be mixing C++ and Google styles. Most of the "make" factories in our codebase look like "MakeFoo"; we almost never create a "make_foo" in our code. Maybe this should be `MakeSourceBuilder`, and similarly `Accumulate` and `Transform` above... unless there's a technical reason/need for the differences in this code.

coryan: PTAL.

PR #4377, opened 2020-06-10
feat: use `SELECT 1` to refresh sessions

### LINKED ISSUE #4022: Use `select 1` instead of GetSession for keepalive
It turns out that keeping sessions alive using a `SELECT 1` is likely more efficient than calling the `GetSession` RPC. So we should likely change our code to do this.

https://github.com/googleapis/google-cloud-cpp-spanner/blob/62bb823c7eb659e6e2305bb42550bccd9913fe17/google/cloud/spanner/internal/session_pool.cc#L160

Internal bug tracking number 152803456

Tagging @mr-salty as likely the best person to take a look at this. Does this seem like a big task? Easy? Major? What do you think?

mr-salty:
Fixes #4022

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4377)


google/cloud/spanner/internal/session_pool.cc
```diff
@@ -157,8 +157,8 @@ void SessionPool::RefreshExpiringSessions() {
     }
   }
   for (auto& refresh : sessions_to_refresh) {
-    AsyncGetSession(cq_, refresh.first, std::move(refresh.second))
-        .then([](future<StatusOr<spanner_proto::Session>> result) {
+    AsyncRefreshSession(cq_, refresh.first, std::move(refresh.second))
```
devbww, line 160: Above we still say "Initiate an async GetSession()".

PR #4397, opened 2020-06-15
fix(storage): handle 408 errors as retryable

### LINKED ISSUE #4396: Treat 408 responses from GCS resumable uploads as transient and retryable
The storage client currently has no special handling for a 408 response from a resumable GCS upload, so it is treated like a generic 4XX signaling a bad request and thus is not retried.

The [GCS documentation](http://cloud/storage/docs/resumable-uploads#practices), however, includes the 408 status code in the list of codes that should be retried.

Searching around, I found some evidence ([1](http://cloud/storage/docs/json_api/v1/status-codes#408_Request_Timeout), [2](https://cloud.google.com/storage/docs/xml-api/reference-status#408%E2%80%94request-timeout)) that a 408 signals the upload connection was broken.

timford:
as per https://cloud.google.com/storage/docs/resumable-uploads#practices.

Fixes #4396.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4397)

### DIFF
--- a/google/cloud/storage/internal/http_response.cc
+++ b/google/cloud/storage/internal/http_response.cc
@@ -76,6 +76,11 @@ Status AsStatus(HttpResponse const& http_response) {
   if (http_response.status_code == HttpStatusCode::kMethodNotAllowed) {
     return Status(StatusCode::kPermissionDenied, http_response.payload);
   }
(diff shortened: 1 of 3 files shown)


coryan:
Thanks @timford !

PR #4401, opened 2020-06-15
fix(spanner): examples on how to delete data

### LINKED ISSUE #4336: Spanner doc missing two examples for delete rows.
Internal reference: b/153943106

## Background

The online documentation for Spanner's `DeleteData` ([the code](https://github.com/googleapis/google-cloud-cpp/blob/f07d03d6f4bc2917608bb987ded48324147b81ad/google/cloud/spanner/samples/samples.cc#L1041)) is here: https://cloud.google.com/spanner/docs/modify-mutation-api#c++_2

The paragraph above mentions 4 ways of deleting rows:
1. Delete all the rows in a table.
2. Delete a single row by specifying the key column values for the row.
3. Delete a group of rows by creating a key range.
4. Delete rows in an interleaved table by deleting the parent rows, if the interleaved table includes ON DELETE CASCADE in its schema definition.

Our sample currently only demonstrates (2) and (3). We actually _do_ cover case (4) also, but since that is completely handled by our schema using "ON DELETE CASCADE", we should add a comment in the code so readers know that it's happening (see the Node.js example that does just this).

## Desired Solution

(Note, these notes are taken from a plan that all the client-library languages will follow, so we should do the same to be consistent.)

The tables that the sample operates are Singers and Albums where Albums is interleaved in Singers and defined with ON DELETE CASCADE. The available keys are:

Singers | 1, 2, 3, 4, 5
-- | --
Albums | (1, 1), (1, 2), (2, 1), (2, 2), (2, 3)

The sample needs to delete all of these rows.

*Sample Logic*

The sample should be updated to do the following:
1. Delete (2, 1) and (2, 3) from Albums at the same time using KeySet
2. Delete 3, 4 from Singers using a KeyRange with closed start of 3 and open end of 5
3. Delete all remaining rows from Singers using KeySet all functionality. This should not specify key 1 or key 2 at all.

The final delete (specified in 3) should have a comment stating that it will also delete the remaining Albums rows because it was defined with ON DELETE CASCADE.

coryan:
Align the `spanner_delete_data` example with other languages. Use 3
different mutations (by key, by range, delete all), and explain how the
last one takes advantage of `ON DELETE CASCADE`. We use the same values
as C#, Java, Node.js, and Python), previously we used the same values as
Go. We use mutations in these examples, there does not seem to be any
consistency across languages on whether they should use SQL statements
or mutations.

Fixes #4336

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4401)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -1035,35 +1035,41 @@ void UpdateData(google::cloud::spanner::Client client) {
 }
 //! [END spanner_update_data]
 
-//! [START spanner_delete_data] [make-key] [make-keybound-closed]
-//! [keyset-add-key]
+//! [START spanner_delete_data]
 void DeleteData(google::cloud::spanner::Client client) {
   namespace spanner = ::google::cloud::spanner;
 
-  // Delete each of the albums by individual key, then delete all the singers
-  // using a key range.
+  // Delete the albums with key (2,1) and (2,3).
+  //! [make-key] [keyset-add-key]
   auto delete_albums = spanner::DeleteMutationBuilder(
                            "Albums", spanner::KeySet()
-                                         .AddKey(spanner::MakeKey(1, 1))
-                                         .AddKey(spanner::MakeKey(1, 2))
                                          .AddKey(spanner::MakeKey(2, 1))
-                                         .AddKey(spanner::MakeKey(2, 2))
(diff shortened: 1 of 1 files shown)


mr-salty:
should be `fix(spanner)`, right?

coryan:
> should be `fix(spanner)`, right?

Yup, fixed.

PR #4009, opened 2020-05-05
ci: avoid flakiness in HMAC integration tests

coryan:
Fixes #3790

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4009)

Issue #3790, opened 2020-04-15
[Flake] HMAC integration tests and examples may be fighting each other


coryan:
We just had a build fail with:

```
[ RUN      ] ServiceAccountIntegrationTest.HmacKeyCRUD
google/cloud/storage/tests/service_account_integration_test.cc:119: Failure
Value of: get_details
Expected: is OK
Actual: Permanent error in GetHmacKey: {
  "error": {
    "code": 404,
    "message": "Access ID not found in project.",
    "errors": [
      {
        "message": "Access ID not found in project.",
        "domain": "global",
        "reason": "notFound"
      }
    ]
  }
}
 [NOT_FOUND] [NOT_FOUND]
[  FAILED  ] ServiceAccountIntegrationTest.HmacKeyCRUD (2539 ms)
```

https://source.cloud.google.com/results/invocations/27093f92-92d4-4ada-9549-9933ca7b0b8e

I think the integration tests and the examples are not working well when executed in parallel, maybe we need two separate HMAC keys (ugh), or change the code to be more friendly to parallelism.

coryan:
Another one I think:

https://source.cloud.google.com/results/invocations/40631501-484f-40fb-8158-e7d01bda59bf/targets

coryan:
This seems like a repeat:

https://source.cloud.google.com/results/invocations/337c764e-d7c4-401f-94df-09ddea90a197

PR #4403, opened 2020-06-16
fix: eliminate data race between storage "service account" tests

devbww:
`storage_service_account_samples` should only delete the HMAC keys
that it creates, rather than everything for the service account,
otherwise it can interfere with other tests running in parallel,
notably `service_account_integration_test`.

We could also try to delete old keys, but it isn't worth it since
these tests are only run against the testbench emulator.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4403)

### DIFF
--- a/google/cloud/storage/examples/storage_service_account_samples.cc
+++ b/google/cloud/storage/examples/storage_service_account_samples.cc
@@ -101,34 +101,30 @@ void ListHmacKeysWithServiceAccount(google::cloud::storage::Client client,
 
 std::string CreateHmacKey(google::cloud::storage::Client client,
                           std::vector<std::string> const& argv) {
-  std::string created_id;
   //! [create hmac key] [START storage_create_hmac_key]
   namespace gcs = google::cloud::storage;
   using ::google::cloud::StatusOr;
-  [](gcs::Client client, std::string const& service_account_email,
-     std::string& access_id) {
+  return [](gcs::Client client, std::string const& service_account_email) {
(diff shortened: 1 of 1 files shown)


coryan:
> FYI, this is fixing a "flake", even though we didn't have an issue created for this flake.

Hmmm... #3790 ?

> This flake is something that @devbww and I periodically saw when running our tests with lots of parallelism.

The fixes in #4009 avoided the problems by *not* running these tests in parallel.  Were you *manually* running the tests with more parallelism than the CI scripts do?

> Thanks operator!

Indeed. We should consider running all the HMAC tests in parallel again.

devbww:
> > FYI, this is fixing a "flake", even though we didn't have an issue created for this flake.
> 
> Hmmm... #3790 ?

Yes, that looks like the same underlying issue.  Thanks for pointing it out.

> > This flake is something that @devbww and I periodically saw when running our tests with lots of parallelism.
> 
> The fixes in #4009 avoided the problems by not running these tests in parallel. Were you manually running the tests with more parallelism than the CI scripts do?

No we weren't manually running the tests, but the execution originated with `ci/kokoro/docker/build-in-docker-cmake.sh` so #4009 did not apply.  Until we started using 96-core machines, I'm guessing that the CMake build never produced parallelism between the offending tests.

> We should consider running all the HMAC tests in parallel again.

I will send a followup PR reversing the effect of #4009.

PR #4405, opened 2020-06-16
doc(spanner): add CreateInstance() example

### LINKED ISSUE #4347: Spanner doc: Add sample showing CreateInstance
Internal reference: b/148418510

The internally referenced issue has a link to a spec doc and more details. Basically no client library in any language has this sample. So we want to coordinate so that all libraries implement a similar sample.

As an example, here's the PR adding this sample for C#: https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/1072

coryan:
Fixes #4347

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4405)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -63,15 +63,49 @@ void GetInstanceCommand(std::vector<std::string> const& argv) {
   GetInstance(std::move(client), argv[0], argv[1]);
 }
 
-//! [create-instance]
+//! [START spanner_create_instance] [create-instance]
 void CreateInstance(google::cloud::spanner::InstanceAdminClient client,
                     std::string const& project_id,
                     std::string const& instance_id,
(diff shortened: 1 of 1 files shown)


coryan:
___
*[google/cloud/spanner/samples/samples.cc, line 129 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4405#-M9xlxITDEkqD7hGf4x-:-M9yAskp34E13rL3pB1p:b95srt7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0e5278593ae013cdda1db7025e49cc0bc8807a23/google/cloud/spanner/samples/samples.cc#L129)):*

Consider calling `CreateInstance()` instead of duplicating the code?
</blockquote></details>

I would need to extract the region from the instance_config, and filter for only regional location configs in the first place, seemed like more work than just calling client.CreateInstance().

PR #4408, opened 2020-06-16
feat: Add GenericRequest::ValueOr<>

dopiera:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4408)

### DIFF
--- a/google/cloud/storage/CMakeLists.txt
+++ b/google/cloud/storage/CMakeLists.txt
@@ -376,6 +376,7 @@ if (BUILD_TESTING)
         internal/binary_data_as_debug_string_test.cc
         internal/bucket_acl_requests_test.cc
         internal/bucket_requests_test.cc
+        internal/complex_option_test.cc
         internal/compute_engine_util_test.cc
         internal/curl_client_test.cc
         internal/curl_handle_factory_test.cc
--- a/google/cloud/storage/client.cc
+++ b/google/cloud/storage/client.cc
@@ -59,9 +59,7 @@ ObjectReadStream Client::ReadObjectImpl(
   auto stream =
       ObjectReadStream(absl::make_unique<internal::ObjectReadStreambuf>(
           request, *std::move(source),
-          request.HasOption<ReadFromOffset>()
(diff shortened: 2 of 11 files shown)


coryan:
___
*[google/cloud/storage/client.cc, line 62 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4408#-M9zKz9EBt1iJLqApNoU:-MA1Rzay3PNJt5rc1hqC:b94j0g7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/72c1c4d31902adc15cc90b77b44d6ac0cb0d1dc7/google/cloud/storage/client.cc#L62)):*

I don't think we can do it. Correct me if I'm wrong, but I think it will still be:
`ValueOr(SomeDefaultConstructedComplexOption).value_or("some actual value")`
</blockquote></details>

I was thinking this:

https://github.com/googleapis/google-cloud-cpp/compare/master...coryan:test-get-option-value-or

coryan:
___
*[google/cloud/storage/client.cc, line 62 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4408#-M9zKz9EBt1iJLqApNoU:-MA1WtLbEPtbeAKP7uzU:b-mfnr3c) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/72c1c4d31902adc15cc90b77b44d6ac0cb0d1dc7/google/cloud/storage/client.cc#L62)):*

Crap - I misread the code - I thought the options are held by a `std::optional<Option>`, so calling `GetOption` would fail. It doesn't, so yes - it makes perfect sense, let me do it.

One quick question - your code does not allow for not setting the `user_project` proto field (it's always set, possibly to an empty string). Is that intentional?
</blockquote></details>

That was more for testing than anything... In proto3 "not set == set to default value" for primitive types (string, ints, enums, etc.), so this becomes a question of (a) what is more readable, (b) what is faster.  I think the code with `value_or()` is marginally more readable, maybe.

PR #4409, opened 2020-06-17
cleanup: use absl::{conjunction,disjunction}

devjgm:
We no longer need our own implementations of conjunection and
disjunction meta functions now that we can use Abseil.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4409)


devjgm:
Bummer. This breaks due to an odd gcc 4.8 compiler bug. I've repro'd the compiler bug, and a possible way that Abseil could work around the bug (if it doesn't break anything else).

https://godbolt.org/z/i7-AmM

devjgm:
I'll re-open this after the next Abseil release, which will have a fix for `absl::disjunction`

PR #4411, opened 2020-06-19
fix: empty version in cmake package

EdvardD:
GOOGLE_CLOUD_CPP_CLIENT_VERSION is not defined so it's replaced by an empty string and version config is invalid

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4411)


EdvardD:
@googlebot I signed it!

Issue #4413, opened 2020-06-19
Combine several sanitizers into a single build


devjgm:
See https://github.com/google/sanitizers/issues/1039#issuecomment-453607726 where kcc says

> asan+lsan+ubsan is fine, the rest should run separately.

We currently don't have an `lsan` build (leak sanitizer). But we probably should.

Today we have separate asan and ubsan builds. But it looks like we could combine those into a single build (maybe we'll call it `xsan`?). And even add lsan into the mix, and we'll end up reducing our CI builds by one.

devjgm:
Perhaps I'm wrong about us not having an 'lsan' build. It looks like it's built into the asan detector https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer

devjgm:
It looks like our ubsan build takes ~90 minutes. Our asan build takes ~60 min. And this xsan build takes ~120 minutes. xsan is probably slightly shorter overall, but not when we factor in the parallelism of running two separate builds (asan + ubsan = double the cores) at the same time.

So at this point, I don't think it's worth combining these builds, so I'm going to close this issue.

PR #4414, opened 2020-06-19
feat: add xsan build

devjgm:
This build runs asan + ubsan, which are all safe to
run together. Once this build is enabled, we can delete the separate
asan and ubsan builds.

Related to https://github.com/googleapis/google-cloud-cpp/issues/4413

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4414)


coryan:
FYI, you are missing `xsan-presubmit.cfg`, the file must exist even if empty.

PR #4418, opened 2020-06-22
feat(spanner/numeric): add google::cloud::spanner::Numeric

### LINKED ISSUE #4356: Implement Spanner NUMERIC type
This is a tracking issue. Details in some internal docs.

devbww:
Add `google::cloud::spanner::Numeric`, a representation of the
Spanner NUMERIC type.  NUMERIC provides an exact numeric value
with 29 decimal digits of integer precision and 9 decimal digits
of fractional precision.

A `Numeric` value can be constructed from, and converted to a
`std::string`, a `double`, or any integral type.  Integers can
be scaled by powers of 10 to avoid precision loss in the C++
values.

DO NOT MERGE.  This PR is based on a development branch of
googleapis/googleapis, so it cannot be merged until the NUMERIC
`TypeCode` is added to the default branch and the development-
branch hooks are removed.

Note: Only adds NUMERIC up to `google::cloud::spanner::Value`.
Server-side support is needed before we can use a NUMERIC in
a database schema.

Fixes #4356.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4418)


google/cloud/spanner/numeric.h
```diff
@@ -0,0 +1,269 @@
...
+ * assert(spanner::ToInteger<int>(n).value() == 1234568);
+ * assert(spanner::ToInteger<int>(n, 2).value() == 123456789);
+ * @endcode
+ */
+///@{
+template <typename T,
+          typename std::enable_if<std::numeric_limits<T>::is_integer &&
+                                      !std::numeric_limits<T>::is_signed,
+                                  int>::type = 0>
+StatusOr<T> ToInteger(Numeric const& n, int exponent = 0) {
```
devjgm, line 187: This is a somewhat large function template at 30 lines, and I'm thinking about possible binary bloat from it being stamped out into the binary for each `T` that this template is instantiated with. If that seems like a legit concern, we could have the majority of the body of this function in a single non-templated `StatusOr<absl::int128> ToIntegerImpl(...)` that returns the value as an `absl::int128`, then the templated wrapper only needs to look at the returned int128 and see if it fits in the requested `T`. Of course, this solution assumes that `int128` is big enough to hold any `T` that the caller provides. I think that'd be a fine assumption.

We'd also need to do the same thing for `absl::uint128`, but it'd follow the same pattern for the same reasons.

WDYT? Do we care about potential binary bloat here?

google/cloud/spanner/numeric.h
```diff
@@ -0,0 +1,269 @@
...
+StatusOr<Numeric> MakeNumeric(std::string s, int exponent);
+
+}  // namespace internal
+
+/**
+ * Construction from a string, in decimal fixed- or floating-point formats
+ * (where there must be digits either before or after the decimal point):
+ *
+ *    [-+]?[0-9]+(.[0-9]*)?([eE][-+]?[0-9]+)?
+ *    [-+]?.[0-9]+([eE][-+]?[0-9]+)?
```
devjgm, line 132: It may be helpful for human readers to see a few examples of of numeric strings so they don't have to interpret this more formal description.

google/cloud/spanner/numeric.h
```diff
@@ -0,0 +1,269 @@
...
+ *
+ * Fails on syntax errors or if the conversion would yield a value outside
+ * the NUMERIC range. If the argument has more than `kFracPrec` digits after
+ * the decimal point it will be rounded, with halfway cases rounding away
+ * from zero.
+ */
+StatusOr<Numeric> MakeNumeric(std::string s);
```
devjgm, line 139: Is a `MakeNumeric` followed by a `ToString` guaranteed to produce the same input string? Or is it canonicalized? I didn't see docs that mentioned this, but possibly I overlooked them.

google/cloud/spanner/numeric.h
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Conversion to the nearest integer value, scaled by 10^`exponent`, rounding
+ * halfway cases away from zero. Fails when the destination type cannot hold
+ * that value.
+ *
+ * @par Example
+ *
+ * @code
+ * spanner::Numeric n = spanner::MakeNumeric(123456789, -2).value();
```
devjgm, line 176: I like this example. It nicely shows a general overview of how to use this class, including it's non-member functions (still part of the class's interface).

It makes me think that this example (or an _additional_ example) may be nice up in the class's documentation so that readers of the class docs get a quick overview of how to use the class, including the non-member functions. i'd vote for adding an additional example up there.

google/cloud/spanner/numeric.h
```diff
@@ -0,0 +1,269 @@
...
+  std::string const& rep = n.ToString();
+  if (exponent != 0) {
+    auto const en = internal::MakeNumeric(rep, exponent);
+    return en ? ToInteger<T>(*en, 0) : en.status();
+  }
+  T v = 0;
+  constexpr auto kDigits = "0123456789";
+  constexpr auto kMin = (std::numeric_limits<T>::min)();
+  bool negate = true;
+  enum { kIntPart, kFracPart } state = kIntPart;
```
devjgm, line 233: C++ parsing is weird :-) Is this guaranteed to work or happens to work?

google/cloud/spanner/numeric_benchmark.cc
```diff
@@ -0,0 +1,127 @@
...
+#include <limits>
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace {
+
+// Run on (96 X 2000.17 MHz CPU s)
```
devjgm, line 27: > Run on (96 X 2000.17 MHz CPU s)

Show off ;-)

google/cloud/spanner/numeric.h
```diff
@@ -0,0 +1,269 @@
...
+  friend StatusOr<Numeric> internal::MakeNumeric(std::string s);
+  explicit Numeric(std::string rep) : rep_(std::move(rep)) {}
+  std::string rep_;  // a valid and canonical NUMERIC representation
+};
+
+namespace internal {
+
+// Like `std::to_string`, but also supports Abseil 128-bit integers.
+template <typename T>
+std::string ToString(T value) {
```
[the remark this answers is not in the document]
devbww: Done.

PR #4424, opened 2020-06-25
chore: remove comment on the license statement

mr-salty:
this is now disallowed when importing code to google.
given there's a full license statement a few lines above, this is
not adding much value so we might as well remove it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4424)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -14,7 +14,7 @@
 
 package(default_visibility = ["//visibility:public"])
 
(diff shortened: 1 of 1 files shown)


mr-salty:
just realized this is not limited to this file, but the rest are fixed on import, so closing this PR

PR #4431, opened 2020-06-25
fix: incorrect metadata key string

scotthart:
per https://google.aip.dev/client-libraries/4222 and https://github.com/googleapis/googleapis/blob/0a602be7b3835b51d59daf8f6f5dc2dc22f69d7e/google/spanner/admin/database/v1/spanner_database_admin.proto#L125

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4431)


coryan:
At least in theory we could (and should) have a test that compares the metadata against the annotations, for example:

https://github.com/googleapis/google-cloud-cpp/blob/faafc869a93c62f35cfacb0e5f63b8dc6f7c64ef/google/cloud/spanner/internal/database_admin_metadata_test.cc#L50-L70

I think this PR was missing such a test, and we should also look at what other functions are missing it.

PR #4448, opened 2020-06-30
fix: CompletionQueue::RunAsync is always async

### LINKED ISSUE #4083: Potential deadlocks in bigtable::MutationBatcher

Yes.

Bigtable

**Describe the bug**

This code is suspect:

https://github.com/googleapis/google-cloud-cpp/blob/a68f42a33dca00c35818ccbfbbe3bec96bc7dd0e/google/cloud/bigtable/mutation_batcher.cc#L185-L189

It calls `TryAdmin()` which performs a lot of work, potentially making RPCs:

https://github.com/googleapis/google-cloud-cpp/blob/a68f42a33dca00c35818ccbfbbe3bec96bc7dd0e/google/cloud/bigtable/mutation_batcher.cc#L192-L207

This can deadlock and/or make no progress.

**To Reproduce**

This is a report from a customer, in their application they need to run about a thousand processes for the deadlock to happen.

**Expected behavior**

The `MutationBatcher` should guarantee progress.

**Operating system:**

Linux, no further details available.

coryan:
Prior to this change we sometimes ran the callable passed to
`CompletionQueue::RunAsync()` in the calling thread and not
in the thread pool, which was hard to reason about.

As part of these changes I also fixed the code to support
move-only callables, and added a test for it.

The code is "future proofed" to support a `void()` callable.
Which is interesting for a number of reasons (think "executors").

Fixes #4083 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4448)

### DIFF
--- a/google/cloud/completion_queue.cc
+++ b/google/cloud/completion_queue.cc
@@ -76,6 +76,38 @@ class AsyncTimerFuture : public internal::AsyncGrpcOperation {
   std::unique_ptr<grpc::Alarm> alarm_;
 };
 
+class AsyncFunction : public internal::AsyncGrpcOperation {
+ public:
+  AsyncFunction(std::unique_ptr<internal::RunAsyncBase> fun,
+                std::unique_ptr<grpc::Alarm> alarm)
+      : fun_(std::move(fun)), alarm_(std::move(alarm)) {}
+
+  void Set(grpc::CompletionQueue& cq,
+           std::chrono::system_clock::time_point deadline, void* tag) {
+    if (alarm_) {
+      alarm_->Set(&cq, deadline, tag);
+    }
+  }
+
+  void Cancel() override {
+    if (alarm_) {
+      alarm_->Cancel();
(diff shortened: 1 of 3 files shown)


google/cloud/completion_queue.cc
```diff
@@ -76,6 +76,38 @@ class AsyncTimerFuture : public internal::AsyncGrpcOperation {
...
+ private:
+  bool Notify(bool) override {
+    fun_->exec();
+    fun_.reset();
+    return true;
+  }
+
+  std::shared_ptr<internal::CompletionQueueImpl> cq_;
+  std::unique_ptr<internal::RunAsyncBase> fun_;
+  // Holds the underlying handle, it might be a nullpotr in tests.
```
devbww, line 107: s/nullpotr/nullptr/

coryan: Fixed, thanks.

PR #4449, opened 2020-07-01
clean(source<T>): we will not use source<T>

coryan:
Remove the existing code for source<T>, it was all in the `internal`
namespace and/or tests.

Update API/ABI baseline to account for the changes in the `internal`
namespace.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4449)


coryan:
@devbww just FYI, I think this does not need to go into the CHANGELOG either, it is all `internal` stuff.

PR #4450, opened 2020-07-01
feat: CompletionQueue::RunAsync with no arguments

coryan:
Support callables that take no arguments (not even a CompletionQueue),
this is already useful in some tests, AsyncRowReader, and I think will
be useful if completion queues will model an executor.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4450)

### DIFF
--- a/google/cloud/completion_queue.h
+++ b/google/cloud/completion_queue.h
@@ -208,6 +208,32 @@ class CompletionQueue {
         absl::make_unique<Wrapper>(impl_, std::forward<Functor>(functor)));
   }
 
+  /**
+   * Asynchronously run a functor on a thread `Run()`ning the `CompletionQueue`.
+   *
+   * @tparam Functor the functor to call on the CompletionQueue thread.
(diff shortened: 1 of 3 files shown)


google/cloud/completion_queue.h
```diff
@@ -208,6 +208,32 @@ class CompletionQueue {
...
 
+  /**
+   * Asynchronously run a functor on a thread `Run()`ning the `CompletionQueue`.
+   *
+   * @tparam Functor the functor to call on the CompletionQueue thread.
+   *   It must satisfy the `void()` signature.
+   * @param functor the value of the functor.
+   * @return an asynchronous operation wrapping the functor; it can be used for
+   *   cancelling but it makes little sense given that it will be completed
+   *   straight away
```
devbww, line 219: s/$/./

coryan: Actually the whole `@return` section (here and in the other overload) was wrong, removed.

PR #4457, opened 2020-07-01
chore: some tweaks to the CHANGELOG.md instructions

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4457)

### DIFF
--- a/release/README.md
+++ b/release/README.md
@@ -55,7 +55,7 @@ git log --no-merges --format="format:* %s" \
    ':(exclude)google/cloud/storage/'
 ```
 
-Any _chore_/_ci_/_test_-tagged PRs in the above lists should probably be
+Any *chore*/*ci*/*test*-tagged PRs in the above lists should probably be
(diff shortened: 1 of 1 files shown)


release/README.md
```diff
@@ -22,28 +22,28 @@ Update `CHANGELOG.md` based on the release notes for Bigtable, Storage,
 Spanner, and the common libraries:
 
 ```bash
-# Summarize the output of this into CHANGELOG.md under the Bigtable header
+# Summarize the output of this into CHANGELOG.md under the "Bigtable" header
```
devjgm, line 25: No change needed, but I'll note that this section looks fairly automateable. I could imagine a small script like ./release/update_changelog.sh that does all this and leave the user with an edited CHANGELOG.md file that they can then manually further tweak and then send the PR.

devbww: Ack.  I'll see what I can come up with.

PR #4460, opened 2020-07-01
chore: expose `absl::nullopt` in our namespace

mr-salty:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4460)

### DIFF
--- a/google/cloud/optional.h
+++ b/google/cloud/optional.h
@@ -24,6 +24,7 @@ inline namespace GOOGLE_CLOUD_CPP_NS {
 
 template <typename T>
 using optional = absl::optional<T>;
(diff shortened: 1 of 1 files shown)


coryan:
As I said elsewhere, maybe we should consider `google::cloud::optional<>` as essentially there for backwards compatibility, and `absl::optional<>` (and its friends) should be the "One True" type we expose and document. If this is true, then adding more backwards compatibility declarations probably does not make sense. 
Thoughts?

mr-salty:
Closing per our discussion, we will migrate towards `absl` names and not introduce any new aliases in our namespace

Issue #4255, opened 2020-05-21
[Flake] Builds can exceed the Bucket creation rate limits


coryan:
https://source.cloud.google.com/results/invocations/193c0819-8f59-40c9-85fc-cb28a5af0fbd/targets

and 

https://cloud.google.com/storage/quotas

I have not thought much about it, probably putting the functions to create these temporary buckets in some kind of retry loop, with a backoff starting at 2 seconds would help.

coryan:
And another: https://source.cloud.google.com/results/invocations/c03056d5-26da-4eb4-8a4f-b9f98552dd49/targets

coryan:
https://source.cloud.google.com/results/invocations/a289a1b8-e1de-41c5-acc6-d6b3678bda0d/targets

coryan:
https://source.cloud.google.com/results/invocations/e47ec087-193a-44c9-b1eb-42c88815dd39

coryan:
The log in this one is unclear, but given what the test was trying to do, I think it is the same problem:

https://source.cloud.google.com/results/invocations/b87487e6-7d27-45c9-9a77-64786f0ff13c

coryan:
https://source.cloud.google.com/results/invocations/6ec40aa5-29ab-4fbb-b59c-23a97c818952/targets

coryan:
I am not planning to do more work on this, so unassigning myself. With the latest changes I hope the incidence will be lower, if there are no repeats we can (as usual) close this.

coryan:
No repeats in 90d, seems like the fixes in #4463 and #4453 have reduced the problem to the point we should not care anymore.

PR #4463, opened 2020-07-02
test: reduce flakiness in CreateBucket/DeleteBucket

coryan:
A single project (and we use a single project in our CI builds) can make
at most one CreateBucket or DeleteBucket request every 2 seconds. While
we have taken measures to minimize the number of times we touch
production (e.g. the CMake builds use the testbench, the Bazel builds
cache tests), sometimes there are enough CI builds hitting production
that we exceed this limit.

In this PR we take some additional measures to reduce the problems:

- The ci/kokoro/docker builds based on Bazel now use the GCS emulator
  (aka testbench) for most tests (a few tests must use production, but
  they do not create buckets AFAIK). This leaves the macOS and Windows
  builds running against production.
- Any loops that create buckets with different configurations now pause
  before each iteration. I do not think this slows down the builds
  because those "fast" calls failed and needed to retry anyway.
- For integration tests that create and/or delete buckets we use a
  longer initial backoff (5s) and longer total retry period (10m), as
  the default (for integration tests) were too short to actually work.
- In the examples, where we often create a temporary bucket, we pause
  before deleting the bucket if the total time is less than 2 seconds.
  Otherwise the delete operation will need to retry anyway.

This should minimize the problems in #4255

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4463)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -130,6 +130,13 @@ if should_run_integration_tests; then
   )
 
   readonly EMULATOR_SCRIPT="run_integration_tests_emulator_bazel.sh"
+
+  echo
+  io::log_yellow "running storage integration tests via Bazel+Emulator"
+  echo
+  "${PROJECT_ROOT}/google/cloud/storage/ci/${EMULATOR_SCRIPT}" \
+    "${BAZEL_BIN}" "${bazel_args[@]}"
+
   # TODO(#441) - remove the for loops below.
   # Sometimes the integration tests manage to crash the Bigtable emulator.
   # Manually restarting the build clears up the problem, but that is just a
@@ -175,6 +182,9 @@ if should_run_integration_tests; then
     # was called above. The one exception is the bigtable_grpc_credentials
     # test, which requires an access token and will be run separately.
     "-//google/cloud/bigtable/..."
+
+    # The Storage integration tests were already run above
+    "-//google/cloud/storage/..."
   )
   for t in "${hmac_service_account_targets[@]}" "${access_token_targets[@]}"; do
     excluded_targets+=("-${t}")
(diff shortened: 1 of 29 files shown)


ci/kokoro/docker/build-in-docker-bazel.sh
```diff
@@ -130,6 +130,13 @@ if should_run_integration_tests; then
   )
 
   readonly EMULATOR_SCRIPT="run_integration_tests_emulator_bazel.sh"
+
+  echo
+  io::log_yellow "running storage integration tests via Bazel+Emulator"
+  echo
```
devbww, line 136: Perhaps this should match lines 148-149 below, and just do a single "==="s banner and no extra blank lines?

google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
```diff
@@ -0,0 +1,95 @@
...
+  exit 1
+fi
+
+# Configure run_emulators_utils.sh to find the instance admin emulator.
+BAZEL_BIN="$1"
+shift
+bazel_test_args=("$@")
+
+# Configure run_emulators_utils.sh to find the instance admin emulator.
+# Configure run_emulators_utils.sh to find the instance admin emulator.
```
devbww, line 32: Duplicated line?

devbww, line 32: s/emulators/testbench/?

google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
```diff
@@ -0,0 +1,95 @@
...
+# the testbench. Note that the HMAC service account is completely invalid and
+# it is not unique to each test, neither is a problem when using the emulator.
+"${BAZEL_BIN}" test "${bazel_test_args[@]}" "${testbench_args[@]}" \
+  --test_tag_filters="integration-test" -- \
+  "//google/cloud/storage/...:all" \
+  "${excluded_targets[@]}"
+exit_status=$?
+
+kill_testbench
+trap '' EXIT
```
devbww, line 93: This appears to depending upon the implementation of `run_testbench_utils`.  Given that we're just about to exit, should we prefer to just let the EXIT handler run by itself?

google/cloud/storage/examples/storage_bucket_acl_samples.cc
```diff
@@ -246,10 +246,13 @@ void RunAll(std::vector<std::string> const& argv) {
...
             << std::endl;
-  auto bucket_metadata = client
-                             .CreateBucketForProject(bucket_name, project_id,
-                                                     gcs::BucketMetadata{})
-                             .value();
+  (void)client
+      .CreateBucketForProject(bucket_name, project_id, gcs::BucketMetadata{})
+      .value();
+  // In GCS a single project cannot create or delete buckets more often than
+  // once every two seconds. We will pause for at least that long before
```
devbww, line 253: s/for at least that long/until at least that time/ perhaps (et seq)?

google/cloud/storage/testing/storage_integration_test.cc
```diff
@@ -44,36 +44,55 @@ StorageIntegrationTest::MakeIntegrationTestClient() {
...
 
+google::cloud::StatusOr<google::cloud::storage::Client>
+StorageIntegrationTest::MakeBucketIntegrationTestClient() {
+  if (UsingTestbench()) return MakeIntegrationTestClient();
+
+  auto constexpr kInitialDelay = std::chrono::seconds(5);
+  auto constexpr kMaximumBackoffDelay = std::chrono::minutes(5);
+  auto constexpr kBackoffScalingFactor = 2.0;
+  return MakeIntegrationTestClient(
+      LimitedTimeRetryPolicy(/*maximum_duration=*/2 * kMaximumBackoffDelay)
```
devbww, line 55: Is the 2 the kBackoffScalingFactor?

coryan: I think what I wrote now is clearer, let me know.

google/cloud/storage/examples/storage_bucket_acl_samples.cc
coryan: Done.

google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
coryan: Good idea, done.

google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
coryan: Fixed, I think.

coryan: Fixed.

ci/kokoro/docker/build-in-docker-bazel.sh
coryan: Done.

PR #4468, opened 2020-07-03
feat: `CreateDatabase` takes an optional encryption key

### LINKED ISSUE #4312: Spanner: Implement CMEK support
Internal tracking: b/157706906

mr-salty:
There's nothing to do for `ListDatabases` and `GetDatabase` since we don't have our own types and just return the protos.

DO NOT SUBMIT - this uses protos from the `googleapis/googleapis` `spanner-cmek` branch; 
this cannot be merged until those protos are live.

Fixes #4312

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4468)

### DIFF
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -75,10 +75,10 @@ def google_cloud_cpp_deps():
         http_archive(
             name = "com_google_googleapis",
             urls = [
-                "https://github.com/googleapis/googleapis/archive/59f97e6044a1275f83427ab7962a154c00d915b5.tar.gz",
+                "https://github.com/googleapis/googleapis/archive/spanner_cmek.tar.gz",
             ],
-            strip_prefix = "googleapis-59f97e6044a1275f83427ab7962a154c00d915b5",
-            sha256 = "5e785c25b1d57973e7481b4da226d7c73056ea22c7545bf6d14dbebf6e99b073",
+            strip_prefix = "googleapis-spanner_cmek",
+            #sha256 = "",
             build_file = "@com_github_googleapis_google_cloud_cpp//bazel:googleapis.BUILD",
         )
 
--- a/external/googleapis/CMakeLists.txt
+++ b/external/googleapis/CMakeLists.txt
@@ -19,10 +19,10 @@ cmake_minimum_required(VERSION 3.5)
 # Give application developers a hook to configure the version and hash
 # downloaded from GitHub.
 set(GOOGLE_CLOUD_CPP_GOOGLEAPIS_URL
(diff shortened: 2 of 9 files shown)


coryan:
FWIW, this is a fixable problem:

```
Error in non-idempotent operation CreateDatabase: KMS Key provided is not usable: ServiceAccount: service-936212892354@gcp-sa-spanner.iam.gserviceaccount.com KmsKeyName: projects/cloud-cpp-testing-resources/locations/us-central1/keyRings/spanner-cmek/cryptoKeys/spanner-cmek-test-key Error: Permission 'cloudkms.cryptoKeyVersions.useToEncrypt' denied on resource 'projects/cloud-cpp-testing-resources/locations/us-central1/keyRings/spanner-cmek/cryptoKeys/spanner-cmek-test-key' (or it may not exist).. [PERMISSION_DENIED]
```

You need to give spanner permission to use the test key you created, we can do this over GVC if you want.

mr-salty:
It works in my client if I point it to staging.  I'm surprised we get this error at all in prod, it seems like it should be disregarding the key entirely.

In any case, I had no plans to spend any time looking at this until the feature is officially released and we're ready to merge into the main branch. Is there a way I can disable tests here? It's just a waste of resources.

coryan:
> 
> In any case, I had no plans to spend any time looking at this until the feature is officially released and we're ready to merge into the main branch. Is there a way I can disable tests here? It's just a waste of resources.

Close the PR, that will disable the CI builds completely.

mr-salty (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4468)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @devjgm)

___
*[google/cloud/kms_key_name.h, line 49 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4468#-MBHlYHItnBWVIi3NdK7-r1-49:-MHJK2tT0Th4PWwit9nR:b6a6giw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0dde3692c9981f342263976f8ea4012218cbd337/google/cloud/kms_key_name.h#L49)):*

Done.  I was actually going to pull this class out and check it into the main branch; there are some existing uses of KMS Key Names in pubsub and storage that I was planning to move over (currently they're using std::string), so I thought this constructor would help... but looking at it again maybe I can get by without it.

I was on the fence about having this be a type at all vs just using `std::string` and having a `std::string MakeKmsKey()` method. in the end I thought the type added just enough to be worthwhile, but I could be convinced otherwise.
</blockquote></details>

just marking this resolved - we decided to use a factory function `MakeKmsKeyName` and make the constructor private. This was also checked into `master` independently.

mr-salty:
closing this PR for now to avoid triggering PR builds when I rebase (which I'm trying to do more frequently since I let it go for a while and it was a bit painful)

Issue #4437, opened 2020-06-27
Add ReadRange ( and maybe ReadFromOffset, ReadLast ) to UploadFile 


vnvo2409:
**What component of `google-cloud-cpp` is this feature request for?**
`google/cloud/storage`

**Describe the solution you'd like**
I would like to add `ReadRange` to `UploadFile`, So I can upload only a part of the local file. This could solve the problem with file with infinite source
https://github.com/googleapis/google-cloud-cpp/blob/42b7087a20dbd8f013e823100d8bed05e154e9d4/google/cloud/storage/client.cc#L129-L130

**Describe alternatives you've considered**
Load the content to an `ObjectWriteStream` but it will be much easier to use `UploadFile`

**Additional context**
I think it is not too much complicated to implement this feature. With `UploadFileSimple`, we need to change this line
https://github.com/googleapis/google-cloud-cpp/blob/42b7087a20dbd8f013e823100d8bed05e154e9d4/google/cloud/storage/client.cc#L115
With `UploadFileResumable`, we need to add a offset to this line
https://github.com/googleapis/google-cloud-cpp/blob/42b7087a20dbd8f013e823100d8bed05e154e9d4/google/cloud/storage/client.cc#L165 
and change the condition of while
https://github.com/googleapis/google-cloud-cpp/blob/42b7087a20dbd8f013e823100d8bed05e154e9d4/google/cloud/storage/client.cc#L175-L176

I could work on this feature

coryan:
This is a nice feature request.  Reusing `ReadRange()` is a bit worrisome, what do you think if we create a new option called `UploadRange()` (and maybe `UploadLimit()`)?

vnvo2409:
So we will have 3 options:
- `UploadRange()` for upload from bytes start to bytes end
- `UploadLimit()` the maximum number of bytes to upload ( for file with infinite sources )
- `UploadFromOffset()` upload from an offset.
I will send you a PR for `UploadFile`

vnvo2409:
Close by https://github.com/googleapis/google-cloud-cpp/pull/4473#event-3530049441

PR #4473, opened 2020-07-05
feat(storage): Add `UploadFromOffset` and `UploadLimit` to `UploadFile`

vnvo2409:
This PR is a part of #4437 
I would like to ask where I should write test for this PR. Thank you

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4473)

### DIFF
--- a/google/cloud/storage/client.cc
+++ b/google/cloud/storage/client.cc
@@ -128,7 +128,7 @@ StatusOr<ObjectMetadata> Client::UploadFileSimple(
   std::string payload(upload_size, char{});
   is.seekg(upload_offset, std::ios::beg);
   is.read(&payload[0], payload.size());
-  if (is.gcount() < payload.size()) {
+  if (static_cast<size_t>(is.gcount()) < payload.size()) {
     std::ostringstream os;
     os << __func__ << "(" << request << ", " << file_name << "): Actual read ("
        << is.gcount() << ") is smaller than upload_size (" << payload.size()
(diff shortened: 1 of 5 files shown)


vnvo2409:
About `fifo` file, it is a little bit more complicated since we have to have 2 threads, one writes data to `fifo` and one uploads the data to GCS server simultaneously. Please review the current changes first and I will add the test for `fifo` later. Thank you.

vnvo2409:
I will wait for #4408. In the mean time, I will do some tests with upload for irregular file

coryan:
@vnvo2409  we merged #4408 earlier today in case you want to pick this up again.

vnvo2409:
I think that we have to write another helper function like `UploadFileSpecial` to deal with special file. After some experiments, C++ api is really bad at dealing with `fifo`:
- `seekg()` does not work.
- We have to read only 1 byte at a time because `gcount(), tellg()` do not work. 
- when we call `fstream.close()`, it causes the thread which is writing to `fifo` stop too. This may cause some errrors since the thread that is writing to `fifo` isn't related to `google-cloud-cpp`.

So I think we should have a seperate PR for the special file. What do you think ?

coryan:
> So I think we should have a separate PR for the special file. What do you think ?

Yes, it seems that getting those to work should be a separate bug and PR.  We might decide to not have them work at all (as opposed to just warning about them).

vnvo2409:
Test failed because
```
/v/google/cloud/storage/client.cc:131:19: error: comparison of integers of different signs: 'streamsize' (aka 'long') and 'size_type' (aka 'unsigned long') [-Werror,-Wsign-compare]
  if (is.gcount() < payload.size()) {
      ~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~
1 error generated.
```

vnvo2409:
Done 🎉

PR #4474, opened 2020-07-06
test: deflake CompletionQueueTest.ShutdownWithPending

### LINKED ISSUE #4384: [Flake] CompletionQueue test flakes
https://source.cloud.google.com/results/invocations/15ce958e-aac5-428b-9e7f-712cd2dceb30/targets

```
 54/263 Test  #31: common_grpc_utils_completion_queue_test ...........................***Failed    1.43 sec
[==========] Running 14 tests from 1 test suite.
[----------] 14 tests from CompletionQueueTest
[ RUN      ] CompletionQueueTest.TimerSmokeTest
[       OK ] CompletionQueueTest.TimerSmokeTest (34 ms)
[ RUN      ] CompletionQueueTest.MockSmokeTest
[       OK ] CompletionQueueTest.MockSmokeTest (5 ms)
[ RUN      ] CompletionQueueTest.ShutdownWithPending
/v/google/cloud/completion_queue_test.cc:122: Failure
  std::future_status::timeout
    Which is: 4-byte object <01-00 00-00>
  timer.wait_for(ms(0))
    Which is: 4-byte object <00-00 00-00>
/v/google/cloud/completion_queue_test.cc:124: Failure
  std::future_status::timeout
    Which is: 4-byte object <01-00 00-00>
  timer.wait_for(ms(0))
    Which is: 4-byte object <00-00 00-00>
```

coryan:
Very rarely this test would fail because the 20ms timer expired before
the very next line could run (scheduling is hard). Fixed the test by (a)
making the timer 500ms, and (b) put a loop to create a new timer if the
first few expired.

In a few cases this would be testing "nothing" as the timer could expire
before the completion queue really is shutdown, but that (rare) false
positive is better (I think) than rare false negatives.

Fixes #4384 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4474)

### DIFF
--- a/google/cloud/completion_queue_test.cc
+++ b/google/cloud/completion_queue_test.cc
@@ -112,14 +112,20 @@ TEST(CompletionQueueTest, ShutdownWithPending) {
   {
     CompletionQueue cq;
     std::thread runner([&cq] { cq.Run(); });
-    timer = cq.MakeRelativeTimer(ms(20)).then(
-        [](future<StatusOr<std::chrono::system_clock::time_point>> result) {
-          // Timer still runs to completion after `Shutdown`.
-          EXPECT_STATUS_OK(result.get().status());
-        });
-    EXPECT_EQ(std::future_status::timeout, timer.wait_for(ms(0)));
+    // In very rare conditions the timer would expire before we get a chance to
+    // call `cq.Shutdown()`, use a little loop to avoid this. In most runs this
+    // loop will exit after the first iteration. By my measurements the loop
+    // would need to run more than once every 10,000 runs. On the other hand,
+    // this should prevent the flake reported in #4246.
+    for (auto i = 0; i != 5; ++i) {
+      timer = cq.MakeRelativeTimer(ms(500)).then(
+          [](future<StatusOr<std::chrono::system_clock::time_point>> f) {
+            // Timer still runs to completion after `Shutdown`.
+            EXPECT_STATUS_OK(f.get().status());
+          });
+      if (timer.wait_for(ms(0)) == std::future_status::timeout) break;
+    }
     cq.Shutdown();
-    EXPECT_EQ(std::future_status::timeout, timer.wait_for(ms(0)));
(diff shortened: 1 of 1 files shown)


google/cloud/completion_queue_test.cc
```diff
@@ -453,6 +459,8 @@ TEST(CompletionQueueTest, ShutdownWithReschedulingTimer) {
 
   cq.Shutdown();
   t.join();
+  // This is a "it did not crash nor hung" test, if we get here it was
```
devbww, line 462: "did not crash or hang" perhaps?  Or "neither crashed nor hung"?

google/cloud/completion_queue_test.cc
```diff
@@ -112,14 +112,20 @@ TEST(CompletionQueueTest, ShutdownWithPending) {
...
-    timer = cq.MakeRelativeTimer(ms(20)).then(
-        [](future<StatusOr<std::chrono::system_clock::time_point>> result) {
-          // Timer still runs to completion after `Shutdown`.
-          EXPECT_STATUS_OK(result.get().status());
-        });
-    EXPECT_EQ(std::future_status::timeout, timer.wait_for(ms(0)));
+    // In very rare conditions the timer would expire before we get a chance to
+    // call `cq.Shutdown()`, use a little loop to avoid this. In most runs this
+    // loop will exit after the first iteration. By my measurements the loop
+    // would need to run more than once every 10,000 runs. On the other hand,
```
devbww, line 118: Should be expect, then, that we never reach `i == 5`?

google/cloud/completion_queue_test.cc
coryan: Fixed.

google/cloud/completion_queue_test.cc
coryan: I think you are right, but I wrote a comment because it is weird to have an assertion line that.

PR #4488, opened 2020-07-08
refactor(GCS+gRPC): simplify resumable upload code

coryan:
The streaming uploads have a number of restrictions on message sizes,
chunk quantums, and final message flags. This refactors the code to keep
all this complexity in one place. It also dtakes advantage of the
`google.storage.v1.ServiceConstants` proto which did not exist or I
missed when we originally implemented this code. The code has more
conditional code now, but does not create more data copies than before.
Unfortunately a copy is more or less unavoidable because protos need to
own the payload and we cannot move it (both because we might need it on
a retry and because it is received as a `std::string const&`).

This will make other changes for #2664 much easier.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4488)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -37,8 +37,6 @@ namespace storage {
 inline namespace STORAGE_CLIENT_NS {
 namespace internal {
 
-std::size_t constexpr GrpcClient::kMaxInsertObjectWriteRequestSize;
-
 bool DirectPathEnabled() {
   auto const direct_path_settings =
       google::cloud::internal::GetEnv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH")
@@ -205,11 +203,8 @@ StatusOr<ObjectMetadata> GrpcClient::InsertObjectMedia(
   google::storage::v1::Object response;
   auto stream = stub_->InsertObject(&context, &response);
   auto proto_request = ToProto(request);
-  // This limit is for the *message*, not just the payload. It includes any
-  // additional information such as checksums. We need to use a stricter limit,
-  // a chunk quantum seems to work in practice.
   std::size_t const maximum_buffer_size =
-      kMaxInsertObjectWriteRequestSize - UploadChunkRequest::kChunkSizeQuantum;
+      google::storage::v1::ServiceConstants_Values_MAX_WRITE_CHUNK_BYTES;
   auto const& contents = request.contents();
 
   // This loop must run at least once because we need to send at least one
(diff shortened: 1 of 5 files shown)


google/cloud/storage/internal/grpc_client.cc
```diff
@@ -205,11 +203,8 @@ StatusOr<ObjectMetadata> GrpcClient::InsertObjectMedia(
   google::storage::v1::Object response;
   auto stream = stub_->InsertObject(&context, &response);
   auto proto_request = ToProto(request);
-  // This limit is for the *message*, not just the payload. It includes any
-  // additional information such as checksums. We need to use a stricter limit,
-  // a chunk quantum seems to work in practice.
   std::size_t const maximum_buffer_size =
-      kMaxInsertObjectWriteRequestSize - UploadChunkRequest::kChunkSizeQuantum;
+      google::storage::v1::ServiceConstants_Values_MAX_WRITE_CHUNK_BYTES;
```
devjgm, line 207: I think we can avoid using the `*_Values_*` enums by accessing the named fields in the `ServiceConstants` class. Like this could be `google::storage::v1::ServiceConstants::MAX_WRITE_CHUNK_BYTES`.

google/cloud/storage/internal/grpc_resumable_upload_session.cc
```diff
@@ -163,6 +86,67 @@ GrpcResumableUploadSession::last_response() const {
...
+    request.clear_insert_object_spec();
+    request.clear_object_checksums();
+
+    next_expected_ += n;
+    return true;
+  };
+
+  std::size_t payload_offset = 0;
+  do {
+    chunk.reserve(maximum_chunk_size);
```
devjgm, line 133: Is there a reason the reserve is in this loop rather than up on line 96 right after `chunk` is declared?

coryan: Yes. It is (potentially) moved-out in the `flush_chunk()` lambda, around line 107.

coryan: I will move this around to try to make it clearer.

google/cloud/storage/internal/grpc_client.cc
coryan: Thanks, fixed.

google/cloud/storage/internal/grpc_resumable_upload_session.cc
coryan: Fixed.

coryan:
Rebased and address comments. PTAL.

PR #4489, opened 2020-07-09
feat(storage): reduce copies during uploads

coryan:
Reduce the number of copies during uploads, change `UploadChunkRequest`
to receive a vector of buffers for the payload, and use that vector when
making the PUT request. We still have copies in the `std::ostream`
layer, those will be fixed in a future PR.

Because most of the time we only send one buffer, I changed CurlRequest
to avoid using read callbacks if possible.

Part of the changes for #2664.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4489)


coryan:
For future reference.

### Bandwidth Improvements

                   count       mean       std        min        25%        50%        75%        max
ApiName Version                                                                                    
XML     baseline  5412.0  42.651381  6.371607   7.954328  39.736107  43.797344  46.950373  54.584919
        develop   6081.0  47.517458  5.341434  12.227134  44.503985  48.244759  51.170239  61.618113

image


### CPU Utilization Improvements

                  count      mean       std       min       25%       50%       75%       max
ApiName Version                                                                              
XML     baseline  5412.0  1.637312  0.252425  1.325035  1.499121  1.560917  1.654678  2.928191
        develop   6081.0  1.355525  0.148113  1.128924  1.269584  1.318617  1.379592  2.169694

image

PR #4490, opened 2020-07-09
fix: workaround min/max macros on Windows

coryan:
Unfortunately our PR builds for Windows are based on Bazel which defines
`NOMINMAX` which hides these problems, and I missed them in my review.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4490)

### DIFF
--- a/google/cloud/storage/client.cc
+++ b/google/cloud/storage/client.cc
@@ -112,10 +112,9 @@ StatusOr<ObjectMetadata> Client::UploadFileSimple(
        << ") is bigger than the size of file source (" << file_size << ")";
     return Status(StatusCode::kInvalidArgument, std::move(os).str());
   }
-  auto upload_size = request.HasOption<UploadLimit>()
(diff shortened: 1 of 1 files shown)


coryan:
> Did something actually break because of this?

Yes.  https://source.cloud.google.com/results/invocations/551b51fe-53da-4a10-a778-9e86e37dd15f

> My understanding of the problem is that this workaround isn't actually needed in .cc files because users' includes don't affect the compilation of _our_ .cc files[1]. This differs from our header files, where a user may easily include a Windows header that defines min/max macros before including one of our headers, which would break if we used a naked `std::min`.

Right. I am guessing that something we include (probably curl or maybe openssl) includes the windows headers that get these macros.

> Of course, thinking through these rules all the time is cumbersome and it's easier just to always do the workaround with parens. I'm curious if you visually spotted this inconsistency, or if this is actually fixing a build break on Windows?

See above.

devjgm:
Thanks. LGTM. I already approved.

PR #4491, opened 2020-07-09
refactor(storage): upload chunks can have many buffers

### LINKED ISSUE #3036: Allow for not copying data in upload retries
Currently member functions `UploadChunk()` and `UploadFinalChunk()` accept a `std::string const &` as the data carrier.

On retries, there is a need to pass only a part of this data to lower layers, which is impossible with this type. We should change it to something that would allow it, e.g.:
* raw `char *` and size
* something along the lines of `boost::buffer<>`
* ...

coryan:
This change allows the `ObjectWriteStreambuf` to only upload a portion
of a larger buffer via a `ResumableUploadSession`, that removes one copy
from the upload path.

In addition, `ResumableUploadSession` now accepts more than one buffer
as the payload, which is a necessary step to remove more additional data
copies from the upload path.

We also use this in the retry layer to avoid a copy when only part of the
data was committed, as the retry can just shrink the buffers without copying
them.

Part of the changes for #2664 and fixes #3036 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4491)


coryan:
For future reference:

## Bandwidth Improvements

                   count       mean       std        min        25%        50%        75%        max
ApiName Version                                                                                     
XML     baseline  6184.0  46.221941  5.782130   3.189308  43.861531  47.199521  49.809935  58.254358
        develop   6768.0  49.879409  4.893566  17.052791  47.530999  50.603315  53.147572  60.879723

image


## CPU Time per Byte Improvements (nanoseconds)

                   count      mean       std       min       25%       50%       75%       max
ApiName Version                                                                               
XML     baseline  6184.0  1.679852  0.299407  1.352087  1.507697  1.567884  1.677669  2.941741
        develop   6768.0  1.123189  0.082659  0.934905  1.065313  1.111653  1.164754  1.738695

￼
image

Issue #4482, opened 2020-07-08
`prefixes` in `ListObjectReponse`


vnvo2409:
**What component of `google-cloud-cpp` is this feature request for?**
`storage`

**Is your feature request related to a problem? Please describe.**
I am implementing a `GetChildren` function will work in a directory-like mode.

**Describe the solution you'd like**
I would like to have access to `prefixes` in the reponse.
**Additional context**
If this is not possible, Is there any way I could access to the raw reponse instead of `ListObjectsReader`

coryan:
I think we completely missed `prefixes` 😞 

The C++ object to process the response (`internal::ListObjectsResponse`) does not have `prefixes` either, so either way we would need to make changes.

I think the "Right Thing":tm: here is to add a new function to `Client` that returns an iterator over the prefixes. Maybe something like:

```{c++}
class Client { public:
  ListObjectPrefixesReader ListMatchingPrefixes(std::string bucket_name, ...);
};
```

`ListObjectPrefixesReader` would be a single pass input range of `std::string`, and the call could always use a suitable `Fields()` value to discard the data it will not use.

The downside is that you need two sets of RPCs: (1) to discover the prefixes that match and (2) to discover the objects that match.  OTOH, prefixes are very different from object names and/or object metadata, so maybe it makes sense to have completely different APIs?

As an alternative we could create a function called `ListObjectNamesAndPrefixes` and return a single pass input range of `std::string` and `Kind` (wrapped in a struct or pair) where `Kind` would tell you if the string represents a prefix or a real object name.

Or thinking about it some more, maybe `absl::variant<std::string, ObjectMetadata>` is an alternative here... that is super hard to use though.

vnvo2409:
> As an alternative we could create a function called ListObjectNamesAndPrefixes and return a single pass input range of std::string and Kind (wrapped in a struct or pair) where Kind would tell you if the string represents a prefix or a real object name.

I think this makes sense because if someone wants `prefixes`, they usually need the name of the object only. I could work on this features but I may need some instructions, What do you think ?

coryan:
Sorry for the delay, we had a quick back and forth on this internally.  It seems we settled on this:

```C++
using ListObjectsAndPrefixesReader =
    internal::PaginationRange<
        absl::variant<std::string, ObjectMetadata>,
        internal::ListObjectsRequest,
        internal::ListObjectsResponse>;

class Client { public:
[...]

PR #4494, opened 2020-07-10
feat(storage): Add `ListObjectsAndPrefixes`

vnvo2409:
This is a draft PR for #4482

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4494)


vnvo2409:
This is just a draft, I test with my bucket on GCS server but I am not too familiar with mock test so it may take sometime to write the tests.

vnvo2409:
How about we add a member variable `std::function<std::vector<T>(Reponse r) get_items_` with the default value is 
```cpp
[](Reponse r) {
  return r.items;
}
```

If you dont want the default value, we could check https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool
```cpp
if(get_items_) {
  result = get_items_(r);
} else {
  result = r.items;
}
```

coryan:
> How about we add a member variable `std::function<std::vector<T>(Reponse r) get_items_` with the default value is
> 
> ```c++
> [](Reponse r) {
>   return r.items;
> }
> ```

I think that is what we need.  nit: maybe it should do `return std::move(r.items_)` by default.

vnvo2409:
Once more question, How do we sort the returned result ? ( in which order, prefixes or objects first, etc, ... )

coryan:
I could kick off a run of the CI builds, but they won't be testing much: the CI builds use the GCS testbench to run, and the testbench (think "simplified emulator") does not have support for prefixes (or filtering!) either:

https://github.com/googleapis/google-cloud-cpp/blob/b308a107402f28e097898e91a07295231878cdcf/google/cloud/storage/testbench/testbench.py#L368

The test you added is skipped when using the testbench.  You still need to write some unit tests.  

You can run the integration tests locally using:

```
RUN_INTEGRATION_TESTS=yes ./ci/kokoro/docker/build.sh clang-tidy
```

I need to fix a bug in the script, but the problem shows up after all  the integration tests are executed anyway.

coryan:
Doh!  there is no bug, run like this for clean results: `RUN_INTEGRATION_TEST=auto ./ci/kokoro/docker/build.sh clang-tidy`

vnvo2409:
my computer is super slow so it will take a while.

vnvo2409:
I add a unit test for `ListObjectsAndPrefixesReader`. I adding something into `testbench.py` to make it work with `gcs::Prefix` and `gcs::Delimiter` and of course remove `if (UsingTestbench()) GTEST_SKIP();`. The only thing left to do now is documentation and sample, which should not be too complicated.

vnvo2409:
All tests passed the last time

vnvo2409:
The loop doesn't run at all, I think maybe there is nothing on the server to list. I will check it

vnvo2409:
The sample runs fine on my machine, I think there isn't anything on the server to list. does the sample use testbench as well ?

vnvo2409:
@coryan 
Indeed, I am a GSoC student who is working for Tensorflow to port some filesystem to modular c api filesystem. Since I would like to get more involved in open source in this summer, please tell me if there is an issue that is suitable for me. Thank you

coryan:
We should probably move this discussion to email, my email is easy enough to discover from the commit logs.

In any case, I just labeled a number of issues in this repository as "good first issue".

If you are interested in something larger, a couple of ideas. In both cases we would need to how this outside google-cloud-cpp, but we can find it a home:

- It would be nice to have a demo showing how to perform client-side encryption for GCS, it should be possible by using tink (https://github.com/google/tink) and Boost.Iostreams (https://www.boost.org/doc/libs/1_73_0/libs/iostreams/doc/index.html)
- I have 1/2 of a demo showing how to read all the metadata from a GCS bucket and put it in Spanner (for quicker searches over arbitrary attributes, not just names).  And then maintain the data up to date using a serverless Cloud Run function. Completing this demo and writing up something on how to use it would be nice too.

megaserg:
It seems `absl::variant` should be added as a dependency to `google/cloud/storage/CMakeLists.txt` as well :)

coryan:
@megaserg said:

> It seems `absl::variant` should be added as a dependency to `google/cloud/storage/CMakeLists.txt` as well :)

This is messy.  The short answer is yes. But I think it was working despite this problem because `absl::variant` was an unnecessary dependency of `google_cloud_cpp_common`.

The dependency was missing in our `pkg-config` files, for a host of unpleasant reasons those are maintained by hand and it is easy to miss these things.  I prepared #5051 to make such mistakes less likely.  Maybe just for me to learn something: could you share what exactly was not working for you?

megaserg:
Thanks! I am trying to add GCS support in [apache/arrow](https://github.com/apache/arrow/). I've built and installed the google-cloud-cpp project as described in doc/packaging.md (on Ubuntu 18.04), except that I made shared libs. Then, in Arrow's cmake config, I linked with `/usr/local/lib/libgoogle_cloud_cpp_common.so;/usr/local/lib/libstorage_client.so`. Then, I wrote some C++ code implementing a filesystem interface for Arrow, using `ListObjectsAndPrefixes`. That code went into the `libarrow.so` (the main artifact of the apache/arrow). `libarrow.so` has built okay, but I got an error when linking tests (which I did not change):

```
Linking CXX executable ../../../debug/arrow-json-test

cd /home/sergey.serebryakov/work/arrow/cpp/debug/src/arrow/json && /usr/bin/cmake -E cmake_link_script CMakeFiles/arrow-json-test.dir/link.txt --verbose=1
/usr/bin/ccache /usr/bin/c++   -Wno-noexcept-type  -fdiagnostics-color=always -ggdb -O0  -Wall -Wno-conversion -Wno-sign-conversion -Wno-unused-variable -Werror -msse4.2  -g  -rdynamic CMakeFiles/arrow-json-test.dir/chunker_test.cc.o CMakeFiles/arrow-json-test.dir/converter_test.cc.o CMakeFiles/arrow-json-test.dir/parser_test.cc.o CMakeFiles/arrow-json-test.dir/reader_test.cc.o  -o ../../../debug/arrow-json-test -Wl,-rpath,/home/sergey.serebryakov/work/arrow/cpp/debug/debug:/usr/local/lib ../../../debug/libarrow_testing.so.18.0.0 ../../../debug/libarrow.so.18.0.0 /usr/lib/x86_64-linux-gnu/libcrypto.so /usr/lib/x86_64-linux-gnu/libssl.so /usr/local/lib/libgoogle_cloud_cpp_common.so /usr/local/lib/libstorage_client.so -ldl /usr/local/lib/libgtest_main.so /usr/local/lib/libgtest.so /usr/local/lib/libgmock.so /usr/lib/x86_64-linux-gnu/libboost_filesystem.so /usr/lib/x86_64-linux-gnu/libboost_system.so -ldl /usr/lib/x86_64-linux-gnu/libcrypto.so ../../../jemalloc_ep-prefix/src/jemalloc_ep/dist//lib/libjemalloc_pic.a -lrt -pthread

../../../debug/libarrow.so.18.0.0: undefined reference to `absl::lts_2020_02_25::variant_internal::ThrowBadVariantAccess()'
```

If I use `ListObjects` instead, the tests build and run just fine.

I did `ldd /usr/local/lib/libstorage_client.so` and found that while it depends on some Abseil libraries, it doesn't depend on `libabsl_bad_variant_internal.so`. I'm actually not sure why everything (libstorage_client.so, libarrow.so) has built just fine until the seemingly unrelated tests :)

coryan:
Thanks for the detailed description, this may help us design better tests to catch such issues in the future.

> Thanks! I am trying to add GCS support in [apache/arrow](https://github.com/apache/arrow/).

Oh cool!

> I've built and installed the google-cloud-cpp project as described in doc/packaging.md (on Ubuntu 18.04), except that I made shared libs.

Okay.

> Then, in Arrow's cmake config, I linked with `/usr/local/lib/libgoogle_cloud_cpp_common.so;/usr/local/lib/libstorage_client.so`. 

First, I believe the order of those dependencies is backwards, you want `libstorage_client.so` *before* `libgoogle_cloud_cpp_common.so`.  I believe (but you need to test this) that reversing the order would fix the problem you are experiencing.

Second, any reason you cannot use the imported target? As in:

```CMake
find_package(CURL REQUIRED)
find_package(storage_client REQUIRED)

# Use `storage_client` as a target, e.g.:
target_link_libraries(<your-target-here-maybe-libarrow?> storage_client)
```

That should work for static libraries, shared libraries, and wherever the library happens to be installed. There is an example of how to use this at:

https://github.com/googleapis/google-cloud-cpp/tree/master/google/cloud/storage/quickstart
https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/storage/quickstart/CMakeLists.txt

FWIW, I tested a similar program, one that actually used `ListObjectsAndPrefixes()`, with shared libraries and it works, even before the #5051 PR.

>Then, I wrote some C++ code implementing a filesystem interface for Arrow, using `ListObjectsAndPrefixes`. That code went into the `libarrow.so` (the main artifact of the apache/arrow). `libarrow.so` has built okay, but I got an error when linking tests (which I did not change):

I think (but maybe I am wrong) this error is a result of (a) we not declaring the direct dependencies of `libstorage_client.so` correctly and (b) you declaring the libraries in the wrong order. Fixing either would fix the problem. Also using the imported CMake target would fix the problem (because it implicitly does (b)).

Using the imported CMake targets is probably "better" as it likely works with static libraries too, and it would continue to work if we ever need to add or change the dependencies for our libraries..

> I did `ldd /usr/local/lib/libstorage_client.so` and found that while it depends on some Abseil libraries, it doesn't depend on `libabsl_bad_variant_internal.so`.

That should be fixed in that PR above, but I recommend you use the imported target instead.

> I'm actually not sure why everything (libstorage_client.so, libarrow.so) has built just fine until the seemingly unrelated tests :)

Most likely because `--no-undefined` is **not** [enabled by default](https://linux.die.net/man/1/ld), that is, undefined symbols are allowed in shared objects and binaries that use them.

megaserg:
Thank you for detailed explanation!

Switching the order like `/usr/local/lib/libstorage_client.so;/usr/local/lib/libgoogle_cloud_cpp_common.so` did not fix the problem. On my machine, `ldd /usr/local/lib/libgoogle_cloud_cpp_common.so` doesn't list `libabsl_bad_variant_internal.so` either. The git tag I used to build google-cloud-cpp libs is `v1.17.0`.

Only when I explicitly include that Abseil lib into list of libraries like `/usr/local/lib/libgoogle_cloud_cpp_common.so;/usr/local/lib/libstorage_client.so;/usr/local/lib/libabsl_bad_variant_access.so` I'm able to build the tests.

Unfortunately Arrow's CMake setup is quite complicated (using variables, multiple files), and my CMake skills are only beginner-level, so I couldn't try your imported target suggestion. Maybe later I'll get more understanding. Thanks again!

coryan:
@megaserg if you even send a PR to the apache arrow project CC me on it, I may be able to make suggestions on how to improve the CMake files.

Issue #4487, opened 2020-07-08
[Flake] vcpkg Failed to take the filesystem lock


devjgm:
In the ~Windows~ Docker quickstart cmake presubmit build:

https://source.cloud.google.com/results/invocations/152e696f-0409-447d-bfe7-19c052095160/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fdocker%2Fquickstart-cmake-presubmit/log

```
[76/77] Building CXX object CMakeFiles/vcpkglib.dir/src/vcpkg/vcpkglib.cpp.o
[77/77] Linking CXX executable vcpkg
Telemetry
---------
vcpkg collects usage data in order to help us improve your experience.
The data collected by Microsoft is anonymous.
You can opt-out of telemetry by re-running the bootstrap-vcpkg script with -disableMetrics,
passing --disable-metrics to vcpkg on the command line,
or by setting the VCPKG_DISABLE_METRICS environment variable.

Read more about vcpkg telemetry at docs/about/privacy.md
There are no outdated packages.
Waiting to take filesystem lock on /h/vcpkg-quickstart/.vcpkg-root...
Failed to take the filesystem lock on /h/vcpkg-quickstart/.vcpkg-root:
    Device or resource busy
Exiting now.
2020-07-08T21:44:00Z: Docker run failed with exit status 1.
```

devjgm:
Seems to be related to https://github.com/microsoft/vcpkg/issues/12286 (filed 2 days ago) and may be fixed in https://github.com/microsoft/vcpkg/issues/12286

coryan:
https://source.cloud.google.com/results/invocations/7fcaf995-e3f7-4bc7-8fc8-107974609db5

ras0219-msft:
Cross-posted to https://github.com/microsoft/vcpkg/issues/12286:

> docker .... --volume /tmpfs/src/github/google-cloud-cpp/cmake-out/home/ci-ubuntu-18.04-quickstart-cmake:/h ....
...
Waiting to take filesystem lock on /h/vcpkg-quickstart/.vcpkg-root...
Failed to take the filesystem lock on /h/vcpkg-quickstart/.vcpkg-root:
Device or resource busy
Exiting now.

[...]

PR #4495, opened 2020-07-10
fix: use flag to prevent vcpkg lock timeout

devjgm:
Fixes https://github.com/googleapis/google-cloud-cpp/issues/4487

The flag was implemented by Microsoft in
https://github.com/microsoft/vcpkg/pull/12227, thanks to @strega-nil!

I believe we were seeing this flake not because of multiple installs
happening at once, but rather because the machines that run our docker
tests were sometimes overloaded and running very slow and thus it
sometimes took more than 1.5 sec to acquire the lock. This is just a
hypothesis.

But regardless, we're now passing `--x-wait-for-lock` to all of our
`vcpkg install` invocations to avoid these flakes.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4495)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-quickstart-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-quickstart-cmake.sh
@@ -57,7 +57,7 @@ fi
   cd "${vcpkg_dir}"
   ./bootstrap-vcpkg.sh
   ./vcpkg remove --outdated --recurse
-  ./vcpkg install google-cloud-cpp
+  ./vcpkg install --x-wait-for-lock google-cloud-cpp
 )
 # Use the new installed/ directory to create the cache.
 rm -fr "${HOME}/vcpkg-quickstart-cache"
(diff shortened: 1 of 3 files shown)


devjgm:
> The changes look good.
> 
> Is the lock used for `remove` as well as `install`? We do a remove before each?/all? installs to clean up old versions.

Good point. The flag works when passed to `vcpkg remove` also, so I think we might as well pass it there too. I made that change and am letting the test re-run.

> 
> > Reviewed 3 of 3 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4495#-:-MBt1lPd8xXg2pXYNX9k:b6iaxpe)_ status: :shipit: complete! all files reviewed, all discussions resolved

strega-nil:
FWIW, this is not necessary in classic mode (which is what i assume you're using); i decided it made the most sense to not lock for existing customers, and only lock for manifest users (since that's the plan for vcpkg moving forward)

devjgm:
> FWIW, this is not necessary in classic mode (which is what i assume you're using); i decided it made the most sense to not lock for existing customers, and only lock for manifest users (since that's the plan for vcpkg moving forward). The lock no longer exists for you, basically.

Oh. TIL about "classic" and "manifest" mode for vcpkg (https://vcpkg.readthedocs.io/en/latest/specifications/manifests/). Thanks for chiming in. I'll just drop this PR.

Thanks for fixing this issue!

PR #4502, opened 2020-07-12
fix: multiple problems under Windows+MSVC+x86

### LINKED ISSUE #4497: google::cloud::storage::ObjectWriteStream fails if > 4 GiB written
google/cloud/storage

**Describe the bug**
In 32-bit builds, uploading streams > 4 GiB fail with:

Permanent error in UploadGenericChunk: Failed to parse Content-Range header. [INVALID_ARGUMENT]

**To Reproduce**

In a 32-bit build:

Use google::cloud::storage::Client::WriteObject to get a stream (m_writer).  Write > 4 GiB to the stream.  Close the stream.

m_writer->metadata().ok() is false and m_writer->last_status().message() is as above.

**Expected behavior**
Success.

**Operating system:**
Windows 10, but not particularly relevant.

Visual Studio 2019, but not particularly relevant.

Irrelevant, underlying problem is in current version: b308a107402f28e097898e91a07295231878cdcf

**Investigation**

The Content-Range header was:

Content-Range: bytes 9135194112-9136473156/546538565

546538565 is rather obviously 9136473157 mod 2^32 so we are looking for a 64 to 32 bit truncation.  It was found in object_streambuf.cc.

**FIX**
Change std::size_t to std::uint64_t in object_streambuf.cc when calculating the upload size.

std::size_t upload_size = upload_session_->next_expected_byte() + actual_size;

_should be_

std::uint64_t upload_size = upload_session_->next_expected_byte() + actual_size;

'auto' would probably work as well; std::size_t is not well defined and is wrong.

Note that next_expected_byte() returns std::uint64_t and UploadFinalChunk() where upload_size is used expects std::uint64_t.

coryan:
Fixed a number of conversions from 64 to 32 bits, use the correct types
for file sizes vs. buffer sizes, etc. I disabled the `/WX` for this
platform because some Abseil (indirect) includes fail to build. I did
not fix all our warnings in tests and benchmarks.

Fixes #4497

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4502)


google/cloud/storage/benchmarks/benchmark_utils.h
```diff
@@ -48,6 +48,13 @@ constexpr std::int64_t kTB = 1000 * kGB;
 /// Parse a string as a byte size, with support for unit suffixes.
 std::int64_t ParseSize(std::string const& val);
 
+/**
+ * Parse a string as a byte size, with support for unit suffixes.
+ *
+ * The size must be small enough for a in-memory buffer.
```
devbww, line 54: s/a/an/

Issue #3645, opened 2020-04-02
Kokoro random timeouts in the build


coryan:
It is unclear why. We invoke bazel about 20 minutes into the build
(after fetching dependencies and such), the build reports no progress
and then seems to be killed due to the timeout:

http://sponge2/1d3eb1e7-26f5-4c23-81f9-f097cf98575e

coryan:
Here is another one:

https://source.cloud.google.com/results/invocations/b5b4a117-8d6c-4190-b84a-8a124c2c9688/targets

devjgm:
FWIW, my suspicion from looking at these logs and a number of other flakes on macos is that I think the macos servers are just way overloaded, and sometimes things run super slow and end up timing out. Not sure how best to fix this.

coryan:
This might be similar, only 60m, but also a timeout for lack of progress:

https://source.cloud.google.com/results/invocations/caff63e5-b2e2-4fb2-a079-421e996d8dc6/details

coryan:
This time the build was trying to upload the cache, maybe we can setup a timeout in the script and just not update the cache?

https://source.cloud.google.com/results/invocations/6efb9ea3-f20d-4c98-b430-a9ef67b01bb9

coryan:
https://source.cloud.google.com/results/invocations/5970321a-7773-4acd-b891-1bfc66cece96/targets

coryan:
This time early in the build:

https://source.cloud.google.com/results/invocations/c5501130-6eea-494c-981d-843043e20e96/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fmacos%2Fquickstart-bazel/log

```
[61 / 254] Compiling external/com_google_protobuf/src/google/protobuf/compiler/java/java_enum_lite.cc [for host]; 82s darwin-sandbox ... (4 actions running)


ERROR: Aborting VM command due to timeout of 3600 seconds
```

coryan:
Another one:

https://source.cloud.google.com/results/invocations/93570f7a-7698-4363-8c79-ae6e568031c6

I am following up with the Kokoro team.

coryan:
One more:

https://source.cloud.google.com/results/invocations/06490f04-0c30-4ff5-a8a3-fb941d95e842

remyabel:
Another instance I think:

[...]

PR #4505, opened 2020-07-13
ci: deflake macOS cache updates

coryan:
The last few failures for #3645 the build timed out because uploading or downloading the cache made no progress in 3600s.  With this change the upload and/or download of the cache may timeout, but that would not break the build because they run with a `|| true` at the end.

Whether the overall build will still timeout for other reason, or the 20 minutes consumed by the cache upload/download timeout proves too much, that remains to be seen. In any case, I believe this would help, if nothing else, it would give us more information as something else would fail.

Part of the work for #3645 

----
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4505)


coryan:
> What's the effect of this change?

Updated the description, will include that in the commit message too.

Issue #4507, opened 2020-07-13
[Flake] macOS build takes over 1 hour to upload (useless) artifacts


coryan:
This build usually takes a few minutes, in this case the build was stuck for over one hour in these steps:

```
[ID: 8993353] Build finished after 178 secs, exit value: 0


Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
```

Note that this is after our build script has finished, I killed the build, which shows Kokoro trying to fetch over 800MiB of artifacts:

```
rsync: connection unexpectedly closed (796110846 bytes received so far) [receiver]
[12:16:25] Collecting build artifacts from build VM
Received signal: 15, build might get interrupted.
Received signal: 15, build might get interrupted.
rsync error: received SIGINT, SIGTERM, or SIGHUP (code 20) at rsync.c(642) [generator=3.1.3]
rsync error: error in rsync protocol data stream (code 12) at io.c(226) [receiver=3.1.3]
Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
[13:16:49] Kokoro builder finished
```

https://source.cloud.google.com/results/invocations/896f01fc-aa98-41c0-a4fd-0ca3008d6c34/targets

I think we need to erase the build artifacts before exiting our builds, like we do for Linux and Windows.

coryan:
Not planning on other changes after #4508. We should close this bug by 2020-10-13 unless the flake repeats before then.

coryan:
No repeats in 90d, the fixes in #4508 seem to have minimized the problem to the point of irrelevancy.

PR #4508, opened 2020-07-13
ci(macos): remove unused artifacts before exit

coryan:
By default Kokoro uploads the data from KOKORO_ARTIFACTS_DIR after
the build script exits. That can be a substantial amount of data, and
we do not use it. It is faster to just remove it.

Part of the work for #4507

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4508)

### DIFF
--- a/ci/kokoro/macos/build.sh
+++ b/ci/kokoro/macos/build.sh
@@ -131,7 +131,20 @@ else
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


coryan:
It seems like this saves a couple of minutes at the end of the macOS builds. So this is useful, even if it is useless to reduce the incidence of the flake (which I think it is not the case).

Issue #4493, opened 2020-07-09
Support startOffset/endOffset is Object: list


coryan:
New parameters for the `Object: list` method (aka `storage::Client::ListObjects()`).

- `startOffset` filters out objects before a given name.
- `endOffset` filters out objects at or after a given name.

More details at:

https://cloud.google.com/storage/docs/json_api/v1/objects/list

vnvo2409:
I will work on this feature

coryan:
Thanks!

PR #4518, opened 2020-07-14
feat(storage): Add startOffset and endOffset

vnvo2409:
Part of the work for #4493

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4518)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -988,7 +988,8 @@ class Client {
    * @param options a list of optional query parameters and/or request headers.
    *     Valid types for this operation include
    *     `IfMetagenerationMatch`, `IfMetagenerationNotMatch`, `UserProject`,
-   *     `Projection`, `Prefix`, `Delimiter`, and `Versions`.
+   *     `Projection`, `Prefix`, `Delimiter`, `StartOffset`, `EndOffset`, and
+   *     `Versions`.
    *
    * @par Idempotency
    * This is a read-only operation and is always idempotent.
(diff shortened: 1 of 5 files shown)


vnvo2409:
Failed because of test bench, I will check it

vnvo2409:
@coryan 
One question: How to set up an integration test with test bench and bazel or run only some tests using cmake ?
Thank you

coryan:
> One question: How to set up an integration test with test bench and bazel or run only some tests using cmake ?

There is no "easy" way to do that, but this works for me:

Create a python virtual environment to install the dependencies for the testbench:

```
cd $HOME
python3 -m virtualenv ~/v-dev
source ~/v-dev/bin/activate
pip3 install flask==1.1.1 Werkzeug==1.0.0 httpbin==0.7.0     gevent==1.4.0 gunicorn==19.10.0 crc32c==2.0
```

In that shell start the testbench on a fixed port (I am using 8585 here) with name for the test bucket:

```
cd google-cloud-cpp
env GOOGLE_CLOUD_CPP_STORAGE_TEST_BUCKET_NAME=bucket gunicorn --bind "0.0.0.0:8585"     --worker-class gevent     --access-logfile -     --pythonpath "${PWD}/google/cloud/storage/testbench" --reload    testbench:application
```

On a separate shell run the integration test you want, typically, the last one I ran was:

```
GOOGLE_CLOUD_PROJECT=test GOOGLE_CLOUD_CPP_STORAGE_TEST_BUCKET_NAME=bucket CLOUD_STORAGE_TESTBENCH_ENDPOINT=http://localhost:8585 \
    ./cmake-build-debug/google/cloud/storage/tests/object_resumable_write_integration_test  --gtest_filter=ObjectResumableWriteIntegrationTest.WriteNotChunked
```

In that example `./cmake-build-debug` is my binary directory for CMake. You can use Bazel too, but remember that Bazel does not pass any environment variables to tests.  You need to pass each environment variable using the `--test_env` option.

Each test requires a slightly different set of environment variables, I am sure you can figure out that part.

vnvo2409:
There is one indentation error in python which cause some errors related `prefixes`. But in the last PR, there were too many tests run at a time and they all passed so I don't know about it. I think it will increase the coverage rate in `ListObjectsAndPrefixes` sample too.

PR #4522, opened 2020-07-14
doc: update region tags for bigtable quickstart

cshaff0524:
Re-add region tags that were removed when this page was refactored and moved. Region tags are required to pull the quickstart code in to the client libraries overview page in the Cloud Bigtable docs.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4522)


cshaff0524:
Yes, that's fine! And thank you.

PR #4525, opened 2020-07-14
doc(bigtable): restore quickstart's region tags

coryan:
Thanks to @cshaff0524 for the original PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4525)

### DIFF
--- a/google/cloud/bigtable/quickstart/quickstart.cc
+++ b/google/cloud/bigtable/quickstart/quickstart.cc
@@ -59,5 +59,4 @@ int main(int argc, char* argv[]) try {
   std::cerr << "Standard C++ exception raised: " << ex.what() << "\n";
   return 1;
 }
-
 // [END bigtable_quickstart]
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/quickstart/quickstart.cc
```diff
@@ -57,3 +59,5 @@ int main(int argc, char* argv[]) try {
   std::cerr << "Standard C++ exception raised: " << ex.what() << "\n";
   return 1;
 }
+
```
devbww, line 62: Could we remove these blank lines (which, it seems, would eliminate the `README.md` edit)?

google/cloud/bigtable/quickstart/quickstart.cc
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// [START bigtable_quickstart]
+
```
devbww, line 16: Would this go away too then?  :-)

Issue #3896, opened 2020-02-12
Create skeleton for `pubsub::Publisher`


coryan:
This is in flux, but the proposed interface for `pubsub::Publisher` is:

```C++
class Publisher { public:
  // Create a new publisher with the given policy.
  Publisher(
      Topic topic,
      std::shared_ptr<PublisherConnection> connection = MakePublisherConnection(),
      BatchingPolicy policy = DefaultBatchingPolicy());

  // Move only, create additional publishers for the same connection to use
  // from multiple threads and/or have different batching policies.
  Publisher(Publisher const&) = delete;
  Publisher& operator=(Publisher const&) = delete;

  // Publish a new message, returning a future, satisfied when the message
  // is sent. The returned value is the message identified returned by Cloud
  // Pub/Sub
  future<StatusOr<std::string>> Publish(Message message);

  // Flush all existing batches, returning a future satisfied when done.
  future<void> Shutdown() &&;
};
```

coryan:
Oops, I implemented `PublisherConnection` but still need `Publisher`.

Issue #3895, opened 2020-02-20
Create skeleton for `pubsub::Subscriber`


coryan:

ziliangpeng:
Hi!

Thanks for implementing the pubsub client in cpp.

Curious what the timeline is for the pubsub library to be ready to use?

coryan:
Cannot commit to a date right now.  I expect we will create milestones and specific plans in the next few weeks.

PR #4538, opened 2020-07-15
feat(pubsub): skeleton stubs to push/pull messages

coryan:
Implement the blocking, synchronous, non-streaming RPCs to publish and
pull messages. That will let us implement skeletons for
`pubsub::Publisher` and `pubsub::Subscriber`.

Part of the work for #3895 and #3896 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4538)

### DIFF
--- a/google/cloud/pubsub/integration_tests/message_integration_test.cc
+++ b/google/cloud/pubsub/integration_tests/message_integration_test.cc
@@ -1,0 +1,150 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


google/cloud/pubsub/integration_tests/message_integration_test.cc
```diff
@@ -0,0 +1,150 @@
...
+    ASSERT_STATUS_OK(messages);
+
+    for (auto const& m : *messages) {
+      SCOPED_TRACE("Search for message " + m.DebugString());
+      auto i = std::find_if(
+          ids->begin(), ids->end(),
+          [&m](std::string const& x) { return m.message().message_id() == x; });
+      EXPECT_STATUS_OK(ack(m.ack_id()));
+      EXPECT_NE(i, ids->end());
+      ids->erase(i);
```
devbww, line 138: `erase(end)` is UB, so we should probably guard this (assuming `ASSERT_NE` is overkill).

coryan: Doh!  assert might be overkill, we probably want to ack all the other messages anyway.

PR #4541, opened 2020-07-16
feat(pubsub): a wrapper for PubsubMessage

coryan:
Introduce `pubsub::Message`, a wrapper for the `PubsubMessage` protobuf.
This wrapper provides more native C++ types to access the message
properties, and a number of factory functions to create messages.

Part of the work for #3895 and #3896

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4541)


google/cloud/pubsub/message_test.cc
```diff
@@ -0,0 +1,177 @@
...
+  EXPECT_EQ("original", m0.data());
+  EXPECT_THAT(m0.attributes(),
+              UnorderedElementsAre(std::make_pair("k1", "v1"),
+                                   std::make_pair("k2", "v2")));
+}
+
+TEST(Message, SetAttributesVectorStdPair) {
+  std::vector<std::pair<std::string, std::string>> const attributes(
+      {{"k1", "v1"}, {"k2", "v2"}});
+  auto const m0 = Message::SetAttributes(
```
devjgm, line 101: Did you consider a fluent builder rather than the static mutator factories? I could imagine this looking like

```cc
auto const m0 = pubsub::MessageBuilder().set_data("original").set_attributes({{"k0, "v0"}}).Build();
```

WDYT?

coryan: Done.

google/cloud/pubsub/message.h
```diff
@@ -0,0 +1,194 @@
...
+    }
+    return r;
+  }
+  //@}
+
+  /// @name Copy and move
+  //@{
+  Message(Message const&) = default;
+  Message& operator=(Message const&) = default;
+  Message(Message&&) = default;
```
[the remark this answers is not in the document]
coryan: Well, quasi-immutable, the only mutation was "move".  I mean, sure, you can do a move and continue to use the object, but that is just bad manners.
Anyway, no longer relevant, since I moved this to use a builder as you suggested.

google/cloud/pubsub/message.h
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * The C++ representation for a Cloud Pub/Sub messages.
+ *
+ * Cloud Pub/Sub applications communicate to each other using messages. Note
+ * that messages must provide at least some data or some attributes. The `From*`
+ * factory functions are designed to encourage applications to always create
+ * valid messages.
+ *
+ * A #Message is immutable other than through move copy or assignment. Once a
```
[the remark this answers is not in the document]
coryan: Fixed, I think (see below).

google/cloud/pubsub/message.h
```diff
@@ -0,0 +1,212 @@
...
+
+namespace pubsub {
+inline namespace GOOGLE_CLOUD_CPP_PUBSUB_NS {
+class MessageBuilder;
+
+/**
+ * The C++ representation for a Cloud Pub/Sub messages.
+ *
+ * Cloud Pub/Sub applications communicate to each other using messages. Note
+ * that messages must provide at least some data or some attributes. The `From*`
```
devjgm, line 48: I think these `From*` factories are gone now.

google/cloud/pubsub/message.h
```diff
@@ -0,0 +1,212 @@
...
+
+  /// Create a message with the data in @p data
+  MessageBuilder& SetData(std::string data) & {
+    proto_.set_data(std::move(data));
+    return *this;
+  }
+
+  /// Create a message with the data in @p data
+  MessageBuilder&& SetData(std::string data) && {
+    (*this).SetData(std::move(data));
```
devjgm, line 124: `this->`?

coryan:
PTAL

google/cloud/pubsub/message.h
```diff
@@ -0,0 +1,212 @@
...
+  template <typename Iterator>
+  MessageBuilder&& SetAttributes(Iterator begin, Iterator end) && {
+    SetAttributes(std::move(begin), std::move(end));
+    return std::move(*this);
+  }
+
+  /// Create a message with the attributes in @p v
+  MessageBuilder& SetAttributes(
+      // NOLINTNEXTLINE(performance-unnecessary-value-param)
+      std::vector<std::pair<std::string, std::string>> v) & {
```
devjgm, line 152: (no change necessary) I wonder if cases like this would work better if the argument was `absl::Span<std::pair<std::string, std::string>>`. I conceptually understand Span, but I don't have a lot of hands-on experience using it, so I'm not sure whether it's intended for cases like this not.

google/cloud/pubsub/message.h
coryan: Fixed.

google/cloud/pubsub/message.h
coryan: Removed.

google/cloud/pubsub/message.h
coryan: Filed #4548 .

PR #4542, opened 2020-07-16
doc: fix Ubuntu code name

Mizux:
Fixing ubuntu code name
ref: https://releases.ubuntu.com/20.04/

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4542)

### DIFF
--- a/ci/generate-markdown/generate-packaging.sh
+++ b/ci/generate-markdown/generate-packaging.sh
@@ -202,9 +202,9 @@ these dependencies.
 - [Fedora 31](#fedora-31)
 - [openSUSE (Tumbleweed)](#opensuse-tumbleweed)
 - [openSUSE (Leap)](#opensuse-leap)
-- [Ubuntu (20.04 - Bionic Beaver)](#ubuntu-2004---focal-fossa)
-- [Ubuntu (18.04 - Bionic Beaver)](#ubuntu-1804---bionic-beaver)
-- [Ubuntu (16.04 - Xenial Xerus)](#ubuntu-1604---xenial-xerus)
+- [Ubuntu (20.04 LTS - Focal Fossa)](#ubuntu-2004-lts---focal-fossa)
+- [Ubuntu (18.04 LTS - Bionic Beaver)](#ubuntu-1804-lts---bionic-beaver)
+- [Ubuntu (16.04 LTS - Xenial Xerus)](#ubuntu-1604-lts---xenial-xerus)
 - [Debian (Buster)](#debian-buster)
 - [Debian (Stretch)](#debian-stretch)
 - [CentOS 8](#centos-8)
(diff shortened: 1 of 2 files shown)


Mizux:
I've just amended the commit to follow the convention...

coryan:
Thank you very much for the PR.  This file is automatically generated by `ci/test-markdown/generate-packaging.sh`, and the build will break if its output does not match `packaging.md`.  Do you mind also updating that file?  I completely understand if that feels like too much work, I will send a PR instead if you prefer.

Mizux:
ok, on my way to do it.

Mizux:
did you mean `ci/generate-markdown/generate-packaging.sh` instead ?

coryan:
Doh, yes.  The comments at the top of `packaging.md` are obviously in need of an update.

Mizux:
I'm note sure concerning the tag since I add LTS -> how can I test link are not broken ?

ci/generate-markdown/generate-packaging.sh
```diff
@@ -202,9 +202,9 @@ these dependencies.
 - [Fedora 31](#fedora-31)
 - [openSUSE (Tumbleweed)](#opensuse-tumbleweed)
 - [openSUSE (Leap)](#opensuse-leap)
-- [Ubuntu (20.04 - Bionic Beaver)](#ubuntu-2004---focal-fossa)
-- [Ubuntu (18.04 - Bionic Beaver)](#ubuntu-1804---bionic-beaver)
-- [Ubuntu (16.04 - Xenial Xerus)](#ubuntu-1604---xenial-xerus)
+- [Ubuntu (20.04 LTS - Focal Fossa)](#ubuntu-2004---focal-fossa)
```
Mizux, line 205: not sure `#ubuntu-2004---focal-fossa` is still working since I add the LTS (which is part of the official name IIRC)

Mizux: Just use chrome extension markdown viewer to retrieve to the potential new tag name

coryan: You can see the packaging.md file in your own branch, and follow the links there:

https://github.com/Mizux/google-cloud-cpp/blob/patch-1/doc/packaging.md#table-of-contents

and:

https://github.com/Mizux/google-cloud-cpp/blob/patch-1/doc/packaging.md#ubuntu-2004-lts---focal-fossa

Mizux:
LGTM IMHO -> PTAL

ci/generate-markdown/generate-packaging.sh
Mizux: tested !

Mizux:
my bad just forgot the s/test-markdown/generate-markdown/ in the shell script
should be fixed now...

PR #4552, opened 2020-07-17
feat(pubsub): a skeleton for PublisherConnection

### LINKED ISSUE #3896: Create skeleton for `pubsub::Publisher`

coryan:
Implements a skeleton for `pubsub::PublisherConnection()` with a single
`Publish()` member function. The function returns a
`future<StatusOr<std::string>>` because eventually needs to be
implemented by an asynchronous RPC, but for now we just return an
immediately satisfied future and use the blocking RPC.

Introduce a `pubsub_testing::MockPublisherStub` so we can write unit
tests for this new class. Also wrote the first unit test.

Fixes #3896

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4552)


google/cloud/pubsub/publisher_connection_test.cc
```diff
@@ -0,0 +1,99 @@
...
+            return make_status_or(response);
+          });
+
+  auto publisher = pubsub_internal::MakePublisherConnection(mock);
+  auto response =
+      publisher
+          ->Publish({topic.FullName(),
+                     MessageBuilder{}.SetData("test-data-0").Build()})
+          .get();
+  EXPECT_EQ(StatusCode::kUnknown, response.status().code());
```
devbww, line 72: Should we look at the message too to make it clearer just what error we're expecting?

PR #4558, opened 2020-07-17
feat(pubsub): a skeleton for SubscriberConnection

### LINKED ISSUE #3895: Create skeleton for `pubsub::Subscriber`

coryan:
Implement an initial version for `pubsub::SubscriberConnection`, leaving
a number of TODO entries for (near) future improvements. Add the
necessary mocks to write unit tests, to do so this change also
introduces the `pubsub_client_mocks` library with user-facing mocks (as
opposed to the mocks that we only expect we would need in our own
tests). Made sure the tests are all running.

Part of the work for #3895

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4558)


google/cloud/pubsub/subscriber_connection_test.cc
```diff
@@ -0,0 +1,95 @@
...
+            EXPECT_EQ(subscription.FullName(), request.subscription());
+            EXPECT_FALSE(request.ack_ids().empty());
+            for (auto& id : request.ack_ids()) {
+              EXPECT_EQ("test-ack-id-0", id);
+            }
+            return Status{};
+          });
+
+  auto subscriber = pubsub_internal::MakeSubscriberConnection(mock);
+  std::atomic_flag received_one{false};
```
devbww, line 57: `#include <atomic>`

coryan: Done, thanks

PR #4591, opened 2020-07-17
feat(storage): disable MD5Hash by default

### LINKED ISSUE #3395: Consider disabling MD5 hashes by default
The MD5 hash seems to be the most CPU intensive part of downloading an object. Considering that some objects do not have a MD5, and we are already computing the CRC32C checksum maybe we should consider disabling this feature by default.

vnvo2409:
Fixes #3395

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4591)

### DIFF
--- a/google/cloud/storage/internal/curl_client.cc
+++ b/google/cloud/storage/internal/curl_client.cc
@@ -510,9 +510,10 @@ StatusOr<ObjectMetadata> CurlClient::InsertObjectMedia(
   }
 
   // If the application has set an explicit hash value we need to use multipart
-  // uploads. We check `DisableCrc32cChecksum` only because `DisableMD5Hash`
-  // always has a value.
-  if (!request.HasOption<DisableCrc32cChecksum>()) {
+  // uploads. `DisableMD5Hash` and `DisableCrc32cChecksum` should not be
+  // dependent on each other.
+  if (!request.GetOption<DisableMD5Hash>().value() ||
+      !request.GetOption<DisableCrc32cChecksum>().value_or(false)) {
     return InsertObjectMediaMultipart(request);
   }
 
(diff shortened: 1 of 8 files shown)


google/cloud/storage/internal/curl_client.cc
```diff
@@ -510,9 +510,9 @@ StatusOr<ObjectMetadata> CurlClient::InsertObjectMedia(
...
 
   // If the application has set an explicit hash value we need to use multipart
-  // uploads.
-  if (!request.HasOption<DisableMD5Hash>() &&
-      !request.HasOption<DisableCrc32cChecksum>()) {
+  // uploads. We check `DisableCrc32cChecksum` only because `DisableMD5Hash`
+  // always has a value.
+  if (!request.HasOption<DisableCrc32cChecksum>()) {
     return InsertObjectMediaMultipart(request);
   }
```
vnvo2409, line 517: I would like to ask something here. When `DisableMD5Hash` is enabled by default, what is the condition to use `InsertObjectMediaMultipart` ?

vnvo2409: The CI will failed because of this condition. So please do not run the CI now. Thank you

google/cloud/storage/internal/curl_client.cc
```diff
@@ -510,9 +510,10 @@ StatusOr<ObjectMetadata> CurlClient::InsertObjectMedia(
...
 
   // If the application has set an explicit hash value we need to use multipart
-  // uploads.
-  if (!request.HasOption<DisableMD5Hash>() &&
-      !request.HasOption<DisableCrc32cChecksum>()) {
+  // uploads. `DisableMD5Hash` and `DisableCrc32cChecksum` should not be
+  // dependent on each other.
+  if (!request.GetOption<DisableMD5Hash>().value() ||
+      !request.GetOption<DisableCrc32cChecksum>().value_or(false)) {
     return InsertObjectMediaMultipart(request);
```
vnvo2409, line 517: I use `||` instead of `&&` because if one of there 2 options (`MD5` or `Crc32`) is disabled, the other one should still be computed as usual to prevent data corruptions.

vnvo2409:
> Why is that not true for CRC32C? Maybe open a bug to fix later?
It is true for `Crc32c` too but since it isn't related to `Md5`, I think I will make another PR for this bug if there are somethings to do with `Crc32c` later.

Issue #4592, opened 2020-07-18
createdBefore has `date` type, not `datetime`


coryan:
From the documentation:

```
A date in RFC 3339 format with only the date part (for instance, "2013-01-15"). This condition is satisfied when an object is created before midnight of the specified date in UTC.
```

https://cloud.google.com/storage/docs/json_api/v1/buckets

We are using the full RFC3339 format.

Issue #4432, opened 2020-06-25
Storage: support customTime metadata and custom timestamp based OLM


jkwlui:
Request to support a new customTime metadata and associated custom timestamp based OLM conditions:

```cpp
std::chrono::time_point tenDays = std::chrono::system_clock::now() + 
                                      std::chrono::days{10};

gcs::BucketLifecycle bucket_lifecycle_rules = gcs::BucketLifecycle{
    {gcs::LifecycleRule(
         gcs::LifecycleRule::ConditionConjunction(
             gcs::LifecycleRule::DaysSinceCustomTime(30),
             gcs::LifecycleRule::CustomTimeBefore(tenDays)),
          gcs::LifecycleRule::Delete())}};

StatusOr<gcs::BucketMetadata> updated_metadata = client.PatchBucket(
    bucket_name,
      gcs::BucketMetadataPatchBuilder().SetLifecycle(bucket_lifecycle_rules));

// New metadata field:
StatusOr<gcs::ObjectMetadata> object_metadata =
    client.GetObjectMetadata(bucket_name, object_name);

gcs::ObjectMetadata desired = *object_metadata;
desired.mutable_metadata().emplace("customTime", "2020-01-01T00:00:00Z");

StatusOr<gcs::ObjectMetadata> updated =
    client.UpdateObject(bucket_name, object_name, desired,
                        gcs::Generation(object_metadata->generation()));
```

coryan:
Notwithstanding the original description, the OLM field needs to have `date` precision.  That is, the code will read:

```cc
gcs::BucketLifecycle bucket_lifecycle_rules = gcs::BucketLifecycle{
    {gcs::LifecycleRule(
         gcs::LifecycleRule::ConditionConjunction(
             gcs::LifecycleRule::DaysSinceCustomTime(30),
             gcs::LifecycleRule::CustomTimeBefore(absl::CivilDay(2020, 7, 29)),
          gcs::LifecycleRule::Delete())}};
```

/cc: @frankyn

Issue #4275, opened 2020-05-26
Support daysSinceNoncurrentTime and noncurrentTimeBefore OLM options


jkwlui:
Basically this change entails

- [x] Adding two fields to `storage::Lifecycle::Condition`:

```C++
struct storage::LifecycleRule::Condition {
 ....
  optional<std::int32_t> days_since_noncurrent_time;
  optional<std::chrono::system_clock::time_point> noncurrent_time_before;
};
```

- [x] Adding helper functions to create conditions with one of these fields set
- [x] Change the existing code to merge these fields as part of `*Conjunction`

frankyn:
Reference documentation

[One-pager](https://docs.google.com/document/d/1sMfcVPaHedPq9OlNlYW4Oy20_oi_jaq5Kcd8WmdJRxM/edit)
[Language Designs](https://docs.google.com/document/d/16nagQmGIhemHlSR8k3IWoIewA9OxClRRJ3knBEKMd3Q/edit#heading=h.90jzlysmds9t)

coryan:
I closed the PR for now (but the branch still exists). When we are ready to release this let me know and I will create a new PR.

coryan:
Marking as currently blocked on an external issue.

coryan:
Notwithstanding the original description, the OLM field needs to have `date` precision.  That is, the code will read:

```cc
struct storage::LifecycleRule::Condition {
 ....
  optional<std::int32_t> days_since_noncurrent_time;
  optional<absl::CivilDay> noncurrent_time_before;
};
```

/cc: @frankyn

PR #4594, opened 2020-07-18
refactor!: move spanner::Date to common library

coryan:
**BREAKING CHANGE**: this change removes `operator<<(std::ostream&, Date const&)` from
the `google::cloud::spanner` namespace. We do not expect many applications to refer to this
symbol explicitly, and the fix is relatively easy too. Nevertheless this changes the API. The 
`google::cloud::spanner::Date` class is now an alias to preserve backwards compatibility in the
common case.

This is part of the work for #4592, #4432 and #4275

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4594)


coryan:
I am going to update the API baseline, obviously g/c/spanner losses some functions.  The `using Date = google::cloud::Date` declaration keeps backwards compatibility for the API *except* for `google::cloud::spanner::operator<<(std::ostream&, google::cloud::spanner::Date)`, referencing that using explicitly would no longer work.

https://source.cloud.google.com/results/invocations/03076ede-fd03-4b84-9632-fc358d0d0c50

I am going to mark this change as a breaking change.

PR #4597, opened 2020-07-20
fix(storage)!: incorrect type in OLM condition

### LINKED ISSUE #4592: createdBefore has `date` type, not `datetime`

coryan:
**BREAKING CHANGE**: changed the `created_before` field in
`LifecycleRuleCondition` from a time point to a date. The service
rejects timestamps, it only accepts dates in YYYY-mm-dd format. We
considered trying to keep the C++ type the same and fixing the
conversions to the underlying protocol, but (a) this was confusing to
any users, and (b) we do not expect anybody was using this field because
it simply did not work before.

Fixes #4592

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4597)


google/cloud/storage/lifecycle_rule.cc
```diff
@@ -62,6 +63,17 @@ LifecycleRuleAction LifecycleRule::SetStorageClass(std::string storage_class) {
   return LifecycleRuleAction{"SetStorageClass", std::move(storage_class)};
 }
 
+LifecycleRuleCondition LifecycleRule::CreatedBefore(std::string const& date) {
```
devjgm, line 66: Q: Do you think it's necessary to have this `CreatedBefore` overload that takes a string? I know we have it today, and users may be passing full time strings, which would fail to parse as a date here and it would throw/crash. This overload is avoiding a build break, but instead there might be a runtime exception/crash.

Since this PR is already a breaking change, I'm wondering if maybe we want to get rid of this overload taking a string and require the caller to pass us a `CivilDay`, which will push any string-to-date error handling onto the caller.

Just a thought. WDYT?

coryan: I was about to push the commit to remove these.

devjgm: I was waiting for this removal, but let me know if you'd like approval before that, or want to do that in a separate PR.

coryan: Sorry for the delay, I had to rebase and fix the API baseline.

coryan:
PTAL

PR #4603, opened 2020-07-21
feat(pubsub): a skeleton for the Publisher client

### LINKED ISSUE #3896: Create skeleton for `pubsub::Publisher`

### LINKED ISSUE #3901: Create mock class for PublisherConnection

coryan:
This PR introduces a very naive version of the `pubsub::Publisher`
client. It also introduces a (somewhat trivial) unit test for the new
class, and a mock for `pubsub::PublisherConnection`. It also introduces
the configuration options for batching and ordering keys, though these
are not used (yet).

This is all going to be more interesting once we implement some amount
of batching.

This really fixes #3896 and fixes #3901 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4603)


google/cloud/pubsub/publisher.h
```diff
@@ -0,0 +1,131 @@
...
+ * queue in `pubsub::ConnectionOptions::DisableBackgroundThreads()`, and (c)
+ * attaching any number of threads to the completion queue.
+ *
+ * TODO(#4586) - add an example on how to do this.
+ *
+ * @par Asynchronous Functions
+ *
+ * Some of the member functions in this class return a `future<T>` (or
+ * `future<StatusOr<T>>`) object.  Readers are probably familiar with
+ * [`std::future<T>`][std-future-link], our version adds a `.then()` function to
```
devbww, line 69: s/, our/. Our/

google/cloud/pubsub/publisher.h
```diff
@@ -0,0 +1,131 @@
...
+  Publisher& operator=(Publisher const&) = default;
+  Publisher(Publisher&&) noexcept = default;
+  Publisher& operator=(Publisher&&) noexcept = default;
+  //@}
+
+  /**
+   * Publishes a message to this publisher's topic
+   *
+   * @par
+   * Note that the message maybe be batched based on the Publisher's
```
devbww, line 113: based on the Publisher's .......?

google/cloud/pubsub/publisher_option_test.cc
```diff
@@ -0,0 +1,56 @@
...
+
+  auto const b = BatchingConfig{}
+                     .set_maximum_hold_time(std::chrono::seconds(12))
+                     .set_maximum_batch_bytes(123)
+                     .set_maximum_message_count(10);
+  EXPECT_EQ(10, b.maximum_message_count());
+  EXPECT_EQ(123, b.maximum_batch_bytes());
+  auto const expected = std::chrono::duration_cast<std::chrono::microseconds>(
+      std::chrono::seconds(12));
+  EXPECT_EQ(expected.count(), b.maximum_hold_time().count());
```
devbww, line 36: Should we be equating counts as opposed to durations?

coryan: Done, note that the error message is pretty much useless in this case. googletest does not know how to format std::chrono::duration types.

google/cloud/pubsub/publisher.h
coryan: Fixed

google/cloud/pubsub/publisher.h
coryan: Fixed

coryan:
PTAL

google/cloud/pubsub/publisher_options.h
```diff
@@ -0,0 +1,92 @@
...
+ public:
+  PublisherOptions() = default;
+
+  BatchingConfig const& batching_config() const { return batching_config_; }
+  PublisherOptions& set_batching_config(BatchingConfig v) {
+    batching_config_ = std::move(v);
+    return *this;
+  }
+
+  bool message_ordering() const { return message_ordering_; }
```
devjgm, line 72: Q: Why the separate enable/disable message ordering functions rather than `set_message_ordering(bool)`? With the accessor named `bool message_ordering()`, we're already asking users to know that `true` means enabled and `false` means disabled, so providing a normal setter seems like it would be reasonable. Just a thought. I'll defer to your preference.

coryan: > With the accessor named `bool message_ordering()`, we're already asking users to know that `true` means enabled and `false` means disabled

Punting to #4609

PR #4608, opened 2020-07-21
feat(pubsub): use asynchronous RPC for Publish

### LINKED ISSUE #4604: Use asynchronous RPC to implement pubsub::PublisherConnection::Publish
What is says in the description, we are current using a blocking API, as we add more
functionality (mainly batching) it will be important to test them with the asynchronous
implementation.

coryan:
Change `pubsub_internal::PublisherStub` to use an asynchronous RPC for
`Publish()`, adjust the unit tests to use the new functions. Change the
default `pubsub::PublisherConnection` to use the configuration in
`pubsub::ConnectionOptions` to either use a completion queue provided by
the application (which must be serviced by one or more threads also provided
by the application), or to create its own completion queue and a (at the moment)
single background thread.

Fixes #4604

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4608)

### DIFF
--- a/google/cloud/pubsub/internal/publisher_stub.cc
+++ b/google/cloud/pubsub/internal/publisher_stub.cc
@@ -62,13 +62,17 @@ class DefaultPublisherStub : public PublisherStub {
     return {};
   }
 
-  StatusOr<google::pubsub::v1::PublishResponse> Publish(
-      grpc::ClientContext& context,
+  future<StatusOr<google::pubsub::v1::PublishResponse>> AsyncPublish(
+      google::cloud::CompletionQueue& cq,
+      std::unique_ptr<grpc::ClientContext> context,
       google::pubsub::v1::PublishRequest const& request) override {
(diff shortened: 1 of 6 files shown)


coryan:
This will conflict with #4603, let's merge that first.

coryan:
> > Change the default pubsub::PublisherConnection to create a thread pool
> 
> I don't think I saw anything there I would characterize as a thread pool.

Good point, I changed the description.

> > The application can control this thread pool using pubsub::ConnectionOptions.
> 
> For example, here we're saying the user can supply a `google::cloud::BackgroundThreads` object, yes?

It is even simpler, the application provides a `g::c::CompletionQueue` and we assume they have provided everything else:

https://github.com/googleapis/google-cloud-cpp/blob/f5a2001ceb6c0778f60000e209b917bf67187478/google/cloud/connection_options.h#L199-L210


PTAL.

Issue #4389, opened 2020-06-11
Make sure time_point -> Timestamp proto conversion handles corner cases


devjgm:
I'm curious: Does this code work with a time_point value of -1ns? According to [the docs](https://github.com/protocolbuffers/protobuf/blob/f1ce8663ac88df54cf212d29ce5123b69203b135/src/google/protobuf/timestamp.proto#L133) I think the answer should look something like:

```cc
auto p = ChronoTimepointToProtoTimestamp(epoch - std::chrono::nanoseconds(1));
EXPECT_EQ(-1, p.seconds());
EXPECT_EQ(999999999, p.nanos());
```

If not, would you please file an issue so that we can follow up later and make sure this works?

_Originally posted by @devjgm in https://github.com/googleapis/google-cloud-cpp/pull/4376/files_

PR #4611, opened 2020-07-21
fix: timestamp proto encoding works before the epoch and with extreme values

devjgm:
Fixes: #4389

Before this PR, encoding a time before the Unix epoch would produce an invalid proto due to likely having a negative nanos field. Additionally extreme values could produce invalid protos beyond the supported range.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4611)


google/cloud/internal/time_utils_test.cc
```diff
@@ -24,46 +24,115 @@ inline namespace GOOGLE_CLOUD_CPP_NS {
...
+  auto const contemporary =
+      TimeFromString("2020-07-21T11:33:03.123456789Z").value();
+
+  struct {
+    absl::Time t;
+    google::protobuf::Timestamp expected;
+  } round_trip_cases[] = {
+      {min_time, min_proto},
+      {epoch - sec(1), MakeProto(-1, 0)},
+      {epoch - sec(1), MakeProto(-1, 0)},
```
devbww, line 71: Duplication?

google/cloud/internal/time_utils_test.cc
```diff
@@ -24,46 +24,115 @@ inline namespace GOOGLE_CLOUD_CPP_NS {
...
+  // nicely with minimal wrapping.
+  // Note: we use microseconds rather than nanoseconds here because we test on
+  // some platforms where the system clock only has microsecond resolution.
+  auto const epoch = std::chrono::system_clock::from_time_t(0);
+  auto const sec = [](std::int64_t n) { return std::chrono::seconds(n); };
+  auto const us = [](std::int64_t n) { return std::chrono::microseconds(n); };
+  auto const contemporary = absl::ToChronoTime(
+      TimeFromString("2020-07-21T11:33:03.123456000Z").value());
+  // Note: we do not test the proto min/max values here because we don't know
+  // if chrono's system_clock can represent that value (it likely can't).
```
devbww, line 112: FWIW, it looks like we also assume `time_point` has microsecond resolution.

google/cloud/internal/time_utils_test.cc
devjgm: Oops. You're right. Fixed.

google/cloud/internal/time_utils_test.cc
devjgm: Yes, that's what I meant in this comment. Updated comment to make it clearer.

devbww: I'd say the range and resolution capabilities are different beasts, but OK.

coryan: FWIW, you can probably static assert whether the `std::chrono` thing has at least microsecond resolution.

```cc
  bool system_clock_has_micros = std::ratio_greater_equal<
      std::micro, std::chrono::system_clock::duration::period>::value;
```

and if you can do that, then you can write a test at full resolution when available, and no-op when it is not...

Totally optional, I do not think it matters.

devjgm: @devbww I may not be following exactly what you're suggesting. If you have wording you'd like me to paste in, feel free to let me know. We do test on some platforms where the system_clock has microsecond resolution, which I mentioned in the comment on line 104, so that's why the test uses microsecond resolution. However, we also test on platforms w/ nanosecond resolution, and on those platforms, they can't represent the full proto min/max range, which is why I don't have a test for that case.

@coryan 
> Totally optional, I do not think it matters.

Thanks. I think that would add a bit of complexity for little additional value in our tests, so I think I'll skip it in this PR.

Issue #4581, opened 2020-07-17
Automatically batch messages into a single RPC


coryan:
Applications may want to setup batching for their message publications, we need to implement:

- [x] A mechanism to control the maximum delay for the messages in the batch
- [x] A mechanism to control the maximum number of messages (or bytes?) in the batch
- [x] A mechanism to force a flush.
- [ ] Separate batches per ordering key.

remyabel:
There are leftover TODOs here:

https://github.com/googleapis/google-cloud-cpp/blob/4307ca4086f42012e8ee208891162ccecb1694a3/google/cloud/pubsub/publisher.cc#L22

PR #4616, opened 2020-07-22
feat(pubsub): batch messages in Publish()

coryan:
With this PR messages are batched in the `PublisherConnection()` layer
before making a `AsyncPublish()` call.  The application can request that
messages are flushed based on message count, total data size, or time.

Each future returned by `Publish()` is satisfied as its own work item in
the completion queue, effectively this means the messages are sent in
order, but the responses are handled in an unspecified order.

Part of the work for #4581 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4616)


coryan:
Rebased and resolved conflicts

Issue #4404, opened 2020-06-16
Implement `DELETE` for to cancel resumable uploads


coryan:
Resumable uploads can be canceled midstream:

https://cloud.google.com/storage/docs/performing-resumable-uploads#cancel-upload

but there is no API to do so. This issue is a request to implement such an API. Note that stale resumable uploads expire after 7d.

vnvo2409:
I will work on this feature.
I think there will be a function `CancelResumableUpload(const std::string& uploadID)`.
What do you think ?

coryan:
> I think there will be a function CancelResumableUpload(const std::string& uploadID).

I believe that sounds about right.

A few observations:

- You will need to create a new `DeleteResumableUploadRequest` and some unit tests for it (that is the name I would use, but if you think of a better one go for it).
- You will need a new function to all the `RawClient` classes.
- This is an idempotent operation, we should retry it with the strict idempotency policy (worst case it fails because it was already deleted, but no data loss)
- We would want to have an integration test for this.
- You will need to change the testbench to support the integration test.

You may want to break this into smaller PRs, I used to do `*Request`, then the RawClient changes, then the testbench changes, then `Client` changes. Up to you but much easier to review and test smaller chunks of code.

vnvo2409:
@coryan 
I think that we could close this issue or am I missing something ?

coryan:
Agree.

PR #4617, opened 2020-07-22
feat(storage): a request type to delete pending resumable uploads

vnvo2409:
This PR is a part of work for #4404

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4617)


vnvo2409:
Some tests are cancelled.

PR #4605, opened 2020-07-21
refactor: refactor `BeginTransaction` to its own private method

mr-salty:
No functional change. We were doing this in 2 places and there will soon
be additional cases where it is needed.

Part of #4516

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4605)

Issue #4516, opened 2020-07-14
atomicity is violated if the first operation in a RW Transaction fails


mr-salty:
Given the following example (pseudocode, we should write an actual test)

```cc
1:  Transaction txn = MakeReadWriteTransaction();
2:  auto result1 = client_->ExecuteDml(txn, "INSERT that will fail (duplicate row)");
3:  auto result2 = client_->ExecuteDml(txn, "INSERT that will succeed");
4:  auto commit_result = client_->Commit(txn, {});
```

In our client library, although the first `ExecuteDml` fails, the second succeeds, as does the `Commit`.
This behavior may be seen as violating the atomicity guarantee - users may expect that the failure of the first `ExecuteDml` means that the transaction cannot successfully Commit.  From http://goto.google.com/spanner-txns#atomicity-consistency-durability

>In addition to the Isolation property, Spanner provides Atomicity (all writes must succeed in order for the transaction to commit; if one or more writes within the transaction fail, the entire transaction fails), [...]

The reason our client library behaves this way is because we defer `Transaction` begin until the first operation that uses the transaction as an optimization. If the first operation on a `Transaction` is a write (DML) that fails, the Spanner backend does not actually begin the transaction or return one in the response, the C++ `Transaction` object remains in its initial state; assuming all subsequent DML operations succeed, the `Commit` will also succeed.

devjgm:
Could we solve this by changing the implementation within the first `ExecuteDml` call (line 2 above) such that if the RPC fails to create a new transaction for us, then we assign some sentinel value to `txn` to indicate "FailedBegin", which would allow us to know that all future operations with `txn` should also fail? Would that fix this issue?

mr-salty:
Updates from the team discussion:

* We can't easily propagate errors because of subtleties involving constraint vs non-constraint errors (BTW I do not see these terms used in any external spanner documentation)
* Changing the Spanner API to return a transaction even when an operation fails would solve the problem, but is probably not feasible, especially not in the short term.
* Always explicitly calling `BeginTransaction` would solve the problem, at the expense of losing the performance optimization we get from "piggybacking" the begin on another operation.
* Although we cannot propagate errors in general due to the first bullet point, we can mark the transaction as "bad" if `BeginTransaction` fails, and cause all subsequent operations using that transaction to fail.

As a first step, we plan to implement support for marking a transaction "bad" and checking the status before using the transaction. These changes should be mostly or completely internal and not user visible.

[...]

PR #4624, opened 2020-07-22
refactor: refactor `BeginTransaction` to its own private method

mr-salty:
No functional change. We were doing this in 2 places and there will soon
be additional cases where it is needed.

This is an alternative to #4605.  I think this is a more idiomatic and
simpler signature for `BeginTransaction` at the cost of a little more
boilerplate at the call sites. My main concern isn't the amount of
boilerplate (it's only a couple lines) but the fact that omitting it at
any site may lead to subtle problems that wouldn't be possible in #4605
where `BeginTransaction` modifies the TransactionSelector directly.

Part of #4516

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4624)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -318,7 +318,7 @@ Status ConnectionImpl::PrepareSession(SessionHolder& session,
  * @param session identifies the Session to use.
  * @param options `TransactionOptions` to use in the request.
  * @param func identifies the calling function for logging purposes.
- *   It should generally be passed the value of the __func__ macro.
+ *   It should generally be passed the value of `__func__`.
  */
 StatusOr<spanner_proto::Transaction> ConnectionImpl::BeginTransaction(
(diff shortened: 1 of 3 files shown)


coryan:
Does this replace #4605? Should we close the older PR?

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -300,6 +312,38 @@ Status ConnectionImpl::PrepareSession(SessionHolder& session,
   return Status();
 }
 
+/**
+ * Performs an explicit `BeginTransaction` in cases where that is needed.
+ *
+ * @param session identifies the Session to use.
+ * @param options `TransactionOptions` to use in the request.
+ * @param func identifies the calling function for logging purposes.
+ *   It should generally be passed the value of the __func__ macro.
```
devbww, line 321: I wouldn't characterize it as a macro.

mr-salty: Done

PR #4663, opened 2020-07-23
fix(pubsub): code now builds with MSVC 2019

coryan:
Another case of `sizeof(int)` < `sizeof(std::size_t)` and protobuf using
`int` to index its vector-like things.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4663)

### DIFF
--- a/google/cloud/pubsub/internal/batching_publisher_connection.cc
+++ b/google/cloud/pubsub/internal/batching_publisher_connection.cc
@@ -128,7 +128,7 @@ void BatchingPublisherConnection::Flush(std::unique_lock<std::mutex> lk) {
   batch.waiters.reserve(pending_.size());
   google::pubsub::v1::PublishRequest request;
   request.set_topic(topic_full_name_);
-  request.mutable_messages()->Reserve(pending_.size());
+  request.mutable_messages()->Reserve(static_cast<int>(pending_.size()));
 
   for (auto& i : pending_) {
     batch.waiters.push_back(std::move(i.response));


google/cloud/pubsub/internal/batching_publisher_connection.cc
```diff
@@ -128,7 +128,7 @@ void BatchingPublisherConnection::Flush(std::unique_lock<std::mutex> lk) {
   batch.waiters.reserve(pending_.size());
   google::pubsub::v1::PublishRequest request;
   request.set_topic(topic_full_name_);
-  request.mutable_messages()->Reserve(pending_.size());
+  request.mutable_messages()->Reserve(static_cast<int>(pending_.size()));
```
rcleveng, line 131: would it make sense to introduce a ssize(C) method as a helper and eventually move to std::ssize(C) for compilers that support C++20?

coryan: TIL: `std::ssize()`.  I am not sure that would help: the compiler is (was?) warning about converting a 64-bit integer (`std::size_t`) to a 32-bit integer (`int`).  I do not believe signed vs. unsigned is the problem, tried with both `std::ptrdiff_t` and with `std::make_signed<std::size>::type`.

But this got me thinking: how do I know the `std::size_t` fits on an integer?  I think we should change the buffering constraints to guarantee that #4671

PR #4670, opened 2020-07-24
feat: store a Status when invalidating Transactions

mr-salty:
after further discussion, @devbww and I reversed the decision in #4545
to use `optional` to mark a `Transaction` invalid, and instead we will
store and propagate the `Status` of the failed `BeginTransaction`.

we also discussed how `Rollback` should function in various failure
scenarios, and I did some testing making manual RPCs to spanner.
When `Rollback` is called with an invalid `Transaction` it now
propagates the `Status` instead of returning `OK`. I also left a TODO
to change the behavior in the `has_begin` case to better match backend
behavior.

part of #4516

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4670)

### DIFF
--- a/google/cloud/spanner/internal/transaction_impl.h
+++ b/google/cloud/spanner/internal/transaction_impl.h
@@ -68,9 +68,9 @@ class TransactionImpl {
   // If the TransactionSelector is in the "begin" state and the operation
   // successfully allocates a transaction ID, then the functor must assign
   // that ID to the selector. If the functor fails to allocate a transaction
-  // ID then it should assign the status of the failing operation to the
-  // selector. All of this is independent of whether the functor itself
-  // succeeds.
+  // ID then it must assign a `Status` that indicates why transaction
+  // allocation failed (i.e. the result of `BeginTransaction`) to the parameter.
+  // All of this is independent of whether the functor itself succeeds.
   //
   // If the TransactionSelector is not in the "begin" state then the functor
   // must not modify it. Rather it should use either the transaction ID or
--- a/google/cloud/spanner/internal/transaction_impl.h
+++ b/google/cloud/spanner/internal/transaction_impl.h
@@ -19,7 +19,7 @@
 #include "google/cloud/spanner/version.h"
 #include "google/cloud/internal/invoke_result.h"
 #include "google/cloud/internal/port_platform.h"
-#include "absl/types/optional.h"
+#include "google/cloud/status_or.h"
 #include <google/spanner/v1/transaction.pb.h>
 #include <condition_variable>
 #include <cstdint>
@@ -35,7 +35,7 @@ namespace internal {
 template <typename Functor>
 using VisitInvokeResult = ::google::cloud::internal::invoke_result_t<
     Functor, SessionHolder&,
-    absl::optional<google::spanner::v1::TransactionSelector>&, std::int64_t>;
+    StatusOr<google::spanner::v1::TransactionSelector>&, std::int64_t>;
 
 /**
  * The internal representation of a google::cloud::spanner::Transaction.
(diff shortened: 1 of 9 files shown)


google/cloud/spanner/internal/transaction_impl.h
```diff
@@ -68,8 +68,9 @@ class TransactionImpl {
   // If the TransactionSelector is in the "begin" state and the operation
   // successfully allocates a transaction ID, then the functor must assign
   // that ID to the selector. If the functor fails to allocate a transaction
-  // ID then it should place the selector into an error state. All of this
-  // is independent of whether the functor itself succeeds.
+  // ID then it should assign the status of the failing operation to the
```
devbww, line 71: "the failing operation" may imply the (for example) `ExecuteSql()` instead of the `BeginTransaction()` that really failed to allocate the transaction ID.  I'm not sure how to tweak it though.

google/cloud/spanner/internal/transaction_impl_test.cc
```diff
@@ -92,11 +92,12 @@ class Client {
 
  private:
   ResultSet Read(SessionHolder& session,
-                 absl::optional<TransactionSelector>& selector,
+                 StatusOr<TransactionSelector>& selector,
                  std::int64_t seqno, std::string const& table,
                  KeySet const& keys, std::vector<std::string> const& columns);
 
   Mode mode_;
+  const Status failed_txn_status_{StatusCode::kInternal, "Bad transaction"};
```
devbww, line 100: If we're going to add a `const` member, we might as well constify `mode_` too.  Otherwise, drop `const`.

Alternatively, this status could be a local constant in `Read()`.

google/cloud/spanner/internal/transaction_impl_test.cc
```diff
@@ -110,20 +111,20 @@ class Client {
...
-        fail_with_throw = (valid_visits_ % 2 == 0);
-        break;
+    // we should only get here in "transaction invalidated" mode.
+    if (mode_ == Mode::kReadFailsAndTxnInvalidated) {
+      EXPECT_EQ(selector.status(), failed_txn_status_);
+      // visits always valid
+      ++valid_visits_;
+      fail_with_throw = (valid_visits_ % 2 == 0);
+    } else {
+      ADD_FAILURE() << "Unexpected mode_ " << static_cast<int>(mode_);
```
devbww, line 127: This would have been (and still will be) picked up by the expectation on `valid_visits_`.  So, unless we want to add failures for all the other places we don't increment `valid_visits_` during this visitor, I would suggest not adding this one.

google/cloud/spanner/testing/matchers.h
```diff
@@ -43,7 +44,7 @@ MATCHER_P2(
     "Verifies a Transaction has the expected Session and Transaction IDs") {
   return google::cloud::spanner::internal::Visit(
       arg, [&](google::cloud::spanner::internal::SessionHolder& session,
-               absl::optional<google::spanner::v1::TransactionSelector>& s,
+               StatusOr<google::spanner::v1::TransactionSelector>& s,
```
devbww: I also had
```
          *result_listener << "Transaction ID missing (expected "
                           << transaction_id << " but found status "
                           << s.status() << ")";
```
below.  That is, also report the status.

mr-salty:
___
*[google/cloud/spanner/internal/transaction_impl_test.cc, line 122 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4670#-MCzZZUv4WebIVjpP1Z-:-MCzbL10DY4P_cg-vLpS:b-90ibms) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9d25ce7de629e877ec1a3f935bf799ddb9f77a4c/google/cloud/spanner/internal/transaction_impl_test.cc#L122)):*

just to underscore the comment I made below, this case should just `break` instead of incrementing `valid_visits_`, correct? I changed it to do that.
</blockquote></details>

This comment didn't show up where I intended it to.  the question is, is `..RemainsBegin` valid here or not? (I think "not")

PR #4673, opened 2020-07-24
ci: use warnings-as-errors on Windows+MSVC+Bazel

coryan:
Enable most warnings (/W3) and treat warnings as errors (/WX) on the
Windows+MSVC+Bazel builds. We already do this on the Windows+MSVC+CMake
builds, but those are post-submit and therefore break from time to time.
Note that only the CMake builds compile for x86, so there are some
warnings we miss anyway.

The change is more intrusive than just adding some --copt options
because we only want *our* code compiled with these options, we do not
want to hear about warnings in our dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4673)

### DIFF
--- a/bazel/BUILD
+++ b/bazel/BUILD
@@ -24,5 +24,5 @@ config_setting(
     },
     values = {
         "cpu": "x64_windows",
-    }
+    },
 )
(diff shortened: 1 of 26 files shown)


coryan:
Dropping this, we have a less intrusive solution in the works.

PR #4677, opened 2020-07-24
fix(storage): missing CreateDefaultClientOptions overload

### LINKED ISSUE #4674: `g::c::s::Client::CreateDefaultClientOptions(ChannelOptions const& channel_options)` appears to be unimplemented
yes

storage

**Describe the bug**

Top-of-trunk `google/cloud/storage/client_options.h` contains a [second signature](https://github.com/googleapis/google-cloud-cpp/blob/9d25ce7de629e877ec1a3f935bf799ddb9f77a4c/google/cloud/storage/client_options.h#L82-L83) for `ClientOptions`: `::ClientOptions(ChannelOptions const& channel_options)`, which would be very useful, but this signature appears to be unimplemented in [`client_options.cc`](https://github.com/googleapis/google-cloud-cpp/blob/9d25ce7de629e877ec1a3f935bf799ddb9f77a4c/google/cloud/storage/client_options.cc#L70).

(this signature would be useful because the internal curl handle pool does not appear to respect the `channel_options` passed to `ClientOptions (std::shared_ptr< oauth2::Credentials > credentials, ChannelOptions channel_options)`  if the handle pool has already been initialized -- for example, by calling `StorageDefaultCredentials()`; so, it is necessary to recapitulate the steps in `CreateDefaultClientOptions`, in order to ensure that the handle pool is created with the correct options for the SSL cert path -- see work-around [here](https://github.com/TileDB-Inc/TileDB/pull/1741))

**To Reproduce**
- build library calling function `g::c::s::ClientOptions(ChannelOptions const& channel_options)` on linux
- symbol `_ZN6google5cloud7storage2v113ClientOptions26CreateDefaultClientOptionsERKNS2_14ChannelOptionsE;` will be listed as an external (`U`) symbol, and library will fail to load at runtime with undef symbol error

**Expected behavior**
Function signature is available.

**Operating system:**
manylinux2010 (CentOS 6 + newer toolchain)

g++ (GCC) 8.3.1 20190311 (Red Hat 8.3.1-3)

1.14, 1.15

**Additional context**
n/a

coryan:
Fixes #4674

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4677)

### DIFF
--- a/google/cloud/storage/client_options_test.cc
+++ b/google/cloud/storage/client_options_test.cc
@@ -14,10 +14,13 @@
 
 #include "google/cloud/storage/client_options.h"
 #include "google/cloud/storage/oauth2/google_credentials.h"
+#include "google/cloud/internal/random.h"
 #include "google/cloud/internal/setenv.h"
 #include "google/cloud/testing_util/assert_ok.h"
 #include "google/cloud/testing_util/scoped_environment.h"
 #include <gmock/gmock.h>
+#include <cstdlib>
+#include <fstream>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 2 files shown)


google/cloud/storage/client_options_test.cc
```diff
@@ -28,13 +31,39 @@ class ClientOptionsTest : public ::testing::Test {
  public:
   ClientOptionsTest()
       : enable_tracing_("CLOUD_STORAGE_ENABLE_TRACING", {}),
-        endpoint_("CLOUD_STORAGE_TESTBENCH_ENDPOINT", {}) {}
+        endpoint_("CLOUD_STORAGE_TESTBENCH_ENDPOINT", {}),
+        generator_(std::random_device{}()) {}
+
+  std::string CreateRandomFileName() {
+    // When running on the internal Google CI systems we cannot write to the
+    // local directory, GTest has a good temporary directory in that case.
```
devbww, line 39: s/, GTest/. GTest/

coryan: Done.

PR #4678, opened 2020-07-24
feat(storage): add a function `DeleteResumableUpload` to `RawClient`

vnvo2409:
This is part of work for #4404 

I am not sure one test is enough but I dont know how to test other clients ?

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4678)


coryan:
I understand why you cannot write more tests. We should really have a test for the logging layer at least, but that is not something you need to fix in this PR.  In any case, PSA: the builds are going to be slower than usual this afternoon, specially for Windows.

vnvo2409:
It seems that I have to rebase before running `ABI/API` dump

vnvo2409:
@coryan 
I would like to ask something. `Cmake` is too slow on my machine ( about 2 hours each run ), so I would like to set up a virtual desktop on Google Cloud Platform. Which cpu configuration ( standard, memory, compute; number of core, ram, etc ) I should choose for an optimization between cost and performance ( I will use it until I run out of credits about 250$ )

PR #4686, opened 2020-07-25
refactor: use absl::FormatTime instead of std::strftime

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4686)


devjgm:
windows builds look good.

Issue #4458, opened 2020-07-01
 Audit and remove unnecessary code that is now redundant in the monorepo with Abseil


devjgm:
This issue may spawn separate issues to clean up each of the redundant things that the audit finds. That's fine.

coryan:
Some related bugs:

- #4375 - absl::Status
- #108 - absl::conjunction
- #4528 - time parsing
- #3990 - stack trace
- #3934 - time parsing

There are probably many more. Search for:

- uses of `std::getline()` as a version of `absl::StrSplit()`
- the string `char const* sep = ` probably as part of something that could be replaced by `absl::StrJoin()`.

devjgm:
We may also want to replace our `google::cloud::Date` class with `absl::CivilDay`

remyabel:
I did not see an issue for `absl::StrJoin` so I created https://github.com/googleapis/google-cloud-cpp/issues/5121.

coryan:
@devjgm I think it would be useful to break down this issue into specific tasks, as-is, we never will know when we are "done".

devjgm:
I think the only real remaining issue here was the time-parsing code changes, so I split that out into https://github.com/googleapis/google-cloud-cpp/issues/5426, and will close this issue.

PR #4689, opened 2020-07-25
refactor: migrate away from using std::tm

devbww:
Remove `google::cloud::internal::AsUtcTm()`, which was only used to
adapt to `std::strftime()` in order to format dates.

Introduce `google::cloud::internal::FormatUtcDate()` instead, which
more closely matches the other `format_time_point.h` interfaces.

The ABI change is only in the internal namespace.

Part of #4458.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4689)


devbww:
Thanks.  I royally messed up the "you need to rebase and update your ABI/API baseline again", and this was the result.

Issue #4363, opened 2020-06-08
Windows build received an interrupt or signal from the OS or another process


coryan:
From time to time we experience build failures ending in:

```
kbuilder@localhost: Permission denied (publickey,password,keyboard-interactive).
rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
rsync error: unexplained error (code 255) at io.c(226) [Receiver=3.1.3]
nc: connection failed, SOCKS error 5
```

https://source.cloud.google.com/results/invocations/9bfa7650-9396-4df2-ab53-c87648297dfe/targets

AFAICT, this is a Kokoro issue I am creating this bug so we have something we can easily point to.

coryan:
https://source.cloud.google.com/results/invocations/6b1b7698-dcd2-4124-b335-6bd35e5fa440


```
2020-06-12T03:25:04.6844720+00:00 Compressing cache tarball
kokoro_log_reader: Received an interrupt or signal from the OS or another process while sleeping: [Errno 4] Interrupted function call. It's possible that the OS or another process is trying to kill kokoro_log_reader. Attempting to continue anyways.

Break signaled


[ID: 5093278] Build finished after 4635 secs, exit value: 1


Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
kbuilder@localhost: Permission denied (publickey,password,keyboard-interactive).
rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
rsync error: unexplained error (code 255) at io.c(226) [Receiver=3.1.3]
nc: connection failed, SOCKS error 5
kex_exchange_identification: Connection closed by remote host
rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
rsync error: unexplained error (code 255) at io.c(226) [Receiver=3.1.3]
Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
[20:46:22] Collecting build artifacts from build VM
Build script failed with exit code: 1
```

coryan:
https://source.cloud.google.com/results/invocations/eb2f1c52-c412-4492-906b-dfe8ed9b28e5

coryan:
https://source.cloud.google.com/results/invocations/232a0387-a1ab-423d-bac7-48c8b135fabe

coryan:
https://source.cloud.google.com/results/invocations/ec16f263-4c0e-412e-8692-4226af1deb63

devbww:
[...]

PR #4692, opened 2020-07-26
ci: reduce root disk usage on Windows

coryan:
We are informed that the problems in #4363 might be caused
by the CI machine running out of space in the boot partition.
This change aims to reduce the chances of that happening,
and to improve our ability to troubleshoot for that case.

Remove temporary files after they are no longer used, for
example, when refreshing the `vcpkg` cache we may need to
create all the `.o` files for the dependencies, but once these
are installed the `.o` files are no longer needed. Likewise, the
tarball for the Bazel cache is not needed after it is extracted.

Add some logging to report how much disk space is left before
and after some of the steps that consume significant disk.

----
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4692)

### DIFF
--- a/ci/kokoro/windows/build-bazel.ps1
+++ b/ci/kokoro/windows/build-bazel.ps1
@@ -132,26 +132,11 @@ Get-CimInstance -Class CIM_LogicalDisk | `
         @{L='FreeSpaceGB';E={"{0:N2}" -f ($_.FreeSpace /1GB)}}, `
         @{L="Capacity";E={"{0:N2}" -f ($_.Size/1GB)}}
 
-Write-Host -ForegroundColor Yellow "`n$(Get-Date -Format o) DEBUG DEBUG DEBUG"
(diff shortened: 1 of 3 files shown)


ci/kokoro/windows/build-bazel.ps1
```diff
@@ -32,6 +32,8 @@ if ($missing.count -ge 1) {
 # Create output directory for Bazel. Bazel creates really long paths,
 # sometimes exceeding the Windows limits. Using a short name for the
 # root of the Bazel output directory works around this problem.
+# We do this in C: (the boot disk) because T: (the tmpfs / build disk)
+# because T: is too small for the Bazel directory.
```
devbww, line 36: Nit: "because T: ... because T:" dup.

coryan: Thanks, I will send a PR to fix this.

PR #4696, opened 2020-07-27
feat(storage): add a function to delete resumable upload in client

vnvo2409:
This PR is a part of #4404 

It seems that it is impossible to implement a test for this function.
- I dont know how to get the `upload_id` from `Object: insert`
- I dont know how to send a signal to interrupt a resumable upload so we can delete that resumable upload.

I think we should implement a function to return the `upload_id` to user. With that function we could actually resume a resumable upload ( if there is an interruption ) and write test for `DeleteResumableUpload`

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4696)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -1256,6 +1256,7 @@ class Client {
    * This operation is always idempotent because it only acts on a specific
    * `upload_id`.
    */
+  template <typename... Options>
   Status DeleteResumableUpload(std::string const& bucket_name,
                                std::string const& object_name,
                                std::string const& upload_id,
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -1246,23 +1246,19 @@ class Client {
   /**
    * Cancel a resumable upload.
    *
-   * @param bucket_name the name of the bucket that contains the object.
-   * @param object_name the name of the object that is uploading.
-   * @param upload_id the id of the resumable upload.
+   * @param upload_session_url the url of the upload session. Returned by
+   * `ObjectWriteStream::resumable_session_id`.
    * @param options a list of optional query parameters and/or request headers.
    *   Valid types for this operation include `UserProject`.
    *
    * @par Idempotency
    * This operation is always idempotent because it only acts on a specific
-   * `upload_id`.
+   * `upload_session_url`.
    */
   template <typename... Options>
-  Status DeleteResumableUpload(std::string const& bucket_name,
-                               std::string const& object_name,
-                               std::string const& upload_id,
+  Status DeleteResumableUpload(std::string const& upload_session_url,
                                Options&&... options) {
-    internal::DeleteResumableUploadRequest request(bucket_name, object_name,
-                                                   upload_id);
+    internal::DeleteResumableUploadRequest request(upload_session_url);
     request.set_multiple_options(std::forward<Options>(options)...);
     return raw_client_->DeleteResumableUpload(request).status();
   }
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -1249,15 +1249,20 @@ class Client {
    * @param bucket_name the name of the bucket that contains the object.
    * @param object_name the name of the object that is uploading.
    * @param upload_id the id of the resumable upload.
+   * @param options a list of optional query parameters and/or request headers.
+   *   Valid types for this operation include `UserProject`.
    *
    * @par Idempotency
-   * This operation is idempotent.
+   * This operation is always idempotent because it only acts on a specific
+   * `upload_id`.
    */
   Status DeleteResumableUpload(std::string const& bucket_name,
                                std::string const& object_name,
-                               std::string const& upload_id) {
+                               std::string const& upload_id,
+                               Options&&... options) {
     internal::DeleteResumableUploadRequest request(bucket_name, object_name,
                                                    upload_id);
+    request.set_multiple_options(std::forward<Options>(options)...);
     return raw_client_->DeleteResumableUpload(request).status();
   }
 
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -1244,6 +1244,24 @@ class Client {
   }
 
   /**
+   * Cancel a resumable upload.
+   *
+   * @param bucket_name the name of the bucket that contains the object.
+   * @param object_name the name of the object that is uploading.
+   * @param upload_id the id of the resumable upload.
+   *
+   * @par Idempotency
+   * This operation is idempotent.
+   */
+  Status DeleteResumableUpload(std::string const& bucket_name,
+                               std::string const& object_name,
+                               std::string const& upload_id) {
+    internal::DeleteResumableUploadRequest request(bucket_name, object_name,
+                                                   upload_id);
+    return raw_client_->DeleteResumableUpload(request).status();
+  }
+
+  /**
    * Downloads a Cloud Storage object to a file.
    *
    * @param bucket_name the name of the bucket that contains the object.
--- a/google/cloud/storage/internal/curl_client.cc
+++ b/google/cloud/storage/internal/curl_client.cc
@@ -707,14 +707,11 @@ CurlClient::RestoreResumableSession(std::string const& session_id) {
 
 StatusOr<EmptyResponse> CurlClient::DeleteResumableUpload(
     DeleteResumableUploadRequest const& request) {
-  CurlRequestBuilder builder(
-      upload_endpoint_ + "/b/" + request.bucket_name() + "/o", upload_factory_);
+  CurlRequestBuilder builder(request.upload_session_url(), upload_factory_);
   auto status = SetupBuilderCommon(builder, "DELETE");
   if (!status.ok()) {
     return status;
   }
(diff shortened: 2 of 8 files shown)


vnvo2409:
How about the `upload_id` ?

coryan:
> How about the `upload_id` ?

Context?  Maybe this answers your question?

https://github.com/googleapis/google-cloud-cpp/blob/603cac26af91e62c18ef6e002f6d9bcaef2054a5/google/cloud/storage/examples/storage_object_resumable_write_samples.cc#L35-L38

https://github.com/googleapis/google-cloud-cpp/blob/603cac26af91e62c18ef6e002f6d9bcaef2054a5/google/cloud/storage/object_stream.h#L312-L323

vnvo2409:
I think we have to simulate an interruption instead.
When I test against real bucket on GCS server using
```
auto stream = client->WriteObject(
      bucket_name_, object_name, NewResumableUploadSession());
auto session_id = stream.resumable_session_id();
auto upload_id = session_id.substr(session_id.find("upload_id=") + strlen("upload_id="));

stream << "This data will not get uploaded, it is too small\n"; 
std::move(stream).Suspend();

auto status = client->DeleteResumableUpload(bucket_name_, object_name, upload_id);
EXPECT_STATUS_OK(status);
```

the return status is
```
Actual: Permanent error in DeleteResumableUpload: {
  "error": {
    "code": 499,
    "message": "clientClosedRequest",
    "errors": [
      {
        "message": "clientClosedRequest",
        "domain": "global",
        "reason": "clientClosedRequest"
      }
    ]
  }
}
 [INVALID_ARGUMENT] [INVALID_ARGUMENT]
```

coryan:
So I tried the same thing using just `curl` and I got the same error...


```
curl --http1.1 --verbose -H'Content-Length: 0' -H 'Content-Type: application/octet-stream' -H "Authorization: Bearer $(gcloud --quiet auth print-access-token)" -H 'Content-Range: bytes */*' -X POST 'https://storage.googleapis.com/upload/storage/v1/b/coryan-test-bucket/o?uploadType=resumable&name=test-02.txt'
```

```
curl --http1.1 --verbose -H'Content-Length: 0' -H 'Content-Type: application/octet-stream' -H "Authorization: Bearer $(gcloud --quiet auth print-access-token)" -H 'Content-Range: bytes */*' -X DELETE 'https://storage.googleapis.com/upload/storage/v1/b/coryan-test-bucket/o?uploadType=resumable&name=test-02.txt&upload_id=.......'
< HTTP/1.1 499 Client Closed Request
```

This is weird, but maybe the error message is to be expected?  Let me ask around, or maybe @frankyn knows the answer.

coryan:
It seems that 499 is the expected (but not well documented) status code for canceling an upload. I have filed a bug internally to fix the documentation.

Meanwhile, I think we will need to accept both 200s and 499 as "success", a bit disgusting I know.

vnvo2409:
So we will return `ok` if we receive 499/200 in `DeleteResumableUpload` or just return the "raw" status and let user decide what to do ?

coryan:
> So we will return `ok` if we receive 499/200 in `DeleteResumableUpload` or just return the "raw" status and let user decide what to do ?

I think `CurlClient` should handle this and return `ok` on 499 and/or 200 (and 200-299) status. When we implement this for gRPC the behavior might change and we should not burden customers with that level of detail.

vnvo2409:
The problem is more complicated than we thought. if we use the upload session after delete it, GCS server will return 499.
```
TEST_F(ObjectIntegrationTest, DeleteResumableUpload) {
  StatusOr<Client> client = MakeIntegrationTestClient();
  ASSERT_STATUS_OK(client);

  auto object_name = MakeRandomObjectName();
  auto stream = client->WriteObject(
      bucket_name_, object_name, NewResumableUploadSession());
  auto session_id = stream.resumable_session_id();
  auto upload_id = session_id.substr(session_id.find("upload_id=") + strlen("upload_id="));

  stream << "This data will not get uploaded, it is too small\n"; 
  std::move(stream).Suspend();

  auto status = client->DeleteResumableUpload(bucket_name_, object_name, upload_id);
  EXPECT_STATUS_OK(status);

  auto client_options = ClientOptions::CreateDefaultClientOptions();
  ASSERT_STATUS_OK(client_options);
  Client client_resumable(client_options->set_maximum_simple_upload_size(0));
  auto stream_resumable = client_resumable.WriteObject(bucket_name_, object_name, RestoreResumableUploadSession(session_id));
  stream_resumable << LoremIpsum();
  stream_resumable.Close();
}
```
`stream_resumable.metadata().status()` is 499

coryan:
> The problem is more complicated than we thought. if we use the upload session after deleting it, GCS server will return 499.

Right. I think we should treat that error code as success **only** for `DeleteResumableUpload()`, everywhere else we should simply treat them as before.

> ```
> auto stream_resumable = client_resumable.WriteObject(bucket_name_, object_name, RestoreResumableUploadSession(session_id));
>   stream_resumable << LoremIpsum();
>   stream_resumable.Close();
> }
> ```
> 
> `stream_resumable.metadata().status()` is 499

I would expect that to fail, after all we intentionally deleted the session.

vnvo2409:
Hmmm. In my opinion, 404 or 499 does not matter, because we cannot mimic the behavior in production. You could think about the case where we call `DeleteResumableUpload` twice on the same `upload_id`. The first call should return `OK`, but the second call is weird because:
- The server returns 499 and we return status `OK` but the second call should be failed because `upload_id` is already deleted ( when testing against real bucket ).
- if we return 4xx ( but not 499 ) , the second call does not return `OK`. This is what I expected but the response code isn't right ( when testing against testbench ).

With me, an ideal test is having 2 threads, one thread will upload and the other thread will send a signal to kill the first thread, ( I am thinking that `Suspend()` has done something with the `upload_id`  so GCS server return 499 ). But this is too complicated for an unit test ?

coryan:
___
*[google/cloud/storage/testbench/testbench.py, line 854 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4696#-MDKMhzu7K4qgVUvxYxN:-MDKqooK1kNRilg-Wtz4:b-himf5q) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a9d6215e7e74b69cc882be1af9d04fd45ac87c2e/google/cloud/storage/testbench/testbench.py#L854)):*
> With me, an ideal test is having 2 threads, one thread will upload and the other thread will send a signal to kill the first thread, ( I am thinking that Suspend() has done something with the upload_id  so GCS server return 499 ). But this is too complicated for an unit test ?

Let's separate the concerns about how we create tests vs. what the right behavior is for the testbench.   The tesbench should work more or less the same regardless of how we torture the test to produce the desired results.

> Hmmm. In my opinion, 404 or 499 does not matter, because we cannot mimic the behavior in production.

Okay, that is always true.  How *much* of the behavior from production we *need* to mimic is the question I typically ask myself.

> You could think about the case where we call DeleteResumableUpload twice on the same upload_id. The first call should return OK, but the second call is weird because:
> 
> The server returns 499 and we return status OK but the second call should be failed because upload_id is already deleted ( when testing against real bucket ).

I think you are saying that production would generate the following responses:

|      | production | testbench | notes |
| ----| --------- | --- | --- |
| --> | POST | POST |  (create a resumable upload) |
| <-- | 200 | 200 | (with the resumable upload id) |
| -->  | DELETE | DELETE | (the just created upload id) |
| <--  | 499 | 200 | the current behavior in the tesbench, could be changed to 499 |
| -->  | DELETE | DELETE | (same upload id) |
| <--  | 499 | 404 | the current behavior in the tesbench, cannot be changed to 499 without a lot more work. |


> if we return 4xx ( but not 499 ) , the second call does not return OK. This is what I expected but the response code isn't right ( when testing against testbench ).
> 

I see, the testbench would need to keep a history of any old upload ids it generated to respond with 499 on the second `DELETE` and that is hard to implement. If it returns 404 that is going to cause trouble... 

Okay, let's leave the code as-is, and file a bug to improve the testbench later.

coryan:
___
*[google/cloud/storage/tests/object_integration_test.cc, line 816 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4696#-MDKMsK40OOA7QdtUgFG:-MDLQGU9BRHXd32DLTnW:brqln2v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a9d6215e7e74b69cc882be1af9d04fd45ac87c2e/google/cloud/storage/tests/object_integration_test.cc#L816)):*

That is probably the right name, except that I chose poorly when I named the full URL `resumable_session_id()`: the two names are too close to each other and confusing.  Unfortunately, I do not think we should change `resumable_session_id()` name without a deprecation period.... let me think for a bit... maybe `resumable_upload_short_id()` and `resumable_upload_full_url()` would be better names, and we can start the process to remove `resumable_session_id()` ... still need to think a bit more.  Let me summon @devjgm to see what he thinks.
</blockquote></details>

I have an idea, what if we change `DeleteResumableUpload()` to accept the value returned by `resumable_session_id()`? It can do its own parsing if needed.  That might prove handy when we enable this for gRPC where the format will be different.

vnvo2409:
I think it is a good idea. To me, it seems to be a little confusing. But I think this is the best solution

PR #4702, opened 2020-07-27
ci: ignore internal symbols in ABI checker

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4702)

### DIFF
--- a/ci/kokoro/docker/check-abi.sh
+++ b/ci/kokoro/docker/check-abi.sh
@@ -55,7 +55,14 @@ check_library() {
   (
     cd "${BINARY_DIR}"
     zcat "${reference_file}" >"${old_dump_file}"
+    # We ignore all symbols in an "::internal::" namespace, because these are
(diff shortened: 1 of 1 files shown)


devjgm:
> Really, cool, thanks!
> 
> Do you know if it would be possible to also ignore `pubsub_internal` ?
> 
> > Reviewed 1 of 1 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4702#-:-MDHBR9e34oIaJltekQ-:b4yugmx)_ status: :shipit: complete! all files reviewed, all discussions resolved

Yep, I think so. After grepping, `internal` and `pubsub_internal` are our only two "internal" namespaces. So I updated the regex to explicitly list both of those. If we end up getting a bunch of differently named internal namespaces, we could perhaps update the regex to be a bit more liberal to cover all of them. But for now, this should be conservative and work.

PR #4706, opened 2020-07-28
fix: explicitly begin failed implicit begin transactions

### LINKED ISSUE #4516: atomicity is violated if the first operation in a RW Transaction fails

mr-salty:
If an operation which has `TransactionSelector::begin` set in order to
implicitly begin a transaction fails, explicitly call `BeginTransaction`.

If that fails, invalidate the `Transaction` so all subsequent operations
fail - most importantly `Commit`, which is necessary to maintain atomicity
since the first operation did not execute.

Otherwise, retry the original operation using the transaction ID
obtained from `BeginTransaction`. That may fail or succeed, but even if
it fails, it may be a "non-constraint error" (read errors and certain
types of write/dml errors) which permit the rest of the `Transaction` to
succeed. Beginning the transaction leaves it to the backend to make
that decision.

Fixes #4516

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4706)


mr-salty:
let me take a look at the build failures... it passed all unit tests and integration tests before I committed.

mr-salty:
PTAL - forgot to check ok() in one place before dereferencing (and the first test failures I looked at sent me down the wrong rathole, it looked like some gmock issue)... should have looked at asan first - and also #4707 which might have made this more obvious.

mr-salty:
well, apparently the gmock thing was an actual issue. It sometimes thinks one of the `GrpcReader` calls is a sequence prerequisite for one of the `SpannerStub` calls, I don't understand why (is there some threading involved here?  even that doesn't seem to fit the symptoms) but I may be holding `InSequence` incorrectly, especially combined with a lambda. I think this more explicit version should work but I'm not able to reproduce the problem reliably.

In any case, I think it's an issue with the test and not the code so I think you can still review it. If the latest commit doesn't fix it I'll take another look in the morning.

mr-salty:
looks like that fixed the test issues although Brad and I were discussing a change to handing the case where we get an "OK" response and no transaction - we should probably handle that differently... although I think that could be done in a follow-up PR, it's a corner case that probably won't occur in practice (and the checked in code does not handle it so it's not a regression).

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -436,33 +440,47 @@ StatusOr<std::vector<ReadPartition>> ConnectionImpl::PartitionReadImpl(
...
+                spanner_proto::PartitionReadRequest const& request) {
+          return stub->PartitionRead(context, request);
+        },
+        request, __func__);
+    if (s->has_begin()) {
+      if (response.ok() && response->has_transaction()) {
+        s->set_id(response->transaction().id());
+      } else {
+        // A transaction should have been returned but was not.
+        auto begin = BeginTransaction(session, s->begin(), __func__);
```
devbww, line 457: I don't think we should do this if the response was OK.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -436,33 +440,47 @@ StatusOr<std::vector<ReadPartition>> ConnectionImpl::PartitionReadImpl(
...
+        if (!begin.ok()) {
+          s = begin.status();  // invalidate the transaction
+          return begin.status();
```
devbww, line 460: And, given that the response was not OK, I think this should return `response.status()`.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -436,33 +440,47 @@ StatusOr<std::vector<ReadPartition>> ConnectionImpl::PartitionReadImpl(
...
+        }
+        s->set_id(begin->id());
+        *request.mutable_transaction() = *s;
+        continue;
```
devbww, line 464: In general, I thought the control flow would look like:
```
{
  auto request = <setup-request>();
  for (;;) {
    *request.mutable_transaction() = *s;
    auto response = <handle-request>(request);
    if (s->has_begin()) {
      if (!response) {
        auto begin = BeginTransaction(...);
        if (!begin) {
          s = begin.status();
          return response;
        }
        s->set_id(begin->id);
        continue;  // try again with id
      }
      if (response->has_transaction()) {
        s->set_id(response->transaction().id());
      }
    }
    return <handle-response>(response);
  } 
} 
```

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -402,6 +409,51 @@ TEST(ConnectionImplTest, ReadImplicitBeginTransaction) {
   EXPECT_THAT(txn, HasSessionAndTransactionId("test-session-name", "ABCDEF00"));
 }
 
+TEST(ConnectionImplTest, ReadImplicitBeginTransactionPermanentFailure) {
+  auto mock = std::make_shared<spanner_testing::MockSpannerStub>();
+
+  auto db = Database("dummy_project", "dummy_instance", "dummy_database_id");
+  auto conn = MakeLimitedRetryConnection(db, mock);
+
+  auto make_reader = []() {
```
devbww, line 418: Nit: I believe we choose to omit the `()` when there are no parameters.

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -402,6 +409,51 @@ TEST(ConnectionImplTest, ReadImplicitBeginTransaction) {
...
+    grpc::Status grpc_status(grpc::StatusCode::PERMISSION_DENIED, "uh-oh");
+    auto reader = absl::make_unique<MockGrpcReader>();
+    EXPECT_CALL(*reader, Read(_)).WillOnce(Return(false));
+    EXPECT_CALL(*reader, Finish()).WillOnce(Return(grpc_status));
+    return reader;
+  };
+  // TODO(salty) some interaction between the spanner calls and the reader
+  // calls seems to confuse gmock and cause some flakiness. Does explicitly
+  // sequencing them fix the issue?
```
devbww, line 427: Is this resolved?

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -402,6 +409,51 @@ TEST(ConnectionImplTest, ReadImplicitBeginTransaction) {
...
+      .WillOnce(Return(MakeTestTransaction()));
+  EXPECT_CALL(*mock, StreamingRead(_, _))
+      .InSequence(s)
+      .WillOnce(Return(ByMove(std::move(reader2))));
+
+  Transaction txn = MakeReadOnlyTransaction(Transaction::ReadOnlyOptions());
+  auto rows = conn->Read({txn, "table", KeySet::All(), {"UserId", "UserName"}});
+  for (auto& row : rows) {
+    EXPECT_EQ(StatusCode::kPermissionDenied, row.status().code());
+    EXPECT_THAT(row.status().message(), HasSubstr("uh-oh"));
```
devbww, line 453: Use `StatusIs()` matcher?

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -677,6 +748,37 @@ TEST(ConnectionImplTest, ExecuteDmlDeleteTooManyTransientFailures) {
   EXPECT_THAT(result.status().message(), HasSubstr("try-again in ExecuteDml"));
 }
 
+// Tests that a Transaction that fails to begin does not successfully commit -
```
devbww, line 751: s/ -/ as/?  Or, two sentences?

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -677,6 +748,37 @@ TEST(ConnectionImplTest, ExecuteDmlDeleteTooManyTransientFailures) {
...
+  Status status(StatusCode::kInvalidArgument, "cannot begin transaction");
+  {
+    InSequence seq;
+    EXPECT_CALL(*mock, BatchCreateSessions(_, _))
+        .WillOnce(Return(MakeSessionsResponse({"session-name"})));
+
+    // The first `ExecuteDml` call tries to implicitly begin the transaction
+    // via `ExecuteSql`, and then explicitly via `BeginTransaction`. Both fail,
+    // and we should receive no further RPC calls - since the transaction is
+    // not valid ther client fails any subsequent operations itself.
```
devbww, line 768: s/ther/the/

mr-salty:
___
*[google/cloud/spanner/internal/connection_impl.cc, line 457 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4706#-MDIxMy6_7MWigBSuS4q-r2-457:-MDMJY_jDWpf-FilHOWl:b-apxsvx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4da49406b20a6420ea7fa8dfb6385dfb6e67fed6/google/cloud/spanner/internal/connection_impl.cc#L457)):*

I don't think we should do this if the response was OK.
</blockquote></details>

my earlier comment: Brad and I were discussing a change to handing the case where we get an "OK" response and no transaction - we should probably handle that differently... although I think that could be done in a follow-up PR, it's a corner case that probably won't occur in practice (and the checked in code does not handle it so it's not a regression).

mr-salty:
___
*[google/cloud/spanner/internal/connection_impl_test.cc, line 453 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4706#-MDIxMy7wgVIf6XymFEF-r2-453:-MDNtKkh1dxN8jQdtkJV:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4da49406b20a6420ea7fa8dfb6385dfb6e67fed6/google/cloud/spanner/internal/connection_impl_test.cc#L453)):*

It looked like a new use.
</blockquote></details>

Done

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -389,19 +398,22 @@ RowStream ConnectionImpl::ReadImpl(
...
-      auto metadata = reader.ok() ? (*reader)->Metadata() : absl::nullopt;
-      if (metadata && metadata->has_transaction()) {
-        s->set_id(metadata->transaction().id());
+      if (reader.ok()) {
+        auto metadata = (*reader)->Metadata();
+        if (metadata && metadata->has_transaction()) {
+          s->set_id(metadata->transaction().id());
+        } else {
+          s = MissingTransactionStatus(__func__);
+          return MakeStatusOnlyResult<RowStream>(s.status());
```
devbww, line 407: Observation: It still seems strange to me to fail an operation that actually succeeded (despite the missing transaction ID), but I guess I'm happy in the knowledge that it never happens.

Issue #4645, opened 2020-07-23
Implement flow control for subscribers


coryan:
We want to add configuration parameters to limit the number of messages outstanding in a subscription. The library should not send more `Pull()` requests if there are:

- More than N messages still being processed (not acked and not nacked), **or**
- The messages being processed contain more than M bytes of data.

coryan:
See #4809 for the details on how to calculate the bytes in a message.

Issue #4583, opened 2020-07-17
Automatically refresh ack deadline for slow callbacks


coryan:
Applications should be able to set a high ack deadline for message processing, the library needs to automatically modify the deadline for the message as long as the callback is still working.

PR #4711, opened 2020-07-28
feat(pubsub): sequence message handle on ack/nack

coryan:
So far we have been using the return from the subscription handler to
schedule the next event. This is not correct if we want to limit the
number of messages being handled and support handlers that use some kind
of thread pool.

This is part of the work for #4583 but just because it makes testing easier.
I believe it will also help with #4645 but that remains to be seen.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4711)

### DIFF
--- a/google/cloud/pubsub/integration_tests/message_integration_test.cc
+++ b/google/cloud/pubsub/integration_tests/message_integration_test.cc
@@ -80,7 +80,7 @@ TEST(MessageIntegrationTest, PublishPullAck) {
   promise<void> ids_empty;
   auto handler = [&](pubsub::Message const& m, AckHandler h) {
     SCOPED_TRACE("Search for message " + m.message_id());
-    std::lock_guard<std::mutex> lk(mu);
+    std::unique_lock<std::mutex> lk(mu);
     auto i = ids.find(m.message_id());
     // Remember that Cloud Pub/Sub has "at least once" semantics, so a dup is
     // perfectly possible, in that case the message would not be in the map of
@@ -93,9 +93,10 @@ TEST(MessageIntegrationTest, PublishPullAck) {
       ++i->second;
       return;
     }
-    std::move(h).ack();
     ids.erase(i);
     if (ids.empty()) ids_empty.set_value();
+    lk.unlock();
+    std::move(h).ack();
   };
 
   auto result = subscriber.Subscribe(subscription, handler);
--- a/google/cloud/pubsub/internal/subscription_session_test.cc
+++ b/google/cloud/pubsub/internal/subscription_session_test.cc
@@ -26,6 +26,7 @@ namespace {
 
 using ::testing::_;
 using ::testing::AtLeast;
+using ::testing::AtMost;
 
 /// @test Verify callbacks are scheduled in the background threads.
 TEST(SubscriptionSessionTest, ScheduleCallbacks) {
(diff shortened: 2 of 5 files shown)


google/cloud/pubsub/internal/subscription_session_test.cc
```diff
@@ -101,6 +102,74 @@ TEST(SubscriptionSessionTest, ScheduleCallbacks) {
...
+    }
+    return make_ready_future(make_status_or(response));
+  };
+  EXPECT_CALL(*mock, AsyncPull(_, _, _)).WillOnce(generate_3);
+
+  // This receives a message, verify that it is the next message in the sequence
+  // and then sets up the expectations for the next set of calls (mostly
+  // AsyncPull).
+  std::atomic<int> expected_message_id{0};
+  auto handler = [&](pubsub::Message const& m, pubsub::AckHandler h) {
```
devbww, line 133: This seems to imply that `handler` will be called after the first mock call, which would break the rule that expectations must be set before the first mock call.

coryan: > This seems to imply that `handler` will be called after the first mock call,

Necessarily so: the values that the handler is called with are generated by the `AsyncPull()` mock.

> which would break the rule that expectations must be set before the first mock call.

What I am trying to test is that once values are consumed (and only then) we expect another mock call to be made, and only once those are consumed we expect the next mock call, and so on and so forth.

Let me take another stab at expressing this with `::testing::InSequence` or something...

coryan: PTAL.

google/cloud/pubsub/integration_tests/message_integration_test.cc
```diff
@@ -93,9 +93,10 @@ TEST(MessageIntegrationTest, PublishPullAck) {
       ++i->second;
       return;
     }
-    std::move(h).ack();
     ids.erase(i);
     if (ids.empty()) ids_empty.set_value();
+    lk.unlock();
```
devbww, line 98: Naive observation: It is a little confusing that we go out of our way to drop the lock before `h.ack()` but not before `h.nack()`.

google/cloud/pubsub/internal/subscription_session_test.cc
```diff
@@ -124,46 +124,61 @@ TEST(SubscriptionSessionTest, SequencedCallbacks) {
...
+  promise<void> enough_messages;
+  std::atomic<int> received_counter{0};
+  auto constexpr kMaximumMessages = 9;
+  auto handler = [&](pubsub::Message const& m, pubsub::AckHandler h) {
+    auto c = received_counter.load();
+    EXPECT_LE(c, kMaximumMessages);
+    SCOPED_TRACE("Running for message " + m.message_id() +
+                 ", counter=" + std::to_string(c));
+    EXPECT_EQ("test-message-id-" + std::to_string(c), m.message_id());
+    if (++received_counter == kMaximumMessages) {
```
devbww, line 168: It looks like this should require an atomic operator.

devbww: I see now that `operator++()` is defined on `std::atomic`.  Sorry about that.

google/cloud/pubsub/integration_tests/message_integration_test.cc
coryan: Yeah, I think this test will need some cleanup. Filed #4715

PR #4712, opened 2020-07-28
ci: parallelize check-abi build steps

devjgm:
The "check-abi" part of the build was done sequentially taking ~7
minutes on my machine. After this change it takes about 2.5 minutes.

I also changed the "update-abi" build to only update the expected ABI
files and skip the compliance check. The rationale for this change is
because if the user is asking to update the expected ABI is probably
because a previous "check-abi" run failed and that's how they know they
need to update the ABI.

One final small change is that the names in the report are changed from
"reference" and "current" to "expected" and "actual", which matches the
file names better.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4712)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -371,11 +371,18 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
     /bin/false
   fi
 
-  # Checking the ABI requires installation, so this is the first opportunity to
-  # run the check.
-  env -C "${PROJECT_ROOT}" \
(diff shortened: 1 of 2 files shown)


devjgm:
> nit: maybe a comment explaining that this script only runs on a Fedora-based build, because that setting is not portable. OTOH, we did not comment on this before.

Done.

devjgm:
> Alternatively, you might capture the individual outputs in the calling script and then present them intact and separated.

I originally thought of that, but that produces bursts of output as processes finish rather than realtime output on the screen. I preferred the realtime output of this on the screen. Plus, if we were actually trying to debug a problem we could pretty easily reconstruct the nicer output by sorting the lines offline.

devjgm:
> OK, but if `blaze test` behaved that way I think we'd complain.
> 
> >

https://github.com/googleapis/google-cloud-cpp/issues/4713

PR #4719, opened 2020-07-29
feat(pubsub): auto refresh ack deadline

### LINKED ISSUE #4583: Automatically refresh ack deadline for slow callbacks

coryan:
Automatically update the Ack deadline for all the events still being
processed (that is, those where the application has neither called
`ack()` nor `nack()`). The application can set a maximum ACK deadline
when the subscription is created, and this applies to the full batch of
messages received by the application even if the messages are
serialized. This is consitent with the golang implementation.

Several tests have different expectations because we avoid sending an
initial `AsyncModifyAckDeadline()` RPC until needed. We expect many
applications to process messages in the initial 10 seconds, so saving
the call seems worthwhile.

Fixes #4583

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4719)

### DIFF
--- a/google/cloud/pubsub/subscription_options.h
+++ b/google/cloud/pubsub/subscription_options.h
@@ -1,0 +1,67 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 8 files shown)


google/cloud/pubsub/subscription_options.h
```diff
@@ -0,0 +1,67 @@
...
+   * @note The deadline applies to each message as it is delivered to the
+   *     application, thus, if the library receives a batch of N messages their
+   *     deadline for all the messages is extended repeatedly. Only once the
+   *     message is delivered to a callback does the deadline become immutable.
+   */
+  std::chrono::seconds max_deadline_time() const { return max_deadline_time_; }
+
+  /// Set the maximum deadline for incoming messages.
+  SubscriptionOptions& set_max_deadline_time(std::chrono::seconds d) {
+    max_deadline_time_ = std::chrono::duration_cast<std::chrono::seconds>(d);
```
devbww, line 52: Nit: It looks like the cast is a no-op ... `seconds` to `seconds`.

coryan: Doh!  I will send a PR in a ... soon.

PR #4728, opened 2020-07-30
feat(pubsub): support move-only message callbacks

### LINKED ISSUE #4556: Support move-only callables in pubsub::Subscriber::Subscribe()
I am about so send a PR with the skeleton for `pubsub::SubscriberConnection`. This initial version uses `std::function<>` to type erase the callback provided by the application. This might be too heavy weight, and a lighter weight type erasure might work better.

coryan:
The previous implementation used `std::function<>` to type-erase the
callable provided by the application, which (AFAIK) requires the
callable to be copy constructible.

Fixes #4556

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4728)

### DIFF
--- a/google/cloud/pubsub/subscriber_connection.h
+++ b/google/cloud/pubsub/subscriber_connection.h
@@ -24,6 +24,7 @@
 #include "google/cloud/pubsub/version.h"
 #include "google/cloud/status_or.h"
 #include <functional>
+#include <memory>
 
 namespace google {
 namespace cloud {
@@ -37,12 +38,16 @@ class SubscriberConnection {
  public:
   virtual ~SubscriberConnection() = 0;
 
-  // TODO(#4556) - consider a lighter weight type-erasure
-  using CallbackType = std::function<void(Message, AckHandler)>;
+  // Type-erase `void(Message, AckHandler)` callables with support for move-only
+  // callables and with less overhead than `std::function<>`.
+  struct Callback {
+    virtual ~Callback() = default;
+    virtual void exec(Message, AckHandler) = 0;
+  };
 
   struct SubscribeParams {
     std::string full_subscription_name;
-    CallbackType callback;
+    std::unique_ptr<Callback> callback;
     SubscriptionOptions options;
   };
   virtual future<Status> Subscribe(SubscribeParams p) = 0;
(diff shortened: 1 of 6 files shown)


google/cloud/pubsub/subscriber_connection.h
```diff
@@ -37,12 +38,16 @@ class SubscriberConnection {
  public:
   virtual ~SubscriberConnection() = 0;
 
-  // TODO(#4556) - consider a lighter weight type-erasure
-  using CallbackType = std::function<void(Message, AckHandler)>;
+  // Type-erase `void(Message, AckHandler)` callables with support for move-only
+  // callables and with less overhead than `std::function<>`.
+  struct Callback {
+    virtual ~Callback() = default;
+    virtual void exec(Message, AckHandler) = 0;
```
devjgm, line 45: Should probably be `Exec`

coryan:
Rebased and resolved conflicts.

google/cloud/pubsub/subscriber_connection.h
```diff
@@ -74,6 +79,23 @@ std::shared_ptr<pubsub::SubscriberConnection> MakeSubscriberConnection(
     std::shared_ptr<SubscriberStub> stub,
     pubsub::ConnectionOptions const& options);
 
+template <typename Callable>
+std::unique_ptr<pubsub::SubscriberConnection::Callback> MakeSubscriberCallback(
+    Callable&& cb) {
+  class Wrapper : public pubsub::SubscriberConnection::Callback {
+   public:
+    explicit Wrapper(absl::decay_t<Callable> cb) : cb_(std::move(cb)) {}
```
devjgm, line 87: The argument of type `absl::decay_t<Callable>` is a value that will be move-constructed due to the one call site down on line 96. Then it will do a second move-construction into `cb_` in this c'tor initialization list. We could avoid the first move-construction by making this c'tor a template

```cc
...
  template <typename U>
  explicit Wrapper(U&& u) : cb_(std::forward<U>(u)) {}
...
```
Of course, doing this would probably require that this class be defined outside the function, becuase IIRC, we can't define templates within a function. But this is already in `pubsub_internal`, so maybe that's not a problem?

Anyway, WDYT?

coryan:
> So to be clear, yes, I'd approve this change. It LGTM. But I'd be just as happy to skip it.

I think you convinced me to not do this, or at least punt on it. I will close this PR and create a new one to remove the TODO.

Issue #4561, opened 2020-07-17
Implement logging decorator for SubscriberStub


coryan:
Implement a decorator that logs every request and response.

PR #4746, opened 2020-07-31
feat: log the result for future<StatusOr<T>>

coryan:
When a function returns `future<StatusOr<T>>` we cannot log the results
immediately, they may not even exist! Previously we just logged whether
the returned future was ready, now we attach a callback to the future
which logs the value whenever it becomes available.

Because this may be much later in the log, the application might need a
token to match the request and response pairs. I am using a per-thread
counter and the thread id as an identifier.

Part of the work for #4561 and #3900

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4746)

### DIFF
--- a/google/cloud/internal/log_wrapper.cc
+++ b/google/cloud/internal/log_wrapper.cc
@@ -14,8 +14,8 @@
 
 #include "google/cloud/internal/log_wrapper.h"
 #include <google/protobuf/text_format.h>
+#include <atomic>
 #include <sstream>
-#include <thread>
 
 namespace google {
 namespace cloud {
@@ -35,10 +35,8 @@ std::string DebugString(google::protobuf::Message const& m,
 }
 
 std::string RequestIdForLogging() {
-  static std::atomic<int> generator{0};
-  std::ostringstream os;
-  os << std::this_thread::get_id() << ":" << ++generator;
-  return std::move(os).str();
+  static std::atomic<std::uint64_t> generator{0};
+  return std::to_string(++generator);
 }
 
 char const* DebugFutureStatus(std::future_status status) {
--- a/google/cloud/internal/log_wrapper.cc
+++ b/google/cloud/internal/log_wrapper.cc
@@ -35,7 +35,7 @@ std::string DebugString(google::protobuf::Message const& m,
 }
 
 std::string RequestIdForLogging() {
-  thread_local std::uint64_t generator = 0;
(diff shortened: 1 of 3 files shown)


google/cloud/internal/log_wrapper.cc
```diff
@@ -32,6 +34,32 @@ std::string DebugString(google::protobuf::Message const& m,
   return str;
 }
 
+std::string RequestIdForLogging() {
+  thread_local std::uint64_t generator = 0;
+  std::ostringstream os;
+  os << std::this_thread::get_id() << ":" << ++generator;
```
devbww, line 40: I don't think this is guaranteed to be unique.  When a thread ends, objects with thread storage duration are destroyed, and the thread ID is available for reuse.  So, you could see something like "X:1" twice.  Perhaps a static atomic is better.

coryan: Good point, fixed.

devbww: `int` can be just 32 bits, I think.  If so, a binary with 1Kqps would overflow in less than a month.

I don't know if that rate or uptime is realistic, but I'd probably burn an extra 32 bits to avoid the issue.

In either case, I'd also suggest an unsigned type so that overflow is at least defined.

google/cloud/internal/log_wrapper.cc
```diff
@@ -35,7 +35,7 @@ std::string DebugString(google::protobuf::Message const& m,
 }
 
 std::string RequestIdForLogging() {
-  thread_local std::uint64_t generator = 0;
+  static std::atomic<int> generator{0};
   std::ostringstream os;
   os << std::this_thread::get_id() << ":" << ++generator;
```
devbww, line 40: It seems like the thread ID is redundant now (and probably more confusing).

coryan: PTAL

google/cloud/internal/log_wrapper.cc
```diff
@@ -14,6 +14,8 @@
 
 #include "google/cloud/internal/log_wrapper.h"
 #include <google/protobuf/text_format.h>
+#include <atomic>
+#include <sstream>
```
devbww, line 18: Nit: This addition now looks unused.

coryan: Fixed too.

Issue #4751, opened 2020-08-01
Add testbench support for GCS+gRPC


vnvo2409:
I will work on this feature by extending Python client to support GCS+gRPC with a different endpoint ( `/grpc` ).
The changes will be:
- Change `GrpcClient` to work with different endpoint when testing with testbench.
- Add `/grpc` endpoint to testbench.
- Add test for existing RPCs like `CreateBucket()`

cc @coryan

vnvo2409:
@coryan 
Some updates:
`HybridClient` is now passing integration test. I am using `grpc_integration_test` with this env set to `media`
https://github.com/googleapis/google-cloud-cpp/blob/70e507792b0edd06c05efd4dc2f32cc9ff46d888/google/cloud/storage/tests/grpc_integration_test.cc#L55

Because it is a large project, I have created a new repository for it https://github.com/vnvo2409/gcs-testbench

I have some questions:
1. Is it possible if I turn this project into a pip package and `google-cloud-cpp` will use it instead of the current testbench ? Because there isn't an offical emulator for gcs that supports both grpc and rest, we could extend this project and I think there will be many projects find this emulator useful.

2. Are all the integration tests inside `google/cloud/storage/tests/` ?

coryan:
> Some updates:
> `HybridClient` is now passing integration test. I am using `grpc_integration_test` with this env set to `media`
> https://github.com/googleapis/google-cloud-cpp/blob/70e507792b0edd06c05efd4dc2f32cc9ff46d888/google/cloud/storage/tests/grpc_integration_test.cc#L55

Really cool, thanks!

> 
> Because it is a large project, I have created a new repository for it https://github.com/vnvo2409/gcs-testbench

Please remember to include the right license and copyright notices in this code. The majority of that code is owned by Google, and the License requires that you reproduce the Copyright notice in any copies of the code:

https://github.com/googleapis/google-cloud-cpp/blob/8d878749ec893b2005e2117a182d4386d30daac5/LICENSE#L89-L128



> 
> I have some questions:
> 
> 1. Is it possible if I turn this project into a pip package and `google-cloud-cpp` will use it instead of the current testbench ? Because there isn't an offical emulator for gcs that supports both grpc and rest, we could extend this project and I think there will be many projects find this emulator useful.

There may be some interest in doing that, let's ask @frankyn if he can suggest a repository to host the testbench separately.

> 2. Are all the integration tests inside `google/cloud/storage/tests/` ?

[...]

PR #4752, opened 2020-08-02
feat(storage): add testbench support for GCS+gRPC

vnvo2409:
This is just a quick and dirty implementation for the GCS+gRPC emulator.
Turn out, we should and must implement a new client instead of our old Python client because:
- `Flask` does not support `HTTP/2.0`
- We could use `protobuf` to generate python code for our testbench ( `google/cloud/storage_v1/proto/storage_*` are generated from `//google/storage/v1:storage_py_grpc` in https://github.com/googleapis/googleapis )

To use this testbench, we need to run 
```bash
python testbench_utils.py 
bazelisk test //google/cloud/storage/tests:grpc_integration_test \
    --test_env=GOOGLE_CLOUD_PROJECT="test" \
    --test_env=CLOUD_STORAGE_GRPC_ENDPOINT=localhost:8585
```

The first 2 tests ( `BucketCRUD` and `ObjectCRUD` should passed ).

Part of work for #4751 .

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4752)


vnvo2409:
All tests should pass now except `ReproLargeInsert` since it is a mixture between production and testbench. 

One more thing, I don't know how to add unknown fields to proto object so I have to disable the test `x_testbench_upload`

vnvo2409:
> It seems unfortunate to commit the generated *_pb2.py and *_pb2_grpc.py files. Would it be possible to obtain those as dependencies (seems unlikely)? Or maybe commit a CMake rule to (re)-generate them as needed?

We've already had the workspace (`"@com_google_googleapis`) for these dependencies indeed.  https://github.com/googleapis/google-cloud-cpp/blob/d3f7d79d1c612b10cdba4e12294aa04c67fead0c/google/cloud/storage/BUILD#L63
Now we need `@com_google_googleapis//google/storage/v1:storage_py_grpc`. We have 2 options:
- Use bazel to generate thoes files and run the test seperate
- Write a bazel rule to run the testbench and we can include `@com_google_googleapis//google/storage/v1:storage_py_grpc` as a dependency.

> Is it possible to run Flask and gRPC in the same python script? Even if they use different ports? For a while we will have to support the HybridClient where some requests use gRPC and some use REST (because that is how production is currently setup, and will be for a few quarters).

This may take some time, I will explore it.

> If we can have a single script supporting both protocols we can use the same global dictionaries for buckets, objects, etc., and at least the requests made through one protocol are visible in the second protocol.

I think it is hard to keep a global dictionaries for buckets and objects because their structures are quite different ( our bucket class vs protobuf class )

coryan:
> > It seems unfortunate to commit the generated *_pb2.py and *_pb2_grpc.py files. Would it be possible to obtain those as dependencies (seems unlikely)? Or maybe commit a CMake rule to (re)-generate them as needed?
> 
> We've already had the workspace (`"@com_google_googleapis`) for these dependencies indeed.
> 
> https://github.com/googleapis/google-cloud-cpp/blob/d3f7d79d1c612b10cdba4e12294aa04c67fead0c/google/cloud/storage/BUILD#L63
> 
> 
> Now we need `@com_google_googleapis//google/storage/v1:storage_py_grpc`. We have 2 options:
> * Use bazel to generate thoes files and run the test seperate
> * Write a bazel rule to run the testbench and we can include `@com_google_googleapis//google/storage/v1:storage_py_grpc` as a dependency.

Okay, I vote for documenting how to use Bazel to generate these files and committing the files. Not pretty, but I would not like to mix CMake and Bazel in the same build.

> 
> > Is it possible to run Flask and gRPC in the same python script? Even if they use different ports? For a while we will have to support the HybridClient where some requests use gRPC and some use REST (because that is how production is currently setup, and will be for a few quarters).
> 
> This may take some time, I will explore it.
> 
> > If we can have a single script supporting both protocols we can use the same global dictionaries for buckets, objects, etc., and at least the requests made through one protocol are visible in the second protocol.
> 
> I think it is hard to keep a global dictionaries for buckets and objects because their structures are quite different ( our class vs protobuf class )

Right. So the gRPC implementation would need to read the protos and convert them to the corresponding REST requests and responses.  This should be less hard than it appears because the protocols are designed to be fairly compatible to each other.  Of course we could do it in the opposite direction (use gRPC for the backend and convert the REST objects to gRPC).  My guess (and this is only a guess) is that the first approach is easier.

vnvo2409:
I think it is impossible Flask and gRPC in the same python script. If I run Flask and gRPC without gunicorn, some tests will failed because of some missing endpoints ( each endpoint is an app ). If I use gunicorn, all tests pass but gRPC does not response at all ( even with a different port )

The real problem is both flask and gunicorn does not support HTTP/2.0 while gRPC requires HTTP/2.0
If we want REST and gRPC in the same python script, we wiil need to use a diffrent framwork rather than Flask and do a quite large refactor in the testbench. 

What do you think ?

coryan:
Thank you for researching the combined Flask + gRPC problems.

> What do you think ?

Here is another a crazy idea: write a script that receives gRPC requests and sends REST to the existing benchmark.  It does not need to keep any state, and would allow us to run tests that combine REST and gRPC in a single application.

vnvo2409:
I've found a microframework that has the same API with Flask and it supports HTTP/2.0 https://pypi.org/project/Quart/.
I am quite interested with the refactor idea. So let I try refactoring the code.

vnvo2409:
> Right. So the gRPC implementation would need to read the protos and convert them to the corresponding REST requests and responses. This should be less hard than it appears because the protocols are designed to be fairly compatible to each other. Of course we could do it in the opposite direction (use gRPC for the backend and convert the REST objects to gRPC). My guess (and this is only a guess) is that the first approach is easier.

To me, it is harder than what we thought. Because, if we convert gRPC to REST object, we will have to convert REST object back to gRPC if `HybridClient` want a gRPC response. So we will have to do both way indeed. The fact that we are using many helper functions like `filtered_response` to hide the implementation of `gcs_bucket`, ... causes many difficulties to incorporate with gRPC. We need a refactor anyway, maybe a big project for me. What do you think ?

coryan:
Yes, this does sound like a big project. I completely understand if you want to tackle something else instead. Let me know one way or the other.

vnvo2409:
I will try refactoring and will let you know later.

vnvo2409:
Good news, I can run gRPC and Flask in the same python script with different port. The trick is to run without `gunicorn` and make both `gRPC server` and `Flask server` global variable to avoid Python garbage collection. Now, we have to refactor only `gcs_bucket`, `gcs_object` so it work better with gRPC

vnvo2409:
It will take a month since I dont have much time now, is it ok ?

coryan:
Yeah, no worries.  Nobody else has time to work on it right now, so you are making the fastest possible progress.

coryan:
I am going to close this PR though, just to keep things tidy, we can reopen when the time comes.

PR #4780, opened 2020-08-05
test: query parameters when resuming upload

### LINKED ISSUE #3300: Use common query parameters in CurlResumableUploadSession::ResetSession
When we reset an upload session we do not use the `userProject`, `quotaUser`, and other similar query parameters from the original request.  We should save these parameters from the original request, and send them in any ResetSession requests.

### LINKED ISSUE #4781: Use query parameters in `UploadChunkRequest`/`InsertObjectRequest`.
Query parameters passed to `Client::WriteObject` are only used to create the resumable session and resume it (when #4780 is merged).

They should also be passed on every individual chunk of data is uploaded (be it via JSON API, i.e. `UploadChunkRequest` or gRPC API - `InsertObjectRequest`).

dopiera:
This fixes #3300 and fixes #4781.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4780)

### DIFF
--- a/google/cloud/storage/tests/curl_resumable_upload_session_integration_test.cc
+++ b/google/cloud/storage/tests/curl_resumable_upload_session_integration_test.cc
@@ -16,6 +16,7 @@
 #include "google/cloud/storage/testing/storage_integration_test.h"
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/testing_util/assert_ok.h"
+#include "google/cloud/testing_util/capture_log_lines_backend.h"
 #include <gmock/gmock.h>
 
 namespace google {
@@ -33,9 +34,13 @@ class CurlResumableUploadIntegrationTest
                        "GOOGLE_CLOUD_CPP_STORAGE_TEST_BUCKET_NAME")
                        .value_or("");
     ASSERT_FALSE(bucket_name_.empty());
+    project_id_ =
+        google::cloud::internal::GetEnv("GOOGLE_CLOUD_PROJECT").value_or("");
+    ASSERT_FALSE(project_id_.empty());
   }
 
   std::string bucket_name_;
+  std::string project_id_;
 };
 
 TEST_F(CurlResumableUploadIntegrationTest, Simple) {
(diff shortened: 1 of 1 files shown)


dopiera (commented):
Eventually it turned out that all f this is unneeded. I added a test showing that all the query parameters are either remembered on the server side or added to `session_id`, which is a url with those parameters set.
 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4780#-:-MF32zDpBAmLWMfvtThU:biui5gs)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @coryan)

___
*[google/cloud/storage/internal/curl_request_builder.h, line 187 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4780#-MDyN665DVcUUlNlxxat:-MF33FqDETQd2BsVPupW:ba3dfob) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2af7d2798ed7982f24eb897e192195b8ca4988e2/google/cloud/storage/internal/curl_request_builder.h#L187)):*

Isn't this something that the builder can determine by itself?  If the URL contains a `?` then it can initialize the query_parameter_separator to `&`?
</blockquote></details>

Removed
___
*[google/cloud/storage/tests/curl_resumable_upload_session_integration_test.cc, line 224 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4780#-MDyMoT15Au6r_pPxR4h:-MF33Igl8eU0kd5m6CLa:ba3dfob) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2af7d2798ed7982f24eb897e192195b8ca4988e2/google/cloud/storage/tests/curl_resumable_upload_session_integration_test.cc#L224)):*

As an alternative, consider capturing the parameters in the emulator (via custom attributes) and then checking the custom attributes here.
</blockquote></details>

Removed
___
*[google/cloud/storage/tests/curl_resumable_upload_session_integration_test.cc, line 258 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4780#-MDyLzzyDzlqm4XvWGSz:-MF33K75-0gtrr0u0VMI:ba3dfob) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2af7d2798ed7982f24eb897e192195b8ca4988e2/google/cloud/storage/tests/curl_resumable_upload_session_integration_test.cc#L258)):*

> ```
>   auto predicate = [this](std::string const& line) {
>     return line.find(" PUT ") != std::string::npos &&
>            line.find("/b/" + bucket_name_ + "/o") != std::string::npos &&
>            line.find("quotaUser=test-quota-user") != std::string::npos;
>   };
>   auto count = std::count_if(backend->log_lines.begin(),
>                              backend->log_lines.end(), predicate);
>   EXPECT_LT(0, count);
> ```

I recently learnt how to use the `::testing::Contains()` matcher to write these checks as:


```cc
EXPECT_THAT(backend->log_lines, Contains(AllOff(HasSubstr(" PUT "), HasSubStr("/b/" + bucket_name_ + "/o"), HasStubstr("quotaUser=test-quota-user"))));
```

Arguably less code, arguably more obscure (for non-googletest/googlemock savants).  Without argument it produces better errors on mismatches.

</blockquote></details>

Removed

Issue #4788, opened 2020-08-05
Consider dropping gcc 4.8 support


devjgm:
Today we have a [CI build that tests using gcc 4.8.5 on centos-7](https://github.com/googleapis/google-cloud-cpp/blob/2c596e310e4de862143bd120b184ea714f305b75/ci/kokoro/docker/build.sh#L178-L188). An upcoming Abseil release will break on this compiler. Abseil dropped gcc 4.8 in August 2019, requiring gcc >= 4.9. We officially [support](https://github.com/googleapis/google-cloud-cpp#supported-platforms) GCC >= 4.9 as well, even though we test with 4.8.5.

Centos 7 is a supported version of Linux. It's default compiler is gcc 4.8.5, which I believe is why we test with it. However, it is possible to install gcc 7 on centos 7 with something like:
 
1. `yum install centos-release-scl`
2. `yum install devtoolset-7`
3. `scl enable devtoolset-7 bash`
4. `gcc --version`, which should show gcc version 7

We should consider dropping gcc 4.8 support. If we did this, would we need a gcc 4.**9** build?

If we did this we'd need to do the following:

- [ ] Install gcc-7 in our centos-7 docker image(s)
- [ ] Update our packaging instructions (which should be automatic since that's generated from the Docker files)
- [ ] Rename our "gcc-4.8" build to indicate the version it's testing

coryan:
*IF* we are dropping GCC-4.8 then I think we can drop GCC-4.9 too: of the distros that we support, the next GCC version installed by default is GCC-5.x on Ubuntu:16.04. (I said "I think", somebody should check my math).

I am not sure what can we realistically do at this time other than dropping GCC-4.8.  It makes me sad that we are forced to do so, without giving customers more notice.  It makes me happy because we can stop writing ugly workarounds for it.

devjgm:
Some data gathering about Linux OS versions and the default compiles available.

## Our "docker" builds
distro | gcc | clang
--------| ---- | ----
centos:7 | 4.8.5 | 3.4.2
fedora:31 | 9.3.1 | 9.0.1
fedora:32 | 10.2.1 | 10.0.0
ubuntu:xenial (16.04) | 5.4.0 | 3.8.0
ubuntu:bionic (18.04) | 7.5.0 | 6.0.0

## Our "install" builds

distro | gcc | clang
--------| ----- | ------
centos:8 | 8.3.1 | 9.0.1
[...]

PR #4796, opened 2020-08-05
support!: drop support for gcc 4.8

devjgm:
Part of: #4788

BREAKING CHANGE: This PR raises our oldest supported GCC version from 4.8 (the default installed on centos-7) to GCC 5.4, which is the oldest gcc that's installed on the Linux distros that are supported by GCE and that we test with. Ubuntu Xenian (16.04) comes with gcc 5.4.

This PR adds a new build named "gcc-5.4". In follow-up changes, I'll enable that build in our CI, and disable the existing "gcc-4.8" build.

This PR also changes our centos:7 "install" test to use gcc-7 (instead of 4.8) using the `devtoolset-7` Software Collection.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4796)


devjgm:
> The changes LG. I think so far we have been using (maybe incorrectly) `doc:` and `doc!:` for the PR tags. Do you mind double checking? It is weird because this changes both documentation and ci strings, so `ci!: ...` may be appropriate too. Give it some thought and decide, I am Okay with either.

I according to https://www.conventionalcommits.org/en/v1.0.0/, I think `docs:` (plural) is "correct", though you're right that we typically use `doc:` (singular). Our codebase has 122 `doc:` and only 32 `docs:`.

I wasn't sure which commit type to use, because as you say, this is a doc change and a CI change. The reason I chose "docs" is because the most important part of this change is the change to the user-facing documentation where we drop support for gcc 4.8. Changes to our "ci" are generally things that users don't care about. But doc changes are user-facing, so I wanted to highlight that.

Alternatively, I think we can make up our own commit types. Maybe we could invent a `support!:` type for something like this?

devjgm:
> TIL: `docs:` vs. `doc:`
> 
> I have a slight preference for `support!:` I think it tells us more about what is happening. I am happy with `doc[s]!:`, on second thought `ci!:` is a bad idea because we skip those in the release notes. Totally your call.
> 
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4796#-:-ME3Aifc9C-2bPlGOQYK:beyc6og)_ status: :shipit: complete! all files reviewed, all discussions resolved

I like `support!:` too. And the conventionalcommits bot seems to be happy with it. So using this one.

PR #4797, opened 2020-08-05
test(GCS+gRPC): verify large chunks create one stream

coryan:
When the application provides a large buffer in `.write()` we want to
upload the full buffer in a single streaming RPC call.  Add a new test
to do this, note that in the test the application-level buffer becomes a
"chunk" but the application never has direct control on those.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4797)


coryan:
Sorry about that, now the description matches the code...

PR #4806, opened 2020-08-06
feat(common): non-proto values in PaginationRange

coryan:
Part of the work for #4568

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4806)

### DIFF
--- a/google/cloud/internal/pagination_range.h
+++ b/google/cloud/internal/pagination_range.h
@@ -21,6 +21,7 @@
 #include <functional>
 #include <iterator>
 #include <string>
+#include <type_traits>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


google/cloud/internal/pagination_range.h
```diff
@@ -29,6 +30,18 @@ namespace cloud {
 inline namespace GOOGLE_CLOUD_CPP_NS {
 namespace internal {
 
+inline bool ComparePaginationValues(google::protobuf::Message const& lhs,
+                                    google::protobuf::Message const& rhs) {
+  return google::protobuf::util::MessageDifferencer::Equals(lhs, rhs);
+}
+
+template <typename T, typename std::enable_if<
```
devjgm, line 38: (optional): I might have written this overload sfinae'ing on the existence of `==`. Maybe like:

```cc
template <typename T>
auto ComparePaginationValues(T const& a, T const& b) -> decltype(a == b) {
return a == b;
}
```
Not sure which way is really "preferred", but something for you to consider. Happy with whatever you decide.

coryan: I will add a separate bug for that... seems intriguing, but I am happy with the code as-is for now.

coryan: #4814

Issue #4819, opened 2020-08-06
Consider: Use `...` as the "run everything" bazel target specification


devjgm:
We can probably change the build target specification to simply `...` here and in most places below.

_Originally posted by @devjgm in https://github.com/googleapis/google-cloud-cpp/pull/4804/files_

We may want to simplify our bazel invocations to use `...`, which will also ensure we're always building and testing everything rather than accidentally forgetting something.

PR #4824, opened 2020-08-07
ci: simplify bazel targets to consistently use `...` to build everything

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/4819

This simplifies our build scripts and future-proofs them so that our bazel builds always build everything. In bazel world `...` is the way to specify that (the trailing `:all` is unnecessary)`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4824)

### DIFF
--- a/ci/check-style.sh
+++ b/ci/check-style.sh
@@ -34,7 +34,7 @@ readonly TIMEFORMAT="... %R seconds"
 problems=""
 printf "%-30s" "Running check-include-guards:"
 time {
-  if ! find google/cloud generator -name '*.h' -print0 |
+  if ! git ls-files -z | grep -zE '\.h$' |
     xargs -0 awk -f "ci/check-include-guards.gawk"; then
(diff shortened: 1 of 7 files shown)


ci/check-style.sh
```diff
@@ -34,7 +34,7 @@ readonly TIMEFORMAT="... %R seconds"
 problems=""
 printf "%-30s" "Running check-include-guards:"
 time {
-  if ! find google/cloud generator -name '*.h' -print0 |
+  if ! git ls-files -z | grep -zE '\.h$' |
```
devbww, line 37: This breaks when you `git rm` a file ... `find` won't find it, but `git ls-files` will list it.

devbww: Clarification: What I said is only the case for an __unstaged__ deletion, which makes it less problematic.

PR #4840, opened 2020-08-07
feature: add more utils, Printer, and ClassGeneratorInterface for generator

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4840)

### DIFF
--- a/generator/internal/class_generator_interface.h
+++ b/generator/internal/class_generator_interface.h
@@ -26,7 +26,7 @@ namespace generator_internal {
  */
 class ClassGeneratorInterface {
  public:
-  virtual ~ClassGeneratorInterface() = default; 
+  virtual ~ClassGeneratorInterface() = default;
   virtual Status Generate() = 0;
 };
 
--- a/generator/internal/class_generator_interface.h
+++ b/generator/internal/class_generator_interface.h
@@ -20,6 +20,10 @@ namespace google {
 namespace cloud {
 namespace generator_internal {
 
+/**
+ * Interface class such that ServiceGenerator can invoke its collection of
+ * class generators polymorphically.
+ */
 class ClassGeneratorInterface {
  public:
   virtual ~ClassGeneratorInterface() {}
--- a/generator/internal/class_generator_interface.h
+++ b/generator/internal/class_generator_interface.h
@@ -1,0 +1,33 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#ifndef GOOGLE_CLOUD_CPP_GENERATOR_INTERNAL_CLASS_GENERATOR_INTERFACE_H
+#define GOOGLE_CLOUD_CPP_GENERATOR_INTERNAL_CLASS_GENERATOR_INTERFACE_H
+
+#include "google/cloud/status.h"
+
+namespace google {
+namespace cloud {
+namespace generator_internal {
+
+class ClassGeneratorInterface {
+ public:
+  virtual ~ClassGeneratorInterface() {}
+  virtual Status Generate() = 0;
+};
+
+}  // namespace generator_internal
+}  // namespace cloud
+}  // namespace google
+
+#endif  // GOOGLE_CLOUD_CPP_GENERATOR_INTERNAL_CLASS_GENERATOR_INTERFACE_H
--- a/generator/internal/class_generator_interface.h
+++ b/generator/internal/class_generator_interface.h
@@ -26,7 +26,7 @@ namespace generator_internal {
  */
 class ClassGeneratorInterface {
  public:
-  virtual ~ClassGeneratorInterface() {}
+  virtual ~ClassGeneratorInterface() = default; 
   virtual Status Generate() = 0;
 };
 
--- a/generator/internal/codegen_utils.cc
+++ b/generator/internal/codegen_utils.cc
@@ -90,13 +90,20 @@ std::string ProtoNameToCppName(absl::string_view proto_name) {
 
 StatusOr<std::vector<std::string>> BuildNamespaces(
     std::map<std::string, std::string> const& vars, NamespaceType ns_type) {
-  // vars["product_path"] must to be present and properly formatted.
   auto iter = vars.find("product_path");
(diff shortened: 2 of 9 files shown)


generator/internal/class_generator_interface.h
```diff
@@ -0,0 +1,33 @@
...
+#ifndef GOOGLE_CLOUD_CPP_GENERATOR_INTERNAL_CLASS_GENERATOR_INTERFACE_H
+#define GOOGLE_CLOUD_CPP_GENERATOR_INTERNAL_CLASS_GENERATOR_INTERFACE_H
+
+#include "google/cloud/status.h"
+
+namespace google {
+namespace cloud {
+namespace generator_internal {
+
+class ClassGeneratorInterface {
```
devjgm, line 23: Please document this interface. I don't care too much about doxygen, but at least comments for other developers on this would be helpful.

generator/internal/codegen_utils.cc
```diff
@@ -87,6 +88,44 @@ std::string ProtoNameToCppName(absl::string_view proto_name) {
   return "::" + absl::StrReplaceAll(proto_name, {{".", "::"}});
 }
 
+std::vector<std::string> BuildNamespaces(
+    std::map<std::string, std::string> const& vars, NamespaceType ns_type) {
+  // vars["product_path"] is guaranteed to be present and properly formatted.
```
devjgm, line 93: What guarantees this? It makes me uncomfortable to have maps with keys that are "guaranteed to exist". If the key must exist, should we use a struct or something with a field that must always exist?

generator/internal/codegen_utils.cc
```diff
@@ -87,6 +88,44 @@ std::string ProtoNameToCppName(absl::string_view proto_name) {
...
+  std::string product_path = vars.find("product_path")->second;
+  std::vector<std::string> v = absl::StrSplit(product_path, '/');
+  auto name = v[v.size() - 2];
```
devjgm, line 96: What guarantees that this element exists in the vector?

generator/internal/codegen_utils.cc
```diff
@@ -87,6 +88,44 @@ std::string ProtoNameToCppName(absl::string_view proto_name) {
...
+
+  auto product_path =
+      std::find_if(command_line_args.begin(), command_line_args.end(),
+                   [](std::pair<std::string, std::string> const& p) {
+                     return p.first == "product_path";
+                   });
+  if (product_path == command_line_args.end() || product_path->second.empty()) {
+    return Status(StatusCode::kInvalidArgument,
+                  "--cpp_codegen_opt=product_path=<path> must be specified.");
+  }
```
devjgm, line 119: `auto& path = product_path->second;`
... then you can just use `path` in the lines below.

generator/internal/codegen_utils.h
```diff
@@ -63,6 +65,21 @@ std::string ServiceNameToFilePath(absl::string_view service_name);
...
+
+/**
+ * Builds namespace hierarchy.
+ */
+std::vector<std::string> BuildNamespaces(
+    std::map<std::string, std::string> const& vars,
+    NamespaceType ns_type = NamespaceType::NORMAL);
+
+/**
+ * Validates command line arguments passed to the microgenerator.
```
devjgm, line 78: Can you document what the `parameters` string must look like? I'd guess a string of `"key=val foo=bar blah=baz"`, but that's just a guess.

generator/internal/printer.h
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Wrapper around a google::protobuf::io::ZeroCopyOutputStream and a
+ * google::protobuf::io::Printer object so that they can be used for
+ * code generation.
+ */
+class Printer {
+ public:
+  Printer(google::protobuf::compiler::GeneratorContext* generator_context,
+          std::string const& file_name)
+      : output_(generator_context->Open(file_name)),
```
devjgm, line 37: Can this line fail?

generator/internal/printer.h
```diff
@@ -0,0 +1,76 @@
...
+   * Print some text after applying variable substitutions.
+   *
+   * If a particular variable in the text is not defined, this will crash.
+   * Variables to be substituted are identified by their names surrounded by
+   * delimiter characters (as given to the constructor).  The variable bindings
+   * are defined by the given map.
+   */
+  void Print(const std::map<std::string, std::string>& variables,
+             absl::string_view text) {
+    printer_->Print(variables, text.data());
```
devjgm, line 51: It's not safe to use a string_view's `.data()` without also consulting its size. `.data()` is not guaranteed to be (and often is not) nul-terminated.

generator/internal/printer.h
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * Like the first Print(), except the substitutions are given as parameters.
+   */
+  template <typename... Args>
+  void Print(absl::string_view text, const Args&... args) {
```
devjgm, line 58: I think we might want `Args&&` and then `std::forward` them. WDYT?

generator/internal/printer_test.cc
```diff
@@ -0,0 +1,75 @@
...
+  MOCK_METHOD(int64_t, ByteCount, (), (const, override));
+  MOCK_METHOD(bool, WriteAliasedRaw, (void const*, int), (override));
+  MOCK_METHOD(bool, AllowsAliasing, (), (const, override));
+};
+
+class PrinterTest : public testing::Test {
+ protected:
+  void SetUp() override {
+    generator_context_ = absl::make_unique<MockGeneratorContext>();
+    output_ = new MockZeroCopyOutputStream();
```
devjgm, line 50: Is this being leaked?  Can it be wrapped in a unique_ptr?

generator/internal/codegen_utils.h
```diff
@@ -63,6 +65,21 @@ std::string ServiceNameToFilePath(absl::string_view service_name);
  */
 std::string ProtoNameToCppName(absl::string_view proto_name);
 
+enum class NamespaceType { NORMAL, INTERNAL };
```
devjgm, line 68: I think these should be named `kNormal` and `kInternal`. After https://github.com/googleapis/google-cloud-cpp/pull/4844 lands, I think clang-tidy will complain about this.

generator/internal/codegen_utils.h
```diff
@@ -63,6 +65,25 @@ std::string ServiceNameToFilePath(absl::string_view service_name);
...
+ */
+std::vector<std::string> BuildNamespaces(
+    std::map<std::string, std::string> const& vars,
+    NamespaceType ns_type = NamespaceType::kNormal);
+
+/**
+ * Validates command line arguments passed to the microgenerator.
+ *
+ * Command line arguments can be passed from the protoc command line via:
+ * '--cpp_codegen_opt=key=value'. This can be specified multiple times to
```
devjgm, line 81: So the `parameters` argument should be a string that looks like `"--cpp_codegen_opt=key=value"` or `"--cpp_codegen_opt=key=value --cpp_codegen_opt=key=value --cpp_codegen_opt=key=value"` ?

PR #4863, opened 2020-08-11
feature(generator): add ServiceGenerator and StubGenerator

scotthart:
Testing for the generated output from stub_generator.cc will be covered by an integration test that is in progress.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4863)


generator/internal/service_generator.cc
```diff
@@ -0,0 +1,69 @@
...
+namespace cloud {
+namespace generator_internal {
+
+ServiceGenerator::ServiceGenerator(
+    google::protobuf::ServiceDescriptor const* service_descriptor,
+    google::protobuf::compiler::GeneratorContext* context,
+    std::map<std::string, std::string> command_line_vars)
+    : descriptor_(service_descriptor), vars_(std::move(command_line_vars)) {
+  SetVars();
+  class_generators_.push_back(
```
devjgm, line 33: Question: I assume `class_generators_` will eventually have more than one element. How will those other class generators be added to this vector?

generator/internal/service_generator.cc
```diff
@@ -0,0 +1,69 @@
...
+      return result;
+    }
+  }
+  return {};
+}
+
+void ServiceGenerator::SetVars() {
+  vars_["stub_class_name"] = absl::StrCat(descriptor_->name(), "Stub");
+  vars_["stub_header_path"] =
+      absl::StrCat(absl::StrCat(vars_["product_path"], "internal/",
```
devjgm, line 50: Here and below, I don't think nested calls to `StrCat` should be necessary.

generator/internal/service_generator.cc
```diff
@@ -0,0 +1,69 @@
...
+#include "generator/internal/codegen_utils.h"
+#include "generator/internal/stub_generator.h"
+#include <google/api/client.pb.h>
+#include <google/protobuf/descriptor.h>
+
+namespace google {
+namespace cloud {
+namespace generator_internal {
+
+ServiceGenerator::ServiceGenerator(
```
devjgm, line 27: It doesn't look like this class needs to have a dep on `ServiceDescriptor`, nor `GeneratorContext`, nor the map of vars. I think this class only needs to have a dep on the `StubGenerator`s. Consider factoring those extra deps into a factory function for the StubGenerator, which will simplify this class and make it more flexible and easier to test. Specifically consider something like:

```cc
// stub_generator.h

// The implementation would look just like this current c'tor, but it would be in the *stub_generator* file.
std::unique_ptr<StubGenerator> MakeStubGenerator(
    ServiceDescriptor const& service, 
    GeneratorContext& context, 
    std::map<string, string> vars);
```

```cc
// service_generator.h
class ServiceGenerator {
 public:
  explicit ServiceGenerator(std::vector<std::unique_ptr<StubGenerator>> v);
  ...
};
```

generator/internal/service_generator.h
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * This collection of ClassGeneratorInterfaces typically contains generators
+ * for Stub, decorators, Connection, and optionally Client classes. Each of
+ * these ClassGeneratorInterfaces has its substituion variable map seeded with
+ * key value pairs common across the service.
+ */
+class ServiceGenerator {
+ public:
+  ServiceGenerator(
+      google::protobuf::ServiceDescriptor const* service_descriptor,
```
devjgm, line 40: Unless we're saying that `nullptr` is allowed, we should prefer references as much as possible.

generator/internal/stub_generator.cc
```diff
@@ -0,0 +1,134 @@
...
+}
+
+Status StubGenerator::GenerateHeader() {
+  // Temporary conditional for unit testing purposes.
+  if (service_descriptor_->name() == "FailureService") {
+    return Status(StatusCode::kInternal, "Failed for testing.");
+  }
+
+  header_.Print(vars_,  // clang-format off
+    "// Generated by the Codegen C++ plugin.\n"
```
devjgm, line 48: Do generated files need to have copyright comments at the top? I don't know, but I'm wondering. What do grpc's or proto-generated files look like? Those might be different because those aren't checked in.

... Here's one example from checked-in golang code, that does include a copyright header: https://github.com/googleapis/google-cloud-go/blob/master/container/apiv1/cluster_manager_client.go

Maybe we should do that too.

Issue #4783, opened 2020-08-05
Improve build times for full builds


coryan:
Some of the full builds (i.e., not PR builds) are fairly slow, nearing the 2 hour limit (#4782 increases the limit for some to 3 hours).

Generally we do not care about the build times for these, except because:

- If they go over the limit the build fails and that seems like an unforced flake error.
- The full builds refresh the build cache, slow builds means out of date caches for 2-3 hours after a PR is merged.

coryan:
Some ideas on how to solve the problem [1]:

- For CMake-based builds we can simply turn on the ccache for full builds. It is currently turned off to keep the caches small and free of "old stuff". ccache has a simple mechanism to limit the cache size which is already turned on.

- For Bazel-based builds we cannot do the same: the cache grows without limit, so over time we would accumulate more and more stale data in the cache, until (and this is not that far away) the process of uploading and downloading the cache will take longer than the build.

It seems like the "Right Thing":tm: &nbsp; for Bazel builds is to turn on [GCS-based caching](https://docs.bazel.build/versions/master/remote-caching.html#google-cloud-storage) which would allow us to delete anything older than (say) 90d to keep the cache sizes under control.  Seems like we could even use these caches for our local builds too, but all of this requires experimentation to see what the performance implications would be.

[1]: I separated these from the problem description because the ideas for solutions might be bad, but the problem is still real.

mr-salty:
FYI there is https://github.com/bazelbuild/bazel/issues/5139 "local caching: implement --disk_cache_size=GiB" although it's been pending for a long time. I wrote a script to prune the cache on my local machine we could adapt. It's currently based on atime but I could make it use a size threshold.

devjgm:
I'm interested in that idea about using GCS as a remote, shared bazel cache. I'm looking into that.

PR #4869, opened 2020-08-12
ci: enable remote build cache for bazel 

devjgm:
This tells all of our bazel builds to use a remote cache in a GCS
bucket. This will apply to all of our *docker* bazel builds, including
presubmit and CI builds. This should make both builds a bit faster, by
shifting CPU work (compiles) to network traffic (fetching cached
artifacts). This is probably a good trade off for our Kokoro machines.

Note: This PR is only changing thigns for our docker builds so we can
get experience with this `--remote_cache` option on bazel. The
ultimately plan would be to eneable this for macOS and Windows as well.

Related to: #4783

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4869)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -49,6 +49,24 @@ if [[ -n "${BAZEL_CONFIG}" ]]; then
   bazel_args+=("--config" "${BAZEL_CONFIG}")
 fi
 
+readonly TEST_KEY_FILE_JSON="/c/kokoro-run-key.json"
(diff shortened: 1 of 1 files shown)


devjgm:
> FYI: I think we should consider a different service account (and more limited privileges) for Bazel, we do not want Bazel accidentally breaking our tests by changing files in the wrong bucket.
> Likewise, we probably don't want one of the tests or benchmarks accidentally corrupting the cache.
> 
> Maybe open a bug to clean this up?

SGTM. Opened https://github.com/googleapis/google-cloud-cpp/issues/4877 to track this.

PR #4872, opened 2020-08-12
feature(generator): add license/copyright boilerplate for generated files

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4872)


generator/internal/codegen_utils.cc
```diff
@@ -140,6 +140,26 @@ ProcessCommandLineArgs(std::string const& parameters) {
   return command_line_args;
 }
 
+std::string CopyrightLicenseFileHeader() {
+  static std::string const kHeader =  // clang-format off
+  "// Copyright 2020 Google LLC\n"
```
devjgm, line 145: I'm not certain, but I think we may want the copyright year to be the current year. FWIW, you can get the current year with

```cc
#include "absl/time/time.h"
...
std::string year_string = absl::FormatTime("%Y", absl::Now(), absl::UTCTimeZone());
```

generator/internal/codegen_utils.cc
```diff
@@ -140,6 +140,26 @@ ProcessCommandLineArgs(std::string const& parameters) {
   return command_line_args;
 }
 
+std::string CopyrightLicenseFileHeader() {
+  static std::string const kHeader =  // clang-format off
```
devjgm, line 144: It would be better to declare this as `static auto constexpr kHeader`, because then it avoids potential static destruction issues w/ std::string.

generator/internal/codegen_utils.cc
coryan: AFAIK the copyright notice should include the year the file was created, and should not change each time we regenerate the file.  Not sure how that will work with our plans to always regenerate these files and to keep them committed.

coryan: Are you thinking that this tool might be used by third-parties with their own code?  If so, the copyright notice is not appropriate.  Maybe make it optional?

devjgm: I'm not thinking about anyone else using this tool. But if next year we start generating files for a new service, 2020 will be the wrong year.

generator/internal/codegen_utils.cc
```diff
@@ -140,6 +140,26 @@ ProcessCommandLineArgs(std::string const& parameters) {
   return command_line_args;
 }
 
+std::string CopyrightLicenseFileHeader() {
+  static auto constexpr kHeader =  // clang-format off
+  "// Copyright 2020 Google LLC\n"
```
devjgm, line 145: @scotthart @coryan how about we at least file an issue about making sure the copyright year is updated or something correctly?

coryan: That SGTM.

scotthart: #4884 created

PR #4873, opened 2020-08-12
doc(pubsub): document publish batch limits

### LINKED ISSUE #4671: pubsub::BatchingPolicy should use `int` for the maximum message count
We are going to create a proto message with N messages, protobuf only accepts `int` as the index for an array, even if more than `std::numeric_limit<int>::max()` would fit in memory.  We need to guarantee the value of `maximum_message_count()` never exceeds `std::numeric_limit<int>::max()` (or maybe 2^31, whichever is smaller probably).

coryan:
Fixes #4671

As noted in the bug, originally I thought we should change the types, but after further thought, this is more of a documentation problem.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4873)

### DIFF
--- a/google/cloud/pubsub/publisher_options.h
+++ b/google/cloud/pubsub/publisher_options.h
@@ -75,6 +75,7 @@ class PublisherOptions {
   }
 
   std::size_t maximum_message_count() const { return maximum_message_count_; }
+
   /**
    * Set the maximum number of messages in a batch.
    *
@@ -91,6 +92,7 @@ class PublisherOptions {
   }
 
   std::size_t maximum_batch_bytes() const { return maximum_batch_bytes_; }
+
   /**
    * Set the maximum size for the messages in a batch.
    *
--- a/google/cloud/pubsub/publisher_options.h
+++ b/google/cloud/pubsub/publisher_options.h
@@ -75,12 +75,32 @@ class PublisherOptions {
   }
 
   std::size_t maximum_message_count() const { return maximum_message_count_; }
+  /**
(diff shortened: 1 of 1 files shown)


google/cloud/pubsub/publisher_options.h
```diff
@@ -75,12 +75,32 @@ class PublisherOptions {
   }
 
   std::size_t maximum_message_count() const { return maximum_message_count_; }
+  /**
```
devbww, line 78: Nit: The "maximum_hold_time" case inserts a blank line here.

google/cloud/pubsub/publisher_options.h
```diff
@@ -75,12 +75,32 @@ class PublisherOptions {
...
+   *
+   * [pubsub-quota-link]: https://cloud.google.com/pubsub/quotas#resource_limits
+   */
   PublisherOptions& set_maximum_message_count(std::size_t v) {
     maximum_message_count_ = v;
     return *this;
   }
 
   std::size_t maximum_batch_bytes() const { return maximum_batch_bytes_; }
+  /**
```
devbww, line 94: Ditto.

google/cloud/pubsub/publisher_options.h
coryan: Fixed, thanks.

google/cloud/pubsub/publisher_options.h
coryan: Fixed too.

PR #4876, opened 2020-08-13
fix: windows breakage from abseil

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4876)


devjgm:
Tested internally. Looks good.

PR #4885, opened 2020-08-13
ci: don't cache curl build actions

devjgm:
We're seeing strange errors in our CI related to libcurl. It's possible
that we're using some cached artifacts that were compiled w/ the wrong
settings.

See also https://docs.bazel.build/versions/2.0.0/remote-caching.html#exclude-specific-targets-from-using-the-remote-cache

Related to PR https://github.com/googleapis/google-cloud-cpp/pull/4869

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4885)

### DIFF
--- a/bazel/curl.BUILD
+++ b/bazel/curl.BUILD
@@ -44,6 +44,7 @@ genrule(
       fi >$@
     """,
     }),
+    tags = ["no-cache"],
 )
 
 CURL_WIN_COPTS = [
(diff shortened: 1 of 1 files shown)


devjgm:
> Do we need to manually propagate the `no-cache` tags? Or just having it on the first `genrule()` target is enough? For a future PR, there may be a better fix where we define some kind of `config_setting()` that reflects the linux distribution and use that in a `swich()`.

I don't know. I figured just tagging all the build rules in this file was safest, but I'm not sure.

PR #4888, opened 2020-08-13
refactor(generator): remove ServiceGenerator in favor of free functions

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4888)


generator/internal/codegen_utils_test.cc
```diff
@@ -13,13 +13,19 @@
...
 #include "generator/internal/codegen_utils.h"
+#include <google/protobuf/descriptor.pb.h>
 #include <gmock/gmock.h>
 
 namespace google {
 namespace cloud {
 namespace generator_internal {
 namespace {
 
+using google::protobuf::DescriptorPool;
```
devjgm, line 24: Style: In using declarations, prefer to use fully qualify, e.g., `using ::foo::Bar;`

PR #4895, opened 2020-08-14
feature(generator): add integration test framework and test stub header

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4895)


generator/integration_tests/generator_integration_test.cc
```diff
@@ -0,0 +1,138 @@
...
+
+    // Path to find .proto files defined for these tests.
+    auto code_path =
+        google::cloud::internal::GetEnv("GOOGLE_CLOUD_CPP_GENERATOR_CODE_PATH")
+            .value_or("/v");
+
+    // Path to location where generated code is written.
+    auto output_path = google::cloud::internal::GetEnv(
+                           "GOOGLE_CLOUD_CPP_GENERATOR_OUTPUT_PATH")
+                           .value_or("/tmp");
```
devjgm, line 80: I think `::testing::TempDir()` is available externally. That'd be what we want.

generator/integration_tests/generator_integration_test.cc
```diff
@@ -0,0 +1,138 @@
...
+    args.emplace_back("--proto_path=" + proto_path);
+    args.emplace_back("--proto_path=" + googleapis_path);
+    args.emplace_back("--proto_path=" + code_path);
+    args.emplace_back("--cpp_codegen_out=" + output_path);
+    args.emplace_back("--cpp_codegen_opt=product_path=google/cloud/test/");
+    args.emplace_back("generator/integration_tests/test.proto");
+
+    c_args_.reserve(args.size());
+    for (auto const& arg : args) {
+      std::cout << "args : " << arg << std::endl;
```
devjgm, line 99: Nit: Prefer "\n".
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rio-endl

generator/integration_tests/generator_integration_test.cc
```diff
@@ -0,0 +1,138 @@
...
+      c_args_.push_back(arg.c_str());
+    }
+  }
+
+  int GenerateOnce() {
```
devjgm, line 104: Consider inlining the body of this function in the `SetUp` function. I think that may simplify the code and semantics. That will also let you get rid of the data members in this fixture, which is nice.

mr-salty:
___
*[generator/integration_tests/generator_integration_test.cc, line 104 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4895#-MEwxXY51d1U7tkS2OQg:-MExPp7bD10DRUhsYdum:brqx6mv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a0e866c7f75e569239e1456f52d9705ba2c04d2b/generator/integration_tests/generator_integration_test.cc#L104)):*

Just a passing comment, but the "should we run integration tests" logic lives in a shell script for other cases.

you could set a static bool in `SetUpTestCase` and have `SetUp` just do `if (skip_all_tests_) GTEST_SKIP();`
</blockquote></details>

Also, consider filing an issue like https://github.com/google/googletest/issues/247 for `GTEST_SKIP`

PR #4910, opened 2020-08-14
feature(generator): add predicate utils to be used in method generation

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4910)

### DIFF
--- a/generator/internal/predicate_utils.cc
+++ b/generator/internal/predicate_utils.cc
@@ -1,0 +1,101 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "generator/internal/predicate_utils.h"
+#include "google/cloud/optional.h"
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace generator_internal {
+
+using ::google::protobuf::Descriptor;
+using ::google::protobuf::FieldDescriptor;
+using ::google::protobuf::MethodDescriptor;
+
+// https://google.aip.dev/client-libraries/4233
+google::cloud::optional<std::pair<std::string, std::string>>
+DeterminePagination(google::protobuf::MethodDescriptor const* method) {
+  std::string paginated_type;
+  Descriptor const* request_message = method->input_type();
+  FieldDescriptor const* page_size =
+      request_message->FindFieldByName("page_size");
+  if (!page_size || page_size->type() != FieldDescriptor::TYPE_INT32) return {};
+  FieldDescriptor const* page_token =
+      request_message->FindFieldByName("page_token");
+  if (!page_token || page_token->type() != FieldDescriptor::TYPE_STRING)
+    return {};
+
+  Descriptor const* response_message = method->output_type();
+  FieldDescriptor const* next_page_token =
+      response_message->FindFieldByName("next_page_token");
+  if (!next_page_token ||
+      next_page_token->type() != FieldDescriptor::TYPE_STRING)
+    return {};
+
+  std::vector<std::tuple<std::string, Descriptor const*, int>>
+      repeated_message_fields;
+  std::cout << "response_message->field_count(): "
+            << response_message->field_count() << "\n";
+  for (int i = 0; i < response_message->field_count(); ++i) {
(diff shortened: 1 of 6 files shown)


generator/internal/predicate_utils.h
```diff
@@ -0,0 +1,194 @@
...
+DeterminePagination(google::protobuf::MethodDescriptor const* method);
+
+/**
+ * Returns true if all predicates return true.
+ */
+template <typename T>
+class GenericAll {
+ public:
+  template <typename... Predicates>
+  explicit GenericAll(Predicates... p) : predicates_({p...}) {}
```
devjgm, line 65: I think we should use perfect forwarding here:

```cc
template <typename... Predicate>
explicit GenericAll(Predicate&&... p) : predicates_({std::forward<Predicate>(p)...}) {}
```

generator/internal/predicate_utils.h
```diff
@@ -0,0 +1,194 @@
...
+             std::function<bool(T const* m)> rhs)
+      : lhs_(std::move(lhs)), rhs_(std::move(rhs)) {}
+  bool operator()(T const* m) const { return lhs_(m) && rhs_(m); }
+
+ private:
+  std::function<bool(T const* m)> lhs_;
+  std::function<bool(T const* m)> rhs_;
+};
+
+using And = GenericAnd<google::protobuf::MethodDescriptor>;
```
devjgm, line 119: Do we need `And` and `All`? It seems like `And` is just the case where `All` is given two predicates. Maybe just renamed All -> And, and then delete And?

Similar comment about Or and Any.

generator/internal/predicate_utils.cc
```diff
@@ -0,0 +1,101 @@
...
+  Descriptor const* response_message = method->output_type();
+  FieldDescriptor const* next_page_token =
+      response_message->FindFieldByName("next_page_token");
+  if (!next_page_token ||
+      next_page_token->type() != FieldDescriptor::TYPE_STRING)
+    return {};
+
+  std::vector<std::tuple<std::string, Descriptor const*, int>>
+      repeated_message_fields;
+  std::cout << "response_message->field_count(): "
```
devjgm, line 49: Was this cout accidentally left in here? In general, we discourage printing output from libraries. My suggestion would be to remove the library output, or replace it with calls to `GCP_LOG` or something.
WDYT?

generator/internal/predicate_utils.cc
```diff
@@ -0,0 +1,101 @@
...
+  if (repeated_message_fields.size() > 1) {
+    auto min_field = std::min_element(
+        repeated_message_fields.begin(), repeated_message_fields.end(),
+        [](std::tuple<std::string, Descriptor const*, int> const& lhs,
+           std::tuple<std::string, Descriptor const*, int> const& rhs) {
+          return std::get<2>(lhs) < std::get<2>(rhs);
+        });
+    int min_field_number = std::get<2>(*min_field);
+    if (min_field_number != std::get<2>(repeated_message_fields[0])) {
+      std::cerr << "Repeated field in paginated response must be first "
```
devjgm, line 73: Perhaps `GCP_LOG(FATAL)` here instead?

devjgm, line 73: It's a somewhat surprising requirement that any field is required to have the lowest field number. Can you add a comment linking to the doc or whatever that talks about why this is in case we wonder about this again in the future?

generator/internal/predicate_utils.h
```diff
@@ -0,0 +1,194 @@
...
+namespace google {
+namespace cloud {
+namespace generator_internal {
+
+/**
+ * Determines if the given method meets the criteria for pagintation.
+ *
+ * https://google.aip.dev/client-libraries/4233
+ */
+bool IsPaginated(google::protobuf::MethodDescriptor const* method);
```
devjgm, line 31: (optional) General comment: I think you should consider switching to const-refs instead of pointers all over (other than where you're overriding a method that requires a pointer). The grpc and proto style of using pointers (a choice made ~20 years ago) is no longer a best practice. Unless we _must_ use a pointer, or want _pointer semantics_ (allow nullptr, disallow binding to temporaries, etc.), I think `const&` is a better choice for non-optional input parameters than is `const*`. Using const-refs would be more consistent with our existing code, and our current style guides. Consistency with grpc and protobuf APIs is a non-goal.

generator/internal/predicate_utils_test.cc
```diff
@@ -0,0 +1,471 @@
...
+  DescriptorPool pool;
+  const FileDescriptor* service_file_descriptor = pool.BuildFile(service_file);
+  EXPECT_DEATH(
+      IsPaginated(service_file_descriptor->service(0)->method(0)),
+      "Repeated field in paginated response must be first appearing and lowest "
+      "field number: google.protobuf.Service.RepeatedOrderMismatch");
+}
+
+TEST(PredicateUtilsTest, PredicatedFragmentTrueString) {
+  PredicatedFragment<int> f = {[](int const*) { return true; }, "True",
```
devjgm, line 452: Use `PredicateTrue`?

PR #4911, opened 2020-08-15
ci: make libcurl cacheable on Linux builds

### LINKED ISSUE #4902: Fix bazel/curl.BUILD to be cacheable
> Do we need to manually propagate the `no-cache` tags? Or just having it on the first `genrule()` target is enough? For a future PR, there may be a better fix where we define some kind of `config_setting()` that reflects the linux distribution and use that in a `swich()`.

I don't know. I figured just tagging all the build rules in this file was safest, but I'm not sure.

_Originally posted by @devjgm in https://github.com/googleapis/google-cloud-cpp/pull/4885#issuecomment-673646273_

Today `bazel/curl.BUILD` marks all those targets with 'no-cache' to avoid breakages we saw in the CI. It would be great if we could do something about that build file to make the targets more cacheable.

coryan:
Make libcurl cacheable on CI builds, though it remains very much
not cache-friendly for manual builds. The build uses a `--define`
Bazel flag to configure the `CURL_CA_BUNDLE` macro, if the
`--define` flag is not provided we default to guessing the location
as before.

I think this is a reasonable tradeoff: the build works by default,
without special flags, but it is not cache-friendly. The CI builds
work more efficiently, but require a special flag.

Fixes #4902 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4911)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -49,6 +49,12 @@ if [[ -n "${BAZEL_CONFIG}" ]]; then
   bazel_args+=("--config" "${BAZEL_CONFIG}")
 fi
 
+if [[ -r "/etc/pki/tls/certs/ca-bundle.crt" ]]; then
+  bazel_args+=("--define" "ca_bundle_style=redhat")
+elif [[ -r "/etc/ssl/certs/ca-certificates.crt" ]]; then
+  bazel_args+=("--define" "ca_bundle_style=debian")
+fi
+
 readonly TEST_KEY_FILE_JSON="/c/kokoro-run-key.json"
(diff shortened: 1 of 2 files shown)


ci/kokoro/docker/build-in-docker-bazel.sh
```diff
@@ -75,6 +81,7 @@ io::log "Fetching dependencies"
 
 echo "================================================================"
 io::log "Compiling and running unit tests"
+echo "bazel test" "${bazel_args[@]}"
```
devbww, line 84: Debugging?

coryan: Yes, but if that is Okay I think we should leave it.  It may be useful in the future when we are trying to understand what the build is actually doing.

PR #4914, opened 2020-08-15
ci: disable bazel tarball cache on Windows

coryan:
Now that we have the GCS cache working, and the libcurl library is
cacheable on Windows we should be able to disable the tarball cache
altogether.

The windows/bazel build reports:

```
INFO: 3121 processes: 3116 remote cache hit, 5 local.
```

https://source.cloud.google.com/results/invocations/4df0be5f-6a84-4874-8e6a-ac49b257ccd6

That seems like a good hit rate :slightly_smiling_face: many of the tests
did not need to run because they reused the cache.

The windows/bazel-release build reports:

```
INFO: 3121 processes: 2331 remote cache hit, 790 local.
```

https://source.cloud.google.com/results/invocations/b36ac25e-cb99-4b25-8c74-cd990c04fd77

That is not nearly as good. OTOH, that build has been all over the place for the last few days.

### DIFF
--- a/ci/kokoro/windows/build-bazel.ps1
+++ b/ci/kokoro/windows/build-bazel.ps1
@@ -78,54 +78,6 @@ if (Test-Path env:TEMP) {
 }
 Write-Host -ForegroundColor Yellow "`n$(Get-Date -Format o) Using ${download_dir} as download directory"
 
-# Download and extract the cache tarball
-if ($RunningCI -and $IsPR -and $CacheConfigured -and $Has7z) {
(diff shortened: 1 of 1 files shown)


coryan:
The second run was under 15m for both:

https://source.cloud.google.com/results/invocations/01b48f7f-fc50-41ef-a90f-a671c4cb5350
https://source.cloud.google.com/results/invocations/fff9a3cb-bd6e-4924-bcb7-b1c74f5cfaf2

PR #4933, opened 2020-08-19
ci: remove inconvenient convenience symlinks

coryan:
Bazel creates "convenience" links in %workspace% pointing to the
location of several interesting directories. These links are wrong when
created inside a Docker container, which is the case for most of our
builds, better to not generate the links in the first place.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4933)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -17,6 +17,10 @@
 #   [1]: https://github.com/bazelbuild/bazel/issues/4341
 build --copt=-DGRPC_BAZEL_BUILD
 
+# Do not create the convenience links, they are incovenient when the build runs
+# inside a docker image.
+build --experimental_convenience_symlinks=ignore
+
 # Clang Sanitizers, use with (for example):
 #
 # --client_env=CXX=clang++ --client_env=CC=clang --config asan


mr-salty:
I do like the title but I also do unadorned `bazel build` and occasionally use the links... 

would adding the flag [here](https://github.com/googleapis/google-cloud-cpp/blob/928b185bd028c09d74d917893639f839ad8b1983/ci/kokoro/docker/build-in-docker-bazel.sh#L43) instead work?

coryan:
I also do the `bazel build` and `bazel test` thing, and the symlinks are a problem for me in that case too. Turned off seems like a better default. However...

> I do like the title but I also do unadorned `bazel build` 

Which will continue to work.

> and occasionally use the links...

Which won't.

Presumably you can use `bazel build --experimental_convenience_symlinks=normal`  and it overrides the value in the `.bazelrc` file?   And maybe set in in `$HOME/.bazelrc`.

> would adding the flag [here](https://github.com/googleapis/google-cloud-cpp/blob/928b185bd028c09d74d917893639f839ad8b1983/ci/kokoro/docker/build-in-docker-bazel.sh#L43) instead work?

Don't know, can you test and send a PR? Remember to add them to the macOS, and Windows scripts too.

PR #4936, opened 2020-08-19
feat: automatically create background thread pools

### LINKED ISSUE #4650: g::c::ConnectionOptions: option to create multiple background threads
It would be useful if ConnectionOptions gave the user a way to create N background threads. Currently the application can do so, but they need to manually create and manage the thread pool, that seems boring.

Note that this is a change for `google::cloud::ConnectionOptions`.

coryan:
Make the size of the background thread pool created by
google::cloud::ConnectionOptions` configurable. A future change will
set the default size to something more sensible than "one", though this
would need to be by service, for example, a subscription or publisher
*should* use several threads, but the admin APIs are probably fine with
one.

Fixes #4650

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4936)

### DIFF
--- a/google/cloud/internal/background_threads_impl.cc
+++ b/google/cloud/internal/background_threads_impl.cc
@@ -25,13 +25,12 @@ AutomaticallyCreatedBackgroundThreads::AutomaticallyCreatedBackgroundThreads()
 
 AutomaticallyCreatedBackgroundThreads::AutomaticallyCreatedBackgroundThreads(
     std::size_t thread_count)
-    : AutomaticallyCreatedBackgroundThreads(
-          thread_count == 0 ? 1 : thread_count, {}) {}
+    : AutomaticallyCreatedBackgroundThreads(thread_count, {}) {}
 
 AutomaticallyCreatedBackgroundThreads::AutomaticallyCreatedBackgroundThreads(
-    std::size_t thread_count, Normalized)
-    : pool_(thread_count) {
-  std::generate_n(pool_.begin(), thread_count, [this] {
+    std::size_t thread_count, NormalizeTag)
+    : pool_(thread_count == 0 ? 1 : thread_count) {
+  std::generate_n(pool_.begin(), pool_.size(), [this] {
     return std::thread([](CompletionQueue cq) { cq.Run(); }, cq_);
   });
 }
--- a/google/cloud/internal/background_threads_impl.cc
+++ b/google/cloud/internal/background_threads_impl.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/internal/background_threads_impl.h"
+#include <algorithm>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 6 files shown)


google/cloud/internal/background_threads_impl.cc
```diff
@@ -13,14 +13,28 @@
...
+    : AutomaticallyCreatedBackgroundThreads(1, {}) {}
+
+AutomaticallyCreatedBackgroundThreads::AutomaticallyCreatedBackgroundThreads(
+    std::size_t thread_count)
+    : AutomaticallyCreatedBackgroundThreads(
+          thread_count == 0 ? 1 : thread_count, {}) {}
+
+AutomaticallyCreatedBackgroundThreads::AutomaticallyCreatedBackgroundThreads(
+    std::size_t thread_count, Normalized)
+    : pool_(thread_count) {
```
devbww, line 33: Is this where the `thread_count == 0 ? 1 : thread_count` would be better applied?

google/cloud/internal/background_threads_impl.h
```diff
@@ -42,14 +43,18 @@ class CustomerSuppliedBackgroundThreads : public BackgroundThreads {
...
   AutomaticallyCreatedBackgroundThreads();
+  explicit AutomaticallyCreatedBackgroundThreads(std::size_t thread_count);
   ~AutomaticallyCreatedBackgroundThreads() override;
 
   CompletionQueue cq() const override { return cq_; }
   void Shutdown();
+  std::size_t pool_size() const { return pool_.size(); }
 
  private:
+  struct Normalized {};
```
devbww, line 54: Nit: I think we'd typically use some variant of "tag" for the name of this type.

google/cloud/internal/background_threads_impl_test.cc
```diff
@@ -72,12 +74,44 @@ TEST(CustomerSuppliedBackgroundThreads, SharesCompletionQueue) {
...
+    actual.cq().RunAsync([&p] {
+      p.set_value(std::this_thread::get_id());
+      // Keep the thread busy for a bit so the other functors schedule in a
+      // different thread.
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    });
+  }
+  std::set<std::thread::id> ids;
+  for (auto& p : promises) ids.insert(p.get_future().get());
+  EXPECT_LE(2, ids.size());
```
devbww, line 111: While this is a good objective for a four-thread pool, it may be flaky, and also seems beside the point of "nothing run on the main thread."

google/cloud/internal/background_threads_impl.cc
coryan: Done.

google/cloud/internal/background_threads_impl.h
coryan: Done, I went with `NormalizeTag` because it is now saying (to me) "please normalize" whereas before it was saying "trust me the value is normalized".

google/cloud/internal/background_threads_impl_test.cc
coryan: Fair enough. Fixed.  FWIW, I did ran it thousands of times to avoid flakiness.

coryan:
PTAL.

PR #4937, opened 2020-08-19
cleanup(pubsub): more familiar idioms in samples

coryan:
The subscription samples need to signal a waiting thread when something
(typically the first message) happens. This change uses condition
variables to do this, maybe less elegant, but easier to understand for
most C++ developers.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4937)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -19,7 +19,6 @@
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/internal/random.h"
 #include "google/cloud/testing_util/example_driver.h"
-#include <atomic>
 #include <tuple>
 #include <utility>
 
@@ -520,31 +519,28 @@ void Subscribe(google::cloud::pubsub::Subscriber subscriber,
   [](pubsub::Subscriber subscriber, pubsub::Subscription const& subscription) {
     std::mutex mu;
     std::condition_variable cv;
-    bool done = false;
     int message_count = 0;
     auto session = subscriber.Subscribe(
         subscription, [&](pubsub::Message const& m, pubsub::AckHandler h) {
-          std::cout << "Received message " << m << std::endl;
+          std::cout << "Received message " << m << "\n";
           std::unique_lock<std::mutex> lk(mu);
           ++message_count;
-          done = true;
           lk.unlock();
           cv.notify_one();
           std::move(h).ack();
         });
     // Wait until at least one message has been received.
     std::unique_lock<std::mutex> lk(mu);
-    cv.wait(lk, [&done] { return done; });
+    cv.wait(lk, [&message_count] { return message_count > 0; });
     lk.unlock();
     // Cancel the subscription session.
     session.cancel();
     // Wait for the session to complete, no more callbacks can happen after this
     // point.
-    std::cout << "GET=" << done << std::endl;
     auto status = session.get();
     // Report any final status, blocking.
     std::cout << "Message count: " << message_count << ", status: " << status
-              << std::endl;
+              << "\n";
   }
   //! [END pubsub_subscriber_async_pull] [subscribe]
   //! [END pubsub_quickstart_subscriber]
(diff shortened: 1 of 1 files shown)


google/cloud/pubsub/samples/samples.cc
```diff
@@ -518,27 +518,33 @@ void Subscribe(google::cloud::pubsub::Subscriber subscriber,
...
+    session.cancel();
+    // Wait for the session to complete, no more callbacks can happen after this
+    // point.
+    std::cout << "GET=" << done << std::endl;
+    auto status = session.get();
     // Report any final status, blocking.
-    std::cout << "Message count = " << count.load()
-              << ", status = " << result.get() << "\n";
+    std::cout << "Message count: " << message_count << ", status: " << status
+              << std::endl;
```
devbww, line 547: Examples below just use `"\n"` here.

google/cloud/pubsub/samples/samples.cc
```diff
@@ -518,27 +518,33 @@ void Subscribe(google::cloud::pubsub::Subscriber subscriber,
...
-        .get();
+    // Wait until at least one message has been received.
+    std::unique_lock<std::mutex> lk(mu);
+    cv.wait(lk, [&done] { return done; });
+    lk.unlock();
+    // Cancel the subscription session.
+    session.cancel();
+    // Wait for the session to complete, no more callbacks can happen after this
+    // point.
+    std::cout << "GET=" << done << std::endl;
```
devbww, line 543: Debugging?  It seems we know `done == true` here in any case.

google/cloud/pubsub/samples/samples.cc
```diff
@@ -518,27 +518,33 @@ void Subscribe(google::cloud::pubsub::Subscriber subscriber,
   using google::cloud::future;
   using google::cloud::StatusOr;
   [](pubsub::Subscriber subscriber, pubsub::Subscription const& subscription) {
-    std::atomic<int> count{0};
```
devbww, line 521: It looks like `<atomic>` is now unused.

mr-salty:
___
*[google/cloud/pubsub/samples/samples.cc, line 537 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4937#-MFBo0AWCYtjTg6l_uUg:-MFBo0AWCYtjTg6l_uUh:bl50t1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/948e285a0aeab340005c393bcc231a1f755f5a6d/google/cloud/pubsub/samples/samples.cc#L537)):*
> ```cpp
>     // Wait until at least one message has been received.
>     std::unique_lock<std::mutex> lk(mu);
>     cv.wait(lk, [&done] { return done; });
> ```

eliminate `done` and just use `message_count > 0` as the condition?

well, it looks like bww and I are both reviewing the same PR (joy), so I'll defer to him.

google/cloud/pubsub/samples/samples.cc
coryan: Sorry, I left those from earlier debugging, fixed.

google/cloud/pubsub/samples/samples.cc
coryan: Removed.

google/cloud/pubsub/samples/samples.cc
coryan: Removed.

coryan:
___
*[google/cloud/pubsub/samples/samples.cc, line 537 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4937#-MFBo0AWCYtjTg6l_uUg:-MFBszjtAeCQUR6VUc2F:bueut62) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/948e285a0aeab340005c393bcc231a1f755f5a6d/google/cloud/pubsub/samples/samples.cc#L537)):*

eliminate `done` and just use `message_count > 0` as the condition?

well, it looks like bww and I are both reviewing the same PR (joy), so I'll defer to him.
</blockquote></details>

I like it, done.

coryan:
PTAL

PR #4947, opened 2020-08-21
feat(pubsub): better defaults for thread pool size

### LINKED ISSUE #4946: Pick better defaults for the number of background threads
We need to have better defaults for `pubsub::Publisher` and `pubsub::Subscriber`, a single thread is not enough for them.
It should be something proportional to the number of cores.

coryan:
For `pubsub::Publisher` and `pubsub::Subscriber` we need larger thread
pool sizes by default. If `std::thread::hardware_concurrency()` returns
`0` we use 4 threads, otherwise using the value returned by this
function, this is aligned with the libraries for other languages.

Fixes #4946

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4947)

### DIFF
--- a/google/cloud/pubsub/publisher_connection.cc
+++ b/google/cloud/pubsub/publisher_connection.cc
@@ -49,11 +49,11 @@ PublisherConnection::~PublisherConnection() = default;
 
 std::shared_ptr<PublisherConnection> MakePublisherConnection(
     Topic topic, PublisherOptions options,
-    ConnectionOptions const& connection_options) {
+    ConnectionOptions connection_options) {
   auto stub = pubsub_internal::CreateDefaultPublisherStub(connection_options,
                                                           /*channel_id=*/0);
   return pubsub_internal::MakePublisherConnection(
-      std::move(topic), std::move(options), connection_options,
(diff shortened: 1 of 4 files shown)


google/cloud/pubsub/publisher_connection.cc
```diff
@@ -65,14 +65,25 @@ inline namespace GOOGLE_CLOUD_CPP_PUBSUB_NS {
...
     std::shared_ptr<PublisherStub> stub) {
   stub = std::make_shared<pubsub_internal::PublisherMetadata>(std::move(stub));
   if (connection_options.tracing_enabled("rpc")) {
     GCP_LOG(INFO) << "Enabled logging for gRPC calls";
     stub = std::make_shared<pubsub_internal::PublisherLogging>(
         std::move(stub), connection_options.tracing_options());
   }
+
+  auto default_thread_pool_size = []() -> std::size_t {
+    auto constexpr kMinThreadPoolSize = 4;
```
devbww, line 78: s/Min/Default/ perhaps (and below)?

PR #4958, opened 2020-08-24
cleanup: document RetryPolicy base class

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4958)

### DIFF
--- a/google/cloud/internal/retry_policy.h
+++ b/google/cloud/internal/retry_policy.h
@@ -26,13 +26,18 @@ inline namespace GOOGLE_CLOUD_CPP_NS {
 namespace internal {
 /**
  * Define the interface for retry policies.
- *
  */
 class RetryPolicy {
  public:
   virtual ~RetryPolicy() = default;
+
+  /// Handle a failure, return `true` if the retry loop can continue.
   virtual bool OnFailure(Status const&) = 0;
+
+  /// Return `true` if no more retries should be attempted.
   virtual bool IsExhausted() const = 0;
+
+  /// Return `true` if the error is not retryable.
   virtual bool IsPermanentFailure(Status const&) const = 0;
 };
 
--- a/google/cloud/internal/retry_policy.h
+++ b/google/cloud/internal/retry_policy.h
@@ -31,14 +31,37 @@ class RetryPolicy {
  public:
   virtual ~RetryPolicy() = default;
 
-  /// Handle a failure, return `true` if the retry loop can continue.
+  //@{
+  /**
+   * @name Control retry loop duration.
+   *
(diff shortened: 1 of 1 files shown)


google/cloud/internal/retry_policy.h
```diff
@@ -26,13 +26,18 @@ inline namespace GOOGLE_CLOUD_CPP_NS {
 namespace internal {
 /**
  * Define the interface for retry policies.
- *
  */
 class RetryPolicy {
  public:
   virtual ~RetryPolicy() = default;
+
+  /// Handle a failure, return `true` if the retry loop can continue.
```
devbww, line 34: `s/, return/. Return./`

google/cloud/internal/retry_policy.h
```diff
@@ -26,13 +26,18 @@ inline namespace GOOGLE_CLOUD_CPP_NS {
...
   virtual bool OnFailure(Status const&) = 0;
+
+  /// Return `true` if no more retries should be attempted.
   virtual bool IsExhausted() const = 0;
+
+  /// Return `true` if the error is not retryable.
```
devbww, line 40: While any documentation on an interface is probably better than none, I find it difficult to distinguish "_return `true` if the retry loop can continue_" from "_return `true` if no more retries should be attempted_" and "_return `true` if the error is not retryable._"  They all sound like the same thing.

coryan: Yes, that is a consequence of how these are intended to be used, and are probably code smells.... basically the loop is like this (paraphrasing):

```
while (!policy->IsExhausted()) {
  auto status = try_rpc();
  if (status.ok()) return status;
  if (!OnFailure(status)) return ModifyError(status);
  // sleep, which may exhaust the policy, even if it was not exhausted in the last call. 
}
return SomeError();
}
```

Happy to improve the descriptions if you have suggestions.

devbww: Perhaps, then, the documentation might be better with the three functions grouped under a single comment, where that comment shows how they will be called in the retry loop?  That might be the easiest way to distinguish their roles.

The example loop you gave looks good, for example, although the `OnFailure(status)` is missing its `policy->`, and `IsPermanentFailure()` isn't mentioned.

google/cloud/internal/retry_policy.h
coryan: Removed.

google/cloud/internal/retry_policy.h
coryan: PTAL

devbww: Much clearer, I reckon.  Thanks.

Issue #4971, opened 2020-08-27
Define retry policy and implement them for `Publisher::Publish()`


coryan:
This is a non-idempotent operation, but apparently all the other libraries retry these calls, we should do the same, or at least offer an option to do the same.

PR #4972, opened 2020-08-27
feat: retry loop for asynchronous RPCs

coryan:
In the Cloud Pub/Sub library (and I think in all future libraries) the
signature for an asynchronous RPC is:

```cc
future<StatusOr<ResponseProto> AsyncFooBar(
    CompletionQueue& cq, std::unique_ptr<grpc::ClientContext> ctx,
    RequestProto const& request);
```

This PR introduces a retry loop for functions with that signature,
previous retry loops required a different return type and arguments. The
older signature was really hard to mock and test.

Part of the work for #4971 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4972)

### DIFF
--- a/google/cloud/CMakeLists.txt
+++ b/google/cloud/CMakeLists.txt
@@ -342,6 +342,7 @@ if (GOOGLE_CLOUD_CPP_ENABLE_GRPC)
         grpc_utils/grpc_error_delegate.h
         grpc_utils/version.h
         internal/async_read_stream_impl.h
+        internal/async_retry_loop.h
         internal/async_retry_unary_rpc.h
         internal/background_threads_impl.cc
         internal/background_threads_impl.h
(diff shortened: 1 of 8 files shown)


coryan:
___
*[google/cloud/internal/async_retry_loop.h, line 162 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4972#-MFl4fVf5n6JitTDD4E1:-MFlA-g26el-y4E1uYUn:bbo0odb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce86d9143babfff9492db4567e8d408a5a74b6a2/google/cloud/internal/async_retry_loop.h#L162)):*

Hmm... future cleanup?  I am not going to change `RetryLoop()` in this PR too, we have hundreds of call sites already for that.
</blockquote></details>

#4975

PR #4976, opened 2020-08-27
feat(pubsub): optionally retry Publisher::Publish

### LINKED ISSUE #4971: Define retry policy and implement them for `Publisher::Publish()`

coryan:
Cloud Pub/Sub libraries are expected to retry `Publish()` requests.
Because these are not idempotent I chose to disable retries by default,
but they can be turned no relatively easily.

Fixes #4971

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4976)

### DIFF
--- a/google/cloud/pubsub/publisher.h
+++ b/google/cloud/pubsub/publisher.h
@@ -113,11 +113,13 @@ class Publisher {
    * Publishes a message to this publisher's topic
    *
    * Note that the message maybe be batched based on the Publisher's
-   * configuration, the message may not be immediately sent.
+   * configuration. It could be delayed until the batch has enough messages,
+   * or enough data, or enough time has elapsed. See the `PublisherOptions`
+   * documentation for more details.
    *
    * @par Idempotency
-   * This is a non-idempotent operation, by default the request is *not* retried
-   * if there is a transient error. Applications wanting to enable retries can
+   * This is a non-idempotent operation. By default the request is *not* retried
+   * if there is a transient error. Applications wanting to retry requests can
    * do so by enabling the `retry_publish_failures()` parameter in the
    * `PublisherOptions`.
    *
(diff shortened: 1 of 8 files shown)


google/cloud/pubsub/publisher.h
```diff
@@ -115,6 +115,12 @@ class Publisher {
    * Note that the message maybe be batched based on the Publisher's
    * configuration, the message may not be immediately sent.
    *
+   * @par Idempotency
+   * This is a non-idempotent operation, by default the request is *not* retried
```
devbww, line 119: s/, by/. By/ perhaps?

coryan: Done.

google/cloud/pubsub/publisher.h
```diff
@@ -113,11 +113,13 @@ class Publisher {
    * Publishes a message to this publisher's topic
    *
    * Note that the message maybe be batched based on the Publisher's
```
devbww, line 115: s/maybe be/may be/

coryan: Ooops, thanks!

google/cloud/pubsub/publisher.h
```diff
@@ -112,7 +112,7 @@ class Publisher {
   /**
    * Publishes a message to this publisher's topic
    *
-   * Note that the message maybe be batched based on the Publisher's
+   * Note that the message may be be batched based on the Publisher's
```
devbww, line 115: s/may be be/may be/  :-)

coryan: LOL

PR #4981, opened 2020-08-28
fix: a few errors in PubSub README.md example

agasparovic-sabre:
* Remove catch without a try block (and eliminate use of exception to print error)
* `topic-id` misspelling

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4981)

### DIFF
--- a/google/cloud/pubsub/README.md
+++ b/google/cloud/pubsub/README.md
@@ -45,9 +45,9 @@ library.
 int main(int argc, char* argv[]) try {
   if (argc != 3) {
     std::string const cmd = argv[0];
-    auto last_slash = std::string(argv[0]).find_last_of('/');
+    auto last_slash = cmd.find_last_of('/');
     std::cerr << "Usage: " << cmd.substr(last_slash + 1)
-              << " <project-id> <topid-ic>\n";
+              << " <project-id> <topic-id>\n";
     return 1;
   }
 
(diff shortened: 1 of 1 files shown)


coryan:
Thanks for sending a PR!  Unfortunately, this README.md file is generated and the source of truth for the code snippet is `pubsub/quickstart/quickstart.cc`.  I can help you figure out to change the source of truth and run the script, or I can apply these fixes for you.. What do you prefer?  I am happy to do either.

> * Remove catch without a try block 

There is a `try {` block here:

https://github.com/googleapis/google-cloud-cpp/blob/d5063ca3f868766e7b511d81f47bd854615b0fb8/google/cloud/pubsub/quickstart/quickstart.cc#L20

hard to spot maybe?

> (and eliminate use of exception to print error)

In general, we prefer to make the examples shorter by throwing on an error: detailed error handling gets in the way of understanding the examples.  This example is (maybe blindly) following that pattern. We know most C++ programmers will want to do something more than just throw, but we expect most will understand what is going on and change the example to fit their needs.

In short, I think we like the code snippet as it is.

> * `topic-id` misspelling

Ah, yes, we should fix that....

agasparovic-sabre:
Fair enough - thanks for the explanation Carlos! If you have the time, I'll take your offer to apply the typo fix. BTW, I also had an additional fix I didn't list - I changed `std::string(argv[0])` to `cmd` (it's on line 23 of the quickstart.cc).

Looking forward to spending more time trying out the library!

coryan:
@agasparovic-sabre just created a new PR to fix this. I noticed the `argv[0]` changes, I chose to remove all the code to simplify the example even more. Look forward to your feedback about the library.

PR #4987, opened 2020-08-30
test: make test utility class thread-safe

coryan:
For Cloud Pub/Sub I need `CaptureLogLinesBackend` to be thread-safe.
Cloud Pub/Sub launches (and uses) multiple threads behind the scenes,
and TSAN detected some racy code in the tests. Most of our tests are
single-threaded, and could get away with a thread-unsafe backend, I
guess the happy days are over.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4987)

### DIFF
--- a/google/cloud/spanner/internal/logging_result_set_reader_test.cc
+++ b/google/cloud/spanner/internal/logging_result_set_reader_test.cc
@@ -80,8 +80,6 @@ TEST_F(LoggingResultSetReaderTest, Read) {
   EXPECT_THAT(log_lines, Contains(HasSubstr("Read")));
   EXPECT_THAT(log_lines, Contains(HasSubstr("test-token")));
 
-  // Clear previous captured lines to ensure the following checks for new
-  // messages.
   result = reader.Read();
   ASSERT_FALSE(result.has_value());
   log_lines = ClearLogLines();
--- a/google/cloud/spanner/internal/logging_result_set_reader_test.cc
+++ b/google/cloud/spanner/internal/logging_result_set_reader_test.cc
@@ -45,11 +45,7 @@ class LoggingResultSetReaderTest : public ::testing::Test {
     logger_id_ = 0;
   }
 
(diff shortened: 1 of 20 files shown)


google/cloud/spanner/internal/logging_result_set_reader_test.cc
```diff
@@ -80,15 +76,16 @@ TEST_F(LoggingResultSetReaderTest, Read) {
...
   EXPECT_EQ("test-token", result->resume_token());
 
-  HasLogLineWith("Read");
-  HasLogLineWith("test-token");
+  auto log_lines = ClearLogLines();
+  EXPECT_THAT(log_lines, Contains(HasSubstr("Read")));
+  EXPECT_THAT(log_lines, Contains(HasSubstr("test-token")));
 
   // Clear previous captured lines to ensure the following checks for new
   // messages.
```
devbww, line 84: This comment now seems out of place.

google/cloud/testing_util/capture_log_lines_backend.cc
```diff
@@ -19,12 +19,20 @@ namespace cloud {
 inline namespace GOOGLE_CLOUD_CPP_NS {
 namespace testing_util {
 
+std::vector<std::string> CaptureLogLinesBackend::ClearLogLines() {
+  std::lock_guard<std::mutex> lk(mu_);
+  auto result = std::move(log_lines_);
+  log_lines_.clear();
+  return result;
```
devbww, line 26: Consider ...
```
std::vector<std::string> result;
{
  result.swap(log_lines_);
}
return result;

google/cloud/spanner/internal/logging_result_set_reader_test.cc
coryan: Done.

google/cloud/testing_util/capture_log_lines_backend.cc
coryan: Done.

PR #4995, opened 2020-08-31
feature(generator): improve Printer missing variable diagnostic

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/4995)

### DIFF
--- a/generator/internal/printer.h
+++ b/generator/internal/printer.h
@@ -14,6 +14,7 @@
 #ifndef GOOGLE_CLOUD_CPP_GENERATOR_INTERNAL_PRINTER_H
 #define GOOGLE_CLOUD_CPP_GENERATOR_INTERNAL_PRINTER_H
 
+#include "google/cloud/internal/port_platform.h"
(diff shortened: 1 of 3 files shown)


generator/internal/printer.h
```diff
@@ -63,6 +65,38 @@ class Printer {
...
 
+  /**
+   * Like Print(), except it accepts diagnostic information from the caller
+   * and rethrows exceptions thrown by protoc in order to provide a more
+   * meaningful diagnostic.
+   */
+  void Print(int line, char const* file,
+             const std::map<std::string, std::string>& variables,
+             std::string const& text) {
+    try {
```
coryan, line 76: Does this even compile with exceptions disabled?  For that matter, do we care?  If protobuf throws to signal errors it seems that this code *requires* exceptions, so we should simply not compile the whole directory on those platforms.

Look at what we do for the `samples` directory in `spanner` and `pubsub` which also require exceptions to compile.

PR #5016, opened 2020-09-03
doc(spanner): add "spanner_query_with_*_parameter" samples

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5016)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -1141,15 +1141,15 @@ void InsertDatatypesData(google::cloud::spanner::Client client) {
           "Venues", {"VenueId", "VenueName", "VenueInfo", "Capacity",
                      "AvailableDates", "LastContactDate", "OutdoorVenue",
                      "PopularityScore", "LastUpdateTime"})
-          .EmplaceRow(4, "Venue 4", "Hello World 1", 1800, available_dates1,
-                      absl::CivilDay(2018, 9, 2), false, 0.85543,
-                      spanner::CommitTimestamp())
-          .EmplaceRow(19, "Venue 19", "Hello World 2", 6300, available_dates2,
-                      absl::CivilDay(2019, 1, 15), true, 0.98716,
-                      spanner::CommitTimestamp())
-          .EmplaceRow(42, "Venue 42", "Hello World 3", 3000, available_dates3,
-                      absl::CivilDay(2018, 10, 1), false, 0.72598,
-                      spanner::CommitTimestamp())
+          .EmplaceRow(4, "Venue 4", spanner::Bytes("Hello World 1"), 1800,
+                      available_dates1, absl::CivilDay(2018, 9, 2), false,
+                      0.85543, spanner::CommitTimestamp())
+          .EmplaceRow(19, "Venue 19", spanner::Bytes("Hello World 2"), 6300,
+                      available_dates2, absl::CivilDay(2019, 1, 15), true,
+                      0.98716, spanner::CommitTimestamp())
+          .EmplaceRow(42, "Venue 42", spanner::Bytes("Hello World 3"), 3000,
+                      available_dates3, absl::CivilDay(2018, 10, 1), false,
+                      0.72598, spanner::CommitTimestamp())
           .Build();
 
   auto commit_result = client.Commit(spanner::Mutations{insert_venues});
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -1160,6 +1125,208 @@ void DeleteData(google::cloud::spanner::Client client) {
...
+  namespace spanner = ::google::cloud::spanner;
+  spanner::Bytes example_bytes("Hello World 1");
+  spanner::SqlStatement select(
+      "SELECT VenueId, VenueName FROM Venues"
+      " WHERE VenueInfo = @venue_info",
+      {{"venue_info", spanner::Value(example_bytes)}});
+  using RowType = std::tuple<std::int64_t, absl::optional<std::string>>;
+  auto rows = client.ExecuteQuery(std::move(select));
+  for (auto const& row : spanner::StreamOf<RowType>(rows)) {
+    if (!row) throw std::runtime_error(row.status().message());
```
coryan, line 1218: The code coverage build says we do not reach this line, which suggests maybe the example is not working as expected?  I have no idea what, the only thing that jumps at me is maybe we need to insert the bytes column using the `spanner::Bytes` wrapper?

google/cloud/spanner/samples/samples.cc
```diff
@@ -1160,6 +1125,208 @@ void DeleteData(google::cloud::spanner::Client client) {
...
+      spanner::MakeTimestamp(std::chrono::system_clock::now()).value();
+  spanner::SqlStatement select(
+      "SELECT VenueId, VenueName, LastUpdateTime FROM Venues"
+      " WHERE LastUpdateTime < @last_update_time",
+      {{"last_update_time", spanner::Value(example_timestamp)}});
+  using RowType = std::tuple<std::int64_t, absl::optional<std::string>,
+                             absl::optional<spanner::Timestamp>>;
+  auto rows = client.ExecuteQuery(std::move(select));
+  for (auto const& row : spanner::StreamOf<RowType>(rows)) {
+    if (!row) throw std::runtime_error(row.status().message());
```
coryan, line 1321: This code is also not reached.  Which is odd given that we are using "now" the data was inserted in the past.  Maybe we should consider `now() + seconds(20)` or something?

coryan (commented):
I think two of the examples do not work as expected, or at least not consistently.  Any thoughts as to why?

google/cloud/spanner/samples/samples.cc
devbww: Right you are.  Thanks for catching that.

[I might say that it is surprising that the `Commit()` succeeds with ill-formatted `BYTES` values.]

google/cloud/spanner/samples/samples.cc
devbww: I see it finding matches (see below), but I agree that it could well be a race.

The Go sample changes the inequality to `<=`.  I'll do the same.

```
Running spanner_query_with_timestamp_parameter sample
2020-09-03T18:09:18.906254766Z [DEBUG] ExecuteStreamingSql() << session: "projects/cloud-cpp-testing-resources/instances/test-instance-11/databases/db-2020-09-03-x3r0pc-u4u8y_6pu/sessions/AN4G3x8CElBR9qMSfrkwTLINfvv8R7QnfmvFDxrUZiaA0-aAyai1xZ_UIZ6LuQ"
transaction {
  single_use {
    read_only {
      strong: true
      return_read_timestamp: true
    }
  }
}
sql: "SELECT VenueId, VenueName, LastUpdateTime FROM Venues WHERE LastUpdateTime < @last_update_time"
params {
  fields {
    key: "last_update_time"
    value {
      string_value: "2020-09-03T18:09:18.905830539Z"
    }
  }
}
param_types {
  key: "last_update_time"
  value {
    code: TIMESTAMP
  }
}
seqno: 1
 (./google/cloud/internal/log_wrapper.h:102)
2020-09-03T18:09:18.906686775Z [DEBUG] ExecuteStreamingSql() >> not null stream (./google/cloud/internal/log_wrapper.h:104)
VenueId: 4      VenueName: Venue 4      LastUpdateTime: 2020-09-03T18:09:18.299631Z
VenueId: 19     VenueName: Venue 19     LastUpdateTime: 2020-09-03T18:09:18.299631Z
VenueId: 42     VenueName: Venue 42     LastUpdateTime: 2020-09-03T18:09:18.299631Z
Query completed for [spanner_query_with_timestamp_parameter]
```

coryan (approved):
The timestamp "race" is probably going to create noise in the code coverage, but it is otherwise harmless.  I thought of alternatives, and I do not like them (adding some slack will make the example more confusing, using `now() - slack` at insert is also confusing).

LGTM.

PR #5023, opened 2020-09-08
feat(pubsub): subscriber retry settings example

### LINKED ISSUE #4622: Create examples for any retry policies
Any retry policy will need examples on how to set different values, and these examples should be linked from the main Doxygen page (see the spanner docs for how we do that in other libraries).

coryan:
Fixes #4622

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5023)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -991,6 +991,62 @@ void SubscriberFlowControlSettings(
   (std::move(subscriber), std::move(subscription));
 }
 
+void SubscriberRetrySettings(std::vector<std::string> const& argv) {
+  namespace examples = ::google::cloud::testing_util;
+  if (argv.size() != 2) {
+    throw examples::Usage{
+        "subscriber-retry-settings <project-id> <subscription-id>"};
+  }
+
+  //! [subscriber-retry-settings]
+  namespace pubsub = google::cloud::pubsub;
+  using google::cloud::future;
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/samples/samples.cc
```diff
@@ -991,6 +991,62 @@ void SubscriberFlowControlSettings(
...
+
+    std::mutex mu;
+    std::condition_variable cv;
+    int count = 0;
+    auto constexpr kExpectedMessageCount = 1;
+    auto handler = [&](pubsub::Message const& m, pubsub::AckHandler h) {
+      std::move(h).ack();
+      {
+        std::lock_guard<std::mutex> lk(mu);
+        std::cout << "Received message " << m << "\n";
```
devbww, line 1030: Opt: Perhaps move the logging outside the lock?

devbww (approved):

google/cloud/pubsub/samples/samples.cc
coryan: That was intentional:

- The output of the example is unreadable otherwise (in a larger application one would need to do something different, of course, but output from inside the lock works here).
- The performance impact is not an issue for an example methinks.

devbww: I wasn't thinking about the performance, but rather it being an example.  But I guess the output really isn't part of the example.  In any case, ack on the serialization.

PR #5027, opened 2020-09-08
feat(pubsub): support fire & forget subscriptions

### LINKED ISSUE #4970: Proposal: refactor pubsub_internal::SubscriptionSession
## Objective

The `pubsub_internal::SubscriptionSession` class has too many responsibilities, and (probably as a result), it is hard to change and test.  This issue proposes a solution, it is partly a notepad for me to write down what are the expected improvements.

## Background

`pubsub_internal::SubscriptionSession` is an internal-only (never exposed to the application) object to handle a single subscription. That is, it is the thing created by `pubsub::Subscriber::Subscribe()`.  It performs the following functions:

- Deliver a `pubsub::Message` and `pubsub::AckHandler` pair to the application for each message received from the service.
- Based on the subscription configuration parameters, control the concurrency for message delivery.
- Track how many callbacks are in progress (not fully handled).
- A callback is considered "in progress" until its `AckHandler::ack()` or `AckHandler::nack()` function is called. Applications can decouple the return of the callback from (presumably asynchronous) message handling.
- Request more data once all the messages received are fully handled.
- Data arrives in batches, the class breaks the batches for delivery to the application.
- Until a message is fully handled, the class refreshes the leases with the service, i.e., extends the message deadline until some application-configurable value.
- (Soon, not yet) Request more data before all the messages received are fully handled, with limits based on message count and total message size.

Writing test for this class is very difficult, as all these different layers of functionality interact with each other, often using multiple threads.  It would be easier if we could test each responsibility separately and then put them together in some sort of pipeline.

## Proposal

Break the class in roughly 4 classes:

* `pubsub_internal::SubscriptionFlowControl` will:
   * Track the number and size of the pending messages
   * Request new messages as needed or rather eagerly, as the flow control parameters permit it to.
   * Pass the received messages to the `SubscriptionLeaseManagement` class.
   * On `Start()` request at least one batch, and save a callback to let `SubscriptionLeaseManagement` know about incoming calls

* `pubsub_internal::SubscriptionLeaseManagement` will:
   * Keep the leases for any incoming messages up to date, this involves:
     * Setup timers to refresh the leases,
     * Remove messages from the list of those needing updates as they are acked/nacked
     * Pass ack/nack messages to the `SubscriptionFlowControl` class.
  * Pass any received message batches to the `SubscriptionMessageQueue` class (via a callback).
  * On `Start()` save the callback for `SubscriptionMessageQueue::OnPull()` and call `Start()` on `FlowControl`.

* `pubsub_internal::SubscriptionMessageQueue` will keep a buffer of messages generated by the `AsyncPull()` requests
  * It will remember the number of `Read()` calls, and respond to them by invoking a callback (typically going to `SubscriptionConcurrencyControl`)
   * The callback just responds with one `::google::pubsub::v1::ReceivedMessage` proto at a time, effectively breaking the batches.

* `pubsub_internal::SubscriptionConcurrencyControl` will invoke the application-level callbacks, subject to the concurrency control settings.
  * It needs to know how many callbacks are pending as it schedules more callbacks when earlier ones complete.
  * While it makes `Read()` requests to ask for more data, it does not expect an immediate response.

* A single instance of `pubsub_internal::ShutdownManager` will be shared by these classes to coordinate their shutdown.

image

## Memory Management

In the diagram ownership is from right to left. That is, the objects on the right own the objects on the left.

Callbacks always use `std::weak_ptr<>`.

The `ShutdownManager` coordinates shutdown, and receives a number of callbacks.

### LINKED ISSUE #4983: Implement a fire+forget subscription test
We need a unit tests for fire+forget subscriptions, where the application does not cancel the subscription but just leaves it running until the completion queue shuts dow.

coryan:
Some applications may want to create a subscription, setup a handler for
messages and errors, and then just leave the subscription "running" in
the background, associated with its completion queue. Until this PR the
subscription was terminated when the associated `future<Status>`
finished.

Fixes #4970 and fixes #4983

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5027)

### DIFF
--- a/google/cloud/pubsub/integration_tests/subscription_lifecycle.cc
+++ b/google/cloud/pubsub/integration_tests/subscription_lifecycle.cc
@@ -1,0 +1,137 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/pubsub/publisher.h"
+#include "google/cloud/pubsub/subscriber.h"
+#include "google/cloud/pubsub/subscription.h"
+#include "google/cloud/pubsub/subscription_admin_client.h"
+#include "google/cloud/pubsub/testing/random_names.h"
+#include "google/cloud/pubsub/topic_admin_client.h"
+#include "google/cloud/pubsub/version.h"
+#include "google/cloud/internal/getenv.h"
+#include "google/cloud/internal/random.h"
+#include "google/cloud/testing_util/assert_ok.h"
+#include "google/cloud/testing_util/status_matchers.h"
+#include <gmock/gmock.h>
+#include <map>
+
+namespace google {
+namespace cloud {
+namespace pubsub {
+inline namespace GOOGLE_CLOUD_CPP_PUBSUB_NS {
+namespace {
+
+using ::google::cloud::testing_util::StatusIs;
+using ::testing::AnyOf;
+using ::testing::ElementsAre;
+using ::testing::ElementsAreArray;
+
+TEST(MessageIntegrationTest, FireAndForget) {
+  auto project_id =
+      google::cloud::internal::GetEnv("GOOGLE_CLOUD_PROJECT").value_or("");
+  ASSERT_FALSE(project_id.empty());
+
+  auto generator = google::cloud::internal::MakeDefaultPRNG();
+  Topic topic(project_id, pubsub_testing::RandomTopicId(generator));
+  Subscription subscription(project_id,
+                            pubsub_testing::RandomSubscriptionId(generator));
+
+  auto topic_admin = TopicAdminClient(MakeTopicAdminConnection());
+  auto subscription_admin =
+      SubscriptionAdminClient(MakeSubscriptionAdminConnection());
+
+  auto topic_metadata = topic_admin.CreateTopic(TopicMutationBuilder(topic));
+  ASSERT_THAT(topic_metadata, AnyOf(StatusIs(StatusCode::kOk),
+                                    StatusIs(StatusCode::kAlreadyExists)));
+
+  struct Cleanup {
+    std::function<void()> action;
+    explicit Cleanup(std::function<void()> a) : action(std::move(a)) {}
+    ~Cleanup() { action(); }
+  };
+  Cleanup cleanup_topic(
+      [topic_admin, &topic]() mutable { topic_admin.DeleteTopic(topic); });
+
+  auto subscription_metadata = subscription_admin.CreateSubscription(
+      topic, subscription,
+      SubscriptionMutationBuilder{}.set_ack_deadline(std::chrono::minutes(2)));
+  ASSERT_THAT(
+      subscription_metadata,
+      AnyOf(StatusIs(StatusCode::kOk), StatusIs(StatusCode::kAlreadyExists)));
+
+  std::mutex mu;
+  std::condition_variable cv;
+  std::set<std::string> received;
+  Status subscription_result;
+  std::set<std::string> published;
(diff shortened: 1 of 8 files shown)


google/cloud/pubsub/integration_tests/subscription_lifecycle.cc
```diff
@@ -0,0 +1,137 @@
+// Copyright 2020 Google LLC
```
devbww, line 1: Would we not typically name this file with an `_test` suffix?

google/cloud/pubsub/integration_tests/subscription_lifecycle.cc
```diff
@@ -0,0 +1,137 @@
...
+#include "google/cloud/pubsub/subscription_admin_client.h"
+#include "google/cloud/pubsub/testing/random_names.h"
+#include "google/cloud/pubsub/topic_admin_client.h"
+#include "google/cloud/pubsub/version.h"
+#include "google/cloud/internal/getenv.h"
+#include "google/cloud/internal/random.h"
+#include "google/cloud/testing_util/assert_ok.h"
+#include "google/cloud/testing_util/status_matchers.h"
+#include <gmock/gmock.h>
+#include <map>
```
devbww, line 27: Unused?

google/cloud/pubsub/integration_tests/subscription_lifecycle.cc
```diff
@@ -0,0 +1,137 @@
...
+  ASSERT_THAT(
+      subscription_metadata,
+      AnyOf(StatusIs(StatusCode::kOk), StatusIs(StatusCode::kAlreadyExists)));
+
+  std::mutex mu;
+  std::condition_variable cv;
+  std::set<std::string> received;
+  Status subscription_result;
+  std::set<std::string> published;
+  std::vector<Status> publish_errors;
```
devbww, line 78:
```
#include <condition_variable>
#include <mutex>
#include <set>
#include <string>
#include <vector>
```

google/cloud/pubsub/internal/subscription_session_test.cc
```diff
@@ -672,6 +608,106 @@ TEST(SubscriptionSessionTest, ShutdownWaitsConditionVars) {
...
+      wait_ack_count(kMessageCount);
+    }
+
+    auto const initial_value = wait_ack_count(2 * kMessageCount);
+    auto const final_value = wait_ack_count(initial_value + 2 * kMessageCount);
+    EXPECT_NE(initial_value, final_value);
+    std::unique_lock<std::mutex> lk(mu);
+    EXPECT_STATUS_OK(status);
+  }
+  SUCCEED();
```
devbww, line 708: Q: Does this serve a purpose?

google/cloud/pubsub/internal/subscription_session.cc
```diff
@@ -38,38 +39,120 @@ class SubscriptionSessionImpl {
...
+  void ScheduleTimer() {
+    GCP_LOG(DEBUG) << __func__ << "()";
+    using TimerArg = future<StatusOr<std::chrono::system_clock::time_point>>;
+
+    std::unique_lock<std::mutex> lk(mu_);
+    auto self = shared_from_this();
+    lk.unlock();
+    auto t = cq_.MakeRelativeTimer(shutdown_polling_period_)
+                 .then([self](TimerArg f) { self->OnTimer(!f.get().ok()); });
+    if (shutdown_state_ == kShutdownByCompletionQueue) return;
```
devbww, line 109: Is `shutdown_state_` not protected by `mu_`?

coryan: Yikes, fixed, thanks.

google/cloud/pubsub/integration_tests/subscription_lifecycle.cc
coryan: Fixed.

google/cloud/pubsub/internal/subscription_session_test.cc
coryan: Not really, removed.

google/cloud/pubsub/integration_tests/subscription_lifecycle.cc
coryan: Removed.

google/cloud/pubsub/integration_tests/subscription_lifecycle.cc
coryan: Renamed

devbww (approved):

Issue #5032, opened 2020-09-09
No way to specify content length in gRPC GCS


dopiera:
In cURL API there is `X-Upload-Content-Length` header, which allows specifying the desired content length of an object. There is no counterpart in gRPC API.

There is a field in the proto, but the comments suggest that it should not be used for this purpose

coryan:
This seems like a dup to #4286 ?

devjgm:
Still blocked.

coryan:
Unblocked and now a feature request only.

PR #5033, opened 2020-09-09
fix: specify content length in gRPC GCS

dopiera:
This doesn't really work - the service ingores this field.

This is related to #5032  

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5033)


coryan (commented):
I am slightly confused.  What is the purpose of this PR?  The change does not work, and it is not clear that it will be implemented as assumed here.
Maybe we should just live with the bug until we have clarity of how to server will implement this feature?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5033#-:-MH3DrnxA_IqTBwV_JNo:bo8sbek)* status: 0 of 2 files reviewed, all discussions resolved

dopiera:
If we're not sure this is how it's going to be fixed, we definitely should not merge it. I probably remembered out previous discussions on it wrong.

Issue #4718, opened 2020-07-29
Make internal::CompletionQueueImpl fully mockable


coryan:
### Problem

Currently `internal::CompletionQueueImpl` is not mockable. Several of its member functions are not virtual:

https://github.com/googleapis/google-cloud-cpp/blob/e8354a6c1b05002a6ef7771cbffabfcf8d92a60b/google/cloud/internal/completion_queue_impl.h#L236

https://github.com/googleapis/google-cloud-cpp/blob/e8354a6c1b05002a6ef7771cbffabfcf8d92a60b/google/cloud/internal/completion_queue_impl.h#L239

and It has some template member functions:

https://github.com/googleapis/google-cloud-cpp/blob/e8354a6c1b05002a6ef7771cbffabfcf8d92a60b/google/cloud/internal/completion_queue_impl.h#L251-L255

We have a class called `MockCompletionQueue` but it is closer to a "fake" than a "mock": it basically allows applications to simulate timer expiration.

### Proposal

We need to change `CompletionQueueImpl` to something like this:

```cc
class CompletionQueueImpl {
 public:
   virtual ~CompletionQueueImpl() = default;
   virtual void Run() = 0;
   virtual void Shutdown() = 0;
   virtual void CancelAll() = 0;
   virtual grpc::CompletionQueue& cq() = 0;

  // Replaces CreateAlarm().
  virtual future<StatusOr<std::chrono::system_clock::time_point>>
      CreateDeadlineTimer(std::chrono::system_clock::time_point tp) = 0;

  // Used by CreateRelativeTimer() to setup timers under control of a mock.
  virtual std::chrono::system_clock::time_point now() = 0;

  // Replaces the template member function StartOperation()
  virtual void StartOperation(
      std::shared_ptr<AsyncGrpcOperation> op, absl::FunctionRef<void(void*)> starter) = 0;
};
```

### The existing `MockCompletionQueueImpl`

This is the more difficult bit, I think we should rename this class to `FakeCompletionQueueImplWithTimerQueue` and implement it (more or less) as follows:

```cc
class FakeCompletionQueueImplWithTimerQueue
  : public CompletionQueueImpl {
 public:
  ~FakeCompletionQueueImplWithTimerQueue() override = default;

  void Run() override { /* block until Shutdown is called */ }
  void Shutdown() override { /* unblock all the Run() things */ }
  void CancelAll() override { /* call op->Notify() on all the ops */ }
[...]

PR #5036, opened 2020-09-10
feat(common): make CompletionQueueImpl mockable [1]

coryan:
Remove all the template member functions from `CompletionQueueImpl`, so
it can be refactored to an abstract class in a future PR. Simplify the
timers for the fake version of `CompletionQueueImpl`.

Part of the work for #4718

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5036)


coryan:
The API baseline reported a removed function:

```
completion_queue.h
namespace google::cloud::v1
CompletionQueue::MakeDeadlineTimer ( std::chrono::_V2::system_clock::time_point
deadline )
```

This is a false positive, the function moved to the `.h` file where it became inline and thus *appears* as removed because it is no longer in the `.so` file.

google/cloud/internal/completion_queue_impl.cc
```diff
@@ -69,8 +153,39 @@ void CompletionQueueImpl::CancelAll() {
...
+  if (shutdown_) {
+    lk.unlock();
+    op->Notify(/*ok=*/false);
+    return;
+  }
+  auto ins =
+      pending_ops_.emplace(reinterpret_cast<std::intptr_t>(tag), std::move(op));
+  if (ins.second) {
+    start(tag);
+    lk.unlock();
```
devbww, line 184: Remove?

google/cloud/internal/completion_queue_impl.cc
```diff
@@ -69,8 +153,39 @@ void CompletionQueueImpl::CancelAll() {
...
+    return;
+  }
+  google::cloud::internal::ThrowRuntimeError(
+      "assertion failure: insertion should succeed");
```
devbww, line 188: Perhaps something involving "duplicate tag" would be helpful?

coryan: Done.

devbww (approved):

PR #5037, opened 2020-09-10
fix(storage): work with unknown SSL version in curl

coryan:
Sometimes libcurl will report the SSL version as `NULL`, typically
because there are multiple SSL engines, or because there is no SSL
version (say when compiled only with Schannel). This change avoids
crashes in that case by returning an empty string for the SSL version
name.

This name is used to determine if SSL requires locking. I think an empty
string works correctly in that case. Only a limited number of *known*
SSL versions do require the locks.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5037)

### DIFF
--- a/google/cloud/storage/internal/curl_wrappers_disable_sigpipe_handler_test.cc
+++ b/google/cloud/storage/internal/curl_wrappers_disable_sigpipe_handler_test.cc
@@ -33,12 +33,12 @@ TEST(CurlWrappers, SigpipeHandlerDisabledTest) {
 #if __has_feature(memory_sanitizer)
   // The memory sanitizer seems to intercept SIGPIPE, simply disable the test
   // in this case.
-  return;
+  GTEST_SKIP();
 #endif  // __has_feature(memory_sanitizer)
 #endif  // __has_fature
 
 #if !defined(SIGPIPE)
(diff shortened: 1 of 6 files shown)


google/cloud/storage/internal/curl_wrappers_disable_sigpipe_handler_test.cc
```diff
@@ -33,12 +33,12 @@ TEST(CurlWrappers, SigpipeHandlerDisabledTest) {
 #if __has_feature(memory_sanitizer)
   // The memory sanitizer seems to intercept SIGPIPE, simply disable the test
   // in this case.
-  return;
+  GTEST_SKIP();
```
devbww, line 36: Did you intend for all the `GTEST_SKIP()` stuff to be part of this PR?

coryan: Yes, I was looking at where these functions were used and got carried away. I can split it if you want.

devbww: Just checking.

devbww (approved):

PR #5038, opened 2020-09-10
fix(common): correct C++ version under MSVC

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5038)

### DIFF
--- a/google/cloud/internal/port_platform.h
+++ b/google/cloud/internal/port_platform.h
@@ -30,9 +30,16 @@
 // readable with indentation.
 // clang-format off
 
-// Microsoft Visual Studio does not define __cplusplus correctly:
+// Microsoft Visual Studio does not define __cplusplus correctly by default:
+//
 //   https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus
-// Define a macro to hide this detail.
+//
+// Here we define a macro to hide this detail.
+//
+// We considered using the `/Zc:__cplusplus` flag with MSVC, but because we
+// use `__cplusplus` in headers we would be forcing downstream code to set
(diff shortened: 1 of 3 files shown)


google/cloud/internal/port_platform.h
```diff
@@ -30,17 +30,20 @@
 // readable with indentation.
 // clang-format off
 
-// Abort compilation if the compiler does not support C++11.
-#ifndef _MSC_VER
 // Microsoft Visual Studio does not define __cplusplus correctly:
```
devbww, line 33: A tidbit from StackOverflow ...

> Thanks to an anonymous commenter for pointing out that there is a `/Zc:__cplusplus` compiler switch, which sets the value of `__cplusplus` to match the variant you select.

... which sounds like it does the right thing with `/std:c++XX` switches.

coryan: I thought about that.  That would change the definition for `__cplusplus` for our users too [1], which could break our users, at least the MSVC folks reports it does, that is why they did not change the default:

> We tried to define `__cplusplus` correctly by default but discovered that a lot of code expects MSVC to always set the macro to “199711L”.

https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus


[1]: we could set the flag when compiling our code, and not when compiling their code, but that is tricky with Bazel, and does not work if we use `__cplusplus` in headers. There are two users in headers, one that I can see how to remove, the other I could not figure it out.  Overall, this seemed safer and relatively clean.

This probably deserves an explanation in the code anyway.

devbww: Thanks.

devbww (approved):

PR #5051, opened 2020-09-12
cleanup: automate Abseil pkg-config dependencies

coryan:
Abseil does not install its own pkg-config `.pc` files, so in our own
`.pc` files we need to list the `-l` options explicitly. With this
change we navigate the dependency graph for the CMake targets to
discover all these flags, instead of hard-coding them in our list.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5051)

### DIFF
--- a/google/cloud/CMakeLists.txt
+++ b/google/cloud/CMakeLists.txt
@@ -169,7 +169,7 @@ add_library(
     version.h)
 target_link_libraries(
     google_cloud_cpp_common PUBLIC absl::memory absl::optional absl::time
-                                   absl::variant Threads::Threads)
+                                   Threads::Threads)
 google_cloud_cpp_add_common_options(google_cloud_cpp_common)
 target_include_directories(
     google_cloud_cpp_common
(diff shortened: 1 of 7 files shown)


devjgm (approved):
Wow, that's great! At what point are you just going to `chsh -s /usr/bin/cmake`?

google/cloud/CMakeLists.txt
```diff
@@ -303,9 +303,9 @@ set(GOOGLE_CLOUD_CPP_PC_DESCRIPTION
 # libraries, so we need to explicitly include them here. Here we include the
 # runtime deps of the absl::optional target (http://github.com/abseil/abseil-cpp
 # /blob/4a851046a0102cd986a5714a1af8deef28a544c4/absl/types/CMakeLists.txt#L164)
-set(GOOGLE_CLOUD_CPP_PC_LIBS
-    "-lgoogle_cloud_cpp_common -labsl_time -labsl_time_zone -labsl_bad_optional_access"
-)
+google_cloud_cpp_absl_pkg_config(absl_pkg_config google_cloud_cpp_common)
+string(CONCAT GOOGLE_CLOUD_CPP_PC_LIBS "-lgoogle_cloud_cpp_common"
```
devbww, line 307: It is a little strange that we depend on the result of `google_cloud_cpp_absl_pkg_config` starting with a space.

coryan: Yeah. Unfortunately `string(JOIN)` is not available until cmake-3.12, and we need to support cmake-3.5 (for now, we should review that).  I guess we could implement our own `google_cloud_cpp_join()` function, but I am lazy.

Issue #3847, opened 2020-03-19
Document what we consider an "API breakage"


coryan:
I think we should create an ADR documenting what is an API breakage, or at least the things we do not consider part of the API.  I am thinking of:

- We make no attempts to preserve the A**B**I.  We change object sizes, add fields, change private members functions and fields without notice. We also may add default parameters to a function without notice.
- For the purposes of this document "API" means the "library API", the guarantees about the REST and/or gRPC APIs offered by GCP are discussed elsewhere.
- We try to preserve the API, but we will break the API if we think it is important, we will document -- to the best of our abilities -- any API breakage.
- Header files are part of the API, but you should not rely on indirect `#includes`.
- We consider type aliases and inline functions part of the API
- We do **not** consider types, functions or objects in the `internal` (or `foo_internal`) namespace to be part of the API, we will change those without notice.
- We do **not** consider types, functions or objects used in our tests part of the API, this includes anything in the `testing` or `foo_testing` namespace.
- We do **not** consider header files in the `internal/` or `testing/` or `testing_utils/` directories to be part of the API.
- This list is not exhaustive, as we discover other things that should not be considered part of the API we will update this list, but still will document any changes.

devjgm:
We should also document what we guarantee wrt to our build deps: https://github.com/googleapis/google-cloud-cpp/pull/5054#issuecomment-692141373

PR #5054, opened 2020-09-14
fix: remove unneeded dep

devjgm:
This was an unnecessary dep that @coryan discovered in
https://github.com/googleapis/google-cloud-cpp/pull/5051 and removed in
the CMakeLists.txt file. This PR cleans up the unneeded dep in the bazel
build.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5054)

### DIFF
--- a/google/cloud/BUILD
+++ b/google/cloud/BUILD
@@ -61,7 +61,6 @@ load(
         ":google_cloud_cpp_common",
         "//google/cloud/testing_util:google_cloud_cpp_testing",
         "@com_google_absl//absl/memory",
-        "@com_google_absl//absl/types:variant",
         "@com_google_googletest//:gtest_main",
     ],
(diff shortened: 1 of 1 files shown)


coryan (approved):

coryan:
I guess technically this is a breaking change.  Somebody might be using `google_cloud_cpp_common` and depending on getting `absl::variant` from it.  I think when we write the full set of rules of what constitutes a breaking change we should say something like:

> We promise that our dependencies are complete, i.e., you do not need to provide additional deps to link our (public) targets.  We do not promise that indirect dependencies are stable, i.e., we can add and remove dependencies to our targets as needed.

devjgm:
Good point. Added that comment to https://github.com/googleapis/google-cloud-cpp/issues/3847

PR #5066, opened 2020-09-16
ci: remove unneeded explicit build step

devjgm:
I don't think there's any need to do an explicit `bazel build`. Any code and tests will be compiled when they need to be run. Additionally, if a `bazel test` invocation uses slightly different flags than the `bazel build` invocation, the cache will not be hit and we'll unnecessarily compile stuff in the `bazel build` step.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5066)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -85,12 +85,6 @@ echo "bazel test" "${bazel_args[@]}"
 "${BAZEL_BIN}" test \
   "${bazel_args[@]}" "--test_tag_filters=-integration-test" ...
 
-echo "================================================================"
-io::log "Compiling all the code, including integration tests"
-# Then build everything else (integration tests, examples, etc). So we can run
-# them next.
-"${BAZEL_BIN}" build "${bazel_args[@]}" ...
-
 should_run_integration_tests() {
   if [[ "${SOURCE_DIR:-}" == "super" ]]; then
     # super builds cannot run the integration tests


mr-salty (commented):
it's probably worth discussing this.

I like the `...` since its (mostly) infallible for making sure nothing gets missed (maybe `...:*` would actually be better in that respect?) That probably doesn't have to happen in every build but I'd like at least one build to do it. Alternatively, maybe we could leverage [`--check_up_to_date`](https://docs.bazel.build/versions/master/command-line-reference.html#flag--check_up_to_date) to ensure tests are building everything?

I assume the flag discrepancy you mentioned is actually happening?  Can we fix that instead?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5066#-:-MHJlw6AEUDSRFaJNfLY:bqknk2t)* status: 0 of 1 files reviewed, all discussions resolved

mr-salty (approved):
That being said, if this will save us significant build time I'm mostly ok with checking this in now then working on the rest.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5066#-:-MHJoYVgBo6dgEYM1uPN:bol0zrk)* status: 0 of 1 files reviewed, all discussions resolved

devjgm:
> it's probably worth discussing this.
> 
> I like the `...` since its (mostly) infallible for making sure nothing gets missed (maybe `...:*` would actually be better in that respect?) That probably doesn't have to happen in every build but I'd like at least one build to do it. Alternatively, maybe we could leverage [`--check_up_to_date`](https://docs.bazel.build/versions/master/command-line-reference.html#flag--check_up_to_date) to ensure tests are building everything?

We have a "`...`" build a few lines up:


> 
> I assume the flag discrepancy you mentioned is actually happening? Can we fix that instead?
> 
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5066#-:-MHJlw6AEUDSRFaJNfLY:bqknk2t)_ status: 0 of 1 files reviewed, all discussions resolved

mr-salty (approved):
:lgtm:
 
I never knew `bazel test` had that behavior (I thought it only build and ran _test targets), but indeed: https://docs.bazel.build/versions/master/user-manual.html#test

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5066#-:-MHJwq3K5U4cvem4SJw-:b8adahk)* status: 0 of 1 files reviewed, all discussions resolved

ci/kokoro/docker/build-in-docker-bazel.sh
```diff
@@ -85,12 +85,6 @@ echo "bazel test" "${bazel_args[@]}"
 "${BAZEL_BIN}" test \
   "${bazel_args[@]}" "--test_tag_filters=-integration-test" ...
 
-echo "================================================================"
-io::log "Compiling all the code, including integration tests"
-# Then build everything else (integration tests, examples, etc). So we can run
-# them next.
-"${BAZEL_BIN}" build "${bazel_args[@]}" ...
```
coryan, line 92: AFAIK all the other `"${BAZEL_BIN}"` commands in this script filter out *something*.  This is (or was), the only command that guaranteed everything built.  Did I miss something?  If I did, just merge and please explain it to me later, no worries.

devjgm: Just a few lines above this we have

https://github.com/googleapis/google-cloud-cpp/blob/e102e5ee35304723716fad99a3b2b1e0c738e20c/ci/kokoro/docker/build-in-docker-bazel.sh#L85-L86

which will do the following:

* It will [build everything](https://docs.bazel.build/versions/master/user-manual.html#test) matching `...` (i.e., everything)
* It will run all the tests, except for those with the `integration-test` tag

So in the best case, the `bazel build` command that this PR removes would do nothing (because the previous command already built everything). But in the worst case, this bazel build will recompile everything again if the flags changed between the "test" and the "build". The flags don't actually change currently, so today, this bazel build is simply doing nothing. However, _iff_ that `bazel test` command were changed to `bazel coverage` (as I'm experimenting with in https://github.com/googleapis/google-cloud-cpp/pull/5065), then the compile flags _will_ be different, and this `bazel build` command will needlessly recompile all targets.

Regardless of whether we move forward w/ the other PR to use `bazel coverage`, it seems like this `bazel build` step is unnecessary.

coryan: Doh, I missed your earlier answer last night, sorry for the noise.

coryan (approved):

Issue #5067, opened 2020-09-16
Explore using bazel for code coverage builds


devjgm:
Our coverage builds usually take ~40 min. These tend to be the long poles that our PRs are blocked on before submitting. The coverage builds today use cmake (without Ninja), and run the tests one-at-a-time. We may be able to speed up the coverage builds if we switched them to use bazel because:

1. We could run the tests in parallel
2. Bazel's remote cache is more up-to-date than the ccache tarballs that we upload during the CI build and download during the PR builds

Bazel's coverage builds (i.e., `bazel coverage `) produce output in files named `coverage.dat`, which appear to be in lcov format. It looks like the codecov.io/bash script knows how to handle these files. These files would live in `cmake-out/home/*coverage/.cache/..../coverage.dat`. 

I'm not sure if this would work, whether codecov.io would understand the output correctly (probably??), or whether it would be any faster.... but it may be worth a try.

FTR: I have a small experiment PR for this at https://github.com/googleapis/google-cloud-cpp/pull/5065

coryan:
> We could run the tests in parallel

Note that CMake can also run tests in parallel. We explicitly disable that because running coverage tests in parallel can corrupt the data:

https://stackoverflow.com/questions/14643589/code-coverage-using-gcov-on-parallel-run

We tried that before in #3530.   It is certainly possible (I would think likely) that Bazel has solved this problem better than our earlier attempts.

As you test these changes: think about coverage build and caching, if a test does not run, would this use previously cached coverage results?  Did your testing verify the effect of reducing code coverage with caching?  Did your testing verify the effect of increasing code coverage with caching?

devjgm:
> 
> As you test these changes: think about coverage build and caching, if a test does not run, would this use previously cached coverage results? Did your testing verify the effect of reducing code coverage with caching? Did your testing verify the effect of increasing code coverage with caching?

If a test is not run due to a cache hit, it'll reuse the cached `coverage.dat` file. The output in the logs look like:

```
//google/cloud/testing_util:status_matchers_test                (cached) PASSED in 0.4s
  /h/.cache/bazel/_bazel_kbuilder/a7ecd8237744645c5d189c197108d6d2/execroot/com_github_googleapis_google_cloud_cpp/bazel-out/k8-fastbuild/testlogs/google/cloud/testing_util/status_matchers_test/coverage.dat
```

So, I believe a cache hit will look exactly the same as no cache hit from a code coverage perspective.

Issue #3530, opened 2020-03-27
Parallelize code coverage builds


devjgm:

devjgm:
I think this issue needs a few more details explaining what needs to change. 

I found:

coryan:
gcov corrupts the `.gcda` files if multiple executables are running in parallel, that is why we current **disable** parallel runs:

https://stackoverflow.com/questions/14643589/code-coverage-using-gcov-on-parallel-run

As you see there, there is a workaround:

https://gcc.gnu.org/onlinedocs/gcc-5.2.0/gcc/Cross-profiling.html

But then the problem becomes using a different value of `GCOV_PREFIX` for each binary, which *might* be possible using the right annotations for each test:

https://cmake.org/cmake/help/v3.6/prop_test/ENVIRONMENT.html

which would translate to adding a directive like this to each spot where we define a test:

```CMake
set_property(TEST ${test} ENVIRONMENT "GCOV_PREFIX=${something_unique}")
```

But this is all theory, somebody has to actually test it, and make sure it integrates with codecov.io and uploads the right things.

coryan:
Just FYI, when serialized:

- The unit tests take 31s
- The bigtable integration tests take 28s
- The storage integration tests take 173s (!!) 
- The tests aganst production take 15s
- The remaining examples take about 2s

These time could be reduced by a factor of 4 if we parallelize them. With -spanner the numbers are even more significant, in this build the integration tests took **1800s** (!!!  -- but it varies a lot, sometimes it takes "as little as" 10 minutes to run the tests):

https://source.cloud.google.com/results/invocations/00c0c50f-08ca-4f97-a96c-e9dd8abf46b1/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp-spanner%2Fmaster%2Fdocker%2Fcoverage-presubmit/log

In short, this build will become the bottleneck as we merge more projects into the monorepo.

coryan:
I think I have bad news for this idea.  I prototyped using a different directory for each test, so they can be parallelized.  It works, but one then needs to merge the `.gcda` tests from each directory and that takes about 2 seconds per directory, we have about 200, so it takes 400 seconds to merge them back.

That is longer than we will probably save from parallelizing these.

For future reference the trick to splitting the output was to add this after each `add_test()`:

```
    string(SHA256 hash "${CMAKE_CURRENT_SOURCE_DIR}/${fname}" )
[...]

Issue #4297, opened 2020-05-28
Enable Spanner+Emulator+Bazel tests


devjgm:
https://github.com/googleapis/google-cloud-cpp/blob/e01df8e3eb7e7d3c61c71dd7c59c1b713f8ff3e1/ci/kokoro/docker/build-in-docker-bazel.sh#L149

currently runs the Bigtable tests with an emulator using bazel (in the `ci/kokoro/docker/build.sh integration` build). However, it doesn't do the same for Spanner (or Storage). It runs the Spanner integration tests, but against prod, not the emulator.

We should probably enable the emulator.

coryan:
May want to fix #4080 before doing this.

devjgm:
Note: This issue may become even more important if we switch our coverage builds to use bazel (https://github.com/googleapis/google-cloud-cpp/pull/5070).

devjgm:
FTR, a the Spanner integration tests hitting production took 4 min today. Not that this is high or low, but it's a data point. That's how much time we could potentially save.

coryan:
I think this was fixed in #6812

PR #5070, opened 2020-09-16
ci: use bazel for the coverage build

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/5067

This changes our "coverage" build to use bazel rather than cmake. The idea is that since bazel can use a remote cache that's updated from the pre- and post-merge builds, the cache hits may be higher. Bazel also knows how to run coverage tests in parallel, which solves https://github.com/googleapis/google-cloud-cpp/issues/3530.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5070)


coryan:
This is really cool.  Note that the CMake builds use the emulator for the spanner tests.  That is faster than production, so this could go even faster if we changed the Bazel builds (or at least this build) to use the emulator too.

I think that explains some of the coverage differences for `google/cloud/spanner/testing/...` too.

coryan:
Also, it seems this excludes all the coverage results from tests (that seems fine), and examples.

Examples are important (to me), I look at them to see how much of our sample code actually is tested.

devjgm:
> This is really cool. Note that the CMake builds use the emulator for the spanner tests. That is faster than production, so this could go even faster if we changed the Bazel builds (or at least this build) to use the emulator too.

Good point. I added a note to https://github.com/googleapis/google-cloud-cpp/issues/4297

> Also, it seems this excludes all the coverage results from tests (that seems fine), and examples.
> Examples are important (to me), I look at them to see how much of our sample code actually is tested.

I added the `--instrument_test_targets` flag, which I think should address this. A new build is running now.

coryan (commented):
### Things we MUST fix in this PR

For Bigtable, I think we did not activate all the tests we could when we moved the Bazel builds to the emulator:

https://codecov.io/gh/googleapis/google-cloud-cpp/pull/5070/changes#L43

That seems like easy to fix and would recover most of the lost coverage.

We lost all coverage for the `gcs2cbt.cc` example.

### This we SHOULD fix in a future PR

Running against production for spanner changed things:

https://codecov.io/gh/googleapis/google-cloud-cpp/pull/5070/tree/google/cloud/spanner

We gained some coverage:

https://codecov.io/gh/googleapis/google-cloud-cpp/pull/5070/src/google/cloud/spanner/samples/samples.cc#L3200

And lost some:

https://codecov.io/gh/googleapis/google-cloud-cpp/pull/5070/src/google/cloud/spanner/samples/samples.cc#L3129

We should really split the tests / samples and run against both production and the emulator to cover both.

On the storage side, we lost all the code for `getrusage()` (it is `#ifdef`-out in Bazel):

https://codecov.io/gh/googleapis/google-cloud-cpp/pull/5070/changes#L164

We should fix that in a future PR that enables this code for Bazel.

### Things we can ignore

I think we also lost the weird tests to trigger failures on `read()` calls:

https://codecov.io/gh/googleapis/google-cloud-cpp/pull/5070/changes#L197

We are testing with a different SSL library:

https://codecov.io/gh/googleapis/google-cloud-cpp/pull/5070/changes#L87

Do you want to remove the code from CMake to support code coverage now or in a future PR?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5070#-:-MHN_NLx5s_PJGgEvr2t:b-qz2r9u)* status: 3 of 8 files reviewed, all discussions resolved (waiting on @coryan)

___
*[ci/kokoro/docker/Dockerfile.fedora-install, line 136 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5070#-MHNZx8U8HRPBNt7nQeT:-MHNZx8U8HRPBNt7nQeU:blf2gn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/20a1bd8b08c4b2989f67646a73a48044d294e971/ci/kokoro/docker/Dockerfile.fedora-install#L136)):*
> ```
> # The Cloud Pub/Sub emulator needs Java, and so does `bazel coverage` :shrug:
> # Bazel needs the '-devel' version with javac.
> RUN dnf makecache && dnf install -y java-latest-openjdk-devel
> ```

FYI, this means the builds will be even faster, because this is rebuilding at least part of the development docker image.

coryan:
I am happy with the changes. As your probably noticed, `codecov.io` does not seem to be. The last 3 builds failed to upload.  Any thoughts?

devjgm:
> I am happy with the changes. As your probably noticed, `codecov.io` does not seem to be. The last 3 builds failed to upload. Any thoughts?

Thanks. I did see the issues on `codecov.io`. So far, I've found no way to debug codecov.io issues. It says there was an error, but doesn't say what the error was nor provide any way for me to debug it. I'll keep looking and googling.

devjgm:
I've no idea why codecov is failing to process these files sometimes, and I see no way to debug it. I asked on the codecov support forum, so we'll see: https://community.codecov.io/t/help-debugging-there-was-an-error-processing-coverage-reports/1984

coryan:
The only thing that comes to mind is comparing the size of the pre-Bazel and post-Bazel uploads.

devjgm:
Due to a mistaken `rebase -i`, I cannot reopen this PR. The new PR is https://github.com/googleapis/google-cloud-cpp/pull/5075

PR #5078, opened 2020-09-17
feature(generator): add generation of logging decorator class

scotthart:
I'm starting to see some common patterns across the stub and logging generation and test code that will likely also be present in the metadata_decorator, connection, and client generation code. In a later PR, I will undertake refactoring some of these common patterns to reduce code duplication.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5078)


generator/internal/logging_decorator_generator.h
```diff
@@ -0,0 +1,70 @@
+// Copyright 2020 Google LLC
```
devjgm, line 1: I forget: how/where are these generators tested? There's no `logging_decorator_generator_test.cc` file, so is there something else that's invoking and testing this?

devjgm (approved):

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5078)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @devjgm)

___
*[generator/internal/logging_decorator_generator.h, line 1 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5078#-MHT74fVydu0MtBdwMBV-r1-1:-MHX6hwBE8EZoC-wVmGY:b-iyrcqe) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/60f08e35e4dbb8765a4d34296a5cb4c99e513ab7/generator/internal/logging_decorator_generator.h#L1)):*

I forget: how/where are these generators tested? There's no `logging_decorator_generator_test.cc` file, so is there something else that's invoking and testing this?
</blockquote></details>

In generator_integration_test.cc, each file is generated from test.proto and compared against its respective golden file. Tests for the contents of the golden files are in the works.

Issue #5072, opened 2020-09-16
[storage] Confusing behavior when using a custom endpoint


megaserg:
**Does this issue affect the google-cloud-cpp project?**
I believe so.

**What component of `google-cloud-cpp` is this related to?**
GCS/storage/CURL client.

**Describe the bug**

- **TL;DR:** Tried to use a custom endpoint, some requests properly go to the endpoint, and some go to `storage.googleapis.com` instead. This is because CURL client 1) automatically switches to XML API unless certain headers are provided 2) ignores the custom endpoint for XML requests in favor of `CLOUD_STORAGE_TESTBENCH_ENDPOINT`.

- **Long version:** I am trying to add GCS support in [apache/arrow](https://github.com/apache/arrow/). I've built and installed the `google-cloud-cpp` project as described in `doc/packaging.md` (on Ubuntu 18.04), except that I made shared libs. 
- Then, I wrote some C++ code implementing a filesystem interface for Arrow, using `google::cloud::storage` client. 
- Then, I wrote a test using [fsouza/fake-gcs-server](https://github.com/fsouza/fake-gcs-server). The test setup creates a bucket, creates a file, and reads the file back. 
- To be able to use `fake-gcs-server`, I launch the server like this:

```
docker run -i --rm -p 4443:4443 fsouza/fake-gcs-server:latest -scheme http
```

and provide a custom endpoint like this:

```
namespace gcs = google::cloud::storage;

StatusOr<gcs::ClientOptions> client_config_;
std::unique_ptr<gcs::Client> client_;

client_config_ = gcs::ClientOptions::CreateDefaultClientOptions();
std::string endpoint = "http://127.0.0.1:4443";
client_config_->set_endpoint(endpoint);
auto creds = gcs::oauth2::CreateAnonymousCredentials();
client_config_->set_credentials(creds);
client_.reset(new gcs::Client(*client_config_));
```

- Then, using `gcs::Client`, I create a bucket `bucket`, insert an object `somefile`, and read an object with `client_->ReadObject("bucket", "somefile");`. The resulting `gcs::ObjectReadStream` has a badbit, and the status just says `Permanent error in Read():  [PERMISSION_DENIED]`.
- It turns out, the `google-cloud-cpp`'s CURL client
  - automatically switches to XML API unless certain headers are provided: https://github.com/googleapis/google-cloud-cpp/blob/v1.17.0/google/cloud/storage/internal/curl_client.cc#L566
[...]

PR #5085, opened 2020-09-18
fix(storage): use correct host header

coryan:
Some of our requests were using the wrong `Host:` header, luckily the
service is forgiving, and the testbench too.

Part of the work for #5072

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5085)

### DIFF
--- a/google/cloud/storage/internal/curl_wrappers.cc
+++ b/google/cloud/storage/internal/curl_wrappers.cc
@@ -261,7 +261,10 @@ std::string ExtractUrlHostpart(std::string url) {
   // be as it is used a handful of times when the CurlClient is created.
   for (std::string scheme_prefix : {"https://", "http://"}) {
     auto p = url.rfind(scheme_prefix, 0);
-    if (p == 0) url.erase(0, scheme_prefix.size());
+    if (p == 0) {
+      url.erase(0, scheme_prefix.size());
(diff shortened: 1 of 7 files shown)


google/cloud/storage/internal/curl_wrappers.cc
```diff
@@ -244,6 +244,41 @@ void CurlInitializeOnce(ClientOptions const& options) {
...
+  //
+  // If we get an invalid-looking URL we simply return an empty string, this is
+  // only used in the CurlClient to generate the `Host: ` header. For invalid
+  // URLs the CurlClient won't work in any case.
+  //
+  // Likewise, this implementation is not super efficient. It does not need to
+  // be as it is used a handful of times when the CurlClient is created.
+  for (std::string scheme_prefix : {"https://", "http://"}) {
+    auto p = url.rfind(scheme_prefix, 0);
+    if (p == 0) url.erase(0, scheme_prefix.size());
```
devbww, line 264: And `break`?

devbww (approved):

google/cloud/storage/internal/curl_wrappers.cc
coryan: Thanks, done.

PR #5090, opened 2020-09-18
cleanup: Fix typos in Doxygen comments

remyabel:
This changes nonretriable (which I don't think is a word) to non-retryable, but
happy to go either way on this.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5090)

### DIFF
--- a/google/cloud/bigtable/instance_admin.h
+++ b/google/cloud/bigtable/instance_admin.h
@@ -324,7 +324,7 @@ class InstanceAdmin {
    * @param instance_update_config config with modified instance.
    *
    * @return a future satisfied when either (a) the instance is updated or (b)
-   *     an unretriable error occurs or (c) polling or retry policy has been
+   *     an non-retryable error occurs or (c) polling or retry policy has been
    *     exhausted.
    *
    * @par Idempotency
@@ -591,7 +591,7 @@ class InstanceAdmin {
    * @param cluster_config cluster with updated values.
    *
    * @return a future satisfied when either (a) the cluster is updated or (b)
-   *     an unretriable error occurs or (c) polling or retry policy has been
+   *     an non-retryable error occurs or (c) polling or retry policy has been
    *     exhausted.
    *
    * @par Idempotency
@@ -757,7 +757,7 @@ class InstanceAdmin {
    * @param instance_id the instance to look the profile in.
    * @param profile_id the id of the profile within that instance.
    * @return a future satisfied when either (a) the profile is fetched or (b)
-   *     an unretriable error occurs or (c) retry policy has been exhausted.
+   *     an non-retryable error occurs or (c) retry policy has been exhausted.
    *
    * @par Idempotency
    * This operation is read-only and therefore it is always idempotent.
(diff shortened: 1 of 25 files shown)


remyabel:
I also spotted two issues which I'll defer judgment on how to fix. In
https://github.com/googleapis/google-cloud-cpp/blob/687adc104118f7fefec9f0c9285de73d239f38ac/google/cloud/internal/async_retry_unary_rpc.h#L48
the tparams do not seem to match the template parameters?

In
https://github.com/googleapis/google-cloud-cpp/blob/1a97c13f03632ca6a49388b8761eeda4e95569b0/google/cloud/storage/signed_url_options.h#L215
the rest of the sentence is cut-off.

coryan (changes requested):
Thanks again for these, I do appreciate them.  A few nits we can solve before kicking off the builds.

The signed url option should say something like "If this option is set, the result is in `https://mybucket.storage.googleapis.com`, while otherwise it is in `https://storage.googleapis.com/mybucket` format".

The `Sig` tparam in async_retry_unary_rpc.h should be removed (aside: we do not really generate Doxygen for `internal/` stuff, though maybe we should, if nothing else for ourselves).

> Reviewed 23 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5090#-:-MHY2GVP1jsIt5kUVmLV:b-lkhuo9)* status: all files reviewed, 2 unresolved discussions (waiting on @remyabel)

___
*[google/cloud/bigtable/instance_admin.h, line 893 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5090#-MHY1J3q9V3W_LW-mZRQ:-MHY1J3q9V3W_LW-mZRR:bqz0ysw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/beb66bc83128a12e72a35a5d93190f17b04edb79/google/cloud/bigtable/instance_admin.h#L893)):*
> ```objc
>    * @return a future satisfied when either (a) the app profile is deleted or
>    *     (b) an non-retryable error occurs or (c) retry policy has been
>    * exhausted.
> ```

please indent ` * exhausted.` -> ` *     exhausted`.
___
*[google/cloud/bigtable/table.h, line 750 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5090#-MHY1ZAZ6FPYl5TiHQiO:-MHY1ZA_-T-NdWtpA0ZG:b-i10fyf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/beb66bc83128a12e72a35a5d93190f17b04edb79/google/cloud/bigtable/table.h#L750)):*
> ```objc
>    * @param filter a filter expression, can be used to select a subset of the
>    *     column families and columns in the row.
>    * @returns a future satisfied when the operation completes, failed
> ```

s/failed/fails/ I think?

google/cloud/bigtable/instance_admin.h
```diff
@@ -324,7 +324,7 @@ class InstanceAdmin {
    * @param instance_update_config config with modified instance.
    *
    * @return a future satisfied when either (a) the instance is updated or (b)
-   *     an unretriable error occurs or (c) polling or retry policy has been
+   *     an non-retryable error occurs or (c) polling or retry policy has been
```
devbww, line 327: s/an non-/a non-/ et seq.

remyabel: Done. Thanks.

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5090#-:-MHYCAYZ54VCTKZR8fTf:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @devbww)

PR #5095, opened 2020-09-19
fix(storage): consistent endpoint computation

### LINKED ISSUE #5094: Inconsistent computation of XML vs. JSON endpoints
The storage library is inconsistent in how it computes the XML vs. JSON endpoints when both `CLOUD_STORAGE_TESTBENCH_ENDPOINT` is set and `ClientOptions::set_endpoint()` is called.

- For XML, the value from `CLOUD_STORAGE_TESTBENCH_ENDPOINT` is used.
- For JSON, the value from `ClientOptions::set_endpoint()` is used.

This is also inconsistent with our other libraries, where the environment variable takes precedence if set.

coryan:
The interaction of `ClientOptions::set_endpoint()` and the
`CLOUD_STORAGE_TESTBENCH_ENDPOINT` environment variable was inconsistent
across endpoints. For JSON endpoints `set_endpoint()` overrode the
`CLOUD_STORAGE_TESTBENCH_ENDPOINT` value, while for XML endpoints it was
the opposite.

In other libraries the environment variable always wins, so we are
changing the behavior here. This behavior was never documented, and it
was buggy, this is not a breaking change per-se, but should be
highlighted in the CHANGELOG.

I refactored the code to compute the different endpoints to standalone
functions. That makes them easier to test.

I used friend functions in `internal::` over member functions in
`ClientOptions` because we will (some day) remove the XML endpoints and
I do not want to make `XmlFoo()` part of the public interface.

Fixes #5094 part of the work for #5072 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5095)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,22 @@
+
 # Changelog
 
 ## v1.19.0 - TBD
+
+### Storage
+
+* fix: consistent computation of XML vs. JSON endpoints
+
+The interaction of `ClientOptions::set_endpoint()` and the
+`CLOUD_STORAGE_TESTBENCH_ENDPOINT` environment variable was inconsistent
+across endpoints. For JSON endpoints `set_endpoint()` overrode the
+`CLOUD_STORAGE_TESTBENCH_ENDPOINT` value, while for XML endpoints it
+was the opposite.
+
+In other libraries the environment variable always wins, so we are changing the
+behavior here. This behavior was never documented, and it was buggy, therefore
+it is not a breaking change. Nonetheless, we think the bug (and the fix) is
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,23 @@
+
 # Changelog
 
 ## v1.19.0 - TBD
 
+### Storage
+
+* fix: consistent computation of XML vs. JSON endpoints
```
devjgm, line 8: This is an interesting pattern you're doing here. Maybe we should do this more often when making a change that requires a nuanced description in the changelog. I like this. Thanks.

devjgm (approved):

Issue #5106, opened 2020-09-21
Build failure with MSAN


coryan:
This started recently:

https://source.cloud.google.com/results/invocations/bb59c9cc-4757-4b1c-8388-609e08ec5916/details

```
ERROR: /v/google/cloud/spanner/BUILD:73:9: Couldn't build file google/cloud/spanner/_objs/database_admin_connection_test/database_admin_connection_test.pic.o: undeclared inclusion(s) in rule '//google/cloud/spanner:database_admin_connection_test':
this rule is missing dependency declarations for the following files included by 'google/cloud/spanner/database_admin_connection_test.cc':
  '/usr/lib64/clang/10.0.1/share/msan_blacklist.txt'
  '/usr/lib64/clang/10.0.1/include/stddef.h'
  '/usr/lib64/clang/10.0.1/include/__stddef_max_align_t.h'
  '/usr/lib64/clang/10.0.1/include/stdint.h'
  '/usr/lib64/clang/10.0.1/include/stdarg.h'
  '/usr/lib64/clang/10.0.1/include/limits.h'
  '/usr/lib64/clang/10.0.1/include/float.h'
```

coryan:
Fixed by clearing some stale cache files.

mr-salty:
I noticed this in the log:

```
2020-09-22T04:11:04Z: Downloading build cache cache-ci-fedora-libcxx-msan-32-msan.tar.gz from cloud-cpp-kokoro-results/build-cache/google-cloud-cpp
ERROR: gcloud crashed (TokenRevokeError): invalid_request
```

does that result in a corrupt/truncated tarball? I don't see any errors related to untarring it so could be a red herring.

also, I wonder if we should `set -x` in all our shell scripts (prints commands being executed)... unsure if it would be helpful in debugging or just noise.

coryan:
The error in the log is unrelated to downloading the tarball and/or extracting it.

devjgm:
This was probably fixed by https://github.com/googleapis/google-cloud-cpp/pull/5108

mr-salty:
@coryan sounds like a vote for `set -x` to me?

coryan:
@mr-salty I can see the argument for `set -x`, OTOH I like the cleaner logs. Send a PR and we can see if the logs are tolerable, or file a bug if you don't have the time right now.

coryan:
[...]

PR #5108, opened 2020-09-22
ci: use per-build prefix for Bazel caches

coryan:
Using a fresh cache for the msan build fixes the problem.  I am splitting each build to its own
cache directory, so we can clear the cache for one build at a time in the future.

Part of the work for #5106

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5108)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -66,11 +66,7 @@ readonly BAZEL_CACHE="https://storage.googleapis.com/cloud-cpp-bazel-cache"
 # not hit this cache.
 if [[ -r "${GOOGLE_APPLICATION_CREDENTIALS}" ]]; then
   io::log "Using bazel remote cache: ${BAZEL_CACHE}"
-  if [[ "${BUILD_NAME:-}" = "msan" ]]; then
-    bazel_args+=("--remote_cache=https://storage.googleapis.com/cloud-cpp-bazel-cache-${BUILD_NAME}")
-  else
-    bazel_args+=("--remote_cache=${BAZEL_CACHE}")
-  fi
+  bazel_args+=("--remote_cache=https://storage.googleapis.com/cloud-cpp-bazel-cache/docker/${BUILD_NAME}")
   bazel_args+=("--google_credentials=${GOOGLE_APPLICATION_CREDENTIALS}")
   # See https://docs.bazel.build/versions/master/remote-caching.html#known-issues
   # and https://github.com/bazelbuild/bazel/issues/3360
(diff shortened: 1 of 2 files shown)


devjgm (approved):

coryan:
FWIW, I am not sure this is the right fix, it built on the PR, but did not build for the CI build:

https://source.cloud.google.com/results/invocations/124e5935-62bb-4fd3-a955-542ec3484b27

mr-salty:
Note that the build did pass but took [97 minutes](https://source.cloud.google.com/results/invocations/0ee69eb6-2e16-42bd-995a-584fe2b4f9a9/targets)

coryan:
> Note that the build did pass but took [97 minutes](https://source.cloud.google.com/results/invocations/0ee69eb6-2e16-42bd-995a-584fe2b4f9a9/targets)

That was a build with a cold cache.  Builds with warm caches are taking more reasonable times, e.g. the [latest](https://source.cloud.google.com/results/invocations/4fca0f8e-c996-4d8f-8aa7-2565db486ffc/targets) took just under 10m.

ci/kokoro/docker/build-in-docker-bazel.sh
```diff
@@ -66,7 +66,7 @@ readonly BAZEL_CACHE="https://storage.googleapis.com/cloud-cpp-bazel-cache"
 # not hit this cache.
 if [[ -r "${GOOGLE_APPLICATION_CREDENTIALS}" ]]; then
   io::log "Using bazel remote cache: ${BAZEL_CACHE}"
-  bazel_args+=("--remote_cache=${BAZEL_CACHE}")
+  bazel_args+=("--remote_cache=https://storage.googleapis.com/cloud-cpp-bazel-cache/docker/${BUILD_NAME}")
```
devjgm, line 69: Did you want to use `${BAZEL_CACHE}` here rather than retyping part of the URL string?

devjgm (approved):

ci/kokoro/docker/build-in-docker-bazel.sh
coryan: Fixed.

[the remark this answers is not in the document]
coryan: > Just so I'm clear, this change is purely to give us the ability to clear this cache in the future _if_ we discover a problem w/ bazel's remote cache. We've currently not seen it cause any problems, nor do we have any need to clear this cache, is that right?

Effectively I cleared the cache by starting from a fresh one.  And yes, we are seeing problems.  What *exactly* are the problems I am not sure.


All I know is that the PR builds with both the tarball and the existing Bazel cache break, that removing the tarball fixed some builds but later they started failing again, and that clearing both caches seems to have fixed the builds for good.  I don't know, maybe the underlying Docker image used to create the tarball matters and Bazel does not include enough of it into the cache inputs.

I am not going to spend time troubleshooting Bazel, clearing the caches worked.

> 
> This will likely reduce our cache hit rate a little, but probably a negligible amount,

I think the amount will be zero: the Bazel builds all have different command-line options (*san vs. libcxx vs. Windows vs. macOS) all of which are caching inputs.  I think none of the artifacts created by build A are used in build B.  I guess maybe some of the `genrule()` artifacts are reusable... hmm... I am wrong it is not zero, but it is something like 1/5,000.

PR #5111, opened 2020-09-22
feature(generator): run unit tests on golden code

scotthart:
This PR adds unit tests for the .golden code.
As part of this:
- the .proto files in generator/integration_test are built in to a library
- the .golden files are copied (as part of the build process) to google/cloud/test/... and built into a library
- unit tests for the stub, logging decorator, and metadata decorator are run using these libraries and new test code (the logging and metadata test code is very similar to the spanner logging and metadata decorator testing).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5111)


generator/integration_tests/BUILD
```diff
@@ -16,19 +16,58 @@ package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])  # Apache 2.0
 
+load("@rules_proto//proto:defs.bzl", "proto_library")
+
+proto_library(
```
devjgm, line 21: As we chatted about offline, after merging this change, please take a look at the test import script and see if you can make the needed changes to save the operator from having to replicate these BUILD changes internally.

devjgm (approved):
As we discussed offline, please consider in the future if we can get to a world where the "golden" files have normal file extensions like `.h` and `.cc`, where they have normal CMake and Bazel build rules to compile and test them, and they're subject to our normal clang-format and clang-tidy checks. As we discussed offline, it seems like this is likely possible and might be a win for us. But I'll leave the thorough thinking to you. Thanks.

PR #5112, opened 2020-09-22
cleanup(spanner): Use CompletionQueue for long running operations

### LINKED ISSUE #4038: Use google::cloud::*::CompletionQueue to implement long running operations.
The implementation of long running operation wrappers works nicely when one uses a `CompletionQueue` (see the Cloud Bigtable code). Once we have refactored those to a good place we should cleanup the code in `google-cloud-cpp-spanner`.

devbww:
Use `google::cloud::CompletionQueue` via (`BackgroundThreads`)
to execute long-running operations in `DatabaseAdminConnection`
and `InstanceAdminConnection`, rather than directly forking a
`std::thread`.

Fixes #4038.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5112)

### DIFF
--- a/google/cloud/spanner/database_admin_connection.cc
+++ b/google/cloud/spanner/database_admin_connection.cc
@@ -120,23 +120,50 @@ std::unique_ptr<PollingPolicy> DefaultAdminPollingPolicy() {
       .clone();
 }
 
+template <typename ValueExtractor>
+struct PollingLoop {
+  std::shared_ptr<internal::DatabaseAdminStub> polling_stub;
+  google::longrunning::Operation operation;
+  std::unique_ptr<PollingPolicy> polling_policy;
+  google::cloud::promise<typename ValueExtractor::ReturnType> promise;
+  char const* location;
+
+  void operator()() {
+    auto stub = std::move(polling_stub);
+    auto result = internal::PollingLoop<ValueExtractor>(
+        std::move(polling_policy),
+        [stub](grpc::ClientContext& context,
+               google::longrunning::GetOperationRequest const& request) {
+          return stub->GetOperation(context, request);
+        },
+        std::move(operation), location);
+    // Drop our reference to stub. Ideally we'd have std::moved into the
+    // internal::PollingLoop() lambda.
+    stub.reset();
+    promise.set_value(std::move(result));
+  }
+};
+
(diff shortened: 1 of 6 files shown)


coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5112)* status: all files reviewed, 2 unresolved discussions (waiting on @devbww)

___
*[google/cloud/spanner/database_admin_connection.cc, line 137 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5112#-MHpbuwa2U3APce5Ak-M:-MHpbuwa2U3APce5Ak-N:b-f3hdxb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/81c821d370bb652a2e4347fa9101b53c7de5dbb6/google/cloud/spanner/database_admin_connection.cc#L137)):*
> ```cpp
>         [stub](grpc::ClientContext& context,
>                google::longrunning::GetOperationRequest const& request) {
>           return stub->GetOperation(context, request);
> ```

While this avoids creating a thread, note that it blocks the background thread while this `GetOperation` is running *and* while the polling backoff is sleeping.  Since we are using a `CompletionQueue` it would be preferable to use `AsyncGetOperation()` (and some asynchronous sleeping) to avoid blocking at all.

As the number of background threads is limited, typically just the one, and the backoff between polling periods is rather long (seconds), this will exhaust the background threads very quickly.  Just one call at a time will be the most you can do.
___
*[google/cloud/spanner/instance_admin_connection.cc, line 54 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5112#-MHpcj-dBCYi4NzOzcWC:-MHpcj-eDNDMIWPXg9p0:b-jz65ym) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/81c821d370bb652a2e4347fa9101b53c7de5dbb6/google/cloud/spanner/instance_admin_connection.cc#L54)):*
> ```cpp
> 
> template <typename ValueExtractor>
> struct PollingLoop {
> ```

Same comments as above, it might be worthwhile refactoring this now, maybe.

coryan:
Should we close this until you are ready to get back to it?

devbww:
Done (and the other one).

PR #5117, opened 2020-09-22
ci: disable tarball cache on full CI Bazel builds

### LINKED ISSUE #5113: Disable tarball cache for all Bazel CI builds
I missed that we were turning on the Bazel tarball caches for all CI builds (I thought it was just CMake).

I believe this is not a good idea, Bazel has no mechanism to prune the local cache. If the CI build downloads the old tarball the cache is populated with old data, it then gets the new data during the CI build, and then gets uploaded with the old and the new data for the PRs **and** for the next full CI build.  Effectively the tarball grows without any limits (note that ccache is limited to 4.0GiB by configuration, so it does not have this problem).

We disabled the download for full CI builds to avoid this, the tarball created in each full CI build was "fresh".

coryan:
Fixes #5113 

----
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5117)

### DIFF
--- a/ci/kokoro/docker/build.sh
+++ b/ci/kokoro/docker/build.sh
@@ -28,6 +28,10 @@ export CMAKE_SOURCE_DIR="."
 
 export GOOGLE_CLOUD_CPP_GENERATOR_RUN_INTEGRATION_TESTS="yes"
 
+# Define the build tool so other scripts (e.g. cache download),
+# currently the possible values are `CMake` or `Bazel`.
+export BUILD_TOOL="CMake"
+
 in_docker_script="ci/kokoro/docker/build-in-docker-cmake.sh"
 
 if [[ $# -ge 1 ]]; then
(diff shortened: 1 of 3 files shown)


ci/kokoro/docker/build.sh
```diff
@@ -28,6 +28,10 @@ export CMAKE_SOURCE_DIR="."
 
 export GOOGLE_CLOUD_CPP_GENERATOR_RUN_INTEGRATION_TESTS="yes"
 
+# Define the build tool so other scripts (e.g. cache download),
+# currently the possible values are `CMake` or `Bazel`.
+export BUILD_TOOL="CMake"
```
devjgm, line 33: NO CHANGE NECESSARY: I wonder if in a future PR we could call the tarball-cache functions from the in-docker build scripts, which would eliminate the need to specify the BUILD_TOOL separately the existing `in_docker_script`.

devjgm (approved):
Thanks for fixing this. This was accidentally enabled last week.

ci/kokoro/docker/build.sh
coryan: Yeah, I have been wondering something like that myself.  We would need to figure out how to pass the credentials file to the in-docker script, and make sure `gsutil` is installed and configured.

PR #5119, opened 2020-09-23
DO NOT REVIEW

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5119)


ci/cspell.json
```diff
@@ -5,6 +5,7 @@
   "language": "en",
   // words - list of words to be always considered correct
   "words": [
+    "absl",
```
coryan, line 8: I think we expected this to be spelled in backticks for comments, .i.e. 

```
// `absl::Foo`
```

scotthart: Thanks, good to know.

Issue #4905, opened 2020-08-14
[Flake] Test timeout, maybe too many buckets?


coryan:
This test failed:

https://source.cloud.google.com/results/invocations/91277c88-f2a0-43ba-8b6c-4ee5922ad2eb

with:

```
==================== Test output for //google/cloud/storage/examples:storage_bucket_samples:

Running ListBucketsForProject() example
artifacts.cloud-cpp-testing-resources.appspot.com
bm-throughput-vs-cpu-6arwin4if5zvxt125ycks1r1vqayzh8ti6oep27fzn
bm-throughput-vs-cpu-gpqfcnmj9s0y6hk5i4gk1reoc425ofphxw407ri6gv
bm-throughput-vs-cpu-y6rmm4i8pi9nfn2ke1qdepxyk9srfmrb98mwybzr48
cloud-cpp-bazel-cache
cloud-cpp-test-examples-00g06ipntirp9x4xndweyjrmz6qoano042v5air
cloud-cpp-test-examples-01292sh2l1hgody1phjcraqv18oe129tpkadvzh
cloud-cpp-test-examples-01gg9byjm2l4qpxeyoa1c71b8lrg8od1os9noz5
cloud-cpp-test-examples-0205o89cf2txzgeyawd9ilxqddx9dr21vcksejd
cloud-cpp-test-examples-05927jl7wc9vsrnwtwvmwogp5jp8n629psqxhta

...
//google/cloud/storage/examples:storage_bucket_samples                  TIMEOUT in 300.0s
  C:/b/rvommcsc/execroot/com_github_googleapis_google_cloud_cpp/bazel-out/x64_windows-fastbuild/testlogs/google/cloud/storage/examples/storage_bucket_samples/test.log
```

The test seems to timeout because the list is too long. I have cleaned up the project, but we probably should automate that process.

PR #5136, opened 2020-09-23
test: use helper to cleanup sample buckets

coryan:
Previously the examples would remove any temporary buckets using just
`client.DeleteBucket()`, but if the sample code accidentally left an
object or two in the temporary bucket the `DeleteBucket()` call fails
and we leak the bucket (and its objects).

Part of the work for #4905

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5136)

### DIFF
--- a/google/cloud/storage/examples/CMakeLists.txt
+++ b/google/cloud/storage/examples/CMakeLists.txt
@@ -54,7 +54,8 @@ if (BUILD_TESTING)
 
     set(storage_examples_unit_tests
         # cmake-format: sort
-        storage_client_mock_samples.cc storage_examples_common_test.cc)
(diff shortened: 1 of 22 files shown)


devbww (approved):
Q. Does that mean samples should not even try to delete bucket contents first?

coryan:
> Q. Does that mean samples should not even try to delete bucket contents first?

Yes.  Unless we think that relying on this function at the end is "obscure". 

FTR, I am going to send another PR (later) to have the samples delete any buckets left over from previous runs, for example, because the previous build crashed or was killed before it had a chance to clean after itself.

PR #5142, opened 2020-09-23
ci: use GCS Bazel cache for macOS quickstart build

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5142)


coryan:
The build time is about 5m with a warm cache: https://source.cloud.google.com/results/invocations/d728f2a8-9eab-41c0-b605-45168382f1b6/targets

devbww (approved):

PR #5144, opened 2020-09-24
generator(refactor): simplify golden testing

scotthart:
This PR removes the .golden suffix from the generator golden test files. This coupled with changing the --product-path parameter in the integration test removes the need to rename the golden files or copy to another location making the cmake and bazel files simpler.

While presubmit clang-tidy does not work due to missing .grpc.pb.h files, clang-tidy when run in cmake does work now that the golden files end with the expected .h and .cc. Minor changes were made to the generators to fix some clang-tidy issues as well as introducing golden versioning files. Versioning files were necessary to supply a #define for the inline namespace macro. Generators for the versioning files will be added later.

Clang-format has been disabled for the golden files as due to version incompatibilities with clang-format in the testing process. This in no way impacts the ability to clang-format generated code for actual services that we plan to check in.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5144)


coryan (approved):
:lgtm:

> Reviewed 24 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5144#-:-MI-I6nG-aNxMsfNyPOi:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devjgm)

___
*[generator/internal/stub_generator.cc, line 75 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5144#-MI-Hpir2g9Ia5iP1wA9:-MI-Hpis7UV88idWyIDG:b-aausu5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4384b607c0e6e82fb3886333482128a0e55808bc/generator/internal/stub_generator.cc#L75)):*
> ```cpp
>   // includes
>   header_.Print(
>       LocalInclude(absl::StrCat(service_vars_["product_path"], "version.h")));
> ```

FYI: for a future PR: consider whether per-library versions are useful, since they all have the same values.  Maybe a single version.h is enough.

PR #5151, opened 2020-09-24
cleanup: fix bucket name prefix in integration tests

coryan:
The integration tests could, but rarely did, use a different prefix for
the random bucket names used in the test. That makes it hard to clean
these up later. In this change we fix the prefix and do some refactoring
to always use `storage::testing::MakeRandomBucketName()`.

Part of the work for #4905

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5151)

### DIFF
--- a/google/cloud/storage/examples/storage_examples_common.cc
+++ b/google/cloud/storage/examples/storage_examples_common.cc
@@ -37,8 +37,7 @@ std::string MakeRandomBucketName(google::cloud::internal::DefaultPRNG& gen) {
   static std::size_t const kMaxBucketNameLength = 63;
   auto const date =
       absl::FormatCivilTime(absl::ToCivilDay(absl::Now(), absl::UTCTimeZone()));
-  auto const full =
-      std::string("cloud-cpp-testing-examples") + '_' + date + '_';
+  auto const full = std::string("cloud-cpp-testing-examples-") + date + '_';
   std::size_t const max_random_characters = kMaxBucketNameLength - full.size();
   // bucket names might also contain `-` and `_` characters, but we do not
   // *need* to use them.
@@ -98,7 +97,7 @@ Status RemoveBucketAndContents(google::cloud::storage::Client client,
 Status RemoveStaleBuckets(
     google::cloud::storage::Client client, std::string const& prefix,
     std::chrono::system_clock::time_point created_time_limit) {
-  std::regex re("^" + prefix + R"re(_\d{4}-\d{2}-\d{2}_.*$)re");
+  std::regex re("^" + prefix + R"re(-\d{4}-\d{2}-\d{2}_.*$)re");
   for (auto& bucket : client.ListBuckets()) {
     if (!bucket) return std::move(bucket).status();
     if (!std::regex_match(bucket->name(), re)) continue;
--- a/google/cloud/storage/examples/storage_examples_common.cc
+++ b/google/cloud/storage/examples/storage_examples_common.cc
@@ -31,13 +31,14 @@ bool UsingTestbench() {
               .empty();
 }
 
-std::string MakeRandomBucketName(google::cloud::internal::DefaultPRNG& gen,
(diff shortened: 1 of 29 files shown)


google/cloud/storage/testing/storage_integration_test.cc
```diff
@@ -121,19 +121,13 @@ std::unique_ptr<RetryPolicy> StorageIntegrationTest::TestRetryPolicy() {
...
 }
 
-std::string StorageIntegrationTest::MakeRandomBucketName(std::string prefix) {
+std::string StorageIntegrationTest::MakeRandomBucketName() {
   // The total length of this bucket name must be <= 63 characters,
-  char constexpr kPrefix[] = "gcs-cpp-test-bucket-";  // NOLINT
+  char constexpr kPrefix[] = "cloud-cpp-testing-";  // NOLINT
   auto constexpr kMaxBucketNameLength = 63;
   static_assert(kMaxBucketNameLength > sizeof(kPrefix),
                 "The bucket prefix is too long");
```
devbww, line 129: It looks like this constant/check no longer applies here.  Indeed, `testing::MakeRandomBucketName()` adds to the prefix, so the check should really happen there (or at least it shouldn't wrap `max_random_characters`).

coryan: Good point, removed.

google/cloud/storage/testing/storage_integration_test.cc
```diff
@@ -122,12 +122,7 @@ std::unique_ptr<RetryPolicy> StorageIntegrationTest::TestRetryPolicy() {
 }
 
 std::string StorageIntegrationTest::MakeRandomBucketName() {
-  // The total length of this bucket name must be <= 63 characters,
-  char constexpr kPrefix[] = "cloud-cpp-testing-";  // NOLINT
-  auto constexpr kMaxBucketNameLength = 63;
-  static_assert(kMaxBucketNameLength > sizeof(kPrefix),
-                "The bucket prefix is too long");
-  return testing::MakeRandomBucketName(generator_, kPrefix);
+  return testing::MakeRandomBucketName(generator_, "cloud-cpp-testing-");
```
devbww, line 125: Vague suggestion:  We might like the trailing dash to be an underscore given than `testing::MakeRandomBucketName()` adds `"YYYY-MM-DD_"`.

devbww (approved):

google/cloud/storage/testing/storage_integration_test.cc
coryan: I like it.  I made that change in this pass, I will clean this further in the next pass (I am putting it off because I will need to change dependencies and I want that to be "pure" refactoring).

google/cloud/storage/examples/storage_examples_common.cc
```diff
@@ -97,7 +98,7 @@ Status RemoveBucketAndContents(google::cloud::storage::Client client,
 Status RemoveStaleBuckets(
     google::cloud::storage::Client client, std::string const& prefix,
     std::chrono::system_clock::time_point created_time_limit) {
-  std::regex re("^" + prefix + R"re(\d{4}-\d{2}-\d{2}_.*$)re");
+  std::regex re("^" + prefix + R"re(_\d{4}-\d{2}-\d{2}_.*$)re");
```
devbww, line 101: Do we still need/want the `prefix` here?

google/cloud/storage/testing/storage_integration_test.cc
coryan: Argh, that won't work...

google/cloud/storage/examples/storage_examples_common.cc
coryan: I think so?  Maybe we can revisit when I refactor the `RemoveStaleBuckets()` function.

Issue #4501, opened 2020-07-11
Include Abseil headers with angle brackets


coryan:
We need to include Abseil headers with angle brackets because on some platforms their headers are not clean under `-Werrror`. We can work around this by disabling `-Werror`, but obviously that is not a great answer because that becomes a transitive problem for our customers: they have to disable -Werror too because our headers are not clean.

For an specific example abseil/abseil-cpp#582 breaks under Windows+MSVC+x86.

devjgm:
I think Abseil will need to do some work to support angle-bracket includes in Bazel builds. We should file a feature request for this.

coryan:
Done: abseil/abseil-cpp#740

coryan:
When compiled with MSVC+x86 some Abseil headers (notably "absl/strings/str_join.h") generate warnings, and break our builds (as we compile with warnings-as-errors). This particular warning is useful, and we do not want to disable it globally. Instead we disable the warning before including "absl/strings/str_join.h", and enable it again after the include.

I have refactored the code to silence the warning, include `"absl/strings/str_join.h"` and re-enable the warnings to a single header (see #5163). When we have a reliable mechanism to disable warnings generated in Abseil headers (e.g. when `#include <absl/...>` works) we can remove that header and replace its call sites with `#include <absl/strings/str_join.h>` (or include with quotes if that works).

Removing this file (or any such file) should be a simple global search&replace followed by `clang-format` to reorder the includes.

devjgm:
Doesn't seem like this is going to happen any time soon. This is difficult to do w/o support from Bazel.

PR #5163, opened 2020-09-25
cleanup: quiet MSVC+x86 in a single place

coryan:
Including "absl/strings/str_join.h" triggers a warning (and we have
warnings-as-errors enabled). The warning is useful, we don't want to
quiet it everywhere, we just want to silence it for that abseil header.
Instead of sprinkling the directives in each include, let's just wrap
the include with one that also silences things. Once Abseil is quiet we
can fix this with a single search+replace+reformat.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5163)


mr-salty (approved):
:lgtm: modulo adding a bit more documentation

> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5163#-:-MI5uw7bEc6X2r8v0NoT:b27gsmd)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[google/cloud/internal/absl_str_join_quiet.h, line 18 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5163#-MI5uKO-4o40AicOEcmT:-MI5uKO-4o40AicOEcmU:b-yn9po8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77b36eda31b901d3509739e9bf535dc1497d9333/google/cloud/internal/absl_str_join_quiet.h#L18)):*
> ```objc
> #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_ABSL_STR_JOIN_QUIET_H
> 
> // TODO(#4501) - fix by doing #include <absl/...>
> ```

can you add some comments about what this file is doing and why?

Also, specifically for the TODO, what happens if/when #4501 does get fixed?  Can we eliminate this file replace inclusions with `#include <abls/strings/str_join.h>` or will this still be necessary for some reason?  I'd state that as well so we don't forget.

mr-salty (commented):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5163)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[google/cloud/internal/absl_str_join_quiet.h, line 18 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5163#-MI5uKO-4o40AicOEcmT:-MI61TpD3U4YzsrMoplT:b32vy8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77b36eda31b901d3509739e9bf535dc1497d9333/google/cloud/internal/absl_str_join_quiet.h#L18)):*

can you add some comments about what this file is doing and why?

Also, specifically for the TODO, what happens if/when #4501 does get fixed?  Can we eliminate this file replace inclusions with `#include <abls/strings/str_join.h>` or will this still be necessary for some reason?  I'd state that as well so we don't forget.
</blockquote></details>

LG

PR #5166, opened 2020-09-28
test(storage): add init bucket and filter_reponse

vnvo2409:
This PR adds:
- Initialization for `Bucket`.
- Some helper functions to support REST `fields` completely. ( Now we can do something like `fields=kind, items(acl(role, entity), name)`)

Related to #4751 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5166)

### DIFF
--- a/google/cloud/storage/emulator/gcs/__init__.py
+++ b/google/cloud/storage/emulator/gcs/__init__.py
@@ -12,10 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Tests entry"""
-
-from tests import test_utils, test_gcs
-
-if __name__ == "__main__":
-    test_utils.run()
-    test_gcs.run()
+from gcs import bucket
--- a/google/cloud/storage/emulator/gcs/bucket.py
+++ b/google/cloud/storage/emulator/gcs/bucket.py
@@ -64,12 +64,6 @@ class Bucket:
             )
             valid &= not bucket_name.startswith("goog")
             valid &= re.search("g[0o][0o]g[1l][e3]", bucket_name) is None
-            valid &= (
-                re.match(
-                    "^[0-9]{1,3}[.][0-9]{1,3}[.][0-9]{1,3}[.][0-9]{1,3}$", bucket_name
-                )
-                is None
-            )
         if not valid:
             utils.error.invalid("Bucket name %s" % bucket_name, context)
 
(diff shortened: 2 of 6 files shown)


coryan (changes requested):
A few requests below.  Let me know what you think.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5166#-:-MIKIeUh8VrL9ixxdtUA:b-js1scm)* status: all files reviewed, 2 unresolved discussions (waiting on @vnvo2409)

___
*[google/cloud/storage/emulator/gcs/bucket.py, line 69 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5166#-MIKHqO8EHfd4UkW_jCM:-MIKHqO8EHfd4UkW_jCN:b-ivz9h4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6464b36fe23dd21fff3b67778a3799f69d978014/google/cloud/storage/emulator/gcs/bucket.py#L69)):*
> ```Python
>             valid &= (
>                 re.match(
>                     "^[0-9]{1,3}[.][0-9]{1,3}[.][0-9]{1,3}[.][0-9]{1,3}$", bucket_name
> ```

Is this trying to match IP addresses?  Why only IPv4 and not IPv6?  Also, it seems like it would never match because this is only used if `if '.' not in bucket_name:` right?  I would just remove it, we do not need the emulator to be *perfect* we need it to be good enough to help us with testing the libraries.
___
*[google/cloud/storage/emulator/gcs/bucket.py, line 181 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5166#-MIKHFZGCgToKZf_Q-3A:-MIKHFZH4maZO_Mwo6_D:bqch8sl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6464b36fe23dd21fff3b67778a3799f69d978014/google/cloud/storage/emulator/gcs/bucket.py#L181)):*
> ```Python
>     @classmethod
>     def init_iam_policy(cls, metadata, context):
>         # TODO(vnvo2409): Check for iamConfiguration.uniformBucketLevelAccess.enabled
> ```

Please create a GitHub issue and change this to `TODO(#<issue-number>)`

vnvo2409:
> Is this trying to match IP addresses? Why only IPv4 and not IPv6? Also, it seems like it would never match because this is only used if `if '.' not in bucket_name:` right? I would just remove it, we do not need the emulator to be perfect we need it to be good enough to help us with testing the libraries.

I copied from the old testbench without thiking too much. I removed it in the latest commit.

> Please create a GitHub issue and change this to `TODO(#<issue-number>)`

This case will most likely not happen in the integration test. So I removed it in the lastest commit too.

coryan (approved):
:lgtm: the build breakage is fixed by #5167. I will wait for that to get fixed before merging these changes.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5166#-:-MIL98h9Dxg-2s7Z4p4y:b-crww9g)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5178, opened 2020-10-03
doc(pubsub): use codition variables in sample code

coryan:
We prefer using condition variables over promises and futures in the
sample code, as C++ developers should be familiar with the former, but
may not be familiar with the latter.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5178)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -1071,22 +1071,29 @@ void CustomThreadPoolSubscriber(std::vector<std::string> const& argv) {
     auto subscription =
         pubsub::Subscription(std::move(project_id), std::move(subscription_id));
 
+    // Because this is an example we want to exit eventually, use a mutex and
+    // condition variable to notify the current thread and stop the example.
     std::mutex mu;
+    std::condition_variable cv;
     int count = 0;
-    google::cloud::promise<void> received_message;
+    auto wait = [&] {
+      std::unique_lock<std::mutex> lk(mu);
+      cv.wait(lk, [&] { return count >= 4; });
+    };
+    auto increase_count = [&] {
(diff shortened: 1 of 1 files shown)


google/cloud/pubsub/samples/samples.cc
```diff
@@ -1071,22 +1071,29 @@ void CustomThreadPoolSubscriber(std::vector<std::string> const& argv) {
     auto subscription =
         pubsub::Subscription(std::move(project_id), std::move(subscription_id));
 
+    // Because this is an example we want to exit eventually, use a mutex and
+    // condition variable to notify the current thread and stop the example.
     std::mutex mu;
+    std::condition_variable cv;
```
devbww, line 1077: Nit: Because it is an example, we should probably also explicitly include `<mutex>` and `<condition_variable>`.

google/cloud/pubsub/samples/samples.cc
```diff
@@ -1071,22 +1071,29 @@ void CustomThreadPoolSubscriber(std::vector<std::string> const& argv) {
...
     int count = 0;
-    google::cloud::promise<void> received_message;
+    auto wait = [&] {
```
devbww, line 1079: s/wait/await_count/ perhaps?

devbww (approved):

google/cloud/pubsub/samples/samples.cc
coryan: Done.

google/cloud/pubsub/samples/samples.cc
coryan: Done.

devbww (approved):

PR #5189, opened 2020-10-07
fix: handling of session URLs in gRPC

### LINKED ISSUE #5030: `session_url` is actually `session_id` in gRPC
When a resumable session is started, a resumable session id is returned.

In cURL mode, this is actually the whole URL with all parameters bound. Unfortunately, in gRPC it is only the id token itself.

This causes the following problems:
* gRPC and cURL implementations cannot be mixed
* some parts of the `GrpcClient` API (e.g. `DeleteResumableSession`) cannot be easily implemented (they also need the bucket and the object, which are parts of the URL)

There are two possible approaches:
* change the internal and external APIs to make the resumable session id an object (perhaps with distinct implementations for gRPC and cURL)
* synthesize the whole URL in the `GrpcClient`

I think we should go with the latter because:
* a session id (a string currently) is a part of the public API
* we will get interchangeability of the cURL and gRPC APIs

The implementation is quite obscure, though, because we need to generate and parse strings.

dopiera:
This fixes #5030.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,6 +152,9 @@ cmake_dependent_option(
     ON "${GOOGLE_CLOUD_CPP_ENABLE_GRPC_EXPRESSION}" OFF)
 mark_as_advanced(GOOGLE_CLOUD_CPP_ENABLE_GRPC)
 
+# Building this target results in all protobufs being compiled.
+add_custom_target(google-cloud-cpp-protos)
+
 if (${GOOGLE_CLOUD_CPP_ENABLE_GRPC})
     add_subdirectory(external/googleapis)
 endif ()
--- a/cmake/CompileProtos.cmake
+++ b/cmake/CompileProtos.cmake
@@ -42,6 +42,9 @@ endfunction ()
 # Use the `PROTO_PATH` option to provide one or more directories to search for
 # proto files in the import.
 #
+# If `LOCAL_INCLUDE` is set to true, `file.pb.cc` file generated for
+# `a/b/file.proto` will include `file.pb.h` rather than `a/b/file.pb.h`.
+#
 # @par Example
 #
 # google_cloud_cpp_generate_proto( MY_PB_FILES "foo/bar/baz.proto"
@@ -53,7 +56,8 @@ endfunction ()
 # `foo/bar/baz.proto` then the directory containing `foo` must be in the search
 # path.
 function (google_cloud_cpp_generate_proto SRCS)
-    cmake_parse_arguments(_opt "" "" "PROTO_PATH_DIRECTORIES" ${ARGN})
+    cmake_parse_arguments(_opt "LOCAL_INCLUDE" "" "PROTO_PATH_DIRECTORIES"
+                          ${ARGN})
     if (NOT _opt_UNPARSED_ARGUMENTS)
         message(SEND_ERROR "Error: google_cloud_cpp_generate_proto() called"
                            " without any proto files")
@@ -73,11 +77,12 @@ function (google_cloud_cpp_generate_proto SRCS)
     endforeach ()
 
     set(${SRCS})
-    foreach (filename ${_opt_UNPARSED_ARGUMENTS})
-        get_filename_component(file_directory "${filename}" DIRECTORY)
+    foreach (file_path ${_opt_UNPARSED_ARGUMENTS})
+        get_filename_component(file_directory "${file_path}" DIRECTORY)
+        get_filename_component(file_name "${file_path}" NAME)
         # This gets the filename without the extension, analogous to $(basename
-        # "${filename}" .proto)
-        get_filename_component(file_stem "${filename}" NAME_WE)
+        # "${file_path}" .proto)
+        get_filename_component(file_stem "${file_path}" NAME_WE)
 
         # Strip all the prefixes in ${_opt_PROTO_PATH_DIRECTORIES} from the
         # source proto directory
@@ -91,15 +96,29 @@ function (google_cloud_cpp_generate_proto SRCS)
         set(pb_cc "${CMAKE_CURRENT_BINARY_DIR}/${D}/${file_stem}.pb.cc")
         set(pb_h "${CMAKE_CURRENT_BINARY_DIR}/${D}/${file_stem}.pb.h")
         list(APPEND ${SRCS} "${pb_cc}" "${pb_h}")
-        add_custom_command(
-            OUTPUT "${pb_cc}" "${pb_h}"
-            COMMAND
-                $<TARGET_FILE:protobuf::protoc> ARGS --cpp_out
-                "${CMAKE_CURRENT_BINARY_DIR}" ${protobuf_include_path}
-                "${filename}"
-            DEPENDS "${filename}" protobuf::protoc
-            COMMENT "Running C++ protocol buffer compiler on ${filename}"
-            VERBATIM)
+
+        if (${_opt_LOCAL_INCLUDE})
+            add_custom_command(
+                OUTPUT "${pb_cc}" "${pb_h}"
+                COMMAND
+                    $<TARGET_FILE:protobuf::protoc> ARGS --cpp_out
+                    "${CMAKE_CURRENT_BINARY_DIR}/${D}" ${protobuf_include_path}
+                    "${file_name}"
+                DEPENDS "${file_path}" protobuf::protoc
+                WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/${D}"
+                COMMENT "Running C++ protocol buffer compiler on ${file_path}"
+                VERBATIM)
+        else ()
+            add_custom_command(
+                OUTPUT "${pb_cc}" "${pb_h}"
+                COMMAND
+                    $<TARGET_FILE:protobuf::protoc> ARGS --cpp_out
+                    "${CMAKE_CURRENT_BINARY_DIR}" ${protobuf_include_path}
+                    "${file_path}"
+                DEPENDS "${file_path}" protobuf::protoc
+                COMMENT "Running C++ protocol buffer compiler on ${file_path}"
+                VERBATIM)
+        endif ()
     endforeach ()
 
     set_source_files_properties(${${SRCS}} PROPERTIES GENERATED TRUE)
@@ -230,16 +249,23 @@ function (google_cloud_cpp_install_proto_library_protos target)
 endfunction ()
 
 function (google_cloud_cpp_proto_library libname)
-    cmake_parse_arguments(_opt "" "" "PROTO_PATH_DIRECTORIES" ${ARGN})
+    cmake_parse_arguments(_opt "LOCAL_INCLUDE" "" "PROTO_PATH_DIRECTORIES"
+                          ${ARGN})
     if (NOT _opt_UNPARSED_ARGUMENTS)
         message(SEND_ERROR "Error: google_cloud_cpp_proto_library() called"
                            " without any proto files")
         return()
     endif ()
 
-    google_cloud_cpp_generate_proto(
-        proto_sources ${_opt_UNPARSED_ARGUMENTS} PROTO_PATH_DIRECTORIES
-        ${_opt_PROTO_PATH_DIRECTORIES})
+    if (${_opt_LOCAL_INCLUDE})
+        google_cloud_cpp_generate_proto(
+            proto_sources ${_opt_UNPARSED_ARGUMENTS} LOCAL_INCLUDE
+            PROTO_PATH_DIRECTORIES ${_opt_PROTO_PATH_DIRECTORIES})
+    else ()
+        google_cloud_cpp_generate_proto(
+            proto_sources ${_opt_UNPARSED_ARGUMENTS} PROTO_PATH_DIRECTORIES
+            ${_opt_PROTO_PATH_DIRECTORIES})
+    endif ()
 
     add_library(${libname} ${proto_sources})
     set_property(TARGET ${libname} PROPERTY PROTO_SOURCES
@@ -251,6 +277,11 @@ function (google_cloud_cpp_proto_library libname)
     target_include_directories(
         ${libname} SYSTEM PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
                                  $<INSTALL_INTERFACE:include>)
+
+    # In some configs we need to only generate the protocol definitions from
+    # `*.proto` files. We achieve this by having this target depend on all proto
+    # libraries. It has to be defined at the top level of the project.
+    add_dependencies(google-cloud-cpp-protos ${libname})
 endfunction ()
 
 function (google_cloud_cpp_grpcpp_library libname)
--- a/external/googleapis/CMakeLists.txt
+++ b/external/googleapis/CMakeLists.txt
@@ -161,10 +161,7 @@ function (googleapis_cpp_add_library proto)
     endif ()
 endfunction ()
 
-add_custom_target(google-cloud-cpp-protos)
-
 function (googleapis_cpp_set_version_and_alias short_name)
-    add_dependencies(google-cloud-cpp-protos "googleapis_cpp_${short_name}")
     add_dependencies("googleapis_cpp_${short_name}" googleapis_download)
     set_target_properties(
         "googleapis_cpp_${short_name}"
--- a/google/cloud/CMakeLists.txt
+++ b/google/cloud/CMakeLists.txt
@@ -177,10 +177,14 @@ target_link_libraries(
                                    Threads::Threads)
 google_cloud_cpp_add_common_options(google_cloud_cpp_common)
 target_include_directories(
-    google_cloud_cpp_common
-    PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
-           $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>
-           $<INSTALL_INTERFACE:include>)
+    google_cloud_cpp_common PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+                                   $<INSTALL_INTERFACE:include>)
+
+# We're putting generated code into ${PROJECT_BINARY_DIR} (e.g. compiled
+# protobufs or build info), so we need it on the include path, however we don't
+# want it checked by linters so we mark it as SYSTEM.
+target_include_directories(google_cloud_cpp_common SYSTEM
+                           PUBLIC $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>)
 target_compile_options(google_cloud_cpp_common
                        PUBLIC ${GOOGLE_CLOUD_CPP_EXCEPTIONS_FLAG})
 
@@ -378,7 +382,6 @@ if (GOOGLE_CLOUD_CPP_ENABLE_GRPC)
     target_include_directories(
         google_cloud_cpp_grpc_utils
         PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
-               $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>
                $<INSTALL_INTERFACE:include>)
     target_compile_options(google_cloud_cpp_grpc_utils
                            PUBLIC ${GOOGLE_CLOUD_CPP_EXCEPTIONS_FLAG})
--- a/google/cloud/storage/BUILD
+++ b/google/cloud/storage/BUILD
@@ -26,6 +26,18 @@ GOOGLE_CLOUD_STORAGE_WIN_COPTS = [
 load(":storage_client.bzl", "storage_client_hdrs", "storage_client_srcs")
 load(":storage_client_grpc.bzl", "storage_client_grpc_hdrs", "storage_client_grpc_srcs")
 
+proto_library(
+    name = "google_cloud_cpp_storage_internal_grpc_resumable_serssion_url_proto",
+    srcs = [
+        "internal/grpc_resumable_upload_session_url.proto",
+    ],
+)
+
+cc_proto_library(
+    name = "google_cloud_cpp_storage_internal_grpc_resumable_serssion_url_cc_proto",
+    deps = [":google_cloud_cpp_storage_internal_grpc_resumable_serssion_url_proto"],
+)
+
 cc_library(
     name = "storage_client_grpc",
     srcs = storage_client_grpc_srcs,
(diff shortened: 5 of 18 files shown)


coryan (changes requested):
> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189)* status: all files reviewed, 10 unresolved discussions (waiting on @dopiera)

___
*[google/cloud/storage/CMakeLists.txt, line 274 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ28JHf08keTBQboucc:-MJ28JHf08keTBQboucd:b-jiap2h) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/CMakeLists.txt#L274)):*
> ```CMake
> if (GOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC)
>     find_package(gRPC REQUIRED)
>     google_cloud_cpp_generate_proto(
> ```

Presumably you need similar changes to the `BUILD` file?
___
*[google/cloud/storage/CMakeLists.txt, line 276 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ28MrQ2KUqnGqjnOV3:-MJ28MrQ2KUqnGqjnOV4:b-fg54qy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/CMakeLists.txt#L276)):*
> ```CMake
>     google_cloud_cpp_generate_proto(
>         google_cloud_cpp_storage_proto_sources
>         grpc_resumable_upload_session_url.proto PROTO_PATH_DIRECTORIES
> ```

FYI: we should consider whether the `.proto` file needs to be installed, my first thought is that it does not.  WDYT?
___
*[google/cloud/storage/grpc_resumable_upload_session_url.proto, line 17 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2AjEF5QGUOlkON21S:-MJ2AjEF5QGUOlkON21T:b6ledsf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/grpc_resumable_upload_session_url.proto#L17)):*
> ```protobuf
> syntax = "proto3";
> 
> package google.cloud.storage;
> ```

Should we version this?
___
*[google/cloud/storage/internal/grpc_client.cc, line 335 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2A-KfCH3fUUcqPMo_:-MJ2A-KfCH3fUUcqPMoa:b-dx2sy4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_client.cc#L335)):*
> ```cpp
>     return resumable_upload_params.status();
>   }
>   auto upload_session_params = DecodeGrpcResumableUploadSessionUrl(upload_url);
> ```

Why do we need to decode twice?
___
*[google/cloud/storage/internal/grpc_resumable_upload_session_url.cc, line 26 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2ADNd7kZX6cg4odxr:-MJ2ADNd7kZX6cg4odxs:b-e4b0zn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_resumable_upload_session_url.cc#L26)):*
> ```cpp
> namespace internal {
> 
> char const* kURIScheme = "grpc://";
> ```

nit:

```cc
auto constexpr kURIScheme = "grpc://";
```
___
*[google/cloud/storage/internal/grpc_resumable_upload_session_url.cc, line 48 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2APuP0pfdxUmzf0Aq:-MJ2APuP0pfdxUmzf0Ar:bf13x7m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_resumable_upload_session_url.cc#L48)):*
> ```cpp
>         "configuration");
>   }
>   std::string const payload =
> ```

nit: `auto const payload =`?
___
*[google/cloud/storage/internal/grpc_resumable_upload_session_url.cc, line 50 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2AMGuAtx1VrTP1w90:-MJ2AMGuAtx1VrTP1w91:b-lydiek) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_resumable_upload_session_url.cc#L50)):*
> ```cpp
>   std::string const payload =
>       upload_session_url.substr(std::strlen(kURIScheme));
>   auto const decoded_vec = UrlsafeBase64Decode(payload);
> ```

Can this fail?
___
*[google/cloud/storage/internal/grpc_resumable_upload_session_url.cc, line 56 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ29_WxC1q-Th5Y9Qvm:-MJ29_WxC1q-Th5Y9Qvn:b85og8n) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_resumable_upload_session_url.cc#L56)):*
> ```cpp
>   if (!proto.ParseFromString(decoded)) {
>     return Status(StatusCode::kInvalidArgument,
>                   "Malformed gRPC resumable uplpad session URL");
> ```

nit: s/uplpad/upload/
___
*[google/cloud/storage/internal/grpc_resumable_upload_session_url.cc, line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2AWXp5zbouqAlfP12:-MJ2AWXp5zbouqAlfP13:bhp2dpi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_resumable_upload_session_url.cc#L63)):*
> ```cpp
> 
> bool IsGrpcResumableSessionUrl(std::string const& upload_session_url) {
>   return upload_session_url.find(kURIScheme) == 0;
> ```

Consider `.rfind(kURIScheme, 0) == 0`, or `absl::StartsWith`.
___
*[google/cloud/storage/internal/grpc_resumable_upload_session_url_test.cc, line 42 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ28uOVB1_imEbJiCAj:-MJ28uOVB1_imEbJiCAk:b-1htmgq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_resumable_upload_session_url_test.cc#L42)):*
> ```cpp
>   EXPECT_FALSE(IsGrpcResumableSessionUrl(""));
>   EXPECT_FALSE(IsGrpcResumableSessionUrl("grpc:/"));
>   EXPECT_FALSE(IsGrpcResumableSessionUrl("https://somrubbish"));
> ```

nit: s/somru/som**e**ru/ ?
___
*[google/cloud/storage/internal/hybrid_client.cc, line 153 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ291j89CeOErSmODNl:-MJ291j89CeOErSmODNm:br4b9r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/hybrid_client.cc#L153)):*
> ```cpp
>   if (internal::IsGrpcResumableSessionUrl(upload_id)) {
>     return grpc_->RestoreResumableSession(upload_id);
>   } else {
> ```

nit: Generally we write:

```cc
if (condition) {
  return ...;
}
return ...;
```

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189)* status: all files reviewed, 5 unresolved discussions (waiting on @coryan)

___
*[google/cloud/storage/CMakeLists.txt, line 274 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ28JHf08keTBQboucc:-MJZX9-53NOyOk0xv0jH:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/CMakeLists.txt#L274)):*

Presumably you need similar changes to the `BUILD` file?
</blockquote></details>

Done.
___
*[google/cloud/storage/CMakeLists.txt, line 276 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ28MrQ2KUqnGqjnOV3:-MJZXAL34gVuA2vONqdQ:by88xv7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/CMakeLists.txt#L276)):*

FYI: we should consider whether the `.proto` file needs to be installed, my first thought is that it does not.  WDYT?
</blockquote></details>

This is only an internal detail, not part of a public API.
___
*[google/cloud/storage/grpc_resumable_upload_session_url.proto, line 17 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2AjEF5QGUOlkON21S:-MJZXPUCBCpMiWTzddMQ:bi1xpt3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/grpc_resumable_upload_session_url.proto#L17)):*

Should we version this?
</blockquote></details>

Initially I thought "no", but we have to if we want to use multiple versions simulateneously and not run into name conflicts. Done.
___
*[google/cloud/storage/internal/grpc_client.cc, line 335 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2A-KfCH3fUUcqPMo_:-MJZYBXe4uOvDOF6-fk3:b-4hoj94) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_client.cc#L335)):*

Why do we need to decode twice?
</blockquote></details>

We don't that's an oversight.
___
*[google/cloud/storage/internal/grpc_resumable_upload_session_url.cc, line 50 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2AMGuAtx1VrTP1w90:-MJZYQ6T9l1ijq9NWk0h:bg33jy1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_resumable_upload_session_url.cc#L50)):*

Can this fail?
</blockquote></details>

It can. Unfortunately, the underlying decoder only throws.
___
*[google/cloud/storage/internal/grpc_resumable_upload_session_url.cc, line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJ2AWXp5zbouqAlfP12:-MJZZ4DJ-P5YO02GC9L8:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0b92bfe615f327bfbb439ef5232c469bef688d58/google/cloud/storage/internal/grpc_resumable_upload_session_url.cc#L63)):*

Consider `.rfind(kURIScheme, 0) == 0`, or `absl::StartsWith`.
</blockquote></details>

Done.

coryan (changes requested):
Looking really good, some improvements (around comments and CMake nits) below.

> Reviewed 10 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-:-MJZhpkE7gdOx1Z_9hUS:buvvnj2)* status: all files reviewed, 4 unresolved discussions (waiting on @dopiera)

___
*[cmake/CompileProtos.cmake, line 59 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJZg7Tt0s5xNesLt6nx:-MJZg7Tt0s5xNesLt6ny:b-x94xnc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f28bc372c4f6cb7001b8d0b92355c1508be0e41/cmake/CompileProtos.cmake#L59)):*
> ```CMake
> # path.
> function (google_cloud_cpp_generate_proto SRCS)
>     cmake_parse_arguments(_opt "" "LOCAL_INCLUDE" "PROTO_PATH_DIRECTORIES"
> ```

You can define arguments that are purely boolean in that first quote:

```cmake
cmake_parse_arguments(_opt "LOCAL_INCLUDE "" "PROTO_PATH_DIRECTORIES" ${ARGN})
```

will define `_opt_LOCAL_INCLUDE` as TRUE if the argument is provided, and false otherwise.  That would make the usage more natural down below...
___
*[google/cloud/CMakeLists.txt, line 182 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJZfm8a4vi9JejCHMJy:-MJZfm8a4vi9JejCHMJz:b-rjusgq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f28bc372c4f6cb7001b8d0b92355c1508be0e41/google/cloud/CMakeLists.txt#L182)):*
> ```CMake
>     google_cloud_cpp_common PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
>                                    $<INSTALL_INTERFACE:include>)
> target_include_directories(google_cloud_cpp_common SYSTEM
> ```

Can you add a comment explaining why this is needed?  Seems non-obvious.  I presume is because we need to ignore warnings in the generated `.pb.h` files? That is even more obscure because there are no protos for the libraries defined here.
___
*[google/cloud/storage/CMakeLists.txt, line 278 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJZglwaEd3OcmpMvIAU:-MJZglwb32tW2NNCx1tK:bi66k38) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f28bc372c4f6cb7001b8d0b92355c1508be0e41/google/cloud/storage/CMakeLists.txt#L278)):*
> ```CMake
>         google_cloud_cpp_storage_protos
>         internal/grpc_resumable_upload_session_url.proto PROTO_PATH_DIRECTORIES
>         ${CMAKE_CURRENT_SOURCE_DIR}/internal LOCAL_INCLUDE TRUE)
> ```

See above, this could become simply `LOCAL_INCLUDE` instead of `LOCAL_INCLUDE TRUE`.
___
*[google/cloud/storage/CMakeLists.txt, line 361 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJZh4DmCMY9wwKNKWLK:-MJZh4DmCMY9wwKNKWLL:be3p4ak) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f28bc372c4f6cb7001b8d0b92355c1508be0e41/google/cloud/storage/CMakeLists.txt#L361)):*
> ```CMake
>                GTest::gtest)
>     google_cloud_cpp_add_common_options(storage_client_testing)
>     target_include_directories(storage_client_testing SYSTEM
> ```

I think this is adding `$<INSTALL_INTERFACE:include>` with both `-isystem` and `-iquote`, is that what we want?  I would expect only one is needed?

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189)* status: all files reviewed, 4 unresolved discussions (waiting on @coryan)

___
*[cmake/CompileProtos.cmake, line 59 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJZg7Tt0s5xNesLt6nx:-MK6JY522joqJ5DDL2M6:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f28bc372c4f6cb7001b8d0b92355c1508be0e41/cmake/CompileProtos.cmake#L59)):*

You can define arguments that are purely boolean in that first quote:

```cmake
cmake_parse_arguments(_opt "LOCAL_INCLUDE "" "PROTO_PATH_DIRECTORIES" ${ARGN})
```

will define `_opt_LOCAL_INCLUDE` as TRUE if the argument is provided, and false otherwise.  That would make the usage more natural down below...


</blockquote></details>

Done.
___
*[google/cloud/CMakeLists.txt, line 182 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJZfm8a4vi9JejCHMJy:-MK6KgwWACcBdbnOjI1Q:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f28bc372c4f6cb7001b8d0b92355c1508be0e41/google/cloud/CMakeLists.txt#L182)):*

Can you add a comment explaining why this is needed?  Seems non-obvious.  I presume is because we need to ignore warnings in the generated `.pb.h` files? That is even more obscure because there are no protos for the libraries defined here.
</blockquote></details>

Done.
___
*[google/cloud/storage/CMakeLists.txt, line 278 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJZglwaEd3OcmpMvIAU:-MJbTDTe8NI3XAbQvmWt:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f28bc372c4f6cb7001b8d0b92355c1508be0e41/google/cloud/storage/CMakeLists.txt#L278)):*

See above, this could become simply `LOCAL_INCLUDE` instead of `LOCAL_INCLUDE TRUE`.
</blockquote></details>

Done.
___
*[google/cloud/storage/CMakeLists.txt, line 361 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MJZh4DmCMY9wwKNKWLK:-MJbTdYTAzcVPw_CW-hC:b-o84mvz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f28bc372c4f6cb7001b8d0b92355c1508be0e41/google/cloud/storage/CMakeLists.txt#L361)):*

I think this is adding `$<INSTALL_INTERFACE:include>` with both `-isystem` and `-iquote`, is that what we want?  I would expect only one is needed?
</blockquote></details>

Yes, that's an oversight.

coryan (commented):
One last nit, but maybe there are good reasons why you implemented things the way you did.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-:-MK6QaqY3Qg23S6ltpXo:b95ysnd)* status: all files reviewed, 2 unresolved discussions (waiting on @dopiera)

___
*[cmake/CompileProtos.cmake, line 280 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MK6Powl85we6nvXXhgR:-MK6Powl85we6nvXXhgS:brxr1rk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9abd3394ae2593d9f1241e1043b32f351ba982b2/cmake/CompileProtos.cmake#L280)):*
> ```CMake
>         ${libname} SYSTEM PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
>                                  $<INSTALL_INTERFACE:include>)
>     add_dependencies(google-cloud-cpp-protos ${libname})
> ```

nit: it is a bit odd to refer to a target here that may or may not have been defined...  
___
*[external/googleapis/CMakeLists.txt, line 165 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MK6PvRI3cXext69KIyD:-MK6PvRI3cXext69KIyE:bo2vnz7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9abd3394ae2593d9f1241e1043b32f351ba982b2/external/googleapis/CMakeLists.txt#L165)):*
> ```CMake
> 
> function (googleapis_cpp_set_version_and_alias short_name)
>     add_dependencies(google-cloud-cpp-protos "googleapis_cpp_${short_name}")
> ```

nit: this is now redundant?  Maybe we should consider keeping this line, removing the redundant line in `CompileProtos.cmake` and adding the right dependency on `google_cloud_cpp_storage_protos` ?

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189)* status: 16 of 18 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[cmake/CompileProtos.cmake, line 280 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MK6Powl85we6nvXXhgR:-MK6W5DG5VBBCNhyWhjw:b-kwfwug) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9abd3394ae2593d9f1241e1043b32f351ba982b2/cmake/CompileProtos.cmake#L280)):*

nit: it is a bit odd to refer to a target here that may or may not have been defined...  
</blockquote></details>

I think it's better to have it defined in `CompileProtos.cmake`. My rationale is:
* let's make it in one place so that we don't need to think about it (the obvious reason)
* when it breaks for whatever reason, it should be pretty simple to figure out why because you'll get see cmake complaining about this line (if not your pointer during the meeting, it would have taken me quite a while to find out about this target at all)

I acknowledge that this is really ugly, to rely on this target being defined, but I think that it's worth trading for better debuggability in the future. WDYT?
___
*[external/googleapis/CMakeLists.txt, line 165 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-MK6PvRI3cXext69KIyD:-MK6W61ICwnPOnYtK4QU:b-enzyci) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9abd3394ae2593d9f1241e1043b32f351ba982b2/external/googleapis/CMakeLists.txt#L165)):*

nit: this is now redundant?  Maybe we should consider keeping this line, removing the redundant line in `CompileProtos.cmake` and adding the right dependency on `google_cloud_cpp_storage_protos` ?
</blockquote></details>

Yup, this is redundant. PTAL at my previous answer also.

dopiera:
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189)_ status: 16 of 18 files reviewed, 2 unresolved discussions (waiting on @coryan)
> I acknowledge that this is really ugly, to rely on this target being defined, but I think that it's worth trading for better debuggability in the future. WDYT?


Alternatively, we may make it conditional on the target's existence, but I think the whole debuggability point is lost then.

coryan (approved):
:lgtm:

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5189#-:-MK6aVWNESe-HGeenmaN:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #4669, opened 2020-07-23
Implement pubsub_publish_ordered_messages sample


coryan:
https://github.com/googleapis/nodejs-pubsub/blob/dcdec71e3a496b041dc8d1d0f03eda85d5d0cf63/samples/publishOrderedMessage.js#L46-L81

coryan:
This sample (`pubsub_publish_ordered_messages`) does not exist, it seems to have been a typo that I picked up when collecting samples from other languages. Recent versions of nodejs call it `pubsub_publish_with_ordering_keys`, like all other languages:

https://github.com/googleapis/nodejs-pubsub/blob/85fec09bfd956adbeeb8a819af796a3cedb3befe/samples/publishOrderedMessage.js#L47-L89

Furthermore, we have this in C++ already:

https://github.com/googleapis/google-cloud-cpp/blob/4b12aea940d4abb1d502fb7cd41419d3c7d1feac/google/cloud/pubsub/samples/samples.cc#L722-L753

Closing as duplicate.

Issue #4656, opened 2020-07-23
Implement pubsub_listen_ordered_messages sample


coryan:
Subscribe to some messages and manually sort them.

https://github.com/googleapis/nodejs-pubsub/blob/dcdec71e3a496b041dc8d1d0f03eda85d5d0cf63/samples/listenForOrderedMessages.js#L49-L114

remyabel:
The region tags for `publishOrderedDate` was [updated](https://github.com/googleapis/nodejs-pubsub/commit/b501325265056319167fb8d5f81e1a0c3582af09) 23 days ago but `listenForOrderedMessages` hasn't been updated in 6 months. I modelled my PR after that sample (including the logic to use a timeout to wait for messages and then sort them by a counterId) so if that is also outdated, I'll wait for updated information how to proceed.

coryan:
I think we are not going to implement any specific examples for receiving ordered messages.  The same code as receiving regular messages "Just Works":tm:

Issue #5044, opened 2020-09-11
Automatically schedule messages with ordering keys one at a time


coryan:
If we receive a message with an ordering key then the topic was configured for ordered delivery: topics configured otherwise reject messages with ordering keys.

The library should automatically schedule such messages "one at a time". We should wait until a message is fully handled before we schedule the next message with the same ordering key.  Messages with different ordering keys can be scheduled in any order.

PR #5194, opened 2020-10-07
feat(pubsub): Implement ordered messages samples

remyabel:
Fixes: #4669
Fixes: #4656

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5194)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -829,8 +829,6 @@ void ListenOrderedMessages(
     // Report any final status, blocking.
     std::cout << "Message count: " << message_count << ", status: " << status
               << "\n";
-
-    // auto constexpr kExpectedMessageCount = 10;
     for (auto const& m : outstanding_messages) {
       std::cout << "Received message " << m.first << "\n";
     }
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -754,6 +754,91 @@ void PublishOrderingKey(google::cloud::pubsub::Publisher publisher,
   (std::move(publisher));
 }
 
+void PublishOrderedMessage(google::cloud::pubsub::Publisher publisher,
+                           std::vector<std::string> const&) {
+  //! [START pubsub_publish_ordered_message] [publish-ordered-message]
+  namespace pubsub = google::cloud::pubsub;
+  using google::cloud::future;
+  using google::cloud::StatusOr;
+  [](pubsub::Publisher publisher) {
+    std::vector<future<void>> done;
+    for (int i = 0; i != 10; ++i) {
+      auto message_id = publisher.Publish(
+          pubsub::MessageBuilder{}
+              .SetData("Hello World! [" + std::to_string(i) + "]")
+              // Have the counter-id start from 1
+              .SetAttributes({{"counter-id", std::to_string(i + 1)}})
+              .Build());
+      done.push_back(message_id.then([i](future<StatusOr<std::string>> f) {
+        auto id = f.get();
+        if (!id) throw std::runtime_error(id.status().message());
+        std::cout << "Message " << i << " published with id=" << *id << "\n"
+                  << " and counter-id=" << (i + 1) << "\n";
+      }));
+    }
+    publisher.Flush();
+    // Block until all the messages are published (optional)
+    for (auto& f : done) f.get();
+  }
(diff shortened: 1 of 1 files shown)


coryan (changes requested):
Thanks for sending this PR.  #4669 was out of date and a duplicate. We can remove that code.  Maybe we should IM about the intent if the other example because I am a bit confused.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5194#-:-MJ48bdS7x_cakaLS0CX:b-40la9p)* status: all files reviewed, 5 unresolved discussions (waiting on @remyabel)

___
*[google/cloud/pubsub/samples/samples.cc, line 757 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5194#-MJ48n81DzKSbZbOJfkg:-MJ48n81DzKSbZbOJfkh:bkj8xrj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b8a525927d823d6a5e4f6b0324c952371f243f03/google/cloud/pubsub/samples/samples.cc#L757)):*
> ```cpp
> }
> 
> void PublishOrderedMessage(google::cloud::pubsub::Publisher publisher,
> ```

See above, we do not need this function.
___
*[google/cloud/pubsub/samples/samples.cc, line 799 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5194#-MJ4AQjQCSUiO4QIapRW:-MJ4AQjQCSUiO4QIapRX:b37vw0u) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b8a525927d823d6a5e4f6b0324c952371f243f03/google/cloud/pubsub/samples/samples.cc#L799)):*
> ```cpp
>   using google::cloud::future;
>   using google::cloud::StatusOr;
>   using ordered_message = std::pair<pubsub::Message, int>;
> ```

nit: generally we `CamelCase` for our own types (namespaces are `snake_case`, functions have complicated rules, types borrowed from the standard match the standard).
___
*[google/cloud/pubsub/samples/samples.cc, line 801 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5194#-MJ4Bdt27xzTlxuoeoM7:-MJ4Bdt27xzTlxuoeoM8:b-tyzucb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b8a525927d823d6a5e4f6b0324c952371f243f03/google/cloud/pubsub/samples/samples.cc#L801)):*
> ```cpp
>   using ordered_message = std::pair<pubsub::Message, int>;
>   [](pubsub::Subscriber subscriber, pubsub::Subscription const& subscription,
>      int timeout) {
> ```

consider using `std::chrono::milliseconds timeout` here, that way the semantics are obvious from the type.
___
*[google/cloud/pubsub/samples/samples.cc, line 816 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5194#-MJ4Aia86c2ZmWNLNLu0:-MJ4Aia86c2ZmWNLNLu1:b2p09ql) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b8a525927d823d6a5e4f6b0324c952371f243f03/google/cloud/pubsub/samples/samples.cc#L816)):*
> ```cpp
>           std::move(h).ack();
>         });
>     std::sort(outstanding_messages.begin(), outstanding_messages.end(),
> ```

I am not sure we need to sort the messages... Can you tell me what is the intent?  If all is working well, the messages should arrive in the same order they were sent.  Currently that is not the case, if nothing else because #5044 
___
*[google/cloud/pubsub/samples/samples.cc, line 822 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5194#-MJ4C-NPACxRNcnCm5Q9:-MJ4C-NPACxRNcnCm5QA:b-eq5lm6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b8a525927d823d6a5e4f6b0324c952371f243f03/google/cloud/pubsub/samples/samples.cc#L822)):*
> ```cpp
>     // Wait until the timeout has been reached.
>     std::unique_lock<std::mutex> lk(mu);
>     cv.wait_for(lk, std::chrono::milliseconds(timeout));
> ```

I think you can say `cv.wait(lk, timeout, [&] { return message_count >= N; });` to stop waiting as soon as N messages have arrived.  If you really want to wait for N milliseconds, maybe just `std::this_thread::sleep_for()`?

remyabel:
Let's close this for now and continue discussion in https://github.com/googleapis/google-cloud-cpp/issues/4656.

PR #5203, opened 2020-10-08
test(storage): store the full Status in ThroughputResult

mr-salty:
This may give us more information if googleapis#5202 recurs.

coryan: one thing that changed is the PrintAsCsv() format, is that going
to mess up any of your analysis? Should I change it?

INSERT,3145728,2097152,4194304,1,0,GRPC,234000,345000,OOR-status-message [OUT_OF_RANGE]

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5203)

### DIFF
--- a/google/cloud/storage/benchmarks/throughput_result.cc
+++ b/google/cloud/storage/benchmarks/throughput_result.cc
@@ -13,16 +13,36 @@
 // limitations under the License.
 
 #include "google/cloud/storage/benchmarks/throughput_result.h"
+#include "absl/strings/str_cat.h"
+#include "absl/strings/str_replace.h"
+#include <sstream>
+#include <string>
 
 namespace google {
 namespace cloud {
 namespace storage_benchmarks {
 
+namespace {
+template <typename T>
+std::string QuoteCsv(T const& element) {
+  std::ostringstream os;
+  os << element;
+  std::string result = os.str();
+  if (result.find_first_of(",\r\n") == std::string::npos) {
+    return result;
+  }
+
+  // Enclose the string in double quotes, and escape other double quotes.
+  return absl::StrCat(R"(")", absl::StrReplaceAll(result, {{R"(")", R"("")"}}),
+                      R"(")");
+}
+}  // namespace
+
 void PrintAsCsv(std::ostream& os, ThroughputResult const& r) {
   os << ToString(r.op) << ',' << r.object_size << ',' << r.app_buffer_size
      << ',' << r.lib_buffer_size << ',' << r.crc_enabled << ',' << r.md5_enabled
      << ',' << ToString(r.api) << ',' << r.elapsed_time.count() << ','
-     << r.cpu_time.count() << ',' << r.status << '\n';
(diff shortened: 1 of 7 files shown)


coryan:
> coryan: 

nit: `@coryan` is the way to mention folks on GitHub.

> one thing that changed is the PrintAsCsv() format, is that going to mess up any of your analysis? Should I change it?

I think error messages could contain `,` and `\n` characters (I have actually seen the latter), and that would certainly mess by analysis.  Can we convert those to `;` or some character like it?

coryan (commented):
Maybe change the `ostream<<` operator to replace newlines and commas by some other character?

devbww (commented):
For what it's worth, these sound like things that a `PrintAsCsv()` function should handle (e.g., by escaping `,` and '\n' chars).

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5203#-:-MJ8LDnJEE3lhi_GCdXq:b-8k2080)* status: :shipit: complete! all files reviewed, all discussions resolved

coryan (approved):

mr-salty (commented):
PTAL - I quoted the string (AFAICT `status` is the only field affected).

Note that [rfc 4180](https://tools.ietf.org/html/rfc4180#section-2) does *not* say you have to quote if the string contains spaces - just comma, double-quote, or CRLF (2.6). Although, I chose to escape CR|LF, and could be convinced to do it for spaces as well.  It's always *permissible* to quote the string. per 2.5.

The character set is also supposed to be limited, although I didn't do anything to strip out other characters:    `TEXTDATA =  %x20-21 / %x23-2B / %x2D-7E`

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5203#-:-MJ92EXd54BiigbZzYYg:b4dx7bn)* status: 4 of 6 files reviewed, all discussions resolved (waiting on @devbww)

google/cloud/storage/benchmarks/throughput_result.cc
```diff
@@ -13,16 +13,36 @@
...
 namespace cloud {
 namespace storage_benchmarks {
 
+namespace {
+template <typename T>
+std::string QuoteCsv(T const& element) {
+  std::ostringstream os;
+  os << element;
+  std::string result = os.str();
+  if (result.find_first_of(",\r\n") == std::string::npos) {
```
devbww, line 31: Double quote?

mr-salty:
___
*[google/cloud/storage/benchmarks/throughput_result.cc, line 31 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5203#-MJ91xXsazuyvwcWOzUF-r2-31:-MJ9gBSCAGbFfP_o4mK_:b7z55tb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/00504971e3a62012972de7db80406e2eb21613bc/google/cloud/storage/benchmarks/throughput_result.cc#L31)):*

Double quote?
</blockquote></details>

good catch, I knew I should have done tests for each case separately (and now I have).

devbww (approved):

PR #5210, opened 2020-10-08
fix(pubsub): streaming pull race conditions

coryan:
Found some race conditions when using this against production.
Apparently the timing is different enough from the emulator that these
do not show up in the emulator even when running thousands of times.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5210)

### DIFF
--- a/google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
+++ b/google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
@@ -531,7 +531,7 @@ TEST(StreamingSubscriptionBatchSourceTest, ShutdownWithPendingRead) {
   AutomaticallyCreatedBackgroundThreads background;
   auto mock = std::make_shared<pubsub_testing::MockSubscriberStub>();
 
-  auto const expected_status = Status{StatusCode::kOk, "fine"};
+  auto const expected_status = Status{};
   FakeStream fake_stream(true, true, expected_status);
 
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
```diff
@@ -493,22 +516,55 @@ TEST(StreamingSubscriptionBatchSourceTest, WriteErrorWaitsForRead) {
...
+  EXPECT_EQ(expected_status, done.get());
+}
+
+TEST(StreamingSubscriptionBatchSourceTest, ShutdownWithPendingRead) {
+  auto subscription = pubsub::Subscription("test-project", "test-subscription");
+  std::string const client_id = "fake-client-id";
+  AutomaticallyCreatedBackgroundThreads background;
+  auto mock = std::make_shared<pubsub_testing::MockSubscriberStub>();
+
+  auto const expected_status = Status{StatusCode::kOk, "fine"};
```
devbww, line 534: Q. Should we ever be constructing an OK `Status` that contains a message?

devbww (approved):

google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
coryan: No, fixed.

PR #5213, opened 2020-10-09
fix: remove test from library sources list

coryan:
This was accidentally (I think) added in #5203.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5213)

PR #5219, opened 2020-10-09
ci: check for diffs post-build

coryan:
This can help us detect any changes introduced by the build scripts,
such as modifications to the `*.bzl` files that happen after we invoke
`ci/check-style.sh`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5219)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -430,6 +430,14 @@ if [[ "${GENERATE_DOCS}" == "yes" ]]; then
   cmake --build "${BINARY_DIR}" --target doxygen-docs -- -j "${NCPU}"
 fi
 
+# Report any differences created by the build, some steps may modify the code
+# *after* the style-checking tools run (e.g. the `*.bzl` file generators do).
+if [[ "${CHECK_STYLE:-}" == "yes" ]]; then
+  echo "================================================================"
+  io::log_yellow "checking for post-build changes is the code"
+  git diff --ignore-submodules=all --color --exit-code .
+fi
+
 if command -v ccache; then
   echo "================================================================"
   io::log_yellow "ccache stats"
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -432,7 +432,7 @@ fi
 
 # Report any differences created by the build, some steps may modify the code
 # *after* the style-checking tools run (e.g. the `*.bzl` file generators do).
-if [[ "${CHECK_STYLE:-}" == "yes" ]]; then
+if [[ "${CHECK_STYLE:-}" == "yes" && "${RUNNING_CI}" == "yes" ]]; then
   echo "================================================================"
   io::log_yellow "checking for post-build changes in the code"
   git diff --ignore-submodules=all --color --exit-code .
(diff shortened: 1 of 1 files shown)


ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -430,6 +430,14 @@ if [[ "${GENERATE_DOCS}" == "yes" ]]; then
   cmake --build "${BINARY_DIR}" --target doxygen-docs -- -j "${NCPU}"
 fi
 
+# Report any differences created by the build, some steps may modify the code
+# *after* the style-checking tools run (e.g. the `*.bzl` file generators do).
+if [[ "${CHECK_STYLE:-}" == "yes" ]]; then
+  echo "================================================================"
+  io::log_yellow "checking for post-build changes is the code"
```
devbww, line 437: "changes is the code"?  s/is/in/ perhaps?

coryan: Doh! Fixed.

ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -430,6 +430,14 @@ if [[ "${GENERATE_DOCS}" == "yes" ]]; then
   cmake --build "${BINARY_DIR}" --target doxygen-docs -- -j "${NCPU}"
 fi
 
+# Report any differences created by the build, some steps may modify the code
+# *after* the style-checking tools run (e.g. the `*.bzl` file generators do).
+if [[ "${CHECK_STYLE:-}" == "yes" ]]; then
+  echo "================================================================"
+  io::log_yellow "checking for post-build changes in the code"
+  git diff --ignore-submodules=all --color --exit-code .
```
devjgm, line 438: I often run builds (including the clang-tidy build) with uncommitted changes. Will this fail those builds?

coryan: Yes. But *after* all the other build steps: the code would have been compiled, the tests executed and then this will exit the script with a non-zero code because there are diffs.

You can `git add` the pending changes and the script to return 0.  That is extra nifty because it breaks if the formatting changes anything.

devjgm: This will mess up my workflow. Maybe I'm the only one. If so, I can live with it. Is this solving a problem that we had? I saw the PR description about the generated `.bzl` files, but diffs in those files will usually result in a bazel build breaking if it's missing files.

coryan: > This will mess up my workflow. Maybe I'm the only one. If so, I can live with it. 

Okay, disabled for non-interactive builds.

>Is this solving a problem that we had? 

 See #5213 for an example of an actual inconsistency between the CMake lists and the Bazel lists.

> I saw the PR description about the generated `.bzl` files, but diffs in those files will usually result in a bazel build breaking if it's missing files.

Unless your `.bzl` file has fewer tests than it should have, in which case it builds happily, when it should not.

coryan: Oh, and in the case of #5213: automatically updating the `.bzl` file broke the build because the CMake file was broken (it built for CMake, but would not build on Bazel).

devjgm: OK. Thanks for clarifying the background on this. I'm happy w/ the code as it is now (only running on CI builds) if you are.

devjgm (approved):

PR #5225, opened 2020-10-12
feat(pubsub): an endurance benchmark

### LINKED ISSUE #4564: Implement an "endurance" test
Write a test that exercises the library and runs for hours. This is to trigger rare leaks, race conditions, or other problems that go undetected by short tests.

The test may need a separate design doc, but it should:

- Run with multiple threads
- Randomly publish messages
- Receive messages and randomly ack or nack them

As part of the CI build we would wish to run a very short version of this test, just to ensure the code compiles and does not crash in an embarrassing way. Running the test for 24 hours or so would be a manual step.

coryan:
Verify the Pub/Sub client library works well when used for hours.

Fixes #4564

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5225)


google/cloud/pubsub/benchmarks/endurance.cc
```diff
@@ -0,0 +1,634 @@
...
+        flow_control.Debug(std::cout);
+        std::cout << std::endl;
+      } else {
+        auto const p = std::minmax_element(samples.begin(), samples.end());
+        auto const sum = std::accumulate(samples.begin(), samples.end(),
+                                         std::chrono::microseconds{0});
+        auto const mean = [](std::chrono::microseconds sum, std::size_t n) {
+          sum /= n;
+          return sum;
+        }(sum, samples.size());
```
devbww, line 315: Why not simply `auto const mean = sum / samples.size()`?

google/cloud/pubsub/benchmarks/endurance.cc
```diff
@@ -0,0 +1,634 @@
...
+  pubsub::TopicAdminClient topic_admin(pubsub::MakeTopicAdminConnection());
+  pubsub::SubscriptionAdminClient subscription_admin(
+      pubsub::MakeSubscriptionAdminConnection());
+
+  auto generator = google::cloud::internal::MakeDefaultPRNG();
+
+  auto const configured_topic = config->topic_id;
+
+  // If there is no pre-defined topic for this test, create one and
+  // automatically remove it at the end of the test.
```
devbww, line 179: It looks like we delete the topic unconditionally.  Should be only do that if we created one?

google/cloud/pubsub/benchmarks/endurance.cc
coryan: LOL, yes of course.  It used to be more complicated and I failed to simplify it fully.

google/cloud/pubsub/benchmarks/endurance.cc
coryan: Good catch, fixed.

devbww (approved):

Issue #5204, opened 2020-10-08
Change Subscriber to receive all parameters in the constructor


coryan:
For consistency with other languages the class will be setup and the `Start()` function will have no arguments.

Makes me wonder if this should be a standalone function.

kamalaboulhosn:
I think ultimately, having a stateful Subscriber class will be useful. There might be some methods coming in the future that would make it easier to have one, e.g., there have been discussions about adding a method to stop tracking a message in flow control and then allowing one to call Ack using the message's AckId. I guess this could be a standalone function as well, though. I do like the symmetry of a Publisher and a Subscriber class.

coryan:
Ack. After #5247 the API became:

```cc
void Handler(pubsub::Message m, pubsub::AckHandler h);

void F(pubsub::Subscription s) {
  pubsub::Subscriber subscriber(pubsub::MakeSubscriberConnection(s));
  auto future = subscriber.Subscribe(Handler);
  future.wait_for(std::chrono::hours(1));
}
```

coryan:
I am assuming it is, closing.

PR #5247, opened 2020-10-14
cleanup(pubsub)!: make Subscriber more consistent

coryan:
**BREAKING CHANGE**

* Change the `pubsub::Subscriber` API. A `Subscriber` is now bound to a specific
  Cloud Pub/Sub subscription, with a fixed set of `SuscriptionOptions`, just
  like a `pubsub::Publisher` is bound to a specific topic and a set of
  `PublisherOptions`. In addition to making publishers and subscribers more
  symmetrical, this makes the library more consistent with the Cloud Pub/Sub
  library for other languages. Finally, note that we are planning to rename
  `SubscriptionOptions` to `SubscriberOptions` in a future PR too.

Part of the work for #5204 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5247)

### DIFF
--- a/google/cloud/pubsub/samples/pubsub_samples_common.cc
+++ b/google/cloud/pubsub/samples/pubsub_samples_common.cc
@@ -59,9 +59,9 @@ google::cloud::testing_util::Commands::value_type CreateSubscriberCommand(
       }
       throw google::cloud::testing_util::Usage{std::move(os).str()};
     }
-    pubsub::Subscription subscription(argv.at(0), argv.at(1));
     google::cloud::pubsub::Subscriber client(
-        google::cloud::pubsub::MakeSubscriberConnection(subscription));
+        google::cloud::pubsub::MakeSubscriberConnection(
+            pubsub::Subscription(argv.at(0), argv.at(1))));
     argv.erase(argv.begin(), argv.begin() + 2);
     command(std::move(client), std::move(argv));
   };
--- a/google/cloud/pubsub/samples/pubsub_samples_common.cc
+++ b/google/cloud/pubsub/samples/pubsub_samples_common.cc
@@ -63,7 +63,7 @@ google::cloud::testing_util::Commands::value_type CreateSubscriberCommand(
     google::cloud::pubsub::Subscriber client(
         google::cloud::pubsub::MakeSubscriberConnection(subscription));
     argv.erase(argv.begin(), argv.begin() + 2);
-    command(std::move(client), std::move(subscription), std::move(argv));
+    command(std::move(client), std::move(argv));
   };
   return google::cloud::testing_util::Commands::value_type{name,
                                                            std::move(adapter)};
--- a/google/cloud/pubsub/samples/pubsub_samples_common.cc
+++ b/google/cloud/pubsub/samples/pubsub_samples_common.cc
@@ -59,9 +59,9 @@ google::cloud::testing_util::Commands::value_type CreateSubscriberCommand(
       }
       throw google::cloud::testing_util::Usage{std::move(os).str()};
     }
-    google::cloud::pubsub::Subscriber client(
-        google::cloud::pubsub::MakeSubscriberConnection());
     pubsub::Subscription subscription(argv.at(0), argv.at(1));
+    google::cloud::pubsub::Subscriber client(
+        google::cloud::pubsub::MakeSubscriberConnection(subscription));
     argv.erase(argv.begin(), argv.begin() + 2);
     command(std::move(client), std::move(subscription), std::move(argv));
   };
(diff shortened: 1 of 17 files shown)


google/cloud/pubsub/subscriber.h
```diff
@@ -116,11 +114,9 @@ class Subscriber {
    * https://en.cppreference.com/w/cpp/utility/functional/function
    */
   template <typename Callable>
-  future<Status> Subscribe(Subscription const& subscription, Callable&& cb,
-                           SubscriptionOptions options = {}) {
+  future<Status> Subscribe(Callable&& cb) {
     std::function<void(Message, AckHandler)> f(std::forward<Callable>(cb));
-    return connection_->Subscribe(
-        {subscription.FullName(), std::move(f), std::move(options)});
+    return connection_->Subscribe({std::move(f)});
```
kamalaboulhosn, line 119: Is the additional SubscriberConnection abstraction that useful if Subscriber is just going to delegate the calls to it? I wonder what benefit we are gaining by having these two classes be separate. We don't have to make the change in this PR, but maybe it's worth us considering as a future change.

coryan: In the C++ libraries all `*Client` classes (which `Subscriber` models) are value types.  But we need to support mocking, so the `*Connection` classes are pure abstract.

google/cloud/pubsub/samples/samples.cc
```diff
@@ -341,18 +341,18 @@ void UpdateDeadLetterSubscription(
 
 void ReceiveDeadLetterDeliveryAttempt(
     google::cloud::pubsub::Subscriber subscriber,
-    google::cloud::pubsub::Subscription const& subscription,
+    google::cloud::pubsub::Subscription const&,
```
devbww, line 344: Can this now-unused parameter be removed from here (and also `Subscribe()`, `SubscribeErrorListener()`, and `SubscribeCustomAttributes()`) along with the suggested change to `examples::SubscriberCommand`?

google/cloud/pubsub/samples/pubsub_samples_common.cc
```diff
@@ -59,9 +59,9 @@ google::cloud::testing_util::Commands::value_type CreateSubscriberCommand(
       }
       throw google::cloud::testing_util::Usage{std::move(os).str()};
     }
-    google::cloud::pubsub::Subscriber client(
-        google::cloud::pubsub::MakeSubscriberConnection());
     pubsub::Subscription subscription(argv.at(0), argv.at(1));
+    google::cloud::pubsub::Subscriber client(
+        google::cloud::pubsub::MakeSubscriberConnection(subscription));
     argv.erase(argv.begin(), argv.begin() + 2);
     command(std::move(client), std::move(subscription), std::move(argv));
```
devbww, line 66: Remove `subscription` parameter?

google/cloud/pubsub/samples/samples.cc
coryan: Good catch, I though about it as I was making these changes and then promptly forgot.

google/cloud/pubsub/samples/pubsub_samples_common.cc
coryan: Done.

google/cloud/pubsub/samples/pubsub_samples_common.cc
```diff
@@ -63,7 +63,7 @@ google::cloud::testing_util::Commands::value_type CreateSubscriberCommand(
     google::cloud::pubsub::Subscriber client(
         google::cloud::pubsub::MakeSubscriberConnection(subscription));
```
devbww, line 64: It is worth `move`ing the `subscription` now (or not giving it a name)?

devbww (approved):

google/cloud/pubsub/samples/pubsub_samples_common.cc
coryan: Doh, fixed.

google/cloud/pubsub/subscriber.h
kamalaboulhosn: Is there a reason it is done this way as opposed to making a mockable `*Client`? It seems like an extra level of indirection where one has to understand the implementation of the `*Client` class in order to see which calls are made to the underlying `*Connection` class. If I'm implementing a publisher or subscriber, I don't want to have to look into the implementation of the `Publisher` or `Subscriber` class in order to write unit tests.

coryan: > Is there a reason it is done this way as opposed to making a mockable `*Client`?

Mockable classes are not value types.  Value types are easier to use in C++.

> It seems like an extra level of indirection where one has to understand the implementation of the `*Client` class in order to see which calls are made to the underlying `*Connection` class.

Not to *use* the `*Client` class, right?

> If I'm implementing a publisher or subscriber, I don't want to have to look into the implementation of the `Publisher` or `Subscriber` class in order to write unit tests.

Fair enough, that is the tradeoff we made: simpler to use `*Client` objects, for a cost in difficulty to write unit tests.

In any case, we are not going to change this for Pub/Sub: consistency across C++ libraries is also important.

Issue #4230, opened 2020-05-18
clang-tidy incorrectly warns about unused using decls in .h files


devjgm:
In our `build-in-docker-cmake.sh` script, we have a couple optimizations around running clang-tidy, because it's super slow in a repo with lots of C++ source files.

In particular, this part: https://github.com/googleapis/google-cloud-cpp/blob/3adfb06ca8f7c2fd167e5da6d7b9875466ad4b69/ci/kokoro/docker/build-in-docker-cmake.sh#L139-L149

Runs clang-tidy only on the changed `.cc` and `.h` files. However, when we run clang-tidy on a `.h` file like `clang-tidy foo.h`, it sometimes complains about unused using declarations that are actually used in other .cc files.

For now, I think I may disable `misc-unused-using-decls` when checking header files. I'm not sure if there are other "unused" warnings that may also fire, but I haven't seen them yet. It's OK if we miss a warning during our presubmit CI runs, because the post-submit will do a full build and will catch any new errors.

devjgm:
There's currently no action to take on this issue. The current problem was fixed by https://github.com/googleapis/google-cloud-cpp/pull/4223.

coryan:
There is a similar problem with `readability-redundant-declaration`, using `friend` in the header may appear as unused.

PR #5248, opened 2020-10-14
test: enable all clang-tidy presubmit warnings for source files

mr-salty:
As discussed in #4230, running clang-tidy on a subset of files produces
some false positives in header files. However, the fix to that also
disabled some warnings for source files. This PR re-enables our full
set of warnings for source files, allowing presubmit to catch more issues
(see #5240 as an example of a break caused by this.)

To test, I forced the relevant section of the build script to run in a local build and added an unnecessary `using` declaration to a `.h` and `.cc ` file.  Note that both are scanned but the warning only appears for the `.cc` file (which would have been missed prior to this PR)

```
2020-10-14T01:27:21Z: Run clang-tidy on changed files in a presubmit build 
clang-tidy -p=cmake-out/ci-fedora-install-31-clang-tidy -checks="-misc-unused-using-decls,-readability-redundant-declaration" google/cloud/log.h 
10458 warnings generated.
/v/google/cloud/log.h:101:12: error: using decl 'make_pair' is unused [misc-unused-using-decls,-warnings-as-errors]
using std::make_pair;
           ^
/v/google/cloud/log.h:101:12: note: remove the using
using std::make_pair;
~~~~~~~~~~~^~~~~~~~~~
Suppressed 10477 warnings (10457 in non-user code, 20 NOLINT).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
1 warning treated as error
2020-10-14T01:27:25Z: Docker run failed with exit status 123. 
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5248)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -141,14 +141,14 @@ if [[ "${CLANG_TIDY:-}" == "yes" && (\
   SOURCE_FILTER_REGEX='google/cloud/.*\.cc$'
 
   # Get the list of modified files.
-  readonly modified=$(git diff --diff-filter=d --name-only "${TARGET_BRANCH}")
+  readonly MODIFIED=$(git diff --diff-filter=d --name-only "${TARGET_BRANCH}")
 
   # Run clang_tidy against files that regex match the first argument (less some
   # exclusions). Any remaining arguments are passed to clang-tidy.
   run_clang_tidy() {
     local -r file_regex="$1"
     shift
-    grep -E "${file_regex}" <<<"${modified}" |
+    grep -E "${file_regex}" <<<"${MODIFIED}" |
       grep -v google/cloud/bigtable/examples/opencensus |
       grep -v generator/integration_tests/golden |
       xargs --verbose -d '\n' -r -n 1 -P "${NCPU}" clang-tidy \
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -146,7 +146,7 @@ if [[ "${CLANG_TIDY:-}" == "yes" && (\
   # Run clang_tidy against files that regex match the first argument (less some
   # exclusions). Any remaining arguments are passed to clang-tidy.
   run_clang_tidy() {
-    readonly file_regex="$1"
+    local -r file_regex="$1"
     shift
     grep -E "${file_regex}" <<<"${modified}" |
       grep -v google/cloud/bigtable/examples/opencensus |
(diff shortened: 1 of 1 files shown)


coryan (approved):
Thank you for sending this PR.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5248#-:-MJawRKX1mOqt6JdL2-t:boepgo3)* status: all files reviewed, 1 unresolved discussion (waiting on @mr-salty)

___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 144 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5248#-MJawdht9MSWBfo6BAE4:-MJawdht9MSWBfo6BAE5:b-770g2w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f82e2c54890830fcb27d7c5a397d7777506d6eb/ci/kokoro/docker/build-in-docker-cmake.sh#L144)):*
> ```Shell
> 
>   # Get the list of modified files.
>   readonly modified=$(git diff --diff-filter=d --name-only "${TARGET_BRANCH}")
> ```

nit: the style guide says s/modified/MODIFIED/ 

https://google.github.io/styleguide/shellguide.html#constants-and-environment-variable-names

up to you.

mr-salty (commented):
FYI I didn't try to make the two invocations run in parallel since they already do `xargs -P "${NCPU}"` individually.  There still could be a benefit where the # of files modified is < `$NCPU` but it didn't seem worth the extra complication. 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5248#-:-MJcL1ZgCHSdjzrWSN5c:b5i00cd)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 144 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5248#-MJawdht9MSWBfo6BAE4:-MJcKJAMByx0p43hLjCg:b-m6csqw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f82e2c54890830fcb27d7c5a397d7777506d6eb/ci/kokoro/docker/build-in-docker-cmake.sh#L144)):*

nit: the style guide says s/modified/MODIFIED/ 

https://google.github.io/styleguide/shellguide.html#constants-and-environment-variable-names

up to you.

</blockquote></details>

This should really be local to the body of the `if` (or maybe `static` inside the function) but sh doesn't have those concepts :)  so, constant is probably the closest match although a couple sections below that muddies the waters: https://google.github.io/styleguide/shellguide.html#read-only-variables

coryan (approved):
:lgtm:

> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5248#-:-MJcLyXyENFCi-q7A2kY:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5253, opened 2020-10-14
feat(pubsub): support subscription filters

### LINKED ISSUE #5200: Add filter setters to the SubscriptionMutationBuilder
Subscriptions can have filters, we are missing them in the admin APIs for Cloud Pub/Sub.

coryan:
Fixes #5200

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5253)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -218,6 +218,31 @@ void CreateSubscription(google::cloud::pubsub::SubscriptionAdminClient client,
   (std::move(client), argv.at(0), argv.at(1), argv.at(2));
 }
 
+void CreateFilteredSubscription(
+    google::cloud::pubsub::SubscriptionAdminClient client,
+    std::vector<std::string> const& argv) {
+  //! [create-filtered-subscription]
+  namespace pubsub = google::cloud::pubsub;
+  [](pubsub::SubscriptionAdminClient client, std::string const& project_id,
+     std::string const& topic_id, std::string const& subscription_id) {
+    auto sub = client.CreateSubscription(
+        pubsub::Topic(project_id, std::move(topic_id)),
+        pubsub::Subscription(project_id, std::move(subscription_id)),
+        pubsub::SubscriptionBuilder{}.set_filter(
+            R"""(attributes.is-even = "false")"""));
+    if (sub.status().code() == google::cloud::StatusCode::kAlreadyExists) {
+      std::cout << "The subscription already exists\n";
+      return;
+    }
+    if (!sub) throw std::runtime_error(sub.status().message());
+
+    std::cout << "The subscription was successfully created: "
+              << sub->DebugString() << "\n";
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/samples/samples.cc
```diff
@@ -218,6 +218,31 @@ void CreateSubscription(google::cloud::pubsub::SubscriptionAdminClient client,
...
+void CreateFilteredSubscription(
+    google::cloud::pubsub::SubscriptionAdminClient client,
+    std::vector<std::string> const& argv) {
+  //! [create-filtered-subscription]
+  namespace pubsub = google::cloud::pubsub;
+  [](pubsub::SubscriptionAdminClient client, std::string const& project_id,
+     std::string const& topic_id, std::string const& subscription_id) {
+    auto sub = client.CreateSubscription(
+        pubsub::Topic(project_id, std::move(topic_id)),
+        pubsub::Subscription(project_id, std::move(subscription_id)),
```
devbww, line 230: It looks like you intended `topic_id` and `subscription_id` to be taken by value.

google/cloud/pubsub/samples/samples.cc
```diff
@@ -218,6 +218,31 @@ void CreateSubscription(google::cloud::pubsub::SubscriptionAdminClient client,
...
+        pubsub::SubscriptionBuilder{}.set_filter(
+            R"""(attributes.is-even = "false")"""));
+    if (sub.status().code() == google::cloud::StatusCode::kAlreadyExists) {
```
devbww, line 233: Is it preferable to make this check inside the `if (!sub) { ... }` block?

google/cloud/pubsub/samples/samples.cc
coryan: Fixed.

google/cloud/pubsub/samples/samples.cc
coryan: Dunno.  As written I read the code as:

> If you get a `kAlreadyExists`, report it and return. If you get any (other) error, throw an exception. Otherwise report success.

Otherwise it reads:

> If you get any error, then check if it is `kAlreadyExists`, in that case reporting it and return, otherwise throw an exception. Otherwise report success.

The first seems slightly more readable, but I realize that might not be an argument that everybody agrees on.  FWIW, it is the pattern we use elsewhere in these examples (naturally if I am wrong that only means we have to change the pattern in more places).

devbww: Ack on the other cases, although I would still say its less readable.  That is, I would tend to interpret a `StatusOr` as a true union/variant and never try to access `status()` until I know I'm in the error case, just like I wouldn't access `value()` until I know I'm in the non-error case.

But not a big deal, given how its implemented.  Thanks.

devbww (approved):

PR #5274, opened 2020-10-19
add error handling example from client->ReadObject() in README.md

mohamedAlaaK:
handles error in quickstart example in README.md

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274)

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,11 @@ int main(int argc, char* argv[]) {
   }
 
   auto reader = client->ReadObject(bucket_name, "quickstart.txt");
+  if (!reader) {
+    std::cerr << "Error reading object: " << reader.status() << "\n";
+    return 1;
+  }
+
   std::string contents{std::istreambuf_iterator<char>{reader}, {}};
   std::cout << contents << "\n";
 
--- a/google/cloud/storage/README.md
+++ b/google/cloud/storage/README.md
@@ -74,6 +74,11 @@ int main(int argc, char* argv[]) {
   }
 
   auto reader = client->ReadObject(bucket_name, "quickstart.txt");
+  if (!reader) {
+    std::cerr << "Error reading object: " << reader.status() << "\n";
+    return 1;
+  }
+
   std::string contents{std::istreambuf_iterator<char>{reader}, {}};
   std::cout << contents << "\n";
 
--- a/google/cloud/storage/quickstart/quickstart.cc
+++ b/google/cloud/storage/quickstart/quickstart.cc
@@ -48,6 +48,11 @@ int main(int argc, char* argv[]) {
   }
 
   auto reader = client->ReadObject(bucket_name, "quickstart.txt");
+  if (!reader) {
+    std::cerr << "Error reading object: " << reader.status() << "\n";
+    return 1;
+  }
+
   std::string contents{std::istreambuf_iterator<char>{reader}, {}};
   std::cout << contents << "\n";
 


mohamedAlaaK:
@googlebot I signed it!

README.md
```diff
@@ -121,6 +121,12 @@ int main(int argc, char* argv[]) {
   }
 
   auto reader = client->ReadObject(bucket_name, "quickstart.txt");
+  if (reader.bad()) {
```
remyabel, line 124: I think this should be `if (!reader)` instead of `if (!reader.bad())`. What do you think?

This quickstart is copied from the storage readme (and consequently from quickstart.cc). I think you should update all three places for consistency.

remyabel: Actually, you should update `quickstart.cc` then run `ci/generate-markdown/generate-storage-readme.sh`

mohamedAlaaK: yes you are right.
i modified it.

coryan (changes requested):
Thank you for sending this PR, and sorry for the late reply. One comment below.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-:-MK5A6ei40qXLh7K1O8C:bejihkv)* status: all files reviewed, 2 unresolved discussions (waiting on @mohamedAlaaK and @remyabel)

___
*[google/cloud/storage/quickstart/quickstart.cc, line 51 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK58t6TEMNXVPqa80-e:-MK58t6UBQFLeCSU5_Ge:b-5dtou1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/google/cloud/storage/quickstart/quickstart.cc#L51)):*
> ```cpp
> 
>   auto reader = client->ReadObject(bucket_name, "quickstart.txt");
>   if (!reader) {
> ```

I think we should move this further down.  One should check for errors during reading, not just opening the stream.  Sadly, `if (!reader)` won't work, because that flag is set on `eof()` and on `failbit()` (which is set on short reads), I think the correct code would be:

```cc
auto reader = ...;
std::string contents{std::istreambuf_iterator<char>{reader}, {}};
if (reader.bad()) {
  std::cerr << "Error reading object: " << reader.status() << "\n";
  return 1;
}
```

mohamedAlaaK:
___
*[google/cloud/storage/quickstart/quickstart.cc, line 51 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK58t6TEMNXVPqa80-e:-MK8v8eJBGQSeQxxERsc:bo8zrkj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/google/cloud/storage/quickstart/quickstart.cc#L51)):*

I think we should move this further down.  One should check for errors during reading, not just opening the stream.  Sadly, `if (!reader)` won't work, because that flag is set on `eof()` and on `failbit()` (which is set on short reads), I think the correct code would be:

```cc
auto reader = ...;
std::string contents{std::istreambuf_iterator<char>{reader}, {}};
if (reader.bad()) {
  std::cerr << "Error reading object: " << reader.status() << "\n";
  return 1;
}
```

</blockquote></details>

Ok i understand the reason behind moving the code 1 line down to check at once for both errors from opening and reading the stream .
but i don't get the exact difference between `!reader` and `reader.bad()`.

coryan (commented):
Hmm... I think I was overthinking this... +kokoro:run 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-:-MK9k_8g-dAVF6L0LNi8:b-7n88n1)* status: all files reviewed, 2 unresolved discussions (waiting on @mohamedAlaaK and @remyabel)

___
*[google/cloud/storage/quickstart/quickstart.cc, line 51 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK58t6TEMNXVPqa80-e:-MK9fG6_55czN3fHVSC5:bx8cira) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/google/cloud/storage/quickstart/quickstart.cc#L51)):*
<details><summary><i>Previously, mohamedAlaaK wrote…</i></summary><blockquote>

Ok i understand the reason behind moving the code 1 line down to check at once for both errors from opening and reading the stream .
but i don't get the exact difference between `!reader` and `reader.bad()`.

</blockquote></details>

There is a useful table at:

https://en.cppreference.com/w/cpp/io/basic_ios/bad

But thinking some more on it, in this case it won't matter, as we are reading one character at a time **and** we are reading through the streambuf, skipping the fail bits altogether.  The only way to report errors while reading from a streambuf is to throw an exception, and that is what we do.


FWIW, The difference is relevant when you do something like:

```cc
char buf[4096];
stream.read(buf, 4096);
```

If the stream has less than 4096 characters then `.read()` will set the `failbit`:

https://en.cppreference.com/w/cpp/io/basic_istream/read

that makes `!stream` return `false`, and most folks would not call that an "error".

coryan (changes requested):
The code looks good, some formatting nits in:

https://source.cloud.google.com/results/invocations/f9388493-c902-47fe-85a2-a64399cc2074/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fdocker%2Fclang-tidy-presubmit/log

```patch
diff --git a/README.md b/README.md
index dfc307bee..21858037c 100644
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ int main(int argc, char* argv[]) {
               << "\n";
     return 1;
   }
-  
+
   std::string contents{std::istreambuf_iterator<char>{reader}, {}};
   std::cout << contents << "\n";
 
diff --git a/google/cloud/storage/quickstart/quickstart.cc b/google/cloud/storage/quickstart/quickstart.cc
index 21f6449f0..c9837ebb6 100644
--- a/google/cloud/storage/quickstart/quickstart.cc
+++ b/google/cloud/storage/quickstart/quickstart.cc
@@ -49,11 +49,10 @@ int main(int argc, char* argv[]) {
 
   auto reader = client->ReadObject(bucket_name, "quickstart.txt");
   if (!reader) {
-    std::cerr << "Error reading object: " << reader.status()
-              << "\n";
+    std::cerr << "Error reading object: " << reader.status() << "\n";
     return 1;
   }
-  
+
   std::string contents{std::istreambuf_iterator<char>{reader}, {}};
   std::cout << contents << "\n";
```

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-:-MK9nUWu3gQL2SXRWtfG:b-mzlztx)* status: all files reviewed, 5 unresolved discussions (waiting on @mohamedAlaaK and @remyabel)

___
*[README.md, line 125 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK9nmsG-4Ue4X_q7lgz:-MK9nmsH3o6TJI4oCQVR:b-wtl25r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/README.md#L125)):*
> ```Markdown
>   auto reader = client->ReadObject(bucket_name, "quickstart.txt");
>   if (!reader) {
>     std::cerr << "Error reading object: " << reader.status()
> ```

this should be all in one line.
___
*[README.md, line 129 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK9nfNXELU77Q28elMu:-MK9nfNY9KM-k7Ia1SnR:b-8gb5ac) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/README.md#L129)):*
> ```Markdown
>     return 1;
>   }
>   
> ```

there are two extra spaces here.
___
*[google/cloud/storage/quickstart/quickstart.cc, line 52 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK9nl5FEMcIZza7Fve4:-MK9nl5FEMcIZza7Fve5:b-wtl25r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/google/cloud/storage/quickstart/quickstart.cc#L52)):*
> ```cpp
>   auto reader = client->ReadObject(bucket_name, "quickstart.txt");
>   if (!reader) {
>     std::cerr << "Error reading object: " << reader.status()
> ```

this should be all in one line.
___
*[google/cloud/storage/quickstart/quickstart.cc, line 56 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK9niSq27tR2PucucPy:-MK9niSq27tR2PucucPz:b-4fgc0c) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/google/cloud/storage/quickstart/quickstart.cc#L56)):*
> ```cpp
>     return 1;
>   }
>   
> ```

two extra spaces here.

mohamedAlaaK:
___
*[google/cloud/storage/quickstart/quickstart.cc, line 51 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK58t6TEMNXVPqa80-e:-MKAG0R9Cy2ri6DKIVS8:b-6twk3j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/google/cloud/storage/quickstart/quickstart.cc#L51)):*

There is a useful table at:

https://en.cppreference.com/w/cpp/io/basic_ios/bad

But thinking some more on it, in this case it won't matter, as we are reading one character at a time **and** we are reading through the streambuf, skipping the fail bits altogether.  The only way to report errors while reading from a streambuf is to throw an exception, and that is what we do.


FWIW, The difference is relevant when you do something like:

```cc
char buf[4096];
stream.read(buf, 4096);
```

If the stream has less than 4096 characters then `.read()` will set the `failbit`:

https://en.cppreference.com/w/cpp/io/basic_istream/read

that makes `!stream` return `false`, and most folks would not call that an "error".
</blockquote></details>

that makes sense. thanks for the explanation.

mohamedAlaaK:
___
*[google/cloud/storage/quickstart/quickstart.cc, line 52 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK9nl5FEMcIZza7Fve4:-MKAGMjF-GSuZKAYM84x:b9n9iu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/google/cloud/storage/quickstart/quickstart.cc#L52)):*

this should be all in one line.
</blockquote></details>

i did it to make it similar to the if condition above. was it because it exceeded the line width?
i will modify it anyways.

mohamedAlaaK (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @coryan and @remyabel)

___
*[README.md, line 125 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK9nmsG-4Ue4X_q7lgz:-MKAHwlqDoYBHXyAKqmj:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/README.md#L125)):*

this should be all in one line.
</blockquote></details>

Done.
___
*[README.md, line 129 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK9nfNXELU77Q28elMu:-MKAHvzJAHc9lHBdmC_K:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/README.md#L129)):*

there are two extra spaces here.
</blockquote></details>

Done.
___
*[google/cloud/storage/quickstart/quickstart.cc, line 56 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-MK9niSq27tR2PucucPy:-MKAHOkkBfZnox0K5AA4:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/da5d289e77b5140c721c2b484589c9cd7624ceeb/google/cloud/storage/quickstart/quickstart.cc#L56)):*

two extra spaces here.
</blockquote></details>

Done.

coryan (commented):
+kokoro:run looking good.

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-:-MKAI3_i11ZJU7ahOomB:buosgca)* status: all files reviewed, 1 unresolved discussion (waiting on @remyabel)

coryan (approved):
:lgtm: Thank you for being patient during the long review.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5274#-:-MKFIA971AtCJRP_wxBM:bk4mc0z)* status: all files reviewed, 1 unresolved discussion (waiting on @remyabel)

PR #5292, opened 2020-10-20
fix(pubsub): handle errors on first Read()

### LINKED ISSUE #5267: Error detection is broken for streaming pulls
Consider a request to a non-existent subscription and/or with invalid permissions.

Setting up the stream (the initial `Start()` call) would succeed.

The `Write()` call would succeed, because *sending* the `Write()` request would be successful.

The initial `Read()` call is the first opportunity the service has to signal an error, and it does.

The code triggers a reconnect loop on `Read()` errors, because it wants to resume that it considers a "successfully connected" stream.  This is incorrect, we should only consider the stream as successfully connected after at least one `Read()` succeeds.

coryan:
For a streaming pull, the service may not be able to report some errors,
e.g, whether the caller has the right permissions, or whether the
subscription exists, until the first `Read()`. This change modifies the
startup sequence to include the first read. Inlining all of this in a
single lambda seemed too ugly, so we unroll into several callbacks.

Fixes #5267

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5292)

### DIFF
--- a/google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
+++ b/google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #include "google/cloud/pubsub/internal/streaming_subscription_batch_source.h"
-#include "google/cloud/internal/async_retry_loop.h"
 #include "google/cloud/log.h"
 #include <ostream>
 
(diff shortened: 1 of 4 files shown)


google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
```diff
@@ -72,55 +70,42 @@ void StreamingSubscriptionBatchSource::ExtendLeases(
...
+    std::shared_ptr<pubsub::BackoffPolicy> backoff_policy) {
+  GCP_LOG(DEBUG) << __func__ << "()";
+  // Starting a stream is a 4 step process.
+  // 1. Create a new SubscriberStub::AsyncPullStream object.
+  // 2. Call Start() on it, which is asynchronous, which might fail, but rarely
+  //    does
+  // 3. Call Write() on it, which is also asynchronous and almost always
+  //    succeeds, could fail if the endpoint is invalid
+  // 4. Call Read() on it, which is also asynchronous, and it is the first
+  //    change to get errors such as kPermissionDenied, or kNotFound.
```
devbww, line 84: s/change/chance/

devbww (approved):

google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
coryan: Thanks, fixed.

PR #5295, opened 2020-10-21
feat(pubsub)!: share connections for publishers

### LINKED ISSUE #5266: Allow applications to share Stub and gRPC channels in Publisher/Subscriber
As currently defined a `pubsub::Publisher` is bound to a specific topic, that is probably the "Right Thing":tm: but it is done by binding the underlying `pubsub::PublisherConnection` to a specific topic too, which means applications cannot share gRPC resources (stubs, channels, sockets) across `Publishers`.

That seems less than ideal, the application should have better control of what resources are bound to what.  Indirectly they do, because gRPC shares sockets with the same attributes, even if created by different channels, but that seems a bit awkward to depend on.

The proposed changes in `pubsub::Publisher` are:

```cc
namespace pubsub = google::cloud::pubsub;

void NewStyleP(pubsub::Topic t1, pubsub::Topic t2) {
  auto connection = pubsub::MakePublisherConnection();
  auto p1 = pubsub::Publisher(
    t1, pubsub::PublisherConfig{}, connection);
  auto p2 = pubsub::Publisher(
    t2, pubsub::PublisherConfig{}, connection);
  p1.Publish(...); // no change
  p2.Publish(...); // no change
}

void CurrentStyleP(pubsub::Topic t1, pubsub::Topic t2) {
  // Sharing is indirect, the application has to reason about gRPC internals to figure out what is going on.
  auto p1 = pubsub::Publisher(pubsub::MakePublisherConnection(t1, pubsub::PublisherConfig{}));
  auto p2 = pubsub::Publisher(pubsub::MakePublisherConnection(t2, pubsub::PublisherConfig{}));
  p1.Publish(...); // no change
  p2.Publish(...); // no change
}
```

We would need similar changes in the `Subscriber`.

coryan:
**BREAKING CHANGE:**

* Allow sharing connections between multiple `pubsub::Publisher` objects.
  Creating a `pubsub::PublisherConnection` no longer requires a `pubsub::Topic`
  or the `pubsub::PublisherOptions`, these are now parameters for the
  `pubsub::Publisher` constructor.

Fixes #5266

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5295)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ are close enough to a GA release that we think highlighting them is important.
 
 * Allow sharing connections between multiple `pubsub::Publisher` objects.
   Creating a `pubsub::PublisherConnection` no longer requires a `pubsub::Topic`
-  or the `pubsub::PublisherOptions`, these are now parameters for the
+  or the `pubsub::PublisherOptions`. These are now parameters for the
   `pubsub::Publisher` constructor.
 
 * Rename `pubsub::SubscriptionOptions` to `pubsub::SubscriberOptions` as these
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@
 While the Pub/Sub library is not GA, and breaking changes are to be expected, we
 are close enough to a GA release that we think highlighting them is important.
 
+* Allow sharing connections between multiple `pubsub::Publisher` objects.
+  Creating a `pubsub::PublisherConnection` no longer requires a `pubsub::Topic`
+  or the `pubsub::PublisherOptions`, these are now parameters for the
+  `pubsub::Publisher` constructor.
+
 * Rename `pubsub::SubscriptionOptions` to `pubsub::SubscriberOptions` as these
   are bound to a specific subscriber object.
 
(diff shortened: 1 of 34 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,11 @@
 While the Pub/Sub library is not GA, and breaking changes are to be expected, we
 are close enough to a GA release that we think highlighting them is important.
 
+* Allow sharing connections between multiple `pubsub::Publisher` objects.
+  Creating a `pubsub::PublisherConnection` no longer requires a `pubsub::Topic`
+  or the `pubsub::PublisherOptions`, these are now parameters for the
```
devbww, line 14: s/, these/. These/

google/cloud/pubsub/publisher_connection.h
```diff
@@ -61,18 +61,17 @@ class PublisherConnection {
 
   /// Wrap the arguments for `Publish()`
   struct PublishParams {
-    Message message;
+    google::pubsub::v1::PublishRequest request;
   };
```
devbww, line 66: Remove extra blank line after following deletion?

devbww (approved):

Issue #4640, opened 2020-07-23
Implement pubsub_resume_publish_with_ordering_keys sample


coryan:

coryan:
Fixed in #5340

PR #5304, opened 2020-10-22
feat(pubsub): sequence batches with ordering keys

### LINKED ISSUE #4803: Sequence batches when using ordering keys
For publishers with ordering keys a batch should not be sent until all previous batches have been acknowledged successfully. And if a batch fails, any queued messages should fail too.

This sounds like a form of "corking".

coryan:
While a batch of messages with the same ordering key is pending (i.e.
sent but the response is not received) any additional messages should be
batched. Add an example showing how to use the `ResumePublish()` API.

Fixes #4803 and #4640

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5304)

### DIFF
--- a/google/cloud/pubsub/integration_tests/subscriber_integration_test.cc
+++ b/google/cloud/pubsub/integration_tests/subscriber_integration_test.cc
@@ -48,6 +48,8 @@ class SubscriberIntegrationTest : public ::testing::Test {
     topic_ = Topic(project_id, pubsub_testing::RandomTopicId(generator_));
     subscription_ = Subscription(
         project_id, pubsub_testing::RandomSubscriptionId(generator_));
+    ordered_subscription_ = Subscription(
+        project_id, pubsub_testing::RandomSubscriptionId(generator_));
 
     auto topic_admin = TopicAdminClient(MakeTopicAdminConnection());
     auto subscription_admin =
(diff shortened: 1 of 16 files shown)


google/cloud/pubsub/integration_tests/subscriber_integration_test.cc
```diff
@@ -62,13 +64,26 @@ class SubscriberIntegrationTest : public ::testing::Test {
     ASSERT_THAT(
         subscription_metadata,
         AnyOf(StatusIs(StatusCode::kOk), StatusIs(StatusCode::kAlreadyExists)));
+    auto ordered_subscription_metadata = subscription_admin.CreateSubscription(
+        topic_, ordered_subscription_,
+        SubscriptionBuilder{}
+            .set_ack_deadline(std::chrono::seconds(30))
+            .enable_message_ordering(true));
+    ASSERT_THAT(
+        subscription_metadata,
```
devbww, line 73: s/subscription_metadata/ordered_subscription_metadata/?

google/cloud/pubsub/publisher.h
```diff
@@ -50,8 +49,17 @@ inline namespace GOOGLE_CLOUD_CPP_PUBSUB_NS {
  * });
  * @endcode
  *
- * @par Performance
+ * @par Message Ordering
+ * A `Publisher` configured to preserve message ordering will sequence the
+ * messages that share a common ordering key (see
+ * `MessageBuilder::SetOrderingKey()`). Messages will be batched by ordering
+ * key, and new batches will wait until the status of the previous batch is
+ * known. Or an error, all pending and queued messages are discarded, and the
```
devbww, line 57: s/Or an error/On an error/?

coryan: Thanks, fixed.

google/cloud/pubsub/integration_tests/subscriber_integration_test.cc
coryan: Yup, thanks.

devbww (approved):

PR #5305, opened 2020-10-22
feat(storage): Move nlohmann/json.hpp includes to cc file

remyabel:
Part of the work for: #2772

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5305)

PR #5311, opened 2020-10-22
feat(storage): move nlohmann/json.hpp includes to cc file

remyabel:
Fixes: #2772

This completes the work of removing the `nlohmann/json.hpp` include from public
headers. `storage_include_test.cc` has been amended to test this.
`PatchBuilder` has been changed to follow the Pimpl idiom and some unused
functions were removed. `SetArrayField` no longer takes a vector but now a
`std::string` which represents a stringified array.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5311)

PR #5313, opened 2020-10-22
fix: include nlohmann with angle brackets

devjgm:
This was accidentally introduced today in https://github.com/googleapis/google-cloud-cpp/pull/5305. A big PR; easy to overlook this. The code works fine externally, but it breaks during our import script.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5313)


coryan:
I think #5311 is making the same change.

coryan (approved):

PR #5327, opened 2020-10-23
feat(pubsub): preserve order for messages with ordering keys

### LINKED ISSUE #5044: Automatically schedule messages with ordering keys one at a time

### LINKED ISSUE #5263: PubSub: message ordering
OKR tracking issue.

coryan:
Messages without an ordering key are scheduled
in parallel, and therefore can be executed out of
order. Messages with an ordering key are scheduled
in sequence for the same ordering key. That is,
messages with key `key1` will be scheduled one
at a time, but messages with other keys may be
scheduled in parallel.

Fixes #5044 and fixes #5263 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327)

### DIFF
--- a/google/cloud/pubsub/BUILD
+++ b/google/cloud/pubsub/BUILD
@@ -25,6 +25,8 @@ cc_library(
     deps = [
         "//google/cloud:google_cloud_cpp_common",
         "//google/cloud:google_cloud_cpp_grpc_utils",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/container:flat_hash_set",
         "@com_google_googleapis//google/pubsub/v1:pubsub_cc_grpc",
     ],
 )
@@ -70,6 +72,7 @@ load(":pubsub_client_unit_tests.bzl", "pubsub_client_unit_tests")
         "//google/cloud:google_cloud_cpp_common",
         "//google/cloud/testing_util:google_cloud_cpp_testing",
         "//google/cloud/testing_util:google_cloud_cpp_testing_grpc",
+        "@com_google_absl//absl/strings:str_format",
         "@com_google_googletest//:gtest_main",
     ],
 ) for test in pubsub_client_unit_tests]
--- a/google/cloud/pubsub/BUILD
+++ b/google/cloud/pubsub/BUILD
@@ -26,7 +26,6 @@ cc_library(
         "//google/cloud:google_cloud_cpp_common",
         "//google/cloud:google_cloud_cpp_grpc_utils",
         "@com_google_absl//absl/container:flat_hash_map",
-        "@com_google_absl//absl/container:flat_hash_set",
         "@com_google_googleapis//google/pubsub/v1:pubsub_cc_grpc",
     ],
 )
--- a/google/cloud/pubsub/CMakeLists.txt
+++ b/google/cloud/pubsub/CMakeLists.txt
@@ -121,8 +121,7 @@ target_include_directories(
 target_link_libraries(
     pubsub_client
     PUBLIC google_cloud_cpp_grpc_utils google_cloud_cpp_common
-           googleapis-c++::pubsub_protos absl::flat_hash_map
-           absl::flat_hash_set)
+           googleapis-c++::pubsub_protos absl::flat_hash_map)
 google_cloud_cpp_add_common_options(pubsub_client)
 set_target_properties(
     pubsub_client PROPERTIES VERSION "${GOOGLE_CLOUD_CPP_VERSION}"
--- a/google/cloud/pubsub/CMakeLists.txt
+++ b/google/cloud/pubsub/CMakeLists.txt
@@ -25,6 +25,11 @@ set(DOXYGEN_PREDEFINED "GOOGLE_CLOUD_CPP_PUBSUB_NS=v1")
 
 include(GoogleCloudCppCommon)
 
+# TODO(#4146) - remove FPHSA_NAME_MISMATCHED manipulation on next absl release
+set(FPHSA_NAME_MISMATCHED Threads) # Quiet warning caused by Abseil
+find_package(absl CONFIG REQUIRED)
+unset(FPHSA_NAME_MISMATCHED)
+
 configure_file(version_info.h.in ${CMAKE_CURRENT_SOURCE_DIR}/version_info.h)
 add_library(
     pubsub_client # cmake-format: sort
@@ -114,8 +119,10 @@ target_include_directories(
            $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>
            $<INSTALL_INTERFACE:include>)
 target_link_libraries(
-    pubsub_client PUBLIC google_cloud_cpp_grpc_utils google_cloud_cpp_common
-                         googleapis-c++::pubsub_protos)
+    pubsub_client
+    PUBLIC google_cloud_cpp_grpc_utils google_cloud_cpp_common
+           googleapis-c++::pubsub_protos absl::flat_hash_map
+           absl::flat_hash_set)
 google_cloud_cpp_add_common_options(pubsub_client)
 set_target_properties(
     pubsub_client PROPERTIES VERSION "${GOOGLE_CLOUD_CPP_VERSION}"
@@ -236,6 +243,7 @@ function (google_cloud_cpp_pubsub_client_define_tests)
                     googleapis-c++::pubsub_client
                     google_cloud_cpp_testing
                     google_cloud_cpp_testing_grpc
+                    absl::str_format
                     GTest::gmock_main
                     GTest::gmock
                     GTest::gtest)
--- a/google/cloud/pubsub/CMakeLists.txt
+++ b/google/cloud/pubsub/CMakeLists.txt
@@ -119,9 +119,8 @@ target_include_directories(
            $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>
            $<INSTALL_INTERFACE:include>)
 target_link_libraries(
-    pubsub_client
-    PUBLIC google_cloud_cpp_grpc_utils google_cloud_cpp_common
-           googleapis-c++::pubsub_protos absl::flat_hash_map)
+    pubsub_client PUBLIC google_cloud_cpp_grpc_utils google_cloud_cpp_common
+                         googleapis-c++::pubsub_protos absl::flat_hash_map)
 google_cloud_cpp_add_common_options(pubsub_client)
 set_target_properties(
     pubsub_client PROPERTIES VERSION "${GOOGLE_CLOUD_CPP_VERSION}"
--- a/google/cloud/pubsub/internal/subscription_message_queue.cc
+++ b/google/cloud/pubsub/internal/subscription_message_queue.cc
@@ -40,18 +40,20 @@ void SubscriptionMessageQueue::Shutdown() {
 void SubscriptionMessageQueue::Read(std::size_t max_callbacks) {
   std::unique_lock<std::mutex> lk(mu_);
   if (!callback_) return;
-  read_count_ += max_callbacks;
+  available_slots_ += max_callbacks;
   DrainQueue(std::move(lk));
 }
 
 future<Status> SubscriptionMessageQueue::AckMessage(std::string const& ack_id,
                                                     std::size_t) {
+  HandlerDone(ack_id);
   source_->AckMessage(ack_id);
   return make_ready_future(Status{});
 }
 
 future<Status> SubscriptionMessageQueue::NackMessage(std::string const& ack_id,
                                                      std::size_t) {
+  HandlerDone(ack_id);
   source_->NackMessage(ack_id);
   return make_ready_future(Status{});
 }
@@ -72,9 +74,17 @@ void SubscriptionMessageQueue::OnRead(
     google::pubsub::v1::StreamingPullResponse r) {
   auto handle_response = [&] {
     shutdown_manager_->FinishedOperation("OnRead");
-    std::move(r.mutable_received_messages()->begin(),
-              r.mutable_received_messages()->end(),
-              std::back_inserter(messages_));
+    for (auto& m : *r.mutable_received_messages()) {
+      auto key = m.message().ordering_key();
+      auto const ordered = !key.empty();
+      auto& queue = queues_[key];
+      queue.messages.push_back(std::move(m));
+      // Queues require ordering are runnable only if there are no running
+      // messages
+      if (!ordered || queue.running == 0) {
+        runnable_queues_.insert(std::move(key));
+      }
+    }
     DrainQueue(std::move(lk));
   };
   auto bulk_nack = [&] {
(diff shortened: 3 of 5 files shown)


coryan:
This is the last feature for GA.. :rocket:

mr-salty (changes requested):
> Reviewed 3 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327)* status: 3 of 5 files reviewed, 7 unresolved discussions (waiting on @coryan)

___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 82 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMPTrBEsK5EDtOoQ3f:-MKMPTrBEsK5EDtOoQ3g:b8ky60y) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L82)):*
> ```
> Queues require
> ```
 
"Queues that require" ...
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 84 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMPa9p9ffORet8Dj5-:-MKMPa9p9ffORet8Dj50:b-cqyk2h) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L84)):*
> ```cpp
>       // Queues require ordering are runnable only if there are no running
>       // messages
>       if (!ordered || queue.running == 0) {
> ```

is it worth adding  `&& queue.size() == 1` here?  should be marginally cheaper than a hash insert.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 116 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMQP_m7WNiowOf5LWo:-MKMQP_n3DeAwO4bUxii:b-byp8e7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L116)):*
> ```cpp
>     queues_.clear();
>     return ack_ids;
>   }(std::move(lk));
> ```

this is a pretty subtle way to drop the lock - it took me a while to get it, I'd rather see an explicit `unlock()`.

I'd probably write this with a swap to minimize the lock holding:

```
QueueByOrderingKey queues;
queues.swap(queues_);  // maybe std::move would also do this?
lk.unlock();

// loop as before, no lambda needed
```
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 142 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMWjbr8Ox8-k8lwJER:-MKMWjbr8Ox8-k8lwJES:b-dbjw40) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L142)):*
> ```cpp
>       0, runnable_queues_.size() - 1)(generator_);
>   auto r = runnable_queues_.begin();
>   for (std::size_t i = 0; i != idx; ++i) ++r;  // std::advance() did not work
> ```

how performance critical is this and how large do we expect `runnable_queues_` and `available_slots_` to be?

My first idea was to have this return a vector, and use `std::sample` so we only walk the list once, although I'm not sure if your `std::advance` issue would stymie using that.

You could also make a linked list of runnable queues instead of having a second map.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 148 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMa91A2REHWk7MSKhX:-MKMa91A2REHWk7MSKhY:b-7j57ii) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L148)):*
> ```
>   if (ql == queues_.end()) return {};
>   auto& q = ql->second;
>   if (q.messages.empty()) {
> ```

IIUC neither of these conditions (can't find the queue, or messages empty) should happen unless there's a bug? I'd add comments or a DCHECK - or maybe even just CHECK.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 167 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMbmtiAPmcuW4x1s6a:-MKMbmtiAPmcuW4x1s6b:b5r4i1z) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L167)):*
> ```cpp
>   std::lock_guard<std::mutex> lk(mu_);
>   auto loc = ordering_key_by_ack_id_.find(ack_id);
>   if (loc == ordering_key_by_ack_id_.end()) return;
> ```

similar comments as above (although in this method I think this *can* happen if `shutdown_` is true so maybe we should check that explicitly?)

mostly, I just want it to be clear if you're checking these results to be defensive (which I *think* is the case modulo `shutdown_`) or if there's some subtle way those cases can happen.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 171 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMckjE6nMtmAwGIghL:-MKMckjE6nMtmAwGIghM:beioldm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L171)):*
> ```cpp
>   ordering_key_by_ack_id_.erase(loc);
>   auto& q = queues_[key];
>   if (--q.running == 0 && !q.messages.empty()) {
> ```

given my comment above this might be an actual bug?  if we're post shutdown_ then queues_[key] may not exist and `running` becomes `-1` here (maybe something else prevents that, and it probably doesn't matter since we're shutting down)

coryan:
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 142 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMWjbr8Ox8-k8lwJER:-MKMkF-rDzYT0Vk3aQn1:b-ejustm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L142)):*

how performance critical is this and how large do we expect `runnable_queues_` and `available_slots_` to be?

My first idea was to have this return a vector, and use `std::sample` so we only walk the list once, although I'm not sure if your `std::advance` issue would stymie using that.

You could also make a linked list of runnable queues instead of having a second map.
</blockquote></details>

`available_slots_` is small-ish, probably comparable to the number of cores, `runnable_queues_` I have no idea.  Probably thousands at most, as we only accept thousands of incoming messages at a time.

`std::samples()` is C++17, not available to us.

I think the core issue you are highlighting is the poor performance of this, thanks, I did not think about that.  A single pass getting messages would be more efficient, but it would be really really hard to get to work right.  We need to release the lock when passing the message to the next layer, as it can callback and that requires acquiring the lock again.  If I pick N messages other threads may consume the available slots, and then I have to re-enqueue the messages, but how do I preserve their order?

I think I can make this work by picking the ordering key for N queues, and processing *those* one at a time, picking a new set of M queues if needed at the end.  Let me give that a try.

coryan (commented):
I completely changed the implementation for something simpler.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-:-MKVkRwT84QXXl4Cits_:b-j94yf4)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @mr-salty)

___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 82 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMPTrBEsK5EDtOoQ3f:-MKMjtKG0tsb7_Jszk-P:bd1ygo1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L82)):*

> ```
> Queues require
> ```
 
"Queues that require" ...
</blockquote></details>

N/A by now I think.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 84 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMPa9p9ffORet8Dj5-:-MKMjuCJ3th-SgOlNhrr:bd1ygo1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L84)):*

is it worth adding  `&& queue.size() == 1` here?  should be marginally cheaper than a hash insert.
</blockquote></details>

N/A by now I think.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 116 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMQP_m7WNiowOf5LWo:-MKMkDc3EPjthGQXXC87:b-pzy150) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L116)):*

this is a pretty subtle way to drop the lock - it took me a while to get it, I'd rather see an explicit `unlock()`.

I'd probably write this with a swap to minimize the lock holding:

```
QueueByOrderingKey queues;
queues.swap(queues_);  // maybe std::move would also do this?
lk.unlock();

// loop as before, no lambda needed
```
</blockquote></details>

N/A by now.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 142 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMWjbr8Ox8-k8lwJER:-MKVXZ2j52UwCkrb-RcB:b-w4oknn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L142)):*

`available_slots_` is small-ish, probably comparable to the number of cores, `runnable_queues_` I have no idea.  Probably thousands at most, as we only accept thousands of incoming messages at a time.

`std::samples()` is C++17, not available to us.

I think the core issue you are highlighting is the poor performance of this, thanks, I did not think about that.  A single pass getting messages would be more efficient, but it would be really really hard to get to work right.  We need to release the lock when passing the message to the next layer, as it can callback and that requires acquiring the lock again.  If I pick N messages other threads may consume the available slots, and then I have to re-enqueue the messages, but how do I preserve their order?

I think I can make this work by picking the ordering key for N queues, and processing *those* one at a time, picking a new set of M queues if needed at the end.  Let me give that a try.
</blockquote></details>

I completely changed the implementation, N/A.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 148 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMa91A2REHWk7MSKhX:-MKMl8xV4Uh7dPrFFjTl:b-oqhhxc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L148)):*

> ```
>   if (ql == queues_.end()) return {};
>   auto& q = ql->second;
>   if (q.messages.empty()) {
> ```

IIUC neither of these conditions (can't find the queue, or messages empty) should happen unless there's a bug? I'd add comments or a DCHECK - or maybe even just CHECK.
</blockquote></details>

`DCHECK` and `CHECK` do not exist in this code, nor do we have anything equivalent...  It was purely defensive.  Left comments for any relevant cases in the new code.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 167 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMbmtiAPmcuW4x1s6a:-MKVXjCzDsJuefnFDR97:b-vbydso) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L167)):*

similar comments as above (although in this method I think this *can* happen if `shutdown_` is true so maybe we should check that explicitly?)

mostly, I just want it to be clear if you're checking these results to be defensive (which I *think* is the case modulo `shutdown_`) or if there's some subtle way those cases can happen.
</blockquote></details>

See new comments.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 171 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMckjE6nMtmAwGIghL:-MKVXmoI9F34PoN-X7_0:b-vbydso) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L171)):*

given my comment above this might be an actual bug?  if we're post shutdown_ then queues_[key] may not exist and `running` becomes `-1` here (maybe something else prevents that, and it probably doesn't matter since we're shutting down)
</blockquote></details>

See new comments.

mr-salty (approved):
:lgtm: definitely like this better, thanks

> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-:-MKa0HkI8kzC9Gx4uBm4:b2h8w62)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/pubsub/internal/subscription_message_queue.h, line 41 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MK_mxSG1AFC49QE1gvd:-MK_mxSH7wwbdrQIx9W_:b9kz3d1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/29380c3ccc908d970c135b999eaf3c4d8846e51f/google/cloud/pubsub/internal/subscription_message_queue.h#L41)):*
> ```
> setups
> ```

s/setups/sets up/
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 116 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMQP_m7WNiowOf5LWo:-MK_oe3WBMNaKQVj38nY:bcmd67c) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L116)):*

N/A by now.
</blockquote></details>

I still think explicit `lk.unlock()` is clearer.

The other part of my suggestion is still possible, although you'd need to do 2 swaps - completely up to you.
___
*[google/cloud/pubsub/internal/subscription_message_queue_test.cc, line 216 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKa-Puo3jy8J2GgRvsX:-MKa-Puo3jy8J2GgRvsY:bjmih0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/29380c3ccc908d970c135b999eaf3c4d8846e51f/google/cloud/pubsub/internal/subscription_message_queue_test.cc#L216)):*
> ```cpp
> TEST(SubscriptionMessageQueueTest, RespectOrderingKeys) {
>   auto mock = std::make_shared<pubsub_testing::MockSubscriptionBatchSource>();
>   EXPECT_CALL(*mock, Shutdown).Times(1);
> ```

FYI `Times(1)` is the default, unless you want to be explicit (occurs elsewhere)

coryan (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @mr-salty)

___
*[google/cloud/pubsub/internal/subscription_message_queue.h, line 41 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MK_mxSG1AFC49QE1gvd:-MKa_ORsE1MlgTXw0CIT:bt2vr1b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/29380c3ccc908d970c135b999eaf3c4d8846e51f/google/cloud/pubsub/internal/subscription_message_queue.h#L41)):*

> ```
> setups
> ```

s/setups/sets up/
</blockquote></details>

Fixed.
___
*[google/cloud/pubsub/internal/subscription_message_queue.cc, line 116 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5327#-MKMQP_m7WNiowOf5LWo:-MKa_ciSA3o1zWhC65Xx:bs7wumb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6687c130fdbbb31910fe450439d7c0ab6666a3f3/google/cloud/pubsub/internal/subscription_message_queue.cc#L116)):*

I still think explicit `lk.unlock()` is clearer.

The other part of my suggestion is still possible, although you'd need to do 2 swaps - completely up to you.
</blockquote></details>

Done.

PR #5329, opened 2020-10-23
cleanup(pubsub): no pure virtuals in `*Connection`

### LINKED ISSUE #4817: Change all *Connection() member functions to non-pure virtuals
We are using pure virtual functions as an aid in development, once the APIs are stable we should change them to non-pure virtual functions, probably returning `Status(StatusCode::kUnimplemented, "....")`.

We do not want pure virtual functions because after GA adding a pure virtual function is a breaking change.

coryan:
We expect applications to derive (typically mocks) from the
`*Connection` classes. To prevent breaking such applications in the
future these classes should not use pure-virtual functions, they should
return `kUnimplemented`.

Fixes #4817

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5329)

### DIFF
--- a/google/cloud/internal/pagination_range.h
+++ b/google/cloud/internal/pagination_range.h
@@ -216,6 +216,22 @@ class PaginationRange {
   bool on_last_page_;
 };
 
+template <typename T>
+struct UnimplementedPaginationRange {};
+
+template <typename T, typename Request, typename Response>
+struct UnimplementedPaginationRange<PaginationRange<T, Request, Response>> {
+  static PaginationRange<T, Request, Response> Create() {
+    return PaginationRange<T, Request, Response>(
+        Request{},
+        [](Request const&) {
+          return StatusOr<Response>(
+              Status{StatusCode::kUnimplemented, "needs-override"});
+        },
+        [](Response) { return std::vector<T>{}; });
+  }
+};
+
 }  // namespace internal
 }  // namespace GOOGLE_CLOUD_CPP_NS
 }  // namespace cloud
--- a/google/cloud/internal/pagination_range_test.cc
+++ b/google/cloud/internal/pagination_range_test.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/internal/pagination_range.h"
+#include "google/cloud/testing_util/status_matchers.h"
 #include <google/bigtable/admin/v2/bigtable_instance_admin.grpc.pb.h>
 #include <gmock/gmock.h>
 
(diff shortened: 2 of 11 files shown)


mr-salty (changes requested):
IIRC we discussed this issue for spanner and came to the opposite conclusion - in any case [`spanner::Connection`](https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/spanner/connection.h) is all pure virtuals).

It's worth revisiting now that we have evidence it's causing pain but I think we should discuss it with the team and apply whatever we decide to all the libraries (maybe it affects generated libraries as well?)

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5329#-:-MKMefLJDf6MrjDgBYyr:b-hhl3z5)* status: 0 of 11 files reviewed, all discussions resolved

coryan:
I thought we had discussed this already and decided to do kUnimplemented. @devjgm because I cannot find a link to the discussion.

devjgm:
I think moving away from pure-virtuals and toward returning `kUnimplemented` is probably a good idea. Because future methods that we add will not be able to be pure, and we started with pure virtuals then we'd end up w/ `Connection` classes that have a seemingly arbitrary mix of pure and impure virtual methods. To help the interface end up a more consistent eventual state, it seems reasonable to start w/ impure virtual methods in public interfaces that we expect customers to subclass.

I think we discussed this and came to this conclusion after we'd shipped the Spanner library.

devjgm (approved):
This change LGTM and matches my recollection of our discussion about this that we had _after_ we launched Spanner.

mr-salty (commented):
[spanner#305](https://github.com/googleapis/google-cloud-cpp-spanner/issues/305) is where the decision was made for spanner initially. There's not much rationale given for pure virtual.

[spanner#1358](https://github.com/googleapis/google-cloud-cpp-spanner/pull/1358#issuecomment-598201751) looks like where Greg suggested changing this. I think it probably makes sense; the [gmock cookbook](https://github.com/google/googletest/blob/master/googlemock/docs/cook_book.md#delegating-calls-to-a-parent-class) recommends pure virtual but it seems like their concern is needing to call the base class implementation which is not an issue here. 

so, I think I agree, and we should change spanner (and others) as well. it might be a breaking change in some super literal sense (code that wouldn't compile before will now compile?) but practically not really?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5329#-:-MK_bNGeBYLf5FnG--8F:b-hlq3ku)* status: 0 of 11 files reviewed, all discussions resolved

mr-salty (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5329#-:-MK_eY1R4LZfV6U0jAGz:bnfp4nl)* status: 0 of 11 files reviewed, all discussions resolved

Issue #4565, opened 2020-07-17
Implement a throughput benchmark


coryan:
We want a test to measure throughput for both publishing and receiving messages.
We also want to measure the "overhead" of either operation when compared to "raw" gRPC.

coryan:
https://cloud.google.com/blog/products/data-analytics/testing-cloud-pubsub-clients-to-maximize-streaming-performance

PR #5333, opened 2020-10-25
refactor: helpers to parse benchmark arguments

coryan:
Remove code duplication between the Storage and Pub/Sub benchmarks. Even
I don't want to cut&paste this code one more time. Did some minor
simplifications, like using `absl::DurationParse` instead of a custom
parser.

Part of the work for #4565

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5333)


google/cloud/testing_util/command_line_parsing.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * The size must be small enough for an in-memory buffer.
+ */
+std::size_t ParseBufferSize(std::string const& val);
+
+/// Parse a string as a duration with support for hours (h), minutes (m), or
+/// second (s) suffixes.
+std::chrono::seconds ParseDuration(std::string const& val);
+
+/// Parse a string as a boolean, with a default value if the string is empty.
```
devbww, line 54: I'm not sure "a default value" is the best term for "the not-present value".  Either way, it is also the result if the boolean fails to parse, whether or not the string is empty.

google/cloud/testing_util/command_line_parsing.cc
```diff
@@ -0,0 +1,161 @@
...
+  if (!absl::ParseDuration(val, &d)) {
+    internal::ThrowInvalidArgument("invalid duration: " + val);
+  }
+  return absl::ToChronoSeconds(d);
+}
+
+absl::optional<bool> ParseBoolean(std::string const& val) {
+  if (val.empty()) {
+    return {};
+  }
```
devbww, line 64: I would suggest omitting this special/redundant case.

google/cloud/testing_util/command_line_parsing.cc
```diff
@@ -0,0 +1,161 @@
...
+  std::ostringstream os;
+  os.setf(std::ios::fixed);
+  os.precision(1);
+  os << (static_cast<double>(size) / static_cast<double>(resolution)) << name;
+  return os.str();
+}
+
+std::string Basename(std::string const& path) {
+  // With C++17 we would use `std::filesystem::path`, until then do the poor's
+  // person version.
```
devbww, line 102: s/poor's person/poor person's/

Although, I might suggest "until then use `find_last_of()`".

devbww (approved):

google/cloud/testing_util/command_line_parsing.cc
coryan: Done.

google/cloud/testing_util/command_line_parsing.cc
coryan: Done.

PR #5335, opened 2020-10-26
chore: update API baselines

coryan:
I think we should update these more often, or we might
introduce and then remove APIs without noticing.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5335)


devbww (approved):
> I think we should update these more often, or we might
> introduce and then remove APIs without noticing.

Perhaps introducing an API should count enough for "check-abi" to fail, but not enough to require a "BREAKING CHANGE" announcement.

PR #5343, opened 2020-10-27
doc: update the API baseline as part of releasing

coryan:
This is an item for discussion.  We may chose to *not* do this.

I think we should update the API baseline before each release.
The purpose of this baseline is to tell us if we are breaking the
API since the last release.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5343)


devjgm (approved):
After discussing offline, I think this proposal looks good.

PR #5347, opened 2020-10-27
chore: update GCS V4 signed URLs conformance tests

### LINKED ISSUE #5073: Update signed URL conformance tests
There is a new version of the signed URL conformance tests, we need to update our copy and (maybe) fix the code.

dopiera:
This fixes #5073.

Turned out we were escaping too much. Json is already escaping the
backslashes for us, so no need to do it on our own. It was verified
againts production, too.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5347)


coryan (approved):
please wait for the pending builds

PR #5365, opened 2020-10-29
feat(pubsub): more efficient message builder

coryan:
Avoids allocations and data copying when the application
just adds a few attributes inline.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5365)

### DIFF
--- a/google/cloud/pubsub/message.h
+++ b/google/cloud/pubsub/message.h
@@ -165,21 +165,30 @@ class MessageBuilder {
     return std::move(SetOrderingKey(std::move(key)));
   }
 
-  /// Add an attribute to the message.
-  template <typename K, typename V>
-  MessageBuilder& AddAttribute(K&& key, V&& value) & {
+  /// Insert an attribute to the message if it does not have the @p key.
+  MessageBuilder& InsertAttribute(std::string const& key,
+                                  std::string const& value) & {
     using value_type =
         google::protobuf::Map<std::string, std::string>::value_type;
-    proto_.mutable_attributes()->insert(
-        value_type{std::forward<K>(key), std::forward<V>(value)});
+    proto_.mutable_attributes()->insert(value_type{key, value});
     return *this;
   }
 
-  /// Add an attribute to the message.
-  template <typename K, typename V>
-  MessageBuilder&& AddAttribute(K&& key, V&& value) && {
-    return std::move(
-        AddAttribute(std::forward<K>(key), std::forward<V>(value)));
+  /// Insert an attribute to the message if it does not have the @p key.
+  MessageBuilder&& InsertAttribute(std::string const& key,
+                                   std::string const& value) && {
+    return std::move(InsertAttribute(key, value));
+  }
+
+  /// Insert or set an attribute on the message.
+  MessageBuilder& SetAttribute(std::string const& key, std::string value) & {
+    (*proto_.mutable_attributes())[key] = std::move(value);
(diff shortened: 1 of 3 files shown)


mr-salty (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5365#-:-MKp7s4u7pyigIy7HUMp:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[google/cloud/pubsub/message.h, line 168 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5365#-MKp7jUf-p6QkQLY2MtP:-MKp7jUf-p6QkQLY2MtQ:b9f67rk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed6ef26add0bf3a127203747d4028125363805ec/google/cloud/pubsub/message.h#L168)):*
> ```objc
>   }
> 
>   /// Add an attribute to the message.
> ```

do these need to be templates or could they just take `std::string&&`?

coryan (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5365)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mr-salty)

___
*[google/cloud/pubsub/message.h, line 168 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5365#-MKp7jUf-p6QkQLY2MtP:-MKpB4_U162gffj-Gj1O:b-8ck3ck) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed6ef26add0bf3a127203747d4028125363805ec/google/cloud/pubsub/message.h#L168)):*

do these need to be templates or could they just take `std::string&&`?
</blockquote></details>

You are right, this is overkill, specially considering `MapPair<>` only has constructors from `std::string const&`:

https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.map#MapPair

I could argue that this is future-proof, but that seems very unlikely. Changed to use plain `std::string` (with the right qualifiers).

mr-salty (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5365#-:-MKpQvHx-ZvAy9xsbSgd:bnfp4nl)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @coryan and @mr-salty)

___
*[google/cloud/pubsub/message.h, line 168 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5365#-MKp7jUf-p6QkQLY2MtP:-MKpQU2PCVYkkaS8D69h:b-jkk03b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed6ef26add0bf3a127203747d4028125363805ec/google/cloud/pubsub/message.h#L168)):*

You are right, this is overkill, specially considering `MapPair<>` only has constructors from `std::string const&`:

https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.map#MapPair

I could argue that this is future-proof, but that seems very unlikely. Changed to use plain `std::string` (with the right qualifiers).


</blockquote></details>

yeah, unfortunate there's not a `&&` `MapPair` constructor... I guess you could leave these parameters as `&&` for future-proofing, and maybe we can talk the proto folks into adding it.

PR #5384, opened 2020-10-31
test(storage): add resumable and copy/compose

vnvo2409:
This PR adds:
- support for resumable upload
- copy object
- compose object
- xml
- streaming data

Part of #4751 

I have some questions:

- It seems that there is a mismatch between the testbench and production. The production states that default projection of patching an object is [NO_ACL](https://cloud.google.com/storage/docs/json_api/v1/objects/patch#parameters). While the integration test assumes that it is FULL. 

https://github.com/googleapis/google-cloud-cpp/blob/7d181694a08d385b8ace39a37e190052a3afb614/google/cloud/storage/tests/object_integration_test.cc#L102-L109

- I could not understand this test. We use `CustomHeader("x-goog-testbench-instructions", "return-503-after-256K")`. So this test will fail when `offset >= 262144`. But I don't see any sign indicates that it will happen.
https://github.com/googleapis/google-cloud-cpp/blob/7d181694a08d385b8ace39a37e190052a3afb614/google/cloud/storage/tests/object_media_integration_test.cc#L713-L746

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5384)


coryan (changes requested):
> I have some questions:
> 
> It seems that there is a mismatch between the testbench and production. The production states that default projection of patching an object is NO_ACL. While the integration test assumes that it is FULL.

It seems the documentation is incomplete. If you patch the ACL then the response includes the ACL too.

> 
> 
> I could not understand this test. We use CustomHeader("x-goog-testbench-instructions", "return-503-after-256K"). So this test will fail when offset >= 262144. But I don't see any sign indicates that it will happen.

The intent is for the emulator to generate an error after 256KiB.  The client library should handle that error by resuming the download AS-IF the error had not happened.  Does that make sense?

The test is weak, in that it passes even if the emulator does not create the expected error, but it was useful when we fixed whatever bug motivated it.

Some questions and suggestions below, take them with a grain of salt, I might be wrong about them.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5384#-:-MKznhAR41V_5GrfykBY:b2eqjs1)* status: all files reviewed, 2 unresolved discussions (waiting on @vnvo2409)

___
*[google/cloud/storage/emulator/database.py, line 65 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5384#-MKzr0Dg4ysIWtWxVazV:-MKzr0DhA_p3Qt7xM_OB:b-p4urlc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/113f9dc2b81cbdaf0db9de501871074212bebf9d/google/cloud/storage/emulator/database.py#L65)):*
> ```Python
>         return self.buckets.values()
> 
>     def delete_bucket(self, request, bucket_name, context):
> ```

Is there a check to prevent deleting non-empty buckets?  If not, maybe add a bug to add that feature later?
___
*[google/cloud/storage/emulator/gcs/holder.py, line 15 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5384#-MKzrTKX--YlBqAgvH-q:-MKzrTKX--YlBqAgvH-r:bgq6af6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/113f9dc2b81cbdaf0db9de501871074212bebf9d/google/cloud/storage/emulator/gcs/holder.py#L15)):*
> ```Python
> # limitations under the License.
> 
> """Implement a class as a placeholder for data."""
> ```

Maybe "Implement a holder for resumable upload's data"?

vnvo2409:
> Is there a check to prevent deleting non-empty buckets?

Done

> Maybe "Implement a holder for resumable upload's data"?

Done

> The intent is for the emulator to generate an error after 256KiB. The client library should handle that error by resuming the download AS-IF the error had not happened. Does that make sense?

It turns out that this test didn't work with bare Flask ( It failed with no retry at all ). In order to get the emulator working with `gunicorn`, I use a pattern called `application factory` which allow us to initialization the database and start the `gRPC` server.
```python
# gunicorn -b localhost:9000 "rest_server:run()" --access-logfile - --worker-class gevent
def run():
     global db
     logging.basicConfig()
     db = database.Database.init()
     # grpc_port = int(os.getenv("GOOGLE_CLOUD_CPP_STORAGE_EMULATOR_GRPC_PORT", 8000))
     # grpc_server.run(grpc_port, db)
     return server
```
I haven't added any `gRPC` code yet so I add some comments to demonstrate how it works with `gRPC`. In addition, `emulator.py` isn't needed any more. I will rename `rest_server.py` to `emulator.py` in the next PR.

vnvo2409:
Regarding `error_injection_integration_test`, How can I test it with `bazel` ? I got an error `Aborting because exceptions are disabled: send() has not been called yet.`. But when I pass `--copt="-DGOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS"` to `bazel`, the test failed to build.

coryan:
> Regarding `error_injection_integration_test`, How can I test it with `bazel` ? I got an error `Aborting because exceptions are disabled: send() has not been called yet.`. But when I pass `--copt="-DGOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS"` to `bazel`, the test failed to build.

The exceptions stuff is plain wrong.  `Terminate()` is adding that to this:

https://github.com/googleapis/google-cloud-cpp/blob/7d181694a08d385b8ace39a37e190052a3afb614/google/cloud/storage/tests/error_injection_integration_test.cc#L76

The test passes for me (see below).  Are you using any options for bazel?  Maybe they incidentally disable dynamic loading, which this test depends on.  We may need a separate bug to track this down.

```
$ bazel --version
bazel 3.7.0
coryan@corvm:~/cpp-develop$ gcc --version

$ gcc (Debian 10.2.0-9) 10.2.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ grep _VERSION_ google/cloud/internal/version_info.h
#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_VERSION_INFO_H
#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_VERSION_INFO_H
#define GOOGLE_CLOUD_CPP_VERSION_MAJOR 1
#define GOOGLE_CLOUD_CPP_VERSION_MINOR 20
#define GOOGLE_CLOUD_CPP_VERSION_PATCH 0
#endif  // GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_VERSION_INFO_H

$ git rev-parse HEAD
7d181694a08d385b8ace39a37e190052a3afb614

$ bazel test //google/cloud/storage/tests:error_injection_integration_test
INFO: Analyzed target //google/cloud/storage/tests:error_injection_integration_test (62 packages loaded, 3729 targets configured).
INFO: Found 1 test target...
INFO: From Generating C++ proto_library @com_google_googleapis//google/iam/v1:options_proto:
google/iam/v1/options.proto:20:1: warning: Import google/api/annotations.proto is unused.
INFO: From Generating C++ proto_library @com_google_googleapis//google/iam/v1:policy_proto:
google/iam/v1/policy.proto:21:1: warning: Import google/api/annotations.proto is unused.
INFO: From Action external/com_google_googleapis/google/storage/v1/storage.grpc.pb.h:
bazel-out/k8-fastbuild/bin/external/com_google_googleapis/external/com_google_googleapis: warning: directory does not exist.
INFO: From Compiling com_github_grpc_grpc/src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc:
external/com_github_grpc_grpc/src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc: In static member function 'static std::vector<grpc_error*> grpc_core::{anonymous}::WeightedTargetLbFactory::ParseChildConfig(const grpc_core::Json&, grpc_core::{anonymous}::WeightedTargetLbConfig::ChildConfig*)':
external/com_github_grpc_grpc/src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc:681:32: warning: comparison of integer expressions of different signedness: 'uint32_t' {aka 'unsigned int'} and 'int' [-Wsign-compare]
  681 |       if (child_config->weight == -1) {
      |           ~~~~~~~~~~~~~~~~~~~~~^~~~~
Target //google/cloud/storage/tests:error_injection_integration_test up-to-date:
  /usr/local/google/home/coryan/.cache/bazel/_bazel_coryan/e2b708c018cdf4ea6e5d73c8bc1aeca0/execroot/com_github_googleapis_google_cloud_cpp/bazel-out/k8-fastbuild/bin/google/cloud/storage/tests/error_injection_integration_test
INFO: Elapsed time: 157.978s, Critical Path: 135.48s
INFO: 1612 processes: 235 internal, 1377 linux-sandbox.
INFO: Build completed successfully, 1612 total actions
//google/cloud/storage/tests:error_injection_integration_test            PASSED in 0.4s

Executed 1 out of 1 test: 1 test passes.
```

coryan (approved):
:lgtm:

> Reviewed 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5384#-:-ML3HPBS4tQlZRkF_Iam:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #4893, opened 2020-08-13
Support customTime for GCS+gRPC


coryan:
I am about to add support for the `customTime` field in `gcs::ObjectMetadata`, the protos do not have it yet, so I am punting on the gRPC implementation.

vnvo2409:
How is the gRPC implementation going ? Since I currently use `protos` as the backend for the testbench, the conversion of `REST` request into `protos` will automatically remove the `customTime` field ( because `protos` does not have that field yet )

coryan:
The gRPC implementation continues to be focused on data operations, I expect these differences will be fixed once the team refocuses on metadata operations.

PR #5392, opened 2020-11-02
test(storage): add emulator rewrite, iam, project

vnvo2409:
This PR adds:
- Object rewrite
- GCS IAM ( copied from the old testbench )
- GCS Project ( copied from the old testbench )
- Temporary workaround for `customTime`
- Fix the behavior of `delete_object`. It should only delete the live version and keep the noncurrent versions. When deleting a bucket, we only check if there is any live version.

I would like to ask some questions:

- Why this sample uses `PatchBucketAcl` instead of `CreateBucketAcl`
https://github.com/googleapis/google-cloud-cpp/blob/7557959a25c7d94070a11ce90bf0a3140ef359b1/google/cloud/storage/examples/storage_bucket_acl_samples.cc#L166-L184

- In addition, I run `clang-tidy` with the new emulator and all the tests passed. Do you prefer a PR to turn the integrations test to the new emulator before of after adding `gRPC` server ( since the `gRPC` tests are running against production now , it won't affect anything )

### DIFF
--- a/google/cloud/storage/emulator/database.py
+++ b/google/cloud/storage/emulator/database.py
@@ -20,15 +20,16 @@ from google.cloud.storage_v1.proto import storage_pb2 as storage_pb2
 
 
 class Database:
-    def __init__(self, buckets, objects, live_generations, uploads):
+    def __init__(self, buckets, objects, live_generations, uploads, rewrites):
         self.buckets = buckets
         self.objects = objects
         self.live_generations = live_generations
         self.uploads = uploads
+        self.rewrites = rewrites
 
     @classmethod
     def init(cls):
-        return cls({}, {}, {}, {})
+        return cls({}, {}, {}, {}, {})
 
     # === BUCKET === #
 
@@ -64,7 +65,7 @@ class Database:
 
     def delete_bucket(self, request, bucket_name, context):
         bucket = self.get_bucket(request, bucket_name, context)
-        if len(self.objects[bucket.metadata.name]) > 0:
+        if len(self.live_generations[bucket.metadata.name]) > 0:
             utils.error.invalid("Deleting non-empty bucket", context)
         del self.buckets[bucket.metadata.name]
         del self.objects[bucket.metadata.name]
@@ -120,6 +121,7 @@ class Database:
             end_offset,
         ) = self.__extract_list_object_request(request, context)
         items = []
+        customTimes = []
         prefixes = set()
         for obj in bucket.values():
             generation = obj.metadata.generation
@@ -139,7 +141,8 @@ class Database:
                 prefixes.add(name[: delimiter_index + 1])
                 continue
             items.append(obj.metadata)
-        return items, list(prefixes)
+            customTimes.append(obj.customTime)
+        return items, list(prefixes), customTimes
 
     def check_object_generation(
         self, request, bucket_name, object_name, is_source, context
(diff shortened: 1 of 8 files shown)


coryan (changes requested):
> Why this sample uses PatchBucketAcl instead of CreateBucketAcl

Probably because the analogous examples in other languages also used `Patch`.

> 
> In addition, I run clang-tidy with the new emulator and all the tests passed.

Yay!

> Do you prefer a PR to turn the integrations test to the new emulator before of after adding gRPC server ( since the gRPC tests are running against production now , it won't affect anything )

Before adding the gRPC server sounds good.

A couple of questions of my own below.

+kokoro:run 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5392#-:-ML9B8nVDbvSmBegAh1M:b-b00np9)* status: all files reviewed, 2 unresolved discussions (waiting on @vnvo2409)

___
*[google/cloud/storage/emulator/gcs/bucket.py, line 362 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5392#-ML9Btcj52Vz6sQ2CqrD:-ML9Btck1VOz1ucoGbuU:bdhzg5f) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/afdd6d7f15efeda936bb5c6b9406760551956734/google/cloud/storage/emulator/gcs/bucket.py#L362)):*
> ```Python
>             payload = simdjson.loads(request.data)
>             role = payload["role"]
>         return self.__upsert_acl(entity, role, False, context)
> ```

Is the `update_only` parameter ever `True` anymore?
___
*[google/cloud/storage/emulator/gcs/object.py, line 75 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5392#-ML9C4_UBIR3924GDR-M:-ML9C4_UBIR3924GDR-N:bmt11xz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/afdd6d7f15efeda936bb5c6b9406760551956734/google/cloud/storage/emulator/gcs/object.py#L75)):*
> ```Python
>         metadata.acl.extend(acls)
> 
>     # TODO(#4893): Remove `customTime`
> ```

Is a separate `customTime` needed because the protos do not have it yet?

If so, consider adding a dictionary.  This is not the last additional field we will have in the REST API that the gRPC protos lag for a bit.

vnvo2409:
> Is the `update_only` parameter ever `True` anymore?

It should be `True` when updating the `ACL` but the `google-cloud-cpp/google/cloud/storage/examples/storage_bucket_acl_samples.cc` sample failed because the `entity` does not exists yet. So I am quite confused about it.

coryan (commented):
> So I am quite confused about it.

Me too.

> It should be True when updating the ACL but the google-cloud-cpp/google/cloud/storage/examples/storage_bucket_acl_samples.cc sample failed because the entity does not exists yet. 

It seems that both "PATCH" and UPDATE allow inserting new entities.  I will ask my colleagues, but I suspect that whatever the answer is (a) we need to implement the current behavior in the emulator, and (b) there behavior is not going to change, too many customers depend on it for better or worse.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5392#-:-MLADZt091dv0RMAPWs-:brhwjk2)* status: all files reviewed, 3 unresolved discussions (waiting on @vnvo2409)

___
*[google/cloud/storage/emulator/gcs/object.py, line 358 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5392#-ML9_z1K1wXQjgNsFJib:-ML9_z1M974Znn4bdaN-:b8sb6oq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/afdd6d7f15efeda936bb5c6b9406760551956734/google/cloud/storage/emulator/gcs/object.py#L358)):*
> ```Python
>         del self.metadata.acl[self.__search_acl(entity, True, context)]
> 
>     # === RESPOSE === #
> ```

nit: s/RESPOSE/RESPONSE/

coryan:
Did you mean to include the `simdjson` changes and `rest_server` -> `emulator` changes in this PR?

vnvo2409:
Sum up what I have pushed.

55afd33 : Add a dictionary `rest_only` to save the unsupported REST metadata and remove the `update_only` parameter.
838a1da : I used `simdjson` because it was fast. However, It doesn't support `python 3.9` and since our request is quite small, the performance gain isn't significant. In the end, I replace it with the builtin `json`.
6b51d53 : Sort the `import` so it looks cleaner.
6c66ad1 : It is a technique from the old testbench indeed.
e69a905 : And lastly, rename the `rest_server.py` to `emulator.py`

And all the tests in the last kokoro run passed.

coryan:
Ack. Sounds good.

coryan (approved):
:lgtm:

> Reviewed 14 of 14 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5392#-:-MLDVLUGDYzLqyP24DyK:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #4010, opened 2020-05-05
[Flake] Integration tests only retry idempotent operations


coryan:
The client libraries, by design, only retry idempotent operations. In the integration tests this can be a problem, as a transient failure breaks the test and the build.  I think this is what happened here:

https://source.cloud.google.com/results/invocations/39ae0dad-4bf4-4efc-aa4e-35e8930cee71/targets

```
FAIL: //google/cloud/bigtable/examples:table_admin_iam_policy_snippets (see /Volumes/BuildData/tmpfs/tmp/bazel/execroot/com_github_googleapis_google_cloud_cpp/bazel-out/darwin-fastbuild/testlogs/google/cloud/bigtable/examples/table_admin_iam_policy_snippets/test.log)
INFO: From Testing //google/cloud/bigtable/examples:table_admin_iam_policy_snippets:
==================== Test output for //google/cloud/bigtable/examples:table_admin_iam_policy_snippets:

Cleaning up old tables

Running GetIamPolicy() example
The IAM Policy for table-admin-snippets-1588544661-irhphmoz is
etag: "\000 \001"


Running SetIamPolicy() example
The IAM Policy for table-admin-snippets-1588544661-irhphmoz is
version: 1
etag: "\007\005\244\305\340x\227\n"
bindings {
  role: "roles/bigtable.user"
  members: "serviceAccount:bigtable-test-iam-sa@cloud-cpp-testing-resources.iam.gserviceaccount.com"
}


Running TestIamPermissions() example
The current user has the following permissions [bigtable.tables.get]

Running AsyncGetIamPolicy() example
IamPolicy details : version: 1
etag: "\007\005\244\305\340x\227\n"
bindings {
  role: "roles/bigtable.user"
  members: "serviceAccount:bigtable-test-iam-sa@cloud-cpp-testing-resources.iam.gserviceaccount.com"
}


Running AsyncSetIamPolicy() example
Waiting for IAM policy update to complete .Standard C++ exception raised: AsyncSetIamPolicynon-idempotent operation failed, last error=Error while setting IAM Policy for projects/cloud-cpp-testing-resources/instances/test-instance/tables/table-admin-snippets-1588544661-irhphmoz
================================================================================
```

Though the error message is not super clear.

coryan:
Here is another one:

https://source.cloud.google.com/results/invocations/dac65eb9-ed4b-4575-827a-8678c2d010e4/targets

devbww:
[...]

PR #5395, opened 2020-11-02
test(spanner): deflake benchmark database creation

devbww:
Retry CreateDatabase() calls during benchmark setup that fail with
UNAVAILABLE. Up to three retries, with backoff intervals of 0s, 1s,
and 2s respectively.

Part of #4010.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5395)

### DIFF
--- a/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
+++ b/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
@@ -188,7 +188,7 @@ int main(int argc, char* argv[]) {
     db = create_future.get();
     if (db) break;
     if (db.status().code() != google::cloud::StatusCode::kUnavailable) break;
-    std::this_thread::sleep_for(retry * std::chrono::seconds(1));
+    std::this_thread::sleep_for(retry * std::chrono::seconds(3));
   }
   std::cout << " DONE\n";
 
--- a/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
+++ b/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
@@ -173,18 +173,26 @@ int main(int argc, char* argv[]) {
     }
     return statements;
   }();
-  auto create_future =
-      admin_client.CreateDatabase(database, additional_statements);
+
   std::cout << "# Waiting for database creation to complete " << std::flush;
-  for (;;) {
-    auto status = create_future.wait_for(std::chrono::seconds(1));
-    if (status == std::future_status::ready) break;
-    std::cout << '.' << std::flush;
+  google::cloud::StatusOr<google::spanner::admin::database::v1::Database> db;
+  int constexpr kMaxCreateDatabaseRetries = 3;
+  for (int retry = 0; retry <= kMaxCreateDatabaseRetries; ++retry) {
+    auto create_future =
+        admin_client.CreateDatabase(database, additional_statements);
+    for (;;) {
+      auto status = create_future.wait_for(std::chrono::seconds(1));
+      if (status == std::future_status::ready) break;
+      std::cout << '.' << std::flush;
+    }
+    db = create_future.get();
+    if (db) break;
+    if (db.status().code() != google::cloud::StatusCode::kUnavailable) break;
+    std::this_thread::sleep_for(retry * std::chrono::seconds(1));
   }
   std::cout << " DONE\n";
 
   bool database_created = true;
-  auto db = create_future.get();
   if (!db) {
     if (user_specified_database &&
         db.status().code() == google::cloud::StatusCode::kAlreadyExists) {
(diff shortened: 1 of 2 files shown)


coryan (approved):
:lgtm: I am not sure if this is convincing, but food for thought:

Why do we treat `CreateDatabase()` as non-idempotent:

https://github.com/googleapis/google-cloud-cpp/blob/7557959a25c7d94070a11ce90bf0a3140ef359b1/google/cloud/spanner/database_admin_connection.cc#L156-L161


The operation can only succeed once, therefore "the system state remains the same after one or several calls":

https://en.wikipedia.org/wiki/Idempotence#Computer_science_meaning

modulo race conditions with `DeleteDatabase`, but that should be out of scope or nothing is idempotent. So maybe we should let the library retry?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5395#-:-MLAA1ZA-B-nM5PzJbhy:b-g7tgiu)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc, line 190 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5395#-MLA92sq0uMzXt6mzdqS:-MLA92sr42DtTmGvDzVT:b-d1ujqq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce91c3dc13bff288282dc3d3c22a1aeb9f75685f/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc#L190)):*
> ```cpp
>     db = create_future.get();
>     if (db) break;
>     if (db.status().code() != google::cloud::StatusCode::kUnavailable) break;
> ```

FYI: if the first request sneaks through and then the socket drops (bizarre, but could happen), then the second retry would fail with `kAlreadyExist`.   That would be a flake, but a far less common flake than what we have today.  Maybe a comment?  Up to you.
___
*[google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc, line 191 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5395#-MLA9eFvBMx_bEeUcrF_:-MLA9eFvBMx_bEeUcrFa:b8d43ju) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce91c3dc13bff288282dc3d3c22a1aeb9f75685f/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc#L191)):*
> ```cpp
>     if (db) break;
>     if (db.status().code() != google::cloud::StatusCode::kUnavailable) break;
>     std::this_thread::sleep_for(retry * std::chrono::seconds(1));
> ```

FYI also: that backoff might be a tad optimistic.

devbww (commented):
> Why do we treat `CreateDatabase()` as non-idempotent:

I would say as long as we expose the difference between `OK` and `ALREADY_EXISTS` to the user that `CreateDatabase()` is non-idempotent.  It would be weird, for example, for a non-racing `DeleteDatabase(D) && CreateDatabase(D)` to return `ALREADY_EXISTS`, as might happen with a simple retry.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5395#-:-MLDkUMd634CqRwyapmH:b2fplqz)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @coryan)

___
*[google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc, line 190 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5395#-MLA92sq0uMzXt6mzdqS:-MLDmbunCNgGiUzYWpdG:bwc3w1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce91c3dc13bff288282dc3d3c22a1aeb9f75685f/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc#L190)):*

FYI: if the first request sneaks through and then the socket drops (bizarre, but could happen), then the second retry would fail with `kAlreadyExist`.   That would be a flake, but a far less common flake than what we have today.  Maybe a comment?  Up to you.
</blockquote></details>

I don't think we need say anything special.  We could also get `kAlreadyExist` simply because the database does already exist, which we do handle in the `user_specified_database` case below.
___
*[google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc, line 191 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5395#-MLA9eFvBMx_bEeUcrF_:-MLDn1sG2FM__r4c9XT4:blkis25) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce91c3dc13bff288282dc3d3c22a1aeb9f75685f/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc#L191)):*

FYI also: that backoff might be a tad optimistic.
</blockquote></details>

Thanks.  Tripled it to multiples of 3 seconds, so the overall extra delay for 3 retries will be 9 seconds.  Much beyond that, I imagine, we'd prefer to fail the test.

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5395#-:-MLE8vDA9cg0hWDxmkp7:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5397, opened 2020-11-03
refactor(generator): service generator access refactor

scotthart:
In order to ensure that certain member variables (raw pointers, maps) of the *Generator classes are not mutated after construction, ServiceGenerator has had protected accessor member functions added and all member variables made private. This has had the side benefit of making the call sites more concise and a modest readability improvement.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5397)


generator/internal/service_code_generator.cc
```diff
@@ -51,7 +57,30 @@ ServiceCodeGenerator::ServiceCodeGenerator(
...
+  for (int i = 0; i < service_descriptor_->method_count(); ++i) {
+    methods_.emplace_back(*service_descriptor_->method(i));
+  }
+}
+
+VarsDictionary const& ServiceCodeGenerator::vars() const {
+  return service_vars_;
+}
+
+std::string ServiceCodeGenerator::vars(std::string const& key) const {
```
devjgm, line 72: Nit: it seems like this method should probably be named `var` since it's returning the value for a single var. But I'll leave it up to you.

devjgm (approved):

PR #5398, opened 2020-11-03
test(storage): use new emulator for storage integration test

vnvo2409:
Part of #4751

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5398)

### DIFF
--- a/ci/kokoro/docker/Dockerfile.fedora
+++ b/ci/kokoro/docker/Dockerfile.fedora
@@ -30,12 +30,10 @@ RUN echo 'root:' | chpasswd
 
 # Install the Python modules needed to run the storage testbench
 RUN pip3 install --upgrade pip
-RUN pip3 install setuptools
-RUN pip3 install Jinja2==2.11.2 MarkupSafe==1.1.1 Werkzeug==1.0.1 \
-    blinker==1.4 brotlipy==0.7.0 cffi==1.14.3 crc32c==2.1 decorator==4.4.2 \
-    flask==1.1.2 gevent==20.9.0 greenlet==0.4.17 gunicorn==20.0.4 \
-    httpbin==0.7.0 itsdangerous==1.1.0 pycparser==2.20 raven==6.10.0 \
-    zope.event==4.5.0 zope.interface==5.1.0
+RUN pip3 install setuptools wheel
+RUN pip3 install git+git://github.com/googleapis/python-storage@gapic-generation
+RUN pip3 install grpcio==1.32.0 flask==1.1.2 pytest==6.0.2 httpbin==0.7.0 \
+    scalpl==0.4.0 crc32c==2.1 gevent==20.9.0 gunicorn==20.0.4
 
 # Install the Cloud SDK and some of the emulators. We use the emulators to run
 # integration tests for the client libraries.
--- a/ci/kokoro/docker/Dockerfile.fedora
+++ b/ci/kokoro/docker/Dockerfile.fedora
@@ -31,7 +31,7 @@ RUN echo 'root:' | chpasswd
 # Install the Python modules needed to run the storage testbench
 RUN pip3 install --upgrade pip
 RUN pip3 install setuptools wheel
-RUN pip3 install git+git://github.com/googleapis/python-storage@gapic-generation
+RUN pip3 install git+git://github.com/googleapis/python-storage@8cf6c62a96ba3fff7e5028d931231e28e5029f1c
 RUN pip3 install grpcio==1.32.0 flask==1.1.2 pytest==6.0.2 httpbin==0.7.0 \
     scalpl==0.4.0 crc32c==2.1 gevent==20.9.0 gunicorn==20.0.4
 
--- a/ci/kokoro/docker/Dockerfile.fedora-install
+++ b/ci/kokoro/docker/Dockerfile.fedora-install
@@ -55,12 +55,10 @@ RUN pip3 install black==19.3b0
 
 # Install the Python modules needed to run the storage testbench
 RUN dnf makecache && dnf install -y python3-devel
-RUN pip3 install setuptools
-RUN pip3 install Jinja2==2.11.2 MarkupSafe==1.1.1 Werkzeug==1.0.1 \
-    blinker==1.4 brotlipy==0.7.0 cffi==1.14.3 crc32c==2.1 decorator==4.4.2 \
-    flask==1.1.2 gevent==20.9.0 greenlet==0.4.17 gunicorn==20.0.4 \
-    httpbin==0.7.0 itsdangerous==1.1.0 pycparser==2.20 raven==6.10.0 \
-    zope.event==4.5.0 zope.interface==5.1.0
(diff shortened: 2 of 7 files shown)


vnvo2409:
@coryan Could you please kick off a CI run so I could see what happend ?

vnvo2409:
Almost tests passed so there is no error in the `Dockerfile`. The problem is all the integration tests passed with cached. So I could not know is there any error with the new testbench. Do you have any solution for this problem @coryan ?

coryan:
To disable the test cache we need to change one of the inputs that Bazel considers.  One approach would be to change this value:

https://github.com/googleapis/google-cloud-cpp/blob/b0020c1b380dda0c94f2adc608dceac8e8f80a1a/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh#L76

The value of the environment variable does not matter, but (I think) any different value will force Bazel to re-run the tests.  Another approach would be to pick a different port for the testbench:

https://github.com/googleapis/google-cloud-cpp/blob/b0020c1b380dda0c94f2adc608dceac8e8f80a1a/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh#L56

vnvo2409:
@coryan 
I think it should be good now

vnvo2409:
Everything should be good now.

coryan:
FWIW, I am running some of the CI builds locally with caching disabled.

coryan (changes requested):
Thanks for the PR, one question around pinning the releases.

We should also retake the discussion around running the Python tests, even if only for the Bazel builds.

> Reviewed 5 of 5 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5398#-:-MLGNhsl5v4maDHYEpS_:b-u58igi)* status: all files reviewed, 1 unresolved discussion (waiting on @vnvo2409)

___
*[ci/kokoro/docker/Dockerfile.fedora, line 34 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5398#-MLGNAWFDetmfUHMVTCn:-MLGNAWFDetmfUHMVTCo:b-zihk1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/cc9d9babbf1f0a19ea248793eef56acdc41e42e7/ci/kokoro/docker/Dockerfile.fedora#L34)):*
> ```
> RUN pip3 install --upgrade pip
> RUN pip3 install setuptools wheel
> RUN pip3 install git+git://github.com/googleapis/python-storage@gapic-generation
> ```

Here and in the other `Dockerfile.*` files: would it be possible to pin this to a specific SHA or tag instead of a branch?  Past experience tell us the builds can break unexpectedly when we try to follow the tip of a branch for any dependency.

Maybe @crwilcox can create a tag for us if we need something symbolic instead of a SHA?

vnvo2409:
> We should also retake the discussion around running the Python tests, even if only for the Bazel builds.

I have just done some experiments between `pytest` and `Bazel`. Here are the pros and cons:
Pros:
- `pytest`: It is super easy. `pip + python tests_entry.py`
- `bazel`: It could cache the tests result so we don't need to rerun the tests in every build.

Cons:
- `pytest`: Each run takes around 7s
- `bazel`: We will need to add a dependecy to our `WORKSPACE` https://github.com/bazelbuild/rules_python#getting-started

In addition, I think we could add the tests to `kokoro/docker/integration`.

> would it be possible to pin this to a specific SHA or tag instead of a branch? Maybe @crwilcox can create a tag for us if we need something symbolic instead of a SHA?

I prefer SHA, because we will need to update the proto files regularly until the upstream proto are GA or at least stable ( they are missing some REST features now ).

What do you think ?

coryan (commented):
> I have just done some experiments between pytest and Bazel. Here are the pros and cons: Pros:
> 
> pytest: It is super easy. pip + python tests_entry.py
> bazel: It could cache the tests result so we don't need to rerun the tests in every build.
> 
> Cons:
> 
> pytest: Each run takes around 7s
> bazel: We will need to add a dependecy to our WORKSPACE https://github.com/bazelbuild/rules_python#getting-started

The only reason I suggested Bazel is because it is already integrated into our build scripts. pytest is fine, but you will need to change the build scripts to invoke pytest, capture the results, stop the build, etc.

> In addition, I think we could add the tests to kokoro/docker/integration.

You get that for free if you put the tests in Bazel (or CMake, but CMake seems harder for this).  Up to you, we should discuss this in a separate PR / issue, so you can make progress here.

>> would it be possible to pin this to a specific SHA or tag instead of a branch? Maybe @crwilcox can create a tag for us if we need something symbolic instead of a SHA?
> 
> I prefer SHA, because we will need to update the proto files regularly until the upstream proto are GA or at least stable ( they are missing some REST features now ).
> What do you think ?

Sounds good, worst case we can change this later to a tag.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5398#-:-MLIUu9t6SMbsLd1JOfv:b-o8ybed)* status: :shipit: complete! all files reviewed, all discussions resolved

coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5398)* status: all files reviewed, 1 unresolved discussion (waiting on @vnvo2409)

vnvo2409:
> Up to you, we should discuss this in a separate PR / issue, so you can make progress here.

I will make a PR for the tests after finishing `grpc`

coryan (approved):
:lgtm: +kokoro:run 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5398#-:-MLIjOE3DUAoHVX9Mc6_:bcbhv2)* status: :shipit: complete! all files reviewed, all discussions resolved

crwilcox:
@vnvo2409 I could make a tag if helpful. The other thing to consider is if we could start including this surface in the client, but keep it separated from user surface. @frankyn to weigh in on that though, as it could be a while until we begin using that code in our python client library.

Also, I don't think we considered before, that it could even make sense to move the generation of that surface to alongside the emulator? Likely depends on the timeline of gRPC support in the python client.

vnvo2409:
> I could make a tag if helpful

IMO, a SHA or a tag does not matter much so I think we can use SHA now.

> Also, I don't think we considered before, that it could even make sense to move the generation of that surface to alongside the emulator.

I'm sorry. I don't understand what you mean here. Could you make it clearer please ? Thank you.

crwilcox:
Sure. We currently are relying on generated code in the client library repo, on a non-maintained branch, that isn't surface used by the client library. We expect that this surface will eventually be used, but we currently haven't started work on a gRPC surface for that client. We could move the generation process alongside the emulator if this is a longer term thing.

vnvo2409:
> We could move the generation process alongside the emulator if this is a longer term thing.

Thank you. It would be really useful, especially when `grpc` are missing some metadata properties compared with `rest`.

PR #5401, opened 2020-11-04
feat(pubsub): round robin publisher connections

coryan:
Allow applications to create multiple channels in a single publisher.
Currently there are other bottlenecks that makes this change a bit
redundant, but it is easier to merge ahead of more changes.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5401)


google/cloud/pubsub/publisher_connection.h
```diff
@@ -125,6 +125,13 @@ std::shared_ptr<pubsub::PublisherConnection> MakePublisherConnection(
     std::unique_ptr<pubsub::RetryPolicy const> retry_policy,
     std::unique_ptr<pubsub::BackoffPolicy const> backoff_policy);
 
+std::shared_ptr<pubsub::PublisherConnection> MakePublisherConnection(
+    pubsub::Topic topic, pubsub::PublisherOptions options,
+    pubsub::ConnectionOptions connection_options,
+    std::vector<std::shared_ptr<PublisherStub>> stubs,
```
devbww, line 131: Is throwing at the `at()` in `Child()` the intended response to an empty vector?  If so, perhaps add a comment that it should not be empty.  If not, it seems like `MakePublisherConnection()` should probably return `nullptr` in that circumstance.

google/cloud/pubsub/internal/publisher_round_robin_test.cc
```diff
@@ -0,0 +1,205 @@
...
+
+namespace google {
+namespace cloud {
+namespace pubsub_internal {
+inline namespace GOOGLE_CLOUD_CPP_PUBSUB_NS {
+namespace {
+
+using ::testing::Return;
+
+TEST(PublisherRoundRobinTest, CreateTopic) {
```
devbww, line 29: Opt nit: These indeed check that each mock is used once, but "round robin" is a slightly stronger condition.

coryan: Fair enough.  I am not sure if the new tests is what you had in mind, if not, let me what are you thinking about.

google/cloud/pubsub/publisher_connection.h
coryan: I am not sure `nullptr` is great, the result goes into the constructor of `Publisher` and `TopicAdminClient`.  On the other hand, largely it does not matter because this is in `internal::` and all the public interfaces create non-empty stub lists. Let me know what you think.

coryan:
PTAL

google/cloud/pubsub/internal/publisher_round_robin_test.cc
devbww: Yes, something like `InSequence` and a multiplier on the call loop was what I was thinking.  Thanks.

google/cloud/pubsub/publisher_connection.h
devbww: Sorry ... I missed that this was inside `pubsub_internal` despite being in a public header.

devbww (approved):

PR #5406, opened 2020-11-04
feat: a faster CompletionQueue::RunAsync

### LINKED ISSUE #5403: Improve performance for CompletionQueue::RunAsync
## Motivation

The current implementation of `CompletionQueue::RunAsync()` takes about 3.9us of CPU (5us) elapsed to handle a call, even with 16 cores:

```
Run on (96 X 2000.15 MHz CPU s)
CPU Caches:
  L1 Data 32K (x48)
  L1 Instruction 32K (x48)
  L2 Unified 1024K (x48)
  L3 Unified 39424K (x2)
Load Average: 3.64, 1.55, 0.90
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
BM_CompletionQueueRunAsync/16/64      1761023 ns       102206 ns        10000
BM_CompletionQueueRunAsync/16/128     1850689 ns       165496 ns        10000
BM_CompletionQueueRunAsync/16/256     2192763 ns       322532 ns        12776
BM_CompletionQueueRunAsync/16/512     2853664 ns       778332 ns         6015
BM_CompletionQueueRunAsync/16/1024    5217779 ns      2661059 ns         1703
BM_CompletionQueueRunAsync/16/2048   10738621 ns      7784985 ns          520
BM_CompletionQueueRunAsync/16/4096   20561458 ns     16853590 ns          267
BM_CompletionQueueRunAsync_BigO       5093.26 N       3943.52 N    
BM_CompletionQueueRunAsync_RMS             12 %            20 %    
```

This is too expensive for Cloud Pub/Sub, we want to handle about 1.2 million events.

I have a prototype for an alternative implementation that takes about 0.3us of CPU (0.8us elapsed):

```
Run on (96 X 2000.15 MHz CPU s)
CPU Caches:
  L1 Data 32K (x48)
  L1 Instruction 32K (x48)
  L2 Unified 1024K (x48)
  L3 Unified 39424K (x2)
Load Average: 1.48, 1.44, 1.29
---------------------------------------------------------------------------------------
Benchmark                                             Time             CPU   Iterations
---------------------------------------------------------------------------------------
BM_CompletionQueueRunAsyncWakeupOnly/16/64       749826 ns        57987 ns        72024
BM_CompletionQueueRunAsyncWakeupOnly/16/128      742435 ns        80440 ns        59046
BM_CompletionQueueRunAsyncWakeupOnly/16/256      847208 ns       114724 ns        36172
BM_CompletionQueueRunAsyncWakeupOnly/16/512      949273 ns       154964 ns        28834
BM_CompletionQueueRunAsyncWakeupOnly/16/1024     984168 ns       189565 ns        22939
BM_CompletionQueueRunAsyncWakeupOnly/16/2048    1710185 ns       410144 ns        12302
BM_CompletionQueueRunAsyncWakeupOnly/16/4096    3244850 ns      1201626 ns         4170
BM_CompletionQueueRunAsyncWakeupOnly_BigO        833.64 N        271.75 N    
BM_CompletionQueueRunAsyncWakeupOnly_RMS             36 %            25 %    
```

## Background

Remember that `CompletionQueue` has an associated pool of threads, this pool is implicit, any thread that calls `CompletionQueue::Run()` joins the pool until the completion queue is shutdown.

The implementation of `CompletionQueue::RunAsync(f)` basically creates a gRPC timer which immediately expires and calls the `f`. This execution happens in one of the threads running `CompletionQueue::Run()`. There is no guarantee of ordering between these calls.

## Overview

This is a proposal to change the completion queue to:

* keep a queue of asynchronous functions to call
* when the queue becomes non-empty the completion queue wakes up one of the threads in the completion queue thread pool to drain this queue
   * If one thread is not keeping up with the growth of the queue we may wake up additional threads
* the woken up threads run until the queue is drained
* we always keep at least one thread running the underlying I/O event loop

### Waking Up Additional Threads

This implementation keeps track of how many threads are running in the completion queue thread pool and how many of those threads are draining the async function queue.

* Let  `P` be the number of threads in the pool,
* Let `D` be the number of those threads that are draining the queue
* If the queue is over the "too long threshold" (initially 1) then:
  * If `P == 1` then we wake up the queue to handle **one** function
    * The handler will need to reschedule itself it there are still functions to handle
  * If `P > 1` then:
    * One of the queues in the pool is woken up to drain the queue and the "too long threshold" is doubled.
* After handling a function we check the queue length
  * If the queue length is less than 1/4 of the "too long threshold" then the threshold is reduced by 1/2.

coryan:
Instead of signaling the underlying grpc::CompletionQueue for each new
function we want to run keep a queue and only wakeup enough threads to
handle the queue.

We do not want to starve the I/O, so we need to take some care when (a)
there is a single thread in the completion queue, and (b) when there are
so many async functions that they could use all the completion queue
threads. In the first case we pretty much use the old approach, in the
second case we never use all the completion queue threads for
asynchronous functions, we reserve at least one thread for I/O.

Fixes #5403, see that bug for details.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5406)


coryan:
Argh, sorry, this is not quite ready.  I want to run more tests before declaring success.

coryan:
Okay, this is now ready for review. It is a large test, but mostly it is test code :relaxed:

google/cloud/completion_queue_test.cc
```diff
@@ -488,6 +489,134 @@ TEST(CompletionQueueTest, RunAsyncThread) {
...
+  std::thread t{[&] { cq.Run(); }};
+
+  // Make each function in RunAsync() block under control of this thread.
+  std::mutex mu;
+  std::condition_variable cv;
+  std::deque<promise<void>> waiting;
+  auto add_promise = [&] {
+    std::unique_lock<std::mutex> lk(mu);
+    waiting.emplace_back();
+    auto f = waiting.back().get_future();
```
devbww, line 557: Any reason this differs from the pushes to `waiting` in the surrounding tests?

google/cloud/completion_queue_test.cc
```diff
@@ -513,6 +642,83 @@ TEST(CompletionQueueTest, NoRunAsyncAfterShutdown) {
   EXPECT_EQ(std::future_status::timeout, f.wait_for(ms(0)));
 }
 
+class RunAsyncTest : public ::testing::TestWithParam<int> {};
+
+TEST_P(RunAsyncTest, Torture) {
```
devbww, line 647: It appears that these torture tests never fail (no expectations).  Is there something we could do to check that they are executing as we expect?  Is the general test timeout sufficient to check that the `wait()` completes, or would something more specific be nice?  Or would these just be better off as benchmarks?

google/cloud/completion_queue_benchmark.cc
```diff
@@ -0,0 +1,133 @@
...
+  void BlockUntilDone() {
+    std::unique_lock<std::mutex> lk(mu_);
+    cv_.wait(lk, [&] { return count_ == 0; });
+  }
+
+  void Done() {
+    std::unique_lock<std::mutex> lk(mu_);
+    if (--count_ == 0) cv_.notify_one();
+    lk.unlock();
+  }
```
devbww, line 68: "Done" is a little ambiguous here.  Perhaps something like `BlockUntilAllDone()` and `OneDone()`?

google/cloud/completion_queue_benchmark.cc
```diff
@@ -0,0 +1,133 @@
...
+// BM_CompletionQueueRunAsync_RMS             24 %             9 %
+
+auto constexpr kMinThreads = 16;
+auto constexpr kMaxThreads = 16;
+auto constexpr kMinExecutions = 1 << 9;
+auto constexpr kMaxExecutions = 1 << 11;
+
+class Wait {
+ public:
+  explicit Wait(int count) : count_(count) {}
```
devbww, line 57: Why not use `std::int64_t count` to obviate the casts below?

google/cloud/completion_queue_benchmark.cc
coryan: Thanks, fixed.

google/cloud/completion_queue_benchmark.cc
coryan: Done.

google/cloud/completion_queue_test.cc
coryan: None, fixed by refactoring.

google/cloud/completion_queue_test.cc
coryan: I added some reasonable expectations, but you are right.  These are largely smoke tests, if they crash (or fail under a sanitizer) we have a problem, otherwise we do not know (but Dijkstra would say that is true of all tests).

I do not feel very strongly about these tests, if you think they do not add enough value, I am happy to remove them.

coryan:
PTAL

google/cloud/completion_queue_test.cc
devbww: I'm happy to keep them.

devbww (approved):

google/cloud/internal/default_completion_queue_impl.cc
```diff
@@ -305,20 +301,21 @@ void DefaultCompletionQueueImpl::DrainRunAsyncOnIdle() {
 
 void DefaultCompletionQueueImpl::WakeUpRunAsyncThread(
     std::unique_lock<std::mutex> lk) {
+  // Nothing to do, just return.
```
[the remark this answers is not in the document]
coryan: Done.

PR #5409, opened 2020-11-05
fix(pubsub): sometimes published oversized batches

### LINKED ISSUE #5408: Publish batches can exceed the byte size limits
After some layers of indirection `Publisher::Publish()`  is implemented by:

https://github.com/googleapis/google-cloud-cpp/blob/9eff7232c7a9dcf5ad3bd41d29b9d42b84f202d0/google/cloud/pubsub/internal/batching_publisher_connection.cc#L69

Which first adds a message:

https://github.com/googleapis/google-cloud-cpp/blob/9eff7232c7a9dcf5ad3bd41d29b9d42b84f202d0/google/cloud/pubsub/internal/batching_publisher_connection.cc#L97

and then verifies if the batch must be flushed:

https://github.com/googleapis/google-cloud-cpp/blob/9eff7232c7a9dcf5ad3bd41d29b9d42b84f202d0/google/cloud/pubsub/internal/batching_publisher_connection.cc#L100

This does not work correctly when sending large messages.  The previous batched messages may be too small, but adding the message may create a batch over the limit. Most of the time this does not matter.  However, Pub/Sub has a hard limit of 10MiB. If the batch setting is something like 9.5MiB, and the new message is about 1MiB this can create a problem.

We need to check if the message fits in the batch *first*, if it does not we should first flush the batch, and only then add it.

There are some complications (aren't there always), we release the lock to flush a batch, so we may get other messages in between.  And the message itself may be too large, in which case we should send it on its own (and let the server validate the size).

coryan:
`pubsub_internal::BatchingPublisherConnection` failed to flush before
accepting a message that would exceed the total bytes limit.

Fixes #5408

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5409)

### DIFF
--- a/google/cloud/pubsub/internal/batching_publisher_connection.cc
+++ b/google/cloud/pubsub/internal/batching_publisher_connection.cc
@@ -68,21 +68,21 @@ struct Batch {
 
 future<StatusOr<std::string>> BatchingPublisherConnection::Publish(
     PublishParams p) {
-  promise<StatusOr<std::string>> pr;
-  auto f = pr.get_future();
   auto const bytes = pubsub_internal::MessageSize(p.message);
   std::unique_lock<std::mutex> lk(mu_);
-  if (!corked_status_.ok()) {
-    struct MoveCapture {
-      promise<StatusOr<std::string>> p;
-      Status status;
-      void operator()() { p.set_value(std::move(status)); }
-    };
-    cq_.RunAsync(MoveCapture{std::move(pr), corked_status_});
-    return f;
-  }
+  do {
+    if (!corked_status_.ok()) return CorkedError();
+    auto const has_capacity =
+        current_bytes_ + bytes < options_.maximum_batch_bytes();
+    if (has_capacity || waiters_.empty()) break;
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/internal/batching_publisher_connection.cc
```diff
@@ -178,6 +176,22 @@ void BatchingPublisherConnection::OnTimer() {
...
+    promise<StatusOr<std::string>> p;
+    Status status;
+    void operator()() { p.set_value(std::move(status)); }
+  };
+  cq_.RunAsync(MoveCapture{std::move(p), corked_status_});
+
+  return f;
+}
+
+void Fits(std::unique_lock<std::mutex> const& lk, std::size_t bytes);
```
devbww, line 193: Remnant of something?

coryan: Yes, thanks!  I am surprised the compiler(s) did not catch that, thanks for pointing it out.

devbww (approved):

PR #5410, opened 2020-11-05
test(pubsub): deflake batching by time unit test

### LINKED ISSUE #5369: [Flake] pubsub unit test times out with strange error
https://source.cloud.google.com/results/invocations/228df956-41e5-4c58-8631-d4f95292854c

```
306/306 Test #370: pubsub_internal_batching_publisher_connection_test .................***Timeout 1500.02 sec
[==========] Running 10 tests from 1 test suite.
[----------] 10 tests from BatchingPublisherConnectionTest
[ RUN      ] BatchingPublisherConnectionTest.DefaultMakesProgress
[       OK ] BatchingPublisherConnectionTest.DefaultMakesProgress (58 ms)
[ RUN      ] BatchingPublisherConnectionTest.BatchByMessageCount
[       OK ] BatchingPublisherConnectionTest.BatchByMessageCount (18 ms)
[ RUN      ] BatchingPublisherConnectionTest.BatchByMessageSize
[       OK ] BatchingPublisherConnectionTest.BatchByMessageSize (23 ms)
[ RUN      ] BatchingPublisherConnectionTest.BatchByMaximumHoldTime
[       OK ] BatchingPublisherConnectionTest.BatchByMaximumHoldTime (12 ms)
[ RUN      ] BatchingPublisherConnectionTest.BatchByFlush
/Volumes/BuildData/tmpfs/src/github/google-cloud-cpp/google/cloud/pubsub/internal/batching_publisher_connection_test.cc:250: Failure
  2
  request.messages_size()
    Which is: 1
[libprotobuf FATAL /Volumes/BuildData/tmpfs/src/github/google-cloud-cpp/cmake-out/macos/external/include/google/protobuf/repeated_field.h:1537] CHECK failed: (index) < (current_size_): 
/Volumes/BuildData/tmpfs/src/github/google-cloud-cpp/google/cloud/pubsub/internal/batching_publisher_connection_test.cc:289: Failure
  "test-message-id-" + std::to_string(i)
    Which is: "test-message-id-1"
  *r
    Which is: "ack-for-test-data-1"


99% tests passed, 1 tests failed out of 306

Total Test time (real) = 1514.54 sec

The following tests FAILED:
	370 - pubsub_internal_batching_publisher_connection_test (Timeout)
Errors while running CTest
make[2]: *** [build/google-cloud-cpp/src/google_cloud_cpp_project-stamp/google_cloud_cpp_project-test] Error 8
make[1]: *** [CMakeFiles/google_cloud_cpp_project.dir/all] Error 2
make: *** [all] Error 2
2020-10-29T14:28:00Z: build script reported errors. 
```

coryan:
The `BatchByMaximumHoldTime` test was flaky, it would fail under heavy
load because it assumed publishing an event took less than 5ms.

Fixes #5369

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5410)

### DIFF
--- a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
+++ b/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
@@ -208,14 +208,15 @@ TEST(BatchingPublisherConnectionTest, BatchByMaximumHoldTime) {
         return make_ready_future(make_status_or(response));
       });
 
-  google::cloud::internal::AutomaticallyCreatedBackgroundThreads background;
-  auto const ordering_key = std::string{};
+  // Start with an inactive message queue, to avoid flakes due to scheduling
+  // problems.
+  CompletionQueue cq;
   auto publisher = BatchingPublisherConnection::Create(
       topic,
       pubsub::PublisherOptions{}
           .set_maximum_hold_time(std::chrono::milliseconds(5))
           .set_maximum_batch_message_count(4),
-      ordering_key, mock, background.cq(), pubsub_testing::TestRetryPolicy(),
+      /*ordering_key=*/{}, mock, cq, pubsub_testing::TestRetryPolicy(),
       pubsub_testing::TestBackoffPolicy());
   auto r0 =
       publisher
@@ -234,8 +235,15 @@ TEST(BatchingPublisherConnectionTest, BatchByMaximumHoldTime) {
             EXPECT_EQ("test-message-id-1", *r);
           });
 
+  // Now that the two messages are queued, we can activate the completion queue.
+  // It should flush the messages in about 5ms.
+  std::thread t{[](CompletionQueue cq) { cq.Run(); }, cq};
+
   r0.get();
   r1.get();
+
+  cq.Shutdown();
+  t.join();
 }
 
 TEST(BatchingPublisherConnectionTest, BatchByFlush) {


devbww (approved):
> it assumed publishing an event took less than 5ms

I can't tell where that assumption was made, but looks good in any case.

coryan:
> I can't tell where that assumption was made, but looks good in any case.

In case you are curious, we tell the batcher to hold messages for at most 5ms:

https://github.com/googleapis/google-cloud-cpp/blob/2a6bf4b7ffab8887bd72463aa3a4d7328c5d367a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc#L216

Then we publish two messages "back to back":

https://github.com/googleapis/google-cloud-cpp/blob/2a6bf4b7ffab8887bd72463aa3a4d7328c5d367a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc#L222

https://github.com/googleapis/google-cloud-cpp/blob/2a6bf4b7ffab8887bd72463aa3a4d7328c5d367a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc#L230

Normally those calls take less than 5ms put together, so the publisher groups them in a single `AsyncPublish` call:

https://github.com/googleapis/google-cloud-cpp/blob/2a6bf4b7ffab8887bd72463aa3a4d7328c5d367a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc#L198

that contains both:

https://github.com/googleapis/google-cloud-cpp/blob/2a6bf4b7ffab8887bd72463aa3a4d7328c5d367a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc#L202

 sadly when the CI machine is sufficiently loaded those two messages are "seen" by the batching threads more than 5ms apart, and they are split into two messages.

devbww:
So the same thing may still happen with a sufficient delay between the two `Publish()` calls, yes?

coryan:
> So the same thing may still happen with a sufficient delay between the two `Publish()` calls, yes?

I do not think so (but check my math), the `BatchingPublisherConnection`:

https://github.com/googleapis/google-cloud-cpp/blob/99d81e603279eff6225cd3db3449ad5876df3ecb/google/cloud/pubsub/internal/batching_publisher_connection_test.cc#L214

 uses this completion queue:

https://github.com/googleapis/google-cloud-cpp/blob/99d81e603279eff6225cd3db3449ad5876df3ecb/google/cloud/pubsub/internal/batching_publisher_connection_test.cc#L213

That is where it sets its timers, performs I/O, etc.  That queue is not active, there are no threads blocked on `.Run()`.  `publisher` can request all the timers it wants, but none of them will execute until the completion queue has at least one thread attached.  That happens here:

https://github.com/googleapis/google-cloud-cpp/blob/99d81e603279eff6225cd3db3449ad5876df3ecb/google/cloud/pubsub/internal/batching_publisher_connection_test.cc#L240

After the messages are queued.  So however long the delay is between, before, or after the messages, they cannot be sent until that point.

Note that the test is verifying that a batch too small to flush because of size:

https://github.com/googleapis/google-cloud-cpp/blob/99d81e603279eff6225cd3db3449ad5876df3ecb/google/cloud/pubsub/internal/batching_publisher_connection_test.cc#L218

gets flushed eventually.  We do not really care how precise the flushing timeout is (maybe we should, but we don't in this test).

devbww:
Ack.  Thanks.

Issue #5284, opened 2020-10-20
FLAKE: gcc-9 GCS signed_url_integration_test HTTP errors


devjgm:
Log: https://g3c.corp.google.com/results/invocations/fc63b4f0-c4cb-46d6-8e5e-737e387c6744/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fdocker%2Fgcc-9/log

```
================================================================
2020-10-20T13:37:38Z: Running the integration tests against production 

================================================================
2020-10-20T13:37:38Z: Create the gcloud configuration for the cloud-cpp tests. 
Updated property [core/project].
Activated service account credentials for: [kokoro-run@cloud-cpp-testing-resources.iam.gserviceaccount.com]
Updated property [core/account].
2020-10-20T13:37:40Z: Revoke service account after creating the access token. 
WARNING: [kokoro-run@cloud-cpp-testing-resources.iam.gserviceaccount.com] appears to be a service account. Service account tokens cannot be revoked, but they will expire automatically. To prevent use of the service account token earlier than the expiration, revoke the parent service account or service account key.
Test project /v/cmake-out/ci-fedora-install-31-gcc-9

...

[ RUN      ] SignedUrlIntegrationTest.CreateV2SignedUrlPut
../../google/cloud/storage/tests/signed_url_integration_test.cc:106: Failure
Expected equality of these values:
  200
  response->status_code
    Which is: 500
../../google/cloud/storage/tests/signed_url_integration_test.cc:110: Failure
Expected equality of these values:
  expected
    Which is: "Lorem ipsum dolor sit amet, consectetur adipiscing\nelit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim\nad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea\ncommodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit\nesse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat\nnon proident, sunt in culpa qui officia deserunt mollit anim id est laborum.\n"
  actual
    Which is: ""
With diff:
@@ -1,6 +1,1 @@
-Lorem ipsum dolor sit amet, consectetur adipiscing
-elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim
[...]

PR #5414, opened 2020-11-05
test(storage): add support for grpc

vnvo2409:
This PR:
- adds support for grpc.
- makes `grpc_integration_test` testing with both `media` and `metadata` since it is our ultimate goal.
- `CI` changes

`gevent` and `gRPC` can not work together, so I change the worker to `sync` and use `threads`. I also add a endpoint `start_grpc` to start the grpc server. The reason is `Gunicorn` will spawn a new subprocess when running `Flask` server. If we start `gRPC` server before the spawn of the subprocess, it's nearly impossible to share the `database` with the new subprocess because Python will copy everything in the memory from the parent process to the subprocess ( So we have 2 separate instance of `database` ). The endpoint will start the `gRPC` server in the same subprocess so there is only one instance of `database`.

Part of #4751 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414)

### DIFF
--- a/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
+++ b/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
@@ -90,4 +90,9 @@ testbench_args=(
   "${excluded_targets[@]}"
 exit_status=$?
 
+if [[ "$exit_status" -ne 0 ]]; then
+  source "${PROJECT_ROOT}/ci/define-dump-log.sh"
+  dump_log testbench.log
+fi
+
 exit "${exit_status}"
--- a/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
+++ b/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
@@ -35,11 +35,9 @@ source "${PROJECT_ROOT}/google/cloud/storage/tools/run_testbench_utils.sh"
 
 # These can only run against production
 production_only_targets=(
-  "//google/cloud/storage/examples:storage_grpc_samples"
   "//google/cloud/storage/examples:storage_policy_doc_samples"
   "//google/cloud/storage/examples:storage_signed_url_v2_samples"
   "//google/cloud/storage/examples:storage_signed_url_v4_samples"
-  "//google/cloud/storage/tests:grpc_integration_test"
   "//google/cloud/storage/tests:key_file_integration_test"
   "//google/cloud/storage/tests:signed_url_integration_test"
 )
@@ -72,6 +70,7 @@ done
 # are missing too.
 testbench_args=(
   "--test_env=CLOUD_STORAGE_TESTBENCH_ENDPOINT=${CLOUD_STORAGE_TESTBENCH_ENDPOINT}"
+  "--test_env=CLOUD_STORAGE_GRPC_ENDPOINT=${CLOUD_STORAGE_GRPC_ENDPOINT}"
   "--test_env=HTTPBIN_ENDPOINT=${HTTPBIN_ENDPOINT}"
   "--test_env=GOOGLE_CLOUD_CPP_STORAGE_TEST_HMAC_SERVICE_ACCOUNT=fake-service-account-sign@example.com"
   "--test_env=GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES=yes"
--- a/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
+++ b/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
@@ -51,7 +51,7 @@ production_only_targets=(
 pushd "${HOME}" >/dev/null
 # Start the testbench on a fixed port, otherwise the Bazel cache gets
 # invalidated on each run.
-start_testbench 8585
+start_testbench 8585 8000
 popd >/dev/null
 
 excluded_targets=(
--- a/google/cloud/storage/ci/run_integration_tests_emulator_cmake.sh
+++ b/google/cloud/storage/ci/run_integration_tests_emulator_cmake.sh
@@ -62,4 +62,9 @@ exit_status=$?
 kill_testbench
 trap '' EXIT
 
+if [[ "$exit_status" -ne 0 ]]; then
+  source "${PROJECT_ROOT}/ci/define-dump-log.sh"
+  dump_log testbench.log
+fi
+
 exit "${exit_status}"
--- a/google/cloud/storage/emulator/database.py
+++ b/google/cloud/storage/emulator/database.py
@@ -33,6 +33,13 @@ class Database:
     def init(cls):
         return cls({}, {}, {}, {}, {})
 
+    def raii(self, grpc_server):
+        self.grpc_server = grpc_server
+
+    def __del__(self):
+        if hasattr(self, "grpc_server") and self.grpc_server is not None:
+            self.grpc_server.stop(None)
+
     # === BUCKET === #
 
     def __check_bucket_metageneration(self, request, bucket, context):
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -43,6 +43,14 @@ def index():
 
 @root.route("/start_grpc")
 def start_grpc():
+    # The reason is `Gunicorn` will spawn a new subprocess ( a worker )
+    # when running `Flask` server. If we start `gRPC` server before
+    # the spawn of the subprocess, it's nearly impossible to share
+    # the `database` with the new subprocess because Python will copy
+    # everything in the memory from the parent process to the subprocess
+    # ( So we have 2 separate instance of `database` ). The endpoint
+    # will start the `gRPC` server in the same subprocess
+    # so there is only one instance of `database`.
     global grpc_port
     if grpc_port == 0:
         port = flask.request.args.get("port", "0")
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -29,7 +29,7 @@ from google.cloud.storage_v1.proto.storage_resources_pb2 import CommonEnums
 from google.protobuf import json_format
 
 db = None
-grpc_is_running = False
+grpc_port = 0
 
 # === DEFAULT ENTRY FOR REST SERVER === #
 root = flask.Flask(__name__)
@@ -43,11 +43,12 @@ def index():
 
 @root.route("/start_grpc")
 def start_grpc():
-    global grpc_is_running
-    if not grpc_is_running:
-        port = flask.request.args.get("port", "8000")
-        grpc_server.run(port, db)
-    return "OK"
+    global grpc_port
+    if grpc_port == 0:
+        port = flask.request.args.get("port", "0")
+        grpc_port = grpc_server.run(port, db)
+        return str(grpc_port)
+    return str(grpc_port)
 
 
 @root.route("/<path:object_name>", subdomain="<bucket_name>")
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -18,6 +18,7 @@ import logging
 import database
 import flask
 import gcs as gcs_type
+import grpc_server
 import httpbin
 import utils
 from werkzeug import serving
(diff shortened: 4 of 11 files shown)


vnvo2409:
@coryan 
Could you kick-off a CI please ?

vnvo2409:
Both `bazel` and `cmake` passed ( I forgot to remove the `storage_grpc_samples` out of the tests against production in `cmake` but it passed with `bazel` ).

`clang-tidy` failed because of formating.

What do you think about this PR ? @coryan

coryan (commented):
> What do you think about this PR ?

I am sorry I have been less responsive than usual, I am super busy with other stuff. The PR is looking good, please fix formatting and there are some questions below.  Let me know those make no sense. 

> I have to disable one test because it always runs against production:

SGTM

> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-:-MLO_quo8-AZ6RDkRynW:bj3lj42)* status: all files reviewed, 4 unresolved discussions (waiting on @vnvo2409)

___
*[google/cloud/storage/emulator/emulator.py, line 45 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-MLOPT8hEj1VI8DxHEQa:-MLOPT8hEj1VI8DxHEQb:b-pky5vb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b446c9eaf952e81a6847a9ce7b71924439811fd6/google/cloud/storage/emulator/emulator.py#L45)):*
> ```Python
> 
> @root.route("/start_grpc")
> def start_grpc():
> ```

Is there a reason we cannot start this immediately?  If there is, can you add a comment explaining why?
___
*[google/cloud/storage/emulator/grpc_server.py, line 30 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-MLOPbLP4KNry7_V5tF8:-MLOPbLP4KNry7_V5tF9:b5vngxr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b446c9eaf952e81a6847a9ce7b71924439811fd6/google/cloud/storage/emulator/grpc_server.py#L30)):*
> ```Python
> 
> 
> class StorageServicer(storage_pb2_grpc.StorageServicer):
> ```

FYI... "Servicer" is a weird name, I see that it is the convention in Python...  :man_shrugging: 
___
*[google/cloud/storage/emulator/grpc_server.py, line 159 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-MLOPnGL7AvhYYfNSenW:-MLOPnGL7AvhYYfNSenX:b-7p4vua) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b446c9eaf952e81a6847a9ce7b71924439811fd6/google/cloud/storage/emulator/grpc_server.py#L159)):*
> ```Python
>     db = database
>     storage_pb2_grpc.add_StorageServicer_to_server(StorageServicer(), server)
>     server.add_insecure_port("[::]:" + port)
> ```

I think this is assuming IPv6, is there a reason why `"localhost:" + port` does not work?
___
*[google/cloud/storage/emulator/gcs/holder.py, line 94 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-MLO_RvP6u1uEdW70kaa:-MLO_RvP6u1uEdW70kab:b6q17jx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b446c9eaf952e81a6847a9ce7b71924439811fd6/google/cloud/storage/emulator/gcs/holder.py#L94)):*
> ```Python
>             ("%s/o/%s" % (bucket.name, metadata.name)).encode("utf-8")
>         ).hexdigest()
>         fake_request = utils.common.FakeRequest.init_protobuf(request, context)
> ```

Could you add a comment explaining  what is going on here?
___
*[google/cloud/storage/tools/run_testbench_utils.sh, line 109 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-MLO_E2e5PQzKaA-xxDl:-MLO_E2f1Sx9b2jOIjJ5:b-golple) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b446c9eaf952e81a6847a9ce7b71924439811fd6/google/cloud/storage/tools/run_testbench_utils.sh#L109)):*
> ```Shell
>   fi
> 
>   grpc_port=8000
> ```

Could we make this an argument received by the function?  And could we make it accept `0` and pick a random port for us?

vnvo2409:
> Is there a reason we cannot start this immediately? If there is, can you add a comment explaining why?

Changed to `localhost`. The homepage uses IPv6 so I copied it.

---

Sum up what I pushed:
f0b8c8f: Formatting

---

964c83e: 
I accidentally pass `None` to `context` when working with gRPC ( probaly because of copy-paste ) so this commit fix it. In addition, It adds a feature to update the `FakeRequest` based on scalar fields in the protobuf request.
```python
            ("%s/o/%s" % (bucket.name, metadata.name)).encode("utf-8")
        ).hexdigest()
        fake_request = utils.common.FakeRequest.init_protobuf(request, context)
```
> Could you add a comment explaining what is going on here?

`init_protobuf` is used to flatten protobuf request. For an example, `InsertObjectRequest`'s structure is
```protobuf
message InsertObjectRequest {
  // The first message of each stream should set one of the following.
  oneof first_message {
    // For resumable uploads. This should be the `upload_id` returned from a
    // call to `StartResumableWriteResponse`.
    string upload_id = 1;

    // For non-resumable uploads. Describes the overall upload, including the
    // destination bucket and object name, preconditions, etc.
    InsertObjectSpec insert_object_spec = 2;
  }
  CommonObjectRequestParams common_object_request_params = 3;

  // A set of parameters common to all Storage API requests.
  CommonRequestParams common_request_params = 4;
  // ...
}
message InsertObjectSpec {
  // Destination object, including its name and its metadata.
  Object resource = 1;

  // Apply a predefined set of access controls to this object.
  CommonEnums.PredefinedObjectAcl predefined_acl = 2;
  // ....
  CommonEnums.Projection projection = 7;
}

```
If we want to extract `predefined_acl`, we will have to do `request-> insert_object_spec-> predefined_acl` which will eventually break all the utils functions. So the `init_protobuf` will create a `FakeRequest` whose structure is something like:
```protobuf
{
//...
CommonEnums.PredefinedObjectAcl predefined_acl = 2;
CommonObjectRequestParams common_object_request_params = 3;
CommonRequestParams common_request_params = 4;
//...
}
```
We could save that `fake_request` and pass it around without worrying about the nested message.

 ---

61d42c8
> Could we make this an argument received by the function? And could we make it accept 0 and pick a random port for us?

Done but I am not so comfortable with `bash` so we will have to wait for the CI.

vnvo2409:
```python
@root.route("/start_grpc")
def start_grpc():
```
> Is there a reason we cannot start this immediately? If there is, can you add a comment explaining why?

The reason is `Gunicorn` will spawn a new subprocess ( a worker ) when running Flask server. If we start `gRPC` server before the spawn of the subprocess, it's nearly impossible to share the `database` with the new subprocess because Python will copy everything in the memory from the parent process to the subprocess ( So we have 2 separate instance of `database` ). The endpoint will start the `gRPC` server in the same subprocess so there is only one instance of `database`. Please let me know if there is something that is unclear.

coryan (changes requested):
The changes look great, thank you.  Just this one thing:

>> Is there a reason we cannot start this immediately? If there is, can you add a comment explaining why?
> 
> The reason is Gunicorn will spawn a new subprocess ( a worker ) when running Flask server. If we start gRPC server before the spawn of the subprocess, it's nearly impossible to share the database with the new subprocess because Python will copy everything in the memory from the parent process to the subprocess ( So we have 2 separate instance of database ). The endpoint will start the gRPC server in the same subprocess so there is only one instance of database. Please let me know if there is something that is unclear.

Your explanation makes perfect sense. But our future-selves will have forgotten it. Putting that comment in the code will make it easier to remember why it is there.

> Reviewed 5 of 6 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-:-MLSkkb8BTZHwXIdtoVJ:b-jo7fr9)* status: all files reviewed, 3 unresolved discussions (waiting on @vnvo2409)

___
*[google/cloud/storage/tools/run_testbench_utils.sh, line 111 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-MLSlFeI6ozAo33RcMZb:-MLSlFeI6ozAo33RcMZc:b-ibhdji) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e83ba689a0d384c670df997da9b57fae9aea036/google/cloud/storage/tools/run_testbench_utils.sh#L111)):*
> ```Shell
>   port="${2:-0}"
>   local grpc_port=""
>   grpc_port=$(curl -s "http://localhost:${testbench_port}/start_grpc?port=${port}")
> ```

I wonder if we need a retry loop here?  I think maybe you can add `--retry-times=5` for curl to do the retry for you.

vnvo2409:
@coryan 
Is there a way to get the `testbench.log` inside kokoro :/ `kokoro/docker/coverage` failed and it only shows `Internal Server Error`. I can not reproduce it locally.

coryan:
> @coryan
> Is there a way to get the `testbench.log` inside kokoro

Unfortunately not. These are all discarded when the build finishes.

> In addition, Issues like #5284 could be solved easier if we have access to `testbench.log`.

Ack.

>If we could not download the log, how about `cat testbench.log` before stopping kokoro if the integration tests failed ?

That SGTM, feel free to add this to your PR so we can troubleshoot.  In case it  there is a bash function to dump only the first N and last N lines of a lot already:

https://github.com/googleapis/google-cloud-cpp/blob/master/ci/define-dump-log.sh

google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh
```diff
@@ -91,4 +90,9 @@ testbench_args=(
   "${excluded_targets[@]}"
 exit_status=$?
 
+if [[ "$exit_status" -ne 0 ]]; then
+  source "${PROJECT_ROOT}/ci/define-dump-log.sh"
+  dump_log testbench.log
```
vnvo2409, line 95: Is it the right location of `testbench.log` ?

coryan: Ooops, no.  I think it is placed in `$HOME`, see above just before `start_testbench`.

vnvo2409: Could I fix it in the next PR ?

coryan: Yes, absolutely, I thought you needed it to troubleshoot something.

vnvo2409: This time, no test failed so I cannot troubleshoot anything. Just to be sure, the extract location of `testbench.log` is `$HOME/testbench.log` right ?

coryan:
___
*[google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh, line 95 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-MLNlXRlGr7Wkmn8tX_Y-r4-95:-MLTV9zZ5ahFRguDSY06:b-903fw7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c20dae017e8950e9a214989984b846ab237d4390/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh#L95)):*
<details><summary><i>Previously, vnvo2409 (Vo Van Nghia) wrote…</i></summary><blockquote>

This time, no test failed so I cannot troubleshoot anything. Just to be sure, the extract location of `testbench.log` is `$HOME/testbench.log` right ?
</blockquote></details>

I believe that is correct.

coryan (commented):
I created #5428 to track restoring this test to the code coverage build:

>            TEST_F(GrpcIntegrationTest, GetObjectMediaNotFound) { 

Otherwise, this :lgtm_strong: Thanks so much for making this multi-step contribution.

> Reviewed 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5414#-:-MLTUtgJ33bxNIDWYI3V:b-2zx18n)* status: all files reviewed, 2 unresolved discussions (waiting on @vnvo2409)

coryan:
I am reverting this PR, sorry, the builds are failing after the merge.

PR #5415, opened 2020-11-05
fix(pubsub): keep corked after ResumePublish

### LINKED ISSUE #5412: Publisher::ResumePublish flushes corked messages
The function should have no effect if there have been no errors publishing previous messages.

Currently it publishes all corked messages, even if there is an `AsyncPublish()` pending:

https://github.com/googleapis/google-cloud-cpp/blob/abe41c90ce26e04807f4bf3ebbe1f9c9e543df35/google/cloud/pubsub/internal/batching_publisher_connection.cc#L110

https://github.com/googleapis/google-cloud-cpp/blob/abe41c90ce26e04807f4bf3ebbe1f9c9e543df35/google/cloud/pubsub/internal/batching_publisher_connection.cc#L117

Seems like `ResumePublish()` should *only* reset `corked_status_` and leave `corked_` alone.

coryan:
We were flushing after each `Publisher::ResumePublish()` which could
reorder messages. We need to track "corked because there was an error"
(which is affected by `Publisher::ResumePublish()`) and "corked because
we are waiting for a response", which should not.

Fixes #5412

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5415)

### DIFF
--- a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
+++ b/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
@@ -838,7 +838,7 @@ TEST(BatchingPublisherConnectionTest, OrderingBatchDiscardOnError) {
     using ms = std::chrono::milliseconds;
     EXPECT_EQ(std::future_status::timeout, r.wait_for(ms(0)));
   }
-  // Trigger the first response, the should all fail with the error response.
+  // Trigger the first response. They should all fail with the error response.
   wait_pending().set_value();
   for (auto& r : responses) {
     EXPECT_THAT(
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/internal/batching_publisher_connection_test.cc
```diff
@@ -786,7 +838,7 @@ TEST(BatchingPublisherConnectionTest, OrderingBatchDiscardOnError) {
     using ms = std::chrono::milliseconds;
     EXPECT_EQ(std::future_status::timeout, r.wait_for(ms(0)));
   }
-  // Trigger the first response.
+  // Trigger the first response, the should all fail with the error response.
```
devbww, line 841: s/, the/. They/?

devbww (approved):

google/cloud/pubsub/internal/batching_publisher_connection_test.cc
coryan: Fixed. Thanks.

PR #5429, opened 2020-11-06
WIP - test build

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5429)


vnvo2409:
It seems that the problem is always caused by an `Internal Server Error` in `ListObject`. I will submit a PR that could dump the `testbench.log` and we could have some insight.

coryan:
@vnvo2409 I need to unblock other work, so I am going to revert #5414.  No rush, we can try again and I will try to help more with the testing (I have been very busy in the last couple of days).

PR #5447, opened 2020-11-08
feat(pubsub): pace publisher in throughtput benchmark

coryan:
Limit the number of messages per second generated by the publishing
threads, this is useful when we want to avoid the synchronization
overhead of message-based flow control.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5447)

### DIFF
--- a/google/cloud/pubsub/benchmarks/throughput.cc
+++ b/google/cloud/pubsub/benchmarks/throughput.cc
@@ -274,7 +274,7 @@ class PublishWorker {
           });
       ++send_count;
       send_bytes.fetch_add(bytes);
-      if (enable_pacing && i != 0 && i % kPacingCount == 0) {
+      if (enable_pacing && (i + 1) % kPacingCount == 0) {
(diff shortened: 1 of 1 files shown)


google/cloud/pubsub/benchmarks/throughput.cc
```diff
@@ -248,6 +265,11 @@ class PublishWorker {
             OnAck();
           });
       ++send_count;
+      if (enable_pacing && i != 0 && i % kPacingCount == 0) {
```
devbww, line 268: Nit: The asymmetry in cases like these always strikes me.  How about `(i + 1) % kPacingCount == 0`?

coryan: I like it, thanks.

coryan:
Rebased, resolved conflicts and addressed comments. PTAL.

devbww (approved):

PR #5452, opened 2020-11-09
feat(spanner): use spanner_commit_stats protos

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5452)

### DIFF
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -75,10 +75,10 @@ def google_cloud_cpp_deps():
         http_archive(
             name = "com_google_googleapis",
             urls = [
-                "https://github.com/googleapis/googleapis/archive/59f97e6044a1275f83427ab7962a154c00d915b5.tar.gz",
+                "https://github.com/googleapis/googleapis/archive/spanner_commit_stats.tar.gz",
             ],
-            strip_prefix = "googleapis-59f97e6044a1275f83427ab7962a154c00d915b5",
-            sha256 = "5e785c25b1d57973e7481b4da226d7c73056ea22c7545bf6d14dbebf6e99b073",
+            strip_prefix = "googleapis-spanner_commit_stats",
+            sha256 = "198b878c673acbce48335b6de7575ff658106d75bb1b0616a0e0f27402a90356",
             build_file = "@com_github_googleapis_google_cloud_cpp//bazel:googleapis.BUILD",
         )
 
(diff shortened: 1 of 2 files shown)


bazel/google_cloud_cpp_deps.bzl
```diff
@@ -75,10 +75,10 @@ def google_cloud_cpp_deps():
         http_archive(
             name = "com_google_googleapis",
             urls = [
-                "https://github.com/googleapis/googleapis/archive/59f97e6044a1275f83427ab7962a154c00d915b5.tar.gz",
+                "https://github.com/googleapis/googleapis/archive/spanner_commit_stats.tar.gz",
             ],
-            strip_prefix = "googleapis-59f97e6044a1275f83427ab7962a154c00d915b5",
-            sha256 = "5e785c25b1d57973e7481b4da226d7c73056ea22c7545bf6d14dbebf6e99b073",
+            strip_prefix = "googleapis-spanner_commit_stats",
+            sha256 = "198b878c673acbce48335b6de7575ff658106d75bb1b0616a0e0f27402a90356",
```
coryan, line 81: FWIW, the SHA256 will change if they ever push a new commit to the `spanner_commit_stats` branch (or they rebase the branch, or otherwise change its contents). This would break the build for anybody who has downloaded the "preview". Not sure how likely it is that new commits appear on that branch.

I think you can omit the `sha256` setting, at a risk of builds failing because the downloaded tarball is bad (git can return something that is not a tarball if its servers are in trouble). I think Bazel does better caching if the `sha256` setting is there.

devbww: Acknowledged.  I think (at least for now) given the benefits and that a change is unlikely, that I'll leave as-is.  If I end up being wrong I will happily accept an "I told you so".  Thanks.

mr-salty (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5452#-:-MLjTCla-A9J8yq6sllX:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[bazel/google_cloud_cpp_deps.bzl, line 81 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5452#-MLfNB8LIt4LrVOq5RhW-r1-81:-MLjSxQwDK6jY6ETZZJt:bt0v4jo) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452253b1a6fca6a48c06c3460bb6f14686e17797/bazel/google_cloud_cpp_deps.bzl#L81)):*

Acknowledged.  I think (at least for now) given the benefits and that a change is unlikely, that I'll leave as-is.  If I end up being wrong I will happily accept an "I told you so".  Thanks.
</blockquote></details>

ha, I commented similar to carlos on brad's other PR.  I disabled sha256 checking in my cmek branch, although in practice they don't update it at all so it doesn't end up mattering.

PR #5453, opened 2020-11-09
test(storage): add emulator unit tests and cleanup

### LINKED ISSUE #4751: Add testbench support for GCS+gRPC

vnvo2409:
Since `Bazel` does not fully support `pytest`, I converted all those tests to `unittest`.

I have some questions
- Should we convert all the code that are using `testbench` to `emulator` ?
- When will we remove the `testbench` directory ?

Part of #4751 

We finally reach a milestone in `Complete GCS+gRPC plugin`. From now, we could implementing `GCS+gRPC` without worrying about testing 🚀

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453)

### DIFF
--- a/ci/kokoro/docker/Dockerfile.fedora
+++ b/ci/kokoro/docker/Dockerfile.fedora
@@ -19,11 +19,8 @@ FROM fedora:${DISTRO_VERSION}
 RUN dnf makecache && \
     dnf install -y ccache clang clang-tools-extra cmake diffutils findutils \
         gcc-c++ git libcxx-devel libcxxabi-devel libasan libubsan libtsan \
-        llvm make openssl-devel pkgconfig python3 python3.8 python3-devel \
-        python3-pip tar unzip wget which zlib-devel
-
-# Link `python` to `python3`
-RUN ln -s $(which python3) /usr/bin/python
+        llvm make openssl-devel pkgconfig python python3.8 python3-devel \
+        python-pip tar unzip wget which zlib-devel
 
 # Sets root's password to the empty string to enable users to get a root shell
 # inside the container with `su -` and no password. Sudo would not work because
@@ -36,7 +33,7 @@ RUN pip3 install --upgrade pip
 RUN pip3 install setuptools wheel
 RUN pip3 install git+git://github.com/googleapis/python-storage@8cf6c62a96ba3fff7e5028d931231e28e5029f1c
 RUN pip3 install flask==1.1.2 httpbin==0.7.0 scalpl==0.4.0 \
-    crc32c==2.1 gevent==20.9.0 gunicorn==20.0.4
+    crc32c==2.1 gunicorn==20.0.4
 
 # Install the Cloud SDK and some of the emulators. We use the emulators to run
 # integration tests for the client libraries.
--- a/ci/kokoro/docker/Dockerfile.fedora
+++ b/ci/kokoro/docker/Dockerfile.fedora
@@ -22,6 +22,9 @@ RUN dnf makecache && \
         llvm make openssl-devel pkgconfig python3 python3.8 python3-devel \
         python3-pip tar unzip wget which zlib-devel
 
+# Link `python` to `python3`
+RUN ln -s $(which python3) /usr/bin/python
+
 # Sets root's password to the empty string to enable users to get a root shell
 # inside the container with `su -` and no password. Sudo would not work because
 # we run these containers as the invoking user's uid, which does not exist in
@@ -32,8 +35,8 @@ RUN echo 'root:' | chpasswd
 RUN pip3 install --upgrade pip
 RUN pip3 install setuptools wheel
 RUN pip3 install git+git://github.com/googleapis/python-storage@8cf6c62a96ba3fff7e5028d931231e28e5029f1c
-RUN pip3 install grpcio==1.32.0 flask==1.1.2 pytest==6.0.2 httpbin==0.7.0 \
-    scalpl==0.4.0 crc32c==2.1 gevent==20.9.0 gunicorn==20.0.4
+RUN pip3 install flask==1.1.2 httpbin==0.7.0 scalpl==0.4.0 \
+    crc32c==2.1 gevent==20.9.0 gunicorn==20.0.4
 
 # Install the Cloud SDK and some of the emulators. We use the emulators to run
 # integration tests for the client libraries.
--- a/ci/kokoro/docker/Dockerfile.fedora-install
+++ b/ci/kokoro/docker/Dockerfile.fedora-install
@@ -26,6 +26,9 @@ RUN dnf makecache && \
         make ninja-build npm openssl-devel pkgconfig protobuf-compiler python3 \
         python3.8 python3-pip ShellCheck tar unzip w3m wget which zlib-devel
 
+# Link `python` to `python3`
+RUN ln -s $(which python3) /usr/bin/python
+
 # Sets root's password to the empty string to enable users to get a root shell
 # inside the container with `su -` and no password. Sudo would not work because
 # we run these containers as the invoking user's uid, which does not exist in
@@ -57,8 +60,8 @@ RUN pip3 install black==19.3b0
 RUN dnf makecache && dnf install -y python3-devel
 RUN pip3 install setuptools wheel
 RUN pip3 install git+git://github.com/googleapis/python-storage@8cf6c62a96ba3fff7e5028d931231e28e5029f1c
-RUN pip3 install grpcio==1.32.0 flask==1.1.2 pytest==6.0.2 httpbin==0.7.0 \
-    scalpl==0.4.0 crc32c==2.1 gevent==20.9.0 gunicorn==20.0.4
+RUN pip3 install flask==1.1.2 httpbin==0.7.0 scalpl==0.4.0 \
+    crc32c==2.1 gevent==20.9.0 gunicorn==20.0.4
 
 # Install cspell for spell checking.
 RUN npm install -g cspell
--- a/ci/kokoro/docker/Dockerfile.fedora-install
+++ b/ci/kokoro/docker/Dockerfile.fedora-install
@@ -23,11 +23,8 @@ RUN dnf makecache && \
     dnf install -y abi-compliance-checker abi-dumper ccache \
         clang clang-tools-extra cmake diffutils doxygen findutils gcc-c++ git \
         grpc-devel grpc-plugins lcov libcxx-devel libcxxabi-devel libcurl-devel \
-        make ninja-build npm openssl-devel pkgconfig protobuf-compiler python3 \
-        python3.8 python3-pip ShellCheck tar unzip w3m wget which zlib-devel
-
-# Link `python` to `python3`
-RUN ln -s $(which python3) /usr/bin/python
+        make ninja-build npm openssl-devel pkgconfig protobuf-compiler python \
+        python3.8 python-pip ShellCheck tar unzip w3m wget which zlib-devel
 
 # Sets root's password to the empty string to enable users to get a root shell
 # inside the container with `su -` and no password. Sudo would not work because
@@ -61,7 +58,7 @@ RUN dnf makecache && dnf install -y python3-devel
 RUN pip3 install setuptools wheel
 RUN pip3 install git+git://github.com/googleapis/python-storage@8cf6c62a96ba3fff7e5028d931231e28e5029f1c
 RUN pip3 install flask==1.1.2 httpbin==0.7.0 scalpl==0.4.0 \
-    crc32c==2.1 gevent==20.9.0 gunicorn==20.0.4
+    crc32c==2.1 gunicorn==20.0.4
 
 # Install cspell for spell checking.
 RUN npm install -g cspell
--- a/ci/kokoro/docker/Dockerfile.fedora-libcxx-msan
+++ b/ci/kokoro/docker/Dockerfile.fedora-libcxx-msan
@@ -19,11 +19,8 @@ FROM fedora:${DISTRO_VERSION}
 # then compile our code.
 RUN dnf makecache && \
     dnf install -y ccache clang clang-tools-extra cmake findutils gcc-c++ \
-        git llvm llvm-devel make ninja-build openssl-devel python3 python3.8 \
-        python3-devel python3-lit python3-pip tar unzip which wget xz
-
-# Link `python` to `python3`
-RUN ln -s $(which python3) /usr/bin/python
+        git llvm llvm-devel make ninja-build openssl-devel python python3.8 \
+        python3-devel python3-lit python-pip tar unzip which wget xz
 
 # Sets root's password to the empty string to enable users to get a root shell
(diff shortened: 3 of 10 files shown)


coryan (changes requested):
> Since Bazel does not fully support pytest, I converted all those tests to unittest.

Thanks.

> I have some questions
> 
> Should we convert all the code that are using testbench to emulator ?

Seems to me all the `google-cloud-cpp` integration tests are already doing so.  What else did you have in mind?

> When will we remove the testbench directory ?

I think we should open a bug to track these things, but I believe we need to move and/or create a new README.md file like then one in the `testbench/` directory.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-:-MLhWqmE949P5ygfcwTk:bv0728b)* status: all files reviewed, 1 unresolved discussion (waiting on @vnvo2409)

___
*[google/cloud/storage/emulator/BUILD, line 15 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-MLhWBNE62Bihl0oKpv3:-MLhWBNF81_q9knfm0bZ:bbm8szr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f905c74458a9cafc886007f596035638490ce6ae/google/cloud/storage/emulator/BUILD#L15)):*
> ```Starlark
> # limitations under the License.
> 
> py_test(
> ```

You need to add some boilerplate I think:

```
package(default_visibility = ["//visibility:public"])

```

vnvo2409:
The CI threw `/usr/bin/env: 'python': No such file or directory`. Do you have any idea why ? @coryan

coryan:
Searching around I found this:

https://github.com/bazelbuild/bazel/issues/8665

We do not install a plain "python" program in the docker images for these builds:

```sh
./ci/kokoro/docker/build.sh asan bash
bash-5.0$ python 
bash: python: command not found
bash-5.0$ python3 --version
Python 3.9.0
```


There is already an open bug for Bazel (https://github.com/bazelbuild/bazel/issues/8446), but I think we will need to workaround it by installing a plain python program.

remyabel:
FYI Ubuntu only ships `/usr/bin/python3` but not `/usr/bin/python`. On Fedora however, `/usr/bin/python` is symlinked to `/usr/bin/python3`. Focal has this [package](https://packages.ubuntu.com/focal/python-is-python3), but I'm not sure how safe the symlink would be on Bionic. WDYT?

coryan:
This was Fedora:33, and I am not seeing the `/usr/bin/python` link:

```
docker run --rm -it fedora:33 bash
# dnf makecache && dnf install python3
# ls -l /usr/bin/py*
lrwxrwxrwx 1 root root     8 Oct  6 14:19 /usr/bin/pydoc -> ./pydoc3
lrwxrwxrwx 1 root root     8 Oct  6 14:19 /usr/bin/pydoc3 -> pydoc3.9
-rwxr-xr-x 1 root root    78 Oct  6 14:19 /usr/bin/pydoc3.9
lrwxrwxrwx 1 root root     9 Oct  6 14:19 /usr/bin/python3 -> python3.9
-rwxr-xr-x 1 root root 15536 Oct  6 14:20 /usr/bin/python3.9
```

Seems like we need to install `python` (aka `python-unversioned-command`) for Fedora:33:

```
# dnf makecache && dnf install python
# ls -l /usr/bin/py*
lrwxrwxrwx 1 root root     8 Oct  6 14:19 /usr/bin/pydoc -> ./pydoc3
lrwxrwxrwx 1 root root     8 Oct  6 14:19 /usr/bin/pydoc3 -> pydoc3.9
-rwxr-xr-x 1 root root    78 Oct  6 14:19 /usr/bin/pydoc3.9
lrwxrwxrwx 1 root root     9 Oct  6 14:46 /usr/bin/python -> ./python3
lrwxrwxrwx 1 root root     9 Oct  6 14:19 /usr/bin/python3 -> python3.9
-rwxr-xr-x 1 root root 15536 Oct  6 14:20 /usr/bin/python3.9

```

remyabel:
> Windows failed because of another reason: Those tests do not support to run on Windows. I have no idea why it happened 

It doesn't look like the python dependencies are being installed for the Windows builds. I think the Windows build only ran the integration tests (and not the testbench/emulator tests) but since you run `bazel test` inside `run_integration_tests_emulator_bazel.sh ` they're getting picked up in the Windows build. Carlos can correct me on this.

> Seems like we need to install python (aka python-unversioned-command) for Fedora:33:

Thanks. It should be [installed by default](https://fedoraproject.org/wiki/Changes/Python_means_Python3), odd that the docker image omits it.

remyabel:
> but since you run bazel test inside run_integration_tests_emulator_bazel.sh they're getting picked up in the Windows build.

Please disregard this. It made sense in my head when I typed it.

coryan:
> Windows failed because of another reason: Those tests do not support to run on Windows. I have no idea why it happened confused

We basically run `bazel test ...` on Windows (and Linux for that matter), so *all* the unit tests try to run.

Looks like we just need to install the `pip` dependencies and things would work, seems like an easy change in `build.bat`, but I really do not have the time to investigate all the details right now.

vnvo2409:
> We basically run `bazel test ...` on Windows (and Linux for that matter), so all the unit tests try to run.

I added `tags=["manual"]` to the `BUILD` file so the tests won't be caught by `...`

> The CI threw /usr/bin/env: 'python': No such file or directory.

Use `RUN ln -s $(which python3) /usr/bin/python` seems to be the easiest way.

I also deleted some redundant (`grpcio`) or unused (`pytest`) dependencies of python in the last commit.

coryan (changes requested):
> Reviewed 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453)* status: all files reviewed, 5 unresolved discussions (waiting on @vnvo2409)

___
*[ci/kokoro/docker/Dockerfile.fedora, line 26 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-MLmuNSU6YUiL-PYAf9j:-MLmuNSU6YUiL-PYAf9k:bjvj2r0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f353c80333c1de4ecc805d3fb25cc6d907e8d6e8/ci/kokoro/docker/Dockerfile.fedora#L26)):*
> ```
> 
> # Link `python` to `python3`
> RUN ln -s $(which python3) /usr/bin/python
> ```

I think we should simply install the `python` package. These Dockefiles are also good documentation to setup a development environment, and we should not be recommending something like this.  Also, seems that this could fail unexpectedly in the future (say because the link is automatically created)
___
*[ci/kokoro/docker/Dockerfile.ubuntu, line 52 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-MLmuqvfBfpekDI7WESJ:-MLmuqvfBfpekDI7WESK:bs4q9op) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f353c80333c1de4ecc805d3fb25cc6d907e8d6e8/ci/kokoro/docker/Dockerfile.ubuntu#L52)):*
> ```
> 
> # Link `python` to `python3`
> RUN ln -s $(which python3) /usr/bin/python
> ```

Ditto.
___
*[google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh, line 34 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-MLmuvAl0dlOaqkne8A2:-MLmuvAl0dlOaqkne8A3:b-zc8kmu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f353c80333c1de4ecc805d3fb25cc6d907e8d6e8/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh#L34)):*
> ```Shell
> 
> # Run the unittests of the emulator before running integration tests.
> "${BAZEL_BIN}" test "${bazel_test_args[@]}" "//google/cloud/storage/emulator:test_utils" \
> ```

Can we use  `:all` and maybe pass `--test_tag_filters=manual`?
___
*[google/cloud/storage/emulator/BUILD, line 15 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-MLhWBNE62Bihl0oKpv3:-MLmv64cCCETVJY-0WQZ:b-gy6gh0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f905c74458a9cafc886007f596035638490ce6ae/google/cloud/storage/emulator/BUILD#L15)):*

You need to add some boilerplate I think:

```
package(default_visibility = ["//visibility:public"])

```
</blockquote></details>

This is still missing I believe.
___
*[google/cloud/storage/emulator/BUILD, line 21 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-MLmv9jd1DDQIQbQkya0:-MLmv9jd1DDQIQbQkya1:bjrnt8r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f353c80333c1de4ecc805d3fb25cc6d907e8d6e8/google/cloud/storage/emulator/BUILD#L21)):*
> ```Starlark
>     imports = ["."],
>     main = "tests/test_gcs.py",
>     tags = ["manual"],
> ```

Can you open a bug to make these unit tests run on Windows?  And add a `TODO(#...) - $why` here?

coryan:
+kokoro:run

vnvo2409:
> Can you open a bug to make these unit tests run on Windows? And add a `TODO(#...) - $why` here?

These tests could run perfectly on Windows. However, I feel it's noop if we run the tests without running the emulator. WDYT ?

coryan:
> > Can you open a bug to make these unit tests run on Windows? And add a `TODO(#...) - $why` here?
> 
> These tests could run perfectly on Windows. However, I feel it's noop if we run the tests without running the emulator. WDYT ?

Yes. I was thinking we would eventually run the Windows CI build against the emulator too.

coryan (commented):
Looks good, let's see what happens with the builds +kokoro:run 

> Reviewed 8 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-:-MLpIIKiAEZQBXr07u0a:b-lyoyr0)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[google/cloud/storage/emulator/BUILD, line 21 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-MLmv9jd1DDQIQbQkya0:-MLpIF7o8YQGHMI_0k_l:b-uxtvcj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f353c80333c1de4ecc805d3fb25cc6d907e8d6e8/google/cloud/storage/emulator/BUILD#L21)):*

Can you open a bug to make these unit tests run on Windows?  And add a `TODO(#...) - $why` here?
</blockquote></details>

Oh cool, you fix that in this PR, thanks!

vnvo2409:
`kokoro/macos/bazel` failed due to the same reason as Windows. I would like to ask if we have to rebuild the `Dockerfile.*` each `kokoro` build or not ? If yes, I think we should move 
```
RUN pip3 install git+git://github.com/googleapis/python-storage@8cf6c62a96ba3fff7e5028d931231e28e5029f1c
RUN pip3 install grpcio==1.32.0 flask==1.1.2 pytest==6.0.2 httpbin==0.7.0 \
     scalpl==0.4.0 crc32c==2.1 gevent==20.9.0 gunicorn==20.0.4
```
to `storage/ci/run_integration_tests_emulator_[bazel/cmake].sh`. It has some benefits:
- we don't have to rewrite these commands for each `Dockerfile.*`
- the dependencies will be installed only when we run the storage integration tests.
- fix `kokoro/macos/bazel` CI 
- save some quota/resources

I will move the unit tests of the emulator to `storage/ci/run_integration_tests_emulator_[bazel/cmake].sh` so we don't have to run the unit tests without running the emulator ( which irritates me )

coryan (commented):
> In addition, I think we should leave Window as-is and file an issue for activating the emulator and running the unit tests at the same time which makes more sense to me.

That SGTM.

>  I would like to ask if we have to rebuild the Dockerfile.* each kokoro build or not ?

Yes, but most builds use cached versions of the docker images and refresh/rebuild the image starting from that cache. 

> to `storage/ci/run_integration_tests_emulator_[bazel/cmake].sh`

I would prefer if we refactor these commands to something like

https://github.com/googleapis/google-cloud-cpp/blob/master/ci/install-cloud-sdk.sh
 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-:-MLxvi6e6AGtCkC9g-xT:b-q6oiow)* status: :shipit: complete! all files reviewed, all discussions resolved

vnvo2409:
> Yes, but most builds use cached versions of the docker images and refresh/rebuild the image starting from that cache.

I am a little bit confused: the dependencies from `pip` are cached with the `Dockerfile.*` or they will be reinstalled in every kokoro run ?

coryan:
> > Yes, but most builds use cached versions of the docker images and refresh/rebuild the image starting from that cache.
> 
> I am a little bit confused: the dependencies from `pip` are cached with the `Dockerfile.*` or they will be reinstalled in every kokoro run ?

Yes 😁 

The builds first download a (cached) docker image created from the  `Dockerfile.*` used in that build, let say `Dockerfile.fedora-install` for this example.  Then they run `docker build -f Dockerfile.fedora-install`, but use the downloaded image as a cache source.

If nothing has changed in `Dockerfile.fedora-install` then the cache satisfies the image and: (a) the `docker build` step is very fast (as it uses the cached image), and (b) the source of the docker image cache is more reliable than downloading N dependencies from M sources (in practice).

If there are changes in `Dockerfile.fedora-install` then some of the steps may still use the cache (which is nice), and the changes are validated as part of the PR (which is important).

The cached docker images are updated as part of the full builds post-PR, which are not required to be fast (though it is nice if they are).

vnvo2409:
After a lot of trials, I think `tags=["manual"]` works best with our needs. Because:
- Without the tag, these tests will run in all CI builds which use bazel ( it isn't what we want because these tests does not belong to the unit tests of the library ).
- We can control when we run the tests, on which platform.

> Can we use :all and maybe pass --test_tag_filters=manual?

It does not work, we have to write all the target manually.

Even after #5498 is closed, we should continue using this tag since it allows us to run the tests only when we need

coryan (approved):
:lgtm: just waiting for the builds.  This is great news.

> Reviewed 1 of 3 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5453#-:-MM1zEjA640pFItEVEJc:b-f6wf3r)* status: :shipit: complete! all files reviewed, all discussions resolved

coryan:
Thanks again for this PR.

Issue #5460, opened 2020-11-09
Cached results can mask integration tests failures


mr-salty:
This issue was actually caused by #5445 but as @coryan noted in #5456 the CI tests continued to use the cached results.

I will disable this test temporarily until we can fix it.

build failure:

```
[ RUN      ] ObjectResumableWriteIntegrationTest.WriteResumeFinalizedUpload
google/cloud/storage/tests/object_resumable_write_integration_test.cc:221: Failure
Value of: os.IsOpen()
  Actual: true
Expected: false
google/cloud/storage/tests/object_resumable_write_integration_test.cc:223: Failure
Value of: os.metadata()
Expected: is OK
Actual: default [UNKNOWN]
[  FAILED  ] ObjectResumableWriteIntegrationTest.WriteResumeFinalizedUpload (908 ms)
```

coryan:
After looking at it some more, it is worse and better: it is a problem with cached results, but the build in question uses **production** (no the emulator) for the storage tests.  The successful runs are indeed cached, but I now suspects the service changed behavior.

vnvo2409:
> but the build in question uses production (no the emulator) for the storage tests

So I suppose there is nothing to do with the new emulator.

> The successful runs are indeed cached

Bazel has an option [`--cache_test_results`](https://docs.bazel.build/versions/master/user-manual.html#flag--cache_test_results). I think we could force `--cache_test_results=no` in an integration test and a production test ( Maybe occasionally only, like 1 specific day in every month, with a little `bash` trick )

> build failure:
```
[ RUN      ] ObjectResumableWriteIntegrationTest.WriteResumeFinalizedUpload
google/cloud/storage/tests/object_resumable_write_integration_test.cc:221: Failure
Value of: os.IsOpen()
  Actual: true
Expected: false
google/cloud/storage/tests/object_resumable_write_integration_test.cc:223: Failure
Value of: os.metadata()
Expected: is OK
Actual: default [UNKNOWN]
[  FAILED  ] ObjectResumableWriteIntegrationTest.WriteResumeFinalizedUpload (908 ms)
```
I've already implemented a feature for dumping the `testbench.log` if there is any failed test. So please add the link to the `build.log` so we could have a better insight about what happend inside the emulator ( I think we do not need it for this issue though).

mr-salty:
[...]

PR #5461, opened 2020-11-09
test: disable broken WriteResumeFinalizedUpload test

mr-salty:
fixes the CI build until #5460 can be resolved.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5461)


devbww (approved):

coryan (approved):
TIL the `DISABLED_` trick.

coryan (commented):
I am fine with the change as-is, but I think 

```cc
if (UsingGrpc()) GTEST_SKIP();
```

Would also fix the problem *and* would continue to test with the REST API.

mr-salty:
I think that's a good idea, although since this is ready to check in let me do that to fix the CI and send another change with your suggestion.

PR #5464, opened 2020-11-10
feat(pubsub): batch acks and nacks in streams

coryan:
It is more efficient to send a large `StreamingPullRequest` protos in
the `Write()` calls. Saves in CPU and (more importantly) I/O. The
batching should have no impact to applications with low throughput, as
we only batch for 100ms and the typical deadlines are 100x larger.

This improved throughput by over 50%, at least in my local tests.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5464)

### DIFF
--- a/google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
+++ b/google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
@@ -359,7 +359,7 @@ void StreamingSubscriptionBatchSource::StartWriteTimer() {
 void StreamingSubscriptionBatchSource::OnWriteTimer(Status const& s) {
   if (!s.ok()) return;
   std::unique_lock<std::mutex> lk(mu_);
-  DrainQueues(std::move(lk), 1);
+  DrainQueues(std::move(lk), true);
   StartWriteTimer();
 }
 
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
```diff
@@ -339,12 +342,27 @@ void StreamingSubscriptionBatchSource::OnWrite(bool ok) {
...
+  using F = future<StatusOr<std::chrono::system_clock::time_point>>;
+  cq_.MakeRelativeTimer(ack_batching_config_.max_hold_time).then([weak](F f) {
+    if (auto self = weak.lock()) self->OnWriteTimer(f.get().status());
+  });
+}
+
+void StreamingSubscriptionBatchSource::OnWriteTimer(Status const& s) {
+  if (!s.ok()) return;
+  std::unique_lock<std::mutex> lk(mu_);
+  DrainQueues(std::move(lk), 1);
```
devjgm, line 362: `1` -> `true`?

devjgm (approved):

google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
coryan: Yup.

Issue #5470, opened 2020-11-11
Change the update-abi build to only update the baselines that actually changed


coryan:
See the comments in #5469

PR #5472, opened 2020-11-11
ci: only update ABI if there were changes

remyabel:
Fixes: #5470

`gzip` files are binary and the default diff tool will show them as different
even if they are "identical". This uses `zdiff` to compare them.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5472)


coryan (approved):
I think this is a an improvement, though maybe it would accept diffs that are not material.  +kokoro:run 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5472#-:-MLrMSCM6pqNu60nF0zb:b-2lwja1)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #4132, opened 2020-05-15
Need a CI build that detects incorrectly installed .so files


devjgm:
We had a bug report (https://github.com/googleapis/google-cloud-cpp/issues/4128) that pointed out that the grpc_utils .so was being installed without a version. The issue was that we were using the wrong variable in the `google/cloud/CMakeLists.txt` file (fix PR: https://github.com/googleapis/google-cloud-cpp/pull/4131).

We should have a CI build to detect when/if this happens. We perhaps need a new `ci/kokoro/install` build to detect these cases.

coryan:
See https://github.com/googleapis/google-cloud-cpp/issues/4265#issuecomment-634147864 it seems like improperly installed `.so` files still "work".  That suggests to me that no build checking for breakage will do the trick.

But we could add a step to the CI scripts that searches for files with unexpected extensions. In principle we only install files matching a few patterns, like with `.*\.h`, `.*\.inc`, `.*\.proto`, `.*\.cmake`, `.*\.pc`, `lib.*\.a$`, `lib.*\.so`, `lib.*\..so\.${GOOGLE_CLOUD_CPP_VERSION_MAJOR}`, `lib.*\..so\.${GOOGLE_CLOUD_CPP_VERSION}`

PR #5488, opened 2020-11-12
ci: detect incorrectly installed files

remyabel:
Fixes: #4132

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -424,6 +424,31 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
     /bin/false
   fi
 
+  io::log_yellow "Verify no extraneous files were installed."
+  # This script is only run on a Fedora-based build, so it's fine that this
+  # setting is not portable.
+  export PKG_CONFIG_PATH="/var/tmp/staging/lib64/pkgconfig:/usr/local/lib64/pkgconfig"
+
+  # Get the version of one of the libraries. These should all be the same, so
+  # it does not matter what we use.
+  GOOGLE_CLOUD_CPP_VERSION=$(pkg-config storage_client --modversion) || true
+  export GOOGLE_CLOUD_CPP_VERSION
+  export GOOGLE_CLOUD_CPP_VERSION_MAJOR="${GOOGLE_CLOUD_CPP_VERSION:0:1}"
+
+  # @E is a parameter substitution that provides an escaped version of the
+  # string, i.e 1.21.0 functionally becomes 1\.21\.0.
+  export GOOGLE_CLOUD_CPP_VERSION_ESCAPED="${GOOGLE_CLOUD_CPP_VERSION@E}"
+
+  files=$(find /var/tmp/staging/ -type f | grep -vE \
+    "\.(h|inc|proto|cmake|pc|a|so|so\.${GOOGLE_CLOUD_CPP_VERSION_ESCAPED}|so\.${GOOGLE_CLOUD_CPP_VERSION_MAJOR})\$")
+  export files
+  if [[ "${#files[@]}" -gt 0 ]]; then
+    io::log_red "Installed files do not match expectation."
+    echo "Found:"
+    echo "${files[@]}"
+    /bin/false
+  fi
+
   if [[ "${CHECK_ABI:-}" == "yes" ]]; then
     io::log_yellow "Checking ABI compatibility."
     libraries=(
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -425,9 +425,7 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
   fi
 
   io::log_yellow "Verify no extraneous files were installed."
-  # This script is only run on a Fedora-based build, so it's fine that this
-  # setting is not portable.
-  export PKG_CONFIG_PATH="/var/tmp/staging/lib64/pkgconfig:/usr/local/lib64/pkgconfig"
+  export PKG_CONFIG_PATH="/var/tmp/staging/lib64/pkgconfig:${PKG_CONFIG_PATH:-}"
 
   # Get the version of one of the libraries. These should all be the same, so
   # it does not matter what we use.
@@ -435,7 +433,8 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
   GOOGLE_CLOUD_CPP_VERSION_MAJOR=$(echo "${GOOGLE_CLOUD_CPP_VERSION}" | cut -d'.' -f1)
 
   # @E is a parameter substitution that provides an escaped version of the
-  # string, i.e 1.21.0 functionally becomes 1\.21\.0.
+  # string, i.e 1.21.0 functionally becomes 1\.21\.0. Otherwise, the regex
+  # match below will fail.
   GOOGLE_CLOUD_CPP_VERSION_ESCAPED="${GOOGLE_CLOUD_CPP_VERSION@E}"
 
   mapfile -t files < <(find /var/tmp/staging/ -type f | grep -vE \
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -432,16 +432,14 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
   # Get the version of one of the libraries. These should all be the same, so
   # it does not matter what we use.
   GOOGLE_CLOUD_CPP_VERSION=$(pkg-config storage_client --modversion) || true
-  export GOOGLE_CLOUD_CPP_VERSION
-  export GOOGLE_CLOUD_CPP_VERSION_MAJOR="${GOOGLE_CLOUD_CPP_VERSION:0:1}"
+  GOOGLE_CLOUD_CPP_VERSION_MAJOR=$(echo "${GOOGLE_CLOUD_CPP_VERSION}" | cut -d'.' -f1)
 
   # @E is a parameter substitution that provides an escaped version of the
   # string, i.e 1.21.0 functionally becomes 1\.21\.0.
-  export GOOGLE_CLOUD_CPP_VERSION_ESCAPED="${GOOGLE_CLOUD_CPP_VERSION@E}"
+  GOOGLE_CLOUD_CPP_VERSION_ESCAPED="${GOOGLE_CLOUD_CPP_VERSION@E}"
 
-  files=$(find /var/tmp/staging/ -type f | grep -vE \
+  mapfile -t files < <(find /var/tmp/staging/ -type f | grep -vE \
     "\.(h|inc|proto|cmake|pc|a|so|so\.${GOOGLE_CLOUD_CPP_VERSION_ESCAPED}|so\.${GOOGLE_CLOUD_CPP_VERSION_MAJOR})\$")
-  export files
   if [[ "${#files[@]}" -gt 0 ]]; then
     io::log_red "Installed files do not match expectation."
     echo "Found:"
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -432,13 +432,8 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
   GOOGLE_CLOUD_CPP_VERSION=$(pkg-config storage_client --modversion) || true
   GOOGLE_CLOUD_CPP_VERSION_MAJOR=$(echo "${GOOGLE_CLOUD_CPP_VERSION}" | cut -d'.' -f1)
 
-  # @E is a parameter substitution that provides an escaped version of the
-  # string, i.e 1.21.0 functionally becomes 1\.21\.0. Otherwise, the regex
-  # match below will fail.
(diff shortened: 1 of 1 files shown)


mr-salty (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488)* status: all files reviewed, 4 unresolved discussions (waiting on @remyabel)

___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 428 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLxwZALE-_jS3AxNUl2:-MLxwXN9DP2x5FsxVtsa:b-5c24ab) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/036951d243798902aaf5d3e307c3b770e9ec0c31/ci/kokoro/docker/build-in-docker-cmake.sh#L428)):*
> ```Shell
> 
>   io::log_yellow "Verify no extraneous files were installed."
>   # This script is only run on a Fedora-based build, so it's fine that this
> ```

I see several ubuntu builds that use this script like [cmake-super](https://github.com/googleapis/google-cloud-cpp/blob/c44f8719fe110844e45d81ef6eb08aa820900f00/ci/kokoro/docker/build.sh#L165-L172)

can we obtain this via other means?
___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 435 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLxzVcn7UTqy-RUFGTQ:-MLxzVcn7UTqy-RUFGTR:b-ar5gcq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/036951d243798902aaf5d3e307c3b770e9ec0c31/ci/kokoro/docker/build-in-docker-cmake.sh#L435)):*
> ```Shell
>   # it does not matter what we use.
>   GOOGLE_CLOUD_CPP_VERSION=$(pkg-config storage_client --modversion) || true
>   export GOOGLE_CLOUD_CPP_VERSION
> ```

I believe you can strike all the `export` directives (aside from `PKG_CONFIG_PATH` above); we're only using these variables in this shell.
___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 436 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLxy2mH9s43B_KP0xhu:-MLxy2mI6PqZ1ATE0ghl:b-thq8b7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/036951d243798902aaf5d3e307c3b770e9ec0c31/ci/kokoro/docker/build-in-docker-cmake.sh#L436)):*
> ```Shell
>   GOOGLE_CLOUD_CPP_VERSION=$(pkg-config storage_client --modversion) || true
>   export GOOGLE_CLOUD_CPP_VERSION
>   export GOOGLE_CLOUD_CPP_VERSION_MAJOR="${GOOGLE_CLOUD_CPP_VERSION:0:1}"
> ```

this will break if the major version reaches 10
___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 444 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLy-5R2C3Z5LPQHbzIF:-MLy-5R2C3Z5LPQHbzIG:bnzsspz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/036951d243798902aaf5d3e307c3b770e9ec0c31/ci/kokoro/docker/build-in-docker-cmake.sh#L444)):*
> ```Shell
>   files=$(find /var/tmp/staging/ -type f | grep -vE \
>     "\.(h|inc|proto|cmake|pc|a|so|so\.${GOOGLE_CLOUD_CPP_VERSION_ESCAPED}|so\.${GOOGLE_CLOUD_CPP_VERSION_MAJOR})\$")
>   export files
> ```

as above; `export` is not required.

remyabel (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @mr-salty)

___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 428 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLxwZALE-_jS3AxNUl2:-MLy6pXu39myOeRLHSKc:bsl5a4w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/036951d243798902aaf5d3e307c3b770e9ec0c31/ci/kokoro/docker/build-in-docker-cmake.sh#L428)):*

I see several ubuntu builds that use this script like [cmake-super](https://github.com/googleapis/google-cloud-cpp/blob/c44f8719fe110844e45d81ef6eb08aa820900f00/ci/kokoro/docker/build.sh#L165-L172)

can we obtain this via other means?
</blockquote></details>

This block of code only runs if `TEST_INSTALL` is true which only runs on the Fedora builds. Let me know if you see otherwise, though.
___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 435 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLxzVcn7UTqy-RUFGTQ:-MLy6wOu2MjYeSIgUG0Y:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/036951d243798902aaf5d3e307c3b770e9ec0c31/ci/kokoro/docker/build-in-docker-cmake.sh#L435)):*

I believe you can strike all the `export` directives (aside from `PKG_CONFIG_PATH` above); we're only using these variables in this shell.
</blockquote></details>

Done.
___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 436 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLxy2mH9s43B_KP0xhu:-MLy6wjO6Ys5CLRqiZlQ:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/036951d243798902aaf5d3e307c3b770e9ec0c31/ci/kokoro/docker/build-in-docker-cmake.sh#L436)):*

this will break if the major version reaches 10
</blockquote></details>

Done.
___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 444 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLy-5R2C3Z5LPQHbzIF:-MLy6xAc-Y1CBGRUTNks:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/036951d243798902aaf5d3e307c3b770e9ec0c31/ci/kokoro/docker/build-in-docker-cmake.sh#L444)):*

as above; `export` is not required.
</blockquote></details>

Done.

ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -424,6 +424,29 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
     /bin/false
   fi
 
+  io::log_yellow "Verify no extraneous files were installed."
+  # This script is only run on a Fedora-based build, so it's fine that this
+  # setting is not portable.
+  export PKG_CONFIG_PATH="/var/tmp/staging/lib64/pkgconfig:/usr/local/lib64/pkgconfig"
```
devjgm, line 430: This script is run from `ubuntu` images also, for example the `clang-3.8` build 

https://github.com/googleapis/google-cloud-cpp/blob/4157e16a81c8a2c8a56af16c1e3f1f5b96a46bb9/ci/kokoro/docker/build.sh#L212-L219

Would this setting become more portable if we simply pre-pended the desired path to whatever this environment variable already was? Like:

```shell
export PKG_CONFIG_PATH="/var/tmp/staging/lib64/pkgconfig:${PKG_CONFIG_PATH:-}"
```

WDYT?

ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -424,6 +424,29 @@ if [[ "${TEST_INSTALL:-}" = "yes" ]]; then
...
+
+  # Get the version of one of the libraries. These should all be the same, so
+  # it does not matter what we use.
+  GOOGLE_CLOUD_CPP_VERSION=$(pkg-config storage_client --modversion) || true
+  GOOGLE_CLOUD_CPP_VERSION_MAJOR=$(echo "${GOOGLE_CLOUD_CPP_VERSION}" | cut -d'.' -f1)
+
+  # @E is a parameter substitution that provides an escaped version of the
+  # string, i.e 1.21.0 functionally becomes 1\.21\.0.
+  GOOGLE_CLOUD_CPP_VERSION_ESCAPED="${GOOGLE_CLOUD_CPP_VERSION@E}"
```
devjgm, line 439: TIL about `@E` -- cool thanks! But can you explain a bit more about why it's needed here? I was playing w/ it in the shell and I couldn't figure out how it did anything interesting with backslashes.

ci/kokoro/docker/build-in-docker-cmake.sh
remyabel: See my comment above in response to Todd. I checked the kokoro logs for the `cmake-super` build as well. With that being I said, I do think the code is an improvement so I added it. We'll need to make similar changes to `check-abi.sh` since that's where I copied the code from.

ci/kokoro/docker/build-in-docker-cmake.sh
remyabel: Done. When testing, I didn't really see any "changes" either but it's necessary to ensure it's escaped properly in the regex. i.e, `1.21.0` doesn't provide the correct behavior versus `1\.21\.0`.

ci/kokoro/docker/build-in-docker-cmake.sh
devjgm: SGTM. Sorry for missing that you and @mr-salty had already discussed this.

ci/kokoro/docker/build-in-docker-cmake.sh
remyabel: Actually I decided to delete it. I tested it again and see no difference. I think I added it while debugging and thought it was needed but it's not.

devjgm: OK. Maybe I'm holding it wrong, but I'm not seeing that `@E` does anything.

```shell
$ echo $BASH_VERSION
5.1.0(1)-rc1
$ FOO="1.2.3"
$ echo 1x2x3 | grep -E "${FOO@E}"  # If this escaped to 1\.2\.3, then the pattern should not match
1x2x3
$
```

Do you see what I'm doing wrong?

devjgm (approved):
Great. Thanks for this! I'm not sure if @mr-salty has any further comments, but this LGTM!

ci/kokoro/docker/build-in-docker-cmake.sh
remyabel: My apologies for the misunderstanding. The bug I encountered was fixed by a different section of code and the parameter substitution was not needed. Removed.

devbww: `@E` doesn't do escaping ... it does expansion of backslash encodings.  For example,
```
$ FOO='\b'
$ echo ${FOO} | od -c
0000000   \   b  \n
0000003
$ echo ${FOO@E} | od -c
0000000  \b  \n
0000002
```

remyabel: Yes I completely misunderstood how the feature worked. Thanks for the clarification.

mr-salty (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-:-MLypg9F4J0_vyvm_lpO:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @devjgm and @mr-salty)

___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 430 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLvoLsrlpx6jjfjtQGj-r2-430:-MLypAu2-NSvUewntXR0:b6gpe11) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bf4d46c4efc8697b3b4c8a75d95ee8a91100eb7a/ci/kokoro/docker/build-in-docker-cmake.sh#L430)):*

SGTM. Sorry for missing that you and @mr-salty had already discussed this.
</blockquote></details>

yeah, I like this better, thanks.
___
*[ci/kokoro/docker/build-in-docker-cmake.sh, line 435 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-MLyp0UL71C11PoN7wk9:-MLyp0UL71C11PoN7wkA:b-cjshvl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3296c3d1d8cd5a8ae519f48a0fb31522455f8ed/ci/kokoro/docker/build-in-docker-cmake.sh#L435)):*
> ```Shell
>   GOOGLE_CLOUD_CPP_VERSION_MAJOR=$(echo "${GOOGLE_CLOUD_CPP_VERSION}" | cut -d'.' -f1)
> 
>   mapfile -t files < <(find /var/tmp/staging/ -type f | grep -vE \
> ```

wow, i learned 2 new things from one line of code.

mr-salty:
not sure why this got stuck, let me try to kick it...

mr-salty (approved):
:lgtm: IDK why the CI build got stuck but it passed after rerun

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5488#-:-MMJCVd7CZlnI-kxPC5H:bojnysc)* status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)

PR #5518, opened 2020-11-18
cleanup(storage): rename `testbench` to `emulator`

### LINKED ISSUE #4751: Add testbench support for GCS+gRPC

vnvo2409:
0aa076d: Rename `UsingTestbench` to `UsingEmulator` in `google/cloud/storage/`
2b0e63b: enable test against emulator for `TEST_F(ObjectFileIntegrationTest, UploadFileBinary)`
Other commits are self-explanatory

Finally, close #4751 🎉

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518)


remyabel:
Does not need to be done in this PR, but I think the dependencies in [setup-development-environment.md](https://github.com/googleapis/google-cloud-cpp/blob/master/doc/setup-development-environment.md) need to be updated as well.

coryan (changes requested):
:confetti_ball: 

This is awesome news.  I am a bit concerned about changes to the "public" interface.  I am going to discuss with my colleagues, I suspect we will need to change some things in a backwards compatible way.

+kokoro:run 

> Reviewed 69 of 69 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-:-MMQYhbO6nHKLNT26JTF:b-xajgz)* status: all files reviewed, 5 unresolved discussions (waiting on @vnvo2409)

___
*[google/cloud/storage/client_options.h, line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMQZ327DQet4g7S3Huw:-MMQZ327DQet4g7S3Hux:b-714jkd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dac65fe2c3c1466e2289121ae7289fdb87df002e/google/cloud/storage/client_options.h#L63)):*
> ```objc
>  * By default, several environment variables are read to configure the client:
>  *
>  * - `CLOUD_STORAGE_EMULATOR_ENDPOINT`: if set, use this http endpoint to
> ```

I think this would make the change a breaking change, and I am not sure we need to do it.  I mean, the naming will be a bit awkward without this change, but OTOH, I am not sure we need to break any build scripts / tests that application developers may have created.

To be clear, I am not sure this is a blocker, but I am not sure it is not either. Let me chat with some folks here, including @devjgm .
___
*[google/cloud/storage/doc/storage-main.dox, line 61 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMQZtFB-s9VZTHPMYku:-MMQZtFB-s9VZTHPMYkv:b-despt6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dac65fe2c3c1466e2289121ae7289fdb87df002e/google/cloud/storage/doc/storage-main.dox#L61)):*
> ```
> - `CLOUD_STORAGE_ENABLE_TRACING=raw-client,http` enables all logging.
> 
> - `CLOUD_STORAGE_EMULATOR_ENDPOINT=...` override the default endpoint used by
> ```

Putting a pin here to think about backwards compatibility.
___
*[google/cloud/storage/emulator/emulator.py, line 680 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMQ_-PU9Nv8X1V29TSG:-MMQ_-PU9Nv8X1V29TSH:bts4k8f) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dac65fe2c3c1466e2289121ae7289fdb87df002e/google/cloud/storage/emulator/emulator.py#L680)):*
> ```Python
>                     None,
>                 )
>                 blob.metadata.metadata["x_emulator_transfer_encoding"] = ":".join(
> ```

This is also something that application developers may be using in their own testing.
___
*[google/cloud/storage/emulator/README.md, line 45 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMQ_KClA2sCT_PpGyhw:-MMQ_KClA2sCT_PpGyhx:b6zy8l2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dac65fe2c3c1466e2289121ae7289fdb87df002e/google/cloud/storage/emulator/README.md#L45)):*
> ```Markdown
> ### return-broken-stream
> 
> Set request headers with `x-goog-emulator-instructions: return-broken-stream`.
> ```

I know folks in other libraries (notably golang, maybe Python and Java too) use these headers to do their own testing.  I think we should not break their code.  What if we supported both `x-goo-emulator-instructions` and `x-goog-testbench-instructions` with the same semantics?
___
*[google/cloud/storage/tools/run_testbench_utils.sh, line 19 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMQaIlPAiTjvsMHpsZ1:-MMQaIlPAiTjvsMHpsZ2:b-9f468u) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dac65fe2c3c1466e2289121ae7289fdb87df002e/google/cloud/storage/tools/run_testbench_utils.sh#L19)):*
> ```Shell
> source module lib/io.sh
> 
> EMULATOR_PID=0
> ```

I note this file is not renamed, any reason why it should not be?

vnvo2409:
> I note this file is not renamed, any reason why it should not be?

I forgot this function. The whole CI will collapse

vnvo2409:
> This is awesome news. I am a bit concerned about changes to the "public" interface. I am going to discuss with my colleagues, I suspect we will need to change some things in a backwards compatible way.

I think we could rename from `emulator` to `testbench` but I prefer the `emulator` 😕

coryan (commented):
To be clear, I like the `emulator` name better too.  Just that in some places we may need to support both the old name and the new name.

+kokoro:run 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-:-MMQg7eG10n3v4jkS6fL:b-twu78s)* status: all files reviewed, 4 unresolved discussions (waiting on @vnvo2409)

vnvo2409:
@coryan Is there any update about this PR ?

coryan (changes requested):
I think we need to preserve backwards compatibility where we can. See below for details.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-:-MMudkoX69JIijkkQucV:bw5kzjl)* status: all files reviewed, 9 unresolved discussions (waiting on @vnvo2409)

___
*[google/cloud/storage/client_options.h, line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMQZ327DQet4g7S3Huw:-MMu_9Vm0tbVsPDDWQ7-:bqtq9h7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dac65fe2c3c1466e2289121ae7289fdb87df002e/google/cloud/storage/client_options.h#L63)):*

I think this would make the change a breaking change, and I am not sure we need to do it.  I mean, the naming will be a bit awkward without this change, but OTOH, I am not sure we need to break any build scripts / tests that application developers may have created.

To be clear, I am not sure this is a blocker, but I am not sure it is not either. Let me chat with some folks here, including @devjgm .

</blockquote></details>

This is definitely a breaking change, I think the best way to proceed is to continue to support the old option, but change the documentation to recommend the new one.

These changes are Okay, but I will be asking for other changes elsewhere.
___
*[google/cloud/storage/client_options.cc, line 34 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMu_Xw-9P0_azvMLKSF:-MMu_Xw-9P0_azvMLKSG:bzaycpq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/157ce4bb7765836e2ab3312e396bcade5aee7135/google/cloud/storage/client_options.cc#L34)):*
> ```cpp
> 
> namespace internal {
> ```

I think we want a helper function like so:

```cc
absl::optional<std::string> GetEmulator() {
  auto emulator = GetEnv("CLOUD_STORAGE_EMULATOR_ENDPOINT");
  if (emulator) return emulator;
  return GetEnv("CLOUD_STORAGE_TESTBENCH_ENDPOINT");
}
```

this function would be used in this file like so:

```cc
std::string JsonEndpoint(ClientOptions const& options) {
  return GetEmulator().value_or(options.endpoint_) + "/storage/" + options.version();
}
```

This would preserve backwards compatibility and we would signal to users (in the documentation) that the new name is preferred.
___
*[google/cloud/storage/well_known_headers_test.cc, line 28 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMubDdPDkMNIBPNG4Ie:-MMubDdPDkMNIBPNG4If:b-rb2rqj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/157ce4bb7765836e2ab3312e396bcade5aee7135/google/cloud/storage/well_known_headers_test.cc#L28)):*
> ```cpp
> /// @test Verify that CustomHeader works as expected.
> TEST(WellKnownHeader, CustomHeader) {
>   CustomHeader header("x-goog-emulator-instructions", "do-stuff");
> ```

FYI, these changes are Okay, but the emulator needs to support the old values for backwards compatibility, more on that later.
___
*[google/cloud/storage/doc/storage-main.dox, line 61 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMQZtFB-s9VZTHPMYku:-MMubPze8GywCjwV8CV1:bgva4aj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dac65fe2c3c1466e2289121ae7289fdb87df002e/google/cloud/storage/doc/storage-main.dox#L61)):*

Putting a pin here to think about backwards compatibility.
</blockquote></details>

I think we should add a description for the old option, something like:

```
- `CLOUD_STORAGE_TESTBENCH_ENDPOINT=...` **DEPRECATED** use `CLOUD_STORAGE_EMULATOR_ENDPOINT` instead.
```
___
*[google/cloud/storage/emulator/emulator.py, line 680 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMQ_-PU9Nv8X1V29TSG:-MMuajPaB1n2lKf7ZEtY:b-rwspe7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dac65fe2c3c1466e2289121ae7289fdb87df002e/google/cloud/storage/emulator/emulator.py#L680)):*

This is also something that application developers may be using in their own testing.
</blockquote></details>

I think we need to preserve the backwards compatible behavior. please keep the old names for these attributes, e.g. `x_testbench_transfer_encoding`.
___
*[google/cloud/storage/emulator/grpc_server.py, line 139 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMubcPY7AsPImboeXyd:-MMubcPY7AsPImboeXye:b-2k7el8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/157ce4bb7765836e2ab3312e396bcade5aee7135/google/cloud/storage/emulator/grpc_server.py#L139)):*
> ```Python
>             request, bucket, context
>         )
>         upload.metadata.metadata["x_emulator_upload"] = "resumable"
> ```

Ditto, please preserve the returned name.  If we must, we can return both the new and old names.
___
*[google/cloud/storage/emulator/README.md, line 41 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMuboM28v-Lg1hPz46f:-MMuboM28v-Lg1hPz46g:bbu3c8m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/157ce4bb7765836e2ab3312e396bcade5aee7135/google/cloud/storage/emulator/README.md#L41)):*
> ```Markdown
> ## Force Failures
> 
> You can force the following failures by using the `x-goog-emulator-instructions` header.
> ```

Maybe add something like:

```
... using the `x-goog-emulator-instructions` header. The `x-goog-testbench-instructions` header is deprecated, but supported for backwards compatibility, please change your code to use `x-goog-emulator-instructions` instead.
```
___
*[google/cloud/storage/emulator/gcs/object.py, line 93 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMucAV9CxAD0gzHTOk1:-MMucAV9CxAD0gzHTOk2:bp3i3ga) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/157ce4bb7765836e2ab3312e396bcade5aee7135/google/cloud/storage/emulator/gcs/object.py#L93)):*
> ```Python
>     ):
>         if context is None:
>             instruction = request.headers.get("x-goog-emulator-instructions")
> ```

maybe:

```py
instruction = request.headers.get("x-goog-emulator-instructions") or request.headers.get("x-goog-testbench-instructions")
```

But if there is a more pythonic way to say "use `a` if present, `b` if not present" then I am all for it.
___
*[google/cloud/storage/emulator/gcs/object.py, line 175 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMucWte5KPSbB9eQCfB:-MMucWte5KPSbB9eQCfC:b-k9lkdr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/157ce4bb7765836e2ab3312e396bcade5aee7135/google/cloud/storage/emulator/gcs/object.py#L175)):*
> ```Python
>             "bucket": bucket.name,
>             "name": object_name,
>             "metadata": {"x_emulator_upload": "simple"},
> ```

As above, we need to preserve backwards compatibility for these.
___
*[google/cloud/storage/examples/storage_examples_common.cc, line 27 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMucfry4gTfnVL33QKF:-MMucfry4gTfnVL33QKG:b-ftkh9v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/157ce4bb7765836e2ab3312e396bcade5aee7135/google/cloud/storage/examples/storage_examples_common.cc#L27)):*
> ```cpp
> 
> bool UsingEmulator() {
>   return !google::cloud::internal::GetEnv("CLOUD_STORAGE_EMULATOR_ENDPOINT")
> ```

I think we should support the backwards compatibility option here.  Thoughts?
___
*[google/cloud/storage/internal/retry_object_read_source.cc, line 74 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMud3jo8WlBnrUOHelL:-MMud3jp3bwPijvsGY-D:bb2f5eu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/157ce4bb7765836e2ab3312e396bcade5aee7135/google/cloud/storage/internal/retry_object_read_source.cc#L74)):*
> ```cpp
>     return result;
>   }
>   bool has_emulator_instructions = false;
> ```

FYI, I thought about asking for backwards compatibility here, but it is not needed, this is our own code to support our own tests, we are fine.
___
*[google/cloud/storage/tests/object_basic_crud_integration_test.cc, line 155 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-MMudUK-14a0D1nh1RPc:-MMudUK-14a0D1nh1RPd:b3rw9a0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/157ce4bb7765836e2ab3312e396bcade5aee7135/google/cloud/storage/tests/object_basic_crud_integration_test.cc#L155)):*
> ```cpp
> StatusOr<Client> CreateNonDefaultClient() {
>   auto emulator =
>       google::cloud::internal::GetEnv("CLOUD_STORAGE_EMULATOR_ENDPOINT");
> ```

FYI, another case where backwards compatibility does not seem necessary.

vnvo2409:
Done,
I've already copied the emulator to a different location and run the test agaist master in order to verify that the emulator supports both `testbench` and `emulator` ( `grpc_integration_test` failed but it is acceptable, cause no one uses grpc right now )

coryan (commented):
+kokoro:run 

With this change a few `x_testbench_*` attributes in the object metadata are becoming `x_emulator_*`.  That is still a breaking change.  I would be Okay if we created both, with the old ones mark as deprecated (maybe to be removed in 90 days or something). Thoughts?

> Reviewed 12 of 12 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-:-MMvZEaBDcnpHWMhgtmz:b2lxrus)* status: :shipit: complete! all files reviewed, all discussions resolved

vnvo2409:
> I would be Okay if we created both, with the old ones mark as deprecated (maybe to be removed in 90 days or something). Thoughts?

On server side, we created only `x_emulator_*`. However, when the client requests the object metadata, `x_testbench_*` will be automatically added into the response ( you could take a look at `rest_metadata` ). I think we could keep supporting `x_testbench_*` even after 90 days. I think there should be no problems.

coryan:
> On server side, we created only `x_emulator_*`. However, when the client requests the object metadata, `x_testbench_*` will be automatically added into the response ( you could take a look at `rest_metadata` ).

I missed that, thanks!

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5518#-:-MMvf4KwD7mhsddq-55g:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5519, opened 2020-11-18
cleanup(spanner): plumb ConnectionOptions into admin connection classes

devbww:
Salvage part of the abandoned #5112 so that `DatabaseAdminConnection`
and `InstanceAdminConnection` both have a `BackgroundThreads` handle
from which they can obtain a `CompletionQueue` to execute the pieces
of long-running operations.

Part of #4038.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5519)

### DIFF
--- a/google/cloud/spanner/database_admin_connection.cc
+++ b/google/cloud/spanner/database_admin_connection.cc
@@ -22,6 +22,7 @@ namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
+
 namespace gcsa = ::google::spanner::admin::database::v1;
 
 using google::cloud::internal::Idempotency;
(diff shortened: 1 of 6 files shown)


google/cloud/spanner/database_admin_connection.cc
```diff
@@ -620,23 +624,25 @@ class DatabaseAdminConnectionImpl : public DatabaseAdminConnection {
   std::unique_ptr<RetryPolicy const> retry_policy_prototype_;
   std::unique_ptr<BackoffPolicy const> backoff_policy_prototype_;
   std::unique_ptr<PollingPolicy const> polling_policy_prototype_;
+  std::unique_ptr<BackgroundThreads> background_threads_;
```
coryan, line 627: Consider a comment here, objects of this class (most of the time) creates a pool of threads, and joins those threads in its destructor.  Those threads should not gain ownership of the object (via a `std::shared_ptr<>`) or deadlocks ensue during shutdown

google/cloud/spanner/instance_admin_connection.cc
```diff
@@ -298,7 +302,9 @@ class InstanceAdminConnectionImpl : public InstanceAdminConnection {
   std::unique_ptr<RetryPolicy const> retry_policy_prototype_;
   std::unique_ptr<BackoffPolicy const> backoff_policy_prototype_;
   std::unique_ptr<PollingPolicy const> polling_policy_prototype_;
+  std::unique_ptr<BackgroundThreads> background_threads_;
```
coryan, line 305: Ditto regarding thread pools and ownership cycles.

coryan (approved):
The comments I mention below are your call.  I am comfortable with the code as-is, but that is because I have made the mistakes I mentioned before, so I *think* I won't make them again (famous last words?)

google/cloud/spanner/database_admin_connection.cc
devbww: Done.

google/cloud/spanner/instance_admin_connection.cc
devbww: Done.

devbww (commented):
Thanks.

PR #5527, opened 2020-11-21
test(storage): support csek for the new emulator

vnvo2409:
This PR adds csek support for the new emulator.

I have some confusions. In the old testbench ( and also in the new emulator ), the base64 encoded encryption key sha256 is saved. However, on the [website](https://cloud.google.com/storage/docs/json_api/v1/objects) and [proto](https://github.com/googleapis/googleapis/blob/37ba54d7ed4da3052ceff96292631c3b6aae0e63/google/storage/v1/storage_resources.proto#L620-L621) definition file, it is written that `SHA256 hash value of the encryption key.` ( not base64 encoded ). Which one should be saved ?

Part of #4751 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5527)

### DIFF
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -416,7 +416,7 @@ def objects_compose(bucket_name, object_name):
             if source_object.get("objectPreconditions") is not None
             else None
         )
-        fake_request = utils.common.FakeRequest(args=dict())
+        fake_request = utils.common.FakeRequest(args=dict(), headers={})
         if generation is not None:
             fake_request.args["generation"] = generation
         if if_generation_match is not None:
@@ -443,6 +443,9 @@ def objects_copy(src_bucket_name, src_object_name, dst_bucket_name, dst_object_n
     dst_bucket = db.get_bucket_without_generation(dst_bucket_name, None).metadata
     src_object = db.get_object(
         flask.request, src_bucket_name, src_object_name, True, None
+    )
+    utils.csek.validation(
+        flask.request, src_object.metadata.customer_encryption.key_sha256, False, None
     )
     dst_metadata = resources_pb2.Object()
     dst_metadata.CopyFrom(src_object.metadata)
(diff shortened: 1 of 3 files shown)


coryan (approved):
> I have some confusions. In the old testbench ( and also in the new emulator ), the base64 encoded encryption key sha256 is saved.

Right.  BTW, I think we pass around base64 keys just because passing binary data through JSON is hard otherwise.

> However, on the website and proto definition file, it is written that SHA256 hash value of the encryption key. ( not base64 encoded ). Which one should be saved ?

The production system only saves the hash value.  In the emulator can do whatever is easier (I think), but seems like saving only the hash value would be less confusing / more in line with the expectations that customers have.  I believe the only reason we save the actual key is because that makes it easier to write the validation function.  In other words, I was lazy, we probably should have two validation functions, one that validates the key against the SHA and only that validates an incoming SHA against the saved SHA.

In any case, the code LGTM. I am happy to merge the code as-is, I can wait for you to fix the key saving in this PR, or we can just record that as a future improvement and move on. Your call, just tell me how to proceed.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5527#-:-MMkZTzPAJokdIGOik6c:b-7kalv4)* status: :shipit: complete! all files reviewed, all discussions resolved

vnvo2409:
> In any case, the code LGTM. I am happy to merge the code as-is, I can wait for you to fix the key saving in this PR, or we can just record that as a future improvement and move on. Your call, just tell me how to proceed.

I think it's better to fix the key saving in this PR.

vnvo2409:
> The production system only saves the hash value.

I misunderstood your explanation. I am asking whether the production system saves the hash value or the base64 encoded of the that hash value. But I look at the tests and it seems that we receive the base64 encoded version. Therefore, the only thing that needs to be changed is the documentation on the [website](https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations)
`customerEncryption.keySha256`
`SHA256 hash value of the encryption key.` to `SHA256 hash value of the encryption key.; encoded using base64`

coryan:
> > The production system only saves the hash value.
> 
> I misunderstood your explanation. I am asking whether the production system saves the hash value or the base64 encoded of the that hash value.

I would think that does not matter?  Both are just different representations of the same data, as long as one keeps straight which one is in use you should be fine.

> But I look at the tests and it seems that we receive the base64 encoded version. Therefore, the only thing that needs to be changed is the documentation on the [website](https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations)
> `customerEncryption.keySha256`
> `SHA256 hash value of the encryption key.` to `SHA256 hash value of the encryption key.; encoded using base64`

I just sent a CL to fix the documentation, we will see what they say, but thank you.

coryan:
I take it we can merge this PR then. If I am wrong we can always fix it with one more PR.

Issue #5521, opened 2020-11-19
Design a StreamRange<T> type


devjgm:
(This issue is just to track some thoughts and a vague need for some abstraction.)

For RPCs that one-way stream their responses, we need a C++ type to wrap that and provide an iterable range-like interface. An example RPC is [`BigQueryRead::ReadRows`](https://github.com/googleapis/googleapis/blob/7ea91db77f53f169a08726bd33efb73042c38284/google/cloud/bigquery/storage/v1/storage.proto#L83). 

An incomplete C++-ish pseudocode version of the class that we want might look like the following
```cc
template <typename T>
class StreamRange {
 public:
  class iterator {
   public:
    using value_type = T;
    using reference = value_type&;
    using pointer = value_type*;
    // ...
    reference operator*();
    pointer operator->();
  };
  iterator begin() const;
  iterator end() const;
};
```
Example usage
```cc
struct Foo { int a; };
StreamRange<Foo> r = ...;
for (auto& foo : r) {
  std::cout << foo.a;
}
```

## Requirements

* The `StreamRange<T>` would need to wrap the `grpc::ClientReader<T>` interface ([docs](https://grpc.io/docs/languages/cpp/basics/#streaming-rpcs)) so that it could keep calling `grpc::ClientReader<T>::Read()`.
* `StreamRange<T>` would need to handle resuming connections that failed partway through the stream.
  * would likely need to hold onto the _original_ "request" object
  * would likely need to hold onto the `Connection` class
  * would need to track progress on the stream so it knows where to resume. For example, if streaming rows back from a service, it may need to track the number of rows its already received so that it could resume at the next row
  * it would likely need some extension mechanism so that we could customize how the "resume" logic works

If the class needs another template parameter to encapsulate the "Resumer" that may be fine... or maybe a lambda? Or ???

## Prior art

There are some classes in our Spanner code that implement all of this logic, but it's currently not bundled in a reusable class. But here are some interesting places to seek inspiration.

* https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/spanner/internal/partial_result_set_reader.h
* https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/spanner/internal/partial_result_set_resume.h
* https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/internal/pagination_range.h
* https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/spanner/row.h#L244

[...]

PR #5532, opened 2020-11-23
feat: introduce a generic StreamRange<T>

devjgm:
Part of https://github.com/googleapis/google-cloud-cpp/issues/5521

This PR is part of the work needed for resumable streaming RPCs. This PR introduces a generic `StreamRange<T>`, which provides an input range of `T` and an iterator class for iterating it. This is a better, and reusable, version of the `RowStreamIterator` and `TupleStreamIterator` classes in spanner [link](https://github.com/googleapis/google-cloud-cpp/blob/43d6687ab1432ad56420c0292283cd3b615cf96f/google/cloud/spanner/row.h#L309). Indeed, those classes could easily be replaced by aliases to `StreamRange<T>`, except for small API breaks that would result, because those Spanner classes have weird constructors that this one doesn't have. We can discuss whether we want to make that change or not later.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532)

### DIFF
--- a/google/cloud/CMakeLists.txt
+++ b/google/cloud/CMakeLists.txt
@@ -373,6 +373,7 @@ if (GOOGLE_CLOUD_CPP_ENABLE_GRPC)
         internal/retry_loop.h
         internal/retry_loop_helpers.cc
         internal/retry_loop_helpers.h
+        internal/stream_range.h
         internal/time_utils.cc
         internal/time_utils.h)
     target_link_libraries(
@@ -447,6 +448,7 @@ if (GOOGLE_CLOUD_CPP_ENABLE_GRPC)
             internal/pagination_range_test.cc
             internal/polling_loop_test.cc
             internal/retry_loop_test.cc
+            internal/stream_range_test.cc
             internal/time_utils_test.cc)
 
         # Export the list of unit tests so the Bazel BUILD file can pick it up.
--- a/google/cloud/google_cloud_cpp_grpc_utils.bzl
+++ b/google/cloud/google_cloud_cpp_grpc_utils.bzl
@@ -39,6 +39,7 @@ google_cloud_cpp_grpc_utils_hdrs = [
     "internal/polling_loop.h",
     "internal/retry_loop.h",
     "internal/retry_loop_helpers.h",
+    "internal/stream_range.h",
     "internal/time_utils.h",
 ]
 
--- a/google/cloud/google_cloud_cpp_grpc_utils_unit_tests.bzl
+++ b/google/cloud/google_cloud_cpp_grpc_utils_unit_tests.bzl
@@ -28,5 +28,6 @@ google_cloud_cpp_grpc_utils_unit_tests = [
     "internal/pagination_range_test.cc",
     "internal/polling_loop_test.cc",
     "internal/retry_loop_test.cc",
+    "internal/stream_range_test.cc",
     "internal/time_utils_test.cc",
 ]
--- a/google/cloud/internal/stream_range.h
+++ b/google/cloud/internal/stream_range.h
@@ -29,7 +29,7 @@ inline namespace GOOGLE_CLOUD_CPP_NS {
 namespace internal {
 
 /// A sentinel type to indicate the successful end of stream.
-using StreamEnd = absl::monostate;
+struct StreamEnd {};
 
 /**
  * The type returned by `StreamReader<T>`.
--- a/google/cloud/internal/stream_range.h
+++ b/google/cloud/internal/stream_range.h
@@ -1,0 +1,204 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_STREAM_RANGE_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_STREAM_RANGE_H
+
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include "absl/types/optional.h"
+#include "absl/types/variant.h"
+#include <functional>
+#include <iterator>
+#include <memory>
+
+namespace google {
+namespace cloud {
+inline namespace GOOGLE_CLOUD_CPP_NS {
+namespace internal {
+
+/// A sentinel type to indicate the successful end of stream.
+using StreamEnd = absl::monostate;
+
+/**
+ * The type returned by `ReaderFunction<T>`.
+ *
+ * This variant may contain one of the following types:
+ *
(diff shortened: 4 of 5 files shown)


coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532)* status: all files reviewed, 4 unresolved discussions (waiting on @devjgm)

___
*[google/cloud/internal/stream_range.h, line 32 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532#-MMn4SKMEN6RIFb-Q7Iw:-MMn4SKMEN6RIFb-Q7Ix:bca6y0w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ad61c400da348798c71a23f5e3b1d10143fa5398/google/cloud/internal/stream_range.h#L32)):*
> ```objc
> 
> /// A sentinel type to indicate the successful end of stream.
> using StreamEnd = absl::monostate;
> ```

Why not just `struct StreamEnd {};` ?
___
*[google/cloud/internal/stream_range.h, line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532#-MMp9QNp-KcSIUac0IM3:-MMp9QNq9-k7R8-dxdXo:bai9hyf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ad61c400da348798c71a23f5e3b1d10143fa5398/google/cloud/internal/stream_range.h#L44)):*
> ```objc
>  */
> template <typename T>
> using StreamReaderResult = absl::variant<StreamEnd, Status, T>;
> ```

Do we need `StreamEnd` here?  Why is `Status == Okay` not a good indication of "end of stream"?
___
*[google/cloud/internal/stream_range.h, line 64 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532#-MMp9dpk6lo9cx0h-RzE:-MMp9dpk6lo9cx0h-RzF:b93eey3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ad61c400da348798c71a23f5e3b1d10143fa5398/google/cloud/internal/stream_range.h#L64)):*
> ```objc
>  */
> template <typename T>
> using StreamReader = StreamReaderResult<T>();
> ```

Shouldn't this be `std::function<StreamReaderResult<T>()>` ? Eventually you will want to associate state, and use something other than plain functions.


Oh, I see, you use this as the template argument for `std::function<>`.  The description should be something like "The signature of a function".  In any case, using `std::function<>` seems cleaner to me.
___
*[google/cloud/internal/stream_range.h, line 124 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532#-MMpANOuEvCccqK-mRKa:-MMpANOuEvCccqK-mRKb:b5hhnyr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ad61c400da348798c71a23f5e3b1d10143fa5398/google/cloud/internal/stream_range.h#L124)):*
> ```objc
>     }
> 
>     Iterator operator++(int) {
> ```

FYI, there are no tests for this version of `operator++()`, consider a test that iterates over the whole range using postfix iteration.
___
*[google/cloud/internal/stream_range.h, line 169 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532#-MMpBCXlEiN7-XdNlSF6:-MMpBCXlEiN7-XdNlSF7:b48ylbe) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ad61c400da348798c71a23f5e3b1d10143fa5398/google/cloud/internal/stream_range.h#L169)):*
> ```objc
>   StreamRange& operator=(StreamRange const&) = delete;
>   // NOLINTNEXTLINE(performance-noexcept-move-constructor)
>   StreamRange(StreamRange&& sr) = default;
> ```

Consider: `noexcept(noexcept(StatusOr<T>(sr.current_)) && noexcept(std::function<StreamReader<T>>(sr.reader_)))` a bit ugly, but more accurate and improves things in the some cases (starting with C++20 I would say the majority of cases).

It might be possible to hide that ugliness is some `constexpr` function.

devjgm (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532)* status: 3 of 5 files reviewed, 4 unresolved discussions (waiting on @coryan)

___
*[google/cloud/internal/stream_range.h, line 32 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532#-MMn4SKMEN6RIFb-Q7Iw:-MMpUUvPAyuOoYO7dnjX:bfmqlwn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ad61c400da348798c71a23f5e3b1d10143fa5398/google/cloud/internal/stream_range.h#L32)):*

Why not just `struct StreamEnd {};` ?
</blockquote></details>

I used the alias to monostate because I _thought_ I would need the equality operators that it provides. However, I didn't need those and an empty struct like you suggest would work fine.... though I've completely removed this now based on other suggestions of yours.
___
*[google/cloud/internal/stream_range.h, line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532#-MMp9QNp-KcSIUac0IM3:-MMpUhAy6XlgtHwyfvrO:b-fca6df) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ad61c400da348798c71a23f5e3b1d10143fa5398/google/cloud/internal/stream_range.h#L44)):*

Do we need `StreamEnd` here?  Why is `Status == Okay` not a good indication of "end of stream"?
</blockquote></details>

Good point. Status==ok would work totally fine. I originally was using `StatusOr<T>`, which cannot hold an OK status without also holding a default-constructed `T`, which I didn't want to use. But since I switched to variant, this will work great, and it's also simpler. Thanks for the suggestion!
___
*[google/cloud/internal/stream_range.h, line 64 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532#-MMp9dpk6lo9cx0h-RzE:-MMpUzbKBuNRAS1oHSuL:bexnttf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ad61c400da348798c71a23f5e3b1d10143fa5398/google/cloud/internal/stream_range.h#L64)):*

Shouldn't this be `std::function<StreamReaderResult<T>()>` ? Eventually you will want to associate state, and use something other than plain functions.


Oh, I see, you use this as the template argument for `std::function<>`.  The description should be something like "The signature of a function".  In any case, using `std::function<>` seems cleaner to me.
</blockquote></details>

Changed to `std::function`.  This also lets me get rid of the `StreamReaderResult` alias, because I can instead use `std::function<T>::result_type`. So, win!
___
*[google/cloud/internal/stream_range.h, line 169 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532#-MMpBCXlEiN7-XdNlSF6:-MMpVDY0BvrCdZRHrC07:b-wzkyjs) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ad61c400da348798c71a23f5e3b1d10143fa5398/google/cloud/internal/stream_range.h#L169)):*

Consider: `noexcept(noexcept(StatusOr<T>(sr.current_)) && noexcept(std::function<StreamReader<T>>(sr.reader_)))` a bit ugly, but more accurate and improves things in the some cases (starting with C++20 I would say the majority of cases).

It might be possible to hide that ugliness is some `constexpr` function.

</blockquote></details>

Done. I still need the `NOLINTNEXTLINE` comments otherwise clang-tidy will still complain when this evaluates to false. ANd since the lines wrap, I'll need additional `NOLINTNEXTLINE` comments. Sight. A little messy, but it seems to work.

coryan (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5532)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5539, opened 2020-11-24
refactor: StreamRange<T> c'tor is now private

devjgm:
Constructors are part of a class's public API, yet in many cases,
they're also very close to an implementation detail. Even for non-public
classes in an `internal` namespace, if/when they are exposed to users
via an alias in a public namespace, those constructors become part of
that user-facing API.

Unless the constructor is necessarily part of the concept or abstraction
represented by the class, it's probably safest to make
implementation-detail constructors private and expose them only via
factory functions. This way, we can alias a private class into a
public namespace and not expose the constructors. And in order to call
the factory function, users would need to spell our `internal`
namespace, which is off limits.

So I think this is a better way to make sure we only expose the
fundamental parts of an API to users. This is also a conservative bet
because we can always expose more later without breaking users, but we
cannot remove things.

Note: This PR also adds a default constructor to `StreamRange<T>` that
constructs an empty range.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5539)

### DIFF
--- a/google/cloud/internal/stream_range.h
+++ b/google/cloud/internal/stream_range.h
@@ -146,13 +146,9 @@ class StreamRange {
   using const_pointer = typename iterator::const_pointer;
 
   /**
-   * Constructs a `StreamRange<T>` that will use the given @p reader.
-   *
-   * @param reader must not be nullptr.
+   * Default-constructs an empty range.
    */
-  explicit StreamRange(StreamReader<T> reader) : reader_(std::move(reader)) {
-    Next();
-  }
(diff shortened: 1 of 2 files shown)


coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5539#-:-MMuZrkp12HEi3fZhiw3:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5540, opened 2020-11-24
refactor: remove PaginationRange's dep on gRPC

devjgm:
`PaginationRange<T>` no longer has a dep on gRPC and it can also work with `Request` and `Response` objects that are either protobufs or plain old structs. This PR should allow GCS to use this `PaginationRange<T>` without taking a dep on GRPC or proto.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5540)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-bazel.sh
+++ b/ci/kokoro/docker/build-in-docker-bazel.sh
@@ -94,11 +94,18 @@ io::log "Fetching dependencies"
   @go_sdk//... \
   @remotejdk11_linux//:jdk
 
+excluded_targets=()
+if [[ "${BUILD_NAME}" == "msan" ]]; then
+  # GoogleTest's `TYPED_TEST` causes msan to complain. Not sure why.
+  excluded_targets+=("-//google/cloud:internal_pagination_range_test")
+fi
+
 echo "================================================================"
 io::log "Compiling and running unit tests"
 echo "bazel ${BAZEL_VERB}" "${bazel_args[@]}"
 "${BAZEL_BIN}" ${BAZEL_VERB} \
-  "${bazel_args[@]}" "--test_tag_filters=-integration-test" ...
+  "${bazel_args[@]}" "--test_tag_filters=-integration-test" \
+  -- ... "${excluded_targets[@]}"
 
 should_run_integration_tests() {
   if [[ "${SOURCE_DIR:-}" == "super" ]]; then
--- a/google/cloud/CMakeLists.txt
+++ b/google/cloud/CMakeLists.txt
@@ -141,6 +141,7 @@ add_library(
     internal/getenv.h
     internal/invoke_result.h
     internal/ios_flags_saver.h
+    internal/pagination_range.h
     internal/parse_rfc3339.cc
     internal/parse_rfc3339.h
     internal/port_platform.h
@@ -245,6 +246,7 @@ if (BUILD_TESTING)
         internal/format_time_point_test.cc
         internal/future_impl_test.cc
         internal/invoke_result_test.cc
+        internal/pagination_range_test.cc
         internal/parse_rfc3339_test.cc
         internal/random_test.cc
         internal/retry_policy_test.cc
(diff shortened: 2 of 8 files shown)


coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5540#-:-MMve_BN3MPqnjr83QAb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

devjgm:
GoogleTest's `TYPED_TEST` causes msan to complain. It looks like an msan bug to me, which is not surprising since msan is a giant PITA to configure, setup, and use (unlike asan, etc). To avoid this I'm just excluding this one test from our `msan` BUILD. If anyone has a better suggestion, I'm open to it.

Error log: https://source.cloud.google.com/results/invocations/a97152f8-9238-4f06-b887-2722a3cf7bf3/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fdocker%2Fmsan-presubmit/log

coryan:
> GoogleTest's `TYPED_TEST` causes msan to complain.

Ugh.

> It looks like an msan bug to me, which is not surprising since msan is a giant

I do not think it is a msan bug.  The issue seems to be that we (or googletest) are using a version of `__cxa_demangle` that was not compiled with msan.  The log reads:

```
    #1 0x7fb6e6829316 in __cxa_demangle (/lib64/libc++abi.so.1+0x13316)
```

That call comes from here:

https://github.com/google/googletest/blob/v1.10.x/googletest/include/gtest/internal/gtest-type-util.h#L91

Because that is not compiled with msan, msan does not know that the data was initialized.

> PITA to configure, setup, and use (unlike asan, etc).

:heavy_plus_sign:  :100: 

> To avoid this I'm just excluding this one test from our `msan` BUILD. If anyone has a better suggestion, I'm open to it.

Do you mind filing a bug to fix our `msan` build to link against the right `libc++abi.so`?

I was going to suggest avoiding `TYPED_TEST` but that also seems like a PITA. I am out of sensible ideas (I mean, "use valgrind" is an alternative, but not very sensible).

Oh... what about a MSAN suppression?  Let me check on that.

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5540#-:-MMx5Gs84z2v4Qzrnnfo:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devjgm)

devjgm:
> I do not think it is a msan bug. The issue seems to be that we (or googletest) are using a version of `__cxa_demangle` that was not compiled with msan. The log reads:
> 
> ```
>     #1 0x7fb6e6829316 in __cxa_demangle (/lib64/libc++abi.so.1+0x13316)
> ```
> 
> That call comes from here:
> 
> https://github.com/google/googletest/blob/v1.10.x/googletest/include/gtest/internal/gtest-type-util.h#L91
> 
> Because that is not compiled with msan, msan does not know that the data was initialized.

We do compile cxxabi with msan ([Dockerfile](https://github.com/googleapis/google-cloud-cpp/blob/d130e77526fea7109b4dfec47e69bf846b4ec0b7/ci/kokoro/docker/Dockerfile.fedora-libcxx-msan#L57)), and we compile googletest with the msan flags in the [.bazelrc](https://github.com/googleapis/google-cloud-cpp/blob/d130e77526fea7109b4dfec47e69bf846b4ec0b7/.bazelrc#L62). So I wasn't sure where the issue was.
> 
> > PITA to configure, setup, and use (unlike asan, etc).
> 
>  
> 
> > To avoid this I'm just excluding this one test from our `msan` BUILD. If anyone has a better suggestion, I'm open to it.
> 
> Do you mind filing a bug to fix our `msan` build to link against the right `libc++abi.so`?

Sure thing. https://github.com/googleapis/google-cloud-cpp/issues/5544

> 
> I was going to suggest avoiding `TYPED_TEST` but that also seems like a PITA. I am out of sensible ideas (I mean, "use valgrind" is an alternative, but not very sensible).

I looked into avoiding `TYPED_TEST` too. It's really just The Right solution for running a bunch of tests on several types. I could factor each `TEST()` body into its own function template. 

> 
> Oh... what about a MSAN suppression? Let me check on that.

I tried the `-fsanitize-blacklist=.msan-blacklist.txt`. That didn't work (I think) because bazel runs the build in a sandbox and so the sandboxed compiler cannot access the `.msan-blacklist.txt` file.

I'm out of (good) ideas other than skipping this test in our msan build.

PR #5541, opened 2020-11-24
fix(pubsub): deadlock during fire & forget shutdown

coryan:
In very rare circumstances (maybe as evidenced by the test) the library
would deadlock trying to shutdown a subscription. I noticed this in less
than 1 run over 10,000, and only when compiled with optimizations (TSAN
and ASAN masked the problem).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5541)

### DIFF
--- a/google/cloud/pubsub/internal/subscription_session_test.cc
+++ b/google/cloud/pubsub/internal/subscription_session_test.cc
@@ -467,7 +467,6 @@ TEST(SubscriptionSessionTest, FireAndForgetShutdown) {
 
   auto mock = std::make_shared<pubsub_testing::MockSubscriberStub>();
   AsyncSequencer<bool> on_read;
-  AsyncSequencer<void> on_cancel;
   AsyncSequencer<Status> on_finish;
 
   auto async_pull_mock = [&](google::cloud::CompletionQueue& cq,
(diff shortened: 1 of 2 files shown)


google/cloud/pubsub/internal/subscription_session_test.cc
```diff
@@ -459,6 +461,78 @@ TEST(SubscriptionSessionTest, FireAndForget) {
...
+  auto finish = on_finish.PopFront();
+  // Shutdown the completion queue, this will disable the timers for the second
+  // async_pull
+  background->cq().Shutdown();
+  finish.set_value(Status{});
+
+  // At this point the streaming pull cannot restart, so there are no pending
+  // operations.  Eventually the session will be finished.
+  shutdown_completed.get_future().get();
+  background.reset(nullptr);
```
devbww, line 533: Is there a reason we need to do this explicitly?

If so, I believe the `reset()` form is probably preferable.

devbww (approved):

google/cloud/pubsub/internal/subscription_session_test.cc
coryan: Good catch. No, we do not need an explicit reset, in fact, we do not need the `unique_ptr<>`, that was a remnant when I wanted to debug this.

devbww (approved):

PR #5538, opened 2020-11-24
refactor!: PaginationRange<T> is now an alias to StreamRange<T>

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/5537

**BREAKING CHANGE**

This change removes a constructor that not intended to be part of the public API from several "Range" types in the Pub/Sub and Spanner APIs. The affected types are:

  * `google/cloud/pubsub/subscription_admin_connection.h`
    * `using ListSubscriptionsRange = google::cloud::internal::PaginationRange`
    * `using ListSnapshotsRange = google::cloud::internal::PaginationRange`
  * `google/cloud/pubsub/topic_admin_connection.h`
    * `using ListTopicsRange = google::cloud::internal::PaginationRange`
    * `using ListTopicSubscriptionsRange = google::cloud::internal::PaginationRange`
    * `using ListTopicSnapshotsRange = google::cloud::internal::PaginationRange`
  * `google/cloud/spanner/database_admin_connection.h`
    * `using ListDatabaseRange = google::cloud::internal::PaginationRange`
    * `using ListBackupOperationsRange = google::cloud::internal::PaginationRange`
    * `using ListDatabaseOperationsRange = google::cloud::internal::PaginationRange`
    * `using ListBackupsRange = google::cloud::internal::PaginationRange`
  * `google/cloud/spanner/instance_admin_connection.h`
    * `using ListInstancesRange = google::cloud::internal::PaginationRange`
    * `using ListInstanceConfigsRange = google::cloud::internal::PaginationRange`

All normal usage of these types will continue to work, i.e., they will all continue to work as ranges. However, the following constructor was removed:

https://github.com/googleapis/google-cloud-cpp/blob/d7d3576c6533dc15bd656b74bed8fc73e7400980/google/cloud/internal/pagination_range.h#L142-L144

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5538)

PR #5545, opened 2020-11-25
refactor!: change gcs to use common PaginationRange<T>

devjgm:
**BREAKING CHANGE**

The common `PaginationRange<T>` class lacks a constructor that existed in the custom range that's being replaced. Specifically, these new ranges will not be constructible with a constructor like:

```cc
  explicit PaginationRange(	
      Request request,	
      std::function<StatusOr<Response>(Request const& r)> loader,	
      std::function<std::vector<T>(Response r)> get_items =	
          [](Response r) { ... })
```

But this constructor was never intended to be part of the public API of these ranges. And in order for a caller to use it correctly they would likely need to reference Request and Response types that are in storage's `internal` namespace. But nonetheless, this appears to be a potentially breaking change.

I also removed `IteratorCompare` tests that now fail in cases where they were asserting behavior that isn't required of input iterators. In particular when an input iterator is incremented, all previous copies of the iterator may be (and often are) invalidated. Also, it's UB to compare iterators from different ranges. With these assertions removed, the whole `IteratorCompare` test was not doing anything unique, so I removed them.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5545)


coryan (commented):
> also removed IteratorCompare tests that now fail in cases where they were asserting behavior that isn't required of input iterators.

Okay... but this highlights an additional breaking change right?

To be clear, I like the new behavior for `operator==()` better, I wish I had implemented that to begin with.  But I didn't, so now we:

* **MUST** document this as a breaking change if we are taking it.
* *SHOULD* consider if the breaking change is worth it:  is the pain for our users to change their code enough of a benefit for them otherwise?  I think the new iterator is probably faster, and less surprising/buggy when there are duplicate elements in the stream.

> In particular when an input iterator is incremented, all previous copies of the iterator may be (and often are) invalidated.

Nevertheless, our previous iterators did not get invalidated, and AFAICT neither are the new iterators.  Not sure what this means, maybe that we should explicitly say that we do not guarantee that old iterators remain valid.

Thoughts?

> Reviewed 15 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5545#-:-MMxJMNd8kO3oLsd2o25:b-aq5naw)* status: :shipit: complete! all files reviewed, all discussions resolved

devjgm:
> > also removed IteratorCompare tests that now fail in cases where they were asserting behavior that isn't required of input iterators.
> 
> Okay... but this highlights an additional breaking change right?
> 
> To be clear, I like the new behavior for `operator==()` better, I wish I had implemented that to begin with. But I didn't, so now we:
> 
> * **MUST** document this as a breaking change if we are taking it.
> * _SHOULD_ consider if the breaking change is worth it:  is the pain for our users to change their code enough of a benefit for them otherwise?  I think the new iterator is probably faster, and less surprising/buggy when there are duplicate elements in the stream.
> 
> > In particular when an input iterator is incremented, all previous copies of the iterator may be (and often are) invalidated.
> 
> Nevertheless, our previous iterators did not get invalidated, and AFAICT neither are the new iterators. Not sure what this means, maybe that we should explicitly say that we do not guarantee that old iterators remain valid.
> 
> Thoughts?

Good questions and suggestions. I did think about these iterator semantic changes and concluded that they were not _API_ breaking changes, though they are potentially _implementation_ breaking changes.

The old `PaginationRange` that's being replaced in this PR includes the following user-facing documentation:

>    * Creating, and particularly incrementing, multiple iterators on the same	
>    * ListHmacKeysReader is **unsupported** and can produce incorrect results.

That's good and it it's inline with the requirements of any [input iterator](https://en.cppreference.com/w/cpp/named_req/InputIterator), which says about incrementing an input iterator

> Any copies of the previous value of i are no longer required to be either dereferenceable or to be **in the domain of ==**
> ...
> "in the domain of ==" means equality comparison is defined between the two iterator values.

However, in the `IteratorCompare` tests, the code tested the behavior that results when incrementing multiple iterators on the same reader. It's fine that it did and even that it had a stronger implementation than input iterator requires. However, AFAICT, that test was simply checking some "unspecified" behavior about the *implementation*.

The other unspecified behavior asserted in `IteratorCompare` was that iterators from *different readers* can be compared. Again, this is something that the "input iterator" concept does not guarantee and any user should NOT rely on this behavior unless the concrete type explicitly adds this stronger guarantee. So in this case also it looked like the test was asserting unspecified behavior about the *implementation*.

I could just remove those two checks from the `IteratorCompare` tests. The rest of the `IteratorCompare` assertions pass. However, it seems that those two assertions are basically the whole point of that test, which is why I removed it.

So my conclusion was that this behavior change was not changing anything about the API that users could/should reasonably depend on. But maybe our "BREAKING CHANGE" comments should refer to potential breakages even from unspecified implementation changes, so I'll document this change in `CHANGELOG.md`.

As to if this change is worth it to users? I say yes, but I wouldn't be sad to be convinced otherwise. If we decide to drop this PR, we probably want to also revert https://github.com/googleapis/google-cloud-cpp/pull/5538, which would have all the same concerns.

We do need `StreamReader<T>` and *some* common `PaginationRange<T>` for all the *new* libraries that we write and generate. It would be nice if we used the same ones in all of our libraries, but I agree that *users* probably wouldn't care about that implementation detail. I could keep the old `PaginationRange`s as they were, perhaps rename them to indicate their "do-not-use-in-new-code-ness", and then add a new `PaginationRange` in common for BigQuery and the generator to use. That would be my second preference, but it would be fine with me.

If none of this is convincing, I can drop this and go the other direction.

coryan:
> > Any copies of the previous value of i are no longer required to be either dereferenceable or to be **in the domain of ==**

Ah, that is an interesting bit that I had not noticed before.

> So my conclusion was that this behavior change was not changing anything about the API that users could/should reasonably depend on.

Thanks, I am convinced now.

> But maybe our "BREAKING CHANGE" comments should refer to potential breakages even from unspecified implementation changes, so I'll document this change in `CHANGELOG.md`.

SGTM.

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5545#-:-MMzcN_d5KnVTmIdqKf8:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5550, opened 2020-11-25
feat: implement Bigtable connection refresh

### LINKED ISSUE #4996: Bigtable latency spikes every hour due to grpc reconnects
Is it possible to implement something similar to the ChannelPrimer in the Bigtable Java client to mitigate hourly spikes in latency due to underlying grpc connection shutdowns & reconnects 

Java client reference: https://github.com/googleapis/java-bigtable/pull/115
Related Github issue for Java client: https://github.com/googleapis/java-bigtable-hbase/issues/1463

dopiera:
This fixes #4996.

This works, however it is not in its final form. Currently, background
threads running `CompletionQueue` are created underhandedly, without
users' control in every Bigtable `{Admin,InstanceAdmin,Data}Client`.

The choice of the `{Admin,InstanceAdmin,Data}Client` is consistent with
Spanner and PubSub, where counterparts of those clients are called
Connections.

The next steps will contain removing the `CompletionQueue` from both the
underlying clients and the user-facing API and instead providing
`BackgroundThreads` to the ctors. I didn't decide to do this in chunks
because otherwise it would either be confusing, which CompletionQueues
are used or we'd have to duplicate the whole API to use the
CompletionQueue provided in the ctors.

More tests will be possible when `BackgroundThreadsFactory` will be
passed in the `ClientOptions`

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550)

### DIFF
--- a/google/cloud/bigtable/client_options.cc
+++ b/google/cloud/bigtable/client_options.cc
@@ -27,6 +27,9 @@ std::shared_ptr<grpc::ChannelCredentials> BigtableDefaultCredentials() {
   }
   return grpc::GoogleDefaultCredentials();
 }
+// As learned from experiments, idle gRPC connections enter IDLE state after 4m.
+std::chrono::milliseconds constexpr kDefaultRefreshPeriod =
+    std::chrono::minutes(3);
 }  // anonymous namespace
 
 namespace google {
@@ -96,7 +99,10 @@ ClientOptions::ClientOptions(std::shared_ptr<grpc::ChannelCredentials> creds)
       admin_endpoint_("bigtableadmin.googleapis.com"),
       instance_admin_endpoint_("bigtableadmin.googleapis.com"),
       tracing_components_(internal::DefaultTracingComponents()),
-      tracing_options_(internal::DefaultTracingOptions()) {
+      tracing_options_(internal::DefaultTracingOptions()),
+      // Refresh connections before the server has a chance to disconnect them
+      // due to being idle.
+      max_conn_refresh_period_(kDefaultRefreshPeriod) {
   static std::string const kUserAgentPrefix = UserAgentPrefix();
   channel_arguments_.SetUserAgentPrefix(kUserAgentPrefix);
   channel_arguments_.SetMaxSendMessageSize(
--- a/google/cloud/bigtable/client_options.h
+++ b/google/cloud/bigtable/client_options.h
@@ -326,6 +326,25 @@ class ClientOptions {
   /// Return the options for use when tracing RPCs.
   TracingOptions const& tracing_options() const { return tracing_options_; }
 
+  /**
+   * Maximum connection refresh period, as set via `set_max_conn_refresh_period`
+   */
+  std::chrono::milliseconds max_conn_refresh_period() {
+    return max_conn_refresh_period_;
+  }
+
+  /**
+   * If set to a positive number, the client will refresh connections at random
+   * moments not more apart from each other than this duration. This is
+   * necessary to avoid all connections simultaneously expiring and causing
+   * latency spikes.
+   */
+  ClientOptions& set_max_conn_refresh_period(
+      std::chrono::milliseconds max_conn_refresh_period) {
+    max_conn_refresh_period_ = max_conn_refresh_period;
+    return *this;
+  }
+
  private:
   friend struct internal::InstanceAdminTraits;
   friend struct ClientOptionsTestTraits;
@@ -348,6 +367,7 @@ class ClientOptions {
   std::string instance_admin_endpoint_;
   std::set<std::string> tracing_components_;
   TracingOptions tracing_options_;
+  std::chrono::milliseconds max_conn_refresh_period_;
 };
 }  // namespace BIGTABLE_CLIENT_NS
 }  // namespace bigtable
--- a/google/cloud/bigtable/internal/common_client.cc
+++ b/google/cloud/bigtable/internal/common_client.cc
@@ -19,20 +19,56 @@ namespace bigtable {
 inline namespace BIGTABLE_CLIENT_NS {
 namespace internal {
 
-std::vector<std::shared_ptr<grpc::Channel>> CreateChannelPool(
-    std::string const& endpoint, bigtable::ClientOptions const& options) {
-  std::vector<std::shared_ptr<grpc::Channel>> result;
-  for (std::size_t i = 0; i != options.connection_pool_size(); ++i) {
-    auto args = options.channel_arguments();
-    if (!options.connection_pool_name().empty()) {
-      args.SetString("cbt-c++/connection-pool-name",
-                     options.connection_pool_name());
-    }
-    args.SetInt("cbt-c++/connection-pool-id", static_cast<int>(i));
-    result.push_back(
-        grpc::CreateCustomChannel(endpoint, options.credentials(), args));
-  }
-  return result;
+ConnectionRefreshState::ConnectionRefreshState(
+    std::chrono::milliseconds max_conn_refresh_period)
+    : max_conn_refresh_period_(max_conn_refresh_period),
+      rng_(std::random_device{}()) {}
+
+std::chrono::milliseconds ConnectionRefreshState::RandomizedRefreshDelay() {
+  std::lock_guard<std::mutex> lk(mu_);
+  return std::chrono::milliseconds(
+      std::uniform_int_distribution<decltype(max_conn_refresh_period_)::rep>(
+          1, max_conn_refresh_period_.count())(rng_));
+}
+
+void ScheduleChannelRefresh(
+    std::shared_ptr<CompletionQueue> const& cq,
+    std::shared_ptr<ConnectionRefreshState> const& state,
+    std::shared_ptr<grpc::Channel> const& channel) {
+  // The timers will only hold weak pointers to the channel or to the
+  // completion queue, so if either of them are destroyed, the timer chain
+  // will simply not continue. Unfortunately, that means that some stray
+  // timers may remain in the `CompletionQueue` for a while, but this is
+  // generally unavoidable because there is no way to cancel individual
+  // timers.
+  std::weak_ptr<grpc::Channel> weak_channel(channel);
+  std::weak_ptr<CompletionQueue> weak_cq(cq);
+  using TimerFuture = future<StatusOr<std::chrono::system_clock::time_point>>;
+  cq->MakeRelativeTimer(state->RandomizedRefreshDelay())
+      .then([weak_channel, weak_cq, state](TimerFuture fut) {
+        if (!fut.get()) {
+          // Timer cancelled.
+          return;
+        }
+        auto channel = weak_channel.lock();
+        if (!channel) return;
+        auto cq = weak_cq.lock();
+        if (!cq) return;
+        cq->AsyncWaitConnectionReady(channel, std::chrono::system_clock::now() +
+                                                  kConnectionReadyTimeout)
+            .then([weak_channel, weak_cq, state](future<Status> fut) {
+              auto conn_status = fut.get();
+              if (!conn_status.ok()) {
+                GCP_LOG(WARNING)
+                    << "Failed to refresh connection. Error: " << conn_status;
+              }
+              auto channel = weak_channel.lock();
+              if (!channel) return;
+              auto cq = weak_cq.lock();
+              if (!cq) return;
+              ScheduleChannelRefresh(cq, state, channel);
+            });
+      });
 }
 
 }  // namespace internal
--- a/google/cloud/bigtable/internal/common_client.h
+++ b/google/cloud/bigtable/internal/common_client.h
@@ -17,6 +17,10 @@
 
 #include "google/cloud/bigtable/client_options.h"
 #include "google/cloud/bigtable/version.h"
+#include "google/cloud/connection_options.h"
+#include "google/cloud/internal/random.h"
+#include "google/cloud/log.h"
+#include "google/cloud/status_or.h"
 #include <grpcpp/grpcpp.h>
 
 namespace google {
@@ -25,9 +29,40 @@ namespace bigtable {
 inline namespace BIGTABLE_CLIENT_NS {
 namespace internal {
 
-/// Create a pool of `grpc::Channel` objects based on the client options.
-std::vector<std::shared_ptr<grpc::Channel>> CreateChannelPool(
-    std::string const& endpoint, bigtable::ClientOptions const& options);
+/**
+ * Time after which we bail out waiting for a connection to become ready.
+ *
+ * This number was copied from the Java client and there doesn't seem to be a
+ * well-founded reason for it to be exactly this. It should not bee too large
+ * since waiting for a connection to become ready is not cancellable.
+ */
+std::chrono::seconds constexpr kConnectionReadyTimeout(10);
+
+/**
+ * State required by timers scheduled by `CommonClient`.
+ *
+ * The scheduled timers might outlive `CommonClient`. They need some shared,
+ * persistent state. Objects of this class implement it.
+ */
+class ConnectionRefreshState {
+ public:
+  explicit ConnectionRefreshState(
+      std::chrono::milliseconds max_conn_refresh_period);
+  std::chrono::milliseconds RandomizedRefreshDelay();
+
+ private:
+  std::mutex mu_;
+  std::chrono::milliseconds max_conn_refresh_period_;
+  google::cloud::internal::DefaultPRNG rng_;
+};
+
+/**
+ * Schedule a chain of timers to refresh the connection.
+ */
+void ScheduleChannelRefresh(
+    std::shared_ptr<CompletionQueue> const& cq,
+    std::shared_ptr<ConnectionRefreshState> const& state,
+    std::shared_ptr<grpc::Channel> const& channel);
 
 /**
  * Refactor implementation of `bigtable::{Data,Admin,InstanceAdmin}Client`.
@@ -54,7 +89,21 @@ class CommonClient {
   //@}
 
   explicit CommonClient(bigtable::ClientOptions options)
-      : options_(std::move(options)), current_index_(0) {}
+      : options_(std::move(options)),
+        current_index_(0),
+        background_threads_(
+            google::cloud::internal::DefaultBackgroundThreads(1)),
+        cq_(std::make_shared<CompletionQueue>(background_threads_->cq())),
+        refresh_state_(std::make_shared<ConnectionRefreshState>(
+            options_.max_conn_refresh_period())) {}
+
+  ~CommonClient() {
+    // This will stop the refresh of the channels.
+    channels_.clear();
+    // TODO(2567): remove this call when the user will have to provide their own
+    // `CompletionQueues`
+    background_threads_->cq().CancelAll();
+  }
 
   /**
    * Reset the channel and stub.
@@ -103,7 +152,7 @@ class CommonClient {
     // introduce attributes in the implementation of CreateChannelPool() to
     // create one socket per element in the pool.
     lk.unlock();
-    auto channels = CreateChannelPool(Traits::Endpoint(options_), options_);
+    auto channels = CreateChannelPool();
     std::vector<StubPtr> tmp;
     std::transform(channels.begin(), channels.end(), std::back_inserter(tmp),
                    [](std::shared_ptr<grpc::Channel> ch) {
@@ -125,6 +174,30 @@ class CommonClient {
     }
   }
 
+  ChannelPtr CreateChannel(std::size_t idx) {
+    auto args = options_.channel_arguments();
+    if (!options_.connection_pool_name().empty()) {
+      args.SetString("cbt-c++/connection-pool-name",
+                     options_.connection_pool_name());
+    }
+    args.SetInt("cbt-c++/connection-pool-id", static_cast<int>(idx));
+    auto res = grpc::CreateCustomChannel(Traits::Endpoint(options_),
+                                         options_.credentials(), args);
+    if (options_.max_conn_refresh_period().count() == 0) {
+      return res;
+    }
+    ScheduleChannelRefresh(cq_, refresh_state_, res);
+    return res;
+  }
+
+  std::vector<std::shared_ptr<grpc::Channel>> CreateChannelPool() {
+    std::vector<std::shared_ptr<grpc::Channel>> result;
+    for (std::size_t i = 0; i != options_.connection_pool_size(); ++i) {
+      result.emplace_back(CreateChannel(i));
+    }
+    return result;
+  }
+
   /// Get the current index for round-robin over connections.
   std::size_t GetIndex() {
     std::size_t current = current_index_++;
@@ -136,10 +209,20 @@ class CommonClient {
   }
 
   std::mutex mu_;
+  std::size_t num_pending_refreshes_{};
   ClientOptions options_;
   std::vector<ChannelPtr> channels_;
   std::vector<StubPtr> stubs_;
   std::size_t current_index_;
+  std::unique_ptr<BackgroundThreads> background_threads_;
+  // Timers, which we schedule for refreshes, need to reference the completion
+  // queue. We cannot make the completion queue's underlying implementation
+  // become owned solely by the operations scheduled on it (because we risk a
+  // deadlock). We solve both problems by holding only weak pointers to the
+  // completion queue in the operations scheduled on it. In order to do it, we
+  // need to hold one instance by a shared pointer.
+  std::shared_ptr<CompletionQueue> cq_;
+  std::shared_ptr<ConnectionRefreshState> refresh_state_;
 };
 
 }  // namespace internal
--- a/google/cloud/bigtable/tests/data_integration_test.cc
+++ b/google/cloud/bigtable/tests/data_integration_test.cc
@@ -600,6 +600,72 @@ TEST_F(DataIntegrationTest, TableApplyWithLogging) {
   google::cloud::LogSink::Instance().RemoveBackend(id);
 }
 
+TEST(ConnectionRefresh, Disabled) {
+  auto client_options = bigtable::ClientOptions().set_max_conn_refresh_period(
+      std::chrono::seconds(0));
+  auto data_client = bigtable::CreateDefaultDataClient(
+      testing::TableTestEnvironment::project_id(),
+      testing::TableTestEnvironment::instance_id(), client_options);
+  // In general, it is hard to show that something has *not* happened, at best
+  // we can show that its side-effects have not happened. In this case we want
+  // to show that the channels have not been refreshed. A side-effect of
+  // refreshing a channel is that it enters the `READY` state, so we check that
+  // this has not taken place and take that as evidence that no refresh has
+  // taken place.
+  //
+  // After the `CompletionQueue` argument is removed from the `Bigtable` API, we
+  // will have an option to provide a mock `CompletionQueue` to the `DataClient`
+  // for test purposes and verify that no timers are created, which will be a
+  // superior way to write this test.
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+  for (std::size_t i = 0; i < client_options.connection_pool_size(); ++i) {
+    auto channel = data_client->Channel();
+    EXPECT_EQ(GRPC_CHANNEL_IDLE, channel->GetState(false));
+  }
+  // Make sure things still work.
+  bigtable::Table table(data_client, testing::TableTestEnvironment::table_id());
+  std::string const row_key = "row-key-1";
+  std::vector<Cell> created{{row_key, kFamily4, "c0", 1000, "v1000"},
+                            {row_key, kFamily4, "c1", 2000, "v2000"}};
+  Apply(table, row_key, created);
+  // After performing some operations, some of the channels should be in ready
+  // state.
+  auto check_if_some_channel_is_ready = [&] {
+    for (std::size_t i = 0; i < client_options.connection_pool_size(); ++i) {
+      if (data_client->Channel()->GetState(false) == GRPC_CHANNEL_READY) {
+        return true;
+      }
+    }
+    return false;
+  };
+  EXPECT_TRUE(check_if_some_channel_is_ready());
+}
+
+TEST(ConnectionRefresh, Frequent) {
+  auto data_client = bigtable::CreateDefaultDataClient(
+      testing::TableTestEnvironment::project_id(),
+      testing::TableTestEnvironment::instance_id(),
+      bigtable::ClientOptions().set_max_conn_refresh_period(
+          std::chrono::milliseconds(100)));
+
+  for (;;) {
+    if (data_client->Channel()->GetState(false) == GRPC_CHANNEL_READY) {
+      // We've found a channel which changed its state from IDLE to READY,
+      // which means that our refreshing mechanism works.
+      break;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
+
+  // Make sure things still work.
+  bigtable::Table table(data_client, testing::TableTestEnvironment::table_id());
+  std::string const row_key = "row-key-1";
+  std::vector<Cell> created{{row_key, kFamily4, "c0", 1000, "v1000"},
+                            {row_key, kFamily4, "c1", 2000, "v2000"}};
+  Apply(table, row_key, created);
+}
+
 }  // namespace
 }  // namespace BIGTABLE_CLIENT_NS
 }  // namespace bigtable


coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550)* status: all files reviewed, 9 unresolved discussions (waiting on @dopiera)

___
*[google/cloud/bigtable/client_options.h, line 302 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-HkGa7FwMfC4UHM3g:-MN-HkGa7FwMfC4UHM3h:b63w004) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/client_options.h#L302)):*
> ```objc
>    * Maximum connection refresh period, as set via `set_max_conn_refresh_period`
>    */
>   std::chrono::seconds max_conn_refresh_period() {
> ```

might be worthwhile to express this in milliseconds, just so the tests are faster... not sure, thoughts?
___
*[google/cloud/bigtable/client_options.cc, line 85 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-AReY8dFqB5y5QqJE:-MN-AReY8dFqB5y5QqJF:b-ir6w23) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/client_options.cc#L85)):*
> ```cpp
>       instance_admin_endpoint_("bigtableadmin.googleapis.com"),
>       // Refresh connections before they expire themselves.
>       max_conn_refresh_period_(std::chrono::seconds(50 * 60)) {
> ```

I think you can say `max_conn_refresh_period_(std::chrono::minutes(50))` which is more readable.  A constant like `kDefaultRefreshPeriod` would be even better. And the connections do not expire by themselves, the server disconnects idle connections.
___
*[google/cloud/bigtable/internal/common_client.h, line 58 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN--eWk6uRmftLWKSHH:-MN--eWk6uRmftLWKSHI:bst4scx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L58)):*
> ```objc
>   explicit CommonClient(bigtable::ClientOptions options)
>       : options_(std::move(options)),
>         rng_(google::cloud::internal::MakeDefaultPRNG()),
> ```

Maybe we are stealing too much entropy for this purpose, consider instead `rng_(std::random_device{}())`
___
*[google/cloud/bigtable/internal/common_client.h, line 150 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN--PC1DUBe1CFD80Z2:-MN--PC1DUBe1CFD80Z3:bt61ul8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L150)):*
> ```objc
>     std::chrono::seconds next_replace_after(
>         std::uniform_int_distribution<std::chrono::seconds::rep>(
>             1, options_.max_conn_refresh_period().count())(rng_));
> ```

Note that `rng_` is not thread safe, should we have a mutex around it?

Consider refactoring the random number generation to a member function, so you can write:

```cc
auto refresh_delay = RandomizedRefreshDelay(...);
background_threads_->cq().MakeRelativeTimer(refresh_delay)...
```
___
*[google/cloud/bigtable/internal/common_client.h, line 153 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-BnZm6yIwWZ3Wocb4:-MN-BnZm6yIwWZ3Wocb5:b-7fzcwx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L153)):*
> ```objc
>     background_threads_->cq()
>         .MakeRelativeTimer(next_replace_after)
>         .then([this, idx](
> ```

Using `this` here seems risky.  Right now the client outlives the completion queue, but that will not be true forever?
___
*[google/cloud/bigtable/internal/common_client.h, line 161 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MMzzT5W2RJZ-j6VO_Jv:-MMzzT5W2RJZ-j6VO_Jw:b-u1cbph) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L161)):*
> ```objc
>           auto channel = CreateChannel(idx);
>           // This is blocking, but it's not a problem given that we're only
>           // blocking a CompletionQueue used solely for refreshing connections.
> ```

Eventually this will be used for the other `Async*()` calls too, right?  Should we add a TODO(#...) here?
___
*[google/cloud/bigtable/internal/common_client.h, line 162 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MMzzojc67oEDLjLWF4H:-MMzzojd3-gb5TqlyLC_:b-27e4yw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L162)):*
> ```objc
>           // This is blocking, but it's not a problem given that we're only
>           // blocking a CompletionQueue used solely for refreshing connections.
>           if (!channel->WaitForConnected(std::chrono::system_clock::now() +
> ```

Maybe we should use `GetState(true)`?  It is not clear from the documentation, but does not seem like it is a blocking operation?

https://grpc.github.io/grpc/cpp/classgrpc_1_1_channel_interface.html#ad2ffa0dffc84f1fe38084c0993c4ca50

Also, please open a bug to replace this with `NotifyOnStateChange()`.
___
*[google/cloud/bigtable/internal/common_client.h, line 163 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-FGH6BflqoTby2C4E:-MN-FGH6BflqoTby2C4F:bhkdop1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L163)):*
> ```objc
>           // blocking a CompletionQueue used solely for refreshing connections.
>           if (!channel->WaitForConnected(std::chrono::system_clock::now() +
>                                          std::chrono::seconds(1))) {
> ```

why is 1 second the right timeout here?  Also, do we need to create a new `Channel` or would calling `GetState(true)` be enough?  Note that we do not block when the pool is first created, it is not clear to me that we must block when refreshing.
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 568 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-A4XLBr92iK6rNlry:-MN-A4XLBr92iK6rNlrz:bp019df) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/tests/data_integration_test.cc#L568)):*
> ```cpp
>   // Make sure something would have failed if were in fact replacing
>   // connections.
>   std::this_thread::sleep_for(std::chrono::seconds(1));
> ```

This is likely to be flaky... the CI machines can become very overloaded.  One approach might be to create a timer to expire *after* the refresh period.  Because there is a single thread in the CQ, the timer would expire *after* the refresh timer has completed, and you can reasonably expect the side-effects of the refresh timer to be visible.

Consider adding a counter with the number of refreshes, maybe only accessible through a `friend` function in `internal::`, to DataClient.

That way you can verify something actually happened.

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550)* status: all files reviewed, 9 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigtable/client_options.h, line 302 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-HkGa7FwMfC4UHM3g:-MNYq7NY8fQe1YS7smJ2:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/client_options.h#L302)):*

might be worthwhile to express this in milliseconds, just so the tests are faster... not sure, thoughts?
</blockquote></details>

Done.
___
*[google/cloud/bigtable/client_options.cc, line 85 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-AReY8dFqB5y5QqJE:-MNYqiUG3-tDpDyIs4_t:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/client_options.cc#L85)):*

I think you can say `max_conn_refresh_period_(std::chrono::minutes(50))` which is more readable.  A constant like `kDefaultRefreshPeriod` would be even better. And the connections do not expire by themselves, the server disconnects idle connections.
</blockquote></details>

Done.
___
*[google/cloud/bigtable/internal/common_client.h, line 58 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN--eWk6uRmftLWKSHH:-MNYquGuCon22_lpf-CY:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L58)):*

Maybe we are stealing too much entropy for this purpose, consider instead `rng_(std::random_device{}())`
</blockquote></details>

Done.
___
*[google/cloud/bigtable/internal/common_client.h, line 150 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN--PC1DUBe1CFD80Z2:-MNYsZ5hCmfFiR-SczIw:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L150)):*

Note that `rng_` is not thread safe, should we have a mutex around it?

Consider refactoring the random number generation to a member function, so you can write:

```cc
auto refresh_delay = RandomizedRefreshDelay(...);
background_threads_->cq().MakeRelativeTimer(refresh_delay)...
```


</blockquote></details>

Done.
___
*[google/cloud/bigtable/internal/common_client.h, line 153 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-BnZm6yIwWZ3Wocb4:-MN_Zn-q3BRF7EDJgIAy:b5v1lmr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L153)):*

Using `this` here seems risky.  Right now the client outlives the completion queue, but that will not be true forever?
</blockquote></details>

It will likely not, fixed.
___
*[google/cloud/bigtable/internal/common_client.h, line 161 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MMzzT5W2RJZ-j6VO_Jv:-MNYwuVGExCZxGuVMR3R:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L161)):*

Eventually this will be used for the other `Async*()` calls too, right?  Should we add a TODO(#...) here?
</blockquote></details>

Done.
___
*[google/cloud/bigtable/internal/common_client.h, line 162 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MMzzojc67oEDLjLWF4H:-MNYwr5x552vAASWs3n5:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L162)):*

Maybe we should use `GetState(true)`?  It is not clear from the documentation, but does not seem like it is a blocking operation?

https://grpc.github.io/grpc/cpp/classgrpc_1_1_channel_interface.html#ad2ffa0dffc84f1fe38084c0993c4ca50

Also, please open a bug to replace this with `NotifyOnStateChange()`.

</blockquote></details>

Done.
___
*[google/cloud/bigtable/internal/common_client.h, line 163 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-FGH6BflqoTby2C4E:-MN_UpKQCJiD8wc73PmD:b9fwj0s) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/internal/common_client.h#L163)):*

why is 1 second the right timeout here?  Also, do we need to create a new `Channel` or would calling `GetState(true)` be enough?  Note that we do not block when the pool is first created, it is not clear to me that we must block when refreshing.
</blockquote></details>

Actually it was supposed to be 10s - this is what Java client does. As for `GetState(true)` - yes, that's enough.
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 568 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN-A4XLBr92iK6rNlry:-MN_ekLt2D4uNpe10rJd:bkoizam) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/625fb0fde428e5827b9d1c8c605c8933b99b00c6/google/cloud/bigtable/tests/data_integration_test.cc#L568)):*

This is likely to be flaky... the CI machines can become very overloaded.  One approach might be to create a timer to expire *after* the refresh period.  Because there is a single thread in the CQ, the timer would expire *after* the refresh timer has completed, and you can reasonably expect the side-effects of the refresh timer to be visible.

Consider adding a counter with the number of refreshes, maybe only accessible through a `friend` function in `internal::`, to DataClient.

That way you can verify something actually happened.

</blockquote></details>

This is the "Disabled" case, so there's always a possibility of a bad thing to happen right after we would have finished the test. Anyway, for better confidence, I started inspecting the channels' states. Do you think this is enough of a check?

coryan (changes requested):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550)* status: all files reviewed, 7 unresolved discussions (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.h, line 143 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_hr5d6zs84dlYlKgM:-MN_hr5d6zs84dlYlKgN:bt2ksj1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L143)):*
> ```objc
>     return std::chrono::milliseconds(
>         std::uniform_int_distribution<std::chrono::milliseconds::rep>(
>             1, options_.max_conn_refresh_period().count())(rng_));
> ```

This is using system-wide entropy on each call, effectively reading from `/dev/urandom` (or the RND instruction or RDSEED or something of that sort).  I do not think that is best practice.

Furthermore, I am not sure if this is called by multiple threads, but seems to be because the caller takes care to put locks around other things: `std::random_device` is **not** thread safe.

I think what you want is:

- Initialize a `google::cloud::internal::DefaultPRNG` with a `std::random_device` in the constructor.
- Protect uses of that `google::cloud::internal::DefaultPRNG` with a mutex.
___
*[google/cloud/bigtable/internal/common_client.h, line 174 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_r3fv9NZsiyBd2MCu:-MN_r3fw1k5iQnXfTGmN:b-qcw9s6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L174)):*
> ```objc
> 
>     timer_fut.then(
>         [this,
> ```

Remind me why is using `this` safe in this context?  Ah, because the completion queue is owned by this object so the callback cannot be used after this object is deleted?
___
*[google/cloud/bigtable/internal/common_client.h, line 175 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_qtSG4dkD9d0x0KLq:-MN_qtSH5lyH7j3iw37w:blxh6q8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L175)):*
> ```objc
>     timer_fut.then(
>         [this,
>          idx](future<StatusOr<std::chrono::system_clock::time_point>> fut) {
> ```

Consider a type alias for the long future (or the `T` inside `future<T>`), I think it is unfortunate this type is so long to spell out, C++14 `auto` sure would be handy here.
___
*[google/cloud/bigtable/internal/common_client.h, line 184 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_rN6X127Bn5R7yRks:-MN_rN6Y0aShk8owCS0e:b-ym1tsm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L184)):*
> ```objc
>               .AsyncWaitConnectionReady(
>                   channels_[idx],
>                   std::chrono::system_clock::now() + std::chrono::seconds(10))
> ```

Move that `std::chrono::seconds(10)` to a `constexpr` variable so we know why 10 is the number of the counting... ?
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 615 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_ss_wAfeJTPODyePE:-MN_ss_wAfeJTPODyePF:bweihd7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/tests/data_integration_test.cc#L615)):*
> ```cpp
>   for (std::size_t i = 0; i < 10U; ++i) {
>     auto channel = data_client->Channel();
>     // It's not a proof but none of the selected channels was actively warmed
> ```

:thinking: 
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 624 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_sjv26TSeLDKH1hpG:-MN_sjv26TSeLDKH1hpH:blmyjdy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/tests/data_integration_test.cc#L624)):*
> ```cpp
>   std::vector<Cell> created{{row_key, kFamily4, "c0", 1000, "v1000"},
>                             {row_key, kFamily4, "c1", 2000, "v2000"}};
>   Apply(table, row_key, created);
> ```

Would you expect that at least one channel is not IDLE after this?  I guess that could flake?
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 633 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_tCkw1Z9dF0iS2F_7:-MN_tCkw1Z9dF0iS2F_8:bqxr9ak) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/tests/data_integration_test.cc#L633)):*
> ```cpp
>       bigtable::ClientOptions().set_max_conn_refresh_period(
>           std::chrono::milliseconds(100)));
> ```

Should we wait at least 100ms?  Do we have any reason to believe it will refresh any sooner?

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550)* status: all files reviewed, 7 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigtable/internal/common_client.h, line 143 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_hr5d6zs84dlYlKgM:-MNcS3BMCpMQekkbqbxs:b-g7hbr4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L143)):*

This is using system-wide entropy on each call, effectively reading from `/dev/urandom` (or the RND instruction or RDSEED or something of that sort).  I do not think that is best practice.

Furthermore, I am not sure if this is called by multiple threads, but seems to be because the caller takes care to put locks around other things: `std::random_device` is **not** thread safe.

I think what you want is:

- Initialize a `google::cloud::internal::DefaultPRNG` with a `std::random_device` in the constructor.
- Protect uses of that `google::cloud::internal::DefaultPRNG` with a mutex.


</blockquote></details>

Oh, I misunderstood you then - I thought you said before to replace the `DefaultPRNG` with `random_device`. I fixed it now.

It has however been protected by a mutex. The `RandomizedRefreshDelay` memfn is called only from `ScheduleChannelRefresh()` with the mutex held. I guess concurrency annotations would be useful here.

Anyway, I discovered a problem - the destructor could potentially block for a couple of minutes when the completion queue is not for this class' sole use. If the `CompletionQueue` is shared with some other component, we can't call `CancelAll()`, because there might be other operation too. This means we either block in the `CommonClient` dtor (for several minutes) or we leave the timers in the `CompletionQueue` and make them noops. The latter is much more compelling, so I now reworked it to do that.
___
*[google/cloud/bigtable/internal/common_client.h, line 174 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_r3fv9NZsiyBd2MCu:-MNcTiFd51pT_u0YoQv5:b-1330n) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L174)):*

Remind me why is using `this` safe in this context?  Ah, because the completion queue is owned by this object so the callback cannot be used after this object is deleted?
</blockquote></details>

The reason is different, i.e. I ensure that this object outlives the timers. I make sure that happens by tracking their number and waiting for this number to drop to 0 in the destructor.

I added a comment in the code for our future selves now.
___
*[google/cloud/bigtable/internal/common_client.h, line 175 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_qtSG4dkD9d0x0KLq:-MNcUX4KBWJZDK8_aP7J:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L175)):*

Consider a type alias for the long future (or the `T` inside `future<T>`), I think it is unfortunate this type is so long to spell out, C++14 `auto` sure would be handy here.
</blockquote></details>

Done
___
*[google/cloud/bigtable/internal/common_client.h, line 184 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_rN6X127Bn5R7yRks:-MNcW5-T1CwLhqcndKvH:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L184)):*

Move that `std::chrono::seconds(10)` to a `constexpr` variable so we know why 10 is the number of the counting... ?
</blockquote></details>

Done.
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 615 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_ss_wAfeJTPODyePE:-MNcYcwB6_WH0mc8f2R5:b-4an05x) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/tests/data_integration_test.cc#L615)):*

:thinking: 
</blockquote></details>

Having reread the comment, I agree it's not great. I tried to rewrite it again, PTAL.
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 624 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_sjv26TSeLDKH1hpG:-MNcYxwY2AzYnjzjIcWI:b-nij41h) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/tests/data_integration_test.cc#L624)):*

Would you expect that at least one channel is not IDLE after this?  I guess that could flake?
</blockquote></details>

That could flake only if the test got frozen for 4 minutes, which I think is longer than the test timeout, so I think that the flakiness likelihood is low enough to ignore it. I added the check.
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 633 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_tCkw1Z9dF0iS2F_7:-MNcb0UP0xuJEDuY_Aa1:bdx9ank) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/tests/data_integration_test.cc#L633)):*

Should we wait at least 100ms?  Do we have any reason to believe it will refresh any sooner?
</blockquote></details>

We pick the refresh periods randomly (but no longer than 100ms). Probability calculus is hard, but I'm guessing we'll be typically looking at 10ms or so (depending on the number of CPUs).

coryan (changes requested):
Still have some thread safety issues below.

> Reviewed 1 of 5 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-:-MNdKMMX6hj8_CiQmgRg:btlpokw)* status: all files reviewed, 4 unresolved discussions (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.h, line 143 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_hr5d6zs84dlYlKgM:-MNdJKpY8ayY40Yl6TT_:befshnf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L143)):*

Oh, I misunderstood you then - I thought you said before to replace the `DefaultPRNG` with `random_device`. I fixed it now.

It has however been protected by a mutex. The `RandomizedRefreshDelay` memfn is called only from `ScheduleChannelRefresh()` with the mutex held. I guess concurrency annotations would be useful here.

Anyway, I discovered a problem - the destructor could potentially block for a couple of minutes when the completion queue is not for this class' sole use. If the `CompletionQueue` is shared with some other component, we can't call `CancelAll()`, because there might be other operation too. This means we either block in the `CommonClient` dtor (for several minutes) or we leave the timers in the `CompletionQueue` and make them noops. The latter is much more compelling, so I now reworked it to do that.
</blockquote></details>

But you still cannot share the completion queue with other components. And that precludes receiving the CQ from the application... The current change is Okay, but we will have trouble in the future.
___
*[google/cloud/bigtable/internal/common_client.h, line 174 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_r3fv9NZsiyBd2MCu:-MNd9XkB6jcHPFXh1rQx:b-9hoo4w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L174)):*

The reason is different, i.e. I ensure that this object outlives the timers. I make sure that happens by tracking their number and waiting for this number to drop to 0 in the destructor.

I added a comment in the code for our future selves now.


</blockquote></details>

Well, technically (and as you probably know), the lifetime of an object ends when the destructor *starts*, so the use of `this` is maybe a technical violation of the spec.  Good news: the lifetime of each member variable of the object has not ended until the destructor *returns*, so things were probably fine-ish.
___
*[google/cloud/bigtable/internal/common_client.h, line 39 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNdEzmD-8wdcNjMjpxu:-MNdEzmD-8wdcNjMjpxv:b8bes8k) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c9bc50037a480f4df967c6d7c56a439ba1da374/google/cloud/bigtable/internal/common_client.h#L39)):*
> ```objc
>  * since waiting for a connection to become ready is not cancellable.
>  */
> constexpr std::chrono::seconds kConnectionReadyTimeout(10);
> ```

consider:

```cc
auto constexpr kConnectionReadyTimeout = std::chrono::seconds(10);
```

in any case, we are `east const` in this project, the `constexpr` goes after the type.
___
*[google/cloud/bigtable/internal/common_client.h, line 82 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNdJrAU12DOJGNu8TO7:-MNdJrAU12DOJGNu8TO8:bzgyfax) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c9bc50037a480f4df967c6d7c56a439ba1da374/google/cloud/bigtable/internal/common_client.h#L82)):*
> ```objc
>     // This will stop the refresh of the channels.
>     channels_.clear();
>     background_threads_->cq().CancelAll();
> ```

Seems like this `CancelAll()` deserves some kind of `TODO(#...) - remove this when the CQ is shared` ?
___
*[google/cloud/bigtable/internal/common_client.cc, line 25 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNdDRMV2jChYIKHZRD5:-MNdDRMW97X2LDbwP-1I:bnlso47) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c9bc50037a480f4df967c6d7c56a439ba1da374/google/cloud/bigtable/internal/common_client.cc#L25)):*
> ```cpp
> std::chrono::milliseconds RandomizedRefreshDelay(
>     std::chrono::milliseconds max_conn_refresh_period) {
>   static google::cloud::internal::DefaultPRNG rng(std::random_device{}());
> ```

Now that it is static, this is no longer protected by a single mutex, that is also a problem, right?
___
*[google/cloud/bigtable/internal/common_client.cc, line 52 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNdDnaJB2BPEoedNZ52:-MNdDnaJB2BPEoedNZ53:b-smghlk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c9bc50037a480f4df967c6d7c56a439ba1da374/google/cloud/bigtable/internal/common_client.cc#L52)):*
> ```cpp
>         }
>         auto channel = weak_channel.lock();
>         if (!channel) {
> ```

FYI: we have started writing short statements like this as:

```cc
if (!channel) return;
```
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 609 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNdEDoM0HAYxlLc9sfc:-MNdEDoM0HAYxlLc9sfd:b-1fj3ua) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c9bc50037a480f4df967c6d7c56a439ba1da374/google/cloud/bigtable/tests/data_integration_test.cc#L609)):*
> ```cpp
>       testing::TableTestEnvironment::project_id(),
>       testing::TableTestEnvironment::instance_id(), client_options);
>   // There doesn't seem to be a reliable way to check if we're really not
> ```

You could make a comment about the difficulty of proving a negative in general, something like:

```cc
// In general, it is hard to show that something has *not* happened, at best we can show that
// its side-effects have not happened.  In this case we want to show that the channels have not
// been refreshed.  A side-effect of refreshing a channel is that it enters the `READY` state, so
// we check that this has not taken place and take that as evidence that no refresh has taken
// place.
```

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550)* status: all files reviewed, 4 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigtable/internal/common_client.h, line 143 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_hr5d6zs84dlYlKgM:-MNe1Ej85oe8kwdzh0O2:bxejudw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L143)):*

But you still cannot share the completion queue with other components. And that precludes receiving the CQ from the application... The current change is Okay, but we will have trouble in the future.
</blockquote></details>

Sorry, I think I'm missing something. This change (i.e. today's rework) is what allows `CommonClient` to operate on a shared `CompletionQueue`. Before this change, we had to either control the `CompletionQueue`'s lifetime or block in the dtor, now there is no such requirement (instead, some noop timers are left scheduled). Why are you saying that we will have trouble in the future?
___
*[google/cloud/bigtable/internal/common_client.h, line 174 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_r3fv9NZsiyBd2MCu:-MNe5fOnCrg6V8w_ySCM:b9lgpew) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L174)):*

Well, technically (and as you probably know), the lifetime of an object ends when the destructor *starts*, so the use of `this` is maybe a technical violation of the spec.  Good news: the lifetime of each member variable of the object has not ended until the destructor *returns*, so things were probably fine-ish.
</blockquote></details>

Yup. Either way, I got rid of this ugliness.
___
*[google/cloud/bigtable/internal/common_client.h, line 39 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNdEzmD-8wdcNjMjpxu:-MNe0DZj1YWC3OqOzfNT:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c9bc50037a480f4df967c6d7c56a439ba1da374/google/cloud/bigtable/internal/common_client.h#L39)):*

consider:

```cc
auto constexpr kConnectionReadyTimeout = std::chrono::seconds(10);
```

in any case, we are `east const` in this project, the `constexpr` goes after the type.

</blockquote></details>

Done.
___
*[google/cloud/bigtable/internal/common_client.h, line 82 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNdJrAU12DOJGNu8TO7:-MNe0qsu5N41InvV6Vje:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c9bc50037a480f4df967c6d7c56a439ba1da374/google/cloud/bigtable/internal/common_client.h#L82)):*

Seems like this `CancelAll()` deserves some kind of `TODO(#...) - remove this when the CQ is shared` ?
</blockquote></details>

Done.
___
*[google/cloud/bigtable/internal/common_client.cc, line 25 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNdDRMV2jChYIKHZRD5:-MNe4DvqAoV44QZdxjpl:b-4l2zrk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c9bc50037a480f4df967c6d7c56a439ba1da374/google/cloud/bigtable/internal/common_client.cc#L25)):*

Now that it is static, this is no longer protected by a single mutex, that is also a problem, right?
</blockquote></details>

Yes - I introduced the same bug in this PR twice :facepalm: 

Fixed
___
*[google/cloud/bigtable/tests/data_integration_test.cc, line 609 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNdEDoM0HAYxlLc9sfc:-MNe3yOa5YH11TuT5lwh:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c9bc50037a480f4df967c6d7c56a439ba1da374/google/cloud/bigtable/tests/data_integration_test.cc#L609)):*

You could make a comment about the difficulty of proving a negative in general, something like:

```cc
// In general, it is hard to show that something has *not* happened, at best we can show that
// its side-effects have not happened.  In this case we want to show that the channels have not
// been refreshed.  A side-effect of refreshing a channel is that it enters the `READY` state, so
// we check that this has not taken place and take that as evidence that no refresh has taken
// place.
```
</blockquote></details>

Done.

coryan (changes requested):
> Reviewed 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550)* status: all files reviewed, 1 unresolved discussion (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.h, line 143 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_hr5d6zs84dlYlKgM:-MNeQ-e028rTmzP5B4jy:b-dmead) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L143)):*

Sorry, I think I'm missing something. This change (i.e. today's rework) is what allows `CommonClient` to operate on a shared `CompletionQueue`. Before this change, we had to either control the `CompletionQueue`'s lifetime or block in the dtor, now there is no such requirement (instead, some noop timers are left scheduled). Why are you saying that we will have trouble in the future?
</blockquote></details>

Because calling `cq().CancelAll()` may cancel timers and other operations that we have no knowledge of?  And might impact how the application works?  Consider code like this:

```cc
void F(CompletionQueue cq) {
  // cq is already in use, there are timers and other things working on it.

  bigtable::Table table(... /* somehow we get to share `cq` */ cq ...);
  // Do stuff with `table`

  // the destructor for `table` will call cq.CancelAll(), that destroys timers and other operations
}
```
___
*[google/cloud/bigtable/internal/common_client.cc, line 25 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNeS6um1exRpYEkSsXf:-MNeS6un9B0hNfuWnYOC:bu17dfc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4d8ac33a05a07f810fbf3ea4cf9cab12b6cba864/google/cloud/bigtable/internal/common_client.cc#L25)):*
> ```cpp
> std::chrono::milliseconds RandomizedRefreshDelay(
>     std::chrono::milliseconds max_conn_refresh_period) {
>   static std::mutex mu;
> ```

And now you have a problem with the google style guide that prohibits static objects with non-trivial destructors....

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[google/cloud/bigtable/internal/common_client.h, line 143 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_hr5d6zs84dlYlKgM:-MNeZN1z1NkQT92z45RL:b-rbsds1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L143)):*

Because calling `cq().CancelAll()` may cancel timers and other operations that we have no knowledge of?  And might impact how the application works?  Consider code like this:

```cc
void F(CompletionQueue cq) {
  // cq is already in use, there are timers and other things working on it.

  bigtable::Table table(... /* somehow we get to share `cq` */ cq ...);
  // Do stuff with `table`

  // the destructor for `table` will call cq.CancelAll(), that destroys timers and other operations
}
```

</blockquote></details>

Oh, yes, that's true. My secret plan was to make `DefaultBackgroundThreads` call `CancellAll` in its destructor as part of the `CompletionQueue` argument removal and stop calling `CancelAll()` here. Does that make sense?
___
*[google/cloud/bigtable/internal/common_client.cc, line 25 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNeS6um1exRpYEkSsXf:-MNeZdu24a3cnPLnoFe8:b-hi9yf9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4d8ac33a05a07f810fbf3ea4cf9cab12b6cba864/google/cloud/bigtable/internal/common_client.cc#L25)):*

And now you have a problem with the google style guide that prohibits static objects with non-trivial destructors....
</blockquote></details>

Ooops, fixed.

coryan (approved):
:lgtm:

> Reviewed 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-:-MNf2N7iA_6YG9a-Rc1r:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[google/cloud/bigtable/internal/common_client.h, line 143 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MN_hr5d6zs84dlYlKgM:-MNf19JqBK458ECbBBcu:b-3mlwp9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/deec432d10459ed1121fb51e58d66ea5124ccbad/google/cloud/bigtable/internal/common_client.h#L143)):*

Oh, yes, that's true. My secret plan was to make `DefaultBackgroundThreads` call `CancellAll` in its destructor as part of the `CompletionQueue` argument removal and stop calling `CancelAll()` here. Does that make sense?
</blockquote></details>

Ack.  That is why I thought the change was fine, but we would need some sort of additional change later.
___
*[google/cloud/bigtable/internal/common_client.cc, line 59 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNf1Yv99Vw2pzbQCvnv:-MNf1Yv99Vw2pzbQCvnw:b3qf0zt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5386f75d05b36ebd1692ce1ae9ea5179c1933cc0/google/cloud/bigtable/internal/common_client.cc#L59)):*
> ```cpp
>         cq->AsyncWaitConnectionReady(channel, std::chrono::system_clock::now() +
>                                                   kConnectionReadyTimeout)
>             .then([weak_channel, weak_cq, state](future<Status> fut) {
> ```

FYI: Almost feels like `ScheduleChannelRefresh()` should be a member function of `ConnectionRefreshState`.  Except that then you would need `shared_from_this()`, and the `shared_ptr<>` cycles are too easy to introduce.

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.cc, line 59 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5550#-MNf1Yv99Vw2pzbQCvnv:-MO6iiTz1_3m9_pMf2UI:bo6xyjw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5386f75d05b36ebd1692ce1ae9ea5179c1933cc0/google/cloud/bigtable/internal/common_client.cc#L59)):*

FYI: Almost feels like `ScheduleChannelRefresh()` should be a member function of `ConnectionRefreshState`.  Except that then you would need `shared_from_this()`, and the `shared_ptr<>` cycles are too easy to introduce.
</blockquote></details>

Yup, also I think `shared_from_this<>` is particularly ugly when combined with inheritance (especially that I don't think we'd have a way to enforce subclasses to only be created via a `shared_ptr<>`.

PR #5552, opened 2020-11-25
feat(pubsub): by default Subscriber::Subscribe retries forever

### LINKED ISSUE #5513: Consider changing default retry policy for Subscription::Subscribe()
It seems that the more appropriate default policy is to try basically forever (a few billion times would have the same effect), and then let the application change the policy if they have good reasons to prefer retrying only a certain number of times, or only for some time.

coryan:
Subscribers have a single operation (`Subscriber::Subscribe()`), after
some discussion, we are changing this to retry basically forever by
default. The application can change this behavior if they want to, and
that is an easier behavior to change than the hard-coded message to
treat some errors as "not really a failure".

Fixes #5513

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5552)


google/cloud/pubsub/subscriber_connection.cc
```diff
@@ -121,8 +123,17 @@ std::shared_ptr<pubsub::SubscriberConnection> MakeSubscriberConnection(
     std::vector<std::shared_ptr<SubscriberStub>> stubs,
     std::unique_ptr<pubsub::RetryPolicy const> retry_policy,
     std::unique_ptr<pubsub::BackoffPolicy const> backoff_policy) {
+  auto default_retry_policy = [] {
+    // Subscribers are special, by default we want to retry essentially forever
```
devbww, line 127: s/, by/. By/

Or s/, by/: by/ perhaps?

devbww (approved):

google/cloud/pubsub/subscriber_connection.cc
coryan: Done (the second).

PR #5554, opened 2020-11-26
refactor: make StreamRange<T> public

devjgm:
`StreamRange<T>` is a user-facing type that our public APIs return to
callers. Users must be able to read code/documentation to learn how to
use this type. Therefore, `StreamRange<T>` should not live in an
`internal` namespace, otherwise, we're forcing our users to reach into
our internal namespace to use a type that we give them.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5554)


coryan (commented):
There are downsides to making the type public: it becomes part of the API, we cannot delete it or replace it by something else in the future.  The refactoring you just did would not be possible because we can remove things from `internal::` but not from the public API (well, we can, but should really try to avoid it).

As another example: renaming `StreamRange<T>::Iterator` to `StreamRange<T>::IteratorImpl` would require declaring a breaking change.


> `StreamRange<T>` is a user-facing type that our public APIs return to
> callers.

Is it?  I would argue that we return specific instantiations of `StreamRange<>`.  Maybe the problem is that we have been abusing type aliases for this, maybe should always do:

```cc
struct FooStream : public internal::StreamRange<Foo> {};
```

or (I am not sure the `@copydoc` trick works):

```cc
/// @copydoc internal::StreamRange<Foo>
struct FooStream : private internal::StreamRange<Foo> /* private inheritance *IS* composition */
{
  /// A single-pass [Input Iterator][input-iterator] with value type `StatusOr<Foo>`.
  ///
  /// [input-iterator]: https://en.cppreference.com/w/cpp/named_req/InputIterator
  using  internal::StreamRange<Foo>::iterator;

  /// Return the beginning of the range
  using internal::begin;
  /// Return the beginning of the range
  using internal::end;
};
```

> Users must be able to read code/documentation to learn how to
> use this type.

The lack of documentation is a problem, I agree we want to fix it.

> Therefore, `StreamRange<T>` should not live in an
> `internal` namespace, otherwise, we're forcing our users to reach into
> our internal namespace to use a type that we give them.

Any thoughts on the approaches sketched above?  Or maybe we want to double check the public members and types and make sure nothing unwanted is leaking as part of this change?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5554#-:-MN3msQK7yzq5AjI_2Tg:b7zvngj)* status: :shipit: complete! all files reviewed, all discussions resolved

devjgm:
Good points to consider. I do think there are some problems with our current approach of exposing internal types via public aliases, but you've suggested several alternative ideas to consider.

I'm going to close this PR for now until we can discuss it more next week in a VC or something. Thanks.

PR #5557, opened 2020-11-30
fix(pubsub): remove race condition between ack() and session.cancel()

bjornleffler:
If you copy this sample code and try to use it in production with the GCP pubsub service, some messages will not be correctly ack'ed. The messages will be re-delivered after the (default of 10s) ackDeadline.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5557)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -379,11 +379,11 @@ void ReceiveDeadLetterDeliveryAttempt(
         [&](pubsub::Message const& m, pubsub::AckHandler h) {
           std::cout << "Received message " << m << "\n";
           std::cout << "Delivery attempt: " << h.delivery_attempt() << "\n";
+          std::move(h).ack();
           std::unique_lock<std::mutex> lk(mu);
           ++message_count;
           lk.unlock();
           cv.notify_one();
-          std::move(h).ack();
         });
     // Wait until at least one message has been received.
     std::unique_lock<std::mutex> lk(mu);
@@ -838,11 +838,11 @@ void Subscribe(google::cloud::pubsub::Subscriber subscriber,
     auto session = subscriber.Subscribe(
         [&](pubsub::Message const& m, pubsub::AckHandler h) {
           std::cout << "Received message " << m << "\n";
+          std::move(h).ack();
           std::unique_lock<std::mutex> lk(mu);
           ++message_count;
           lk.unlock();
           cv.notify_one();
-          std::move(h).ack();
         });
     // Wait until at least one message has been received.
     std::unique_lock<std::mutex> lk(mu);
@@ -877,12 +877,12 @@ void SubscribeErrorListener(google::cloud::pubsub::Subscriber subscriber,
         subscriber
             .Subscribe([&](pubsub::Message const& m, pubsub::AckHandler h) {
               std::cout << "Received message " << m << "\n";
+              std::move(h).ack();
               std::unique_lock<std::mutex> lk(mu);
               ++message_count;
               done = true;
               lk.unlock();
               cv.notify_one();
-              std::move(h).ack();
             })
             // Setup an error handler for the subscription session
             .then([&](future<google::cloud::Status> f) {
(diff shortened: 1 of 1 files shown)


remyabel:
To make conventional commits happy, you'll need to change your title to something like: `fix(pubsub): ...` and remove the period at the end

coryan (commented):
> If you copy this sample code and try to use it in production with the GCP pubsub service, some messages will not be correctly ack'ed. The messages will be re-delivered after the (default of 10s) ackDeadline.

I do not think these changes fix that.   Acks are potentially delayed (by tens of milliseconds), so the application could shutdown before they are sent.  And acks are best-effort anyway, the client library may be unable to send the acks because of some transient error.  And the service may redeliver the messages even if the ack is successfully sent (and delivered), if the service could avoid that it would have stronger guarantees than "at least once" delivery.

I am happy to merge the changes if they make all those problems *less likely*, but they are not going to make them go away.

Separately, if you could tell me more about the context that these changes were copied into, that might help me think of alternative solutions. Things like "Is the application going to stop after receiving N messages or keep running forever?"

+kokoro:run 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5557#-:-MNORKD56ERpamRRNJeo:b-vvk2iw)* status: :shipit: complete! all files reviewed, all discussions resolved

bjornleffler:
I agree that my change doesn't fix anything for the samples.cc file. It does fix issues for anyone that tries to copy the code and use it elsewhere. I tried to copy the code from samples.cc and was surprised that about half the messages were delivered 2-4 times and took 10s of seconds.

Ultimately, my application just receives pubsub messages in an infinite loop. Multiple deliveries happened when I tried to receive a single message within an infinite loop. Mostly to check if pubsub was working.

coryan:
> I agree that my change doesn't fix anything for the samples.cc file. It does fix issues for anyone that tries to copy the code and use it elsewhere. I tried to copy the code from samples.cc and was surprised that about half the messages were delivered 2-4 times and took 10s of seconds.

I understand, but I think the problem with our sample code is elsewhere.  Acking the messages earlier may hide a deeper problem.  We are probably showing (or not showing) some other aspect of the problem that is more important.

> 
> Ultimately, my application just receives pubsub messages in an infinite loop.

Why a loop?  I would expect that most applications that receive events "forever" do something like this:

```cc
void DoStuff(pubsub::Message const& m);

void ReceiveForever(pubsub::Subscriber s) {
  auto session = s.Subscribe([](pubsub::Message const& m, pubsub::AckHandler h) {
    DoStuff(m);
    std::move(h).ack();
  }).then([](auto f) {
    std::cerr << "Error while waiting for session: " << f.get() << std::endl;
  });
 session.get();
}
```

Our examples are maybe too complicated by the fact that they need to stop eventually.  We may need to change the base case to really run forever, or maybe we need to stop the examples outside the example code.

> Multiple deliveries happened when I tried to receive a single message within an infinite loop.

Depending on how you wrote the infinite loop that might be a more serious bug in the library (or not).  If you create and delete the `pubsub::Subscriber` inside the loop then missing `.ack()` calls is expected.  If you create a single `pubsub::Subscriber` and re-use it inside the loop missing `.acks()` calls should be rare.

In other words, if you wrote your infinite loop like so:


```cc
void L1() {
  for (;;) {
    auto subscriber = pubsub::Subscriber(...);
    auto session = subscriber.Subscribe(...);
     // handle a few events and then run the loop again
  }
}
```

Then we need to change the examples to show that this is **not** how you should be writing the loops (in fact, we don't need you to write loops in the first place).

If you wrote the loop like so:

```cc
void L2() {
  auto subscriber = pubsub::Subscriber(...);
  for (;;) {
    auto session = subscriber.Subscribe(...);
     // handle a few events and then run the loop again
  }
}
```

and you are still seeing a lot of resends then the library needs some work.

>Mostly to check if pubsub was working.

Ack (pun intended).

bjornleffler:
Thank you for explaining all that so clearly. Sounds like I should re-write the code using your ReceiveForever() logic.

I wrote the code like the following, thinking that if the session somehow failed, then the loop would create another one and keep going.
```
void loop() {
  auto subscriber = pubsub::Subscriber(...);
  while (true) {
    auto session = subscriber.Subscribe(...);
    // Wait until at least one message has been received.
    session.cancel();
  }
}
```

I'll write a much smaller test to see if I can reproduce this.

coryan:
@bjornleffler I am inclined to close this PR.  I still think we need to change our examples to steer users into the easier implementation path (see #5598), but this change is not doing that and not really resolving the issue.  Thoughts?

bjornleffler:
Yes, please close the PR. Thanks again for your help and explanations.

PR #5558, opened 2020-11-30
fix: inspect returned grpc::Status

devjgm:
In some environments `grpc::Status` is marked as "must_use_result", and
these test lines would cause a breakage due to `-Wunused-result` if we
didn't do something with the `grpc::Status` that's returned from these
`LogWrapper()` calls.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5558)

### DIFF
--- a/google/cloud/internal/log_wrapper_test.cc
+++ b/google/cloud/internal/log_wrapper_test.cc
@@ -296,7 +296,8 @@ TEST(LogWrapper, StatusValue) {
   grpc::ClientContext context;
   btproto::MutateRowRequest request;
   btproto::MutateRowResponse response;
-  LogWrapper(mock, &context, request, &response, "in-test", {});
+  auto const r = LogWrapper(mock, &context, request, &response, "in-test", {});
+  EXPECT_TRUE(r.ok());
 
   auto const log_lines = backend->ClearLogLines();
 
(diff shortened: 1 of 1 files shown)


google/cloud/internal/log_wrapper_test.cc
```diff
@@ -317,7 +318,9 @@ TEST(LogWrapper, StatusValueError) {
   grpc::ClientContext context;
   btproto::MutateRowRequest request;
   btproto::MutateRowResponse response;
-  LogWrapper(mock, &context, request, &response, "in-test", {});
+  auto const r = LogWrapper(mock, &context, request, &response, "in-test", {});
+  EXPECT_EQ(r.error_code(), status.error_code());
```
coryan, line 322: I would use `EXPECT_EQ(status, r);` in this case.  Also consider `StatusIs()` which may produce better error messages.

coryan (approved):

google/cloud/internal/log_wrapper_test.cc
devjgm: The external `grpc::Status` doesn't have `op==`, and `StatusIs` doesn't work with `grpc::Status`, which has differently named methods like `error_message()` instead of `message(). I tried both of those already :smile:

coryan: Doh!

PR #5578, opened 2020-12-03
fix(storage): allow endings without \r\n in emulator multipart

shaffeeullah:
The Storage multipart upload API allows for a multipart ending without "\r\n". The [RFC](https://tools.ietf.org/html/rfc7578) also does not specify that "\r\n” is required. This change allows for this flexibility in the emulator as well.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5578)


coryan (approved):
Thanks!

PR #5592, opened 2020-12-08
feat(storage): implement "public access prevention"

### LINKED ISSUE #5491: storage: add support for PublicAccessPrevention
**What component of `google-cloud-cpp` is this feature request for?**
Storage

We need to add support for a new field `publicAccessPrevention` in the bucket IAM configuration for storage. Full details are available in [this doc](https://docs.google.com/document/d/1o_ElQg4c8enEx8OYa3irrUYFKADmHFbRn7dC9g287sI/edit?usp=sharing). This should include ability to create buckets with this field set, get the field, and set the field. We'd also like an integration test that verifies that the feature works as expected.

Please wait for approval from myself or @shaffeeullah before merging or releasing this change so we can coordinate around release dates.

coryan:
This is a new metadata field in the storage Bucket. This PR also fixes
some typos and blunders in the existing tests for
`BucketIamConfiguration`.

Fixes #5491

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5592)

### DIFF
--- a/google/cloud/storage/bucket_metadata.cc
+++ b/google/cloud/storage/bucket_metadata.cc
@@ -52,10 +52,16 @@ std::ostream& operator<<(std::ostream& os,
 
 std::ostream& operator<<(std::ostream& os, BucketIamConfiguration const& rhs) {
   os << "BucketIamConfiguration={";
+  char const* sep = "";
+  if (rhs.public_access_prevention.has_value()) {
+    os << "public_access_prevention=" << *rhs.public_access_prevention;
+    sep = ", ";
+  }
   if (rhs.uniform_bucket_level_access.has_value()) {
-    os << "uniform_bucket_level_access=" << *rhs.uniform_bucket_level_access;
+    os << sep
+       << "uniform_bucket_level_access=" << *rhs.uniform_bucket_level_access;
     return os << "}";
-  };
+  }
   return os << "}";
 }
 
@@ -309,6 +315,10 @@ BucketMetadataPatchBuilder& BucketMetadataPatchBuilder::SetIamConfiguration(
     BucketIamConfiguration const& v) {
   internal::PatchBuilder iam_configuration;
 
+  if (v.public_access_prevention.has_value()) {
+    iam_configuration.SetStringField("publicAccessPrevention",
+                                     *v.public_access_prevention);
+  }
   if (v.uniform_bucket_level_access.has_value()) {
     internal::PatchBuilder uniform_bucket_level_access;
     uniform_bucket_level_access.SetBoolField(
--- a/google/cloud/storage/bucket_metadata.cc
+++ b/google/cloud/storage/bucket_metadata.cc
@@ -33,7 +33,7 @@ std::ostream& operator<<(std::ostream& os, CorsEntry const& rhs) {
   os << "CorsEntry={";
   char const* sep = "";
   if (rhs.max_age_seconds.has_value()) {
-    os << "max_age_seconds=" << *rhs.max_age_seconds;
+    os << sep << "max_age_seconds=" << *rhs.max_age_seconds;
     sep = ", ";
   }
   return os << sep << "method=[" << absl::StrJoin(rhs.method, ", ")
@@ -54,7 +54,7 @@ std::ostream& operator<<(std::ostream& os, BucketIamConfiguration const& rhs) {
   os << "BucketIamConfiguration={";
   char const* sep = "";
   if (rhs.public_access_prevention.has_value()) {
-    os << "public_access_prevention=" << *rhs.public_access_prevention;
+    os << sep << "public_access_prevention=" << *rhs.public_access_prevention;
     sep = ", ";
   }
   if (rhs.uniform_bucket_level_access.has_value()) {
--- a/google/cloud/storage/bucket_metadata.h
+++ b/google/cloud/storage/bucket_metadata.h
@@ -208,6 +208,10 @@ struct BucketIamConfiguration {
   absl::optional<std::string> public_access_prevention;
 };
 
+inline std::string PublicAccessPreventionEnforced() { return "enforced"; }
+
+inline std::string PublicAccessPreventionUnspecified() { return "unspecified"; }
+
 //@{
 /// @name Comparison operators for BucketIamConfiguration.
 inline bool operator==(BucketIamConfiguration const& lhs,
--- a/google/cloud/storage/bucket_metadata.h
+++ b/google/cloud/storage/bucket_metadata.h
@@ -205,18 +205,23 @@ std::ostream& operator<<(std::ostream& os, UniformBucketLevelAccess const& rhs);
  */
 struct BucketIamConfiguration {
   absl::optional<UniformBucketLevelAccess> uniform_bucket_level_access;
+  absl::optional<std::string> public_access_prevention;
 };
 
 //@{
 /// @name Comparison operators for BucketIamConfiguration.
 inline bool operator==(BucketIamConfiguration const& lhs,
                        BucketIamConfiguration const& rhs) {
-  return lhs.uniform_bucket_level_access == rhs.uniform_bucket_level_access;
+  return lhs.uniform_bucket_level_access == rhs.uniform_bucket_level_access &&
+         lhs.public_access_prevention == rhs.public_access_prevention;
 }
 
 inline bool operator<(BucketIamConfiguration const& lhs,
                       BucketIamConfiguration const& rhs) {
-  return lhs.uniform_bucket_level_access < rhs.uniform_bucket_level_access;
+  return std::tie(lhs.uniform_bucket_level_access,
+                  lhs.public_access_prevention) <
+         std::tie(rhs.uniform_bucket_level_access,
+                  rhs.public_access_prevention);
 }
 
 inline bool operator!=(BucketIamConfiguration const& lhs,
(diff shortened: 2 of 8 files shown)


google/cloud/storage/bucket_metadata.cc
```diff
@@ -51,15 +51,20 @@ std::ostream& operator<<(std::ostream& os, BucketPolicyOnly const& rhs) {
 
 std::ostream& operator<<(std::ostream& os, BucketIamConfiguration const& rhs) {
   os << "BucketIamConfiguration={";
+  char const* sep = "";
+  if (rhs.public_access_prevention.has_value()) {
+    os << "public_access_prevention=" << *rhs.public_access_prevention;
```
devbww, line 56: Opt: Perhaps we should stream `sep` here, to make the position of this element non-special.

google/cloud/storage/bucket_metadata.h
```diff
@@ -212,18 +212,23 @@ struct BucketIamConfiguration {
...
                        BucketIamConfiguration const& rhs) {
-  return lhs.uniform_bucket_level_access == rhs.uniform_bucket_level_access;
+  return lhs.uniform_bucket_level_access == rhs.uniform_bucket_level_access &&
+         lhs.public_access_prevention == rhs.public_access_prevention;
 }
 
 inline bool operator<(BucketIamConfiguration const& lhs,
                       BucketIamConfiguration const& rhs) {
-  return lhs.uniform_bucket_level_access < rhs.uniform_bucket_level_access;
+  return std::tie(lhs.uniform_bucket_level_access,
```
devbww, line 228: Opt: If we use `tie` for `operator<`, perhaps we should also use it for `operator==`.

google/cloud/storage/bucket_metadata_test.cc
```diff
@@ -641,8 +646,22 @@ TEST(BucketMetadataTest, SetIamConfigurationUBLA) {
...
+  auto expected = CreateBucketMetadataForTest();
+  auto copy = expected;
+  BucketIamConfiguration new_configuration;
+  new_configuration.public_access_prevention = "enforced";
+  copy.set_iam_configuration(new_configuration);
+  ASSERT_TRUE(copy.has_iam_configuration());
+  EXPECT_EQ(new_configuration, copy.iam_configuration());
+  EXPECT_NE(expected, copy)
+      << "expected = " << expected.iam_configuration()
+      << "\n  actual=" << copy.iam_configuration() << "\n";
```
devbww, line 660: Perhaps a space before the `=`, like in `"expected ="`?

devbww (approved):

google/cloud/storage/bucket_metadata_test.cc
coryan: Fixed here and elsewhere.

google/cloud/storage/bucket_metadata.h
coryan: Done

google/cloud/storage/bucket_metadata.cc
coryan: Done

google/cloud/storage/bucket_metadata.h
```diff
@@ -212,18 +212,25 @@ struct BucketIamConfiguration {
   ///     `uniform_bucket_level_access`
   absl::optional<BucketPolicyOnly> bucket_policy_only;
   absl::optional<UniformBucketLevelAccess> uniform_bucket_level_access;
+  absl::optional<std::string> public_access_prevention;
```
tritone, line 215: Is it typical to use a plain string for a field like this in C++? For some languages, we've used an Enum or at least supplied constants for the valid values.

google/cloud/storage/examples/storage_bucket_samples.cc
```diff
@@ -435,6 +435,81 @@ void GetUniformBucketLevelAccess(google::cloud::storage::Client client,
...
+  [](gcs::Client client, std::string const& bucket_name) {
+    StatusOr<gcs::BucketMetadata> bucket_metadata =
+        client.GetBucketMetadata(bucket_name);
+    if (!bucket_metadata) {
+      throw std::runtime_error(bucket_metadata.status().message());
+    }
+
+    if (bucket_metadata->has_iam_configuration() &&
+        bucket_metadata->iam_configuration()
+            .public_access_prevention.has_value()) {
```
tritone, line 499: My understanding is that, after this change, the IAM config and the PAP field should always be present for a bucket. So I don't think this check should be necessary (unless I'm missing something about how things work in this library). @shaffeeullah correct me if I'm wrong.

google/cloud/storage/examples/storage_bucket_samples.cc
```diff
@@ -585,6 +660,17 @@ void RunAll(std::vector<std::string> const& argv) {
   std::cout << "\nRunning GetUniformBucketLevelAccess() example" << std::endl;
   GetUniformBucketLevelAccess(client, {bucket_name});
 
+  std::cout << "\nRunning SetPublicAccessPreventionEnforced() example"
+            << std::endl;
+  SetPublicAccessPreventionEnforced(client, {bucket_name});
```
tritone, line 665: So it looks like these tests don't check that the metadata has been updated correctly when re-fetched. Probably fine?

google/cloud/storage/bucket_metadata_test.cc
```diff
@@ -958,27 +977,31 @@ TEST(BucketMetadataPatchBuilder, ResetDefaultAcl) {
...
-  builder.ResetEncryption();
+  builder.ResetIamConfiguration();
 
   auto actual = builder.BuildPatch();
   auto json = nlohmann::json::parse(actual);
-  ASSERT_EQ(1U, json.count("encryption")) << json;
-  ASSERT_TRUE(json["encryption"].is_null()) << json;
+  ASSERT_EQ(1U, json.count("iamConfiguration")) << json;
+  ASSERT_TRUE(json["iamConfiguration"].is_null()) << json;
 }
```
tritone, line 1006: I don't see an integration test for this feature (aside from the sample test), am I missing it?

tritone (commented):
Sorry for the slow review, and thanks for taking care of this @coryan !

google/cloud/storage/internal/grpc_client_bucket_metadata_test.cc
```diff
@@ -359,8 +359,10 @@ TEST(GrpcClientBucketMetadata, BucketIamConfigurationRoundtrip) {
   auto tp = std::chrono::system_clock::time_point{} +
             std::chrono::duration_cast<std::chrono::system_clock::duration>(
                 std::chrono::seconds(1234) + std::chrono::nanoseconds(5678000));
-  auto const expected =
-      BucketIamConfiguration{{}, UniformBucketLevelAccess{true, tp}};
+  auto const expected = BucketIamConfiguration{
+      /*.bucket_policy_only=*/{},
+      /*.uniform_bucket_level_access=*/UniformBucketLevelAccess{true, tp},
+      /*.public_access_prevention=*/{}};
```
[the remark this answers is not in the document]
coryan: you cannot skip fields in C++, and the (deprecated, but not yet removed) BPO field is first. The comments `/*.foo=*/` is to help programmers see what the fields are.

google/cloud/storage/bucket_metadata.h
coryan: > Is it typical to use a plain string for a field like this in C++?

Yes, we do that for all the other fields.

> For some languages, we've used an Enum or at least supplied constants for the valid values.

I can add the constants (strings), you are going to have a bad day with enums when they add a new value.

google/cloud/storage/examples/storage_bucket_samples.cc
coryan: This is an example, not a test.  As long as the example runs I am happy.

google/cloud/storage/bucket_metadata_test.cc
coryan: Not missing it, but frankly we do not have an integration tests for each field. Unit tests + a generic integration test is how we roll.

google/cloud/storage/internal/grpc_client_bucket_metadata_test.cc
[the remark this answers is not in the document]
tritone: Ahh okay, thanks for clarifying.

google/cloud/storage/examples/storage_bucket_samples.cc
tritone: Sounds good.

google/cloud/storage/bucket_metadata_test.cc
tritone: Got it, if that's how it works in this library then it's fine with me.

google/cloud/storage/bucket_metadata.h
coryan: Constants added.

google/cloud/storage/bucket_metadata_test.cc
coryan: Added the integration tests, had to disable it because it does not pass against production with my personal project 🤷

google/cloud/storage/examples/storage_bucket_samples.cc
coryan: This is defensive, yes.  Having said that: even for all the old buckets?  I doubt the millions (at a guess, but maybe billions) of existing buckets will get the new field added to their metadata.  Even if they do, the application can use a `fields` attribute to selectively return just a few of the fields in the metadata, I would like the code not to crash in that case.

coryan:
Rebased to resolve conflicts.

coryan:
Okay, this has been dragging for too long and still has several weeks before it is ready. Closing for now, will reopen when needed.

Issue #5426, opened 2020-11-06
Replace custom time-parsing code with Abseil >= Sept 2020 LTS 


devjgm:
After we can take a dep on the Sept 2020 LTS (https://github.com/abseil/abseil-cpp/releases/tag/20200923.2), we should replace all of our custom time-parsing code with Abseil's time parsing functions. We need to wait on that LTS for some time-parsing fixes in Abseil's code.

This issue is being split out from https://github.com/googleapis/google-cloud-cpp/issues/4458#issuecomment-720472425

devbww:
For the record, I have a pending PR that I believe will cover this.

bduclaux:
Also, StatusOr is now in Abseil. See https://abseil.io/blog/2020-091021-status

coryan:
@bduclaux yes, we are aware, see #4375 (unfortunately the `StatusOr<>` is Abseil is incompatible in a couple of ways)

PR #5603, opened 2020-12-10
refactor: use absl::ParseTime rather than custom impl

devjgm:
Fixes: #530
Fixes: #5426

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5603)

### DIFF
--- a/google/cloud/internal/parse_rfc3339.cc
+++ b/google/cloud/internal/parse_rfc3339.cc
@@ -30,7 +30,8 @@ std::chrono::system_clock::time_point ParseRfc3339(
   // will handle `T` and `Z` in upper or lower case.
   auto const upper = absl::AsciiStrToUpper(timestamp);
   if (!absl::ParseTime(absl::RFC3339_full, upper, &t, &err)) {
-    google::cloud::internal::ThrowInvalidArgument(err);
+    google::cloud::internal::ThrowInvalidArgument(
+        "Error parsing RFC-3339 timestamp: '" + timestamp + "': " + err);
   }
   return absl::ToChronoTime(t);
 }
--- a/google/cloud/internal/parse_rfc3339.cc
+++ b/google/cloud/internal/parse_rfc3339.cc
@@ -13,180 +13,10 @@
 // limitations under the License.
 
 #include "google/cloud/internal/parse_rfc3339.h"
+#include "google/cloud/internal/absl_str_cat_quiet.h"
 #include "google/cloud/internal/throw_delegate.h"
-#include <array>
-#include <cctype>
-#include <cstdio>
-#include <iomanip>
-#include <iostream>
-#include <sstream>
-
-namespace {
-[[noreturn]] void ReportError(std::string const& timestamp, char const* msg) {
-  std::ostringstream os;
-  os << "Error parsing RFC 3339 timestamp: " << msg
-     << " Valid format is YYYY-MM-DD[Tt]HH:MM:SS[.s+](Z|[+-]HH:MM), got="
-     << timestamp;
-  google::cloud::internal::ThrowInvalidArgument(os.str());
-}
-
-bool IsLeapYear(int year) {
-  // NOLINTNEXTLINE(readability-magic-numbers)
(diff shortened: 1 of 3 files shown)


coryan (approved):

devjgm:
Adding @devbww as someone who may have thoughts about this.

But if the only reason we were waiting on the new Abseil release was because of the case sensitivity issues, I think I'd rather take this code change now, even at the cost of the use of to-upper. Thoughts?

google/cloud/internal/parse_rfc3339.cc
```diff
@@ -14,218 +14,25 @@
...
-#include <sstream>
-
-namespace {
-[[noreturn]] void ReportError(std::string const& timestamp, char const* msg) {
-  std::ostringstream os;
-  os << "Error parsing RFC 3339 timestamp: " << msg
-     << " Valid format is YYYY-MM-DD[Tt]HH:MM:SS[.s+](Z|[+-]HH:MM), got="
-     << timestamp;
-  google::cloud::internal::ThrowInvalidArgument(os.str());
-}
```
devbww, line 31: The version of this PR that I've had laying around kept (a slightly tweaked) `ReportError()`-like message because the error string from `absl::ParseTime()` is notoriously terse.

google/cloud/internal/parse_rfc3339.cc
```diff
@@ -14,218 +14,25 @@
...
-  time_point += fractional_seconds;
-  time_point -= offset;
-  time_point -= kLocalTimeOffset;
-  return time_point;
+  std::string err;
+  absl::Time t;
+  // TODO(#5593): Once we take a dep on Abseil >= Sept 2020 LTS, we can remove
+  // this call to `absl::AsciiStrToUpper()`, because then `absl::ParseTime()`
+  // will handle `T` and `Z` in upper or lower case.
+  auto const upper = absl::AsciiStrToUpper(timestamp);
```
devbww, line 31: > if the only reason we were waiting on the new Abseil release was because of the case sensitivity issues, I think I'd rather take this code change now, even at the cost of the use of to-upper. Thoughts?

Eliminating the need for the "to upper" by updating the Abseil dep was all my version was waiting for.  It always seemed to be "any day now", but if you think that day is still far off, so be it.

google/cloud/internal/parse_rfc3339.cc
```diff
@@ -14,218 +14,25 @@
...
-#include "google/cloud/internal/diagnostics_pop.inc"
-}  // anonymous namespace
+#include "absl/strings/ascii.h"
+#include "absl/time/time.h"
 
 namespace google {
 namespace cloud {
 inline namespace GOOGLE_CLOUD_CPP_NS {
 namespace internal {
 std::chrono::system_clock::time_point ParseRfc3339(
```
devbww, line 24: I also updated the header to remove the specific reference to "Google Cloud Storage" with ...
```
 * Google Cloud libraries often use RFC-3339 for timestamps. This function is
 * used to parse them and convert to `std::chrono::system_clock::time_point`,
 * the C++ class used to represent timestamps. Depending on the underlying
 * C++ implementation the timestamp may lose precision. C++ does not specify
 * the precision of the system clock, though most implementations have sub-
 * second precision, and nanoseconds is common. The RFC-3339 spec allows for
 * arbitrary precision in fractional seconds, though it would be surprising
 * to see femtosecond timestamp for Internet events.
```

google/cloud/internal/parse_rfc3339.cc
devjgm: Good point. Fixed.

google/cloud/internal/parse_rfc3339.cc
devjgm: I do think it's worth taking this change now so we get the majority of the refactoring improvement, and the only price is a single extra "to upper". I'd rather wait on eliminating that to upper than wait on the rest of the refactoring.

google/cloud/internal/parse_rfc3339.cc
devjgm: Done.

devbww (approved):

Issue #3046, opened 2019-09-12
Determine namespace(s) to use for "utility" classes.


mr-salty:
Our "utility" classes are currently in the namespace `google::cloud::internal`.  This is fairly unwieldy to type, but attempting to abbreviate it also has issues.

* using-declarations can result in ADL problems (this was actually seen with `make_unique` with a C++17 compiler).
* writing `internal::foo` tends not to work because individual projects have their own more deeply nested `internal` namespaces (e.g. `google::cloud::spanner::internal`).

It would be nice to have a short namespace (similar to `std` or `absl`) so we can name these types unambiguously and without too much cruft.

We should also update/eliminate existing using declarations in this project and others (i.e. spanner).

coryan:
Is there something actionable here?  Are we going to do something?

devjgm:
The pattern we're using now is:

* `google::cloud::foo` (the main product namespace)
* `google::cloud::foo_inernal` (internal bits; a peer of the public namespace)
* `google::cloud::foo_mocks` (optional public namespace for mocks; again a peer of the public namespace)

I don't think it's worth re-namespacing our existing libraries (Spanner, Bigtable, GCS) to match this pattern. But new products (like Pub/Sub), including generated ones, should use this scheme.

At this point, I don't think there's any further work for this issue. Closing.

mr-salty:
we could rename our existing libraries and use aliases so we can migrate slowly.

PR #5609, opened 2020-12-10
chore(spanner): rename spanner::internal -> spanner::spanner_internal

devjgm:
Related to #3046

This naming is also the direction we are moving now with new libraries.

/cc: @scotthart, for consideration w/ the generator

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5609)


devjgm:
Closing because this namespace should have been a peer of `google::cloud::spanner`, rather than a subnamespace. Doing that change is a lot more work and likely not worth the effort. Dropping this.

PR #5612, opened 2020-12-11
feature: add ConnectionOptionsOverride class

scotthart:
This is a stab at creating a version of `google::cloud::ConnectionOptions` that could be used to override defaults when passed to a Connection factory function that is "defaults aware". It is not intended to be submitted as is, but to provoke discussion and feedback.

I decided to make the new type a specialization of `google::cloud::ConnectionOptions<void>` because I thought it might provide some compatibility benefits to things that could accept a `google::cloud::ConnectionOptions<T>` if such things exist. It could just as easily be a completely separate type.

My thought was new libraries would write Connection factory functions that accepted the new type. And, if desired, we could add overloads to existing Connection factory functions that accepted the new type.

### DIFF
--- a/google/cloud/connection_options.cc
+++ b/google/cloud/connection_options.cc
@@ -24,14 +24,14 @@ namespace google {
 namespace cloud {
 inline namespace GOOGLE_CLOUD_CPP_NS {
 namespace internal {
-std::set<std::string> DefaultTracingComponents() {
+absl::optional<std::set<std::string>> DefaultTracingComponents() {
   auto tracing =
       google::cloud::internal::GetEnv("GOOGLE_CLOUD_CPP_ENABLE_TRACING");
   if (!tracing.has_value()) return {};
   return absl::StrSplit(*tracing, ',');
 }
 
-TracingOptions DefaultTracingOptions() {
+absl::optional<TracingOptions> DefaultTracingOptions() {
   auto tracing_options =
       google::cloud::internal::GetEnv("GOOGLE_CLOUD_CPP_TRACING_OPTIONS");
   if (!tracing_options) return {};
--- a/google/cloud/connection_options.h
+++ b/google/cloud/connection_options.h
@@ -20,6 +20,7 @@
 #include "google/cloud/status_or.h"
 #include "google/cloud/tracing_options.h"
 #include "google/cloud/version.h"
+#include "absl/types/optional.h"
 #include <grpcpp/grpcpp.h>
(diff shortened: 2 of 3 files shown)


devjgm (commented):
Thanks, Scott! This is a great start, and it gives us something concrete to focus on and discuss. Here are a few high-level design comments.

* `ConnectionImpl` seems to have two personalities, that of a "bag of optionals", and that of a class with fields with default values. I think we should simplify that class by making it _only_ a "bag of optionals. That is, I think it should have all `optional<T>` data members, setters that accept `optional<T>` (so an option can be _unset_), and accessors that return `optional<T>` (we could ref-qualify the getters to make it more efficient). We do not need `has_foo()` methods because the optional encapsulates that information. This class would have _no default values_ for anything, i.e., the "default" would be a disengaged optional. I recommend following the style of [spanner/query_options.h](https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/spanner/query_options.h)

* Once we have that simpler implementation described above, then the existing `ConnectionOptions<T>` class should hold one of those and implement all of its logic about setting default values, using the Traits classes, etc. That's complexity that would _only_ affect the "old" `ConnectionOptions<T>` class.

* What about naming? I'm not sure what the new one should be named at this point. I'd really like `ConnectionOptions`, but that's taken. Maybe there's a reasonable migration strategy to get there, but I'm not sure yet. We should discuss options. But until then, I think you should implement the _new class in the `internal` namespace_, so that nothing undecided leaks into the public API. You could still have the micro-generator reach into the internal namespace for the time being. I might consider temporarily naming it `internal::ConnectionOptions2` so the name is clearly distinguished from the old one, it's private, and it'll be easy to find and rename later, but whatever.

* The constructor for `internal::ConnectionOptions2` should not set any defaults. All defaults are a disengaged optional. Indeed, `internal::ConnectionOptions2` should likely not have any constructor other than the default constructor: Any setting of fields can be done with the setters, and if we ref-qualify them and `return *this`, callers could use a nice fluent syntax to set the fields they want.

* I don't think we should use a `ConnectionOptions<void>` specialization for this new options type because it has different semantics than the base template -- there are no defaults, the setters/getters have different signatures (take/return optional<T>), and no `has_foo()` methods.

devjgm:
I believe we should close this PR in favor of the plan described in https://github.com/googleapis/google-cloud-cpp/issues/5738. @scotthart feel free to reopen if I'm mistaken.

PR #5620, opened 2020-12-11
cleanup: rename spanner::internal to spanner_internal

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620)

### DIFF
--- a/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
+++ b/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
@@ -48,6 +48,7 @@
 namespace {
 
 namespace spanner = ::google::cloud::spanner;
+namespace spanner_internal = ::google::cloud::spanner_internal;
 using ::google::cloud::Status;
 using ::google::cloud::spanner_benchmarks::Config;
 
@@ -468,18 +469,18 @@ class ExperimentImpl {
   }
 
   std::pair<std::vector<spanner::Client>,
-            std::vector<std::shared_ptr<spanner::internal::SpannerStub>>>
+            std::vector<std::shared_ptr<spanner_internal::SpannerStub>>>
   CreateClientsAndStubs(Config const& config,
                         spanner::Database const& database) {
     std::vector<spanner::Client> clients;
-    std::vector<std::shared_ptr<spanner::internal::SpannerStub>> stubs;
+    std::vector<std::shared_ptr<spanner_internal::SpannerStub>> stubs;
     std::cout << "# Creating clients and stubs " << std::flush;
     for (int i = 0; i != config.maximum_clients; ++i) {
       auto options = spanner::ConnectionOptions().set_channel_pool_domain(
           "task:" + std::to_string(i));
       clients.emplace_back(
           spanner::Client(spanner::MakeConnection(database, options)));
-      stubs.emplace_back(spanner::internal::CreateDefaultSpannerStub(
+      stubs.emplace_back(spanner_internal::CreateDefaultSpannerStub(
           database, options, /*channel_id=*/0));
       std::cout << '.' << std::flush;
     }
@@ -591,7 +592,7 @@ class ReadExperiment : public Experiment {
   Status Run(Config const& config, spanner::Database const& database) override {
     // Create enough clients and stubs for the worst case
     std::vector<spanner::Client> clients;
-    std::vector<std::shared_ptr<spanner::internal::SpannerStub>> stubs;
+    std::vector<std::shared_ptr<spanner_internal::SpannerStub>> stubs;
     std::tie(clients, stubs) = impl_.CreateClientsAndStubs(config, database);
 
     // Capture some overall getrusage() statistics as comments.
@@ -602,7 +603,7 @@ class ReadExperiment : public Experiment {
       auto const thread_count = impl_.ThreadCount(config);
       auto const client_count = impl_.ClientCount(config);
       if (use_stubs) {
-        std::vector<std::shared_ptr<spanner::internal::SpannerStub>>
+        std::vector<std::shared_ptr<spanner_internal::SpannerStub>>
             iteration_stubs(stubs.begin(), stubs.begin() + client_count);
         RunIterationViaStubs(config, iteration_stubs, thread_count);
         continue;
@@ -619,7 +620,7 @@ class ReadExperiment : public Experiment {
  private:
   void RunIterationViaStubs(
       Config const& config,
-      std::vector<std::shared_ptr<spanner::internal::SpannerStub>> const& stubs,
+      std::vector<std::shared_ptr<spanner_internal::SpannerStub>> const& stubs,
       int thread_count) {
     std::vector<std::future<std::vector<RowCpuSample>>> tasks(thread_count);
     int task_id = 0;
@@ -639,7 +640,7 @@ class ReadExperiment : public Experiment {
   std::vector<RowCpuSample> ReadRowsViaStub(
       Config const& config, int thread_count, int client_count,
       spanner::Database const& database,
-      std::shared_ptr<spanner::internal::SpannerStub> const& stub) {
+      std::shared_ptr<spanner_internal::SpannerStub> const& stub) {
     auto session = [&]() -> google::cloud::StatusOr<std::string> {
       Status last_status;
       for (int i = 0; i != 10; ++i) {
@@ -686,7 +687,7 @@ class ReadExperiment : public Experiment {
       for (auto const& name : columns) {
         request.add_columns(name);
       }
-      *request.mutable_key_set() = spanner::internal::ToProto(key);
+      *request.mutable_key_set() = spanner_internal::ToProto(key);
 
       int row_count = 0;
       google::spanner::v1::PartialResultSet result;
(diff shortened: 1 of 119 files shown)


devjgm (approved):
LGTM, but I want to record why I'm weakly opposed to peer-internal namespaces, and ever so slightly prefer `foo::foo_internal`.

* I've never seen this peer-internal namespace strategy before. All "internal" or "detail" namespaces I've seen have been nested within the thing which they are details of. This certainly isn't data supporting my opinion, but it may explain my mild apprehension. :-)

* peer-internal namespaces seem to add extra verbosity with lots of opening and closing namespaces, with sub inline namespaces, in order to forward declare stuff that will be needed.

* friend declarations get a bit messier since they need to include the macro w/ the inline namespace name.

* Having to type `foo::Bar` from within the internal namespace in order to access the main namespace seems unnecessarily verbose. One could argue that this is a feature and makes the code more readable, but then that person probably should also argue that we should qualify even from _within_ the same namespace. So I don't see this as a readability improvement; I see it just as extra typing.

But that's all I've got :-) Nothing major; mostly just slight discomfort... but it's all "internal", so it doesn't really matter, I'm pretty sure I'm in the minority w/ my opinion, so LGTM and thanks for all the work!!!

mr-salty (commented):
I'll defer to what Greg said, the only other thing I was going to comment about is whether we care about blank lines, I saw some inconsistency but I think *extra* blank lines might help readability. i.e. if you only look at the top and bottom of the file we now usually have this:

```
namespace google {
namespace cloud {
namespace spanner {
inline namespace SPANNER_CLIENT_NS {

... [bunch of lines where spanner ends and spanner_internal begins] ...

}  // namespace SPANNER_CLIENT_NS
}  // namespace spanner_internal
}  // namespace cloud
}  // namespace google
```

which looks kind of weird since it's asymmetrical. There is the comment there but it might be easy to gloss over. In some places you have a blank line before `spanner` and/or after `spanner_internal` which I think would make me pause and consider that the namespaces don't all start/end in the same place.

```
namespace google {
namespace cloud {

namespace spanner {
inline namespace SPANNER_CLIENT_NS {

... [bunch of lines where spanner ends and spanner_internal begins] ...

}  // namespace SPANNER_CLIENT_NS
}  // namespace spanner_internal

}  // namespace cloud
}  // namespace google
```

in any case, it's pretty minor and I use `%` in vim a lot... but we should be consistent.

> Reviewed 6 of 119 files at r1, 1 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-:-MOItu_wCVWpLfUivhH9:b4wwjyk)* status: 7 of 119 files reviewed, all discussions resolved (waiting on @coryan)

mr-salty (commented):
I guess I lied, I have one more comment which is I think tests should be in the `spanner` namespace not `spanner_internal`.  Ideally tests should only use the public API of the class so I'd rather have the places where that's not true stand out.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-:-MOIw32e3pKJYTkmiS16:b-4jbvy9)* status: 7 of 119 files reviewed, all discussions resolved (waiting on @coryan)

devjgm:
> I guess I lied, I have one more comment which is I think tests should be in the `spanner` namespace not `spanner_internal`. Ideally tests should only use the public API of the class so I'd rather have the places where that's not true stand out.

Question: Do you mean that tests should *always* be in the public namespace, or that tests should be in the same namespace as the thing they're testing? For example, the tests for `spanner::Value` should be in the `spanner` namespace, but the tests for `spanner_internal::ConnectionImpl` should be in `spanner_internal`?

mr-salty:
yeah, what you said - same namespace as the thing they're testing. thanks for the clarification, i wasn't considering internal tests.

mr-salty (approved):
ah, I see - I didn't realize the tests I was commenting on were in `spanner_internal`, I was trying to go too fast through all those files.  I did find one case of the opposite (test is in `internal`) but doesn't use the namespace, but that was pre-existing and we can fix it up later.  I might have added some using-declarations to avoid all the qualifications but that's personal preference...

I think my comment above about inconsistent blank lines after namespaces still applies in some places but I'd have an easier time identifying them with a script than in this UI... and I don't think it matters that much.

:lgtm: 

> Reviewed 97 of 119 files at r1, 3 of 10 files at r2, 14 of 14 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-:-MOZUd0ZAOuOJiVM1okt:b-6bxmgv)* status: all files reviewed, 3 unresolved discussions (waiting on @coryan)

___
*[google/cloud/spanner/bytes.h, line 30 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-MOZJo5J6bA4mxxIa98l:-MOZJo5J6bA4mxxIa98m:b-hsnvde) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3d11c91aa6ea9a8906f1d966a1378d1f83ee653/google/cloud/spanner/bytes.h#L30)):*
> ```objc
> namespace spanner_internal {
> inline namespace SPANNER_CLIENT_NS {
> struct BytesInternals;
> ```

since this class is completely opaque and contains "Internal" in its name, we might just declare it inside the `spanner` namespace and save 4 more lines here - WDYT?
___
*[google/cloud/spanner/keys.h, line 216 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-MOZNTKt8-Zay7V_pgq4:-MOZNTKt8-Zay7V_pgq5:bz727mg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3d11c91aa6ea9a8906f1d966a1378d1f83ee653/google/cloud/spanner/keys.h#L216)):*
> ```objc
> namespace spanner_internal {
> inline namespace SPANNER_CLIENT_NS {
> struct KeySetInternals {
> ```

`KeySetInternals` feels less compelling when it's inlined in the header but my other suggestions were also convoluted. 
___
*[google/cloud/spanner/internal/tuple_utils_test.cc, line 22 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-MOZUEO_-ia-bskzC85B:-MOZUEO_-ia-bskzC85C:b-maksn0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3d11c91aa6ea9a8906f1d966a1378d1f83ee653/google/cloud/spanner/internal/tuple_utils_test.cc#L22)):*
> ```cpp
> namespace google {
> namespace cloud {
> namespace spanner {
> ```

this should be in namespace `spanner_internal`, then we can remove the qualifiers below.

coryan (commented):
Rebase, address comments, and fix formatting.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-:-MQIWHXjDgxtmL7i749y:b56tybo)* status: 117 of 119 files reviewed, 3 unresolved discussions (waiting on @mr-salty)

___
*[google/cloud/spanner/bytes.h, line 30 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-MOZJo5J6bA4mxxIa98l:-MQIWK_j3CoaHKKxb05W:b-tw711s) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3d11c91aa6ea9a8906f1d966a1378d1f83ee653/google/cloud/spanner/bytes.h#L30)):*

since this class is completely opaque and contains "Internal" in its name, we might just declare it inside the `spanner` namespace and save 4 more lines here - WDYT?
</blockquote></details>

I do not disagree, but I think somebody else could make a pass fixing these `*Internal` helpers.  BTW, we could.
___
*[google/cloud/spanner/keys.h, line 216 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-MOZNTKt8-Zay7V_pgq4:-MQIW_vd2WrDshNdIfD6:bb6sh2i) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3d11c91aa6ea9a8906f1d966a1378d1f83ee653/google/cloud/spanner/keys.h#L216)):*

`KeySetInternals` feels less compelling when it's inlined in the header but my other suggestions were also convoluted. 
</blockquote></details>

Ack.
___
*[google/cloud/spanner/internal/tuple_utils_test.cc, line 22 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-MOZUEO_-ia-bskzC85B:-MQIWbqmDj89uKjBQgU9:bt2vr1b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3d11c91aa6ea9a8906f1d966a1378d1f83ee653/google/cloud/spanner/internal/tuple_utils_test.cc#L22)):*

this should be in namespace `spanner_internal`, then we can remove the qualifiers below.
</blockquote></details>

Fixed.

coryan:
___
*[google/cloud/spanner/bytes.h, line 30 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5620#-MOZJo5J6bA4mxxIa98l:-MQIuiig5hJ8r-88Liyq:b-3du7c4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3d11c91aa6ea9a8906f1d966a1378d1f83ee653/google/cloud/spanner/bytes.h#L30)):*

I do not disagree, but I think somebody else could make a pass fixing these `*Internal` helpers.  BTW, we could.
</blockquote></details>

Argh.  What I mean to say is "We cloud save more than 4 lines."  If `FooInternal` is not part of the API, then we can just forward declare them in the body of the class.

Issue #5601, opened 2020-12-09
generator: determine how to generate ConnectionOptions in a multi-service product


scotthart:
Given that the current `google::cloud::ConnectionOptions` supports only a single endpoint, determine how `google::cloud::<product>::ConnectionOptions` should be defined/generated in a product with multiple services where not all the services have the same endpoint.

devjgm:
We might want to see if we can factor the default endpoint out of ConnectionOptions. We might be able to remove that requirement from `ConnectionTraits` ([link](https://github.com/googleapis/google-cloud-cpp/blob/00671b0972e27d66301a56801e3f7859e56d3439/google/cloud/connection_options.h#L77)), and have the default end-point be set within the `MakeConnection()` method itself. The `ConnectionOptions` class itself could retain (for API compat) `endpoint()` and `set_endpoint()` methods, but these would only be for users who wish to _override_ the endpoint. The default wouldn't be set there.

Just a thought.

scotthart:
Would we also need to factor out the other defaults provided in ConnectionOptions? number of transport channels and user agent string prefix?

Issue #4528, opened 2020-07-14
Update spanner retries/timeouts to use settings from json files


devjgm:
The default timeout and retry settings for our spanner clients should come from the following official json files:

* Client: https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/spanner_grpc_service_config.json
* DatabaseAdminClient: https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/database/v1/spanner_admin_database_grpc_service_config.json
* InstanceAdminClient: https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/instance/v1/spanner_admin_instance_grpc_service_config.json

We should audit where our current default settings are different, and then update them to match what's recommended in the json files.

coryan:
FWIW, I think for Bigtable we wrote some program to read similar files to these and generate header files that we then just compile.

Also, our retry policies are per-client, not per method.  It would be a fairly large change to modify the policies to know what function is being called and what values they should use in that case.

hengfengli:
Hi @devjgm, is there any update with this issue?

mr-salty:
I don't think anyone has looked at this but I can take care of it

mr-salty:
@scotthart Greg tells me the generator will be able to parse these files and generate code?  So, I'm going to plan to do the easy thing for now and just hardcode the values from the json files, I assume they're not expected to change super often anyways.

devjgm:
Is this done to the extent that we can do anything?

mr-salty:
I think we can/should still do something with this (I'm going to have to hack something for PDML regardless).  I wanted to have a group discussion about it but haven't gotten it together yet.

devjgm:
I'm a bit confused about what remains on this issue. @mr-salty can you please clearly summarize what needs to be done in order to close this issue?

mr-salty:
The work item for this bug (as stated in the original report) is "update [timeout and retry settings] to match what's recommended in the json files".

However, that is blocked on deciding how we want to approach retries and timeouts in general (i.e. what we discussed during our meeting today). I'm not sure if that merits a separate github issue or not. Really, this issue likely applies to all of our client libraries.

coryan:
We will revisit, it may need changes in the generator and some infrastructure to have per-function defaults.

devbww:
Blocked by #8473.

coryan:
Closing. Applications can now override these on a single function if needed.

PR #5623, opened 2020-12-12
refactor(generator): product wide retry and connection option types to service specific

scotthart:
In order to support generation of multiple services in a "product", RetryPolicy and ConnectionOption types need to be specific to each particular service in order to unblock progress on the microgenerator.

This implementation for ConnectionOptions is a temporary measure pending resolution of #5601. RetryPolicy will also need the flexibility for service specific policies, whether generated from the service_config.json files authored for each service, or some other solution per resolution of #4528.

It is likely both the code generated for ConnectionOptions and RetryPolicy will be revisited before using the generator to produce code for a service to be designated as GA.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5623)


coryan (approved):
> Reviewed 20 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5623)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5632, opened 2020-12-15
feat(storage): propagate custom header with resumable upload PUT requests

### LINKED ISSUE #5421: Propagate custom header with resumable upload PUT requests
A customer could use custom headers with the PUT requests in a resumable upload.

sumeetoc:
The Pull Request consists of below addition:

1. Propagate Custom Header in subsequent PUT requests when using Resumable uploads
2. Test case using GCS emulator

Fixes #5421

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632)

### DIFF
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -679,7 +679,7 @@ def resumable_upload_chunk(bucket_name):
     if content_length != len(data):
         utils.error.invalid("content-length header", None)
     content_range = request.headers.get("content-range")
-    custom_header_value = request.headers.get("custom_header_key")
+    custom_header_value = request.headers.get("x-goog-emulator-custom-header")
     if content_range is not None:
         items = list(utils.common.content_range_split.match(content_range).groups())
         if len(items) != 2 or (items[0] == items[1] and items[0] != "*"):
@@ -749,7 +749,7 @@ def resumable_upload_chunk(bucket_name):
             upload.transfer
         )
         blob.metadata.metadata["x_emulator_upload"] = "resumable"
-        blob.metadata.metadata["x-goog-emulator-custom-header"] = str(custom_header_value)
+        blob.metadata.metadata["x_emulator_custom_header"] = str(custom_header_value)
         db.insert_object(upload.request, bucket_name, blob, None)
         projection = utils.common.extract_projection(
             upload.request, CommonEnums.Projection.NO_ACL, None
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -749,7 +749,7 @@ def resumable_upload_chunk(bucket_name):
             upload.transfer
         )
         blob.metadata.metadata["x_emulator_upload"] = "resumable"
-        blob.metadata.metadata["x_emulator_debug_custom_header"] = str(custom_header_value)
+        blob.metadata.metadata["x-goog-emulator-custom-header"] = str(custom_header_value)
         db.insert_object(upload.request, bucket_name, blob, None)
         projection = utils.common.extract_projection(
             upload.request, CommonEnums.Projection.NO_ACL, None
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -749,7 +749,7 @@ def resumable_upload_chunk(bucket_name):
             upload.transfer
         )
         blob.metadata.metadata["x_emulator_upload"] = "resumable"
-        blob.metadata.metadata["x_emulator_debug_custom_header"] = custom_header_value
+        blob.metadata.metadata["x_emulator_debug_custom_header"] = str(custom_header_value)
         db.insert_object(upload.request, bucket_name, blob, None)
         projection = utils.common.extract_projection(
             upload.request, CommonEnums.Projection.NO_ACL, None
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -679,6 +679,7 @@ def resumable_upload_chunk(bucket_name):
     if content_length != len(data):
         utils.error.invalid("content-length header", None)
     content_range = request.headers.get("content-range")
+    custom_header_value = request.headers.get("custom_header_key")
     if content_range is not None:
         items = list(utils.common.content_range_split.match(content_range).groups())
         if len(items) != 2 or (items[0] == items[1] and items[0] != "*"):
@@ -748,6 +749,7 @@ def resumable_upload_chunk(bucket_name):
             upload.transfer
         )
         blob.metadata.metadata["x_emulator_upload"] = "resumable"
+        blob.metadata.metadata["x_emulator_debug_custom_header"] = custom_header_value
         db.insert_object(upload.request, bucket_name, blob, None)
         projection = utils.common.extract_projection(
             upload.request, CommonEnums.Projection.NO_ACL, None
--- a/google/cloud/storage/tests/object_file_integration_test.cc
+++ b/google/cloud/storage/tests/object_file_integration_test.cc
@@ -628,43 +628,44 @@ TEST_F(ObjectFileIntegrationTest, UploadPortionRegularFile) {
 TEST_F(ObjectFileIntegrationTest, ResumableUploadFileCustomHeader) {
   // Test relies on emulator for capturing custom header
   if (!UsingEmulator()) GTEST_SKIP();
- 
+
   StatusOr<Client> client = MakeIntegrationTestClient();
   ASSERT_STATUS_OK(client);
- 
+
   auto file_name = ::testing::TempDir() + MakeRandomFilename();
   auto object_name = MakeRandomObjectName();
- 
+
   // We will construct the expected response while streaming the data up.
   std::ostringstream expected;
   // Create a file with the contents to upload.
   std::ofstream os(file_name, std::ios::binary);
   WriteRandomLines(os, expected);
   os.close();
- 
+
   // Create a CustomHeader object to send in request.
-  CustomHeader custom_header("custom_header_key","custom_header_value");
- 
-  StatusOr<ObjectMetadata> meta = client->UploadFile(
-      file_name, bucket_name_, object_name, custom_header,
-      IfGenerationMatch(0), NewResumableUploadSession());
- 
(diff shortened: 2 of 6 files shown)


google/cloud/storage/emulator/emulator.py
```diff
@@ -679,6 +679,7 @@ def resumable_upload_chunk(bucket_name):
     if content_length != len(data):
         utils.error.invalid("content-length header", None)
     content_range = request.headers.get("content-range")
+    custom_header_value = request.headers.get("custom_header_key")
```
vnvo2409, line 682: What if the user uses a different header key ?

vnvo2409 (commented):
I think you are probably taking things too literally. A custom header is a header which isn't supported by the library and has to be injected using `CustomHeader` ( On the other side, offical header has a separate option for itself ).

What we want is: when a user injects a `CustomHeader` while doing a resumable upload, we want that `CustomHeader` propagates to the last upload PUT requests ( because resumable upload requires a series of PUT requests ).

coryan (changes requested):
some suggestions below on naming the test headers. looking good, let's see what the builds say +kokoro:run 

@vnvo2409 we discussed the feature with @dobbster offline, the use-case they have in mind requires sending the header in each request.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632#-:-MOagHga4ZETpxhR8GTN:bq8rpda)* status: all files reviewed, 2 unresolved discussions (waiting on @sumeetoc and @vnvo2409)

___
*[google/cloud/storage/emulator/emulator.py, line 682 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632#-MO_J9joc_pn4T1BHeTw-r1-682:-MOafiOqBbyE7sUowYrz:bxjpco2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f571d447fb1d8e00f60234ac05b2b10417da7e65/google/cloud/storage/emulator/emulator.py#L682)):*
<details><summary><i>Previously, vnvo2409 (Vo Van Nghia) wrote…</i></summary><blockquote>

What if the user uses a different header key ?
</blockquote></details>

I think what we need is a way to capture a well known header for testing, not necessarily a way to capture any possible header.

I would suggest that we capture `x-goog-emulator-custom-header` for consistency with the metadata field.
___
*[google/cloud/storage/tests/object_file_integration_test.cc, line 646 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632#-MOafO3L9oiL_2GmgbSb:-MOafO3L9oiL_2GmgbSc:b-xdmstc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f571d447fb1d8e00f60234ac05b2b10417da7e65/google/cloud/storage/tests/object_file_integration_test.cc#L646)):*
> ```cpp
>  
>   // Create a CustomHeader object to send in request.
>   CustomHeader custom_header("custom_header_key","custom_header_value");
> ```

I think header names are traditionally separated by `-`, so this would be `custom-header-key`. I would use `x-goog-emulator-custom-header` to make it painfully obvious that this is a test.

google/cloud/storage/tests/object_file_integration_test.cc
```diff
@@ -625,6 +625,51 @@ TEST_F(ObjectFileIntegrationTest, UploadPortionRegularFile) {
...
+      IfGenerationMatch(0), NewResumableUploadSession());
+ 
+  ASSERT_STATUS_OK(meta);
+  EXPECT_EQ(object_name, meta->name());
+  EXPECT_EQ(bucket_name_, meta->bucket());
+  auto expected_str = expected.str();
+  ASSERT_EQ(expected_str.size(), meta->size());
+ 
+  ASSERT_TRUE(meta->has_metadata("x-goog-emulator-custom-header"));
+  EXPECT_EQ("custom_header_value", meta->metadata("x-goog-emulator-custom-header"));
```
vnvo2409, line 659: Ahh sorry, I misunderstood the code :sweat_smile:. One more comment, we have to wrap this code around
```
if (UsingEmulator())
```

google/cloud/storage/tests/object_file_integration_test.cc
```diff
@@ -625,6 +625,51 @@ TEST_F(ObjectFileIntegrationTest, UploadPortionRegularFile) {
   EXPECT_EQ(0, std::remove(file_name.c_str()));
 }
 
+TEST_F(ObjectFileIntegrationTest, ResumableUploadFileCustomHeader) {
+  // Test relies on emulator for capturing custom header
+  if (!UsingEmulator()) GTEST_SKIP();
```
coryan, line 630: @vnvo2409 FYI, the test is skipped if not using the emulator.

vnvo2409: Ah sorry again 😓 I am not at my best today.

coryan (changes requested):
There are formatting changes needed, see the `clang-tidy` build results.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632#-:-MOaoupW9XGBE67lgWrT:b95oyjk)* status: all files reviewed, 4 unresolved discussions (waiting on @coryan, @sumeetoc, and @vnvo2409)

___
*[google/cloud/storage/tests/object_file_integration_test.cc, line 659 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632#-MO_J9joc_pn4T1BHeU--r1-659:-MOapWww3BpUqHrwRLdw:b-dlvr75) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f571d447fb1d8e00f60234ac05b2b10417da7e65/google/cloud/storage/tests/object_file_integration_test.cc#L659)):*
<details><summary><i>Previously, vnvo2409 (Vo Van Nghia) wrote…</i></summary><blockquote>

Ahh sorry, I misunderstood the code :sweat_smile:. One more comment, we have to wrap this code around
```
if (UsingEmulator())
```
</blockquote></details>

Answered elsewhere, the whole test is disabled outside the emulator.

dobbster:
> I think header names are traditionally separated by -, so this would be custom-header-key. I would use x-goog-emulator-custom-header to make it painfully obvious that this is a test.
I had assumed it was the metadata field that needed renaming, but on second thoughts, it made more sense to have a consistent key name for our emulator. My bad, thanks for pointing that out!

The metadata field itself is named `x_emulator_custom_header` as the underscored naming convention seemed apt for the emulator, let me know if this needs changes as well.

> @vnvo2409 FYI, the test is skipped if not using the emulator.
This is right, the test case is skipped entirely if the emulator is not running. This seemed to be the best way.

> There are formatting changes needed, see the clang-tidy build results.
I ran the new changes with clang-tidy as well so hopefully this took care of it!

coryan (changes requested):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632)* status: all files reviewed, 4 unresolved discussions (waiting on @sumeetoc and @vnvo2409)

___
*[google/cloud/storage/internal/curl_resumable_upload_session.h, line 37 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632#-MOb55ZwEzr_Wjz6ENhn:-MOb55ZwEzr_Wjz6ENho:b-p5v8w7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2da68dddfb3872df87ced54b5165d21fa945e91f/google/cloud/storage/internal/curl_resumable_upload_session.h#L37)):*
> ```objc
>       : client_(std::move(client)),
>         session_id_(std::move(session_id)),
>         custom_header_(custom_header) {}
> ```

you probably noticed this already, but `clang-tidy` is reminding us that this should be:

```cc
    custom_header_(std::move(custom_header))
```
___
*[google/cloud/storage/tests/object_file_integration_test.cc, line 630 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632#-MO_J9joc_pn4T1BHeU--r1-630:-MOag-rbCwT_udekwTOB:b-vqkjuf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f571d447fb1d8e00f60234ac05b2b10417da7e65/google/cloud/storage/tests/object_file_integration_test.cc#L630)):*
<details><summary><i>Previously, vnvo2409 (Vo Van Nghia) wrote…</i></summary><blockquote>

Ah sorry again 😓 I am not at my best today.
</blockquote></details>

No worries.

dobbster:
> you probably noticed this already, but `clang-tidy` is reminding us that this should be:
> 
> ```c++
>     custom_header_(std::move(custom_header))
> ```
Ah yes, I almost overlooked this

Awaiting for any further modifications needed from the build results

coryan (approved):
:lgtm: +kokoro:run 

> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5632#-:-MObRBHxB1y17rCJyp95:bcbhv2)* status: all files reviewed, 2 unresolved discussions (waiting on @vnvo2409)

google/cloud/storage/tests/object_file_integration_test.cc
remyabel: Sorry to butt in, but it seems strange to have marked this comment as off-topic. "Mark as obsolete" would've made more sense here.

coryan: Hmm... yeah, you are right.  OTOH, I do not recall explicitly marking it as "off-topic" and/or "obsolete" or anything else.  Is this a new feature that I am misusing or something?

vnvo2409: No I think it's not a new feature but it's because we rarely use it. I think we usually mark a review as resolve. Hiding a review is when I feel that there is a serious problem with it. In addition, Github does not have "obsolete" option indeed.

image

remyabel: Yeah "obsolete" was a freudian slip. "Resolved" is what I mean. But "off-topic" makes it sound like the comment is disruptive (when it's clearly not).

vnvo2409: > But "off-topic" makes it sound like the comment is disruptive (when it's clearly not).

Ah I did not think about it. But there is a "Spam" option for that kind of comments though. And "Resolved" is not really this case so I just go with "off-topic".

PR #5637, opened 2020-12-15
docs: homogenize region tags

Strykrol:
These samples' region tags slightly differ from identical snippets in different languages. The following is the remedy:

1. Merge this PR to fix GitHub source (keep old tag for now)
2. Fix doc code includes to represent the better region tag
3. Submit a second PR to clean-up redundant region tag wrapping

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5637)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -2456,6 +2456,7 @@ void FieldAccessOnStructParameters(google::cloud::spanner::Client client) {
 }
 //! [END spanner_field_access_on_struct_parameters]
 
+//! [START spanner_field_access_on_nested_struct_parameters]
 //! [START spanner_field_access_on_nested_struct]
 void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
   namespace spanner = ::google::cloud::spanner;
@@ -2491,6 +2492,7 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
   std::cout << "Query completed for [spanner_field_access_on_nested_struct]\n";
 }
 //! [END spanner_field_access_on_nested_struct]
+//! [END spanner_field_access_on_nested_struct_parameters]
 
 void ExampleStatusOr(google::cloud::spanner::Client client) {
   //! [example-status-or]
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -681,6 +681,7 @@ void CreateBackupCommand(std::vector<std::string> const& argv) {
   CreateBackup(std::move(client), argv[0], argv[1], argv[2], argv[3]);
 }
 
+//! [restore-database] [START spanner_restore_backup]
```
coryan, line 684: We need to remove the `[restore-database]` doxygen tag from this line or the next line.  If we eventually want to remove the `spanner_restore_database` region tag I would recommend moving the doxygen tag.  If we want to keep both region tags, then maybe removing `[restore-database]` from the new lines is better.  More on this can be found at go/cloud-cxx:snippets-dd

google/cloud/spanner/samples/samples.cc
```diff
@@ -700,6 +701,7 @@ void RestoreDatabase(google::cloud::spanner::DatabaseAdminClient client,
             << " backup '" << backup.FullName() << "'.\n";
 }
 //! [restore-database] [END spanner_restore_database]
+//! [restore-database] [END spanner_restore_backup]
```
coryan, line 704: See above.

coryan (changes requested):
Thanks for the PR!

google/cloud/spanner/samples/samples.cc
Strykrol: This is one of multiple PRs spanning several repositories; I can't be certain how quickly those will get turned around (I'd rather not move the tag for now). 

I'm can remove the tag from the new line(s), and when I do a 2nd PR pass to remove the "old" `spanner_restore_database` I will move the doxygen tag at that time. SGTY?

coryan: > I'm can remove the tag from the new line(s), and when I do a 2nd PR pass to remove the "old" spanner_restore_database I will move the doxygen tag at that time. SGTY?

Yup.

coryan (approved):
LGTM, thanks!

PR #5639, opened 2020-12-16
feature(generator): accept commit hash from command line arg

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5639)


coryan (approved):
Is it possible to add some help message or other documentation about how this is used (or will be used)?  Can you also add some details about the motivation to the PR so we know /why/ we made this change in the future?

scotthart:
Currently, the generator uses the googleapis commit hash to generate html links for doxygen markup. This PR adds a mechanism to specify that hash per invocation of the generator.

PR #5659, opened 2020-12-23
fix(generator): handle package local and non-imported deduced lro response types

### LINKED ISSUE #5656: generator: deducing lro return types not in descriptor pool
Currently, when specifying a metadata or response type in an `option (google.longrunning.operation_info)`, the named type does not have to be imported and thus will not be present in the descriptor pool for the generator to interrogate. Provide a workaround when encountering these scenarios.

### LINKED ISSUE #5657: generator: prepend package to locally defined lro operation_info types
When specifying `option (google.longrunning.operation_info)` the respone and/or metadata types only need to be fully qualified if found in a different package. When the generator fails to find an unqualified type, try again prepending the package to the type before failing.

scotthart:
When specifying the metadata_type or response_type in an `option (google.longrunning.operation_info)` annotation, the types only need to be fully qualified if they exist in a different package than the rpc. Thus, for unqualified types the generator will concatenate the rpc package with the type specified, if necessary.
Furthermore, it is not required that those types source .proto file is imported into the current file, so the generator cannot verify that the type exists at all if the source .proto is not tangentially imported somewhere in the dependency tree. The generator will make a best effort to find a descriptor for the type specified. If no descriptor can be found the generator will just handle it as a string literal.

fixes #5657 
fixes #5656

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5659)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -84,6 +84,22 @@ std::string FormatDoxygenLink(google::protobuf::Descriptor const& message_type,
 }
 
 absl::variant<std::string, google::protobuf::Descriptor const*>
+FullyQualifyMessageType(google::protobuf::MethodDescriptor const& method,
+                        std::string message_type) {
+  google::protobuf::Descriptor const* output_type =
+      method.file()->pool()->FindMessageTypeByName(message_type);
+  if (output_type != nullptr) {
+    return output_type;
+  }
+  output_type = method.file()->pool()->FindMessageTypeByName(
+      absl::StrCat(method.file()->package(), ".", message_type));
+  if (output_type != nullptr) {
+    return output_type;
+  }
+  return message_type;
+}
+
+absl::variant<std::string, google::protobuf::Descriptor const*>
 DeduceLongrunningOperationResponseType(
     google::protobuf::MethodDescriptor const& method,
     google::longrunning::OperationInfo const& operation_info) {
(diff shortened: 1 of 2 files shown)


coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5659#-:-MPCL21J9LWdVXiDKW-F:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @scotthart)

___
*[generator/internal/descriptor_utils.cc, line 94 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5659#-MPCKhAa1LcuIKjqbHoS:-MPCKhAa1LcuIKjqbHoT:bpwf254) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4d79ee35de0acaacf988a25e33fc84e2e4ce7497/generator/internal/descriptor_utils.cc#L94)):*
> ```cpp
>           ? operation_info.metadata_type()
>           : operation_info.response_type();
>   google::protobuf::Descriptor const* output_type =
> ```

nit: consider `auto const* output_type = ...`
___
*[generator/internal/descriptor_utils.cc, line 102 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5659#-MPCKcIfCq6BN3HOJPDH:-MPCKcIfCq6BN3HOJPDI:b89p562) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4d79ee35de0acaacf988a25e33fc84e2e4ce7497/generator/internal/descriptor_utils.cc#L102)):*
> ```cpp
>       absl::StrCat(method.file()->package(), deduced_response_type));
>   if (output_type != nullptr) {
>     return output_type;
> ```

nit: this branch is not reached, which presumably you wanted to test?

Issue #2567, opened 2019-04-27
Remove CompletionQueue parameter from Async*() functions.


coryan:
Instead the `*Client` should take the `CompletionQueue` as a parameter in its constructor, probably wrapped in a "Concurrency Policy" or something similar.

If no `CompletionQueue` is provided the functions should create a completion queue on demand, and block until the result is ready, effectively the default is "NoConcurrency" or "Synchronous".

dopiera:
FWIW, here is a POC showing how this can be done:
https://github.com/dopiera/google-cloud-cpp/commit/51dbf0f5835795203333dcc2a912a71d89f902ad

dopiera:
I've spent some significant amount of time trying different approaches to implement it.

My goals were:
* we need something that is going to make it nearly impossible for us to:
  * forget to wait for the operation in the "Synchronous" case
  * forget to call `Run` on the `CompletionQueue` in the "Synchronous" case
  * use a wrong `CompletionQueue` (a temporary instead of a provided one)
* will not require us to effectively duplicate every asynchronous test

I think I achieved all of the goals in https://github.com/dopiera/google-cloud-cpp/commit/396ec5d220117f42d7d36da09ea0adb08b2c4c17.

Unfortunately there is IoC, but it's synchronous code, so the current codebase can almost be converted by a script.

The proposal is to introduce a `CompletionQueueProvider` to every client. It would optionally hold a `CompletionQueue` executed by an external thread(pool).
`CompletionQueueProvider` has one member function: `WithCompletionQueue(Functor)`, which synchronously executes the functor with either the user-provided `CompletionQueue` or with a temporary one. The functor should start the asynchronous operation and return a future (just like all `Async*` operations are implemented). In the "Synchronous" case it also adds a `then` to the returned future, which will shut down the temporary `CompletionQueue` after the operation is complete.

dopiera:
@coryan Carlos, do you think that this approach is the way to go? If yes, I will.

coryan:
I really like the `WithCompletionQueue()` function. The names do not speak to me, so I am going to propose slightly different ones:

- Rename `CompletionQueueProvider` to `CompletionQueueExecutor`.
- Rename `WithCompletionQueue` to `Post`.

I am basically borrowing the names from the `executors` proposal (http://chriskohlhoff.github.io/executors/).

I would like (if possible) to hide this behind some helper functions, the `ExecutionPolicy` should not be a public API:

```C++
[...]

PR #5667, opened 2020-12-30
feat: Cancel all connection refresh timers in dtor

dopiera:
This is a part of #2567.

Timers created by `CommonClient` to refresh connections are now
cancelled in the `CommonClient` dtor. This is needed to make sure that
no stray timers remain in the completion queue after the destruction of
`CommonClient`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)

### DIFF
--- a/google/cloud/bigtable/CMakeLists.txt
+++ b/google/cloud/bigtable/CMakeLists.txt
@@ -257,6 +257,7 @@ if (BUILD_TESTING)
         internal/async_retry_multi_page_test.cc
         internal/async_retry_unary_rpc_and_poll_test.cc
         internal/bulk_mutator_test.cc
+        internal/common_client_test.cc
         internal/google_bytes_traits_test.cc
         internal/logging_admin_client_test.cc
         internal/logging_data_client_test.cc
--- a/google/cloud/bigtable/bigtable_client_unit_tests.bzl
+++ b/google/cloud/bigtable/bigtable_client_unit_tests.bzl
@@ -46,6 +46,7 @@ bigtable_client_unit_tests = [
     "internal/async_retry_multi_page_test.cc",
     "internal/async_retry_unary_rpc_and_poll_test.cc",
     "internal/bulk_mutator_test.cc",
+    "internal/common_client_test.cc",
     "internal/google_bytes_traits_test.cc",
     "internal/logging_admin_client_test.cc",
     "internal/logging_data_client_test.cc",
--- a/google/cloud/bigtable/internal/common_client.cc
+++ b/google/cloud/bigtable/internal/common_client.cc
@@ -20,9 +20,11 @@ inline namespace BIGTABLE_CLIENT_NS {
 namespace internal {
 
 ConnectionRefreshState::ConnectionRefreshState(
+    std::shared_ptr<CompletionQueue> const& cq,
     std::chrono::milliseconds max_conn_refresh_period)
     : max_conn_refresh_period_(max_conn_refresh_period),
-      rng_(std::random_device{}()) {}
+      rng_(std::random_device{}()),
+      timers_(std::make_shared<OutstandingTimers>(cq)) {}
 
 std::chrono::milliseconds ConnectionRefreshState::RandomizedRefreshDelay() {
   std::lock_guard<std::mutex> lk(mu_);
@@ -37,38 +39,87 @@ void ScheduleChannelRefresh(
     std::shared_ptr<grpc::Channel> const& channel) {
   // The timers will only hold weak pointers to the channel or to the
   // completion queue, so if either of them are destroyed, the timer chain
-  // will simply not continue. Unfortunately, that means that some stray
-  // timers may remain in the `CompletionQueue` for a while, but this is
-  // generally unavoidable because there is no way to cancel individual
-  // timers.
+  // will simply not continue.
   std::weak_ptr<grpc::Channel> weak_channel(channel);
   std::weak_ptr<CompletionQueue> weak_cq(cq);
   using TimerFuture = future<StatusOr<std::chrono::system_clock::time_point>>;
-  cq->MakeRelativeTimer(state->RandomizedRefreshDelay())
-      .then([weak_channel, weak_cq, state](TimerFuture fut) {
-        if (!fut.get()) {
-          // Timer cancelled.
-          return;
-        }
-        auto channel = weak_channel.lock();
-        if (!channel) return;
-        auto cq = weak_cq.lock();
-        if (!cq) return;
-        cq->AsyncWaitConnectionReady(channel, std::chrono::system_clock::now() +
-                                                  kConnectionReadyTimeout)
-            .then([weak_channel, weak_cq, state](future<Status> fut) {
-              auto conn_status = fut.get();
-              if (!conn_status.ok()) {
-                GCP_LOG(WARNING)
-                    << "Failed to refresh connection. Error: " << conn_status;
-              }
-              auto channel = weak_channel.lock();
-              if (!channel) return;
-              auto cq = weak_cq.lock();
-              if (!cq) return;
-              ScheduleChannelRefresh(cq, state, channel);
-            });
-      });
+  auto timer_future =
+      cq->MakeRelativeTimer(state->RandomizedRefreshDelay())
+          .then([weak_channel, weak_cq, state](TimerFuture fut) {
+            if (!fut.get()) {
+              // Timer cancelled.
+              return;
+            }
+            auto channel = weak_channel.lock();
+            if (!channel) return;
+            auto cq = weak_cq.lock();
+            if (!cq) return;
+            cq->AsyncWaitConnectionReady(
+                  channel,
+                  std::chrono::system_clock::now() + kConnectionReadyTimeout)
+                .then([weak_channel, weak_cq, state](future<Status> fut) {
+                  auto conn_status = fut.get();
+                  if (!conn_status.ok()) {
+                    GCP_LOG(WARNING) << "Failed to refresh connection. Error: "
+                                     << conn_status;
+                  }
+                  auto channel = weak_channel.lock();
+                  if (!channel) return;
+                  auto cq = weak_cq.lock();
+                  if (!cq) return;
+                  ScheduleChannelRefresh(cq, state, channel);
+                });
+          });
+  state->timers().RegisterTimer(std::move(timer_future));
+}
+
+void OutstandingTimers::RegisterTimer(future<void> fut) {
+  std::unique_lock<std::mutex> lk(mu_);
+  if (shutdown_) {
+    lk.unlock();
+    fut.cancel();
+    return;
+  }
+
+  auto id = id_generator_++;
+  auto self = shared_from_this();
+  auto timer = fut.then([self, id](future<void>) {
+    // If the completion queue is being destroyed, we can afford not
+    // ignoring this continuation. Most likely nobody cares anymore.
+    auto cq = self->weak_cq_.lock();
+    if (!cq) return;
+    // Do not run in-line to avoid deadlocks when the timer is immediately
+    // satisfied.
+    cq->RunAsync([self, id] { self->DeregisterTimer(id); });
+  });
+  bool const inserted =
+      timers_.emplace(std::make_pair(id, std::move(timer))).second;
+  if (!inserted) Terminate("Duplicate timer identifier");
+}
+
+void OutstandingTimers::DeregisterTimer(std::uint64_t id) {
+  std::unique_lock<std::mutex> lk(mu_);
+  // `CancelAll` might have emptied the `timers_` map, so `id` might not point
+  // to a valid timer, but it's OK.
+  timers_.erase(id);
+}
+
+void OutstandingTimers::CancelAll() {
+  absl::flat_hash_map<std::uint64_t, future<void>> to_cancel;
+  {
+    std::lock_guard<std::mutex> lk(mu_);
+    if (shutdown_) {
+      // Already cancelled
+      return;
+    }
+    shutdown_ = true;
+    // We don't want to fire the timer continuations with the lock held to avoid
+    // deadlocks, so we shouldn't call `cancel()` here.
+    to_cancel.swap(timers_);
+  }
+  for (auto& fut : to_cancel) {
+    fut.second.cancel();
+  }
 }
 
 }  // namespace internal
--- a/google/cloud/bigtable/internal/common_client.h
+++ b/google/cloud/bigtable/internal/common_client.h
@@ -18,10 +18,12 @@
 #include "google/cloud/bigtable/client_options.h"
 #include "google/cloud/bigtable/version.h"
 #include "google/cloud/connection_options.h"
+#include "google/cloud/internal/absl_flat_hash_map_quiet.h"
 #include "google/cloud/internal/random.h"
 #include "google/cloud/log.h"
 #include "google/cloud/status_or.h"
 #include <grpcpp/grpcpp.h>
+#include <list>
 
 namespace google {
 namespace cloud {
@@ -38,6 +40,30 @@ namespace internal {
  */
 std::chrono::seconds constexpr kConnectionReadyTimeout(10);
 
+class OutstandingTimers
+    : public std::enable_shared_from_this<OutstandingTimers> {
+ public:
+  explicit OutstandingTimers(std::shared_ptr<CompletionQueue> const& cq)
+      : weak_cq_(cq) {}
+  // Register a timer. It will automatically deregister on completion.
+  void RegisterTimer(future<void> fut);
+  // Cancel all currently registered timers and all which will be registered in
+  // the future.
+  void CancelAll();
+
+ private:
+  void DeregisterTimer(std::uint64_t id);
+  std::mutex mu_;
+  bool shutdown_ = false;           // GUARDED_BY(mu_)
+  std::uint64_t id_generator_ = 0;  // GUARDED_BY(mu_)
+  absl::flat_hash_map<std::uint64_t,
+                      future<void>> timers_;  // GUARDED_BY(mu_)
+  // Object of this class is owned by timers continuations, which means it
+  // cannot have an owning reference to the `CompletionQueue` because  it would
+  // otherwise create a risk of a deadlock on the completion queue destruction.
+  std::weak_ptr<CompletionQueue> weak_cq_;  // GUARDED_BY(mu_)
+};
+
 /**
  * State required by timers scheduled by `CommonClient`.
  *
@@ -47,13 +73,16 @@ std::chrono::seconds constexpr kConnectionReadyTimeout(10);
 class ConnectionRefreshState {
  public:
   explicit ConnectionRefreshState(
+      std::shared_ptr<CompletionQueue> const& cq,
       std::chrono::milliseconds max_conn_refresh_period);
   std::chrono::milliseconds RandomizedRefreshDelay();
+  OutstandingTimers& timers() { return *timers_; }
 
  private:
   std::mutex mu_;
   std::chrono::milliseconds max_conn_refresh_period_;
   google::cloud::internal::DefaultPRNG rng_;
+  std::shared_ptr<OutstandingTimers> timers_;
 };
 
 /**
@@ -95,14 +124,13 @@ class CommonClient {
             google::cloud::internal::DefaultBackgroundThreads(1)),
         cq_(std::make_shared<CompletionQueue>(background_threads_->cq())),
         refresh_state_(std::make_shared<ConnectionRefreshState>(
-            options_.max_conn_refresh_period())) {}
+            cq_, options_.max_conn_refresh_period())) {}
 
   ~CommonClient() {
     // This will stop the refresh of the channels.
     channels_.clear();
-    // TODO(2567): remove this call when the user will have to provide their own
-    // `CompletionQueues`
-    background_threads_->cq().CancelAll();
+    // This will cancel all pending timers.
+    refresh_state_->timers().CancelAll();
   }
 
   /**
--- a/google/cloud/bigtable/internal/common_client_test.cc
+++ b/google/cloud/bigtable/internal/common_client_test.cc
@@ -1,0 +1,111 @@
+// Copyright 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+
+#include "google/cloud/bigtable/internal/common_client.h"
+#include "google/cloud/testing_util/assert_ok.h"
+#include <gmock/gmock.h>
+
+namespace google {
+namespace cloud {
+namespace bigtable {
+inline namespace BIGTABLE_CLIENT_NS {
+namespace internal {
+
+using TimerFuture = future<StatusOr<std::chrono::system_clock::time_point>>;
+
+class OutstandingTimersTest : public ::testing::Test {
+ public:
+  OutstandingTimersTest() : thread_([this] { cq_.Run(); }) {}
+  ~OutstandingTimersTest() override {
+    cq_.Shutdown();
+    thread_.join();
+  }
+
+ protected:
+  CompletionQueue cq_;
+  std::thread thread_;
+};
+
+TEST_F(OutstandingTimersTest, Trivial) {
+  std::make_shared<OutstandingTimers>(std::make_shared<CompletionQueue>(cq_));
+}
+
+TEST_F(OutstandingTimersTest, TimerFinishes) {
+  auto registry = std::make_shared<OutstandingTimers>(
+      std::make_shared<CompletionQueue>(cq_));
+  promise<void> continuation_promise;
+  auto t = cq_.MakeRelativeTimer(std::chrono::milliseconds(2))
+               .then([&](TimerFuture fut) {
+                 EXPECT_STATUS_OK(fut.get());
+                 continuation_promise.set_value();
+               });
+  registry->RegisterTimer(std::move(t));
+  continuation_promise.get_future().get();
+  // This should be a noop.
+  registry->CancelAll();
+  // Calling it twice shouldn't hurt.
+  registry->CancelAll();
+}
+
+TEST_F(OutstandingTimersTest, TimerIsCancelled) {
+  auto registry = std::make_shared<OutstandingTimers>(
+      std::make_shared<CompletionQueue>(cq_));
+  promise<void> continuation_promise;
+  auto t =
+      cq_.MakeRelativeTimer(std::chrono::hours(10)).then([&](TimerFuture fut) {
+        auto res = fut.get();
+        EXPECT_FALSE(res);
+        EXPECT_EQ(StatusCode::kCancelled, res.status().code());
+        continuation_promise.set_value();
+      });
+  registry->RegisterTimer(std::move(t));
+  registry->CancelAll();
+  continuation_promise.get_future().get();
+}
+
+TEST_F(OutstandingTimersTest, TimerOutlivesRegistry) {
+  auto registry = std::make_shared<OutstandingTimers>(
+      std::make_shared<CompletionQueue>(cq_));
+  promise<void> continuation_promise;
+  auto t = cq_.MakeRelativeTimer(std::chrono::milliseconds(10))
+               .then([&](TimerFuture fut) {
+                 auto res = fut.get();
+                 EXPECT_STATUS_OK(res);
+                 continuation_promise.set_value();
+               });
+  registry->RegisterTimer(std::move(t));
+  registry.reset();
+  continuation_promise.get_future().get();
+}
+
+TEST_F(OutstandingTimersTest, TimerRegisteredAfterCancelAllGetCancelled) {
+  auto registry = std::make_shared<OutstandingTimers>(
+      std::make_shared<CompletionQueue>(cq_));
+  promise<void> continuation_promise;
+  auto t =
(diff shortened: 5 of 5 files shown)


coryan (commented):
This is changing the semantics of destroying any object owning background threads.  Before, the destructor would wait until all operations finished, now it cancels them.

We would need to announce any changes in behavior in the CHANGELOG.md file, but I cannot say with any confidence how these classes would be affected:

- `pubsub::Publisher`: there are timers to flush any batched messages, presumably those will fail to be delivered now?
- `pubsub::Subscriber`: the CQ is used to schedule messages in separate thread pools, what happens to them (I think this is Okay because it is similar to the normal shutdown process)?  The shutdown process for fire&forget subscriptions may be faster now, that seems better.
- `spanner::Client`: I think this is Okay, the session pool is destroyed before the CQ anyway.
- `spanner::InstanceAdminClient`: my guess is no change, but there are pending long running operations to think through.
- `spanner::DatabaseAdminClient`: same as the previous one.

Any thoughts?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-:-MPpJ-Z82I6jIP3OTfCU:bhfm1d0)* status: :shipit: complete! all files reviewed, all discussions resolved

dopiera (commented):
I'm afraid you're right. We need to solve it properly, which means selectively cancelling the timers. To be fair, I wasn't even aware that cancellation was added. PTAL.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-:-MPuACuG-S_0S0esvYs8:bk96mdc)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @coryan)

coryan (changes requested):
> Reviewed 4 of 7 files at r2, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 5 unresolved discussions (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.h, line 47 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPzBOjS2q9rNSqZaeSn:-MPzBOjS2q9rNSqZaeSo:b-1zruyt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.h#L47)):*
> ```objc
>   // A dummy object whose destruction will remove the timer from this
>   // registry.
>   using TimerDeleter = std::shared_ptr<void>;
> ```

I am not sure if this is more useful than a `std::function<void()>`, you are not really relying on the destructor to do any work, you assume that `CancelAll()` will do the necessary work.

But see below, I think there is even a simpler way.
___
*[google/cloud/bigtable/internal/common_client.cc, line 43 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPz84gJCAbiKedD5rw3:-MPz84gK4D8YKL5T0-GH:b-y1ewb4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.cc#L43)):*
> ```cpp
>   // will simply not continue. Unfortunately, that means that some stray
>   // timers may remain in the `CompletionQueue` for a while, but this is
>   // generally unavoidable because there is no way to cancel individual
> ```

This comment seems obsolete.
___
*[google/cloud/bigtable/internal/common_client.cc, line 95 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPz8hoFAGAJkYAWCs9m:-MPz8hoG-kLYX-X-rXC4:bxvznb4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.cc#L95)):*
> ```cpp
>   }
>   timers_.emplace_back(std::move(fut));
>   auto iter = --timers_.end();
> ```

This business with the iterator seems overly complicated.  What do you think of:

- Assign a 64-bit integer to each timer.
- Keep the timers in a `std::unordered_map<std::int64_t, future<void>>` (or the `absl::flat_hash_map<>` version)
- When a timer is cancelled you just remove the entry with "its" integer.
- When canceling all the timers just swap the maps to do any work outside the lock.
- Just rely on the 
___
*[google/cloud/bigtable/internal/common_client.cc, line 100 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPzA7Re804QZxTJxr_2:-MPzA7Re804QZxTJxr_3:bf05nnc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.cc#L100)):*
> ```cpp
>       // Using `this` here is completely arbitrary. It could have been
>       // 0xdeadbeef. The address is not touched anyway.
>       reinterpret_cast<void*>(this), [iter, weak_self](void*) {
> ```

Instead of creating another `shared_ptr<>` to do this work you could just attach that work to the future, something like:

```cc
void OutstandingTimers::RegisterTimer(future<void> f) {
  auto index = ++index_generator;
  auto result = f.then([w = weak_from_this(), index](future<void> f) {
    // When the future is satisfied, remove the timer from the list
    auto self = w.lock();
    self->RemoveTimer(index);
  });
  timers_.insert(index, result);
}
```
___
*[google/cloud/bigtable/internal/common_client.cc, line 109 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPz8Qcg-1bNDgzxktdu:-MPz8Qcg-1bNDgzxktdv:b-ahhhj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.cc#L109)):*
> ```cpp
> 
> void OutstandingTimers::CancelAll() {
>   std::vector<future<void>> to_cancel;
> ```

Seems like you could just `swap()` the lists and do the work outside the lock after the swap?

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 5 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigtable/internal/common_client.h, line 47 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPzBOjS2q9rNSqZaeSn:-MPzcRFG5CZMijTXVitO:bo3ay2l) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.h#L47)):*

I am not sure if this is more useful than a `std::function<void()>`, you are not really relying on the destructor to do any work, you assume that `CancelAll()` will do the necessary work.

But see below, I think there is even a simpler way.
</blockquote></details>

Yup, your way is clearly better.
___
*[google/cloud/bigtable/internal/common_client.cc, line 43 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPz84gJCAbiKedD5rw3:-MPzcU1e2-jz0ZWdLMax:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.cc#L43)):*

This comment seems obsolete.
</blockquote></details>

Done.
___
*[google/cloud/bigtable/internal/common_client.cc, line 95 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPz8hoFAGAJkYAWCs9m:-MPzfC8K5nmA1JzDRhMF:b-u6ai6f) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.cc#L95)):*

This business with the iterator seems overly complicated.  What do you think of:

- Assign a 64-bit integer to each timer.
- Keep the timers in a `std::unordered_map<std::int64_t, future<void>>` (or the `absl::flat_hash_map<>` version)
- When a timer is cancelled you just remove the entry with "its" integer.
- When canceling all the timers just swap the maps to do any work outside the lock.
- Just rely on the 

</blockquote></details>

I think that whether a simple `swap` is possible or not, doesn't depend on whether we use iterators or IDs. Instead, it depends on whether we clear the collection on `CancelAll` or gradually (when each timer finishes). I changed that part now.

Honestly, I think the `unordered_map` introduces more complexity than there currently is. Rather than using the iterators directly, we're reinventing them by introducing a `std::int64` serving the same purpose. This `std::int64` "iterator" has the same validity guarantees as `std::list<>::iterator`, except we introduce an additional moving part and we need to care about how we generate those IDs.
___
*[google/cloud/bigtable/internal/common_client.cc, line 100 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPzA7Re804QZxTJxr_2:-MPzcZzbBC-jJ9TL4oxf:b-uybn1m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.cc#L100)):*

Instead of creating another `shared_ptr<>` to do this work you could just attach that work to the future, something like:

```cc
void OutstandingTimers::RegisterTimer(future<void> f) {
  auto index = ++index_generator;
  auto result = f.then([w = weak_from_this(), index](future<void> f) {
    // When the future is satisfied, remove the timer from the list
    auto self = w.lock();
    self->RemoveTimer(index);
  });
  timers_.insert(index, result);
}
```


</blockquote></details>

Yes, this is much cleaner, I didn't think about it.
___
*[google/cloud/bigtable/internal/common_client.cc, line 109 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MPz8Qcg-1bNDgzxktdu:-MPzccO9EgxTAzGBpHI4:bejtfxt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d059c39b49e88536461ec52b843778b61a93ebc6/google/cloud/bigtable/internal/common_client.cc#L109)):*

Seems like you could just `swap()` the lists and do the work outside the lock after the swap?
</blockquote></details>

There's a race condition if I simply swapped them. The timer continuations will try to deregister themselves and call `erase(iter)` on an empty list, which would be an error.

But I think it's avoidable if I just check if `cancel_all` is true. PTAL.

coryan (changes requested):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 3 unresolved discussions (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.h, line 53 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQ2_h6x6lP3pG-bPSTz:-MQ2_h6x6lP3pG-bPSU-:bdwlgbi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3e64632f077516d4b36f470b65a682a93b8ca6fd/google/cloud/bigtable/internal/common_client.h#L53)):*
> ```objc
>  private:
>   std::mutex mu_;
>   bool cancel_all_ = false;         // GUARDED_BY(mu_)
> ```

nit: the name of this flag is confusing, it really means something like "this thing is now shutdown and not accepting more work".. maybe `shutdown_`?
___
*[google/cloud/bigtable/internal/common_client.cc, line 84 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQ2_B9e4orX85TYxTfh:-MQ2_B9e4orX85TYxTfi:bblvjro) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3e64632f077516d4b36f470b65a682a93b8ca6fd/google/cloud/bigtable/internal/common_client.cc#L84)):*
> ```cpp
> 
>   timers_.emplace_back(future<void>());
>   auto iter = --timers_.end();
> ```

why not `std::back()`?
___
*[google/cloud/bigtable/internal/common_client.cc, line 87 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQ2a8tQEtx937_bocUd:-MQ2a8tRAevOyzBTaosi:b5i25xf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3e64632f077516d4b36f470b65a682a93b8ca6fd/google/cloud/bigtable/internal/common_client.cc#L87)):*
> ```cpp
>   lk.unlock();
>   std::weak_ptr<OutstandingTimers> weak_self(shared_from_this());
>   *iter = fut.then([weak_self, iter](future<void>) {
> ```

There is a race condition here... consider:

- thread A is about to call `CancelAll()`
- thread B is about to set `*iter = ...`
- thread A starts running, swaps the pointer and calls `cancel()`
- thread B calls `.then()` but the timer is already canceled and satisfied.
- thread A releases the lock and (since `CancelAll()` is called) invalidates all the iterators.
- thread B calls `*iter = ...`.

I think you want to do then `.then()` before grabbing the lock, and do all the other work inside the lock.

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigtable/internal/common_client.cc, line 84 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQ2_B9e4orX85TYxTfh:-MQ2jtD5BOwjj8KfKRcd:bep02zn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3e64632f077516d4b36f470b65a682a93b8ca6fd/google/cloud/bigtable/internal/common_client.cc#L84)):*

why not `std::back()`?
</blockquote></details>

It returns a reference to the element, not the iterator. I figured using `emplace_front()` and `begin()` is probably more readable.
___
*[google/cloud/bigtable/internal/common_client.cc, line 87 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQ2a8tQEtx937_bocUd:-MQ2kB3_BBFTgqhNj02g:b-2jup0n) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3e64632f077516d4b36f470b65a682a93b8ca6fd/google/cloud/bigtable/internal/common_client.cc#L87)):*

There is a race condition here... consider:

- thread A is about to call `CancelAll()`
- thread B is about to set `*iter = ...`
- thread A starts running, swaps the pointer and calls `cancel()`
- thread B calls `.then()` but the timer is already canceled and satisfied.
- thread A releases the lock and (since `CancelAll()` is called) invalidates all the iterators.
- thread B calls `*iter = ...`.

I think you want to do then `.then()` before grabbing the lock, and do all the other work inside the lock.

</blockquote></details>

Ooops, you're right. I need to hold the lock to create the iterator (be it an actual iterator or an ID), though, so `then()` has to happen after the lock is held. It shouldn't happen with the lock held, however.

I think the easiest way to fix this is to revert to what it used to be, i.e. always let the timers deregister themselves. That way we reduce the number of control flows and hopefully increase simplicity. This comes at the cost of `CancelAll()`leaving those tombstones like it used to in the original implementation.

coryan (changes requested):
> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 1 unresolved discussion (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.cc, line 87 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQ2a8tQEtx937_bocUd:-MQ34Df84ue0i7ybUJK0:bls1jqm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3e64632f077516d4b36f470b65a682a93b8ca6fd/google/cloud/bigtable/internal/common_client.cc#L87)):*

Ooops, you're right. I need to hold the lock to create the iterator (be it an actual iterator or an ID), though, so `then()` has to happen after the lock is held. It shouldn't happen with the lock held, however.

I think the easiest way to fix this is to revert to what it used to be, i.e. always let the timers deregister themselves. That way we reduce the number of control flows and hopefully increase simplicity. This comes at the cost of `CancelAll()`leaving those tombstones like it used to in the original implementation.
</blockquote></details>

You have a different race condition now?  Maybe less critical:

- thread 1 is in `RegisterTimer()` with the lock held.
- thread 2 is in `CancelAll()` trying to acquire the lock.
- thread 1 calls `emplace_front()`, `timers_.begin()` and `lk.unlock()`.
- thread 2 now starts running, calling `cancel()` on all the timers, note that the first element has a null timer so `cancel()` is a no-op.
- thread 1 calls `*iter = ...`, but that cancelation is never executed.

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[google/cloud/bigtable/internal/common_client.cc, line 87 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQ2a8tQEtx937_bocUd:-MQ4fn-d45cs4zHRY8wy:bw0r8bm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3e64632f077516d4b36f470b65a682a93b8ca6fd/google/cloud/bigtable/internal/common_client.cc#L87)):*

You have a different race condition now?  Maybe less critical:

- thread 1 is in `RegisterTimer()` with the lock held.
- thread 2 is in `CancelAll()` trying to acquire the lock.
- thread 1 calls `emplace_front()`, `timers_.begin()` and `lk.unlock()`.
- thread 2 now starts running, calling `cancel()` on all the timers, note that the first element has a null timer so `cancel()` is a no-op.
- thread 1 calls `*iter = ...`, but that cancelation is never executed.


</blockquote></details>

True.

I didn't come up with any idea to fix it in a non-ugly manner, which makes me think that going back to my original implementation (with the `std::shared_ptr<>`) might not be a bad idea. In fact, I think it is now the simplest.

It seems that the root of all this complexity is `future<>::then()`, which may execute the continuation in-line. The `shared_ptr<>`-based implementation doesn't have this problem. I'm tempted to go back to it.

FWIW, if we went with the future continuations, the least complicated of the ideas which I had is:
* pass a `std::shared_ptr<bool> erased` to the continuation, which will be set to true after the continuation is executed
* reacquire the lock after calling `future<>::then()`
* `if (*erased)  // do nothing`
* `if (shutdown_)  { lk.unlock(); self_unregistering.cancel(); }`
*  otherwise `*iter = std::move(self_unregistering);`

IMO, we should go back to the `shared_ptr<>`. WDYT?

coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 1 unresolved discussion (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.cc, line 87 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQ2a8tQEtx937_bocUd:-MQ81HRM2lBOGGMCq-J0:bojulq1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3e64632f077516d4b36f470b65a682a93b8ca6fd/google/cloud/bigtable/internal/common_client.cc#L87)):*

True.

I didn't come up with any idea to fix it in a non-ugly manner, which makes me think that going back to my original implementation (with the `std::shared_ptr<>`) might not be a bad idea. In fact, I think it is now the simplest.

It seems that the root of all this complexity is `future<>::then()`, which may execute the continuation in-line. The `shared_ptr<>`-based implementation doesn't have this problem. I'm tempted to go back to it.

FWIW, if we went with the future continuations, the least complicated of the ideas which I had is:
* pass a `std::shared_ptr<bool> erased` to the continuation, which will be set to true after the continuation is executed
* reacquire the lock after calling `future<>::then()`
* `if (*erased)  // do nothing`
* `if (shutdown_)  { lk.unlock(); self_unregistering.cancel(); }`
*  otherwise `*iter = std::move(self_unregistering);`

IMO, we should go back to the `shared_ptr<>`. WDYT?
</blockquote></details>

Consider this instead:

- Using a hash map for `timers_`.
- Store a CQ in OutstandingTimers to delay some operations and avoid deadlocks (alternatively, consider `std::recursive_mutex`).


```cc
void OutstandingTimers::RegisterTimer(future<void> fut) {
  if (shutdown_) {
    lk.unlock();
    fut.cancel();
    return;
  }
  std::weak_ptr<OutstandingTimers> weak(shared_from_this());
  auto id = ++id_generator_;
  auto f = fut.then([weak, id](future<void>) {
    if (auto self = weak.lock()) self->AsyncDeregisterTimer(id);
  });
  timers_->insert(id, 
}

void OutstandingTimers::AsyncDeregisterTimer(std::int64_t id) {
  std::weak_ptr<OutstandingTimers> weak(shared_from_this());
  // Run in a separate thread to avoid deadlocks when the timer is immediately satisfied.
  cq_->RunAsyncDeregisterTimer([weak, id]() { if (auto self = weak.lock()) self->DeregisterTimer(id); });
}

void OutstandingTimers::DeregisterTimer(std::int64_t id) {
  timers_.erase(id);
}

void OutstandingTimers::CancelAll() {
  absl::flat_hash_map<std::int64_t, future<void>> timers;
  {
    timers.swap(timers_);
    shutdown_ = true;
  }
  for (auto& kv : timers) kv->second.cancel();
```

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[google/cloud/bigtable/internal/common_client.cc, line 87 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQ2a8tQEtx937_bocUd:-MQ8ckiV4RfRKNNdXCcO:b-9q5g40) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3e64632f077516d4b36f470b65a682a93b8ca6fd/google/cloud/bigtable/internal/common_client.cc#L87)):*

Consider this instead:

- Using a hash map for `timers_`.
- Store a CQ in OutstandingTimers to delay some operations and avoid deadlocks (alternatively, consider `std::recursive_mutex`).


```cc
void OutstandingTimers::RegisterTimer(future<void> fut) {
  if (shutdown_) {
    lk.unlock();
    fut.cancel();
    return;
  }
  std::weak_ptr<OutstandingTimers> weak(shared_from_this());
  auto id = ++id_generator_;
  auto f = fut.then([weak, id](future<void>) {
    if (auto self = weak.lock()) self->AsyncDeregisterTimer(id);
  });
  timers_->insert(id, 
}

void OutstandingTimers::AsyncDeregisterTimer(std::int64_t id) {
  std::weak_ptr<OutstandingTimers> weak(shared_from_this());
  // Run in a separate thread to avoid deadlocks when the timer is immediately satisfied.
  cq_->RunAsyncDeregisterTimer([weak, id]() { if (auto self = weak.lock()) self->DeregisterTimer(id); });
}

void OutstandingTimers::DeregisterTimer(std::int64_t id) {
  timers_.erase(id);
}

void OutstandingTimers::CancelAll() {
  absl::flat_hash_map<std::int64_t, future<void>> timers;
  {
    timers.swap(timers_);
    shutdown_ = true;
  }
  for (auto& kv : timers) kv->second.cancel();
```

</blockquote></details>

I rewrote it this way, but it adds complexity elsewhere - we need to be really cautious to make sure that the continuations do not hold an indirect owning reference to the completion queue.

I think it works, but honestly, but I think it's even more brittle now. If it was up to me, I'd still go with `shared_ptr<>`s , but if you prefer it this way, I'm happy too.

coryan (changes requested):
> Reviewed 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 2 unresolved discussions (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.h, line 46 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQCuJ91E6ikdyDKwTUw:-MQCuJ91E6ikdyDKwTUx:b-mmsviz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1980823d5ecb6ea08e49f6b8f4869887dcddcf58/google/cloud/bigtable/internal/common_client.h#L46)):*
> ```objc
>     : public std::enable_shared_from_this<OutstandingTimers> {
>  public:
>   explicit OutstandingTimers(std::shared_ptr<CompletionQueue> const& cq)
> ```

I am not sure I follow, even if this class owns a `CompletionQueue`, there are no cycles because the futures (and timers) all own weak references to this class, right?  Yes, those `weak_ptr<>` get upgraded, but those are for relatively short duration.
___
*[google/cloud/bigtable/internal/common_client.cc, line 95 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQCtdSuDpKc39f3XFYz:-MQCtdSvBV9rsypih2a-:b-6lk49e) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1980823d5ecb6ea08e49f6b8f4869887dcddcf58/google/cloud/bigtable/internal/common_client.cc#L95)):*
> ```cpp
>     // Do not run in-line to avoid deadlocks when the timer is immediately
>     // satisfied.
>     cq->RunAsync([self, id, weak_self] {
> ```

You probably do not want to copy `self` here?

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigtable/internal/common_client.h, line 46 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQCuJ91E6ikdyDKwTUw:-MQECc_W3oRHxWcx05Ov:b-nsq3dt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1980823d5ecb6ea08e49f6b8f4869887dcddcf58/google/cloud/bigtable/internal/common_client.h#L46)):*

I am not sure I follow, even if this class owns a `CompletionQueue`, there are no cycles because the futures (and timers) all own weak references to this class, right?  Yes, those `weak_ptr<>` get upgraded, but those are for relatively short duration.
</blockquote></details>

These continuations - yes. The problem is with the ones which `ScheduleChannelRefresh()` creates. Those hold a `ChannelRefreshState`, which holds `OutstandingTimers`.
___
*[google/cloud/bigtable/internal/common_client.cc, line 95 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQCtdSuDpKc39f3XFYz:-MQECYIf8DKG7k6O90uq:b-jtno6d) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1980823d5ecb6ea08e49f6b8f4869887dcddcf58/google/cloud/bigtable/internal/common_client.cc#L95)):*

You probably do not want to copy `self` here?
</blockquote></details>

Actually, I do not want `weak_self` :)

coryan (approved):
:lgtm:

> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-:-MQELj9N3RvVrojUv04o:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @dopiera)

___
*[google/cloud/bigtable/internal/common_client.h, line 46 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQCuJ91E6ikdyDKwTUw:-MQELDgaBI-mBZAONwVg:byrr4hn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1980823d5ecb6ea08e49f6b8f4869887dcddcf58/google/cloud/bigtable/internal/common_client.h#L46)):*

These continuations - yes. The problem is with the ones which `ScheduleChannelRefresh()` creates. Those hold a `ChannelRefreshState`, which holds `OutstandingTimers`.
</blockquote></details>

Ack.  That seems like a different problem to fix (maybe).
___
*[google/cloud/bigtable/internal/common_client.cc, line 95 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQCtdSuDpKc39f3XFYz:-MQELJUb-N_KHW8Zv5FE:bpshhzm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1980823d5ecb6ea08e49f6b8f4869887dcddcf58/google/cloud/bigtable/internal/common_client.cc#L95)):*

Actually, I do not want `weak_self` :)
</blockquote></details>

Hmm... that does create a loop, as `cq` will  own the lambda for "a short while" (until it is dequeued and executed), but no leaks I think (on shutdown the queue is cleared).

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan and @dopiera)

___
*[google/cloud/bigtable/internal/common_client.cc, line 95 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQCtdSuDpKc39f3XFYz:-MQH2vfj2AtO6x_1D3u4:b2nha9w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1980823d5ecb6ea08e49f6b8f4869887dcddcf58/google/cloud/bigtable/internal/common_client.cc#L95)):*

Hmm... that does create a loop, as `cq` will  own the lambda for "a short while" (until it is dequeued and executed), but no leaks I think (on shutdown the queue is cleared).
</blockquote></details>

ACK

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigtable/internal/common_client.h, line 46 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5667#-MQCuJ91E6ikdyDKwTUw:-MQH3rEM58RZm-SCV-kT:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1980823d5ecb6ea08e49f6b8f4869887dcddcf58/google/cloud/bigtable/internal/common_client.h#L46)):*

Ack.  That seems like a different problem to fix (maybe).
</blockquote></details>

Done.

PR #5681, opened 2021-01-06
fix: allow DefaulBackgroundThreads to self-destroy

dopiera:
The last reference to `AutomaticallyCreatedBackgroundThreads` may be in
the `CompletionQueue` owned by itself. This means that
`~AutomaticallyCreatedBackgroundThreads()` will be executed in the
thread belonging to it. Before this PR this resulted in a deadlock. This
PR fixes it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5681)

### DIFF
--- a/google/cloud/internal/background_threads_impl.cc
+++ b/google/cloud/internal/background_threads_impl.cc
@@ -37,12 +37,28 @@ AutomaticallyCreatedBackgroundThreads::AutomaticallyCreatedBackgroundThreads(
 
 AutomaticallyCreatedBackgroundThreads::
     ~AutomaticallyCreatedBackgroundThreads() {
-  Shutdown();
-}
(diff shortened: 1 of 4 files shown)


coryan:
When we have ran into this problem before we solved it by separating the ownership of the thread pool from the users of the CQ.  For example, in pubsub we do this:

https://github.com/googleapis/google-cloud-cpp/blob/36bfcefb1ac8db778e5903424ea5422aa61f2e28/google/cloud/pubsub/publisher_connection.cc#L35

In spanner we did this:

https://github.com/googleapis/google-cloud-cpp-spanner/pull/1397

coryan:
Closing for now, we can reopen later.

PR #5718, opened 2021-01-20
test(spanner): add an additional transient failure test

mr-salty:
If the intial read from a streaming operation fails when also
trying to implicitly start a transaction, ensure the library properly
recovers and completes the operation (using an explicit
`BeginTransaction` call).

This was definitely a gap in our test coverage; @thiagotnunes brought a
Java customer issue (https://github.com/googleapis/java-spanner/issues/799)
to my attention, so I wanted to ensure C++ users were not susceptible.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5718)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl_test.cc
+++ b/google/cloud/spanner/internal/connection_impl_test.cc
@@ -482,7 +482,6 @@ TEST(ConnectionImplTest, ReadImplicitBeginTransactionOnePermanentFailure) {
   auto failing_reader = MakeFailingReader(grpc_status);
   auto constexpr kText = R"pb(
     metadata: {
-      transaction: { id: "ABCDEF00" }
       row_type: {
         fields: {
           name: "UserId",
--- a/google/cloud/spanner/internal/connection_impl_test.cc
+++ b/google/cloud/spanner/internal/connection_impl_test.cc
@@ -415,6 +415,69 @@ TEST(ConnectionImplTest, ReadImplicitBeginTransactionOneTransientFailure) {
   auto conn = MakeConnection(
       db, {mock},
       spanner::ConnectionOptions{grpc::InsecureChannelCredentials()});
+  grpc::Status grpc_status(grpc::StatusCode::UNAVAILABLE, "uh-oh");
+  auto failing_reader = MakeFailingReader(grpc_status);
+  auto constexpr kText = R"pb(
+    metadata: {
+      transaction: { id: "ABCDEF00" }
+      row_type: {
+        fields: {
+          name: "UserId",
+          type: { code: INT64 }
+        }
+        fields: {
+          name: "UserName",
+          type: { code: STRING }
+        }
+      }
+    }
+    values: { string_value: "12" }
+    values: { string_value: "Steve" }
+    values: { string_value: "42" }
+    values: { string_value: "Ann" }
+  )pb";
+  auto ok_reader = MakeReader({kText});
+
+  // n.b. these calls are explicitly sequenced because using the scoped
+  // `InSequence` object causes gMock to get confused by the reader calls.
+  Sequence s;
+  EXPECT_CALL(*mock, BatchCreateSessions(_, HasDatabase(db)))
+      .InSequence(s)
+      .WillOnce(Return(MakeSessionsResponse({"test-session-name"})));
+  EXPECT_CALL(*mock, StreamingRead(_, AllOf(HasSession("test-session-name"),
+                                            HasBeginTransaction())))
+      .InSequence(s)
+      .WillOnce(Return(ByMove(std::move(failing_reader))));
+  EXPECT_CALL(*mock, StreamingRead(_, AllOf(HasSession("test-session-name"),
+                                            HasBeginTransaction())))
+      .InSequence(s)
+      .WillOnce(Return(ByMove(std::move(ok_reader))));
+  spanner::Transaction txn =
+      MakeReadOnlyTransaction(spanner::Transaction::ReadOnlyOptions());
+  auto rows = conn->Read(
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -408,6 +408,73 @@ TEST(ConnectionImplTest, ReadImplicitBeginTransaction) {
...
 }
 
+TEST(ConnectionImplTest, ReadImplicitBeginTransactionOneTransientFailure) {
+  auto mock = std::make_shared<spanner_testing::MockSpannerStub>();
+  auto db =
+      spanner::Database("dummy_project", "dummy_instance", "dummy_database_id");
+  auto conn = MakeConnection(
+      db, {mock},
+      spanner::ConnectionOptions{grpc::InsecureChannelCredentials()});
+  grpc::Status grpc_status(grpc::StatusCode::PERMISSION_DENIED, "uh-oh");
```
coryan, line 418: `PERMISSION_DENIED` is not usually transient, did I miss something?

devbww: If an RPC with a piggy-backed "begin transaction" fails for any reason, we do an explicit BeginTransaction() and then (if that succeeded) try the RPC again.  We need to do that to maintain the transaction semantics.

mr-salty (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5718)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/spanner/internal/connection_impl_test.cc, line 418 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5718#-MRTkBFZZJp9VyIH6-DQ-r1-418:-MRYWloc3JtaooGOVlkt:bclfaw8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f5c89e8bfe396b9d080133ac0c940a7bd4286277/google/cloud/spanner/internal/connection_impl_test.cc#L418)):*

If an RPC with a piggy-backed "begin transaction" fails for any reason, we do an explicit BeginTransaction() and then (if that succeeded) try the RPC again.  We need to do that to maintain the transaction semantics.
</blockquote></details>

good catch, it was supposed to be `UNAVAILABLE` as in the bug I referenced, but I made a copy/paste error. 

However, I would have said exactly what Brad said, that the error didn't really matter in this context because we will do the explicit `BeginTransaction` regardless of whether the failure was transient vs. not.  The truth turned out to be more complicated -  [`PartialResultSetResume::Read()`](https://github.com/googleapis/google-cloud-cpp/blob/36bfcefb1ac8db778e5903424ea5422aa61f2e28/google/cloud/spanner/internal/partial_result_set_resume.cc#L26-L45)  will retry the RPC if it's a transient error.  Anyways, I added tests for both cases although I don't know that the `OnePermanentFailure` case ever arises in practice.

I'm a bit concerned about whether the retry invalidates any assumptions we made. I think it does not, because we still cannot get in the bad state where the operation succeeds without starting nor invalidating the transaction. If it keeps retrying and getting the same transient error, eventually it will stop retrying, at which point we will try to do a separate `BeginTransaction`, and possibly doing the operation again.  I think that ends up potentially doing more retries than we should, but it's correct. But, maybe `PartialResultSetResume` should have a way to just return an error in this case, I'm not going to attempt to tackle that but we can discuss it.
___
*[google/cloud/spanner/internal/connection_impl_test.cc, line 422 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5718#-MRYip0ZD8MIRGVVNHz_:-MRYip0ZD8MIRGVVNHza:b-ojg0zd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f5c89e8bfe396b9d080133ac0c940a7bd4286277/google/cloud/spanner/internal/connection_impl_test.cc#L422)):*
> ```
>       transaction: { id: "ABCDEF00" }
> ```

I noticed this... since this response only comes after the explicit `BeginTransaction`, the `transaction` here is just silently ignored. Is that a bug (should we fail with `INTERNAL`), WAI (it's harmless?), or somewhere in the middle (maybe log it?)

coryan (approved):
:lgtm: but you probably want @devbww to answer the question below and do a deeper review.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5718#-:-MRZy9Au4CqgDQ0vOH1S:b-pyp632)* status: all files reviewed, 1 unresolved discussion

mr-salty (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5718)* status: all files reviewed, 1 unresolved discussion (waiting on @mr-salty)

___
*[google/cloud/spanner/internal/connection_impl_test.cc, line 422 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5718#-MRYip0ZD8MIRGVVNHz_:-MRauK5F0t3pQ4Cdk9_Z:b-hqjbz9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f5c89e8bfe396b9d080133ac0c940a7bd4286277/google/cloud/spanner/internal/connection_impl_test.cc#L422)):*

> ```
>       transaction: { id: "ABCDEF00" }
> ```

I noticed this... since this response only comes after the explicit `BeginTransaction`, the `transaction` here is just silently ignored. Is that a bug (should we fail with `INTERNAL`), WAI (it's harmless?), or somewhere in the middle (maybe log it?)
</blockquote></details>

I did mean to remove this (as @devbww pointed out in chat); what should happen was a separate side discussion.

devbww (approved):

PR #5722, opened 2021-01-20
refactor!: removed old bigquery code

devjgm:
In preparation for a supported BigQuery library, this PR removes the old experimental BQ library. This is "breaking" if
someone were using it. However, this library was only "experimental", so I doubt any users will actually have production workloads using this.

Note: I left the bigquery protos in the cmake files so that we continue compiling and linking these protos since we'll need them again soon.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5722)


coryan (approved):
LGTM

Issue #5726, opened 2021-01-22
Proposal: consistent names for libraries and targets in CMake


coryan:
## Status

This has been implemented. We are waiting until 2022-02-15 to retire the legacy Bazel rules, CMake packages, and CMake rules.

## Background

The Bigtable client library is called `libbigtable_client.{a,so}`, the CMake target is called `bigtable_client`.  The storage library follows the same pattern.

These names are prone to conflicts. Any other application or library provider could create a `storage_client` target.

When it came time to write Spanner we created an alias `googleapis-c++::spanner_client`, which at least is less prone to conflicts, but it is too redundant: it is unlikely that our Java googleapis will have CMake targets.  PubSub also uses the `googleapis-c++::` prefix.

The common libraries (`libgoogle_cloud_cpp_common.{a,so}` and `libgoogle_cloud_cpp_grpc_utils.{a,so}`) do not have targets with the `googleapis-c++::` prefix.

The proto libraries are exported with the `googleapis-c++::` prefix.

## Proposal

* For CMake we will use a prefix (`google-cloud-cpp::`) consistently across **all** our exported libraries.
* For Bazel we will recommend that applications use `//:bigtable`, `//:pubsub`, etc. as the target they link.
* For CMake we will recommend `google-cloud-cpp::bigtable`, `google-cloud-cpp::storage`
* For CMake packages will be imported with:
  ```cmake
  find_package(google_cloud_cpp_bigtable)
  find_package(google_cloud_cpp_pubsub)
  find_package(google_cloud_cpp_storage)
  ```
* Though rarely used, proto libraries will be imported with:
  ```cmake
  find_package(google_cloud_cpp_googleapis)
  ```

### T: As Soon as Possible

* Export all libraries with the `google-cloud-cpp::` prefix
* Change all our examples and tests to use this prefix
* Add aliases for the old names (`storage_client`, `bigtable_client`, etc.)
* Declare these aliases as deprecated and announce that they will be removed ~in 2 * X months~ by a certain date.
* Include this bug as a reference
* We should explore using the `DEPRECATION` [property](https://cmake.org/cmake/help/latest/prop_tgt/DEPRECATION.html) to warn users.
* Change the Bazel in `google/cloud/*/BUILD` to use the `deprecation` directive and reference this bug (as well as the top-level targets), see #3701 
* Rename the physical libraries to `libgoogle_cloud_cpp_${name}.a` (or `.so`)
* Create backwards compatibility physical libraries.

[...]

PR #5741, opened 2021-01-25
test: verify googleapis targets remain usable

coryan:
This creates a test to verify the targets created in
`external/googleapis` remain usable.

Part of the work for #5726

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5741)

### DIFF
--- a/ci/verify_exported_targets/CMakeLists.txt
+++ b/ci/verify_exported_targets/CMakeLists.txt
@@ -35,20 +35,17 @@ find_package(firestore_client REQUIRED)
 find_package(storage_client REQUIRED)
 find_package(spanner_client REQUIRED)
 find_package(pubsub_client REQUIRED)
-pkg_check_modules(google_cloud_cpp_common IMPORTED_TARGET REQUIRED
-                  google_cloud_cpp_common)
-pkg_check_modules(google_cloud_cpp_grpc_utils IMPORTED_TARGET REQUIRED
-                  google_cloud_cpp_grpc_utils)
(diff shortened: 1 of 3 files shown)


ci/verify_exported_targets/CMakeLists.txt
```diff
@@ -71,11 +68,78 @@ add_test_case(t120 storage_client)
...
+set(pkg_config_modules
+    google_cloud_cpp_common google_cloud_cpp_grpc_utils bigtable_client
+    pubsub_client spanner_client storage_client)
+foreach (module ${pkg_config_modules})
+    pkg_check_modules(${module} IMPORTED_TARGET REQUIRED ${module})
+    add_test_case(test_pc_${module} PkgConfig::${module})
+endforeach ()
+
+# This one is special because it does not depend on google_cloud_cpp_common.
+pkg_check_modules(firestore_client IMPORTED_TARGET REQUIRED firestore_client)
```
devjgm, line 84: (no change necessary) Given Firestore's experimental status, I'm not sure we need to do anything special to maintain backward compat for it. However, since we're doing these checks for the other libraries, this Firestore check likely won't add any extra constraints on us. So, I'd be happy to remove these Firestore checks, or keep them, but I did want to comment to record that there was a question about them here.

devjgm (approved):

ci/verify_exported_targets/CMakeLists.txt
coryan: Ack.

PR #5746, opened 2021-01-25
feat: mark legacy Bazel targets as deprecated

coryan:
We want users to use the top-level targets such as `//:spanner`, or `//:storage`. This change marks the existing targets as deprecated, and recommends using the new targets.

Part of the work for #5726 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5746)


BUILD
```diff
@@ -20,30 +20,80 @@ exports_files([
...
+cc_library(
+    name = "bigtable",
+    deps = [
+        "//google/cloud/bigtable:bigtable_client_internal",
+    ],
+)
+
 cc_library(
     name = "bigtable_client",
+    deprecation = "this target will be removed on or around 2022-02-15, please use //:spanner instead.",
```
devbww, line 32: s/spanner/bigtable/

BUILD
```diff
@@ -20,30 +20,80 @@ exports_files([
...
+    name = "pubsub_mocks",
     deps = [
-        "//google/cloud/bigtable:bigtable_client",
+        "//google/cloud/pubsub:pubsub_mocks_internal",
     ],
 )
 
 cc_library(
     name = "pubsub_client",
+    deprecation = "this target will be removed on or around 2022-02-15, please use //:spanner instead.",
```
devbww, line 55: s/spanner/pubsub/

devbww (approved):

google/cloud/bigtable/BUILD
```diff
@@ -38,14 +42,23 @@ cc_library(
...
+
 load(":bigtable_client_testing.bzl", "bigtable_client_testing_hdrs", "bigtable_client_testing_srcs")
 
 cc_library(
     name = "bigtable_client_testing",
     srcs = bigtable_client_testing_srcs,
     hdrs = bigtable_client_testing_hdrs,
     deps = [
-        ":bigtable_client",
+        "//:bigtable",
```
devjgm, line 61: Within this file, using the locally defined `:bigtable_client_internal` seems fine and somewhat more expected.

devjgm (approved):

google/cloud/bigtable/BUILD
coryan: Done.  What is the style for sub-directories (examples, integration tests, benchmarks, etc.)?

devjgm: Not 100% sure, but I think sub-packages would be fully qualified, like `//google/cloud/bigtable/examples:foo`. I can't recall ever seeing relative build targets in BUILD files, other than for targets defined within the local build file.

devbww: I was going to comment on this, but after seeing that the target also needs to be referred to from `//google/cloud/bigtable` sub-packages (and similarly for the other clients), I favored the simple `//:bigtable` everywhere.

But your choice.

devjgm: FWIW, referring to `//google/cloud/bigtable:bigtable_client_internal` in subpackages would be fine with me, and IMO, preferable to having the bigtable package itself link w/ `//:bigtable`.

But I agree not a big deal. It's not a user-facing issue either way, and we can change this in a separate PR.

devbww: > FWIW, referring to //google/cloud/bigtable:bigtable_client_internal in subpackages would be fine with me, and IMO, preferable to having the bigtable package itself link w/ //:bigtable.
> 
> But I agree not a big deal. It's not a user-facing issue either way, and we can change this in a separate PR.

Just for the record, the merged version does not appear to use the "internal" target from the subpackages.

PR #5754, opened 2021-01-27
feat(bigtable): restore from backups in other instance

coryan:
Allow applications to specify a different instance as the location for a
backup.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5754)


coryan:
The lack of testing is annoying, but we ran into this with the existing backup functions: the emulator does not support them, and the quota / performance does not allow running them against production.

devjgm (approved):

PR #5757, opened 2021-01-27
cleanup(common): consistent CMake target names

coryan:
Export the libraries as `google-cloud-cpp::common` and
`google-cloud-cpp::grpc_utils`. The legacy names are still exported for
backwards compatibility. These legacy target names generate warnings
with CMake >= 3.18

Part of the work for #5726

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5757)


coryan:
Ping

devjgm (approved):
Sorry I missed this.

PR #5769, opened 2021-01-28
cleanup(storage): consistent CMake target names

coryan:
Export the storage client library as `google-cloud-cpp::storage`. The
legacy name for this library (`pubsub_client` is also exported for
backwards compatibility. The legacy name generates warnings with
CMake >= 3.18.

I also created an alias and exported name for the GCS+gRPC library. This
library is *not* installed in a normal build, so its name is not that
important. Nevertheless, I called it
`google-cloud-cpp::experimental-gcs-grpc` to make clear that it is not
ready for production.

Part of the changes for #5726

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5769)


scotthart (approved):
> Reviewed 11 of 12 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5769)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #5593, opened 2020-12-08
Upgrade to gRPC >= 1.34.0


devjgm:
This issue is to record some stuff so I don't forget it over the coming holidays.

We want to upgrade to gRPC >= 1.34.0, because this version of gRPC fixes its Abseil dep so that we can upgrade _our_ Abseil dep (grpc no longer looks for that dynamic_annotations.so library).

I tried to do this upgrade in https://github.com/googleapis/google-cloud-cpp/pull/5579, which worked except for a few outstanding issues.

1. The build breaks on windows. This seems to be a gRPC bug: https://github.com/grpc/grpc/issues/24904
2. [FIXED in /pull/5599] Our build breaks on MacOS ([console log](https://source.cloud.google.com/results/invocations/bb9d75fd-5bee-4177-a101-e68d65921073/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fmacos%2Fcmake-super-presubmit/log)). I'm told that this may be related to the TLS version range that gRPC supports (https://github.com/grpc/grpc/pull/23165  may be relevant). Running with `GRPC_VERBOSITY=DEBUG` in the env may help us debug.

coryan:
gRPC 1.35.0 is out, did that fix the Abseil problems?

devjgm:
I'm testing grpc 1.35.0 now. It looks like 1.35.0 starts using `StatusOr` from Aseil, which means we _must_ upgrade Abseil at the same time.

devjgm:
With this update, we see a tsan data race that looks like the following (slightly cleaned up). 

Note: To reproduce this error, patch in PR #5791, and then run `ci/kokoro/docker/build.sh tsan`. The following error is from the `//google/cloud:internal_background_threads_impl_test` BUILD target.

```
WARNING: ThreadSanitizer: data race 
  Atomic write of size 4 at 0x7f5eecedef50 by main thread (mutexes: write M611):
    #0 __tsan_atomic32_fetch_add <null> 
    #1 std::__atomic_base<int>::fetch_add(int, std::memory_order) /usr/bin/../lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/atomic_base.h:514:16 
    #2 absl::lts_2020_09_23::synchronization_internal::Waiter::Post() /proc/self/cwd/external/com_google_absl/absl/synchronization/internal/waiter.cc:174:14 
[...]

PR #5791, opened 2021-02-02
chore: upgrade grpc and abseil

devjgm:
Fixes: #5593

Upgrades gRPC to 1.35.0
Upgrades Abseil to 20200923.3
Installs the new RE2 dep

Upgrading gRPC requires the new dep on RE2 and it also requires an newer version of Abseil that includes `StatusOr`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5791)


doc/packaging.md
```diff
@@ -231,9 +231,9 @@ We need a recent version of Abseil.
 
 ```bash
```
coryan, line 232: We are missing `Abseil` as a dependency further up in this file.  We probably want to update the minimum version for gRPC too.  I doubt we still compile with 1.16

coryan (approved):
One comment below. We could fix that in a future PR if you prefer. Modulo the build problems this :lgtm: 

> Reviewed 23 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5791#-:-MSYtH7w9XH7aZdkh9q_:bhq8q1q)* status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)

doc/packaging.md
devjgm: Ack. Thanks. I filed https://github.com/googleapis/google-cloud-cpp/issues/5796 to track this.

PR #5811, opened 2021-02-04
feat: make StreamRange<T> public

devjgm:
This class is moving from `google/cloud/internal/` to `google/cloud/`, and also in a non-internal namespace. This is now a user-facing class that we want users to be able to use and read the documentation for. We expect to have user-facing functions _directly_ returning objects of type `google::cloud::StreamRange<Foo>`, so it needs to be public.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5811)


google/cloud/stream_range.h
```diff
@@ -85,13 +91,13 @@ class StreamRange {
   template <typename U>
   static constexpr bool IsMoveNoexcept() {
     return noexcept(StatusOr<U>(std::declval<U>()))&& noexcept(
-        StreamReader<U>(std::declval<StreamReader<U>>()));
+        internal::StreamReader<U>(std::declval<internal::StreamReader<U>>()));
   }
 
  public:
   /// An input iterator for a `StreamRange<T>`
```
coryan, line 98: nit: maybe remind users that this is not part of the public API and they should be using the `iterator` alias below.

devjgm: Good idea. Done.

google/cloud/stream_range.h
```diff
@@ -186,22 +196,46 @@ class StreamRange {
...
+   * end of stream, or a non-OK `Status` to indicate an error, or a `T`. The
+   * `internal::StreamReader` will not be invoked again after it returns a
+   * `Status`.
+   *
+   * @par Example: Printing integers from 1-10.
+   *
+   * @code
+   * int counter = 0;
+   * auto reader = [&counter]() -> internal::StreamReader<int>::result_type {
+   *   if (counter++ < 10) return counter;
```
devbww, line 216: Nit for moved example: `if (++counter <= 10) return counter;` is a little nicer.

coryan (approved):
:lgtm:

> Reviewed 5 of 6 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5811#-:-MSjNpAF463gcDs18nbb:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)

google/cloud/stream_range.h
devjgm: Changed here and above.

Issue #5819, opened 2021-02-05
Consider NOT caching ~/.cache/baze/ on our Docker builds


devjgm:
We currently cache all of `~/.cache/` on our Docker Linux builds:

https://github.com/googleapis/google-cloud-cpp/blob/c8e235437e944b4a8623f7e554586ef4100df2cc/ci/kokoro/docker/upload-cache.sh#L42-L43

That directory includes bazel's cache, and that can be ~9GB in my experiments (probably bigger too). We're already using GCS as a remote cache for bazel

https://github.com/googleapis/google-cloud-cpp/blob/c8e235437e944b4a8623f7e554586ef4100df2cc/ci/kokoro/docker/build-in-docker-bazel.sh#L69

So we effectively have two caches of bazel's data.

We should consider no longer caching all of `~/.cache/` and instead only caching specific subdirectories, like `~/.cache/vcpkg`.

HOWEVER...

It's possible that this may slow down the bazel builds. I'm not sure if we get a performance win by downloading bazel's local cache in one go rather than piece-by-piece as I think will happen with the "remote_cache". But there's one way we can find out -- we could try it! :-)

devjgm:
Fixed in https://github.com/googleapis/google-cloud-cpp/pull/5822

PR #5822, opened 2021-02-05
ci: only cache certain .cache/ dirs

devjgm:
Rather than caching all of ~/.cache, we only cache certain subdirs under
there. This will effectively STOP caching ~/.cache/bazel. But from the
rationale in https://github.com/googleapis/google-cloud-cpp/issues/5819,
that may be a good thing.

NOTE: This PR may make the bazel builds SLOWER, so we may end up reverting this PR. That's fine.
We won't know if this makes things slower until after this commit makes it through a full CI build where we upload 
new cache files that no longer contain the ~/.cache/bazel dir. Then the following builds will NOT be using it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5822)

### DIFF
--- a/ci/kokoro/docker/upload-cache.sh
+++ b/ci/kokoro/docker/upload-cache.sh
@@ -39,8 +39,12 @@ maybe_dirs=(
   # This is where ccache stores its files, if present we want to back it up
   "${HOME_DIR}/.ccache"
 
-  # Bazel and vcpkg put their caches within here.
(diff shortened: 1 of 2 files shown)


coryan (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5822)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5841, opened 2021-02-10
ci: add Z switch so mounts work on selinux systems

remyabel:
Otherwise I get a "permission denied" error. See [docker documentation](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5841)


coryan (approved):
Thanks!

Issue #2857, opened 2019-07-05
Use [[deprecated]] attributes.


dopiera:
These attributes are available since C++14.

We could add a macro or wait until dropping support for C++11.

devjgm:
Now that we have Abseil, we could use `ABSL_DEPRECATED("Use Bar instead")`

https://github.com/abseil/abseil-cpp/blob/fbdff6f3ae0ba977a69f172e85ecaede535e70f6/absl/base/attributes.h#L630-L655

devbww:
> Now that we have Abseil, we could use `ABSL_DEPRECATED("Use Bar instead")`

After writing that, @devjgm followed up with, _"I think it would be good for us to understand the implications of ABSL_DEPRECATED on our customers before using it, because it has the ability to break our customers' builds."_

So, I'm reopening this issue.

remyabel:
There are still instances of `TODO(#2857)` in bigtable/instance_admin.h and storage/client.h, those TODO's need to be updated as well.

coryan:
Thanks, I prepared a PR.  I am also asking a colleagues for ideas to automate this TODO <-> bug linking, maybe a bot that warns us when we are about to leave TODOs stale, or reopens bugs that still have pending TODOs.

PR #5844, opened 2021-02-11
refactor(spanner)!: avoid chrono::time_point for Spanner timestamps

devbww:
Introduce new overloads for `DatabaseAdminClient::CreateBackup()` and
`DatabaseAdminClient::UpdateBackupExpireTime()` that take a `Timestamp`
instead of a `std::chrono::system_clock::time_point`. This guarantees
that we can set time instants with full Spanner resolution (ns). Mark
the old interfaces as deprecated.

Also avoid `time_point` when constructing `Timestamp` values in tests.

Relatedly, `Timestamp` has been crafted in such a way that conversion
from a `protobuf::Timestamp` cannot fail, so remove that possibility
in the signature of `internal::TimestampFromProto()` by eliminating
`StatusOr<>` from the return type.

**BREAKING CHANGE:**

Field `DatabaseAdminConnection::CreateBackupParams::expire_time` has
changed from `std::chrono::system_clock::time_point` to `Timestamp`.
This affects mocking of `CreateBackup()` in Spanner connection tests.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5844)


google/cloud/spanner/database_admin_connection.h
```diff
@@ -174,11 +175,11 @@ class DatabaseAdminConnection {
     /// The name of the database.
     Database database;
     std::string backup_id;
-    std::chrono::system_clock::time_point expire_time;
+    Timestamp expire_time;
     // The backup will contain an externally consistent copy of the database
     // at `version_time`. If `version_time` is not specified, the system will
     // set `version_time` to the `create_time` of the backup.
-    absl::optional<std::chrono::system_clock::time_point> version_time;
+    absl::optional<Timestamp> version_time;
```
devjgm, line 182: Changing the types of these fields will be a hard break for customers right away. Maybe we should leave the old fields, and _add_ the new fields. We could mark the old fields deprecated, and indicate that they are ignored. WDYT?

google/cloud/spanner/database_admin_client.h
```diff
@@ -392,12 +394,37 @@ class DatabaseAdminClient {
...
+   * case letters, numbers, underscore (`_`) or dashes (`-`), that is, they must
+   * belong to the `[a-z0-9_-]` character set.
+   *
+   * The @p expire_time must be at least 6 hours and at most 366 days from the
+   * time the `CreateBackup()` request is processed.
+   *
+   * @return A `google::cloud::future` that becomes satisfied when the operation
+   *   completes on the service. Note that this can take minutes in some cases.
+   */
+  ABSL_DEPRECATED("Use the Timestamp overload instead")
```
devjgm, line 423: My understanding of this attribute is that it will NOT produce a warning by default under clang for gcc, but it will cause a warning from clang-tidy? I believe there's also a compiler flag that if enabled will cause this to emit a warning. So the end result is that for most of our callers will this be silent and not break their builds. However, for customers using clang-tidy they'll start seeing this warning. Additionally, for any customers compiling with the "show deprecation warnings" flag enabled, their build might break (if using -Werror).

Does this sound right to you?

google/cloud/spanner/database_admin_connection.h
```diff
@@ -174,11 +175,11 @@ class DatabaseAdminConnection {
     /// The name of the database.
     Database database;
     std::string backup_id;
-    std::chrono::system_clock::time_point expire_time;
+    Timestamp expire_time;
```
devjgm, line 178: Terrible idea ahead: If `Timestamp` were _implicitly_ constructible from `std::chrono::system_clock::time_point`, then existing code like
```cc
std::chrono::system_clock::time_point tp;
auto p = CreateBackupParams{db, id, tp, {}};
```
Would continue to compile even if we changed the `expire_time` field to `Timestamp`. It's probably still a breaking change technically, but perhaps less so.

devjgm (commented):
Thanks, Brad. I just quickly looked at the .h files and have a couple questions for discussion.

google/cloud/spanner/database_admin_connection.h
devbww: Done.

(Aside: It only matters for `expire_time` ... `version_time` is a new field.)

google/cloud/spanner/database_admin_client.h
devbww: I do see warnings like ...
```
google/cloud/spanner/integration_tests/backup_integration_test.cc:146:30: warning: 'CreateBackup' is deprecated: Use the Timestamp overload instead [-Wdeprecated-declarations]
```
from clang (but I don't know how `deprecated-declarations` was enabled, in this case in ubsan, or why it might not be in other builds).

So, I guess I'll say that sounds right, and working as advertised.

google/cloud/spanner/database_admin_connection.h
devbww: We've taken the "leave the old fields" approach, so this is moot.  The possibility of the conversion failing would have been problematic anyway.

google/cloud/spanner/database_admin_client.h
devjgm: I couldn't find those warnings in the build logs. Which build? Do you have a link?

If those show up all the time, might it break users who compile with `-Werror`?

To be clear, my concern is that if adding a formal deprecation tag warning here may break users, we may want to consider whether we actually want to use that tag, or if we just want to document that an overload/field is going away. I don't think we've used `ABSL_DEPRECATED()` yet, and I'm afraid I don't yet really understand its implications. If you do, please feel free to just enlighten me so that we can feel confident using it now and going forward.

devjgm (commented):
Overall this LGTM.

However, I _think_ this is no longer a breaking change, though it does announce a deprecation and may deserve to be called out clearly in the next Release Notes. You may want to update this PR description to reflect that.

Also, in the past for changes like this we've sometimes included an edit to the `CHANGELOG.md` file in the affecting PR, so that the PR author can write clear notes to users about what's actually changing. Consider doing that.

devbww:
> However, I think this is no longer a breaking change, though it does announce a deprecation and may deserve to be called out clearly in the next Release Notes. You may want to update this PR description to reflect that.

I will update the PR description during the "squash and merge".

> Also, in the past for changes like this we've sometimes included an edit to the CHANGELOG.md file in the affecting PR, so that the PR author can write clear notes to users about what's actually changing. Consider doing that.

My plan was to defer that until merging into the main branch.  Updating `CHANGELOG.md` in this branch will just cause more conflicts.

devbww:
> I couldn't find those warnings in the build logs. Which build? Do you have a link?

They won't be in the build logs because we don't call the deprecated overloads.

> If those show up all the time, might it break users who compile with -Werror?

Yes, I imagine that under `-Wdeprecated-declarations` and `-Werror` that the build would break.  But it seems it would be what they asked for.

> To be clear, my concern is that if adding a formal deprecation tag warning here may break users, we may want to consider whether we actually want to use that tag, or if we just want to document that an overload/field is going away. I don't think we've used ABSL_DEPRECATED() yet, and I'm afraid I don't yet really understand its implications. If you do, please feel free to just enlighten me so that we can feel confident using it now and going forward.

I'm happy to take it away.  I don't claim to understand the full implications.  But if we choose to do that we should probably reopen [#2857](https://github.com/googleapis/google-cloud-cpp/issues/2857).

devjgm:
> > However, I think this is no longer a breaking change, though it does announce a deprecation and may deserve to be called out clearly in the next Release Notes. You may want to update this PR description to reflect that.
> 
> I will update the PR description during the "squash and merge".
> 
> > Also, in the past for changes like this we've sometimes included an edit to the CHANGELOG.md file in the affecting PR, so that the PR author can write clear notes to users about what's actually changing. Consider doing that.
> 
> My plan was to defer that until merging into the main branch. Updating `CHANGELOG.md` in this branch will just cause more conflicts.

I once again failed to realize that this PR was to a non-master branch. Your plan SGTM.

devjgm:
> > I couldn't find those warnings in the build logs. Which build? Do you have a link?
> 
> They won't be in the build logs because we don't call the deprecated overloads.

Ahh. Since we need to make sure those overloads continue working, I think it would be good if we had _some_ test that called them, even once.

> 
> > If those show up all the time, might it break users who compile with -Werror?
> 
> Yes, I imagine that under `-Wdeprecated-declarations` and `-Werror` that the build would break. But it seems it would be what they asked for.
> 
> > To be clear, my concern is that if adding a formal deprecation tag warning here may break users, we may want to consider whether we actually want to use that tag, or if we just want to document that an overload/field is going away. I don't think we've used ABSL_DEPRECATED() yet, and I'm afraid I don't yet really understand its implications. If you do, please feel free to just enlighten me so that we can feel confident using it now and going forward.
> 
> I'm happy to take it away. I don't claim to understand the full implications. But if we choose to do that we should probably reopen [#2857](https://github.com/googleapis/google-cloud-cpp/issues/2857).

We could reopen #2857 or create a new issue, but I think it would be good for us to understand the implications of `ABSL_DEPRECATED` on our customers before using it, because it has the ability to _break_ our customers' builds. And to be clear, I'm only saying that _I_ don't understand the implications.

devjgm (dismissed):

devbww:
> Since we need to make sure those overloads continue working, I think it would be good if we had some test that called them, even once.

OK, done.  PTAL.

This, of course, requires the removal of the deprecation attributes, lest our clang-tidy build fails.  (I also find it irksome when packages elicit their own deprecation warnings during builds.  I'm looking at you, gtest.)

devbww:
> We could reopen #2857 or create a new issue

Reopened.

devjgm (approved):

PR #5847, opened 2021-02-14
ci: improve vcpkg download and usage

### LINKED ISSUE #5818: Cleanup vcpkg commands on windows CI
In https://github.com/googleapis/google-cloud-cpp/pull/5816 we simplified the `vcpkg` commands a bit, and enabled caching of the `vcpkg` binary so we don't have to compile it each time. These changes were made to the Linux and macOS builds. We should make analogous changes to the Windows builds' uses of `vcpkg`.

Specifically likely these lines need to be updated (maybe more):

https://github.com/googleapis/google-cloud-cpp/blob/0b42bde91ec6af4aebe083758553d52a12526ef9/ci/kokoro/windows/build-cmake-dependencies.ps1#L59-L77

coryan:
Use a tarball (instead of `git clone`) to get the vcpkg ports data.
Download the vcpkg binary from GitHub to speed up the build. Use the
vcpkg binary cache instead of uploading the full `installed` directory,
take advantage of `gsutil rsync` for uploading and downloading the vcpkg
binary cache.

Fixes #5818

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5847)


devjgm (approved):
Thanks for this cleanup!

PR #5848, opened 2021-02-15
fix: use correct variable to disable abseil tests

remyabel:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5848)


coryan (approved):
Thanks!

/cc: @devjgm it is a bit sad that Abseil is special w.r.t. disabling tests.  The documentation actually recommends using `-DBUILD_TESTING=OFF`, but scanning the code that does not seem to be enough.

devjgm:
We should file an issue against abseil about this.

remyabel:
The tests are off by default anyway so this change is not needed after the bug is reported to abseil so I will close it.

PR #5860, opened 2021-02-16
feat: add support for get_pointer

devjgm:
This gives us the ability to mutate options directly within `Options`
class. This also gives us the ability to get options that will not move
so that we can return references to them.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5860)

### DIFF
--- a/google/cloud/options.h
+++ b/google/cloud/options.h
@@ -107,6 +107,16 @@ class Options {
   }
 
   /**
+   * Gets a pointer to the option specified by `T`, else nullptr.
+   */
+  template <typename T>
+  T* get_pointer() {
+    auto it = m_.find(typeid(T));
+    if (it == m_.end()) return nullptr;
+    return absl::any_cast<T>(&it->second);
+  }
+
+  /**
    * Gets the option of type `T` if set, else a newly constructed default `T`.
    *
    * If the specified option `T` is not set, a new `T` will be constructed with
--- a/google/cloud/options.h
+++ b/google/cloud/options.h
@@ -120,7 +120,17 @@ class Options {
   }
 
   /**
-   * Gets a pointer to the option specified by `T`, else nullptr.
+   * Gets a const pointer to the option specified by `T`, else nullptr.
+   */
+  template <typename T>
+  T const* get_pointer() const {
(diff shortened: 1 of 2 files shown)


devbww (approved):

devbww (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5860)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devjgm)

devjgm:
Thanks, Brad.

I'm actually thinking about adding another function to this interface

```cc
  /**
   * Gets a reference to the option specified by `T`, creating it if necessary.
   */
  template <typename T, typename... U>
  T& get_or_set(U&&... u) {
    auto it = m_.find(typeid(T));
    if (it != m_.end()) return *absl::any_cast<T>(&it->second);
    set<T>(std::forward<U>(u)...);
    return *get_pointer<T>();
  }
```

These functions in this PR showed up as I tried to use this class to re-implement `ConnectionOptions`. Functions like these may make the code cleaner and more efficient.

WDYT about these 3 new functions (`get_pointer` x2 and `get_or_set`)? I think I need to sleep on 'em and see what they look like tomorrow.

devjgm:
Nevermind. I'm not going to add that `get_or_set` function. I'll consider it later if it's really necessary. So no change from what you already reviewed.

coryan:
Do we think customers should be using the pointer accessors?  If not, maybe we should have a friend standalone function in `internal::` that we use in the implementation of more complicated classes based on `Options`?

devjgm:
Two thoughts:

1. Should? I don't know. Probably depends on their use case. But I think the pointer accessors are safe to be used.
2. Right after merging this, I thought of a different way to use `Options` in my upcoming PR where change `ConnectionOptions`, and I no longer need these functions. So in the spirit of err on the side of smaller interfaces, I thinking of rolling this PR back. Not because I think the change is bad, but more because it's easier to add things later than remove them later.

I assume there's no objection to a rollback?

Issue #5738, opened 2021-01-22
Implement "ConnectionOptions2"


devjgm:
(A quick late-Friday brain dump so we don't lose too much state.)

# Background

The current `ConnectionOptions` class is a template

https://github.com/googleapis/google-cloud-cpp/blob/a94d60293ec829a432bee142c4272b64aea40360/google/cloud/connection_options.h#L41-L43

Its template parameter is a traits type that defines three service customizable functions, for example

https://github.com/googleapis/google-cloud-cpp/blob/a94d60293ec829a432bee142c4272b64aea40360/google/cloud/pubsub/connection_options.h#L27-L36

# The issue

The problem that we're running into with the microgenerator is that no all services have a single `end_point`, some have multiple endpoints, for example, there may be separate end points for data and admin connections. For example, the Bigtable `ClientOptions` class:

https://github.com/googleapis/google-cloud-cpp/blob/a94d60293ec829a432bee142c4272b64aea40360/google/cloud/bigtable/client_options.h#L85-L92

@scotthart wrote a nice doc about this issue and outlined several options (internal only doc, sorry: go/cloud-cxx:connection-options2-choice)

# The solution

We discussed the options in Scott's doc above and as a team we decided on a solution that looks basically like this.

* We're going to create a _new_ "Connection Options" class. Until we settle on a final name, it should be named `internal::ConnectionOptions2`. This name is "internal", clearly not the final name, and it should be easy for us to find references to this name to rename later.
* `internal::ConnetionOptions2` will be a "bag of optionals" for all of its fields. That is, any option may be set, or not. The default state for all options will be unset (this matches optional's default). Note: Fields that have a native "unset" state do not need to be wrapped in an optional, for example, `std::shared_ptr<T>` could be unset with `nullptr`.
* Default values will be applied by the relevant "MakeConnection" function. For example, it would be the responsibility of `spanner::MakeConnection(..., internal::ConnectionOptions2)` to use the _default_ endpoint of `spanner.googleapis.com` if the relevant option is unset in the argument. **This is the point that fixes the issue above**. This way, the data connection and admin connections can use the same `internal::ConnectionOptions2` class, and yet they can apply their own defaults to  unset fields.
* What about the existing `ConnectionOptions` class? It will continue to exist for the current services that use it for backward compatibility. It will (probably) be re-implemented on top of the new `internal::ConnectionOptions2` class, but it will preserve its existing interface and semantics.
[...]

PR #5864, opened 2021-02-17
feat: ConnectionOptions<T> can convert to `internal::Options`

devjgm:
Part of https://github.com/googleapis/google-cloud-cpp/issues/5738

This give us and users a path to migrate from the old `ConnectionOptions<T>` instances so the new `google::cloud::Options` class. After this PR, we can start adding overloads to existing functions that take `ConnectionOptions<T>` to also take `Options`, and have the `Options` overload be preferred.

## Attention

Since this is the first PR where we're really _using_ the `Options` class, I'll make a few observations and ask for feedback.

1. I don't love that `opts.set<GrpcUserAgentPrefixOption>(std::set<std::string>{user_agent_prefix_});` requires the user to spell the `std::set` part. The reason for this is because `Options::set<T>()` takes a variadic forwarding reference, and passing _only_ an initializer list (e.g., `opts.set<GrpcUserAgentPrefixOption>({user_agent_prefix_});`) will not be able to deduce `U`. We _could_ "fix" this if `Options` takes a position that all option structs _must_ have one data member named `.value` (which we're currently agnostic about). In this case, we could use the type of the `.value` member to _be_ the type of set's argument, thus avoiding the forwarding reference, and thus enabling caller-side initializer lists. Thoughts?
2. `Options` only "gets" _values_, so that requires a copy. This is fine and easy, but it could become an issue if these options become large and get copied a lot. For example, consider an option whose value is a set/map, and the caller wants to insert a value. That would require making a copy, inserting the value, then moving the new value over the old one. Just something to consider.
3. The endpoint option is gRPC specific: `GrpcEndpointOption`. Is that fine? I think it's OK. If we make it transport agnostic, then we'll need to find another header for it to live in.
4. `GrpcUserAgentPrefixOption` has a _std::set_ for all the prefixes. This means that the user can't control the order of the prefixes. Is this OK with us? I think it's probably fine. I don't know if/why the order would matter. If we do care about the order, we could use a `std::deque`.

Note: I did not change the implementation of `ConnectionOptions<T>` to simply hold a single `google::cloud::Options` instance and re-implement its getters/setters in terms of `Options`, because there were enough little semantic differences that made it messy. For example, `channel_pool_domain()` is now implemented as a single key in `GrpcChannelArgumentsOption`, which makes it very difficult for the `channel_pool_domain()` getter to return a const-reference (to what)? I could stuff a default empty string in there for that option, but then an empty channel_pool_domain is used as a signal to not set that channel attribute.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864)


coryan (commented):
Need to leave, some comments below, but I did not complete the review.

> Reviewed 7 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-:-MTl20t-2DDIXsbNPMVt:b-i2fdhw)* status: 7 of 8 files reviewed, 4 unresolved discussions (waiting on @devjgm)

___
*[google/cloud/connection_options.h, line 60 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-MTl1-776RxS90eIRwNF:-MTl1-776RxS90eIRwNG:bcrl1bf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452c0b3bdde038a793175368a433e0853cbcffc9/google/cloud/connection_options.h#L60)):*
> ```objc
> 
>   /// Moves all the settings in this class into an `google::cloud::Options`.
>   Options ToOptions() && {
> ```

can we make this a friend in `internal::`?  We should try to not expose APIs that we need for our implementation in the public APIs of other classes.
___
*[google/cloud/connection_options.h, line 74 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-MTl18X4-cGETLlHBEWI:-MTl18X4-cGETLlHBEWJ:b-roh631) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452c0b3bdde038a793175368a433e0853cbcffc9/google/cloud/connection_options.h#L74)):*
> ```objc
>     if (!channel_pool_domain_.empty()) {
>       opts.set<GrpcChannelArgumentsOption>(std::map<std::string, std::string>{
>           {GRPC_ARG_CHANNEL_POOL_DOMAIN, std::move(channel_pool_domain_)}});
> ```

This macro does not exist in sufficiently old versions of gRPC, see the Debian buster build ...
___
*[google/cloud/grpc_options.h, line 67 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-MTl1hlNBO-GB3Kkvdlb:-MTl1hlNBO-GB3Kkvdlc:bqzadwo) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452c0b3bdde038a793175368a433e0853cbcffc9/google/cloud/grpc_options.h#L67)):*
> ```objc
>  * number of users running particular versions of their system or library.
>  */
> struct GrpcUserAgentPrefixOption {
> ```

User agent prefixes apply to both gRPC and REST clients.
___
*[google/cloud/grpc_options.h, line 96 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-MTl1o4TBoSb5Lttcijp:-MTl1o4TBoSb5Lttcijq:bttltn1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452c0b3bdde038a793175368a433e0853cbcffc9/google/cloud/grpc_options.h#L96)):*
> ```objc
>  * - rpc-streams
>  */
> struct GrpcTracingComponentsOption {
> ```

I think this can be generalized too, feel free to push these generalizations to a future PR, but be aware of the API-breakage that might cause.

coryan (commented):
Sorry for the incomplete review, I had to run to do an interview. Some suggestions below.  It might be easier to start with these options in the `internal::` namespace` to avoid breaking APIs as we stabilize the design for all libraries.

If you agree with these changes, but would prefer to do them in future PRs that works too.

> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-:-MTlHHZ10wNR9_5AHCGo:buug4id)* status: all files reviewed, 4 unresolved discussions (waiting on @devjgm)

___
*[google/cloud/grpc_options.h, line 67 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-MTl1hlNBO-GB3Kkvdlb:-MTlH3zhAH5GO8aF01Wg:b-i36k6o) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452c0b3bdde038a793175368a433e0853cbcffc9/google/cloud/grpc_options.h#L67)):*

User agent prefixes apply to both gRPC and REST clients.
</blockquote></details>

Consider refactoring to a common place.

devjgm (commented):
Done. Moved all the `Options`-related code into `internal::`.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-:-MTqciP_-k5zAvBjpzV8:b60oy2y)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @coryan)

___
*[google/cloud/connection_options.h, line 60 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-MTl1-776RxS90eIRwNF:-MTqcpP88gbidLkuDp_i:bawgmoz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452c0b3bdde038a793175368a433e0853cbcffc9/google/cloud/connection_options.h#L60)):*

can we make this a friend in `internal::`?  We should try to not expose APIs that we need for our implementation in the public APIs of other classes.
</blockquote></details>

Moved to a non-member in `internal::`. Note: That once we have finalized the `Options` API and class, I think we should consider moving this conversion function back into the class. I think a user-visible `ToOptions()` function would be useful for users so that they can easily migrate from the old `ConnectionOptions` -> the new `Options`.
___
*[google/cloud/connection_options.h, line 74 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-MTl18X4-cGETLlHBEWI:-MTqd7uS6WlRN5ILyJpW:b-ab1zf2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452c0b3bdde038a793175368a433e0853cbcffc9/google/cloud/connection_options.h#L74)):*

This macro does not exist in sufficiently old versions of gRPC, see the Debian buster build ...
</blockquote></details>

Thanks. Moved back to using the string and I kept the original comment that was there. I thought our gRPC requirements had moved forward enough that this macro would be available everywhere, so I wanted to check that. I guess not :-)
___
*[google/cloud/grpc_options.h, line 67 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-MTl1hlNBO-GB3Kkvdlb:-MTqdLiY-gsyCNvECB5H:bvvrisb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452c0b3bdde038a793175368a433e0853cbcffc9/google/cloud/grpc_options.h#L67)):*

Consider refactoring to a common place.
</blockquote></details>

Moved to a `common_options.h` header.
___
*[google/cloud/grpc_options.h, line 96 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-MTl1o4TBoSb5Lttcijp:-MTqdUll2nxysicuojcu:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/452c0b3bdde038a793175368a433e0853cbcffc9/google/cloud/grpc_options.h#L96)):*

I think this can be generalized too, feel free to push these generalizations to a future PR, but be aware of the API-breakage that might cause.
</blockquote></details>

Done.

coryan (approved):
:lgtm:

> Reviewed 12 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5864#-:-MTqgVnFCm7M-xAIdra0:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

google/cloud/internal/common_options.h
```diff
@@ -0,0 +1,68 @@
...
+#include <string>
+
+namespace google {
+namespace cloud {
+inline namespace GOOGLE_CLOUD_CPP_NS {
+
+/**
+ * Change the endpoint.
+ *
+ * In almost all cases the a suitable default will be chosen automatically.
```
devbww, line 30: s/ the//

google/cloud/internal/common_options.h
```diff
@@ -0,0 +1,68 @@
...
+ * Applications may need to be changed to (1) test against a fake or simulator,
+ * or (2) to use a beta or EAP version of the service.
```
devbww, line 32: "changed to (1) test ..., or (2) use a ..."

OR

"changed (1) to test ..., or (2) to use a ..."

google/cloud/internal/grpc_options.h
```diff
@@ -0,0 +1,97 @@
...
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_GRPC_OPTIONS_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_GRPC_OPTIONS_H
+
+#include "google/cloud/background_threads.h"
+#include "google/cloud/internal/options.h"
+#include "google/cloud/tracing_options.h"
+#include "google/cloud/version.h"
+#include <grpcpp/grpcpp.h>
+#include <map>
+#include <set>
```
devbww, line 24: Unused?

devjgm:
Addressed in follow-up PR https://github.com/googleapis/google-cloud-cpp/pull/5875

Issue #5588, opened 2020-12-08
CMEK keys must be colocated with Spanner Instances where they are used


mr-salty:
keys used to encrypt databases must be in the same location as the corresponding Spanner Instance.

However, `spanner_testing::PickRandomInstance` might return an Instance from any location, and there's currently no (easy) way to determine the location, so it's difficult to write tests/examples that use it.

mr-salty:
This will need to be resolved before we can merge CMEK into the main branch (#4312)

mr-salty:
unfortunately I didn't manage to knock this off, so reassigning to Brad.

Currently the key location is [hardcoded to us-central1](https://github.com/googleapis/google-cloud-cpp/blob/c4efc5cc5116f76b9da5e97a4fb00bbb31a71424/google/cloud/spanner/integration_tests/database_admin_integration_test.cc#L44) and there is a [hack](https://github.com/googleapis/google-cloud-cpp/blob/c4efc5cc5116f76b9da5e97a4fb00bbb31a71424/google/cloud/spanner/testing/pick_random_instance.cc#L46-L49) that forces the integration tests to only select instances that happen to be in us-central1. The hack excludes 6 of our 16 [test instances](http://go/xlwei1-spanner?project=cloud-cpp-testing-resources), including 2 that are in us-central1 and would otherwise be usable. While it would be possible to merge that hack into `master` in a pinch, it would potentially cause integration tests to slow down due to restricting the number of instances. It's also fragile, since it depends on specific named instances being in specific locations.

My tentative plan was to make `PickRandomInstance()` return the location, so we can use it when constructing the key name. I believe Carlos recently did something similar for samples, where we also [hardcode the location](https://github.com/googleapis/google-cloud-cpp/blob/c4efc5cc5116f76b9da5e97a4fb00bbb31a71424/google/cloud/spanner/samples/samples.cc#L3231-L3232) currently (that line could use a TODO).

Note that it's also necessary to manually create keys with the correct ring name and key name in each location via the [console](https://pantheon.corp.google.com/security/kms?project=cloud-cpp-testing-resources) or gcloud. As an alternative, we might auto-generate the KMS API and use it to create keys on the fly, or perhaps the build scripts could do so via gcloud.

[...]

PR #5868, opened 2021-02-18
refactor(spanner): reorganize backup_integration_test

devbww:
Refactor the `BackupTest` class, mostly in preparation for adding
new PITR and CMEK backup/restore tests.

Cleanup some handling of timestamps by using `absl::Time` for all
time computation, and `spanner::Timestamp` for all expectations.
Remove the assertion that "the server only preserves micros" as the
loss of precision is really due to funneling through `time_point`.

Also only run `CleanupStaleInstances()` once when the test starts,
rather than before every test case.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5868)

### DIFF
--- a/google/cloud/spanner/integration_tests/backup_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/backup_integration_test.cc
@@ -18,13 +18,14 @@
 #include "google/cloud/spanner/testing/cleanup_stale_instances.h"
 #include "google/cloud/spanner/testing/pick_instance_config.h"
 #include "google/cloud/spanner/testing/policies.h"
-#include "google/cloud/spanner/testing/random_backup_name.h"
 #include "google/cloud/spanner/testing/random_database_name.h"
 #include "google/cloud/spanner/testing/random_instance_name.h"
+#include "google/cloud/spanner/timestamp.h"
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/internal/random.h"
 #include "google/cloud/internal/time_utils.h"
 #include "google/cloud/testing_util/assert_ok.h"
+#include "absl/time/time.h"
 #include <gmock/gmock.h>
 #include <chrono>
 #include <regex>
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/integration_tests/backup_integration_test.cc
```diff
@@ -36,116 +37,115 @@ namespace spanner {
...
+    // Make sure we're using a correct regex.
+    EXPECT_EQ(2, instance_name_regex.mark_count());
+    auto generator = google::cloud::internal::MakeDefaultPRNG();
+    Instance in(ProjectId(), spanner_testing::RandomInstanceName(generator));
+    auto fq_instance_name = in.FullName();
+    std::smatch m;
+    EXPECT_TRUE(std::regex_match(fq_instance_name, m, instance_name_regex));
+    EXPECT_EQ(3, m.size());
+
+    if (RunSlowBackupTests()) {
```
coryan, line 77: You can probably perform this cleanup regardless of whether slow tests are enabled or not, up to you.

google/cloud/spanner/integration_tests/backup_integration_test.cc
```diff
@@ -36,116 +37,115 @@ namespace spanner {
...
+      spanner_testing::TestBackoffPolicy(),
+      spanner_testing::TestPollingPolicy()));
 
-  Database db(project_id_, instance_id, database_id);
-
-  auto database_future = database_admin_client_.CreateDatabase(db);
-  auto database = database_future.get();
+  Instance in(ProjectId(), spanner_testing::RandomInstanceName(generator_));
+  auto instance_config = spanner_testing::PickInstanceConfig(
+      in.project_id(), std::regex(".*us-west.*"), generator_);
```
coryan, line 114: Maybe a comment to document why `us-west` is special?  I have some vague recollection that we have more room to create instances there or something.

google/cloud/spanner/integration_tests/backup_integration_test.cc
```diff
@@ -36,116 +37,115 @@ namespace spanner {
...
+  ASSERT_FALSE(instance_config.empty()) << "could not get an instance config";
+  EXPECT_STATUS_OK(
+      instance_admin_client
+          .CreateInstance(CreateInstanceRequestBuilder(in, instance_config)
```
coryan, line 118: Do you know why this test needs a new instance?  I mean, it is going to create databases and backups with random names, an existing instance should be able to do that?  Or is there some limitation on the number of backups per instance we need to workaround?

coryan (approved):
LGTM. There is some stuff in this test that is failure obscure, maybe I just don't remember why we made the decisions we made anymore.  Consider some comments (left some ideas below).

google/cloud/spanner/integration_tests/backup_integration_test.cc
devbww: I'll leave it as is for now, but will reassess when I return here soon for the PITR/CMEK tests.

google/cloud/spanner/integration_tests/backup_integration_test.cc
devbww: https://github.com/googleapis/google-cloud-cpp-spanner/pull/1403 appears to be the origin, but doesn't list a reason.

google/cloud/spanner/integration_tests/backup_integration_test.cc
devbww: No, I don't know.  My thought is there was no better way to ensure a "us-west" instance (for whatever reason that was).

I am working on a filtering capability for `PickRandomInstance()`, which will be required for the CMEK merge (at least for the tests in their current state).  See #5588.  At that time I can also simplify this (if we believe "us-west" is still a feature).

PR #5874, opened 2021-02-19
cleanup(spanner): tidy up spanner backup/restore samples

devbww:
This is mostly to ease the upcoming merge of backup/restore
samples for PITR and CMEK, which include some general fixes
like eliminating unused using declarations, adding missing
newlines, and correcting tag names.  Make those changes here
first so that the future merges are all about the features.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5874)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -671,8 +671,6 @@ void CreateBackup(google::cloud::spanner::DatabaseAdminClient client,
                   std::string const& project_id, std::string const& instance_id,
                   std::string const& database_id,
                   std::string const& backup_id) {
-  using ::google::cloud::future;
-  using ::google::cloud::StatusOr;
   google::cloud::spanner::Database database(project_id, instance_id,
                                             database_id);
   auto backup = client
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -1779,7 +1770,7 @@ void QueryNewColumn(google::cloud::spanner::Client client) {
     std::cout << "AlbumId: " << std::get<1>(*row) << "\t";
     auto marketing_budget = std::get<2>(*row);
     if (marketing_budget) {
-      std::cout << "MarketingBudget: " << marketing_budget.value() << "\n";
+      std::cout << "MarketingBudget: " << *marketing_budget << "\n";
```
coryan, line 1773: nit: unrelated?

coryan (approved):

google/cloud/spanner/samples/samples.cc
devbww: Yes.  And no ... I should have mentioned in the description an attempt to make the samples more `StatusOr`-idiomatic.  The underlying issue is that I probably should not have attempted any cleanup in a feature branch, but it is difficult to resist.

PR #5882, opened 2021-02-20
ci: prefer curl which is avail by default more often

devjgm:
Curl seems to be more commonly available on default installs of
operating systems, including macOS, so it's nice to prefer curl rather
than wget so we can avoid the user having to install wget when we could
have used curl.

Note: most of our CI scripts were already depending on curl, so this
shouldn't introduce a new dep, rather I think it should remove a dep.

Also, I'm not using the "--silent" option because I think the download progress info may be useful. But if others disagree and think this should be silent, I can change to that.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5882)


coryan (approved):
Thanks.

PR #5885, opened 2021-02-21
test(spanner): support filtering of `PickRandomInstance()` candidates

### LINKED ISSUE #5588: CMEK keys must be colocated with Spanner Instances where they are used

devbww:
Support filtering of `PickRandomInstance()` candidates (1) on the server
side, by exposing the `ListInstances` filter, and (2) on the client side,
via a predicate over `Instance` and `InstanceConfig`.

This will be needed to support the upcoming CMEK integration tests (at
least in their current configuration), where the selected instance must
be in the single location we use to store the encryption keys.

Also use server-side filtering to implement the restriction of instance
IDs to "test-instance-*".

Fixes #5588.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885)

### DIFF
--- a/google/cloud/spanner/testing/pick_random_instance.cc
+++ b/google/cloud/spanner/testing/pick_random_instance.cc
@@ -26,19 +26,22 @@ inline namespace SPANNER_CLIENT_NS {
 StatusOr<std::string> PickRandomInstance(
     google::cloud::internal::DefaultPRNG& generator,
     std::string const& project_id, std::string const& filter,
-    std::function<bool(
-        google::spanner::admin::instance::v1::Instance const&,
-        google::spanner::admin::instance::v1::InstanceConfig const&)> const&
-        predicate) {
+    InstancePredicate predicate) {
+  if (!predicate) {
+    predicate =
+        [](google::spanner::admin::instance::v1::Instance const&,
+           google::spanner::admin::instance::v1::InstanceConfig const&) {
+          return true;
+        };
+  }
+
   spanner::InstanceAdminClient client(spanner::MakeInstanceAdminConnection());
 
+  // Fetch the instance configurations for use in the predicate.
   std::vector<google::spanner::admin::instance::v1::InstanceConfig> configs;
-  if (predicate) {
-    // Fetch the instance configurations for use in the callback.
-    for (auto& config : client.ListInstanceConfigs(project_id)) {
-      if (!config) return std::move(config).status();
-      configs.push_back(*std::move(config));
-    }
+  for (auto& config : client.ListInstanceConfigs(project_id)) {
+    if (!config) return std::move(config).status();
+    configs.push_back(*std::move(config));
   }
 
   // We only pick instance IDs starting with "test-instance-" for isolation
@@ -51,17 +54,13 @@ StatusOr<std::string> PickRandomInstance(
   for (auto& instance :
        client.ListInstances(project_id, filter + name_filter)) {
     if (!instance) return std::move(instance).status();
-    auto instance_id = instance->name().substr(instance_prefix.size());
-    if (!predicate) {
-      instance_ids.push_back(instance_id);
-    } else {
-      for (auto const& config : configs) {
-        if (instance->config() == config.name()) {
-          if (predicate(*instance, config)) {
-            instance_ids.push_back(instance_id);
-          }
-          break;
+    for (auto const& config : configs) {
+      if (instance->config() == config.name()) {
+        if (predicate(*instance, config)) {
+          auto instance_id = instance->name().substr(instance_prefix.size());
+          instance_ids.push_back(std::move(instance_id));
         }
+        break;
       }
     }
   }
--- a/google/cloud/spanner/testing/pick_random_instance.cc
+++ b/google/cloud/spanner/testing/pick_random_instance.cc
@@ -13,37 +13,56 @@
 // limitations under the License.
 
 #include "google/cloud/spanner/testing/pick_random_instance.h"
-#include "google/cloud/spanner/connection_options.h"
 #include "google/cloud/spanner/create_instance_request_builder.h"
 #include "google/cloud/spanner/instance_admin_client.h"
 #include "google/cloud/internal/getenv.h"
+#include <vector>
 
 namespace google {
 namespace cloud {
 namespace spanner_testing {
 inline namespace SPANNER_CLIENT_NS {
+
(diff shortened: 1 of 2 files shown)


coryan (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885)* status: all files reviewed, 2 unresolved discussions (waiting on @devbww)

___
*[google/cloud/spanner/testing/pick_random_instance.h, line 39 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885#-MU3oThl6hzQJ5Fbd-U5:-MU3oThl6hzQJ5Fbd-U6:bflb44m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c548816511b6707b598e890dd6af28a41b77130/google/cloud/spanner/testing/pick_random_instance.h#L39)):*
> ```objc
>     google::cloud::internal::DefaultPRNG& generator,
>     std::string const& project_id, std::string const& filter = "",
>     std::function<bool(
> ```

Consider a type alias, it was hard for me to find the `predicate` in this case.
___
*[google/cloud/spanner/testing/pick_random_instance.cc, line 36 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885#-MU3oyCFEcZV0b32IWoH:-MU3oyCFEcZV0b32IWoI:b-vo1r04) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c548816511b6707b598e890dd6af28a41b77130/google/cloud/spanner/testing/pick_random_instance.cc#L36)):*
> ```cpp
> 
>   std::vector<google::spanner::admin::instance::v1::InstanceConfig> configs;
>   if (predicate) {
> ```

I think if you take `predicate` by value you can do something like:

```cc
if (!predicate) {
  predicate = [](auto, auto) { return true; }
}
```

And the code below gets simpler.  Well, you can't use `auto` in the lambda, because C++11, but you get the idea.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/spanner/testing/pick_random_instance.h, line 39 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885#-MU3oThl6hzQJ5Fbd-U5:-MU4qyie9fRK97QI08pm:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c548816511b6707b598e890dd6af28a41b77130/google/cloud/spanner/testing/pick_random_instance.h#L39)):*

Consider a type alias, it was hard for me to find the `predicate` in this case.
</blockquote></details>

Done.
___
*[google/cloud/spanner/testing/pick_random_instance.cc, line 36 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885#-MU3oyCFEcZV0b32IWoH:-MU4r1TXCnQDDd_HnHxM:b-qbrc30) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c548816511b6707b598e890dd6af28a41b77130/google/cloud/spanner/testing/pick_random_instance.cc#L36)):*

I think if you take `predicate` by value you can do something like:

```cc
if (!predicate) {
  predicate = [](auto, auto) { return true; }
}
```

And the code below gets simpler.  Well, you can't use `auto` in the lambda, because C++11, but you get the idea. 

</blockquote></details>

I was trying to avoid (1) copying the `std::function`, and (2) fetching the instance configs every time, but this is test code, so done.

coryan (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885)* status: all files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[google/cloud/spanner/testing/pick_random_instance.cc, line 36 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885#-MU3oyCFEcZV0b32IWoH:-MU4vqcpEv7XC-18xosQ:biebb7i) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c548816511b6707b598e890dd6af28a41b77130/google/cloud/spanner/testing/pick_random_instance.cc#L36)):*

I was trying to avoid (1) copying the `std::function`, and (2) fetching the instance configs every time, but this is test code, so done.
</blockquote></details>

Ack.  Thanks for making the change.
___
*[google/cloud/spanner/testing/pick_random_instance.cc, line 58 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885#-MU4vyht4Z3dwzWjRsIx:-MU4vyht4Z3dwzWjRsIy:b7pz1xx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/397c6864f9476e5a8c845c5eae955ce72a3f2b32/google/cloud/spanner/testing/pick_random_instance.cc#L58)):*
> ```cpp
>     if (!instance) return std::move(instance).status();
>     for (auto const& config : configs) {
>       if (instance->config() == config.name()) {
> ```

nit: if you wanted to, you can organize the configs into some kind of map or hash as they are loaded and avoid the double loop.  But as you said elsewhere, this is test code.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devbww)

___
*[google/cloud/spanner/testing/pick_random_instance.cc, line 58 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5885#-MU4vyht4Z3dwzWjRsIx:-MU59H5y3Fjca0sidz7g:b-3bdnog) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/397c6864f9476e5a8c845c5eae955ce72a3f2b32/google/cloud/spanner/testing/pick_random_instance.cc#L58)):*

nit: if you wanted to, you can organize the configs into some kind of map or hash as they are loaded and avoid the double loop.  But as you said elsewhere, this is test code.
</blockquote></details>

Thanks.  But yes ... test code.  And I think we might also find that linear search outperforms associative lookup when the number of configs is small (as expected), and particularly so when it is within another loop and cache effects kick in when data is more contiguous.

PR #5891, opened 2021-02-22
cleanup: make CMake builds less verbose

coryan:
I made a number of changes to quiet up the builds, including:

- stop cspell from listing all the files in the repository
- stop `cmake --install` from listing all the installed files
- do not dump the clang-tidy configuration
- avoid logging actions that will not happen

I tried enabling `ctest --quiet`, but it does not print failures in that
case, we need something more involved, like logging to a file and
printing the file if there is an error.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5891)

### DIFF
--- a/ci/check-style.sh
+++ b/ci/check-style.sh
@@ -146,7 +146,7 @@ time {
 printf "%-30s" "Running cspell:"
 time {
   if ! git ls-files -z | grep -zE '\.(cc|h)$' |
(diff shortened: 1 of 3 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5891)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5895, opened 2021-02-23
cleanup(spanner): cleanup instance_admin_integration_test

devbww:
Carry over similar changes from backup_integration_test.  In
particular, only run `CleanupStaleInstances()` once, when the
test starts, rather than before every test case.  But follow
a previous suggestion to do so unconditionally, instead of
only when "slow" tests are enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5895)

### DIFF
--- a/google/cloud/spanner/integration_tests/instance_admin_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/instance_admin_integration_test.cc
@@ -23,10 +23,10 @@
 #include "google/cloud/testing_util/assert_ok.h"
 #include "google/cloud/testing_util/status_matchers.h"
 #include <gmock/gmock.h>
-#include <chrono>
-#include <ctime>
+#include <algorithm>
 #include <regex>
 #include <string>
+#include <vector>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/integration_tests/instance_admin_integration_test.cc
```diff
@@ -39,88 +39,100 @@ using ::testing::AnyOf;
...
+::testing::Environment* const kCleanupEnv =
+    ::testing::AddGlobalTestEnvironment(new CleanupStaleInstances);
+
 class InstanceAdminClientTest : public testing::Test {
  public:
-  InstanceAdminClientTest() : client_(MakeInstanceAdminConnection()) {}
+  InstanceAdminClientTest()
+      : generator_(internal::MakeDefaultPRNG()),
+        client_(MakeInstanceAdminConnection()) {
+    static_cast<void>(kCleanupEnv);
```
coryan, line 97: So arguably we could do these with:

```cc
  static int const kCleanup = [] {
    // cleanup code here ...
  }();
```

coryan (approved):
Sorry, maybe I should not have marked this ready for review. That looked like the "review here" button to me (I blame the lack of coffee).

google/cloud/spanner/integration_tests/instance_admin_integration_test.cc
devbww: Arguably, yes (modulo any difference between running something when the first test is being constructed -v- before any test is constructed).

It looks like we have one other use of `testing::Environment`, which is doing what we handle here with the `static env_var` things here.  I'd be happy to try to eliminate both `testing::Environment` uses separately if you like.

PR #5896, opened 2021-02-23
fix(generator): correct exit code and update hashes

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5896)


ci/check-golden-md5-hashes.sh
```diff
@@ -36,6 +36,7 @@ hash_check_result=$(md5sum --check --quiet "${PROJECT_ROOT}/ci/etc/generator-gol
 
 if [[ ${hash_check_result} != "" ]]; then
   io::log_yellow "Run 'ci/kokoro/docker/build.sh generate-libraries' to update."
+  exit 1
```
devjgm, line 39: FYI, it may be helpful here to also output the non-empty `${hash_check_result}` string so the user can see which hashes didn't match.

devjgm (approved):

coryan (approved):

PR #5897, opened 2021-02-23
generator(feature): default to nonidempotent for rpcs missing http option

scotthart:
During the survey of generating code for all grpc services, it was found that not all rpcs have the google.api.http option. This is used to determine default idempotency for each rpc. When the option is not present, prefer a conservative kNonIdempotent value.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5897)


coryan (approved):
In addition to this change, should we submit bugs for the protos missing this option?

scotthart (commented):
It's not clear to me that the option is required. Rpcs, that I have found, lacking the option include some commented as experimental and others that were bidirectional streaming. https://google.aip.dev/client-libraries/4222

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5897#-:-MUEV65U4f17HkhAt_dj:b-wpu65h)* status: 0 of 2 files reviewed, all discussions resolved

PR #5900, opened 2021-02-23
cleanup: internal/*_options.h were not in the internal namespace

devjgm:
These options were in the `google/cloud/internal/` directory, but they were accidentally not in the `google::cloud::internal` namespace. Fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5900)

### DIFF
--- a/google/cloud/internal/grpc_options.h
+++ b/google/cloud/internal/grpc_options.h
@@ -27,6 +27,8 @@ namespace google {
 namespace cloud {
 inline namespace GOOGLE_CLOUD_CPP_NS {
 
+namespace internal {
+
 /**
  * The gRPC credentials used by clients configured with this object.
  */
@@ -81,6 +83,8 @@ using BackgroundThreadsFactory =
 struct GrpcBackgroundThreadsOption {
   BackgroundThreadsFactory value;
 };
+
+}  // namespace internal
 
 namespace internal {
 
--- a/google/cloud/connection_options_test.cc
+++ b/google/cloud/connection_options_test.cc
@@ -48,26 +48,27 @@ TEST(ConnectionOptionsTest, Credentials) {
   auto expected = grpc::InsecureChannelCredentials();
   TestConnectionOptions options(expected);
   EXPECT_EQ(expected.get(), options.credentials().get());
-  EXPECT_EQ(expected, ToOptions(options).get<GrpcCredentialOption>()->value);
+  EXPECT_EQ(expected,
+            ToOptions(options).get<internal::GrpcCredentialOption>()->value);
 
   auto other_credentials = grpc::InsecureChannelCredentials();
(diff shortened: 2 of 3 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5900)* status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)

___
*[google/cloud/internal/common_options.h, line 26 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5900#-MUEeqmN2KO8mhWbIQtO:-MUEeqmN2KO8mhWbIQtP:b20wi3r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9eb8eac593f014bc7c6a106b5a97af391aedb1c8/google/cloud/internal/common_options.h#L26)):*
> ```objc
> namespace cloud {
> inline namespace GOOGLE_CLOUD_CPP_NS {
> ```

nit: are the blank lines between the namespaces intentional?

devjgm (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5900)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @scotthart)

___
*[google/cloud/internal/common_options.h, line 26 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5900#-MUEeqmN2KO8mhWbIQtO:-MUEfxxh1oiRVnBHESFB:b-90clyh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9eb8eac593f014bc7c6a106b5a97af391aedb1c8/google/cloud/internal/common_options.h#L26)):*

nit: are the blank lines between the namespaces intentional?
</blockquote></details>

Yes, they were intentional, but now that you bring it up, they probably shouldn't have been. I removed the blanks from common_options.h, because the internal namespace covered the whole file. However, I left the blanks in grpc_options.h because it only covered part of the file and the leading blank line matched the trailing blank line.

google/cloud/internal/grpc_options.h
```diff
@@ -82,6 +84,8 @@ struct GrpcBackgroundThreadsOption {
   BackgroundThreadsFactory value;
 };
 
+}  // namespace internal
+
 namespace internal {
```
devbww, line 89: Is it strange to close the `internal` namespace, and then immediately re-open it?

devjgm: A little strange, yes. The reason I did that is because the `internal` namespace being opened on line 89 should always remain internal, whereas the `internal` namespace closing on line 87 will be removed when these options move out of the `internal` namespace. I wanted to distinguish between the "temporarily" internal namespace and the permanent internal namespace so that we didn't accidentally publicly expose something.

PR #5908, opened 2021-02-24
cleanup: reenable DeleteResumableUpload test

remyabel:
Per #5030 it's already been fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5908)

### DIFF
--- a/google/cloud/storage/tests/object_integration_test.cc
+++ b/google/cloud/storage/tests/object_integration_test.cc
@@ -898,10 +898,6 @@ TEST_F(ObjectIntegrationTest, DeleteAccessControlFailure) {
 }
 
 TEST_F(ObjectIntegrationTest, DeleteResumableUpload) {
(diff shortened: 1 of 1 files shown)


coryan:
Looks like the gRPC plugin is still missing this feature:

https://github.com/googleapis/google-cloud-cpp/blob/82fa1e1b1adcc7119de6ba7c511f6f2c719608aa/google/cloud/storage/internal/grpc_client.cc#L368-L371

Let me know how you want to proceed.

remyabel:
I mistakenly thought the test was passing when I ran it locally (apparently I didn't read it hard enough). I think the original issue or new issue should be opened.

coryan:
> I think the original issue or new issue should be opened.

I agree, created #5908 and I am sending a PR to reference the right issue from the TODO.

Issue #5923, opened 2021-02-25
Remove the async functions from the Bigtable admin APIs


coryan:
In 2021-06-01 (or shortly after) we are planning to remove all the functions matching `Async*` in `bigtable::TableAdmin` and `bigtable::InstanceAdmin`. These functions were already marked as **experimental**, they allow applications to perform administrative operations in a non-blocking manner.

We do not think these functions are sufficiently useful to justify the additional maintenance and API clutter. Starting with the v1.25.0 release,  and depending on your compiler settings, using these functions may issue a deprecation warning at the call site.

If your compiler is emitting a diagnostic, consider disabling the warning at the call site, most compilers support `#pragmas` for this purpose.  If it is not possible to disable the warning at each call site, and you do need to disable the warnings, consider adding `-DGOOGLE_CLOUD_CPP_DISABLE_pragma to

The data APIs remain experimental, we know we want to remove the `CompletionQueue` parameter but that is a separate issue (see #2567)

PR #5924, opened 2021-02-25
cleanup(bigtable): deprecate admin async functions

### LINKED ISSUE #2857: Use [[deprecated]] attributes.

coryan:
All the `Async*` functions in bigtable are marked as experimental. We do
not think the admin APIs need async versions, so it is time to mark them
as deprecated.

I wanted to use `ABSL_DEPRECATED()`, but that macro only works for
Clang, and we have users on GCC and MSVC to worry about.

Part of the work for #5923. Fixes #2857 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5924)


devjgm (approved):

devjgm (approved):

devbww:
> I wanted to use ABSL_DEPRECATED(), but that macro only works for
> Clang, and we have users on GCC and MSVC to worry about.

A rhetorical question, but doesn't Abseil have GCC and MSVC users to worry about too?

coryan:
> > I wanted to use ABSL_DEPRECATED(), but that macro only works for
> > Clang, and we have users on GCC and MSVC to worry about.
> 
> A rhetorical question, but doesn't Abseil have GCC and MSVC users to worry about too?

Yes, and I think we should contribute the support for other compilers.  I did not want to wait for a full cycle of Abseil + gRPC releases before starting to use it.  We can replace the macro when they are ready.

/FYI @derekmauro

google/cloud/internal/attributes.h
```diff
@@ -0,0 +1,52 @@
...
+ */
+#if defined(GOOGLE_CLOUD_CPP_DEPRECATED)
+#error "Applications should not define GOOGLE_CLOUD_CPP_DEPRECATED"
+#elif GOOGLE_CLOUD_CPP_DISABLE_DEPRECATION_WARNINGS
+// Do nothing, the default definition effectively
+#elif GOOGLE_CLOUD_CPP_CPP_VERSION >= 201402L
+#define GOOGLE_CLOUD_CPP_DEPRECATED(message) [[deprecated(message)]]
+#elif defined(_MSC_VER)
+#define GOOGLE_CLOUD_CPP_DEPRECATED(message) __declspec(deprecated(message))
+#elif defined(__clag__) || defined(__GNUC__)
```
derekmauro, line 43: I think you mean `__clang__`.

derekmauro (commented):
I didn't realize our implementation of ABSL_DEPRECATED was stale. Thanks for letting me know, I'll look into updating it.

google/cloud/internal/attributes.h
coryan: Fixed, thanks!

PR #5925, opened 2021-02-25
feat: implement storage::GrpcClient::SetBucketIamPolicy

remyabel:
Fixes: #4160

I added the version field to SetBucketIamPolicyRequest but I think I may be confusing google::iam::v1::Policy, SetBucketIamPolicyRequest and google::storage::v1::SetIamPolicyRequest. 

I removed the test from grpc_client_failures_test.cc as I'm not entirely sure what the purpose of the test is, I am happy to add it back with a proper test if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5925)

### DIFF
--- a/google/cloud/storage/emulator/grpc_server.py
+++ b/google/cloud/storage/emulator/grpc_server.py
@@ -144,6 +144,11 @@ class StorageServicer(storage_pb2_grpc.StorageServicer):
         bucket.delete_notification(notification_id, context)
         return Empty()
 
+    def SetIamPolicy(self, request, context):
+        bucket_name = request.bucket
+        bucket = db.get_bucket(request, bucket_name, context)
+        return bucket.set_iam_policy(request, context)
+
     # === OBJECT === #
 
     def InsertObject(self, request_iterator, context):
--- a/google/cloud/storage/internal/bucket_requests.cc
+++ b/google/cloud/storage/internal/bucket_requests.cc
@@ -296,10 +296,9 @@ StatusOr<IamPolicy> ParseIamPolicyFromString(std::string const& payload) {
 SetBucketIamPolicyRequest::SetBucketIamPolicyRequest(
     std::string bucket_name, google::cloud::IamPolicy const& policy)
     : bucket_name_(std::move(bucket_name)) {
-  nlohmann::json iam{
-      {"kind", "storage#policy"},
-      {"etag", policy.etag},
-  };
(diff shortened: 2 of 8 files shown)


remyabel:
I forgot to add an integration test. I will reopen after that.

coryan:
Cool, the failure tests are there to exercise the "cannot reach the server" code path.  I think after this you can change them to only accept `kUnavailable`.

remyabel:
I've disabled the failures test for now. I think passing an empty `IamPolicy` is resulting in this error:

```
C++ exception with description "[json.exception.type_error.306] cannot use value() with number" thrown in the test body.
```

But I haven't been able to identify the problem.

coryan:
> I've disabled the failures test for now. I think passing an empty `IamPolicy` is resulting in this error:
> 
> ```
> C++ exception with description "[json.exception.type_error.306] cannot use value() with number" thrown in the test body.
> ```
> 

Sometimes I really wish the C++ runtime printed a stack trace for these exceptions that leak everywhere.

> But I haven't been able to identify the problem.

Can gdb and/or lldb give you a stack trace where the exception is thrown?

remyabel:
gdb lead me around in circles, since debugging within gtest is difficult. However, I printed the `payload_policy.status()` (not sure why I didn't think of this earlier as I used it to find a previous bug):

```
DEBUG: Invalid IamPolicy payload, expected array for 'bindings' field. payload={"bindings":null,"etag":"","kind":"storage#policy","version":0} [INVALID_ARGUMENT]
```

In other words, `NativeIamPolicy::CreateFromJson` doesn't like a default constructed `IamPolicy`. How would you like me to proceed?

coryan:
Seems like we need to fix the `NativeIamPolicy::CreateFromJson` function, a default-constructed `IamPolicy` should be very rare in the wild, but the library should not crash if it sees one.

remyabel:
Sure, just to be 100% clear, `NativeIamPolicy::CreateFromJson` has error checking to make sure the json has valid values. So the new behavior is that it should be able to accept a null `bindings`?

remyabel:
Actually, I think the larger issue here is that I'm mixing the `SetBucketIamPolicy` and the `SetNativeBucketIamPolicy` API's up. Trying to mix the two is clearly causing issues. I think I will close this and implement `SetNativeBucketIamPolicy` instead.

PR #5926, opened 2021-02-25
test(spanner): extend samples and integration-test timeouts

devbww:
We now run the GOOGLE_CLOUD_CPP_SPANNER_SLOW_INTEGRATION_TESTS
overnight, with a couple of new ones either recently submitted
or in the works, so extend the timeout.

The change from 15m to 60m could cause a failure delay if you
hit a hang without even running the slow tests, but so be it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5926)

### DIFF
--- a/google/cloud/spanner/integration_tests/BUILD
+++ b/google/cloud/spanner/integration_tests/BUILD
@@ -20,7 +20,7 @@ load(":spanner_client_integration_tests.bzl", "spanner_client_integration_tests"
 
 [cc_test(
     name = test.replace("/", "_").replace(".cc", ""),
-    timeout = "long",
+    timeout = "eternal",
     srcs = [test],
     tags = [
         "integration-test",
--- a/google/cloud/spanner/samples/BUILD
+++ b/google/cloud/spanner/samples/BUILD
@@ -23,7 +23,7 @@ load(":spanner_client_unit_samples.bzl", "spanner_client_unit_samples")
 
 [cc_test(
     name = test.replace("/", "_").replace(".cc", ""),
-    timeout = "long",
+    timeout = "eternal",
     srcs = [test],
     tags = [
         "integration-test",


coryan:
Is there any way to set these *just* for the nightly builds?  I think you can change the defaults using `bazel test --test_timeout`, maybe we can pass those in the build script?

Also, do you want to adjust this number?

https://github.com/googleapis/google-cloud-cpp/blob/8211af48121998cd0e444099208abd575669bd03/ci/kokoro/docker/common.cfg#L17

devbww (commented):
> Is there any way to set these just for the nightly builds?

Not within a `BUILD` file, I don't believe.

> I think you can change the defaults using bazel test --test_timeout, maybe we can pass those in the build script?

Yes, that would be possible.  At the moment I don't think it is worth it, but could easily be convinced otherwise.

> Also, do you want to adjust this number?

Not that I'm aware of, but perhaps we'll find out.

For what it's worth, I've been running with these "eternal" settings while developing the new tests/samples.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5926#-:-MUQ0OfzEMxUZkZ2S4PJ:b-58o3l1)* status: 0 of 2 files reviewed, all discussions resolved

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5926)* status: :shipit: complete! all files reviewed, all discussions resolved

coryan (approved):
> > I think you can change the defaults using bazel test --test_timeout, maybe we can pass those in the build script?
> 
> Yes, that would be possible.  At the moment I don't think it is worth it, but could easily be convinced otherwise.
> 

Let's try it out your way.

> > Also, do you want to adjust this number?
> 
> Not that I'm aware of, but perhaps we'll find out.

That is the total timeout for the build. Kokoro will kill the build after 120 minutes, including all the download time, compiling all the code, and then running all the tests.  If you think the tests and samples are close to 60m each, you are cutting it pretty close.

> For what it's worth, I've been running with these "eternal" settings while developing the new tests/samples.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5926#-:-MUQ1Wb86XVIb2nqHlwA:bawnfkv)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5932, opened 2021-02-26
doc(ADR): deprecated functions do not have examples

coryan:
Let's discuss in this PR, and if we agree I will merge (after changing dates).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5932)

### DIFF
--- a/doc/adr/2021-03-01-deprecated-functions-do-not-have-examples.md
+++ b/doc/adr/2021-03-01-deprecated-functions-do-not-have-examples.md
@@ -1,6 +1,6 @@
 **Title**: Deprecated functions do not have examples.
 
-**Status**: proposed
+**Status**: accepted
 
 **Context**: we write examples for all important functions, typically the ones
 that result in one or more RPCs. From time to time we need to deprecate a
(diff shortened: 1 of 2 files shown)


remyabel:
This is technically a separate idea but related, but I wonder if this will cause bug reports asking where the missing examples are. They may have bookmarks or links that are directed towards the `latest` documentation and it will create a 404. We could have the doxygen 404 page point towards an errata (or in this case the `CHANGELOG` will be enough) with something like:

```
Trying to find something that's no longer here? Please see the `CHANGELOG` to see if we removed it.
```

devjgm (approved):
This LGTM. I think it's _good_ that our examples show users the *right*/recommended way to do something, and that's never the deprecated way.

But for posterity, and for other `google-cloud-cpp` developers, I do think deprecated methods should still be *tested* in our unit tests. This is to ensure that they continue to function even during the period when they're deprecated. But I understand that this is different from the point of this ADR, which LGTM.

Thanks.

coryan:
> But for posterity, and for other `google-cloud-cpp` developers, I do think deprecated methods should still be _tested_ in our unit tests. This is to ensure that they continue to function even during the period when they're deprecated. But I understand that this is different from the point of this ADR, which LGTM.

Good point, added that to the text.

PR #5937, opened 2021-02-27
feat(storage): implement {Get,Set}NativeBucketIamPolicy

remyabel:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5937)

### DIFF
--- a/google/cloud/iam_policy.h
+++ b/google/cloud/iam_policy.h
@@ -28,7 +28,7 @@ inline namespace GOOGLE_CLOUD_CPP_NS {
  *
  * @see
  * https://cloud.google.com/resource-manager/reference/rest/Shared.Types/Policy
- *     for more information about a AIM policies.
+ *     for more information about IAM policies.
  *
  * @see https://tools.ietf.org/html/rfc7232#section-2.3 for more information
  *     about ETags.
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -322,6 +322,7 @@ def bucket_notification_delete(bucket_name, notification_id):
 
 @gcs.route("/b/<bucket_name>/iam")
 def bucket_get_iam_policy(bucket_name):
+    db.insert_test_bucket(None)
     bucket = db.get_bucket(flask.request, bucket_name, None)
     response = json_format.MessageToDict(bucket.iam_policy)
     response["kind"] = "storage#policy"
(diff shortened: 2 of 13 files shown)


remyabel:
This currently does not work. I ran into a snag with my munderstanding.The [JSON API](https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy) basically expects a `Policy` and that's how the emulator parses it. However, the proto seems to be in the form of:

```
    iam_request: {
      resource: "test-bucket-name"
      policy: {
        bindings: {
          role: "test-role"
          members: "user:test@example.com"
        }
        etag: "test-etag"
        version: 3
      }
    }
```

Changing the emulator would break the `bucket_integration_test`, so I am missing some key element here.

coryan:
> This currently does not work. I ran into a snag with my misunderstanding.The [JSON API](https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy) basically expects a `Policy` and that's how the emulator parses it.

Yes. It seems the JSON API flattens the policy request, and lifts the fields inside `Policy` to the request body. Compare the proto definitions:

https://github.com/googleapis/googleapis/blob/c7b5032f6ee39de6c0359b912012f99dfccb49b5/google/iam/v1/iam_policy.proto#L98-L110

https://github.com/googleapis/googleapis/blob/c7b5032f6ee39de6c0359b912012f99dfccb49b5/google/iam/v1/policy.proto#L88-L123

with the JSON resource that you quoted:

https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy#request-body

> Changing the emulator would break the `bucket_integration_test`, so I am missing some key element here.

I think there needs to be a layer in the emulator that deals with these differences between the proto and JSON representations of the request. They are basically the same contents, with just one level of nesting removed for JSON.

remyabel:
I have zero idea why, but I'm getting `Actual: Permanent error in GetNativeBucketIamPolicy: Method not implemented! [UNIMPLEMENTED] [UNIMPLEMENTED]` for `stub_->GetBucketIamPolicy`. Perhaps this PR is a bit premature.

remyabel:
I figured out the issue. I embarrassingly named the functions incorrectly and was looking in the wrong locations while debugging.

This PR should be ready to review with one caveat: setting the etag seems to make the emulator unhappy (maybe the tag needs to be in a certain format?) and since it is set to the timestamp anyway, I remove it from the request. I'm also not sure what the best way of testing it has the correct value is since it returns the timestamp of the current date.

coryan (approved):
:lgtm: Thanks!

> Reviewed 6 of 10 files at r1, 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5937#-:-MUdhR0TAlpUxh46Q0lM:b-g2a5fb)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5939, opened 2021-02-27
refactor: simplify the Options API, easier to use, improved efficiency

devjgm:
Part of https://github.com/googleapis/google-cloud-cpp/issues/5738
Fixes my concerns described in https://github.com/googleapis/google-cloud-cpp/pull/5864

The API changes in this PR

* "Option" structs now must contain a public `Type` member typedef indicating the type of the option's value. It no longer needs to have a public `.value` data member, and we no longer care how to construct one -- indeed, we never construct an "option" class at all.
* All uses of `absl::optional` are gone, as any "getter" method that might have to indicate that the value was not there. Instead, the only pure "getter" is `.get_or<T>(default_value)`, which returns by value, and can never fail because the caller must provide a default value to return if the option was not set (note: the default can simply be `{}`, but it's good that this is explicit at the call site).
* Since `absl::optional` is gone, I added `.has<T>()` to allow users to check if an option was set.
* Since `absl::optional` is gone, callers no longer have to dereference the value through something like `->value`. Instead, all accessors directly return the *value* (by copy or reference, depending on the method) that the user wanted, making call sites clearer and more succinct.
* Added a `.lookup<T>(init_value)` method that **returns a reference to the contained value**, inserting one if it was not already present. This behavior is similar to `std::map::insert`. This method makes it easy for users to do things like appending something to a potentially already present option. No pointers, or unnecessary copies are required here.

Note: It's possible that we might still want reference/pointer access to an option's value via a `const` accessor. If we end up wanting this for performance reasons, we could add this later similar to what was done in https://github.com/googleapis/google-cloud-cpp/pull/5860. We could even add it as a non-member friend that's `internal` so only we can use it if we want.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5939)

### DIFF
--- a/google/cloud/connection_options.h
+++ b/google/cloud/connection_options.h
@@ -262,8 +262,7 @@ Options MakeOptions(ConnectionOptions<ConnectionTraits>&& old) {
   opts.set<GrpcCredentialOption>(std::move(old.credentials_));
   opts.set<EndpointOption>(std::move(old.endpoint_));
   opts.set<GrpcNumChannelsOption>(old.num_channels_);
-  opts.set<UserAgentPrefixOption>(
-      std::set<std::string>{std::move(old.user_agent_prefix_)});
+  opts.set<UserAgentPrefixOption>({std::move(old.user_agent_prefix_)});
   opts.set<GrpcTracingOptionsOption>(std::move(old.tracing_options_));
   opts.set<GrpcBackgroundThreadsOption>(old.background_threads_factory());
   if (!old.tracing_components_.empty()) {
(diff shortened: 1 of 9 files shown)


devjgm:
No rush on this, but I added @coryan explicitly because he's had many thoughts and suggestions about this API, so I wanted to make sure he had a chance to comment on this proposed refactoring.

coryan (approved):
:lgtm:

> Reviewed 6 of 7 files at r1, 8 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5939#-:-MUdCXaQ5k4vufde8Ick:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)

___
*[google/cloud/internal/options.h, line 43 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5939#-MUdArZb9fB3YSZotCln:-MUdArZb9fB3YSZotClo:bbvv48m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1e32eacf045b0e37cdfa0969ddb796c0566c7ab2/google/cloud/internal/options.h#L43)):*
> ```objc
>  * An "Option" is any struct that has a public `Type` member typedef. By
>  * convention they are named like "FooOption". Each library (e.g., spanner,
>  * storage) may define their own set of options. Additionally, there may be
> ```

nit: s/there may be/there are/

devjgm:
> nit: s/there may be/there are/

Fixed.

PR #5940, opened 2021-02-28
cleanup: internal::MakeOptions accepts by value

devjgm:
A value (copy) is needed because we move the fields from the old
`ConnectionOptions` to the new `Options`. Previously it was requiring an
rvalue reference, which annoyingly required the caller to have to make
an explicit copy sometimes (e.g., if the caller only had a const-ref and
couldn't move).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5940)


coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5940#-:-MUdCxbZ2NEl2qR8u_D9:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5947, opened 2021-03-01
refactor: add Options::get<T>()

devjgm:
It turns out that we actually can provide a const `get<T>()` method that
returns a const-ref by defining that it returns a reference to a static
value-initialzied default value. I think this will be very useful for
cases where we want to query an option with changing or setting it. For
example, in a number of places we inspect the `TracingComponentsOption`
which is a `std::set<std::string>` to see if it contains an element.
Using `.get<T>()` would let us do this without copying the `set` every
time. I believe this will also allow us to re-implement
`ConnectionOptions` in terms of `Options`, since it can now return
const-refs, even to unset options.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5947)

### DIFF
--- a/google/cloud/internal/options.h
+++ b/google/cloud/internal/options.h
@@ -50,9 +50,10 @@ namespace internal {
  * - `.set<T>(x)`    -- Sets the option `T` to value `x`
  * - `.has<T>()`     -- Returns true iff option `T` is set
  * - `.unset<T>()`   -- Removes the option `T`
+ * - `.get<T>()`     -- Gets a const-ref to the value of option `T`
  * - `.get_or<T>(x)` -- Gets the value of option `T`, or `x` if no value was set
- * - `.lookup<T>(x)` -- Gets a reference to option `T`'s value, initializing it
- *                      to `x` if it was no set (`x` is optional).
+ * - `.lookup<T>(x)` -- Gets a non-const-ref to option `T`'s value, initializing
+ *                      it to `x` if it was no set (`x` is optional).
  *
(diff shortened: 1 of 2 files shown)


devjgm:
FYI, here's a diff of how we can change `ConnectionOptions` after this PR: https://github.com/googleapis/google-cloud-cpp/compare/master...devjgm:reimplement-connection-options?expand=1

Note: that that diff also contains the changes in this PR, because it's not merged yet.

coryan (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5947#-:-MUixyu4CUtqqaIP_kY7:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[google/cloud/internal/options.h, line 164 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5947#-MUixeiT8e0HT2Y7zQys:-MUixeiT8e0HT2Y7zQyt:bv592ij) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a1206cab157ebfdc69a4bd6af37906108a57704f/google/cloud/internal/options.h#L164)):*
> ```objc
>   template <typename T>
>   ValueTypeT<T> const& get() const {
>     static auto const* const kDefaultValue = new ValueTypeT<T>{};
> ```

FYI, remind me to write a rant about the style guide...

PR #5949, opened 2021-03-01
refactor: reimplement ConnectionOptions in terms of Options more

devjgm:
This refactoring takes advantage of the fact that `Options` now offers a
`get<T>()` const-ref accessor.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949)

### DIFF
--- a/google/cloud/connection_options.h
+++ b/google/cloud/connection_options.h
@@ -16,6 +16,7 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_CONNECTION_OPTIONS_H
 
 #include "google/cloud/completion_queue.h"
+#include "google/cloud/internal/algorithm.h"
 #include "google/cloud/internal/background_threads_impl.h"
 #include "google/cloud/internal/common_options.h"
 #include "google/cloud/internal/grpc_options.h"
@@ -56,24 +57,29 @@ class ConnectionOptions {
   /// The default options, using an explicit credentials object.
   explicit ConnectionOptions(
       std::shared_ptr<grpc::ChannelCredentials> credentials)
-      : credentials_(std::move(credentials)),
-        endpoint_(ConnectionTraits::default_endpoint()),
-        num_channels_(ConnectionTraits::default_num_channels()),
-        tracing_components_(internal::DefaultTracingComponents()),
-        tracing_options_(internal::DefaultTracingOptions()),
+      : opts_(internal::Options{}
+                  .set<internal::GrpcCredentialOption>(std::move(credentials))
+                  .set<internal::TracingComponentsOption>(
+                      internal::DefaultTracingComponents())
+                  .set<internal::GrpcTracingOptionsOption>(
+                      internal::DefaultTracingOptions())
+                  .template set<internal::EndpointOption>(
+                      ConnectionTraits::default_endpoint())
+                  .template set<internal::GrpcNumChannelsOption>(
+                      ConnectionTraits::default_num_channels())),
         user_agent_prefix_(ConnectionTraits::user_agent_prefix()) {}
 
   /// Change the gRPC credentials value.
   ConnectionOptions& set_credentials(
       // NOLINTNEXTLINE(performance-unnecessary-value-param) TODO(#4112)
       std::shared_ptr<grpc::ChannelCredentials> v) {
-    credentials_ = std::move(v);
+    opts_.set<internal::GrpcCredentialOption>(std::move(v));
     return *this;
   }
 
   /// The gRPC credentials used by clients configured with this object.
   std::shared_ptr<grpc::ChannelCredentials> credentials() const {
-    return credentials_;
+    return opts_.get<internal::GrpcCredentialOption>();
   }
 
   /**
@@ -87,12 +93,14 @@ class ConnectionOptions {
    */
   // NOLINTNEXTLINE(performance-unnecessary-value-param) TODO(#4112)
   ConnectionOptions& set_endpoint(std::string v) {
-    endpoint_ = std::move(v);
+    opts_.set<internal::EndpointOption>(std::move(v));
     return *this;
   }
 
   /// The endpoint used by clients configured with this object.
-  std::string const& endpoint() const { return endpoint_; }
+  std::string const& endpoint() const {
+    return opts_.get<internal::EndpointOption>();
+  }
 
   /**
    * The number of transport channels to create.
@@ -104,11 +112,13 @@ class ConnectionOptions {
    *
    * The default value is set by `ConnectionTraits::default_num_channels()`.
    */
-  int num_channels() const { return num_channels_; }
+  int num_channels() const {
+    return opts_.get<internal::GrpcNumChannelsOption>();
+  }
 
   /// Set the value for `num_channels()`.
   ConnectionOptions& set_num_channels(int num_channels) {
-    num_channels_ = num_channels;
+    opts_.set<internal::GrpcNumChannelsOption>(num_channels);
     return *this;
   }
 
@@ -120,28 +130,31 @@ class ConnectionOptions {
    * be enabled by clients configured with this option.
    */
   bool tracing_enabled(std::string const& component) const {
-    return tracing_components_.find(component) != tracing_components_.end();
+    return internal::Contains(opts_.get<internal::TracingComponentsOption>(),
+                              component);
   }
 
   /// Enable tracing for @p component in clients configured with this object.
   ConnectionOptions& enable_tracing(std::string const& component) {
-    tracing_components_.insert(component);
+    opts_.lookup<internal::TracingComponentsOption>().insert(component);
     return *this;
   }
 
   /// Disable tracing for @p component in clients configured with this object.
   ConnectionOptions& disable_tracing(std::string const& component) {
-    tracing_components_.erase(component);
+    opts_.lookup<internal::TracingComponentsOption>().erase(component);
     return *this;
   }
 
   /// Return the set of tracing components.
   std::set<std::string> const& components() const {
-    return tracing_components_;
+    return opts_.get<internal::TracingComponentsOption>();
   }
 
   /// Return the options for use when tracing RPCs.
-  TracingOptions const& tracing_options() const { return tracing_options_; }
+  TracingOptions const& tracing_options() const {
+    return opts_.get<internal::GrpcTracingOptionsOption>();
+  }
 
   /**
    * Define the gRPC channel domain for clients configured with this object.
@@ -229,8 +242,7 @@ class ConnectionOptions {
     return *this;
   }
 
-  using BackgroundThreadsFactory =
-      std::function<std::unique_ptr<BackgroundThreads>()>;
+  using BackgroundThreadsFactory = internal::BackgroundThreadsFactory;
   BackgroundThreadsFactory background_threads_factory() const {
     if (background_threads_factory_) return background_threads_factory_;
     auto const s = background_thread_pool_size_;
@@ -241,14 +253,12 @@ class ConnectionOptions {
   template <typename T>
   friend internal::Options internal::MakeOptions(ConnectionOptions<T>);
 
-  std::shared_ptr<grpc::ChannelCredentials> credentials_;
-  std::string endpoint_;
-  int num_channels_;
-  std::set<std::string> tracing_components_;
-  TracingOptions tracing_options_;
-  std::string channel_pool_domain_;
+  internal::Options opts_;
 
+  // These are fields that have different semantics than the equivalent ones in
+  // the new `Options` class.
   std::string user_agent_prefix_;
+  std::string channel_pool_domain_;
   std::size_t background_thread_pool_size_ = 0;
   BackgroundThreadsFactory background_threads_factory_;
 };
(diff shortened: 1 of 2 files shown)


devjgm (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/connection_options.h, line 60 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjW-z5Bs3OIO5nXLXk:-MUjdgldAeNkahxcO0aA:bvqbdqa) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L60)):*
I think this is consistent with what I previously said. When a user constructs an `Options` it will be "empty" and will have no defaults. That's different than what is happening here. This is a `ConnectionOptions` class, and it has different semantics. `ConnectionOptions` _does_ have default values. With this PR, `ConnectionOptions` is implemented in terms of `Options`, so it sets these defaults in the `Options`, to preserve its semantics. 

> Is that just because we want to simplify the "legacy" ConnectionOptions implementation, but we will continue to use ConnectionOptions as before?

Yes, I want to "simplify" the `ConnectionOptions` implementation. I used quotes around "simplify" because I realize it's not a ton simpler. But I also wanted to use `Options` to prove that it _could_ do the things that we wanted it to do.

I do not want to keep `ConnectionOptions` around. When we can, it should be deprecated, and users should be encouraged to use `Options` in all new code. But to be clear, `Options` will not have any default values.

> How will that fit with an overload for spanner::MakeConnection(internal::Options)?

`Options` will not have any default values. Unless, we converted a `ConnectionOptions` -> `Options`, in which case it will have all the values that the `ConnectionOptions` had.

`spanner::MakeConnection` will apply all of its own default values to the `Options` that it gets. it should look something like this:

```cc
// NEW
Connection spanner::MakeConnection(internal::Options opts) {
  // look at `opts` and apply appropriate defaults if an option is unset
}

// Deprecated
Connection spanner::MakeConnection(spanner::ConnectionOptions old) {
  return spanner::MakeConnection(internal::MakeOptions(old));
}
```
___
*[google/cloud/connection_options_test.cc, line 45 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjUUqNBIHD7agguujl:-MUjfz_Q-a08RBHNrHhT:b-egvlm3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options_test.cc#L45)):*

nit: this renaming made it harder to see what had changed in the test, no need to roll it back, just note for the future.
</blockquote></details>

Ack. Sorry. Too many things were named "options" in here, and I was lost.

coryan:
___
*[google/cloud/connection_options.h, line 60 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjW-z5Bs3OIO5nXLXk:-MUjguccEiLADNEiuAG7:b-650ot2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L60)):*

I think this is consistent with what I previously said. When a user constructs an `Options` it will be "empty" and will have no defaults. That's different than what is happening here. This is a `ConnectionOptions` class, and it has different semantics. `ConnectionOptions` _does_ have default values. With this PR, `ConnectionOptions` is implemented in terms of `Options`, so it sets these defaults in the `Options`, to preserve its semantics. 

> Is that just because we want to simplify the "legacy" ConnectionOptions implementation, but we will continue to use ConnectionOptions as before?

Yes, I want to "simplify" the `ConnectionOptions` implementation. I used quotes around "simplify" because I realize it's not a ton simpler. But I also wanted to use `Options` to prove that it _could_ do the things that we wanted it to do.

I do not want to keep `ConnectionOptions` around. When we can, it should be deprecated, and users should be encouraged to use `Options` in all new code. But to be clear, `Options` will not have any default values.

> How will that fit with an overload for spanner::MakeConnection(internal::Options)?

`Options` will not have any default values. Unless, we converted a `ConnectionOptions` -> `Options`, in which case it will have all the values that the `ConnectionOptions` had.

`spanner::MakeConnection` will apply all of its own default values to the `Options` that it gets. it should look something like this:

```cc
// NEW
Connection spanner::MakeConnection(internal::Options opts) {
  // look at `opts` and apply appropriate defaults if an option is unset
}

// Deprecated
Connection spanner::MakeConnection(spanner::ConnectionOptions old) {
  return spanner::MakeConnection(internal::MakeOptions(old));
}
```
</blockquote></details>

Won't the defaults be in two places then, or at least set from two places?  I imagine the implementation will be something like:

```cc
// NEW
Connection spanner::MakeConnection(internal::Options opts) {
  ...
  auto endpoint = opts.get_or<GrpcEndpoint>("spanner.googleapis.com");
  ...
}

// Deprecated
Connection spanner::MakeConnection(spanner::ConnectionOptions old) {
  return spanner::MakeConnection(internal::MakeOptions(old));
}
```

devjgm (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/connection_options.h, line 60 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjW-z5Bs3OIO5nXLXk:-MUji_r89k3aOaISoi1K:b-j3o4ps) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L60)):*

Won't the defaults be in two places then, or at least set from two places?  I imagine the implementation will be something like:

```cc
// NEW
Connection spanner::MakeConnection(internal::Options opts) {
  ...
  auto endpoint = opts.get_or<GrpcEndpoint>("spanner.googleapis.com");
  ...
}

// Deprecated
Connection spanner::MakeConnection(spanner::ConnectionOptions old) {
  return spanner::MakeConnection(internal::MakeOptions(old));
}
```

</blockquote></details>

Yes, until we get rid of `ConnectionOptions`. Then the defaults will only be in one place. And all the new services (e.g., the generated ones) will have defaults only in one place, because they won't use `ConnectionOptions` at all.

coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949)* status: all files reviewed, 2 unresolved discussions (waiting on @devjgm)

___
*[google/cloud/connection_options.h, line 260 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjtNlbDK0JN_8hR_nV:-MUjtNlbDK0JN_8hR_nW:bo9v3yt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L260)):*
> ```objc
>   // These are fields that have different semantics than the equivalent ones in
>   // the new `Options` class.
>   std::string user_agent_prefix_;
> ```

nit (maybe): why does `user_agent_prefix_` needs to be separate, you could do something like:

```cc
ConnectionOptions& add_user_agent_prefix(std::string p) {
  opts_.set<UserAgentPrefix>(p + " " + opts_.get<UserAgentPrefix>());
  return *this;
}
```

maybe?
___
*[google/cloud/connection_options.h, line 270 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjuPOL7UxUdv5lDyRk:-MUjuPOMBK9CyjwQ24uw:b-8dfd76) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L270)):*
> ```objc
> template <typename ConnectionTraits>
> Options MakeOptions(ConnectionOptions<ConnectionTraits> old) {
>   Options opts = std::move(old.opts_);
> ```

The `background_thread_pool_size_` member function is not used here, that seems odd.

devjgm (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/connection_options.h, line 260 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjtNlbDK0JN_8hR_nV:-MUjvi1QBHIJiiuCtnjp:b-y28v34) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L260)):*

nit (maybe): why does `user_agent_prefix_` needs to be separate, you could do something like:

```cc
ConnectionOptions& add_user_agent_prefix(std::string p) {
  opts_.set<UserAgentPrefix>(p + " " + opts_.get<UserAgentPrefix>());
  return *this;
}
```

maybe?

</blockquote></details>

That would work. That's how I first implemented it. But `internal::UserAgentPrefixOption` is a `std::set<std::string>` so it the actual implementation of this gets a little uglier with some uses of `*opts.get<...>().begin()` and the assumption that the set is never empty. Now, I think this code *does* assume/require that the user-agent prefix set would never be empty, so this assumption should work. It just started to look like I was trying too hard.

Of course, the problem may also be that `UserAgentPrefixOption` should *not* be a `std::set`, and it should instead be a `std::string`, like this is. However, since multiple user-agent prefixes can be specified, it seems like using a collection of some sort seems desirable going forward, which is why I landed on set.

WDYT? I can easily change to your above suggestion if you prefer, and I could also easily change the type of `UserAgentPrefixOption` if we think it's wrong.
___
*[google/cloud/connection_options.h, line 270 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjuPOL7UxUdv5lDyRk:-MUjwiV6CXvPwVgE8oT5:b-dkw8mz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L270)):*

The `background_thread_pool_size_` member function is not used here, that seems odd.
</blockquote></details>

That value is accounted for when we call 
```
  opts.set<GrpcBackgroundThreadsFactoryOption>(
      old.background_threads_factory());
```
below. That's the only place that value is used.

Note: I currently haven't exposed an option in `grpc_options.h` to tweak "background_thread_pool_size" on its own. The thinking is that since we already expose `GrpcBackgroundThreadsFactoryOption`, users could tweak the number of threads by changing the threads factory to something that uses more threads. Our `internal::AutomaticallyCreatedBackgroundThreads` class is not public, but if we like this approach and decide to, we could expose that or some other easy-to-use factory function for users to use.

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-:-MUk5xyN4d3AEtmUaagJ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[google/cloud/connection_options.h, line 260 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjtNlbDK0JN_8hR_nV:-MUk3HyI7gLp5vlIMsc1:blkd2t0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L260)):*
> Of course, the problem may also be that UserAgentPrefixOption should not be a std::set, and it should instead be a std::string, like this is. However, since multiple user-agent prefixes can be specified, it seems like using a collection of some sort seems desirable going forward, which is why I landed on set.

That is assuming the order does not matter, and uniqueness is important.

> WDYT? I can easily change to your above suggestion if you prefer, and I could also easily change the type of UserAgentPrefixOption if we think it's wrong.

Maybe a followup PR, I am not sure what the right representation for the `UserAgent` header should be.  I think the following properties are desirable:

- Applications can add their own.
- Intermediate libraries (think something like Tensorflow, built on top of our GCS library) could add their own.
- Applications cannot remove the prefixes added by libraries (because we lose the ability to troubleshoot things).
- The end result is a string that we put in some HTTP header (or a similar thing).

Seems like order should matter, maybe a `std::vector<>` is better than `std::set<>`, seems like it will always be "small", maybe a dozen elements. Seems like a `std::string` could handle this just fine.
___
*[google/cloud/connection_options.h, line 270 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjuPOL7UxUdv5lDyRk:-MUk5X7EBEZRPko_RQi3:bqubtxj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L270)):*

That value is accounted for when we call 
```
  opts.set<GrpcBackgroundThreadsFactoryOption>(
      old.background_threads_factory());
```
below. That's the only place that value is used.

Note: I currently haven't exposed an option in `grpc_options.h` to tweak "background_thread_pool_size" on its own. The thinking is that since we already expose `GrpcBackgroundThreadsFactoryOption`, users could tweak the number of threads by changing the threads factory to something that uses more threads. Our `internal::AutomaticallyCreatedBackgroundThreads` class is not public, but if we like this approach and decide to, we could expose that or some other easy-to-use factory function for users to use.
</blockquote></details>

Oh, I see, thanks.

devjgm:
> _[google/cloud/connection_options.h, line 260 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5949#-MUjtNlbDK0JN_8hR_nV:-MUk3HyI7gLp5vlIMsc1:blkd2t0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c755f4793c564407c976ced7901e1e13962b45b8/google/cloud/connection_options.h#L260)):_
> 
> > Of course, the problem may also be that UserAgentPrefixOption should not be a std::set, and it should instead be a std::string, like this is. However, since multiple user-agent prefixes can be specified, it seems like using a collection of some sort seems desirable going forward, which is why I landed on set.
> 
> That is assuming the order does not matter, and uniqueness is important.
> 
> > WDYT? I can easily change to your above suggestion if you prefer, and I could also easily change the type of UserAgentPrefixOption if we think it's wrong.
> 
> Maybe a followup PR, I am not sure what the right representation for the `UserAgent` header should be. I think the following properties are desirable:
> 
> * Applications can add their own.
> * Intermediate libraries (think something like Tensorflow, built on top of our GCS library) could add their own.
> * Applications cannot remove the prefixes added by libraries (because we lose the ability to troubleshoot things).
> * The end result is a string that we put in some HTTP header (or a similar thing).
> 
> Seems like order should matter, maybe a `std::vector<>` is better than `std::set<>`, seems like it will always be "small", maybe a dozen elements. Seems like a `std::string` could handle this just fine.

Filed https://github.com/googleapis/google-cloud-cpp/issues/5952 to track this. But TL;DR: I agree now that `std::set` was the wrong choice. Will fix in another PR.

PR #5955, opened 2021-03-02
refactor: remove Options::get_or

devjgm:
(sorry for all these `Options` API changes, but it's still internal and these should happen before it's public.)

After adding a const-ref-returning `Options::get<T>()`, I think there's
very little need for `get_or`. AFAICT, all uses of "get_or" are more
clearly written as "get", and they're more efficient. Additionally, the
functionality provided by "get_or" is "convenient" at best, and could be
completely implemented as a non-member helper function, thus improving
encapsulation.

It's always easier to add APIs than remove them, so we should err on the
side of removing this before the API is public.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5955)

### DIFF
--- a/google/cloud/internal/common_options_test.cc
+++ b/google/cloud/internal/common_options_test.cc
@@ -33,7 +33,7 @@ using ::testing::ContainsRegex;
 template <typename T, typename ValueType = typename T::Type>
 void TestOption(ValueType const& expected) {
   auto opts = Options{}.template set<T>(expected);
-  EXPECT_EQ(expected, opts.template get_or<T>({}))
+  EXPECT_EQ(expected, opts.template get<T>())
       << "Failed with type: " << typeid(T).name();
 }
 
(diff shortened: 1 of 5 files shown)


coryan (approved):
:lgtm: 

This reminds me (and only tangential to this PR), before deciding the "final" API for `Options` we should:

- Implement something like `spanner_internal::MakeConnection(internal::Options)` just to see how it fits when receiving one of these things.
- Use `Options` in a generated library.
- Solve a problem that would be difficult to fix with the current approach, maybe GCS+gRPC and credentials or setting the CA_PATH for libcurl and the equivalent for gRPC.
 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5955#-:-MUoHw6b1kAPOSdkdQhA:b-rie4p5)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5956, opened 2021-03-02
test(spanner): matching test instance with test resources

devbww:
Undo the addition of a predicate to `PickRandomInstance()` (from
to, say, one in a specific location where test resources were
configured.

A better tactic is to configure test resources in all possible
test-instance locations.  So, introduce `InstanceLocation()`, which
can be used to select location-specific test resources for the
randomly selected instance.

This will be needed to support the upcoming CMEK integration tests
and samples, where the location of the encryption keys must match
the location of the instance.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5956)


google/cloud/spanner/testing/instance_location.cc
```diff
@@ -0,0 +1,39 @@
...
+
+namespace google {
+namespace cloud {
+namespace spanner_testing {
+inline namespace SPANNER_CLIENT_NS {
+
+StatusOr<std::string> InstanceLocation(spanner::Instance const& in) {
+  spanner::InstanceAdminClient client(spanner::MakeInstanceAdminConnection());
+  auto instance = client.GetInstance(in);
+  if (!instance) return instance.status();
```
coryan, line 26: nit: s/return instance.status()/return std::move(instance).status()/

google/cloud/spanner/testing/instance_location.cc
```diff
@@ -0,0 +1,39 @@
...
+  auto instance_config = client.GetInstanceConfig(instance->config());
+  if (!instance_config) return instance_config.status();
```
coryan, line 28: ditto.

coryan (approved):
A couple of performance nits, your call.

google/cloud/spanner/testing/instance_location.cc
devbww: Done.

google/cloud/spanner/testing/instance_location.cc
devbww: Done.

coryan (approved):

PR #5964, opened 2021-03-03
ci: fix Ubuntu:16.04 images

### LINKED ISSUE #5963: ubuntu 16.04 builds failing to create updated docker image
An example from the `clang-3.8` build, which is using `Dockerfile.ubuntu` with version `16.04`: https://fusion2.corp.google.com/invocations/23a9b9c2-54e0-4ffd-9fb2-10893ac2819b/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fdocker%2Fclang-3.8/log
```
...
Reading state information...
E: Unable to locate package clang-9
E: Unable to locate package llvm-9
The command '/bin/sh -c apt-get update &&     apt-get --no-install-recommends install -y         automake         build-essential         ccache         clang         clang-9         cmake         ctags         curl         gawk         git         gcc         g++         cmake         libcurl4-openssl-dev         libssl-dev         libtool         llvm-9         lsb-release         make         ninja-build         pkg-config         python3         python3-dev         python3-pip         tar         unzip         zip         wget         zlib1g-dev         apt-utils         ca-certificates         apt-transport-https' returned a non-zero code: 100
```

It looks like this has been failing since Feb 2, 2021 when https://github.com/googleapis/google-cloud-cpp/pull/5794 was submitted.

We need to fix so that clang-9 and llvm-9 are only installed on ubuntu >= 18.04, not the 16.04 builds. Maybe we just need two separate docker files.

coryan:
The images were failing to build, as these images are only used in two
builds I took the opportunity to reduce the dependencies for the images.

Fixes #5963

### DIFF
--- a/ci/kokoro/docker/Dockerfile.ubuntu-xenial
+++ b/ci/kokoro/docker/Dockerfile.ubuntu-xenial
@@ -1,0 +1,50 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


devjgm (approved):
Thank you very much for grabbing this issue and fixing it so quickly!

PR #5665, opened 2020-12-30
feat: remove `CompletionQueue` from InstanceAdmin

dopiera:
This is a part of #2567.

This PR removes the `CompletionQueue` parameter from the `InstanceAdmin`
API. Apart from the obvious changes, additional option to
`bigtable::ConnectioOptions` was added to allow for custom worker
threads.

`AutomaticallyCreatedBackgroundThreads` now cancels all outstanding
operations in the destructor. Without this, destroying this object after
using it in clients with client refresh was taking multiple minutes.

A non-trivial concurrency change in the `CompletionQueue` was necessary.
Before this PR the `StartOp` passed to `StartOperation` was called with
the `CompletionQueue` lock held. This became problematic with this
change because the same completion queue is used for connection refresh
and all user activity. User activity, apart from calling the appropriate
method on the stub was also scheduling timers, yielding a deadlock.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5665)

PR #5967, opened 2021-03-04
feat: bind `CompletionQueue` with InstanceAdmin

dopiera:
This is a part of #2567.

Eventually, most `Async*` calls will be removed from the Bigtable API
but those which will remain (in `Table`) will have the `CompletionQueue`
argument removed. Instead, the `CompletionQueue` will be bound with
the `Table` object. In the same spirit, we're binding it with
`InstanceAdmin` in this PR. This allows to eliminate creating threads on
all calls which return `future<>`s but do not take a `CompletionQueue`
argument.

A decision was made to store the background threads in `InstanceAdmin`
rather than `CommonClient` even though `CommonClient` seems to resemble
other clients better. This is because `CommonClient` must not be the
owner of the background threads executing the completion queue
associated with itself because at some point the completion queue may hold
the last reference to the `CommonClient`. This is problematic because it
could potentially lead to a situation where `CompletionQueue` would
trigger `CommonClient`'s destructor, which in turn would trigger the
same `CompletionQueue`'s destructor, which would be really hard to
handle.

This is a reiteration of #5665.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967)

### DIFF
--- a/google/cloud/bigtable/client_options.cc
+++ b/google/cloud/bigtable/client_options.cc
@@ -13,6 +13,7 @@
 
 #include "google/cloud/bigtable/client_options.h"
 #include "google/cloud/bigtable/internal/client_options_defaults.h"
+#include "google/cloud/internal/background_threads_impl.h"
 #include "google/cloud/internal/build_info.h"
 #include "google/cloud/internal/getenv.h"
 #include "absl/strings/str_split.h"
@@ -157,6 +158,25 @@ std::string ClientOptions::UserAgentPrefix() {
   std::string agent = "gcloud-cpp/" + version_string() + " " +
                       google::cloud::internal::compiler();
   return agent;
+}
+
+ClientOptions& ClientOptions::DisableBackgroundThreads(
+    google::cloud::CompletionQueue const& cq) {
+  background_threads_factory_ = [cq] {
+    return absl::make_unique<
+        ::google::cloud::internal::CustomerSuppliedBackgroundThreads>(cq);
+  };
+  return *this;
+}
+
+ClientOptions::BackgroundThreadsFactory
+ClientOptions::background_threads_factory() const {
+  if (background_threads_factory_) return background_threads_factory_;
+  auto const s = background_thread_pool_size_;
+  return [s] {
+    return absl::make_unique<
+        ::google::cloud::internal::AutomaticallyCreatedBackgroundThreads>(s);
+  };
 }
 
 }  // namespace BIGTABLE_CLIENT_NS
--- a/google/cloud/bigtable/client_options.h
+++ b/google/cloud/bigtable/client_options.h
@@ -16,6 +16,8 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGTABLE_CLIENT_OPTIONS_H
 
 #include "google/cloud/bigtable/version.h"
+#include "google/cloud/background_threads.h"
+#include "google/cloud/completion_queue.h"
 #include "google/cloud/status.h"
 #include "google/cloud/tracing_options.h"
 #include <grpcpp/grpcpp.h>
@@ -375,6 +377,36 @@ class ClientOptions {
     return *this;
   }
 
+  /**
+   * Set the number of background threads.
+   *
+   * @note this value is not used if `DisableBackgroundThreads()` is called.
+   */
+  ClientOptions& set_background_thread_pool_size(std::size_t s) {
+    background_thread_pool_size_ = s;
+    return *this;
+  }
+  std::size_t background_thread_pool_size() const {
+    return background_thread_pool_size_;
+  }
+
+  /**
+   * Configure the connection to use @p cq for all background work.
+   *
+   * Connections need to perform background work on behalf of the application.
+   * Normally they just create a background thread and a `CompletionQueue` for
+   * this work, but the application may need more fine-grained control of their
+   * threads. In this case the application can provide the `CompletionQueue` and
+   * it assumes responsibility for creating one or more threads blocked on
+   * `CompletionQueue::Run()`.
+   */
+  ClientOptions& DisableBackgroundThreads(
+      google::cloud::CompletionQueue const& cq);
+
+  using BackgroundThreadsFactory =
+      std::function<std::unique_ptr<BackgroundThreads>()>;
+  BackgroundThreadsFactory background_threads_factory() const;
+
  private:
   friend struct internal::InstanceAdminTraits;
   friend struct ClientOptionsTestTraits;
@@ -399,6 +431,8 @@ class ClientOptions {
   TracingOptions tracing_options_;
   std::chrono::milliseconds max_conn_refresh_period_;
   std::chrono::milliseconds min_conn_refresh_period_;
+  std::size_t background_thread_pool_size_ = 0;
+  BackgroundThreadsFactory background_threads_factory_;
 };
 
 }  // namespace BIGTABLE_CLIENT_NS
--- a/google/cloud/bigtable/instance_admin.cc
+++ b/google/cloud/bigtable/instance_admin.cc
@@ -134,14 +134,8 @@ future<StatusOr<InstanceList>> InstanceAdmin::AsyncListInstances(
 
 future<StatusOr<btadmin::Instance>> InstanceAdmin::CreateInstance(
     InstanceConfig instance_config) {
-  CompletionQueue cq;
-  std::thread([](CompletionQueue cq) { cq.Run(); }, cq).detach();
-
-  return AsyncCreateInstance(cq, std::move(instance_config))
-      .then([cq](future<StatusOr<btadmin::Instance>> f) mutable {
-        cq.Shutdown();
-        return f.get();
-      });
+  auto cq = background_threads_->cq();
+  return AsyncCreateInstance(cq, std::move(instance_config));
 }
 
 future<StatusOr<google::bigtable::admin::v2::Instance>>
@@ -172,15 +166,9 @@ InstanceAdmin::AsyncCreateInstance(CompletionQueue& cq,
 future<StatusOr<btadmin::Cluster>> InstanceAdmin::CreateCluster(
     ClusterConfig cluster_config, std::string const& instance_id,
     std::string const& cluster_id) {
-  CompletionQueue cq;
-  std::thread([](CompletionQueue cq) { cq.Run(); }, cq).detach();
-
+  auto cq = background_threads_->cq();
   return AsyncCreateCluster(cq, std::move(cluster_config), instance_id,
-                            cluster_id)
-      .then([cq](future<StatusOr<btadmin::Cluster>> f) mutable {
-        cq.Shutdown();
-        return f.get();
-      });
+                            cluster_id);
 }
 
 future<StatusOr<google::bigtable::admin::v2::Cluster>>
@@ -213,14 +201,8 @@ InstanceAdmin::AsyncCreateCluster(CompletionQueue& cq,
 
 future<StatusOr<google::bigtable::admin::v2::Instance>>
 InstanceAdmin::UpdateInstance(InstanceUpdateConfig instance_update_config) {
-  CompletionQueue cq;
-  std::thread([](CompletionQueue cq) { cq.Run(); }, cq).detach();
-
-  return AsyncUpdateInstance(cq, std::move(instance_update_config))
-      .then([cq](future<StatusOr<btadmin::Instance>> f) mutable {
-        cq.Shutdown();
-        return f.get();
-      });
+  auto cq = background_threads_->cq();
+  return AsyncUpdateInstance(cq, std::move(instance_update_config));
 }
 
 future<StatusOr<google::bigtable::admin::v2::Instance>>
@@ -501,14 +483,8 @@ future<StatusOr<ClusterList>> InstanceAdmin::AsyncListClusters(
 
 future<StatusOr<google::bigtable::admin::v2::Cluster>>
 InstanceAdmin::UpdateCluster(ClusterConfig cluster_config) {
-  CompletionQueue cq;
-  std::thread([](CompletionQueue cq) { cq.Run(); }, cq).detach();
-
-  return AsyncUpdateCluster(cq, std::move(cluster_config))
-      .then([cq](future<StatusOr<btadmin::Cluster>> f) mutable {
-        cq.Shutdown();
-        return f.get();
-      });
+  auto cq = background_threads_->cq();
+  return AsyncUpdateCluster(cq, std::move(cluster_config));
 }
 
 future<StatusOr<google::bigtable::admin::v2::Cluster>>
@@ -636,14 +612,8 @@ InstanceAdmin::AsyncGetAppProfile(CompletionQueue& cq,
 future<StatusOr<btadmin::AppProfile>> InstanceAdmin::UpdateAppProfile(
     std::string const& instance_id, std::string const& profile_id,
     AppProfileUpdateConfig config) {
-  CompletionQueue cq;
-  std::thread([](CompletionQueue cq) { cq.Run(); }, cq).detach();
-
-  return AsyncUpdateAppProfile(cq, instance_id, profile_id, std::move(config))
-      .then([cq](future<StatusOr<btadmin::AppProfile>> f) mutable {
-        cq.Shutdown();
-        return f.get();
-      });
+  auto cq = background_threads_->cq();
+  return AsyncUpdateAppProfile(cq, instance_id, profile_id, std::move(config));
 }
 
 future<StatusOr<google::bigtable::admin::v2::AppProfile>>
--- a/google/cloud/bigtable/instance_admin.h
+++ b/google/cloud/bigtable/instance_admin.h
@@ -134,7 +134,8 @@ class InstanceAdmin {
         rpc_backoff_policy_prototype_(
             DefaultRPCBackoffPolicy(internal::kBigtableInstanceAdminLimits)),
         polling_policy_prototype_(
-            DefaultPollingPolicy(internal::kBigtableInstanceAdminLimits)) {}
+            DefaultPollingPolicy(internal::kBigtableInstanceAdminLimits)),
+        background_threads_(client_->BackgroundThreadsFactory()()) {}
 
   /**
    * Create a new InstanceAdmin using explicit policies to handle RPC errors.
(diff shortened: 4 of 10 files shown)


coryan (changes requested):
> Reviewed 17 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967)* status: all files reviewed, 9 unresolved discussions (waiting on @dopiera)

___
*[google/cloud/bigtable/instance_admin.h, line 234 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvAQOFE4nipxCYYd1U:-MUvAQOFE4nipxCYYd1V:btk17jw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.h#L234)):*
> ```objc
>    *     is not subject to any SLA or deprecation policy.
>    *
>    * @param cq the completion queue that will execute the asynchronous calls,
> ```

In the last couple of weeks we marked all these functions as deprecated, and announced they will be removed BY 2021-06-01 (in about 90d from now):

https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/version.h#L23-L26

I do not think it makes sense to break customers to fix the `CompletionQueue` parameter, and then break them again when we remove the function.  I also think there are no customers using these functions, but we should be disciplined about these things.

With that said:

- Having a `CompletionQueue` member would be useful, because the long running operations will still need them.
- I expect some of the functions will not be removed, but will become private, to be used in the polling loops of the long running operations.

So maybe leave the functions as-is, and just use the completion queue member in the long running operations? (notes on this below).
___
*[google/cloud/bigtable/instance_admin.h, line 251 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBQs25SFYMaGTUupM:-MUvBQs25SFYMaGTUupN:b8j3xsp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.h#L251)):*
> ```objc
>   GOOGLE_CLOUD_CPP_BIGTABLE_ADMIN_ASYNC_DEPRECATED
>   future<StatusOr<google::bigtable::admin::v2::Instance>> AsyncCreateInstance(
>       CompletionQueue& cq, bigtable::InstanceConfig instance_config);
> ```

please revert.
___
*[google/cloud/bigtable/instance_admin.cc, line 137 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBYeC2IQYD_UGh0ar:-MUvBYeC2IQYD_UGh0as:b-qv62x9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.cc#L137)):*
> ```cpp
> future<StatusOr<btadmin::Instance>> InstanceAdmin::CreateInstance(
>     InstanceConfig instance_config) {
>   CompletionQueue cq;
> ```

This is a place where we could use the member variable instead of creating a CQ on the fly... We would just pass that variable to `AsyncCreateInstance(background_thread_->cq(), ...)`
___
*[google/cloud/bigtable/instance_admin.cc, line 175 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBmKq-tnkzFZ73F9k:-MUvBmKq-tnkzFZ73F9l:bco2l7z) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.cc#L175)):*
> ```cpp
>     ClusterConfig cluster_config, std::string const& instance_id,
>     std::string const& cluster_id) {
>   CompletionQueue cq;
> ```

Another good use of the new `background_threads_`
___
*[google/cloud/bigtable/instance_admin.cc, line 216 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBt-S6CYWsztO8vI8:-MUvBt-S6CYWsztO8vI9:bs4q9op) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.cc#L216)):*
> ```cpp
> future<StatusOr<google::bigtable::admin::v2::Instance>>
> InstanceAdmin::UpdateInstance(InstanceUpdateConfig instance_update_config) {
>   CompletionQueue cq;
> ```

Ditto.
___
*[google/cloud/bigtable/instance_admin.cc, line 504 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBvKQCxH3qnNClT5W:-MUvBvKQCxH3qnNClT5X:bs4q9op) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.cc#L504)):*
> ```cpp
> future<StatusOr<google::bigtable::admin::v2::Cluster>>
> InstanceAdmin::UpdateCluster(ClusterConfig cluster_config) {
>   CompletionQueue cq;
> ```

Ditto.
___
*[google/cloud/bigtable/instance_admin.cc, line 639 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBwpT13yQeYJqhoQw:-MUvBwpT13yQeYJqhoQx:bs4q9op) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.cc#L639)):*
> ```cpp
>     std::string const& instance_id, std::string const& profile_id,
>     AppProfileUpdateConfig config) {
>   CompletionQueue cq;
> ```

Ditto.
___
*[google/cloud/bigtable/internal/common_client.h, line 134 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvCGRB9oJ7zIRsEDXO:-MUvCGRB9oJ7zIRsEDXP:b-9selnp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/internal/common_client.h#L134)):*
> ```objc
>             options_.max_conn_refresh_period())) {}
> 
>   ~CommonClient() {
> ```

I do not think I understand this change
___
*[google/cloud/bigtable/tests/instance_admin_async_future_integration_test.cc, line 515 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvCf5BDCHhL7WaXYuo:-MUvCf5BDCHhL7WaXYup:ba4r3tf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c818688b0b7398060edfa72d5f885b631cfe9fd6/google/cloud/bigtable/tests/instance_admin_async_future_integration_test.cc#L515)):*
> ```cpp
> 
> TEST_F(InstanceAdminAsyncFutureIntegrationTest, CustomWorkers) {
>   CompletionQueue cq;
> ```

It might be simpler to use a `AutomaticallyCreatedBackgroundThreads` object here to create and cleanup the thread and completion queue.

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967)* status: all files reviewed, 9 unresolved discussions (waiting on @coryan and @dopiera)

___
*[google/cloud/bigtable/instance_admin.h, line 234 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvAQOFE4nipxCYYd1U:-MUzUlXN62z-3TBev2kH:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.h#L234)):*

In the last couple of weeks we marked all these functions as deprecated, and announced they will be removed BY 2021-06-01 (in about 90d from now):

https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/version.h#L23-L26

I do not think it makes sense to break customers to fix the `CompletionQueue` parameter, and then break them again when we remove the function.  I also think there are no customers using these functions, but we should be disciplined about these things.

With that said:

- Having a `CompletionQueue` member would be useful, because the long running operations will still need them.
- I expect some of the functions will not be removed, but will become private, to be used in the polling loops of the long running operations.

So maybe leave the functions as-is, and just use the completion queue member in the long running operations? (notes on this below).

</blockquote></details>

Done
___
*[google/cloud/bigtable/instance_admin.h, line 251 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBQs25SFYMaGTUupM:-MUzUnjV4rL27fhZJfOr:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.h#L251)):*

please revert.
</blockquote></details>

Done.
___
*[google/cloud/bigtable/instance_admin.cc, line 137 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBYeC2IQYD_UGh0ar:-MUzUu-L7MQ3abNdnJiz:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.cc#L137)):*

This is a place where we could use the member variable instead of creating a CQ on the fly... We would just pass that variable to `AsyncCreateInstance(background_thread_->cq(), ...)`
</blockquote></details>

Done.
___
*[google/cloud/bigtable/instance_admin.cc, line 175 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBmKq-tnkzFZ73F9k:-MUzUzc03Ptmi4sc7QVu:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.cc#L175)):*

Another good use of the new `background_threads_`
</blockquote></details>

Done.
___
*[google/cloud/bigtable/instance_admin.cc, line 216 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvBt-S6CYWsztO8vI8:-MUzV0BS3KyOiwqzVXUi:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/instance_admin.cc#L216)):*

Ditto.
</blockquote></details>

Done.
___
*[google/cloud/bigtable/internal/common_client.h, line 134 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvCGRB9oJ7zIRsEDXO:-MUzD7S5DP1t62Z7YTN_:b1awgdp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/337aad332954046a998b90189127583c4bf9a658/google/cloud/bigtable/internal/common_client.h#L134)):*

I do not think I understand this change
</blockquote></details>

I have no idea, how it got here, probably some weird rebase failure (not sure if on mine or git's end). This is why the tests were failing.

The rename of `cq_` to `refresh_cq_` was intentional, however. The purpose is to highlight that this completion queue is only used for channel refresh. If you wonder why we're not using the same completion queue for operations and refresh, the reason is that this client may be reused between various `InstanceAdmin`s.

dopiera (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967)* status: 6 of 10 files reviewed, 9 unresolved discussions (waiting on @coryan and @dopiera)

___
*[google/cloud/bigtable/tests/instance_admin_async_future_integration_test.cc, line 515 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvCf5BDCHhL7WaXYuo:-MUzWNv34NQnLJqBcZ6f:bl7xpkq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c818688b0b7398060edfa72d5f885b631cfe9fd6/google/cloud/bigtable/tests/instance_admin_async_future_integration_test.cc#L515)):*

It might be simpler to use a `AutomaticallyCreatedBackgroundThreads` object here to create and cleanup the thread and completion queue.
</blockquote></details>

The test checks if the supplied `CompletionQueue` is used and that it is entirely controlled by the user.

I'm testing it by running an operation on a delivered `CompletionQueue`, which no thread is `Run()`ning and checking if operations don't complete. This allows this test to be falsely positive, but I guess it's good enough, since this will not make it flaky.

In order to do it, I need to `Run()` the `CompletionQueue` after the operation is started and I don't think `AutomaticallyCreatedBackgroundThreads` allows me to.

coryan (commented):
Looking good, let me think about the breaking change tomorrow.

> Reviewed 4 of 11 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-:-MUziiGa-KSjTdSJPKOd:bp2r2wm)* status: all files reviewed, 1 unresolved discussion (waiting on @dopiera)

___
*[ci/test-abi/google_cloud_cpp_bigtable.expected.abi.dump.gz, line 0 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUzhqiH6odfz-QlTPBP:-MUzhqiH6odfz-QlTPBQ:b-sgalmi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/53b36231b4eb014abe1a9042f32b81394ecb7b6e/ci/test-abi/google_cloud_cpp_bigtable.expected.abi.dump.gz#L0)):*
I think this was motivated by the new pure virtual in `InstanceAdminClient`?  That is a breaking change, applications need to inherit from that class to implement their own mocks.

At the very least the description of the PR changes to `feat!: ` (or `feat(bigtable)!: ` if you prefer.  It would be nice to update the CHANGELOG at the same time.  And I need to think when I am not super tired if there is a way to avoid the breakage.
___
*[google/cloud/bigtable/tests/instance_admin_async_future_integration_test.cc, line 515 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MUvCf5BDCHhL7WaXYuo:-MUzYGNR6f-3WCPwfnsS:b-8b1kps) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c818688b0b7398060edfa72d5f885b631cfe9fd6/google/cloud/bigtable/tests/instance_admin_async_future_integration_test.cc#L515)):*

The test checks if the supplied `CompletionQueue` is used and that it is entirely controlled by the user.

I'm testing it by running an operation on a delivered `CompletionQueue`, which no thread is `Run()`ning and checking if operations don't complete. This allows this test to be falsely positive, but I guess it's good enough, since this will not make it flaky.

In order to do it, I need to `Run()` the `CompletionQueue` after the operation is started and I don't think `AutomaticallyCreatedBackgroundThreads` allows me to.
</blockquote></details>

Ack.

dopiera (commented):
The obvious workaround seems to add a default implementation returning `CompletionQueue{}`.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-:-MUzl2YG7ykzpav02RrD:bvwn89n)* status: all files reviewed, 1 unresolved discussion (waiting on @dopiera)

coryan (commented):
> The obvious workaround seems to add a default implementation returning `CompletionQueue{}`.

I am looking at this log:

https://source.cloud.google.com/results/invocations/5998dc08-bd8f-46b4-bceb-5670cb54cb3a/targets

it says:

```
instance_admin_client.h
namespace google::cloud::bigtable::v1
[+] class InstanceAdminClient  1

         Change                                 Effect
                        Recompilation of a client program may be broken with
  Pure virtual method   the error message: cannot allocate an object of
1 Options ( ) has been  abstract type InstanceAdminClient because the following
  added to this class.  virtual functions are pure within InstanceAdminClient:
                        virtual Options ( ).
```


Maybe you mean returning `ClientOptions{}` because `CompletionQueue{}` won't work.  But `ClientOptions{}` won't work either, as this returns `ClientOptions const&` and we cannot return a reference to a temporary.

Now, the only thing we are using this new `Options()` is to return the `BackgroundThreadsFactory`.  I think we should do this:

* Instead of adding an `Options()` virtual, add a `BackgroundThreadsFactory()` virtual.
* It should return a `BackgroundThreadsFactory` by value.
* It should have a default implementation, it can return a factory that does not work (a default initialized `std::function<>` for example, or a `std::function<>` that always calls `ThrowRuntimeError()`).
* We need to make sure this new virtual does not become part of the public interface.  Make the new virtual member function `private`, so only `friends` can call it.  Fortunately `InstanceAdmin` is already a friend.

WDYT?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-:-MV1mSxu8po3P1R4PMSU:bfzvqom)* status: all files reviewed, 1 unresolved discussion (waiting on @dopiera)

___
*[google/cloud/bigtable/instance_admin.h, line 138 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-MV1p34YAP1J2dpwi15M:-MV1p34Z8DeX75Xaa4_e:blb21ac) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/53b36231b4eb014abe1a9042f32b81394ecb7b6e/google/cloud/bigtable/instance_admin.h#L138)):*
> ```objc
>         polling_policy_prototype_(
>             DefaultPollingPolicy(internal::kBigtableInstanceAdminLimits)),
>         background_threads_(client_->Options().background_threads_factory()()) {
> ```

FYI: Ah, so there are two thread pools, one for the `InstanceAdmin` object and one for the `CommonClient` object. Less than ideal, but Okay.

dopiera (commented):
This was my initial idea, but I went with `ClientOptions` because I was worried of this obscure detail becoming a part of the interface, but you're right - `private` and `friend` solve that problem.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-:-MVDn6Wo6xcko3no854e:b-52nr2y)* status: 5 of 10 files reviewed, 1 unresolved discussion (waiting on @coryan and @dopiera)

coryan (approved):
:lgtm: Thanks!

> Reviewed 6 of 6 files at r4.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5967#-:-MVF3Xof78nyuftm6TII:b-g2a5fb)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5973, opened 2021-03-04
fix(generator): ambiguous inline namespace issue between library and proto code

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5973)


devjgm:
I know there was some discussion about this issue on internal chat this morning. But for external people (and future us) looking at this change, it would be helpful to have a description of what the problem is/what and how/why this was the solution (and if there were alternative solutions consider). In the PR description is probably fine or a linked issue. Your call. Thanks.

devjgm (approved):

scotthart:
fixes #5974

PR #5984, opened 2021-03-05
DO NOT SUBMIT -- Change spanner to use Options

devjgm:
Related to https://github.com/googleapis/google-cloud-cpp/issues/5738

This PR changes all the Spanner code use the new `internal::Options` class. Functions accepting the old `spanner::ConnectionOptions` would still work, but they would quickly call the preferred `Options` overload.

**The intent of this PR is to show the design of how a product would use the new `Options`. This PR is still sloppy and would not be submitted as-is.** 

## The main design points are

* The `Options` class is still in `internal::` so please try to imagine all the code with `internal::` removed.
* All factory functions will accept -- and prefer -- an `internal::Options` object. An empty, default-constructed `Options` will be fine.
* Once a factory function gets an `options` argument (which is likely empty), it will call `spanner_internal::DefaultOptions()` to set and possibly override options as is appropriate for the service. See `MakeConnection` in client.cc as an example.
* End-users will never need to call `spanner_internal::DefaultOptions()`. _We_ make calls to `DefaultOptions()` internally in our tests some places, but that should not affect users, and it's never used in user-facing samples.
* The `spanner_internal::DefaultOptions()` function (see `.../spanner/internal/options.cc`) is the *one* place where have all the logic for setting and overriding options. This allows us to have logic that, say, overrides an explicitly specified option with an environment variable, which is something we actually need to do with `SPANNER_EMULATOR_HOST`. Today we cannot do this within the "options" framework, instead we needed a separate `EmulatorOverrides()` function for this special case.
* Each user-facing factory function that accepts an `Options` will also call `internal::CheckExpectedOptions<...>()` to log and warn the user to help with debugging if the user specified unexpected options.
* All the functions logically below the user-facing factory functions can assume that any `Options` they get are fully populated with any relevant defaults, and they can simply access options via the `Option::get<T>()` function. This function is `const`.
* If we have a service that needs/wants different defaults for different services, we need to provide a separate implementation of `spanner_internal::DefaultOptions()` for each. The name doesn't matter, it can be changed to be service-specific if we want.

## Open questions

1. Do we like this design? This is the test of the https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/internal/options.h API, and the ideas in https://github.com/googleapis/google-cloud-cpp/issues/5738
2. The user-facing factory functions accept `internal::Options` by value, because they may need to modify defaults and override things. This by-copy argument allows us to use move semantics. All other users of `Options` can be `const&` because they'll only use the `const` `.get<T>()` accessor. Do we like this? We could accept by const-ref everywhere and just make the copy internal to our factory.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5984)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -13,13 +13,18 @@
 // limitations under the License.
 
 #include "google/cloud/spanner/client.h"
+#include "google/cloud/spanner/connection_options.h"
 #include "google/cloud/spanner/internal/connection_impl.h"
+#include "google/cloud/spanner/internal/options.h"
 #include "google/cloud/spanner/internal/spanner_stub.h"
 #include "google/cloud/spanner/internal/status_utils.h"
 #include "google/cloud/spanner/retry_policy.h"
 #include "google/cloud/spanner/transaction.h"
 #include "google/cloud/backoff_policy.h"
+#include "google/cloud/internal/compiler_info.h"
 #include "google/cloud/internal/getenv.h"
+#include "google/cloud/internal/grpc_options.h"
+#include "google/cloud/internal/options.h"
 #include "google/cloud/internal/retry_loop.h"
 #include "google/cloud/log.h"
 #include <grpcpp/grpcpp.h>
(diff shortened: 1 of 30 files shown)


google/cloud/spanner/client.cc
```diff
@@ -324,30 +329,54 @@ QueryOptions Client::OverlayQueryOptions(QueryOptions const& preferred) {
...
   stubs.reserve(num_channels);
   for (int channel_id = 0; channel_id < num_channels; ++channel_id) {
-    stubs.push_back(spanner_internal::CreateDefaultSpannerStub(
-        db, connection_options, channel_id));
+    stubs.push_back(
+        spanner_internal::CreateDefaultSpannerStub(db, options, channel_id));
   }
   return spanner_internal::MakeConnection(
-      db, std::move(stubs), connection_options, std::move(session_pool_options),
+      db, std::move(stubs), options, std::move(session_pool_options),
```
coryan, line 356: nit: `std::move(options)` ?

google/cloud/spanner/client.h
```diff
@@ -652,9 +653,12 @@ class Client {
  *     by the `Connection`.
  */
 std::shared_ptr<Connection> MakeConnection(
-    Database const& db,
-    ConnectionOptions const& connection_options = ConnectionOptions(),
-    SessionPoolOptions session_pool_options = SessionPoolOptions());
+    Database const& db, internal::Options options = {},
+    SessionPoolOptions session_pool_options = {});
+// DEPRECATED
+std::shared_ptr<Connection> MakeConnection(
```
coryan, line 659: nit: here and below remember we have `GOOGLE_CLOUD_CPP_DEPRECATED(msg)` now.

google/cloud/spanner/client_test.cc
```diff
@@ -386,10 +387,10 @@ TEST(ClientTest, MakeConnectionOptionalArguments) {
   auto conn = MakeConnection(db);
   EXPECT_NE(conn, nullptr);
 
-  conn = MakeConnection(db, ConnectionOptions());
+  conn = MakeConnection(db, internal::Options{});
   EXPECT_NE(conn, nullptr);
 
-  conn = MakeConnection(db, ConnectionOptions(), SessionPoolOptions());
+  conn = MakeConnection(db, internal::Options{}, SessionPoolOptions{});
```
coryan, line 393: I wonder if `SessionPoolOptions` can be subsumed into `Options{}`, we added it because it was not possible to easily add stuff to `ConnectionOptions`.

google/cloud/spanner/internal/options.cc
```diff
@@ -0,0 +1,61 @@
...
+  }
+  if (!opts.has<internal::GrpcCredentialOption>()) {
+    opts.set<internal::GrpcCredentialOption>(grpc::GoogleDefaultCredentials());
+  }
+  if (!opts.has<internal::GrpcBackgroundThreadsFactoryOption>()) {
+    opts.set<internal::GrpcBackgroundThreadsFactoryOption>(
+        internal::DefaultBackgroundThreadsFactory);
+  }
+  if (opts.get<internal::GrpcNumChannelsOption>() < 1) {
+    opts.set<internal::GrpcNumChannelsOption>(1);
```
coryan, line 46: This changes the default number of channels from 4 to 1, intentional?

coryan (commented):
> Do we like this design?

I do.

> The user-facing factory functions accept `internal::Options` by value .... Do we like this?

I do.

google/cloud/spanner/client_test.cc
devjgm: Yes, I think it can. Good idea. Likely also the retry and backoff policies. I think we'll be able to get the `MakeConnection` factories to look like `MakeConnection(Database, Options opts = {})` with no other overloads.

devjgm:
Breaking this PR into smaller ones.

PR #5985, opened 2021-03-05
refactor(spanner): add spanner_internal::DefaultOptions() in preparation for using `Options`

devjgm:
Related to #5738
This PR breaks off a piece of: https://github.com/googleapis/google-cloud-cpp/pull/5984

This PR adds the `spanner_internal::DefaultOptions()` function (`google/cloud/spanner/internal/options.h`). All the defaults for spanner moved into the body of this function. The existing `ConnectionOptionsTraits` class was changed to pull its defaults out of `spanner_internal::DefaultOptions()`, so that we don't have defaults living in two places.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5985)


coryan (changes requested):
Why is this labeled `chore: `?  I would have called it `feat(spanner):` (or maybe `feat(spanner)!:` see below)...

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5985#-:-MV3Mo0C7HGEaByGsGbe:bajj229)* status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)

___
*[google/cloud/spanner/connection_options.cc, line 28 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5985#-MV3LKc5Bv-AslXr18Zg:-MV3LKc6CF244De9i-XZ:bbrxuve) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1542ced933bd274e934605251f399d24ae0d55d1/google/cloud/spanner/connection_options.cc#L28)):*
> ```cpp
> 
> std::string ConnectionOptionsTraits::default_endpoint() {
>   static auto const* const kEndpoint = new auto(
> ```

FYI, every time I see this pattern of not freeing the object I get :angry: :angry: :angry:  the style guide is soooo wrong about this.
___
*[google/cloud/spanner/connection_options_test.cc, line 61 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5985#-MV3LqdZCgorXA9x-sZW:-MV3LqdZCgorXA9x-sZX:b-bnlski) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a4f1e417cf3e69d26f169d8d816d8d21ed86d6df/google/cloud/spanner/connection_options_test.cc#L61)):*
> ```cpp
> TEST(DefaultEndpoint, EnvironmentWorks) {
>   EXPECT_NE("invalid-endpoint", ConnectionOptionsTraits::default_endpoint());
>   ScopedEnvironment env("GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT",
> ```

I think this is a breaking change, not a show stopping breaking change, but we should pause and think it through.

An application could have:

- call `spanner::MakeConnection()`, get one object,
- then call `setenv("GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT", ....)` set the environment variable variable to a different value,
- call `spanner::MakeConnection()`, and get a  new object pointing to a different endpoint.

Why would they do that?  I do not know, I can concoct a reason (maybe because they are using spanner through a wrapper and this is the only way to change the endpoint).

Regardless of whether there was a good reason or not: the change is breaking.  Some alternatives:

- Is it that important to optimize `default_endpoint()`?  If so, we should label this PR as `feat(spanner)!: ` and would be nice if you updated `CHANGELOG.md` in this PR.
- If it is not that important, then keep the existing functionality (however misguided) and don't cache the initial result.

devjgm (commented):
There's no user-facing feature in this PR. But `refactor` would be better. I'll change it.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5985#-:-MV3RgPaBkGhEs5XW3rc:b-f53tcf)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[google/cloud/spanner/connection_options.cc, line 28 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5985#-MV3LKc5Bv-AslXr18Zg:-MV3P43QEZ86dEDKU9IF:b24v65v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1542ced933bd274e934605251f399d24ae0d55d1/google/cloud/spanner/connection_options.cc#L28)):*

FYI, every time I see this pattern of not freeing the object I get :angry: :angry: :angry:  the style guide is soooo wrong about this.
</blockquote></details>

:shrug: Looks like correct to me, regardless of style. If there's a correct alternative, I'd be interested to learn it.
___
*[google/cloud/spanner/connection_options_test.cc, line 61 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5985#-MV3LqdZCgorXA9x-sZW:-MV3RK8q5D5b_sXsea-9:b-r252pm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a4f1e417cf3e69d26f169d8d816d8d21ed86d6df/google/cloud/spanner/connection_options_test.cc#L61)):*

I think this is a breaking change, not a show stopping breaking change, but we should pause and think it through.

An application could have:

- call `spanner::MakeConnection()`, get one object,
- then call `setenv("GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT", ....)` set the environment variable variable to a different value,
- call `spanner::MakeConnection()`, and get a  new object pointing to a different endpoint.

Why would they do that?  I do not know, I can concoct a reason (maybe because they are using spanner through a wrapper and this is the only way to change the endpoint).

Regardless of whether there was a good reason or not: the change is breaking.  Some alternatives:

- Is it that important to optimize `default_endpoint()`?  If so, we should label this PR as `feat(spanner)!: ` and would be nice if you updated `CHANGELOG.md` in this PR.
- If it is not that important, then keep the existing functionality (however misguided) and don't cache the initial result.

</blockquote></details>

It doesn't need to break in this PR. But I think it will be worth breaking after we add support for `MakeConnection()` (and friends) accepting the new `Options`, because then callers will have a supported alternative way to get the behavior you describe. I can break it in that PR, with a note in `CHANGELOG.md` at that point.

coryan:
___
*[google/cloud/spanner/connection_options_test.cc, line 61 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5985#-MV3LqdZCgorXA9x-sZW:-MV3T7_306CPU43knJti:bnjawjx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a4f1e417cf3e69d26f169d8d816d8d21ed86d6df/google/cloud/spanner/connection_options_test.cc#L61)):*

It doesn't need to break in this PR. But I think it will be worth breaking after we add support for `MakeConnection()` (and friends) accepting the new `Options`, because then callers will have a supported alternative way to get the behavior you describe. I can break it in that PR, with a note in `CHANGELOG.md` at that point.
</blockquote></details>

This PR is introducing the change in behavior, right?  Let's describe the change here.

devjgm:
> This PR is introducing the change in behavior, right? Let's describe the change here.

I don't think there's any change in behavior now that I changed the PR to no longer cache the endpoint in a function-local static. I also added back the `EnvironmentWorks` test to show this (though that file now has zero edits, so it's gone from the PR).

coryan (approved):
Doh, my UI was stuck in the previous revision. :lgtm: 

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5985#-:-MV3ZHKy3FPcB8PdIIyy:bz2bppl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5992, opened 2021-03-07
refactor(spanner): implement SessionPoolOptions in terms of Options

devjgm:
This will allow us to remove `SessionPoolOptions` in factory functions that take the new `Options` class. The new options are in `spanner_internal` for now. They'll be made public as soon as we decide to make `Options` public.

There should be no API or semantic change in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5992)

### DIFF
--- a/google/cloud/spanner/internal/options.cc
+++ b/google/cloud/spanner/internal/options.cc
@@ -56,24 +56,26 @@ internal::Options DefaultOptions(internal::Options opts) {
                       google::cloud::internal::CompilerFeatures() + ")");
 
   // Sets Spanner-specific options from session_pool_options.h
-  if (!opts.has<spanner::SessionPoolMaxSessionsPerChannelOption>()) {
-    opts.set<spanner::SessionPoolMaxSessionsPerChannelOption>(100);
+  if (!opts.has<spanner_internal::SessionPoolMaxSessionsPerChannelOption>()) {
+    opts.set<spanner_internal::SessionPoolMaxSessionsPerChannelOption>(100);
   }
-  if (!opts.has<spanner::SessionPoolActionOnExhaustionOption>()) {
-    opts.set<spanner::SessionPoolActionOnExhaustionOption>(
+  if (!opts.has<spanner_internal::SessionPoolActionOnExhaustionOption>()) {
+    opts.set<spanner_internal::SessionPoolActionOnExhaustionOption>(
         spanner::ActionOnExhaustion::kBlock);
   }
-  if (!opts.has<spanner::SessionPoolKeepAliveIntervalOption>()) {
-    opts.set<spanner::SessionPoolKeepAliveIntervalOption>(
+  if (!opts.has<spanner_internal::SessionPoolKeepAliveIntervalOption>()) {
+    opts.set<spanner_internal::SessionPoolKeepAliveIntervalOption>(
         std::chrono::minutes(55));
   }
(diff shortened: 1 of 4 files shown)


coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5992#-:-MVD4uBy2PQexSWeF17S:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @devjgm)

___
*[google/cloud/spanner/session_pool_options.h, line 49 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5992#-MVD3V7O1uXqZBOzXXr6:-MVD3V7O1uXqZBOzXXr7:b-29s1g2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/260a7564c16b28fa6a80c5e9a29d6d7a8db5cc57/google/cloud/spanner/session_pool_options.h#L49)):*
> ```objc
>  * the number of sessions (`max_sessions_per_channel` * number of channels).
>  *
>  * @note This option is to be used with the `google/cloud/internal/options.h`
> ```

nit: I think we should reference the type instead of the header.  There is some nuisance with cross-project references (as we define projects for `Doxygen), but that is a problem for both.
___
*[google/cloud/spanner/session_pool_options.h, line 136 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5992#-MVD44sx5zaf8g3nNPR8:-MVD44sx5zaf8g3nNPR9:b-k5ornt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/260a7564c16b28fa6a80c5e9a29d6d7a8db5cc57/google/cloud/spanner/session_pool_options.h#L136)):*
> ```objc
> 
>   /**
>    * Enforce the stated constraints on the option values, altering them if
> ```

FYI, unrelated to these changes... this is an odd function to have in the public interface.  And it does not help all that much... The setters can be called after it is called, invalidating the constraints...  I think this should never have been a public member function (a friend in `spanner_internal::` sounds right).

devjgm:
> nit: I think we should reference the type instead of the header. There is some nuisance with cross-project references (as we define projects for `Doxygen), but that is a problem for both.

You're right. I left the path to give me something easy to search for when options moves out of internal. However, a fully qualified class name is just as easy. So changed.

> 
> _[google/cloud/spanner/session_pool_options.h, line 136 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5992#-MVD44sx5zaf8g3nNPR8:-MVD44sx5zaf8g3nNPR9:b-k5ornt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/260a7564c16b28fa6a80c5e9a29d6d7a8db5cc57/google/cloud/spanner/session_pool_options.h#L136)):_
> 
> > ```objc
> >   /**
> >    * Enforce the stated constraints on the option values, altering them if
> > ```
> 
> FYI, unrelated to these changes... this is an odd function to have in the public interface. And it does not help all that much... The setters can be called after it is called, invalidating the constraints... I think this should never have been a public member function (a friend in `spanner_internal::` sounds right).

Agreed. The good news is that we don't need this function at all anymore in the new `Options`-based code. Though this will remain in the legacy `SessionPoolOptions` class. But once we start using `Options`, `EnforceConstraits()` will never be called anymore.

PR #5993, opened 2021-03-07
feat(common): make RetryPolicy and BackupPolicy interfaces public.

devjgm:
Moves the following classes from `google::cloud::internal` to `google::cloud`

* `google::cloud::internal::RetryPolicy` -> `google::cloud::RetryPolicy`. This is a newly public symbol, but it likely should have public anyway since we had other user-facing aliases referring to this symbol.
* `google::cloud::internal::BackoffPolicy` -> `google::cloud::BackoffPolicy`. This symbol was already public via an alis.
* `google::cloud::internal::ExpoentialBackoffPolicy` -> `google::cloud::ExponentialBackoffPolicy`. This symbol was already public via an alias.

This PR will allow us to create policy options for use with `Options` so we can remove factory overloads that only give users the ability to inject policies. We want to allow this via `Options` instead.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5993)


devjgm:
closing this for now. may re-open again later.

PR #5994, opened 2021-03-08
refactor(spanner): adds internal-only spanner-specific Retry and Backoff options

devjgm:
These options are internal-only and spanner specific. They will not be
used by customers and we can change them if/when we have other
user-facing options that we want our customers to use. In the mean time,
our implementation can use these options to pass around RetryPolicy and
BackoffPolicy objects inside an `Options`, without having to pass them
as separate arguments.

Note: these options hold the policies in `std::shared_ptr<T>` rather
than unique_ptr. This is because `Options` uses `absl::any`, which
doesn't work with move-only types. However, I think shared_ptr should be
fine for these.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5994)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -66,18 +66,20 @@ class DefaultPartialResultSetReader : public PartialResultSetReader {
 
 namespace spanner_proto = ::google::spanner::v1;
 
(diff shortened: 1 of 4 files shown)


coryan (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5994)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5995, opened 2021-03-08
refactor(spanner): remove unnecessary MakeConnection factory

devjgm:
There's no need for a factory to construct
`spanner_internal::ConnectionImpl`. We can just construct it directly
from `client.cc`. I don't think the factory was adding any value, but it
was adding an extra layer of things to plumb parameters through.
This also makes the code easier to follow.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5995)


coryan (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5995)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5996, opened 2021-03-08
refactor(generator): streaming reads to accept proto by const-ref

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/5996)


devjgm (approved):
Modulo the broken builds, the change LGTM

PR #6005, opened 2021-03-09
ci: troubleshoot macOS clock skew

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6005)


coryan:
Does not seem to work, and I am out of ideas.

PR #6016, opened 2021-03-10
cleanup(spanner): new version time in spanner_create_backup sample

devbww:
A newly changed requirement on the PITR `CreateBackup()` sample
(see #5943 for the previous change) to hide the source of the
`version_time` from the snippet altogether by passing it as a
parameter.  Let's extend that decision to `expire_time` too.

In the process, let's avoid cross-clock-domain issues as well,
by only using timestamps originated by the service.  In these
cases we only need "now", from `SELECT CURRENT_TIMESTAMP()`
instead of from `absl::Now()`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6016)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -30,7 +30,6 @@
 #include "google/cloud/spanner/update_instance_request_builder.h"
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/internal/random.h"
-#include "absl/time/clock.h"
 #include "absl/time/time.h"
 #include "absl/types/optional.h"
 #include <chrono>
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -439,6 +438,27 @@ void InstanceTestIamPermissionsCommand(std::vector<std::string> const& argv) {
   InstanceTestIamPermissions(std::move(client), argv[0], argv[1]);
 }
 
+google::cloud::spanner::Timestamp DatabaseNow(std::string const& project_id,
+                                              std::string const& instance_id,
+                                              std::string const& database_id) {
+  auto client = google::cloud::spanner::Client(
+      google::cloud::spanner::MakeConnection(google::cloud::spanner::Database(
```
coryan, line 445: Consider refactoring to accept a `spanner::Client` parameter and reuse the parameter between example calls in `RunAll()`.

FYI, we generally use an existing client for these examples because it was simply too slow to create a client in each example (and that was when we had a fraction of the examples we have now).  Two or three calls as in this PR is not a big deal, but it can grow over time.

google/cloud/spanner/samples/samples.cc
```diff
@@ -1354,13 +1363,14 @@ void QueryWithStringParameter(google::cloud::spanner::Client client) {
...
+void QueryWithTimestampParameter(
+    google::cloud::spanner::Client client,
+    google::cloud::spanner::Timestamp last_update_time) {
   namespace spanner = ::google::cloud::spanner;
-  auto example_timestamp = spanner::MakeTimestamp(absl::Now()).value();
   spanner::SqlStatement select(
       "SELECT VenueId, VenueName, LastUpdateTime FROM Venues"
       " WHERE LastUpdateTime <= @last_update_time",
-      {{"last_update_time", spanner::Value(example_timestamp)}});
+      {{"last_update_time", spanner::Value(last_update_time)}});
```
coryan, line 1373: This example may be hard to grok when reading it at cloud.google.com, the context of where `last_update_time` comes from is lost. Consider putting the body of the query inline the example.  Alternatively, one can reopen a region (in this case the `[START spanner_query_with_timestamp_parameter]`), so you could refactor this example to call a version of `DatabaseNow()` which is included in the example.  I say I version because I would pass an already built `spanner::Client` to it.

coryan (commented):
Some thoughts below, if I am off base let me know.

google/cloud/spanner/samples/samples.cc
devbww: Done, although it really only helps the `QueryWithTimestampParameter()` call.

google/cloud/spanner/samples/samples.cc
devbww: I think losing the particular value of `last_update_time` from the snippet is the goal.  The value is arbitrary.  The snippet is only about how to interpolate it into a query.  Yes?

coryan (approved):

PR #6019, opened 2021-03-11
refactor(generator): use new Options

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019)


generator/internal/stub_factory_generator.cc
```diff
@@ -51,9 +51,13 @@ Status StubFactoryGenerator::GenerateHeader() {
   auto result = HeaderOpenNamespaces(NamespaceType::kInternal);
   if (!result.ok()) return result;
 
+  HeaderPrint(  // clang-format off
+    "internal::Options Resolve$service_name$Options(internal::Options options);\n\n");
```
devjgm, line 55: I think this should be `Default$service_name$Options()`, and the argument should have a default value of `= {}`. This lets us either fill in an existing options with the defaults, or construct a new object fully filled in with defaults.

devjgm, line 55: Nit: By convention so far, I've been calling the variable of type `Options` simply `opts`. This has allowed by to distinguish it from existing variables of type `ConnectionOption`, which are typically all named `options`. And now I've gotten used to the short name `opts` for these new things, and it looks nice and normal now. So FWIW, consider just `opts`.

generator/internal/stub_factory_generator.cc
```diff
@@ -51,9 +51,13 @@ Status StubFactoryGenerator::GenerateHeader() {
...
+  //clang-format on
+
   HeaderPrint(  // clang-format off
     "std::shared_ptr<$stub_class_name$>\n"
-    "CreateDefault$stub_class_name$($product_namespace$::$connection_options_name$ const& options);\n\n");
+    "CreateDefault$stub_class_name$(internal::Options options);\n\n");
```
devjgm, line 60: I think we're _really_ close (any day now) close to making `Options` non-internal. When we do that all these sites will need to be renamed, along with all the other call sites outside of the generator code. We'll have to do this all in one big PR.

Do you imagine that will be easy to do w/ the generator and its artifacts? Or would it be easier to hold off on this PR for a couple days until `Options` is non-internal? Either's fine with me as long as the generator code will respond well to a large rename PR.

generator/internal/stub_factory_generator.cc
```diff
@@ -73,22 +77,53 @@ Status StubFactoryGenerator::GenerateCc() {
...
+    "  }\n"
+    "  if (!options.has<internal::GrpcCredentialOption>()) {\n"
+    "    options.set<internal::GrpcCredentialOption>(grpc::GoogleDefaultCredentials());\n"
+    "  }\n"
+    "  if (!options.has<internal::GrpcBackgroundThreadsFactoryOption>()) {\n"
+    "    options.set<internal::GrpcBackgroundThreadsFactoryOption>(\n"
+    "        internal::DefaultBackgroundThreadsFactory);\n"
+    "  }\n"
+    "  if (!options.has<internal::GrpcNumChannelsOption>()) {\n"
+    "    options.set<internal::GrpcNumChannelsOption>(4);\n"
```
devjgm, line 108: This value of 4 here may be spanner specific. I'm not sure what a good default value is for other services. Not sure why 4 is better than 1 or N. You may want to start a thread w/ the team to see if anyone has any thoughts on this question.

generator/internal/stub_factory_generator.cc
```diff
@@ -73,22 +77,53 @@ Status StubFactoryGenerator::GenerateCc() {
...
+
   // factory function implementation
   CcPrint(  // clang-format off
     "std::shared_ptr<$stub_class_name$>\n"
-    "CreateDefault$stub_class_name$($product_namespace$::$connection_options_name$ const& options) {\n"
-    "  auto channel =\n"
-    "      grpc::CreateCustomChannel(options.endpoint(), options.credentials(),\n"
-    "                                options.CreateChannelArguments());\n"
+    "CreateDefault$stub_class_name$(internal::Options options) {\n"
+    "  options = Resolve$service_name$Options(options);\n"
```
devjgm, line 122: The argument should be moved into the function.

generator/internal/stub_factory_generator.cc
```diff
@@ -116,11 +151,13 @@ Status StubFactoryGenerator::GenerateCc() {
...
+    "  if (internal::Contains(\n"
+    "      options.get<internal::TracingComponentsOption>(), \"rpc\")) {\n"
     "    GCP_LOG(INFO) << \"Enabled logging for gRPC calls\";\n"
-    "    stub = std::make_shared<$logging_class_name$>(std::move(stub),\n"
-    "                                                  options.tracing_options(),\n"
-    "                                                  options.components());\n"
+    "    stub = std::make_shared<$logging_class_name$>(\n"
+    "        std::move(stub),\n"
+    "        options.get<internal::GrpcTracingOptionsOption>(),\n"
+    "        options.get<internal::TracingComponentsOption>());\n"
```
devjgm, line 160: I wouldn't expect the stub to need to know tracing components.

google/cloud/bigquery/big_query_read_connection.gcpcxx.pb.h
```diff
@@ -78,11 +69,10 @@ class BigQueryReadConnection {
 };
 
 std::shared_ptr<BigQueryReadConnection> MakeBigQueryReadConnection(
-    BigQueryReadConnectionOptions const& options =
-        BigQueryReadConnectionOptions());
+    internal::Options const& options = internal::Options());
```
devjgm, line 72: This is the only overload that we want. Per our discussion at the team meeting, we want to avoid the other overloads that take separate policy arguments.

devjgm, line 72: Just `= {}`. It's simpler, and the type is already mentioned, so there's no need to mention it again.

Also, the argument should be passed by value, because it will be modified by the implementation.

google/cloud/bigquery/integration_tests/big_query_read_integration_test.cc
```diff
@@ -33,15 +34,15 @@ using ::testing::HasSubstr;
...
+    connection_options_.set<internal::TracingComponentsOption>({"rpc"});
     retry_policy_ =
         absl::make_unique<BigQueryReadLimitedErrorCountRetryPolicy>(1);
     backoff_policy_ = absl::make_unique<ExponentialBackoffPolicy>(
         std::chrono::seconds(1), std::chrono::seconds(1), 2.0);
     idempotency_policy_ = MakeDefaultBigQueryReadConnectionIdempotencyPolicy();
   }
   std::vector<std::string> ClearLogLines() { return log_.ExtractLines(); }
-  BigQueryReadConnectionOptions connection_options_;
+  internal::Options connection_options_;
```
devjgm, line 45: s/connection_option_/opts_/g

google/cloud/iam/integration_tests/iam_credentials_integration_test.cc
```diff
@@ -46,7 +46,7 @@ class IamCredentialsIntegrationTest : public ::testing::Test {
     ASSERT_FALSE(invalid_iam_service_account_.empty());
   }
   std::vector<std::string> ClearLogLines() { return log_.ExtractLines(); }
-  IAMCredentialsConnectionOptions rpc_tracing_options_;
+  internal::Options rpc_tracing_options_;
```
devjgm, line 49: "rcp_tracing_options_" seems like a misleading name. Probably want just `opts_`.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019)* status: 0 of 32 files reviewed, 8 unresolved discussions (waiting on @devjgm)

___
*[generator/internal/stub_factory_generator.cc, line 55 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTcSovDStBk3nBowF3-r1-55:-MVcBYDv8hAUGTlcVrw4:b3kz4iw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/generator/internal/stub_factory_generator.cc#L55)):*

Nit: By convention so far, I've been calling the variable of type `Options` simply `opts`. This has allowed by to distinguish it from existing variables of type `ConnectionOption`, which are typically all named `options`. And now I've gotten used to the short name `opts` for these new things, and it looks nice and normal now. So FWIW, consider just `opts`.
</blockquote></details>

re:opts: Per the style guide, I thought we were supposed to prefer not to abbreviate names?
___
*[generator/internal/stub_factory_generator.cc, line 60 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTcSovDStBk3nBowF3-r1-60:-MVcDaeS0JcQLJA269FF:b-kscwzg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/generator/internal/stub_factory_generator.cc#L60)):*

I think we're _really_ close (any day now) close to making `Options` non-internal. When we do that all these sites will need to be renamed, along with all the other call sites outside of the generator code. We'll have to do this all in one big PR.

Do you imagine that will be easy to do w/ the generator and its artifacts? Or would it be easier to hold off on this PR for a couple days until `Options` is non-internal? Either's fine with me as long as the generator code will respond well to a large rename PR. 
</blockquote></details>

Waiting is fine with me.
___
*[generator/internal/stub_factory_generator.cc, line 108 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTcSovDStBk3nBowF3-r1-108:-MVcEaN23tq4HQRXZWtx:b-sf97m0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/generator/internal/stub_factory_generator.cc#L108)):*

This value of 4 here may be spanner specific. I'm not sure what a good default value is for other services. Not sure why 4 is better than 1 or N. You may want to start a thread w/ the team to see if anyone has any thoughts on this question.
</blockquote></details>

I'll add it as a discussion item for the next team meeting.
___
*[generator/internal/stub_factory_generator.cc, line 122 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTcSovDStBk3nBowF3-r1-122:-MVcExV3-b2m67A6pZQu:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/generator/internal/stub_factory_generator.cc#L122)):*

The argument should be moved into the function.
</blockquote></details>

Done.
___
*[generator/internal/stub_factory_generator.cc, line 160 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTcSovDStBk3nBowF3-r1-160:-MVcDiNNDfjRM-lGcyc9:b2r17l8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/generator/internal/stub_factory_generator.cc#L160)):*

I wouldn't expect the stub to need to know tracing components.
</blockquote></details>

This adds the logging decorator to the existing stub, which does need that information.
___
*[google/cloud/bigquery/big_query_read_connection.gcpcxx.pb.h, line 72 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTmD7HmafEh4halcBt-r1-72:-MVcCDOj6p7Q2I8lAhYL:btjidxo) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/google/cloud/bigquery/big_query_read_connection.gcpcxx.pb.h#L72)):*

Just `= {}`. It's simpler, and the type is already mentioned, so there's no need to mention it again.

Also, the argument should be passed by value, because it will be modified by the implementation.
</blockquote></details>

I was going to fold in the policies into Options in a separate PR, to keep the size down, provided this one was merged in a timely fashion. Given that I will hold this PR until Options is no longer internal, I will go ahead and do the needful with the policies.

I think we still want the overload that takes a stub and policies for the case where the user adds their own custom decorators to the stub.

We had a discussion the other day about accepting by-value or by-const ref where the parameter was merely passed through one function to another. How is this different?
___
*[google/cloud/bigquery/integration_tests/big_query_read_integration_test.cc, line 45 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTmD7HmafEh4halcBu-r1-45:-MVcIF0F0a6xF8vq6KPL:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/google/cloud/bigquery/integration_tests/big_query_read_integration_test.cc#L45)):*

s/connection_option_/opts_/g
</blockquote></details>

Done.
___
*[google/cloud/iam/integration_tests/iam_credentials_integration_test.cc, line 49 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTmD7HmafEh4halcBz-r1-49:-MVcIFWD8sVWxOBLFYKZ:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/google/cloud/iam/integration_tests/iam_credentials_integration_test.cc#L49)):*

"rcp_tracing_options_" seems like a misleading name. Probably want just `opts_`.
</blockquote></details>

Done.

devjgm (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019)* status: 0 of 32 files reviewed, 8 unresolved discussions (waiting on @devjgm and @scotthart)

___
*[generator/internal/stub_factory_generator.cc, line 55 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTcSovDStBk3nBowF3-r1-55:-MVcJ9LUDg9hqQBFfl59:b-vj154r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/generator/internal/stub_factory_generator.cc#L55)):*

re:opts: Per the style guide, I thought we were supposed to prefer not to abbreviate names?
</blockquote></details>

The style guide would allow either here. "opts" is a common abbreviation for options. Just like we allow (nay prefer) `it` to refer to an iterator. It would be hard to argue that `opts.get<Foo>()` is less clear than `options.get<Foo>()`, especially in cases where the longer one may cause more line wrapping, which also hurts readability.

Not a big deal either way though.
___
*[generator/internal/stub_factory_generator.cc, line 108 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTcSovDStBk3nBowF3-r1-108:-MVcK4Jx0uz4BASRkuo3:b-ucfkn4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/generator/internal/stub_factory_generator.cc#L108)):*

I'll add it as a discussion item for the next team meeting.
</blockquote></details>

Thanks.
___
*[generator/internal/stub_factory_generator.cc, line 160 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTcSovDStBk3nBowF3-r1-160:-MVcKcNRAADXjN6Ek3qN:b-psfpn7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/generator/internal/stub_factory_generator.cc#L160)):*

This adds the logging decorator to the existing stub, which does need that information.
</blockquote></details>

Ahh, thanks. Interesting. The spanner stub does not need the components, so I was confused.

https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/spanner/internal/spanner_stub.cc
___
*[google/cloud/bigquery/big_query_read_connection.gcpcxx.pb.h, line 72 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTmD7HmafEh4halcBt-r1-72:-MVcL0jr1ICtq8nGjng7:b-jz78i8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/af37b901d2a0966ed9de1965c89ffe844529ce7a/google/cloud/bigquery/big_query_read_connection.gcpcxx.pb.h#L72)):*
> I was going to fold in the policies into Options in a separate PR, to keep the size down, provided this one was merged in a timely fashion. Given that I will hold this PR until Options is no longer internal, I will go ahead and do the needful with the policies.

SGTM. Thanks for being flexible.

> I think we still want the overload that takes a stub and policies for the case where the user adds their own custom decorators to the stub.

Everything the user can modify will be passed via options, including the policy objects. So there will be no need for overloads that accept the policy objects.

> We had a discussion the other day about accepting by-value or by-const ref where the parameter was merely passed through one function to another. How is this different?

This `opts` param isn't merely passed through. It's modified by calling `DefaultFooOptions`. So we can say that this factory function itself *needs* a copy.

generator/integration_tests/golden/internal/golden_kitchen_sink_stub_factory.gcpcxx.pb.cc
```diff
@@ -28,11 +34,34 @@ namespace cloud {
...
+  return options;
+}
+
 std::shared_ptr<GoldenKitchenSinkStub>
-CreateDefaultGoldenKitchenSinkStub(golden::GoldenKitchenSinkConnectionOptions const& options) {
-  auto channel =
-      grpc::CreateCustomChannel(options.endpoint(), options.credentials(),
-                                options.CreateChannelArguments());
+CreateDefaultGoldenKitchenSinkStub(internal::Options options) {
+  options = ResolveGoldenKitchenSinkOptions(std::move(options));
```
devjgm, line 60: There should be no need to call `ResolveGoldenKitchenSinkOptions()` here, because the user-facing "MakeConnection" factory would have already set all the defaults as appropriate. Therefore, this function can/should accept `internal::Options const& opts` by const-ref, because it's only reading from the object.

FWIW: The approach I've taken with Spanner is this:

- Spanner's user-facing "make connection" factory functions accept `internal::Options opts` *by value*, because they need a copy because they're going to tweak the options to set any missing defaults appropriately. This is the only place -- in the one "connection factory layer" -- where defaults are resolved.
- All other *non-user-facing* functions that accept an `Optoins` do so by const-ref, because they don't need to modify anything.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019)* status: 0 of 32 files reviewed, 9 unresolved discussions (waiting on @devjgm and @scotthart)

___
*[generator/integration_tests/golden/internal/golden_kitchen_sink_stub_factory.gcpcxx.pb.cc, line 60 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-MVTcSovDStBk3nBowEt-r2-60:-MVcYtD_Djp8dCeDZiDP:b-mgkbqa) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c5604a205b1c936f90827277d7b0ded7f61fd5cf/generator/integration_tests/golden/internal/golden_kitchen_sink_stub_factory.gcpcxx.pb.cc#L60)):*

There should be no need to call `ResolveGoldenKitchenSinkOptions()` here, because the user-facing "MakeConnection" factory would have already set all the defaults as appropriate. Therefore, this function can/should accept `internal::Options const& opts` by const-ref, because it's only reading from the object.

FWIW: The approach I've taken with Spanner is this:

- Spanner's user-facing "make connection" factory functions accept `internal::Options opts` *by value*, because they need a copy because they're going to tweak the options to set any missing defaults appropriately. This is the only place -- in the one "connection factory layer" -- where defaults are resolved.
- All other *non-user-facing* functions that accept an `Optoins` do so by const-ref, because they don't need to modify anything.
</blockquote></details>

Adding the policies as members of the Options will necessitate percolating the Options resolution up to the Connection layer. I'll adjust this accordingly when I make the policy options changes.

devjgm:
FYI, as of https://github.com/googleapis/google-cloud-cpp/pull/6042 (merged) the `Options` are now public.

scotthart (commented):
Rebased; policy options added; libraries generated 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6019#-:-MVydqQe2RcjuMa8xHA6:b3x8tzh)* status: 0 of 67 files reviewed, 9 unresolved discussions (waiting on @devjgm and @scotthart)

google/cloud/bigquery/big_query_read_options.gcpcxx.pb.h
```diff
@@ -0,0 +1,54 @@
...
+#include "google/cloud/options.h"
+#include "google/cloud/version.h"
+#include <memory>
+
+namespace google {
+namespace cloud {
+namespace bigquery {
+inline namespace GOOGLE_CLOUD_CPP_GENERATED_NS {
+
+struct BigQueryReadRetryPolicyOption {
```
devjgm, line 33: Would it be possible to inject a generic comment here like "/// An option to use with `google::cloud::Options`"? Some tie back go `google::cloud::Options` may be helpful for customers.

A followup PR is fine.

google/cloud/iam/iam_credentials_connection.gcpcxx.pb.cc
```diff
@@ -63,36 +54,31 @@ IAMCredentialsConnection::SignJwt(
...
 }
 
-namespace {
-std::unique_ptr<IAMCredentialsRetryPolicy> DefaultRetryPolicy() {
+std::unique_ptr<IAMCredentialsRetryPolicy> DefaultIAMCredentialsRetryPolicy() {
   return IAMCredentialsLimitedTimeRetryPolicy(std::chrono::minutes(30)).clone();
 }
 
-std::unique_ptr<BackoffPolicy> DefaultBackoffPolicy() {
+std::unique_ptr<BackoffPolicy> DefaultIAMCredentialsBackoffPolicy() {
```
devjgm, line 61: These "Default" functions are not needed anymore. The body of them can just be inlined in the body of the "FooDefaultOptions()" function.

Fine to do in a followup PR if you prefer.

google/cloud/bigquery/integration_tests/big_query_read_integration_test.cc
```diff
@@ -34,15 +36,15 @@ using ::testing::Not;
...
     retry_policy_ =
         absl::make_unique<BigQueryReadLimitedErrorCountRetryPolicy>(1);
     backoff_policy_ = absl::make_unique<ExponentialBackoffPolicy>(
         std::chrono::seconds(1), std::chrono::seconds(1), 2.0);
     idempotency_policy_ = MakeDefaultBigQueryReadConnectionIdempotencyPolicy();
   }
   std::vector<std::string> ClearLogLines() { return log_.ExtractLines(); }
-  BigQueryReadConnectionOptions connection_options_;
+  Options options_;
   std::unique_ptr<BigQueryReadRetryPolicy> retry_policy_;
```
devjgm, line 48: My guess is that these `foo_policy_` members are no longer needed since that all now lives in `Options`. Actually, if these policies matter for the tests, they should actually be set in the `options_` data member too.

Fine to handle in a followup PR.

devjgm (approved):
LGTM. Thanks!

I left a few comments, but I see this is a huge PR, and it's fine with me if you'd rather address the comments in a follow-up PR, but I'll let you decide what's easiest.

PR #6022, opened 2021-03-11
refactor(spanner): add spanner_internal::MakeConnection

devjgm:
This is the function that *will be* the preferred user-facing
"MakeConnection" function. It's requires a `spanner::Database` argument,
and nothing else. A single, optional `Options` argument may be specified
to configure aspects of the returned `Connection`. Appropriate defaults
will be used for any unspecified option, so `{}` is a valid (and the
default) argument.

The other `MakeConnection` overloads will continue to exist and they
will simply forward to this one new/preferred overload. The old
overloads will NOT be deprecated, so we will not break existing callers.
However, we will clearly document all the overloads to make it clear
that the new overload should be preferred, so that customers do not
struggle to decide which overload to choose.

Of course, this is all "internal" for now, but it will likely be moved
out of internal and made public soon (before the April release).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6022)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -354,9 +354,10 @@ inline namespace SPANNER_CLIENT_NS {
 
 std::shared_ptr<spanner::Connection> MakeConnection(spanner::Database const& db,
                                                     internal::Options opts) {
-  internal::CheckExpectedOptions<internal::CommonOptions, internal::GrpcOptions,
+  internal::CheckExpectedOptions<internal::CommonOptionList,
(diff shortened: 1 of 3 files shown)


coryan (approved):
> Reviewed 5 of 5 files at r1, 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6022)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devjgm)

PR #6023, opened 2021-03-11
feat(internal): add internal-only `TypeList`

devjgm:
Also adds the ability to concatenate type lists. This makes it so we
don't need to use `std::tuple` and `std::tuple_cat` when we only care
about the types and we don't care about values. This is important
because `std::tuple_cat` breaks with large type lists on clang-3.8.

I plan to use this in `google/cloud/internal/options.h` in a following
PR to avoid breakages like this: https://source.cloud.google.com/results/invocations/f5efd65c-943c-4050-85d2-5ee41289e702/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fdocker%2Fclang-3.8-presubmit/log

Note: If/when we get rid of clang-3.8, we could easily replace this with
calls to `<tuple>` stuff, if we think that's simpler.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6023)

### DIFF
--- a/google/cloud/CMakeLists.txt
+++ b/google/cloud/CMakeLists.txt
@@ -164,6 +164,7 @@ add_library(
     internal/throw_delegate.cc
     internal/throw_delegate.h
     internal/tuple.h
+    internal/type_list.h
     internal/user_agent_prefix.cc
     internal/user_agent_prefix.h
     internal/utility.h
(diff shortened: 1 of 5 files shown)


coryan (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6023#-:-MVXf6V698gF_EUScqxU:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devjgm)

PR #6025, opened 2021-03-12
refactor: rework the Status/StatusOr<T> gmock matchers

devbww:
Because `StatusOr<T>` does not have a printer (and should not), gmock
will render its value as "N-byte object <...>", which is not very
useful. Accordingly, `StatusIs` will now print the enclosed `Status`
directly so that a failing expectation will not require further
explanation.

This means that we no longer need workarounds like ...
```
  EXPECT_THAT(v, IsOk()) << v.status();
```
or
```
  EXPECT_THAT(v.status(), IsOk())
```
just to be able to see the entire `Status<T>::status()` in a failed
expectation.

It also means we don't have to, nor want to, print a `Status` code
or message during either of their sub-matches.

Also generate an explanation even when a `StatusIs` match succeeds.
This is necessary when the `StatusIs` matcher is used as an argument
to another matcher that can change a success into a failure (e.g.,
`Not()`).  This means that ...
```
  EXPECT_THAT(v, Not(StatusIs(StatusCode::kInternal)));
```
will now describe the failing internal Status.

Finally, implement `EXPECT_STATUS_OK` using `StatusIs`. This means the
expectation output is the same for all matcher varieties.  The helper
macros are now exported by `google/cloud/testing_util/status_matchers.h`,
although `assert_ok.h` remains for backwards compatibility.


google/cloud/testing_util/assert_ok.h
```diff
@@ -15,39 +15,7 @@
...
-}  // namespace GOOGLE_CLOUD_CPP_NS
-}  // namespace cloud
-}  // namespace google
-
-#define ASSERT_STATUS_OK(val) \
-  ASSERT_PRED_FORMAT1(::google::cloud::testing_util::IsOkPredFormat, val)
-
-#define EXPECT_STATUS_OK(val) \
-  EXPECT_PRED_FORMAT1(::google::cloud::testing_util::IsOkPredFormat, val)
+// Deprecated. Use google/cloud/testing_util/status_matchers.h directly.
```
devjgm, line 18: `testing_util/` is not a public user-facing directory, so I don't think we need to bother with deprecations here. We can just clean up our code and then delete this file.

WDYT?

devjgm (approved):
Nice cleanup. Thanks!

devbww (commented):
I say let's leave it as-is in this PR.  As you note, it requires some cleanup, which I already have separately pending.  Then assert_ok* can disappear in another follow-up PR.

All that said, perhaps we should be exposing Status matchers to users.  I'll raise it as a discussion issue.

PR #6044, opened 2021-03-15
common: fix ambiguous match bazel error on macos 11 intel

vnvo2409:
On macos 11 intel, we will have the following compilation problem:
```
ERROR: Analysis of target '//google/cloud/pubsub/samples:samples' failed; build aborted: /Users/vovannghia/programming/google-cloud-cpp/google/cloud/testing_util/BUILD:43:11: Illegal ambiguous match on configurable attribute "defines" in //google/cloud/testing_util:google_cloud_cpp_testing:
@bazel_tools//src/conditions:darwin_x86_64
@bazel_tools//src/conditions:darwin
```
This PR removes `@bazel_tools//src/conditions:darwin` configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6044)

### DIFF
--- a/google/cloud/testing_util/BUILD
+++ b/google/cloud/testing_util/BUILD
@@ -56,7 +56,7 @@ cc_library(
             "GOOGLE_CLOUD_CPP_HAVE_RUSAGE_THREAD",
         ],
         (
-            "@bazel_tools//src/conditions:darwin_x86_64",
+            "@bazel_tools//src/conditions:darwin",
             "@bazel_tools//src/conditions:freebsd",
             "@bazel_tools//src/conditions:openbsd",
         ): [
--- a/google/cloud/testing_util/BUILD
+++ b/google/cloud/testing_util/BUILD
@@ -57,7 +57,6 @@ cc_library(
         ],
         (
             "@bazel_tools//src/conditions:darwin_x86_64",
-            "@bazel_tools//src/conditions:darwin",
             "@bazel_tools//src/conditions:freebsd",
             "@bazel_tools//src/conditions:openbsd",
         ): [


devjgm:
Thanks for the PR. I'm guessing we haven't tested on Big Sur, otherwise, we'd probably have caught this. I'm not an expert on this stuff, so let me ask a couple questions, please. Would removing the `conditions:darwin_x86_64` and leaving `@bazel_tools//src/conditions:darwin` also have fixed the ambiguous problem? I can understand why the condition is ambiguous, but I'm wondering which one we want to keep. Keeping `@bazel_tools//src/conditions:darwin` would seem to more closely match the other ones like `:freebsd`.

vnvo2409:
We could remove one of them and it does not make any difference. 

I remove `darwin` because we don't have a M1 machine yet and we don't know if it supports `GOOGLE_CLOUD_CPP_HAVE_GETRUSAGE`, so I think it's better to let it fall to the default condition. What do you think ?

coryan:
> We could remove one of them and it does not make any difference.
> 
> I remove `darwin` because we don't have a M1 machine yet and we don't know if it supports `GOOGLE_CLOUD_CPP_HAVE_GETRUSAGE`, so I think it's better to let it fall to the default condition. What do you think ?

`getrusage(2)` is very widely supported on Unices, regardless of the underlying CPU, let's use the more generic `@bazel_tools//src/conditions:darwin`.

vnvo2409:
> `getrusage(2)` is very widely supported on Unices, regardless of the underlying CPU, let's use the more generic `@bazel_tools//src/conditions:darwin`.

Done, please take a look !

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6044#-:-MVvONvj7Xj03I5eSP6H:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @vnvo2409)

PR #6012, opened 2021-03-09
feat(bigtable): Bind CompletionQueue with Table

dopiera:
This is a part of #2567 and a continuation of #5967 and #6004.

This change is a part of the removal of the `CompletionQueue` argument
from all `Table::Async*` calls. Instead, `CompletionQueue` will have to
be bound with the `Table` object. Current `Table::Async*` member
functions have been deprecated and overrides without the
`CompletionQueue` have been added.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6012)

PR #6004, opened 2021-03-09
feat: bind `CompletionQueue` with `TableAdmin`

dopiera:
This is a part of #2567 and a continuation of #5967.

This MR makes all synchronous operations on `TableAdmin`
execute on a `CompletionQueue` bound with that `TableAdmin`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6004)

PR #6047, opened 2021-03-16
cleanup(bigtable)!: make *Client::BackgroundThreadsFactory() pure

devbww:
Change the recently-added `BackgroundThreadsFactory()` functions (see
#5967, #6004, and #6012) in `bigtable::AdminClient`, `bigtable::DataClient`,
and `bigtable::InstanceAdminClient` to be pure virtual.

This restores these class to be pure interfaces, rather than having one
small piece of concrete behavior.  A default implementation that returns
a null factory, which throws when called, is never useful.

This may seem like a breaking change, but it was actually the introduction
of the member function that was breaking, as derived classes already had
to override the function in order to use the client (for example, to pass
a `DataClient` to the `bigtable::Table` constructor).

Failure to override will at least now be a compilation error, rather than
a runtime error (`std::bad_function_call`).

This was discovered while trying to use the latest Bigtable client in
another package, and seeing exceptions thrown.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6047)


devbww:
I'm open to addressing this in a different way, but at least this PR raises the issue.

coryan:
> I'm open to addressing this in a different way, but at least this PR raises the issue.

I think your analysis is correct.  A couple of things:

- This PR should be marked as `refactor(bigtable)!` (or maybe `cleanup(bigtable)!`) to signal it is a breaking change.  I understand the previous PRs should have been marked that way too, but we can make this one right.
- Please update `CHANGELOG.md` explaining the breaking change.
- As you probably noticed, you need to update the API baseline.

devbww:
Thanks.  Done.  PTAL.

coryan (approved):

Issue #4758, opened 2020-08-03
[flake] spanner samples or integration test failure: Duplicate name in schema


mr-salty:
Seen (at least) twice:
* 2020-07-28 (sample) [log](https://source.cloud.google.com/results/invocations/a27264ec-f166-42d7-8db2-e3cbab0815c2/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmaster%2Fdocker%2Fubsan/log)
* 2019-10-29 (integration test) - see also #4034 

The failure looks like:

```
FAIL: //google/cloud/spanner/samples:samples (see /h/.cache/bazel/_bazel_kbuilder/a7ecd8237744645c5d189c197108d6d2/execroot/com_github_googleapis_google_cloud_cpp/bazel-out/k8-fastbuild/testlogs/google/cloud/spanner/samples/samples/test.log)
INFO: From Testing //google/cloud/spanner/samples:samples:
==================== Test output for //google/cloud/spanner/samples:samples:
Running instance admin samples on test-instance-1

[...]

Running spanner_create_storing_index sample
Duplicate name in schema: AlbumsByAlbumTitle2.
================================================================================
[...]
INFO: Build completed, 1 test FAILED, 13 total actions
//google/cloud/spanner/samples:samples                                   FAILED in 51.5s
```

coryan:
I think this is another form of the same problem:

https://source.cloud.google.com/results/invocations/55596056-51c4-45fc-bfb6-9d6111f4f4ff

```
[4,194 / 4,197] 105 / 108 tests; Testing //google/cloud/bigtable/tests:data_integration_test; 86s local ... (3 actions running)
FAIL: //google/cloud/spanner/samples:samples (see C:/b/rvommcsc/execroot/com_github_googleapis_google_cloud_cpp/bazel-out/x64_windows-opt/testlogs/google/cloud/spanner/samples/samples/test.log)
INFO: From Testing //google/cloud/spanner/samples:samples:
==================== Test output for //google/cloud/spanner/samples:samples:
Running instance admin samples on test-instance-15

... ... ... snip snip snip ... ... ...
Running spanner get-database-ddl sample
Database metadata is:
statements: "CREATE TABLE Singers (\n  SingerId INT64 NOT NULL,\n  FirstName STRING(1024),\n  LastName STRING(1024),\n  SingerInfo BYTES(MAX),\n) PRIMARY KEY(SingerId)"
[...]

PR #6052, opened 2021-03-16
test: log last RPCs on integration test failures

coryan:
Part of the work for #4758

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6052)


google/cloud/testing_util/integration_test.cc
```diff
@@ -0,0 +1,30 @@
...
+#include "google/cloud/testing_util/integration_test.h"
+#include "google/cloud/log.h"
+
+namespace google {
+namespace cloud {
+inline namespace GOOGLE_CLOUD_CPP_NS {
+namespace testing_util {
+
+void IntegrationTest::TearDown() {
+  if (HasFailure()) ::google::cloud::LogSink::Instance().Flush();
```
devbww, line 24: This is part of the reason I wasn't overly keen on the idea ... this log dump may be long after the failing expectation of interest, so the relevant traces may have disappeared, or at least won't be the most recent ones.  But, that shouldn't stop us from doing what we can.

coryan: I tried to connect an event listener, but had no luck getting it to work. This looks promising:

https://github.com/google/googletest/blob/b7d472f1225c5a64943821d8483fecb469d3f382/googletest/include/gtest/gtest.h#L1126

devbww: Ack.  I failed to even find such an interface, so, yes, promising.  I'll see if I can learn anything.  Thanks,

google/cloud/testing_util/integration_test.cc
```diff
@@ -0,0 +1,30 @@
...
+
+#include "google/cloud/testing_util/integration_test.h"
+#include "google/cloud/log.h"
+
+namespace google {
+namespace cloud {
+inline namespace GOOGLE_CLOUD_CPP_NS {
+namespace testing_util {
+
+void IntegrationTest::TearDown() {
```
devbww, line 23: If we're going to be careful to call `IntegrationTest::Setup()` from the sub-class `Setup()` functions (which will just resolve to `::testing::Test::Setup()`), then we should probably call `::testing::Test::TearDown()` explicitly before this `TearDown()` override returns.

coryan: Done.

google/cloud/testing_util/integration_test.cc
```diff
@@ -0,0 +1,31 @@
...
+  if (HasFailure()) ::google::cloud::LogSink::Instance().Flush();
```
devjgm, line 24: Googletest allows setup/teardown logic in constructors and destructors, which is nice because we don't have to worry about subclasses remember to call the super class's function at the right point. If the logic that's expected to go into this `IntegrationTest` fixture would work in its c'tor or d'tor, consider using those instead. Of course, there are some technical differences that sometimes requires SetUp/TearDown, in which case, of course use the one that works :-)

https://github.com/google/googletest/blob/master/docs/faq.md#should-i-use-the-constructordestructor-of-the-test-fixture-or-setupteardown-ctorvssetup

devjgm (approved):

google/cloud/testing_util/integration_test.cc
coryan: I like that better.

devjgm (approved):

devbww (approved):

PR #6055, opened 2021-03-17
fix(pubsub): deadlocks for cancel during session startup

### LINKED ISSUE #6053: [Flake] Pub/Sub samples timeout in test
https://source.cloud.google.com/results/invocations/29229a70-3153-456e-8e9e-f55f127bafad

```
7/7 Test #207: pubsub_samples ...............................***Timeout 1500.10 sec
```

coryan:
If an application canceled a session before the first `Write()` call
completes in the startup sequence.

Changed `pubsub/samples/samples.cc` to make it easier to troubleshoot
this problem, I am keeping the change because it feels generally useful.

Fixes #6053

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6055)

### DIFF
--- a/google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
+++ b/google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
@@ -141,15 +141,21 @@ void StreamingSubscriptionBatchSource::OnInitialWrite(RetryLoopState const& rs,
     OnInitialError(std::move(rs));
     return;
   }
-  shutdown_manager_->StartOperation(__func__, "InitialRead", [&] {
-    auto weak = WeakFromThis();
-    stream_->Read().then(
-        [weak,
-         rs](future<absl::optional<google::pubsub::v1::StreamingPullResponse>>
-                 f) {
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
```diff
@@ -141,15 +141,21 @@ void StreamingSubscriptionBatchSource::OnInitialWrite(RetryLoopState const& rs,
...
+            [weak,
+             rs](future<
+                 absl::optional<google::pubsub::v1::StreamingPullResponse>>
+                     f) {
+              if (auto s = weak.lock())
+                s->OnInitialRead(std::move(rs), f.get());
+            });
+      });
+  // This is very rare, but if the session enters shutdown while the initial
+  // setup is in progress it can happen.
```
devbww, line 157: Perhaps "This is very rare, but it can happen if ...".

devbww (approved):

google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
coryan: Done.

PR #6056, opened 2021-03-17
ci: change sourced shell scripts to "modules" in /ci/kokoro/lib

devjgm:
Part of: https://github.com/googleapis/google-cloud-cpp/issues/3911

This change only changes the shell libraries used by the code in /ci/kokoro/... There are a few remaining scripts to be converted to this form after this PR.

The benefit of this change is that our scripts are sourced in a uniform way, with absolute paths, and each can declare the deps that they need rather than relying on the caller sourcing them in the correct order.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6056)


coryan (approved):
A bit over my head, but LGTM.

ci/kokoro/lib/gcloud.sh
```diff
@@ -30,38 +34,45 @@ if [[ -z "${GCLOUD_ARGS:-}" ]]; then
   )
 fi
 
+GCLOUD_BIN="/usr/local/google-cloud-sdk/bin/gcloud"
+if [ ! -x "${GCLOUD_BIN}" ]; then
+  # If the bin wasn't found at the above path, hope it's found in the path.
```
devbww, line 39: A few too many "path"s perhaps?  Maybe `s/in the path/in ${PATH}/`?

ci/kokoro/docker/dump-reports.sh
```diff
@@ -16,13 +16,7 @@
...
 set -eu
 
-if [[ -z "${PROJECT_ROOT+x}" ]]; then
-  readonly PROJECT_ROOT="$(
-    cd "$(dirname "$0")/../../.."
-    pwd
-  )"
-fi
-source "${PROJECT_ROOT}/ci/kokoro/define-docker-variables.sh"
+source module /ci/kokoro/lib/docker-variables.sh
```
devbww, line 19: Does this file need a `source "$(dirname "$0")/../../lib/init.sh"`?

ci/kokoro/lib/gcloud.sh
devjgm: Fixed.

ci/kokoro/docker/dump-reports.sh
devjgm: Yes. Good catch! I looked for more cases of this too:

```shell
$ grep -r -l 'source module ' ci/kokoro/ | xargs grep -L -P 'source .+/lib/init.sh'
ci/kokoro/install/run-installed-programs.sh
ci/kokoro/lib/cache.sh
```
The two files shown there are themselves libraries, not binaries, so they are included via `source` and the calling binary is the one responsible for sourcing `lib/init.sh`. So I think the one you found was the only one. Thanks.

devbww (approved):

Issue #6060, opened 2021-03-17
Cleanup `std::move()` of `const` values


coryan:
This may indicate we are trying to optimize something that cannot be optimized.

https://github.com/googleapis/google-cloud-cpp/blob/fcbbc055d3100b67096857f4fae9ee09aa700c79/google/cloud/storage/parallel_upload.h#L936

https://github.com/googleapis/google-cloud-cpp/blob/fcbbc055d3100b67096857f4fae9ee09aa700c79/google/cloud/iam_bindings.h#L48

https://github.com/googleapis/google-cloud-cpp/blob/fcbbc055d3100b67096857f4fae9ee09aa700c79/google/cloud/storage/parallel_upload.h#L757

https://github.com/googleapis/google-cloud-cpp/blob/fcbbc055d3100b67096857f4fae9ee09aa700c79/google/cloud/storage/parallel_upload.h#L802

https://github.com/googleapis/google-cloud-cpp/blob/fcbbc055d3100b67096857f4fae9ee09aa700c79/google/cloud/storage/parallel_upload.h#L684

devjgm:
We disable

https://github.com/googleapis/google-cloud-cpp/blob/912f4a396d3f4f8cd7d7ad4b978bbe364f9aabbc/.clang-tidy#L21-L26

Is something warning about these?

coryan:
> Is something warning about these?

Yes, see the internal email thread.

This is different (I think) from the `clang-tidy` warning.  The function is holding a `Foo const&` and then passing it to another function via `std::move()`.  That might be the right thing to do (say because you want to signal that you are no longer going to use the thing), but also might be an attempt to move the object, but that is not what happens.

remyabel:
Is there an existing document on when it's appropriate to use `std::move` in this codebase?

devjgm:
@remyabel we don't have a document like that, but it's possible one would be helpful. I've personally considered when-to-move to be more of a black-and-white (and simpler) issue than some checkers do.

In my mind, I think reasoning about about when to `move` should be able to be done locally _at the call site_, without having to care about the _implementation detail_ about how a function chooses to accept its parameter or the type of the moved object. I use the following two principles that _I think_ always work and result in correct and the most efficient move-related code (someone please tell me if I'm wrong):

[...]

PR #6064, opened 2021-03-18
fix(storage): move disabling consts

coryan:
Part of the work for #6060

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6064)

### DIFF
--- a/google/cloud/storage/parallel_upload.h
+++ b/google/cloud/storage/parallel_upload.h
@@ -79,12 +79,12 @@ struct CreateParallelUploadShards;
  *     return the value of the first element of type T
  */
 template <typename T, typename Tuple, typename Enable = void>
-struct ExtractFirstOccurenceOfTypeImpl {
+struct ExtractFirstOccurrenceOfTypeImpl {
   absl::optional<T> operator()(Tuple const&) { return absl::optional<T>(); }
 };
 
 template <typename T, typename... Options>
-struct ExtractFirstOccurenceOfTypeImpl<
+struct ExtractFirstOccurrenceOfTypeImpl<
     T, std::tuple<Options...>,
     typename std::enable_if<
         Among<typename std::decay<Options>::type...>::template TPred<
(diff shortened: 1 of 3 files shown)


google/cloud/storage/parallel_upload.h
```diff
@@ -750,9 +748,9 @@ Composer ResumableParallelUploadState::CreateComposer(
   auto get_metadata_options = StaticTupleFilter<
       Among<DestinationPredefinedAcl, EncryptionKey, KmsKeyName, QuotaUser,
             UserIp, UserProject, WithObjectMetadata>::TPred>(options);
-  auto composer =
-      [client, bucket_name, object_name, compose_options, get_metadata_options,
-       prefix](std::vector<ComposeSourceObject> const& sources) mutable
+  auto composer = [client, bucket_name, object_name, compose_options,
+                   get_metadata_options,
+                   prefix](std::vector<ComposeSourceObject> sources) mutable
```
devjgm, line 753: The `Composer` alias is defined as

https://github.com/googleapis/google-cloud-cpp/blob/7cb938b386da73a3adc088646c19aa18bd6c581f/google/cloud/storage/parallel_upload.h#L124-L127

So will changing this signature here to accept by value actually eliminate a copy somewhere?

devjgm (approved):

google/cloud/storage/parallel_upload.h
coryan: That is either terrible or hilarious... see the lambda does get assigned to a `Composer` but that goes to:


https://github.com/googleapis/google-cloud-cpp/blob/7cb938b386da73a3adc088646c19aa18bd6c581f/google/cloud/storage/parallel_upload.h#L226-L227

which has a different signature, but for extra fun it was used here:

https://github.com/googleapis/google-cloud-cpp/blob/7cb938b386da73a3adc088646c19aa18bd6c581f/google/cloud/storage/parallel_upload.cc#L253

I think all this stuff needs some deep cleaning.

PR #6066, opened 2021-03-18
fix(storage): use after move problems

coryan:
We were using objects after move-assigning from them. Found using
Clang's static analyzer.

Also cleaned up a (mostly harmless) unused assignment.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6066)

### DIFF
--- a/google/cloud/storage/lifecycle_rule.cc
+++ b/google/cloud/storage/lifecycle_rule.cc
@@ -104,7 +104,6 @@ std::ostream& operator<<(std::ostream& os, LifecycleRuleCondition const& rhs) {
   }
   if (rhs.custom_time_before.has_value()) {
     os << sep << "custom_time_before=" << *rhs.custom_time_before;
-    sep = ", ";
   }
   return os << "}";
 }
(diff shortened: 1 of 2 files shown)


google/cloud/storage/lifecycle_rule.cc
```diff
@@ -141,11 +140,12 @@ void LifecycleRule::MergeConditions(LifecycleRuleCondition& result,
...
-      std::vector<std::string> a = *std::move(result.matches_storage_class);
+      std::vector<std::string> a;
+      a.swap(*result.matches_storage_class);
       std::sort(a.begin(), a.end());
       std::vector<std::string> b = *rhs.matches_storage_class;
       std::sort(b.begin(), b.end());
+      std::vector<std::string> tmp;
       std::set_intersection(a.begin(), a.end(), b.begin(), b.end(),
                             std::back_inserter(tmp));
       result.matches_storage_class.emplace(std::move(tmp));
```
devjgm, line 151: The previous code w/ the move looked a bit more idiomatic, and likely how we'd write this if we wrote something similar again tomorrow. The only issue was re-using the moved-from `absl::optional` on line 151. I'd imagine another fix would be the previous code exactly as it was, but change line 151:

```diff
- result.matches_storage_class.emplace(std::move(tmp));
+ result.matches_storage_class = std::move(tmp);
```

Though any fix is fine w/ me, and I'll defer to your sensibilities.

devjgm (approved):

PR #6080, opened 2021-03-20
ci: add basic Cloud Build files

devjgm:
This PR adds basic configs for using Google Cloud Build with this repo. This PR does not remove or change any of our existing CI builds. The purpose of this is to experiment with simple Cloud Build builds to see if they seem valuable for this repo.

I copied the `ci/kokoro/docker/Dockerfile.fedora-install` -> `ci/cloudbuild/Dockerfile.fedora` (note no "-install"), and this is the only docker image that the build's _currently_ use, though we can/will add more. Other than copying an existing `Dockerfile` I have not copied other build script logic. Instead, I'm seeing if we start fresh can we end up with a simpler build system without so much complexity embedded in our build scripts.

## Overview

`ci/cloudbuild` contains all the files. The `cloudbuild.yaml` is the only config file for Cloud Build. It builds the specified docker image with kaniko, then invokes the `ci/cloudbuild/build.sh <buildname>` script to kick off the specified build, which is implemented by a _simple_ (the goal is to keep these focused and simple) script in `ci/cloudbuild/builds/<buildname>.sh`.

The `ci/cloudbuild/build.sh` script is invoked from within a Cloud Build docker image, AND it can be invoked directly by a user from the shell in order to (1) submit builds to Cloud Build, or (2) run builds locally on the local machine (or within the user's specified Docker image).

## Setup

The Cloud Builds run as the invoking user's default project. In order to change the project, the user would need to use `gcloud` commands to select a different project. 

## Next steps

1. Configure some triggers in the Cloud Build console to automatically run these builds pre-/post- submit
2. Add more builds, to cover things like publishing docs, code coverage, etc.
3. Add *integration test* builds.
4. Add *install* / "quickstart" builds.
5. Add some caching. Bazel builds should use the bazel remote cache. CMake builds should do something with caching the ccache data.
6. Figure out if there are build artifacts that we want to save from each build. Save them.
7. Figure out how best to run these builds _locally_. The more logic we put in `cloudbuild.yaml`, the more difficult it may be to run builds locally. Though we can try `cloud-build-local`, which may make this easier.
8. More stuff that I'm not currently thinking of.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6080)


ci/cloudbuild/build.sh
```diff
@@ -0,0 +1,94 @@
...
+#
+#   Build names:
+
+set -euo pipefail
+
+source "$(dirname "$0")/../lib/init.sh"
+source module ci/lib/io.sh
+cd "${PROJECT_ROOT}"
+
+# Use getop to parse and normalize all the args.
```
remyabel, line 40: s/getop/getopt/

ci/cloudbuild/build.sh
```diff
@@ -0,0 +1,94 @@
...
+  --)
+    shift
+    break
+    ;;
+  esac
+done
+readonly DISTRO
+
+if (($# == 0)); then
+  # Extracts the usage from the file comment starting at line 17.
```
remyabel, line 66: Would it be better to just add a `print_usage` function?

devjgm: Done. Called from two more places now too.

ci/cloudbuild/build.sh
devjgm: Fixed.

> candidly, I do not know if either is more popular, I just noticed that CLion, and VSCode, and I think github treat `.Dockerfile` as an extension

Done. Changed to `<distro>.Dockerfile`

ci/cloudbuild/builds/cmake+gcc.sh
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+export CC=gcc
+export CXX=g++
+
+cmake -GNinja -S . -B cmake-out
```
coryan, line 22: For a future PR, the current CI scripts use a different binary output directory for each build (i.e. a different `-B` flag).  That allows us to run `./ci/kokoro/docker/build.sh asan` and `./ci/kokoro/docker/build.sh clang-tidy` in the same workspace, without mixing `.o` files from each build.
You may be able to keep these scripts simple by hiding the mount directory through some clever use of `--volume`.

ci/cloudbuild/cloudbuild.yaml
```diff
@@ -0,0 +1,45 @@
...
+  args: [
+    '-c',
+    'docker pull us-central1-docker.pkg.dev/$PROJECT_ID/google-cloud-cpp-cloudbuild-docker/${_DISTRO}-image:tag1 || exit 0'
+  ]
+- name: 'gcr.io/cloud-builders/docker'
+  args: [
+    'build',
+    '-f', 'ci/cloudbuild/${_DISTRO}.Dockerfile',
+    '-t', 'us-central1-docker.pkg.dev/$PROJECT_ID/google-cloud-cpp-cloudbuild-docker/${_DISTRO}-image:tag1',
+    '--cache-from', 'us-central1-docker.pkg.dev/$PROJECT_ID/google-cloud-cpp-cloudbuild-docker/${_DISTRO}-image:tag1',
```
coryan, line 37: ah... I see, you are caching from the previous push... neat.

Note how a separate PR that changes `fedora.Dockerfile` will dirty up your cache.  I think everything works, just maybe not as well as it could.

coryan (approved):

ci/cloudbuild/cloudbuild.yaml
devjgm: Yes, I'm attempting to do some caching of the docker image. I'd like to switch to Kaniko, though I cannot figure out why it doesn't work. I filed https://github.com/GoogleContainerTools/kaniko/issues/1604 about that.

Maybe what we want is something like post-submit builds push built images. Pre-submit PR builds pull images and `--cache-from` them, but then do not push their images.

coryan: Ack, it might be worth trying `:edge` for Kaniko instead of `:latest`.

devjgm: Huh. I'd tried that before and it didn't work. However, I tried again today with `:edge` AND I changed to `gcr.io` instead of using the artifact registry and it seems to work. So, cool!

coryan (approved):

PR #6084, opened 2021-03-21
feat(ci): add caching to GCB builds

devjgm:
Adds a `ci/cloudbuild/cache.sh` script that is responsible for `save`'ing and `restore`'ing our build caches to/from GCS.
There are also two new build steps in the `cloudbuild.yaml` file that call this script at the appropriate times.

NOTE: The `~/.cache/bazel` directory is not currently cached. It was huge and it wasn't clear if it was beneficial. I'll look into
fixing that later, or just using the normal remote build cache for bazel.

This PR also creates a home directory mount of `/h` that's visible to all build steps so they can all read/write to the homedir.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6084)

### DIFF
--- a/ci/cloudbuild/cache.sh
+++ b/ci/cloudbuild/cache.sh
@@ -1,0 +1,145 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This script saves a cache tarball to GCS, and restores one from GCS. It is
+# intended to be called from a build step in the `cloudbuild.yaml` file. It
+# executes in the context of build. Typically the "restore cache" step will be
+# called in one step, followed by the actual build of the code, followed by the
+# "save cache" step.
+#
+# Usage: ./cache <save|restore>
+#
+#   Options:
+#     --bucket_url=<url>    URL to the GCS bucket where the cache should be saved
+#     --key=<key>           The primary key to use when caching
+#     --fallback_key=<key>  If the primary key isn't found when restoring, try this key
+#     --path=<path>         A path to be backed up. May be specified multiple times.
+#     -h|--help             Print this help message
+
+set -euo pipefail
+
+source "$(dirname "$0")/../lib/init.sh"
+source module ci/lib/io.sh
+cd "${PROJECT_ROOT}"
+
+function print_usage() {
+  # Extracts the usage from the file comment starting at line 17.
+  sed -n '17,/^$/s/^# \?//p' "${PROGRAM_PATH}"
+}
+
+# Use getopt to parse and normalize all the args.
+PARSED="$(getopt -a \
+  --options="h" \
+  --longoptions="bucket_url:,key:,fallback_key:,path:,help" \
+  --name="${PROGRAM_NAME}" \
+  -- "$@")"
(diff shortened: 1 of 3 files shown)


ci/cloudbuild/cloudbuild.yaml
```diff
@@ -15,21 +15,51 @@
...
+- name: 'gcr.io/${PROJECT_ID}/google-cloud-cpp-cloudbuild-docker/${_DISTRO}-image:${BUILD_ID}'
   entrypoint: 'ci/cloudbuild/build.sh'
   args: [ '${_BUILD_NAME}' ]
+
+  # Caches the homedir, /h, in the specified GCS bucket.
+- name: 'gcr.io/cloud-builders/gsutil'
+  entrypoint: 'bash'
+  args: [
+    '-c',
+    'gsutil -m rsync -r -c -e /h "gs://${_CACHE_BUCKET}/${PROJECT_ID}/${_DISTRO}-${_BUILD_NAME}/h"'
```
coryan, line 64: Consider `-q` to make the logs less verbose.

ci/cloudbuild/cloudbuild.yaml
```diff
@@ -15,21 +15,51 @@
...
+
+  # Runs the specified build inside the docker image that was created in the
+  # first step.
+- name: 'gcr.io/${PROJECT_ID}/google-cloud-cpp-cloudbuild-docker/${_DISTRO}-image:${BUILD_ID}'
   entrypoint: 'ci/cloudbuild/build.sh'
   args: [ '${_BUILD_NAME}' ]
+
+  # Caches the homedir, /h, in the specified GCS bucket.
+- name: 'gcr.io/cloud-builders/gsutil'
+  entrypoint: 'bash'
```
coryan, line 61: What is the default entrypoint for `gcr.io/cloud-builders/gsutil`?  I would expect it to be `gsutil` and this can become:

```
- name: 'grc.io/cloud-builders/gsutil'
  args: [ '-m', '-q', 'rsync', '-r', '-c', '-e', '/h', 'gs://.....']
```

devjgm: yes, `gsutil` is the entrypoint. I was using the `bash -c` version because it above I needed to always `exit 0`, and down here it shortened the command quite a bit. I've now changed things around a bit more. I could still change this second usage of `gsutil` to pass an array to `args:` if you think that's preferable.

ci/cloudbuild/cloudbuild.yaml
```diff
@@ -15,21 +15,51 @@
 options:
   machineType: 'N1_HIGHCPU_32'
   diskSizeGb: '512'
+  env: [ 'HOME=/h' ]
```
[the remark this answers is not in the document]
devjgm: GCB made this part quite nice.

ci/cloudbuild/cloudbuild.yaml
devjgm: done.

ci/cloudbuild/cloudbuild.yaml
coryan: Ack. Sounds good as it is.

coryan (approved):

ci/cloudbuild/cache.sh
```diff
@@ -0,0 +1,144 @@
...
+  gsutil -m cp cache.tar.gz "${url}"
+}
+
+function restore_cache() {
+  for url in "${PRIMARY_CACHE_URL}" "${FALLBACK_CACHE_URL}"; do
+    if gsutil stat "${url}"; then
+      io::log "Fetching cache url ${url}"
+      gsutil -m cp "${url}" cache.tar.gz || continue
+      io::log "  Cache size: $(du -sh cache.tar.gz)"
+      tar -zxf cache.tar.gz && break
```
coryan, line 125: FWIW, `gsutil cp` supports `-` as a source or destination, so you could do `gsutil cp "${url}" - | tar -zxf -`

coryan (approved):

ci/cloudbuild/cache.sh
devjgm: Thanks. Done.

PR #6087, opened 2021-03-22
feature(spanner): customer-managed encryption

devbww:
`DatabaseAdminClient::CreateDatabase()`, `CreateBackup()`, and
`RestoreDatabase()` now take a `google::cloud::spanner::EncryptionConfig`
argument to control the encryption of the resulting database or backup.

For backwards compatibility the `EncryptionConfig` argument defaults to
`DefaultEncryption`, which means either (1) create the database using
Google encryption, (2) create the backup using the encryption of the
source database, or (3) restore the database using the encryption of
the source backup.

Adds samples for tags: `spanner_create_database_with_encryption_key`
`spanner_create_backup_with_encryption_key` and
`spanner_restore_backup_with_encryption_key`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6087)


scotthart (approved):
> Reviewed 11 of 11 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6087)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devbww)

google/cloud/spanner/database_admin_client.h
```diff
@@ -226,14 +230,17 @@ class DatabaseAdminClient {
...
    * @return A `google::cloud::future` that becomes satisfied when the operation
    *   completes on the service. Note that this can take minutes in some cases.
    *
    * @par Example
    * @snippet samples.cc restore-database
    */
   future<StatusOr<google::spanner::admin::database::v1::Database>>
-  RestoreDatabase(Database db, Backup const& backup);
+  RestoreDatabase(Database db, Backup const& backup,
+                  EncryptionConfig encryption_config = DefaultEncryption());
```
devjgm, line 243: You probably know this and decided otherwise, but I'll say that default constructing a variant constructs a value-initialized instance of its first member, which is a `DefaultEncryption`. It's good that you arranged things this way. That way, `EncryptionConfig x = {}`, does the right thing and is equivalent to `EncryptionConfig x = DefaultEncryption{}`. So these arguments could be written either way. I'll let you choose the one you prefer. I'm happy either way.

devjgm (approved):

google/cloud/spanner/database_admin_client.h
devbww: Yes ... I thought this way made it clearer to the peruser of the declaration exactly which behavior they would get in the defaulted case.

devbww:
There have been some second thoughts about what should (not) be included in the samples, but I'll address that post-merge.

PR #6092, opened 2021-03-23
ci: add yaml files for current triggers

devjgm:
The list of triggers and yaml files will grow, but for now, I'm just
copying what currently exists in the UI.

This PR also adds a script, `trigger.sh`, to make it easier for us to
manipulate these trigger yaml files.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6092)

### DIFF
--- a/ci/cloudbuild/trigger.sh
+++ b/ci/cloudbuild/trigger.sh
@@ -65,27 +65,20 @@ PARSED="$(getopt -a \
   -- "$@")"
 eval set -- "${PARSED}"
 
-while true; do
-  case "$1" in
-  -d | --describe)
-    describe_trigger "$2"
-    exit
-    ;;
-  -i | --import)
-    import_trigger "$2"
-    exit
-    ;;
-  -l | --list)
-    list_triggers
-    exit
-    ;;
-  -h | --help)
-    print_usage
-    exit
-    ;;
-  *)
-    print_usage
-    exit
-    ;;
-  esac
-done
+case "$1" in
+-d | --describe)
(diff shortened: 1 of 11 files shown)


ci/cloudbuild/trigger.sh
```diff
@@ -0,0 +1,91 @@
...
+
+# Use getopt to parse and normalize all the args.
+PARSED="$(getopt -a \
+  --options="d:i:lh" \
+  --longoptions="describe:,import:,list,help" \
+  --name="${PROGRAM_NAME}" \
+  -- "$@")"
+eval set -- "${PARSED}"
+
+while true; do
```
devbww, line 68: Why a loop?

devjgm: Just left over from when I thought I was going to handle multiple flags at once. But I can add the loop later if it's needed. removed for now.

ci/cloudbuild/triggers/clang-tidy.yaml
```diff
@@ -0,0 +1,20 @@
+---
+createTime: '2021-03-20T22:56:56.896890729Z'
+disabled: true
```
devbww, line 3: I assume this is still intentional.

devbww (approved):

ci/cloudbuild/triggers/clang-tidy.yaml
devjgm: yes it is intentional.

PR #6096, opened 2021-03-23
feat(storage): self-signed JWTs for service accounts

### LINKED ISSUE #6073: Use self-signed JWT for service account authentication (with JSON files)
For `google/cloud/storage/oauth2` implement self-signed JWTs as described in:

https://google.aip.dev/auth/4111

coryan:
Fixes #6073.

This is interesting because:

- It is simpler, no remote RPCs to create the bearer token.
- It removes a remote HTTPS request from the authorization path, making the first request a bit faster.
- It works even if `https://oauth2.googleapis.com/token` is having an outage.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6096)


devjgm (approved):
Nice!

PR #6099, opened 2021-03-23
ci: add cmake-install GCB build

devjgm:
Adds a `cmake-install` GCB build that compiles and installs our repo following the instructions at https://github.com/googleapis/google-cloud-cpp/blob/master/doc/packaging.md#installing-google-cloud-cpp (pretty closely). It then validates the installed artifacts:

* It looks at the installed directory structure.
* It looks at the file extensions of the installed files
* It validates all installed pkg-config files with `pkg-config --validate <file>`

Any error breaks the build, but it tries to continue to display as many errors as possible.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6099)

### DIFF
--- a/ci/cloudbuild/builds/cmake-install.sh
+++ b/ci/cloudbuild/builds/cmake-install.sh
@@ -1,0 +1,145 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


coryan (approved):
LGTM.

Will there be a separate build to test the components?

devjgm:
> LGTM.
> 
> Will there be a separate build to test the components?

There could be. What do you think? Should we test and verify `cmake --build x --target install` AND the component installs? Or just one (which one)? WDYT?

I'm going to check this one in for now to prevent new pkg-config issues, but we can change it.

coryan:
> > LGTM.
> > Will there be a separate build to test the components?
> 
> There could be. What do you think? Should we test and verify `cmake --build x --target install` AND the component installs? Or just one (which one)? WDYT?

It is all coming back to me.  There is a separate build to test the component installs:

https://github.com/googleapis/google-cloud-cpp/blob/f86bdde68f95c7c3cd278d0273f3086ef5fa11d7/ci/kokoro/install/Dockerfile.fedora-components#L96-L111

In other words, this build is good, we don't need to change it.  When/If we move the `ci/kokoro/install` builds we need to retake this conversation.

PR #6104, opened 2021-03-24
chore: building blocks for patching googleapis after download

devbww:
When developing a new feature based on simple changes to the
googleapis repo, it may well be easier to apply a patch to the
downloaded proto files rather than dealing with branches.

Those googleapis branches are often not kept up to date, so it
can be difficult to rebase branches in this repo, which hinders
developing at head. Patching also allows for the possibility of
layering changes from multiple googleapis branches.

This PR doesn't change any behavior, but it does leave indicators
for where patches can be applied in both the CMake and bazel
builds, as well as adding the "patch" binary to the docker images.

I recently used this mechanism to check whether a proposed proto
rearrangement would affect the build for the Cloud Spanner client.
An additional lesson from that effort was that any changes to the
shape of the proto tree made by the patch, which would require edits
to the list of proto files in external/googleapis/CMakeLists.txt,
would also necessitate patching the downloaded bazel BUILD files.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6104)

### DIFF
--- a/external/googleapis/CMakeLists.txt
+++ b/external/googleapis/CMakeLists.txt
@@ -157,18 +157,22 @@ foreach (proto ${EXTERNAL_GOOGLEAPIS_PROTO_FILES})
 endforeach ()
 
 include(ExternalProject)
+
 ExternalProject_Add(
     googleapis_download
     EXCLUDE_FROM_ALL ON
     PREFIX "${CMAKE_BINARY_DIR}/external/googleapis"
     URL ${GOOGLE_CLOUD_CPP_GOOGLEAPIS_URL}
     URL_HASH SHA256=${GOOGLE_CLOUD_CPP_GOOGLEAPIS_SHA256}
-             # Scaffolding for patching googleapis after download. For example:
-             # PATCH_COMMAND patch -p1 --input=/v/external/googleapis.patch
-             # NOTE: This should only be used while developing with a new
-             # protobuf message. No changes to `PATCH_COMMAND` should ever be
-             # committed to the main branch.
-    PATCH_COMMAND ""
+    PATCH_COMMAND
+        ""
+        # ~~~
+        # Scaffolding for patching googleapis after download. For example:
+        #   PATCH_COMMAND patch -p1 --input=/v/external/googleapis.patch
+        # NOTE: This should only be used while developing with a new
+        # protobuf message. No changes to `PATCH_COMMAND` should ever be
+        # committed to the main branch.
+        # ~~~
     CONFIGURE_COMMAND ""
     BUILD_COMMAND ""
     INSTALL_COMMAND ""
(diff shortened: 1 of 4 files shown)


coryan:
Hey, sorry for the delay.  I see how this is useful.  I also see how this is a bit dangerous, it makes **our HEAD** use protos that do not match production. Do you mind if I mull this until tomorrow?  If other folks think I am overly cautious then we should probably merge.

devbww:
Let me clarify that I don't think any patches should ever be checked-in.  I could certainly add comments to that effect.

This is only intended for early development on unreleased proto messages.  The thought was that leaving the breadcrumbs for where to apply patches, and to already have the patch tool in the image, would just make such development a little easier.

coryan:
Ah, yes, a couple of comments saying "never commit these changes: would help.

devbww:
Thanks.  Added comments.  PTAL.

devbww:
> Thanks. Added comments. PTAL.

It looks like the formatter likes to (further) indent the `CMakeLists.txt` comments.  Should we try to avoid that?

devjgm:
> > Thanks. Added comments. PTAL.
> 
> It looks like the formatter likes to (further) indent the `CMakeLists.txt` comments. Should we try to avoid that?

Yikes. Maybe move the comment above and attache it to the `ExternalProject_Add` command.

coryan:
Sometimes you can surround a block with   `# ~~~` to disable the formatter .

coryan (approved):

devbww:
> Sometimes you can surround a block with `# ~~~` to disable the formatter .

Thanks.  Those fences do disable re-flowing of the comment (which I need for the example anyway), but they do not affect the chosen indentation level.

It looks like it assumes the comment is related to the previous line, so it is indented to the level of that keyword's arguments (in this case, the `URL_HASH`).  So moving the comment down one line does something reasonable ... aligning with the `PATCH_COMMAND` argument.

external/googleapis/CMakeLists.txt
```diff
@@ -157,18 +157,22 @@ foreach (proto ${EXTERNAL_GOOGLEAPIS_PROTO_FILES})
...
     PREFIX "${CMAKE_BINARY_DIR}/external/googleapis"
     URL ${GOOGLE_CLOUD_CPP_GOOGLEAPIS_URL}
     URL_HASH SHA256=${GOOGLE_CLOUD_CPP_GOOGLEAPIS_SHA256}
-             # Scaffolding for patching googleapis after download. For example:
-             # PATCH_COMMAND patch -p1 --input=/v/external/googleapis.patch
-             # NOTE: This should only be used while developing with a new
-             # protobuf message. No changes to `PATCH_COMMAND` should ever be
-             # committed to the main branch.
-    PATCH_COMMAND ""
+    PATCH_COMMAND
```
coryan, line 167: FYI, the formatting was probably confused because `PATCH_COMMAND` is not a known argument, you can add them here:

https://github.com/googleapis/google-cloud-cpp/blob/67de7b3da3c4f52f25fa24d693854c7945792b22/.cmake-format.py#L20-L47

coryan (approved):

external/googleapis/CMakeLists.txt
devbww: It looks like `PATCH_COMMAND` is there.

PR #6107, opened 2021-03-25
chore: add patch to Dockerfile.fedora-libcxx-msan

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6107)


devbww:
I'm not sure how I missed this in #6104, but I did switch between the Bazel-native patch implementation and the external patch command at some point.  I'll blame that.

devjgm (approved):

PR #6111, opened 2021-03-25
ci: garbage collect docker images in GCB

devjgm:
This will delete the BUILD_ID-specific docker image at the end of a
successful build. It will also clean up 4-week old images that were left
around from failed builds that didn't get to the cleanup step.

This PR also changes the image name to something nicer and shorter, and this will let us manually
clean up all the old images that are now lying around.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6111)

### DIFF
--- a/ci/cloudbuild/cloudbuild.yaml
+++ b/ci/cloudbuild/cloudbuild.yaml
@@ -28,7 +28,7 @@ substitutions:
   _BUILD_NAME: 'unknown'
   _CACHE_BUCKET: '${PROJECT_ID}_cloudbuild'
   _CACHE_TYPE: '${_PR_NUMBER:-main}'
-  _IMAGE: '${PROJECT_ID}_cloudbuild/${_DISTRO}'
+  _IMAGE: 'cloudbuild/${_DISTRO}'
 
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/cloudbuild.yaml
```diff
@@ -38,7 +39,7 @@ steps:
     '--context=dir:///workspace/ci',
     '--dockerfile=ci/cloudbuild/${_DISTRO}.Dockerfile',
     '--cache=true',
-    '--destination=gcr.io/${PROJECT_ID}/google-cloud-cpp-cloudbuild-docker/${_DISTRO}-image:${BUILD_ID}',
+    '--destination=gcr.io/${PROJECT_ID}/${_IMAGE}:${BUILD_ID}',
```
coryan, line 42: So the name of the image already includes the project?  Any reason to repeat it above?

devjgm: I think not. Changed to simply "cloudbuild".

coryan (approved):

PR #6112, opened 2021-03-25
ci: add integration tests to GCB

devjgm:
This runs most integration tests in Google Cloud Build using bazel. They
use emulators when possible. Some tests are not yet enabled because they
require a bit more work, but there are TODOs mentioning what's missing.

Note that these integration tests do not require any special service account files.
The cloud build account has the needed access rights to run these. And team members can run
these from their workstation, again without needing service account files, with a command like:

```
./ci/cloudbuild/build.sh --project cloud-cpp-testing-resources integration
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6112)

### DIFF
--- a/ci/cloudbuild/builds/integration.sh
+++ b/ci/cloudbuild/builds/integration.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+source "$(dirname "$0")/../../lib/init.sh"
+source module ci/cloudbuild/builds/lib/bazel.sh
+source module ci/etc/integration-tests-config.sh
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/builds/integration.sh
```diff
@@ -0,0 +1,90 @@
...
+args+=(
+  "--test_timeout=600"
+  "--test_env=GOOGLE_CLOUD_PROJECT=${GOOGLE_CLOUD_PROJECT}"
+  "--test_env=GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES=yes"
+  "--test_env=GOOGLE_CLOUD_CPP_EXPERIMENTAL_LOG_CONFIG=lastN,100,WARNING"
+  "--test_env=GOOGLE_CLOUD_CPP_ENABLE_TRACING=rpc"
+  "--test_env=CLOUD_STORAGE_ENABLE_TRACING=rpc"
+)
+
+# TODO(#6083): Enable storage integration tests. The difficulty with these is
```
coryan, line 42: nit: for a future PR, maybe enable all the GCS integration tests *except* the ones that use these key files? I think there are only two or three.

coryan (approved):

ci/cloudbuild/builds/integration.sh
devjgm: I was going to do that, but the way the way that `bazel_test_args` is used in [run_integration_tests_emulator_bazel.sh](https://github.com/googleapis/google-cloud-cpp/blob/976a96397d0bb2ef3b8ba22f29acb3e187684a01/google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh#L42) I cannot pass `-//foo/bar/...`. So this script will need to be refactored a bit first, then I'll be able to enable those tests. Just planned to do that in a separate PR

PR #6116, opened 2021-03-25
feature(bigquery): add quickstart

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6116)

### DIFF
--- a/google/cloud/bigquery/quickstart/quickstart.cc
+++ b/google/cloud/bigquery/quickstart/quickstart.cc
@@ -1,0 +1,74 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! [START bigquerystorage_quickstart]
+#include "google/cloud/bigquery/bigquery_read_client.gcpcxx.pb.h"
+#include <iostream>
+#include <stdexcept>
+
+namespace {
+void ProcessRowsInAvroFormat(
+    ::google::cloud::bigquery::storage::v1::AvroSchema const&,
+    ::google::cloud::bigquery::storage::v1::AvroRows const&) {
+  // Code to deserialize avro rows should be added here.
+}
+}  // namespace
+
+int main(int argc, char* argv[]) try {
+  if (argc != 3) {
+    std::cerr << "Usage: " << argv[0] << " <project-id> <table>\n";
+    return 1;
+  }
+
+  // project_id should be in the format "projects/<your-gcp-project>"
+  std::string const project_id = argv[1];
+  // table_id should be in the format:
+  // "projects/<project-table-resides-in>/datasets/<dataset-table_resides-in>/tables/<table
+  // name>" The project values in project_id and table_id do not have to be
+  // identical.
+  std::string const table_id = argv[2];
+
+  // Create a namespace alias to make the code easier to read.
+  namespace bigquery = google::cloud::bigquery;
+  constexpr int kMaxReadStreams = 1;
+  // Create the ReadSession.
+  auto client =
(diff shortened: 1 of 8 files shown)


devjgm:
I think this PR needs to be rebased. Sorry about that.

google/cloud/bigquery/quickstart/quickstart.cc
```diff
@@ -0,0 +1,74 @@
...
+}
+}  // namespace
+
+int main(int argc, char* argv[]) try {
+  if (argc != 3) {
+    std::cerr << "Usage: " << argv[0] << " <project-id> <table>\n";
+    return 1;
+  }
+
+  // project_id should be in the format "projects/<your-gcp-project>"
```
coryan, line 34: For other resources (Bigtable tables, Spanner databases, etc.) we call the fully qualified path `$resource name` (e.g. table name, database name), and the unqualified id, `$resource id` (e.g. table id, database id).

In other words, consider `std::string const project_name = argv[1];` and change the comments accordingly.

coryan (commented):
Do you want this to be compiled and tested in one of the integration tests?  Right now it is not, the quickstart guides usually have their own build steps.  Chat with @devjgm about whether you should include this *only* in the GCB builds or if we want to add them to the other builds too.

Note that there are separate `quickstart` scripts for  CMake and Bazel on each of Windows, macOS and Linux.

devjgm:
> Do you want this to be compiled and tested in one of the integration tests? Right now it is not, the quickstart guides usually have their own build steps. Chat with @devjgm about whether you should include this _only_ in the GCB builds or if we want to add them to the other builds too.
> 
> Note that there are separate `quickstart` scripts for CMake and Bazel on each of Windows, macOS and Linux.

Happy to chat. FTR, for GCB-only builds, integration tests can go in https://github.com/googleapis/google-cloud-cpp/blob/master/ci/cloudbuild/builds/integration.sh

I don't have quickstarts running yet, but they are just about next on my list, and I expect to have those running in soon. However, I wouldn't want to block @scotthart. If Scott thinks that doing something w/ GCB would be _helpful_ for him, that's great. Otherwise, if doing things all in kokoro is faster for him, that's fine too.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6116)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigquery/quickstart/quickstart.cc, line 34 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6116#-MWf2QGbKoPKDiRe_6uD-r1-34:-MWjDp0n6MSLs9E7x6mR:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9624b73109cfbc781690994db74168c64c78e5cf/google/cloud/bigquery/quickstart/quickstart.cc#L34)):*

For other resources (Bigtable tables, Spanner databases, etc.) we call the fully qualified path `$resource name` (e.g. table name, database name), and the unqualified id, `$resource id` (e.g. table id, database id).

In other words, consider `std::string const project_name = argv[1];` and change the comments accordingly.

</blockquote></details>

Done.
___
*[google/cloud/bigquery/quickstart/quickstart.cc, line 53 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6116#-MWf2QGbKoPKDiRe_6uD-r1-53:-MWjDqcKETktehlnGPkV:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9624b73109cfbc781690994db74168c64c78e5cf/google/cloud/bigquery/quickstart/quickstart.cc#L53)):*

The project is hard-coded, you probably want to use `argv[1]`.
</blockquote></details>

Done.

coryan (commented):
I think you need to ignore this directory here:

https://github.com/googleapis/google-cloud-cpp/blob/1ee16f3c9a4f88c32e0acc1360ae7365dab420d1/.bazelignore#L9

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6116#-:-MWjE8AkDs-YU_5XDeSy:b-v36c3i)* status: 2 of 7 files reviewed, all discussions resolved (waiting on @scotthart)

coryan (approved):
> Reviewed 5 of 7 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6116)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

PR #6121, opened 2021-03-26
ci(gcb): distro can now be looked up from trigger file

devjgm:
This makes it easier to run a build because `--distro` need not be
specified. Example:

```
build.sh asan  # Runs build in GCB using `fedora`
build.sh --local asan  # Runs build on local machine; distro unnecessary
```

I also removed the `-d` shorthand for distro since it will likely be
rarely used, and foresee wanting to use that short flag to mean
`--docker` (upcoming).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6121)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -16,18 +16,18 @@
 #
 # ## Run builds using Google Cloud Build ##
 #
-# This script runs builds using Google Cloud Build. It expects a single
-# argument of the build name to run. Valid build names are the basenames of the
-# scripts in the `builds/` directory. An optional `--distro=<distro>` flag may
-# be specified to indicate which `<distro>.Dockerfile` to run the named build
-# in, and this build is submitted to Cloud Build to run. A distro of "local"
-# (the default) indicates to run the build locally, without docker, using the
-# invoking user's operating system and environment.
+# This script runs the builds defined in `ci/cloudbuild/builds/` on the local
+# machine (if `--local` is specified) or in the cloud using Google Cloud Build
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/build.sh
```diff
@@ -94,33 +94,43 @@ if [[ "${LOCAL_BUILD}" = "true" ]]; then
...
-    exit 1
-  elif [ ! -x "ci/cloudbuild/builds/${BUILD_NAME}.sh" ]; then
-    echo "Unknown build name: ${BUILD_NAME}"
+  exit
+fi
+
+# If --distro wasn't specified, look it up from the build's trigger file.
+if [[ -z "${DISTRO}" ]]; then
+  trigger_file="ci/cloudbuild/triggers/${BUILD_NAME}.yaml"
+  DISTRO="$(grep _DISTRO "${trigger_file}" | awk '{print $2}' || true)"
```
coryan, line 103: nit: (maybe for a future PR) we should consider something like https://kislyuk.github.io/yq/ instead of parsing the YAML file with grep.

coryan (approved):

ci/cloudbuild/build.sh
devjgm: Thanks. I did consider `yq` already, but it's a dep that most of us likely don't have on our machines, and it's not just one `apt install` away. If we ever need to do non-trivial parsing of yaml, then I think it'd be worth using something like `yq`, but until then, I suspect this is probably OK.... well, that was my thinking anyway :man_shrugging: :smile:

PR #6123, opened 2021-03-27
ci(gcb): print info about machine doing local build

devjgm:
These builds can run in a variety of environments, in Google Cloud
Build, a local docker, or on the local machine. It can be useful to
see some stats about the local machine runing the build.

Example output from a cloud build run:

```
Step #2: ====================
Step #2: |   Machine Info   |
Step #2: ====================
Step #2:     local: Sat Mar 27 17:32:03 UTC 2021 (+0000)
Step #2:       utc: Sat Mar 27 17:32:03 UTC 2021
Step #2:    kernel: #41~18.04.1-Ubuntu SMP Fri Feb 26 22:23:13 UTC 2021
Step #2:        os: PRETTY_NAME="Fedora 33 (Container Image)"
Step #2:     nproc: 32
Step #2:       mem: 28.21 GiB
Step #2:       gcc: gcc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)
Step #2:     clang: clang version 11.0.0 (Fedora 11.0.0-2.fc33)
Step #2:        cc: cc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)
Step #2:     cmake: cmake version 3.19.7
Step #2:     bazel: bazel 3.5.0
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6123)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -116,6 +116,21 @@ if [[ "${LOCAL_BUILD}" = "true" ]]; then
     print_usage
     exit 1
   fi
+  function mem_total() {
+    awk '$1 == "MemTotal:" {printf "%0.2f GiB", $2/1024/1024}' /proc/meminfo
+  }
+  io::log_h1 "Machine Info"
+  TZ=/etc/localtime printf "%10s %s (%s)\n" "local:" "$(date)" "$(date +%z)"
+  printf "%10s %s\n" "utc:" "$(date -u)"
+  printf "%10s %s\n" "kernel:" "$(uname -v)"
+  printf "%10s %s\n" "os:" "$(grep PRETTY_NAME /etc/os-release)"
+  printf "%10s %s\n" "nproc:" "$(nproc)"
+  printf "%10s %s\n" "mem:" "$(mem_total)"
+  printf "%10s %s\n" "gcc:" "$(gcc --version 2>&1 | head -1)"
+  printf "%10s %s\n" "clang:" "$(clang --version 2>&1 | head -1)"
+  printf "%10s %s\n" "cc:" "$(cc --version 2>&1 | head -1)"
+  printf "%10s %s\n" "cmake:" "$(cmake --version 2>&1 | head -1)"
+  printf "%10s %s\n" "bazel:" "$(bazel --version 2>&1 | head -1)"
   io::log_h1 "Starting local build: ${BUILD_NAME}"
   readonly TIMEFORMAT="==> 🕑 ${BUILD_NAME} completed in %R seconds"
   time "${PROGRAM_DIR}/builds/${BUILD_NAME}.sh"


coryan (approved):
LGTM as-is.

There are, of course, other pieces of information you could capture. From the top of my head, the linker, and the `libstdc++` and/or `libc++` versions would be interesting.

devjgm:
> LGTM as-is.
> 
> There are, of course, other pieces of information you could capture. From the top of my head, the linker, and the `libstdc++` and/or `libc++` versions would be interesting.

Great suggestions. The point of this change is to give us a place to add things like this when we think of them. Sadly, I'm not sure of the right way to detect these versions, so it may be better to add these later.

ci/cloudbuild/build.sh
```diff
@@ -116,6 +116,21 @@ if [[ "${LOCAL_BUILD}" = "true" ]]; then
     print_usage
     exit 1
   fi
+  function mem_total() {
+    awk '$1 == "MemTotal:" {printf "%0.2f GiB", $2/1024/1024}' /proc/meminfo
+  }
+  io::log_h1 "Machine Info"
+  TZ=/etc/localtime printf "%10s %s (%s)\n" "local:" "$(date)" "$(date +%z)"
+  printf "%10s %s\n" "utc:" "$(date -u)"
```
devbww, line 124: It might be nice if these were guaranteed to show the same time, with something like ...
```
now=$(date +%s)
TZ=/etc/localtime printf "%10s %s (%s)\n" "local:" "$(date -d @${now})" "$(date +%z -d @${now})"
printf "%10s %s\n" "utc:" "$(date -u -d @${now})"
```
And then you could also avoid the need for the extra `+%z` call by using RFC3339 output like ...
```
TZ=/etc/localtime printf "%10s %s\n" "local:" "$(date -d @${now} --rfc-3339=seconds)"
printf "%10s %s\n" "utc:" "$(date -u -d @${now} --rfc-3339=seconds)"
```
But, who cares about "localtime" (and whether `TZ=/etc/localtime` is the right way to force it)?  Perhaps we should be running all these scripts under `export TZ=:UTC`?

devjgm: > It might be nice if these were guaranteed to show the same time, with something like ...
> 
> ```
> now=$(date +%s)
> TZ=/etc/localtime printf "%10s %s (%s)\n" "local:" "$(date -d @${now})" "$(date +%z -d @${now})"
> printf "%10s %s\n" "utc:" "$(date -u -d @${now})"
> ```
> 
> And then you could also avoid the need for the extra `+%z` call by using RFC3339 output like ...
> 
> ```
> TZ=/etc/localtime printf "%10s %s\n" "local:" "$(date -d @${now} --rfc-3339=seconds)"
> printf "%10s %s\n" "utc:" "$(date -u -d @${now} --rfc-3339=seconds)"
> ```
> 
> But, who cares about "localtime" (and whether `TZ=/etc/localtime` is the right way to force it)? Perhaps we should be running all these scripts under `export TZ=UTC`?

*Our* logs are always in UTC, but others' log output is sometimes in localtime. We see this often with Kokoro where timestamps are mixed between UTC, localtime, and America/Los_Angeles. Even the GCB UI will sometimes show timestamps in localtime. The idea was to highlight to the reader of the logs whether or not localtime was different, and what the difference was.

But, I do like your suggestion of forcing UTC for everything, so I changed out GCBs and docker builds to set `TZ=:UTC` so that all logs will be in UTC (hopefully). The Cloud Build web UI will still show localtime sometimes.

I removed the print for localtime.

I added a line that logs the time that Google thinks it is, with the idea that this could help us diagnose time sync issues that may affect our integration tests (which we've seen happen on Kokoro).

See https://github.com/googleapis/google-cloud-cpp/pull/6124

PR #6124, opened 2021-03-28
ci(gcb): set TZ=:UTC; log google_time

devjgm:
Per @devbww's comments on
https://github.com/googleapis/google-cloud-cpp/pull/6123, we set
`TZ=:UTC` for docker and GCB builds so our logs -- and others' logs --
show up in UTC to make them easily comparable. This PR also logs the
time that google.com thinks it is, which would help us diagnose issues
caused by time synchronization issues (which we've seen on Kokoro).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6124)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -124,7 +124,7 @@ if [[ "${LOCAL_BUILD}" = "true" ]]; then
     curl -sI google.com | grep "^Date:" | cut -f2- -d:
   }
   io::log_h1 "Machine Info"
-  printf "%10s %s\n" "utc:" "$(date -u --rfc-3339=seconds)"
+  printf "%10s %s\n" "host:" "$(date -u --rfc-3339=seconds)"
   printf "%10s %s\n" "google:" "$(date -ud "$(google_time)" --rfc-3339=seconds)"
   printf "%10s %s\n" "kernel:" "$(uname -v)"
   printf "%10s %s\n" "os:" "$(grep PRETTY_NAME /etc/os-release)"
@@ -172,7 +172,7 @@ if [[ "${DOCKER_BUILD}" = "true" ]]; then
     "--rm"
     "--user=$(id -u):$(id -g)"
     "--env=USER=$(id -un)"
-    "--env=TZ=:UTC"
+    "--env=TZ=UTC0"
     # Mounts an empty volume over "build-out" to isolate builds from each
     # other. Doesn't affect GCB builds, but it helps our local docker builds.
     "--volume=/workspace/build-out"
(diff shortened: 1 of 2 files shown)


devjgm:
Example output:
```
====================
|   Machine Info   |
====================
      utc: 2021-03-28 14:10:52+00:00
   google: 2021-03-28 14:10:52+00:00
   kernel: #41~18.04.1-Ubuntu SMP Fri Feb 26 22:23:13 UTC 2021
       os: PRETTY_NAME="Fedora 33 (Container Image)"
    nproc: 32
      mem: 28.21 GiB
      gcc: gcc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)
    clang: clang version 11.0.0 (Fedora 11.0.0-2.fc33)
       cc: cc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)
    cmake: cmake version 3.19.7
    bazel: bazel 3.5.0
```

ci/cloudbuild/build.sh
```diff
@@ -168,6 +172,7 @@ if [[ "${DOCKER_BUILD}" = "true" ]]; then
     "--rm"
     "--user=$(id -u):$(id -g)"
     "--env=USER=$(id -un)"
+    "--env=TZ=:UTC"
```
devbww, line 175: I'm having second thoughts about the precise incantation.  `TZ=:UTC` will probably only work on tzdb-based implementations that compile the "backward" rules, and I'm not sure who does that.

We're probably better off with `TZ=UTC0`, which will work on any system that supports the POSIX TZ spec, and that includes any Unix-like libc, plus Windows I believe.  Abseil/CCTZ will actually fail to load "UTC0" (unlike bare "UTC", which we special case), but it will then fall back to UTC, so that's OK.  [But I think I'll tweak CCTZ so that loading "UTC0" returns success too.]

ci/cloudbuild/build.sh
```diff
@@ -119,9 +119,13 @@ if [[ "${LOCAL_BUILD}" = "true" ]]; then
...
     awk '$1 == "MemTotal:" {printf "%0.2f GiB", $2/1024/1024}' /proc/meminfo
   }
+  function google_time() {
+    # Extracts the time that Google thinks it is.
+    curl -sI google.com | grep "^Date:" | cut -f2- -d:
+  }
   io::log_h1 "Machine Info"
-  TZ=/etc/localtime printf "%10s %s (%s)\n" "local:" "$(date)" "$(date +%z)"
-  printf "%10s %s\n" "utc:" "$(date -u)"
+  printf "%10s %s\n" "utc:" "$(date -u --rfc-3339=seconds)"
```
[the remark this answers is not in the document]
devjgm: Done. Though named just "host" and "google" to keep alignment nice.

ci/cloudbuild/build.sh
devjgm: Done.

devbww (approved):

PR #6132, opened 2021-03-29
ci(gcb): add quickstart-cmake and quickstart-bazel builds

devjgm:
The `quickstart-cmake` builds use ccache to speed up the building of the `vcpkg` binary itself. And it uses `VCPKG_BINARY_SOURCES` to speed up vcpkg's installation of `google-cloud-cpp`. 

I'll enable these triggers after this PR merges.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6132)


coryan (commented):
You need to fix some shell check warnings, but otherwise LGTM.

coryan (approved):

PR #6135, opened 2021-03-30
ci(gcb): add builds for old clang and gcc 

devjgm:
The builds are named with `38` and `54` because GCB triggers cannot contain dots. Alternative names:
* `clang-3-8`
* `clang-old`

Happy to rename if other names are better.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6135)


coryan (approved):
:lgtm: 

Something to think about as you redesign the build matrix, there is some overlap between these builds and the `ci/kokoro/install` builds for Ubuntu:16.04.

Something else to think about: Ubuntu:16.04 is reaching "end of standard support" (not the same as EOL) at the end of April.  We should try to understand if/when GCE will stop supporting Ubuntu:16.04 and act accordingly.

> Reviewed 1 of 7 files at r1, 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6135#-:-MX26R9H87DqapuYO89S:brysmla)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devjgm)

devjgm:
> :lgtm:
> 
> Something to think about as you redesign the build matrix, there is some overlap between these builds and the `ci/kokoro/install` builds for Ubuntu:16.04.
> 
> Something else to think about: Ubuntu:16.04 is reaching "end of standard support" (not the same as EOL) at the end of April. We should try to understand if/when GCE will stop supporting Ubuntu:16.04 and act accordingly.
> 
> > Reviewed 1 of 7 files at r1, 6 of 6 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6135#-:-MX26R9H87DqapuYO89S:brysmla)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devjgm)

Good suggestions. I'll try to make a table today showing the build matrix that I *think* we're shooting for so we can see what may need to change.

PR #6138, opened 2021-03-30
ci: prevent flakes in Bazel builds

coryan:
Download the dependencies of a Bazel build before starting the build,
with retries. This adds little or no cost to a normal build, as the
dependencies would have been downloaded anyway, but prevents builds from
flaking out when there is some kind of network error that prevents the
download from working.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6138)

### DIFF
--- a/ci/cloudbuild/builds/lib/bazel.sh
+++ b/ci/cloudbuild/builds/lib/bazel.sh
@@ -14,13 +14,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This bash library has various helper functions for our bazel-based builds.
+# This bash library has various helper functions for our bazel-based builds
+# and automatically pre-fetches all dependencies for the project.
 
 # Make our include guard clean against set -o nounset.
 test -n "${CI_CLOUDBUILD_BUILDS_LIB_BAZEL_SH__:-}" || declare -i CI_CLOUDBUILD_BUILDS_LIB_BAZEL_SH__=0
 if ((CI_CLOUDBUILD_BUILDS_LIB_BAZEL_SH__++ != 0)); then
   return 0
 fi # include guard
+
+source module ci/lib/io.sh
+
+io::log_h2 "Prefetching bazel deps"
+# Bazel downloads all the dependencies of a project, as well as a number of
+# development tools during startup. In automated builds these downloads fail
+# from time to time due to transient network problems. Running `bazel fetch` at
+# the beginning of the build prevents such transient failures from flaking the
+# build.
+"./ci/retry-command.sh" 3 120 bazel fetch ... \
+  @local_config_platform//... \
+  @local_config_cc_toolchains//... \
+  @local_config_sh//... \
+  @go_sdk//... \
+  @remotejdk11_linux//:jdk
 
 # Outputs a list of args that should be given to all bazel invocations. To read
 # this into an array use `mapfile -t my_array < <(bazel::common_args)`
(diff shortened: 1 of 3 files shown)


ci/cloudbuild/builds/lib/bazel.sh
```diff
@@ -40,3 +40,17 @@ function bazel::common_args() {
...
   printf "%s\n" "${args[@]}"
 }
+
+# Bazel downloads all the dependencies of a project, as well as a number of
+# development tools during startup. In automated builds these downloads fail
+# from time to time due to transient network problems. Running `bazel fetch` at
+# the beginning of the build prevents such transient failures from flaking the
+# build.
+function bazel::fetch_dependencies() {
+  "./ci/retry-command.sh" 3 120 bazel fetch ... \
```
devjgm, line 50: (comment) I was hoping the network flakiness was mostly a kokoro issue and that we'd be able to avoid this w/ GCB. Oh well. (maybe I blame too much on kokoro ;-) )

Q: Why fetch and not sync? I think fetch requires us to specify specific targets, none of which make sense to me. Would just doing `bazel sync` be sufficient? (I don't know; asking)

ci/cloudbuild/builds/lib/bazel.sh
```diff
@@ -40,3 +40,17 @@ function bazel::common_args() {
   fi
   printf "%s\n" "${args[@]}"
 }
+
+# Bazel downloads all the dependencies of a project, as well as a number of
+# development tools during startup. In automated builds these downloads fail
+# from time to time due to transient network problems. Running `bazel fetch` at
+# the beginning of the build prevents such transient failures from flaking the
+# build.
+function bazel::fetch_dependencies() {
```
devjgm, line 49: Do we want this function to be called by all of our bazel builds? Basically by everything that sources this file? I tend to think so. In which case, I'm thinking that we may want to just list this command above (but after the include guard) outside of a function so that it's executed automatically when this file is sourced. We do something similar in the [`lib/cmake.sh`](https://github.com/googleapis/google-cloud-cpp/blob/4755d7ad262d25f430533ae84dccb3764b80badd/ci/cloudbuild/builds/lib/cmake.sh#L30) library. WDYT?

ci/cloudbuild/builds/lib/bazel.sh
coryan: > (comment) I was hoping the network flakiness was mostly a kokoro issue and that we'd be able to avoid this w/ GCB. Oh well. (maybe I blame too much on kokoro ;-) )

I am just going to leave this here:

https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing


> 
> Q: Why fetch and not sync? I think fetch requires us to specify specific targets, none of which make sense to me. Would just doing `bazel sync` be sufficient? (I don't know; asking)

As I recall `sync` does not download the indirect dependencies nor the SDKs (e.g. `@go_sdk//...`).  Candidly, I don't fully understand why the SDKs are needed, but they are.

devjgm (approved):
> As I recall sync does not download the indirect dependencies nor the SDKs (e.g. @go_sdk//...). Candidly, I don't fully understand why the SDKs are needed, but they are.

On my machine I `rm -rf ~/.cache/bazel`, then `bazel sync`, and then I do see files for the `go_sdk`, `remotejdk11`, etc. It looks like it's recursively fetching all the deps. Possibly the behavior changed.

ci/cloudbuild/builds/lib/bazel.sh
```diff
@@ -40,3 +40,15 @@ function bazel::common_args() {
   fi
   printf "%s\n" "${args[@]}"
 }
+
```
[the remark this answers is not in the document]
coryan: Done.

Added #6140 to track the `sync` vs. `fetch` issue.

PR #6169, opened 2021-04-01
feat(storage): reduce overhead in pooled handle factory

coryan:
Replace `std::vector<>` with `std::deque<>` because when the pool grows
too large we want to remove the *oldest* handles (fron the front). Also,
use a while loop during cleanup because it makes the intent clearer. I
expect that while loop to run at most once, but why force humans to
think this through?

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6169)


google/cloud/storage/internal/curl_handle_factory.cc
```diff
@@ -89,6 +86,7 @@ PooledCurlHandleFactory::~PooledCurlHandleFactory() {
 }
 
 CurlPtr PooledCurlHandleFactory::CreateHandle() {
+  std::cerr << __func__ << "()" << std::endl;
```
devjgm, line 89: Left in debugging?

devjgm (approved):

PR #6051, opened 2021-03-16
fix(bigtable): multiple region tags on same line

billyjacobson:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6051)

PR #6172, opened 2021-04-01
chore: April 2021 release notes and ABI update

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6172)


CHANGELOG.md
```diff
@@ -75,6 +88,49 @@
   `BackgroundThreadsFactory()` member function, requiring anyone who derives
   from those classes to implement that function. This would typically only
   affect users who create mock clients for testing purposes.
+* fix(bigtable): multiple region tags on same line (#6051)
```
coryan, line 91: These changes are not breaking, right? So we can split them with a `**OTHER CHANGES**:` header (or whatever we have used in the past)?

Also, #6051 was probably mislabeled, it is a cleanup and uninteresting to customers.

coryan (commented):
I think we should separate the breaking and non-breaking changes for bigtable?

coryan (approved):

Issue #6161, opened 2021-03-31
Convert google/cloud/storage to use g::c::Options


coryan:
This is needed to implement a number of things beyond GUAC.

coryan:
I think this is done. The remaining step is to mark the old functions as deprecated and #6642 is all about that.

PR #6209, opened 2021-04-08
refactor(storage): use g::c::Options for GrpcClient

coryan:
Change `storage::internal::GrpcClient` and
`storage::internal::HybridClient` to use `google::cloud::Options`. I had
to define exactly where the defaults are applied to an `Options` object,
for now this is in the `*Client::Create()` factory functions, it may get
refined later.

Some additional refactoring in `storage::ClientOptions` and
`storage::internal::CurlClient` to share the initialization of default
options.

Part of the work for #6161

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6209)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -59,9 +59,9 @@ bool DirectPathEnabled() {
                         [](absl::string_view v) { return v == "storage"; });
 }
 
-Options FillWithDefaultsGrpc(Options options) {
-  options = FillWithDefaults(std::make_shared<oauth2::AnonymousCredentials>(),
-                             std::move(options));
+Options DefaultOptionsGrpc(Options options) {
+  options = DefaultOptions(std::make_shared<oauth2::AnonymousCredentials>(),
+                           std::move(options));
   auto env = google::cloud::internal::GetEnv("CLOUD_STORAGE_GRPC_ENDPOINT");
   if (env.has_value()) {
     options.set<GrpcCredentialOption>(grpc::InsecureChannelCredentials());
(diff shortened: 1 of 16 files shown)


google/cloud/storage/internal/grpc_client.cc
```diff
@@ -48,60 +59,54 @@ bool DirectPathEnabled() {
                         [](absl::string_view v) { return v == "storage"; });
 }
 
-std::string GrpcEndpoint() {
+Options FillWithDefaultsGrpc(Options options) {
+  options = FillWithDefaults(std::make_shared<oauth2::AnonymousCredentials>(),
```
devjgm, line 63: I assume you saw the naming pattern used by Spanner (https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/spanner/internal/defaults.h) and the generator, but following that pattern, this would be named `DefaultOptions()`, with a ` = {}` default argument. I'm  happy to go with whatever name you choose, but just pointing out the difference in case you didn't see that.

google/cloud/storage/internal/grpc_client.cc
```diff
@@ -48,60 +59,54 @@ bool DirectPathEnabled() {
...
-      stub_(
-          google::storage::v1::Storage::NewStub(CreateGrpcChannel(options_))) {}
+std::shared_ptr<GrpcClient> GrpcClient::Create(Options options) {
+  return Create(std::move(options), /*channel_id=*/0);
+}
+
+std::shared_ptr<GrpcClient> GrpcClient::Create(Options options,
+                                               int channel_id) {
+  // Cannot use std::make_shared<> as the constructor is private.
+  return std::shared_ptr<GrpcClient>(
```
devjgm, line 101: Are you planning to call `internal::CheckExpectedOptions()` from somewhere? That should probably be in the first user-facing function that accepts user-provided `options`. I'd imagine that's also the function where you'd call `FillWithDefaults`, which seems like maybe here?

Fine to do in a separate PR.

devjgm (approved):

google/cloud/storage/internal/grpc_client.cc
coryan: Done.  I was not paying attention to be candid.

google/cloud/storage/internal/grpc_client.cc
coryan: > Are you planning to call `internal::CheckExpectedOptions()` from somewhere?

Yes.

> Fine to do in a separate PR.

That was my plan, I do not know *exactly* what that first entry point will be, there are a bit too many ways to create `storage::Client` things right now.

PR #6223, opened 2021-04-10
fix(ci): gitignore pattern that was too broad

devjgm:
We were ignoring all `bazel-*` files, which included some trigger files
that I thought were checked in. The triggers are active and running, but
apparently they weren't in git.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6223)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ build-out/
 cmake-out/
 
 # Common bazel output directories
-bazel-*
+/bazel-*
 !bazel-*.cfg
 
 # Used by some IDEs and LSP plugins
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -6,7 +6,7 @@ build-out/
 cmake-out/
 
 # Common bazel output directories
-bazel-*
+/bazel-*
```
coryan, line 9: If you build in one of the `quickstart/` directories this could be a problem.  There is no way to win.

coryan (approved):

.gitignore
devjgm: I added a rule to ignore `/google/cloud/*/quickstart/bazel-*`, which seems to cover that case too.

coryan (approved):

PR #6229, opened 2021-04-12
ci(gcb): add useful tags to all builds

devjgm:
This will let us filter on builds based on whether they're CI, PR, or
manually triggered builds. I'm also tagged based on distro and build
name, which may be useful too.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6229)


devjgm:
Example of tags showing up in the GCB UI:

image

coryan (approved):

Issue #6163, opened 2021-04-01
[GCB] Add install builds


devjgm:
We need to test our code correctly compiles, installs, and is usable on a variety of Linux distros. We call these "install" builds. In the kokoro world these live here https://github.com/googleapis/google-cloud-cpp/tree/master/ci/kokoro/install

We should run the quickstarts to verify that the installed artifacts work, and can be linked with.

*We need to make sure to test `make` also*

PR #6239, opened 2021-04-13
ci(gcb): add demo-centos-7 install build

devjgm:
Part of #6163

This is the first example of how I plan to move what we currently call
"install" builds (defined in `ci/kokoro/install`) over to GCB where I
plan to call them "demo" builds. The idea is that we'll have a single
script called `demo-install.sh` (in this PR) that installs
`google-cloud-cpp`, and then tests the install by running the
quickstarts. We'll then have a number of `demo-<os>.Dockerfile` files
(`demo-centos-7.Dockerfile` is included in this PR), and a number of
trigger files, which associate a `demo-<os>` distro with the one
`demo-install.sh` script.

In subsequent PRs, I'll add more `demo-<os>.Dockerfile`s, and more
triggers for the other platforms.

Separately, I'll also update the packaging scripts to use these new
files for generating the `packaging.md` file.

Note about naming: I thought that "install" was a slightly ambiguous
term for these builds, and thought "demo" was a little more descriptive
since we use these builds to demonstrate how to install the deps and
install our library on various platforms. Happy to change the naming
back to "install" if others dislike the new naming.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6239)

### DIFF
--- a/ci/cloudbuild/builds/demo-install.sh
+++ b/ci/cloudbuild/builds/demo-install.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This build script serves two main purposes:
+#
+# 1. It demonstrates to users how to install `google-cloud-cpp`. We will
+#    extract part of this script into user-facing markdown documentation.
+# 2. It verifies that the installed artifacts work by compiling and running the
+#    quickstart programs against the installed artifacts.
+
+set -eu
+
+source "$(dirname "$0")/../../lib/init.sh"
+source module ci/cloudbuild/builds/lib/cmake.sh
+source module ci/etc/quickstart-config.sh
+source module ci/lib/io.sh
+
+## [START packaging.md]
+
+# Pick a location to install the artifacts, e.g., `/usr/local` or `/opt`
+export PREFIX="${HOME}/google-cloud-cpp-installed"
+cmake -DBUILD_TESTING=OFF -DCMAKE_INSTALL_PREFIX="${PREFIX}" -H. -Bcmake-out
+cmake --build cmake-out -- -j "$(nproc)"
+cmake --build cmake-out --target install
+
+## [END packaging.md]
+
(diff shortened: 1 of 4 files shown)


ci/cloudbuild/builds/demo-install.sh
```diff
@@ -0,0 +1,65 @@
...
+#    quickstart programs against the installed artifacts.
+
+set -eu
+
+source "$(dirname "$0")/../../lib/init.sh"
+source module ci/cloudbuild/builds/lib/cmake.sh
+source module ci/etc/quickstart-config.sh
+source module ci/lib/io.sh
+
+## [START packaging.md]
```
coryan, line 31: nit: the use of `START` and `END` markers clashes with the cloud site documentation tags.  You may be in a good position to fix this as part of this PR series.

ci/cloudbuild/builds/demo-install.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Pick a location to install the artifacts, e.g., `/usr/local` or `/opt`
+export PREFIX="${HOME}/google-cloud-cpp-installed"
```
coryan, line 34: nit: export?  Do we need to define this variable at all?

ci/cloudbuild/triggers/demo-centos-7-ci.yaml
```diff
@@ -0,0 +1,13 @@
+filename: ci/cloudbuild/cloudbuild.yaml
```
coryan, line 1: Q: I assume you cannot enable these triggers until the code is merged, right?

coryan (approved):

ci/cloudbuild/builds/demo-install.sh
devjgm: Filed https://github.com/googleapis/google-cloud-cpp/issues/6241 to track this.

ci/cloudbuild/builds/demo-install.sh
devjgm: Export not needed and removed. This build script runs as a non-root user, so I can't install in `/usr/local`. I needed to change it to some dir that's writable.

I tried using `DESTDIR`, which almost works, but I could never get it to actually work correctly.

Let me know if you have another idea about how I could avoid the `-DCMAKE_INSTALL_PREFIX`

ci/cloudbuild/triggers/demo-centos-7-ci.yaml
devjgm: Correct. Well, I _could_ enable them and they would work on _my_ PR because my PR has the right scripts. However, if someone else submitted a PR, the trigger would fail for them, so I tend not to enable triggers until after I merge the required scripts.

ci/cloudbuild/builds/demo-install.sh
coryan: Showing the `-DCMAKE_INSTALL_PREFIX` LGTM, I just did not want to imply that an export was required.

coryan (approved):

PR #6243, opened 2021-04-13
ci!: generator/ is no longer built by default in CMake

devjgm:
Our `generator/` code is an implementation that our users will rarely,
if ever, need to build or use directly. So it seems it would be better
to not compile it by default in our CMake builds. This PR changes the
generator to not be build by default in CMake builds.

It also explicitly enables the generator in our clang-tidy builds
(kokoro and GCB) so that the generator is checked. All the sanitizer
builds use bazel so they will be unaffected.

This is technically a breaking change, but it's very unlikely to affect
anyone in practice.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6243)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,15 @@
 
 ## v1.27.0 - TBD
 
+### Misc.
+
+**BREAKING CHANGES**:
+* In #6243 we stopped compiling the code in `generator/` by default in CMake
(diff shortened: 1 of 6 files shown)


coryan (approved):
Please update CHANGELOG.md now, so the person making the next release has an easier life.

PR #6245, opened 2021-04-13
ci(gcb): add all remaining demo builds

devjgm:
Fixes: #6163

This PR adds "demo" builds (new parlance) for all the old "install" builds (old parlance).

This looks like a big PR, but it's pretty mechanical. Mostly copying the old "install" Dockerfiles, and generating new trigger yaml files.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6245)

### DIFF
--- a/ci/cloudbuild/builds/demo-install.sh
+++ b/ci/cloudbuild/builds/demo-install.sh
@@ -39,7 +39,7 @@ cmake --build cmake-out --target install
 ## [END packaging.md]
 
 # Tells pkg-config where the artifacts are, so the Makefile builds work.
-export PKG_CONFIG_PATH="${PREFIX}/lib64/pkgconfig:${PKG_CONFIG_PATH:-}"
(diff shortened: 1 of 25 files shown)


coryan (approved):
I should have thought about this earlier.  "demo" is fine, but maybe "packaging" is a better name? Your call.

devjgm:
> I should have thought about this earlier. "demo" is fine, but maybe "packaging" is a better name? Your call.

"Packaging" was my first thought as well. I even implemented half of these using that name. The reasons I changed the name and began to favor "demo"

1. While we use these builds to generate our "packaging" instructions, these are useful for more than just packaging. Anyone who would like to install our library in a usable way on a distro could/should follow these. In that regard, these are kind of like demonstrations for how to install the library on various distros.
2. "demo" is shorter, and that matters a little because we have limited real estate in the GH "checks" UI. "packaging-ubuntu-bionic-pr" is getting pretty long.

Anyway, my mind is still open on naming, but that's why I landed here. A potential rename should be in a separate PR anyway.

PR #6250, opened 2021-04-14
ci(gcb): enable noex build

devjgm:
The triggers will be active after I merge this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6250)

### DIFF
--- a/ci/cloudbuild/builds/noex.sh
+++ b/ci/cloudbuild/builds/noex.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


ci/cloudbuild/builds/noex.sh
```diff
@@ -0,0 +1,35 @@
...
+source module ci/cloudbuild/builds/lib/bazel.sh
+
+export CC=gcc
+export CXX=g++
+
+mapfile -t args < <(bazel::common_args)
+args+=("--copt=-fno-exceptions")
+
+# Skip examples that require exceptions for error handling.
+bazel test "${args[@]}" --test_tag_filters=-integration-test ... -- \
```
coryan, line 29: Are you planning to have a CMake build without exceptions?  There are a number of `if (GOOGLE_CLOUD_CPP_ENABLE_CXX_EXCEPTIONS)` statements that would go untested if we move the `noex` build to Bazel.

devjgm: I wasn't, but now that you mention it, we probably should. I'll probably just change this build to use cmake.... coming up.

devjgm: Done. PTAL

coryan (approved):

PR #6255, opened 2021-04-14
ci(gcb): triggers will work on all branches we care about

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6255)

### DIFF
--- a/ci/cloudbuild/trigger.sh
+++ b/ci/cloudbuild/trigger.sh
@@ -50,7 +50,7 @@ github:
   name: google-cloud-cpp
   owner: googleapis
   push:
-    branch: ^(master|main|preview-.*|v\\d+\\..*)$
+    branch: ^(master|main|v\\d+\\..*)$
 name: ${name}-ci
 substitutions:
   _BUILD_NAME: ${name}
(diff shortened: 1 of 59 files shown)


ci/cloudbuild/trigger.sh
```diff
@@ -50,7 +50,7 @@ github:
   name: google-cloud-cpp
   owner: googleapis
   push:
-    branch: ^master$
+    branch: ^(master|main|preview-.*|v\\d+\\..*)$
```
devbww, line 53: I presume this will be OK if there are no "preview-.*" branches.

That said, I think they were hard to use.  For googleapis at least, the "patch after fetch" method seemed much easier to use for development.  So, I'd be happy not to see them immortalized here (at least until we have a better use for them).

devjgm: I believe this regex should be fine even without preview- branches. I think they just apply this regex to incomping PRs to see if the trigger should run.... Indeed, it works with `main` specified, which doesn't yet exist.

If our plan is to do away with `preview-` branches, then I'm happy to remove this alternate. However, if we plan to keep them, even for a while longer, it seems low-cost to keep this regex and run some tests on PRs to the preview- branch. We could always remove it later if/when we're done with these branches. I don't think there's much immortal here -- pretty easy to add/remove at any point.

WDYT?

coryan (approved):

ci/cloudbuild/trigger.sh
devbww: > If our plan is to do away with preview- branches, then I'm happy to remove this alternate.

I'm removing them now.

devjgm: well, OK then :-) I'll do the same here.

devjgm: Gone. PTAL.

devbww (approved):

PR #6257, opened 2021-04-14
ci(gcb): add shared library build

devjgm:
Also fixed an issue where the pubsub protos were missing a needed dep.

This PR refactors the logic to run the CMake and Makefile quickstart
builds into a library, because it turns out that this is a useful way to
verify that our installed artifacts work, and we end up calling this
from a few places.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6257)


coryan (approved):
Nice refactoring.

PR #6259, opened 2021-04-15
ci(gcb): add msan build

devjgm:
This build is a little strange because it requires a custom-built
libc++, which we build in the Dockerfile.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6259)

### DIFF
--- a/ci/cloudbuild/builds/msan.sh
+++ b/ci/cloudbuild/builds/msan.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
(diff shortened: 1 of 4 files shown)


coryan (approved):
The description had a typo (s/bec**ua**se/bec**au**se/), I fixed the PR, you should double check the commit message.

I like the name for the Dockerfile.

devjgm:
> The description had a typo (s/bec**ua**se/bec**au**se/), I fixed the PR, you should double check the commit message.

Thanks. Fixed.

PR #6261, opened 2021-04-15
chore(spanner): quash warning about a leaked mocked object

### LINKED ISSUE #6145: [Flake] Spanner unit test fails with "leaked" mock object
The log says:

```
../../google/cloud/spanner/database_admin_connection_test.cc:917: ERROR: this mock object (used in test DatabaseAdminClientTest.CreateBackupWithEncryption) should be deleted but never is. Its address is @0xd49970.
ERROR: 1 leaked mock object found at program exit. Expectations on a mock object is verified when the object is destructed. Leaking a mock means that its expectations aren't verified, which is usually a test bug. If you really intend to leak a mock, you can suppress this error using testing::Mock::AllowLeak(mock_object), or you may use a fake or stub instead of a mock.
```

https://source.cloud.google.com/results/invocations/b4c30a87-27e9-4640-b290-f47f10282ce8

I think this is related to #4444, the mocks remain referenced by a (detached) thread.  Let the record reflect that I think detached threads are a "Bad Thing":tm:, but with that out of the way, I think the changes proposed in #4038 these should go away.

devbww:
Suppress the warning about leaking the mock stub until #4038
is fixed.

Fixes #6145.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6261)

### DIFF
--- a/google/cloud/spanner/database_admin_connection_test.cc
+++ b/google/cloud/spanner/database_admin_connection_test.cc
@@ -912,6 +912,10 @@ TEST(DatabaseAdminConnectionTest, CreateBackupSuccess) {
 /// @test Verify that using an encryption key works.
 TEST(DatabaseAdminClientTest, CreateBackupWithEncryption) {
   auto mock = std::make_shared<MockDatabaseAdminStub>();
+  // Suppress a false leak.
+  // TODO(#4038): After we fix the issue #4038, we won't need to use
+  // `AllowLeak()` any more.
+  Mock::AllowLeak(mock.get());
   Database dbase("test-project", "test-instance", "test-database");
 
   EXPECT_CALL(*mock, CreateBackup(_, _))


coryan (approved):

google/cloud/spanner/database_admin_connection_test.cc
```diff
@@ -912,6 +912,10 @@ TEST(DatabaseAdminConnectionTest, CreateBackupSuccess) {
 /// @test Verify that using an encryption key works.
 TEST(DatabaseAdminClientTest, CreateBackupWithEncryption) {
   auto mock = std::make_shared<MockDatabaseAdminStub>();
+  // Suppress a false leak.
+  // TODO(#4038): After we fix the issue #4038, we won't need to use
+  // `AllowLeak()` any more.
+  Mock::AllowLeak(mock.get());
```
coryan, line 918: It occurs to me (too late I know) that we may have the same problem with other mocks in this test ... I am not sure why this mock would be any more leaky than the others... 🤷

devbww: My rationale (without actually checking) was that only certain test cases hit the detached-thread-creating interfaces, and therefore to play a wait-and-see game for now.

PR #6262, opened 2021-04-15
feature(generator): add standalone executable with config file support

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6262)


generator/BUILD
```diff
@@ -60,8 +60,32 @@ load(":google_cloud_cpp_generator_unit_tests.bzl", "google_cloud_cpp_generator_u
...
+
+cc_proto_library(
+    name = "generator_config_cc_proto",
+    deps = [":generator_config_proto"],
+)
+
+cc_binary(
+    name = "google-cloud-cpp-codegen",
+    srcs = ["standalone_main.cc"],
+    includes = ["."],
```
devjgm, line 82: Are these `includes = ["."]` lines needed? I suspect they aren't or shouldn't be needed if we should be including headers with paths relative to the project root.

generator/standalone_main.cc
```diff
@@ -0,0 +1,132 @@
...
+
+namespace {
+google::cloud::StatusOr<google::cloud::cpp::generator::GeneratorConfiguration>
+GetConfig(std::string const& filepath) {
+  std::ifstream input(filepath);
+  std::stringstream buffer;
+  buffer << input.rdbuf();
+
+  google::cloud::cpp::generator::GeneratorConfiguration generator_config;
+  auto parse_result = google::protobuf::TextFormat::ParseFromString(
```
devjgm, line 48: I think there's also a `ParseFromIstream()`, which may be simpler since you're starting with a `std::ifstream`

generator/standalone_main.cc
```diff
@@ -0,0 +1,132 @@
...
+#include "generator/generator.h"
+#include "generator/generator_config.pb.h"
+#include <google/protobuf/compiler/command_line_interface.h>
+#include <google/protobuf/text_format.h>
+#include <fstream>
+#include <future>
+#include <iostream>
+#include <sstream>
+
+ABSL_FLAG(std::string, config_file, "./generator/generator_config.textproto",
```
devjgm, line 29: These relative path defaults imply that this binary should be run from a specific directory so that the default paths work. I recommend avoiding this issue by omitting this default path value, and the user can just specify `--config_file` with the appropriate relative path from *their* cwd.

generator/standalone_main.cc
```diff
@@ -0,0 +1,132 @@
...
+                               "Unable to parse textproto file.");
+}
+}  // namespace
+
+int main(int argc, char** argv) {
+  absl::ParseCommandLine(argc, argv);
+
+  auto proto_path = absl::GetFlag(FLAGS_protobuf_proto_path);
+  auto googleapis_path = absl::GetFlag(FLAGS_googleapis_proto_path);
+  auto googleapis_commit_hash = absl::GetFlag(FLAGS_googleapis_commit_hash);
```
devjgm, line 62: Are these flags required? Consider: adding some documentation about how to use this binary. Maybe as comments in this file, or as a `Usage()` function or something.

generator/standalone_main.cc
```diff
@@ -0,0 +1,132 @@
...
+    args.emplace_back("--cpp_codegen_opt=googleapis_commit_hash=" +
+                      googleapis_commit_hash);
+    args.emplace_back("--cpp_codegen_opt=copyright_year=" +
+                      service.initial_copyright_year());
+    for (auto const& omit_rpc : service.omitted_rpcs()) {
+      args.emplace_back("--cpp_codegen_opt=omit_rpc=" + omit_rpc);
+    }
+    args.emplace_back(service.service_proto_path());
+    GCP_LOG(INFO) << "Generating service code using: "
+                  << absl::StrJoin(args.begin(), args.end(), ";") << "\n";
```
devjgm, line 97: FYI, StrJoin accepts a range, so this can simply be `absl::StrJoin(args, ';')`

devjgm (approved):

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6262)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @devjgm)

___
*[generator/BUILD, line 82 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6262#-MYIV5D40Lfc9nr_b0Wk-r1-82:-MYQQWC25yHvZzsZUoym:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/935d0ed051a2c6e0e2d51787081b2465b463de59/generator/BUILD#L82)):*

Are these `includes = ["."]` lines needed? I suspect they aren't or shouldn't be needed if we should be including headers with paths relative to the project root.
</blockquote></details>

Done.
___
*[generator/standalone_main.cc, line 29 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6262#-MYIV5D40Lfc9nr_b0Wp-r1-29:-MYQQXc88KvcADTW0tur:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/935d0ed051a2c6e0e2d51787081b2465b463de59/generator/standalone_main.cc#L29)):*

These relative path defaults imply that this binary should be run from a specific directory so that the default paths work. I recommend avoiding this issue by omitting this default path value, and the user can just specify `--config_file` with the appropriate relative path from *their* cwd. 
</blockquote></details>

Done.
___
*[generator/standalone_main.cc, line 48 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6262#-MYIV5D40Lfc9nr_b0Wp-r1-48:-MYQQZP34tIdIcOW18bl:b4d45ce) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/935d0ed051a2c6e0e2d51787081b2465b463de59/generator/standalone_main.cc#L48)):*

I think there's also a `ParseFromIstream()`, which may be simpler since you're starting with a `std::ifstream`
</blockquote></details>

There's a google::protobuf::MessageLite::ParseFromIstream member function, but as it's not a drop in replacement, I'm thinking it expects the input in a different format than textproto.
___
*[generator/standalone_main.cc, line 62 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6262#-MYIV5D40Lfc9nr_b0Wp-r1-62:-MYQQYOM9DlB6IiK37gB:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/935d0ed051a2c6e0e2d51787081b2465b463de59/generator/standalone_main.cc#L62)):*

Are these flags required? Consider: adding some documentation about how to use this binary. Maybe as comments in this file, or as a `Usage()` function or something.
</blockquote></details>

Done.

PR #6263, opened 2021-04-15
cleanup: use wildcard matcher (`_`) sparingly

coryan:
Many of our mocks said `EXPECT_CALL(..., Foo(_, _))` when one can simply
say `EXPECT_CALL(..., Foo)` for the same effect. Furthermore, we were
inconsistent about this, with newer libraries generally ommitting the
extra `_` wildcards.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6263)


devjgm (approved):
Cleaner. Thanks.

PR #6264, opened 2021-04-15
ci: introduce io::log_cmdline

devbww:
Add `io::log_cmdline` for logging command execution. It prefixes
the command with `${PS4}`, just like `set -x` would in the shell.
This helps to delineate the command from normal output.

Also print the command in a bold typeface when possible, making
it even easier to distinguish.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6264)


coryan (approved):

devjgm (approved):

devjgm (approved):
also, I like the use of bold.

PR #6267, opened 2021-04-15
ci(gcb): add gcs-grpc build

devjgm:
From my understanding of the existing gcs-grpc build, I _think_ this is equivalent and testing the right things. However, is there a way for me to verify that the correct gcs + grpc logic/tests were run? Here's [output from a manual run](https://pantheon.corp.google.com/cloud-build/builds/3ff77a4e-3058-4134-8a22-8293cfdae02c?project=cloud-cpp-testing-resources) I did of this build in case this will show you whether or not the correct things ran.

Context: **current** `gcs-grpc` build:
https://github.com/googleapis/google-cloud-cpp/blob/21f3d910814fd9b5df5e511eecf3c141e2a3e1db/ci/kokoro/docker/build.sh#L278-L292

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6267)

### DIFF
--- a/ci/cloudbuild/builds/gcs-grpc.sh
+++ b/ci/cloudbuild/builds/gcs-grpc.sh
@@ -27,6 +27,8 @@ mapfile -t args < <(bazel::common_args)
 bazel test "${args[@]}" --test_tag_filters=-integration-test google/cloud/storage/...
 
 mapfile -t integration_args < <(integration::args)
+# "media" says to use the hybrid gRPC/REST client. For more details see
+# https://github.com/googleapis/google-cloud-cpp/issues/6268
 GOOGLE_CLOUD_CPP_STORAGE_GRPC_CONFIG=media # Uses the gRPC data plane
 io::log_h2 "Running Storage integration tests (with emulator)"
 google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh \
--- a/ci/cloudbuild/builds/gcs-grpc.sh
+++ b/ci/cloudbuild/builds/gcs-grpc.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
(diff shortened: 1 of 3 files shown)


ci/cloudbuild/builds/gcs-grpc.sh
```diff
@@ -0,0 +1,31 @@
...
+
+export CC=clang
+export CXX=clang++
+
+mapfile -t args < <(bazel::common_args)
+bazel test "${args[@]}" --test_tag_filters=-integration-test ...
+
+mapfile -t integration_args < <(integration::args)
+GOOGLE_CLOUD_CPP_STORAGE_GRPC_CONFIG=media # Uses the gRPC data plane
+integration::bazel_with_emulators test "${args[@]}" "${integration_args[@]}"
```
coryan, line 31: Soon (~next week) we would want to run the storage integration tests against production.  I assume you would be Okay if we changed `integration::bazel_with_emulators` to support switching one service to production?

Right now we are using the emulator because the GCS+gRPC service was down for about a month, but it is back up as of this week.

devjgm: > Soon (~next week) we would want to run the storage integration tests against production. I assume you would be Okay if we changed `integration::bazel_with_emulators` to support switching one service to production?

I just changed this test script to only run the GCS integration tests, which seems reasonable since the purpose of this build is testing GCS + gRPC.

Regarding changing to test against production: Would that be a change in `google/cloud/storage/ci/run_integration_tests_emulator_bazel.sh` or in this build script (`ci/cloudbuild/builds/gcs-grpc.sh`)? Either way, a change in either of those to hit production instead would be fine.

> 
> Right now we are using the emulator because the GCS+gRPC service was down for about a month, but it is back up as of this week.

Should I make the change to hit production right now? How would I do that?

coryan: > Should I make the change to hit production right now? How would I do that?

You need to revert the changes in #6063, it would make my life easier if you would. But it is more involved than adding a build.

devjgm: > > Should I make the change to hit production right now? How would I do that?
> 
> You need to revert the changes in #6063, it would make my life easier if you would. But it is more involved than adding a build.

I took a stab at doing this, but ended up with a few extra errors that weren't obvious. It doesn't look difficult to enable, but it's probably best done in a separate, focused PR. I think this whole script would amount to something like:

```shell
mapfile -t args < <(bazel::common_args)
mapfile -t integration_args < <(integration::args)
GOOGLE_CLOUD_CPP_STORAGE_GRPC_CONFIG=media # Uses the gRPC data plane
bazel test "${args[@]}" "${integration_args[@]}" google/cloud/storage/...
```
--
That said, is the PR as-is a reasonable replacement for the kokoro build we have today named `gcs-grpc`, or is it missing more stuff?

coryan (approved):
I think it is close enough, we can iterate from here.

PR #6269, opened 2021-04-15
cleanup: use wildcard matcher (_) even more sparingly

devbww:
A small followup to #6263.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6269)


coryan (approved):
Thanks! Sorry I missed these.

PR #6270, opened 2021-04-15
ci(gcb): enable coverage build

devjgm:
This seems to be working in my tests. For now, it's configured to upload
coverage data to devjgm's personal codecov account at
https://app.codecov.io/gh/devjgm/google-cloud-cpp/. Once we see that
this build works and uploads coverage data correctly, we can change it
to use the _real_ token, which can be done in Secret Manager.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6270)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -158,6 +158,13 @@ if [[ -z "${BUILD_NAME}" ]]; then
   exit 1
 fi
 
+# Info about the git repo that is used by some builds, e.g., coverage. These
+# will be automatically set by GCB for triggered builds, but we need to compute
+# them ourselves for manually started builds and docker builds. See
+# https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values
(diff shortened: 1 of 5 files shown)


coryan (approved):
LGTM.

This is a case where you may want to disable the Kokoro build pretty much at the same time?  I do not think we want two builds uploading data to codecov.io?

devjgm:
> LGTM.
> 
> This is a case where you may want to disable the Kokoro build pretty much at the same time? I do not think we want two builds uploading data to codecov.io?

Oops. I followed up about this offline with...

Re: the addition of the GCB coverage build (https://github.com/googleapis/google-cloud-cpp/pull/6270) The GCB coverage build is configured to write to my personal codecov account (https://app.codecov.io/gh/devjgm/google-cloud-cpp/) so that it won't conflict w/ the kokoro coverage build. Once it's merged we should be able to look at the coverage (in my account) to make sure the numbers look right and that it works. Then we can disable the kokoro coverage build and change the GCB build to use the main project account simply by changing the token that we have stored in Secret Manager.

PR #6281, opened 2021-04-16
ci(gcb): misc cleanups

devjgm:
These are a couple small cleanups and documentation improvements that I
had lying around.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6281)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -242,7 +242,9 @@ if [[ "${DOCKER_FLAG}" = "true" ]]; then
   )
   cmd=(ci/cloudbuild/build.sh --local "${BUILD_NAME}")
   if [[ "${SHELL_FLAG}" = "true" ]]; then
-    printf "To run the build manually:\n  %s\n\n" "${cmd[*]}"
+    printf "To run the build manually:\n  "
+    printf " %q" "${cmd[@]}"
+    printf "\n\n"
     cmd=("bash")
   fi
   docker run "${run_flags[@]}" "${image}" "${cmd[@]}"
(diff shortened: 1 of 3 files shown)


ci/cloudbuild/build.sh
```diff
@@ -242,8 +242,7 @@ if [[ "${DOCKER_FLAG}" = "true" ]]; then
   )
   cmd=(ci/cloudbuild/build.sh --local "${BUILD_NAME}")
   if [[ "${SHELL_FLAG}" = "true" ]]; then
-    io::log "Starting shell, to manually run the requested build use:"
-    echo "==> ${cmd[*]}"
+    printf "To run the build manually:\n  %s\n\n" "${cmd[*]}"
```
devbww, line 245: `%q` perhaps?

devbww (approved):

ci/cloudbuild/build.sh
devjgm: That would output something like `ci/cloudbuild/build.sh\ --local\ asan`, which with those escaped spaces, is not a copy-n-pasteable command. So I _think_ we want `%s`.

Context: The point of this output is that when you're trying to debug some build, and you drop to a shell, it may be useful to know the exact command that the build would have run, so that you can run it yourself and them maybe poke around at the output.

devbww: Then I think you want:
```
printf "To run the build manually:\n"
printf " %q" "${cmd[@]}"
printf "\n\n";
```

devjgm: Thanks. Done.

ci/cloudbuild/build.sh
```diff
@@ -242,7 +242,9 @@ if [[ "${DOCKER_FLAG}" = "true" ]]; then
   )
   cmd=(ci/cloudbuild/build.sh --local "${BUILD_NAME}")
   if [[ "${SHELL_FLAG}" = "true" ]]; then
-    printf "To run the build manually:\n  %s\n\n" "${cmd[*]}"
+    printf "To run the build manually:\n  "
```
devbww, line 245: Nit: Previously you indented the command by 2, but this will result in 3.

devjgm: I know. 3 looked nicer.

PR #6284, opened 2021-04-19
ci(gcb): write logs to our own bucket

devjgm:
By default GCB writes logs to a bucket they GCB controls and that we
have no access to. This PR changes our builds to write to our own
bucket, which will give us the ability to control the lifetime and
permissions for the build logs.

NOTE: This PR doesn't change anything wrt to log visibility or anything. This PR simply makes our build logs live in our bucket that we control.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6284)

### DIFF
--- a/ci/cloudbuild/cloudbuild.yaml
+++ b/ci/cloudbuild/cloudbuild.yaml
@@ -39,6 +39,7 @@ substitutions:
   _CACHE_TYPE: '${_PR_NUMBER:-main}'
   _IMAGE: 'cloudbuild/${_DISTRO}'
   _TRIGGER_TYPE: 'manual'
+  _LOGS_BUCKET_SUFFIX: 'publiclogs'
 
 tags: [ '${_TRIGGER_TYPE}', '${_BUILD_NAME}', '${_DISTRO}' ]
 timeout: 3600s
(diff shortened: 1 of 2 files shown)


ci/cloudbuild/cloudbuild.yaml
```diff
@@ -48,6 +49,8 @@ availableSecrets:
   - versionName: projects/${PROJECT_ID}/secrets/CODECOV_TOKEN/versions/latest
     env: 'CODECOV_TOKEN'
 
+logsBucket: 'gs://${PROJECT_ID}_${_LOGS_BUCKET_SUFFIX}/logs/google-cloud-cpp/${_CACHE_TYPE}/${COMMIT_SHA}/${_DISTRO}-${_BUILD_NAME}'
```
coryan, line 52: nit: the use of `_CACHE_TYPE` sounds odd...

coryan (approved):

ci/cloudbuild/cloudbuild.yaml
devjgm: Yep. Re-using it here suggests that it's a bad name. It results to a PR number for PR builds, `main` for CI builds, and `manual-user@domain` for manually triggered builds. I'll look into renaming that to something clearer in a separate PR.

PR #6285, opened 2021-04-19
ci(gcb): rename _CACHE_TYPE -> _TRIGGER_SOURCE

devjgm:
`_CACHE_TYPE` somewhat made sense when it was only used for the caching
URLs, but now that it's also used for the logging bucket directory it
needs a cache-agnostic name. This name seems a bit better, but I'm open
to other names.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6285)


coryan (approved):
Thanks.

PR #6286, opened 2021-04-19
refactor!: remove gcpcxx.pb suffix from generated files

### LINKED ISSUE #6074: Consider removing ".gcpcxx.pb" from generated file names
This is a discussion that we started internally, but I'm moving it to an issue so we don't forget.

I think we probably want to remove the `.gcpcxx.pb` component from autogenerated source files, because it adds noise and an implementation detail that our customers need not be concerned with. As far as our customers are concerned, our C++ libraries are checked in C++ source files, that are all intended directly for customers to read and use. Whether we happen to hand-write a particular file or autogenerate it is an implementation detail that the customer should not care about. Absent a _customer benefit_ from including `.gcpcxx.pb` in path names, I think we should keep things simpler and remove that component.

It's possible that it may benefit _us_ to be able to distinguish generated files from hand-written ones. But again, I'd say that this is an implementation detail that we should shield customers from. We don't have use use the pathname as the indicator for this since the generated files already have comments, we could use something like `grep -rl "Generated by the Codegen C++ plugin"`

I'm not the expert on the generator, but it looks like changing this should be relatively straight forward since this component has already been nicely factored out into its own function:

https://github.com/googleapis/google-cloud-cpp/blob/055f7a35f3acf0025652135d3368a69c3a843e03/generator/internal/codegen_utils.cc#L42

Changing that, plus a few `git mv`s, and I'd imagine that's about it.

I'd be curious to hear if there are good reasons to keep this path component that I've overlooked. Otherwise, we probably want to remove it.

/cc: @scotthart for comments, as the expert on the generator

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6286)

fixes #6074


scotthart:
fixes #6074

ci/check-golden-md5-hashes.sh
```diff
@@ -23,7 +23,9 @@ if [[ "${CHECK_GENERATED_CODE_HASH}" != "yes" ]]; then
   exit 0
 fi
 
-num_current_files=$(find "${PROJECT_ROOT}"/generator/integration_tests/golden -name '*.gcpcxx.*' | wc -l)
+num_current_files=$(find "${PROJECT_ROOT}"/generator/integration_tests/golden/ \
+  -path "${PROJECT_ROOT}"/generator/integration_tests/golden/tests -prune \
+  -false -o -name '*golden*.h' -o -name '*golden*.cc' | wc -l)
```
devbww, line 28: `-false -o ...` looks fishy.

generator/integration_tests/golden/CMakeLists.txt
```diff
@@ -141,6 +141,7 @@ function (google_cloud_cpp_generator_define_golden_tests)
         tests/golden_thing_admin_idempotency_policy_test.cc
         tests/golden_thing_admin_logging_decorator_test.cc
         tests/golden_thing_admin_metadata_decorator_test.cc
+        tests/golden_thing_admin_option_defaults_test.cc
```
coryan, line 144: Is this new?

generator/integration_tests/golden/tests/golden_thing_admin_option_defaults_test.cc
```diff
@@ -32,15 +32,15 @@ TEST(GoldenThingAdminDefaultOptions, DefaultEndpoint) {
 }
 
 TEST(GoldenThingAdminDefaultOptions, EnvVarEndpoint) {
-  internal::SetEnv("GOOGLE_CLOUD_CPP_GOLDEN_KITCHEN_SINK_ENDPOINT",
```
coryan, line 35: These changes look unrelated?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6286)* status: 0 of 134 files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[ci/check-golden-md5-hashes.sh, line 28 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6286#-MYfg0Bdd_QuZbIV5kc3-r2-28:-MYfxQPx-CNdOtYBiuOs:b-5tulfe) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0252dccc108d9dbb82cf829fb831c9ddd79a5b16/ci/check-golden-md5-hashes.sh#L28)):*

`-false -o ...` looks fishy.
</blockquote></details>

It was necessary to add the -false to prevent the directory name from being emitted in the output, which would cause the count to be off by one.

generator/internal/codegen_utils.cc
```diff
@@ -115,7 +115,7 @@ std::string CurrentCopyrightYear() {
   return kCurrentCopyrightYear;
 }
 
-std::string GeneratedFileSuffix() { return ".gcpcxx.pb"; }
+std::string GeneratedFileSuffix() { return ""; }
```
coryan, line 118: Should we remove this function?  Seems pretty useless now.

coryan (commented):
There is a change here that seems unrelated, if you need to have it as part of this PR, just add an explanation in the description.

There are also some additional cleanups, but you can do those in a future PR.

scotthart (commented):
Reverted testing fixes for now; will add in a separate PR.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6286#-:-MYg3FQq2iuxF7TxK5um:beqx1bv)* status: 0 of 136 files reviewed, 3 unresolved discussions (waiting on @coryan and @devbww)

___
*[generator/integration_tests/golden/CMakeLists.txt, line 144 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6286#-MYfg0Bdd_QuZbIV5kc8-r3-144:-MYg34u08yjfICAUyZLS:b-d86860) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7293b0e9a54cd1d29f81f4eba368865c2cf6fe4b/generator/integration_tests/golden/CMakeLists.txt#L144)):*

Is this new?
</blockquote></details>

Reverted for now; will add in a separate PR.
___
*[generator/internal/codegen_utils.cc, line 118 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6286#-MYfg0BeXVZ7Iq2EXWop-r3-118:-MYg3BE1EZKRlKo4eQ-W:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7293b0e9a54cd1d29f81f4eba368865c2cf6fe4b/generator/internal/codegen_utils.cc#L118)):*

Should we remove this function?  Seems pretty useless now.
</blockquote></details>

Done.

devbww:
> It was necessary to add the -false to prevent the directory name from being emitted in the output, which would cause the count to be off by one.

Sorry ... I was confused by implicit precedence.

coryan (approved):

Issue #5821, opened 2021-02-05
generator: modify generator to run emitted code through clang-format


scotthart:
Currently the generated code does is not run through clang-format, nor are the existing golden files. Modify the existing integration test to compare the emitted code to golden code after applying clang-format to both.

scotthart:
Currently running clang-format over emitted code is an additional step in the build process. Determine, and implement if possible, a mechanism to plumb the emitted code through clang-format as part of code generation.

coryan:
We are happy with how it works today.

PR #6290, opened 2021-04-19
refactor(generator): generate libraries using generator config file

scotthart:
Switch generate-libraries build to use new standalone binary for the microgenerator and the new textproto config file.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6290)

### DIFF
--- a/generator/generate-libraries.sh
+++ b/generator/generate-libraries.sh
@@ -26,41 +26,26 @@ fi
 readonly BAZEL_BIN=${BAZEL_BIN:-/usr/local/bin/bazel}
 readonly BAZEL_OUTPUT_BASE=$("${BAZEL_BIN}" info output_base)
 readonly BAZEL_BIN_DIR=$("${BAZEL_BIN}" info bazel-bin)
-readonly BAZEL_DEPS_GOOGLEAPIS_HASH=$("${BAZEL_BIN}" query 'kind(http_archive, //external:com_google_googleapis)' --output=build | sed -n 's/^.*strip_prefix = "googleapis-\(\S*\)"\,.*$/\1/p')
+readonly BAZEL_DEPS_GOOGLEAPIS_HASH=$("${BAZEL_BIN}" query \
+  'kind(http_archive, //external:com_google_googleapis)' --output=build |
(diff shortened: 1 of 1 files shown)


generator/generate-libraries.sh
```diff
@@ -26,41 +26,36 @@ fi
...
+  --googleapis_commit_hash="${BAZEL_DEPS_GOOGLEAPIS_HASH}" \
+  --protobuf_proto_path="${BAZEL_OUTPUT_BASE}"/external/com_google_protobuf/src \
+  --googleapis_proto_path="${BAZEL_OUTPUT_BASE}"/external/com_google_googleapis \
+  --output_path=. \
+  --config_file=generator/generator_config.textproto
 
-  find "${product_path}" \( -name '*.cc' -o -name '*.h' \) -exec clang-format -i {} \;
-done
+io::log_yellow "clang-format generated code."
+find google/cloud -path google/cloud/bigtable -prune -o \
```
devjgm, line 49: Consider for another PR: When the generator binary writes the file perhaps it should just directly clang-format it at that point. It could write the file then call `clang-format -i...` or it could pipe its output through clang-format.

devjgm (approved):

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6290)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @devjgm)

___
*[generator/generate-libraries.sh, line 49 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6290#-MYgX04bUBkLv4zvJPS1-r1-49:-MYkhlVQ-gMqBAsX2GRy:b-8ukjt1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1933f035bdfab4d73b56a3d13ea96050c08c9931/generator/generate-libraries.sh#L49)):*

Consider for another PR: When the generator binary writes the file perhaps it should just directly clang-format it at that point. It could write the file then call `clang-format -i...` or it could pipe its output through clang-format.
</blockquote></details>

Updated #5821 to track this.

PR #6296, opened 2021-04-20
test(storage): fix emulator offset handling

tritone:
Object list calls from the Go library weren't working against
the emulator because the endOffset is sent as an empty string
by the library. This matches the actual service behavior for
this.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6296)


coryan:
The log-linker-pr failure is expected, no worries on that.

coryan (approved):
LGTM, just waiting for the builds.

Issue #6105, opened 2021-03-24
Prototype Unified Authentication Client


coryan:
## Objective

Before we embark in a full implementation of the unified authentication client (aka GUAC), we should prototype the ideas outlined in the design doc. This document can be found (inside Google) at go/cloud-cxx:guac-dd, but the main ideas are reproduced here for ease of access (the names might be different, that is not that important).

## Overview

We will separate the classes that *configure* authentication from the classes that *implement* authentication.  A small hierarchy will represent configuration.  Applications should not need to access this hierarchy directly, instead they would use factory functions.  Something along the lines of (everything in the `google::cloud` namespace unless otherwise specified):

```cc
/// The base class for authentication **configuration**
class Credentials {};

/// A factory function that creates the default configured credentials
std::shared_ptr<Credentials> GoogleDefaultCredentials();

/// A factory function that creates credentials for service account impersonation.
std::shared_ptr<Credentials> ServiceAccountImpersonation(
    std::string target_account, std::shared_ptr<Credentials> credentials);

namespace internal {
class ServiceAccountImpersonationConfig : public Credentials { ... };
class GoogleDefaultCredentialsConfig : public Credentials { ... };
}
```

These configurations are used by gRPC-based libraries to create the right credential types via a visitor:

```cc
namespace internal {
class GrpcVisitor : public CredentialsVisitor {
  public:
    // for illustration only, the actual result would be more complex...
    std::shared_ptr<GrpcAuthenticationStrategy> result() && { return std::move(result_); }

    void visit(GoogleDefaultCredentialsConfig&) override {
      result_.reset(new AuthenticateUsingChannelCredentials(grpc::GoogleDefaultCredentials()));
    }
    void visit(ServiceAccountImpersonationConfig&) override {
      result_.reset(new ...);
    }

  private:
    std::shared_ptr<GrpcAuthenticationStrategy> result_;
};
}
```

A similar approach is used by REST-based libraries (storage is the only one) via a different visitor.

coryan:
I am pausing this while I work on documenting the different gRPC credentials and how they work in practice (see #6154, #6151, etc.)

coryan:
Fixed in #6323

PR #6299, opened 2021-04-20
feat(common): bootstrap Unified Authentication Client

coryan:
Create the initial classes for the "Unified Authentication Client". In
this change we create (non-public) classes to represent either (a) I
want to use the default google credentials, or (b) I want to use an
access token. Future PRs will use these preferences to initialize the
client libraries.

These two cases are intereseting because, while easy to implement, they
represent two main cases for gRPC: a `grpc::ChannelCredentials` that
remains usable for a long time, and short-lived access tokens that are
provided as `grpc::CallCredentials` in each RPC (though probably
computed only once an hour or so).

Part of the changes for #6105

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6299)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6299)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[google/cloud/internal/credentials.cc, line 35 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6299#-MYlSM7YCD3A8guyc2uc:-MYlSM7Z5C-VKCx3_j6v:b-7l8t9u) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7b87f7744ca27fc2216f95ec2427095d5cba8be0/google/cloud/internal/credentials.cc#L35)):*
> ```cpp
> std::shared_ptr<Credentials> AccessTokenCredentials(
>     std::string const& access_token,
>     std::chrono::system_clock::time_point expiration) {
> ```

Why take expiration by-value and not by-ref (const)?

coryan:
___
*[google/cloud/internal/credentials.cc, line 35 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6299#-MYlSM7YCD3A8guyc2uc:-MYlTH4HEMf7gNajyoYu:b-fepcvi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7b87f7744ca27fc2216f95ec2427095d5cba8be0/google/cloud/internal/credentials.cc#L35)):*

Why take expiration by-value and not by-ref (const)?
</blockquote></details>

https://stackoverflow.com/questions/37918168/pass-stdchronoduration-by-value-or-by-reference-to-const

coryan:
___
*[google/cloud/internal/credentials.cc, line 35 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6299#-MYlSM7YCD3A8guyc2uc:-MYlVf89-w6ZZiTtHvxu:b-k4ema0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7b87f7744ca27fc2216f95ec2427095d5cba8be0/google/cloud/internal/credentials.cc#L35)):*

https://stackoverflow.com/questions/37918168/pass-stdchronoduration-by-value-or-by-reference-to-const
</blockquote></details>

But I can change it if you have a strong opinion.

google/cloud/internal/credentials.h
```diff
@@ -0,0 +1,108 @@
...
+namespace google {
+namespace cloud {
+inline namespace GOOGLE_CLOUD_CPP_NS {
+namespace internal {
+class CredentialsVisitor;
+}  // namespace internal
+
+// TODO(#6293) - move unified credentials out of internal namespace
+namespace internal {
+class Credentials {
```
devjgm, line 35: Does this represent multiple credentials or a single credential?

google/cloud/internal/credentials.h
```diff
@@ -0,0 +1,108 @@
...
+ public:
+  virtual ~Credentials() = 0;
+
+ private:
+  friend class CredentialsVisitor;
+  virtual void dispatch(CredentialsVisitor& visitor) = 0;
+};
+
+std::shared_ptr<Credentials> GoogleDefaultCredentials();
+std::shared_ptr<Credentials> AccessTokenCredentials(
```
devjgm, line 45: FWIW: These look like factories, and in Spanner we tended to standardize on `MakeFoo()` naming for factories.

google/cloud/internal/credentials.h
```diff
@@ -0,0 +1,108 @@
...
+    std::string const& access_token,
+    std::chrono::system_clock::time_point expiration);
```
devjgm, line 47: Is there a reason this argument isn't a `struct AccessToken` (defined below)?

google/cloud/internal/credentials.h
```diff
@@ -0,0 +1,108 @@
...
+struct AccessToken {
+  std::string token;
+  std::chrono::system_clock::time_point expiration;
+};
+
+/// A generator for access tokens.
+using AccessTokenSource = std::function<AccessToken()>;
+
+/// A generic generator of access tokens.
+std::shared_ptr<Credentials> DynamicAccessTokenCredentials(
```
devjgm, line 68: Do you think these factories, all of which return a `Credentials` (shared_ptr), would make sense all named `MakeCredentials()` and overloaded based on the argument. Like WDYT about

```cc
std::shared_ptr<Credentials> MakeCredentials();  // Retruns Google Default Credentials
std::shared_ptr<Credentials> MakeCredentials(AccessToken const& tok);
std::shared_ptr<Credentials> MakeCredentails(AccessTokenSource source);
...
```

devjgm (approved):
All LGTM as-is. A few API questions for your consideration.

google/cloud/internal/credentials.h
coryan: Ack. Let's talk about that below.

google/cloud/internal/credentials.h
coryan: No idea. I think common usage is `Credentials`, that is what gRPC calls them, that is how they are documented in wikipedia, e.g.:

https://en.wikipedia.org/wiki/Basic_access_authentication

I think if we start using `Credential` (singular) it will create more confusion than helping.

google/cloud/internal/credentials.h
coryan: I am seriously considering keeping `struct AccessToken` inside the `internal` namespace.

google/cloud/internal/credentials.h
coryan: With respect to the `Make*` vs. not, I have very little opinion, I think `Make*` may avoid ambiguity because too many functions are called `GoogleDefaultCredentials()` already (in different namespaces). I will make that change.

There will be at least a few that take no arguments, e.g., `MakeComputeEngineCredentials()` and `MakeGoogleDefaultCredentials()`.  Some will take only a string, which of course we could wrap in a struct to make the overload works.  But that seems like too much wrapping.

devjgm: A leading `Make` SGTM. Avoiding overloading for the reasons you mentioned SGTM.

PR #6304, opened 2021-04-21
feat(common): support unified credentials in gRPC

coryan:
To use the unified authentication client credentials with gRPC we need
to map said credentials to the gRPC authentication abstractions. These
come in two flavors:

- For short-lived tokens (such as access tokens), one needs to create a
  `grpc::CallCredentials` object and pass it to `grpc::ClientContext`
  before each call.
- For long-lived credentials (e.g. `grpc::GoogleDefaultCredentials`) the
  credentials are associated with the gRPC channel and gRPC takes care
  of refreshing them.

This suggests that we need an abstraction to (a) create properly
configured channels when the unified credentials can be represented by a
`grpc::ChannelCredentials`, and (b) create (and if necessary cache) the
right `grpc::CallCredentials` and then set them before each RPC.

This change introduces the first versions of these components, and
creates a mapping function. For backwards compatibility we can also map
a `std::shared_ptr<grpc::ChannelCredentials>`.

Part of the changes for #6105

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6304)

### DIFF
--- a/google/cloud/internal/grpc_channel_credentials_authentication.h
+++ b/google/cloud/internal/grpc_channel_credentials_authentication.h
@@ -1,0 +1,48 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_GRPC_CHANNEL_CREDENTIALS_AUTHENTICATION_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_GRPC_CHANNEL_CREDENTIALS_AUTHENTICATION_H
+
+#include "google/cloud/internal/unified_grpc_credentials.h"
+#include "google/cloud/version.h"
+#include <grpcpp/grpcpp.h>
+
+namespace google {
+namespace cloud {
+inline namespace GOOGLE_CLOUD_CPP_NS {
+namespace internal {
+
+class GrpcChannelCredentialsAuthentication : public GrpcAuthenticationStrategy {
+ public:
+  explicit GrpcChannelCredentialsAuthentication(
+      std::shared_ptr<grpc::ChannelCredentials> c)
+      : credentials_(std::move(c)) {}
+  ~GrpcChannelCredentialsAuthentication() override = default;
+
+  std::shared_ptr<grpc::Channel> CreateChannel(
+      std::string const& endpoint,
+      grpc::ChannelArguments const& arguments) override;
+  Status Setup(grpc::ClientContext&) override;
+
+ private:
+  std::shared_ptr<grpc::ChannelCredentials> credentials_;
+};
+
+}  // namespace internal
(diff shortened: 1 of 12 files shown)


google/cloud/internal/unified_grpc_credentials.h
```diff
@@ -0,0 +1,55 @@
...
+inline namespace GOOGLE_CLOUD_CPP_NS {
+namespace internal {
+
+class GrpcAuthenticationStrategy {
+ public:
+  virtual ~GrpcAuthenticationStrategy() = default;
+
+  virtual std::shared_ptr<grpc::Channel> CreateChannel(
+      std::string const& endpoint, grpc::ChannelArguments const& arguments) = 0;
+  virtual Status Setup(grpc::ClientContext&) = 0;
```
devjgm, line 37: Is `Setup` allowed to be called multiple times on an instance? The name suggests once, but the fact that it's a method like this suggests multiple times.

If it's intended to be called exactly once, would we be able to remove this function from this interface and instead rely on some factory function to "set up" the appropriate subclass exactly once before returning it? Basically, I'm wondering if this function needs to be part of this interface.

But also, I may not be understanding this strategy class (is it an abstract factory?)

google/cloud/internal/grpc_channel_credentials_authentication.h
```diff
@@ -0,0 +1,48 @@
...
+ public:
+  explicit GrpcChannelCredentialsAuthentication(
+      std::shared_ptr<grpc::ChannelCredentials> c)
+      : credentials_(std::move(c)) {}
+  ~GrpcChannelCredentialsAuthentication() override = default;
+
+  std::shared_ptr<grpc::Channel> CreateChannel(
+      std::string const& endpoint,
+      grpc::ChannelArguments const& arguments) override;
+  Status Setup(grpc::ClientContext&) override;
```
devjgm, line 37: Nit: I believe "setup" is a noun and "set up" is a verb. That's why Googletest uses "SetUp", which frankly annoys me a little, but :man_shrugging: I don't really even care if you change it, because the uppercase U annoys me about as much as the noun-verb thing.... so, just saying. Whatever you decide SGTM.

google/cloud/internal/unified_grpc_credentials.h
coryan: > Is `Setup` allowed to be called multiple times on an instance? The name suggests once, but the fact that it's a method like this suggests multiple times.

You should call it once per `grpc::ClientContext`, to see this in use:

https://github.com/coryan/google-cloud-cpp/blob/c43945c7190a58deea41adb8933ef77ec625548c/google/cloud/storage/internal/grpc_client.cc#L91-L92

and

https://github.com/coryan/google-cloud-cpp/blob/c43945c7190a58deea41adb8933ef77ec625548c/google/cloud/storage/internal/grpc_client.cc#L130-L131

> If it's intended to be called exactly once, would we be able to remove this function from this interface and instead rely on some factory function to "set up" the appropriate subclass exactly once before returning it? 

What is being set up is the `grpc::ClientContext`.  Maybe `ApplyAuthToContext` or `ConfigureThisContext` are better names?

> But also, I may not be understanding this strategy class (is it an abstract factory?)

I think "Strategy" is the right name, it abstracts two (maybe some day more) implementations of "how to configure gRPC channels and context to have proper authentication".

google/cloud/internal/grpc_channel_credentials_authentication.h
coryan: All the more reason to maybe use a completely different name.

google/cloud/internal/unified_grpc_credentials.h
devjgm: Ahh, then I misunderstood its purpose. A new name sounds good. `ConfigureContext()` sounds pretty good to me. Otherwise, whatever more descriptive name you prefer SGTM.

coryan: Done.

coryan:
PTAL

devjgm (approved):

PR #6314, opened 2021-04-21
ci(gcb): add check-abi build

devjgm:
This build does not yet include an "update abi" mode, but that'll come
in a later PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6314)

### DIFF
--- a/ci/cloudbuild/builds/check-abi.sh
+++ b/ci/cloudbuild/builds/check-abi.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+source "$(dirname "$0")/../../lib/init.sh"
+source module ci/cloudbuild/builds/lib/cmake.sh
+
+export CC=gcc
+export CXX=g++
+
+INSTALL_PREFIX=/var/tmp/google-cloud-cpp
+# TODO(#6313): Compile with `-Og`, which abi-dumper wants.
+cmake -GNinja \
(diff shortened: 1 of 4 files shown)


ci/cloudbuild/dockerfiles/fedora.Dockerfile
```diff
@@ -139,6 +139,17 @@ RUN curl -sSL https://github.com/matus-chochlik/ctcache/archive/0ad2e227e8a981a9
     cp clang-tidy-cache /usr/local/bin/clang-tidy-cache && \
     cd /var/tmp && rm -fr build
 
+# Installs Universal Ctags (which is different than the default "Exuberant
```
coryan, line 142: TIL

ci/cloudbuild/builds/check-abi.sh
```diff
@@ -0,0 +1,90 @@
...
+  # namespace. See: https://en.wikipedia.org/wiki/Name_mangling
+  if ! abi-compliance-checker \
+    -skip-internal-symbols "(8internal|15pubsub_internal|16spanner_internal)\d" \
+    -src -l "${lib}" \
+    -old "cmake-out/${expected_dump_file}" \
+    -new "cmake-out/${actual_dump_file}"; then
+    io::log_red "ABI Compliance error: ${lib}"
+    ((++errors))
+    report="cmake-out/compat_reports/${lib}/expected_to_actual/src_compat_report.html"
+    io::log "Report file: ${report}"
```
coryan, line 85: Maybe a `io::log_h1` or `io::log_h2`, I am not sure, just mention it because these dumps are long and sometimes hard to fine where one starts...

coryan (approved):

ci/cloudbuild/builds/check-abi.sh
```diff
@@ -0,0 +1,90 @@
...
+set -eu
+
+source "$(dirname "$0")/../../lib/init.sh"
+source module ci/cloudbuild/builds/lib/cmake.sh
+
+export CC=gcc
+export CXX=g++
+
+INSTALL_PREFIX=/var/tmp/google-cloud-cpp
+# TODO(#6313): Compile with `-Og`, which abi-dumper wants.
```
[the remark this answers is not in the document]
devjgm: It would, but the build breaks when I do that. Wanted to troubleshoot that separately.

ci/cloudbuild/builds/check-abi.sh
devjgm: There's a `io::log_h2` on line 66 that shows up right before these reports get dumped. So actually, they're quite readable now.

PR #6316, opened 2021-04-22
ci(gcb): create secrets if necessary

devjgm:
Our GCB builds expect secrets for the `CODECOV_TOKEN` and the
`LOG_LINKER_PAT`. These exist in our main GCB project, but not in our
personal projects. This PR creates these secrets with empty data if
they don't already exist.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6316)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -278,8 +278,7 @@ if [[ "${project}" != "cloud-cpp-testing-resources" ]]; then
   for secret in "CODECOV_TOKEN" "LOG_LINKER_PAT"; do
     if ! gcloud --project "${project}" secrets describe "${secret}" >/dev/null; then
       io::log_yellow "Adding missing secret ${secret} to ${project}"
-      gcloud --project "${project}" secrets create "${secret}" --data-file=- \
-        </dev/null
+      echo | gcloud --project "${project}" secrets create "${secret}" --data-file=-
     fi
   done
 fi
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/build.sh
```diff
@@ -264,8 +269,20 @@ fi
...
+account="$(gcloud config get-value account 2>/dev/null)"
+# The cloudbuild.yaml file expects certain "secrets" to be present in the
+# project's "Secret Manager". This is true for our main production project, but
+# for personal projects we may need to create them (with empty strings).
+if [[ "${project}" != "cloud-cpp-testing-resources" ]]; then
+  for secret in "CODECOV_TOKEN" "LOG_LINKER_PAT"; do
+    if ! gcloud --project "${project}" secrets describe "${secret}" >/dev/null; then
+      io::log_yellow "Adding missing secret ${secret} to ${project}"
+      gcloud --project "${project}" secrets create "${secret}" --data-file=- \
+        </dev/null
```
devbww, line 282: Just `--data-file=/dev/null`?

devbww (approved):

ci/cloudbuild/build.sh
devjgm: I'm embarrassed that my chain of refactoring stopped just short of the clear expression that you suggest here. Changing... oh, to my surprise that doesn't work. Well, it creates a secret but without any "versions" (i.e., no values associated w/ the secret key), and GCB requires at least one version, so this doesn't work. However, I did change this line to read `echo | gcloud ...`, which is still shorter and clearer than what I had before.

devbww: > Changing... oh, to my surprise that doesn't work. Well, it creates a secret but without any "versions"

That indeed sounds wacky.

PR #6319, opened 2021-04-22
fix(spanner): propagate request_options in ReadPartition serialization

devbww:
When the `ReadRequest` protobuf grew a `RequestOptions` field containing
the request priority, and `ReadOptions` was extended to match, we should
have also extended `ReadPartition` (de)serialization. Do that now so
that the request options are propagated from `Client::PartitionRead()`
through to `Client::Read(ReadPartition)`.


google/cloud/spanner/read_partition.cc
```diff
@@ -36,6 +36,45 @@ ReadPartition::ReadPartition(std::string transaction_id, std::string session_id,
   *proto_.mutable_key_set() = spanner_internal::ToProto(std::move(key_set));
   proto_.set_limit(read_options.limit);
   proto_.set_partition_token(std::move(partition_token));
+  if (read_options.request_priority) {
+    auto* request_options = proto_.mutable_request_options();
+    switch (*read_options.request_priority) {
+      case spanner::RequestPriority::kLow:
```
coryan, line 42: maybe too late for this, but introducing a new enum like this is less future proof than just using an alias for the enum [*].  If they ever add a `PRIORITY_URGENT` level you will need to change this code to make it usable.



[*]: Frankly I think using enums in proto APIs is not the best idea, strings or integers are easier to evolve.

google/cloud/spanner/read_partition.cc
```diff
@@ -36,6 +36,45 @@ ReadPartition::ReadPartition(std::string transaction_id, std::string session_id,
...
+        request_options->set_priority(
+            google::spanner::v1::RequestOptions_Priority_PRIORITY_LOW);
```
coryan, line 44: I think the intended enum name in the generated code is `google::spanner::v1::RequestOptions::PRIORITY_LOW`:

https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#enum

the additional constants `...RequestOptions_Priority_PRIORITY_LOW` are not defined in the spec and (I would guess) can change.

google/cloud/spanner/read_partition.cc
```diff
@@ -36,6 +36,45 @@ ReadPartition::ReadPartition(std::string transaction_id, std::string session_id,
...
+  }
+}
+
+google::cloud::spanner::ReadOptions ReadPartition::ReadOptions() const {
+  google::cloud::spanner::ReadOptions options;
+  options.index_name = proto_.index();
+  options.limit = proto_.limit();
+  if (proto_.has_request_options()) {
+    switch (proto_.request_options().priority()) {
+      case google::spanner::v1::RequestOptions_Priority_PRIORITY_LOW:
```
coryan, line 64: Same thing about enum names.

coryan (approved):
A bit of a nit regarding generated enum names, I trust you will fix if needed or let me know why I am wrong.

google/cloud/spanner/read_partition.cc
devbww: Ack.

google/cloud/spanner/read_partition.cc
devbww: I'll adjust the naming in a followup PR.

google/cloud/spanner/read_partition.cc
devbww: I'll adjust the naming in a followup PR.

PR #6322, opened 2021-04-22
ci(gcb): add generator hash checks to checkers build

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6322)

### DIFF
--- a/ci/cloudbuild/builds/checkers.sh
+++ b/ci/cloudbuild/builds/checkers.sh
@@ -41,3 +41,11 @@ CHECK_STYLE=yes NCPU="$(nproc)" RUNNING_CI="${GOOGLE_CLOUD_BUILD:-no}" \
 
 io::log_h2 "Verifying Markdown"
 CHECK_MARKDOWN=yes ci/check-markdown.sh
+io::log "done"
+
+io::log_h2 "Verifying googleapis commit has for generated code"
+CHECK_GENERATED_CODE_HASH=yes ci/check-generated-code-hash.sh
+
+io::log_h2 "Verifying generator golden file md5 hashes"
+CHECK_GENERATED_CODE_HASH=yes ci/check-golden-md5-hashes.sh
+io::log "done"


ci/cloudbuild/builds/checkers.sh
```diff
@@ -41,3 +41,11 @@ CHECK_STYLE=yes NCPU="$(nproc)" RUNNING_CI="${GOOGLE_CLOUD_BUILD:-no}" \
 
 io::log_h2 "Verifying Markdown"
 CHECK_MARKDOWN=yes ci/check-markdown.sh
+io::log "done"
+
+io::log_h2 "Verifying googleapis commit has for generated code"
+CHECK_GENERATED_CODE_HASH=yes ci/check-generated-code-hash.sh
+
+io::log_h2 "Verifying generator golden file md5 hashes"
+CHECK_GENERATED_CODE_HASH=yes ci/check-golden-md5-hashes.sh
```
coryan, line 50: I still do not see the need to keep these MD5s, the files are committed to git, we are already re-generating the files, all we need to do is run `git diff --exit-code` to see if anything changed.

coryan (approved):

ci/cloudbuild/builds/checkers.sh
devjgm: You may be right. I haven't looked into the contents of these scripts much yet; I was just trying to replicate kokoro behavior in GCB so nothing fell through the cracks.

I _think_ the issue is that current `generate-libraries` build in `ci/kokoro/docker/build.sh` is only run manually. And this MD5 check exists to break the build to alert us to the fact that we need to manually re-run the `generate-libraries` build.

It sounds like what you're proposing is that we have automatic PR/CI builds that run `generate-libraries`, then at that point, we can simply look to see if any files changed, and if so, break the build. I think that would work too and be simpler. I'd be happy to look into doing that once the kokoro builds are gone so that I only need to futz with one build system.

devjgm: Following up. This was done in https://github.com/googleapis/google-cloud-cpp/pull/6344

PR #6328, opened 2021-04-22
ci(gcb): add convenience script for manual testing

coryan:
This build is intended only for manual testing before sending a PR. It
runs the formatting checks, a clang-tidy build, and all the integration
tests that do not depend on production.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6328)

### DIFF
--- a/ci/cloudbuild/builds/smoke-test.sh
+++ b/ci/cloudbuild/builds/smoke-test.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


devjgm:
We had similar PRs cross in the aether. Actually, you were a bit faster, but close! My PR is https://github.com/googleapis/google-cloud-cpp/pull/6329 and the main difference is that it runs the cmake+emulator+integration tests in several builds that are actually triggered, so I think it'll result in more coverage.

However, my PR does not include `checkers` in the clang-tidy run, so you'd still need to run that separately. WDYT?

coryan:
My preference is to use your PR, looking right now.

PR #6335, opened 2021-04-23
ci: remove docker pull noise from main step

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6335)


devjgm (approved):
So the noise is still there but now in a different build step? Still an improvement.

coryan:
> So the noise is still there but now in a different build step? Still an improvement.

Yeah, I tried to do a `docker pull -q` and did not work (something something permissions).   Could not find an option to silence it.  Maybe `docker pull -q --network=cloudbuild ...` would do the trick, we can try that in a future PR.

Issue #6346, opened 2021-04-23
Consider using dependabot or renovatebot to update dependencies


coryan:
We have a number of dependencies which are listed in our `*.Dockerfile` files, `*.cmake` files, and in `WORKSPACE` files. We also have one `requirements.txt` file for Python.

Updating these files is usually easy, but remembering that they need to be updated (or receiving a notification) is tedious and easy to forget.

There are bots to automate these processes for Python, and at least renovatebot promises support for `Bazel` files.  There are "techniques" to deal with tarballs when the bots do not support them directly.

I will document my experience with these bots in this bug and then make a recommendation.

coryan:
dependabot is a GitHub product. It does not support most of the stuff we need, but seems easy to enable the "reminder" PRs. I enabled dependabot in #6339, configured to track Abseil and `google-cloud-cpp` (we have self-referencing `WORKSPACE` files in the quickstarts).

It created two PRs:

- #6343 is a reminder to update the `WORKSPACE` files. Updating the PR with my own commits was not that difficult (see this [blog post] for a tutorial).
  - The experience was frustrating because I made at least two mistakes.
  - At some point the Kokoro builds stopped working.
- #6342 is a reminder to update Abseil.
  - I did not even try to get that to work, closing for now.

[blog post]: https://tighten.co/blog/adding-commits-to-a-pull-request/

coryan:
* Renovatebot did a good job with #6354.
* I have disabled googleapis/googleapis updates for it. I just closed the PR, may want to add [ignoreDeps] for it.
* It may be possible to update the Dockerfiles using [regexManagers]


[regexManagers]: https://docs.renovatebot.com/configuration-options/#regexmanagers
[ignoreDeps]: https://docs.renovatebot.com/configuration-options/#ignoredeps

coryan:
Testing changes to the Renovatebot configuration seems fairly hard, one can validate the format of the configuration file with the `renovate-config-validator` tool (distributed via `npm install renovate`).

But the validation is not complete, and there is no telling what the *effect* of the changes will be, i.e., whether anything will be updated or not.

coryan:
Looks line renovate is updating Dockerfiles, yay!  For example #6374

coryan:
Updating most of the `ci/cloudbuild/dockerfiles/*.Dockerfile` requires manually updating `doc/packaging.md` too.  It would be nice to automate this.

coryan:
[...]

PR #6349, opened 2021-04-24
ci: enable renovate

coryan:
This is an alternative to dependabot.

Part of the work for #6346 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6349)


coryan:
We got an automated PR from renovate, let's use that.

Issue #6384, opened 2021-04-26
[GCB] We need a nightly build


devjgm:
This build runs some of the slow integration tests that we can't or don't want to run on every PR/CI.

PR #6386, opened 2021-04-26
ci(gcb): add integration-daily build to GCB

devjgm:
Fixes: #6384

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6386)

### DIFF
--- a/ci/cloudbuild/builds/integration-nightly.sh
+++ b/ci/cloudbuild/builds/integration-nightly.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ATTENTION: This build is different than most because it's designed to be run
+# as a "manual" trigger that's run on a schedule in GCB. It will be scheduled
+# to right once a day at night. For details about scheduled manual triggers see
+# https://cloud.google.com/build/docs/automating-builds/create-manual-triggers.
+#
+# NOTE: This build script will not have a trigger file in the
+# `ci/cloudbuild/triggers` directory. "Manual" triggers exist only within the
+# GCB UI at the time of this writing. Users with the appropriate access can run
+# this build by hand with:
+#   `ci/cloudbuild/build.sh --distro fedora integration-nightly --project cloud-cpp-testing-resources`
+
+set -eu
+
+source "$(dirname "$0")/../../lib/init.sh"
+source module ci/cloudbuild/builds/lib/bazel.sh
+source module ci/cloudbuild/builds/lib/integration.sh
+source module ci/lib/io.sh
+
+export CC=clang
+export CXX=clang++
+
+mapfile -t args < <(bazel::common_args)
+bazel test "${args[@]}" --test_tag_filters=-integration-test ...
+
+export ENABLE_BIGTABLE_ADMIN_INTEGRATION_TESTS="yes"
+export GOOGLE_CLOUD_CPP_SPANNER_SLOW_INTEGRATION_TESTS="instance,backup"
+mapfile -t integration_args < <(integration::bazel_args)
+integration::bazel_with_emulators test "${args[@]}" "${integration_args[@]}"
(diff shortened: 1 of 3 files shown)


ci/cloudbuild/builds/integration-nightly.sh
```diff
@@ -0,0 +1,44 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ATTENTION: This build is different than most because it's designed to be run
+# as a "manual" trigger that's run on a schedule in GCB. It will be scheduled
+# to right once a day at night. For details about scheduled manual triggers see
```
devbww, line 19: "At night" is in the eye of the beholder.

devjgm: Just copying the existing name. Happy to rename to whatever you suggest.

coryan, line 19: s/right once a day/run once a day/

coryan: consider "once a day, at a time when most of the `google-cloud-cpp` developers are not working" or something like that.

devjgm: SGTM. Changed all references to "nightly" to "daily".

devjgm: Fixed

coryan (approved):

devjgm:
Hmm. The spanner tests keep timing out. Example: https://pantheon.corp.google.com/cloud-build/builds;region=global/ef3cc3f7-4af8-4330-ac91-81f4074ca002;step=3?project=cloud-cpp-testing-resources

Will look into this and try to get one to completely successfully before merging this.

devjgm:
I think I just need to bump the spanner test timeout, because I found a previous run of this test from Kokoro where the times were:

```
//google/cloud/spanner/integration_tests:backup_integration_test         PASSED in 1831.0s
...
//google/cloud/spanner/samples:samples                                   PASSED in 1506.8s
```

coryan:
I think #5926 extended the integration test time for these tests, but in the GCB builds we override that:

https://github.com/googleapis/google-cloud-cpp/blob/7d81e4ead9f66cbb88d684397b757fe13b661b34/ci/cloudbuild/builds/lib/integration.sh#L42

while the Kokoro builds run with the default for Spanner.

devjgm:
> I think #5926 extended the integration test time for these tests, but in the GCB builds we override that:
> 
> https://github.com/googleapis/google-cloud-cpp/blob/7d81e4ead9f66cbb88d684397b757fe13b661b34/ci/cloudbuild/builds/lib/integration.sh#L42
> 
> while the Kokoro builds run with the default for Spanner.

Thanks.

I can bump the spanner test timeout in the GCB build, but we currently have a 3600s timeout on the whole GCB pipeline. 

https://github.com/googleapis/google-cloud-cpp/blob/7d81e4ead9f66cbb88d684397b757fe13b661b34/ci/cloudbuild/cloudbuild.yaml#L45

And that may not be enough.

I'm seeing if I can make that a parameter so that it's only bumped up for the `integration-daily` test and not the others.

devjgm:
> > I think #5926 extended the integration test time for these tests, but in the GCB builds we override that:
> > https://github.com/googleapis/google-cloud-cpp/blob/7d81e4ead9f66cbb88d684397b757fe13b661b34/ci/cloudbuild/builds/lib/integration.sh#L42
> > 
> > while the Kokoro builds run with the default for Spanner.
> 
> Thanks.
> 
> I can bump the spanner test timeout in the GCB build, but we currently have a 3600s timeout on the whole GCB pipeline.
> 
> https://github.com/googleapis/google-cloud-cpp/blob/7d81e4ead9f66cbb88d684397b757fe13b661b34/ci/cloudbuild/cloudbuild.yaml#L45
> 
> And that may not be enough.
> 
> I'm seeing if I can make that a parameter so that it's only bumped up for the `integration-daily` test and not the others.

Actually, as long as the tests run in parallel, (they should), 3600s is probably enough overall. I'll leave that for now and see if we hit it.

ci/cloudbuild/builds/lib/integration.sh
```diff
@@ -153,7 +153,7 @@ function integration::bazel_with_emulators() {
   # - GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT
   io::log_h2 "Running Spanner integration tests"
   bazel "${verb}" "${args[@]}" --test_tag_filters=integration-test \
-    google/cloud/spanner/...
+    --test_timeout=3600 google/cloud/spanner/...
```
devbww, line 156: Why do we need that?  Do the BUILD-file values not apply?

devjgm: It's set to 600 above, so I need to override it here.

Also, `eternal` implies a timeout of 3600s, so this is setting it to the equivalent of eternal.

https://docs.bazel.build/versions/master/test-encyclopedia.html#role-of-the-test-runner

devjgm: ... of course, I could remove the `--test_timeoout` above and always rely on the per-test size setting... that may be better.

devjgm:
FYI, I got a successful run: https://pantheon.corp.google.com/cloud-build/builds/9293a306-94bf-49e6-b978-5f8836ef43f4?project=cloud-cpp-testing-resources ~35 minutes

Issue #6164, opened 2021-04-01
[GCB] Expose build logs publically


devjgm:
This is an OSS project with OSS contributors, and it would be best if OSS contributors could see the build errors for their PRs, so we need a way to make the GCB logs public. Note: that non-googler PRs will not be triggered automatically. This is by design, for security. This is OK. However, once a project owner triggers the build, we'd like the resulting logs to be public.

FWIW, The kokoro logs are public using ResultStore.

devjgm:
It's not easy to make the GCB logs public. However, it should be possible. My current plan is to do the following:

1. Change the GCS bucket where our GCB builds log (https://github.com/googleapis/google-cloud-cpp/pull/6284).
2. Make the logs bucket publicly **readable** (not writeable).This will let any user view our pre-submit and post-submit logs. In particular, users will be able to view the PR build logs for their own PRs.
3. Have a "bot" that posts a comment to each PR saying something like `"GCB Build Logs will be available at http://storage.googleapis.com/cloud-cpp-testing-resources_publiclogs/logs/google-cloud-cpp/<PR-NUMBER>/<COMMIT-SHA>/"`. The "bot" posting that could be a special GCB build, or a GitHub Action, or maybe a GitHub App. Whatever posts the comment would need the appropriate access to post a comment, so, whichever mechanism makes this easist is probably the one we should use. This is likely a pretty simple bot.

devjgm:
1. (above) is done
2. Is being tracked by internal bug: b/185928485
3. Is done

devjgm:
2. above was finished. Yay! I enabled our public build logs.

D'oh! There was no clickable web UI for users to browse GCS to see the logs for their PR. BUMMER!

@coryan saved the day by quickly whipping up https://github.com/googleapis/google-cloud-cpp/pull/6390 which creates an index.html page for each PR+commit that produces a clickable UI for users. Awesome! That closes this issue.

PR #6390, opened 2021-04-27
ci: create an index.html file with the logs for each PR

coryan:
Post the GCB logs as an HTML file at a well known location.

Part of the work for #6164 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6390)


ci/cloudbuild/notifiers/logs/function/function.cc
```diff
@@ -0,0 +1,102 @@
...
+  std::string prefix = "logs/google-cloud-cpp/" + pr_number + "/";
+
+  // TODO(coryan) - this should be in a OCC loop
+  std::ostringstream os;
+  os << "<body>\n";
+  os << "<ul>";
+  for (auto const& object :
+       client.ListObjects(bucket_name, gcs::Prefix(prefix))) {
+    if (!object) throw std::runtime_error(object.status().message());
+    os << "<li><a href=\"" << object->media_link() << "\">" << object->name()
```
devjgm, line 93: It looks like the `media_link()` gives us a URL that results in the browser always downloading the file rather than displaying it in the browser. I think that's convenient than displaying the log contents. If we use a URL like `https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/6390/77fcbdb76087b20a8aa24f1fa069e5577f99137e/debian-github-log-linker/log-5a2129a1-7085-4e97-b51f-5b01a7f38746.txt` my browser (Chrome) shows the contents, which is nice.

It looks like we may be able to construct that URL ourselves just from a fixed prefix w/ the scheme and domain, and `object->name()`.

devjgm, line 93: We may want to skip objects whose `object->name()` ends with `-step-N.txt`. The individual step output I think is not important to end-users. It's a GCB implementation detail. The one log file w/o the step suffix contains all the output for all steps merged, which seems like all the end-users would want. And if we omit the step-output, it'll greatly reduce the log links and make the list more manageable.

devjgm (commented):
This is awesome, and doesn't even look very complicated. Nice! I think this is going to save the day.

I realize you're still experimenting and stuff, but here are two thoughts for you to consider.

ci/cloudbuild/notifiers/logs/function/function.cc
coryan: Done.

coryan: Done.

devjgm (approved):

PR #6393, opened 2021-04-27
ci: build with older versions of each dep

### LINKED ISSUE #6395: [GCB] vcpkg errors with GCS
Maybe we need to install `zip` and some other tools, or maybe the build is configured incorrectly?

```
Step #3: sh: zip: command not found
Step #3: gsutil failed to upload with exit code: 256
Step #3: CommandException: No URLs matched: /workspace/cmake-out/vcpkg/buildtrees/crc32c/x64-linux.zip
Step #3: Uploaded binaries to 0 GCS remotes.
Step #3: sh: zip: command not found
Step #3: Failed to store binary cache /h/.cache/vcpkg/archives/92/926697d4df9301ed4ba18664196f961651db3267.zip: Invalid cross-device link
```

https://pantheon.corp.google.com/cloud-build/builds/12b88506-fd54-496a-9c69-c83c01c98c0b;step=3?project=cloud-cpp-testing-resources

coryan:
This build compiles with older versions of each dependency. Its only
purpose is to verify compilation (and basic testing), so we only run
this in one platform. I used vcpkg to fetch the older versions of the
deps, including their dependencies at the time.

The only magic number is the `vcpkg` baseline, which is obtained by
looking at the history of:

https://github.com/microsoft/vcpkg/blob/master/versions/baseline.json

and finding the commit that has all the versions we want. In this build
vcpkg reports:

```
    abseil[core]:x64-linux -> 2020-09-23#3
    benchmark[core]:x64-linux -> 1.5.2
    c-ares[core]:x64-linux -> 1.17.1
    crc32c[core]:x64-linux -> 1.1.1
    curl[core,non-http,openssl,ssl]:x64-linux -> 7.74.0#3
    grpc[core]:x64-linux -> 1.33.1#3
    gtest[core]:x64-linux -> 1.10.0#4
    nlohmann-json[core]:x64-linux -> 3.9.1
    openssl[core]:x64-linux -> 1.1.1j
    protobuf[core]:x64-linux -> 3.14.0#3
    re2[core]:x64-linux -> 2020-10-01
    upb[core]:x64-linux -> 2020-08-19
    zlib[core]:x64-linux -> 1.2.11#9
```

Fixes #6395 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6393)

### DIFF
--- a/ci/cloudbuild/builds/cmake-deps.sh
+++ b/ci/cloudbuild/builds/cmake-deps.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+source "$(dirname "$0")/../../lib/init.sh"
+source module ci/cloudbuild/builds/lib/cmake.sh
+source module ci/cloudbuild/builds/lib/integration.sh
+
+export CC=clang
+export CXX=clang++
+
+mkdir -p cmake-out
+if [[ ! -d cmake-out/vcpkg ]]; then
+  git -C cmake-out clone https://github.com/microsoft/vcpkg.git
+fi
+VCPKG_ROOT="$(pwd)/cmake-out/vcpkg"
+cmake -GNinja -S . -B cmake-out/build \
+  "-DCMAKE_TOOLCHAIN_FILE=${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake" \
+  "-DVCPKG_MANIFEST_DIR=ci/etc/oldest-deps" \
+  "-DVCPKG_FEATURE_FLAGS=versions,manifest"
+cmake --build cmake-out/build
+env -C cmake-out/build ctest -LE "integration-test" --parallel "$(nproc)"
+
+integration::ctest_with_emulators "cmake-out/build"
--- a/ci/cloudbuild/builds/cmake-oldest-deps.sh
+++ b/ci/cloudbuild/builds/cmake-oldest-deps.sh
@@ -23,13 +23,12 @@ source module ci/cloudbuild/builds/lib/integration.sh
 export CC=clang
 export CXX=clang++
 
-mkdir -p cmake-out
 if [[ ! -d cmake-out/vcpkg ]]; then
+  mkdir -p cmake-out
   git -C cmake-out clone https://github.com/microsoft/vcpkg.git
 fi
-VCPKG_ROOT="$(pwd)/cmake-out/vcpkg"
 cmake -GNinja -S . -B cmake-out/build \
-  "-DCMAKE_TOOLCHAIN_FILE=${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake" \
+  "-DCMAKE_TOOLCHAIN_FILE=${PROJECT_ROOT}/cmake-out/vcpkg/scripts/buildsystems/vcpkg.cmake" \
   "-DVCPKG_MANIFEST_DIR=ci/etc/oldest-deps" \
   "-DVCPKG_FEATURE_FLAGS=versions,manifest"
 cmake --build cmake-out/build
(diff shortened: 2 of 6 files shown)


coryan:
@devjgm I would like your input re: naming for this build, and the location of this config file.  Also any hints on how to create the triggers?

ci/cloudbuild/builds/cmake-deps.sh
```diff
@@ -0,0 +1,38 @@
...
+source "$(dirname "$0")/../../lib/init.sh"
+source module ci/cloudbuild/builds/lib/cmake.sh
+source module ci/cloudbuild/builds/lib/integration.sh
+
+export CC=clang
+export CXX=clang++
+
+mkdir -p cmake-out
+if [[ ! -d cmake-out/vcpkg ]]; then
+  git -C cmake-out clone https://github.com/microsoft/vcpkg.git
```
devjgm, line 28: Do we need to compile the vcpkg binary? Or does that happen automatically when we use the `vcpkg.cmake` file?

I'm doing this in https://github.com/googleapis/google-cloud-cpp/blob/master/ci/cloudbuild/builds/quickstart-cmake.sh, but maybe I don't need to bootstrap it.

ci/cloudbuild/builds/cmake-deps.sh
```diff
@@ -0,0 +1,38 @@
...
+fi
+VCPKG_ROOT="$(pwd)/cmake-out/vcpkg"
```
devjgm, line 30: Nit: No biggie, but I tend to use `${PROJECT_ROOT}` in other scripts rather than `$(pwd)`. Either's fine w/ me, but fyi.

ci/cloudbuild/builds/cmake-deps.sh
```diff
@@ -0,0 +1,38 @@
...
+set -eu
+
+source "$(dirname "$0")/../../lib/init.sh"
+source module ci/cloudbuild/builds/lib/cmake.sh
+source module ci/cloudbuild/builds/lib/integration.sh
+
+export CC=clang
+export CXX=clang++
+
+mkdir -p cmake-out
```
devjgm, line 26: This dir will always exist in our builds (I think). If not, I think it's a bug somewhere.

devjgm (commented):
Re: naming. 

Hmm. I may go with something like `cmake-old-desps` or `cmake-trailing-deps` or "oldest" is fine with me too.

Regarding creating triggers:

```
ci/cloudbuild/trigger.sh --generate cmake-old-deps
git status # trigger files should exist now
```

The trigger files are always created using a distro of `fedora` by default. You can change that by hand if you want to use a different distro. That command only creates the trigger files. GCB still doesn't know about the triggers. This is a feature, otherwise the triggers would fire and fail on other folks' PRs.

I usually import the triggers (i.e., tell GCB about them) right as I'm merging the PR:

```
trigger.sh --import triggers/cmake-old-deps-pr.yaml
trigger.sh --import triggers/cmake-old-deps-ci.yaml
```
You can test this build easily without triggers using
```
build.sh --distro fedora cmake-old-deps # optionally --docker
```
You can also use the local trigger files even before you've imported them to GCB
```
build.sh -t cmake-old-deps-pr
```

All this and a bit more at https://github.com/googleapis/google-cloud-cpp/blob/master/ci/cloudbuild/README.md

ci/cloudbuild/builds/cmake-deps.sh
coryan: I know that the vcpkg binary is compiled automatically in manifest mode.  I am not sure what happens if there is no `vcpkg.json` to point to.

ci/cloudbuild/builds/cmake-deps.sh
coryan: Fixed

ci/cloudbuild/builds/cmake-deps.sh
coryan: I think for `--docker` builds that is true, for `--project` builds that is not true:

https://pantheon.corp.google.com/cloud-build/builds/12b88506-fd54-496a-9c69-c83c01c98c0b;step=3?project=cloud-cpp-testing-resources

coryan:
@devjgm I am not sure if #6395 is introduced in this PR or is an existing bug. Happy to help either way.

ci/cloudbuild/builds/cmake-deps.sh
devjgm: .... and now I remember that. Thanks :-)

devjgm:
> @devjgm I am not sure if #6395 is introduced in this PR or is an existing bug. Happy to help either way.

Try with `ubuntu-bionic`. That's what I use for the `quickstart-cmake` builds that also use vcpkg.

coryan:
> 
> Try with `ubuntu-bionic`. That's what I use for the `quickstart-cmake` builds that also use vcpkg.

As noted in #6395 that does not work, CMake is too old to support manifest mode in that platform.  Installing `zip(1)` fixes the problems, I have included that fix in this PR.

PTAL.

PR #6413, opened 2021-04-29
ci: fix mount point for gcloud config

coryan:
I suggested using a read-only mount, but one needs write permissions to
run the Pub/Sub emulator.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6413)


devjgm (approved):
I tested and didn't notice a problem but the results may have been cached for me.

coryan:
> I tested and didn't notice a problem but the results may have been cached for me.

Maybe my previous version of the file was "wrong" and needed to be rewritten and yours was Okay? :shrug:

PR #6414, opened 2021-04-29
ci: fix abi compat_report to be written in cmake-out

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6414)


devbww (approved):
Thanks.

devjgm:
> Thanks.

YW. Thanks for reporting the bug. FYI, I changed the code to fix it a different way that I like a little better.

PR #6416, opened 2021-04-29
ci: avoid root-owned /h/.config/gcloud dir

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6416)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -231,7 +231,9 @@ if [[ "${DOCKER_FLAG}" = "true" ]]; then
   out_dir="${PROJECT_ROOT}/build-out/${DISTRO_FLAG}-${BUILD_NAME}"
   out_home="${out_dir}/h"
   out_cmake="${out_dir}/cmake-out"
-  mkdir -p "${out_home}" "${out_cmake}"
+  # Create every directory that docker will mount as a volume, otherwise the
+  # docker daemon will create it as a root-owned directory.
+  mkdir -p "${out_cmake}" "${out_home}" "${out_home}/.config/gcloud"
   image="gcb-${DISTRO_FLAG}:latest"
   io::log_h2 "Building docker image: ${image}"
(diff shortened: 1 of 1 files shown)


coryan (approved):

ci/cloudbuild/build.sh
```diff
@@ -231,7 +231,9 @@ if [[ "${DOCKER_FLAG}" = "true" ]]; then
   out_dir="${PROJECT_ROOT}/build-out/${DISTRO_FLAG}-${BUILD_NAME}"
   out_home="${out_dir}/h"
   out_cmake="${out_dir}/cmake-out"
-  mkdir -p "${out_home}" "${out_cmake}"
+  # Create every directory that docker will mount as a volume, otherwise the
+  # docker daemon will create it as a root-owned directory.
+  mkdir -p "${out_cmake}" "${out_home}" "${out_home}/.config/gcloud"
```
devbww, line 236: If you're going to `-p` then there isn't much point in mentioning prefix directories.

devjgm: I'd rather include all 3 mount points explicitly. I think it makes the intent clearer and the code match the comment.

devbww: Then I guess it should say `${HOME}/.config/gcloud` to match the code then.

devjgm: `${HOME}/.config/gcloud` is a different path. The path that will be mounted is down in `build-out/<distro>-<build>/h/.config/gcloud`. Outside of docker, we spell that with `${out_home}`.

devbww: I understand.  I'm just highlighting that "code match the comment" doesn't really apply.  But OK.

devjgm: OK. Removed `${out_home}` from the mkdir line. PTAL.

devbww (approved):

PR #6420, opened 2021-04-30
ci: remove dupe builds

devjgm:
These were the first 4 builds that I added when testing GCB. Now that we
have all the builds, the functionality here, {bazel, cmake} x {gcc,
clang}, is duplicated several times over in other builds. We still have
this cross product covered even without these builds, and removing them
will ease our GCB quota issues.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6420)


coryan (approved):

devjgm:
... of course, I forgot to delete the triggers... duh. Trying again.

devjgm:
we could consider removing `integration.sh` too. All the sanitizer builds run them.

coryan:
> we could consider removing `integration.sh` too. All the sanitizer builds run them.

Agree.

Issue #6309, opened 2021-04-21
Create unified authorization client abstractions for "impersonate a service account"


coryan:
We need a subtype of `google::cloud::[internal::]Credentials` that represents "impersonate a service account".

This will need to call the IAM Service Account Credentials Service:

https://cloud.google.com/iam/docs/reference/credentials/rest

to generate an access token. There are gRPC and REST versions of this RPC, we probably need to implement both (ugh).

We may need to break this down into smaller tasks, this involves:

- [x] Implement the gRPC version of this class, it is tempting to use the micro-generated library, but that adds an unfortunate cycle, it does not require retries, that would be handled by the caller anyway. It does require metadata decorations.
- [x] Implement the REST version of this class, same as gRPC for metadata an retries.
- [x] Implement the `google::cloud::internal::ServiceAccountImpersonationConfig` class, which just stores the data necessary to do the work.
- [x] Weave the new class through the visitors.
- [x] Implement an integration test for REST
- [x] Implement an integration test for gRPC

PR #6425, opened 2021-04-30
feat(storage): minimal stub for IAM credentials service

coryan:
A minimal "stub" to request access tokens for service account
impersonation using REST. This will be used to implement service account
impersonation in the storage library, without requiring any gRPC
linkage.

Part of the work for #6309

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6425)


google/cloud/storage/internal/minimal_iam_credentials_rest.cc
```diff
@@ -0,0 +1,151 @@
...
+        google::cloud::internal::ParseRfc3339(
+            parsed["expireTime"].get<std::string>())};
+  }
+
+ private:
+  std::string MakeRequestUrl(GenerateAccessTokenRequest const& request) {
+    return absl::StrCat(endpoint_, "projects/-/serviceAccounts/",
+                        request.service_account, ":generateAccessToken");
+  }
+  static std::string NormalizeEndpoint(std::string endpoint) {
```
devjgm, line 83: This function could probably be moved to the .cc file in an unnamed namespace. I don't see any need for it to be in a header.

devjgm (approved):

google/cloud/storage/internal/minimal_iam_credentials_rest.cc
coryan: I tricked you: it already is in the `.cc` file... the whole class is in the anonymous namespace in the `.cc` file.. :wink:

devjgm: D'oh! You got me :-)

Issue #6427, opened 2021-04-30
Running demo-*-pr builds locally creates binaries in the source directory


coryan:
Running something like:

```console
> git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
```

```console
./ci/cloudbuild/build.sh -t demo-centos-7-pr --docker
````

Creates binaries in the source directory:

```console
> git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	core
	google/cloud/bigtable/quickstart/quickstart
	google/cloud/pubsub/quickstart/quickstart
	google/cloud/spanner/quickstart/quickstart
	google/cloud/storage/quickstart/quickstart

nothing added to commit but untracked files present (use "git add" to track)
```

devjgm:
Ahh, it's because of this Makefile build:

https://github.com/googleapis/google-cloud-cpp/blob/e8ee1692aed7058fab883347bb303387f75dfde3/ci/cloudbuild/builds/lib/quickstart.sh#L54-L55

We need to point that output somewhere else. I'll take a look at that.

PR #6428, opened 2021-04-30
ci: quickstart Makefile doesn't leave binary in repo

devjgm:
Fixes: #6427

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6428)

### DIFF
--- a/ci/cloudbuild/builds/lib/quickstart.sh
+++ b/ci/cloudbuild/builds/lib/quickstart.sh
@@ -54,8 +54,7 @@ function quickstart::build_cmake_and_make() {
     makefile_bin_dir="${PROJECT_ROOT}/cmake-out/quickstart-makefile-${lib}"
     mkdir -p "${makefile_bin_dir}"
     PKG_CONFIG_PATH="${prefix}/lib64/pkgconfig:${prefix}/lib/pkgconfig:${PKG_CONFIG_PATH:-}" \
-      make -C "${src_dir}"
-    mv "${src_dir}/quickstart" "${makefile_bin_dir}"
+      make -C "${src_dir}" BIN="${makefile_bin_dir}"
   done
 }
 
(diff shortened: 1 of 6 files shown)


ci/cloudbuild/builds/lib/quickstart.sh
```diff
@@ -45,14 +45,17 @@ function quickstart::build_cmake_and_make() {
...
+    cmake_bin_dir="${PROJECT_ROOT}/cmake-out/quickstart-cmake-${lib}"
+    cmake -H"${src_dir}" -B"${cmake_bin_dir}" "-DCMAKE_PREFIX_PATH=${prefix}"
+    cmake --build "${cmake_bin_dir}"
 
     echo
     io::log "[ Make ]"
+    makefile_bin_dir="${PROJECT_ROOT}/cmake-out/quickstart-makefile-${lib}"
+    mkdir -p "${makefile_bin_dir}"
     PKG_CONFIG_PATH="${prefix}/lib64/pkgconfig:${prefix}/lib/pkgconfig:${PKG_CONFIG_PATH:-}" \
       make -C "${src_dir}"
```
coryan, line 57: What do you think of changing the Makefile to say something like:

```
BIN=.

....

$(BIN)/quickstart: ...
```

and then pass `BIN=${cmake_bin_dir}` as an argument to Make?  I would think it is better to never store the files in the source rather than move them, maybe?

coryan (approved):

ci/cloudbuild/builds/lib/quickstart.sh
devjgm: That seems better. My make-fu is rusty, so I'm glad you gave me some pointers here.

devjgm: Done. PTAL.

PR #6429, opened 2021-04-30
feat(common): credentials for service account impersonation

coryan:
This introduces a unified representation for service account
impersonation credentials. Service account impersonation credentials
allow a program to run with a given set of credentials (the base
credentials), but make RPCs using the credentials of a target service
account. This is often used to delegate billing to another service
account, or to support some forms of multi-tenancy.

Part of the changes for #6309

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6429)

### DIFF
--- a/google/cloud/internal/credentials.cc
+++ b/google/cloud/internal/credentials.cc
@@ -43,9 +43,8 @@ MakeImpersonateServiceAccountCredentials(
   opts = MergeOptions(
       std::move(opts),
       Options{}
-          .set<LifetimeOption>(std::chrono::seconds(std::chrono::hours(1)))
           .set<ScopesOption>({"https://www.googleapis.com/auth/cloud-platform"})
-          .set<DelegatesOption>({}));
+          .set<LifetimeOption>(std::chrono::seconds(std::chrono::hours(1))));
   return std::make_shared<ImpersonateServiceAccountConfig>(
       std::move(base_credentials), std::move(target_service_account),
       std::move(opts));
--- a/google/cloud/internal/credentials.cc
+++ b/google/cloud/internal/credentials.cc
@@ -36,6 +36,30 @@ std::shared_ptr<Credentials> MakeAccessTokenCredentials(
   return std::make_shared<AccessTokenConfig>(access_token, expiration);
 }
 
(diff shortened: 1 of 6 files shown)


google/cloud/internal/credentials.cc
```diff
@@ -36,6 +36,30 @@ std::shared_ptr<Credentials> MakeAccessTokenCredentials(
...
+std::shared_ptr<ImpersonateServiceAccountConfig>
+MakeImpersonateServiceAccountCredentials(
+    std::shared_ptr<Credentials> base_credentials,
+    std::string target_service_account, Options opts) {
+  opts = MergeOptions(
+      std::move(opts),
+      Options{}
+          .set<LifetimeOption>(std::chrono::seconds(std::chrono::hours(1)))
+          .set<ScopesOption>({"https://www.googleapis.com/auth/cloud-platform"})
+          .set<DelegatesOption>({}));
```
devjgm, line 48: I think there should be no need to set an empty option like this, unless you're relying on `.has<T>()` always returning true. IIRC, `.get<T>()` and `.lookup<T>()` should do the Right Thing, even if this option didn't previously exist.

google/cloud/internal/credentials.h
```diff
@@ -47,6 +48,18 @@ std::shared_ptr<Credentials> MakeAccessTokenCredentials(
     std::string const& access_token,
     std::chrono::system_clock::time_point expiration);
 
+struct DelegatesOption {
```
devjgm, line 51: Q: Is the plan that all of the user-visible options for the new credentials library will exist in this single `credentials.h` file? Or will they be spread around in a few files? I ask because it would be nice for them to exist in a single file. This file (`credentials.h`) is fine with me. Otherwise, consider a common location like `credentials_options.h` or similar.

devjgm (approved):

google/cloud/internal/credentials.cc
coryan: Fixed.

google/cloud/internal/credentials.h
coryan: I was planning to keep them all in `credentials.h`, we will see, I expect maybe another handful of options.

PR #6453, opened 2021-05-04
fix(storage): update emulator to use MultipartDecoder

frankyn:
I saw string manipulation to handle multipart and found MultipartDecoder class as a potential replacement supported in `requests-toolbelt`.

Additional changes:
1. Updated emulator README.md on how to run test.
2. Ran linter

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6453)

### DIFF
--- a/google/cloud/storage/emulator/gcs/bucket.py
+++ b/google/cloud/storage/emulator/gcs/bucket.py
@@ -169,8 +169,8 @@ class Bucket:
                 )
             cls.__insert_predefined_acl(metadata, predefined_acl, context)
         if len(metadata.default_object_acl) == 0:
-            predefined_default_object_acl = utils.acl.extract_predefined_default_object_acl(
-                request, context
+            predefined_default_object_acl = (
+                utils.acl.extract_predefined_default_object_acl(request, context)
             )
             if (
                 predefined_default_object_acl
--- a/google/cloud/storage/emulator/gcs/bucket.py
+++ b/google/cloud/storage/emulator/gcs/bucket.py
@@ -169,9 +169,9 @@ class Bucket:
                 )
             cls.__insert_predefined_acl(metadata, predefined_acl, context)
         if len(metadata.default_object_acl) == 0:
-            predefined_default_object_acl = (
-                utils.acl.extract_predefined_default_object_acl(request, context)
-            )
+            predefined_default_object_acl =  utils.acl.extract_predefined_default_object_acl(
+                request, context)
+
             if (
                 predefined_default_object_acl
                 == CommonEnums.PredefinedObjectAcl.PREDEFINED_OBJECT_ACL_UNSPECIFIED
--- a/google/cloud/storage/emulator/gcs/bucket.py
+++ b/google/cloud/storage/emulator/gcs/bucket.py
@@ -169,9 +169,9 @@ class Bucket:
                 )
             cls.__insert_predefined_acl(metadata, predefined_acl, context)
         if len(metadata.default_object_acl) == 0:
-            predefined_default_object_acl =  utils.acl.extract_predefined_default_object_acl(
-                request, context)
-
+            predefined_default_object_acl = utils.acl.extract_predefined_default_object_acl(
+                request, context
+            )
             if (
                 predefined_default_object_acl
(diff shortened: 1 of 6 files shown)


google/cloud/storage/emulator/gcs/bucket.py
```diff
@@ -169,8 +169,8 @@ def init(cls, request, context, rest_only=None):
                 )
             cls.__insert_predefined_acl(metadata, predefined_acl, context)
         if len(metadata.default_object_acl) == 0:
-            predefined_default_object_acl = utils.acl.extract_predefined_default_object_acl(
-                request, context
+            predefined_default_object_acl = (
+                utils.acl.extract_predefined_default_object_acl(request, context)
```
coryan, line 173: As you probably noticed the builds do not like this formatting for some reason?

coryan (changes requested):

google/cloud/storage/emulator/tests/test_gcs.py
```diff
@@ -314,7 +314,7 @@ def test_init_multipart(self):
         request = utils.common.FakeRequest(
             args={},
             headers={"content-type": "multipart/related; boundary=foo_bar_baz"},
-            data=b'--foo_bar_baz\r\nContent-Type: application/json; charset=UTF-8\r\n{"name": "object", "metadata": {"key": "value"}}\r\n--foo_bar_baz\r\nContent-Type: image/jpeg\r\n123456789\r\n--foo_bar_baz--\r\n',
+            data=b'--foo_bar_baz\r\nContent-Type: application/json; charset=UTF-8\r\n\r\n{"name": "object", "metadata": {"key": "value"}}\r\n--foo_bar_baz\r\nContent-Type: image/jpeg\r\n\r\n123456789\r\n--foo_bar_baz--\r\n',
```
tritone, line 317: What's the reason for this change? Will it affect whether request from other clients are acceptable to the emulator?

google/cloud/storage/emulator/utils/common.py
```diff
@@ -245,43 +249,19 @@ def parse_multipart(request):
...
-            media = media[2:]
-        return {"content-type": value}, media
-
-    boundary = boundary.encode("utf-8")
     body = extract_media(request)
-    parts = body.split(b"--" + boundary)
-    if parts[-1] != b"--\r\n" and parts[-1] != b"--":
-        utils.error.missing("end marker (--%s--) in media body" % boundary, None)
-    resource = parse_metadata(parts[1])
+    try:
```
tritone, line 253: Awesome to be able to delete all this stuff!

google/cloud/storage/emulator/tests/test_gcs.py
vnvo2409: IIRC, the reason why I use `\r\n\r\n` in the first place was because I noticed that sometime `google-cloud-cpp` uses `\r\n` and sometime `\r\n\r\n`. You could see my comment here:

https://github.com/googleapis/google-cloud-cpp/blob/e678e4ee830ff171a4e2ec95070525fc9d4b107c/google/cloud/storage/emulator/utils/common.py#L273

frankyn: Thanks @vnvo2409, I missed that comment. Maybe that's something that can be fixed in the C++ client instead, but the GCS API does allow it if different requests from the C++ client use it.

coryan: I think (a) each header is required to end with `\r\n`, and (b) there is supposed to be an empty line after all the headers and before the payload.  In other words, the C++ client is *required* (but maybe does not) send a `\r\n\r\n` at the end of the last header.

It seems to do so:

https://github.com/googleapis/google-cloud-cpp/blob/afc9d24e8ceb3a19826e4d2f533950460f3c6a18/google/cloud/storage/internal/curl_client.cc#L1375

Since the tests pass with the new multipart decoder I am inclined to say that whatever bugs we have in the C++ client library can be fixed later.

coryan (commented):
The changes LGTM.  I will wait until the Windows and macOS builds complete and then merge.
Thanks for the change, really neat.

coryan (approved):

PR #6455, opened 2021-05-04
feat(common): a wrapper for streaming write RPCs

coryan:
Introduce a wrapper for streaming write RPCs. This is similar to
`google::cloud::internal::StreamingReadRpc<>`, with some important
differences:

- There is an extra template parameter to represent the final response.
  Note that this value only makes sense when the operation completes
  successfully.
- There is a `WritesDone()` member function to close the stream, which
  returns a `StatusOr<ResponseType>`. Sometimes the application wants
  to close the stream without sending more data.

Part of the changes for #6309 but really all of GUAC for the storage library.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6455)

### DIFF
--- a/google/cloud/internal/streaming_write_rpc.h
+++ b/google/cloud/internal/streaming_write_rpc.h
@@ -141,4 +141,4 @@ class StreamingWriteRpcImpl
 }  // namespace cloud
 }  // namespace google
 
-#endif  // GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_STREAMING_READ_RPC_H
+#endif  // GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_STREAMING_WRITE_RPC_H
--- a/google/cloud/internal/streaming_write_rpc.h
+++ b/google/cloud/internal/streaming_write_rpc.h
@@ -1,0 +1,144 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_STREAMING_WRITE_RPC_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_STREAMING_WRITE_RPC_H
+
+#include "google/cloud/grpc_error_delegate.h"
+#include "google/cloud/status.h"
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include "absl/types/variant.h"
+#include <grpcpp/grpcpp.h>
+#include <grpcpp/support/sync_stream.h>
+#include <memory>
+
+namespace google {
+namespace cloud {
+inline namespace GOOGLE_CLOUD_CPP_NS {
+namespace internal {
+
+/**
+ * Defines the interface for wrappers around gRPC streaming write RPCs.
+ *
+ * We wrap the gRPC classes used for streaming write RPCs to (a) simplify the
+ * memory management of auxiliary data structures, (b) enforce the "rules"
+ * around calling `Finish()` before deleting an RPC, (c) allow us to mock the
+ * classes, and (d) allow us to decorate the streaming RPCs, for example for
+ * logging.
+ *
+ * This class defines the interface for these wrappers.  The canonical
+ * implementation is `StreamingWriteRpcImpl<RequestType, ResponseType>`
+ */
+template <typename RequestType, typename ResponseType>
+class StreamingWriteRpc {
+ public:
+  /**
+   * Possible results for a Write() operation.
+   *
+   * A Write() operation can have three possible outcomes:
+   * - The Write was successful, continue using the stream, this is represented
+   *   by the `absl::monostate` branch
+   * - The streaming RPC is now closed, either with an error (the `Status`
+   *   branch), or with a successful final response (the `ResponseType`) branch.
+   *   maybe as a result of the Write() as result.
+   *
+   * @note the stream may be closed because of the `Write()` call (e.g., if it
+   * contains the `last_message()` bit), or because the service has closed the
+   * stream, or because of network or service errors.
+   */
+  using WriteOutcome = absl::variant<absl::monostate, Status, ResponseType>;
+
+  virtual ~StreamingWriteRpc() = default;
+
+  /// Cancel the RPC, this is needed to terminate the RPC "early".
+  virtual void Cancel() = 0;
+
+  /// Return the next element, or the final RPC status.
+  virtual WriteOutcome Write(RequestType const& r, grpc::WriteOptions o) = 0;
+
+  /// Half-close the stream and wait for a response.
+  virtual StatusOr<ResponseType> WritesDone() = 0;
+};
+
+/// Report the errors in a standalone function to minimize includes
+void StreamingWriteRpcReportUnhandledError(Status const& status,
+                                           char const* tname);
+
+/**
+ * Implement `StreamingWriteRpc<RequestType, ResponseType>` using the gRPC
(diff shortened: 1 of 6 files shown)


google/cloud/internal/streaming_write_rpc.h
```diff
@@ -0,0 +1,144 @@
...
+  virtual ~StreamingWriteRpc() = default;
+
+  /// Cancel the RPC, this is needed to terminate the RPC "early".
+  virtual void Cancel() = 0;
+
+  /// Return the next element, or the final RPC status.
+  virtual WriteOutcome Write(RequestType const& r, grpc::WriteOptions o) = 0;
+
+  /// Half-close the stream and wait for a response.
+  virtual StatusOr<ResponseType> WritesDone() = 0;
```
devjgm, line 72: I'm wondering about the `Write()` and `WritesDone()` methods and noticing that they both (potentially) return a `ResponseType`, and it's not clear to me if I should call `WritesDone()` if I already got a `ResponseType` from `Write()`. This may just be the nature of gRPC streaming writes, but I'm also wondering if there's any way to simplify it, especially given that I'm not burdened by actually understanding how any of this works ;-)

What do you think about something like:

```c++
  // Writes data. Can be called until Status returns not OK
  virtual Status Write(RequestType const&, WriteOptions) = 0;

  // Returns the error status or ResponseType. Callers should always call this function to close the connection.
  // Call this after writing data with `Write()`
  virtual StatusOr<ResponseType> Close();
```

Does anything like this even make sense? Even if we don't like this API, it'll help tell me if I even understand the concepts ;-) 

No requirements here, just a thought for discussion.

google/cloud/internal/streaming_write_rpc.h
```diff
@@ -0,0 +1,144 @@
...
+   *   by the `absl::monostate` branch
+   * - The streaming RPC is now closed, either with an error (the `Status`
+   *   branch), or with a successful final response (the `ResponseType`) branch.
+   *   maybe as a result of the Write() as result.
+   *
+   * @note the stream may be closed because of the `Write()` call (e.g., if it
+   * contains the `last_message()` bit), or because the service has closed the
+   * stream, or because of network or service errors.
+   */
+  using WriteOutcome = absl::variant<absl::monostate, Status, ResponseType>;
```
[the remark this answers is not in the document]
coryan: There is an actual difference:

* A completed `Read()` either gets a value or the final status for the stream.
* A completed `Write()` can get:
  * nothing: the `Write()` was successful and the stream is still open for more `Write()` calls.
  * the response for the RPC: the `Write()` was successful, and the streaming closed as a side-effect ([example](https://github.com/googleapis/googleapis/blob/8e2d88f5367380b7d28e7208270522ff3675c710/google/storage/v1/storage.proto#L1362))
  * an error: the `Write()` failed or not, but in any case the stream is now closed and finished with an error status.

google/cloud/internal/streaming_write_rpc.h
coryan: I think I like  `Close()` better.  For background (only): I think `WritesDone()` (as opposed to `Close()`) is motivated by bi-directional streams where you can "half-close" the writes, and continue reading for a long time afterwards.  In raw gRPC a bidirectional stream IS-A (as in public inheritance) write stream and also a read stream:

https://grpc.github.io/grpc/cpp/classgrpc_1_1_client_reader_writer_interface.html

In that context is kind of makes sense to decouple the "close" vs. "half-close".

I am not sure we need to preserve those properties.  I doubt there is a realistic use-case where somebody uses a bi-directional RPC but at other times they use a streaming-read RPC (with the same return types).  If needed, we can use composition to fit one style into the other.

coryan:
PTAL

google/cloud/internal/streaming_write_rpc.h
devjgm: The naming of Close vs WritesDone wasn't the main point. Either of those names is fine with me. My main point/question was about how the API would be used. It seems a little complicated with both Write and Close returning a ResultType, and it's not clear to me if I'm supposed to always call Close or what.

But I don't think it's problem. Likely just some of my own misunderstanding. It's your call, and fine as-is.

devjgm (approved):

google/cloud/internal/streaming_write_rpc.h
coryan: Maybe a short GVC to discuss?  All the alternatives are pretty ugly, and I have no strong opinions on this API. And I agree the proposed one is confusing.

Let's talk about the underlying gRPC APIs for a second:

- You make a streaming writes RPC
- You pass a pointer to a `ResponseType` object, where the result of the streaming write will be stored **when and if** the streaming RPC ends successfully.
- This pointer must last until the end of the streaming RPC
- You get back a `std::unique_ptr<grpc::ClientWriter<RequestType>>`.

This `ClientWriter<W>` has several member functions, `.Write()` to send more data, `.WritesDone()` to indicate that you will send no more data, `.Finish()` to collect the status when the stream closes.

There are ~two~three ways to gracefully terminate a streaming RPC:

- The service may have defined a specially message attribute that indicates "this is the last message", in that case `.Write()` returns `true` and you are supposed to know to call `.Finish()` at that time (it seems you can call `.WritesDone()` without harm too).
- The "application" (or library really) can call `.WritesDone()` (on the underlying gRPC type)
- The "application" (or library really) can call `.Write()` and provide the `last_message()` flag, which is equivalent (we are told in the documentation) to calling `.Write()` and then `.WritesDone()`

There is also a way to "ungracefully" terminate the stream, you call `.Write()` (or `.WritesDone()`) it returns `false` and then `.Finish()` returns the error status.  Except that it could return success, maybe it shouldn't, but it could.

If (and only if) `Finish()` returns success you can examine the object of `ResponseType` for its value.

As usual, there is a `gpc::ClientContext&` that must remain valid through the life of the streaming RPC.

So depending on the specific service there are two ways to use the proposed `internal::StreamingWriteRpc<>`:

```cc
StatusOr<ResponseType> SendStuff(
       StreamingWriteRpc<RequestType, ResponseType> stream, 
      std::vector<RequestType> stuff) {
  for (auto& request : stuff) {
    auto w = stream.Write(request, grpc::WriteOptions{});
    if (absl::holds_alternative<absl::monostate>(w)) continue;
    if (absl::holds_alternative<Status>(w)) return absl::get<Status>(w);
    assert(!absl::holds_alternative<RequestType>(w));
  }
  return stream.Close();
}


StatusOr<ResponseType> SendStuffWithClosingMessage(
       StreamingWriteRpc<RequestType, ResponseType> stream, 
      std::vector<RequestType> stuff, RequestType closing) {
  for (auto& request : stuff) {
    auto w = stream.Write(request, grpc::WriteOptions{});
    if (absl::holds_alternative<absl::monostate>(w)) continue;
    if (absl::holds_alternative<Status>(w)) return absl::get<Status>(w);
    assert(!absl::holds_alternative<RequestType>(w));
  }
  auto w = stream.Write(closing, grpc::WriteOptions{}
    .set_last_message() // optional, but probably a good idea.
  );
  assert(!absl::holds_alternative<absl::monostate>(w));
  if (absl::holds_alternative<Status>(w)) return absl::get<Status>(w);
  return absl::get<ResponseType>(w);
}
```

Maybe we want two different streaming classes depending on how it should be used?  In one case you want to use `.Close()`, in the other you shouldn't?

devjgm: A GVC SGTM. But let me jot down some code here first that we can talk about. Thinking about the API that I think I'd _want_ as a user of this API, it would look something like this:

```c++
StatusOr<ResponseType> SendStuff(
    StreamWriterRpc<RequestType, ResponseType> stream,
    std::vector<RequestType> stuff) {
  for (auto const& request : stuff) {
    if (!stream.Write(request, grpc::WriteOptions{})) break;
  }
  return stream.Close();
}
```

I think that code would look exactly the same even if using `grpc::WriteOptions{}.set_last_message()`. What I like about this is that
1. There's only one place to get the `ResponseType` -- from `Close()`.
2. There's only one way to write the loops: Iterate calling `Write()`, then call `Close()` -- always. 
3. No new vocabulary; just vanilla `Status` and `StatusOr`

coryan: FTR, we had a short meeting. I like @devjgm proposal better.  The tricky bit is when `.Close()` is this proposal calls `.WritesDone()` or not. We convinced ourselves that:

- If the application made a `.Write()` call with the `grpc::WriteOptions::last_message()` flag we should not call `.WritesDone()`.
- Otherwise we always call `.WritesDone()`, including the case where there is a service-specific "closing flag" in the proto we can safely call `.WritesDone()` in `.Close()`.

devjgm (approved):

Issue #6458, opened 2021-05-04
Consider if there's a way cancel existing PR builds when a new commit is pushed


devjgm:
When we push a new commit for a PR, there may still be a bunch of queued builds that will all run first. It may be nice if we could configure GCB and/or Kokoro to cancel existing builds when a new commit is pushed.

For GCB there is a `gcloud builds cancel` command that we may be able to leverage. Possibly our logs notifier could do this automatically for us? Or maybe we could just provide a command-line tool that we run manually when needed.

Not sure what to do about kokoro.

devjgm:
Update: @coryan made a change to Kokoro that made it so our ongoing kokoro builds will be cancelled when a new commit is pushed. This is great.

I'm not looking to see if there's something we can do for GCB.

devjgm:
https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/cancelot looks promising

PR #6470, opened 2021-05-05
ci(gcb): cancel ongoing builds on new PR push

devjgm:
Fixes: #6458

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6470)

### DIFF
--- a/ci/cloudbuild/cloudbuild.yaml
+++ b/ci/cloudbuild/cloudbuild.yaml
@@ -143,10 +143,10 @@ steps:
         xargs -r -P 4 -L 32 gcloud container images delete -q --force-delete-tags
       exit 0
 
-  # Cancels all ongoing builds for the current build name and PR that don't
-  # match the new COMMIT_SHA. This effectively means that anytime we push a new
-  # commit, all ongoing builds are cancelled in favor of starting builds for
-  # the new commit. This will only happen for PRs (i.e., not for CI builds).
+  # Cancels all ongoing builds for the current PR that don't match the new
+  # COMMIT_SHA. This effectively means that anytime we push a new commit, all
+  # ongoing builds are cancelled in favor of starting builds for the new
+  # commit. This will only happen for PRs (i.e., not for CI builds).
 - name: 'gcr.io/google.com/cloudsdktool/cloud-sdk'
   waitFor: [ '-' ]
   entrypoint: 'bash'
--- a/ci/cloudbuild/cloudbuild.yaml
+++ b/ci/cloudbuild/cloudbuild.yaml
@@ -143,20 +143,6 @@ steps:
         xargs -r -P 4 -L 32 gcloud container images delete -q --force-delete-tags
       exit 0
 
-  # Cancels in-progress builds for previous commits in the current PR so we can
-  # free up resources to start running the builds for the new commit.
-- name: 'gcr.io/google.com/cloudsdktool/cloud-sdk'
-  waitFor: [ '-' ]
-  entrypoint: 'bash'
-  args:
-    - '-c'
-    - |
-      set -xe
-      ctime="$(gcloud builds describe --format 'value(create_time)' '${BUILD_ID}')"
-      gcloud builds list --ongoing --format='value(id)' --filter \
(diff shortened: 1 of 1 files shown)


devjgm:
Here's a log to a build that detected that a newer commit was queued up and so it cancelled itself:
https://pantheon.corp.google.com/cloud-build/builds;region=global/9389a873-1133-44a2-a8d1-d7300a343ef0;step=8?project=cloud-cpp-testing-resources

devjgm:
Here's an example of a newer build that detected that older builds were still running and cancelled them:
https://pantheon.corp.google.com/cloud-build/builds;region=global/8a9f9719-e0e7-44c3-b58f-df58fee77117;step=7?project=cloud-cpp-testing-resources

devjgm:
Here's [builds that were cancelled](https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/6470/443b76b7a325a8c3faf90240a71b1f2d8fa28b97/index.html)
Here's [the newer builds that replaced the cancelled builds](https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/6470/363e90356b0ec064767cbe5d6dbafb70101bca80/index.html)

ci/cloudbuild/cloudbuild.yaml
```diff
@@ -142,3 +142,38 @@ steps:
         xargs -r printf "gcr.io/${PROJECT_ID}/${_IMAGE}@%s\n" | \
         xargs -r -P 4 -L 32 gcloud container images delete -q --force-delete-tags
       exit 0
+
+  # Cancels *this* specific build one is already queued up for a newer commit.
```
coryan, line 146: s/build one/build **if** one/ ?

ci/cloudbuild/cloudbuild.yaml
```diff
@@ -142,3 +142,38 @@ steps:
...
+      ctime="$(gcloud builds describe --format 'value(create_time)' '${BUILD_ID}')"
+      query="tags=pr"
+      query+=" AND tags=${_PR_NUMBER}"
+      query+=" AND tags=${_BUILD_NAME}"
+      query+=" AND substitutions.COMMIT_SHA != ${COMMIT_SHA}"
+      query+=" AND create_time > ${ctime}"
+      gcloud builds list --limit 1 --format='value(id)' --filter "${query}" | \
+        grep . && gcloud builds cancel '${BUILD_ID}' || exit 0
+
+  # Cancels in-progress builds for previous commits in the current PR so we can
```
coryan, line 164: This one took more thinking to convince myself it was safe.  You are cancelling *all* the builds for the previous SHA.  That might include a build that does not yet have a replacement, but presumably the replacement will be created shortly?

coryan (approved):
One nit and a question, I think you can merge whatever the answer is, but I am curious if my reasoning is correct.

ci/cloudbuild/cloudbuild.yaml
devjgm: Fixed. Thanks.

ci/cloudbuild/cloudbuild.yaml
devjgm: yes, that's the idea. Actually, it cancels all builds for _all_ previous SHAs, which could be more than one. Yes, even if a replacement build doesn't exist yet, one should be created.

My reasoning was that once a new commit is pushed, we really don't care about any builds on the previous commits, so cancel 'em all.

devjgm: Also, I did consider only cancelling old instances of the current build. For example, the `gcc-58-pr` build would only cancel old instances of that particular build. In a way, that seems safer.

The reason I didn't do that was because we hit the GCB max quota limit on every PR push, so many newer builds wouldn't even be scheduled to run until the in-progress older instances finished and freed up a slot. So instead, this allows _any_ build from the "new commit" to cancel _all_ of the builds for older commits to hopefully free up more quota quickly.

Note: If there was no quota limitation, I probably wouldn't have done it this way.

Issue #3446, opened 2020-03-06
Storage: Test Bench does not support gzip'd body's in request


frankyn:
While getting the Java GCS client to work with the Storage Test Bench, I found that the test bench does not support gzip compressed request body's. 

This is a feature request to introduce support for gzip content.

Fragment code that helped me get around this (only 1 example it should be addressed for all JSON requests):
```python
def check_content_encoding(request, data):
    if request.headers.get('Content-Encoding', '') == 'gzip':
        return gzip.decompress(data)
    return data


def extract_media(request):
    """Extract the media from a flask Request.

    To avoid race conditions when using greenlets we cannot perform I/O in the
    constructor of GcsObjectVersion, or in any of the operations that modify
    the state of the service.  Because sometimes the media is uploaded with
    chunked encoding, we need to do I/O before finishing the GcsObjectVersion
    creation. If we do this I/O after the GcsObjectVersion creation started,
    the the state of the application may change due to other I/O.

    :param request:flask.Request the HTTP request.
    :return: the full media of the request.
    :rtype: str
    """
    data = None
    if request.environ.get('HTTP_TRANSFER_ENCODING', '') == 'chunked':
        return check_content_encoding(request, request.environ.get('wsgi.input').read())
    else:
        return check_content_encoding(request, request.data)
```

PR #6472, opened 2021-05-05
fix(storage): support gzip request body in emulator

frankyn:
Adding a middleware to decompress request bodies that are gzip'd which is used by Go and Java from integration testing against the emulator. 

Fixes: https://github.com/googleapis/google-cloud-cpp/issues/3446

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6472)


tritone (approved):
Thanks for implementing this @frankyn , looks great!

coryan (approved):
nit: the description say "from trials", maybe "from integration tests" ?

PR #6474, opened 2021-05-05
feat(storage): support method overrides in emulator

frankyn:
Cloud Storage supports an override header `X-Http-Method-Override` for PATCH requests over POST and the Java Storage library uses it. Wow. This adds support to the Storage emulator as well.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6474)


coryan (approved):
Approved.
Part of me wonders why Java needed to do this. Then I remind myself not to ask questions unless I really want the answer.

PR #6477, opened 2021-05-05
fix(storage): address behavior bugs

frankyn:
Fix two issues:

1. Allow object name in either resource or query parameter for resumable uploads to support Storage libs (they mix and match yay...)
2. Fix a code typo.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6477)

### DIFF
--- a/google/cloud/storage/emulator/gcs/holder.py
+++ b/google/cloud/storage/emulator/gcs/holder.py
@@ -59,21 +59,31 @@ class DataHolder(types.SimpleNamespace):
 
     @classmethod
     def init_resumable_rest(cls, request, bucket):
-        name = request.args.get("name", "")
+        name = request.args.get("name", None)
         rest_only = {}
-        if len(request.data) > 0 and request.data.decode("utf-8") != "{}":
-            if name != "":
-                utils.error.invalid("name argument in non-empty payload", None)
+        if len(request.data) > 0:
             data = json.loads(request.data)
             rest_only = cls.__extract_rest_only(data)
             metadata = json_format.ParseDict(data, resources_pb2.Object())
+            if name != None and name != metadata.name:
+                utils.error.invalid(
+                    "Value '%s' in content does not agree with value '%s'."
+                    % (metadata.name, name),
+                    context=None,
+                )
+            if not metadata.name and name:
+                metadata.name = name
         else:
             metadata = resources_pb2.Object()
+            if not metadata.name:
+                utils.error.invalid("No object name", context=None)
             metadata.name = name
         if metadata.content_type == "":
             metadata.content_type = request.headers.get(
                 "x-upload-content-type", "application/octet-stream"
             )
+        if metadata.name == "":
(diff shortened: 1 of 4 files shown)


google/cloud/storage/emulator/gcs/holder.py
```diff
@@ -59,21 +59,31 @@ def init_upload(
...
-        if len(request.data) > 0 and request.data.decode("utf-8") != "{}":
-            if name != "":
-                utils.error.invalid("name argument in non-empty payload", None)
+        if len(request.data) > 0:
             data = json.loads(request.data)
             rest_only = cls.__extract_rest_only(data)
             metadata = json_format.ParseDict(data, resources_pb2.Object())
+            if name != None and name != metadata.name:
+                utils.error.invalid(
+                    "Value '%s' in content does not agree with value '%s'."
```
coryan, line 70: AFAIK this is not an error, the name parameter "Overrides the object metadata's name value, if any. ":

https://cloud.google.com/storage/docs/json_api/v1/objects/insert#parameters

google/cloud/storage/emulator/gcs/holder.py
```diff
@@ -59,21 +59,31 @@ def init_upload(
...
             metadata = resources_pb2.Object()
+            if not metadata.name:
+                utils.error.invalid("No object name", context=None)
             metadata.name = name
         if metadata.content_type == "":
             metadata.content_type = request.headers.get(
                 "x-upload-content-type", "application/octet-stream"
             )
+        if metadata.name == "":
+            utils.error.invalid("No object name", context=None)
```
coryan, line 86: It might be useful to use a different error message, if only to troubleshoot the CI failures.

coryan (changes requested):

google/cloud/storage/emulator/gcs/holder.py
frankyn: That's correct for multipart uploads. When I tested the different cases against the service, it fails in this way with resumable uploads, maybe this is a bug then.

Example 1:
```bash
export BUCKET_NAME="anima-frank"
curl -X POST -H "Authorization: Bearer $(gcloud auth application-default print-access-token)" \
    -H "Content-Type: application/json" \
    "https://storage.googleapis.com/upload/storage/v1/b/${BUCKET_NAME}/o?uploadType=resumable&name=t" \
    -d '{"name": "test"}'

# response
# {
#   "error": {
#     "code": 400,
#     "message": "Value 'test' in content does not agree with value 't'. This can happen when a value set through a parameter is inconsistent with a value set in the request.",
#     "errors": [
#       {
#         "message": "Value 'test' in content does not agree with value 't'. This can happen when a value set through a parameter is inconsistent with a value set in the request.",
#         "domain": "global",
#         "reason": "invalidParameter"
#       }
#     ]
#   }
# }
```

frankyn:
I'll pick this back up when I'm back. Thanks for your patience. Closing PR for now.

PR #6484, opened 2021-05-06
refactor(storage): use streaming RPC wrappers

coryan:
Use the `google::cloud::internal::Streaming{Read,Write}Rpc` wrappers,
they are (allegedly) easier to mock and use. I will be splitting the
`storage::internal::GrpcClient` class even more in a future PR, and this
change makes that easier.

Part of the work for #6309

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6484)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -25,6 +25,8 @@
 #include "google/cloud/internal/format_time_point.h"
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/internal/invoke_result.h"
+#include "google/cloud/internal/streaming_read_rpc.h"
+#include "google/cloud/internal/streaming_write_rpc.h"
 #include "google/cloud/internal/time_utils.h"
 #include "google/cloud/log.h"
 #include "absl/algorithm/container.h"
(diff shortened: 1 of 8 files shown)


google/cloud/storage/internal/grpc_client.cc
```diff
@@ -286,28 +297,40 @@ StatusOr<BucketMetadata> GrpcClient::LockBucketRetentionPolicy(
...
   auto const& contents = request.contents();
+  auto const contents_size = static_cast<std::int64_t>(contents.size());
+  std::int64_t const maximum_buffer_size =
+      google::storage::v1::ServiceConstants::MAX_WRITE_CHUNK_BYTES;
 
   // This loop must run at least once because we need to send at least one
   // Write() call for empty objects.
-  std::size_t offset = 0;
-  do {
+  for (std::int64_t offset = 0, n = 0; offset < contents_size; offset += n) {
```
devjgm, line 325: Since this loop must run at least once, is `contents_size` guaranteed to be > 0?

devjgm (approved):

google/cloud/storage/internal/grpc_client.cc
coryan: Nope, thanks for catching that. Fixed now.

devjgm (approved):

PR #6485, opened 2021-05-06
refactor(storage): split gRPC Client vs. Stub

coryan:
Separate the wrappers around the gRPC code to its own file. That makes
it easier to add decorators in the style of other gRPC-based libraries,
particularly it is easier to add the auth decorator, which should be
used only when needed.

Part of the work for #6309

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6485)


coryan:
Apologies for the monster PR, the ways to split this are uglier (I think).

devjgm (approved):

PR #6488, opened 2021-05-07
feat(storage): support service account impersonation

### LINKED ISSUE #6309: Create unified authorization client abstractions for "impersonate a service account"

coryan:
This weaves the support for service account impersonation through the
gRPC and REST adapters. This change includes an integration test for
`google::cloud::storage` using this new credential type.

Part of the work for #6309, some cleanup afterwards and I can close it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6488)

### DIFF
--- a/google/cloud/internal/credentials.h
+++ b/google/cloud/internal/credentials.h
@@ -84,6 +84,7 @@ class CredentialsVisitor {
   virtual ~CredentialsVisitor() = default;
   virtual void visit(GoogleDefaultCredentialsConfig&) = 0;
   virtual void visit(AccessTokenConfig&) = 0;
+  virtual void visit(ImpersonateServiceAccountConfig&) = 0;
 
   static void dispatch(Credentials& credentials, CredentialsVisitor& visitor);
 };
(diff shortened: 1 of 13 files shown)


scotthart (approved):
> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6488)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

Issue #6502, opened 2021-05-08
[GCB] Make Fedora version configurable


coryan:
## What?

I want to make the Fedora version used by each build configurable, using a flag for `build.sh` or a substitution for the triggers.

## Why?

Because when updating Fedora sometimes we need to keep one or more builds behind.  Right now this is all the Bazel+GCC builds, as BoringSSL does not compile with GCC 11 and Bazel.

## Any alternatives?

We could move some of those builds to a different distro, one of the Ubuntus for example, but that is (probably) more confusing.

We could wait until BoringSSL is ready. But we will have this problem on every update, and updating all the builds in sync will always be problematic.

We could use a more stable distro for our builds, e.g., one of the Ubuntu LTS releases.  Those are typically stable over 5 years.  I have never written this down, but I think Fedora offers better tradeoffs:

- The releases are stable for a year, which means the development tools are never "too old".
- There are six months of overlap between "current" and "latest", which gives us time to update.
- We are *forced* to update after a year, which means we keep up with the current compiler releases.
- It includes a lot of the development tools (and dependencies) we need.

I think the work with take 2 or 3 PRs, so I opened an issue.

devjgm:
Another alternative to consider: Create separate `fedora-33.Dockerfile` and `fedora-34.Dockerfile` (etc.) dockerfiles, and use each one where appropriate.

Pros:
* Consistency w/ the existing model used for other distros, such as `ubuntu-bionic.Dockerfile`, etc.
* Doesn't add a flag to `build.sh` that only applies to certain distros but not others
* Allows us the option to vary the details of each fedora image

Cons:
* If we truly want all the steps of the fedora images to be identical except for the distro version, then having multiple Dockerfiles would cause duplication

coryan:
> Another alternative to consider: Create separate `fedora-33.Dockerfile` and `fedora-34.Dockerfile` (etc.) dockerfiles, and use each one where appropriate.

Yup, that just occurred to me today.  Let me change the PR to do that.

coryan:
I think this is now fixed.

PR #6503, opened 2021-05-08
ci: make Fedora version configurable

coryan:
Change the dockerfiles to make *only* the Fedora version configurable.
The other builds have the version fixed in their name, e.g.,
`ubuntu-bionic.Dockerfile` obviously is intended to be used exclusively
with `ubuntu:bionic` (aka `ubuntu:18.04`).

Change the build.sh script to accept a `--distro-version` flag to set
the `DISTRO_VERSION` build argument in `docker build`. Also changed the
triggers to set `_DISTRO_VERSION` substitution, so this value is picked
up correctly in builds with `-t` flag. The substitution is not used in
the GCB builds yet.

Part of the work for #6502

### DIFF
--- a/ci/cloudbuild/dockerfiles/demo-centos-7.Dockerfile
+++ b/ci/cloudbuild/dockerfiles/demo-centos-7.Dockerfile
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG DISTRO_VERSION=7
-FROM centos:${DISTRO_VERSION}
+FROM centos:7
 ARG NCPU=4
 
 ## [BEGIN packaging.md]
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -18,6 +18,7 @@
 #
 #   Options:
 #     --distro=<name>      The distro name to use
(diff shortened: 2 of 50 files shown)


coryan:
Another way to implement this would be to create a `fedora-33.Dockerfile` and `fedora-34.Dockerfile`... maybe we should do that instead?

coryan:
PTAL

ci/cloudbuild/dockerfiles/demo-centos-7.Dockerfile
```diff
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG DISTRO_VERSION=7
```
devjgm, line 15: Thanks for cleaning up these `DISTRO_VERSION`s. I originally copied docker files over as-is, but planned to clean this up later. Apparently I forgot, and even forgot to file an issue so I don't forget ;-)

devjgm (approved):

devjgm:
FWIW, when I make mass changes to yaml files like this, I often just re-import all of them w/ something like `ls *.yaml | xargs -n1 ../trigger.sh --import`

Issue #6461, opened 2021-05-05
Bigtable AsyncBulkApply does not respect backoff policy


dbolduc:
`AsyncRetryBulkApply::OnFinish` needs a delay before retrying `StartIterationIfNeeded` as determined by the backoff policy.

PR #6510, opened 2021-05-10
fix(bigtable): async bulk apply retry, backoff policy fix 

dbolduc:
fixes: #6461

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6510)

### DIFF
--- a/google/cloud/bigtable/internal/async_bulk_apply.cc
+++ b/google/cloud/bigtable/internal/async_bulk_apply.cc
@@ -92,7 +92,7 @@ void AsyncRetryBulkApply::OnFinish(CompletionQueue cq, Status const& status) {
   auto self = this->shared_from_this();
   cq.MakeRelativeTimer(rpc_backoff_policy_->OnCompletion(status))
       .then([self, cq](TimerFuture result) {
-        if (auto tp = result.get()) {
+        if (result.get()) {
           self->StartIteration(std::move(cq));
         } else {
           self->SetPromise();
--- a/google/cloud/bigtable/internal/async_bulk_apply_test.cc
+++ b/google/cloud/bigtable/internal/async_bulk_apply_test.cc
@@ -22,6 +22,7 @@
 #include "google/cloud/testing_util/status_matchers.h"
 #include "google/cloud/testing_util/validate_metadata.h"
 #include <gmock/gmock.h>
+#include <algorithm>
 #include <iterator>
 
 namespace google {


google/cloud/bigtable/internal/async_bulk_apply_test.cc
```diff
@@ -46,15 +46,42 @@ class AsyncBulkApplyTest : public bigtable::testing::TableTestFixture {
...
+TEST_F(AsyncBulkApplyTest, NoMutations) {
+  bigtable::BulkMutation mut;
+
+  auto bulk_apply_future = internal::AsyncRetryBulkApply::Create(
+      cq_, rpc_retry_policy_->clone(), rpc_backoff_policy_->clone(),
+      *idempotent_mutation_policy_, metadata_update_policy_, client_,
+      "my-app-profile", "my-table", std::move(mut));
+
+  ASSERT_EQ(0U, bulk_apply_future.get().size());
+};
```
coryan, line 82: nit: redundant semi-colon

google/cloud/bigtable/internal/bulk_mutator.cc
```diff
@@ -127,7 +127,7 @@ std::vector<int> BulkMutatorState::OnRead(
   return res;
 }
 
-void BulkMutatorState::OnFinish(google::cloud::Status finish_status) {
+void BulkMutatorState::OnFinish(google::cloud::Status const& finish_status) {
```
coryan, line 130: This change makes it impossible for callers to transfer `finish_status` when they have no more use for it, forcing an additional copy that may be unneeded.

In addition, this change seems unrelated: you could do it in a separate PR.  Future you would appreciate if each PR is about one thing only.

google/cloud/bigtable/internal/async_bulk_apply_test.cc
```diff
@@ -288,6 +285,179 @@ TEST_F(AsyncBulkApplyTest, AsyncBulkApplyFailureRetry) {
...
+
+  auto bulk_apply_future = internal::AsyncRetryBulkApply::Create(
+      cq_, rpc_retry_policy_->clone(), std::move(mock),
+      *idempotent_mutation_policy_, metadata_update_policy_, client_,
+      "my-app-profile", "my-table", std::move(mut));
+
+  SimulateIteration();
+  ASSERT_EQ(1U, cq_impl_->size());
+
+  // cancel the future
```
coryan, line 450: I would say "cancel the pending operation"

google/cloud/bigtable/internal/async_bulk_apply.cc
```diff
@@ -78,17 +71,33 @@ void AsyncRetryBulkApply::StartIterationIfNeeded(CompletionQueue cq) {
...
 void AsyncRetryBulkApply::OnRead(
     google::bigtable::v2::MutateRowsResponse response) {
   state_.OnRead(response);
 }
 
-void AsyncRetryBulkApply::OnFinish(CompletionQueue cq, Status status) {
-  state_.OnFinish(std::move(status));
-  StartIterationIfNeeded(std::move(cq));
+void AsyncRetryBulkApply::OnFinish(CompletionQueue cq, Status const& status) {
+  state_.OnFinish(status);
```
coryan, line 83: I might have written this as:

```cc
void AsyncRetryBulkApply::OnFinish(CompletionQueue cq, Status status) {
  auto const is_retryable = status.ok() || rpc_retry_policy_->OnFailure(status);
  state_.OnFinish(std::move(status));
  if (!state_.HasPendingMutations() || !is_retryable) {
    promise_.set_value(std::move(state_).OnRetryDone());
    return;
  }
... ... ...  
```

Mostly because the copy of `Status` (which has a `std::string`) is more expensive than checking whether it is retryable.

google/cloud/bigtable/internal/async_bulk_apply.cc
```diff
@@ -78,17 +71,33 @@ void AsyncRetryBulkApply::StartIterationIfNeeded(CompletionQueue cq) {
...
+      (!status.ok() && !rpc_retry_policy_->OnFailure(status))) {
+    promise_.set_value(std::move(state_).OnRetryDone());
+    return;
+  }
+
+  auto self = this->shared_from_this();
+  cq.MakeRelativeTimer(rpc_backoff_policy_->OnCompletion(status))
+      .then(
+          [self,
+           cq](future<StatusOr<std::chrono::system_clock::time_point>> result) {
```
coryan, line 94: Consider an alias for `future<StatusOr<std::chrono::system_clock...>>`. I do wish C++14 was more popular and we could write `[self, cq](auto result) {`

google/cloud/bigtable/internal/async_bulk_apply.cc
```diff
@@ -78,17 +71,33 @@ void AsyncRetryBulkApply::StartIterationIfNeeded(CompletionQueue cq) {
...
+            if (auto tp = result.get()) {
+              self->StartIteration(std::move(cq));
+            } else {
+              self->promise_.set_value(std::move(self->state_).OnRetryDone());
```
coryan, line 98: The double `self->` here is odd, consider (but this is your call) a private `OnRetryDone()` function

google/cloud/bigtable/internal/async_bulk_apply_test.cc
```diff
@@ -46,15 +46,42 @@ class AsyncBulkApplyTest : public bigtable::testing::TableTestFixture {
...
+    cq_impl_->SimulateCompletion(true);
+    // state == PROCESSING
+    cq_impl_->SimulateCompletion(true);
+    // state == PROCESSING, 1 read
+    cq_impl_->SimulateCompletion(false);
+    // state == FINISHING
+    cq_impl_->SimulateCompletion(true);
+  }
+
+  void SimulateIterationThatRetries() {
```
coryan, line 61: `SimulateIterationAndRetry()` maybe?  Though in general I think the more "unrolled" tests are the better (you do not want your test to get so complicated that they need test themselves 😬 ), so maybe just not have this function at all.  Your call.

google/cloud/bigtable/internal/async_bulk_apply_test.cc
```diff
@@ -288,6 +285,179 @@ TEST_F(AsyncBulkApplyTest, AsyncBulkApplyFailureRetry) {
...
+TEST_F(AsyncBulkApplyTest, TooManyFailures) {
+  bigtable::BulkMutation mut{
+      bigtable::SingleRowMutation("foo2",
+                                  {bigtable::SetCell("f", "c", 0_ms, "v2")}),
+      bigtable::SingleRowMutation("foo3",
+                                  {bigtable::SetCell("f", "c", 0_ms, "v3")}),
+  };
+
+  // We give up on the 5th error.
+  const int limited_error_count = 4;
```
coryan, line 297: Consider instead:

```
auto constexpr kErrorCount = 4;
```

Also: is there a benefit from testing with 4 errors vs. 1 or 2 or 3? Pick the smallest number that would do the trick.

google/cloud/bigtable/internal/async_bulk_apply_test.cc
```diff
@@ -288,6 +285,179 @@ TEST_F(AsyncBulkApplyTest, AsyncBulkApplyFailureRetry) {
...
+
+  for (int retry = 0; retry < limited_error_count; ++retry) {
+    SimulateIterationThatRetries();
+    ASSERT_EQ(1U, cq_impl_->size());
+  }
+
+  SimulateIteration();
+
+  auto failures = bulk_apply_future.get();
+  EXPECT_EQ(2U, failures.size());
```
coryan, line 328: Since these test were written we have gained a bunch of useful helpers, nowadays we would say:

```cc
EXPECT_THAT(failures, ElementsAre(StatusIs(StatusCode::kUnavailable), StatusIs(StatusCode::kUnavailable)));
```

On failure the expression I propose would be a lot more informative.

coryan (changes requested):
A bunch of nit picks, generally looks really good.

google/cloud/bigtable/internal/async_bulk_apply_test.cc
dbolduc: Done.

google/cloud/bigtable/internal/async_bulk_apply_test.cc
dbolduc: Done.

google/cloud/bigtable/internal/async_bulk_apply.cc
dbolduc: Done. but with `OnFinish(status)` because we need the value later in the method

google/cloud/bigtable/internal/async_bulk_apply.cc
dbolduc: Done.

google/cloud/bigtable/internal/async_bulk_apply_test.cc
dbolduc: Ack. My thinking for `SimulateIteration` was that if we change the exact way `AsyncBulkApply` uses the `CompletionQueue` (i.e. add/remove an operation behind the scenes), we would only need to update one place in the test.

I unrolled `SimulateIterationThatRetries` into:
 ```c++
 SimulateIteration();
// Start backoff policy timer before a retry
 cq_impl_->SimulateCompletion(ok)
 ```

google/cloud/bigtable/internal/async_bulk_apply_test.cc
dbolduc: Done.

google/cloud/bigtable/internal/async_bulk_apply_test.cc
dbolduc: Done.

google/cloud/bigtable/internal/async_bulk_apply_test.cc
```diff
@@ -46,15 +48,43 @@ class AsyncBulkApplyTest : public bigtable::testing::TableTestFixture {
...
   std::shared_ptr<RPCRetryPolicy const> rpc_retry_policy_;
   std::shared_ptr<RPCBackoffPolicy const> rpc_backoff_policy_;
   std::shared_ptr<IdempotentMutationPolicy> idempotent_mutation_policy_;
   MetadataUpdatePolicy metadata_update_policy_;
 };
 
-TEST_F(AsyncBulkApplyTest, AsyncBulkApplySuccess) {
+std::vector<Status> StatusOnly(std::vector<FailedMutation> const& failures) {
+  std::vector<Status> v;
+  std::transform(failures.begin(), failures.end(), std::back_inserter(v),
```
devjgm, line 71: `#include <algorithm>`

devjgm (commented):
Wow, this looks great! And a rather meaty first(-ish) PR. Great job.

Overall, this LGTM, but I'm going to defer to @coryan for actual approval since he knows this code way better than I do.

coryan:
I concur with @devjgm otherwise LGTM.

PR #6515, opened 2021-05-10
ci: rename publish docs head version to HEAD

devjgm:
Uses a less offensive term, and a clearer term since the docs are indeed
for the newest unreleased version.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6515)


devjgm:
Example: https://staging.googleapis.dev/cpp/google-cloud-pubsub/HEAD/index.html

devbww (approved):

PR #6523, opened 2021-05-11
cleanup(spanner): execute create-instance sample during auto run

devbww:
Actually execute the "spanner_create_instance" sample during an
"GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES=yes" integration test, rather
than some similar looking code.

This also opens a pathway for upcoming create-instance additions.

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -102,61 +102,6 @@ void CreateInstanceCommand(std::vector<std::string> const& argv) {
   CreateInstance(std::move(client), argv[0], argv[1], argv[2], region);
 }
 
-std::string PickInstanceLocation(
-    google::cloud::spanner::InstanceAdminClient client,
-    std::string const& project_id) {
-  std::mt19937_64 generator(std::random_device{}());
-  std::string ret{};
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -3287,6 +3231,43 @@ int RunOneCommand(std::vector<std::string> argv) {
...
+  std::string const included_prefix = "regional-us-";
+  // Exclude the US configs where we keep most of our test instances.
+  std::vector<std::string> const excluded{"regional-us-central1",
+                                          "regional-us-east1"};
+
+  std::string const config_prefix =
+      "projects/" + project_id + "/instanceConfigs/";
+
+  // Use a reservoir sampler of size k==1 to pick a region at random.
+  std::mt19937_64 generator(std::random_device{}());
```
coryan, line 3246: nit: this is only called from a function that already has a generator, why not pass it in?

coryan (approved):

google/cloud/spanner/samples/samples.cc
devbww: Done.  Thanks.

PR #6539, opened 2021-05-12
cleanup(spanner): make samples PickConfig() less restrictive

devbww:
If PickConfig() cannot find an instance config that satisfies its
filters, have it return the first instance config (if one exists).
This makes it work naturally when run against the Spanner emulator,
or any other test service where the configs are not so predictable.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6539)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -3248,11 +3248,13 @@ std::string PickConfig(google::cloud::spanner::InstanceAdminClient client,
   std::cout << "\n" << __func__ << ":\n";
 
   std::string ret;
+  std::string first;
   int i = 0;
   for (auto const& instance_config : client.ListInstanceConfigs(project_id)) {
     if (!instance_config) break;
     if (instance_config->name().rfind(config_prefix, 0) != 0) continue;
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -3248,11 +3248,13 @@ std::string PickConfig(google::cloud::spanner::InstanceAdminClient client,
   std::cout << "\n" << __func__ << ":\n";
 
   std::string ret;
+  std::string first;
   int i = 0;
   for (auto const& instance_config : client.ListInstanceConfigs(project_id)) {
     if (!instance_config) break;
     if (instance_config->name().rfind(config_prefix, 0) != 0) continue;
     auto const config = instance_config->name().substr(config_prefix.size());
+    if (first.empty()) first = config;
```
coryan, line 3257: I am not sure you need another `first` variable if you change this to `if (ret.empty()) ret = config;` ?

coryan (approved):

google/cloud/spanner/samples/samples.cc
devbww: Yes, but I thought that was less clear.

PR #6547, opened 2021-05-12
ci: simplify generator integration tests

devjgm:
1. This PR removes the `generator/ci/` scripts because there's no
   emulator for the generator, so there's no need to have the
   "run_integration_tests_emulator*" scripts. It's simpiler just to
   inline the execution of the generator integration tests in
   `lib/integration.sh`.
2. Fixed the generator integration tests to run with CMake. Previously
   they were passing, but they were not being run because
   GOOGLE_CLOUD_CPP_GENERATOR_RUN_INTEGRATION_TESTS was not set. This PR
   enables these tests and fixes a few places where the needed paths
   were wrong.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6547)


devjgm:
/cc: @scotthart FYI. Should be no change for you, other than if you need to edit the generator's integration tests, look in `ci/cloudbuild/builds/lib/integration.sh`.

coryan (approved):

PR #6549, opened 2021-05-12
chore: upgrade vcpkg to latest release

coryan:
AFAICT this will use the same version for all builds.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6549)

### DIFF
--- a/ci/cloudbuild/builds/quickstart-cmake.sh
+++ b/ci/cloudbuild/builds/quickstart-cmake.sh
@@ -26,11 +26,10 @@ export CXX=g++
 
 io::log_h2 "Installing vcpkg"
 vcpkg_dir="${HOME}/vcpkg-quickstart"
-vcpkg_sha="105456798402aa5f494ffeb3b19dd0d870656d39"
 vcpkg_bin="${vcpkg_dir}/vcpkg"
 mkdir -p "${vcpkg_dir}"
-io::log "Downloading vcpkg@${vcpkg_sha} into ${vcpkg_dir}..."
-curl -sSL "https://github.com/microsoft/vcpkg/archive/${vcpkg_sha}.tar.gz" |
+io::log "Downloading vcpkg into ${vcpkg_dir}..."
+curl -sSL "https://github.com/microsoft/vcpkg/archive/2021.04.30.tar.gz" |
   tar -C "${vcpkg_dir}" --strip-components=1 -zxf -
 env CC="ccache ${CC}" CXX="ccache ${CXX}" "${vcpkg_dir}"/bootstrap-vcpkg.sh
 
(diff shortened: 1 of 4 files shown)


ci/cloudbuild/builds/quickstart-cmake.sh
```diff
@@ -26,11 +26,10 @@ export CXX=g++
 
 io::log_h2 "Installing vcpkg"
 vcpkg_dir="${HOME}/vcpkg-quickstart"
-vcpkg_sha="105456798402aa5f494ffeb3b19dd0d870656d39"
 vcpkg_bin="${vcpkg_dir}/vcpkg"
 mkdir -p "${vcpkg_dir}"
-io::log "Downloading vcpkg@${vcpkg_sha} into ${vcpkg_dir}..."
-curl -sSL "https://github.com/microsoft/vcpkg/archive/${vcpkg_sha}.tar.gz" |
+io::log "Downloading vcpkg into ${vcpkg_dir}..."
+curl -sSL "https://github.com/microsoft/vcpkg/archive/2021.04.30.tar.gz" |
```
devjgm, line 32: Should we do the same thing in 

https://github.com/googleapis/google-cloud-cpp/blob/aaf4ff98498b4cadb253d093901624960b876e38/ci/cloudbuild/builds/cmake-oldest-deps.sh#L29

Or is that case intentionally different?

coryan: Good question... it is intentionally different, because we are doing "history" in `cmake-oldest-deps`. Maybe my intentions are bad ... meaning I do not recall testing if a tarball would work in this case.

coryan: This is what I got when using a tarball:

```
/workspace/ci/etc/oldest-deps
Error: Couldn't find explicitly specified baseline `"6d9ed568117dd958c543b3ab8d3ed692965cac34"` in the baseline file, and there was no baseline at that commit or the commit didn't exist.
Error: while checking out baseline '6d9ed568117dd958c543b3ab8d3ed692965cac34:versions/baseline.json':
fatal: not a git repository: '/workspace/cmake-out/vcpkg/.git'

This may be fixed by updating vcpkg to the latest master via `git pull`.
Failed to determine the current commit:
fatal: not a git repository: '/workspace/cmake-out/vcpkg/.git'
```

It seems that vcpkg needs the git history to implement "versions".  I am going to add a comment, but otherwise leave as in.

devjgm: SGTM. Thanks.

devjgm (approved):

Issue #6516, opened 2021-05-10
ReadObject with empty or small range results in error


shaileshkm:
**Does this issue affect the google-cloud-cpp project?**
If the problem is with the Google Cloud service exposed by the `google-cloud-cpp` libraries instead of the client libraries themselves, you may consider opening a support request instead. The `google-cloud-cpp` developers cannot help you troubleshoot problems with the service itself.

Doing read_stream.status().message.data() returns
0x7fffeef27558 <std::string::_Rep::_S_empty_rep_storage+24> ""
I suspect this is SDK issue.

**What component of `google-cloud-cpp` is this related to?**
For example, is this related to bigtable (i.e., something in `google/cloud/bigtable`), or GCS (i.e., something in `google/cloud/storage`)?

Storage

**Describe the bug**

If I run ReadObject on a zero sized object, error is set on stream with empty buffer in status.

**To Reproduce**
1. Create empty/0 byte object.
2. `gcs::ObjectReadStream read_stream = client.ReadObject(_bucket_name.data(), _object_name.data(), gcs::ReadRange(start, start+readsize));
`
where readsize is 0.
3. Check read_stream && read_stream.IsOpen()
4. Obtain read_stream.status().message().data()

**Expected behavior**
My expectations are:
- If I'm able to create a 0 sized object in store, I should be able to read it. SDK minimally should check if the object is accessible.
- Even if it decides to set an error, it should be legible message.

**Operating system:**
If you are using a Linux distribution please include the name and version of the distribution too.
SUSE Linux Enterprise Server 12 SP4

**What compiler and version are you using?**
Please include the output of `g++ -v` or `clang++ -v` or the equivalent command-line flag.
cc version 9.2.1 20191018 

**What version of `google-cloud-cpp` are you using?**
Please include the output from `git rev-parse HEAD` if you are compiling from source, or the version number from the applicable `google/cloud/*/version.h` file.
1.24

**Additional context**

coryan:
Thanks for the detailed bug report.  

> If I'm able to create a 0 sized object in store, I should be able to read it. SDK minimally should check if the object is accessible.
> Even if it decides to set an error, it should be legible message.

I think those things are true.  Unfortunately there is no way to express "read 0 bytes" in the JSON protocol, none that comes to mind immediately. You can read 1 byte, you can even read all the bytes (and it returns 0 bytes), but asking for exactly 0 bytes cannot be expressed.

[...]

PR #6551, opened 2021-05-12
test(storage): more detailed tests for ReadRange

coryan:
Add some tests for `gcs::ReadRange()`, `gcs::ReadFromOffset()`, and
`gcs::ReadLast()`.

Part of the work for #6516

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6551)

### DIFF
--- a/google/cloud/storage/tests/object_read_range_integration_test.cc
+++ b/google/cloud/storage/tests/object_read_range_integration_test.cc
@@ -106,7 +106,6 @@ TEST_F(ObjectReadRangeIntegrationTest, ReadFromOffset) {
   } cases[] = {
       {0, contents.substr(0)},
       {4, contents.substr(4)},
-      {0, contents},
       {size - 1, contents.substr(size - 1)},
   };
 
--- a/google/cloud/storage/tests/object_read_range_integration_test.cc
+++ b/google/cloud/storage/tests/object_read_range_integration_test.cc
@@ -1,0 +1,165 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


google/cloud/storage/tests/object_read_range_integration_test.cc
```diff
@@ -0,0 +1,165 @@
...
+      {0, size, contents},
+  };
+
+  for (auto const& test : cases) {
+    SCOPED_TRACE("Testing range [" + std::to_string(test.begin) + "," +
+                 std::to_string(test.end) + ")");
+    auto reader = client->ReadObject(bucket_name(), object_name,
+                                     ReadRange(test.begin, test.end));
+    auto actual = std::string{std::istreambuf_iterator<char>(reader), {}};
+    EXPECT_THAT(reader.status(), IsOk());
```
devbww, line 81: Should you do this before constructing `actual`?

google/cloud/storage/tests/object_read_range_integration_test.cc
```diff
@@ -0,0 +1,165 @@
...
+  auto size = static_cast<std::int64_t>(insert->size());
+
+  // Read several small portions of the object, expecting specific results.
+  struct Test {
+    std::int64_t offset;
+    std::string expected;
+  } cases[] = {
+      {0, contents.substr(0)},
+      {4, contents.substr(4)},
+      {0, contents},
```
devbww, line 109: Why is this different from the first case?

devbww (approved):

google/cloud/storage/tests/object_read_range_integration_test.cc
coryan: I am like 99.4% certain that it is better to check `.status()` after constructing `actual`:

* For small objects like these it does not matter, a failed download fails immediately, and `actual` is empty because the iterator stops immediately,
* For larger objects the download may start successfully (at which point `.status().code()` is `kOkay`) and fail by the end, and once it fails the `.status()` never changes.

google/cloud/storage/tests/object_read_range_integration_test.cc
coryan: Doh, removed.

google/cloud/storage/tests/object_read_range_integration_test.cc
devbww: Okie dokie.  (2) sounds good, and if the `string` construction in (1) works on a failed reader then that sounds good too.  Thanks.

Issue #6543, opened 2021-05-12
Enable warnings-as-errors in Doxygen


coryan:
We disabled "warnings as errors" in Doxygen in the Fedora:32 to Fedora:33 transition because there were too many false positives.  It seems like those are under control again, but the "true positives" are non-zero and need to be manually fixed:

https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/6536/f8e9fcd1010cdd84e0b20a2a8b664d94bcfab5ad/fedora-34-publish-docs/log-3b70a6c6-2124-46bc-be74-93b2f07aa8d9.txt

```
Step #3: /workspace/google/cloud/stream_range.h:101: warning: explicit link request to 'iterator' could not be resolved
...

Step #3: T+35.970s [3/9] Generate logging HTML documentation
Step #3: warning: source '/workspace/google/cloud/logging/samples' is not a readable file or directory... skipping.
Step #3: warning: source '/workspace/google/cloud/logging/quickstart' is not a readable file or directory... skipping.
...
Step #3: /workspace/google/cloud/logging/logging_service_v2_client.h:199: warning: explicit link request to 'google::logging::v2::DeleteLogRequest' could not be resolved
Step #3: /workspace/google/cloud/logging/logging_service_v2_client.h:227: warning: explicit link request to 'google::logging::v2::ListLogEntriesRequest' could not be resolved
Step #3: /workspace/google/cloud/logging/logging_service_v2_client.h:247: warning: explicit link request to 'google::logging::v2::ListLogsRequest' could not be resolved
Step #3: /workspace/google/cloud/logging/logging_service_v2_client.h:236: warning: explicit link request to 'google::logging::v2::ListMonitoredResourceDescriptorsRequest' could not be resolved
Step #3: /workspace/google/cloud/logging/logging_service_v2_client.h:213: warning: explicit link request to 'google::logging::v2::WriteLogEntriesRequest' could not be resolved
Step #3: /workspace/google/cloud/logging/logging_service_v2_client.h:215: warning: explicit link request to 'google::logging::v2::WriteLogEntriesResponse' could not be resolved
Step #3: /workspace/google/cloud/logging/logging_service_v2_client.h:136: warning: explicit link request to 'google::logging::v2::WriteLogEntriesResponse' could not be resolved
...
Step #3: /workspace/google/cloud/spanner/read_partition.cc:23: warning: no matching class member found for 
[...]

PR #6552, opened 2021-05-12
ci(doc): fix some doxygen errors

devjgm:
Related to: #6543

This fixes two sets of doxygen errors:

1. It fixes the `fatal error: 'stddef.h'` errors by setting the clang
   resource-dir path.
2. After these errors, we got a bunch of errors about unable to find
   includes like `google/proto/status.pb.h`. These are fixed by
   installing our library first, then telling doxygen where to find the
   headers we installed.

In my local tests, the only remaining errors seem to be legit warnings
that we likely need to fix (in subsequent PRs).

**NOTE** The doxygen build is now slower. It was ~5 minutes. Now it's ~10 minutes. 

Example docs: https://staging.googleapis.dev/cpp/google-cloud-storage/HEAD/index.html

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6552)

### DIFF
--- a/ci/cloudbuild/builds/publish-docs.sh
+++ b/ci/cloudbuild/builds/publish-docs.sh
@@ -19,10 +19,16 @@ set -eu
 source "$(dirname "$0")/../../lib/init.sh"
 source module ci/cloudbuild/builds/lib/cmake.sh
 
+export CC=clang
+export CXX=clang++
+
+install_dir="$(mktemp -d)"
 version=""
 doc_args=(
   "-DCMAKE_BUILD_TYPE=Debug"
   "-DGOOGLE_CLOUD_CPP_GEN_DOCS_FOR_GOOGLEAPIS_DEV=ON"
+  "-DCMAKE_INSTALL_PREFIX=${install_dir}"
+  "-DDOXYGEN_CLANG_OPTIONS=-resource-dir=$(clang -print-resource-dir)"
 )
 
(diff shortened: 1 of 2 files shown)


ci/cloudbuild/builds/publish-docs.sh
```diff
@@ -50,11 +56,13 @@ if [[ "${TRIGGER_TYPE}" == "ci" ]]; then
...
 fi
 
-export CC=clang
-export CXX=clang++
 cmake -GNinja "${doc_args[@]}" -S . -B cmake-out
+cmake --build cmake-out --target install
 cmake --build cmake-out --target doxygen-docs
 
+# Cleans up the installed artifacts, which are no longer needed.
+test -d "${install_dir}" && rm -rf "${install_dir}"
```
devbww, line 64: The `test -d` looks unnecessary.

devbww (approved):

ci/cloudbuild/builds/publish-docs.sh
devjgm: It was. It was an attempt to make the `rm -rf` a bit safer, like in case `${install_dir}` was empty. But in that case, there's nothing to guard against. So, removed.

devjgm:
I'm going to close this PR in favor of https://github.com/googleapis/google-cloud-cpp/pull/6560, which is similar, but I think is a bit cleaner.

PR #6553, opened 2021-05-13
feature(spanner): low-cost instances

devbww:
Support the provisioning of Spanner instances using `processing_units`
as an alternative to `node_count`. 1 node == 1,000 processing units.

Note: This commit is for a preview release. DO NOT MERGE AS IS.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6553)

### DIFF
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -87,7 +87,7 @@ def google_cloud_cpp_deps():
             # committed to the main branch.
             patch_tool = "patch",
             patch_args = ["-p1"],
-            patches = [],
+            patches = ["googleapis.patch"],
         )
 
     # Load protobuf.
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -86,19 +86,56 @@ void CreateInstance(google::cloud::spanner::InstanceAdminClient client,
                   .Build())
           .get();
   if (!instance) throw std::runtime_error(instance.status().message());
-  std::cout << "Created instance [" << in << "]\n";
+  std::cout << "Created instance [" << in << "]:\n" << instance->DebugString();
 }
 //! [END spanner_create_instance] [create-instance]
 
+// [START spanner_create_instance_with_processing_units]
+void CreateInstanceWithProcessingUnits(
+    google::cloud::spanner::InstanceAdminClient client,
+    std::string const& project_id, std::string const& instance_id,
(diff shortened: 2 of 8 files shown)


bazel/google_cloud_cpp_deps.bzl
```diff
@@ -87,7 +87,7 @@ def google_cloud_cpp_deps():
             # committed to the main branch.
             patch_tool = "patch",
             patch_args = ["-p1"],
-            patches = [],
+            patches = ["googleapis.patch"],
```
coryan, line 90: Maybe that could be `@com_github_googleapis_google_cloud_cpp//external:googleapis.patch`, dunno, just guessing.

google/cloud/spanner/samples/samples.cc
```diff
@@ -3414,6 +3451,18 @@ void RunAllSlowInstanceTests(
...
   std::cout << "\nRunning delete-instance sample" << std::endl;
   DeleteInstance(instance_admin_client, project_id, crud_instance_id);
+
+  if (!emulator) {
+    std::cout
+        << "\nRunning spanner_create_instance_with_processing_units sample"
+        << std::endl;
+    CreateInstanceWithProcessingUnits(
+        instance_admin_client, project_id, crud_instance_id,
+        "Test Low-Cost Instance", "regional-us-central1");
```
coryan, line 3461: I assume the hard-coded config is necessary.  Consider a TODO to pick a location at random, or this will create contention across the builds.

coryan (commented):
A few things to consider, but LGTM

bazel/google_cloud_cpp_deps.bzl
devbww: I'm not sure either, but given that `googleapis.patch` will never be merged my guess is that wouldn't work.

google/cloud/spanner/samples/samples.cc
devbww: Yes ... my understanding is that this will currently only work in that region.  We already have the randomly chosen `config` to change this to before merging, although the "TODO" for that is currently only in my head.  I'll make a pending edit to do that, but will deliver the existing tested version first.  Thanks.

devjgm:
@devbww is there anything else you need or are waiting for on this PR?

devbww:
> @devbww is there anything else you need or are waiting for on this PR?

No, nothing to do for now.  I can close.

The next step will be when the proto is released and we update our googleapis version.  Then this (or a new variant) will come along that we'll be able to merge for the actual release.

PR #6558, opened 2021-05-13
refactor: move gRPC credential types example

coryan:
This example shows how to use the IAM library to get different
credential types and then use those with a client library (in this case
Spanner). It did not belong in the "iam/" directory, it is an example
showing how to use two libraries at the same time.

I am planning to add even more examples of errr... uncommon gRPC credentials here.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6558)


coryan:
Ping.

devbww (approved):

PR #6560, opened 2021-05-13
ci(docs): fix doxygen warnings from clang

devjgm:
Related to: #6543

This fixes doxygen + clang being unable find certain resources like
`stddef.h` by telling clang where its `-resource-dir` is. After fixing
that error, there were a bunch of missing includes of proto/grpc
headers. Those were fixed by making the `doxygen-docs` CMake target
depend on `google-cloud-cpp-protos` so that those would be generated
before trying to generate the docs. Then some include path tweaks to
make things work.

I also enabled HTML timestamping so that we can see when new docs get
published.

This change supersedes
https://github.com/googleapis/google-cloud-cpp/pull/6552, which I'm
goign to close.


coryan (approved):

devjgm:
Step #3: GCB: https://console.cloud.google.com/cloud-build/builds;region=global/49c54674-4cbb-4696-a04d-1d086968c844;tab=detail?project=cloud-cpp-testing-resources
Step #3: Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/6560/3d78388851a5c4db4a14cda055ab2fda51fd8c24/fedora-34-publish-docs/log-49c54674-4cbb-4696-a04d-1d086968c844.txt

There are still some includes that aren't found, but much fewer. I don't really understand why these aren't found, but this is probably no big deal or an issue for a future PR.

PR #6565, opened 2021-05-14
cleanup: prevent behavior change from a gRPC change

jacobsa:
This class currently encourages an ownership hack that works around a gRPC API
issue that is [being fixed](https://github.com/grpc/proposal/pull/238). At that
point the tests that use this class will start failing, because they manage it
using **two different `std::unique_ptr` objects**.

Prevent them failing by overriding the upcoming `Destroy` method to do nothing.
This preserves the existing behavior. Because I haven't used the `override`
keyword, this works before and after the gRPC change, decoupling this issue
from the import of the gRPC change.

Longer term it would be better to make the use of this class not require the
ownership hack at all, but that is a larger project. In the meantime this
change unblocks fixing the need for such hacks for other subclasses that don't
want to further promulgate the issue.

@coryan @devbww

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6565)

### DIFF
--- a/google/cloud/bigtable/testing/mock_response_reader.h
+++ b/google/cloud/bigtable/testing/mock_response_reader.h
@@ -79,8 +79,10 @@ class MockResponseReader : public grpc::ClientReaderInterface<Response> {
  *
  * Note that using this mock often requires special memory management. The
  * google mock library requires all mocks to be destroyed. In contrast, grpc
- * specializes `std::unique_ptr<>` to *not* delete objects of type
- * `grpc::ClientAsyncResponseReaderInterface<T>`:
+ * previously specialized `std::unique_ptr<>` to *not* delete objects of type
+ * `grpc::ClientAsyncResponseReaderInterface<T>` (and our override of the
+ * Destroy method below preserves that behavior even after
+ * https://github.com/grpc/proposal/pull/238):
  *
  *
  *     https://github.com/grpc/grpc/blob/608188c680961b8506847c135b5170b41a9081e8/include/grpcpp/impl/codegen/async_unary_call.h#L305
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/testing/mock_response_reader.h
```diff
@@ -111,6 +113,15 @@ class MockAsyncResponseReader
   MOCK_METHOD(void, StartCall, (), (override));
   MOCK_METHOD(void, ReadInitialMetadata, (void*), (override));
   MOCK_METHOD(void, Finish, (Response*, grpc::Status*, void*), (override));
+
+  // Preserve the behavior from before https://github.com/grpc/proposal/pull/238
+  // of not destroying the object when
+  // std::unique_ptr<grpc::ClientAsyncResponseReaderInterface> goes out of
+  // scope.
+  //
+  // TODO(jacobsa): mark this with the override keyword once the method exists
```
coryan, line 122: nit: we do not create `TODO(person)`, I created #6566 for your, please change this to `// TODO(#6566)`

coryan (commented):
LGTM, waiting for the builds to approve and merge.

coryan:
The publicly accessible build logs will be up in a few minutes, but the gist of the `checkers-pr` failure is:

```
/workspace/google/cloud/bigtable/testing/mock_response_reader.h:122:11 - Unknown word (jacobsa)
```

Which I already asked you to change.

google/cloud/bigtable/testing/mock_response_reader.h
jacobsa: I was wondering about that. Done, thanks.

jacobsa:
@coryan: Fixed the TODO. Do I need to do something to make the checks run again? Sorry, unfamiliar with the process.

jacobsa:
@coryan: sorry, I think you need to do it once more after I fixed the commit message lint error.

coryan:
Unfortunately you look like an external contributor from Google Cloud Build's point of view, one of us needs to trigger them.

the conventionalcommits.org check is annoying (but we can ignore it), it is complaining about your **commit** title.  I have found that a second commit makes it not care and just use the PR title (as it should). Leave it be, I can merge with the failure outstanding.

coryan:
Ooops, I forgot about kicking off the Kokoro builds too.

coryan (approved):

coryan:
Thanks for the PR!

jacobsa:
Thanks!

PR #6571, opened 2021-05-14
fix: GOOGLE_CLOUD_CPP_ENABLE_EXAMPLES depends condition

devjgm:
`GOOGLE_CLOUD_CPP_ENABLE_EXAMPLES` is dependent on
`GOOGLE_CLOUD_CPP_ENABLE_CXX_EXCEPTIONS`, but that was unset prior to
its first use, so `GOOGLE_CLOUD_CPP_ENABLE_EXAMPLES` was never getting
enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6571)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,6 +227,10 @@ endif ()
 # also includes the BUILD_TESTING option, which is on by default.
 include(CTest)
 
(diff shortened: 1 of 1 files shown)


coryan (approved):
Ooops, thanks!

PR #6573, opened 2021-05-14
fix: all remaining doxygen errors

devjgm:
Part of: #6543

This fixes all the remaining doxygen errors by ensuring that the needed protos are all generated before doxygen runs.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6573)


coryan (approved):
Thanks!

PR #6574, opened 2021-05-14
fix(storage): address behavior bugs

frankyn:
Round two of #6477

cc: @tritone 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6574)

### DIFF
--- a/google/cloud/storage/emulator/gcs/holder.py
+++ b/google/cloud/storage/emulator/gcs/holder.py
@@ -65,7 +65,11 @@ class DataHolder(types.SimpleNamespace):
         if len(request.data) > 0:
             data = json.loads(request.data)
             data_name = data.get("name", None)
-            if query_name is not None and data_name is not None and query_name != data_name:
+            if (
+                query_name is not None
+                and data_name is not None
+                and query_name != data_name
+            ):
                 utils.error.invalid(
                     "Value '%s' in content does not agree with value '%s'."
                     % (data_name, query_name),
--- a/google/cloud/storage/emulator/gcs/holder.py
+++ b/google/cloud/storage/emulator/gcs/holder.py
@@ -59,31 +59,28 @@ class DataHolder(types.SimpleNamespace):
 
     @classmethod
     def init_resumable_rest(cls, request, bucket):
-        name = request.args.get("name", None)
+        query_name = request.args.get("name", None)
         rest_only = {}
+        metadata = resources_pb2.Object()
         if len(request.data) > 0:
             data = json.loads(request.data)
-            rest_only = cls.__extract_rest_only(data)
-            metadata = json_format.ParseDict(data, resources_pb2.Object())
-            if name != None and name != metadata.name:
+            data_name = data.get("name", None)
+            if query_name is not None and data_name is not None and query_name != data_name:
                 utils.error.invalid(
                     "Value '%s' in content does not agree with value '%s'."
-                    % (metadata.name, name),
+                    % (data_name, query_name),
                     context=None,
                 )
-            if not metadata.name and name:
-                metadata.name = name
-        else:
-            metadata = resources_pb2.Object()
-            if not name:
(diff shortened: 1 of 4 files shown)


google/cloud/storage/emulator/gcs/holder.py
```diff
@@ -59,21 +59,31 @@ def init_upload(
...
+        name = request.args.get("name", None)
         rest_only = {}
-        if len(request.data) > 0 and request.data.decode("utf-8") != "{}":
-            if name != "":
-                utils.error.invalid("name argument in non-empty payload", None)
+        if len(request.data) > 0:
             data = json.loads(request.data)
             rest_only = cls.__extract_rest_only(data)
             metadata = json_format.ParseDict(data, resources_pb2.Object())
+            if name != None and name != metadata.name:
```
tritone, line 68: Can you add a comment to explain why the logic here is needed?

Also `name is not None` in python usually. :)

google/cloud/storage/emulator/gcs/holder.py
```diff
@@ -59,21 +59,31 @@ def init_upload(
...
+                utils.error.invalid(
+                    "Value '%s' in content does not agree with value '%s'."
+                    % (metadata.name, name),
+                    context=None,
+                )
+            if not metadata.name and name:
```
tritone, line 74: Can this be moved outside the `if` to dedup with L80 in the else?

tritone (commented):
thanks Frank!

google/cloud/storage/emulator/gcs/holder.py
coryan, line 68: Question: when should one use `is not None` vs. `!= None` ?

google/cloud/storage/emulator/gcs/holder.py
```diff
@@ -59,21 +59,31 @@ def init_upload(
...
+                metadata.name = name
         else:
             metadata = resources_pb2.Object()
+            if not name:
```
coryan, line 78: I think you are going to do this check in line 85 anyway :shrug:

google/cloud/storage/emulator/tests/test_gcs.py
```diff
@@ -415,6 +418,160 @@ def test_update_and_patch(self):
...
 
 class TestHolder(unittest.TestCase):
+    def test_init_resumable_rest_incorrect_usage(self):
+        bucket_metadata = json.dumps({"name": "bucket-test"})
+        environ = create_environ(
+            base_url="http://localhost:8080",
+            content_length=len(bucket_metadata),
+            data=bucket_metadata,
+            content_type="application/json",
+            method="GET",
```
coryan, line 428: Question: these are all `GET` requests, but you should be testing stuff for `POST`?

coryan (commented):
The code looks good, and it seems to pass the tests.  I do have a few questions that maybe should be answered, even if the answer is "Carlos, you don't understand it works like blah blah".

google/cloud/storage/emulator/tests/test_gcs.py
frankyn: Whoops, method is ignored in tests because I only need the request object to pass to the underlying `init_resumable_rest()` method call. Addressing issue. Ty

google/cloud/storage/emulator/gcs/holder.py
frankyn: `is not` - compares object addresses
`!=` - compares object values `__eq__(self, obj)`  or literal values

Python uses [String Interning](https://en.wikipedia.org/wiki/String_interning), but it's an implementation detail IIUC so I decided to use `!=` to compare string literal values and `is not` for `None`.

I honestly took this for granted, and found a helpful article that walked me through it: https://realpython.com/python-is-identity-vs-equality/

coryan (approved):

PR #6583, opened 2021-05-16
feat: an example using identity tokens with gRPC

### LINKED ISSUE #6482: Verify OIDC for gRPC works with Cloud Run
Create a Cloud Run service that receives authenticated gRPC requests.

Use the unified credentials to generate a (Google) OIDC token. Send the request, expect a successful response.

Related to #2786

coryan:
Application developers may want to deploy gRPC services to Cloud Run, to
authenticate with these services one must use `GenerateIdentityToken()`
in the IAM credentials service, and pass the resulting token to the gRPC
library. This example uses a small "Hello World"-type gRPC service to
demonstrate this use-case.

Fixes #6482

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6583)

### DIFF
--- a/google/cloud/examples/BUILD
+++ b/google/cloud/examples/BUILD
@@ -16,6 +16,12 @@ package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])  # Apache 2.0
 
+load(
+    "@com_google_googleapis_imports//:imports.bzl",
+    "cc_grpc_library",
+    "cc_proto_library",
+)
+
 # This library defines main(). It's declared as a library so it can be part of
 # a cc_test and cc_binary target.
 cc_library(
@@ -41,15 +47,49 @@ cc_test(
     deps = [":gcs2cbt_lib"],
 )
 
+proto_library(
+    name = "hello_world_protos",
+    srcs = [
+        "hello_world_grpc/hello_world.proto",
+    ],
+    visibility = [
+        "//:__pkg__",
+    ],
+    deps = [],
+)
+
+cc_proto_library(
+    name = "hello_world_cc_proto",
+    visibility = [
+        "//:__pkg__",
+    ],
+    deps = [":hello_world_protos"],
+)
+
+cc_grpc_library(
+    name = "hello_world_cc_grpc",
+    srcs = [":hello_world_protos"],
+    grpc_only = True,
+    visibility = [
+        "//:__pkg__",
+    ],
+    deps = [":hello_world_cc_proto"],
+)
+
 cc_test(
     name = "grpc_credential_types",
     timeout = "long",
     srcs = ["grpc_credential_types.cc"],
+    copts = [
+        "-I$(GENDIR)",
+    ],
     tags = [
         "integration-test",
         "integration-test-production",
     ],
     deps = [
+        ":hello_world_cc_grpc",
+        ":hello_world_cc_proto",
         "//:experimental-iam",
         "//:spanner",
         "//google/cloud:google_cloud_cpp_common",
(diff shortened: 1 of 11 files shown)


google/cloud/examples/BUILD
```diff
@@ -41,15 +47,49 @@ cc_test(
...
+    ],
+    deps = [":hello_world_protos"],
+)
+
+cc_grpc_library(
+    name = "hello_world_cc_grpc",
+    srcs = [":hello_world_protos"],
+    grpc_only = True,
+    visibility = [
+        "//:__pkg__",
```
devjgm, line 74: I think this visibility rule is granting access to this target from the targets defined in our top-level `BUILD` file, is that the intent? If the goal is to simply make this target private to the current package (i.e. `//google/cloud/example`), then I think you may want `//visibility:private`

google/cloud/examples/BUILD
```diff
@@ -41,15 +47,49 @@ cc_test(
...
+    ],
+    deps = [":hello_world_cc_proto"],
+)
+
 cc_test(
     name = "grpc_credential_types",
     timeout = "long",
     srcs = ["grpc_credential_types.cc"],
+    copts = [
+        "-I$(GENDIR)",
```
devjgm, line 84: I'm surprised this `-I` is needed. I don't think we usually need that w/ proto deps. Do you know why it's needed in this case?

google/cloud/examples/BUILD
coryan: Fixed.

google/cloud/examples/BUILD
coryan: But we do need something similar in our proto deps:

https://github.com/googleapis/google-cloud-cpp/blob/ebb6247af79d63df0e4ebc33c2cf18334a5223f2/bazel/googleapis.BUILD#L19-L29

and:

https://github.com/googleapis/google-cloud-cpp/blob/ebb6247af79d63df0e4ebc33c2cf18334a5223f2/google/cloud/BUILD#L100

google/cloud/examples/BUILD
```diff
@@ -41,15 +47,49 @@ cc_test(
     deps = [":gcs2cbt_lib"],
 )
 
+proto_library(
+    name = "hello_world_protos",
+    srcs = [
+        "hello_world_grpc/hello_world.proto",
+    ],
+    visibility = [
+        "//:__pkg__",
```
devjgm, line 56: Probably want `//visibility:private` here too

google/cloud/examples/BUILD
devjgm: Oh, so this is to enable angle-bracket includes? If we could accomplish this similar to how `googleapis_system_includes` works using `includes = [ "$(GENDIR)" ]` on the proto/grpc targets instead of the users of that target, that may be preferable. But I'll let you see what works and pick one.

devjgm (approved):

google/cloud/examples/BUILD
coryan: Doh!

google/cloud/examples/BUILD
coryan: Neither rule supports that parameter:

https://github.com/grpc/grpc/blob/cc9326c93650f5e853fe02922ddb476af70ce303/bazel/cc_grpc_library.bzl#L29
https://docs.bazel.build/versions/master/be/c-cpp.html#cc_proto_library

I could add an empty `cc_library()` rule with just that definition, hardly seems better, do you think it would be? I like :shrug: no idea :shrug:

devjgm:
What u have is fine.

devjgm (approved):
Thanks.

PR #6590, opened 2021-05-17
cleanup: prevent behavior change from a gRPC change

jacobsa:
This is the same story as ee82393b. I didn't realize that there were two
other copies of this class.

@coryan

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6590)


coryan (approved):
LGTM, seems like we need to do some refactoring (see #6592)

PR #6596, opened 2021-05-17
ci: support new gRPC examples in publish-docs

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6596)

### DIFF
--- a/cmake/GoogleCloudCppCommon.cmake
+++ b/cmake/GoogleCloudCppCommon.cmake
@@ -62,6 +62,7 @@ else ()
         set(DOXYGEN_SEARCH_INCLUDES YES)
         set(DOXYGEN_INCLUDE_PATH
             "${PROJECT_SOURCE_DIR}" "${PROJECT_BINARY_DIR}"
+            "${PROJECT_BINARY_DIR}/google/cloud/examples"
             "${PROJECT_BINARY_DIR}/external/googleapis")
         set(DOXYGEN_GENERATE_LATEX NO)
         set(DOXYGEN_GRAPHICAL_HIERARCHY NO)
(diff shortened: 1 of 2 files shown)


cmake/GoogleCloudCppCommon.cmake
```diff
@@ -62,6 +62,7 @@ else ()
         set(DOXYGEN_SEARCH_INCLUDES YES)
         set(DOXYGEN_INCLUDE_PATH
             "${PROJECT_SOURCE_DIR}" "${PROJECT_BINARY_DIR}"
+            "${PROJECT_BINARY_DIR}/google/cloud/examples"
```
devjgm, line 65: I believe the full path to the generated include file should be `${PROJECT_BINARY_DIR}/google/cloud/examples/hello_world_grpc/hello_world.grpc.pb.h` and it's included via `<google/cloud/examples/hello_world_grpc/hello_world.grpc.pb.h>`, so I don't understand why this include path would be needed. But if it fixes the build... LGTM.

devjgm (approved):

cmake/GoogleCloudCppCommon.cmake
coryan: I think I can explain why. The `.grpc.pb.h` file reads:

```cc
#include "hello_world_grpc/hello_world.pb.h"
```

Why?  Because `protobuf` has strong opinions.  Could we fix that?  Maybe.  It would probably require invoking `protoc` from the top-level directory instead of the currrent dir.

devjgm: Let's certainly not call protoc ourselves. But thanks for explaining.

PR #6597, opened 2021-05-17
ci: log cache.sh command

devjgm:
Sometimes it's useful to use the cached artifacts that were used in a GCB build even when running locally in docker. This may help reproduce build breaks if they're cache related. This prints the command to run locally to pull the production cache

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6597)

### DIFF
--- a/ci/cloudbuild/cache.sh
+++ b/ci/cloudbuild/cache.sh
@@ -39,6 +39,9 @@ function print_usage() {
   sed -n '17,/^$/s/^# \?//p' "${PROGRAM_PATH}"
 }
 
+io::log "Executing command:"
+printf "env -C '%s' %q\n" "$(pwd)" "$@"
+
 # Use getopt to parse and normalize all the args.
 PARSED="$(getopt -a \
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/cache.sh
```diff
@@ -39,6 +39,11 @@ function print_usage() {
   sed -n '17,/^$/s/^# \?//p' "${PROGRAM_PATH}"
 }
 
+io::log "Running:"
+printf "env -C '%s' %s" "$(pwd)" "$0"
+printf " %q" "$@"
+printf "\n"
```
coryan, line 45: nit: can one say `printf " %q\n" "$@"` ?

coryan (approved):

ci/cloudbuild/cache.sh
devjgm: No. That would run `" %q\n` for each arg, printing each one, escaped, on its own line.

PR #6603, opened 2021-05-18
doc(adr): handling token refresh failures

### LINKED ISSUE #6337: Consider interaction of service account impersonation credentials and retry loops
When impersonating a service account via the [IAM Credentials Service] one may need to retry the request to create the token.  In principle, we should let the caller perform the retry, as they know what retry policy is in effect. This does not work when the caller is performing a non-idempotent operation. In that case the caller will not retry, but they could have, the operation in question never started.

I think we should let this be, and just record the decision (ADR?), this issue is to record any discussion.

[IAM Credentials Service]: https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken

coryan:
Fixes #6337

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6603)

### DIFF
--- a/doc/adr/2021-05-18-credential-failures-participate-in-retry-loop.md
+++ b/doc/adr/2021-05-18-credential-failures-participate-in-retry-loop.md
@@ -1,0 +1,29 @@
+**Title**: Credential refresh failures participate in retry loop.
+
+**Status**: accepted
+
+**Context**: many credential types require fetching tokens from a remote
+service. For example, service account impersonation credentials must obtain an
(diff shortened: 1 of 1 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6603)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

devjgm (approved):

PR #6609, opened 2021-05-18
ci: reintroduce logging of important build commands

devbww:
Rebrand `io::log_cmdline` as `io::log_and_run`, which, as the
name suggests, now also executes the command so as to eliminate
duplication at the call site.  This version also adds quoting.

Introduce `io::log_and_run` into the `ci/cloudbuild/build.sh`
script. Additions to any of the `ci/cloudbuild/builds/*.sh`
scripts can be made as the need arises.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6609)


devjgm (approved):
LGTM as-is. I'd also be OK w/ a shorter name like `io::run` since the `io::` part gives some hint that something will be logged also. But I'm happy w/ whatever you decide.

devbww:
> I'd also be OK w/ a shorter name like `io::run`

Done.

PR #6611, opened 2021-05-19
feat(storage): add Retry Test API to emulator

frankyn:
cc: @tritone 

Adding Retry Test API support to the emulator

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6611)

### DIFF
--- a/google/cloud/storage/emulator/README.md
+++ b/google/cloud/storage/emulator/README.md
@@ -88,14 +88,19 @@ instructions.
 
 ## Retry Test API
 
-The GCS Emulator at the time of writing (5/19/21), only provided support
-for reproducing transient errors by passing instructions through a header
-which is not easily implementable across the 8 language libraries that Cloud
-Storage currently supports. Retry Test API introduces a method to supply failure
-configuration and supplies GCS libraries with a unique identifier which is
-used to identify how to fail a specified GCS API operation. In addition,
-the Retry Test API performs accounting to tell a developer if expected
-failures occured.
+The "Retry Test API" offers a mechanism to describe more complex retry scenarios
+while sending a single, constant header through all the HTTP requests from a
+test program. Retry Test provides accounting of failures used to validate
+the expected failures were experienced by the emulator and not accidentally missed.
+
+Previous versions of the GCS emulator used a custom header in the RPC to
+control the behavior of each RPC, for some test scenarios this required sending
+different header with the first retry attempt vs. subsequent attempts. Producing
+different headers in each attempt is not easy to implement with some client libraries.
+
+Sending a constant header with all RPCs can be implemented across all client libraries,
+and to some degree decouples the test setup from the test execution.
+
 
 ### Creating a new Retry Test
 
--- a/google/cloud/storage/emulator/README.md
+++ b/google/cloud/storage/emulator/README.md
@@ -84,3 +84,40 @@ For N==1 and N==2 behave like `return-305-after-256K`, for `N>=3` ignore the
 failure instruction and return successfully. This is used to test failures during
 retry, the client cooperates by sending the retry counter in the failure
 instructions.
+
+
+## Retry Test API
+
+### Creating a new Retry Test
+
+```bash
+    curl -X POST "http://localhost:9000/retry_test" -H 'Content-Type: application/json' \
+         -d '{"instructions":{"storage.buckets.list": ["return-503"]}}'
+```
+
+### Get a Retry Test resource
+
+```bash
+    curl -X GET "http://localhost:9000/retry_test/1d05c20627844214a9ff7cbcf696317d"
+```
+
+### Delete a Retry Test resource
+
+```bash
+    curl -X DELETE "http://localhost:9000/retry_test/1d05c20627844214a9ff7cbcf696317d"
+```
+
+### Causing a failure using x-retry-test-id header
+
+```bash
+    curl -H "x-retry-test-id: 1d05c20627844214a9ff7cbcf696317d" "http://localhost:9100/storage/v1/b?project=test"
+```
+
+### Forced Failures Supported
+
+| Failure Id              | Description
+| ----------------------- | ---
+| return-X                | Emulator will fail with HTTP code provided for `X`, e.g. return-503 returns a 503
+| return-X-after-YK       | Not Supported. Return X after YKiB of uploaded data
+| return-broken-stream    | Emulator will fail after sending 10 bytes
+| return-reset-connection | Emulator will fail with a reset connection
--- a/google/cloud/storage/emulator/README.md
+++ b/google/cloud/storage/emulator/README.md
@@ -88,29 +88,50 @@ instructions.
 
 ## Retry Test API
 
+The GCS Emulator at the time of writing (5/19/21), only provided support
+for reproducing transient errors by passing instructions through a header
+which is not easily implementable across the 8 language libraries that Cloud
+Storage currently supports. Retry Test API introduces a method to supply failure
+configuration and supplies GCS libraries with a unique identifier which is
+used to identify how to fail a specified GCS API operation. In addition,
+the Retry Test API performs accounting to tell a developer if expected
+failures occured.
+
 ### Creating a new Retry Test
 
+The following cURL request will create a Retry Test resource which emits a 503
+when a buckets list operation is received by the emulator with the returned
+retry test ID.
+
 ```bash
-    curl -X POST "http://localhost:9000/retry_test" -H 'Content-Type: application/json' \
-         -d '{"instructions":{"storage.buckets.list": ["return-503"]}}'
+curl -X POST "http://localhost:9000/retry_test" -H 'Content-Type: application/json' \
+     -d '{"instructions":{"storage.buckets.list": ["return-503"]}}'
 ```
 
 ### Get a Retry Test resource
 
+Get Retry Test resource by id "1d05c20627844214a9ff7cbcf696317d".
+
 ```bash
-    curl -X GET "http://localhost:9000/retry_test/1d05c20627844214a9ff7cbcf696317d"
+curl -X GET "http://localhost:9000/retry_test/1d05c20627844214a9ff7cbcf696317d"
 ```
 
 ### Delete a Retry Test resource
 
+Delete Retry Test resource by id "1d05c20627844214a9ff7cbcf696317d".
+
 ```bash
-    curl -X DELETE "http://localhost:9000/retry_test/1d05c20627844214a9ff7cbcf696317d"
+curl -X DELETE "http://localhost:9000/retry_test/1d05c20627844214a9ff7cbcf696317d"
 ```
 
 ### Causing a failure using x-retry-test-id header
 
+The following cURL request will attempt to list buckets and the emulator will emit
+a `503` error once based on the Retry Test created above. Subsequent list buckets
+operations will succeed.
+
 ```bash
-    curl -H "x-retry-test-id: 1d05c20627844214a9ff7cbcf696317d" "http://localhost:9100/storage/v1/b?project=test"
+curl -H "x-retry-test-id: 1d05c20627844214a9ff7cbcf696317d" "http://localhost:9100/storage/v1/b?project=test"
 ```
 
 ### Forced Failures Supported
--- a/google/cloud/storage/emulator/README.md
+++ b/google/cloud/storage/emulator/README.md
@@ -101,7 +101,6 @@ different headers in each attempt is not easy to implement with some client libr
 Sending a constant header with all RPCs can be implemented across all client libraries,
 and to some degree decouples the test setup from the test execution.
 
-
 ### Creating a new Retry Test
 
 The following cURL request will create a Retry Test resource which emits a 503
--- a/google/cloud/storage/emulator/database.py
+++ b/google/cloud/storage/emulator/database.py
@@ -16,16 +16,18 @@ import collections
 import json
 import os
 import uuid
+import re
 
 import gcs
 import utils
 
+from schema import Schema, SchemaError, And, Use
 from google.cloud.storage_v1.proto import storage_pb2 as storage_pb2
 
 
 class Database:
     def __init__(
-        self, buckets, objects, live_generations, uploads, rewrites, retry_tests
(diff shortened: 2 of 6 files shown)


google/cloud/storage/emulator/emulator.py
```diff
@@ -42,6 +43,55 @@ def index():
     return "OK"
 
 
+"""
+TODO:
```
coryan, line 47: Create an issue to track this TODO and name it here `TODO(#1234)`

google/cloud/storage/emulator/tests/test_utils.py
```diff
@@ -472,6 +474,151 @@ def test_error_propagation(self):
...
+            response = self.conn.getresponse()
+
+        self.conn.request("DELETE", "/retry_test/%s" % (retry_test_id))
+        response = self.conn.getresponse()
+        body = response.read().decode("utf-8")
+        self.assertIn("Deleted %s" % (retry_test_id), body)
+        self.conn.close()
+
+    def test_retry_test_broken_stream(self):
+
```
coryan, line 550: nit: blank line

google/cloud/storage/emulator/tests/test_utils.py
```diff
@@ -472,6 +474,151 @@ def test_error_propagation(self):
...
+        self.conn.close()
+
+        self.conn.request("DELETE", "/retry_test/%s" % (retry_test_id))
+        response = self.conn.getresponse()
+        body = response.read().decode("utf-8")
+        self.assertIn("Deleted %s" % (retry_test_id), body)
+        self.conn.close()
+
+    def test_retry_test_503(self):
+
```
coryan, line 580: nit: blank line?

google/cloud/storage/emulator/README.md
```diff
@@ -84,3 +84,40 @@ For N==1 and N==2 behave like `return-305-after-256K`, for `N>=3` ignore the
 failure instruction and return successfully. This is used to test failures during
 retry, the client cooperates by sending the retry counter in the failure
 instructions.
+
+
+## Retry Test API
+
```
coryan, line 90: Can you write something here about the *motivation* for these new APIs and resources?  Why do we create "named" Retry Test objects instead of passing the instructions via `x-goog-emulator-instructions`?  What problem does this new approach solve?
I can make some guesses, but why guess when we can write it down so it is clear for our future selves.

google/cloud/storage/emulator/README.md
```diff
@@ -84,3 +84,40 @@ For N==1 and N==2 behave like `return-305-after-256K`, for `N>=3` ignore the
...
+### Creating a new Retry Test
+
+```bash
+    curl -X POST "http://localhost:9000/retry_test" -H 'Content-Type: application/json' \
```
coryan, line 94: I think block quoting via indentation and via backticks is overkill, maybe?

google/cloud/storage/emulator/tests/test_utils.py
```diff
@@ -472,6 +474,151 @@ def test_error_propagation(self):
...
+        self.conn.close()
+
+        self.conn.request("DELETE", "/retry_test/%s" % (retry_test_id))
+        response = self.conn.getresponse()
+        body = response.read().decode("utf-8")
+        self.assertIn("Deleted %s" % (retry_test_id), body)
+        self.conn.close()
+
+    def test_retry_test_reset_connection(self):
+
```
coryan, line 522: nit: blank line does not seem idiomatic?

google/cloud/storage/emulator/emulator.py
```diff
@@ -99,6 +149,7 @@ def root_put_object_with_bucket(bucket_name, object_name):
 
 
 @gcs.route("/b", methods=["GET"])
+@retry_test(method="storage.buckets.list")
```
coryan, line 152: Okay, I think I understand how this works, you can wrap any other API with the retry test wrapper, with minimal effort.  Very neat. Still does not answer why we need a "retry test" resource...

google/cloud/storage/emulator/tests/test_utils.py
```diff
@@ -472,6 +474,151 @@ def test_error_propagation(self):
...
+        if self.emulator is None:
+            self.skipTest("CLOUD_STORAGE_EMULATOR_ENDPOINT is not set")
+        self.conn = http.client.HTTPConnection(
+            self.emulator[len("http://") :], timeout=10
+        )
+
+    def tearDown(self):
+        self.conn.close()
+
+    def test_retry_test_resource_crud(self):
```
coryan, line 492: We seem to be missing a test to verify that *all* APIs are wrapped. In case I forgot to add `@retry_test(method="fancy.new.method")` when we add `fancy.new.method` to the APIs.

Do you think it is possible to list all the "methods" in the JSON / XML APIs and write a test to verify each of them got a wrapper, maybe using some `flask` introspection or using the discovery doc?  If it is a lot of work, please open a bug as a future enhancement.

google/cloud/storage/emulator/utils/common.py
```diff
@@ -330,3 +338,88 @@ def enforce_patch_override(request):
...
+                database.dequeue_next_instruction(test_id, method)
+                fd = request.environ["gunicorn.socket"]
+                fd.setsockopt(
+                    socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0)
+                )
+                fd.close()
+                sys.exit(1)
+            elif items[0] == "broken-stream":
+
+                def retry_test_response_handler(data):
```
coryan, line 395: This is fairly confusing to me... I think you are declaring a local function so the value returned below is a different one?

Consider instead having two functions and picking what to return:

```py
def handle_retry_test_instruction(database, request, method):
  def simple_handler(data):
    return data

  def streaming_handler(data):
    ...
   return __wrap_in_response_instance(data, streamer())

  handler = simple_handler # Use simple_handler by default
   ...
  elif items[0] == "broken-stream":
    handler = streaming_handler
  ... ... ...
  ..
  return handler
```

coryan (changes requested):
See below, I like the general direction btw.

google/cloud/storage/emulator/README.md
frankyn: Yea... looks like i just couldn't decide which to use. I'll address this issue.

google/cloud/storage/emulator/emulator.py
```diff
@@ -42,6 +43,55 @@ def index():
     return "OK"
 
 
+"""
+TODO:
+    1. Introducing failures into uploads with return-XXX-after-YYYk
+"""
+
+# Needs to be defined in emulator.py to keep context of flask and db global variables
```
tritone, line 51: This is very neat!

google/cloud/storage/emulator/emulator.py
```diff
@@ -42,6 +43,55 @@ def index():
...
+    test_instruction_set = payload.get("instructions", None)
+    if not test_instruction_set:
+        return flask.Response("instructions is not defined", status=400)
+    retry_test = db.insert_retry_test(test_instruction_set)
+    retry_test_response = json.dumps(retry_test)
+    return flask.Response(retry_test_response, status=200)
+
+
+@root.route("/retry_test/<test_id>", methods=["GET"])
+def get_retry_test(test_id):
```
tritone, line 84: What happens for a test that doesn't exist?

google/cloud/storage/emulator/database.py
```diff
@@ -251,3 +256,68 @@ def insert_rewrite(self, rewrite):
...
+
+    def get_retry_test(self, retry_test_id):
+        retry_test = self.retry_tests.get(retry_test_id)
+        if retry_test is None:
+            utils.error.notfound("Retry Test %s" % retry_test_id, context=None)
+        return self.__to_serializeable_retry_test(retry_test)
+
+    def insert_retry_test(self, instructions):
+        retry_test_id = uuid.uuid4().hex
+        self.retry_tests[retry_test_id] = {
```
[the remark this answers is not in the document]
frankyn: terribly, i didn't capture the case.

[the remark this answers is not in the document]
cojenco: +1, I accidentally created a retry test with a typo in the instructions just now.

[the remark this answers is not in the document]
frankyn: Added validation for methods and instruction format: 
https://github.com/googleapis/google-cloud-cpp/pull/6611/commits/9af654d815d2aee15fc696038321b174228ed6db

google/cloud/storage/emulator/utils/common.py
frankyn: Addressed in: https://github.com/googleapis/google-cloud-cpp/pull/6611/commits/b996679af727df8380ef25478f7a839446249473

frankyn: One more refactor: https://github.com/googleapis/google-cloud-cpp/pull/6611/commits/e729f58c793acfeda31d52ffd4a662c316564ce6

google/cloud/storage/emulator/tests/test_utils.py
frankyn: Filed: https://github.com/googleapis/google-cloud-cpp/issues/6625

google/cloud/storage/emulator/README.md
coryan: I am now really confused.  The motivation is that passing instructions in a header is hard, but passing the `x-retry-test-id: $arbitrary_string` is easy?

google/cloud/storage/emulator/README.md
```diff
@@ -84,3 +84,61 @@ For N==1 and N==2 behave like `return-305-after-256K`, for `N>=3` ignore the
 failure instruction and return successfully. This is used to test failures during
 retry, the client cooperates by sending the retry counter in the failure
 instructions.
+
+
+## Retry Test API
+
+The GCS Emulator at the time of writing (5/19/21), only provided support
```
coryan, line 91: Dates in YYYY-MM-DD format please.

google/cloud/storage/emulator/utils/common.py
```diff
@@ -330,3 +338,87 @@ def enforce_patch_override(request):
...
+        return data
+    return FlaskResponse(to_wrap)
+
+
+def __get_streamer_response_fn(database, method, conn, test_id):
+    def response_handler(data):
+        def streamer():
+            database.dequeue_next_instruction(test_id, method)
+            d = __extract_data(data)
+            chunk_size = 4
```
coryan, line 363: seems like `chunk_size` and `10` should be maybe parameters?  With default values?

google/cloud/storage/emulator/README.md
```diff
@@ -84,3 +84,61 @@ For N==1 and N==2 behave like `return-305-after-256K`, for `N>=3` ignore the
...
+for reproducing transient errors by passing instructions through a header
+which is not easily implementable across the 8 language libraries that Cloud
+Storage currently supports. Retry Test API introduces a method to supply failure
+configuration and supplies GCS libraries with a unique identifier which is
+used to identify how to fail a specified GCS API operation. In addition,
```
coryan, line 96: This second motivation seems important enough by itself, and I assume the logs are more readable too.

coryan (approved):
I think the motivation could be made crisper. But otherwise this LGTM.

google/cloud/storage/emulator/README.md
tritone: The tricky part is that in several languages, configuring the client to send an extra header happens per-library call or per-client, rather than per-request sent. Therefore, it's hard to configure the client to send an extra header for the first N attempts and then stop sending it for the last request. Having the client send the same header for every request for a given test (while the emulator tracks when to stop forcing failures) is easier.

google/cloud/storage/emulator/database.py
```diff
@@ -269,13 +279,38 @@ def __to_serializeable_retry_test(cls, retry_test):
...
+    def __validate_instructions(self, instructions):
+        expected_instructions_format = [
+            utils.common.retry_return_error_code.pattern,
+            utils.common.retry_return_error_connection.pattern,
+            utils.common.retry_return_error_after_bytes.pattern,
+        ]
+        search_expression = re.compile(
+            "(%s)" % ")|(".join(expected_instructions_format)
+        )
+        expected_schema = Schema(
```
tritone, line 300: Have not seen this technique using a schema in python before; it feels more reminiscent of unmarshaling in Go. Pretty cool.

tritone (approved):
Looking good!

google/cloud/storage/emulator/README.md
coryan: Okay, let's say that then:

```md
## Retry Test API

The "Retry Test API" offers a mechanism to describe more complex retry scenarios while sending a single, constant header through all the HTTP requests from a test program.

Previous versions of the GCS emulator used a custom header in the RPC to control the behavior of each RPC,
for some test scenarios this required sending different header with the first retry attempt vs. subsequent attempts. Producing different headers in each attempt is not easy to implement with some client libraries.

Sending a constant header with all RPCs can be implemented across all client libraries, and to some degree decouples the test setup from the test execution.
```

coryan (changes requested):

google/cloud/storage/emulator/README.md
```diff
@@ -84,3 +84,66 @@ For N==1 and N==2 behave like `return-305-after-256K`, for `N>=3` ignore the
...
+the expected failures were experienced by the emulator and not accidentally missed.
+
+Previous versions of the GCS emulator used a custom header in the RPC to
+control the behavior of each RPC, for some test scenarios this required sending
+different header with the first retry attempt vs. subsequent attempts. Producing
+different headers in each attempt is not easy to implement with some client libraries.
+
+Sending a constant header with all RPCs can be implemented across all client libraries,
+and to some degree decouples the test setup from the test execution.
+
```
coryan, line 103: nit: blank line ?

coryan (approved):
LGTM

google/cloud/storage/emulator/README.md
frankyn: ... i love extra empty lines.

PR #6613, opened 2021-05-19
ci: allocate a docker tty only when we have one

devbww:
Conditionalize the docker `--tty` flag so that we only allocate
a tty inside the image when we have a tty to connect it to. The
major upshot of this is that you can now pipe shell commands into
`ci/cloudbuild/build.sh --docker --trigger=<trigger> --docker-shell`.

Previously, under the unconditional `--tty=true`, docker would set
`TERM=xterm`, which was bad when there was no external tty, and
possibly wrong when there was one.  So, we now also conditionally
set `${TERM}` within the image depending on whether there is an
external tty and what its terminal type is, and otherwise default
to `TERM-dumb`. This should eliminate the xterm control sequences
we see in log files (and correct things for those who do not use
an xterm-compatible terminal).


devjgm (approved):
LGTM. If the control sequences you mentioned are the ones in GCB logs ([example](https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/6613/1a09b8395a62ee2977323c3c61c680c3acc37efa/fedora-33-check-api/log-fada2903-509c-4259-a4cd-2a033d86cd8b.txt)), then I don't think this will fix that because this code is only run when we do local `--docker` builds.

Although, after seeing this PR, I do wonder if we unconditionally set `TERM=dumb` in our `cloudbuild.yaml` file, if that would eliminate those control codes. May be worth a try.

devbww:
Yes ... I was just experimenting with that.  It seems to me that exporting the same TERM I currently have for `docker run` also to `gcloud builds submit` should cover that, while also preserving the coloring when possible.  But I'm still verifying.

devbww:
FYI:  Nothing I've tried so far (`TERM=dumb` in `options: env:` in `cloudbuild.yaml`; `gcloud config set disable_color True`; and variations of those) has been able to defeat the colorized output from `Step #0`.  I'll keep experimenting, but if you have additional ideas, I'm all ears.

devjgm:
> FYI: Nothing I've tried so far (`TERM=dumb` in `options: env:` in `cloudbuild.yaml`; `gcloud config set disable_color True`; and variations of those) has been able to defeat the colorized output from `Step #0`. I'll keep experimenting, but if you have additional ideas, I'm all ears.

Wild guess: Step 0 is the kaniko step, and it uses the logrus golang logging library. It looks like that library checks for some env var to disable log coloring:

https://github.com/sirupsen/logrus/blob/b50299cfaaa1bca85be76c8984070e846c7abfd2/text_formatter.go#L117-L127

So maybe try setting `CLICOLOR=0` in our `cloudbuild.yaml` file?

devbww:
> So maybe try setting `CLICOLOR=0`

That didn't work because it looks like kaniko is forcing color on the logging library.  (Aside: It looks like the color support is hardwired ANSI.)

But that did lead me to the flag that can force text (`--log-format=text`).  Unfortunately, it doesn't look like there is a way to set the logging library back to "auto" mode (color if on a tty), so I'll play a little with making the value of the flag conditional.

ci/cloudbuild/build.sh
```diff
@@ -313,6 +315,7 @@ subs+=("_BUILD_NAME=${BUILD_NAME}")
 subs+=("_TRIGGER_SOURCE=manual-${account}")
 subs+=("_PR_NUMBER=") # Must be empty or a number, and this is not a PR
 subs+=("_LOGS_BUCKET=${project}_cloudbuild")
+subs+=("_LOG_FORMAT=$([[ -t 0 ]] && echo color || echo text)")
```
devjgm, line 318: This doesn't look quite right to me. It looks like this is checking if the _local_ script is running with a tty, and based on that it's telling the _remote_ GCB build whether or not to use color.

Additionally, for triggered builds, this part of the script won't be run at all. A triggered build enters this script with the `--local` flag. So my guess is that in `cloudbuild.yaml`, kaniko is being run with `--log-format=""`, which seems like it uses `text` in that case?

If this is correct, maybe we should just change the `cloudbuild.yaml` file to unconditionally pass `--log-format=text` for GCB builds, which I think, will never have a tty. WDYT?

devbww: I'm happy to go with the unconditional `--log-format=text`, so I'll update shortly.

That said, I can get both `text` and `color` behavior respectively with, say, ...
```
ci/cloudbuild/build.sh --trigger=asan-pr </dev/null 
ci/cloudbuild/build.sh --trigger=asan-pr
```

I don't think we'd ever generate `--log-format=""`.

devjgm: _We_ wouldn't generate `--log-format=""`, but GCB would when it triggers a PR/CI build because it doesn't use `build.sh` to start that build. So the `_LOG_FORMAT` substitution would be unset and would expand to the empty string when referenced in `cloudbuild.yaml`.

It is a bit confusing because `build.sh --trigger <file>` just means to look up the `_DISTRO` and `_BUILD_NAME` from the named trigger file, then submit the build to GCB. That's different than a build that's automatically triggered for a PR or CI build.

Regardless, I _think_ GCB will never run our builds with a TTY, so we might as well just tell Kaniko to use text logs unconditionally.

devjgm (approved):

PR #6617, opened 2021-05-20
feat: custom trust store with unified credentials

### LINKED ISSUE #6311: Support SSL parameters for g::c::internal::GrpcAuthenticationStrategy
The unified authentication client for gRPC needs to call `grpc::SslCredentials`, applications may want to configure that call by setting the SSL parameters:

https://grpc.github.io/grpc/cpp/structgrpc_1_1_ssl_credentials_options.html

Note this includes support for mutual TLS authentication (see [aip/4114](https://google.aip.dev/auth/4114)).

coryan:
Some applications may not be able to use the system's trust store (aka
CA certificates, aka root of trust files). For example, because the
application is running in a lightweight container without system
configuration files. And because the credentials need to use this trust
store to create and/or refresh access tokens, we need a unified
configuration parameter to set the trust store.

This PR introduces `google:cloud:CAInfo` an option that should point
to a file with the CA certificate bundle (typically a PEM file). For
REST-based libraries we simply point libcurl (and indirectly OpenSSL) to
this file. For gRPC-based library we load the contents of this file and
feed them to gRPC.

Fixes #6311 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6617)


devjgm:
/gcbrun 

(just using this build to test handling GCB notifications)

ci/cloudbuild/builds/lib/integration.sh
```diff
@@ -30,6 +30,11 @@ source module ci/lib/io.sh
 export PATH="${HOME}/.local/bin:${PATH}"
 python3 -m pip install --quiet --user -r "${PROJECT_ROOT}/google/cloud/storage/emulator/requirements.txt"
 
+echo "================================================================"
```
devjgm, line 33: We've avoided these echo lines in the GCB builds by using `io::log_h2` and friends instead. Consider switching to that, or if you prefer to keep it quieter keep the `io::log_yellow` and just drop this echo line.

google/cloud/credentials.h
```diff
@@ -203,6 +203,49 @@ struct UnifiedCredentialsOption {
...
 
+/**
+ * Configures a custom CA (Certificates Authority) certificates file.
+ *
+ * Most applications should use the system's root certificates and should avoid
+ * setting this option unnecessarily. A common exception to this recommendation
+ * are containerized applications. These often deploy without system's root
+ * certificates and need to explicitly configure a root of trust.
+ *
+ * The value of this option should be the name of a file in [PEM format].
```
devjgm, line 214: I realize that `CAINFO` is a term that curl uses, and it may be good to follow suit w/ this name. But "CAINFO" sounds pretty ambiguous to me. I didn't realize this was a file path until I saw it used w/ `ifstream` below (and yes, I must have missed this sentence). Would naming this something like `CARootPemOption` be clearer? Or different and bad? WDYT?

ci/cloudbuild/builds/lib/integration.sh
coryan: Done.

google/cloud/credentials.h
coryan: You are right, changed the name, PTAL.

devjgm (approved):

PR #6620, opened 2021-05-23
feat(pubsub): add ability to set the ack deadline extension

### LINKED ISSUE #6283: Give applications better control over message ack deadline extension
In Pub/Sub the service resends messages that are not acked before its ack deadline.  This deadline can be controlled by the service configuration, and extended by the library based on subscription settings.  Currently the C++ client library extends the deadlines by the maximum allowed interval, this can be a problem.  Consider a message with a relatively long deadline, and an application that crashes shortly after the deadline is extended.  The service will need to wait for the full deadline before resending the message.  It would be better if the library extended the deadline "just enough" to dead with failures.

The client libraries in other languages track the time required to ack a message, and only extend the message to p99 of this latency.

dbolduc:
Fixes #6283 

Note that this change gives the users the ability to set their desired deadline extension (in seconds). It does not calculate and use the p99 latency, like the other Pub/Sub libraries do.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620)

### DIFF
--- a/google/cloud/pubsub/internal/subscription_lease_management.cc
+++ b/google/cloud/pubsub/internal/subscription_lease_management.cc
@@ -20,7 +20,6 @@ namespace pubsub_internal {
 inline namespace GOOGLE_CLOUD_CPP_PUBSUB_NS {
 
 std::chrono::seconds constexpr SubscriptionLeaseManagement::kMinimumAckDeadline;
-std::chrono::seconds constexpr SubscriptionLeaseManagement::kMaximumAckDeadline;
 std::chrono::seconds constexpr SubscriptionLeaseManagement::kAckDeadlineSlack;
 
 void SubscriptionLeaseManagement::Start(BatchCallback cb) {
@@ -101,7 +100,7 @@ void SubscriptionLeaseManagement::RefreshMessageLeases(
 
   std::vector<std::string> ack_ids;
   ack_ids.reserve(leases_.size());
-  auto extension = kMaximumAckDeadline;
+  auto extension = max_deadline_extension_;
   auto const now = std::chrono::system_clock::now();
   for (auto const& kv : leases_) {
     // This message lease cannot be extended any further, and we do not want to
--- a/google/cloud/pubsub/internal/subscription_lease_management.h
+++ b/google/cloud/pubsub/internal/subscription_lease_management.h
@@ -36,7 +36,6 @@ class SubscriptionLeaseManagement
  public:
   static auto constexpr kAckDeadlineSlack = std::chrono::seconds(2);
   static auto constexpr kMinimumAckDeadline = std::chrono::seconds(10);
-  static auto constexpr kMaximumAckDeadline = std::chrono::seconds(600);
 
   /**
    * A wrapper to create timers.
@@ -50,7 +49,8 @@ class SubscriptionLeaseManagement
       google::cloud::CompletionQueue cq,
       std::shared_ptr<SessionShutdownManager> shutdown_manager,
       std::shared_ptr<SubscriptionBatchSource> child,
-      std::chrono::seconds max_deadline_time) {
+      std::chrono::seconds max_deadline_time,
+      std::chrono::seconds max_deadline_extension) {
     auto timer_factory =
         [cq](std::chrono::system_clock::time_point tp) mutable {
           return cq.MakeDeadlineTimer(tp).then(
@@ -61,7 +61,8 @@ class SubscriptionLeaseManagement
     return std::shared_ptr<SubscriptionLeaseManagement>(
         new SubscriptionLeaseManagement(
             std::move(cq), std::move(shutdown_manager),
-            std::move(timer_factory), std::move(child), max_deadline_time));
+            std::move(timer_factory), std::move(child), max_deadline_time,
+            max_deadline_extension));
   }
 
   static std::shared_ptr<SubscriptionLeaseManagement> CreateForTesting(
@@ -69,11 +70,13 @@ class SubscriptionLeaseManagement
       std::shared_ptr<SessionShutdownManager> shutdown_manager,
       TimerFactory timer_factory,
       std::shared_ptr<SubscriptionBatchSource> child,
-      std::chrono::seconds max_deadline_time) {
+      std::chrono::seconds max_deadline_time,
+      std::chrono::seconds max_deadline_extension) {
     return std::shared_ptr<SubscriptionLeaseManagement>(
         new SubscriptionLeaseManagement(
             std::move(cq), std::move(shutdown_manager),
-            std::move(timer_factory), std::move(child), max_deadline_time));
+            std::move(timer_factory), std::move(child), max_deadline_time,
+            max_deadline_extension));
   }
 
   void Start(BatchCallback cb) override;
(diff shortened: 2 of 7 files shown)


dbolduc (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @dbolduc)

___
*[google/cloud/pubsub/internal/subscription_lease_management.h, line 53 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620#-MaMiGCLETsUTuNjmXUR:-MaMiGCLETsUTuNjmXUS:b-rmxu3e) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a0c0971ad5af7a886dc3eec710f5d9283909ce21/google/cloud/pubsub/internal/subscription_lease_management.h#L53)):*
> ```objc
>       std::shared_ptr<SubscriptionBatchSource> child,
>       std::chrono::seconds max_deadline_time,
>       std::chrono::seconds max_deadline_extension) {
> ```

This could be a `DeadlineExtension` wrapper class.

For now it would only have
* `std::chrono::seconds max_extension_`
* `std::chrono::seconds GetMaxExtension() const` that just returns `max_extension_`

In the future we could
* Add `double latency_percentile_`
* Add a data structure to store the latencies (probably a `map<int, int>` for latency seconds -> count)
* Add `void AddLatency(std::chrono::seconds s)`
* Add `std::chrono::seconds CalculateLatencyPercentile(double n) const`
* Update the logic in `GetMaxExtension()` to calculate the extension based on the Nth percentile latency and `max_extension_`.

As far as testing goes we would:
* Test the `DeadlineExtension` member functions in their own unit test.
* Pass in a `MockDeadlineExtension` into `CreateForTesting`. This would allow us to manually set the return value for `GetMaxExtension()`. Note: we can assume the max extension is calculated correctly, because that would be covered in the unit test


Ultimately, I decided not to implement an intermediate wrapper class, because I think it would unnecessarily obfuscate the code. If there is no desire to use the latency to calculate the ack deadline extension, we will regret having the unneeded abstraction. If there is desire, this comment can serve as an equally good jumping off point to begin the implementation.
___
*[google/cloud/pubsub/internal/subscription_lease_management.cc, line 103 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620#-MaMhMSc3O2byFysnRy-:-MaMhMSc3O2byFysnRy0:by9pirg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a0c0971ad5af7a886dc3eec710f5d9283909ce21/google/cloud/pubsub/internal/subscription_lease_management.cc#L103)):*
> ```cpp
>   std::vector<std::string> ack_ids;
>   ack_ids.reserve(leases_.size());
>   auto extension = max_deadline_extension_;
> ```

could be something like `deadline_extension_.GetMaxExtension()`. see above comment.

coryan (approved):
:lgtm: You should check that `using foo = std::...` is Okay, as opposed to `using foo = ::std::...`.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620#-:-MaYNdJBBbkO_pjYQwG5:b-if8xq9)* status: all files reviewed, 4 unresolved discussions (waiting on @dbolduc)

___
*[google/cloud/pubsub/subscriber_options_test.cc, line 25 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620#-MaYMiVX7tP03YXP-Goj:-MaYMiVX7tP03YXP-Gok:b-nnnz3e) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a0c0971ad5af7a886dc3eec710f5d9283909ce21/google/cloud/pubsub/subscriber_options_test.cc#L25)):*
> ```cpp
> namespace {
> 
> using seconds = std::chrono::seconds;
> ```

I am no expert in the google style guide, can you check what it says about using declarations?
___
*[google/cloud/pubsub/internal/subscription_lease_management.h, line 53 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620#-MaMiGCLETsUTuNjmXUR:-MaYMvDv3uSGe2zdd7MR:bfv5l7h) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a0c0971ad5af7a886dc3eec710f5d9283909ce21/google/cloud/pubsub/internal/subscription_lease_management.h#L53)):*

This could be a `DeadlineExtension` wrapper class.

For now it would only have
* `std::chrono::seconds max_extension_`
* `std::chrono::seconds GetMaxExtension() const` that just returns `max_extension_`

In the future we could
* Add `double latency_percentile_`
* Add a data structure to store the latencies (probably a `map<int, int>` for latency seconds -> count)
* Add `void AddLatency(std::chrono::seconds s)`
* Add `std::chrono::seconds CalculateLatencyPercentile(double n) const`
* Update the logic in `GetMaxExtension()` to calculate the extension based on the Nth percentile latency and `max_extension_`.

As far as testing goes we would:
* Test the `DeadlineExtension` member functions in their own unit test.
* Pass in a `MockDeadlineExtension` into `CreateForTesting`. This would allow us to manually set the return value for `GetMaxExtension()`. Note: we can assume the max extension is calculated correctly, because that would be covered in the unit test


Ultimately, I decided not to implement an intermediate wrapper class, because I think it would unnecessarily obfuscate the code. If there is no desire to use the latency to calculate the ack deadline extension, we will regret having the unneeded abstraction. If there is desire, this comment can serve as an equally good jumping off point to begin the implementation.
</blockquote></details>

If you want to preserve this comment it might be better to create an issue in GitHub.
___
*[google/cloud/pubsub/internal/subscription_lease_management_test.cc, line 106 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620#-MaYNKiB4NLoZXnPoc07:-MaYNKiB4NLoZXnPoc08:b-rmuygk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a0c0971ad5af7a886dc3eec710f5d9283909ce21/google/cloud/pubsub/internal/subscription_lease_management_test.cc#L106)):*
> ```cpp
> 
>   batch_callback(GenerateMessages("0-", 3));
>   ASSERT_THAT(timers, SizeIs(1));
> ```

TIL: `SizeIs()`

dbolduc (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620)* status: all files reviewed, 4 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[google/cloud/pubsub/subscriber_options_test.cc, line 25 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620#-MaYMiVX7tP03YXP-Goj:-MaYaMorBvbbSDdZL7bP:b-gmigul) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a0c0971ad5af7a886dc3eec710f5d9283909ce21/google/cloud/pubsub/subscriber_options_test.cc#L25)):*

I am no expert in the google style guide, can you check what it says about using declarations?
</blockquote></details>

The closest I can find is that it explicitly says not to do: `using ::std::chrono::seconds;` for local convenience.

Also, we have hundreds of instances of `using T = std::something` and no instances of `using T = ::std::something`

So I'm going to leave it as is.
___
*[google/cloud/pubsub/internal/subscription_lease_management.h, line 53 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620#-MaMiGCLETsUTuNjmXUR:-MaYbMXCAV0NEbjMsbiq:bb74x83) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a0c0971ad5af7a886dc3eec710f5d9283909ce21/google/cloud/pubsub/internal/subscription_lease_management.h#L53)):*

If you want to preserve this comment it might be better to create an issue in GitHub.
</blockquote></details>

True

devjgm (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620)* status: all files reviewed, 4 unresolved discussions

___
*[google/cloud/pubsub/subscriber_options_test.cc, line 25 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6620#-MaYMiVX7tP03YXP-Goj:-MaZm6IlBnwPGRgjA5zq:b-1evb6b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a0c0971ad5af7a886dc3eec710f5d9283909ce21/google/cloud/pubsub/subscriber_options_test.cc#L25)):*

The closest I can find is that it explicitly says not to do: `using ::std::chrono::seconds;` for local convenience.

Also, we have hundreds of instances of `using T = std::something` and no instances of `using T = ::std::something`

So I'm going to leave it as is.
</blockquote></details>

ftr, I'd suggest just:

```
using std::chrono::seconds;
```

Issue #6627, opened 2021-05-24
Upgrading Abseil to 20210324 breaks our Makefile quickstart builds


devjgm:
When upgrading https://github.com/abseil/abseil-cpp/archive/20210324.1.tar.gz our Makefile quickstart builds fail with a bunch of errors like:
```
2021-05-24T19:57:17Z: [ Make ]
make: Entering directory '/workspace/google/cloud/bigtable/quickstart'
g++ -std=c++11 -I/h/google-cloud-cpp-installed/include -I/usr/local/include -L/h/google-cloud-cpp-installed/lib -L/usr/local/lib -o /workspace/cmake-out/quickstart-makefile-bigtable/quickstart quickstart.cc -lgoogle_cloud_cpp_bigtable -lgoogle_cloud_cpp_grpc_utils -labsl_time.2103.0.1 -labsl_civil_time.2103.0.1 -labsl_strings.2103.0.1 -labsl_strings_internal.2103.0.1 -labsl_base.2103.0.1 -labsl_spinlock_wait.2103.0.1 [...]
...
/usr/bin/ld: cannot find -labsl_time.2103.0.1
/usr/bin/ld: cannot find -labsl_civil_time.2103.0.1
/usr/bin/ld: cannot find -labsl_strings.2103.0.1
...
```

Those link lines are coming from the `.pc` files that we installed, such as this one for `google_cloud_cpp_common`
```
...
Name: Google Cloud C++ Client Library Common Components
Description: Common Components used by the Google Cloud C++ Client Libraries.
Requires:
Version: 1.28.0

[...]

PR #6628, opened 2021-05-24
chore(deps): upgrade abseil to 20210324.1

devjgm:
Fixes: #6627

This version of Abseil adds proper pkgconfig support, so we no longer
need dynamically compute the Abseil `.so` files to link with, instead we
can just list the pkg-config deps that we "require" from Abseil. Indeed,
our old way of computing the libraries no longer works, so this change
was required in the same PR as the Abseil upgrade.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6628)


devbww:
```
ld: illegal thread local variable reference to regular symbol __ZN4absl12lts_2021032413base_internal19thread_identity_ptrE for architecture x86_64
```
from `kokoro/macos/bazel`.

ci/generate-markdown/generate-packaging.sh
```diff
@@ -199,7 +199,7 @@ cc_binary(
 
 | Library | Minimum version | Description |
 | ------- | --------------: | ----------- |
-| [Abseil][abseil-gh] | 20200923, Patch 3 | Abseil C++ common library |
```
coryan, line 202: We have at least one build that tests with 20200923.  I think we should not drop support for the older version, if we do, we likely have to drop support for older gRPC versions too.

devjgm: The issue is that this version of Abseil renames their old `.so` files and adds proper `pkg-config` support. So ideally we should start using their pkg-config files, as I've done in this version of the PR, but doing that makes it so our `Makefile` builds will no longer work w/ previous releases of Abseil.

We could also not using Abseil's `pkg-config` files that are added in this new release. But doing that would require that we change our CMake function to understand Abseil's new `.so` naming.

coryan: I think I follow how we end up here.

I am trying to think about this from the perspective of customers reading this document, maybe for the first time.  They read `Abseil >= 20210324` and also read `gRPC >= 1.35` which supposedly works with (requires?) `Abseil == 20200923`.  That seems complicated.  Which one is it?

Anyway, feel free to ignore this.  The thing that @devbww points out seems more important.

devjgm:
> ```
> ld: illegal thread local variable reference to regular symbol __ZN4absl12lts_2021032413base_internal19thread_identity_ptrE for architecture x86_64
> ```
> 
> from `kokoro/macos/bazel`.

It looks like an Abseil issue. I filed https://github.com/abseil/abseil-cpp/issues/965

devjgm:
Closing this PR for now. Abseil is working up a fix for `ld: illegal thread local variable reference to regular symbol __ZN4absl12lts_2021032413base_internal19thread_identity_ptrE for architecture x86_64` and they'll cut another patch release.

PR #6631, opened 2021-05-25
refactor(pubsub): remove timer factory

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6631)

### DIFF
--- a/google/cloud/pubsub/internal/subscription_session.cc
+++ b/google/cloud/pubsub/internal/subscription_session.cc
@@ -203,16 +203,9 @@ future<Status> CreateTestingSubscriptionSession(
       "test-client-id", options, std::move(retry_policy),
       std::move(backoff_policy));
 
-  auto cq = executor;  // need a copy to make it mutable
-  auto timer = [cq](std::chrono::system_clock::time_point) mutable {
-    return cq.MakeRelativeTimer(std::chrono::milliseconds(50))
-        .then([](future<StatusOr<std::chrono::system_clock::time_point>> f) {
-          return f.get().status();
-        });
-  };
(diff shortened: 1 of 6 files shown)


google/cloud/pubsub/internal/subscription_session.cc
```diff
@@ -203,16 +203,9 @@ future<Status> CreateTestingSubscriptionSession(
...
-  auto timer = [cq](std::chrono::system_clock::time_point) mutable {
-    return cq.MakeRelativeTimer(std::chrono::milliseconds(50))
-        .then([](future<StatusOr<std::chrono::system_clock::time_point>> f) {
-          return f.get().status();
-        });
-  };
-  auto lease_management = SubscriptionLeaseManagement::CreateForTesting(
-      executor, shutdown_manager, timer, std::move(batch),
-      options.max_deadline_time(), options.max_deadline_extension());
+  auto lease_management = SubscriptionLeaseManagement::Create(
```
coryan, line 206: I wonder if the enclosing function (`CreateTestingSubscriptionSession()`) is even needed now?

coryan (approved):

google/cloud/pubsub/internal/subscription_session.cc
dbolduc: I think you are right. I have moved the method, which exists only for testing, into the test file.

PTAL

coryan (approved):

google/cloud/pubsub/internal/subscription_session.cc
dbolduc: I think I messed up my `std::move()`'s

PR #6640, opened 2021-05-27
feat(storage): recommend `g::c::Options`

coryan:
With this PR the constructors for `google::cloud::storage::Client` using
`google::cloud::Options` become public. The documentation, examples, and
tests change to recommend using these constructors.

Part of the work for #6161

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6640)


devjgm (approved):
Net reduction of about 100 lines. Nice.

PR #6651, opened 2021-05-30
cleanup(storage): simplify gRPC plugin initialization

coryan:
Now that the plugin is initialized with the unified credentials it can
never "fail" (it can return a client that always fails). Likewise, the
credentials are easier to initialize, setting the unified credentials
works for both `internal::CurlClient` and `internal::GrpcClient`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6651)


dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6651)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #6658, opened 2021-05-31
ci: use typos-cli instead of cspell

devjgm:
Typos (https://crates.io/crates/typos) is a modern source-code spell
checker that is crazy fast and seems to work very well out of the box.
By default it checks all of our code, respecting our `.gitignore`,
including source code itself (not just comments). On my laptop `cspell`
took ~30 seconds and `typos` takes ~0.2 seconds, which is convenient
because most of us run this build frequently to fix formatting, etc.

It also supports the ability to write the changes back to disk
(`--write-changes`), which we could consider as we gain confidence in
it. I used this mode to fix all the typos in this PR, but it had one
case of bad changes where it changed `Expection` -> `Exception`, whereas
in the particular context we wanted `Expectation`. I tweaked this by
hand, and decided that simply reporting the misspellings rather than
fixing them is probably safer, and it matches what we did w/ cspell.

**NOTE**: This PR is somewhat like a proposal or a "consider". We certainly
could decide that sticking w/ `cspell` is better. Though I *think* `typos` seems
better in all the ways I've considered and I think it's worth trying. It would be
easy to go back to cspell if we ever want to.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6658)


remyabel:
+1 This looks way superior. They use a hand-rolled parser rather than regex so naturally it's going to be faster and better at catching context-aware typos.

dbolduc (approved):
> Reviewed 24 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6658)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devjgm)

PR #6660, opened 2021-06-01
ci: run sed checkers in parallel

devjgm:
This PR speeds up the `Running Abseil header fixes`, `Running include fixes`, and `Running whitespace fixes` steps by making each step work in parallel with `xargs -P`. This cuts the overall `checkers-pr` time in half on my laptop. For the mentioned three steps they drop from about 15 seconds to about 3 seconds.

Speed improvements to this build are valuable because this is a build that most of us run frequently, and certainly before sending a PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6660)

### DIFF
--- a/ci/cloudbuild/builds/checkers.sh
+++ b/ci/cloudbuild/builds/checkers.sh
@@ -115,7 +115,7 @@ time {
 printf "%-30s" "Running Abseil header fixes:" >&2
 time {
   expressions=("-e" "'s;#include \"absl/strings/str_\(cat\|replace\|join\).h\";#include \"google/cloud/internal/absl_str_\1_quiet.h\";'")
-  expressions+=("-e" "'s;#include \"absl/container/\\(flat_hash_map\\).h\";#include \"google/cloud/internal/absl_\\1_quiet.h\";'")
+  expressions+=("-e" "'s;#include \"absl/container/\(flat_hash_map\).h\";#include \"google/cloud/internal/absl_\1_quiet.h\";'")
   git ls-files -z |
     grep -zv 'google/cloud/internal/absl_.*quiet.h$' |
     grep -zE '\.(h|cc)$' |
--- a/ci/cloudbuild/builds/checkers.sh
+++ b/ci/cloudbuild/builds/checkers.sh
@@ -20,22 +20,27 @@ source "$(dirname "$0")/../../lib/init.sh"
 source module ci/lib/io.sh
 source module ci/cloudbuild/builds/lib/git.sh
 
-# Replaces a file only if it changed. This is needed because sed -i and perl -i
-# will modify the mtime even if no edits are made, which will cause code to be
-# recompiled unnecessarily.
-function replace_original_if_changed() {
-  if [[ $# != 2 ]]; then
-    return 1
-  fi
-  local original="$1"
-  local reformatted="$2"
-  if cmp -s "${original}" "${reformatted}"; then
-    rm -f "${reformatted}"
-  else
-    chmod --reference="${original}" "${reformatted}"
-    mv -f "${reformatted}" "${original}"
-  fi
+# Runs a single sed expression over the given files, editing them in place, and
+# updating the file's mtime only if edits were made. Tis function is exported
+# so it can be run in subshells, such as with xargs -P. Example:
+#
+#   sed_edit 's/foo/bar/g' hello.txt
+#
+function sed_edit() {
+  local expression="$1"
+  shift
+  for file in "$@"; do
+    local tmp="${file}.tmp"
+    cp -p "${file}" "${tmp}"
+    sed -i -e "${expression}" "${tmp}"
+    if cmp -s "${file}" "${tmp}"; then
+      rm -f "${tmp}"
+    else
+      mv -f "${tmp}" "${file}"
+    fi
+  done
 }
+export -f sed_edit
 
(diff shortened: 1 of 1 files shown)


devjgm:
**BEFORE**
```
Running typos:                ... 0.227 seconds
Running markdown generators:  ... 0.144 seconds
Running check-include-guards: ... 0.043 seconds
Running cmake-format:         ... 2.961 seconds
Running Abseil header fixes:  ... 4.671 seconds
Running clang-format:         ... 1.208 seconds
Running buildifier:           ... 0.041 seconds
Running black:                ... 0.672 seconds
Running shfmt:                ... 0.018 seconds
Running shellcheck:           ... 1.587 seconds
Running include fixes:        ... 4.688 seconds
Running whitespace fixes:     ... 6.336 seconds
==> 🕑 checkers completed in 22.769 seconds
```

**AFTER**
```
Running typos:                ... 0.205 seconds
Running markdown generators:  ... 0.143 seconds
Running check-include-guards: ... 0.040 seconds
Running cmake-format:         ... 2.859 seconds
Running Abseil header fixes:  ... 0.284 seconds
Running clang-format:         ... 1.168 seconds
Running buildifier:           ... 0.042 seconds
Running black:                ... 0.643 seconds
Running shfmt:                ... 0.017 seconds
Running shellcheck:           ... 1.479 seconds
Running include fixes:        ... 0.282 seconds
Running whitespace fixes:     ... 0.456 seconds
==> 🕑 checkers completed in 7.717 seconds
```

ci/cloudbuild/builds/checkers.sh
```diff
@@ -20,22 +20,39 @@ source "$(dirname "$0")/../../lib/init.sh"
...
+#   sed_edit -e 's/foo/bar/g' -e 's,baz,blah,' hello.txt
+#
+function sed_edit() {
+  local expressions=()
+  local files=()
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+    -e)
+      expressions+=("-e" "$2")
+      shift 2
```
devbww, line 36: Nit: There may only be one argument.

ci/cloudbuild/builds/checkers.sh
```diff
@@ -97,15 +114,12 @@ time {
 # includes.
 printf "%-30s" "Running Abseil header fixes:" >&2
 time {
+  expressions=("-e" "'s;#include \"absl/strings/str_\(cat\|replace\|join\).h\";#include \"google/cloud/internal/absl_str_\1_quiet.h\";'")
```
devbww, line 117: Nit: The double quoting is a pain.

ci/cloudbuild/builds/checkers.sh
```diff
@@ -97,15 +114,12 @@ time {
...
   git ls-files -z |
     grep -zv 'google/cloud/internal/absl_.*quiet.h$' |
     grep -zE '\.(h|cc)$' |
-    while IFS= read -r -d $'\0' file; do
-      sed -e 's;#include "absl/strings/str_\(cat\|replace\|join\).h";#include "google/cloud/internal/absl_str_\1_quiet.h";' \
-        -e 's;#include "absl/container/\(flat_hash_map\).h";#include "google/cloud/internal/absl_\1_quiet.h";' \
-        "${file}" >"${file}.tmp"
-      replace_original_if_changed "${file}" "${file}.tmp"
-    done
+    xargs -P "$(nproc)" -n 50 -0 bash -c "sed_edit ${expressions[*]} \$0 \$@"
```
devbww, line 122: Q: Why do we want to limit the number of processes to the number of processing units?  It seems that most of the time they'll be waiting for I/O.  [I have a recollection of asking this before on a similar usage, but I'm afraid I don't recall the response.]

devbww, line 122: It looks like `$0` will not be quoted there.

devbww (approved):

ci/cloudbuild/builds/checkers.sh
devjgm: Agreed. Not sure how to make that better. Open to suggestions.

ci/cloudbuild/builds/checkers.sh
devjgm: testing `$2` first now

ci/cloudbuild/builds/checkers.sh
devjgm: I believe we previously talked about why `-n 50`. I don't recall talking about `-P $(nproc)`, though we certainly could have and I forgot.

So why in this PR, because that what all the other xargs do in this file. Why do _they_ choose nproc? Not sure. Seems to be a reasonable choice unless we have some better intuition or data. I'd be happy to try a different number in a follow up PR to see if that makes a difference.

devjgm: Quoting now.

devbww: It seems to me that they're related.  If you want to run things in parallel (`-P`), you should specify the batch size (`-n`).  And the best batch size seems to be related to the relative cost of processing one file -v- the cost of starting a new process ... too low and the process management dominates, and too high and parallelism is restricted.  But after picking a batch size, you might as well schedule as many batches as you can (`-P 0`), assuming scheduling is sub-linear.

PR #6664, opened 2021-06-01
chore(deps): upgrade abseil to 20210324.2

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/6627

This version of Abseil adds proper pkgconfig support, so we no longer
need dynamically compute the Abseil `.so` files to link with, instead we
can just list the pkg-config deps that we "require" from Abseil. Indeed,
our old way of computing the libraries no longer works, so this change
was required in the same PR as the Abseil upgrade.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6664)


ci/generate-markdown/generate-packaging.sh
```diff
@@ -200,7 +200,7 @@ cc_binary(
 
 | Library | Minimum version | Description |
 | ------- | --------------: | ----------- |
-| [Abseil][abseil-gh] | 20200923, Patch 3 | Abseil C++ common library |
+| [Abseil][abseil-gh] | 20210324, Patch 2 | Abseil C++ common library |
```
coryan, line 203: I am not sure this is accurate, we are still building with 20200923.  I know it would be messy to describe what does not work with that version, so this is your call.

coryan (approved):
Just our of curiosity, what changed in the API/ABI dumps?

devjgm:
> Just our of curiosity, what changed in the API/ABI dumps?

Abseil's inline namespace changed, so uses of things like `absl::optional` have a different name in our compiled artifacts. So I imagine we'll need to update these ABI dumps everytime we update Abseil

ci/generate-markdown/generate-packaging.sh
devjgm: Where would this version still be building with 20200923? The issue is that Abseil's 20210324 LTS is the first release that includes real `pkg-config` files, so if we change our `pkg-config` files to use theirs, then ours will not work w/ previous versions of Abseil that don't have pkg-config files. Conversely, if we don't change to use their new pkg-config files, then our code may not work w/ newer versions of Abseil.

Note that the issue is only `Makefile` builds that happen to use `pkg-config`. Our cmake and bazel builds are fine building against older Abseil.

Thoughts about how we should proceed?

coryan: > Where would this version still be building with 20200923?

This build:

https://github.com/googleapis/google-cloud-cpp/blob/main/ci/cloudbuild/builds/cmake-oldest-deps.sh

Uses this file:

https://github.com/googleapis/google-cloud-cpp/blob/main/ci/etc/oldest-deps/vcpkg.json

Which points to:

https://github.com/microsoft/vcpkg/blob/6d9ed568117dd958c543b3ab8d3ed692965cac34/versions/baseline.json#L11-L14


> The issue is that Abseil's 20210324 LTS is the first release that includes real `pkg-config` files, so if we change our `pkg-config` files to use theirs, then ours will not work w/ previous versions of Abseil that don't have pkg-config files. Conversely, if we don't change to use their new pkg-config files, then our code may not work w/ newer versions of Abseil.

Yes, I think we discussed that already.

> Thoughts about how we should proceed?

I am not sure how to fit this in a simple table: we should leave the minimum version at `20200923` with a footnote saying that you need `20210324.2` to use `pkg-config`.

devjgm: Good idea. Done. PTAL

coryan (approved):
LGTM

coryan:
Remember to fix #6627 as part of this PR.

devjgm:
> Remember to fix #6627 as part of this PR.

Thanks for the reminder. I would have forgotten. Added to the PR description.

PR #6672, opened 2021-06-02
chore(deps): renovate-bot also updates packaging.md

### LINKED ISSUE #6657: add post update task to renovate-bot so that it regenerates packaging.md
We have [PR's from renovate-bot](https://github.com/googleapis/google-cloud-cpp/pull/6656) that dont pass `checkers-pr`. This is because the `packaging.md` file is not touched by renovate-bot, but it is regenerated by `checkers-pr`. If a dependency in `packaging.md` is changed, the build breaks.

It looks like renovate-bot has [postUpgradeTasks](https://docs.renovatebot.com/configuration-options/#postupgradetasks) which could be promising? maybe we can use this to run `ci/generate-markdown/generate-packaging.sh > doc/packaging.md`?

dbolduc:
Fixes #6657 

(we cant run a postUpgradeTask, but having renovate update `packaging.md` will probably catch all of the breaks due to checkers-pr)

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6672)

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -6,7 +6,10 @@
   ],
   "regexManagers": [
     {
-      "fileMatch":  [".*\\.Dockerfile$"],
+      "fileMatch":  [
+        ".*\\.Dockerfile$",
+        ".*packaging.md$"
+      ],
       "matchStrings": [
         "https://github\\.com/(?<depName>.*?)/archive/(?<currentValue>.*?)\\.tar\\.gz"
       ],
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -8,7 +8,7 @@
     {
       "fileMatch":  [
         ".*\\.Dockerfile$",
-        ".*packaging.md$"
+        ".*doc/packaging\\.md$"
       ],
       "matchStrings": [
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -6,7 +6,10 @@
   ],
   "regexManagers": [
     {
-      "fileMatch":  [".*\\.Dockerfile$"],
+      "fileMatch":  [
+        ".*\\.Dockerfile$",
+        ".*packaging.md$"
```
coryan, line 11: Probably won't change anything, but shouldn't we escape the `.`, as in: `"packaging\\.md$"`.  And do you need to match `doc/foo-packaging.md` also?

dbolduc: Done.

coryan (approved):

.github/renovate.json
```diff
@@ -6,7 +6,10 @@
   ],
   "regexManagers": [
     {
-      "fileMatch":  [".*\\.Dockerfile$"],
+      "fileMatch":  [
+        ".*\\.Dockerfile$",
+        ".*doc/packaging\\.md$"
```
devbww, line 11: For what it's worth, if the regexp's aren't anchored (and the trailing `$` implies they are not, but I don't know), then the leading `.*` is useless.

Issue #6687, opened 2021-06-03
Upgrade our vcpkg to a version that supports Apple M1


devjgm:
We're using the following vcpkg version (2021.04.30):

```
cloudbuild/builds/quickstart-cmake.sh
32:curl -sSL "https://github.com/microsoft/vcpkg/archive/2021.04.30.tar.gz" |

kokoro/macos/build-quickstart-cmake.sh
31:curl -sSL "https://github.com/microsoft/vcpkg/archive/2021.04.30.tar.gz" |

kokoro/windows/build-cmake-dependencies.ps1
51:$vcpkg_version = "2021.04.30"
```

This version doesn't understand the arm64 arch. 

There's now https://github.com/microsoft/vcpkg/releases/tag/2021.05.12, which I suspect will work.

PR #6688, opened 2021-06-03
ci: upgrade to vcpkg 2021.05.12 which understands Apple M1

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/6687

This version of vcpkg sees the `vcpkg.json` file at the top of our repo
and thinks we're using it in manifest mode, so we need to run the vcpkg
commands from w/in the vcpkg dir.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6688)

### DIFF
--- a/ci/kokoro/windows/build-cmake-dependencies.ps1
+++ b/ci/kokoro/windows/build-cmake-dependencies.ps1
@@ -126,7 +126,7 @@ if ($RunningCI -and $HasBuildCache) {
 
 # Integrate installed packages into the build environment.
 Push-Location "${vcpkg_dir}"
-&"vcpkg.exe" integrate install
+&".\vcpkg.exe" integrate install
 if ($LastExitCode) {
     Write-Host -ForegroundColor Red "vcpkg integrate failed with exit code $LastExitCode"
     Exit ${LastExitCode}
@@ -136,7 +136,7 @@ Pop-Location
 # Remove old versions of the packages.
 Write-Host -ForegroundColor Yellow "`n$(Get-Date -Format o) Cleanup outdated vcpkg packages."
 Push-Location "${vcpkg_dir}"
-&"vcpkg.exe" remove ${vcpkg_flags} --outdated --recurse
+&".\vcpkg.exe" remove ${vcpkg_flags} --outdated --recurse
 if ($LastExitCode) {
     Write-Host -ForegroundColor Red "vcpkg remove --outdated failed with exit code $LastExitCode"
     Exit ${LastExitCode}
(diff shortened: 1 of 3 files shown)


coryan (approved):

ci/kokoro/windows/build-cmake-dependencies.ps1
```diff
@@ -48,7 +48,7 @@ if ($args.count -ge 1) {
     $vcpkg_flags=("--triplet", "${env:VCPKG_TRIPLET}")
 }
 $vcpkg_dir = "cmake-out\${vcpkg_base}"
-$vcpkg_version = "2021.04.30"
+$vcpkg_version = "2021.05.12"
```
coryan, line 51: Ugh, I did not notice the build failures, looks like you need a call to `Set-Location ${vcpkg_dir}` around line 134.

devjgm:
The windows builds still break because they think `vcpkg` is running in manifest mode. I imagine this is because the vcpkg dir is in `cmake-out/vcpkg`, which means the project's `vcpkg.json` file is found in a parent directory of that.

To fix this we may be able to download vcpkg some other dir. Or we may be able to tell vcpkg to NOT use manifest mode.

devjgm:
> This is just a guess, but maybe `Push-Location` does not do what one would think it does:
> 
> https://stackoverflow.com/questions/32734408/when-to-use-push-location-versus-set-location
> https://stackoverflow.com/questions/11246068/why-dont-net-objects-in-powershell-use-the-current-directory
> https://stackoverflow.com/questions/44162443/current-directory-apparently-not-current-directory/56462392#56462392

I know *VERY* little about this/windows/powershell, but it appears like `Set-Location` behaves the same way.

```console
PS C:\Windows\system32> Write-Host ([Environment]::CurrentDirectory)
C:\Windows\system32
PS C:\Windows\system32> Set-Location c:\
PS C:\> Write-Host ([Environment]::CurrentDirectory)
C:\Windows\system32
PS C:\> Get-Location

Path
----
C:\
```

ci/kokoro/windows/build-cmake-dependencies.ps1
devjgm: After trying things like `Push-Location` (which I think behaves the same as `Set-Location`), I couldn't get it to work. So I disabled manifest mode here with `$Env:VCPKG_FEATURE_FLAGS = "-manifests"`

coryan: SGTM.

coryan (approved):

PR #6694, opened 2021-06-04
ci(macos): cleanup logging output

devjgm:
There should be no semantic changes in the builds; just logging
additions/changes. Overview:

* The output now starts with a "Machine Info" section, similar to our
GCB builds
* We no longer echo `====` lines, instead we use `io::log_h1` (or h2)
* Reordered a few sections

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6694)


ci/kokoro/macos/build-bazel.sh
```diff
@@ -19,9 +19,7 @@ source "$(dirname "$0")/../../lib/init.sh"
 source module /ci/etc/integration-tests-config.sh
 source module /ci/lib/io.sh
 
-echo
-echo "================================================================"
-io::log_yellow "update or install Bazel."
+io::log_h2 "update or install Bazel."
```
devbww, line 22: I guess I'd remove all the trailing periods from things that are now in boxes.

ci/kokoro/macos/build.sh
```diff
@@ -99,11 +76,47 @@ else
   exit 1
 fi
 
+cd "${PROJECT_ROOT}"
+NCPU="$(sysctl -n hw.logicalcpu)"
+readonly NCPU
+
+function google_time() {
+  curl -sI google.com | grep "^Date:" | cut -f2- -d: | tr -d '\r' | xargs
```
devbww, line 84: This seems overly convoluted, down to the `xargs` whose purpose seems to be to remove the leading space.

How about just ...
```
  curl -sI google.com | sed -n 's/Date: \(.*\)\r/\1/p'
```

Also, should we update the `google_time()` in `ci/cloudbuild/build.sh` to also be wary of the CR?

devbww (approved):

ci/kokoro/macos/build-bazel.sh
devjgm: done

ci/kokoro/macos/build.sh
devjgm: done

devbww (approved):

PR #6698, opened 2021-06-04
ci(macos): install ccache when needed

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6698)

### DIFF
--- a/ci/kokoro/macos/build-quickstart-cmake.sh
+++ b/ci/kokoro/macos/build-quickstart-cmake.sh
@@ -22,6 +22,9 @@ source module /ci/etc/quickstart-config.sh
 
 io::log_h2 "Update or install dependencies"
 
+# vcpkg needs this
+brew install pkg-config || pkg-config --version
+
 # Fetch vcpkg at the specified hash.
 vcpkg_dir="${HOME}/vcpkg-quickstart"
 mkdir -p "${vcpkg_dir}"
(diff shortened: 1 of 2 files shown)


devbww (approved):

devjgm:
Added install of `pkg-config` also, which vcpkg needs during the `quickstart-cmake` build

ci/kokoro/macos/build-quickstart-cmake.sh
```diff
@@ -22,6 +22,9 @@ source module /ci/etc/quickstart-config.sh
 
 io::log_h2 "Update or install dependencies"
 
+# vcpkg needs this
+brew install pkg-config || pkg-config --version
```
devbww, line 26: I'm not sure what these idioms mean.  If we fail to install `pkg-config`, we print its version?

devjgm: You're right that this is a bit strange. I'm mostly just following the established pattern that was already here. I _think_ it's written like this because `brew install <foo>` commonly fails if foo is already installed but a newer version is available. It fails because we're told `brew` not to update packages (with `HOMEBREW_NO_AUTO_UPDATE=1`). But in the case of this failure, we want to proceed so we execute the command (that we think is installed) and prints its version. It's similar to `|| true`, except that it will fail the the `brew install` fails for some other reason and the command actually does not exist.

Anyway, I think we may want to revisit the need for any of these brew commands now that we're running on more modern kokoro mac machines. We may be able to get rid of all of them (or many).

Issue #6699, opened 2021-06-05
cmake-oldest-deps build is broken


devjgm:
GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/0684b9b8-3891-44b1-82f6-527a15409384;tab=detail?project=cloud-cpp-testing-resources
Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/4032b07e9f3556d4d29d17ee2bb214d5d136292c/ubuntu-focal-cmake-oldest-deps/log-0684b9b8-3891-44b1-82f6-527a15409384.txt

This error is reproducible locally with:

```console
$ ./ci/cloudbuild/build.sh -t cmake-oldest-deps-pr --docker
...
-------------------
|   Configuring   |
-------------------
-- Running vcpkg install
-- Running vcpkg install - failed
CMake Error at cmake-out/vcpkg/scripts/buildsystems/vcpkg.cmake:815 (message):
  vcpkg install failed.  See logs for more information:
  /workspace/cmake-out/build/vcpkg-manifest-install.log
Call Stack (most recent call first):
  /usr/share/cmake-3.16/Modules/CMakeDetermineSystem.cmake:93 (include)
  CMakeLists.txt:26 (project)


CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
...
```

The error is from vcpkg. The cmake-oldest-deps build clones vcpkg at head https://github.com/googleapis/google-cloud-cpp/blob/4032b07e9f3556d4d29d17ee2bb214d5d136292c/ci/cloudbuild/builds/cmake-oldest-deps.sh#L33
so the problem could be some recent commit to vcpkg that happened today.

Here are commits from today (June 4): https://github.com/microsoft/vcpkg/commits/master

devjgm:
Here's vcpkg's manifest install log

```console
$ cat /workspace/cmake-out/build/vcpkg-manifest-install.log
Detecting compiler hash for triplet x64-linux...
The following packages will be built and installed:
[...]

PR #6700, opened 2021-06-05
fix(cmake-oldest-deps): use vcpkg release tarball rather than clone

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/6699

This PR changes the `cmake-oldest-deps` build to use a release tarball
of vcpkg, the same as our other builds that use vcpkg, instead of
cloning the whole vcpkg repo, which picks up new changes from vcpkg that
can break our build unexpectedly.

This change appears to work in my local tests, but I'm concerned that
it's wrong because the previous code had a comment saying that cloning
the repo was required. So I suppose this PR is just for discussion to
see if this is the right direction or if we should fix the issue some
other way.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6700)

### DIFF
--- a/ci/cloudbuild/builds/cmake-oldest-deps.sh
+++ b/ci/cloudbuild/builds/cmake-oldest-deps.sh
@@ -23,18 +23,13 @@ source module ci/cloudbuild/builds/lib/integration.sh
 export CC=clang
 export CXX=clang++
 
+io::log_h2 "Installing vcpkg"
 vcpkg_root="${PROJECT_ROOT}/cmake-out/vcpkg"
-if [[ ! -d "${vcpkg_root}" ]]; then
-  mkdir -p "${vcpkg_root}"
-  # To enable versioning we need to clone the vcpkg history. It seems that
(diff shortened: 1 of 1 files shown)


devjgm:
Nope, I was wrong. I thought it was working locally, but that was due to some vcpkg cache I had locally. After a full clean build it failed w/ this PR due to:

```
Error: Couldn't find explicitly specified baseline `"6d9ed568117dd958c543b3ab8d3ed692965cac34"` in the baseline file, and there was no baseline at that commit or the commit didn't exist.
Error: while checking out baseline '6d9ed568117dd958c543b3ab8d3ed692965cac34:versions/baseline.json':
fatal: not a git repository: '/workspace/cmake-out/vcpkg/.git'

This may be fixed by updating vcpkg to the latest master via `git pull`.
Failed to determine the current commit:
fatal: not a git repository: '/workspace/cmake-out/vcpkg/.git'
```

PR #6711, opened 2021-06-07
cleanup(bigtable)!: remove admin async integration tests

### LINKED ISSUE #5923: Remove the async functions from the Bigtable admin APIs

dbolduc:
Fixes #5923

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6711)


coryan:
Can we remove the untested functions at the same time?

google/cloud/bigtable/instance_admin.cc
```diff
@@ -84,63 +84,15 @@ StatusOr<InstanceList> InstanceAdmin::ListInstances() {
...
-                  std::back_inserter(res.failed_locations));
-        return res;
-      });
-}
-
 future<StatusOr<btadmin::Instance>> InstanceAdmin::CreateInstance(
     InstanceConfig instance_config) {
   auto cq = background_threads_->cq();
-  return AsyncCreateInstance(cq, std::move(instance_config));
+  return AsyncCreateInstanceImpl(cq, std::move(instance_config));
```
dbolduc, line 90: We could move `AsyncCreateInstanceImpl` into `CreateInstance` and just have one method.
@coryan Do you think that would be cleaner?

coryan: Let's just do the remove in this PR and we can cleanup further in future PRs.

coryan (changes requested):
The changes look good.  I think you should update `CHANGELOG.md` as part of this PR. Easier to describe the changes now that in a month when we create a new release.

I changed the PR description to `cleanup(bigtable)!: ` because it is a breaking change.

CHANGELOG.md
```diff
@@ -79,6 +65,12 @@
 
 ## v1.29.0 - TBD
 
+### Bigtable:
+
+**BREAKING CHANGES**:
+* Experimental functions matching `Async*` in `bigtable::TableAdmin` and
+  `bigtable::InstanceAdmin` have been removed from the public API.
```
coryan, line 72: Consider linking #5923 from here.

coryan (approved):

PR #6716, opened 2021-06-08
ci: add a seconds delta to the "io" timestamps

devbww:
Append "(+\<delta\>s)" to the `io::internal::timestamp()` to more
clearly indicate the duration since the beginning of the build.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6716)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -250,6 +250,7 @@ if [[ "${DOCKER_FLAG}" = "true" ]]; then
     "--env=USER=$(id -un)"
     "--env=TERM=$([[ -t 0 ]] && echo "${TERM:-dumb}" || echo dumb)"
     "--env=TZ=UTC0"
+    "--env=CI_LIB_IO_FIRST_TIMESTAMP=${CI_LIB_IO_FIRST_TIMESTAMP:-}"
     "--env=CODECOV_TOKEN=${CODECOV_TOKEN:-}"
     "--env=BRANCH_NAME=${BRANCH_NAME}"
     "--env=COMMIT_SHA=${COMMIT_SHA}"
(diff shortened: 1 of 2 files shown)


ci/cloudbuild/build.sh
```diff
@@ -250,6 +250,7 @@ if [[ "${DOCKER_FLAG}" = "true" ]]; then
     "--env=USER=$(id -un)"
     "--env=TERM=$([[ -t 0 ]] && echo "${TERM:-dumb}" || echo dumb)"
     "--env=TZ=UTC0"
+    "--env=CI_LIB_IO_FIRST_TIMESTAMP=${CI_LIB_IO_FIRST_TIMESTAMP:-}"
```
devjgm, line 253: Is this just to enable manual testing of the code, or is it needed for something in the normal functioning of this feature?

devjgm (approved):
What an interesting coincidence: Just this past Friday I started a branch to add exactly this feature, but I then got sidetracked by other stuff and didn't finish it. 

Good idea! Thanks!

ci/cloudbuild/build.sh
devbww: I believe it is needed to get the original "first timestamp" through the `docker run` from one `io.sh` environment to a child.

devjgm: I think you're right, that _would_ be needed, but I don't believe any of our builds do that. None of our builds re-invoke `ci/cloudbuild/build.sh`. If that's all it's for, I suspect we don't need to pass along that environment variable.

devbww: It happens for me all the time.  Say ...
```
ci/cloudbuild/build.sh --docker --trigger=asan-pr
  docker run ... --env=...  ... ci/cloudbuild/build.sh --local asan
    ci/cloudbuild/build.sh --local asan
```

devjgm: I see what you want now. I was thinking of something slightly different. But now that I understand, I think what you have here looks great. Thanks. LGTM.

PR #6722, opened 2021-06-08
cleanup(ci): use bash colon builtin to set vars with defaults

devjgm:
Uses the bash `:` builtin and `:=` in the parameter expansion to set
some bash variables with default values in a more succinct way.

See:
- https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Bourne-Shell-Builtins
- https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6722)


coryan (approved):
TIL

PR #6730, opened 2021-06-09
feat(generator): support unified credentials

### LINKED ISSUE #6721: Support unified credentials in generated libraries
We need to support unified credentials in the generated libraries.  This will happen in several steps:

- [x] Change the generated `*Connection` class to hold an instance of `google::cloud::BackgroundThreads`, this is needed because the gRPC-based unified credentials may perform asynchronous requests to refresh any access tokens.
- [x] Change the generated factory functions for `*Stub` to consume a `google::cloud::CompletionQueue`, this would be the completion queue associated with any background threads, see the previous point.
- [x] Change the generator to create a `*Auth` decorator similar to `google::cloud::storage::internal::StorageAuth`
- [x] Change the generator to use the `*Auth` decorator in the `*Connection` factory function.

coryan:
The generator creates a `*Stub` decorator to inject the necessary
credentials in the `grpc::ClientContext`. This decorator is optional, as
many credential types simply initialize the `grpc::Channel` and never
change the policies after this point.

Fixes #6721

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6730)


coryan:
I think this is too large, I am going to break it down into smaller PRs.

PR #6744, opened 2021-06-10
feat(generator): generators for auth decorator

coryan:
This PR just introduces code to generate the `*Auth` decorators. These
are not used by default because doing so would either (1) break the
build, or (2) require to update all the golden files and libraries in a
huge PR.

Part of the work for #6721.
If you are interested in the generated code from this (none of it is included with this PR), see #6730.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6744)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6744)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #6753, opened 2021-06-11
cleanup(internal): use ssl_options_

remyabel:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6753)


coryan:
Thanks!

remyabel:
Hmm, it is crashing in two different tests. Guessing this is going to require a bit of debugging.

PR #6758, opened 2021-06-13
refactor: use helper to join testing::TempDir()

coryan:
With googletest >= 1.11 ::testing::tempDir() does not seem to end in a
forward slash. That is a problem because in many places we did
`::testing::TempDir() + path` to create a filepath, which can be invalid
or confuse things.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6758)

### DIFF
--- a/generator/integration_tests/generator_integration_test.cc
+++ b/generator/integration_tests/generator_integration_test.cc
@@ -91,6 +91,9 @@ class GeneratorIntegrationTest
     output_path_ = google::cloud::internal::GetEnv(
                        "GOOGLE_CLOUD_CPP_GENERATOR_OUTPUT_PATH")
                        .value_or(::testing::TempDir());
+    if (output_path_.empty() || output_path_.back() != '/') {
+      output_path_ += '/';
+    }
 
     google::cloud::generator::Generator generator;
     google::protobuf::compiler::CommandLineInterface cli;
--- a/generator/integration_tests/generator_integration_test.cc
+++ b/generator/integration_tests/generator_integration_test.cc
@@ -91,7 +91,7 @@ class GeneratorIntegrationTest
     output_path_ = google::cloud::internal::GetEnv(
                        "GOOGLE_CLOUD_CPP_GENERATOR_OUTPUT_PATH")
                        .value_or(::testing::TempDir());
-    if (output_path_.empty() || output_path_.back() != '/') {
+    if (!output_path_.empty() && output_path_.back() != '/') {
       output_path_ += '/';
     }
 
--- a/google/cloud/internal/filesystem.cc
+++ b/google/cloud/internal/filesystem.cc
@@ -170,6 +170,26 @@ std::uintmax_t file_size(std::string const& path,
   return static_cast<std::uintmax_t>(stat.st_size);
 }
 
+std::string PathAppend(std::string const& directory, std::string const& path) {
+#if _WIN32
+  auto constexpr kSeparator = '\\';
+#else
+  auto constexpr kSeparator = '/';
+#endif
+  if (path.empty()) return directory;
+  if (directory.empty()) return path;
+  if (directory.back() != kSeparator && path.front() != kSeparator) {
+    return directory + kSeparator + path;
+  }
+  if (directory.back() != kSeparator || path.front() != kSeparator) {
+    return directory + path;
+  }
+  auto r = directory;
+  r.pop_back();
+  r += path;
+  return r;
+}
+
 }  // namespace internal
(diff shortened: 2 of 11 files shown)


google/cloud/internal/filesystem.cc
```diff
@@ -170,6 +170,26 @@ std::uintmax_t file_size(std::string const& path,
   return static_cast<std::uintmax_t>(stat.st_size);
 }
 
+std::string PathAppend(std::string const& directory, std::string const& path) {
+#if _WIN32
+  auto constexpr kSeparator = '\\';
```
devbww, line 175: Observation: The Windows APIs allow `'/'` as a separator, so the conditional compilation isn't really necessary.

generator/integration_tests/generator_integration_test.cc
```diff
@@ -91,6 +91,9 @@ class GeneratorIntegrationTest
     output_path_ = google::cloud::internal::GetEnv(
                        "GOOGLE_CLOUD_CPP_GENERATOR_OUTPUT_PATH")
                        .value_or(::testing::TempDir());
+    if (output_path_.empty() || output_path_.back() != '/') {
```
devbww, line 94: I'm not sure that an empty output path means we should write to the root directory, as opposed to the working directory.

google/cloud/internal/filesystem.cc
```diff
@@ -170,6 +170,26 @@ std::uintmax_t file_size(std::string const& path,
...
+#else
+  auto constexpr kSeparator = '/';
+#endif
+  if (path.empty()) return directory;
+  if (directory.empty()) return path;
+  if (directory.back() != kSeparator && path.front() != kSeparator) {
+    return directory + kSeparator + path;
+  }
+  if (directory.back() != kSeparator || path.front() != kSeparator) {
```
devbww, line 184: `directory.back() != kSeparator` implies `path.front() != kSeparator` in this context.  I'm not sure whether making the redundant check renders the code more or less understandable.

google/cloud/internal/filesystem.h
```diff
@@ -167,6 +167,9 @@ inline bool is_other(file_status s) noexcept {
 std::uintmax_t file_size(std::string const& path);
 std::uintmax_t file_size(std::string const& path, std::error_code& ec) noexcept;
 
+/// Append @p path (which may have subdirectories) to @p directory
```
devbww, line 170: Whether `path` can be absolute seems more important than whether it can include directories.

google/cloud/internal/filesystem_test.cc
```diff
@@ -28,12 +28,13 @@ namespace internal {
 namespace {
 using ::testing::HasSubstr;
 
-std::string CreateRandomFileName() {
+// When running on the internal Google CI systems we cannot write to the local
+// directory, GTest has a good temporary directory in that case.
```
devbww, line 32: s/, /, but / perhaps.

google/cloud/internal/filesystem.cc
coryan: Yes, but `directory.back() == kSeparator` says nothing about whether `path.front()` matches or not?

google/cloud/internal/filesystem.h
coryan: Fixed

google/cloud/internal/filesystem_test.cc
coryan: Done

google/cloud/internal/filesystem.cc
coryan: Yes, but I do not think we want `PathAppend("foo\\", "bar")` to return `"foo\\/bar"?

generator/integration_tests/generator_integration_test.cc
coryan: PTAL

google/cloud/internal/filesystem.h
```diff
@@ -167,6 +167,9 @@ inline bool is_other(file_status s) noexcept {
 std::uintmax_t file_size(std::string const& path);
 std::uintmax_t file_size(std::string const& path, std::error_code& ec) noexcept;
 
+/// Append @p path (even if it is an absolute directory) to @p directory
```
devbww, line 170: Nit: The directory-ness of path isn't a factor.

devbww (approved):

google/cloud/internal/filesystem.h
coryan: Fixed.

devjgm (approved):

Issue #6641, opened 2021-05-27
[Flake] //google/cloud/storage/emulator:test_utils fails


devjgm:
Since https://github.com/googleapis/google-cloud-cpp/pull/6611, this test seems to be failing occasionally.

GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/54295d0e-3835-4d29-9679-97f790b95e1d;tab=detail?project=cloud-cpp-testing-resources
Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/2424ec1c03d1dc03bc689846a47ca655fc7b3e10/fedora-34-asan/log-54295d0e-3835-4d29-9679-97f790b95e1d.txt

```
[8 / 10] Testing //google/cloud/storage/emulator:test_gcs; 1s remote-cache, linux-sandbox ... (4 actions running)
FAIL: //google/cloud/storage/emulator:test_utils (see /h/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/execroot/com_github_googleapis_google_cloud_cpp/bazel-out/k8-fastbuild/testlogs/google/cloud/storage/emulator/test_utils/test.log)
INFO: From Testing //google/cloud/storage/emulator:test_utils:
==================== Test output for //google/cloud/storage/emulator:test_utils:
.....................F.
======================================================================
FAIL: test_retry_test_reset_connection (__main__.TestRetryTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/h/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/sandbox/linux-sandbox/3873/execroot/com_github_googleapis_google_cloud_cpp/bazel-out/k8-fastbuild/bin/google/cloud/storage/emulator/test_utils.runfiles/com_github_googleapis_google_cloud_cpp/google/cloud/storage/emulator/tests/test_utils.py", line 545, in test_retry_test_reset_connection
    self.assertIn("Deleted %s" % (retry_test_id), body)
AssertionError: 'Deleted 323f324fe41f4c5aa0983f46581f528b' not found in '"Retry Test 323f324fe41f4c5aa0983f46581f528b does not exist."\n'
```

@frankyn FYI

devjgm:
I'm able to reproduce this flake locally with a patch like

## Repro

```diff
[...]

PR #6792, opened 2021-06-16
test(storage): deflake integration tests

coryan:
Use a different instance of the emulator to run the **emulator** tests
vs. the tests that **use** the emulator. The emulator tests
intentionally crash some of the emulator workers, which then causes
problems for the regular tests. Note that the emulator tests themselves
are still flaky, but they are very fast so using `--flake-test_attempts`
mostly avoids the problem.

Part of the work for #6641

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6792)

### DIFF
--- a/google/cloud/storage/emulator/ci/run_integration_tests_bazel.sh
+++ b/google/cloud/storage/emulator/ci/run_integration_tests_bazel.sh
@@ -27,8 +27,6 @@ fi
 # Configure run_emulators_utils.sh to find the instance admin emulator.
 BAZEL_BIN="$1"
 shift
-# ignore BAZEL_VERB="$1"
-shift
 
 bazel_test_args=("$@")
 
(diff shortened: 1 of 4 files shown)


google/cloud/storage/emulator/ci/run_integration_tests_bazel.sh
```diff
@@ -0,0 +1,56 @@
...
+
+if [[ $# -lt 1 ]]; then
+  echo "Usage: $(basename "$0") <bazel-program> [bazel-test-args]"
+  exit 1
+fi
+
+# Configure run_emulators_utils.sh to find the instance admin emulator.
+BAZEL_BIN="$1"
+shift
+# ignore BAZEL_VERB="$1"
```
devbww, line 30: Should we expand on why?

And if it should always be there, should we expect at least 2 arguments above?

devbww (approved):

PR #6795, opened 2021-06-16
ci: increase the GCB timeout

devbww:
We've been seeing some timeouts from the `integration-daily`
build since disabling CI-build test-result caching.  This
should not be too surprising given that the slow Spanner
backup/instance tests and samples already take a good part
of the allotted 1h, and are subject to scheduling delays in
the backend.  So, increase the timeout to 2h.

I couldn't, for the life of me, figure out how to parameterize
the timeout with a `gcloud builds submit --substitutions=`,
and chose not to move the value out to a `--timeout=` flag,
so this simply raises the limit for all builds.  Alternative
ideas are welcomed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6795)

### DIFF
--- a/ci/cloudbuild/cloudbuild.yaml
+++ b/ci/cloudbuild/cloudbuild.yaml
@@ -43,7 +43,7 @@ substitutions:
   _TRIGGER_TYPE: 'manual'
   _LOGS_BUCKET: 'cloud-cpp-community-publiclogs'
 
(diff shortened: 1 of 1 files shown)


devjgm (approved):
I think this way of doing it is fine and simple. LGTM. Thanks.

PR #6829, opened 2021-06-21
chore: retrieve repo via vcpkg by commit hash

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6829)


ci/cloudbuild/builds/lib/vcpkg.sh
```diff
@@ -24,15 +24,16 @@ source module ci/lib/io.sh
 
 TIMEFORMAT="==> 🕑 vcpkg installed in %R seconds"
 time {
-  VCPKG_RELEASE_VERSION="2021.05.12"
```
coryan, line 27: Why only this file?  There are other places that need similar fixes right?

coryan (approved):

PR #6840, opened 2021-06-22
Revert "chore(deps): update dependency grpc/grpc to v1.38.1"

devjgm:
Reverts googleapis/google-cloud-cpp#6827

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6840)

### DIFF
--- a/ci/cloudbuild/dockerfiles/demo-centos-7.Dockerfile
+++ b/ci/cloudbuild/dockerfiles/demo-centos-7.Dockerfile
@@ -125,7 +125,7 @@ RUN curl -sSL https://github.com/c-ares/c-ares/archive/cares-1_14_0.tar.gz | \
 
 # ```bash
 WORKDIR /var/tmp/build/grpc
-RUN curl -sSL https://github.com/grpc/grpc/archive/v1.38.1.tar.gz | \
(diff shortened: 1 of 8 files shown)


devjgm:
This revert is because after this PR our `demo-centos-7-pr` build fails. The issue seems to be related to `pkg-config`. Likely some diff in https://github.com/grpc/grpc/compare/v1.37.1...v1.38.1

devbww (approved):

devjgm:
Following up: The underlying issue was that centos-7's `pkg-config` is too old and suffers from a bad performance bug. I fixed this in https://github.com/googleapis/google-cloud-cpp/pull/6846, so our centos-7 image now has a newer version of `pkg-config`. We should be able to accept this grpc upgrade now, but I'll wait to see if renovate sends another upgrade PR for it.

coryan:
Renovate tries to minimize the "noise", and only sends one PR (ever) for each version.  I think renovate would *not* send a new PR unless:

* There is a newer release, such as 1.38.2 or 1.39.0
* You rename the old PRs to upgrade to 1.38.1

I just renamed the old PRs, let's see what happens.

PR #6855, opened 2021-06-24
fix(common): synchronize cancel for long-running operations

### LINKED ISSUE #6854: data race in AsyncPollingLoopImpl
Write of size 8 at 0x7b3c00002200 by **thread T21**:
```
  void OnPoll(future<StatusOr<Operation>> f) {
    auto op = f.get();
    if (op && op->done()) {
      return promise_.set_value(*std::move(op));
    }
    // Update the polling policy even on successful requests, so we can stop
    // after too many polling attempts.
    if (!polling_policy_->OnFailure(op.status())) {
      if (op) {
        return promise_.set_value(Status(
            StatusCode::kDeadlineExceeded,
            location_ + "() - polling loop terminated by polling policy"));
      }
      return promise_.set_value(std::move(op).status());
    }
    if (op) op->Swap(&op_);  // <--- HERE
    Wait();
  }
```
Previous read of size 8 at 0x7b3c00002200 by **main thread**:
```
  void DoCancel() {
    if (op_.name().empty()) return;  // <--- HERE
    google::longrunning::CancelOperationRequest request;
    request.set_name(op_.name());
    // Cancels are best effort, so we use weak pointers.
    auto w = WeakFromThis();
    cancel_(cq_, absl::make_unique<grpc::ClientContext>(), request)
        .then([w](future<Status> f) {
          if (auto self = w.lock()) self->OnCancel(f.get());
        });
  }
```
Location is heap block of size 232 at 0x7b3c000021c0 allocated by main thread:
```
future<StatusOr<Operation>> AsyncPollingLoop(
    google::cloud::CompletionQueue cq, future<StatusOr<Operation>> op,
    AsyncPollLongRunningOperation poll, AsyncCancelLongRunningOperation cancel,
    std::unique_ptr<PollingPolicy> polling_policy, std::string location) {
  auto loop = std::make_shared<AsyncPollingLoopImpl>(  // <--- HERE
      std::move(cq), std::move(poll), std::move(cancel),
      std::move(polling_policy), std::move(location));
  return loop->Start(std::move(op));
}
```

From `//google/cloud/spanner:database_admin_connection_test` `DatabaseAdminConnectionTest.CreateBackupCancel`, which exercises cancellation of a long-running operation.

GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/1c0a45a1-4d2b-4994-90ce-c6b8c90c084e;tab=detail?project=cloud-cpp-testing-resources
Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/9d7f46eba8d0d8ce4221ac0af2eaca67aab17887/fedora-34-tsan/log-1c0a45a1-4d2b-4994-90ce-c6b8c90c084e.txt

devbww:
Cancellation of a long-running operation (1) happens asynchronously
with the flow of callbacks, and (2) requires access to the currently
running operation name. This means we must serialize reads/writes of
that name, so introduce a member mutex.

And given that we now have a mutex, use it to guard `canceled_`
as well, avoiding the need for atomic operations. Also mark the
operation as cancelled when that happens before the operation name
is first set.

Finally, update the `DatabaseAdminConnectionTest.CreateBackupCancel`
test to allow for `AsyncCancelOperation()` not being called before
the polling policy expires.

Fixes #6854.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6855)

### DIFF
--- a/google/cloud/internal/async_polling_loop.cc
+++ b/google/cloud/internal/async_polling_loop.cc
@@ -60,7 +60,6 @@ class AsyncPollingLoopImpl
     google::longrunning::CancelOperationRequest request;
     {
       std::unique_lock<std::mutex> lk(mu_);
-      canceled_ = true;
       if (op_name_.empty()) return;
       request.set_name(op_name_);
     }
@@ -83,7 +82,7 @@ class AsyncPollingLoopImpl
     {
       std::unique_lock<std::mutex> lk(mu_);
       if (canceled_) return Cancelled();
-      op_name_ = std::move(op)->name();
+      op_name_ = std::move(*op->mutable_name());
     }
     Wait();
   }
@@ -134,7 +133,7 @@ class AsyncPollingLoopImpl
     if (op) {
       std::unique_lock<std::mutex> lk(mu_);
       if (canceled_) return Cancelled();
-      op_name_ = std::move(op)->name();
+      op_name_ = std::move(*op->mutable_name());
     }
     Wait();
   }
--- a/google/cloud/internal/async_polling_loop.cc
+++ b/google/cloud/internal/async_polling_loop.cc
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 #include "google/cloud/internal/async_polling_loop.h"
-#include <atomic>
+#include <mutex>
+#include <string>
 
 namespace google {
 namespace cloud {
@@ -33,10 +34,10 @@ class AsyncPollingLoopImpl
       : cq_(std::move(cq)),
         poll_(std::move(poll)),
         cancel_(std::move(cancel)),
-        canceled_(false),
         polling_policy_(std::move(polling_policy)),
         location_(std::move(location)),
-        promise_(null_promise_t{}) {}
+        promise_(null_promise_t{}),
+        canceled_(false) {}
 
   future<StatusOr<Operation>> Start(future<StatusOr<Operation>> op) {
     auto self = shared_from_this();
@@ -56,9 +57,13 @@ class AsyncPollingLoopImpl
   }
 
   void DoCancel() {
-    if (op_.name().empty()) return;
     google::longrunning::CancelOperationRequest request;
-    request.set_name(op_.name());
+    {
+      std::unique_lock<std::mutex> lk(mu_);
+      canceled_ = true;
+      if (op_name_.empty()) return;
+      request.set_name(op_name_);
+    }
     // Cancels are best effort, so we use weak pointers.
     auto w = WeakFromThis();
     cancel_(cq_, absl::make_unique<grpc::ClientContext>(), request)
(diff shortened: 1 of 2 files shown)


google/cloud/internal/async_polling_loop.cc
```diff
@@ -120,22 +131,29 @@ class AsyncPollingLoopImpl
       }
       return promise_.set_value(std::move(op).status());
     }
-    if (op) op->Swap(&op_);
+    if (op) {
+      std::unique_lock<std::mutex> lk(mu_);
+      if (canceled_) return Cancelled();
+      op_name_ = std::move(op)->name();
```
coryan, line 137: consider:

```cc
  op_name_ = std::move(*op->mutable_name());
```

google/cloud/internal/async_polling_loop.cc
```diff
@@ -56,9 +57,13 @@ class AsyncPollingLoopImpl
   }
 
   void DoCancel() {
-    if (op_.name().empty()) return;
     google::longrunning::CancelOperationRequest request;
-    request.set_name(op_.name());
+    {
+      std::unique_lock<std::mutex> lk(mu_);
+      canceled_ = true;
```
coryan, line 63: I think we should set this just on `OnCancel()`: the cancel request might fail after all.

google/cloud/internal/async_polling_loop.cc
```diff
@@ -69,12 +74,17 @@ class AsyncPollingLoopImpl
...
+    canceled_ = true;
   }
 
   void OnStart(StatusOr<Operation> op) {
     if (!op || op->done()) return promise_.set_value(std::move(op));
-    op_ = *std::move(op);
+    {
+      std::unique_lock<std::mutex> lk(mu_);
+      if (canceled_) return Cancelled();
+      op_name_ = std::move(op)->name();
```
coryan, line 86: Consider `op_name_ = std::move(*op->mutable_name());`

coryan (approved):
Thanks for fixing this!  Do you think it is possible to write some unit test for this?  I cannot think of a reasonable one.

I left a few nits below.

coryan (approved):

google/cloud/internal/async_polling_loop.cc
devbww: Done.

google/cloud/internal/async_polling_loop.cc
devbww: Done.

google/cloud/internal/async_polling_loop.cc
devbww: Removed, because it looks like it was insufficient for the purpose, but I still think something should be done here ... otherwise, a cancel before `OnStart()` is triggered, and `op_name_` filled in, will be ignored.

devbww:
> Do you think it is possible to write some unit test for this? I cannot think of a reasonable one.

I could not either, at least without intrusively adding hooks to control ordering.  So the conclusion I came to is that `tsan` is our friend, albeit scheduling dependent.

PR #6864, opened 2021-06-24
fix(storage): alternative endpoints and Host header

coryan:
Some alternatives for `storage.googleapis.com` require setting the
`Host:` header to `storage.googleapis.com`. This change (I think)
simplifies the host-header computation: we let `libcurl` do its thing in
most cases (it pulls it from the URL), and we override it only in the
cases where we know it is needed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6864)


google/cloud/storage/internal/curl_client.cc
```diff
@@ -104,8 +104,22 @@ bool XmlEnabled() {
 
 }  // namespace
 
+std::string HostHeader(Options const& options, char const* service) {
+  // This function returns an empty string if In most cases libcurl will fill
+  // out the correct `Host: ` header based on the URL.  The main exception is
+  // `https://restricted.googleapis.com` Hard-code the Host header. Applications
```
devbww, line 110: s/Hard-code/where we hard-code/ perhaps?

google/cloud/storage/internal/curl_client.h
```diff
@@ -31,6 +31,20 @@ inline namespace STORAGE_CLIENT_NS {
 namespace internal {
 class CurlRequestBuilder;
 
+/**
+ * Computes the `Host: ` header given the options.
```
devbww, line 35: s/options/options and service/ perhaps?

devbww (approved):

google/cloud/storage/internal/curl_client.cc
```diff
@@ -104,8 +104,22 @@ bool XmlEnabled() {
 
 }  // namespace
 
+std::string HostHeader(Options const& options, char const* service) {
+  // This function returns an empty string if In most cases libcurl will fill
```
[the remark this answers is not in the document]
coryan: I changed all this text, hopefully it is clearer now.

google/cloud/storage/internal/curl_client.cc
coryan: See above

google/cloud/storage/internal/curl_client.h
coryan: Done

devbww (approved):

PR #6867, opened 2021-06-24
cleanup(bigtable): fix AsyncReadRowsWithLimit example

dbolduc:
`AsyncReadRowsWithLimit()` contained identical code to `AsyncReadRows()`. They were both using the same [overload](https://github.com/googleapis/google-cloud-cpp/blob/a5d37eee2c1b5010738d51e3b7611d03cc6afa13/google/cloud/bigtable/table.h#L926) of `Table::AsyncReadRows`

I decided to make the limit settable via command line. And did the same for the synchronous version.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6867)

### DIFF
--- a/google/cloud/bigtable/examples/read_snippets.cc
+++ b/google/cloud/bigtable/examples/read_snippets.cc
@@ -107,11 +107,11 @@ void PrepareReadSamples(google::cloud::bigtable::Table table) {
 }
 
 void ReadRowsWithLimit(google::cloud::bigtable::Table table,
-                       std::vector<std::string> const&) {
+                       std::vector<std::string> const& argv) {
   //! [read rows with limit]
   namespace cbt = google::cloud::bigtable;
   using google::cloud::StatusOr;
-  [](cbt::Table table) {
+  [](cbt::Table table, std::int64_t limit) {
     // Create the range of rows to read.
     auto range = cbt::RowRange::Range("phone#4c410523#20190501",
                                       "phone#4c410523#20190502");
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/examples/read_snippets.cc
```diff
@@ -121,8 +121,8 @@ void ReadRowsWithLimit(google::cloud::bigtable::Table table,
         cbt::Filter::ColumnRangeClosed("stats_summary", "connected_wifi",
                                        "connected_wifi"),
         cbt::Filter::Latest(1));
-    // Read and print the first 5 rows in the range.
-    for (StatusOr<cbt::Row> const& row : table.ReadRows(range, 5, filter)) {
+    // Read and print the first rows in the range, within the row limit.
```
coryan, line 124: I think this example is shared with other libraries, check with @billyjacobson to make sure we are not making the example harder to follow in C++ vs. other languages.

dbolduc: This example actually doesn't have a DevRel region tag. It only has a tag for our dox.

You may be thinking of this [Limit Cells Per Row example](https://cloud.google.com/bigtable/docs/samples/bigtable-filters-limit-cells-per-row) found in `data_filter_snippets.cc`. (why are there so many different files -- I do not know.)

I'm also noticing that the filter example I linked does not look like the other languages' examples... (I will fix)

dbolduc: Actually none of the filter examples conform to what other libraries use. I opened #6868

coryan (approved):
SGTM, we can fix the other examples elsewhere.

PR #6870, opened 2021-06-25
fix(storage): crashes with tracing enabled

coryan:
With some MSVC configurations (2019+Debug, 2017+Debug, or 2017+Release)
enabling HTTP tracing creates a crash. I think because in other
compilers the move constructor and assignments are not used, and luckily
that means the pointers were stable, but this is not the case for those
versions/configurations of MSVC.

The integration test crashes before this change and succeeds after the
change.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6870)

### DIFF
--- a/google/cloud/storage/internal/curl_handle.cc
+++ b/google/cloud/storage/internal/curl_handle.cc
@@ -148,7 +148,8 @@ void CurlHandle::ResetSocketCallback() {
 
 void CurlHandle::EnableLogging(bool enabled) {
   if (enabled) {
-    SetOption(CURLOPT_DEBUGDATA, &debug_info_);
+    debug_info_ = std::make_shared<DebugInfo>();
+    SetOption(CURLOPT_DEBUGDATA, debug_info_.get());
     SetOption(CURLOPT_DEBUGFUNCTION, &CurlHandleDebugCallback);
     SetOption(CURLOPT_VERBOSE, 1L);
   } else {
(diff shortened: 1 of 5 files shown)


dbolduc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6870)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[google/cloud/storage/internal/curl_handle.h, line 50 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6870#-Md2W1ik0Q9SV9goIo3U:-Md2W1ik0Q9SV9goIo3V:ba1b7wx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/61cfbd56475ccf5d00fea534fbf7a7aec561a893/google/cloud/storage/internal/curl_handle.h#L50)):*
> ```
>   // `std::shared_ptr<>`. For other callbacks they callback
>   // is set up only once the object is stable and won't move
> ```

I don't know what you are trying to say.

"For other callbacks, they are set up" <-- if you are talking about the pointer
"Other callbacks are set up ..." <-- if you are talking about the callback

coryan (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6870)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @dbolduc)

___
*[google/cloud/storage/internal/curl_handle.h, line 50 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6870#-Md2W1ik0Q9SV9goIo3U:-Md2m-ac-SgaJXtQHRNl:b73eokx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/61cfbd56475ccf5d00fea534fbf7a7aec561a893/google/cloud/storage/internal/curl_handle.h#L50)):*

> ```
>   // `std::shared_ptr<>`. For other callbacks they callback
>   // is set up only once the object is stable and won't move
> ```

I don't know what you are trying to say.

"For other callbacks, they are set up" <-- if you are talking about the pointer
"Other callbacks are set up ..." <-- if you are talking about the callback

</blockquote></details>

Yeah, that was not very clear. PTAL.

dbolduc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6870)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[google/cloud/storage/internal/curl_handle.h, line 50 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6870#-Md2pBm8AREY5tWhWjVs:-Md2pBm8AREY5tWhWjVt:b-4id4sc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2902825f8382d4ff3fd3b2a0593085c7265599f3/google/cloud/storage/internal/curl_handle.h#L50)):*
> ```objc
>   // * For the `debug` callback (used rarely), we use a `std::shared_ptr<>`.
>   // * For the `socket` callback, the only classes that use it are
>   //   `CurlDownloadRequest` and `CurlRequest`, those classes guarantee the
> ```

s/, those/. Those

dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6870)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @coryan)

PR #6882, opened 2021-06-27
feat(storage): share client in throughput benchmark

coryan:
Now that the GCS+gRPC plugin supports multiple channels it makes more
sense to use the same client in all threads, at least by default.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6882)

### DIFF
--- a/google/cloud/storage/benchmarks/throughput_experiment.cc
+++ b/google/cloud/storage/benchmarks/throughput_experiment.cc
@@ -324,35 +324,33 @@ class DownloadObjectRawGrpc : public ThroughputExperiment {
 
 }  // namespace
 
+using ::google::cloud::storage_experimental::DefaultGrpcClient;
+
 std::vector<std::unique_ptr<ThroughputExperiment>> CreateUploadExperiments(
     ThroughputOptions const& options,
+    google::cloud::storage::Client rest_client,
+    google::cloud::storage::Client grpc_client,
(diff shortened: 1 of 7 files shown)


google/cloud/storage/benchmarks/throughput_experiment.cc
```diff
@@ -369,23 +367,24 @@ std::vector<std::unique_ptr<ThroughputExperiment>> CreateUploadExperiments(
...
+    google::cloud::storage::Client rest_client,
+    google::cloud::storage::Client grpc_client,
     google::cloud::Options const& client_options, int thread_id) {
-  gcs::Client rest_client(client_options);
+  if (options.client_per_thread) rest_client = gcs::Client(client_options);
+#if GOOGLE_CLOUD_CPP_STORAGE_HAVE_GRPC
+  if (options.client_per_thread) {
+    grpc_client = DefaultGrpcClient(client_options);
+  }
+#else
```
devbww, line 378: A perhaps-naive question: Why is this different from `CreateUploadExperiments()`?

devbww (approved):

google/cloud/storage/benchmarks/throughput_experiment.cc
coryan: Not naive at all, fixed.

PR #6888, opened 2021-06-29
ci(spanner): run spanner integration tests against prod

devbww:
Add a hook for running integration tests without the emulator,
execute the Spanner tests from it, and call the hook during
the `integration-daily` build.

The Spanner samples/tests have cases that are skipped or
short-circuited when running against the emulator, so we lost
some coverage when we move to using the emulator exclusively
from the "normal" builds.  Now we'll at least hit those cases
during `integration-daily`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6888)

### DIFF
--- a/ci/cloudbuild/builds/integration-daily.sh
+++ b/ci/cloudbuild/builds/integration-daily.sh
@@ -23,7 +23,7 @@
 # `ci/cloudbuild/triggers` directory. "Manual" triggers exist only within the
 # GCB UI at the time of this writing. Users with the appropriate access can run
 # this build by hand with:
-#   `ci/cloudbuild/build.sh --distro fedora integration-daily --project cloud-cpp-testing-resources`
+#   `ci/cloudbuild/build.sh --distro fedora-34 integration-daily --project cloud-cpp-testing-resources`
 
 set -eu
 
@@ -44,3 +44,4 @@ export GOOGLE_CLOUD_CPP_SPANNER_SLOW_INTEGRATION_TESTS="instance,backup"
 export GOOGLE_CLOUD_CPP_IAM_QUOTA_LIMITED_INTEGRATION_TESTS="yes"
 mapfile -t integration_args < <(integration::bazel_args)
 integration::bazel_with_emulators test "${args[@]}" "${integration_args[@]}"
+integration::bazel_without_emulators test "${args[@]}" "${integration_args[@]}"
--- a/ci/cloudbuild/builds/integration-daily.sh
+++ b/ci/cloudbuild/builds/integration-daily.sh
@@ -44,4 +44,7 @@ export GOOGLE_CLOUD_CPP_SPANNER_SLOW_INTEGRATION_TESTS="instance,backup"
 export GOOGLE_CLOUD_CPP_IAM_QUOTA_LIMITED_INTEGRATION_TESTS="yes"
 mapfile -t integration_args < <(integration::bazel_args)
 integration::bazel_with_emulators test "${args[@]}" "${integration_args[@]}"
-integration::bazel_without_emulators test "${args[@]}" "${integration_args[@]}"
+
+io::log_h2 "Running Spanner integration tests (against prod)"
+bazel test "${args[@]}" "${integration_args[@]}" \
+  --test_tag_filters="integration-test" google/cloud/spanner/...
--- a/ci/cloudbuild/builds/lib/integration.sh
+++ b/ci/cloudbuild/builds/lib/integration.sh
@@ -220,22 +220,6 @@ function integration::bazel_with_emulators() {
     //google/cloud/examples/...
 }
 
-# Runs integration tests with bazel, like above, but without using emulators.
-function integration::bazel_without_emulators() {
-  if [[ $# == 0 ]]; then
-    io::log_red "error: bazel verb required"
-    return 1
-  fi
-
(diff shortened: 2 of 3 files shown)


ci/cloudbuild/builds/integration-daily.sh
```diff
@@ -44,3 +44,4 @@ export GOOGLE_CLOUD_CPP_SPANNER_SLOW_INTEGRATION_TESTS="instance,backup"
 export GOOGLE_CLOUD_CPP_IAM_QUOTA_LIMITED_INTEGRATION_TESTS="yes"
 mapfile -t integration_args < <(integration::bazel_args)
 integration::bazel_with_emulators test "${args[@]}" "${integration_args[@]}"
+integration::bazel_without_emulators test "${args[@]}" "${integration_args[@]}"
```
devjgm, line 47: Rather than introducing this new function, maybe the daily build should run everything without emulators. Like what if we replaced lines 46 and 47 with simply (we could remove line 40 too):

```shell
bazel test "${args[@]}" "${integration_args[@]}" ...
```

It seems like that may be a once-per-day way to cover ourselves that we never forget to run integration tests against actual prod _sometimes_. WDYT?

devjgm: One more thought: If some of our integration tests do not easily work against production today, we could start with only spanner by replacing the `bazel_without_emulators` call with 

```shell
# Run Spanner's integration tests against production.
bazel test --test_tag_filters="integration-test" "${args[@]}" "${integration_args[@]}" google/cloud/spanner/...
```

ci/cloudbuild/builds/lib/integration.sh
```diff
@@ -220,6 +220,22 @@ function integration::bazel_with_emulators() {
...
+    return 1
+  fi
+
+  local verb="$1"
+  local args=("${@:2}")
+
+  io::log_h2 "Running Spanner integration tests"
+  bazel "${verb}" "${args[@]}" \
+    --test_tag_filters="integration-test" -- \
+    "//google/cloud/spanner/...:all"
```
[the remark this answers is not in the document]
devbww: I have some vague recollection that there is/was a difference in regard to the handling of "manual" tagged tests, but I cannot find areference for that now.  In any case, this line is now gone.

The pattern does occur in three `run_integration_tests_emulator_bazel.sh` files, which we can think about some other time.

ci/cloudbuild/builds/lib/integration.sh
```diff
@@ -220,6 +220,22 @@ function integration::bazel_with_emulators() {
     //google/cloud/examples/...
 }
 
+# Runs integration tests with bazel, like above, but without using emulators.
+function integration::bazel_without_emulators() {
```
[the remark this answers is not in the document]
devbww: Thanks.  I'll go with one of @devjgm's suggestions.

ci/cloudbuild/builds/integration-daily.sh
devbww: Indeed, some of the tests failed when I tried the `...`.
```
//google/cloud/bigtable/examples:bigtable_grpc_credentials                  FAILED in 0.4s
//google/cloud/examples:grpc_credential_types                               FAILED in 0.4s
//google/cloud/pubsub/benchmarks:endurance                                  FAILED in 1.6s
//google/cloud/pubsub/benchmarks:throughput                                 FAILED in 2.0s
//google/cloud/pubsub/integration_tests:schema_admin_integration_test       FAILED in 0.9s
//google/cloud/pubsub/integration_tests:subscriber_integration_test         FAILED in 14.1s
//google/cloud/pubsub/integration_tests:subscription_admin_integration_test FAILED in 3.9s
//google/cloud/pubsub/integration_tests:topic_admin_integration_test        FAILED in 2.2s
//google/cloud/pubsub/samples:samples                                       FAILED in 2.4s
//google/cloud/storage/examples:storage_service_account_samples             FAILED in 1.2s
//google/cloud/storage/tests:grpc_integration_test                          FAILED in 17.2s
//google/cloud/storage/tests:service_account_integration_test               FAILED in 1.9s
```
I did not look into the root causes (although one was just complaining that "The GOOGLE_CLOUD_CPP_TEST_HELLO_WORLD_SERVICE_ACCOUNT environment variable is not set", implying that some "--test_env" settings should be moved to `integration::bazel_args()`), nor into which were cascading failures.

So, I've gone with your "we could start with only spanner" suggestion, although by running it in addition to `bazel_without_emulators` instead of replacing it.  PTAL.

I'm still not enthralled with having to edit build scripts to get any pre-merge coverage for the Spanner prod-only features, but at least this will give us some daily, post-merge confidence that things are working.

devjgm (approved):

ci/cloudbuild/builds/integration-daily.sh
devjgm: > I'm still not enthralled with having to edit build scripts to get any pre-merge coverage for the Spanner prod-only features, but at least this will give us some daily, post-merge confidence that things are working.

We could consider _one_ pr and ci build that runs integration tests without emulators. If we did, we may need to think about test latency, quotas, and possibly flakes. Thought w/ test caching _enabled_ for PR builds (but not CI builds), that may not be too big of an issue.

PR #6892, opened 2021-06-29
feat(storage): micro-optimization for downloads

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6892)

### DIFF
--- a/google/cloud/storage/internal/object_streambuf.cc
+++ b/google/cloud/storage/internal/object_streambuf.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/storage/object_stream.h"
 #include "absl/memory/memory.h"
 #include <cstring>
+#include <iterator>
 #include <sstream>
 
 namespace google {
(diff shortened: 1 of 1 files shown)


devjgm (approved):

google/cloud/storage/internal/object_streambuf.cc
```diff
@@ -68,9 +68,10 @@ StatusOr<ObjectReadStreambuf::int_type> ObjectReadStreambuf::Peek() {
   // assert(n <= current_ios_buffer_.size())
   current_ios_buffer_.resize(read_result->bytes_received);
 
-  for (auto const& kv : read_result->response.headers) {
+  auto hint = headers_.end();
+  for (auto& kv : read_result->response.headers) {
     hash_validator_->ProcessHeader(kv.first, kv.second);
-    headers_.emplace(kv.first, kv.second);
+    hint = headers_.emplace_hint(hint, kv.first, std::move(kv.second));
```
devbww, line 74: Given that the `kv.first` values are in order, it seems like a better hint for the next insertion would be the successor of the last insertion.

devbww, line 74: How come you can't move `kv.first` too?

coryan: It is a `std::string const`, because `value_type` for associative containers is `std::pair<K const, V>` as you probably know better than I do.

devbww: Apparently not.

coryan: Done. Keeping my fingers crossed ... hopefully all the compilers we use support `std::next()`.

PR #6895, opened 2021-06-29
feature(spanner): low-cost instances

devbww:
Support the provisioning of Spanner instances using `processing_units`
as an alternative to `node_count`. 1 node == 1,000 processing units.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6895)


google/cloud/spanner/samples/samples.cc
```diff
@@ -86,19 +86,56 @@ void CreateInstance(google::cloud::spanner::InstanceAdminClient client,
...
+                  .SetProcessingUnits(500)
+                  .SetLabels({{"cloud_spanner_samples", "true"}})
+                  .Build())
+          .get();
+  if (!instance) throw std::runtime_error(instance.status().message());
+  std::cout << "Created instance [" << in << "]:\n" << instance->DebugString();
+
+  instance = client.GetInstance(in);
+  if (!instance) throw std::runtime_error(instance.status().message());
+  std::cout << "Instance " << in << " has " << instance->processing_units()
```
coryan, line 117: Does the example require this call to `.GetInstance()`?  That seems slightly odd, as the `.CreateInstance()` operation should return processing units too.  If it is not needed, may elide it to keep the example simpler.  If it is need it maybe a comment explaining why?

coryan (approved):

google/cloud/spanner/samples/samples.cc
devbww: I have a recollection of asking that very question, but I cannot find a trace of the answer.  So, I've asked again and will report back here.

If I was pressed to dig up a memory, I think it would be that the `CreateInstance()` actually didn't return the processing units, whether for good reason or bad.

coryan: Feel free to merge as-is and cleanup when you get an answer.  I did not mean to block you.

devbww: The response was, _"the call for getInstance() was to essentially have a test for the API surfaces that were touched by the published protos.  Also to show how to use the field mask."_

The "field mask" stuff is hidden from C++ users (at least while we have a non-generated admin API).  But otherwise we don't have anything showing `GetInstance()` returning `processing_units`, so I'm inclined to leave it because of that.  And also to maintain some similarity with the sample in other languages.

coryan: SGTM.

PR #6901, opened 2021-06-30
cleanup(spanner): remove empty options.cc

dbolduc:
---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6901)


dbolduc:
@devjgm - if I can successfully run `ci/cloudbuild/build.sh -t clang-tidy-pr --docker` does that mean I can delete this file?

dbolduc:
The builds all succeeded, so I have to think this change is benign

coryan (approved):

devjgm:
> @devjgm - if I can successfully run `ci/cloudbuild/build.sh -t clang-tidy-pr --docker` does that mean I can delete this file?

My guess is that the clang-tidy that came w/ fedora 33 showed the false positive that this empty file was working around. And now that we use fedora 34 it's no longer needed. So if all the builds succeed (and they do), then I agree that this file is not needed anymore.

PR #6913, opened 2021-07-01
cleanup(storage): simplify integration test cleanup

coryan:
Most of our integration tests create some objects, which then need to be
deleted when the test ends. In addition to having to remember to delete
the object, some of the tests are a bit unnatural, as they should use
`ASSERT_*()` but they use `EXPECT_*()` to ensure the cleanup step runs.

With this change the cleanup is performed by a common `TearDown()`
function, the tests just add objects (and buckets if applicable) to a
list that will be deleted at the end.

In some tests the `DeleteObject()` function is actually part of the
test, i.e., we are testing the `DeleteObject()` function. In these cases
I left the code alone. I also left unchanged some tests where objects
are created without a `gcs::ObjectMetadata` becoming available,
typically because the test is using an streaming upload  implicitly
finalized by the stream's destructor.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6913)

### DIFF
--- a/google/cloud/storage/testing/storage_integration_test.cc
+++ b/google/cloud/storage/testing/storage_integration_test.cc
@@ -38,6 +38,20 @@ static bool UseGrpcForMedia() {
   return v.find("media") != std::string::npos;
 }
 
+void StorageIntegrationTest::TearDown() {
+  // The client configured to create and delete buckets is good for our
+  // purposes.
+  auto client = MakeBucketIntegrationTestClient();
+  if (!client) return;
+  for (auto& o : objects_to_delete_) {
+    (void)client->DeleteObject(o.bucket(), o.name(),
+                               Generation(o.generation()));
+  }
+  for (auto& b : buckets_to_delete_) {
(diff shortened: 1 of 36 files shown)


google/cloud/storage/testing/storage_integration_test.cc
```diff
@@ -38,6 +38,20 @@ static bool UseGrpcForMedia() {
   return v.find("media") != std::string::npos;
 }
 
+void StorageIntegrationTest::TearDown() {
```
devbww, line 41: https://github.com/google/googletest/blob/master/docs/faq.md#should-i-use-the-constructordestructor-of-the-test-fixture-or-setupteardown-ctorvssetup indicates a preference for `~StorageIntegrationTest()` for cleanup of this variety.  What do you think?

coryan: Good idea, done.

devbww (approved):

PR #6918, opened 2021-07-01
fix(iam): SetIamPolicy sample from attempting concurrent changes

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6918)

### DIFF
--- a/google/cloud/iam/samples/iam_samples.cc
+++ b/google/cloud/iam/samples/iam_samples.cc
@@ -543,13 +543,13 @@ void AutoRun(std::vector<std::string> const& argv) {
   QueryGrantableRoles(
       {absl::StrCat("//iam.googleapis.com/projects/", project_id,
                     "/serviceAccounts/", service_account_id)});
-  SetIamPolicy({service_account_name});
   ListRoles({absl::StrCat("projects/", project_id)});
   QueryTestablePermissions(
       {absl::StrCat("//iam.googleapis.com/projects/", project_id,
                     "/serviceAccounts/", service_account_id)});
 
   if (RunQuotaLimitedSamples()) {
+    SetIamPolicy({service_account_name});
     CreateServiceAccount({project_id, "sample-account-id", "SampleAccount",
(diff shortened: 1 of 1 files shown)


devbww (approved):

google/cloud/iam/samples/iam_samples.cc
```diff
@@ -543,13 +543,13 @@ void AutoRun(std::vector<std::string> const& argv) {
...
       {absl::StrCat("//iam.googleapis.com/projects/", project_id,
                     "/serviceAccounts/", service_account_id)});
-  SetIamPolicy({service_account_name});
   ListRoles({absl::StrCat("projects/", project_id)});
   QueryTestablePermissions(
       {absl::StrCat("//iam.googleapis.com/projects/", project_id,
                     "/serviceAccounts/", service_account_id)});
 
   if (RunQuotaLimitedSamples()) {
+    SetIamPolicy({service_account_name});
```
coryan, line 552: This will probably work, but the example could be much better.  This code:

https://github.com/googleapis/google-cloud-cpp/blob/970c2fe99b35757b826fedd8211bb49ab26e5ae6/google/cloud/iam/samples/iam_samples.cc#L234

Should really show how to implement a OCC loop, something like:

```cc
auto backoff = std::chrono::seconds(1);
for (int i = 0; i != kMaxAttempts; ++i) {
  auto current = client.GetIamPolicy(name);
  if (!current) continue;
  auto policy = *current;
  // change the policy, e.g., add or remove an entry
  auto updated = client.SetIamPolicy(name, policy);
  if (updated.ok()) break; // success!
  if (updated.code() != StatusCode::kAborted) {
    std::this_thread::sleep_for(backoff);
    backoff *= 2;
    continue;
  }
  throw std::runtime_error(updated.status().message()); // unrecoverable
}
```

coryan (approved):
I just noticed that each example creates its own Client.  That is probably Okay, just be aware that in other libraries we found that to be too slow.

PR #6921, opened 2021-07-01
cleanup(bigtable)!: remove async functions that take `CompletionQueue` as a param

### LINKED ISSUE #2567: Remove CompletionQueue parameter from Async*() functions.

dbolduc:
Fixes #2567 

Note: this leaves a lot of `*Impl()`'s that are called only once. I will refactor these in a subsequent PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6921)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,19 @@
 
 ## v1.30.0 - TBD
 
+### Bigtable:
+
+**BREAKING CHANGES**:
+* `Async*` functions in `Table` that took `CompletionQueue&` as a parameter
+  have been removed. Users that need to update their code should use the
+  identically named functions which do not require a `CompletionQueue&`. If
+  users absolutely require a custom `CompletionQueue`, one can be supplied to
+  the `DataClient` used to construct the `Table`, via
+  `ClientOptions::DisableBackgroundThreads(CompletionQueue const&)`.
+  See [#2567][issue-2567] for more details.
+
+[issue-2567]: https://github.com/googleapis/google-cloud-cpp/issues/2567
+
 ## v1.29.0 - 2021-07
 
 ### Bigtable:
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -65,6 +65,19 @@
...
 
+### Bigtable:
+
+**BREAKING CHANGES**:
+* `Async*` functions in `Table` that took `CompletionQueue&` as a parameter
+  have been removed. Users that need to update their code should use the
+  identically named functions which do not require a `CompletionQueue&`. If
+  users absolutely require a custom `CompletionQueue`, one can be supplied to
+  the `DataClient` used to construct the `Table`, via
+  `ClientOptions::DisableBackgroundThreads(CompletionQueue const&)`.
```
dbolduc, line 76: 1. Is this even worth saying?
2. The change to prefer `Options` > `ClientOptions` will be done in this release. I am making a mental note to come back and update `CHANGELOG.md` accordingly

coryan: In general, I think "You cannot use Foo() anymore, use Bar() instead" is more friendly/useful than "Foo() was removed."  

It is also better to assume that folks have not memorized our reference manual, so they may not be familiar with every option in the library.

With those caveats, pointing folks to the *recommended* way to do things is the "Right Thing":tm:

coryan: The change is Okay as-is, we just need to update the CHANGELOG on the PRs where we start recommending `Options{}` instead.

coryan (approved):
Nice!

PR #6929, opened 2021-07-02
cleanup(internal): GCP_LOG(FATAL) terminates execution

### LINKED ISSUE #5693: GCP_LOG(FATAL) should terminate execution
After `GCP_LOG(FATAL) << streamables` completes (and flushes) its output, it should terminate the process.

Then, cleanup all call sites that are currently followed by some form of `std::exit()`.

remyabel:
Fixes #5693

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6929)

### DIFF
--- a/google/cloud/internal/log_impl.cc
+++ b/google/cloud/internal/log_impl.cc
@@ -14,6 +14,7 @@
 
 #include "google/cloud/internal/log_impl.h"
 #include "google/cloud/internal/getenv.h"
+#include "google/cloud/terminate_handler.h"
 
 namespace google {
 namespace cloud {
@@ -26,17 +27,24 @@ void StdClogBackend::Process(LogRecord const& lr) {
   if (lr.severity >= Severity::GCP_LS_WARNING) {
     std::clog << std::flush;
   }
+  if (lr.severity == Severity::GCP_LS_FATAL) {
+    google::cloud::Terminate("Encountered fatal error.");
+  }
 }
 
 void StdClogBackend::Flush() { std::clog << std::flush; }
 
 void CircularBufferBackend::ProcessWithOwnership(LogRecord lr) {
   std::unique_lock<std::mutex> lk(mu_);
-  auto const needs_flush = lr.severity >= min_flush_severity_;
+  auto const severity = std::move(lr.severity);
+  auto const needs_flush = severity >= min_flush_severity_;
   buffer_[index(end_)] = std::move(lr);
   ++end_;
   if (end_ - begin_ > buffer_.size()) ++begin_;
   if (needs_flush) FlushImpl(std::move(lk));
+  if (severity == Severity::GCP_LS_FATAL) {
+    google::cloud::Terminate("Encountered fatal error.");
+  }
 }
 
 void CircularBufferBackend::Flush() {
--- a/google/cloud/spanner/row.cc
+++ b/google/cloud/spanner/row.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/status.h"
 #include "google/cloud/status_or.h"
 #include <algorithm>
+#include <cstdlib>
 #include <iterator>
 #include <utility>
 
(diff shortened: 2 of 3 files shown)


google/cloud/spanner/row.cc
```diff
@@ -51,6 +52,7 @@ Row::Row(std::vector<Value> values,
   if (values_.size() != columns_->size()) {
     GCP_LOG(FATAL) << "Row's value and column sizes do not match: "
                    << values_.size() << " vs " << columns_->size();
+    std::exit(EXIT_FAILURE);
```
coryan, line 55: Why the call to `std::exit`?  Presumably we want that to happen in the `GCP_LOG(FATAL)` line?

google/cloud/internal/log_impl.cc
```diff
@@ -26,17 +27,24 @@ void StdClogBackend::Process(LogRecord const& lr) {
...
+    google::cloud::Terminate("Encountered fatal error.");
+  }
 }
 
 void StdClogBackend::Flush() { std::clog << std::flush; }
 
 void CircularBufferBackend::ProcessWithOwnership(LogRecord lr) {
   std::unique_lock<std::mutex> lk(mu_);
-  auto const needs_flush = lr.severity >= min_flush_severity_;
+  auto const severity = std::move(lr.severity);
```
coryan, line 39: I do not believe you want a `std::move()` here.  First, it has no effect, it is an enum.  Second, if it *had* any effect the assignment to `buffer_[index(end_)]` would lose information.

google/cloud/internal/log_impl.cc
```diff
@@ -26,17 +27,24 @@ void StdClogBackend::Process(LogRecord const& lr) {
   if (lr.severity >= Severity::GCP_LS_WARNING) {
     std::clog << std::flush;
   }
+  if (lr.severity == Severity::GCP_LS_FATAL) {
```
coryan, line 30: Consider an alternative implementation.  Create a new `LogBackend` which *only* calls terminate if the severity is `GCP_LS_FATAL`, and initialize it here:

https://github.com/googleapis/google-cloud-cpp/blob/8f6afc3c9d503ebb634f189fd94f915c2e308b34/google/cloud/log.cc#L200

Maybe we can split `GOOGLE_CLOUD_CPP_EXPERIMENTAL_LOG_CONFIG` by `:` first, so people can write:

```
export GOOGLE_CLOUD_CPP_EXPERIMENTAL_LOG_CONFIG=terminate-on-fatal:clog
```

or 

```
export GOOGLE_CLOUD_CPP_EXPERIMENTAL_LOG_CONFIG=terminate-on-fatal:lastN,20,ERROR
```

What do you think?

devbww: By "`GOOGLE_CLOUD_CPP_EXPERIMENTAL_LOG_CONFIG=terminate-on-fatal`", are you suggesting the behavior should be conditional?

coryan: Hmmm... yes, I was thinking that.  I see where that would be weird or wrong.  Maybe we can start with the behavior being opt-in (after all, it does not crash today!) and then move to opt-out?

devbww: Given that all but two of the existing `GCP_LOG(FATAL)` calls are immediately followed by `std::exit()`, and the intuition that in those other two cases the author thought it would not return, I'm not sure that we need such a migration strategy.

If my intuition is wrong (we can check with the authors), then perhaps an easier thing to do would be to just change those two to `GCP_LOG(ERROR)` first.

coryan: I should have looked, thanks for doing so.  Yes, unconditional without migration SGTM then.

remyabel:
Let me revisit this later. Hold on.

PR #6941, opened 2021-07-06
feat(storage): CPU usage in throughput benchmark

### LINKED ISSUE #6884: Change benchmarks to measure downloads using existing data and GCS+gRPC
We need a GCS+gRPC benchmark that measures download bandwidth using existing data, so we can run the benchmark from a cluster of machines all using the same dataset.

The `throughput_vs_cpu` benchmark creates its own bucket, then uploads an object, and then downloads it, so it does not fit the bill.

The `shard_throughput` benchmark does not support GCS+gRPC, and does not measure CPU time.

It seems like we need a blend of the two.

Note that we *also* need a too to populate the datasets, in the past I have used the existing benchmarks with options to "do not delete the data you create".

coryan:
Capture CPU usage in the `aggregate_throughput_benchmark`. This is
useful when trying to compare the costs of different protocols.

Fixes #6884

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6941)

### DIFF
--- a/google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc
+++ b/google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc
@@ -22,6 +22,7 @@
 #include "google/cloud/internal/random.h"
 #include "google/cloud/options.h"
 #include "google/cloud/testing_util/command_line_parsing.h"
+#include "google/cloud/testing_util/timer.h"
 #include "absl/time/time.h"
 #include <algorithm>
 #include <atomic>
@@ -34,7 +35,9 @@
 
 namespace {
 using ::google::cloud::storage_experimental::DefaultGrpcClient;
+using ::google::cloud::testing_util::CpuAccounting;
 using ::google::cloud::testing_util::FormatSize;
+using ::google::cloud::testing_util::Timer;
 namespace gcs = google::cloud::storage;
 namespace gcs_bm = google::cloud::storage_benchmarks;
 using gcs_bm::AggregateThroughputOptions;
@@ -215,10 +218,12 @@ int main(int argc, char* argv[]) {
   using clock = std::chrono::steady_clock;
   auto const start = clock::now();
   auto const deadline = clock::now() + options->running_time;
+  auto timer = Timer{CpuAccounting::kPerProcess};
   while (clock::now() < deadline) {
     std::this_thread::sleep_for(options->reporting_interval);
-    std::cout << current_time() << "," << accumulate_bytes_received()
-              << std::endl;
+    auto const usage = timer.Sample();
+    std::cout << current_time() << "," << accumulate_bytes_received() << ","
+              << usage.cpu_time.count() << std::endl;
   }
   auto const elapsed = clock::now() - start;
   auto const bytes_received = accumulate_bytes_received();
--- a/google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc
+++ b/google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc
@@ -35,7 +35,6 @@
 
 namespace {
 using ::google::cloud::storage_experimental::DefaultGrpcClient;
-using ::google::cloud::testing_util::CpuAccounting;
 using ::google::cloud::testing_util::FormatSize;
 using ::google::cloud::testing_util::Timer;
 namespace gcs = google::cloud::storage;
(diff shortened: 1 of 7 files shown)


dbolduc (dismissed):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6941)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[google/cloud/testing_util/timer.cc, line 29 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6941#-Mdxj0n9DpT2LdHBYUvr:-Mdxj0n9DpT2LdHBYUvs:bqf6lyi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e8bab17e3ab3e1eabceb6bd221d327699b13dbf0/google/cloud/testing_util/timer.cc#L29)):*
> ```cpp
>   return accounting == CpuAccounting::kPerThread ? RUSAGE_THREAD : RUSAGE_SELF;
> #else
>   (void)accounting;
> ```

this seems weird to me. would a NOLINTNEXTLINE above `int rusage_who()` be possible/preferable?

coryan (commented):
PTAL

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6941#-:-Mdxq9980i8HYt1GArcZ:bb71pro)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @dbolduc)

___
*[google/cloud/testing_util/timer.cc, line 29 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6941#-Mdxj0n9DpT2LdHBYUvr:-MdxpRPJBJPAS_wn7EXO:b8fxz72) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e8bab17e3ab3e1eabceb6bd221d327699b13dbf0/google/cloud/testing_util/timer.cc#L29)):*

this seems weird to me. would a NOLINTNEXTLINE above `int rusage_who()` be possible/preferable?
</blockquote></details>

AFAIK, that disables clang-tidy warnings, but won't work with GCC's `-Wunused-parameter`, and I am not sure it would work with `Clang` `-Wunused-parameter` either.  Both GCC and Clang turn that one with `-Wextra`, which we use btw.  And not that it matters, but it won't work with  with `/W4` and MSVC.  Happy to use another portable way to suppress the warning, if you know of one.

google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc
```diff
@@ -215,10 +218,12 @@ int main(int argc, char* argv[]) {
   using clock = std::chrono::steady_clock;
   auto const start = clock::now();
   auto const deadline = clock::now() + options->running_time;
```
devbww, line 220: Off topic, but `s/clock::now()/start/` perhaps?

But then again, perhaps all the `clock::now()`/`start`/`deadline` stuff should disappear in favor of `timer.Sample().elapsed_time`.

google/cloud/testing_util/timer.h
```diff
@@ -27,21 +27,28 @@ namespace cloud {
...
+
 class Timer {
  public:
   struct Snapshot {
     std::chrono::microseconds elapsed_time;
     std::chrono::microseconds cpu_time;
   };
 
-  Timer();
+  Timer() : Timer(CpuAccounting::kPerThread) {}
```
devbww, line 42: Do we need a default constructor?

google/cloud/testing_util/timer.cc
```diff
@@ -22,19 +22,21 @@ namespace testing_util {
...
 #if GOOGLE_CLOUD_CPP_HAVE_GETRUSAGE
 namespace {
-int rusage_who() {
+int rusage_who(CpuAccounting accounting) {
 #if GOOGLE_CLOUD_CPP_HAVE_RUSAGE_THREAD
-  return RUSAGE_THREAD;
+  return accounting == CpuAccounting::kPerThread ? RUSAGE_THREAD : RUSAGE_SELF;
 #else
+  (void)accounting;
   return RUSAGE_SELF;
```
devbww, line 30: Opt:  Perhaps @dbolduc's comment would be addressed with something like:
```
switch (accounting) {
  case CpuAccounting::kPerThread:
#if GOOGLE_CLOUD_CPP_HAVE_RUSAGE_THREAD
    return RUSAGE_THREAD;
#else
    return RUSAGE_SELF;
#endif
  case CpuAccounting::kPerProcess:
    return RUSAGE_SELF;
}
```

coryan: I like that better, with some adjustments to make the compiler(s) happy.

google/cloud/testing_util/timer.h
```diff
@@ -27,21 +27,29 @@ namespace cloud {
 inline namespace GOOGLE_CLOUD_CPP_NS {
 namespace testing_util {
 
+enum class CpuAccounting {
```
devbww, line 30: Perhaps this is private now.

google/cloud/testing_util/timer.cc
```diff
@@ -22,19 +22,22 @@ namespace testing_util {
...
 #if GOOGLE_CLOUD_CPP_HAVE_GETRUSAGE
 namespace {
-int rusage_who() {
-#if GOOGLE_CLOUD_CPP_HAVE_RUSAGE_THREAD
-  return RUSAGE_THREAD;
-#else
+int rusage_who(CpuAccounting accounting) {
+  switch (accounting) {
+    case CpuAccounting::kPerThread:
+      return RUSAGE_THREAD;
```
devbww, line 28: Does this imply `RUSAGE_THREAD` is available on all `getrusage()` platforms (i.e., no need for `GOOGLE_CLOUD_CPP_HAVE_RUSAGE_THREAD`)?

coryan: Nope, I just forgot about macOS.  Will send a fix in a jiffy.

google/cloud/testing_util/timer.h
coryan: That requires moving `rusage_who()` to the class too, which I thought was a good idea.

coryan:
TIL: the compiler can figure out if you don't use a member variable, wow.

coryan:
PTAL.

devbww (approved):
> Reviewed 5 of 7 files at r2, 1 of 2 files at r3, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6941)* status: all files reviewed, 3 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[google/cloud/testing_util/timer.cc, line 34 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6941#-Me1OzzS4mHt6wjIaKwf:-Me1OzzS4mHt6wjIaKwg:b-99ypxz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce4845cb7641111f5ee4b54316f4624955fcb7db/google/cloud/testing_util/timer.cc#L34)):*
> ```cpp
>       std::chrono::steady_clock::now() - start_);
> #if !GOOGLE_CLOUD_CPP_HAVE_GETRUSAGE
>   return Snapshot{elapsed, std::chrono::microseconds()};
> ```

Nitty nit:  We're careful to `std::move(elapsed)` below.

PR #6958, opened 2021-07-08
test(spanner): use text protos to test admin-connection APIs

devbww:
Use full text protos, which set every field, to test the Spanner
GetDatabase/ListDatabases and GetInstanceConfig/ListInstanceConfigs
admin-connection APIs, rather than only individually assigning and
checking certain fields.  While these APIs are quite thin, this
gives us a little more confidence that we're not missing anything.

It also allows us to add new fields without changing the control
flow in the test.  That was the primary goal in this case, so as
to get coverage for the new "configurable leader placement" fields
`google.spanner.admin.database.v1.Database.default_leader` and
`google.spanner.admin.instance.v1.InstanceConfig.leader_options`.

In addition, the `GetDatabaseWithEncryption` test becomes redundant.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6958)

### DIFF
--- a/google/cloud/spanner/database_admin_connection_test.cc
+++ b/google/cloud/spanner/database_admin_connection_test.cc
@@ -180,63 +180,44 @@ TEST(DatabaseAdminConnectionTest, HandleCreateDatabaseError) {
 }
 
 /// @test Verify that the successful case works.
-TEST(DatabaseAdminConnectionTest, GetDatabaseSuccess) {
-  auto mock = std::make_shared<MockDatabaseAdminStub>();
-  std::string const database_name =
-      "projects/test-project/instances/test-instance/databases/test-database";
+TEST(DatabaseAdminConnectionTest, GetDatabase) {
+  auto constexpr kResponseText = R"pb(
+    name: "projects/project/instances/instance/databases/database"
+    state: READY
+    create_time { seconds: 1625696199 nanos: 123456789 }
+    restore_info {
+      source_type: BACKUP
+      backup_info {
+        backup: "projects/project/instances/instance/backups/backup"
+        create_time { seconds: 1625696099 nanos: 987564321 }
+        source_database: "projects/project/instances/instance/databases/database"
+        version_time { seconds: 1625696099 nanos: 987564321 }
+      }
+    }
+    encryption_config {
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/database_admin_connection_test.cc
```diff
@@ -436,55 +417,167 @@ TEST(DatabaseAdminConnectionTest, UpdateDatabaseGetOperationError) {
 
 /// @test Verify that we can list databases in multiple pages.
 TEST(DatabaseAdminConnectionTest, ListDatabases) {
-  auto mock = std::make_shared<MockDatabaseAdminStub>();
-  Instance in("test-project", "test-instance");
-  std::string const expected_parent = in.FullName();
+  gcsa::Database expected_databases[5];
+  ASSERT_TRUE(TextFormat::ParseFromString(
+      R"pb(
```
coryan, line 422: FYI, some old compilers get confused when using raw strings in the `ASSERT_TRUE` macros, that is why we use a separate `auto constexpr kText = R"pb( ... )pb"` variable.

google/cloud/spanner/database_admin_connection_test.cc
```diff
@@ -436,55 +417,169 @@ TEST(DatabaseAdminConnectionTest, UpdateDatabaseGetOperationError) {
 
 /// @test Verify that we can list databases in multiple pages.
 TEST(DatabaseAdminConnectionTest, ListDatabases) {
-  auto mock = std::make_shared<MockDatabaseAdminStub>();
-  Instance in("test-project", "test-instance");
+  constexpr char const* kDatabaseText[5] = {
```
coryan, line 420: nit: here and below, I think east const spelling would be `char const* constexpr` ?

coryan (approved):

google/cloud/spanner/database_admin_connection_test.cc
devbww:
```
google/cloud/spanner/database_admin_connection_test.cc:420:15: error: expected unqualified-id                                                                   
  char const* constexpr kDatabaseText[5] = {
              ^
```

coryan: TIL, thanks for checking.

Issue #6946, opened 2021-07-07
Use Spanner's base64 encoding/decoding in storage


coryan:
We have two sets of Base64 encoding/decoding functions, some in Spanner:

https://github.com/googleapis/google-cloud-cpp/blob/569e1dc7f2073df76a27d0a82759ca718904d62b/google/cloud/spanner/bytes.h#L159-L160

and some in Storage:

https://github.com/googleapis/google-cloud-cpp/blob/569e1dc7f2073df76a27d0a82759ca718904d62b/google/cloud/storage/internal/openssl_util.h#L31-L39

We should refactor the implementation in spanner to `google/cloud/internal` and use those in the storage library. Using OpenSSL for this is too complicated, and requires error handling for the OpenSSL abstractions.

I think this will remove some of the exceptions mentioned in #6933

PR #6963, opened 2021-07-09
fix(GCS+gRPC): no hashes in Object resource

### LINKED ISSUE #4156: Implement CRC32C checksum in GrpcResumableUploadSession
With gRPC we can compute the checksum as data is streamed and send it with the last message to finish the upload. That provides better data integrity as the upload will not create any object if the checksum is mismatched.

coryan:
The proto file says that neither the `md5_hash` nor the `crc32c` fields
should be set when uploading an `Object`:

https://github.com/googleapis/googleapis/blob/1eade8161f2ef102282869bc3eb18841807b2b46/google/storage/v1/storage_resources.proto#L691-L694
https://github.com/googleapis/googleapis/blob/1eade8161f2ef102282869bc3eb18841807b2b46/google/storage/v1/storage_resources.proto#L676-L679

Also improved the documentation on how to obtain these hashes for tests.

I disabled some tests that used to pass because the service does not ignore
the hashes in the `Object` resource. The documentation says it does (see
the links), so my tests were passing by accident and the service can change
at any time.  Once I fix #4156 and #4157 these tests can be enabled again.

Part of the work for #6946, and will eventually help with #4156 and #4157

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6963)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -784,14 +784,6 @@ void SetResourceOptions(google::storage::v1::Object& resource,
     resource.set_content_type(
         request.template GetOption<ContentType>().value());
   }
-
-  if (request.template HasOption<Crc32cChecksumValue>()) {
-    resource.mutable_crc32c()->set_value(GrpcClient::Crc32cToProto(
-        request.template GetOption<Crc32cChecksumValue>().value()));
-  }
-  if (request.template HasOption<MD5HashValue>()) {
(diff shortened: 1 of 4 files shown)


scotthart (approved):
> Reviewed 2 of 2 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6963)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #6979, opened 2021-07-13
test(spanner): troubleshoot backup timeouts

coryan:
Part of the work for #4306

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6979)

### DIFF
--- a/google/cloud/spanner/integration_tests/backup_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/backup_integration_test.cc
@@ -23,6 +23,7 @@
 #include "google/cloud/internal/absl_str_cat_quiet.h"
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/internal/random.h"
+#include "google/cloud/log.h"
 #include "google/cloud/testing_util/integration_test.h"
 #include "google/cloud/testing_util/status_matchers.h"
 #include "absl/time/time.h"
@@ -38,6 +39,7 @@ namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
 namespace {
 
+using fs = std::future_status;
 using ::google::cloud::testing_util::IsOk;
 using ::google::cloud::testing_util::StatusIs;
 using ::testing::HasSubstr;
@@ -104,6 +106,11 @@ TEST_F(BackupTest, BackupTest) {
 
   // Cancel the CreateBackup operation.
   backup_future.cancel();
+  // TODO(#4306) - CreateBackup() times out (rarely) log for troubleshoot
+  if (backup_future.wait_for(std::chrono::minutes(5)) != fs::ready) {
+    GCP_LOG(WARNING) << "Potential timeout in CreateBackup()" << db;
+    google::cloud::LogSink::Instance().Flush();
+  }
   auto cancelled_backup = backup_future.get();
   if (cancelled_backup) {
     EXPECT_STATUS_OK(database_admin_client_.DeleteBackup(*cancelled_backup));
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/integration_tests/backup_integration_test.cc
```diff
@@ -130,6 +137,11 @@ TEST_F(BackupTest, BackupTest) {
       << "Database " << db.database_id()
       << " not found in the backup operation list.";
 
+  // TODO(#4306) - CreateBackup() times out (rarely) log for troubleshoot
+  if (backup_future.wait_for(std::chrono::minutes(5)) != fs::ready) {
+    GCP_LOG(WARNING) << "Potential timeout in CreateBackup()" << db;
```
devbww, line 142: I'm not sure what "potential timeout" messages are buying us.  Sometimes these operations just don't complete in the hour we've allotted.  But, OK.

devbww, line 142: Nit: Could probably do with a space between `CreateBackup()` and the database name (et seq).

google/cloud/spanner/integration_tests/backup_integration_test.cc
```diff
@@ -424,10 +448,15 @@ TEST_F(BackupTest, BackupTestWithCMEK) {
...
-          .get();
+  auto pending_create = database_admin_client_.CreateDatabase(
+      db, /*extra_statements=*/{}, encryption_config);
+
+  // TODO(#4306) - this test times out (rarely) additional logs to troubleshoot
+  if (pending_create.wait_for(std::chrono::minutes(5)) != fs::ready) {
+    GCP_LOG(WARNING) << "Potential timeout in CreateDatabase()" << in;
+    google::cloud::LogSink::Instance().Flush();
+  }
+  auto database = pending_create.get();
```
devbww, line 459: Have we ever seen `CreateDatabase()` taking a long time?

devbww (approved):

google/cloud/spanner/integration_tests/backup_integration_test.cc
coryan: I am not sure I follow.  Are you implying that these flakes are "normal" and therefore we should ignore them?

If so, I would be glad to change the tests to say something like "this operation is taking so long that we will simply declare success on this run and move on".  The downside if that these timeouts become "common", then we would not notice.

I am actually more interested in the output from the next line, where the last N RPCs would be logged. I am hoping those will help us identify if there was something particular about the instance / database / environment that triggered the timeout.

coryan: Fixed.

google/cloud/spanner/integration_tests/backup_integration_test.cc
coryan: Ooops, I changed my mind about what needed extra logging and did not revert this change. Fixed.

coryan:
PTAL.  If you remain unconvinced by my rationale I will just close the PR.

google/cloud/spanner/integration_tests/backup_integration_test.cc
coryan: Ack. Closing.

devbww: It certainly might be nice to know we were actually in a `CreateBackup()` poll loop when the timeout happened, but I'm unaware of any hook we could use to dump the logging buffer when that happens.

PR #6983, opened 2021-07-14
cleanup(storage): simplify ObjectReadStreambuf

coryan:
Simplify the implementation of `storage::internal::ObjectReadStreambuf`.
The class had two different paths to get more data from the underlying
`internal::ObjectReadSource`, this change makes all reads through the
`xsgetn()` function.

In addition, the class was sloppy with the hash validator, potentially
using it after it was moved-from, luckily this was harmless, but no
reason to keep doing so.

I added some integration tests to cover corner cases that were only
detected by accident in other tests.

Part of the work for #4156 and #4157

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6983)

### DIFF
--- a/google/cloud/storage/internal/object_read_streambuf.cc
+++ b/google/cloud/storage/internal/object_read_streambuf.cc
@@ -113,6 +113,9 @@ void ObjectReadStreambuf::ThrowHashMismatchDelegate(const char* function_name) {
 }
 
 bool ObjectReadStreambuf::ValidateHashes(char const* function_name) {
+  // This function is called once the stream is "closed" (either an explicit
+  // `Close()` call or a permanent error). After this point the validator is
+  // not usable.
   auto validator = std::move(hash_validator_);
   hash_validator_result_ = std::move(*validator).Finish();
   if (!hash_validator_result_.is_mismatch) return true;
--- a/google/cloud/storage/internal/object_read_streambuf.cc
+++ b/google/cloud/storage/internal/object_read_streambuf.cc
@@ -65,135 +65,6 @@ void ObjectReadStreambuf::Close() {
   }
 }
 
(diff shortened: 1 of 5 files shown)


google/cloud/storage/internal/object_read_streambuf.cc
```diff
@@ -65,100 +65,93 @@ void ObjectReadStreambuf::Close() {
...
+  // obviously not ideal, but it is the best we can do when the application
+  // disables the standard mechanism to signal errors.
   throw HashMismatchError(msg, received_hash(), computed_hash());
-#else
-  return traits_type::eof();
 #endif  // GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
 }
 
+bool ObjectReadStreambuf::ValidateHashes(char const* function_name) {
+  auto validator = std::move(hash_validator_);
```
devjgm, line 116: It's interesting that calling `ValidateHashes()` "consumes" the `hash_validator_` member. I assume there's some guarantee about the way that this code (or streams in general) work that ensures that this function cannot be called twice or that any other functions that use `hash_validator_` can be called after this function. If there's a link to some docs that explain this it may be worth including here. Otherwise, even a small comment like "// `hash_validator_` cannot be used after this function because ...".

devjgm (approved):

google/cloud/storage/internal/object_read_streambuf.cc
coryan: Done. FWIW, we were already consuming `*hash_validator_`.  The call to `std::move(*hash_validator_).Finish()` effectively made the validator unusable, it was just harder to see.

devjgm: I see that now. The code is indeed clearer now. And thanks for the comment.

devjgm (approved):

PR #6987, opened 2021-07-14
cleanup: make user agent prefixes conform

dbolduc:
Our different libraries had slightly different user agent prefixes. This makes them more similar and removes some duplicate code.

This does change functionality slightly for Bigtable, Storage... but I think it is for the better. Correct me if my assumption is wrong.

Also in storage, we use `'/'` as a delimiter instead of `' '`.  `' '` seems more consistent to me, but I did not change it. I would be happy to change it though.
https://github.com/googleapis/google-cloud-cpp/blob/575c671c3e52352035f663e84570c3b8d6e29768/google/cloud/storage/client_options.h#L182-L190

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6987)


google/cloud/bigtable/client_options.cc
```diff
@@ -93,9 +93,9 @@ ClientOptions& ClientOptions::set_connection_pool_size(std::size_t size) {
 }
 
 std::string ClientOptions::UserAgentPrefix() {
-  std::string agent = "gcloud-cpp/" + version_string() + " " +
-                      google::cloud::internal::compiler();
-  return agent;
+  static auto const kUserAgentPrefix =
+      google::cloud::internal::UserAgentPrefix();
```
devbww, line 97: These ones seem to break the "no static-duration objects unless they are trivially destructible" convention.

devjgm: The common way to avoid the issue that @devbww mentions is to heap-allocate the object so that the static thing is only a pointer or reference. Like:

```cc
  static auto const* kUserAgentPrefix = new auto(google::cloud::internal::UserAgentPrefix());
  return *kUserAgentPrefix;
```

coryan:
FWIW, fixing storage would be nice. If you do not have time, feel free to open a separate bug and I will look into it.

google/cloud/bigtable/client_options.cc
dbolduc: I decided to move the static optimization into the internal function.... although now I am noticing that I used `static auto* const` instead of `static auto const*` as @devjgm suggested...

dbolduc:
> FWIW, fixing storage would be nice. If you do not have time, feel free to open a separate bug and I will look into it.

Done.

google/cloud/bigtable/client_options.cc
devjgm: I think we want both consts because the pointer and the thing it points to are const.

dbolduc: Done?

devjgm (approved):

PR #6991, opened 2021-07-14
feat(GCS+gRPC): option to configure plugin

coryan:
Make it easier to experiment with configuration options for the GCS+gRPC
plugin. This is useful in benchmarking, and might be useful in the final
form of the plugin.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991)

### DIFF
--- a/google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc
+++ b/google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc
@@ -142,8 +142,10 @@ int main(int argc, char* argv[]) {
     auto channels = options->grpc_channel_count;
     if (channels == 0) channels = (std::max)(options->thread_count / 4, 4);
     client = DefaultGrpcClient(
-        google::cloud::Options{}.set<google::cloud::GrpcNumChannelsOption>(
-            channels));
+        google::cloud::Options{}
+            .set<google::cloud::GrpcNumChannelsOption>(channels)
+            .set<google::cloud::storage_experimental::GrpcPluginOption>(
+                options->grpc_plugin_config));
   }
 #endif  // GOOGLE_CLOUD_CPP_STORAGE_HAVE_GRPC
   std::vector<gcs::ObjectMetadata> objects;
--- a/google/cloud/storage/benchmarks/aggregate_throughput_options.cc
+++ b/google/cloud/storage/benchmarks/aggregate_throughput_options.cc
@@ -68,6 +68,11 @@ ParseAggregateThroughputOptions(std::vector<std::string> const& argv,
        [&options](std::string const& val) {
          options.grpc_channel_count = std::stoi(val);
        }},
+      {"--grpc-plugin-config",
+       "low-level experimental settings for the GCS+gRPC plugin",
+       [&options](std::string const& val) {
+         options.grpc_plugin_config = val;
+       }},
   };
   auto usage = BuildUsage(desc, argv[0]);
 
--- a/google/cloud/storage/benchmarks/aggregate_throughput_options.h
+++ b/google/cloud/storage/benchmarks/aggregate_throughput_options.h
@@ -34,6 +34,7 @@ struct AggregateThroughputOptions {
   std::size_t read_buffer_size = 4 * kMiB;
   ApiName api = ApiName::kApiGrpc;
   int grpc_channel_count = 0;
+  std::string grpc_plugin_config;
   bool exit_after_parse = false;
 };
 
--- a/google/cloud/storage/benchmarks/aggregate_throughput_options_test.cc
+++ b/google/cloud/storage/benchmarks/aggregate_throughput_options_test.cc
@@ -35,6 +35,7 @@ TEST(AggregateThroughputOptions, Basic) {
           "--read-buffer-size=1MiB",
           "--api=XML",
           "--grpc-channel-count=16",
+          "--grpc-plugin-config=default",
       },
       "");
   ASSERT_STATUS_OK(options);
--- a/google/cloud/storage/grpc_plugin.h
+++ b/google/cloud/storage/grpc_plugin.h
@@ -31,6 +31,30 @@ namespace storage_experimental {
 inline namespace STORAGE_CLIENT_NS {
 
 /**
+ * Low-level experimental settings for the GCS+gRPC plugin.
+ *
+ * Possible values for the string include:
+ *
+ * - "default" or "none": do not use any special settings with gRPC
+ * - "dp": equivalent to setting both "pick-first-lb" and
+ *   "enable-dns-srv-queries"
+ * - "alts": same settings as "dp", but use the experimental ALTS credentials
+ * - "enable-dns-srv-queries": set the `grpc.dns_enable_srv_queries` channel
+ *   argument to `1`, see [dns-query-arg].
+ * - "disable-dns-srv-queries": set the `grpc.dns_enable_srv_queries` channel
+ *   argument to `0`, see [dns-query-arg].
+ * - "exclusive": use a exclusive channel for each stub.
+ *
+ * Unknown values are ignored.
+ *
+ * [dns-query-arg]:
+ * https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#ga247ed6771077938be12ab24790a95732
+ */
+struct GrpcPluginOption {
+  using Type = std::string;
+};
+
+/**
  * Create a `google::cloud::storage::Client` object configured to use gRPC.
  *
  * @note the Credentials parameter in the configuration is ignored. The gRPC
--- a/google/cloud/storage/grpc_plugin.h
+++ b/google/cloud/storage/grpc_plugin.h
@@ -35,15 +35,17 @@ inline namespace STORAGE_CLIENT_NS {
  *
  * Possible values for the string include:
  *
- * - "default" or "none": do not use any special settings with gRPC
- * - "dp": equivalent to setting both "pick-first-lb" and
- *   "enable-dns-srv-queries"
- * - "alts": same settings as "dp", but use the experimental ALTS credentials
+ * - "default" or "none": do not use any special settings with gRPC.
+ * - "dp": enable Google Direct Access (formerly 'Direct Path') equivalent to
+ *   setting both "pick-first-lb" and "enable-dns-srv-queries".
+ * - "alts": same settings as "dp", but use the experimental ALTS credentials.
  * - "enable-dns-srv-queries": set the `grpc.dns_enable_srv_queries` channel
  *   argument to `1`, see [dns-query-arg].
  * - "disable-dns-srv-queries": set the `grpc.dns_enable_srv_queries` channel
  *   argument to `0`, see [dns-query-arg].
- * - "exclusive": use a exclusive channel for each stub.
+ * - "pick-first-lb": configure the gRPC load balancer to use the "pick_first"
+ *   policy.
+ * - "exclusive": use an exclusive channel for each stub.
  *
  * Unknown values are ignored.
  *
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/storage/internal/grpc_client.h"
+#include "google/cloud/storage/grpc_plugin.h"
 #include "google/cloud/storage/internal/grpc_object_read_source.h"
 #include "google/cloud/storage/internal/grpc_resumable_upload_session.h"
 #include "google/cloud/storage/internal/openssl_util.h"
@@ -56,14 +57,6 @@ auto constexpr kDirectPathConfig = R"json({
     }]
   })json";
 
-bool DirectPathEnabled() {
-  auto const direct_path_settings =
-      google::cloud::internal::GetEnv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH")
-          .value_or("");
-  return absl::c_any_of(absl::StrSplit(direct_path_settings, ','),
-                        [](absl::string_view v) { return v == "storage"; });
-}
-
 int DefaultGrpcNumChannels() {
   auto constexpr kMinimumChannels = 4;
   auto const count = std::thread::hardware_concurrency();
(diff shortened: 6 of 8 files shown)


dbolduc (changes requested):
> Reviewed 7 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991)* status: 7 of 8 files reviewed, 5 unresolved discussions (waiting on @coryan)

___
*[google/cloud/storage/grpc_plugin.h, line 39 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-MebGYTn7vDUA13_aZ7l:-MebG6aw1WgkxEqxjUWU:b-i1vrvi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/grpc_plugin.h#L39)):*
> ```objc
>  *
>  * - "default" or "none": do not use any special settings with gRPC
>  * - "dp": equivalent to setting both "pick-first-lb" and
> ```

maybe mention DirectPath?
"dp" wasn't obvious to me, but also me is stoopid.
___
*[google/cloud/storage/grpc_plugin.h, line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-Meb_ygI6bVxt8SYIdOn:-Meb_ygI6bVxt8SYIdOo:bu3hs3q) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/grpc_plugin.h#L44)):*
> ```objc
>  * - "enable-dns-srv-queries": set the `grpc.dns_enable_srv_queries` channel
>  *   argument to `1`, see [dns-query-arg].
>  * - "disable-dns-srv-queries": set the `grpc.dns_enable_srv_queries` channel
> ```

I'm not sure why we need both enable and disable? I'd assume that not setting enable implies disable.
___
*[google/cloud/storage/grpc_plugin.h, line 46 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-MebtBof2X5tpS7XJa67:-MebtBogCPHuUmgy07zr:bd997n7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/grpc_plugin.h#L46)):*
> ```objc
>  * - "disable-dns-srv-queries": set the `grpc.dns_enable_srv_queries` channel
>  *   argument to `0`, see [dns-query-arg].
>  * - "exclusive": use a exclusive channel for each stub.
> ```

s/a exclusive/an exclusive/
___
*[google/cloud/storage/grpc_plugin.h, line 53 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-MecE9ZA0TvN8ZRoGkoa:-MecE9ZA0TvN8ZRoGkob:b-kufeen) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/grpc_plugin.h#L53)):*
> ```objc
>  * https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#ga247ed6771077938be12ab24790a95732
>  */
> struct GrpcPluginOption {
> ```

Given that this is in an experimental namespace, I think it's fine. But I am not sure this is how we want to deal with `grpc::ChannelArguments` in the long term across all of our libraries.

I am currently dealing with this in Bigtable, but I do not yet have a plan on how to proceed. So let me think out loud here for a bit:

I think the main trade off is going to be between dealing with options that are very specific to each library (like this one) or having very general options that all libraries can use (like `GrpcChannelArgumentsOption`).

* The plus of having general options living in common is that libraries can reuse it easily. For example, DirectPath will need to be implemented in Bigtable and Spanner. Wouldn't it be nice if there was only 1 implementation instead of 3? It is not obvious to me how this implementation can be applied to the other libraries other than copy-paste.

* The minus is that having a bunch of options in common can lead to clutter, making it more confusing for the users. There might be a channel argument that is only pertinent to one library, although it could be set for others. In this case, should the `Option` belong only in the library that uses it? I am leaning yes, because I think it is on us to provide the proper guidance for how to use these settings. I would like to know what you think though.

So how do we deal with channel arguments? Maybe I should start with what I don't think we should do. I don't think we should have one option to rule them all that has `Type = grpc::ChannelArguments`. While this would make our code maximally simple, it would require our users to know the exact right settings. This is clearly a deal breaker.

I also don't think we want an option with `Type = std::map<std::string, int>` (similar to what we do now with [`GrpcChannelArgumentsOption`](https://github.com/googleapis/google-cloud-cpp/blob/15a0d4787efbfaea3f9d0c23bc389af039be0474/google/cloud/grpc_options.h#L58-L60). It would make it easy to [convert](https://github.com/googleapis/google-cloud-cpp/blob/15a0d4787efbfaea3f9d0c23bc389af039be0474/google/cloud/grpc_options.cc#L26) `Options` <----> `grpc::ChannelArguments` but again, the generality requires the users to know the key of the gRPC argument to set. Also, it's actually [not recommended](https://github.com/grpc/grpc/blob/master/include/grpcpp/support/channel_arguments.h#L95-L97) to set custom integer values (although it is possible). Using strings is the recommended way for users to differentiate channels.

I think the other thing to consider is whether the individual channel arguments should be broken out into their own options. Or if they can be consolidated. (e.g. a `DirectPathOption` might set the int `GRPC_ARG_DNS_ENABLE_SRV_QUERIES` and the string `GRPC_ARG_SERVICE_CONFIG`). The logic in the code might be simpler if the `Options` were logically independent... But I think the users would benefit from simple, consolidated `Options`. (What you have done putting a bunch of these options into one string though.... that might be too far for me. If it is just us using it, it's fine. If it's the users.... I'd like to see this broken up).

What I propose (which you might already be doing/thinking) is for each library to define the channel arguments they need, packaged as they see fit, in an experimental namespace. Then when it's all laid out, we can see what is shared across multiple libraries that would benefit from being moved into common. Maybe there is a way for each library to build off of [this method](https://github.com/googleapis/google-cloud-cpp/blob/15a0d4787efbfaea3f9d0c23bc389af039be0474/google/cloud/grpc_options.cc#L26) to deal with its unique options.

(cover your eyes now if you don't want to read about Bigtable)

We cannot yet go from `grpc::ChannelArguments` --> `Options`. I need to do this in order to implement [this method](https://github.com/googleapis/google-cloud-cpp/blob/15a0d4787efbfaea3f9d0c23bc389af039be0474/google/cloud/bigtable/client_options.h#L184) in bigtable using `Options`. While this would certainly benefit from having a general option type like `Type = std::map<std::string, int>` I do not think it should have any impact on our design moving forward. We should not accommodate for backwards compatibility of `ClientOptions` (which we wish to stop using). I think the code will require a gnarly `switch` statement...... but so be it.
___
*[google/cloud/storage/internal/grpc_client_test.cc, line 20 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-MebXb2-9eEc6KzvSzSh:-MebXb2-9eEc6KzvSzSi:b-3tf711) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/internal/grpc_client_test.cc#L20)):*
> ```cpp
> #include "google/cloud/storage/internal/object_metadata_parser.h"
> #include "google/cloud/testing_util/is_proto_equal.h"
> #include "google/cloud/testing_util/scoped_environment.h"
> ```

@devbww points out that this include is no longer needed

coryan (commented):
PTAL

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-:-MedvSrX9GoMwrqvgk0L:bb71pro)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @dbolduc and @devbww)

___
*[google/cloud/storage/grpc_plugin.h, line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-Meb_ygI6bVxt8SYIdOn:-Medu7DWEYEb8vd90627:b-68zs9u) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/grpc_plugin.h#L44)):*

I'm not sure why we need both enable and disable? I'd assume that not setting enable implies disable.
</blockquote></details>

I want to be able to say `alts,disable-dns-srv-queries` or just `enable-dns-srv-queries`.  And I do not know what is the default for `grpc.dns_enable_srv_queries` so I want to be able to swich away from the default.  The point is: I do not know which one of these options have performance effects, I want to be able to tweak them.
___
*[google/cloud/storage/grpc_plugin.h, line 46 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-MebtBof2X5tpS7XJa67:-MedvBdH92xJA9XLuKOD:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/grpc_plugin.h#L46)):*

s/a exclusive/an exclusive/
</blockquote></details>

Done.
___
*[google/cloud/storage/grpc_plugin.h, line 53 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-MecE9ZA0TvN8ZRoGkoa:-MedvCabAuwV4p65o0e6:b-c401kl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/grpc_plugin.h#L53)):*

Given that this is in an experimental namespace, I think it's fine. But I am not sure this is how we want to deal with `grpc::ChannelArguments` in the long term across all of our libraries.

I am currently dealing with this in Bigtable, but I do not yet have a plan on how to proceed. So let me think out loud here for a bit:

I think the main trade off is going to be between dealing with options that are very specific to each library (like this one) or having very general options that all libraries can use (like `GrpcChannelArgumentsOption`).

* The plus of having general options living in common is that libraries can reuse it easily. For example, DirectPath will need to be implemented in Bigtable and Spanner. Wouldn't it be nice if there was only 1 implementation instead of 3? It is not obvious to me how this implementation can be applied to the other libraries other than copy-paste.

* The minus is that having a bunch of options in common can lead to clutter, making it more confusing for the users. There might be a channel argument that is only pertinent to one library, although it could be set for others. In this case, should the `Option` belong only in the library that uses it? I am leaning yes, because I think it is on us to provide the proper guidance for how to use these settings. I would like to know what you think though.

So how do we deal with channel arguments? Maybe I should start with what I don't think we should do. I don't think we should have one option to rule them all that has `Type = grpc::ChannelArguments`. While this would make our code maximally simple, it would require our users to know the exact right settings. This is clearly a deal breaker.

I also don't think we want an option with `Type = std::map<std::string, int>` (similar to what we do now with [`GrpcChannelArgumentsOption`](https://github.com/googleapis/google-cloud-cpp/blob/15a0d4787efbfaea3f9d0c23bc389af039be0474/google/cloud/grpc_options.h#L58-L60). It would make it easy to [convert](https://github.com/googleapis/google-cloud-cpp/blob/15a0d4787efbfaea3f9d0c23bc389af039be0474/google/cloud/grpc_options.cc#L26) `Options` <----> `grpc::ChannelArguments` but again, the generality requires the users to know the key of the gRPC argument to set. Also, it's actually [not recommended](https://github.com/grpc/grpc/blob/master/include/grpcpp/support/channel_arguments.h#L95-L97) to set custom integer values (although it is possible). Using strings is the recommended way for users to differentiate channels.

I think the other thing to consider is whether the individual channel arguments should be broken out into their own options. Or if they can be consolidated. (e.g. a `DirectPathOption` might set the int `GRPC_ARG_DNS_ENABLE_SRV_QUERIES` and the string `GRPC_ARG_SERVICE_CONFIG`). The logic in the code might be simpler if the `Options` were logically independent... But I think the users would benefit from simple, consolidated `Options`. (What you have done putting a bunch of these options into one string though.... that might be too far for me. If it is just us using it, it's fine. If it's the users.... I'd like to see this broken up).

What I propose (which you might already be doing/thinking) is for each library to define the channel arguments they need, packaged as they see fit, in an experimental namespace. Then when it's all laid out, we can see what is shared across multiple libraries that would benefit from being moved into common. Maybe there is a way for each library to build off of [this method](https://github.com/googleapis/google-cloud-cpp/blob/15a0d4787efbfaea3f9d0c23bc389af039be0474/google/cloud/grpc_options.cc#L26) to deal with its unique options.

(cover your eyes now if you don't want to read about Bigtable)

We cannot yet go from `grpc::ChannelArguments` --> `Options`. I need to do this in order to implement [this method](https://github.com/googleapis/google-cloud-cpp/blob/15a0d4787efbfaea3f9d0c23bc389af039be0474/google/cloud/bigtable/client_options.h#L184) in bigtable using `Options`. While this would certainly benefit from having a general option type like `Type = std::map<std::string, int>` I do not think it should have any impact on our design moving forward. We should not accommodate for backwards compatibility of `ClientOptions` (which we wish to stop using). I think the code will require a gnarly `switch` statement...... but so be it.
</blockquote></details>

Ack.  Do you want to start a design doc?
___
*[google/cloud/storage/internal/grpc_client_test.cc, line 20 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-MebXb2-9eEc6KzvSzSh:-MedvGzmCB0FXzStw5CP:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/internal/grpc_client_test.cc#L20)):*

@devbww points out that this include is no longer needed
</blockquote></details>

Done.

dbolduc (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @coryan, @dbolduc, and @devbww)

___
*[google/cloud/storage/grpc_plugin.h, line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6991#-Meb_ygI6bVxt8SYIdOn:-MeeQyr20GEXldVmSINC:b-a9qhvh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/15856cc2fb4661f17fd26a3b84914a30f46ee999/google/cloud/storage/grpc_plugin.h#L44)):*

I want to be able to say `alts,disable-dns-srv-queries` or just `enable-dns-srv-queries`.  And I do not know what is the default for `grpc.dns_enable_srv_queries` so I want to be able to swich away from the default.  The point is: I do not know which one of these options have performance effects, I want to be able to tweak them.
</blockquote></details>

understood. I believe the default is `false`, set here:
https://github.com/grpc/grpc/blob/434f90b5e571d4989675547434f5daaf47ba3a26/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L139-L140

but I am not certain.

dbolduc (approved):

PR #6994, opened 2021-07-15
feat(storage): capture peer addres for REST

coryan:
This is most useful in benchmarks, but could be handy
when troubleshooting connection problems. Specially
if proxies or VPC-SC are involved.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6994)

### DIFF
--- a/google/cloud/storage/internal/curl_handle.h
+++ b/google/cloud/storage/internal/curl_handle.h
@@ -119,14 +119,11 @@ class CurlHandle {
     return AsStatus(e, __func__);
   }
 
-  StatusOr<long> GetResponseCode() {  // NOLINT(google-runtime-int)
-    long code;                        // NOLINT(google-runtime-int)
-    auto e = curl_easy_getinfo(handle_.get(), CURLINFO_RESPONSE_CODE, &code);
-    if (e == CURLE_OK) {
-      return code;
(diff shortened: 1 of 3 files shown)


google/cloud/storage/internal/curl_handle.h
```diff
@@ -119,14 +119,11 @@ class CurlHandle {
...
-    auto e = curl_easy_getinfo(handle_.get(), CURLINFO_RESPONSE_CODE, &code);
-    if (e == CURLE_OK) {
-      return code;
-    }
-    return AsStatus(e, __func__);
-  }
+  /// Get the HTTP response code, or an error.
+  StatusOr<std::int32_t> GetResponseCode();
+
+  /// Get the peer, or just "unknown".
```
devbww, line 125: Q.  Is the `"unknown"` something that `CURLINFO_PRIMARY_IP` provides, or is this the `"[error-fetching-peer]"` case?

devbww (approved):

google/cloud/storage/internal/curl_handle.h
coryan: Thanks, fixed the comments to avoid confusion.

PR #6996, opened 2021-07-15
refactor(storage): split hash functions vs. validation

coryan:
For the GCS+gRPC plugin we need to split the hash computation from its
validation. With GCS+gRPC we can send the hashes in the last chunk,
which means we cannot do the validation at the same time (we get the
server-side hashes later). Furthermore, for just validation having all
the hashes in a single string was a neat trick, but now we need to
change their format before uploading. Better to separate the
computation, and to keep the hashes in a data structure that is easier
to operate with.

Part of the work for #4156 and for #4157

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6996)


google/cloud/storage/internal/hash_values.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "google/cloud/internal/absl_str_cat_quiet.h"
+
+namespace google {
+namespace cloud {
+namespace storage {
+inline namespace STORAGE_CLIENT_NS {
+namespace internal {
+
+std::string Format(HashValues const& values) {
+  if (values.md5.empty()) return values.crc32c;
```
devjgm, line 25: Do you think it'd be useful to always include the "crc32c=" and "md5=" prefixes in the output, even if one is empty? It seems like it may be useful to know which one is being printed. Consider always formatting using the `StrCat` call on line 27.

devjgm (approved):

google/cloud/storage/internal/hash_values.cc
coryan: These strings are returned by user-visible functions (see `ObjectReadStream::received_hash()/computed_hash()` and their counterparts in `ObjectWriteStream`).  I did not want to change their format without more consideration about any backwards compatibility implications.  I mean, the change you suggest may be harmless or not break anybody, I just wanted to avoid the extra thinking.

Issue #6832, opened 2021-06-22
[Flake] cannot start GCS emulator


devjgm:
GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/c2cb00f4-aca1-4355-95e2-2059fd49e4f4;tab=detail?project=cloud-cpp-testing-resources
Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/54f4b5cc564e656d059211ae71982f080c8f5fdc/fedora-34-tsan/log-c2cb00f4-aca1-4355-95e2-2059fd49e4f4.txt

```
Step #3: INFO: Build completed successfully, 8 total actions
Step #3: [ -------- ] Integration test environment set-up
Step #3: Launching Cloud Storage emulator in the background
Step #3: Cannot connect to emulator; aborting test.
Step #3: [2021-06-22 12:14:45 +0000] [26827] [INFO] Starting gunicorn 20.1.0
Step #3: [2021-06-22 12:14:45 +0000] [26827] [INFO] Listening at: http://0.0.0.0:8585 (26827)
Step #3: [2021-06-22 12:14:45 +0000] [26827] [INFO] Using worker: gthread
Step #3: [2021-06-22 12:14:45 +0000] [26835] [INFO] Booting worker with pid: 26835
Step #3: [ -------- ] Integration test environment tear-down.
Step #3: Killing emulator server [26827] ... done.
Step #3: [ ======== ] Integration test environment tear-down.
Step #3: ==> ðŸ•‘ tsan completed in 401.803 seconds
Finished Step #3
ERROR
ERROR: build step 3 "gcr.io/cloud-cpp-testing-resources/cloudbuild/fedora-34:c2cb00f4-aca1-4355-95e2-2059fd49e4f4" failed: step exited with non-zero status: 1
```

coryan:
@scotthart recorded another instance in a duplicate bug that I created:

https://pantheon.corp.google.com/cloud-build/builds;region=us-east1/da3f25f2-9de7-43ed-bac6-5e2280e3a2e3;step=3?project=cloud-cpp-testing-resources

coryan:
I am going to declare this fixed.  We have an explanation and a fix for the flake, and it has not repeated since that fix.

PR #7002, opened 2021-07-16
test: avoid using old emulator logs in startup

coryan:
I think this will deflake the storage builds.  Consider this:

- The script to startup the emulator uses the log file to discover the
  port where it is running.
- The script launches the emulator in the background and then
  immediately tries to read its log
- The script fails with "cannot connect", which it is only attempted
  after the port is found.
- The script then prints an empty emulator log, which could not happen
  if the port was found!

But really what is going on is that the shell calls `fork(2)` to start
the emulator, but before it gets to call `exec(2)`, or before it even
redirects `stdout` the parent process reads the file, finds the *old*
ports, then tries to connect. That fails, of course, then it prints the
log file, but that might be empty, or incomplete, or completely
different.

I got paranoid and fixed all the other scripts to start emulators.

Here is a log showing this:

```
2021-07-16T00:00:13Z (+362s)
---------------------------------------------------------
|   Running Storage integration tests (with emulator)   |
---------------------------------------------------------
2021-07-16T00:00:13Z (+362s): Launching Cloud Storage emulator on port 0
2021-07-16T00:00:13Z (+362s): Cannot connect to emulator; aborting test.
2021-07-16T00:00:13Z (+362s): curl connection test result.
curl: (7) Failed to connect to localhost port 39413: Connection refused
Killing emulator server [25990] ... 2021-07-16T00:00:13Z (+362s): ===> ccache stats
```

Part of the work for #6832

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7002)

### DIFF
--- a/google/cloud/bigtable/tools/run_emulator_utils.sh
+++ b/google/cloud/bigtable/tools/run_emulator_utils.sh
@@ -82,12 +82,10 @@ function start_emulators() {
   trap kill_emulators EXIT
 
   local -r CBT_EMULATOR_CMD="/usr/local/google-cloud-sdk/platform/bigtable-emulator/cbtemulator"
-  rm -f emulator.log
   "${CBT_EMULATOR_CMD}" -port "${emulator_port}" >emulator.log 2>&1 </dev/null &
   EMULATOR_PID=$!
 
   # The path to this command must be set by the caller.
-  rm -f instance-admin-emulator.log
   "${CBT_INSTANCE_ADMIN_EMULATOR_CMD}" "${instance_admin_emulator_port}" >instance-admin-emulator.log 2>&1 </dev/null &
   INSTANCE_ADMIN_EMULATOR_PID=$!
 
--- a/google/cloud/bigtable/tools/run_emulator_utils.sh
+++ b/google/cloud/bigtable/tools/run_emulator_utils.sh
@@ -82,10 +82,12 @@ function start_emulators() {
   trap kill_emulators EXIT
 
   local -r CBT_EMULATOR_CMD="/usr/local/google-cloud-sdk/platform/bigtable-emulator/cbtemulator"
+  rm -f emulator.log
   "${CBT_EMULATOR_CMD}" -port "${emulator_port}" >emulator.log 2>&1 </dev/null &
   EMULATOR_PID=$!
 
   # The path to this command must be set by the caller.
+  rm -f instance-admin-emulator.log
   "${CBT_INSTANCE_ADMIN_EMULATOR_CMD}" "${instance_admin_emulator_port}" >instance-admin-emulator.log 2>&1 </dev/null &
   INSTANCE_ADMIN_EMULATOR_PID=$!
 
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/tools/run_emulator_utils.sh
```diff
@@ -82,10 +82,12 @@ function start_emulators() {
   trap kill_emulators EXIT
 
   local -r CBT_EMULATOR_CMD="/usr/local/google-cloud-sdk/platform/bigtable-emulator/cbtemulator"
+  rm -f emulator.log
```
devbww, line 85: I don't think this can make any difference.  Whether `emulator.log` exists or not here, the next command will `O_CREAT|O_TRUNC` it before `CBT_EMULATOR_CMD` is executed.

google/cloud/pubsub/ci/lib/pubsub_emulator.sh
```diff
@@ -59,6 +59,7 @@ function pubsub_emulator::start() {
   readonly PUBSUB_EMULATOR_ARGS=(
     "beta" "emulators" "pubsub" "start"
     "--project=${GOOGLE_CLOUD_PROJECT:-fake-project}")
+  rm -f emulator.log
   if [[ ! -x "${PUBSUB_EMULATOR_CMD}" ]]; then
     echo 1>&2 "The Cloud Pub/Sub emulator does not seem to be installed, aborting"
     cat emulator.log >&2
```
devbww, line 65: This seems misplaced either way.

coryan: Fixed.

google/cloud/bigtable/tools/run_emulator_utils.sh
coryan: Fixed.

devjgm: I think the key insight in this PR is that the invocation of the emulator is done in the background, and the `emulator.log` file is opened (and truncated) also in that background process. So that creates a race condition with the latter function that reads the `emulator.log` file looking for a port number, since the backgrounded process may not have truncated the log yet. The fix proposed here is to remove the log file before running the emulator in the background.

This seems like a reasonable fix to me, and it seems like it applies here just as much as in the other emulator startup scripts in this PR.

But I apologize if I've misunderstood the comment or situation here.

devbww: I apologize ... I missed the location of the fork.  I agree that the (re)creation of the log file happens in the child process the way things stand, and therefore that removing them first indeed has an effect.

devbww (approved):

devjgm (approved):
You may want to add `google/cloud/bigtable/tools/run_emulator_utils.sh` with similar edits back to this PR.

google/cloud/bigtable/tools/run_emulator_utils.sh
coryan: In the specific case of the **bigtable** emulator I was over enthusiastic and @devbww is right, there is no reading back of the log file, so removing it is not useful.  In the other emulators it seems to matter.

coryan:
Let's merge this and we can talk about whether Bigtable needs it later.

devjgm:
LGTM

PR #7008, opened 2021-07-16
test(pubsub): deflake flow control test

### LINKED ISSUE #7006: [Flake] Pubsub Flow Control times out
Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/70f92a9acc238daa2fc13ddf8dbb37156be8c0ce/fedora-34-libcxx/log-0660cabf-78f0-424d-bcbf-69b96658c8e0.txt
```
Step #3: TIMEOUT: //google/cloud/pubsub:publisher_connection_test (Summary)
Step #3:       /h/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/execroot/com_github_googleapis_google_cloud_cpp/bazel-out/k8-fastbuild/testlogs/google/cloud/pubsub/publisher_connection_test/test.log
Step #3: INFO: From Testing //google/cloud/pubsub:publisher_connection_test:
Step #3: ==================== Test output for //google/cloud/pubsub:publisher_connection_test:
Step #3: Running main() from gmock_main.cc
Step #3: [==========] Running 11 tests from 1 test suite.
Step #3: [----------] Global test environment set-up.
Step #3: [----------] 11 tests from PublisherConnectionTest
Step #3: [ RUN      ] PublisherConnectionTest.Basic
Step #3: E0716 14:56:45.948159349      12 credentials_generic.cc:35]  Could not get HOME environment variable.
Step #3: [       OK ] PublisherConnectionTest.Basic (19 ms)
Step #3: [ RUN      ] PublisherConnectionTest.Metadata
Step #3: E0716 14:56:45.967836880      12 credentials_generic.cc:35]  Could not get HOME environment variable.
Step #3: [       OK ] PublisherConnectionTest.Metadata (26 ms)
Step #3: [ RUN      ] PublisherConnectionTest.Logging
Step #3: E0716 14:56:45.994248927      12 credentials_generic.cc:35]  Could not get HOME environment variable.
Step #3: [       OK ] PublisherConnectionTest.Logging (18 ms)
Step #3: [ RUN      ] PublisherConnectionTest.FlowControl
Step #3: E0716 14:56:46.012879941      12 credentials_generic.cc:35]  Could not get HOME environment variable.
Step #3: -- Test timed out at 2021-07-16 15:01:45 UTC --
```

I was able to reproduce once locally with: 
```
bazel test --nocache_test_results --runs_per_test=100 //google/cloud/pubsub:publisher_connection_test
```
But even that will only fail every couple of runs.

dbolduc:
Fixes #7006 

The issue was that (very rarely) the hold timer would expire and [flush](https://github.com/googleapis/google-cloud-cpp/blob/8db6bac678ab450c4d46fee68d94c94bbc3be94a/google/cloud/pubsub/internal/batching_publisher_connection.cc#L163) the batch of messages before all messages in the test had been published. This would lead to the mock's `AsyncPublish` being called more than once.

This change makes it so that we ack all batches, instead of just [the first batch](https://github.com/googleapis/google-cloud-cpp/blob/8db6bac678ab450c4d46fee68d94c94bbc3be94a/google/cloud/pubsub/publisher_connection_test.cc#L156).

Other fixes considered:
* adding a `bool AsyncSequencer::Empty()`. This seems like it was against the design of the sequencer, which [allows for push and pop to be called in any order](https://github.com/googleapis/google-cloud-cpp/blob/8db6bac678ab450c4d46fee68d94c94bbc3be94a/google/cloud/testing_util/async_sequencer.h#L84). I don't think we can know its size when we pop.
* increasing the hold time: `PublisherOption::set_maximum_hold_time(std::chrono::seconds(1))`. Tests based on timers are generally bad (although this would be no worse than before). Also, the one thing I noticed was that if I set the value to 10s, the test would take 10s. I am wondering if this is a design bug, or if we don't care because the value is supposed to be on the order of microseconds.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7008)

### DIFF
--- a/google/cloud/pubsub/publisher_connection_test.cc
+++ b/google/cloud/pubsub/publisher_connection_test.cc
@@ -155,7 +155,7 @@ TEST(PublisherConnectionTest, FlowControl) {
   EXPECT_THAT(rejected.get(), StatusIs(StatusCode::kFailedPrecondition));
 
   publisher->Flush({});
-  for (int i = 0; i < publish_count; i++) {
+  for (int i = 0; i < publish_count; ++i) {
     publish.PopFront().set_value();
   }
   for (auto& p : pending) {
--- a/google/cloud/pubsub/publisher_connection_test.cc
+++ b/google/cloud/pubsub/publisher_connection_test.cc
@@ -123,11 +123,13 @@ TEST(PublisherConnectionTest, FlowControl) {
   testing_util::ScopedLog log;
 
   AsyncSequencer<void> publish;
+  int publish_count = 0;
   EXPECT_CALL(*mock, AsyncPublish)
       .Times(AtLeast(1))
       .WillRepeatedly([&](google::cloud::CompletionQueue&,
                           std::unique_ptr<grpc::ClientContext>,
                           google::pubsub::v1::PublishRequest const& request) {
+        ++publish_count;
         return publish.PushBack().then([request](future<void>) {
           google::pubsub::v1::PublishResponse response;
           for (auto const& m : request.messages()) {
@@ -153,7 +155,9 @@ TEST(PublisherConnectionTest, FlowControl) {
   EXPECT_THAT(rejected.get(), StatusIs(StatusCode::kFailedPrecondition));
 
   publisher->Flush({});
-  publish.PopFront().set_value();
+  for (int i = 0; i < publish_count; i++) {
+    publish.PopFront().set_value();
+  }
   for (auto& p : pending) {
     EXPECT_THAT(p.get(), IsOk());
   }


google/cloud/pubsub/publisher_connection_test.cc
```diff
@@ -153,7 +155,9 @@ TEST(PublisherConnectionTest, FlowControl) {
   EXPECT_THAT(rejected.get(), StatusIs(StatusCode::kFailedPrecondition));
 
   publisher->Flush({});
-  publish.PopFront().set_value();
+  for (int i = 0; i < publish_count; i++) {
+    publish.PopFront().set_value();
+  }
```
devbww, line 160: Perhaps there should be an `AsyncSequencer::empty()`, removing the need to keep its size externally?

devbww: Oops ... I see you mentioned that explicitly in the description.

google/cloud/pubsub/publisher_connection_test.cc
```diff
@@ -153,7 +155,9 @@ TEST(PublisherConnectionTest, FlowControl) {
   EXPECT_THAT(rejected.get(), StatusIs(StatusCode::kFailedPrecondition));
 
   publisher->Flush({});
-  publish.PopFront().set_value();
+  for (int i = 0; i < publish_count; i++) {
```
devjgm, line 158: Style nit, but JFYI: Prefer preincrement unless you explicitly need post-increment semantics.
https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

devjgm (approved):

google/cloud/pubsub/publisher_connection_test.cc
dbolduc: Fixed.

PR #7025, opened 2021-07-20
fix(storage): use hash values in InsertObject()

### LINKED ISSUE #7014: Audit use of pre-computed hashes for uploads and downloads
After implementing the changes for #4156 and #4157 I am not convinced that the existing uses of `Crc32cChecksumValue()` and `MD5HashValue()` work correctly.

We need to review the existing unit and integration tests, and verify they are passing for the right reasons. If needed, we need to write new integration tests.

Labeled as `cleanup` for now, but if the audit reveals a problem then we should change the type of this issue to a p1 bug.

coryan:
If the application provides a `gcs::Crc32ChecksumValue()` or
`gcs::MD5HashValue()` to `InsertObject()` we should use those values to
create the metadata portion of the upload. Otherwise we are losing an
opportunity to validate the data integrity during the upload.

I discovered this while refactoring and restructuring the integration
tests for CRC32C checksums and MD5 hashes. Apologies for the massive
changes in the tests for what amounts to a small fix.

The tests now have 3 portions, one for `InsertObject()`, one for
`WriteObject()` and one for `ReadObject()`. Unlike before, the code for
XML and JSON tests is shared, though this required some changes in the
emulator to make it easier to detect what hash fields (if any) were
including in the upload.

For `InsertObject()` and `WriteObject()` we test 5 scenarios: (1) using
the default settings, which are enabled for CRC32C, and disabled for
MD5, (2) explicitly disabling the hash, (3) explicitly enabling the
hash, (4) setting the hash to the correct value, and (5) setting the
hash to the incorrect value.

For `WriteObject()` we also test detecting an invalid hash, both in the
case where the library computes the hash and the case where the
application provides it.

For `ReadObject()` it is simpler, because we cannot provide expected
hashes. So we only test the cases where the hash check is in its default
setting (enabled for CRC32, disabled for MD5) and that hash mistmatches
are detected when the hash is enabled.

Fixes #7014

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7025)

### DIFF
--- a/google/cloud/storage/internal/curl_client.cc
+++ b/google/cloud/storage/internal/curl_client.cc
@@ -536,7 +536,9 @@ StatusOr<ObjectMetadata> CurlClient::InsertObjectMedia(
   // uploads. `DisableMD5Hash` and `DisableCrc32cChecksum` should not be
   // dependent on each other.
   if (!request.GetOption<DisableMD5Hash>().value() ||
-      !request.GetOption<DisableCrc32cChecksum>().value_or(false)) {
+      !request.GetOption<DisableCrc32cChecksum>().value_or(false) ||
+      request.HasOption<MD5HashValue>() ||
+      request.HasOption<Crc32cChecksumValue>()) {
     return InsertObjectMediaMultipart(request);
   }
 
--- a/google/cloud/storage/emulator/emulator.py
+++ b/google/cloud/storage/emulator/emulator.py
@@ -48,6 +48,7 @@ def index():
 TODO(#6615): Introducing failures into uploads with return-XXX-after-YYYk
 """
 
+
 # Needs to be defined in emulator.py to keep context of flask and db global variables
 def retry_test(method):
     global supported_methods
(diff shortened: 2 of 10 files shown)


coryan:
If you are reviewing this: you might be better off thinking of the integration tests as completely new code.

google/cloud/storage/internal/curl_client.cc
```diff
@@ -536,7 +536,9 @@ StatusOr<ObjectMetadata> CurlClient::InsertObjectMedia(
   // uploads. `DisableMD5Hash` and `DisableCrc32cChecksum` should not be
   // dependent on each other.
   if (!request.GetOption<DisableMD5Hash>().value() ||
-      !request.GetOption<DisableCrc32cChecksum>().value_or(false)) {
+      !request.GetOption<DisableCrc32cChecksum>().value_or(false) ||
```
devjgm, line 539: For `DisableCrc32cChecksum` we check `.value_or(false)`, but for `DisableMD5Hash` we just check `.value()`. Is that difference intentional?

devjgm (approved):

google/cloud/storage/internal/curl_client.cc
coryan: Yes, but a bit weird.  `DisableMD5Hash` is always set, its default constructor sets its value to `true` whereas `DisableCrc32cChecksum` defaults the value to `absl::nullopt`... groan.
I will clean this up in a future PR (#7029)

PR #7028, opened 2021-07-20
feat(bigtable): cap outstanding mutation number for mutation batcher

### LINKED ISSUE #6205: Throttle mutation batcher to avoid "full queue" errors
A customer got the following error using mutation batcher:

```
Error You have reached the limit of total mutations in your queue. Throttle your usage and wait for operations to finish. [RESOURCE_EXHAUSTED] while saving mutation in table ....
```

The customer reports that the batcher had sent 4466426 mutations with 60215 failures.

We should throttle the requests.

dbolduc:
Fixes #6205 as our users **should not** run into that exact "queued mutations" error any more.

One thing I noticed is that we do not do any bounding for the [knobs](https://github.com/dbolduc/google-cloud-cpp/blob/feda70036bd9847a190249fd0fbef271300045c3/google/cloud/bigtable/mutation_batcher.h#L63-L67) to set the limit (they can, for example, set mutations per request to be a gajillion). This is why I said "should not" instead of "can not".

I would sleep easier if we did not let them set values above the Bigtable imposed limits.... I will fix in a later PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7028)

### DIFF
--- a/google/cloud/bigtable/mutation_batcher.cc
+++ b/google/cloud/bigtable/mutation_batcher.cc
@@ -22,8 +22,10 @@ namespace cloud {
 namespace bigtable {
 inline namespace BIGTABLE_CLIENT_NS {
 
-// Cloud Bigtable doesn't accept more than this.
+// Cloud Bigtable doesn't accept more than this in a single request.
 auto constexpr kBigtableMutationLimit = 100000;
+// Maximum mutations that can be outstanding in the Cloud Bigtable front end.
+auto constexpr kBigtableOutstandingMutationLimit = 300000;
 // Let's make the default slightly smaller, so that overheads or
 // miscalculations don't tip us over.
 auto constexpr kDefaultMaxSizePerBatch =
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/mutation_batcher.cc
```diff
@@ -22,8 +22,10 @@ namespace cloud {
 namespace bigtable {
 inline namespace BIGTABLE_CLIENT_NS {
 
-// Cloud Bigtable doesn't accept more than this.
+// Cloud Bigtable doesn't accept more than this in a single request.
 auto constexpr kBigtableMutationLimit = 100000;
+// Maximum mutations that can be outstanding in the Cloud Bigtable front end.
+auto constexpr kBigtableOutstandingMutationLimit = 300000;
```
coryan, line 28: There is something odd about tracking in one process a limit that is system-wide... I may only send 100,000 mutations, but if you sent 250,000 I am affected too.

I guess I can prevent hurting myself, but this seems brittle.

coryan (approved):

PR #7031, opened 2021-07-20
feat(GCS+gRPC): upload checksums with last chunk

### LINKED ISSUE #4156: Implement CRC32C checksum in GrpcResumableUploadSession

### LINKED ISSUE #4157: Implement MD5 checksum in GrpcResumableUploadSession
With gRPC we can compute the checksum as data is streamed and send it with the last message to finish the upload. That provides better data integrity as the upload will not create any object if the checksum is mismatched.

coryan:
gRPC supports sending the object checksums in the last chunk of a
resumable upload. This changes the `internal::ResumableUploadSession`
class to receive these checksums, and weaves them through all the
decorators and tests. I also changed `internal::ObjectWriteStream` to
compute and store the checksums, and only validate them later.

Fixes #4156 and fixes #4157

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7031)

### DIFF
--- a/google/cloud/storage/client.cc
+++ b/google/cloud/storage/client.cc
@@ -94,7 +94,7 @@ ObjectWriteStream Client::WriteObjectImpl(
     ObjectWriteStream error_stream(
         absl::make_unique<internal::ObjectWriteStreambuf>(
             std::move(error), 0, internal::CreateNullHashFunction(),
-            internal::CreateNullHashValidator(),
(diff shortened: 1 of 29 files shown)


devjgm (approved):
LGTM

PR #7034, opened 2021-07-20
cleanup(storage): use Disable{MD5*,Crc32c*} consistently

### LINKED ISSUE #7029: Cleanup uses of `Disable{MD5Hash,Crc32Checksum}`
The uses are not symmetric, as MD5 hashes are disabled by default, and CRC32C checksums are enabled by default.  I think my decision to use the default constructor of `DisableMD5Hash` made the rest of the code too complicated.

coryan:
All GCS options are implemented as an `absl::optional<T>`, in the case
of `DisableMD5Hash` and `DisableCrc32cChecksum` this is
`absl::optional<bool>`. The MD5 hashes are disabled by default, this is
achieved by having the default constructor for `DisableMD5Hash`
initialize the option to `true`. CRC32C checksums are enabled by
default, the default constructor initializes the optional to
`absl::nullopt` (this is consistent with all the other options).

After this change we always use these options by doing:

```cc
auto value = request.GetOption<Disable*>().value_or(false);
```

This makes the usage consistent, and only the constructor encodes
what the default is.

Fixes #7029

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7034)

### DIFF
--- a/google/cloud/storage/internal/hash_function.cc
+++ b/google/cloud/storage/internal/hash_function.cc
@@ -45,7 +45,7 @@ std::unique_ptr<HashFunction> CreateHashFunction(
   if (request.RequiresRangeHeader()) return CreateNullHashFunction();
   return CreateHashFunction(
       request.GetOption<DisableCrc32cChecksum>().value_or(false),
-      request.GetOption<DisableMD5Hash>().value_or(true));
+      request.GetOption<DisableMD5Hash>().value_or(false));
 }
 
 std::unique_ptr<HashFunction> CreateHashFunction(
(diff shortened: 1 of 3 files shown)


google/cloud/storage/internal/hash_function.cc
```diff
@@ -45,7 +45,7 @@ std::unique_ptr<HashFunction> CreateHashFunction(
   if (request.RequiresRangeHeader()) return CreateNullHashFunction();
   return CreateHashFunction(
       request.GetOption<DisableCrc32cChecksum>().value_or(false),
-      request.GetOption<DisableMD5Hash>().value_or(true));
+      request.GetOption<DisableMD5Hash>().value_or(false));
```
devjgm, line 48: This default changed from `true` to `false`. Is this an intentional change in behavior?

devjgm (commented):
Thanks. Just one question.

google/cloud/storage/internal/hash_function.cc
coryan: Should not be.  The default constructor for `DisableMD5Hash` sets the value to `true`.  The other constructors either set it to either `true` or `false`, the `_or()` branch is never taken.  It is confusing, but I thought it was less confusing than using `.value_or(false)` for CRC32C and `.value_or(true)` for MD5.  I can change it if you prefer, I think both can be messy.

devjgm: No, it's fine as is. Thanks.

devjgm (approved):

PR #7036, opened 2021-07-21
cleanup(bigtable): strictly enforce bigtable limits in mutation batcher

### LINKED ISSUE #7030: Strictly enforce Bigtable mutation limits
`MutationBatcher` allows users to [set values](https://github.com/googleapis/google-cloud-cpp/blob/e49eefac8a856d52e66bda996c07f4fcb06374d4/google/cloud/bigtable/mutation_batcher.h#L63-L67) above the maximum allowed [mutations per batch](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/bigtable/mutation_batcher.cc#L26) and [outstanding mutations](https://github.com/googleapis/google-cloud-cpp/blob/feda70036bd9847a190249fd0fbef271300045c3/google/cloud/bigtable/mutation_batcher.cc#L28). We should never let these values exceed the Bigtable imposed limits.

dbolduc:
Fixes #7030

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7036)

### DIFF
--- a/google/cloud/bigtable/mutation_batcher.h
+++ b/google/cloud/bigtable/mutation_batcher.h
@@ -155,6 +155,7 @@ class MutationBatcher {
    */
   std::pair<future<void>, future<Status>> AsyncApply(CompletionQueue& cq,
                                                      SingleRowMutation mut);
+
   /**
    * Asynchronously wait until all submitted mutations complete.
    *
--- a/google/cloud/bigtable/mutation_batcher.h
+++ b/google/cloud/bigtable/mutation_batcher.h
@@ -61,10 +61,7 @@ class MutationBatcher {
     Options();
 
     /// A single RPC will not have more mutations than this.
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/mutation_batcher.h
```diff
@@ -161,7 +155,6 @@ class MutationBatcher {
    */
   std::pair<future<void>, future<Status>> AsyncApply(CompletionQueue& cq,
                                                      SingleRowMutation mut);
-
```
devbww, line 164: I'd probably restore this separating line.

devbww (approved):

google/cloud/bigtable/mutation_batcher.h
dbolduc: Whoops. Thanks for catching that

PR #7058, opened 2021-07-26
feat(common): make GCP_LOG(FATAL) terminate execution

### LINKED ISSUE #5693: GCP_LOG(FATAL) should terminate execution

devbww:
`GCP_LOG(FATAL)` now calls `std::abort()` during destruction
of its `Logger`, so control will never return to the caller.
This makes it more closely resemble the behavior of Abseil
logging, which eliminates some confusion between authors and
code readers, and which may also be a feature down the line.

This required some refactoring in a few places because the
control flow is more hidden, and `[[unreachable]]` isn't a
standard attribute yet.

Some, but not all, existing `GCP_LOG(FATAL)` invocations
were immediately followed by calls to `std::exit(1)`. These
were removed. The streaming arguments were also purged of
any trailing newlines and flushes, which `GCP_LOG()` already
handles.

Finally, `storage::AssertOptionSuccess()` was changed to
use `GCP_LOG(FATAL)` instead of `ThrowLogicError()`,
because (as implied in the `TODO`) an assertion failure
is not really something that should be caught.

Fixes #5693.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7058)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -72,8 +72,7 @@ std::string CppTypeToString(FieldDescriptor const* field) {
       return ProtoNameToCppName(field->message_type()->full_name());
   }
   GCP_LOG(FATAL) << "FieldDescriptor " << std::string(field->cpp_type_name())
-                 << " not handled.";
-  std::exit(1);
+                 << " not handled";
 }
 
 std::string FormatDoxygenLink(google::protobuf::Descriptor const& message_type,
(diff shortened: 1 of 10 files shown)


dbolduc (approved):

generator/internal/descriptor_utils.cc
```diff
@@ -238,29 +237,23 @@ void SetResourceRoutingMethodVars(
...
-    case google::api::HttpRule::kPost:
-    case google::api::HttpRule::kDelete:
-    case google::api::HttpRule::kPatch:
-      return "kNonIdempotent";
-      break;
-    default:
-      GCP_LOG(FATAL) << __FILE__ << ":" << __LINE__
-                     << ": google::api::HttpRule not handled" << std::endl;
-      std::exit(1);
+  if (method.options().HasExtension(google::api::http)) {
```
coryan, line 240: FYI, this file seems to have several additional cleanups, unrelated to the main change. Fine by me, but a bit harder to grok.

google/cloud/log_test.cc
```diff
@@ -292,6 +292,11 @@ TEST(LogSinkTest, DisabledLogsMakeNoCalls) {
   EXPECT_EQ(0, counter);
 }
 
+TEST(LogFatalTest, NoReturn) {
+  LogSink::EnableStdClog();
```
coryan, line 296: Is this needed because the messages are not logged otherwise?  There is something odd about crashing the application without logging.  Maybe `FATAL` should always log to `std::clog`?  Or maybe it should log to `std::clog` if there are no other backends configured?

coryan (commented):
I am mostly Okay with these changes.  The fact that we can call `abort()` without logging is slightly disconcerting.  Thoughts?

generator/internal/descriptor_utils.cc
devbww: I did not intend to change any behavior here ... just invert some tests so that `GCP_LOG(FATAL)` is not the last thing in a non-void function.  This is the "[[unreachable]] refactoring" I mentioned in the description, although undoubtedly poorly expressed.  It would be nice if the compiler could clearly see that a FATAL log does not return.

google/cloud/log_test.cc
devbww: I concur that having `FATAL` logs (or indeed, perhaps anything about `WARNING`) always at least end up on `std::clog` might be a feature.  But this PR is not changing that behavior.  Even before a `GCP_LOG(FATAL) << "foo"; std::exit(1)` could silently exit if there was no enabled log sink.  So, I think we should consider that separately.

generator/internal/descriptor_utils.cc
coryan: Ah, SGTM then.

google/cloud/log_test.cc
coryan: Yeah, I did not mean you should fix that in this PR.  Separate change for it SGTM too.

coryan (approved):

Issue #7052, opened 2021-07-23
Upgrading to grpc 1.39.0 makes our `cmake-install` build too slow


devjgm:
Step #3: GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/16a86463-87c8-4b20-ae98-bc8913b3de7a;tab=detail?project=cloud-cpp-testing-resources
Step #3: Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/7046/2fcd9ef25cf6c6221ea8af73def9e9d0ebe05ea5/fedora-34-cmake-install/log-16a86463-87c8-4b20-ae98-bc8913b3de7a.txt

We upgraded grpc from 1.38.1 to 1.39.0 and our `cmake-install` build became unusably slow, so we rolled it back. To roll this change forward `git revert ba41d5a99818922b09eee5da148786c1d68c17f5` (ba41d5a99818922b09eee5da148786c1d68c17f5)

The slow line is:
https://github.com/googleapis/google-cloud-cpp/blob/727bc4f4676a75a97cec91247bcdd78741cf7204/ci/verify_deprecated_targets/CMakeLists.txt#L79

The line is slow when checking a module that links with grpc, like `google_cloud_cpp_grpc_utils`

I suspect there may be some change in this version of grpc to its pkg-config files. Maybe it's also exposing a bug in `pkg-config` itself, maybe similar to https://github.com/googleapis/google-cloud-cpp/pull/6846

NOTE: We need to figure this out because we need to be able to upgrade gRPC

devjgm:
@veblush FYI

devjgm:
We see this in a docker image w/ the following info:

```
Step #3: ====================
Step #3: |   Machine Info   |
Step #3: ====================
Step #3:      host: 2021-07-23 14:50:49+00:00
Step #3:    google: 2021-07-23 14:50:49+00:00
Step #3:    kernel: #49~18.04.1-Ubuntu SMP Fri Jun 18 21:33:55 UTC 2021
Step #3:        os: PRETTY_NAME="Fedora 34 (Container Image)"
Step #3:     nproc: 32
Step #3:       mem: 125.81 GiB
Step #3:      term: dumb
Step #3:       gcc: gcc (GCC) 11.1.1 20210531 (Red Hat 11.1.1-3)
Step #3:     clang: clang version 12.0.0 (Fedora 12.0.0-2.fc34)
[...]

PR #7061, opened 2021-07-26
ci: avoid fedora's `pkgconf` which is slow

devjgm:
Fixes: #7052

This avoids Fedora's `pkgconf` (a drop-in replacement for `pkg-config`)
because it's too slow when dealing with `.pc` files with lots of
`Requires:` deps, which we have with Abseil. Instead, we use the normal
`pkg-config` binary, which seems to work better.

After this PR, we should be able to upgrade to gRPC 1.39.0 (`git revert
ba41d5a99818922b09eee5da148786c1d68c17f5`)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7061)

### DIFF
--- a/ci/cloudbuild/builds/cmake-install.sh
+++ b/ci/cloudbuild/builds/cmake-install.sh
@@ -149,6 +149,7 @@ done < <(find "${INSTALL_PREFIX}" -type f -print0)
 
 for repo_root in "ci/verify_current_targets" "ci/verify_deprecated_targets"; do
   out_dir="cmake-out/$(basename "${repo_root}")-out"
+  rm -f "${out_dir}/CMakeCache.txt"
   io::log_h2 "Verifying CMake targets in repo root: ${repo_root}"
   cmake -GNinja -DCMAKE_PREFIX_PATH="${INSTALL_PREFIX}" \
     -S "${repo_root}" -B "${out_dir}" -Wno-dev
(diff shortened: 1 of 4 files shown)


ci/cloudbuild/builds/cmake-install.sh
```diff
@@ -149,6 +149,7 @@ done < <(find "${INSTALL_PREFIX}" -type f -print0)
 
 for repo_root in "ci/verify_current_targets" "ci/verify_deprecated_targets"; do
   out_dir="cmake-out/$(basename "${repo_root}")-out"
+  rm -f "${out_dir}/CMakeCache.txt"
```
devjgm, line 152: This is so that we can run the `cmake-install` build multiple times on a machine where this file is persistent (e.g., our `--docker` builds). The problem is that this file caches the _temporary_ install location from the previous invocation.

coryan: For a separate PR: maybe the install location should not be temporary?  `/h/test-install` or `cmake-out/test-install`?

coryan (approved):

ci/cloudbuild/builds/cmake-install.sh
devjgm: I think we want a fully clean install each time. Otherwise, if we _removed_ an artifact in one build, the artifact may still exist on disk and we may not notice that we're still depending on the removed artifact.... well, our CI builds in GCB would notice, but we wouldn't notice on our local machines until we blew away the install directory and rebuilt. Anyway, that's why I originally installed in a new dir every time.

PR #7064, opened 2021-07-26
fix(storage): cleanup interrupted downloads

### LINKED ISSUE #7051: CurlRequest::MakeRequestImpl could crash when SetOption for CURLOPT_BUFFERSIZE
If the problem is with the Google Cloud service exposed by the `google-cloud-cpp` libraries instead of the client libraries themselves, you may consider opening a support request instead. The `google-cloud-cpp` developers cannot help you troubleshoot problems with the service itself.
Yes.

For example, is this related to bigtable (i.e., something in `google/cloud/bigtable`), or GCS (i.e., something in `google/cloud/storage`)?
google/cloud/storage

**Describe the bug**
Curl lib setopt.c checks data->state.buffer and return CURLE_BAD_FUNCTION_ARGUMENT(https://github.com/curl/curl/pull/5842/files).
CurlRequestBuilder uses handle_ from a PooledCurlHandleFactory that reuses handle.
The handle could still have data->state.buffer set and would cause runtime error when
CurlRequest::MakeRequestImpl is called to SetOption for CURLOPT_BUFFERSIZE.

**To Reproduce**
1. Call function 'Curl::Client::CreateResumableSessionGeneric'
2. See runtime error.

**Expected behavior**
No runtime error.

**Operating system:**
If you are using a Linux distribution please include the name and version of the distribution too.

Please include the output of `g++ -v` or `clang++ -v` or the equivalent command-line flag.

Please include the output from `git rev-parse HEAD` if you are compiling from source, or the version number from the applicable `google/cloud/*/version.h` file.

**Additional context**
In EasyPerform -> easy_perform in easy.c either easy_events or easy_transfer would be called to initiate a transfer and possibly initialize data->state.buffer. In curl_multi_remove_handle, the multi_done step would call Curl_safefree for
data->state.buffer. So not sure why data->state.buffer would be set.

coryan:
Downloads can be interrupted by the application. Sometimes this results
in incomplete cleanup of the handle, which when reused would start
failing some `curl_easy_setopt()` calls.

Fixes #7051

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7064)

### DIFF
--- a/google/cloud/storage/tests/curl_download_request_integration_test.cc
+++ b/google/cloud/storage/tests/curl_download_request_integration_test.cc
@@ -30,6 +30,8 @@ inline namespace STORAGE_CLIENT_NS {
 namespace internal {
 namespace {
 
+using ::google::cloud::testing_util::IsOk;
+
 std::string HttpBinEndpoint() {
   return google::cloud::internal::GetEnv("HTTPBIN_ENDPOINT")
       .value_or("https://nghttp2.org/httpbin");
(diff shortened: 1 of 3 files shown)


google/cloud/storage/tests/curl_download_request_integration_test.cc
```diff
@@ -72,6 +74,65 @@ TEST(CurlDownloadRequestTest, SimpleStream) {
   EXPECT_EQ(kDownloadedLines, count);
 }
 
+// Run one attempt of the Regression7501 test. This is wrapped in a retry loop,
+// as integration tests flake due to unrelated (and unavoidable) problems, e.g.,
+// trying to setup connections.
+Status AttemptRegression7501() {
```
k7z45, line 80: Do you mean 7051 vs. 7501?

google/cloud/storage/tests/curl_download_request_integration_test.cc
```diff
@@ -72,6 +74,65 @@ TEST(CurlDownloadRequestTest, SimpleStream) {
...
+  // Download the maximum number of lines supported by httpbin.org
+  auto constexpr kDownloadedLines = 100;
+  auto constexpr kTestPoolSize = 32;
```
k7z45, line 83: So this test is making the assumption that the same handle is used across multiple download requests. If so, would you like to assert that?

google/cloud/storage/tests/curl_download_request_integration_test.cc
```diff
@@ -72,6 +74,65 @@ TEST(CurlDownloadRequestTest, SimpleStream) {
...
+
+  {
+    auto r_partial_close = make_download();
+    auto read = r_partial_close.Read(buffer, kBufferSize);
+    if (!read) return std::move(read).status();
+    auto close = r_partial_close.Close();
+    if (!close) return std::move(close).status();
+  }
+
+  auto r_full = make_download();
```
k7z45, line 111: Do you need another no close read before the full read?

coryan: No, I want to check that a handle where (a) not all the data is downloaded, and (b) you do call `Close()`, is usable.

google/cloud/storage/tests/curl_download_request_integration_test.cc
coryan: Done.  But I would have been happy to just test "the factory returns a usable handle".  If the handle is usable because it is brand new, or because it was completely cleaned up then 🤷 .

google/cloud/storage/tests/curl_download_request_integration_test.cc
coryan: Fixed.

devjgm (approved):

Issue #189, opened 2018-01-18
Write a "single RPC" latency test for the Bigtable Client.


coryan:
We need a latency benchmark for the Bigtable C++ Client library.  The benchmark will need to:

1. Work with the emulator, production, and with an embedded server that implements the `google.bigtable.v2.Bigtable` service.
2. Accept the project id, and instance id as command-line parameters.
2. The table id start with `perf` followed by 8 random (but valid) characters.
3. Create the table during startup.  The table should contain a single column family `cf`, set to retain a single version.
4.  The test should populate the table with 10,000,000 unique keys, in the `user[0-9]{7}` format, e.g., `user1234567`, `user0000001`.
    * Each row should contain 10 columns named `field0` through `field9`.
    * Each column should contain a single value with 100 characters (randomized).
    * When loading data to production or the emulator it is recommended that
       * The upload is parallelized, using multiple threads, e.g., thread 0 creates the rows with keys in the `user0000000` to `user0999999` range.
       * Each thread uses `BulkApply()` with 1,000 entries per batch.
5. The test should execute for 30 minutes, the running time must be configurable.
6. The test should perform random reads and writes, approximately 50% of the operations should be reads, and 50% should be writes.
7. The operation should select a random key to operate on.
8. The test should capture the execution time (latency, in microseconds) for each operation.  It should also capture if the operation was successful or failed.
    * Keep the results for reads vs. writes separate.
    * Notice that with latency ~5msec you will need approximately 360,000 entries to keep all the results.
9. The test should delete the table when done.
10. The test should produce a CSV file as output, including the results.  Please talk to @sduskis about the format of this file, but it must include:
    * Latency for load inserts, run reads, run updates. Including min, max, p50, p90, p95, p99 and p99.9.
    * Overall throughput for load and run phases.
    * Execution information: which client (language) and version, start time, what type of test.

PR #7070, opened 2021-07-27
feat(bigtable): add benchmark to measure MutationBatcher throughput

### LINKED ISSUE #7024: Add a benchmark in Bigtable for `SingleRowMutation` write throughput
To inform #7023 we need data to help us determine defaults for our `MutationBatcher::Options`. At a minimum the benchmark should let us compare throughput (mutations / s) of `MutationBatcher::AsyncApply()` for the following variables:
* mutation batch size (either bytes or # of mutations)
* concurrent outgoing requests

... with total number of mutations held fixed.

Optionally, the benchmark can:
* also measure latency (s) per request
* make the thread count configurable
* compare results to using `Table::AsyncApply()` directly.
* make the node count for the cluster configurable. (probably out of scope).

dbolduc:
Fixes #7024 

A new benchmark to measure the throughput of `bigtable::MutationBatcher` given maximum number of mutations per batch and maximum number of outstanding batches.

#### Shards / Splits / Row keys:

There are no initial splits provided to the table. I considered splitting based on:
* % of table - e.g. always split into 10 shards with 10% of the rows in each.
* \# of rows in split - e.g. split such that each shard will contain 1,000,000 rows
* \# of threads - e.g. launch T threads. split such that each shard contains 1/T of the data

(Our other benchmarks launch 10 threads and split the 10M rows into 10 shards. This is where my example numbers are coming from) 

But my current thinking is that it wouldn't be helpful at this time. I assume that the advantage of initial splits come when you concurrently send batches that hit different shards (where the mutations within a batch all belong to the same shard). Please correct me if I'm wrong.

The `MutationBatcher` is fed mutations from many threads. It simply [appends](https://github.com/googleapis/google-cloud-cpp/blob/1d18bca029224cab43a2e63d9a00716033cfc5df/google/cloud/bigtable/mutation_batcher.cc#L86) new mutations into the current batch. So I think the only way (presently) to guarantee mutations from a single shard make it into a single batch is to only send mutations from that shard across all threads. But then we are not getting the advantage of concurrent batches going to _different_ shards.

Initial splits might be worth adding. It would especially be worth adding if we change the guts of the `MutationBatcher` to try to batch mutations with splits in mind. I think it would come with a `--shard-count` option. And the row keys would be something like: `"row" + shard_index + mutation_index`. 

#### Cutoff deadline:

I want to be able to run the benchmark in a loop with random settings. Some of these settings are going to be very slow. I do not want to wait for them. I want to say something like "the best settings do M mutations in ~25s , so let me set a `--max-time=60s`. 

There might be a way to do this in bash with `timeout` but 1) I don't know how, 2) I want to get some sort of throughput measure if the program does get cut off.

This does complicate the code a little bit. I am obviously open to a better way to do this.

#### Logging progress:

Logging the progress of the writes also complicates the code a bit. But it's nice to know the program is doing something.

#### Things I didn't do:
* Accept an App Profile ID. Should I?
* Add ability to set maximum size in bytes of a batch (instead of maximum requests per batch). I didn't think this was necessary, as all mutations are of the same size. So I don't know if anything would be gained here.
* Make a way to compare this to using `Table::Apply` or `Table::BulkApply` directly. Our other benchmarks record throughput for manually sent batches. We can use those to get a feel for the throughput of the direct table methods.
* Make gRPC channel count configurable for the `DataClient`. It just uses the default.
* Have a setting for random row keys (instead of sequenced). I originally had this in, but I realized we are not even sending the batcher sequenced row keys, as discussed above in the initial splits section.
* Report total batches sent by the `MutationBatcher`. I think this data is valuable, but getting it is not yet native to the batcher. My plan is to first: get the benchmark done. Then second: modify `MutationBatcher` to record batches sent + update this benchmark to report the data.
* Use a version of `MutationBatcher::AsyncApply()` that uses the table's background threads. This method hasn't been made yet. Similar to the total batches, I want to first get the benchmark done before adding this feature.

#### Things that are highly questionable:
* I set the background thread pool size for the table to be equal to `max_batches`. I thought this made sense, but maybe I need to cap it. We probably don't want hundreds of threads if we set the option too high...
* I left `column_family` and `column` in the options struct... so they are essentially hard coded.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070)

### DIFF
--- a/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
+++ b/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
@@ -1,0 +1,274 @@
+// Copyright 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/bigtable/benchmarks/mutation_batcher_throughput_options.h"
+#include "google/cloud/bigtable/mutation_batcher.h"
+#include "google/cloud/bigtable/table.h"
+#include "google/cloud/bigtable/table_admin.h"
+#include "google/cloud/internal/getenv.h"
+#include "google/cloud/status_or.h"
+#include "google/cloud/testing_util/command_line_parsing.h"
+#include "google/cloud/testing_util/timer.h"
+#include "absl/time/time.h"
+#include <chrono>
+#include <iomanip>
+#include <sstream>
+#include <string>
+#include <thread>
+
+namespace {
+
+namespace cbt = google::cloud::bigtable;
+using cbt::benchmarks::MutationBatcherThroughputOptions;
+using cbt::benchmarks::ParseMutationBatcherThroughputOptions;
+using google::cloud::Status;
+using google::cloud::StatusCode;
+using google::cloud::internal::GetEnv;
+
+char const kDescription[] =
+    R"""(A benchmark to measure the throughput of the `MutationBatcher` class.
+
+The purpose of the program is to determine effective default settings for
+`MutationBatcher::Options` that maximize throughput. The specific settings,
+which are the main inputs to this program, are maximum mutations per batch and
+maximum concurrent batches in flight.
+
+The program is designed to be run repeatedly. It has the ability to cut itself
+off (if really bad options are supplied). It also has the ability to use a
+pre-existing table instead of creating a new one then deleting it when the
+program is done.
+
+The program will:
+
+1) Conditionally create a table.
+2) Echo your configuration settings.
+3) Configure a `MutationBatcher`.
+4) Spin off some number of threads.
+5) Send mutations to a `MutationBatcher` from each thread. The `MutationBatcher`
+   will apply these mutations to the table. Progress will be logged for those
+   who are impatient.
+6) Conditionally stop batching mutations if the program exceeds a supplied deadline.
+7) Report the total time it took to apply the mutations.
+8) Report the total number of successful and failed mutations.
+9) Conditionally delete the table.
+)""";
+
+google::cloud::StatusOr<MutationBatcherThroughputOptions> ParseArgs(
+    int argc, char* argv[]) {
+  auto const auto_run =
+      GetEnv("GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES").value_or("") == "yes";
+  if (auto_run) {
+    for (auto const& var : {"GOOGLE_CLOUD_PROJECT",
+                            "GOOGLE_CLOUD_CPP_BIGTABLE_TEST_INSTANCE_ID"}) {
+      auto const value = GetEnv(var).value_or("");
+      if (!value.empty()) continue;
+      std::ostringstream os;
+      os << "The environment variable " << var << "is not set or empty";
+      return Status(StatusCode::kUnknown, std::move(os).str());
+    }
+    auto const project_id = GetEnv("GOOGLE_CLOUD_PROJECT").value();
+    auto const instance_id =
+        GetEnv("GOOGLE_CLOUD_CPP_BIGTABLE_TEST_INSTANCE_ID").value();
+    return ParseMutationBatcherThroughputOptions(
+        {
+            std::string(argv[0]),
+            "--project-id=" + project_id,
+            "--instance-id=" + instance_id,
+            "--mutation-count=1000",
+            "--max-batches=3",
+        },
+        kDescription);
+  }
+
+  return ParseMutationBatcherThroughputOptions({argv, argv + argc},
+                                               kDescription);
+}
+
+cbt::SingleRowMutation MakeMutation(
+    MutationBatcherThroughputOptions const& options, int key_width,
+    std::int64_t const& row_key) {
+  std::ostringstream os;
+  os << "row" << std::setw(key_width) << std::setfill('0') << row_key;
+  return cbt::SingleRowMutation(
+      std::move(os).str(),
+      {cbt::SetCell(options.column_family, options.column, "value")});
+}
+
+}  // namespace
+
+int main(int argc, char* argv[]) {
+  auto options = ParseArgs(argc, argv);
+  if (!options) {
+    std::cerr << options.status() << "\n";
+    return 1;
+  }
+  if (options->exit_after_parse) return 0;
+
+  namespace cbt = google::cloud::bigtable;
+  using google::cloud::CompletionQueue;
+  using google::cloud::future;
+  using google::cloud::Options;
+  using google::cloud::Status;
+  using google::cloud::testing_util::Timer;
+
+  cbt::TableAdmin admin(cbt::CreateDefaultAdminClient(options->project_id, {}),
+                        options->instance_id);
+
+  // Create a new table if one was not supplied
+  auto table_id = options->table_id;
+  if (table_id.empty()) {
+    using google::cloud::internal::Sample;
+    auto generator = google::cloud::internal::MakeDefaultPRNG();
+    std::string table_id_chars =
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_";
+    table_id = "table-" + Sample(generator, 8, table_id_chars);
+
+    std::cout << "# Creating Table\n";
+    auto status = admin.CreateTable(
+        table_id,
+        cbt::TableConfig(
+            {{options->column_family, cbt::GcRule::MaxNumVersions(10)}}, {}));
+    if (!status) {
+      std::cout << status.status() << std::endl;
+      return -1;
+    }
+    std::cout << "#\n";
+  } else {
+    auto status = admin.GetTable(table_id, cbt::TableAdmin::NAME_ONLY);
+    if (!status) {
+      std::cout << status.status() << std::endl;
+      return -1;
+    }
+  }
+
+  auto opts = Options{}.set<google::cloud::GrpcBackgroundThreadPoolSizeOption>(
+      options->max_batches);
+  auto table = cbt::Table(
+      cbt::CreateDefaultDataClient(options->project_id, options->instance_id,
+                                   cbt::ClientOptions(std::move(opts))),
+      table_id);
+
+  cbt::MutationBatcher batcher(
+      table, cbt::MutationBatcher::Options{}
+                 .SetMaxBatches(options->max_batches)
+                 .SetMaxMutationsPerBatch(options->batch_size));
+
+  std::cout << "# Project ID: " << options->project_id
+            << "\n# Instance ID: " << options->instance_id
+            << "\n# Table ID: " << table_id << "\n# Cutoff Time: "
+            << absl::FormatDuration(absl::FromChrono(options->max_time))
+            << "\n# Thread Count: " << options->thread_count
+            << "\n# Total Mutations: " << options->mutation_count
+            << "\n# Mutations per Batch: " << options->batch_size
+            << "\n# Concurrent Batches: " << options->max_batches << std::endl;
+
+  std::atomic<int> fails{0};
+  std::atomic<int> successes{0};
+  std::atomic<bool> timeout{false};
+
+  // Conditionally start the timeout thread
+  std::thread* timeout_thread = nullptr;
+  if (options->max_time.count() > 0) {
+    auto deadline = std::chrono::system_clock::now() + options->max_time;
+    timeout_thread = new std::thread([deadline, &timeout] {
+      while (std::chrono::system_clock::now() < deadline) {
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+      }
+      timeout = true;
+    });
+  }
+
+  // Start the MutationBatcher thread
+  CompletionQueue cq;
+  std::thread cq_runner([&cq] { cq.Run(); });
+
+  int key_width = 0;
+  for (std::int64_t i = options->mutation_count - 1; i > 0; i /= 10) ++key_width;
+  auto write = [&options, &batcher, &cq, &fails, &successes, &timeout,
+                key_width](std::int64_t const& start, std::int64_t const& end,
+                           bool const& log) {
+    // One write thread will log its progress
+    if (log) std::cout << "#\n# Writing" << std::flush;
+    auto progress_period = (std::max)(1L, (end - start) / 20);
+
+    for (std::int64_t i = start; i < end; ++i) {
+      // Stop writing if we hit the cutoff deadline
+      if (timeout) break;
+
+      auto mut = MakeMutation(*options, key_width, i);
+      auto admission_completion = batcher.AsyncApply(cq, mut);
+      auto& admission_future = admission_completion.first;
+      auto& completion_future = admission_completion.second;
+      completion_future.then([&fails, &successes](future<Status> fut) {
+        auto status = fut.get();
+        if (!status.ok())
+          ++fails;
+        else
+          ++successes;
+      });
+      admission_future.get();
+
+      if (log && (i - start) % progress_period == 0)
+        std::cout << "." << std::flush;
+    }
+    if (log) std::cout << "\n#" << std::endl;
+  };
+
+  auto timer = Timer::PerThread();
+
+  std::vector<std::future<void>> tasks;
+  std::int64_t start = 0;
+  for (int j = 0; j < options->thread_count; ++j) {
+    auto end =
+        (std::min)(options->mutation_count,
+                   start + options->mutation_count / options->thread_count);
+    tasks.emplace_back(
+        std::async(std::launch::async, write, start, end, j == 0));
+    start = end;
+  }
+
+  for (auto& t : tasks) t.get();
+
+  // Wait for all mutations to complete
+  batcher.AsyncWaitForNoPendingRequests().get();
+
+  // Join the MutationBatcher thread
+  cq.Shutdown();
+  cq_runner.join();
+
+  auto snapshot = timer.Sample();
+  std::cout << "MutationCount,BatchSize,MaxBatches,ElapsedMicroseconds,"
+               "Successes,Fails\n"
+            << options->mutation_count << "," << options->batch_size << ","
+            << options->max_batches << "," << snapshot.elapsed_time.count()
+            << "," << successes << "," << fails << "\n";
+
+  // If we created a table, delete it.
+  if (options->table_id.empty()) {
+    std::cout << "#\n# Deleting Table\n";
+    auto status = admin.DeleteTable(table_id);
+    if (!status.ok()) {
+      std::cout << status << std::endl;
+      return -1;
+    }
+  }
+
+  // Detach the timeout thread if it is running
+  if (timeout_thread) {
+    timeout_thread->detach();
+    delete timeout_thread;
+  }
+
+  return 0;
+}
--- a/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
+++ b/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
@@ -21,6 +21,7 @@
 #include "google/cloud/testing_util/command_line_parsing.h"
 #include "google/cloud/testing_util/timer.h"
 #include "absl/time/time.h"
+#include <algorithm>
 #include <chrono>
 #include <iomanip>
 #include <sstream>
--- a/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
+++ b/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
@@ -16,6 +16,8 @@
 #include "google/cloud/bigtable/mutation_batcher.h"
 #include "google/cloud/bigtable/table.h"
 #include "google/cloud/bigtable/table_admin.h"
+#include "google/cloud/bigtable/testing/random_names.h"
+#include "google/cloud/internal/background_threads_impl.h"
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/status_or.h"
 #include "google/cloud/testing_util/command_line_parsing.h"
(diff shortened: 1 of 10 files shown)


google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
```diff
@@ -0,0 +1,274 @@
...
+
+  // Start the MutationBatcher thread
+  CompletionQueue cq;
+  std::thread cq_runner([&cq] { cq.Run(); });
+
+  int key_width = 0;
+  for (auto i = options->mutation_count - 1; i > 0; i /= 10) ++key_width;
+  auto write = [&options, &batcher, &cq, &fails, &successes, &timeout,
+                key_width](std::int64_t const& start, std::int64_t const& end,
+                           bool const& log) {
```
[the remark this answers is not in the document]
coryan: Ditto for the `std::int64_t` parameters.

google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
```diff
@@ -0,0 +1,275 @@
...
+    // One write thread will log its progress
+    if (log) std::cout << "#\n# Writing" << std::flush;
+    auto progress_period = (std::max)(1L, (end - start) / 20);
+
+    for (std::int64_t i = start; i < end; ++i) {
```
coryan, line 206: I would just write `for (auto i = start; i != end; ++i)` 🤷

google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
```diff
@@ -0,0 +1,275 @@
...
+
+  // Join the MutationBatcher thread
+  cq.Shutdown();
+  cq_runner.join();
+
+  auto snapshot = timer.Sample();
+  std::cout << "MutationCount,BatchSize,MaxBatches,ElapsedMicroseconds,"
+               "Successes,Fails\n"
+            << options->mutation_count << "," << options->batch_size << ","
+            << options->max_batches << "," << snapshot.elapsed_time.count()
```
coryan, line 255: The `PerThread` timer may be overkill here.  Consider just using a steady clock if you are not interested in CPU time.  Or use the `PerProcess` timer and report the CPU time?

coryan (changes requested):

dbolduc (commented):
I also (hopefully) dealt with the `std::max` errors on the kokoro builds. PTAL.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-:-Mffdpj72kjpov8g3Z4j:bv6efqq)* status: 0 of 10 files reviewed, 14 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 43 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-43:-Mff4Bqy9-_6uVSQsKgu:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L43)):*

"effective ... that maximize throughput" maybe one or the other?
</blockquote></details>

Done.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 48 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-48:-Mff4kVFBKVbuHUfu6U9:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L48)):*

Could be slow because the options are bad or because the server is having a bad day, maybe?
</blockquote></details>

Done.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 144 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-144:-MffA8czEWQW4F5TQYTE:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L144)):*

s/return -1/return 1/ unless you really want the exit status to be 255 or something. Consider `return EXIT_FAILURE;` if you want to make the C++ language lawyers happy.
</blockquote></details>

Done.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 150 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-150:-MffCcapBGaJs93fxYVy:b-80tlqk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L150)):*

Will that be clear enough?  Do you wan to print the name of the table and so forth?
</blockquote></details>

Sure.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 181 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-181:-Mff_FuN2RqZLCemjPTY:bf6jvoh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L181)):*

Why not just a regular `std::thread`?  You can use `std::thread::joinable()` to join if needed:

https://en.cppreference.com/w/cpp/thread/thread/joinable

</blockquote></details>

Ah. My previous code reveals my ignorance of how to use `std::thread::operator=`. Thanks for sharing the reference link.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 200 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-200:-MffDql42qEsANgPIHBl:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L200)):*

Ditto for the `std::int64_t` parameters.
</blockquote></details>

Done.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 210 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-210:-MffEOlMClJaiNAwdaGW:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L210)):*

can you `std::move(mut)` here?
</blockquote></details>

Done.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 215 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-215:-MffEjhLC82E4-RXA83d:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L215)):*

I think we need braces in this `if/else`.
</blockquote></details>

Done.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 223 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-223:-MffF6pj1dicTjgyyS7N:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L223)):*

Ditto, braces here.  We only use `if(foo) ...;` when it fits in a single line.
</blockquote></details>

Done.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 232 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r3-232:-MffKeb91v8tOlVkpljf:bs9en3d) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c41e963c3193b6e6cd182a0d62eff8a22b6027dd/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L232)):*

Consider:

```cc
std::int64_t start = 0;
std::vector<std::future<void>> tasks(options->thread_count);
for (auto& t : tasks) {
  auto end = ...;
  t = std::async(std::launch::async, write, start, end, start == 0);
  start = end;
}
```

You could use a `std::transform()` too if you are into that sort of thing.

</blockquote></details>

`Ctrl+C` + `Ctrl+V`.... Done.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 206 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r4-206:-MffDqCT1gzDJ4eW9CmA:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f3d9c809b55174cc72af817b9145575598b7456/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L206)):*

I would just write `for (auto i = start; i != end; ++i)` 🤷 
</blockquote></details>

Done.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 255 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r4-255:-MffLulW1UmkYFw31GD5:bcufdt2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f3d9c809b55174cc72af817b9145575598b7456/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L255)):*

The `PerThread` timer may be overkill here.  Consider just using a steady clock if you are not interested in CPU time.  Or use the `PerProcess` timer and report the CPU time?
</blockquote></details>

Done. Also changed the elapsed time to print in seconds.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 268 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r4-268:-MffYc_a7VOU2e6WOW1Q:b-ws34r3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f3d9c809b55174cc72af817b9145575598b7456/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L268)):*

This is UB, and most likely a crash waiting to happen.  If you detach the thread while it still holds a reference to `timeout` it will access an object that has gone out of scope.  Better to join the thread, with a flag to terminate it early if needed.

Better yet.  You can create a timer in the completion queue, something like:

```cc
auto timer = cq.CreateAbsoluteTimer(deadline).then([&timeout](auto f) { timeout = true; });
```

and then do:

```
timer.cancel();
timer.get();
```

to cancel and wait for the timer before shutting down the CQ.

</blockquote></details>

Ah. Makes sense. Thanks for the suggestion. This is much safer.

How I chose to deal with the no-deadline case is a bit of a hack. But I'm okay with it.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_options.h, line 33 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OK-r4-33:-MffFQGYDWHxE1SJ9NpZ:b-egvaqu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f3d9c809b55174cc72af817b9145575598b7456/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_options.h#L33)):*

nit: maybe the `table_id` should be after the `instance_id`.
</blockquote></details>

Agreed. This ordering was unintentional. Fixed.

coryan (changes requested):
>  But then we are not getting the advantage of concurrent batches going to different shards.

Maybe we should consider keeping a mutation batcher per shard / split.  Or maybe just deliver the mutations in order. I think there is a large search space for optimization here.

> Reviewed 7 of 10 files at r1, 1 of 1 files at r2, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-:-Mfh5cQ_BjU2fop6QkDU:bvgsaxo)* status: all files reviewed, 4 unresolved discussions (waiting on @dbolduc)

___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 206 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-MfdurCOx7A_LGmOF1OI-r4-206:-Mfh34ib6ly_QiEyqa7G:bfwv67) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f3d9c809b55174cc72af817b9145575598b7456/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L206)):*

Done.
</blockquote></details>

nit: I think almost everywhere else we write `i != end`.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 135 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-Mfh1YJt7wCL_gAcRWJF:-Mfh1YJt7wCL_gAcRWJG:bxu73ys) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77585fba61ad42030b01223b593036e6e52cd743/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L135)):*
> ```cpp
>     std::string table_id_chars =
>         "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_";
>     table_id = "table-" + Sample(generator, 8, table_id_chars);
> ```

I see your self-test creates a table.  Over time these can leak: this program may crash or timeout before having a chance to delete the table.  In other programs that create persistent resources as part of the integration test we try to name the resources something like `foo-YYYY-MM-DD-${random_stuff}` and then delete them if they are older than 2 days.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 154 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-Mfh23zx4HxZ38LYj7nH:-Mfh23zx4HxZ38LYj7nI:b-zeqpwg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77585fba61ad42030b01223b593036e6e52cd743/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L154)):*
> ```cpp
>         return 1;
>       }
>       std::cout << table.status() << std::endl;
> ```

I was suggesting that the table name should always be printed, because the error won't tell you what failed, just the type of error.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 186 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-Mfh4B_6CTOvKehxbBxA:-Mfh4B_6CTOvKehxbBxB:bw24xgq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77585fba61ad42030b01223b593036e6e52cd743/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L186)):*
> ```cpp
>   // Start the MutationBatcher thread
>   CompletionQueue cq;
>   std::thread cq_runner([&cq] { cq.Run(); });
> ```

While you have N threads generating mutations, you only have 1 thread sending RPCs and handling responses to the RPCs. That may be less than ideal.

Consider:

```cc
std::vector<std::vector> background(options->background_threads);
std::generate_n(background.begin(), background.size(), [&cq] { return std::thread{[](CompletionQueue q) { q.Run(); }, cq}; });
```
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_options.cc, line 139 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-Mfh3Po415eOUNhDtq7r:-Mfh3Po415eOUNhDtq7s:bvi2uf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77585fba61ad42030b01223b593036e6e52cd743/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_options.cc#L139)):*
> ```cpp
>     return make_status(os);
>   }
>   if (options.batch_size <= 0 || options.batch_size > 100000) {
> ```

is there a constant somewhere you can use instead of this number?  Otherwise, consider a comment. You and I know where 100,000 comes from, other folks may not.

dbolduc (commented):
The design has changed significantly in this latest commit. I have seen up to a 6x improvement in throughput relative to the previous design. The big differences are that:
1) There is now 1 mutation batcher used per thread.
2) Initial splits are provided to the table
3) The number of background threads used to run the batchers is now configurable. (now it can be > 1)
4) I no longer write to `atomic<int> fails`, `atomic<int> successes` on every response from the server. This was really dumb.

I think ~3x of the speedup is attributable to having multiple splits and batchers. This allows the server to process the requests efficiently because:
1) Each individual batch belongs to a single shard (if your thread count = shard count. otherwise there are edge cases, but they are inconsequential)
2) Batches can be send to different shards at the same time, allowing the backend to parallelize the work in each shard.

I think ~2x of the speedup came from correcting the poor use of atomics, based on some empirical data I collected.

#### Weird things still

* Our deadline timer is run using the batcher background threads. I think this is fine, because we don't care so much about results that timeout anyways. But maybe it should have it's own dedicated thread?
* `GrpcBackgroundThreadPoolSizeOption` is still set to the value of `--max-batches`. Even though there could be N batchers that each have their own max batches in flight limit.... I think this value is now wrong. But I do not know what it should be. I guess it could be yet another configuration parameter? 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-:-Mfs2GBC6-wxug44gKxB:b89cj6t)* status: 4 of 10 files reviewed, 4 unresolved discussions (waiting on @coryan)

___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 135 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-Mfh1YJt7wCL_gAcRWJF:-Mfrswzp5GRO1ht_PjIY:b-rgkg66) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77585fba61ad42030b01223b593036e6e52cd743/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L135)):*

I see your self-test creates a table.  Over time these can leak: this program may crash or timeout before having a chance to delete the table.  In other programs that create persistent resources as part of the integration test we try to name the resources something like `foo-YYYY-MM-DD-${random_stuff}` and then delete them if they are older than 2 days.
</blockquote></details>

Ah. This makes sense. Fixed.

I notice that our [other benchmarks](https://github.com/googleapis/google-cloud-cpp/blob/1295b30097a21451f4b7cf0582660989b91200d7/google/cloud/bigtable/benchmarks/setup.cc#L47-L51) do not exactly match [the pattern](https://github.com/googleapis/google-cloud-cpp/blob/1295b30097a21451f4b7cf0582660989b91200d7/google/cloud/bigtable/testing/random_names.cc#L25) we look for when we [clean up stale tables](https://github.com/googleapis/google-cloud-cpp/blob/1295b30097a21451f4b7cf0582660989b91200d7/google/cloud/bigtable/testing/cleanup_stale_resources.cc#L25-L44). (None of them use the prefix "tbl"). 

The exact prefix of the table was a [requirement](https://github.com/googleapis/google-cloud-cpp/issues/189) many years ago. We could make the regex more general.... but I think we should just update the other benchmarks to use the "tbl" prefix. They needed to be cleaned up, anyway.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 154 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-Mfh23zx4HxZ38LYj7nH:-MfryqYDD03UDkAH2_Ro:bt2vr1s) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77585fba61ad42030b01223b593036e6e52cd743/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L154)):*

I was suggesting that the table name should always be printed, because the error won't tell you what failed, just the type of error.
</blockquote></details>

Fixed?
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 186 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-Mfh4B_6CTOvKehxbBxA:-MfryvK4DAK3zoMBiiBw:b-jk23nl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77585fba61ad42030b01223b593036e6e52cd743/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L186)):*

While you have N threads generating mutations, you only have 1 thread sending RPCs and handling responses to the RPCs. That may be less than ideal.

Consider:

```cc
std::vector<std::vector> background(options->background_threads);
std::generate_n(background.begin(), background.size(), [&cq] { return std::thread{[](CompletionQueue q) { q.Run(); }, cq}; });
```

</blockquote></details>

We discussed this offline. I have added the configuration option `--batcher-thread-count` to address this.
___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_options.cc, line 139 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-Mfh3Po415eOUNhDtq7r:-Mfrz9fh7CK3ohMsrfyT:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77585fba61ad42030b01223b593036e6e52cd743/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_options.cc#L139)):*

is there a constant somewhere you can use instead of this number?  Otherwise, consider a comment. You and I know where 100,000 comes from, other folks may not.
</blockquote></details>

Done.

coryan (approved):
:lgtm: I expect we will need to iterate on this over time, but looks good, thanks!

> Reviewed 5 of 6 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-:-Mft2QQN8ZypC7JLeBB_:b-j0qsqa)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dbolduc)

___
*[google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc, line 135 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7070#-Mfh1YJt7wCL_gAcRWJF:-Mft1t1u5lmJRmEFt2Bq:b-5xouc0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/77585fba61ad42030b01223b593036e6e52cd743/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc#L135)):*

Ah. This makes sense. Fixed.

I notice that our [other benchmarks](https://github.com/googleapis/google-cloud-cpp/blob/1295b30097a21451f4b7cf0582660989b91200d7/google/cloud/bigtable/benchmarks/setup.cc#L47-L51) do not exactly match [the pattern](https://github.com/googleapis/google-cloud-cpp/blob/1295b30097a21451f4b7cf0582660989b91200d7/google/cloud/bigtable/testing/random_names.cc#L25) we look for when we [clean up stale tables](https://github.com/googleapis/google-cloud-cpp/blob/1295b30097a21451f4b7cf0582660989b91200d7/google/cloud/bigtable/testing/cleanup_stale_resources.cc#L25-L44). (None of them use the prefix "tbl"). 

The exact prefix of the table was a [requirement](https://github.com/googleapis/google-cloud-cpp/issues/189) many years ago. We could make the regex more general.... but I think we should just update the other benchmarks to use the "tbl" prefix. They needed to be cleaned up, anyway.
</blockquote></details>

SGTM, please file a bug to fix the other benchmarks and tests.

Issue #6682, opened 2021-06-03
release script usage of hub requires oauth token


scotthart:
release/release.sh invokes `hub`, which will check for an oauth token in ~/.config/hub. Despite some version of hub prompting for username/password when ~/.config/hub is not found, these credentials are ignored. There must exists a ~/.config/hub with the proper oauth token.

devjgm:
I think hub now requires a personal access token rather than a password. https://github.blog/2020-07-30-token-authentication-requirements-for-api-and-git-operations/

PR #7071, opened 2021-07-27
chore(release): change release.sh to use `gh` instead of `hub`

devjgm:
Fixes: #6682

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7071)

### DIFF
--- a/release/release.sh
+++ b/release/release.sh
@@ -1,11 +1,24 @@
 #!/bin/bash
 #
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


release/release.sh
```diff
@@ -1,11 +1,24 @@
 #!/bin/bash
 #
+# Copyright 2020 Google LLC
```
coryan, line 3: FYI: I am not sure the line between `#!/bin/bash` and the Copyright notice is needed, in fact, we do have many files that do not have it.  Your call.

coryan (approved):

release/release.sh
devjgm: I never noticed that inconsistency before. It looks like we have 80 shell scripts total, 54 use a blank line, 26 do not. Not sure which way we prefer, but I'd be happy to clean it up in a separate PR. I don't have a strong preference which way we go... well, https://google.github.io/styleguide/shellguide.html#file-header implies a blank line may be preferred, so maybe that's a nudge.

coryan: SGTM, don't forget the `.py` files.

PR #7075, opened 2021-07-28
cleanup: use a blank comment line between shebang and copyright

devjgm:
This fixes all files that start w/ a `#!`. 

Context: https://github.com/googleapis/google-cloud-cpp/pull/7071#discussion_r677835834

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7075)

### DIFF
--- a/ci/etc/integration-tests-config.sh
+++ b/ci/etc/integration-tests-config.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+#
 # Copyright 2020 Google LLC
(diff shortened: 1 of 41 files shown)


coryan (approved):
I assume none of these was using a line number to store the "Usage" message?

devjgm:
> I assume none of these was using a line number to store the "Usage" message?

I looked for those cases and didn't see any in these files.

PR #7076, opened 2021-07-28
fix(dep): fetch protobuf tarball from the 'protocolbuffers' org

devjgm:
We were fetching the protobuf dep from the "google" github org, and that
worked for years, but today it broke. Changing to fetching from the
`protocolbuffers/` org seems to work and seems like the right thing to
do anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7076)

### DIFF
--- a/ci/cloudbuild/dockerfiles/demo-debian-buster.Dockerfile
+++ b/ci/cloudbuild/dockerfiles/demo-debian-buster.Dockerfile
@@ -97,7 +97,7 @@ RUN curl -sSL https://github.com/nlohmann/json/archive/v3.9.1.tar.gz | \
 
 # ```bash
 WORKDIR /var/tmp/build/protobuf
-RUN curl -sSL https://github.com/google/protobuf/archive/v3.15.8.tar.gz | \
+RUN curl -sSL https://github.com/protocolbuffers/protobuf/archive/v3.15.8.tar.gz | \
     tar -xzf - --strip-components=1 && \
     cmake \
         -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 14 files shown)


ci/cloudbuild/dockerfiles/demo-fedora.Dockerfile
```diff
@@ -127,7 +127,7 @@ RUN curl -sSL https://github.com/nlohmann/json/archive/v3.9.1.tar.gz | \
 
 # ```bash
 WORKDIR /var/tmp/build/protobuf
-RUN curl -sSL https://github.com/google/protobuf/archive/v3.15.8.tar.gz | \
+RUN curl -sSL https://github.com/protocolbuffers/protobuf/archive/v3.15.8.tar.gz | \
```
coryan, line 130: Unrelated, but why is this version so different?

ci/cloudbuild/dockerfiles/demo-debian-buster.Dockerfile
```diff
@@ -97,7 +97,7 @@ RUN curl -sSL https://github.com/nlohmann/json/archive/v3.9.1.tar.gz | \
 
 # ```bash
 WORKDIR /var/tmp/build/protobuf
-RUN curl -sSL https://github.com/google/protobuf/archive/v3.15.8.tar.gz | \
+RUN curl -sSL https://github.com/protocolbuffers/protobuf/archive/v3.15.8.tar.gz | \
```
coryan, line 100: See below, very different version, I wonder why.

coryan (commented):
Drive by comments.

dbolduc (approved):

coryan (approved):

PR #7077, opened 2021-07-28
feat(storage): support setting HTTP version

coryan:
This is an experimental feature to set the HTTP version. Disabling
HTTP/2 can result in better load balancing as load balancers are not
aware of HTTP/2 streams.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7077)

### DIFF
--- a/google/cloud/storage/options.h
+++ b/google/cloud/storage/options.h
@@ -44,7 +44,7 @@ inline namespace STORAGE_CLIENT_NS {
  *
  * [libcurl's default]: https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html
  */
-struct CurlHttpVersion {
+struct CurlHttpVersionOption {
   using Type = std::string;
 };
 }  // namespace STORAGE_CLIENT_NS
@@ -236,7 +236,7 @@ using ClientOptionList = ::google::cloud::OptionList<
     MaximumCurlSocketRecvSizeOption, MaximumCurlSocketSendSizeOption,
     DownloadStallTimeoutOption, RetryPolicyOption, BackoffPolicyOption,
     IdempotencyPolicyOption, CARootsFilePathOption,
-    storage_experimental::CurlHttpVersion>;
+    storage_experimental::CurlHttpVersionOption>;
 
 }  // namespace STORAGE_CLIENT_NS
 }  // namespace storage
--- a/google/cloud/storage/options.h
+++ b/google/cloud/storage/options.h
@@ -44,7 +44,7 @@ inline namespace STORAGE_CLIENT_NS {
  *
  * [libcurl's default]: https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html
  */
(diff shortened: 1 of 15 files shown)


google/cloud/storage/options.h
```diff
@@ -28,6 +28,28 @@
...
+ * - "1.0": use HTTP/1.0, this is not recommended as would require a new
+ *   connection for each request.
+ * - "1.1": use HTTP/1.1, this may be useful when the overhead of HTTP/2 is
+ *   unacceptable. Note that this may require additional connections.
+ * - "2TLS": use HTTP/2 with TLS
+ * - "2.0": use HTTP/2 with our without TLS.
+ *
+ * [libcurl's default]: https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html
+ */
+struct CurlHttpVersion {
```
devjgm, line 47: I think we previously adopted the style that options end w/ an "Options" suffix. So consider making this `CurlHttpVersionOption`.

google/cloud/storage/options.h
```diff
@@ -28,6 +28,28 @@
...
+  using Type = std::string;
```
devjgm, line 48: (optional) Did you consider making this option value a `long` and requiring that it's value be one of the CURL codes, like `CURL_HTTP_VERSION_2_0`? That may require that the caller include some curl header to access that symbol, which could be a bit annoying, but it would avoid ambiguity in the string format (like "2" works but "1" doesn't, but "1.0" and "2.0" both work).

devjgm (approved):

google/cloud/storage/options.h
coryan: No, the thought did not cross my mind, thanks for bringing it up. On balance I think I will keep the string.  This is in `storage_experimental` partly because exposing the underlying implementation (libcurl) seems wrong.  If (big "if" I know) we used a different HTTP library: should we expose the values in that library?  If we use gRPC, what is the value of this option?  Using the CURL enums would be even worse (I think).

I do not have answers to those questions, so string and experimental keep things more loosely coupled until I do.

google/cloud/storage/options.h
coryan: Done.

google/cloud/storage/options.h
devjgm: Keeping string SGTM. I only suggested it because "Curl" is already in the option name, so the implementation is already exposed.

devjgm (approved):

google/cloud/storage/options.h
devbww: I was going to suggest `int major; int minor` and taking "Curl" out of the option name, but I came back to preferring `string`.  Maybe "Curl" could still go in any case.

coryan: Doh! Yeah, that `Curl` prefix was no good. Removed.

devjgm (approved):

PR #7081, opened 2021-07-28
feat(storage): release download handles sooner

### LINKED ISSUE #6160: Release CURL handle as soon as the download completes
In a `CurlDownloadRequest` the curl handle is released in the destructor:

https://github.com/googleapis/google-cloud-cpp/blob/9ebc00b04b4fd2767c5bee6acdd1172407e6b9eb/google/cloud/storage/internal/curl_download_request.h#L56-L57

They can be released earlier, i.e., once the download completes.  That makes the resources usable to other streams sooner.

coryan:
Release the CURL* handles as soon as we are done using them. The
downloads used to hold on to these until the `CurlDownloadRequest`
object was deleted, which in turn could live until the
`storage::ObjectReadStream` was deleted.

Fixes #6160

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7081)

### DIFF
--- a/google/cloud/storage/internal/curl_download_request.cc
+++ b/google/cloud/storage/internal/curl_download_request.cc
@@ -62,10 +62,7 @@ CurlDownloadRequest::CurlDownloadRequest()
 
 CurlDownloadRequest::~CurlDownloadRequest() {
   CleanupHandles();
-  if (factory_) {
-    factory_->CleanupHandle(std::move(handle_));
-    factory_->CleanupMultiHandle(std::move(multi_));
-  }
+  ReleaseHandles();
 }
 
 StatusOr<HttpResponse> CurlDownloadRequest::Close() {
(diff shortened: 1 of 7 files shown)


google/cloud/storage/internal/curl_download_request.cc
```diff
@@ -98,23 +95,29 @@ StatusOr<HttpResponse> CurlDownloadRequest::Close() {
 }
 
 StatusOr<ReadSourceResult> CurlDownloadRequest::Read(char* buf, std::size_t n) {
+  GCP_LOG(WARNING) << __func__ << "()";
```
coryan, line 98: @devjgm hilariously, removing this like breaks the test.  Considering this is a single-threaded program I am starting to think that we should stop using computers.

coryan:
I will be making many changes to this branch.  Closing because I do not want to waste CI resources.

PR #7082, opened 2021-07-29
cleanup(storage): simplify CurlDownloadRequest creation

coryan:
Other than in tests, we always used the class in a `std::unique_ptr<>`.
We can make the class non-copyable and non-moveable, and simply return a
`std::unique_ptr<CurlDownloadRequest>` from the `CurlRequestBuilder`.
This will also workaround some bugs in MSVC (or at least I believe they
are bugs).

This is part of the work for #6160

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7082)

### DIFF
--- a/google/cloud/storage/internal/curl_download_request.h
+++ b/google/cloud/storage/internal/curl_download_request.h
@@ -48,12 +48,14 @@ extern "C" std::size_t CurlDownloadRequestHeader(char* contents,
  */
 class CurlDownloadRequest : public ObjectReadSource {
  public:
-  explicit CurlDownloadRequest();
+  CurlDownloadRequest(CurlHeaders headers, CurlHandle handle, CurlMulti multi);
 
   ~CurlDownloadRequest() override;
 
-  CurlDownloadRequest(CurlDownloadRequest&&) = default;
-  CurlDownloadRequest& operator=(CurlDownloadRequest&& rhs) = default;
+  CurlDownloadRequest(CurlDownloadRequest&&) = delete;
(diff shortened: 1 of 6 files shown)


google/cloud/storage/internal/curl_download_request.h
```diff
@@ -48,12 +48,14 @@ extern "C" std::size_t CurlDownloadRequestHeader(char* contents,
...
-  explicit CurlDownloadRequest();
+  CurlDownloadRequest(CurlHeaders headers, CurlHandle handle, CurlMulti multi);
 
   ~CurlDownloadRequest() override;
 
-  CurlDownloadRequest(CurlDownloadRequest&&) = default;
-  CurlDownloadRequest& operator=(CurlDownloadRequest&& rhs) = default;
+  CurlDownloadRequest(CurlDownloadRequest&&) = delete;
+  CurlDownloadRequest& operator=(CurlDownloadRequest&&) = delete;
+  CurlDownloadRequest(CurlDownloadRequest const&) = delete;
```
devjgm, line 57: I suspect that this class used to be copyable. Do you think that may have been the issue? A copy showing up where it wasn't expected?

devjgm (approved):

google/cloud/storage/internal/curl_download_request.h
devjgm: Bummer. There's also the `buffer_` member, which seems like a `=default` move c'tor may not do the right thing with.

Regardless, I like the change of making this not moveable or copyable and avoiding all these guesses.

PR #7083, opened 2021-07-29
doc(iam): polish GA release

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7083)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,12 +59,6 @@
   is now GA.
 * *IAM* -- The [IAM libray](https://github.com/googleapis/google-cloud-cpp/tree/main/google/cloud/iam) is now GA.
 
-### BigQuery:
-
-**BREAKING CHANGES**:
-* The `experimental-bigquery` target has been replaced with the target
-  `bigquery`. Current usages will need to be updated.
-
 ### Bigtable:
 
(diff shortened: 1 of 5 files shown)


devjgm (approved):

CHANGELOG.md
```diff
@@ -77,6 +78,12 @@
 
 [issue-2567]: https://github.com/googleapis/google-cloud-cpp/issues/2567
 
+### IAM:
+
+**BREAKING CHANGES**:
+* The `experimental-iam` target has been replaced with the target `iam`. Current
```
devjgm, line 84: I believe this only applies to bazel builds, since the `experimental-iam` target still exists in cmake. You could clarify that here. Or, you could add the `experimental-{iam,bigquery}` targets back to //BUILD so there is no breaking change this release, and the bazel and cmake targets would remain in sync. Then after next weeks release you could then remove the `experimental-{iam,bigquery}` targets from BOTH bazel and cmake, and add the **BREAKING CHANGES** doc for the following month's release. I think that latter option is what I'd do, but I think I've caused you enough headache, so I'll approve and let you decide how you want to proceed :-) Thanks.

devjgm (approved):

PR #7087, opened 2021-07-29
feat(common): always clog GCP_LOG(FATAL) messages

devbww:
It is desirable that FATAL log messages are never discarded,
especially now that GCP_LOG(FATAL) calls `std::abort` (see #7058).

To that end, ...

- Don't allow FATAL logging to be compile-time disabled.

- Always instantiate a `StdClogBackend` in `DefaultLogBackend()`.

- Give that default `StdClogBackend` a "mininum log severity" from
  the first of the following conditions:
    * If `${GOOGLE_CLOUD_CPP_EXPERIMENTAL_LOG_CONFIG}` is valid,
      use the lowest enabled severity.
    * If `${GOOGLE_CLOUD_CPP_ENABLE_CLOG}` is a valid severity,
      use that.
    * If `${GOOGLE_CLOUD_CPP_ENABLE_CLOG}` is otherwise set,
      use the lowest enabled severity (backwards compatibility).
    * If `${GOOGLE_CLOUD_CPP_ENABLE_CLOG}` is unset, use FATAL.

This means that FATAL messages should always be logged. But note
that `LogSink::DisableStdClog()` and `LogSink::ClearBackends()`
can remove that `StdClogBackend`, so they should be used with care
(undoubtedly only from tests).

Also tidy up some tests so that the disposition of `LogBackend`s
is a little clearer.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7087)

### DIFF
--- a/google/cloud/internal/log_impl.cc
+++ b/google/cloud/internal/log_impl.cc
@@ -22,9 +22,11 @@ namespace internal {
 
 void StdClogBackend::Process(LogRecord const& lr) {
   std::lock_guard<std::mutex> lk(mu_);
-  std::clog << lr << "\n";
-  if (lr.severity >= Severity::GCP_LS_WARNING) {
-    std::clog << std::flush;
+  if (lr.severity >= min_severity_) {
+    std::clog << lr << "\n";
+    if (lr.severity >= Severity::GCP_LS_WARNING) {
+      std::clog << std::flush;
+    }
   }
 }
 
--- a/google/cloud/internal/log_impl_test.cc
+++ b/google/cloud/internal/log_impl_test.cc
@@ -68,62 +68,85 @@ TEST(DefaultLogBackend, CircularBuffer) {
   ScopedEnvironment config(kLogConfig, "lastN,5,WARNING");
   ScopedEnvironment clog(kEnableClog, absl::nullopt);
   auto be = DefaultLogBackend();
-  EXPECT_TRUE(!!be);
   auto const* buffer = dynamic_cast<CircularBufferBackend*>(be.get());
   ASSERT_NE(buffer, nullptr);
   EXPECT_EQ(5, buffer->size());
   EXPECT_EQ(Severity::GCP_LS_WARNING, buffer->min_flush_severity());
+  auto const* clog_be = dynamic_cast<StdClogBackend*>(buffer->backend().get());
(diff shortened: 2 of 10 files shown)


google/cloud/internal/log_impl_test.cc
```diff
@@ -68,62 +68,85 @@ TEST(DefaultLogBackend, CircularBuffer) {
...
 }
 
 TEST(DefaultLogBackend, CLog) {
   ScopedEnvironment config(kLogConfig, "clog");
   ScopedEnvironment clog(kEnableClog, absl::nullopt);
   auto be = DefaultLogBackend();
-  EXPECT_TRUE(!!be);
-  EXPECT_NE(dynamic_cast<StdClogBackend*>(be.get()), nullptr);
+  auto const* clog_be = dynamic_cast<StdClogBackend*>(be.get());
+  ASSERT_NE(clog_be, nullptr);
```
coryan, line 85: nit: here and below, consider `ASSERT_THAT(clog_be, IsNotNull());`

coryan: And the one above too.

google/cloud/internal/log_impl.cc
```diff
@@ -22,9 +22,11 @@ namespace internal {
 
 void StdClogBackend::Process(LogRecord const& lr) {
   std::lock_guard<std::mutex> lk(mu_);
-  std::clog << lr << "\n";
-  if (lr.severity >= Severity::GCP_LS_WARNING) {
-    std::clog << std::flush;
+  if (lr.severity >= min_severity_) {
```
coryan, line 25: optional: `if (lr.severity < min_severity_) return;` reduces indentation.

coryan (approved):
Thanks for this change.  A couple of minor nits, completely your choice.

google/cloud/internal/log_impl_test.cc
devbww: Done.

google/cloud/internal/log_impl.cc
devbww: I often find the single-entry/exit and indentation (within reason) clearer, and this is such an example, so I left it alone.

PR #7091, opened 2021-07-30
feat(storage): more tuning parameters for benchmark

coryan:
Add options to prefer HTTP/1.1 over HTTP/2 when using REST, and to use a
different `gcs::Client` when using either gRPC or REST.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7091)

### DIFF
--- a/google/cloud/storage/benchmarks/aggregate_throughput_options.cc
+++ b/google/cloud/storage/benchmarks/aggregate_throughput_options.cc
@@ -75,6 +75,14 @@ ParseAggregateThroughputOptions(std::vector<std::string> const& argv,
        [&options](std::string const& val) {
          options.grpc_plugin_config = val;
        }},
+      {"--rest-http-version", "change the preferred HTTP version",
+       [&options](std::string const& val) { options.rest_http_version = val; }},
+      {"--client-per-thread",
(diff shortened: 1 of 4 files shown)


google/cloud/storage/benchmarks/aggregate_throughput_options.cc
```diff
@@ -75,6 +75,14 @@ ParseAggregateThroughputOptions(std::vector<std::string> const& argv,
        [&options](std::string const& val) {
          options.grpc_plugin_config = val;
        }},
+      {"--rest-http-version", "change the preferred HTTP version",
+       [&options](std::string const& val) { options.rest_http_version = val; }},
+      {"--client-per-thread",
+       "use a different storage::Client object in each thread",
+       [&options](std::string const& val) {
+         options.client_per_thread =
+             testing_util::ParseBoolean(val).value_or("false");
```
dbolduc, line 84: I would just check if the flag is present

dbolduc (approved):

google/cloud/storage/benchmarks/aggregate_throughput_options.cc
coryan: Done.

PR #7093, opened 2021-07-31
testing(common): two small ScopedLog::Backend::Process() optimizations

devbww:
1. Split the `LogRecord::message` outside the lock.

2. Move the split lines into `log_lines_` instead of copying.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7093)


coryan (approved):
I am about to sign off, so I did a drive-by review. LGTM.

PR #7099, opened 2021-08-02
ci(check-api): add IAM and BigQuery to check-api build

devjgm:
I also needed to teach the script to skip the api-compatibility check when the "expected" file was missing (but it still gets created for the subsequent runs).

I also changed `gzip` to use `-n`, otherwise, the timestamp (and file name) are included in the compressed archive, which results in a different md5sum for every run of this script, even when nothing changes. This results in git thinking that the compressed dump files changed even when they didn't.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7099)


coryan (approved):

ci/cloudbuild/builds/check-api.sh
```diff
@@ -72,27 +74,29 @@ for lib in "${libraries[@]}"; do
...
+    zcat "${expected_dump_path}" >"cmake-out/${expected_dump_file}"
+    # We ignore all symbols in internal namespaces, because these are not part
+    # of our public API. We do this by specifying a regex that matches against
+    # the mangled symbol names. For example, 8 is the number of characters in
+    # the string "internal", and it should again be followed by some other
+    # number indicating the length of the symbol within the "internal"
+    # namespace. See: https://en.wikipedia.org/wiki/Name_mangling
+    report="cmake-out/compat_reports/${lib}/expected_to_actual/src_compat_report.html"
+    if ! abi-compliance-checker \
+      -skip-internal-symbols "(8internal|15pubsub_internal|16spanner_internal)\d" \
```
dbolduc, line 87: were you going to add BQ, IAM to this?

devjgm: done

dbolduc (approved):

scotthart (approved):
> Reviewed 8 of 9 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7099)* status: all files reviewed, 1 unresolved discussion (waiting on @dbolduc)

Issue #4216, opened 2020-05-18
GCS+gRPC plugin should timeout uploads with no progress


coryan:
As we do for REST, the timeout should not be based on the total time for the upload, but a minimum effective bandwidth.

PR #7108, opened 2021-08-03
feat(GCS+gRPC): initial support for download timeouts

coryan:
Timeouts for downloads (or any streaming operation) are tricky. A
download can take anywhere from a few milliseconds, to tens of thousands
of seconds. For the REST-based implementation we do not use timeouts for
the full operation, we use a "minimum observed bandwidth" requirement.
There does not seem to a an easy way to implement this in gRPC. While I
work in an implementation setting a timeout (of any kind) is a good
start: at least the programs won't hang forever waiting for stalled
downloads.

Part of the work for #4216

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7108)

### DIFF
--- a/google/cloud/storage/CMakeLists.txt
+++ b/google/cloud/storage/CMakeLists.txt
@@ -602,6 +602,7 @@ if (GOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC)
             internal/grpc_client_failures_test.cc
             internal/grpc_client_insert_object_media_test.cc
             internal/grpc_client_object_request_test.cc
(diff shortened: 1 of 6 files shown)


devjgm:
I was experimenting and broke your windows builds. sorry. I restarted them.

coryan:
> I was experimenting and broke your windows builds. sorry. I restarted them.

Thanks!  I was about to look and dreading the thought I had broken just Windows.

devjgm (approved):

PR #7113, opened 2021-08-04
feat(storage): make benchmark more robust on stalls

coryan:
Change the benchmark to use a single queue (LIFO) with the set of
objects to download, and have all the threads service this queue.
Previously we separated the objects in groups and assigned them to
threads. The previous approach wastes threads if one (or a small
fraction) of the downloads gets stalled. I think the new approach better
models a well-designed application.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7113)

### DIFF
--- a/google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc
+++ b/google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc
@@ -82,7 +82,6 @@ google::cloud::StatusOr<AggregateThroughputOptions> ParseArgs(int argc,
 struct TaskConfig {
   gcs::Client client;
   std::seed_seq::result_type seed;
-  std::vector<gcs::ObjectMetadata> objects;
 };
 
 using Counters = std::map<std::string, std::int64_t>;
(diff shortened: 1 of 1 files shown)


dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7113)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @coryan)

___
*[google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc, line 257 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7113#-MgEEV8F7x7Ut2ArsxOQ:-MgEEV8F7x7Ut2ArsxOR:b85clji) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/02f004a5096e4afb7cac582e5683b2eaaec63c15/google/cloud/storage/benchmarks/aggregate_throughput_benchmark.cc#L257)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>

> ```
>     std::transform(configs.begin(), configs.end(), tasks.begin(),
>                    [&task](TaskConfig const& c) {
>                      return std::async(std::launch::async, task, std::cref(c));
>                    });
> ```
</details>


if you did something like this, you can get rid of the`task` lambda:
```
std::transform(configs.begin(), configs.end(), tasks.begin(),
               [&iteration](TaskConfig const& c) {
                 return std::async(std::launch::async, &Iteration::DownloadTask, iteration, std::cref(c));
               });
```
The aesthetic judgment is up to you. I don't have a preference.

PR #7116, opened 2021-08-04
ci(checkers): trim trailing blank lines from text files

devjgm:
This PR
* Updates the `ci/cloudbuild/builds/checkers.sh` script to trim trailing blank lines from text files
* Fixes all the files in the repo that had trailing blank lines
* Updates the generator so that it doesn't generate files w/ trailing blank lines

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7116)

### DIFF
--- a/ci/cloudbuild/builds/checkers.sh
+++ b/ci/cloudbuild/builds/checkers.sh
@@ -141,7 +141,7 @@ time {
   # http://sed.sourceforge.net/sed1line.txt.
   expressions+=("-e" ":a" "-e" "'/^\n*$/{\$d;N;ba'" "-e" "'}'")
   git ls-files -z | grep -zv '\.gz$' |
-    xargs -P "$(nproc)" -n 50 -0 grep -ZPL "\bDO NOT EDIT\b" |
+    (xargs -P "$(nproc)" -n 50 -0 grep -ZPL "\bDO NOT EDIT\b" || true) |
     xargs -P "$(nproc)" -n 50 -0 bash -c "sed_edit ${expressions[*]} \"\$0\" \"\$@\""
 }
 
--- a/ci/cloudbuild/builds/checkers.sh
+++ b/ci/cloudbuild/builds/checkers.sh
@@ -137,8 +137,14 @@ time {
 # snippets).
 printf "%-30s" "Running whitespace fixes:" >&2
 time {
+  expressions=("-e" "'s/[[:blank:]]\+$//'")
+  # The next three expressions trim trailing blank lines, courtesy of
+  # http://sed.sourceforge.net/sed1line.txt.
+  expressions+=("-e" ":a")
+  expressions+=("-e" "'/^\n*$/{\$d;N;ba'")
+  expressions+=("-e" "'}'")
   git ls-files -z | grep -zv '\.gz$' |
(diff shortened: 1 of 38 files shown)


ci/cloudbuild/builds/checkers.sh
```diff
@@ -131,14 +131,18 @@ time {
     xargs -P "$(nproc)" -n 50 -0 bash -c "sed_edit ${expressions[*]} \"\$0\" \"\$@\""
 }
 
-# Apply transformations to fix whitespace formatting in files not handled by
-# clang-format(1) above.  For now we simply remove trailing blanks.  Note that
-# we do not expand TABs (they currently only appear in Makefiles and Makefile
-# snippets).
+# Applies the following fixes in text files that do not say "DO NOT EDIT":
```
coryan, line 134: Something about barbers that only shave those that do not shave themselves... j/k

coryan: also, s/do not say/unless they say/ maybe?

ci/cloudbuild/builds/checkers.sh
```diff
@@ -131,14 +131,18 @@ time {
...
+#   - Removes trailing whitespace on lines
+#   - Removes trailing blank lines in files
 printf "%-30s" "Running whitespace fixes:" >&2
 time {
+  expressions=("-e" "'s/[[:blank:]]\+$//'")
+  # Trims trailing blank lines, courtesy of
+  # http://sed.sourceforge.net/sed1line.txt.
+  expressions+=("-e" ":a" "-e" "'/^\n*$/{\$d;N;ba'" "-e" "'}'")
```
coryan, line 142: I of course do not understand if that works, or why, nor do I understand why anybody would want to understand why it works either.  If we ever have to fix this, I would remove the whole thing and start over in C++ or awk or python or something.

coryan (approved):

ci/cloudbuild/builds/checkers.sh
devjgm: "do not shave themselves"... hilarious! I didn't realize it, but it's true here :-)

Also fixed wording.

ci/cloudbuild/builds/checkers.sh
devjgm: Ack to all of that. I don't understand it either. If it breaks, we can just remove the line and go back to not caring (which I'm fine w/ even now).

devbww: It boils down to ...
```
while (!eof) {
  read line into pattern space;
  while (pattern space matches /^\n*$/) {
    if (eof) {
      clear pattern space;
      break;
    }
    append line to pattern space;
  }
  print pattern space;
}
```
So, we're accumulating empty lines into the pattern space, and if we reach end-of-file before anything else we'll discard them.

The only thing I might suggest is coalescing it all into a single `-e` argument:
```
  expressions+=('-e' ':a;/^\n*$/{$d;N;ba;}')
```

devjgm: Done.

ci/cloudbuild/builds/checkers.sh
devjgm: Used a character class in the regex to avoid the shaving barbers problem, now `\b[D]O NOT EDIT\b` won't match this file itself.

devbww (approved):

PR #7118, opened 2021-08-05
fix(spanner): fix Client::OverlayQueryOptions() to merge correctly

devbww:
`Client::OverlayQueryOptions()` was misbehaving on two fronts:
- The environment option for `optimizer_statistics_package` was
  using the value of `${SPANNER_OPTIMIZER_VERSION}` instead of
  `${SPANNER_OPTIMIZER_STATISTICS_PACKAGE}`.
- It was not setting an output value for `request_priority` at all.

At least part of the reason for these errors was that the test:
- assumed the behavior could be changed by modifying the environment
  whereas the implementation only read the environment once, and
- only a single `MockConnection`, with accumulated parameter matches,
  was used across all test cases.

So, fix the implementation, and revamp the test to be more direct
(which eliminates the mocking) and to exercise RequestPriority.

Also add a note to `query_options.h` to update `OverlayQueryOptions()`
whenever new options are added.

Finally, add documentation for `${SPANNER_OPTIMIZER_STATISTICS_PACKAGE}`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7118)

### DIFF
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -643,6 +643,11 @@ class Client {
       SqlStatement statement, QueryOptions const& opts = {});
 
  private:
+  friend class OverlayQueryOptionsTester;
+  static QueryOptions OverlayQueryOptions(
+      QueryOptions const& preferred, QueryOptions const& fallback,
+      absl::optional<std::string> const& optimizer_version_env,
+      absl::optional<std::string> const& optimizer_statistics_package_env);
   QueryOptions OverlayQueryOptions(QueryOptions const&);
 
   std::shared_ptr<Connection> conn_;
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -643,11 +643,6 @@ class Client {
       SqlStatement statement, QueryOptions const& opts = {});
 
  private:
-  friend class OverlayQueryOptionsTester;
-  static QueryOptions OverlayQueryOptions(
-      QueryOptions const& preferred, QueryOptions const& fallback,
-      absl::optional<std::string> const& optimizer_version_env,
-      absl::optional<std::string> const& optimizer_statistics_package_env);
   QueryOptions OverlayQueryOptions(QueryOptions const&);
 
   std::shared_ptr<Connection> conn_;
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/client.h
```diff
@@ -643,6 +643,11 @@ class Client {
       SqlStatement statement, QueryOptions const& opts = {});
 
  private:
+  friend class OverlayQueryOptionsTester;
+  static QueryOptions OverlayQueryOptions(
```
coryan, line 647: Alternatively, put this static function in the `spanner_internal` namespace and you don't need the `friend` helper, your call.

google/cloud/spanner/query_options.h
```diff
@@ -30,6 +30,9 @@ inline namespace SPANNER_CLIENT_NS {
  * These QueryOptions allow users to configure features about how their SQL
  * queries executes on the server.
  *
+ * Note: If you add an attribute here, remember to update the implementation
```
coryan, line 33: Ugh... that will make it to the public documentation too... Can you add the comment to a `.cc` file or to the beginning of the `private:` section?  If not, can you add it after the Doxygen block as non-Doxygen comment?

google/cloud/spanner/client_test.cc
```diff
@@ -1026,59 +1034,86 @@ TEST(ClientTest, ProfileQueryWithOptionsSuccess) {
...
-      {{}, {}, {}, {}},
-      {"env", {}, {}, "env"},
-      {{}, "client", {}, "client"},
-      {{}, {}, "function", "function"},
-      {"env", "client", {}, "client"},
-      {"env", {}, "function", "function"},
-      {{}, "client", "function", "function"},
-      {"env", "client", "function", "function"},
-  };
+  constexpr auto OverlayQueryOptions =  // NOLINT(readability-identifier-naming)
```
coryan, line 1037: I am sure you understand the tradeoffs between testing the function that computes the query options vs. verifying the correct query options are sent to the `SpannerConnection`.  This note is just a nudge to consider these tradeoffs one last time before you pull the trigger.  I am good either way.

coryan (commented):
A lot of nits.  The thing about notes to library developers in the Doxygen documentation needs to be fixed, the other stuff is optional.

google/cloud/spanner/client.h
devbww: Done.

google/cloud/spanner/query_options.h
devbww: There is no `.cc` for this interface, so `private:` section it is.

devjgm (approved):
Thanks for fixing.

coryan (approved):

google/cloud/spanner/client_test.cc
devbww: There is a separate test, `ConnectionImplTest.QueryOptions`, for verifying how a final `QueryOptions` value is propagated into the protos.  It is deficient in the `request_priority` department however, so I'll fix that in another PR.

PR #7119, opened 2021-08-05
feat(common): capture thread that creates log record

coryan:
When using the `lastN` backend one wants the id of the thread that
created the log, not the id of the thread that is flushing the logs.
Arguably we may also want a `lastNInThread` logger, at least some of the
the time, but that is a matter for a separate PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7119)

### DIFF
--- a/google/cloud/internal/log_impl_test.cc
+++ b/google/cloud/internal/log_impl_test.cc
@@ -35,32 +35,34 @@ auto constexpr kEnableClog = "GOOGLE_CLOUD_CPP_ENABLE_CLOG";
 TEST(CircularBufferBackend, Basic) {
   auto be = std::make_shared<ScopedLog::Backend>();
   CircularBufferBackend buffer(3, Severity::GCP_LS_ERROR, be);
-  buffer.ProcessWithOwnership(LogRecord{
-      Severity::GCP_LS_INFO, "test_function()", "file", 1, {}, "msg 1"});
-  buffer.ProcessWithOwnership(LogRecord{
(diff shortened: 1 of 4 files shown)


google/cloud/internal/log_impl_test.cc
```diff
@@ -35,32 +35,34 @@ auto constexpr kEnableClog = "GOOGLE_CLOUD_CPP_ENABLE_CLOG";
...
+  buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_DEBUG, "msg 2"));
+  buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_TRACE, "msg 3"));
+  buffer.ProcessWithOwnership(
+      test_log_record(Severity::GCP_LS_WARNING, "msg 4"));
+  buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_INFO, "msg 5"));
   EXPECT_THAT(be->ExtractLines(), IsEmpty());
 
-  buffer.ProcessWithOwnership(LogRecord{
-      Severity::GCP_LS_ERROR, "test_error()", "file", 1, {}, "msg 6"});
+  buffer.ProcessWithOwnership(LogRecord{Severity::GCP_LS_ERROR,
```
devjgm, line 51: Is there a reason that this call can't also use `test_log_record()`?

devjgm (approved):

google/cloud/internal/log_impl_test.cc
coryan: Nope, fixed.

PR #7128, opened 2021-08-06
fix(GCS+gRPC): return status on stream closure

coryan:
The GCS+gRPC version of `ObjectReadSource` was behaving differently from
the REST version. It was returning the last bytes received before the
stream closed, which sets `IsOpen()` to return `false`. The caller
checks `IsOpen` and assumes the last returned status indicates whether
there was an error.

It is possible we want the return type to be `StatusAnd`, but that would
require changing all kinds of APIs that we expose for mocking. Better to
have GCS+gRPC and REST behave consistently, if a bit weirdly.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7128)

### DIFF
--- a/google/cloud/storage/internal/grpc_object_read_source.cc
+++ b/google/cloud/storage/internal/grpc_object_read_source.cc
@@ -100,19 +100,9 @@ StatusOr<ReadSourceResult> GrpcObjectReadSource::Read(char* buf,
     headers.insert(h.begin(), h.end());
   }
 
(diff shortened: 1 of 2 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7128)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #7134, opened 2021-08-06
refactor(stream_range): remove unnecessary noexcept spec

devjgm:
`=default` will implicitly set an appropriate noexcept specification, so the `IsMoveNoexcept()` function is unnecessary and possibly even subtly wrong. The clang-tidy warning `performance-noexcept-move-constructor` incorrectly triggers on this case, so the NOLINTs are still needed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7134)

### DIFF
--- a/google/cloud/stream_range.h
+++ b/google/cloud/stream_range.h
@@ -86,14 +86,6 @@ StreamRange<T> MakeStreamRange(StreamReader<T>);
  */
 template <typename T>
 class StreamRange {
-  // Helper that returns true if StreamRange's move constructor and assignment
-  // operator should be declared noexcept.
-  template <typename U>
-  static constexpr bool IsMoveNoexcept() {
-    return noexcept(StatusOr<U>(std::declval<U>()))&& noexcept(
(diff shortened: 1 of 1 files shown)


google/cloud/stream_range.h
```diff
@@ -165,9 +157,9 @@ class StreamRange {
   StreamRange(StreamRange const&) = delete;
   StreamRange& operator=(StreamRange const&) = delete;
   // NOLINTNEXTLINE(performance-noexcept-move-constructor)
-  StreamRange(StreamRange&&) noexcept(IsMoveNoexcept<T>()) = default;
+  StreamRange(StreamRange&&) = default;
```
coryan, line 160: nit: maybe you can move the `// NOLINTNEXTLINE(...)` to `// NOLINT` ... on the next line. Ha! that rhymes.

coryan (approved):

google/cloud/stream_range.h
devjgm: I tried but it's still too long and will wrap.

PR #7135, opened 2021-08-06
chore: back port CI changes

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7135)

### DIFF
--- a/ci/kokoro/macos/builds/bazel.sh
+++ b/ci/kokoro/macos/builds/bazel.sh
@@ -41,7 +41,6 @@ bazel_args=(
 
 readonly CONFIG_DIR="${KOKORO_GFILE_DIR:-/private/var/tmp}"
 readonly TEST_KEY_FILE_JSON="${CONFIG_DIR}/kokoro-run-key.json"
-readonly TEST_KEY_FILE_P12="${CONFIG_DIR}/kokoro-run-key.p12"
 readonly BAZEL_CACHE="https://storage.googleapis.com/cloud-cpp-bazel-cache"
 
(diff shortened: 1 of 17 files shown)


coryan:
It seems like there are two unrelated changes in this PR. One is a change of the CI scripts to backport recent changes to `v1.30.x`, another is the change described here to change the CMake targets.  I think we should split them in two PRs, thoughts?

scotthart (commented):
The only reason I am back porting the scripts is so that CI runs successfully with my cmake changes.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7135#-:-MggYLX8CL-8TVKJ6zAZ:b-qfc37i)* status: 0 of 17 files reviewed, all discussions resolved

ci/kokoro/macos/builds/bazel.sh
```diff
@@ -84,8 +83,7 @@ bazelisk build "${bazel_args[@]}" ...
 
 should_run_integration_tests() {
   if [[ -r "${GOOGLE_APPLICATION_CREDENTIALS}" && -r \
-    "${TEST_KEY_FILE_JSON}" && -r \
-    "${TEST_KEY_FILE_P12}" ]]; then
```
devjgm, line 88: fyi, the p12 key files do not exist anymore, so this test will be false and integration tests will not be run.

devjgm: Please ignore. I was confused.

coryan (approved):

PR #7138, opened 2021-08-07
cleanup(noexcept): remove noexcept from defaulted functions

devjgm:
Using `=default` implicitly sets the Right noexcept specification for a
function. So by declaring a function explicitly with `noexcept =
default` we're either being redundant, or we're possibly _incorrectly_
declaring the specification. For example, a class template should likely
not have a `noexcept(true)` move constructor because the member `T` may
not be noexcept. Omitting the specification and letting `=default` do
the Right Thing is safe, whereas declaring with `noexcept = default`
would be wrong for some `T`.

Similarly, `noexcept = delete` is not necessary since deleted functions
can't throw anyway, and functions can't be overloaded based on the
exception specification.

In general, I think the right style rule to follow is: Only declare a
noexcept specification if/when implementing the body of the function. If
using `=default`, rely on _it_ generating the correct noexcept
specification.

The `performance-noexcept-move-constructor` clang-tidy checkers gets all
the `=default` cases wrong when they're in class templates, so we have
to ignore them.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7138)


coryan (approved):
Looks like everything will build.

Unrelated to this PR: the build time is close to 90m now, we really need to think about how to keep build times under control as we add libraries.

PR #7141, opened 2021-08-08
test(spanner): extend and generalize ConnectionImplTest.QueryOptions

devbww:
Extend `ConnectionImplTest.QueryOptions` to cover all `Client::*` calls
that are affected by `spanner::QueryOptions` and the matching elements of
`spanner::ReadOptions` and `spanner::CommitOptions`, plus the stand-alone
`spanner::RequestPriorityOption`.

Also generalize the protobuf builder so that it will be easier to add
new option fields in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7141)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl_test.cc
+++ b/google/cloud/spanner/internal/connection_impl_test.cc
@@ -83,6 +83,28 @@ using ::testing::Unused;
 
 namespace spanner_proto = ::google::spanner::v1;
 
+// A simple protobuf builder.
+template <typename T>
+class ProtoBuilder {
+ public:
+  template <typename ArgT0, typename... ArgT>
+  ProtoBuilder&& Set(void (T::*set)(ArgT0&&, ArgT&&...), ArgT0&& v,
+                     ArgT&&... args) && {
+    (t_.*set)(std::forward<ArgT0>(v), std::forward<ArgT>(args)...);
+    return std::move(*this);
+  }
+  template <typename ArgT>
+  ProtoBuilder&& Set(void (T::*set)(ArgT), ArgT v) && {
+    (t_.*set)(std::move(v));
+    return std::move(*this);
+  }
+
+  T Build() && { return std::move(t_); }
+
+ private:
+  T t_;
+};
+
 // Matchers for mock calls.
 MATCHER_P(HasSession, session, "request has expected session name") {
   return arg.session() == session;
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -83,6 +83,28 @@ using ::testing::Unused;
 
 namespace spanner_proto = ::google::spanner::v1;
 
+// A simple protobuf builder.
```
coryan, line 86: I like this, I may not call it "simple", but I like it.

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -801,23 +867,84 @@ TEST(ConnectionImplTest, QueryOptions) {
...
+              Return(ByMove(absl::make_unique<NiceMock<MockGrpcReader>>())))
+          .RetiresOnSaturation();
+
+      // Commit().
+      EXPECT_CALL(*mock, Commit(_, commit_request_matcher))
+          .Times(1)
+          .RetiresOnSaturation();
+    }
+
+    (void)conn->ExecuteQuery(sql_params);
```
coryan, line 939: Maybe these should be separate tests, but I can see how that would be a lot more boilerplate code. No action needed.

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -706,88 +728,132 @@ TEST(ConnectionImplTest, ExecuteQueryImplicitBeginTransaction) {
...
   struct {
-    // Given this for SqlParams::query_options ...
+    // Given these QueryOptions ...
     spanner::QueryOptions options;
-
-    // ... expect these for ExecuteSqlRequest query_options/request_options.
+    // ... expect these protos.
     spanner_proto::ExecuteSqlRequest::QueryOptions qo_proto;
     spanner_proto::RequestOptions ro_proto;
   } test_cases[] = {
```
coryan, line 742: At some point maybe you would need to split this test cases into separate function, and use `TEST_P()` with the parameter being the values in this array.  Just a thought for the future, right now it is probably six of one, half a dozen of the other.

coryan (approved):

google/cloud/spanner/internal/connection_impl_test.cc
devbww: Acknowledged.

PR #7146, opened 2021-08-09
test(pubsub): deflake pubsub default batching test

### LINKED ISSUE #7021: [Flake] pubsub batching_publisher_connection_test timeout (macOS)
https://source.cloud.google.com/results/invocations/3d8fa2d2-6f23-4318-b8d9-9d6356611c8f/targets

```
342/343 Test #377: storage_parallel_uploads_test .............................................   Passed   15.34 sec
343/343 Test #173: pubsub_internal_batching_publisher_connection_test ........................***Timeout 1500.08 sec
[==========] Running 12 tests from 1 test suite.
[----------] 12 tests from BatchingPublisherConnectionTest
[ RUN      ] BatchingPublisherConnectionTest.DefaultMakesProgress


99% tests passed, 1 tests failed out of 343

Total Test time (real) = 1507.66 sec

The following tests FAILED:
	173 - pubsub_internal_batching_publisher_connection_test (Timeout)
Errors while running CTest
2021-07-19T15:07:49Z (+2671s): build script reported errors.
```

dbolduc:
Fixes #7021 

Basically the same as #7008, the problem was that super rarely the two messages would be sent in two batches instead of one, due to the hold timer. And, previously, we would only get the future for one batch.

I considered using a CQ that is manually started after the messages are ready, instead of the `AutomaticallyCreatedBackgroundThreads`. I decided against this because I'd rather have no reliance on timers at all.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7146)

### DIFF
--- a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
+++ b/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
@@ -85,11 +85,7 @@ TEST(BatchingPublisherConnectionTest, DefaultMakesProgress) {
   google::cloud::internal::AutomaticallyCreatedBackgroundThreads background;
   auto const ordering_key = std::string{};
   auto publisher = BatchingPublisherConnection::Create(
-      topic,
-      pubsub::PublisherOptions{}
-          .set_maximum_batch_message_count(4)
-          .set_maximum_hold_time(std::chrono::milliseconds(50)),
-      ordering_key, mock, background.cq());
+      topic, pubsub::PublisherOptions{}, ordering_key, mock, background.cq());
 
   // We expect the responses to be satisfied in the context of the completion
   // queue threads. This is an important property, the processing of any
--- a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
+++ b/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
@@ -64,7 +64,7 @@ TEST(BatchingPublisherConnectionTest, DefaultMakesProgress) {
   pubsub::Topic const topic("test-project", "test-topic");
 
   AsyncSequencer<void> async;
-  auto publish_count = 0;
+  std::atomic<int> publish_count{0};
   EXPECT_CALL(*mock, AsyncPublish)
       .Times(AtLeast(1))
       .WillRepeatedly([&](google::pubsub::v1::PublishRequest const& request) {
@@ -113,7 +113,7 @@ TEST(BatchingPublisherConnectionTest, DefaultMakesProgress) {
   publisher->Flush({});
   // Use the CQ threads to satisfy the AsyncPull future, like we do in the
   // normal code.
-  background.cq().RunAsync([&async, publish_count] {
+  background.cq().RunAsync([&async, &publish_count] {
     for (auto i = 0; i != publish_count; ++i) {
       async.PopFront().set_value();
     }
--- a/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
+++ b/google/cloud/pubsub/internal/batching_publisher_connection_test.cc
@@ -64,9 +64,11 @@ TEST(BatchingPublisherConnectionTest, DefaultMakesProgress) {
   pubsub::Topic const topic("test-project", "test-topic");
 
   AsyncSequencer<void> async;
+  auto publish_count = 0;
   EXPECT_CALL(*mock, AsyncPublish)
       .Times(AtLeast(1))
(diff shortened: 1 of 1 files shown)


google/cloud/pubsub/internal/batching_publisher_connection_test.cc
```diff
@@ -64,9 +64,11 @@ TEST(BatchingPublisherConnectionTest, DefaultMakesProgress) {
   pubsub::Topic const topic("test-project", "test-topic");
 
   AsyncSequencer<void> async;
+  auto publish_count = 0;
```
coryan, line 67: This is thread safe for fairly obscure reasons (see below), maybe a comment?

google/cloud/pubsub/internal/batching_publisher_connection_test.cc
```diff
@@ -64,9 +64,11 @@ TEST(BatchingPublisherConnectionTest, DefaultMakesProgress) {
...
   EXPECT_CALL(*mock, AsyncPublish)
       .Times(AtLeast(1))
       .WillRepeatedly([&](google::pubsub::v1::PublishRequest const& request) {
+        ++publish_count;
```
coryan, line 71: This happens in "the" thread created by `background` below.

google/cloud/pubsub/internal/batching_publisher_connection_test.cc
```diff
@@ -83,11 +85,7 @@ TEST(BatchingPublisherConnectionTest, DefaultMakesProgress) {
   google::cloud::internal::AutomaticallyCreatedBackgroundThreads background;
```
coryan, line 85: This creates a single thread by default.

google/cloud/pubsub/internal/batching_publisher_connection_test.cc
```diff
@@ -115,7 +113,11 @@ TEST(BatchingPublisherConnectionTest, DefaultMakesProgress) {
   publisher->Flush({});
   // Use the CQ threads to satisfy the AsyncPull future, like we do in the
   // normal code.
-  background.cq().RunAsync([&async] { async.PopFront().set_value(); });
+  background.cq().RunAsync([&async, publish_count] {
+    for (auto i = 0; i != publish_count; ++i) {
```
coryan, line 117: This happens in the same thread as the `AsyncPublish()`, because `background` creates a single thread, otherwise you would need some synchronization.

coryan (commented):
FYI..

dbolduc:
Ack. I opted for synchronization over a comment

google/cloud/pubsub/internal/batching_publisher_connection_test.cc
```diff
@@ -64,9 +64,11 @@ TEST(BatchingPublisherConnectionTest, DefaultMakesProgress) {
   pubsub::Topic const topic("test-project", "test-topic");
 
   AsyncSequencer<void> async;
+  std::atomic<int> publish_count{0};
```
coryan, line 67: Okay, pretty soon you will realize that I am dumb, because I always barely understand these synchronization primitives, and each time I learn enough to understand something, I forget in about 5 seconds flat.

google/cloud/pubsub/internal/batching_publisher_connection_test.cc
coryan: If we are assuming there are multiple threads involved, I think atomics are not guaranteed to save the day.  The change may not be visible when this thread runs, so you may get "0" when loading `publish_count`, even if it was incremented twice and "async" contains two elements:

https://stackoverflow.com/questions/42003798/how-do-i-make-memory-stores-in-one-thread-promptly-visible-in-other-threads

And therefore "Bad Things[tm]" happen (probably a deadlock).  

Well, it is complicated, again.  It probably works, but for other obscure reasons: `RunAsync()` only runs once `Flush()` completes and that only after the two `PublishAsync()` complete.  `RunAsync()` uses mutexes to push the the lambda to the CQ thread.  Those mutexes will preserve the "happens before" relationships and make the changes to `publish_count` (and anything else) visible to that other thread...  But the atomic itself was not enough.

If we are going to assume synchronization is needed, then maybe just add a new `.PopFrontSafe() -> optional<promise<T>>` or `.Empty()` to the `AsyncSequencer` and use it instead of a separate counter.

Thoughts?

coryan (commented):
Sorry for the synchronization nitpicking, not sure why it matters to me today.

dbolduc:
I will return to this shortly, but I don't like how it is sitting in the PR queue

PR #7150, opened 2021-08-10
fix(statusor)!: remove status() accessor returning non-const lvalue 

devjgm:
This function would allow callers to assign to the contained `Status`
and potentially violate class invariants. For example:

```cc
StatusOr<Foo> sor;  // constructed with kUnknown and no Foo
sor.status() = Status{};  // BUG
auto v = sor.value();  // :-|
```

This PR fixes the above API bug by removing the `status()` accessor that
returns the lvalue to non-const T.

I also remove the `status()` overload that returns the rvalue-ref to
const-T, because I don't think it adds any new functionality to the
class, and a "move" isn't actually happening behind the scenes. Also,
the two access that remain (`const &` and `&&`) are the same two that
`absl::StatusOr` [1] has, so we match there.

[1]: https://github.com/abseil/abseil-cpp/blob/bf31a10b65d945665cecfb9d8807702ae4a7fde1/absl/status/statusor.h#L492-L493

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7150)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,16 @@
 
 ## v1.31.0 - TBD
 
+## [Common Libraries](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/README.md)
+
+**BREAKING CHANGES**:
+
+* `google::cloud::StatusOr<T>` had an accessor that returned an lvalue
+  reference to non-const `Status`, which could allow callers to modify the
+  contained `Status` and thus break invariants of the `StatusOr` class. This
+  function was removed. If your code previously relied on `sor.status() =
+  new_status` you should change it to `sor = new_status`.
+
 ## v1.30.0 - 2021-08
 
 ### New GA Libraries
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -54,6 +54,16 @@
 
 ## v1.31.0 - TBD
 
+## [Common Libraries](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/README.md)
+
+**BREAKING CHANGES**:
+
+* `google::cloud::StatusOr<T>` had an accessor that returned an lvalue
+  reference to non-const `Status`, which could allow callers to modify the
```
coryan, line 62: s/, which could allow/. This allowed/ ?

coryan (approved):

CHANGELOG.md
devjgm: Done.

PR #7169, opened 2021-08-13
doc: require an explicit flag to enable Doxygen

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7169)


devjgm (approved):
Might this speed up some cmake builds?

coryan:
I expect a few seconds / milliseconds faster as it does not need to search for the Doxygen program.

PR #7177, opened 2021-08-14
ci: add Debian:11 (Bullseye) build files

coryan:
The release is imminent, it is easy to set up a build for it. In about a
year we should retire the Debian:9 (Stretch) builds as this is under LTS
support and will be EOL in June.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7177)


devjgm:
May want to update https://github.com/googleapis/google-cloud-cpp/blob/main/ci/generate-markdown/generate-packaging.sh too

coryan:
> May want to update https://github.com/googleapis/google-cloud-cpp/blob/main/ci/generate-markdown/generate-packaging.sh too

I was of two minds on this, but that nudge is enough. Done.  I also fixed some stale and ambiguous release names.

devjgm (approved):

PR #7184, opened 2021-08-17
feat(spanner): add generated admin APIs

devbww:
Add the generated Spanner database/instance admin APIs.  These are
just the raw, generated sources.  The hand edits needed to build them
into the Spanner client library will be added later.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7184)


coryan (approved):
FWIW, I just scanned the generated code.

PR #7195, opened 2021-08-19
test(spanner): extend timeouts for "slow" database operations

### LINKED ISSUE #4306: Spanner backup integration test timeout

devbww:
Creating/restoring backups are not only long-running operations,
but their service time is variable (apparently because they are
batch-scheduled in the backend).

So:
- Extend the timeout for "eternal" integration tests in the
  `integration-daily` build (where they run against prod) from
  1h (the default) to 2h.  This now matches the GCB timeout.
- Extend the per-operation polling timeout from 30m to 60m.

Fixes #4306.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7195)


coryan (approved):
:crossed_fingers:

PR #7199, opened 2021-08-20
refactor: move gRPC Channel Argument helpers to common

dbolduc:
If a library wants to default a gRPC channel argument, it will need to check if the user has set it. Move the functions that do this into common.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7199)

### DIFF
--- a/google/cloud/bigtable/client_options_test.cc
+++ b/google/cloud/bigtable/client_options_test.cc
@@ -35,6 +35,12 @@ struct ClientOptionsTestTraits {
   }
 };
 
+namespace {
+using ::google::cloud::internal::GetIntChannelArgument;
+using ::google::cloud::internal::GetStringChannelArgument;
+using ::testing::HasSubstr;
+}  // namespace
+
 TEST(ClientOptionsTest, ClientOptionsDefaultSettings) {
   bigtable::ClientOptions client_options_object = bigtable::ClientOptions();
   EXPECT_EQ("bigtable.googleapis.com", client_options_object.data_endpoint());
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/client_options_test.cc
```diff
@@ -37,32 +37,12 @@ struct ClientOptionsTestTraits {
 namespace {
 
 using ::testing::HasSubstr;
+// NOLINTNEXTLINE(readability-identifier-naming)
+auto const GetInt = ::google::cloud::internal::GetIntChannelArgument;
```
devjgm, line 41: Rather than declaring these here, I think it would be clearer if this file just included `google/cloud/grpc_options.h` and then directly called the new functions with their new names.

dbolduc: Done.

devjgm (approved):

google/cloud/bigtable/client_options_test.cc
```diff
@@ -307,8 +278,9 @@ TEST(ClientOptionsTest, SetGrpclbFallbackTimeoutMS) {
   bigtable::ClientOptions client_options_object = bigtable::ClientOptions();
   ASSERT_STATUS_OK(client_options_object.SetGrpclbFallbackTimeout(
       std::chrono::milliseconds(5)));
-  auto const actual = GetInt(client_options_object.channel_arguments(),
-                             GRPC_ARG_GRPCLB_FALLBACK_TIMEOUT_MS);
+  auto const actual = ::google::cloud::internal::GetIntChannelArgument(
```
devbww, line 281: Nit: Do we want to fully-qualify names like this?

google/cloud/bigtable/client_options_test.cc
```diff
@@ -380,34 +356,36 @@ TEST(ClientOptionsTest, SetServiceConfigJSON) {
...
   bigtable::ClientOptions client_options_object = bigtable::ClientOptions();
   client_options_object.SetUserAgentPrefix("test_prefix");
-  auto const actual = GetString(client_options_object.channel_arguments(),
-                                GRPC_ARG_PRIMARY_USER_AGENT_STRING);
+  auto const actual = ::google::cloud::internal::GetStringChannelArgument(
+      client_options_object.channel_arguments(),
+      GRPC_ARG_PRIMARY_USER_AGENT_STRING);
   ASSERT_TRUE(actual.has_value());
-  EXPECT_THAT(*actual, HasSubstr("test_prefix"));
+  EXPECT_THAT(*actual, ::testing::HasSubstr("test_prefix"));
```
devbww, line 372: Nit: The `using ::testing::HasSubstr` was probably a feature.

google/cloud/bigtable/client_options_test.cc
dbolduc: I guess not.

google/cloud/bigtable/client_options_test.cc
dbolduc: Reverted.

devbww (approved):

PR #7200, opened 2021-08-20
feat(generator): customization hooks for endpoint configuration

devbww:
Add two ServiceConfiguration fields to customization the generation
of `<Service>DefaultOptions()`.

- `service_endpoint_env_var`: Overrides the default environment
  variable name, `GOOGLE_CLOUD_CPP_<SERVICE>_ENDPOINT`, used to
  override the default value of the `EndpointOption`.

- `emulator_endpoint_env_var`: Enables an extra override of the
  `EndpointOption` from an additional environment variable. When
  used, it also sets the `GrpcCredentialOption` with insecure
  channel credentials.

Use both of these new configuration fields for the generation of
the Spanner database/instance-admin services so that they match
the use of the existing `GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT`
and `SPANNER_EMULATOR_HOST` environment variables respectively.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7200)

### DIFF
--- a/generator/generator.h
+++ b/generator/generator.h
@@ -40,6 +40,8 @@ namespace generator {
  *   --cpp_codegen_out=. \
  *   --cpp_codegen_opt=product_path=google/cloud/spanner
  *   --cpp_codegen_opt=copyright_year=2020
+ *   --cpp_codegen_opt=service_endpoint_env_var=SERVICE_ENDPOINT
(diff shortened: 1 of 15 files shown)


coryan (approved):
LGTM

PR #7203, opened 2021-08-20
cleanup(generator): use separator lines consistently

devbww:
In a recent change to the generator I noticed that we sometimes add
multiple separator lines when only one would do, and don't add one
when one is appropriate.  This remedies that.

While spacing isn't that important, we do submit the generated code
to the repo, so it is nice to be as consistent as possible.  A more
clear-cut generation strategy (like always using leading separators,
except for the initial comment block) might help keep things that
way, but I've just done the minimal tweaks here.  (I tweaked up some
spacing in the generator proper, too.)

Also correct some indentations, although this is even less important
as it is typically fixed during a post `clang-format` phase.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7203)


scotthart (approved):
> Reviewed 156 of 156 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7203)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devbww)

generator/internal/stub_generator.cc
```diff
@@ -159,13 +160,14 @@ Status StubGenerator::GenerateHeader() {
...
                        // clang-format on
                        And(IsNonStreaming, Not(IsLongrunningOperation))),
-         MethodPattern({{R"""(
-  future<StatusOr<google::longrunning::Operation>> Async$method_name$(
+         MethodPattern(
+             {{R"""(  future<StatusOr<google::longrunning::Operation>> Async$method_name$(
       google::cloud::CompletionQueue& cq,
       std::unique_ptr<grpc::ClientContext> context,
       $request_type$ const& request) override;
+
```
coryan, line 168: I was hoping we would settle on "code fragments never end with a blank line" so we could inject the blank line at the beginning of the code fragment (if needed), but as long as this is consistent I am okay.

coryan (approved):

generator/internal/stub_generator.cc
devbww: Yeah, I think that is what I was trying to get at with the "always using leading separators, except for the initial comment block" in the PR description.

PR #7204, opened 2021-08-20
cleanup: define inline ns to be GOOGLE_CLOUD_CPP_NS

devjgm:
This is a cleanup related to #5976

There's no semantic change here.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7204)

### DIFF
--- a/google/cloud/pubsub/version.h
+++ b/google/cloud/pubsub/version.h
@@ -18,32 +18,6 @@
 #include "google/cloud/pubsub/version_info.h"
 #include "google/cloud/version.h"
 
-#define GOOGLE_CLOUD_CPP_PUBSUB_NS                                     \
-  GOOGLE_CLOUD_CPP_VEVAL(GOOGLE_CLOUD_CPP_PUBSUB_CLIENT_VERSION_MAJOR, \
-                         GOOGLE_CLOUD_CPP_PUBSUB_CLIENT_VERSION_MINOR)
-
-namespace google {
-/**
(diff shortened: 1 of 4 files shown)


google/cloud/pubsub/version.h
```diff
@@ -18,32 +18,6 @@
 #include "google/cloud/pubsub/version_info.h"
 #include "google/cloud/version.h"
 
-#define GOOGLE_CLOUD_CPP_PUBSUB_NS                                     \
-  GOOGLE_CLOUD_CPP_VEVAL(GOOGLE_CLOUD_CPP_PUBSUB_CLIENT_VERSION_MAJOR, \
-                         GOOGLE_CLOUD_CPP_PUBSUB_CLIENT_VERSION_MINOR)
-
-namespace google {
-/**
```
coryan, line 26: I can see why you would remove these, but they show up in the documentation for Pub/Sub:

https://googleapis.dev/cpp/google-cloud-pubsub/latest/namespaces.html

If you remove them from here, the documentation will be gone.

coryan (commented):
Where does Doxygen get the namespace documentation from now?

google/cloud/pubsub/version.h
devjgm: Oh, I didn't know that was used for anything. Added them back.

coryan: It is possible they are redundant, but let's separate that cleanup if that is the case.

coryan (approved):

Issue #6307, opened 2021-04-21
Convert google/cloud/bigtable to use g::c::Options


coryan:
Change the library to *prefer* `google::cloud::Options` in its initialization, the overloads taking `google::cloud::bigtable::ClientOptions` should remain usable for backwards compatibility.

PR #7206, opened 2021-08-21
refactor(bigtable): replace grpc::ChannelArguments with Options

dbolduc:
Part of the work for #6307 

Note that:
1. `std::string connection_pool_name_` cannot be removed because `ClientOptions::connection_pool_name()` returns a `std::string const&`. (Well, maybe it can, but I don't know how to do it)
2. The functionality of setting a connection pool name is captured by both `GrpcChannelArgumentsOption` and `GrpcChannelArgumentsNativeOption`. (just add a string to the map, or use `SetString()` on the native object)

We could add language that tells users to avoid the `ClientOptions` setters, in favor of doing all configuring through `Options`. But I sort of think we should just wait until every use of `ClientOptions` has an `Options` alternative, and then we add language that tells users to avoid the `ClientOptions` class altogether. A second opinion on this strategy would be appreciated.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7206)

### DIFF
--- a/google/cloud/bigtable/client_options.h
+++ b/google/cloud/bigtable/client_options.h
@@ -177,13 +177,14 @@ class ClientOptions {
 
   /// Access all the channel arguments.
   grpc::ChannelArguments channel_arguments() const {
-    return channel_arguments_;
+    return ::google::cloud::internal::MakeChannelArguments(opts_);
   }
 
   /// Set all the channel arguments.
   ClientOptions& set_channel_arguments(
       grpc::ChannelArguments const& channel_arguments) {
-    channel_arguments_ = channel_arguments;
+    opts_.set<GrpcChannelArgumentsNativeOption>(channel_arguments);
+    opts_.unset<GrpcChannelArgumentsOption>();
     return *this;
   }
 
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/client_options.h
```diff
@@ -177,13 +177,13 @@ class ClientOptions {
...
   grpc::ChannelArguments channel_arguments() const {
-    return channel_arguments_;
+    return ::google::cloud::internal::MakeChannelArguments(opts_);
   }
 
   /// Set all the channel arguments.
   ClientOptions& set_channel_arguments(
       grpc::ChannelArguments const& channel_arguments) {
-    channel_arguments_ = channel_arguments;
+    opts_.set<GrpcChannelArgumentsNativeOption>(channel_arguments);
```
dbolduc, line 186: I considered unsetting `GrpcChannelArgumentsOptions` here.
```
opts_.unset<GrpcChannelArgumentsOption>();
```

The `grpc::ChannelArguments` object that gets made includes both options (and `UserAgentProductsOption`). So we could have a situation like this:
```
ClientOptions client_options(Options{}.set<GrpcChannelArgumentsOption>({{"key", "value"}});
client_options.set_channel_arguments(channel_args);
auto returned = client_options.channel_arguments();
// returned != channel_args
// because `returned` contains the extra ("key", "value") pair.
```

But I know we don't want to unset `UserAgentProductsOption`. And it's a bit weird to unset `GrpcChannelArgumentsOption` and not `UserAgentProductsOption`.

I suspect this is an inconsequential difference, as we will recommend against these methods..... but if someone else thinks differently, I would be happy to unset the map option.

coryan: Ack

coryan (approved):

Issue #7179, opened 2021-08-16
Restore instructions for generating compile_commands.json


remyabel:
Sometimes users will want to generate this while developing on the project locally, see https://github.com/googleapis/google-cloud-cpp/issues/3381. The instructions that were there previously were removed. The old instructions were bound to be outdated anyway as the project evolves and the "conventional" way of generating the file is cumbersome. Greg's suggestion to use Bazel ended up being straight forward and seems future-proof, so I propose adding it to the contributing docs.

PR #7210, opened 2021-08-23
docs: add instructions on generating compile_commands.json

remyabel:
Fixes: #7179

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7210)

### DIFF
--- a/doc/contributor/howto-guide-setup-development-workstation.md
+++ b/doc/contributor/howto-guide-setup-development-workstation.md
@@ -148,10 +148,7 @@ Name this script `bazel-compile-commands.sh`, add it to `~/.local/bin` and
 ```console
 #!/bin/bash
 readonly TMP_DIR="$(mktemp -d "/tmp/bazel-compdb-temp-XXXXXXX")"
-cleanup() {
-  rm -rf "${TMP_DIR}"
-}
-trap cleanup EXIT
+trap "rm -rf ${TMP_DIR}" EXIT
 readonly REPO="https://github.com/grailbio/bazel-compilation-database"
 git clone --depth=1 "${REPO}" "${TMP_DIR}"
 CC=clang CXX=clang++ "${TMP_DIR}/generate.sh"
--- a/doc/contributor/howto-guide-setup-development-workstation.md
+++ b/doc/contributor/howto-guide-setup-development-workstation.md
@@ -123,8 +123,12 @@ alternatives.
 ./ci/install-cloud-sdk.sh
 ```
 
-If you use an editor that uses `compile_commands.json` for autocomplete, there
-is a straight-forward way to generate it using Bazel. First, install Bazelisk:
+If you use an editor that uses `compile_commands.json` for autocomplete, one
+can generate it with CMake using
+[CMAKE_EXPORT_COMPILE_COMMANDS](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html).
+If you prefer using Bazel, the following steps will generate one for you.
+
+First, install Bazelisk:
 
 ```console
 go get github.com/bazelbuild/bazelisk
--- a/doc/contributor/howto-guide-setup-development-workstation.md
+++ b/doc/contributor/howto-guide-setup-development-workstation.md
@@ -123,6 +123,42 @@ alternatives.
 ./ci/install-cloud-sdk.sh
 ```
 
+If you use an editor that uses `compile_commands.json` for autocomplete, there
+is a straight-forward way to generate it using Bazel. First, install Bazelisk:
+
+```console
+go get github.com/bazelbuild/bazelisk
+# Alternatively, if you have Go 1.17 or later
+go install github.com/bazelbuild/bazelisk@latest
+```
+
+Add Go binaries to your PATH:
+
+```console
+export PATH=$PATH:$(go env GOPATH)/bin
+```
+
+Name this script `bazel-compile-commands.sh`, add it to `~/.local/bin` and
+`chmod +x` it:
+
+```console
+#!/bin/bash
+readonly TMP_DIR="$(mktemp -d "/tmp/bazel-compdb-temp-XXXXXXX")"
+cleanup() {
+  rm -rf "${TMP_DIR}"
+}
+trap cleanup EXIT
+readonly REPO="https://github.com/grailbio/bazel-compilation-database"
+git clone --depth=1 "${REPO}" "${TMP_DIR}"
+CC=clang CXX=clang++ "${TMP_DIR}/generate.sh"
+```
+
+Then from the directory where you've cloned `google-cloud-cpp`:
+
+```console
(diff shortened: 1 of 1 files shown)


doc/contributor/howto-guide-setup-development-workstation.md
```diff
@@ -123,6 +123,42 @@ alternatives.
 ./ci/install-cloud-sdk.sh
 ```
 
+If you use an editor that uses `compile_commands.json` for autocomplete, there
+is a straight-forward way to generate it using Bazel. First, install Bazelisk:
```
coryan, line 127: I think "straightforward" does not need a hyphen?

doc/contributor/howto-guide-setup-development-workstation.md
```diff
@@ -123,6 +123,42 @@ alternatives.
...
+
+```console
+export PATH=$PATH:$(go env GOPATH)/bin
+```
+
+Name this script `bazel-compile-commands.sh`, add it to `~/.local/bin` and
+`chmod +x` it:
+
+```console
+#!/bin/bash
```
coryan, line 145: This seems more complicated than this:

https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html

Is it because installing the deps is easier with Bazel vs. CMake?  Maybe we should say something like:

If you are using CMake, read about CMAKE_EXPORT_COMPILE_COMMANDS (<<<< link) to generate a compilation database.  If you prefer using Bazel, the following steps will generate one for you.

WDYT?

coryan (changes requested):

doc/contributor/howto-guide-setup-development-workstation.md
remyabel: Could be wrong, but there currently isn't any way to pass extra Cmake flags to the docker build? Alternatively, one could have a separate CMake build just for `compile_commands.json` but that got unwieldly. So I looked into using ninja directly, but also ran into issues with that. Even though the Bazel approach looks like it requires more steps on the surface I found it to actually be the most straightforward overall. But this is based on fuzzy memory and trying to figure out the cleanest way of doing this for months. I will add the link either way.

remyabel: I remember now. There were weird issues with file paths when trying to generate the file within Docker. For some reason, the Bazel approach does not have that issue.

coryan: Ah, yes, there would be issues with the paths, as inside Docker we mount the source and $HOME in weird directories.  I think this looks fine.  Let me summon @devjgm because he uses Bazel for development and may want to chime in.

doc/contributor/howto-guide-setup-development-workstation.md
```diff
@@ -123,6 +123,46 @@ alternatives.
...
+readonly TMP_DIR="$(mktemp -d "/tmp/bazel-compdb-temp-XXXXXXX")"
+cleanup() {
+  rm -rf "${TMP_DIR}"
+}
+trap cleanup EXIT
```
devjgm, line 154: I think just `trap "rm -rf ${TMP_DIR}" EXIT` should work.

doc/contributor/howto-guide-setup-development-workstation.md
```diff
@@ -123,6 +123,46 @@ alternatives.
...
+readonly REPO="https://github.com/grailbio/bazel-compilation-database"
+git clone --depth=1 "${REPO}" "${TMP_DIR}"
+CC=clang CXX=clang++ "${TMP_DIR}/generate.sh"
+```
+
+Then from the directory where you've cloned `google-cloud-cpp`:
+
+```console
+bazel-compile-commands.sh --sandbox_writable_path=$HOME/.ccache
```
devjgm, line 163: I don't use this `--sandbox_writable_path` arg when I do this. What's it for?

doc/contributor/howto-guide-setup-development-workstation.md
```diff
@@ -123,6 +123,46 @@ alternatives.
...
+```console
+go get github.com/bazelbuild/bazelisk
+# Alternatively, if you have Go 1.17 or later
+go install github.com/bazelbuild/bazelisk@latest
+```
+
+Add Go binaries to your PATH:
+
+```console
+export PATH=$PATH:$(go env GOPATH)/bin
```
devjgm, line 142: Thanks for the PR @remyabel. I like this. I'm just wondering if we want to just commit this script to our repo so it's even easier for folks to run. The instructions could just be: "To generate a `compile_commands.json` file run `ci/compile-commands.sh`. Re-run that command whenever files are added or deleted from the repo." The body of the script would be pretty much what you have now, except it could also install bazelisk iff needed.

Note: we could also include the link to https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html as well, though I believe using bazel to generate the compile commands file would be fine even if you use cmake to build the code, but maybe someone (@coryan ?) could confirm that.

doc/contributor/howto-guide-setup-development-workstation.md
remyabel: On my system at least, Bazel will complain about not having write permissions to the `.ccache` directory. This allows it to be accessed within the sandbox. Maybe it won't be necessary on everyone's system, but it's a common enough issue based on a google search.

doc/contributor/howto-guide-setup-development-workstation.md
remyabel: I guess. It would add a maintenance burden and the script is more of an example than authoritative steps. I copy/pasted the script you wrote (see the original issue). Would have to evaluate what the easiest option for everyone is. This happens to be the easiest option for me now. Overtime as the project developed, the instructions for generating `compile_commands.json` were removed, then the ability to add extra cmake commands to docker were removed (and not added back yet, I think I saw an issue on that) and as mentioned in another comment having a separate build just for it is kind of unwieldly as it requires compiling the entire project and that takes forever. I know ninja has something like `compdb`, but that requires fiddling. And when running it inside Docker, the paths are messed up. The current approach seems quickest and fastest so far and least likely to become obsolete as the code base changes overtime IMO.

devjgm (approved):
OK, whether we commit this as as cript or not doesn't need to block this PR. This is clearly an improvement, so LGTM. Thanks.

doc/contributor/howto-guide-setup-development-workstation.md
devjgm: LGTM

devjgm (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7210)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @coryan and @remyabel)

coryan (approved):

PR #7212, opened 2021-08-24
feat(spanner): support for JSON as a column type

devbww:
`google::cloud::spanner::JSON` is just a strongly-typed string.
That is, a `JSON` value can be constructed from, and converted
to a `std::string`. `JSON` values can also be compared (by string)
for equality, and streamed.

There is no syntax checking of JSON strings in the client library.
All checking is done on the server. The user is expected to build
`JSON` values from well-formatted strings, probably produced by
a separate JSON library. Similarly, `JSON` values should be parsed
by that separate library.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7212)

### DIFF
--- a/google/cloud/spanner/integration_tests/data_types_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/data_types_integration_test.cc
@@ -30,6 +30,9 @@ namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
 namespace {
 
+using ::google::cloud::testing_util::IsOk;
+using ::google::cloud::testing_util::StatusIs;
+using ::testing::HasSubstr;
 using ::testing::UnorderedElementsAreArray;
 
 absl::Time MakeTime(std::time_t sec, int nanos) {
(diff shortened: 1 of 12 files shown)


google/cloud/spanner/integration_tests/database_admin_integration_test.cc
```diff
@@ -207,6 +208,12 @@ TEST_F(DatabaseAdminClientTest, DatabaseBasicCRUD) {
           FirstName  STRING(1024),
           LastName   STRING(1024),
           SingerInfo BYTES(MAX)
+      )""");
+  if (!emulator_) {
```
coryan, line 212: Should we just create a separate test for JSON?  All these `if (!emulator_)` things sprinkled over the test are hard to grok.

google/cloud/spanner/json.h
```diff
@@ -0,0 +1,89 @@
...
+ * JavaScript Object Notation) defines a small set of formatting rules
+ * for the portable representation of structured data. See RFC 7159.
+ *
+ * A `JSON` value can be constructed from, and converted to a `std::string`.
+ * `JSON` values can be compared (by string) for equality, and streamed.
+ *
+ * There is no syntax checking of JSON strings in this interface. The user
+ * is expected to only construct `JSON` values from well-formatted strings.
+ */
+class JSON {
```
coryan, line 39: I think this should be `Json`:


> prefer to capitalize abbreviations as single words, e.g., StartRpc() rather than StartRPC().

https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

coryan (commented):
If my comments do not make sense just let me know.

coryan (approved):

google/cloud/spanner/integration_tests/database_admin_integration_test.cc
devbww: At least for the time being, I think I prefer this over duplication.  But part of the "grokness" issue depends on how long these conditionals last.  I'm still hoping that the emulator will support some of these features soon, but who knows.

google/cloud/spanner/json.h
devbww: Done, not that I particularly like that rule, or at least when applied to a single word.

google/cloud/spanner/integration_tests/data_types_integration_test.cc
```diff
@@ -408,6 +445,39 @@ TEST_F(DataTypeIntegrationTest, InsertAndQueryWithStruct) {
   EXPECT_EQ(data, v[0]);
 }
 
+// Verify maximum JSON nesting.
+TEST_F(DataTypeIntegrationTest, JsonMaxNesting) {
+  // TODO(#6873): Remove this check when the emulator supports JSON.
+  if (UsingEmulator()) GTEST_SKIP();
+
+  // The default value of the backend max-nesting-level flag.
+  int const k_spanner_json_max_nesting_level = 100;
```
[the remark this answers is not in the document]
devbww: I'll leave it.  Compile-time-constness isn't particular germane to the case.

PR #7237, opened 2021-08-27
fix(bigtable): ensure connection refresh period is valid

### LINKED ISSUE #7235: Bigtable connection refresh options can lead to UB
`ClientOptions` internally enforced a rule that the min connection refresh period <= max connection refresh period. The implementation with `Options` does not do that.

We can:
1. quickly change the `Make*Client` methods to return `StatusOr<*Client>`, which would fail if a min is provided that is greater than a max.
2. edit the defaulting method to enforce the rule

At the moment, I think doing 2 makes the most sense. The only unclear case for what to do is when the user supplies a min that is greater than a max. I think we should just flip the bounds for this case. I am open to other opinions though.

dbolduc:
Fixes #7235 

I went with @coryan's suggestion to just use the maximum instead of flipping the bounds when an invalid range is supplied. It led to simpler code and probably simpler documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7237)

### DIFF
--- a/google/cloud/bigtable/internal/defaults.cc
+++ b/google/cloud/bigtable/internal/defaults.cc
@@ -33,13 +33,11 @@ namespace internal {
 namespace {
 
 // As learned from experiments, idle gRPC connections enter IDLE state after 4m.
-auto constexpr kDefaultMaxRefreshPeriod =
-    std::chrono::milliseconds(std::chrono::minutes(3));
+auto constexpr kDefaultMaxRefreshPeriod = std::chrono::minutes(3);
 
 // Applications with hundreds of clients seem to work better with a longer
 // delay for the initial refresh. As there is no particular rush, start with 1m.
-auto constexpr kDefaultMinRefreshPeriod =
-    std::chrono::milliseconds(std::chrono::minutes(1));
+auto constexpr kDefaultMinRefreshPeriod = std::chrono::minutes(1);
 
 static_assert(kDefaultMinRefreshPeriod <= kDefaultMaxRefreshPeriod,
               "The default period range must be valid");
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/internal/defaults.cc
```diff
@@ -111,11 +109,17 @@ Options DefaultOptions(Options opts) {
   if (!opts.has<GrpcNumChannelsOption>()) {
     opts.set<GrpcNumChannelsOption>(DefaultConnectionPoolSize());
   }
-  if (!opts.has<MinConnectionRefreshOption>()) {
-    opts.set<MinConnectionRefreshOption>(kDefaultMinRefreshPeriod);
-  }
-  if (!opts.has<MaxConnectionRefreshOption>()) {
-    opts.set<MaxConnectionRefreshOption>(kDefaultMaxRefreshPeriod);
+
+  bool has_min = opts.has<MinConnectionRefreshOption>();
```
coryan, line 113: nit: consider adding the calls to the `if()`, maybe.

Separately, consider `auto const has_min = ...`.

google/cloud/bigtable/internal/defaults_test.cc
```diff
@@ -37,6 +37,11 @@ using ::google::cloud::internal::GetStringChannelArgument;
 using ::google::cloud::testing_util::ScopedEnvironment;
 using ::testing::Contains;
 using ::testing::HasSubstr;
+using secs = std::chrono::seconds;
+using mins = std::chrono::minutes;
+
+/// Gives us more informative errors
+int GetMs(std::chrono::milliseconds ms) { return static_cast<int>(ms.count()); }
```
coryan, line 44: nit: you could add an overload to `PrintTo()` so googletest learns about printing `chrono` types.  Or add a `operator<<(std::ostream& os, std::chrono::duration<R, P> x) { return os << absl::FromChrono(x); }` ..

google/cloud/bigtable/options.h
```diff
@@ -82,6 +82,9 @@ struct MinConnectionRefreshOption {
  * Maximum time in ms to refresh connections.
  *
  * The server will disconnect idle connections before this time.
+ *
+ * @note If this value is less than the value of `MinConnectionRefreshOption`,
```
coryan, line 86: Don't you need a similar warnings for the other value?

coryan (approved):

google/cloud/bigtable/internal/defaults.cc
dbolduc: Fixed. The new code is 9 lines longer (🤮), but it doesn't do any unnecessary default initializations. It is more flexible too. If we ever want to change the behavior to flip the bounds of an invalid range, there is a designated `else` block for it.

google/cloud/bigtable/internal/defaults_test.cc
dbolduc: I tried both options.... and failed. 

Gonna ship the code anyway ¯\\\_(ツ)_/¯

google/cloud/bigtable/options.h
dbolduc: I don't think so. Either
1. the min option is unset -> we give the user a default value
2. the min option is set -> it is used.

the max value on the other hand may be set AND change if the min value is set higher than it.

google/cloud/bigtable/internal/defaults_test.cc
devbww: I would suggest `absl::FromChrono()` instead.

dbolduc: Fixed. (I think)

Issue #6982, opened 2021-07-13
Move GCS+gRPC plugin to v2 protos


coryan:
The `v2` protos for GCS+gRPC were just published:

https://github.com/googleapis/googleapis/tree/master/google/storage/v2

We would like to move the GCS+gRPC plugin to use these protos. I anticipate at least the following steps:

- [x] Pick a new minimum version of Clang (DONE: 6.0) and create a build for it
- [x] Remove the clang-38 build and update documentation to say Clang-6.0 is the new minimum
- [x] Update the googleapis baseline.
- [x] Update all the builds to require Protobuf >= 3.15 
- [x] Add the new protos to the storage_protos library.
- [x] Remove the functionality not available in `v2/` protos.
- [x] Remove any tests or integration tests using functionality (e.g. metadata operations) that these protos do not implement.
- [x] Change the emulator to use the new protos.
- [x] Rewrite the `StorageStub` (and friends) and the `internal::GrpcClient` to use these new protos.
- [x] Figure out what to do with the `QuotaUser` field.

coryan:
I picked Clang-6.0 as the new minimum because it is the earliest default Clang in any of the distros we support.  Ubuntu:16.04 ships with Clang-3.8, and Ubuntu:18.04 ships with Clang-6.0... So 6.0 it is.

vnvo2409:
Hi @coryan , please ping if there is anything I can help on this issue :smile:

coryan:
@vnvo2409 thanks for the offer.  Please be patient with us, I just deleted a lot of the code you wrote, but we will add it back (with some -- hopefully minor -- changes).

coryan:
Okay, I figured out the answer to quota user, we need to add this as as `x-quota-user` header in the `grpc::ClientContext`:

https://cloud.google.com/apis/docs/system-parameters

PR #7255, opened 2021-09-02
cleanup(GCS+gRPC): rename stub functions to match protos

coryan:
I thought the previous PR was too large already, and pushed this cleanup
for later. Now it is "later". All these changes are in `internal`
classes, nothing is breaking.

Part of the work for #6982

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7255)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.h
+++ b/google/cloud/storage/internal/grpc_client.h
@@ -56,11 +56,10 @@ class GrpcClient : public RawClient,
   // them is very different from the standard retry loop. Also note that these
   // are virtual functions only because we need to override them in the unit
   // tests.
-  // TODO(#6982) - rename WriteObjectStream to WriteStream.
-  using InsertStream = ::google::cloud::internal::StreamingWriteRpc<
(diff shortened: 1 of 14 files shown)


google/cloud/storage/internal/grpc_client.h
```diff
@@ -56,7 +56,7 @@ class GrpcClient : public RawClient,
   // them is very different from the standard retry loop. Also note that these
   // are virtual functions only because we need to override them in the unit
   // tests.
-  // TODO(#6982) - rename InsertStream to WriteStream.
+  // TODO(#6982) - rename WriteObjectStream to WriteStream.
```
devjgm, line 59: Is this TODO in the right place, because it appears to still be attached to `InsertStream`

devjgm (approved):

google/cloud/storage/internal/grpc_client.h
coryan: Fixed by renaming this alias too.

devjgm (approved):

PR #7160, opened 2021-08-11
chore: update main with v1.30.1 changes

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7160)

PR #7263, opened 2021-09-02
fix(storage): use CA info options in credentials

coryan:
I accidentally dropped the `options_` parameter when moving the request
construction from the constructor to when the request is issued. This
restores the use of the `options_` parameter and adds unit tests to
prevent such breaks in the future.

Fixes #7160

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7263)


devjgm (approved):

coryan:
The GCS+gRPC failure is a known bug in the service. And GCS+gRPC is not GA. I am going to ignore it.

PR #7270, opened 2021-09-03
fix(pubsub): dont move PublisherOptions that we are still using

### LINKED ISSUE #7006: [Flake] Pubsub Flow Control times out

dbolduc:
we `std::move(options)` in the `make_connection` lambda which captures by reference. Then we keep using `options`... 

I am not sure how to test for this change.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7270)

### DIFF
--- a/google/cloud/pubsub/publisher_connection.cc
+++ b/google/cloud/pubsub/publisher_connection.cc
@@ -138,13 +138,13 @@ std::shared_ptr<pubsub::PublisherConnection> MakePublisherConnection(
     if (options.message_ordering()) {
       auto factory = [topic, options, sink, cq](std::string const& key) {
         return BatchingPublisherConnection::Create(
-            topic, options, key, SequentialBatchSink::Create(std::move(sink)),
-            cq);
+            std::move(topic), std::move(options), key,
+            SequentialBatchSink::Create(std::move(sink)), std::move(cq));
       };
       return OrderingKeyPublisherConnection::Create(std::move(factory));
     }
     return RejectsWithOrderingKey::Create(BatchingPublisherConnection::Create(
-        std::move(topic), options, {}, sink, std::move(cq)));
+        topic, options, {}, std::move(sink), std::move(cq)));
   };
   auto connection = make_connection();
   if (options.full_publisher_rejects() || options.full_publisher_blocks()) {
--- a/google/cloud/pubsub/publisher_connection.cc
+++ b/google/cloud/pubsub/publisher_connection.cc
@@ -138,13 +138,12 @@ std::shared_ptr<pubsub::PublisherConnection> MakePublisherConnection(
     if (options.message_ordering()) {
       auto factory = [topic, options, sink, cq](std::string const& key) {
         return BatchingPublisherConnection::Create(
-            std::move(topic), std::move(options), key,
-            SequentialBatchSink::Create(std::move(sink)), std::move(cq));
+            topic, options, key, SequentialBatchSink::Create(sink), cq);
       };
       return OrderingKeyPublisherConnection::Create(std::move(factory));
     }
     return RejectsWithOrderingKey::Create(BatchingPublisherConnection::Create(
-        topic, options, {}, std::move(sink), std::move(cq)));
+        std::move(topic), options, {}, std::move(sink), std::move(cq)));
   };
   auto connection = make_connection();
   if (options.full_publisher_rejects() || options.full_publisher_blocks()) {
--- a/google/cloud/pubsub/publisher_connection.cc
+++ b/google/cloud/pubsub/publisher_connection.cc
@@ -144,7 +144,7 @@ std::shared_ptr<pubsub::PublisherConnection> MakePublisherConnection(
       return OrderingKeyPublisherConnection::Create(std::move(factory));
     }
     return RejectsWithOrderingKey::Create(BatchingPublisherConnection::Create(
-        std::move(topic), std::move(options), {}, sink, std::move(cq)));
+        std::move(topic), options, {}, sink, std::move(cq)));
   };
   auto connection = make_connection();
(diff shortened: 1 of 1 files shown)


coryan (dismissed):

google/cloud/pubsub/publisher_connection.cc
```diff
@@ -138,13 +138,13 @@ std::shared_ptr<pubsub::PublisherConnection> MakePublisherConnection(
     if (options.message_ordering()) {
       auto factory = [topic, options, sink, cq](std::string const& key) {
         return BatchingPublisherConnection::Create(
-            topic, options, key, SequentialBatchSink::Create(std::move(sink)),
-            cq);
+            std::move(topic), std::move(options), key,
```
coryan, line 141: I do not think you want to `std::move` here. `factory` may be called multiple times, once for each ordering key that is discovered (dynamically).  You want a copy, and note that they are captured by copy too.

google/cloud/pubsub/publisher_connection.cc
```diff
@@ -138,13 +138,13 @@ std::shared_ptr<pubsub::PublisherConnection> MakePublisherConnection(
...
+            SequentialBatchSink::Create(std::move(sink)), std::move(cq));
       };
       return OrderingKeyPublisherConnection::Create(std::move(factory));
     }
     return RejectsWithOrderingKey::Create(BatchingPublisherConnection::Create(
-        std::move(topic), std::move(options), {}, sink, std::move(cq)));
+        topic, options, {}, std::move(sink), std::move(cq)));
```
coryan, line 147: This is called at most once, and therefore a `std::move(topic)` was appropriate.

google/cloud/pubsub/publisher_connection.cc
```diff
@@ -138,13 +138,13 @@ std::shared_ptr<pubsub::PublisherConnection> MakePublisherConnection(
...
   };
   auto connection = make_connection();
```
coryan, line 149: I am beginning to think this is too clever for its own good.

Maybe we should do this (starting around line 133):


```cc
auto cp = background->cq();
auto sink = DefaultBatchSink::Create(stub, cq, std::move(retry_policy), std::move(backoff_policy));
auto connection = RejectsWithOrderingKey::Create(BatchingPublisherConnection::Create(topic, options, {}, sink, cq);
if (options.message_ordering()) {
  auto factory = [topic, options, sink, cq](std::string const& key) {
    return BatchingPublisherConnection::Create(topic, options, key, SequentialBatchSink(std::move(sink), std::move(cq)));
  };
  connection = OrderKeyPublisherConnection(std::move(factory));
}
```

It makes a few more copies than it should, but it is probably more readable.  Alternatively, moving the lambda to a function may make it more obvious.  Your call.

coryan (changes requested):

google/cloud/pubsub/publisher_connection.cc
dbolduc: I think the code is fine. I like that `BatchingPublisherConnection::Create()` is only called once

google/cloud/pubsub/publisher_connection.cc
dbolduc: whoops

google/cloud/pubsub/publisher_connection.cc
dbolduc: Fixed.

coryan (approved):

Issue #6306, opened 2021-04-21
Convert google/cloud/pubsub to use g::c::Options


coryan:
Change the library to *prefer* `google::cloud::Options` in its initialization, the overloads taking `google::cloud::pubsub::ClientOptions` should remain usable for backwards compatibility.

PR #7273, opened 2021-09-03
feat(pubsub): introduce PublisherOptionList (g::c::Options)

dbolduc:
Start of the work for #6306 

Some choices:
1. used a private enum for `FullPublisherAction`. I considered having 3 `bool` options, which might conflict, but ultimately decided that although the enum is verbose, it is clearest. It is also consistent with [this](https://github.com/googleapis/google-cloud-cpp/blob/67500487ff3d9c3acb24a5a0228f19a83753443c/google/cloud/spanner/options.h#L117-L125) in spanner. (I could use a check on doxygen comments for enum values)
2. default values are just hardcoded in (instead of using constants which have hardcoded values). This is not what was done in bigtable, but it is what was done in spanner.
3. test for equality of default values instead of testing that default values are > 0. This is not what is done in `publisher_options_test`, but I personally prefer it.
4. forward declaration of `PublisherOptions` is ugly. Not sure how to avoid it though.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7273)


google/cloud/pubsub/options.h
```diff
@@ -0,0 +1,170 @@
...
+  using Type = std::chrono::microseconds;
+};
+
+/**
+ * The maximum number of messages in a batch.
+ *
+ * @note Application developers should keep in mind that Cloud Pub/Sub
+ *    sets [limits][pubsub-quota-link] on the size of a batch (1,000 messages)
+ *    The library makes no attempt to validate the value provided in this
+ *    function.
```
coryan, line 74: s/function/option/ ?

google/cloud/pubsub/options.h
```diff
@@ -0,0 +1,170 @@
...
+  using Type = std::size_t;
+};
+
+/**
+ * The maximum size for the messages in a batch.
+ *
+ * @note Application developers should keep in mind that Cloud Pub/Sub
+ *    sets [limits][pubsub-quota-link] on the size of a batch (10MB). The
+ *    library makes no attempt to validate the value provided in this
+ *    function.
```
coryan, line 88: Ditto.

google/cloud/pubsub/options.h
```diff
@@ -0,0 +1,170 @@
...
+ * the throughput. Therefore, the behavior is disabled by default.
+ *
+ * @see the documentation for the `Publisher` class for details.
+ */
+struct MessageOrderingOption {
+  using Type = bool;
+};
+
+/// Actions taken by a full publisher.
+enum class FullPublisherAction {
```
coryan, line 145: Hmmm... I was trying to not make these enums public, I cannot recall why. Maybe because I hate `enum`:

* They are the devil to evolve
* They are a code smell for "somebody is using switch when they should be using virtual functions"

But I cannot recall why I did not want them in this case.... Can we hide them in an opaque type?

```cc
class FullPublisherAction {
 public:
   // maybe == and operator!=
  static FullPublisherAction BlockIfFull() { return FullPublisherAction(kBlocks); }
  static FullPublisherAction RejectIfFull() { return FullPublisherAction(kRejects); }
  static FullPublisherAction None() { return FullPublisherAction(kIgnored); }

 private:
   enum Discriminant { kIgnored, kRejects, kBlocks };
   explicit FullPublisherAction(Discriminant d) : discriminant_(d) {}
   friend ... internal::DoStuff(FullPublisherAction a);
};
```

google/cloud/pubsub/publisher_options.h
```diff
@@ -46,7 +60,19 @@ inline namespace GOOGLE_CLOUD_CPP_PUBSUB_NS {
  */
 class PublisherOptions {
  public:
-  PublisherOptions() = default;
+  PublisherOptions() : PublisherOptions(Options{}) {}
+  /**
```
coryan, line 64: nit: blank line before comment block

google/cloud/pubsub/publisher.cc
```diff
@@ -19,8 +19,9 @@ namespace cloud {
 namespace pubsub {
 inline namespace GOOGLE_CLOUD_CPP_PUBSUB_NS {
 
-Publisher::Publisher(std::shared_ptr<PublisherConnection> connection,
-                     PublisherOptions)
+Publisher::Publisher(
+    std::shared_ptr<PublisherConnection> connection,
+    PublisherOptions)  // NOLINT(performance-unnecessary-value-param)
```
coryan, line 24: Q: do you know why this is needed now?

google/cloud/pubsub/options.h
coryan: I am not convinced that a public enum is a bad idea, but I am not convinced it is a good idea.  Can you provide an argument one way or the other?

coryan (changes requested):

google/cloud/pubsub/options.h
dbolduc: So quick thoughts are that 

1. user code doesn't change much either way.
```c++
opts.set<FullPublisherActionOption>(FullPublisherAction::kBlockIfFull); // or kBlocks
// vs.
opts.set<FullPublisherActionOption>(FullPublisherAction::BlockIfFull());
```
 
2. Spanner has a similar enum https://github.com/googleapis/google-cloud-cpp/blob/67500487ff3d9c3acb24a5a0228f19a83753443c/google/cloud/spanner/options.h#L117-L125

3. I'm going to guess it's easier to deprecate a method from the class than it is to deprecate a value from the `enum`. Also it seems slightly better to grow a `class` by adding methods than by appending values to an `enum`.

4. Currently, the value is used as an enum. Would it ever grow? For example something that could take in a callback? Like do generic user provided action when Publisher is full, instead of one of the three actions we have listed for you.

5. the class is more complex for us to write. It is ever so slightly more complicated for a user to understand.

dbolduc: If there is an action to be taken, other than the ones we list, then the `class` makes sense. Here is a stupid one:

`static FullPublisherAction RejectIfFullWithProbability(double p);`

What might a real action be?

coryan: Let's go with the public enum.

dbolduc:
I wanted to get this done but I do not have the patience to set up a development workstation on my windows laptop right now. So I am temporarily closing this, to be reopened in one week.

google/cloud/pubsub/publisher.cc
dbolduc: A: Honestly, no... `clang-tidy` yelled at me, so I changed it. (It also yelled at me to make the change in `flow_controlled_publisher_connection_test.cc`)

PR #7276, opened 2021-09-07
feat(generator): generate a SetIamPolicy() overload with an OCC loop

### LINKED ISSUE #7168: Implement generic IAM update loop to common
A lot of services have RPCs to `GetIamPolicy()`, `SetIamPolicy()` and `TestIamPermissions()`. 

`SetIamPolicy()` is special because to use it well one needs to implement an [OCC](https://en.wikipedia.org/wiki/Optimistic_concurrency_control) loop.  We have implemented such a loop for Cloud Spanner:

https://github.com/googleapis/google-cloud-cpp/blob/7bdf353c1eaed1d1056f119d6c2d1be2d15b58b0/google/cloud/spanner/instance_admin_client.h#L241-L267

This code in spanner is almost completely generic, could be implemented over any `*Client` class that implements the `GetIamPolicy()` and `SetIamPolicy()` functions.  Its signature would be something like:

```cc
template <typename Client, typename ResourceName>
StatusOr<google::iam::v1::Policy> SetIamPolicy(
  Client& client,
  ResourceName const& name,
  IamUpdater const& updater,
  RetryPolicy retry_policy = ... some default ...
  BackoffPolicy backoff_policy = ... some other default ...) {

  while (!policy is exhausted) {
    auto current = client->GetIamPolicy(name ...);
    auto changed = updater(name);
    if (!changed.has_value()) return ...; // no changes requested
    result = client->SetIamPolicy(name, changed);
    if (result.ok()) return result;
  }
  return last_status;
}
```

This function could be part of the public API, but I anticipate that initially it would not be.

Should help with #6919 and eliminate problems like #7166

devbww:
For generated clients with standard `GetIamPolicy()` and `SetIamPolicy()`
operations, also generate a `SetIamPolicy()` overload that implements an
optimistic concurrency control loop, where the `Policy.etag` field is
used to prevent simultaneous updates of a policy from overwriting each
other.

Fixes #7168.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7276)

### DIFF
--- a/generator/internal/client_generator.cc
+++ b/generator/internal/client_generator.cc
@@ -176,27 +176,26 @@ Status ClientGenerator::GenerateHeader() {
         auto get_method_name = get_iam_policy_extension_->name();
         HeaderPrint({
             PredicatedFragment<void>(""),
-            {"  /**\n"
-             "   * Updates the IAM policy for @p resource using an"
-             " optimistic concurrency control loop.\n"
-             "   *\n"
-             "   * The loop fetches the current policy for @p resource, and"
-             " passes it to @p updater, which should return the new policy."
-             " This new policy should use the current etag so that the"
-             " read-modify-write cycle can detect races and rerun the update"
-             " when there is a mismatch. If the new policy does not have an"
-             " etag, the existing policy will be blindly overwritten. If"
-             " @p updater does not yield a policy, the control loop is"
-             " terminated and kCancelled is returned.\n"
-             "   *\n"
-             "   * @param resource  Required. The resource for which the"
-             " policy is being specified. See the operation documentation"
-             " for the appropriate value for this field.\n"
-             "   * @param updater  Required. Functor to map the current policy"
-             " to a new one.\n"
-             "   * @param options  Optional. Options to control the loop."
-             " Expected options are:\n"
-             "   *       - `$service_name$BackoffPolicyOption`\n"},
+            {R"""(  /**
+   * Updates the IAM policy for @p resource using an optimistic concurrency
+   * control loop.
+   *
+   * The loop fetches the current policy for @p resource, and passes it to @p
+   * updater, which should return the new policy. This new policy should use the
+   * current etag so that the read-modify-write cycle can detect races and rerun
+   * the update when there is a mismatch. If the new policy does not have an
+   * etag, the existing policy will be blindly overwritten. If @p updater does
+   * not yield a policy, the control loop is terminated and kCancelled is
+   * returned.
+   *
+   * @param resource  Required. The resource for which the policy is being
+   * specified. See the operation documentation for the appropriate value for
+   * this field.
+   * @param updater  Required. Functor to map the current policy to a new one.
+   * @param options  Optional. Options to control the loop. Expected options
+   * are:
+   *       - `$service_name$BackoffPolicyOption`
+)"""},
             {"   * @return " + response_type + "\n"},
             {"   */\n"},
             {"  StatusOr<" + response_type + ">\n"},
--- a/generator/internal/client_generator.cc
+++ b/generator/internal/client_generator.cc
@@ -33,7 +33,29 @@ ClientGenerator::ClientGenerator(
     google::protobuf::compiler::GeneratorContext* context)
     : ServiceCodeGenerator("client_header_path", "client_cc_path",
                            service_descriptor, std::move(service_vars),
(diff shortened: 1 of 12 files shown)


coryan (approved):
One alternative for the comment formatting, your call.

generator/internal/client_generator.cc
```diff
@@ -142,6 +168,44 @@ Status ClientGenerator::GenerateHeader() {
...
+            set_iam_policy_extension_->output_type()->full_name());
+        auto set_method_name = set_iam_policy_extension_->name();
+        auto get_method_name = get_iam_policy_extension_->name();
+        HeaderPrint({
+            PredicatedFragment<void>(""),
+            {"  /**\n"
+             "   * Updates the IAM policy for @p resource using an"
+             " optimistic concurrency control loop.\n"
+             "   *\n"
+             "   * The loop fetches the current policy for @p resource, and"
```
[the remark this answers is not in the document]
devbww: Done, although in the long run I suspect we may well be better off relying on `clang-format` to do all the formatting rather than only part of it.

scotthart (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7276)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[generator/internal/client_generator.cc, line 76 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7276#-Mj0JeWh6NZEFe9EPT1X:-Mj0JeWiAFe5-mtr7bTZ:birmspf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4d3950b937814e6f7899e9100c6f544687186e0b/generator/internal/client_generator.cc#L76)):*
> ```cpp
>                        "google/cloud/polling_policy.h",
>                        "google/cloud/status_or.h", "google/cloud/version.h"});
>   if (get_iam_policy_extension_ && set_iam_policy_extension_) {
> ```

There's no guarantee (that I'm aware of) that either or both GetIamPolicy and SetIamPolicy rpcs have method_signature options. It's just proved true so far in the handful of services we've encountered. It might be better to use the proto request/response typed rpcs directly.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7276)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @scotthart)

___
*[generator/internal/client_generator.cc, line 76 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7276#-Mj0JeWh6NZEFe9EPT1X:-Mj0Rbrn8Mm1MA6Mk7oS:baia5is) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4d3950b937814e6f7899e9100c6f544687186e0b/generator/internal/client_generator.cc#L76)):*

There's no guarantee (that I'm aware of) that either or both GetIamPolicy and SetIamPolicy rpcs have method_signature options. It's just proved true so far in the handful of services we've encountered. It might be better to use the proto request/response typed rpcs directly.
</blockquote></details>

The logic was that if they don't have the "resource" and "resource,policy" signature extensions respectively, then we don't want to generate a "resource,updater" extension either.  That is, if the service only uses the proto request/response types, then that is all you get.

scotthart (approved):
> Reviewed 10 of 12 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7276)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devbww)

google/cloud/google_cloud_cpp_common.bzl
```diff
@@ -26,6 +26,7 @@ google_cloud_cpp_common_hdrs = [
     "iam_binding.h",
     "iam_bindings.h",
     "iam_policy.h",
+    "iam_updater.h",
```
devjgm, line 29: This seems like the wrong place for these headers, since they seem to depend on gRPC and this list is for the "common" deps, which shouldn't have a dep on grpc. I suspect the other iam-related headers that already existed here may be in the wrong place too.

devbww: `iam_updater.h` moved in https://github.com/googleapis/google-cloud-cpp/pull/7287.

PR #7283, opened 2021-09-07
refactor(storage): reduce indirection on restoring uploads

coryan:
The `RawClient::RestoreResumableUpload()` function is not really used,
and it lacks a parameter with the original request (which may have set
options, such as custom headers). With this cleanup the function is
deprecated. This enables a change to fully support standard parameters
when restoring resumable uploads.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7283)

### DIFF
--- a/google/cloud/storage/internal/curl_client.cc
+++ b/google/cloud/storage/internal/curl_client.cc
@@ -231,10 +231,8 @@ CurlClient::FullyRestoreResumableSession(ResumableUploadRequest const& request,
   auto session = absl::make_unique<CurlResumableUploadSession>(
       shared_from_this(), request, session_id);
   auto response = session->ResetSession();
-  if (response.status().ok()) {
-    return std::unique_ptr<ResumableUploadSession>(std::move(session));
-  }
-  return std::move(response).status();
+  if (!response) std::move(response).status();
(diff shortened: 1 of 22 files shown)


google/cloud/storage/internal/curl_client.cc
```diff
@@ -225,6 +225,18 @@ StatusOr<ResumableUploadResponse> CurlClient::QueryResumableUpload(
   return AsStatus(*response);
 }
 
+StatusOr<std::unique_ptr<ResumableUploadSession>>
+CurlClient::FullyRestoreResumableSession(ResumableUploadRequest const& request,
+                                         std::string const& session_id) {
+  auto session = absl::make_unique<CurlResumableUploadSession>(
+      shared_from_this(), request, session_id);
+  auto response = session->ResetSession();
+  if (response.status().ok()) {
```
devbww, line 234: Nit: Do we prefer to say this over just `if (response)`?

devbww (approved):

google/cloud/storage/internal/curl_client.cc
coryan: Yes, we do. Thanks.

PR #7289, opened 2021-09-08
test(spanner): add integration tests to g/c/spanner/admin

devbww:
Add integration tests to the generated Spanner admin API tree. Note,
however, that these are currently only direct copies of the existing
tests for the hand-written APIs from g/c/spanner, and will be updated
separately to actually use the generated APIs.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7289)


coryan (approved):
BTW, only 40% of the test code runs with the emulator?  That is an indictment on the emulator feature coverage.

devbww:
> BTW, only 40% of the test code runs with the emulator? That is an indictment on the emulator feature coverage.

A combination of `!emulator` and `!daily` methinks.  But it is certainly an issue.

PR #7296, opened 2021-09-09
ci(gcb): increase the GCB timeout

devbww:
Increase the GCB timeout by 5m so that it is strictly greater than
the bazel timeout for the `integration-daily` Spanner integration
tests against prod, giving the bazel timeout the chance to fire first
and produce its test summary.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7296)

### DIFF
--- a/ci/cloudbuild/cloudbuild.yaml
+++ b/ci/cloudbuild/cloudbuild.yaml
@@ -43,7 +43,7 @@ substitutions:
   _TRIGGER_TYPE: 'manual'
   _LOGS_BUCKET: 'cloud-cpp-community-publiclogs'
 
-timeout: 7500s
+timeout: 9000s
 tags: [
   '${_TRIGGER_TYPE}',
   '${_BUILD_NAME}',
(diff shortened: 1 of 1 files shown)


coryan (approved):
5m will do it?

devjgm (approved):

devbww:
> 5m will do it?

That's a good question.  I was going to say that it doesn't take that long for the bazel timeout to be serviced, but the real issue is how long it takes for the cloud build to get to that particular bazel invocation.  So, I'll bump it to 30m and we'll see how that goes.  The cloud build timeout should be looked at as just a fail-safe anyway, so, within reason, its magnitude isn't too important.

BTW, what I'm wondering about at the moment is what effect running dual instances of the Spanner admin integration tests (hand-written plus generated APIs) will have.  Without being CPU bound, the overall test time should only be a function of the long pole, so additional tests should not change the run time too much.  But we know that slow operations, like backups, do interfere with each other in the backend, so that rule may not hold (and perhaps the interference is exacerbated should the same test instance be chosen).

PR #7301, opened 2021-09-10
cleanup: upstream patches from conan recipe

### LINKED ISSUE #7298: Upstream patch from conan-center
It makes me sad when package maintainers have to patch our code:

https://github.com/conan-io/conan-center-index/tree/master/recipes/google-cloud-cpp/all/patches

I see no reason why we cannot apply that patch in our repository: the paths will change for local builds, but we treat those as an implementation detail.

coryan:
For our builds this change has (basically) no effect. For builds
importing `google-cloud-cpp` via `git submodule` or `FetchContent` or
something similar, this isolates the downloaded googleapis files to a
subdirectory of the build.

Fixes #7298 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7301)


coryan:
Ping

PR #7303, opened 2021-09-11
feat(GCS+gRPC): reduce data copying in downloads

### LINKED ISSUE #4325: Reduce data copying in GCS+gRPC plugin
It seems like we make about 50% more data copies in the GCS+gRPC plugin vs. the REST implementation.

The code here:

https://github.com/googleapis/google-cloud-cpp/blob/76620ea9e5df9633c2fa5b4ffeb63d73761edd9e/google/cloud/storage/internal/grpc_object_read_source.cc#L59-L68

Makes more copies than it needs to, we could remove this line:

https://github.com/googleapis/google-cloud-cpp/blob/76620ea9e5df9633c2fa5b4ffeb63d73761edd9e/google/cloud/storage/internal/grpc_object_read_source.cc#L66

If we just keep `spill_` and an offset indicating what portion has been processed already.

coryan:
Fixes #4325

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7303)

### DIFF
--- a/google/cloud/storage/internal/grpc_object_read_source.cc
+++ b/google/cloud/storage/internal/grpc_object_read_source.cc
@@ -74,8 +74,8 @@ StatusOr<ReadSourceResult> GrpcObjectReadSource::Read(char* buf,
       if (response.has_checksummed_data()) {
         // Sometimes protobuf bytes are not strings, but the explicit conversion
         // always works.
-        self.spill_ = std::string(std::move(
-            *response.mutable_checksummed_data()->mutable_content()));
+        self.spill_ = std::string(
+            std::move(*response.mutable_checksummed_data()->mutable_content()));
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/grpc_object_read_source.cc
```diff
@@ -94,7 +97,7 @@ StatusOr<ReadSourceResult> GrpcObjectReadSource::Read(char* buf,
     }
   };
 
-  spill_ = update_buf(std::move(spill_));
+  spill_view_ = update_buf(std::move(spill_view_));
```
devbww, line 100: Nit: `absl::string_view` will never (I claim) have special move behavior, so it probably isn't worth indicating that it might.

devbww (approved):

PR #7305, opened 2021-09-12
feat(generator): a better default polling policy for generated LROs

devbww:
Add support for constructing a `google::cloud::GenericPollingPolicy` from
`std::shared_ptr<BasePolicy>` values (in addition to the existing support
for `ConcretePolicy` values).

This means that a polling policy can be created from dynamic retry-policy
and backoff-policy values.

Use this in generated `$service_name$DefaultOptions()` functions to build
the default `$service_name$PollingPolicyOption`, for services with long-
running operations, from the prevailing retry and backoff policies (be
they defaulted or not) instead of from some other hard-coded policies.

One slight operation difference for completely defaulted policies is that
the LRO polling policy will now start its doubling exponential backoff at
1s (matching the backoff policy for retry loops) instead of 10s.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7305)

### DIFF
--- a/generator/integration_tests/golden/internal/golden_kitchen_sink_option_defaults.cc
+++ b/generator/integration_tests/golden/internal/golden_kitchen_sink_option_defaults.cc
@@ -62,13 +62,11 @@ Options GoldenKitchenSinkDefaultOptions(Options options) {
         golden::GoldenKitchenSinkLimitedTimeRetryPolicy(
             std::chrono::minutes(30)).clone());
   }
-
   if (!options.has<golden::GoldenKitchenSinkBackoffPolicyOption>()) {
     options.set<golden::GoldenKitchenSinkBackoffPolicyOption>(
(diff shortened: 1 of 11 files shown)


coryan (approved):
Supporting `shared_ptr<T>` as parameters for the Polling Policy SGTM.

I am not sure the default *values* for a `RetryPolicy` and `BackoffPolicy` are quite right for a LRO polling policy.  But those can be changed without impacting the API, so let's merge this change and we can discuss good/better values for the policies later.

devbww:
> Supporting `shared_ptr<T>` as parameters for the Polling Policy SGTM.
> 
> I am not sure the default _values_ for a `RetryPolicy` and `BackoffPolicy` are quite right for a LRO polling policy. But those can be changed without impacting the API, so let's merge this change and we can discuss good/better values for the policies later.

Ack.

PR #7310, opened 2021-09-13
cleanup(docs): initial capitalization

dbolduc:
Capitalize initial letter for `@name`, `@note`, `@returns`, and `@warning` doxygen comments.

I didn't touch `@param [[:alpha:]]* [[:alpha:]]`, but I would be happy to. By my count there are:

| Initial Capitalization | Lines | Files |
|-|-|-|
| Lowercase | 455 | 68 |
| Uppercase | 60 | 6 |

I think using Uppercase with would be slightly better, but I don't really want to change the majority of the instances.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7310)

### DIFF
--- a/google/cloud/bigtable/table.h
+++ b/google/cloud/bigtable/table.h
@@ -504,7 +504,7 @@ class Table {
    * @param row_key the row to read.
    * @param filter a filter expression, can be used to select a subset of the
    *     column families and columns in the row.
-   * @returns A tuple, the first element is a boolean, with value `false` if the
+   * @returns a tuple, the first element is a boolean, with value `false` if the
    *     row does not exist.  If the first element is `true` the second element
    *     has the contents of the Row.  Note that the contents may be empty
    *     if the filter expression removes all column families and columns.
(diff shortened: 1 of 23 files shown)


google/cloud/grpc_options.h
```diff
@@ -96,7 +96,7 @@ struct GrpcTracingOptionsOption {
 /**
  * The size of the background thread pool
  *
- * @note this is ingored if `GrpcBackgroundThreadsFactoryOption` is set.
+ * @note This is ingored if `GrpcBackgroundThreadsFactoryOption` is set.
```
devbww, line 99: Might as well `s/ingored/ignored/` while we're here.

google/cloud/bigtable/table.h
```diff
@@ -504,7 +504,7 @@ class Table {
    * @param row_key the row to read.
    * @param filter a filter expression, can be used to select a subset of the
    *     column families and columns in the row.
-   * @returns a tuple, the first element is a boolean, with value `false` if the
+   * @returns A tuple, the first element is a boolean, with value `false` if the
```
devbww, line 507: The `@returns` ones I'm a little less clear on.  For example, this generates something that looks like ...

> **Returns**
>     a tuple, the first element is a boolean, with value false if the row does not exist.
>     If the first element is true the second element has the contents of the Row.
>     Note that the contents may be empty if the filter expression removes all column
>     families and columns.

Is the `Returns` a heading, or the first word of a sentence?  If the latter we don't want to capitalize the "a".  But, if the former, we should do more than just capitalize the "a".

So, I guess I would suggest leaving the `@returns` for now.

google/cloud/grpc_options.h
dbolduc: Done.

devbww (approved):

PR #7319, opened 2021-09-15
feat(pubsub): introduce SubscriberOptionList

dbolduc:
part of the work for #6306 

**Some choices:**

A value of `0` supplied to `MaxConcurrencyOption` uses `DefaultMaxConcurrency()`. Perhaps it would be simpler if a `0` value just becomes `1` or `4`, and an unset option uses `DefaultMaxConcurrency()`. But I know the value cannot stay `0`.

I made new Options for `MaxOutstanding(Bytes|Messages)Option` even though they seem very similar to `MaxPending(Bytes|Messages)Option`. I think it is best to separate the publisher options from the subscriber options, and to rely on the `PublisherOptionList` and `SubscriberOptionList` to check that the user has set the correct one. Maybe publisher/subscriber should be mentioned in the option name?

I am including the `@snippet`s for `MaxOutstanding(Bytes|Messages)Option` even though they point to a sample in which the new options are not used. I know that the code it points to will be updated in the future, and I would rather keep the tag, instead of removing it and potentially forgetting to add it back in later.

I have been convinced to not check for equality for default option values. I will update the `PublisherDefaults` test to do the same in a subsequent PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7319)

### DIFF
--- a/google/cloud/pubsub/options.h
+++ b/google/cloud/pubsub/options.h
@@ -154,11 +154,120 @@ struct FullPublisherActionOption {
   using Type = FullPublisherAction;
 };
 
-/// The list of options specific to publishers
+/// The list of options specific to publishers.
 using PublisherOptionList =
     OptionList<MaxHoldTimeOption, MaxBatchMessagesOption, MaxBatchBytesOption,
                MaxPendingMessagesOption, MaxPendingBytesOption,
                MessageOrderingOption, FullPublisherActionOption>;
+
+/**
+ * The maximum deadline for each incoming message.
+ *
+ * Configure how long the application has to respond (ACK or NACK) to an
+ * incoming message. Note that this might be longer, or shorter, than the
+ * deadline configured in the server-side subscription.
+ *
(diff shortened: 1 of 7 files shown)


google/cloud/pubsub/options.h
```diff
@@ -154,13 +154,122 @@ struct FullPublisherActionOption {
...
+ * The Cloud Pub/Sub C++ client library will schedule parallel callbacks as long
+ * as the number of outstanding callbacks is less than this maximum.
+ *
+ * Note that this controls the number of callbacks *scheduled*, not the number
+ * of callbacks actually executing at a time. The application needs to create
+ * (or configure) the background threads pool with enough parallelism to execute
+ * more than one callback at a time.
+ *
+ * Some applications may want to share a thread pool across many subscriptions,
+ * the additional level of control (scheduled vs. running callbacks) allows
```
devbww, line 243: s/, the/. The/

devbww (approved):

PR #7320, opened 2021-09-16
test(pubsub): update pubsub options tests

dbolduc:
Make the publisher options default test check only that a default is set (instead of checking the exact value of a default)

Also add a mini test to ensure we are outputting warnings if an option not in the expected `OptionList` is supplied to the `PublisherOptions` constructor.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7320)

### DIFF
--- a/google/cloud/pubsub/internal/defaults_test.cc
+++ b/google/cloud/pubsub/internal/defaults_test.cc
@@ -28,13 +28,11 @@ using std::chrono::seconds;
 
 TEST(OptionsTest, PublisherDefaults) {
   auto opts = DefaultPublisherOptions(Options{});
-  EXPECT_EQ(ms(10), opts.get<pubsub::MaxHoldTimeOption>());
-  EXPECT_EQ(100, opts.get<pubsub::MaxBatchMessagesOption>());
-  EXPECT_EQ(1024 * 1024L, opts.get<pubsub::MaxBatchBytesOption>());
-  EXPECT_EQ((std::numeric_limits<std::size_t>::max)(),
-            opts.get<pubsub::MaxPendingBytesOption>());
-  EXPECT_EQ((std::numeric_limits<std::size_t>::max)(),
-            opts.get<pubsub::MaxPendingMessagesOption>());
+  EXPECT_LT(0, opts.get<pubsub::MaxHoldTimeOption>().count());
+  EXPECT_LT(0L, opts.get<pubsub::MaxBatchMessagesOption>());
+  EXPECT_LT(0L, opts.get<pubsub::MaxBatchBytesOption>());
+  EXPECT_LT(0L, opts.get<pubsub::MaxPendingBytesOption>());
+  EXPECT_LT(0L, opts.get<pubsub::MaxPendingMessagesOption>());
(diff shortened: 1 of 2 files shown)


devbww (approved):

google/cloud/pubsub/internal/defaults_test.cc
```diff
@@ -27,13 +27,11 @@ using ms = std::chrono::milliseconds;
...
 TEST(OptionsTest, PublisherDefaults) {
   auto opts = DefaultPublisherOptions(Options{});
-  EXPECT_EQ(ms(10), opts.get<pubsub::MaxHoldTimeOption>());
-  EXPECT_EQ(100, opts.get<pubsub::MaxBatchMessagesOption>());
-  EXPECT_EQ(1024 * 1024L, opts.get<pubsub::MaxBatchBytesOption>());
-  EXPECT_EQ((std::numeric_limits<std::size_t>::max)(),
-            opts.get<pubsub::MaxPendingBytesOption>());
-  EXPECT_EQ((std::numeric_limits<std::size_t>::max)(),
-            opts.get<pubsub::MaxPendingMessagesOption>());
+  EXPECT_LT(0, opts.get<pubsub::MaxHoldTimeOption>().count());
```
coryan, line 30: Actually, the default values here matter, because there are common standards across all Pub/Sub libraries we want to follow.

coryan (changes requested):
Please restore the test to check for specific defaults, annoying I know.

google/cloud/pubsub/internal/defaults_test.cc
dbolduc: This defaults test copied checks from `publisher_options_test.cc`.
https://github.com/googleapis/google-cloud-cpp/blob/0eafcc84660df9070e08dbdf6807a91d6a8efa65/google/cloud/pubsub/publisher_options_test.cc#L25-L26
https://github.com/googleapis/google-cloud-cpp/blob/0eafcc84660df9070e08dbdf6807a91d6a8efa65/google/cloud/pubsub/publisher_options_test.cc#L47-L48
https://github.com/googleapis/google-cloud-cpp/blob/0eafcc84660df9070e08dbdf6807a91d6a8efa65/google/cloud/pubsub/publisher_options_test.cc#L58-L59
etc.

Should I also update the original test to check for equality of default values?

What about `subscriber_options_test.cc`?
https://github.com/googleapis/google-cloud-cpp/blob/0eafcc84660df9070e08dbdf6807a91d6a8efa65/google/cloud/pubsub/subscriber_options_test.cc#L31-L37

coryan: We only need one test that verifies the defaults match the defaults in other client libraries.  If we have one test that checks `default == specific-value` I am happy.  If all our tests amount to `default != 0` then I am not happy.

Does that make sense?

dbolduc: It does make sense. It implies that you have been unhappy with the tests for a while ;)

dbolduc:
@coryan PTAL

coryan (approved):

PR #7321, opened 2021-09-16
feat(spanner): add factory functions for instance/database/backup

devbww:
With the move to generated APIs for Spanner admin, the results of
of admin operations now only contain the raw resource strings from
the protobufs. So, it is helpful to be able to build `Instance`,
`Database`, and `Backup` objects directly from those strings.

And to complete the picture, introduce `google::cloud::Project`,
which will be helpful to encapsulate the slew of "projects/"
concatenations we've accumulated.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7321)


google/cloud/spanner/backup.h
```diff
@@ -17,7 +17,8 @@
 
 #include "google/cloud/spanner/instance.h"
 #include "google/cloud/spanner/version.h"
-#include <ostream>
+#include "google/cloud/status_or.h"
+#include <iosfwd>
```
coryan, line 21: Thanks!

coryan (approved):
Looks like this is mixing a bit of refactoring for `KmsKeyName`, an other changes.  Closely related though.

Curious why the ABI dump needed to change as part of this PR?

devbww:
> Looks like this is mixing a bit of refactoring for `KmsKeyName`, an other changes. Closely related though.

Yeah.  I couldn't help but try to make them all more alike.

> Curious why the ABI dump needed to change as part of this PR?

It looks like because I changed the existing model for all this from `MakeKmsKeyName(std::string)` to `MakeKmsKeyName(std::string const&)`.  Not really a source incompatibility.

PR #7328, opened 2021-09-17
cleanup(spanner): prefer obj.FullName() for producing resource names

devbww:
Eliminate some hand construction of resource names when we have higher
level APIs for doing just that.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7328)

### DIFF
--- a/google/cloud/kms_key_name_test.cc
+++ b/google/cloud/kms_key_name_test.cc
@@ -25,7 +25,6 @@ namespace {
 
 using ::google::cloud::testing_util::IsOk;
 using ::google::cloud::testing_util::StatusIs;
-using ::testing::Eq;
 
 TEST(KmsKeyNameTest, FromComponents) {
   KmsKeyName key("test-project", "some-location", "a-key-ring", "a-key-name");
@@ -36,11 +35,8 @@ TEST(KmsKeyNameTest, FromComponents) {
 }
 
 TEST(KmsKeyNameTest, MakeKmsKeyName) {
-  auto const k_valid_key_name_name =
-      KmsKeyName("p1", "l1", "r1", "n1").FullName();
-  auto valid_key = MakeKmsKeyName(k_valid_key_name_name);
-  ASSERT_THAT(valid_key, IsOk());
-  EXPECT_THAT(valid_key->FullName(), Eq(k_valid_key_name_name));
+  auto key = KmsKeyName("p1", "l1", "r1", "n1");
+  EXPECT_EQ(key, MakeKmsKeyName(key.FullName()).value());
 
   for (std::string invalid : {
            "projects/p1/locations/l1/keyRings/r1/carlosKey/n1",
--- a/google/cloud/kms_key_name_test.cc
+++ b/google/cloud/kms_key_name_test.cc
@@ -36,11 +36,11 @@ TEST(KmsKeyNameTest, FromComponents) {
 }
 
 TEST(KmsKeyNameTest, MakeKmsKeyName) {
-  auto constexpr kValidKeyNameName =
-      "projects/p1/locations/l1/keyRings/r1/cryptoKeys/n1";
-  auto valid_key = MakeKmsKeyName(kValidKeyNameName);
+  auto const k_valid_key_name_name =
+      KmsKeyName("p1", "l1", "r1", "n1").FullName();
+  auto valid_key = MakeKmsKeyName(k_valid_key_name_name);
   ASSERT_THAT(valid_key, IsOk());
-  EXPECT_THAT(valid_key->FullName(), Eq(kValidKeyNameName));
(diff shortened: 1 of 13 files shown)


dbolduc:
I have a comment and a question.

1. I think we should keep the written out strings in these tests: `KmsKeyName`, `MakeProject`, `MakeBackup`, `MakeDatabase`, `MakeInstance`. I think it is useful to see the full string so that we can contrast it with the list of invalid strings. If you disagree, I would at least suggest that we drop the `k_` prefix from `auto const k_*`
2. (as an example) Why do we prefer using `Instance(Project("p1"), "i1")` over just `Instance("p1", "i1")` ? This question also applies to `Database` and `Backup` constructors (I know `Backup(std::string project_id, std::string instance_id, std::string backup_id)` doesn't exist, but maybe it should)

devbww:
> 1. I think we should keep the written out strings in these tests: `KmsKeyName`, `MakeProject`, `MakeBackup`, `MakeDatabase`, `MakeInstance`. I think it is useful to see the full string so that we can contrast it with the list of invalid strings.

I think the tests you're reaching for would be more related to the existing `FullName()` interfaces.  These `Make<Resource>()` tests should, I think, be testing that you can reconstruct a resource from its full name, independent of the actual full-name syntax.  So, I've updated them to directly look like `EXPECT_EQ(x, MakeXxx(x.Fullname()).value())`.  PTAL.

For the `Make<Resource>()` failure cases, we do have to build invalid strings by hand.

> 2. (as an example) Why do we prefer using `Instance(Project("p1"), "i1")` over just `Instance("p1", "i1")` ? This question also applies to `Database` and `Backup` constructors (I know `Backup(std::string project_id, std::string instance_id, std::string backup_id)` doesn't exist, but maybe it should)

I think we should prefer to use, say, a project-id string only in the constructor for `Project`.  Everything else that wants a project value should take a `Project` (and ditto for the others).  Using "stronger" types is generally a good thing.  Having the `Instance` constructor (for example) take a project-id string is an historical artifact that, I believe, we'd be better off eliminating.

dbolduc (approved):
Ack. Thanks for clarifying.

> Reviewed 12 of 13 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7328#-:-Mk0ygC72J-gthKJXtvb:bloppw6)* status: all files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[google/cloud/spanner/create_instance_request_builder_test.cc, line 56 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7328#-MjtMuiL9jKaWUHrpbJj:-MjtMuiM9CUgbBTGDDte:b713ljr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7770dbb26d930ccfdca985f28e16a287d12fc71d/google/cloud/spanner/create_instance_request_builder_test.cc#L56)):*
> ```cpp
>   EXPECT_EQ(in.project().FullName(), req.parent());
>   EXPECT_EQ(in.instance_id(), req.instance_id());
>   EXPECT_EQ(expected_name, req.instance().name());
> ```

`in.FullName()`
___
*[google/cloud/spanner/create_instance_request_builder_test.cc, line 78 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7328#-MjtNRet8m4aYnlta0j5:-MjtNRet8m4aYnlta0j6:b713ljr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7770dbb26d930ccfdca985f28e16a287d12fc71d/google/cloud/spanner/create_instance_request_builder_test.cc#L78)):*
> ```cpp
>   EXPECT_EQ(in.project().FullName(), req.parent());
>   EXPECT_EQ(in.instance_id(), req.instance_id());
>   EXPECT_EQ(expected_name, req.instance().name());
> ```

`in.FullName()`
___
*[google/cloud/spanner/create_instance_request_builder_test.cc, line 35 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7328#-Mk0xsyl6BogIugicriC:-Mk0xsyl6BogIugicriD:b-w8p7cp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a84cb1e4e156e996954f9f862006d75bdf6418de/google/cloud/spanner/create_instance_request_builder_test.cc#L35)):*
> ```cpp
>   EXPECT_EQ(in.project().FullName(), req.parent());
>   EXPECT_EQ(in.instance_id(), req.instance_id());
>   EXPECT_EQ(expected_name, req.instance().name());
> ```

nit: might as well just inline `in.FullName()` if we are doing that with `in.project().FullName()` and `in.instance_id()`

devbww:
> nit: might as well just inline in.FullName() if we are doing that with in.project().FullName() and in.instance_id()

Done.  Thanks.

PR #7335, opened 2021-09-20
feat(pubsub): add policy options; default common options

dbolduc:
Part of the work for #6306 

This change may seem strange on its own, because the code it is "replacing" is not touched in this PR. I did this because the full change is too large, and I want the PR's to be manageable. A (somewhat accurate) draft of the bigger picture can be found [here](https://github.com/dbolduc/google-cloud-cpp/commits/pubsub-options-big-picture).

Here are the sources for the default values:
* The logic related to `PUBSUB_EMULATOR_HOST` comes from `internal/emulator_overrides*` which will eventually be removed.
* The defaults for options from `GrpcOptionList` and `CommonOptionList` come from: [connection_options.cc](https://github.com/googleapis/google-cloud-cpp/blob/99a1cfd88cd67ed0e88c6c50712964ba8b3306c7/google/cloud/pubsub/connection_options.cc), except for `GrpcBackgroundThreadPoolSizeOption` which comes from [here (publisher)](https://github.com/googleapis/google-cloud-cpp/blob/99a1cfd88cd67ed0e88c6c50712964ba8b3306c7/google/cloud/pubsub/publisher_connection.cc#L123-L131) and [here(subscriber)](https://github.com/googleapis/google-cloud-cpp/blob/99a1cfd88cd67ed0e88c6c50712964ba8b3306c7/google/cloud/pubsub/subscriber_connection.cc#L147-L155)
* The defaults for options from `PolicyOptionList` come from: [here (publisher)](https://github.com/googleapis/google-cloud-cpp/blob/99a1cfd88cd67ed0e88c6c50712964ba8b3306c7/google/cloud/pubsub/publisher_connection.cc#L112-L113) and  [here (subscriber)](https://github.com/googleapis/google-cloud-cpp/blob/99a1cfd88cd67ed0e88c6c50712964ba8b3306c7/google/cloud/pubsub/subscriber_connection.cc#L136-L137)** (Note the special retry policy for the subscriber)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7335)

### DIFF
--- a/google/cloud/pubsub/internal/defaults.cc
+++ b/google/cloud/pubsub/internal/defaults.cc
@@ -13,8 +13,15 @@
 // limitations under the License.
 
 #include "google/cloud/pubsub/internal/defaults.h"
+#include "google/cloud/pubsub/internal/default_retry_policies.h"
 #include "google/cloud/pubsub/options.h"
+#include "google/cloud/common_options.h"
+#include "google/cloud/connection_options.h"
+#include "google/cloud/grpc_options.h"
+#include "google/cloud/internal/getenv.h"
+#include "google/cloud/internal/user_agent_prefix.h"
 #include "google/cloud/options.h"
+#include "default_retry_policies.h"
 #include <chrono>
 #include <limits>
 #include <thread>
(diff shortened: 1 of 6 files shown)


google/cloud/pubsub/internal/defaults.cc
```diff
@@ -25,16 +32,59 @@ namespace pubsub_internal {
...
-  return n == 0 ? kDefaultMaxConcurrency : n;
+  return n == 0 ? kDefaultThreadCount : n;
+}
+
+Options DefaultCommonOptions(Options opts) {
+  auto emulator = internal::GetEnv("PUBSUB_EMULATOR_HOST");
+  if (emulator.has_value()) {
+    opts.set<EndpointOption>(*emulator).set<GrpcCredentialOption>(
+        grpc::InsecureChannelCredentials());
+  }
```
devbww, line 48: Perhaps move this next to the other setting for `GrpcCredentialOption`?

devbww (approved):

google/cloud/pubsub/internal/defaults.cc
dbolduc: the `EndpointOption`s are colocated 🤷‍♂️

PR #7336, opened 2021-09-20
feat(pubsub): configure TopicAdminConnection with Options

dbolduc:
Part of the work for #6306 

This PR adds a method to configure `TopicAdminConnection` using `g::c::Options`. Users who might have relied on calling `MakeTopicAdminConnection({})` will not be broken because of the introduction of a non-constructible backwards compatibility type. (Credit to @devbww for the idea and implementation).

This PR also adds `Options` alternatives for `CreateDefaultPublisherStub()` and `CreateChannel()`. The old methods are still used by the rest of the `Make*Connection` methods. Leaving them allows for incremental changes. I put in some TODO's to remind myself to remove the old methods when I am done, though.

FYI: The same work for `TopicAdminConnection` will also be done for:
* `PublisherConnection`
* `SubscriberConnection`
* `SubscriptionAdminConnection`
* `SchemaAdminConnection`

(I started with `TopicAdminConnection`, because it didn't seem that involved)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7336)


google/cloud/pubsub/topic_admin_connection.h
```diff
@@ -158,6 +159,45 @@ class TopicAdminConnection {
   virtual ListTopicSnapshotsRange ListTopicSnapshots(ListTopicSnapshotsParams);
 };
 
+/**
+ * Creates a new `TopicAdminConnection` object to work with `TopicAdminClient`.
+ *
+ * @warning This method exists solely for backwards compatibility. It prevents
+ *     existing code, which calls `MakeTopicAdminConnection({})` from breaking,
+ *     due to ambiguity. Please use the `MakeTopicAdminConnection` method which
+ *     accepts `google::cloud::Options` instead.
```
dbolduc, line 168: I should change this to say "Please use `MakeTopicAdminConnection()` instead."

(I will wait to make the edit, because I suspect there are other things that will need to change)

google/cloud/pubsub/internal/backwards_compatibility_type.h
```diff
@@ -0,0 +1,45 @@
...
+namespace pubsub_internal {
+inline namespace GOOGLE_CLOUD_CPP_PUBSUB_NS {
+
+/**
+ * This is a non-constructible type. It is used to disambiguate overloaded
+ * methods that are passed an empty initializer list.
+ *
+ * See `MakeTopicAdminConnection()` for an example use case.
+ */
+struct BackwardsCompatibilityTypePleaseIgnore {
```
devbww, line 32: I'm not sure that "backwards compatibility" is the salient feature of this type, and therefore that there might be a better name.

google/cloud/pubsub/internal/backwards_compatibility_type.h
```diff
@@ -0,0 +1,45 @@
...
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_PUBSUB_INTERNAL_BACKWARDS_COMPATIBILITY_TYPE_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_PUBSUB_INTERNAL_BACKWARDS_COMPATIBILITY_TYPE_H
+
+#include "google/cloud/pubsub/version.h"
+#include "google/cloud/options.h"
+
+namespace google {
+namespace cloud {
+namespace pubsub_internal {
```
devbww, line 23: Do we think this type might be useful in similar circumstances in other libraries?  If so, it might be an idea to make it independent of pubsub.

google/cloud/pubsub/internal/backwards_compatibility_type.h
dbolduc: Do we like `NonConstructibleTypePleaseIgnore`?

google/cloud/pubsub/internal/backwards_compatibility_type.h
dbolduc: I agree that it could be useful in the future for other libraries.

My thinking was just that we could start with it in `pubsub_internal`, and only move it into `g::c::internal` if a use case in another library arises. I don't think that would constitute a breaking change, because nobody should be using the type.

I don't feel strongly either way, though.

google/cloud/pubsub/topic_admin_connection_test.cc
```diff
@@ -21,6 +21,7 @@
 #include "google/cloud/testing_util/scoped_log.h"
 #include "google/cloud/testing_util/status_matchers.h"
 #include "google/cloud/testing_util/validate_metadata.h"
+#include "internal/defaults.h"
```
dbolduc, line 24: #include "google/cloud/pubsub/internal/defaults.h"

coryan:
___
*[google/cloud/pubsub/internal/backwards_compatibility_type.h, line 32 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7336#-Mk4Sq9nGRQ8tLZLzOtv-r1-32:-Mk8Pve31editXF10qok:b-z5ag9k) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/12e9ae1f4b05d65f6c460edb66655529258b0552/google/cloud/pubsub/internal/backwards_compatibility_type.h#L32)):*

I think I would just go with just `NonConstructible` (`Unconstructible`?), and limit the description to something about cases where we want to disallow the instantiation of an object.  (Might it sometimes not be about backwards compatibility?)  That is, I'd suggest leaving mention of "initializer list" for the particular case that uses the unconstructible type that way (and drop mention of `MakeTopicAdminConnection()`).  I'd also drop the `PleaseIgnore`.
</blockquote></details>

:thumbsup: to what @devbww said.

devbww (approved):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7336)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dbolduc)

___
*[google/cloud/pubsub/topic_admin_connection.h, line 165 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7336#-Mk8kdVS3TDZsoOzsqI6:-Mk8kdVS3TDZsoOzsqI7:blqw9q7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/605c2a9f6e1c5c284b1c6c44c9afb2bcce14c358/google/cloud/pubsub/topic_admin_connection.h#L165)):*
> ```objc
>  * Creates a new `TopicAdminConnection` object to work with `TopicAdminClient`.
>  *
>  * @note This method exists solely for backwards compatibility. It prevents
> ```

FYI, I believe we try to avoid the term "method" for C++.

PR #7338, opened 2021-09-21
feat(pubsub): configure SchemaAdminConnection with Options

dbolduc:
Part of the work for #6306

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7338)

### DIFF
--- a/google/cloud/pubsub/schema_admin_connection.cc
+++ b/google/cloud/pubsub/schema_admin_connection.cc
@@ -14,9 +14,12 @@
 
 #include "schema_admin_connection.h"
 #include "google/cloud/pubsub/internal/default_retry_policies.h"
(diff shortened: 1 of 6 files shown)


google/cloud/pubsub/schema_admin_connection.cc
```diff
@@ -14,9 +14,12 @@
 
 #include "schema_admin_connection.h"
```
devbww, line 15: This include looks unconventional.

devbww (approved):

google/cloud/pubsub/schema_admin_connection.cc
dbolduc: Good catch.

PR #7342, opened 2021-09-22
feat(pubsub): configure SubscriberConnection with Options

dbolduc:
Part of the work for #6306 (so close....)

Half of `subscription_session_test.cc` used different test policies. Changing it to the test policies that every other test uses seems to be fine.

Only cleanups remain:
* tidying up code
* preferring the new methods in examples, benchmarks, quickstart
* deprecating old methods

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7342)

### DIFF
--- a/google/cloud/pubsub/subscriber_connection.cc
+++ b/google/cloud/pubsub/subscriber_connection.cc
@@ -14,10 +14,12 @@
 
 #include "google/cloud/pubsub/subscriber_connection.h"
 #include "google/cloud/pubsub/internal/default_retry_policies.h"
+#include "google/cloud/pubsub/internal/defaults.h"
 #include "google/cloud/pubsub/internal/subscriber_logging.h"
 #include "google/cloud/pubsub/internal/subscriber_metadata.h"
 #include "google/cloud/pubsub/internal/subscriber_round_robin.h"
 #include "google/cloud/pubsub/internal/subscription_session.h"
+#include "google/cloud/pubsub/options.h"
 #include "google/cloud/pubsub/retry_policy.h"
 #include "google/cloud/internal/random.h"
 #include "google/cloud/log.h"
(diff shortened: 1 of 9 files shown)


google/cloud/pubsub/subscriber_connection.cc
```diff
@@ -85,78 +99,39 @@ class SubscriberConnectionImpl : public pubsub::SubscriberConnection {
...
-}
-
-std::shared_ptr<pubsub::SubscriberConnection> MakeSubscriberConnection(
-    pubsub::Subscription subscription, pubsub::SubscriberOptions options,
-    pubsub::ConnectionOptions connection_options,
-    std::vector<std::shared_ptr<SubscriberStub>> stubs,
-    std::unique_ptr<pubsub::RetryPolicy const> retry_policy,
-    std::unique_ptr<pubsub::BackoffPolicy const> backoff_policy) {
-  auto default_retry_policy = [] {
-    // Subscribers are special: by default we want to retry essentially forever
```
coryan, line 127: Where did this code go?

dbolduc: Sorry, this was confusing because of how I broke up the PRs. But the replacement code was introduced in #7335.

We default the subscriber only retry policy value here:
https://github.com/googleapis/google-cloud-cpp/blob/6112ac217b43961b2f19accac6070fae1f05f93c/google/cloud/pubsub/internal/defaults.cc#L132-L141

and test for the value here:
https://github.com/googleapis/google-cloud-cpp/blob/6112ac217b43961b2f19accac6070fae1f05f93c/google/cloud/pubsub/internal/defaults_test.cc#L168-L171

dbolduc: And our entry point calls `DefaultSubscriberOptions()` https://github.com/googleapis/google-cloud-cpp/blob/9479521e1fd47167a5849f5c42cad114d7410ce1/google/cloud/pubsub/subscriber_connection.cc#L54

coryan: Ack, SGTM

coryan (approved):

PR #7345, opened 2021-09-22
feat(common): add convenience method for generating a BackgroundThreadFactory via a CQ

dbolduc:
This change does 3 things.

1. moves definition of `google::cloud::BackgroundThreadsFactory` out of `grpc_options.h` and into its own file.
2. adds a `CustomBackgroundThreads(CompletionQueue cq)` convenience method for generating background threads that just return that CQ.
3. implements the new method in our code, to clean things up.

I think we want the function for convenience. Users that relied on the simplicity of `ConnectionOptions::DisableBacgkroundThreads(cq)` in pubsub or `ClientOptions::DisableBackgroundThreads(cq)` in bigtable will appreciate not having to construct the factory themselves. I think the difference in our code speaks for itself.

I am obviously open to changing the name and/or the location of the function.

Other names considered were: `DisableBackgroundThreads(cq)`, `DisableBackgroundThreadsFactory(cq)`, `CustomBackgroundThreadsFactory(cq)`.

I considered putting the function in `grpc_options.h` and `background_threads.h`, but neither seemed to be the right place for it.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7345)

### DIFF
--- a/google/cloud/background_threads_factory_test.cc
+++ b/google/cloud/background_threads_factory_test.cc
@@ -14,6 +14,9 @@
 
 #include "google/cloud/background_threads_factory.h"
 #include <gmock/gmock.h>
+#include <chrono>
+#include <future>
+#include <thread>
 
 namespace google {
 namespace cloud {
--- a/google/cloud/background_threads_factory_test.cc
+++ b/google/cloud/background_threads_factory_test.cc
@@ -1,0 +1,47 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/background_threads_factory.h"
+#include <gmock/gmock.h>
+
+namespace google {
+namespace cloud {
+inline namespace GOOGLE_CLOUD_CPP_NS {
+namespace {
+
+TEST(BackgroundThreadsFactory, CustomBackgroundThreads) {
+  using ms = std::chrono::milliseconds;
+  CompletionQueue cq;
+  auto background = CustomBackgroundThreads(cq)();
+
+  // Schedule some work that cannot execute because there is no thread draining
+  // the completion queue.
+  promise<std::thread::id> p;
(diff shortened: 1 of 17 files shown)


devbww:
> I considered putting the function in grpc_options.h and background_threads.h, but neither seemed to be the right place for it.

I think I would vote for putting it in `background_threads.h`, short of some other consideration.

> Other names considered were: DisableBackgroundThreads(cq), DisableBackgroundThreadsFactory(cq), CustomBackgroundThreadsFactory(cq).

There is a bit of a mismatch between "Custom" versus "CustomerSupplied"/"AutomaticallyCreated", but perhaps that would be clarified if `DefaultBackgroundThreads()` was moved here too.

(Aside: Perhaps "BackgroundThreads" has always been confusing given that it is really a `CompletionQueue` that is being run somehow.)

google/cloud/background_threads_factory_test.cc
```diff
@@ -0,0 +1,50 @@
...
+  // the completion queue.
+  promise<std::thread::id> p;
+  auto background_thread_id = p.get_future();
+  background->cq().RunAsync(
+      [&p](CompletionQueue&) { p.set_value(std::this_thread::get_id()); });
+  EXPECT_NE(std::future_status::ready, background_thread_id.wait_for(ms(10)));
+
+  // Verify we can create our own threads to drain the completion queue.
+  std::thread t([&cq] { cq.Run(); });
+  EXPECT_EQ(t.get_id(), background_thread_id.get());
```
devbww, line 41: There must be a race here, right?

coryan:
> > Other names considered were: DisableBackgroundThreads(cq), DisableBackgroundThreadsFactory(cq), CustomBackgroundThreadsFactory(cq).
> 
> There is a bit of a mismatch between "Custom" versus "CustomerSupplied"/"AutomaticallyCreated", but perhaps that would be clarified if `DefaultBackgroundThreads()` was moved here too.

`CustomerSuppliedBackgroundThreads` is probably better.

> 
> (Aside: Perhaps "BackgroundThreads" has always been confusing given that it is really a `CompletionQueue` that is being run somehow.)

You are not wrong.  My choice of names here was probably less than ideal, what we want to express is:

* Background operations (mostly asynchronous I/O, but also timers, and sometimes pure code) need a `CompletionQueue` to run.  This is largely imposed by gRPC because asynchronous I/O and timers happen through completion queues.
* The completion queue needs one or more threads to "service" them.
* The thread pool servicing the completion queue can be created by the library or provided by the application.

If we take this opportunity to provide more meaningful names, I am all for it.

google/cloud/background_threads_factory_test.cc
devbww: Ignore that ... I'm just confused.

dbolduc:
The discussion here is valuable, but I suspect we will take a different approach. So I am closing this PR.

Issue #7348, opened 2021-09-23
Deprecate and remove pubsub configuration classes and methods that predated `Options`


dbolduc:
basically #6642, but in pubsub.

With #6306 implemented, we can start to deprecate `PublisherOptions` and `SubscriberOptions`, as they are no longer needed or wanted.

Issue #7347, opened 2021-09-23
Update pubsub samples to use new Make*Connection functions


dbolduc:
Pubsub has a [publisher](https://github.com/googleapis/google-cloud-cpp/blob/7414d8ad4517560e71ffe306d9ddd2ccd5a3dc8d/google/cloud/pubsub/samples/samples.cc#L1349) and [subscriber](https://github.com/googleapis/google-cloud-cpp/blob/7414d8ad4517560e71ffe306d9ddd2ccd5a3dc8d/google/cloud/pubsub/samples/samples.cc#L1597) sample to demonstrate the use of custom thread pools.

These are configured using `ConnectionOptions::DisableBackgroundThreads(cq)`. We want to update them to use the new functions configured with `Options`, but we are blocked by #7346 

We also need to update the documentation in [Publisher](https://github.com/googleapis/google-cloud-cpp/blob/acb7e732ea1d827a826b162d8661d3767a7b1dc5/google/cloud/pubsub/publisher.h#L67-L72) and [Subscriber](https://github.com/googleapis/google-cloud-cpp/blob/acb7e732ea1d827a826b162d8661d3767a7b1dc5/google/cloud/pubsub/subscriber.h#L58-L63)

Issue #7344, opened 2021-09-22
Update pubsub quickstart


dbolduc:
When we release the next version of our library (and vcpkg has the update), change: https://github.com/googleapis/google-cloud-cpp/blob/7414d8ad4517560e71ffe306d9ddd2ccd5a3dc8d/google/cloud/pubsub/quickstart/quickstart.cc#L31-L32

to:
```c++
  auto publisher = pubsub::Publisher(
      pubsub::MakePublisherConnection(pubsub::Topic(project_id, topic_id)));
```

also, regenerate the README

PR #7350, opened 2021-09-23
cleanup(pubsub): remove unused code; tidy up other code; edit documentation

### LINKED ISSUE #6306: Convert google/cloud/pubsub to use g::c::Options

dbolduc:
Fixes #6306 

The only remaining tasks are captured by: #7344, #7347 , #7348

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7350)


google/cloud/pubsub/internal/defaults.cc
```diff
@@ -61,10 +60,14 @@ Options DefaultCommonOptions(Options opts) {
     opts.set<GrpcTracingOptionsOption>(internal::DefaultTracingOptions());
   }
   if (!opts.has<pubsub::RetryPolicyOption>()) {
-    opts.set<pubsub::RetryPolicyOption>(DefaultRetryPolicy()->clone());
+    opts.set<pubsub::RetryPolicyOption>(
+        std::make_shared<pubsub::LimitedTimeRetryPolicy>(
```
devbww, line 64: It might be worth sticking with the `clone()` idiom rather than using `make_shared` directly.

coryan (approved):

google/cloud/pubsub/internal/defaults.cc
dbolduc: Done. Maybe you can explain to me why we might prefer one over the other?

devbww: The big difference is that `clone()` is polymorphic.  While that doesn't matter in these particular cases, I reckoned that if we have `clone()` anyway then we might as well use it, even when we know the compile-time type.

Aside: I think you missed a couple of occurrences in the `test_retry_policies.cc` edits, but not worth worrying about now I'd proffer.

dbolduc: yup, I missed those 🤦‍♂️

PR #7366, opened 2021-09-27
fix(storage): eliminate dangling function references

devbww:
`absl::FunctionRef` is a non-owning wrapper, so the temporary function
objects used to initialize the `parsers[]` array will be destroyed before
they can be used. So, change to the `std::function` value type instead.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7366)


coryan (approved):
Oops, thanks!

PR #7369, opened 2021-09-27
test(storage): accept normalized PAP values

coryan:
Part of the work for #7365 

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7369)

### DIFF
--- a/google/cloud/storage/bucket_metadata.cc
+++ b/google/cloud/storage/bucket_metadata.cc
@@ -32,10 +32,10 @@ namespace internal {
 absl::optional<std::string> const& NormalizePap(
     absl::optional<std::string> const& pap) {
   static auto const* normalized =
-      new absl::optional<std::string>(PublicAccessInherited());
+      new absl::optional<std::string>(PublicAccessPreventionInherited());
   if (!pap.has_value()) return pap;
-  if (*pap == PublicAccessPreventionUnspecified()) return pap;
-  return *normalized;
+  if (*pap == PublicAccessPreventionUnspecified()) return *normalized;
+  return pap;
 }
 }  // namespace internal
 
--- a/google/cloud/storage/bucket_metadata.h
+++ b/google/cloud/storage/bucket_metadata.h
@@ -217,7 +217,7 @@ struct BucketIamConfiguration {
 //@{
 /// @name Public Access Prevention helper functions
 inline std::string PublicAccessPreventionEnforced() { return "enforced"; }
-inline std::string PublicAccessInherited() { return "inherited"; }
+inline std::string PublicAccessPreventionInherited() { return "inherited"; }
 inline std::string PublicAccessPreventionUnspecified() { return "unspecified"; }
 //@}
 
--- a/google/cloud/storage/bucket_metadata_test.cc
+++ b/google/cloud/storage/bucket_metadata_test.cc
@@ -166,6 +166,18 @@ BucketMetadata CreateBucketMetadataForTest() {
   return internal::BucketMetadataParser::FromString(text).value();
 }
 
+TEST(BucketMetadataTest, NormalizePap) {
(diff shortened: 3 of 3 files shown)


dbolduc (changes requested):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7369)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @coryan and @dbolduc)

___
*[google/cloud/storage/bucket_metadata.cc, line 37 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7369#-MkdUq360-Bd3N9DOQGL:-MkdUq360-Bd3N9DOQGM:b-tfc1sz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f6cc8ae0827b09163a3aac05d1e5b6a6c6e4539d/google/cloud/storage/bucket_metadata.cc#L37)):*
> ```
>   if (*pap == PublicAccessPreventionUnspecified()) return pap;
>   return *normalized;
> ```

maybe I'm not getting it, but don't we want both "inherited" and "unspecified" to map to the same value? which I think is "inherited".

wouldn't that look like:
```c++
if (*pap == PublicAccessPreventionUnspecified()) return *normalized;
return pap;
```

dbolduc:
also, I know this code's lifespan should be very short, but do should we add any tests for it?

coryan (commented):
Thanks, PTAL.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7369#-:-MkddUnO7ga3qARgK044:br4ikdp)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @dbolduc)

___
*[google/cloud/storage/bucket_metadata.cc, line 37 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7369#-MkdUq360-Bd3N9DOQGL:-MkddSWoDvcpCX-qyacG:b-lx4fjp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f6cc8ae0827b09163a3aac05d1e5b6a6c6e4539d/google/cloud/storage/bucket_metadata.cc#L37)):*

> ```
>   if (*pap == PublicAccessPreventionUnspecified()) return pap;
>   return *normalized;
> ```

maybe I'm not getting it, but don't we want both "inherited" and "unspecified" to map to the same value? which I think is "inherited".

wouldn't that look like:
```c++
if (*pap == PublicAccessPreventionUnspecified()) return *normalized;
return pap;
```


</blockquote></details>

Fixed.

dbolduc (approved):

google/cloud/storage/bucket_metadata.cc
```diff
@@ -28,6 +28,16 @@ namespace google {
 namespace cloud {
 namespace storage {
 inline namespace STORAGE_CLIENT_NS {
+namespace internal {
+absl::optional<std::string> const& NormalizePap(
+    absl::optional<std::string> const& pap) {
+  static auto const* normalized =
+      new absl::optional<std::string>(PublicAccessPreventionInherited());
```
devjgm, line 35: Consider just `new auto(PublicAccessPreventionInherited())`, because I'm not seeing any need to wrap the returned string in an `absl::optional`, just to be unconditionally dereferenced a few lines below.

coryan: The dereference is (I think) changing from `absl::optional<..> const*` to `absl::optional<..> const&`, right?

devjgm: Oh, I'm sorry. I missed that this was returning a const-ref. My mistake.

PR #7378, opened 2021-09-30
fix: substitute build metadata (git commit) in bazel builds

### LINKED ISSUE #4035: Add git commit hash to bazel builds
We try to stamp our builds with a "git commit hash". [Here's how we get this info](https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/CMakeLists.txt#L30) in our CMake builds. This info is exposed in C++ in the `g::c::s::internal::BuildMetadata()` function [here](https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/internal/build_info.h#L73).

The problem is that it's unclear how to get this info from bazel, so currently our bazel builds always report a git hash of "unknown".
https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/BUILD#L26

This issue is to try to do the following:

Figure out how to get the git commit hash (i.e., ` git rev-parse --short HEAD`) when running a bazel build. This should work when our library is built directly, AND when it's build as a dep of some other project.

Some notes:

* Running the `git` command directly in a `genrule` doesn't work because it is executed in an environment without the `.git/` directory structure. It *might* work to add `.git/**` to the `srcs` attribute of the `genrule`, but I haven't tried this.

* Using bazel's `--workspace_status_command=` argument *does* work. Almost. It works when one specifies that flag. And we could add that flag to all of our bazel invocations by adding it to our `.bazelrc` file. However, this would not get picked up and used when our project is a dep of another, which is probably the common case.

Related to PR googleapis/google-cloud-cpp-spanner#278

devbww:
- Use bazel's `workspace_status_command` to report the git commit.
- Substitute it into `google::cloud::internal::build_metadata()`.
- Test that `build_metadata()` does not contain a "@" delimiter.

Fixes #4035.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7378)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# NOTE: This file is not loaded if google-cloud-cpp is built as part of a
+# larger project.
+
 # To workaround a bug in Bazel [1], gRPC requires this flag on macOS builds,
 # and there is (AFAIK) no way to "inherit" their definitions.
 #   [1]: https://github.com/bazelbuild/bazel/issues/4341
--- a/.bazelrc
+++ b/.bazelrc
@@ -21,6 +21,14 @@ build --copt=-DGRPC_BAZEL_BUILD
 # inside a docker image.
 build --experimental_convenience_symlinks=ignore
 
+# Use host-OS-specific config lines from bazelrc files.
+build --enable_platform_specific_config=true
+
+# Specify a binary that Bazel runs before each build. The program can report
+# information about the status of the workspace, such as the current source
+# control revision.
+build:linux --workspace_status_command=bazel/workspace_status_command.sh
+
 # Clang Sanitizers, use with (for example):
 #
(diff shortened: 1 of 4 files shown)


.bazelrc
```diff
@@ -21,6 +21,14 @@ build --copt=-DGRPC_BAZEL_BUILD
 # inside a docker image.
 build --experimental_convenience_symlinks=ignore
 
+# Use host-OS-specific config lines from bazelrc files.
```
coryan, line 24: AFAIK, this file is not loaded if `google-cloud-cpp` is built as part of a larger project.   That is not a blocker for this change, but maybe we should document the limitations somewhere?

coryan (approved):

.bazelrc
devbww: Added such a note.

[Indeed, if this code was part of a bigger project they may not be using `bazel` or `git`, so they'd have to figure out how to generate their own build identifier.]

PR #7383, opened 2021-09-30
feat(generator): add "generator" to API client header "gccl" semver

devbww:
1. Strip the leading "v" from the value of the "gccl" component in the
   `X-Goog-Api-Client` header so that what remains is a valid semantic
   version string (see https://semver.org/).

2. Add a "generator" build identifier to the "gccl" semver for requests
   sent from generated clients.

The upshot is that the library-version component of the client header
will be `gccl/<major>.<minor>.<patch>+<git-hash>.generator`, allowing
us to identify requests from generated clients.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7383)


coryan (approved):
Make sure you edit the commit message to s/semv**a**r/semver/

PR #7390, opened 2021-10-01
fix(storage): prevent crashes on double Close()

coryan:
The library was crashing if the application called `.Close()` twice on a
`storage::ObjectWriteStream`. Arguably one should not do that, but I
think a crash is really sad behavior in this case, and it is trivial to
return the status from the previous `.Close()`.

I think @devbww asked me about this in the original review, and I had
convinced myself the pre-conditions prevented the code from running
twice. I was wrong.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7390)

### DIFF
--- a/google/cloud/storage/internal/object_write_streambuf.cc
+++ b/google/cloud/storage/internal/object_write_streambuf.cc
@@ -75,6 +75,7 @@ bool ObjectWriteStreambuf::ValidateHash(ObjectMetadata const& meta) {
     auto function = std::move(hash_function_);
     hash_values_ = std::move(*function).Finish();
   }
+  if (!hash_validator_) return !hash_validator_result_.is_mismatch;
   auto validator = std::move(hash_validator_);
   validator->ProcessMetadata(meta);
   hash_validator_result_ = std::move(*validator).Finish(hash_values_);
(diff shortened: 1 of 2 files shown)


devbww (approved):

google/cloud/storage/internal/object_write_streambuf.cc
```diff
@@ -75,6 +75,7 @@ bool ObjectWriteStreambuf::ValidateHash(ObjectMetadata const& meta) {
     auto function = std::move(hash_function_);
     hash_values_ = std::move(*function).Finish();
   }
+  if (!hash_validator_) return !hash_validator_result_.is_mismatch;
```
dbolduc, line 78: not necessary, but should we log some sort of warning here?

coryan: Thanks.  I am not sure how the warning would help.  I mean, yes, maybe they would see it and stop calling `.Close()` twice. But would it matter now that `.Close()` is idempotent?

dbolduc: Ah, point taken.

Issue #7394, opened 2021-10-02
Treat certain known `kInternal` errors as retryable


dbolduc:
We do this in [spanner](https://github.com/googleapis/google-cloud-cpp/blob/6f6967cb56fe1765b493b46c6b54c8dd59d41f73/google/cloud/spanner/retry_policy.h#L39-L51). It will be added to bigtable shortly. As far as I can tell, any service that performs streaming reads will need to be updated. 

My guess is that we will want to refactor the implementation into common. (see [here](https://github.com/dbolduc/google-cloud-cpp/commit/a9eba809e804453ad80281e24aea59ed73fdcb93) for what I had in mind)

We need to:
* determine exactly which messages should be retried. (I think [these](https://github.com/dbolduc/google-cloud-cpp/blob/a9eba809e804453ad80281e24aea59ed73fdcb93/google/cloud/internal/retry_policy.cc#L28-L30))
* determine which libraries need this. (and maybe make a note for generated libraries that the hand-written `retry_traits.h` might need this)
* write/refactor the code

Also it would be nice to have some link to an internal (within Google) issue, such that if the internal issue is fixed, we would know that we can remove the message check in our code. (Also, if someone decides to edit the error message, we could potentially learn about it and update our code)

dbolduc:
I think we should refactor [spanner's retry traits logic](https://github.com/googleapis/google-cloud-cpp/blob/f779bb99395bbdfe3aac51864ff2abf2a0b6dffc/google/cloud/spanner/retry_policy.h#L39-L50) to invoke the common helper function. I do not think we should widely add this processing of `kInternal` errors to all services with streaming reads, until someone raises it as an issue.

PR #7395, opened 2021-10-02
fix(bigtable): retry internal errors known to be transient

dbolduc:
The Bigtable part of #7394. 

I wanted to do it all at once, but I think it's best to just do the Bigtable piece and deliver value for that library now for known Bigtable error messages. (Instead of debating whether we can drop/change some spanner messages or debating which file the common function should go in).

The code and test for `IsTransientInternalError()` should be easy to refactor into common. (I decided to just have a redundant test specifically for RST_STREAM, for when we pluck out the Basic test)

I will consolidate what I have learned about the issue and share with the team offline when I have time, so we can decide how to address the rest of the issue.

### DIFF
--- a/google/cloud/bigtable/internal/bulk_mutator.cc
+++ b/google/cloud/bigtable/internal/bulk_mutator.cc
@@ -98,18 +98,17 @@ std::vector<int> BulkMutatorState::OnRead(
     auto const index = static_cast<std::size_t>(entry.index());
     auto& annotation = annotations_[index];
     annotation.has_mutation_result = true;
-    auto const& status = entry.status();
-    auto const code = static_cast<grpc::StatusCode>(status.code());
+    auto const& status = MakeStatusFromRpcError(entry.status());
     // Successful responses are not even recorded, this class only reports
     // the failures.  The data for successful responses is discarded, because
     // this class takes ownership in the constructor.
-    if (grpc::StatusCode::OK == code) {
+    if (status.ok()) {
       res.push_back(annotation.original_index);
       continue;
     }
     auto& original = *mutations_.mutable_entries(static_cast<int>(index));
     // Failed responses are handled according to the current policies.
-    if (SafeGrpcRetry::IsTransientFailure(code) &&
+    if (SafeGrpcRetry::IsTransientFailure(status) &&
         (annotation.idempotency == Idempotency::kIdempotent)) {
       // Retryable requests are saved in the pending mutations, along with the
       // mapping from their index in pending_mutations_ to the original
--- a/google/cloud/bigtable/rpc_retry_policy.cc
+++ b/google/cloud/bigtable/rpc_retry_policy.cc
@@ -14,12 +14,30 @@
 
 #include "google/cloud/bigtable/rpc_retry_policy.h"
 #include "google/cloud/grpc_error_delegate.h"
+#include "absl/strings/match.h"
 #include <sstream>
 
 namespace google {
 namespace cloud {
 namespace bigtable {
 inline namespace BIGTABLE_CLIENT_NS {
+namespace internal {
+bool IsTransientInternalError(Status const& status) {
+  // Treat the unexpected termination of the gRPC connection as retryable.
+  // There is no explicit indication of this, but it will result in an
+  // INTERNAL status with one of the `kTransientFailureMessages`.
+  if (status.code() == StatusCode::kInternal) {
+    static constexpr char const* kTransientFailureMessages[] = {
+        "RST_STREAM", "Received Rst Stream",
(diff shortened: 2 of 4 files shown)


google/cloud/bigtable/rpc_retry_policy.h
```diff
@@ -27,34 +28,31 @@ namespace cloud {
...
-
   static inline bool IsOk(google::cloud::Status const& status) {
     return status.ok();
   }
   static inline bool IsTransientFailure(google::cloud::Status const& status) {
-    return IsTransientFailure(status.code());
+    auto const code = status.code();
+    return code == StatusCode::kAborted || code == StatusCode::kUnavailable ||
+           code == StatusCode::kDeadlineExceeded ||
+           IsTransientInternalError(status);
```
devbww, line 46: Can't you just call `google::cloud::internal::IsTransientInternalError()` rather than having a bigtable copy of it?

dbolduc: Well I was thinking 2 things.
1. I wanted to avoid discussion about whether the function should be defined in `g/c/i/retry_policy.h`. Maybe it belongs in `g/c/grpc_error_delegate.h`. (This has backfired, as we are now discussing whether the function belongs in `g/c/bt/rpc_retry_policy.h` 😆)
2. Maybe there will be a transient internal error for a service that should not be retried by the others. Then refactoring into common might not be ideal. (Currently, this is not the case, which is why I think the function does belong in common)

Anyway, the function has been moved. `spanner/` is untouched. PTAL.

google/cloud/bigtable/rpc_retry_policy.cc
```diff
@@ -14,12 +14,14 @@
 
 #include "google/cloud/bigtable/rpc_retry_policy.h"
 #include "google/cloud/grpc_error_delegate.h"
+#include "absl/strings/match.h"
```
coryan, line 17: now unused?

google/cloud/internal/retry_policy.cc
```diff
@@ -0,0 +1,41 @@
...
+
+bool IsTransientInternalError(Status const& status) {
+  // Treat the unexpected termination of the gRPC connection as retryable.
+  // There is no explicit indication of this, but it will result in an
+  // INTERNAL status with one of the `kTransientFailureMessages`.
+  if (status.code() == StatusCode::kInternal) {
+    static constexpr char const* kTransientFailureMessages[] = {
+        "RST_STREAM", "Received Rst Stream",
+        "Received unexpected EOS on DATA frame from server"};
+    for (auto const& message : kTransientFailureMessages) {
```
coryan, line 31: nit: consider

```cc
  if (status.code() != Statuscode::kInternal) return false;
  return std::any_of(
      std::begin(kTransienFailureMessages),
      std::end(kTransientFailureMessages),
     [&status](char const* message) {
        return absl::StrContains(status.message(), message);
   });
```

google/cloud/bigtable/rpc_retry_policy.cc
dbolduc: Thanks. Fixed.

google/cloud/internal/retry_policy.cc
dbolduc: Updated. I did consider an uglier version of it.

google/cloud/internal/retry_policy.cc
```diff
@@ -0,0 +1,42 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/internal/retry_policy.h"
+#include "absl/strings/match.h"
+
```
coryan, line 17: nit: `#include <algorithm>`

coryan (approved):
LGTM, I think maybe @devbww should take another look.

google/cloud/bigtable/internal/bulk_mutator.cc
```diff
@@ -98,18 +98,17 @@ std::vector<int> BulkMutatorState::OnRead(
     auto const index = static_cast<std::size_t>(entry.index());
     auto& annotation = annotations_[index];
     annotation.has_mutation_result = true;
-    auto const& status = entry.status();
-    auto const code = static_cast<grpc::StatusCode>(status.code());
+    auto const& status = MakeStatusFromRpcError(entry.status());
```
devbww, line 101: Avoid capturing value by reference?

devbww (approved):

PR #7398, opened 2021-10-04
ci: use CodeQL for security scans

coryan:
This sets up a GitHub action to run the CodeQL security scans. It only
runs once a day because it is too slow for PRs. It is unclear if this is
useful (beyond bragging rights), but it is worth testing for a while.

Previous results (no alerts) visible at:

https://github.com/coryan/google-cloud-cpp/runs/3792838206?check_suite_focus=true
https://github.com/coryan/google-cloud-cpp/security/code-scanning

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7398)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -5,8 +5,6 @@ on:
   # that much, start with a time that allows for easier troubleshooting.
   schedule:
     - cron: '00 22 * * *'
-  push:
-    branches: ['ci-experiment-codeql']
 
 jobs:
   analyze:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -30,10 +30,10 @@ jobs:
         with:
           path: "build/vcpkg"
           repository: "microsoft/vcpkg"
-      - name: Read vcpkg version
+          fetch-depth: 0
+      - name: Checkout pinned vcpkg version
         run: >
-          cat ci/etc/vcpkg-commit.txt;
-          git -C build/vcpkg checkout $(<ci/etc/vcpkg-commit.txt);
+          git -C build/vcpkg checkout -q $(<ci/etc/vcpkg-commit.txt)
       - name: cache-vcpkg
         id: cache-vcpkg
         uses: actions/cache@v2
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,84 @@
+name: "CodeQL"
+
+on:
+  # Run the analysis once every 24 hours. The actual time does not matter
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,84 @@
+name: "CodeQL"
+
+on:
+  # Run the analysis once every 24 hours. The actual time does not matter
+  # that much, start with a time that allows for easier troubleshooting.
+  schedule:
+    - cron: '00 22 * * *'
+  push:
+    branches: ['ci-experiment-codeql']
```
dbolduc, line 9: main?

coryan: This is temporary.  I will remove the whole `push` section.

coryan: Done now.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Checkout vcpkg
+        uses: actions/checkout@v2
+        with:
+          path: "build/vcpkg"
+          repository: "microsoft/vcpkg"
+          ref: "7e7bcf89171b7ed84ece845b1fa596a018e462f2"
```
dbolduc, line 31: We might want to do something like [this](https://github.community/t/read-from-file-and-assign-to-env-variable/174936/2) to read the vcpkg commit from `ci/etc/vckpg-commit.txt`

(if we keep this action, that is)

coryan: That cannot work, each step runs in its own docker container, you cannot pass variables from one to the next. But I can add a step that performs the checkout at the right SHA. PTAL

dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7398)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[.github/workflows/codeql-analysis.yml, line 31 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7398#-MlBnAFr9PiSgX6CIjKp-r2-31:-MlC-t3R0_islTT_F_nf:bdahrph) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0dae126125e9c4f304f1838e4784da04a10cd8d8/.github/workflows/codeql-analysis.yml#L31)):*

That cannot work, each step runs in its own docker container, you cannot pass variables from one to the next. But I can add a step that performs the checkout at the right SHA. PTAL
</blockquote></details>

Ah, thanks for finding a fix that doesn't require us to update the vcpkg commit in multiple places.

PR #7409, opened 2021-10-06
doc: update CHANGELOG and version numbers for v1.32.1

dbolduc:
No need to update the ABI, right?

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7409)


coryan:
> No need to update the ABI, right?

I would just run the build to updated them.

CHANGELOG.md
```diff
@@ -69,6 +69,12 @@
 
 ## v1.33.0 - TBD
 
+## v1.32.1 - 2021-10
+
+### [Pub/Sub](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/pubsub/README.md)
+
+* fix(pubsub): do not overwrite ConnectionOptions ([#7406](https://github.com/googleapis/google-cloud-cpp/pull/7406))
```
coryan, line 76: I should have called this earlier: this is what it does not do, but there are many things it does not do.  Maybe "respect non-default values in pubsub::ConnectionOptions"?

coryan (approved):
a bit nit picky, but I would be confused by the CHANGELOG contents.

coryan (approved):

PR #7444, opened 2021-10-12
doc(doxygen): use `omit_this_inline_ns` as doxygen inline namespace name

devjgm:
Related to: https://github.com/googleapis/google-cloud-cpp/issues/5976

Doxygen doesn't know how to collapse C++ inline namespaces, so it renders them as normal namespace names. For example, see https://googleapis.dev/cpp/google-cloud-storage/latest/classgoogle_1_1cloud_1_1storage_1_1v1_1_1Client.html where the class name is `google::cloud::storage::v1::Client`, which _includes_ the `v1` inline namespace. In https://github.com/googleapis/google-cloud-cpp/issues/5976 we're planning to change the inline namespace name, and this will break all of our doxygen links, which also must include the inline namespace name in a symbol's fully qualified name.

This PR changes the inline namespace name WITHIN DOXYGEN ONLY to be something that is agnostic of any particular library version, so that we do not need to change doxygen docs with each release. I also chose a name that suggests to users that they should **omit** the inline namespace name: `omit_this_inline_ns`. 

So for example:

OLD | NEW
--- | ---
`google::cloud::storage::v1::Client` | `google::cloud::storage::omit_this_inline_ns::Client`

Alternatively, we could also name the inline namespace something like `vX_Y_Z` or something, which does look nicer, but also may seem to suggest that we want users to spell it.

Example:
image

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7444)

### DIFF
--- a/google/cloud/bigtable/version.h
+++ b/google/cloud/bigtable/version.h
@@ -39,7 +39,7 @@ namespace cloud {
  */
 namespace bigtable {
 /**
- * The inlined, versioned namespace for the Cloud Bigtable C++ client APIs.
+ * Versioned inline namespace that users should generally avoid spelling.
  *
  * Applications may need to link multiple versions of the Cloud Bigtable C++
  * client, for example, if they link a library that uses an older version of
--- a/google/cloud/grpc_utils/version.h
+++ b/google/cloud/grpc_utils/version.h
@@ -23,6 +23,7 @@ namespace cloud {
  * Contains all the Cloud C++ gRPC Utilities APIs.
  */
 namespace grpc_utils {
+/// Versioned inline namespace that users should generally avoid spelling.
 inline namespace GOOGLE_CLOUD_CPP_NS {
 using ::google::cloud::version;
 using ::google::cloud::version_major;
--- a/google/cloud/pubsub/version.h
+++ b/google/cloud/pubsub/version.h
@@ -33,7 +33,7 @@ namespace cloud {
  */
 namespace pubsub {
 /**
- * The inlined, versioned namespace for the Cloud Pubsub C++ client APIs.
+ * Versioned inline namespace that users should generally avoid spelling.
  *
  * Applications may need to link multiple versions of the Cloud pubsub C++
  * client, for example, if they link a library that uses an older version of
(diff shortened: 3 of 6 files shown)


devjgm:
> Can you check if this generates a page for the namespace, i.e. the equivalent of:
> 
> https://googleapis.dev/cpp/google-cloud-storage/latest/namespacegoogle_1_1cloud_1_1storage_1_1v1.html

Yes, it does generate a page like that.

> 
> If so, maybe we should add a filter for these symbols (like we do for `internal`) so they are removed?

Good suggestion. Sadly, I couldn't get this to work. If I omit the symbols in `omit_this_inline_ns` then we can no longer reference those symbols. I tried various spellings and alternatives around this idea and couldn't get it to work.

> 
> Regardless of how we mark this, I think we should:
> 
> * Create a GitHub discussion titled "What is `omit_this_inline_ns`?" so the search engines can find it.

Good idea. Done: https://github.com/googleapis/google-cloud-cpp/discussions/7447

> * And, we should have a `.dox` page that describes the namespace and says why Doxygen is terrible and we need this workaround.

Creating a `.dox` page adds additional documentation to the end of the documentation for the inline namespaces that already exists in the source file. It doesn't replace it. So I think the best we can do is to tweak the documentation for these inline namespaces in the source itself. So I started the documentation for each inline namespace with `Versioned inline namespace that users should generally avoid spelling.`, which should show up as the one-line summary in the generated documentation.

> 
> Also, I wonder if it is possible to trick Doxygen into using an illegal C++ character and name the namespace `$versioned-inline-ns$` or something like that.

I didn't try with dollar signs specifically, but I tried a bunch of other options last night (parens, angle brackets, square brackets, etc), and none worked. I think it needs to be valid C++ because it's still using clang to parse the files.

> 
> [START bikeshed]
> 
> s/omit_this_inline_ns/_inline_ns_for_doxygen_/

I prefer keeping the word "omit" in the name to suggest what users' should do, but I don't feel strongly. If you prefer another name, that's fine with me and I can change it in the code and in the Discussion linked above.

Issue #7526, opened 2021-10-28
Generate Bigtable Admin APIs


dbolduc:
Configure the generator to generate the Instance Admin and Table Admin APIs into `bigtable/admin/`

Consider copying and converting the current admin integration tests in `bigtable/admin/integration_tests/`

PR #7550, opened 2021-11-02
feat(bigtable): add generated admin APIs

dbolduc:
Part of the work for #7526 

This change configures the generator to generate (and format) APIs for the Bigtable Instance Admin and Table Admin services. Note that `TableAdmin` requires an `AsyncCheckConsistency` function for backwards compatibility.

All `.cc` and `.h` code is generated, except for `retry_traits.h` (which is needed for `publish-docs-pr`).

The admin code is not built at this time. A future PR will actually build the new APIs.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7550)


generator/generator_config.textproto
```diff
@@ -21,6 +21,22 @@ service {
...
+  initial_copyright_year: "2021"
+  emulator_endpoint_env_var: "BIGTABLE_INSTANCE_ADMIN_EMULATOR_HOST"
+}
+
+service {
+  service_proto_path: "google/bigtable/admin/v2/bigtable_table_admin.proto"
+  product_path: "google/cloud/bigtable/admin"
+  initial_copyright_year: "2021"
+  emulator_endpoint_env_var: "BIGTABLE_EMULATOR_HOST"
+  gen_async_rpcs: ["CheckConsistency"]
```
coryan, line 37: Consider removing the "snapshot" APIs:

https://github.com/googleapis/googleapis/blob/bcfd386c224fded5f5c5be7a17bc5884a2a52f59/google/bigtable/admin/v2/bigtable_table_admin.proto#L70-L74

We do not support them in the current APIs, and have no plans to (these APIs never left "private preview").

coryan (changes requested):

generator/generator_config.textproto
dbolduc: Ah, thanks for pointing this out.

coryan (approved):

PR #7554, opened 2021-11-02
ci: build generated bigtable admin APIs

dbolduc:
Part of the work for #7526 

This change allows the new APIs to be built with bazel, cmake.

This change does not add any new build targets. The new admin APIs are built as a part of `google_cloud_cpp_bigtable`

The mocks will be added to `CMakeLists.txt` in a later PR where they are actually used.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7554)


google/cloud/bigtable/CMakeLists.txt
```diff
@@ -50,6 +43,45 @@ include(CreateBazelConfig)
 # the client library
 add_library(
     google_cloud_cpp_bigtable # cmake-format: sort
+    admin/bigtable_instance_admin_client.cc
```
coryan, line 46: Can you add this new directory (or at least the generated files) to the list of ignored files for code coverage purposes:

https://github.com/googleapis/google-cloud-cpp/blob/ffbe748b73a016e864f9a2b7bff371759adf70c5/.codecov.yml#L1

https://docs.codecov.com/docs/ignoring-paths

coryan (approved):
If the plan is to add tests later, then please merge as-is.

If the plan is to not have tests for these files, let's change `.codecov.yml` to ignore them.

google/cloud/bigtable/CMakeLists.txt
dbolduc: The plan is to add integration tests for the admin APIs. I don't think that will get done this week due to the summit, so I will have ignore the files from coverage for now.

PR #7571, opened 2021-11-09
test(bigtable): convert admin integration tests to use generated code

### LINKED ISSUE #7526: Generate Bigtable Admin APIs

dbolduc:
Fixes #7526 

Most of the changes in here involve methods that go from accepting an ID (e.g. `table-id` ) to accepting a full name (e.g. `projects/project-id/instances/instance-id/tables/table-id`)

There is also a change to the `WaitForConsistency` test in `table_admin_integration_test.cc`. Our generated clients do not have a method like `TableAdmin::WaitForConsistency`, so I implemented a simple version in a retry loop. I did not create an asynchronous retry loop, making asynchronous calls. It seemed too complicated, and I did not want to write a test to test the test.

Note that `bigtable/testing/table_integration_test.*` still uses the old APIs. That will be updated to the generated APIs in a future PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7571)

### DIFF
--- a/google/cloud/bigtable/admin/integration_tests/admin_backup_integration_test.cc
+++ b/google/cloud/bigtable/admin/integration_tests/admin_backup_integration_test.cc
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "google/cloud/bigtable/instance_admin.h"
+#include "google/cloud/bigtable/admin/bigtable_instance_admin_client.h"
+#include "google/cloud/bigtable/admin/bigtable_table_admin_client.h"
 #include "google/cloud/bigtable/testing/table_integration_test.h"
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/internal/random.h"
@@ -28,7 +29,7 @@
 
 namespace google {
 namespace cloud {
-namespace bigtable {
+namespace bigtable_admin {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace {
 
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/admin/integration_tests/table_admin_integration_test.cc
```diff
@@ -349,15 +451,29 @@ TEST_F(TableAdminIntegrationTest, CreateListGetDeleteTableWithLogging) {
...
-  std::vector<bigtable::ColumnFamilyModification> column_modification_list = {
-      bigtable::ColumnFamilyModification::Create(
-          "newfam", GC::Intersection(GC::MaxAge(std::chrono::hours(7 * 24)),
-                                     GC::MaxNumVersions(1))),
-      bigtable::ColumnFamilyModification::Update("fam", GC::MaxNumVersions(2)),
-      bigtable::ColumnFamilyModification::Drop("foo")};
-
-  auto table_modified =
-      table_admin->ModifyColumnFamilies(table_id, column_modification_list);
+  using Mod = btadmin::ModifyColumnFamiliesRequest::Modification;
```
coryan, line 454: nit: `using btadmin::ModifyColumnFamiliesRequest::Modification;` seems like would work?

google/cloud/bigtable/admin/integration_tests/table_admin_integration_test.cc
```diff
@@ -274,64 +335,105 @@ TEST_F(TableAdminIntegrationTest, WaitForConsistencyCheck) {
...
 
   // Create a consistency token after modifying the table.
+  auto consistency_token_resp = client_.GenerateConsistencyToken(table_name);
+  ASSERT_STATUS_OK(consistency_token_resp);
   auto consistency_token =
-      table_admin.GenerateConsistencyToken(random_table_id);
-  ASSERT_STATUS_OK(consistency_token);
+      std::move(*consistency_token_resp->mutable_consistency_token());
+
+  // A retry loop that checks if the data replication has completed.
```
coryan, line 343: For a future PR: this is the kind of stuff that, while not hard to implement, it is tedious for customers to implement.  Maybe we should have a standalone function?  Something like:

```cc
namespace google::cloud::bigtable_admin {
future<Status> AsyncWaitForConsistency(BlahClient client, std::string consistency_token);
future<Status> AsyncWaitForConsistency(BlahClient client, std::string consistency_token, RetryPolicy retry, BackoffPolicy backoff);
}
```

google/cloud/bigtable/admin/integration_tests/admin_backup_integration_test.cc
```diff
@@ -37,102 +38,129 @@ using ::google::protobuf::util::TimeUtil;
 using ::testing::Contains;
 using ::testing::Not;
 namespace btadmin = ::google::bigtable::admin::v2;
-namespace bigtable = ::google::cloud::bigtable;
 
 class AdminBackupIntegrationTest
     : public bigtable::testing::TableIntegrationTest {
```
coryan, line 43: I think this means you are still using the hand-crafted clients for the base class `SetUp()` and cleanup. Consider a future PR to switch the base classes to use the new admin APIs.

google/cloud/bigtable/admin/integration_tests/table_admin_integration_test.cc
coryan: And FWIW, there is a lot of untested / unused code in your sample loop.

coryan (approved):

google/cloud/bigtable/admin/integration_tests/admin_backup_integration_test.cc
dbolduc: Yes. The plan is to convert the `TableIntegrationTest` implementation in the next PR as part of the "prefer generated APIs" issue. I edited the PR description to reflect that, but probably after you had looked at the code.

google/cloud/bigtable/admin/integration_tests/table_admin_integration_test.cc
dbolduc: Apparently no, because `Modification` is not a class. The generated proto code is:
```c++
typedef ModifyColumnFamiliesRequest_Modification Modification;
```
I could save a line (in two places) by just saying:
```c++
std::vector<btadmin::ModifyColumnFamiliesRequest::Modification> mods(3)
```
... but I think that's a bit harder to read

google/cloud/bigtable/admin/integration_tests/table_admin_integration_test.cc
dbolduc: Ack and I agree. We will keep `TableAdmin::WaitForConsistency()` for as long as we keep `TableAdmin`... but if we are ever going to deprecate `TableAdmin`, we need a useful alternative (like the one you propose).

It is true that the loop is not tested. Maybe it is already too complicated. But I am going to leave it as is. I think tests for all the behaviors of such a loop should be done in unit tests with mocks.

I wouldn't say there is "unused" code. If the loop fails for whatever reason, we want to have a more informative error as to why.

PR #7579, opened 2021-11-10
test: enable codecov for bigtable/admin

dbolduc:
Unsurprisingly, I forgot to re-enable codecov for `bigtable/admin/` after the integration tests were converted.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7579)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,6 +4,3 @@ coverage:
   status:
     project: off
     patch: off
-
-ignore:
-  - "google/cloud/bigtable/admin/*"


coryan:
Do we want this coverage?  Are there RPC flavors not covered in other generated libraries that do have tests?

dbolduc:
Closing this PR as it will not help us identify code that we want to test but are not testing currently.

devbww:
> Do we want this coverage? Are there RPC flavors not covered in other generated libraries that do have tests?

Does that imply we should figure out if coverage can be ignored for other generated trees too?  Would we have to re-evaluate such decisions every time the service definitions change?  That is, would it be easier to just include everything?

PR #7585, opened 2021-11-12
ci: ignore generated code in coverage build

devbww:
Ignore directories containing generated code.  This also ignores
some hand-written pieces like `integration_tests` and `samples`,
but we're not particularly interested in coverage data for those
either.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7585)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,5 +5,15 @@ coverage:
     project: off
     patch: off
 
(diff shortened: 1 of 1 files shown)


coryan (approved):
Thanks, we can refine this later if we think hand-crafted examples matter (for example).

PR #7588, opened 2021-11-12
fix(spanner): avoid use-after-move bugs

devjgm:
@devbww and I debugged a use-after move issue that boiled down to these
two Row iterators inspecting a data member (e.g., `row_` and `tup_`) on
subsequent invocations of `op++`, after a caller may have rightfully
moved the object. To avoid this issue, we keep an extra bool in both
iterators to avoid the need to inspect the StatusOr data member again.

Note: This issue didn't show up in our code or builds. It showed up only
when trying to make a different change to `Status`.

Kudos to @devbww for helping to track this down.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7588)

### DIFF
--- a/google/cloud/spanner/row_test.cc
+++ b/google/cloud/spanner/row_test.cc
@@ -330,13 +330,17 @@ TEST(RowStreamIterator, MovedFromValueOk) {
   EXPECT_NE(it, end);
   auto row = std::move(*it);
   EXPECT_STATUS_OK(row);
-  EXPECT_EQ(1, row->size());
+  auto val = row->get("num");
+  EXPECT_STATUS_OK(val);
+  EXPECT_EQ(Value(1), *val);
 
   ++it;
   EXPECT_NE(it, end);
   row = std::move(*it);
   EXPECT_STATUS_OK(row);
-  EXPECT_EQ(1, row->size());
+  val = row->get("num");
+  EXPECT_STATUS_OK(val);
+  EXPECT_EQ(Value(2), *val);
 
   ++it;
   EXPECT_EQ(it, end);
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/row_test.cc
```diff
@@ -404,6 +428,32 @@ TEST(TupleStreamIterator, Error) {
   EXPECT_EQ(it, end);
 }
 
+TEST(TupleStreamIterator, MovedFromValueOk) {
+  std::vector<Row> rows;
+  rows.emplace_back(MakeTestRow({{"num", Value(42)}}));
+  rows.emplace_back(MakeTestRow({{"num", Value(42)}}));
```
devbww, line 434: Use a distinct value?

google/cloud/spanner/row_test.cc
```diff
@@ -318,6 +318,30 @@ TEST(RowStreamIterator, RangeForLoop) {
...
+  rows.emplace_back(MakeTestRow({{"num", Value(2)}}));
+
+  RowRange range(MakeRowStreamIteratorSource(rows));
+  auto it = range.begin();
+  auto end = range.end();
+
+  EXPECT_NE(it, end);
+  auto row = std::move(*it);
+  EXPECT_STATUS_OK(row);
+  EXPECT_EQ(1, row->size());
```
devbww, line 333: Look at the values?

google/cloud/spanner/row_test.cc
devjgm: Done.

google/cloud/spanner/row_test.cc
devjgm: Done.

devbww (approved):

Issue #7429, opened 2021-10-08
Improve error messages by supporting ErrorInfo and AIP-193


devjgm:
https://google.aip.dev/193 describes the recommended way for Google services to return additional error information to users. The additional data is sent in the https://github.com/googleapis/api-common-protos/blob/master/google/rpc/status.proto `details` field, which is a **repeated** `google.protobuf.Any`. This is exposed to in C++ via `grpc::Status`'s `.error_details()` method (https://grpc.github.io/grpc/cpp/classgrpc_1_1_status.html), which returns `std::string`. As of today, our [`google::cloud::Status`](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/status.h) class drops the error details on the ground.

So I think we need to do the following:
- [x] Attach "error_details" to [`google::cloud::Status`](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/status.h). This is covered by https://github.com/googleapis/google-cloud-cpp/issues/1912
- [ ] We need to provide some API that easily (or automatically?) extract and print [`ErrorInfo`](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L110) fields *IF* they are present in the Status's "error_details". We should consider options for how we want to do this. 
- [ ] We should change `op<<` for `g::c::Status` to log the extra "ErrorInfo" stuff by default

devjgm:
This is a useful link for where grpc does a lot of manipulation of its Status object using upb and absl in some places too: https://github.com/grpc/grpc/blob/master/src/core/lib/gprpp/status_helper.cc

devjgm:
I believe the error_details is set to the serialized `google::rpc::Status` from which it was created, and that would have an Any that may contain the `ErrorInfo`.

Related to: https://github.com/grpc/grpc/issues/24007

devjgm:
Some answers to questions that I had

* Any grpc Status object will have _at most_ on `ErrorInfo` object. Even though the underlying field is a repeated field of Anys, there will be a max of one ErrorInfo.
* We do need to give users programmatic access to the "Error Info" fields (reason, domain, metadata).
* We need to change `Status::op<<` to print the ErrorInfo details if its present.
* We cannot add a protobuf or grpc dep to `status.cc`, because that would suck for GCS users.
* Even JSON requests will gain an "ErrorInfo" (equivalent) JSON object

Issue #4612, opened 2020-07-21
Add a function to convert from `g::c::Status` -> `absl::Status`


devjgm:
Due to https://github.com/googleapis/google-cloud-cpp/issues/4375, we will *probably* not be adopting `absl::Status`, and instead we'll keep our custom `google::cloud::Status` class https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/status.h

It's likely that customers may want to convert from our Status to `absl::Status`, so we may want to provide a function like:

```cc
absl::Status ToAbslStatus(google::cloud::Status const& s);
```

**Q: Where should this function live?**

`google/cloud/status.h` is an obvious choice, but that would require that all of our customers would also get the Abseil dep, even when they don't need it (i.e., non-Google customers). Other thoughts:

* `google/cloud/absl_status.h`
* `google/cloud/absl_utils.h` -- Maybe there are more Abseil helpers we have or may need?
* ???

/cc @curiousg23

devjgm:
We now have a hard dep on Abseil, so I suspect that concern mentioned above is likely no longer an issue.

coryan:
While the dependency on Abseil is there, adding this function exposes `absl::Status` in our public API, maybe harmless, but we had decided to only use the `absl::` names that come from `std::` in our public API. we can change our minds, of course.    🤷

devjgm:
At this point, we've decided that we don't want to expose `absl::Status` in our headers (at least not publically). So we will not be adding these conversion functions in this repo. We'll work with folks who may need them to figure out a good alternative.

Issue #1912, opened 2019-01-31
Preserve `grpc::Status::error_details()`


devjgm:
[`grpc::Status`](https://grpc.io/grpc/cpp/classgrpc_1_1_status.html) has an `error_details()` member function, which contains some binary form of additional error information. I believe it's usually some serialized protobuf. So simply printing it is probably not the right answer.

At this point `google::cloud::Status` does not support additional payloads. If at some point we have a Status object that does support additional payloads, we may want to preserve the error details.

PR #7592, opened 2021-11-13
refactor(common): Status uses absl::Status

devjgm:
Related to: https://github.com/googleapis/google-cloud-cpp/issues/4612
Related to: https://github.com/googleapis/google-cloud-cpp/issues/1912

There is no user-facing API change in this PR. This change only
_implements_ `g::c::Status` in terms of `absl::Status`. This will give
us the ability to do the following:

1. Easily convert between our Status and Abseil's.
2. Preserve grpc's `error_details()`.

One downside of this PR is that we still need to keep our own `message_`
data member, which is a copy of the one in contained `absl::Status`,
becuase our `g::c::Status::message()` function returns a `std::string
const&`. I can see no other way to do this, and it is a con.

Note: `absl::Status` does not like use-after-move, so I needed to remove a
couple tests that were relying on unspecified behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7592)

### DIFF
--- a/google/cloud/BUILD
+++ b/google/cloud/BUILD
@@ -40,6 +40,7 @@ cc_library(
     hdrs = google_cloud_cpp_common_hdrs,
     deps = [
         "@com_google_absl//absl/memory",
+        "@com_google_absl//absl/status",
         "@com_google_absl//absl/time",
         "@com_google_absl//absl/types:optional",
         "@com_google_absl//absl/types:variant",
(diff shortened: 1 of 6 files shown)


coryan (changes requested):
My gut says no, but let's talk about it.

devjgm:
We're going to go in a different direction. Alternative PR coming up.

Issue #2344, opened 2019-03-26
Remove `grpc::Status` from `PollingPolicy`


dopiera:
`google::cloud::Status` should be used instead.

It's not feasible yet because there are places which use `grpc::Status`, but eventually they should go away after we migrate to futures.

coryan:
To what @dopiera said I would add: because at least in theory customers could add their own policies, we should consistently use one `*::Status` class.

coryan:
We will re-examine this bug as part of "Modernizing Bigtable"

PR #7594, opened 2021-11-13
cleanup(bigtable): add ability to convert to common policies

### LINKED ISSUE #7529: Normalize Bigtable Policies
Bigtable has its own version of retry, backoff, and polling policies. We need a way to convert them from `g::c::bigtable::RPCRetryPolicy` -> `g::c::internal::RetryPolicy`, for example.

**Why?** - To modernize Bigtable.
We want the connection layer to implement the retry loops. Generated connection layers can only use common policies. Also, we should still prefer that handwritten connection layers use common policies, to minimize code maintenance.

**Note to self:** the classes will probably look like [this](https://github.com/dbolduc/google-cloud-cpp/blob/a93d12b77eccbbfebd9916b1ab653038d70f1801/google/cloud/bigtable/table.h#L866-L916). But they definitely won't be defined in `table.h`. And I probably want to add a virtual `IsExhausted` method to `bigtable::RPCRetryPolicy` that returns `false` by default, but `impl_.IsExhausted()` where possible.

dbolduc:
Fixes #7529 

There is no `static_assert()` checking that the type  `T` passed to `MakeCommonRetryPolicy<T>();` is derived from `TraitBasedRetryPolicy<U>`. (I am not sure this is even possible). I think this is fine, because the function is not user facing.

I think it is best to leave the `grpc::Status`es in the tests, because that is the main way we pass errors to the policies. There is a separate issue to clean up these functions. (#2344)

This PR introduces `bigtable_internal`. I did this because it is easier to read `internal::Foo` than `google::cloud::internal::Foo`. I am allowed to convert everything else from `bigtable::internal` to `bigtable_internal` on Dec. 2nd.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7594)

### DIFF
--- a/google/cloud/bigtable/CMakeLists.txt
+++ b/google/cloud/bigtable/CMakeLists.txt
@@ -21,8 +21,8 @@ set(DOXYGEN_PROJECT_BRIEF "A C++ Client Library for Google Cloud Bigtable")
 set(DOXYGEN_PROJECT_NUMBER "${PROJECT_VERSION}")
 set(DOXYGEN_EXAMPLE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/examples"
                          "${CMAKE_CURRENT_SOURCE_DIR}/quickstart")
-set(DOXYGEN_EXCLUDE_SYMBOLS "benchmarks" "bigtable_admin_internal" "internal"
-                            "testing" "examples")
+set(DOXYGEN_EXCLUDE_SYMBOLS "benchmarks" "bigtable_admin_internal"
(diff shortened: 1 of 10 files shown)


dbolduc:
@coryan points out that I have dropped the `Setup(grpc::ClientContext&)` functionality for no good reason.

I am marking this as a draft until I add them back in.

dbolduc:
I thought adding `Setup` back in would take like 10 minutes. It seems much more involved. Closing for now.

PR #7596, opened 2021-11-14
cleanup: explicitly list Doxygen sources

coryan:
So far, we have been recursively scanning the source directory of each
library and then excluding some files. This is tedious, particularly in
the opt-level `google/cloud` directory. And also brittle, as adding a
new directory requires changes in multiple places.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7596)

### DIFF
--- a/google/cloud/iam/CMakeLists.txt
+++ b/google/cloud/iam/CMakeLists.txt
@@ -19,7 +19,6 @@ set(DOXYGEN_PROJECT_BRIEF "A C++ Client Library for Google Cloud IAM")
 set(DOXYGEN_PROJECT_NUMBER "${PROJECT_VERSION}")
 set(DOXYGEN_EXAMPLE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/samples
                          ${CMAKE_CURRENT_SOURCE_DIR}/quickstart)
-set(DOXYGEN_EXCLUDE_PATTERNS "*/*_test.cc")
 set(DOXYGEN_EXCLUDE_SYMBOLS "internal" "iam_internal" "iam_testing" "examples")
 
 # Creates the proto headers needed by doxygen.
--- a/google/cloud/logging/CMakeLists.txt
+++ b/google/cloud/logging/CMakeLists.txt
@@ -17,7 +17,6 @@
 set(DOXYGEN_PROJECT_NAME "Google Cloud Logging C++ Client")
 set(DOXYGEN_PROJECT_BRIEF "A C++ Client Library for Google Cloud Logging")
 set(DOXYGEN_PROJECT_NUMBER "${PROJECT_VERSION} (Beta)")
-set(DOXYGEN_EXCLUDE_PATTERNS "*/*_test.cc")
 set(DOXYGEN_EXCLUDE_SYMBOLS "internal" "logging_internal" "logging_testing"
                             "examples")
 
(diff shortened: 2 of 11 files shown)


google/cloud/iam/CMakeLists.txt
```diff
@@ -19,7 +19,6 @@ set(DOXYGEN_PROJECT_BRIEF "A C++ Client Library for Google Cloud IAM")
 set(DOXYGEN_PROJECT_NUMBER "${PROJECT_VERSION}")
 set(DOXYGEN_EXAMPLE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/samples
                          ${CMAKE_CURRENT_SOURCE_DIR}/quickstart)
-set(DOXYGEN_EXCLUDE_PATTERNS "*/*_test.cc")
 set(DOXYGEN_EXCLUDE_SYMBOLS "internal" "iam_internal" "iam_testing" "examples")
```
dbolduc, line 22: not introduced in this PR, but: s/examples/samples/

google/cloud/logging/CMakeLists.txt
```diff
@@ -17,7 +17,6 @@
 set(DOXYGEN_PROJECT_NAME "Google Cloud Logging C++ Client")
 set(DOXYGEN_PROJECT_BRIEF "A C++ Client Library for Google Cloud Logging")
 set(DOXYGEN_PROJECT_NUMBER "${PROJECT_VERSION} (Beta)")
-set(DOXYGEN_EXCLUDE_PATTERNS "*/*_test.cc")
 set(DOXYGEN_EXCLUDE_SYMBOLS "internal" "logging_internal" "logging_testing"
                             "examples")
```
dbolduc, line 21: logging has no `examples`

dbolduc (commented):
I am a little confused by when we want to exclude the samples/examples in the libraries. And whether we want to even modify the CMakeLists.txt for the generated libraries. I made comments where I saw discrepancies, but I might not be understanding what we want to do.

Also, what about BigQuery? https://github.com/googleapis/google-cloud-cpp/blob/cfa2fbe422f72fbe09ea19f884ec81befdc3c821/google/cloud/bigquery/CMakeLists.txt#L22

google/cloud/spanner/CMakeLists.txt
```diff
@@ -22,11 +22,7 @@ set(DOXYGEN_PROJECT_NUMBER "${PROJECT_VERSION}")
 set(DOXYGEN_EXAMPLE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/samples
                          ${CMAKE_CURRENT_SOURCE_DIR}/quickstart)
 set(DOXYGEN_EXCLUDE_SYMBOLS "internal" "spanner_admin_internal"
-                            "spanner_internal" "spanner_testing")
-set(DOXYGEN_EXCLUDE_PATTERNS
-    "*/google/cloud/spanner/README.md" "*/google/cloud/spanner/internal/*"
-    "*/google/cloud/spanner/benchmarks/*" "*/google/cloud/spanner/testing/*"
-    "*/google/cloud/spanner/*_test.cc")
+                            "spanner_internal" "spanner_testing" "examples")
```
dbolduc, line 25: s/examples/samples/ ?

google/cloud/iam/CMakeLists.txt
coryan: This is **symbols** not directories.  There are none (yet), but when we create helper classes in `google/cloud/$product/samples` they are typically called `google::cloud::$product::*examples*`.  And we (at least I) do not want these to be exported as part of the documentation of the library.

google/cloud/logging/CMakeLists.txt
coryan: I know.  I am trying to make all the CMake files look as-if they had been generated, or closer to it.

google/cloud/iam/CMakeLists.txt
dbolduc: Ok, thanks for clarifying the point of the changes.

I think [the documentation](https://www.doxygen.nl/manual/config.html#cfg_exclude_symbols) confused me. I guess the note about `*/test/*` only belongs in EXCLUDE_PATTERNS.

PR #7597, opened 2021-11-14
feat: automate expected directory list

coryan:
Automatically update the list of expected directories when we add a
generated library.


generator/standalone_main.cc
```diff
@@ -52,6 +53,69 @@ GetConfig(std::string const& filepath) {
...
+  auto const l = path.find_last_of('/');
+  if (l == std::string::npos) return path;
+  return path.substr(l + 1);
+}
+
+std::string Dirname(std::string const& path) {
+  auto const l = path.find_last_of('/');
+  if (l == std::string::npos) return path;
+  return path.substr(0, l);
+}
```
devbww, line 67: Nit: These may not matter in the particular use cases, but `Dirname("/foo") == ""` and `Dirname("foo") == "foo"` seem wrong without some note about the supported shape of `path`.

coryan: Done.

devbww: > The return value for ... paths without `/` is unspecified

Perhaps you should drop the test for `npos` then.  The above is equivalent to `return path.substr(0, path.find_last_of('/'));`.

coryan: Sure.

devjgm (approved):

PR #7599, opened 2021-11-15
feat: automatically update feature list

coryan:
When adding new generated libraries we need to update the feature list
used by the CI builds (currently just `cmake-install`). With this change
that update is automated, reducing the work for each library.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7599)

### DIFF
--- a/ci/cloudbuild/builds/cmake-install.sh
+++ b/ci/cloudbuild/builds/cmake-install.sh
@@ -27,12 +27,26 @@ export CXX=clang++
 INSTALL_PREFIX="$(mktemp -d)"
 readonly INSTALL_PREFIX
 
+mapfile -t feature_list < <(cat ci/etc/full_feature_list)
+# Add the hand-crafted libraries and custom features to the full list
+feature_list+=(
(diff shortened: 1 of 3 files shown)


ci/cloudbuild/builds/cmake-install.sh
```diff
@@ -27,12 +27,26 @@ export CXX=clang++
...
 
+mapfile -t feature_list < <(cat ci/etc/full_feature_list)
+# Add the hand-crafted libraries and custom features to the full list
+feature_list+=(
+  bigtable
+  pubsub
+  spanner
+  storage
+  experimental-storage-grpc
+  pubsublite
```
dbolduc, line 38: should `pubsublite` be in `ci/etc/full_feature_list` ?

coryan: Will be, once I add `pubsublite` to the `generator/generator_config.textproto` file.

dbolduc (approved):

PR #7600, opened 2021-11-15
feat: use cloud.google.com for stable doc links

coryan:
When linking to the documentation of the different protos use the
reference documentation in `https://cloud.google.com/$product/docs`.
This gives us stable links that do not change as we pin to newer
versions of the APIs.  The downside is that the proto description in
`cloud.google.com` may be newer than the version we used to generate the
code, specially for documents generated with older versions of the
library.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7600)


coryan:
Close for now, no consensus.

PR #7603, opened 2021-11-16
feat(common): add internal-only payload support to Status

devjgm:
Related to: #7429
Related to: #1912

This PR adds the ability for non-OK `g::c::Status` objects to hold arbitrary payloads. This does not add any new public API to Status; it is all internal. There's no behavior change in this PR, because nothing is currently setting the payloads. A following PR will start using the payload features.

This payload support will allow us to add and extract the `ErrorInfo` protobuf from Status objects that came from gRPC. We can also do the same for errors from REST services that have an "ErrorInfo" equivalent. We can also include extra stringified data to be logged when we print Status objects, without forcing a protobuf dep on the main `Status` class (which would be bad for gRPC. We would also be able to convert to/from `absl::Status` without losing any data if we choose to do so. All of this will be part of followup PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7603)

### DIFF
--- a/google/cloud/BUILD
+++ b/google/cloud/BUILD
@@ -40,6 +40,7 @@ cc_library(
     hdrs = google_cloud_cpp_common_hdrs,
     deps = [
         "@com_google_absl//absl/memory",
+        "@com_google_absl//absl/status",
         "@com_google_absl//absl/time",
         "@com_google_absl//absl/types:optional",
         "@com_google_absl//absl/types:variant",
--- a/google/cloud/BUILD
+++ b/google/cloud/BUILD
@@ -40,7 +40,6 @@ cc_library(
     hdrs = google_cloud_cpp_common_hdrs,
     deps = [
         "@com_google_absl//absl/memory",
-        "@com_google_absl//absl/status",
         "@com_google_absl//absl/time",
         "@com_google_absl//absl/types:optional",
         "@com_google_absl//absl/types:variant",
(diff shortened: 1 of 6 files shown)


coryan (approved):
LGTM.

I think you could hide the `GetPayload()` helper if you added a `google/cloud/internal/status_impl.h` file, which would only be included in the `.cc` files that needed it.  But I am just being nit picky.

devjgm:
> LGTM.
> 
> I think you could hide the `GetPayload()` helper if you added a `google/cloud/internal/status_impl.h` file, which would only be included in the `.cc` files that needed it. But I am just being nit picky.

That initially seemed like overkill to me. But now that I think about the next PRs, we probably need somewhere to define the keys that use for various payloads. That would pollute status.h more than I'd like. So I may see about changing this to be implemented in a `google/cloud/internal/status_payload.h` file, which could also hold whatever payload keys we want to define for ourselves.

devjgm:
> > LGTM.
> > I think you could hide the `GetPayload()` helper if you added a `google/cloud/internal/status_impl.h` file, which would only be included in the `.cc` files that needed it. But I am just being nit picky.
> 
> That initially seemed like overkill to me. But now that I think about the next PRs, we probably need somewhere to define the keys that use for various payloads. That would pollute status.h more than I'd like. So I may see about changing this to be implemented in a `google/cloud/internal/status_payload.h` file, which could also hold whatever payload keys we want to define for ourselves.

On further thought, I think I'm going to leave this as-is. We can easily put keys in a file like `google/cloud/internal/status_payload_keys.h` or something to keep them out of the main `status.h`. If that doesn't work, we can also easily refactor later since this is all internal.

PR #7606, opened 2021-11-16
feat(generator): more stable googleapis proto links

coryan:
We want to minimize the number of changes when we change the commit SHA
for googleapis/googleapis.  With this change we use a Doxygen custom
command in the header files. The CMake file defines this command to
include the embedded SHA. The arguments to the command *should* be more
stable, but will change when proto definitions are moved in their
`.proto` file.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7606)


scotthart (approved):
> Reviewed 32 of 32 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7606)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #7620, opened 2021-11-17
refactor(common): attach Status proto as g::c::Status payload when available

devjgm:
Fixes: #1912

This PR attaches serialized `google.rpc.Status` protos to our `g::c::Status` protobuf messages when it's available. This is normally sent to us from gRPC as the value of its `grpc::Status::error_details()` field.

There's no new public API in this PR. There should also be no new behavior change, because nothing is yet looking at these payload fields (other than `op==`). In following PRs, I'll add a new API to extract the `ErrorInfo`, and to log the `ErrorInfo`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7620)


google/cloud/grpc_error_delegate.cc
```diff
@@ -62,25 +63,34 @@ google::cloud::StatusCode MapStatusCode(grpc::StatusCode const& code) {
 }  // namespace
 
 google::cloud::Status MakeStatusFromRpcError(grpc::Status const& status) {
+  auto e = status.error_details();
+  if (!e.empty()) {
+    google::rpc::Status proto;
+    if (!proto.ParseFromString(e)) {
+      return google::cloud::Status(StatusCode::kInternal,
```
coryan, line 70: Is that what we want?  We would be dropping the error message reported by the service in favor of the `kInternal`?  I think maybe you want to return a status with the original `status.error_code()` and `status.error_message()` and make the error details say "we got some error details but they were in an invalid format, so we had to discard them".

google/cloud/grpc_error_delegate.h
```diff
@@ -42,7 +43,7 @@ google::cloud::Status MakeStatusFromRpcError(grpc::StatusCode code,
  * libraries in `google-cloud-cpp` represent these errors using a
  * `google::cloud::Status`.
  */
-google::cloud::Status MakeStatusFromRpcError(google::rpc::Status const& status);
+google::cloud::Status MakeStatusFromRpcError(google::rpc::Status const& proto);
```
coryan, line 46: This becomes a public API, I think I made a mistake when I made `MakeStatusFromRpcError()` public, can we avoid adding the overload to the public API too?

coryan (changes requested):
Thanks, this is pretty clean modulo some comments.

google/cloud/grpc_error_delegate.cc
devjgm: Your suggestion is probably better. Changed. The reason I originally used `kInternal` was because this situation indicates a bad `grpc::Status` object, so it's unclear if its code and message are to be trusted either.

google/cloud/grpc_error_delegate.h
devjgm: This was already a public API. I just renamed the param to `proto`.

devjgm (commented):
PTAL

google/cloud/grpc_error_delegate.h
coryan: Doh!

coryan (approved):

PR #7622, opened 2021-11-17
refactor(common): Status ignores message if code==OK

devjgm:
With this PR, `google::cloud::Status` now ignores the `message` it's constructed with if the `code` is OK. This behavior matches `absl::Status` (which we don't use, but FWIW). We believe the previous behavior where the message was _not_ dropped on OK statuses was a bug, because that can result in "OK" Statuses being not equal. Consider:

```cc
Status ok1{StatusCode::kOk, ""};
Status ok2{StatusCode::kOk, "blah"};
assert(ok1.ok() && ok2.ok());
assert(ok1 != ok2);  // OLD BEHAVIOR -- CONSIDERED A BUG
assert(ok1 == ok2);  // NEW BEHAVIOR -- ALL OKs ARE EQUIVALENT
```
This new behavior differs from `grpc::Status`, which _does_ allow message strings on an OK status. But I imagine this is an oversight due to how it was implemented, and not actually desired behavior.

Requiring that all OK statuses are equivalent enables certain simplifications and optimizations, which we would like to enable.

Note: We (and Abseil) already disallow Status "payloads" on OK statuses.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7622)

### DIFF
--- a/google/cloud/status.h
+++ b/google/cloud/status.h
@@ -89,9 +89,8 @@ class Status {
    * Ignores @p message if @p code is `StatusCode::kOk`.
    */
   explicit Status(StatusCode code, std::string message)
-      : code_(code), message_(std::move(message)) {
-    if (ok()) message_.clear();
-  }
+      : code_(code),
+        message_(code_ == StatusCode::kOk ? "" : std::move(message)) {}
 
   bool ok() const { return code_ == StatusCode::kOk; }
   StatusCode code() const { return code_; }
--- a/google/cloud/status.h
+++ b/google/cloud/status.h
@@ -68,11 +68,12 @@ absl::optional<std::string> GetPayload(Status const&, std::string const& key);
 
  * This class is typically used to indicate whether or not a function or other
  * operation completed successfully. Success is indicated by an "OK" status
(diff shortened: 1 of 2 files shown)


dbolduc (approved):

google/cloud/status.h
```diff
@@ -82,8 +83,15 @@ class Status {
...
-  explicit Status(StatusCode status_code, std::string message)
-      : code_(status_code), message_(std::move(message)) {}
+  /**
+   * Constructs a Status with the given @p code and @p message.
+   *
+   * Ignores @p message if @p code is `StatusCode::kOk`.
+   */
+  explicit Status(StatusCode code, std::string message)
+      : code_(code), message_(std::move(message)) {
+    if (ok()) message_.clear();
```
coryan, line 93: nit:

```cc
  explicit Status(StatusCode code, std::string message)
      : code_(code), message_(ok() ? std::move(message) : std::string{}) {
```

https://godbolt.org/z/WjKdcW7sE

devjgm: fixed

PR #7624, opened 2021-11-18
feat(common): add public g::c::GetStatusDetailProto<T>(status) API

devjgm:
Part of: https://github.com/googleapis/google-cloud-cpp/issues/7429

This PR introduces a new user-facing API to allow users to access "error details" protos, which are defined in https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto, that may be attached to `google::cloud::Status` objects. Since these are protobufs, they need to to be declared in a file that can include and link with protobufs, which `status.h` *cannot* do since GCS does not link protobuf.

The new API is defined in `google/cloud/status_detail_proto.h`. The public API consists of one function template:

```cc
template <typename T>
absl::optional<T> GetStatusDetailProto(Status const& s);
```
Users would use it like:

```cc
#include "google/cloud/status_details.h"

google::cloud::Status status = ...
auto ei = google::cloud::GetStatusDetailProto<google::rpc::ErrorInfo>(status);
if (ei) {
  // ei->reason() ...
}
```

This PR also defines a non-public part of the API in the `google::cloud::internal` namespace. This function template will be used in a follow up PR, which is why it's factored out into a separate helper function.

Since this is introducing a new *public* API, feel free to nitpick the naming, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7624)

### DIFF
--- a/google/cloud/BUILD
+++ b/google/cloud/BUILD
@@ -98,6 +98,7 @@ cc_library(
         "@com_google_googleapis//:googleapis_system_includes",
         "@com_google_googleapis//google/iam/credentials/v1:credentials_cc_grpc",
         "@com_google_googleapis//google/longrunning:longrunning_cc_grpc",
+        "@com_google_googleapis//google/rpc:error_details_cc_proto",
         "@com_google_googleapis//google/rpc:status_cc_proto",
     ],
 )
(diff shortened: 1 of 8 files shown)


devjgm:
Changed this PR back to a draft. I want to chat about the design of the public API before committing this, but I'm OOO today so this will have to wait until tomorrow.

PR #7628, opened 2021-11-18
refactor: g::c::Status now logs code first then message

devjgm:
Upcoming we are going to implement a change where we sometimes log
additional "error info" data when streaming a Status object. This
additional data will have newlines in it, and it would seem weird if
status code came last one the final line of the logged message. It
seems more natural for the streamed output to start with the code,
followed by the message, and any additional data. This format also
matches `absl::Status`, fwiw.

I'm not picky about the specific format syntax. I chose something like
`UNKNOWN: some message`, but I'd be just as happy to go back to using
square brackes around the code, with or without a colon.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7628)


coryan (approved):
I changed the PR description: s/longs/logs/ please do the same in the commit message.

coryan (approved):

PR #7633, opened 2021-11-18
feat(storage): default payload format for CreateNotification

### LINKED ISSUE #1098: Make format optional in CreateNotification()
The API requires the parameter, but we can default to `JSON_API_V1` if the application does not specify a format.

/cc: @frankyn who suggested this improvement.

coryan:
Fixes #1098 

I tried to use `@copydoc` but it does not render correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7633)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -3027,6 +3027,44 @@ class Client {
    * @param topic_name the Google Cloud Pub/Sub topic that will receive the
    *     notifications. This requires the full name of the topic, i.e.:
    *     `projects/<PROJECT_ID>/topics/<TOPIC_ID>`.
+   * @param metadata define any optional parameters for the notification, such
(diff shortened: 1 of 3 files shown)


google/cloud/storage/client.h
```diff
@@ -3016,6 +3016,44 @@ class Client {
     return std::move(result).value().items;
   }
 
+  /**
+   * Creates a new notification config for a Bucket.
+   *
+   * Cloud Pub/Sub Notifications sends information about changes to objects
```
devbww, line 3022: s/sends/send/

devbww (approved):

google/cloud/storage/client.h
coryan: Fixed, thanks

PR #7639, opened 2021-11-19
refactor: optimize Status moving error data to heap-allocated Impl

devjgm:

Background: "OK" statuses are the common case, and they don't have message strings,
payload data, etc. Only non-OK Statuses need this extra data, so paying
the cost of these fields for OK statuses is wasteful.

This PR refactors `google::cloud::Status` to hold an opaque `Impl*` that
holds all the error data for non-OK statuses. OK statuses are
represented by `impl_ == nullptr`, which requires no allocations (again,
this is the common case). This allows us to associate whatever data we
need with non-OK Statuses without hurting the OK case.

**What's the moved-from state?** We continue to not define the moved-from state for
`Status`. It's a valid state, but unspecified. However, the implementation in this PR
is such that **all moved-from `Status` objects are "OK"**. This matches the default state
when a `Status{}` is constructed. Note that this is a **change in behavior** from the code
previously whereby a moved-from Status would retain its `code()` (and therefore `ok()`) state,
while losing its payload and message. 

Note: We will be attaching more "ErrorInfo" data with non-OK statuses in
an upcoming PR, so this refactoring will help with that.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7639)

### DIFF
--- a/google/cloud/status.cc
+++ b/google/cloud/status.cc
@@ -26,26 +26,6 @@ std::string StatusWhat(Status const& status) {
 }
 }  // namespace
 
-namespace internal {
-
-// Sets the given `payload`, indexed by the given `key`, on the given `Status`,
-// IFF the status is not OK. Payloads are considered in equality comparisons.
-// The keyspace used here is separate from other keyspaces (e.g.,
-// `absl::Status`), so we only need to coordinate keys with ourselves.
-void SetPayload(Status& s, std::string key, std::string payload) {
-  if (!s.ok()) s.payload_[std::move(key)] = std::move(payload);
-}
-
-// Returns the payload associated with the given `key`, if available.
-absl::optional<std::string> GetPayload(Status const& s,
-                                       std::string const& key) {
-  auto it = s.payload_.find(key);
-  if (it == s.payload_.end()) return absl::nullopt;
-  return it->second;
-}
-
-}  // namespace internal
-
 std::string StatusCodeToString(StatusCode code) {
   switch (code) {
     case StatusCode::kOk:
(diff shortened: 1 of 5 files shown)


google/cloud/status.h
```diff
@@ -82,42 +82,40 @@ absl::optional<std::string> GetPayload(Status const&, std::string const& key);
...
-    if (s.ok()) return os << s.code();
-    return os << s.code() << ": " << s.message();
-  }
+  explicit Status(StatusCode code, std::string message);
+
+  bool ok() const { return impl_ == nullptr; }
+  StatusCode code() const;
+  std::string const& message() const;
+
+  friend bool operator==(Status const& a, Status const& b);
```
coryan, line 102: nit: you can optimize (if function call overhead still matters these days) the OK case with something like:

```
friend bool operator==(Status const& a, Status const& b) {
  return (a.ok() && b.ok()) || a.EqualsTo(b);
}
friend bool operator!=(Status const& a, Status const& b) { return !(a == b); }
```

google/cloud/status.h
```diff
@@ -82,42 +82,40 @@ absl::optional<std::string> GetPayload(Status const&, std::string const& key);
...
   friend void internal::SetPayload(Status&, std::string, std::string);
   friend absl::optional<std::string> internal::GetPayload(Status const&,
                                                           std::string const&);
 
-  StatusCode code_{StatusCode::kOk};
-  std::string message_;
-  std::unordered_map<std::string, std::string> payload_;
+  class Impl;
+  // A null `impl_` is an OK status. Only non-OK Statuses allocate an Impl.
+  Impl* impl_{nullptr};
```
coryan, line 113: Why not use `std::unique_ptr<>` and keep `~Status()` in the `.cc` file?

google/cloud/status.cc
```diff
@@ -91,6 +71,119 @@ std::ostream& operator<<(std::ostream& os, StatusCode code) {
...
+
+// Deep copy
+Status::Status(Status const& other)
+    : impl_(other.ok() ? nullptr : new auto(*other.impl_)) {}
+
+// Deep copy
+Status& Status::operator=(Status const& other) {
+  if (this != &other) {
+    delete impl_;
+    impl_ = other.ok() ? nullptr : new auto(*other.impl_);
```
coryan, line 120: this is not exception-safe.  Should `auto(*other.impl_)` throw (or `new auto(...)` throw) the current object was already modified.  That is why we do the copy + swap stuff.

google/cloud/status_or.h
```diff
@@ -98,14 +98,22 @@ class StatusOr final {
...
   // NOLINTNEXTLINE(performance-noexcept-move-constructor)
-  StatusOr(StatusOr&&) = default;
+  StatusOr(StatusOr&& other)
+      : status_(std::move(other.status_)), value_(std::move(other.value_)) {
+    other.status_ = MakeDefaultStatus();
+  }
   // NOLINTNEXTLINE(performance-noexcept-move-constructor)
-  StatusOr& operator=(StatusOr&&) = default;
+  StatusOr& operator=(StatusOr&& other) {
+    status_ = std::move(other.status_);
```
coryan, line 112: Any reason to avoid the "move-constructor + swap idiom"?

coryan (commented):
I think the exception-safe idioms are actually easier to implement than the non-exception safe approach :shrug:

A few other thoughts for you to consider.

google/cloud/status.h
devjgm: std::unique_ptr requires a `Impl` to be fully defined, which would mean the definition of Impl needs to be pulled out into som other header. `std::shared_ptr` allow (I think) a forward decl of `Impl` only, so it would work, but shared_ptr implements "sharing" semantics which we don't need. But I think I could switch to shared_ptr still w/ a custom implementation of the copy c'tor and assignment to implement deep copy.

google/cloud/status.h
devjgm: Done.

google/cloud/status.h
devjgm: @coryan pointed out that I was mistaken, and `std::unique_ptr` does work here. Changed. Looks better to me.

google/cloud/status.cc
devjgm: Better with unique_ptr?

google/cloud/status_or.h
devjgm: now using `=default` move operators.

coryan (approved):

google/cloud/status.cc
coryan: Yes, thanks!

google/cloud/status.cc
```diff
@@ -91,6 +71,103 @@ std::ostream& operator<<(std::ostream& os, StatusCode code) {
   return os << StatusCodeToString(code);
 }
 
+// Encapsulates the implementation of a non-OK status. OK Statuses are
+// represented by a nullptr Status::impl_, as an optimization for the common
+// case of OK Statuses. This class holds all the data associated with a non-OK
+// Status so we don't have to worry about bloating the common OK case.
+class Status::Impl {
+  using PayloadType = std::unordered_map<std::string, std::string>;
```
devbww, line 79: Explicit access specifier?

devjgm: Actually, I wanted it to be private. It's only there to save some typing and readability in the body of `Impl`. I didn't intend to expose it to others as a usable member alias.

PR #7641, opened 2021-11-20
feat(generator): reduce required editing on scaffold

coryan:
The generated `CMakeLists.txt` files can use `file(GLOB)` to eliminate
the need to edit most generated files. We still can edit these files, of
course, but now it is not required in most cases.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7641)

### DIFF
--- a/doc/adding_new_libraries.md
+++ b/doc/adding_new_libraries.md
@@ -78,17 +78,30 @@ bazel run \
 ci/cloudbuild/build.sh -t checkers-pr
 ```
 
-## Manually add the C++ files to the CMakeLists.txt file
+## Create any custom source files
 
 If the `generator/generator_config.textproto` entry for the service does not
 enumerate the `retryable_status_codes`, you need to manually create the file as
-`google/cloud/$library/internal/<service_name>_retry_traits.h`. Add this file
-**and** any generated C++ files to the `CMakeLists.txt` file. There should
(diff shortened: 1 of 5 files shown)


doc/adding_new_libraries.md
```diff
@@ -78,18 +78,31 @@ bazel run \
...
-`google/cloud/$library/internal/<service_name>_retry_traits.h`. Add this file
-**and** any generated C++ files to the `CMakeLists.txt` file. There should
-be markers in the file, search for `EDIT HERE`.
+`google/cloud/$library/internal/<service_name>_retry_traits.h`.
+
+Likewise, for services using streaming operations you may need to implement the
+streaming `*Updater` function. Use `google/cloud/bigquery/streaming.cc` for
+inspiration.
+
+## Validate and Generate the `*.bzl` files
```
devjgm, line 91: I believe bazel can use file globs also. It's not a "recommended" approach for bazel, but here we're talking about generated libraries, trying to reduce manual steps, and we're now using globs in cmake also. Something to consider.

devjgm (approved):

doc/adding_new_libraries.md
coryan: Ack.  I am not sure which approach would be easier to import into google3. #7642

PR #7648, opened 2021-11-21
ci: parallelize CMake quickstart builds

coryan:
Each quickstart build compiles a single file with CMake, running them
sequentially is wasteful, and will not scale if we add quickstarts for
generated libraries.

We could try something similar for quickstart + Bazel.  My first attempts
did not work well as each `bazel build` invocation compiles all the
dependencies (Abseil, gRPC, protobuf) in parallel.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7648)

### DIFF
--- a/ci/verify_quickstart/CMakeLists.txt
+++ b/ci/verify_quickstart/CMakeLists.txt
@@ -1,0 +1,66 @@
+# ~~~
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


ci/verify_quickstart/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ~~~
+
+cmake_minimum_required(VERSION 3.14)
+project(verify-quickstart CXX)
+
+set(LIBRARIES bigquery bigtable iam pubsub spanner storage)
```
devjgm, line 20: This is another place where we have a list of libraries. Will this list need to be updated as we add new libraries? If so, would it be possible to have the cmake invocation command pass in the list of libraries with a `-D` flag or something?

devjgm (approved):

ci/verify_quickstart/CMakeLists.txt
coryan: Fixed

devjgm (approved):

PR #7650, opened 2021-11-22
cleanup: use script to update external/googleapis

coryan:
Use a script to update the list of `.proto` files use by each library,
as well as the list of libraries each library depends on.

The more complicated case was the `google-cloud-cpp::logging_type_protos`
library, its name does not match the Bazel rules where it would generate
`*logging_type_type_protos`. I introduced an `INTERFACE` library for
backwards compatibility.

In the bigquery library we are exposing both the `v1beta1/` and `v1/`
versions of `connection.proto`. I will figure out how to clean that up
in a future PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7650)


scotthart (approved):
> Reviewed 25 of 26 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7650)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

Issue #7644, opened 2021-11-20
Add ErrorInfo support to GCS


devjgm:
This is a follow on issue related to https://github.com/googleapis/google-cloud-cpp/issues/7429. In that issue, we added ErrorInfo support to `g::c::Status`, per https://google.aip.dev/193. That's now done for gRPC based services.

GCS is not gRPC, and so it needs the ability to fill in the ErrorInfo class that's (optionally) passed to `g::c::Status`. Per @coryan 's comment in https://github.com/googleapis/google-cloud-cpp/pull/7640

> For a future PR: I think we should do the same thing for GCS, at least here:
https://github.com/googleapis/google-cloud-cpp/blob/75d5b339f6f3efc3eb5ef74c0dc88eb094bdf81d/google/cloud/storage/internal/http_response.cc#L25

> Something like:

```cc
Status AsStatus(HttpResponse const& http_response) {
  if (HttpStatusCode::kMinContinue <= http_response.status_code &&
      http_response.status_code < HttpStatusCode::kMinSuccess) {
    // We treat the 100s (e.g. 100 Continue) as OK results. They normally are
    // ignored by libcurl, so we do not really expect to see them.
    return Status{};
  }
  if (HttpStatusCode::kMinSuccess <= http_response.status_code &&
      http_response.status_code < HttpStatusCode::kMinRedirects) {
    // We treat the 200s as Okay results.
    return Status{};
  }
  // Parse the payload as a JSON object, see https://cloud.google.com/apis/design/errors#http_mapping
  auto parsed = nlohmann::json::parse(http_response.payload, nullptr, false);
  ErrorInfo error_info = ExtractErrorInfo(parsed);
  return Status(/* map HTTP code */, parsed.value("message", ""), error_info);
```

devjgm:
From https://cloud.google.com/apis/design/errors#http_mapping, an example response with "error info" data is:

```json
{
  "error": {
    "code": 400,
    "message": "API key not valid. Please pass a valid API key.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "API_KEY_INVALID",
        "domain": "googleapis.com",
        "metadata": {
          "service": "translate.googleapis.com"
        }
      }
    ]
  }
}
```

devjgm:
[...]

PR #7654, opened 2021-11-23
feat: GCS parse "error info" from JSON when available

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/7644

Follows https://cloud.google.com/apis/design/errors#http_mapping and parses the "error info" json object from responses when it's available, and uses it to set the fields of the returned `g::c::Status`.

**Note to reviewers**: I'm new to using nlohmann::json, so please review that code closely and tell me what I'm doing wrong. I may be doing things very inefficiently or not following the right best practices. Please let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7654)

### DIFF
--- a/google/cloud/storage/internal/http_response.cc
+++ b/google/cloud/storage/internal/http_response.cc
@@ -146,7 +146,8 @@ StatusCode MapHttpCodeToStatus(long code) {  // NOLINT(google-runtime-int)
 //   ]
 // See also https://cloud.google.com/apis/design/errors#http_mapping
 ErrorInfo MakeErrorInfo(nlohmann::json const& details) {
-  static auto const kErrorInfoType = "type.googleapis.com/google.rpc.ErrorInfo";
+  static auto const* const kErrorInfoType =
+      "type.googleapis.com/google.rpc.ErrorInfo";
   for (auto const& e : details.items()) {
     auto const& v = e.value();
     if (v.value("@type", "") != kErrorInfoType) continue;
--- a/google/cloud/storage/internal/http_response.cc
+++ b/google/cloud/storage/internal/http_response.cc
@@ -146,7 +146,7 @@ StatusCode MapHttpCodeToStatus(long code) {  // NOLINT(google-runtime-int)
 //   ]
 // See also https://cloud.google.com/apis/design/errors#http_mapping
 ErrorInfo MakeErrorInfo(nlohmann::json const& details) {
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/http_response.cc
```diff
@@ -50,87 +50,147 @@ Status AsStatus(HttpResponse const& http_response) {
...
+//       "domain": "..."
+//       "metdata": {
+//         "key1": "value1"
+//         ...
+//       }
+//     }
+//   ]
+// See also https://cloud.google.com/apis/design/errors#http_mapping
+ErrorInfo MakeErrorInfo(nlohmann::json const& details) {
+  static auto const* const kErrorInfoType =
```
coryan, line 149: nit: `auto constexpr kErrorInfoType = "...";` seems like enough?

google/cloud/storage/internal/http_response.cc
```diff
@@ -50,87 +50,147 @@ Status AsStatus(HttpResponse const& http_response) {
...
+      "type.googleapis.com/google.rpc.ErrorInfo";
+  for (auto const& e : details.items()) {
+    auto const& v = e.value();
+    if (v.value("@type", "") != kErrorInfoType) continue;
+    auto reason = v.value("reason", "");
+    auto domain = v.value("domain", "");
+    auto metadata_json = v.value("metadata", nlohmann::json::object());
```
coryan, line 156: nit: I do not worry about copying here, this should be the uncommon path.  But if one was worried about copying, there is a copy of something like a small `std::map<>`.  You could do something like:

```cc
  auto domain = v.value("domain", "");
  if (!v.contains("metadata")) return ErrorInfo{std::move(reason), std::move(domain), {});
  auto metadata = std::unordered_map<std::string, std::string>{};
  for (auto const& m : v["metadata"].items()) {
    metadata[m.key()] = m.value();
  }
...
```

coryan (approved):
The code LGTM. There is more copying than needed, but this is not in the critical path.

coryan (approved):

google/cloud/storage/internal/http_response.cc
devjgm: Thanks. Due to the likely minimal copying that would save, I think I'll leave the code as is, because I think it's possibly slightly easier to read as is.

Issue #7664, opened 2021-11-24
Generator doxygen comments include broken links


coryan:
The documentation as rendered looks like this:

https://googleapis.dev/cpp/google-cloud-iam/1.33.0/classgoogle_1_1cloud_1_1iam_1_1IAMClient.html#a317a11faf4b5ad519550a6764dcfac3c

It seems like this is attempting to use markdown link aliases, e.g. `[ServiceAccount][google.iam.admin.v1.ServiceAccount]`, but the link aliases are not defined in the Doxygen block:

https://github.com/googleapis/google-cloud-cpp/blob/7bcb5f746b4591c21e263c9d66864ba54f4f0fd7/google/cloud/iam/iam_client.h#L106-L121

Note that the previous references are to the `v1.33.0` release, but this is still the case on `main`.

coryan:
Ah, it seems these link aliases are taken verbatim from the proto comments:

https://github.com/googleapis/googleapis/blob/9a04fde55ca0b57bdb34cbc1716e52274a690169/google/iam/admin/v1/iam.proto#L79

coryan:
Since this is just copying the comments verbatim, it is really a feature request, not a bug.

coryan:
One possible "fix" is to make these links not look like links.  Maybe the move to stored the docs in cloud.google.com will fix this too.

coryan:
This looks fixed.

coryan:
It is not fixed, for example look at:

https://googleapis.dev/cpp/google-cloud-secretmanager/latest/classgoogle_1_1cloud_1_1secretmanager_1_1SecretManagerServiceClient.html#aab81903782c1a0f0354bbfd851ace46d

Apparently I fixed the request and response types used in each method, but not any additional references.

PR #7665, opened 2021-11-24
feat(generator): link each method input and output types

coryan:
The protobuf documentation seems to use reference-style markdown links
for protos in the comments, i.e., each reference to `Foo` is formatted
as `[Foo][google.blah.blah.v1.Foo]`.  This formatting is very
distracting if left verbatim, but it can be turned into a link if we
define `google.blah.blah.v1.Foo` as a link reference in Doxygen's
markdown.

With this change the input and output types of each method get such
reference link definitions. The definitions themselves are not visible,
but they make the generated Doxygen output easier to read.

This does not solve the problem when the referenced entity is not an
input or output type of the method, or when it references another
method. But it works in enough cases to make the change worthwhile.

Part of the work for #7664

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7665)


coryan:
Sorry about the churn, should be in a good place to review now.

dbolduc (approved):

dbolduc:
You are adding links for many methods that don't need them, but I think it's harmless.

PR #7669, opened 2021-11-28
feat(common): add support for call-tree-specific options

devbww:
Provide access to the prevailing `Options` for an operation without having
to plumb function parameters through the various and sundry layers, like
connection.

This also means that helper libraries can determine things like whether a
tracing component is enabled, without having to distort their APIs.

For example, if all `ServiceClient::Operation()` calls are implemented like
```
class ServiceClient {
 public
  ServiceClient(std::shared_ptr<ServiceConnection> connection,
                Options options = {})
      : connection_(std::move(connection)),
        options_(ServiceDefaultOptions(std::move(options))) {}

  T Operation(..., Options options = {}) {
    internal::OptionsSpan span(internal::MergeOptions(options, options_));
    ...
  }

 private:
  std::shared_ptr<ServiceConnection> connection_;
  Options options_;
};
```
then the connection layer, or any other internal library, can retrieve the
prevailing options using `internal::CurrentOptions()`.  If the operation is
asynchronous, the invocation-time `Options` are also installed during its
callbacks.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7669)

### DIFF
--- a/google/cloud/internal/async_read_stream_impl.h
+++ b/google/cloud/internal/async_read_stream_impl.h
@@ -17,6 +17,7 @@
 
 #include "google/cloud/grpc_error_delegate.h"
 #include "google/cloud/internal/completion_queue_impl.h"
+#include "google/cloud/options.h"
 #include "google/cloud/version.h"
 #include <grpcpp/support/async_stream.h>
 #include <memory>
@@ -150,10 +151,12 @@ class AsyncReadStreamImpl
      private:
       void Cancel() override {}  // LCOV_EXCL_LINE
       bool Notify(bool ok) override {
+        OptionsSpan span(options_);
         control_->OnStart(ok);
         return true;
       }
       std::shared_ptr<AsyncReadStreamImpl> control_;
+      Options options_ = CurrentOptions();
     };
 
     context_ = std::move(context);
@@ -195,10 +198,12 @@ class AsyncReadStreamImpl
      private:
       void Cancel() override {}  // LCOV_EXCL_LINE
       bool Notify(bool ok) override {
+        OptionsSpan span(options_);
         control_->OnRead(ok, std::move(response));
         return true;
       }
       std::shared_ptr<AsyncReadStreamImpl> control_;
+      Options options_ = CurrentOptions();
     };
 
     auto callback = std::make_shared<NotifyRead>(this->shared_from_this());
@@ -243,10 +248,12 @@ class AsyncReadStreamImpl
      private:
       void Cancel() override {}  // LCOV_EXCL_LINE
       bool Notify(bool ok) override {
+        OptionsSpan span(options_);
         control_->OnFinish(ok, MakeStatusFromRpcError(status));
         return true;
       }
       std::shared_ptr<AsyncReadStreamImpl> control_;
+      Options options_ = CurrentOptions();
     };
 
     auto callback = std::make_shared<NotifyFinish>(this->shared_from_this());
@@ -281,10 +288,12 @@ class AsyncReadStreamImpl
      private:
       void Cancel() override {}  // LCOV_EXCL_LINE
       bool Notify(bool ok) override {
+        OptionsSpan span(options_);
         control_->OnDiscard(ok, std::move(response));
         return true;
       }
       std::shared_ptr<AsyncReadStreamImpl> control_;
+      Options options_ = CurrentOptions();
     };
 
     auto callback = std::make_shared<NotifyDiscard>(this->shared_from_this());
--- a/google/cloud/options.cc
+++ b/google/cloud/options.cc
@@ -46,18 +46,21 @@ namespace {
 // The prevailing options for the current operation.  Thread local, so
 // additional propagation must be done whenever work for the operation
 // is done in another thread.
-thread_local Options current_options;
+Options& ThreadLocalOptions() {
+  thread_local Options current_options;
+  return current_options;
+}
 
 }  // namespace
 
-Options const& CurrentOptions() { return current_options; }
+Options const& CurrentOptions() { return ThreadLocalOptions(); }
 
 OptionsSpan::OptionsSpan(Options opts) : opts_(std::move(opts)) {
   using std::swap;
-  swap(opts_, current_options);
+  swap(opts_, ThreadLocalOptions());
 }
 
-OptionsSpan::~OptionsSpan() { current_options = std::move(opts_); }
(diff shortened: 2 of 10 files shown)


google/cloud/internal/async_read_stream_impl.h
```diff
@@ -150,10 +151,12 @@ class AsyncReadStreamImpl
      private:
       void Cancel() override {}  // LCOV_EXCL_LINE
       bool Notify(bool ok) override {
+        OptionsSpan span(options_);
```
coryan, line 154: I wonder if we should set these spans in the completion queue instead.  The `Start()` function is a good place to capture the current options:

https://github.com/googleapis/google-cloud-cpp/blob/b92933ba81c2e497e7daf239ce9872080da2828e/google/cloud/internal/default_completion_queue_impl.cc#L238-L241

and the notifications would be a good place to create the span:

https://github.com/googleapis/google-cloud-cpp/blob/b92933ba81c2e497e7daf239ce9872080da2828e/google/cloud/internal/default_completion_queue_impl.cc#L175-L179

google/cloud/options.cc
```diff
@@ -40,6 +41,24 @@ Options MergeOptions(Options a, Options b) {
   return a;
 }
 
+namespace {
+
+// The prevailing options for the current operation.  Thread local, so
+// additional propagation must be done whenever work for the operation
+// is done in another thread.
+thread_local Options current_options;
```
coryan, line 49: This obviously has a non-trivial destructor.  I am certain you know better than I do whether this is accepted by the Google Style Guide.  Just asking to maybe learn something.

coryan (commented):
A suggestion for a different implementation which might be (maybe) a more targeted change and easier to maintain as we add asynchronous operations.

google/cloud/options.h
```diff
@@ -306,6 +307,40 @@ void CheckExpectedOptions(Options const& opts, char const* caller) {
...
+ *
+ * @param opts the `Options` to install.
+ */
+class ABSL_MUST_USE_RESULT OptionsSpan {
+ public:
+  explicit OptionsSpan(Options opts);
+  ~OptionsSpan();
+
+ private:
+  Options opts_;
```
devjgm, line 341: nit: If I understand this correctly, `opts_` is actually the previous options that will be _restored_ after the span goes out of scope. That is, `opts_` is _not_ the same as `opts`, which was passed to the c'tor. If you think this distinction is worth capturing, consider renaming this data member to something like `restore_` or `old_opts_` or something.

google/cloud/options.h
```diff
@@ -306,6 +307,40 @@ void CheckExpectedOptions(Options const& opts, char const* caller) {
...
+ *     assert(internal::CurrentOptions().get<IntOption>() == 2);
+ *   }
+ *   assert(internal::CurrentOptions().get<IntOption>() == 1);
+ * }
+ * assert(!internal::CurrentOptions().has<IntOption>());
+ * @endcode
+ *
+ * @param opts the `Options` to install.
+ */
+class ABSL_MUST_USE_RESULT OptionsSpan {
```
devjgm, line 335: Q: Is `ABSL_MUST_USE_RESULT` needed here, or nice to have, or... what's this here for?

devjgm (commented):
Looks like a nice solution that's hidden from our users but will make our implementation plumbing much easier. Thanks! Just a couple small questions. The PR LGTM, but I'll let Carlos do the real review.

google/cloud/internal/async_read_stream_impl.h
devbww: I considered that, but thought that changing the `tag` map (which is really just a collection of `AsyncGrpcOperation`s) to a map of `AsyncGrpcOperation`/`Options` pairs was both a little more complicated, and belied the fact that the `Options` logically belonged to the operation.

It is certainly the case that the behavior is now distributed to the `AsyncGrpcOperation` sub-classes, but I've been thinking of that as a feature.  For example, the existence of the `internal::CompletionQueueImpl` interface implies that this behavior should not belong in a particular instantiation (the `DefaultCompletionQueueImpl`) of that interface.

Perhaps experience with this will lead to a better understanding, but for now I'd vote for leaving it as is.

google/cloud/options.cc
devbww: The non-trivial destructor for the namespace-scoped object is not a problem because the object is only accessible to its thread, and is not destroyed until the thread terminates.  So, there is no issue with destruction order, and indeed the Style Guide says as much.

What the Style Guide still does require for namespace-scoped objects, thread local or not, is a compile-time-constant initializer.  In the thread-local case I guess this is to guard against threads being created before global initialization is complete.  The workaround is the same in both cases ... to convert the namespace-scoped object to a function local, so that it is initialized on the first call, no matter when that occurs.  I've done that.  PTAL.

google/cloud/options.h
devbww: I had considered that, but concluded that the meaning of the member changes with time.  For the majority of its life it is indeed the "options to restore", but after the initializer list is complete it is the "options to install" (i.e., it is the `opts` passed to the constructor), and after restoration it is a "moved from options".  All in all, it seemed better to give it a generic name, and let the code specify its meaning at any point.

google/cloud/options.h
devbww: It is there to try to call out usages like `OptionsSpan(opts);` where you forget the variable name, and the temporary span dies at the semicolon.  For example, see https://github.com/abseil/abseil-cpp/blob/master/absl/base/attributes.h#L410-L414.  (Previously we might have used a macro trick to do something similar.)

That said, when I tried to test that, it didn't seem to trigger a warning.  But I'm not sure that wasn't just an artifact of which compiler+version I used, so I decided to leave it anyway.  But, happy to remove also.

google/cloud/internal/async_read_stream_impl.h
coryan: > For example, the existence of the `internal::CompletionQueueImpl` interface implies that this behavior should not belong in a particular instantiation (the `DefaultCompletionQueueImpl`) of that interface.

FWIW, the only reason for the interface is mocking :shrug:

> Perhaps experience with this will lead to a better understanding, but for now I'd vote for leaving it as is.

SGTM.

google/cloud/options.cc
coryan: Ack, thanks for looking into it.

coryan (approved):

PR #7673, opened 2021-11-30
feat(generator): document details about `*Client`

### LINKED ISSUE #7672: Add information about thread safety of `*Client` classes
Hand-crafted classes have a comment like this:

```cc
/**
...
...
 * @par Thread Safety
 *
 * Instances of this class created via copy-construction or copy-assignment
 * share the underlying pool of connections. Access to these copies via multiple
 * threads is guaranteed to work. Two threads operating on the same instance of
 * this class is not guaranteed to work.
...
...
 */
```

It seems that generated code does not include this comment.  It should, as customers often wonder how or if these classes can be used in multiple threads.

coryan:
Customers often wonder about the performance, thread-safety, and
connection pooling in our `*Client` classes. This changes the generator
to include this information in the generated code.

Fixes #7672

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7673)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -292,7 +292,31 @@ std::string FormatClassCommentsFromServiceComments(
                    absl::StrReplaceAll(
                        ChompByValue(service_source_location.leading_comments),
                        {{"\n\n", "\n///\n/// "}, {"\n", "\n/// "}}),
-                   "\n///");
+                   R"""(
+///
+/// @par Equality
+///
+/// Instances of this class created via copy-construction or copy-assignment
+/// always compare equal. Instances created with equal
+/// `std::shared_ptr<*Connection>` objects compare equal. Objects that compare
+/// equal share the same underlying resources.
+///
+/// @par Performance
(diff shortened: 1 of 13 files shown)


generator/internal/descriptor_utils.cc
```diff
@@ -292,7 +292,31 @@ std::string FormatClassCommentsFromServiceComments(
...
+///
+/// Creating a new instance of this class is a relatively expensive operation,
+/// new objects establish new connections to the service. In contrast,
+/// copy-construction, move-construction, and the corresponding assignment
+/// operations are relatively efficient as the copies share all underlying
+/// resources.
+///
+/// @par Thread Safety
+///
+/// Concurrent access to different instances of this class, even if the compare
```
dbolduc, line 314: s/they/the/

dbolduc (approved):

generator/internal/descriptor_utils.cc
coryan: Fixed, thanks.

PR #7683, opened 2021-12-02
feat(generator): add options support to generated clients

devbww:
The generated client constructor now takes a defaulted `Options` argument,
which is merged with the default options for the service, and then stored
as a member.

Similarly, the generated client operations also take a defaulted `Options`
argument, which is then merged with the client options from above, and
installed as the prevailing options for the duration of the operation by
instantiating an `OptionsSpan`.

Any module within the scope of that operation, or its continuations, can
then retrieve those prevailing options.  So, for example, a metadata
decorator could add a header based on a per-operation option, or a retry
policy might modify its behavior based on another option.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7683)

### DIFF
--- a/ci/cloudbuild/builds/cmake-install.sh
+++ b/ci/cloudbuild/builds/cmake-install.sh
@@ -51,6 +51,9 @@ mapfile -t actual_dirs < <(env -C "${INSTALL_PREFIX}" find -type d)
 mapfile -t expected_dirs < <(cat ci/etc/expected_install_directories)
 expected_dirs+=(
   ./include/google/api
+  ./include/google/bigtable
+  ./include/google/bigtable/admin
+  ./include/google/bigtable/admin/v2
   ./include/google/bigtable/v2
   ./include/google/cloud/bigquery/connection
   ./include/google/cloud/bigquery/connection/v1
@@ -62,6 +65,7 @@ expected_dirs+=(
   ./include/google/cloud/bigquery/reservation
   ./include/google/cloud/bigquery/reservation/v1
   ./include/google/cloud/bigquery/v2
+  ./include/google/cloud/bigtable
   ./include/google/cloud/bigtable/internal
   ./include/google/cloud/dialogflow
   ./include/google/cloud/dialogflow/v2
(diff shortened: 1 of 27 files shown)


ci/cloudbuild/builds/cmake-install.sh
```diff
@@ -51,6 +51,9 @@ mapfile -t actual_dirs < <(env -C "${INSTALL_PREFIX}" find -type d)
 mapfile -t expected_dirs < <(cat ci/etc/expected_install_directories)
 expected_dirs+=(
   ./include/google/api
+  ./include/google/bigtable
```
coryan, line 54: nit: this change seems unrelated, but I guess you had to do it as part of regenerating all the libraries?

generator/internal/client_generator.cc
```diff
@@ -91,7 +91,7 @@ Status ClientGenerator::GenerateHeader() {
     "$class_comment_block$\n"
     "class $client_class_name$ {\n"
     " public:\n"
-    "  explicit $client_class_name$(std::shared_ptr<$connection_class_name$> connection);\n"
+    "  explicit $client_class_name$(std::shared_ptr<$connection_class_name$> connection, Options options = {});\n"
```
coryan, line 94: I think using an unqualified `Options` here is Okay, because this code lives in `google::cloud::$library$::v1_X_Y` any `Options` class or struct (say one generated by protos) would be in `google::cloud::$library$::v1` or some such namespace.

Is my reasoning correct?

coryan (approved):
LGTM.  I a question for you, if you agree with my reasoning please go ahead and merge.

ci/cloudbuild/builds/cmake-install.sh
devbww: Correct.  That is, apart from the generator changes themselves, everything else here was either generated or required by a test because of the regeneration.

generator/internal/client_generator.cc
devbww: A little simpler that that, I think, in that I don't think we're expecting to find anything other than `google::cloud::Options`.  We already do this for `Options` in existing library-qualified namespaces, and also for other "common" symbols.

PR #7692, opened 2021-12-03
fix(pubsub): Change AsyncReadWriteStreamAuth to be usable with unique_ptr

dpcollins-google:
Also fix a bug where if it was cancelled while auth was being set up, neither the spawning context nor the created stream would be cancelled.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7692)

### DIFF
--- a/google/cloud/internal/async_read_write_stream_auth.h
+++ b/google/cloud/internal/async_read_write_stream_auth.h
@@ -20,7 +20,7 @@
 #include "google/cloud/version.h"
 #include <functional>
 #include <memory>
-#include <thread>
+#include <mutex>
 
 namespace google {
 namespace cloud {
@@ -47,14 +47,8 @@ class AsyncStreamingReadWriteRpcAuth
 
   future<bool> Start() override {
     using Result = StatusOr<std::unique_ptr<grpc::ClientContext>>;
-
-    std::unique_ptr<grpc::ClientContext> context;
-    {
-      std::lock_guard<std::mutex> g{state_->mu};
-      context = std::move(state_->initial_context);
-    }
     auto weak = std::weak_ptr<SharedState>(state_);
-    return auth_->AsyncConfigureContext(std::move(context))
+    return auth_->AsyncConfigureContext(state_->ReleaseInitialContext())
         .then([weak](future<Result> f) mutable {
           if (auto state = weak.lock()) return state->OnStart(f.get());
           return make_ready_future(false);
@@ -83,7 +77,18 @@ class AsyncStreamingReadWriteRpcAuth
 
  private:
   struct SharedState {
-    SharedState(StreamFactory factory, std::unique_ptr<grpc::ClientContext> initial_context) : factory(std::move(factory)), initial_context(std::move(initial_context)), stream(absl::make_unique<AsyncStreamingReadWriteRpcError<Request, Response>>(Status(StatusCode::kFailedPrecondition, "Stream is not yet started."))) {}
+    SharedState(StreamFactory factory,
+                std::unique_ptr<grpc::ClientContext> initial_context)
+        : factory(std::move(factory)),
+          initial_context(std::move(initial_context)),
+          stream(absl::make_unique<
+                 AsyncStreamingReadWriteRpcError<Request, Response>>(
+              Status(StatusCode::kInternal, "Stream is not yet started."))) {}
+
+    std::unique_ptr<grpc::ClientContext> ReleaseInitialContext() {
+      std::lock_guard<std::mutex> g{mu};
+      return std::move(initial_context);
+    }
 
     future<bool> OnStart(StatusOr<std::unique_ptr<grpc::ClientContext>> context) {
       std::lock_guard<std::mutex> g{mu};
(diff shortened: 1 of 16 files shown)


google/cloud/internal/async_read_write_stream_auth.h
```diff
@@ -28,79 +29,97 @@ namespace internal {
...
+
+    void Cancel() {
+      std::lock_guard<std::mutex> g{mu};
+      if (cancelled) return;
+      cancelled = true;
+      if (initial_context) initial_context->TryCancel();
+      stream->Cancel();
+    }
+
+    const StreamFactory factory;
```
coryan, line 114: east const or no `const`.  I am not sure const-qualifying this adds enough value 🤷

google/cloud/internal/async_read_write_stream_auth.h
```diff
@@ -28,79 +29,97 @@ namespace internal {
...
+    std::mutex mu;
+    std::unique_ptr<grpc::ClientContext> initial_context;  // ABSL_GUARDED_BY(mu)
+    std::unique_ptr<AsyncStreamingReadWriteRpc<Request, Response>> stream;  // ABSL_GUARDED_BY(mu)
+    bool cancelled = false;  // ABSL_GUARDED_BY(mu)
+  };
+
+  const std::shared_ptr<GrpcAuthenticationStrategy> auth_;
```
coryan, line 121: east const

google/cloud/internal/async_read_write_stream_auth.h
```diff
@@ -20,6 +20,7 @@
 #include "google/cloud/version.h"
 #include <functional>
 #include <memory>
+#include <thread>
```
coryan, line 23: is this for `std::mutex`?  Why not just `<mutex>`?

google/cloud/internal/async_read_write_stream_auth.h
```diff
@@ -28,79 +29,97 @@ namespace internal {
...
 
   future<bool> Start() override {
     using Result = StatusOr<std::unique_ptr<grpc::ClientContext>>;
 
-    auto weak =
-        std::weak_ptr<AsyncStreamingReadWriteRpcAuth>(this->shared_from_this());
-    return auth_->AsyncConfigureContext(std::move(context_))
+    std::unique_ptr<grpc::ClientContext> context;
+    {
+      std::lock_guard<std::mutex> g{state_->mu};
```
coryan, line 53: I think I would have written:

```cc
std::unique_lock<std::mutex> g{state_->mu}
auto context = std::move(state_->initial_context);
g.unlock();
auto weak = ...
...
```

Alternatively, add a `ReleaseContext()` function to `SharedState`?

google/cloud/internal/async_read_write_stream_auth.h
```diff
@@ -28,79 +29,97 @@ namespace internal {
...
+    return state_->Finish();
   }
 
  private:
-  future<bool> OnStart(StatusOr<std::unique_ptr<grpc::ClientContext>> context) {
-    if (!context) {
-      stream_ =
-          absl::make_unique<AsyncStreamingReadWriteRpcError<Request, Response>>(
+  struct SharedState {
+    SharedState(StreamFactory factory, std::unique_ptr<grpc::ClientContext> initial_context) : factory(std::move(factory)), initial_context(std::move(initial_context)), stream(absl::make_unique<AsyncStreamingReadWriteRpcError<Request, Response>>(Status(StatusCode::kFailedPrecondition, "Stream is not yet started."))) {}
```
coryan, line 86: I am sure the robots are unhappy about this....

coryan: I wonder if `kInternal` is better than `kFailedPrecondition` in this case.  After all, using this initial stream would be a bug in the implementation of the client library?  Thoughts?

google/cloud/internal/async_read_write_stream_auth.h
```diff
@@ -28,79 +29,97 @@ namespace internal {
...
+      std::lock_guard<std::mutex> g{mu};
+      if (cancelled) return;
+      cancelled = true;
+      if (initial_context) initial_context->TryCancel();
+      stream->Cancel();
+    }
+
+    const StreamFactory factory;
+    std::mutex mu;
+    std::unique_ptr<grpc::ClientContext> initial_context;  // ABSL_GUARDED_BY(mu)
```
coryan, line 116: Are these `ABSL_GUARDED_BY` comments useful?  What else would they be guarded by?

coryan (changes requested):

google/cloud/internal/async_read_write_stream_auth.h
dpcollins-google: They could be unguarded. I'm hoping I can find a way to get ABSL_GUARDED_BY to do the thing its supposed to with lock guards, then eventually add the annotation proper back. But thats a task for another day.

google/cloud/internal/async_read_write_stream_auth.h
dpcollins-google: Done. It does, because it explains why it is not mutex guarded: its value never changes.

google/cloud/internal/async_read_write_stream_auth.h
dpcollins-google: Done.

google/cloud/internal/async_read_write_stream_auth.h
dpcollins-google: Because I don't know the threading headers by heart yet :) done.

google/cloud/internal/async_read_write_stream_auth.h
dpcollins-google: Personal preference to avoid unlocking methods because scopes more clearly deliniate the locked region. But a release method is cleaner

google/cloud/internal/async_read_write_stream_auth.h
dpcollins-google: Changed to kInternal, I don't have a strong preference.

coryan (commented):
LGTM, but the robots are still unhappy.

coryan (approved):
LGTM. Thanks!

PR #7694, opened 2021-12-04
feat(storage): capture metadata info in downloads

### LINKED ISSUE #7677: Include metadata attributes on successful object downloads
When we download the media (data) of a GCS object we get several (but not all) object metadata attributes as HTTP headers. Some of these are extremely useful, e.g., generation, and the total object size.  The mechanism is different for GCS (where really the full metadata is included in the first response), but the data is there.  It would be nice to expose these through stable APIs and not just through debugging stuff (like `ObjectReadStream::headers()`)

coryan:
A lot of useful metadata information is included the the metadata
headers for a GCS download over REST, and as an initial `Object` proto
for gRPC. This information is needed to perform consistent reads from
versioned objects (or detect inconsistent reads from objects without
versioning).  I will need this information as part of the changes for
Arrow too.

Fixes #7677

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7694)

### DIFF
--- a/google/cloud/storage/tests/object_read_headers_integration_test.cc
+++ b/google/cloud/storage/tests/object_read_headers_integration_test.cc
@@ -51,6 +51,79 @@ class ObjectReadHeadersIntegrationTest
   std::string bucket_name_;
 };
 
+TEST_F(ObjectReadHeadersIntegrationTest, CaptureMetadataXml) {
+  StatusOr<Client> client = MakeIntegrationTestClient();
+  ASSERT_STATUS_OK(client);
+
+  auto const object_name = MakeRandomObjectName();
+
+  auto insert = client->InsertObject(bucket_name(), object_name, LoremIpsum(),
+                                     IfGenerationMatch(0));
+  ASSERT_THAT(insert, IsOk());
+  ScheduleForDelete(*insert);
+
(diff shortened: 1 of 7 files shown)


google/cloud/storage/tests/object_read_headers_integration_test.cc
```diff
@@ -51,6 +51,79 @@ class ObjectReadHeadersIntegrationTest
...
+  auto const object_name = MakeRandomObjectName();
+
+  auto insert = client->InsertObject(bucket_name(), object_name, LoremIpsum(),
+                                     IfGenerationMatch(0));
+  ASSERT_THAT(insert, IsOk());
+  ScheduleForDelete(*insert);
+
+  auto is = client->ReadObject(bucket_name(), object_name,
+                               Generation(insert->generation()));
+  EXPECT_EQ(insert->generation(), is.generation().value_or(0));
```
dbolduc, line 67: why are the `.value_or()`s necessary?

dbolduc (approved):

google/cloud/storage/tests/object_read_headers_integration_test.cc
dbolduc: Nevermind, I see it now.

PR #7695, opened 2021-12-04
ci: update all library READMES with a single script

coryan:
Use some markers in each library `README.md` file to inject the
`quickstart.cc` code. With this change it is now possible to update all
the libraries using a single script.

I also changed the scaffold generation to include these markers for
future libraries.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7695)


generator/internal/scaffold_generator.cc
```diff
@@ -261,10 +261,64 @@ void GenerateReadme(std::ostream& os,
...
+
+The [quickstart/](quickstart/README.md) directory contains a minimal environment
+to get started using this client library in a larger project. The following
+"Hello World" program is used in this quickstart, and should give you a taste of
+this library.
+
+<!-- inject-quickstart-start -->
+<!-- inject-quickstart-end -->
+
+* Packaging maintainers or developers that prefer to install the library in a
```
devbww, line 300: s/that/who/ perhaps?

devbww (approved):

generator/internal/scaffold_generator.cc
coryan: Done here and everywhere else (maybe I went too far, let me know).

devbww (approved):

PR #7696, opened 2021-12-04
fix(common): backoff policy clones initial state

dbolduc:
When calling `.clone()` on an exponential backoff policy, copy the initial state instead of the current state. This may result in a change of behavior, but we see it as a bug that ought to be fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7696)

### DIFF
--- a/google/cloud/bigtable/rpc_backoff_policy.cc
+++ b/google/cloud/bigtable/rpc_backoff_policy.cc
@@ -30,7 +30,8 @@ ExponentialBackoffPolicy::ExponentialBackoffPolicy(
 }
 
 std::unique_ptr<RPCBackoffPolicy> ExponentialBackoffPolicy::clone() const {
-  return std::unique_ptr<RPCBackoffPolicy>(new ExponentialBackoffPolicy(*this));
+  return std::unique_ptr<RPCBackoffPolicy>(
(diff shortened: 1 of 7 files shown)


coryan (approved):
Consider updating CHANGELOG.md as part of this PR, it will be harder to remember what we wanted to say when we release this.

google/cloud/bigtable/rpc_backoff_policy.cc
```diff
@@ -30,7 +30,8 @@ ExponentialBackoffPolicy::ExponentialBackoffPolicy(
 }
 
 std::unique_ptr<RPCBackoffPolicy> ExponentialBackoffPolicy::clone() const {
-  return std::unique_ptr<RPCBackoffPolicy>(new ExponentialBackoffPolicy(*this));
+  return std::unique_ptr<RPCBackoffPolicy>(
+      new ExponentialBackoffPolicy(initial_delay_, maximum_delay_));
```
devbww, line 34: Possible follow-up note:  It seems to me that copy construction/assignment for these "clone-able" classes is an attractive nuisance (they don't support equality after all), and that we'd be well served trying to delete that capability.

PR #7699, opened 2021-12-06
chore: Actually disable nlohmann/json test compilation which doesn't work on mac

dpcollins-google:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7699)


coryan:
This change seems cool, but I am not sure why it would have any effect on macOS.  The files are all Dockerfiles for Linux?  Oh, maybe you are using docker on macOS and compiling the Linux images?

coryan (approved):

PR #7704, opened 2021-12-07
refactor(spanner): prefer AsyncRetryLoop to StartRetryAsyncUnaryRpc

dbolduc:
Prefer `AsyncRetryLoop` which uses our `CompletionQueue` instead of the `grpc::CompletionQueue`. The mocks are also nicer.

With this PR, the only remaining uses of `StartRetryAsyncUnaryRpc` are in Bigtable. Eventually they will be removed as part of the effort to modernize Bigtable.

Also change the return type of `SessionPool::AsyncDeleteSession` to drop the `protobuf::Empty`. This function is not called outside of an integration test at the moment, but we plan to use it later.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7704)

### DIFF
--- a/google/cloud/spanner/integration_tests/session_pool_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/session_pool_integration_test.cc
@@ -36,7 +36,7 @@ struct SessionPoolFriendForTest {
                                                   num_sessions);
   }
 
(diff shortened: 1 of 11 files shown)


devbww (approved):
Thanks.

PR #7706, opened 2021-12-07
feat(spanner): spanner::Client construction from Options

devbww:
Add a new `spanner::Client` constructor using `google::cloud::Options`.
This can then be used to merge with per-operation `Options` in a follow
up PR.

`ClientOptions` construction is still supported for compatibility, but
it is immediately converted to an `Options` value.  `QueryOptions`
is still a first-class citizen, however, because of its use in the
`Connection` interface.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7706)


google/cloud/spanner/client.h
```diff
@@ -128,10 +130,31 @@ class Client {
...
+  explicit Client(std::shared_ptr<Connection> conn, Options opts = {})
       : conn_(std::move(conn)), opts_(std::move(opts)) {}
 
-  /// No default construction. Use `Client(std::shared_ptr<Connection>)`
+  /**
+   * Constructs a `Client` object using the specified @p conn and @p opts.
+   *
+   * Note that the previous constructor, which uses the new way to represent
+   * options of all varieties (`google::cloud::Options`), is now preferred.
+   */
```
dbolduc, line 141: mark as `@deprecated`?

google/cloud/spanner/client.h
```diff
@@ -128,10 +130,31 @@ class Client {
...
+      : conn_(std::move(conn)), opts_(opts) {}
+
+  /**
+   * Constructs a `Client` object using the specified @p conn and an empty
+   * initializer list.
+   *
+   * @note This constructor exists solely for backwards compatibility. It
+   *     prevents existing code that uses `Client(conn, {})` from breaking
+   *     due to ambiguity introduced by the `Options` overload.
+   */
```
dbolduc, line 152: In pubsub, I introduced this constructor as `@deprecated`, to scare away any users reading the documentation. In the long run, we do want to remove the function, but, we also want `Client(conn, {})` to be valid code that calls the `Options` constructor. (So I'm not sure using `@deprecated` was correct).

Changing from a `@note` to a `@warning` is also possible. I'll let you decide if you think either of these are improvements.

dbolduc (approved):
Just left some FYIs, which you probably already know and decided to improve on.

google/cloud/spanner/client_options.h
```diff
@@ -35,6 +36,12 @@ class ClientOptions {
   ClientOptions(ClientOptions&&) = default;
   ClientOptions& operator=(ClientOptions&&) = default;
 
+  /**
+   * Convert the `ClientOptions` to the new, recommended way to represent
+   * options of all varieties, `google::cloud::Options`.
+   */
+  explicit operator Options() const;
```
dbolduc, line 43: in pubsub a conversion method like this was hidden away in `internal` with the logic being that `ClientOptions` was to be deprecated, so no `public` methods should be added to it.

google/cloud/spanner/client.h
devbww: I planned to leave deprecating `ClientOptions` to a later stage.

google/cloud/spanner/client.h
devbww: Yeah, I don't think `@deprecated` is best for the reasons you cite.  And indeed, `@warning` falls to the same logic methinks.  The `@note` is really just an answer to "What's this?", and it all goes away when `ClientOptions` does.

google/cloud/spanner/client_options.h
devbww: I was thinking that a customer might want to call this as part of their own migration, but I'm just guessing.  In any case, with `ClientOptions` on the road to removal I'm not too concerned about adding to the API.

Issue #7720, opened 2021-12-09
Consider renaming BUILD -> BUILD.bazel


devjgm:
Bazel supports build files named either `BUILD` or `BUILD.bazel`. The bazel team doesn't seem to make an explicit recommendation, but this issue https://github.com/bazelbuild/bazel/issues/4517 seems to show that some of the community recommends the more explicit `BUILD.bazel` naming.

We might want to consider renaming our build files to `BUILD.bazel`. This could also be an advantage when this code is used in an environment where _blaze_ `BUILD` files are _also_ needed. 😉

PR #7722, opened 2021-12-09
refactor: rename BUILD -> BUILD.bazel

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/7720

Bazel supports `BUILD` and `BUILD.bazel`. The latter seems to be preferable because it's less likely to conflict with something else, and it will make our imports a bit easier. I also changed `WORKSPACE` -> `WORKSPACE.bazel`. 

This change should be transparent to our users.

NOTES: 
* There are still generic uses of the term "BUILD" file. This is common in bazel's docs, and I think it's fine in ours as well. That is, we can still say "update the BUILD file", and it means to update whichever bazel build file you have (either `BUILD` or `BUILD.bazel`). There were a few places where we referred to a _specific_ file, and in those cases I updated the documentation.
* We still have some `bazel/*.BUILD` files. I left these because they already have specific names. We _could_ rename them to something like `foo.BUILD.bazel` or `foo.bazel` instead, but that didn't seem much clearer.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7722)


devbww (commented):
I knew I should have assigned the issue to myself before working on this, or at least check that no one else had.  That said, I also had edits to `ci/generate-markdown/generate-packaging.sh` (and therefore to `doc/packaging.md`), in the "we provide Bazel files" bit.

coryan (commented):
Can you check the renovate bot documentation and verify it works with `WORKSPACE.bazel` files?

release/README.md
```diff
@@ -106,8 +106,8 @@ send the PR for review against `main`. You need to:
   `google/cloud/internal/version_info.h` file.
 
 **NOTE:** The Renovate bot will automatically update the Bazel deps in the
-quickstart `WORKSPACE` files after it sees the new release published. Watch for
-this PR to come through, kick off the builds, approve, and merge it.
+quickstart `WORKSPACE.bazel` files after it sees the new release published.
```
coryan, line 109: Answering my own question, seems like it does:

https://docs.renovatebot.com/modules/manager/bazel/

coryan (commented):
LGTM, but seems like @devbww would be a better reviewer

release/README.md
devjgm: Yes, I did check that ahead of time.

devjgm:
> I knew I should have assigned the issue to myself before working on this, or at least check that no one else had. That said, I also had edits to `ci/generate-markdown/generate-packaging.sh` (and therefore to `doc/packaging.md`), in the "we provide Bazel files" bit.

Ahh, sorry for wasting your time! We should have coordinated better. Done.

devbww (approved):
> Ahh, sorry for wasting your time! We should have coordinated better.

No, not at all ... I was the one working outside the lines.

Issue #7528, opened 2021-10-28
Prefer generated bigtable admin APIs


dbolduc:
Switch `TableAdmin` -> `BigtableTableAdminClient` and `InstanceAdmin` -> `BigtableInstanceAdminClient` everywhere except for tests of the legacy code.

PR #7729, opened 2021-12-11
cleanup(bigtable): prefer generated instance admin API

dbolduc:
Part of the work for #7528 

Some things to note:
* In `examples/bigtable_instance_admin_snippets.cc`, the project ID was previously consumed by `InstanceAdmin`. With the generated APIs, we have to pass it to each command. This explains some of the changes around the argument parsing.
* Funcitonal change to `TestIamPermissions`. Previously it was only used for testing permissions on an instance (so the inputs were project ID, instance ID). Now we test permissions on any resource by passing in the fully qualified resource name. (so you can test permissions for a cluster).
* Oh and I decided to just add the Table Admin mocks now alongside the Instance Admin mocks, even though they aren't used yet.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7729)

### DIFF
--- a/google/cloud/bigtable/CMakeLists.txt
+++ b/google/cloud/bigtable/CMakeLists.txt
@@ -211,6 +211,8 @@ create_bazel_config(google_cloud_cpp_bigtable)
 if (BUILD_TESTING)
     add_library(
         bigtable_client_testing # cmake-format: sort
+        admin/mocks/mock_bigtable_instance_admin_connection.h
+        admin/mocks/mock_bigtable_table_admin_connection.h
         testing/cleanup_stale_resources.cc
         testing/cleanup_stale_resources.h
         testing/embedded_server_test_fixture.cc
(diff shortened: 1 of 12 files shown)


google/cloud/bigtable/CMakeLists.txt
```diff
@@ -211,6 +211,8 @@ create_bazel_config(google_cloud_cpp_bigtable)
 if (BUILD_TESTING)
     add_library(
         bigtable_client_testing # cmake-format: sort
+        admin/mocks/mock_bigtable_instance_admin_connection.h
```
coryan, line 214: Maybe in a separate PR, but I think we would want a separate library for public mocks (as opposed to helpers for our own tests / mocks for stubs, etc.)

coryan (approved):

google/cloud/bigtable/CMakeLists.txt
dbolduc: Yeah you're right. I'll add `google_cloud_cpp_bigtable_mocks` in a future PR.

PR #7733, opened 2021-12-13
refactor(bigtable): prefer generated Table Admin

dbolduc:
Part of the work for #7528 

Same story as #7729, except with `TableAdmin`. `TableAdmin` holds a project ID and an instance ID. 

Some things to note:
* I passed on redoing the `WaitForConsistency` test now.
* There are still references to `TableAdmin` in `testing/embedded_test_fixture.*`. I do not yet have a plan for what to do with that class.
* I am not going to call the issue done yet, because I think I should take another pass at the documentation. (e.g. `bigtable-hello-table-admin.dox` references `g::c::bigtable::GcRule`. It should reference the proto.)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7733)

### DIFF
--- a/google/cloud/bigtable/examples/bigtable_table_admin_backup_snippets.cc
+++ b/google/cloud/bigtable/examples/bigtable_table_admin_backup_snippets.cc
@@ -12,26 +12,30 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "google/cloud/bigtable/admin/bigtable_table_admin_client.h"
 #include "google/cloud/bigtable/examples/bigtable_examples_common.h"
-#include "google/cloud/bigtable/table_admin.h"
+#include "google/cloud/bigtable/iam_binding.h"
+#include "google/cloud/bigtable/resource_names.h"
 #include "google/cloud/bigtable/testing/cleanup_stale_resources.h"
 #include "google/cloud/bigtable/testing/random_names.h"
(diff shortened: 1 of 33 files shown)


google/cloud/bigtable/examples/bigtable_table_admin_backup_snippets.cc
```diff
@@ -207,17 +278,19 @@ void SetIamPolicy(google::cloud::bigtable::TableAdmin const& admin,
...
+   argv.at(5));
 }
 
 void RunAll(std::vector<std::string> const& argv) {
   namespace examples = ::google::cloud::bigtable::examples;
   namespace cbt = ::google::cloud::bigtable;
+  namespace cbta = ::google::cloud::bigtable_admin;
 
   if (!argv.empty()) throw examples::Usage{"auto"};
   if (!examples::RunAdminIntegrationTests()) return;
```
coryan, line 296: Note that these examples are not executed on each PR, you may want to run them manually to make sure they work.

coryan (approved):

google/cloud/bigtable/examples/bigtable_table_admin_backup_snippets.cc
dbolduc: can confirm that the `RunAll` example works :+1:

PR #7737, opened 2021-12-14
chore: update cloud sdk

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7737)

### DIFF
--- a/ci/install-cloud-sdk.sh
+++ b/ci/install-cloud-sdk.sh
@@ -16,8 +16,8 @@
 
 set -euo pipefail
 
-readonly GOOGLE_CLOUD_CPP_CLOUD_SDK_VERSION="348.0.0"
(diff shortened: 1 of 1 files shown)


coryan (approved):
If you feel confident this fixes the issues in #441 we can remove the retry loop in:

https://github.com/googleapis/google-cloud-cpp/blob/13ec1e946ae1baad6bcae952daf5910649dcfd0a/ci/cloudbuild/builds/lib/integration.sh#L263

dbolduc:
> If you feel confident this fixes the issues in #441 we can remove the retry loop in:
> 
> https://github.com/googleapis/google-cloud-cpp/blob/13ec1e946ae1baad6bcae952daf5910649dcfd0a/ci/cloudbuild/builds/lib/integration.sh#L263

I have another PR to do this

Issue #7740, opened 2021-12-14
[Flake] Bigtable table admin snippets crashes


dbolduc:
GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/db23ecec-bcf4-4063-875f-d70c52b51b15;tab=detail?project=cloud-cpp-testing-resources
Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/d85214a59aae81f8330c71964ae569b274372224/fedora-msan-msan/log-db23ecec-bcf4-4063-875f-d70c52b51b15.txt

```
Running GenerateConsistencyToken() example
libc++abi: terminating
*** SIGABRT received at time=1639503165 on cpu 4 ***
PC: @     0x7fa5f8fc92a2  (unknown)  raise
    @     0x7fa5fde468d0        112  absl::lts_20211102::WriteFailureInfo()
    @     0x7fa5fde45ef7        112  absl::lts_20211102::AbslFailureSignalHandler()
    @           0x481727        384  SignalAction()
generated token is : TokenFor-projects/cloud-cpp-testing-resources/instances/test-instance/tables/tbl-2021-12-14-bneu2z8oylx60lsn08rboy9h9gjbpwe95x

Running CheckConsistency() example
Table is consistent with token TokenFor-projects/cloud-cpp-testing-resources/instances/test-instance/tables/tbl-2021-12-14-bneu2z8oylx60lsn08rboy9h9gjbpwe95x

Running DropRowsByPrefix() example
All rows starting with test-prefix/foo/bar/ successfully deleted

Running DropAllRows() example
    @         0xfffffffa  (unknown)  (unknown)
```

The culprit could be any combination of the following:
* #7733, which reimplemented the test in terms of the generated admin APIs
* #7737, which introduced a new bigtable emulator
* #7738, which removed the retry loop around bigtable integration tests. (Potentially, this retry loop has always been covering up these errors)

dbolduc:
One can reproduce this bug more consistently (~90% of the time) by:
1. Making the sample exit faster, by [dropping the processing](https://github.com/googleapis/google-cloud-cpp/blob/d2a68192ebc9c4d3c015e30ca32ad109f32ca9b6/google/cloud/bigtable/examples/table_admin_snippets.cc#L679-L683) after the client call.
[...]

Issue #7732, opened 2021-12-13
Add `WaitForConsistency` helper method for Bigtable generated Admin APIs


dbolduc:
Add the equivalent of: https://github.com/googleapis/google-cloud-cpp/blob/cd9e90fee3f9a4d03d745ff66e003a5b1a903d22/google/cloud/bigtable/table_admin.h#L819-L820
... for generated admin APIs. It is a common use case. Having a helper method would help us make the claim that the generated APIs are better.

```cc
namespace google::cloud::bigtable_admin {
future<Status> AsyncWaitForConsistency(CompletionQueue cq, BigtableTableAdminClient client, std::string const& table_name, std::string const& consistency_token, Options const& options = {});
}  // namespace google::cloud::bigtable_admin
```

dbolduc:
Summarizing a conversation with @coryan...

The proposed API is not super user friendly. The best way to use the code involves sharing the threads, CQ that the client uses with this method. Sample code would look something like this:
```cc
std::shared_ptr<BackgroundThreads> background_threads = SomehowMakeBackgroundThreads();
auto options = Options{}.set<GrpcCompletionQueueOption>(background_threads->cq());
auto connection = MakeBigtableTableAdminConnection(options);
auto client = BigtableTableAdminClient(connection)

// Wait for consistency with the same CQ that the client uses.
auto f = WaitForConsistency(background_threads->cq(), client, table_name, token);
```

But ideally, setting up the background threads is a job for the client library.

One option is to move the background threads into the `Client` layer, and give the client an accessor to get the threads' CQ.

Issue #7530, opened 2021-10-28
Consider implementing `TableAdmin`, `AdminClient` in terms of generated `BigtableTableAdminConnection`


dbolduc:
The design and tradeoffs for this change are described in #5934.

This is a breaking change (for users extending `AdminClient`) that involves stripping away all protected/public member functions except for `std::string AdminClient::project()`. We are likely to make this change, as it will probably be the least-hassle way for current users to adopt GUAC.

**Mini-Gotcha:** async stub call
[This public member function](https://github.com/googleapis/google-cloud-cpp/blob/1cd25c596d12d69ee5ab32d955aa26fd138210d3/google/cloud/bigtable/table_admin.h#L819-L820) relies on async stub calls. We do not generate async stub calls. I think it can be easily replaced by a custom polling loop that sends synchronous requests, running on a background thread in `TableAdmin`.

PR #7742, opened 2021-12-15
ci: workaround bigtable table admin snippets bug

dbolduc:
There seems to be lifetime issues around calling `TableAdmin::WaitForConsistency()`. Before #7733, the object's lifetime existed outside the scope of the sample method, so we never saw the bug surface. See #7740 for more details.

This PR adds a workaround for that bug. It is a dumb band-aid, but it seems more appropriate than retrying the entire CI build for Bigtable when the error is encountered.

I think the wisest course of action, moving forward, is to:
1. Implement #7732. This will not fix the lifetime issues, but it will allow us to drop the workaround.
2. Fix the lifetime issues, by reimplementing `TableAdmin` on top of the generated APIs (#7530), naturally as part of the effort to Modernize Bigtable.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7742)

### DIFF
--- a/google/cloud/bigtable/examples/table_admin_snippets.cc
+++ b/google/cloud/bigtable/examples/table_admin_snippets.cc
@@ -694,7 +694,13 @@ void WaitForConsistencyCheck(
     fut.get();  // simplify example by blocking until operation is done.
   }
   //! [wait for consistency check]
-  (std::move(old_admin), argv.at(2));
+  (old_admin, argv.at(2));
+
+  // TODO(#7740) - remove this extra call when lifetime issues are fixed.
+  //
+  // Use old_admin outside of the lambda to extend its lifetime long enough to
+  // avoid a crash.
+  (void)old_admin.ListTables(google::bigtable::admin::v2::Table::NAME_ONLY);
 }
 
 void CheckConsistency(
--- a/google/cloud/bigtable/examples/table_admin_snippets.cc
+++ b/google/cloud/bigtable/examples/table_admin_snippets.cc
@@ -20,7 +20,9 @@
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/testing_util/crash_handler.h"
 #include "google/bigtable/admin/v2/bigtable_table_admin.pb.h"
+#include <chrono>
 #include <sstream>
+#include <thread>
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/examples/table_admin_snippets.cc
```diff
@@ -694,7 +694,13 @@ void WaitForConsistencyCheck(
...
   }
   //! [wait for consistency check]
-  (std::move(old_admin), argv.at(2));
+  (old_admin, argv.at(2));
+
+  // TODO(#7740) - remove this extra call when lifetime issues are fixed.
+  //
+  // Use old_admin outside of the lambda to extend its lifetime long enough to
+  // avoid a crash.
+  (void)old_admin.ListTables(google::bigtable::admin::v2::Table::NAME_ONLY);
```
devbww, line 703: Do we need to make a call, or just not `move` it in the parameter list?

dbolduc: It still crashes without the `move` :/

devbww: Perhaps the `ListTables()` is lasting long enough that the extended lifetime of `old_admin` avoids the problem?  Maybe a sleep would suffice too.

For what it's worth ...
```
Step #3: libc++abi: terminating
Step #3: *** SIGABRT received at time=1639503165 on cpu 4 ***
Step #3: PC: @     0x7fa5f8fc92a2  (unknown)  raise
Step #3:     @     0x7fa5fde468d0        112  absl::lts_20211102::WriteFailureInfo()
Step #3:     @     0x7fa5fde45ef7        112  absl::lts_20211102::AbslFailureSignalHandler()
Step #3:     @           0x481727        384  SignalAction()
```
implies to me that "libc++abi" called `abort(3)`.  It would be nice if it told us more than "terminating".

You said you can reproduce it ~2% of the time, yes?  Is so, perhaps one of asan/msan/tsan might shed some light.

google/cloud/bigtable/examples/table_admin_snippets.cc
```diff
@@ -694,7 +696,11 @@ void WaitForConsistencyCheck(
     fut.get();  // simplify example by blocking until operation is done.
   }
   //! [wait for consistency check]
-  (std::move(old_admin), argv.at(2));
+  (old_admin, argv.at(2));
```
coryan, line 699: Consider creating a new function that receives `old_admin` as a parameter, but extends its lifetime until after the futures are destroyed.

Something like:

```cc
void WaitForConsistencyCheckImpl(
    google::cloud::bigtable::TableAdmin const& admin,
   std::vector<std::string> const& argv) { /* current code here */ }

void WaitForConsistencyCheck(
    google::cloud::bigtable_admin::BigtableTableAdminClient const&,
    std::vector<std::string> const& argv) {
  auto old_admin = google::cloud::bigtable::TableAdmin(
      google::cloud::bigtable::MakeAdminClient(argv.at(0)), argv.at(1));
  WaitForConsistencyCheckImpl(old_admin, argv);
}
```

dbolduc: This does not fix the problem. It appears that `old_admin` needs to exist for a certain amount of time. Not just exist for the scope of the method.

dbolduc:
I will soon implement the helper function and update the sample. Closing this PR.

devbww (approved):
> I think adding the sleep is a better way to hide the flake than re-running the bigtable emulator tests

OK (although it looks like you need to rebase to have the PR tests avoid the new-year problem).

PR #7745, opened 2021-12-16
feature(rest): port storage/internal curl types to internal for future use

scotthart:
These files have been copied over from storage/internal "functionally as is". Only include paths, namespaces, etc. have been modified. Eventually when GCS is refactored to use the new REST library the copies of these files in storage/internal will be removed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7745)


coryan (changes requested):
These files are not compiled anywhere, is that intentional?  If so, what is the next step?  Add them to what library?

And once you add them to a (new?) library: how do you avoid the duplication?

google/cloud/BUILD.bazel
```diff
@@ -157,3 +157,30 @@ load(":google_cloud_cpp_grpc_utils_benchmarks.bzl", "google_cloud_cpp_grpc_utils
         "@com_google_googleapis//google/spanner/v1:spanner_cc_grpc",
     ],
 ) for test in google_cloud_cpp_grpc_utils_benchmarks]
+
+load(":google_cloud_cpp_rest_internal.bzl", "google_cloud_cpp_rest_internal_hdrs", "google_cloud_cpp_rest_internal_srcs")
+
+cc_library(
+    name = "google_cloud_cpp_rest_internal",
```
coryan, line 164: nit: the default visibility here is "public", consider changing the visibility of this target to "packages" or something similar.

google/cloud/CMakeLists.txt
```diff
@@ -606,3 +606,155 @@ if (GOOGLE_CLOUD_CPP_ENABLE_GRPC)
...
+        PROPERTIES EXPORT_NAME "google-cloud-cpp::rest-internal"
+                   VERSION ${PROJECT_VERSION} SOVERSION
+                                              ${PROJECT_VERSION_MAJOR})
+    add_library(google-cloud-cpp::rest-internal ALIAS
+                google_cloud_cpp_rest_internal)
+
+    include(CreateBazelConfig)
+    create_bazel_config(google_cloud_cpp_rest_internal YEAR 2021)
+
+    # Define the benchmarks in a function so we have a new scope for variable
```
coryan, line 646: nit: s/benchmarks/unit tests/ ?

google/cloud/internal/curl_options.h
```diff
@@ -0,0 +1,162 @@
...
+ * them.
+ *
+ * @warning The library may create more connections than this option configures,
+ * for example if your application requests many simultaneous downloads.
+ */
+struct ConnectionPoolSizeOption {
+  using Type = std::size_t;
+};
+
+/**
```
coryan, line 72: This looks to me like a storage-only option.

google/cloud/internal/curl_options.h
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * When using formatted I/O operations (typically `operator>>(std::istream&...)`
+ * this option controls the size of the in-memory buffer kept to satisfy any I/O
+ * requests.
+ *
+ * Applications seeking optional performance for downloads should avoid
+ * formatted I/O, and prefer using `std::istream::read()`. This option has no
+ * effect in that case.
+ */
+struct DownloadBufferSizeOption {
```
coryan, line 83: same here, this is specific to GCS.

google/cloud/internal/curl_options.h
```diff
@@ -0,0 +1,162 @@
...
+ * When using formatted I/O operations (typically `operator<<(std::istream&...)`
+ * this option controls the size of the in-memory buffer kept before a chunk is
+ * uploaded. Note that GCS only accepts chunks in multiples of 256KiB, so this
+ * option is always rounded up to the next such multiple.
+ *
+ * Applications seeking optional performance for downloads should avoid
+ * formatted I/O, and prefer using `std::istream::write()`. This option has no
+ * effect in that case.
+ */
+struct UploadBufferSizeOption {
```
coryan, line 99: And here.

google/cloud/internal/curl_options.h
```diff
@@ -0,0 +1,162 @@
...
+ * Disables automatic OpenSSL locking.
+ *
+ * With older versions of OpenSSL any locking must be provided by locking
+ * callbacks in the application or intermediate libraries. The C++ client
+ * library automatically provides the locking callbacks. If your application
+ * already provides such callbacks, and you prefer to use them, set this option
+ * to `false`.
+ *
+ * @note This option is only useful for applications linking against
+ * OpenSSL 1.0.2.
```
coryan, line 113: I wonder if OpenSSL 1.0.2 is a thing we can ignore now?

coryan: Maybe just file a bug to follow up on this.

coryan (approved):

ci/etc/expected_install_directories
```diff
@@ -53,5 +53,6 @@
 ./lib64/cmake/google_cloud_cpp_iam
 ./lib64/cmake/google_cloud_cpp_logging
 ./lib64/cmake/google_cloud_cpp_pubsublite
+./lib64/cmake/google_cloud_cpp_rest_internal
```
devbww, line 56: `generate-libraries` seems to remove this for me.

coryan: My naming is awful, as usual.  `ci/etc/expected_install_directories` only contains the directories created by automatically generated libraries.  `ci/cloudbuilds/builds/cmake-install.sh` contains the list of directories created by hand-crafted libraries.

coryan: Meaning: we should move that hand-crafted directory to the `cmake-install.sh` script.

PR #7761, opened 2021-12-18
feat(storage): enable c-ares for libcurl under Bazel

### LINKED ISSUE #7462: asynchronous dns is not enabled in bazel builds of curl
Yes

All libraries built with bazel that depend on curl.

**Describe the bug**
The bazel build currently does not set `USE_ARES` or `USE_THREADS_POSIX` or `USE_THREADS_WIN32` in `bazel/curl.BUILD` which results in synchronous dns lookups which ultimately results in this issue: https://stackoverflow.com/questions/9191668/error-longjmp-causes-uninitialized-stack-frame

**To Reproduce**
The is difficult to reproduce since one would need to run into a timeout during dns lookup. This then causes an `alarm` to be issued causing a `longjmp` which crashes on modern linux systems.

**Expected behavior**
The bazel build of curl should use asynchronous dns lookup.

**Operating system:**
Ubuntu 18.04 and 20.04 but other modern linux os will be affected as well.

g++ 7

1.21 but this seems to not have changed in more recent version.

**Additional context**

coryan:
Fixes #7462

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7761)

### DIFF
--- a/bazel/curl.BUILD
+++ b/bazel/curl.BUILD
@@ -439,8 +439,9 @@ cc_library(
     }),
     visibility = ["//visibility:public"],
     deps = [
-        # Use the same version of zlib that gRPC does.
+        # Use the same version of zlib and c-ares that gRPC does.
         "//external:madler_zlib",
+        "//external:cares",
         ":define-ca-bundle-location",
     ] + select({
         ":windows": [],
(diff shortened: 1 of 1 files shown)


dbolduc (approved):
🤷‍♂️

PR #7768, opened 2021-12-21
feat(bigtable): add wait for consistency helper method

### LINKED ISSUE #7732: Add `WaitForConsistency` helper method for Bigtable generated Admin APIs

dbolduc:
Part of the work for #7732

This method will be needed in order to modernize `TableAdmin`. It currently lives in the `bigtable_admin_internal` namespace. We are not convinced that having a `CompletionQueue` parameter is very user friendly. Ideally, as @coryan points out, the work done for this method would be run on the same threads / CQ that the client uses.

It is also worth considering whether this belongs in a `bigtable_experimental` namespace as the method seems valuable, although it may change.

The implementation is `AsyncRetryLoop`, but with a few changes.
1. All templates have been resolved
2. The response is processed. An OK response that is not yet consistent counts as a transient failure.
3. Polling policy is used instead of Retry and Backoff policies. This is consistent with the previous functionality. 

This new version of the method also allows for the `future` to be cancelled.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768)


google/cloud/bigtable/admin/wait_for_consistency_test.cc
```diff
@@ -0,0 +1,381 @@
...
+  promise<bool> WaitForRequest() {
+    std::unique_lock<std::mutex> lk(mu_);
+    cv_.wait(lk, [&] { return !requests_.empty(); });
+    auto p = std::move(requests_.front());
+    requests_.pop_front();
+    return p;
+  }
+
+ private:
+  // We will simulate the asynchronous requests by pushing promises into this
```
coryan, line 228: There is a helper class to do this sort of thing:

https://github.com/googleapis/google-cloud-cpp/blob/5d19d3d16eb458414ff96162f3080b09d40369df/google/cloud/testing_util/async_sequencer.h#L30-L65

If it does not work in this case, maybe add a comment explaining why?

google/cloud/bigtable/admin/wait_for_consistency_test.cc
```diff
@@ -0,0 +1,381 @@
...
+  std::deque<promise<bool>> requests_;
+  int cancel_count_ = 0;
+};
+
+TEST_F(AsyncWaitForConsistencyCancelTest, CancelAndSuccess) {
+  std::string const table_name =
+      bigtable::TableName("test-project", "test-instance", "test-table");
+  std::string const token = "test-token";
+
+  auto fake = std::make_shared<testing_util::FakeCompletionQueueImpl>();
```
coryan, line 241: Instead of using the fake completion queue, consider using the mock:

https://github.com/googleapis/google-cloud-cpp/blob/5d19d3d16eb458414ff96162f3080b09d40369df/google/cloud/testing_util/mock_completion_queue_impl.h#L28

It is easier to grok what it does (we wrote the fake when it was not possible to mock a `google::cloud::CompletionQueue`).

google/cloud/bigtable/examples/table_admin_snippets.cc
```diff
@@ -657,44 +658,49 @@ void DropRowsByPrefix(
...
-    StatusOr<std::string> consistency_token =
-        admin.GenerateConsistencyToken(table_id);
-    if (!consistency_token) {
-      throw std::runtime_error(consistency_token.status().message());
+  [](cbta::BigtableTableAdminClient admin, std::string const& project_id,
+     std::string const& instance_id, std::string const& table_id) {
+    std::string table_name = cbt::TableName(project_id, instance_id, table_id);
+
+    // First, generate the consistency token
+    StatusOr<google::bigtable::admin::v2::GenerateConsistencyTokenResponse>
```
coryan, line 676: I would just write `auto consistency_token = admin.GenerateConsistencyToken(table_name);`

google/cloud/bigtable/examples/table_admin_snippets.cc
```diff
@@ -657,44 +658,49 @@ void DropRowsByPrefix(
...
-          } else {
-            std::cout
-                << "Table is not yet consistent, Please try again later with"
-                << " the same token (" << *consistency_token << ")\n";
-          }
-        });
-    fut.get();  // simplify example by blocking until operation is done.
+    std::string consistency_token = result->consistency_token();
+
+    // Spin off a thread to run the operation. An application will likely want
```
coryan, line 683: Ugh... Okay, I see why this works now, you have a completely different thread pool for `AsyncWaitForConsistency`.  That is ... not desirable, spinning one of these up for each request would be expensive, ideally an application should have a single completion queue and thread pool.

I hate that you are blocked on this, but the API for `AsyncWaitForConsistency()` is not something I would like us to expose to customers (even if none of them uses it).

coryan: Maybe if `AsyncWaitForConsistency()` used the internal CQ, maybe with an accessor in `internal::` or some kind of `friend` function or struct.

coryan (changes requested):

google/cloud/bigtable/admin/wait_for_consistency_test.cc
dbolduc: Ah, thanks for the reminder. I opened #7771 to add the cancel functionality to `AsyncSequencer`.

I propose that I leave the PR as is for now. The code in this test is stolen from [`async_retry_loop_test.cc`](https://github.com/googleapis/google-cloud-cpp/blob/5d19d3d16eb458414ff96162f3080b09d40369df/google/cloud/internal/async_retry_loop_test.cc#L58-L98).  I think it will be easiest to cleanup both of these test fixtures at the same time. (FYI: cleanup means using `AsyncSequencer` and `MockCompletionQueueImpl`)

google/cloud/bigtable/admin/wait_for_consistency_test.cc
dbolduc: See above comment in the `AsyncSequencer` thread

dbolduc:
To summarize the changes from the previous commit:
* the function was moved into an internal namespace. (Maybe it belongs in an experimental namespace?)
* the sample and integration test were reverted.

The motivation for this decision is explained in the PR description, and in this [comment](https://github.com/googleapis/google-cloud-cpp/issues/7732#issuecomment-999210591) on the original issue.

google/cloud/bigtable/admin/internal/wait_for_consistency.cc
```diff
@@ -0,0 +1,184 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
```
coryan, line 7: nit: s/http/https/

google/cloud/bigtable/admin/wait_for_consistency_test.cc
coryan: SGTM, as long as we have an issue for this cleanup.

google/cloud/bigtable/admin/wait_for_consistency_test.cc
coryan: SGTM.

google/cloud/bigtable/admin/internal/wait_for_consistency.cc
```diff
@@ -0,0 +1,184 @@
...
+
+using RespType = StatusOr<bigtable::admin::v2::CheckConsistencyResponse>;
+using TimerFuture = future<StatusOr<std::chrono::system_clock::time_point>>;
+
+// This class borrows heavily from `google::cloud::internal::AsyncRetryLoop`
+class AsyncWaitForConsistencyImpl
+    : public std::enable_shared_from_this<AsyncWaitForConsistencyImpl> {
+ public:
+  AsyncWaitForConsistencyImpl(CompletionQueue cq,
+                              bigtable_admin::BigtableTableAdminClient client,
```
coryan, line 35: Should this parameter  `std::shared_ptr<bigtable_admin::BigtableTableAdminConnection>` ?  Otherwise you will have the ownership cycle again?

google/cloud/bigtable/admin/internal/wait_for_consistency.cc
```diff
@@ -0,0 +1,184 @@
...
+                              std::string table_name,
+                              std::string consistency_token,
+                              Options const& options)
+      : cq_(std::move(cq)),
+        client_(std::move(client)),
+        table_name_(std::move(table_name)),
+        consistency_token_(std::move(consistency_token)),
+        options_(BigtableTableAdminDefaultOptions(std::move(options))),
```
coryan, line 43: Should you take the `options` parameter by value then?

google/cloud/bigtable/admin/internal/wait_for_consistency.cc
```diff
@@ -0,0 +1,184 @@
...
+    kWaiting,
+    kDone,
+  };
+
+  void StartAttempt() {
+    auto self = this->shared_from_this();
+    auto op =
+        client_.AsyncCheckConsistency(table_name_, consistency_token_, options_)
+            .then([self](future<RespType> f) { self->OnAttempt(f.get()); });
+    SetWaiting(std::move(op));
```
coryan, line 71: What happens if the `.then()` parameter is called before `SetWaiting()` gets a chance to run?

google/cloud/bigtable/admin/internal/wait_for_consistency.cc
```diff
@@ -0,0 +1,184 @@
...
+  void OnAttempt(RespType result) {
+    SetIdle();
+    // A successful attempt, set the value and finish the loop.
+    if (result.ok() && result->consistent()) {
+      SetDone(Status());
+      return;
+    }
+    auto status = std::move(result).status();
+    if (!polling_policy_->OnFailure(status)) {
+      if (status.ok()) {
```
coryan, line 83: Maybe:

```cc
if (!status.ok()) return SetDone(std::move(status));
return SetDone(Status(StatusCode::kDeadline...)
```

google/cloud/bigtable/admin/internal/wait_for_consistency.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "google/cloud/bigtable/admin/bigtable_table_admin_client.h"
+
+namespace google {
+namespace cloud {
+namespace bigtable_admin_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * Checks consistency of a table with multiple calls using a separate thread
```
coryan, line 26: s/a separate thread/background threads/

google/cloud/bigtable/admin/internal/wait_for_consistency.cc
```diff
@@ -0,0 +1,184 @@
...
+        SetDone(std::move(status));
+      }
+      return;
+    }
+    if (Cancelled()) return;
+    auto self = this->shared_from_this();
+    auto op =
+        cq_.MakeRelativeTimer(polling_policy_->WaitPeriod())
+            .then([self](TimerFuture f) { self->OnBackoffTimer(f.get()); });
+    SetWaiting(std::move(op));
```
coryan, line 96: As above.  The OS scheduler could insert a long sleep between lines 95 and 96, and by the time you call `SetWaiting()` then the timer has expired, and a new operation started.  There is an inherent race condition, it seems you need a mutex around this critical section.

coryan (changes requested):

dbolduc:
marking as a draft until potential races in `AsyncRetryLoop` are resolved

dbolduc (commented):
The major changes involve:
* updating the retry loop to follow the same design as `AsyncRetryLoop`.
* modernizing the test (also plagiarized from async retry loop)
* accepting a Connection instead of a Client

PTAL

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-:-Ms6t1Ar0ci8bnPg2z0u:b-7q6iil)* status: 0 of 8 files reviewed, 12 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[google/cloud/bigtable/admin/wait_for_consistency_test.cc, line 228 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-MrQXrx9ckrlb8aWE4ey-r2-228:-Ms5EZwi6C_OXEGlrgHT:ba1x42i) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bc62725707b34dca801cd0ad6f10be7fc5cba90e/google/cloud/bigtable/admin/wait_for_consistency_test.cc#L228)):*

SGTM, as long as we have an issue for this cleanup.
</blockquote></details>

Nevermind, I cleaned up the test in this commit.
___
*[google/cloud/bigtable/admin/internal/wait_for_consistency.h, line 26 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-MrXICCHqob6JjbPWmav-r4-26:-Ms4iInQ6Pli2tVOe7QD:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0d75d9c4344a0c4667303fdb5d6fee4ba2638c59/google/cloud/bigtable/admin/internal/wait_for_consistency.h#L26)):*

s/a separate thread/background threads/
</blockquote></details>

Done.
___
*[google/cloud/bigtable/admin/internal/wait_for_consistency.h, line 42 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-MrXICCHqob6JjbPWmav-r4-42:-Ms4iKJs8IKURhj_4Tnt:btvorlv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0d75d9c4344a0c4667303fdb5d6fee4ba2638c59/google/cloud/bigtable/admin/internal/wait_for_consistency.h#L42)):*

This no longer applies?
</blockquote></details>

Yes. Thanks, I forgot.
___
*[google/cloud/bigtable/admin/internal/wait_for_consistency.cc, line 7 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-MrXICCHqob6JjbPWmau-r4-7:-Ms4i82JBy5shXJb0i0j:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0d75d9c4344a0c4667303fdb5d6fee4ba2638c59/google/cloud/bigtable/admin/internal/wait_for_consistency.cc#L7)):*

nit: s/http/https/
</blockquote></details>

Done.
___
*[google/cloud/bigtable/admin/internal/wait_for_consistency.cc, line 35 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-MrXICCHqob6JjbPWmau-r4-35:-Ms4i_if9MCLqMo9pEla:b-m1iflk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0d75d9c4344a0c4667303fdb5d6fee4ba2638c59/google/cloud/bigtable/admin/internal/wait_for_consistency.cc#L35)):*

Should this parameter  `std::shared_ptr<bigtable_admin::BigtableTableAdminConnection>` ?  Otherwise you will have the ownership cycle again?
</blockquote></details>

I changed it.

(You don't have to read past here)

To answer your question: There would be no cycle using Client at the moment because the background threads are owned outside of the Client. But we would run into one later.

In the long term, we want our public API to look like:

```cc
namespace google::cloud::bigtable_admin {
future<Status> AsyncWaitForConsistency(bigtable_admin::BigtableTableAdminClient client, std::string table_name, std::string consistency_token, Options options);
}  // namespace google::cloud::bigtable_admin
```

We would need:
1. The background threads to live in the Client instead of the Connection. (to break the cycle).
2. An accessor to get the CQ from the Client.
3. An accessor to get the Connection from the Client.

If we have all these things, then the public function can just call the internal implementation like so: 

```cc
namespace google::cloud::bigtable_admin {
future<Status> AsyncWaitForConsistency(bigtable_admin::BigtableTableAdminClient client, std::string table_name, std::string consistency_token, Options options) {
  return bigtable_admin_internal::AsyncWaitForConsistency(client->cq(), client->connection(), std::move(table_name), std::move(consistency_token), std::move(options));
  // Maybe I need to change the name of the internal function ...
}
}  // namespace google::cloud::bigtable_admin
```
___
*[google/cloud/bigtable/admin/internal/wait_for_consistency.cc, line 43 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-MrXICCHqob6JjbPWmau-r4-43:-Ms4ixjV2O5oixexBDv0:bgwt35a) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0d75d9c4344a0c4667303fdb5d6fee4ba2638c59/google/cloud/bigtable/admin/internal/wait_for_consistency.cc#L43)):*

Should you take the `options` parameter by value then?
</blockquote></details>

Yes. Fixed.
___
*[google/cloud/bigtable/admin/internal/wait_for_consistency.cc, line 71 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-MrXICCHqob6JjbPWmau-r4-71:-Ms5EM93CS-yTgM1vplU:bu0xuhd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0d75d9c4344a0c4667303fdb5d6fee4ba2638c59/google/cloud/bigtable/admin/internal/wait_for_consistency.cc#L71)):*

What happens if the `.then()` parameter is called before `SetWaiting()` gets a chance to run?
</blockquote></details>

Resolved by using the updated `AsyncRetryLoop` desi
___
*[google/cloud/bigtable/admin/internal/wait_for_consistency.cc, line 83 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-MrXICCHqob6JjbPWmau-r4-83:-Ms4y2ro7Dz7ktfFuzjG:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0d75d9c4344a0c4667303fdb5d6fee4ba2638c59/google/cloud/bigtable/admin/internal/wait_for_consistency.cc#L83)):*

Maybe:

```cc
if (!status.ok()) return SetDone(std::move(status));
return SetDone(Status(StatusCode::kDeadline...)
```

</blockquote></details>

Done.
___
*[google/cloud/bigtable/admin/internal/wait_for_consistency.cc, line 96 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7768#-MrXICCHqob6JjbPWmau-r4-96:-Ms5EU-p72PjzdZwRMfD:bgjsi0o) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0d75d9c4344a0c4667303fdb5d6fee4ba2638c59/google/cloud/bigtable/admin/internal/wait_for_consistency.cc#L96)):*

As above.  The OS scheduler could insert a long sleep between lines 95 and 96, and by the time you call `SetWaiting()` then the timer has expired, and a new operation started.  There is an inherent race condition, it seems you need a mutex around this critical section.
</blockquote></details>

Resolved by using the updated `AsyncRetryLoop` design

google/cloud/bigtable/admin/internal/wait_for_consistency.cc
```diff
@@ -0,0 +1,185 @@
+// Copyright 2021 Google LLC
```
coryan, line 1: nit: now it is 2022.

google/cloud/bigtable/admin/internal/wait_for_consistency.h
```diff
@@ -0,0 +1,52 @@
+// Copyright 2021 Google LLC
```
coryan, line 1: nit: s/2021/2022/

google/cloud/bigtable/admin/internal/wait_for_consistency_test.cc
```diff
@@ -0,0 +1,423 @@
+// Copyright 2021 Google LLC
```
coryan, line 1: s/2021/2022/

coryan (changes requested):
Because of this:

https://github.com/googleapis/google-cloud-cpp/blob/eec706292cb1f815f67d4ef92571efb035fc8de7/.codecov.yml#L16

Your test and code is not being included in the code coverage metrics. Which I would like to see to find out if the tests are sufficiently complete.  Can we move these files to `google/cloud/bigtable/internal/`?

dbolduc:
> Your test and code is not being included in the code coverage metrics. Which I would like to see to find out if the tests are sufficiently complete. Can we move these files to `google/cloud/bigtable/internal/`?

Done. I'm glad you remembered because I didn't. Going forward, I will keep handwritten code out of `bigtable/admin/`.

coryan (approved):
😮‍💨

devbww:
Remember we have a convention to try to refrain from merging PRs between cutting the release branch and merging the "bump version" PR.

dbolduc:
Sorry, I couldn't hold it in any longer

PR #7769, opened 2021-12-21
feat(pubsublite): Add stream auth to stream factories

dpcollins-google:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7769)

### DIFF
--- a/google/cloud/internal/async_read_write_stream_auth.h
+++ b/google/cloud/internal/async_read_write_stream_auth.h
@@ -18,7 +18,6 @@
 #include "google/cloud/internal/async_read_write_stream_impl.h"
 #include "google/cloud/internal/unified_grpc_credentials.h"
 #include "google/cloud/version.h"
-#include "absl/functional/bind_front.h"
 #include <functional>
 #include <memory>
 #include <mutex>
--- a/google/cloud/internal/async_read_write_stream_auth.h
+++ b/google/cloud/internal/async_read_write_stream_auth.h
@@ -128,6 +128,20 @@ class AsyncStreamingReadWriteRpcAuth
   std::shared_ptr<SharedState> const state_;
 };
 
+template <typename Request, typename Response>
+std::unique_ptr<AsyncStreamingReadWriteRpc<Request, Response>>
+MakeAuthorizedStreamingReadWriteRpc(
+    CompletionQueue const& cq, std::unique_ptr<grpc::ClientContext> context,
+    std::shared_ptr<google::cloud::internal::GrpcAuthenticationStrategy> auth,
+    PrepareAsyncReadWriteRpc<Request, Response> async_call) {
+  return absl::make_unique<AsyncStreamingReadWriteRpcAuth<Request, Response>>(
+      std::move(context), std::move(auth),
+      [=](std::unique_ptr<grpc::ClientContext> ctx) {
+        return MakeStreamingReadWriteRpc<Request, Response>(cq, std::move(ctx),
+                                                            async_call);
+      });
+}
+
 }  // namespace internal
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace cloud
--- a/google/cloud/internal/async_read_write_stream_auth.h
+++ b/google/cloud/internal/async_read_write_stream_auth.h
@@ -18,6 +18,7 @@
 #include "google/cloud/internal/async_read_write_stream_impl.h"
 #include "google/cloud/internal/unified_grpc_credentials.h"
 #include "google/cloud/version.h"
+#include "absl/functional/bind_front.h"
 #include <functional>
 #include <memory>
 #include <mutex>
@@ -127,20 +128,6 @@ class AsyncStreamingReadWriteRpcAuth
   std::shared_ptr<GrpcAuthenticationStrategy> const auth_;
   std::shared_ptr<SharedState> const state_;
 };
-
-template <typename Request, typename Response>
-std::unique_ptr<AsyncStreamingReadWriteRpc<Request, Response>>
-MakeAuthorizedStreamingReadWriteRpc(
-    CompletionQueue const& cq, std::unique_ptr<grpc::ClientContext> context,
-    std::shared_ptr<google::cloud::internal::GrpcAuthenticationStrategy> auth,
-    PrepareAsyncReadWriteRpc<Request, Response> async_call) {
-  return absl::make_unique<AsyncStreamingReadWriteRpcAuth<Request, Response>>(
-      std::move(context), std::move(auth),
-      [=](std::unique_ptr<grpc::ClientContext> ctx) {
-        return MakeStreamingReadWriteRpc<Request, Response>(cq, std::move(ctx),
-                                                            async_call);
-      });
-}
 
 }  // namespace internal
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
--- a/google/cloud/pubsublite/internal/stream_factory.h
+++ b/google/cloud/pubsublite/internal/stream_factory.h
@@ -15,6 +15,7 @@
 #ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_PUBSUBLITE_INTERNAL_STREAM_FACTORY_H
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_PUBSUBLITE_INTERNAL_STREAM_FACTORY_H
 
+#include "absl/functional/bind_front.h"
 #include "google/cloud/internal/async_read_write_stream_auth.h"
 #include "google/cloud/internal/async_read_write_stream_impl.h"
 #include "google/cloud/version.h"
--- a/google/cloud/pubsublite/internal/stream_factory.h
+++ b/google/cloud/pubsublite/internal/stream_factory.h
@@ -35,6 +35,11 @@ template <class Request, class Response>
 using StreamFactory =
     std::function<std::unique_ptr<BidiStream<Request, Response>>()>;
 
+template <class Request, class Response>
+using ContextStreamFactory =
+    std::function<std::unique_ptr<BidiStream<Request, Response>>(
+        std::unique_ptr<grpc::ClientContext> context)>;
+
 using ClientMetadata = std::unordered_map<std::string, std::string>;
 
 inline std::unique_ptr<grpc::ClientContext> MakeGrpcClientContext(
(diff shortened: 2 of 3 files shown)


google/cloud/internal/async_read_write_stream_auth.h
```diff
@@ -128,6 +128,20 @@ class AsyncStreamingReadWriteRpcAuth
...
 
+template <typename Request, typename Response>
+std::unique_ptr<AsyncStreamingReadWriteRpc<Request, Response>>
+MakeAuthorizedStreamingReadWriteRpc(
+    CompletionQueue const& cq, std::unique_ptr<grpc::ClientContext> context,
+    std::shared_ptr<google::cloud::internal::GrpcAuthenticationStrategy> auth,
+    PrepareAsyncReadWriteRpc<Request, Response> async_call) {
+  return absl::make_unique<AsyncStreamingReadWriteRpcAuth<Request, Response>>(
+      std::move(context), std::move(auth),
+      [=](std::unique_ptr<grpc::ClientContext> ctx) {
```
coryan, line 139: Note that `async_call` is a `absl::FunctionRef`:

https://github.com/googleapis/google-cloud-cpp/blob/6b4e66f77f88246fe39a244035ff0e1bf9b261f5/google/cloud/internal/async_read_write_stream_impl.h#L166-L169

Which should not be deep copied:

https://github.com/abseil/abseil-cpp/blob/f523d0dd69806d8057a898bb2595910558156aaa/absl/functional/function_ref.h#L36-L39

google/cloud/pubsublite/internal/stream_factory.h
```diff
@@ -48,14 +49,16 @@ inline std::unique_ptr<grpc::ClientContext> MakeGrpcClientContext(
 inline StreamFactory<pubsublite::v1::PublishRequest,
                      pubsublite::v1::PublishResponse>
 MakeStreamFactory(
-    // NOLINTNEXTLINE(performance-unnecessary-value-param)
-    std::shared_ptr<pubsublite::v1::PublisherService::StubInterface> stub,
+    std::shared_ptr<pubsublite::v1::PublisherService::StubInterface> const&
```
coryan, line 52: Maybe we should change things a bit.  Consider defining:

```cc
using StreamFactory =
    std::function<std::unique_ptr<BidiStream<Request, Response>>(
       google::cloud::CompletionQueue const& cq, std::unique_ptr<grpc::ClientContext> context)>;
```

Then you can define:

```cc
inline StreamFactory<pubsublite::v1::PublishRequest, pubsublite::v1::PublishResponse>
MakeStreamFactory(std::shared_ptr<pubsublite::v1::PublisherService::StubInterface> s) {
  return [stub = std::move(s)](
      google::cloud::CompletionQueue const& cq, std::unique_ptr<grpc::ClientContext> context) {
         return MakeStreamingReadWriteRpc(cq, std::move(context)
             [stub](grpc::ClientContext* context, grpc::CompletionQueue* cq) {
                 return stub->PrepareAsyncSubscribe(context, cq);
             }
      };
}
```

And you can define a generic decorator:

```cc
template <typename Request, typename Response>
inline StreamFactory<Request, Response>
AuthDecorator(
  std::shared_ptr<google::cloud::internal::GrpcAuthenticationStrategy> a,
  StreamFactory<Request, Response> c) {
  return [auth = std::move(s), child = std::move(c)](
    google::cloud::CompletionQueue const& cq, std::unique_ptr<grpc::ClientContext> context) {
      return absl::make_unique<AsyncStreamingReadWriteRpcAuth<Request, Response>>(
          cq, std::move(context), std::move(child));
    };
}
```

coryan: You can implement the move-captures in C++11 using a helper struct inside the function, or just make more copies.

coryan (changes requested):

google/cloud/pubsublite/internal/stream_factory.h
dpcollins-google: I've restructured this to not duplicate lambda definitions. I still want the interface to have all the arguments pre-bound, but there's less code duplication now and less tight coupling.

google/cloud/internal/async_read_write_stream_auth.h
dpcollins-google: Ack, reverted this change.

google/cloud/pubsublite/internal/stream_factory.h
```diff
@@ -45,74 +52,111 @@ inline std::unique_ptr<grpc::ClientContext> MakeGrpcClientContext(
...
+    return absl::make_unique<
+        internal::AsyncStreamingReadWriteRpcAuth<Request, Response>>(
+        std::move(context), auth,
+        [=](std::unique_ptr<grpc::ClientContext> ctx) {
+          return underlying(std::move(ctx));
+        });
+  };
+}
+
+template <class Obj, class R, class... Args>
```
coryan, line 77: I would have just written:

```cc
template <class Functor, class Request, class Response>
inline ContextStreamFactory<Request, Response> MakeRawStreamFactory(
    google::cloud::CompletionQueue const& cq, Function fun) {
  return [=](std::unique_ptr<grpc::ClientContext> context)
             -> std::unique_ptr<BidiStream<Request, Response>> {
    return internal::MakeStreamingReadWriteRpc<Request, Response>(
        cq, std::move(context), fun);
  };
}
```

and pass a lambda as `fun`, for example:

```cc
inline StreamFactory<pubsublite::v1::SubscribeRequest,
                     pubsublite::v1::SubscribeResponse>
MakeStreamFactory(
    std::shared_ptr<pubsublite::v1::SubscriberService::StubInterface> const&
        stub,
    std::shared_ptr<google::cloud::internal::GrpcAuthenticationStrategy> const&
        auth,
    ClientMetadata const& metadata = {}) {
  return AddClientMetadata(
      metadata,
      AddAuthContext(auth, MakeRawStreamFactory(
                               cq, [stub](grpc::ClientContext* ctx, grpc::CompletionQueue* cq) {
          return stub->PrepareAsyncSubscribe(ctx, cq); }
   )));
}
```

coryan (changes requested):
I still think this approach with factory functions is a dead-end.  At the end, if Pub/Sub does not look like the other libraries, I will simply not accept the changes.

Maybe I do not understand what is the final API you are intending to offer and the role of these functions in them.  Maybe you could write that down in a short doc or a bug?

dpcollins-google:
> I still think this approach with factory functions is a dead-end. At the end, if Pub/Sub does not look like the other libraries, I will simply not accept the changes.
> 
> Maybe I do not understand what is the final API you are intending to offer and the role of these functions in them. Maybe you could write that down in a short doc or a bug?

This is intended to be used in the implementation of the API, similar to how ResumableStreamingReadRPC takes a std::function stream factory here https://github.com/googleapis/google-cloud-cpp/blob/18f0e30b69dcd1e6cb66d855fe54b90b73b3f937/google/cloud/internal/resumable_streaming_read_rpc.h#L33, to create streams in a generic way, a similar helper for bidi streams will need to be developed to handle streaming retries in a generic manner for code reuse between the 4 bidi streaming RPCs used by Pub/Sub Lite.

coryan:
Let's get this code in and iterate on it.  You do need to make the robots happy, after that I will approve.

google/cloud/pubsublite/internal/stream_factory.h
```diff
@@ -45,74 +52,111 @@ inline std::unique_ptr<grpc::ClientContext> MakeGrpcClientContext(
...
 
+template <class Request, class Response>
+StreamFactory<Request, Response> AddClientMetadata(
+    ClientMetadata const& metadata,
+    ContextStreamFactory<Request, Response> underlying) {
+  return [=] { return underlying(MakeGrpcClientContext(metadata)); };
+}
+
+template <typename Request, typename Response>
+inline ContextStreamFactory<Request, Response> AddAuthContext(
```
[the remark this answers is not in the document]
dpcollins-google: Done.

google/cloud/pubsublite/internal/stream_factory.h
dpcollins-google: The way you've written it, MakeRawStreamFactory needs to provide the request and response arguments at the call site, i.e. `MakeRawStreamFactory<pubsublite::v1::SubscribeRequest, pubsublite::v1::SubscribeResponse>` since there's no where but the functor where the types can be deduced. I can make this change, but I think the current structure is more clear.

coryan (approved):

coryan:
Looks like this needs to be rebased.

dpcollins-google:
Rebased, adding automerge tag

coryan:
FWIW, GitHub has a native "Enable auto-merge" button.  Both should do the same.

coryan:
Oh, and if the builds get stuck I can restart one build or two, no need to restart them all.

coryan:
Looks like it needs a little more work:

```
FAILED: google/cloud/pubsublite/CMakeFiles/pubsublite_internal_stream_factory_test.dir/internal/stream_factory_test.cc.o 
/usr/bin/cmake -E __run_co_compile --launcher=/usr/bin/ccache --tidy="/usr/local/bin/clang-tidy-wrapper;--extra-arg-before=--driver-mode=g++" --source=../google/cloud/pubsublite/internal/stream_factory_test.cc -- /usr/bin/clang++ -DGOOGLE_CLOUD_CPP_HAVE_GETRUSAGE=1 -DGOOGLE_CLOUD_CPP_HAVE_RUSAGE_THREAD=1 -DGTEST_LINKED_AS_SHARED_LIBRARY=1 -I../ -isystem . -isystem external/googleapis -isystem google/cloud/pubsublite -Wall -Wextra -Wconversion -Wno-sign-conversion -Werror -pthread -std=gnu++11 -MD -MT google/cloud/pubsublite/CMakeFiles/pubsublite_internal_stream_factory_test.dir/internal/stream_factory_test.cc.o -MF google/cloud/pubsublite/CMakeFiles/pubsublite_internal_stream_factory_test.dir/internal/stream_factory_test.cc.o.d -o google/cloud/pubsublite/CMakeFiles/pubsublite_internal_stream_factory_test.dir/internal/stream_factory_test.cc.o -c ../google/cloud/pubsublite/internal/stream_factory_test.cc
/workspace/cmake-out/../google/cloud/pubsublite/internal/stream_factory.h:134:74: error: the parameter 'auth' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
    std::shared_ptr<google::cloud::internal::GrpcAuthenticationStrategy> auth,
                                                                         ^
    const                                                               &
35459 warnings generated.
Suppressed 35566 warnings (35458 in non-user code, 108 NOLINT).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
1 warning treated as error
```

coryan:
I think we can close this in favor of #7974

coryan (changes requested):
I think we should not implement these changes and actually call the generated factory functions.

PR #7782, opened 2021-12-23
feat(common): include the request/response type name in the RPC log

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7782)

### DIFF
--- a/google/cloud/internal/log_wrapper_test.cc
+++ b/google/cloud/internal/log_wrapper_test.cc
@@ -60,6 +60,7 @@ TEST(LogWrapper, DefaultOptions) {
   TracingOptions tracing_options;
   // clang-format off
   std::string const text =
+      R"pb(google.spanner.v1.Mutation { )pb"
       R"pb(insert { )pb"
       R"pb(table: "Singers" )pb"
       R"pb(columns: "SingerId" )pb"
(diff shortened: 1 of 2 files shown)


google/cloud/internal/log_wrapper_test.cc
```diff
@@ -60,6 +60,7 @@ TEST(LogWrapper, DefaultOptions) {
   TracingOptions tracing_options;
   // clang-format off
   std::string const text =
+      R"pb(google.spanner.v1.Mutation { )pb"
```
coryan, line 63: Feels like this could be a single raw string, maybe for another PR.

google/cloud/internal/log_wrapper_test.cc
```diff
@@ -118,7 +121,9 @@ TEST(LogWrapper, Truncate) {
   TracingOptions tracing_options;
   tracing_options.SetOptions("truncate_string_field_longer_than=8");
   // clang-format off
-  std::string const text =R"pb(insert { )pb"
+  std::string const text =
```
coryan, line 124: Same here.

coryan (approved):
Very good idea, thanks.

google/cloud/internal/log_wrapper_test.cc
devbww: I think the lack of newlines dooms this to being some form of concatenation.

google/cloud/internal/log_wrapper_test.cc
devbww: Ditto.

PR #7783, opened 2021-12-23
test: modernize RetryLoop tests

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7783)

### DIFF
--- a/google/cloud/internal/async_retry_loop_test.cc
+++ b/google/cloud/internal/async_retry_loop_test.cc
@@ -15,7 +15,8 @@
 #include "google/cloud/internal/async_retry_loop.h"
 #include "google/cloud/internal/background_threads_impl.h"
 #include "google/cloud/options.h"
-#include "google/cloud/testing_util/fake_completion_queue_impl.h"
+#include "google/cloud/testing_util/async_sequencer.h"
+#include "google/cloud/testing_util/mock_completion_queue_impl.h"
 #include "google/cloud/testing_util/status_matchers.h"
 #include <gmock/gmock.h>
 #include <deque>
@@ -27,6 +28,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
 namespace {
 
+using ::google::cloud::testing_util::AsyncSequencer;
 using ::google::cloud::testing_util::IsOk;
 using ::google::cloud::testing_util::StatusIs;
 using ::testing::AllOf;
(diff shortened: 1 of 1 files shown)


dbolduc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7783)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[google/cloud/internal/async_retry_loop_test.cc, line 114 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7783#-MrcHJTc3_-B2rw5-IO6:-MrcHJTc3_-B2rw5-IO7:brskkfl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c01b513824357cd2db9bd5d3c2c361ea1edd77ab/google/cloud/internal/async_retry_loop_test.cc#L114)):*
> ```cpp
>           return SimulateRelativeTimer(d);
>         });
>     EXPECT_CALL(*mock, MakeDeadlineTimer)
> ```

is this called?
___
*[google/cloud/internal/async_retry_loop_test.cc, line 402 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7783#-MrcHheA0ZVXdttl0kzg:-MrcHheA0ZVXdttl0kzh:b-jsdza7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c01b513824357cd2db9bd5d3c2c361ea1edd77ab/google/cloud/internal/async_retry_loop_test.cc#L402)):*
> ```cpp
>   future<StatusOr<int>> actual = AsyncRetryLoop(
>       LimitedErrorCountRetryPolicy<TestRetryablePolicy>(5).clone(),
>       ExponentialBackoffPolicy(ms(1), ms(2), 2.0).clone(),
> ```

nit: I would replace all of these with `TestRetryPolicy(), TestBackoffPolicy(), ...` including the ones with 2000ms backoffs below, now that we are using the mock CQ and a sequencer



_Code quote:_
```cpp
      LimitedErrorCountRetryPolicy<TestRetryablePolicy>(5).clone(),
      ExponentialBackoffPolicy(ms(1), ms(2), 2.0).clone(),
```
___
*[google/cloud/internal/async_retry_loop_test.cc, line 409 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7783#-MrcFfSiEP7EJn7RzZUH:-MrcFfSiEP7EJn7RzZUI:b2dw9gd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c01b513824357cd2db9bd5d3c2c361ea1edd77ab/google/cloud/internal/async_retry_loop_test.cc#L409)):*
> ```cpp
>       42, "test-location");
> 
>   // First simulate a regular request that results in a
> ```

... transient failure
___
*[google/cloud/internal/async_retry_loop_test.cc, line 487 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7783#-MrcFpjFAXb9sTxUqy9f:-MrcFpjFAXb9sTxUqy9g:b-kh2wvn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c01b513824357cd2db9bd5d3c2c361ea1edd77ab/google/cloud/internal/async_retry_loop_test.cc#L487)):*
> ```cpp
>   EXPECT_EQ(0, CancelCount());
>   actual.cancel();
>   EXPECT_EQ(1, CancelCount());
> ```

So I think the point of this check originally, was to make sure that the timer was being cancelled and not the request. Thus the request cancel count was 0.

I think this means I was too narrow in my implementation of `CancelCount()`. I think we actually want to store a map from names to counts, so we can check that `CancelCount("Request") == 0`

google/cloud/internal/async_retry_loop_test.cc
```diff
@@ -55,46 +58,66 @@ std::unique_ptr<BackoffPolicy> TestBackoffPolicy() {
...
+  future<TimerResult> SimulateDeadlineTimer(
+      std::chrono::system_clock::time_point tp) {
+    return sequencer_.PushBack("DeadlineTimer")
+        .then([tp](future<Status> g) -> TimerResult {
+          auto status = g.get();
+          if (!status.ok()) return status;
+          return tp;
+        });
+  }
+
```
dbolduc, line 85: Should have pointed this out in the previous review, but this function is unused.

google/cloud/internal/async_retry_loop_test.cc
```diff
@@ -55,46 +58,66 @@ std::unique_ptr<BackoffPolicy> TestBackoffPolicy() {
...
-    auto p = std::move(requests_.front());
-    requests_.pop_front();
-    return p;
+    auto p = sequencer_.PopFrontWithName();
+    EXPECT_EQ("Request", p.second);
+    return std::move(p.first);
+  }
+  promise<Status> WaitForTimer() {
+    auto p = sequencer_.PopFrontWithName();
+    EXPECT_THAT(p.second, AnyOf("DeadlineTimer", "RelativeTimer"));
```
dbolduc, line 105: `EXPECT_EQ(p.second, "RelativeTimer");`

google/cloud/internal/async_retry_loop_test.cc
```diff
@@ -55,46 +58,66 @@ std::unique_ptr<BackoffPolicy> TestBackoffPolicy() {
...
+
 class AsyncRetryLoopCancelTest : public ::testing::Test {
  protected:
-  int CancelCount() {
-    std::lock_guard<std::mutex> lk(mu_);
-    int n = cancel_count_;
-    cancel_count_ = 0;
-    return n;
-  }
+  std::size_t CancelCount() { return sequencer_.CancelCount(); }
```
dbolduc, line 65: FYI: With the changes from #7784, this will need to become something like:
```cc
std::size_t RequestCancelCount() { return sequencer_.CancelCount("Request"); }
...
TEST(blah, blah) {
  EXPECT_EQ(0, RequestCancelCount());
}
```
or
```cc
std::size_t CancelCount(std::string name) { return sequencer_.CancelCount(name); }
....
TEST(blah, blah) {
  EXPECT_EQ(0, CancelCount("Request"));
}
```

and you could also check the timer's cancel count.

dbolduc (approved):

PR #7784, opened 2021-12-23
test: count async sequencer cancels by name

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7784)

### DIFF
--- a/google/cloud/testing_util/async_sequencer.h
+++ b/google/cloud/testing_util/async_sequencer.h
@@ -95,8 +95,9 @@ class AsyncSequencer {
   std::size_t CancelCount() { return CancelCount("unnamed"); }
   std::size_t CancelCount(std::string name) {
     std::lock_guard<std::mutex> lk(mu_);
-    auto n = cancel_counts_[name];
-    cancel_counts_[std::move(name)] = 0;
+    std::size_t n = 0;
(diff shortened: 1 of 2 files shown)


google/cloud/testing_util/async_sequencer.h
```diff
@@ -91,10 +92,11 @@ class AsyncSequencer {
     return p;
   }
 
-  std::size_t CancelCount() {
+  std::size_t CancelCount() { return CancelCount("unnamed"); }
+  std::size_t CancelCount(std::string name) {
     std::lock_guard<std::mutex> lk(mu_);
-    auto n = cancel_count_;
-    cancel_count_ = 0;
+    auto n = cancel_counts_[name];
```
coryan, line 98: nit: if you want to save a lookup:

```cc
  auto& count = cancel_counts_[std::move(name)];
  std::size_t n = 0;
  std::swap(n, count);
  return n;
```

coryan (approved):

google/cloud/testing_util/async_sequencer.h
dbolduc: I do want to save a lookup. Thanks.

PR #7787, opened 2021-12-28
cleanup(pubsublite): use bzl config files

dbolduc:
Generate .bzl config files from CMake (instead of defining globs in BUILD.bazel) for pubsublite.

cc: @dpcollins-google - We decided to go this route for libraries with handwritten code. Globs get a little more complicated when we have `examples`, `benchmarks`, `integration_tests`, `quickstart`s, etc. It also makes our libraries more consistent. Sorry if this inconveniences you.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7787)


google/cloud/pubsublite/CMakeLists.txt
```diff
@@ -123,7 +125,7 @@ target_compile_options(google_cloud_cpp_pubsublite_mocks
                        INTERFACE ${GOOGLE_CLOUD_CPP_EXCEPTIONS_FLAG})
 
 file(
-    GLOB test_files
```
coryan, line 126: I think we should stop using globs in the CMake file too.

coryan (changes requested):

google/cloud/pubsublite/CMakeLists.txt
dbolduc: Yeah, only doing half of the change didn't make sense. Fixed.

coryan (approved):

Issue #3932, opened 2020-04-24
Remove grpc_utils/ backwards compatibility code.


coryan:
We left headers, type aliases and code build script hacks to maintains backwards compatibility when we moved the symbols out of `google::cloud::grpc_utils::` (and the files out of `google/cloud/grpc_utils/`).

We do not have specific plans to clean this up, for now this is just a placeholder to point the TODOs to.

coryan:
Add `#warning` directives and start a timer to remove them.

coryan:
I updated the files and announced they will be removed around 2023-01-01

coryan:
Still waiting for the passage of time.

scotthart:
This will be part of the v3 release.

PR #7799, opened 2022-01-04
cleanup: deprecate grpc_utils/ files and symbols

coryan:
When we moved the symbols from `google::cloud::grpc_utils::*` to
`google::cloud::*` (and same thing with files) we created backwards
compatibility targets, headers and symbols, but did not specify when
they would get removed.  This adds all the warnings in all the right
places.

Part of the work for #3932

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7799)

### DIFF
--- a/google/cloud/grpc_utils/async_operation.h
+++ b/google/cloud/grpc_utils/async_operation.h
@@ -16,7 +16,13 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_GRPC_UTILS_ASYNC_OPERATION_H
 
 #include "google/cloud/async_operation.h"
-#include "google/cloud/version.h"
+#include "google/cloud/grpc_utils/version.h"
+#warning \
+    "We moved the helpers in the google::cloud::grpc_utils namespace to " \
+      "`google::cloud`, please use them directly. The aliases to maintain " \
+      "backwards compatibility will be removed on or about 2023-01-01. "    \
+      "More details at "                                                    \
+      "https://github.com/googleapis/google-cloud-cpp/issues/3932")
 
 namespace google {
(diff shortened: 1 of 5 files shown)


google/cloud/grpc_utils/version.h
```diff
@@ -15,7 +15,14 @@
 #ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_GRPC_UTILS_VERSION_H
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_GRPC_UTILS_VERSION_H
 
+#include "google/cloud/internal/attributes.h"
```
dbolduc, line 18: is this needed?

google/cloud/grpc_utils/async_operation.h
```diff
@@ -16,7 +16,13 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_GRPC_UTILS_ASYNC_OPERATION_H
 
 #include "google/cloud/async_operation.h"
-#include "google/cloud/version.h"
+#include "google/cloud/grpc_utils/version.h"
```
dbolduc, line 19: why did this `#include` change, but the others didn't?

dbolduc (approved):

google/cloud/grpc_utils/async_operation.h
coryan: Fixed.

google/cloud/grpc_utils/version.h
coryan: Nope, I thought I would use `[[deprecated]]` attributes, but at the end I did not.

PR #7802, opened 2022-01-04
refactor(internal): improve release.sh release notes

devjgm:
Parsing the `CHANGELOG.md` file to extract the release notes is now done
with awk, and it's more robust than before, because it stops on a line
that looks like `## vX.Y.Z` rather than just `## `. This also uses
`gh`'s explicit `--title` flag to set the title to only the release tag
name (e.g. "v1.35.0").

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7802)

### DIFF
--- a/release/release.sh
+++ b/release/release.sh
@@ -130,31 +130,19 @@ function run() {
 }
 
 # Outputs the release notes for the given tag. Looks for the release notes in
-# the CHANGELOG.md file starting at a heading that looks like "## <tag>". For
-# example `get_release notes v0.6.0` would look for a line like "## v0.6.0". If
-# no release notes could be found, a default string is output telling the
-# operator to insert the release notes manually.
+# the CHANGELOG.md file starting at a heading that looks like "## <tag>", and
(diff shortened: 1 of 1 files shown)


release/release.sh
```diff
@@ -130,31 +130,19 @@ function run() {
...
 
 # Outputs the release notes for the given tag. Looks for the release notes in
-# the CHANGELOG.md file starting at a heading that looks like "## <tag>". For
-# example `get_release notes v0.6.0` would look for a line like "## v0.6.0". If
-# no release notes could be found, a default string is output telling the
-# operator to insert the release notes manually.
+# the CHANGELOG.md file starting at a heading that looks like "## <tag>", and
+# ending at the next heading that looks like `##` with a version number. For
+# example `get_release_notes v0.6.0` would look for a line like "## v0.6.0" and
+# stop at "## v0.5.0".
```
devbww, line 136: Presumes a little about the shape of `CHANGELOG.md`, but OK.

devbww (approved):

PR #7825, opened 2022-01-06
fix(generator): handle empty method signatures

### LINKED ISSUE #7814: Generator needs to handle empty method signatures
For some reason, this exists:
https://github.com/googleapis/googleapis/blob/1236b08b701363015cbc876a850b60a061f23bb3/google/cloud/billing/v1/cloud_billing.proto#L55

Generator currently crashes because the `parameter_descriptor` is null in this case: https://github.com/googleapis/google-cloud-cpp/blob/a36bcfcb5e80ff9bc4ed9111fc18b4c89dd1d02e/generator/internal/descriptor_utils.cc#L186-L188

dbolduc:
Fixes #7814 

If we are given an empty parameter list, generate a client method that only takes `Options`.

I tried to break the change up into useful commits for the reviewer. They are:
1. logic changes
2. output for the generated client code
3. all other generated code (noise)
4. update a mock, which I forgot to do in the first commit.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7825)

### DIFF
--- a/generator/integration_tests/generator_integration_test.cc
+++ b/generator/integration_tests/generator_integration_test.cc
@@ -97,7 +97,7 @@ class GeneratorIntegrationTest : public testing::TestWithParam<std::string> {
                           "Codegen C++ Generator");
 
     product_path_ = "generator/integration_tests/golden/";
-    copyright_year_ = CurrentCopyrightYear();
+    copyright_year_ = "2022";
     omit_rpc1_ = "Omitted1";
     omit_rpc2_ = "GoldenKitchenSink.Omitted2";
     service_endpoint_env_var_ = "GOLDEN_KITCHEN_SINK_ENDPOINT";
--- a/generator/integration_tests/generator_integration_test.cc
+++ b/generator/integration_tests/generator_integration_test.cc
@@ -97,9 +97,7 @@ class GeneratorIntegrationTest : public testing::TestWithParam<std::string> {
                           "Codegen C++ Generator");
 
     product_path_ = "generator/integration_tests/golden/";
-    // The golden files were last generated on 2021, update next time you
-    // re-generate said files.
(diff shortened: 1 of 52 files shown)


generator/integration_tests/generator_integration_test.cc
```diff
@@ -97,9 +97,7 @@ class GeneratorIntegrationTest : public testing::TestWithParam<std::string> {
                           "Codegen C++ Generator");
 
     product_path_ = "generator/integration_tests/golden/";
-    // The golden files were last generated on 2021, update next time you
-    // re-generate said files.
-    copyright_year_ = "2021";
+    copyright_year_ = CurrentCopyrightYear();
```
coryan, line 100: I would rather you did 2022, that way we will not have the same breaks on 2023-01-01 we had in 2022-01-01.

Frankly, I would rather we pinned these to a specific year, but since I don't know how to do that ...

generator/integration_tests/test.proto
```diff
@@ -63,6 +63,14 @@ service GoldenThingAdmin {
     option (google.api.method_signature) = "parent";
   }
 
+  // Does Nothing.
```
coryan, line 66: if this was the las method you would have fewer spurious changes due to line numbers...  Just saying.

coryan (changes requested):

generator/integration_tests/generator_integration_test.cc
dbolduc: Done.

generator/integration_tests/test.proto
dbolduc: You're right. I moved it into the last `service`.
(less overall changes, but half the line numbers changed in the latest commit 🤷‍♂️)

PR #7833, opened 2022-01-06
feat(generator): add options to disable client class

### LINKED ISSUE #5935: generator: add option to exclude Client class generation
In order to better support services in which we only want to generate the stack from the Connection down, provide a mechanism to invoke the microgenerator such that no Client classes are emitted.

Alternately, we could continue to have the microgenerator emit the Client files for such services and just ignore them.

coryan:
Fixes #5935

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7833)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -641,11 +641,8 @@ std::vector<std::unique_ptr<GeneratorInterface>> MakeGenerators(
     std::vector<std::pair<std::string, std::string>> const& vars) {
   std::vector<std::unique_ptr<GeneratorInterface>> code_generators;
   VarsDictionary service_vars = CreateServiceVars(*service, vars);
-  bool omit_client = false;
-  for (auto const& v : vars) {
-    if (v.first == "omit_client" && v.second == "true") omit_client = true;
-  }
-  if (!omit_client) {
+  auto found = service_vars.find("omit_client");
+  if (found == service_vars.end() || found->second != "true") {
     code_generators.push_back(absl::make_unique<ClientGenerator>(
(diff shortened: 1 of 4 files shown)


generator/internal/descriptor_utils.cc
```diff
@@ -641,12 +641,18 @@ std::vector<std::unique_ptr<GeneratorInterface>> MakeGenerators(
     std::vector<std::pair<std::string, std::string>> const& vars) {
   std::vector<std::unique_ptr<GeneratorInterface>> code_generators;
   VarsDictionary service_vars = CreateServiceVars(*service, vars);
+  bool omit_client = false;
+  for (auto const& v : vars) {
```
devbww, line 645: Should you look in `service_vars` instead (and then use `find()`)?

generator/standalone_main.cc
```diff
@@ -209,6 +209,9 @@ int main(int argc, char** argv) {
     for (auto const& retry_code : service.retryable_status_codes()) {
       args.emplace_back("--cpp_codegen_opt=retry_status_code=" + retry_code);
     }
+    if (service.omit_client()) {
+      args.emplace_back("--cpp_codegen_opt=omit_client=true");
```
devbww, line 213: Opt: You may want to add a case to `generator/internal/codegen_utils_test.cc`.

generator/internal/descriptor_utils.cc
coryan: Done, for whatever reason my previous attempt at this failed.

generator/standalone_main.cc
coryan: Done.

devbww (approved):

PR #7839, opened 2022-01-07
impl(common): add an (internal) option to setup grpc::ClientContext

dbolduc:
Add ability to configure the `grpc::ClientContext` before each RPC call. For now, this has been developed in an `internal` namespace. The options may never be made public, but I figured it wouldn't hurt to document them anyway.
 
Part of the work for #7529. [This comment](https://github.com/googleapis/google-cloud-cpp/issues/7529#issuecomment-1006772608) is worth reading to understand the context of this change. Especially the **Downside** section, which could invalidate the entire PR.

Some Notes/Choices:
* `GrpcSetupPollOption` takes effect on cancel requests. (Even though there is no equivalent function in Bigtable). I could use a second opinion on whether the option should apply to both cancels and polls, or just polls.
* There is no configuration done on the context used when we [generate access tokens](https://github.com/googleapis/google-cloud-cpp/blob/be1bc0a9de3df833ecce0602f86f364ae8d1e384/google/cloud/internal/grpc_impersonate_service_account.cc#L46).
* [`AsyncRetryUnaryRpc`](https://github.com/googleapis/google-cloud-cpp/blob/be1bc0a9de3df833ecce0602f86f364ae8d1e384/google/cloud/internal/async_retry_unary_rpc.h#L150) does not configure the context. All calls to this class will be replaced by `AsyncRetryLoop`, which does configure the context. (So such a change is unnecessary)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7839)

### DIFF
--- a/google/cloud/grpc_options.cc
+++ b/google/cloud/grpc_options.cc
@@ -22,6 +22,18 @@ namespace cloud {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
 
+void ConfigureContext(grpc::ClientContext& context, Options const& opts) {
+  if (opts.has<GrpcSetupOption>()) {
+    opts.get<GrpcSetupOption>()(context);
+  }
+}
+
+void ConfigurePollContext(grpc::ClientContext& context, Options const& opts) {
+  if (opts.has<GrpcSetupPollOption>()) {
+    opts.get<GrpcSetupPollOption>()(context);
+  }
+}
+
 grpc::ChannelArguments MakeChannelArguments(Options const& opts) {
   auto channel_arguments = opts.get<GrpcChannelArgumentsNativeOption>();
   for (auto const& p : opts.get<GrpcChannelArgumentsOption>()) {
--- a/google/cloud/grpc_options.h
+++ b/google/cloud/grpc_options.h
@@ -157,6 +157,42 @@ using GrpcOptionList =
 
 namespace internal {
 
+/**
+ * A generic function to directly configure a gRPC context.
+ *
+ * This function takes effect before the context is used to make any requests.
+ *
+ * @warning It is NOT recommended to call `set_auth_context()` or
+ *     `set_credentials()` directly on the context. Instead, use the Google
+ *     Unified Auth Credentials library, via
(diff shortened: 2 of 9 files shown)


google/cloud/grpc_options.cc
```diff
@@ -22,6 +22,18 @@ namespace cloud {
...
 
+void ConfigureContext(grpc::ClientContext& context, Options const& opts) {
+  if (opts.has<GrpcSetupOption>()) {
+    opts.get<GrpcSetupOption>()(context);
+  }
+}
+
+void ConfigurePollContext(grpc::ClientContext& context, Options const& opts) {
+  if (opts.has<GrpcSetupPollOption>()) {
+    opts.get<GrpcSetupPollOption>()(context);
```
coryan, line 33: Consider:

```cc
void F() {
  grpc::ClientContext context;
  std::function<void(grpc::ClientContext&)> crashme;
  google::cloud::internal::ConfigureContext(
    context, google::cloud::Options{}.set<google::cloud::GrpcSetupOption>(crashme);
}
```

Do you want to check that the `std::function<>` is non-empty or something?

dbolduc: (I did not think about this). Umm, I could add a check... but we do not check that a custom retry policy is not `nullptr` before using it either. ([1](https://github.com/googleapis/google-cloud-cpp/blob/33ce1b369bb37e7a97326c06fdf7c61d0933bfb5/google/cloud/bigquery/internal/bigquery_read_option_defaults.cc#L57)) ([2](https://github.com/googleapis/google-cloud-cpp/blob/33ce1b369bb37e7a97326c06fdf7c61d0933bfb5/google/cloud/bigquery/bigquery_read_connection.cc#L71))

Should we think of passing an empty `std::function` as any different than passing a null pointer?

We could start checking for both....

dbolduc: Whoops, I did not answer your question, I just said it back at you. 

My answer is a weak no, we do not want to check that it is not empty.

coryan: SGTM.

coryan (approved):

PR #7844, opened 2022-01-07
fix: add mock library aliases

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7844)


dbolduc (changes requested):
I think you are missing `accessapproval`

coryan:
> I think you are missing `accessapproval`

Good catch, fixed.

dbolduc (approved):

PR #7852, opened 2022-01-08
fix(generator): long-running ops and routing header

coryan:
Fix the generated code for long-running operations without a routing
header. It was calling `Foo()` when it needed to call `**Async**Foo()`

The interesting change is in `generator/internal/metadata_decorator_generator.cc`, the rest is just how testing for this stuff works.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7852)

### DIFF
--- a/generator/integration_tests/golden/golden_kitchen_sink_client.h
+++ b/generator/integration_tests/golden/golden_kitchen_sink_client.h
@@ -116,10 +116,10 @@ class GoldenKitchenSinkClient {
   ///  not specified, the token's lifetime will be set to a default value of one
   ///  hour.
   /// @param options  Optional. Operation options.
-  /// @return @googleapis_link{google::test::admin::database::v1::GenerateAccessTokenResponse,generator/integration_tests/test.proto#L880}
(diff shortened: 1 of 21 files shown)


dbolduc (approved):
I disabled auto-merge.

optionally, you could add a test to: [golden_kitchen_sink_metadata_decorator_test.cc](https://github.com/googleapis/google-cloud-cpp/blob/18742f8ee21676edc99049fa2290b136b97c9e00/generator/integration_tests/golden/tests/golden_kitchen_sink_metadata_decorator_test.cc)

... but the code compiling is probably good enough.

devbww (approved):

coryan:
> I disabled auto-merge.
> 
> optionally, you could add a test to: [golden_kitchen_sink_metadata_decorator_test.cc](https://github.com/googleapis/google-cloud-cpp/blob/18742f8ee21676edc99049fa2290b136b97c9e00/generator/integration_tests/golden/tests/golden_kitchen_sink_metadata_decorator_test.cc)
> 
> ... but the code compiling is probably good enough.

I added a test, but I think the right place was `golden_thing_admin_metadata_decorator_test.cc`.

Issue #7853, opened 2022-01-08
`GenericPollingPolicy` clones its current state, not its initial state


dbolduc:
I noticed this for `bigtable::GenericPollingPolicy`. But it's also true of the common policy:
https://github.com/googleapis/google-cloud-cpp/blob/18742f8ee21676edc99049fa2290b136b97c9e00/google/cloud/polling_policy.h#L84-L86

Also, there are no tests for the common polling policy... We should probably add some...

PR #7854, opened 2022-01-08
fix(bigtable): polling policy clones initial state

dbolduc:
The bigtable part of #7853

`bigtable::GenericPollingPolicy<>` was cloning its current state, instead of its initial state. The idea is just to initially clone the objects. The current state of the polling policy is maintained by those clones. When the polling policy itself is cloned, it uses the initial objects to copy its initial state.

Also add a test for `WaitPeriod` to get better coverage.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7854)

### DIFF
--- a/google/cloud/bigtable/polling_policy.h
+++ b/google/cloud/bigtable/polling_policy.h
@@ -107,27 +107,27 @@ class PollingPolicy {
 template <typename Retry = LimitedTimeRetryPolicy,
           typename Backoff = ExponentialBackoffPolicy>
 class GenericPollingPolicy : public PollingPolicy {
-  static_assert(std::is_convertible<Retry*, RPCRetryPolicy*>::value,
+  static_assert(std::is_base_of<RPCRetryPolicy, Retry>::value,
                 "Retry must derive from RPCRetryPolicy");
-  static_assert(std::is_convertible<Backoff*, RPCBackoffPolicy*>::value,
+  static_assert(std::is_base_of<RPCBackoffPolicy, Backoff>::value,
                 "Backoff must derive from RPCBackoffPolicy");
 
  public:
   explicit GenericPollingPolicy(internal::RPCPolicyParameters defaults)
-      : rpc_retry_policy_(Retry(defaults)),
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/polling_policy.h
```diff
@@ -107,40 +107,54 @@ class PollingPolicy {
 template <typename Retry = LimitedTimeRetryPolicy,
           typename Backoff = ExponentialBackoffPolicy>
 class GenericPollingPolicy : public PollingPolicy {
+  static_assert(std::is_convertible<Retry*, RPCRetryPolicy*>::value,
```
coryan, line 110: Consider: https://en.cppreference.com/w/cpp/types/is_base_of

google/cloud/bigtable/polling_policy.h
```diff
@@ -107,40 +107,54 @@ class PollingPolicy {
...
+                "Retry must derive from RPCRetryPolicy");
+  static_assert(std::is_convertible<Backoff*, RPCBackoffPolicy*>::value,
+                "Backoff must derive from RPCBackoffPolicy");
+
  public:
   explicit GenericPollingPolicy(internal::RPCPolicyParameters defaults)
       : rpc_retry_policy_(Retry(defaults)),
-        rpc_backoff_policy_(Backoff(defaults)) {}
+        rpc_backoff_policy_(Backoff(defaults)),
```
coryan, line 118: Consider renaming the old variables to `rpc_retry_prototype_` and `rpc_backoff_prototype_` following:

https://en.wikipedia.org/wiki/Prototype_pattern

google/cloud/bigtable/polling_policy.h
dbolduc: Done.

google/cloud/bigtable/polling_policy.h
dbolduc: Done.

coryan (approved):

PR #7864, opened 2022-01-10
impl(common): Create thread annotations which work with libc++

dpcollins-google:
These annotations are no-ops when not building with a libc++ version that has thread annotations.

Also enable -Wthread-safety and libc++ annotations in ubsan mode

Also annotate all classes which previously used comments for this, and change them to pass this linter.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7864)


coryan:
I think I will take a pass on this.  Not see the value and I do not need to learn another non-standard-C++ version.

coryan:
Slightly more canonical commit / description.

devjgm:
@dpcollins-google offline, you said that this PR was the result of an error you got from an earlier PR. Can you link to that error so we can understand what this is fixing?

Issue #7835, opened 2022-01-06
Incorrect handling of resumable upload responses without a range header


coryan:
The client library treats 308 responses as an unrecoverable error:

https://github.com/googleapis/google-cloud-cpp/blob/49b5cc2423bbbb6b763f07dccf8b9aee3a04524c/google/cloud/storage/internal/retry_resumable_upload_session.cc#L76

which can be triggered from here:

https://github.com/googleapis/google-cloud-cpp/blob/49b5cc2423bbbb6b763f07dccf8b9aee3a04524c/google/cloud/storage/internal/resumable_upload_session.cc#L50-L57

However the spec allows for an empty range header:

> If Cloud Storage has not yet persisted any bytes, the 308 response does **not have a Range header**. In this case, you should start your upload from the beginning.

https://cloud.google.com/storage/docs/performing-resumable-uploads

PR #7868, opened 2022-01-10
cleanup(storage)!: uploads track committed_size

### LINKED ISSUE #6880: Resumable uploads have confusing account for "last committed byte"
In the REST-based resumable uploads we define the "last committed byte" as the value returned by the `Range` header of a successful chunk upload. This is the 1-based index of the highest-byte committed by the service.   We use `0` to represent "nothing committed yet". And it is trivial to make a "off by one error" when converting to C++ style ranges and 0-based indexing.

In the GRPC-based resumable uploads the service returns the "committed_size", that is the number of bytes already written. 

I think it would be cleaner to use this definition everywhere.

coryan:
Fixing how the `storage` client library deals with reponses lacking a
`Range` header will break any existing unit tests that mock a
`ResumableUploadSession`. This creates an opportunity to cleanup the
semantics of `ResumableUploadResponse`. This change performs that
cleanup.

**BREAKING CHANGE:** with this PR any use of the
`storage::internal::ResumableUploadResponse` type require changes.
Applications should have little need for this type, outside mocks, so
the changes should not affect production code.

Nevertheless, we apologize for the inconvenience, and while we would have
preferred to avoid breaking changes, it was inevitable to introduce some
breaking changes to fix a data loss bug.

If you are affected by this change, we expect that any existing tests
will fail to compile. This was an intentional choice, the semantics of
some values are changing, we think it is better to have these issues
detected at compile-time rather than during test execution.

We have updated the examples to use the new mocks, and the release notes
include more details on how to change any affected tests.

Fixes #6880, part of the work for #7835

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,51 @@
 
 ## v1.36.0 - TBD
 
+### [Storage](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/storage/README.md)
+
+**BREAKING CHANGE:** with this release any use of the
+`storage::internal::ResumableUploadResponse` type require changes. Applications
+should have little need for this type, outside mocks, so the changes should not
+affect production code.
+
+Nevertheless, we apologize for the inconvenience, and while we would have
+preferred to avoid breaking changes, it was inevitable to introduce some
+breaking changes to fix a data loss bug.
+
+If you are affected by this change, you will need to change your tests following
+this guide. Any place where you return a `ResumableUploadResponse` needs to
+change from:
+
+```cc
+storage::internal::ResumableUploadResource{
+  /*upload_session_url=*/std::string{"typically-unused"},
+  /*last_committed_byte=*/std::uint64_t{value},
+  /*payload=*/absl::nullopt, // or some gcs::ObjectMetadata value
+  /*upload_state=*/ResumableUploadResponse::kInProgress, // or kDone
+  /*annotations=*/std::string{"typically-unused"}
+}
+```
+
+to:
+
+```cc
+storage::internal::ResumableUploadResource{
+  /*upload_session_url=*/std::string{"typically-unused"},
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -90,6 +90,51 @@
...
+storage::internal::ResumableUploadResource{
+  /*upload_session_url=*/std::string{"typically-unused"},
+  /*upload_state=*/ResumableUploadResponse::kInProgress, // or kDone
+  /*committed_size=*/value + 1, // or absl::nullopt
+  /*payload=*/absl::nullopt, // or some gcs::ObjectMetadata value
+  /*annotations=*/std::string{"typically-unused"}
+}
+```
+
+That is, you need to re-order the fields **and** change increase the
```
devbww, line 130: s/change increase/increase/?

google/cloud/storage/internal/resumable_upload_session.cc
```diff
@@ -24,6 +24,33 @@ namespace cloud {
 namespace storage {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
+namespace {
+StatusOr<std::uint64_t> ParseRangeHeader(std::string const& range) {
+  // We expect a `Range:` header in the format described here:
+  //    https://cloud.google.com/storage/docs/json_api/v1/how-tos/resumable-upload
+  // that is the value should match `bytes=0-[0-9]+`:
+
```
devbww, line 32: Opt nit: Remove blank line?

devbww (approved):

CHANGELOG.md
coryan: Done.

google/cloud/storage/internal/resumable_upload_session.cc
coryan: Done.

PR #7877, opened 2022-01-11
fix(storage)!: handle missing range header in uploads

### LINKED ISSUE #7835: Incorrect handling of resumable upload responses without a range header

coryan:
Resumable upload responses use the `Range` header to report how many
bytes were committed. Sometimes, maybe once every few million uploads,
the header is missing. This indicates that nothing was committed.
The library was treating this as a permanent error. The correct behavior
is to query the status of the upload and try again.

Fixes #7835

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7877)

### DIFF
--- a/google/cloud/storage/internal/retry_resumable_upload_session.cc
+++ b/google/cloud/storage/internal/retry_resumable_upload_session.cc
@@ -62,20 +62,20 @@ StatusOr<ResumableUploadResponse> RetryResumableUploadSession::UploadFinalChunk(
 // A description of resumable uploads can be found at:
 //     https://cloud.google.com/storage/docs/performing-resumable-uploads
 //
-// A description of the gRPC analog can be found in the proto file:
+// A description of the gRPC analog can be found in the proto file. Pay
+// particular attention to the documentation for `WriteObject()`,
+// `WriteObjectRequest`, `StartResumablewrite()` and `QueryResumableWrite()`:
 //    https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto
 //
-// Particularly the documentation for `WriteObject()`, `WriteObjectRequest`,
-// `StartResumablewrite()` and `QueryResumableWrite()`.
-//
 // At a high level one starts a resumable upload by creating a "session". These
-// sessions are persistent, they survive disconnections from the service, one
+// sessions are persistent (they survive disconnections from the service). One
 // can even resume uploads after shutting down and restarting an application.
 // Their current state can be queried using a simple RPC (or a PUT request
 // without payload).
 //
-// Resumable uploads make progress by sending "chunks", a single PUT request in
-// REST-based transports, a client-side streaming RPC for gRPC-based transports.
+// Resumable uploads make progress by sending "chunks", either a single PUT
+// request in REST-based transports, or a client-side streaming RPC for
+// gRPC-based transports.
 //
 // Resumable uploads complete when the application sends the last bytes of the
 // object. In the client library we mostly start uploads without knowing the
@@ -152,14 +152,15 @@ RetryResumableUploadSession::UploadGenericChunk(char const* caller,
       continue;
     }
 
-    // Sometimes the upload completes in a regular `UploadChunk()` or event a
+    // While normally a `UploadFinalChunk()` call completes an upload, sometimes
+    // the upload can complete in a regular `UploadChunk()` or a
     // `ResetSession()` call. For example, the server can detect a completed
     // upload "early" if the application includes the X-Upload-Content-Length`
     // header.
     if (result->upload_state == ResumableUploadResponse::kDone) return result;
 
     // This indicates that the response was missing a `Range:` header, or that
-    // the range header was in the wrong format, either way, treat that as a
+    // the range header was in the wrong format. Either way, treat that as a
     // (transient) failure and start over.
     if (!result->committed_size.has_value()) {
       if (operation != &reset) {
--- a/google/cloud/storage/internal/retry_resumable_upload_session.cc
+++ b/google/cloud/storage/internal/retry_resumable_upload_session.cc
@@ -24,7 +24,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
 
 namespace {
-StatusOr<ResumableUploadResponse> ReturnError(Status&& last_status,
+StatusOr<ResumableUploadResponse> ReturnError(Status const& last_status,
                                               RetryPolicy const& retry_policy,
                                               char const* error_message) {
   std::ostringstream os;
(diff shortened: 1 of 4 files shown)


google/cloud/storage/internal/retry_resumable_upload_session.cc
```diff
@@ -57,112 +57,164 @@ StatusOr<ResumableUploadResponse> RetryResumableUploadSession::UploadFinalChunk(
...
+      auto delay = backoff_policy->OnCompletion();
+      std::this_thread::sleep_for(delay);
+      operation = &reset;
       continue;
     }
-    last_status = std::move(result).status();
-    if (!retry_policy->OnFailure(last_status)) {
-      return ReturnError(std::move(last_status), *retry_policy, __func__);
+
+    // Sometimes the upload completes in a regular `UploadChunk()` or event a
```
devbww, line 155: "or event a"?

google/cloud/storage/internal/retry_resumable_upload_session.cc
```diff
@@ -57,112 +57,164 @@ StatusOr<ResumableUploadResponse> RetryResumableUploadSession::UploadFinalChunk(
...
+    // `ResetSession()` call. For example, the server can detect a completed
+    // upload "early" if the application includes the X-Upload-Content-Length`
+    // header.
+    if (result->upload_state == ResumableUploadResponse::kDone) return result;
+
+    // This indicates that the response was missing a `Range:` header, or that
+    // the range header was in the wrong format, either way, treat that as a
```
devbww, line 162: s/, either way/. Either way/

google/cloud/storage/internal/retry_resumable_upload_session.cc
```diff
@@ -57,112 +57,164 @@ StatusOr<ResumableUploadResponse> RetryResumableUploadSession::UploadFinalChunk(
...
+// Implements the retry loop for a resumable upload session.
+//
+// A description of resumable uploads can be found at:
+//     https://cloud.google.com/storage/docs/performing-resumable-uploads
+//
+// A description of the gRPC analog can be found in the proto file:
+//    https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto
+//
+// Particularly the documentation for `WriteObject()`, `WriteObjectRequest`,
+// `StartResumablewrite()` and `QueryResumableWrite()`.
```
devbww, line 69: Perhaps this would be better before the colon in the previous block.

google/cloud/storage/internal/retry_resumable_upload_session.cc
```diff
@@ -57,112 +57,164 @@ StatusOr<ResumableUploadResponse> RetryResumableUploadSession::UploadFinalChunk(
...
+//
+// At a high level one starts a resumable upload by creating a "session". These
+// sessions are persistent, they survive disconnections from the service, one
```
devbww, line 72: Perhaps ...

_These sessions are persistent (they survive disconnections from the service).  One can even ..._

google/cloud/storage/internal/retry_resumable_upload_session.cc
```diff
@@ -57,112 +57,164 @@ StatusOr<ResumableUploadResponse> RetryResumableUploadSession::UploadFinalChunk(
...
+// can even resume uploads after shutting down and restarting an application.
+// Their current state can be queried using a simple RPC (or a PUT request
+// without payload).
+//
+// Resumable uploads make progress by sending "chunks", a single PUT request in
```
devbww, line 77: Perhaps ...

_Resumable uploads make progress by sending "chunks", either a ..., or a ..._

if I understand correctly.

google/cloud/storage/internal/retry_resumable_upload_session_test.cc
```diff
@@ -597,67 +557,163 @@ TEST_F(RetryResumableUploadSessionTest, HandleTransientPartialFailures) {
...
+  ::testing::InSequence sequence;
+  // Simulate a successful upload to have the number of committed bytes be
+  // non-zero.
+  EXPECT_CALL(*mock, UploadChunk).WillOnce([&](ConstBufferSequence const&) {
+    return make_status_or(ResumableUploadResponse{
+        "", ResumableUploadResponse::kInProgress, quantum, {}, {}});
+  });
+
+  // Simulate an upload that "succeeds", but returns a missing value for the
+  // committed size, this should trigger a reset.
```
devbww, line 588: s/size, this/size. This/

google/cloud/storage/internal/retry_resumable_upload_session.cc
coryan: Fixed I hope.

google/cloud/storage/internal/retry_resumable_upload_session.cc
coryan: Fixed.

google/cloud/storage/internal/retry_resumable_upload_session.cc
coryan: Done. And thanks.

google/cloud/storage/internal/retry_resumable_upload_session.cc
coryan: Fixed.

google/cloud/storage/internal/retry_resumable_upload_session.cc
coryan: Thanks.

google/cloud/storage/internal/retry_resumable_upload_session_test.cc
coryan: Fixed.

coryan:
PTAL.

devbww (approved):

Jseph:
___
*[google/cloud/storage/internal/retry_resumable_upload_session.cc, line 140 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7877#-MtAKYAz5IBP6C2b6OND:-MtAKYAz5IBP6C2b6ONE:b-3tfc4j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e45022d3a3b2f377acc9cb6d61bc85adf3567df7/google/cloud/storage/internal/retry_resumable_upload_session.cc#L140)):*
> ```cpp
>     auto const pre_operation_committed_size = committed_size_;
>     auto result = (*operation)(buffers);
>     AppendDebug("upload", static_cast<std::uint64_t>(result.status().code()));
> ```

This is potentially misleading because this may be the result of an upload or a reset.

Jseph:
___
*[google/cloud/storage/internal/retry_resumable_upload_session.cc, line 164 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7877#-MtAKvfI7ritFBpmL7Cy:-MtAKvfI7ritFBpmL7Cz:b-wr5223) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e45022d3a3b2f377acc9cb6d61bc85adf3567df7/google/cloud/storage/internal/retry_resumable_upload_session.cc#L164)):*
> ```cpp
>     // This indicates that the response was missing a `Range:` header, or that
>     // the range header was in the wrong format. Either way, treat that as a
>     // (transient) failure and start over.
> ```

Nit: The combination of `start over` and the reset call below led me to believe that the upload is being restarted from the beginning. This isn't the case, the upload session is just querying GCS for its current state.



_Code quote:_
```cpp
start over
```

coryan:
___
*[google/cloud/storage/internal/retry_resumable_upload_session.cc, line 140 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7877#-MtAKYAz5IBP6C2b6OND:-MtALl0FCeJXs969aWQ4:bmq8xca) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e45022d3a3b2f377acc9cb6d61bc85adf3567df7/google/cloud/storage/internal/retry_resumable_upload_session.cc#L140)):*
<details><summary><i>Previously, Jseph wrote…</i></summary><blockquote>

This is potentially misleading because this may be the result of an upload or a reset.
</blockquote></details>

True.  I was (and still am) tempted to remove these debugging statements.  They are a drag and they served their purpose.  Thoughts?

coryan:
___
*[google/cloud/storage/internal/retry_resumable_upload_session.cc, line 164 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7877#-MtAKvfI7ritFBpmL7Cy:-MtAM2x92nNkuBdSq05I:b-c0m60c) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e45022d3a3b2f377acc9cb6d61bc85adf3567df7/google/cloud/storage/internal/retry_resumable_upload_session.cc#L164)):*
<details><summary><i>Previously, Jseph wrote…</i></summary><blockquote>

Nit: The combination of `start over` and the reset call below led me to believe that the upload is being restarted from the beginning. This isn't the case, the upload session is just querying GCS for its current state.
</blockquote></details>

Will send a PR to clarify that.

Jseph:
___
*[google/cloud/storage/internal/retry_resumable_upload_session.cc, line 192 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7877#-MtAMNSUEMZ2pjCFjR82:-MtAMNSUEMZ2pjCFjR83:b7mu884) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e45022d3a3b2f377acc9cb6d61bc85adf3567df7/google/cloud/storage/internal/retry_resumable_upload_session.cc#L192)):*
> ```cpp
> 
>     // On a full write we can return immediately.
>     if (written == size) return result;
> ```

It's also possible in the case of racing processes uploading to the same session that `written > size` in which case I think the thing to do is also return result?  That should cause the object_write_streambuf to go into a bad state.  Otherwise I think you get stuck in a loop trying to do an upload with 0 bytes until the retries are exhausted.



_Code quote:_
```cpp
if (written == size) return result;
```

Jseph:
___
*[google/cloud/storage/internal/retry_resumable_upload_session.cc, line 140 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7877#-MtAKYAz5IBP6C2b6OND:-MtAWVIVBsZ_DZPjhC03:b-hy8kp7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e45022d3a3b2f377acc9cb6d61bc85adf3567df7/google/cloud/storage/internal/retry_resumable_upload_session.cc#L140)):*

True.  I was (and still am) tempted to remove these debugging statements.  They are a drag and they served their purpose.  Thoughts?
</blockquote></details>

Yeah seems like the error code doesn't matter to the user (retries are successful) or they are returned anyway as the result of this function.

PR #7884, opened 2022-01-12
impl: add HttpPayload class

scotthart:
Skeleton of CurlImpl added with minimal implementation necessary for HttpPayload to compile.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7884)

### DIFF
--- a/google/cloud/internal/curl_impl.h
+++ b/google/cloud/internal/curl_impl.h
@@ -1,0 +1,90 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_CURL_IMPL_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_CURL_IMPL_H
+
+#include "google/cloud/internal/curl_handle.h"
+#include "google/cloud/internal/curl_handle_factory.h"
+#include "google/cloud/internal/curl_wrappers.h"
+#include "google/cloud/internal/rest_request.h"
+#include "google/cloud/internal/rest_response.h"
+#include "google/cloud/options.h"
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include <utility>
+
+namespace google {
+namespace cloud {
+namespace rest_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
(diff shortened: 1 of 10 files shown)


google/cloud/internal/curl_impl.h
```diff
@@ -0,0 +1,90 @@
...
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include <utility>
+
+namespace google {
+namespace cloud {
+namespace rest_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// This class encapsulates use of libcurl and manages all the necessary state
```
coryan, line 33: Looks like this is a placeholder so you can have an interface to code to while making progress on other things?  Maybe a TODO and a bug would be nice if that is the case.

google/cloud/internal/curl_impl.h
```diff
@@ -0,0 +1,90 @@
...
+// of a request and its associated response.
+class CurlImpl {
+ public:
+  enum class HttpMethod { kDelete, kGet, kPatch, kPost, kPut };
```
coryan, line 37: Does this need to be an enum?  Can we just use a string with maybe some constants?  If not, then I think we are missing a few:

https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods

google/cloud/internal/curl_impl.h
```diff
@@ -0,0 +1,90 @@
...
+  HttpStatusCode status_code() const { return http_code_; }
+  std::multimap<std::string, std::string> const& headers() const {
+    return headers_;
+  }
+
+  std::string const& url() const { return url_; }
+
+  // NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+  std::string MakeEscapedString(std::string const&) {
+    // TODO(sdhart): implement this function
```
coryan, line 61: nit: s/TODO(sdhart)/TODO(#github-issue)/

google/cloud/internal/curl_impl.h
```diff
@@ -0,0 +1,90 @@
...
+#include <utility>
+
+namespace google {
+namespace cloud {
+namespace rest_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// This class encapsulates use of libcurl and manages all the necessary state
+// of a request and its associated response.
+class CurlImpl {
```
coryan, line 35: If this encapsulates a request, maybe `CurlHttpRequest` would be a better name?

google/cloud/internal/http_payload.cc
```diff
@@ -0,0 +1,60 @@
...
+std::multimap<std::string, std::string> CurlHttpPayload::Trailers() const {
+  // TODO(sdhart): This currently returns all the previously available headers
+  // and any received trailers. Given that there should be a header listing
+  // what trailers to expect, this function could be improved such that it
+  // interrogates the "Trailer:" header and construct a map of only the
+  // trailers.
+  return impl_->headers();
+}
+
+StatusOr<std::string> ReadAll(std::unique_ptr<HttpPayload> payload,
```
coryan, line 42: If you are trying to add placeholders or an scaffold to build around, this function seems premature.

coryan (commented):
I am a bit loss, maybe we can chat tomorrow.  The description ("add HttpPayload") does not match the PR (which adds many more classes).  And it is unclear why those other classes need to be added at the same time (or why they are needed at all, why do we have CurlImpl and CurlHandle?  They have many common methods..).

PR #7913, opened 2022-01-12
fix(generator): trim spaces from parameter name before lookup

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7913)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -179,11 +179,11 @@ void SetMethodSignatureMethodVars(
   for (int i = 0; i < method_signature_extension.size(); ++i) {
     google::protobuf::Descriptor const* input_type = method.input_type();
     std::vector<std::string> parameters =
-        absl::StrSplit(method_signature_extension[i], ',');
+        absl::StrSplit(method_signature_extension[i], ',', absl::SkipEmpty());
     std::string method_signature;
     std::string method_request_setters;
     for (auto& parameter : parameters) {
-      if (parameter.empty()) continue;
+      absl::StripAsciiWhitespace(&parameter);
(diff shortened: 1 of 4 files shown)


coryan (approved):

generator/internal/descriptor_utils.cc
```diff
@@ -179,11 +179,11 @@ void SetMethodSignatureMethodVars(
   for (int i = 0; i < method_signature_extension.size(); ++i) {
     google::protobuf::Descriptor const* input_type = method.input_type();
     std::vector<std::string> parameters =
-        absl::StrSplit(method_signature_extension[i], ',');
+        absl::StrSplit(method_signature_extension[i], ',', absl::SkipEmpty());
```
devbww, line 182: `SkipEmpty` seems like something you wouldn't want to do in a comma-separated list of names.  Is that really OK?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7913)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[generator/internal/descriptor_utils.cc, line 182 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7913#-MtEuzXhxYNLcvI8G-7T-r1-182:-MtFCRFq8Xk51tFySo2O:b-l06c9t) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3d57c70f12ebd09a70edb7d21e96d26e4344414c/generator/internal/descriptor_utils.cc#L182)):*

`SkipEmpty` seems like something you wouldn't want to do in a comma-separated list of names.  Is that really OK?
</blockquote></details>

Yes. If they are empty it's a result of human error editing the proto annotation.

PR #7915, opened 2022-01-12
feat(eventarc): generate library

### LINKED ISSUE #7903: Eventarc API: `eventarc.googleapis.com`

devbww:
Fixes #7903.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7915)


external/googleapis/update_libraries.sh
```diff
@@ -49,6 +49,11 @@ declare -A -r LIBRARIES=(
       "@com_google_googleapis//google/cloud/clouddms/logging/v1:logging_proto"
   )"
   ["dialogflow"]="@com_google_googleapis//google/cloud/dialogflow/v2:dialogflow_proto"
+  ["eventarc"]="$(
+    printf ",%s" \
+      "@com_google_googleapis//google/cloud/eventarc/v1:eventarc_cc_grpc" \
```
coryan, line 54: No action needed: normally we use the `*_proto` (vs. `*_cc_grpc`) rules here.  Seems like both work, and `*_cc_grpc` are easier to find.  We may want to fix all these in a separate PR.

coryan (approved):
Sounds like the links required minimal manipulation.  🎉

external/googleapis/update_libraries.sh
devbww: Ack.

PR #7926, opened 2022-01-13
fix(generator): multiple method signatures doxygen

### LINKED ISSUE #7923: Generator does not document parameters for functions with multiple method signatures properly
Instead of adding the documentation for the single method signature in question, we loop over all method signatures. https://github.com/googleapis/google-cloud-cpp/blob/b2abf92792951c7f75ad34775113b865b5ec9af6/generator/internal/descriptor_utils.cc#L325

So when we have a proto like [this](https://github.com/googleapis/googleapis/blob/master/google/cloud/recommender/v1/recommender_service.proto#L118-L119), it leads to generated code like this: https://github.com/googleapis/google-cloud-cpp/blob/2de054cf38c42ccbded6d8bc0a666b70e15d5542/google/cloud/recommender/recommender_client.h#L226-L247

which breaks our `publish-docs` build.

dbolduc:
Fixes #7923 

Fix the generator so that the doxygen parameter comments are limited to the method signature in question (instead of all of them). I could have added an extra paremeter that was -1 for a protobuf request or N for the Nth method signature. But that seemed like bad code. So I took the opportunity to refactor the logic a bit, and clean up some things.

The diff is not great here. The old `FormatProtobufRequestParameters` was actually only two lines of code, so I inlined it. The other method was simplified a bit. Also, there is one less `enum` in the world.

In order to update the unit test so that it would have previously failed, I added a comment and method signature to the unit test proto. This explains why some of the line numbers in unrelated tests had to change.

I did not add an RPC with multiple method signatures to `generator/integration_tests/test.proto`. I think the unit test is sufficient.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7926)

### DIFF
--- a/generator/internal/client_generator.cc
+++ b/generator/internal/client_generator.cc
@@ -118,54 +118,43 @@ Status ClientGenerator::GenerateHeader() {
     auto method_signature_extension =
         method.options().GetRepeatedExtension(google::api::method_signature);
     for (int i = 0; i < method_signature_extension.size(); ++i) {
-      std::string method_string =
+      std::string const method_string =
           absl::StrCat("  $method_name$($method_signature", i,
                        "$Options options = {});\n\n");
+      std::string const signature = method_signature_extension[i];
       HeaderPrintMethod(
           method,
           {MethodPattern(
-               {
-                   {FormatMethodCommentsFromRpcComments(
-                       method, MethodParameterStyle::kApiMethodSignature)},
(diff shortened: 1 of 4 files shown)


devbww (approved):
Thanks.  I hadn't noticed that you had filed an issue, and was just beginning to work on this problem (which hit me for `google/cloud/iot`).

Issue #7795, opened 2022-01-04
Support bidir streaming in the generator


coryan:
The generator needs to support bidirectional streaming, at least for the `*Stub` and `*Connection` classes.  If the stream has a very "custom" resume protocol we will need some way to suppress the `*Client` class (see #5935) and hand-craft parts of the library.

PR #7933, opened 2022-01-13
feat(generator): support bidir streams in Connection

coryan:
Part of the work for #7795

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7933)

### DIFF
--- a/generator/internal/connection_generator.cc
+++ b/generator/internal/connection_generator.cc
@@ -56,6 +56,9 @@ Status ConnectionGenerator::GenerateHeader() {
        HasStreamingReadMethod() || HasPaginatedMethod()
            ? "google/cloud/stream_range.h"
            : "",
+       HasBidirStreamingMethod()
+           ? "google/cloud/internal/async_read_write_stream_impl.h"
+           : "",
        "google/cloud/version.h"});
   HeaderSystemIncludes(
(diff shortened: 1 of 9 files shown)


generator/internal/connection_generator.cc
```diff
@@ -357,6 +388,20 @@ Status ConnectionGenerator::GenerateCc() {
...
   for (auto const& method : methods()) {
+    if (IsBidirStreaming(method)) {
+      CcPrintMethod(
+          method, __FILE__, __LINE__,
+          R"""(  std::unique_ptr<::google::cloud::internal::AsyncStreamingReadWriteRpc<
+      $request_type$,
+      $response_type$>>
+  Async$method_name$() override {
+    auto cq = background_->cq();
+    return stub_->Async$method_name$(cq, absl::make_unique<grpc::ClientContext>());
```
devbww, line 399: Should we avoid giving the `CompletionQueue` a name so as to avoid an extra ref-count inc/dec on its `impl_`?

coryan: Maybe there is a cast or another trick that I am missing?

devbww: Ack.  (Aside: If we could change `Async$method_name$` to take its `CompletionQueue` by value, I would vote to do so.)

devbww (approved):

PR #7940, opened 2022-01-14
impl(generator): support additional proto files for lro return types

### LINKED ISSUE #7922: Support long running operations with not-imported metadata types
Some services use unimported message type names for the long running operation metadata and results, for example:

https://github.com/googleapis/googleapis/blob/8675f5ec8bf7573cd8329488a55df55566ee03e2/google/appengine/v1/appengine.proto#L997-L1004

That type is defined here:

https://github.com/googleapis/googleapis/blob/8675f5ec8bf7573cd8329488a55df55566ee03e2/google/appengine/v1/operation.proto#L32

Note that this proto is not `import`-ed into `appengine.proto`, also note that the name is not fully qualified.  For this code the generator uses the `metadata_type`, instead of the `result_type`, as the `future<StatusOr<T>>` returned by the operation. 

Possible fixes include:

- Just use `future<Status>`? Or `future<Status<google::cloud::Empty>>` which is trivial to convert to the previous type.
- Fully qualify the type name and trust that the `metadata_type` proto file was `imported`
- Fully qualify the type name and try to generate the right `include` (how? no idea).

scotthart:
Allows additional proto files to be specified in the config. The files are used for type lookups during code generation and to format include statements in the <service_name>_stub.h files. The proto files are also run through protoc in order to produce the accompanying pb.h and pb.cc files.

fixes #7922

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7940)

### DIFF
--- a/generator/generator_config.proto
+++ b/generator/generator_config.proto
@@ -53,6 +53,11 @@ message ServiceConfiguration {
   // does not provide good application-level concepts (consider chunking for
   // rows or values).
   bool omit_client = 11;
+
+  // If additional proto files are needed that aren't imported by the service
+  // file, add them using this field. This typically happens when an annotation
+  // lists a type that is not defined in any of the imported proto files.
+  repeated string additional_proto_files = 12;
 }
 
 message GeneratorConfiguration {
--- a/generator/integration_tests/generator_integration_test.cc
+++ b/generator/integration_tests/generator_integration_test.cc
@@ -107,6 +107,7 @@ class GeneratorIntegrationTest : public testing::TestWithParam<std::string> {
     retry_code1_ = "GoldenKitchenSink.kInternal";
     retry_code2_ = "kUnavailable";
     retry_code3_ = "GoldenThingAdmin.kDeadlineExceeded";
+    additional_proto_file_ = "generator/integration_tests/backup.proto";
 
     std::vector<std::string> args;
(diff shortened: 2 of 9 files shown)


coryan (commented):
Quick turnaround, thanks.

In my testing I found that the generator could not find the fully qualified name for a non-qualified name. For example, consider this method:

https://github.com/googleapis/googleapis/blob/04fe61ccc4c0e5a68a13849165e0c22a44ab2712/google/appengine/v1/appengine.proto#L177-L186

The generator (before your patches, without the necessary import) creates this code:

https://github.com/coryan/google-cloud-cpp/blob/56431557c79d44bfc1c758f3ed7cfa290000c293/google/cloud/appengine/services_connection.h#L64-L66

Note how `OperationMetadataV1` is not fully qualified.  The C++ compiler cannot find the type, without or without the `#include <.../appengine/v1/operation.pb.h>`.  Once I added the import, the generated code changed:

https://github.com/coryan/google-cloud-cpp/commit/370e514f46805210ab5c19b0716fbf041f57ba44

The full branch with all the changes is at:

https://github.com/coryan/google-cloud-cpp/commits/feat-appengine-generate-library

I do not see anything in your PR to fix the fact that the metadata annotation remains unqualified?  If you want, I can test your PR with the appengine code and see what it generates?

Issue #7929, opened 2022-01-13
Support streaming write operations in the generator


coryan:
I am not sure if there are streaming writes in any of the libraries that we plan to automatically generate. Seems likely.  But I need this feature to automatically generate the `*Stub` classes for the `storage/v2` APIs anyway.

coryan:
I am going to close this, as I think I have implemented all we need to for streaming-write RPCS.  There are only two services that require streaming writes (see below).

- `google.actions.sdk.v2` is not a cloud service, it is part of [Google Assistant](https://developers.google.com/assistant/actions/api/reference/rest)
- `google.storage.v2.WriteObject` requires a custom connection and client (which we have already implemented), generating the `*Stub` and decorators automatically is all we need. 
- `google.storage.v1.InsertObject` will never go GA 

```
cat googleapis/api-index-v1.json | jq '.apis | map(select(.services != null)) | map(.services | map(select(.methods != null)) | map(.methods | map(select(.mode == "CLIENT_STREAMING") | [.mode, .fullName]  ))) | flatten ' | grep -v CLIENT_STREAMING  | grep -v beta
[
  "google.actions.sdk.v2.ActionsSdk.CreateVersion",
  "google.actions.sdk.v2.ActionsSdk.WriteDraft",
  "google.actions.sdk.v2.ActionsSdk.WritePreview",
  "google.storage.v1.Storage.InsertObject",
  "google.storage.v2.Storage.WriteObject"
]
```

PR #7941, opened 2022-01-14
feat(generator): streaming-write RPCs for Stubs

coryan:
Change the generator to support streaming write (aka client-side
streaming) RPCs in the `*Stub` class and its decorators. Changed the
golden files to include one such method. Change the unit tests for the
golden files to test the generated code.

The generated libraries are unaffected by this change, as none of the
libraries we have uses streaming write RPCs.

Part of the work for #7929

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7941)

### DIFF
--- a/generator/integration_tests/golden/internal/golden_kitchen_sink_stub.h
+++ b/generator/integration_tests/golden/internal/golden_kitchen_sink_stub.h
@@ -21,6 +21,7 @@
 
 #include "google/cloud/internal/async_read_write_stream_impl.h"
 #include "google/cloud/internal/streaming_read_rpc.h"
+#include "google/cloud/internal/streaming_write_rpc.h"
 #include "google/cloud/status_or.h"
 #include "google/cloud/version.h"
 #include <generator/integration_tests/test.grpc.pb.h>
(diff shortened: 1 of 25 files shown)


generator/integration_tests/golden/internal/golden_kitchen_sink_stub.h
```diff
@@ -71,6 +72,12 @@ class GoldenKitchenSinkStub {
       google::cloud::CompletionQueue const& cq,
       std::unique_ptr<grpc::ClientContext> context) = 0;
 
+ virtual std::unique_ptr<::google::cloud::internal::StreamingWriteRpc<
+      google::test::admin::database::v1::WriteObjectRequest,
+      google::test::admin::database::v1::WriteObjectResponse>>
+  WriteObject(
+      std::unique_ptr<grpc::ClientContext> context) = 0;
+
```
devbww, line 80: Aside: It looks like we're generating a post-declaration separator here, when we don't know if we actually need one.

devbww (approved):

generator/integration_tests/golden/internal/golden_kitchen_sink_stub.h
coryan: Yes. Some of the generators use pre-declaration/definition separators, others use post.  I went with what was already there (at least I hope I did).  I agree that pre-declaration/definition separators are easier to reason about.

Issue #7670, opened 2021-11-29
Implement the Pub/Sub Lite client library


dpcollins-google:
**What component of `google-cloud-cpp` is this feature request for?**
pubsublite

**Is your feature request related to a problem? Please describe.**
pubsublite should be usable from c++

**Describe the solution you'd like**
A hand written client library for Pub/Sub Lite in C++ sharing the Cloud Pub/Sub API surface like it does in other languages.

dpcollins-google:
Not sure if this can be reassigned to me for now?

coryan:
I am going to close this because it is not specific enough.  Seems like we need to breakdown the remaining work in a series of  issues.  Completing this library is now part of our general backlog, fear not.

PR #7966, opened 2022-01-15
feat(pubsublite): add TopicStats service

coryan:
This service does not use any streaming RPCs, it is suitable for code
generation.

Part of the work for #7670

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7966)


google/cloud/pubsublite/CMakeLists.txt
```diff
@@ -85,7 +85,28 @@ add_library(
...
+    internal/topic_stats_metadata_decorator.cc
+    internal/topic_stats_metadata_decorator.h
+    internal/topic_stats_option_defaults.cc
+    internal/topic_stats_option_defaults.h
+    internal/topic_stats_retry_traits.h
+    internal/topic_stats_stub.cc
+    internal/topic_stats_stub.h
+    internal/topic_stats_stub_factory.cc
+    internal/topic_stats_stub_factory.h
+    mocks/mock_topic_stats_connection.h
```
dbolduc, line 102: this should be in the mock library

google/cloud/pubsublite/google_cloud_cpp_pubsublite.bzl
```diff
@@ -29,6 +29,18 @@ google_cloud_cpp_pubsublite_hdrs = [
...
     "internal/admin_stub_factory.h",
     "internal/stream_factory.h",
+    "internal/topic_stats_auth_decorator.h",
+    "internal/topic_stats_logging_decorator.h",
+    "internal/topic_stats_metadata_decorator.h",
+    "internal/topic_stats_option_defaults.h",
+    "internal/topic_stats_retry_traits.h",
+    "internal/topic_stats_stub.h",
+    "internal/topic_stats_stub_factory.h",
+    "mocks/mock_topic_stats_connection.h",
```
dbolduc, line 39: "

dbolduc (changes requested):
Also, do we have a code coverage plan? Right now we ignore all of `google/cloud/pubsublite`. But if there is hand written code, we will want to test that.

A not so great option is to move this and the Admin API into a subdirectory, then ignore that subdirectory. Food for thought.

google/cloud/pubsublite/CMakeLists.txt
coryan: Thanks for catching that, fixed.

coryan:
> Also, do we have a code coverage plan? Right now we ignore all of `google/cloud/pubsublite`. But if there is hand written code, we will want to test that.

Eventually we will need more complicated rules for this directory, I think maybe using this:

https://docs.codecov.com/docs/path-regexp

dbolduc (approved):
codecov regex SGTM

PR #7968, opened 2022-01-15
impl(bigtable): convert set of policies into options

### LINKED ISSUE #7529: Normalize Bigtable Policies

dbolduc:
Fixes #7529 

With this change, the pieces are finally in place to make the breaking change that implements `TableAdmin` / `InstanceAdmin` in terms of the generated admin clients.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7968)


google/cloud/bigtable/internal/convert_policies_test.cc
```diff
@@ -0,0 +1,190 @@
...
+    };
+    std::chrono::milliseconds OnCompletion(grpc::Status const&) override {
+      return {};
+    }
+
+   private:
+    bool* called_;
+  };
+
+  class PollingWithSetup : public PollingPolicy {
```
coryan, line 118: This is what mock classes were invented for?  You would need to mock `clone()` to return a mock that expects a single call to `Setup()`?

google/cloud/bigtable/internal/convert_policies.h
```diff
@@ -0,0 +1,72 @@
...
+#include "google/cloud/bigtable/rpc_backoff_policy.h"
+#include "google/cloud/bigtable/rpc_retry_policy.h"
+#include "google/cloud/options.h"
+
+namespace google {
+namespace cloud {
+namespace bigtable_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
```
coryan, line 28: Thanks for the detailed comment, it helped.

google/cloud/bigtable/internal/convert_policies_test.cc
dbolduc: D'oh. I redid the test with mocks, and it no longer reads like a logic puzzle.

Also, I split it up into two separate tests: one for retry + backoff; one for polling.

google/cloud/bigtable/internal/convert_policies.cc
```diff
@@ -0,0 +1,81 @@
...
+#include "google/cloud/grpc_options.h"
+
+namespace google {
+namespace cloud {
+namespace bigtable_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+Options MakeGrpcSetupOptions(
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
+    std::shared_ptr<bigtable::RPCRetryPolicy> retry,
```
coryan, line 27: I would have made these parameters `std::shared_ptr<T> const&`.  Your call.

coryan (approved):

Issue #7963, opened 2022-01-15
Implement all operations in StorageStub and its decorators


coryan:
I am planning to use the code generator for this, but it will require some refactor and probably will add functions slowly.

coryan:
I changed my mind. Instead of generating all the operations in one big change, I am going to add them as needed.  They are trivial to generate after all the previous changes.

PR #7970, opened 2022-01-15
feat(GCS+gRPC): add generated logging decorator

coryan:
Introduce the `StorageLogging` decorator for `StorageStub`. I used the
generator to create this code, with most methods removed afterwards. I
only included minimal tests for this class: the code is generated and we
trust the generator.

Part of the work for #7963

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7970)


google/cloud/storage/internal/storage_stub_factory_test.cc
```diff
@@ -27,10 +28,15 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
...
 
+using ::google::cloud::testing_util::ScopedLog;
 using ::google::cloud::testing_util::StatusIs;
+using ::testing::Contains;
+using ::testing::HasSubstr;
 
 // All the tests have nearly identical structure. They create a stub pointing to
-// an invalid endpoint. Getting a kUnavailable error is the expected result.
+// an invalid endpoint. The stubs have full decorators. We expect that they
+// will result in a call to a service.  Getting a kUnavailable error is the
```
dbolduc, line 38: Maybe combine these two sentences. "We expect that they will make a call to the service which results in a kUnavailable error."

Or just get rid of the "We expect that they will result in a call to a service" sentence altogether.

dbolduc (approved):

google/cloud/storage/internal/storage_stub_factory_test.cc
coryan: I changed the comments a bit.  I am going to make a couple of big changes to these tests, so getting the language "just so" is not that important.

coryan: Meaning: thanks for the suggestions, they are good, but I am going to keep changing the comments anyway.

PR #7985, opened 2022-01-16
ci: explicitly list targets instrumented for code coverage

coryan:
This simplifies the addition of new generated libraries.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7985)


.codecov.yml
```diff
@@ -12,37 +12,5 @@ coverage:
 #     generator/generator_config.textproto | sort -u
```
dbolduc, line 12: should we modify this comment block now?

```
# Ignore generated code that lives with handcrafted code.
```

dbolduc (approved):

.codecov.yml
coryan: Done.

PR #7992, opened 2022-01-17
feat(GCS+gRPC): add metadata decorator

coryan:
Use the generator to create a metadata decorator. Manually remove the
unused methods (for now).  The testing is fairly minimal because we
trust the generator.  It is not zero because I do not trust my "manually
remove" step as much as I trust the generator.

Part of the work for #7963

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/7992)


google/cloud/storage/internal/storage_metadata_decorator.cc
```diff
@@ -0,0 +1,75 @@
+// Copyright 2021 Google LLC
```
dbolduc, line 1: nit: s/2021/2022/

google/cloud/storage/internal/storage_metadata_decorator.h
```diff
@@ -0,0 +1,66 @@
+// Copyright 2021 Google LLC
```
dbolduc, line 1: nit: s/2021/2022/

google/cloud/storage/internal/storage_metadata_decorator.h
```diff
@@ -0,0 +1,66 @@
...
+  StatusOr<google::storage::v2::QueryWriteStatusResponse> QueryWriteStatus(
+      grpc::ClientContext& context,
+      google::storage::v2::QueryWriteStatusRequest const& request) override;
+
+ private:
+  void SetMetadata(grpc::ClientContext& context,
+                   std::string const& request_params);
+  std::shared_ptr<StorageStub> child_;
+  std::string api_client_header_;
+};  // StorageMetadata
```
dbolduc, line 58: (Not about this PR)

I am just now noticing that the generator adds this comment in the logging and metadata headers.... I don't think I like it....

https://github.com/googleapis/google-cloud-cpp/blob/354d2945c08b8308f57ea056502517d294e89d71/generator/internal/logging_decorator_generator.cc#L166
https://github.com/googleapis/google-cloud-cpp/blob/354d2945c08b8308f57ea056502517d294e89d71/generator/internal/metadata_decorator_generator.cc#L161

dbolduc (approved):

google/cloud/storage/internal/storage_metadata_decorator.cc
coryan: Fixed.

google/cloud/storage/internal/storage_metadata_decorator.h
coryan: Fixed

google/cloud/storage/internal/storage_metadata_decorator.h
coryan: Ack.  I am going to leave this as-is because I want the code to move towards what the generator creates.

PR #8001, opened 2022-01-18
refactor(GCS+gRPC): move stubs to `storage_internal`

coryan:
Move the `StorageStub` class and its decorators to the
`google::cloud::storage_internal` namespace. This is the namespace used
by the generator. It seems better to move the types than trying to tweak
the generator to use a special namespace just in this case.

Part of the work for #7963

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8001)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.h
+++ b/google/cloud/storage/internal/grpc_client.h
@@ -25,6 +25,13 @@
 
 namespace google {
 namespace cloud {
+
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+class StorageStub;
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
+}  // namespace storage_internal
+
 namespace storage {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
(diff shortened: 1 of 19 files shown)


google/cloud/storage/internal/grpc_client.h
```diff
@@ -183,13 +188,14 @@ class GrpcClient : public RawClient,
...
-  explicit GrpcClient(std::shared_ptr<StorageStub> stub, Options opts);
+  explicit GrpcClient(std::shared_ptr<storage_internal::StorageStub> stub,
+                      Options opts);
 
  private:
   Options options_;
   ClientOptions backwards_compatibility_options_;
   std::unique_ptr<google::cloud::BackgroundThreads> background_;
-  std::shared_ptr<StorageStub> stub_;
+  std::shared_ptr<google::cloud::storage_internal::StorageStub> stub_;
```
dbolduc, line 198: nit: will just `storage_internal::StorageStub` do?

dbolduc (changes requested):
This is the first time we are introducing the namespace. I think you need to exclude it from the doxygen symbols:

https://github.com/googleapis/google-cloud-cpp/blob/c9b4b5dd74f3cf2e1694826f579305ac32427a17/google/cloud/storage/CMakeLists.txt#L25

google/cloud/storage/internal/grpc_client.h
coryan: Done.

coryan:
> This is the first time we are introducing the namespace. I think you need to exclude it from the doxygen symbols:

Done.

coryan:
PTAL

dbolduc (approved):
oops, I missed the latest commit. thanks for the nudge.

PR #8007, opened 2022-01-18
impl: add httpbody.proto to CMake build

devjgm:
Part of: #7892

This is needed by the upcoming library for Google Cloud Build.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8007)

### DIFF
--- a/external/googleapis/CMakeLists.txt
+++ b/external/googleapis/CMakeLists.txt
@@ -526,7 +526,6 @@ string(
            " google_cloud_cpp_api_client_protos"
            " google_cloud_cpp_api_field_behavior_protos"
            " google_cloud_cpp_api_http_protos"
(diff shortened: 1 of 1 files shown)


external/googleapis/CMakeLists.txt
```diff
@@ -523,6 +526,7 @@ string(
            " google_cloud_cpp_api_client_protos"
            " google_cloud_cpp_api_field_behavior_protos"
            " google_cloud_cpp_api_http_protos"
+           " google_cloud_cpp_api_httpbody_protos"
```
coryan, line 529: This list is used for backwards compatibility (see #5726).  We probably do not want to grow it.

devjgm: Removed.

coryan (approved):

PR #8013, opened 2022-01-19
feat(generator): connection respects per call policies

### LINKED ISSUE #8002: Generator accepts per-call retry/backoff/polling policies
Our retry/backoff/polling policies are inextricably linked to the `Connection` layer. We set them when the `Connection` layer is created: https://github.com/googleapis/google-cloud-cpp/blob/5ba920c3c76db1b414adaaa991e05d37b9c6169e/google/cloud/accessapproval/access_approval_connection.cc#L102-L108

We do not override them with `internal::CurrentOptions()` set from the `Client` either: https://github.com/googleapis/google-cloud-cpp/blob/5ba920c3c76db1b414adaaa991e05d37b9c6169e/google/cloud/accessapproval/access_approval_connection.cc#L153

Seems like we can just add getters like this to `ServiceConnectionImpl`:
```cc
std::unique_ptr<RetryPolicy> retry_policy() {
  auto options = internal::CurrentOptions();
  if (options.has<ServiceRetryPolicyOption>()) {
    return options.get<ServiceRetryPolicyOption>()->clone();
  }
  return retry_policy_prototype_->clone();
}
```
then we just s/`retry_policy_prototype_->clone()`/`retry_policy()`/ everywhere else

dbolduc:
Fixes #8002 

First commit has the logic changes in `generator/`
Second commit has the changes to the golden files
Third commit has the changes to generated libraries in `google/cloud/...`

I decided to use `if` blocks instead of predicates, per [this discussion](https://github.com/googleapis/google-cloud-cpp/pull/7885#discussion_r784161046).

We have the following types of calls:
1. Sync Loops - (**R**, **B**, I)
2. LRO Polling Loops - (R, B, I, **P**)
3. Paginated (StreamRange) - (R, B, I)
4. Streaming Reads (StreamRange) - (R, B)
5. Async Loop - (R, B, I)

Letters in parentheses are the policies used in such a call.
R = Retry, B = Backoff, I = Idempotency, P = Polling.

The **bold** letters are the ones I actually tested for. I did not feel like writing any other tests. I would begrudgingly add more tests, if asked nicely.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8013)

### DIFF
--- a/generator/integration_tests/golden/tests/plumbing_test.cc
+++ b/generator/integration_tests/golden/tests/plumbing_test.cc
@@ -25,6 +25,7 @@ namespace golden {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace {
 
+using ::testing::AtLeast;
 using ::testing::Return;
 using ms = std::chrono::milliseconds;
 
@@ -53,7 +54,9 @@ TEST(PlumbingTest, RetryLoopUsesPerCallPolicies) {
     auto clone = absl::make_unique<MockRetryPolicy>();
     // We will just say the policy is never exhausted, and use a permanent error
     // to break out of the loop.
-    EXPECT_CALL(*clone, IsExhausted).WillRepeatedly(Return(false));
+    EXPECT_CALL(*clone, IsExhausted)
+        .Times(AtLeast(1))
+        .WillRepeatedly(Return(false));
     EXPECT_CALL(*clone, OnFailureImpl);
     return clone;
   });
--- a/generator/integration_tests/golden/tests/plumbing_test.cc
+++ b/generator/integration_tests/golden/tests/plumbing_test.cc
@@ -1,0 +1,123 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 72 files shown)


generator/integration_tests/golden/tests/plumbing_test.cc
```diff
@@ -0,0 +1,123 @@
...
+  auto call_b = std::make_shared<MockBackoffPolicy>();
+  auto call_options = Options{}
+                          .set<GoldenThingAdminRetryPolicyOption>(call_r)
+                          .set<GoldenThingAdminBackoffPolicyOption>(call_b);
+
+  EXPECT_CALL(*call_r, clone).WillOnce([] {
+    auto clone = absl::make_unique<MockRetryPolicy>();
+    // We will just say the policy is never exhausted, and use a permanent error
+    // to break out of the loop.
+    EXPECT_CALL(*clone, IsExhausted).WillRepeatedly(Return(false));
```
coryan, line 56: Consider `.Times(AtLeast(1)).WillRepeatedly(Return(false));` ?

coryan (approved):
There are a couple of PRs in flight that you may need to coordinate with.

coryan (approved):

generator/integration_tests/golden/tests/plumbing_test.cc
dbolduc: sure

PR #8033, opened 2022-01-19
fix(storage): improve JSON validation

### LINKED ISSUE #8027: [Flake] error in JSON library in storage signed_url_integration_test.cc
I ran `asan-pr` locally, and ran into the following flake in this test: https://github.com/googleapis/google-cloud-cpp/blob/9c793dae470fd82a9ea86572fb0c8ab4a12ea736/google/cloud/storage/tests/signed_url_integration_test.cc#L132

```
[ RUN      ] SignedUrlIntegrationTest.CreateV4SignedUrlGet
  what():  [json.exception.type_error.306] cannot use value() with string
```

The error message reminds me of this PR: #7702

coryan:
Prefer using `!is_object()` over `is_discarded()` to detect parsing
errors. In many cases this change made no difference (e.g. where there
was an implicit check via `.count()`). In a few places the new tests
triggered latent bugs. One of these bugs was real, but it is hard to say
which one.

Fixes #8027

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8033)

### DIFF
--- a/google/cloud/storage/oauth2/google_credentials_test.cc
+++ b/google/cloud/storage/oauth2/google_credentials_test.cc
@@ -445,17 +445,23 @@ TEST_F(GoogleCredentialsTest, LoadUnknownTypeCredentials) {
 }
 
 TEST_F(GoogleCredentialsTest, LoadInvalidCredentials) {
-  std::string filename = google::cloud::internal::PathAppend(
-      ::testing::TempDir(), "invalid-credentials.json");
-  std::ofstream os(filename);
-  std::string contents_str = R"""( not-a-json-object-string )""";
-  os << contents_str;
-  os.close();
-  ScopedEnvironment adc_env_var(GoogleAdcEnvVar(), filename.c_str());
+  char const* const test_cases[] = {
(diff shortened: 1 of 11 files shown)


google/cloud/storage/oauth2/google_credentials_test.cc
```diff
@@ -445,17 +445,23 @@ TEST_F(GoogleCredentialsTest, LoadUnknownTypeCredentials) {
...
-
-  auto creds = GoogleDefaultCredentials();
-  EXPECT_THAT(creds, StatusIs(StatusCode::kInvalidArgument,
-                              HasSubstr("credentials file " + filename)));
+  char const* const test_cases[] = {
+      R"""( not-a-json-object-string )""",
+      R"js("valid-json-but-not-an-object")js",
+  };
+  for (auto const* contents : test_cases) {
+    SCOPED_TRACE("Testing with: " + std::string{contents});
```
dbolduc, line 453: TIL: `SCOPED_TRACE()`

dbolduc (approved):
Thanks for going through and fixing these.

PR #8041, opened 2022-01-20
cleanup: reduce manual steps to update libraries

coryan:
Make it easier to update the list of protos in `google/api` and
`google/rpc`. Also use `update_libraries.sh` to update the list of
dependencies for each proto.  While most have no dependencies, keeping
the list for `google/api/service.proto` seems daunting.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8041)


external/googleapis/update_libraries.sh
```diff
@@ -137,6 +137,23 @@ declare -A -r LIBRARIES=(
...
+  ["api_billing"]="@com_google_googleapis//google/api:billing_proto"
+  ["api_distribution"]="@com_google_googleapis//google/api:distribution_proto"
+  ["api_endpoint"]="@com_google_googleapis//google/api:endpoint_proto"
+  ["api_log"]="@com_google_googleapis//google/api:log_proto"
+  ["api_logging"]="@com_google_googleapis//google/api:logging_proto"
+  ["api_monitored_resource"]="@com_google_googleapis//google/api:monitored_resource_proto"
+  ["api_monitoring"]="@com_google_googleapis//google/api:monitoring_proto"
+  ["api_quota"]="@com_google_googleapis//google/api:quota_proto"
+  ["api_usage"]="@com_google_googleapis//google/api:usage_proto"
+  ["api_service"]="@com_google_googleapis//google/api:service_proto"
```
dbolduc, line 156: In all of these, do we prefer things like `service_cc_proto`?

coryan: Done. I am not sure if would make any difference?  We certainly cannot use `*_cc_grpc` because they do not exist.

dbolduc (approved):
I cannot read CMake, but I get the gist of what's happening

coryan:
I just realized that the `*.lists` files do not look right.

coryan:
Closing for now.  I may get back to it later.

PR #8053, opened 2022-01-21
feat(generator): handle proto fields named with C++ keywords

devbww:
Motivated by a new service with a proto containing a "namespace"
field.  The C++ proto compiler adds a trailing underscore to such
fields and their associated accessors, so we should do the same.
And we should also do it in the parameters to "method_signature"
overloads, and in generated @param annotations.

If anyone has any better ideas for working around the cmake issue,
or for some alternative method to access libprotoc's cpp keyword
mapping, I'm all ears.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8053)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -38,28 +38,20 @@
 #include <regex>
 #include <string>
 
-// #include <google/protobuf/compiler/cpp/cpp_helpers.h> doesn't work with the
-// existing cmake support because it includes other files that are not part of
-// ${libprotoc_headers}. So, we just directly declare what we want to use.
-namespace google {
-namespace protobuf {
-namespace compiler {
-namespace cpp {
-std::string ResolveKeyword(std::string const&);
-}  // namespace cpp
-}  // namespace compiler
-}  // namespace protobuf
-}  // namespace google
-
 using ::google::protobuf::FieldDescriptor;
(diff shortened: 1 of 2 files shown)


generator/internal/descriptor_utils.cc
```diff
@@ -38,9 +38,23 @@
 #include <regex>
 #include <string>
 
+// #include <google/protobuf/compiler/cpp/cpp_helpers.h> doesn't work with the
```
coryan, line 41: Seems to me that protobuf does not install that header at all, and therefore it is not intended for public  consumption.  I assume either they have a bug (they intend to install this, but just don't for some reason), or we are using an implementation detail.  May be worth checking with the protobuf folks about how they intend implementors of C++ plugins to use this?

devbww: Changed to use our own keyword-mapping function, which can be enhanced as the need arises.  For now, just "namespace".

coryan (approved):

generator/internal/descriptor_utils.cc
devbww: It turns out that the function I really should have been using from `cpp_helpers.h` is also declared in `cpp_names.h`, which is public.  So, I'm going to close this and send a new PR using that instead.

PR #8057, opened 2022-01-21
test(bigtable): split instance admin test in two

dbolduc:
Split the test in two. I will keep the new one, and eventually delete the old one. I tried to have the new one be named "instance_admin_test.cc" and the old one be named "instance_admin_legacy_test.cc", but I could not get the diff that I wanted.

If we think this is being too careful, we can go with #8055 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8057)

### DIFF
--- a/google/cloud/bigtable/instance_admin_new_test.cc
+++ b/google/cloud/bigtable/instance_admin_new_test.cc
@@ -1,0 +1,98 @@
+// Copyright 2022 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/instance_admin_new_test.cc
```diff
@@ -0,0 +1,98 @@
+// Copyright 2022 Google Inc.
```
coryan, line 1: s/Inc./LLC/

coryan (approved):
nit: "new test" is planned obsolescence ... the next test will be the "newer test"?  Okay if it is temporary.

google/cloud/bigtable/instance_admin_new_test.cc
dbolduc: It is temporary. I am not doing things a great way, but in the end it will not matter.

I am going to merge and make the s/Inc./LLC/ in a subsequent PR so I don't have to wait for builds.

Issue #7534, opened 2021-10-29
Consider implementing `InstanceAdmin`, `InstanceAdminClient` in terms of generated `BigtableInstanceAdminConnection`


dbolduc:
Basically the same as #7530, but without needing to deal with an async stub call

PR #8061, opened 2022-01-22
impl(bigtable): implement connection mechanism in legacy InstanceAdmin

dbolduc:
Part of the work for #7534.

This PR adds the mechanism for storing generated connections in `InstanceAdmin`. This code is unused and sits side by side with the old way (using `InstanceAdminClient`). This PR is one step in the conversion process. The big picture can be found [here](https://github.com/googleapis/google-cloud-cpp/compare/main...dbolduc:bigtable-convert-instance-admin).

Previous code could have one `InstanceAdminClient` that is shared by N `InstanceAdmin`s. In this case, we want the N `InstanceAdmin`s to all share the same `bigtable_admin::BigtableInstanceAdminConnection`. So we make the connection upon construction of the `InstanceAdminClient` and give the `InstanceAdmin` access to it.

Temporary things:
* The added constructor will make an `InstanceAdmin` that cannot make RPC calls. That is okay, it will function by the release.
* Every `InstanceAdminClient` initiates an extra connection it does not use. This is fine, we will get rid of the legacy stuff before release.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8061)

### DIFF
--- a/google/cloud/bigtable/instance_admin_client.cc
+++ b/google/cloud/bigtable/instance_admin_client.cc
@@ -325,13 +325,12 @@ class DefaultInstanceAdminClient : public InstanceAdminClient {
         stub->AsyncGetOperation(context, request, cq).release());
   }
 
- protected:
+ private:
   std::shared_ptr<bigtable_admin::BigtableInstanceAdminConnection> connection()
       override {
     return connection_;
   }
 
- private:
   google::cloud::BackgroundThreadsFactory BackgroundThreadsFactory() override {
     return impl_.BackgroundThreadsFactory();
   }
--- a/google/cloud/bigtable/instance_admin_client.cc
+++ b/google/cloud/bigtable/instance_admin_client.cc
@@ -41,7 +41,9 @@ namespace btadmin = ::google::bigtable::admin::v2;
 class DefaultInstanceAdminClient : public InstanceAdminClient {
  public:
   DefaultInstanceAdminClient(std::string project, Options options)
-      : project_(std::move(project)), impl_(std::move(options)) {}
+      : project_(std::move(project)), impl_(options) {
+    MakeConnection(std::move(options));
+  }
 
   std::string const& project() const override { return project_; }
   std::shared_ptr<grpc::Channel> Channel() override { return impl_.Channel(); }
--- a/google/cloud/bigtable/instance_admin_client.cc
+++ b/google/cloud/bigtable/instance_admin_client.cc
@@ -41,9 +41,10 @@ namespace btadmin = ::google::bigtable::admin::v2;
 class DefaultInstanceAdminClient : public InstanceAdminClient {
  public:
   DefaultInstanceAdminClient(std::string project, Options options)
-      : project_(std::move(project)), impl_(options) {
-    MakeConnection(std::move(options));
-  }
+      : project_(std::move(project)),
+        impl_(options),
+        connection_(bigtable_admin::MakeBigtableInstanceAdminConnection(
+            std::move(options))) {}
 
   std::string const& project() const override { return project_; }
   std::shared_ptr<grpc::Channel> Channel() override { return impl_.Channel(); }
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/instance_admin_client.h
```diff
@@ -87,7 +88,11 @@ class InstanceAdminClient {
   // them protected, so the mock classes can override them, and then make the
   // classes that do use them friends.
  protected:
-  friend class InstanceAdmin;
+  void MakeConnection(Options options) {
```
coryan, line 91: This function seems to be used only once?  Should we just remove it and do the work in the header?  Alternatively, a static function that returns the value to put in `connection_`?

google/cloud/bigtable/instance_admin_new_test.cc
```diff
@@ -12,85 +12,94 @@
...
 
 namespace google {
 namespace cloud {
 namespace bigtable {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
-namespace {
 
-using ::testing::ReturnRef;
+// Helper class for checking that the legacy API still functions correctly
+class InstanceAdminTester {
```
coryan, line 25: No action needed, just for you to think about: this is a "test smell", you need this to verify that the `private` sections of a class behave a certain way, the question should be: what behavior of the class do I need to test?  Is that behavior observable?  If not, why do I want a test? (sometimes the answer is: the behavior is only observable in terms of performance and testing for that is difficult, so this may be Okay... ).

coryan (approved):

google/cloud/bigtable/instance_admin_new_test.cc
dbolduc: Hmm. (I think) I want to test that sharing one `InstanceAdminClient` across multiple `InstanceAdmin`s only shares one `BigtableInstanceAdminConnection` under the hood.
https://github.com/googleapis/google-cloud-cpp/blob/be590f9306c7fde14b906791eb751da256129e21/google/cloud/bigtable/instance_admin_new_test.cc#L93-L101

If we didn't do this, maybe it takes longer to make new `InstanceAdmin`s then it should. Or maybe we start using too many threads or something. So I think it is like you say: there is an observable difference, but it is not easy to test for it.

google/cloud/bigtable/instance_admin_client.h
dbolduc: I moved `connection_` into the implementation class. The base class has a virtual accessor (for `InstanceAdmin` to use).

My plan is to make this a pure virtual function. I do not feel bad about adding a pure virtual function, because we are going to break the class anyway. I did not make it a pure virtual function at this moment because I did not feel like making changes in derived classes (`mock_instance_admin_client.h`, `logging_instance_admin_client.h`).

I dismissed your approval. PTAL.

google/cloud/bigtable/instance_admin_client.cc
```diff
@@ -322,13 +325,20 @@ class DefaultInstanceAdminClient : public InstanceAdminClient {
         stub->AsyncGetOperation(context, request, cq).release());
   }
 
+ protected:
+  std::shared_ptr<bigtable_admin::BigtableInstanceAdminConnection> connection()
```
coryan, line 329: This changes the access modifier for `connection()`, that is Okay, but is it intentional?  Does the derived class need to use it?  If not, (seems that way) keep it `private`?

google/cloud/bigtable/instance_admin_client.h
```diff
@@ -325,6 +325,14 @@ class InstanceAdminClient {
                     google::longrunning::GetOperationRequest const& request,
                     grpc::CompletionQueue* cq) = 0;
   //@}
+
+ private:
+  friend class InstanceAdmin;
```
coryan, line 330: Do you need the `friend`?

coryan (changes requested):

google/cloud/bigtable/instance_admin_client.cc
dbolduc: Argh. This was an oversight. Thanks for catching it.

google/cloud/bigtable/instance_admin_client.h
dbolduc: yes. https://github.com/googleapis/google-cloud-cpp/blob/a62ea5760134a6ec48959777a21943bf95f2b590/google/cloud/bigtable/instance_admin.h#L148

coryan (approved):

PR #8062, opened 2022-01-22
test(bigtable): relax logging tests

dbolduc:
Instead of checking that nothing at all is logged, only check that the logging decorator is not in use. (I have tests breaking due to extraneous logs from `CheckExpectedOptions<>`)

All of this has been prophesied by @devbww

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8062)


devbww (commented):
As you know, I'm with you, so I'll let others have their say first.

coryan (approved):

PR #8067, opened 2022-01-23
cleanup(generator): simplify golden files upkeep

coryan:
Change the generator so we can use as single command to update the
golden files.  With this in place, comparing the golden files becomes
very similar to the process of comparing the generated files, so I
change the builds to do that instead.  That produces nice diffs when the
golden files change, as opposed to the unreadable output from
googletest.  I then removed the integration test program for the
generator, since it was no longer needed.

I also documented how to manually run the generator to produce the
golden files. Before, the only known way to do it involved running one
of the builds in "--docker-shell" mode, waiting for a full build, for
the unit tests to pass (note this required the golden files to compile),
then for the integration test to fail, then copy the golden files from
`/tmp`.  During development the golden files would (often) not compile,
so one needed to revert the golden files and start over.


generator/integration_tests/README.md
```diff
@@ -0,0 +1,19 @@
+# Updating the golden files
+
+We keep a set of "golden" files to ensure the generator has not changed its
+output in unexpected ways.  If you need to update these files, use this command:
```
devbww, line 4: Question to ponder:  Now that we have actual generated interfaces to compare against, do the "golden" files have any value-add?

coryan: I do not think they add much.  Let's talk about this on the next team meeting.

coryan: FWIW, I would rather merge this as-is and discuss the removal later.  I am making more than one change to the generator and the current workflows are annoying.

scotthart (approved):

PR #8075, opened 2022-01-24
fix(pubsublite): Restructure stream_factory.h to use generated stubs

dpcollins-google:
These are still useful to provide the input factory in a structured way for the (eventual) ResumableStreamingReadWriteRpc matching the patterns of https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/internal/resumable_streaming_read_rpc.h

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8075)

### DIFF
--- a/google/cloud/pubsublite/internal/stream_factory.h
+++ b/google/cloud/pubsublite/internal/stream_factory.h
@@ -32,12 +32,8 @@ namespace pubsublite_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 template <class Request, class Response>
-using BidiStream =
-    ::google::cloud::AsyncStreamingReadWriteRpc<Request, Response>;
-
-template <class Request, class Response>
-using StreamFactory =
-    std::function<std::unique_ptr<BidiStream<Request, Response>>()>;
+using StreamFactory = std::function<
+    std::unique_ptr<AsyncStreamingReadWriteRpc<Request, Response>>()>;
 
 using ClientMetadata = std::unordered_map<std::string, std::string>;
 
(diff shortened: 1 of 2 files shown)


google/cloud/pubsublite/internal/stream_factory.h
```diff
@@ -48,71 +52,41 @@ inline std::unique_ptr<grpc::ClientContext> MakeGrpcClientContext(
...
+                  google::cloud::CompletionQueue const& cq,
+                  ClientMetadata const& metadata) {
+  return
+      [=] { return stub->AsyncPublish(cq, MakeGrpcClientContext(metadata)); };
 }
 
 inline StreamFactory<pubsublite::v1::SubscribeRequest,
                      pubsublite::v1::SubscribeResponse>
 MakeStreamFactory(
     // NOLINTNEXTLINE(performance-unnecessary-value-param)
```
coryan, line 65: I think the NOLINT is no longer needed?

coryan (changes requested):
A couple of nits, one noted below.  It also seems that `BidiStream<>` is used only once, maybe you have future plans for it, but if not, consider removing it.

google/cloud/pubsublite/internal/stream_factory.h
dpcollins-google: Done.

dpcollins-google:
> A couple of nits, one noted below. It also seems that `BidiStream<>` is used only once, maybe you have future plans for it, but if not, consider removing it.

removed

coryan (approved):
Thanks!

PR #8087, opened 2022-01-24
impl(generator): skip method signatures that contain deprecated fields

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8087)


coryan (approved):
Thanks!

generator/internal/descriptor_utils.cc
```diff
@@ -229,10 +235,12 @@ void SetMethodSignatureMethodVars(
       }
       method_signature += ", ";
     }
-    std::string key = "method_signature" + std::to_string(i);
+    if (field_deprecated) continue;
```
devbww, line 238: Doesn't this mean that when an existing field is newly deprecated, we would generate a breaking change?

PR #8088, opened 2022-01-25
refactor(GCS+gRPC): reorder GrpcObjectRequestParser

coryan:
Sort the functions in this class so they match the order of declaration
in the `StorageStub`. A bit silly, but makes it easier to grok the PR
changes, and possibly allows me to send more PRs in parallel.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8088)


dbolduc (approved):
"I would like to believe some of these PRs actually add value, and
not just churn code around." 

j/k 😉

PR #8099, opened 2022-01-25
ci: add bazel-latest build

devjgm:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8099)

### DIFF
--- a/ci/cloudbuild/builds/bazel-latest.sh
+++ b/ci/cloudbuild/builds/bazel-latest.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+#
+# Copyright 2022 Google LLC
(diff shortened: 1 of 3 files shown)


coryan (approved):
CI-only changes are usually `ci: add bazel-latest build`, we use `impl:` for implementation details without observable behavior for customers (e.g., adding something in `internal::` that is not used yet).

devjgm:
> CI-only changes are usually `ci: add bazel-latest build`, we use `impl:` for implementation details without observable behavior for customers (e.g., adding something in `internal::` that is not used yet).

Right. I forgot. Fixed.

Issue #8115, opened 2022-01-26
Cloud Channel API: `cloudchannel.googleapis.com`


devjgm:

devjgm:
UPDATE:
PR adding this: https://github.com/googleapis/google-cloud-cpp/pull/8116

Breaks on macos due to:

```
bazel-out/darwin-fastbuild/bin/external/com_google_googleapis/google/cloud/channel/v1/common.pb.h:545:33: error: expected member name or ';' after declaration specifiers
  static constexpr CustomerType DOMAIN =
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
/Applications/Xcode_12.0.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/math.h:732:17: note: expanded from macro 'DOMAIN'
#define DOMAIN          1
                        ^
```
https://source.cloud.google.com/results/invocations/8ee5da3b-6410-4185-a0d2-b38f5282069c/targets/cloud-cpp%2Fgithub%2Fgoogle-cloud-cpp%2Fmain%2Fmacos%2Fbazel-presubmit/log

That is, protobuf-generated code uses a symbol named `DOMAIN` that conflicts w/ a macro definition.

PR #8116, opened 2022-01-26
feat(channel): generate library

devjgm:
Fixes: #8115

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8116)


.typos.toml
```diff
@@ -4,6 +4,7 @@ extend-exclude = [
     "google/cloud/bigtable/internal/readrowsparser_acceptance_tests.inc",
     # The source proto files have one or more typos in their comments
     "google/cloud/accessapproval/access_approval_client.h",
+    "google/cloud/channel/cloud_channel_client.h",
```
coryan, line 7: Consider sending CLs to fix the typos upstream, I just merged my fixes for storage_transfer_client.h

BUILD.bazel
```diff
@@ -171,6 +171,13 @@ cc_library(
     ],
 )
 
+cc_library(
```
coryan, line 174: Seems like you are missing the `experimental-channel_mocks` rule.

coryan (changes requested):

.typos.toml
devjgm: It's not clear that it's a real typo. It complained about "add-ons" (the ons part). Maybe it's a `typos` bug. Not sure. In a way, it may be good to teach typos to ignore all auto-generated libraries.

BUILD.bazel
devjgm: Thanks. Added.

.typos.toml
coryan: The comments go straight to our Doxygen documentation.  The number of typos is small, but I would rather not ignore them and fix the problems upstream.  Less embarrassing for everyone involved.  It would be nice if the protos in googleapis/googleapis used some kind of typo detector too.

coryan (approved):

.typos.toml
devjgm: Makes sense. I'll keep that in mind. In this case, `typos` complained about "Add-ons", which I think is spelled correctly; it's likely a bug in `typos`.

devbww: Can't we add an `extend-words` entry to our `.typos.toml` file?

devjgm: Generally, we could do that. In this case, however, the typo is actually "ons" from the word "add-ons". So if we used `extend-words` we'd have to teach it that "ons" is a valid word, which is generally not true. Extending the allowed words to allow "add-ons" does not work.

FTR, the actual output was:
```
  --> ./google/cloud/channel/cloud_channel_client.h:813:10
    |
813 |   /// add-ons, or entitlements for Google Cloud's development platform.
    |           ^^^
    |
... 0.784 seconds
```

coryan (approved):

PR #8122, opened 2022-01-26
impl(storage): make patch objects accessible

coryan:
So far we have been using strings to represent "patch" requests. For
JSON this works, just store the string representation of a JSON patch.
For gRPC this will not work anymore, we need the actual patch to compute
the corresponding gRPC request.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8122)

### DIFF
--- a/google/cloud/storage/internal/patch_builder_details.h
+++ b/google/cloud/storage/internal/patch_builder_details.h
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC
+// Copyright 2022 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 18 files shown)


google/cloud/storage/internal/patch_builder_details.h
```diff
@@ -0,0 +1,48 @@
+// Copyright 2018 Google LLC
```
devjgm, line 1: This copyright year may need to be updated since this is a brand new file.

devjgm (approved):

google/cloud/storage/internal/patch_builder_details.h
coryan: Ooops, fixed.

PR #8133, opened 2022-01-27
feat(debugger): generate library

devjgm:
Fixes: #8118

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8133)


google/cloud/debugger/README.md
```diff
@@ -2,8 +2,9 @@
 
 :construction:
 
-This directory contains an idiomatic C++ client library for the
-[Stackdriver Debugger API][cloud-service-docs], a service to Examines the call stack and variables of a running application without stopping or slowing it down.
+This directory contains an idiomatic C++ client library for the [Stackdriver
```
coryan, line 5: nit: this makes it sound like you can debug C++ applications in GCP, which is not what the service is for.  I think we should merge this as-is and think about how we describe this library, the Cloud Logging, and the (eventual) Cloud Monitoring libraries.  Something along the lines of "these are the building blocks to implement C++ logging to cloud, but you should not use this directly".

coryan (approved):

google/cloud/debugger/README.md
devjgm: ack. good point.

PR #8135, opened 2022-01-27
refactor: move run_emulator_utils.sh to common for use with rest

scotthart:
The GCS emulator runs a local copy of httpbin. The Rest library integration tests will use it as well.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8135)


coryan (approved):
Thanks for refactoring this in a separate PR.

PR #8157, opened 2022-01-29
fix(rest): initial partial read could overflow spill buffer

scotthart:
The intent of handling a response by first writing all bytes to the spill buffer was to enable the status code and headers to be interrogated via the `RestResponse` class, without having to read all the bytes in the response body. The initial implementation was based on the model that libcurl would write all of CURLOPT_BUFFERSIZE bytes at once into the spill buffer. This was incorrect and in order to handle multiple writes to the spill buffer new monitoring of when the status line and headers were received was added.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8157)

### DIFF
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -181,6 +181,8 @@ CurlImpl::CurlImpl(CurlHandle handle,
       multi_(factory_->CreateMultiHandle()),
       in_multi_(false),
       paused_(false),
+      status_line_received_(false),
+      all_headers_received_(false),
       buffer_({nullptr, 0}),
       spill_offset_(0),
       options_(std::move(options)) {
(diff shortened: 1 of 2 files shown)


google/cloud/internal/curl_impl.cc
```diff
@@ -334,8 +337,20 @@ std::size_t CurlImpl::WriteCallback(void* ptr, std::size_t size,
   return WriteToUserBuffer(ptr, size, nmemb);
 }
 
+// libcurl invokes the HEADERFUNCTION exactly once for each header line received
+// and only a complete header line is passed to the function. Additionally, the
+// status line and blank lines preceding and following the headers are passed
+// to this function.
 std::size_t CurlImpl::HeaderCallback(char* contents, std::size_t size,
                                      std::size_t nitems) {
+  if (nitems == 2 && contents[0] == '\r' && contents[1] == '\n') {
```
coryan, line 346: nit: I think you want `nitems * size == 2`?

google/cloud/internal/curl_impl.cc
```diff
@@ -334,8 +337,20 @@ std::size_t CurlImpl::WriteCallback(void* ptr, std::size_t size,
...
+    if (!status_line_received_) {
+      status_line_received_ = true;
+    } else {
+      all_headers_received_ = true;
+      http_code_ = handle_.GetResponseCode();
```
coryan, line 351: Thanks, this solves a mystery for me.

google/cloud/internal/curl_impl.cc
```diff
@@ -576,9 +591,11 @@ StatusOr<std::size_t> CurlImpl::ReadImpl(absl::Span<char> output) {
 
   Status status;
   if (buffer_.empty()) {
-    status = PerformWorkUntil([this] {
-      return curl_closed_ || paused_ || spill_offset_ >= spill_.max_size();
-    });
+    // Once we have received the status and all the headers, we have read
+    // enough to satisfy calls to any of RestResponse's methods and can
```
coryan, line 595: s/methods and/methods, and/ ?

google/cloud/internal/curl_impl.cc
```diff
@@ -241,21 +243,22 @@ CurlImpl::~CurlImpl() {
...
   }
-  // As this function is only called on the first Read, the spill buffer should
-  // be empty so we can always copy the full contents of `ptr` into the spill
-  // buffer.
-  std::memcpy(spill_.data(), ptr, size * nmemb);
-  spill_offset_ = size * nmemb;
+  // As this function can be called multiple times on the first Read, the spill
+  // buffer may be written to several times before it is full.
+  std::memcpy(spill_.data() + spill_offset_, ptr, size * nmemb);
+  spill_offset_ += size * nmemb;
```
coryan, line 261: Just an observation: this is starting to look similar to how we write into the application's buffer.

coryan (approved):

PR #8158, opened 2022-01-29
feat(generator): merge connection options into client options

### LINKED ISSUE #8054: Merge connection options into client options, before defaulting the client options.
In generated libraries, policy options passed to `MakeConnection(Options options)` are overwritten by the defaults in the `Client`.

See: (internal) ["design" doc](https://docs.google.com/document/d/1y9BfD2STOLAFnjgB6vH7V70ks1E4tyOGSak2yvzqI2I/) describing the problem, and the proposed solution.

devbww:
Store the Connection options so that they may be made available to
merge into the Client options, and then use them to implement the
four polling-policy data elements.  (This is the generator version
of #8090.)

Rules of thumb for `Options` values:
- `MergeOptions()` LHS should not include the service defaults
- `MergeOptions()` RHS should include service defaults
- An `OptionsSpan` should include the service defaults
- stored `Options` (data members) should include the service defaults

Code can then normally just look at the `CurrentOptions()`, knowing
that they will include the service defaults.  For example, the policy
options will be available.  A special exception to this, however, is
when testing the connection layer in isolation (without a client), as
no one has created an `OptionsSpan` in that case.  We could add an
`OptionsSpan` to the connection-layer calls, but it would be a no-op
in practice.  So, tests need to ensure required options are available.

Fixes #8054.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8158)


devbww:
Don't bother looking at the "regenerated" commits.

coryan (approved):
Thank you for separating the "generated" commits, made it much easier to review things.

Do you think it would be possible to add some sort of unit test to the generated golden files?  We don't write tests for each generated library, and we explain this to ourselves by testing the golden files once.  I am Okay we decide to add them in a separate PR.

dbolduc (commented):
Thanks for doing the fix

devbww:
> Do you think it would be possible to add some sort of unit test to the generated golden files?

Yes, I look into it.  Thanks.

PR #8170, opened 2022-01-30
feat(generator): always hand-craft bidir streaming body

coryan:
This changes the generator to **not** emit the body of the bidirectional streaming RPC.  There are very few services that need these RPCs, many are hand-crafted in full, many need a very unique implementation for this function, some do not need a retry / resume loop at all.

We will be implementing these functions by hand, in a `internal/streaming.cc` file (or similar).  In this PR only the golden files and `google/cloud/logging` are affected.  I am not implementing interesting versions of the function body, that will come in separate PRs.

Part of the work for #7795.  I added some TODO entries to prevent accidentally declaring GA for the logging library.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8170)

### DIFF
--- a/generator/integration_tests/golden/streaming.cc
+++ b/generator/integration_tests/golden/streaming.cc
@@ -41,7 +41,6 @@ void GoldenKitchenSinkTailLogEntriesStreamingUpdater(
   // implementation here
 }
 
-
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace golden
 }  // namespace cloud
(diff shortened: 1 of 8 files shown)


generator/integration_tests/golden/streaming.cc
```diff
@@ -25,6 +41,7 @@ void GoldenKitchenSinkTailLogEntriesStreamingUpdater(
   // implementation here
 }
 
+
```
devbww, line 44: Nit: Extra blank line.

coryan: Fixed.

devbww (approved):

Issue #5934, opened 2021-02-26
Proposal: modernize the Bigtable internals


coryan:
## Status

In Review | Approved | **In Progress** | Abandoned

This proposal is under review, we have not committed to implementing it. It
needs to wait until the `Async*` are retired in any case.

## Background

Bigtable is the first library we implemented, and its internals show it. We should consider a cleanup to incorporate the
lessons from the last 3 years into its design. We should do this without breaking customers, or at least minimizing the
breakage.

Bigtable is a "resource centric" library, there are classes to represent a *Table* or an *Instance*, whereas newer
libraries are "service centric". If we were writing Bigtable again there would be a `bigtable::Client` class and would
receive the name of the table in each function (as opposed to caching it in the `bigtable::Table` class).

Bigtable is hard to mock. The only place where application developers can mock behavior is the (confusingly
named) `bigtable::DataClient` class. This forces them to think of low-level gRPC concepts,
like `grpc::ClientReaderInterface<T>`, and they need to implement low-level chunking protocols to meet the expectations
of `bigtable::Table`.

The `bigtable::*Client` classes are pure virtual, and the majority of their member functions are `protected`. While
using these classes to mock behavior ranges from "awkward" to "unnecessarily difficult", some people do manage to
implement mocks through them.

Bigtable uses streaming (read) RPCs, including `MutateRows()` and `ReadRows()`. `ReadRows()` is complicated because a
single row (or even cell!) can be chunked across several responses. Moreover, the protocol to resume is to restart the
read from the last fully received *row*, there is no simple "token" to resume from the last received chunk.

We should be able to use the micro-generator for the Bigtable administrative APIs. There are a number of helper types (
builders for update and create requests), that we may want to preserve as add-ons.

## Design Overview

The design is guided by two principles:

* We should preserve the user-facing classes, `bigtable::Table`,
  `bigtable::TableAdmin`, and `bigtable::InstanceAdmin`. With the existing API, semantics, and performance tradeoffs.

* We should continue to support construction from the corresponding
  `std::shared_ptr<bigtable::*Client>`.

With those in mind the proposal can be summarized as: re-implement everything using `*Connection`-style APIs, change
the `bigtable::*Client` to be data-only classes used to construct a `*Connection`.

### Code sketches

A class like `bigtable::Table` will change in its implementation, it will hold a `std::shared_ptr<>`
of `bigtable::Connection`:

```cc
class bigtable::Table {
 public:
[...]

PR #8172, opened 2022-01-31
feat(bigtable): cheap Table creation with different resource name

dbolduc:
Related to #5934

We have customers that use our `Table` clients to interact with tables spread across various instances and projects. Because the project and instance ids are stored in the `DataClient`, the customer is forced to create multiple connections, which is redundant and slow. While this issue will be addressed as part of the effort to modernize the Data API, that will take time. This PR doesn't hurt, and it adds value now.

This PR stores the project and instance ids in the `Table`. Customers can now produce a copy of a `Table` with a modified resource name, without needing to make a new connection to the service.

```cc
// Old code
auto t1 = Table(MakeDataClient("project-1", "instance-A"), "table-X");
auto t2 = Table(MakeDataClient("project-2", "instance-B"), "table-Y"); // initiates an extra connection

// New code
auto t1 = Table(MakeDataClient("project-1", "instance-A"), "table-X");
auto t2 = t1.WithNewTarget("project-2", "instance-B", "table-Y"); // cheap
```

I think the PR is in line with our thread safety guarantees, but the reviewer should check this. https://github.com/googleapis/google-cloud-cpp/blob/9e5ac6776b638a123455e2c8e60b076e51252acd/google/cloud/bigtable/table.h#L86-L90

Should there be another version of this method that sets the app profile id as well?

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8172)

### DIFF
--- a/google/cloud/bigtable/table.h
+++ b/google/cloud/bigtable/table.h
@@ -366,15 +366,8 @@ class Table {
    */
   Table WithNewTarget(std::string project_id, std::string instance_id,
                       std::string table_id) const {
-    auto table = *this;
-    table.instance_id_ = std::move(instance_id);
-    table.project_id_ = std::move(project_id);
-    table.table_id_ = std::move(table_id);
-    table.table_name_ =
-        TableName(table.project_id_, table.instance_id_, table.table_id_);
-    table.metadata_update_policy_ =
-        MetadataUpdatePolicy(table.table_name_, MetadataParamTypes::TABLE_NAME);
-    return table;
+    return WithNewTarget(std::move(project_id), std::move(instance_id),
+                         app_profile_id_, std::move(table_id));
   }
 
   /**
--- a/google/cloud/bigtable/table.h
+++ b/google/cloud/bigtable/table.h
@@ -361,6 +361,8 @@ class Table {
   /**
    * Returns a Table that reuses the connection and configuration of this
    * Table, but with a different resource name.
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/table.h
```diff
@@ -351,10 +354,26 @@ class Table {
...
   std::string const& table_name() const { return table_name_; }
   std::string const& app_profile_id() const { return app_profile_id_; }
-  std::string const& project_id() const { return client_->project_id(); }
-  std::string const& instance_id() const { return client_->instance_id(); }
+  std::string const& project_id() const { return project_id_; }
+  std::string const& instance_id() const { return instance_id_; }
   std::string const& table_id() const { return table_id_; }
 
+  /// Override the project id
+  void set_project_id(std::string project_id) {
```
coryan, line 362: An alternative would be:

```cc
Table WithNewTarget(std::string instance_id, std::string project_id, std::string table_id) const {
  auto table = *this;
  table.instance_id_ = std::move(instance_id);
  table.project_id_ = std::move(project_id);
  table.table_id_ = std::move(table_id);
  return table;
}
```

coryan (commented):
It strikes me as odd that I can change the project and instance ids, but not the table id?  That seems like a fairly uncommon use-case?

coryan (approved):

PR #8187, opened 2022-01-31
feat(generator): *Client with bidir streaming RPCs

### LINKED ISSUE #7795: Support bidir streaming in the generator

coryan:
For simple uses of bidirectional streaming RPCs we can generate a
`*Client` function for bidirectional streaming RPCs. For more complex
cases we will need to write a custom `*Client` class anyway. Recall that
there are about 10 services that use these RPCs, and some of these
services (Pub/Sub, Pub/Sub Lite, Firestore) are fully hand-crafted
anyway.

Fixes #7795

### DIFF
--- a/generator/integration_tests/golden/tests/golden_kitchen_sink_client_test.cc
+++ b/generator/integration_tests/golden/tests/golden_kitchen_sink_client_test.cc
@@ -31,14 +31,30 @@ namespace {
 
 using ::google::cloud::testing_util::IsProtoEqual;
 using ::google::cloud::testing_util::StatusIs;
+using ::google::test::admin::database::v1::AppendRowsRequest;
+using ::google::test::admin::database::v1::AppendRowsResponse;
 using ::testing::ElementsAreArray;
 using ::testing::UnorderedElementsAreArray;
+
+class MockAppendRowsStream
+    : public AsyncStreamingReadWriteRpc<AppendRowsRequest, AppendRowsResponse> {
+ public:
+  MOCK_METHOD(void, Cancel, (), (override));
+  MOCK_METHOD(future<bool>, Start, (), (override));
+  MOCK_METHOD(future<absl::optional<AppendRowsResponse>>, Read, (), (override));
+  MOCK_METHOD(future<bool>, Write,
+              (AppendRowsRequest const&, grpc::WriteOptions), (override));
+  MOCK_METHOD(future<bool>, WritesDone, (), (override));
+  MOCK_METHOD(future<Status>, Finish, (), (override));
+};
 
 TEST(GoldenKitchenSinkClientTest, CopyMoveEquality) {
   auto conn1 =
       std::make_shared<golden_mocks::MockGoldenKitchenSinkConnection>();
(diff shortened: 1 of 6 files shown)


generator/internal/client_generator.cc
```diff
@@ -343,6 +353,20 @@ Status ClientGenerator::GenerateCc() {
...
   for (google::protobuf::MethodDescriptor const& method : methods()) {
+    if (IsBidirStreaming(method)) {
+      CcPrintMethod(method, __FILE__, __LINE__,
+                    R"""(
+std::unique_ptr<::google::cloud::AsyncStreamingReadWriteRpc<
+    $request_type$,
+    $response_type$>>
+$client_class_name$::Async$method_name$(::google::cloud::Options options) {
+  ::google::cloud::internal::OptionsSpan span(
+      ::google::cloud::internal::MergeOptions(std::move(options), options_));
```
dbolduc, line 364: observation: we do not fully qualify `Options`, `internal::OptionsSpan`, or `internal::MergeOptions` elsewhere in this file.

coryan: Fixed.

generator/integration_tests/golden/tests/golden_kitchen_sink_client_test.cc
```diff
@@ -257,6 +288,62 @@ TEST(GoldenKitchenSinkClientTest, ListServiceAccountKeys) {
   EXPECT_STATUS_OK(response);
 }
 
+TEST(GoldenKitchenSinkClientTest, AsyncAppendRows) {
```
devbww, line 291: Does this count as the "generator golden current options" test you asked for?  I was just going to start looking into it.

coryan: Maybe?  It only tests that bidirectional streaming RPCs do the right thing.  One could imagine that this is done correctly for such RPCs, but incorrectly for regular RPCs, or for streaming reads, or streaming writes, or long running (I would say etc. but I think that is the full list).

devbww: Ack.  I'll work on hitting a call of each variety.

coryan:
@dbolduc PTAL

generator/internal/client_generator.cc
```diff
@@ -360,8 +360,8 @@ std::unique_ptr<::google::cloud::AsyncStreamingReadWriteRpc<
     $request_type$,
     $response_type$>>
 $client_class_name$::Async$method_name$(::google::cloud::Options options) {
```
dbolduc, line 362: nit: `s/::google::cloud::Options/Options/`

dbolduc (approved):

dbolduc (approved):

PR #8189, opened 2022-01-31
docs: introduce new libraries

dbolduc:
I thought the list looked better with things like "Cloud Vision" listed under "C" instead of "V".

Names are in accordance with:
https://cloud.google.com/guides/style/branding-list

The following names  were not obvious. I used my best judgment.
* Anthos GKE API
* Compute Engine OS Config
* Compute Engine OS Login
* Retail

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8189)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -184,6 +184,73 @@ your tests.
 
 [#7835]: https://github.com/googleapis/google-cloud-cpp/issues/7835
 
+### New Libraries
+
+We are introducing client libraries for many (>50) GCP services. These libraries
+are still under development.
+
+<details>
+<summary> Expand to see the full list...</summary>
+<br>
+
+* [Access Approval](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/accessapproval/README.md)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -184,6 +184,73 @@ your tests.
 
 [#7835]: https://github.com/googleapis/google-cloud-cpp/issues/7835
 
+### New Libraries
+
+We are introducing client libraries for many (>50) GCP services. These libraries
+are still under development.
```
coryan, line 190: I would say: "While we do not anticipate any API changes to these libraries before declaring them GA, we are releasing them early in case they elicit some feedback that requires changes."

CHANGELOG.md
```diff
@@ -184,6 +184,73 @@ your tests.
...
+
+<details>
```
coryan, line 192: I think this idea, thanks.

CHANGELOG.md
dbolduc: Fixed and added a bullet for GKE / container

coryan (approved):

PR #8196, opened 2022-01-31
fix(rest): add missing include for std::tolower

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8196)

### DIFF
--- a/google/cloud/internal/rest_response.cc
+++ b/google/cloud/internal/rest_response.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/internal/curl_impl.h"
 #include "google/cloud/log.h"
 #include <nlohmann/json.hpp>
+#include <cctype>
 #include <iostream>
 
 namespace google {
--- a/google/cloud/internal/rest_response.cc
+++ b/google/cloud/internal/rest_response.cc
@@ -17,7 +17,6 @@
 #include "google/cloud/internal/curl_impl.h"
 #include "google/cloud/log.h"
 #include <nlohmann/json.hpp>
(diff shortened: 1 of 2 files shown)


google/cloud/internal/rest_response.cc
```diff
@@ -17,6 +17,7 @@
 #include "google/cloud/internal/curl_impl.h"
 #include "google/cloud/log.h"
 #include <nlohmann/json.hpp>
+#include <cctype>
```
devbww, line 20: I don't see `std::tolower` in the file.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8196)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[google/cloud/internal/rest_response.cc, line 20 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8196#-MumVoYcijgdZORsKbXT-r1-20:-MunaX4T2ZEyyWgOi_Ma:b10p2yp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b92127d67795d89bb96a8a727d4c0096b3bb9f2d/google/cloud/internal/rest_response.cc#L20)):*

I don't see `std::tolower` in the file.
</blockquote></details>

Reverted the changes here; made the change in the correct file.

coryan:
Maybe we should use #8200, I tested that on my VM.

coryan (approved):

PR #8204, opened 2022-02-01
docs(iam): added comments on private_key_data

Sita04:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8204)

### DIFF
--- a/google/cloud/iam/samples/iam_samples.cc
+++ b/google/cloud/iam/samples/iam_samples.cc
@@ -208,6 +208,10 @@ std::string CreateServiceAccountKey(std::vector<std::string> const& argv) {
             TYPE_GOOGLE_CREDENTIALS_FILE,
         google::iam::admin::v1::ServiceAccountKeyAlgorithm::KEY_ALG_RSA_2048);
     if (!response) throw std::runtime_error(response.status().message());
+    // The private_key_data field in the response contains the 
+    // service account key in JSON format.
+    // TODO(Developer): Save the key to a secure location.
+    // You cannot download it later.
     std::cout << "ServiceAccountKey successfully created: "
               << response->DebugString() << "\n";
     return response->name();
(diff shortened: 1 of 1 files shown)


google/cloud/iam/samples/iam_samples.cc
```diff
@@ -208,6 +208,10 @@ std::string CreateServiceAccountKey(std::vector<std::string> const& argv) {
             TYPE_GOOGLE_CREDENTIALS_FILE,
         google::iam::admin::v1::ServiceAccountKeyAlgorithm::KEY_ALG_RSA_2048);
     if (!response) throw std::runtime_error(response.status().message());
+    // The private_key_data field in the response contains the 
+    // service account key in JSON format.
+    // TODO(Developer): Save the key to a secure location.
```
coryan, line 213: We do not leave `TODO(Developer)` marks in our examples.  Also note that these examples are executable as-is, the parameters are received from the command-line, so the developer may not see this comment.  I think instead we should do something like:

```cc
  std::cout << "ServiceAccountKey successfully created: "
                  << response->DebugString() << "\n";
                  <<  "Please save the key in a secure location, as they cannot be downloaded later\n";
```

coryan:
FYI, the robots do not like the extra space at the end of one of your comments.  Public versions of the logs are available from  https://github.com/googleapis/google-cloud-cpp/pull/8204#issuecomment-1026826591

google/cloud/iam/samples/iam_samples.cc
Sita04: Thank you for the insight. Modified the comment.

coryan:
These instructions may help you speed up the edit -> test -> edit cycle:

https://github.com/googleapis/google-cloud-cpp/blob/main/doc/contributor/howto-guide-forks-and-pull-requests.md#reproducing-ci-build-failures

Please refrain from starting more builds until at least `checkers` passes locally on your workstation. We have limited capacity for Windows and macOS builds.

coryan:
Thanks for the PR!  I will fix the build problem and merge this later today. We are in the middle of a release so please don't merge quite yet.

PR #8207, opened 2022-02-01
cleanup: prefer @platforms rather than @bazel_tools

devjgm:
The former is the preferred way to specify platforms. Note that in the
testing_util case, I dropped the support for freebsd and openbsd. I'm
not sure why those cases were there or if we need them. We don't seem to
have those cases elsewhere. I can add them back if they are needed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8207)

### DIFF
--- a/google/cloud/testing_util/BUILD.bazel
+++ b/google/cloud/testing_util/BUILD.bazel
@@ -16,7 +16,6 @@ package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])  # Apache 2.0
 
-load("@bazel_skylib//lib:selects.bzl", "selects")
 load(":google_cloud_cpp_testing.bzl", "google_cloud_cpp_testing_hdrs", "google_cloud_cpp_testing_srcs")
 
 config_setting(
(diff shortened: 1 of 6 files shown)


google/cloud/testing_util/BUILD.bazel
```diff
@@ -44,22 +43,12 @@ cc_library(
...
+        "@platforms//os:linux": [
             "GOOGLE_CLOUD_CPP_HAVE_GETRUSAGE",
             "GOOGLE_CLOUD_CPP_HAVE_RUSAGE_THREAD",
         ],
-        (
-            "@bazel_tools//src/conditions:darwin",
-            "@bazel_tools//src/conditions:freebsd",
-            "@bazel_tools//src/conditions:openbsd",
-        ): [
+        "@platforms//os:macos": [
```
coryan, line 51: As I am sure you know, macOS != FreeBSD and != OpenBSD.  So this is not strictly preserving the previous behavior.  OTOH, I am not sure the previous approach worked. 🤷   I think we should make this change and fix it if it is a problem.

coryan (approved):

google/cloud/testing_util/BUILD.bazel
devjgm: Yes, I realize this is not preserving behavior. But I wasn't sure why that behavior was needed or why the special cases for freebsd and openbsd were added. We don't claim to support those platforms, so I'm not sure if any of it actually works. I'm happy to add those cases back if we want.

coryan: Sorry to be unclear, I think we should merge as-is.

PR #8225, opened 2022-02-03
cleanup: simplify Bazel targets for new libraries

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8225)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -20,117 +20,126 @@ exports_files([
     "LICENSE",
 ])
 
-cc_library(
-    name = "experimental-accessapproval",
-    deps = [
-        "//google/cloud/accessapproval:google_cloud_cpp_accessapproval",
-    ],
-)
+EXPERIMENTAL_LIBRARIES = [
+    "accessapproval",
+    "accesscontextmanager",
+    "apigateway",
+    "appengine",
+    "artifactregistry",
+    "asset",
+    "assuredworkloads",
+    "automl",
+    "billing",
+    "binaryauthorization",
+    "channel",
+    "cloudbuild",
+    "composer",
+    "container",
+    "containeranalysis",
+    "datacatalog",
+    "datamigration",
+    "debugger",
+    "dlp",
+    "eventarc",
+    "filestore",
+    "functions",
+    "gameservices",
+    "gkehub",
+    "iap",
+    "ids",
+    "iot",
+    "kms",
+    "logging",
+    "managedidentities",
+    "memcache",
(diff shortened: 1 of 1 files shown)


BUILD.bazel
```diff
@@ -20,117 +20,126 @@ exports_files([
...
+]
+
+[cc_library(
+    name = "experimental-{library}".format(library = library),
+    deps = [
+        "//google/cloud/{library}:google_cloud_cpp_{library}".format(library = library),
+    ],
+) for library in EXPERIMENTAL_LIBRARIES]
+
+[cc_library(
```
dbolduc, line 105: optionally, you can combine these into the same for loop

BUILD.bazel
```diff
@@ -20,117 +20,126 @@ exports_files([
...
+    name = "experimental-{library}".format(library = library),
+    deprecation = "this library is now GA, please use //:{library} instead.".format(library = library),
+    tags = ["manual"],
     deps = [
-        "//google/cloud/assuredworkloads:google_cloud_cpp_assuredworkloads",
+        "//google/cloud/{library}:google_cloud_cpp_{library}".format(library = library),
     ],
-)
+) for library in TRANSITION_LIBRARIES]
```
dbolduc, line 119: same option

BUILD.bazel
```diff
@@ -20,117 +20,126 @@ exports_files([
...
+    "workflows",
+]
+
+TRANSITION_LIBRARIES = [
+    "secretmanager",
+    "tasks",
+]
+
+GA_LIBRARIES = [
+]
```
dbolduc, line 96: do you want to add `bigquery`, `iam`, `pubsub`, `spanner` to here?

(leaving their deprecated targets below)

dbolduc (approved):
just some options that you might have already considered and decided against

BUILD.bazel
coryan: I do not think you can, these are list comprehensions, if you write `[x(a) y(a) for a in list]` that is a syntax error, if you write `[(x(a), y(a)) for a in list]` that is a list of tuples, and you don't want tuples...

BUILD.bazel
coryan: I am leaving those for a later refactoring.

BUILD.bazel
coryan: Ditto, but if you know of a Python / Starlark trick to get these working I would be glad to know.

PR #8227, opened 2022-02-03
feat(language): generate library

### LINKED ISSUE #8185: Cloud Natural Language API: `language.googleapis.com`

coryan:
Fixes #8185

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8227)


generator/generator_config.textproto
```diff
@@ -385,6 +385,14 @@ service {
   retryable_status_codes: ["kDeadlineExceeded", "kUnavailable"]
 }
 
+# Language
+service {
+  service_proto_path: "google/cloud/language/v1/language_service.proto"
+  product_path: "google/cloud/language"
+  initial_copyright_year: "2022"
+  retryable_status_codes: ["kUnavailable"]
```
dbolduc, line 393: https://github.com/googleapis/googleapis/blob/2b37083027fcfa408ee75eee4264cad7c2de0a67/google/cloud/language/v1/language_grpc_service_config.json#L35-L38

I dont remember where we landed with the AIP vs. service config json rules. But I think if this one doesn't have `kDeadlineExceeded`, then no service should have it...

dbolduc (changes requested):

google/cloud/language/quickstart/quickstart.cc
```diff
@@ -0,0 +1,51 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/language/language_client.h"
+#include <iostream>
+#include <stdexcept>
+
+auto constexpr kText = R"""(
```
dbolduc, line 19: forgot to add to review:

it's weird to start the string with a `'\n'`, but who cares.

generator/generator_config.textproto
coryan: I don't think we reached a decision, but I think (a) we should never have `kDeadlineExceeded`, and (b) this is easy to change after the fact.

dbolduc: Ack.

dbolduc (approved):

PR #8229, opened 2022-02-03
feat(impl): Create a generic helper for async continuations

dpcollins-google:
This is the equivalent of guava's Futures.transformAsync (https://guava.dev/releases/21.0/api/docs/com/google/common/util/concurrent/Futures.html#transformAsync-com.google.common.util.concurrent.ListenableFuture-com.google.common.util.concurrent.AsyncFunction-) and makes it easier to write sequences of asynchronous operations

This is intended for internal use only, not to add to the external api, even if doing so would lead to a cleaner chaining (with .then_async)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8229)

### DIFF
--- a/google/cloud/internal/future_then_async.h
+++ b/google/cloud/internal/future_then_async.h
@@ -1,0 +1,89 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_FUTURE_THEN_ASYNC_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_FUTURE_THEN_ASYNC_H
+/**
+ * @file
+ *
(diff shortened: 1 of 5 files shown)


google/cloud/internal/future_then_async.h
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * This function allows for generic asynchronous continuations using
+ * google::cloud::future. In particular, it allows chaining of async
+ * operations without increasing the nesting level.
+ * 
+ * Usage of this utility would look like the below, which is a series of
+ * asynchronous operations written in an almost linear style:
+ * 
+ * future<std::string> f1 = ...
+ * future<int> f2 = then_async(std::move(f1), [](std::string value) -> future<int> {
```
coryan, line 30: Note that `google::cloud::future<>` already unwraps:

https://github.com/googleapis/google-cloud-cpp/blob/6600f0fda6f6da0c8d248beadbc519db23310f94/google/cloud/future_generic.h#L95-L96

I think your implementation could be:

```cc
template <typename T, typename F>
auto then_async(future<T> f, F&& f) /* -> this could be ugly, look at internal::then_helper<F, T>::future_t */ {
  return f.then([g = std::forward<Functor>(f)](future<T> h) { return g(h.get()); });
}
```

google/cloud/internal/future_then_async.h
```diff
@@ -0,0 +1,89 @@
...
+template <class T>
+using future_value_type_t = typename future_value_type<T>::type;
+
+void link_future_and_promise(future<void> fut, promise<void> p) {
+    fut.then([p]{ p.set_value(); });
+}
+
+template <class T>
+void link_future_and_promise(future<T> fut, promise<T> p) {
+    fut.then([p](T val){ p.set_value(std::move(val)); });
```
coryan, line 63: `promise<T>` is move-only, and this is C++ 11, you need to implement a helper class to move-capture.

coryan (changes requested):

PR #8231, opened 2022-02-03
feat(dataproc): generate library

### LINKED ISSUE #8182: Cloud Dataproc API: `dataproc.googleapis.com`

dbolduc:
Fixes #8182 

The [service config json](https://github.com/googleapis/googleapis/blob/master/google/cloud/dataproc/v1/dataproc_grpc_service_config.json) has different retryable codes per RPC. I went with the strictest list for each service, and discounted `kDeadlineExceeded`. This left everything as `kUnavailable`.

The endpoint is interesting. I did not see it in documentation, but if you send a request with a region, the service responds that you should be using the regional endpoint. [Other client libraries do the same thing](https://cloud.google.com/dataproc/docs/reference/libraries#using_the_client_library).

I enabled the Dataproc API and created two clusters in region "us-central1" using the UI. (as described in [this quickstart](https://cloud.google.com/dataproc/docs/quickstarts/quickstart-explorer-create)).

```console
# Command
bazel run //google/cloud/dataproc/quickstart:quickstart -- dbolduc-test us-central1

# Output
cluster-6f37
cluster-8d27
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8231)


google/cloud/dataproc/README.md
```diff
@@ -38,24 +42,34 @@ this library.
...
 
-  auto const project = google::cloud::Project(argv[1]);
-  for (auto r : client.List /*EDIT HERE*/ (project.FullName())) {
-    if (!r) throw std::runtime_error(r.status().message());
-    std::cout << r->DebugString() << "\n";
+  std::string const project_id = argv[1];
+  std::string const region = argv[2];
+
+  gc::Options options;
+  if (region != "global") {
```
coryan, line 63: Can we hard-code the region to "global" and just ask for the project-id in the command-line?

google/cloud/dataproc/README.md
```diff
@@ -3,7 +3,11 @@
 :construction:
 
 This directory contains an idiomatic C++ client library for the
-[Cloud Dataproc API][cloud-service-docs], a service to Manages Hadoop-based clusters and jobs on Google Cloud Platform.
+[Cloud Dataproc API][cloud-service-docs], a fully managed and highly scalable
+service for running Apache Spark, Apache Flink, Presto, and 30+ open source
+tools and frameworks. Use Dataproc for data lake modernization, ETL, and secure
+data science, at planet scale, fully integrated with Google Cloud, at a fraction
+of the cost.
```
coryan, line 11: I think this description may give the impression that you can run C++ applications in Cloud Dataproc, which I think you cannot?  Maybe we should say something like: "Use this library to manage Cloud Dataproc clusters and other resources, to run C++ applications in Cloud Dataproc we recommend you use ..." (if we have a thing to fill the blank), or say "Note that this library allows you to *manage* Cloud Dataproc resources, it does not provide APIs to run C++ applications in Cloud Dataproc".

coryan (commented):
Looks pretty good, a couple of comments / questions.

google/cloud/dataproc/README.md
dbolduc: I would not. Setting `region="global"` does not list all clusters in any region. "global" itself is a region (like "us-central1"), so it would only list clusters that were created in the "global" region. When you create a cluster using the UI, the default region is something like "us-central1", not "global".

coryan: Yeah... I was trying to simplify the code / save vertical space.

In any case, this suggests we need a helper function (for a future PR), something like:

```cc
namespace google::cloud::dataproc {
gc::Options WithRegionalEndpoint(Options options, std::string const& region);
}
```

dbolduc: Ack, I see the issue you opened.

dbolduc:
Our `publish-docs` build fails because `<tbody>` is not handled by Doxygen. See https://www.doxygen.nl/manual/htmlcmds.html

I decided to just change the generator in this PR to comment out the tag. See the "fix doxygen" commit.

PTAL

coryan (approved):

PR #8236, opened 2022-02-04
fix(generator): Connection base-class operations should fail

devbww:
While we do not test the behavior of the generated "ServiceConnection"
base classes, all operations should fail with `kUnimplemented`. That
was not the case for paginated calls, which previously returned an
empty sequence instead.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8236)


dbolduc (changes requested):
Can we return this instead:
```cc
return google::cloud::internal::MakeUnimplementedPaginationRange<
    StreamRange<$range_output_type$>>();
```

https://github.com/googleapis/google-cloud-cpp/blob/306d12ba8d23ac9897c06a1c0e706bdef47d7ded/google/cloud/internal/pagination_range.h#L182-L189

where maybe we change the message to "not implemented" instead of "needs-override"

devbww:
> Can we return this instead:

Done.  PTAL.

devbww:
Note: Unfortunately the change in google/cloud/internal/pagination_range.h from "needs-override" to "not implemented" was incorrectly grouped in the "regeneration" commit ef7628a.

dbolduc (approved):

Issue #7936, opened 2022-01-13
We should run generated quickstarts


dbolduc:
We compile generated quickstarts, but we never run them. https://github.com/googleapis/google-cloud-cpp/blob/ac535f71e6c1baa3a81caa1fb565d5d0e29e1133/ci/etc/quickstart-config.sh#L25-L32

It will also take some effort to figure out what arguments should be used, and whether any setup is necessary to run a given quickstart.

----

**Cleanup**

- [x] Change `GOOGLE_CLOUD_CPP_TEST_REGION` from `GOOGLE_CLOUD_REGION`. The former matches our other test variable names. The latter was me not thinking and copying `GOOGLE_CLOUD_PROJECT` which we use because it is used in other libraries and tools.

After #8254 the following libraries will need work:

**Cannot run in our CI Builds**

- [x] `google/cloud/assuredworkloads/`:
- [x] `google/cloud/servicecontrol/`
- [x] `google/cloud/accesscontextmanager/`:
- [x] `google/cloud/policytroubleshooter/`: 
- [x] `google/cloud/channel/`: requires a partner account id

**Requires Custom Resources**

- [x] `google/cloud/debugger/`
- [x] google/cloud/appengine/
- [x] google/cloud/gameservices/
- [x] google/cloud/iot/
- [x] google/cloud/recommender/
- [x] google/cloud/resourcemanager/
- [x] google/cloud/talent/

**Requires Permissions unavailable to a Service Account**

- [x] google/cloud/shell/
- [x] google/cloud/oslogin/ (with my credentials using `$USER@google.com` worked)

**Requires X-User-Project** (#8245) 

- [x] `google/cloud/securitycenter/`
- [x] google/cloud/accessapproval/
- [x] google/cloud/automl/
- [x] google/cloud/dlp/
- [x] google/cloud/language/
- [x] google/cloud/orgpolicy/
- [x] google/cloud/retail/
- [x] google/cloud/texttospeech/
- [x] google/cloud/translate/
- [x] google/cloud/videointelligence/
- [x] google/cloud/vision/
- [x] google/cloud/webrisk/
- [x] google/cloud/asset/

**Hangs**

- [x] google/cloud/cloudbuild/

**Fails**

- [x] google/cloud/pubsublite/
- [x] google/cloud/workflows/

coryan:
## Proposal

I will change the generator to create test targets for the quickstart programs.  These test targets will:

[...]

PR #8245, opened 2022-02-04
feat(generator): support `x-goog-user-project`

coryan:
Many services cannot be used directly from a user account. User accounts
have no default project associated with them for quota and billing, and
some services do not have any resources attached to a specific project
either. For example, compute-only services like Text-to-Speech. In this
case the recommendation is to use the x-goog-user-project header, or use
service account impersonation. The latter is fairly difficult, and would
make the "quickstart" not very quick.

This PR introduces a new option to configure this header. The default
comes from an environment variable.

Part of the work for #7936, or at least motivated by it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8245)

### DIFF
--- a/generator/internal/option_defaults_generator.cc
+++ b/generator/internal/option_defaults_generator.cc
@@ -92,12 +92,19 @@ Status OptionDefaultsGenerator::GenerateCc() {
   //  clang-format on
 
   CcPrint(  // clang-format off
-  {{"\n"
-    "Options $service_name$DefaultOptions(Options options) {\n"
-    "  if (!options.has<EndpointOption>()) {\n"
-    "    auto env = internal::GetEnv(\"$service_endpoint_env_var$\");\n"
-    "    options.set<EndpointOption>(env && !env->empty() ? *env : \"$service_endpoint$\");\n"
-    "  }\n"},
+  {{R"""(
+Options $service_name$DefaultOptions(Options options) {
+  if (!options.has<EndpointOption>()) {
+    auto env = internal::GetEnv("$service_endpoint_env_var$");
+    options.set<EndpointOption>(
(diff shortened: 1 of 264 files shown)


scotthart (approved):

generator/internal/option_defaults_generator.cc
```diff
@@ -92,12 +92,19 @@ Status OptionDefaultsGenerator::GenerateCc() {
...
+  {{R"""(
+Options $service_name$DefaultOptions(Options options) {
+  if (!options.has<EndpointOption>()) {
+    auto env = internal::GetEnv("$service_endpoint_env_var$");
+    options.set<EndpointOption>(
+        env && !env->empty() ? *env : "$service_endpoint$");
+  }
+  if (!options.has<UserProjectOption>()) {
+    auto env = internal::GetEnv("GOOGLE_CLOUD_CPP_USER_PROJECT");
+    if (env.has_value() && !env->empty()) options.set<UserProjectOption>(*env);
```
devbww, line 104: We should probably use either `env` or `env.has_value()` consistently when they're this close together.

generator/internal/option_defaults_generator.cc
```diff
@@ -92,12 +92,19 @@ Status OptionDefaultsGenerator::GenerateCc() {
...
+  }
```
devbww, line 105: Should we be adding this to the non-generated default options?

coryan: Yes.  I am planning to do that.

generator/internal/option_defaults_generator.cc
coryan: Ack.  I will send a PR to clean this up, I think there is some opportunity for refactoring too.

PR #8247, opened 2022-02-04
ci: reduce CTest output on interactive builds

coryan:
This will make it easier to find stuff when running the builds locally.
I expect the CI builds are not affected, and still have long outputs. I
had to refactor some stuff in the CI scripts to make sure ctest got the
same args everywhere, that required favoring `-j` over `--parallel`
because `--parallel` is not supported by the oldest version of CTest we
happen to use.

I also found that gcc-54.sh was still around, but unused, easier to
remove it than fixing it.

With these changes running the unit test locally produces this output:

```
Test project /workspace/cmake-out
412/412 Test  #41: common_completion_queue_benchmark
100% tests passed, 0 tests failed out of 412
```

Previously it generated over 800 lines of informational messages.
We still get some progress (`CTest` overwrites the same line over and
over), and we still get the full errors on failure.

You are welcome.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8247)


ci/cloudbuild/builds/cmake-oldest-deps.sh
```diff
@@ -36,6 +36,18 @@ io::log_h2 "Building"
...
 
 io::log_h2 "Testing"
-env -C cmake-out/build ctest -LE "integration-test" --parallel "$(nproc)"
+
+ctest_flags=(
+  # Print the full output on failures
+  --output-on-failure
+  # Skip tests with the "integration-test" label, the intent is to run only
+  # unit tests
+  -LE "integration-test"
```
devbww, line 45: Perhaps this should be factored out so that the `ctest` invocation looks the same as in the other files?

ci/cloudbuild/builds/cmake-oldest-deps.sh
```diff
@@ -36,6 +36,18 @@ io::log_h2 "Building"
 cmake --build cmake-out/build
 
 io::log_h2 "Testing"
-env -C cmake-out/build ctest -LE "integration-test" --parallel "$(nproc)"
+
+ctest_flags=(
```
devbww, line 40: Perhaps `s/ctest_flags/ctest_args/` to match the others files?

ci/cloudbuild/builds/gcc-63.sh
```diff
@@ -28,7 +28,8 @@ cmake -GNinja -H. -Bcmake-out \
   -DGOOGLE_CLOUD_CPP_ENABLE="$(features::always_build_cmake)" \
   -DBUILD_SHARED_LIBS=yes -DGOOGLE_CLOUD_CPP_ENABLE_CCACHE=ON
 cmake --build cmake-out
+mapfile -t ctest_args < <(ctest::common_args)
 (
   cd cmake-out
-  ctest --output-on-failure -LE "integration-test" -j "$(nproc)"
+  ctest "${ctest_args[@]}" -LE "integration-test"
```
devbww, line 34: Can we use `env -C cmake-out ctest ...`, instead of the sub-shell, to match the other files?

ci/kokoro/macos/builds/cmake-vcpkg.sh
```diff
@@ -71,9 +71,17 @@ io::log_h2 "Compiling with ${NCPU} cpus"
...
+  --output-on-failure
+  # Run many tests in parallel, use -j for compatibility with old versions
+  -j "$(nproc)"
+  # Make the output shorter on interactive tests
+  --progress
+)
 (
   cd "${BINARY_DIR}"
-  ctest -LE integration-test --output-on-failure -j "${NCPU}"
+  ctest "${ctest_args[@]}" -LE "integration-test"
```
devbww, line 84: Can we use env -C "${BINARY_DIR}" ctest ..., instead of the sub-shell, to match the other files?

ci/kokoro/macos/builds/cmake-vcpkg.sh
```diff
@@ -101,10 +109,9 @@ if should_run_integration_tests; then
     export CLOUD_STORAGE_ENABLE_TRACING="raw-client"
 
     cd "${BINARY_DIR}"
```
devbww, line 111: Ditto on `env -C`.  We could even use `env` to do all those `export`s.

ci/kokoro/windows/build-cmake.ps1
```diff
@@ -88,7 +88,12 @@ if (Test-Path env:RUNNING_CI) {
 # Get the number of processors to parallelize the tests
 $NCPU=(Get-CimInstance Win32_ComputerSystem).NumberOfLogicalProcessors
 
-$ctest_flags = @("--output-on-failure", "-j", $NCPU, "-C", $env:CONFIG)
+$ctest_flags = @(
```
devbww, line 91: `s/ctest_flags/ctest_args/` to match?

devbww:
Thanks.

> You are welcome.

ci/cloudbuild/builds/cmake-oldest-deps.sh
coryan: Done

ci/cloudbuild/builds/cmake-oldest-deps.sh
coryan: Done

ci/cloudbuild/builds/gcc-63.sh
coryan: Unfortunately we cannot, this runs on Ubuntu:16.04 where `env` does not support `-C`.  I will add a comment.

ci/kokoro/macos/builds/cmake-vcpkg.sh
coryan: Same thing, macOS does not support this flag (just tried on my laptop).

ci/kokoro/macos/builds/cmake-vcpkg.sh
coryan: I moved the exports, but we cannot do the -C

ci/kokoro/windows/build-cmake.ps1
coryan: Done, and also s/cmake_flags/cmake_args/ so the file is internally consistent.

devbww (approved):

PR #8254, opened 2022-02-05
fix: make quickstarts run and add to CI

coryan:
Another pass cleaning up quickstart programs and adding them to the CI
builds.

I know I am not being very systematic on these, maybe I should be sending one PR per quickstart or something.  But I think speed is more important than carefully curated code archaeological records in this case.

Part of the work for #7936

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8254)

### DIFF
--- a/google/cloud/composer/README.md
+++ b/google/cloud/composer/README.md
@@ -55,7 +55,7 @@ int main(int argc, char* argv[]) try {
       composer::EnvironmentsClient(composer::MakeEnvironmentsConnection());
 
   auto const parent =
-      std::string("projects/") + argv[0] + "/locations/" + argv[1];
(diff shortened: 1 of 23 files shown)


google/cloud/composer/README.md
```diff
@@ -55,7 +55,7 @@ int main(int argc, char* argv[]) try {
       composer::EnvironmentsClient(composer::MakeEnvironmentsConnection());
 
   auto const parent =
-      std::string("projects/") + argv[0] + "/locations/" + argv[1];
```
dbolduc, line 58: ugh, this was me, without a doubt.

coryan: No worries, that is why you asked for testing.

dbolduc (approved):

PR #8255, opened 2022-02-05
fix(oauth2): use correct expiration slack value

jmouradi-google:
In the client library, OAuth tokens are currently cached until they have 500s of time remaining. However, as per [public documentation](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances), "The metadata server caches access tokens until they have 5 minutes of remaining time..."

This leaves a 200s-window in which RefreshingCredentialsWrapper objects can repeatedly attempt to refresh credentials from the metadata server, but receive the same token each time. 

This pull request corrects the value of slack from 500s to 300s to prevent this from happening.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8255)


coryan:
Thanks for sending a PR.  LGTM, let me start the builds for you.

coryan (approved):

PR #8256, opened 2022-02-05
fix: add Options save/restore to StreamRange<T>

### LINKED ISSUE #8248: Capture options span for long running operations and paginated APIs.
Consider a loop like:

```cc
for (auto a : client.ListAssets(...)) {
  // do stuff with `a`
}
```

The `ListAssets()` call creates an `OptionsSpan`:

https://github.com/googleapis/google-cloud-cpp/blob/4074a0e3479f45e9ee2b257d41089a2787e6786b/google/cloud/asset/asset_client.cc#L48-L52

the next call is:

https://github.com/googleapis/google-cloud-cpp/blob/4074a0e3479f45e9ee2b257d41089a2787e6786b/google/cloud/asset/internal/asset_connection_impl.cc#L71-L100

That defines a lambda:

https://github.com/googleapis/google-cloud-cpp/blob/4074a0e3479f45e9ee2b257d41089a2787e6786b/google/cloud/asset/internal/asset_connection_impl.cc#L84-L92

The first time this lambda is called it happens in the context of the `*Client::ListAssets()` invocation, and thus has the right options span.  Subsequent calls are triggered when `operator++` has consumed the first page of results.  At that time the relevant `OptionsSpan` is no longer available.

The `*Connection::ListAssets()` function should capture the span for the first call, and reuse it for future calls.

devbww:
Save the options in force during `StreamRange<T>` construction,
and then restore them over `internal::StreamReader<T>` callbacks.
This means that streaming/paginated operations will have the right
options installed during all their RPCs.

As well as adding tests to check this behavior for `StreamRange`
and `MakePaginationRange`, also add checks to verify the matching,
existing behavior for `RetryLoop`, `AsyncLongRunningOperation`,
and `AsyncPollingLoop`.

Fixes #8248.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8256)

### DIFF
--- a/google/cloud/internal/async_polling_loop.cc
+++ b/google/cloud/internal/async_polling_loop.cc
@@ -46,7 +46,7 @@ class AsyncPollingLoopImpl
     auto self = shared_from_this();
     auto w = WeakFromThis();
     auto const& options = CurrentOptions();
-    promise_ = promise<StatusOr<Operation>>([w, options] {
+    promise_ = promise<StatusOr<Operation>>([w, options]() mutable {
       if (auto self = w.lock()) {
         OptionsSpan span(std::move(options));
         self->DoCancel();
--- a/google/cloud/internal/async_polling_loop.cc
+++ b/google/cloud/internal/async_polling_loop.cc
@@ -45,8 +45,12 @@ class AsyncPollingLoopImpl
   future<StatusOr<Operation>> Start(future<StatusOr<Operation>> op) {
     auto self = shared_from_this();
     auto w = WeakFromThis();
-    promise_ = promise<StatusOr<Operation>>([w] {
-      if (auto self = w.lock()) self->DoCancel();
+    auto const& options = CurrentOptions();
+    promise_ = promise<StatusOr<Operation>>([w, options] {
+      if (auto self = w.lock()) {
+        OptionsSpan span(std::move(options));
+        self->DoCancel();
+      }
     });
     op.then([self](future<StatusOr<Operation>> f) { self->OnStart(f.get()); });
     return promise_.get_future();
--- a/google/cloud/internal/async_polling_loop_test.cc
+++ b/google/cloud/internal/async_polling_loop_test.cc
@@ -146,8 +146,12 @@ TEST(AsyncPollingLoopTest, ImmediateCancel) {
   auto pending =
       AsyncPollingLoop(cq, p.get_future(), MakePoll(mock), MakeCancel(mock),
                        std::move(policy), "test-function");
-  pending.cancel();
+  {
+    OptionsSpan overlay(Options{}.set<StringOption>("uh-oh"));
+    pending.cancel();
+  }
   p.set_value(starting_op);
+  OptionsSpan overlay(Options{}.set<StringOption>("uh-oh"));
   auto actual = pending.get();
   EXPECT_THAT(actual, StatusIs(StatusCode::kCancelled,
                                AllOf(HasSubstr("test-function"),
@@ -184,11 +188,11 @@ TEST(AsyncPollingLoopTest, PollThenSuccess) {
   EXPECT_CALL(*policy, WaitPeriod)
       .WillRepeatedly(Return(std::chrono::milliseconds(1)));
   OptionsSpan span(Options{}.set<StringOption>("PollThenSuccess"));
-  auto actual =
-      AsyncPollingLoop(cq, make_ready_future(make_status_or(starting_op)),
-                       MakePoll(mock), MakeCancel(mock), std::move(policy),
-                       "test-function")
-          .get();
+  auto pending = AsyncPollingLoop(
+      cq, make_ready_future(make_status_or(starting_op)), MakePoll(mock),
+      MakeCancel(mock), std::move(policy), "test-function");
+  OptionsSpan overlay(Options{}.set<StringOption>("uh-oh"));
+  auto actual = pending.get();
   ASSERT_THAT(actual, IsOk());
   EXPECT_THAT(*actual, IsProtoEqual(expected));
 }
@@ -272,11 +276,11 @@ TEST(AsyncPollingLoopTest, PollThenEventualSuccess) {
   EXPECT_CALL(*policy, WaitPeriod)
       .WillRepeatedly(Return(std::chrono::milliseconds(1)));
   OptionsSpan span(Options{}.set<StringOption>("PollThenEventualSuccess"));
-  auto actual =
-      AsyncPollingLoop(cq, make_ready_future(make_status_or(starting_op)),
-                       MakePoll(mock), MakeCancel(mock), std::move(policy),
-                       "test-function")
-          .get();
+  auto pending = AsyncPollingLoop(
+      cq, make_ready_future(make_status_or(starting_op)), MakePoll(mock),
+      MakeCancel(mock), std::move(policy), "test-function");
+  OptionsSpan overlay(Options{}.set<StringOption>("uh-oh"));
+  auto actual = pending.get();
   ASSERT_THAT(actual, IsOk());
   EXPECT_THAT(*actual, IsProtoEqual(expected));
 }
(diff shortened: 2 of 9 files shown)


coryan (approved):
Thanks!  I feared this would require changes to the generator, but this is much cleaner.

devbww:
Unfortunately the tests were a little deficient .. they should override/undo the call-time `OptionsSpan` before they do anything with the operation result.  Fixing that revealed a problem in the "async" path.  I'm looking into it.

devbww:
Added an options restore to the cancellation path.  PTAL.

google/cloud/internal/async_polling_loop.cc
```diff
@@ -45,8 +45,12 @@ class AsyncPollingLoopImpl
   future<StatusOr<Operation>> Start(future<StatusOr<Operation>> op) {
     auto self = shared_from_this();
     auto w = WeakFromThis();
-    promise_ = promise<StatusOr<Operation>>([w] {
-      if (auto self = w.lock()) self->DoCancel();
+    auto const& options = CurrentOptions();
+    promise_ = promise<StatusOr<Operation>>([w, options] {
+      if (auto self = w.lock()) {
+        OptionsSpan span(std::move(options));
```
coryan, line 51: The `operator()` is `const` in this lambda, the `std::move()` has no effect.  Do you want a mutable lambda?

google/cloud/internal/async_retry_loop.h
```diff
@@ -189,8 +189,12 @@ class AsyncRetryLoopImpl
 
   future<T> Start() {
     auto weak = std::weak_ptr<AsyncRetryLoopImpl>(this->shared_from_this());
-    result_ = promise<T>([weak] {
-      if (auto self = weak.lock()) self->Cancel();
+    auto const& options = CurrentOptions();
+    result_ = promise<T>([weak, options] {
+      if (auto self = weak.lock()) {
+        OptionsSpan span(std::move(options));
```
coryan, line 195: Ditto

coryan (commented):
I am too tired to do a full review, but I caught a couple of things.

google/cloud/internal/async_polling_loop.cc
devbww: Good catch!

I see you disabled the "performance-move-const-arg" clang-tidy in #3458, which I approved, so I guess we just need to be careful.

google/cloud/internal/async_polling_loop.cc
```diff
@@ -45,8 +45,12 @@ class AsyncPollingLoopImpl
...
-    promise_ = promise<StatusOr<Operation>>([w] {
-      if (auto self = w.lock()) self->DoCancel();
+    auto const& options = CurrentOptions();
+    promise_ = promise<StatusOr<Operation>>([w, options]() mutable {
+      if (auto self = w.lock()) {
+        OptionsSpan span(std::move(options));
+        self->DoCancel();
+      }
     });
     op.then([self](future<StatusOr<Operation>> f) { self->OnStart(f.get()); });
```
coryan, line 55: I think we want `OnStart()` to run with the options in effect during `Start()`?  If so, it seems that we want to capture `CurrentOptions()` in the constructor of this class, and create a span with them on each `OnFoo()` callback

coryan: See for example lines 79, and 123.

google/cloud/internal/async_polling_loop_test.cc
```diff
@@ -234,31 +247,40 @@ TEST(AsyncPollingLoopTest, PollThenEventualSuccess) {
...
-  auto actual =
-      AsyncPollingLoop(cq, make_ready_future(make_status_or(starting_op)),
-                       MakePoll(mock), MakeCancel(mock), std::move(policy),
-                       "test-function")
-          .get();
+  OptionsSpan span(Options{}.set<StringOption>("PollThenEventualSuccess"));
+  auto pending = AsyncPollingLoop(
+      cq, make_ready_future(make_status_or(starting_op)), MakePoll(mock),
+      MakeCancel(mock), std::move(policy), "test-function");
+  OptionsSpan overlay(Options{}.set<StringOption>("uh-oh"));
```
coryan, line 282: Unfortunately I think this test does not catch the problems because all the asynchronous callbacks run immediately.  I think you want to use an `AsyncSequencer<T>` to gain finer grained control of the steps in this test, and reset the Options span each time.  This outline will probably need some work:

```cc
TEST(AsyncPollingLoopTest, PollThenEventualSuccess) {
  // NEW
  AsyncSequencer<Status> sequencer;

  Instance instance;
  instance.set_name("test-instance-name");
  google::longrunning::Operation starting_op;
  starting_op.set_name("test-op-name");
  google::longrunning::Operation expected = starting_op;
  expected.set_done(true);
  expected.mutable_metadata()->PackFrom(instance);
  auto mock_cq = std::make_shared<MockCompletionQueueImpl>();
  EXPECT_CALL(*mock_cq, MakeRelativeTimer)
      .WillRepeatedly([&](std::chrono::nanoseconds) {
        return sequencer.PushBack("MakeRelativeTimer").then([](future<Status> f) {
          auto status = f.get();
          if (status.ok()) return make_status_or(std::chrono::system_clock::now());
          return status; // Maybe needs an explicit conversion?
      });
  CompletionQueue cq(mock_cq);

  auto mock = std::make_shared<MockStub>();
  EXPECT_CALL(*mock, AsyncGetOperation)
      .WillOnce([&](CompletionQueue&, std::unique_ptr<grpc::ClientContext>,
                    google::longrunning::GetOperationRequest const&) {
        EXPECT_EQ(CurrentOptions().get<StringOption>(),
                  "PollThenEventualSuccess");
        return sequencer.PushBack("GetOperation").then([&](future<Status> f) {
          auto status = f.get();
          if (status.ok()) return make_status_or(starting_op);
          return status; // Maybe needs an explicit conversion?
        });
      })
... ... ...

  auto policy = absl::make_unique<MockPollingPolicy>();
  EXPECT_CALL(*policy, clone()).Times(0);
  EXPECT_CALL(*policy, OnFailure).WillRepeatedly(Return(true));
  EXPECT_CALL(*policy, WaitPeriod)
      .WillRepeatedly(Return(std::chrono::milliseconds(1)));
  OptionsSpan span(Options{}.set<StringOption>("PollThenEventualSuccess"));
  auto pending = AsyncPollingLoop(
      cq, make_ready_future(make_status_or(starting_op)), MakePoll(mock),
      MakeCancel(mock), std::move(policy), "test-function");

  // Now you can control the execution of the polling loop one callback at a time.

  // TODO this may not be a good example, but a good start maybe, with some C++17 liberties
  auto [p, name] = sequencer.PopFrontWithName();
  EXPECT_EQ(name, "GetOperation");
  OptionsSpan overlay(Options{}.set<StringOption>("uh-oh"));
  p.set_value(Status(StatusCode::kUnavailable, "try-again"));

  [p, name] = sequencer.PopFrontWithName();
  EXPECT_EQ(CurrentOptions().get<StringOption>(), "uh-oh"); // TODO - remove this, just pointing out what I expect
  EXPECT_EQ(name, "MakeRelativeTimer");
  p.set_value(Status()); // timers should succeed or the loop is aborted 

    
}

```

google/cloud/internal/async_retry_loop.h
```diff
@@ -189,8 +189,12 @@ class AsyncRetryLoopImpl
 
   future<T> Start() {
     auto weak = std::weak_ptr<AsyncRetryLoopImpl>(this->shared_from_this());
-    result_ = promise<T>([weak] {
-      if (auto self = weak.lock()) self->Cancel();
+    auto const& options = CurrentOptions();
```
coryan, line 192: Same comments as in the polling loop.  I think you want to capture these options so they can be set in the context of the callbacks, e.g., when entering (or before calling) `OnAttempt()`, `OnOperation`, or `OnBackoff`.

coryan (changes requested):

google/cloud/internal/async_polling_loop.cc
coryan: BTW.  I could be convinced that we should turn this back on.  @devjgm also has opinions on this

google/cloud/internal/async_retry_loop.h
devbww: See above.  `OnAttempt()`, `OnOperation()`, and `OnBackoff()` are already called with the loop-start-time options.

google/cloud/internal/async_polling_loop.cc
coryan: SGTM

google/cloud/internal/async_polling_loop.cc
coryan: Aahhh, I forgot about those.

google/cloud/internal/async_retry_loop.h
coryan: Thanks

coryan (approved):

PR #8257, opened 2022-02-05
fix: correct uses of `target_compatible_with`

coryan:
Perhaps misleadingly, the list of constraints provided to
`target_compatible_with` form a conjunction: all these constraints must
be satisfied for the rule to be compatible:

> A list of constraint_values that **must** be present

https://docs.bazel.build/versions/main/be/common-definitions.html#common.target_compatible_with

What wanted to do in all cases was to exclude one platform, and
fortunately Bazel provides `@platforms//:incompatible` for that purpose.

Part of the work for #7936

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8257)

### DIFF
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -26,6 +26,17 @@ def google_cloud_cpp_deps():
     override the version of the dependencies they want to use.
     """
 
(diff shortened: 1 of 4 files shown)


devjgm (approved):
Oops. Thanks for fixing.

PR #8258, opened 2022-02-05
fix(pubsublite): correct quickstart and enable in CI

coryan:
The quickstart had a number of problems, starting with using the wrong
endpoint.  Pub/Sub Lite happens to be zonal, it requires a zone as a
location argument, and it uses regional endpoints.

Part of the work for #7936

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8258)


google/cloud/pubsublite/endpoint.h
```diff
@@ -0,0 +1,35 @@
...
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace pubsublite {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+StatusOr<std::string> EndpointFromZone(std::string const& zone_id);
```
dbolduc, line 28: I think we also want one for regions. So like:

```cc
/// A region looks like: `us-east4`
StatusOr<std::string> EndpointFromRegion(std::string const& region_id);
/// A zone looks like: `us-east4-c`
StatusOr<std::string> EndpointFromZone(std::string const& zone_id);
```

(here's why I think that). Create Topic UI:
image

I made one of each, and they are displayed like this:
image

dbolduc: Or maybe `EndpointFromLocation`? that could get messy though.

coryan: Thanks.  I added the new helper, but left the quickstart as-is.

ci/etc/integration-tests-config.ps1
```diff
@@ -12,13 +12,21 @@
...
 $env:GOOGLE_CLOUD_REGION="us-central1"
+# Some quickstart programs require a zone.
+$env:GOOGLE_CLOUD_CPP_TEST_ZONE="us-central1-a"
+
+# Enable the self-test for the sample programs. Normally the example drivers
+# require the name of the example to run as a command-line argument, with this
+# environment variable the sample drivers run all the examples.
 $env:GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES="yes"
+
+# A number of options to improve logging during the CI builds. The are useful
```
dbolduc, line 28: s/The/They/

ci/etc/integration-tests-config.sh
```diff
@@ -20,13 +20,21 @@ if ((CI_ETC_INTEGRATION_TESTS_CONFIG_SH__++ != 0)); then
...
 export GOOGLE_CLOUD_REGION="us-central1"
+# Some quickstart programs require a zone.
+export GOOGLE_CLOUD_CPP_TEST_ZONE="us-central1-a"
+
+# Enable the self-test for the sample programs. Normally the example drivers
+# require the name of the example to run as a command-line argument, with this
+# environment variable the sample drivers run all the examples.
 export GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES="yes"
+
+# A number of options to improve logging during the CI builds. The are useful
```
dbolduc, line 36: s/The/They/

dbolduc (approved):

ci/etc/integration-tests-config.ps1
coryan: Fixed.

ci/etc/integration-tests-config.sh
coryan: Fixed

PR #8267, opened 2022-02-05
chore: upgrade and fix typos-cli

devjgm:
This upgrades us to the newest version of typos-cli. The previous
version we were using (1.3.1) was also broken, but we didn't notice
because I think all of our `gcb-checkers` docker images had an old
version that worked. Aside: I think somehow the cached docker images
don't get invalidated properly.

Anyway, the string "ba" showed up as a typo (in the old and new
version). I changed it to "bx", which I think makes the word weird
enough that `typos` doesn't flag it as a typo. I also upgrade to a newer
version of typos to make sure all the docker images get rebuilt.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8267)

### DIFF
--- a/ci/cloudbuild/builds/checkers.sh
+++ b/ci/cloudbuild/builds/checkers.sh
@@ -121,7 +121,7 @@ time {
   # Removes trailing whitespace on lines
   expressions=("-e" "'s/[[:blank:]]\+$//'")
   # Removes trailing blank lines (see http://sed.sourceforge.net/sed1line.txt)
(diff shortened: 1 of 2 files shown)


devbww (approved):

coryan:
this it is just too ironic to ignore: s/becuase/because/ in the commit message

devjgm:
> this it is just too ironic to ignore: s/becuase/because/ in the commit message

:-) Fixed

PR #8268, opened 2022-02-06
fix(rest): correct detection of all headers received

scotthart:
Instead of trying detect by ourselves when the status and all headers have been received, rely on libcurl to signal that via the first call to WriteCallback. libcurl will invoke HeaderCallback up until that point.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8268)


coryan (approved):

coryan:
Looking at the libcurl docs, it seems that a non-zero status code also indicated the headers are fully received.

PR #8276, opened 2022-02-06
impl(oauth2): add support for OAuth2 compute engine credentials

scotthart:
In order to support and test the compute engine credentials:
- RestClient factory functions moved inside classes for dependency injection
- RestClient mock added

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8276)


google/cloud/testing_util/mock_rest_client.h
```diff
@@ -0,0 +1,105 @@
...
+      std::vector<absl::Span<char const>> const& payload) override {
+    return impl_->Post(request, payload);
+  }
+  StatusOr<std::unique_ptr<rest_internal::RestResponse>> Put(
+      rest_internal::RestRequest const& request,
+      std::vector<absl::Span<char const>> const& payload) override {
+    return impl_->Put(request, payload);
+  }
+
+  struct Impl : public RestClient {
```
coryan, line 67: What is the motivation for this extra indirection in the `MockRestClient`?

google/cloud/internal/rest_client_integration_test.cc
```diff
@@ -124,7 +124,7 @@ class RestClientIntegrationTest : public ::testing::Test {
 
 TEST_F(RestClientIntegrationTest, Get) {
   options_.set<UnifiedCredentialsOption>(MakeInsecureCredentials());
-  auto client = GetDefaultRestClient(url_, {});
+  auto client = DefaultRestClient::GetRestClient(url_, {});
```
coryan, line 127: These changes seem unrelated?

google/cloud/internal/oauth2_refreshing_credentials_wrapper.h
```diff
@@ -0,0 +1,86 @@
...
+
+namespace google {
+namespace cloud {
+namespace oauth2_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * Wrapper for refreshable parts of a Credentials object.
+ */
+class RefreshingCredentialsWrapper {
```
coryan, line 33: Looks like this class could be implemented and reviewed in an initial PR?  Maybe we can do that to reduce the size?

coryan (changes requested):
This PR is too large, please break down. It also seems to contain code that we decided to implement differently in previous PRs?

I left a number of comments because the code could be simplified as your `RestClient` is much easier to mock that `HttpRequest`.

PR #8283, opened 2022-02-07
feat(generator): support parameters named "options"

coryan:
The googleapis RPCs can define "method signatures" which decompose a
proto `request` object into its key parameters. The generator maps these
to overloads, and adds the `Options options` parameter. This does not
work when one of the arguments defined in the method signature is
already called `options`.  This changes the extract `Options options`
parameter to `Options opts`, which, as of this writing, does not have
any conflicts in the googleapis protos.

Part of the work for #8175 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8283)

### DIFF
--- a/generator/internal/client_generator.cc
+++ b/generator/internal/client_generator.cc
@@ -197,8 +197,8 @@ Status ClientGenerator::GenerateHeader() {
    * specified. See the operation documentation for the appropriate value for
    * this field.
    * @param updater  Required. Functor to map the current policy to a new one.
-   * @param opts  Optional. Override the default class-level options, such as
-   *     retry and backoff policies.
+   * @param opts  Optional. Override the class-level options, such as retry and
+   *    backoff policies.
 )"""},
             {"   * @return " + response_type + "\n"},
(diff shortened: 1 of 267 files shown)


generator/internal/client_generator.cc
```diff
@@ -197,17 +197,16 @@ Status ClientGenerator::GenerateHeader() {
    * specified. See the operation documentation for the appropriate value for
    * this field.
    * @param updater  Required. Functor to map the current policy to a new one.
-   * @param options  Optional. Options to control the loop. Expected options
-   * are:
-   *       - `$service_name$BackoffPolicyOption`
+   * @param opts  Optional. Override the default class-level options, such as
+   *     retry and backoff policies.
```
devbww, line 201: Nit: This is subtly different from the comment generated for the normal `Options` parameters by including "default".  I would suggest just removing it here.

coryan: Fixed.

devbww (approved):

PR #8289, opened 2022-02-07
chore: fix west->east const

devjgm:
These were fixed using a new clang-format with
`--qualifier-alignment=right`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8289)


coryan (approved):
Nice.  Fewer than I would have expected.

PR #8291, opened 2022-02-08
impl(rest): add POST with form_data method

scotthart:
This adds an overload to RestClient::Post that accepts a collection of key/value pairs representing form data. The values of each pair are x-www-form-urlencoded and the all pairs are assembled into a payload such `key1=value1&key2=value2&keyN=valueN`. This addition was the result of preserving the libcurl encapsulation in `CurlImpl` and `curl_easy_escape` requiring an easy handle as an argument.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8291)


coryan (approved):
Feels like there is something not quite orthogonal if we need a new method for a new kind of `Post()` request.

I disabled auto-merge so you can give it a thought.

Issue #8284, opened 2022-02-07
Support `x-goog-user-project` in hand-crafted libraries


coryan:
Or at least in the parts that are not obsoleted:

- [x] Bigtable
- [x] Pub/Sub
- [x] Spanner
- [x] Storage: already supported using `userProject`, a per-call option, but maybe we want to support configuring this via `google::cloud::Options`.

coryan:
I am going to assert that the current support in Storage is good enough.

PR #8294, opened 2022-02-08
refactor: use helpers to populate default options 

coryan:
All the libraries need to initialize some common values in the default
`Options`. For example, `EndpointOption`, or `TracingOptions`. With this
change all the generated libraries use some common helpers to do this
work.

Things are a bit complicated because the common options are split across
the `google_cloud_cpp_common` and `google_cloud_cpp_grpc_utils`. I wrote
two separate functions to make that doable.

Part of the work for #8284 (or at least motivated by it).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8294)


google/cloud/internal/populate_common_options.cc
```diff
@@ -0,0 +1,67 @@
...
+        opts.set<EndpointOption>(default_endpoint);
+      }
+    }
+  }
+  if (!emulator_env_var.empty()) {
+    auto e = GetEnv(emulator_env_var.c_str());
+    if (e && !e->empty()) {
+      opts.set<EndpointOption>(*std::move(e));
+    }
+  }
```
dbolduc, line 52: I think this is simpler, now that we are not setting the `GrpcCredentialOption` in here.
```cc
  if (!emulator_env_var.empty()) {
    auto e = GetEnv(emulator_env_var.c_str());
    if (e && !e->empty()) {
      opts.set<EndpointOption>(*std::move(e));
    }
  } else {
    if (!opts.has<EndpointOption>()) {
      opts.set<EndpointOption>(std::move(default_endpoint));
    }
  }
```

google/cloud/internal/populate_common_options.cc
```diff
@@ -0,0 +1,67 @@
...
+  auto e = GetEnv("GOOGLE_CLOUD_CPP_USER_PROJECT");
+  if (e.has_value() && !e->empty()) opts.set<UserProjectOption>(*std::move(e));
```
dbolduc, line 54: I think @devbww did not like the use of both `e` and `e.has_value()`  so close together.

google/cloud/internal/populate_common_options_test.cc
```diff
@@ -0,0 +1,121 @@
...
+  } test_cases[] = {
+      {"empty-0", empty, absl::nullopt, absl::nullopt, "default"},
+      {"empty-1", empty, absl::nullopt, "emulator", "emulator"},
+      {"empty-2", empty, "default", absl::nullopt, "default"},
+      {"empty-3", empty, "default", "emulator", "emulator"},
+      {"with-ep-0", with_ep, absl::nullopt, absl::nullopt, "with-ep"},
+      {"with-ep-1", with_ep, absl::nullopt, "emulator", "emulator"},
+      {"with-ep-2", with_ep, "default", absl::nullopt, "with-ep"},
+      {"with-ep-3", with_ep, "default", "emulator", "emulator"},
+  };
```
dbolduc, line 66: maybe add one more case where the emulator endpoint is `""` instead of `absl::nullopt` ?

dbolduc (changes requested):

google/cloud/internal/populate_common_options_test.cc
coryan: Done.

google/cloud/internal/populate_common_options.cc
coryan: Fixed.

google/cloud/internal/populate_common_options.cc
coryan: The second `if()` would have no effect, we already assigned an `EndpointOption` in the `if()` starting at line 35.

Changed the code again to (maybe) make it easier to follow.

dbolduc: I am sorry. I gave you bad code. (notably it never referenced `endpoint_env_var`)

Here is code that actually passes your tests. I claim that it is easier to follow:

```cc
  if (!emulator_env_var.empty()) {
    auto e = internal::GetEnv(emulator_env_var.c_str());
    if (e && !e->empty()) {
      opts.set<EndpointOption>(*std::move(e));
    }
  }
  if (!opts.has<EndpointOption>()) {
    auto e = internal::GetEnv(endpoint_env_var.c_str());
    if (e && !e->empty()) {
      opts.set<EndpointOption>(*std::move(e));
    } else {
      opts.set<EndpointOption>(std::move(default_endpoint));
    }
  }
```

dbolduc: The tests do not test that setting `EndpointOption == ""` should be overwritten btw. I don't think this is a problem, the same way I didn't think we should have to test for an empty `std::function`. But just an FYI

dbolduc (approved):

google/cloud/internal/populate_common_options.cc
```diff
@@ -0,0 +1,67 @@
...
+#include "google/cloud/internal/populate_common_options.h"
+#include "google/cloud/common_options.h"
+#include "google/cloud/internal/getenv.h"
+#include "google/cloud/internal/user_agent_prefix.h"
+#include "absl/strings/str_split.h"
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
```
devbww, line 24: Nitty nit, if it is not too late ... insert separator line.

google/cloud/internal/populate_common_options.cc
```diff
@@ -0,0 +1,67 @@
...
+std::set<std::string> DefaultTracingComponents() {
```
devbww, line 25: Nit: Move this down so it matches the header declaration order (and the similar ordering with `DefaultTracingOptions()` in the "grpc_options" version).

coryan: I will do those in a separate PR.  I want to merge this so I can unblock other changes in my queue.

PR #8295, opened 2022-02-08
cleanup: cut more output from interactive builds

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8295)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -25,6 +25,7 @@
 #     -l|--local           Run the build in the local environment
 #     -d|--docker          Run the build in a local docker (default)
 #     -s|--docker-shell    Run a shell in the build's docker container
+#     --verbose            Print additional information during the build
 #     -h|--help            Print this help message
 #
 #   Note: flags may be specified in any order and with or without an equals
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/build.sh
```diff
@@ -190,31 +196,39 @@ if [[ "${LOCAL_FLAG}" = "true" ]]; then
...
+    # link may be publicly accessible. In manual builds this information is
+    # never useful (both are "none").
+    io::log_h1 "Log Links"
+    printf "GCB: %s\n" "${CONSOLE_LOG_URL:-none}"
+    printf "Raw: %s\n" "${RAW_LOG_URL:-none}"
+  fi
+
+  if [[ "${TRIGGER_TYPE}" != "manual" || "${VERBOSE_FLAG}" == "true" ]]; then
+    # Prints information about the machine and compiler. In manual builds this
+    # information is almost never useful.
```
devjgm, line 218: I think this info can be useful on manual builds, because they all run in Docker containers so the OS and compilers versions and stuff are not always obvious. My opinion would be to continue printing this, unless others think this is too noisy during manual builds.

devjgm (approved):
There's been a number of PRs reducing build output. Are these just nice cleanups or is there a problem or bigger objective driving them that I missed?

ci/cloudbuild/build.sh
coryan: Ooops, you triggered the merge before we could discuss, my trap has sprung, mua ha ha.

Obviously **I** think it is too noisy, most of the time I am running a build locally this information gooes by unread.  If I need it I can run again with `--verbose`.  But we can change it back, adding this to the team meeting agenda.

PR #8306, opened 2022-02-08
refactor: enabled performance-move-const-arg clang-tidy

devbww:
This enables a handful of "remove std::move()" suggestions:
- std::move of the const variable/expression has no effect
- std::move of the trivially-copyable type has no effect
- passing result of std::move as a const& argument; no move will happen

In a few cases we remove the `const` from the variable so that the
move does have an effect (which is a performance win).  But otherwise,
eliminating all the no-op moves makes the code easier to understand.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8306)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -18,13 +18,6 @@
 #      avoiding C arrays often makes the code less readable, and std::array is
 #      not a drop-in replacement because it doesn't deduce the size.
 #
-#  -performance-move-const-arg: This warning requires the developer to
-#      know/care more about the implementation details of types/functions than
-#      should be necessary. For example, `A a; F(std::move(a));` will trigger a
-#      warning IFF `A` is a trivial type (and therefore the move is
-#      meaningless). It would also warn if `F` accepts by `const&`, which is
-#      another detail that the caller need not care about.
-#
 #  -readability-redundant-declaration: A friend declaration inside a class
 #      counts as a declaration, so if we also declare that friend outside the
 #      class in order to document it as part of the public API, that will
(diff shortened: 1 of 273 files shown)


devjgm:
I vote against merging this PR. From a spotcheck of the changes, the code looks better "before", with the std::moves. AFAICT, all the "remove std::move()" suggestions are noise and none of them improve the code or performance in anyway. For example, the three warnings you mention the PR description are all all actively bad recommendations from this check, IMO. This warning is widely considered bad advice. We disable this warning internally as well for the same reasons.

devjgm:
I should have added that I'm sorry to vote against this. You clearly put a lot of work into this PR cleaning things up. Thanks for that.

Also, the warning about:
```cc
const string s;
return std::move(s);  // Warning: std::move of the const variable has no effect
```
is potentially useful, because it warns about things that the _caller_ controls, like the use of `std::move` and the `const`ness of the argument. So warning there could be helpful because the caller could easily do something about it. However, the other warnings are noise because they warn the caller about things that they potentially/likely don't control and could potentially change in the future, like how the function accepts the argument (by value or reference) and whether the argument happens to be trivial or not. Keeping the `std::move` in these latter cases does no harm and future-proofs the code.

I wish we could configure the warning to only warn in the one case, but sadly there's no option for that. It should be easy to add an option upstream to control that if we wanted to.

devjgm:
FTR, https://reviews.llvm.org/D119370 added an option to let us disable `performance-move-const-arg.CheckMoveToConstRef`. There are now two options for this check (https://clang.llvm.org/extra/clang-tidy/checks/performance-move-const-arg.html) that should let us turn off all the noise and keep the useful warnings... though we'd need to upgrade our version of clang-tidy, which maybe we can do.

Just following up for posterity.

PR #8308, opened 2022-02-08
ci: speedup macos build shutdown

coryan:
We were not cleaning up the artifacts directory on errors. Kokoro copies
this directory very slowly, and this was making our failures slower than
they needed to be.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8308)


devbww (approved):

coryan:
the builds still fail but we saved 11 minutes waiting for the failure.

PR #8326, opened 2022-02-09
doc(pubsub): add region tag for subscription with filter

coryan:
Use the existing example, it does what we want it just needed the right
region tags.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8326)


dbolduc (approved):
is "_subscrption_wit_" in the title intentional?

coryan:
> is "_subscrption_wit_" in the title intentional?

Thanks, fixed.

PR #8328, opened 2022-02-10
impl(bigtable): extract or create CQ from options

dbolduc:
The use in `DefaultAdminClient` will be:

```cc
class DefaultAdminClient : public google::cloud::bigtable::AdminClient {
 public:
  // Or maybe I pass in AdminClientParams intstead of Options
  DefaultAdminClient(std::string project, Options options)
     : project_(std::move(project)) {
    auto params = AdminClientParams(std::move(options));
    cq_ = std::move(params.cq);
    background_threads_ = std::move(params.background_threads);
    connection_ = bigtable_admin::MakeBigtableTableAdminConnection(
        std::move(params.options));
  }

  std::string project_;
  CompletionQueue cq_;
  std::unique_ptr<BackgroundThreads> background_threads_;
  std::shared_ptr<bigtable_admin::BigtableTableAdminConnection> connection_;
};

```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8328)


google/cloud/bigtable/admin_client.h
```diff
@@ -308,6 +308,37 @@ std::shared_ptr<AdminClient> CreateDefaultAdminClient(std::string project,
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace bigtable
+namespace bigtable_internal {
```
coryan, line 311: Please move this class to a header in `google/cloud/bigtable/internal`.

google/cloud/bigtable/admin_client_test.cc
```diff
@@ -23,6 +24,15 @@ namespace bigtable {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace {
 
+using ::testing::IsNull;
+using ::testing::NotNull;
+
+void CompareCQ(CompletionQueue const& a, CompletionQueue const& b) {
+  auto impl_a = google::cloud::internal::GetCompletionQueueImpl(a);
+  auto impl_b = google::cloud::internal::GetCompletionQueueImpl(b);
+  EXPECT_EQ(impl_a, impl_b);
```
coryan, line 33: Note that all failures will point to this line of code, and it will be more difficult to troubleshoot problems.  There is a clever way to use matchers and move the `EXPECT_*()` to your test code.

coryan (approved):
A nit about code location, please fix and then merge.

google/cloud/bigtable/admin_client_test.cc
coryan: Well, ADL is complicated, but I would think yes.  You are also risking an ODR violation if we ever add `operator==` in the header file for that class 🤷

PR #8329, opened 2022-02-10
feat(common): Add converting constructors for future

dpcollins-google:
When the value type of a future is convertible to the value type of another, this allows the futures to interconvert.

This is useful with, for example, make_ready_future, where one could return `make_ready_future(absl::nullopt)` for a return type of `future<absl::optional<Result>>`

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8329)

### DIFF
--- a/google/cloud/future_generic.h
+++ b/google/cloud/future_generic.h
@@ -24,6 +24,7 @@
 #include "google/cloud/internal/future_fwd.h"
 #include "google/cloud/internal/future_impl.h"
 #include "google/cloud/internal/future_then_meta.h"
+#include "google/cloud/internal/utility.h"
 #include "google/cloud/version.h"
 
 namespace google {
@@ -64,8 +65,9 @@ class future final : private internal::future_base<T> {
    * Creates a future from a future whose result type is convertible to this
    * future's result type.
    */
-  template <class U,
-            typename Enable = std::enable_if<std::is_constructible<T, U>{}>>
+  // NOLINTNEXTLINE(google-explicit-constructor)
+  template <class U, typename Enable = internal::enable_if_t<
+                         std::is_constructible<T, U>::value>>
   future(future<U>&& rhs)
       : future<T>(rhs.then([](future<U> other) { return T(other.get()); })) {}
 
--- a/google/cloud/future_generic.h
+++ b/google/cloud/future_generic.h
@@ -65,10 +65,9 @@ class future final : private internal::future_base<T> {
    * Creates a future from a future whose result type is convertible to this
    * future's result type.
    */
-  // NOLINTNEXTLINE(google-explicit-constructor)
   template <class U, typename Enable = internal::enable_if_t<
                          std::is_constructible<T, U>::value>>
-  future(future<U>&& rhs)
+  explicit future(future<U>&& rhs)
       : future<T>(rhs.then([](future<U> other) { return T(other.get()); })) {}
 
   /**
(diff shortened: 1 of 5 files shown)


google/cloud/future_generic.h
```diff
@@ -60,6 +60,15 @@ class future final : private internal::future_base<T> {
   // NOLINTNEXTLINE(google-explicit-constructor)
   future(future<future<T>>&& rhs) noexcept(false);
 
+  /**
+   * Creates a future from a future whose result type is convertible to this
+   * future's result type.
+   */
+  template <class U,
+            typename Enable = std::enable_if<std::is_constructible<T, U>{}>>
```
coryan, line 68: This is allowing code to compile when it shouldn't: https://godbolt.org/z/zPzefx6GG
Note the values of `ShouldBe0` and `ShouldBe1`.  You want  `is_constructible<>::value` and you want `std::enable_if<>::type`.

google/cloud/future_generic.h
```diff
@@ -60,6 +60,15 @@ class future final : private internal::future_base<T> {
...
+  future(future<U>&& rhs)
```
coryan, line 69: This needs an `explicit`

google/cloud/future_void.h
```diff
@@ -60,6 +60,13 @@ class future<void> final : private internal::future_base<void> {
   // NOLINTNEXTLINE(google-explicit-constructor)
   future(future<future<void>>&& rhs) noexcept(false);
 
+  /**
+   * Creates a future from a future whose result type is convertible to this
+   * future's result type.
+   */
+  template <class T>
+  future(future<T>&& rhs) : future<void>(rhs.then([](future<T>) {})) {}
```
coryan, line 68: This needs an `explicit`

coryan (changes requested):
The robots have things to say, I agree with my robot overlords.

google/cloud/future_generic.h
dpcollins-google: So, the purpose was for this to be an implicit conversion- such that it could operate on function return values. But I could make it explicit if you'd prefer

google/cloud/future_generic.h
dpcollins-google: Good catch, will fix. Do you have the ability to write nc tests?

coryan: Not at this time.  And I am not sure how we would write them.  But you should be able to write a `static_assert()` that verifies SFINAE works as expected?

google/cloud/future_generic.h
coryan: My gut says we should not allow implicit conversions, I might be wrong, but we can always add them later, while removing them is a breaking change.  Please make these explicit

dpcollins-google: ack

dpcollins-google: Done.

google/cloud/future_void.h
dpcollins-google: Done.

coryan:
Can you please rebase?  We have been making a lot of changes to our builds and you are missing some support scripts / changes.

coryan:
Ping.

google/cloud/future_generic.h
dpcollins-google: Done.

dpcollins-google:
> Can you please rebase? We have been making a lot of changes to our builds and you are missing some support scripts / changes.

Done.

Issue #5116, opened 2020-09-22
PKCS #12 keys are not supported in gRPC


dopiera:
gRPC doesn't support PKCS #12 service account keys. This means that creating `storage::GrpcClient` from a `.p12` file doesn't work.

We should probably fix this by implementing the support in gRPC (boring SSL seems to have support for it, so it should be pretty easy).

Alternatively, we could try to transform the `.p12` keys in the `storage` library, but I'm afraid there is no way to do it (the `.p12` doesn't contain the service account email).

devjgm:
Closing because we're not going to support PKCS #12

Issue #12, opened 2017-11-26
Implement (sync) Table::Apply(SingleRowMutation)


coryan:
At the end of this task `bigtable::Table` will provide an `Apply()` member function to apply mutations to a single row.  The member functions should:

* Use the policies in effect to retry any recoverable RPC failures.
* Use the policies in effect to determine if a mutation is idempotent.
* Report an error when the operation fails after the policies "give up".

coryan:
For what is worth, this is blocked behind #13, #14, #15 and #16.

coryan:
#13, #14, and #16 are done.  I just closed #15 as invalid.  I am calling this done.

mbrukman:
@coryan — replaced the bulletted list in the issue description with checkboxes and added issue references; PTAL and verify if my mapping is correct.

coryan:
Nice, thanks.  LGTM.

PR #8343, opened 2022-02-11
impl(oauth2): port service account credential prerequisites

scotthart:
This PR brings over several oauth2/ssl utility files from google/cloud/storage. These are all necessary for implementing OAuth2.0 service account credentials.

Other than changes to namespaces and include paths, the contents of these files have been copied as-is from google/cloud/storage. Ultimately these file locations will be their permanent homes after GCS has been refactored to use the new REST library.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343)


coryan (changes requested):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343)* status: all files reviewed, 5 unresolved discussions (waiting on @scotthart)

___
*[google/cloud/internal/oauth2_credential_constants.h, line 33 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343#-MvclYKfDMQqeW3KUJ4q:-MvclYKg61tYbsfs4EVH:bbgn65d) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/caf13d665d3c6bd12f076f3600a17b88da3a001f/google/cloud/internal/oauth2_credential_constants.h#L33)):*
> ```objc
>  */
> // NOLINTNEXTLINE(readability-identifier-naming)
> enum class JwtSigningAlgorithms { RS256 };
> ```

Let's remove this enum.  It serves no purpose, if we ever need a new signing algorithm we can create a function or enum at the time.
___
*[google/cloud/internal/openssl_util.h, line 65 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343#-Mvcn2-42xHpBViz9GRd:-Mvcn2-42xHpBViz9GRe:b-t8s4yl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/caf13d665d3c6bd12f076f3600a17b88da3a001f/google/cloud/internal/openssl_util.h#L65)):*
> ```objc
> template <typename Collection>
> inline std::string UrlsafeBase64Encode(Collection const& bytes) {
>   std::string b64str = Base64Encode(bytes);
> ```

Optional: I think it might be cleaner to just use `google::cloud::internal::Base64Encoder` directly here, and remove the `Base64Encode()` functions?
___
*[google/cloud/internal/openssl_util.cc, line 75 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343#-MvcmJiM9xRJ5-zR1BVm:-MvcmJiM9xRJ5-zR1BVn:b209tuv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/caf13d665d3c6bd12f076f3600a17b88da3a001f/google/cloud/internal/openssl_util.cc#L75)):*
> ```cpp
> 
>   EVP_MD const* digest_type = nullptr;
>   switch (alg) {
> ```

If we remove the `alg` argument we can cut all this code, maybe it requires renaming the function `SignUsingSha256`, the type of the arguments are implicit and don't need to be in the name again.  
___
*[google/cloud/internal/openssl_util.cc, line 173 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343#-MvcnbvP3nvY4idIv8IP:-MvcnbvP3nvY4idIv8IQ:bcm2c8f) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/caf13d665d3c6bd12f076f3600a17b88da3a001f/google/cloud/internal/openssl_util.cc#L173)):*
> ```cpp
> }
> 
> std::vector<std::uint8_t> MD5Hash(std::string const& payload) {
> ```

I suspect this is very GCS specific, I doubt it will see usage in other services. Please remove.
___
*[google/cloud/internal/write_base64.h, line 31 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343#-MvcoECr4oknbjSGm511:-MvcoECr4oknbjSGm512:b-wrlzs1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/caf13d665d3c6bd12f076f3600a17b88da3a001f/google/cloud/internal/write_base64.h#L31)):*
> ```objc
>  * If it fails, it will throw an exception on badbit.
>  */
> void WriteBase64AsBinary(std::string const& filename, char const* data);
> ```

AFAICT this is only used in the tests for `.p12` key files.  We have no plans to support those with GUAC (they don't work with gRPC see #5116).  I think we should remove this code and all the support for `PKCS #12`.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343)* status: 12 of 16 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/testing_util/credentials_constants.h, line 1 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343#-MveuxBT7EFhA8IpwII5:-Mvf0MbyD6k0k9Adq0Qx:bf7teh9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/701a380170aa86a637924167a213492467432206/google/cloud/testing_util/credentials_constants.h#L1)):*

Is this file used anywhere?  It is only for testing and does not appear to be used in `openssl_utils_test.cc`?
</blockquote></details>

it's used in make_jwt_assertion_test and another test file "Not Yet Appearing in this Film", but we don't need a separate file for this constant used in 2 places.
___
*[google/cloud/testing_util/credentials_constants.h, line 41 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343#-MveuX6w9EebySkCzADe:-Mvf0K5f7uFOo_R2gnzO:bsv4n8l) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/701a380170aa86a637924167a213492467432206/google/cloud/testing_util/credentials_constants.h#L41)):*

Are any of these two constants used in the new code?  I would expect they are not?
</blockquote></details>

No, they are not.

coryan (approved):
:lgtm:

> Reviewed 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343#-:-Mvf0w_W5r8Lsl9HtfoO:bnfp4nl)* status: 15 of 16 files reviewed, all discussions resolved (waiting on @coryan)

coryan (approved):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8343)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

PR #8355, opened 2022-02-12
fix(spanner): add Options save/restore to PartialResultSetSource

devbww:
Save the options in force during `PartialResultSetSource` construction,
and then restore them over `PartialResultSetReader` calls.  This means
that `NextRow()`, `Finish()` and `TryCancel()` will have the right
options installed during any RPCs they execute.

Tighten up the test by installing a non-matching `OptionsSpan` before
any `NextRow()` calls are made, and by expecting no `TryCancel()`s in
cases that complete their enumeration.

This is the Spanner equivalent of #8256.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8355)

### DIFF
--- a/google/cloud/spanner/internal/partial_result_set_source.cc
+++ b/google/cloud/spanner/internal/partial_result_set_source.cc
@@ -84,6 +84,7 @@ PartialResultSetSource::~PartialResultSetSource() {
     // If there is actual data in the streaming RPC Finish() can deadlock, so
     // before trying to read the final status we need to cancel the streaming
     // RPC.
+    internal::OptionsSpan span(options_);
     reader_->TryCancel();
     // The user didn't iterate over all the data; finish the stream on their
     // behalf, but we have no way to communicate error status.
@@ -95,6 +96,7 @@ PartialResultSetSource::~PartialResultSetSource() {
 }
 
 Status PartialResultSetSource::ReadFromStream() {
+  internal::OptionsSpan span(options_);
   auto result_set = reader_->Read();
   if (!result_set) {
     // Read() returns false for end of stream, whether we read all the data or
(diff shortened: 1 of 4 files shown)


coryan (approved):
This got me thinking, should we make analogous changes to the resumable streaming RPC:

https://github.com/googleapis/google-cloud-cpp/blob/18082df93f4de5e52b97464eadc21f8e40a539a9/google/cloud/internal/resumable_streaming_read_rpc.h#L93-L99

devbww:
> This got me thinking, should we make analogous changes to the resumable streaming RPC

I'll look into it.

coryan (approved):

devbww:
> > This got me thinking, should we make analogous changes to the resumable streaming RPC
> 
> I'll look into it.

I had thought that the `StreamRange` changes would handle this, but it looks like there is a hole when the `StreamRange` destroys its `reader_`, as that might cause a cancellation RPC without the right options if the reader is not yet finished.  I'm investigating fixing that, and making the generated and Spanner cases look as similar as possible.

devbww:
> > > This got me thinking, should we make analogous changes to the resumable streaming RPC
> > 
> > 
> > I'll look into it.
> 
> I had thought that the `StreamRange` changes would handle this, but it looks like there is a hole when the `StreamRange` destroys its `reader_`, as that might cause a cancellation RPC without the right options if the reader is not yet finished. I'm investigating fixing that, and making the generated and Spanner cases look as similar as possible.

I don't think there was really a hole, as destroying the unfinished streaming reader only calls directly into gRPC to try to cancel the stream.  That is, there is no cancellation RPC or other code that might look at our `Options`.  But, we should probably always restore the operation-start options whenever the user can call back into the library, no matter what code we think that executes.  So, I created #8403.

coryan:
> 
> I don't think there was really a hole, as destroying the unfinished streaming reader only calls directly into gRPC to try to cancel the stream.

I was thinking about `google::cloud::internal::ResumableStreamingReadRpc<>`, I just created #8404, you can tell me there if the changes are not needed.

PR #8370, opened 2022-02-14
chore(deps): update storage testbench

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8370)


coryan:
> Is this something we could teach renovate-bot to do?

Maybe.  Usually I create the releases, and because I am about to use a new version in our repo, so I am not worried about missing an update, but would be nice to save me the typing.

PR #8376, opened 2022-02-15
impl(oauth2): add factory functions to create oauth2 credentials

scotthart:
This PR adds a collection of factory functions to create OAuth2 credentials from various sources. Per previous conversations, P12 file support has been removed.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376)


google/cloud/internal/oauth2_google_credentials.cc
```diff
@@ -0,0 +1,291 @@
...
+    Options const& options) {
+  std::ifstream ifs(path);
+  if (!ifs.is_open()) {
+    // We use kUnknown here because we don't know if the file does not exist, or
+    // if we were unable to open it for some other reason.
+    return Status(StatusCode::kUnknown, "Cannot open credentials file " + path);
+  }
+  std::string contents(std::istreambuf_iterator<char>{ifs}, {});
+  auto cred_json = nlohmann::json::parse(contents, nullptr, false);
+  if (cred_json.is_discarded()) {
```
dbolduc, line 63: Not a full review, just a comment. I think we mostly stopped using `.is_discarded()` in #8033.

There are still some uses of `.is_discarded()` in `storage`, though. I will defer to @coryan for when to use one vs. the other.
https://github.com/googleapis/google-cloud-cpp/search?q=is_discarded%28%29

coryan (changes requested):
> Reviewed 4 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376)* status: 4 of 8 files reviewed, 3 unresolved discussions (waiting on @coryan, @dbolduc, and @scotthart)

___
*[google/cloud/internal/oauth2_google_credentials.h, line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376#-MvxJI3I7hNywWG8FW6W:-MvxIBfUDvYiRQzfeFtF:b-9e75f4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e83ff5126041ad26e83882d14cb292011e039884/google/cloud/internal/oauth2_google_credentials.h#L44)):*
> ```objc
>  * about Application Default %Credentials.
>  */
> StatusOr<std::shared_ptr<Credentials>> GoogleDefaultCredentials(
> ```

I get why we need `GoogleDefaultCredentials()`, I am not sure why we need the other functions.  If I recall correctly, the plan is to *only* expose GUAC / unified credentials to application developers.  That would be implemented via something like this:

https://github.com/googleapis/google-cloud-cpp/blob/c11a6cce5344fb55c048cfaff7a8b702b53503e5/google/cloud/storage/internal/unified_rest_credentials.cc#L35-L38

Only that function would need to call the functions in this file.  In some cases we would not bother writing them (e.g. `CreateAnonymousCredentials()`), in others we would probably simplify them, in others we would write the single variant we need, e.g., `CreateAuthorizedUserCredentials*()`, and in most cases we would just return `std::shared_ptr<Credentials>` because the credentials mapper just converts `StatusOr<std::shared_ptr<Credentials>>` to `ErrorCredentials` on error, so we could do the same on its helper functions.
___
*[google/cloud/internal/oauth2_google_credentials.cc, line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376#-Mvumnuq4Ezi185qPkd6-r1-63:-MvxJ-jQAP1z5-_8uPY4:b-8fx2u7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e83ff5126041ad26e83882d14cb292011e039884/google/cloud/internal/oauth2_google_credentials.cc#L63)):*

Not a full review, just a comment. I think we mostly stopped using `.is_discarded()` in #8033.

There are still some uses of `.is_discarded()` in `storage`, though. I will defer to @coryan for when to use one vs. the other.
https://github.com/googleapis/google-cloud-cpp/search?q=is_discarded%28%29
</blockquote></details>

:+1: 
___
*[google/cloud/internal/oauth2_google_credentials.cc, line 160 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376#-MvxKXWvCx_DlMcZbQVt:-MvxKXWvCx_DlMcZbQVu:bxlgyym) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e83ff5126041ad26e83882d14cb292011e039884/google/cloud/internal/oauth2_google_credentials.cc#L160)):*
> ```cpp
>   auto gce_creds = std::make_shared<ComputeEngineCredentials>();
>   auto override_val =
>       google::cloud::internal::GetEnv(internal::GceCheckOverrideEnvVar());
> ```

I think if we were writing this code today we would use something like `oauth2_internal::ForceGceOption` and then write:


```cc
  // 3) Check for implicit environment-based credentials (GCE, GAE ...)
  if (options.get<oauth2_internal::ForceGceOption>() || gce_creds->AuthorizationHeader().ok()) {
      return gce_creds;
  }
```

That way the tests can just use the `options` to force GCE credentials.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376)* status: 4 of 8 files reviewed, 3 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[google/cloud/internal/oauth2_google_credentials.h, line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376#-MvxJI3I7hNywWG8FW6W:-Mvynn64-pYLA31oHfQT:b-y22ssv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e83ff5126041ad26e83882d14cb292011e039884/google/cloud/internal/oauth2_google_credentials.h#L44)):*

I get why we need `GoogleDefaultCredentials()`, I am not sure why we need the other functions.  If I recall correctly, the plan is to *only* expose GUAC / unified credentials to application developers.  That would be implemented via something like this:

https://github.com/googleapis/google-cloud-cpp/blob/c11a6cce5344fb55c048cfaff7a8b702b53503e5/google/cloud/storage/internal/unified_rest_credentials.cc#L35-L38

Only that function would need to call the functions in this file.  In some cases we would not bother writing them (e.g. `CreateAnonymousCredentials()`), in others we would probably simplify them, in others we would write the single variant we need, e.g., `CreateAuthorizedUserCredentials*()`, and in most cases we would just return `std::shared_ptr<Credentials>` because the credentials mapper just converts `StatusOr<std::shared_ptr<Credentials>>` to `ErrorCredentials` on error, so we could do the same on its helper functions.

</blockquote></details>

Removed the factory functions not necessary to creating `GoogleDefaultCredentials`. Relocated some to the future unified_rest_credentials PR.
___
*[google/cloud/internal/oauth2_google_credentials.cc, line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376#-Mvumnuq4Ezi185qPkd6-r1-63:-Mvyo1-tCUuPndO7wadH:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e83ff5126041ad26e83882d14cb292011e039884/google/cloud/internal/oauth2_google_credentials.cc#L63)):*

:+1: 
</blockquote></details>

Done.
___
*[google/cloud/internal/oauth2_google_credentials.cc, line 160 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376#-MvxKXWvCx_DlMcZbQVt:-Mvyo1f2ExGSkMCl6TVw:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e83ff5126041ad26e83882d14cb292011e039884/google/cloud/internal/oauth2_google_credentials.cc#L160)):*

I think if we were writing this code today we would use something like `oauth2_internal::ForceGceOption` and then write:


```cc
  // 3) Check for implicit environment-based credentials (GCE, GAE ...)
  if (options.get<oauth2_internal::ForceGceOption>() || gce_creds->AuthorizationHeader().ok()) {
      return gce_creds;
  }
```

That way the tests can just use the `options` to force GCE credentials.
</blockquote></details>

Done.

coryan (approved):
:lgtm:

> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8376#-:-MvzHIeE18_IdtlrB3Fc:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @dbolduc)

PR #8097, opened 2022-01-25
refactor(bigtable)!: modernize Instance Admin

dbolduc:
Fixes #7534.

This PR is a doozy. Apologies in advance to the reviewer. Changing the connection layer meant that the unit test for the `InstanceAdmin` calls had to be written from scratch.

Implement `InstanceAdmin` on top of the generated Connection class. Reduce `InstanceAdminClient` to be a Connection + project id. This gives us all the features of the generated libraries (like GUAC. yay).

For some reason, `check-api-pr` worked for me locally? I feel like it shouldn't have though.

A minor detail to note:
`internal/async_retry_unary_rpc_and_poll.h` was used by both `InstanceAdmin` and `TableAdmin`. Now it is only used by `TableAdmin`.

The test (`internal/async_retry_unary_rpc_and_poll_test.cc`) was written in terms of `InstanceAdmin`. I decided to delete it entirely, even though the class is still in use, because it is not worth temporarily rewriting. The class will be deleted upon conversion of `TableAdmin`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8097)

PR #8383, opened 2022-02-15
refactor(bigtable)!: modernize TableAdmin

### LINKED ISSUE #7530: Consider implementing `TableAdmin`, `AdminClient` in terms of generated `BigtableTableAdminConnection`

dbolduc:
Fixes #7530

This is the `TableAdmin` analog of #8097. I deleted unused, internal files as part of this PR.

The tests were basically written from scratch, but I did borrow protos as strings for the following tests: [CreateTable](https://github.com/googleapis/google-cloud-cpp/blob/2adcc6cd5103af9c71c0d6e78e58b50cfb798e27/google/cloud/bigtable/table_admin_legacy_test.cc#L380-L397), [ModifyColumnFamilies](https://github.com/googleapis/google-cloud-cpp/blob/2adcc6cd5103af9c71c0d6e78e58b50cfb798e27/google/cloud/bigtable/table_admin_legacy_test.cc#L791-L801)

Note that the `WaitForConsistency()` helper was based off of `g::c::i::AsyncRetryLoop`. But it does not have the changes from #8256. These changes are not necessary because we pass in the options as a function argument, not via an `OptionsSpan`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8383)


google/cloud/bigtable/table_admin_test.cc
```diff
@@ -17,10 +17,14 @@
 #include "google/cloud/bigtable/admin/bigtable_table_admin_options.h"
 #include "google/cloud/bigtable/admin/mocks/mock_bigtable_table_admin_connection.h"
 #include "google/cloud/bigtable/testing/mock_policies.h"
+#include "google/cloud/internal/background_threads_impl.h"
 #include "google/cloud/internal/time_utils.h"
-#include "google/cloud/testing_util/async_sequencer.h"
-#include "google/cloud/testing_util/mock_completion_queue_impl.h"
+#include "google/cloud/testing_util/is_proto_equal.h"
+#include "google/cloud/testing_util/status_matchers.h"
+#include "google/bigtable/admin/v2/bigtable_table_admin.pb.h"
```
dbolduc, line 24: this include shouldn't be here. I will wait for review comments before making the change though.

coryan (approved):
Some minor performance improvement suggestions, but otherwise LGTM.

PR #8391, opened 2022-02-16
fix(rest): formating headers

scotthart:
Some GCP services are very particular how the Host header is formatted.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8391)

### DIFF
--- a/google/cloud/internal/rest_client.cc
+++ b/google/cloud/internal/rest_client.cc
@@ -23,6 +23,8 @@
 #include "google/cloud/internal/rest_options.h"
 #include "google/cloud/log.h"
 #include "absl/strings/match.h"
+#include "absl/strings/str_split.h"
+#include "absl/strings/strip.h"
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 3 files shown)


google/cloud/internal/rest_client.cc
```diff
@@ -53,6 +55,15 @@ Status MakeRequestWithPayload(
   return impl.MakeRequest(http_method, payload);
 }
 
+std::string FormatHostHeaderValue(absl::string_view hostname) {
+  if (!absl::ConsumePrefix(&hostname, "https://")) {
+    absl::ConsumePrefix(&hostname, "http://");
+  }
+  std::vector<std::string> splits =
+      absl::StrSplit(hostname, absl::MaxSplits(absl::ByChar('/'), 1));
+  return splits[0];
```
devbww, line 64: Is it worth materializing `splits[1]`?  For example, consider ...
```
  return std::string(hostname.substr(0, hostname.find('/')));
```

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8391)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[google/cloud/internal/rest_client.cc, line 64 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8391#-Mw-mI7UtSE6iTNOZY5M-r1-64:-Mw2XQij-azo9wK2qCZc:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2ff462fadfcd8160ce959347eab9691220c4ea6e/google/cloud/internal/rest_client.cc#L64)):*

Is it worth materializing `splits[1]`?  For example, consider ...
```
  return std::string(hostname.substr(0, hostname.find('/')));
```
</blockquote></details>

Done.

devbww (approved):

PR #8392, opened 2022-02-16
refactor: include examples from top-level CMake

coryan:
This will allow us to add examples to a generated library, without
having to modify the generated `CMakeLists.txt` file.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8392)

### DIFF
--- a/cmake/GoogleCloudCppCommon.cmake
+++ b/cmake/GoogleCloudCppCommon.cmake
@@ -18,14 +18,8 @@
 get_filename_component(GOOGLE_CLOUD_CPP_SUBPROJECT
                        "${CMAKE_CURRENT_SOURCE_DIR}" NAME)
 
-# C++ Exceptions are enabled by default, but allow the user to turn them off.
-include(EnableCxxExceptions)
-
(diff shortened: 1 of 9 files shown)


cmake/GoogleCloudCppCommon.cmake
```diff
@@ -18,15 +18,9 @@
 get_filename_component(GOOGLE_CLOUD_CPP_SUBPROJECT
                        "${CMAKE_CURRENT_SOURCE_DIR}" NAME)
 
-# C++ Exceptions are enabled by default, but allow the user to turn them off.
-include(EnableCxxExceptions)
```
dbolduc, line 22: The CMake novice asks if this deletion changes behavior at all

dbolduc (approved):

cmake/GoogleCloudCppCommon.cmake
coryan: It is already added in line 278 of the top-level CMakeList.txt file.

PR #8395, opened 2022-02-16
cleanup: remove crash handler

### LINKED ISSUE #4236: Figure a better way to run all examples and unit tests w/ the CrashHandler
In https://github.com/googleapis/google-cloud-cpp/pull/4150 I'm adding a "crash handler" library that lives in `google/cloud/testing_util/crash_handler`. It currently requires the `void InstallCrashHandler(char const* argv0);` function to be called early in `main()`. This means that any unit tests or example binary that wants to use the crash handler needs to define a `main()` and call this function, as is done in my PR linked above.

In https://github.com/googleapis/google-cloud-cpp/pull/4150#pullrequestreview-414616652 @coryan raised good questions about how we could make this easier to apply to all examples, tests, etc. The linked comment has some concerns.

But regardless, if we can figure out a better way to make this crash handler apply across the board for our tests, that'd be great.

dbolduc:
Fixes #4236 

We will just remove the crash handler completely. It was not super useful for debugging, and our examples don't crash anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8395)


devbww:
Given that the examples `throw`, you might want to add the
```
int main(int argc, char* argv[]) try {
  ...
  return example.Run(argc, argv);
} catch (std::exception const& ex) {
  std::cerr << ex.what() << "\n";
  ::google::cloud::LogSink::Instance().Flush();
  return 1;
}
```
wrapping instead.

devbww:
> our examples don't crash anymore.

:-)

dbolduc:
> Given that the examples `throw`, you might want to add the `/*code*/` wrapping instead.

Oops. Thanks.

devbww (approved):

PR #8397, opened 2022-02-16
impl(rest): add content-length header to requests with payloads

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8397)

### DIFF
--- a/google/cloud/internal/rest_client.cc
+++ b/google/cloud/internal/rest_client.cc
@@ -48,9 +48,16 @@ Status MakeRequestWithPayload(
       concatenated_payload += std::string(p.begin(), p.end());
     }
     encoded_payload = impl.MakeEscapedString(concatenated_payload);
+    impl.SetHeader(absl::StrCat("content-length: ", encoded_payload.size()));
     return impl.MakeRequest(http_method,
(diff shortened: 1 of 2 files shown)


google/cloud/internal/rest_client.cc
```diff
@@ -48,9 +48,16 @@ Status MakeRequestWithPayload(
       concatenated_payload += std::string(p.begin(), p.end());
     }
     encoded_payload = impl.MakeEscapedString(concatenated_payload);
+    impl.SetHeader(absl::StrCat("content-length: ", encoded_payload.size()));
     return impl.MakeRequest(http_method,
                             {{encoded_payload.data(), encoded_payload.size()}});
   }
+
+  int content_length = 0;
```
coryan: This should be a `std::size_t`, on MSVC `int` is too narrow (in theory, in practice it is fine, but the compiler will warn).

coryan (approved):

google/cloud/internal/rest_client.cc
coryan: Ah, you changed that already. Ignore this.

PR #8404, opened 2022-02-17
fix(common): use original options when resuming streaming RPCs

coryan:
From the discussion on https://github.com/googleapis/google-cloud-cpp/pull/8355#issuecomment-1042714059

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8404)

### DIFF
--- a/google/cloud/internal/resumable_streaming_read_rpc.h
+++ b/google/cloud/internal/resumable_streaming_read_rpc.h
@@ -16,6 +16,7 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_RESUMABLE_STREAMING_READ_RPC_H
 
 #include "google/cloud/internal/streaming_read_rpc.h"
+#include "google/cloud/options.h"
 #include "google/cloud/version.h"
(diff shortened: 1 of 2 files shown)


devbww (changes requested):
> I was thinking about `google::cloud::internal::ResumableStreamingReadRpc<>`, I just created https://github.com/googleapis/google-cloud-cpp/pull/8404, you can tell me there if the changes are not needed.

No, I don't think these changes are needed.  That is, these new `OptionsSpans` are redundant, and we probably should avoid creating them, if only for performance reasons.  It is also nice to have the span stuff all in one place, and as close to the user as possible.

In this case, a `ResumableStreamingReadRpc` is always immediately transferred into a `StreamReader<T>` functor, and that is then moved into a `StreamRange<T>` via `MakeStreamRange<T>()`.  From there, the reader is only ever invoked/destroyed under the restored options, as the user handles the `StreamRange<T>`, due to the changes already made to `google/cloud/stream_range.h`.

coryan:
Thanks for looking into it.

PR #8407, opened 2022-02-17
cleanup(common): better names for MergeOptions parameters

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8407)


devbww:
Strangely enough, I have a "merge-options-argument-names" branch open with the same intent.  I was waiting on a reply from a mathematician friend about whether there were conventional names for the operation and its arguments.  But short of ever receiving an answer, I happy to go with these ones.

devbww (approved):

PR #8412, opened 2022-02-18
fix(pubsub): restore MakeSchemaAdminConnection overload

coryan:
I removed the implementation by accident. It looked unused, and I
thought it was in the `pubsub_internal` namespace.  Good thing I caught
this before the release.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8412)

### DIFF
--- a/google/cloud/pubsub/schema_admin_connection.h
+++ b/google/cloud/pubsub/schema_admin_connection.h
@@ -102,7 +102,8 @@ class SchemaAdminConnection {
  *     existing code that calls `MakeSchemaAdminConnection({})` from breaking,
  *     due to ambiguity.
  *
- * @deprecated Please use `MakeSchemaAdminConnection()` instead.
+ * @deprecated Please use the `MakeSchemaAdminConnection` function that accepts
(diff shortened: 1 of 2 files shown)


google/cloud/pubsub/schema_admin_connection.h
```diff
@@ -102,7 +102,8 @@ class SchemaAdminConnection {
  *     existing code that calls `MakeSchemaAdminConnection({})` from breaking,
  *     due to ambiguity.
  *
- * @deprecated Please use `MakeSchemaAdminConnection()` instead.
+ * @deprecated Please use the `MakeSchemaAdminConnection` function that accepts
+ *     `google::cloud::Options` instead.
```
dbolduc, line 106: My thinking was that if someone wanted to call `MakeSchemaAdminConnection({})`, they should really be calling `MakeSchemaAdminConnection()`.

But if someone is looking through the documentation wondering which overload to use, it is probably best to redirect them to the `Options` one.

dbolduc (approved):
Oops. Why didn't `check-api` catch this?

PR #8416, opened 2022-02-18
feat(GCS+gRPC): implement `UpdateBucket()`

### LINKED ISSUE #4163: Implement storage::GrpcClient::UpdateBucket

coryan:
Since this uses the same RPC as `PatchBucket()` I just needed to
implement the `ToProto()` helper, the RPC itself was easy after that.
As usual, I also implemented the unit and integration test for this new
RPC.

Fixes #4163

### DIFF
--- a/google/cloud/storage/internal/grpc_bucket_request_parser.h
+++ b/google/cloud/storage/internal/grpc_bucket_request_parser.h
@@ -38,6 +38,8 @@ struct GrpcBucketRequestParser {
 
   static StatusOr<google::storage::v2::UpdateBucketRequest> ToProto(
       PatchBucketRequest const& request);
+  static StatusOr<google::storage::v2::UpdateBucketRequest> ToProto(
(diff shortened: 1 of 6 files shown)


google/cloud/storage/internal/grpc_bucket_request_parser.h
```diff
@@ -38,6 +38,8 @@ struct GrpcBucketRequestParser {
 
   static StatusOr<google::storage::v2::UpdateBucketRequest> ToProto(
       PatchBucketRequest const& request);
+  static StatusOr<google::storage::v2::UpdateBucketRequest> ToProto(
```
dbolduc, line 41: this can not return a bad status. do we want this signature?
```cc
google::storage::v2::UpdateBucketRequest ToProto(UpdateBucketRequest const& request);
```

dbolduc (approved):

PR #8421, opened 2022-02-19
fix(oauth2): compute engine credential query

scotthart:
Querying the gce metadata needs to be a GET instead of a POST and must be able to follow redirects.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8421)

### DIFF
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -654,10 +654,8 @@ Status CurlImpl::MakeRequest(CurlImpl::HttpMethod method,
   using HttpMethod = CurlImpl::HttpMethod;
   handle_.SetOption(CURLOPT_CUSTOMREQUEST, HttpMethodAsChar(method));
   handle_.SetOption(CURLOPT_UPLOAD, 0L);
-  if (options_.has<CurlFollowLocationOption>()) {
-    handle_.SetOption(CURLOPT_FOLLOWLOCATION,
-                      options_.get<CurlFollowLocationOption>() ? 1L : 0L);
-  }
+  handle_.SetOption(CURLOPT_FOLLOWLOCATION,
+                    options_.get<CurlFollowLocationOption>() ? 1L : 0L);
 
   if (method == HttpMethod::kGet) {
     handle_.SetOption(CURLOPT_NOPROGRESS, 1L);
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -654,6 +654,7 @@ Status CurlImpl::MakeRequest(CurlImpl::HttpMethod method,
   using HttpMethod = CurlImpl::HttpMethod;
   handle_.SetOption(CURLOPT_CUSTOMREQUEST, HttpMethodAsChar(method));
   handle_.SetOption(CURLOPT_UPLOAD, 0L);
+  handle_.SetOption(CURLOPT_FOLLOWLOCATION, 1L);
 
   if (method == HttpMethod::kGet) {
     handle_.SetOption(CURLOPT_NOPROGRESS, 1L);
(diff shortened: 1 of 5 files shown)


google/cloud/internal/oauth2_compute_engine_credentials.cc
```diff
@@ -127,14 +127,13 @@ ComputeEngineCredentials::DoMetadataServerGetRequest(std::string const& path,
   request.SetPath(path);
   request.AddHeader("metadata-flavor", "Google");
   if (recursive) request.AddQueryParameter("recursive", "true");
-  return rest_client_->Post(request,
-                            std::vector<std::pair<std::string, std::string>>{});
+  return rest_client_->Get(request);
```
coryan, line 130: Sorry I did not catch this.

google/cloud/internal/curl_impl.cc
```diff
@@ -654,6 +654,7 @@ Status CurlImpl::MakeRequest(CurlImpl::HttpMethod method,
   using HttpMethod = CurlImpl::HttpMethod;
   handle_.SetOption(CURLOPT_CUSTOMREQUEST, HttpMethodAsChar(method));
   handle_.SetOption(CURLOPT_UPLOAD, 0L);
+  handle_.SetOption(CURLOPT_FOLLOWLOCATION, 1L);
```
coryan, line 657: Do we want this for *all* requests?  Should this be an option, maybe in the `*_internal::` namespace?

coryan (approved):

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8421)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[google/cloud/internal/curl_impl.cc, line 657 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8421#-MwEu3-lVgU6zQ6MOVPB-r1-657:-MwIdrx38gFlZBx5YHxl:b8kiixu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b71488c00ca4777a9b29c9271f5d8b49942ab38f/google/cloud/internal/curl_impl.cc#L657)):*

Do we want this for *all* requests?  Should this be an option, maybe in the `*_internal::` namespace?
</blockquote></details>

Added option to allow library users to specify FOLLOWLOCATION.No owr
___
*[google/cloud/internal/oauth2_compute_engine_credentials.cc, line 130 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8421#-MwEu3-lVgU6zQ6MOVPC-r1-130:-MwIe9qJ-OfYWMZwhig-:boj0p82) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b71488c00ca4777a9b29c9271f5d8b49942ab38f/google/cloud/internal/oauth2_compute_engine_credentials.cc#L130)):*

Sorry I did not catch this.
</blockquote></details>

No worries; that's what testing is for.

google/cloud/internal/curl_impl.cc
```diff
@@ -654,6 +654,10 @@ Status CurlImpl::MakeRequest(CurlImpl::HttpMethod method,
   using HttpMethod = CurlImpl::HttpMethod;
   handle_.SetOption(CURLOPT_CUSTOMREQUEST, HttpMethodAsChar(method));
   handle_.SetOption(CURLOPT_UPLOAD, 0L);
+  if (options_.has<CurlFollowLocationOption>()) {
```
devbww, line 657: When the `!has` case is equivalent to `T{}`, and with a `bool` option in particular, I would proffer that it is better (faster and clearer) to drop the `has` check.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8421)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @coryan and @devbww)

___
*[google/cloud/internal/curl_impl.cc, line 657 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8421#-MwEu3-lVgU6zQ6MOVPB-r2-657:-MwIl_JI5P9UIPa1IRzO:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b8854fcf0dff6f87f728d2a79d7953bd74be7ca0/google/cloud/internal/curl_impl.cc#L657)):*

When the `!has` case is equivalent to `T{}`, and with a `bool` option in particular, I would proffer that it is better (faster and clearer) to drop the `has` check.
</blockquote></details>

Done.

PR #8437, opened 2022-02-23
fix(bigtable): cast to std::string which works with string and Cord

devbww:
This fixes the import.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8437)

### DIFF
--- a/google/cloud/bigtable/async_row_reader.h
+++ b/google/cloud/bigtable/async_row_reader.h
@@ -375,7 +375,8 @@ class AsyncRowReader : public std::enable_shared_from_this<
       }
     }
     if (!response.last_scanned_row_key().empty()) {
-      last_read_row_key_ = std::move(*response.mutable_last_scanned_row_key());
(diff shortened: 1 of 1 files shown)


dbolduc:
Hmm do we need to do the same thing here?

https://github.com/googleapis/google-cloud-cpp/blob/083d568806ae061126ba49fbb5f0bcbb2a8d64f5/google/cloud/bigtable/row_reader.cc#L144

devbww (commented):
No, because that `last_read_row_key_` was of the right type.  Which suggested a better fix.  PTAL.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8437#-:-Mw_7RxY9llCYivYEETd:bv1h138)* status: 0 of 1 files reviewed, all discussions resolved

dbolduc (approved):
Ah. This is better.

PR #8233, opened 2022-02-04
feat(pubsublite): resumable async streaming read write rpc

18suresha:
Implements `Read` and `Write` of a generic resumable bidi rpc async stream following https://github.com/googleapis/google-cloud-cpp/pull/8450.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8233)

PR #8441, opened 2022-02-23
feat(pubsublite): resumable async read write rpc stream boilerplate

18suresha:
this serves to reduce the LOC count of https://github.com/googleapis/google-cloud-cpp/pull/8233 by including the boilerplate code of the latter

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8441)


dpcollins-google (approved):

google/cloud/pubsublite/google_cloud_cpp_pubsublite_mocks.bzl
```diff
@@ -18,6 +18,9 @@
 
 google_cloud_cpp_pubsublite_mocks_hdrs = [
     "mocks/mock_admin_connection.h",
+    "internal/mock_async_reader_writer.h",
```
coryan, line 21: Please see my comments in the other PR

coryan (changes requested):
This is helpful, thanks.  I left comments in the other PR about some of the files here.  Please address them so we can start merging bite-sized changes.

google/cloud/pubsublite/google_cloud_cpp_pubsublite_mocks.bzl
18suresha: yup just pushed some changes

google/cloud/pubsublite/internal/futures.h
```diff
@@ -0,0 +1,51 @@
...
+  // Note: this drops any exceptions contained in `future<U>`
+  future<T> operator()(future<U>) {
+    return std::move(f);
+  }
+
+  future<T> f;
+};
+
+/**
+ * This is a templatized helper to chain futures together with an intended
```
coryan, line 37: Well, yes. I can see that.  My question is why do you need this instead of some `.then()` calls and lambdas? Is that because you need to capture-by-move the returned future? That is, because this does not work:

```
future<T> f  = ...;
future<U> g = ...;
return g.then([ff = std::move(f)](future<U>) mutable { return std::move(ff); }
```

If so, maybe add something like "We often need to capture and return futures from `.then()` continuations, but without C++14 support, this is very tedious, this class and function reduce the effort" to the comment.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -0,0 +1,129 @@
...
+using StreamInitializer = std::function<future<StatusOr<
+    std::unique_ptr<AsyncStreamingReadWriteRpc<RequestType, ResponseType>>>>(
+    std::unique_ptr<AsyncStreamingReadWriteRpc<RequestType, ResponseType>>)>;
+
+using AsyncSleeper = std::function<future<void>(std::chrono::milliseconds)>;
+
+using RetryPolicyFactory = std::function<std::unique_ptr<RetryPolicy>()>;
+
+/**
+ * The purpose of this class is to encapsulate the retry logic of a generic
```
coryan, line 60: You have not answered my question about how this is used.  Without answering that question we do not know if this is the right API.  We can strip this class from this PR if you think that discussion will take a while, or we can figure out what the API should be and then merge this PR. Your call.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -0,0 +1,129 @@
...
+   * If `true` is returned, the call was successful. If `false` is returned,
+   * the call failed, but the user may call `Write` again unless `Start` had
+   * finished with a permanent error `Status`.
+   */
+  virtual future<bool> Write(RequestType const&, grpc::WriteOptions) = 0;
+
+  /**
+   * Finishes the streaming RPC.
+   *
+   * This will cause any outstanding `Read` or `Write` to fail.
```
coryan, line 119: Note that this is not the behavior in gRPC:

https://grpc.github.io/grpc/cpp/classgrpc_1_1internal_1_1_client_async_streaming_interface.html#a0c4fd4be2228fa196ae446502828e23a

In particular, calling `Finish()` for the underlying stream is only possible if:

> there are no more messages to be received from the server (this can be known implicitly by the calling code, or explicitly from an earlier call to [AsyncReaderInterface::Read](https://grpc.github.io/grpc/cpp/classgrpc_1_1internal_1_1_async_reader_interface.html#adf87c602036d69158950b7299d0aae70) that yielded a failed result, e.g. cq->Next(&read_tag, &ok) filled in 'ok' with 'false').

Is your intention is to call `Finish()` (in the underlying stream) while a `Read()` call is outstanding?  Then you will also need to (a) call `Cancel()` to terminate the `Read()` call, then (b) drain all the `Read()` calls that return `true`, and finally (c) call `Finish()` after the first `Read()` that returns `false`.

I am not sure your comments / API proposal are necessarily bad or not.  At this point I am just asking what you want the API to do.

coryan (changes requested):

google/cloud/pubsublite/internal/futures.h
dpcollins-google: +1, this is the reason for this helper.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
dpcollins-google: This is the intention yes, which could be more clear in documentation. It is to abstract away handling of how Read() and Write() calls to this object correspond to Read() and Write() calls on the underlying stream object, which may change over time as streams fail.

The intention of this is that, by the time Finish() returns, the future returned from `Start` will have the final status (if it doesn't already due to a permanent error), any outstanding Read() and Write() futures from this object will be failed, and any future calls to Read() or Write() will fail immediately.

This is largely to encapsulate the retry logic needed for classes like StreamingSubscriptionBatchSource, of which there will be 4 in Pub/Sub Lite code, to prevent needing to write complicated retry logic multiple times.

 https://github.com/googleapis/google-cloud-cpp/blob/c9f7f317a245fbeac2af634953ccc3d6eb6e639b/google/cloud/pubsub/internal/streaming_subscription_batch_source.h#L63

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -59,6 +59,34 @@ using RetryPolicyFactory = std::function<std::unique_ptr<RetryPolicy>()>;
...
  * The purpose of this class is to encapsulate the retry logic of a generic
  * bidirectional asynchronous RPC stream.
+ *
+ * Example (unrealistic) usage:
+ * ```
+ * std::string ExampleReadUsage() {
+ *   ResumableAsyncStreamingReadWriteRpc<Request, Response> resumable_stream;
+ *   auto start = resumable_stream.Start();
+ *   size_t num_attempts = 0;
+ *   while (num_attempts < 3) {
```
dpcollins-google, line 69: this shouldn't be a loop.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -59,6 +59,34 @@ using RetryPolicyFactory = std::function<std::unique_ptr<RetryPolicy>()>;
...
+ * finished, so there had to be a permanent error in the retry loop";
+ *       }
+ *     } else {
+ *       return read_response->Stringify();
+ *     }
+ *     // we are ready to call `Read` again due to the object's internal retry
+ * mechanism.
+ *     ++num_attempts;
+ *   }
+ *   resumable_stream.Finish().get();
```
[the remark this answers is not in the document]
dpcollins-google: +1 for `Close`

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -59,6 +59,34 @@ using RetryPolicyFactory = std::function<std::unique_ptr<RetryPolicy>()>;
...
+ *     if (!read_response) {
+ *       // the second conditional is redundant since the only way `Start` can
+ * // finish without a user calling `Finish` is if a retry loop finishes with a
+ * // permanent error (not ok) status.
+ *       if (start.is_ready() && !start.get().ok()) {
+ *         return "The read response optional was not engaged, and `Start`
+ * finished, so there had to be a permanent error in the retry loop";
+ *       }
+ *     } else {
+ *       return read_response->Stringify();
```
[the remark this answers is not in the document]
dpcollins-google: I think a better example would be:

```
using Underlying = std::unique_ptr<StreamingReadWriteRpc<Req, Res>>;
future<StatusOr<Underlying>> Initialize(Underlying to_init) {
  auto initial_write = to_init.Write(GetInitialRequest());
  auto to_init_shared = std::make_shared<Underlying>(std::move(to_init));
  auto initial_read = initial_write.then([=](future<bool> write_ok){
    if (!write_ok.get()) return make_ready_future(absl::optional<Res>());
    return (**to_init_shared).Read();
  });
  return initial_read.then([=](future<absl::optional<Res>> read){
    auto read_res = read.get();
    if (!read.has_value()) return future<StatusOr<Underlying>>((**to_init_shared).Finish());
    return make_ready_future(StatusOr<Underlying>(std::move(*to_init_shared)));
  });
}

Status Example() {
  std::unique_ptr<ResumableAsyncStreamingReadWriteRpc<Req, Res>> resumable_stream = MakeResumableStream(&Initialize);
  future<Status> final_status = stream->Start();
  if (final_status.is_ready()) return final_status.get();
  while (!final_status.is_ready()) {
    auto message_future_1 = stream->Write(GetMessage());
    if (!message_future_1.get()) continue;
    auto message_future_2 = stream->Write(GetMessage());
    if (!message_future_2.get()) continue;
    auto response_1 = stream->Read().get();
    if (!response_1.has_value()) continue;
    ProcessResponse(*response_1);
    auto response_2 = stream->Read().get();
    if (!response_2.has_value()) continue;
    ProcessResponse(*response_2);
    resumable_stream.Finish().get();
    return Status();
  }
  return final_status.get();
}
```

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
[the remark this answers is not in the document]
18suresha: yea this has to do with the 'unrealistic' part
I just wanted a loop to show how you can automatically retry your `Read` call. `num_attempts` is intended as just a counter variable without any legitimate usage beyond that despite its name

yup it would be the same for `Write`

18suresha: https://github.com/googleapis/google-cloud-cpp/pull/8441#discussion_r814015630

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -0,0 +1,191 @@
...
+   * is successfully shut down (in which case in contains an ok status),
+   * or when the retry policies to resume the stream are exhausted. The
+   * latter includes the case where the stream fails with a permanent
+   * error.
+   *
+   * While the stream is usable immediately after this function returns,
+   * any `Read()` or `Write()` calls will fail until the stream is initialized
+   * successfully.
+   */
+
```
coryan, line 118: nit: remove extra blank line

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -0,0 +1,191 @@
...
+ *     if (!response_2.has_value()) continue;
+ *     ProcessResponse(*response_2);
+ *     resumable_stream.Shutdown().get();
+ *     return Status();
+ *   }
+ *   return final_status.get();
+ * }
+ * @endcode
+ */
+
```
coryan, line 99: nit: remove extra blank line

google/cloud/pubsublite/internal/futures.h
coryan: Can we change the comments to say that?  Maybe something along these lines:

```cc
/**
  * A helper to capture-by-move futures into a second future continuation.
  *
  * Given two futures `future<U> r` and `future<T> f` we often want to write:
   * @code
   * f.then([tmp = std::move(f)](future<T>) mutable { return std::move(tmp); } 
   * @encode
  *
  * Unfortunately we cannot, as the project needs to support C++11. This is 
  * a helper to avoid repetition on this pattern.
  */
```

google/cloud/pubsublite/internal/futures.h
```diff
@@ -0,0 +1,55 @@
...
+ * internal return type of `T` irrespective of the type passed in of
+ * `future<U>`. Without this, the previous solution was to wrap a shared pointer
+ * around the future we wanted to chain (no generalized lambda capture in C++11)
+ * and capture that into the original future's continuation. This helper removes
+ * and encapsulates that complexity/effort by moving the future into a functor
+ * as a member variable and returning that upon invoking the continuation.
+ */
+template <class T>
```
coryan, line 45: s/class T/typename T/

google/cloud/pubsublite/internal/futures.h
```diff
@@ -0,0 +1,55 @@
...
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_PUBSUBLITE_INTERNAL_FUTURES_H
+
+#include "google/cloud/future.h"
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace pubsublite_internal {
+
+template <class T>
```
coryan, line 25: s/class T/typename T/ here and elsewhere in this file.

coryan (changes requested):

coryan:
FWIW, looks good to me, let's see what the robots say.

coryan (approved):

PR #8446, opened 2022-02-24
fix(storage): json["projectTeam"] might be present but null

martinzink:
We are working on google-cloud integration for our project [Apache Nifi Minifi CPP](https://github.com/apache/nifi-minifi-cpp)
And I found that the `projectTeam` is an optional field in the `ObjectMetadata` json, however it will crash if it is present but null.
This makes impossible to make integration tests using the [fake-gcs-server](https://github.com/fsouza/fake-gcs-server), because the response to a ResumableUploadRequest is something like this due to the datastructure in [google-cloud-go](https://github.com/googleapis/google-cloud-go/blob/main/storage/acl.go#L60)
```
{
   "bucket":"my-bucket",
   "name":"my-key",
   "contentType":"",
   "contentEncoding":"",
   "acl":[
      {
         "entity":"projectOwner",
         "entityId":"",
         "role":"OWNER",
         "domain":"",
         "email":"",
         "projectTeam":null
      }
   ],
   "created":"0001-01-01T00:00:00Z",
   "updated":"0001-01-01T00:00:00Z",
   "deleted":"0001-01-01T00:00:00Z"
}
```
which will crash with the following error 
```
  what():  [json.exception.type_error.306] cannot use value() with null
Signal: SIGABRT (Aborted)
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8446)

### DIFF
--- a/google/cloud/storage/internal/access_control_common_parser.cc
+++ b/google/cloud/storage/internal/access_control_common_parser.cc
@@ -36,10 +36,12 @@ Status AccessControlCommonParser::FromJson(AccessControlCommon& result,
   result.self_link_ = json.value("selfLink", "");
   if (json.count("projectTeam") != 0) {
     auto tmp = json["projectTeam"];
-    ProjectTeam p;
-    p.project_number = tmp.value("projectNumber", "");
-    p.team = tmp.value("team", "");
-    result.project_team_ = std::move(p);
+    if (!tmp.is_null()) {
+      ProjectTeam p;
+      p.project_number = tmp.value("projectNumber", "");
+      p.team = tmp.value("team", "");
+      result.project_team_ = std::move(p);
+    }
   }
   return Status();
 }
--- a/google/cloud/storage/internal/access_control_common_parser.cc
+++ b/google/cloud/storage/internal/access_control_common_parser.cc
@@ -36,12 +36,12 @@ Status AccessControlCommonParser::FromJson(AccessControlCommon& result,
   result.self_link_ = json.value("selfLink", "");
   if (json.count("projectTeam") != 0) {
     auto tmp = json["projectTeam"];
-    if (!tmp.is_null()) {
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/access_control_common_parser.cc
```diff
@@ -36,10 +36,12 @@ Status AccessControlCommonParser::FromJson(AccessControlCommon& result,
   result.self_link_ = json.value("selfLink", "");
   if (json.count("projectTeam") != 0) {
     auto tmp = json["projectTeam"];
-    ProjectTeam p;
-    p.project_number = tmp.value("projectNumber", "");
-    p.team = tmp.value("team", "");
-    result.project_team_ = std::move(p);
+    if (!tmp.is_null()) {
```
coryan, line 39: Could you change this to `if (tmp.is_null()) return Status{};` ?

coryan (commented):
First thanks for the PR, and including a test!  A small nit below, let me know what you think.

/gcbrun

google/cloud/storage/internal/access_control_common_parser.cc
martinzink: sure thing, I changed it in https://github.com/googleapis/google-cloud-cpp/pull/8446/commits/739c7f0e3af09ee680c62ab9e4565a224e0966c2

coryan:
The public CI build results are sometimes hard to find, see https://github.com/googleapis/google-cloud-cpp/pull/8446#issuecomment-1049844494 where it links to:

https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/8446/739c7f0e3af09ee680c62ab9e4565a224e0966c2/index.html

coryan (approved):

PR #8447, opened 2022-02-24
cleanup(storage): report transfer size in benchmark

### LINKED ISSUE #4349: Storage throughput benchmarks should reports bytes transferred
Currently the benchmark downloads only the full object. It would be good to test downloading only a portion of the object, which requires distinguishing the size of the object vs. the bytes transferred. This is also useful when the transfer has a partial failure and is interrupted in the middle.

coryan:
For the most part this just creates a placeholder to report the transfer
size. While the transfer size and object size could be different in case
of (unrecoverable) errors during download, the real motivation is to
support partial downloads in a future PR.

Fixes #4349

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8447)

### DIFF
--- a/google/cloud/storage/benchmarks/throughput_experiment.cc
+++ b/google/cloud/storage/benchmarks/throughput_experiment.cc
@@ -16,9 +16,6 @@
 #include "google/cloud/storage/benchmarks/benchmark_utils.h"
 #include "google/cloud/storage/client.h"
 #include "google/cloud/storage/grpc_plugin.h"
-#if GOOGLE_CLOUD_CPP_STORAGE_HAVE_GRPC
-#include "google/cloud/storage/internal/grpc_client.h"
-#endif  // GOOGLE_CLOUD_CPP_STORAGE_HAVE_GRPC
 #include "google/cloud/grpc_error_delegate.h"
 #include "google/cloud/internal/getenv.h"
 #include "absl/algorithm/container.h"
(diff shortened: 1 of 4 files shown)


dbolduc (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8447)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @dbolduc)

___
*[google/cloud/storage/benchmarks/throughput_experiment.cc, line 92 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8447#-MwhHoQ21hO2zVQdocG3:-MwhHoQ21hO2zVQdocG4:bn69qe7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/51b37d418d73cf1b40f6a522698485c407d04707/google/cloud/storage/benchmarks/throughput_experiment.cc#L92)):*
> ```cpp
>       auto len = config.app_buffer_size;
>       if (offset + static_cast<std::int64_t>(len) > config.object_size) {
>         len = static_cast<std::size_t>(config.object_size - offset);
>       }
> ```

Unrelated nit. I am not thinking deeply about the difference between `std::size_t` and `std::int64_t`, but seems like this could be simpler. No need to change though.
```
auto len = std::min<std::size_t>(config.object_size - offset, config.app_buffer_size);
```

dbolduc (approved):

PR #8449, opened 2022-02-24
refactor(bigtable)!: mark InstanceAdminClient as final

### LINKED ISSUE #8419: Make `bigtable::*AdminClient` classes `final`
We do not want users to extend the `AdminClient` or `InstanceAdminClient` classes. Currently they are virtual interfaces, and we define a `Default*AdminClient` in an anonymous namespace in the `.cc` file.

I think we should move the implementation into the base class. (Still not offer any public constructors). And mark those classes as `final`. I think this makes the intention of our API the most clear.

dbolduc:
Fixes #8419

The `InstanceAdmin` equivalent of #8424. I also added a note in the changelog, because this is technically a break since the last release.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8449)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,7 +108,7 @@ The library has been expanded to include the following services:
 
 ### [Bigtable](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/bigtable/README.md)
 
-**BREAKING CHANGE:** The `bigtable::TableAdminClient` interface has changed
+**BREAKING CHANGE:** The `bigtable::AdminClient` interface has changed
 significantly. Any code that extends this class or calls its experimental public
 APIs (`reset()`, `Channel()`) will be broken. For the most part, this should
 only affect customers who mock this class in their tests. Code that calls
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -140,6 +140,10 @@ information on these new classes, see our [Architecture Design] document.
 Again, we apologize for making this breaking change, but we believe it is in the
 best long-term interest of our customers.
 
+**BREAKING CHANGE:** The `bigtable::InstanceAdminClient` class has been marked
+as `final`. It should not be extended for mocking. Customers should instead use
```
coryan, line 144: Maybe: s/It should not be extended for mocking/After the changes to this class in v1.3X.YY, we do not expect there is any need (or point) in extending this class./ ??

coryan (approved):

CHANGELOG.md
dbolduc: I like the suggestion. Thanks.

PR #8450, opened 2022-02-24
init commit

18suresha:
Implementation of `Start`, `Shutdown`, and retry logic for interface from https://github.com/googleapis/google-cloud-cpp/pull/8441


google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,258 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+      ResumableAsyncStreamingReadWriteRpc<RequestType, ResponseType>&&) =
+      delete;
+  ResumableAsyncStreamingReadWriteRpcImpl& operator=(
+      ResumableAsyncStreamingReadWriteRpc<RequestType, ResponseType>&&) =
+      delete;
+
+  future<Status> Start() override {
+    future<Status> status_future;
+    {
+      std::lock_guard<std::mutex> g{mu_};
```
coryan, line 219: nit: elsewhere in the project we call these variables `lk`.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,258 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+    return root_future;
+  }
+
+ private:
+  using UnderlyingStream =
+      std::unique_ptr<AsyncStreamingReadWriteRpc<RequestType, ResponseType>>;
+
+  enum class State { kRetrying, kInitialized, kShutdown };
+
+  void SetReadWriteFuturesLockHeld() {
```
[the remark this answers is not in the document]
coryan: Adjust the `lk` parameter to `std::unique_lock<std::mutex> const&` or `std::unique_lock<std:::mutex>&` as appropriate.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,258 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+    {
+      std::lock_guard<std::mutex> g{mu_};
+      switch (stream_state_) {
+        case State::kShutdown:
+          return make_ready_future();
+        case State::kRetrying:
+          stream_state_ = State::kShutdown;
+          root_future = root_future.then(ChainFuture(
+              retry_promise_->get_future().then([this](future<void>) {
+                std::lock_guard<std::mutex> g{mu_};
```
coryan, line 252: It would seem more natural to just call `CompleteStream(Status{})` and have that function acquire the mutex.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,258 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+
+  future<bool> Write(RequestType const&, grpc::WriteOptions) override {
+    // TODO(18suresha) implement
+    return make_ready_future(false);
+  }
+
+  future<void> Shutdown() override {
+    promise<void> root_promise;
+    future<void> root_future = root_promise.get_future();
+    {
```
coryan, line 243: Seems like it would be better to refactor all this code to a function so this reads:

```cc
future<void> Shutdown() override {
  promise<void> p;
  auto future = ScheduleShutdown( // <<< think of a better name
    p.get_future());
 p.set_value();
 return future;
}

...
...

future<void> ScheduleShutdown(future<void> root_future) {
  // as before
}
```

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,258 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+      write_reinit_done.swap(write_reinit_done_);
+    }
+    mu_.unlock();
+    if (read_reinit_done) read_reinit_done->set_value();
+
+    if (write_reinit_done) write_reinit_done->set_value();
+    mu_.lock();
+  }
+
+  void CompleteStreamLockHeld(Status status) {
```
coryan, line 299: Maybe a more meaningful name would be `OnShutdownCompleted()`?  Dunno.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
```diff
@@ -0,0 +1,462 @@
...
+    ::google::cloud::pubsublite_testing::MockAsyncReaderWriter<FakeRequest,
+                                                               FakeResponse>;
+
+using AsyncReadWriteStreamReturnType =
+    std::unique_ptr<AsyncStreamingReadWriteRpc<FakeRequest, FakeResponse>>;
+
+using ResumableAsyncReadWriteStream = std::unique_ptr<
+    ResumableAsyncStreamingReadWriteRpc<FakeRequest, FakeResponse>>;
+
+const Status kFailStatus = Status(StatusCode::kUnavailable, "Unavailable");
```
coryan, line 66: east const

coryan, line 66: I think the Google Style Guide prohibits globals with non-trivial destructors, you probably want to use a function.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
```diff
@@ -0,0 +1,462 @@
...
+using ::testing::InSequence;
+using ::testing::MockFunction;
+using ::testing::Return;
+using ::testing::StrictMock;
+
+using ::google::cloud::pubsublite_testing::MockBackoffPolicy;
+using ::google::cloud::pubsublite_testing::MockRetryPolicy;
+
+using ms = std::chrono::milliseconds;
+const unsigned int kFutureWaitMs = 100;
```
coryan, line 45: consider:

```
auto constexpr kFutureWait = std::chrono::milliseconds(100);
```

But I wonder why are you introducing the delay?  Is waiting for 100 milliseconds vs. 50 or 10 or 0 any different in these tests?

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
```diff
@@ -0,0 +1,462 @@
...
+
+  EXPECT_CALL(stream2_ref, Finish)
+      .WillOnce(Return(ByMove(make_ready_future(Status()))));
+  auto shutdown = stream_->Shutdown();
+  shutdown.get();
+  EXPECT_THAT(status_future.get(), IsOk());
+}
+
+TEST_F(ResumableAsyncReadWriteStreamingRpcTest,
+       SingleStartFailurePermanentError) {
```
coryan, line 158: I am missing a test for "too many transient failures during start".

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
```diff
@@ -0,0 +1,462 @@
...
+
+  auto backoff_policy = absl::make_unique<StrictMock<MockBackoffPolicy>>();
+  auto& backoff_policy_ref = *backoff_policy;
+  EXPECT_CALL(*backoff_policy_, clone)
+      .WillOnce(Return(ByMove(std::move(backoff_policy))));
+
+  EXPECT_CALL(stream_factory_, Call)
+      .WillOnce(Return(ByMove(std::move(stream1))));
+
+  promise<bool> start_promise;
```
coryan, line 120: Just FYI, it does not seem like you need it in these tests. We have a helper class to test asynchronous stuff:

https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/testing_util/async_sequencer.h

coryan (changes requested):
Thank you for being patient and breaking down the code.

I will make some changes to your PR description. In general the title should be < 72 characters wide.  We typically reserve `feat` for user-visible features, and `impl` for new implementation details that are not (by themselves) user-visible.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,258 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+      if (stream_state_ == State::kShutdown) {
+        return FinishRetryPromiseLockHeld();
+      }
+      assert(stream_state_ == State::kRetrying);
+    }
+
+    if (!retry_policy->IsExhausted() && retry_policy->OnFailure(status)) {
+      sleeper_(backoff_policy->OnCompletion())
+          .then([this, retry_policy, backoff_policy](future<void>) {
+            Initialize(retry_policy, backoff_policy);
```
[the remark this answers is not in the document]
18suresha: `sleeper_` is just supposed to invoke a pure time delay.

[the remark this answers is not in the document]
coryan: That won't work.  You would be blocking the thread, i.e., the thread cannot do any other useful work, while sleeping. This can be a problem because typically there is a limited number of background threads dedicated to this work.

You should use this:

 https://github.com/googleapis/google-cloud-cpp/blob/291e9dd4d3a66ecb3133d79ae58abff3491079cc/google/cloud/completion_queue.h#L95-L97

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
18suresha: the magic number used for the delay is never actually used. Regardless of the value, the sleeper mock in the test suite either returns an immediate future with zero time delay or returns the future of a promise that's satisfied manually later

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
```diff
@@ -0,0 +1,462 @@
...
+  EXPECT_CALL(stream2_ref, Start)
+      .WillOnce(Return(ByMove(make_ready_future(true))));
+
+  EXPECT_CALL(initializer_, Call).WillOnce(InitializeInline());
+
+  start_promise.set_value(false);
+
+  // start future doesn't finish until
+  // permanent error or `Finish`is called
+  EXPECT_EQ(status_future.wait_for(ms(kFutureWaitMs)),
```
coryan, line 147: Here is where `kFutureWaitMs` is used.  That blocks the current thread (the testing thread) for at least `kFutureWaitMs`.  If there is no reason to block it for that long, we should make the time shorter, so the test can finish faster.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
coryan: I do not think that is accurate, see: https://github.com/googleapis/google-cloud-cpp/pull/8450/files#r814918494

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,258 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+  // `Read` or `Write` finish with a failure. The returned future is only
+  // completed when the invoked retry loop completes on success or permanent
+  // error.
+  absl::optional<promise<void>> read_reinit_done_;   // ABSL_GUARDED_BY(mu_)
+  absl::optional<promise<void>> write_reinit_done_;  // ABSL_GUARDED_BY(mu_)
+  // The below two member variables are promises that finish their future when
+  // an internal `Read` or `Write` is finished respectively. This allows us to
+  // perform retry logic (calling `Finish` internally) right when there are no
+  // more outstanding `Read`s or `Write`s.
+  absl::optional<promise<void>> in_progress_read_;   // ABSL_GUARDED_BY(mu_)
```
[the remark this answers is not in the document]
18suresha: yea but they're important to the Shutdown and retry loop logic so I kept them in here

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,258 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+
+  void Initialize(std::shared_ptr<RetryPolicy> retry_policy,
+                  std::shared_ptr<BackoffPolicy> backoff_policy) {
+    future<bool> start_future;
+
+    {
+      std::lock_guard<std::mutex> g{mu_};
+      if (stream_state_ == State::kShutdown)
+        return FinishRetryPromiseLockHeld();
+      stream_ = stream_factory_();
```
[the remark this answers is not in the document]
18suresha: decided to use `std::shared_ptr<std::unique_ptr<StreamType>>` to facilitate only initializing `stream_` on success
otherwise `stream_` would have to be a `std::shared_ptr` which we didn't think made much sense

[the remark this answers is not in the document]
coryan: I don't follow: `std::shared_ptr<StreamType>` does not make sense, but `std::shared_ptr<std::unique_ptr<StreamType>>` makes sense?  Please don't do that.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
18suresha: but wouldn't that require an extra unlock before invoking the function, along with that function acquiring the lock again (specifically around line 282)?

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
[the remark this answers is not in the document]
18suresha: the reason being that upon successful initialization, it only makes sense for the class to have ownership of the underlying stream
though it sounds like the weirdness of the double indirection via `std::shared_ptr<std::unique_ptr<StreamType>>` outweighs the reasoning behind keeping the member variable as an `std::unique_ptr`?

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
coryan: I see. SGTM then.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +182,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+    lk.lock();
+  }
+
+  void CompleteUnsatisfiedOpsLockHeld(Status status,
+                                      std::unique_lock<std::mutex>& lk) {
+    SetReadWriteFuturesLockHeld(lk);
+    status_promise_.set_value(std::move(status));
+  }
+
+  void FinishRetryPromiseLockHeld(std::unique_lock<std::mutex>& lk) {
```
coryan, line 312: There is no need to append `LockHeld` to these functions.  Obviously the lock is held because they receive the lock guard as a parameter.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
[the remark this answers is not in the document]
dpcollins-google: This is going to invoke MakeRelativeTimer under the hood, but is a smaller, simpler API. There's no reason to care if the completion queue is shut down and this returns an error- that just means we'll create a new stream, which should fail with a permanent error if, for example, the completion queue is shut down.

I don't think there's any need to handle more complexity here.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +182,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+    assert(retry_promise_.has_value());
+    promise<void> retry_promise = std::move(*retry_promise_);
```
dpcollins-google, line 314: std::move(retry_promise_.value())

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
dpcollins-google: This will, in practice, never happen, as the streams should retry forever on transient errors like network outages. But we can add a test that this fails permanently when IsExhausted returns true

coryan: Hmmm... If I understand what you are suggesting then if the customer misconfigures the library (for example, but not limited to: set the endpoint to something that rejects connections) they would get no notification whatsoever, the library will just keep retrying after getting `kUnavailable`.

In other libraries we set a limit (time or error count base) on retries for unsuccessful streams (i.e., those that get no data ever).  If the stream gets some data we allow then retry regardless of the error code.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +182,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+        return FinishRetryPromiseLockHeld(lk);
+      }
+      stream = std::make_shared<UnderlyingStream>(stream_factory_());
+      start_future = (*stream)->Start();
+    }
+
+    auto start_future_result =
+        start_future.then([this, stream](future<bool> start_future) {
+          if (start_future.get()) return make_ready_future(Status());
+          std::lock_guard<std::mutex> lk{mu_};
```
dpcollins-google: remove

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +182,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+      std::unique_lock<std::mutex> lk{mu_};
+      if (stream_state_ == State::kShutdown) {
+        return FinishRetryPromiseLockHeld(lk);
+      }
+      stream = std::make_shared<UnderlyingStream>(stream_factory_());
+      start_future = (*stream)->Start();
+    }
+
+    auto start_future_result =
+        start_future.then([this, stream](future<bool> start_future) {
```
dpcollins-google: remove

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +182,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+          if (start_future.get()) return make_ready_future(Status());
+          std::lock_guard<std::mutex> lk{mu_};
+          return (*stream)->Finish();
+        });
+
+    auto initializer_future =
+        start_future_result.then([this, stream](future<Status> future_status) {
```
dpcollins-google, line 390: remove

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +182,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+          Status status = future_status.get();
+          if (!status.ok()) {
+            return make_ready_future(
+                StatusOr<std::shared_ptr<UnderlyingStream>>(status));
+          }
+          std::lock_guard<std::mutex> lk{mu_};
```
dpcollins-google: remove

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -48,9 +48,10 @@ using AsyncStreamFactory = std::function<
  * from AsyncStreamFactory.
  */
 template <typename RequestType, typename ResponseType>
-using StreamInitializer = std::function<future<StatusOr<
-    std::unique_ptr<AsyncStreamingReadWriteRpc<RequestType, ResponseType>>>>(
-    std::unique_ptr<AsyncStreamingReadWriteRpc<RequestType, ResponseType>>)>;
+using StreamInitializer = std::function<future<StatusOr<std::shared_ptr<
```
dpcollins-google, line 51: revert this. This should still take and return the unique_ptr

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +182,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+
+    auto initializer_future =
+        start_future_result.then([this, stream](future<Status> future_status) {
+          Status status = future_status.get();
+          if (!status.ok()) {
+            return make_ready_future(
+                StatusOr<std::shared_ptr<UnderlyingStream>>(status));
+          }
+          std::lock_guard<std::mutex> lk{mu_};
+          return initializer_(stream);
```
dpcollins-google, line 396: initializer_(std::move(*stream))

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +182,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+        });
+
+    initializer_future.then(
+        [this, stream, retry_policy,
```
dpcollins-google, line 400: remove capture of stream

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +182,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+    FinishRetryPromiseLockHeld(lk);
+    if (stream_state_ == State::kShutdown) return;
+    stream_state_ = State::kShutdown;
+
+    CompleteUnsatisfiedOpsLockHeld(status, lk);
+  }
+
+  void OnInitialize(
+      StatusOr<std::shared_ptr<UnderlyingStream>> start_initialize_response,
+      std::shared_ptr<UnderlyingStream> stream,
```
dpcollins-google, line 349: remove this argument, you don't need to call Finish(), that's the initializer's responsibility

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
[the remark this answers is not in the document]
coryan: > though it sounds like the weirdness of the double indirection via `std::shared_ptr<std::unique_ptr<StreamType>>` outweighs the reasoning behind keeping the member variable as an `std::unique_ptr`?

Yes.  I think the recommendation to prefer `std::unique_ptr<>` over `std::shared_ptr<>` is basically sound, but I do not see it as a suicide pact.  When the complexity of using `std::unique_ptr<>` gets too high I just shrug and switch to `std::shared_ptr<>`.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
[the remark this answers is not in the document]
coryan: Looking at my own code, I think you are right.  For some reason I recall that not working in the past.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
18suresha: added a test in which a permanent error occurred via `IsExhausted` returning `true`

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
```diff
@@ -0,0 +1,575 @@
+// Copyright 2020 Google LLC
```
coryan, line 1: nit: s/2020/2022/

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,266 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+          CompleteUnsatisfiedOps(Status(), lk);
+          root_future = root_future.then([stream](future<void>) {
+            return future<void>(stream->Finish());
+          });
+      }
+    }
+    return root_future;
+  }
+
+  void SetReadWriteFutures(std::unique_lock<std::mutex>& lk) {
```
coryan, line 290: Optional: I think I would have received the argument by value, and force the caller to reacquire the mutex if they needed to.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,266 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+    lk.unlock();
+    if (read_reinit_done) read_reinit_done->set_value();
+
+    if (write_reinit_done) write_reinit_done->set_value();
+    lk.lock();
+  }
+
+  void CompleteUnsatisfiedOps(Status status, std::unique_lock<std::mutex>& lk) {
+    SetReadWriteFutures(lk);
+    status_promise_.set_value(std::move(status));
```
coryan, line 307: This can (implicitly) call user-level code while holding a mutex, that easily leads to deadlocks.  Seems like you would want to `lk.unlock()` before calling `status_promise_.set_value()`?

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,266 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+  }
+
+  void FinishRetryPromise(std::unique_lock<std::mutex>& lk) {
```
coryan, line 310: Optional: here I would also received the `lk` argument by value.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
coryan: Sorry I missed that earlier.

coryan (changes requested):
I think this is very close.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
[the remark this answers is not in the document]
dpcollins-google: I think in particular, shared_ptrs to streams are a bit scary, since they have some fairly tight constraints on calling their methods (specifically, that there can be exactly one future outstanding to read/write at a time).

Do you think what is done here adds significant complexity over the gain from knowing we're the only owner? If not, I'd rather leave unique_ptrs in the interface and members.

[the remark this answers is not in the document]
coryan: AFAIK, the only thing preventing two outstanding `Read()` calls (or `Write()` calls) is a comment in the header:

https://github.com/googleapis/google-cloud-cpp/blob/faecca1324a6c42af2a4f5fdd5018c9ebb9bea73/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h#L122-L124

 How is the `std::unique_ptr<>` helping more or less with that?  And furthermore, how would this `std::shared_ptr<>` leave the confines of this class and become reachable from multiple points?

Admittedly the use of `std::shared_ptr<std::unique_ptr<>>` is also fairly constrained, but I will not understand the code or motivation by Monday.  Forget about 6 months or 2 years from now.

I think we have enough complexity to carry in our heads, cutting one level of indirection seems like a good thing to me.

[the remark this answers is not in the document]
dpcollins-google: In particular, the initializer is required to return ownership of the unique_ptr to the caller, at which point, if it has outstanding futures, it can no longer enforce their contracts, so it must not have any outstanding futures. 

If it's a shared_pointer, one could quite easily store a reference to the underlying stream after initialization is complete, and think that is a perfectly fine thing to do.

I don't feel very strongly about this, other than unique_ptr is the more constrained API that gives us greater confidence that we are the only valid owner, which shared_ptr doesn't let you reason about. I think from a code archaeology perspective, down the road when modifying this, if I see a shared_ptr member, I assume someone else must be potentially accessing it, which it is confusing that in this case that is not true. 

Given the above, has your opinion changed at all on this? If not, we can make this change.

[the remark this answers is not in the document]
coryan: Hmm... I think you are saying that `std::shared_ptr<std::unique_ptr<>>` can be reasoned about locally, as it is used in a single function, while changing the initializer to return `std::shared_ptr<>` forces us to reason about something across multiple functions.  Fair enough.  That deserves at least a comment?

Finally, can we remove those temporary futures.  Chaining `.then()` calls seems more readable (module the formatting robots doing a semi-decent job).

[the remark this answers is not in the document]
18suresha: yup, just pushed

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+  }
+
+ private:
+  using UnderlyingStream =
+      std::unique_ptr<AsyncStreamingReadWriteRpc<RequestType, ResponseType>>;
+
+  enum class State { kRetrying, kInitialized, kShutdown };
+
+  future<void> ConfigureShutdownOrder(future<void> root_future) {
+    {
```
coryan, line 256: nit: is this scope useful?  Seems to be basically the whole function.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+    absl::optional<promise<void>> read_reinit_done;
+    absl::optional<promise<void>> write_reinit_done;
+
+    {
+      read_reinit_done.swap(read_reinit_done_);
+      write_reinit_done.swap(write_reinit_done_);
+    }
+    lk.unlock();
+    if (read_reinit_done) read_reinit_done->set_value();
+
```
coryan, line 300: nit remove blank line.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -181,6 +181,274 @@ class ResumableAsyncStreamingReadWriteRpc {
...
+  void OnInitialize(StatusOr<UnderlyingStream> start_initialize_response,
+                    std::shared_ptr<RetryPolicy> retry_policy,
+                    std::shared_ptr<BackoffPolicy> backoff_policy) {
+    if (!start_initialize_response.ok()) {
+      AttemptRetry(std::move(start_initialize_response.status()), retry_policy,
+                   backoff_policy);
+      return;
+    }
+    auto stream = std::move(*start_initialize_response);
+    {
```
coryan, line 357: nit: this scope does not seem useful.

coryan (approved):
A couple of nits that I trust you will clean up.  Enjoy the weekend.

Issue #3317, opened 2020-01-14
Set the `ClientContext` authority to support PSC and VPC-SC


coryan:
Originally reported for Cloud Bigtable as follows:

----
Using Cloud Bigtable with a user-provided JSON Web Token (JWT) for authentication **and** aiming to a custom endpoint does not work.

Reportedly one gets:

```
Error accessing Cloud Bigtable: Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.
```

To fix it, one must call `context->set_authority(<custom-endpoint-here>);`

----

But it is a more general problem.  When using custom endpoints, such as those defined by [PSC] (Private Service Connect) or [VPC-SC] (Virtual Private Cloud - Service Controls) we need to specify the `authority` in each gRPC request. This is the HTTP/2 analog of a `Host` header, and allows applications to access a service while using a custom endpoint. For example, the target endpoint could be `pubsub-custom.p.googleapis.com`, but the authority field remains `pubsub.googleapis.com`.

[PSC]: https://cloud.google.com/vpc/docs/private-service-connect
[VPC-SC]: https://cloud.google.com/vpc-service-controls

This will require several changes:

- [x] We need a new option (I am picking `AuthorityOption`), which defaults to the *default* endpoint for a library, for example, for Pub/Sub it defaults to `pubsub.googleapis.com` regardless of what `EndpointOption` and/or any environment variable is set to.
- [x] We need to change the generator to use this new option when the metadata is decorated.
- [x] We need to change Pub/Sub to use the common functions to initialize the default `Options`.
- [x] We need to change Pub/Sub to use this new option in its metadata decorators
- [x] We need to change Bigtable to use this new option in any metadata decorators
- [x] Storage+gRPC metadata decorator is generated, it should require no changes
- [x] Storage+REST should use this option to set the `Host:` header, with some suitable default if not set

PR #8453, opened 2022-02-25
feat: better support for PSC and VPC-SC

coryan:
When using [PSC] (Private Service Connect) or [VPC-SC] (Virtual Private
Cloud - Service Controls) we need to specify the `authority` in each
gRPC request. This is the HTTP/2 analog of a `Host` header, and allows
applications to access a service while using a custom endpoint. For
example, the target endpoint could be `pubsub-custom.p.googleapis.com`,
but the authority field remains `pubsub.googleapis.com`.

[PSC]: https://cloud.google.com/vpc/docs/private-service-connect
[VPC-SC]: https://cloud.google.com/vpc-service-controls

Part of the work for #3317

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8453)


dbolduc (approved):
Thanks for figuring this out. The changes LGTM.

I was confused by the scope of the PR (generated metadata decorators + spanner) but I see that you have a handle on the remaining work in the issue description.

PR #8456, opened 2022-02-25
feat(pubsub): support x-goog-user-project

coryan:
Change the metadata decorators to inject `x-goog-user-project` from the
current option span.

For the admin APIs, use the per-call options to configure a span in the
`*Client` class.  For `pubsub::Publisher` and `pubsub::Subscriber` it is
less clear which call "triggers" an RPC, and far less clear how to merge
options from multiple RPCs.  I have decided to use the per-connection
options in this case. We can always add more fine-grained options later,
but this takes care of the common case.

Part of the work for #8284 and will help with #3317

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8456)

### DIFF
--- a/google/cloud/pubsub/internal/publisher_metadata_test.cc
+++ b/google/cloud/pubsub/internal/publisher_metadata_test.cc
@@ -29,6 +29,10 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace {
 
 using ::google::cloud::testing_util::ValidateMetadataFixture;
+using ::testing::_;
+using ::testing::Contains;
+using ::testing::Not;
+using ::testing::Pair;
 
 class PublisherMetadataTest : public ::testing::Test {
  protected:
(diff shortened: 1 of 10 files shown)


google/cloud/pubsub/internal/publisher_metadata_test.cc
```diff
@@ -38,6 +42,22 @@ class PublisherMetadataTest : public ::testing::Test {
...
+
+  void ValidateTestUserProject(grpc::ClientContext& context) {
+    auto md = validate_metadata_fixture_.GetMetadata(context);
+    EXPECT_THAT(md, Contains(Pair("x-goog-user-project", "test-project")));
+  }
+
+  static Options TestOptions(std::string const& user_project) {
+    return user_project.empty()
+               ? Options{}
+               : Options{}.set<UserProjectOption>("test-project");
```
devbww, line 58: Did you mean to say `set<UserProjectOption>(user_project)` in these?

coryan: Yes, thanks for catching that.

devbww (approved):

PR #8457, opened 2022-02-25
doc: expanding metadata fields in storage_get_metadata sample

sydney-munro:
b/153524869

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8457)

### DIFF
--- a/google/cloud/storage/examples/storage_object_samples.cc
+++ b/google/cloud/storage/examples/storage_object_samples.cc
@@ -259,7 +259,18 @@ void GetObjectMetadata(google::cloud::storage::Client client,
 
     std::cout << "The metadata for object " << object_metadata->name()
               << " in bucket " << object_metadata->bucket() << " is "
-              << *object_metadata << "\n";
+              << "\n CacheControl : " << object_metadata->cache_control()
+              << "\n ComponentCount : " << object_metadata->component_count()
+              << "\n ContentDisposition : "
+              << object_metadata->content_disposition()
+              << "\n ContentEncoding : " << object_metadata->content_encoding()
+              << "\n ContentLanguage : " << object_metadata->content_language()
+              << "\n ContentType : " << object_metadata->content_type()
+              << "\n Crc32c : " << object_metadata->crc32c()
+              << "\n Generation : " << object_metadata->generation()
+              << "\n KmsKeyName : " << object_metadata->kms_key_name()
+              << "\n Md5Hash : " << object_metadata->md5_hash()
+              << "\n MediaLink : " << object_metadata->media_link();
   }
   //! [get object metadata] [END storage_get_metadata]
   (std::move(client), argv.at(0), argv.at(1));


google/cloud/storage/examples/storage_object_samples.cc
```diff
@@ -259,7 +259,18 @@ void GetObjectMetadata(google::cloud::storage::Client client,
 
     std::cout << "The metadata for object " << object_metadata->name()
               << " in bucket " << object_metadata->bucket() << " is "
-              << *object_metadata << "\n";
+              << "\n CacheControl : " << object_metadata->cache_control()
```
coryan, line 262: nit: if `CacheControl` goes into the next line then the previous `" is "` should be `" is:"` ?

Also, `CacheControl` is not the name of the field in the JSON API (where it is `cacheControl`), in the HTTP headers (`Cache-Control`), nor in the C++ API (`cache_control()`).  Seems sad to introduce a name that matches none of the other fields the application developer might be familiar with?

google/cloud/storage/examples/storage_object_samples.cc
```diff
@@ -259,7 +259,18 @@ void GetObjectMetadata(google::cloud::storage::Client client,
...
+              << "\n ContentDisposition : "
+              << object_metadata->content_disposition()
+              << "\n ContentEncoding : " << object_metadata->content_encoding()
+              << "\n ContentLanguage : " << object_metadata->content_language()
+              << "\n ContentType : " << object_metadata->content_type()
+              << "\n Crc32c : " << object_metadata->crc32c()
+              << "\n Generation : " << object_metadata->generation()
+              << "\n KmsKeyName : " << object_metadata->kms_key_name()
+              << "\n Md5Hash : " << object_metadata->md5_hash()
+              << "\n MediaLink : " << object_metadata->media_link();
```
coryan, line 273: maybe add a `<< "\n"` so this program outputs a newline at the end?

google/cloud/storage/examples/storage_object_samples.cc
coryan, line 262: The message is now technically incorrect, this is a selection of the metadata, not the full metadata, it is missing (for example) the storage class and the custom `metadata()`.   Maybe say "Selected metadata attributes for object "?

coryan (changes requested):
I remain unconvinced, if we think it would be beneficial for all languages to output the same things, then so be it.

Seems to me this is missing a number of attributes, the output when I run this example (at `main`) is below.  The new version is missing all the timestamps (though the old format is terrible, and also inconsistent!), the storage class, any custom metadata, and so on.

```
bazel run //google/cloud/storage/examples:storage_object_samples -- get-object-metadata coryan-test-bucket hello.txt
[snip snip snip]
Executing tests from //google/cloud/storage/examples:storage_object_samples
-----------------------------------------------------------------------------
The metadata for object hello.txt in bucket coryan-test-bucket is ObjectMetadata={name=hello.txt, acl=[], bucket=coryan-test-bucket, cache_control=, component_count=0, content_disposition=, content_encoding=, content_language=, content_type=, crc32c=NT3Yvg==, etag=CMyY9Ia9i/ICEAE=, event_based_hold=false, generation=1627671555279948, id=coryan-test-bucket/hello.txt/1627671555279948, kind=storage#object, kms_key_name=, md5_hash=sZRqySSS0jR8YjW00mERhA==, media_link=https://storage.googleapis.com/download/storage/v1/b/coryan-test-bucket/o/hello.txt?generation=1627671555279948&alt=media, , metageneration=1, name=hello.txt, retention_expiration_time=1970-01-01T00:00:00Z, self_link=https://www.googleapis.com/storage/v1/b/coryan-test-bucket/o/hello.txt, size=6, storage_class=MULTI_REGIONAL, temporary_hold=false, time_created=1627671555391000000, time_deleted=0, time_storage_class_updated=1627671555391000000, updated=1627671555391000000}

```

PR #8461, opened 2022-02-26
impl(rest): add AppendPath methods

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8461)

### DIFF
--- a/google/cloud/internal/rest_request.cc
+++ b/google/cloud/internal/rest_request.cc
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 #include "google/cloud/internal/rest_request.h"
+#include "google/cloud/internal/absl_str_cat_quiet.h"
 #include "google/cloud/log.h"
+#include "absl/strings/strip.h"
 #include <cctype>
 
 namespace google {
@@ -35,6 +37,18 @@ RestRequest::RestRequest(std::string path, HttpHeaders headers,
 
 RestRequest& RestRequest::SetPath(std::string path) & {
   path_ = std::move(path);
+  return *this;
+}
+
+RestRequest& RestRequest::AppendPath(std::string path) & {
+  if (path_.empty()) return SetPath(std::move(path));
+  if (path_.back() != '/' && path.front() != '/') {
+    path_ = absl::StrCat(path_, "/", path);
+  } else if (path_.back() == '/' && path.front() == '/') {
+    path_ = absl::StrCat(path_, absl::StripPrefix(path, "/"));
+  } else {
+    path_ = absl::StrCat(path_, path);
+  }
   return *this;
 }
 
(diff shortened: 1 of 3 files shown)


scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8461)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @devjgm)

___
*[google/cloud/internal/rest_request.h, line 55 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8461#-MwnmLyPn_Np7rrvYr7D-r1-55:-Mws12jt3ae1jvjFghO1:b9288qt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0a8704600bbb42c72745412507321cc5ab805044/google/cloud/internal/rest_request.h#L55)):*

I now see that many functions in this file use the `inline` keyword. Why is that? Unless there's a functional reason that I'm not aware of, I think we should remove them for consistency w/ existing style. It's fine to do in a separate PR, though.
</blockquote></details>

I have no idea why I added those `inline`s. I blame either too much or too little caffeine that day.
I think it's a uncontroversial change that doesn't need a separate PR.
___
*[google/cloud/internal/rest_request.cc, line 43 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8461#-MwnmLyPn_Np7rrvYr7C-r1-43:-Mws1NgmAGoi4fD6o3Y-:bmyyhwb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0a8704600bbb42c72745412507321cc5ab805044/google/cloud/internal/rest_request.cc#L43)):*

"Append" operations often can't move from the argument, because their data usually gets copied into some other existing buffer, so I'd expect to see the argument passed by const-ref rather than value. In this implementation I do see one std::move, but the other code paths are all copies, and the move path looks like an implementation detail. I suspect we want the parameter to be a const-ref.

What do you think?
</blockquote></details>

So far while using this method I've mostly passed it prvalues. I would prefer to keep it by-value for now.
___
*[google/cloud/internal/rest_request.cc, line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8461#-MwnmLyPn_Np7rrvYr7C-r1-44:-Mws1_F5A33FKMojY3Sb:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0a8704600bbb42c72745412507321cc5ab805044/google/cloud/internal/rest_request.cc#L44)):*

Consider something like:

```cc
path_ = absl::StrCat(absl::StripSuffix(path_, "/"), "/", absl::StripPrefix(path, "/"));
return *this;
```
Most of those are string_view operations, so it should be pretty efficient.
</blockquote></details>

Done.

devjgm (approved):
Thanks. LGTM.

> So far while using this method I've mostly passed it prvalues. I would prefer to keep it by-value for now.

OK. Leaving it as you have it is OK if you want.

I don't think passing in prvalues affects the decision, because whether the argument type is a value or const-ref, that prvalue will be used to move-construct a new std::string. What matters is how the implementation uses the argument. As a rule of thumb, if the implementation needs to own its own copy of the data, pass by value. Otherwise, pass by reference. "Append" methods usually do not need to own the passed data because they copy the data into some other existing buffer (like std::string::append).

PR #8465, opened 2022-02-27
ci: update and check API baseline for all GA libraries

coryan:
Use the list of GA libraries from the top-level `BUILD.bazel` file to
determine which libraries should have an API baseline.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8465)


ci/cloudbuild/builds/check-api.sh
```diff
@@ -51,16 +59,11 @@ function dump_abi() {
...
-  "google_cloud_cpp_grpc_utils"
-  "google_cloud_cpp_iam"
-  "google_cloud_cpp_spanner"
-  "google_cloud_cpp_storage"
-  "google_cloud_cpp_pubsub"
+feature_list+=(
+  "common"
+  "grpc_utils"
 )
+mapfile -t libraries < <(for f in "${feature_list[@]}"; do echo "google_cloud_cpp_${f}"; done)
```
devjgm, line 66: No change necessary, but it looks like this should work similar to what you did above using `printf`, like `... <(printf "google_cloud_cpp_%s\n" "${feature_list[@]}")`

devjgm (approved):

devjgm (approved):

Issue #8459, opened 2022-02-25
Mark some RPCs as "experimental"


coryan:
This is motivated by bi-dir streaming RPCs, where we are unconvinced as to the final form of the RPC, but it may be useful as new RPCs are introduced too.

It would be nice to mark some RPCs are "experimental", such that their signature changes from:

```cc
class FooServiceClient {
public:
  StatusOr<Response> DoThing(Request const&, Options = {});
```

to 

```cc
class FooServiceClient {
public:
  StatusOr<Response> DoThing(google::cloud::Experimental, Request const&, Options = {});
```

that extra `google::cloud::Experimental` parameter would serve as a warning that the API may change.

dbolduc:
@coryan is there more to do here?

e.g. we could generate an experimental API for streaming writes

coryan:
No, closing.

PR #8471, opened 2022-02-28
feat(generator): all bidir streams are experimental

coryan:
Change the bidirectional streaming RPCs to take a `ExperimentalTag`
parameter. We are not certain if these APIs are ergonomic enough, and
may want to change them for something easier to use.  Until we make
these decisions, it seems sad to keep all the other RPCs from reaching
"GA" status. In the case of BigQuery, the overall library is GA, we
don't want to revert the GA status based on a single RPC.

This is not a breaking change, as none of the libraries affected by the
change are GA.  It does reset the clock for these libraries until T+28d.

Related to #8459

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8471)

### DIFF
--- a/generator/internal/client_generator.cc
+++ b/generator/internal/client_generator.cc
@@ -71,10 +71,12 @@ Status ClientGenerator::GenerateHeader() {
 
   // includes
   HeaderPrint("\n");
-  HeaderLocalIncludes({vars("connection_header_path"), "google/cloud/future.h",
-                       "google/cloud/options.h",
-                       "google/cloud/polling_policy.h",
-                       "google/cloud/status_or.h", "google/cloud/version.h"});
+  HeaderLocalIncludes(
+      {vars("connection_header_path"),
+       HasBidirStreamingMethod() ? "google/cloud/experimental_tag.h" : "",
+       "google/cloud/future.h", "google/cloud/options.h",
+       "google/cloud/polling_policy.h", "google/cloud/status_or.h",
+       "google/cloud/version.h"});
   if (get_iam_policy_extension_ && set_iam_policy_extension_) {
     HeaderLocalIncludes({"google/cloud/iam_updater.h"});
   }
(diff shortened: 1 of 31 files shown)


generator/internal/client_generator.cc
```diff
@@ -121,7 +123,7 @@ Status ClientGenerator::GenerateHeader() {
   std::unique_ptr<::google::cloud::AsyncStreamingReadWriteRpc<
       $request_type$,
       $response_type$>>
-  Async$method_name$(Options opts = {});
+  Async$method_name$(ExperimentalTag, Options opts = {});
```
dbolduc, line 126: So we are signing the user up for at least one break. Is that a good thing? We could avoid it by doing like:

```
AsyncFoo(Options opts = {}, ExperimentalTag tag = {});
```

(but then they are less likely to notice the tag)

google/cloud/experimental_tag.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_EXPERIMENTAL_TAG_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_EXPERIMENTAL_TAG_H
+
+#include "google/cloud/version.h"
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+struct ExperimentalTag {};
```
dbolduc, line 24: can we say what this thing means?

"If this tag is present as a function parameter, the function is experimental. Experimental functions are subject to change without notice, including removal."

generator/internal/client_generator.cc
coryan: I want this tag to be a *required* parameter, so folks cannot accidentally use these APIs without knowing that they are experimental (I guess they could say `using GA = ExperimentalTag;` but that would be just self-defeating).

To avoid the break we can keep an overload:

```cc
[[deprecated("use non-experimental version")]] Blah AsyncFoo(ExperimentalTag, Options opts = {}) {
 return AsyncFoo(std::move(opts));
 }

Blah AsyncFoo(Options opts);
```

dbolduc: Thanks for clarifying.

dbolduc (approved):

google/cloud/experimental_tag.h
devbww: s/, they/. They/

PR #8477, opened 2022-03-01
fix: use CMAKE_CURRENT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

martinzink:
Recent commits broke the possibility to use the project with CMake as an ExternalProject(FetchContent in my case).
In https://github.com/googleapis/google-cloud-cpp/commit/f4c898f47838dffade2f0cdf4a182e0347a34612 and https://github.com/googleapis/google-cloud-cpp/commit/14a0ad2bf1574a87277778b2b146a80672872ccb in the root CMakeLists.txt the folder structure is checked relative to CMAKE_SOURCE_DIR which works if the project is standalone, however if used as an external project CMAKE_SOURCE_DIR refers to the topmost project's source dir.

I suggest we change these to CMAKE_CURRENT_SOURCE_DIR since `add_subdirectory` already works relative to that.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8477)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -300,9 +300,9 @@ function (google_cloud_cpp_enable_features)
                 add_subdirectory(google/cloud/storage)
             endif ()
         else ()
-            if (NOT IS_DIRECTORY "${CMAKE_SOURCE_DIR}/google/cloud/${feature}"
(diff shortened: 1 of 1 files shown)


coryan:
Thanks for the PR!  It looks good to me, let's get the CI started.

coryan (approved):

coryan:
Thanks again for the PR!

PR #8478, opened 2022-03-01
feat(pubsublite): partition publisher implementation

18suresha:
This PR contains the core publisher tests for when we have an initialized publisher.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8478)

PR #8504, opened 2022-03-04
clean(pubsublite): add base interfaces and promise wrapper

18suresha:
this PR contains changes and helpers that were in https://github.com/googleapis/google-cloud-cpp/pull/8478 for `pubsublite_internal::ResumableAsyncStreamingReadWriteRpc`

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8504)


dpcollins-google (approved):

google/cloud/pubsublite/internal/base_interface.h
```diff
@@ -0,0 +1,46 @@
...
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace pubsublite_internal {
+
+// A non-copyable, non-movable base interface. Using this as a base class for
+// your interface reduces its boilerplate, while complying with the style guide
+// that `If a base class clearly isn't copyable or movable, derived classes
+// naturally won't be either.`
+class BaseInterface {
```
coryan, line 29: This class has two code smells.  Anything called `Base*` and/or `*Interface` is suspect.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -358,39 +352,35 @@ class ResumableAsyncStreamingReadWriteRpcImpl
...
-        root_future =
-            root_future.then(ChainFuture(in_progress_read_->get_future()));
-      }
-      if (in_progress_write_.has_value()) {
-        root_future =
-            root_future.then(ChainFuture(in_progress_write_->get_future()));
-      }
-
-      root_future.then([this](future<void>) { FinishOnStreamFail(); });
+    AsyncRoot root;
```
coryan, line 355: Please use a separate PR for these `AsyncRoot` refactorings.

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
```diff
@@ -253,7 +236,7 @@ class ResumableAsyncStreamingReadWriteRpcImpl
         });
   }
 
-  future<bool> Write(RequestType const& r, grpc::WriteOptions o) override {
+  future<bool> Write(RequestType const& r) override {
```
coryan, line 239: This is uncontroversial, but needs to be in a separate `cleanup(pubsublite): `-type PR.

google/cloud/pubsublite/internal/lifecycle_interface.h
```diff
@@ -0,0 +1,62 @@
...
+#include "google/cloud/pubsublite/internal/base_interface.h"
+#include "google/cloud/future.h"
+#include "google/cloud/status.h"
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace pubsublite_internal {
+
+class LifecycleInterface : public BaseInterface {
```
coryan, line 27: Another code smell.  Why is this not called `Lifecycle`?  Why do we need this base class at all?  This smells like a base class to reuse implementation, not a real thing we want to model.

coryan (changes requested):

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
18suresha: https://github.com/googleapis/google-cloud-cpp/pull/8506

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h
18suresha: https://github.com/googleapis/google-cloud-cpp/pull/8505

google/cloud/pubsublite/internal/lifecycle_interface.h
18suresha: Down the line, there will be classes with their own `Start` and `Shutdown` methods (inheriting from `LifecycleInterface`). Those methods will respectively call `Start` and `Shutdown` of the class' member variables (ex. `ResumableAsyncStreamingReadWriteRpc`) who also inherit from `LifecycleInterface`.
An example of this interface's usage is here https://github.com/googleapis/google-cloud-cpp/blob/ede01f6a2372f0f7e29d2cc6449e5de5ca0d13de/google/cloud/pubsublite/internal/lifecycle_helper.h 
Modeling this through `LifecycleInterface` allows us to have helpers like the above that can be reused.

coryan: I have separate objections to this:

1) `FooInterface` is almost always a bad name.  `Foo` is the right name, that is what the consumer of the class will operate and think about.  `FooImpl` is the concrete (or one of the concrete) implementations.

2) What are these things?  "Lifecycle" is not a meaningful name.  Are they "streams"?  Background workers? Long running tasks?  What?  Are you naming these for what you can do on them or for what they represent?

3) The fact that they have common functions is  not enough.  If we are going to use base classes they need to represent a thing.  An interesting entity that you can use in different contexts or that has many different implementations. As a guideline (but not a hard and fast rule) read on Liskov substitution.

`BaseInterface` is even worse.  It saves some typing (maybe), but it is not a usable class or interface, it is an implementation detail.  Public inheritance is supposed to represent IS-A relationships, this is not a case of that.  If you feel you need to prohibit copying and assignment then call `BaseInterface` something meaningful, like`DisableCopy`, and consider using private inheritance.  Or use CRTP to make it clear what this does:

```cc
template <typename T>
class DisableCopy {
 public:
   DisableCopy(DisableCopy&&) = delete;
   DisableCopy& operator=(DisableCopy&&) = delete;
   DisableCopy(DisableCopy const&) = delete;
   DisableCopy& operator=(DisableCopy const&) = delete;
};

class Foo : private DisableCopy<Foo> {
};
```

I still would think that writing such a class is a waste of time, but I least I don't find such use confusing.  Yes, I know the style guide spouts ridiculous opinions about private inheritance... I.Don't.Care.

dpcollins-google: > FooInterface is almost always a bad name. Foo is the right name, that is what the consumer of the class will operate and think about. FooImpl is the concrete (or one of the concrete) implementations.

Sure. I'm fine calling this "HasLifecycle" or "ObjectLifecycle". The point is to allow for common handling when you have multiple dependencies that start up/shut down in the same way. Java calls this "Service" but that's not a more descriptive name. 

> What are these things? "Lifecycle" is not a meaningful name. Are they "streams"? Background workers? Long running tasks? What? Are you naming these for what you can do on them or for what they represent?

Anything that needs asynchronous startup and shutdown outside of just a destructor call. Most of the time these will be things that wrap streams in an abstraction layer.

> BaseInterface is even worse. It saves some typing (maybe) but it is not a usable class or interface

It saves quite a bit of typing which is the main purpose, but it is usable as an interface as well: it is a class whose only virtual method is a destructor that nonetheless is non-slicable. When you have token types with no functions other than their destructor, they are quite easy to slice accidentally. You can choose to delete the virtual destructor instead, but then you need to explicitly override it in all inheritors.

If you really hate this we can remove it, I just don't quite see why it's so objectionable.

coryan: > Sure. I'm fine calling this "HasLifecycle" or "ObjectLifecycle". The point is to allow for common handling when you have multiple dependencies that start up/shut down in the same way. Java calls this "Service" but that's not a more descriptive name.

If there is no clear thing they represent that is a code smell.  I remain unconvinced.  Maybe later if it looks like it solves some problem we can not solve with some less coupling.

> If you really hate this we can remove it, I just don't quite see why it's so objectionable.

Please do remove it.

Actually, this PR seems empty after you remove those things.  Let me know if I missed anything.

18suresha: To try to give more reasoning behind `LifecyleInterface` (will change name if the class goes through), it represents objects that have a `Start` behavior and a `Shutdown` behavior. The reason for the abstraction is that objects at multiple layers of the stack are 'lifecycle' objects.
For example, a multipartition publisher would use a single partition publisher which would use a resumable stream. 'would use' akin to a 'has a' relationship.
With the above 'has a' relationship, we know that for each `foo` that has `bar[]`, `foo.Start()` would call `bar.forEach(obj => obj.Start())`.
Having the `LifecycleInterface` abstraction allows us to uniformly handle these dependencies/behaviors.

coryan: So if you have a multi-partition publisher, presumably it already holds all the single-partition objects (whatever their type is) in some collection?  Can you just use `std::for_each()` or `std::transform()` in that collection?  Or if that is too cumbersome, a generic function like:

```cc
template <typename It>
auto StartAll(It begin, It end) {
  std::vector<future<Status>> started;
  std::transform(begin, end, std::back_inserter(started), [](auto s) { return s.Start(); });
  return started;
}
```

18suresha: it would look like this
https://github.com/googleapis/google-cloud-cpp/blob/2df67980805371b26d79d889aede792da5b46f91/google/cloud/pubsublite/internal/lifecycle_helper.h
because not only do we need to `Start` everything, we also need to keep track of failures across any of the dependencies and satisfy the correct futures and modify state accordingly

coryan: When I made my comments I had read that code already.  I remain unconvinced.  This is getting boring.  Find a name and description of what `Lifecycle` represents or don't introduce a base class.  Classes should represent things, not just convenient groupings of functions.

dpcollins-google: Hello,

I can think of a couple of workarounds for your objections to LifecycleInterface, please let me know which you would prefer. We can delete BaseInterface if required as it primarily serves to reduce the amount of typing required.

1) We rename it to `Service`, as it represents a logical service which can be started, stopped and can fail. This would be similar to the guava (java) library of the same name https://guava.dev/releases/19.0/api/docs/com/google/common/util/concurrent/Service.html which serves a similar purpose of putting a usable name on object lifecycle operations so they can be manipulated in the abstract. We could also give it whatever other name you would find acceptable if you have a suggestion.

2) We can remove LifecycleInterface and instead rely on all interfaces having the same "shape" for use in common lifecycle operations. This would make LifecycleHelper significantly more complex, as it would then have to contain, instead of a vector of interface references, a vector of c-style struct interfaces with functions, which can be constructed from lambdas in the constructor:

```
class LifecycleHelper {
 public:
  template <class LifecycleT>
  void AddDependency(LifecycleT& lifecycle) {
    std::lock_guard<std::mutex> g{mu_};
    dependencies_.push_back([&]{ return lifecycle.Start(); }, [&]{ return lifecycle.Shutdown(); });
  }
  
  future<Status> Start() {
    std::vector<future<Status>> start_futures;
    {
    std::lock_guard<std::mutex> g{mu_};
    for (LifecycleMethods& dependency : dependencies_) {
      start_futures.push_back(dependency.start());
    }
    }
    for (future<Status>& start_future: start_futures) {
      start_future.then([this](future<Status> status_future) {
        Status s = status_future.get();
        if (!s.ok()) Abort(std::move(s));
      });
    }
    std::lock_guard<std::mutex> g{mu_};
    return status_promise_->get_future();
  }
            
    ... other methods ...
 private:
  struct LifecycleMethods {
    std::function<future<Status>> start;
    std::function<future<void>> shutdown;
  };
  std::mutex mu_;
  std::vector<LifecycleMethods> dependencies_;
  ... other members ...
};
```

3) We can inline the identical lifecycle behavior into every class which uses it. I would really prefer to avoid this, as it is quite a lot of code duplication and is quite error prone to implement multiple times (probably ~10 over the course of implementing this library) 

Please let us know which approach you would prefer.

coryan: `Service` sounds much better than `Lifecycle`.  Not super crisp, but with a description at least makes it something that I can follow and understand what sort of entities are involved.  I would gladly go with that.

Optional: are we sure we need inheritance and we could not use external polymorphism for these things?  Could `LifecycleHelper` (which I guess would be renamed `ServiceHelper` or better yet `ServiceComposite`) be a template on the type of its "services"?

dpcollins-google: We'll change Lifecycle -> Service. As I mentioned in option 2, you need some way to address all the services, which may have different types. You can do other things like having a pack of argument types then taking dependencies of those types, but that prevents something we would like to do for the next layer of code, a publisher which wraps one partition publisher for each partition, from adding dependencies if new ones come about (i.e. if the number of partitions increases). It also makes the fact that a given object implements the `Service` shape much more implicit.

18suresha: just pushed with these updates

google/cloud/pubsublite/internal/service.h
```diff
@@ -0,0 +1,76 @@
...
+#include "google/cloud/future.h"
+#include "google/cloud/status.h"
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace pubsublite_internal {
+
+/**
+ * A `Service` object is one that has a `Start` method that must be called
```
coryan, line 27: Please describe what the class represents, not the functions it has, I can read those.  Comments are for humans, they should explain the intent.  Maybe something like this?

```cc
/**
  * A `Service` is anything that can start, run for a while, possibly with an error, and then can shutdown.
  *
  * Examples include resumable streaming RPCs of different types, and compositions of
  * these streaming RPCs.
  */
```

google/cloud/pubsublite/internal/service.h
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Starts the lifecycle of the object.
+   *
+   * The future returned by this function is satisfied when the object
+   * encounters an error during one of its operations or when `Shutdown` is
+   * called on it. The value the future is satisfied with details what happened
+   * to the object, i.e., `Shutdown` was called, encountered a permanent error,
+   * etc.
+   *
+   * Must be called before any other method and may only be called once.
```
coryan, line 52: The only other method is `Shutdown()`?  Does this requirement apply to member functions added in derived classes?  That seems odd (usually the base class / interface cannot impose restrictions on how the children behave).  If this is the case, then please say that this also applies to the children classes.

coryan (changes requested):

google/cloud/pubsublite/internal/service.h
18suresha: yup applies to derived classes, added a note.

coryan (approved):

Issue #3398, opened 2020-02-22
Record some recent performance results in a GitHub issue.


coryan:
It would be nice to have performance results recorded in a GitHub issue so folks know what to expect. Maybe something like:

- Download times (effective bandwidth) for different read sizes.
- Download times (effective bandwidth) with MD5 enable vs. MD5 disabled.
- Upload times for different sizes in the `stream.write(...)` calls.
- Download times for regional buckets in the same region as the VM vs. a "remote" regional bucket.

coryan:
Details available internally at [go/coryan:gcs+grpc-2022-q1](https://goto.google.com/coryan:gcs+grpc-2022-q1)

PR #8520, opened 2022-03-09
test(storage): stream benchmark results

coryan:
I am starting to run benchmarks that take hours to complete.  Sometimes
it is desirable to capture partial results from these benchmarks to
debug the analysis scripts, or even run preliminary analysis. With this
change the throughput-vs-cpu benchmark streams its results as it runs,
instead of printing the results at the end of the test.

Motivated by #3398

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8520)

### DIFF
--- a/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
+++ b/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
@@ -23,6 +23,7 @@
 #include "google/cloud/internal/format_time_point.h"
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/internal/random.h"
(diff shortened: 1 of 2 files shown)


google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -200,15 +201,19 @@ int main(int argc, char* argv[]) {
     return 1;
   }
 
+  std::mutex mu;
```
devbww, line 204: Nit: Perhaps something like `// serialize output` might clarify this (assuming I understand it).

devbww (approved):

PR #8522, opened 2022-03-10
docs: Document service endpoint environment variables

### LINKED ISSUE #8320: Document service endpoint environment variables
Each generated service comes with its own environment variable to set the endpoint. We do not document this anywhere. All endpoint environment variables for a library should be documented in the `google/cloud/${library}/README.md`

dbolduc:
Fixes #8320

All of the changes in the first commit were the result of running a script, which Googlers can find [here](https://paste.googleplex.com/5867102910021632). Reviewers should not hesitate to point out nits, because it will be easy to regenerate these changes.

I also decided to update the scaffolding so we (hopefully) don't forget to document these variables in future libraries.

Note that Bigtable, Pub/Sub, Spanner, and Storage already document their endpoint variables.


coryan (approved):
Did you find any duplicates?

PR #8529, opened 2022-03-10
feat(pubsublite): service composite implementation

18suresha:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8529)

### DIFF
--- a/google/cloud/pubsublite/internal/service_composite.h
+++ b/google/cloud/pubsublite/internal/service_composite.h
@@ -1,0 +1,105 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_PUBSUBLITE_INTERNAL_SERVICE_COMPOSITE_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_PUBSUBLITE_INTERNAL_SERVICE_COMPOSITE_H
+
+#include "google/cloud/pubsublite/internal/futures.h"
+#include "google/cloud/pubsublite/internal/service.h"
+#include <mutex>
+#include <vector>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace pubsublite_internal {
+
+/**
+ * `ServiceComposite` wraps the lifetimes of one or more `Service`s into a
+ * single object. This enables a user to manage the lifetimes of several
+ * dependencies through a single object.
+ */
+class ServiceComposite : public Service {
+ public:
+  template <class... ServiceT>
+  explicit ServiceComposite(ServiceT&... dependencies)
+      : dependencies_{{std::ref(static_cast<Service&>(dependencies))...}} {}
+
+  future<Status> Start() override {
+    for (Service& dependency : dependencies_) {
+      dependency.Start().then([this](future<Status> status_future) {
+        Status s = status_future.get();
+        if (!s.ok()) Abort(std::move(s));
+      });
+    }
+    std::lock_guard<std::mutex> g{mu_};
+    return status_promise_->get_future();
+  }
+
+  /**
+   * @note Can be safely called more than once.
+   */
+  void Abort(Status s) {
+    assert(!s.ok());
+    promise<Status> start_promise{null_promise_t{}};
+    {
+      std::lock_guard<std::mutex> g{mu_};
+      if (!status_promise_) return;
+      start_promise = std::move(*status_promise_);
+      status_promise_.reset();
+      final_status_ = s;
+    }
+    start_promise.set_value(std::move(s));
+  }
+
+  /**
+   *
+   * @return a `Status` of `kOk` if and only if the current lifecycle is in the
+   * running phase
+   */
+  Status status() {
+    std::lock_guard<std::mutex> g{mu_};
+    return final_status_;
+  }
+
+  future<void> Shutdown() override {
+    {
+      std::lock_guard<std::mutex> g{mu_};
+      if (shutdown_) return make_ready_future();
+      shutdown_ = true;
+    }
+    Abort(Status{StatusCode::kAborted, "`Shutdown` called"});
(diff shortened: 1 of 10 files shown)


google/cloud/pubsublite/internal/service_composite.h
```diff
@@ -0,0 +1,105 @@
...
+    Abort(Status{StatusCode::kAborted, "`Shutdown` called"});
+    future<void> root_future = make_ready_future();
+    for (auto const& dependency : dependencies_) {
+      root_future = root_future.then(ChainFuture(dependency.get().Shutdown()));
+    }
+    return root_future;
+  }
+
+ private:
+  std::vector<std::reference_wrapper<Service>> const dependencies_;
```
coryan, line 91: I think this should be `std::vector<Service*>` the `std::reference_wrapper` is just hiding what you are really doing.

google/cloud/pubsublite/internal/service_composite.h
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  future<void> Shutdown() override {
+    {
+      std::lock_guard<std::mutex> g{mu_};
+      if (shutdown_) return make_ready_future();
+      shutdown_ = true;
+    }
+    Abort(Status{StatusCode::kAborted, "`Shutdown` called"});
+    future<void> root_future = make_ready_future();
```
coryan, line 83: Why serialize all the `Shutdown()` calls?  I think you want something like:

https://github.com/GoogleCloudPlatform/cpp-samples/blob/f9e1a7038909ab5d83f639a4931d13792e90c487/getting-started/gke/index_gcs.cc#L211-L243

And then do:

```cc
std::vector<future<void>> shutdown;
std::transform(dependencies_.begin(), dependencies_.end(), std::back_inserter(shutdown), [](auto s) { return s->Shutdown(); });
return when_all(std::move(shutdown));
```

coryan (changes requested):

google/cloud/pubsublite/internal/service_composite.h
```diff
@@ -0,0 +1,105 @@
...
+      start_promise = std::move(*status_promise_);
+      status_promise_.reset();
+      final_status_ = s;
+    }
+    start_promise.set_value(std::move(s));
+  }
+
+  /**
+   *
+   * @return a `Status` of `kOk` if and only if the current lifecycle is in the
```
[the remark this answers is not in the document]
18suresha: `Shutdown` for `Service` objects don't really have a success/failure result, only completion since it returns `future<void>`

dpcollins-google:
___
*[google/cloud/pubsublite/internal/service_composite.h, line 83 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8529#-Mxooz-8nSRe16VNwely-r2-83:-Mxp7PH--iAPGG0uTX8N:b6oaav3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/10eef81019fa03bf2d528f92711cb05b47075a49/google/cloud/pubsublite/internal/service_composite.h#L83)):*

Why serialize all the `Shutdown()` calls?  I think you want something like:

https://github.com/GoogleCloudPlatform/cpp-samples/blob/f9e1a7038909ab5d83f639a4931d13792e90c487/getting-started/gke/index_gcs.cc#L211-L243

And then do:

```cc
std::vector<future<void>> shutdown;
std::transform(dependencies_.begin(), dependencies_.end(), std::back_inserter(shutdown), [](auto s) { return s->Shutdown(); });
return when_all(std::move(shutdown));
```

</blockquote></details>

That's exactly what this is doing. It's only serializing the calls to make the futures- note that ChainFuture takes a future (i.e. already started) and adds it to the `then` callback.

Is there a "when_all" future helper? If so, we can use that instead.

google/cloud/pubsublite/internal/service_composite.h
18suresha: done

dpcollins-google:
___
*[google/cloud/pubsublite/internal/service_composite.h, line 61 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8529#-MxpbR9UBWskDJLAIyb3:-MxpbR9UBWskDJLAIyb4:bdrtyiq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3ed5b92d26d513a622c44aebc25cc6e497f71b9/google/cloud/pubsublite/internal/service_composite.h#L61)):*
> ```objc
>    * This will add a `Service` dependency for the current object to manage. It
>    * is only added if the current object hasn't been `Shutdown` yet.
>    * @param dependency
> ```

add a note that Start() must have already been called?



_Code quote:_
```objc
* @param dependency
```

dpcollins-google:
___
*[google/cloud/pubsublite/internal/service_composite_test.cc, line 34 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8529#-Mxpb_M_CpakzzbATZsX:-Mxpb_M_CpakzzbATZsY:b-dg8tca) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e3ed5b92d26d513a622c44aebc25cc6e497f71b9/google/cloud/pubsublite/internal/service_composite_test.cc#L34)):*
> ```cpp
> using ::google::cloud::pubsublite_testing::MockService;
> 
> TEST(ServiceTest, SingleDependencyGood) {
> ```

can you add a create destroy test for this as well?



_Code quote:_
```cpp
TEST(ServiceTest, SingleDependencyGood) {
```

dpcollins-google (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8529)* status: 0 of 10 files reviewed, 5 unresolved discussions (waiting on @18suresha and @coryan)

google/cloud/pubsublite/internal/service_composite.h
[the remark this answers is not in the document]
coryan: Right.  What I mean is what happens if you all the services finish running and the future returned by `Start()` are all satisfied with success?  Is there a reason that cannot happen?  Can you write it down?

google/cloud/pubsublite/internal/service_composite.h
```diff
@@ -0,0 +1,159 @@
...
+      if (status_.ok())
+        status_ = Status{StatusCode::kAborted, "`Shutdown` called"};
+    }
+    if (status_promise) status_promise->set_value(Status());
+
+    AsyncRoot root;
+    future<void> root_future = root.get_future();
+
+    std::lock_guard<std::mutex> g{mu_};
+    for (auto* dependency : dependencies_) {
```
coryan, line 137: I had a question here, but I think it was answered, I missed the effect of `ChainFuture()`.

coryan (approved):

google/cloud/pubsublite/internal/service_composite.h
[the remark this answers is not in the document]
18suresha: Oh I see. Well the services can only finish in two ways since they are `Service`s, i.e., their `Start` encounters an error or `Shutdown` is called on the higher level `ServiceComposite` object. If I understand your question correctly, the answer would be that it cannot happen. Added a comment.

PR #8530, opened 2022-03-10
feat(kms): support the EKM service

### LINKED ISSUE #8483: New `ekm_service.proto` for google/cloud/kms
Probably just needs a new entry in `generator_config.textproto`

coryan:
This is a new gRPC service in the same googleapis directory (`google/cloud/kms/v1`) so it goes into the same library.

Fixes #8483.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8530)


dbolduc:
1. I think we want a note somewhere in the CHANGELOG that this service has been added. Like we did in #8422 for bigquery
2. Can you add the new environment variable here: https://github.com/googleapis/google-cloud-cpp/blob/70e0a81b4095320037381bbcd6132724bcda25af/google/cloud/kms/doc/main.dox#L44-L46

This makes me think we may want/need a section in our howto doc for "adding new services to GA libraries". Even if it's just a two item checklist. What do you think?

dbolduc (approved):

PR #8531, opened 2022-03-10
feat(pubsublite): single partition publisher boilerplate and `Start`/`Shutdown` implementation

18suresha:
This PR implements the boilerplate for the publisher and the mocks used for later tests
Note that only the batching logic currently has tests to keep the LOC count below 1000. 

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8531)


coryan (approved):
An `ABSL_LOCKS_REQUIRED()` comment is now obsolete. Please fix and then merge.

I suspect that the inner `std::deque<>` in all those `std::deueue<std::dequeue<MessageWithPromise>>` could be a vector or a class with some functionality.  Finally, if you are going to be slicing collections, maybe we should be using `std::list<>` and its `splice()` member function.

> Reviewed 1 of 3 files at r6, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8531#-:-MyHbrqc12VzT-uaWdZB:b-mk21bz)* status: all files reviewed, 1 unresolved discussion (waiting on @18suresha)

___
*[google/cloud/pubsublite/internal/partition_publisher.cc, line 83 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8531#-MyHcOvr7N1yreTZG-H7:-MyHcOvr7N1yreTZG-H8:bfr2jnz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/df8bac7e6a4cd22fbc614282d3606296f894695b/google/cloud/pubsublite/internal/partition_publisher.cc#L83)):*
> ```cpp
> 
> std::deque<PartitionPublisher::MessageWithPromise>
> PartitionPublisher::UnbatchAll() {  // ABSL_LOCKS_REQUIRED(mu_)
> ```

I think the comment is now incorrect.
___
*[google/cloud/pubsublite/internal/partition_publisher.cc, line 107 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8531#-MyHfDNEAtH4YUU_CbHx:-MyHfDNF96xsC30gyTqX:bqpac03) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/df8bac7e6a4cd22fbc614282d3606296f894695b/google/cloud/pubsublite/internal/partition_publisher.cc#L107)):*
> ```cpp
> std::deque<std::deque<PartitionPublisher::MessageWithPromise>>
> PartitionPublisher::CreateBatches(std::deque<MessageWithPromise> messages,
>                                   BatchingOptions const& options) {
> ```

FYI, this function and its tests could have been in a separate PR.

PR #8532, opened 2022-03-11
feat(common): `g::c::future` support for coroutines

coryan:
Integrate `google::cloud::future<T>` with C++20 coroutines.  This
feature is optional, and only enabled if the compiler supports C++20.
Because some C++20 compilers disable coroutines by default, this feature
is disabled if `__cpp_impl_coroutine` is not defined, or it is 0.

### DIFF
--- a/google/cloud/internal/future_coroutines.h
+++ b/google/cloud/internal/future_coroutines.h
@@ -1,0 +1,131 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


google/cloud/internal/future_coroutines.h
```diff
@@ -0,0 +1,131 @@
...
+#if GOOGLE_CLOUD_CPP_CPP_VERSION >= 202002L && __cpp_impl_coroutine
+#include "google/cloud/future_generic.h"
+#include "google/cloud/future_void.h"
+#include "google/cloud/version.h"
+#include <coroutine>
+#include <memory>
+
+/// Specialize `std::coroutine_traits` for `google::cloud::future<T>`.
+template <typename T, typename... Args>
+requires(!std::is_void_v<T> && !std::is_reference_v<T>) /**/
```
devjgm, line 28: Is `/**/` intentional?

devjgm (approved):
I'm new to C++20, but from what I can tell this LGTM.

google/cloud/internal/future_coroutines.h
coryan: yes, `clang-format` does a poor job with these otherwise.

Issue #6767, opened 2021-06-14
Consider making CMAKE_CXX_STANDARD a required user-provided option 


devjgm:
Today our CMake builds set the C++ language standard to 11 by default
https://github.com/googleapis/google-cloud-cpp/blob/d8eedaea719d9dd96d7f16dadd5b5b646eec3b8e/CMakeLists.txt#L37-L39

Users can override that with something like `cmake -DCMAKE_CXX_STANDARD=17 ...`, but if they don't specify something, we choose 11. Defaulting to C++11 will bias our telemetry data about C++ standard usage. And 11 may not be the correct default if users are using, say, C++17 in their own code (though it should work correctly in most cases).

We should consider not defaulting `CMAKE_CXX_STANDARD` to 11 and instead *requiring* that users explicitly specify the standard that they want to use.

Of course, we should continue to _support_ C++11, and we'd need to have some builds that verify that our code works in a C++11 environment.

remyabel:
> Defaulting to C++11 will bias our telemetry data about C++ standard usage. 

Drive-by opinion. If the primary motivation is telemetry, I feel like there's still a problem. If users have to explicitly specify it and they say copy/paste from the instructions, they might just leave `-DCMAKE_CXX_STANDARD=11` in which as the same effect as it being default. Those using C++17 will explicitly change it but the numbers might look similar to before (this is just my intuition, I'm probably wrong). How about a big fat CMake warning/info that says "The default standard is C++11. If this is not what was intended, please change it..."

devjgm:
Well, I wouldn't say the primary motivation is telemetry. But to the extent that we have telemetry, making it more accurate would be good.

Maybe our instructions could say `-DCMAKE_CXX_STANDARD=<select one of 11, 14, 17, or 20>` or something, so that users can't copy-n-paste our chosen default.

Alternatively, maybe we should simply make our default the newest supported C++ standard, and then users can explicitly set it to anything they want all the way back to 11... but the default would be 20 or 17 or the latest standard supported by the compiler we're testing on.

devjgm:
Today we default to the oldest C++ standard we support (C++11), as described in the first comment in this issue. I'm now starting to like the idea of defaulting to the *newest* C++ standard available, while of course still supporting users who explicitly need an older standard (back to C++11). What might this actually look like?

**Option 1: Default `CMAKE_CXX_STANDARD` in our `CMakeLists.txt` to the newest standard available.**

[...]

PR #8537, opened 2022-03-15
feat!: let CMAKE_CXX_STANDARD remain unset

devjgm:
Fixes: https://github.com/googleapis/google-cloud-cpp/issues/6767

Rather than us explicitly setting (defaulting) `CMAKE_CXX_STANDARD=11`, we leave it unset and rely on the compiler's default C++ language standard. This will let customers more easily build w/ the "correct" standard for their platform.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8537)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,12 +68,15 @@ https://github.com/googleapis/google-cloud-cpp/issues/8234.
 
 ## v1.39.0 - TBD
 
-In this release our CMake build no longer sets `CMAKE_CXX_STANDARD` to `11`
-when it was unspecified by the caller on Linux and Windows. Instead, we leave
-it unset, relying on the compiler's default C++ language standard. On macOS, we
-continue to default `CMAKE_CXX_STANDARD=11`, since the compiler's default of
-C++98 is too old. For more details, see:
-https://github.com/googleapis/google-cloud-cpp/issues/6767.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -68,6 +68,16 @@ https://github.com/googleapis/google-cloud-cpp/issues/8234.
 
 ## v1.39.0 - TBD
 
+**BREAKING CHANGES**
+
+* In this release our CMake build no longer sets `CMAKE_CXX_STANDARD` to `11`
```
coryan, line 73: consider:

```md
* Starting with this release Linux and Windows builds based on CMake
   no longer set `CMAKE_CXX_STANDARD` to `11`.  We rely on the compiler's
   default C++ language standard.  Note that all the compilers we support default
   to C++14.
  * on macOS the default is C++98, as `google-cloud-cpp` requires C++ >= 11, we continue
     to set `CMAKE_CXX_STANDARD` on that platform.
  * This only changes the default C++ version, we continue to test and support C++11.
  * For more details, see [#6767](https://github.com/googleapis/google-cloud-cpp/issues/6767). 
```

devjgm: Considered. Liked. Accepted. Pushed.

coryan (approved):

PR #8538, opened 2022-03-15
fix(storage): use correct offset and size types

### LINKED ISSUE #8536: ARM32 compile problem..
I want to use google-cloud-cpp library on ARM32.

So, I try to compile source on ARM32(ubuntu 20.04). 
(And When I try it on ARM64(ubuntu20.04), it compiles successfully. But some library which I used doesn't support ARM64, so I have to use ARM32)

It returns 2 errors. But I don't know how to fix it.

Please give me information to resolve it.

Here are the error log.
**/root/google-cloud-cpp/google/cloud/storage/parallel_upload.cc:409:30: error: conversion from 'long long int' to 'std::streamsize' {aka 'int'} may change value [-Werror=conversion]
  409 |     istream.read(buf.data(), to_copy);
      |                              ^~~~~~~
/root/google-cloud-cpp/google/cloud/storage/parallel_upload.cc:413:32: error: conversion from 'long long int' to 'std::streamsize' {aka 'int'} may change value [-Werror=conversion]
  413 |     ostream_.write(buf.data(), to_copy);
      |                                ^~~~~~~**

coryan:
Use the correct types of stream offsets (`std::streamoff`) vs.
stream buffer sizes (`std::streamsize`).  Cleanup some uses of
`std::uintmax_t` where `std::int64_t` is more reasonable.

Fixes #8536

### DIFF
--- a/google/cloud/storage/parallel_upload.cc
+++ b/google/cloud/storage/parallel_upload.cc
@@ -381,7 +381,8 @@ Status ParallelUploadFileShard::Upload() {
     left_to_upload_ = 0;
     return status;
   };
-  auto const already_uploaded = ostream_.next_expected_byte();
+  auto const already_uploaded =
+      static_cast<std::int64_t>(ostream_.next_expected_byte());
   if (already_uploaded > left_to_upload_) {
     return fail(StatusCode::kInternal, "Corrupted upload state, uploaded " +
                                            std::to_string(already_uploaded) +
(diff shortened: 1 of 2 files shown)


coryan:
Manually tested on Windows+x86 and @ahnjs1987 reports this works on ARM32 too.

devjgm (approved):

Issue #3701, opened 2020-04-07
Restrict visibility of google/cloud/*/BUILD targets


devjgm:
In https://github.com/googleapis/google-cloud-cpp/issues/3532 we're adding user-facing bazel build targets to the top-level https://github.com/googleapis/google-cloud-cpp/blob/master/BUILD file. In order to encourage/require users to only use the build targets in this top-level file we should do the following:

1. Initially, have the top-level *and* google/cloud/*/BUILD targets work for customers.
1. Message to users that this build-target change is coming, and that users should change their build targets from `//google/cloud/$x:$x_client` -> `//:$x_client`
1. Wait some period of time. 60 days? 90 days?
1. Change the default build visibility to `"//:__subpackages__"` in `google/cloud/*/BUILD`.

coryan:
An observation: if we want to hide some targets (e.g. `//google/cloud/spanner:spanner_client_testing`) we may want to mark them as deprecated ahead of time.

devjgm:
Ultimately we want to hide everything in `//google/...`. We want to expose _only_ the top-level `//:library` targets

coryan:
Yeah, I am just saying that this will require the usual amount of planning.  I am trying to keep the scope of #5726 as small as possible, so I am not doing anything to hinder this bug, but I am not doing anything to make it easier either.

coryan:
We also need to think of experimental features like GCS+gRPC: is is on a separate library target, maybe we should name it `//experimental:gcs_grpc` or something `//:experimental_gcs_grpc`.

devjgm:
Then to

> An observation: if we want to hide some targets (e.g. //google/cloud/spanner:spanner_client_testing) we may want to mark them as deprecated ahead of time.

I guess I'd say "yes". Any target in `google/cloud/.../BUILD` should be marked deprecated early so that we can change their default build visibility later.

Actually, maybe there's nothing we need to do. Our https://github.com/googleapis/google-cloud-cpp#api-breaking-changes doc says 

> Any file or symbol that lives within a directory or namespace containing "internal", "impl", "test", "detail", "benchmark", "sample", or "example", is explicitly not part of our public API.

Would this apply to `//google/cloud/foo:foo_test`, et al? I think the spirit of that wording was that it should apply.

WDYT?

coryan:
[...]

PR #8550, opened 2022-03-16
cleanup: mark google/cloud/ bazel targets as deprecated

devjgm:
Part of: #3701

We want users to only use targets in the top-level //BUILD.bazel file.
Very likely nobody is using these targets even right now. But a 90+ day
deprecation would be nice. We could extend this to a year if people feel
that's necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8550)

### DIFF
--- a/google/cloud/BUILD.bazel
+++ b/google/cloud/BUILD.bazel
@@ -17,7 +17,7 @@
 package(
     default_deprecation = """
     These targets are not for public use and will be removed sometime around
-    July 1, 2022. Only use Bazel targets defined in the top-level //BUILD.bazel
+    2023-03-16. Only use Bazel targets defined in the top-level //BUILD.bazel
     file. See https://github.com/googleapis/google-cloud-cpp/issues/3701 for
     more details.
     """,
--- a/google/cloud/BUILD.bazel
+++ b/google/cloud/BUILD.bazel
@@ -12,7 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-package(default_visibility = ["//visibility:public"])
+# TODO(#3701) Change the default_visibility to "//visibility:private", and
+# widen visibility as needed with per-target visibility.
+package(
+    default_deprecation = """
+    These targets are not for public use and will be removed sometime around
+    July 1, 2022. Only use Bazel targets defined in the top-level //BUILD.bazel
+    file. See https://github.com/googleapis/google-cloud-cpp/issues/3701 for
+    more details.
+    """,
+    default_visibility = ["//visibility:public"],
+)
 
 licenses(["notice"])  # Apache 2.0
 
--- a/google/cloud/grpc_utils/BUILD.bazel
+++ b/google/cloud/grpc_utils/BUILD.bazel
@@ -28,7 +28,7 @@ cc_library(
     ],
     deprecation = (
         "please use //google/cloud:google_cloud_cpp_grpc_utils instead." +
-        " This target will be removed on or about 2023-03-16. More details at" +
+        " This target will be removed on or about 2023-01-01. More details at" +
         " https://github.com/googleapis/google-cloud-cpp/issues/3932"
     ),
     tags = ["manual"],
--- a/google/cloud/grpc_utils/BUILD.bazel
+++ b/google/cloud/grpc_utils/BUILD.bazel
@@ -12,17 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# TODO(#3701) Change the default_visibility to "//visibility:private", and
-# widen visibility as needed with per-target visibility.
-package(
-    default_deprecation = """
-    These targets are not for public use and will be removed sometime around
-    July 1, 2022. Only use Bazel targets defined in the top-level //BUILD.bazel
-    file. See https://github.com/googleapis/google-cloud-cpp/issues/3701 for
-    more details.
-    """,
-    default_visibility = ["//visibility:public"],
-)
+package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])  # Apache 2.0
 
@@ -38,7 +28,7 @@ cc_library(
     ],
     deprecation = (
         "please use //google/cloud:google_cloud_cpp_grpc_utils instead." +
-        " This target will be removed on or about 2023-01-01. More details at" +
+        " This target will be removed on or about 2023-03-16. More details at" +
         " https://github.com/googleapis/google-cloud-cpp/issues/3932"
     ),
     tags = ["manual"],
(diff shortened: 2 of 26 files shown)


google/cloud/BUILD.bazel
```diff
@@ -12,7 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-package(default_visibility = ["//visibility:public"])
+# TODO(#3701) Change the default_visibility to "//visibility:private", and
+# widen visibility as needed with per-target visibility.
+package(
+    default_deprecation = """
+    These targets are not for public use and will be removed sometime around
+    July 1, 2022. Only use Bazel targets defined in the top-level //BUILD.bazel
```
coryan, line 20: nit: 2022-07-01 is probably more readable for folks outside the US.

google/cloud/grpc_utils/BUILD.bazel
```diff
@@ -12,7 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-package(default_visibility = ["//visibility:public"])
+# TODO(#3701) Change the default_visibility to "//visibility:private", and
+# widen visibility as needed with per-target visibility.
+package(
+    default_deprecation = """
+    These targets are not for public use and will be removed sometime around
+    July 1, 2022. Only use Bazel targets defined in the top-level //BUILD.bazel
```
coryan, line 20: The only target here is slated for deprecation in 2023-01-01 (see around line 41), it seems like we should stick to that date.

google/cloud/BUILD.bazel
```diff
@@ -12,7 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-package(default_visibility = ["//visibility:public"])
+# TODO(#3701) Change the default_visibility to "//visibility:private", and
+# widen visibility as needed with per-target visibility.
+package(
```
coryan, line 17: This file happens to have some libraries that have (AFAIK) no corresponding top-level target: `google_cloud_cpp_common` and `google_cloud_cpp_grpc_utils`.

Maybe for them we should decide if we are going to make these libraries public (I think we should).  And I think the deprecation period may need to be longer?

coryan (changes requested):
There are some `*_testing` libraries that get deprecated with this change.  I think they should be deprecated, maybe the time period is too short.

This also deprecates the only want to link directly against `google_cloud_cpp_common` and `google_cloud_cpp_grpc_utils`.  Those seem to be part of our public API. We should add top-level targets for them.  The deprecation period also feels short for them.

google/cloud/testing_util/BUILD.bazel
```diff
@@ -12,7 +12,17 @@
...
+# widen visibility as needed with per-target visibility.
+package(
+    default_deprecation = """
+    These targets are not for public use and will be removed sometime around
+    July 1, 2022. Only use Bazel targets defined in the top-level //BUILD.bazel
+    file. See https://github.com/googleapis/google-cloud-cpp/issues/3701 for
+    more details.
+    """,
+    default_visibility = ["//visibility:public"],
+)
```
coryan, line 25: Argh, also here, there are rules here that are already deprecated.

devjgm:
> This also deprecates the only want to link directly against `google_cloud_cpp_common` and `google_cloud_cpp_grpc_utils`. Those seem to be part of our public API. We should add top-level targets for them. The deprecation period also feels short for them.

I was debating whether "common" was actually part of our public API, or if the intent is that you just get them as part of your dep on one of our public user-facing libraries. If they are intended to be part of our public API on their own, then I agree we need top-level targets for them. However, it's not 100% clear to me that we want that.

google/cloud/BUILD.bazel
devjgm: Good point. I'll change to that format once we know how long to deprecate these. I'd be happy to pick 2023-01-01 for all of this to keep it simple. We could also pick a full year from now. Or something else. WDYT?

google/cloud/testing_util/BUILD.bazel
devjgm: Yes, I saw those. But to keep things simpler, I was leaning toward just handling these deprecations as whole files, even if an already deprecated thing stays around a bit longer. I could remove the other deprecation notices though.

coryan:
> I was debating whether "common" was actually part of our public API, or if the intent is that you just get them as part of your dep on one of our public user-facing libraries. If they are intended to be part of our public API on their own, then I agree we need top-level targets for them. However, it's not 100% clear to me that we want that.

I think the ship named "Should we make this part of our public API" sailed, picked up cargo in 5 ports, came back, delivered the cargo, and sailed again. 😄

What follows is not as black and white, but arguably, there are APIs in these libraries useful when using C++ in GCP, and can be used even without any client library, e.g.:
  - `g::c::Project`, `g::c::KmsKeyName` (and I expect we will add `g::c::Location` at some point).
  - `g::c::future` and `g::c::CompletionQueue` might be useful outside the libraries too.

google/cloud/testing_util/BUILD.bazel
coryan: Delaying the deprecation to simplify things SGTM.  Removing them right away also SGTM.

devjgm:
> > I was debating whether "common" was actually part of our public API, or if the intent is that you just get them as part of your dep on one of our public user-facing libraries. If they are intended to be part of our public API on their own, then I agree we need top-level targets for them. However, it's not 100% clear to me that we want that.
> 
> I think the ship named "Should we make this part of our public API" sailed, picked up cargo in 5 ports, came back, delivered the cargo, and sailed again. 😄

For the C++ code, I agree. For the build targets, I don't think it's clear, but I'm happy to defer to your preference for how to proceed. I'll prepare a PR to add them to //BUILD.bazel.

> 
> What follows is not as black and white, but arguably, there are APIs in these libraries useful when using C++ in GCP, and can be used even without any client library, e.g.:
> 
> * `g::c::Project`, `g::c::KmsKeyName` (and I expect we will add `g::c::Location` at some point).
> * `g::c::future` and `g::c::CompletionQueue` might be useful outside the libraries too.

devjgm:
@coryan PTAL.

Some of the noteworthy changes.
* The bazel config_settings that were already deprecated I removed in https://github.com/googleapis/google-cloud-cpp/pull/8556
* I pushed back the deprecation date a full year to 2023-03-16, which should cover us in all these cases.
* I changed the date format to something more universal.
* I exposed the common and grpc_utils targets in the top-level BUILD file in https://github.com/googleapis/google-cloud-cpp/pull/8552

coryan (approved):

PR #8552, opened 2022-03-16
feat: expose build targets for common libraries

devjgm:
Since #3701 says we want users to only use top-level build targets like
`//:foo`, we need to expose the targets from our
`google/cloud/BUILD.bazel` file.

I'm happy to bikeshe the target names here.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8552)


coryan (approved):
The names (and this change in general) LGTM.

PR #8555, opened 2022-03-16
Feature/low level publisher implementation

18suresha:
tests will be added in a later PR


coryan (changes requested):
> Reviewed 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555)* status: all files reviewed, 16 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/partition_publisher.h, line 36 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyMzRyy4tYJgr_eg60C:-MyMzRyy4tYJgr_eg60D:b-ngns62) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher.h#L36)):*
> ```objc
>     : public Publisher<google::cloud::pubsublite::v1::Cursor> {
>  private:
>   using ResumableStream = std::unique_ptr<ResumableAsyncStreamingReadWriteRpc<
> ```

Judging from the code below, it seems that 

```cc
  using ResumableStream = ResumableAsyncStreamingReadWriteRpc<
      google::cloud::pubsublite::v1::PublishRequest,
      google::cloud::pubsublite::v1::PublishResponse>;
```

would be more useful.
___
*[google/cloud/pubsublite/internal/partition_publisher.cc, line 79 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyMxFnY8yJ-JX02TInu:-MyMxFnY8yJ-JX02TInv:bq31gou) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher.cc#L79)):*
> ```cpp
>   {
>     std::lock_guard<std::mutex> g{mu_};
>     for (auto& batch :
> ```

This is:

```cc
auto batches = CreateBatches(...);
std::move(batches.begin(), batches.end(), std::back_inserter(unsent_batches_));
```
___
*[google/cloud/pubsublite/internal/partition_publisher.cc, line 109 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyMyKVxE7AteMX6yPXx:-MyMyKVxE7AteMX6yPXy:blr9zgq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher.cc#L109)):*
> ```cpp
>       *publish_request.mutable_message_publish_request();
>   for (auto& message_with_future : in_flight_batches_.back()) {
>     *message_publish_request.add_messages() = message_with_future.message;
> ```

Why copy `message_with_future.message`? Should we `std::move()` it?  If a copy is necessary, please add a comment explaining the rationale.
___
*[google/cloud/pubsublite/internal/partition_publisher.cc, line 123 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyMtfME3J_nc4VCT_lZ:-MyMtfME3J_nc4VCT_l_:b-luxxnd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher.cc#L123)):*
> ```cpp
>   AsyncRoot root;
>   if (!service_composite_.status().ok()) return;
>   std::lock_guard<std::mutex> g{mu_};
> ```

What does this guard do?
___
*[google/cloud/pubsublite/internal/partition_publisher.cc, line 128 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyMuaEr-SQNQA4H4WTr:-MyMuaEr-SQNQA4H4WTs:b-66q2yp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher.cc#L128)):*
> ```cpp
>       .then([this](future<absl::optional<PublishResponse>>
>                        optional_response_future) {
>         auto optional_response = optional_response_future.get();
> ```

This `.then()` body is large enough that it deserves its own function.  Maybe:

```cc
.then(ChainFuture(resumable_stream_->Read())
.then([this](future<absl::optional<PublishResponse>> f) {
  auto response = f.get();
  return OnRead(std::move(response));
});
```

Also, avoid encoding the type of a variable in its name.  I can see it is an optional and a future by looking at the type.  And there is no need for very long names when the scope of the variable is short.. 
___
*[google/cloud/pubsublite/internal/partition_publisher.cc, line 216 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyMtpH39lAAm_OEKyiq:-MyMtpH39lAAm_OEKyir:bms9gsf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher.cc#L216)):*
> ```cpp
> 
> void PartitionPublisher::SatisfyOutstandingMessages() {
>   auto messages_with_futures = UnbatchAll(std::unique_lock<std::mutex>{mu_});
> ```

Here and elsewhere, the variable name should be `messages_with_promises`.  We changed the type but neglected the variables.

Generally I find such naming a bit of a code smell, `x_with_promises` tells me about the structure of the object, but not about what their role is.  I think it is more interesting that these messages have not received an `ack` from the service than the fact that they have an associated promise. `unacked_messages`?  
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 208 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN0ULh4tgDDWV1WrNs:-MyN0ULh4tgDDWV1WrNt:b-hdiyfv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L208)):*
> ```cpp
>   auto batches =
>       TestCreateBatches(std::move(message_with_futures), std::move(options));
>   EXPECT_EQ(batches.size(), ceil(((double)messages.size()) / 3.0));
> ```

You can make these changes in a separate PR, sorry I missed them earlier:

* We would say `std::ceil()` also add `#include <cmath>` at the top
* Seems like it would be clearer if you just defined some constants in this test:

```cc
auto constexpr kMessageCount = 10;
auto constexpr kBatchMessageCount = 3;
auto constexpr kBatchCount = 4; // std::ceil(10 / 3)
```
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 209 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN1qymCDM4c9_z6xIk:-MyN1qymCDM4c9_z6xIl:bqh7goe) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L209)):*
> ```cpp
>       TestCreateBatches(std::move(message_with_futures), std::move(options));
>   EXPECT_EQ(batches.size(), ceil(((double)messages.size()) / 3.0));
>   for (unsigned int i = 0; i < batches.size(); ++i) {
> ```

I think something like this would be clearer.

```cc
ASSERT_EQ(batches.size(), kBatchCount);
EXPECT_THAT(batches[0], ElementsAre(IsProtoEqual(messages[0]), IsProtoEqual(messages[1]), IsProtoEqual(messages[2])));
EXPECT_THAT(batches[1], ElementsAre(IsProtoEqual(messages[3]), IsProtoEqual(messages[4]), IsProtoEqual(messages[4])));
EXPECT_THAT(batches[2], ElementsAre(IsProtoEqual(messages[6]), IsProtoEqual(messages[7]), IsProtoEqual(messages[8])));
EXPECT_THAT(batches[3], ElementsAre(IsProtoEqual(messages[9])));
```
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 252 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN2diD4LtcMhUx3Brq:-MyN2diD4LtcMhUx3Brr:b-z67k35) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L252)):*
> ```cpp
>   auto batches =
>       TestCreateBatches(std::move(message_with_futures), std::move(options));
>   EXPECT_EQ(batches.size(), ceil(((double)messages.size()) / 3.0));
> ```

Same comments as above to improve the tests.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 288 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN2h1h-qLsOzbq1WQj:-MyN2h1h-qLsOzbq1WQk:bcxwok6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L288)):*
> ```cpp
>   }
> 
>   unsigned int batch_boundary_ = 5;
> ```

Sounds like this should be a constant, like `kAlarmDuration`.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 294 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN5zVI7G6FYAtdIT51:-MyN5zVI7G6FYAtdIT52:b-ymfx9k) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L294)):*
> ```cpp
>   MockAlarmRegistry alarm_;
>   std::function<void()> leaked_alarm_;
>   MockResumableAsyncReaderWriter<PublishRequest, PublishResponse>*
> ```

Let's just make it a rule: use `std::unique_ptr<>` or `std::shared_ptr<>`.  The raw pointers in these tests are just gross.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 316 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN32Tk-Xvv3HmpM_d6:-MyN32Tk-Xvv3HmpM_d7:b54ddip) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L316)):*
> ```cpp
>   future<Status> publisher_start_future = publisher_->Start();
> 
>   auto* underlying_stream = new StrictMock<AsyncReaderWriter>;
> ```

Is there a reason we cannot use `absl::make_unique<>` here?
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 334 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN3U5kCwGB_b4jObcT:-MyN3U5kCwGB_b4jObcU:b-on6bty) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L334)):*
> ```cpp
> 
>   std::vector<future<StatusOr<Cursor>>> publish_message_futures;
>   for (unsigned int i = 0; i < 11; ++i) {
> ```

use a range-for here.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 347 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN3Za0A41SdXyNtBZ5:-MyN3Za0A41SdXyNtBZ6:bwowe5k) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L347)):*
> ```cpp
>   shutdown_future.get();
> 
>   for (unsigned int i = 0; i < individual_publish_messages.size(); ++i) {
> ```

Use a `range-for` here.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 437 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN3iVn1gP1zUV19svD:-MyN3iVn1gP1zUV19svE:b1n6j1n) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L437)):*
> ```cpp
>   future<Status> publisher_start_future = publisher_->Start();
> 
>   auto* underlying_stream = new StrictMock<AsyncReaderWriter>;
> ```

Unless there is a good reason, `absl::make_unique<>`
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 494 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN49v97Z5J99JwJtaf:-MyN49v97Z5J99JwJtag:bp205tq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L494)):*
> ```cpp
>   future<Status> publisher_start_future = publisher_->Start();
> 
>   auto* underlying_stream = new StrictMock<AsyncReaderWriter>;
> ```

here and elsewhere `absl::make_unique<>`.

coryan (changes requested):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555)* status: all files reviewed, 15 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/partition_publisher.cc, line 250 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MySBV7J8KEFfS4t-rwp:-MySBV7J8KEFfS4t-rwq:bb9uhcl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/26933b3569ee4308d6d976cda725bebda1ee3950/google/cloud/pubsublite/internal/partition_publisher.cc#L250)):*
> ```cpp
>         return (*shared_stream)->Finish();
>       })
>       .then([this, shared_stream](future<Status> status_future)
> ```

optional: this is another case where `status_future` is a longer name that tells me nothing new.  `f` would have been just fine.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 334 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN3U5kCwGB_b4jObcT:-MySCF3L0Mt-CWrMGXTK:b-buszup) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L334)):*

use a range-for here.
</blockquote></details>

What I meant is that you can use `for (auto& m : individual_publish_messages) {`
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 347 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN3Za0A41SdXyNtBZ5:-MySCWFK8JKfqUV-BxDV:b6zo8wh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L347)):*

Use a `range-for` here.
</blockquote></details>

Same here, you are iterating over `individual_publish_messages`, just do that.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 112 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MySDbLf-BpluZTkdStX:-MySDbLf-BpluZTkdStY:bwqt7yi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/26933b3569ee4308d6d976cda725bebda1ee3950/google/cloud/pubsublite/internal/partition_publisher_test.cc#L112)):*
> ```cpp
>   }
> 
>   static std::vector<PubSubMessage> GetMessageDequeFromBatch(
> ```

Can you tell me why this function is named as it is?  Seems like it returns a vector, not a deque.  This is one of the reasons you don't want to encode the types into the names of things.  Just call it `GetMessagesFromBatch(...)`. 
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 356 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MySC_w69aaxm3En0ZqE:-MySC_w69aaxm3En0ZqF:bjoj3as) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/26933b3569ee4308d6d976cda725bebda1ee3950/google/cloud/pubsublite/internal/partition_publisher_test.cc#L356)):*
> ```cpp
> 
>   std::vector<PubSubMessage> individual_publish_messages;
>   for (unsigned int i : {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11}) {
> ```

The classical for loop was fine to initialize `individual_publish_messages`.  Though maybe I would have written 

```cc
auto constexpr kMessageCount = 2 * kBatchBoundary + 1;  // We want two full batches and a partial one.
for (int i = 0; i != kMessageCount; ++i) {
...
}
```

In C++14 I would have written:

```cc
auto constexpr kMessageCount = 2 * kBatchBoundary + 1;  // We want two full batches and a partial one.
std::generate_n(std::back_inserter(individual_publish_messages, kMessageCount, [i = 0]() mutable {
    PubSubMessage message;
    *message.mutable_key() = "key";
    *message.mutable_data() = std::to_string(i++);
    return message;
});
```
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 668 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MySFY-tAqOHDv9PfqYy:-MySFY-tAqOHDv9PfqYz:b-7unqa7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/26933b3569ee4308d6d976cda725bebda1ee3950/google/cloud/pubsublite/internal/partition_publisher_test.cc#L668)):*
> ```cpp
>   read_promise.set_value(absl::optional<PublishResponse>());
> 
>   EXPECT_CALL(alarm_token_ref_, Destroy);
> ```

I think you are creating `alarm_token_ref_` and `resumable_stream_ref_` just so you can create expectations after you `std::move()` the mocks.  Is there a reason you cannot add the expectations before the `std::move()`?

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555)* status: all files reviewed, 15 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 668 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MySFY-tAqOHDv9PfqYy:-MyVgIh6AukMz-QBoReV:b710j26) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/26933b3569ee4308d6d976cda725bebda1ee3950/google/cloud/pubsublite/internal/partition_publisher_test.cc#L668)):*

I think you are creating `alarm_token_ref_` and `resumable_stream_ref_` just so you can create expectations after you `std::move()` the mocks.  Is there a reason you cannot add the expectations before the `std::move()`? 
</blockquote></details>

Yup, we use `InSequence seq;` to ensure that all our expectations occur in the order we specify. For example, we only want `DestroyToken` to be called upon destruction.

coryan (changes requested):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555)* status: all files reviewed, 8 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 208 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN0ULh4tgDDWV1WrNs:-MyZOAITBZOovl77M3eW:b30hci0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L208)):*

You can make these changes in a separate PR, sorry I missed them earlier:

* We would say `std::ceil()` also add `#include <cmath>` at the top
* Seems like it would be clearer if you just defined some constants in this test:

```cc
auto constexpr kMessageCount = 10;
auto constexpr kBatchMessageCount = 3;
auto constexpr kBatchCount = 4; // std::ceil(10 / 3)
```

</blockquote></details>

My comments about `ceil` remain unaddressed.  You could say "I am planning to fix that in a separate PR", or "I think it is not important", or just fix it.  But unless you say something I don't know what you are planning to do.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 668 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MySFY-tAqOHDv9PfqYy:-MyZR33_B86zvGgECvix:b47226f) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/26933b3569ee4308d6d976cda725bebda1ee3950/google/cloud/pubsublite/internal/partition_publisher_test.cc#L668)):*
> Yup, we use InSequence seq; to ensure that all our expectations occur in the order we specify. 

That does not answer the question.  Or at least: the immediate followup question is why couldn't you set the expectations using that `InSequence` but before the moving-from the smart pointers that hold these mocks.

I am again wondering if the benefits of using `std::unique_ptr<>` for some of these things outweigh the costs.  Yes one part of the code got clearer, at the cost of making these tests much harder to follow and reason about.  Unless we can clean up the tests, I think we should go back and make the initializers consume `std::shared_ptr<>`.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 304 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyZQ-Nx06rxDK06xIEQ:-MyZQ-Nx06rxDK06xIER:ba4mc4e) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/079d0b4078af672959e5ce0fd3e6b8b143432459/google/cloud/pubsublite/internal/partition_publisher_test.cc#L304)):*
> ```cpp
>         [&](StreamInitializer<PublishRequest, PublishResponse> const&
>                 initializer) {
>           initializer_ = std::move(initializer);
> ```

a `std::move()` for a `T const&` may not do what you think it does.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 316 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyZQX8FA1hMT_-TuqZC:-MyZQX8G8M1HrNdpK7Vg:bg5q087) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/079d0b4078af672959e5ce0fd3e6b8b143432459/google/cloud/pubsublite/internal/partition_publisher_test.cc#L316)):*
> ```cpp
>   StrictMock<MockAlarmRegistryCancelToken>& alarm_token_ref_;
>   MockAlarmRegistry alarm_;
>   std::function<void()> leaked_alarm_;
> ```

the name of this variable is a sign that something fishy is going on.  I would prefer if we could work without "leaking" things (whatever that means in this context), but if we must, you need to add some comments explaining why
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 321 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyZQHMrBXSVQB2KZxxY:-MyZQHMrBXSVQB2KZxxZ:bwqn1gn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/079d0b4078af672959e5ce0fd3e6b8b143432459/google/cloud/pubsublite/internal/partition_publisher_test.cc#L321)):*
> ```cpp
>       resumable_stream_ = absl::make_unique<StrictMock<
>           MockResumableAsyncReaderWriter<PublishRequest, PublishResponse>>>();
>   MockResumableAsyncReaderWriter<PublishRequest, PublishResponse>&
> ```

I would prefer if these `*_ref` variables did not have to exist.  But if they must, please add some comments explaining why they remain valid until the destruction of this class.

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555)* status: all files reviewed, 8 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 208 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyN0ULh4tgDDWV1WrNs:-My_oIcx3dIbkUWF3pPC:bgvk2db) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4006e06b401dae5fcef7cb134b7bbf91c1ceeb1c/google/cloud/pubsublite/internal/partition_publisher_test.cc#L208)):*

My comments about `ceil` remain unaddressed.  You could say "I am planning to fix that in a separate PR", or "I think it is not important", or just fix it.  But unless you say something I don't know what you are planning to do.
</blockquote></details>

oh yea sorry, was going to address it in a separate PR, but I should've mentioned what I'm doing
I made the change to use `std::ceil` and store the magic numbers in variables
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 668 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MySFY-tAqOHDv9PfqYy:-My_oDi65ORh9LV9Ys6B:bodo78q) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/26933b3569ee4308d6d976cda725bebda1ee3950/google/cloud/pubsublite/internal/partition_publisher_test.cc#L668)):*

> Yup, we use InSequence seq; to ensure that all our expectations occur in the order we specify. 

That does not answer the question.  Or at least: the immediate followup question is why couldn't you set the expectations using that `InSequence` but before the moving-from the smart pointers that hold these mocks.

I am again wondering if the benefits of using `std::unique_ptr<>` for some of these things outweigh the costs.  Yes one part of the code got clearer, at the cost of making these tests much harder to follow and reason about.  Unless we can clean up the tests, I think we should go back and make the initializers consume `std::shared_ptr<>`.

</blockquote></details>

The `InSequence seq;` enforces that `EXPECT_CALL`s actually occur in the order they were invoked. Making those calls before `std::move` would be an invalid order of how these mocked calls are invoked.
We can't setup the `EXPECT_CALL`s before moving the smart pointers because that occurs in the publisher constructor and we still need to `EXPECT_CALL` on these moved objects toward the end of test cases.

Perhaps we could use lambda capturing to create functors that can be called in place of those later `EXPECT_CALL`s, but I think this will decrease readability of the test suite, be more complicated to implement, and `InSequence seq;` in the functor will only enforce order in the functor's scope rather than the whole test case.

___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 304 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyZQ-Nx06rxDK06xIEQ:-My_o-BiE4j-fIuEmJjG:bu10esi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/079d0b4078af672959e5ce0fd3e6b8b143432459/google/cloud/pubsublite/internal/partition_publisher_test.cc#L304)):*

a `std::move()` for a `T const&` may not do what you think it does.
</blockquote></details>

changed it to a pass-by-value argument as it is in the header
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 316 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyZQX8FA1hMT_-TuqZC:-My_nrwd8VArI9DOeYuc:b-qoehy3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/079d0b4078af672959e5ce0fd3e6b8b143432459/google/cloud/pubsublite/internal/partition_publisher_test.cc#L316)):*

the name of this variable is a sign that something fishy is going on.  I would prefer if we could work without "leaking" things (whatever that means in this context), but if we must, you need to add some comments explaining why
</blockquote></details>

I renamed the variable and explained why we're using it

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555)* status: 2 of 3 files reviewed, 8 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 321 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyZQHMrBXSVQB2KZxxY:-MybNQjc3Rvq1jbcrNV0:b51kggp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/079d0b4078af672959e5ce0fd3e6b8b143432459/google/cloud/pubsublite/internal/partition_publisher_test.cc#L321)):*

I would prefer if these `*_ref` variables did not have to exist.  But if they must, please add some comments explaining why they remain valid until the destruction of this class.
</blockquote></details>

added comments

coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555)* status: 2 of 3 files reviewed, 5 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 668 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MySFY-tAqOHDv9PfqYy:-MygTQaL0IkJbkfFaDb4:b-rbypqw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/26933b3569ee4308d6d976cda725bebda1ee3950/google/cloud/pubsublite/internal/partition_publisher_test.cc#L668)):*
> The InSequence seq; enforces that EXPECT_CALLs actually occur in the order they were invoked.

I understand that you want these expectations to be satisfied in an specific order, you do not need to repeat that.  

> Making those calls before std::move would be an invalid order of how these mocked calls are invoked.

 Why is that the case?  You could setup *all* the `EXPECT_CALL()`, in the right sequence, before calling `initializer(_(std::move(underlying_stream_))`, or before any other calls.

I am going to be blunt: as soon as I have some time I will rip out the `std::unique_ptr<>` from `ResumableAsyncStreamingReadWriteRpc`.  It is not worth the complexity it introduces in the tests.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 321 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MyZQHMrBXSVQB2KZxxY:-MygVeOz1K0U0XRfWAz3:bq7a0k0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/079d0b4078af672959e5ce0fd3e6b8b143432459/google/cloud/pubsublite/internal/partition_publisher_test.cc#L321)):*

added comments
</blockquote></details>

Thanks.  Keeping pointers to a `std::move'd` object held by a `std::unique_ptr<>` is inherently dangerous.  If you insist on this approach, your comments should also explain why this is safe.  How do you know these `std::move'd` object will not go out of scope for the lifetime of the class?  And be advised, if this requires you to know a lot about how the classes under test are implemented, then that is a code smell too.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 201 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MygUSBg2mdPrVIAWOkW:-MygUSBg2mdPrVIAWOkX:b-qvqdl6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d9340d71a86feeccfa1489fedbd7e56e1f919470/google/cloud/pubsublite/internal/partition_publisher_test.cc#L201)):*
> ```cpp
>   std::deque<PubSubMessage> messages;
>   unsigned int const num_messages = 10;
>   unsigned int const max_batch_message_count = 3;
> ```

You can just add:


```cc
auto const expected_batch_count = (num_messages + max_batch_message_count - 1) / max_batch_message_count;
```

and avoid all the casting to `double` and `std::ceil()` calls.
___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 256 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MygV6-E7_lmnBNeB_ut:-MygV6-E7_lmnBNeB_uu:bjv2e06) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d9340d71a86feeccfa1489fedbd7e56e1f919470/google/cloud/pubsublite/internal/partition_publisher_test.cc#L256)):*
> ```cpp
>   // all messages are of same size so <message size> * 3 is enough for 3
>   // messages
>   unsigned int const max_batch_message_count = 3;
> ```

Again, you can calculate the number of batches at compile-time, without casts, without calling `std::ceil()`.

coryan (approved):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @18suresha)

___
*[google/cloud/pubsublite/internal/partition_publisher_test.cc, line 668 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8555#-MySFY-tAqOHDv9PfqYy:-Mym3s1U8Oc5Cvaog2rn:b-56rjba) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/26933b3569ee4308d6d976cda725bebda1ee3950/google/cloud/pubsublite/internal/partition_publisher_test.cc#L668)):*
>> I am going to be blunt: as soon as I have some time I will rip out the std::unique_ptr<> from ResumableAsyncStreamingReadWriteRpc. It is not worth the complexity it introduces in the tests.
> 
> I disagree that we should optimize for not needing the test to have a lifetime comment, resumable stream lives as long as the publisher_ object, in exchange for less clear ownership of the stream object. 

First, I think the code now explains why it works, so we should proceed.

With that said: tests have to be maintained, if they are hard to read, they will be hard to change. I understand you disagree.  This is just too much for me to keep in my head.  As I said, as soon as I have the time, I will change the code.  It was not expressed as a question.  It is a statement of what I will do.

PR #8560, opened 2022-03-17
ci: more hand-crafted code to clang-tidy build

coryan:
The `pubsublite` directory contains hand-crafted code and should be
checked with `clang-tidy`. Incidentally this regenerates all the `.bzl`
files, so we can check for differences as part of this build.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8560)

### DIFF
--- a/ci/cloudbuild/builds/clang-tidy.sh
+++ b/ci/cloudbuild/builds/clang-tidy.sh
@@ -31,7 +31,7 @@ mapfile -t cmake_args < <(cmake::common_args)
 cmake "${cmake_args[@]}" \
   -DCMAKE_CXX_CLANG_TIDY=/usr/local/bin/clang-tidy-wrapper \
   -DCMAKE_CXX_STANDARD=11 \
-  -DGOOGLE_CLOUD_CPP_ENABLE_GENERATOR=ON \
+  -DGOOGLE_CLOUD_CPP_ENABLE="bigtable;bigquery;generator;iam;logging;pubsub;pubsublite;spanner;storage" \
   -DGOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC=ON
 cmake --build cmake-out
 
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/builds/clang-tidy.sh
```diff
@@ -40,6 +40,11 @@ env -C cmake-out ctest "${ctest_args[@]}" -LE integration-test
 
 integration::ctest_with_emulators "cmake-out"
 
+# This build should fail if any generated .bzl files differ from what was
+# checked in.
+io::log_h2 "Highlight generated code differences"
+git diff --exit-code
```
devbww, line 46: As much as I sympathize with the intent, this makes a `clang-tidy` build fail if I have any pending changes, and it is difficult to discern whether it is only my pending edits that caused the failure, or whether there was an actual `clang-tidy` update.

Last time I commented on a similar issue with `checkers`, no one else seemed to care, so I wrote a wrapper that changed failure to success if the diffs before were the same as the diffs after.  Does anyone care more now?

devbww: BTW, such a wrapper is impossible in the `clang-tidy` case because it can, of course, fail for more reasons than updating a `.bzl` file, so you can't use diffs to determine the cause of the failure.

coryan: Maybe we should disable that check for interactive builds?

devbww: Sounds good to me, although I'm not sure how to distinguish an interactive build.

coryan: Here is an example:

https://github.com/googleapis/google-cloud-cpp/blob/d6f3a211f78d2c7978630cad1091ed0be095187b/ci/cloudbuild/build.sh#L206-L214

devbww: OK, thanks.  I'll experiment with that, and followup with a PR if it looks good.

Issue #8559, opened 2022-03-17
ObjectWriteStreambuf Corrupted When Bytes Not Fully Uploaded


Jseph:
**Does this issue affect the google-cloud-cpp project?**
Yes

**What component of `google-cloud-cpp` is this related to?**
GCS

**Describe the bug**
The object write streambuf implements std::basic_streambuf<char> and writes are performed using the streaming operator which triggers xsputn().  It makes a call to FlushRoundChunk and assumes that all bytes that aren't uploaded will be stored in current_ios_buffer_ and the internal stream pointers will get updated accordingly.

FlushRoundChunk makes a call to upload_session_->UploadChunk() and assumes that a good response means that all `rounded_size` bytes in `buffers` have been uploaded because they are later discarded.  The check [here](https://github.com/googleapis/google-cloud-cpp/blob/f8c472bafb9d87e5966078b24915bcf110639462/google/cloud/storage/internal/object_write_streambuf.cc#L203) should probably just check that `actual_next_byte < expected_next_byte`.  and return an error if this isn't the case, though more complicated logic could be added to see if there is room in the buffer for the uncommitted bytes and stick them there and return success.

This issue is triggered (or at least made more prevalent) by [this commit]([https://github.com/googleapis/google-cloud-cpp/commit/38d1d2a327dbbcb4c85e0cf88349c2a4fe143ed8](https://www.google.com/url?q=https://github.com/googleapis/google-cloud-cpp/commit/38d1d2a327dbbcb4c85e0cf88349c2a4fe143ed8&sa=D&source=buganizer&usg=AOvVaw1zoElsjZMnHcQiZR71O_Mp) which started return OK when no bytes were committed. 

**To Reproduce**

Very hard to do because it relies on very uncomon GCS behavior but here's how it's triggered (reach out to me internally for some logs that show this)

1. Create a resumable upload and start writing data.
2. Start writing bytes to the stream.
3. Either a) get GCS to partially upload the bytes or b) get GCS to return a retriable failure and then have reset_session succeed before the retry policy is exhausted in retry_resumable_upload_session.
4. Continue writing bytes to the stream.

The finalized object will be smaller than expected with a gap corresponding to the dropped bytes in 3.

**Expected behavior**
No dropped bytes.  The stream goes into a bad state if bytes are dropped, allowing the user to resume the session again and avoid corruption.

**Operating system:**
Na

**What compiler and version are you using?**
Na

[...]

PR #8562, opened 2022-03-17
fix(storage): return an error on short writes

coryan:
Very rarely the retry policy is exhausted during an upload. If this
happens when the last request was a query to find the status of the
upload the `UploadChunk()` succeeds with a short write.

We need to determine if that is a good idea (it might be). In the
interim, we need to signal to the application that the stream is no
longer usable. They can use `.Suspend()` to reset the stream (maybe in
another process) if that can be useful.

Part of the work for #8559

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8562)

### DIFF
--- a/google/cloud/storage/internal/object_write_streambuf.cc
+++ b/google/cloud/storage/internal/object_write_streambuf.cc
@@ -180,7 +180,6 @@ void ObjectWriteStreambuf::FlushRoundChunk(ConstBufferSequence buffers) {
   // GCS upload returns an updated range header that sets the next expected
   // byte. Check to make sure it remains consistent with the bytes stored in the
   // buffer.
-  auto first_buffered_byte = upload_session_->next_expected_byte();
   auto expected_next_byte = upload_session_->next_expected_byte() + actual_size;
   last_response_ = upload_session_->UploadChunk(payload);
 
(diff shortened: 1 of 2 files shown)


devbww (approved):

google/cloud/storage/internal/object_write_streambuf.cc
```diff
@@ -200,8 +199,10 @@ void ObjectWriteStreambuf::FlushRoundChunk(ConstBufferSequence buffers) {
...
     // to deal with this mess, so let's not duplicate that code here.
     auto actual_next_byte = upload_session_->next_expected_byte();
-    if (actual_next_byte < expected_next_byte &&
-        actual_next_byte < first_buffered_byte) {
+    if (actual_next_byte < expected_next_byte) {
+      // TODO(#8559) - if actual_next_byte < first_buffered_byte and there is
+      //   enough space in the buffer we could save the bytes for a future
+      //   write.
       std::ostringstream error_message;
       error_message << "Could not continue upload stream. GCS requested byte "
```
k7z45, line 207: Does this error message need to be updated?
If actual next byte is less than expected next byte, that means some bytes have not been committed and therefore might not be uploaded yet?

PR #8571, opened 2022-03-21
feat(common): make coroutine support optional

coryan:
Applications can disable coroutine support by adding
`-DGOOGLE_CLOUD_CPP_HAVE_COROUTINES=0` to the build.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8571)

### DIFF
--- a/google/cloud/future_coroutines_test.cc
+++ b/google/cloud/future_coroutines_test.cc
@@ -14,7 +14,7 @@
 // limitations under the License.
 
 #include "google/cloud/future.h"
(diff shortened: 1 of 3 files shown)


dbolduc:
Should this be documented somewhere? I don't have a good answer for where it should go. It looks like we do not document `GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS` either. Maybe they both belong in the top level README?

dbolduc (approved):
SGTM

PR #8575, opened 2022-03-22
feat(pubsublite): alarm registry implementation

18suresha:
This PR provides an implementation for `AlarmRegistry`, an asynchronous way to run a function at a specified time interval.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575)


google/cloud/pubsublite/internal/alarm_registry.h
```diff
@@ -41,6 +42,8 @@ class AlarmRegistry {
   AlarmRegistry& operator=(AlarmRegistry&&) = delete;
   // When CancelToken is destroyed, the alarm will not be running and will never
   // run again.
+  // All `CancelToken`s returned from `RegisterAlarm` must be destroyed before
```
dpcollins-google, line 45: This is a bad constraint to need- remove it

dpcollins-google: You also don't appear to need this anywhere

google/cloud/pubsublite/internal/alarm_registry_impl.cc
```diff
@@ -0,0 +1,78 @@
...
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace pubsublite_internal {
+
+AlarmRegistryImpl::AlarmRegistryImpl(google::cloud::CompletionQueue cq)
+    : cq_{std::move(cq)} {}
+
+void AlarmRegistryImpl::OnAlarm(CompletionQueue cq,
+                                std::chrono::milliseconds period,
+                                std::function<void()> const& on_alarm,
+                                std::shared_ptr<bool> const& alarm_status,
```
dpcollins-google, line 31: call this `shutdown`

google/cloud/pubsublite/internal/alarm_registry_impl.cc
```diff
@@ -0,0 +1,78 @@
...
+                                std::shared_ptr<std::mutex> const& mu) {
+  {
+    std::lock_guard<std::mutex> g{*mu};
+    if (!*alarm_status) return;
+  }
+  cq.MakeRelativeTimer(period).then(
+      [cq, period, on_alarm, alarm_status,
+       mu](future<StatusOr<std::chrono::system_clock::time_point>> f) {
+        bool is_ok = f.get().ok();
```
dpcollins-google, line 40: you don't actually need to set alarm_status to false: you're the only one looking at it. Just call:

```if (!f.get().ok()) return;```

Comment (dpcollins-google) [reply google/cloud/pubsublite/internal/alarm_registry_impl.cc:40]: maybe add a LOG(INFO) as well?

File: google/cloud/pubsublite/internal/alarm_registry.h (hunk 1 above)
Comment (18suresha) [reply google/cloud/pubsublite/internal/alarm_registry.h:45]: oh yea I think this was leftover from when I had `period` and `on_alarm` in a member variable vector

File: google/cloud/pubsublite/internal/alarm_registry_impl.h [hunk 4]
```diff
@@ -0,0 +1,71 @@
...
+  };
+
+  std::unique_ptr<AlarmRegistry::CancelToken> RegisterAlarm(
+      std::chrono::milliseconds period,
+      std::function<void()> on_alarm) override;
+
+ private:
+  // static with arguments rather than member variables, so parameters aren't
+  // bound to object lifetime
+  static void OnAlarm(CompletionQueue cq, std::chrono::milliseconds period,
```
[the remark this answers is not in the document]
18suresha: Added `AlarmState` as a private struct.

google/cloud/pubsublite/internal/alarm_registry_impl.h
```diff
@@ -0,0 +1,76 @@
...
+  std::unique_ptr<AlarmRegistry::CancelToken> RegisterAlarm(
+      std::chrono::milliseconds period,
+      std::function<void()> on_alarm) override;
+
+ private:
+  struct AlarmState {
+    CompletionQueue cq;
+    std::chrono::milliseconds period;
+    std::shared_ptr<std::mutex> mu;
+    // make on_alarm unique_ptr/shared_ptr?
```
dpcollins-google, line 59: none of these need to be shared_ptr- just give the CancelToken a shared_ptr<AlarmState>

google/cloud/pubsublite/internal/alarm_registry_impl.cc
```diff
@@ -0,0 +1,77 @@
...
+}
+
+std::unique_ptr<AlarmRegistry::CancelToken> AlarmRegistryImpl::RegisterAlarm(
+    std::chrono::milliseconds period, std::function<void()> on_alarm) {
+  // mu guards shutdown
+  auto mu = std::make_shared<std::mutex>();
+  auto shutdown = std::make_shared<bool>(false);
+  std::unique_ptr<AlarmRegistry::CancelToken> cancel_token =
+      absl::make_unique<CancelTokenImpl>(mu, shutdown);
+  OnAlarm(std::make_shared<AlarmState>(AlarmState{
```
dpcollins-google, line 69: this won't compile. You need to do:

OnAlarm(absl::WrapUnique(new AlarmState{...}))

google/cloud/pubsublite/internal/alarm_registry_impl.cc
```diff
@@ -0,0 +1,77 @@
...
+      cq_, period, std::move(mu), std::move(on_alarm), std::move(shutdown)}));
+  return cancel_token;
```
dpcollins-google, line 71:
```
std::shared_ptr<AlarmState> state{new AlarmState{...}};
ScheduleAlarm(state);
return absl::make_unique<CancelTokenImpl>(state);
```

google/cloud/pubsublite/internal/alarm_registry_impl.h
```diff
@@ -0,0 +1,76 @@
...
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace pubsublite_internal {
+
+// TODO(18suresha): description
+class AlarmRegistryImpl : public AlarmRegistry {
+ public:
+  explicit AlarmRegistryImpl(google::cloud::CompletionQueue cq);
+
+  // TODO(18suresha): implement destructor sanity check?
+  //  ~AlarmRegistryImpl() override;
```
dpcollins-google, line 35: this isn't needed.

google/cloud/pubsublite/internal/alarm_registry_impl_test.cc
```diff
@@ -0,0 +1,123 @@
...
+      .WillOnce(Return(ByMove(timer.get_future())));
+  temp_promise.set_value(std::chrono::system_clock::time_point{});
+
+  token = nullptr;
+
+  timer.set_value(std::chrono::system_clock::time_point{});
+  EXPECT_TRUE(flag);
+}
+
+TEST(AlarmRegistryImpl, TokenDestroyedAfterFiveRuns) {
```
dpcollins-google, line 92: Add a test for the token being destroyed during a run, and the destructor should block until the run completes. You can use std::thread synchronized with a few promise<void> to test this.

google/cloud/pubsublite/internal/alarm_registry_impl_test.cc
```diff
@@ -0,0 +1,123 @@
...
+using ::testing::InSequence;
+;
+using ::testing::Return;
+using ::testing::StrictMock;
+
+TEST(AlarmRegistryImpl, TokenDestroyedBeforeRun) {
+  auto cq = std::make_shared<StrictMock<MockCompletionQueueImpl>>();
+  AlarmRegistryImpl alarm{CompletionQueue{cq}};
+  bool flag = false;
+  auto fun = [&flag]() { flag = true; };
```
dpcollins-google, line 42: Use a StrictMock<MockFunction<void()>> instead.

google/cloud/pubsublite/internal/alarm_registry_impl.h
```diff
@@ -0,0 +1,78 @@
...
+               std::function<void()> alarm_func, bool is_shutdown)
+        : cq{std::move(cmq)},
+          period{prd},
+          on_alarm{std::move(alarm_func)},
+          shutdown{is_shutdown} {}
+    CompletionQueue cq;
+    std::chrono::milliseconds period;
+    std::mutex mu;
+    std::function<void()> on_alarm;  // ABSL_GUARDED_BY(*mu_)
+    bool shutdown;                   // ABSL_GUARDED_BY(*mu_)
```
dpcollins-google, line 50: alarm state should always start as not shut down. just use bool shutdown = false;

google/cloud/pubsublite/internal/alarm_registry_impl.h
```diff
@@ -0,0 +1,78 @@
...
+ public:
+  explicit AlarmRegistryImpl(google::cloud::CompletionQueue cq);
+
+  std::unique_ptr<AlarmRegistry::CancelToken> RegisterAlarm(
+      std::chrono::milliseconds period,
+      std::function<void()> on_alarm) override;
+
+ private:
+  struct AlarmState {
+    AlarmState(CompletionQueue cmq, std::chrono::milliseconds prd,
```
dpcollins-google, line 40: please don't use golang style abbreviations like this. Just use reasonable variable names. `period` is fine. so is `completion_queue`.

dpcollins-google: Ah, this is for not overloading the same name. This is fine to do, just use the same name and the compiler will figure it out.

google/cloud/pubsublite/internal/alarm_registry_impl.h
```diff
@@ -0,0 +1,78 @@
...
+  // run again.
+  class CancelTokenImpl : public AlarmRegistry::CancelToken {
+   public:
+    explicit CancelTokenImpl(std::shared_ptr<AlarmState>);
+
+    ~CancelTokenImpl() override;
+
+   private:
+    std::shared_ptr<AlarmState> state_;  // ABSL_GUARDED_BY(*mu_)
+  };
```
dpcollins-google, line 70: This doesn't need to be public. You're returning an instance of the interface.

google/cloud/pubsublite/internal/alarm_registry_impl.h
```diff
@@ -0,0 +1,78 @@
...
+    std::mutex mu;
+    std::function<void()> on_alarm;  // ABSL_GUARDED_BY(*mu_)
+    bool shutdown;                   // ABSL_GUARDED_BY(*mu_)
+  };
+
+  // static with arguments rather than member variables, so parameters aren't
+  // bound to object lifetime
+  static void ScheduleAlarm(std::shared_ptr<AlarmState> const& state);
+
+  google::cloud::CompletionQueue cq_;
```
dpcollins-google, line 57: const, you're only ever making a copy of this.

google/cloud/pubsublite/internal/alarm_registry_impl_test.cc
```diff
@@ -0,0 +1,177 @@
...
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace pubsublite_internal {
+namespace {
+
+using google::cloud::CompletionQueue;
+using google::cloud::testing_util::MockCompletionQueueImpl;
+using ::testing::ByMove;
+using ::testing::InSequence;
+;
```
dpcollins-google, line 35: strange semicolon, delete?

google/cloud/pubsublite/internal/alarm_registry_impl_test.cc
```diff
@@ -0,0 +1,177 @@
...
+
+  promise<StatusOr<std::chrono::system_clock::time_point>> timer;
+  EXPECT_CALL(*cq_, MakeRelativeTimer(std::chrono::nanoseconds(kAlarmPeriod)))
+      .WillOnce(Return(ByMove(timer.get_future())));
+  auto token = alarm_.RegisterAlarm(kAlarmPeriod, fun_.AsStdFunction());
+
+  EXPECT_CALL(fun_, Call);
+
+  auto temp_promise = std::move(timer);
+  timer = promise<StatusOr<std::chrono::system_clock::time_point>>{};
```
dpcollins-google, line 81: just do `promise<StatusOr<std::chrono::system_clock::time_point>> timer2;` instead of moving out of timer

google/cloud/pubsublite/internal/alarm_registry_impl_test.cc
```diff
@@ -0,0 +1,177 @@
...
+
+  promise<StatusOr<std::chrono::system_clock::time_point>> timer;
+  EXPECT_CALL(*cq_, MakeRelativeTimer(std::chrono::nanoseconds(kAlarmPeriod)))
+      .WillOnce(Return(ByMove(timer.get_future())));
+  auto token = alarm_.RegisterAlarm(kAlarmPeriod, fun_.AsStdFunction());
+
+  for (unsigned int i = 0; i < 5; ++i) {
+    EXPECT_CALL(fun_, Call);
+    auto temp_promise = std::move(timer);
+    timer = promise<StatusOr<std::chrono::system_clock::time_point>>{};
```
dpcollins-google, line 102: timer = {}; should work?

google/cloud/pubsublite/internal/alarm_registry_impl_test.cc
```diff
@@ -0,0 +1,177 @@
...
+
+TEST(AlarmRegistryImpl, TokenDestroyedDuringSecondRun) {
+  InSequence seq;
+
+  // not using StrictMock because of ON_CALL
+  auto cq = std::make_shared<MockCompletionQueueImpl>();
+  AlarmRegistryImpl alarm{CompletionQueue{cq}};
+  StrictMock<MockFunction<void()>> fun;
+
+  // using ctr to assert that MakeRelativeTimer is called at least twice
```
dpcollins-google, line 121: delete. Don't use ON_CALL, there should be no non-determinism

google/cloud/pubsublite/internal/alarm_registry_impl_test.cc
```diff
@@ -0,0 +1,177 @@
...
+
+  auto temp_promise = std::move(timer);
+  timer = promise<StatusOr<std::chrono::system_clock::time_point>>{};
+  temp_promise.set_value(std::chrono::system_clock::time_point{});
+
+  EXPECT_EQ(ctr, 2);
+
+  promise<void> in_alarm_function;
+  promise<void> destroy_finished;
+  auto destroy_token = [&in_alarm_function, &destroy_finished, &token]() {
```
dpcollins-google, line 147: initialize the thread inline. It will block until in_alarm_function is engaged.

google/cloud/pubsublite/internal/alarm_registry_impl.cc
```diff
@@ -0,0 +1,79 @@
...
+              "`MakeRelativeTimer` returned a non-ok `StatusOr`:",
+              status.status().message());
+          return;
+        }
+        {
+          std::lock_guard<std::mutex> g{state->mu};
+          if (state->shutdown) return;
+          state->on_alarm();
+        }
+        ScheduleAlarm(std::move(state));
```
dpcollins-google, line 50: std::move does nothing

dpcollins-google (approved):
> Reviewed 1 of 5 files at r1, 2 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575)* status: 3 of 7 files reviewed, 5 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/alarm_registry_impl.h, line 42 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-MysYGA38YW33kvOciUj:-MysYGA38YW33kvOciUk:b-pzp31p) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/57a01d5cf4ec006f112f4b6037059e892ff5e91d/google/cloud/pubsublite/internal/alarm_registry_impl.h#L42)):*
> ```objc
>     AlarmState(CompletionQueue completion_queue,
>                std::chrono::milliseconds period,
>                std::function<void()> alarm_func)
> ```

nit. on_alarm



_Code quote:_
```objc
alarm_func
```
___
*[google/cloud/pubsublite/internal/alarm_registry_impl.cc, line 64 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-MysYoH_AYJ-wOmwwzBC:-MysYoH_AYJ-wOmwwzBD:bbvndxd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/57a01d5cf4ec006f112f4b6037059e892ff5e91d/google/cloud/pubsublite/internal/alarm_registry_impl.cc#L64)):*
> ```cpp
>     std::chrono::milliseconds period, std::function<void()> on_alarm) {
>   std::shared_ptr<AlarmState> state{
>       new AlarmState{cq_, period, std::move(on_alarm)}};
> ```

since you added a constructor, you can just say:

`auto state = std::make_shared<AlarmState>(...);`



_Code quote:_
```cpp
  std::shared_ptr<AlarmState> state{
      new AlarmState{cq_, period, std::move(on_alarm)}};
```
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 71 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-MysZ4PX5Lj5C0FN-Fws:-MysZ4PX5Lj5C0FN-Fwt:b4bsm9z) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/57a01d5cf4ec006f112f4b6037059e892ff5e91d/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L71)):*
> ```cpp
>   InSequence seq;
> 
>   promise<StatusOr<std::chrono::system_clock::time_point>> timer;
> ```

please call these `timer1` and `timer2` instead of `timer` and `timer1`



_Code quote:_
```cpp
  promise<StatusOr<std::chrono::system_clock::time_point>> timer;
  EXPECT_CALL(*cq_, MakeRelativeTimer(std::chrono::nanoseconds(kAlarmPeriod_)))
      .WillOnce(Return(ByMove(timer.get_future())));
  auto token = alarm_.RegisterAlarm(kAlarmPeriod_, fun_.AsStdFunction());

  EXPECT_CALL(fun_, Call);

  promise<StatusOr<std::chrono::system_clock::time_point>> timer1;
```
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 138 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-MysZQf52G9GAoyDQktg:-MysZQf52G9GAoyDQkth:bt09qsh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/57a01d5cf4ec006f112f4b6037059e892ff5e91d/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L138)):*
> ```cpp
>   promise<StatusOr<std::chrono::system_clock::time_point>> timer2;
> 
>   std::thread first{[&in_alarm_function, &destroy_finished, &token]() {
> ```

here and elsewhere: you can just use `[&] { ... }` for lambdas in unit tests



_Code quote:_
```cpp
&in_alarm_function, &destroy_finished, &token
```
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 138 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-MysZY9CDdvlbklqKoT3:-MysZY9CDdvlbklqKoT4:b1du9cf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/57a01d5cf4ec006f112f4b6037059e892ff5e91d/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L138)):*
> ```cpp
>   promise<StatusOr<std::chrono::system_clock::time_point>> timer2;
> 
>   std::thread first{[&in_alarm_function, &destroy_finished, &token]() {
> ```

nit. please move this above creation of timer2



_Code quote:_
```cpp
std::thread first{[&in_alarm_function, &destroy_finished, &token]() {
```

coryan (changes requested):
> Reviewed 1 of 5 files at r2, 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575)* status: all files reviewed, 10 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/alarm_registry_impl.h, line 29 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0mnaP8OMbMJW3y2Du:-Mz0mnaP8OMbMJW3y2Dv:bn32kep) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl.h#L29)):*
> ```objc
> namespace pubsublite_internal {
> 
> // TODO(18suresha): description
> ```

Please add the description.
___
*[google/cloud/pubsublite/internal/alarm_registry_impl.h, line 32 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0nLynCzdi59lSBHWN:-Mz0nLynCzdi59lSBHWO:b5ascl7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl.h#L32)):*
> ```objc
> class AlarmRegistryImpl : public AlarmRegistry {
>  public:
>   explicit AlarmRegistryImpl(google::cloud::CompletionQueue cq);
> ```

On naming, this class (and its interface) does not look like an "alarm" to me.  It seems to be a "periodic timer", or "periodic timer queue".
___
*[google/cloud/pubsublite/internal/alarm_registry_impl.cc, line 58 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0nxnL9jRSbL5bjf_Z:-Mz0nxnL9jRSbL5bjf__:b60tqur) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl.cc#L58)):*
> ```cpp
>   // run after the destructor finishes
>   std::lock_guard<std::mutex> g{state_->mu};
>   state_->shutdown = true;
> ```

While this will shutdown the alarm "eventually" that might be long (if the period is long).  If you wanted to shutdown the alarm quicker you should need to add  a `future<...> timer_` to `state_`, initialize it with the value returned by `cq.MakeRelativeTimer()`, and then do something like `state_->timer.cancel()`.
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 71 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-MysZ4PX5Lj5C0FN-Fws:-Mz0rjsW-qDcC8sU7FmJ:b-nlk1yx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/57a01d5cf4ec006f112f4b6037059e892ff5e91d/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L71)):*
<details><summary><i>Previously, dpcollins-google wrote…</i></summary><blockquote>

please call these `timer1` and `timer2` instead of `timer` and `timer1`
</blockquote></details>

See below, these could be simplified (or at least made to look more like other timers with asynchronous stuff) using an `AsyncSequencer`.
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 138 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-MysZQf52G9GAoyDQktg:-Mz0oegt5AP4zCC6beOD:bdf20fo) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/57a01d5cf4ec006f112f4b6037059e892ff5e91d/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L138)):*
<details><summary><i>Previously, dpcollins-google wrote…</i></summary><blockquote>

here and elsewhere: you can just use `[&] { ... }` for lambdas in unit tests
</blockquote></details>

Please address this comment.
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 48 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0ouH427po2mhjBnNm:-Mz0ouH427po2mhjBnNn:b2j4rbr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L48)):*
> ```cpp
>   StrictMock<MockFunction<void()>> fun_;
>   // 10,000 seconds to ensure this test has no time dependence
>   std::chrono::milliseconds const kAlarmPeriod_{10000 * 1000};
> ```

let the computer deal with the units: `std::chrono::milliseconds(10000 * std::chrono::seconds(1))`

if this is really a constant for the file, there is no need to make it a member variable?
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 91 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0pV_c4sXf9zJmrk0x:-Mz0pV_c4sXf9zJmrk0y:bqspxbk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L91)):*
> ```cpp
>   InSequence seq;
> 
>   promise<StatusOr<std::chrono::system_clock::time_point>> timer;
> ```

Most of these tests could be simplified if you used `AsyncSequencer`

https://github.com/googleapis/google-cloud-cpp/blob/2688fd9b5b1ffc3daf4c66192c9d2251284d9385/google/cloud/testing_util/async_sequencer.h#L65-L66



You could write:

```cc
google::cloud::testing_util::AsyncSequencer<void> sequencer;
EXPECT_CALL(*cq_, MakeRelativeTimer(std::chrono::nanoseconds(kAlarmPeriod_)).Times(5).WillRepeatedly(
    [&sequencer](std::chrono::nanoseconds) { return sequencer.PushBack().then([](future<void>) { return make_status_or(std::chrono::system_clock::time_point{}); });

auto token = alarm_.RegisterAlarm(kAlarmPeriod_, fun_.AsStdFunction());

::testing::InSequence seq; // Note this got moved.
for (int i = 0; i != 5; ++i) {
  EXPECT_CALL(fun_, Call);
  sequencer.PopFront().set_value();
}  
```

dpcollins-google:
> While this will shutdown the alarm "eventually" that might be long (if the period is long). If you wanted to shutdown the alarm quicker you should need to add a future<...> timer_ to state_, initialize it with the value returned by cq.MakeRelativeTimer(), and then do something like state_->timer.cancel().

This code actually fences future runs- since the lock is held while the callback is called, you have the guarantee by the time the destructor runs to completion that 1) the callback is not running (since then we wouldnt be able to get the lock) and 2) the callback will never run again (since we check the `shutdown` variable before running the callback). Yes there is an outstanding alarm on the completion queue, but it will be a no-op when it does run.

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575)* status: 4 of 7 files reviewed, 10 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/alarm_registry_impl.h, line 32 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0nLynCzdi59lSBHWN:-Mz1nzE_161vkfDv5iDE:bahepws) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl.h#L32)):*

On naming, this class (and its interface) does not look like an "alarm" to me.  It seems to be a "periodic timer", or "periodic timer queue".
</blockquote></details>

I think 'alarm' is suitable since something has to "ring" for something (the `on_alarm` function) to occur. I understand that an 'alarm' doesn't necessarily ring periodically, though I would prefer to keep the naming the same to avoid refactoring previously merged code and code currently in review.

If need be, I can of course rename and refactor.
___
*[google/cloud/pubsublite/internal/alarm_registry_impl.cc, line 58 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0nxnL9jRSbL5bjf_Z:-Mz1oZTVEd-A4_Drje9t:bzbt6wk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl.cc#L58)):*

While this will shutdown the alarm "eventually" that might be long (if the period is long).  If you wanted to shutdown the alarm quicker you should need to add  a `future<...> timer_` to `state_`, initialize it with the value returned by `cq.MakeRelativeTimer()`, and then do something like `state_->timer.cancel()`.
</blockquote></details>

I added some code to account for this, is it along the lines of what you were thinking?
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 71 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-MysZ4PX5Lj5C0FN-Fws:-Mz1ooBH1fcbrF78fPg7:bs3nvtd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/57a01d5cf4ec006f112f4b6037059e892ff5e91d/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L71)):*

See below, these could be simplified (or at least made to look more like other timers with asynchronous stuff) using an `AsyncSequencer`.
</blockquote></details>

I used `AsyncSequencer<StatusOr<std::chrono::system_clock::time_point>>` because one of the test cases sets the value to an error.

coryan (changes requested):
> Reviewed 1 of 2 files at r13, 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575)* status: all files reviewed, 8 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/alarm_registry_impl.h, line 32 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0nLynCzdi59lSBHWN:-Mz26tk293RiVgRlP99A:b61ht7h) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl.h#L32)):*

I think 'alarm' is suitable since something has to "ring" for something (the `on_alarm` function) to occur. I understand that an 'alarm' doesn't necessarily ring periodically, though I would prefer to keep the naming the same to avoid refactoring previously merged code and code currently in review.

If need be, I can of course rename and refactor.
</blockquote></details>

Please add a bug so we remember to rename this later.
___
*[google/cloud/pubsublite/internal/alarm_registry_impl.cc, line 58 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0nxnL9jRSbL5bjf_Z:-Mz27tWF-FIU_Z_lasYp:bgvadzz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl.cc#L58)):*

I added some code to account for this, is it along the lines of what you were thinking?
</blockquote></details>

Yes, thanks.  Needs a few tweaks, but generally in the right direction.
___
*[google/cloud/pubsublite/internal/alarm_registry_impl.cc, line 44 at r14](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz27Jqz-soRsryvNMS_:-Mz27Jqz-soRsryvNMSa:bryzq51) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/98cd8ef4207cf4ce3190d168c8a52dce0ff22704/google/cloud/pubsublite/internal/alarm_registry_impl.cc#L44)):*
> ```cpp
>               std::lock_guard<std::mutex> g{state->mu};
>               if (state->shutdown) return;
>               state->on_alarm();
> ```

I should have noticed this before.  Calling external code while holding a mutex is a recipe for a deadlock.  Is it possible to move the `on_alarm()` call outside?
___
*[google/cloud/pubsublite/internal/alarm_registry_impl.cc, line 49 at r14](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz279ug2Huzy4iO8M2b:-Mz279ug2Huzy4iO8M2c:b-jvy4it) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/98cd8ef4207cf4ce3190d168c8a52dce0ff22704/google/cloud/pubsublite/internal/alarm_registry_impl.cc#L49)):*
> ```cpp
>           });
>   std::lock_guard<std::mutex> g{state->mu};
>   state->timer = std::move(temp);
> ```

I think you may want to check `state_->shutdown` something like:

```cc
std::lock_guard<std::mutex> g{state->mu};
if (!state_->shutdown) {
  state->timer = std::move(temp);
}
temp.cancel();
```
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 71 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-MysZ4PX5Lj5C0FN-Fws:-Mz28Bl_3f3Lx0td-wpz:br9zmg7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/57a01d5cf4ec006f112f4b6037059e892ff5e91d/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L71)):*

I used `AsyncSequencer<StatusOr<std::chrono::system_clock::time_point>>` because one of the test cases sets the value to an error.
</blockquote></details>

Your call: some tests may be simpler with `AsyncSequencer<void>`.  OTOH, I can see how some would be simpler with your choice, and I can see how using the same `AsyncSequencer<>` in all tests would be easier to grok too.
___
*[google/cloud/pubsublite/internal/alarm_registry_impl_test.cc, line 42 at r14](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz282j41nxPH1YSZTph:-Mz282j41nxPH1YSZTpi:bvebk4z) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/98cd8ef4207cf4ce3190d168c8a52dce0ff22704/google/cloud/pubsublite/internal/alarm_registry_impl_test.cc#L42)):*
> ```cpp
> 
> // 10,000 seconds to ensure this test has no time dependence
> std::chrono::milliseconds constexpr kAlarmPeriod{10000 *
> ```

nit: generally we prefer to write:

```cc
auto constexpr kAlarmPeriod = std::chrono::milliseconds{...
```

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575)* status: 5 of 7 files reviewed, 8 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/alarm_registry_impl.h, line 32 at r11](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz0nLynCzdi59lSBHWN:-Mz2UV3H4OXofoIvodet:b-vrsmfg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4ecfca07c989cc6782122ff3b4862c91283f6d3e/google/cloud/pubsublite/internal/alarm_registry_impl.h#L32)):*

Please add a bug so we remember to rename this later.
</blockquote></details>

https://github.com/googleapis/google-cloud-cpp/issues/8601
___
*[google/cloud/pubsublite/internal/alarm_registry_impl.cc, line 44 at r14](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz27Jqz-soRsryvNMS_:-Mz2UWt87RZmeGrcJu81:bpvhy3k) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/98cd8ef4207cf4ce3190d168c8a52dce0ff22704/google/cloud/pubsublite/internal/alarm_registry_impl.cc#L44)):*

I should have noticed this before.  Calling external code while holding a mutex is a recipe for a deadlock.  Is it possible to move the `on_alarm()` call outside?
</blockquote></details>

hm but won't that only occur if the destructor for `CancelToken` is called in `on_alarm()`'s downcall?
We need `on_alarm` to be under lock, so that the destructor is only run when `on_alarm()` isn't running.
Additionally, `~CancelToken` won't be called in `on_alarm()`'s downcall.

coryan (changes requested):
> Reviewed 2 of 2 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575)* status: all files reviewed, 6 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/alarm_registry_impl.cc, line 44 at r14](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz27Jqz-soRsryvNMS_:-Mz5Y9BYCVAtRXrdGWfz:b-3qkih8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/98cd8ef4207cf4ce3190d168c8a52dce0ff22704/google/cloud/pubsublite/internal/alarm_registry_impl.cc#L44)):*
> hm but won't that only occur of the destructor for CancelToken is called in on_alarm()'s downcall?

 "that" is ambiguous, sorry, I am not sure what you are referring to.  

I am seeing a mutex held and a call to external code.  That is always a code smell.  I am suggesting there is a simple way to make it easier to reason about the code.

> We need on_alarm to be under lock, so that the destructor is only run when on_alarm() isn't running.

The lambda has a copy of `state`, the destructor of the lambda cannot be called while its `operator()` is executing (having the destructor and `operator()` running at the same time would already be undefined behavior).  Therefore, the destructor of `on_alarm` cannot be called either until the lambda returns.

> Additionally, ~CancelToken won't be called in on_alarm()'s downcall.

I am not sure why that matters, but it seems that it could happen:

```cc
void F(AlarmRegistry& r) {
  struct Foo {
    std::unique_ptr<AlarmRegister::CancelToken> token;
  };
  auto evil = std::make_shared<Foo>();
  evil->token = r.RegisterAlarm(std::chrono::milliseconds(10), [evil] { token.reset(); });
}
```

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575)* status: 6 of 7 files reviewed, 6 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[google/cloud/pubsublite/internal/alarm_registry_impl.cc, line 44 at r14](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575#-Mz27Jqz-soRsryvNMS_:-Mz6-44aCaSBNCLzeo2r:b-4q1voj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/98cd8ef4207cf4ce3190d168c8a52dce0ff22704/google/cloud/pubsublite/internal/alarm_registry_impl.cc#L44)):*

> hm but won't that only occur of the destructor for CancelToken is called in on_alarm()'s downcall?

 "that" is ambiguous, sorry, I am not sure what you are referring to.  

I am seeing a mutex held and a call to external code.  That is always a code smell.  I am suggesting there is a simple way to make it easier to reason about the code.

> We need on_alarm to be under lock, so that the destructor is only run when on_alarm() isn't running.

The lambda has a copy of `state`, the destructor of the lambda cannot be called while its `operator()` is executing (having the destructor and `operator()` running at the same time would already be undefined behavior).  Therefore, the destructor of `on_alarm` cannot be called either until the lambda returns.

> Additionally, ~CancelToken won't be called in on_alarm()'s downcall.

I am not sure why that matters, but it seems that it could happen:

```cc
void F(AlarmRegistry& r) {
  struct Foo {
    std::unique_ptr<AlarmRegister::CancelToken> token;
  };
  auto evil = std::make_shared<Foo>();
  evil->token = r.RegisterAlarm(std::chrono::milliseconds(10), [evil] { token.reset(); });
}
```

</blockquote></details>

by 'that' I meant deadlock

I think there's a misunderstanding? By 'destructor', I meant the destructor of the `CancelToken` returned by `RegisterAlarm`. The semantics of the `AlarmRegistry` interface state, 'When CancelToken is destroyed, the alarm will not be running and will never run again.' The reason being that if there was a chance that `on_alarm` was in the middle of its final run after `~CancelToken` is finished, the user has no idea when `on_alarm` will finish and cannot run code safely under the assumption that `on_alarm` is done. With `AlarmRegistryImpl`, the way we achieve that constraint is that we want `on_alarm` to be invoked/running only when its respective `AlarmState` is not shutdown. Since, the `AlarmState` can only be set to shutdown via `CancelToken`'s destructor, we need to guarantee that the downcall of `~CancelToken` is only ran when `on_alarm` isn't running. This is done by guarding `on_alarm` and the corresponding `AlarmState` under the same mutex.

I added a description comment to `AlarmRegistryImpl::RegisterAlarm`: 'The returned `CancelToken` should not be destructed in `on_alarm()`'s downcall.' This is a property of the use cases we had in mind for this class, and since this is an internal class, we can ensure that we follow this design.

coryan (approved):
> Reviewed 1 of 1 files at r16, 3 of 3 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8575)* status: all files reviewed, 5 unresolved discussions (waiting on @18suresha and @dpcollins-google)

PR #8580, opened 2022-03-22
cleanup!: deprecate public spanner bazel targets

devjgm:
Part of #3701

Marks all Spanner BUILD targets as private by default. `cc_test` targets
that were previously public are now private. The
`//google/cloud/spanner:spanner_client_testing` target was never
intended to be public and is now deprecated to be deleted in about a
year.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8580)


google/cloud/spanner/BUILD.bazel
```diff
@@ -77,6 +77,18 @@ cc_library(
     ],
 )
 
+# TODO(#3701): Delete this target after 2023-04-01.
```
devbww, line 80: Hopefully what we'll also do is rename `spanner_client_testing_private` back to `spanner_client_testing`.

devbww (approved):

google/cloud/spanner/BUILD.bazel
devjgm: Agreed, but I didn't want to capture all those details in the TODO.

PR #8581, opened 2022-03-22
cleanup!: deprecate public bigtable bazel targets

devjgm:
Part of #3701

Marks all Bigtable BUILD targets as private by default. cc_test targets
that were previously public are now private. The
//google/cloud/bigtable:bigtable_client_testing target was never
intended to be public and is now deprecated to be deleted in about a
year.


dbolduc (approved):

devjgm:
Updated this to keep `bigtable_client_testing` as a public target thanks to @devbww 's eagle eyes. PTAL

coryan (approved):

google/cloud/bigtable/BUILD.bazel
```diff
@@ -60,12 +60,10 @@ cc_library(
...
 
 cc_library(
-    name = "bigtable_client_testing_private",
+    name = "bigtable_client_testing",
     srcs = bigtable_client_testing_srcs,
     hdrs = bigtable_client_testing_hdrs,
-    visibility = [
-        ":__subpackages__",
-    ],
+    visibility = ["//visibility:public"],
```
dbolduc, line 66: Thanks for correcting this.

PR #8584, opened 2022-03-23
cleanup!: restrict testing_util build visibility

devjgm:
Part of #3701

Change the default build visibility to private. Expose the previous targets as public but deprecated.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8584)


coryan (commented):
I would have expected these needed the `*_internal` + alias treatment too?

devjgm:
> I would have expected these needed the `*_internal` + alias treatment too?

I can add that later today. I didn't bother originally because https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/testing_util/README.md says

> The functions and types provided here are **not intended for general use, and are subject to change without notice.**

coryan (approved):
I had forgotten that from the README, SGTM.

devjgm:
> I had forgotten that from the README, SGTM.

Sorry, too late :-) Exposed publicly and deprecated. We can proceed either way with minimal effort. Do you have a preference how you'd like to proceed?

coryan (approved):

PR #8585, opened 2022-03-23
cleanup!: deprecate //google/cloud:{common,grpc_utils} targets

devjgm:
Part of #3701

Changes the default build visibility for `google/cloud/BUILD.bazel` to
be private. Deprecates the `//google/cloud` targets
`:google_cloud_cpp_common` and `:google_cloud_cpp_common_grpc_utils`
directding users to their top-level targets in `//:` instead.

Updates all of our code to use the undeprecated top-level targets.

Note to reviewers: The only manual edits were to:
* BUILD.bazel
* google/cloud/BUILD.bazel

The rest were generated w/ `perl -p -i -e ...`

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8585)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -249,13 +249,13 @@ cc_library(
 cc_library(
     name = "common",
     deps = [
-        "//google/cloud:google_cloud_cpp_common",
+        "//google/cloud:google_cloud_cpp_common_private",
     ],
 )
 
 cc_library(
     name = "grpc_utils",
     deps = [
-        "//google/cloud:google_cloud_cpp_grpc_utils",
+        "//google/cloud:google_cloud_cpp_grpc_utils_private",
     ],
 )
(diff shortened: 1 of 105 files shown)


coryan (approved):
Probably need to fix the generator for a lot of these files too:

https://github.com/googleapis/google-cloud-cpp/blob/5600e7bba791e1e3c964b97f26a3f31c672f76a9/generator/internal/scaffold_generator.cc#L365-L366

devjgm:
> Probably need to fix the generator for a lot of these files too:
> 
> https://github.com/googleapis/google-cloud-cpp/blob/5600e7bba791e1e3c964b97f26a3f31c672f76a9/generator/internal/scaffold_generator.cc#L365-L366

Good catch. Thanks. Fixed. Re-testing.

PR #8597, opened 2022-03-25
refactor: move sha-256 algorithm to `google::cloud::internal`

18suresha:
Pub/Sub Lite will use the algorithm for its routing policy, so moving the file into `google::cloud::internal` from `google::cloud::storage;:internal`


coryan (changes requested):
The additional dependency on `OpenSSL::Crypto` (aka `@boringssl//:crypto`) gave me some pause.  I think it is fine since anything using `google_cloud_cpp_common` probably needs that dependency too (via gRPC or CURL).  Introducing `absl::Span` also gave me pause, but for an internal API that seems fine.

There are some changes requested below, I think they should be fast to address.

> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8597#-:-Mz17UX9BYexi3awNAnv:b-v254bq)* status: all files reviewed, 4 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/CMakeLists.txt, line 218 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8597#-Mz13C4U0B7GpYmgtur2:-Mz13C4U0B7GpYmgtur3:b-3z3nbw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1801ff981f88f4829a7b28e81ae35fa032f63140/google/cloud/pubsublite/CMakeLists.txt#L218)):*
> ```CMake
>     # the GTest::gmock target, and the target names are also weird.
>     find_package(GTest CONFIG REQUIRED)
> ```

This change is unrelated?  Please remove.
___
*[google/cloud/storage/oauth2/service_account_credentials.h, line 24 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8597#-Mz14GRlEWH4BkcpFtMv:-Mz14GRlEWH4BkcpFtMw:b-c3t2gh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6f8f957d5e610bb65113d6ebd43fccb7a914b9d8/google/cloud/storage/oauth2/service_account_credentials.h#L24)):*
> ```objc
> #include "google/cloud/storage/oauth2/refreshing_credentials_wrapper.h"
> #include "google/cloud/storage/version.h"
> #include "google/cloud/internal/sha256_hash.h"
> ```

Looks like this is unused?  Maybe move it to the `.cc` file if it is used there, or remove it altogether.
___
*[google/cloud/internal/sha256_hash.h, line 31 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8597#-Mz15JCu6WPHZgMyTi9k:-Mz15JCu6WPHZgMyTi9l:bjowsmm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6f8f957d5e610bb65113d6ebd43fccb7a914b9d8/google/cloud/internal/sha256_hash.h#L31)):*
> ```objc
> 
> /// Return the SHA256 hash (as raw bytes) of @p str.
> std::array<std::uint8_t, 32> Sha256Hash(std::string const& str);
> ```

I know where the 32 comes from, but a future reader might not.  This suggests we should have a type alias and/or a comment explaining where the magic comes from.
___
*[google/cloud/internal/sha256_hash.h, line 37 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8597#-Mz14u5A1EEIJhNDv3jf:-Mz14u5A1EEIJhNDv3jg:b-r2dokj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6f8f957d5e610bb65113d6ebd43fccb7a914b9d8/google/cloud/internal/sha256_hash.h#L37)):*
> ```objc
> 
> /// Return @p bytes encoded as a lowercase hexadecimal string.
> std::string HexEncode(absl::Span<std::uint8_t const> const& bytes);
> ```

Just `absl::Span<std::uint8_t const> bytes`.

coryan:
___
*[google/cloud/storage/oauth2/service_account_credentials.h, line 24 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8597#-Mz14GRlEWH4BkcpFtMv:-Mz1EAr0CJMcItggVD3B:b-givrrf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6f8f957d5e610bb65113d6ebd43fccb7a914b9d8/google/cloud/storage/oauth2/service_account_credentials.h#L24)):*

hm when I remove it from the header and add it to the source file, the tests don't build
</blockquote></details>

Leaving it alone, I will clean it up later.

coryan (approved):
> Reviewed 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8597)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dpcollins-google)

dpcollins-google (approved):

google/cloud/storage/internal/openssl_util.cc
```diff
@@ -54,7 +54,7 @@ std::string Base64Encode(std::string const& str) {
   return std::move(enc).FlushAndPad();
 }
 
-std::string Base64Encode(std::vector<std::uint8_t> const& bytes) {
+std::string Base64Encode(absl::Span<std::uint8_t const> const& bytes) {
```
dpcollins-google, line 57: any parameters of absl::Span type should be accepted by value not reference

PR #8602, opened 2022-03-26
cleanup(spanner): tweak spanner convenience namespace aliases

devbww:
Move the proto version outside the `spanner_proto` and `gcsa`
namespace aliases used in the implementation.  This means that
these Spanner proto names won't require special transformation
rules when imported into an environment that uses a different
gRPC service namespace.

Also remove the `giam` alias altogether, as it wasn't pulling
its weight.

I contemplated removing the `gcsa` alias too, partly because
it has different definitions in different files (or to at least
move the `database` and `instance` parts outside the aliases),
but that can wait for another rainy day.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8602)

### DIFF
--- a/google/cloud/spanner/client_test.cc
+++ b/google/cloud/spanner/client_test.cc
@@ -38,7 +38,7 @@ namespace spanner {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace {
 
-namespace spanner_proto = ::google::spanner::v1;
+namespace spanner_proto = ::google::spanner;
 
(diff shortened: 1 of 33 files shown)


coryan (approved):
LGTM.  I will just note that `gcsa` vs. `google::spanner::admin::blah` saves some wear and tear on the keyboard (and the typist), but `google::spanner` vs. `spanner_proto` is less of a win.

devbww:
> `google::spanner` vs. `spanner_proto` is less of a win.

Indeed.

Issue #8613, opened 2022-03-29
Build wtih Bazel fails on Windows 10


Nogbit:
**Does this issue affect the google-cloud-cpp project?**
Yes

**What component of `google-cloud-cpp` is this related to?**
Powershell ... `bazel build //...`

**Describe the bug**
The build was running fine for a very long time and then fails. 

```
INFO: From Compiling upb/upb.c:
external/upb/upb/upb.c(223): warning C4116: unnamed type definition in parentheses
ERROR: C:/users/me/documents/stuff/gcpcpp/google-cloud-cpp/google/cloud/BUILD.bazel:19:8: Executing genrule //google/cloud:generate_build_info failed: (Exit 127): bash.exe failed: error executing command C:\WINDOWS\system32\bash.exe -c ... (remaining 1 argument skipped)
/bin/bash: $'\r': command not found
INFO: Elapsed time: 1153.890s, Critical Path: 34.10s
INFO: 2310 processes: 319 internal, 1991 local.
FAILED: Build did NOT complete successfully
```

**To Reproduce**
1. Clone this repo, switch to branch  `v1.38.x`
2. Start powershell
3. `cd` to this clone
4. ` bazel build //...`
5. See error

**Expected behavior**
Build should not error

**Operating system:**
Windows 10

**What compiler and version are you using?**
bazel

**What version of `google-cloud-cpp` are you using?**
`643fa047a549422ebaa40f9173ab0c2c8372aede`

Nogbit:
I get the same when doing a much simpler build

`cd google\cloud\automl` and then `bazel build ...`

```
INFO: From Compiling src/core/ext/upbdefs-generated/envoy/config/bootstrap/v3/bootstrap.upbdefs.c:
C:\users\marc\_bazel_marc\gg6hr4cz\execroot\com_github_googleapis_google_cloud_cpp\external\upb\upb/msg_internal.h(741): warning C4090: 'function': different 'const' qualifiers
ERROR: C:/users/marc/documents/gamesoft/gcpcpp/google-cloud-cpp2/google-cloud-cpp2/google/cloud/BUILD.bazel:19:8: Executing genrule //google/cloud:generate_build_info failed: (Exit 127): bash.exe failed: error executing command C:\WINDOWS\system32\bash.exe -c ... (remaining 1 argument skipped)
/bin/bash: $'\r': command not found
INFO: Elapsed time: 113.978s, Critical Path: 8.35s
[...]

PR #8614, opened 2022-03-29
cleanup: force LF line endings on git checkout

dbolduc:
Seems like Windows might automatically convert line endings from LF to CRLF, depending on the `core.autocrlf` git setting. (See: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration)

I followed instructions [here](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings) to add a `.gitattributes` file to prevent this from happening. (I think...)

Inspired by #8613

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8614)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Force the use of Unix LF style line endings
+* text eol=lf
+
+# Our ABI dumps should not be modified. Mark them as binary.
+*.gz binary


.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Force the use of Unix LF style line endings
+* text eol=lf
```
coryan, line 2: I am 99.94% sure this is not what we want.  On Windows CRLF is the native EOL, if you edit a file on Windows that is what the editor expects to see, and what it will (most likely) save.

dbolduc:
that's a monkey for me

PR #8617, opened 2022-03-29
feat(spanner): add initial PostgreSQL samples

devbww:
These samples are sufficiently different from the "Google Standard"
dialect ones that they deserve their own source file (not to mention
that `samples.cc` is becoming unmanageable).  If there ends up being
any significant commonality between the samples implementations, we
can factor it out later.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8617)

### DIFF
--- a/google/cloud/spanner/samples/CMakeLists.txt
+++ b/google/cloud/spanner/samples/CMakeLists.txt
@@ -16,7 +16,7 @@
 
 function (spanner_client_define_samples)
     set(spanner_client_integration_samples # cmake-format: sort
-                                           samples.cc)
+                                           pg_samples.cc samples.cc)
     set(spanner_client_unit_samples # cmake-format: sort
                                     mock_execute_query.cc)
 
--- a/google/cloud/spanner/samples/CMakeLists.txt
+++ b/google/cloud/spanner/samples/CMakeLists.txt
@@ -16,7 +16,7 @@
 
 function (spanner_client_define_samples)
     set(spanner_client_integration_samples # cmake-format: sort
-                                           pg_samples.cc samples.cc)
+                                           postgresql_samples.cc samples.cc)
     set(spanner_client_unit_samples # cmake-format: sort
                                     mock_execute_query.cc)
 
--- a/google/cloud/spanner/samples/pg_samples.cc
+++ b/google/cloud/spanner/samples/pg_samples.cc
@@ -1,0 +1,371 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/spanner/admin/database_admin_client.h"
+#include "google/cloud/spanner/client.h"
+#include "google/cloud/spanner/database.h"
+#include "google/cloud/spanner/instance.h"
+#include "google/cloud/spanner/testing/pick_random_instance.h"
+#include "google/cloud/spanner/testing/random_database_name.h"
+#include "google/cloud/internal/getenv.h"
+#include "google/cloud/internal/random.h"
+#include "google/cloud/log.h"
+#include "absl/time/clock.h"
+#include "absl/time/time.h"
+#include <cstdint>
+#include <exception>
+#include <functional>
+#include <map>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace {
+namespace samples {
+
+// [START spanner_postgresql_create_database]
+void CreateDatabase(google::cloud::spanner_admin::DatabaseAdminClient client,
+                    google::cloud::spanner::Database const& database) {
+  google::spanner::admin::database::v1::CreateDatabaseRequest request;
+  request.set_parent(database.instance().FullName());
+  request.set_create_statement("CREATE DATABASE " + database.database_id());
+  request.set_database_dialect(
+      google::spanner::admin::database::v1::DatabaseDialect::POSTGRESQL);
+  auto db = client.CreateDatabase(request).get();
+  if (!db) throw std::runtime_error(db.status().message());
+  std::cout << "Database " << db->name() << " created.\n";
+}
+// [END spanner_postgresql_create_database]
+
+void CreateTables(google::cloud::spanner_admin::DatabaseAdminClient client,
+                  google::cloud::spanner::Database const& database) {
+  std::vector<std::string> statements = {
+      R"""(
+        CREATE TABLE Singers (
+            SingerId   BIGINT NOT NULL,
+            FirstName  CHARACTER VARYING(1024),
+            LastName   CHARACTER VARYING(1024),
+            singerInfo BYTEA,
+            PRIMARY KEY(singerid)
+        )
+      )""",
+      R"""(
+        CREATE TABLE Albums (
+            AlbumId    BIGINT NOT NULL,
+            SingerId   BIGINT NOT NULL,
+            AlbumTitle CHARACTER VARYING,
+            PRIMARY KEY(SingerId, AlbumId),
+            FOREIGN KEY(SingerId) REFERENCES Singers(SingerId)
+        )
+      )""",
+  };
+  auto metadata =
+      client.UpdateDatabaseDdl(database.FullName(), statements).get();
+  if (!metadata) throw std::runtime_error(metadata.status().message());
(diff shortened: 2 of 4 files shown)


google/cloud/spanner/samples/pg_samples.cc
```diff
@@ -0,0 +1,371 @@
...
+  };
+  auto it = commands.find(argv[0]);
+  if (it == commands.end()) {
+    throw std::runtime_error(argv[0] + ": Unknown command");
+  }
+  it->second(std::move(argv));
+  return 0;
+}
+
+void SampleBanner(std::string const& name) {
```
dbolduc, line 269: We *could* factor this out as it appears in `samples.cc` as well. (We could also refactor the `Command`/`make_command_entry` stuff too). Probably not something we want to do in this PR. Maybe not something we ever want to do.

google/cloud/spanner/samples/pg_samples.cc
```diff
@@ -0,0 +1,371 @@
...
+int main(int ac, char* av[]) try {
+  const auto* const emulator_host = "SPANNER_EMULATOR_HOST";
+  if (google::cloud::internal::GetEnv(emulator_host).has_value()) {
+    return 0;  // emulator does not support PostgreSQL dialect
+  }
+  const auto* const auto_run = "GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES";
+  if (google::cloud::internal::GetEnv(auto_run).value_or("") == "yes") {
+    return RunAll();
+  }
+  std::string program(ac ? (ac--, *av++) : "pg_samples");
```
dbolduc, line 361: TIL `argc` can be 0

google/cloud/spanner/samples/pg_samples.cc
```diff
@@ -0,0 +1,371 @@
...
+      {"add-column", Command(samples::AddColumn)},
+      {"create-storing-index", Command(samples::CreateStoringIndex)},
+      {"insert-data", Command(samples::InsertData)},
+      {"query-with-parameter", Command(samples::QueryWithParameter)},
+      {"dml-getting-started-update", Command(samples::DmlGettingStartedUpdate)},
+      {"drop-database", Command(samples::DropDatabase)},
+  };
+  auto it = commands.find(argv[0]);
+  if (it == commands.end()) {
+    throw std::runtime_error(argv[0] + ": Unknown command");
```
dbolduc, line 263: So how can the user learn which commands are available? Well they can look in the code, but I don't think they should have to. Can we print all commands here?

google/cloud/spanner/samples/pg_samples.cc
```diff
@@ -0,0 +1,371 @@
...
+  }
+  const auto* const auto_run = "GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES";
+  if (google::cloud::internal::GetEnv(auto_run).value_or("") == "yes") {
+    return RunAll();
+  }
+  std::string program(ac ? (ac--, *av++) : "pg_samples");
+  if (ac == 0) {
+    throw std::runtime_error("Usage: " + program +
+                             " <command> [<argument> ...]");
+  }
```
dbolduc, line 365: I think we also want to print the available commands in this failure case.

dbolduc (changes requested):

google/cloud/spanner/samples/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@
 
 function (spanner_client_define_samples)
     set(spanner_client_integration_samples # cmake-format: sort
-                                           samples.cc)
+                                           pg_samples.cc samples.cc)
```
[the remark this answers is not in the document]
devbww: Meh.  OK, `postgresql_samples.cc` it is.

google/cloud/spanner/samples/pg_samples.cc
```diff
@@ -0,0 +1,371 @@
...
+    return budget ? *budget : 0;
+  };
+
+  // A helper to update the budget for the given album and singer.
+  auto update_budget = [&](google::cloud::spanner::Transaction txn,
+                           std::int64_t album_id, std::int64_t singer_id,
+                           std::int64_t budget) {
+    auto sql = google::cloud::spanner::SqlStatement(
+        "UPDATE Albums SET MarketingBudget = $1"
+        "  WHERE SingerId = $3 AND AlbumId = $2",
```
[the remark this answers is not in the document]
devbww: It doesn't look like there is a definitive source of what these samples are supposed to look like (just their names), but this one says it should be similar to the non-pg version, where these were ordered this way, although the non-pg versions uses names to bind instead of indices.  I could imagine that explicitly using out-of-order indices is a teaching point, but probably not worth it.  So, yes, swapped.

google/cloud/spanner/samples/pg_samples.cc
devbww: I added a "help" command to list all the available commands, and mention using it in this error message.

google/cloud/spanner/samples/pg_samples.cc
devbww: I added a "help" command to list all the available commands, and mention using it in this error message.

google/cloud/spanner/samples/pg_samples.cc
devbww: Yeah.  It can be instructive to try executing common programs with no arguments and seeing how many survive the ordeal.

google/cloud/spanner/samples/pg_samples.cc
devbww: I mentioned possible refactorings down the line in the original PR description, but, yes, we should certainly postpone that decision until we see where things fall.  (At the moment it looks like it would be a net complication.)

dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8617)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @dbolduc and @devbww)

___
*[google/cloud/spanner/samples/pg_samples.cc, line 177 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8617#-MzMbCCaMfKkwE7fzoo2-r1-177:-MzQ7EDK1nsS92IuEXd1:bfgb1hd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5ac6fd3fd601379ddc7c6b580a09c78a064b873c/google/cloud/spanner/samples/pg_samples.cc#L177)):*

It doesn't look like there is a definitive source of what these samples are supposed to look like (just their names), but this one says it should be similar to the non-pg version, where these were ordered this way, although the non-pg versions uses names to bind instead of indices.  I could imagine that explicitly using out-of-order indices is a teaching point, but probably not worth it.  So, yes, swapped.
</blockquote></details>

Hm, for the record I was thinking:
```
    auto sql = google::cloud::spanner::SqlStatement(
        "UPDATE Albums SET MarketingBudget = $1"
        "  WHERE SingerId = $2 AND AlbumId = $3",
        {{"p1", google::cloud::spanner::Value(budget)},
         {"p2", google::cloud::spanner::Value(singer_id)},
         {"p3", google::cloud::spanner::Value(album_id)}});
```

But this is all a nit anyway.
___
*[google/cloud/spanner/samples/pg_samples.cc, line 263 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8617#-MzMbCCaMfKkwE7fzoo2-r1-263:-MzQ7N4H3EnD8uVUscUP:bjtm9te) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5ac6fd3fd601379ddc7c6b580a09c78a064b873c/google/cloud/spanner/samples/pg_samples.cc#L263)):*

I added a "help" command to list all the available commands, and mention using it in this error message.
</blockquote></details>

works for me

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8617)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @dbolduc)

___
*[google/cloud/spanner/samples/pg_samples.cc, line 177 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8617#-MzMbCCaMfKkwE7fzoo2-r1-177:-MzQZs4kAkuOVga_BCa-:b-pwtq33) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5ac6fd3fd601379ddc7c6b580a09c78a064b873c/google/cloud/spanner/samples/pg_samples.cc#L177)):*

Hm, for the record I was thinking:
```
    auto sql = google::cloud::spanner::SqlStatement(
        "UPDATE Albums SET MarketingBudget = $1"
        "  WHERE SingerId = $2 AND AlbumId = $3",
        {{"p1", google::cloud::spanner::Value(budget)},
         {"p2", google::cloud::spanner::Value(singer_id)},
         {"p3", google::cloud::spanner::Value(album_id)}});
```

But this is all a nit anyway.
</blockquote></details>

I was going to say that it makes a little more sense this way because the `Albums` key is `(AlbumId, SingerId)` (for better or worse), but it looks like its the other way around.  I'm going to change that.

PR #8620, opened 2022-03-30
test: move SetEnv()/UnSetEnv() to the testing_util library

devbww:
We should never be modifying the environment from a service library,
so move `SetEnv()` and `UnSetEnv()` to `testing_util`, next to
`ScopedEnvironment`.

And while we're here, add warnings that these environment-modifying
functions should only be called when the process is single-threaded.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8620)


devjgm (approved):

coryan (approved):
Thanks!

PR #8623, opened 2022-03-30
test: add bazel testonly attributes to test libraries and byproducts

devbww:
Use the bazel testonly attribute to ensure that test code does not
end up in the production libraries.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8623)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -115,6 +115,7 @@ GA_LIBRARIES = [
 
 [cc_library(
     name = "experimental-{library}_mocks".format(library = library),
+    testonly = 1,
     deps = [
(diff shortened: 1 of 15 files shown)


BUILD.bazel
```diff
@@ -115,6 +115,7 @@ GA_LIBRARIES = [
 
 [cc_library(
     name = "experimental-{library}_mocks".format(library = library),
+    testonly = 1,
```
coryan, line 118: nit: should these be `True`?

coryan (approved):

BUILD.bazel
devbww: On.  Ehh, I mean Yes.  Oops, I mean True.

PR #8627, opened 2022-03-31
feat(spanner): add remaining PostgreSQL samples

devbww:
Add all the remaining PostgreSQL samples, with the exception of the
cases that add PG NUMERIC values.

Also correct the key ordering for the Albums table, and ensure we
delete the test database should an error be thrown.

### DIFF
--- a/google/cloud/spanner/samples/postgresql_samples.cc
+++ b/google/cloud/spanner/samples/postgresql_samples.cc
@@ -13,9 +13,11 @@
 // limitations under the License.
 
 #include "google/cloud/spanner/admin/database_admin_client.h"
+#include "google/cloud/spanner/bytes.h"
 #include "google/cloud/spanner/client.h"
 #include "google/cloud/spanner/database.h"
 #include "google/cloud/spanner/instance.h"
+#include "google/cloud/spanner/numeric.h"
 #include "google/cloud/spanner/testing/pick_random_instance.h"
 #include "google/cloud/spanner/testing/random_database_name.h"
 #include "google/cloud/internal/getenv.h"
@@ -34,6 +36,15 @@
 
 namespace {
 namespace samples {
+
+// [START spanner_postgresql_create_clients]
+void CreateClients(google::cloud::spanner::Database const& database) {
+  google::cloud::spanner_admin::DatabaseAdminClient database_admin_client(
+      google::cloud::spanner_admin::MakeDatabaseAdminConnection());
+  google::cloud::spanner::Client client(
+      google::cloud::spanner::MakeConnection(database));
+}
+// [END spanner_postgresql_create_clients]
 
 // [START spanner_postgresql_create_database]
 void CreateDatabase(google::cloud::spanner_admin::DatabaseAdminClient client,
@@ -57,7 +68,7 @@ void CreateTables(google::cloud::spanner_admin::DatabaseAdminClient client,
             SingerId   BIGINT NOT NULL,
             FirstName  CHARACTER VARYING(1024),
             LastName   CHARACTER VARYING(1024),
-            singerInfo BYTEA,
+            SingerInfo BYTEA,
             PRIMARY KEY(singerid)
         )
       )""",
@@ -66,8 +77,16 @@ void CreateTables(google::cloud::spanner_admin::DatabaseAdminClient client,
             AlbumId    BIGINT NOT NULL,
             SingerId   BIGINT NOT NULL,
             AlbumTitle CHARACTER VARYING,
-            PRIMARY KEY(AlbumId, SingerId),
+            PRIMARY KEY(SingerId, AlbumId),
             FOREIGN KEY(SingerId) REFERENCES Singers(SingerId)
+        )
+      )""",
+      R"""(
+        CREATE TABLE users (
+            user_id    BIGINT NOT NULL,
+            user_name  CHARACTER VARYING(1024),
+            active     BOOLEAN,
+            PRIMARY KEY(user_id)
         )
       )""",
   };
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/postgresql_samples.cc
```diff
@@ -207,6 +215,444 @@ void DmlGettingStartedUpdate(google::cloud::spanner::Client client) {
...
+          .Build();
+  auto commit =
+      client.Commit(google::cloud::spanner::Mutations{insert_singers});
+  if (!commit) throw std::runtime_error(commit.status().message());
+  std::cout << "Insert was successful.\n";
+
+  // DML statements must also follow the PostgreSQL case rules.
+  commit = client.Commit(
+      [&client](google::cloud::spanner::Transaction txn)
+          -> google::cloud::StatusOr<google::cloud::spanner::Mutations> {
```
coryan, line 322: This got me thinking, maybe we should have overloads for `client.Commit()` that takes a `std::function<Status(google::cloud::spanner::Transaction txn)>`

google/cloud/spanner/samples/postgresql_samples.cc
```diff
@@ -207,6 +215,444 @@ void DmlGettingStartedUpdate(google::cloud::spanner::Client client) {
...
+        // should verify that all statements completed successfully.
+        if (!result->status.ok()) return result->status;
+        return google::cloud::spanner::Mutations{};
+      });
+  if (!commit) throw std::runtime_error(commit.status().message());
+  std::cout << "Update was successful.\n";
+}
+// [END spanner_postgresql_batch_dml]
+
+void DropTables(google::cloud::spanner_admin::DatabaseAdminClient client,
```
coryan, line 254: Seems odd that there are no region tags or doxygen tags for this sample.  Do we really need it?

google/cloud/spanner/samples/postgresql_samples.cc
```diff
@@ -207,6 +215,444 @@ void DmlGettingStartedUpdate(google::cloud::spanner::Client client) {
...
+    std::cout << "Table: " << std::get<2>(*row)
+              << " (User defined type: " << user_defined_type << ")\n";
+  }
+}
+// [END spanner_postgresql_information_schema]
+
+// [START spanner_postgresql_numeric_data_type]
+void NumericDataType(google::cloud::spanner::Client client) {
+  /*
+   * Remaining samples to add:
```
coryan, line 614: Is this something we want to appear in the documentation / snippet?

coryan (changes requested):

google/cloud/spanner/samples/postgresql_samples.cc
devbww: We only need it to clear state for the subsequent samples, which want to recreate these tables in a different way.  Indeed, the "auto run" calls `DropTables()` twice.  There is nothing PostgreSQL specific about it though, so there is no "spanner_postgresql_*" tag defined for dropping tables.

I did make up such a tag for the "banner" line output however, but I could just as easily get rid of that, or replace it with something less tag-like.

(Some of the other languages completely create a new database, and create and populate tables, before every sample and according to its needs, and then tear the whole thing down again before starting the next sample, but that is really slow.)

google/cloud/spanner/samples/postgresql_samples.cc
devbww: They're not grabbing any of this yet, and it will be filled out before they do.

But even if they were grabbing it, I think the answer would be "yes".

google/cloud/spanner/samples/postgresql_samples.cc
```diff
@@ -260,13 +726,25 @@ CommandType HelpCommand(std::map<std::string, CommandType>& commands) {
...
+      {"create-clients", Command(samples::CreateClients)},
       {"create-database", Command(samples::CreateDatabase)},
       {"create-tables", Command(samples::CreateTables)},
       {"add-column", Command(samples::AddColumn)},
-      {"create-storing-index", Command(samples::CreateStoringIndex)},
       {"insert-data", Command(samples::InsertData)},
       {"query-with-parameter", Command(samples::QueryWithParameter)},
       {"dml-getting-started-update", Command(samples::DmlGettingStartedUpdate)},
+      {"batch-dml", Command(samples::BatchDml)},
+      {"drop-tables", Command(samples::DropTables)},
```
coryan, line 737: Your call, but if this is internal-only, maybe we can omit it here.

google/cloud/spanner/samples/postgresql_samples.cc
coryan: Ack.

google/cloud/spanner/samples/postgresql_samples.cc
coryan: It was just a brainwave. SGTM.

google/cloud/spanner/samples/postgresql_samples.cc
```diff
@@ -207,6 +215,444 @@ void DmlGettingStartedUpdate(google::cloud::spanner::Client client) {
...
+  }
+}
+// [END spanner_postgresql_information_schema]
+
+// [START spanner_postgresql_numeric_data_type]
+void NumericDataType(google::cloud::spanner::Client client) {
+  /*
+   * Remaining samples to add:
+   *   - Insert a Venue with a valid for the Revenues column.
+   *   - insert a Venue with a NULL value for the Revenues column.
```
coryan, line 616: nit: s/insert/Insert/ in this line and the next?

google/cloud/spanner/samples/postgresql_samples.cc
coryan: That seems odd.  I do not think our own TODO lists should be part of the example code we display to customers.

coryan (approved):

google/cloud/spanner/samples/postgresql_samples.cc
devbww: PS: `CreateTables()` falls into the same category.

google/cloud/spanner/samples/postgresql_samples.cc
devbww: Fair enough.  I moved the comment outside the tags.

google/cloud/spanner/samples/postgresql_samples.cc
devbww: The person exercising individual samples from the command line will still want to call `CreateTables()` and `DropTables()`, but I agree that maintaining state between samples makes that process error prone.

I'm going to move them out of the `samples` namespace, and give them non-tag-like banners.

google/cloud/spanner/samples/postgresql_samples.cc
devbww: Done.

Issue #7555, opened 2021-11-02
Move builds to Fedora:35


coryan:
Fedora:35 is out:

https://fedoraproject.org/wiki/Releases

We can now start moving the builds to this release, Fedora:34 (the version we currently use) will go EOL around 2022-05-25.

- [x] `msan` `ci/cloudbuild/dockerfiles/fedora-msan.Dockerfile`
- [x] `demo-fedora` `ci/cloudbuild/dockerfiles/demo-fedora.Dockerfile`
- [x] `cxx11`: `ci/cloudbuild/dockerfiles/fedora-34-cxx11.Dockerfile`
- [x] `asan`
- [x] `tsan`
- [x] `ubsan`
- [x] `xsan`
- [x] `check-api`
- [x] `shared`
- [x] `libcxx`
- [x] `noex`
- [x] `coverage`
- [x] `cmake-install`
- [x] `bazel-oldest`
- [x] `bazel-targets`
- [x] `gcs-grpc`
- [x] `install-gcs-grpc`
- [x] `integration-production`
- [x] `quickstart-production`
- [x] `publish-docs`
- [x] `integration-daily`
- [x] `rotate-keys`

coryan:
Note that Docker <= 20.10.9 does not work with Fedora:35:

https://medium.com/nttlabs/ubuntu-21-10-and-fedora-35-do-not-work-on-docker-20-10-9-1cd439d9921

The version of Docker on GCB is 20.10.6:

https://cloud.google.com/build/docs/release-notes

There are workarounds:

https://github.com/AkihiroSuda/clone3-workaround

I would rather wait another month or two before implementing the workarounds.

coryan:
The docker server is updated to 20.10.12:

https://console.cloud.google.com/cloud-build/builds;region=us-east1/26cbe831-7cbb-43b4-bede-4c2787b4e072;tab=detail?project=cloud-cpp-testing-resources

And I am updating the instructions to install the same version in our workstations.

PR #8670, opened 2022-04-02
ci: move `cmake-install` build to Fedora:35

coryan:
Part of the work for #7555

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8670)


devbww (approved):
I notice `integration-daily` is not in your #7555 list.  Should we edit its Cloud Build trigger too?

coryan:
> I notice `integration-daily` is not in your #7555 list. Should we edit its Cloud Build trigger too?

Yes we should. I will update the list and update the triggers once I have tested everything else.

Issue #7658, opened 2021-11-23
Long Running Operation APIs don't have a way to determine when the operation is created


dpcollins-google:
**Does this issue affect the google-cloud-cpp project?**
Yes

**What component of `google-cloud-cpp` is this related to?**
Any that use google.longrunning.Operation in their APIs.

**Describe the bug**
The purpose of long running operations is that they are long running and have two phases: submission, and completion. There is no way with the current api, which returns a future with the response_type, to determine when submission has succeeded. This means it is impossible to write a short running command line tool that would create the operation and print its identifier for later use without blocking on the result.

dpcollins-google:
The specific issue for me is with pubsublite, where the operation for SeekSubscription will never complete until the user connects a subscriber client for the first time or seeks again. This means it is impossible to write an infrastructure tool with this API that creates a subscription seeked to a predetermined location.

coryan:
It should be possible to split the LRO into a `*Client::SubmitFoo()` function, returning a `future<StatusOr<Operation>>` and a `AwaitFoo()` overload set, returning `future<StatusOr<ReturnType>>` and consuming either `StatusOr<Operation>` or `future<StatusOr<Operation>>`.

coryan:
Some ideas on this. Where today we have something like:

```cc
  future<StatusOr<google::spanner::admin::database::v1::Backup>> CreateBackup(
      google::spanner::admin::database::v1::CreateBackupRequest const& request,
      Options opts = {});
```

In the future we would have two new functions:

```cc
  future<StatusOr<google::spanner::admin::database::v1::Backup>> StartCreateBackup(
      google::spanner::admin::database::v1::CreateBackupRequest const& request,
      Options opts = {});
  future<StatusOr<google::spanner::admin::database::v1::Backup>> AwaitCreateBackup(
      google::longrunning::Operation op,
      Options opts = {});
```

And the old function would be implemented using:

```cc
  future<StatusOr<google::spanner::admin::database::v1::Backup>> CreateBackup(
      google::spanner::admin::database::v1::CreateBackupRequest const& request,
      Options opts = {}) {
    return StartCreateBackup(request, opts).then([options = std::move(opts)](auto f) mutable {
      auto op = f.get();
      if (op) return AwaitCreateBackup(*std::move(op), std::move(options));
      return make_ready_future(StatusOr<google::spanner::admin::database::v1::Backup>>(std::move(op).status()));
    });
  }
```

[...]

PR #8674, opened 2022-04-03
ci(spanner): tweak response to polling timeout in CreateBackup() test

devbww:
Reacting to a "terminated by polling policy" error is difficult at
best as the caller won't know the actual state of the operation, and
will have little they can do to regain state.  In the case at hand,
we stop waiting for the CreateBackup(), but it continues none the
less, and we no longer have a handle by which the operation can be
canceled.  Attempting to drop the database while a backup operation
is pending will fail, so don't even try.  Instead, we rely upon the
short expiration time of test backups, and the eventual cleanup of
test databases (2 days) to reclaim the lost resources.

In general, the client library probably shouldn't be synthesizing
a return status for an operation.  Rather, it should be accurately
reflecting the state of the service. Leave a comment to that effect,
which we can discuss and follow up on later.

Also, the previous increase in the "lastN" tracing to 256 was not
adequate to capture a ~100x polling loop, so bump it to 1024. There
are 6 trace events per loop, so that should be sufficient.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8674)

### DIFF
--- a/ci/etc/integration-tests-config.ps1
+++ b/ci/etc/integration-tests-config.ps1
@@ -30,7 +30,7 @@ $env:GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES="yes"
 
 # A number of options to improve logging during the CI builds. They are useful
 # when troubleshooting problems.
-$env:GOOGLE_CLOUD_CPP_EXPERIMENTAL_LOG_CONFIG="lastN,256,WARNING"
+$env:GOOGLE_CLOUD_CPP_EXPERIMENTAL_LOG_CONFIG="lastN,1024,WARNING"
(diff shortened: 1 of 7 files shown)


coryan (approved):
I suspect #7658 may be part of the solution here.  We want to start the operation, be able to asynchronously wait for it, and if that fails still be able to cancel or do something with the operation.

devbww:
> I suspect #7658 may be part of the solution here. We want to start the operation, be able to asynchronously wait for it, and if that fails still be able to cancel or do something with the operation.

Perhaps.

In the general (non-#7658) case, I think we can drop the "or do something with" part, and just make it so a polling timeout does a cancel (and then the cancellation or completion status is returned).

PR #8695, opened 2022-04-06
fix(common): revamp the async polling loop part deux

devbww:
Do not fabricate a deadline-exceeded `Status` when the polling policy
says enough is enough.  That left the operation in an unknowable state.
Rather, cancel the operation and wait for the next poll to yield us an
accurate status (which may end up being OK).

That is, a polling limit is really a request for auto-cancellation.

(See #7762 for the part-un work on ensuring an explicit cancel yielded an
accurate status.)

In the case of the Spanner backup tests, for example, this means we can do
something meaningful after hitting the poll limit of a `CreateBackup()`
call, as we'll know that the backup is no longer pending.

Update the poll-loop mocking tests to account for the new sequence of RPCs.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8695)

### DIFF
--- a/google/cloud/internal/async_polling_loop.cc
+++ b/google/cloud/internal/async_polling_loop.cc
@@ -133,23 +133,14 @@ class AsyncPollingLoopImpl
     if (op && op->done()) {
       return promise_.set_value(*std::move(op));
     }
-    // Update the polling policy even on successful requests, so we can stop
-    // after too many polling attempts.
+    // Update the polling policy even on successful requests, so we can
+    // cancel the operation after too many polling attempts.
     if (!polling_policy_->OnFailure(op.status())) {
-      if (op) {
-        // We should not be fabricating a `Status` value here. Rather, we
-        // should cancel the operation and wait for the next poll to return
-        // an accurate status to the user, otherwise they will have no idea
-        // how to react.
-        return promise_.set_value(Status(
-            StatusCode::kDeadlineExceeded,
-            location_ + "() - polling loop terminated by polling policy"));
+      if (!op) {
+        return promise_.set_value(std::move(op).status());
       }
-      return promise_.set_value(std::move(op).status());
-    }
-    if (op) {
-      std::unique_lock<std::mutex> lk(mu_);
-      op_name_ = std::move(*op->mutable_name());
+      // Consider trying to reset the backoff wait period.
+      DoCancel();
     }
     return Wait();
   }
@@ -165,7 +156,7 @@ class AsyncPollingLoopImpl
   promise<StatusOr<Operation>> promise_;
 
   // `delayed_cancel_` and `op_name_`, in contrast, are also used from
-  // `DoCancel()`, which is called asynchronously, so they need locking.
+  // `DoCancel()`, which can be called asynchronously, so they need locking.
   std::mutex mu_;
   bool delayed_cancel_;  // GUARDED_BY(mu_)
   std::string op_name_;  // GUARDED_BY(mu_)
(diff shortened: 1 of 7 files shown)


google/cloud/internal/async_polling_loop.cc
```diff
@@ -133,23 +133,14 @@ class AsyncPollingLoopImpl
...
+      if (!op) {
+        return promise_.set_value(std::move(op).status());
       }
-      return promise_.set_value(std::move(op).status());
-    }
-    if (op) {
-      std::unique_lock<std::mutex> lk(mu_);
-      op_name_ = std::move(*op->mutable_name());
+      // Consider trying to reset the backoff wait period.
+      DoCancel();
```
coryan, line 143: There is no guarantee that calling `Cancel()` will actually cancel anything, right?  The `Cancel()` could fail, could be ignored, could take several more minutes to resolve?  This means the operation could be running for much longer than the polling policy prescribes (which sounds bad), including "forever" (which sounds really bad). And there are still cases where you get back an error and don't know what the state of the long running operation is.

coryan (commented):
Maybe I am too paranoid about the reliability of canceling long running operations.

google/cloud/internal/async_polling_loop.cc
```diff
@@ -133,23 +133,14 @@ class AsyncPollingLoopImpl
...
-      if (op) {
-        // We should not be fabricating a `Status` value here. Rather, we
-        // should cancel the operation and wait for the next poll to return
-        // an accurate status to the user, otherwise they will have no idea
-        // how to react.
-        return promise_.set_value(Status(
-            StatusCode::kDeadlineExceeded,
-            location_ + "() - polling loop terminated by polling policy"));
+      if (!op) {
+        return promise_.set_value(std::move(op).status());
```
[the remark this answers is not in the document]
devbww: > After a cancel, the polling operation could fail with a transient and this would be returning with that error, and not the actual status of the long running operation?

Correct, but this is no worse than before; we always returned any transient error after the policy became exhausted.  I added a comment to the effect.

[We could consider also sending the cancel request in the exhausted/poll-failed case.]

google/cloud/internal/async_polling_loop.cc
devbww: > There is no guarantee that calling `Cancel()` will actually cancel anything, right? The `Cancel()` could fail, could be ignored, could take several more minutes to resolve? 

Right, although it is "best effort".

> This means the operation could be running for much longer than the polling policy prescribes (which sounds bad), including "forever" (which sounds really bad).

Correct, but this is no worse than before, where we always abandoned the operation, leaving it to run as long as it wanted.  Now, while the poll continues to succeed, we will continue to request cancellation, which is the anti-abandon behavior that I'll claim is a feature.

If a user really wants to abandon an operation, they probably want to use lower-level primitives ala #7658.

> And there are still cases where you get back an error and don't know what the state of the long running operation is.

Yes, the previously mentioned transient-at-exhaustion (which we could think about addressing).

coryan: > Correct, but this is no worse than before, where we always abandoned the operation, leaving it to run as long as it wanted.

Yes.  I was more worried about the application blocking on `.get()` or never getting a calling in a `.then()`.

> Now, while the poll continues to succeed, we will continue to request cancellation, which is the anti-abandon behavior that I'll claim is a feature.

I counter-claim: returning from `.get()` once the polling policy has expired is a feature too.

So, I like the fact that we don't abandon things, but I don't like the fact that now we are (maybe surprisingly) working -- and maybe blocking -- beyond the constraints in the polling policy. 

Would it be possible to make the new behavior optional?  If not, I think we should leave this alone and give folks this choice (to cancel on a polling expiration) once we implement #7658.

> If a user really wants to abandon an operation, they probably want to use lower-level primitives ala #7658.

I will make some notes there.

Issue #7990, opened 2022-01-17
Dialogflow API: `dialogflow.googleapis.com`


coryan:

coryan:
Note that this API already has proto libraries in `external/googleapis`, it may need some refactoring.  In addition, it has a bidirectional streaming RPCs:

* `google.cloud.dialogflow.cx.v3.Sessions.StreamingDetectIntent`
* `google.cloud.dialogflow.v2.Sessions.StreamingDetectIntent`


```
cat  ~/googleapis/api-index-v1.json | jq '.apis | map(select(.services != null)) | map(.services | map(select(.methods != null)) | map(.methods | map(select(.mode == "BIDIRECTIONAL_STREAMING") | [.mode, .fullName]  ))) | flatten ' | grep -v BIDIRECTIONAL_STREAMING  | grep -v beta
[
  "google.cloud.apigeeconnect.v1.Tether.Egress",
  "google.cloud.bigquery.storage.v1.BigQueryWrite.AppendRows",
  "google.cloud.dialogflow.cx.v3.Sessions.StreamingDetectIntent",
  "google.cloud.dialogflow.v2.Sessions.StreamingDetectIntent",
  "google.cloud.mediatranslation.v1alpha1.SpeechTranslationService.StreamingTranslateSpeech",
  "google.cloud.pubsublite.v1.CursorService.StreamingCommitCursor",
  "google.cloud.pubsublite.v1.PartitionAssignmentService.AssignPartitions",
  "google.cloud.pubsublite.v1.PublisherService.Publish",
  "google.cloud.pubsublite.v1.SubscriberService.Subscribe",
  "google.cloud.speech.v1.Speech.StreamingRecognize",
  "google.devtools.build.v1.PublishBuildEvent.PublishBuildToolEventStream",
  "google.firestore.v1.Firestore.Listen",
  "google.firestore.v1.Firestore.Write",
  "google.logging.v2.LoggingServiceV2.TailLogEntries",
  "google.pubsub.v1.Subscriber.StreamingPull"
]

coryan:
I think this is blocked by #8459

PR #8700, opened 2022-04-07
feat(generator): workaround dialogflow + doxygen problems

coryan:
Motivated by #7990.  I tried all kinds of other workarounds before settling for this monstrosity.

I cannot figure out how to pass arbitrary strings, with spaces and newlines, as parameters to the generator, if somebody tells me how to do that, we can move these out of the code and into `generator_config.textproto`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8700)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -450,9 +450,6 @@ std::string FormatApiMethodSignatureParameters(
         input_type->FindFieldByName(parameter);
     google::protobuf::SourceLocation loc;
     parameter_descriptor->GetSourceLocation(&loc);
-    //    if (loc.leading_comments.find("Context ID>`") != std::string::npos) {
-    //      std::cerr << "DEBUG DEBUG\n" << loc.leading_comments << std::endl;
-    //    }
     auto comment = absl::StrReplaceAll(
         loc.leading_comments,
         {{kDialogflowCXSesionIdProto, kDialogflowCXSesionIdCpp},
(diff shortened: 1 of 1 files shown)


generator/internal/descriptor_utils.cc
```diff
@@ -339,6 +339,104 @@ std::string FormatClassCommentsFromServiceComments(
   return absl::StrReplaceAll(doxygen_formatted_comments, {{"///  ", "/// "}});
 }
 
+auto constexpr kDialogflowCXSesionIdProto = R"""(
```
dbolduc, line 342: s/Sesion/Session/

generator/internal/descriptor_utils.cc
```diff
@@ -339,6 +339,104 @@ std::string FormatClassCommentsFromServiceComments(
...
+ Format: `projects/<Project ID>/locations/<Location ID>/agents/<Agent
+ ID>/sessions/<Session ID>` or `projects/<Project ID>/locations/<Location
+ ID>/agents/<Agent ID>/environments/<Environment ID>/sessions/<Session ID>`.
+)""";
+
+auto constexpr kDialogflowCXSesionIdCpp = R"""(
```
dbolduc, line 348: s/Sesion/Session/

dbolduc (commented):
Yeah, this is pretty ugly. Does @scotthart have a suggestion for how to make this nicer?

dbolduc (approved):
I don't have a better suggestion so..... approved

coryan:
The arguments to the generator are all munched into a single string.  I think this is the way to go for now.

PR #8707, opened 2022-04-08
feat(pubsublite): publisher connection impl

18suresha:
also templatizes `MockPartitionPublisher` -> `MockPublisher<T>` so that we can mock various types of publishers with the same code

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)


dpcollins-google (approved):
> Reviewed 1 of 5 files at r1, 2 of 6 files at r17, 1 of 1 files at r18, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)* status: 4 of 13 files reviewed, 6 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/options.h, line 39 at r21](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-PybyT1hvMc-uQJTpz:-N-PybyT1hvMc-uQJTq-:boais46) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7152db0f21f0fea94485a85e8c8bff6182d13a3d/google/cloud/pubsublite/options.h#L39)):*
> ```objc
> 
> /**
>  * Handler for if the publisher connection shuts down with a non-`kOk` status.
> ```

reword to `Handler to be called if a connection fails with a permanent error.`



_Code quote:_
```objc
if the publisher connection
```
___
*[google/cloud/pubsublite/internal/publisher_connection_impl_test.cc, line 49 at r21](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-PzYv32BlpDariDmEx:-N-PzYv32BlpDariDmEy:b-a73yud) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7152db0f21f0fea94485a85e8c8bff6182d13a3d/google/cloud/pubsublite/internal/publisher_connection_impl_test.cc#L49)):*
> ```cpp
>  protected:
>   PublisherConnectionImplTest()
>       : publisher_{absl::make_unique<
> ```

please restructure to remove the publisher_ member variable.
```
publisher_ref_{*new StrictMock<MockPublisher<MessageMetadata>>()} {
  EXPECT_CALL(*publisher_, Start).WillOnce(Return(ByMove(status_promise_.get_future())));
  conn_ = absl::make_unique<PublisherConnectionImpl>(absl::WrapUnique(&publisher_ref_), ...);
}

...

std::unique_ptr<google::cloud::pubsub::PublisherConnection> conn_;
```



_Code quote:_
```cpp
      : publisher_{absl::make_unique<
            StrictMock<MockPublisher<MessageMetadata>>>()},
        publisher_ref_{[this]() -> StrictMock<MockPublisher<MessageMetadata>>& {
          EXPECT_CALL(*publisher_, Start)
              .WillOnce(Return(ByMove(p_.get_future())));
          return *publisher_;
        }()},
```
___
*[google/cloud/pubsublite/internal/publisher_connection_impl_test.cc, line 65 at r21](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-Pzokc-QHiarZl1UUB:-N-Pzokc-QHiarZl1UUC:b-hlqbge) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7152db0f21f0fea94485a85e8c8bff6182d13a3d/google/cloud/pubsublite/internal/publisher_connection_impl_test.cc#L65)):*
> ```cpp
> 
>   std::unique_ptr<StrictMock<MockPublisher<MessageMetadata>>> publisher_;
>   promise<Status> p_;
> ```

Give a real name like `status_promise_`



_Code quote:_
```cpp
p_
```
___
*[google/cloud/pubsublite/internal/publisher_connection_impl.cc, line 39 at r21](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-Pz1tN3V96BwvSClE-:-N-Pz1tN3V96BwvSClE0:b-pl7f3s) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7152db0f21f0fea94485a85e8c8bff6182d13a3d/google/cloud/pubsublite/internal/publisher_connection_impl.cc#L39)):*
> ```cpp
>     if (status.ok()) return;
>     auto code = static_cast<absl::StatusCode>(status.code());
>     failure_handler(absl::Status{code, std::move(status).message()});
> ```

you have a status. no need to reconstruct it. Just `failure_handler(std::move(status));`



_Code quote:_
```cpp
    auto code = static_cast<absl::StatusCode>(status.code());
    failure_handler(absl::Status{code, std::move(status).message()});
```

coryan (changes requested):
> Reviewed 2 of 8 files at r8, 2 of 8 files at r14, 1 of 6 files at r17, 4 of 4 files at r22, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)* status: all files reviewed, 9 unresolved discussions (waiting on @18suresha)

___
*[google/cloud/pubsublite/options.h, line 33 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YrPzY6DPBht6YiNYi:-N-YrPzY6DPBht6YiNYj:b-ifkjhd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L33)):*
> ```objc
>  * protobuf `PubSubMessage`.
>  */
> using PublishMessageTransformer =
> ```

Is this something that application developers are required to use or reference?  Why?  Seems like it should be an implementation detail and thus live in `pubsublite_internal`?
___
*[google/cloud/pubsublite/options.h, line 40 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YraZx4O5_DhU887Nv:-N-YraZx4O5_DhU887Nw:b-n4w04m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L40)):*
> ```objc
>  * Handler to be called if a connection fails with a permanent error.
>  */
> using FailureHandler = std::function<void(Status)>;
> ```

Same question as above.
___
*[google/cloud/pubsublite/internal/multipartition_publisher_test.cc, line 171 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-Ys35GCl4uuzArhl9x:-N-Ys35H6G7pryxTU6hR:bvb37u2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/internal/multipartition_publisher_test.cc#L171)):*
> ```cpp
>   MultipartitionPublisherTest()
>       : partition_publisher_0_{std::make_shared<
>             StrictMock<MockPublisher<Cursor>>>()},
> ```

FYI, these changes seemed unrelated to me.  I think you are refactoring/generalizing the `MockPublisher` at the same time as you add the new class... I encourage you to break down such changes into separate PRs, but this is not such a large refactor as to mandate a split.
___
*[google/cloud/pubsublite/internal/publisher_connection_impl.h, line 33 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YsCWODMa50VEHsCmC:-N-YsCWODMa50VEHsCmD:b79wuhq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/internal/publisher_connection_impl.h#L33)):*
> ```objc
>  */
> class PublisherConnectionImpl
>     : public ::google::cloud::pubsub::PublisherConnection {
> ```

Normally we do not use an initial `::` for names like this, why here?

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)* status: 12 of 13 files reviewed, 9 unresolved discussions (waiting on @18suresha and @coryan)

___
*[google/cloud/pubsublite/options.h, line 33 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YrPzY6DPBht6YiNYi:-N-Yv0xtAUkXG9wiPicN:bfjuiis) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L33)):*

Is this something that application developers are required to use or reference?  Why?  Seems like it should be an implementation detail and thus live in `pubsublite_internal`?
</blockquote></details>

It facilitates the Pub/Sub -> Pub/Sub Lite use case for customers, and is intended to match what's in other client libraries. Customers can customize how they transform these messages if they choose to do so.
___
*[google/cloud/pubsublite/options.h, line 40 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YraZx4O5_DhU887Nv:-N-YvSX-AKIqApjbQk9J:bo4ujrc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L40)):*

Same question as above.
</blockquote></details>

This is something that application developers can customize so they have a hook into permanent failures. It's not required as there will be a minimal default implementation but it is something that customers should have access to so they can react to permanent failures with this layer as they see fit
___
*[google/cloud/pubsublite/internal/publisher_connection_impl.h, line 33 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YsCWODMa50VEHsCmC:-N-YzLTj4lq_JMyHnSrt:b-zb8auc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/internal/publisher_connection_impl.h#L33)):*

Normally we do not use an initial `::` for names like this, why here?
</blockquote></details>

oversight, fixed

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)* status: 12 of 13 files reviewed, 9 unresolved discussions (waiting on @18suresha and @coryan)

___
*[google/cloud/pubsublite/options.h, line 40 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YraZx4O5_DhU887Nv:-N-ZZ2SI-e_YG2Sdd-pf:b2l5qrh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L40)):*

This is something that application developers can customize so they have a hook into permanent failures. It's not required as there will be a minimal default implementation but it is something that customers should have access to so they can react to permanent failures with this layer as they see fit
</blockquote></details>

As in it's not required for them to customize it

coryan (changes requested):
> Reviewed 1 of 1 files at r23, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)* status: all files reviewed, 8 unresolved discussions (waiting on @18suresha)

___
*[google/cloud/pubsublite/options.h, line 33 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YrPzY6DPBht6YiNYi:-N-Z_5jECUThmGh1tNnq:bia0gt6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L33)):*

It facilitates the Pub/Sub -> Pub/Sub Lite use case for customers, and is intended to match what's in other client libraries. Customers can customize how they transform these messages if they choose to do so.
</blockquote></details>

I think this is the second time I ask: it would really help if we had a design document explaining the public API we want to expose.  It would save you and me a lot of time discussing what needs and not needs to be in the public APIs.

With that said: is there going to be a default version of this?  I ask because if it is an option we would probably define:

```cc
struct PublishMessageTransformerOption {
  using Type = std::function< ... as in your PR ...>;
};
```

if there is no default, we would probably make it a required argument and the code you have is fine.
___
*[google/cloud/pubsublite/options.h, line 40 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YraZx4O5_DhU887Nv:-N-Za1iU2-ObB4NUGzFW:bqjal71) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L40)):*

As in it's not required for them to customize it
</blockquote></details>

Not following.  They get the permanent failures in the `future<StatusOr<std::string>>` result.  Do you want them to be able to ignore the permanent failure?  They can do that using a custom retry policy.  Why do we need an additional mechanism for this?

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)* status: all files reviewed, 8 unresolved discussions (waiting on @18suresha and @coryan)

___
*[google/cloud/pubsublite/options.h, line 33 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YrPzY6DPBht6YiNYi:-N-_qwRY2dRyqXSkT6k4:bhomivs) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L33)):*

I think this is the second time I ask: it would really help if we had a design document explaining the public API we want to expose.  It would save you and me a lot of time discussing what needs and not needs to be in the public APIs.

With that said: is there going to be a default version of this?  I ask because if it is an option we would probably define:

```cc
struct PublishMessageTransformerOption {
  using Type = std::function< ... as in your PR ...>;
};
```

if there is no default, we would probably make it a required argument and the code you have is fine.
</blockquote></details>

copying what I had written in the doc:
There should be a user-facing `PublishMessageTransformerOption` of the type `std::function<StatusOr<google::cloud::pubsublite::v1::PubSubMessage>(google::cloud::pubsub::Message)>`. With this, end users can specify how they would like these messages to be transformed for their specific use case. If the user does not specify this, then there will be a default implementation that just maps corresponding entries in each protobuf (https://github.com/googleapis/google-cloud-cpp/blob/fdb3cfd47b7b4f4ac22c3a8f916caff168fab355/google/cloud/pubsublite/internal/default_publish_message_transformer.cc). More specifically, it provides a workaround for users to publish Pub/Sub Lite messages which are inexpressible as Pub/Sub messages with the standard transforms, ex. those with binary attributes.
___
*[google/cloud/pubsublite/options.h, line 40 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YraZx4O5_DhU887Nv:-N-_r2fsBrpsAw-crIlI:b27k7wc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L40)):*

Not following.  They get the permanent failures in the `future<StatusOr<std::string>>` result.  Do you want them to be able to ignore the permanent failure?  They can do that using a custom retry policy.  Why do we need an additional mechanism for this?
</blockquote></details>

`FailureHandler` is meant for when the permanent failure propagates to user-facing objects. `RetryPolicy` is used at the lowest layer and is not meant to be a user-specifiable option.
copying what I had written in the doc:
There should be a user-facing `FailureHandlerOption`. When looking at the `google::cloud::pubsub::PublisherConnection API`, there is no certain way to know if the underlying connection encountered a permanent failure. `StatusOr<>`s are returned through the `Publish` API, but this is semantically meant for failures of individual `Publish` calls which may or may not be due to a permanent failure on the connection. End users require a hook into the immediate permanent failure of a connection, thus we intend to expose the `FailureHandlerOption`. The type will be `std::function<void(Status)>`. Example use cases for end users include, attempting to instantiate another connection, updating their own error logs, etc.

coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)* status: all files reviewed, 8 unresolved discussions (waiting on @18suresha)

___
*[google/cloud/pubsublite/options.h, line 33 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YrPzY6DPBht6YiNYi:-N-bydvl4NnZwoWTMYgK:b-tkrq7w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L33)):*

copying what I had written in the doc:
There should be a user-facing `PublishMessageTransformerOption` of the type `std::function<StatusOr<google::cloud::pubsublite::v1::PubSubMessage>(google::cloud::pubsub::Message)>`. With this, end users can specify how they would like these messages to be transformed for their specific use case. If the user does not specify this, then there will be a default implementation that just maps corresponding entries in each protobuf (https://github.com/googleapis/google-cloud-cpp/blob/fdb3cfd47b7b4f4ac22c3a8f916caff168fab355/google/cloud/pubsublite/internal/default_publish_message_transformer.cc). More specifically, it provides a workaround for users to publish Pub/Sub Lite messages which are inexpressible as Pub/Sub messages with the standard transforms, ex. those with binary attributes.
</blockquote></details>

Okay, looks like this should be wrapped by a `*Option`, but also looks like this will happen in a future PR, that SGTM.
___
*[google/cloud/pubsublite/options.h, line 40 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YraZx4O5_DhU887Nv:-N-bym_k1Ajti72Npgpa:b63jwg5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L40)):*
> FailureHandler is meant for when the permanent failure propagates to user-facing objects.

I think the `Publish()` function will return a `future<Status<std::string>> f`, applications can use:

```cc
auto publisher.Publish(...).then([handler](auto g) {
  auto id = g.get(); 
  if (!g) return handler(g.status());
  return id;
});
```

if they want to trap the error with a handler.

> RetryPolicy is used at the lowest layer and is not meant to be a user-specifiable option.

In all the other C++ client libraries they are very much a user-specifiable option. 

> copying what I had written in the doc:

I think the applications already have a way to call a function for errors, they don't need a second one.

> There should be a user-facing FailureHandlerOption. 

That is assuming the hypothesis.  I am asking why we should have a FailureHandlerOption and your answer is because the document says we should... You know better than to attempt circular arguments.

>When looking at the google::cloud::pubsub::PublisherConnection API, there is no certain way to know if the underlying connection encountered a permanent failure. StatusOr<>s are returned through the Publish API, but this is semantically meant for failures of individual Publish calls which may or may not be due to a permanent failure on the connection. End users require a hook into the immediate permanent failure of a connection, thus we intend to expose the FailureHandlerOption. The type will be std::function<void(Status)>. Example use cases for end users include, attempting to instantiate another connection, updating their own error logs, etc.

If users really need that (I doubt it) they can use the RetryPolicy.  That is why we have it, and that is why it is customizable.

Please remove this additional handler.

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)* status: all files reviewed, 8 unresolved discussions (waiting on @18suresha and @coryan)

___
*[google/cloud/pubsublite/options.h, line 40 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YraZx4O5_DhU887Nv:-N-cIvlI6sxYRcdZnWbC:bu9krgz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L40)):*
> auto publisher.Publish(...).then([handler](auto g) {
>   auto id = g.get(); 
>   if (!g) return handler(g.status());
>   return id;
> });
Here the error from `Publish` might not be due to a permanent connection, it could be that they sent a malformed message. They can check for the type of error in their handler, but I don't think it should be the user's responsibility to identify what is and what is not identified as a permanent failure on our end, rather they should just provide what they want to be done in such a scenario. Additionally, the permanent failure might happen before the user calls `Publish` (during stream initialization for example), so the user shouldn't have to rely on calling `Publish` to identify if an error is present and if that error is due to a permanent failure on the connection.

> your answer is because the document says we should
my bad about the wording and copying and pasting from the doc
'end users require a hook' should read end users should have a hook because of [previous sentence in doc]'

The main point is that users should be able to react to a permanent failure instead of attempting to identify one first through an API whose purpose is not intended for that as `PublisherConnection` doesn't provide an intended detection mechanism for it.

coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707)* status: all files reviewed, 7 unresolved discussions (waiting on @18suresha)

___
*[google/cloud/pubsublite/options.h, line 40 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8707#-N-YraZx4O5_DhU887Nv:-N-cP4Zd59kMAJk7miYn:b-wypz6t) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/67389fab1bf0c2a457579db6859bb43913c9677c/google/cloud/pubsublite/options.h#L40)):*
> The main point is that users should be able to react to a permanent failure instead of attempting to identify one first through an API whose purpose is not intended for that as PublisherConnection doesn't provide an intended detection mechanism for it.

That is what the retry policies are for.  We don't need a second mechanism.

dpcollins-google:
We are not planning on giving users control over the retry policies for any of the streams. They are in no position to decide which codes are retryable for our RPCs, and the layer that a retry policy gives them access to (individual streaming RPCs) is not a layer of abstraction they should need to care about.

This (not letting users have control over retry policies) actually gets even worse on the subscribe side, since certain ErrorInfo information indicates a different retry behavior is needed for the subscribe stream when an admin seek is performed.

In general, we need to expose a hook for users to know when their connection, not an individual RPC stream which they shouldn't care about, has failed. RetryPolicy does not fit this use case.

dpcollins-google:
@coryan I've written up a document providing the rationale behind this approach, please take a look. go/psl-cpp-permanent-error-handler

google/cloud/pubsublite/internal/publisher_connection_impl.cc
```diff
@@ -0,0 +1,80 @@
...
+        google::cloud::pubsublite::MessageMetadata>>
+        publisher,
+    PublishMessageTransformer transformer)
+    : publisher_{std::move(publisher)},
+      message_transformer_{std::move(transformer)},
+      service_composite_{publisher_.get()} {
+  // memory safe since this will at latest occur in the downcall of
+  // `ServiceComposite::Shutdown`
+  service_composite_.Start().then([this](future<Status>) {
+    // shutdown in case of failure to finish all outstanding `Publish` futures
```
dpcollins-google, line 37: nit. Can we add:

Status status = fut.get().status();
if (!status.ok()) {
  GCP_LOG(WARNING) << "Publisher failed permanently: " << status;
}

google/cloud/pubsublite/internal/publisher_connection_impl.h
```diff
@@ -0,0 +1,66 @@
...
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * A connection implementation for publishing messages to a single `Topic`.
+ */
+class PublisherConnectionImpl
+    : public google::cloud::pubsub::PublisherConnection {
+ public:
+  PublisherConnectionImpl(
+      std::unique_ptr<google::cloud::pubsublite_internal::Publisher<
```
dpcollins-google, line 36: nit. google::cloud::pubsublite_internal:: prefix shouldn't be needed.

google/cloud/pubsublite/options.h
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Converts a Google Cloud Pub/Sub `Message` into a Google Cloud Pub/Sub Lite
+ * protobuf `PubSubMessage`.
+ */
+using PublishMessageTransformer =
+    std::function<StatusOr<google::cloud::pubsublite::v1::PubSubMessage>(
+        google::cloud::pubsub::Message)>;
+
+/**
```
dpcollins-google, line 37: delete

dpcollins-google (approved):

google/cloud/pubsublite/BUILD.bazel
```diff
@@ -26,6 +26,7 @@ cc_library(
     deps = [
         "//:common",
         "//:grpc_utils",
+        "//:pubsub",
```
devbww, line 29: It seems strange for "pubsub light" to depend on "pubsub heavy".  Perhaps there is some refactoring that should happen here.

coryan: #8859

PR #8736, opened 2022-04-11
feat(pubsublite): publisher connection factory

18suresha:
depends on https://github.com/googleapis/google-cloud-cpp/pull/8707 and https://github.com/googleapis/google-cloud-cpp/pull/8818

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736)


dpcollins-google (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736)* status: 0 of 14 files reviewed, 17 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[google/cloud/pubsublite/options.h, line 39 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-U11MxEhptwSrhnlgH:-N-U11MxEhptwSrhnlgI:bd5e6wo) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f983e49bc836546f446a024f6f70f3813190b06/google/cloud/pubsublite/options.h#L39)):*
> ```objc
>         google::cloud::pubsub::Message)>;
> 
> struct PublishMessageTransformerOption {
> ```

I think we need some documentation on all of these. See the golang versions for good examples. https://pkg.go.dev/cloud.google.com/go/pubsublite/pscompat#PublishSettings

Note that we don't need to support timeout



_Code quote:_
```objc
struct PublishMessageTransformerOption {
  using Type = PublishMessageTransformer;
};

/**
 * Handler to be called if a connection fails with a permanent error.
 */
using FailureHandler = std::function<void(Status)>;

struct FailureHandlerOption {
  using Type = FailureHandler;
};

struct MaximumBatchMessageCountOption {
  using Type = std::int64_t;
};

struct MaximumBatchBytesOption {
  using Type = std::int64_t;
};

struct PublishFlushAlarmPeriodOption {
};
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 77 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-U1YQc9kg226aLeoC1:-N-U1YQc9kg226aLeoC2:b-dihy0v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f983e49bc836546f446a024f6f70f3813190b06/google/cloud/pubsublite/publisher_connection.cc#L77)):*
> ```cpp
> using google::protobuf::Value;
> 
> const PublishMessageTransformer kDefaultMessageTransformer =
> ```

make this a free function, and pass &FunctionName for the functor. Also, put it in a pubsublite/internal header and test it



_Code quote:_
```cpp
const PublishMessageTransformer kDefaultMessageTransformer =
    [](Message const& message) {
      PubSubMessage m;
      m.set_key(message.ordering_key());
      *m.mutable_data() = message.data();
      for (auto kv : message.attributes()) {
        AttributeValues av;
        av.add_values(std::move(kv.second));
        (*m.mutable_attributes())[kv.first] = std::move(av);
      }
      return m;
    };
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 90 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YFxYc84r7TLtyL2RP:-N-YFxYc84r7TLtyL2RQ:b-nv3tda) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L90)):*
> ```cpp
>     };
> 
> class PubSubLiteRetryPolicy : public RetryPolicy {
> ```

put in an internal header and test it



_Code quote:_
```cpp
class PubSubLiteRetryPolicy : public RetryPolicy {
 public:
  bool OnFailure(Status const& s) override {
    return retryable_codes_.find(static_cast<std::uint8_t>(s.code())) !=
           retryable_codes_.end();
  }

  bool IsExhausted() const override { return false; }

  bool IsPermanentFailure(Status const& s) const override {
    return retryable_codes_.find(static_cast<std::uint8_t>(s.code())) ==
           retryable_codes_.end();
  }

 private:
  std::unordered_set<std::uint8_t> retryable_codes_{
      static_cast<std::uint8_t>(StatusCode::kDeadlineExceeded),
      static_cast<std::uint8_t>(StatusCode::kAborted),
      static_cast<std::uint8_t>(StatusCode::kInternal),
      static_cast<std::uint8_t>(StatusCode::kUnavailable),
      static_cast<std::uint8_t>(StatusCode::kUnknown),
      static_cast<std::uint8_t>(StatusCode::kResourceExhausted)};
};
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 90 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YGNDoD3euakY7fYKI:-N-YGNDoD3euakY7fYKJ:b-8fafoy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L90)):*
> ```cpp
>     };
> 
> class PubSubLiteRetryPolicy : public RetryPolicy {
> ```

StreamRetryPolicy, no need to prefix with PubSubLite
___
*[google/cloud/pubsublite/publisher_connection.cc, line 105 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YGGDN30rAri8UqU2C:-N-YGGDN30rAri8UqU2D:b96coy5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L105)):*
> ```cpp
> 
>  private:
>   std::unordered_set<std::uint8_t> retryable_codes_{
> ```

just use StatusCode as the value type



_Code quote:_
```cpp
std::uint8_t
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 138 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YSRWw93AuOF7Hom0D:-N-YSRWw93AuOF7Hom0E:b-qht3ox) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L138)):*
> ```cpp
> }
> 
> std::string GetSerializedContext() {
> ```

make this function take `const std::string& framework`, and set it to the "framework" key

You should pass "CLOUD_PUBSUB_SHIM" as the value for this field.



_Code quote:_
```cpp
std::string GetSerializedContext() 
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 142 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YRpVq712iJtNqh3SY:-N-YRpVq712iJtNqh3SZ:b-62gmwr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L142)):*
> ```cpp
>   auto& metadata_map = *context.mutable_fields();
>   Value lang;
>   *lang.mutable_string_value() = "cpp";
> ```

all caps



_Code quote:_
```cpp
version_minor
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 157 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YSrTU7KcTko9iPiN2:-N-YSrTU7KcTko9iPiN3:b12jegj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L157)):*
> ```cpp
> std::unique_ptr<PublisherConnection> MakePublisherConnection(Topic topic,
>                                                              Options opts) {
>   if (!opts.has<GrpcNumChannelsOption>()) {
> ```

put a comment here explaining why we do this.

// Default to 20 channels per client to avoid limitations on streams and requests per-channel.



_Code quote:_
```cpp
  if (!opts.has<GrpcNumChannelsOption>()) {
    opts.set<GrpcNumChannelsOption>(20);
  }
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 160 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YTS2iAcCDEVMI38x4:-N-YTS2iAcCDEVMI38x5:bfelxdt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L160)):*
> ```cpp
>     opts.set<GrpcNumChannelsOption>(20);
>   }
>   if (!opts.has<PublishMessageTransformerOption>()) {
> ```

Change the structure, here and below for FailureHandler, to this:

```
PublishMessageTransformer transformer = pubsublite_internal::DefaultPublishMessageTransformer;
if (opts.has<PublishMessageTransformerOption>()) {
  transformer = opts.get<PublishMessageTransformerOption>();
}
```



_Code quote:_
```cpp
  if (!opts.has<PublishMessageTransformerOption>()) {
    opts.set<PublishMessageTransformerOption>(kDefaultMessageTransformer);
  }
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 164 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YU4LdBD87nyFX8qtF:-N-YU4LdBD87nyFX8qtG:bga33uo) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L164)):*
> ```cpp
>   }
>   if (!opts.has<FailureHandlerOption>()) {
>     opts.set<FailureHandlerOption>([](Status const&) {});
> ```

LOG(WARNING) the status with "Pub/Sub Lite connection failed: "



_Code quote:_
```cpp
[](Status const&) {}
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 168 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YUSSu22b0RXA4dQUp:-N-YUSSu22b0RXA4dQUq:b-wihdnj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L168)):*
> ```cpp
>   if (!opts.has<EndpointOption>()) {
>     auto endpoint = GetEndpoint(topic.location());
>     if (!endpoint) return nullptr;
> ```

you should never return nullptr. Since this factory can fail, we should instead return StatusOr<std::unique_ptr<PublisherConnection>>
___
*[google/cloud/pubsublite/publisher_connection.cc, line 171 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YUmiM5I2DYCdtdjE-:-N-YUmiM5I2DYCdtdjE0:bexyy36) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L171)):*
> ```cpp
>     opts.set<EndpointOption>(*std::move(endpoint));
>   }
>   CompletionQueue cq = MakeBackgroundThreadsFactory(opts)()->cq();
> ```

you need to save it somewhere to keep it alive. I think you need to move ContainingPublisherConnection from google/cloud/pubsub/publisher_connection.cc to pubsub_internal and use that



_Code quote:_
```cpp
MakeBackgroundThreadsFactory(opts)
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 187 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YXYVy4taz0lazCsd3:-N-YXYVy4taz0lazCsd4:bkqc2l6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L187)):*
> ```cpp
>   std::function<std::shared_ptr<
>       Publisher<google::cloud::pubsublite::v1::Cursor>>(std::uint32_t)>
>       partition_publisher_factory = [topic, backoff_policy, cq, opts,
> ```

Please do all parsing of the google::cloud::Options in this function if possible? I.e. don't bind it to any lambdas
___
*[google/cloud/pubsublite/publisher_connection.cc, line 198 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YVydq4sNzGYbg0gaR:-N-YVydq4sNzGYbg0gaS:b-fjou23) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L198)):*
> ```cpp
>                         initializer) {
>               ClientMetadata metadata;
>               metadata["x-goog-request-params"] = absl::StrCat(
> ```

Can you extract this to a helper? `MakeClientMetadata(const Topic& topic, std::uint32_t partition)`



_Code quote:_
```cpp
              metadata["x-goog-request-params"] = absl::StrCat(
                  "partition=", partition, "&", "topic=", topic.FullName());
              metadata["x-goog-pubsub-context"] = GetSerializedContext();
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 203 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YWDyT64Bmb_EWddBM:-N-YWDyT64Bmb_EWddBN:bl5tg5w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L203)):*
> ```cpp
>               return absl::make_unique<ResumableAsyncStreamingReadWriteRpcImpl<
>                   PublishRequest, PublishResponse>>(
>                   []() { return absl::make_unique<PubSubLiteRetryPolicy>(); },
> ```

tiny nit. you don't need () if you take no arguments and dont specify the return type



_Code quote:_
```cpp
()
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 205 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YWUr999NHvZ8ZrB7f:-N-YWUr999NHvZ8ZrB7g:bzefm7o) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L205)):*
> ```cpp
>                   []() { return absl::make_unique<PubSubLiteRetryPolicy>(); },
>                   backoff_policy, sleeper,
>                   MakeStreamFactory(CreateDefaultPublisherServiceStub(cq, opts),
> ```

You should call MakeStreamFactory once and bind to the lambda. I.e.

```
auto stream_factory= MakeStreamFactory(...);
auto partition_publisher_factory = [stream_factory, ...](std::uint32_t partition) {
```
___
*[google/cloud/pubsublite/publisher_connection.cc, line 209 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YXV2U7MV5UJr86aVr:-N-YXV2U7MV5UJr86aVs:b-qk9yqw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L209)):*
> ```cpp
>                   std::move(initializer));
>             },
>             CreateBatchingOptions(opts), std::move(ipr), alarm_registry);
> ```

Call this once and bind BatchingOptions instead



_Code quote:_
```cpp
CreateBatchingOptions(opts)
```

google/cloud/pubsublite/internal/containing_publisher_connection.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "google/cloud/pubsub/publisher_connection.h"
+#include "google/cloud/options.h"
+
+namespace google {
+namespace cloud {
+namespace pubsublite_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// ported from google::cloud::pubsub
```
dpcollins-google, line 26: please don't port, just move the pubsub version to a header in the pubsub internal folder, then import it

google/cloud/pubsublite/internal/default_publish_message_transformer_test.cc
```diff
@@ -0,0 +1,85 @@
...
+
+class DefaultPublishMessageTransformerTest : public ::testing::Test {
+ protected:
+  DefaultPublishMessageTransformerTest()
+      : pubsub_message_{MessageBuilder{}
+                            .SetData("dataaaa")
+                            .SetOrderingKey("keyyyy")
+                            .InsertAttribute("attr1_key", "attr1_key")
+                            .InsertAttribute("lang", "cpp")},
+        psl_message_{[] {
```
dpcollins-google, line 40: psl_message_ doesn't need to be a member- make a free function ExamplePslMessage() that returns it

google/cloud/pubsublite/internal/default_publish_message_transformer_test.cc
```diff
@@ -0,0 +1,85 @@
...
+
+using google::cloud::internal::Base64Encoder;
+using google::cloud::pubsub::MessageBuilder;
+using google::cloud::pubsublite::v1::PubSubMessage;
+using google::protobuf::Timestamp;
+
+class DefaultPublishMessageTransformerTest : public ::testing::Test {
+ protected:
+  DefaultPublishMessageTransformerTest()
+      : pubsub_message_{MessageBuilder{}
```
dpcollins-google, line 35: ditto for pubsub_message_. I don't think you need a harness here

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,215 @@
...
+            GCP_LOG(INFO) << "`MakeRelativeTimer` returned a non-ok status: "
+                          << status.status();
+          };
+        });
+  };
+
+  auto publisher_service_stub = CreateDefaultPublisherServiceStub(cq, opts);
+  auto batching_options = CreateBatchingOptions(opts);
+
+  std::function<std::shared_ptr<
```
dpcollins-google, line 163: auto

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,215 @@
...
+      Publisher<google::cloud::pubsublite::v1::Cursor>>(std::uint32_t)>
+      partition_publisher_factory = [topic, backoff_policy, batching_options,
+                                     cq, publisher_service_stub,
+                                     sleeper](std::uint32_t partition) {
+        InitialPublishRequest ipr;
+        *ipr.mutable_topic() = topic.FullName();
+        ipr.set_partition(partition);
+        AlarmRegistryImpl alarm_registry{cq};
+        auto stream_factory = MakeStreamFactory(
```
dpcollins-google, line 172: put this on line 162, outside the lambda

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,215 @@
...
+            publisher_service_stub, cq, MakeClientMetadata(topic, partition));
+        return std::make_shared<PartitionPublisher>(
+            [backoff_policy, publisher_service_stub, cq, sleeper,
```
dpcollins-google, line 175: [=] (Stream...

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,215 @@
...
+  };
+
+  auto publisher_service_stub = CreateDefaultPublisherServiceStub(cq, opts);
+  auto batching_options = CreateBatchingOptions(opts);
+
+  std::function<std::shared_ptr<
+      Publisher<google::cloud::pubsublite::v1::Cursor>>(std::uint32_t)>
+      partition_publisher_factory = [topic, backoff_policy, batching_options,
+                                     cq, publisher_service_stub,
+                                     sleeper](std::uint32_t partition) {
```
dpcollins-google, line 167: [=](...

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,215 @@
...
+        InitialPublishRequest ipr;
```
dpcollins-google, line 168: InitialPublishRequest request;

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,215 @@
...
+    }
+    opts.set<EndpointOption>(*std::move(endpoint));
+  }
+  std::unique_ptr<BackgroundThreads> background_threads =
+      MakeBackgroundThreadsFactory(opts)();
+  CompletionQueue cq = background_threads->cq();
+  std::shared_ptr<BackoffPolicy const> backoff_policy =
+      std::make_shared<ExponentialBackoffPolicy>(std::chrono::milliseconds{10},
+                                                 std::chrono::seconds{10}, 2.0);
+  AsyncSleeper sleeper = [cq](std::chrono::milliseconds dur) mutable {
```
dpcollins-google, line 149: `dur` -> sleep_duration

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,215 @@
...
+  if (opts.has<PublishMessageTransformerOption>()) {
+    transformer = opts.get<PublishMessageTransformerOption>();
+  }
+  FailureHandler failure_handler = [](Status const& s) {
+    GCP_LOG(WARNING) << "Pub/Sub Lite connection failed: " << s;
+  };
+  if (opts.has<FailureHandlerOption>()) {
+    failure_handler = opts.get<FailureHandlerOption>();
+  }
+  return static_cast<std::unique_ptr<PublisherConnection>>(
```
dpcollins-google, line 199: static_cast not needed, its an implicit conversion

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,215 @@
...
+}
+
+std::string GetSerializedContext(std::string const& framework) {
+  Struct context;
+  auto& metadata_map = *context.mutable_fields();
+  Value lang;
+  *lang.mutable_string_value() = "CPP";
+  metadata_map["language"] = std::move(lang);
+  Value framework_val;
+  *framework_val.mutable_string_value() = framework;
```
dpcollins-google, line 108: anywhere you have "mutable_xyz" where xyz is a string, do "set_xyz" instead.

dpcollins-google (approved):
Approved contingent on changes

google/cloud/pubsublite/publisher_connection.cc
18suresha: `auto stream_factory = MakeStreamFactory(publisher_service_stub, cq, MakeClientMetadata(topic, partition));` requires `partition` which is a parameter of the lambda

google/cloud/pubsublite/publisher_connection.cc
18suresha: I originally had that but it didn't work. I think it's because the return type of the function is `StatusOr<std::unique_ptr<PublisherConnection>>` instead of `std::unique_ptr<PublisherConnection>` so the compiler isn't double implicitly casting to `std::unique_ptr<>` and then to `StatusOr<std::unique_ptr<>>`

dpcollins-google: sgtm

google/cloud/pubsublite/publisher_connection.cc
dpcollins-google: you're right, sorry

google/cloud/pubsublite/internal/stream_retry_policy.h
```diff
@@ -0,0 +1,61 @@
...
+
+  bool IsPermanentFailure(Status const& s) const override {
+    return retryable_codes_.find(s.code()) == retryable_codes_.end();
+  }
+
+ private:
+  class StatusCodeHash {
+   public:
+    std::size_t operator()(StatusCode const& code) const {
+      return hash_(static_cast<std::uint8_t>(code));
```
dpcollins-google, line 44: has no state, just put inline: std::hash<std::uint8_t>{}(...)

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,208 @@
...
+
+std::string GetSerializedContext(std::string const& framework) {
+  Struct context;
+  auto& metadata_map = *context.mutable_fields();
+  Value lang;
+  lang.set_string_value("CPP");
+  metadata_map["language"] = std::move(lang);
+  Value framework_val;
+  framework_val.set_string_value(framework);
+  metadata_map["framework"] = framework_val;
```
dpcollins-google, line 109: anywhere you do a map accessor: it creates an implicit instance. just use for example:

metadata_map["framework"].set_string_value(framework);

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,208 @@
...
+  Value minor_version;
```
dpcollins-google, line 110: major_version and minor_version never put in the map

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -0,0 +1,208 @@
...
+  minor_version.set_number_value(version_minor());
+  Value major_version;
+  major_version.set_number_value(version_major());
+  std::string serialized_context;
+  context.SerializePartialToString(&serialized_context);
+  Base64Encoder encoder;
```
dpcollins-google, line 116: nit. can you please add a method to Base64Encoder:
```
void PushBack(absl::string_view string) {
   for (char const c : string) PushBack(c);
}
```

dpcollins-google (approved):

google/cloud/internal/base64_transforms.h
```diff
@@ -36,6 +37,9 @@ class Base64Encoder {
     buf_[len_++] = c;
     if (len_ == buf_.size()) Flush();
   }
+  void PushBack(absl::string_view string) {
+    for (char const c : string) PushBack(c);
+  }
```
devbww, line 42: Notice that you've enshrined a `signed char` (in all probabilities) to `unsigned char` conversion here.  Base64 is intended for "binary data", which means `absl::string_view` was probably the wrong representation in the first place, and why raw `char` was never mentioned here before.

coryan (changes requested):
I would prefer if we can split the refactor in `google/cloud/pubsub` to a separate PR.

> Reviewed 1 of 6 files at r9, 3 of 9 files at r10, 4 of 9 files at r14, 1 of 2 files at r15, 1 of 4 files at r16, 3 of 4 files at r18, 4 of 8 files at r21, 8 of 8 files at r22, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-:-N0gEdCY0c6HHEtf3t1J:bykjklg)* status: all files reviewed, 44 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[`google/cloud/pubsub/publisher_connection.cc` line 95 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gDYrvE0_245Rpgh81:-N0gDYrvE0_245Rpgh82:b360jah) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsub/publisher_connection.cc#L95)):*
> ```cpp
>   }
>   return std::make_shared<
>       google::cloud::pubsub_internal::ContainingPublisherConnection>(
> ```

nit: for consistency replace `google::cloud::pubsub_internal` with `pubsub_internal`.
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 168 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YUSSu22b0RXA4dQUp:-N0gM3IK0URjPLXnp4Og:btaq907) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L168)):*
<details><summary><i>Previously, dpcollins-google wrote…</i></summary><blockquote>

you should never return nullptr. Since this factory can fail, we should instead return StatusOr<std::unique_ptr<PublisherConnection>>
</blockquote></details>

Alternatively: is there an endpoint we could use when the location cannot be parsed?  Or is there a way we can validate the topic name and take a type that represents a validated name as a parameter?  Yet another alternative: don't validate the location ... just pass it to the server and let it validate whether it is valid or not.
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 149 at r13](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-Q-Ms96IPNnNkfsFiV-r13-149:-N0gNNlv5fg5nc3x7wZF:bazi9up) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e4e263ccd6d428ff43e89c8b3931e1541ec25cb1/google/cloud/pubsublite/publisher_connection.cc#L149)):*
<details><summary><i>Previously, dpcollins-google wrote…</i></summary><blockquote>

`dur` -> sleep_duration
</blockquote></details>

nit: it is not exactly a sleep time, it does not block the thread, maybe `backoff_time` is better?
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 199 at r13](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-Q-Ms96IPNnNkfsFiV-r13-199:-N0gZTB64XT_FKbDoU8E:b-mubzf8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e4e263ccd6d428ff43e89c8b3931e1541ec25cb1/google/cloud/pubsublite/publisher_connection.cc#L199)):*
<details><summary><i>Previously, dpcollins-google wrote…</i></summary><blockquote>

sgtm
</blockquote></details>

This is not quite a `static_cast<>`, I mean, obviously it is, but it is rare to use `static_cast` for this type of conversion. Just say:

```cc
return std::unique_ptr<PublisherConnection>(...
```
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 43 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gK2Nh3ud6puTmtjKK:-N0gK2Nh3ud6puTmtjKL:bjlfnog) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/publisher_connection.cc#L43)):*
> ```cpp
> using google::cloud::internal::Base64Encoder;
> using google::cloud::internal::MakeBackgroundThreadsFactory;
> ```

nit: We do not leave blank lines between groups like these.
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 71 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gKl3U3dG3X4jxAj4w:-N0gKl3U3dG3X4jxAj4x:b-2a2ysz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/publisher_connection.cc#L71)):*
> ```cpp
> using google::protobuf::Struct;
> 
> BatchingOptions CreateBatchingOptions(Options const& opts) {
> ```

nit: the naming convention in C++ is `MakeBatchingOptions()`, e.g. `std::make_shared<>` or `std::make_pair` or `std::make_unique`.
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 117 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gLFbR6kZa4lrR5XN7:-N0gLFbR6kZa4lrR5XN8:bp6f3fm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/publisher_connection.cc#L117)):*
> ```cpp
>     // Default to 20 channels per client to avoid limitations on streams and
>     // requests per-channel.
>     opts.set<GrpcNumChannelsOption>(20);
> ```

Why 20?  Why not 19, or 21 or 42 or 1000?  There is probably a rationale for this value, it would be good to have it documented.  If it is just "experience shows" then that is fine.
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 129 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gMdDl3d7tbRTWamyx:-N0gMdDl3d7tbRTWamyy:b1ngdam) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/publisher_connection.cc#L129)):*
> ```cpp
>   }
>   opts = google::cloud::internal::PopulateCommonOptions(
>       std::move(opts), "GOOGLE_CLOUD_PUBSUBLITE_IGNORED_OVERRIDE", "",
> ```

This is introducing `GOOGLE_CLOUD_PUBSUBLITE_IGNORED_OVERRIDE` as an environment variable to override the default endpoint, and then ignoring it.  If you do not want an override, then an empty name is what you want:

```cc
opts = google::cloud::internal::PopulateCommonOptions(std::move(opts), /*endpoint_env_var=*/{}, /*emulator_end_var=*/{}, "pubsublite.googleapis.com");
```
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 135 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gRanE-_Cv0GwDjDUa:-N0gRanFC5pqEON9z7jv:bg0exeb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/publisher_connection.cc#L135)):*
> ```cpp
>       MakeBackgroundThreadsFactory(opts)();
>   CompletionQueue cq = background_threads->cq();
>   std::shared_ptr<BackoffPolicy const> backoff_policy =
> ```

In other libraries the backoff policy is one of the options the application can provide in `opts`.  I think we would want the same here.  To unblock you, please file a bug (e.g. "Consider supporting BackoffPolicyOption in Pub/Sub Lite"), and add a TODO here.

nit: Despite the additional `const`, I would have used `auto const backoff_policy = std::make_shared<...>`.
___
*[`google/cloud/pubsublite/internal/default_publish_message_transformer.h` line 28 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gSExZ44EqNaWX2rAW:-N0gSExZ44EqNaWX2rAX:bxry2en) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/internal/default_publish_message_transformer.h#L28)):*
> ```objc
> GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
> 
> constexpr char kEventTimestampAttribute[] =
> ```

Hmm... I think just a function returning `std::string` would be fine, we do not need to overoptimize this.
___
*[`google/cloud/pubsublite/internal/default_publish_message_transformer.cc` line 43 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gSYZA7VYjXQ99YGqF:-N0gSYZA7VYjXQ99YGqG:b-kqht22) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/internal/default_publish_message_transformer.cc#L43)):*
> ```cpp
>                       "Not able to parse event time."};
>       }
>     } else {
> ```

most often we would add a `continue` and avoid the `} else {`.
___
*[`google/cloud/pubsublite/internal/stream_retry_policy.h` line 44 at r14](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YxqALtmEVZVfi58tV-r14-44:-N0gT4I54zjRh2zYvIct:b-c31f6h) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a2cd7b746dc1f0e5684193871b3905f97980bfac/google/cloud/pubsublite/internal/stream_retry_policy.h#L44)):*
<details><summary><i>Previously, dpcollins-google wrote…</i></summary><blockquote>

has no state, just put inline: `std::hash<std::uint8_t>{}(...)`
</blockquote></details>

Any reason why `std::hash<int>{}` would not work here? 
___
*[`google/cloud/pubsublite/internal/stream_retry_policy.h` line 47 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gToEDB2eTQyLVSpeh:-N0gToEDB2eTQyLVSpei:bm2jkr6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/internal/stream_retry_policy.h#L47)):*
> ```objc
>     }
>   };
>   std::unordered_set<StatusCode, StatusCodeHash> retryable_codes_{
> ```

There are some surprising status codes in this list.  I am sure the list is correct, or at least, I am sure you know better than I do what should be in the list.  It would be useful to link a document or something so our future selves are not surprised by this.
___
*[`google/cloud/pubsublite/internal/stream_retry_policy_test.cc` line 22 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gURMi9YW6hmh6U-z9:-N0gURMi9YW6hmh6U-zA:b7iib38) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/internal/stream_retry_policy_test.cc#L22)):*
> ```cpp
> namespace pubsublite_internal {
> 
> constexpr unsigned int kNumStatusCodes = 16;
> ```

errr: https://github.com/googleapis/google-cloud-cpp/blob/79a15e3b1ca7b51d7705005ce0f19f531f9b7f62/google/cloud/status.h#L46 I think that should be `17`.  And I am not sure why you need to make this an `unsigned int`.
___
*[`google/cloud/pubsublite/internal/stream_retry_policy_test.cc` line 26 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gUck22gN4K1g-q6Nj:-N0gUck22gN4K1g-q6Nk:bgshst3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/internal/stream_retry_policy_test.cc#L26)):*
> ```cpp
> TEST(StreamRetryPolicyTest, Codes) {
>   StreamRetryPolicy retry_policy;
>   for (unsigned int i = 0; i != kNumStatusCodes; ++i) {
> ```

Consider:

```cc
struct Case {
  StatusCode code;
   bool expected;
} cases[] = {
    {kOk, false},
    {kCancelled, false},
    {kUnknown, true},
    {kInvalidArgument, false},
    {kDeadlineExceeded, true},
    {kNotFound, false},
    {kAlreadyExists, false},
    {kPermissionDenied, false},
    {kUnauthenticated, false},
    {kResourceExhausted, true},
    {kFailedPrecondition, false},
    {kAborted, true},
    {kOutOfRange, false},
    {kUnimplemented, false},
    {kInternal, true},
    {kUnavailable, true},
    {kDataLoss, false},
};

for (auto const& case : cases) {
  SCOPED_TRACE("Testing " + StatusCodeToString(case.code));
  EXPECT_EQ(retry_policy.OnFailure(Status(case.code, ""), case.expected));
  EXPECT_EQ(retry_policy.IsPermanentFailure(Status(case.code, ""), case.expected));
  EXPECT_FALSE(retry_policy.IsExhausted());
}
```

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736)* status: 14 of 25 files reviewed, 44 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[`google/cloud/pubsublite/publisher_connection.cc` line 168 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YUSSu22b0RXA4dQUp:-N0glqf099wQO6nOt0G4:b-4yqr78) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L168)):*

Alternatively: is there an endpoint we could use when the location cannot be parsed?  Or is there a way we can validate the topic name and take a type that represents a validated name as a parameter?  Yet another alternative: don't validate the location ... just pass it to the server and let it validate whether it is valid or not.
</blockquote></details>

there are no alternative endpoint I'd rather the user get errors directly from the factory function if possible rather than failed RPC stream initializations 
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 135 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gRanE-_Cv0GwDjDUa:-N0gjPZJ9YZvEfpSOxjf:b-vrskwz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/publisher_connection.cc#L135)):*

In other libraries the backoff policy is one of the options the application can provide in `opts`.  I think we would want the same here.  To unblock you, please file a bug (e.g. "Consider supporting BackoffPolicyOption in Pub/Sub Lite"), and add a TODO here.

nit: Despite the additional `const`, I would have used `auto const backoff_policy = std::make_shared<...>`.

</blockquote></details>

https://github.com/googleapis/google-cloud-cpp/issues/8819
___
*[`google/cloud/pubsublite/internal/stream_retry_policy_test.cc` line 22 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gURMi9YW6hmh6U-z9:-N0gjTVZDomNnSJYlqDr:b-b087r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/internal/stream_retry_policy_test.cc#L22)):*

errr: https://github.com/googleapis/google-cloud-cpp/blob/79a15e3b1ca7b51d7705005ce0f19f531f9b7f62/google/cloud/status.h#L46 I think that should be `17`.  And I am not sure why you need to make this an `unsigned int`.
</blockquote></details>

oops I just saw the `15` at the bottom and went off that

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736)* status: 14 of 25 files reviewed, 44 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[`google/cloud/pubsublite/publisher_connection.cc` line 168 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YUSSu22b0RXA4dQUp:-N0go_4P79V0b60oAKf9:brx64yf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L168)):*
> there are no alternative endpoint I'd rather the user get errors directly from the factory function if possible rather than failed RPC stream initializations
*there are no alternative endpoints and I'd rather the user get errors directly from the factory function if possible rather than failed RPC stream initializations

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736)* status: 14 of 25 files reviewed, 44 unresolved discussions (waiting on @18suresha, @coryan, and @dpcollins-google)

___
*[`google/cloud/pubsublite/publisher_connection.cc` line 168 at r8](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N-YUSSu22b0RXA4dQUp:-N0gogQO3MWW9RyzcDDE:b-p6o3vf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/be320809189f40e48281b27da3c58b40120aefaf/google/cloud/pubsublite/publisher_connection.cc#L168)):*
> there are no alternative endpoint I'd rather the user get errors directly from the factory function if possible rather than failed RPC stream initializations

\* there are no alternative endpoints and I'd rather the user get errors directly from the factory function if possible rather than failed RPC stream initializations

coryan (changes requested):
You need to add some TODO entries and cleanup some more nits. I trust you will do that before merging.

> Reviewed 4 of 7 files at r23, 4 of 5 files at r24, 1 of 2 files at r25, 1 of 1 files at r26, 1 of 1 files at r28, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-:-N0h1r-Y1vK4GPnVChlM:b7nn7jo)* status: all files reviewed, 37 unresolved discussions (waiting on @18suresha and @dpcollins-google)

___
*[`google/cloud/pubsublite/publisher_connection.cc` line 117 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gLFbR6kZa4lrR5XN7:-N0gyeTICEzVCVzoMz_y:b292awo) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/publisher_connection.cc#L117)):*

Why 20?  Why not 19, or 21 or 42 or 1000?  There is probably a rationale for this value, it would be good to have it documented.  If it is just "experience shows" then that is fine.
</blockquote></details>

Sorry, but the comment does not answer the question. Why is 2,000 enough? Why not 1,600 or 100?
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 45 at r28](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gyUGK4kfXhE3kj3Tq:-N0gyUGLBSPOJcvN4FXm:bcjugsh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f1cc35891c6ae3b1cc118f0f291e65d4539d86c/google/cloud/pubsublite/publisher_connection.cc#L45)):*
> ```cpp
> using google::cloud::pubsub::PublisherConnection;
> using google::cloud::pubsub_internal::ContainingPublisherConnection;
> ```

Please fix all the other blank lines too.
___
*[`google/cloud/pubsublite/publisher_connection.cc` line 137 at r28](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gyyrV6aR_sDDQseN_:-N0gyyrV6aR_sDDQseNa:b-7vzcre) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7f1cc35891c6ae3b1cc118f0f291e65d4539d86c/google/cloud/pubsublite/publisher_connection.cc#L137)):*
> ```cpp
>   CompletionQueue cq = background_threads->cq();
> 
>   // TODO(18suresha): consider supporting `BackoffPolicyOption`
> ```

Our TODO entries typically point to the bug, this should be `TODO(#8819)`
___
*[`google/cloud/pubsublite/internal/stream_retry_policy.h` line 47 at r22](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-N0gToEDB2eTQyLVSpeh:-N0gzeUa87w58F08_78C:b-52g1bl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/04d28bb0217234c5104af5f2c87053728b6c0869/google/cloud/pubsublite/internal/stream_retry_policy.h#L47)):*

There are some surprising status codes in this list.  I am sure the list is correct, or at least, I am sure you know better than I do what should be in the list.  It would be useful to link a document or something so our future selves are not surprised by this.
</blockquote></details>

Hmm... Okay, that is what Java does, but the service definition is:


https://github.com/googleapis/googleapis/blob/22d09aeac7301ed6f89db3941d939d99b9ea5da8/google/cloud/pubsublite/v1/pubsublite_grpc_service_config.json#L35-L39

which does not include `kResourceExhausted`, nor does it include `kCancelled`.   Furthermore, the guidelines for gRPC:

https://grpc.github.io/grpc/core/md_doc_statuscodes.html

suggest that `kResourceExhausted` may be retryable if the server generates it, but not retryable if generated by the gRPC library.

Please open another bug with a TODO so you can make progress and we can resolve those details elsewhere.

Pro Tip: There is a way to create bugs from a GitHub comment.



_Code quote:_
```objc
https://github.com/googleapis/java-pubsublite/blob/main/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/ErrorCodes.java
```

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8736#-:-N0h1zbq3VTLBgQU2h29:bnfp4nl)* status: all files reviewed, 37 unresolved discussions (waiting on @18suresha and @dpcollins-google)

PR #8738, opened 2022-04-12
impl(rest): better errors in SignUsingSha256()

coryan:
Add the OpenSSL errors if there are any failures. Also some (I think)
clean ups on the use of `std::unique_ptr<>`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8738)

### DIFF
--- a/google/cloud/internal/openssl_util.cc
+++ b/google/cloud/internal/openssl_util.cc
@@ -57,6 +57,7 @@ std::unique_ptr<EVP_MD_CTX, OpenSslDeleter> GetDigestCtx() {
 
 std::string CaptureSslErrors() {
   std::string msg;
+  char const* sep = "";
   while (auto code = ERR_get_error()) {
     // OpenSSL guarantees that 256 bytes is enough:
     //   https://www.openssl.org/docs/man1.1.1/man3/ERR_error_string_n.html
@@ -65,7 +66,9 @@ std::string CaptureSslErrors() {
     auto constexpr kMaxOpenSslErrorLength = 256;
     std::array<char, kMaxOpenSslErrorLength> buf{};
     ERR_error_string_n(code, buf.data(), buf.size());
+    msg += sep;
     msg += buf.data();
+    sep = ", ";
   }
   return msg;
 }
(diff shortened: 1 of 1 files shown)


google/cloud/internal/openssl_util.cc
```diff
@@ -16,118 +16,142 @@
...
+  std::string msg;
+  while (auto code = ERR_get_error()) {
+    // OpenSSL guarantees that 256 bytes is enough:
+    //   https://www.openssl.org/docs/man1.1.1/man3/ERR_error_string_n.html
+    //   https://www.openssl.org/docs/man1.0.2/man3/ERR_error_string_n.html
+    // we could not find a macro or constant to replace the 256 literal.
+    auto constexpr kMaxOpenSslErrorLength = 256;
+    std::array<char, kMaxOpenSslErrorLength> buf{};
+    ERR_error_string_n(code, buf.data(), buf.size());
+    msg += buf.data();
```
devbww, line 68: Should there be some kind of separator?

google/cloud/internal/openssl_util.cc
```diff
@@ -16,118 +16,142 @@
...
-  if (digest_sign_success_code !=
-      EVP_DigestSignFinal(digest_ctx.get(),
-                          nullptr,  // unsigned char *sig
-                          &signed_str_size)) {
+  // The OpenSSL "C" APIs uses `size_t`, I (coryan) believe all platforms we
+  // support satisfy this assertion, and suspect this is true in all platforms.
+  // However, if the code ever finds its way to one where it is not true, it
+  // would be better to get a compile-time error.
+  static_assert(std::is_same<std::size_t, size_t>::value,
+                "Expect std::size_t == size_t");
```
devbww, line 132: The C++ standard requires that ::size_t and std::size_t are identical.

google/cloud/internal/openssl_util.cc
coryan: Done.

google/cloud/internal/openssl_util.cc
coryan: I suspected that would be the case, but TIL.

devbww (approved):

Issue #8621, opened 2022-03-30
Remove `internal::ResumableUploadSession`


coryan:
## Objective

Fix the current problems with resumable uploads in the storage library, and change the code to make such problems less
likely in the future.

## Overview

This proposes removing `ResumableUploadSession` from the implementation. This will happen in multiple steps, the PRs
would get too large otherwise. These steps are probably going to change as we discover problems. With those caveats we
can outline them as:

1. Change the functions that create a `ResumableUploadSession` to return both the session and the current state of the
   session.
2. Change the client library to stop using `last_response()`, `done()`, or
   `next_expected_byte()`. I expect at this point the bug report in [#8559] will be fixed.
3. Add new member functions to `RawClient` that implement `UploadChunk()` and
   `UploadFinalChunk()`.
    - These functions will receive the offset as an input.
    - At this point the implementation of `RetryResumableUploadSession` will effectively move to `RetryClient`.
4. Remove `ResumableUploadSession`.

[#8559]: https://github.com/googleapis/google-cloud-cpp/issues/8559

## Background

Resumable uploads allow applications to upload objects to GCS even if the connection to GCS is lost, or even if the
application performing the upload restarts. The latter requires the source data to be persistent and the application to
checkpoint its intent.

At a high level one starts a resumable upload by creating a "session". In REST these are represented as a URI, in gRPC
these are a simple string. The client library always represents them as a string, and this is exposed in the application
developers. As noted above, these sessions are persistent. Their current state can be queried using a simple RPC (or a
PUT request without payload). This allows an application to resume an upload even after a shutdown.

When querying a resumable upload the response includes the last committed byte. If the upload was finalized, the
response includes the full object metadata. This allows applications to recover, even if they crash immediately after
finalizing an upload.

A description of resumable uploads for REST can be found at:
https://cloud.google.com/storage/docs/performing-resumable-uploads

A description of the gRPC analog can be found in the proto file. Pay particular attention to the documentation
for `WriteObject()`,
`WriteObjectRequest`, `StartResumablewrite()` and `QueryResumableWrite()`:
https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto

Resumable uploads make progress by sending "chunks", either via a single PUT request in REST-based transports, or via a
client-side streaming RPC for gRPC-based transports. All the chunks in an upload, except the last one, must be sized in
multiples of the upload quantum (256 KiB as we write this).

[...]

PR #8747, opened 2022-04-13
impl(storage): support for partial writes in `UploadChunkRequest`

coryan:
This changes the `internal::UploadChunkRequest` class to support partial
writes. The class remains immutable, but now provides a function to
return a new `UploadChunkRequest` with the remaining data (but the
original options) after a partial write.

Part of the work for #8621

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8747)

### DIFF
--- a/google/cloud/storage/internal/curl_resumable_upload_session_test.cc
+++ b/google/cloud/storage/internal/curl_resumable_upload_session_test.cc
@@ -78,16 +78,16 @@ TEST(CurlResumableUploadSessionTest, Simple) {
                   "none-match-etag");
         EXPECT_EQ(test_url, request.upload_session_url());
         EXPECT_THAT(request.payload(), MatchesPayload(payload));
-        EXPECT_EQ(0, request.source_size());
-        EXPECT_EQ(0, request.range_begin());
+        EXPECT_FALSE(request.upload_size().has_value());
+        EXPECT_EQ(0, request.offset());
         return make_status_or(ResumableUploadResponse{
             "", ResumableUploadResponse::kInProgress, size, {}, {}});
       })
       .WillOnce([&](UploadChunkRequest const& request) {
         EXPECT_EQ(test_url, request.upload_session_url());
         EXPECT_THAT(request.payload(), MatchesPayload(payload));
-        EXPECT_EQ(2 * size, request.source_size());
-        EXPECT_EQ(size, request.range_begin());
+        EXPECT_EQ(2 * size, request.upload_size().value_or(0));
+        EXPECT_EQ(size, request.offset());
         return make_status_or(ResumableUploadResponse{
             "", ResumableUploadResponse::kDone, 2 * size, {}, {}});
       });
@@ -168,8 +168,9 @@ TEST(CurlResumableUploadSessionTest, Empty) {
       .WillOnce([&](UploadChunkRequest const& request) {
         EXPECT_EQ(test_url, request.upload_session_url());
         EXPECT_THAT(request.payload(), MatchesPayload(payload));
-        EXPECT_EQ(0, request.source_size());
-        EXPECT_EQ(0, request.range_begin());
+        EXPECT_TRUE(request.upload_size().has_value());
+        EXPECT_EQ(0, request.upload_size().value_or(0));
+        EXPECT_EQ(0, request.offset());
         return make_status_or(ResumableUploadResponse{
             "", ResumableUploadResponse::kDone, size, {}, {}});
       });
--- a/google/cloud/storage/internal/curl_resumable_upload_session_test.cc
+++ b/google/cloud/storage/internal/curl_resumable_upload_session_test.cc
@@ -86,6 +86,7 @@ TEST(CurlResumableUploadSessionTest, Simple) {
       .WillOnce([&](UploadChunkRequest const& request) {
         EXPECT_EQ(test_url, request.upload_session_url());
         EXPECT_THAT(request.payload(), MatchesPayload(payload));
+        EXPECT_TRUE(request.upload_size().has_value());
         EXPECT_EQ(2 * size, request.upload_size().value_or(0));
         EXPECT_EQ(size, request.offset());
         return make_status_or(ResumableUploadResponse{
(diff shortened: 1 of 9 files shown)


coryan:
Build errors fixed, PTAL.

google/cloud/storage/internal/curl_resumable_upload_session_test.cc
```diff
@@ -78,16 +78,16 @@ TEST(CurlResumableUploadSessionTest, Simple) {
...
+        EXPECT_EQ(0, request.offset());
         return make_status_or(ResumableUploadResponse{
             "", ResumableUploadResponse::kInProgress, size, {}, {}});
       })
       .WillOnce([&](UploadChunkRequest const& request) {
         EXPECT_EQ(test_url, request.upload_session_url());
         EXPECT_THAT(request.payload(), MatchesPayload(payload));
-        EXPECT_EQ(2 * size, request.source_size());
-        EXPECT_EQ(size, request.range_begin());
+        EXPECT_EQ(2 * size, request.upload_size().value_or(0));
```
devbww, line 89: Nit: Below we also expect `request.upload_size().has_value()` before the `value_or()`.

google/cloud/storage/internal/object_requests.cc
```diff
@@ -408,16 +408,25 @@ std::string UploadChunkRequest::RangeHeader() const {
...
-    os << "/" << source_size();
+    os << "/" << *upload_size();
   }
   return std::move(os).str();
 }
 
+UploadChunkRequest UploadChunkRequest::RemainingChunk(
+    std::uint64_t new_offset) const {
+  UploadChunkRequest result = *this;
+  if (new_offset < offset()) return result;
```
devbww, line 424: Opt: Would it be clearer to use `result.offset_` there?

google/cloud/storage/internal/object_requests.h
```diff
@@ -403,30 +403,32 @@ class UploadChunkRequest
...
         full_object_hashes_(std::move(full_object_hashes)) {}
 
   std::string const& upload_session_url() const { return upload_session_url_; }
-  std::uint64_t range_begin() const { return range_begin_; }
-  std::uint64_t range_end() const { return range_begin_ + payload_size() - 1; }
-  std::uint64_t source_size() const { return source_size_; }
-  bool last_chunk() const { return last_chunk_; }
-  std::size_t payload_size() const { return TotalBytes(payload_); }
+  std::uint64_t offset() const { return offset_; }
+  absl::optional<std::uint64_t> upload_size() const { return upload_size_; }
```
devbww, line 424: `#include "absl/types/optional.h"`

devbww (approved):

Jseph:
*[google/cloud/storage/internal/object_requests.cc, line 424 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8747#-N-dqMyH2gb3zwHwSv_x:-N-dqMyH2gb3zwHwSv_y:b-pkr6mk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/68ec9ca58ad1f7c19f701a9c713d4f5cc7fa8d75/google/cloud/storage/internal/object_requests.cc#L424)):*
> ```cpp
>     std::uint64_t new_offset) const {
>   UploadChunkRequest result = *this;
>   if (new_offset < offset()) return result;
> ```

This behavior seems surprising to me.  As a caller I would probably expect a bad status if this happened.  Consider either returning a status_or with something like  absl::StatusCode::kOutOfRange or documenting this behavior in the function comment.

Jseph:
*[google/cloud/storage/internal/object_requests.cc, line 424 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8747#-N-dqMyH2gb3zwHwSv_x:-N-drL_W1kyCIMmAADi_:byimdwa) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/68ec9ca58ad1f7c19f701a9c713d4f5cc7fa8d75/google/cloud/storage/internal/object_requests.cc#L424)):*
<details><summary><i>Previously, Jseph wrote…</i></summary><blockquote>

This behavior seems surprising to me.  As a caller I would probably expect a bad status if this happened.  Consider either returning a status_or with something like  absl::StatusCode::kOutOfRange or documenting this behavior in the function comment.
</blockquote></details>

Or have consistent behavior with what happens below where we're out of range because `new_offset > offset() + payload_.size()` and just set `result.offset_ = new_offset` and `result.payload_ = {}`.

google/cloud/storage/internal/object_requests.h
coryan: Fixed.

google/cloud/storage/internal/object_requests.cc
coryan: Fixed.

google/cloud/storage/internal/curl_resumable_upload_session_test.cc
coryan: Added here, though it is implicit (here, but not below) as it compares to a different value than the `0` passed to `value_or()`.

coryan (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8747)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @devbww and @Jseph)

___
*[google/cloud/storage/internal/object_requests.cc, line 424 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8747#-N-dqMyH2gb3zwHwSv_x:-N-hMs_r2uq64UX798Uf:bvepsg8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/68ec9ca58ad1f7c19f701a9c713d4f5cc7fa8d75/google/cloud/storage/internal/object_requests.cc#L424)):*
<details><summary><i>Previously, Jseph wrote…</i></summary><blockquote>

Or have consistent behavior with what happens below where we're out of range because `new_offset > offset() + payload_.size()` and just set `result.offset_ = new_offset` and `result.payload_ = {}`.
</blockquote></details>

I did the second.  I think I will end up removing the function from this class and move it to the one place where it will be used.  In that location we guarantee that`new_offset` is in range and return a `Status` otherwise.

PR #8748, opened 2022-04-14
test(spanner): a more flexible PickInstanceConfig()

devbww:
Change PickInstanceConfig() to take a predicate over InstanceConfig
values instead of a regex to match against InstanceConfig.name. This
provides a flexibility that will be needed for an upcoming feature
where we will need to select an instance config based upon more
complex conditions.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8748)

### DIFF
--- a/google/cloud/spanner/testing/pick_instance_config.h
+++ b/google/cloud/spanner/testing/pick_instance_config.h
@@ -17,7 +17,9 @@
 
 #include "google/cloud/spanner/version.h"
 #include "google/cloud/internal/random.h"
-#include <regex>
+#include "google/cloud/project.h"
+#include <google/spanner/admin/instance/v1/spanner_instance_admin.pb.h>
+#include <functional>
 #include <string>
 
 namespace google {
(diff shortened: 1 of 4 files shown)


google/cloud/spanner/testing/pick_instance_config.h
```diff
@@ -17,20 +17,29 @@
...
+ * Returns the name of one instance config that satisfies the given
+ * predicate from amongst all those that exist within the given project.
+ *
+ * If multiple instance configs qualify, the one returned is chosen
+ * at random using the PRNG. If none qualify, the first candidate is
+ * returned. If there are no candidates, the empty string is returned.
+ */
+std::string PickInstanceConfig(
+    Project const& project, google::cloud::internal::DefaultPRNG& generator,
+    std::function<
```
coryan, line 40: Consider a type alias.

coryan (approved):

google/cloud/spanner/testing/pick_instance_config.h
devbww: I did, but it seemed like it never really helped anyone.  No one utters it, for example.

PR #8750, opened 2022-04-14
tests(pubsublite): integration tests

18suresha:
depends on https://github.com/googleapis/google-cloud-cpp/pull/8736

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8750)


coryan (changes requested):

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,166 @@
...
+
+using google::cloud::pubsub::MessageBuilder;
+
+using google::cloud::pubsub::PublisherConnection;
+using google::cloud::pubsublite::v1::CreateTopicRequest;
+using google::cloud::pubsublite::v1::DeleteTopicRequest;
+using google::cloud::pubsublite::v1::ListTopicsRequest;
+using google::cloud::pubsublite_internal::MakeLocation;
+
+int constexpr kNumMessages = 10000;
```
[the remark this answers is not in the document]
18suresha: The single partition publishers batch up messages and send them out, so I want to test that they're able to successfully send batches once they reach capacity and when there are some leftover messages.

The default max number of messages per batch is 1000, and with 3 partitions per test topic, I changed `kNumMessages` to 5000 (3000 + some leftovers).

Does that work better given the reasoning?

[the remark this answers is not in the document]
coryan: Sure.  The running time for this is a bit high:

```
//google/cloud/pubsublite/integration_tests:publisher_integration_test   PASSED in 8.1s
```

It would be nice if it was closer to 2 or 3 seconds.  Optional: consider changing the batch size to something smaller and dropping the number of messages.

Separately: I assume there is a timeout that triggers if the batch is not full?

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,171 @@
...
+    admin_connection_->DeleteTopic(std::move(req));
+  }
+
+  void GarbageCollect() {
+    ListTopicsRequest req;
+    req.set_parent("projects/" + project_id_ + "/locations/" + location_id_);
+    auto topics = admin_connection_->ListTopics(std::move(req));
+    std::string full_topic_prefix =
+        "projects/" + project_id_ + "/locations/" + location_id_ + "/topics/" +
+        "pub-int-test-" +
```
coryan, line 109: I should have mentioned this, but maybe a helper function:

```cc
std::string TestTopicPrefix(std::chrono::system_clock::time_point tp) {
 return blah.
}

...

  topic_prefix_{TestTopicPrefix(std::chrono::system_clock::now())}

...

auto full_topic_prefix = ... + TestTopicPrefix(std::chrono::system_clock::now() - std::chrono::hours(48));

```

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,171 @@
...
+#include <regex>
+
+namespace google {
+namespace cloud {
+namespace pubsublite {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace {
+
+using google::cloud::pubsub::MessageBuilder;
+
```
coryan, line 38: nit: no block separators for these using directives.

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,171 @@
...
+#include <gmock/gmock.h>
+#include <regex>
+
+namespace google {
+namespace cloud {
+namespace pubsublite {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace {
+
+using google::cloud::pubsub::MessageBuilder;
```
coryan, line 37: nit: I think our using directives usually are anchored, e.g., `using ::google::cloud::pubsub::MessageBuilder;`

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,171 @@
...
+
+using google::cloud::pubsub::PublisherConnection;
+using google::cloud::pubsublite::v1::CreateTopicRequest;
+using google::cloud::pubsublite::v1::DeleteTopicRequest;
+using google::cloud::pubsublite::v1::ListTopicsRequest;
+using google::cloud::pubsublite_internal::MakeLocation;
+
```
coryan, line 44: since we already have these many directives, consider adding:

```cc
using ::std::chrono::system_clock;
using ::std::chrono::hours;
```

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,171 @@
...
+auto constexpr kNumMessages = 5000;
+auto constexpr kThroughputCapacityMiB = 4;
+auto constexpr kGiB = static_cast<std::int64_t>(1024 * 1024 * 1024LL);
+auto constexpr kPartitionStorage = 30 * kGiB;
+
+class PublisherIntegrationTest : public testing_util::IntegrationTest {
+ protected:
+  PublisherIntegrationTest()
+      : tp_{std::chrono::system_clock::now()},
```
coryan, line 54: Seems like this is only used in the next line, please remove it.

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,171 @@
...
+  std::string project_id_;
+  std::string location_id_;
+  std::shared_ptr<AdminServiceConnection> admin_connection_;
+  std::string topic_name_;
+  std::unique_ptr<PublisherConnection> publisher_;
+};
+
+TEST_F(PublisherIntegrationTest, BasicGoodWithoutKey) {
+  std::vector<future<StatusOr<std::string>>> results;
+  for (unsigned int i = 0; i != kNumMessages; ++i) {
```
coryan, line 146: I missed these earlier.  Do they need to be `unsigned int i` and `unsigned i`?  Why not just `int`?

coryan (commented):
Sorry I missed a few things in the earlier review.  It is pretty close mostly nits / small improvements.

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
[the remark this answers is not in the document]
18suresha: > consider changing the batch size

I set the max batch size to `250` via the options passed to the factory.

> I assume there is a timeout that triggers

Not exactly, what happens is that the single partition publisher gathers up all the individual messages and batches them up. If the first four 4 batches are full and the 5th batch isn't, there are still 5 batches that will be sent out (the 5th batch just isn't full but that's fine).

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
18suresha: typo
`int` is fine

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,173 @@
...
+using ::std::chrono::hours;
+using ::std::chrono::system_clock;
+
+auto constexpr kNumMessages = 125;
+auto constexpr kThroughputCapacityMiB = 4;
+auto constexpr kGiB = static_cast<std::int64_t>(1024 * 1024 * 1024LL);
+auto constexpr kPartitionStorage = 30 * kGiB;
+auto constexpr kMaxNumMessagesPerBatch = 25;
+
+std::string TestTopicPrefix(std::chrono::system_clock::time_point tp) {
```
coryan, line 52: nit: since we added the `using` stuff above, we can save some typing here.

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,173 @@
...
+      }
+    }
+  }
+
+  std::string RandomTopicName() {
+    auto generator =
+        google::cloud::internal::DefaultPRNG(std::random_device{}());
+    return topic_prefix_ +
+           google::cloud::internal::Sample(
+               generator, 42, "abcdefghijlkmnopqrstuvwxyz0123456789_-") +
```
coryan, line 129: It would be nice to document where the magic `42` comes from.  Probably something to do with the length of the prefix and the maximum topic id length?

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -0,0 +1,173 @@
...
+    admin_connection_->DeleteTopic(std::move(req));
+  }
+
+  void GarbageCollect() {
+    ListTopicsRequest req;
+    req.set_parent("projects/" + project_id_ + "/locations/" + location_id_);
+    auto topics = admin_connection_->ListTopics(std::move(req));
+    std::string full_topic_prefix =
+        "projects/" + project_id_ + "/locations/" + location_id_ + "/topics/" +
+        TestTopicPrefix(std::chrono::system_clock::now() -
```
coryan, line 112: nit: maybe just `system_clock::now() - hours(48)`, same think in other places.

coryan (approved):
Some small nits, please fix before merging.

PR #8754, opened 2022-04-14
cleanup: split google/cloud/storage/CMakeLists.txt file

coryan:
This file was too large, it was hard to find the code related to each
library (we have two, maybe three soon).  I created 2 `.cmake` files,
one for each library, and included them from the `.txt` file. The file
for each library first defines the library, including install targets
(or commands), and then any testing targets.


google/cloud/storage/google_cloud_cpp_storage_grpc.cmake
```diff
@@ -0,0 +1,205 @@
...
+            COMPONENT google_cloud_cpp_development)
+# With CMake-3.12 and higher we could avoid this separate command (and the
+# duplication).
+install(
+    TARGETS google_cloud_cpp_storage_grpc
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+            COMPONENT google_cloud_cpp_development
+            NAMELINK_ONLY
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+            COMPONENT google_cloud_cpp_development)
```
dbolduc, line 152: ordering nit: this block is moved above the `google_cloud_cpp_install_headers()` call in the other file. move if you like.

dbolduc (approved):

google/cloud/storage/google_cloud_cpp_storage_grpc.cmake
coryan: Fixed.

PR #8759, opened 2022-04-15
fix(pubsublite): cleanup client on multipartition publisher failure

18suresha:
Without this change, then `Publish` futures created before the first num-partitions poll fail will only be satisfied once `Shutdown` is called. Additionally, this change shuts down the alarm on first poll failure. If the alarm kept running, it would be useless.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8759)

### DIFF
--- a/google/cloud/pubsublite/internal/multipartition_publisher.cc
+++ b/google/cloud/pubsublite/internal/multipartition_publisher.cc
@@ -62,6 +62,19 @@ future<Status> MultipartitionPublisher::Start() {
   return start;
 }
 
+void MultipartitionPublisher::CleanupOutstandingResources(
+    Status const& status) {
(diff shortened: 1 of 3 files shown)


dpcollins-google:
I don't think this should be merged. The specified behavior: `Publish futures created before the first num-partitions poll fails will only be satisfied once Shutdown is called` is WAI, the client has permanently failed and they must discard the connection (and call shutdown). They can detect the error through the permanent error handler.

devjgm (approved):

18suresha:
closed because the premise of why we would want this PR also applies to `PartitionPublisher` and can be resolved by `FailureHandler`
if this change later becomes required though, a broader restructuring should be made to this layer and lower layers

PR #8760, opened 2022-04-15
fix(pubsublite): async ordering fixes on resumable stream and lifetime of underlying stream

18suresha:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8760)

### DIFF
--- a/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
+++ b/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
@@ -215,6 +215,60 @@ TEST_F(ResumableAsyncReadWriteStreamingRpcTest,
 }
 
 TEST_F(ResumableAsyncReadWriteStreamingRpcTest,
+       SingleStartFailurePermanentErrorWithAsyncLoop) {
+  struct Loop {
+    // this should run if and only if the stream is not shutdown
+    void InvokeLoop() {
+      stream->Read().then([this](future<absl::optional<FakeResponse>>) {
+        to_call();
+        if (status_future.is_ready() && !status_future.get().ok()) return;
+        InvokeLoop();
+      });
+    }
+    ResumableAsyncReadWriteStream stream;
+    std::function<void()> to_call;
+    future<Status> status_future;
+  };
+  InSequence seq;
+
+  auto stream1 = absl::make_unique<StrictMock<AsyncReaderWriter>>();
+  auto& stream1_ref = *stream1;
+
+  auto mock_retry_policy = absl::make_unique<StrictMock<MockRetryPolicy>>();
+  auto& retry_policy_ref = *mock_retry_policy;
+  EXPECT_CALL(retry_policy_factory_, Call)
+      .WillOnce(Return(ByMove(std::move(mock_retry_policy))));
+
+  auto backoff_policy = absl::make_unique<StrictMock<MockBackoffPolicy>>();
+  EXPECT_CALL(*backoff_policy_, clone)
+      .WillOnce(Return(ByMove(std::move(backoff_policy))));
+
+  EXPECT_CALL(stream_factory_, Call)
+      .WillOnce(Return(ByMove(std::move(stream1))));
+
+  promise<bool> start_promise;
+  EXPECT_CALL(stream1_ref, Start)
+      .WillOnce(Return(ByMove(start_promise.get_future())));
+
+  auto* stream = stream_.get();
+  StrictMock<MockFunction<void()>> placeholder_func;
+  Loop loop{std::move(stream_), placeholder_func.AsStdFunction(),
+            stream->Start()};
+  loop.InvokeLoop();
+
+  EXPECT_CALL(stream1_ref, Finish)
+      .WillOnce(Return(ByMove(make_ready_future(FailStatus()))));
+  EXPECT_CALL(retry_policy_ref, IsExhausted).WillOnce(Return(false));
+  EXPECT_CALL(retry_policy_ref, OnFailure(FailStatus()))
+      .WillOnce(Return(false));
+  EXPECT_CALL(placeholder_func, Call);
+
+  start_promise.set_value(false);
+  auto shutdown = stream->Shutdown();
+  shutdown.get();
+}
+
+TEST_F(ResumableAsyncReadWriteStreamingRpcTest,
        SingleStartFailureExhaustedPermanentError) {
   InSequence seq;
 
--- a/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
+++ b/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
@@ -244,7 +244,10 @@ TEST_F(ResumableAsyncReadWriteStreamingRpcTest,
   std::function<void()> loop = [&] {
     stream.Read().then([&](future<absl::optional<FakeResponse>>) {
       to_call();
-      if (status_future.is_ready()) return;
+      if (status_future.is_ready()) {
+        EXPECT_FALSE(status_future.get().ok());
+        return;
+      }
       loop();
     });
   };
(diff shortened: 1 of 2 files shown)


google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
```diff
@@ -214,6 +214,60 @@ TEST_F(ResumableAsyncReadWriteStreamingRpcTest,
   shutdown.get();
 }
 
+TEST_F(ResumableAsyncReadWriteStreamingRpcTest,
+       SingleStartFailurePermanentErrorWithAsyncLoop) {
+  struct Loop {
+    // this should run if and only if the stream is not shutdown
+    void InvokeLoop() {
+      stream->Read().then([this](future<absl::optional<FakeResponse>>) {
```
dpcollins-google, line 222: just make this a lambda with [&] capture

dpcollins-google: Actually, just call this inline, and ref capture for the inner lambda

dpcollins-google: ahhhhh its recursive. You can do the following:

```
// Repeatedly call stream->Read asynchronously until status future is ready
std::function<void()> loop;
loop = [&]{
  stream->Read().then([this](auto) {
      to_call();
      if (status_future.is_ready() && !status_future.get().ok()) return;
      loop();
  }
}
```

google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc
```diff
@@ -214,6 +214,60 @@ TEST_F(ResumableAsyncReadWriteStreamingRpcTest,
...
+        to_call();
+        if (status_future.is_ready() && !status_future.get().ok()) return;
```
dpcollins-google, line 224: remove this.

dpcollins-google: `!status_future.get().ok()`, you should stop looping even if it is OK (but fail the test)

dpcollins-google (approved):

coryan:
FYI: https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting

Try to keep your PR titles under 50 characters, and definitely under 72.  Ditto for commit messages.

coryan (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8760)* status: all files reviewed, 5 unresolved discussions (waiting on @18suresha)

___
*[google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h, line 428 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8760#-N-hq9_t861_2jUW4jky:-N-hq9_u8b79pzwLMAX2:b-9jboab) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9208fe66e5af65b72379c144776f3f35b098b1b3/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h#L428)):*
> ```objc
>               root_future.then(ChainFuture(in_progress_write_->get_future()));
>         }
>         std::shared_ptr<AsyncStreamingReadWriteRpc<RequestType, ResponseType>>
> ```

nit: 

```cc
auto stream = std::move(stream_);
```
___
*[google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc, line 234 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8760#-N-hrw_d4x5WSKLEw_2Y:-N-hrw_d4x5WSKLEw_2Z:b-4nequ6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9208fe66e5af65b72379c144776f3f35b098b1b3/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc#L234)):*
> ```cpp
> 
>   EXPECT_CALL(stream_factory_, Call)
>       .WillOnce(Return(ByMove(std::move(stream1))));
> ```

C++ memory management is hard only if you make it so.  Please stop using these `_ref` variables.  You can just write:

```cc
EXPECT_CALL(stream_factory_, Call)
  .WillOnce([&]  {
      auto stream = absl::make_unique<AsyncReaderWriter>();
      EXPECT_CAL(*stream, Start).WillOnce(Return(MyMove(start_promise.get_future()));
      return stream;
  });
```
___
*[google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc, line 258 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8760#-N-hsmtMCk7doa6dbSQA:-N-hrFIL9vJxo9CLA6AM:b-refaec) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9208fe66e5af65b72379c144776f3f35b098b1b3/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc#L258)):*
> ```cpp
>   EXPECT_CALL(stream1_ref, Finish)
>       .WillOnce(Return(ByMove(make_ready_future(FailStatus()))));
>   EXPECT_CALL(retry_policy_ref, IsExhausted).WillOnce(Return(false));
> ```

As above, seems like this is just complicating memory management.

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8760)* status: all files reviewed, 5 unresolved discussions (waiting on @18suresha and @coryan)

___
*[google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc, line 234 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8760#-N-hrw_d4x5WSKLEw_2Y:-N-hua8z5R9yNcofTakV:birho46) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9208fe66e5af65b72379c144776f3f35b098b1b3/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc_test.cc#L234)):*

C++ memory management is hard only if you make it so.  Please stop using these `_ref` variables.  You can just write:

```cc
EXPECT_CALL(stream_factory_, Call)
  .WillOnce([&]  {
      auto stream = absl::make_unique<AsyncReaderWriter>();
      EXPECT_CAL(*stream, Start).WillOnce(Return(MyMove(start_promise.get_future()));
      return stream;
  });
```

</blockquote></details>

there's an `EXPECT_CALL(stream1_ref, Finish).WillOnce(Return(ByMove(make_ready_future(FailStatus()))));` at the end of the test case
we can just make that expectation in the lambda (and do the same for other references we have) but that will prevents us from leveraging `InSequence seq` on the whole test case since many of the expectations don't happen at object instantiation.

The validity of these references is dependent on the validity of the class we're testing. Expectations on these references should work if we implemented the class correctly, otherwise it will lead to memory errors that are likely to fail the test and/or be caught by sanitizers. Thus, if the use of these references cause errors, we know that there's something wrong with the implementation of our class.

18suresha (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8760)* status: all files reviewed, 5 unresolved discussions (waiting on @18suresha and @coryan)

___
*[google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h, line 428 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8760#-N-hq9_t861_2jUW4jky:-N-i3Ume14nQfR1pm0H9:b-z37flv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9208fe66e5af65b72379c144776f3f35b098b1b3/google/cloud/pubsublite/internal/resumable_async_streaming_read_write_rpc.h#L428)):*

nit: 

```cc
auto stream = std::move(stream_);
```

</blockquote></details>

`stream_` is a unique ptr and we're explicitly moving it into a shared ptr

coryan (approved):
I think you are too worried about the sequence of your mock calls.  I expect I will need to rewrite all these tests to be simpler to reason about.

PR #8764, opened 2022-04-15
refactor(GCS+gRPC): write resumable session using RU

coryan:
This continues moving the implementation of `ResumableUploadSession` in
terms of session-less "resumable uploads".  The expectation is that
a session-less implementation will be easier to reason about and have
fewer problems when resuming uploads.

Part of the work for #8621

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8764)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -119,17 +119,16 @@ std::unique_ptr<GrpcClient::WriteObjectStream> GrpcClient::CreateUploadWriter(
 
 StatusOr<ResumableUploadResponse> GrpcClient::QueryResumableSession(
     QueryResumableUploadRequest const& request) {
-  OptionsSpan span(options_);
-  grpc::ClientContext context;
-  ApplyQueryParameters(context, request, "resource");
-  auto const timeout = options_.get<TransferStallTimeoutOption>();
-  if (timeout.count() != 0) {
-    context.set_deadline(std::chrono::system_clock::now() + timeout);
-  }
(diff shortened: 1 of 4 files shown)


google/cloud/storage/internal/grpc_client.cc
```diff
@@ -457,26 +475,22 @@ StatusOr<CreateResumableSessionResponse> GrpcClient::CreateResumableSession(
...
 }
 
 StatusOr<QueryResumableUploadResponse> GrpcClient::QueryResumableUpload(
-    QueryResumableUploadRequest const&) {
-  return Status(StatusCode::kUnimplemented, "TODO(#8621)");
+    QueryResumableUploadRequest const& request) {
+  OptionsSpan span(options_);
+  grpc::ClientContext context;
+  ApplyQueryParameters(context, request, "resource");
+  auto const timeout = options_.get<TransferStallTimeoutOption>();
```
devbww, line 486: Nitty nit ... and it is only in carried-forward code ... and it doesn't look like it makes a difference in this case ... and it looks like there are similar instances, so it undoubtedly shouldn't be address here ... but, if we're going to instantiate an `OptionsSpan`, we should probably use `CurrentOptions()` for access.

devbww (approved):

google/cloud/storage/internal/grpc_client.cc
coryan: I am sure you are right, I created #8770 let's continue the conversation there.

PR #8765, opened 2022-04-15
fix(storage): build on MSVC+x86

coryan:
Part of the work for #8621

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8765)

### DIFF
--- a/google/cloud/storage/internal/object_requests.cc
+++ b/google/cloud/storage/internal/object_requests.cc
@@ -426,7 +426,8 @@ UploadChunkRequest UploadChunkRequest::RemainingChunk(
     result.payload_.clear();
     return result;
   }
-  // No chunk can be larger than what fits in memory, and thus this
+  // No chunk can be larger than what fits in memory, and because `new_offset`
+  // is the [offset_, offset_ + payload_size()], this static_cast<> is safe:
   PopFrontBytes(result.payload_,
                 static_cast<std::size_t>(new_offset - result.offset_));
   result.offset_ = new_offset;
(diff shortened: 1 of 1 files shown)


google/cloud/storage/internal/object_requests.cc
```diff
@@ -421,12 +421,14 @@ std::string UploadChunkRequest::RangeHeader() const {
...
     std::uint64_t new_offset) const {
   UploadChunkRequest result = *this;
-  if (new_offset < offset_) {
+  if (new_offset < offset_ || new_offset >= offset_ + payload_size()) {
     result.offset_ = new_offset;
     result.payload_.clear();
     return result;
   }
-  PopFrontBytes(result.payload_, new_offset - result.offset_);
+  // No chunk can be larger than what fits in memory, and thus this
```
devbww, line 429: "And thus this ..."?

coryan: Fixed.

google/cloud/storage/internal/object_requests.cc
```diff
@@ -421,12 +421,15 @@ std::string UploadChunkRequest::RangeHeader() const {
...
   UploadChunkRequest result = *this;
-  if (new_offset < offset_) {
+  if (new_offset < offset_ || new_offset >= offset_ + payload_size()) {
     result.offset_ = new_offset;
     result.payload_.clear();
     return result;
   }
-  PopFrontBytes(result.payload_, new_offset - result.offset_);
+  // No chunk can be larger than what fits in memory, and because `new_offset`
+  // is the [offset_, offset_ + payload_size()], this static_cast<> is safe:
```
devbww, line 430: Nit: That interval is open on the end, so `)` would be a better closer.

s/is the/is in/

devbww (approved):

google/cloud/storage/internal/object_requests.cc
devbww: Oops ... auto merge.

PR #8771, opened 2022-04-15
refactor(spanner): extract instance ID generation into helper func

devbww:
This facilitates an upcoming change where we will need to generate
another form of instance ID, which has a different prefix.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8771)

### DIFF
--- a/google/cloud/spanner/testing/random_instance_name.h
+++ b/google/cloud/spanner/testing/random_instance_name.h
@@ -24,8 +24,8 @@ namespace cloud {
 namespace spanner_testing {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-/// Create a random instance name given a PRNG generator.
-std::string RandomInstanceName(google::cloud::internal::DefaultPRNG& generator);
+/// Returns a random instance ID given a PRNG generator.
+std::string RandomInstanceName(internal::DefaultPRNG&);
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace spanner_testing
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/testing/random_instance_name.h
```diff
@@ -24,8 +24,8 @@ namespace cloud {
 namespace spanner_testing {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-/// Create a random instance name given a PRNG generator.
-std::string RandomInstanceName(google::cloud::internal::DefaultPRNG& generator);
+/// Returns a random instance ID given a PRNG generator.
+std::string RandomInstanceName(internal::DefaultPRNG&);
```
dbolduc, line 28: do you want to change this to `RandomInstanceId` ?

dbolduc (approved):
Also, I notice that there is a `testing/random_database_name_test.cc` but no `testing/random_instance_name_test.cc`. Your call on whether or not to add unit tests for this business.

google/cloud/spanner/testing/random_instance_name.h
devbww: I do want to, yes.  (You may have noticed some other name changes I've been making to try to reduce confusion.)  However, in this case I decided that changing the function/file name wasn't worth it.  I might revisit that choice later though.

devbww:
> Also, I notice that there is a `testing/random_database_name_test.cc` but no `testing/random_instance_name_test.cc`. Your call on whether or not to add unit tests for this business.

Yeah ... in the database case there are some conditions between names, prefixes, and prefix regexs that are amenable to testing.  There is nothing directly like that in the instance case, so no test.

[That said, we usually draw the line at testing the test code.  It is supposed to be simple enough to not need it.]

PR #8773, opened 2022-04-17
fix(pubsublite): multipartition publisher ordering

18suresha:
in the following case:
thread 1: call `Publish` on `5` messages before publishers are instantiated
alarm thread: instantiate publishers and start loop of calling `PartitionPublisher::Publish` on above messages
alarm thread: call `PartitionPublisher::Publish` on first message
thread 1: call `Publish` on `6`th message which is `PartitionPublisher::Publish`ed due to publishers having been instantiated
alarm thread: `2`nd to `5`th messages are `PartitionPublisher::Publish`ed

this leads to incorrect ordering

This PR implements `MultipartitionPublisher::Publish` via an internal buffer to remove these types of race conditions.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8773)

### DIFF
--- a/google/cloud/pubsublite/internal/multipartition_publisher.cc
+++ b/google/cloud/pubsublite/internal/multipartition_publisher.cc
@@ -93,17 +93,17 @@ void MultipartitionPublisher::HandleNumPartitions(
     new_partition_publishers.push_back(publisher_factory_(partition));
     service_composite_.AddServiceObject(new_partition_publishers.back().get());
   }
-  std::vector<PublishState> initial_publish_buffer;
   {
     std::lock_guard<std::mutex> g{mu_};
     std::move(new_partition_publishers.begin(), new_partition_publishers.end(),
               std::back_inserter(partition_publishers_));
-    initial_publish_buffer.swap(initial_publish_buffer_);
+    for (auto& state : initial_publish_buffer_) {
+      state.num_partitions = num_partitions;
+      messages_.push_back(std::move(state));
+    }
+    initial_publish_buffer_.clear();
   }
-  for (auto& state : initial_publish_buffer) {
-    state.num_partitions = num_partitions;
-    RouteAndPublish(std::move(state));
-  }
+  PublishLoop();
 }
 
 void MultipartitionPublisher::TriggerPublisherCreation() {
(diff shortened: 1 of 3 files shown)


google/cloud/pubsublite/internal/multipartition_publisher.cc
```diff
@@ -169,24 +164,46 @@ void MultipartitionPublisher::RouteAndPublish(PublishState state) {
...
+    std::lock_guard<std::mutex> g{mu_};
+    if (in_publish_loop_) return;
+    messages_left = !messages_.empty();
+    if (messages_left) in_publish_loop_ = true;
+  }
+  while (messages_left) {
+    PublishState state;
+    {
+      std::lock_guard<std::mutex> g{mu_};
+      state = std::move(messages_.front());
```
devjgm, line 179: Is there some guarantee that `messages_` is not empty at this point, or should there be a check right before calling `.front()` (under the lock) that it's still not empty to avoid the UB of calling front on a potentially empty deque?

google/cloud/pubsublite/internal/multipartition_publisher.cc
```diff
@@ -169,24 +164,46 @@ void MultipartitionPublisher::RouteAndPublish(PublishState state) {
       .then(OnPublish{std::move(state.publish_promise), partition});
 }
 
+void MultipartitionPublisher::PublishLoop() {
```
dpcollins-google, line 167: Can we give this a better name, `TryPublishMessages`?

dpcollins-google, line 167: Suggested restructuring.
```
{
    std::lock_guard<std::mutex> g{mu_};
    if (in_publish_loop_) return;
    in_publish_loop_ = true;
}
while (true) {
  std::deque<PublishState> messages;
  uint32_t num_partitions;
  {
    std::lock_guard<std::mutex> g{mu_};
    if (messages_.empty()) {
      in_publish_loop = false;
      return;
    }
    messages.swap(messages_);
    num_partitions = static_cast<std::uint32_t>(partition_publishers_.size());
  }
  for (PublishState& state : messages) {
    state.num_partitions = num_partitions;
    RouteAndPublish(std::move(state));
  }
}
```

google/cloud/pubsublite/internal/multipartition_publisher.cc
```diff
@@ -169,24 +164,46 @@ void MultipartitionPublisher::RouteAndPublish(PublishState state) {
...
     if (partition_publishers_.empty()) {
-      initial_publish_buffer_.push_back(std::move(state));
-      return initial_publish_buffer_.back().publish_promise.get_future();
+      messages_.push_back(std::move(state));
+      return messages_.back().publish_promise.get_future();
     }
-    state.num_partitions =
-        static_cast<std::uint32_t>(partition_publishers_.size());
+    to_return = state.publish_promise.get_future();
+    messages_.push_back(std::move(state));
```
dpcollins-google, line 204: suggested restructuring:

```
to_return = state.publish_promise.get_future();
messages_.push_back(std::move(state));
if (partition_publishers_.empty()) return to_return;
```

google/cloud/pubsublite/internal/multipartition_publisher.cc
18suresha: by using the `in_publish_loop_` flag variable, only one thread can execute `PublishLoop` at a time, and all other threads can only push to `messages_`
so our non-empty check before the loop and at the end of each iteration guarantee that it's non-empty when we call `front()` on it

18suresha: also restructured the code as per Dan's comments below, so we don't have to worry about this UB

google/cloud/pubsublite/internal/multipartition_publisher.cc
```diff
@@ -169,24 +164,43 @@ void MultipartitionPublisher::RouteAndPublish(PublishState state) {
...
-    // `initial_publish_buffer_` is flushed when publishers are created
-    if (partition_publishers_.empty()) {
-      initial_publish_buffer_.push_back(std::move(state));
-      return initial_publish_buffer_.back().publish_promise.get_future();
-    }
-    state.num_partitions =
-        static_cast<std::uint32_t>(partition_publishers_.size());
+    to_return = state.publish_promise.get_future();
+    messages_.push_back(std::move(state));
+    if (partition_publishers_.empty()) return to_return;
```
dpcollins-google, line 201: nit. add a comment that message will be published whenever partition count is read and publishers are created.

dpcollins-google (approved):

devjgm (approved):

PR #8784, opened 2022-04-25
feat(spanner): add spanner::Value support for Type.type_annotation

devbww:
Add support for passing arguments through the `absl::optional` and
`std::vector` constructors for `spanner::Value`, to be applied when
building the prototypical `T` value used to set the protobuf `Type`
member should the `optional` or the `vector` be empty.

The `optional` case means we also need to pass `T` construction
arguments through `MakeNullValue<T>()`.

The Spanner C++ types are in one-to-one correspondence with the
protobuf `Type.code`.  But `message Type` now has a new field,
`type_annotation`, that disambiguates the SQL type that Spanner
will use for particular values of the C++ type.  Therefore, we need
to be able to set the `type_annotation` field when originating such
values.

Normally this isn't a problem as we have the value at hand, but in
the cases of an empty `optional` or `vector` we need to construct
a `T`.  Previously we used the default constructor to do that, but
now we can pass through construction arguments, which can influence
the chosen `type_annotation`.

For example, `spanner::MakeNullValue<T>(args...)`.

So, legitimize the use of the `MakeTypeProto()` argument value
when building the `message Type`.  Previously we only used its C++
type for overload dispatch.

Also clarify that `type_annotation` does not affect the result of
`TypeProtoIs()`, which only concerns the C++ type.

The first use of these construction arguments will be in an upcoming
change to `spanner::Numeric`, but for now we just test that arguments
can be passed in the right places.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8784)

### DIFF
--- a/google/cloud/spanner/value.h
+++ b/google/cloud/spanner/value.h
@@ -228,9 +228,10 @@ class Value {
    * Constructs a non-null instance if `opt` has a value, otherwise constructs
    * a null instance with the specified type `T`.
    */
-  template <typename T>
-  explicit Value(absl::optional<T> opt)
-      : Value(PrivateConstructor{}, std::move(opt)) {}
+  template <typename T, typename... Args>
+  explicit Value(absl::optional<T> opt, Args&&... args)
+      : Value(PrivateConstructor{}, std::move(opt),
+              std::forward<Args>(args)...) {}
 
   /**
    * Constructs an instance from a Spanner ARRAY of the specified type and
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/value.h
```diff
@@ -228,9 +228,10 @@ class Value {
    * Constructs a non-null instance if `opt` has a value, otherwise constructs
    * a null instance with the specified type `T`.
    */
```
dbolduc, line 230: Do you think we should mention when `args` are used and when they are ignored?

google/cloud/spanner/value.h
```diff
@@ -244,8 +245,9 @@ class Value {
    *   vector's elements must have exactly the same field names. Any mismatch
    *   in in field names results in undefined behavior.
```
dbolduc, line 246: same `args` comment

also just noticed "in in field names"

PR #8792, opened 2022-04-26
cleanup: use new CompileProtos support for CMake proto-list/deps

devbww:
Note: A few `google/cloud/*/CMakeLists.txt` files remain using the
old proto-list/deps lists because of an issue I don't yet understand.
I'll address them separately.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8792)


coryan (approved):
Thanks!

PR #8834, opened 2022-04-30
feat(storage): new option to disable decompressive transcoding

### LINKED ISSUE #8305: Add option to inject `Accept-Encoding: gzip`.
Probably can be accepted by all requests (where it would have no effect), but only relevant for `ReadObject()`.

coryan:
For objects stored in gzip format (and with contentEncoding == "gzip")
GCS over HTTP automatically decompresses the object during download.
Some applications may want to read the object in compressed format.
Support this use-case with a new option (`AcceptEncoding`) for
`Client::ReadObject()`, and a new helper function
(`AcceptEncodingGzip()`) that returns this option with the correct value.

Fixes #8305

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8834)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -1064,8 +1064,8 @@ class Client {
    *     Valid types for this operation include `DisableCrc32cChecksum`,
    *     `DisableMD5Hash`, `IfGenerationMatch`, `EncryptionKey`, `Generation`,
    *     `IfGenerationMatch`, `IfGenerationNotMatch`, `IfMetagenerationMatch`,
-   *     `IfMetagenerationNotMatch`, `ReadFromOffset`, `ReadRange`, `ReadLast`
-   *     and `UserProject`.
+   *     `IfMetagenerationNotMatch`, `ReadFromOffset`, `ReadRange`, `ReadLast`,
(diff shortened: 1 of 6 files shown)


dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8834)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/storage/well_known_headers.h` line 295 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8834#-N0zqTBZD_SYmhX4mAwG:-N0zqTB_06HvQwYRWKxi:b70rgx8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed4e3775b1a1c2a01e7315b0d17d860778adc9dc/google/cloud/storage/well_known_headers.h#L295)):*
> ```objc
>  *     compressed with `gzip` and have their `content_encoding()` attribute
>  *     set accordingly. At the time of this writing GCS does not decompress
>  *     objects stored with other compression algorithms, nor does not it
> ```

s/nor does not it/nor does it/

PR #8850, opened 2022-05-02
fix: correct check for storage vs. rest enabled

coryan:
We cannot enable the storage library and disable the `rest-internal`
library, unfortunately the check was based on the wrong variables.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8850)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,13 +272,18 @@ cmake_dependent_option(
     ON "GOOGLE_CLOUD_CPP_ENABLE_REST_EXPRESSION" OFF)
 mark_as_advanced(GOOGLE_CLOUD_CPP_ENABLE_REST)
 
-# Prevent users for setting incompatible flags.
+# Detect incompatible flags and let the customer know what to do.
 if (storage IN_LIST GOOGLE_CLOUD_CPP_ENABLE AND NOT GOOGLE_CLOUD_CPP_ENABLE_REST)
     message(
         FATAL_ERROR
-            "If storage is enabled (e.g. -DGOOGLE_CLOUD_ENABLE=storage), "
-            "the REST library cannot be disabled "
-            "(e.g. -DGOOGLE_CLOUD_CPP_ENABLE_REST=OFF)")
+            "If storage is enabled (e.g. -DGOOGLE_CLOUD_CPP_ENABLE=storage),"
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -273,7 +273,7 @@ cmake_dependent_option(
 mark_as_advanced(GOOGLE_CLOUD_CPP_ENABLE_REST)
 
 # Prevent users for setting incompatible flags.
-if (GOOGLE_CLOUD_CPP_ENABLE_STORAGE AND NOT GOOGLE_CLOUD_CPP_ENABLE_REST)
+if (storage IN_LIST GOOGLE_CLOUD_CPP_ENABLE AND NOT GOOGLE_CLOUD_CPP_ENABLE_REST)
```
dbolduc, line 276: just asking, do we need this to be a
```
((GOOGLE_CLOUD_CPP_ENABLE_STORAGE OR storage IN_LIST GOOGLE_CLOUD_CPP_ENABLE) AND NOT GOOGLE_CLOUD_CPP_ENABLE_REST)
```

or does `(storage IN_LIST GOOGLE_CLOUD_CPP_ENABLE)` imply `GOOGLE_CLOUD_CPP_ENABLE_STORAGE` ?

coryan: Does not matter what the value of `GOOGLE_CLOUD_CPP_ENABLE_STORAGE` is.  The storage library is compiled if and only if `storage IN_LIST GOOGLE_CLOUD_CPP_ENABLE`.

CMakeLists.txt
```diff
@@ -273,7 +273,7 @@ cmake_dependent_option(
 mark_as_advanced(GOOGLE_CLOUD_CPP_ENABLE_REST)
 
 # Prevent users for setting incompatible flags.
```
dbolduc, line 275: nit: s/for/from/

dbolduc: s/users/customers/ ?

dbolduc (approved):

SamuelMarks:
FYI: @coryan this broke vcpkg, see #8919

SamuelMarks:
FYI: @coryan this broke vcpkg [and other package managers?], see #8919

PR #8856, opened 2022-05-02
feat(spanner): add support for spanner::PgNumeric

devbww:
The Google-Spanner variety of NUMERIC has 29 decimal digits of integer
precision and 9 decimal digits of fractional precision, while the new
PostgreSQL variety has 131072 and 16383 digits respectively (and supports
NaN).

So:
 - Introduce `spanner::Decimal<DecimalMode>`, which is a templated
   refactoring of the old `spanner::Numeric` code but that handles
   either precision specification.
 - Re-brand `spanner::Numeric` as an alias for the Google variety of
   `Decimal`.
 - Add `spanner::PgNumeric` as an alias for the PostgreSQL variety.

This pivots from a previous decision to handle both varieties is a single
C++ type.  Now there are two unrelated types, which will be used to set
both the `Type.code` and `Type.type_annotation` protobuf fields in an
upcoming change to `spanner::Value`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8856)


google/cloud/spanner/numeric.cc
```diff
@@ -103,157 +102,33 @@ void Round(std::deque<char>& int_rep, std::deque<char>& frac_rep,
...
 }  // namespace spanner_internal
 
 namespace spanner {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-constexpr std::size_t Numeric::kIntPrec;
-constexpr std::size_t Numeric::kFracPrec;
-
-Numeric::Numeric() : rep_("0") {}
+template <>
```
coryan, line 123: Maybe we should define

```cc
template <std::size_t IntPrecision, std::size_t FracPrecision, bool HasNaN>
class Decimal {
...
};
```

google/cloud/spanner/numeric.h
```diff
@@ -118,25 +137,195 @@ class Numeric {
...
+                 std::size_t frac_prec);
+void Round(std::deque<char>& int_rep, std::deque<char>& frac_rep,
+           std::size_t frac_prec);
+
+Status InvalidArgument(std::string message);
+Status OutOfRange(std::string message);
+Status DataLoss(std::string message);
+
+template <spanner::DecimalMode Mode>
+StatusOr<spanner::Decimal<Mode>> MakeDecimal(std::string s) {
```
coryan, line 211: Can we refactor this function to the `.cc` file?  Maybe:

```cc
StatusOr<std::string> MakeDecimalRep(std::string s, std::size_t int_precision, std::size_t frac_precision);

template <...>
StatusOr<spanner::Decimal<...>> MakeDecimal(std::string s) {
  auto rep = MakeDecimalRep(std::move(s), Decimal<...>::kIntPrecision, Decimal<...>::kFracPrecision);
  if (!rep) return std::move(rep).status();
  return Decimal<...>(std::move(rep));
}
```

google/cloud/spanner/numeric.h
```diff
@@ -31,83 +40,93 @@ namespace google {
 namespace cloud {
 namespace spanner {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
-class Numeric;  // defined below
+
+// The supported `Decimal` modes.
+enum class DecimalMode {
```
coryan, line 45: As mentioned below, seems like this should be modeled with traits.  I also believe we will need a similar type for BigQuery, with slightly different precision and fraction part (I think):

https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types

Which argues for refactoring to `google/cloud/internal/`

coryan (changes requested):

google/cloud/spanner/numeric.h
devbww: Yes.  Done.

google/cloud/spanner/numeric.h
devbww: We can certainly extract common bits when the time comes.  One thing to note, however, is the mismatch between [max-total-digits + max-fractional-digits] and [max-integral-digits + max-fractional-digits].

google/cloud/spanner/numeric.cc
coryan: Ack.

google/cloud/spanner/numeric.h
coryan: Thanks.

coryan (approved):

PR #8864, opened 2022-05-03
refactor(storage): ListBuckets to use RestClient

scotthart:
When the RestClient is enabled in GCS, ListBuckets now uses the rest library in lieu of GCS  libcurl client.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8864)

### DIFF
--- a/google/cloud/storage/internal/rest_client.cc
+++ b/google/cloud/storage/internal/rest_client.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/storage/internal/rest_client.h"
 #include "google/cloud/storage/internal/bucket_access_control_parser.h"
 #include "google/cloud/storage/internal/bucket_metadata_parser.h"
+#include "google/cloud/storage/internal/bucket_requests.h"
 #include "google/cloud/storage/internal/curl_request_builder.h"
 #include "google/cloud/storage/internal/hmac_key_metadata_parser.h"
 #include "google/cloud/storage/internal/logging_client.h"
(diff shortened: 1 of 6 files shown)


google/cloud/storage/internal/rest_client.cc
```diff
@@ -36,13 +38,34 @@ namespace storage {
...
+namespace rest = google::cloud::rest_internal;
+
+template <typename ReturnType>
+StatusOr<ReturnType> ParseFromRestResponse(
+    StatusOr<std::unique_ptr<rest::RestResponse>> response) {
+  if (!response.ok()) return std::move(response).status();
+  if ((*response)->StatusCode() >= rest::kMinNotSuccess) {
+    return rest::AsStatus(std::move(**response));
+  }
+  return ReturnType::FromRestResponse(std::move(**response));
```
coryan, line 50: Consider:

```cc
template <typename ReturnType>
StatusOr<ReturnType> ParseFromRestResponse(
    StatusOr<std::unique_ptr<rest::RestResponse>> response) {
  if (!response.ok()) return std::move(response).status();
  if ((*response)->StatusCode() >= rest::kMinNotSuccess) {
    return rest::AsStatus(std::move(**response));
  }
  auto r = *std::move(response);
  auto payload = rest::ReadAll(std::move(*r).ExtractPayload());
  if (!payload.ok()) return std::move(payload).status();
  return ReturnType::FromRestResponse(*payload);
}
```

That way you won't have to modify most `*Request` types?

coryan (approved):

PR #8865, opened 2022-05-03
fix(storage): set CURLOPT_BUFFERSIZE once per handle

coryan:
It seems that, despite all existing precautions, we are still getting
errors on `curl_setopt(..., CURLOPT_BUFFERSIZE, value)` because the
option cannot be set until some internal cleanup happens.  Since `value`
is always the same, we can set it when the handle is created, and avoid
the cleanup problems.

Part of the work for #7051

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8865)

### DIFF
--- a/google/cloud/storage/internal/curl_wrappers.cc
+++ b/google/cloud/storage/internal/curl_wrappers.cc
@@ -211,6 +211,15 @@ bool SslLockingCallbacksInstalled() {
 #endif  // GOOGLE_CLOUD_CPP_SSL_REQUIRES_LOCKS
 }
 
+CurlPtr MakeCurlPtr() {
+  auto handle = CurlPtr(curl_easy_init(), &curl_easy_cleanup);
+  // We get better performance using a slightly larger buffer (128KiB) than the
+  // default buffer size set by libcurl (16KiB).  We ignore errors because
+  // failing to set this parameter just affects performance by a small amount.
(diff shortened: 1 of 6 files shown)


dbolduc (approved):

google/cloud/storage/internal/curl_wrappers.cc
```diff
@@ -211,6 +211,15 @@ bool SslLockingCallbacksInstalled() {
 #endif  // GOOGLE_CLOUD_CPP_SSL_REQUIRES_LOCKS
 }
 
+CurlPtr MakeCurlPtr() {
+  auto handle = CurlPtr(curl_easy_init(), &curl_easy_cleanup);
+  // We get better performance using a slightly larger buffer (128KiB) than the
+  // default buffer size set by libcurl (16KiB).  We ignore errors because
+  // failing to set this parameter just affects performance by a small amount.
+  (void)curl_easy_setopt(handle.get(), CURLOPT_BUFFERSIZE, 128 * 1024L);
```
dbolduc, line 219: quick question about `128 * 1024L`... do we want this to be `CURL_MAX_WRITE_SIZE` ?

coryan: No, `CURLOPT_BUFFERSIZE` defaults to 
`CURL_MAX_WRITE_SIZE` and that is set to 16KiB (unless you override it at compile-time in libcurl).

Issue #8798, opened 2022-04-26
Prepare for Bigtable DataConnection implementation


dbolduc:
- [ ] Add new options (in internal) and default them.
- [ ] Add empty `DataConnection` (in internal), `DataConnectionImpl`
- [ ] Implement [policy getters](https://github.com/googleapis/google-cloud-cpp/blob/8a2298482d4ecbdcbc3406afa65c3d4f44b1d749/google/cloud/bigquery/internal/bigquery_read_connection_impl.h#L69-L97) for `DataConnectionImpl`
- [ ] Add private `Table` constructor that accepts a `DataConnection`
- [ ] Introduce value parameters to Table integration tests. The values will loop over: {`Table` constructed with `DataClient`, `Table` constructed with `DataConnection`}
- [ ] Add MockBigtableStub

PR #8867, opened 2022-05-03
impl(bigtable): add data options

dbolduc:
Part of the work for #8798

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8867)

### DIFF
--- a/google/cloud/bigtable/internal/defaults.cc
+++ b/google/cloud/bigtable/internal/defaults.cc
@@ -32,6 +32,8 @@ namespace internal {
 
 namespace {
 
+auto constexpr kBackoffScaling = 2.0;
+
 // As learned from experiments, idle gRPC connections enter IDLE state after 4m.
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/internal/defaults.cc
```diff
@@ -32,6 +32,8 @@ namespace internal {
 
 namespace {
 
+auto constexpr kBackoffScaling = 2.0;
```
coryan, line 35: This is the only default that gets a constant? 🤷

coryan (approved):

google/cloud/bigtable/internal/defaults.cc
dbolduc: Yes. https://github.com/googleapis/google-cloud-cpp/blob/6d936867bdfc5cd3071e5cb35d9a0291d22a106a/generator/internal/option_defaults_generator.cc#L84-L128

PR #8872, opened 2022-05-04
feat(spanner): spanner::Value support for spanner::PgNumeric

devbww:
Add support for `spanner::PgNumeric` to build a `spanner::Value`
with `Type { .code=NUMERIC, .type_annotation=PG_NUMERIC }`.

`TypeCode::NUMERIC` now takes the `type_annotation` into account
for equality tests.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8872)

### DIFF
--- a/google/cloud/spanner/value.h
+++ b/google/cloud/spanner/value.h
@@ -66,6 +66,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
  * BYTES        | `google::cloud::spanner::Bytes`
  * JSON         | `google::cloud::spanner::Json`
  * NUMERIC      | `google::cloud::spanner::Numeric`
+ * NUMERIC(PG)  | `google::cloud::spanner::PgNumeric`
  * TIMESTAMP    | `google::cloud::spanner::Timestamp`
  * DATE         | `absl::CivilDay`
  * ARRAY        | `std::vector<T>`  // [1]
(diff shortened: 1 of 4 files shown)


google/cloud/spanner/value.h
```diff
@@ -363,7 +365,27 @@ class Value {
   static bool TypeProtoIs(std::string const&, google::spanner::v1::Type const&);
   static bool TypeProtoIs(Bytes const&, google::spanner::v1::Type const&);
   static bool TypeProtoIs(Json const&, google::spanner::v1::Type const&);
-  static bool TypeProtoIs(Numeric const&, google::spanner::v1::Type const&);
+  template <DecimalMode Mode>
+  static bool TypeProtoIs(Decimal<Mode> const&,
+                          google::spanner::v1::Type const& type) {
```
devbww, line 370: @devjgm ... You seemed intent on making these templates, with conditional behavior (albeit at compile time), so I've done that.  For my money they are clearer and cleaner as non-template overloads for the two candidates, `Numeric` and `PgNumeric`.  But I'll let you decide.

devjgm: I don't recall exactly the discussion about this function. I do recall having a preference for `Decimal<M>` being a template rather than two separate, non-templated classes, which is indeed what you did, so I'm happy.

At this point, I have no preference about whether this function is a template or overloaded. Either is fine with me. Happy if you change it or keep it, I'll approve either way.

devjgm (approved):

google/cloud/spanner/value.h
devbww: OK, thanks.  Switched to overloads.

PR #8873, opened 2022-05-04
cleanup: remove remnant IAM deprecation macros

devbww:
`GOOGLE_CLOUD_CPP_BIGTABLE_IAM_DEPRECATED` disappeared in #8652.
`GOOGLE_CLOUD_CPP_IAM_DEPRECATED` disappeared in #8667.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8873)


cmake/GoogleCloudCppCommon.cmake
```diff
@@ -82,8 +82,6 @@ elseif (GOOGLE_CLOUD_CPP_GENERATE_DOXYGEN)
     # to be noops or have simple values.
     set(DOXYGEN_PREDEFINED
         "GOOGLE_CLOUD_CPP_DEPRECATED(x)="
-        "GOOGLE_CLOUD_CPP_IAM_DEPRECATED(x)="
-        "GOOGLE_CLOUD_CPP_BIGTABLE_IAM_DEPRECATED(x)="
         "GOOGLE_CLOUD_CPP_SPANNER_ADMIN_API_DEPRECATED(x)="
```
dbolduc, line 85: Hmm.. I recently added https://github.com/googleapis/google-cloud-cpp/blob/b09c2f33bac5331eb20b2412fd9d4c691bce98f1/google/cloud/bigtable/version.h#L22-L26 and the doxygen seems fine (although I don't remember what the original problem was). Maybe it is fine because it reduces to `GOOGLE_CLOUD_CPP_DEPRECATED`? If so, then maybe we can remove the spanner macro from this list.

dbolduc (approved):
Oops, I missed this in #8667. Thanks for noticing and cleaning up!

cmake/GoogleCloudCppCommon.cmake
devbww: My understanding is that it should be a problem if doxygen saw any code that used `GOOGLE_CLOUD_CPP_BIGTABLE_DATA_CLIENT_DEPRECATED`.  So you may want to investigate.

[The original problem is that doxygen doesn't usually expand macros during its parse, so those attribute macros look like syntax errors.]

dbolduc: Ack. I think the `GOOGLE_CLOUD_CPP_BIGTABLE_DATA_CLIENT_DEPRECATED` macro is only used in tests and internal classes. So we should be fine.

PR #8875, opened 2022-05-04
chore(deps): update storage testbench

coryan:
This will help with some cleanups and additional testing.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8875)


ci/cloudbuild/builds/lib/integration.sh
```diff
@@ -30,7 +30,7 @@ source module ci/cloudbuild/builds/lib/git.sh
 export PATH="${HOME}/.local/bin:${PATH}"
 python3 -m pip uninstall -y --quiet googleapis-storage-testbench Werkzeug
 python3 -m pip install --upgrade --user --quiet --disable-pip-version-check \
-  Werkzeug==2.0.3 "git+https://github.com/googleapis/storage-testbench@v0.16.0"
```
devbww, line 33: Q. Does that mean we can stop uninstalling Werkzeug in the previous command?

coryan: I think that is true in a followup PR, or the next time we have to change this.

PR #8882, opened 2022-05-05
test(spanner): complete spanner_postgresql_numeric_data_type sample

devbww:


google/cloud/spanner/samples/postgresql_samples.cc
```diff
@@ -557,28 +559,106 @@ void InformationSchema(
...
- *   - Insert a Venue with a NaN value for the Revenues column.
- *   - Mutations can also be used to insert/update NUMERIC values,
- *     including NaN values.
- */
 // [START spanner_postgresql_numeric_data_type]
 void NumericDataType(google::cloud::spanner::Client client) {
-  // Get all Venues and inspect the Revenues values.
+  // Insert a Venue with a valid value for the Revenue column.
+  google::cloud::spanner::DmlResult dml_result;
+  auto commit = client.Commit(
```
coryan, line 566: I think this example would benefit from `namespace spanner = ::google::cloud::spanner`.

google/cloud/spanner/samples/postgresql_samples.cc
```diff
@@ -557,28 +559,106 @@ void InformationSchema(
...
-    if (!std::get<1>(*row).has_value()) continue;
-    std::cout << "Revenues of " << std::get<0>(*row) << ": "
-              << *std::get<1>(*row) << "\n";
+    std::cout << "Revenue of " << std::get<0>(*row) << ": ";
+    if (std::get<1>(*row).has_value()) {
+      std::cout << *std::get<1>(*row);
+    } else {
+      std::cout << "NULL";
+    }
+    std::cout << "\n";
```
coryan, line 661: Seems to me that 100 lines is a tad long for an example, but I assume that is the requirement. 🤷‍♂️

coryan (approved):

google/cloud/spanner/samples/postgresql_samples.cc
devbww: I originally thought that it might benefit from not having them, but we do use such an alias in the non-postgres samples, so I'll revisit.  Thanks.

google/cloud/spanner/samples/postgresql_samples.cc
devbww: Yes, I agree.  But yes, I don't make the rules.

PR #8896, opened 2022-05-06
fix(storage): handle full but not finalized uploads

### LINKED ISSUE #8868: Segfault or Uncommitted GCS Object in ResumableUpload
Yes

GCS

**Describe the bug**
There appears to be an edge case where calls to ObjectWriteStream::Close result in a stream that is still in a good and open state.  A subsequent call to Close will result in a segfault.  I've been working on debugging a core dump and I think I've figured out how this happens.

**To Reproduce**
1. Create a resumable upload ObjectWriteStream write_stream
2. Write `N` bytes to the internal buffer
3. Call Close()
4. Call write_stream.IsOpen() returns true
5. Call write_stream.good() returns true
6. Call Close()
7. segfault

Here's what I think is happening during step 3 that gets the stream into this state.  Sorry if some of this ends up a bit stale due to the ongoing refactor but AFAICT it's still an issue in the live version of the code.

1. [FlushFinal()](https://github.com/googleapis/google-cloud-cpp/blob/143a5d7ce4c553d1e1bbb48b9ea5141932c323a0/google/cloud/storage/internal/object_write_streambuf.cc#L135) is called.
2. [hash_function_](https://github.com/googleapis/google-cloud-cpp/blob/143a5d7ce4c553d1e1bbb48b9ea5141932c323a0/google/cloud/storage/internal/object_write_streambuf.cc#L144) is moved and can no longer be used.
3. client_->UploadChunk() is called with a request to upload the bytes in the buffer and finalize the upload.
4. This is handled by [RetryClient::UploadChunk](https://github.com/googleapis/google-cloud-cpp/blob/143a5d7ce4c553d1e1bbb48b9ea5141932c323a0/google/cloud/storage/internal/retry_client.cc#L433).
5. Attempt to upload and finalize the object fails with a retirable error.  Behind the scenes GCS succeeded in committing the provided bytes, but failed to finalize the upload session and actually create the GCS object.
6. The status of the upload session is queried and GCS replies that the upload is in progress with all `N` bytes written.
7. [This line](https://github.com/googleapis/google-cloud-cpp/blob/143a5d7ce4c553d1e1bbb48b9ea5141932c323a0/google/cloud/storage/internal/retry_client.cc#L553) determines that the upload was successful and returns the result.  Note that the upload is still in progress.
8. Since response was good [last_status_](https://github.com/googleapis/google-cloud-cpp/blob/143a5d7ce4c553d1e1bbb48b9ea5141932c323a0/google/cloud/storage/internal/object_write_streambuf.cc#L152) is set to a good and in progress state.
9. I'm not sure what happens [here](https://github.com/googleapis/google-cloud-cpp/blob/143a5d7ce4c553d1e1bbb48b9ea5141932c323a0/google/cloud/storage/internal/object_write_streambuf.cc#L155) because it's different than the code I was testing, but it's a bug either way because either `IsOpen()` will still evaluate to `true` even though the stream was closed, or the stream will claim it has closed successfully despite the destination object not being created.  Pretty sure metadata will be nullopt and the stream stays open but I haven't traced that behavior to be sure.

**Expected behavior**
When the final chunk is uploaded the client library should retry until the resumable upload session is closed or return an error and put the stream in a bad state.

**Operating system:**
Na

Na

I was debugging a core with gcloud-cpp/v1.40.0 but I think this is still present in the most recent code.

**Additional context**
Please ping me if you need any additional info on this.

coryan:
In some very rare circumstances a resumable upload may report that all
bytes have been received (and thus there is no more need to send data),
but fail to send back the finalized object metadata. In those cases we
need to retry sending the message to finalize the object.

Fixes #8868 

/FYI: @Jseph

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8896)

### DIFF
--- a/google/cloud/storage/internal/object_write_streambuf_test.cc
+++ b/google/cloud/storage/internal/object_write_streambuf_test.cc
@@ -460,6 +460,10 @@ TEST(ObjectWriteStreambufTest, Regression8868) {
   EXPECT_EQ(close->committed_size.value_or(0), quantum);
   EXPECT_TRUE(close->payload.has_value());
 
+  // Before the fixes for #8868 this second call (which is legal, though maybe
+  // a bit silly) would crash.  Basically the class assumed that the final
+  // UploadChunk() would always return an error or a full payload.  That is now
+  // guaranteed by the RetryClient.
   close = streambuf.Close();
   ASSERT_STATUS_OK(close);
   EXPECT_FALSE(streambuf.IsOpen());
--- a/google/cloud/storage/internal/object_write_streambuf_test.cc
+++ b/google/cloud/storage/internal/object_write_streambuf_test.cc
@@ -414,6 +414,59 @@ TEST(ObjectWriteStreambufTest, CreatedForFinalizedUpload) {
   ASSERT_STATUS_OK(close_result);
 }
 
+/// @test A regression test for #8868.
+///    https://github.com/googleapis/google-cloud-cpp/issues/8868
+TEST(ObjectWriteStreambufTest, Regression8868) {
+  auto mock = absl::make_unique<testing::MockClient>();
+
+  auto const quantum = UploadChunkRequest::kChunkSizeQuantum;
+  auto const payload = std::string(quantum, '0');
+  using ::testing::Return;
+
+  ::testing::InSequence sequence;
+  // Simulate an upload chunk that has some kind of transient error.
+  EXPECT_CALL(*mock, UploadChunk)
+      .WillOnce(Return(Status(StatusCode::kUnavailable, "try-again")));
+  // This should trigger a `QueryResumableUpload()`, simulate the case where
+  // all the data is reported as "committed", but the payload is not reported
+  // back.
+  EXPECT_CALL(*mock, QueryResumableUpload)
+      .WillOnce(Return(QueryResumableUploadResponse{quantum, absl::nullopt}));
+  EXPECT_CALL(*mock, UploadChunk)
+      .WillOnce(
+          Return(QueryResumableUploadResponse{quantum, ObjectMetadata()}));
+
+  using us = std::chrono::microseconds;
+  auto retry = RetryClient::Create(
+      std::move(mock),
+      Options{}
+          .set<Oauth2CredentialsOption>(oauth2::CreateAnonymousCredentials())
+          .set<RetryPolicyOption>(LimitedErrorCountRetryPolicy(3).clone())
+          .set<BackoffPolicyOption>(
+              ExponentialBackoffPolicy(us(1), us(2), 2).clone())
+          .set<IdempotencyPolicyOption>(
+              AlwaysRetryIdempotencyPolicy{}.clone()));
+  ObjectWriteStreambuf streambuf(
+      std::move(retry), ResumableUploadRequest(), "test-only-upload-id",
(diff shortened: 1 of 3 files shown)


google/cloud/storage/internal/object_write_streambuf_test.cc
```diff
@@ -414,6 +414,59 @@ TEST(ObjectWriteStreambufTest, CreatedForFinalizedUpload) {
...
+  ASSERT_STATUS_OK(close);
+  EXPECT_FALSE(streambuf.IsOpen());
+  EXPECT_EQ(close->committed_size.value_or(0), quantum);
+  EXPECT_TRUE(close->payload.has_value());
+
+  close = streambuf.Close();
+  ASSERT_STATUS_OK(close);
+  EXPECT_FALSE(streambuf.IsOpen());
+  EXPECT_EQ(close->committed_size.value_or(0), quantum);
+  EXPECT_TRUE(close->payload.has_value());
```
dbolduc, line 467: why are we closing twice?

google/cloud/storage/internal/retry_client.cc
```diff
@@ -549,8 +549,18 @@ StatusOr<QueryResumableUploadResponse> RetryClient::UploadChunk(
     }
 
     committed_size = *result->committed_size;
+
+    if (committed_size != expected_committed_size || request.last_chunk()) {
+      // If we still have to send data, restart the loop. On the last chunk,
+      // event if the service reports all the data as received, we need to keep
```
dbolduc, line 555: s/event/even/

scotthart (approved):

google/cloud/storage/internal/retry_client_test.cc
```diff
@@ -507,6 +510,68 @@ TEST(RetryClientTest, UploadChunkMissingRangeHeaderInQueryResumableUpload) {
...
+                              StrictIdempotencyPolicy().clone()));
+
+  auto const quantum = UploadChunkRequest::kChunkSizeQuantum;
+  auto const payload = std::string(quantum, '0');
+
+  // Simulate an upload chunk that has some kind of transient error.
+  EXPECT_CALL(*mock, UploadChunk)
+      .Times(AtLeast(2))
+      .WillRepeatedly(Return(Status(StatusCode::kUnavailable, "try-again")));
+  // This should trigger a `QueryResumableUpload()`, simulate the case the
```
dbolduc, line 563: s/case the/case where the/

google/cloud/storage/internal/object_write_streambuf_test.cc
coryan: Added a comment to explain why.

google/cloud/storage/internal/retry_client_test.cc
coryan: Fixed.

google/cloud/storage/internal/retry_client.cc
coryan: Fixed.

google/cloud/storage/internal/object_write_streambuf_test.cc
dbolduc: Thanks

dbolduc (approved):

PR #8911, opened 2022-05-06
doc(pubsub): deprecate `PublisherOptions`

coryan:
Recommend using `google::cloud::Options` instead. As usual, I documented
the class as deprecated, but did not add a `[[deprecated]]` attribute
for the class, just for each member function.

Part of the work for #7348

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8911)


google/cloud/pubsub/publisher.h
```diff
@@ -189,6 +185,19 @@ class Publisher {
     connection_->ResumePublish({std::move(ordering_key)});
   }
 
+  /// @deprecated Use `Publisher(connection)` instead.
+  GOOGLE_CLOUD_CPP_DEPRECATED("use `Publisher(connection)` instead")
+  explicit Publisher(std::shared_ptr<PublisherConnection> connection,
+                     std::initializer_list<internal::NonConstructible>)
+      : Publisher(std::move(connection)) {}
```
dbolduc, line 192: optional: At the moment, this is unnecessary, as `Publisher(conn, {})` will hit the `PublisherOptions` overload. It only comes into play once we add `Publisher(std::shared_ptr<PublisherConnection> connection, Options options)`.

We can keep it though.

google/cloud/pubsub/publisher_options.cc
```diff
@@ -20,6 +20,10 @@ namespace cloud {
 namespace pubsub {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+PublisherOptions::PublisherOptions() {
```
dbolduc, line 23: nit: we could save these lines by only offering a `Publisher(Options opts = {})`. This is fine though.

google/cloud/pubsub/publisher_options.h
```diff
@@ -86,7 +93,13 @@ class PublisherOptions {
    * to [quota limits].
    *
    * [quota limits]: https://cloud.google.com/pubsub/quotas#resource_limits
+   *
```
dbolduc, line 96: nit: delete this line?

dbolduc (approved):

google/cloud/pubsub/publisher.h
coryan: Removed.

google/cloud/pubsub/publisher_options.h
coryan: Done.

google/cloud/pubsub/publisher_options.cc
coryan: Will keep as-is, mostly laziness.

PR #8933, opened 2022-05-10
refactor(spanner): use struct for transaction callback parameters

devbww:
Introduce `TransactionContext` to encapsulate the read-only callback
parameters to the transaction visitation functor.

When we added the first parameter, `seqno`, we said, "Great, a single
extra parameter."  When we added the second parameter, `tag`, we
said, "This might get out of hand, but there's no time to fix it now."

A third parameter is on the horizon, so let's take the opportunity to
add a "context" struct now to house all the additional transaction
data.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8933)

### DIFF
--- a/google/cloud/spanner/internal/transaction_impl.h
+++ b/google/cloud/spanner/internal/transaction_impl.h
@@ -32,11 +32,16 @@ namespace cloud {
 namespace spanner_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+struct TransactionContext {
+  std::string const& tag;
+  std::int64_t seqno;
+};
+
 template <typename Functor>
 using VisitInvokeResult = ::google::cloud::internal::invoke_result_t<
     Functor, SessionHolder&,
-    StatusOr<google::spanner::v1::TransactionSelector>&, std::string const&,
-    std::int64_t>;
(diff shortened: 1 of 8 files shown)


google/cloud/spanner/internal/transaction_impl.h
```diff
@@ -32,11 +32,16 @@ namespace cloud {
 namespace spanner_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+struct TransactionContext {
+  std::string const& tag;
```
coryan, line 36: This gives me pause.  Should this be an `absl::string_view`?  Or maybe `std::string` and replace the `tag_` and `seqno_` member variables in `TransactionImpl` with a `TransactionContext context_`?

devbww: Using a single, member `TransactionContext` wouldn't work.  The sequence number is per-visit, and multiple visitations can be happening concurrently.

We could use `string_view` for the tag, but unless callbacks need to dissect it (which I'll claim they don't, as it is structure-less ... they simply add it to the request), it doesn't really buy us anything over a `const&`.

coryan (approved):

Issue #8937, opened 2022-05-11
Support asynchronous streaming read RPCs in the generator


coryan:
This is motivated by #568.  We need to implement:

- [x] A `google::cloud::internal::AsyncStreamingReadRpc<Response>` wrapper.
- [x] A `google::cloud::internal::AsyncStreamingReadRpcAuth<Response>` wrapper.
- [x] A `google::cloud::internal::AsyncStreamingReadRpcLogging<Response>` wrapper.
- [x] Change the generator to support asynchronous streaming read functions in `stub_generator.cc`
- [x] Change the generator to support asynchronous streaming read functions in `auth_decorator_generator.cc`
- [x] Change the generator to support asynchronous streaming read functions in `logging_decorator_generator.cc`
- [x] Change the generator to support asynchronous streaming read functions in `metadata_decorator_generator.cc`

PR #8942, opened 2022-05-11
feat(common): async streaming read RPC `*Auth` decorator

coryan:
Part of the work for #8937 

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8942)

### DIFF
--- a/google/cloud/google_cloud_cpp_rest_internal.bzl
+++ b/google/cloud/google_cloud_cpp_rest_internal.bzl
@@ -17,7 +17,6 @@
 """Automatically generated source lists for google_cloud_cpp_rest_internal - DO NOT EDIT."""
 
 google_cloud_cpp_rest_internal_hdrs = [
-    "internal/async_streaming_read_rpc_auth.h",
     "internal/binary_data_as_debug_string.h",
     "internal/curl_handle.h",
     "internal/curl_handle_factory.h",
--- a/google/cloud/google_cloud_cpp_rest_internal.bzl
+++ b/google/cloud/google_cloud_cpp_rest_internal.bzl
@@ -17,6 +17,7 @@
 """Automatically generated source lists for google_cloud_cpp_rest_internal - DO NOT EDIT."""
 
 google_cloud_cpp_rest_internal_hdrs = [
+    "internal/async_streaming_read_rpc_auth.h",
     "internal/binary_data_as_debug_string.h",
     "internal/curl_handle.h",
     "internal/curl_handle_factory.h",
--- a/google/cloud/google_cloud_cpp_rest_internal.cmake
+++ b/google/cloud/google_cloud_cpp_rest_internal.cmake
@@ -21,7 +21,6 @@ find_package(OpenSSL REQUIRED)
 # the library
 add_library(
     google_cloud_cpp_rest_internal # cmake-format: sort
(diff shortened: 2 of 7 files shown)


google/cloud/google_cloud_cpp_rest_internal.cmake
```diff
@@ -21,6 +21,7 @@ find_package(OpenSSL REQUIRED)
 # the library
 add_library(
     google_cloud_cpp_rest_internal # cmake-format: sort
+    internal/async_streaming_read_rpc_auth.h
```
dbolduc, line 24: seems wrong

google/cloud/google_cloud_cpp_rest_internal.bzl
```diff
@@ -17,6 +17,7 @@
 """Automatically generated source lists for google_cloud_cpp_rest_internal - DO NOT EDIT."""
 
 google_cloud_cpp_rest_internal_hdrs = [
+    "internal/async_streaming_read_rpc_auth.h",
```
dbolduc, line 20: seems wrong

google/cloud/internal/async_streaming_read_rpc_auth.h
```diff
@@ -0,0 +1,122 @@
...
+
+    std::unique_ptr<grpc::ClientContext> ReleaseInitialContext() {
+      std::lock_guard<std::mutex> g{mu};
+      return std::move(initial_context);
+    }
+
+    future<bool> OnStart(
+        StatusOr<std::unique_ptr<grpc::ClientContext>> context) {
+      std::lock_guard<std::mutex> g{mu};
+      if (cancelled) return make_ready_future(false);
```
dbolduc, line 82: `Cancel()` is untested. I think we want a test for the case where `Cancel()` is called before the `AsyncConfigureContext()` future is ready.

google/cloud/internal/async_streaming_read_rpc_auth.h
```diff
@@ -0,0 +1,122 @@
...
+      if (context) {
+        stream = factory(*std::move(context));
+      } else {
+        stream = absl::make_unique<AsyncStreamingReadRpcError<Response>>(
+            std::move(context).status());
+      }
```
dbolduc, line 88: I think we also want a test case for when the auth strategy returns a bad status

dbolduc (changes requested):

google/cloud/google_cloud_cpp_rest_internal.cmake
dbolduc: Aside: it is surprising to me that our builds do not catch this...

google/cloud/google_cloud_cpp_rest_internal.bzl
coryan: Fixed

google/cloud/internal/async_streaming_read_rpc_auth.h
coryan: Done.

dbolduc (approved):

google/cloud/google_cloud_cpp_rest_internal.cmake
dbolduc: Thanks for explaining

PR #8944, opened 2022-05-12
chore(pubsublite): coverage ignores generated code

dbolduc:
https://docs.codecov.com/docs/ignoring-paths

<hr />

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8944)


.codecov.yml
```diff
@@ -7,21 +7,19 @@ coverage:
 
 # Ignore generated code that lives with handcrafted code.
 ignore:
+  - "google/cloud/*/internal/*_auth_decorator.cc"
+  - "google/cloud/*/internal/*_auth_decorator.h"
+  - "google/cloud/*/internal/*_logging_decorator.cc"
+  - "google/cloud/*/internal/*_logging_decorator.h"
+  - "google/cloud/*/internal/*_metadata_decorator.cc"
+  - "google/cloud/*/internal/*_metadata_decorator.h"
+  - "google/cloud/*/internal/*_stub.cc"
```
coryan, line 16: https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/pubsub/internal/publisher_stub.cc

dbolduc: That was careless of me. I updated the regex to match things more closely.

Aside: seems like we could generate the stubs and decorators for pubsub and spanner if we wanted to.

coryan (approved):

.codecov.yml
coryan: It is a bit of a minefield, easy enough to miss files.  And yes, if somebody wants to move those stubs to use the micro-generator  they would be welcome to.

PR #8948, opened 2022-05-12
refactor(bigtable): use variant<status, T> not StatusOr<optional<T>>

dbolduc:
refactor `RowReader` / `RowReaderIterator` to deal in `absl::variant<Status, Row>`s instead of `StatusOr<absl::optional<Row>>`s. This is more consistent with the interface `StreamRange` expects. In a subsequent PR I will probably implement RowReader in terms of a StreamRange (which will allow me to delete `bigtable/internal/rowreader_iterator.{h,cc}`)

 | previous type: `StatusOr<absl::optional<Row>>` | new type: `absl::variant<Status, Row>` |
|-|-|
| has `Row` | has `Row`|
| bad Status | has bad Status |
| has `absl::nullopt` | has OK Status|

<hr />

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8948)

### DIFF
--- a/google/cloud/bigtable/internal/legacy_row_reader_impl.cc
+++ b/google/cloud/bigtable/internal/legacy_row_reader_impl.cc
@@ -116,9 +116,7 @@ absl::variant<Status, bigtable::Row> LegacyRowReaderImpl::Advance() {
     }
 
     auto status = absl::get<Status>(std::move(variant));
-    if (status.ok()) {
-      return Status();
-    }
+    if (status.ok()) return Status{};
 
     // In the unlikely case when we have already reached the requested
     // number of rows and still receive an error (the parser can throw
@@ -126,7 +124,7 @@ absl::variant<Status, bigtable::Row> LegacyRowReaderImpl::Advance() {
     // retry and we have no good value for rows_limit anyway.
     if (rows_limit_ != bigtable::RowReader::NO_ROWS_LIMIT &&
         rows_limit_ <= rows_count_) {
-      return Status();
+      return Status{};
     }
 
     if (!last_read_row_key_.empty()) {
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/internal/legacy_row_reader_impl.cc
```diff
@@ -105,25 +105,28 @@ bool LegacyRowReaderImpl::NextChunk() {
...
-    grpc::Status status = AdvanceOrFail(row);
+    auto variant = AdvanceOrFail();
+    if (absl::holds_alternative<bigtable::Row>(variant)) {
+      return absl::get<bigtable::Row>(std::move(variant));
+    }
+
+    auto status = absl::get<Status>(std::move(variant));
     if (status.ok()) {
-      return row;
+      return Status();
```
coryan, line 120: nit: nowadays we would write `if (status.ok()) return Status{};`

coryan: Here and some places below.

google/cloud/bigtable/internal/rowreaderiterator.cc
```diff
@@ -41,14 +41,14 @@ RowReaderIterator& RowReaderIterator::operator++() {
 }
 
 void RowReaderIterator::Advance() {
-  auto status_or_optional_row = owner_->Advance();
-  if (!status_or_optional_row) {
-    row_ = StatusOr<bigtable::Row>(std::move(status_or_optional_row).status());
+  auto variant = owner_->Advance();
+  if (absl::holds_alternative<bigtable::Row>(variant)) {
```
coryan, line 45: You could do this with a visitor (but double check my ref-ref):

```cc
  struct Visitor {
    StatusOr<bigtable::Row> operator()(Status&& s) { return s; }
    StatusOr<bigtable::Row> operator()(bigtable::Row&& r) { return r; }
  };
  row_ = absl::visit(Visitor{}, owner_->Advance());
}
```

coryan (approved):

google/cloud/bigtable/internal/legacy_row_reader_impl.cc
dbolduc: Fixed.

google/cloud/bigtable/internal/rowreaderiterator.cc
dbolduc: Won't Fix. I am going to delete the class shortly.

Also... what happens if your code is given an OK Status? 🤔

PR #8962, opened 2022-05-13
fix(generator): async streaming read MD respects routing header

dbolduc:
My stub factory tests are failing without this

<hr />

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8962)

### DIFF
--- a/generator/integration_tests/golden/internal/golden_kitchen_sink_metadata_decorator.cc
+++ b/generator/integration_tests/golden/internal/golden_kitchen_sink_metadata_decorator.cc
@@ -114,7 +114,7 @@ GoldenKitchenSinkMetadata::AsyncTailLogEntries(
     google::cloud::CompletionQueue const& cq,
     std::unique_ptr<grpc::ClientContext> context,
     google::test::admin::database::v1::TailLogEntriesRequest const& request) {
-  SetMetadata(*context);
+  SetMetadata(*context, {});
   return child_->AsyncTailLogEntries(cq, std::move(context), request);
 }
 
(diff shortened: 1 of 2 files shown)


generator/integration_tests/golden/internal/golden_kitchen_sink_metadata_decorator.cc
```diff
@@ -114,7 +114,7 @@ GoldenKitchenSinkMetadata::AsyncTailLogEntries(
     google::cloud::CompletionQueue const& cq,
     std::unique_ptr<grpc::ClientContext> context,
     google::test::admin::database::v1::TailLogEntriesRequest const& request) {
-  SetMetadata(*context);
+  SetMetadata(*context, {});
```
devbww, line 117: Q. Should we really be setting the request params to `""` -v- not setting them at all?

dbolduc: Hm, so looking at this [internal discussion](https://groups.google.com/a/google.com/g/cloud-spanner-users/c/omhDdnBBb1c/m/LnbkGUBHEwAJ), I think the answer is no. If there's no resource targeted, then there's no gain from setting `x-goog-request-params=""`.

We do this for a few flavors of RPC. I reverted the change for async streaming reads in this PR. I will send a subsequent PR to change the other cases.

dbolduc: opened #8963. I don't think it will conflict with this PR.

devbww (approved):

PR #8968, opened 2022-05-14
fix(storage): avoid crashes when parsing `ErrorInfo`

### LINKED ISSUE #8965: google storage client json throw exception when http request fails
GCS

**Describe the bug**
When http response fails, it would rely on json library to get detailed error message.
Code here: https://github.com/googleapis/google-cloud-cpp/blob/643fa047a549422ebaa40f9173ab0c2c8372aede/google/cloud/storage/internal/http_response.cc#L172

Sample http response-1:
```
{\"error\":\"invalid_grant\",\"error_description\":\"Invalid grant: account not found\"}
```
Exception stacktrace:
```
  what():  [json.exception.type_error.306] cannot use value() with string
```
Using some logging, the exception happens: https://github.com/googleapis/google-cloud-cpp/blob/643fa047a549422ebaa40f9173ab0c2c8372aede/google/cloud/storage/internal/http_response.cc#L189

Sample http response-2:
```
Service Unavailable
```
Exception stacktrace:
```
  what():  [json.exception.type_error.306] cannot use value() with string
```
With some logging the error happens:
https://github.com/googleapis/google-cloud-cpp/blob/643fa047a549422ebaa40f9173ab0c2c8372aede/google/cloud/storage/internal/http_response.cc#L175

**Expected behavior**
No exception would be thrown.

**Operating system:**
ubuntu 20

clang v13.0.0

v1.38.0

**Additional information**
I tried to reproduce locally with the above error response, but cannot repro it.
I want you to give me some intuitions.

coryan:
Maybe unsurprisingly, the error payloads are less structured that we
expected. This PR performs additional validation to avoid crashes while
trying to parse these payloads as a `ErrorInfo`.  If we could reliably
use exceptions I would just parse the code and handle any exception in
a single line of code, alas, we cannot.

Fixes #8965

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8968)

### DIFF
--- a/google/cloud/storage/internal/http_response.cc
+++ b/google/cloud/storage/internal/http_response.cc
@@ -148,17 +148,23 @@ StatusCode MapHttpCodeToStatus(long code) {  // NOLINT(google-runtime-int)
 ErrorInfo MakeErrorInfo(nlohmann::json const& details) {
   static auto constexpr kErrorInfoType =
       "type.googleapis.com/google.rpc.ErrorInfo";
+  if (!details.is_array()) return ErrorInfo{};
   for (auto const& e : details.items()) {
     auto const& v = e.value();
+    if (!v.is_object()) continue;
+    if (!v.contains("@type") || !v["@type"].is_string()) continue;
     if (v.value("@type", "") != kErrorInfoType) continue;
-    auto reason = v.value("reason", "");
-    auto domain = v.value("domain", "");
-    auto metadata_json = v.value("metadata", nlohmann::json::object());
+    if (!v.contains("reason") || !v["reason"].is_string()) continue;
+    if (!v.contains("domain") || !v["domain"].is_string()) continue;
+    if (!v.contains("metadata") || !v["metadata"].is_object()) continue;
+    auto const& metadata_json = v["metadata"];
     auto metadata = std::unordered_map<std::string, std::string>{};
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/http_response.cc
```diff
@@ -186,11 +194,13 @@ Status AsStatus(HttpResponse const& http_response) {
...
-  auto error = json.value("error", nlohmann::json::object());
-  auto message = error.value("message", http_response.payload);
-  auto details = error.value("details", nlohmann::json::object());
-  auto error_info = MakeErrorInfo(details);
-  return Status(status_code, std::move(message), std::move(error_info));
+  if (!json.contains("error")) return default_error();
+  auto const& e = json["error"];
+  if (!e.is_object()) return default_error();
+  if (!e.contains("message") || !e.contains("details")) return default_error();
+  if (!e["message"].is_string()) return default_error();
```
dbolduc, line 201: We could return a `Status(code, e.value("message", ""))` if `e.contains("message") && e["message"].is_string() && !e.contains("details")`, but it's also fine to return the whole payload as a message.

dbolduc (approved):

google/cloud/storage/internal/http_response.cc
coryan: Ack

PR #8969, opened 2022-05-14
cleanup(bigtable): better example code for GC rules

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8969)


coryan (approved):
I find both approaches "meh".  Initializing protos is always ugly.  For a future PR: Maybe we should add support to the `bigtable::TableConfig` classes to create the protos we need for the different RPCs in `bigtable/admin/`.

dbolduc:
> I find both approaches "meh". Initializing protos is always ugly. For a future PR: Maybe we should add support to the `bigtable::TableConfig` classes to create the protos we need for the different RPCs in `bigtable/admin/`.

Ack. The support exists today. the `*.as_proto()` functions are all public. I think the `cbt::GcRule::MaxAge(duration)` is easier to work with than setting the count of seconds directly in the proto.

I am going to merge this. I might end up putting things back to the way they were later.

PR #8976, opened 2022-05-16
refactor(spanner): use common RST_STREAM status helper

### LINKED ISSUE #7394: Treat certain known `kInternal` errors as retryable

dbolduc:
Fixes #7394 (well, this is all I am going to do on it unless it becomes a bigger problem)

Refactor spanner to point to the [common helper](https://github.com/googleapis/google-cloud-cpp/blob/0ff878ea71ccea66e1e2c7cf74e703683ec7ad94/google/cloud/internal/retry_policy.cc#L24-L30).

Note that we drop the "Connection closed with unknown cause" case. It was fixed 4 years ago. Googlers can see: https://b.corp.google.com/issues/27794742#comment14

We also drop the "HTTP/2 error code: INTERNAL_ERROR" case. It is covered by matching on "Received Rst Stream". Googlers can see: https://b.corp.google.com/issues/33017979 for an example

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8976)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl_test.cc
+++ b/google/cloud/spanner/internal/connection_impl_test.cc
@@ -2006,9 +2006,9 @@ TEST(ConnectionImplTest, ExecutePartitionedDmlRetryableInternalErrors) {
       .WillOnce(Return(ByMove(MakeReader(
           {}, {grpc::StatusCode::INTERNAL,
                "Received unexpected EOS on DATA frame from server"}))))
-      .WillOnce(
-          Return(ByMove(MakeReader({}, {grpc::StatusCode::INTERNAL,
-                                        "HTTP/2 error code: INTERNAL_ERROR"}))))
+      .WillOnce(Return(ByMove(MakeReader(
+          {}, {grpc::StatusCode::INTERNAL,
+               "HTTP/2 error code: INTERNAL_ERROR\nReceived Rst Stream"}))))
       .WillOnce(Return(ByMove(MakeReader({kTextResponse}))));
 
   auto result = conn->ExecutePartitionedDml(
--- a/google/cloud/spanner/retry_policy.h
+++ b/google/cloud/spanner/retry_policy.h
@@ -32,23 +32,9 @@ struct SafeGrpcRetry {
     return status.ok();
   }
   static inline bool IsTransientFailure(google::cloud::Status const& status) {
-    if (status.code() == StatusCode::kUnavailable ||
-        status.code() == StatusCode::kResourceExhausted) {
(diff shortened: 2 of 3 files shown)


coryan (approved):

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -2006,9 +2006,9 @@ TEST(ConnectionImplTest, ExecutePartitionedDmlRetryableInternalErrors) {
       .WillOnce(Return(ByMove(MakeReader(
           {}, {grpc::StatusCode::INTERNAL,
                "Received unexpected EOS on DATA frame from server"}))))
-      .WillOnce(
-          Return(ByMove(MakeReader({}, {grpc::StatusCode::INTERNAL,
-                                        "HTTP/2 error code: INTERNAL_ERROR"}))))
+      .WillOnce(Return(ByMove(MakeReader(
+          {}, {grpc::StatusCode::INTERNAL,
+               "HTTP/2 error code: INTERNAL_ERROR\nReceived Rst Stream"}))))
```
devbww, line 2011: Opt: Perhaps a newline in an error message isn't the best choice?

dbolduc: @devbww and I talked offline. While it is weird to have a newline in the error message, we are not the ones creating this error message. We are just handling it. So it's fine to include a newline in our tests.

PR #8977, opened 2022-05-16
refactor(rest): remove RestEndpointOption from Rest library

scotthart:
common_options.h already defines an `google::cloud::EndpointOption`, having an additional `google::cloud::rest_internal::RestEndpointOption` is unnecessary as the endpoint used to create the `RestClient` should be resolved by the library using the `RestClient`.

Additionally, add comments emphasizing the immutability of an endpoint in a `CurlRestClient` and add comments emphasizing the relationship between `EndpoinOption` and `AuthorityOption`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8977)

### DIFF
--- a/google/cloud/common_options.h
+++ b/google/cloud/common_options.h
@@ -30,7 +30,9 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
  *
  * In almost all cases a suitable default will be chosen automatically.
  * Applications may need to be changed to (1) test against a fake or simulator,
- * or (2) use a beta or EAP version of the service.
+ * or (2) use a beta or EAP version of the service. When using a beta or EAP
+ * version of the service, the AuthorityOption should also be set to the usual
+ * hostname of the service.
  */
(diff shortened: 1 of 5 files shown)


coryan (commented):
In the storage library you may be using gRPC and REST both at the same time, but the endpoints need to be different (`https://storage.googleapis.com` for REST and `storage.googleapis.com` for gRPC, maybe with a weird `google-c2p:///` prefix).

How would that work?

coryan (changes requested):
Actually I think my question is a blocker.

coryan (approved):
> google/cloud/storage/options.h defines a RestEndpointOption that would be used.

I see, I take it that `google::cloud::storage::RestEndpointOption` would need to survive any migration to use `RestClient` as the implementation for the storage library?  I guess there would need to be some special code to handle the initialization of `storage::internal::HybridClient`.  Some of the details escape me, but I think we can figure them out.

PR #8988, opened 2022-05-17
refactor(storage): configure AuthorityOption when using RestClient

scotthart:


google/cloud/storage/internal/rest_client.cc
```diff
@@ -104,11 +104,45 @@ Status AddAuthorizationHeader(Options const& options,
...
+  std::unique_ptr<rest::RestClient> iam_rest_client;
+  auto curl_options = options;
+  auto storage_endpoint = RestEndpoint(options);
+  auto iam_endpoint = IamRestEndpoint(options);
+
+  if (!options.has<AuthorityOption>() &&
+    absl::StrContains(storage_endpoint, "googleapis.com")) {
+    storage_rest_client = rest::MakePooledRestClient(
+        storage_endpoint,
+        options.set<AuthorityOption>("storage.googleapis.com"));
```
coryan, line 118: Consider the case where `options == Options{}.set<IamEndpointOption>("foo").set<RestEndpointOption>("bar.googleapis.com")`

Then line 118 will set the `AuthorityOption` to `storage.googleapis.com` and that will get used in line 131.  Is that what we want?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8988)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @coryan)

*[a discussion](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8988#-N2IWX6SCIZG-henibMC:-N2IWX6TAfxnkerN3Xm8:b-9pmqlc) (no related file):*
Modified the RestClient constructor to also accept an injected RawClient in case we want to inject a mock.

___
___
*[`google/cloud/storage/internal/rest_client.cc` line 118 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8988#-N2Ek_jktEZREyLQ6r59-r1-118:-N2IWN3E8qEBGXrihPxC:b-2t3vvt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/88362ca6b087d5aa3c6690cb0e383dd19736815a/google/cloud/storage/internal/rest_client.cc#L118)):*

Consider the case where `options == Options{}.set<IamEndpointOption>("foo").set<RestEndpointOption>("bar.googleapis.com")`

Then line 118 will set the `AuthorityOption` to `storage.googleapis.com` and that will get used in line 131.  Is that what we want?
</blockquote></details>

It's not. I'll modify things to make sure we preserve the original options accordingly.

coryan (approved):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8988)* status: all files reviewed, 2 unresolved discussions (waiting on @scotthart)

*[a discussion](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8988#-N2IWX6SCIZG-henibMC:-N2Ll8unE_JF6KA3H6ls:b-z09zjt) (no related file):*

Modified the RestClient constructor to also accept an injected RawClient in case we want to inject a mock.
</blockquote></details>

Ack.  I think you mean you can inject two `RestClient`'s, but that is just nit picking.

The current code works, and it is perfectly serviceable.  I am Okay if you merge this as-is. It could be cleaned up a bit (in my opinion).  For a future PR, consider accepting *constructed* `std::shared_ptr<RestClient>` as inputs, and move the code to decorate the `Options` to standalone functions (I will try to explain better below).

___
___
*[`google/cloud/storage/internal/rest_client.h` line 44 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8988#-N2LldNq9OoXWcwH-_Sd:-N2LldNr1wLDKsH0f646:b6j8xy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/52dd6a75feb45a68b73db88443db797663800ba1/google/cloud/storage/internal/rest_client.h#L44)):*
> ```objc
>       Options options,
>       std::function<std::unique_ptr<google::cloud::rest_internal::RestClient>(
>           std::string, Options)> const& rest_client_factory_function =
> ```

Also for a future PR, you could just change this to be:

```cc
static std::shared_ptr<RestClient> Create(
    Options options, std::shared_ptr<rest_internal::RestClient> storage_rest_client, std::shared_ptr<rest_internal::RestClient> iam_rest_client);
``` 

With a couple of helper functions:

```cc
static Options StorageRestClientOptions(Options const& options) {
  auto endpoint = RestEndpoint(options);
  if (options.has<AuthorityOption> || !absl::StrContains(endpoint, "googleapis.com")) return options;
  return Options(options).set<AuthorityOption>("storage.googleapis.com");
}

static Options IamRestClientOptions(Options const& options) {
  auto endpoint = IamEndpoint(options);
  if (options.has<AuthorityOption> || !absl::StrContains(endpoint, "googleapis.com")) return options;
  return Options(options).set<AuthorityOption>("iamcredentials.googleapis.com");
}
```

Then this would be used via:

```cc
static std::shared_ptr<RestClient> Create(Options options) {
  auto storage = rest_internal::MakePooledRestClient(StorageRestClientOptions(options));
  auto iam = rest_internal::MakePooledRestClient(IamRestClientOptions(options));
  return Create(std::move(options), std::move(storage), std::move(iam));
}
```

I think that would make testing simpler, and the code easier to follow.  You can throw some $2 words about decoupling if somebody is into that kind of stuff.

coryan (approved):
:lgtm:

> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8988#-:-N2NlD7V8EJIRoazPmE5:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @scotthart)

___
*[`google/cloud/storage/internal/rest_client.cc` line 134 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8988#-N2Nl70r1gPpXguMR-km:-N2Nl70r1gPpXguMR-kn:b1szcb5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c114e03e434129acf02cce3113e1deb8f47a9bd8/google/cloud/storage/internal/rest_client.cc#L134)):*
> ```cpp
>   return Options(options).set<AuthorityOption>("storage.googleapis.com");
> }
> Options RestClient::ResolveIamAuthority(Options const& options) {
> ```

nit: blank line between functions.

PR #9001, opened 2022-05-17
fix(common): override the AuthorityOption along with EndpointOption

devbww:
I needed this to get things to work when setting the `endpoint_env_var`
to use the Spanner staging service.  Please tell me why I'm wrong.

### DIFF
--- a/google/cloud/internal/populate_common_options.cc
+++ b/google/cloud/internal/populate_common_options.cc
@@ -35,6 +35,7 @@ Options PopulateCommonOptions(Options opts, std::string const& endpoint_env_var,
   if (!opts.has<EndpointOption>()) {
     auto e = internal::GetEnv(endpoint_env_var.c_str());
     if (e && !e->empty()) {
+      opts.set<AuthorityOption>(*e);
       opts.set<EndpointOption>(*std::move(e));
     } else {
       opts.set<EndpointOption>(default_endpoint);
--- a/google/cloud/internal/populate_common_options_test.cc
+++ b/google/cloud/internal/populate_common_options_test.cc
@@ -54,26 +54,28 @@ TEST(PopulateCommonOptions, Endpoint) {
     Options initial;
     absl::optional<std::string> default_env;
     absl::optional<std::string> emulator_env;
-    std::string expected;
+    std::string expected_endpoint;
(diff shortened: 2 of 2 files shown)


google/cloud/internal/populate_common_options.cc
```diff
@@ -35,6 +35,7 @@ Options PopulateCommonOptions(Options opts, std::string const& endpoint_env_var,
   if (!opts.has<EndpointOption>()) {
     auto e = internal::GetEnv(endpoint_env_var.c_str());
     if (e && !e->empty()) {
+      opts.set<AuthorityOption>(*e);
```
dbolduc, line 38: (This isn't the sort of feedback you were looking for, but) do you want to override an already set `AuthorityOption`? We do not override an already set `EndpointOption`.

coryan:
> I needed this to get things to work when setting the `endpoint_env_var` to use the Spanner staging service.

Ack.

> Please tell me why I'm wrong.

On [VPC-SC](https://cloud.google.com/vpc-service-controls/docs) environments (which I expect are more commonly used than the staging service) the endpoint is something like `restricted.googleapis.com` while the authority **must** be `spanner.googleapis.com` (or whatever service you are trying to use).  The [docs](https://cloud.google.com/vpc-service-controls/docs/set-up-private-connectivity) are not very explicit, but that is my recollection.

Maybe consider an additional environment variable `GOOGLE_CLOUD_CPP_USE_ENDPOINT_FOR_AUTHORITY=<boolean>`

google/cloud/internal/populate_common_options.cc
devbww: > do you want to override an already set `AuthorityOption`?

I think so, but I'm open to be corrected.

> We do not override an already set `EndpointOption`.

Yeah, and that sounds like a mistake to me.  The environment variable should always win over anything set in the code.

devbww:
> Maybe consider an additional environment variable `GOOGLE_CLOUD_CPP_USE_ENDPOINT_FOR_AUTHORITY=<boolean>`

Or we could decouple them completely with some kind of `authority_env_var` that simply contains the desired value.

In any case, I'll close this, and open an item for discussion instead.

PR #9004, opened 2022-05-17
chore: renovate updates scaffolding

dbolduc:
Tested here: https://github.com/dbolduc/test-renovate/pull/10/files

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9004)

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -8,7 +8,8 @@
     {
       "fileMatch": [
         ".*\\.Dockerfile$",
-        ".*doc/packaging\\.md$"
+        ".*doc/packaging\\.md$",
+        ".*generator/internal/scaffold_generator\\.cc$"
       ],
       "matchStrings": [
         "https://github\\.com/(?<depName>.*?)/archive/(?<currentValue>.*?)\\.tar\\.gz",


scotthart (dismissed):

.github/renovate.json
```diff
@@ -8,7 +8,8 @@
     {
       "fileMatch": [
         ".*\\.Dockerfile$",
-        ".*doc/packaging\\.md$"
+        ".*doc/packaging\\.md$",
+        ".*generator/internal/scaffold_generator\\.cc$"
```
coryan, line 12: This is not quite right.  We use a different manager to update the `.bzl` files:

https://docs.renovatebot.com/modules/manager/bazel/

It needs to do fancy stuff with the SHA256 too.

coryan (changes requested):
I am not sure if it can be made to work, but this does not look quite right to me.

.github/renovate.json
dbolduc: renovate-bot doesn't believe me when I say `scaffold_generator.cc` is a WORKSPACE file. I give up. Closing.

PR #9021, opened 2022-05-19
feat(common): add functions for url encoding and decoding

scotthart:
This functionality exists in libcurl, but requires creating a curl handle, which we're trying to better encapsulate. I've put off rolling my own prior to this, but in order to url encode object names for storage it's necessary. This implementation matches that of curl_easy_escape which escapes all printable ASCII characters except for [a-zA-Z0-9-_.~].

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9021)

### DIFF
--- a/google/cloud/url_encoded.cc
+++ b/google/cloud/url_encoded.cc
@@ -33,7 +33,6 @@ std::array<const char[4], 95> constexpr kEncode = {
 }  // namespace
 
 std::string UrlEncode(std::string const& input) {
-  if (input.empty()) return {};
   std::stringstream encoded;
   for (auto const& c : input) {
     encoded << kEncode[c - 0x20];
@@ -42,7 +41,6 @@ std::string UrlEncode(std::string const& input) {
 }
 
 std::string UrlDecode(std::string const& input) {
-  if (input.empty()) return {};
   std::stringstream decoded;
   for (std::size_t i = 0; i < input.size(); ++i) {
     if (input[i] == '%') {
--- a/google/cloud/url_encoded.cc
+++ b/google/cloud/url_encoded.cc
@@ -1,0 +1,61 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/url_encoded.h"
+#include <array>
+#include <sstream>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace {
+std::array<const char[4], 95> constexpr kEncode = {
+    "%20", "%21", "%22", "%23", "%24", "%25", "%26", "%27", "%28", "%29", "%2A",
+    "%2B", "%2C", "-",   ".",   "%2F", "0",   "1",   "2",   "3",   "4",   "5",
+    "6",   "7",   "8",   "9",   "%3A", "%3B", "%3C", "%3D", "%3E", "%3F", "%40",
+    "A",   "B",   "C",   "D",   "E",   "F",   "G",   "H",   "I",   "J",   "K",
+    "L",   "M",   "N",   "O",   "P",   "Q",   "R",   "S",   "T",   "U",   "V",
+    "W",   "X",   "Y",   "Z",   "%5B", "%5C", "%5D", "%5E", "_",   "%60", "a",
+    "b",   "c",   "d",   "e",   "f",   "g",   "h",   "i",   "j",   "k",   "l",
+    "m",   "n",   "o",   "p",   "q",   "r",   "s",   "t",   "u",   "v",   "w",
+    "x",   "y",   "z",   "%7B", "%7C", "%7D", "~"};
+}  // namespace
+
+std::string UrlEncode(std::string const& input) {
+  if (input.empty()) return {};
+  std::stringstream encoded;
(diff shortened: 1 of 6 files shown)


google/cloud/url_encoded_test.cc
```diff
@@ -0,0 +1,76 @@
...
+using ::testing::IsEmpty;
+
+TEST(UrlEncodeString, AllPrintableAsciiCharacters) {
+  std::string escaped_printable_ascii_chars =
+      R"(%20%21%22%23%24%25%26%27%28%29%2A%2B%2C-.%2F0123456789%3A%3B%3C%3D%3E%3F%40ABCDEFGHIJKLMNOPQRSTUVWXYZ%5B%5C%5D%5E_%60abcdefghijklmnopqrstuvwxyz%7B%7C%7D~)";
+  std::string printable_ascii_chars =
+      R"( !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~)";
+  auto result = UrlEncode(printable_ascii_chars);
+  EXPECT_EQ(result, escaped_printable_ascii_chars);
+}
```
devbww, line 32: Perhaps combining each encode/decode test pair into a single "round trip" test would be a feature?

google/cloud/url_encoded.cc
```diff
@@ -0,0 +1,61 @@
...
+    "A",   "B",   "C",   "D",   "E",   "F",   "G",   "H",   "I",   "J",   "K",
+    "L",   "M",   "N",   "O",   "P",   "Q",   "R",   "S",   "T",   "U",   "V",
+    "W",   "X",   "Y",   "Z",   "%5B", "%5C", "%5D", "%5E", "_",   "%60", "a",
+    "b",   "c",   "d",   "e",   "f",   "g",   "h",   "i",   "j",   "k",   "l",
+    "m",   "n",   "o",   "p",   "q",   "r",   "s",   "t",   "u",   "v",   "w",
+    "x",   "y",   "z",   "%7B", "%7C", "%7D", "~"};
+}  // namespace
+
+std::string UrlEncode(std::string const& input) {
+  if (input.empty()) return {};
```
devbww, line 36: Why special-case empty input?

google/cloud/url_encoded.cc
```diff
@@ -0,0 +1,61 @@
...
+  std::stringstream encoded;
+  for (auto const& c : input) {
+    encoded << kEncode[c - 0x20];
+  }
+  return encoded.str();
+}
+
+std::string UrlDecode(std::string const& input) {
+  if (input.empty()) return {};
```
devbww, line 45: Ditto on empty input.

google/cloud/url_encoded.h
```diff
@@ -0,0 +1,38 @@
...
+#include "google/cloud/version.h"
+#include <string>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// Encodes the given string and returns a new string such that all input
+// characters that are not in [a-zA-z-._~] are converted to their "URL escaped"
+// version "%NN" where NN is the ASCII hex value.
```
devbww, line 27: Do we need to say anything about what happens should an input character be outside [32 ... 126]?

google/cloud/url_encoded.h
```diff
@@ -0,0 +1,38 @@
...
+std::string UrlEncode(std::string const& input);
+
+// Decodes the given string, interpreting any "%NN" sequence as an "URL escaped"
+// character and substitutes the ASCII character corresponding to hex value NN.
```
devbww, line 31: Do we need to say anything about what happens should the input be malformed (e.g., a `%` not followed by two hex digits)?

google/cloud/url_encoded.cc
```diff
@@ -0,0 +1,61 @@
...
+    "L",   "M",   "N",   "O",   "P",   "Q",   "R",   "S",   "T",   "U",   "V",
+    "W",   "X",   "Y",   "Z",   "%5B", "%5C", "%5D", "%5E", "_",   "%60", "a",
+    "b",   "c",   "d",   "e",   "f",   "g",   "h",   "i",   "j",   "k",   "l",
+    "m",   "n",   "o",   "p",   "q",   "r",   "s",   "t",   "u",   "v",   "w",
+    "x",   "y",   "z",   "%7B", "%7C", "%7D", "~"};
+}  // namespace
+
+std::string UrlEncode(std::string const& input) {
+  if (input.empty()) return {};
+  std::stringstream encoded;
```
coryan, line 37: I would think `std::string` and `operator+=` would be enough for this?

google/cloud/url_encoded.cc
```diff
@@ -0,0 +1,61 @@
...
+  for (auto const& c : input) {
+    encoded << kEncode[c - 0x20];
```
coryan, line 39: maybe `static_assert(' ' == 0x20, "This library is only designed to work with ASCII-based character sets")`?  The code could land in a EBCDIC system someday :fear:

google/cloud/url_encoded.h
coryan: I think the answer is yes, for example, GCS object names are UTF-8, and they can appear in the URL.

coryan (changes requested):

google/cloud/url_encoded.cc
devbww: You might consider something like ...

https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/internal/base64_transforms.cc#L27-L37

google/cloud/url_encoded.h
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_URL_ENCODED_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_URL_ENCODED_H
+
+#include "google/cloud/version.h"
+#include <string>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
```
devbww, line 23: Q.  Does this need to be externally visible?

PR #9025, opened 2022-05-19
feat(GCS+gRPC): adjust default channel count for DirectPath

coryan:
When using DirectPath it makes little or no sense to use more than one
gRPC channel.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9025)


dbolduc (approved):
Your stance on `absl::StartsWith` is clear.

PR #9032, opened 2022-05-20
fix(common)!: adjust env/option/default override order for endpoints

devbww:
Make the preference order be (1) `${emulator_env}`, (2) `${endpoint_env}`,
(3) `EndpointOption`, and (4) default value.

The changes to the tests are good indications that the new ordering is more
logical.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9032)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,16 @@ to grant, change, and revoke access to all
 [resource-types-with-policies]: https://cloud.google.com/iam/docs/resource-types-with-policies
 [iam-policy-link]: https://cloud.google.com/iam/docs/manage-access-other-resources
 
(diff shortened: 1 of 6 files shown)


dbolduc (approved):

coryan (approved):
Do we need to make similar changes to other services?  I am wondering if Storage or Pub/Sub are not doing the Right Thing.

devbww:
> Do we need to make similar changes to other services? I am wondering if Storage or Pub/Sub are not doing the Right Thing.

The only other one I found that looked questionable was `google/cloud/storage/internal/grpc_client.cc:DefaultOptionsGrpc()`, but I didn't want to touch it without further advice.

And, of course, I may have missed others.

PR #9033, opened 2022-05-20
impl(bigtable): add unimplemented Connection API

dbolduc:
Part of the work for #8798 

Add the DataConnection API as defined in this (internal) [document](https://docs.google.com/document/d/1NjGK7_1gfkxavxFxIHcKoDyxPWGITnYL5JBH18ZNG6A/edit?resourcekey=0-A9nSevk5ZyWVGzhxtXE93Q).

The one difference is that I forgot about resources (app profile id + table name) in the document. We do not want to store resources in the Connection. So the Table must pass the resource names with each request.

I tried my best to minimize the `NOLINT` eye sore. `NOLINTBEGIN`/`NOLINTEND` [is a thing](https://clang.llvm.org/extra/clang-tidy/), but not until clang-tidy 15.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9033)

### DIFF
--- a/google/cloud/bigtable/internal/data_connection.cc
+++ b/google/cloud/bigtable/internal/data_connection.cc
@@ -16,6 +16,7 @@
 #include "google/cloud/bigtable/internal/bigtable_stub_factory.h"
 #include "google/cloud/bigtable/internal/data_connection_impl.h"
 #include "google/cloud/bigtable/internal/defaults.h"
+#include "google/cloud/bigtable/internal/row_reader_impl.h"
 #include "google/cloud/bigtable/options.h"
 #include "google/cloud/background_threads.h"
 #include "google/cloud/common_options.h"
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/internal/data_connection.cc
```diff
@@ -26,9 +27,122 @@ namespace google {
...
+std::vector<bigtable::FailedMutation> DataConnection::BulkApply(
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
+    std::string const&, std::string const&, bigtable::BulkMutation mut) {
+  return MakeUnimplementedFailedMutations(mut.size());
+}
+
+future<std::vector<bigtable::FailedMutation>> DataConnection::AsyncBulkApply(
+    // NOLINTNEXTLINE(performance-unnecessary-value-param)
+    std::string const&, std::string const&, bigtable::BulkMutation mut) {
+  return make_ready_future(MakeUnimplementedFailedMutations(mut.size()));
```
coryan, line 69: I see why you need that function now, LGTM.

coryan (approved):

PR #9044, opened 2022-05-22
test(spanner): check EndpointOption setting in a compatible environment

devbww:
After #9032, setting the `EndpointOption` only has an effect when there is
no prevailing `${endpoint_env}`, which there might be if we are also running
integration tests against an alternate service.

So, remove `${endpoint_env}` (`GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT`
for Spanner), and `${emulator_env}` for good measure (`SPANNER_EMULATOR_HOST`),
from the environment of the relevant unit test for `MakeConnection()`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9044)

### DIFF
--- a/google/cloud/spanner/client_test.cc
+++ b/google/cloud/spanner/client_test.cc
@@ -21,6 +21,7 @@
 #include "google/cloud/spanner/timestamp.h"
 #include "google/cloud/spanner/value.h"
 #include "google/cloud/testing_util/is_proto_equal.h"
+#include "google/cloud/testing_util/scoped_environment.h"
 #include "google/cloud/testing_util/status_matchers.h"
 #include "absl/memory/memory.h"
 #include "absl/types/optional.h"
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/client_test.cc
```diff
@@ -395,12 +397,19 @@ TEST(ClientTest, MakeConnectionOptionalArguments) {
   EXPECT_EQ(conn->options().get<EndpointOption>(),
             spanner_internal::DefaultOptions().get<EndpointOption>());
 
-  conn = MakeConnection(db, Options{}.set<EndpointOption>("endpoint"));
-  ASSERT_NE(conn, nullptr);
-  ASSERT_TRUE(conn->options().has<EndpointOption>());
-  EXPECT_NE(conn->options().get<EndpointOption>(),
-            spanner_internal::DefaultOptions().get<EndpointOption>());
-  EXPECT_EQ(conn->options().get<EndpointOption>(), "endpoint");
+  {
```
coryan, line 400: nit: remove the braces?  There is no code after you close the explicit scope.

coryan (approved):

google/cloud/spanner/client_test.cc
devbww: My thought: One day there might be extra code, or something might get moved, and in the meantime it documents the desired extent of the effect.

PR #9051, opened 2022-05-24
impl(bigtable): add RowReader factory for testing

### LINKED ISSUE #8797: Make `bigtable::RowReader` mockable
- [x] refactor class into a pimpl class
- [ ] add MockRowReader factory function

This is mostly done in: https://github.com/googleapis/google-cloud-cpp/compare/main...dbolduc:mockable-row-reader

... except we want to call the current implementation: `LegacyRowReaderImpl` (and leave `DefaultRowReaderImpl` for the implementation that interacts with a Stub)

dbolduc:
Fixes #8797 

Similar to what we decided for `StreamRange`'s, add a simple factory function to construct a `RowReader` that accepts a vector of rows + a final status. We are not adding a google mock object that extends `RowReaderImpl` at this time.

I kept `google_cloud_cpp_bigtable_mocks` as an INTERFACE library because cmake is hard.

### DIFF
--- a/google/cloud/bigtable/mocks/mock_row_reader_test.cc
+++ b/google/cloud/bigtable/mocks/mock_row_reader_test.cc
@@ -1,0 +1,148 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


google/cloud/bigtable/mocks/mock_row_reader_test.cc
```diff
@@ -0,0 +1,148 @@
...
+        rows.emplace_back((*std::move(sor)).row_key());
+      }
+    }
+  }
+
+  std::vector<std::string> rows;
+  Status final_status;
+};
+
+TEST(MakeTestRowReader, Empty) {
```
coryan, line 52: Testing the test function .. we must go deeper.

coryan (approved):

PR #9056, opened 2022-05-24
doc: update generated endpoint documentation during generate-libraries

devbww:
Teach the `generate-libraries` build how to update the generated endpoint
documentation.

Run it to update the `baremetalsolution` doc, which is the only one that
still has the required injection markers.  A followup PR will re-add these
markers to all the other `main.dox` files.

This, and the recent addition of `external/googleapis/update_libraries.sh`
to the `generate-libraries` build, means the the "How-to Guide" for adding
a generated library can be simplified, but that remains a TODO for now.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9056)

### DIFF
--- a/google/cloud/baremetalsolution/doc/main.dox
+++ b/google/cloud/baremetalsolution/doc/main.dox
@@ -41,8 +41,8 @@ which should give you a taste of the Bare Metal Solution API C++ client library 
 
 <!-- inject-endpoint-env-vars-start -->
 
-- `GOOGLE_CLOUD_CPP_BARE_METAL_SOLUTION_ENDPOINT=...` changes the default endpoint
-  (baremetalsolution.googleapis.com) used by `BareMetalSolutionConnection`.
+- `GOOGLE_CLOUD_CPP_BARE_METAL_SOLUTION_ENDPOINT=...` changes the default `EndpointOption`,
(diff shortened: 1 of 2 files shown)


google/cloud/baremetalsolution/doc/main.dox
```diff
@@ -41,8 +41,8 @@ which should give you a taste of the Bare Metal Solution API C++ client library
 
 <!-- inject-endpoint-env-vars-start -->
 
-- `GOOGLE_CLOUD_CPP_BARE_METAL_SOLUTION_ENDPOINT=...` changes the default endpoint
-  (baremetalsolution.googleapis.com) used by `BareMetalSolutionConnection`.
+- `GOOGLE_CLOUD_CPP_BARE_METAL_SOLUTION_ENDPOINT=...` changes the default `EndpointOption`,
```
dbolduc, line 44: I don't think it is accurate to say it changes the default `EndpointOption`.

What about just: "`BLAH_ENDPOINT=...` overrides the `EndpointOption` used by `MakeBlahConnection()`."

devbww: Fair enough.  I took what you had, but kept the mention of the default value.  PTAL.

dbolduc (approved):

PR #9060, opened 2022-05-24
ci: enable GCS+gRPC production integration tests

### LINKED ISSUE #7257: Restore integration tests disabled for GCS+gRPC
There is a small number of bugs in the (pre-release not GA) GCS+gRPC service, specially with the v2/ protos. I have disabled the tests that fail, this is a bug to reference from the TODO entries.

coryan:
The integration tests against production were disabled while the protos
stabilized. There are no anticipated breaking changes at this point.

We cannot pass the options created for a `GrpcClient` to a `CurlClient`.
The latter has a very subtle interpretation of `AuthorityOption` as it
supports two separate services: `iamcredentials.googleapis.com` is
used to sign blobs.

A number of tests remain disabled because they use metadata operations
in gRPC, which are not ready in production. Changed their TODO entries
to point to the right bug.

Fixes #7257

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9060)

### DIFF
--- a/google/cloud/storage/grpc_plugin.cc
+++ b/google/cloud/storage/grpc_plugin.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/storage/grpc_plugin.h"
 #include "google/cloud/storage/internal/grpc_client.h"
 #include "google/cloud/storage/internal/hybrid_client.h"
+#include "google/cloud/internal/getenv.h"
 
 namespace google {
 namespace cloud {
@@ -22,12 +23,15 @@ namespace storage_experimental {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 google::cloud::storage::Client DefaultGrpcClient(Options opts) {
-  opts = google::cloud::storage::internal::DefaultOptionsGrpc(std::move(opts));
-  auto config = opts.get<GrpcPluginOption>();
+  using ::google::cloud::internal::GetEnv;
+  auto const config = GetEnv("GOOGLE_CLOUD_CPP_STORAGE_GRPC_CONFIG")
+                          .value_or(opts.get<GrpcPluginOption>());
   if (config == "none" || config.empty()) {
     return google::cloud::storage::Client(std::move(opts));
   }
   if (config == "metadata") {
+    opts =
+        google::cloud::storage::internal::DefaultOptionsGrpc(std::move(opts));
     return storage::internal::ClientImplDetails::CreateClient(
         storage::internal::GrpcClient::Create(opts));
   }
--- a/google/cloud/storage/grpc_plugin_test.cc
+++ b/google/cloud/storage/grpc_plugin_test.cc
@@ -1,0 +1,129 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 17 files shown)


google/cloud/storage/grpc_plugin.cc
```diff
@@ -15,19 +15,23 @@
...
-  auto config = opts.get<GrpcPluginOption>();
+  using ::google::cloud::internal::GetEnv;
+  auto const config = GetEnv("GOOGLE_CLOUD_CPP_STORAGE_GRPC_CONFIG")
+                          .value_or(opts.get<GrpcPluginOption>());
   if (config == "none" || config.empty()) {
     return google::cloud::storage::Client(std::move(opts));
   }
   if (config == "metadata") {
+    opts =
+        google::cloud::storage::internal::DefaultOptionsGrpc(std::move(opts));
```
dbolduc, line 34: I am confused. Is the point that we do not want to `DefaultOptionsGrpc` if `config == "media"` ? I would have guessed that `HybridClient` wants the options defaulted.

Also, pulling the env var stuff out of `DefaultOptionsGrpc()` means we have no test that the env var overrides the `GrpcPluginOption`... that seems less good to me.

google/cloud/storage/internal/grpc_client_test.cc
```diff
@@ -182,24 +179,6 @@ std::shared_ptr<GrpcClient> CreateTestClient(
   return GrpcClient::CreateMock(std::move(stub), TestOptions());
 }
 
-TEST_F(GrpcClientTest, DefaultOptionsGrpcNoEnv) {
```
dbolduc, line 185: :/

google/cloud/storage/grpc_plugin.cc
dbolduc: Sorry, I just read the PR description. You explained the point.

It would still be nice if we could unit test the env var behavior. Can we default the authority option differently within `DefaultOptionsGrpc()` depending on the config?

coryan: > Sorry, I just read the PR description. You explained the point.
> 
> It would still be nice if we could unit test the env var behavior. Can we default the authority option differently within `DefaultOptionsGrpc()` depending on the config?

That would spread the knowledge of what `GrpcPluginOption` does across multiple files.  Does not seem very desirable.  I created a new unit test for the `DefaultGrpcClient()` function.

google/cloud/storage/grpc_plugin_test.cc
```diff
@@ -0,0 +1,129 @@
...
+using ::google::cloud::storage::internal::GrpcClient;
+using ::google::cloud::storage::internal::HybridClient;
+using ::google::cloud::storage::internal::RetryClient;
+using ::google::cloud::testing_util::ScopedEnvironment;
+using ::testing::IsNull;
+using ::testing::NotNull;
+
+TEST(GrpcPluginTest, MetadataConfigCreatesGrpc) {
+  // Disable the RestClient, while this is not the default, it is enabled in
+  // some of our CI builds.
```
dbolduc, line 40: maybe: `// Explicitly disable the RestClient, which may be enabled by our CI builds.` here and elsewhere

google/cloud/storage/grpc_plugin_test.cc
```diff
@@ -0,0 +1,129 @@
...
+  auto rest = ScopedEnvironment("GOOGLE_CLOUD_CPP_STORAGE_HAVE_REST_CLIENT",
+                                absl::nullopt);
+  // Disable logging, also enabled in most of our CI builds
```
dbolduc, line 43: `// Explicitly disable logging, which may be enabled by our CI builds.`

dbolduc (approved):
That works. some nits about the comments

PR #9062, opened 2022-05-24
doc: update generated endpoint documentation

devbww:
Restore the "inject-endpoint-env-vars" markers to the `main.dox` files,
and run the `generate-libraries` build to update the descriptions of the
various endpoint environment variables.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9062)

### DIFF
--- a/google/cloud/accessapproval/doc/main.dox
+++ b/google/cloud/accessapproval/doc/main.dox
@@ -40,8 +40,13 @@ which should give you a taste of the Access Approval API C++ client library API.
 
 ## Environment Variables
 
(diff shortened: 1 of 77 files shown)


dbolduc (approved):
Thanks for doing this.

For the record, the `<!-- inject-endpoint-env-vars-* -->` tags were never there to begin with... We only thought to add them about ~90 libraries in, when you suggested the operation should be idempotent. I was too lazy to go through and update the already-generated documentation.

devbww:
No worries, and thanks for correcting the history.

The next time someone adds a new library, we should consider how all of this will simplify the "How-to Guide: Adding generated libraries" process.

PR #9064, opened 2022-05-24
ci: automatically start GCB builds for trusted contributors

coryan:
I am mostly thinking about Renovate Bot contributions, but it applies
to anything else that the `trusted-contributor` bot knows about.

### DIFF
--- a/.github/trusted-contribution.yml
+++ b/.github/trusted-contribution.yml
@@ -1,3 +1,5 @@
 annotations:
   - type: label
     text: "kokoro:force-run"
+  - type: comment
+    text: "/gcbrun"


dbolduc (approved):
I wonder if we have to explicitly list renovate-bot like this does: https://github.com/googleapis/java-redis/blob/48f52a344c3b8db445ba81c2e5cac60e5eaabe7c/.github/trusted-contribution.yml

coryan:
> I wonder if we have to explicitly list renovate-bot like this does: https://github.com/googleapis/java-redis/blob/48f52a344c3b8db445ba81c2e5cac60e5eaabe7c/.github/trusted-contribution.yml

We can try and find out.  But FWIW, this is already starting the Kokoro builds.

Issue #8799, opened 2022-04-26
Implement unary RPCs for Modern Bigtable Data API


dbolduc:
- [x] Apply
- [x] CheckAndMutateRow
- [x] ReadModifyWriteRow
- [x] AsyncApply
- [ ] AsyncCheckAndMutateRow
- [x] AsyncReadModifyWriteRow

Must be done after #8798.

PR #9070, opened 2022-05-25
impl(bigtable): implement DataConnection::Apply()

dbolduc:
Part of the work for #8799 , #8798 

This PR implements DataConnection::Apply(). It includes getters for call vs. connection options. The Table::Apply() piece will be addressed in a later PR, when I have set up the integration tests to exercise the new code path.

The old code path is [here](https://github.com/googleapis/google-cloud-cpp/blob/e6a8a6ffe1396cdd16ce4405d8578fbd9420d2cc/google/cloud/bigtable/table.cc#L72)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9070)

### DIFF
--- a/google/cloud/bigtable/internal/data_connection_impl.cc
+++ b/google/cloud/bigtable/internal/data_connection_impl.cc
@@ -15,6 +15,8 @@
 #include "google/cloud/bigtable/internal/data_connection_impl.h"
 #include "google/cloud/bigtable/internal/defaults.h"
 #include "google/cloud/background_threads.h"
+#include "google/cloud/idempotency.h"
+#include "google/cloud/internal/retry_loop.h"
 #include <memory>
 
 namespace google {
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/internal/data_connection_impl.cc
```diff
@@ -31,6 +33,33 @@ DataConnectionImpl::DataConnectionImpl(
...
+  mut.MoveTo(request);
+
+  auto idempotent_policy = idempotency_policy();
+  bool const is_idempotent = std::all_of(
+      request.mutations().begin(), request.mutations().end(),
+      [&idempotent_policy](google::bigtable::v2::Mutation const& m) {
+        return idempotent_policy->is_idempotent(m);
+      });
+
+  auto sor = google::cloud::internal::RetryLoop(
```
coryan, line 51: nit: I think we avoid names like this.  `response` may be better.  Your call.

coryan (approved):

google/cloud/bigtable/internal/data_connection_impl.cc
dbolduc: hmm.. I think I will keep using `sor` but feel self-conscious about it.

PR #9072, opened 2022-05-25
feat(GCS+gRPC): timeout downloads on lack of progress

### LINKED ISSUE #7300: Timeout based on progress vs. completion
The REST implementation of the storage library times out requests based on "lack of progress", i.e., if no data has been received or transmitted in X seconds then we timeout the request.  gRPC does not have a native API to implement such timeouts, all timeouts are based on "completion", including timeouts for streaming RPCs which can take minutes or even hours to actually complete.

The gRPC team has indicated that no timeouts based on progress or per-Read / per-Write call will be implemented in the foreseeable future.

To implement "lack of progress" timeouts with GCS+gRPC we would need use the asynchronous APIs for streaming RPCs, and implement the timeouts ourselves.  This is complicated because using the asynchronous APIs usually requires having a set of background threads too.

coryan:
Previously we timed out a download if it failed to complete the download
in a given time. This requires setting excessive timeouts for
applications that download both small objects (which take milliseconds)
and large objects (which can take minutes).

Furthermore, some applications start a download and then consume the
data slowly.  The timeout values for such applications would need to be
too large too.

This PR uses the asynchronous API to download objects. This API allows
us to timeout a `Read()` call, and abort the download if it is not
making progress.

Fixes #7300

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9072)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -363,15 +363,17 @@ StatusOr<std::unique_ptr<ObjectReadSource>> GrpcClient::ReadObject(
   }
   auto context = absl::make_unique<grpc::ClientContext>();
   ApplyQueryParameters(*context, request);
-  auto const timeout = options_.get<DownloadStallTimeoutOption>();
-  if (timeout.count() != 0) {
-    context->set_deadline(std::chrono::system_clock::now() + timeout);
-  }
   auto proto_request = GrpcObjectRequestParser::ToProto(request);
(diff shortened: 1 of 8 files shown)


dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9072)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #9075, opened 2022-05-26
test(spanner): limit special instance-config filtering to prod

devbww:
When selecting an `InstanceConfig` to use in the samples that create
a new `Instance`, we have some special filtering to avoid locations
where we run integration tests.  This filtering is specific to the
production Spanner service, and so should not be applied when using
an alternative like the staging service.

This allows the `spanner_create_database_with_encryption_key` sample
to run on the staging service, as its encryption key is tied to a
location whose `InstanceConfig` was previously discarded by the filter.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9075)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -3728,21 +3728,27 @@ void RunAllSlowInstanceTests(
 
   if (!run_slow_instance_tests) return;
 
+  auto const prod_service = google::cloud::internal::GetEnv(
+                                "GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT")
+                                .value_or("")
+                                .empty();
   auto const config_id =
       Basename(google::cloud::spanner_testing::PickInstanceConfig(
           google::cloud::Project(project_id), generator,
-          [](google::spanner::admin::instance::v1::InstanceConfig const&
-                 config) {
+          [prod_service](
+              google::spanner::admin::instance::v1::InstanceConfig const&
+                  config) {
             auto config_id = Basename(config.name());
             // Skip non-US configs. They are too slow.
-            if (absl::StartsWith(config_id, "regional-us-")) {
-              // Exclude US configs where we keep most test instances.
-              if (config_id != "regional-us-central1" &&
-                  config_id != "regional-us-east1") {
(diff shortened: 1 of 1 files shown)


reinasalondra (approved):

google/cloud/spanner/samples/samples.cc
```diff
@@ -3728,21 +3728,27 @@ void RunAllSlowInstanceTests(
...
+  auto const prod_service = google::cloud::internal::GetEnv(
+                                "GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT")
+                                .value_or("")
+                                .empty();
   auto const config_id =
       Basename(google::cloud::spanner_testing::PickInstanceConfig(
           google::cloud::Project(project_id), generator,
-          [](google::spanner::admin::instance::v1::InstanceConfig const&
-                 config) {
+          [prod_service](
```
coryan, line 3738: nit: maybe time to move the functor out?

google/cloud/spanner/samples/samples.cc
```diff
@@ -3728,21 +3728,27 @@ void RunAllSlowInstanceTests(
...
+                  config) {
             auto config_id = Basename(config.name());
             // Skip non-US configs. They are too slow.
-            if (absl::StartsWith(config_id, "regional-us-")) {
-              // Exclude US configs where we keep most test instances.
-              if (config_id != "regional-us-central1" &&
-                  config_id != "regional-us-east1") {
-                return true;
+            if (!absl::StartsWith(config_id, "regional-us-")) return false;
+            if (prod_service) {
```
coryan, line 3744: maybe:

```cc
if (!prod_service) return true;
return config_id != "regional-us-central1" && config_id != "regional-us-east1";
```

coryan (approved):

google/cloud/spanner/samples/samples.cc
devbww: OK, done.

google/cloud/spanner/samples/samples.cc
devbww: Done.

coryan (approved):

PR #9077, opened 2022-05-26
fix(GCS+gRPC): use "upsert" semantics for `*AccessControl`

### LINKED ISSUE #9055: Fix semantics for `*AccessControl` over gRPC
Labeling this as a bug, as these functions should behave the same over JSON vs. gRPC.

Turns out JSON always succeeds for `*AccessControl: insert`: if the entity does not exist it is created, if it exists, it is updated.
Same thing for `*AccessControl: update` and `AccessControl: patch`.

This is not the case for `AccessControl: delete`, if the object does not exist, JSON returns a 404 error.

coryan:
Over JSON the semantics are "update-or-insert" for `*AccessControl`
operations, when we emulate these functions over gRPC we should
follow the same semantics.

Fixes #9055

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9077)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -42,6 +42,21 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
 namespace {
 
+template <typename AccessControl>
+StatusOr<std::vector<AccessControl>> UpsertAcl(std::vector<AccessControl> acl,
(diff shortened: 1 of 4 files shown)


dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9077)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #9078, opened 2022-05-26
feat(generator): `SetIamPolicy()` can be idempotent

### LINKED ISSUE #2595: SetIamPolicy should use etag to determine idempotency.
The InstanceAdmin::SetIamPolicy should examine the `etag` field to determine if the operation is idempotent, we currently assume that it never is.  We need to verify the server actually uses the etag parameter though.

coryan:
The `SetIamPolicy()` operations can be idempotent if the request has an
ETag value. The ETag value is a pre-condition, that changes when the
operation succeeds.  In other words, the operation can succeed at
most once: all such operations are idempotent.

Fixes #2595

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9078)

### DIFF
--- a/generator/internal/idempotency_policy_generator.cc
+++ b/generator/internal/idempotency_policy_generator.cc
@@ -155,7 +155,25 @@ Status IdempotencyPolicyGenerator::GenerateCc() {
     "  }\n");
   // clang-format on
 
+  auto is_set_iam_policy = [](google::protobuf::MethodDescriptor const& m) {
+    return m.output_type()->full_name() == "google.iam.v1.Policy" &&
+           m.input_type()->full_name() == "google.iam.v1.SetIamPolicyRequest";
+  };
   for (auto const& method : methods()) {
+    // SetIamPolicy() methods are common. They are idempotent if the request
+    // has an `ETag` pre-condition.  This is one of the few (only?) cases where
+    // a policy based on the request contents is known.
+    if (is_set_iam_policy(method)) {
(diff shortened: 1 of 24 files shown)


generator/internal/idempotency_policy_generator.cc
```diff
@@ -155,7 +155,25 @@ Status IdempotencyPolicyGenerator::GenerateCc() {
     "  }\n");
   // clang-format on
 
+  auto is_set_iam_policy = [](google::protobuf::MethodDescriptor const& m) {
+    return m.output_type()->full_name() == "google.iam.v1.Policy" &&
+           m.input_type()->full_name() == "google.iam.v1.SetIamPolicyRequest";
+  };
```
devbww, line 161: Q. Perhaps this deserves a `IsSetIamPolicy()` helper in `predicate_utils.h`?

We do have some similar logic in `ClientGenerator::ClientGenerator()`, but it isn't as simple.

devbww (approved):

generator/internal/idempotency_policy_generator.cc
coryan: The `ClientGenerator()` is trying to do something more interesting, the signature extensions are needed to use the `SetIamPolicy(resource, policy)` and `GetIamPolicy(resource)` overloads, and to use them in the loop for `SetIamPolicy(resource, updater)`.

 By the time a request gets to the idempotency policy the request is always in the form of a request protos and we do not need the additional complexity.

I am going to leave this as-is.

Issue #9081, opened 2022-05-27
Deprecate and remove `bigtable::*EndpointOption`s


dbolduc:
Deprecate these options: https://github.com/googleapis/google-cloud-cpp/blob/7b535eeb2db520e47988a02e6601199268a2fa08/google/cloud/bigtable/options.h#L54-L73

We do not need them. The endpoint should be configured using `EndpointOption` like all other libraries do.

We will still need to have `internal` versions of these options. These will be set by `ClientOptions` and respected by `DataClient`, `AdminClient`, `InstanceAdminClient` to maintain backwards compatibility. They just do not need to be public.

Also, it would be nice to clean up the logic around defaulting. Namely: the logic in `bigtable_internal::DefaultOptions()` is mostly unnecessary. Aside from the endpoint it sets some fields needed only for the Data API, and does the same work as the generated function to default options for the Admin APIs. The one difference is that the Instance Admin has 2 emulator environment variables. That is annoying, but still there is a simpler way to write the code.

dbolduc:
These are deprecated in documentation, but no compiler warnings are emitted.

https://github.com/googleapis/google-cloud-cpp/blob/c805c29d9d4121a81b89dfa1bbc0f1e15e3ca12a/google/cloud/bigtable/options.h#L91

mpeddada1:
Fixed in https://github.com/googleapis/google-cloud-cpp/pull/15856

PR #9082, opened 2022-05-27
feat(bigtable): configure `*Client`s with `EndpointOption`

### LINKED ISSUE #9027: Bigtable (legacy) `*Client` classes should respect `EndpointOption` when configured directly with `Options`
When calling something like `MakeDataClient("blah", "blah", options)`, customers can only configure the endpoint using `bigtable::DataEndpointOption`, not `g::c::EndpointOption` (like every other library does). This is an artifact of handling `bigtable::ClientOptions` which contains information for three different endpoint options. (If you pass an `EndpointOption` to the `bigtable::ClientOptions` it would not know which one to set)

I think both `google::cloud::EndpointOption` and `bigtable::DataEndpointOption` should work when passed to this function.

<hr />

Actually, we should just retire the `bigtable::*EndpointOption`s. They do nothing for us. We can use `internal` options to store the configuration of `ClientOptions` to maintain backwards compatibility.

dbolduc:
Fixes #9027 and does a tiny bit of the work for #9081 

This change:
1. Makes it possible to use `EndpointOption` with `DataClient`, `AdminClient`, and `InstanceAdminClient`
2. Recommends against using `DataEndpointOption`, `AdminEndpointOption`, and `InstanceAdminEndpointOption` to configure the above classes in documentation only.

Note that our `Default*Options()` functions expect `DefaultOptions()` to provide input via `*EndpointOption`. (for example [here](https://github.com/googleapis/google-cloud-cpp/blob/7b535eeb2db520e47988a02e6601199268a2fa08/google/cloud/bigtable/internal/defaults.cc#L199-L200)). I will clean this stuff up as part of #9081.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9082)

### DIFF
--- a/google/cloud/bigtable/internal/defaults.cc
+++ b/google/cloud/bigtable/internal/defaults.cc
@@ -81,9 +81,15 @@ Options DefaultOptions(Options opts) {
 
   if (opts.has<EndpointOption>()) {
     auto const& ep = opts.get<EndpointOption>();
-    opts.set<DataEndpointOption>(ep);
-    opts.set<AdminEndpointOption>(ep);
-    opts.set<InstanceAdminEndpointOption>(ep);
+    if (!opts.has<DataEndpointOption>()) {
+      opts.set<DataEndpointOption>(ep);
+    }
+    if (!opts.has<AdminEndpointOption>()) {
+      opts.set<AdminEndpointOption>(ep);
+    }
+    if (!opts.has<InstanceAdminEndpointOption>()) {
+      opts.set<InstanceAdminEndpointOption>(ep);
+    }
   }
 
   auto emulator = GetEnv("BIGTABLE_EMULATOR_HOST");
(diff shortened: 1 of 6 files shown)


google/cloud/bigtable/internal/defaults.cc
```diff
@@ -79,6 +79,13 @@ int DefaultConnectionPoolSize() {
 Options DefaultOptions(Options opts) {
   using ::google::cloud::internal::GetEnv;
 
+  if (opts.has<EndpointOption>()) {
+    auto const& ep = opts.get<EndpointOption>();
+    opts.set<DataEndpointOption>(ep);
```
coryan, line 84: What if the application already provided `DataEndpointOption` for some reason?

coryan: Seems to me this is a breaking change.  The behavior is now different.

dbolduc: Fixed. PTAL

coryan (approved):

Issue #7976, opened 2022-01-15
Cloud Document AI API: `documentai.googleapis.com`


dbolduc:
- [x] Make the quickstart work, and run it in our ci builds.
- [ ] Add a helper function for setting the EndpointOption.

dbolduc:
Still to-do:

- [ ] Make the quickstart work, and run it in our ci builds.
- [ ] Add a helper function for setting the `EndpointOption`.

devbww:
We don't need the helper function for setting the EndpointOption now.  You can pass the location directly to `MakeDocumentProcessorServiceConnection()` instead.

PR #9083, opened 2022-05-27
fix(documentai)!: revert GA declaration

dbolduc:
Declaring GA should have been blocked on these items: https://github.com/googleapis/google-cloud-cpp/issues/7976#issuecomment-1047073239

Our metrics show that no adoption of this library. So nobody should be affected by reverting the GA declaration.


dbolduc:
Given #9108, this PR will not be necessary.

PR #9091, opened 2022-05-27
fix(bigtable): remove unnecessary sleeps from BulkApply

### LINKED ISSUE #9089: `bigtable::Table::BulkApply` sleeps unnecessarily
https://github.com/googleapis/google-cloud-cpp/blob/30c06a2c3ee6b99bded6670382d79d98034019fa/google/cloud/bigtable/table.cc#L171

We should check that there are no pending mutations before sleeping, instead of after. For example, we could send one successful request (with no failed mutations) and the library would sleep before noticing that there are no pending mutations.

dbolduc:
Fixes #9089 ... although this should be tested properly.

We now check the loop condition before sleeping instead of after.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9091)

### DIFF
--- a/google/cloud/bigtable/table.cc
+++ b/google/cloud/bigtable/table.cc
@@ -151,6 +151,7 @@ future<Status> Table::AsyncApply(SingleRowMutation mut) {
 }
 
 std::vector<FailedMutation> Table::BulkApply(BulkMutation mut) {
+  if (mut.empty()) return {};
   grpc::Status status;
 
   // Copy the policies in effect for this operation.  Many policy classes change
@@ -162,7 +163,7 @@ std::vector<FailedMutation> Table::BulkApply(BulkMutation mut) {
 
   bigtable::internal::BulkMutator mutator(app_profile_id_, table_name_,
                                           *idempotent_policy, std::move(mut));
-  while (mutator.HasPendingMutations()) {
+  while (true) {
     grpc::ClientContext client_context;
     backoff_policy->Setup(client_context);
     retry_policy->Setup(client_context);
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/table.cc
```diff
@@ -171,6 +172,7 @@ std::vector<FailedMutation> Table::BulkApply(BulkMutation mut) {
     if (!status.ok() && !retry_policy->OnFailure(status)) {
```
coryan, line 172: My proposal was:

```cc
while (mutator.HasPendingMutations()) { 
....  
  status = mutator.MakeOneRequest(*client_, client_context);
  if (status.ok()) continue;
  if (!retry_policy->OnFailure(status)) break;
...
}
```

I think your approach saves the initial `HasPendingMutations()` probably worth it.

coryan (approved):

google/cloud/bigtable/table.cc
```diff
@@ -171,6 +172,7 @@ std::vector<FailedMutation> Table::BulkApply(BulkMutation mut) {
...
       break;
     }
+    if (!mutator.HasPendingMutations()) break;
     auto delay = backoff_policy->OnCompletion(status);
     std::this_thread::sleep_for(delay);
```
devbww, line 177: Why do we backoff on success at all?

dbolduc: It's just how the RPC works. The response itself has fields for errors. So the overall response can be OK, but still point us to failed mutations that we should retry: https://github.com/googleapis/googleapis/blob/1848f0849e4c18168e39c796127c705eaa03b776/google/bigtable/v2/bigtable.proto#L411-L427

coryan: Because this API can be partially successful.  The RPC returns successfully with a list of mutations that failed with a transient error.

devbww: Thanks.

coryan (approved):

PR #9098, opened 2022-05-29
fix(GCS):ccccccgelgvtgjvdjnrtblllejltbtneukgbrlthcddj Silence false positive maybe uninitialized from GCC

tobim:


devbww:
> > What revision of GCC 11 are you using? In what mode?
> 
> GCC 11.3 with the CMake scaffold. I didn't set any language specific options,
> so I think the dialect should be `gnu++17`.

I am unable to reproduce it with GCC 11.3.0, across a variety of optimization levels.

> `abs::optional` should indeed be an alias to `std::optional`, so it might really be
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80635. I agree that the proposed
> workaround is not pretty, but neither is `-Wno-error-maybe-uninitialized`.

I'd still say the flag is less non-pretty, and also guaranteed to work no matter exactly how/where the flow analysis fails.

I would also suggest that the "maybe" in the diagnostics name means it probably shouldn't be made an error.

coryan:
I am inclined to close this without merging.  I am not sure why @tobim's version of GCC emits the warning while our version does not (vendor patches?  something in the standard library or systems headers that changes the compiler state?) But sprinkling the code with pragmas to disable a diagnostic that we have good reason to believe is a compiler bug (or a limitation of its static analysis) seems wasteful.  Is not like we are ignoring the warning everywhere, we have builds with GCC >= 11, with the warning enabled, and they pass.  We also have builds using  MemorySanitizer, that would detect the problem if it was real.

In the environments where the warning is emitted, this only seems to affect tests.  The applications can still compile and link the library with all warnings enabled. 

In other words:

- in all cases the applications can use the library with `-Wall -Werror -Wmaybe-initialized` (the last one may be implied by `-Wall`)
- all our builds are able to compile the library and its tests with `-Wall -Werror -Wmaybe-initialzed`
- in some cases compiling the library tests requires `-Wall -Werror -Wno-error=maybe-initialized`

That seems like a good set of tradeoffs to me.

> 
> I would also suggest that the "maybe" in the diagnostics name means it probably shouldn't be made an error.

Good point a better workaround is `-Wno-error=maybe-unitialized`.  Leaves the warning there, but allows the code to continue.

PR #9104, opened 2022-05-30
feat: if available use compile features to request C++ >= 11

coryan:
As long as CMake >= 3.8, this will automatically require C++ >= 11 to
compile the library and their dependencies.  This is only relevant with
Apple Clang, the last compiler we support that defaults to C++98 (and
apparently will always default to C++98).  All other compilers we
support default to C++14 or higher.

While this only works for CMake >= 3.8, I expect this is almost always
the case on macOS.  In other words, this can fail to automate things in
theory, but should work well in practice.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9104)

### DIFF
--- a/cmake/EnableWerror.cmake
+++ b/cmake/EnableWerror.cmake
@@ -40,6 +40,13 @@ endfunction ()
 
 function (google_cloud_cpp_add_common_options target)
     google_cloud_cpp_silence_warnings_in_deps(${target})
+    # Require C++ >= 11.  Unfortunately CMake can only express such requirements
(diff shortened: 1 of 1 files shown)


devbww (commented):
Should this belong in `EnableWerror.cmake`?

coryan:
> Should this belong in `EnableWerror.cmake`?

Probably not, but really the problem is that the `google_cloud_cpp_add_common_options()` function does not belong in this file, the name and what it does is not exclusively related to `Werror`.  I created #9105 because all the modules need a cleanup and reorganization.

devbww (approved):

Issue #8802, opened 2022-04-26
Implement BulkApply for Modern Bigtable Data APi


dbolduc:
- [ ] Add stub support to `BulkMutator`
- [ ] Implement `DataConnection::BulkApply()`
- [ ] Implement `Table::BulkApply()`

PR #9114, opened 2022-05-31
impl(bigtable): BulkMutator can use Stub

dbolduc:
Part of the work for #8802 

Add ability for `BulkMutator` to make calls using `BigtableStub`. Move the test code for the calls that use `DataClient` into `legacy_bulk_mutator_test.cc` (the only change is to the license year). Rewrite the tests to use the stub in `bulk_mutator_test.cc`.

### DIFF
--- a/google/cloud/bigtable/internal/bulk_mutator.cc
+++ b/google/cloud/bigtable/internal/bulk_mutator.cc
@@ -79,7 +79,7 @@ google::bigtable::v2::MutateRowsRequest const& BulkMutatorState::BeforeStart() {
 }
 
 std::vector<int> BulkMutatorState::OnRead(
-    google::bigtable::v2::MutateRowsResponse& response) {
+    google::bigtable::v2::MutateRowsResponse&& response) {
   std::vector<int> res;
   for (auto& entry : *response.mutable_entries()) {
     // The type of `entry.index()` is a 64-bit int. But we can never create more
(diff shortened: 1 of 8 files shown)


google/cloud/bigtable/internal/bulk_mutator.cc
```diff
@@ -79,7 +79,7 @@ google::bigtable::v2::MutateRowsRequest const& BulkMutatorState::BeforeStart() {
 }
 
 std::vector<int> BulkMutatorState::OnRead(
-    google::bigtable::v2::MutateRowsResponse& response) {
+    google::bigtable::v2::MutateRowsResponse&& response) {
```
devbww, line 82: Should all these simply be `s/&&//` instead?

google/cloud/bigtable/internal/bulk_mutator_test.cc
```diff
@@ -11,13 +11,10 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
```
devbww, line 14: Nit: Restore separator line.

google/cloud/bigtable/internal/bulk_mutator.cc
dbolduc: After discussing further with @devbww, I am no longer convinced that using `&&` is much of a constraint.

So changed `OnRead` to pass by value, which is recommended by the style guide and more consistent with the rest of the library.

google/cloud/bigtable/internal/bulk_mutator_test.cc
dbolduc: Done.

google/cloud/bigtable/internal/bulk_mutator.cc
```diff
@@ -197,14 +197,42 @@ grpc::Status BulkMutator::MakeOneRequest(bigtable::DataClient& client,
...
+  struct UnpackVariant {
+    BulkMutatorState& state;
+    bool operator()(btproto::MutateRowsResponse&& r) {
+      state.OnRead(std::move(r));
+      return true;
+    }
+    bool operator()(Status&& s) {
+      state.OnFinish(std::move(s));
+      return false;
+    }
```
devbww, line 226: If these work with `&&` then they should work just as well without them too.

dbolduc: Fixed.

FWIW, I copied this: https://github.com/googleapis/google-cloud-cpp/blob/41d558bcd48aaba612f49d955d83db83d6f226e1/google/cloud/stream_range.h#L182-L194

PR #9118, opened 2022-05-31
feat(rest): add HasUnreadData method

scotthart:
Necessary for refactoring GCS to use rest library.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9118)


coryan (approved):
I think we used to call this `IsOpen()`.  I am Okay with the new name, it is a bit unusual maybe.  It may overstate the case when `curl_closed_ == false` but reading returns 0 bytes.

devbww:
ENOMERGE between release prep and version bump.  :-(

Issue #568, opened 2018-05-25
Implement asynchronous APIs for GCS


coryan:
The application should be able to overlap computation and I/O.  The application should be able to request that specific operations are executed asynchronously, for example:

```C++
namespace gcs = google::cloud::storage;
void AppCode(Bucket bucket, gcs::CompletionQueue cq) {
  std::future<gcs::BucketMetadata> meta = bucket.GetMetadata(cq);
  // .. do other stuff here, and when ready:
  meta.get(); // block until the response is ready.
}
```

This is a large task, it requires implementing completion queues, and doubles (if using `std::experimental::future`) or triples if using (`std::future` and callbacks) the number of APIs.

There is an existing design doc for the Bigtable library that is relevant.

coryan:
This is becoming more important.  It would solve problems with deleting and composing multiple objects.

One solution might be to hide the completion queue completely, and just offer a configuration for the number of background threads (but none of the thread properties). That would work for REST and gRPC.

coryan:
This is too open ended for a bug.  I am going to open more specific bugs and close this.

PR #9120, opened 2022-06-01
impl(GCS+gRPC): add `*Connection` class for async client

coryan:
Part of the work for #568

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9120)

### DIFF
--- a/google/cloud/storage/internal/async_connection_impl.h
+++ b/google/cloud/storage/internal/async_connection_impl.h
@@ -1,0 +1,88 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_STORAGE_INTERNAL_ASYNC_CONNECTION_IMPL_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_STORAGE_INTERNAL_ASYNC_CONNECTION_IMPL_H
+
+#include "google/cloud/storage/idempotency_policy.h"
+#include "google/cloud/storage/internal/async_connection.h"
+#include "google/cloud/storage/internal/storage_stub.h"
+#include "google/cloud/storage/options.h"
+#include "google/cloud/storage/retry_policy.h"
+#include "google/cloud/completion_queue.h"
+#include "google/cloud/future.h"
+#include "google/cloud/version.h"
+#include <memory>
+
+namespace google {
+namespace cloud {
+namespace storage_internal {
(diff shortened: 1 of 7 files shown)


google/cloud/storage/internal/async_connection_impl.h
```diff
@@ -0,0 +1,88 @@
...
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+class AsyncConnectionImpl : public AsyncConnection {
+ public:
+  explicit AsyncConnectionImpl(CompletionQueue cq,
+                               std::shared_ptr<StorageStub> stub,
+                               Options options);
+  ~AsyncConnectionImpl() override = default;
+
+  Options options() const override { return options_; }
```
dbolduc, line 40: do you want a test for this? maybe that one option from `DefaultOptionGrpc()` is set?

dbolduc (approved):

google/cloud/storage/internal/async_connection_impl.h
```diff
@@ -0,0 +1,88 @@
...
+
+  future<Status> AsyncDeleteObject(
+      storage::internal::DeleteObjectRequest const& request) override;
+
+ private:
+  std::unique_ptr<storage::RetryPolicy> retry_policy() {
+    auto const& options = internal::CurrentOptions();
+    if (options.has<storage::RetryPolicyOption>()) {
```
dbolduc, line 48: ugh, forgot to add this comment to my review.

The current options vs connection options behavior isn't tested. I decided to test it in bigtable, but [it was annoying](https://github.com/googleapis/google-cloud-cpp/blob/9c6256e77b03e83cee73faea2cf4880f807612df/google/cloud/bigtable/internal/data_connection_impl_test.cc#L113-L182). Your call on whether or not to add a test.

google/cloud/storage/internal/async_connection_impl.h
coryan: We check that `AuthorityOption` is set to the right value.  Is that what you meant?  I included a comment to explain that too.

google/cloud/storage/internal/async_connection_impl_test.cc
```diff
@@ -0,0 +1,95 @@
...
+                        UnorderedElementsAre(
+                            Pair("x-goog-quota-user", "test-quota-user"),
+                            Pair("x-goog-fieldmask", "field1,field2")));
+            EXPECT_THAT(request.bucket(), "projects/_/buckets/test-bucket");
+            EXPECT_THAT(request.object(), "test-object");
+            return make_ready_future(PermanentError());
+          });
+  CompletionQueue cq;
+  auto connection = MakeTestConnection(cq, mock);
+  OptionsSpan span(connection->options());
```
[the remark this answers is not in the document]
dbolduc: Ah. I think this is kind of a problem we will have to deal with eventually. I added a discussion item for the next meeting.

It's probably cleanest to just:
```c++
ConnectionImpl::Foo(FooRequest const& r) {
  auto options = internal::MergeOptions(internal::CurrentOptions(), options_);
  OptionsSpan span(options);
  
  return internal::RetryLoop(
    options.get<RetryPolicyOption>()->clone(),
    options.get<BackoffPolicyOption>()->clone(),
    options.get<IdempotencyPolicyOption>()->clone()->Foo(r),
    /* etc */ );
}
```
and drop the [policy getters](https://github.com/googleapis/google-cloud-cpp/blob/5f84aede7a748d723cfbabe979f376925b37127c/google/cloud/accessapproval/internal/access_approval_connection_impl.h#L87-L120) altogether.

PR #9132, opened 2022-06-01
refactor(storage): most of the object methods to use rest library

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9132)


coryan (approved):
Very neat code, thanks.

For a future PR, maybe we should make sure these functions are called as part of the `coverage` build?

PR #9152, opened 2022-06-02
refactor(bigtable): remove/isolate some test fixtures

dbolduc:
I am going to hurt myself on these eventually....

Drop test fixtures from `internal_data_connection_impl.cc`. Also do some light editing like `connection`->`conn` and adding an argument to the mutation factory functions.

Also split `table_test.cc` into tests that use the DataConnection and those that use the DataClient in a fixture. A lot of the DataClient tests should be salvaged, but I don't want to deal with them right now.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9152)

### DIFF
--- a/google/cloud/bigtable/legacy_table_test.cc
+++ b/google/cloud/bigtable/legacy_table_test.cc
@@ -1,0 +1,267 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


devbww (approved):

google/cloud/bigtable/legacy_table_test.cc
```diff
@@ -0,0 +1,267 @@
...
+namespace bigtable {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace {
+
+namespace btproto = ::google::bigtable::v2;
+using ::google::cloud::testing_util::FakeCompletionQueueImpl;
+using ::testing::HasSubstr;
+
+/// Define types and functions used in the tests.
+namespace {
```
coryan, line 32: An anonymous namespace inside the anonymous namespace?  We must go deeper.

PR #9155, opened 2022-06-03
impl(bigtable): implement DataConnection::BulkApply()

dbolduc:
Part of the work for #8802 

Here is the Table implementation of this retry loop: https://github.com/googleapis/google-cloud-cpp/blob/2a00f3d17425a15987674ba79548680be4330bed/google/cloud/bigtable/table.cc#L153-L180

The tests are mostly from: [table_bulk_apply_test.cc](https://github.com/googleapis/google-cloud-cpp/blob/2a00f3d17425a15987674ba79548680be4330bed/google/cloud/bigtable/table_bulk_apply_test.cc), but hopefully they are a bit clearer

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9155)

### DIFF
--- a/google/cloud/bigtable/internal/data_connection_impl.cc
+++ b/google/cloud/bigtable/internal/data_connection_impl.cc
@@ -78,12 +78,16 @@ std::vector<bigtable::FailedMutation> DataConnectionImpl::BulkApply(
   if (mut.empty()) return {};
   bigtable::internal::BulkMutator mutator(
       app_profile_id, table_name, *idempotency_policy(), std::move(mut));
-  auto retry = retry_policy();
-  auto backoff = backoff_policy();
+  // We wait to allocate the policies until they are needed as a
+  // micro-optimization.
+  std::unique_ptr<DataRetryPolicy> retry;
+  std::unique_ptr<BackoffPolicy> backoff;
   do {
     auto status = mutator.MakeOneRequest(*stub_);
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/internal/data_connection_impl.cc
```diff
@@ -71,6 +72,24 @@ bigtable::RowReader DataConnectionImpl::ReadRows(
   return MakeRowReader(std::move(impl));
 }
 
+std::vector<bigtable::FailedMutation> DataConnectionImpl::BulkApply(
+    std::string const& app_profile_id, std::string const& table_name,
+    bigtable::BulkMutation mut) {
+  if (mut.empty()) return {};
+  bigtable::internal::BulkMutator mutator(
+      app_profile_id, table_name, *idempotency_policy(), std::move(mut));
+  auto retry = retry_policy();
```
coryan, line 81: nit: this always allocate a retry and backoff policy, even though on success (the common case?) they are not used.  Most likely not important, but if you are looking for micro-optimizations, there is one.

coryan (approved):

google/cloud/bigtable/internal/data_connection_impl.cc
dbolduc: Done.

Issue #4512, opened 2020-07-14
Reduce data copies for Client::InsertObject


coryan:
The code path for InsertObject avoids copies using `std::move()`, and passing the request as a `const&`, but the default path uses JSON and includes checksums, so it ends here:

https://github.com/googleapis/google-cloud-cpp/blob/e235f46521219904f6886c36562d253ad5e852e3/google/cloud/storage/internal/curl_client.cc#L1291

Where we make some expensive copies. It would be more efficient to create multiple buffers and use `CurlRequest::MakeUploadRequest()`.

coryan:
I will try to look at this over the next few months.

coryan:
Hmmm. I am missing the comment, largely fixed by #9518

coryan:
Strike 3, closing.

PR #9158, opened 2022-06-03
refactor(storage): use multiple spans for insert object

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9158)


coryan (approved):
Let's tag #4512 in this PR.  I think it effectively fixes that problem for the `RestClient`, and that will be the default in the glorious future.

PR #9169, opened 2022-06-06
feat(common): include status error payloads in the RPC log

devbww:
Include the `google.rpc.Status.details` messages when adding a failing
`Status` value to the RPC log.

See https://cloud.google.com/apis/design/errors#error_payloads for the
details on what is available.

`Status` already has smarts for dealing with a `google.rpc.ErrorInfo`
message.  I was never sure why this support didn't extend to the other
error-detail messages, but @devjgm assured me that `ErrorInfo` was all
that was required for user access.

But let's at least add all the available information to the RPC log.
This will help us understand failures, like the notoriously inscrutable
"Invalid request", where we'll now be able to see which field was bad
and a description of what went wrong.

### DIFF
--- a/google/cloud/internal/log_wrapper.cc
+++ b/google/cloud/internal/log_wrapper.cc
@@ -14,8 +14,11 @@
 
 #include "google/cloud/internal/log_wrapper.h"
 #include "google/cloud/internal/absl_str_cat_quiet.h"
+#include "google/cloud/internal/status_payload_keys.h"
 #include <google/protobuf/text_format.h>
+#include <google/rpc/error_details.pb.h>
 #include <atomic>
+#include <sstream>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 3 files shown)


google/cloud/internal/log_wrapper.cc
```diff
@@ -36,6 +39,66 @@ std::string DebugString(google::protobuf::Message const& m,
...
+  auto payload =
+      internal::GetPayload(status, internal::kStatusPayloadGrpcProto);
+  google::rpc::Status proto;
+  if (payload && proto.ParseFromString(*payload)) {
+    // See https://cloud.google.com/apis/design/errors#error_payloads
+    for (google::protobuf::Any const& any : proto.details()) {
+      std::string details;
+      switch (status.code()) {
+        case StatusCode::kInvalidArgument:
+          details = DebugString<google::rpc::BadRequest>(any, options);
```
coryan, line 62: I wonder if we should do this based on the actual contents:

```cc
if (any.Is<google::rpc::BadRequest>()) {
  details = DebugString<google::rpc::BadRequest>(any, options);
} else if (any.Is<google::rpc::PreconditionFailure>()) {
  details = DebugString<google::rpc::PreconditionFailure>(any, options);
}
...
```

devbww: Unfortunately there appear to be multiple details protos in practice, although in the cases I've seen that the extra protos have been redundant.

For example, a "Not found" error has a `ResourceInfo` message (good), but also a `DebugInfo` message whose `.detail` field just contains the `.DebugString()` of the `ResourceInfo`.

That means in the proposal we would either need to be overly careful with the order of the `Is<M>()` tests, or change `else if` to `if` in that chain and allow for multiple messages in the log, neither of which I preferred.

So, I chose to follow the specification (at least for now), and only use the single details proto that matches the status code.  We could, of course, revisit that if we learn there might be additional, useful information in any extra messages.

coryan (approved):
SGTM, thanks for explaining.

Issue #9150, opened 2022-06-02
Timeout uploads on progress vs. completion


coryan:
This is analogous to #7300, but for uploads. These always involve streaming write RPCs.

PR #9195, opened 2022-06-07
impl(storage): helpers to create `OptionSpan`

coryan:
To add per-operation options we need some helpers to create a span from
the parameter pack received by each operation.  With these helpers we
will be able to more easily add spans. For example:

```cc
class Client {
 public:
...
  template <typename... RequestOptions>
  StatusOr<ObjectMetadata> GetObjectMetadata(
      std::string const& bucket_name, std::string const& object_name,
      RequestOptions&&... options) {
    auto const span = internal::MakeOptionSpan(
        options(), std::forward<RequestOptions>(options)...);
    ...
    ..
  }
};
```

Part of the work for #9150

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9195)

### DIFF
--- a/google/cloud/storage/internal/make_options_span.h
+++ b/google/cloud/storage/internal/make_options_span.h
@@ -1,0 +1,92 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_STORAGE_INTERNAL_MAKE_OPTIONS_SPAN_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_STORAGE_INTERNAL_MAKE_OPTIONS_SPAN_H
+
+#include "google/cloud/storage/version.h"
+#include "google/cloud/options.h"
+#include <utility>
+
+namespace google {
+namespace cloud {
+namespace storage {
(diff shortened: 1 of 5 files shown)


google/cloud/storage/internal/make_options_span_test.cc
```diff
@@ -0,0 +1,72 @@
...
+  auto const& current = CurrentOptions();
+  EXPECT_EQ("u-p-default", current.get<UserProjectOption>());
+  EXPECT_EQ("a-default", current.get<AuthorityOption>());
+}
+
+TEST(MakeOptionSpanTest, Overrides) {
+  auto const span =
+      MakeOptionsSpan(SimulateRawClientOptions(),
+                      Options{}.set<EndpointOption>("test-endpoint"),
+                      Options{}.set<AuthorityOption>("a-override-1"),
```
devbww, line 46: Perhaps trying an `Options{}.set<X>(x).set<Y>(y)` here would mix things up a little more?

google/cloud/storage/internal/make_options_span.h
```diff
@@ -0,0 +1,92 @@
...
+}
+
+template <typename Head, typename... Tail>
+static Options GroupOptions(Head&&, Tail&&... t) {
+  return GroupOptions(std::forward<Tail>(t)...);
+}
+///@}
+
+/**
+ * Create an option span for a `google::cloud::storage::Client` operation.
```
devbww, line 70: s/option span/`OptionsSpan`/ perhaps.

google/cloud/storage/internal/make_options_span.h
```diff
@@ -0,0 +1,92 @@
...
+namespace google {
+namespace cloud {
+namespace storage {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+///@{
+/**
+ * The `GroupOptions()` overload set groups all the `google::cloud::Options`
+ * present in a parameter pack into a single `google::cloud::Options`.
```
devbww, line 31: Perhaps mention that option values later in the pack are preferred when there are duplicates?

devbww (approved):

google/cloud/storage/internal/make_options_span.h
coryan: Done.

google/cloud/storage/internal/make_options_span.h
coryan: Done.

Issue #7691, opened 2021-12-02
add storage support for per-operation options


devbww:
Add an `Options options = {}` argument to the constructor of the client class(es).  Merge these options with the default options for the service, and store them as a member of the client class.

Add an `Options options = {}` argument to each operation within each client.  These options should then be merged with the client options from above, and installed as the prevailing options for the duration of the operation by instantiating an `internal::OptionsSpan`.

You could then use `internal::CurrentOptions()` to obtain (a `const&` to) the prevailing options from anywhere you might need them.  Any cleanup for call paths where `Options` have been passed explicitly is discretionary.

Similar support for the generated client classes was added in #7683, so you might be able to use that as an example.

----

I (@coryan) will steal some of the description to break down the work a bit. 

- [x] Change all the `storage::internal::*Request` classes to ignore `google::cloud::Options` in `set_multiple_options()`
- [x] Create some helpers to convert parameter packs into a `google::cloud::Options` and to create spans 
- [x] Change all the member functions in `storage::Client` to create an `OptionsSpan`
- [x] Add tests to verify all these spans are created
- [x] Change `storage::internal::CurlClient` to use `google::cloud::internal::CurrentOptions()` 
- [x] Change `storage::internal::RetryClient` to use `google::cloud::internal::CurrentOptions()`
- [x] Change `storage::internal::GrpcClient` to stop creating its own spans
- [x] Change `storage::internal::RestClient` to use `google::cloud::internal::CurrentOptions()`
- [ ] Update the documentation in `storage::Client`

devbww:
Note: The options stored in the client class should now include the `connection->options()`.  So something like ...
```
ServiceClient::ServiceClient(std::shared_ptr<ServiceConnection> connection, Options opts = {})
    : connection_(std::move(connection)),
      opts_(internal::MergeOptions(std::move(opts), 
                                   service_internal::DefaultOptions(connection_->options()))) {
  ...
}

T ServiceClient::Operation(..., Options opts = {}) {
  internal::OptionsSpan span(internal::MergeOptions(std::move(opts), opts_));
  ...
}
```

coryan:
## Background

All operations in the storage library already have a parameter pack:

https://github.com/googleapis/google-cloud-cpp/blob/5cb53865c0e9b113bd38fbdfb7ab8268bffc415e/google/cloud/storage/client.h#L1114-L1117

There are several alternatives to add `google::cloud::Options` to this set.

[...]

PR #9202, opened 2022-06-08
feat(storage): per-operation options / BucketAccessControl

coryan:
Support per-operation `google::cloud::Options` for operations related to
`BucketAccessControl` resources.

Part of the work for #7691

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9202)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -1675,6 +1675,7 @@ class Client {
   template <typename... Options>
   StatusOr<std::vector<BucketAccessControl>> ListBucketAcl(
       std::string const& bucket_name, Options&&... options) {
+    auto const span = MakeSpan(std::forward<Options>(options)...);
     internal::ListBucketAclRequest request(bucket_name);
     request.set_multiple_options(std::forward<Options>(options)...);
     auto items = raw_client_->ListBucketAcl(request);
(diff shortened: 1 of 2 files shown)


dbolduc (approved):

google/cloud/storage/client.h
```diff
@@ -1675,6 +1675,7 @@ class Client {
   template <typename... Options>
   StatusOr<std::vector<BucketAccessControl>> ListBucketAcl(
       std::string const& bucket_name, Options&&... options) {
+    auto const span = MakeSpan(std::forward<Options>(options)...);
```
devbww, line 1678: For the record, I'm a little concerned about returning `OptionSpan`s.  It seems to depend on copy elision to do the right thing, but I'm still thinking it through.  Remember that the `OptionSpan` ctor and dtor write to a (thread) global, and its set/restore operations need to be FIFO.  Perhaps `OptionSpan` should not be copyable/movable.

A safer tactic would be to have the pack consumer just return an `Options`, and then have these functions instantiate the `OptionsSpan` directly.

coryan: Fair enough.  I would prefer to merge the current PRs and then make one PR fixing this.  Is that Okay with you?  Otherwise I will pull them and create one large PR, redoing the work with a few PRs already merged is more convoluted than I wish.



nit: I think you meant LIFO.

devbww: > Fair enough. I would prefer to merge the current PRs and then make one PR fixing this. Is that Okay with you?

Of course.  Whatever is easiest.

> nit: I think you meant LIFO.

Indeed.

PR #9216, opened 2022-06-08
cleanup: consistency

dbolduc:
I dont think there is a difference either way, but we tend to write `auto stub = stub_;`. I was likely the one who introduced this inconsistency.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9216)

### DIFF
--- a/generator/internal/connection_impl_generator.cc
+++ b/generator/internal/connection_impl_generator.cc
@@ -454,7 +454,7 @@ future<Status>)"""
 future<StatusOr<$response_type$>>)""",
                       R"""(
 $connection_class_name$Impl::Async$method_name$($request_type$ const& request) {
-  auto& stub = stub_;
+  auto stub = stub_;
   return google::cloud::internal::AsyncRetryLoop(
       retry_policy(), backoff_policy(),
       idempotency_policy()->$method_name$(request),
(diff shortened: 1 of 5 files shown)


generator/internal/connection_impl_generator.cc
```diff
@@ -454,7 +454,7 @@ future<Status>)"""
 future<StatusOr<$response_type$>>)""",
                       R"""(
 $connection_class_name$Impl::Async$method_name$($request_type$ const& request) {
-  auto& stub = stub_;
+  auto stub = stub_;
```
devbww, line 457: I'm not sure that your way isn't better.

If we assume that `stub_` has an initial ref count of 1, then the new, copy-into-`stub` way will increase that to 2.  Then, the copy into the functor will increase it to 3.  Finally, when `stub` is destroyed as this function returns, the ref count falls back to 2: one for `stub_` and the other for the stub in the retry functor.

In the old way, `stub` doesn't take a new reference ... it just gives the `stub_` member another name.  Copying that into the functor will increase the ref count to 2, but "destroying" `stub` does nothing.  So, we got to the desired state without going to 3 and back.  Sounds better to me.

dbolduc:
Closing. I will make things consistent, but the way @devbww says (saving a copy of the `shared_ptr`)

PR #9217, opened 2022-06-08
impl(bigtable): modern `Table::AsyncApply()`

dbolduc:
Part of the work for #8799

Basically a copy of `Table::Apply()`. Especially the unit tests. (I did make the `Apply()` unit tests slightly more rigorous by using mock policies instead of relying on the behavior of the default policies).

Also introduce test params to the async test fixture, exactly as we did in `data_integration_test.cc`. (Yes, I meant to change this one's name to `DataIntegrationTest`):

https://github.com/googleapis/google-cloud-cpp/blob/0d01585a8fb44a8990188705c073a01d9234ed29/google/cloud/bigtable/tests/data_integration_test.cc#L38-L44


google/cloud/bigtable/internal/data_connection_impl.cc
```diff
@@ -82,6 +83,40 @@ Status DataConnectionImpl::Apply(std::string const& app_profile_id,
...
+  mut.MoveTo(request);
+
+  auto idempotent_policy = idempotency_policy();
+  bool const is_idempotent = std::all_of(
+      request.mutations().begin(), request.mutations().end(),
+      [&idempotent_policy](google::bigtable::v2::Mutation const& m) {
+        return idempotent_policy->is_idempotent(m);
+      });
+
+  auto stub = stub_;
```
devbww, line 101: Were you convinced by my argument to prefer a reference?

coryan (commented):
What @devbww says.

dbolduc:
I am merging this PR as is. I opened #9223 and assigned it to myself. I will fix it before adding the other async functions.

PR #9220, opened 2022-06-08
feat(storage): per-operation options for `CurlClient`

coryan:
Use the options from `CurrentOptions()` to control the behavior of
`CurlClient`.  In the implementation of some long-lived objects
(streaming writes, streaming reads), we need to copy the options used
when the object is created, and then set the span before making RPCs. I
tried to make this copy as soon as possible.

Some tests bypass `storage::Client` and need to be adjusted to create a
span. Ditto for some standalone functions, like parallel uploads. There
is an existing bug to add better support for per-operation options for
the standalone functions too.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9220)

### DIFF
--- a/google/cloud/storage/internal/curl_client.cc
+++ b/google/cloud/storage/internal/curl_client.cc
@@ -126,15 +126,14 @@ std::string HostHeader(Options const& options, char const* service) {
 
 Status CurlClient::SetupBuilderCommon(CurlRequestBuilder& builder,
                                       char const* method, char const* service) {
+  auto const& current = CurrentOptions();
   auto auth_header =
-      CurrentOptions().get<Oauth2CredentialsOption>()->AuthorizationHeader();
-  if (!auth_header.ok()) {
-    return std::move(auth_header).status();
-  }
+      current.get<Oauth2CredentialsOption>()->AuthorizationHeader();
+  if (!auth_header.ok()) return std::move(auth_header).status();
   builder.SetMethod(method)
-      .ApplyClientOptions(CurrentOptions())
+      .ApplyClientOptions(current)
       .AddHeader(auth_header.value())
-      .AddHeader(HostHeader(CurrentOptions(), service))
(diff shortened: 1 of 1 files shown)


google/cloud/storage/internal/curl_client.cc
```diff
@@ -124,14 +127,14 @@ std::string HostHeader(Options const& options, char const* service) {
...
+      CurrentOptions().get<Oauth2CredentialsOption>()->AuthorizationHeader();
   if (!auth_header.ok()) {
     return std::move(auth_header).status();
   }
   builder.SetMethod(method)
-      .ApplyClientOptions(opts_)
+      .ApplyClientOptions(CurrentOptions())
       .AddHeader(auth_header.value())
-      .AddHeader(HostHeader(opts_, service))
+      .AddHeader(HostHeader(CurrentOptions(), service))
```
devbww, line 137: It is probably worth doing `auto const& options = CurrentOptions();` once as finding the options is like doing a map lookup by tid.

devbww (approved):

google/cloud/storage/internal/curl_client.cc
coryan: > It is probably worth doing `auto const& options = CurrentOptions();` once

Done.

> as finding the options is like doing a map lookup by tid.

That would be a really sad implementation of thread local storage.  My recollection is that there is a register with the thread context, and some googling seems to confirm that:

https://chao-tic.github.io/blog/2018/12/25/tls

Once your find the control block for the thread, all the TLS variables are in per-thread array, and should be an indirection (or two or three) into that register.

devbww: Yes ... right you are.  I'm a Neanderthal.  Certainly lessens the need to factor `CurrentOptions()`, but still not a bad thing.

PR #9229, opened 2022-06-08
impl(bigtable): modern `AsyncRowSampler`

dbolduc:
Most of the work for #9173 

Convert `AsyncRowSampler` to use the Stub and common retry/backoff policies.

The diff on the unit test came out pretty rough :(

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9229)


coryan (approved):
Yeah, it was easier to just read the test as if it was new code.

PR #9230, opened 2022-06-08
impl(storage): use `CurrentOptions()` for `RetryClient`

coryan:
The most interesting options in this case are the `RetryPolicyOption`
and `BackoffPolicyOption`.

Because now the policies are obtained from the `CurrentOptions()` and
these are initialized from `mock->options()`, the mocks need to return
sensible values for the options.

Some tests need to set the OptionsSpan, or the unit under test will not
work.

Part of the work for #7691

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9230)

### DIFF
--- a/google/cloud/storage/client.cc
+++ b/google/cloud/storage/client.cc
@@ -51,7 +51,7 @@ std::shared_ptr<internal::RawClient> Client::CreateDefaultInternalClient(
   if (enable_logging) {
     client = std::make_shared<internal::LoggingClient>(std::move(client));
   }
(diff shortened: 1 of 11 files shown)


scotthart (approved):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9230)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #9234, opened 2022-06-09
impl(bigtable): add constructors for RowKeySample

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9234)

### DIFF
--- a/google/cloud/bigtable/row_key_sample.h
+++ b/google/cloud/bigtable/row_key_sample.h
@@ -25,6 +25,9 @@ namespace bigtable {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 /// A simple wrapper to represent the response from `Table::SampleRowKeys()`.
 struct RowKeySample {
+  RowKeySample() = default;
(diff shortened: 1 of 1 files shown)


dbolduc:
Hmm so instead of adding a constructor, can we instead change the test?

So like this:
https://github.com/googleapis/google-cloud-cpp/blob/afdb4f552e6386f82464035b790791a3c8b170fd/google/cloud/bigtable/internal/async_row_sampler_test.cc#L60-L66

becomes:
```c++
absl::optional<v2::SampleRowKeysResponse> MakeResponse(std::string row_key,
                                                       std::int64_t offset) {
 v2::SampleRowKeysResponse r;
  r.set_row_key(std::move(row_key));
  r.set_offset_bytes(offset);
  return absl::make_optional(r);
};
```

then we drop all the initializer lists

devbww:
+1

dbolduc:
@scotthart Instead of making you copy/paste the code in my comment, I opened #9235

PR #9236, opened 2022-06-09
impl(rest): add support for treating specified error codes as non-failures

scotthart:
Need to support GCS using 308 for upload resume, but in a generic fashion.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9236)

### DIFF
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -628,11 +628,11 @@ StatusOr<std::size_t> CurlImpl::ReadImpl(absl::Span<char> output) {
     TRACE_STATE() << ", status=" << status << ", http code=" << http_code_
                   << "\n";
 
-    if (!internal::Contains(ignored_http_error_codes_, http_code_) &&
-        !status.ok()) {
-      return status;
+    if (status.ok() ||
+        internal::Contains(ignored_http_error_codes_, http_code_)) {
+      return bytes_read;
     }
-    return bytes_read;
+    return status;
   }
   TRACE_STATE() << ", http code=" << http_code_ << "\n";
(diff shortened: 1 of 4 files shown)


google/cloud/internal/curl_rest_client_integration_test.cc
```diff
@@ -146,6 +148,31 @@ TEST_F(RestClientIntegrationTest, Get) {
   EXPECT_GT(body->size(), 0);
 }
 
+TEST_F(RestClientIntegrationTest, GetWithIgnoredErrorCode) {
+  options_.set<UnifiedCredentialsOption>(MakeInsecureCredentials());
+  auto client =
+      MakeDefaultRestClient(url_, Options{}.set<IgnoredHttpErrorCodes>({308}));
```
coryan, line 154: neat.

google/cloud/internal/curl_impl.cc
```diff
@@ -626,7 +627,11 @@ StatusOr<std::size_t> CurlImpl::ReadImpl(absl::Span<char> output) {
         static_cast<HttpStatusCode>(http_code_), {});
     TRACE_STATE() << ", status=" << status << ", http code=" << http_code_
                   << "\n";
-    if (!status.ok()) return status;
+
+    if (!internal::Contains(ignored_http_error_codes_, http_code_) &&
```
coryan, line 631: I think it might be easier to revert all the `not`s at this point:

```cc
  if (status.ok()) return bytes_read;
  if (internal::Contains(ignored_http_error_codes_, http_code_)) return bytes_read;
  return status;
```

coryan (approved):
A small readability thing that I leave up to you.

PR #9247, opened 2022-06-10
doc(storage): document per-operation options

### LINKED ISSUE #7691: add storage support for per-operation options

coryan:
Fixes #7691

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9247)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -36,6 +36,7 @@
 #include "google/cloud/storage/upload_options.h"
 #include "google/cloud/storage/version.h"
 #include "google/cloud/internal/throw_delegate.h"
+#include "google/cloud/options.h"
 #include "google/cloud/status.h"
 #include "google/cloud/status_or.h"
 #include "absl/meta/type_traits.h"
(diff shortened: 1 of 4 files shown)


google/cloud/storage/client.h
```diff
@@ -195,24 +196,32 @@ struct ClientImplDetails;
...
+ * modify the request, you can specify options that override the local behavior
+ * of the library.  For example, you can override the local retry policy:
+ *
+ * @snippet storage_client_per_operation_samples.cc change-retry-policy
+ *
  * @par Retry, Backoff, and Idempotency Policies
  *
  * The library automatically retries requests that fail with transient errors,
- * and follows the
+ * follows the
```
devbww, line 224: "and follows" sounded better.

google/cloud/storage/client.h
```diff
@@ -160,28 +161,28 @@ struct ClientImplDetails;
...
  * auto stream = gcs.ReadObject(
  *     "my-bucket", "my-object", gcs::Generation(generation));
  * @endcode
  *
- * Each function documents the types accepted as optional parameters. These
- * options can be specified in any order. Specifying an option that is not
- * applicable to a member function results in a compile-time error.
+ * Each function documents the types accepted as optional request options. These
+ * parameters can be specified in any order. Specifying a request option that is
+ * not applicable to a member function results in a compile-time error.
```
devbww, line 183: Is this true?  If so, I'm not sure it is a feature.

coryan: We can have a discussion about whether we should change this, but this is not new behavior. Note that this is referring to the static options, so something like this would fail to compile:

```cc
  client.GetBucketMetadata("my-bucket",
     gcs::Generation(7) // compile-time error, buckets do not have generations
 );
```

while this would compile (and this is new behavior):

```cc
  client.GetBucketMetadata("my-bucket",
    g::Options{}.set<pubsub::RetryPolicyOption(...) // ignored
  );
```

devbww: That's cool.  I was only thinking about the behavior for `Options`.

devbww (approved):

PR #9251, opened 2022-06-10
fix: relax parsing for GCE informational fields

coryan:
We parse (but largely ignore) some metadata information about the GCE
service account. This parsing was very strict, rejecting the data if
fields were missing.  We now accept missing fields, and field them
with blank data (or empty lists).  We also accept the responses from
Cloud Shell, where the `scopes` field is a newline separated string,
instead of an array of strings.  This may be a bug in Cloud Shell, but
it makes the library unusable.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9251)

### DIFF
--- a/google/cloud/internal/oauth2_compute_engine_credentials.cc
+++ b/google/cloud/internal/oauth2_compute_engine_credentials.cc
@@ -15,10 +15,10 @@
 #include "google/cloud/internal/oauth2_compute_engine_credentials.h"
 #include "google/cloud/internal/compute_engine_util.h"
 #include "google/cloud/internal/curl_options.h"
-#include "google/cloud/internal/getenv.h"
 #include "google/cloud/internal/oauth2_credential_constants.h"
 #include "google/cloud/internal/oauth2_credentials.h"
 #include "google/cloud/internal/rest_response.h"
+#include "absl/strings/str_split.h"
 #include <nlohmann/json.hpp>
 #include <set>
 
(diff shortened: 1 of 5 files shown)


google/cloud/storage/oauth2/compute_engine_credentials.cc
```diff
@@ -13,36 +13,20 @@
...
 #include <nlohmann/json.hpp>
 
 namespace google {
 namespace cloud {
 namespace storage {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace oauth2 {
+
 StatusOr<ServiceAccountMetadata> ParseMetadataServerResponse(
     storage::internal::HttpResponse const& response) {
```
dbolduc, line 26: so the common REST library has code like:
https://github.com/googleapis/google-cloud-cpp/blob/4ab2d009a738726dcde670b7e6d40f3480d8528c/google/cloud/internal/oauth2_compute_engine_credentials.cc#L32-L33

Has this sort of processing already happened?

dbolduc (approved):

google/cloud/storage/oauth2/compute_engine_credentials.cc
coryan: Yes.

google/cloud/internal/oauth2_compute_engine_credentials.cc
```diff
@@ -31,25 +31,31 @@ StatusOr<ServiceAccountMetadata> ParseMetadataServerResponse(
...
+  auto scopes = [&]() -> std::set<std::string> {
+    if (!body.contains("scopes")) return {};
+    auto const& s = body["scopes"];
+    if (s.is_string()) {
+      return absl::StrSplit(s.get<std::string>(), '\n', absl::SkipWhitespace());
+    }
+    if (!s.is_array()) return {};
+    std::set<std::string> result;
+    for (auto const& i : s) {
+      if (!i.is_string()) return {};
```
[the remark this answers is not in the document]
coryan: Maybe?  You could argue that "if we cannot parse `scopes` as an array of strings we return an empty list of scopes" is a reasonable action to take.  I added a comment to explain that.

PR #9252, opened 2022-06-10
fix: handle transient failures in the GCE credentials

### LINKED ISSUE #9249: `GoogleDefaultCredentials()` does not handle temporary unavailability in metadata server
This code runs only once, when the credentials are created:

https://github.com/googleapis/google-cloud-cpp/blob/5d88584a32ea0ada7c212c96beb08634ee8637f4/google/cloud/internal/oauth2_google_credentials.cc#L158
https://github.com/googleapis/google-cloud-cpp/blob/5d88584a32ea0ada7c212c96beb08634ee8637f4/google/cloud/storage/oauth2/google_credentials.cc#L169

Basically this is querying the metadata server to verify that the application is running in an environment that supports "GCE" [*] credentials.  But the metadata server could be temporarily unavailable, maybe it had too many requests, maybe it was just slow, or restarting.  In any case, we should return something that retries requests later.

[*] Not only GCE supports a metadata server, GAE, GKE, Cloud Run, Cloud Shell, all have metadata servers and thus support GCE credentials.

coryan:
Getting the authorization token from the GCE credentials can fail
when the credentials are created.  The metadata service may be
experiencing some kind of transient failure.  It is better to return the
credentials, and let the retry loops handle the problem.  If the failure
is not transient, the retry loop will handle that too.

Fixes #9249

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9252)

### DIFF
--- a/google/cloud/internal/oauth2_google_credentials.cc
+++ b/google/cloud/internal/oauth2_google_credentials.cc
@@ -34,10 +34,6 @@ namespace oauth2_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace {
 
-char constexpr kAdcLink[] =
-    "https://developers.google.com/identity/protocols/"
-    "application-default-credentials";
-
 // Parses the JSON at `path` and creates the appropriate
 // Credentials type.
 //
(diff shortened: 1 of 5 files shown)


scotthart (approved):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9252)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #9255, opened 2022-06-11
fix(storage): respect policies in deprecated constructor

dbolduc:
This deprecated constructor
https://github.com/googleapis/google-cloud-cpp/blob/2c83a865b680bb7c498b00770c3bbf5cfb0480a4/google/cloud/storage/client.h#L3236-L3244

... consolidates the retry/backoff/idempotency policies as options
https://github.com/googleapis/google-cloud-cpp/blob/2c83a865b680bb7c498b00770c3bbf5cfb0480a4/google/cloud/storage/client_options.h#L51-L55

... but when the options get to the `RetryClient`, only the logging configuration is used.
https://github.com/googleapis/google-cloud-cpp/blob/2c83a865b680bb7c498b00770c3bbf5cfb0480a4/google/cloud/storage/client.cc#L45-L58

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9255)

### DIFF
--- a/google/cloud/storage/internal/retry_client.cc
+++ b/google/cloud/storage/internal/retry_client.cc
@@ -94,19 +94,22 @@ typename Signature<MemberFunction>::ReturnType MakeCall(
 }  // namespace
 
 std::shared_ptr<RetryClient> RetryClient::Create(
-    std::shared_ptr<RawClient> client) {
+    std::shared_ptr<RawClient> client, Options policies) {
   // Cannot use `std::make_shared<>` because the constructor is private.
-  return std::shared_ptr<RetryClient>(new RetryClient(std::move(client)));
+  return std::shared_ptr<RetryClient>(
+      new RetryClient(std::move(client), std::move(policies)));
 }
 
-RetryClient::RetryClient(std::shared_ptr<RawClient> client)
-    : client_(std::move(client)) {}
+RetryClient::RetryClient(std::shared_ptr<RawClient> client, Options policies)
+    : client_(std::move(client)), policies_(std::move(policies)) {}
 
 ClientOptions const& RetryClient::client_options() const {
   return client_->client_options();
 }
 
-Options RetryClient::options() const { return client_->options(); }
+Options RetryClient::options() const {
+  return google::cloud::internal::MergeOptions(policies_, client_->options());
+}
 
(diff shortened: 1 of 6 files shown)


google/cloud/storage/internal/retry_client.cc
```diff
@@ -94,19 +94,22 @@ typename Signature<MemberFunction>::ReturnType MakeCall(
...
+RetryClient::RetryClient(std::shared_ptr<RawClient> client, Options policies)
+    : client_(std::move(client)), policies_(std::move(policies)) {}
 
 ClientOptions const& RetryClient::client_options() const {
   return client_->client_options();
 }
 
-Options RetryClient::options() const { return client_->options(); }
+Options RetryClient::options() const {
+  return google::cloud::internal::MergeOptions(policies_, client_->options());
```
dbolduc, line 111: I'm wondering if this is better off as:
```c++
Options RetryClient::options() const {
  auto opts = client_->options();
  if (policies_.has<RetryPolicyOption>()) {
    opts.set<RetryPolicyOption>(policies_.get<RetryPolicyOption>());
  }
  if (policies_.has<BackoffPolicyOption>()) {
    opts.set<BackoffPolicyOption>(policies_.get<BackoffPolicyOption>());
  }
  if (policies_.has<IdempotencyPolicyOption>()) {
    opts.set<IdempotencyPolicyOption>(policies_.get<IdempotencyPolicyOption>());
  }
  return opts;
}
```
given that there are only 3 possible policy options, but N client options.

coryan: We should optimize `MergeOptions()` for the case when `policies_` is empty.  I think that is a fairly common case (and *the* common case here), and we can turn that function from O(n) to O(1).

coryan (approved):

google/cloud/storage/internal/retry_client.cc
dbolduc: @coryan can we assume that `client_->options()` doesn't change from when the class is constructed? If that's true we can do the merge on construction, and just return that result every time `options()` is called.

coryan: I think you can make that assumption.  It is true for all concrete implementations of the `RawClient` class except `MockClient`, and we can say that `MockClient` needs to respect that contract.

Issue #9133, opened 2022-06-01
Implement AsyncClient::ReadObject()


coryan:
This should just read a range, and return all the bytes, i.e, the API should be something like:

```cc
struct ReadRangeResponse {
  absl::optional<ObjectMetadata> object_metadata;
  std::multimap<std::string, std::string> request_metadata; // aka headers
  std::vector<std::string> payload; // the actual data
};

class AsyncClient {
 public:
  template <typename... RequestOptions>
  future<StatusOr<ReadRangeResponse>>
  ReadObject(std::string bucket_name, std::string object_name, RequestOptions&&... options);
```

PR #9277, opened 2022-06-15
impl(GCS+gRPC): async accumulator for `ReadObject()`

coryan:
The asynchronous version of `ReadObject()` will return all the bytes. We
need a function to asynchronously accumulate all the values from a
single `AsyncReadObject()` call to the `StorageStub`. This function does
to retry or resume the download, that will come later. In case you are
wonder, we will require applications to only make "ranged reads", where
the number of bytes downloaded is controlled by the application.

Part of the work for #9133

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9277)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9277)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #9278, opened 2022-06-15
impl(storage): use options to select REST implementation

coryan:
Use a new (internal-only option) to select between `RestClient` and
`CurlClient`. This will make it easier to write benchmarks that compare
these implementations side-by-side.  To support the CI builds, the
environment variable (`GOOGLE_CLOUD_CPP_STORAGE_HAVE_REST_CLIENT`) sets
the default value for the new option.

### DIFF
--- a/google/cloud/storage/internal/rest_client.cc
+++ b/google/cloud/storage/internal/rest_client.cc
@@ -891,453 +891,147 @@ StatusOr<QueryResumableUploadResponse> RestClient::UploadChunk(
 
 StatusOr<ListBucketAclResponse> RestClient::ListBucketAcl(
     ListBucketAclRequest const& request) {
-  auto const& current = CurrentOptions();
-  RestRequestBuilder builder(
-      absl::StrCat("storage/", current.get<TargetApiVersionOption>(), "/b/",
-                   request.bucket_name(), "/acl"));
-  auto auth = AddAuthorizationHeader(current, builder);
-  if (!auth.ok()) return auth;
-  request.AddOptionsToHttpRequest(builder);
-  return ParseFromRestResponse<ListBucketAclResponse>(
-      storage_rest_client_->Get(std::move(builder).BuildRequest()));
+  return curl_client_->ListBucketAcl(request);
 }
 
 StatusOr<BucketAccessControl> RestClient::GetBucketAcl(
     GetBucketAclRequest const& request) {
-  auto const& current = CurrentOptions();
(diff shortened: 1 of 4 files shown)


google/cloud/storage/internal/rest_client.cc
```diff
@@ -891,453 +891,147 @@ StatusOr<QueryResumableUploadResponse> RestClient::UploadChunk(
...
-  auto const& current = CurrentOptions();
-  RestRequestBuilder builder(
-      absl::StrCat("storage/", current.get<TargetApiVersionOption>(), "/b/",
-                   request.bucket_name(), "/acl"));
-  auto auth = AddAuthorizationHeader(current, builder);
-  if (!auth.ok()) return auth;
-  request.AddOptionsToHttpRequest(builder);
-  return ParseFromRestResponse<ListBucketAclResponse>(
-      storage_rest_client_->Get(std::move(builder).BuildRequest()));
+  return curl_client_->ListBucketAcl(request);
```
dbolduc, line 894: I do not think @scotthart will appreciate this change

google/cloud/storage/options.h
```diff
@@ -64,6 +65,11 @@ struct CAPathOption {
   using Type = std::string;
 };
 
+/// This is only intended for testing of the library. Not for public use.
+struct RestConfiguration {
+  using Type = std::string;
```
devbww, line 70: Q.  I see this is set-to/checked-as "rest", which seems redundant ... RestConfiguration == "rest".  Perhaps it should be "TransportOption", or something like that?

google/cloud/storage/internal/rest_client.cc
coryan: Ooops, bad merge, fixed.

dbolduc (approved):

google/cloud/storage/options.h
```diff
@@ -64,6 +65,11 @@ struct CAPathOption {
   using Type = std::string;
 };
 
+/// This is only intended for testing of the library. Not for public use.
+struct RestConfiguration {
```
devbww, line 69: Nit: These have traditionally been called "....Option".

google/cloud/storage/options.h
coryan: Turned this into a boolean.

google/cloud/storage/options.h
coryan: Done

PR #9291, opened 2022-06-16
fix(GCS+gRPC): all `co_await` should be callbacks

coryan:
I tried to cheat for the timers and avoid a callback, but it produces
deadlocks if the completion queue has a single thread.

Part of the work for #9133

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9291)

### DIFF
--- a/google/cloud/storage/internal/async_accumulate_read_object.cc
+++ b/google/cloud/storage/internal/async_accumulate_read_object.cc
@@ -128,7 +128,6 @@ class AsyncAccumulateReadObjectPartialHandle
         self->OnReadTimer(f.get(), std::move(response));
       }
     };
-    auto self = shared_from_this();
     tm.then(ByMove{shared_from_this(), std::move(response)});
   }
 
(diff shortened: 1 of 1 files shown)


google/cloud/storage/internal/async_accumulate_read_object.cc
```diff
@@ -114,7 +121,19 @@ class AsyncAccumulateReadObjectPartialHandle
...
     tm.cancel();
-    if (tm.get()) return OnTimeout("Read()");
+    struct ByMove {
+      std::shared_ptr<AsyncAccumulateReadObjectPartialHandle> self;
+      absl::optional<Response> response;
+      void operator()(future<bool> f) {
+        self->OnReadTimer(f.get(), std::move(response));
+      }
+    };
+    auto self = shared_from_this();
```
dbolduc, line 131: nit: unnecessary line

dbolduc (approved):

google/cloud/storage/internal/async_accumulate_read_object.cc
coryan: Thanks, fixed.

PR #9302, opened 2022-06-17
impl(GCS+gRPC): convert `AsyncReadObject()` results

coryan:
In the `storage` library we do not expose protos directly to the
application developer. This change introduces a function to convert the
protos accumulated by `AsyncAccumulateReadObjectFull()` into types that
are consistent with the public APIs.

Part for the work for #9133

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9302)

### DIFF
--- a/google/cloud/storage/async_object_responses.h
+++ b/google/cloud/storage/async_object_responses.h
@@ -29,7 +29,7 @@ namespace storage_experimental {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 /// Represents the response from reading a subset of an object.
-struct ReadObjectRangeResponse {
+struct AsyncReadObjectRangeResponse {
   /**
    * The final status of the download.
    *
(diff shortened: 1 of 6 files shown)


google/cloud/storage/async_object_responses.h
```diff
@@ -0,0 +1,77 @@
...
+#include <string>
+#include <vector>
+
+namespace google {
+namespace cloud {
+namespace storage_experimental {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/// Represents the response from reading a subset of an object.
+struct ReadObjectRangeResponse {
```
dbolduc, line 32: Does this have any relation to `internal::ReadObjectRangeRequest`? If no, are you concerned that a future developer will trip on these names?

google/cloud/storage/async_object_responses.h
```diff
@@ -0,0 +1,77 @@
...
+   *     response.contents.begin(), response.contents.end(), std::string{},
+   *     [](auto a, auto b) { a += b; return a; });
+   * @endcode
+   */
+  std::vector<std::string> contents;
+
+  /**
+   * Per-request metadata and annotations.
+   *
+   * These are intended as debugging tools, they are subject to change without
```
dbolduc, line 66: s/, they/. They/

dbolduc (approved):

google/cloud/storage/async_object_responses.h
coryan: Fixed.

google/cloud/storage/async_object_responses.h
coryan: Hmmm... I added an `Async` prefix to avoid that.

Issue #7688, opened 2021-12-02
add bigtable support for per-operation options


devbww:
Add an `Options options = {}` argument to the constructor of the client class(es).  Merge these options with the default options for the service, and store them as a member of the client class.

Add an `Options options = {}` argument to each operation within each client.  These options should then be merged with the client options from above, and installed as the prevailing options for the duration of the operation by instantiating an `internal::OptionsSpan`.

You could then use `internal::CurrentOptions()` to obtain (a `const&` to) the prevailing options from anywhere you might need them.  Any cleanup for call paths where `Options` have been passed explicitly is discretionary.

Similar support for the generated client classes was added in #7683, so you might be able to use that as an example.

devbww:
Note: The options stored in the client class should now include the `connection->options()`.  So something like ...
```
ServiceClient::ServiceClient(std::shared_ptr<ServiceConnection> connection, Options opts = {})
    : connection_(std::move(connection)),
      opts_(internal::MergeOptions(std::move(opts), 
                                   service_internal::DefaultOptions(connection_->options()))) {
  ...
}

T ServiceClient::Operation(..., Options opts = {}) {
  internal::OptionsSpan span(internal::MergeOptions(std::move(opts), opts_));
  ...
}
```

devbww:
Addendum: Please also ensure that the `OptionsSpan` instantiated during `ServiceClient::Operation()` is saved and then restored over any work done on behalf of the operation but after `Operation()` returns.  Typically this means for async, paginated, or streaming RPCs.  (These cases may already be handled via changes in the common library.)

PR #9307, opened 2022-06-19
impl(bigtable): `Table` accepts `Options`

dbolduc:
Part of the work for #7688

I do not think it makes sense to add Options to the constructors that accept a `DataClient`, because no options will apply to the calls.

The constructor is lacking documentation. I will add it later.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9307)

### DIFF
--- a/google/cloud/bigtable/table.cc
+++ b/google/cloud/bigtable/table.cc
@@ -48,6 +48,7 @@ static_assert(std::is_copy_assignable<bigtable::Table>::value,
 
 Status Table::Apply(SingleRowMutation mut) {
   if (connection_) {
+    google::cloud::internal::OptionsSpan span(options_);
     return connection_->Apply(app_profile_id_, table_name_, std::move(mut));
   }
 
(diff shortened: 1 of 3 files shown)


devbww:
> I cannot find the issue for adding client-level options for the Bigtable Data API

Maybe #7688??

dbolduc:
> Maybe #7688??

D'oh. I didn't read past the title, and I assumed the scope was only call options. Not client options + call options. Thanks.

google/cloud/bigtable/table.cc
```diff
@@ -48,6 +48,7 @@ static_assert(std::is_copy_assignable<bigtable::Table>::value,
 
 Status Table::Apply(SingleRowMutation mut) {
   if (connection_) {
+    google::cloud::internal::OptionsSpan span(options_);
```
devbww, line 51: Could/should these be outside the `if (connection_)` conditional?

dbolduc: The spans could be outside the if conditionals, but there is no point. The code outside of the if conditionals does not care about `CurrentOptions()`. I'd rather have them inside the if blocks.

devbww: > The code outside of the if conditionals does not care about `CurrentOptions()`.

OK, although I'll assert that it is difficult to know that, and it may be better to be safe than sorry.

devbww (approved):

PR #9272, opened 2022-06-13
refactor(bigtable): untemplate `AsyncRowReader`

dbolduc:
Part of the work for #9174 

Our virtual connection base class cannot have templates. So we need to convert the arguments passed into `Table::AsyncReadRows(...)`.

I could use some help on whether the arguments to the `std::function` should be `const&` qualified or not:
```c++
// e.g. :
using RowFunctor = std::function<future<bool>(bigtable::Row const&)>;
using FinishFunctor = std::function<void(Status const&)>;
```

For the record, we `std::move` the row into the RowFunctor: https://github.com/dbolduc/google-cloud-cpp/blob/eba7d22730fd530caff55cb7469e54376b99e446/google/cloud/bigtable/internal/async_row_reader.cc#L106

but we do not `std::move` the status into the FinishFunctor:
https://github.com/dbolduc/google-cloud-cpp/blob/eba7d22730fd530caff55cb7469e54376b99e446/google/cloud/bigtable/internal/async_row_reader.cc#L83

does this imply that I should be doing:
```c++
using RowFunctor = std::function<future<bool>(bigtable::Row)>;
using FinishFunctor = std::function<void(Status const&)>;
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9272)

PR #9313, opened 2022-06-20
refactor(bigtable): actually type erase `AsyncReadRows` this time

dbolduc:
Type erase the `RowFunctor` and `FinishFunctor` templates for `Table::AsyncReadRows`.

We define aliases for the two types. (I personally think it's nicer to have aliases vs. saying `std::function<...>`). Then we move the no-longer-templated implementation from `table.h` to `table.cc`, dropping the `static_assert`s.

I should have done this in #9272.

### DIFF
--- a/google/cloud/bigtable/table.h
+++ b/google/cloud/bigtable/table.h
@@ -15,6 +15,7 @@
 #ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGTABLE_TABLE_H
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGTABLE_TABLE_H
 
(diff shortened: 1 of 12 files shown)


google/cloud/bigtable/table.h
```diff
@@ -831,12 +829,8 @@ class Table {
    * @par Example
    * @snippet data_async_snippets.cc async read rows
    */
-  template <typename RowFunctor, typename FinishFunctor>
```
coryan, line 834: I did not think about this before: in #9272 we imposed an additional requirement: anything that goes into a `std::function<>` must be copy-constructible, but `AsyncRowReader` required (only?) move-constructible.

Thougths?

PR #9326, opened 2022-06-22
feat(pubsub): update subscription builders

coryan:
We provide a few helpers to build the protos. While this may not be a
good idea in the long run, it is the current plan, so we need these
updates from time to time.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9326)


google/cloud/pubsub/subscription_builder.h
```diff
@@ -123,6 +123,67 @@ class PushConfigBuilder {
   std::set<std::string> paths_;
 };
 
+/**
+ * A helper class to build `google::pubsub::v1::BigQueryConfig` protos.
+ *
+ * Makes it easier to build admin requests to
```
devbww, line 129: "to ..." ?

PR #9336, opened 2022-06-23
feat(bigtable): modern `Table` constructors

dbolduc:


google/cloud/bigtable/table.h
```diff
@@ -192,6 +176,63 @@ class Table {
...
+   * @param options Configuration options for the table. The relevant options
+   *     are enumerated in `DataPolicyOptionList`. They include:
+   *   - `DataRetryPolicyOption`
+   *   - `DataBackoffPolicyOption`
+   *   - `IdempotentMutationPolicyOption`
+   *
+   * [app-profile-docs]: https://cloud.google.com/bigtable/docs/app-profiles
+   */
+  explicit Table(std::shared_ptr<DataConnection> conn, std::string project_id,
+                 std::string instance_id, std::string app_profile_id,
```
coryan, line 222: It is a bit weird to have the `app_profile_id` in the middle ... I guess because the same `project_id` and `instance_id` apply to both?  Still, that seems error-prone when compared to the previous constructor.

Since this is about a "modern" API: if we were writing this from scratch, there would be a `Table` class to represent the `project_id`, `instance_id`, `table_id` triplet (like `pubsub::Topic` or `google::cloud::Project` or `spanner::Database), since `Table` is taken maybe `TableName`?

```cc
class TableName {
  public:
    TableName(std::string project_id, std::string instance_id, std::string table_id) ...

    ...
   std::string FullName() const { return "projects/" + project_id_ + "/instances/" + instance_id_ + "/tables/" + table_id_; }
```

Likewise, we would have an `AppProfileName` (or just `AppProfile`). These constructors should consume those classes.  Moreover, it seems that `AppProfileName` should be passed via the `Options`?

coryan (changes requested):

google/cloud/bigtable/table.h
dbolduc: Thanks for the review. I agree with everything you said. There is not much to salvage in this PR, so I am going to close it and start working on the changes.

> It is a bit weird to have the `app_profile_id` in the middle ... I guess because the same `project_id` and `instance_id` apply to both? Still, that seems error-prone when compared to the previous constructor.

I agree. I think I made a bad choice. I was copying the effective order of this: https://github.com/googleapis/google-cloud-cpp/blob/ee484a2ef86d3e851602430ff72c29899b61141e/google/cloud/bigtable/table.h#L222-L223 

> Since this is about a "modern" API: if we were writing this from scratch, there would be a `Table` class to represent the `project_id`, `instance_id`, `table_id` triplet (like `pubsub::Topic` or `google::cloud::Project` or `spanner::Database), since `Table`is taken maybe`TableName`?

Sadly `TableName` is also taken: https://github.com/googleapis/google-cloud-cpp/blob/ee484a2ef86d3e851602430ff72c29899b61141e/google/cloud/bigtable/resource_names.h#L35-L37

I am thinking `bigtable::TableResource` + `bigtable::Instance`, and I punt on admin API stuff like `bigtable::Cluster` or `bigtable::Backup` for now. I opened #9348 for this task.

> Likewise, we would have an `AppProfileName` (or just `AppProfile`). These constructors should consume those classes. Moreover, it seems that `AppProfileName` should be passed via the `Options`?

nit: the Data API only asks for an `app_profile_id`. (I imagine it infers the instance name from the `table_name`). Having an `AppProfileName` would only be of use in the Instance Admin API.

I think passing it as an Option is a good idea, because it is in fact optional. I opened #9349 for this task.

PR #9337, opened 2022-06-23
ci: add new build to test with C++14

coryan:
Before we can stop testing with C++11 we need to start testing with
C++14.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9337)

### DIFF
--- a/ci/cloudbuild/builds/cxx14.sh
+++ b/ci/cloudbuild/builds/cxx14.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2021 Google LLC
+# Copyright 2022 Google LLC
 #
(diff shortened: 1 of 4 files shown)


ci/cloudbuild/builds/cxx14.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
```
dbolduc, line 3: nit

dbolduc (approved):

ci/cloudbuild/builds/cxx14.sh
coryan: Fixed, thanks

Issue #9312, opened 2022-06-19
Spanner Postgresql samples duplicate column name


dbolduc:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/4de02137-cd60-4843-b80a-5a5604419a4b;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/d12b9c4fb6ca5967875703422d52ff036e531ad5/fedora-35-integration-daily/log-4de02137-cd60-4843-b80a-5a5604419a4b.txt

```
//google/cloud/spanner/samples:postgresql_samples                        FAILED in 65.6s
...
Running spanner_postgresql_add_column sample at 2022-06-19T07:15:36Z
Database projects/cloud-cpp-testing-resources/instances/test-instance-14/databases/db-2022-06-19-40ajg3sx4qu538dh dropped.

Duplicate column name albums.marketingbudget.
```

This might be a problem with the service... Googlers can see b/192520720

dbolduc:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/06eb5e5a-10e0-49c9-951e-c147885a75b9;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/97298d7ffaf51db6fdb7c5e32271d520faaefc1c/fedora-36-integration-daily/log-06eb5e5a-10e0-49c9-951e-c147885a75b9.txt

coryan:
90d without repeats, and it is really a problem in the service. Closing.

PR #9342, opened 2022-06-23
Change column name for add samples of spanner postres.

irvifa:
Fix https://github.com/googleapis/google-cloud-cpp/issues/9312

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9342)

### DIFF
--- a/google/cloud/spanner/samples/postgresql_samples.cc
+++ b/google/cloud/spanner/samples/postgresql_samples.cc
@@ -67,7 +67,7 @@ void AddColumn(google::cloud::spanner_admin::DatabaseAdminClient client,
   std::vector<std::string> statements = {
       R"""(
         ALTER TABLE Albums
-            ADD COLUMN MarketingBudget BIGINT
+            ADD COLUMN NumberOfSongs BIGINT
       )""",
   };
   auto metadata =


google/cloud/spanner/samples/postgresql_samples.cc
```diff
@@ -67,7 +67,7 @@ void AddColumn(google::cloud::spanner_admin::DatabaseAdminClient client,
   std::vector<std::string> statements = {
       R"""(
         ALTER TABLE Albums
-            ADD COLUMN MarketingBudget BIGINT
+            ADD COLUMN NumberOfSongs BIGINT
```
devbww, line 70: I don't believe that addresses the issue.  And it also breaks the test.

irvifa: Isn't the column already defined before? Not sure if there's breaking changes on the latest library that expect the exception to be thrown if the columns actually is a duplicate. Is there anyway to run the test locally without using gcb?

irvifa: Not sure if this is related but at least the proto definition says that it should return [`ALREADY_EXIST`](https://github.com/googleapis/googleapis/blob/6379d5fe706781af6682447f77f20d18b4db05b2/google/spanner/admin/database/v1/spanner_database_admin.proto#L573) so I guess this is expected behaviour and the test should expect to throw an exception for the same column name

devbww: > Isn't the column already defined before?

No.  That's the problem.  Otherwise it would happen all the time.

It's presence seems to be an artifact of the `UpdateDatabaseDdl()` being replayed behind the scenes, but we are yet to understand why/when that happens, or how you're supposed to deal with it.

> Is there anyway to run the test locally without using gcb?

Yes, you should be able to build/run the test locally, although you'll need a Spanner instance to point it at.

coryan: > Not sure if this is related but at least the proto definition says that it should return [`ALREADY_EXIST`](https://github.com/googleapis/googleapis/blob/6379d5fe706781af6682447f77f20d18b4db05b2/google/spanner/admin/database/v1/spanner_database_admin.proto#L573) so I guess this is expected behaviour and the test should expect to throw an exception for the same column name

That is correct.  But in our case it fails (very rarely) even though the column does not exist before sending the RPC, and even though we send the RPC only once.  We suspect a bug in the back end, but do not know for sure yet (and yes, the back end team has been informed).

>  Is there anyway to run the test locally without using gcb?

PR #9351, opened 2022-06-25
feat(common): truncation support for plain strings in the RPC log

devbww:
Add `DebugString(std::string, TracingOptions const&)` for adding a plain
string to the RPC log, but with the `truncate_string_field_longer_than`
behavior of `google::protobuf::TextFormat::Printer`.

This will be used in an upcoming change that will log a `resume_token`
from the Spanner streaming read/query protocol, and which is extremely
long.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9351)

### DIFF
--- a/google/cloud/internal/log_wrapper.cc
+++ b/google/cloud/internal/log_wrapper.cc
@@ -99,9 +99,10 @@ std::string DebugString(Status const& status, TracingOptions const& options) {
   return std::move(os).str();
 }
 
-std::string RequestIdForLogging() {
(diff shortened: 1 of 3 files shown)


coryan (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9351)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devbww)

Issue #9327, opened 2022-06-22
Pub/Sub Exactly Once Delivery


coryan:
## Objective

Give customers access to a new feature in Pub/Sub: exactly once delivery. This
feature guarantee that successfully acknowledged messages are not redelivered,
nor are messages redelivered before their ack deadline expires.

## Background

Previously Pub/Sub guaranteed "at least once" delivery, which meant all ack/nack
messages were handled on a best-effort basis.  Errors in ack/nack messages were
not exposed to the application, and no retry attempts are made on these RPCs.

Some documents refer to this feature as "Exactly Once Subscribe" (EOS). Others
use "Exactly Once Delivery" (EOD). Sigh.

The proto changes are deceptively small.

First, when a subscription is created it can be configured to use "exactly once
delivery":

* https://github.com/googleapis/googleapis/commit/f02f4392673ec85157120cf451ce32fcb613ad5a

Second, if you use the streaming API to ack and nack messages you get back
responses with the status of your nack/ack:

* https://github.com/googleapis/googleapis/commit/1f707ab5f82f83644f2812f5f3d94462c72f9870

Note that all client libraries, including the C++ client libraries do **not**
use the streaming pull RPC to ack/nack messages.  We use the `Acknowledge()` and
`ModifyAckDeadline()` RPCs.  Which unfortunately can accept multiple ack ids,
but cannot return how was each ack id handled.  The service has chosen to tunnel
this information via the `ErrorInfo`.  It becomes part of the client library's
job to parse the `ErrorInfo` and deliver the errors to the application.

The C++ client library provides the following class so applications can ack/nack
messages:

```cc
class AckHandler {
 public:
  ~AckHandler();

  AckHandler(AckHandler&&) = default;
  AckHandler& operator=(AckHandler&&) = default;

  void ack() &&;
  void nack() &&;
  std::int32_t delivery_attempt() const;
}
```

## Proposal

We will introduce a new class to ack/nack messages but with (possible) errors:

```cc
class ExactlyOnceAckHandler {
 public:
  ~ExactlyOnceAckHandler();

  ExactlyOnceAckHandler(ExactlyOnceAckHandler&&) = default;
  ExactlyOnceAckHandler& operator=(ExactlyOnceAckHandler&&) = default;

  future<Status> ack() &&;
  future<Status> nack() &&;
  std::int32_t delivery_attempt() const;
}
```

Note that `ack() &&` and `nack() &&` now return a `future<Status>`. See below
[...]

PR #9357, opened 2022-06-27
impl(pubsub): handle streaming pull subscription properties

coryan:
Part of the work for #9327

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9357)

### DIFF
--- a/google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
+++ b/google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
@@ -279,7 +279,7 @@ void StreamingSubscriptionBatchSource::OnRead(
   pending_read_ = false;
   if (response && stream_state_ == StreamState::kActive && !shutdown_) {
     if (response->has_subscription_properties()) {
-      auto enabled =
+      auto const enabled =
           response->subscription_properties().exactly_once_delivery_enabled();
       if (exactly_once_delivery_enabled_ != enabled) {
         exactly_once_delivery_enabled_ = enabled;
@@ -336,9 +336,9 @@ void StreamingSubscriptionBatchSource::DrainQueues(
   auto stream = stream_;
   pending_write_ = true;
 
-  auto stream_ack_deadline = std::chrono::seconds(0);
+  auto stream_ack_deadline = std::int32_t{0};
   if (needs_stream_ack_deadline_update_) {
-    stream_ack_deadline = min_deadline_time_;
+    stream_ack_deadline = static_cast<std::int32_t>(min_deadline_time_.count());
     needs_stream_ack_deadline_update_ = false;
   }
   std::vector<std::pair<std::string, std::chrono::seconds>> deadlines;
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
```diff
@@ -278,6 +278,14 @@ void StreamingSubscriptionBatchSource::OnRead(
   std::unique_lock<std::mutex> lk(mu_);
   pending_read_ = false;
   if (response && stream_state_ == StreamState::kActive && !shutdown_) {
+    if (response->has_subscription_properties()) {
+      auto enabled = absl::make_optional(
+          response->subscription_properties().exactly_once_delivery_enabled());
```
devbww, line 283: Opt: I'm torn on whether dropping the `optional` on the local variable would make this a little clearer/simpler?
```
  auto enabled = response->subscription_properties().exactly_once_delivery_enabled());
  if (exactly_once_delivery_enabled_ != enabled) {
    exactly_once_delivery_enabled_ = enabled;
    needs_stream_ack_deadline_update_ = true;
  }
```

google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
```diff
@@ -323,16 +331,22 @@ void StreamingSubscriptionBatchSource::OnFinish(Status status) {
 
 void StreamingSubscriptionBatchSource::DrainQueues(
     std::unique_lock<std::mutex> lk, bool /*force_flush*/) {
-  if (deadlines_queue_.empty()) return;
+  if (deadlines_queue_.empty() && !needs_stream_ack_deadline_update_) return;
   if (stream_state_ != StreamState::kActive || pending_write_) return;
   auto stream = stream_;
   pending_write_ = true;
 
+  auto stream_ack_deadline = std::int32_t{0};
```
devbww, line 339: Opt: It would be nicer, I reckon, to make `stream_ack_deadline` a `std::chrono::seconds`, and leave the `.count()` to the same expression as the `set_stream_ack_deadline_seconds()`.  The longer we stay in the `chrono` domain, the better.

[Aside: It is a shame that the proto field is called "deadline" (which generally implies a time) instead of "timeout" (which generally implies a duration).  And I guess we might as well mirror their naming.]

devbww (approved):

google/cloud/pubsub/internal/streaming_subscription_batch_source.cc
coryan: `¯\_(ツ)_/¯` I changed it to your approach.

google/cloud/pubsub/internal/streaming_subscription_batch_source.h
```diff
@@ -167,7 +170,9 @@ class StreamingSubscriptionBatchSource
   bool pending_read_ = false;
   Status status_;
   std::shared_ptr<AsyncPullStream> stream_;
+  bool exactly_once_delivery_enabled_ = false;
```
devbww, line 173: This one, I thought, needed to stay as `optional` to catch the transition from unengaged to false.  But I'm surmising.

devbww (approved):

PR #9358, opened 2022-06-27
test(storage): make benchmark more memory efficient

coryan:
The benchmark was allocating a (constant) buffer of random data to
upload on each thread. The default size of this buffer is 64 MiB.
This is a good default, as around 64 MiB the performance plateaus.
We also are starting to run benchmarks with 64 and more threads,
and each thread has several buffers (gRPC, DirectPath, XML, Json,
each gets a different buffer) that quickly pushes memory consumption
too high.  Better to share the buffer across all these threads.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9358)

### DIFF
--- a/google/cloud/storage/benchmarks/throughput_experiment.cc
+++ b/google/cloud/storage/benchmarks/throughput_experiment.cc
@@ -38,13 +38,12 @@ std::string ExtractPeer(
   return p == headers.end() ? std::string{"unknown"} : p->second;
 }
 
-using SharedContents = std::shared_ptr<std::string>;
-
 class UploadObject : public ThroughputExperiment {
  public:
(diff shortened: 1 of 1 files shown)


google/cloud/storage/benchmarks/throughput_experiment.cc
```diff
@@ -38,11 +38,13 @@ std::string ExtractPeer(
   return p == headers.end() ? std::string{"unknown"} : p->second;
 }
 
+using SharedContents = std::shared_ptr<std::string>;
```
devbww, line 41: Opinion: I'm not sure this helps readability.

devbww (approved):

PR #9359, opened 2022-06-27
impl(spanner): fix the handling of PartialResultSet.resume_token

### LINKED ISSUE #8523: clarify protocol around PartialResultSet.resume_token
At the moment we assume that any `PartialResultSet` that completes rows also includes a `resume_token` that will allow us to resume the stream at a point just past the last complete row.  However, it is unclear whether this is guaranteed by the service.

Without such a guarantee, we would need to buffer completed rows until we have a covering `resume_token` (or, presumably, until the buffer becomes too large, at which point we'd probably have to give up).

Implementation-wise, any action here is a little complicated because resume tokens are handled by `PartialResultSetResume` while row buffering (and value merging) is handled by `PartialResultSetSource`.

devbww:
The service only adds a `resume_token` when the `PartialResultSet` ends
with a completed row (except when the `PartialResultSet` also completes
the stream, in which case the `resume_token` remains empty). Therefore,
`resume_token` and `chunked_value` are mutually exclusive.

This means the library has to buffer rows completed within a non-token
`PartialResultSet` if it is to maintain resumability. This mainly affects
the algorithm and data used by `PartialResultSetSource::ReadFromStream()`.
For example, the storage of the token moves from `PartialResultSetResume`
to `PartialResultSetSource`, and the latter gains a `deque` of complete,
but undeliverable rows.

Introduce the `StreamingResumabilityBufferSizeOption` to control the size
of the buffer. If this size is exceeded, complete rows will be delivered
to the user, which makes the stream non-resumable until a new `resume_token`
is received.

Correct the `ConnectionImplTest.ReadSuccess` test case, which was checking
that we did the wrong thing. :-)

Add new `PartialResultSetResume` test cases for stream resumption.

Loop within the `PartialResultSetSourceTest.MultipleResponses` test to
validate that different buffering limits do not affect a complete stream.

Fixes #8523.

While we're here, canonicalize the shape of protobuf text in tests, which
includes giving resume tokens meaningful content, and also include some
additional expectations that `BeginTransaction()` is not called.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9359)

### DIFF
--- a/google/cloud/spanner/internal/logging_result_set_reader_test.cc
+++ b/google/cloud/spanner/internal/logging_result_set_reader_test.cc
@@ -27,6 +27,7 @@ namespace spanner_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace {
 
+using ::testing::_;
 using ::testing::Contains;
 using ::testing::HasSubstr;
 
@@ -46,7 +47,7 @@ TEST_F(LoggingResultSetReaderTest, TryCancel) {
 
 TEST_F(LoggingResultSetReaderTest, Read) {
   auto mock = absl::make_unique<spanner_testing::MockPartialResultSetReader>();
-  EXPECT_CALL(*mock, Read())
+  EXPECT_CALL(*mock, Read(_))
       .WillOnce([] {
         google::spanner::v1::PartialResultSet result;
         result.set_resume_token("test-token");
@@ -54,7 +55,7 @@ TEST_F(LoggingResultSetReaderTest, Read) {
       })
       .WillOnce([] { return absl::optional<PartialResultSet>{}; });
   LoggingResultSetReader reader(std::move(mock), TracingOptions{});
-  auto result = reader.Read();
+  auto result = reader.Read("");
   ASSERT_TRUE(result.has_value());
   EXPECT_EQ("test-token", result->result.resume_token());
 
@@ -62,8 +63,9 @@ TEST_F(LoggingResultSetReaderTest, Read) {
   EXPECT_THAT(log_lines, Contains(HasSubstr("Read")));
   EXPECT_THAT(log_lines, Contains(HasSubstr("test-token")));
 
-  result = reader.Read();
+  result = reader.Read("test-token");
   ASSERT_FALSE(result.has_value());
+
   log_lines = log_.ExtractLines();
   EXPECT_THAT(log_lines, Contains(HasSubstr("(optional-with-no-value)")));
 }
--- a/google/cloud/spanner/internal/partial_result_set_source.cc
+++ b/google/cloud/spanner/internal/partial_result_set_source.cc
@@ -14,185 +14,242 @@
 
 #include "google/cloud/spanner/internal/partial_result_set_source.h"
 #include "google/cloud/spanner/internal/merge_chunk.h"
+#include "google/cloud/spanner/options.h"
 #include "google/cloud/log.h"
+#include "absl/container/fixed_array.h"
 
 namespace google {
 namespace cloud {
 namespace spanner_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+namespace {
+
+using Values = google::protobuf::RepeatedPtrField<google::protobuf::Value>;
+
+// Efficiently move values from one repeated field to another. Starts at
+// index `start`, but always clears all of `src`. This is worth optimizing
+// because it is on the primary path for getting bulk data to the user.
+void ExtractSubrangeAndAppend(Values& src, int start, Values& dst) {
+  auto* const src_arena = src.GetArena();
+  auto* const dst_arena = dst.GetArena();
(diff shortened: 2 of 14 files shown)


google/cloud/spanner/internal/partial_result_set_source.cc
```diff
@@ -14,186 +14,243 @@
...
+    }
+    rows_.push_back(RowFriend::MakeRow(std::move(values), columns_));
+    values.clear();
   }
 
+  // If we didn't combine all the values, leave the remainder for next time.
+  auto* rem_values = result_set->result.mutable_values();
+  ExtractSubrangeAndAppend(values_, values_pos, *rem_values);
+  values_.Swap(rem_values);
+
```
dbolduc, line 253: Here I am confused. Don't we already move out all the values and clear the source, so `result_set->result.values().empty() == true` ? Can we remove this code?
https://github.com/googleapis/google-cloud-cpp/blob/441dba6359ab5a2997cfc2c4bff487414ae2648b/google/cloud/spanner/internal/partial_result_set_source.cc#L187
https://github.com/googleapis/google-cloud-cpp/blob/441dba6359ab5a2997cfc2c4bff487414ae2648b/google/cloud/spanner/internal/partial_result_set_source.cc#L194
https://github.com/googleapis/google-cloud-cpp/blob/441dba6359ab5a2997cfc2c4bff487414ae2648b/google/cloud/spanner/internal/partial_result_set_source.cc#L53

dbolduc (approved):

google/cloud/spanner/internal/partial_result_set_source.cc
```diff
@@ -14,186 +14,243 @@
...
+                    " despite our having asked it not to");
     }
-    auto& front = new_values[0];
-    auto merge_status = MergeChunk(*chunk_, std::move(front));
-    if (!merge_status.ok()) {
-      return merge_status;
+    values_back_incomplete_ = false;
+    values_.Clear();
+  }
+
```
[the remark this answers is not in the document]
devbww: I'm hesitant to because `PartialResultSetSource` doesn't really know anything about chunking.  Take a look at `merge_chunk.h` and `merge_chunk.cc` and let me know if you think moving the example there would have a benefit.  Thanks.

google/cloud/spanner/internal/partial_result_set_source.cc
dbolduc: Ah, thanks for explaining. I see it now.

devbww: Just as a further clarification, if we get to here then `values_pos` is probably equal to `values_.size()`.  That is, we've combined all the values into rows, and there is nothing left over.  The exception is when we've exceeded the buffer size, and yielded rows without having a resume token to suppress their replay.

devbww: And P.S. ... thanks for slogging through all of that!

google/cloud/spanner/internal/partial_result_set_source.cc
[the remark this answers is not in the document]
dbolduc: I am just letting you know that the example was easier for me to grok than the prose. I think it's fine to remove the example, because there are [tests](https://github.com/googleapis/google-cloud-cpp/blob/6f512195c0a14a34ba7637f46d0e4e3f34a91179/google/cloud/spanner/internal/partial_result_set_source_test.cc#L525-L529) that do the same job as the example.

> `PartialResultSetSource` doesn't really know anything about chunking.

This is an odd claim when the logic depends on the `chunked_value`.... 🤷‍♂️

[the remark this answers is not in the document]
devbww: > I am just letting you know that the example was easier for me to grok than the prose. I think it's fine to remove the example, because there are [tests](https://github.com/googleapis/google-cloud-cpp/blob/6f512195c0a14a34ba7637f46d0e4e3f34a91179/google/cloud/spanner/internal/partial_result_set_source_test.cc#L525-L529) that do the same job as the example.
> 
> > `PartialResultSetSource` doesn't really know anything about chunking.
> 
> This is an odd claim when the logic depends on the `chunked_value`.... man_shrugging

I guess what I'm saying is that, while it knows values can be split between responses, it doesn't know anything about how they are reconstituted.

google/cloud/spanner/internal/logging_result_set_reader_test.cc
```diff
@@ -46,24 +47,25 @@ TEST_F(LoggingResultSetReaderTest, TryCancel) {
 
 TEST_F(LoggingResultSetReaderTest, Read) {
   auto mock = absl::make_unique<spanner_testing::MockPartialResultSetReader>();
-  EXPECT_CALL(*mock, Read())
+  EXPECT_CALL(*mock, Read(_))
```
coryan, line 50: FYI, I you can just say `EXPECT_CALL(*mock, Read)`

Issue #9121, opened 2022-06-01
Support `routing_parameters` in the generator


dbolduc:
At the moment, these are only used in the Bigtable Data API (which we handwrite) and Cloud Run (which we do not generate). Here is an example:
https://github.com/googleapis/googleapis/blob/36a7e1dd18aecda768b560a7e5c155f79c867370/google/bigtable/v2/bigtable.proto#L65-L73

For how we should process these fields, see the `google.api.routing` proto: https://github.com/googleapis/googleapis/blob/177c62af7f181a42b91def98dec2e46efdb07b5e/google/api/routing.proto

PR #9360, opened 2022-06-27
impl(generator): process the `RoutingRule` proto

dbolduc:
Part of the work for #9121 

Our metadata generator will interact with the `ExplicitRoutingInfo` data struct in order to generate the code. We use this intermediate object instead of generating code directly from the `google.api.RoutingRule` because the translation is complex enough to warrant unit testing.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9360)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -552,7 +552,7 @@ absl::optional<ResourceRoutingInfo> ParseResourceRoutingHeader(
 ExplicitRoutingInfo ParseExplicitRoutingHeader(
     google::protobuf::MethodDescriptor const& method) {
   ExplicitRoutingInfo info;
-  if (!method.options().HasExtension(google::api::routing)) return {};
+  if (!method.options().HasExtension(google::api::routing)) return info;
   google::api::RoutingRule rule =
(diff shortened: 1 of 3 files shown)


generator/internal/descriptor_utils.cc
```diff
@@ -548,6 +549,50 @@ absl::optional<ResourceRoutingInfo> ParseResourceRoutingHeader(
   return ResourceRoutingInfo{match[0], match[1], match[2], http_rule.body()};
 }
 
+ExplicitRoutingInfo ParseExplicitRoutingHeader(
+    google::protobuf::MethodDescriptor const& method) {
+  ExplicitRoutingInfo info;
+  if (!method.options().HasExtension(google::api::routing)) return {};
```
devbww, line 555: `return info` may help with NRVO.

devbww (approved):

generator/internal/descriptor_utils.cc
dbolduc: Thanks. TIL: RVO / NRVO https://www.fluentcpp.com/2016/11/28/return-value-optimizations/

PR #9364, opened 2022-06-28
test(storage): control benchmark pacing

coryan:
Sometimes we only want to get a few samples per second or per minute.
For example, if we are running the benchmark for multiple days getting
hundreds of samples per second can be overwhelming. With this change the
user can control the maximum sampling rate.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9364)

### DIFF
--- a/google/cloud/storage/benchmarks/throughput_options.cc
+++ b/google/cloud/storage/benchmarks/throughput_options.cc
@@ -186,9 +186,10 @@ google::cloud::StatusOr<ThroughputOptions> ParseThroughputOptions(
        [&options](std::string const& val) {
          absl::Duration d;
          if (absl::ParseDuration(val, &d)) {
+           options.minimum_sample_delay = absl::ToChronoMilliseconds(d);
+         } else {
            options.minimum_sample_delay = std::chrono::milliseconds(-1);
          }
-         options.minimum_sample_delay = absl::ToChronoMilliseconds(d);
        }},
   };
   auto usage = BuildUsage(desc, argv[0]);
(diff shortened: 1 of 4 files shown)


google/cloud/storage/benchmarks/throughput_options.cc
```diff
@@ -309,6 +322,12 @@ google::cloud::StatusOr<ThroughputOptions> ParseThroughputOptions(
     return make_status(os);
   }
 
+  if (options.minimum_sample_delay < std::chrono::milliseconds(-1)) {
```
devbww, line 325: So, -1ms is valid?

google/cloud/storage/benchmarks/throughput_options.cc
```diff
@@ -177,6 +178,18 @@ google::cloud::StatusOr<ThroughputOptions> ParseThroughputOptions(
...
+      {"--minimum-sample-delay",
+       "configure the minimum time between samples."
+       " Sometimes we only want to collect a few samples per second."
+       " This can make the data resulting from a multi-day run more"
+       " manageable.",
+       [&options](std::string const& val) {
+         absl::Duration d;
+         if (absl::ParseDuration(val, &d)) {
+           options.minimum_sample_delay = std::chrono::milliseconds(-1);
+         }
```
devbww, line 190: I presume there should be an `else` there (and that the arms of the conditional would then need to be reversed).

coryan: Fixed.

PR #9368, opened 2022-06-28
feat(generator): support explicit routing headers

### LINKED ISSUE #9121: Support `routing_parameters` in the generator

dbolduc:
Fixes #9121 

The first and last commits are interesting. The middle two are not.

We do not pass our variables through `service_method_vars`. It is easier to work with the data directly.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9368)

### DIFF
--- a/generator/internal/metadata_decorator_generator.cc
+++ b/generator/internal/metadata_decorator_generator.cc
@@ -20,6 +20,7 @@
 #include "generator/internal/predicate_utils.h"
 #include "generator/internal/printer.h"
 #include <google/protobuf/descriptor.h>
+#include <algorithm>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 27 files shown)


generator/internal/metadata_decorator_generator.cc
```diff
@@ -58,12 +58,13 @@ std::string SetMetadataText(google::protobuf::MethodDescriptor const& method,
     if (std::all_of(
             kv.second.begin(), kv.second.end(),
             [](RoutingParameter const& rp) { return rp.pattern == "(.*)"; })) {
-      for (auto i = 0; i != kv.second.size(); ++i){
-        auto const& field = kv.second[i].field_name;
-        text += i == 0 ? "  " : " else ";
-        text += "if (!request." + field + "().empty()) {\n";
-        text += "    params.push_back(\"" + kv.first + "=\" + request." + field + "());\n";
+      bool initial_if = true;
```
coryan, line 61: consider:

```cc
char const* sep = "  ";
for (auto const& rp : kv.second) {
  text += sep;
  ....
  ...
  sep = " else ";
}
```

generator/internal/metadata_decorator_generator.cc
```diff
@@ -32,14 +33,64 @@ std::string SetMetadataText(google::protobuf::MethodDescriptor const& method,
...
+    }
+    // If the method does not have a `google.api.routing` or `google.api.http`
+    // annotation, we do not send the "x-goog-request-params" header.
+    return "  SetMetadata(" + context + ");";
   }
-  // If the method does not have a `google.api.http` annotation, we do not send
-  // the "x-goog-request-params" header.
-  return "  SetMetadata(" + context + ");";
+
+  // clang-format off
```
coryan, line 50: Ah, yes, that is the thing turning off formatting.

coryan (commented):
Do you want to add tests for the async cases?  Or streaming?

generator/internal/metadata_decorator_generator.cc
dbolduc: Done.

generator/internal/metadata_decorator_generator.cc
```diff
@@ -58,12 +58,13 @@ std::string SetMetadataText(google::protobuf::MethodDescriptor const& method,
     if (std::all_of(
             kv.second.begin(), kv.second.end(),
             [](RoutingParameter const& rp) { return rp.pattern == "(.*)"; })) {
-      for (auto i = 0; i != kv.second.size(); ++i){
-        auto const& field = kv.second[i].field_name;
-        text += i == 0 ? "  " : " else ";
-        text += "if (!request." + field + "().empty()) {\n";
-        text += "    params.push_back(\"" + kv.first + "=\" + request." + field + "());\n";
+      bool initial_if = true;
+      for (auto const& rp : kv.second){
```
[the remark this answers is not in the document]
dbolduc: It was a deliberate choice to keep the lines intact

dbolduc:
> Do you want to add tests for the async cases? Or streaming?

I dropped the routing annotation on a streaming read and unary rpc where we generate async versions of the functions. `IsContextMDValid` revealed that if no routing matches were found, I wasn't sending any metadata at all. Yay for testing? Boo for self esteem.

coryan (approved):
I think we still need to figure out streaming writes and streaming read-write RPCs, but those are (a) very rare, and (b) already broken I think.

PR #9369, opened 2022-06-28
deflake(pubsub): mock completion queue to avoid unwanted timers

coryan:
Instead of relying on "long enough" timeouts use a mock completion
queue to control exactly when the timer expire (never in these tests).

Part of the work for #9327, and yes, this is the second deflake, the first one was a temporary fix to unblock other development.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9369)

### DIFF
--- a/google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
+++ b/google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
@@ -33,8 +33,10 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace {
 
 using ::google::cloud::internal::AutomaticallyCreatedBackgroundThreads;
+using ::google::cloud::internal::RunAsyncBase;
 using ::google::cloud::testing_util::AsyncSequencer;
 using ::google::cloud::testing_util::IsOk;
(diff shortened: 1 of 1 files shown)


google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
```diff
@@ -805,11 +806,28 @@ TEST(StreamingSubscriptionBatchSourceTest, StateOStream) {
...
+  auto mock_cq = std::make_shared<MockCompletionQueueImpl>();
+  EXPECT_CALL(*mock_cq, MakeRelativeTimer)
+      .WillRepeatedly([&](std::chrono::nanoseconds) {
+        return aseq.PushBack("MakeRelativeTimer").then([](auto) {
+          return make_status_or(std::chrono::system_clock::now());
+        });
+      });
+  EXPECT_CALL(*mock_cq, RunAsync)
+      .WillRepeatedly([&](std::unique_ptr<RunAsyncBase> f) {
+        aseq.PushBack("RunAsync").then([function = std::move(f)](auto) mutable {
```
dbolduc, line 819: it begins...

dbolduc (approved):

PR #9375, opened 2022-06-28
impl(pubsub): retry ack/nacks with exactly-once

coryan:
If the streaming pull corresponds to a exactly-once delivery
subscription we need to retry the ack/nack messages.

Part of the work for #9327

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9375)


google/cloud/pubsub/internal/exactly_once_policies.h
```diff
@@ -0,0 +1,54 @@
...
+  bool IsPermanentFailure(Status const&) const override;
+
+ private:
+  std::string const ack_id_;
+  std::chrono::steady_clock::time_point deadline_;
+};
+
+/// Build a backoff policy suitable to retry ack/nack messages when exactly-once
+/// is enabled.
+std::unique_ptr<pubsub::BackoffPolicy> ExactlyOnceBackoffPolicy();
```
devbww, line 47: `#include <memory>`

google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
```diff
@@ -512,7 +514,7 @@ TEST(StreamingSubscriptionBatchSourceTest, AckMany) {
               cq, std::move(context), request);
           using Request = ::google::pubsub::v1::StreamingPullRequest;
           // Add expectations for Write() calls with empty subscriptions, only
-          // the first call has a non-empty value and it is already set.
```
devbww, line 515: While we're tweaking, `s/subscriptions, only/subscriptions. Only/`

devbww (approved):

google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
coryan: Done, here and elsewhere.

PR #9379, opened 2022-06-29
impl(pubsub): use unary RPCs to refresh leases

coryan:
So far we have been using the `StreamingPull()` RPC to receive data and
to refresh the message leases. To implement exactly-once delivery we
need to retry the message lease refreshes, and it is better to do this
using a separate unary RPC because (1) it is easier to implement,
(2) it is more consistent with other libraries, and (3) the preferred
approach by the Pub/Sub service folks.

Part of the work for #9327

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9379)

### DIFF
--- a/google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
+++ b/google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
@@ -39,12 +39,12 @@ using ::google::cloud::testing_util::IsOk;
 using ::google::cloud::testing_util::MockCompletionQueueImpl;
 using ::google::cloud::testing_util::StatusIs;
 using ::testing::_;
+using ::testing::AllOf;
 using ::testing::AtLeast;
 using ::testing::AtMost;
 using ::testing::ByMove;
 using ::testing::ElementsAre;
 using ::testing::HasSubstr;
-using ::testing::IsEmpty;
 using ::testing::Property;
 using ::testing::Return;
 using ::testing::Unused;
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
```diff
@@ -579,112 +562,159 @@ TEST(StreamingSubscriptionBatchSourceTest, AckMany) {
...
+  auto mock_cq = std::make_shared<MockCompletionQueueImpl>();
+  EXPECT_CALL(*mock_cq, MakeRelativeTimer)
+      .WillRepeatedly([&](std::chrono::nanoseconds) {
+        return aseq.PushBack("MakeRelativeTimer").then([](auto) {
+          return make_status_or(std::chrono::system_clock::now());
+        });
+      });
+  EXPECT_CALL(*mock_cq, RunAsync)
+      .WillRepeatedly([&](std::unique_ptr<RunAsyncBase> f) {
+        aseq.PushBack("RunAsync").then([function = std::move(f)](auto) mutable {
```
devbww, line 575: Woo hoo!

devbww (approved):

google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
dbolduc: I can only imagine the glee with which @coryan typed that line

Issue #9349, opened 2022-06-24
Add `AppProfileIdOption` for bigtable Data API.


dbolduc:
The App Profile Id is optional, so let's pass it as an option, instead of having a `Table` constructor with and without an `app_profile_id` field. The default value will be the empty string.

```c++
namespace google::cloud::bigtable {
struct AppProfileIdOption {
  using Type = std::string;
};
} // namespace google::cloud::bigtable
```

dbolduc:
So, our connection methods all look like:

```c++
Foo(std::string const& app_profile_id, std::string const& table_name, FooArgs...);
```

They have been in our public API for a few days, but have not been released yet.

I think we should drop the `app_profile_id` parameter, and instead pass the value in the `OptionsSpan` as an `AppProfileIdOption`.

PR #9382, opened 2022-06-29
feat(bigtable): add `AppProfileIdOption`

dbolduc:
Part of the work for #9349 

For now, if the `AppProfileIdOption` is supplied to the `Table` constructor, we copy the value into `Table::app_profile_id_`, and pass that along to the Connection when a call is made.

In a follow up PR, I will modify the Connection API to drop the `app_profile_id` argument from its calls, and instead look for the value in `CurrentOptions()`. I think I will drop `Table::app_profile_id_` in that PR as well.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9382)

### DIFF
--- a/google/cloud/bigtable/options.h
+++ b/google/cloud/bigtable/options.h
@@ -52,6 +52,15 @@ namespace bigtable {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 /**
+ * The application profile id needed for using the replication API.
+ *
+ * @see https://cloud.google.com/bigtable/docs/app-profiles
+ */
+struct AppProfileIdOption {
+  using Type = std::string;
+};
+
+/**
  * The endpoint for data operations.
  *
  * @deprecated Please use `google::cloud::EndpointOption` instead.
--- a/google/cloud/bigtable/options.h
+++ b/google/cloud/bigtable/options.h
@@ -52,9 +52,23 @@ namespace bigtable {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 /**
- * The application profile id needed for using the replication API.
+ * The application profile id.
  *
- * @see https://cloud.google.com/bigtable/docs/app-profiles
+ * An application profile, or app profile, stores settings that tell your Cloud
+ * Bigtable instance how to handle incoming requests from an application. When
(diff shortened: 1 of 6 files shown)


google/cloud/bigtable/table.h
```diff
@@ -37,6 +37,7 @@
 #include "google/cloud/status.h"
 #include "google/cloud/status_or.h"
 #include "absl/meta/type_traits.h"
+#include "options.h"
```
coryan, line 40: s/"options.h"/"google/cloud/options.h"/

coryan (changes requested):

google/cloud/bigtable/options.h
```diff
@@ -51,6 +51,15 @@ namespace cloud {
 namespace bigtable {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+/**
+ * The application profile id needed for using the replication API.
```
devbww, line 55: "Option" and "needed" sound like opposites.

google/cloud/bigtable/table.h
```diff
@@ -953,12 +953,11 @@ class Table {
...
       std::shared_ptr<bigtable::DataConnection>, std::string, std::string,
-      std::string, std::string, Options);
+      std::string, Options);
   explicit Table(std::shared_ptr<bigtable::DataConnection> conn,
                  std::string project_id, std::string instance_id,
-                 std::string app_profile_id, std::string table_id,
-                 Options options = {})
-      : app_profile_id_(std::move(app_profile_id)),
+                 std::string table_id, Options options = {})
+      : app_profile_id_(options.get<AppProfileIdOption>()),
```
devbww, line 960: Should this wait until `options_` is constructed?

google/cloud/bigtable/options.h
dbolduc: Lol. It is _needed_ to use an _optional_ feature. I updated the comment to avoid this confusion.

(not related to your comment, but) I also added some more information on app profiles.

google/cloud/bigtable/table.h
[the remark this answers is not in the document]
dbolduc: D'oh. Fixed.

coryan (approved):

Issue #9381, opened 2022-06-29
BigQuery `AppendRows` API could accept a `write_stream`


dbolduc:
We cannot currently send routing metadata for streaming writes. Mainly because we must configure the `grpc::ClientContext` before we know what the first request will be.

In BigQuery, the semantics are to add extra [routing](https://github.com/googleapis/googleapis/blob/79287704c3d7c2f239b52ecfb74b2efaf7e675d2/google/cloud/bigquery/storage/v1/storage.proto#L178-L181) information in the initial request, via the `write_stream` [field](https://github.com/googleapis/googleapis/blob/79287704c3d7c2f239b52ecfb74b2efaf7e675d2/google/cloud/bigquery/storage/v1/storage.proto#L414-L431). (Note that `write_stream` is also listed as a [method signature)](https://github.com/googleapis/googleapis/blob/79287704c3d7c2f239b52ecfb74b2efaf7e675d2/google/cloud/bigquery/storage/v1/storage.proto#L182)

Maybe we should add `std::string const& write_stream` as a parameter to the client call. Then we could set `internal::GrpcSetupOption` to `[write_stream] (auto& context) { context.AddMetadata("x-goog-request-params", write_stream); }` in order to achieve routing. I do not know whether we should try to modify the initial request to set this field.

coryan:
We do not have time for this.

Issue #9373, opened 2022-06-28
`testing_util::ValidateMetadataFixture` should respect explicit routing


dbolduc:
This is not a problem yet, but we [validate metadata expectations](https://github.com/googleapis/google-cloud-cpp/blob/91aae8c5bbd5182534ba716a07a3d5a1a3bdade0/google/cloud/testing_util/validate_metadata.cc#L93) on the `x-goog-request-params` by only looking at the `google.api.http` rules. Well, `google.api.routing` is the new thing, and when that annotation is present it takes priority over the `google.api.http` rules.

The mismatch between checking only the http rules vs. checking the routing and the http rule can lead to:
  * false positives where the metadata is correct, but does not meet the expectations from the http rule.
  * true negatives where the metadata is not correct, but it meets the expectations from the http rule.

The explicit routing rules depend on the values set in the request. So we would need to have a function that takes in the context, the method name, **and** the request itself. It might be helpful to move `generator_internal::ExplicitRoutingInfo` into `google::cloud::internal` so it can be shared by the generator and by testing_utils.

PR #9383, opened 2022-06-29
refactor(testing_util): prepare metadata validation to handle explicit routing

dbolduc:
Part of the work for #9373 

**REVIEWER**: the actual refactor is in the second commit, which is mostly moving things from the source to the header. The first commit contains updates to all the tests.

For my next trick, I am adding a templated `Request` parameter to `IsContextMDValid`, and dropping the return value. Explicit routing looks into the contents of a request so we will need this information.

The case of streaming writes is weird because there is no request when we add metadata to the `grpc::ClientContext`. So we just send empty requests in our tests. Note that `BigQueryWriteClient::AppendRows` would not pass this test. (This has been true forever. I opened #9381 to address it).

Also, make the error message for the metadata matcher more informative (like it used to be).
```
Value of: actual
Expected: contains at least one element that has a first field that is equal to "table_name", and has a second field that matches the pattern: "projects/*/instances/*/tables/*"
  Actual: { ("table_name", "proijslkfj/ps/snfinstces/sles/slk") }
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9383)

### DIFF
--- a/google/cloud/testing_util/validate_metadata.cc
+++ b/google/cloud/testing_util/validate_metadata.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/status_or.h"
 #include <google/api/annotations.pb.h>
 #include <google/protobuf/descriptor.h>
+#include <gmock/gmock.h>
 #include <grpcpp/completion_queue.h>
 #include <grpcpp/generic/generic_stub.h>
 #include <grpcpp/server.h>
@@ -68,9 +69,15 @@ RoutingHeaders ExtractMDFromHeader(std::string header) {
   return res;
 }
 
-MATCHER_P(ContainsStdRegex, pattern, "ContainsRegex using std::regex") {
-  std::regex regex(pattern);
-  return std::regex_search(arg, regex);
+/**
+ * We do not use `::testing::MatchesRegex` because our Windows builds use a
+ * googletest built with `GTEST_USES_SIMPLE_RE`, instead of
+ * `GTEST_USES_POSIX_RE`.
+ */
+MATCHER_P(MatchesGlob, glob, "matches the glob: \"" + glob + "\"") {
+  // Translate the glob into a regex pattern.
+  std::regex regex(absl::StrReplaceAll(glob, {{"*", "[^/]+"}}));
+  return std::regex_match(arg, regex);
 }
 
 /**
@@ -110,15 +117,14 @@ RoutingHeaders FromHttpRule(google::api::HttpRule const& http,
            std::sregex_iterator(pattern.begin(), pattern.end(), subst_re);
        i != std::sregex_iterator(); ++i) {
     std::string const& param = (*i)[1].str();
-    std::string const& expected_pattern =
-        absl::StrReplaceAll((*i)[2].str(), {{"*", "[^/]+"}});
+    std::string const& expected_pattern = (*i)[2].str();
     headers.insert(std::make_pair(param, expected_pattern));
   }
   return headers;
 }
 
 RoutingHeaders ExtractRoutingHeaders(
-    std::string const& method,
+    std::string const& method, google::protobuf::Message const&,
     absl::optional<std::string> const& resource_name) {
   auto const* method_desc =
       google::protobuf::DescriptorPool::generated_pool()->FindMethodByName(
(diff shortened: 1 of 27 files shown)


google/cloud/testing_util/validate_metadata.h
```diff
@@ -15,20 +15,55 @@
...
+
+std::map<std::string, std::string> ExtractMDFromHeader(std::string header);
+
+std::map<std::string, std::string> FromHttpRule(
+    google::api::HttpRule const& http,
+    absl::optional<std::string> const& resource_name);
+
+template <typename Request>
+std::map<std::string, std::string> ExtractRoutingHeaders(
+    std::string const& method, Request const&,
```
coryan, line 55: All protos inherit from `google::protobuf::Message`.  Could use use `google::protobuf::Message const&` here and move this code back to the `.cc` file?

google/cloud/testing_util/validate_metadata.h
```diff
@@ -15,20 +15,60 @@
...
```
coryan, line 55: Consider consuming a `google::protobuf::Message const& request` instead and keeping in the .cc file.

google/cloud/testing_util/validate_metadata.h
```diff
@@ -15,20 +15,60 @@
...
 
+/**
+ * We do not use `::testing::MatchesRegex` because our Windows builds use a
+ * googletest built with `GTEST_USES_SIMPLE_RE`, instead of
+ * `GTEST_USES_POSIX_RE`.
+ */
+MATCHER_P(MatchesMetadataRegex, pattern,
+          "matches the pattern: \"" + pattern + "\"") {
+  // Translate the pattern into a form that can be used by std::regex
+  std::regex regex(absl::StrReplaceAll(pattern, {{"*", "[^/]+"}}));
```
devbww, line 43: The inputs strings are "globs" (hence the need for the replacement), so I would avoid the terms "Regex" in the matcher name and "pattern" in its argument.

And then, it becomes clearer that the stuff about Windows is an implementation detail, rather than something that should be documented about the matcher.

google/cloud/testing_util/validate_metadata.h
dbolduc: Ahh. I didn't think of that. Thanks for the suggestion.

I reverted the commit that moved the definitions from the source into the header. It makes this PR a little too granular, but that's ok. I will pad my contributor stats.

google/cloud/testing_util/validate_metadata.h
dbolduc: Done.

google/cloud/testing_util/validate_metadata.h
dbolduc: Thanks. I was confused about how to refer to these things. I will use your lingo.

google/cloud/testing_util/validate_metadata.cc
```diff
@@ -68,9 +69,15 @@ RoutingHeaders ExtractMDFromHeader(std::string header) {
...
 }
 
-MATCHER_P(ContainsStdRegex, pattern, "ContainsRegex using std::regex") {
-  std::regex regex(pattern);
-  return std::regex_search(arg, regex);
+/**
+ * We do not use `::testing::MatchesRegex` because our Windows builds use a
+ * googletest built with `GTEST_USES_SIMPLE_RE`, instead of
+ * `GTEST_USES_POSIX_RE`.
+ */
```
devbww, line 76: I would still suggest not having this as a main/doxygen-level comment on `MatchesGlob`.  The decision not to use `testing::MatchesRegex` is only an artifact of how the implementation converts globs to regexs.  That is, move it to the body.

dbolduc: So is there is a clean way for me to convert the glob into a form that is accepted by `::testing::MatchesRegex` on all platforms? (clean meaning no `#ifdef`s)

if so, let's use that.

if not, I think the comment makes sense where it is.

coryan (approved):

google/cloud/testing_util/validate_metadata.cc
devbww: > is there a clean way for me to convert the glob

I don't think so.  glob `*` conversion requires character classes, which are not supported by `GTEST_USES_SIMPLE_RE`.

> I think the comment makes sense where it is.

I still disagree, no matter what the details of the internal glob->regex conversion.  But, if I haven't convinced you yet, so be it.

PR #9386, opened 2022-06-29
impl(pubsub): retry lease extensions for exactly-once

coryan:
When exactly-once delivery is enabled we want to retry lease extensions.
The spec says "try 3 times", I have questions about why that number, but
for now I am just going to hard-code it, it will be easy to change
later if needed.

Part of the work for #9327

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9386)


google/cloud/pubsub/internal/extend_leases_with_retry_test.cc
```diff
@@ -0,0 +1,175 @@
...
+            {"test-003", "PERMANENT_ERROR_INVALID_BLAH"},
+        }));
+      });
+  for (int i = 0; i != 3; ++i) {
+    EXPECT_CALL(*mock_cq, MakeRelativeTimer).WillOnce(timer);
+    EXPECT_CALL(*mock, AsyncModifyAckDeadline(
+                           _, _,
+                           Property(&ModifyAckDeadlineRequest::ack_ids,
+                                    ElementsAre("test-001", "test-002"))))
+        .WillRepeatedly(transient);
```
dbolduc, line 156: s/WillRepeatedly/WillOnce/

google/cloud/pubsub/internal/extend_leases_with_retry_test.cc
```diff
@@ -0,0 +1,175 @@
...
+  request.add_ack_ids("test-002");
+  auto result = ExtendLeasesWithRetry(mock, CompletionQueue(mock_cq), request);
+  EXPECT_THAT(result.get(), StatusIs(StatusCode::kUnknown));
+}
+
+TEST(ExtendLeasesWithRetry, FailureTooManyTransients) {
+  auto mock = std::make_shared<MockSubscriberStub>();
+  auto mock_cq = std::make_shared<MockCompletionQueueImpl>();
+
+  auto timer = [] {
```
dbolduc, line 127: nit: is there any reason why `timer` and `transient` are defined out here? it is inconsistent with the rest of the tests

google/cloud/pubsub/internal/extend_leases_with_retry.cc
```diff
@@ -0,0 +1,171 @@
...
+    Status const& status) {
+  if (ExactlyOnceRetryable(status.code())) return request;
+
+  auto permanent_error = [](std::string const& ack_id, Status const& status) {
+    auto const& m = status.error_info().metadata();
+    auto f = m.find(ack_id);
+    auto const permanent =
+        f == m.end() || f->second.rfind("TRANSIENT_FAILURE_", 0) != 0;
+    if (permanent) {
+      GCP_LOG(WARNING)
```
dbolduc, line 44: Not a big deal, but there is nothing to control whether this operation logs or not. You could introduce Options and check for some new `TracingComponentsOption`. but eh.

google/cloud/pubsub/internal/extend_leases_with_retry.cc
```diff
@@ -0,0 +1,171 @@
...
+#include <memory>
+
+namespace google {
+namespace cloud {
+namespace pubsub_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace {
+
+// This value is hard-coded in the spec.
+auto constexpr kMaxRetryAttempts = 3;
```
dbolduc, line 29: Is the spec: "try three times" or "retry three times" ?

If the former, I would think this value should be 2.

google/cloud/pubsub/internal/extend_leases_with_retry.cc
dbolduc: Nevermind. It looks like we have a lot of things that log outside of our logging classes

google/cloud/pubsub/internal/extend_leases_with_retry.cc
coryan: Fixed.

google/cloud/pubsub/internal/extend_leases_with_retry.cc
coryan: Ack, but I think you now agree this is fine.

google/cloud/pubsub/internal/extend_leases_with_retry_test.cc
coryan: When using `WillRepeatedly()` you cannot use `ByMove()`, but now that we are using `WillOnce()` this is not needed

google/cloud/pubsub/internal/extend_leases_with_retry_test.cc
coryan: Fixed

dbolduc (approved):

PR #9396, opened 2022-06-30
impl(pubsub): exactly-once delivery retries lease extensions

coryan:
This finally uses the lease extensions on a subscription with
exactly-once delivery enabled.

Part of the work for #9327

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9396)

### DIFF
--- a/google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
+++ b/google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
@@ -623,6 +623,28 @@ std::unique_ptr<pubsub_testing::MockAsyncPullStream> MakeExactlyOnceStream(
   return stream;
 }
 
+// Wait until the exactly once stream is ready.  Refactors some repetitive code.
+// The promise returned here will trigger a `Write()` and `Read()` call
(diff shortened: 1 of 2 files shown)


google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
```diff
@@ -623,6 +623,28 @@ std::unique_ptr<pubsub_testing::MockAsyncPullStream> MakeExactlyOnceStream(
   return stream;
 }
 
+// Wait until the exactly once stream is ready.  Refactors some repetitive code.
+// The promise returned here will trigger a `Write()` and `Read()` call
+// corresponding to the initial update of the stream's deadline (as this is a
+// exactly once stream), and the loop for `Read()`.
```
devbww, line 629: s/a exactly once stream/an exactly-once stream/

devbww (approved):

google/cloud/pubsub/internal/streaming_subscription_batch_source_test.cc
coryan: Fixed, thanks

PR #9397, opened 2022-06-30
cleanup(bigtable): ergonomic string ctor for TableResource

dbolduc:
I am sorry, but writing code like `TableResource(InstanceResource(Project("p"), "i"), "t)"` is just too painful for me. I am sure it would be painful for our customers too.

So add a constructor that accepts strings. `TableResource("p", "i", "t")`

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9397)

### DIFF
--- a/google/cloud/bigtable/table_resource.cc
+++ b/google/cloud/bigtable/table_resource.cc
@@ -26,8 +26,9 @@ TableResource::TableResource(InstanceResource instance, std::string table_id)
 
 TableResource::TableResource(std::string project_id, std::string instance_id,
                              std::string table_id)
-    : instance_(Project(std::move(project_id)), std::move(instance_id)),
-      table_id_(std::move(table_id)) {}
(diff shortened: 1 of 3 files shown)


coryan (approved):
Not opposed to this notion, though it opens some questions:

* Should there be a `TableResource(Project, std::string, std::string)` overload too?
* What happens when we have resources with 5 components (they do exist)?

dbolduc:
> * Should there be a `TableResource(Project, std::string, std::string)` overload too?

`spanner::Database` doesn't have it: https://github.com/googleapis/google-cloud-cpp/blob/fdfaca2bb37f5f31b08e11296ec86d81dc68c52e/google/cloud/spanner/database.h#L45-L59

I didn't add one. But it would be reasonable to have one.

> * What happens when we have resources with 5 components (they do exist)?

Let's take it on a case by case basis. Maybe it makes sense to have 5 constructors. Maybe it doesn't.

PR #9400, opened 2022-06-30
test(storage): more tweaks to ranged reads

coryan:
The benchmark would perform a range read with 50% probability, but some
ranged reads were "full" as the range was larger than the object. I
found it hard to reason about this, so now only the range size controls
whether the read is full or not.

I also rediscovered a quirk of the JSON API, that it is better avoided:
reading the last 0 bytes actually returns all the bytes. Maybe we should
fix this in the client library itself, for now just avoid running into
it.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9400)

### DIFF
--- a/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
+++ b/google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
@@ -369,15 +369,18 @@ void RunThread(ThroughputOptions const& options, std::string const& bucket_name,
       options.minimum_read_size, options.maximum_read_size,
       options.read_size_quantum);
 
-  auto const read_range_enabled =
-      options.minimum_read_size != options.maximum_read_size;
   auto read_range_generator = [&](auto& g, std::int64_t object_size)
       -> absl::optional<std::pair<std::int64_t, std::int64_t>> {
-    if (!read_range_enabled || !std::bernoulli_distribution{}(g)) {
-      return absl::nullopt;
-    }
     auto offset = (std::min)(object_size, read_offset_generator(g));
     auto size = (std::min)(object_size - offset, read_size_generator(g));
+    // This makes it easier to control the ratio of ranged vs. full reads from
+    // the command-line. To make more full reads happen set the read range size
+    // to be larger than the object sizes. The larger this read range size is,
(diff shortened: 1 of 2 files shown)


google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -369,15 +369,18 @@ void RunThread(ThroughputOptions const& options, std::string const& bucket_name,
...
-    }
     auto offset = (std::min)(object_size, read_offset_generator(g));
     auto size = (std::min)(object_size - offset, read_size_generator(g));
+    // This makes it easier to control the ratio of ranged vs. full reads from
+    // the command-line. To make more full reads happen set the read range size
+    // to be larger than the object sizes. The larger this read range size is,
+    // the higher the proportion of full range reads.
+    if (offset == 0 && size == object_size) return absl::nullopt;
+    // The REST API has a quirk: reading the last 0 bytes returns all the bytes,
+    // just read the *first* 0 bytes in that case.
```
devbww, line 382: s/bytes, just/bytes. Just/

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
```diff
@@ -369,15 +369,18 @@ void RunThread(ThroughputOptions const& options, std::string const& bucket_name,
...
+    if (offset == object_size) return std::make_pair(0, 0);
```
devbww, line 383: Should you check `size == 0` there too?

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
coryan: Done.

google/cloud/storage/benchmarks/storage_throughput_vs_cpu_benchmark.cc
coryan: I think it is implied, or it is now that I fixed some validation :grin:

devbww (approved):

Issue #9309, opened 2022-06-19
Bigtable example code uses `DataConnection`


dbolduc:
After #8860 is done, we should recommend the `DataConnection` API over the `DataClient` API in all sample code.

This also applies to our internal code (integration tests and such)

dbolduc:
Only the quickstart remains.

Issue #9308, opened 2022-06-19
Show how to mock `bigtable::Table` calls


dbolduc:
After #8860 is done, we should offer clear documentation for how to mock the behavior of `Table` using the `MockDataConnection`. This will be similar to `spanner/samples/mock_execute_query.cc` and `bigquery/samples/bigquery_read_mock.cc`, but for bigtable.

We should offer examples for:
* `Table::Apply` (because it's normal)
* `Table::ReadRows` (because it's weird)
* `Table::AsyncReadRows` (because it's weird)

Issue #7434, opened 2021-10-09
Prefer using unified credentials when using emulator(s)


coryan:
Typically we switch the credentials to `grpc::InsecureChannelCredentials()` when the emulator overrides are in effect. Maybe we should use the equivalent `google::cloud::MakeInsecureCredentials()`.

devjgm:
Still want to do this.

coryan:
Closing, no time to work on this for now.

PR #9403, opened 2022-06-30
feat(bigtable): modern `Table` constructor

### LINKED ISSUE #1161: Converge programmatic credentials story for C++ libraries
At the moment there exists gRPC and Apiary credentials story which breaks causes a break in consistency between providing credentials. 

**Goal**: User can set Google service account credentials programmatically consistently across libraries in this repository while hiding what happens underneath between gRPC and Apiary based libraries.

### LINKED ISSUE #6465: Use unified credentials in bigtable
Change the `bigtable` library to use unified credentials.

### LINKED ISSUE #8860: Make Modern Bigtable Data API public
We are nearly ready to do this. The remaining work is to:
- [x] Add `Options` to the `Table` constructor that accepts `DataConnection`

Then we can shuffle the code around:
- [x] move `DataConnection`, data policy options from `bigtable_internal` -> `bigtable`
- [ ] move `Table::Table(std::shared_ptr<DataConnection>, ...)` from `private` -> `public`
- [x] move `MockDataConnection` out of `internal`
- [x] move `MockRowReader` out of `internal`

dbolduc:
Fixes #8860
Technically this fixes #6465 which fixes #1161. Although we should probably update our [credentials sample](https://github.com/googleapis/google-cloud-cpp/blob/c8a4920f64e5fdafb180d9ae1e30bcf4f7bbb3ff/google/cloud/bigtable/examples/bigtable_grpc_credentials.cc). (I think of that task as #7434).

Release the `Table` constructor that accepts a `DataConnection`. Yay, we are close to having a good mocking story. and a good credentials story. (I say close because there is still the pesky legacy API which does not have these things and lives alongside the "new shinny").

For now I just list the relevant options in the constructor's documentation. I would like to link to examples of how the `Options` are used. But the examples must exist first. (#9309 tracks this work).

I would also like to provide guidance on how to mock the `Table` calls in this constructor's documentation. (which is blocked on #9308).

I think this feature warrants a detailed CHANGELOG entry, but I am going to wait until the other modernization work is done before writing the post.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9403)

### DIFF
--- a/google/cloud/bigtable/options.h
+++ b/google/cloud/bigtable/options.h
@@ -145,17 +145,17 @@ using DataLimitedErrorCountRetryPolicy =
     ::google::cloud::internal::LimitedErrorCountRetryPolicy<
         bigtable::internal::SafeGrpcRetry>;
 
-/// Option to use with `google::cloud::Options`.
+/// Option to configure the retry policy used by `Table`.
 struct DataRetryPolicyOption {
   using Type = std::shared_ptr<DataRetryPolicy>;
 };
 
-/// Option to use with `google::cloud::Options`.
+/// Option to configure the backoff policy used by `Table`.
 struct DataBackoffPolicyOption {
   using Type = std::shared_ptr<BackoffPolicy>;
 };
 
-/// Option to use with `google::cloud::Options`.
+/// Option to configure the idempotency policy used by `Table`.
 struct IdempotentMutationPolicyOption {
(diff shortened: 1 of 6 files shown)


coryan (approved):

dbolduc:
Another build failed to trigger. We go again.

dbolduc:
oh duh, because the trigger is disabled...

Issue #4964, opened 2020-08-25
Consider fully qualifying using-declarations in examples


coryan:
Our examples and code snippets typically have several using declarations to make the code more readable, e.g.:

```cc
//! [foo-tag]
using google::cloud::StatusOr;
// ... ... more example ...
//! [foo-tag]
```

As @mr-salty says:

https://github.com/googleapis/google-cloud-cpp/pull/4963#pullrequestreview-474934853

maybe we should fully qualify those.

devjgm:
+1 to a cleanup to fully qualify these.

mr-salty:
More generally, I'd propose we recommend following https://abseil.io/tips/ in our code. Internally, these are "recommended but not required". They do have solid rationale behind them and are subject to discussion and review before being published, and in my experience most code does adhere to them. I think the same would be reasonable for our codebase.

https://github.com/googleapis/google-cloud-cpp/blob/master/CONTRIBUTING.md#style says:
> This repository follows the Google C++ Style Guide, with some additional constraints specified in the [style guide](https://github.com/googleapis/google-cloud-cpp/blob/master/doc/cpp-style-guide.md). Please make sure your contributions adhere to the style guide.

we could add something there, or to the linked (local) style guide variations

PR #9407, opened 2022-07-01
cleanup(pubsub): use fully-qualified google::protobuf types

devbww:
This makes it easier to do any downstream mapping of names
in environments with non-standard protobuf namespaces.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9407)

### DIFF
--- a/google/cloud/pubsub/snapshot_builder.h
+++ b/google/cloud/pubsub/snapshot_builder.h
@@ -60,7 +60,8 @@ class SnapshotBuilder {
   ///@{
   SnapshotBuilder& add_label(std::string const& key,
                              std::string const& value) & {
-    using value_type = protobuf::Map<std::string, std::string>::value_type;
+    using value_type =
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/snapshot_builder.h
```diff
@@ -60,7 +60,8 @@ class SnapshotBuilder {
   ///@{
   SnapshotBuilder& add_label(std::string const& key,
                              std::string const& value) & {
-    using value_type = protobuf::Map<std::string, std::string>::value_type;
+    using value_type =
+        google::protobuf::Map<std::string, std::string>::value_type;
```
dbolduc, line 64: The ghosts of Cloud C++ past suggest that we should fully qualify these #4964

dbolduc: Ack.

PR #9408, opened 2022-07-01
fix(generator): handle explicit routing params for nested fields

dbolduc:
Handle fields that look like `field.sub_field`. I re-read the explicit routing parameters spec and they allow for these nested fields.

We copy the implementation from the `HttpRule`: https://github.com/googleapis/google-cloud-cpp/blob/d67dce6e8a69a78f4f47045172b6145875dee348/generator/internal/descriptor_utils.cc#L255-L264

In order to test this, `IsContextMDValid` needs to learn how to generically access this field name. (It did not need to before because the `HttpRule` only checks that a routing key exists and matches some known pattern. It does not set its expectations by looking at the request's contents like the `RoutingRule` does.)

I added a new test for the nested case. And updated all explicit routing tests to both use `IsContextMDValid` and to verify the expected headers manually. This sorta-kinda tests `IsContextMDValid`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9408)

### DIFF
--- a/google/cloud/testing_util/validate_metadata.cc
+++ b/google/cloud/testing_util/validate_metadata.cc
@@ -31,6 +31,10 @@
 #include <iterator>
 #include <regex>
 
+// Undefine a Windows macro, which conflicts with
+// `protobuf::Reflection::GetMessage`
+#undef GetMessage
+
 namespace google {
 namespace cloud {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
--- a/google/cloud/testing_util/validate_metadata.cc
+++ b/google/cloud/testing_util/validate_metadata.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/internal/absl_str_replace_quiet.h"
 #include "google/cloud/log.h"
 #include "google/cloud/status_or.h"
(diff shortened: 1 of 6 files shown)


google/cloud/testing_util/validate_metadata.cc
```diff
@@ -83,6 +85,23 @@ MATCHER_P(MatchesGlob, glob, "matches the glob: \"" + glob + "\"") {
...
+// solution.
+std::string GetField(  // NOLINT(misc-no-recursion)
+    std::deque<std::string> fields,
+    google::protobuf::Descriptor const* input_type,
+    google::protobuf::Message const& msg) {
+  if (fields.empty()) {
+    GCP_LOG(FATAL) << "Empty field name defined in RoutingRule.";
+  }
+  auto const& field_name = fields.front();
+  fields.pop_front();
```
devbww, line 98: Doesn't that invalidate `field_name`?

dbolduc: d'oh. `msan` agrees with you.

devbww (approved):

PR #9413, opened 2022-07-01
impl(pubsub): use `ExactlyOnceAckHandler`

coryan:
While the implementation always uses `ExactlyOnceAckHandler`, the
application will receive the `*AckHandler` type that it requests. For
applications continuing to use `pubsub::AckHandler` the implementation
is wrapped, and any errors are logged.

Part of the work for #9327

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9413)

### DIFF
--- a/google/cloud/pubsub/internal/subscription_session.h
+++ b/google/cloud/pubsub/internal/subscription_session.h
@@ -42,7 +42,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
  *
  * Applications provide a callable compatible with this type to receive
  * messages.  They acknowledge (or reject) messages using
- * `ExactlyOnceAckHandler`. This is a move-only type to support asynchronously
+ * `ExactlyOnceAckHandler`.  This is a move-only type to support asynchronous
  * acknowledgments.
  */
 using ExactlyOnceApplicationCallback =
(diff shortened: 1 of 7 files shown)


google/cloud/pubsub/internal/subscription_session.h
```diff
@@ -34,10 +35,28 @@ namespace cloud {
...
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+// TODO(#9327) - move this to application_callback.h
+/**
+ * Defines the interface for application-level callbacks with exactly-once
+ * delivery.
+ *
+ * Applications provide a callable compatible with this type to receive
+ * messages.  They acknowledge (or reject) messages using
+ * `ExactlyOnceAckHandler`. This is a move-only type to support asynchronously
```
dbolduc, line 45: nit: s/asynchronously/asynchronous/

dbolduc: also super nit: earlier in the paragraph you put two spaces after the period, but here you put one.

dbolduc (approved):

google/cloud/pubsub/internal/subscription_session.h
coryan: Fixed, thanks

PR #9418, opened 2022-07-02
cleanup(rest): move code to matching `.cc` file

coryan:
The `AsStatus()` functions were declared in `rest_response.h`, but
defined in `curl_rest_response.cc`.  That seems unnecessarily confusing.
The test were also in an mismatched file.

I also refactored a function that `clang-tidy` claims has high
complexity to make it less so.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9418)


devbww:
> The `AsStatus()` functions were declared in `rest_response.h`, but defined in `rest_response.cc`. That seems unnecessarily confusing.

s/defined in `rest_response.cc`/defined in `curl_rest_response.cc`/

devbww (approved):

PR #9422, opened 2022-07-05
ci: missing export in build.sh

coryan:
When building with `--cloud=...` we need to export the
`VERBOSE_FLAG` environment variable. In most command-line builds this
is exported via the `docker run --env=...` commands, so it was hard to
detect. To make it easier to detect such problems, I reordered the
`export` commands to match the `docker run` order.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9422)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -179,10 +179,11 @@ fi
 CODECOV_TOKEN="$(tr -d '[:space:]' <<<"${CODECOV_TOKEN:-}")"
 LOG_LINKER_PAT="$(tr -d '[:space:]' <<<"${LOG_LINKER_PAT:-}")"
 
-export TRIGGER_TYPE
+export CODECOV_TOKEN
 export BRANCH_NAME
 export COMMIT_SHA
-export CODECOV_TOKEN
+export TRIGGER_TYPE
+export VERBOSE_FLAG
 export LOG_LINKER_PAT
 
 # --local is the most fundamental build mode, in that all other builds


ci/cloudbuild/build.sh
```diff
@@ -179,10 +179,11 @@ fi
 CODECOV_TOKEN="$(tr -d '[:space:]' <<<"${CODECOV_TOKEN:-}")"
 LOG_LINKER_PAT="$(tr -d '[:space:]' <<<"${LOG_LINKER_PAT:-}")"
 
-export TRIGGER_TYPE
+export CODECOV_TOKEN
 export BRANCH_NAME
 export COMMIT_SHA
-export CODECOV_TOKEN
+export TRIGGER_TYPE
+export VERBOSE_FLAG
```
devbww, line 186: How does even an exported VERBOSE_FLAG make it through to a cloud build?  I thought that environment was fully controlled.

coryan: If you run `ci/cloudbuild/build.sh -t clang-tidy-pr --cloud=...` then this `ci/cloudbuild/build.sh` first runs locally, then calls `gcloud builds submit` to start "something" (I will explain in a second) in the google cloud build:

https://github.com/googleapis/google-cloud-cpp/blob/3f7dd3768dbc16c39d72bd2ec6e3c4519234018a/ci/cloudbuild/build.sh#L281

The "something" is defined here:

https://github.com/googleapis/google-cloud-cpp/blob/3f7dd3768dbc16c39d72bd2ec6e3c4519234018a/ci/cloudbuild/build.sh#L275

Which actually starts `ci/cloud/build.sh` *again*, but this time in the GCB environment:

https://github.com/googleapis/google-cloud-cpp/blob/3f7dd3768dbc16c39d72bd2ec6e3c4519234018a/ci/cloudbuild/cloudbuild.yaml#L98

That in turn calls `clang-tidy.sh`:

https://github.com/googleapis/google-cloud-cpp/blob/3f7dd3768dbc16c39d72bd2ec6e3c4519234018a/ci/cloudbuild/build.sh#L234

This time "locally" (that is local w.r.t. the GCB environment) and at that point we need the environment variables to be set.

I won't say  "hopefully that makes sense", but maybe "hopefully that explains things"?

devbww: > Which actually starts `ci/cloud/build.sh` again, but this time in the GCB environment

Hang on.  How does this ever get a verbose flag?

devbww: > whether it gets it or not, the `clang-tidy.sh` script still checks for the environment variable.

Then I reckon `clang-tidy.sh` would (also) do well not to depend on `VERBOSE_FLAG` even being in the environment.

PR #9430, opened 2022-07-06
chore: modernize bigtable changelog update

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9430)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,7 +135,7 @@ interface that more closely matches the client surface of `Table`. Read more
 about `*Connection` classes in our
 [Architecture Design][architecture-connection] document.
 
-**What are the benefits of `DataConnection`?**
+#### What are the benefits of `DataConnection`?
 
 The new API greatly simplifies mocking. Every `Table::Foo(..)` call has an
 associated `DataConnection::Foo(...)` call. This allows you to set expectations
@@ -149,7 +149,7 @@ also enables the use of some common library features, such as our
 [`UnifiedCredentialsOption`][guac-dox]. Also, any new features will be added to
 the `DataConnection` API first.
 
-**Do I need to update my code?**
+#### Do I need to update my code?
 
 No. If the benefits are not appealing enough, you do not need to update your
 code. All code that currently uses `DataClient` will continue to function as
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -127,6 +127,95 @@ case it elicits some feedback that requires changes.
...
 
+### [Bigtable](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/bigtable/README.md)
+
+We introduced a [new constructor][modern-table-ctor] for `Table` which accepts a
+`DataConnection` instead of a `DataClient`. The `DataConnection` is a new
+interface that more closely matches the client surface of `Table`. Read more
+about `*Connection` classes in our
+[Architecture Design][architecture-connection] document.
+
+**What are the benefits of `DataConnection`?**
```
coryan, line 138: Any reason we should not use `#### What are the benefits of `DataConnection`?

## [Bigtable](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/bigtable/README.md)

Blah

#### What are the benefits of `DataConnection`?

The new API

CHANGELOG.md
```diff
@@ -127,6 +127,95 @@ case it elicits some feedback that requires changes.
...
+**Do I need to update my code?**
+
+No. If the benefits are not appealing enough, you do not need to update your
+code. All code that currently uses `DataClient` will continue to function as
+before. This includes uses of `testing::MockDataClient`.
+
+However, if you are using `testing::MockDataClient` to mock the behavior of
+`Table` in your tests:
+
+1. I am impressed, that is not easy to do.
```
coryan, line 161: Funny, but maybe remove it? Your third point below is saying the same thing.

CHANGELOG.md
```diff
@@ -127,6 +127,95 @@ case it elicits some feedback that requires changes.
...
+3. The retry, backoff, and idempotency policies are packaged in `Options` as
+   `shared_ptr`s, instead of passed by value as variadic parameters to
+   `Table(..., Policies&&)`.
+4. App profiles are also passed via `Options`.
+
+**What can't the new API do?**
+
+The new [policies][cbt-modern-policies] do not have a
+`Setup(grpc::ClientContext&)` interface. This interface has not been included
+because we believe that setting up the `grpc::ClientContext`...
```
coryan, line 205: Maybe we should move this CHANGELOG entry to a new document, say `bigtable/doc/migrating-from-dataclient.dox` (or `.md`?) and link that document from the CHANGELOG?

CHANGELOG.md
```diff
@@ -127,6 +127,95 @@ case it elicits some feedback that requires changes.
...
+
+1. I am impressed, that is not easy to do.
+2. Be aware that we have announced our intention to remove classes derived from
+   `DataClient` on or around 2023-05. Your tests will break then.
+3. Please consider using `bigtable_mocks::MockDataConnection`. It will greatly
+   simplify your tests.
+
+**How do I update existing `DataClient` code?**
+
+```c++
```
coryan, line 169: nit: we usually say `cc` instead of `c++` for these blocks.

CHANGELOG.md
```diff
@@ -127,6 +127,95 @@ case it elicits some feedback that requires changes.
...
+namespace gc = ::google::cloud;
+namespace cbt = ::google::cloud::bigtable;
+
+// An option which affects the `DataClient` / `DataConnection`.
+auto options = gc::Options{}.set<gc::GrpcNumChannels>(4);
+
+// Existing code...
```
coryan, line 176: Maybe wrap these in a function?

```cc
void OldVersion() {
  auto connection = cbt::MakeDataConnection(options);
...
}

void NewCode() {
  auto connection = cbt::MakeDataConnection(...);
...
}
```

You may want to consider a simpler example first, this gives folks sticker shock.  Start with something that has no profiles and no app profile and then add introduce the additional complexity.

CHANGELOG.md
dbolduc: I changed it to use an h4. Although we do things like `**BREAKING CHANGES:** / **Other Changes:**` in h3's.

google/cloud/bigtable/doc/migrating-from-dataclient.dox
```diff
@@ -0,0 +1,90 @@
...
+cbt::Table UpdatedCode() {
+  auto connection = cbt::MakeDataConnection();
+  return cbt::Table(
+      connection, cbt::TableResource("project-id", "instance-id", "table-id"));
+}
+```
+
+Note that the `DataConnection` is not associated with any resource ids. They
+are instead packaged as a `TableResource` and passed as a single object to
+`Table`.
```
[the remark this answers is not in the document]
dbolduc: The change is really that the customer can access many different tables without needing to use `Table::WithNewTarget`. But given that the functionality was possible with `DataClient`, I do not think it needs a note.

PR #9435, opened 2022-07-06
cleanup(storage): enforce cognitive complexity limits

coryan:
Enable the clang-tidy `readability-function-cognitive-complexity` check
(as an error) for `google/cloud/storage/`.  I think we should turn this
on everywhere, but needed to start somewhere.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9435)


devbww (approved):
> I think we should turn this on everywhere

FWIW, I'm in two minds myself.

PR #9440, opened 2022-07-06
cleanup(bigtable): enable `function-cognitive-complexity` check

coryan:
There were only 3 functions over the limit, only one seems hard to fix.
I simply disabled the check for it.

### DIFF
--- a/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
+++ b/google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
@@ -126,6 +126,47 @@ cbt::SingleRowMutation MakeMutation(
       {cbt::SetCell(options.column_family, options.column, "value")});
 }
 
+StatusOr<std::string> CreateTableIfNeeded(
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/benchmarks/mutation_batcher_throughput_benchmark.cc
```diff
@@ -154,52 +194,18 @@ int main(int argc, char* argv[]) {
...
-        cbt::TableName(options->project_id, options->instance_id, table_id));
-    r.set_view(google::bigtable::admin::v2::Table_View_NAME_ONLY);
-    auto table = admin.GetTable(std::move(r));
-    if (!table) {
-      std::cout << "Error trying to get Table " << table_id << ":\n"
-                << table.status() << std::endl;
-      return 1;
-    }
-  }
+  auto table_id = CreateTableIfNeeded(admin, *options, key_width);
```
dbolduc, line 197: I was skeptical about the `readability-function-cognitive-complexity` project, but this is definitely nicer. Thanks.

dbolduc (approved):

PR #9452, opened 2022-07-11
fix(spanner): avoid evaluation-order issue in function arguments

devbww:
The behavior of
```
  extern void f(std::unique_ptr<T>, T);
  std::unique_ptr<T> p = ...;
  f(std::move(p), *p);
```
is undefined, so introduce a sequence point between dereferencing
the `unique_ptr` and moving it.


coryan (approved):
I wish there was `clang-tidy` check for these:

https://github.com/llvm/llvm-project/issues/21193

Issue #4926, opened 2020-08-17
RPC timeouts and retry loops across services


coryan:
The `google::cloud::bigtable::RPCRetryPolicy` classes have special member functions to setup a timeout:

https://github.com/googleapis/google-cloud-cpp/blob/81adbdabca2da794875fd567591fddc2cf0c8394/google/cloud/bigtable/rpc_retry_policy.h#L98-L101

https://github.com/googleapis/google-cloud-cpp/blob/81adbdabca2da794875fd567591fddc2cf0c8394/google/cloud/bigtable/rpc_retry_policy.cc#L53-L57

This is used in the synchronous retry loop, e.g.:

https://github.com/googleapis/google-cloud-cpp/blob/81adbdabca2da794875fd567591fddc2cf0c8394/google/cloud/bigtable/table.cc#L90-L95

But not in the asynchronous retry loop:

https://github.com/googleapis/google-cloud-cpp/blob/81adbdabca2da794875fd567591fddc2cf0c8394/google/cloud/internal/async_retry_unary_rpc.h#L145-L149

That asynchronous retry loop is used in spanner too, and the `RetryPolicy` for spanner does not have a `Setup()` member function.  Furthermore, the same retry policy class is used in storage, where we would't want to introduce it (because that would create an unwanted dependency on gRPC).

We need a cleaner way to setup per-RPC timeouts across all services, with good support for REST, and we need to use that on the asynchronous and synchronous loops.

coryan:
nit: the retry policies do not sleep today, the retry loop does, and the sleep time comes from the backoff policy.  Your bigger point about this making retry loops hard to test stands.

devjgm:
@devbww any update on this?

devbww:
> any update on this?

No.  Indeed, I am only starting to work on it directly today.

coryan:
Turns out #5959 might be directly related to this.

devbww:
http://docs/document/d/1ooIl9RQGO4Kyz21nghcNtR9nOBMD5dlRkCQqVCFY8fU?resourcekey=0-93iJsZmiyp0wP_-Rzlyn4Q

devjgm:
Still need this. @dbolduc has some thoughts here too.

coryan:
We still want this, but not scheduled yet.

coryan:
Let's consider this for 2023/Q4.

coryan:
Included in the 2024 roadmap document.

[...]

PR #9453, opened 2022-07-12
impl(storage): configure gRPC ClientContext using `CurrentOptions`

dbolduc:
Configure the `grpc::ClientContext`s created by a `storage::internal::GrpcClient` using the `CurrentOptions()`.

Note that, as of writing this, the only option that does anything is `google::cloud::internal::GrpcSetupOption`. And we do not set this option. So we are only gaining a hook for debugging. And we are doing some extra map lookups on every client call. That is not obviously great.

But, if we ever add an option for setting metadata or for setting a deadline we can have it take effect in `google::cloud::internal::ConfigureContext(..., Options)`. for gRPC. (See https://github.com/googleapis/google-cloud-cpp/issues/4926#issuecomment-1125348091 for how I see things playing out)

### DIFF
--- a/google/cloud/storage/internal/grpc_configure_client_context_test.cc
+++ b/google/cloud/storage/internal/grpc_configure_client_context_test.cc
@@ -27,6 +27,7 @@ namespace {
 using ::google::cloud::testing_util::ValidateMetadataFixture;
 using ::testing::Contains;
 using ::testing::IsEmpty;
+using ::testing::MockFunction;
 using ::testing::Pair;
 
 class GrpcConfigureClientContext : public ::testing::Test {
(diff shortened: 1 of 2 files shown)


coryan (approved):

google/cloud/storage/internal/grpc_configure_client_context_test.cc
```diff
@@ -101,6 +102,19 @@ TEST_F(GrpcConfigureClientContext, ApplyQueryParametersQuotaUserAndUserIp) {
   }
 }
 
+TEST_F(GrpcConfigureClientContext, ApplyQueryParametersGrpcOptions) {
+  MockFunction<void(grpc::ClientContext&)> mock;
+  EXPECT_CALL(mock, Call);
+
+  google::cloud::internal::OptionsSpan span(
+      Options{}.set<google::cloud::internal::GrpcSetupOption>(
```
devbww, line 110: A word of caution on what seems to be going on here.

An "internal option" implies to me that we're using thread locals to (conveniently?) plumb state through call layers. 
 I reckon such things would be better passed explicitly.  Options should be reserved for truly optional things from the user.  [Off soap box, and apologies if I'm mistaken.]

dbolduc: Ack. The change is confusing on the surface. We are not using `internal::GrpcSetupOption` for state.

Basically, I want to test that current options (supplied by the user) will be used when setting up the `grpc::ClientContext`. But there are no such public options right now. So we test the one option that does anything: `internal::GrpcSetupOption`. In the future, I could see `GrpcSetupOption` being released, or see us adding other options like a `MetadataOption`. https://github.com/googleapis/google-cloud-cpp/issues/4926#issuecomment-1125348091

PR #9457, opened 2022-07-12
feat(rest): reduce contention in `PooledCurlHandleFactory`

### LINKED ISSUE #9431: Reduce locking contention in handle factory
Under load (the report includes 31,000+ threads trying to use the handle factory) the mutex in the handle factory becomes contended.  We need to reduce contention to support these applications.  The suggestions are:

- Use more fine grained locks
- Reduce the size of the critical sections, perform work that does not need locking outside the locks
- For optional work (e.g., reduce the size of the pool) use try lock and delay the work until the lock is not contended

I think we should refactor the code to use the handle factory in `rest_internal` before making this change, otherwise we would be making the change twice.

- [ ] Refactor `CurlPtr` and friends
- [ ] Refactor `CurlHandle` and `CurlHandleFactory`
- [ ] Use different mutexes to protect variables unrelated by some invariant
- [ ] Release locks as soon as possible
- [ ] Reduce size of critical sections in `CleanupHandle()` and `CleanupMultiHandle()`.

coryan:
Use separate mutexes for each component. Move as much work as possible
outside the critical sections. And finally amortize the cost of
releasing the handles to one thread.

Of these changes the most important seems to be moving the work to
fetch the local IP address outside the critical sections, as this may
involve a DNS lookup. The rest are at least not harmful, and could
help when under pressure.

I added a benchmark.  I ran this to 32,768 simultaneous threads using
the pool. It works and only takes a few milliseconds, but this is not
a number of threads we can safely run in the CI environment.

Fixes #9431

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9457)


scotthart (approved):
> Reviewed 7 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9457)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

Issue #4122, opened 2020-05-14
Run bigtable integration tests against production on nightly builds.


coryan:
Most of the Bigtable integration tests are running against the emulator for both CMake and Bazel. It would be desirable if the `ci/kokoro/docker/integration-nightly` builds ran *all* the tests against production.

coryan:
Note that this would be easy to do now, as the driver scripts use `FORCE_TEST_IN_PRODUCTION` to decide which services use their emulator or run against production.

devjgm:
Still need to do this.

PR #9464, opened 2022-07-13
test(bigtable): fix integration tests for prod

dbolduc:
Part of the work for #4122 

The app profile routing must be supplied in a request or it gets rejected.

The IAM test was sending id's when it should have been sending names.

Replace the sync retry loop to wait for consistency with the actual helper function in our public API.

### DIFF
--- a/google/cloud/bigtable/admin/integration_tests/table_admin_integration_test.cc
+++ b/google/cloud/bigtable/admin/integration_tests/table_admin_integration_test.cc
@@ -16,6 +16,8 @@
 #include "google/cloud/bigtable/admin/bigtable_table_admin_client.h"
 #include "google/cloud/bigtable/resource_names.h"
 #include "google/cloud/bigtable/testing/table_integration_test.h"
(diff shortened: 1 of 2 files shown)


coryan (approved):

google/cloud/bigtable/admin/integration_tests/table_admin_integration_test.cc
```diff
@@ -325,39 +329,14 @@ TEST_F(TableAdminIntegrationTest, WaitForConsistencyCheck) {
...
-      }
-      std::this_thread::sleep_for(backoff.OnCompletion());
-    } while (!retry.IsExhausted());
-    return status;
-  };
-
-  // Verify that our clusters are eventually consistent.
-  auto result = wait_for_consistency();
-  ASSERT_STATUS_OK(result);
+  // Verify that our clusters are eventually consistent. This calls
```
coryan, line 332: nit: we should not verifying anything about the clusters, that is the Bigtable service team job.  We are verifying that our functions complete under the assumption that Bigtable works.

PR #9475, opened 2022-07-14
test: stricter routing metadata validation

dbolduc:
The main change is to check for (unordered) equality of the routing header against our expectations. Previously, we were checking that the actual header is a superset of the expected headers.

We skip validation of the routing header for any client-side streams.

We do some light refactoring/renaming of method names vs method descriptors.

Then we update two bigtable tests to set realistic resource values, so the metadata is extracted properly. (Our tests were setting the table name to `""`, which is not something the library will ever do.)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9475)

### DIFF
--- a/google/cloud/testing_util/validate_metadata.cc
+++ b/google/cloud/testing_util/validate_metadata.cc
@@ -42,11 +42,14 @@ namespace testing_util {
 
 namespace {
 
+using ::google::protobuf::DescriptorPool;
 using ::testing::Contains;
 using ::testing::IsEmpty;
+using ::testing::Matcher;
 using ::testing::Not;
 using ::testing::NotNull;
 using ::testing::Pair;
+using ::testing::UnorderedElementsAreArray;
 
 using RoutingHeaders = std::map<std::string, std::string>;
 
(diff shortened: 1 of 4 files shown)


google/cloud/testing_util/validate_metadata.cc
```diff
@@ -315,14 +314,30 @@ void ValidateMetadataFixture::IsContextMDValid(
...
+  }
+
   // Check if the metadata in the context satisfied the expectations.
+  std::vector<Matcher<std::pair<std::string, std::string>>> matchers;
+  matchers.reserve(expected.size());
   for (auto const& param : expected) {
-    EXPECT_THAT(actual, Contains(Pair(param.first, MatchesGlob(param.second))));
+    matchers.push_back(Pair(param.first, MatchesGlob(param.second)));
   }
+  EXPECT_THAT(actual, UnorderedElementsAreArray(matchers));
```
devbww, line 340: This seems like something that should be expressible directly in gmock, not that I know how to do so off the top of my head.

dbolduc: Hmm. It may be `UnorderedPointwise(...)`. Let me see if it works.

devbww: Did you have any luck?  I don't want to hold you up unnecessarily.

devbww: After chatting, LGTM.

devbww (approved):

google/cloud/testing_util/validate_metadata.cc
dbolduc: To summarize, this approach with `UnorderedPointwise` works: https://github.com/dbolduc/google-cloud-cpp/commit/bbc062015c88e0a32f682acc076d4c00e06b7bc6

I think the code is cleaner. But I think the error messages are less helpful.

I could not find a good way to generate error messages using `UnorderedPointwise` (trying things like `ExplainMatchResult` outputting to `*result_listener`).

I will leave the PR as is because there are other things I could be doing instead of trying to achieve "gMock correctness".

PR #9484, opened 2022-07-15
impl(storage): helper class to generate invocation ids

coryan:
We will be adding telemetry to GCS requests, where all the retry
attempts for the same application call will share an "invocation id".
This will be included in the `x-goog-api-client` header, under the
`gccl-invocation-id/` field. The design doc calls for a very specific
format for these ids, which this helper class generates.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9484)


dbolduc (approved):

google/cloud/storage/internal/invocation_id_generator.cc
```diff
@@ -0,0 +1,67 @@
...
+  auto constexpr kVersion = static_cast<std::uint8_t>(4 << 4);
+  auto constexpr kVersionMask = static_cast<std::uint8_t>(0b0000'1111);
+  auto constexpr kVariant = static_cast<std::uint8_t>(1 << 5);
+  auto constexpr kVariantMask = static_cast<std::uint8_t>(0b0001'1111);
+
+  auto o = [this]() {
+    auto constexpr kIdBitCount = 128;
+    auto constexpr kArraySize = kIdBitCount / 8;
+    std::array<std::uint8_t, kArraySize> buf;
+    std::uniform_int_distribution<std::uint8_t> d(0, 255);
```
devbww, line 45: I'm seeing a static assertion failure with a platform of interest.
```
    static_assert(__libcpp_random_is_valid_inttype<_IntType>::value, "IntType must be an integer type larger than char");
```

devbww: For what it's worth, cppreference says ...
_"IntType - The result type generated by the generator. The effect is undefined if this is not one of short, int, long, long long, unsigned short, unsigned int, unsigned long, or unsigned long long."_

coryan: Blegh.  Okay, we can use `int` and cast to the desired result.  The output should be in range.  I can send a PR, but I need to dig myself from all the upgrade PRs that are breaking things right now.

coryan: See #9509. Let's see what the builds say.

PR #9494, opened 2022-07-15
feat(spanner): set UpdateDatabaseDdlRequest.operation_id

devbww:
Set `UpdateDatabaseDdlRequest.operation_id` (if not already set) to a
client-generated random operation ID.

The documentation says, "Specifying an explicit operation ID simplifies
determining whether the statements were executed in the event that the
UpdateDatabaseDdl() call is replayed, or the return value is otherwise
lost."

It was recommended that we do this as part of understanding #4758.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9494)


generator/integration_tests/golden/golden_thing_admin_client.cc
```diff
@@ -21,6 +21,7 @@
 #include <memory>
 #include "generator/integration_tests/golden/golden_thing_admin_options.h"
 #include <thread>
+#include "google/cloud/internal/operation_id.h"
```
coryan, line 24: nit: mm... I guess `clang-format` is disabled in this directory, but that sounds like an unfortunate order

generator/internal/client_generator.cc
```diff
@@ -339,6 +339,9 @@ Status ClientGenerator::GenerateCc() {
     CcLocalIncludes({vars("options_header_path")});
     CcSystemIncludes({"thread"});
   }
+  if (HasUpdateDatabaseDdlMethod()) {
```
coryan, line 342: Ah, this is where we would fix the `#include` ordering if needed.  For a separate PR: do we understand why `#include <thread>` is needed?

google/cloud/internal/operation_id.cc
```diff
@@ -0,0 +1,51 @@
...
+#include "google/cloud/internal/random.h"
+#include "absl/strings/ascii.h"
+#include <mutex>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+std::string OperationId(std::string const& prefix) {
```
coryan, line 25: FYI: I have a very similar thing in the works, see #9484.

coryan (approved):
Consider this instead. Write some helper functions, to help applications initialize these protos, something like:

```cc
namespace google::cloud::spanner {

google::spanner::admin::database::v1::UpdateDatabaseDdlRequest
WithUniqueId(
   google::spanner::admin::database::v1::UpdateDatabaseDdlRequest request) {
  request.set_operation_id(
    google::cloud::internal::OperationId("UpdateDatabaseDdl"));
  return request;
}

google::spanner::admin::database::v1::UpdateDatabaseDdlRequest
MakeUpdatedDdlRequest(
  std::string const& database, std::vector<std::string> const& statements) {
  google::spanner::admin::database::v1::UpdateDatabaseDdlRequest request;
  request.set_database(database);
  *request.mutable_statements() = {statements.begin(), statements.end()};
  return WithUniqueId(std::move(request);
}
```

That reduces the number of service-specific changes in the generator, makes it possible to write tests with and without a unique operation id.  But it does require better examples / documentation, and passes some of the responsibility to the application developers.

Also, I think this makes the `UpdatedDatabaseDdl()` operation idempotent when the `operation_id` is non-empty.  Should we add a idempotency policy that provides this functionality?  That also seems like hand-crafted code.

generator/integration_tests/golden/golden_thing_admin_client.cc
devbww: Right ... we do not format the golden files.

But if we're going to rely on a post formatting stage at all in order to make generation simpler/easier, then I would assert that this falls into that category.  If we want to complicate the generator, then we should get rid of post formatting altogether.

generator/internal/client_generator.cc
devbww: `#include <thread>` is needed for the `std::this_thread::sleep_for()` call in the `SetIamPolicy()` OCC-loop overload.

devbww:
> Consider this instead.

If setting the `operation_id` is _"**the** right thing"_ to do, then I think it should be unconditionally handled by the generated code.

If it is not _"**the** right thing"_, then I don't think the generated code should do anything (except allow the user to set it, which it already does).

So, given that I don't know what the right thing is, I'm going to close this until the dust clears.

devbww:
> Also, I think this makes the UpdatedDatabaseDdl() operation idempotent when the operation_id is non-empty.

I forgot to add that adding our own replays would seem to admit the very behavior we're questioning.

PR #9500, opened 2022-07-18
doc(pubsub): create BigQuery subscription sample

kamalaboulhosn:
Note that this was not added to the auto samples run. The reason is that it requires a BigQuery table to exist or to be created. It looks like there isn't library support for creating tables in C++. If there is a way to do it and it is preferred that this sample be added to the auto-run samples, I can do that.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9500)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -340,6 +340,32 @@ void CreatePushSubscription(
   (std::move(client), argv.at(0), argv.at(1), argv.at(2), argv.at(3));
 }
 
+void CreateBigQuerySubscription(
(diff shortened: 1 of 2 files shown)


coryan:
> For anyone downloading and running the samples, they probably won't have access to the table you created and would instead need to create a table themselves.

That is true for nearly all our examples.  The application would need to provide a project id and table id, the question was really whether all our CI builds (I should have been more explicit, we have over 50 builds that run for each PR, most of them using the emulator) needed to have separate tables.  It looks like a single table will work for our CI.  I will just do that.

PR #9519, opened 2022-07-19
impl(rest): improve write performance

scotthart:
Using recorded perf statistics, attempt to improve write performance by making inner loops less cpu intensive on the success paths and mimicking libcurl's use of curl_multi_poll with a longer timeout.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9519)

### DIFF
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -776,10 +776,12 @@ Status CurlImpl::MakeRequest(CurlImpl::HttpMethod method,
     if (!status.ok()) return OnTransferError(std::move(status));
     status = handle_.SetOption(CURLOPT_UPLOAD, 1L);
     if (!status.ok()) return OnTransferError(std::move(status));
+#if CURL_AT_LEAST_VERSION(7, 51, 0)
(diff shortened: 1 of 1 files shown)


google/cloud/internal/curl_impl.cc
```diff
@@ -528,49 +532,40 @@ StatusOr<int> CurlImpl::PerformWork() {
...
 }
 
 Status CurlImpl::WaitForHandles(int& repeats) {
-  int const timeout_ms = 1;
-  std::chrono::milliseconds const timeout(timeout_ms);
+  int const timeout_ms = 1000;
   int numfds = 0;
   CURLMcode result =
-      curl_multi_wait(multi_.get(), nullptr, 0, timeout_ms, &numfds);
+      curl_multi_poll(multi_.get(), nullptr, 0, timeout_ms, &numfds);
```
coryan, line 564: `curl_multi_poll()` was added in CURL 7.66.0, and we currently say we support CURL >= 7.47.0 (in fact we test with 7.29.0, for RHEL:7 / CentOS:7)

PR #9524, opened 2022-07-19
fix(generator): streaming writes with routing instructions

coryan:
The generator cannot create code to capture the routing instructions on
asynchronous streaming writes.  The generated functions do not have a
request to extract the routing instructions from.  Just inject the
basic metadata, the caller is responsible for adding the routing
instructions.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9524)

### DIFF
--- a/generator/internal/metadata_decorator_generator.cc
+++ b/generator/internal/metadata_decorator_generator.cc
@@ -316,6 +316,10 @@ $metadata_class_name$::$method_name$(
       continue;
     }
     if (IsBidirStreaming(method)) {
+      // Asynchronous streaming writes do not consume a request. Typically, the
+      // first `Write()` call contains any relevant data to "start" the stream.
+      // Thus, the decorator cannot add any routing instructions. The caller
+      // should initialize `context` with any such instructions.
       CcPrintMethod(method, __FILE__, __LINE__,
                     R"""(
 std::unique_ptr<::google::cloud::AsyncStreamingReadWriteRpc<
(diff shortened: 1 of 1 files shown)


generator/internal/metadata_decorator_generator.cc
```diff
@@ -397,16 +397,21 @@ std::unique_ptr<::google::cloud::internal::AsyncStreamingReadRpc<
...
           R"""(
 std::unique_ptr<::google::cloud::internal::AsyncStreamingWriteRpc<
     $request_type$, $response_type$>>
 $metadata_class_name$::Async$method_name$(
     google::cloud::CompletionQueue const& cq,
     std::unique_ptr<grpc::ClientContext> context) {
-)""",
-          SetMetadataText(method, kPointer),
-          R"""(
+  SetMetadata(*context);
```
dbolduc, line 414: hm, so the comment you wrote also applies to async streaming read-writes. It seems like we should either copy the comment, or have both use `SetMetadataText(method, kPointer)`. Then in that method we would say like:
```cc
std::string SetMetadataText(...) {
  std::string const context = ...;

  // blah blah blah
  // comment about client streaming
  // blah blah blah
  if (method.client_streaming()) return "  SetMetadata(" + context + ");";
  ...
}
```

generator/internal/metadata_decorator_generator.cc
```diff
@@ -397,16 +397,21 @@ std::unique_ptr<::google::cloud::internal::AsyncStreamingReadRpc<
       continue;
     }
     if (IsStreamingWrite(method)) {
+      // Asynchronous streaming writes do not consume a request, typically the
```
dbolduc, line 400: s/, typically/. Typically/

generator/internal/metadata_decorator_generator.cc
```diff
@@ -397,16 +397,21 @@ std::unique_ptr<::google::cloud::internal::AsyncStreamingReadRpc<
...
+      // first `Write()` call contains any relevant data to "start" the stream.
+      // Thus, the decorator cannot add any routing instructions. The caller
+      // should initialize `context` with any such instructions. At this time,
```
dbolduc, line 403: I think I would just say: "The handwritten caller should initialize \`context\` with any such instructions." and delete the rest of the paragraph.

If you decide to keep the rest of the paragraph then, `s/, ut/, but/` a few lines down.

generator/internal/metadata_decorator_generator.cc
coryan: I chose to copy the comment

generator/internal/metadata_decorator_generator.cc
coryan: Fixed.

generator/internal/metadata_decorator_generator.cc
coryan: Option 1

dbolduc (approved):

PR #9553, opened 2022-07-22
cleanup: use relative links in CHANGELOG

coryan:
Relative links respect the current branch or tag, and they are easier to
write too.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9553)


dbolduc (commented):
might be a good idea to change this too: https://github.com/googleapis/google-cloud-cpp/blob/ee5513c79bb646cbb4cb4e1163dbb30f18c23bdb/release/changes.sh#L93

dbolduc (approved):

PR #9556, opened 2022-07-22
refactor: split streaming write RPC implementation

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9556)


google/cloud/internal/streaming_write_rpc_impl.h
```diff
@@ -0,0 +1,121 @@
...
+#include "google/cloud/internal/streaming_write_rpc.h"
+#include "google/cloud/status.h"
+#include <grpcpp/grpcpp.h>
+#include <grpcpp/support/sync_stream.h>
+#include <memory>
+#include <utility>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
```
devbww, line 28: `#include "google/cloud/version.h"`

google/cloud/internal/streaming_write_rpc_impl.h
```diff
@@ -0,0 +1,121 @@
...
+namespace internal {
+
+/// Report the errors in a standalone function to minimize includes
```
devbww, line 31: I'm not sure what "in a standalone function to minimize includes" is supposed to convey.  I might instead emphasis that this a `Status` we can only surface via logging.

devbww (approved):

google/cloud/internal/streaming_write_rpc_impl.h
coryan: Done

google/cloud/internal/streaming_write_rpc_impl.h
coryan: Done

Issue #5782, opened 2021-01-30
The google-cloud-cpp::*-mocks targets are not exported


coryan:
These targets are not exported.  In practice it does not matter much, because using the library gets you the same `-I` flags, but the applications are missing the `-lgtest*` and `-lgmock*` options.

coryan:
We need to revisit this in 2022-06 when we will drop CMake 3.5.

coryan:
I will try to do this on my next operator week.

coryan:
Soon (2023-05-01) we will drop CMake 3.10 and I think this gets easier.

dbolduc:
Only `compute` remains. We probably want to install all `compute_<resource>_mocks` in a `google_cloud_cpp_compute_mocks` module.

If we want to use the helper, we would need to generalize it.

Unassigning because my operator shift is over.

scotthart:
We need to finish out compute.

PR #9568, opened 2022-07-26
fix(bigtable): the mocks library is no longer header-only

coryan:
The library is no longer header-only, it needs to become a real library.

Part of the work for #5782

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9568)


dbolduc (approved):
Thanks

Issue #9469, opened 2022-07-13
Connection pool release count should adjust based on outstanding connections


coryan:
Regarding #9457 one of our customers points out:

> The PR generally looks good to me, the only concern I have is only keeping half of the max pool size when the pool is full, which sounds a bit aggressive. Would it be better to keep something like 90% of the max pool size?

I suspect there is no perfect answer here, but certainly we could do better than we are doing right now.

When there are many threads using connections compared to the size of the pool you probably want to remove more of the pool to make room for the connections that will return shortly after (newer connections are "better").  But when there are fewer threads using connections you want to release as little as possible.

I thought about adding a configuration parameter, but that does not work because the workload changes over time.  I think we need to track of all outstanding connections and return more when the ratio of outstanding connections to the pool size is high, and return less when the ratio is lower.  That requires an API change, because the pool currently does not know about connections that had an error (they are never returned to the pool), they should now be returned with a flag saying "I am not using this, and it is bad, so just release it and updated your counters"

PR #9575, opened 2022-07-27
impl(rest): always discard handles via HandleFactory

coryan:
When a CURL* handle has an error we don't want to return it to the pool.
The error probably indicates a bad server (in the load balancer) or a
bad connection, and it is time to set up a new connection. Previously,
we just discarded the handle, how let the factory do so. In a future
PR we will use this to update counters in the factory, and adjust how
aggressively the pool releases its handles.

Part of the work for #9469

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9575)

### DIFF
--- a/google/cloud/internal/curl_handle.cc
+++ b/google/cloud/internal/curl_handle.cc
@@ -135,13 +135,13 @@ CurlHandle CurlHandle::MakeFromPool(CurlHandleFactory& factory) {
 void CurlHandle::ReturnToPool(CurlHandleFactory& factory, CurlHandle h) {
   CurlPtr tmp(nullptr, curl_easy_cleanup);
   h.handle_.swap(tmp);
-  factory.CleanupHandle(std::move(tmp), CurlHandleFactory::kKeep);
+  factory.CleanupHandle(std::move(tmp), HandleDisposition::kKeep);
 }
 
 void CurlHandle::DiscardFromPool(CurlHandleFactory& factory, CurlHandle h) {
   CurlPtr tmp(nullptr, curl_easy_cleanup);
   h.handle_.swap(tmp);
-  factory.CleanupHandle(std::move(tmp), CurlHandleFactory::kDiscard);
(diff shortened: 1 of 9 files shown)


scotthart (approved):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9575)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/internal/curl_handle_factory.h` line 36 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9575#-N80iGHaATIlg-YVVtZn:-N80iGHaATIlg-YVVtZo:b54wzyn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/eb6ce9a17fd1350ffa058d7425dde6b2cfa66fdd/google/cloud/internal/curl_handle_factory.h#L36)):*
> ```objc
> class CurlHandleFactory {
>  public:
>   enum HandleDisposition : int {
> ```

Why not an `enum class`?

coryan (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9575)* status: all files reviewed, 1 unresolved discussion (waiting on @scotthart)

___
*[`google/cloud/internal/curl_handle_factory.h` line 36 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9575#-N80iGHaATIlg-YVVtZn:-N80m-Oq5pEykNiVFhy-:b-rmt8hw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/eb6ce9a17fd1350ffa058d7425dde6b2cfa66fdd/google/cloud/internal/curl_handle_factory.h#L36)):*

Why not an `enum class`?
</blockquote></details>

And why the silly `: int`.  Done, but I moved it out of the class because my `:` key is only rated for so many key presses.

PR #9576, opened 2022-07-27
feat(GCS+gRPC): more efficient `WriteObject()` stall timeouts

### LINKED ISSUE #9558: Implement stall timeouts using watchdog timers
The current implementation using the asynchronous APIs. This moves the computation (e.g. decrypting the incoming data) to the I/O threads.  This seems to bottleneck reads and writes.  

Using a watchdog timer that cancels the stream on a timeout seems to perform better.

- [x] `ReadObject()`
- [x] `InsertObject()`
- [x] `WriteObject()`

coryan:
This changes the implementation of `WriteObject()` to use the blocking
API and a watchdog timer to implement stall timeouts. If the timer
expires before the `Write()` operation completes a background thread
cancels the RPC.  If the operation completes, we cancel the timer.

Fixes #9558

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9576)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -91,15 +91,6 @@ StatusOr<ObjectAccessControl> FindDefaultObjectAccessControl(
                                            "> in object id " + response->id());
 }
 
-std::chrono::milliseconds DefaultTransferStallTimeout(
-    std::chrono::milliseconds value) {
-  if (value != std::chrono::milliseconds(0)) return value;
-  // We need a large value for `wait_for()`, but not so large that it can easily
-  // overflow.  Fortunately, uploads automatically cancel (server side) after
-  // 7 days, so waiting for 14 days will not create spurious timeouts.
-  return std::chrono::milliseconds(std::chrono::hours(24) * 14);
-}
-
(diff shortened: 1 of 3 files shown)


google/cloud/storage/internal/grpc_client.cc
```diff
@@ -129,39 +120,34 @@ Status TimeoutError(std::chrono::milliseconds timeout, std::string const& op) {
...
+      if (!f.get()) return false;
       writer->Cancel();
-      pending.then(WaitForIdle{std::move(writer)});
+      return true;
+    });
+    (void)writer->Write(google::storage::v2::WriteObjectRequest{},
+                        grpc::WriteOptions().set_last_message());
+    watchdog.cancel();
+    if (watchdog.get()) {
+      writer->Close();
```
dbolduc, line 139: do we need another watchdog that might cancel this `Close()` operation? we use one a few lines below here.

coryan: No.  If we get to this point the streaming RPC is already canceled.  Setting up another watchdog, just to cancel the request again seems unlikely to speed up things.

dbolduc (approved):

PR #9579, opened 2022-07-28
test(storage): use shorter random object names

coryan:
The computers do not care, but it was hard to inspect the output with
the very long names.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9579)

### DIFF
--- a/google/cloud/storage/benchmarks/benchmark_utils.cc
+++ b/google/cloud/storage/benchmarks/benchmark_utils.cc
@@ -149,6 +149,15 @@ std::string MakeRandomBucketName(google::cloud::internal::DefaultPRNG& gen) {
   return storage::testing::MakeRandomBucketName(gen, RandomBucketPrefix());
 }
 
+std::string MakeRandomObjectName(google::cloud::internal::DefaultPRNG& gen) {
+  auto constexpr kObjectNameLength = 32;
+  return google::cloud::internal::Sample(gen, kObjectNameLength,
+                                         "abcdefghijklmnopqrstuvwxyz"
+                                         "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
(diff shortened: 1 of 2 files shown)


google/cloud/storage/benchmarks/benchmark_utils.cc
```diff
@@ -149,6 +149,15 @@ std::string MakeRandomBucketName(google::cloud::internal::DefaultPRNG& gen) {
   return storage::testing::MakeRandomBucketName(gen, RandomBucketPrefix());
 }
 
+std::string MakeRandomObjectName(google::cloud::internal::DefaultPRNG& gen) {
+  auto constexpr kObjectNameLength = 32;
+  return google::cloud::internal::Sample(gen, kObjectNameLength,
+                                         "abcdefghijklmnopqrstuvwxyz"
+                                         "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                                         "0123456789") +
+         ".txt";
```
dbolduc, line 158: wouldn't this be a file name?

dbolduc (approved):

google/cloud/storage/benchmarks/benchmark_utils.cc
coryan: I am not sure what you are asking.  But there are no files in GCS.  Just buckets and objects.

dbolduc: I was just curious why there is a `".txt"` on the end.

Why not use: https://github.com/googleapis/google-cloud-cpp/blob/e2dd43461df4eb213e827653ca2c5b50657a8b0f/google/cloud/storage/testing/random_names.cc#L48-L58

PR #9582, opened 2022-07-28
impl(rest): tune HandleFactory cleanup

### LINKED ISSUE #9469: Connection pool release count should adjust based on outstanding connections

coryan:
Before this change, the pool would always remove half the handles to
make room for additional handles returning to the pool. We want to
remove fewer handles when there are few excess handles, and remove
more handles (but not all handles) when there are more excess handles.

Fixes #9469 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9582)

### DIFF
--- a/google/cloud/internal/curl_handle_factory.cc
+++ b/google/cloud/internal/curl_handle_factory.cc
@@ -84,11 +84,7 @@ void DefaultCurlHandleFactory::SetCurlOptions(CURL* handle) {
 
 PooledCurlHandleFactory::PooledCurlHandleFactory(std::size_t maximum_size,
                                                  Options const& o)
-    : maximum_size_(maximum_size),
-      cainfo_(CAInfo(o)),
-      capath_(CAPath(o)),
-      active_handles_(0),
-      active_multi_handles_(0) {}
+    : maximum_size_(maximum_size), cainfo_(CAInfo(o)), capath_(CAPath(o)) {}
 
 PooledCurlHandleFactory::~PooledCurlHandleFactory() = default;
 
@@ -104,10 +100,10 @@ CurlPtr PooledCurlHandleFactory::CreateHandle() {
     SetCurlOptions(handle.get());
     return handle;
   }
+  ++active_handles_;
   lk.unlock();
   auto curl = MakeCurlPtr();
   SetCurlOptions(curl.get());
-  ++active_handles_;
   return curl;
 }
 
@@ -121,6 +117,7 @@ void PooledCurlHandleFactory::CleanupHandle(CurlPtr h, HandleDisposition d) {
     std::unique_lock<std::mutex> lk(last_client_ip_address_mu_);
     last_client_ip_address_ = ip;
   }
+  std::unique_lock<std::mutex> lk(handles_mu_);
   if (d == HandleDisposition::kDiscard) {
     --active_handles_;
     return;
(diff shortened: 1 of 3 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9582)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

google/cloud/internal/curl_handle_factory.h
```diff
@@ -154,6 +155,8 @@ class PooledCurlHandleFactory : public CurlHandleFactory {
   std::deque<CurlMulti> multi_handles_;
   mutable std::mutex last_client_ip_address_mu_;
   std::string last_client_ip_address_;
+  std::atomic<std::size_t> active_handles_;
+  std::atomic<std::size_t> active_multi_handles_;
```
devbww, line 159: Opt: It looks like these are manipulated within, or very close to regions already protected by their matching mutexes, so you may want to take advantage of that to remove the need for `std::atomic`.

coryan: Done. And fixed a bug in the process.

google/cloud/internal/curl_handle_factory.cc
```diff
@@ -116,22 +117,38 @@ void PooledCurlHandleFactory::CleanupHandle(CurlPtr h, HandleDisposition d) {
...
   }
-  if (d == HandleDisposition::kDiscard) return;
+  std::unique_lock<std::mutex> lk(handles_mu_);
+  if (d == HandleDisposition::kDiscard) {
+    --active_handles_;
+    return;
+  }
   // Use a temporary data structure to release any excess handles *after* the
   // lock is released.
   std::vector<CurlPtr> released;
```
devbww, line 127: It looks like you'll now want to move this declaration up too.

google/cloud/internal/curl_handle_factory.h
```diff
@@ -19,6 +19,7 @@
 #include "google/cloud/options.h"
 #include "google/cloud/version.h"
 #include "absl/types/optional.h"
+#include <atomic>
```
devbww, line 22: Remove now?

google/cloud/internal/curl_handle_factory.cc
coryan: Doh! Fixed.

google/cloud/internal/curl_handle_factory.h
coryan: Done.

devbww (approved):

PR #9583, opened 2022-07-28
doc(bigtable): document connection pooling model

### LINKED ISSUE #1553: Document the connection pooling model in the Bigtable library.
We should give guidance to our users about how to get more connections, how are they used, etc. This is important for them to know how to get the most out of the library.

dbolduc:
Fixes #1553

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9583)


coryan (approved):
Consider using the active voice, e.g. "If this option is not set, the class creates between 1 and 64 channels." 

Also consider being explicit about what is subject to change, e.g., "... between 1 and 64 channels. The specific algorithm and number of channels is an implementation detail, and subject to change without notice."

PR #9617, opened 2022-08-02
impl: check if options are empty

dbolduc:
Part of the work for #7688 

Add an internal helper to see if Options are empty or not.

There are two code paths in the `bigtable::Table` client. One that uses `client_` and one that uses `connection_`. Per operation options only apply to the path that uses the `connection_`. I do not want to check if `opts.has<T>` for each `T` option. So, we just check to see if the options are empty.

The code in bigtable will look vaguely like this:

```cc
Status Table::Foo(Options opts) {
  if (!connection_ && !google::cloud::internal::IsEmpty(opts)) {
    return Status(StatusCode::kFailedPrecondition,
                   "Per-operation options only apply to `Table`s constructed "
                   "with a `DataConnection`.");
  }
  // etc...
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9617)

### DIFF
--- a/google/cloud/options.h
+++ b/google/cloud/options.h
@@ -33,6 +33,7 @@ namespace internal {
 Options MergeOptions(Options, Options);
 void CheckExpectedOptionsImpl(std::set<std::type_index> const&, Options const&,
                               char const*);
+// TODO(#8800) - Remove when bigtable::Table no longer uses bigtable::DataClient
 bool IsEmpty(Options const&);
 template <typename T>
 inline T const& DefaultValue() {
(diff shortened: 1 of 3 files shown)


google/cloud/options.h
```diff
@@ -209,6 +210,7 @@ class Options {
   friend Options internal::MergeOptions(Options, Options);
   friend void internal::CheckExpectedOptionsImpl(
       std::set<std::type_index> const&, Options const&, char const*);
+  friend bool internal::IsEmpty(Options const&);
```
devbww, line 213: Did you consider just an `Options::empty()`?

dbolduc: I don't want to grow the public API. Eventually, I want to remove this function. I should probably attach an issue to it, to help me remember my intentions...

devbww (approved):

PR #9622, opened 2022-08-02
test(storage): improvements in upload benchmark

coryan:
I want to use the `aggregate_upload_throughput` benchmark to compare
gRPC vs. DirectPath vs. REST with different timeout parameters and
other options. It seemed simpler to add a `google::cloud::Options` to
the benchmark configuration, that will require fewer changes in the
future, and simplifies (I think) the client initialization.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9622)

### DIFF
--- a/google/cloud/storage/benchmarks/aggregate_upload_throughput_benchmark.cc
+++ b/google/cloud/storage/benchmarks/aggregate_upload_throughput_benchmark.cc
@@ -32,10 +32,8 @@ using ::google::cloud::storage_experimental::DefaultGrpcClient;
 using ::google::cloud::testing_util::FormatSize;
 using ::google::cloud::testing_util::Timer;
 namespace gcs = ::google::cloud::storage;
-namespace gcs_experimental = ::google::cloud::storage_experimental;
 namespace gcs_bm = ::google::cloud::storage_benchmarks;
 using gcs_bm::AggregateUploadThroughputOptions;
-using gcs_bm::ApiName;
 using gcs_bm::FormatBandwidthGbPerSecond;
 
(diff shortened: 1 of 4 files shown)


google/cloud/storage/benchmarks/aggregate_upload_throughput_benchmark.cc
```diff
@@ -298,6 +285,22 @@ int main(int argc, char* argv[]) {
...
+  auto p = headers.find(":grpc-context-peer");
+  if (p == headers.end()) {
+    p = headers.find(":curl-peer");
+  }
+  return p == headers.end() ? std::string{"[peer-unknown]"} : p->second;
+}
+
+std::string ExtractUploadId(std::string v) {
+  auto constexpr kRestField = "upload_id=";
+  auto const pos = v.find(kRestField);
```
devbww, line 299: Q.  Should this be more like a `StartsWith`/`ConsumePrefix`?

devbww (approved):

google/cloud/storage/benchmarks/aggregate_upload_throughput_benchmark.cc
coryan: There is a lot of (unfortunately variable) stuff before `upload_id=` that I also need to remove, I do not think `ConsumePrefix()` works in this case, and this seems simple enough.

devbww: Ack.  I was confused that if there is stuff before the `upload_id=`, why there might not be stuff after the value too.  But I'm only assuming so.

PR #9623, opened 2022-08-02
feat(bigtable): per-operation Options (pt. 1)

dbolduc:
Most of the work for #7688 

This PR handles all calls except for `Table::(Async)ReadModifyWriteRow(...)`, which are a little bit more involved.

If `Options` are supplied to a `Table` implemented by `DataClient`, we will immediately return a `kInvalidArgument` error.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9623)

### DIFF
--- a/google/cloud/bigtable/data_connection.cc
+++ b/google/cloud/bigtable/data_connection.cc
@@ -25,21 +25,23 @@
 
 namespace google {
 namespace cloud {
-namespace bigtable {
+namespace bigtable_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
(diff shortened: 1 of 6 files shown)


coryan (changes requested):
FWIW, I think there is some authoritative document describing what error code to use when, but I cannot recall its location right now (the gRPC github repo has a copy I think).

devbww (approved):
Don't forget to update the status code in the description.

dbolduc:
> Don't forget to update the status code in the description.

Thanks. At no point did updating the PR description cross my mind.

PR #9636, opened 2022-08-05
cleanup(bigtable): default options like the generator does

dbolduc:
Follow the lead of the generator and:
1. Remove `bigtable_internal::MakeDataConnection(...)` (which I wasn't even using).
2. Default the `Options` exactly once - in `bigtable::MakeDataConnection(...)`. Adjust tests accordingly.

I also modified a unit test to check that `Options` are defaulted by `MakeDataConnection`. Ideally we wouldn't construct a `DataConnection` in the test, we would just verify that defauted options are passed to it... but that is ok. If we didn't have this test and we forgot to default the options [here](https://github.com/googleapis/google-cloud-cpp/blob/e14aed697ecc8924d4d2fafeea45fefd152065eb/google/cloud/bigtable/data_connection.cc#L145), we would only learn about it from an integration test segfaulting.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9636)

### DIFF
--- a/google/cloud/bigtable/internal/data_connection_impl_test.cc
+++ b/google/cloud/bigtable/internal/data_connection_impl_test.cc
@@ -22,6 +22,7 @@
 #include "google/cloud/grpc_options.h"
 #include "google/cloud/internal/async_streaming_read_rpc_impl.h"
 #include "google/cloud/testing_util/mock_backoff_policy.h"
+#include "google/cloud/testing_util/scoped_environment.h"
 #include "google/cloud/testing_util/status_matchers.h"
 #include <google/protobuf/text_format.h>
 #include <gmock/gmock.h>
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/internal/data_connection_impl_test.cc
```diff
@@ -1636,6 +1631,16 @@ TEST(DataConnectionTest, AsyncReadRowFailure) {
   EXPECT_THAT(resp, StatusIs(StatusCode::kPermissionDenied));
 }
 
+TEST(MakeDataConnection, DefaultsOptions) {
+  auto conn = bigtable::MakeDataConnection(
```
devbww, line 1635: > It makes an actual connection to the service

Might it be possible to move it to some place marked as an integration test?

dbolduc: I want to check that the options are defaulted by `MakeDataConnection()` in a unit test.

I think my comment was inaccurate, and it reveals that I don't know what happens when a gRPC Stub is created. The test does not depend on the presence of a service/emulator i.e. it would pass even if I set the `EndpointOption` to some nonsense.

google/cloud/bigtable/internal/data_connection_impl_test.cc
```diff
@@ -1636,6 +1631,16 @@ TEST(DataConnectionTest, AsyncReadRowFailure) {
...
+      Options{}
+          .set<UnifiedCredentialsOption>(MakeInsecureCredentials())
+          .set<GrpcNumChannelsOption>(1));
+  auto options = conn->options();
+  EXPECT_EQ(options.get<AuthorityOption>(), "bigtable.googleapis.com");
+  EXPECT_EQ(options.get<GrpcNumChannelsOption>(), 1);
```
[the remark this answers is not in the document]
dbolduc: We check that the unset `AuthorityOption` gets defaulted to the right value. Then I set `GrpcNumChannelOption` to 1 so that:
* we can verify that a user supplied option is not overridden by the defaults
* the minimum number of stubs are created

google/cloud/bigtable/internal/data_connection_impl_test.cc
dbolduc: Ack on the fishiness. I think the value added by the unit test is worth the fishiness.

> something that would at least stay on the same host?

I followed the example of [this test](https://github.com/googleapis/google-cloud-cpp/blob/df5eb3727a6a466d2073f7683766f89e1c4da3af/google/cloud/spanner/internal/spanner_stub_test.cc#L46). I think that is what you are suggesting?

devbww (approved):
Thanks.

Issue #9168, opened 2022-06-06
Consider dropping or simplifying `ConnectionImpl::*_policy()` getters


dbolduc:
We do not need the `else` branch that uses the connection's `options_` in this code:
https://github.com/googleapis/google-cloud-cpp/blob/292265193edcbe6737cde4053145610a5f3e267d/google/cloud/tpu/internal/tpu_connection_impl.h#L89-L119

These policy options are defaulted in the connection, and passed back to the client. This means that any `OptionsSpan` the client creates will have the policy options we should use present.

We may need to update some unit tests for the `ConnectionImpl`s which do not have a client present to create such a span.

devbww:
> We may need to update some unit tests for the `ConnectionImpl`s which do not have a client present to create such a span.

For the record, here is the comment that describes the current behavior of those tests.

https://github.com/googleapis/google-cloud-cpp/blob/292265193edcbe6737cde4053145610a5f3e267d/generator/internal/connection_impl_generator.cc#L113-L117

coryan:
@dbolduc thinks there is not enough value. Closing.

coryan:
This is now fixed.

PR #9653, opened 2022-08-08
refactor(bigtable): `InstanceAdmin` receives Connection `Options`

dbolduc:
Part of the work for #9168 

With this change, we copy options from the connection to the legacy client upon construction. These options make it into an `OptionSpan` for each legacy client call.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9653)

### DIFF
--- a/google/cloud/bigtable/instance_admin_test.cc
+++ b/google/cloud/bigtable/instance_admin_test.cc
@@ -31,8 +31,8 @@ class InstanceAdminTester {
     return admin.connection_;
   }
 
-  static Options Policies(bigtable::InstanceAdmin const& admin) {
-    return admin.policies_;
+  static Options Opts(bigtable::InstanceAdmin const& admin) {
+    return admin.options_;
   }
 };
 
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/instance_admin_test.cc
```diff
@@ -31,8 +31,8 @@ class InstanceAdminTester {
     return admin.connection_;
   }
 
-  static Options Policies(bigtable::InstanceAdmin const& admin) {
-    return admin.policies_;
+  static Options Opts(bigtable::InstanceAdmin const& admin) {
```
devbww, line 34: Can we use `Options` instead of `Opts`?

devbww (approved):

google/cloud/bigtable/instance_admin_test.cc
dbolduc: No :(

PR #9656, opened 2022-08-08
refactor(bigtable): `TableAdmin` receives Connection `Options`

dbolduc:
Part of the work for #9168. This PR unblocks the generator refactor.

Same story as #9653 - merge in options coming from the connection and put them in an `OptionsSpan` for each client call.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9656)

### DIFF
--- a/google/cloud/bigtable/table_admin.h
+++ b/google/cloud/bigtable/table_admin.h
@@ -166,8 +166,10 @@ class TableAdmin {
             DefaultRPCBackoffPolicy(internal::kBigtableTableAdminLimits)),
         polling_prototype_(
             DefaultPollingPolicy(internal::kBigtableTableAdminLimits)),
-        policies_(bigtable_internal::MakeTableAdminOptions(
-            retry_prototype_, backoff_prototype_, polling_prototype_)) {}
+        options_(google::cloud::internal::MergeOptions(
+            bigtable_internal::MakeTableAdminOptions(
+                retry_prototype_, backoff_prototype_, polling_prototype_),
+            connection_->options())) {}
 
   /**
    * Create a new TableAdmin using explicit policies to handle RPC errors.
@@ -212,8 +214,10 @@ class TableAdmin {
         polling_prototype_(
             DefaultPollingPolicy(internal::kBigtableTableAdminLimits)) {
     ChangePolicies(std::forward<Policies>(policies)...);
-    policies_ = bigtable_internal::MakeTableAdminOptions(
-        retry_prototype_, backoff_prototype_, polling_prototype_);
+    options_ = google::cloud::internal::MergeOptions(
+        bigtable_internal::MakeTableAdminOptions(
+            retry_prototype_, backoff_prototype_, polling_prototype_),
+        connection_->options());
   }
 
   TableAdmin(TableAdmin const&) = default;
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/table_admin.h
```diff
@@ -1115,10 +1121,14 @@ class TableAdmin {
   std::string project_id_;
   std::string instance_id_;
   std::string instance_name_;
+  //@{
+  /// These prototypes are only used as temporary storage during construction of
+  /// the class, where they are consolidated as common policies in `options_`.
```
devbww, line 1126: Can't we avoid doing that?

google/cloud/bigtable/table_admin_test.cc
```diff
@@ -56,8 +56,8 @@ class TableAdminTester {
     return admin.background_threads_;
   }
 
-  static Options Policies(bigtable::TableAdmin const& admin) {
-    return admin.policies_;
+  static Options Opts(bigtable::TableAdmin const& admin) {
```
devbww, line 59: Can we use `Options()`?

google/cloud/bigtable/table_admin.h
dbolduc: Probably. But that is not something I am going to do in this PR.

google/cloud/bigtable/table_admin_test.cc
dbolduc: No! :'(

coryan: Why not?  It conflicts with the name?  Can you say:

```cc
static ::google::cloud::Options Options(bigtable::TableAdmin const& admin) ...
```

Or can you rename the function to `ExtractOptions()`?

dbolduc: `::google::cloud::Options Options(...)` works.

I clearly did not put up a good fight against the compiler.

devbww: As should just `cloud::Options`, yes?  If we depend on namespace lookup at all (as we do), then we might as well use it I reckon.

google/cloud/bigtable/table_admin_test.cc
```diff
@@ -207,7 +216,7 @@ TEST_F(TableAdminTest, LegacyConstructorSetsCustomCQ) {
 TEST_F(TableAdminTest, LegacyConstructorDefaultsPolicies) {
   auto admin_client = MakeAdminClient(kProjectId, TestOptions());
   auto admin = TableAdmin(std::move(admin_client), kInstanceId);
-  auto policies = TableAdminTester::Policies(admin);
+  auto policies = TableAdminTester::Opts(admin);
```
devbww, line 219: Should we call these locals `options` too?

dbolduc: Yes, I think so. Fixed.

google/cloud/bigtable/table_admin_test.cc
dbolduc: I used `::google::cloud::Options` for no particular reason.

PTAL

google/cloud/bigtable/table_admin.h
```diff
@@ -166,8 +166,10 @@ class TableAdmin {
             DefaultRPCBackoffPolicy(internal::kBigtableTableAdminLimits)),
         polling_prototype_(
             DefaultPollingPolicy(internal::kBigtableTableAdminLimits)),
-        policies_(bigtable_internal::MakeTableAdminOptions(
-            retry_prototype_, backoff_prototype_, polling_prototype_)) {}
+        options_(google::cloud::internal::MergeOptions(
```
devbww, line 169: Do we need/like qualification beyond `internal::MergeOptions()` (et seq)?

devbww (approved):

google/cloud/bigtable/table_admin.h
dbolduc: `bigtable::internal` is a thing. So I think we need at least `cloud::internal::MergeOptions`.

PR #9669, opened 2022-08-10
feat(spanner): fine-grained access control

devbww:
Introduce `spanner::SessionCreatorRoleOption`, which can be passed to
`spanner::MakeConnection()` to set the `session_template.creator_role`
used when creating sessions for the connection.

Add integration test for creating and listing roles.

Add samples:
 - spanner_add_and_drop_database_roles
 - spanner_read_data_with_database_role
 - spanner_list_database_roles
 - spanner_enable_fine_grained_access

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9669)

### DIFF
--- a/google/cloud/spanner/options.h
+++ b/google/cloud/spanner/options.h
@@ -85,6 +85,14 @@ using SpannerPolicyOptionList =
                SpannerPollingPolicyOption>;
 
 /**
+ * Option for `google::cloud::Options` to set the database role used for
+ * session creation.
+ */
+struct SessionCreatorRoleOption {
+  using Type = std::string;
+};
+
+/**
  * Option for `google::cloud::Options` to set the minimum number of sessions to
  * keep in the pool.
  *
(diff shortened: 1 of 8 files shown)


google/cloud/spanner/options.h
```diff
@@ -84,6 +84,14 @@ using SpannerPolicyOptionList =
     OptionList<spanner::SpannerRetryPolicyOption, SpannerBackoffPolicyOption,
                SpannerPollingPolicyOption>;
 
+/**
+ * Option for `google::cloud::Options` to set the database role used for
+ * session creation.
```
coryan, line 89: Consider a link to the spanner documentation explaining what these what these roles are for.

google/cloud/spanner/samples/samples.cc
```diff
@@ -666,6 +666,110 @@ void GetDatabaseDdl(google::cloud::spanner_admin::DatabaseAdminClient client,
...
+  auto& condition = *binding.mutable_condition();
+  condition.set_expression("resource.name.endsWith(\"/databaseRoles/" +
+                           database_role + "\")");
+  condition.set_title(title);
+
+  auto new_policy =
+      client.SetIamPolicy(database.FullName(), *std::move(policy));
+  if (!new_policy) throw std::runtime_error(new_policy.status().message());
+  std::cout << "Enabled fine-grained access in IAM. New policy has version "
+            << new_policy->version() << "\n";
```
coryan, line 769: I do not think that version is what you think it is.  It is more like a code revision tracker than a state change tracker.  You want something like the `etag` field.

coryan (approved):
Some suggestions, but LGTM as-is.

google/cloud/spanner/options.h
devbww: I'm afraid I'm unaware of, and cannot discover, any released documentation.  We can, of course, add something as it becomes available.

google/cloud/spanner/samples/samples.cc
devbww: I don't think I think it is anything in particular.  It doesn't seem, at least, that the sample implies that it isn't a code-revision tracker.

That said, the shape of the output is defined by the internal "Fine Grained Access Control Samples" document, which is followed by all the other languages, so I'm not sure I'm in a position to change anything.

coryan: You could try to talk the team into changing this example. As it is, it will print either `1` or `3` for the version.

coryan: Please merge as-is though, we can implement (or not) both my suggestions at a later time.

devbww: > You could try to talk the team into changing this example.

I can certainly do that.  I'm not sure what I would write other than your comment verbatim however, so I'm also happy to point you at the doc.  Let me know offline.

> As it is, it will print either 1 or 3 for the version.

I thought it would always print (at least) 3, given that's what we raised the minimum version to above.

PR #9678, opened 2022-08-11
cleanup(pubsublite): nicer integration test

dbolduc:
Do a pass on this integration test.
* Use `SetUp` and `TearDown` instead of `::testing::Test`'s ctor and dtor.
* Assert before using the values in `absl::optional<T>` or `StatusOr<T>`.
* Make calls using clients instead of connections (this unblocks my other PR)

I am confused what is under test though... I think it is `pubsublite::MakePublisherConnection(...)` ? If I understood, I would probably rename the test cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9678)

### DIFF
--- a/google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
+++ b/google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
@@ -12,14 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "google/cloud/pubsublite/admin_connection.h"
-#include "google/cloud/pubsublite/internal/location.h"
+#include "google/cloud/pubsub/publisher.h"
+#include "google/cloud/pubsublite/admin_client.h"
+#include "google/cloud/pubsublite/endpoint.h"
 #include "google/cloud/pubsublite/options.h"
 #include "google/cloud/pubsublite/publisher_connection.h"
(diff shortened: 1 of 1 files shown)


google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
```diff
@@ -48,119 +45,105 @@ auto constexpr kThroughputCapacityMiB = 4;
...
+             generator, 1, "abcdefghijlkmnopqrstuvwxyz0123456789");
+}
+
+void GarbageCollect(AdminServiceClient client, std::string const& parent) {
+  auto stale_prefix =
+      parent + "/topics/" + TestTopicPrefix(system_clock::now() - hours(48));
+  auto topics = client.ListTopics(parent);
+  for (auto const& topic : topics) {
+    ASSERT_STATUS_OK(topic);
+    if (!std::regex_search(topic->name(), std::regex{kTopicRegex})) continue;
```
coryan, line 73: nit: I think we should build the regex outside the loop?

coryan (approved):

google/cloud/pubsublite/integration_tests/publisher_integration_test.cc
dbolduc: You're right. Fixed.

PR #9694, opened 2022-08-16
feat(GCS+gRPC): routing headers for WriteObject

dbolduc:
Fixes b/238793833

Routing rules are here: https://github.com/googleapis/googleapis/blob/b7cb84f5d42e6dba0fdcc2d8689313f6a8c9d7b9/google/storage/v2/storage.proto#L340-L349

I did the logic by hand, instead of using what the generator might produce. I did this because the two fields are a `oneof` that our code is already split on.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9694)


coryan (approved):
Thanks!  Saved me some time here.

PR #9701, opened 2022-08-17
ci: fix output of checkers build

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9701)

### DIFF
--- a/ci/cloudbuild/builds/checkers.sh
+++ b/ci/cloudbuild/builds/checkers.sh
@@ -191,38 +191,40 @@ time {
   done
 }
 
-printf "%-30s" "Updating generated doxygen:" >&2
-# Update the service's endpoint environment variables.
-for dox in google/cloud/*/doc/main.dox; do
-  lib="${dox%/doc/main.dox}"
-  inject_start="<!-- inject-endpoint-env-vars-start -->"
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/builds/checkers.sh
```diff
@@ -191,38 +191,40 @@ time {
...
-    fi
-    make_connection_re='Make.*?Connection()'
-    make_connection=$(grep -Pom1 "${make_connection_re}" "${connection_h}")
-    env_vars+=("- \`${variable}=...\` overrides the")
-    env_vars+=("  \`EndpointOption\` (which defaults to ${endpoint})")
-    env_vars+=("  used by \`${make_connection}()\`.")
-    env_vars+=("")
-  done
-  sed -i -f - "${dox}" <<EOT
+printf "%-30s" "Updating doxygen landing pages:" >&2
```
devbww, line 194: Thanks.

Totally optional: If you can think of something that begins with "Running", it would conform more with the existing output.

devbww (approved):

ci/cloudbuild/builds/checkers.sh
coryan: Done. PTAL

devbww (approved):
Thanks.

PR #9718, opened 2022-08-23
fix(bigtable): `DataConnection` refreshes channels

### LINKED ISSUE #9712: Fix bigtable channel refreshing for `DataConnection`
We use a temporary `shared_ptr<CompletionQueue>` when we create the Stubs. https://github.com/googleapis/google-cloud-cpp/blob/18937be27e09b01ada592e090014e5b239aae674/google/cloud/bigtable/internal/bigtable_stub_factory.cc#L60

So `DataConnection` never refreshes its channels. It always backs out, because there are no more references to the CQ: https://github.com/googleapis/google-cloud-cpp/blob/18937be27e09b01ada592e090014e5b239aae674/google/cloud/bigtable/internal/connection_refresh_state.cc#L74-L75
 
We can fix the problem by updating the channel refresh API to accept a `shared_ptr<internal::CompletionQueueImpl>` instead of a `shared_ptr<CompletionQueue>`. The CQImpl is also owned by the `DataConnection`'s background threads, so it will not go out of scope until the class goes out of scope.

dbolduc:
Fixes #9712 

This is mostly a refactor. Update the channel refresh API to accept a `shared_ptr<internal::CompletionQueueImpl>` instead of a `shared_ptr<CompletionQueue>`.

Now, we will not back out of the channel refresh loop too early, because the background threads hold onto the CQ object for the lifetime of the stub.

Add a regression test for this behavior that fails at HEAD.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9718)

### DIFF
--- a/google/cloud/bigtable/internal/bigtable_channel_refresh_test.cc
+++ b/google/cloud/bigtable/internal/bigtable_channel_refresh_test.cc
@@ -71,6 +71,60 @@ TEST(BigtableChannelRefresh, Disabled) {
       cq, Options{}.set<bigtable::MaxConnectionRefreshOption>(ms(0)));
 }
 
+TEST(BigtableChannelRefresh, Continuations) {
+  using ms = std::chrono::milliseconds;
+  using ns = std::chrono::nanoseconds;
+  using ::testing::ReturnRef;
+
+  promise<StatusOr<std::chrono::system_clock::time_point>> p;
+
+  auto mock_cq = std::make_shared<MockCompletionQueueImpl>();
+  grpc::CompletionQueue grpc_cq;
+  EXPECT_CALL(*mock_cq, cq).WillRepeatedly(ReturnRef(grpc_cq));
+
+  ::testing::InSequence s;
+  EXPECT_CALL(*mock_cq, MakeRelativeTimer)
+      .WillOnce([&p](std::chrono::nanoseconds duration) {
+        EXPECT_LE(ns(ms(500)).count(), duration.count());
+        EXPECT_LE(duration.count(), ns(ms(1000)).count());
+        // This is a regression test for
+        // https://github.com/googleapis/google-cloud-cpp/issues/9712
+        //
+        // We delay returning the future until the Stub is fully constructed.
+        return p.get_future();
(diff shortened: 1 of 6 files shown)


google/cloud/bigtable/internal/bigtable_channel_refresh_test.cc
```diff
@@ -71,6 +72,59 @@ TEST(BigtableChannelRefresh, Disabled) {
...
+
+  auto mock_cq = std::make_shared<MockCompletionQueueImpl>();
+  grpc::CompletionQueue grpc_cq;
+  EXPECT_CALL(*mock_cq, cq).WillRepeatedly(ReturnRef(grpc_cq));
+
+  ::testing::InSequence s;
+  EXPECT_CALL(*mock_cq, MakeRelativeTimer)
+      .WillOnce([&p](std::chrono::nanoseconds duration) {
+        EXPECT_LE(ns(ms(500)).count(), duration.count());
+        EXPECT_LE(duration.count(), ns(ms(1000)).count());
```
devbww, line 90: Does `.count()` need to be involved?  That is, can't we compare the durations directly?

dbolduc: We could compare the durations directly. I think the idea was to have (arguably) better error messages. e.g.
```
Expected: (ns(ms(1000)).count()) <= (duration.count()), actual: 1000000000 vs 852000000
```
vs.
```
Expected: (ms(1000)) <= (duration), actual: 8-byte object <E8-03 00-00 00-00 00-00> vs 8-byte object <00-8E AB-25 00-00 00-00>
```

devbww: Fair enough.  Perhaps consider these options:

1. Leave it as-is.
2. Wrap each side of duration expectations in `absl::FromChrono`, so you even get even nicer output, including units.

Either of these (`.count()` or wrapping) could disappear when durations become streamable in C++20.

devbww (approved):

PR #9723, opened 2022-08-24
fix: update instance admin iam test to skip service account check in emulator

arunmoezhi:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9723)


google/cloud/spanner/admin/integration_tests/instance_admin_integration_test.cc
```diff
@@ -220,9 +220,10 @@ TEST_F(InstanceAdminClientTest, InstanceIam) {
   ASSERT_FALSE(in.project_id().empty());
   ASSERT_FALSE(in.instance_id().empty());
 
-  ASSERT_FALSE(internal::GetEnv("GOOGLE_CLOUD_CPP_SPANNER_TEST_SERVICE_ACCOUNT")
-                   .value_or("")
-                   .empty());
+  auto test_iam_service_account_ =
```
devbww, line 223: `error: invalid case style for variable 'test_iam_service_account_'`

It should not have a trailing underscore (it looks like you copied it from a place where it is a member).

That said, it doesn't look like this test uses the value of `${GOOGLE_CLOUD_CPP_SPANNER_TEST_SERVICE_ACCOUNT}`, so perhaps we'd be better off simply removing the assertion altogether.

PR #9724, opened 2022-08-24
impl(oauth2): allow for optional signing account in SignBlob

scotthart:
This is necessary in order to be consistent with the current `storage::oauth2::SignBlob` functionality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9724)

### DIFF
--- a/google/cloud/internal/oauth2_service_account_credentials.cc
+++ b/google/cloud/internal/oauth2_service_account_credentials.cc
@@ -225,11 +225,13 @@ ServiceAccountCredentials::AuthorizationHeader() {
 }
 
 StatusOr<std::vector<std::uint8_t>> ServiceAccountCredentials::SignBlob(
-    std::string const& signing_account, std::string const& blob) const {
-  if (signing_account != info_.client_email) {
-    return Status(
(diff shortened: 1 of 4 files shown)


coryan (approved):

google/cloud/internal/oauth2_service_account_credentials.cc
```diff
@@ -225,11 +225,13 @@ ServiceAccountCredentials::AuthorizationHeader() {
...
 StatusOr<std::vector<std::uint8_t>> ServiceAccountCredentials::SignBlob(
-    std::string const& signing_account, std::string const& blob) const {
-  if (signing_account != info_.client_email) {
-    return Status(
-        StatusCode::kInvalidArgument,
-        "The current_credentials cannot sign blobs for " + signing_account);
+    absl::optional<std::string> const& signing_account,
+    std::string const& blob) const {
+  if (signing_account.has_value() &&
+      signing_account.value() != info_.client_email) {
```
devbww, line 231: Nitty nit: Typically we use dereferencing instead of `value()` when it is a consequent of `has_value()`.

PR #8689, opened 2022-04-05
test(spanner): re-enable DataTypeIntegrationTest.JsonMaxNesting

devbww:
Restore verification of the (newly-lowered) JSON nesting limit.
See #8327.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/8689)

PR #9727, opened 2022-08-24
test(spanner): cater to the service vagaries on nested JSON

devbww:
Reduce the "spanner_json_max_nesting_level" from 90 to 80, now that
the service is crashing on nesting depths somewhere in [82..90].
It appears that 80 will be the new documented limit.

And that means we temporarily expect `kAlreadyExists` when trying to
re-add a value with a nesting depth of 81.

See #7212, #8327, and #8689 for previous action in this area.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9727)

### DIFF
--- a/google/cloud/spanner/integration_tests/data_types_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/data_types_integration_test.cc
@@ -432,17 +432,17 @@ TEST_F(DataTypeIntegrationTest, InsertAndQueryWithStruct) {
 // Verify maximum JSON nesting.
 TEST_F(DataTypeIntegrationTest, JsonMaxNesting) {
   // The default value of the backend spanner_json_max_nesting_level flag.
-  int const k_spanner_json_max_nesting_level = 80;
+  int const spanner_json_max_nesting_level = 80;
 
-  // Nested arrays that exceed `k_spanner_json_max_nesting_level` by one.
+  // Nested arrays that exceed `spanner_json_max_nesting_level` by one.
   std::string bad_json;
-  for (int i = 0; i != k_spanner_json_max_nesting_level + 1; ++i)
+  for (int i = 0; i != spanner_json_max_nesting_level + 1; ++i)
     bad_json.append(1, '[');
   bad_json.append("null");
-  for (int i = 0; i != k_spanner_json_max_nesting_level + 1; ++i)
+  for (int i = 0; i != spanner_json_max_nesting_level + 1; ++i)
     bad_json.append(1, ']');
 
-  // Nested arrays that match `k_spanner_json_max_nesting_level`.
+  // Nested arrays that match `spanner_json_max_nesting_level`.
   std::string good_json = bad_json.substr(1, bad_json.size() - 2);
 
   std::vector<Json> const good_data = {Json(good_json)};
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/integration_tests/data_types_integration_test.cc
```diff
@@ -431,8 +431,8 @@ TEST_F(DataTypeIntegrationTest, InsertAndQueryWithStruct) {
 
 // Verify maximum JSON nesting.
 TEST_F(DataTypeIntegrationTest, JsonMaxNesting) {
-  // The default value of the backend max-nesting-level flag.
-  int const k_spanner_json_max_nesting_level = 90;
+  // The default value of the backend spanner_json_max_nesting_level flag.
+  int const k_spanner_json_max_nesting_level = 80;
```
coryan, line 435: The name looks funny, if this is intended to be a compile-time constant maybe:

```
auto constexpr kSpannerJsonMaxNestingLevel = 80;
```

google/cloud/spanner/integration_tests/data_types_integration_test.cc
```diff
@@ -456,11 +456,21 @@ TEST_F(DataTypeIntegrationTest, JsonMaxNesting) {
     // The emulator has no such limitation, so it tries to re-insert.
     EXPECT_THAT(result, StatusIs(StatusCode::kAlreadyExists));
   } else {
+#if 1
+    // NOTE: The service is currently (August 2022) in a state where there
+    // is no workable "spanner_json_max_nesting_level".  It accepts 80, but
+    // crashes on some yet-to-be discovered value in [82..90].  81 results
+    // in ALREADY_EXISTS.  We really want to get out of the "JsonMaxNesting"
```
coryan, line 463: Let's do that.  Set the constant to 20.  Or remove this test altogether.  The teams that own this stuff should write their own tests.

coryan (approved):
I would rather remove this test.  But I guess this is just the second strike.

google/cloud/spanner/integration_tests/data_types_integration_test.cc
devbww: Yeah ... without the `constexpr` that is what's required (I believe).  But I was happy with that as "spanner_json_max_nesting_level" is the true name of the Spanner flag, so it helped when I needed to re-search.

That said, see below.

google/cloud/spanner/integration_tests/data_types_integration_test.cc
devbww: I'm totally on board with that (removing the test altogether).  But I propose that we wait until the current issue is addressed.

google/cloud/spanner/integration_tests/data_types_integration_test.cc
coryan: You can at least remove the `k_` prefix in that case.

google/cloud/spanner/integration_tests/data_types_integration_test.cc
coryan: SGTM

google/cloud/spanner/integration_tests/data_types_integration_test.cc
devbww: Done.

PR #9758, opened 2022-09-01
refactor: make HTTP logs safer and more readable

coryan:
Shorten some excessively long messages, simplify the output format, and
truncate bearer tokens in the logs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9758)

### DIFF
--- a/google/cloud/internal/curl_wrappers.cc
+++ b/google/cloud/internal/curl_wrappers.cc
@@ -280,7 +280,7 @@ std::string DebugSendHeader(char const* data, std::size_t size) {
       trailer = payload.substr(nl_pos);
       body = payload.substr(bearer_pos, nl_pos - bearer_pos);
     }
-    auto marker = body.size() > limit ? "...<truncated>..." : "";
+    auto const* marker = body.size() > limit ? "...<truncated>..." : "";
     body = absl::ClippedSubstr(std::move(body), 0, limit);
     return absl::StrCat(">> curl(Send Header): ", prefix, body, marker,
                         trailer);
--- a/google/cloud/internal/curl_wrappers.cc
+++ b/google/cloud/internal/curl_wrappers.cc
@@ -264,19 +264,28 @@ std::string DebugRecvHeader(char const* data, std::size_t size) {
 }
 
 std::string DebugSendHeader(char const* data, std::size_t size) {
-  auto header = absl::string_view{data, size};
-  auto const match = absl::string_view{": Bearer "};
-  auto const limit = match.size() + 32;
-  auto const bearer_pos = header.find(match);
+  // libcurl delivers multiple headers in a single payload, separated by '\n'.
+  auto const payload = absl::string_view{data, size};
+  // We want to truncate the portion of the payload following this ": Bearer" to
+  // at most 32 characters, skipping everything else until any newline.
+  auto const bearer = absl::string_view{": Bearer "};
+  auto const limit = bearer.size() + 32;
+  auto const bearer_pos = payload.find(bearer);
   if (bearer_pos != std::string::npos) {
-    auto const nl_pos = header.find('\n', bearer_pos);
-    auto const n = (std::min)(limit, nl_pos - bearer_pos);
-    return absl::StrCat(
-        ">> curl(Send Header): ", header.substr(0, bearer_pos + n),
-        n == limit ? "...<truncated>..." : "",
-        nl_pos == std::string::npos ? "" : header.substr(nl_pos));
+    auto const nl_pos = payload.find('\n', bearer_pos);
+    auto const prefix = payload.substr(0, bearer_pos);
+    auto trailer = absl::string_view{};
+    auto body = payload.substr(bearer_pos);
+    if (nl_pos != std::string::npos) {
+      trailer = payload.substr(nl_pos);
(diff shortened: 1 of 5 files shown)


google/cloud/internal/curl_wrappers.cc
```diff
@@ -168,6 +169,17 @@ class CurlInitializer {
   CurlInitializer() { curl_global_init(CURL_GLOBAL_ALL); }
   ~CurlInitializer() { curl_global_cleanup(); }
 };
+
+auto constexpr kMaxDebugLength = 128;
```
devbww, line 173: Is it better to make this explicitly `std::size_t` (and then let `std::min` deduce it)?

google/cloud/internal/curl_wrappers.cc
```diff
@@ -243,26 +255,59 @@ std::size_t CurlAppendHeaderData(CurlReceivedHeaders& received_headers,
...
+std::string DebugSendHeader(char const* data, std::size_t size) {
+  auto header = absl::string_view{data, size};
+  auto const match = absl::string_view{": Bearer "};
+  auto const limit = match.size() + 32;
+  auto const bearer_pos = header.find(match);
+  if (bearer_pos != std::string::npos) {
+    auto const nl_pos = header.find('\n', bearer_pos);
+    auto const n = (std::min)(limit, nl_pos - bearer_pos);
+    return absl::StrCat(">> curl(Send Header): ", header.substr(0, bearer_pos),
+                        header.substr(bearer_pos, n), "...<truncated>...",
```
devbww, line 275: `header.substr(0, bearer_pos) + header.substr(bearer_pos, n)` == `header.substr(0, bearer_pos + n)`.

devbww, line 275: Nothing is truncated if `n` == `nl_pos - bearer_pos`, right?

google/cloud/internal/curl_wrappers.cc
```diff
@@ -243,26 +255,59 @@ std::size_t CurlAppendHeaderData(CurlReceivedHeaders& received_headers,
...
+                        header.substr(nl_pos), "\n");
```
devbww, line 276: This seems to append a trailing `"\n"` that we wouldn't get in the `bearer_pos == std::string::npos` case.

coryan: I think you meant `nl_pos == std::string::npos`, I fixed that.  Otherwise, this is in the body of a `if (bearer_pos != std::string::npos)` ?

google/cloud/internal/curl_wrappers.cc
coryan: Fixed

coryan: Fixed

google/cloud/internal/curl_wrappers.cc
```diff
@@ -243,26 +255,60 @@ std::size_t CurlAppendHeaderData(CurlReceivedHeaders& received_headers,
...
+  auto const match = absl::string_view{": Bearer "};
+  auto const limit = match.size() + 32;
+  auto const bearer_pos = header.find(match);
+  if (bearer_pos != std::string::npos) {
+    auto const nl_pos = header.find('\n', bearer_pos);
+    auto const n = (std::min)(limit, nl_pos - bearer_pos);
+    return absl::StrCat(
+        ">> curl(Send Header): ", header.substr(0, bearer_pos + n),
+        n == limit ? "...<truncated>..." : "",
+        nl_pos == std::string::npos ? "" : header.substr(nl_pos));
```
devbww, line 277: >> This seems to append a trailing `"\n"` that we wouldn't get in the `bearer_pos == std::string::npos` case.
> 
> I think you meant `nl_pos == std::string::npos`, I fixed that. Otherwise, this is in the body of a `if (bearer_pos != std::string::npos)` ?

I still meant `bearer_pos == std::string::npos` ... that is, comparing to the final `return` (outside the `if`) case.  But the extra `"\n"` is gone now, so all is good.

google/cloud/internal/curl_wrappers.cc
```diff
@@ -243,26 +255,60 @@ std::size_t CurlAppendHeaderData(CurlReceivedHeaders& received_headers,
...
+  auto header = absl::string_view{data, size};
+  auto const match = absl::string_view{": Bearer "};
+  auto const limit = match.size() + 32;
+  auto const bearer_pos = header.find(match);
+  if (bearer_pos != std::string::npos) {
+    auto const nl_pos = header.find('\n', bearer_pos);
+    auto const n = (std::min)(limit, nl_pos - bearer_pos);
+    return absl::StrCat(
+        ">> curl(Send Header): ", header.substr(0, bearer_pos + n),
+        n == limit ? "...<truncated>..." : "",
```
devbww, line 276: >> Nothing is truncated if `n` == `nl_pos - bearer_pos`, right?
> 
> Fixed

You chose to test `n == limit`, which then mistakenly announces truncation when `limit == nl_pos - bearer_pos`.

coryan: Sorry about that, I believe it is fixed now.

devbww (approved):

PR #9761, opened 2022-09-01
cleanup(storage): make logs more readable

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9761)


google/cloud/internal/binary_data_as_debug_string.cc
```diff
@@ -13,72 +13,24 @@
...
-  if (count != 0) {
-    flush();
-  }
-  return result;
+  auto dump = std::string{
+      data, max_output_bytes == 0 ? size : std::min(size, max_output_bytes)};
+  std::transform(dump.begin(), dump.end(), dump.begin(),
+                 [](auto c) { return std::isprint(c) ? c : '.'; });
+  if (max_output_bytes == 0 || size <= max_output_bytes) return dump;
+  return dump + "...<truncated>...";
```
devbww, line 31: This case is always interesting.  If the data exceeded the maximum output size by, say, 1, then we add an additional 17 bytes to announce that fact, whereas we could have just added 1 more byte and seen all the data.

coryan: Yeah... I figured we could improve on that later.

PR #9769, opened 2022-09-02
test(spanner): create a PostgreSQL version of DatabaseIntegrationTest

devbww:
Add a `PgDatabaseIntegrationTest` test fixture that creates the
database using `DatabaseDialect::POSTGRESQL` and with PostgreSQL-
specific column types.  This enables the testing of types with
PG type annotation codes, like `spanner::PgNumeric`, from
`spanner/integration_tests:data_types_integration_test`.  Then,
add "WriteRead" and "WriteReadArray" tests for `PgNumeric` that
were previously missing.

Also move some (negative) checks for JSON index/key and PG NUMERIC
key support to the data-types integration test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9769)

### DIFF
--- a/google/cloud/spanner/integration_tests/data_types_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/data_types_integration_test.cc
@@ -12,12 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "google/cloud/spanner/admin/database_admin_client.h"
 #include "google/cloud/spanner/client.h"
 #include "google/cloud/spanner/database.h"
 #include "google/cloud/spanner/mutations.h"
 #include "google/cloud/spanner/testing/database_integration_test.h"
 #include "google/cloud/spanner/timestamp.h"
-#include "google/cloud/internal/getenv.h"
 #include "google/cloud/testing_util/status_matchers.h"
 #include "absl/memory/memory.h"
 #include "absl/time/time.h"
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/integration_tests/data_types_integration_test.cc
```diff
@@ -63,35 +65,43 @@ StatusOr<T> WriteReadData(Client& client, T const& data,
   return actual;
 }
 
-class DataTypeIntegrationTest
-    : public spanner_testing::DatabaseIntegrationTest {
- public:
+template <typename T>
```
coryan, line 68: Did you consider composition instead of inheritance?  I am guessing you did. Can you add a comment explaining why it does not work?

coryan (approved):

google/cloud/spanner/integration_tests/data_types_integration_test.cc
devbww: GoogleTest fixtures are all based on inheriting from `testing::Test`.  It looks to me like it would be more difficult (or at least less familiar) to buck that trend.

coryan: I was thinking along these lines:

```cc
template <typename T>
class DataTypeIntegrationTest : public ::testing::Test {
 private:
   T base_impl_;
};
```

devbww: Yeah, maybe, I guess.  If `T` isn't a `testing::Test` it would mean agreeing on some new interface.  But if it is, I'm not sure that it is buying you much.

coryan: Up to you, I don't need to look at this code that often.

Issue #5673, opened 2021-01-05
Enable GCS+gRPC tests for metadata operations against production


vnghia:
Currently, GCS + gRPC is not GA, some `rpc` ( e.g `InsertBucketAccessControl` ) are unimplemented. This issue is a reminder for enabling tests against production when GCS + gRPC is GA.

- [ ] #9800
- [ ] #9801
- [ ] #9802
- [ ] #9803
- [ ] #9804
- [ ] #9805
- [ ] #9806
- [x] Manually verify HMAC keys work

coryan:
The title was a bit confusing because media operations are already being tested against production.

coryan:
The remaining work is tracked in the separate bugs listed above.

PR #9788, opened 2022-09-08
refactor(GCS+gRPC): helpers for bucket names

coryan:
Motivated by #5673

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9788)


google/cloud/storage/internal/grpc_bucket_name.h
```diff
@@ -0,0 +1,36 @@
...
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace storage {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+std::string BucketNameToProto(std::string const& bucket_name);
+std::string BucketNameFromProto(std::string const& bucket_name);
```
devbww, line 28: I'm not sure "proto" enters into this.  Would these be better as something like `BucketNameToId(bucket_name)` and `BucketIdToName(bucket_id)`?

dbolduc (approved):
* you could add unit tests
* you could add a `NotifcationNameToProto(std::string const& bucket_name, std::string const& notification_id)`

google/cloud/storage/internal/grpc_bucket_name.h
coryan: I think the name should hint that it is gRPC related, as the `id` and `name` are identical for REST.  Please let me know if the new names make sense.

coryan:
> * you could add unit tests

Done.

> * you could add a `NotifcationNameToProto(std::string const& bucket_name, std::string const& notification_id)`

Seems like not enough value for money.

devbww (approved):

PR #9790, opened 2022-09-08
fix(GCS+gRPC): correctly parse Bucket `name` field

coryan:
I thought I was being clever by reusing the `id` field. That does not work if the `id` field is filtered out.

Motivated by #5673

### DIFF
--- a/google/cloud/storage/internal/grpc_bucket_metadata_parser.cc
+++ b/google/cloud/storage/internal/grpc_bucket_metadata_parser.cc
@@ -21,6 +21,7 @@
 #include "google/cloud/internal/time_utils.h"
 #include "absl/algorithm/container.h"
 #include "absl/strings/match.h"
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/grpc_bucket_metadata_parser.cc
```diff
@@ -21,6 +21,7 @@
 #include "google/cloud/internal/time_utils.h"
 #include "absl/algorithm/container.h"
 #include "absl/strings/match.h"
+#include "absl/strings/strip.h"
```
dbolduc, line 24: unused?

dbolduc (approved):

google/cloud/storage/internal/grpc_bucket_metadata_parser.cc
coryan: Fixed.

PR #9798, opened 2022-09-08
fix(common): better defaults for curl initialization

coryan:
By default, initialize libcurl with locking and signal handlers that work for most applications. In particular, leaving the default handler for `SIGPIPE` usually results in the application crashes when a socket is closed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9798)

### DIFF
--- a/google/cloud/internal/curl_wrappers.cc
+++ b/google/cloud/internal/curl_wrappers.cc
@@ -307,7 +307,10 @@ void CurlInitializeOnce(Options const& options) {
     //
     //     https://curl.haxx.se/libcurl/c/threadsafe.html
     //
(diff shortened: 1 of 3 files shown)


dbolduc (commented):
Should we have an `Options CurlDefaultOptions(Options)`, where we can default these values to true (and test that the defaults are correct)?

Maybe we call it here: https://github.com/googleapis/google-cloud-cpp/blob/171f936071a216701ef530835bfb24a68bc9ff33/google/cloud/internal/curl_rest_client.cc#L102

dbolduc (approved):
Thanks

PR #9850, opened 2022-09-14
ci: placeholders to change default MSVC version (#9829)

coryan:
This change just creates placeholder Kokoro configuration files. In a separate change I will remove the `bazel*.cfg` files without an explicit MSVC version. Then I will switch the default to 2022.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9850)


coryan:
In case this looks familiar: I am backporting CI changes to the v2.2.x branch.

devbww (approved):

PR #9855, opened 2022-09-15
fix(storage): error message for resumable uploads

### LINKED ISSUE #9750: Error in validatecommittedsize
During a transfer using the cpp library, we observed this error.

ValidateCommittedSize: the server indicates that 41943040000 bytes are committed  but given the current request no more than 41810919424 should be. This could be caused by multiple applications trying to use the same resumable upload, but could be a bug in the library or the service. If you believe this is a bug, please contact support or report the bug at [https://github.com/googleapis/google-cloud-cpp/issues/new](https://www.google.com/url?sa=D&q=https%3A%2F%2Fgithub.com%2Fgoogleapis%2Fgoogle-cloud-cpp%2Fissues%2Fnew) Include as much information as you can including this message, session_id

I have the session id and other information to share out of band(don't want to post on github)

coryan:
One of the error conditions is actually a rare, but unavoidable, problem when resuming uploads.  Improve the error message to clarify this situation.

Fixes #9750

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9855)


google/cloud/storage/internal/retry_client.cc
```diff
@@ -120,14 +122,24 @@ Status ValidateCommittedSize(UploadChunkRequest const& request,
...
-       << " same resumable upload, but could be a bug in the library or"
-       << " the service. If you believe this is a bug, please contact support"
-       << " or report the bug at"
-       << " https://github.com/googleapis/google-cloud-cpp/issues/new"
-       << " Include as much information as you can including this message";
-    os << ", session_id=" << request.upload_session_url();
+       << expected_committed_size << " are expected be."
+       << " Most likely your application resumed an upload, and the client"
+       << " library queried the service to find the current persisted bytes."
+       << " In some cases, the service is still writing in data in the"
```
dbolduc, line 128: s/in data/data/

google/cloud/storage/internal/retry_client.cc
```diff
@@ -120,14 +122,24 @@ Status ValidateCommittedSize(UploadChunkRequest const& request,
...
+       << " background and conservatively reports the bytes persisted."
+       << " Then the next write reports a much higher number of bytes"
+       << " persisted than expected. It is not possible for the client library"
+       << " to recover from this situation. The application needs to resume the"
+       << " upload."
```
dbolduc, line 133: add a `\n` ? Or is that a bad idea in error messages?

google/cloud/storage/internal/retry_client.cc
```diff
@@ -120,14 +122,24 @@ Status ValidateCommittedSize(UploadChunkRequest const& request,
...
+       << " This could also be caused by multiple instances of a distributed"
+       << " application trying to use the same resumable upload, this is a bug"
+       << " in the application."
```
dbolduc, line 136: same `\n` question

dbolduc (approved):

google/cloud/storage/internal/retry_client.cc
coryan: I want to avoid those in error messages.

google/cloud/storage/internal/retry_client.cc
coryan: Ditto

google/cloud/storage/internal/retry_client.cc
coryan: Fixed and some rewording.

Issue #9854, opened 2022-09-15
Restore GCB builds for v1.42.x branch


coryan:
Our CI builds are configured to run for `main` and all the `v[0-9].*` branches.  For example:

https://github.com/googleapis/google-cloud-cpp/blob/8c3c51072f6943c4b94612eee42e1595a4e04357/ci/cloudbuild/triggers/clang-tidy-pr.yaml#L6

As the build matrix evolves, the CI scripts in the branches are no longer usable.  The branch may be missing scripts that are needed for new builds.  Likewise, the triggers for old builds that no longer apply are removed.

We could backport the CI scripts, but that will require back porting the fixes (e.g., `clang-tidy` cleanups) in the code too.  It seems better to preserve the triggers for the LTS branch (or any branch where we need to create a patch release).

I think we need to change all the existing triggers to only apply to `main`.  And then re-create the old triggers for `v1.42.x`, but have them only apply to the v1.42.x branch. We will need to do something similar for the `v2.*.x` branches when we decide to create patch releases.

PR #9857, opened 2022-09-15
ci: GCB triggers are branch-specific

coryan:
Made the GCB triggers specific to the `main` branch. When we need builds in a branch (for LTS and/or patch releases), we will use the `convert-to-branch-triggers.sh` script to create separate triggers that apply to this branch. I will update the release documentation in a future PR.

Part of the work for #9854

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9857)

### DIFF
--- a/ci/cloudbuild/trigger.sh
+++ b/ci/cloudbuild/trigger.sh
@@ -50,7 +50,7 @@ github:
   name: google-cloud-cpp
   owner: googleapis
   push:
-    branch: ^(master|main|v\\d+\\..*)$
+    branch: ^main$
 name: ${name}-ci
 substitutions:
   _BUILD_NAME: ${name}
@@ -69,7 +69,7 @@ github:
   name: google-cloud-cpp
   owner: googleapis
   pullRequest:
-    branch: ^(master|main|v\\d+\\..*)$
+    branch: ^main$
     commentControl: COMMENTS_ENABLED_FOR_EXTERNAL_CONTRIBUTORS_ONLY
 name: ${name}-pr
 substitutions:
--- a/ci/cloudbuild/triggers/asan-ci.yaml
+++ b/ci/cloudbuild/triggers/asan-ci.yaml
@@ -3,7 +3,7 @@ github:
   name: google-cloud-cpp
   owner: googleapis
   push:
-    branch: ^(master|main|v\d+\..*)$
+    branch: ^main$
 name: asan-ci
 substitutions:
   _BUILD_NAME: asan
--- a/ci/cloudbuild/triggers/asan-pr.yaml
+++ b/ci/cloudbuild/triggers/asan-pr.yaml
@@ -3,7 +3,7 @@ github:
   name: google-cloud-cpp
   owner: googleapis
   pullRequest:
-    branch: ^(master|main|v\d+\..*)$
+    branch: ^main$
     commentControl: COMMENTS_ENABLED_FOR_EXTERNAL_CONTRIBUTORS_ONLY
 name: asan-pr
 substitutions:
--- a/ci/cloudbuild/triggers/bazel-oldest-ci.yaml
+++ b/ci/cloudbuild/triggers/bazel-oldest-ci.yaml
@@ -3,7 +3,7 @@ github:
   name: google-cloud-cpp
   owner: googleapis
   push:
-    branch: ^(master|main|v\d+\..*)$
+    branch: ^main$
 name: bazel-oldest-ci
 substitutions:
   _BUILD_NAME: bazel-oldest
--- a/ci/cloudbuild/triggers/bazel-oldest-pr.yaml
+++ b/ci/cloudbuild/triggers/bazel-oldest-pr.yaml
@@ -3,7 +3,7 @@ github:
   name: google-cloud-cpp
   owner: googleapis
   pullRequest:
-    branch: ^(master|main|v\d+\..*)$
+    branch: ^main$
     commentControl: COMMENTS_ENABLED_FOR_EXTERNAL_CONTRIBUTORS_ONLY
 name: bazel-oldest-pr
 substitutions:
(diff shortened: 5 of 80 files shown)


ci/cloudbuild/convert-to-branch-triggers.sh
```diff
@@ -0,0 +1,78 @@
...
+#     -h|--help           Print this help message
+
+set -euo pipefail
+
+source "$(dirname "$0")/../lib/init.sh"
+source module ci/lib/io.sh
+
+function print_usage() {
+  # Extracts the usage from the file comment starting at line 17.
+  sed -n '17,/^$/s/^# \?//p' "${PROGRAM_PATH}"
```
devbww, line 34: Consider moving the usage text into a here document within `print_usage()`.  That avoids having to reopen the script (which is never reliable) and magic numbers like 17.  And it may even be clearer here (I think so), than in the comment.

ci/cloudbuild/convert-to-branch-triggers.sh
```diff
@@ -0,0 +1,78 @@
...
+
+function convert_triggers() {
+  local branch="$1"
+  local name_prefix="${branch//./-}"
+  shift
+  for file in "$@"; do
+    sed \
+      -e "s/^name: /name: ${name_prefix}-/" \
+      -e "s/branch: .*/branch: ${branch}/" "${file}" |
+      sponge "${file}"
```
devbww, line 45: Consider `sed -i` (`--in-place`), which removes the need for `sponge`.

ci/cloudbuild/convert-to-branch-triggers.sh
```diff
@@ -0,0 +1,78 @@
...
+      exit 0
+      ;;
+    --)
+      shift
+      break
+      ;;
+  esac
+done
+
+if [[ $# -lt 1 ]]; then
```
devbww, line 71: s/-lt/-ne/

ci/cloudbuild/convert-to-branch-triggers.sh
```diff
@@ -0,0 +1,78 @@
...
+  print_usage
+  exit 1
+fi
+BRANCH="$1"
```
devbww, line 75: This script should only be run within the specified branch, right?  If so, consider discovering the branch instead.

ci/cloudbuild/convert-to-branch-triggers.sh
```diff
@@ -0,0 +1,78 @@
...
+
+git ls-files -z -- ci/cloudbuild/triggers/*.yaml |
+  xargs -P "$(nproc)" -n 50 -0 bash -c "convert_triggers \"${BRANCH}\" \"\$@\""
```
devbww, line 78: Given that there are (currently) only 79 of these files, we may want to reduce the `-n` flag (if parallelism helps at all).

coryan: Done

ci/cloudbuild/convert-to-branch-triggers.sh
coryan: Fixed

ci/cloudbuild/convert-to-branch-triggers.sh
coryan: Done.

ci/cloudbuild/convert-to-branch-triggers.sh
coryan: Done.  For whatever reasons the other scripts use the comment trick :shrug:

ci/cloudbuild/convert-to-branch-triggers.sh
coryan: N/A now that the branch is discovered.

ci/cloudbuild/convert-to-branch-triggers.sh
```diff
@@ -31,48 +27,55 @@ source module ci/lib/io.sh
...
       break
       ;;
   esac
 done
 
-if [[ $# -lt 1 ]]; then
-  print_usage
-  exit 1
+if [[ -z "${BRANCH}" ]]; then
+  BRANCH="$(git branch --show-current)"
```
devbww, line 77: Is the current branch really what we're looking for?  For example, in your current context I imagine you want to get "main" and not "ci-triggers-are-branch-specific" (and, say, "v1.42.x" if you were branched from that).

coryan: That is why I added `--branch`. If I am running in `v1.42.x` I do not want `main`.  If I am running in `fix-v1.42.x-triggers` I do not want that branch (I want `v1.42.x` instead).

devbww: Right, but I figured there must be some way to discover the name you really want from where you are running (not that I've put any thought into it beyond that).

coryan: "the name you really want" can only be defined by convention.  If you are in a branch that looks like a release branch, then you probably want that branch.  If you are not in such a branch, you probably want the "parent" branch.  Leaving aside the fact that there may be multiple parent branches...  I found this horror:

https://stackoverflow.com/questions/3161204/how-to-find-the-nearest-parent-of-a-git-branch

I think that shell pipeline can be optimized (combining `grep` and `sed` calls), but frankly this is not that worthwhile.  I suspect most of the time we will just use the `--branch` option to be explicit.

coryan: I removed the script. Let's discuss that in a separate PR and get the triggers committed first.

dbolduc (approved):

devbww (approved):
> I removed the script. Let's discuss that in a separate PR and get the triggers committed first.

Fair enough.  I plan to experiment on the "parent branch" question this afternoon.  I'll let you know what conclusions I draw.

PR #9860, opened 2022-09-15
chore: update googleapis SHA circa 2022-09-15

devbww:
Mostly to pick up `{Create,Update,Delete}InstanceConfig()`
for the `spanner_admin::InstanceAdminClient`. This should
allow us to merge the "CMMR Phase 2" work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9860)

### DIFF
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -85,11 +85,11 @@ def google_cloud_cpp_deps():
         http_archive(
             name = "com_google_googleapis",
             urls = [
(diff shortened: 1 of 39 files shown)


devbww:
> Temporarily edit a generated file to fix the "publish-docs" build

The "generate-libraries" build, of course, undoes that.  So, I guess I'm stuck until I can get an upstream change made.  Sigh.

coryan:
Comment changes upstream move relatively quickly.

With that said, I think we have some places in the generator where we "fix" the comments so Doxygen can consume them. Maybe we need to do the same thing here?

coryan (approved):
LGTM

devbww:
Thanks.

PR #9872, opened 2022-09-19
feat(spanner): add support for CMMR phase 2

devbww:
Tests and samples for user-managed instance configurations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9872)

### DIFF
--- a/google/cloud/spanner/testing/cleanup_stale_instances.cc
+++ b/google/cloud/spanner/testing/cleanup_stale_instances.cc
@@ -91,6 +91,52 @@ Status CleanupStaleInstances(Project const& project) {
   return Status();
 }
 
+Status CleanupStaleInstanceConfigs(Project const& project) {
+  std::regex name_regex(R"(projects/.+/instanceConfigs/)"
+                        R"(custom-temporary-config-(\d{4}-\d{2}-\d{2})-.+)");
+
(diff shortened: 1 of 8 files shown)


google/cloud/spanner/testing/cleanup_stale_instances.cc
```diff
@@ -91,6 +91,52 @@ Status CleanupStaleInstances(Project const& project) {
...
+    auto name = config->name();
+    std::smatch m;
+    if (std::regex_match(name, m, name_regex)) {
+      auto date_str = m[1];
+      if (date_str < cutoff_date) {
+        configs.push_back(name);
+      }
+    }
+  }
+  // Let it fail if we have too many leaks.
```
coryan, line 128: I am Okay with this, but curious about the rationale.  Seems like an opportunity for flakes that we do not need?

coryan (approved):

google/cloud/spanner/testing/cleanup_stale_instances.cc
devbww: Yes, I agree.  At the moment I am only duplicating the behavior of `CleanupStaleInstances()`, which I didn't originate, but I will more than happily followup by removing both cases.  Thanks.

PR #9875, opened 2022-09-19
test(spanner): remove "too many stale" checks

devbww:
These have never caught any problems, and, even if they did, the remedy would probably just be to remove the stales objects anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9875)

### DIFF
--- a/google/cloud/spanner/testing/cleanup_stale_instances.cc
+++ b/google/cloud/spanner/testing/cleanup_stale_instances.cc
@@ -74,10 +74,6 @@ Status CleanupStaleInstances(Project const& project) {
       }
     }
   }
-  // Let it fail if we have too many leaks.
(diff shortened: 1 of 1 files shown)


dbolduc (approved):
The returned `Status` does not seem necessary, but who cares. Seems like we would rather have the regex defined somewhere and checked in a unit test (or not checked at all).

devbww:
> The returned `Status` does not seem necessary, but who cares. Seems like we would rather have the regex defined somewhere and checked in a unit test (or not checked at all).

Yes ... I mostly agree with you.  The remaining failures, as errors in the test code, and not the code under test, could simply be fatal errors.  I'll ponder on it some.  Thanks.

Issue #7142, opened 2021-08-09
Support mocking for the storage library without exposing implementation details


coryan:
## Background

Some customers need to "mock" the storage library for their own testing.  We have examples on how to to this, but they use names in the `google::cloud::storage::internal::` namespace:

https://github.com/googleapis/google-cloud-cpp/blob/ee2e0e51f0ed63a941d5f8e7ebb9226553813c24/google/cloud/storage/examples/storage_client_mock_samples.cc#L116-L145

Normally `gcs::internal::` would be reserved for implementation details, and would be subject to change without notice.  Exposing and documenting how they work muddles the message to our customers ("don't use `internal` because it changes, except for these bits, maybe") and complicate the life for the library developers ("these bits of `internal` need to remain backwards compatible").

Even when the types and functions do not change, the sequence of calls may change, possibly breaking a mock, for an example see #6983 and #7042.

## Overview

I think we should change the mechanism used to mock the `gcs::Client` class. We should define a public API for mocking. Keeping in mind that there are 50 odd RPCs in the GCS surface, we should also try to avoid defining one `*Request` and `*Response` type for each RPC. Specially considering there are half a dozen or more "optional" parameters with each RPC.

I think we should define the following APIs.

First, a class that applications can use to mock `gcs::Client`, the class would live in a public namespace, such as:

```cc
namespace google::cloud::storage_mocks {
class MockClient {
public:
  // details about methods below
};
} // namespace google::cloud::storage_mocks
```

Note that this class does not require an interface. It does not expose any of the APIs defined in `gcs::internal`.

Second, a factory function to use this class from testing code:

```cc
namespace google::cloud::storage_mocks {

google::cloud::storage::Client MakeMockClient(std::shared_ptr<MockClient> mock);

} // namespace google::cloud::storage_mocks
```

Applications do not need to know how a `MockClient` is mapped to a client (we will describe that below), but once they have a `gcs::Client` object they can use it in their test code as usual.

Third, all the methods for `MockClient` will have exactly the same signature:

```cc
namespace google::cloud::storage_mocks {
class MockClient {
public:
  MOCK_METHOD(MockHttpResponse, GetObjectMetadata, (MockHttpRequest const&), ());
};
} // namespace google::cloud::storage_mocks
```

The `HttpResponse` and `HttpRequest` classes are provided as part of the mocking library.  I think we should just clone the API (but not the implementation) from the functions framework:

```cc
class MockHttpRequest {
 public:
[...]

PR #9886, opened 2022-09-21
feat(storage): easier mocks with `BucketMetadata`

coryan:
Populating all the fields for `BucketMetadata` was somewhat difficult. The class tried to prevent applications from setting fields that only the server would set.  I thought about adding a `BucketMetadataBuilder`, but that would duplicate a lot of code without preventing anything. In addition, all the other libraries use protos that allow any field to be set, and that does not seemt to create a lot of trouble or confusion.

I also stopped using the "composition-by-private-inheritance" base because it does not save enough code.  I expect this class will go away once I make similar changes to `ObjectMetadata`.

Part of the work for #8929, I also think it will help with #7142

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9886)

### DIFF
--- a/google/cloud/storage/bucket_metadata.h
+++ b/google/cloud/storage/bucket_metadata.h
@@ -35,8 +35,6 @@ namespace cloud {
 namespace storage {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
-struct BucketMetadataParser;
-struct GrpcBucketMetadataParser;
 struct GrpcBucketRequestParser;
 }  // namespace internal
 
@@ -574,7 +572,7 @@ std::ostream& operator<<(std::ostream& os,
 /**
  * Represents a Google Cloud Storage Bucket Metadata object.
  */
-class BucketMetadata : private internal::CommonMetadata<BucketMetadata> {
+class BucketMetadata {
  public:
   BucketMetadata() = default;
 
@@ -710,7 +708,12 @@ class BucketMetadata : private internal::CommonMetadata<BucketMetadata> {
   }
   ///@}
 
-  using CommonMetadata::etag;
+  std::string const& etag() const { return etag_; }
+  /// @note This is only intended for mocking.
+  BucketMetadata& set_etag(std::string v) {
+    etag_ = std::move(v);
+    return *this;
+  }
 
   /**
    * @name Get and set the IAM configuration.
@@ -743,8 +746,19 @@ class BucketMetadata : private internal::CommonMetadata<BucketMetadata> {
   }
   ///@}
 
-  using CommonMetadata::id;
-  using CommonMetadata::kind;
+  std::string const& id() const { return id_; }
+  /// @note This is only intended for mocking
+  BucketMetadata& set_id(std::string v) {
+    id_ = std::move(v);
+    return *this;
+  }
+
+  std::string const& kind() const { return kind_; }
+  /// @note This is only intended for mocking
+  BucketMetadata& set_kind(std::string v) {
+    kind_ = std::move(v);
+    return *this;
+  }
 
   /// @name Accessors and modifiers to the `labels`.
   ///@{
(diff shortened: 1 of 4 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9886)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

google/cloud/storage/bucket_metadata.h
```diff
@@ -819,8 +833,15 @@ class BucketMetadata : private internal::CommonMetadata<BucketMetadata> {
     return *this;
   }
 
+  /// Returns the location type (e.g. regional vs. dual region).
   std::string const& location_type() const { return location_type_; }
```
devbww, line 838: Nit: Similar blocks omit this separator line.

google/cloud/storage/internal/bucket_metadata_parser.cc
```diff
@@ -189,17 +195,16 @@ std::map<std::string, std::string> ParseLabels(nlohmann::json const& json) {
...
-Status ParseProjectNumber(std::int64_t& project_number,
-                          nlohmann::json const& json) {
-  auto p = internal::ParseLongField(json, "projectNumber");
-  if (!p) return std::move(p).status();
-  project_number = *p;
+Status ParseOwner(BucketMetadata& meta, nlohmann::json const& json) {
+  if (!json.contains("owner")) return Status{};
+  Owner o;
+  o.entity = json["owner"].value("entity", "");
+  o.entity_id = json["owner"].value("entityId", "");
```
devbww, line 202: Q.  Are we at all concerned about repeated lookups?

devbww (approved):
Q.  Did you noticed the codecov warning?  I wonder if we're doing something wrong (and perhaps the new uploader now tells us).

coryan:
> Q. Did you noticed the codecov warning? I wonder if we're doing something wrong (and perhaps the new uploader now tells us).

If you mean this one:

> Current head https://github.com/googleapis/google-cloud-cpp/commit/74f4ba0f7404a9cffa1f0714152f28040c57918c differs from pull request most recent head https://github.com/googleapis/google-cloud-cpp/commit/aa7232fc8376bd05196626afed08f828b89bf4fc.

I ignore those, they have been around since approximately forever.   I am not sure we are doing anything "wrong".  My interpretation is that they happen whenever a PR starts is based on an older "HEAD" than the current one (e.g. because something was merged after the PR was created).

google/cloud/storage/internal/bucket_metadata_parser.cc
coryan: I doubt it is in the critical path for anything, but there is no reason to duplicate work either.  Fixed.

google/cloud/storage/bucket_metadata.h
coryan: Fixed (at least for new functions)

google/cloud/storage/internal/bucket_metadata_parser.cc
devbww: Without wanting to derail you any further, it seems like we could still go from 2 lookups to 1 using `json.find("owner")` (not that I'm familiar with the API).

devbww:
> > Current head [74f4ba0](https://github.com/googleapis/google-cloud-cpp/commit/74f4ba0f7404a9cffa1f0714152f28040c57918c) differs from pull request most recent head [aa7232f](https://github.com/googleapis/google-cloud-cpp/commit/aa7232fc8376bd05196626afed08f828b89bf4fc).
> 
> I ignore those, they have been around since approximately forever. I am not sure we are doing anything "wrong". My interpretation is that they happen whenever a PR starts is based on an older "HEAD" than the current one (e.g. because something was merged after the PR was created).

I was more interested in the other one `Patch has no changes to coverable lines.`  I guess it is not actually a warning, but shouldn't "head" and "base" be different?

google/cloud/storage/internal/bucket_metadata_parser.cc
coryan: Tracking that in #9887

coryan:
> 
> I was more interested in the other one `Patch has no changes to coverable lines.` I guess it is not actually a warning, but shouldn't "head" and "base" be different?

Candidly I don't know.  I tend to ignore those and only worry about large-ish changes in coverage.

PR #9902, opened 2022-09-23
cleanup: use [[deprecated]]

dbolduc:
we require C++14, so simplify the deprecated macro

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9902)

### DIFF
--- a/google/cloud/internal/attributes.h
+++ b/google/cloud/internal/attributes.h
@@ -21,9 +21,6 @@
  * Marks a class, struct, enum, function, member function, or variable as
  * deprecated.
  *
- * This macro injects an attribute equivalent to C++14's `[[deprecated(msg)]]`.
- * On C++11 it uses compiler extensions to implement equivalent behavior.
- *
  * @note If you need to temporarily disable the deprecation warnings in your
  *     application, consider adding a pre-processor flag to define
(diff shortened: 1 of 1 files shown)


google/cloud/internal/attributes.h
```diff
@@ -36,14 +33,9 @@
...
 #define GOOGLE_CLOUD_CPP_DEPRECATED(message) [[deprecated(message)]]
-#elif defined(_MSC_VER)
-#define GOOGLE_CLOUD_CPP_DEPRECATED(message) __declspec(deprecated(message))
-#elif defined(__clang__) || defined(__GNUC__)
-#define GOOGLE_CLOUD_CPP_DEPRECATED(message) \
-  __attribute__((deprecated(message)))
-#endif  // C++ <= 11
+#endif
 
 #ifndef GOOGLE_CLOUD_CPP_DEPRECATED
```
devbww, line 40: Now that there is an unconditional `#else` branch, we might as well move the do-nothing `GOOGLE_CLOUD_CPP_DEPRECATED()` definition to the "Do nothing" branch.

dbolduc: Good point. Fixed.

devbww (approved):

PR #9905, opened 2022-09-23
impl(GCS+gRPC): automatically refresh gRPC channels

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9905)

### DIFF
--- a/google/cloud/internal/async_connection_ready.cc
+++ b/google/cloud/internal/async_connection_ready.cc
@@ -27,7 +27,6 @@ AsyncConnectionReadyFuture::AsyncConnectionReadyFuture(
     : cq_(std::move(cq)), channel_(std::move(channel)), deadline_(deadline) {}
 
 future<Status> AsyncConnectionReadyFuture::Start() {
-  channel_->GetState(true);
   RunIteration();
   return promise_.get_future();
 }
--- a/google/cloud/internal/async_connection_ready.cc
+++ b/google/cloud/internal/async_connection_ready.cc
@@ -27,7 +27,8 @@ AsyncConnectionReadyFuture::AsyncConnectionReadyFuture(
     : cq_(std::move(cq)), channel_(std::move(channel)), deadline_(deadline) {}
 
 future<Status> AsyncConnectionReadyFuture::Start() {
-  RunIteration(channel_->GetState(true));
+  channel_->GetState(true);
+  RunIteration();
   return promise_.get_future();
 }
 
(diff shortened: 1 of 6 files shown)


google/cloud/internal/async_connection_ready.h
```diff
@@ -50,14 +51,30 @@ class AsyncConnectionReadyFuture
 
   // gRPC uses an anonymous type for the gRPC channel state enum :shrug:.
   using ChannelStateType = decltype(GRPC_CHANNEL_READY);
```
devbww, line 53: Now unused (at least here)?

google/cloud/storage/internal/storage_round_robin.cc
```diff
@@ -262,6 +281,32 @@ std::shared_ptr<StorageStub> StorageRoundRobin::Child() {
   return child;
 }
 
+void StorageRoundRobin::Refresh(
+    std::size_t index,
+    std::weak_ptr<google::cloud::internal::CompletionQueueImpl> wcq) {
+  auto cq = wcq.lock();
+  if (!cq) return;
+  auto deadline = std::chrono::system_clock::now() + kRefreshPeriod;
+  // An invalid index, stop the loop.  There is no need to check lock the mutex,
```
devbww, line 290: s/check lock/lock/?

devbww (approved):

google/cloud/internal/async_connection_ready.h
coryan: Removed.

google/cloud/storage/internal/storage_round_robin.cc
coryan: Fixed.

google/cloud/internal/async_connection_ready.cc
```diff
@@ -27,7 +27,8 @@ AsyncConnectionReadyFuture::AsyncConnectionReadyFuture(
     : cq_(std::move(cq)), channel_(std::move(channel)), deadline_(deadline) {}
 
 future<Status> AsyncConnectionReadyFuture::Start() {
-  RunIteration(channel_->GetState(true));
+  channel_->GetState(true);
```
dbolduc, line 30: nit: I think this call is superfluous

devbww: I looked at that too, but it causes the channel to try to make a connection if it is currently IDLE.

dbolduc: won't it get called here anyways? https://github.com/googleapis/google-cloud-cpp/blob/55f718928d4ff14515dde192763715c89a7746bf/google/cloud/internal/async_connection_ready.cc#L74

google/cloud/storage/internal/storage_round_robin.cc
```diff
@@ -262,6 +281,32 @@ std::shared_ptr<StorageStub> StorageRoundRobin::Child() {
...
 }
 
+void StorageRoundRobin::Refresh(
+    std::size_t index,
+    std::weak_ptr<google::cloud::internal::CompletionQueueImpl> wcq) {
+  auto cq = wcq.lock();
+  if (!cq) return;
+  auto deadline = std::chrono::system_clock::now() + kRefreshPeriod;
+  // An invalid index, stop the loop.  There is no need to lock the mutex, as
+  // the channels do not change after the class is initialized.
```
devbww, line 291: Q.  How is an invalid index injected in the first place?  If `channels_` doesn't change after initialization, couldn't we have caught this earlier?

coryan: This is just defensive.  It really cannot (or should not) happen, but if it does then I would rather not crash and just stop using the invalid index.

devbww: OK, thanks, although I've never been a big fan of that philosophy.

google/cloud/internal/async_connection_ready.cc
coryan: Removed

PR #9912, opened 2022-09-23
ci: enable CMake policies up to 3.24

coryan:
The main motivation is to improve the customer experience with CMake under Windows.  Until CMake 3.15 switching the MSVC runtime was a bit of a pain.  We used `cmake_minimum_required(VERSION 3.10)` which has two effects:

- It requires CMake to be >= 3.10
- It guarantees the CMake 3.10 behavior, even if you use 3.15 or 3.24

We do not want the 3.10 behavior on Windows, it requires too many steps to switch the MSVC runtime library.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9912)

### DIFF
--- a/cmake/CompileProtos.cmake
+++ b/cmake/CompileProtos.cmake
@@ -272,13 +272,13 @@ function (google_cloud_cpp_load_protodeps var file)
     set(targets_to_omit
         "google-cloud-cpp::cloud_orgpolicy_v1_orgpolicy_protos"
         "google-cloud-cpp::cloud_oslogin_common_common_protos"
-        "google-cloud-cpp::identity_accesscontextmanager_type_type_protos")
+        "google-cloud-cpp::identity_accesscontextmanager_type_type_protos"
+        "google-cloud-cpp::cloud_recommender_v1_recommender_protos")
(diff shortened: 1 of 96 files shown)


cmake/CompileProtos.cmake
```diff
@@ -272,13 +272,13 @@ function (google_cloud_cpp_load_protodeps var file)
     set(targets_to_omit
         "google-cloud-cpp::cloud_orgpolicy_v1_orgpolicy_protos"
         "google-cloud-cpp::cloud_oslogin_common_common_protos"
-        "google-cloud-cpp::identity_accesscontextmanager_type_type_protos")
+        "google-cloud-cpp::identity_accesscontextmanager_type_type_protos"
+        "google-cloud-cpp::cloud_recommender_v1_recommender_protos")
```
devbww, line 276: Keep sorted?

devbww (approved):

cmake/CompileProtos.cmake
coryan: Done.

Issue #9909, opened 2022-09-23
timeout in storage ThreadIntegrationTest.Unshared


devbww:
```
Note: Randomizing tests' orders with a seed of 87453 .
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from ThreadIntegrationTest
[ RUN      ] ThreadIntegrationTest.ReuseConnections
[       OK ] ThreadIntegrationTest.ReuseConnections (23007 ms)
[ RUN      ] ThreadIntegrationTest.Unshared
-- Test timed out at 2022-09-23 19:05:12 UTC --
```
```
 //google/cloud/storage/tests:thread_integration_test                    TIMEOUT in 900.0s
```
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/ac5407df-2d28-41de-8e6d-af401e101383;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/52b8433e9ea7691750ffe770cc084dc46beff378/fedora-36-gcs-grpc/log-ac5407df-2d28-41de-8e6d-af401e101383.txt

devbww:
"Fixed" #9918.

dbolduc:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/e3c02776-bfd9-4efd-b804-7c23bf395af0;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/10a02afe861d77a7ecf377a45e82fa30e056aa27/fedora-36-gcs-grpc/log-e3c02776-bfd9-4efd-b804-7c23bf395af0.txt

:(

dbolduc:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/365c09a0-5209-4577-ae19-374a906f97f6;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/73347b9b34b5c0337f90978caad858543664e310/fedora-37-bazel-gcs-grpc/log-365c09a0-5209-4577-ae19-374a906f97f6.txt

coryan:
More than 90d without repeats. Closing.

PR #9918, opened 2022-09-25
test(storage): reduce runtime of ThreadIntegrationTest.Unshared

devbww:
Lower the number of objects created/deleted in the test, which brings down the average runtime from about to 12.5m to 10m, and that reduces the possibility of exceeding the "long" timeout of 15m.

"Fixes" #9909.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9918)

### DIFF
--- a/google/cloud/storage/tests/thread_integration_test.cc
+++ b/google/cloud/storage/tests/thread_integration_test.cc
@@ -115,7 +115,7 @@ TEST_F(ThreadIntegrationTest, Unshared) {
   // feature.
   auto const thread_count =
       (std::min)(32U, (std::max)(8U, std::thread::hardware_concurrency()));
-  auto const object_count = 100 * thread_count;
+  auto const object_count = 75 * thread_count;
   std::vector<std::string> objects(object_count);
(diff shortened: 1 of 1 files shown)


google/cloud/storage/tests/thread_integration_test.cc
```diff
@@ -115,7 +115,7 @@ TEST_F(ThreadIntegrationTest, Unshared) {
   // feature.
   auto const thread_count =
       (std::min)(32U, (std::max)(8U, std::thread::hardware_concurrency()));
-  auto const object_count = 100 * thread_count;
+  auto const object_count = 75 * thread_count;
```
coryan, line 118: If you want to set this to 50 or 25 that would probably work too.

coryan (approved):

google/cloud/storage/tests/thread_integration_test.cc
devbww: OK, 50 it is.  Thanks.

PR #9952, opened 2022-09-29
feat(storage): small optimization in `InsertObject()`

coryan:
Optimize the common case when creating valid MIME separators for JSON multipart uploads.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9952)


google/cloud/storage/internal/generate_message_boundary.h
```diff
@@ -58,6 +67,24 @@ template <typename RandomStringGenerator,
 std::string GenerateMessageBoundary(
     std::string const& message, RandomStringGenerator&& random_string_generator,
     int initial_size, int growth_size) {
+  // This is not the only place we make this assumption. Duplicating the static
+  // assertion should make it easier to find the code if we ever need to fix it.
+  static_assert(std::numeric_limits<unsigned char>::max() == 255,
+                "required for efficient GenerateMessageBoundary");
```
devbww, line 73: Do we need to make the assumption?  Couldn't we size the `seen` array with `std::numeric_limits<unsigned char>::max()`?

google/cloud/storage/internal/generate_message_boundary.h
```diff
@@ -58,6 +67,24 @@ template <typename RandomStringGenerator,
...
+  static char const kBoundaryChars[] =
+      "abcdefghijklmnopqrstuvwxyz0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+  auto constexpr kSkip = std::size_t{64};
+  std::array<bool, 255> seen;
+  std::fill(seen.begin(), seen.end(), false);
+  for (auto i = kSkip - 1; i < message.size(); i += kSkip) {
+    seen[static_cast<unsigned char>(message[i])] = true;
+  }
+  for (auto c : kBoundaryChars) {
+    if (c == '\0') continue;
```
devbww, line 84: Nit: I don't have a nice workaround for this, but it seems like `kBoundaryChars` is not living up to its name when this check is required.

google/cloud/storage/internal/generate_message_boundary.h
```diff
@@ -58,6 +67,24 @@ template <typename RandomStringGenerator,
...
     int initial_size, int growth_size) {
+  // This is not the only place we make this assumption. Duplicating the static
+  // assertion should make it easier to find the code if we ever need to fix it.
+  static_assert(std::numeric_limits<unsigned char>::max() == 255,
+                "required for efficient GenerateMessageBoundary");
+
+  static char const kBoundaryChars[] =
+      "abcdefghijklmnopqrstuvwxyz0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+  auto constexpr kSkip = std::size_t{64};
+  std::array<bool, 255> seen;
```
devbww, line 78: s/255/256/?

dbolduc: oops. never trust my code.

google/cloud/storage/internal/generate_message_boundary.h
coryan: Would work everywhere except on Cray computers where `unsigned char` was 64-bits 😀 

This is mostly to make it easier to find the code when/if we try to port to a machine with 16-bit chars (which unlike Cray computers still exist).

google/cloud/storage/google_cloud_cpp_storage.bzl
```diff
@@ -228,4 +228,5 @@ google_cloud_cpp_storage_srcs = [
     "version.cc",
     "well_known_headers.cc",
     "well_known_parameters.cc",
+    "internal/generate_message_boundary.cc",
```
devbww, line 231: Sort?

google/cloud/storage/google_cloud_cpp_storage.cmake
```diff
@@ -228,7 +228,7 @@ add_library(
     well_known_headers.cc
     well_known_headers.h
     well_known_parameters.cc
-    well_known_parameters.h)
+    well_known_parameters.h internal/generate_message_boundary.cc)
```
devbww, line 231: Sort?

google/cloud/storage/internal/generate_message_boundary.h
```diff
@@ -33,13 +35,15 @@ namespace internal {
...
- * of the class that uses it.
+ * The algorithm is:
+ * - First, try using fixed strings consisting of 64 copies of any valid
+ *   boundary character. We can verify if one of those would work by searching
+ *   for the boundary character every 64 bytes in the message. If the character
+ *   never appears in that search, then there is no substring in the message
+ *   consisting of 64 copies of the character.
+ * - If that fails, generate a short random string, and search for it in the
+ *   message. If the message has that string, append some more random characters
+ *   and keep searching.
```
devbww, line 46: Possible note for the future: Given that the message actually contained the length-n random string, it may be better to generate an entirely new length-n+k random string rather than appending k new random chars to the existing string.

google/cloud/storage/internal/generate_message_boundary_benchmark.cc
```diff
@@ -0,0 +1,78 @@
...
+#include <random>
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace storage {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+namespace {
+
```
devbww, line 28: Any benchmark results to add here?

google/cloud/storage/internal/generate_message_boundary.h
devbww: Ack.  But platforms with 16-bit `char` (still) exist?  I'm surprised.

google/cloud/storage/internal/generate_message_boundary.h
coryan: The advantage of adding `k` characters is that you don't need to search from the beginning.  If the first time you find `foo` is at position 1000, then you know  `foobar` is not found before position 1000 either. That means the algorithm is always making progress.

google/cloud/storage/google_cloud_cpp_storage.cmake
coryan: Yeah, sorry, I wanted to push the code so we could have the offline discussion

google/cloud/storage/internal/generate_message_boundary_benchmark.cc
coryan: Discussing that elsewhere.

coryan:
Closing for now, I need to redo this.

PR #9954, opened 2022-09-29
fix(spanner): remove session from pool upon "not found" refresh failure

### LINKED ISSUE #4026: remove the Session from the allocation pool if refresh fails with "not found"
We currently use a "send-and-forget" `GetSession()` call to refresh expiring `Session`s.  See `SessionPool::RefreshExpiringSessions()`.

If the `GetSession()` call fails with `kNotFound`, we could remove the session from the allocation pool (although this should never happen if refresh is working correctly).

devbww:
When the "SELECT 1" session-refresh query fails with "Session not found", remove the bad session from the pool, if possible.  Previously we did nothing because (1) it shouldn't happen, and (2) we have other mechanisms in place to deal with bad sessions (which must remain).  But it is better dealt with directly if we can.

Fixes #4026.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9954)

### DIFF
--- a/google/cloud/spanner/internal/session_pool.cc
+++ b/google/cloud/spanner/internal/session_pool.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/spanner/internal/session_pool.h"
 #include "google/cloud/spanner/internal/connection_impl.h"
 #include "google/cloud/spanner/internal/session.h"
+#include "google/cloud/spanner/internal/status_utils.h"
 #include "google/cloud/spanner/options.h"
 #include "google/cloud/completion_queue.h"
 #include "google/cloud/internal/async_retry_loop.h"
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/internal/session_pool.cc
```diff
@@ -152,20 +152,43 @@ void SessionPool::RefreshExpiringSessions() {
...
-          // We simply discard the response as handling IsSessionNotFound()
-          // by removing the session from the pool is problematic (and would
-          // not eliminate the possibility of IsSessionNotFound() elsewhere).
-          // The last-use time has already been updated to throttle attempts.
-          // TODO(#4026): Re-evaluate these decisions.
-          (void)result.get();
-        });
+    auto handler =
+        [pool, session_name = refresh.second](
+            future<StatusOr<google::spanner::v1::ResultSet>> result) {
```
coryan, line 159: Optional: use `auto result`

coryan (approved):

PR #9965, opened 2022-10-02
fix(storage): respect MIME message boundary size limits

### LINKED ISSUE #9961: MIME separators cannot exceed 70 characters
The storage library sometimes performs JSON multipart uploads and needs a valid MIME separator.

Thought extremely unlikely in practice, in theory the current implementation could exceed the 70 character limit for the "boundary":

https://www.rfc-editor.org/rfc/rfc2046#section-5.1.1

coryan:
Fixes #9961


google/cloud/storage/internal/generate_message_boundary.cc
```diff
@@ -0,0 +1,47 @@
...
+  return candidate;
+}
+
+std::string GenerateMessageBoundaryCandidate(
+    google::cloud::internal::DefaultPRNG& generator) {
+  auto constexpr kSize = 64;
+  auto candidate = std::string{
+      "abcdefghijklmnopqrstuvwxyz0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"};
+  std::shuffle(candidate.begin(), candidate.end(), generator);
+  return candidate.substr(0, kSize);
```
devbww, line 40: Given that there are only 62 chars in `candidate`, the `substr()` is a no-op (after making a copy).  Was that intentional?

google/cloud/storage/internal/generate_message_boundary_benchmark.cc
```diff
@@ -0,0 +1,156 @@
...
+//
+// clang-format off
+// Run on (96 X 2000 MHz CPU s)
+// CPU Caches:
+//  L1 Data 32 KiB (x48)
+//  L1 Instruction 32 KiB (x48)
+//  L2 Unified 1024 KiB (x48)
+//  L3 Unified 39424 KiB (x2)
+// ---------------------------------------------------------------------------------------------
+// Benchmark                                                   Time             CPU   Iterations
```
devbww, line 41: Opinion: I usually think that the minimum is the best value to record (instead of mean or median) for such micro-benchmarks.

google/cloud/storage/internal/generate_message_boundary_benchmark.cc
```diff
@@ -0,0 +1,156 @@
...
+    benchmark::DoNotOptimize(test(message()));
+  }
+}
+
+BENCHMARK_F(GenerateBoundaryFixture, BestCase)
+(benchmark::State& state) {
+  auto test = [](std::string const& message) {
+    auto count = 0;
+    for (std::size_t i = 0; i < message.size(); i += 64) {
+      count += static_cast<int>(message[i] == 'Z');
```
devbww, line 141: Is `if (message[i] == 'Z') ++count`, or even `count += (message[i] == 'Z') ? 1 : 0`, clearer (leaving any optimizations to the compiler)?

coryan: Done.

google/cloud/storage/internal/generate_message_boundary.cc
coryan: Fixed.

Issue #5871, opened 2021-02-18
Support creating hundreds of simultaneous Bigtable Table clients


coryan:
In some of their tests our customers need to do things like this:

```cc
void F() {

  std::string instance_id = "test-instance";
  std::string project_id = "test-project";

  std::mutex mu;
  std::vector<cbt::DataClient> clients;
  std::vector<std::thread> tasks;

  for (int i = 0; i != 200; ++i) {
    tasks.push_back(std::thread{[&] {
      std::lock_guard lk(mu);
      clients.push_back(CreateDefaultDataClient(instance_id, project_id, ...));
    }}); 
    tasks.push_back(std::thread{[&](int i) {
      std::this_thread::sleep_for( /* random time */);
      std::unique_lock lk(mu);
      if (clients.size() < i + 1) return;
      auto data_client = clients[i];
      lk.unlock();
      cbt::Table table(client, "table_id");
      table.Apply( /* stuff */);
    }, i}); 
  }
  for (auto& t : tasks) t.join();
}
```

They are reporting deadlocks in this code (or code equivalent to it).

coryan:
Googlers can look at cl/358000434 for more details.

coryan:
We will revisit in the future.

dbolduc:
## Problem Diagnosis

So this isn't a deadlock, and the issue doesn't have to do with the threads. What is actually happening is that the `DataClient`s destruction may get held up while waiting for the result of the channel refreshing operation: https://github.com/googleapis/google-cloud-cpp/blob/18937be27e09b01ada592e090014e5b239aae674/google/cloud/bigtable/internal/connection_refresh_state.cc#L76-L78

Note that this timeout is 10s long: https://github.com/googleapis/google-cloud-cpp/blob/18937be27e09b01ada592e090014e5b239aae674/google/cloud/bigtable/internal/connection_refresh_state.cc#L30

So we are seeing delays that could be as high as 10s * N, as we wait for each client to go out of scope. If N is large enough this starts looking like a deadlock.

## Simpler Repro

We can write a simpler test that exposes this behavior:
```cc
// This test will take longer than 10s as we are guaranteed to hit a refresh
// that will not complete. Ideally it would only take a second or so.
TEST(DeadlocksTest, DataClientNoThreads) {
  auto constexpr kRefreshPeriod = std::chrono::milliseconds(500);
  auto options =
      Options{}
[...]

PR #9967, opened 2022-10-03
fix(bigtable): simplify channel refresh

dbolduc:
The `DataConnection` part of #5871. (I do not intend to fix this for `DataClient`).

Simplify the channel refresh logic. Stop using `bigtable::ConnectionRefreshState`.

This PR diverges from the `storage` implementation, which puts this functionality in `StorageRoundRobinStub`. Modifying `BigtableChannelRefresh` seemed like a smaller change given the current state of the library.

Eventually we would like the generator to know how to refresh channels (and generate round robin stubs). That will be easiest if the `bigtable` and `storage` implementations are in parity. I will consider refactoring at a later time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9967)


dbolduc:
I will just close this until I am ready to pick it up again. I know where to find it.

PR #6175, opened 2021-04-01
cleanup(bigtable): remove DirectPath specific endpoint

mohanli-ml:
Since the bigtable main endpoint `bigtable.googleapis.com` has rollout to DirectPath, we no longer need the DirectPath specific endpoing `directpath-bigtable.googleapis.com`, and we should move all the tests to use the main endpoint.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/6175)

PR #9978, opened 2022-10-04
feat(bigtable): support `GOOGLE_CLOUD_ENABLE_DIRECT_PATH`

### LINKED ISSUE #4382: Support DirectPath for Cloud Bigtable
See the changes in #4379, we need similar changes for Cloud Bigtable

dbolduc:
Fixes #4382 and basically undoes #6175

I set up a VM with the magic subnet. If I set the endpoint as such and make some request, I see that my gRPC peer is `2001:4860:8040:...etc....`.

Note that `google-c2p:///bigtable.googleapis.com` requires the project to be allow-listed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9978)

### DIFF
--- a/google/cloud/bigtable/internal/defaults.cc
+++ b/google/cloud/bigtable/internal/defaults.cc
@@ -22,6 +22,8 @@
 #include "google/cloud/internal/getenv.h"
 #include "google/cloud/internal/user_agent_prefix.h"
 #include "google/cloud/options.h"
+#include "absl/algorithm/container.h"
+#include "absl/strings/str_split.h"
 #include <chrono>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/internal/defaults.cc
```diff
@@ -135,6 +137,21 @@ Options DefaultOptions(Options opts) {
...
+  if (absl::c_any_of(absl::StrSplit(direct_path, ','),
+                     [](absl::string_view v) { return v == "bigtable"; })) {
+    opts.set<DataEndpointOption>(
+            "google-c2p:///directpath-bigtable.googleapis.com")
+        .set<AuthorityOption>("directpath-bigtable.googleapis.com");
+
+    // When using DirectPath the gRPC library already does load balancing across
+    // multiple sockets, it makes little sense to perform additional load
+    // balancing in the client library.
+    if (!opts.has<GrpcNumChannelsOption>()) opts.set<GrpcNumChannelsOption>(1);
```
coryan, line 152: The ideal number of channels may not be `1`.  The service controls how many subchannels are created.  AFAIK, this is not exposed anywhere, but may be documented somewhere.  Consider a future PR to update this to `ceil($ncores / $subchannels)`

coryan (approved):

google/cloud/bigtable/internal/defaults.cc
dbolduc: Ack. I couldn't find the number. I will ask around.

PR #9981, opened 2022-10-05
feat(generator): make idempotency policy non-abstract

coryan:
The generated `*IdempotencyPolicy` class cannot be pure abstract. If it
is, any custom idempotency policy breaks when we generate new member
functions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9981)


generator/internal/idempotency_policy_generator.cc
```diff
@@ -1,3 +1,4 @@
+
```
devbww, line 1: Delete extra blank line.

devbww (approved):

generator/internal/idempotency_policy_generator.cc
coryan: Ooops, fixed.

PR #9992, opened 2022-10-06
feat(GCS+gRPC): use alternative names in ACLs

### LINKED ISSUE #9900: Improve `*AccessControl` operations for GCS+gRPC
The GCS+gRPC APIs for `*AccessControl` need some improvements to handle entities with alternative names.

coryan:
Some entities in ACLs have more than one name. For example, `project-owners-${PROJECT_NUMBER}` can also be called `project-owners-${PROJECT_ID}`.  The application can refer to the entity by either name.  The client library is supposed to fetch the ACL (using something like `GetBucket()`), find the entity in the returned ACL, apply some changes, and then update the ACL (using something like `UpdateBucket()`).

gRPC will return both names for entities that have them. All the client library needs to do is check for both.

Fixes #9900


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9992)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #9997, opened 2022-10-07
feat(storage): faster `InsertObject()` uploads

coryan:
With this change the client library will optimistically use a randomly generated string as a message boundary, but will not validate if the string appears in the payload.

The new behavior makes sense as (1) it saves about 2ms of client CPU time for 128MiB uploads, (2) if there is a collision this will be detected by the service via the upload checksums, and (3) the probability of a collision is small.

The current implementation has a population of more than $10^{100}$ strings to pick from. The probability of finding this string in a EB of data is less than $1 / 10^{95}$.  You could upload a EB of data per second for a billion years and still not find a collision.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9997)

### DIFF
--- a/google/cloud/storage/internal/curl_client.cc
+++ b/google/cloud/storage/internal/curl_client.cc
@@ -1250,7 +1250,7 @@ StatusOr<ObjectMetadata> CurlClient::InsertObjectMediaMultipart(
   request.ForEachOption(no_content_type);
 
   // 2. Pick a separator that does not conflict with the request contents.
-  auto boundary = PickBoundary(request.contents());
+  auto boundary = MakeBoundary();
   builder.AddHeader("content-type: multipart/related; boundary=" + boundary);
   builder.AddQueryParameter("uploadType", "multipart");
   builder.AddQueryParameter("name", request.object_name());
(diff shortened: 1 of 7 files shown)


google/cloud/storage/options.h
```diff
@@ -286,6 +286,30 @@ struct IdempotencyPolicyOption {
...
+ * When this option is enabled the client library will validate the message,
+ * if it finds a collision, it will repeatedly generate new random strings until
+ * one that does not collide is found. The space of possible random strings is
+ * large enough this should be found in one or two attempts.
+ *
+ * [single request upload]:
+ * https://cloud.google.com/storage/docs/uploads-downloads [JSON multipart
+ * uploads]: https://cloud.google.com/storage/docs/json_api/v1/objects/insert
+ */
+struct ValidateInsertObjectBoundary {
```
dbolduc, line 309: Given the math, I would recommend against adding this option.

Or we could rename it to `SlowDownInsertObjectOption` so customers know what they are signing up for. /s

coryan: Removed.

google/cloud/storage/internal/curl_client.cc
```diff
@@ -1250,7 +1250,7 @@ StatusOr<ObjectMetadata> CurlClient::InsertObjectMediaMultipart(
   request.ForEachOption(no_content_type);
 
   // 2. Pick a separator that does not conflict with the request contents.
```
devbww, line 1252: Does this comment now overstate the properties of the returned value?

google/cloud/storage/internal/rest_client.cc
```diff
@@ -359,7 +356,7 @@ StatusOr<ObjectMetadata> RestClient::InsertObjectMediaMultipart(
   }
 
   // 2. Pick a separator that does not conflict with the request contents.
```
devbww, line 358: Ditto.

google/cloud/storage/internal/curl_client.cc
coryan: PTAL

devbww (approved):
Obligatory reminder to update the commit message.

PR #9998, opened 2022-10-07
feat: add mock library w/ StreamRange

### LINKED ISSUE #9052: Add public library to hold common mocks and test helpers
We will probably call it `google_cloud_cpp_mocks`.

We may want to move some symbols over from `google_cloud_cpp_testing` (e.g. [status matchers](https://github.com/googleapis/google-cloud-cpp/blob/69fdc8d2a28c5a0784b16eacbf3dacff2d309a8a/google/cloud/testing_util/status_matchers.h))

We know we want to include a:
```c++
template <typename T>
StreamRange<T> MakeTestStreamRange(std::vector<T> values, Status final_status);
```

dbolduc:
Fixes #9052 

as discussed in go/cloud-cxx:mock-stream-range

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/9998)


google/cloud/google_cloud_cpp_common.cmake
```diff
@@ -214,6 +214,34 @@ install(
     DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/google_cloud_cpp_common"
     COMPONENT google_cloud_cpp_development)
 
+# Create a header-only library for the mocks. We use a CMake `INTERFACE` library
+# for these, a regular library would not work on macOS (where the library needs
+# at least one .o file). Unfortunately INTERFACE libraries are a bit weird in
+# that they need absolute paths for their sources.
+set(relative_mock_files mocks/mock_stream_range.h)
```
coryan, line 221: nit: should we add the magic comment to keep the "list" of mocks sorted?

google/cloud/google_cloud_cpp_common.cmake
```diff
@@ -214,6 +214,34 @@ install(
...
+    INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+              $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>
+              $<INSTALL_INTERFACE:include>)
+target_compile_options(google_cloud_cpp_mocks
+                       INTERFACE ${GOOGLE_CLOUD_CPP_EXCEPTIONS_FLAG})
+add_library(google-cloud-cpp::mocks ALIAS google_cloud_cpp_mocks)
+
+create_bazel_config(google_cloud_cpp_mocks YEAR "2022")
+
+google_cloud_cpp_install_headers("google_cloud_cpp_mocks"
```
coryan, line 242: FYI, I think this will install the `.h` file, but we are not installing the targets (see #5782).  Can you check that the header installs correctly (maybe look at the installed files in one of the `demo-*` builds?)

coryan (approved):

google/cloud/google_cloud_cpp_common.cmake
dbolduc: I run
```
ci/cloudbuild.sh -t demo-fedora-pr -s
# (inside docker shell now)

ci/cloudbuild.sh --local --build demo-install
# (build runs)

ls /h/google-cloud-cpp-installed/include/google/cloud/mocks/
# Output:
#    mock_stream_range.h
```

And if I run:
```
grep -rl mocks /h/google-cloud-cpp-installed/lib64/cmake/
# Output:
#    /h/google-cloud-cpp-installed/lib64/cmake/google_cloud_cpp_pubsub/pubsub-targets.cmake
```

I take it that #5782 involves getting the other mock libraries to show up.

google/cloud/google_cloud_cpp_common.cmake
dbolduc: Done.

google/cloud/google_cloud_cpp_common.cmake
dbolduc: Maybe done? But does it need a pkgconfig thing like we do with `rest_internal`?

Should it have its own target separate from `google_cloud_cpp_common-targets` ?

coryan (dismissed):

google/cloud/BUILD.bazel
```diff
@@ -188,6 +189,21 @@ load(":google_cloud_cpp_grpc_utils_benchmarks.bzl", "google_cloud_cpp_grpc_utils
     ],
 ) for test in google_cloud_cpp_grpc_utils_benchmarks]
 
+load(":google_cloud_cpp_mocks.bzl", "google_cloud_cpp_mocks_hdrs")
+
+cc_library(
+    name = "google_cloud_cpp_mocks",
```
devbww, line 195: `testonly`?

dbolduc: Done.

google/cloud/google_cloud_cpp_common.cmake
dbolduc: pkg-config has been left for a follow up PR.

I did move the targets out of `google_cloud_cpp_common-targets.cmake` and into a new `google_cloud_cpp_mocks-targets.cmake`. I think we want to do this to avoid installing the mock targets into the `google_cloud_cpp_runtime` component.

PTAL, because your original review was from a lifetime ago

google/cloud/google_cloud_cpp_common.cmake
```diff
@@ -214,6 +214,42 @@ install(
     DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/google_cloud_cpp_common"
     COMPONENT google_cloud_cpp_development)
 
+# Create a header-only library for the mocks. We use a CMake `INTERFACE` library
+# for these, a regular library would not work on macOS (where the library needs
+# at least one .o file). Unfortunately INTERFACE libraries are a bit weird in
```
coryan, line 219: I think this comment is now out of date?

google/cloud/google_cloud_cpp_common.cmake
coryan: There is no `.pc` file, but there is no `-config.cmake` file either.  Is that intentional?

coryan (commented):
Are you planning to add a test to `cmake-install` for this library?

google/cloud/google_cloud_cpp_common.cmake
dbolduc: At the moment, it is a header only library.

coryan: yes, but you are not using the absolute paths anymore?

dbolduc: ah, thanks. I did not read the comment closely.

dbolduc:
> There is no .pc file, but there is no -config.cmake file either. Is that intentional?
>
> Are you planning to add a test to `cmake-install` for this library?

I decided to just do the pkgconfig stuff in this PR. (and test it in an environment where `find_dependency(GTest)` finds `gmock_main`.)

coryan (approved):

PR #10002, opened 2022-10-07
chore: add resourcemanager to typos

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10002)


devbww (commented):
Should we be adding word exceptions rather than excluding whole files?

coryan:
> Should we be adding word exceptions rather than excluding whole files?

If we could we would, but last time I checked the fancy "typos" thing could not do that.

coryan (approved):

PR #10005, opened 2022-10-09
test(storage): minor coverage improvements

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10005)

### DIFF
--- a/google/cloud/storage/bucket_cors_entry_test.cc
+++ b/google/cloud/storage/bucket_cors_entry_test.cc
@@ -1,0 +1,63 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


dbolduc (approved):
just curious, why do `Struct{/*.field=*/val}` instead of `Struct{.field=val}` ?

coryan:
> just curious, why do `Struct{/*.field=*/val}` instead of `Struct{.field=val}` ?

Because the latter is a C++20-ism:

https://en.cppreference.com/w/cpp/language/aggregate_initialization

In plain `C` you can do that since C99 and some compilers support it.

dbolduc:
Ah thanks. I was looking at https://en.cppreference.com/w/c/language/struct_initialization but missed the distinction between C/C++

Issue #7689, opened 2021-12-02
add pubsub support for per-operation options


devbww:
Add an `Options options = {}` argument to the constructor of the client class(es).  Merge these options with the default options for the service, and store them as a member of the client class.

Add an `Options options = {}` argument to each operation within each client.  These options should then be merged with the client options from above, and installed as the prevailing options for the duration of the operation by instantiating an `internal::OptionsSpan`.

You could then use `internal::CurrentOptions()` to obtain (a `const&` to) the prevailing options from anywhere you might need them.  Any cleanup for call paths where `Options` have been passed explicitly is discretionary.

Similar support for the generated client classes was added in #7683, so you might be able to use that as an example.

coryan:
I need this to fix the x-goog-user-project anyway.

coryan:
What is supposed to happen to the options used to initialize a `Connection`?

devbww:
> What is supposed to happen to the options used to initialize a Connection?

At least they should be returned by the `Options()` accessor.  After that, I'd say it is up to the service.

In the generated `ConnectionImpl`s they are only used to get policies when `CurrentOptions()` doesn't have them (but that is only an allowance for testing here: https://github.com/googleapis/google-cloud-cpp/blob/2a40c30e50b30d0f61f41882681ccb60161c7d58/generator/internal/connection_impl_generator.cc#L112-L116).

In the Spanner `ConnectionImpl` though, they are used to check if `rpc-streams` tracing is enabled, and to create the `SessionPool`.

coryan:
I have not been following the ins and outs of this feature as I should.  This bug is asking me to:

Save *all* the options, with their defaults, at the `Client` level, and then set these options (and maybe the per-call options) using an `internal::OptionsSpan`.  That means the options set at the connection level never have an effect.  I recall that we changed our minds about how options-per-call are implemented, did you update this bug to reflect the latest approach?

In Pub/Sub some of the options are *only* used while constructing the `*Connection` + `*Stub` stack, it is impossible to override them for each RPC.  For example, consider this:

https://github.com/googleapis/google-cloud-cpp/blob/207a602a02c5d5bd44546650f84912c1c6a396da/google/cloud/pubsub/publisher_connection.cc#L141-L148

or this:

https://github.com/googleapis/google-cloud-cpp/blob/207a602a02c5d5bd44546650f84912c1c6a396da/google/cloud/pubsub/publisher_connection.cc#L89-L100

[...]

PR #10043, opened 2022-10-13
feat(pubsub): implement per-call options for `Subscriber`

coryan:
Part of the work for #7689

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10043)

### DIFF
--- a/google/cloud/pubsub/subscriber.cc
+++ b/google/cloud/pubsub/subscriber.cc
@@ -22,19 +22,17 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 Subscriber::Subscriber(std::shared_ptr<SubscriberConnection> connection,
                        Options opts)
     : connection_(std::move(connection)),
-      options_(google::cloud::internal::MergeOptions(std::move(opts),
-                                                     connection_->options())) {}
+      options_(
+          internal::MergeOptions(std::move(opts), connection_->options())) {}
 
 future<Status> Subscriber::Subscribe(ApplicationCallback cb, Options opts) {
-  google::cloud::internal::OptionsSpan span(
-      google::cloud::internal::MergeOptions(std::move(opts), options_));
+  internal::OptionsSpan span(internal::MergeOptions(std::move(opts), options_));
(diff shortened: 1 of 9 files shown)


google/cloud/pubsub/subscriber_test.cc
```diff
@@ -76,6 +92,103 @@ TEST(SubscriberTest, SubscribeWithOptions) {
...
+    EXPECT_EQ(current.get<TestOptionA>(), "override-a");
+    EXPECT_EQ(current.get<TestOptionB>(), "override-b2");
+    EXPECT_EQ(current.get<TestOptionC>(), "test-c");
+    return make_ready_future(Status{});
+  });
+
+  Subscriber subscriber(mock, Options{}.set<TestOptionA>("override-a"));
+  ASSERT_STATUS_OK(subscriber
+                       .Subscribe([](Message const&, AckHandler const&) {},
+                                  Options{}.set<TestOptionB>("override-b1"))
```
devbww, line 183: Perhaps these should also be setting `TestOptionA` too, to check that call overrides client.

google/cloud/pubsub/subscriber.cc
```diff
@@ -0,0 +1,44 @@
...
+
+namespace google {
+namespace cloud {
+namespace pubsub {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+Subscriber::Subscriber(std::shared_ptr<SubscriberConnection> connection,
+                       Options opts)
+    : connection_(std::move(connection)),
+      options_(google::cloud::internal::MergeOptions(std::move(opts),
```
devbww, line 25: Q.  Do we like `google::cloud::` qualifications when we're already in `google::cloud`?

google/cloud/pubsub/subscriber_test.cc
coryan: Done

google/cloud/pubsub/subscriber.cc
coryan: Fixed.

devbww (approved):

PR #10046, opened 2022-10-13
fix(rest): add response error payload to Status message

### LINKED ISSUE #9947: Rest Error Message are Less Useful Now
Yes

google/cloud/storage

**Describe the bug**
After the recent refactor errors from rest requests appear to originate [here](https://github.com/googleapis/google-cloud-cpp/blob/1db13d0b0b3025764dc01ed692f9286eb6957e34/google/cloud/internal/curl_impl.cc#L657) when the HTTP error code returned isn't explicit ignored.  This means the body of the response isn't parsed and error details aren't returned.  I have observed this impacting QueryResumableUpload when 409 is returned but I imagine it affects a lot of other errors as well.

**To Reproduce**
Get a resumable upload in a bad state so that queries return:

```
HTTP/2 409 
x-guploader-uploadid: [redacted]
content-type: application/json; charset=UTF-8
date: Tue, 27 Sep 2022 00:02:08 GMT
vary: Origin
vary: X-Origin
cache-control: no-cache, no-store, max-age=0, must-revalidate
expires: Mon, 01 Jan 1990 00:00:00 GMT
pragma: no-cache
content-length: 399
server: UploadServer

{
  "error": {
    "code": 409,
    "message": "The object has already been created in an earlier attempt and was overwritten, possibly due to a race condition.",
    "errors": [
      {
        "message": "The object has already been created in an earlier attempt and was overwritten, possibly due to a race condition.",
        "domain": "global",
        "reason": "conflict"
      }
    ]
  }
}
```

Resuming the upload will produce an error `ABORTED: Permanent error in QueryResumableUpload: Received HTTP status code: 409` without any of the useful error details.

**Expected behavior**
The status message should be `message` from the JSON and ErrorInfo should contain the rest of the detailed information.

**Operating system:**
N/a

N.a

Head: 1db13d0b0b3025764dc01ed692f9286eb6957e34

**Additional context**

scotthart:
The response payload was read into the spill buffer, but was not being included in creation of the Status value.

Using `testing_util::ScopedEnvironment env("GOOGLE_CLOUD_CPP_STORAGE_USE_LEGACY_HTTP", "yes");` on the StreamingWriteFailure test, I modified the checks in the tests to verify the error message the legacy implementation provided. The current implementation with the Rest Library now satisfies those same additional checks.

fixes #9947

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10046)

### DIFF
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -655,7 +655,8 @@ StatusOr<std::size_t> CurlImpl::ReadImpl(absl::Span<char> output) {
   if (curl_closed_) {
     OnTransferDone();
     status = google::cloud::rest_internal::AsStatus(
-        static_cast<HttpStatusCode>(http_code_), {});
+        static_cast<HttpStatusCode>(http_code_),
+        std::string(spill_.data(), spill_offset_));
     TRACE_STATE() << ", status=" << status << ", http code=" << http_code_
                   << "\n";
 
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -30,7 +30,8 @@
 // `TRACE_STATE()` macro is only needed by the library developers when
 // troubleshooting this class.
 #define TRACE_STATE()                                                       \
-  GCP_LOG(TRACE) << __func__ << "(), buffer_.size()=" << buffer_.size()     \
+  GCP_LOG(TRACE) << __func__ << "(), remaining_buffer_.size()="             \
+                 << remaining_buffer_.size()                                \
                  << ", spill_.max_size()=" << spill_.max_size()             \
                  << ", spill_offset_=" << spill_offset_                     \
                  << ", closing=" << closing_ << ", closed=" << curl_closed_ \
@@ -192,7 +193,7 @@ CurlImpl::CurlImpl(CurlHandle handle,
       download_stall_timeout_(0),
       handle_(std::move(handle)),
       multi_(factory_->CreateMultiHandle()),
-      buffer_({nullptr, 0}),
+      remaining_buffer_({nullptr, 0}),
       options_(std::move(options)) {
   CurlInitializeOnce(options_);
   ApplyOptions(options_);
@@ -275,7 +276,7 @@ std::size_t CurlImpl::WriteAllBytesToSpillBuffer(void* ptr, std::size_t size,
 
 std::size_t CurlImpl::WriteToUserBuffer(void* ptr, std::size_t size,
                                         std::size_t nmemb) {
-  if (buffer_.empty()) {
+  if (remaining_buffer_.empty()) {
     TRACE_STATE()
         << " buffer_offset_ >= buffer_size_ *** PAUSING HANDLE *** << \n";
     paused_ = true;
@@ -287,7 +288,7 @@ std::size_t CurlImpl::WriteToUserBuffer(void* ptr, std::size_t size,
   // of absl::Span<char>.
   // Use the spill buffer first, if there is any...
   (void)DrainSpillBuffer();
-  std::size_t free = buffer_.size();
+  std::size_t free = remaining_buffer_.size();
   if (free == 0) {
     TRACE_STATE()
         << " (buffer_size_ - buffer_offset_) == 0 *** PAUSING HANDLE *** \n";
@@ -299,8 +300,8 @@ std::size_t CurlImpl::WriteToUserBuffer(void* ptr, std::size_t size,
 
   // Copy the full contents of `ptr` into the application buffer.
   if (size * nmemb < free) {
-    buffer_ =
-        WriteToBuffer(std::move(buffer_),
+    remaining_buffer_ =
+        WriteToBuffer(std::move(remaining_buffer_),
                       absl::Span<char>(static_cast<char*>(ptr), size * nmemb));
 
     TRACE_STATE() << ", copy full"
@@ -309,8 +310,9 @@ std::size_t CurlImpl::WriteToUserBuffer(void* ptr, std::size_t size,
   }
 
   // Copy as much as possible from `ptr` into the application buffer.
-  buffer_ = WriteToBuffer(std::move(buffer_),
-                          absl::Span<char>(static_cast<char*>(ptr), free));
+  remaining_buffer_ =
+      WriteToBuffer(std::move(remaining_buffer_),
+                    absl::Span<char>(static_cast<char*>(ptr), free));
   // The rest goes into the spill buffer.
   spill_offset_ = size * nmemb - free;
   std::memcpy(spill_.data(), static_cast<char*>(ptr) + free, spill_offset_);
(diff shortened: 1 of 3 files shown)


google/cloud/storage/tests/object_resumable_write_integration_test.cc
```diff
@@ -311,6 +312,15 @@ TEST_F(ObjectResumableWriteIntegrationTest, StreamingWriteFailure) {
       AnyOf(Eq(StatusCode::kFailedPrecondition), Eq(StatusCode::kAborted)))
       << " status=" << os.metadata().status();
 
+  if (UsingEmulator() &&
```
coryan, line 315: It would be more interesting to test the case where this runs against production (vs. the emulator) where the error is parsed and we get something valid, e.g.:

```cc
if (os.metadata().status().code() == StatusCode::kFailedPrecondition) {
  if (UsingEmulator()) {
    EXPECT_THAT(os.metadata().error_info().reason(), Not(IsEmpty()));
    EXPECT_THAT(os.metadata().error_info().domain(), Not(IsEmpty()));
   // maybe something else?
  } else {
    // Your code here
  }
}
```

google/cloud/internal/curl_impl.cc
```diff
@@ -655,7 +655,8 @@ StatusOr<std::size_t> CurlImpl::ReadImpl(absl::Span<char> output) {
   if (curl_closed_) {
     OnTransferDone();
     status = google::cloud::rest_internal::AsStatus(
-        static_cast<HttpStatusCode>(http_code_), {});
+        static_cast<HttpStatusCode>(http_code_),
+        std::string(spill_.data(), spill_offset_));
```
coryan, line 659: There is no guarantee that the spill buffer has the error data.  Nor that it is large enough to hold all the error data.

coryan (changes requested):
Unless I am missing something, this may not always work.

scotthart (commented):
I renamed buffer_ to something more expressive after stubbing my toe on the semantic difference between the output var in ReadImpl and buffer_.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10046#-:-NEMn9o5Et_8wfq7O6w2:b-jcnz50)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/internal/curl_impl.cc` line 659 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10046#-NEIVY8AHOeBEUZPwvWc-r2-659:-NEMmsGK3hWuLgzgge2g:bu3v1w8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bf22195ee195d4a0b830a1fa98760ddb8b9aefb3/google/cloud/internal/curl_impl.cc#L659)):*

There is no guarantee that the spill buffer has the error data.  Nor that it is large enough to hold all the error data.
</blockquote></details>

Good point. There could be data in the user provided buffer as well.
___
*[`google/cloud/storage/tests/object_resumable_write_integration_test.cc` line 315 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10046#-NEIVY8AHOeBEUZPwvWd-r2-315:-NEMn1M6-xyCyvTbcmMf:bi0kgsy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bf22195ee195d4a0b830a1fa98760ddb8b9aefb3/google/cloud/storage/tests/object_resumable_write_integration_test.cc#L315)):*

It would be more interesting to test the case where this runs against production (vs. the emulator) where the error is parsed and we get something valid, e.g.:

```cc
if (os.metadata().status().code() == StatusCode::kFailedPrecondition) {
  if (UsingEmulator()) {
    EXPECT_THAT(os.metadata().error_info().reason(), Not(IsEmpty()));
    EXPECT_THAT(os.metadata().error_info().domain(), Not(IsEmpty()));
   // maybe something else?
  } else {
    // Your code here
  }
}
```

</blockquote></details>

Added check for production error message.

coryan:
*[`google/cloud/internal/curl_impl.cc` line 659 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10046#-NEIVY8AHOeBEUZPwvWc-r2-659:-NEMnY8xBq5WmPt902K7:bd9qr41) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bf22195ee195d4a0b830a1fa98760ddb8b9aefb3/google/cloud/internal/curl_impl.cc#L659)):*

Good point. There could be data in the user provided buffer as well.
</blockquote></details>

But also in neither.  Maybe you got lucky / unlucky and stopped reading data from libcurl (i.e. when detecting the end of the headers) before a full error message was in the user data and/or the spill buffer.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10046)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/internal/curl_impl.cc` line 659 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10046#-NEIVY8AHOeBEUZPwvWc-r2-659:-NENT4kX0khKd6-IBGdL:b6pqxwi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bf22195ee195d4a0b830a1fa98760ddb8b9aefb3/google/cloud/internal/curl_impl.cc#L659)):*

But also in neither.  Maybe you got lucky / unlucky and stopped reading data from libcurl (i.e. when detecting the end of the headers) before a full error message was in the user data and/or the spill buffer.
</blockquote></details>

So this block of code only executes if `curl_closed_ == true`. The only places we set `curl_closed_ = true` is in `~CurlImpl` and in `CurlImpl::PerformWork` after a call to `curl_multi_perform` sets the number of `running_handles` to `0` indicating that there are no longer any transfers in progress. Considering that our write callback *must* handle all bytes libcurl passes to it or libcurl errors, this seems definitive that there is no more data to be read for this request/response.

coryan:
*[`google/cloud/internal/curl_impl.cc` line 659 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10046#-NEIVY8AHOeBEUZPwvWc-r2-659:-NEajgmc9L1Pfi3XDNEo:b3by7t3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bf22195ee195d4a0b830a1fa98760ddb8b9aefb3/google/cloud/internal/curl_impl.cc#L659)):*

So this block of code only executes if `curl_closed_ == true`. The only places we set `curl_closed_ = true` is in `~CurlImpl` and in `CurlImpl::PerformWork` after a call to `curl_multi_perform` sets the number of `running_handles` to `0` indicating that there are no longer any transfers in progress. Considering that our write callback *must* handle all bytes libcurl passes to it or libcurl errors, this seems definitive that there is no more data to be read for this request/response.
</blockquote></details>

Granted.  But then maybe a previous call to `ReadImpl()` consumed part of the response.  This would still return a `google::cloud::Status()` with a partial payload.


Regardless of whether any of these conditions, we can can make this work correctly by *removing* code, see #10051.   There is no need to convert an HTTP status code to a `google::cloud::Status` because we do that at a higher level:

https://github.com/googleapis/google-cloud-cpp/blob/6d3a1c5e7e8b19a80989f4c1faa03814b22d0836/google/cloud/storage/internal/rest_client.cc#L73-L80

PR #10051, opened 2022-10-17
fix(rest): return complete payloads for errors

### LINKED ISSUE #9947: Rest Error Message are Less Useful Now

coryan:
We need the full payload when returning completing a `HttpRequest` with a non-OK status code. The payload may contain a full ErrorInfo, and we want the application to have access to those.

`rest_internal::CurlImpl` was converting the HTTP status code unless it was in a list of values ignored. Then `storage::RestClient` used a per-method predicate to convert some values to errors.  It seems simpler to *never* convert an HTTP status code to a `Status` in the `CurlImpl` layer. After all, an HTTP request that completes with a status code was able to send and receive the call.

Fixes #9947 

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10051)

### DIFF
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -181,7 +181,6 @@ void CurlImpl::ApplyOptions(Options const& options) {
   transfer_stall_minimum_rate_ = options.get<TransferStallMinimumRateOption>();
   download_stall_timeout_ = options.get<DownloadStallTimeoutOption>();
   transfer_stall_minimum_rate_ = options.get<DownloadStallMinimumRateOption>();
-  ignored_http_error_codes_ = options.get<IgnoredHttpErrorCodes>();
 }
 
 CurlImpl::CurlImpl(CurlHandle handle,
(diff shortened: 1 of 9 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10051)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #10054, opened 2022-10-17
fix(rest): too many debug headers

coryan:
We were injecting a `:curl-peer` pseudo-header (usually used for troubleshooting) for each `Read()` call.  The existing tests did not catch this because this headers were never exposed to the application layer either.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10054)


scotthart (approved):
> Reviewed 6 of 7 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10054)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

Issue #6819, opened 2021-06-21
Printing usage should not result in an error


dbolduc:
In our [base Example class](https://github.com/googleapis/google-cloud-cpp/blob/59d7cf6a8d3b521caf304c274fc127d636d102c3/google/cloud/testing_util/example_driver.cc) and several other places printing the usage results in a shell error.

This is not very important, but it seems mean to instruct users to run a command that results in an error:
https://github.com/googleapis/google-cloud-cpp/blob/59d7cf6a8d3b521caf304c274fc127d636d102c3/google/cloud/spanner/samples/README.md#L55-L59

As far as I can tell we want a successful print usage call to look like this:
```console
$ ./target --help
Usage: blah blah blah
$ echo $?
0
```

And for the case of examples, which have commands, we do the following to successfully print usage for a given command:
```console
$ ./example command --help
```

devjgm:
Punt once.

coryan:
Strike 2

Dhruv1107:
Hi @dbolduc and @devjgm ,
This issue seems to be a good initiation of my journey towards contribution to FOSS projects.
Can you please assign this to me?

dbolduc:
> Can you please assign this to me?

Happily.

<hr />

Let's start by just building and running an example. I'd pick the pubsub samples, which uses the `google::cloud::testing_util::Example` class:
https://github.com/googleapis/google-cloud-cpp/blob/eb5ad369a7c5fc35aba1f4678eed44b686610c5e/google/cloud/pubsub/samples/samples.cc#L2375
https://github.com/googleapis/google-cloud-cpp/blob/eb5ad369a7c5fc35aba1f4678eed44b686610c5e/google/cloud/pubsub/samples/samples.cc#L2521

We can run the example with [Bazel](https://bazel.build/) using this command:  (Warning: the initial build will be *really* slow)

```sh
$ bazel run //google/cloud/pubsub/samples:samples

# Output
Missing command
Usage: samples <command> [arguments]

Commands:
    create-avro-schema <project-id> <schema-id>
    ....etc....
```

We can then print the result of the previous command
```sh
$ echo $?

# Output
1
```

When that is working, we are ready to start solving the problem. We like that the program exits with a failed status when no commands are supplied. But we want to be able to pass it a `--help` and have it both
* print the usage
* return a 0 (success)

```sh
$ bazel run //google/cloud/pubsub/samples:samples -- --help

# Output
Usage: blah blah blah....
[...]

PR #10057, opened 2022-10-18
cleanup: change exit status to 0 when --help is used

Dhruv1107:
Hey @dbolduc, I have cleaned up the code for the issue "[Printing usage should not result in an error](https://github.com/googleapis/google-cloud-cpp/issues/6819)" .

Can you please review the changes?

Cheers,
Dhruv

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10057)


google/cloud/testing_util/example_driver.cc
```diff
@@ -61,10 +61,14 @@ int Example::Run(int argc, char const* const argv[]) try {
 
   std::string const command_name = argv[1];
   auto command = commands_.find(command_name);
-  if (commands_.end() == command) {
+  if (commands_.end() == command && command_name != "--help") {
     PrintUsage(argv[0], "Unknown command: " + command_name);
     return 1;
   }
+  else if(commands_.end() == command && command_name == "--help") {
+    PrintUsage(argv[0], "");
```
devbww, line 69: This is the first time we'll be passing an empty `msg` to `PrintUsage()`, which means it will output a blank line.  We should probably avoid that somehow.

devbww, line 69: Aside: It looks like a number of `testing_util::Example` users do not handle "--help" in their commands (by throwing a `testing_util::Usage`), so `full_usage_` is not going to be complete.

google/cloud/testing_util/example_driver.cc
```diff
@@ -61,10 +61,14 @@ int Example::Run(int argc, char const* const argv[]) try {
...
+    return 0;
+  }
 
   command->second({argv + 2, argv + argc});
```
devbww, line 73: In the `catch (Usage const& u)` case for this (i.e., `command --help`), I presume we also want to `return 0`.

google/cloud/testing_util/example_driver.cc
```diff
@@ -61,10 +61,14 @@ int Example::Run(int argc, char const* const argv[]) try {
 
   std::string const command_name = argv[1];
   auto command = commands_.find(command_name);
-  if (commands_.end() == command) {
+  if (commands_.end() == command && command_name != "--help") {
     PrintUsage(argv[0], "Unknown command: " + command_name);
     return 1;
   }
+  else if(commands_.end() == command && command_name == "--help") {
```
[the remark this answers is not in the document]
Dhruv1107: Done

google/cloud/testing_util/example_driver.cc
Dhruv1107: Done

google/cloud/testing_util/example_driver.cc
Dhruv1107: Done

google/cloud/testing_util/example_driver.cc
```diff
@@ -85,9 +89,16 @@ int Example::Run(int argc, char const* const argv[]) try {
 void Example::PrintUsage(std::string const& cmd, std::string const& msg) {
   auto last_slash = cmd.find_last_of('/');
   auto program = cmd.substr(last_slash + 1);
+  if(msg.length() != 0) {
```
devbww, line 92: It would be better to factor out the common parts of the if/else branches.  It also doesn't look like we need an extra newline after `full_usage_`.  And, prefer `!empty()` over `length() != 0`.  So, ...
```
  if (!msg.empty()) std::cerr << msg << "\n";
  std::cerr << "Usage: " << program << " <command> [arguments]\n\n"
            << "Commands:\n" << full_usage_;
```

google/cloud/testing_util/example_driver.cc
```diff
@@ -61,17 +61,21 @@ int Example::Run(int argc, char const* const argv[]) try {
 
   std::string const command_name = argv[1];
   auto command = commands_.find(command_name);
-  if (commands_.end() == command) {
-    PrintUsage(argv[0], "Unknown command: " + command_name);
-    return 1;
-  }
+    if (commands_.end() == command) {
```
devbww, line 64: Fix indentation.

Are you able to run the "checkers" build to automatically fix these?

google/cloud/testing_util/example_driver.cc
devbww: Sorry about not being clearer.  Indeed, I don't think there is anything actionable for you here.  It is more a reminder for us that any user of `testing_util::Example` is expected to handle "--help" for all their non-auto commands.

google/cloud/testing_util/example_driver.cc
Dhruv1107: I have fixed the indentation. How do I run "checkers" build, do we have any documentation for that?

coryan: https://github.com/googleapis/google-cloud-cpp/tree/main/doc/contributor

google/cloud/testing_util/example_driver.cc
```diff
@@ -61,17 +61,21 @@ int Example::Run(int argc, char const* const argv[]) try {
...
+        return 1;
+    }
 
   command->second({argv + 2, argv + argc});
 
   return 0;
 } catch (Usage const& u) {
   PrintUsage(argv[0], u.what());
-  return 1;
+  return 0;
```
devbww, line 78: Perhaps this was bad advice.  It causes `//google/cloud/testing_util:example_driver_test` to fail, at least.

I'm not sure where that leaves us though, because I imagine that we want `samples command foo` to fail (when "foo" is a bad argument), but #6819 seems to want `samples command --help` to succeed, in addition to `samples --help`.

For example, with this PR applied we see ...
```
$ bazel-bin/google/cloud/iam/samples/iam_samples --help
Usage: iam_samples <command> [arguments]

Commands:
    :
$ echo $?
0
```
Good.  But ...
```
$ bazel-bin/google/cloud/iam/samples/iam_samples list-service-accounts
list-service-accounts <project-id>
Usage: iam_samples <command> [arguments]

Commands:
    :
$ echo $?
0
```
should probably fail, while ...
```
$ bazel-bin/google/cloud/iam/samples/iam_samples list-service-accounts --help
list-service-accounts <project-id>
Usage: iam_samples <command> [arguments]

Commands:
    :
$ echo $?
0
```
should probably succeed (if we agree with the whole premise).  But they both go through this single `return` because they both throw a `Usage`.

[Aside: The output in those last two "list-service-accounts" commands isn't the greatest either.]

Anyway, maybe @dbolduc, as the originator of #6819, can tell us more about how this should behave.

Dhruv1107:
Hi @dbolduc, can you help us on this issue?

dbolduc:
I agree with @devbww on what the behavior should be.

> can you help us on this issue?

One thing we could do is maintain the usage string for each command separately, instead of concatenating them as `full_usage_` at the start:
https://github.com/googleapis/google-cloud-cpp/blob/59d7cf6a8d3b521caf304c274fc127d636d102c3/google/cloud/testing_util/example_driver.cc#L28-L39

It might be simplest to look for the `--help` argument and **extract** it if it exists. (so `--help` could be anywhere in `argv`). Then if there is an `argv[1]` after that, we assume it is the command. If it is a valid command, we would print the usage for that command only and `return 0`. Else, we would print something like "Unknown Command: blah" then the full usage and `return 0`.

If there is no `--help` and the command is not valid, we should `return 1`.

<hr />

Here are some test cases for you. Assume:
* The executable is named `target`.
* There is one command: `foo`.
* The command `foo` takes one argument.
* The correct usage string would be: `./target foo bar`

```
./target --help
# Prints full usage, returns 0

./target foo --help
# Prints foo usage, returns 0

./target --help foo
# Prints foo usage, returns 0. (equivalent to the above)

./target foo bar --help
# Prints foo usage, returns 0

./target unknown --help
# Prints full usage, returns 0

./target
# Prints full usage, returns 1

./target unknown
# Prints full usage, returns 1

./target foo
# Prints foo usage, returns 1

./target foo bar baz
# Prints foo usage, returns 1
```

devbww:
And I agree with @dbolduc on all those test cases.  My only comment is that it looks like it could take non-trivial changes to the interfaces to realize those results.

dbolduc:
I am closing this PR, because it has not been updated in >30 days.

@Dhruv1107 if you are still working on this, please feel free to reopen the PR.

Dhruv1107:
@dbolduc I don't have the permission to reopen the PR? Can you please reopen it?

Dhruv1107:
Hi @devbww, I have made the changes for all the extra test cases that you have mentioned. Can you please review?

devbww:
> Hi @devbww, I have made the changes for all the extra test cases that you have mentioned. Can you please review?

It looks like //google/cloud/testing_util:example_driver_test is failing.  It would be nice is everything was passing before we spend much time reviewing.

Dhruv1107:
@devbww I have made the changes to fix the errors in //google/cloud/testing_util:example_driver_test.

coryan:
No activity in 30d.  Closing.

Issue #7187, opened 2021-08-17
Create synchronous / blocking pull


coryan:
We have a trickle of customer requests for synchronous / blocking pulls.  Whether they are a great idea or not is somewhat irrelevant, customers need them.  The motivations include, but are not limited to:

* Some applications use blocking pulls with other Pub/Sub services, re-architecting the application is not feasible. 
* Some applications just need to pull a few messages and then exit.

----
## Objective

Provide a "blocking pull" API for C++ customers needing a simpler interface to the service.

## Background

We currently offer "streaming pull" APIs for Cloud Pub/Sub.  The application provides a callback, and the library invokes the callback when events are received.  There are two versions, one for regular subscriptions:

https://github.com/googleapis/google-cloud-cpp/blob/55b3648aead2f57a32ba684f77b6ad0609b03b8e/google/cloud/pubsub/subscriber.h#L125

and one for subscriptions where the application expects "exactly once" semantics:

https://github.com/googleapis/google-cloud-cpp/blob/55b3648aead2f57a32ba684f77b6ad0609b03b8e/google/cloud/pubsub/subscriber.h#L157

Note that both can receive messages with exactly once semantics. And (if needed) the `*AckHandler` will perform retries for messages with exactly once semantics, even if the application uses the simpler API.

We have two versions of this API because exactly-once was added later, and the original `AckHandler` cannot report errors:

https://github.com/googleapis/google-cloud-cpp/blob/55b3648aead2f57a32ba684f77b6ad0609b03b8e/google/cloud/pubsub/ack_handler.h#L65

Furthermore, changing it to return an error would have been a breaking change.

The question arises as to which of these two APIs to offer for blocking pulls.  Or whether to offer both.

## Overview

We propose using a single API, but one that returns errors.  A new function in the subscriber would look like so:

```cc
namespace google::cloud::pubsub { // implicit unless otherwise noticed
class Subscriber {
 public:
  ... ... ...
  std::pair<PullAckHandler, Message> Pull();
};
```

The `Message` class already exists, the `PullAckHandler` is new.  It will meet the following interface:

```cc
class google::cloud::pubsub::PullAckHandler {
 public:
  future<Status> ack() &&;
  future<Status> nack() &&;
  std::int32_t delivery_attempt() const;
```

The reader may note that this matches the interface as `ExactlyOnceAckHandler`:

[...]

PR #10072, opened 2022-10-19
cleanup(pubsub): signature for bidir streaming RPC

coryan:
Use the generator signature for `StreamingPull`, the only bidir streaming RPC in this service. The affected function is in `pubsub_internal`, so the change is non-breaking, but it does require changing a large number of tests.

Part of the work for #7187

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10072)


devbww (approved):
Squintingly.

PR #10079, opened 2022-10-20
ci: enable renovate PRs for pre-releases

coryan:
Testing with pre-releases will help gRPC, Protobuf, and Abseil detect integration problems before they create a release. It may also help us if we need a fix in "the next release", but something else is broken.

If this becomes too noisy we can just disable unstable releases.


devjgm (approved):
I'm not an owner anymore, but this LGTM.

dbolduc (approved):

PR #10084, opened 2022-10-21
cleanup: use `git ls-files` to filter files

coryan:
When possible, use `git ls-files` to filter files by extension or name. It should be possible to use `--exclude` to exclude files, but I left that for another day.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10084)


dbolduc (approved):
TIL I do not need to type out `--include ...`

PR #10085, opened 2022-10-21
cleanup(generator): separate golden from hand-crafted files

coryan:
Move the hand-crafted files out of the
`generator/integration_tests/golden/` subdirectory, including CMake and Bazel support files.  I think this will make it easier to work in these files, as they will format and work as regular files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10085)


generator/integration_tests/BUILD.bazel
```diff
@@ -59,6 +59,70 @@ cc_grpc_library(
...
+        "golden/internal/*.cc",
+    ]),
+)
+
+filegroup(
+    name = "golden_hdrs",
+    srcs = glob([
+        "golden/*.h",
+        "golden/internal/*.h",
+        "golden/mocks/*.h",
```
dbolduc, line 78: should these be split out into their own `google_cloud_cpp_generator_golden_mocks` library?

generator/integration_tests/BUILD.bazel
```diff
@@ -59,6 +59,70 @@ cc_grpc_library(
     name = "google_cloud_cpp_generator_testing_cc_grpc",
     srcs = [":google_cloud_cpp_generator_testing_proto"],
     grpc_only = True,
-    visibility = [":__subpackages__"],
     deps = [":google_cloud_cpp_generator_testing_cc_proto"],
 )
+
+filegroup(
```
dbolduc, line 65: TIL. `filegroup`. It looks tidier than what we currently do with: `SOURCE_GLOB = "**/*.cc"` in our generated `BUILD.bazel`s. I do not think it is worth updating though.

generator/integration_tests/CMakeLists.txt
```diff
@@ -44,12 +51,102 @@ target_link_libraries(
...
-# to build all the tests everytime they create a new package or when the package
-# is installed from source.
-if (BUILD_TESTING)
-    add_subdirectory(golden)
-endif (BUILD_TESTING)
+file(
+    GLOB google_cloud_cpp_generator_golden_files
+    RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}"
+    "golden/*.h" "golden/*.cc" "golden/internal/*.h" "golden/internal/*.cc"
+    "golden/mocks/*.h")
```
dbolduc, line 61: same comment about splitting out the mocks. it's fine if you don't think this is worth doing.

generator/integration_tests/CMakeLists.txt
```diff
@@ -44,12 +51,102 @@ target_link_libraries(
...
+    tests/golden_thing_admin_stub_test.cc
+    tests/plumbing_test.cc)
+
+# Export the list of unit tests to a .bzl file so we do not need to maintain the
+# list in two places.
+export_list_to_bazel("google_cloud_cpp_generator_golden_tests.bzl"
+                     "google_cloud_cpp_generator_golden_tests" YEAR "2020")
+
+# Create a custom target so we can say "build all the tests"
+add_custom_target(generator-golden-tests)
```
dbolduc, line 127: hm do we do this anywhere else? is it necessary?

dbolduc (approved):

generator/integration_tests/CMakeLists.txt
coryan: Removed.

generator/integration_tests/BUILD.bazel
coryan: Ack

generator/integration_tests/BUILD.bazel
coryan: I will leave these as-is until I find a reason to split them.

generator/integration_tests/CMakeLists.txt
coryan: Ack

PR #10093, opened 2022-10-24
feat(generator): optimize round-robin decorators

coryan:
Thanks to @devbww for the suggestion

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10093)

### DIFF
--- a/generator/internal/round_robin_decorator_generator.cc
+++ b/generator/internal/round_robin_decorator_generator.cc
@@ -372,9 +372,10 @@ $round_robin_class_name$::AsyncCancelOperation(
   CcPrint(R"""(
 std::shared_ptr<$stub_class_name$>
 $round_robin_class_name$::Child() {
-  std::lock_guard<std::mutex> lk(mu_);
-  auto child = children_[current_];
+  std::unique_lock<std::mutex> lk(mu_);
+  auto const& child = children_[current_];
   current_ = (current_ + 1) % children_.size();
+  lk.unlock();
   return child;
 }
 )""");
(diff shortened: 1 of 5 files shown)


generator/internal/round_robin_decorator_generator.cc
```diff
@@ -372,9 +372,10 @@ future<Status>
   CcPrint(R"""(
 std::shared_ptr<$stub_class_name$>
 $round_robin_class_name$::Child() {
-  std::lock_guard<std::mutex> lk(mu_);
-  auto child = children_[current_];
+  std::unique_lock<std::mutex> lk(mu_);
+  auto const& child = children_[current_];
```
devbww, line 376: And perhaps `auto const current = current_;` and then `return children_[current];` would move some more arithmetic outside the lock.

generator/internal/round_robin_decorator_generator.cc
```diff
@@ -372,9 +372,10 @@ future<Status>
...
   current_ = (current_ + 1) % children_.size();
```
devbww, line 377: You might also consider strength reduction to ...
```
  if (++current_ == children_.size()) current_ = 0;
```

devbww (approved):
Some more nits on the nit, if they're convincing.

generator/internal/round_robin_decorator_generator.cc
coryan: I think those would make no difference with modern compilers:

https://godbolt.org/z/vaa5Yq3hr

But if you think they can make the intent better, I would be happy to use that.

generator/internal/round_robin_decorator_generator.cc
coryan: Ditto

coryan: I am going to go with your approach.  easier to read I think.

Issue #10095, opened 2022-10-24
add tests for PG "JSONB[]" when supported


devbww:
Support for column types that are arrays of PG JSONB values will be delayed.  That is, scalar JSONB values will be released first.

This issue will be used to mark code that needs to be updated when `JSONB[]` support is released.

PR #10099, opened 2022-10-25
test(spanner): loosen message expectation on column not found

devbww:
It looks like different versions/configs of the backend respond in slightly different ways, so loosen the #10095 workaround expectation. In particular, this accounts for the current state of the staging service.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10099)

### DIFF
--- a/google/cloud/spanner/integration_tests/data_types_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/data_types_integration_test.cc
@@ -417,8 +417,9 @@ TEST_F(PgDataTypeIntegrationTest, WriteReadArrayJson) {
   auto result = WriteReadData(*client_, data, "ArrayJsonValue");
   {
     // TODO(#10095): Remove this when JSONB[] is supported.
-    auto matcher = testing_util::StatusIs(
-        StatusCode::kNotFound, testing::HasSubstr("Column not found in table"));
+    auto matcher = StatusIs(StatusCode::kNotFound,
+                            AnyOf(HasSubstr("Column not found in table"),
+                                  HasSubstr("is not a column in")));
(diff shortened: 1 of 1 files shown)


dbolduc (approved):

google/cloud/spanner/integration_tests/data_types_integration_test.cc
```diff
@@ -417,8 +417,9 @@ TEST_F(PgDataTypeIntegrationTest, WriteReadArrayJson) {
   auto result = WriteReadData(*client_, data, "ArrayJsonValue");
   {
     // TODO(#10095): Remove this when JSONB[] is supported.
-    auto matcher = testing_util::StatusIs(
-        StatusCode::kNotFound, testing::HasSubstr("Column not found in table"));
+    auto matcher = StatusIs(StatusCode::kNotFound,
```
coryan, line 420: Or maybe not match anything about the string?  Is the ErrorInfo any better?

coryan (approved):

google/cloud/spanner/integration_tests/data_types_integration_test.cc
devbww: > Or maybe not match anything about the string?

Maybe, but I think we're probably best off being more specific, particularly in the case of a temporary workaround.

> Is the ErrorInfo any better?

I did teach the logger how to try to extract a `ResourceInfo` on a `kNotFound` error, but there was nothing in the log in this case, so its presence is at least something that we can't depend on.  [Aside: We may want to think about `testing_util/status_matchers.h` support for looking into `Status` payloads.]

So, for better or worse, I think I'll leave it as-is for now.

PR #10103, opened 2022-10-26
cleanup(bigtable): use generator for round robin stub

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10103)


coryan (approved):
That worked better than I expected.

Issue #9016, opened 2022-05-19
Sample code and quickstarts should print full errors


dbolduc:
Our sample/quickstart code only prints the error message, discarding the error code and error info. They look something like this (in the `StatusOr<>` case):
https://github.com/googleapis/google-cloud-cpp/blob/08f1d4a76d92d79c8a5200a3c13e4453e8ad29a4/google/cloud/accessapproval/quickstart/quickstart.cc#L31

Printing the full error gives customers more information to debug issues they might encounter. Our sample/quickstart code should reflect the code we want our customers to write.

We can just make this public: https://github.com/googleapis/google-cloud-cpp/blob/08f1d4a76d92d79c8a5200a3c13e4453e8ad29a4/google/cloud/status.cc#L22-L26
... and do a tedious find and replace.

- [x] Fix `examples/`
- [x] Fix `google/cloud/bigquery/samples/`
- [x] Fix `google/cloud/iam/samples/`
- [x] Fix `google/cloud/pubsub/samples/`
- [x] Fix `google/cloud/bigtable/examples/`
- [x] Fix `google/cloud/storage/examples/`

devbww:
How about throwing the whole `Status` value, and then just streaming it when caught?

coryan:
We will start fixing these as we see them.

coryan:
Consider breaking this bug into smaller bugs (one per service?) to make them easier to handle.

Qarthak:
@coryan @devbww Is this issue closed? I'm new to open source and coming from a C/C++ background I think I can work on this. But from what I know, a merge means the issue has been handled right?

coryan:
> But from what I know, a merge means the issue has been handled right?

Some teams break down larger bugs into smaller issues that can be solved in a single PRs.  Some go as far as *requiring* that a PR fixes exactly one bug.  We do neither.  Many issues are resolved by a single PR, some larger issues require multiple small PRs to resolve.

Generally, we let each person follow whatever approach works best for organizing / planning their development.

@Qarthak if you need some help breaking this down, this outline may help:

- [ ] Fix `examples/`
- [ ] Fix `google/cloud/bigquery/samples/`
- [ ] Fix `google/cloud/iam/samples/`
- [ ] Fix `google/cloud/pubsub/samples/`
- [ ] Fix `google/cloud/bigtable/examples/`
- [ ] Fix `google/cloud/storage/examples/`

I got that using:

```
[...]

PR #10108, opened 2022-10-27
cleanup(examples): throw full Status values

Qarthak:
Throw the full status as the error, instead of creating `std::runtime_error` with just the message of the status

Fixed this issue(#9016) for examples folder

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10108)

### DIFF
--- a/examples/grpc_credential_types.cc
+++ b/examples/grpc_credential_types.cc
@@ -100,7 +100,7 @@ google::iam::credentials::v1::GenerateAccessTokenResponse UseAccessToken(
     auto token = client.GenerateAccessToken(
         "projects/-/serviceAccounts/" + service_account, /*delegates=*/{},
         /*scope=*/{"https://www.googleapis.com/auth/cloud-platform"}, duration);
-    if (!token) throw std::runtime_error(token.status().message());
+    if (!token) throw token.status();
 
     auto const expiration =
         std::chrono::system_clock::from_time_t(token->expire_time().seconds());
@@ -119,7 +119,7 @@ google::iam::credentials::v1::GenerateAccessTokenResponse UseAccessToken(
                 credentials)));
     for (auto instance :
          admin.ListInstances(google::cloud::Project(project_id).FullName())) {
-      if (!instance) throw std::runtime_error(instance.status().message());
+      if (!instance) throw instance.status();
       std::cout << "Instance: " << instance->name() << "\n";
     }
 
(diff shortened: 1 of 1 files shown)


examples/grpc_credential_types.cc
```diff
@@ -100,7 +100,7 @@ google::iam::credentials::v1::GenerateAccessTokenResponse UseAccessToken(
     auto token = client.GenerateAccessToken(
         "projects/-/serviceAccounts/" + service_account, /*delegates=*/{},
         /*scope=*/{"https://www.googleapis.com/auth/cloud-platform"}, duration);
-    if (!token) throw std::runtime_error(token.status().message());
+    if (!token) throw token.status();
```
coryan, line 103: Blegh, `clang-tidy` is unhappy about this:

https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/10108/7f850251b5795674be31a845ed64d3d5b6d9f4b4/fedora-36-clang-tidy/log-13dfbf57-59cd-4d1a-a9cd-a1f82e42f7ce.txt

This seems to fix the error for me:

```cc
  if (!token) throw std::move(token).status();
```

Make the same change below too.

devbww: Aside: It looks like there is nothing here to catch these exceptions, so throwing a `Status` will probably result in less information being reported, contrary to the goal.

Qarthak: Oh yeah you're right. Should we add something like this towards the end?
https://github.com/googleapis/google-cloud-cpp/blob/69561feebfd679a6546a829764e711bb93d96665/google/cloud/spanner/samples/postgresql_samples.cc#L1068-L1075

Maybe we can skip catching the std::exception as even if it's not caught, we get a
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to create CurlHandle
Aborted
```
instead of a
```
Failed to create CurlHandle
```

devbww: Yes, something like that.

I'm not sure of the sensibilities of skipping the `std::exception` handler.  You're correct that we see the same information either way, but I reckon the answer depends on whether we think of "Failed to create CurlHandle" (or any of the other `runtime_error` cases) as an `exit(1)` or an `abort()`.

I'd vote for a normal termination, albeit indicating an error (i.e., `exit(1)`), but lets see if anyone else has an opinion.

coryan: What @devbww said sounds good to me.

dbolduc:
I am closing this PR, because it has not been updated in >30 days.

@Qarthak if you are still working on this, please feel free to reopen the PR.

PR #10109, opened 2022-10-27
impl(generator): extract additional info from http option for rest

scotthart:
Expanded parsing of the http option in proto files needed for generating rest stubs.

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -244,27 +244,54 @@ void SetMethodSignatureMethodVars(
   }
 }
 
-void SetResourceRoutingMethodVars(
+void SetHttpResourceRoutingMethodVars(
     google::protobuf::MethodDescriptor const& method,
     VarsDictionary& method_vars) {
-  auto result = ParseResourceRoutingHeader(method);
-  if (result.has_value()) {
-    method_vars["method_request_url_path"] = result->url_path;
-    method_vars["method_request_url_substitution"] = result->url_substitution;
-    std::string param = result->param_key;
-    method_vars["method_request_param_key"] = param;
-    std::vector<std::string> chunks;
-    auto const* input_type = method.input_type();
-    for (auto const& sv : absl::StrSplit(param, '.')) {
-      auto const chunk = std::string(sv);
-      auto const* chunk_descriptor = input_type->FindFieldByName(chunk);
-      chunks.push_back(FieldName(chunk_descriptor));
-      input_type = chunk_descriptor->message_type();
+  struct HttpInfoVisitor {
+    HttpInfoVisitor(google::protobuf::MethodDescriptor const& method,
+                    VarsDictionary& method_vars)
+        : method(method), method_vars(method_vars) {}
+
+    void operator()(HttpExtensionInfo const& info) {
+      method_vars["method_request_url_path"] = info.url_path;
+      method_vars["method_request_url_substitution"] = info.url_substitution;
+      std::string param = info.request_field_name;
+      method_vars["method_request_param_key"] = param;
+      std::vector<std::string> chunks;
+      auto const* input_type = method.input_type();
+      for (auto const& sv : absl::StrSplit(param, '.')) {
+        auto const chunk = std::string(sv);
+        auto const* chunk_descriptor = input_type->FindFieldByName(chunk);
+        chunks.push_back(FieldName(chunk_descriptor));
+        input_type = chunk_descriptor->message_type();
+      }
+      method_vars["method_request_param_value"] =
+          absl::StrJoin(chunks, "().") + "()";
+      method_vars["method_request_body"] = info.body;
+      method_vars["method_http_verb"] = info.http_verb;
+      method_vars["method_rest_path"] =
+          absl::StrCat("absl::StrCat(\"", info.path_prefix, "\",request.",
+                       method_vars["method_request_param_value"], ",\"",
+                       info.path_suffix, "\")");
     }
-    method_vars["method_request_param_value"] =
-        absl::StrJoin(chunks, "().") + "()";
-    method_vars["method_request_body"] = result->body;
-  }
+
+    void operator()(HttpSimpleInfo const& info) {
+      method_vars["method_http_verb"] = info.http_verb;
+      method_vars["method_request_param_value"] = method.full_name();
+      method_vars["method_rest_path"] = absl::StrCat("\"", info.url_path, "\"");
+    }
+
+    void operator()(absl::monostate) {
+      method_vars["method_http_verb"] = method.full_name();
+      method_vars["method_request_param_value"] = method.full_name();
+      method_vars["method_rest_path"] = method.full_name();
+    }
+    google::protobuf::MethodDescriptor const& method;
+    VarsDictionary& method_vars;
+  };
+
+  auto info = ParseHttpExtension(method);
+  absl::visit(HttpInfoVisitor(method, method_vars), info);
 }
 
 std::string DefaultIdempotencyFromHttpOperation(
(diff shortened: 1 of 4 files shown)


AdrianaDJ (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

AdrianaDJ (commented):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

AdrianaDJ (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109)* status: all files reviewed, 1 unresolved discussion (waiting on @scotthart)

___
*[`generator/internal/descriptor_utils.h` line 98 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109#-NFPhIFoCNGWAttqyVYe:-NFPhIFp08QQoJgfXk0o:blvcawl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e5fce94115ceb16e2231c69b724999c8de18b91f/generator/internal/descriptor_utils.h#L98)):*
> ```objc
>  */
> std::string FormatMethodCommentsProtobufRequest(
>     google::protobuf::MethodDescriptor const& method);
> ```

firebase-adminsdk-kvtrl@titanium-genre-213714.iam.gserviceaccount.com

generator/internal/descriptor_utils.cc
```diff
@@ -244,27 +244,54 @@ void SetMethodSignatureMethodVars(
...
+    void operator()(HttpSimpleInfo const& info) {
+      method_vars["method_http_verb"] = info.http_verb;
+      method_vars["method_request_param_value"] = method.full_name();
+      method_vars["method_rest_path"] = absl::StrCat("\"", info.url_path, "\"");
+    }
+
+    void operator()(absl::monostate) {
+      method_vars["method_http_verb"] = method.full_name();
+      method_vars["method_request_param_value"] = method.full_name();
+      method_vars["method_rest_path"] = method.full_name();
```
dbolduc, line 287: 1. TIL `absl::monostate`
2. These values kinda look like nonsense. Do we have to set them? If so, should we test this case?

generator/internal/descriptor_utils.cc
```diff
@@ -244,27 +244,54 @@ void SetMethodSignatureMethodVars(
...
+        auto const chunk = std::string(sv);
+        auto const* chunk_descriptor = input_type->FindFieldByName(chunk);
+        chunks.push_back(FieldName(chunk_descriptor));
+        input_type = chunk_descriptor->message_type();
+      }
+      method_vars["method_request_param_value"] =
+          absl::StrJoin(chunks, "().") + "()";
+      method_vars["method_request_body"] = info.body;
+      method_vars["method_http_verb"] = info.http_verb;
+      method_vars["method_rest_path"] =
```
dbolduc, line 272: should we be verifying the value of `method_rest_path`?

generator/internal/descriptor_utils.cc
```diff
@@ -244,27 +244,54 @@ void SetMethodSignatureMethodVars(
...
+                       info.path_suffix, "\")");
     }
-    method_vars["method_request_param_value"] =
-        absl::StrJoin(chunks, "().") + "()";
-    method_vars["method_request_body"] = result->body;
-  }
+
+    void operator()(HttpSimpleInfo const& info) {
+      method_vars["method_http_verb"] = info.http_verb;
+      method_vars["method_request_param_value"] = method.full_name();
```
dbolduc, line 280: why don't we need to do the same `chunks` shenanigans that we do for the `HttpExtensionInfo` case above?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @AdrianaDJ and @dbolduc)

___
*[`generator/internal/descriptor_utils.cc` line 272 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109#-NFP_DchNZSV7rGew_IO-r2-272:-NFUrAxwANY3sObOLBOG:bqtrlxr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f07bb2580bcdb7c7cc8efc3f6caed8bdfc857c5b/generator/internal/descriptor_utils.cc#L272)):*

should we be verifying the value of `method_rest_path`?
</blockquote></details>

Yes. Checks added to the existing tests.
___
*[`generator/internal/descriptor_utils.cc` line 280 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109#-NFP_DchNZSV7rGew_IO-r2-280:-NFUqoNn7ObZBNSc2_Vv:bibp4ci) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f07bb2580bcdb7c7cc8efc3f6caed8bdfc857c5b/generator/internal/descriptor_utils.cc#L280)):*

why don't we need to do the same `chunks` shenanigans that we do for the `HttpExtensionInfo` case above?
</blockquote></details>

I added comments to all 3 visitors to better explain their function.
___
*[`generator/internal/descriptor_utils.cc` line 287 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109#-NFP_DchNZSV7rGew_IO-r2-287:-NFUqtWl6i9iPM249GQN:b-run58x) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f07bb2580bcdb7c7cc8efc3f6caed8bdfc857c5b/generator/internal/descriptor_utils.cc#L287)):*

1. TIL `absl::monostate`
2. These values kinda look like nonsense. Do we have to set them? If so, should we test this case?
</blockquote></details>

It's useful as a diagnostic that we've encountered an url we don't parse correctly.

dbolduc (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @AdrianaDJ and @scotthart)

___
*[`generator/internal/descriptor_utils.cc` line 280 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109#-NFP_DchNZSV7rGew_IO-r2-280:-NFV846o6MSgADAPJ7UB:b1ztq4c) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f07bb2580bcdb7c7cc8efc3f6caed8bdfc857c5b/generator/internal/descriptor_utils.cc#L280)):*

I added comments to all 3 visitors to better explain their function.
</blockquote></details>

Thanks. I am still slightly confused by the output in the simple case, but it is probably because I am dumb. Should this one end with a "()" like the other one does?

Also I think this was the service that prompted us to add the `FindFieldByName()` logic: https://servicedirectory.googleapis.com/$discovery/rest?version=v1 . I wonder if it works... but even if it doesn't we can always fix it later.
___
*[`generator/internal/descriptor_utils.cc` line 287 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109#-NFP_DchNZSV7rGew_IO-r2-287:-NFV8nriCxh3oVJu6olB:b-hyrc3t) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f07bb2580bcdb7c7cc8efc3f6caed8bdfc857c5b/generator/internal/descriptor_utils.cc#L287)):*

It's useful as a diagnostic that we've encountered an url we don't parse correctly.
</blockquote></details>

Ack, thanks. I will defer to you on whether its better to have the generator break while emitting code, or while compiling the emitted code.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @AdrianaDJ and @dbolduc)

___
*[`generator/internal/descriptor_utils.cc` line 280 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109#-NFP_DchNZSV7rGew_IO-r2-280:-NFVAf5O0WvADNqC16r4:beiion7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f07bb2580bcdb7c7cc8efc3f6caed8bdfc857c5b/generator/internal/descriptor_utils.cc#L280)):*

Thanks. I am still slightly confused by the output in the simple case, but it is probably because I am dumb. Should this one end with a "()" like the other one does?

Also I think this was the service that prompted us to add the `FindFieldByName()` logic: https://servicedirectory.googleapis.com/$discovery/rest?version=v1 . I wonder if it works... but even if it doesn't we can always fix it later.
</blockquote></details>

There's no need for () because no accessor method on a proto field is being called to format the string.
___
*[`generator/internal/descriptor_utils.cc` line 287 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10109#-NFP_DchNZSV7rGew_IO-r2-287:-NFVAuvU0GUJ7woB05AX:b-8uo695) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f07bb2580bcdb7c7cc8efc3f6caed8bdfc857c5b/generator/internal/descriptor_utils.cc#L287)):*

Ack, thanks. I will defer to you on whether its better to have the generator break while emitting code, or while compiling the emitted code.
</blockquote></details>

Better? I'm undecided on this as well. Currently this is the simplest way to indicate failure, as the caller returns void. This could be refactored later if we deem it useful to propagate up a Status or some such.

dbolduc (approved):

Issue #10114, opened 2022-10-28
Generate basic samples for `*Client` classes


coryan:
We need simple examples showing how to make the most common configuration changes to generated files:

- [x] Change the default endpoint
- [x] Change the default credentials
- [ ] Change the retry or backoff policy
- [ ] Change the polling policy for LRO

PR #10121, opened 2022-10-28
cleanup: use more precise globs for generated files

coryan:
The previous globs would capture too many sources and headers.

Part of the work for #10114

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10121)

### DIFF
--- a/generator/internal/scaffold_generator.cc
+++ b/generator/internal/scaffold_generator.cc
@@ -259,24 +259,37 @@ package(default_visibility = ["//visibility:private"])
 
 licenses(["notice"])  # Apache 2.0
 
-SOURCE_GLOB = "**/*.cc"
+filegroup(
+    name = "srcs",
+    srcs = glob([
+        "*.cc",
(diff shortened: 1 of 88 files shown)


generator/internal/scaffold_generator.cc
```diff
@@ -259,24 +259,37 @@ package(default_visibility = ["//visibility:private"])
...
-    hdrs = glob(
-        include = [HEADER_GLOB],
-        exclude = [MOCK_HEADER_GLOB],
-    ),
+    srcs = [
+        ":srcs",
+    ],
+    hdrs = [
+        ":hdrs",
+    ],
```
devbww, line 292: If you remove the trailing comma from the one element lists, these will format down to one line ... e.g., `srcs = [":srcs"],`.

coryan: Done.

devbww (approved):

PR #10125, opened 2022-10-28
feat(generator): build scripts for samples

coryan:
The scaffold generator should also generate CMake and Bazel build scripts for client samples.

Part of the work for #10114

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10125)

### DIFF
--- a/generator/internal/scaffold_generator.h
+++ b/generator/internal/scaffold_generator.h
@@ -74,6 +74,10 @@ void GenerateQuickstartBuild(
     std::ostream& os, std::map<std::string, std::string> const& variables);
 void GenerateQuickstartBazelrc(
     std::ostream& os, std::map<std::string, std::string> const& variables);
+void GenerateSamplesBuild(std::ostream& os,
+                          std::map<std::string, std::string> const& variables);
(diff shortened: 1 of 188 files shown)


generator/internal/scaffold_generator.h
```diff
@@ -74,6 +74,10 @@ void GenerateQuickstartBuild(
     std::ostream& os, std::map<std::string, std::string> const& variables);
 void GenerateQuickstartBazelrc(
     std::ostream& os, std::map<std::string, std::string> const& variables);
+void GenerateSamplesBuild(std::ostream& os,
```
devbww, line 77: s/Build/Bazel/ perhaps??

devbww (approved):

generator/internal/scaffold_generator.h
coryan: I will let it be. We already have a `GenerateQuickstartBuild()` function, and calling that `GenerateQuickstartBazel()` would be confusing considering we also have `GenerateQuickstartWorkspace()`.

PR #10129, opened 2022-10-29
doc: add endpoint override snippets to generated libs

coryan:
Modify the landing pages for all the generated libraries to include a
snippet showing how to override the endpoint for the `*Client`.

In libraries where there is more than one `*Client` injecting a
snippet for each `*Client` looked too cluttered. The script injects the
first (in alphabetical order) snippet into the landing page. The
remaining snippets are referenced from that page, but are in
separate pages.

Part of the work for #10114 

The main page for a service now looks like this (click on the image to embiggen):

image

While each one of the referenced pages looks like this:

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10129)

### DIFF
--- a/ci/generate-markdown/update-library-landing-dox.sh
+++ b/ci/generate-markdown/update-library-landing-dox.sh
@@ -27,8 +27,9 @@ fi
 cd "${PROJECT_ROOT}"
 readonly LIBRARY=$1
 
-lib="google/cloud/${LIBRARY}"
-if [[ ! -r "${lib}/doc/main.dox" ]]; then
+readonly LIB="google/cloud/${LIBRARY}"
+readonly MAIN_DOX="google/cloud/${LIBRARY}/doc/main.dox"
+if [[ ! -r "${MAIN_DOX}" ]]; then
   exit 0
 fi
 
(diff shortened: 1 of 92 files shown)


ci/generate-markdown/update-library-landing-dox.sh
```diff
@@ -52,10 +53,43 @@ while IFS= read -r -d $'\0' option_defaults_cc; do
...
+      echo
+      echo "Follow these links to find examples for other \\c *Client classes:"
+      for sample_cc in "${samples_cc[@]}"; do
+        sed -n "/main-dox-marker: / s;// main-dox-marker: \(.*\); [\1](@ref \1-endpoint-snippet);p" "${sample_cc}"
+      done
+    fi
+  fi
+  echo
+  sed -n '/<!-- inject-endpoint-snippet-end -->/,$p' "${MAIN_DOX}"
+) | sponge "${MAIN_DOX}"
```
devbww, line 85: I might suggest that these could also be handled with `sed -i`, but that is a question for another day.

devbww (approved):

ci/generate-markdown/update-library-landing-dox.sh
coryan: Ack.

PR #10148, opened 2022-10-31
refactor: make rest internal lib punctuation consistent with other libs

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10148)


coryan (approved):

coryan:
For the record, I think this is Okay because we announced that any symbols (including  target names) with `internal` are not part of the public API.

PR #10150, opened 2022-10-31
chore(generator): add env var to reduce scope of generate-libraries build

scotthart:
Quality of life improvement for doing generator development.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10150)

### DIFF
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -291,6 +291,13 @@ fi
 # Default to --docker mode since no other mode was specified.
 DOCKER_FLAG="true"
 
+# Default to generating all libraries.
+if [ -n "${GENERATE_GOLDEN_ONLY+set}" ]; then
+  GENERATE_GOLDEN_ONLY=${GENERATE_GOLDEN_ONLY:-}
+else
+  GENERATE_GOLDEN_ONLY=""
+fi
+
 # Uses docker to locally build the specified image and run the build command.
 if [[ "${DOCKER_FLAG}" = "true" ]]; then
   io::log_h1 "Starting docker build: ${BUILD_FLAG}"
@@ -346,6 +353,9 @@ if [[ "${DOCKER_FLAG}" = "true" ]]; then
     # Makes the host's gcloud credentials visible inside the docker container,
     # which we need for integration tests.
     "--volume=${HOME}/.config/gcloud:/h/.config/gcloud:Z"
+    # Makes the generate-libraries build ONLY touch golden files in the
+    # generator dir.
+    "--env=GENERATE_GOLDEN_ONLY=${GENERATE_GOLDEN_ONLY}"
   )
   # All GOOGLE_CLOUD_* env vars will be passed to the docker container.
   for e in $(env); do
--- a/ci/cloudbuild/build.sh
+++ b/ci/cloudbuild/build.sh
@@ -191,13 +191,6 @@ export TRIGGER_TYPE
 export VERBOSE_FLAG
 export LOG_LINKER_PAT
 
-# Default to generating all libraries.
-if [ -n "${GENERATE_GOLDEN_ONLY+set}" ]; then
-  GENERATE_GOLDEN_ONLY=${GENERATE_GOLDEN_ONLY:-}
-else
-  GENERATE_GOLDEN_ONLY=""
-fi
-
 # --local is the most fundamental build mode, in that all other builds
 # eventually call this one. For example, a --docker build will build the
 # specified docker image, then in a container from that image it will run the
@@ -355,7 +348,7 @@ if [[ "${DOCKER_FLAG}" = "true" ]]; then
     "--volume=${HOME}/.config/gcloud:/h/.config/gcloud:Z"
     # Makes the generate-libraries build ONLY touch golden files in the
     # generator dir.
-    "--env=GENERATE_GOLDEN_ONLY=${GENERATE_GOLDEN_ONLY}"
+    "--env=GENERATE_GOLDEN_ONLY=${GENERATE_GOLDEN_ONLY-}"
   )
   # All GOOGLE_CLOUD_* env vars will be passed to the docker container.
   for e in $(env); do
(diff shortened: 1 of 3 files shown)


ci/cloudbuild/build.sh
```diff
@@ -191,6 +191,13 @@ export TRIGGER_TYPE
 export VERBOSE_FLAG
 export LOG_LINKER_PAT
 
+# Default to generating all libraries.
+if [ -n "${GENERATE_GOLDEN_ONLY+set}" ]; then
+  GENERATE_GOLDEN_ONLY=${GENERATE_GOLDEN_ONLY:-}
+else
+  GENERATE_GOLDEN_ONLY=""
+fi
```
devbww, line 199: This seems to be the equivalent of `: ${GENERATE_GOLDEN_ONLY=}`, which can be read as "if GENERATE_GOLDEN_ONLY is unset, then set it to the empty string".

But, you could just use `${GENERATE_GOLDEN_ONLY-}` below instead.

ci/cloudbuild/builds/generate-libraries.sh
```diff
@@ -32,22 +32,39 @@ bazel run --action_env=GOOGLE_CLOUD_CPP_ENABLE_CLOG=yes \
...
+fi
+
+if [ -z "${GENERATE_GOLDEN_ONLY}" ]; then
+  # TODO(#5821): The generator should run clang-format on its output files itself
+  # so we don't need this extra step.
+  io::log_h2 "Formatting generated code"
+  git ls-files -z -- '*.h' '*.cc' |
+    xargs -P "$(nproc)" -n 1 -0 clang-format -i
+else
+  io::log_red "Only formatting generator golden code."
```
devbww, line 54: s/generator/generated/

ci/cloudbuild/builds/generate-libraries.sh
```diff
@@ -32,22 +32,39 @@ bazel run --action_env=GOOGLE_CLOUD_CPP_ENABLE_CLOG=yes \
...
+    'generator/integration_tests/golden/internal/*.cc' \
+    'generator/integration_tests/golden/*.cc' |
+    xargs -P "$(nproc)" -n 1 -0 clang-format -i
+fi
+
+if [ -z "${GENERATE_GOLDEN_ONLY}" ]; then
+  io::log_h2 "Updating protobuf lists/deps"
+  external/googleapis/update_libraries.sh
+else
+  io::log_red "Skipping update of generated libraries dependencies."
```
devbww, line 66: "Skipping update of protobuf lists/deps" perhaps, to match non-skipping message.

ci/cloudbuild/cloudbuild.yaml
```diff
@@ -25,6 +25,7 @@ options:
     'COMMIT_SHA=${COMMIT_SHA}',
     'PR_NUMBER=${_PR_NUMBER}',
     'TRIGGER_TYPE=${_TRIGGER_TYPE}',
+    'GENERATE_GOLDEN_ONLY=""',
```
devbww, line 28: Is this important, given that the above code doesn't distinguish between unset and empty?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10150)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @devbww)

___
*[`ci/cloudbuild/build.sh` line 199 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10150#-NFjitGAYhefRfcY_7E_-r2-199:-NFnwjRB6C5x3phANrVS:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/044645c5d7fbe32fce486264ebe2e15291f36347/ci/cloudbuild/build.sh#L199)):*

This seems to be the equivalent of `: ${GENERATE_GOLDEN_ONLY=}`, which can be read as "if GENERATE_GOLDEN_ONLY is unset, then set it to the empty string".

But, you could just use `${GENERATE_GOLDEN_ONLY-}` below instead.
</blockquote></details>

Done.
___
*[`ci/cloudbuild/cloudbuild.yaml` line 28 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10150#-NFk48C47o5RAaXGh21w-r2-28:-NFnwuuZ549_J2uKLoYe:b-8ebus3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/044645c5d7fbe32fce486264ebe2e15291f36347/ci/cloudbuild/cloudbuild.yaml#L28)):*

Is this important, given that the above code doesn't distinguish between unset and empty?
</blockquote></details>

So the code in build.sh only passes the env var when using docker, for cloud build, this is the only way I can prevent getting the following error:
`Step #3: /workspace/ci/cloudbuild/builds/generate-libraries.sh: line 35: GENERATE_GOLDEN_ONLY: unbound variable`
___
*[`ci/cloudbuild/builds/generate-libraries.sh` line 54 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10150#-NFjitGAYhefRfcY_7Ea-r2-54:-NFnx8giAMLm6fn1f85S:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/044645c5d7fbe32fce486264ebe2e15291f36347/ci/cloudbuild/builds/generate-libraries.sh#L54)):*

s/generator/generated/
</blockquote></details>

Done.
___
*[`ci/cloudbuild/builds/generate-libraries.sh` line 66 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10150#-NFjitGAYhefRfcY_7Ea-r2-66:-NFnx9FFBhmQn8lgJkiw:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/044645c5d7fbe32fce486264ebe2e15291f36347/ci/cloudbuild/builds/generate-libraries.sh#L66)):*

"Skipping update of protobuf lists/deps" perhaps, to match non-skipping message.
</blockquote></details>

Done.

devbww (approved):

PR #10156, opened 2022-11-01
cleanup: eliminate ODR violations in samples

devbww:
Samples that have their own `main()`, and don't use gtest anyway, should not link with `@com_google_googletest//:gtest_main`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10156)


coryan (approved):
If you are interested in this, there are common libraries that link `gtest_main`, those may need fixing too.  We may need to do the same thing for CMake, I did not look.

devbww:
> We may need to do the same thing for CMake

Oops.  Done.  PTAL.

> there are common libraries that link `gtest_main`, those may need fixing too.

OK, I'll take a look separately.

coryan (approved):

PR #10172, opened 2022-11-03
ci(batch): remove suppression of batch quickstart on Windows

### LINKED ISSUE #10067: remove suppression of batch quickstart on Windows
#10056 caused an intermittent[*] problem on Windows as the non-experimental target for the Batch library is not yet available in vcpkg.

[*] Intermittent because we only build a random subset of quickstarts on Windows (see #9923).

So, for the time being we avoid building the Batch quickstart altogether (see #10068).  This issue is a reminder to reverse that.

devbww:
Fixes #10067.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10172)


dbolduc (approved):

dbolduc:
Actually, don't we need to update the vcpkg version in our repo first?

devbww:
> Actually, don't we need to update the vcpkg version in our repo first?

Right you are.

PR #10175, opened 2022-11-03
Add storage_mocks library to Google Cloud Storage

zapatoshoe:
I'm not sure why the mock client wasn't already exposed, but I'm adding it here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10175)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -265,6 +265,14 @@ cc_library(
 )
 
 cc_library(
+    name = "storage_mocks",
+    testonly = True,
+    deps = [
+        "//google/cloud/storage:storage_client_testing",
+    ],
+)
+
+cc_library(
     name = "experimental-storage-grpc",
     tags = ["manual"],
     deps = [


dbolduc:
Hey, thanks for the PR, but I don't think we can accept it.

> I'm not sure why the mock client wasn't already exposed, but I'm adding it here.

I have an answer, but I do not think you will be thrilled to hear it....

Adding a public `storage_mocks` library is something we want to do, and that work is tracked in #7142 

The tl;dr for why we can't right now is that `//google/cloud/storage:storage_client_testing` contains **both** helpers for mocking the `storage::Client` (which we "support" and want customer to use) **and** helpers for the google-cloud-cpp library maintainers to test our `internal` code (which we do not support nor want customers to use).

To make things even more messy, the mocks which we "support" in `google::cloud::storage::testing` use `internal` types.... (which we tell customers not to use 😖).

So we will not add a new `storage_mocks` until we have decided the exact mocking API we want to expose for the long run. (e.g. no `internal` types).

As for what you should do now if you want to mock the storage library? You can either depend on `//google/cloud/storage:storage_client_testing` anyway, and follow the [storage mocking guide](https://googleapis.dev/cpp/google-cloud-storage/2.3.0/storage-mocking.html) (which you might be doing already). Or you can wait for #7142 to be completed. I don't have an expected date for when the work will be completed. I doubt it will happen in 2022. It is a very difficult problem to solve, that requires rearchitecting the library.

zapatoshoe:
Thanks for the quick response. Will proceed with the former option you suggested.

Issue #10169, opened 2022-11-03
Handle client methods with identical signatures in the generator


dbolduc:
Per https://google.aip.dev/client-libraries/4232, about the `google.api.method_signature` annotation:

"An RPC can provide this annotation more than once to specify multiple signatures. Order matters here: In some situations, it may not be possible to generate an overload for every signature provided. In this situation, client library generators must follow a "first match wins" strategy (generate an overload for the first signature in the conflict set, and drop the rest)."

Here is a case in the wild: https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v2/cloud_speech.proto#L125-L133

`uri` and `content` are both strings, so we would generate the same signature.

PR #10178, opened 2022-11-04
test(generator): check for absence of method var keys

dbolduc:
Motivated by #10169 

This will be useful to check that we do not include repeated message signatures. Verify that a method key is not present in the `VarsDictionary`. Start with the two deprecated functions in the unit test.

https://github.com/googleapis/google-cloud-cpp/blob/f3d08b736565af0d5bef916c8134b52c77a579a4/generator/internal/descriptor_utils_test.cc#L455-L463

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10178)

### DIFF
--- a/generator/internal/descriptor_utils_test.cc
+++ b/generator/internal/descriptor_utils_test.cc
@@ -37,7 +37,9 @@ using ::google::protobuf::DescriptorPool;
 using ::google::protobuf::FileDescriptor;
 using ::google::protobuf::FileDescriptorProto;
 using ::google::protobuf::MethodDescriptor;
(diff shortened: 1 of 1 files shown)


devbww (approved):

dbolduc:
D'oh I just noticed this test:
https://github.com/googleapis/google-cloud-cpp/blob/40c2dd3e2d59200de70d7870aa7e2fed2ca6ed4b/generator/internal/descriptor_utils_test.cc#L595-L609

I might revert this PR for the sake of cleanliness...

PR #10180, opened 2022-11-04
test: legible output of `std::chrono` values

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10180)

### DIFF
--- a/google/cloud/internal/type_traits.h
+++ b/google/cloud/internal/type_traits.h
@@ -16,7 +16,7 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_TYPE_TRAITS_H
 
 #include "google/cloud/version.h"
-#include <ostream>
+#include <iosfwd>
 #include <type_traits>
 #include <utility>
 
@@ -40,22 +40,18 @@ namespace internal {
 template <typename...>
 using void_t = void;
 
+//@{
 /**
  * A helper to detect if a type is output streamable.
  */
-template <typename T>
-class IsOStreamable {
- private:
-  template <typename U>
-  static auto test(int)
-      -> decltype(std::declval<std::ostream&>() << std::declval<U>(),
-                  std::true_type());
-  template <typename>
-  static auto test(...) -> std::false_type;
+template <typename T, typename = void>
+class IsOStreamable : public std::false_type {};
 
- public:
-  static constexpr bool kValue = decltype(test<T>(0))::value;
-};
+template <typename T>
+class IsOStreamable<
+    T, void_t<decltype(std::declval<std::ostream&>() << std::declval<T>())>>
+    : public std::true_type {};
+//@}
 
 }  // namespace internal
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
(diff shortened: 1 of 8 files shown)


google/cloud/internal/type_traits.h
```diff
@@ -16,6 +16,9 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_TYPE_TRAITS_H
 
 #include "google/cloud/version.h"
+#include <ostream>
```
coryan, line 19: Probably `#include <iosfwd>` would do in this case.

google/cloud/testing_util/chrono_matchers.h
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Provide `std::chrono::duration` and `std::chrono::time_point` output
+ * streaming when otherwise unsupported. This enables the printing of
+ * legible chrono values from googletest, rather than raw bytes, when a
+ * match fails.
+ *
+ * It is strictly undefined behavior to add to any `std` namespace, but
+ * it works, and, in this test-only case, is well worth it.
+ */
+namespace std {
```
coryan, line 53: If you are willing to open the namespace, then adding `PrintTo()` overloads is also an option:

https://github.com/google/googletest/blob/main/docs/advanced.md#teaching-googletest-how-to-print-your-values

Unless you worry about a `PrintTo()` already in the `std::chrono()` namespace

google/cloud/internal/type_traits.h
```diff
@@ -37,6 +40,23 @@ namespace internal {
 template <typename...>
 using void_t = void;
 
+/**
```
coryan, line 43: Alternatively:

```cc
template <typename T, typename AlwaysVoid = void>
class IsOStreamable : public std::false_type {};

template <typename T>
class IsOStreamable<T, void_t<decltype(std::declval<std::ostream&>() << std::declval<U>())>> : public std::true_type {};
```

https://godbolt.org/z/oPqW6e4nq

coryan (approved):
LGTM, and this seems easier to adopt through all tests than my proposal.
A couple of (purely optional) comments.

google/cloud/internal/type_traits.h
devbww: Done.

google/cloud/testing_util/chrono_matchers.h
devbww: The win with the existing incantation is that normal, outside-googletest streaming also works.  I admit that that whole thing is currently wrapped in "chrono_matcher" clothing, but having it also for non-googletest tests is nice.  We could change the filename.

google/cloud/internal/type_traits.h
devbww: Done.

google/cloud/testing_util/chrono_matchers.h
coryan: Ack.  Your call I think.

coryan (approved):

google/cloud/testing_util/chrono_matchers.h
devbww: Renamed to "chrono_output".

PR #10182, opened 2022-11-04
impl(generator): handle conflicting method overloads

### LINKED ISSUE #10169: Handle client methods with identical signatures in the generator

dbolduc:
Fixes #10169

### DIFF
--- a/generator/internal/service_code_generator.h
+++ b/generator/internal/service_code_generator.h
@@ -171,9 +171,8 @@ class ServiceCodeGenerator : public GeneratorInterface {
    * the number of method signatures to emit may be fewer than the number in
    * the proto file.
    */
-  bool IsDeprecatedMethodSignature(
-      google::protobuf::MethodDescriptor const& method,
-      int method_signature_number) const;
+  bool OmitMethodSignature(google::protobuf::MethodDescriptor const& method,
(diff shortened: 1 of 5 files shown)


coryan (approved):

generator/internal/service_code_generator.h
```diff
@@ -171,9 +171,8 @@ class ServiceCodeGenerator : public GeneratorInterface {
    * the number of method signatures to emit may be fewer than the number in
```
devbww, line 171: Perhaps this comment needs revisiting now?

dbolduc: Fixed.

PR #10186, opened 2022-11-07
impl(pubsub): an `AckHandler` for blocking pull

coryan:
We want to offer a similar API to `nack()` and `ack()` messages when received via a blocking pull vs. a callback.  This `AckHandler` (once wrapped) will be able to provide the same API customers use for streaming calls.  It will also maintain the message lease in the background, as we do for the callback API.

Part of the work for #7187

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10186)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10186)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

Issue #9340, opened 2022-06-23
Generate library for `beyondcorp.googleapis.com`


coryan:
This should live in `google/cloud/beyondcorp/`.  There are multiple services, all in:

https://github.com/googleapis/googleapis/tree/master/google/cloud/beyondcorp

coryan:
The only bit remaining is supporting Bazel on Windows, which is blocked on external bugs (`cl.exe` does not support long paths, or Bazel generates too long of a path, or the proto names are too long, take your pick).

coryan:
I filed something with the Bazel team: https://github.com/bazelbuild/bazel/issues/15961

PR #10189, opened 2022-11-07
cleanup: break crc32c dependency on workspace name

coryan:
With this change crc32c works regardless of the name given to `google-cloud-cpp` workspace, even when imported.

Part of the changes for #9340

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10189)


bazel/configure_template.bzl.patch
```diff
@@ -0,0 +1,42 @@
...
++# distributed under the License is distributed on an "AS IS" BASIS,
++# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
++# See the License for the specific language governing permissions and
++# limitations under the License.
++
++"""
++Define a rule to configure .in template files.
++
++We use custom *.BUILD files for two external dependencies that do not natively support Bazel: crc32c and libcurl.
++Both require configuring a `.in` file. This creates a rules to perform this configuration.
```
devbww, line 22: It doesn't look we use this for libcurl.  Should we refrain from mentioning it?

devbww (approved):

PR #10197, opened 2022-11-08
doc(bigtable): create page for configuration options

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10197)

### DIFF
--- a/google/cloud/bigtable/options.h
+++ b/google/cloud/bigtable/options.h
@@ -69,6 +69,8 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
  *
  * @see https://cloud.google.com/bigtable/docs/replication-overview#app-profiles
  *     for how app profiles are used to achieve replication.
+ *
+ * @ingroup bigtable-options
  */
 struct AppProfileIdOption {
   using Type = std::string;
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/options.h
```diff
@@ -145,17 +151,29 @@ using DataLimitedErrorCountRetryPolicy =
     ::google::cloud::internal::LimitedErrorCountRetryPolicy<
         bigtable::internal::SafeGrpcRetry>;
 
-/// Option to configure the retry policy used by `Table`.
+/**
+ * Option to configure the retry policy used by `Table`.
+ *
+ * @ingroup bigtable-options
+ */
 struct DataRetryPolicyOption {
```
dbolduc, line 159: every generated service has an equivalent of this option too. Will they all get these `*-options.dox`?

dbolduc (commented):
Should we be adding the `OptionList<>`s? Have they fallen out of favor?

coryan:
> Should we be adding the `OptionList<>`s? Have they fallen out of favor?

I am not sure why customers would care for them.  AFAIK they are only used to implement some checks, and probably should not be in the public namespace.  I have no idea if we should remove them, move them to `*_internal`.  Until we decide, we should leave them undocumented (created #10206) to track.

google/cloud/bigtable/options.h
coryan: I would like to, yes, but I did the hand-crafted libraries first.

dbolduc (approved):
Ack. I think the `OptionList<>`s were also added for discoverability of Options. https://github.com/googleapis/google-cloud-cpp/blob/3baceacc1fde7f36d478df153cc113b857798335/google/cloud/accessapproval/access_approval_connection.h#L118-L125

But these doxygen ref groups probably do a better job.

PR #10207, opened 2022-11-08
cleanup: shorten Bazel project name

### LINKED ISSUE #9340: Generate library for `beyondcorp.googleapis.com`

coryan:
Using a shorter project name will avoid problems on Windows, where Bazel
can create paths exceeding the MSVC hardcoded limits (260 characters).
Note that, AFAICT, this is a bug in the compiler. Windows supports
longer paths, but the compiler cannot use them for some reason. Bazel
does not help either, by creating deeply nested names for the object
files.

Fixes  #9340

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10207)

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-workspace(name = "com_github_googleapis_google_cloud_cpp")
+workspace(name = "google_cloud_cpp")
 
 load("//bazel:google_cloud_cpp_deps.bzl", "google_cloud_cpp_deps")
 
--- a/ci/kokoro/windows/builds/quickstart-bazel.ps1
+++ b/ci/kokoro/windows/builds/quickstart-bazel.ps1
@@ -40,8 +40,8 @@ function Get-Released-Quickstarts {
 
     Push-Location "${project_root}/google/cloud/bigtable/quickstart"
     bazelisk $bazel_common_flags version | Out-Null
-    bazelisk $bazel_common_flags query --noshow_progress --noshow_loading_progress "filter(/quickstart:quickstart$, kind(cc_binary, @com_github_googleapis_google_cloud_cpp//google/...))" |
-        ForEach-Object { $_.replace("@com_github_googleapis_google_cloud_cpp//google/cloud/", "").replace("/quickstart:quickstart", "") } |
+    bazelisk $bazel_common_flags query --noshow_progress --noshow_loading_progress "filter(/quickstart:quickstart$, kind(cc_binary, @google_cloud_cpp//google/...))" |
+        ForEach-Object { $_.replace("@google_cloud_cpp//google/cloud/", "").replace("/quickstart:quickstart", "") } |
         # The following quickstarts have problems building on Windows:
         #   TODO(#8145) - asset (TRUE/FALSE macros)
         #   TODO(#10067) - batch (awaiting non-experimental target)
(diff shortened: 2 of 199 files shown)


WORKSPACE.bazel
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-workspace(name = "com_github_googleapis_google_cloud_cpp")
+workspace(name = "google_cloud_cpp")
```
dbolduc, line 15: I think this is a breaking change.

Customers that depend on `@com_github_googleapis_google_cloud_cpp//...` have to update their BUILD files.

is there some alias we can leave?

coryan: I do not think this is breaking.  Customers can continue to use something like this:

```
http_archive(
    name = "com_github_googleapis_google_cloud_cpp",
    sha256 = "245e198e29c4ec19734cc99ef631daaefbdb874307fc3743e22514ee8bcb36c4",
    strip_prefix = "google-cloud-cpp-2.4.0",
    url = "https://github.com/googleapis/google-cloud-cpp/archive/v2.4.0.tar.gz",
)
```

And use the `@com_github_googleapis_google_cloud_cpp//...` in their build scripts (or Bazel files).  It used to be impossible to use any other name because we hard-coded `@com_github_googleapis_google_cloud_cpp` in our own BUILD files, but I have managed to remove that restriction (using `Label()` when possible, changing some stuff to be "patches" instead of references).

No, we cannot leave an alias.  There is a long discussion in https://github.com/bazelbuild/bazel/issues/3219 explaining what the Bazel team thinks is a solution, and everybody else telling them it is not.

dbolduc (approved):
FYI: I think this fixes the issue linked in the PR description

coryan:
> FYI: I think this fixes the issue linked in the PR description

I was missing one change in CI, fixed, and will update the description.

WORKSPACE.bazel
dbolduc: Please check my logic here. I could still be misunderstanding....

Considering this example: https://github.com/googleapis/google-cloud-cpp/blob/8a8ba10eb3ee764f8ca70fd776f5d7dbb9acf1c9/bazel/google_cloud_cpp_deps.bzl#L50-L59

Let's say abseil changes their project name to `com_google_abseil_new`. All of our code keeps working. No problem.

But if we change our `load_deps()` to check for `"com_google_abseil_new"`, it would be a breaking change. (e.g. if a downstream app is pinning abseil to a specific version, we would ignore it in favor of our default version of abseil.).

So are we to blame for the break? or is it really abseil....

coryan: > But if we change our `load_deps()` to check for `"com_google_abseil_new"`

Why would we do that?  We are trying to find out if something else has loaded Abseil **with the name we intend to use**.  The name that Abseil calls itself is not material here.

A different question would be:

- What if a downstream application starts loading Abseil with the `com_google_abseil_new` name?

They could do that. And that may break things if they loaded a different version (ODR and diamond dependencies get messy quickly).  In the Bazel world it is the job of the leaf WORKSPACE to satisfy all the dependencies consistently.  That is cumbersome, and thus the invention of `bzlmod`.

For example, if gRPC starts using a different name for Abseil then we need to use the same name, or introduce some kind of repository mapping (and/or `bind()`) to make sure gRPC uses the same version we loaded.

dbolduc: > In the Bazel world it is the job of the leaf WORKSPACE to satisfy all the dependencies consistently.

Ah. I did not realize this was the paradigm. That clears things up. Thanks.

PR #10220, opened 2022-11-09
ci: split fedora dockerfiles by build system

### LINKED ISSUE #10217: Consider using a dedicated Dockerfile for bazel builds
Our bazel builds (all except for `msan`) use `fedora-36.Dockerfile`. @coryan points out that there is an opportunity to speed up the builds by instead using a dedicated `fedora-36-bazel.Dockerfile`, which installs the minimal build tools needed for bazel. (and thus does not build our dependencies from source with cmake).

`fedora-msan.Dockerfile` is a good example of such a minimal Dockerfile.

dbolduc:
Fixes #10217 

This separates the Fedora 36 Dockerfile into bazel and cmake components. It will speed up bazel builds where we `--docker-clean` a bit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10220)

### DIFF
--- a/ci/cloudbuild/dockerfiles/fedora-36-cmake.Dockerfile
+++ b/ci/cloudbuild/dockerfiles/fedora-36-cmake.Dockerfile
@@ -179,16 +179,8 @@ ENV CLOUDSDK_PYTHON=python3.10
 RUN /var/tmp/ci/install-cloud-sdk.sh
 ENV CLOUD_SDK_LOCATION=/usr/local/google-cloud-sdk
 ENV PATH=${CLOUD_SDK_LOCATION}/bin:${PATH}
-# The Cloud Pub/Sub emulator needs Java, and so does `bazel coverage` :shrug:
-# Bazel needs the '-devel' version with javac.
-RUN dnf makecache && dnf install -y java-latest-openjdk-devel
-
-RUN curl -o /usr/bin/bazelisk -sSL "https://github.com/bazelbuild/bazelisk/releases/download/v1.15.0/bazelisk-linux-${ARCH}" && \
-    chmod +x /usr/bin/bazelisk && \
-    ln -s /usr/bin/bazelisk /usr/bin/bazel
-
-RUN dnf makecache && dnf install -y "dnf-command(debuginfo-install)"
-RUN dnf makecache && dnf debuginfo-install -y libstdc++
+# The Cloud Pub/Sub emulator needs Java :shrug:
+RUN dnf makecache && dnf install -y java-latest-openjdk
 
 # Some of the above libraries may have installed in /usr/local, so make sure
 # those library directories will be found.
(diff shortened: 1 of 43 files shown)


coryan (approved):

devbww (commented):
Some questions:

Will `ci/cloudbuild/dockerfiles/fedora-36.Dockerfile` go away (after updating the "integration-daily" and "rotate-keys" manual triggers)?  Aside: distro fedora-36 is also referenced in comments in `build.sh` and `builds/integration-daily.sh`.

Can `fedora-36-cxx14.Dockerfile` or `fedora-36-cxx20.Dockerfile` benefit from any similar trimmage?

ci/cloudbuild/dockerfiles/fedora-36-cmake.Dockerfile
```diff
@@ -0,0 +1,187 @@
...
+    ./configure --prefix=/usr/local && \
+    make && \
+    make install && \
+    cd /var/tmp && rm -fr build
+
+# Installs the abi-dumper with the integer overflow fix from
+# https://github.com/lvc/abi-dumper/pull/29. We can switch back to `dnf install
+# abi-dumper` once it has the fix.
+WORKDIR /var/tmp/build
+RUN curl -sSL https://github.com/lvc/abi-dumper/archive/814effec0f20a9613441dfa033aa0a0bc2a96a87.tar.gz | \
```
coryan, line 169: You need to update this SHA now.

coryan (changes requested):

ci/cloudbuild/dockerfiles/fedora-36-cmake.Dockerfile
dbolduc: Fixed. Note that the SHA was updated for `fedora-36-cmake.Dockerfile` in the first commit, during the rebase. (The SHA in `fedora-36.Dockerfile` was updated manually).

I also restored bazelisk in `fedora-36-cmake.Dockerfile`. I should not have removed it.

dbolduc:
> Will `ci/cloudbuild/dockerfiles/fedora-36.Dockerfile` go away (after updating the "integration-daily" and "rotate-keys" manual triggers)? Aside: distro fedora-36 is also referenced in comments in `build.sh` and `builds/integration-daily.sh`.

yes. Fixed the comments.

> Can `fedora-36-cxx14.Dockerfile` or `fedora-36-cxx20.Dockerfile` benefit from any similar trimmage?

Good catch. It looks like they (and `fedora-36-cmake.Dockerfile`) could use trimming. (I see at least a few `lib*san`s). I am going to wait until after the triggers are updated on GCB to trim these, so that the builds will actually fail, if I trim too much.

devbww (approved):

PR #10222, opened 2022-11-10
fix(pubsublite): wait for reads during PartitionPublisher shutdown

tmdiep:
Fixes a race during destruction of PartitionPublisher that can cause crashes. It now blocks until the result of the last Read is handled.

Internal: b/254929621#comment4

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10222)


google/cloud/pubsublite/internal/partition_publisher.h
```diff
@@ -96,9 +100,9 @@ class PartitionPublisher
   std::deque<std::deque<MessageWithPromise>>
       unsent_batches_;  // ABSL_GUARDED_BY(mu_)
   std::deque<std::deque<MessageWithPromise>>
-
       in_flight_batches_;  // ABSL_GUARDED_BY(mu_)
   bool writing_ = false;   // ABSL_GUARDED_BY(mu_)
+  absl::optional<promise<void>> reading_;  // ABSL_GUARDED_BY(mu_)
```
coryan, line 105: The robots would like you to reformat this code:

```
diff --git a/google/cloud/pubsublite/internal/partition_publisher.h b/google/cloud/pubsublite/internal/partition_publisher.h
index a750951..961caf0 100644
--- a/google/cloud/pubsublite/internal/partition_publisher.h
+++ b/google/cloud/pubsublite/internal/partition_publisher.h
@@ -100,8 +100,8 @@ class PartitionPublisher
   std::deque<std::deque<MessageWithPromise>>
       unsent_batches_;  // ABSL_GUARDED_BY(mu_)
   std::deque<std::deque<MessageWithPromise>>
-      in_flight_batches_;  // ABSL_GUARDED_BY(mu_)
-  bool writing_ = false;   // ABSL_GUARDED_BY(mu_)
+      in_flight_batches_;                  // ABSL_GUARDED_BY(mu_)
+  bool writing_ = false;                   // ABSL_GUARDED_BY(mu_)
   absl::optional<promise<void>> reading_;  // ABSL_GUARDED_BY(mu_)
 
   std::unique_ptr<AlarmRegistry::CancelToken> cancel_token_;
```

* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/10222/3fa3d30ab2912de7f5cd8ef7a65400013805f429/checkers-checkers/log-62d838c2-28df-4bed-8c67-fdd20038f626.txt

google/cloud/pubsublite/internal/partition_publisher.cc
```diff
@@ -170,6 +182,24 @@ void PartitionPublisher::Read() {
       });
 }
 
+void PartitionPublisher::OnReadStart() {
+  std::lock_guard<std::mutex> g{mu_};
+  if (!reading_) {
```
coryan, line 187: nit: we would probably write

```cc
if (!reading_) reading_.emplace();
```

google/cloud/pubsublite/internal/partition_publisher.cc
```diff
@@ -170,6 +182,24 @@ void PartitionPublisher::Read() {
...
+  }
+}
+
+void PartitionPublisher::OnReadEnd() {
+  absl::optional<promise<void>> read_done;
+  {
+    std::lock_guard<std::mutex> g{mu_};
+    read_done.swap(reading_);
+  }
+  if (read_done) {
```
coryan, line 198: Same nit, we tend to write `if (read_done) read_done->set_value();`

google/cloud/pubsublite/internal/partition_publisher_test.cc
```diff
@@ -591,6 +592,45 @@ TEST_F(PartitionPublisherTest, PublishAfterShutdown) {
...
+  EXPECT_CALL(alarm_token_ref_, Destroy);
+  EXPECT_CALL(resumable_stream_ref_, Shutdown)
+      .WillOnce(Return(ByMove(make_ready_future())));
+  auto shutdown_future = publisher_->Shutdown();
+  start_promise.set_value(Status());
+  EXPECT_EQ(publisher_start_future.get(), Status());
+
+  // shutdown future does not finish until read response is received.
+  EXPECT_EQ(shutdown_future.wait_for(std::chrono::milliseconds(25)),
+            std::future_status::timeout);
```
coryan, line 629: Hmm, any reason that `wait_for()` needs to be for 25ms?  Wouldn't any amount of time give you the same confidence?  If so, consider something shorter, like 2ms.

google/cloud/pubsublite/internal/partition_publisher.cc
```diff
@@ -94,8 +95,14 @@ void PartitionPublisher::Flush() {
...
-  return service_composite_.Shutdown().then(
-      [this](future<void>) { SatisfyOutstandingMessages(); });
+  future<void> shutdown = service_composite_.Shutdown();
+  {
+    std::lock_guard<std::mutex> g{mu_};
+    if (reading_) {
+      shutdown = shutdown.then(ChainFuture(reading_->get_future()));
+    }
+  }
+  return shutdown.then([this](future<void>) { SatisfyOutstandingMessages(); });
```
coryan, line 105: Why is it Okay to hold on to `this` in that continuation?  What guarantees the lifetime of the object is long enough?  Do you mind adding a comment explaining it?

coryan (changes requested):
Thanks for the PR!  Generally it looks good.  Some formatting things that you will need to fix, the robots are pretty stubborn (I left a comment).
Some small nits to help us grok the code.

google/cloud/pubsublite/internal/partition_publisher.cc
dpcollins-google: This could use a comment, but its part of the interface of Service that you wait on the shutdown future before destroying the object. There's nothing enforcing this in the type system however.

google/cloud/pubsublite/internal/partition_publisher.cc
tmdiep: Done

google/cloud/pubsublite/internal/partition_publisher.cc
tmdiep: Done

google/cloud/pubsublite/internal/partition_publisher_test.cc
tmdiep: I changed this to `EXPECT_FALSE(shutdown_future.is_ready());`

The wait was from an earlier iteration where there was a false positive test pass for no implementation changes. But this has now been resolved.

coryan (approved):
Thanks again for the PR.

PR #10228, opened 2022-11-11
feat(speech): generate speech v2

### LINKED ISSUE #10168: Generate library for speech v2
https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v2

dbolduc:
Fixes #10168 

Note that moving speech v1 is not a prerequisite for making this change. The types will not conflict, and we know where we want to put speech v2.

I tried to come up with a general glob solution. It deserves scrutiny.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10228)


google/cloud/speech/CMakeLists.txt
```diff
@@ -50,10 +50,24 @@ google_cloud_cpp_grpcpp_library(
 external_googleapis_set_version_and_alias(speech_protos)
 target_link_libraries(google_cloud_cpp_speech_protos PUBLIC ${proto_deps})
 
+unset(mocks_globs)
+unset(source_globs)
+set(service_dirs ";v2/")
```
coryan, line 55: if you are planning to use something like a list, then do so:

```
set(service_dirs "" "v2/")
set(internal_dirs "" "internal/")
set(extensions "*.h" ".cc")
```

But all of this seems overly complicated, I think something like

```
file(GLOB source_files RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}"
       "*.h" "*.cc" "internal/*.h" "intermal/*.cc"
       "v2/*.h" "v2/*.cc" "v2/internal/*.h" "v2/intermal/*.cc")
```

Is easier to read.

google/cloud/speech/v2/samples/speech_client_samples.cc
```diff
@@ -0,0 +1,90 @@
+// Copyright 2022 Google LLC
```
coryan, line 1: Normally the scaffold would generate a CMakeLists.txt and BUILD file in this directory.  You need to manually create them:

https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/speech/samples/BUILD.bazel

google/cloud/speech/v2/speech_options.h
```diff
@@ -0,0 +1,86 @@
...
+
+namespace google {
+namespace cloud {
+namespace speech_v2 {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * Use with `google::cloud::Options` to configure the retry policy.
+ *
+ * @ingroup google-cloud-speech-v2-options
```
coryan, line 37: This is a new Doxygen group, you need to create a page to host these.

coryan (changes requested):
I see no changes the `main.dox` which would include the samples for this new client.  I think you will need to make some changes to the `ci/generate-markdown` scripts and maybe add new placeholders in `main.dox`.

Likewise, you need a new group for the Options documentation.

Finally, the samples are not getting compiled.  You need to add a BUILD and CMakeLists.txt file (the scaffold steps would have done that before).  **And** you need to modify the top-level CMake file around here:

https://github.com/googleapis/google-cloud-cpp/blob/1fbb8ebd02018d441bad97563e3711772a239cfc/CMakeLists.txt#L358-L365

google/cloud/speech/CMakeLists.txt
dbolduc: Ack, I unrolled the `internal_dirs` and `extensions`, but I left the `service_dirs` loop. I think it offers a general solution that we can use in other libraries. e.g. bigquery might (eventually) look like:

```
set(service_dirs "analyticshub/v1/" "connection/v1/"
                 "datatransfer/v1/" "migration/v2/"
                 "reservation/v1/" "storage/v1/" "v2/")
```

google/cloud/speech/v2/samples/speech_client_samples.cc
dbolduc: I opted to build the samples from the library's top level BUILD and CMakeLists.txt. The top level build files will need to know where the files are anyway. And we don't need to play the `add_subdirectory` game.

dbolduc:
> I see no changes the `main.dox` which would include the samples for this new client. I think you will need to make some changes to the `ci/generate-markdown` scripts and maybe add new placeholders in `main.dox`.

Fixed by #10241 + regeneration

google/cloud/speech/doc/main.dox
```diff
@@ -42,6 +42,10 @@ which should give you a taste of the Cloud Speech-to-Text API C++ client library
 
 <!-- inject-endpoint-env-vars-start -->
 
+- `GOOGLE_CLOUD_CPP_SPEECH_ENDPOINT=...` overrides the
```
coryan, line 45: Seems like some sort of `sort -u` would be useful in the script that updates these things?

coryan (approved):

google/cloud/speech/doc/main.dox
dbolduc: I lean more towards including the namespace, such as: `speech_v2::MakeSpeechConnection()`

PR #10235, opened 2022-11-11
ci: make the `check-api` build functional

### LINKED ISSUE #8484: The `check-api` build is missing API changes
What is says in the title.  Removing a function from `servicemanagement` went undetected.

coryan:
We were missing changes to support Fedora:34 and higher. I also made
some changes to make the script easier to read. During troubleshooting,
I added a number of flags that seem generally useful: I limited the
number of headers that are part of each library's "public" headers, I
added flags to find headers and other libraries, and I change the script
to also dump "unused" symbols.

I also changed the ABI dumps to include the git revision. I think this
makes the reports more readable, as they used to say "comparing foo
with foo".

I also installed the `libstdc++` debug symbols, it seems the
`abi-dumper` script may use them for some symbols.

Fixes #8484

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10235)


ci/cloudbuild/dockerfiles/fedora-36.Dockerfile
```diff
@@ -166,7 +166,7 @@ RUN curl -sSL https://github.com/universal-ctags/ctags/archive/refs/tags/p5.9.20
 # https://github.com/lvc/abi-dumper/pull/29. We can switch back to `dnf install
 # abi-dumper` once it has the fix.
 WORKDIR /var/tmp/build
-RUN curl -sSL https://github.com/lvc/abi-dumper/archive/814effec0f20a9613441dfa033aa0a0bc2a96a87.tar.gz | \
+RUN curl -sSL https://github.com/lvc/abi-dumper/archive/16bb467cd7d343dd3a16782b151b56cf15509594.tar.gz | \
```
devbww, line 169: That SHA also appears in `fedora-36-cxx20.Dockerfile`.

ci/cloudbuild/builds/check-api.sh
```diff
@@ -51,11 +51,37 @@ cmake --install cmake-out
 function dump_abi() {
   local library="$1"
   local prefix="$2"
+  local public_headers="${prefix}/include/google/cloud/${library#google_cloud_cpp_}"
+  if [[ "${library}" == "google_cloud_cpp_common" ]] || [[ "${library}" == "google_cloud_cpp_grpc_utils" ]]; then
```
devbww, line 55: It is simpler to use expression disjunction rather than command disjunction.  That is, ...
    ``[[ exp1 || exp2 ]]``
instead of ...
    ``[[ exp1 ]] || [[ exp2 ]]``

coryan: Done.

ci/cloudbuild/dockerfiles/fedora-36.Dockerfile
coryan: Removed from there, it was not used.

devbww (approved):

PR #10240, opened 2022-11-14
impl(rest): add user facing RestOptions

scotthart:
This adds a user visible set of RestOptions for use with REST transport. These are based on the system parameters found at https://cloud.google.com/apis/docs/system-parameters. If we need to add additional options from these parameters, we can do so later as they are needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10240)

### DIFF
--- a/google/cloud/rest_options.h
+++ b/google/cloud/rest_options.h
@@ -1,0 +1,70 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_REST_OPTIONS_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_REST_OPTIONS_H
+
+#include "google/cloud/options.h"
+#include "google/cloud/version.h"
+#include <chrono>
+#include <cstdint>
+#include <memory>
+#include <string>
+
+namespace google {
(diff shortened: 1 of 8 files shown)


google/cloud/rest_options.h
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Configure the QuotaUser system parameter.
+ *
+ * A pseudo user identifier for charging per-user quotas. If not specified, the
+ * authenticated principal is used. If there is no authenticated principal, the
+ * client IP address will be used. When specified, a valid API key with service
+ * restrictions must be used to identify the quota project. Otherwise, this
+ * parameter is ignored.
+ */
+struct QuotaUser {
```
coryan, line 38: There are similar constructs for GCS, e.g.:

https://github.com/googleapis/google-cloud-cpp/blob/e83dc4dc9b2a307723abba6b60f01fe841234a0a/google/cloud/storage/well_known_parameters.h#L524-L527

Which means there would be two different ways to set this on a storage request, e.g.:

```cc
namespace gcs = ::google::cloud::storage;
namespace g = ::google::cloud;
auto F(gcs::Client client) {
  return client.GetObjectMetadata(
      "my-bucket", "my-object", gcs::QuotaUser("user1"), g::Options{}.set<g::QuotaUser>("user2"));
```

How are we planning to handle this?  There are many reasonable questions (e.g. "deprecate the gcs version over time", or "ignore the g::QuotaUser in GCS and document this").  Leaving it undefined is not really a good plan methinks?

coryan (approved):

google/cloud/rest_options.h
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Configure the QuotaUser system parameter.
+ *
+ * A pseudo user identifier for charging per-user quotas. If not specified, the
+ * authenticated principal is used. If there is no authenticated principal, the
+ * client IP address will be used. When specified, a valid API key with service
+ * restrictions must be used to identify the quota project. Otherwise, this
+ * parameter is ignored.
+ */
```
coryan, line 37: I just realized these are missing something like:

```cc
/**
  *....
  *
  * @ingroup options
  */
```

So they will automatically appear in:

https://googleapis.dev/cpp/google-cloud-common/HEAD/group__options.html

dbolduc:
Do we need to expose a `//:rest` library for bazel?

Issue #10170, opened 2022-11-03
Support clients for multiple versions of the same GCP service


dbolduc:
See https://github.com/googleapis/google-cloud-cpp/blob/main/doc/adr/2022-11-11-multiple-versions-of-GCP-service-in-one-library.md

We need to:

- [x] Update the markdown generator to pick up endpoints from `google/cloud/*_options_defaults.cc` 
- [x] Update the markdown generator to generate samples for the new paths
  -  we will need to rename the refs to include version, because they will conflict otherwise
- [x] Add a flag to the generator to generate the following in a library's top-level headers:
  -  `#include "google/cloud/foo/*.h"` headers
  -  `using ::google::cloud::foo_vN;` directives
- [x] Update purely generated libraries' `product_path`, and enable the flag.
- [x] Update scaffolding to follow the lead of `speech/v2` w/r/t `service_dirs` in [bazel](https://github.com/googleapis/google-cloud-cpp/blob/4482ada4a0027cdfcb2a5695acaba2bb3ee4d9d9/google/cloud/speech/BUILD.bazel#L19-L22) and [cmake](https://github.com/googleapis/google-cloud-cpp/blob/4482ada4a0027cdfcb2a5695acaba2bb3ee4d9d9/google/cloud/speech/CMakeLists.txt#L27)
- [x] Update quickstarts to use versioned clients
- [ ] Update samples to use versioned clients (IAM, BQ)

See this branch for what some of the changes should look like. (Next to add them to the generator, and break down the work into logical PRs) https://github.com/googleapis/google-cloud-cpp/compare/main...dbolduc:google-cloud-cpp:speech-reorg-dev?expand=1

devbww:
See #10278 (vmwareengine/v1) for another example of the proposed layout (without backwards-compatibility headers).

dbolduc:
This is the script I used to generate #10857 

```sh
#!/bin/bash
#
# Usage: ./relocate.sh apigateway kms osconfig
#
# If anything goes wrong, it will likely be in the "update builds" step.
#
# Verify the result with:
# ```
# ci/cloudbuild/build.sh -t cmake-install-pr
# bazel build //google/cloud/...
# ```

libraries=("$@")

function edit_config() {
  library=$1
  service_dirs=("\"\"") # Remember these. They are needed in Step 5.
[...]

PR #10241, opened 2022-11-14
doc: use qualified client name in samples

dbolduc:
Part of the work for #10170 

We will need this for speech, which has conflicting client names.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10241)


generator/internal/sample_generator.cc
```diff
@@ -48,7 +48,7 @@ Status SampleGenerator::GenerateHeader() {
   HeaderSystemIncludes({"iostream", "fstream", "string", "vector"});
 
   HeaderPrint(R"""(
-// main-dox-marker: $client_class_name$
+// main-dox-marker: $product_namespace$::$client_class_name$
```
coryan, line 51: TIL

PR #10244, opened 2022-11-14
cleanup: use C++14 for clang-tidy

dbolduc:
To be honest, I am not sure what this does.

<hr />

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10244)

### DIFF
--- a/ci/cloudbuild/builds/clang-tidy.sh
+++ b/ci/cloudbuild/builds/clang-tidy.sh
@@ -26,11 +26,11 @@ export CTCACHE_DIR=~/.cache/ctcache
 mapfile -t cmake_args < <(cmake::common_args)
 
 # See https://github.com/matus-chochlik/ctcache for docs about the clang-tidy-cache
-# Note: we use C++11 for this build because we don't want tidy suggestions that
+# Note: we use C++14 for this build because we don't want tidy suggestions that
 # require a newer C++ standard.
 cmake "${cmake_args[@]}" \
   -DCMAKE_CXX_CLANG_TIDY=/usr/local/bin/clang-tidy-wrapper \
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/builds/clang-tidy.sh
```diff
@@ -26,11 +26,11 @@ export CTCACHE_DIR=~/.cache/ctcache
 mapfile -t cmake_args < <(cmake::common_args)
 
 # See https://github.com/matus-chochlik/ctcache for docs about the clang-tidy-cache
-# Note: we use C++11 for this build because we don't want tidy suggestions that
+# Note: we use C++14 for this build because we don't want tidy suggestions that
 # require a newer C++ standard.
 cmake "${cmake_args[@]}" \
   -DCMAKE_CXX_CLANG_TIDY=/usr/local/bin/clang-tidy-wrapper \
-  -DCMAKE_CXX_STANDARD=11 \
+  -DCMAKE_CXX_STANDARD=14 \
```
coryan, line 33: This is probably a good idea, or at least a good cleanup.  It may have no effect, because we already require C++14 or higher, regardless of the value in CMAKE_CXX_STANDARD:

https://github.com/googleapis/google-cloud-cpp/blob/2a871404314c0c63e1ed57f49fbc50e00be82a51/cmake/EnableWerror.cmake#L43-L46

coryan (approved):

ci/cloudbuild/builds/clang-tidy.sh
dbolduc: I was surprised when the change had no effect. But your explanation makes sense.

PR #10253, opened 2022-11-15
impl(rest): change ServerTimeout type to match type of literals

scotthart:
`Options{}.set<ServerTimeoutOption>(3.14)` creates a double not a float, which in turn results in a narrowing conversion warning. Change the option type to double as a quality of life improvement.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10253)

### DIFF
--- a/google/cloud/rest_options.h
+++ b/google/cloud/rest_options.h
@@ -60,7 +60,7 @@ struct UserIpOption {
  * @ingroup rest-options
  */
 struct ServerTimeoutOption {
-  using Type = float;
+  using Type = double;
 };
 
 /// The complete list of options accepted by `CurlRestClient`
--- a/google/cloud/rest_options.h
+++ b/google/cloud/rest_options.h
@@ -18,8 +18,6 @@
 #include "google/cloud/options.h"
 #include "google/cloud/version.h"
 #include <chrono>
-#include <cstdint>
-#include <memory>
 #include <string>
 
 namespace google {
@@ -54,13 +52,13 @@ struct UserIpOption {
 };
 
 /**
- * Timeout (in seconds) for the server to finish processing the request. This
- * system param only applies to REST APIs for which client-side timeout is not
- * applicable.
+ * Timeout for the server to finish processing the request. This system param
+ * only applies to REST APIs for which client-side timeout is not applicable.
+ *
  * @ingroup rest-options
  */
 struct ServerTimeoutOption {
-  using Type = double;
+  using Type = std::chrono::milliseconds;
 };
 
 /// The complete list of options accepted by `CurlRestClient`


google/cloud/rest_options.h
```diff
@@ -56,7 +56,7 @@ struct UserIpOption {
  * applicable.
  */
 struct ServerTimeoutOption {
-  using Type = float;
+  using Type = double;
```
coryan, line 59: I just realized what this was.  Why not `using Type = std::chrono::milliseconds` or some other type that is used to denote durations?  We can convert that to the right units when sending it over HTTP.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10253)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/rest_options.h` line 59 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10253#-NGxHfv3B7EK78NZaIZw-r1-59:-NGxTdh49dLXht65Fftv:b-hi7h2r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dee896b7a54e03f66cd1e33af9f401d1024fc421/google/cloud/rest_options.h#L59)):*

I just realized what this was.  Why not `using Type = std::chrono::milliseconds` or some other type that is used to denote durations?  We can convert that to the right units when sending it over HTTP.
</blockquote></details>

I was just going off of https://cloud.google.com/apis/docs/system-parameters which says about ServerTimeout:
```
Timeout (in seconds, float value) for the server to finish processing the request. This system param only applies to REST APIs for which client-side timeout is not applicable.
```

coryan:
*[`google/cloud/rest_options.h` line 59 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10253#-NGxHfv3B7EK78NZaIZw-r1-59:-NGxUwJg7iQCfnXIl2fa:btbkkah) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dee896b7a54e03f66cd1e33af9f401d1024fc421/google/cloud/rest_options.h#L59)):*

I was just going off of https://cloud.google.com/apis/docs/system-parameters which says about ServerTimeout:
```
Timeout (in seconds, float value) for the server to finish processing the request. This system param only applies to REST APIs for which client-side timeout is not applicable.
```
</blockquote></details>

Yes.  That is the best they can do over HTTP/gRPC.  But the C++ representation does not need to be a floating point?  And representing durations as integers or floats is a "Known Problem" [tm] the units are implicit, implicit conversions can result in bad values, etc.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10253)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/rest_options.h` line 59 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10253#-NGxHfv3B7EK78NZaIZw-r1-59:-NGxZWvzB9gQoJ8VU0Ai:bhsp6ko) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dee896b7a54e03f66cd1e33af9f401d1024fc421/google/cloud/rest_options.h#L59)):*

Yes.  That is the best they can do over HTTP/gRPC.  But the C++ representation does not need to be a floating point?  And representing durations as integers or floats is a "Known Problem" [tm] the units are implicit, implicit conversions can result in bad values, etc.
</blockquote></details>

Changed to milliseconds.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10253)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

___
*[`google/cloud/rest_options.h` line 59 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10253#-NGxHfv3B7EK78NZaIZw-r1-59:-NGxcVNN7XHYu5xPI4Gg:b-nfmcqu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/dee896b7a54e03f66cd1e33af9f401d1024fc421/google/cloud/rest_options.h#L59)):*

Thanks. Now help me convince folks that this code is fine:

```cc
g::c::Options F() {
  using namespace std::chrono_literals;
  return Options{}.set<g::c::ServerTimeoutOption>(3s + 14ms);
}
```

</blockquote></details>

it's a perfectly cromulent expression

devbww (approved):
Nit: The `Options{}.set<ServerTimeoutOption>("3.14")` from the PR description wouldn't have compiled, right?  That is, it wasn't a string.

Issue #5915, opened 2021-02-24
Support third-party identity


coryan:
We need to support third-party identities, like gRPC does in:

https://github.com/grpc/proposal/blob/master/L77-core-cpp-third-party-identity-support-for-call-credentials.md

More details available internally at go/guac-3pi

- [x] Create token source for files
- [x] Create token source for OIDC
- [x] Create token source for AWS v1
- [x] Implement token exchange
- [x] Implement service account impersonation workflow
- [x] Add support in `oauth2_internal::MapCredentials()`
- [x] Add support in `storage::oauth2::MapCredentials()`
- [x] Add support in `g::c::internal::CreateAuthenticationStrategy`
- [x] Add support in `g::c::oauth2_internal::GoogleDefaultCredentials()`
- [x] Consider a new factory function `g::c::MakeExternalAccountCredentials()`
- [x] Manually test with Azure
- [x] Manually test with AWS
- [x] Propagate `g::c::Options` from factory function to the token sources and token exchange

coryan:
Still needed.

coryan:
Strike 2, keep open.

PR #10259, opened 2022-11-16
impl(common): classes to propagate error information

coryan:
Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10259)


google/cloud/internal/error_metadata.h
```diff
@@ -0,0 +1,158 @@
...
+};
+
+std::string Format(absl::string_view message, ErrorContext const& context);
+
+/**
+ * Stores an error context for later use.
+ *
+ * In most cases an error context is used in a single call tree, and therefore
+ * using `absl::string_view` gives a good performance tradeoff.  Sometimes one
+ * needs to capture the error context for use after the function returns.  In
```
devbww, line 111: It's a little more insidious than "after the function returns", methinks.  `ErrorContext` is a non-owning object, and its lifetime needs to be contained within the lifetime of its ctor argument, so its doc block probably deserves a big warning to that effect (or something like ABSL_ATTRIBUTE_LIFETIME_BOUND perhaps).

devbww (commented):
I'm not sure of the consensus on this yet, but ...

google/cloud/internal/error_metadata.h
coryan: I think I will switch this to have simpler lifetime semantics.  The use-cases I have in mind are not in the critical path and are not worth the extra thinking required to avoid problems.

In the future we may care about avoiding allocations for a bag of attributes that is unlikely to be used.

coryan:
PTAL

google/cloud/internal/error_metadata.cc
```diff
@@ -0,0 +1,35 @@
...
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+std::string Format(absl::string_view message, ErrorContext const& context) {
+  if (context.empty()) return std::string{message};
+  auto format = [](std::string* out, auto const& i) {
+    absl::StrAppend(out, i.first, "=", i.second);
+  };
```
devbww, line 28: I believe this is the equivalent of `absl::PairFormatter("=")`.

coryan: Fixed.

devbww (approved):

PR #10265, opened 2022-11-17
impl(generator): support experimental clients

### LINKED ISSUE #10263: Generate experimental clients
Sometimes we want to add a non-GA client to an already-GA library. (e.g. if the new service is still in private preview).

To achieve this, we should add a new `bool experimental` field to `ServiceConfiguration`.

Services with that field set, will add an `ExperimentalTag` to the client constructor and the connection factory function.

dbolduc:
Fixes #10263 

I originally used an `$experimental_tag$`, but decided the generator code was not as easy to read.

I am compromising and not adding a new golden service in the generator, just to verify that these two `ExperimentalTag`s show up. Instead, I decided to just turn it on for speech v2. :shrug: 

If someone twists my arm hard enough, I will add a new golden service for testing.

### DIFF
--- a/generator/internal/client_generator.cc
+++ b/generator/internal/client_generator.cc
@@ -71,12 +71,13 @@ Status ClientGenerator::GenerateHeader() {
 
   // includes
   HeaderPrint("\n");
-  HeaderLocalIncludes(
-      {vars("connection_header_path"),
-       HasBidirStreamingMethod() ? "google/cloud/experimental_tag.h" : "",
-       "google/cloud/future.h", "google/cloud/options.h",
-       "google/cloud/polling_policy.h", "google/cloud/status_or.h",
-       "google/cloud/version.h"});
+  HeaderLocalIncludes({vars("connection_header_path"),
+                       HasBidirStreamingMethod() || IsExperimental()
+                           ? "google/cloud/experimental_tag.h"
+                           : "",
+                       "google/cloud/future.h", "google/cloud/options.h",
+                       "google/cloud/polling_policy.h",
+                       "google/cloud/status_or.h", "google/cloud/version.h"});
   if (get_iam_policy_extension_ && set_iam_policy_extension_) {
     HeaderLocalIncludes({"google/cloud/iam_updater.h"});
   }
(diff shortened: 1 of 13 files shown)


generator/internal/client_generator.cc
```diff
@@ -89,32 +90,35 @@ Status ClientGenerator::GenerateHeader() {
...
+
+  //@{
+  // @name Copy and move support
+  $client_class_name$($client_class_name$ const&) = default;
+  $client_class_name$& operator=($client_class_name$ const&) = default;
+  $client_class_name$($client_class_name$&&) = default;
+  $client_class_name$& operator=($client_class_name$&&) = default;
+  //@}
+
+  //@{
```
coryan, line 112: unrelated: these should be `///@{` to actually start a group, ditto for closing the group.

generator/internal/client_generator.cc
```diff
@@ -354,7 +358,11 @@ Status ClientGenerator::GenerateCc() {
 
   CcPrint(  // clang-format off
     "\n"
-    "$client_class_name$::$client_class_name$(std::shared_ptr<$connection_class_name$> connection, Options opts)"
+    "$client_class_name$::$client_class_name$("
```
coryan, line 361: nit: looks like this section could benefit from raw strings too.

generator/internal/service_code_generator_test.cc
```diff
@@ -83,12 +84,44 @@ class TestGenerator : public ServiceCodeGenerator {
...
+      .WillRepeatedly(Return(nullptr));
+
+  TestGenerator g1(service_file_descriptor->service(0),
+                   generator_context.get());
+  EXPECT_FALSE(g1.IsExperimental());
+
+  TestGenerator g2(
+      service_file_descriptor->service(0), generator_context.get(),
+      {{"header_path_key", "header_path"}, {"experimental", "true"}});
+
```
coryan, line 118: There are no expectations for `g2`?

generator/internal/connection_generator.cc
```diff
@@ -247,6 +249,7 @@ class $connection_class_name$ {
 std::shared_ptr<$connection_class_name$> Make$connection_class_name$(
 )""");
   HeaderPrint("    ");
+  if (IsExperimental()) HeaderPrint("ExperimentalTag, ");
```
coryan, line 252: Consider refactoring the code to generate the `Make$connection_class_name$` arguments to some helper functions.  So maybe this will just be:

```cc
  HeaderPrint("    " + ConnectionFactoryFunctionArguments());
```

just a thought.
This code is getting complex enough that maybe we should refactor it to its

coryan (approved):

generator/internal/client_generator.cc
dbolduc: Ack. I'll send a follow up PR that touches all the files.

generator/internal/client_generator.cc
dbolduc: Done.

generator/internal/connection_generator.cc
dbolduc: Done, but I don't love it.

coryan (approved):

PR #10269, opened 2022-11-17
ci: parallelize API report generation

### LINKED ISSUE #10234: Parallelize API report generation
Separately, we parallelize the processes to create the ABI dumps, but we do not parallelize the generation of the comparison reports.

I think on failure we want to see the output of a single report in the logs.  But it takes about 6 seconds per report, and there are about 100 of them.  That is nearly 10 minutes.  We should parallelize the report generation, and only print the failures sequentially.

devbww:
Run `abi-dumper; abi-compliance-checker` in parallel for each library. This reduces the run time of "check-api" (in a totally unscientific, one-off, "ninja: no work to do" test) from around 20m30s to about 7m30s.

Also take this opportunity to improve the output by discarding the noise ("cmake --install" messages, "missed type id" errors, etc.) and by including the elapsed time for each library.

On the latter front, the outliers are ...
```
bigtable completed in 4m2s
dialogflow_cx completed in 6m7s
dialogflow_es completed in 6m56s
```
where we can see that the "6m56" dominates the overall run time.

Fixes #10234

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10269)

### DIFF
--- a/ci/cloudbuild/builds/check-api.sh
+++ b/ci/cloudbuild/builds/check-api.sh
@@ -30,8 +30,8 @@ mapfile -t feature_list < <(bazelisk --batch query \
   sed -e 's;//:;;')
 enabled="$(printf ";%s" "${feature_list[@]}")"
 # These two are not libraries that require enabling.
-enabled="${enabled/;common}"
-enabled="${enabled/;grpc_utils}"
+enabled="${enabled/;common/}"
+enabled="${enabled/;grpc_utils/}"
 enabled="${enabled:1}"
 
 INSTALL_PREFIX=/var/tmp/google-cloud-cpp
@@ -68,7 +68,8 @@ function check_abi() {
     public_headers="${prefix}/include/google/cloud"
   fi
 
-  local version=$(git rev-parse --short HEAD)
+  local version
+  version=$(git rev-parse --short HEAD)
   local actual_dump_file="${library}.actual.abi.dump"
   local -a dump_options=(
     # The source .so file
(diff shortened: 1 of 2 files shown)


ci/cloudbuild/builds/check-api.sh
```diff
@@ -81,57 +92,65 @@ function dump_abi() {
...
+           time check_abi \${0} ${INSTALL_PREFIX} ${PROJECT_ROOT}"
+
+# A count of the number of libraries that fail the api compliance check.
+# This will become the script's exit code.
+errors=0
+
+for library in "${libraries[@]}"; do
+  report="cmake-out/compat_reports/${library}/src_compat_report.html"
+  if grep --silent "<td class='compatible'>100%</td>" "${report}"; then
+    io::log_green "ABI Compliance OK: ${library}"
```
coryan, line 146: Ah, that is clever.  I could not figure out how to capture the status code from a call to `xargs`.

ci/generate-markdown/generate-readme.sh
```diff
@@ -36,7 +36,7 @@ file="README.md"
 (
   mapfile -t libraries < <(bazelisk --batch query \
     --noshow_progress --noshow_loading_progress \
-    'kind(cc_library, //:all) except filter("experimental|mocks|common|grpc_utils|internal", kind(cc_library, //:all))' |
+    'kind(cc_library, //:all) except filter("experimental|mocks|common|grpc_utils", kind(cc_library, //:all))' |
```
coryan, line 39: is this change related?

coryan (approved):

ci/generate-markdown/generate-readme.sh
devbww: Tangentially, at least.  While investigating the _"common is not a valid feature in google-cloud-cpp"_ CMake warning elicited through similar code in check-api.sh, I noticed that there isn't a `//:*internal*` target, so there's no reason to try to filter it out.  So I slipped in this update. :-)

PR #10270, opened 2022-11-17
impl(common): helper functions to create `Status`

coryan:
These will make it easier to return errors, with less syntactic noise. A future PR will make it easier to create error infos used in the library.

Motivated by #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10270)


google/cloud/internal/make_status.h
```diff
@@ -0,0 +1,49 @@
...
+#include "google/cloud/status.h"
+#include "google/cloud/version.h"
+#include <string>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+Status CancelledError(std::string msg, ErrorInfo info);
```
devbww, line 27: Should we default (`info = {}`) the `ErrorInfo`?

coryan: Done.

devbww (approved):

PR #10273, opened 2022-11-18
impl(generator): refactor generated rest stub to use EndpointOption

### LINKED ISSUE #10214: Leverage Options to override rest transport endpoint
Current generated RestStubs hard code the endpoint from the proto file annotation. Modify the emitted code to check the passed Options for a possible override for the default endpoint.

scotthart:
fixes #10214

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10273)

### DIFF
--- a/generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc
+++ b/generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc
@@ -31,8 +31,8 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 DefaultGoldenKitchenSinkRestStub::DefaultGoldenKitchenSinkRestStub(Options options)
     : rest_client_(rest_internal::MakePooledRestClient(
-      options.get<EndpointOption>(), options)),
-      options_(std::move(options)) {}
+          options.get<EndpointOption>(), options)),
+          options_(std::move(options)) {}
 
 DefaultGoldenKitchenSinkRestStub::DefaultGoldenKitchenSinkRestStub(
     std::shared_ptr<rest_internal::RestClient> rest_client, Options options)
--- a/generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc
+++ b/generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc
@@ -32,7 +32,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 DefaultGoldenKitchenSinkRestStub::DefaultGoldenKitchenSinkRestStub(Options options)
     : rest_client_(rest_internal::MakePooledRestClient(
           options.get<EndpointOption>(), options)),
-          options_(std::move(options)) {}
+      options_(std::move(options)) {}
 
 DefaultGoldenKitchenSinkRestStub::DefaultGoldenKitchenSinkRestStub(
     std::shared_ptr<rest_internal::RestClient> rest_client, Options options)
--- a/generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc
+++ b/generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc
@@ -17,6 +17,7 @@
 // source: generator/integration_tests/test.proto
 
 #include "generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.h"
+#include "google/cloud/common_options.h"
 #include "google/cloud/internal/absl_str_cat_quiet.h"
 #include "google/cloud/internal/rest_stub_helpers.h"
 #include "google/cloud/status_or.h"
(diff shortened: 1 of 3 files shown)


generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc
```diff
@@ -30,7 +31,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 DefaultGoldenKitchenSinkRestStub::DefaultGoldenKitchenSinkRestStub(Options options)
     : rest_client_(rest_internal::MakePooledRestClient(
-          "https://goldenkitchensink.googleapis.com", options)),
+      options.get<EndpointOption>(), options)),
```
coryan, line 34: I think the default value for generated libraries is something like `compute.googleapis.com`, or `bigquery.googleapis.com`.  Do I get that right?
If so, do you need to add the `http://` prefix?

generator/internal/stub_rest_generator.cc
```diff
@@ -152,11 +153,10 @@ Status StubRestGenerator::GenerateCc() {
   auto result = CcOpenNamespaces(NamespaceType::kInternal);
   if (!result.ok()) return result;
 
-  // TODO(#10214): Emit code to check options for a potential endpoint override.
   CcPrint(R"""(
 Default$stub_rest_class_name$::Default$stub_rest_class_name$(Options options)
     : rest_client_(rest_internal::MakePooledRestClient(
-          "https://$service_endpoint$", options)),
+      options.get<EndpointOption>(), options)),
```
devbww, line 159: Restore indentation?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10273)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc` line 34 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10273#-NHBVvvPK2SIxDt1r57Q-r1-34:-NHBXb874x0tVTrVc9Fy:b-2m280b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e8e79b5f64419360b9592be7eb6748a29be08f28/generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc#L34)):*

I think the default value for generated libraries is something like `compute.googleapis.com`, or `bigquery.googleapis.com`.  Do I get that right?
If so, do you need to add the `http://` prefix?
</blockquote></details>

You're correct about the default endpoint option value for generated libraries. I'm looking at whether or not I need to sanitize the option value (i.e. add the https:// if it's missing). And I'd prefer to do that in the RestClient rather than here. These types of runtime issues will be verified in the eventual integration tests running against some server.

generator/internal/stub_rest_generator.cc
```diff
@@ -152,12 +153,11 @@ Status StubRestGenerator::GenerateCc() {
...
   if (!result.ok()) return result;
 
-  // TODO(#10214): Emit code to check options for a potential endpoint override.
   CcPrint(R"""(
 Default$stub_rest_class_name$::Default$stub_rest_class_name$(Options options)
     : rest_client_(rest_internal::MakePooledRestClient(
-          "https://$service_endpoint$", options)),
-      options_(std::move(options)) {}
+          options.get<EndpointOption>(), options)),
+          options_(std::move(options)) {}
```
devbww, line 160: It looks like you restored the one line, but then added indentation to the next.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10273)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @coryan and @devbww)

___
*[`generator/internal/stub_rest_generator.cc` line 159 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10273#-NHBVvvPK2SIxDt1r57S-r1-159:-NHBbUh97FflWuAN1aW4:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e8e79b5f64419360b9592be7eb6748a29be08f28/generator/internal/stub_rest_generator.cc#L159)):*

Restore indentation?
</blockquote></details>

Done.
___
*[`generator/internal/stub_rest_generator.cc` line 160 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10273#-NHBVvvPK2SIxDt1r57S-r2-160:-NHBbUzi5h1s05AiLDZf:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/9173c254cd288ebe03764178f53f01840f23cca3/generator/internal/stub_rest_generator.cc#L160)):*

It looks like you restored the one line, but then added indentation to the next.
</blockquote></details>

Done.

coryan (approved):
> Reviewed 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10273)* status: all files reviewed, 2 unresolved discussions (waiting on @devbww)

___
*[`generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc` line 34 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10273#-NHBVvvPK2SIxDt1r57Q-r1-34:-NHUaUSKCEN7PBVBW1Ez:b-svh06d) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e8e79b5f64419360b9592be7eb6748a29be08f28/generator/integration_tests/golden/internal/golden_kitchen_sink_rest_stub.cc#L34)):*

You're correct about the default endpoint option value for generated libraries. I'm looking at whether or not I need to sanitize the option value (i.e. add the https:// if it's missing). And I'd prefer to do that in the RestClient rather than here. These types of runtime issues will be verified in the eventual integration tests running against some server.
</blockquote></details>

Hmm... Okay, maybe it will work.

PR #10274, opened 2022-11-18
impl(common): helper functions to parse JSON configuration

coryan:
To support workload/workforce identity federation (aka BYOID, aka external accounts) we will need to valid the JSON configuration. This involves a lot of "is this field there and is it a string", or "if the field is not there use this default, otherwise it must be a string". These two functions save a lot of repetition in the validation code.

Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10274)


devbww (commented):
My only comment is that the functions themselves are rather generic compared to the filename/namespace.  Might we prefer some kind of "JSON utils" library?

coryan:
> My only comment is that the functions themselves are rather generic compared to the filename/namespace. Might we prefer some kind of "JSON utils" library?

We can always refactor if we find more need for them.  Being in `*_internal` makes it easy to do so.

devbww (approved):
Sure enough.

Issue #10237, opened 2022-11-12
Consider grouping services by library in generator config


dbolduc:
Our repo has a structure to it. We have services contained in libraries. If the generator configuration reflects that, it will help us maintain consistency and confidence in our documentation across all libraries. I am thinking something like:

```proto
message LibraryConfiguration {
  string name = 1;
  string display_name = 2;
  string description = 3;
  string cloud_service_root = 4;
  string cloud_service_docs = 5;
  string cloud_quickstart = 6;
  bool experimental = 7;
  repeated ServiceConfiguration service = 8;

  // we can add more knobs, but we should be cautious about doing so.
}

message GeneratorConfiguration {
  repeated LibraryConfiguration library = 1;

  // Support standalone services, for libraries that have handwritten documentation.
  repeated ServiceConfiguration service = 2;
}
```

At the moment, we generate many things in the scaffolding. https://github.com/googleapis/google-cloud-cpp/blob/a590dc461ad7a725ee0664fb9f61a1d5a4f5b04b/generator/internal/scaffold_generator.cc#L118-L132

I think (at least) the following should be generated for each `LibraryConfiguration`:
* `README.md`
* `doc/main.dox`
* `doc/options.dox`
* `quickstart/.bazelrc`
* `samples/BUILD.bazel`
* `samples/CMakeLists.txt`

We do not want to move things to the generator config, just because. Some things are easier to maintain in place, and impossible to generalize. An obvious example would be the `quickstart/quickstart.cc`'s.

I explored this some in a branch. I started with the `README.md` and `doc/main.dox`:
https://github.com/googleapis/google-cloud-cpp/compare/main...dbolduc:google-cloud-cpp:generator-library-config-dev?expand=1

coryan:
There is a separate effort to have a single source of truth for all client libraries (across all languages).  

I have also been thinking that we could generate `main.dox` from `README.md` with a few alterations to make the generation "easy".

dbolduc:
The individual service generators know their client names / namespaces / filepaths / endpoints. But there is not a great way to get that information back into `standalone_main.cc` in order to generate documentation type stuff.

I like the concept of the `main-dox-marker` https://github.com/googleapis/google-cloud-cpp/blob/ba4669ee64106d6269a451bddcfc6c86cf6cde5a/google/cloud/accessapproval/samples/access_approval_client_samples.cc#L29

Maybe we could extend it and have each service generate a data file that makes the extraction super-duper easy.
```
[...]

PR #10278, opened 2022-11-19
feat(vmwareengine): generate library

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10278)


devbww:
Note: There were a few hand edits to the generated code for things we need to teach the generator to do.  Some (moving files for the new layout) are hidden in the "Run generators" commit, while others (fixing BUILD.bazel and CMakeLists.txt) are obvious in the "Manually update" commit.  Anyway, the idea was to have a concrete, working example on which to base the generator changes and HowTo tweaks.

google/cloud/vmwareengine/BUILD.bazel
```diff
@@ -0,0 +1,75 @@
...
+service_dirs = [
+    "v1/",
+]
+
+internal_dirs = [
+    "",
+    "internal/",
+]
+
+src_dirs = [s + i for s in service_dirs for i in internal_dirs]
```
coryan, line 28: No action required: I think in almost all cases it will be easier to just unroll these lists.  Most services will have 1 or 2 versions, so we can just say:

```
service_dirs=["v1/"]
srcs_dirs = ["v1/", "v1/internal/"]
```

or (in a few cases):

```
service_dirs=["v1/", "v2/"]
srcs_dirs = ["v1/", "v1/internal/", "v2/", "v2/internal/"]
```

That is easier to grok and maintain that the for loops (I think).  If we must use for loops we should consider:

```
service_dirs = ["v1/", "v2/"]
src_dirs = service_dirs + [i + "internal/" for i in service_dirs]
```

google/cloud/vmwareengine/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
...
+# ~~~
+
+include(GoogleapisConfig)
+set(DOXYGEN_PROJECT_NAME "VMware Engine API C++ Client")
+set(DOXYGEN_PROJECT_BRIEF "A C++ Client Library for the VMware Engine API")
+set(DOXYGEN_PROJECT_NUMBER "${PROJECT_VERSION}")
+set(DOXYGEN_EXCLUDE_SYMBOLS "internal")
+set(DOXYGEN_EXAMPLE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/quickstart)
+
+unset(mocks_globs)
```
coryan, line 24: Ditto, in most cases unrolling these loops will be easier to grok.

coryan (commented):
Some comments for future consideration, but this is good AS-IS.,

google/cloud/vmwareengine/BUILD.bazel
devbww: I believe I concur. I'll chat to @dbolduc about it, and see what can be done under cover of https://github.com/googleapis/google-cloud-cpp/issues/10170.

google/cloud/vmwareengine/CMakeLists.txt
devbww: Ack.

devbww:
> Some comments for future consideration, but this is good AS-IS.

Did you miss clicking "Approve"?

coryan (approved):

coryan:
> > Some comments for future consideration, but this is good AS-IS.
> 
> Did you miss clicking "Approve"?

Doh!  Fixed.

google/cloud/vmwareengine/BUILD.bazel
dbolduc: It's not a question about how many versions. It is a question about how many subdirectories.

I concede that the most common size of the list will probably be 2. (if/when we move the services to a new dir and leave backwards compat headers). The worst case (that I know of) is `bigquery` which will need:

```
set(service_dirs "" "analyticshub/v1/" "connection/v1/"
                 "datatransfer/v1/" "migration/v2/"
                 "reservation/v1/" "storage/v1/" "v2/")
```

The crazy plan I am considering is to
1. #10237
2. factor out the `cc_library(...)`'s into a `google/cloud/foo/foo.bzl`. (I think we only need the `deps`. There are libraries that are [only compatible on certain platforms](https://github.com/googleapis/google-cloud-cpp/blob/7d35ce8ea8a14d66c7a1f64acd7d36eef3812626/google/cloud/asset/BUILD.bazel#L44-L48). I hope this could be captured in the generator configuration. Which would also generate [documentation](https://github.com/googleapis/google-cloud-cpp/blame/main/google/cloud/asset/README.md#L11-L17) that is not stale.)
3. generate the `BUILD.bazel` file. (we know all the subdirectories; they are defined in the generator configuration).

The `defs.bzl` would be initially created by the scaffold generator, and maintained by hand. We might do a similar thing with `CMakeLists.txt` -> `lib.cmake` for the quickstart configuration.

Am I actively working on this crazy plan? No. It is just a thought that I would explore if I had infinite time.

PR #10280, opened 2022-11-19
fix(common): create default gRPC credentials only if needed

coryan:
For backwards compatibility with the pre-Unified Credentials world we were calling `grpc::GoogleDefaultCredentials()` if `g::c::GrpcCredentialsOption` was not set. In many cases the option is not set, but `g::c::UnifiedCredentialsOption` is. If that is the case, the unified options are preferred. Calling
`grpc::GoogleDefaultCredentials()` produces spurious errors, and these credentials won't be used.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10280)


devbww (approved):
Thanks.

PR #10284, opened 2022-11-20
doc(spanner): deprecate old MakeConnection() overloads

devbww:
Raise the disfavor for `ConnectionOptions` and `SessionPoolOptions` arguments to `spanner::MakeConnection()` from @note to @deprecated.

Also removed some stale inclusions of `connection_options.h` and `session_pool_options.h`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10284)

### DIFF
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -961,8 +961,8 @@ std::shared_ptr<spanner::Connection> MakeConnection(spanner::Database const& db,
  * The returned connection object should not be used directly, rather it should
  * be given to a `Client` instance, and methods should be invoked on `Client`.
  *
- * @note Prefer using the `MakeConnection()` overload that accepts
- *     `google::cloud::Options`.
+ * @deprecated Please use the `MakeConnection()` overload that accepts
+ *     `google::cloud::Options` instead.
  *
  * @see `Connection`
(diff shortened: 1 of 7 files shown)


dbolduc (approved):

google/cloud/spanner/client.h
```diff
@@ -961,8 +961,8 @@ std::shared_ptr<spanner::Connection> MakeConnection(spanner::Database const& db,
  * The returned connection object should not be used directly, rather it should
  * be given to a `Client` instance, and methods should be invoked on `Client`.
  *
- * @note Prefer using the `MakeConnection()` overload that accepts
- *     `google::cloud::Options`.
+ * @deprecated Please use the `MakeConnection()` overload that accepts
```
coryan, line 964: Should we add the `[[deprecated]]` attribute too?

coryan (approved):

google/cloud/spanner/client.h
devbww: I planned to leave that until the old `DatabaseAdminClient` goes away.  But, maybe it would be better to try to kill them both with the same stone.  I'll think on it.

PR #10287, opened 2022-11-21
ci: create build for external accounts integration

coryan:
We need integration tests for external account support. The existing CI systems run on Google infrastructure and therefore not suitable for these tests. GitHub Actions are a bit underpowered, but if we keep the number of tests in this builds small, and the caches are warm, the builds are fast enough.

Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10287)

### DIFF
--- a/.github/workflows/external-account-integration.yml
+++ b/.github/workflows/external-account-integration.yml
@@ -1,0 +1,55 @@
+name: "External Account Integration"
+
+# Build on pull requests and pushes to `main`. The PR builds will be
+# non-blocking for now, but that is configured elsewhere.
+on:
+  pull_request:
+  push:
+    branches: [ 'main' ]
+
+jobs:
+  # A minimal build to validate external account (aka Workload/Workforce
+  # Identity Federation, aka WIF, aka BYOID).  As the name implies, external
+  # accounts support non-Google sources of identity, such as AWS, Azure, or
+  # GitHub. Most of our builds use Google Cloud Build (GCB), which is not
+  # usable in this case.
+  identity-federation-integration-test:
+    name: external-account-integration-test
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: install ninja
+        run: sudo apt install ninja-build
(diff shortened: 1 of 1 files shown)


.github/workflows/external-account-integration.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: cache-vcpkg
+        id: cache-vcpkg
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/vcpkg
+          key: |
+            vcpkg-${{ steps.vcpkg-version.outputs.version }}-${{ hashFiles('vcpkg.json') }}
+          restore-keys: |
+            vcpkg-${{ steps.vcpkg-version.outputs.version }}-
```
dbolduc, line 45: I thought the trailing `-` was weird, but TIL: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-multiple-restore-keys

.github/workflows/external-account-integration.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      # Configuring CMake installs the dependencies from the vcpkg cache, or
```
dbolduc, line 47: Do we need to `vcpkg install google-cloud-cpp[core,iam,storage]`?

dbolduc (approved):

.github/workflows/external-account-integration.yml
coryan: Ack

.github/workflows/external-account-integration.yml
coryan: No.  We don't want to compile the version of `google-cloud-cpp` shipped with vcpkg, just the dependencies of `google-cloud-cpp`.  That happens automagically, thanks to the `vcpkg.json` in our top-level directory.

dbolduc: Ah.

Issue #10283, opened 2022-11-20
Set up infrastructure to compile with OpenTelemetry


dbolduc:
These tasks correspond to [Milestone 1](https://docs.google.com/document/d/1WfWxVt8PuVhmXbjJnNDpb5eijTkM2LQVEXL2mwJRLcE/edit?resourcekey=0-Cfq_Vsq-PEhFntaVdvUG9Q#bookmark=id.oy8jxjrvrcyc) of an internal design doc.

- [ ] Update fedora-36-cmake.Dockerfile
- [ ] Add bazel `bool_flag`, internal `config_setting`s (unused)
- [ ] Load OTel bazel rule; set flags in top level .bazelrc
- [ ] Add no-otel-bazel trigger, build.
- [ ] cmake install: find dependency + update build script
- [ ] Add bazel + cmake switch and enabled/disabled test (no deps yet)
- [ ] Add minimal file with 1 OT symbol. add deps for bazel, cmake
- [ ] Update CentOS-7, Ubuntu dockerfiles

PR #10294, opened 2022-11-22
impl(otel): bazel loads OpenTelemetry

dbolduc:
Part of the work for #10283

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10294)

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -34,3 +34,7 @@ grpc_deps()
 load("@com_github_grpc_grpc//bazel:grpc_extra_deps.bzl", "grpc_extra_deps")
 
 grpc_extra_deps()
+
+load("@io_opentelemetry_cpp//bazel:repository.bzl", "opentelemetry_cpp_deps")
+
+opentelemetry_cpp_deps()
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -186,3 +186,14 @@ def google_cloud_cpp_deps():
             patch_args = ["-p1"],
             patches = [Label("//bazel:configure_template.bzl.patch")],
         )
+
+    # Open Telemetry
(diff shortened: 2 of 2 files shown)


WORKSPACE.bazel
```diff
@@ -34,3 +34,7 @@ grpc_deps()
 load("@com_github_grpc_grpc//bazel:grpc_extra_deps.bzl", "grpc_extra_deps")
 
 grpc_extra_deps()
+
+load("@io_opentelemetry_cpp//bazel:repository.bzl", "opentelemetry_cpp_deps")
```
coryan, line 38: Will we need to change all the quickstart WORKSPACE files too?  Not now, but it a future PR?
Maybe it is time to create the multiple workspace functions like TensorFlow does.

bazel/google_cloud_cpp_deps.bzl
```diff
@@ -186,3 +186,14 @@ def google_cloud_cpp_deps():
             patch_args = ["-p1"],
             patches = [Label("//bazel:configure_template.bzl.patch")],
         )
+
+    # Open Telemetry
+    if "io_opentelemetry_cpp" not in native.existing_rules():
```
coryan, line 191: The renovate configuration will also need a PR, one where both CMake and Bazel get changed in a single PR.

coryan (approved):

bazel/google_cloud_cpp_deps.bzl
dbolduc: Ack. I added this as an extra task to the GH issue.

WORKSPACE.bazel
dbolduc: > Will we need to change all the quickstart WORKSPACE files too? Not now, but it a future PR?

Yeah, we will. That is a task for Milestone 2.

> Maybe it is time to create the multiple workspace functions like TensorFlow does.

:shrug:

PR #10309, opened 2022-11-28
chore: teach renovate how to parse googletest version

dbolduc:
Manually parse the version for `googletest`. [Previous runs](https://app.renovatebot.com/dashboard#github/googleapis/google-cloud-cpp/908580241) would fail with:
> DEBUG: Dependency google/googletest has unsupported value release-1.12.1

I used the same `renovate.json` in my test repo and the bots sent me this: https://github.com/dbolduc/test-renovate/pull/13

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -69,7 +69,7 @@
     {
       "matchPackageNames": ["com_google_googletest", "google/googletest"],
       "groupName": "GoogleTest",
-      "versioning": "loose"
+      "versioning": "regex:^release-(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)?$"
     },
     {
       "matchManagers": ["pip_requirements"],


coryan (approved):

.github/renovate.json
```diff
@@ -69,7 +69,7 @@
     {
       "matchPackageNames": ["com_google_googletest", "google/googletest"],
       "groupName": "GoogleTest",
-      "versioning": "loose"
+      "versioning": "regex:^release-(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)?$"
```
devbww, line 72: If the patch is optional, should its separating period be inside the group?

dbolduc: Yeah, that is strange. I copied the example from here without thinking about it: https://docs.renovatebot.com/modules/versioning/#using-a-custom-regex-versioning-scheme

I'll send a follow up to just make the patch mandatory.

PR #10317, opened 2022-11-28
feat(pubsub): blocking pulls

### LINKED ISSUE #7187: Create synchronous / blocking pull

coryan:
This adds blocking pulls to the Pub/Sub public APIs. It also includes a minimal example, and an integration test.

Fixes #7187

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10317)

### DIFF
--- a/google/cloud/pubsub/integration_tests/subscriber_integration_test.cc
+++ b/google/cloud/pubsub/integration_tests/subscriber_integration_test.cc
@@ -523,6 +523,32 @@ TEST_F(SubscriberIntegrationTest, ExactlyOnce) {
   EXPECT_STATUS_OK(result.get());
 }
 
+TEST_F(SubscriberIntegrationTest, BlockingPull) {
+  auto publisher = Publisher(MakePublisherConnection(topic_));
+  auto subscriber =
+      Subscriber(MakeSubscriberConnection(exactly_once_subscription_));
+
+  std::set<std::string> ids;
+  for (auto const* data : {"message-0", "message-1", "message-2"}) {
+    auto response =
+        publisher.Publish(MessageBuilder{}.SetData(data).Build()).get();
+    EXPECT_STATUS_OK(response);
+    if (response) ids.insert(*std::move(response));
+  }
+  EXPECT_THAT(ids, Not(IsEmpty()));
+
+  auto const count = 2 * ids.size();
+  for (std::size_t i = 0; i != count && !ids.empty(); ++i) {
+    auto response = subscriber.Pull();
+    EXPECT_STATUS_OK(response);
+    if (!response) continue;
+    auto ack = std::move(response->handler).ack().get();
+    EXPECT_STATUS_OK(ack);
+    ids.erase(response->message.message_id());
+  }
+  EXPECT_THAT(ids, IsEmpty());
+}
+
 /// @test Verify the backwards compatibility `v1` namespace still exists.
 TEST_F(SubscriberIntegrationTest, BackwardsCompatibility) {
   auto connection = ::google::cloud::pubsub::v1::MakeSubscriberConnection(
--- a/google/cloud/pubsub/pull_response.h
+++ b/google/cloud/pubsub/pull_response.h
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2022 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -35,7 +35,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
  * the invocation succeeds. With best-efforts subscriptions, the service *may*
  * redeliver the message, even after a successful `handler.ack()` invocation.
  *
- * If `handler` is not an rvalue, you may need to use `std::move(ack).ack()`
+ * If `handler` is not an rvalue, you may need to use `std::move(handler).ack()`
  * or `std::move(handler).nack()`.
  *
  * @see https://cloud.google.com/pubsub/docs/exactly-once-delivery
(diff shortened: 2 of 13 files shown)


google/cloud/pubsub/pull_response.h
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * If the application invokes `handler.nack()` or allows `handler` to go out
+ * of scope, then the service will redeliver the message.
+ *
+ * With exactly-once delivery subscriptions, the service will stop
+ * redelivering the message once the application invokes `handler.ack()` and
+ * the invocation succeeds. With best-efforts subscriptions, the service *may*
+ * redeliver the message, even after a successful `handler.ack()` invocation.
+ *
+ * If `handler` is not an rvalue, you may need to use `std::move(ack).ack()`
```
dbolduc, line 38: s/std::move(ack)/std::move(handler)/ ?

google/cloud/pubsub/pull_response.h
```diff
@@ -0,0 +1,55 @@
+// Copyright 2020 Google LLC
```
dbolduc, line 1: nit

google/cloud/pubsub/integration_tests/subscriber_integration_test.cc
```diff
@@ -523,6 +523,32 @@ TEST_F(SubscriberIntegrationTest, ExactlyOnce) {
...
+  std::set<std::string> ids;
+  for (auto const* data : {"message-0", "message-1", "message-2"}) {
+    auto response =
+        publisher.Publish(MessageBuilder{}.SetData(data).Build()).get();
+    EXPECT_STATUS_OK(response);
+    if (response) ids.insert(*std::move(response));
+  }
+  EXPECT_THAT(ids, Not(IsEmpty()));
+
+  auto const count = 2 * ids.size();
```
dbolduc, line 540: Why do we do 2*N iterations for an exactly-once subscription?

Is it to make sure everything gets acked, in the failure case where we `continue`?

google/cloud/pubsub/subscriber.h
```diff
@@ -157,6 +157,30 @@ class Subscriber {
   future<Status> Subscribe(ExactlyOnceApplicationCallback cb,
                            Options opts = {});
 
+  /**
+   * Pulls one message from @p subscription.
+   *
+   * @par Idempotency
+   * @parblock
+   * This is an idempotent operation; it only reads messages from the service.
+   * Will make multiple attempts to pull a message from the service, subject
```
dbolduc, line 166: s/Will/It will/ ?

coryan: Fixed

google/cloud/pubsub/integration_tests/subscriber_integration_test.cc
coryan: Yes. For low values of "make sure".

google/cloud/pubsub/pull_response.h
coryan: Fixed

google/cloud/pubsub/pull_response.h
coryan: Fixed.

dbolduc (approved):

Issue #10316, opened 2022-11-28
Use decorators to implement credential caching and logging


coryan:
As I work in #5915 it is clear that I will need good logging for credentials (see #3429) and I don't want to repeat the code to cache credentials either.

Using a decorator seems like a better solution, but we need to change some signatures (these are Okay, they are in `*_internal` namespaces).

## Overview

The `g::c::oauth2_credentials::Credentials` class will change from:

```cc
class google::cloud::oauth2_internal::Credentials {
 public:
   virtual StatusOr<std::pair<std::string, std::string>> AuthorizationHeader();
};
```

to:

```cc
// Already existing ... just a reminder for this bug
struct google::cloud::internal::AccessToken {
  std::string token; // empty for anonymous credentials
  std::chrono::system_clock::time_point expiration;
};

class google::cloud::oauth2_internal::Credentials {
 public:
   virtual StatusOr<internal::AccessToken> GetToken(std::chrono::system_clock::time_point now) = 0;
};
```

Returning an access token makes it possible to log the returned token (truncated for security reasons) and its expiration time.  Passing in the current time makes it easier to test the implementation.

## Work Breakdown

I am planning to

1. Introduce the `GetToken()` function and a caching decorator
1. Change the existing functions returning `g::c::oauth2_internal::RefreshingCredentialsWrapper::TemporaryToken` to return `internal::AccessToken`
1. Change the existing classes, remove `AuthorizationHeader()` and the custom caching loops
    * This will be a large CL
1. Remove `g::c::oauth2_internal::RefreshingCredentialsWrapper`
1. Add a logging decorator for credentials, enabled only if some tracing component is present (`auth`?)

## Future Work

There may be some opportunities to add a `type` field to `AccessToken`.  Normally this would be `Bearer`, but could be `None` or some other magical string to indicate the credentials are anonymous.

We may consider a different tracing component to enable low-level (HTTP headers and traffic) for `auth` requests.

PR #10319, opened 2022-11-29
impl(common): a new decorator to cache credentials

coryan:
Multiple credential types are expensive to create. Fortunately, they typically can be cached for about an hour. This PR adds a new member function to `oauth2_internal::Credentials` that works better with decorators for caching. The PR also includes a caching decorator, but this is not used yet. Future PRs will change the code to use the new function and decorator.

Part of the work for #10316

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10319)


google/cloud/internal/oauth2_cached_credentials_test.cc
```diff
@@ -0,0 +1,180 @@
...
+TEST(CachedCredentials, SignBlob) {
+  auto mock = std::make_shared<MockCredentials>();
+  auto const expected = std::vector<std::uint8_t>{1, 2, 3};
+  EXPECT_CALL(*mock, SignBlob(absl::make_optional(std::string("test-account")),
+                              "test-blob"))
+      .WillOnce(Return(expected));
+  CachedCredentials tested(mock);
+  auto actual = tested.SignBlob("test-account", "test-blob");
+  ASSERT_STATUS_OK(actual);
+  EXPECT_THAT(*actual, ElementsAreArray(expected.begin(), expected.end()));
```
dbolduc, line 159: nit: I think we can just say `ElementsAreArray(expected)`

dbolduc (approved):

PR #10322, opened 2022-11-29
impl(otel): define GOOGLE_CLOUD_CPP_HAVE_OPEN_TELEMETRY

dbolduc:
Part of the work for #10283 

define `GOOGLE_CLOUD_CPP_HAVE_OPEN_TELEMETRY` if OpenTelemetry is present in cmake, or requested in bazel. Verify that the defines switch works correctly with a test.

(Aside: will my `otel-disabled-bazel` build run on presubmit??)

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10322)

### DIFF
--- a/google/cloud/google_cloud_cpp_common.cmake
+++ b/google/cloud/google_cloud_cpp_common.cmake
@@ -150,7 +150,6 @@ target_link_libraries(
 # Take a dependency on the OpenTelemetry API, if it is available.
 find_package(opentelemetry-cpp CONFIG)
 if (opentelemetry-cpp_FOUND)
-    target_link_libraries(google_cloud_cpp_common PUBLIC opentelemetry-cpp::api)
     target_compile_definitions(
         google_cloud_cpp_common
         PUBLIC # Enable OpenTelemetry features in google-cloud-cpp
--- a/google/cloud/google_cloud_cpp_common.cmake
+++ b/google/cloud/google_cloud_cpp_common.cmake
@@ -147,6 +147,15 @@ target_link_libraries(
            absl::variant
            Threads::Threads
            OpenSSL::Crypto)
+# Take a dependency on the OpenTelemetry API, if it is available.
+find_package(opentelemetry-cpp CONFIG)
+if (opentelemetry-cpp_FOUND)
+    target_link_libraries(google_cloud_cpp_common PUBLIC opentelemetry-cpp::api)
+    target_compile_definitions(
+        google_cloud_cpp_common
+        PUBLIC # Enable OpenTelemetry features in google-cloud-cpp
+               GOOGLE_CLOUD_CPP_HAVE_OPEN_TELEMETRY)
+endif ()
 google_cloud_cpp_add_common_options(google_cloud_cpp_common)
 target_include_directories(
     google_cloud_cpp_common PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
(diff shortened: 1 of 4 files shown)


google/cloud/google_cloud_cpp_common.cmake
```diff
@@ -147,6 +147,14 @@ target_link_libraries(
            absl::variant
            Threads::Threads
            OpenSSL::Crypto)
+# Take a dependency on the OpenTelemetry API, if it is available.
+find_package(opentelemetry-cpp CONFIG)
```
coryan, line 151: I think we can fix this later, but this may create problems with vcpkg (and other package managers).  Basically, if somebody does:

```
vcpkg install opentelemetry-cpp
vcpkg install google-cloud-cpp
```

Then `google-cloud-cpp` has a dependency on `opentelemetry-cpp` that the package manager knows nothing about.  While if they do:

```
vcpkg remove opentelemetry-cpp
vcpkg install google-cloud-cpp
```

Then they get a different set of deps.  I think you want something like:

```
if (experimental-opentelemetry IN_LIST GOOGLE_CLOUD_CPP_ENABLE)
  find_package(opentelemetry-cpp CONFIG QUIET)
  if (opentelemetry-cpp_FOUND)
... ...
endif ()

Comment (coryan) [reply google/cloud/google_cloud_cpp_common.cmake:151]: Hmmm, we have to fix this before the next release, or we will not be able to push this to vcpkg, and Conda will pick up the change automatically.

**Review by coryan — APPROVED** [review]

File: google/cloud/google_cloud_cpp_common.cmake (hunk 1 above)
Comment (dbolduc) [reply google/cloud/google_cloud_cpp_common.cmake:151]: Argh. Good point. It makes sense for customers to explicitly request the feature anyway.

This PR is currently harmless (we are not taking a dep on opentelemetry, yet, despite the comment saying otherwise). But I will slap the `do not merge` label on it anyway, and add the cmake feature in a separate PR.

File: google/cloud/internal/open_telemetry_enabled_test.cc [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gmock/gmock.h>
+
+TEST(OpenTelemetry, VerifyEnabled) {
+#ifdef GOOGLE_CLOUD_CPP_HAVE_OPEN_TELEMETRY
```
coryan, line 18: Feels that a better test would be something like:

```cc
#ifdef GOOGLE_CLOUD_CPP_HAVE_OPEN_TELEMETRY
#include <opentelemetry/version.h>
#include <gmock/gmock.h>

TEST(OpenTelemetry, VerifyUsable) {
  auto version = std::string{OPENTELEMETRY_VERSION};
  EXPECT_THAT(version, Not(IsEmpty());
  ... ... Maybe use a function here ... ... 
}
```

dbolduc: I am testing a different thing. I want a test that verifies that `GOOGLE_CLOUD_CPP_HAVE_OPEN_TELEMETRY` is defined. If it was not defined than the thing you proposed is an empty test fixture, which will quietly succeed.

We should have a test like the one you propose. But we are not ready for it yet.

coryan: But that is an odd test.  You are basically testing that CMake / Bazel works as advertised?

dbolduc: It is a weird test. But it is the only thing that will notice if we accidentally remove the `defines` / `target_compile_definitions`

dbolduc:
> You are basically testing that CMake / Bazel works as advertised?

Ok, I am convinced this test is goofy and isn't buying us much. Closing.

Issue #9171, opened 2022-06-06
C++ Pub/Sub stops receiving


BrianPez:
Using the online documentation and the code in the sample repo, I've implemented PubSub for a single incoming and multiple outgoing.  The pull PubSub is an ordering subscription (would much prefer an ordered and deliever once, but that doesn't seem allowed).  I have dead letter configured for each PubSub.  Repeatably, the code will stop pulling messages.  Sometimes in 5 minutes, sometimes after a couple of days.  This morning, there were 55k messages waiting to be pulled.  No messages in the dead letter.  Based on the website, I expected something from the then() handler in the subscription, but there is no error messages.  Here is my code.

```c++
/* These are based on https://github.com/googleapis/google-cloud-cpp/blob/HEAD/google/cloud/pubsub/samples/samples.cc */
/* https://cloud.google.com/pubsub/docs/handling-failures#dead-letter_topic */
/* https://cloud.google.com/blog/products/application-development/running-cplusplus-apps-with-the-help-of-pubsub-and-gke */
/* https://cloud.google.com/pubsub/docs/samples/pubsub-subscriber-error-listener#code-sample */
#include <google/cloud/pubsub/publisher.h>
#include <google/cloud/storage/client.h>
#include <google/cloud/pubsub/subscriber.h>
#include <google/cloud/pubsub/subscription_admin_client.h>
#include <google/cloud/pubsub/topic_admin_client.h>
#include <nlohmann/json.hpp>

    void createTopic(const std::string& topicId, const std::string& dlTopicId, google::cloud::pubsub::TopicAdminClient& client)
    {
        namespace pubsub = ::google::cloud::pubsub;

        [](pubsub::TopicAdminClient client, std::string project_id, std::string topic_id, std::string dl_topic_id)
            {
                auto topic = client.CreateTopic(pubsub::TopicBuilder(pubsub::Topic(project_id, topic_id)));
                // Note that kAlreadyExists is a possible error when the library retries.
                if (topic.status().code() == google::cloud::StatusCode::kAlreadyExists)
                {
                    ;
                }
                else if (!topic)
                {
                    throw std::runtime_error(topic.status().message());
                }
                else
                {
                    LOGn << "The topic " << topic_id << " was successfully created: " << topic->DebugString();
                }

                auto dl_topic = client.CreateTopic(pubsub::TopicBuilder(pubsub::Topic(project_id, dl_topic_id)));
[...]

PR #10327, opened 2022-11-29
feat(pubsub): add option to override subscription

coryan:
For historical reasons, a `pubsub::SubscriberConnection` has a subscription assigned to it.  With this change that subscription becomes the default, and one can provide an override via the `Options` argument.

Motivated by #10320 and #9171

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10327)

### DIFF
--- a/google/cloud/pubsub/options.h
+++ b/google/cloud/pubsub/options.h
@@ -369,7 +369,8 @@ struct ShutdownPollingPeriodOption {
  * Override the default subscription for a request.
  *
  * Some applications need to receive messages from multiple subscriptions. In
- * these case they can use this option to override the default
+ * this case they can use this option to override the default subscription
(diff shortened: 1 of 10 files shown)


google/cloud/pubsub/options.h
```diff
@@ -364,12 +365,25 @@ struct ShutdownPollingPeriodOption {
   using Type = std::chrono::milliseconds;
 };
 
+/**
+ * Override the default subscription for a request.
+ *
+ * Some applications need to receive messages from multiple subscriptions. In
+ * these case they can use this option to override the default
```
devbww, line 372: s/these case/these cases/ or /this case/

devbww, line 372: s/$/./

devbww (approved):

PR #10331, opened 2022-11-30
impl(otel): add experimental cmake option

dbolduc:
Part of the work for #10283 

Add a top level cmake option to enable open telemetry as an experimental feature. Use it in the `clang-tidy` build. Leave a `TODO` / `message()` placeholder to speak to the intended use of the feature.

Arguably the option is unnecessary. We could just use `-DGOOGLE_CLOUD_CPP_ENABLE`.... but I think being able to append a `-DGOOGLE_CLOUD_CPP_EXPERIMENTAL_OPENTELEMETRY` adds some amount of convenience to the customers.

Aside: I decided I no longer want to break up `opentelemetry` into two words in our code. I will update the existing uses (e.g. `//:experimetnal-open_telemetry` -> `//:experimental-opentelemetry).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,6 +199,10 @@ option(GOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC
        "${GOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC_DEFAULT}")
 mark_as_advanced(GOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC)
 
+option(GOOGLE_CLOUD_CPP_ENABLE_EXPERIMENTAL_OPENTELEMETRY
+       "Compile google-cloud-cpp with OpenTelemetry (EXPERIMENTAL)")
+mark_as_advanced(GOOGLE_CLOUD_CPP_ENABLE_EXPERIMENTAL_OPENTELEMETRY)
+
 # We no longer build the generator by default, but if it was explicitly
 # requested, we add it to the list of enabled libraries.
 if (GOOGLE_CLOUD_CPP_ENABLE_GENERATOR)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -199,6 +199,10 @@ option(GOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC
        "${GOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC_DEFAULT}")
 mark_as_advanced(GOOGLE_CLOUD_CPP_STORAGE_ENABLE_GRPC)
 
+option(GOOGLE_CLOUD_CPP_ENABLE_EXPERIMENTAL_OPENTELEMETRY
```
coryan, line 202: We are moving away from the `GOOGLE_CLOUD_CPP_ENABLE_*` options and trying to consolidate *only* on `GOOGLE_CLOUD_CPP_ENABLE`.  Do you have a reason to support both in this case?

ci/cloudbuild/builds/clang-tidy.sh
```diff
@@ -31,7 +31,7 @@ mapfile -t cmake_args < <(cmake::common_args)
 cmake "${cmake_args[@]}" \
   -DCMAKE_CXX_CLANG_TIDY=/usr/local/bin/clang-tidy-wrapper \
   -DCMAKE_CXX_STANDARD=14 \
-  -DGOOGLE_CLOUD_CPP_ENABLE="bigtable;bigquery;generator;iam;logging;pubsub;pubsublite;spanner;storage" \
+  -DGOOGLE_CLOUD_CPP_ENABLE="bigtable;bigquery;experimental-opentelemetry;generator;iam;logging;pubsub;pubsublite;spanner;storage" \
```
coryan, line 34: Aside: we should consider building *all* features here.

CMakeLists.txt
dbolduc: I have a weak argument.

1. I assume an option is more discoverable. (I assume there is some way to list all options for a cmake project).
2. Being able to append a `-DGOOGLE_CLOUD_CPP_EXPERIMENTAL_OPENTELEMETRY` adds some amount of convenience to the customers? (vs. appending to `GOOGLE_CLOUD_CPP_ENABLE`)

If you are unconvinced, I'll just close this thing.

coryan: I am unconvinced.

Issue #10321, opened 2022-11-29
publish docs flake


dbolduc:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/0112f35c-741d-40b9-9427-0666b615ad54;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/1dfeae7fb2ef76c96ede2f2ad6753617ecfe1ff3/fedora-36-cmake-publish-docs/log-0112f35c-741d-40b9-9427-0666b615ad54.txt

```
T+242.898s [66/99] Generate websecurityscanner HTML documentation
FAILED: google/cloud/websecurityscanner/CMakeFiles/websecurityscanner-docs /workspace/cmake-out/google/cloud/websecurityscanner/CMakeFiles/websecurityscanner-docs 
cd /workspace/google/cloud/websecurityscanner && /usr/bin/cmake -E make_directory /workspace/cmake-out/google/cloud/websecurityscanner && /usr/bin/doxygen /workspace/cmake-out/google/cloud/websecurityscanner/Doxyfile.websecurityscanner-docs
error: source '/workspace/cmake-out/google/cloud/servicemanagement/html/graph_legend.dot' is not a readable file or directory... skipping.
 (warning treated as error, aborting now)
T+242.969s [67/99] Generate servicemanagement HTML documentation
```

I don't really understand this one. I want to call it a race.

dbolduc:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/a36027a9-8588-4237-9f2c-d8f6ca34cc9b;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/212eb81c02030e13783eec95233cc0103814a6d9/fedora-36-cmake-publish-docs/log-a36027a9-8588-4237-9f2c-d8f6ca34cc9b.txt

coryan:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/08198399-b011-4e31-abe9-80114545d22f;tab=detail?project=cloud-cpp-testing-resources
[...]

PR #10333, opened 2022-11-30
cleanup: doxygen does not generate .dot files

dbolduc:
Motivated by #10321 

disable `dot(1)`

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10333)


devbww (approved):

cmake/GoogleCloudCppCommon.cmake
```diff
@@ -86,6 +86,7 @@ elseif (GOOGLE_CLOUD_CPP_GENERATE_DOXYGEN)
         # Any additional includes that the library needs
         ${GOOGLE_CLOUD_CPP_DOXYGEN_EXTRA_INCLUDES})
     set(DOXYGEN_GENERATE_LATEX NO)
+    set(DOXYGEN_HAVE_DOT NO)
```
devbww, line 89: Is there a benefit to actually having dot?

dbolduc: I think it makes the diagrams that we see at the top of this page, for example: https://googleapis.dev/cpp/google-cloud-common/HEAD/group__guac.html

I think we tried to turn it off with our other settings too but doxygen has since evolved.

PR #10338, opened 2022-11-30
impl(scaffold): support new directory structure

dbolduc:
Part of the work for #10170 

Drop the `samples/{BUILD.bazel,CMakeLists.txt}`. Instead build the samples from the library's top level build files.

Support a `service.product_path()` like `google/cloud/foo/bar/baz/quux/v1alpha7` by using `LibraryPath(...)` to strip it to `google/cloud/foo/`.

The changes were based off of @devbww's #10278. These scaffolding changes were used to generate: https://github.com/dbolduc/google-cloud-cpp/tree/feat-paymentgateway-generate-library

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10338)

### DIFF
--- a/generator/internal/scaffold_generator.cc
+++ b/generator/internal/scaffold_generator.cc
@@ -127,6 +127,7 @@ std::map<std::string, std::string> ScaffoldVars(
   vars["copyright_year"] = service.initial_copyright_year();
   vars["library"] = library;
   vars["product_options_page"] = OptionsGroup(service.product_path());
+  vars["service_subdirectory"] = ServiceSubdirectory(service.product_path());
   vars["site_root"] = SiteRoot(service);
   vars["library_prefix"] = experimental ? "experimental-" : "";
   vars["doxygen_version_suffix"] = experimental ? " (Experimental)" : "";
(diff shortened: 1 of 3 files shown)


generator/internal/scaffold_generator.cc
```diff
@@ -305,25 +304,30 @@ package(default_visibility = ["//visibility:private"])
 
 licenses(["notice"])  # Apache 2.0
 
+service_dirs = [
```
devbww, line 307: Is this the conclusion from https://github.com/googleapis/google-cloud-cpp/pull/10278#discussion_r1027103136?

dbolduc: I think the for loops will be easier to maintain, because they are more generic. We maintain one list for each library: `service_dirs`. (And the generator_config knows the values, so I bet we will automate this maintenance one day).

I updated the definitions to be more succinct, as suggested in the comment (sorta). I figured I might as well update `vmwareengine/BUILD.bazel` too.

devbww: SGTM.

We should remember to update the couple of google3 BUILD files to match after the next import.

devbww (approved):

generator/internal/scaffold_generator.cc
```diff
@@ -379,10 +386,20 @@ include(GoogleapisConfig)
...
 set(DOXYGEN_PROJECT_BRIEF "A C++ Client Library for the $title$")
 set(DOXYGEN_PROJECT_NUMBER "$${PROJECT_VERSION}$doxygen_version_suffix$")
-set(DOXYGEN_EXCLUDE_SYMBOLS "internal" "$library$_internal" "$library$_testing"
-                            "examples")
-set(DOXYGEN_EXAMPLE_PATH $${CMAKE_CURRENT_SOURCE_DIR}/samples
-                         $${CMAKE_CURRENT_SOURCE_DIR}/quickstart)
+set(DOXYGEN_EXCLUDE_SYMBOLS "internal")
+set(DOXYGEN_EXAMPLE_PATH $${CMAKE_CURRENT_SOURCE_DIR}/quickstart)
+
+unset(mocks_globs)
```
coryan, line 392: I see you applied the changes to `vmwareengine/BUILD.bazel`, I am guessing as a test? Why not to vmwareengine/CMakeLists.txt`?

dbolduc: > Why not to `vmwareengine/CMakeLists.txt`?

`vmwareengine/CMakeLists.txt` matches the scaffold, as is.

> I am guessing as a test?

I updated `vmwareengine/BUILD.bazel` for the sake of consistency. Of course, I forgot about `speech/BUILD.bazel`: https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/speech/BUILD.bazel#L24-L29

PR #10349, opened 2022-12-02
impl(common): more consistent HTTP clients

coryan:
For external accounts we will need HTTP clients to exchange subject tokens into access tokens.  Some (most?) subject token sources also need HTTP clients. I think it will be simpler if the subject token sources receive a `RestClient` factory when asked to fetch a subject token, as opposed to binding a factory to the time when the source is created.

The downside is an extra (unused) parameter in the subject token sources that do not use HTTP.

Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10349)


scotthart (approved):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10349)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #10356, opened 2022-12-02
impl: simplify `oauth2_internal::ComputeEngineCredentials`

coryan:
The implementation of `ComputeEngineCredentials` queries the metadata service to support self-signed URLs in `google/cloud/storage`. This query only needs to be done once, once the data is successfully retrieved it is not going to change as it is immutable in the GCE instance.

It also makes no sense to fail the token creation if the metadata query fails.

Part of the work for #10316

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10356)

### DIFF
--- a/google/cloud/internal/oauth2_compute_engine_credentials_test.cc
+++ b/google/cloud/internal/oauth2_compute_engine_credentials_test.cc
@@ -254,7 +254,7 @@ TEST(ComputeEngineCredentialsTest, FailedRetrieveServiceAccountInfo) {
   EXPECT_EQ(actual, alias);
   // Response 2
   actual = credentials.AccountEmail();
-  EXPECT_THAT(actual, alias);
+  EXPECT_EQ(actual, alias);
 }
 
 /// @test Mock a failed refresh response.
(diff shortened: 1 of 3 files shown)


google/cloud/internal/oauth2_compute_engine_credentials_test.cc
```diff
@@ -220,220 +224,145 @@ TEST_F(ComputeEngineCredentialsTest, ParseMetadataServerResponse) {
...
   // Response 1
-  auto status = credentials.AuthorizationHeader();
-  EXPECT_THAT(status, Not(IsOk()));
+  auto actual = credentials.AccountEmail();
+  EXPECT_EQ(actual, alias);
   // Response 2
-  status = credentials.AuthorizationHeader();
-  EXPECT_THAT(status, Not(IsOk()));
+  actual = credentials.AccountEmail();
+  EXPECT_THAT(actual, alias);
```
dbolduc, line 257: nit: s/THAT/EQ/

dbolduc (approved):

google/cloud/internal/oauth2_compute_engine_credentials_test.cc
coryan: Done

PR #10359, opened 2022-12-03
ci: check-api skips gRPC stubs

dbolduc:
This will unblock future googleapis SHA updates.

Here is an example of a failed `check-api` build:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/b815c373-d930-44db-bc1d-fe5443e66735;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/10358/6c38859de975e74e9a540fca8a3706d699d2339e/fedora-36-cmake-check-api/log-b815c373-d930-44db-bc1d-fe5443e66735.txt

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10359)

### DIFF
--- a/ci/cloudbuild/builds/check-api.sh
+++ b/ci/cloudbuild/builds/check-api.sh
@@ -114,6 +114,12 @@ function check_abi() {
       # by some other number indicating the length of the symbol within the
       # "internal" namespace. See: https://en.wikipedia.org/wiki/Name_mangling
       -skip-internal-symbols "(8internal|_internal)\d"
+      # We also ignore the raw gRPC Stub class. The generated gRPC headers that
+      # contain these classes are installed alongside our headers. When a new
+      # RPC is added to a service, these classes gain a pure virtual method. Our
+      # customers do not use these classes directly, so this should not
+      # constitute a breaking change.
+      -skip-internal-types "::StubInterface"
       # The library to compare
       -l "${library}"
       # Compared the saved baseline vs. the dump for the current version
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/builds/check-api.sh
```diff
@@ -114,6 +114,12 @@ function check_abi() {
       # by some other number indicating the length of the symbol within the
       # "internal" namespace. See: https://en.wikipedia.org/wiki/Name_mangling
       -skip-internal-symbols "(8internal|_internal)\d"
+      # We also ignore the raw gRPC Stub class. The generated gRPC headers that
+      # contain these classes are installed alongside our headers. When a new
+      # RPC is added to a service, these classes gain a pure virtual method. Our
+      # customers do not use these classes directly, so this should not
+      # constitute a breaking change.
+      -skip-internal-types "::StubInterface"
```
coryan, line 122: This is an unmangled symbol, but based on the previous comment we should be skipping mangled symbols.  Which one is the "Right Thing"[tm]?

dbolduc: I think either will work.
https://github.com/lvc/abi-compliance-checker/blob/8caa4417e5ccbd1cbf6add727842e205d6568cb4/abi-compliance-checker.pl#L620-L626

`-skip-internal-types` is easier to read. The reason I did not move the `"(8internal|_internal)\d"` business into types, as `"(::internal::|_internal::)"` is because I did not feel like testing it 😬

coryan (approved):

ci/cloudbuild/builds/check-api.sh
dbolduc: For future record, this is from the logs linked in the PR description: 
image

PR #10369, opened 2022-12-05
fix(storage): scopes should disable self-signed JWTs

coryan:
The storage service does not support self-signed JWTs with scopes. With this change, self-signed JWTs will be automatically disabled when using the legacy `storage::oauth2::Credentials`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10369)

### DIFF
--- a/google/cloud/internal/oauth2_service_account_credentials.cc
+++ b/google/cloud/internal/oauth2_service_account_credentials.cc
@@ -80,8 +80,9 @@ StatusOr<ServiceAccountCredentialsInfo> ParseServiceAccountCredentials(
       // "token_uri" attribute in the JSON object.  In this case, we try using
       // the default value.
       credentials.value(token_uri_key, default_token_uri),
-      /*scopes*/ {},
-      /*subject*/ {}};
+      /*scopes=*/absl::nullopt,
+      /*subject=*/absl::nullopt,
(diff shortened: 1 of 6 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10369)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/storage/oauth2/service_account_credentials.cc` line 137 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10369#-NIZAeCI2kbFws6xe0_T:-NIZAeCI2kbFws6xe0_U:bxf0q4w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c8afd89ca651de715cd4843308022bfc3be9e600/google/cloud/storage/oauth2/service_account_credentials.cc#L137)):*
> ```cpp
>   // Storage has more stringent requirements w.r.t. self-signed JWTs
>   // than most services (which the base class
>   // Disable them in the implementation class
> ```

Either some formatting or wording is off in this comment block.



_Code quote:_
```cpp
  // Storage has more stringent requirements w.r.t. self-signed JWTs
  // than most services (which the base class
  // Disable them in the implementation class
```

PR #10370, opened 2022-12-05
refactor: success predicates for `rest_internal::RestResponse`

coryan:
We often need to determine if a `RestResponse` contains a successful HTTP response or an error.  These helpers make the code easier to read and provide more consistent treatment of `RestResponse`. Callers that deviate from the standard success/failure codes can write their own functions (e.g., GCS treats 308 as "success" in some RPCs).

Part of the work for #5915 

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10370)

### DIFF
--- a/google/cloud/internal/rest_response.cc
+++ b/google/cloud/internal/rest_response.cc
@@ -139,7 +139,7 @@ StatusCode MapHttpCodeToStatus(std::int32_t code) {
 bool IsHttpSuccess(RestResponse const& response) {
   static_assert(HttpStatusCode::kMinSuccess < HttpStatusCode::kMinNotSuccess,
                 "Invalid HTTP code success range");
-  return response.StatusCode() <= HttpStatusCode::kMinNotSuccess &&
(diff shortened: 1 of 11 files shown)


google/cloud/internal/rest_response.cc
```diff
@@ -136,6 +136,17 @@ StatusCode MapHttpCodeToStatus(std::int32_t code) {
   return StatusCode::kUnknown;
 }
 
+bool IsHttpSuccess(RestResponse const& response) {
+  static_assert(HttpStatusCode::kMinSuccess < HttpStatusCode::kMinNotSuccess,
+                "Invalid HTTP code success range");
+  return response.StatusCode() <= HttpStatusCode::kMinNotSuccess &&
```
devbww, line 142: It feels like either `kMinNotSuccess` is lying, or that should be `<'.  That is, "== kMinNotSuccess" shouldn't be a success.  (??)

coryan: Good catch, fixed.

PR #10377, opened 2022-12-05
cleanup(ci): feature lists

dbolduc:
@coryan said about the `clang-tidy` build: "we should consider building _all_ features here."

I am not totally convinced (the build would be slower, and `cmake-install` does that job already). If asked, I can update this PR to have `clang-tidy` build all libraries. ( by using `full_list_cmake` instead of `always_build_cmake`).

Also, `ci/etc/full_feature_list` is `libraries`. so there is no need to include the `always_build` features twice. <details><summary>And...</summary>I am going to add `experimental-opentelemetry` to the `always_build` list in a future PR.</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10377)


ci/cloudbuild/builds/lib/features.sh
```diff
@@ -32,7 +32,7 @@ function features::always_build() {
     pubsub
```
coryan, line 32: You seem to have changed the permissions for this file.  Was that intentional?  I think the previous permissions were fine, and arguably "safer"?

coryan (approved):

ci/cloudbuild/builds/lib/features.sh
dbolduc: d'oh, unintentional. Good catch. Fixed.

PR #10389, opened 2022-12-07
fix(pubsublite): update retryable error codes for Admin and TopicStats clients

tmdiep:
For consistency with Pub/Sub Lite clients in other languages. For example, Java:
https://github.com/googleapis/java-pubsublite/blob/c4a9ff8e42f188e648d1bbf39333cd41615b1fa1/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/v1/stub/AdminServiceStubSettings.java#L745

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10389)

### DIFF
--- a/generator/generator_config.textproto
+++ b/generator/generator_config.textproto
@@ -1145,7 +1145,7 @@ service {
   product_path: "google/cloud/pubsublite"
   initial_copyright_year: "2021"
   gen_async_rpcs: ["GetTopicPartitions"]
-  retryable_status_codes: ["kInternal", "kUnavailable"]
+  retryable_status_codes: ["kAborted", "kInternal", "kUnavailable", "kUnknown"]
 }
 
 service {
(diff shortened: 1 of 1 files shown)


coryan:
Thanks for the PR, but I am shocked to see that `kAborted` and `kUnknown` are used as retryable codes:

- `kAborted` is explicitly documented as "retryable at another level" (for example via an OCC loop).
- `kUnknown` has no clear semantics.  Why would that be considered retryable?

https://grpc.github.io/grpc/core/md_doc_statuscodes.html

If anything I would suggest that Java is doing this wrong and they need to fix their code.

tmdiep:
I'll discuss this with the team. While `kAborted` and `kUnknown` are definitely retryable codes in Pub/Sub Lite's bidi stream protocol, this may not be true for the unary RPCs of these two services.

PR #10391, opened 2022-12-07
impl: support external accounts mapping

coryan:
Create the class derived from `google::cloud::Credentials` that will represent external accounts. These classes are mapped to the implementation using `google::cloud::internal::CredentialsVisitor`, which now supports external accounts. I change the integration test for external accounts to use the credentials to make a GCS request.

Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10391)

### DIFF
--- a/google/cloud/internal/credentials_impl.cc
+++ b/google/cloud/internal/credentials_impl.cc
@@ -84,6 +84,9 @@ ServiceAccountConfig::ServiceAccountConfig(std::string json_object,
     : json_object_(std::move(json_object)),
       options_(PopulateAuthOptions(std::move(opts))) {}
 
+ExternalAccountConfig::ExternalAccountConfig(std::string j, Options o)
+    : json_object_(std::move(j)), options_(PopulateAuthOptions(std::move(o))) {}
+
 }  // namespace internal
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace cloud
--- a/google/cloud/internal/credentials_impl.cc
+++ b/google/cloud/internal/credentials_impl.cc
@@ -84,8 +84,10 @@ ServiceAccountConfig::ServiceAccountConfig(std::string json_object,
     : json_object_(std::move(json_object)),
       options_(PopulateAuthOptions(std::move(opts))) {}
 
-ExternalAccountConfig::ExternalAccountConfig(std::string j, Options o)
-    : json_object_(std::move(j)), options_(PopulateAuthOptions(std::move(o))) {}
+ExternalAccountConfig::ExternalAccountConfig(std::string json_object,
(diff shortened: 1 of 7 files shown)


google/cloud/internal/unified_grpc_credentials.cc
```diff
@@ -19,10 +19,30 @@
 #include "google/cloud/internal/grpc_channel_credentials_authentication.h"
 #include "google/cloud/internal/grpc_impersonate_service_account.h"
 #include "google/cloud/internal/grpc_service_account_authentication.h"
-#include "google/cloud/internal/time_utils.h"
 #include "absl/memory/memory.h"
+#include <grpcpp/security/credentials.h>
```
dbolduc, line 23: are we relying on the `grpc::experimental` namespace being included by the grpc header? should we define it ourselves, just in case?
 
```cc
namespace grpc::experimental {}
```

google/cloud/internal/credentials_impl.cc
```diff
@@ -84,6 +84,9 @@ ServiceAccountConfig::ServiceAccountConfig(std::string json_object,
     : json_object_(std::move(json_object)),
       options_(PopulateAuthOptions(std::move(opts))) {}
 
+ExternalAccountConfig::ExternalAccountConfig(std::string j, Options o)
+    : json_object_(std::move(j)), options_(PopulateAuthOptions(std::move(o))) {}
```
dbolduc, line 88: nit: I applaud you for saving a line, but `j -> json_object`, `o -> opts` would be more informative and more consistent with the method signature for `ServiceAccountConfig`.

google/cloud/internal/unified_grpc_credentials.cc
coryan: No.  This was just because I wanted to preempt any IWYU questions.  We should not be reopening "other" namespaces, and there are plenty of `grpc::experimental` things

google/cloud/internal/credentials_impl.cc
coryan: Fixed

google/cloud/internal/unified_grpc_credentials.cc
dbolduc: ok, I just worry that we wouldn't control whether `using namespace grpc::experimental;` breaks one day, because the namespace is not defined. :shrug:

dbolduc (approved):

google/cloud/internal/unified_grpc_credentials.cc
coryan: I expect we will get a PR to update gRPC and detect the breakage early, before our customers with high likelihood.

PR #10392, opened 2022-12-07
impl: propagate `g::c::Options` through credentials

coryan:
Most credentials consume a `g::c::Options`.  In some places we were not propagating these options from the `g::c::Make*Credentials()` call all the way to the implementation class.

This change fixes that problem and starts the support for `g::c::Options` in all credentials. The last bit would be useful to enable logging when needed.

Part of the work for #10316

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10392)


devbww (approved):

google/cloud/internal/credentials_impl.cc
```diff
@@ -13,25 +13,76 @@
...
+  // First set any defaults that may be missing.
+  options =
+      MergeOptions(std::move(options),
+                   Options{}
+                       .set<ScopesOption>({kDefaultScope})
+                       .set<AccessTokenLifetimeOption>(kDefaultTokenLifetime));
+  // Then apply any overrides.
+  return MergeOptions(
+      Options{}.set<TracingComponentsOption>(DefaultTracingComponents()),
+      std::move(options));
```
dbolduc, line 41: why are these overrides?

coryan: I think because https://github.com/googleapis/google-cloud-cpp/blob/main/doc/adr/2022-11-16-option-precedence-hierarchy.md `DefaultTracingComponents()` reads the environment variable.

dbolduc: If no environment variables are set, it overrides the customer supplied options.

Maybe we should be defaulting `options` before they get here.

Issue #3429, opened 2020-03-04
Add logging to the credential refresh RPCs


coryan:
We need a way to log the credential refresh RPCs, without it, it is really hard to troubleshoot problems with credentials. cf #3426 .

kaustubhpatil2611:
Hello @coryan ,
I would like to take up this issue! Can i work on it?

coryan:
Certainly, feel free to send us a PR. Remember to sign the CLA so we can accept your contribution.

coryan:
Seems like this will be fixed as part of the migration to the `rest-internal` library.

coryan:
@scotthart is working on it

PR #10408, opened 2022-12-08
impl(common): decorator to log REST credentials

coryan:
This will be used in the initialization of
`oauth2_internal::Credentials`.  I am planning to log any usage of cached credentials, and also any refreshes.  Using `ComputeEngineCredentials` as an example, and with all types implicitly in the `oauth2_internal` namespace, the full typical stack would be:

```
LoggingCredentials(cached) -> CachedCredentials ->
     LoggingCredentials(refresh) -> ComputeEngineCredentials
```

The logging elements are optional. They will be inserted only if the `g::c::TracingComponentsOption` includes `auth`.

Part of the work for #10316 and for #3429

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10408)


google/cloud/internal/oauth2_logging_credentials.cc
```diff
@@ -0,0 +1,80 @@
...
+                   << token.status();
+    return token;
+  }
+  if (now <= token->expiration) {
+    GCP_LOG(DEBUG) << __func__ << "(" << phase_ << "), token=" << *token
+                   << ", token will expire in "
+                   << absl::FormatDuration(
+                          absl::FromChrono(token->expiration - now));
+    return token;
+  }
```
devbww, line 48: Nit: I'd make this "good" case the final one by inverting the test.

google/cloud/internal/oauth2_logging_credentials.h
```diff
@@ -0,0 +1,67 @@
...
+namespace oauth2_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * Log credentials usage and refreshes.
+ *
+ * Debugging problems with authentication can be fairly difficult. This
+ * decorator is automatically inserted (twice) if
+ * `google::cloud::TracingComponentsOption` includes `auth`. The decorator is
+ * inserted before and after the caching layer, to show whether an cached token
```
devbww, line 33: s/an cached/a cached/

google/cloud/internal/oauth2_logging_credentials.h
```diff
@@ -0,0 +1,67 @@
...
+ * @note To prevent leaking authentication secrets, the tokens are not logged in
+ * full.
+ *
+ * @see https://cloud.google.com/docs/authentication/ for an overview of
+ * authenticating to Google Cloud Platform APIs.
+ */
+class LoggingCredentials : public Credentials {
+ public:
+  LoggingCredentials(std::string phase, TracingOptions tracing_options,
+                     std::shared_ptr<Credentials> impl);
```
devbww, line 45: `#include <memory>`

google/cloud/internal/oauth2_logging_credentials.h
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @note To prevent leaking authentication secrets, the tokens are not logged in
+ * full.
+ *
+ * @see https://cloud.google.com/docs/authentication/ for an overview of
+ * authenticating to Google Cloud Platform APIs.
+ */
+class LoggingCredentials : public Credentials {
+ public:
+  LoggingCredentials(std::string phase, TracingOptions tracing_options,
```
devbww, line 44: `#include <string>`

google/cloud/internal/oauth2_logging_credentials.h
```diff
@@ -0,0 +1,67 @@
...
+  StatusOr<internal::AccessToken> GetToken(
+      std::chrono::system_clock::time_point now) override;
+  StatusOr<std::vector<std::uint8_t>> SignBlob(
+      absl::optional<std::string> const& signing_service_account,
+      std::string const& string_to_sign) const override;
+  std::string AccountEmail() const override;
+  std::string KeyId() const override;
+
+ private:
+  std::string const phase_;
```
devbww, line 57: Nit: We usually think (I think) the `const` isn't worth sacrificing copyability (or, if you actually don't want copyability, to do that explicitly).

coryan: Fixed

google/cloud/internal/oauth2_logging_credentials.h
coryan: Fixed

google/cloud/internal/oauth2_logging_credentials.cc
coryan: Done

google/cloud/internal/oauth2_logging_credentials.h
coryan: Fixed, thanks

google/cloud/internal/oauth2_logging_credentials.h
coryan: Fixed

devbww (approved):

PR #10412, opened 2022-12-09
feat: support logging for unified Rest credentials

coryan:


google/cloud/internal/oauth2_decorate_credentials.h
```diff
@@ -25,6 +26,16 @@ namespace cloud {
 namespace oauth2_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+/// Add a full stack of logging (if needed) and caching decorators to the
+/// credentials.
+std::shared_ptr<oauth2_internal::Credentials> Decorate(
+    std::shared_ptr<oauth2_internal::Credentials> impl, Options const& opts);
+
+/// Add only a logging decorator to the credentials.
```
devbww, line 34: Perhaps this could also do with the "if needed" restriction.

"If needed" might also be expanded to something like "if requested by the `Options`" in both cases.

devbww (approved):

PR #10416, opened 2022-12-09
impl(oauth2): use client factory for impersonation credentials

coryan:
This will make things easier to test, I am thinking of `oauth2_internal::MapCredentials()` or `oauth2_internal::GoogleDefaultCredentials()`, which are not easy to mock at this time.

Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10416)

### DIFF
--- a/google/cloud/internal/oauth2_minimal_iam_credentials_rest_test.cc
+++ b/google/cloud/internal/oauth2_minimal_iam_credentials_rest_test.cc
@@ -209,17 +209,11 @@ TEST(MinimalIamCredentialsRestTest, GenerateAccessTokenSuccess) {
   std::string scope = "my_scope";
   std::string delegate = "my_delegate";
   std::string response = R"""({
-  "accessToken": "my_access_token",
-  "expireTime": "2022-10-12T07:20:50.52Z"
-})""";
-
-  auto mock_credentials = std::make_shared<MockCredentials>();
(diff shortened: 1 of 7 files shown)


google/cloud/internal/oauth2_minimal_iam_credentials_rest_test.cc
```diff
@@ -210,46 +203,56 @@ TEST(ParseGenerateAccessTokenResponse, InvalidExpireTimeFormat) {
...
-  "expireTime": "2022-10-12T07:20:50.52Z"
-})""";
-
-  auto* mock_response = new MockRestResponse();
-  EXPECT_CALL(*mock_response, StatusCode)
-      .WillRepeatedly(Return(rest_internal::HttpStatusCode::kOk));
-  EXPECT_CALL(std::move(*mock_response), ExtractPayload).WillOnce([&] {
-    return testing_util::MakeMockHttpPayloadSuccess(response);
+    "accessToken": "my_access_token",
+    "expireTime": "2022-10-12T07:20:50.52Z"})""";
```
devbww, line 213: Nit: Do we like the extra newline that used to appear in `response`?

coryan: I think the test is more readable as-is. And the newline  makes no difference when parsing the JSON object.

PR #10417, opened 2022-12-09
feat(common): options for `Make*Credentials()`

coryan:
This changes all the `g::c::Credential` factory functions to accept a `g::c::Options` parameter. At this time this can be used to initialize logging in the credential layer, but eventually it could gain

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10417)


google/cloud/credentials.h
```diff
@@ -82,8 +83,12 @@ struct UnifiedCredentialsOption {
  * and sometimes even errors.
  *
  * @ingroup guac
+ *
+ * @param opts optional configuration values.  Note that the effect of these
+ *     parameters depends on the underlying transport.  For example
```
devbww, line 88: s/$/,/ et seq.

devbww (approved):
Description: s/eventually it could gain$/eventually it could gain <something>./

google/cloud/credentials.h
coryan: Thanks. Fixed.

PR #10421, opened 2022-12-10
fix: add missing `<cstdint>` includes

trofi:
Without the change build fails on this week's `gcc-13` as:

    In file included from google/cloud/internal/curl_handle.cc:15:
    google/cloud/internal/curl_handle.h:107:8:
      error: 'int32_t' in namespace 'std' does not name a type
      107 |   std::int32_t GetResponseCode();
          |        ^~~~~~~

`gcc-13` cleaned up `<string>` not to include `<cstdint>` for newer c++ standards. Hence the build failures.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10421)

### DIFF
--- a/google/cloud/internal/curl_handle.h
+++ b/google/cloud/internal/curl_handle.h
@@ -20,6 +20,7 @@
 #include "google/cloud/version.h"
 #include "absl/functional/function_ref.h"
 #include <curl/curl.h>
+#include <cstdint>
 #include <string>
 #include <type_traits>
 
--- a/google/cloud/internal/oauth2_credentials.h
+++ b/google/cloud/internal/oauth2_credentials.h
@@ -20,6 +20,7 @@
 #include "google/cloud/status_or.h"
 #include "google/cloud/version.h"
 #include <chrono>
+#include <cstdint>
 #include <string>
 #include <vector>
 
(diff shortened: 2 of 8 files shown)


coryan:
Thanks for the PR!  Seems like the robots think you need to sort some headers. The build logs are listed above, but they are tricky to find, so here is a direct link:

https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/10421/081386c1fce4a888b188fcc89cb5d0f7a1234da5/checkers-checkers/log-06f36380-a1f3-4b44-8ea2-f1342a8a712a.txt

Are you using a particular Linux distro to build with GCC 13?  I would like to expand our build matrix to cover this case, and I would also like to avoid installing GCC from source.

trofi:
I'm afraid I'm using locally built `gcc-13`. Other people used ubuntu's ppa with weekly gcc snapshots, but I'm not familiar with precise locations of those. I wonder if `include-what-you-use` could provide a similar signal for missing headers.

Applied the following header sorting and force-pushed:

```
Step #3: -#include <cstdint>
Step #3:  #include <chrono>
Step #3: +#include <cstdint>
```

Thank you!

trofi:
Fixed commit message to match PR (now has `fix: ...` prefix)

coryan (approved):

PR #10431, opened 2022-12-13
cleanup: shorter names for CMake targets

coryan:
The CMake targets need to be unique across the whole project. That is trivial for libraries, but tests often have conflicting names. A `foo_test.cc` will typically result in a target name of `foo` regardless of what directory it is in. Note that only the target name needs to be globally unique, the output file (`foo_test`) is already in the same directory as `foo_test.cc` and not likely to conflict.

In some places we used something inspired by the path (relative to the project root or to `google/cloud`) to prefix the target name. In other places we used something inspired by the associated library. I think it is simpler to always use the path. And it also yields shorter target.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10431)

### DIFF
--- a/.github/workflows/external-account-integration.yml
+++ b/.github/workflows/external-account-integration.yml
@@ -59,7 +59,7 @@ jobs:
       - name: build
         run: |
           cmake --build "${{runner.temp}}/build" -j 2 \
-            --target rest_internal_internal_external_account_integration_test
+            --target common_internal_external_account_integration_test
 
       # Configure Workload Identity Federation and generate an access token.
       - id: 'auth'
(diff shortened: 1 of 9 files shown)


.github/workflows/external-account-integration.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: build
         run: |
           cmake --build "${{runner.temp}}/build" -j 2 \
-            --target rest_internal_internal_external_account_integration_test
+            --target common_internal_external_account_integration_test
```
devbww, line 62: "common" between "internal" and "external"  :-)

devbww (approved):

.github/workflows/external-account-integration.yml
coryan: Better than `internal_internal_external_` ? :shrug:  I think I would like to remove the `common_` prefix in a future PR.

PR #10434, opened 2022-12-13
impl(oauth2): support IMDSv2 in AWS token sources

coryan:
This is the easiest configuration to implement for AWS token sources.

Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10434)


coryan:
Sorry, I just realized I misread the documentation.  The IDMSv2 case is not the easy case, it is the most complicated case. Sigh... I need to close this.

PR #10439, opened 2022-12-14
impl(oauth2): building blocks for AWS external accounts

coryan:
As with the other external accounts, we need to build a subject token for AWS external accounts. Building this subject token requires obtaining a number of pieces of information. These can be found on environment variables or via the VM's metadata service. The functions to get them are complex enough that they deserve their own tests.

Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10439)

### DIFF
--- a/google/cloud/internal/external_account_token_source_aws.cc
+++ b/google/cloud/internal/external_account_token_source_aws.cc
@@ -19,6 +19,7 @@
 #include "google/cloud/internal/json_parsing.h"
 #include "google/cloud/internal/make_status.h"
 #include "absl/strings/match.h"
+#include <chrono>
 
 namespace google {
 namespace cloud {
@@ -89,7 +90,7 @@ StatusOr<ExternalAccountTokenSourceAwsInfo> ParseExternalAccountTokenSourceAws(
             *environment_id,
             ". Consider updating `google-cloud-cpp`, as a new version may",
             " support this environment. If you find this is not the case,",
-            " please file a feature at"
+            " please file a feature request at"
             " https://github.com/googleapis/google-cloud-cpp/issues"),
         GCP_ERROR_INFO().WithContext(ec));
   }
@@ -120,12 +121,13 @@ StatusOr<std::string> FetchMetadataToken(
     HttpClientFactory const& client_factory, Options const& opts,
     internal::ErrorContext const& /*ec*/) {
   if (info.imdsv2_session_token_url.empty()) return std::string{};
+  auto const ttl =
+      std::chrono::duration_cast<std::chrono::seconds>(kDefaultMetadataTokenTtl)
+          .count();
+  auto request = rest_internal::RestRequest()
+                     .SetPath(std::move(info.imdsv2_session_token_url))
+                     .AddHeader(kMetadataTokenTtlHeader, std::to_string(ttl));
   auto client = client_factory(opts);
-  auto request =
-      rest_internal::RestRequest()
-          .SetPath(std::move(info.imdsv2_session_token_url))
-          .AddHeader(kMetadataTokenTtlHeader,
-                     std::to_string(kDefaultMetadataTokenTtl.count()));
   auto response = client->Put(request, {});
   if (!response) return std::move(response).status();
   if (IsHttpError(**response)) return AsStatus(std::move(**response));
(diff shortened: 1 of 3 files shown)


google/cloud/internal/external_account_token_source_aws.cc
```diff
@@ -95,6 +115,94 @@ StatusOr<ExternalAccountTokenSourceAwsInfo> ParseExternalAccountTokenSourceAws(
...
+    ExternalAccountTokenSourceAwsInfo const& info,
+    HttpClientFactory const& client_factory, Options const& opts,
+    internal::ErrorContext const& /*ec*/) {
+  if (info.imdsv2_session_token_url.empty()) return std::string{};
+  auto client = client_factory(opts);
+  auto request =
+      rest_internal::RestRequest()
+          .SetPath(std::move(info.imdsv2_session_token_url))
+          .AddHeader(kMetadataTokenTtlHeader,
+                     std::to_string(kDefaultMetadataTokenTtl.count()));
```
devbww, line 128: Nit: In order to avoid depending on the exact `duration` type of a named object in the places where we need to convert it to a particular variety or count, I would always suggest something like
```
  std::chrono::duration_cast<std::chrono::seconds>(kDefaultMetadataTokenTtl).count()
```
In this case a glorified no-op, but it makes what is going on clearer I reckon (the alternative being putting units in the name, which feels like going in the wrong direction).

google/cloud/internal/external_account_token_source_aws.cc
```diff
@@ -95,6 +115,94 @@ StatusOr<ExternalAccountTokenSourceAwsInfo> ParseExternalAccountTokenSourceAws(
...
+  }
+
+  auto payload = GetMetadata(info.region_url, metadata_token, cf, opts);
+  if (!payload) return std::move(payload).status();
+  if (payload->empty()) {
+    return InvalidArgumentError(
+        absl::StrCat("invalid (empty) region returned from ", info.region_url),
+        GCP_ERROR_INFO().WithContext(ec));
+  }
+  // The metadata service returns an availability zone, we must remove the last
```
devbww, line 152: s/zone, we/zone, so we/, or "but we", or s/zone, we/zone. We/.

google/cloud/internal/external_account_token_source_aws.cc
```diff
@@ -95,6 +115,94 @@ StatusOr<ExternalAccountTokenSourceAwsInfo> ParseExternalAccountTokenSourceAws(
...
+  return *std::move(payload);
+}
+
+StatusOr<ExternalAccountTokenSourceAwsSecrets> FetchSecrets(
+    ExternalAccountTokenSourceAwsInfo const& info,
+    std::string const& metadata_token, HttpClientFactory const& cf,
+    Options const& opts, internal::ErrorContext const& ec) {
+  auto access_key_id_env = internal::GetEnv("AWS_ACCESS_KEY_ID");
+  auto secret_access_key_env = internal::GetEnv("AWS_SECRET_ACCESS_KEY");
+  auto session_token_env = internal::GetEnv("AWS_SESSION_TOKEN");
```
devbww, line 164: Nit: Move inside the `if`?

google/cloud/internal/external_account_token_source_aws.cc
```diff
@@ -95,6 +115,94 @@ StatusOr<ExternalAccountTokenSourceAwsInfo> ParseExternalAccountTokenSourceAws(
...
+        /*session_token=*/session_token_env.value_or("")};
+  }
+
+  // This code fetches the security credentials from the metadata services in an
+  // AWS EC2 instance, i.e., a VM. The requests and responses are documented in:
+  //  https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials
+  auto role = GetMetadata(info.url, metadata_token, cf, opts);
+  if (!role) return std::move(role).status();
+  auto path = info.url;
+  if (path.back() != '/') path.push_back('/');
```
devbww, line 178: Do we need check `path.empty()`?

google/cloud/internal/external_account_token_source_aws.cc
```diff
@@ -42,8 +43,27 @@ namespace {
 auto constexpr kDefaultUrl =
     "http://169.254.169.254/latest/meta-data/iam/security-credentials";
 
+auto constexpr kMetadataTokenTtlHeader = "X-aws-ec2-metadata-token-ttl-seconds";
+auto constexpr kDefaultMetadataTokenTtl = std::chrono::seconds(900);
```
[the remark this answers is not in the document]
coryan: Done

google/cloud/internal/external_account_token_source_aws.cc
coryan: I am going to leave this as-is for now. This will only be used with an `info` initialized with `ParseExternalAccountTokenSourceAws()` and I will be adding some code to prevent empty `url` values to that function.

devbww (approved):

PR #10441, opened 2022-12-14
cleanup(pubsub): simplify emulator workarounds

### LINKED ISSUE #4792: Cleanup workarounds for gaps in the Cloud Pub/Sub emulator
The Cloud Pub/Sub emulator lacks a number of features. This issue will help us document any workarounds setup for them.

- [ ] UpdateTopic() not implemented (b/162942530)
- [ ] UpdateSubscription() does not support labels (b/162942530)
- [ ] CreateSnapshot() does not support server-side assigned names (b/163124116)
- [ ] UpdateSnapshot() is not supported (b/163146286)
- [ ] DetachSubscription() is not supported (b/163313381)

coryan:
Just throw from the samples and have a little wrapper capture and ignore the right exceptions when running with the emulator.

Fixes #4792 - or at least we will not be worrying about unimplemented features in the emulator after this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10441)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -163,7 +163,7 @@ void UpdateTopic(google::cloud::pubsub::TopicAdminClient client,
         pubsub::TopicBuilder(
             pubsub::Topic(std::move(project_id), std::move(topic_id)))
             .add_label("test-key", "test-value"));
-    if (!topic) return;  // TODO(#4792) - emulator lacks UpdateTopic()
+    if (!topic) throw std::move(topic).status();
 
     std::cout << "The topic was successfully updated: " << topic->DebugString()
               << "\n";
@@ -220,7 +220,7 @@ void DetachSubscription(google::cloud::pubsub::TopicAdminClient client,
      std::string subscription_id) {
     auto response = client.DetachSubscription(pubsub::Subscription(
         std::move(project_id), std::move(subscription_id)));
-    if (!response.ok()) return;  // TODO(#4792) - not implemented in emulator
+    if (!response.ok()) throw std::move(response).status();
 
     std::cout << "The subscription was successfully detached: "
               << response->DebugString() << "\n";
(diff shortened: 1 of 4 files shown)


google/cloud/pubsub/samples/samples.cc
```diff
@@ -2069,6 +2065,18 @@ void AutoRun(std::vector<std::string> const& argv) {
 
   auto const snapshot_id = RandomSnapshotId(generator);
 
+  auto ignore_emulator_failures = [](auto lambda) {
+    using ::google::cloud::StatusCode;
+    try {
+      lambda();
+    } catch (google::cloud::Status const& s) {
+      if (UsingEmulator() && s.code() == StatusCode::kUnimplemented) return;
```
devbww, line 2073: Re `return` ... Might not an "unimplemented" error cascade to a non-"unimplemented" error?

coryan: You mean like a `CreateFoo()` fails and then the next `UpdateFoo()` fails but not with `kUnimplemented`?
In this PR the existing case (in `AutoRunProtobuf()`), which already had a `if (UsingEmulator()) return;`, regardless of failures.  The PR makes things slightly less tested (we used to test 3 functions that all failed), but I can live with that.  Of course that could happen in the future, when that happens I expect that:

- When somebody introduces a PR that triggers this behavior the build will fail and they will need to do something (e.g., group all the related functions in a `AutoRunFoo()` helper so the first failure exits the function, and we just wrap the `AutoRunFoo()` call)
- If we ever changed the emulator to a new version that is also a PR and we will notice in at least some of the builds

devbww: > You mean like a `CreateFoo()` fails and then the next `UpdateFoo()` fails but not with `kUnimplemented`?

Yes.

> When somebody introduces a PR that triggers this behavior the build will fail and they will need to do something

OK.

devbww (approved):

PR #10442, opened 2022-12-14
impl(oauth2): helper functions to compute `HMAC_SHA256()`

coryan:
As part of the AWS external account support we need to generate a signed request:

> Construct the AWS signed request (AWS Signature Version 4) using the
> `GetCallerIdentity`

(from)
https://google.aip.dev/auth/4117#determining-the-subject-token-in-aws

These signed requests are documented in:

https://docs.aws.amazon.com/general/latest/gr/create-signed-request.html

They require computing multiple `HMAC_SHA256()` values. The implementation is simple if you already have a dependency on OpenSSL (as we do). A longer description of these can be found in Wikipedia:

https://en.wikipedia.org/wiki/HMAC

Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10442)


google/cloud/internal/sha256_hmac.h
```diff
@@ -0,0 +1,59 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_SHA256_HMAC_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_SHA256_HMAC_H
+
+#include "google/cloud/internal/sha256_type.h"
+#include "google/cloud/version.h"
+#include "absl/types/span.h"
+#include <array>
```
devbww, line 21: Unused.

google/cloud/internal/sha256_hmac.h
```diff
@@ -0,0 +1,59 @@
...
+#include <cstdint>
+#include <string>
+#include <vector>
```
devbww, line 24: Unused.

google/cloud/internal/sha256_hmac_test.cc
```diff
@@ -0,0 +1,69 @@
...
+  auto data = std::string{"The quick brown fox jumps over the lazy dog"};
+  // Compare against the HEX-encoded value just because they are easier to read.
+  auto const expected = std::string{
+      "f7bc83f430538424b13298e6aa6fb143ef4d59a14946175997479dbc2d1a3cd8"};
+  EXPECT_EQ(expected, HexEncode(Sha256Hmac("key", data)));
+  EXPECT_EQ(expected, HexEncode(Sha256Hmac("key", absl::Span<char>(data))));
+  EXPECT_EQ(expected,
+            HexEncode(Sha256Hmac("key", absl::Span<char const>(data))));
+  std::vector<std::uint8_t> v(data.begin(), data.end());
+  EXPECT_EQ(expected,
```
devbww, line 39: `EXPECT_EQ(expected, HexEncode(Sha256Hmac("key", v)));` too, like for `data`?

google/cloud/internal/sha256_hmac_test.cc
```diff
@@ -0,0 +1,69 @@
...
+  auto data = std::string{"The quick brown fox jumps over the lazy dog"};
+  // Compare against the HEX-encoded value just because they are easier to read.
+  auto const expected = std::string{
+      "f61f354e69e4e35dd936913f098993f4a07254ef9c156e26842f07e998d6e61e"};
+  auto key = Sha256Hmac("key", data);
+  EXPECT_EQ(expected, HexEncode(Sha256Hmac(key, data)));
+  EXPECT_EQ(expected, HexEncode(Sha256Hmac(key, absl::Span<char>(data))));
+  EXPECT_EQ(expected, HexEncode(Sha256Hmac(key, absl::Span<char const>(data))));
+  std::vector<std::uint8_t> v(data.begin(), data.end());
+  EXPECT_EQ(expected, HexEncode(Sha256Hmac(key, absl::Span<std::uint8_t>(v))));
```
devbww, line 60: Ditto.

google/cloud/internal/sha256_hmac.cc
```diff
@@ -0,0 +1,74 @@
...
+Sha256Type Sha256Hmac(std::string const& key, absl::Span<char const> data) {
+  return Sha256HmacImpl(absl::Span<char const>(key),
+                        reinterpret_cast<unsigned char const*>(data.data()),
+                        data.size());
+}
+
+Sha256Type Sha256Hmac(std::string const& key,
+                      absl::Span<std::uint8_t const> data) {
+  return Sha256HmacImpl(absl::Span<char const>(key),
+                        reinterpret_cast<unsigned char const*>(data.data()),
```
devbww, line 54: It is worth noting (in our heads, if nowhere else) that if `std::uint8_t` exists, as we assume, that it must be a synonym for `unsigned char`.  So, the cast is fine.

google/cloud/internal/sha256_hmac.cc
```diff
@@ -0,0 +1,74 @@
...
+namespace internal {
+namespace {
+
+static_assert(EVP_MAX_MD_SIZE >= Sha256Type().size(),
+              "EVP_MAX_MD_SIZE is too small");
+
+template <typename T>
+Sha256Type Sha256HmacImpl(absl::Span<T const> key, unsigned char const* data,
+                          std::size_t count) {
+  std::array<unsigned char, EVP_MAX_MD_SIZE> digest;
```
devbww, line 32: `#include <array>`

google/cloud/internal/sha256_hmac.h
coryan: Removed

google/cloud/internal/sha256_hmac.h
coryan: Removed

google/cloud/internal/sha256_hmac_test.cc
coryan: Done

google/cloud/internal/sha256_hmac_test.cc
coryan: Done

google/cloud/internal/sha256_hmac.cc
devbww: "When `std::uint8_t` exists it must be `unsigned char`" perhaps?

devbww (approved):

PR #10444, opened 2022-12-15
cleanup(spanner): do not provide stub functions for unused RPCs

devbww:
We do not provide a stub function for `Spanner.Read` because the client implementation doesn't use it.  So, do the same for `Spanner.GetSession` and `Spanner.ListSessions`, which are similarly unused.

This will make it clearer that `GetSession` (like `Read`) does not need to be considered when adding routing-header support in the future.

### DIFF
--- a/google/cloud/spanner/internal/logging_spanner_stub.cc
+++ b/google/cloud/spanner/internal/logging_spanner_stub.cc
@@ -60,29 +60,6 @@ LoggingSpannerStub::AsyncBatchCreateSessions(
       cq, std::move(context), request, __func__, tracing_options_);
 }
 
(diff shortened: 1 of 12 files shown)


coryan (approved):
Have you considered generating all these stubs?

devbww:
> Have you considered generating all these stubs?

Not so far, but I will.  Thanks.

devbww:
> > Have you considered generating all these stubs?
> 
> Not so far, but I will. Thanks.

This would, at least, require some new generator support.  Spanner data requests, for example, all have a "google-cloud-resource-prefix" header that names the database.

PR #10452, opened 2022-12-16
impl(oauth2): create AWS V4 signed request

coryan:
Create the AWS V4 signed request as a JSON object because it is easier to test. In the actual implementation this becomes a string, that is then URL-encoded (yikes!) and passed to Google's Security Token Service.

Part of the work for #5915

### DIFF
--- a/google/cloud/internal/external_account_token_source_aws.cc
+++ b/google/cloud/internal/external_account_token_source_aws.cc
@@ -14,11 +14,16 @@
 
 #include "google/cloud/internal/external_account_token_source_aws.h"
 #include "google/cloud/internal/absl_str_cat_quiet.h"
+#include "google/cloud/internal/absl_str_replace_quiet.h"
 #include "google/cloud/internal/external_account_source_format.h"
+#include "google/cloud/internal/format_time_point.h"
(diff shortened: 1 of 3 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10452)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/internal/external_account_token_source_aws.h` line 113 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10452#-NJO4TzR2YwS3Uv0IKGb:-NJO4TzS5K14EQcG-Kzp:b6myqcb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f4c68dcbb63ef2471784d3e33711932452b51f64/google/cloud/internal/external_account_token_source_aws.h#L113)):*
> ```objc
>     ExternalAccountTokenSourceAwsSecrets const& secrets,
>     std::chrono::system_clock::time_point now, std::string const& target,
>     bool debug);
> ```

Maybe default this to false?



_Code quote:_
```objc
bool debug
```

PR #10455, opened 2022-12-17
impl(rest): add RestCompletionQueue

scotthart:
This is part of the work to handle async operations libraries with REST transport. This particular class is typically used as a member of a type that implements CompletionQueueImpl. In current form, it's a simple type that could probably be merged into the forthcoming RestCompletionQueueImpl class, but it's necessary to keep it separate to manage dependencies. This class is the in the rest_internal library, which does not have dependencies on protobuf or grpc. RestCompletionQueueImpl will have to reside in rest_protobuf_internal library as that's where everything else completion queue related is located.
In the future, this class would leverage curl_multi_handle to manage multiple ongoing HTTP requests in single thread.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455)


google/cloud/internal/rest_completion_queue.h
```diff
@@ -0,0 +1,74 @@
...
+  // Prevents further retrieval from or mutation of the RestCompletionQueue.
+  void Shutdown();
+
+  // Attempts to get the next tag from the queue before the deadline is reached.
+  // If a tag is retrieved, tag is set to the retrieved value, ok is set true,
+  //   and kGotEvent is returned.
+  // If the deadline is reached and no tag is available, kTimeout is returned.
+  // If Shutdown has been called, tag is set to nullptr, ok is set to false,
+  //    and kShutdown is returned.
+  QueueStatus GetNext(void** tag, bool* ok,
```
devbww, line 57: `void*` has an odor.  Can we use something like `absl::any` instead to hide some of that?  Or should all these be templates so any casts can be internal?  Or ...?

google/cloud/internal/rest_completion_queue.h
```diff
@@ -0,0 +1,74 @@
...
+                      std::chrono::system_clock::time_point deadline);
+  void AddTag(void* tag);
+  void RemoveTag(void* tag);
+  std::size_t size() const;
```
devbww, line 61: I would avoid `size()` in a concurrent API.  Its value is obsolete immediately after you obtain it, so it is an attractive nuisance.

google/cloud/internal/rest_completion_queue.h
```diff
@@ -0,0 +1,74 @@
...
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_REST_COMPLETION_QUEUE_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_REST_COMPLETION_QUEUE_H
+
+#include "google/cloud/future.h"
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include "absl/memory/memory.h"
+#include <chrono>
+#include <deque>
+#include <memory>
```
devbww, line 24: Unused.

google/cloud/internal/rest_completion_queue.h
```diff
@@ -0,0 +1,74 @@
...
+#include <mutex>
+#include <thread>
```
devbww, line 26: Unused.

google/cloud/internal/rest_completion_queue.cc
```diff
@@ -0,0 +1,80 @@
...
+    void** tag, bool* ok, std::chrono::system_clock::time_point deadline) {
+  auto check_queue = [&]() mutable {
+    std::lock_guard<std::mutex> lock(mu_);
+    if (shutdown_) {
+      *tag = nullptr;
+      *ok = false;
+      return QueueStatus::kShutdown;
+    }
+
+    if (!pending_tags_.empty()) {
```
devbww, line 40: Nit: I'd reverse this check so that `kGotEvent`, the "good" case, is the fallthough case.

google/cloud/internal/rest_completion_queue.cc
```diff
@@ -0,0 +1,80 @@
...
+      auto* front = std::move(pending_tags_.front());
+      pending_tags_.pop_front();
+      *tag = front;
+      *ok = true;
+      return QueueStatus::kGotEvent;
+    }
+    return QueueStatus::kTimeout;
```
devbww, line 47: Nit: It is a little weird for this to say `kTimeout` when `check_queue()` has got nothing to do with time.

google/cloud/internal/rest_completion_queue.cc
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  // This is a naive implementation of using the deadline to return if no tags
+  // are present.
+  auto initial_check = check_queue();
+  if (initial_check != QueueStatus::kTimeout) return initial_check;
+  std::this_thread::sleep_until(deadline);
```
devbww, line 54: `sleep_until()` has an odor, the prime one being: don't you want this to unblock early if someone does an `AddTag()` in the meantime?

It will also make the API harder to test.

google/cloud/internal/rest_completion_queue_test.cc
```diff
@@ -0,0 +1,99 @@
...
+
+namespace google {
+namespace cloud {
+namespace rest_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace {
+
+TEST(RestCompletionQueue, AddTag) {
+  RestCompletionQueue cq;
+  int tag;
```
devbww, line 27: It is probably nicer to reverse such declarations so that the queue is destroyed before the objects it (kind of) contains.

google/cloud/internal/rest_completion_queue.cc
```diff
@@ -0,0 +1,80 @@
...
+#include <thread>
+
+namespace google {
+namespace cloud {
+namespace rest_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+void RestCompletionQueue::Shutdown() {
+  std::lock_guard<std::mutex> lock(mu_);
+  shutdown_ = true;
```
devbww, line 27: Empty `pending_tags_` actively?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455)* status: 0 of 6 files reviewed, 9 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/internal/rest_completion_queue.h` line 24 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5t-r2-24:-NJWZUgO9kfvmmFdjiql:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.h#L24)):*

Unused.
</blockquote></details>

Done.
___
*[`google/cloud/internal/rest_completion_queue.h` line 26 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5t-r2-26:-NJWZWbD85-t-9lyAdGn:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.h#L26)):*

Unused.
</blockquote></details>

Done.
___
*[`google/cloud/internal/rest_completion_queue.h` line 56 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5t-r2-56:-NJWaY7j9kN2f5Xma5BI:b-encs1w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.h#L56)):*

`ok` seems redundant.
</blockquote></details>

Perhaps. I'm modeling this class on the interface of grpc::CompletionQueue as that's what functionality it is replacing in a REST transport world. I would prefer to leave this out parameter for now, and if after everything is implemented and working if it can be removed, then we can do so.
___
*[`google/cloud/internal/rest_completion_queue.h` line 57 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5t-r2-57:-NJWZY7Q4JOLCvABHu6I:b2v9i9j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.h#L57)):*

`void*` has an odor.  Can we use something like `absl::any` instead to hide some of that?  Or should all these be templates so any casts can be internal?  Or ...?
</blockquote></details>

the void* is a key into a `std::unordered_map<void*, std::shared_ptr<AsyncRestOperation>>` in RestCompletionQueueImpl
___
*[`google/cloud/internal/rest_completion_queue.h` line 61 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5t-r2-61:-NJWZyOn4yrnq9wcLTff:b-1r1aob) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.h#L61)):*

I would avoid `size()` in a concurrent API.  Its value is obsolete immediately after you obtain it, so it is an attractive nuisance.
</blockquote></details>

I only added it for testing purposes.
___
*[`google/cloud/internal/rest_completion_queue.cc` line 27 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5s-r2-27:-NJW_8f5E4F3WLbTusL2:b-80tlqk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.cc#L27)):*

Empty `pending_tags_` actively?
</blockquote></details>

Sure.
___
*[`google/cloud/internal/rest_completion_queue.cc` line 40 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5s-r2-40:-NJW_bYrE5ocf6lJvleG:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.cc#L40)):*

Nit: I'd reverse this check so that `kGotEvent`, the "good" case, is the fallthough case.
</blockquote></details>

Done.
___
*[`google/cloud/internal/rest_completion_queue.cc` line 47 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5s-r2-47:-NJWb3kT8EUwZIKSjZOd:bb0x8ap) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.cc#L47)):*

Nit: It is a little weird for this to say `kTimeout` when `check_queue()` has got nothing to do with time.
</blockquote></details>

It's less about what the lambda is named, but what the behavior of GetNext is.
___
*[`google/cloud/internal/rest_completion_queue.cc` line 54 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5s-r2-54:-NJW_pbxE_SzI89EcGmj:b1uyshe) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.cc#L54)):*

`sleep_until()` has an odor, the prime one being: don't you want this to unblock early if someone does an `AddTag()` in the meantime?

It will also make the API harder to test.
</blockquote></details>

I do want want it to unblock early, eventually. I imagine adding in a condition variable could achieve this, but I'd prefer this naive implementation for now and iterate on it later.
___
*[`google/cloud/internal/rest_completion_queue_test.cc` line 27 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5u-r2-27:-NJWbDWb8jGFAT_jaEZb:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue_test.cc#L27)):*

It is probably nicer to reverse such declarations so that the queue is destroyed before the objects it (kind of) contains.
</blockquote></details>

Done.

coryan:
*[`google/cloud/internal/rest_completion_queue.h` line 56 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5t-r2-56:-NJes51ACvhAeBrfyvG0:bk5wpj3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.h#L56)):*
> I'm modeling this class on the interface of grpc::CompletionQueue as that's what functionality it is replacing in a REST transport world.

Why?  `grpc::CompletionQueue` has a horrible interface, that is why we have to create a complicated wrapper.  Can you show me some code of how this class will be used?  Maybe I can propose a simpler API to achieve what you want.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455)* status: 0 of 6 files reviewed, 9 unresolved discussions (waiting on @coryan and @devbww)

___
*[`google/cloud/internal/rest_completion_queue.h` line 56 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5t-r2-56:-NJfqzs47lCHPU4DZ9xP:bajxwzw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.h#L56)):*

> I'm modeling this class on the interface of grpc::CompletionQueue as that's what functionality it is replacing in a REST transport world.

Why?  `grpc::CompletionQueue` has a horrible interface, that is why we have to create a complicated wrapper.  Can you show me some code of how this class will be used?  Maybe I can propose a simpler API to achieve what you want.


</blockquote></details>

a) because this class is merely a placeholder, in the internal section of our repo that we can iterate on once we get a better understanding of what a libcurl multi handle completion queue would look like
b) its presence helps keep the library dependencies straight
c) it unblocks the work that we originally wanted to get done this quarter

I'll create a draft PR that has all my intended changes to help give a better overview of where i'm headed with this.

coryan:
*[`google/cloud/internal/rest_completion_queue.h` line 56 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10455#-NJSxMAuOV4NkQO62A5t-r2-56:-NJfy1oj6XH51at_GF7d:bxvwr5a) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/431069509d33d0b8094377643ef760bb10f7cfa7/google/cloud/internal/rest_completion_queue.h#L56)):*

a) because this class is merely a placeholder, in the internal section of our repo that we can iterate on once we get a better understanding of what a libcurl multi handle completion queue would look like
b) its presence helps keep the library dependencies straight
c) it unblocks the work that we originally wanted to get done this quarter

I'll create a draft PR that has all my intended changes to help give a better overview of where i'm headed with this.
</blockquote></details>

Looking at your PR it appears that you are only using `CompletionQueue::Run()` and `CompletionQueue::RunAsync()`.  If that is the case then I think all you need to implement is the timer queue that we discussed over IM.  The implementation of `RunAsync()` already runs the functor using a timer.  Your `RestCompletionQueueImpl` would need to be something like:

```cc
class RestCompletionQueueImpl : public CompletionQueueImpl {
 private:
    TimerQueue<std::chrono::system_clock> tq_;
 public:
    future<...> MakeRelativeTimer(... d) { return MakeDeadlineTimer(std::chrono::system_clock::now() + d); }
    future<...> MakeDeadlineTimer(std::chrono::system_clock::time_point tp) { return tq_.Schedule(tp); }
    void Run() { tq_.Service(); }
    void Shutdown() { tq_.Shutdown(); }
};
```

One may be concerned that the thread pool assigned to these things is not enough for all the `RunAsync()` calls.  If that is the case you could do something like this in your code (I am ignoring the lifetime issues with all the references, you need to fix that too, but we can do that separately):

```cc
future<StatusOr<google::longrunning::Operation>>
Default$stub_rest_class_name$::Async$method_name$(
      CompletionQueue& cq,
      rest_internal::RestContext& rest_context,
      $request_type$ const& request) {
  promise<StatusOr<google::longrunning::Operation>> p;
  future<StatusOr<google::longrunning::Operation>> f = p.get_future();
  // Create a thread to execute the HTTP request
  std::thread thread{[this](auto promise) {
    promise.set_value(rest_internal::$method_http_verb$<google::longrunning::Operation>(*service, rest_context, request);
    }, std::move(promise));
  // Join the thread (in the CQ pool) when the operation complete. At that point the remaining work in the thread should be minimal.
  return f.then([t = std::move(thread), cq](auto f) { cq.RunAsync([t = std::move(t)]() { t.join(); }); return f.get(); });
}
```

The main idea is to launch a new thread for each HTTP request, and use the CQ thread pool (which could be a single thread) to just *join* those threads.  Waiting until the thread is almost done to request the join is done via a `future::then()`.

---- Timer Queue ---

```cc
template <typename Clock>
class TimerQueue {
 public:
  using TimePoint = typename Clock::time_point;
  using FutureType = StatusOr<TimePoint>;

  TimerQueue() = default;
  TimerQueue(TimerQueue const&) = delete;
  TimerQueue& operator=(TimerQueue const&) = delete;

  future<FutureType> Schedule(TimePoint tp) {
    auto p = PromiseType();
    auto f = p.get_future();
    if (shutdown_)
      return make_ready_future(
          FutureType(Status(StatusCode::kCancelled, "queue already shutdown")));
    (void)timers_.emplace(tp, std::move(p));
    lk.unlock();
    cv_.notify_all();
    return f;
  }

  void Shutdown() {
    std::unique_lock<std::mutex> const lk(mu_);
    shutdown_ = true;
    cv_.notify_all();
  }

  void Service() {
    while (true) {
      auto const ne = NextExpiration();
      cv_.wait_until(
          lk, ne, [this, ne] { return shutdown_ || ne >= NextExpiration(); });
      if (shutdown_) return NotifyShutdown(std::move(lk));
      ExpireTimers(std::move(lk), Clock::now());
    }
  }

 private:
  void ExpireTimers(std::unique_lock<std::mutex> lk, TimePoint tp) {
    auto p = std::move(timers_.begin()->second);
    timers_.erase(timers_.begin());
    auto const has_expired_timers =
        !timers_.empty() && timers_.begin()->first < tp;
    lk.unlock();
    // We do not want to expire all the timers in this thread, as this would
    // serialize the expiration.  Instead, we pick another thread to continue
    // expiring timers.
    if (has_expired_timers) cv_.notify_one();
    p.set_value(tp);
  }

  void NotifyShutdown(std::unique_lock<std::mutex> lk) {
    auto timers = std::move(timers_);
    lk.unlock();
    for (auto& kv : timers) {
      kv.second.set_value(
          Status(StatusCode::kCancelled, "timer queue shutdown"));
    }
  }

  TimePoint NextExpiration() {
    auto constexpr kMaxSleep = std::chrono::hours(1);
    if (timers_.empty()) return Clock::now() + kMaxSleep;
    return timers_.begin()->first;
  }
  using PromiseType = promise<StatusOr<TimePoint>>;
  std::mutex mu_;
  std::condition_variable cv_;
  std::multimap<TimePoint, PromiseType> timers_;
  bool shutdown_ = false;
};
```

google/cloud/internal/rest_completion_queue.h
```diff
@@ -0,0 +1,184 @@
...
+                << std::endl;
+      auto p = std::move(timers_.begin()->second);
+      timers_.erase(timers_.begin());
+      auto const has_expired_timers =
+          !timers_.empty() && timers_.begin()->first < tp;
+      lk.unlock();
+      // We do not want to expire all the timers in this thread, as this would
+      // serialize the expiration.  Instead, we pick another thread to continue
+      // expiring timers.
+      if (has_expired_timers) cv_.notify_one();
```
coryan, line 143: Maybe this should be 

```cc
if (!has_expired_timers) return;
lk.unlock();
cv_.notify_one();
p.set_value(tp);
```

scotthart:
The TimerQueue was added in #10499.

PR #10456, opened 2022-12-18
impl(generator): support forwarding headers

dbolduc:
Part of the work for #10170 

If this is too much for one PR, just say the words, and I will try to chop it up into smaller pieces.

This PR adds support for generating backwards compatibility include headers and using directives, by adding a `forwarding_product_path` field in the generator config.

Now we can relocate a service from `foo` to `foo/v1` by setting:
- `product_path: google/cloud/foo/v1`
- `forwarding_product_path: google/cloud/foo`

... and then manually updating the build files.

The `check-api` build should catch if there are any discrepancies in the API when we start moving libraries. But I added a backwards compatibility unit test that checks that the legacy types are defined, just to be safe.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10456)


generator/internal/service_code_generator.cc
```diff
@@ -253,19 +253,21 @@ void ServiceCodeGenerator::CcSystemIncludes(
   GenerateSystemIncludes(cc_, system_includes);
 }
 
-Status ServiceCodeGenerator::HeaderOpenNamespaces(NamespaceType ns_type) {
+Status ServiceCodeGenerator::HeaderOpenNamespaces(NamespaceType ns_type,
+                                                  ProductPathType pp_type) {
```
coryan, line 257: This use of enums looks odd.  It may be simpler to have two functions `HeaderOpenNamespaces()` and `HeaderOpenForwardingNamespaces()`.

coryan (approved):
Only one actionable comment. I think I could live with the code as-is, but it could be improved.

generator/integration_tests/golden_config.textproto
```diff
@@ -26,4 +26,5 @@ service {
   retryable_status_codes: ["GoldenKitchenSink.kInternal", "kUnavailable", "GoldenThingAdmin.kDeadlineExceeded"]
   generate_round_robin_decorator: true
   generate_rest_transport: true
+  forwarding_product_path: "generator/integration_tests/golden"
```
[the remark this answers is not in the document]
dbolduc: I am aware, and I agree with your overall point.

> This means the only golden file integration test uses the `forwarding_product_path` feature. We have no tests without it.

Yes but this is true for every feature. Note that only `forwarding_headers_test.cc` uses the forwarding headers. The other unit tests use the `golden_v1` types and headers directly.

generator/internal/service_code_generator.cc
dbolduc: Done.

coryan (approved):

PR #10457, opened 2022-12-19
impl(oauth2): subject token source for AWS credentials

coryan:
Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10457)

### DIFF
--- a/google/cloud/internal/external_account_token_source_aws_test.cc
+++ b/google/cloud/internal/external_account_token_source_aws_test.cc
@@ -38,6 +38,7 @@ using ::google::cloud::testing_util::MockRestClient;
 using ::google::cloud::testing_util::MockRestResponse;
 using ::google::cloud::testing_util::ScopedEnvironment;
 using ::google::cloud::testing_util::StatusIs;
+using ::testing::_;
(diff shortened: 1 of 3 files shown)


google/cloud/internal/external_account_token_source_aws_test.cc
```diff
@@ -103,6 +114,159 @@ ExternalAccountTokenSourceAwsInfo MakeTestInfoImdsV2() {
...
+  EXPECT_THAT(
+      subject_token->token,
+      HasSubstr(nlohmann::json{{"key", "host"},
+                               {"value", "sts.expected-region.amazonaws.com"}}
+                    .dump()));
+}
+
+TEST(ExternalAccountTokenSource, SourceImdsv2Failure) {
+  // To simplify the test we provide all the parameters via environment
+  // variables and avoid using imdsv2.
```
dbolduc, line 160: This test seems to use imdsv2. Should we remove the comment and not set any env vars?

dbolduc (approved):

PR #10464, opened 2022-12-20
impl(oauth2): fix AWS subject token generation

coryan:
This fixes a number of mistakes in the AWS subject token generation. I was able to test these changes manually using a VM in AWS.

The largest changes are:

- If present, the `x-amz-security-token` header must be signed, this required a new test
- It is also a recommended that the `x-goog-cloud-target-resource` header is signed too
- The subject token must be URL-encoded. Effectively this URL-encodes the subject token twice, as it is included in form payload.
- The signature should be based on the `YYYYMMDD` date, not the `YYYYMMDDTHHMMSSZ` timestamp
- I was missing a separator line for the canonical request
- s/AWS/AWS4/

Part of the work for #5915

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10464)

### DIFF
--- a/google/cloud/internal/external_account_token_source_aws.cc
+++ b/google/cloud/internal/external_account_token_source_aws.cc
@@ -14,6 +14,7 @@
 
 #include "google/cloud/internal/external_account_token_source_aws.h"
 #include "google/cloud/internal/absl_str_cat_quiet.h"
+#include "google/cloud/internal/absl_str_join_quiet.h"
 #include "google/cloud/internal/absl_str_replace_quiet.h"
 #include "google/cloud/internal/external_account_source_format.h"
 #include "google/cloud/internal/format_time_point.h"
@@ -24,7 +25,11 @@
 #include "google/cloud/internal/sha256_hmac.h"
 #include "absl/strings/match.h"
 #include "absl/strings/str_split.h"
+#include "absl/strings/string_view.h"
+#include <algorithm>
+#include <array>
 #include <chrono>
+#include <utility>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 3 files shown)


google/cloud/internal/external_account_token_source_aws.cc
```diff
@@ -80,9 +85,20 @@ StatusOr<internal::SubjectToken> Source(
...
+  using CharMapping = std::pair<absl::string_view, absl::string_view>;
+  auto const mappings = std::array<CharMapping, 25>{{
+      {" ", "%20"}, {"\"", "%22"}, {"#", "%23"},  {"$", "%24"}, {"%", "%25"},
+      {"&", "%26"}, {"+", "%2B"},  {",", "%2C"},  {"/", "%2F"}, {":", "%3A"},
+      {";", "%3B"}, {"<", "%3C"},  {"=", "%3D"},  {">", "%3E"}, {"?", "%3F"},
+      {"@", "%40"}, {"[", "%5B"},  {"\\", "%5C"}, {"]", "%5D"}, {"^", "%5E"},
+      {"`", "%60"}, {"{", "%7B"},  {"|", "%7C"},  {"}", "%7D"}, {"\177", "%7F"},
+  }};
+
+  return internal::SubjectToken{absl::StrReplaceAll(subject, mappings)};
```
devbww, line 101: Possible followup: Perhaps we should make an `internal/url_encoder.cc`.

google/cloud/internal/external_account_token_source_aws.cc
```diff
@@ -304,16 +320,35 @@ nlohmann::json ComputeSubjectToken(
...
+        {"host", host},
+        {"x-amz-date", timestamp},
+        {"x-goog-cloud-target-resource", target},
+    };
+    // The session token may be empty, in which case we do not need to include
+    // it.
+    if (!secrets.session_token.empty()) {
+      headers.emplace_back("x-amz-security-token", secrets.session_token);
+    }
+    std::sort(headers.begin(), headers.end());
```
devbww, line 334: Q: Is this required?

devbww (approved):

google/cloud/internal/external_account_token_source_aws.cc
coryan: Yes:

https://docs.aws.amazon.com/general/latest/gr/create-signed-request.html#:~:text=use%20lowercase%20characters%2C-,must%20appear%20in%20alphabetical%20order,-%2C%20and%20must%20be

Arguably we could insert these in order ourselves, but I decided to be lazy and more explicit than not in this code.

google/cloud/internal/external_account_token_source_aws.cc
coryan: Ack.

PR #10473, opened 2022-12-22
refactor(generator): consolidate some stub member functions

dbolduc:
All gRPC stubs have the same public methods. I think there are enough stubs to justify this consolidation.

It would be nice if we could do something similar for the method signatures in the `.cc` files... but I have not thought that far ahead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10473)

### DIFF
--- a/generator/internal/stub_generator_base.cc
+++ b/generator/internal/stub_generator_base.cc
@@ -1,0 +1,148 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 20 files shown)


generator/internal/stub_generator_base.cc
```diff
@@ -0,0 +1,148 @@
...
+    google::protobuf::compiler::GeneratorContext* context)
+    : ServiceCodeGenerator(header_path_key, cc_path_key, service_descriptor,
+                           std::move(service_vars),
+                           std::move(service_method_vars), context) {}
+
+void StubGeneratorBase::HeaderPrintPublicMethods() {
+  for (auto const& method : methods()) {
+    if (IsStreamingWrite(method)) {
+      HeaderPrintMethod(method, __FILE__, __LINE__, R"""(
+  std::unique_ptr<::google::cloud::internal::StreamingWriteRpc<
```
devbww, line 37: I think I see that this is preserving previous behavior, but the rule of thumb is only to fully qualify a name in a top-level using declaration, or where it might be required.  So, I'd vote for dropping the leading `::` here and elsewhere.

dbolduc: Ack. I could do that in a follow up PR

devbww: SGTM.

devbww (approved):

PR #10474, opened 2022-12-23
impl(rest): make the CurlImpl spill buffer circular

### LINKED ISSUE #7957: REST: determine if libcurl supports adding multiple headers
Currently we call curl_slist_append once per header we need to add. If libcurl can support adding multiple headers at once, update the implementation of CurlImpl accordingly.

### LINKED ISSUE #7958: REST: consider using a circular buffer for CurlImpl spill buffer
The spill buffer is currently implemented as an array which requires a memmove every time we read some bytes from it. Using a circular buffer would avoid this.

### LINKED ISSUE #8059: REST: refactor moving bytes between buffers
In `CurlImpl::WriteToUserBuffer` we juggle bytes between 3 different buffers, look to refactor this such that all buffers are represented as `absl::Span<char>` possibly like:

```
std::pair<absl::Span<char>, absl::Span<char>> CopyWithSpill(absl::Span<T> destination, absl::Span<T> spill, absl::Span<T> source) { ...}

std::tie(buffer_, spill_) = CopyWithSpill(std::move(buffer_), std::move(spill_), absl::Span<char>(ptr, nmemb * size));
```

### LINKED ISSUE #9200: Remove check for empty Authorization header
Determine root cause for calling CurlImpl::SetHeader with "authorization: ", fix it, and remove existing guard clause.

devbww:
Eliminate "shift down" logic in the spill buffer and WriteVector.  While this probably doesn't make a measurable difference ...

```
  //google/cloud:internal_curl_rest_client_integration_test
    old over 100 runs: max = 5.0s, min = 2.6s, avg = 4.6s, dev = 0.3s
    new over 100 runs: max = 5.0s, min = 2.3s, avg = 4.6s, dev = 0.4s
```

..., removing redundant copying is always a good thing.  Fixes #7958.

Move the spill buffer code into its own class so that it might be tested individually.

Initialize `download_stall_minimum_rate_` (from its option) instead of overwriting `transfer_stall_minimum_rate_`.

Do not store an entire `Options` object when we only need the value of the `CurlFollowLocationOption`.

Close #8059 by commenting on how the libcurl API does not support data buffers owned by the caller.  This means there will always be additional copying of the request/response data into/from libcurl.

Close #9200 by removing the TODO for an already-answered question.

Close #7957 by removing the TODO for an already-answered question.

Group the `extern "C"` libcurl callbacks, and make them static within the cc file.  Also immediately convert the libcurl ptr/size pairs to `absl::Span<char>`.

Cleanup some comments that seem to refer to older versions of the code, and add some missing includes.  Rename some variables, like `buffer` and `ptr`, to give a clearer indiction of their roles.  Remove the trailing newlines from all `GCP_LOG()`ing.

And, the piece de resistance, which is sure to excite reviewers, reorder function declarations/definitions into a clearer, top-down layout.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10474)

### DIFF
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -45,23 +45,18 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 namespace {
 
-std::string UserAgentSuffix() {
-  static auto const* const kUserAgentSuffix = new auto([] {
-    return absl::StrCat(internal::UserAgentPrefix(), " ", curl_version());
-  }());
-  return *kUserAgentSuffix;
-}
-
-std::string NormalizeEndpoint(std::string endpoint) {
-  if (!endpoint.empty() && endpoint.back() != '/') endpoint.push_back('/');
-  return endpoint;
-}
-
 char const* InitialQueryParameterSeparator(std::string const& url) {
   // Abseil <= 20200923 does not implement StrContains(.., char)
   // NOLINTNEXTLINE(abseil-string-find-str-contains)
   if (url.find('?') != std::string::npos) return "&";
   return "?";
+}
+
+std::string UserAgentSuffix() {
+  static auto const* const kUserAgentSuffix = new auto([] {
+    return absl::StrCat(internal::UserAgentPrefix(), " ", curl_version());
+  }());
+  return *kUserAgentSuffix;
 }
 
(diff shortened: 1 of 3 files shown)


coryan (approved):
A bit hard to follow with all the changes in a single PR, but LGTM.

devbww:
> A bit hard to follow with all the changes in a single PR, ...

Yeah ... I lost my head.

PR #10497, opened 2023-01-05
ci: fast checkers can skip markdown steps

dbolduc:
Follow up to #10495 

Only run scripts from `ci/generate-markdown/` if the script has changed OR if a library has changes made in it. This shaves off another 10s from the checkers build.

I did not bother to be more fine-grained with the script changes. If any file in `ci/generate-markdown/` has changed, we just regenerate everything.

## Results

Running on my gLaptop where the only changes are in `checkers.sh`.

```sh
$ ci/cloudbuild/build.sh -t checkers-pr

Running markdown generators:                      ... 6.082 seconds
Running doxygen landing-page updates:             ... 5.056 seconds
==> 🕑 checkers completed in 125.783 seconds
```

```sh
$ env GOOGLE_CLOUD_CPP_FAST_CHECKERS=1 ci/cloudbuild/build.sh -t checkers-pr

Running markdown generators:                      ... 0.000 seconds
Running doxygen landing-page updates:             ... 0.000 seconds
==> 🕑 checkers completed in 1.724 seconds
```

The `0.000 seconds` is a bit of a lie, because I make the library list outside of the timed step.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10497)

### DIFF
--- a/ci/cloudbuild/builds/checkers.sh
+++ b/ci/cloudbuild/builds/checkers.sh
@@ -69,6 +69,13 @@ git_files() {
   fi
 }
 
+CI_CHANGES="$(git_files -z -- "ci/generate-markdown/")"
+if [ -z "${CI_CHANGES}" ]; then
+  mapfile -t GIT_LIBS < <(comm -12 <(features::libraries) <(git_files google/cloud/*/ | cut -d "/" -f3 | uniq))
+else
+  mapfile -t GIT_LIBS < <(features::libraries)
+fi
+
 # This controls the output format from bash's `time` command, which we use
 # below to time blocks of the script. A newline is automatically included.
 readonly TIMEFORMAT="... %R seconds"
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/builds/checkers.sh
```diff
@@ -69,6 +69,13 @@ git_files() {
   fi
 }
 
+CI_CHANGES="$(git_files -z -- "ci/generate-markdown/")"
```
coryan, line 72: If you are going to do dependency management, then use a tool for dependency management.  We have at least two.

doc/packaging.md also depends on the `ci/**/demo-*.Dockerfile` files.

dbolduc: > `doc/packaging.md` also depends on the `ci/**/demo-*.Dockerfile` files.

Argh, you're right. I was afraid of this sort of dependency blow up.

My gut says that `FAST_BUT_SOMETIMES_WRONG_CHECKERS` is still a value gain. But I will think it over some more....

coryan: Maybe what we should have is a `quick-format` script.  Refactor the formatting functions to a common library in ci/cloudbuild/builds/lib, then use the functions from `checkers.sh`, and create a `quick-format.sh` script that only formats what you want.

dbolduc: > Refactor the formatting functions to a common library

Oooh, I really like this idea. It would make it way easier to run the formatters after `generate-libraries` (#5821), which I especially want to do as part of (#10237). https://github.com/dbolduc/google-cloud-cpp/blob/4af53dd5eb2a73529788b0997097b14ed105279b/ci/cloudbuild/builds/generate-libraries.sh#L70-L103

(Not working on this now. But maybe I will do it on some rainy day when I am operator)

dbolduc:
Didn't get to this in my operator cycle. Maybe I will get to it in the next one.

PR #10499, opened 2023-01-05
impl: add TimerQueue

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499)

### DIFF
--- a/google/cloud/internal/timer_queue.h
+++ b/google/cloud/internal/timer_queue.h
@@ -1,0 +1,141 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_TIMER_QUEUE_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_TIMER_QUEUE_H
+
+#include "google/cloud/future.h"
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include <chrono>
+#include <map>
+#include <mutex>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+template <typename Clock>
+class TimerQueue {
+ public:
+  using TimePoint = typename Clock::time_point;
+  using FutureType = StatusOr<TimePoint>;
+
+  TimerQueue() = default;
+  TimerQueue(TimerQueue const&) = delete;
+  TimerQueue& operator=(TimerQueue const&) = delete;
+
+  future<FutureType> Schedule(TimePoint tp) {
+    auto p = PromiseType();
+    auto f = p.get_future();
+    {
+      std::lock_guard<std::mutex> lk(mu_);
+      if (shutdown_)
+        return make_ready_future(FutureType(
+            Status(StatusCode::kCancelled, "Queue already shutdown.")));
+      (void)timers_.emplace(tp, std::move(p));
+      timer_scheduled_ = true;
+    }
+    cv_.notify_all();
+    return f;
+  }
+
+  void Shutdown() {
+    std::unique_lock<std::mutex> const lk(mu_);
+    shutdown_ = true;
+    cv_.notify_all();
+  }
+
+  // The fn parameter is executed after the timer has been expired and is
+  // intended for testing purposes only.
+  void Service(std::function<void(void)> const& fn = {}) {
+    while (true) {
+      std::unique_lock<std::mutex> lk(mu_);
+      if (NextExpiration() > Clock::now()) {
+        cv_.wait_until(lk, NextExpiration(), [this] {
+          return shutdown_ || timer_scheduled_ ||
+                 NextExpiration() <= Clock::now();
+        });
+        timer_scheduled_ = false;
+      }
+      if (shutdown_) return NotifyShutdown(std::move(lk));
+      ExpireTimers(std::move(lk), Clock::now(), fn);
+    }
+  }
+
+  void CancelAll() {
+    std::multimap<TimePoint, PromiseType> timers;
+    {
+      std::lock_guard<std::mutex> lk(mu_);
+      timers = std::move(timers_);
+    }
+    for (auto& kv : timers) {
+      kv.second.set_value(Status(StatusCode::kCancelled, "Timer cancelled."));
+    }
+  }
+
+  // For testing purposes.
+  std::int64_t expired_timer_counter() const {
+    return expired_timer_counter_.load();
+  }
+
+ private:
+  void ExpireTimers(std::unique_lock<std::mutex> lk, TimePoint tp,
+                    std::function<void(void)> const& fn = {}) {
+    if (timers_.empty() || timers_.begin()->first > tp) return;
+    auto p = std::move(timers_.begin()->second);
+    timers_.erase(timers_.begin());
+    auto const has_expired_timers =
+        !timers_.empty() && timers_.begin()->first < tp;
+    lk.unlock();
+    // We do not want to expire all the timers in this thread, as this would
+    // serialize the expiration.  Instead, we pick another thread to continue
+    // expiring timers.
+    if (has_expired_timers) cv_.notify_one();
+    p.set_value(tp);
+    if (fn) fn();
+    ++expired_timer_counter_;
+  }
+
+  void NotifyShutdown(std::unique_lock<std::mutex> lk) {
+    auto timers = std::move(timers_);
+    lk.unlock();
+    for (auto& kv : timers) {
+      kv.second.set_value(
+          Status(StatusCode::kCancelled, "Timer queue shutdown."));
+    }
+  }
+
+  TimePoint NextExpiration() {
+    auto constexpr kMaxSleep = std::chrono::seconds(3600);
+    if (timers_.empty()) return Clock::now() + kMaxSleep;
+    return timers_.begin()->first;
+  }
+
+  using PromiseType = promise<StatusOr<TimePoint>>;
+  std::mutex mu_;
+  std::condition_variable cv_;
+  std::atomic<std::int64_t> expired_timer_counter_{0};
+  std::multimap<TimePoint, PromiseType> timers_;
+  bool shutdown_ = false;
+  bool timer_scheduled_ = false;
+};
+
+}  // namespace internal
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
+}  // namespace cloud
+}  // namespace google
+
+#endif  // GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_TIMER_QUEUE_H
--- a/google/cloud/internal/timer_queue.h
+++ b/google/cloud/internal/timer_queue.h
@@ -27,104 +27,42 @@ namespace cloud {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
 
-template <typename Clock>
 class TimerQueue {
  public:
-  using TimePoint = typename Clock::time_point;
-
   TimerQueue() = default;
   TimerQueue(TimerQueue const&) = delete;
   TimerQueue& operator=(TimerQueue const&) = delete;
 
-  future<StatusOr<TimePoint>> Schedule(TimePoint tp) {
-    auto p = PromiseType();
-    auto f = p.get_future();
-    bool should_notify;
-    {
-      std::lock_guard<std::mutex> lk(mu_);
-      if (shutdown_)
-        return make_ready_future(StatusOr<TimePoint>(
-            Status(StatusCode::kCancelled, "Queue already shutdown.")));
-      auto iter = timers_.emplace(tp, std::move(p));
-      should_notify = iter == timers_.begin();
-    }
-    if (should_notify) cv_.notify_all();
-    return f;
-  }
+  // Adds a timer to the queue.
+  future<StatusOr<std::chrono::system_clock::time_point>> Schedule(
+      std::chrono::system_clock::time_point tp);
 
-  void Shutdown() {
-    std::unique_lock<std::mutex> const lk(mu_);
-    shutdown_ = true;
-    cv_.notify_all();
-  }
+  // Signals all threads that have called Service to return. Does not modify
+  // remaining timers. If desired, CancelAll can be called.
+  void Shutdown();
 
-  // Threads monitoring the queue and expiring timers should call Service.
-  void Service() {
-    while (true) {
-      std::unique_lock<std::mutex> lk(mu_);
-      auto const until = NextExpiration(lk);
-      auto predicate = [this, until, &lk] {
-        auto const ne = NextExpiration(lk);
-        return shutdown_ || ne < until || HasExpiredTimers(lk, Clock::now());
-      };
-      // TODO(#10512): Only wake one thread to service the next timer instead
-      // of all of them.
-      cv_.wait_until(lk, until, predicate);
-      if (shutdown_) return NotifyShutdown(std::move(lk));
-      if (HasExpiredTimers(lk, Clock::now())) {
-        ExpireTimers(std::move(lk), Clock::now());
-      }
-    }
-  }
+  // Timers added via Schedule should be managed by one or more threads that
+  // call Service. Calls to Service only return after Shutdown has been called.
+  void Service();
 
-  void CancelAll() {
-    std::multimap<TimePoint, PromiseType> timers;
-    {
-      std::lock_guard<std::mutex> lk(mu_);
-      std::swap(timers, timers_);
-    }
-    for (auto& timer : timers) {
-      timer.second.set_value(
-          Status(StatusCode::kCancelled, "Timer cancelled."));
-    }
-  }
+  // Cancels all timers.
+  void CancelAll();
 
  private:
-  bool HasExpiredTimers(std::unique_lock<std::mutex> const&, TimePoint tp) {
-    return !timers_.empty() && timers_.begin()->first < tp;
-  }
-  void ExpireTimers(std::unique_lock<std::mutex> lk, TimePoint tp) {
-    auto p = std::move(timers_.begin()->second);
-    timers_.erase(timers_.begin());
-    auto const has_expired_timers =
-        !timers_.empty() && timers_.begin()->first < tp;
-    lk.unlock();
-    // We do not want to expire all the timers in this thread, as this would
-    // serialize the expiration.  Instead, we pick another thread to continue
-    // expiring timers.
-    if (has_expired_timers) cv_.notify_one();
-    p.set_value(tp);
-  }
+  // Requires a lock to be held before calling.
+  bool HasExpiredTimer(std::unique_lock<std::mutex> const&,
+                       std::chrono::system_clock::time_point tp);
+  void ExpireTimer(std::unique_lock<std::mutex> lk,
+                   std::chrono::system_clock::time_point tp);
+  void NotifyShutdown(std::unique_lock<std::mutex> lk);
+  // Requires a lock to be held before calling.
+  std::chrono::system_clock::time_point NextExpiration(
+      std::unique_lock<std::mutex> const&);
 
-  void NotifyShutdown(std::unique_lock<std::mutex> lk) {
-    auto timers = std::move(timers_);
-    lk.unlock();
-    for (auto& kv : timers) {
-      kv.second.set_value(
-          Status(StatusCode::kCancelled, "Timer queue shutdown."));
-    }
-  }
-
-  TimePoint NextExpiration(std::unique_lock<std::mutex> const&) {
-    auto constexpr kMaxSleep = std::chrono::seconds(3600);
-    if (timers_.empty()) return Clock::now() + kMaxSleep;
-    return timers_.begin()->first;
-  }
-
-  using PromiseType = promise<StatusOr<TimePoint>>;
+  using PromiseType = promise<StatusOr<std::chrono::system_clock::time_point>>;
   std::mutex mu_;
   std::condition_variable cv_;
-  std::multimap<TimePoint, PromiseType> timers_;
+  std::multimap<std::chrono::system_clock::time_point, PromiseType> timers_;
   bool shutdown_ = false;
 };
 
--- a/google/cloud/internal/timer_queue.h
+++ b/google/cloud/internal/timer_queue.h
@@ -31,24 +31,24 @@ template <typename Clock>
 class TimerQueue {
  public:
   using TimePoint = typename Clock::time_point;
-  using FutureType = StatusOr<TimePoint>;
 
   TimerQueue() = default;
   TimerQueue(TimerQueue const&) = delete;
   TimerQueue& operator=(TimerQueue const&) = delete;
 
-  future<FutureType> Schedule(TimePoint tp) {
+  future<StatusOr<TimePoint>> Schedule(TimePoint tp) {
     auto p = PromiseType();
     auto f = p.get_future();
+    bool should_notify;
     {
       std::lock_guard<std::mutex> lk(mu_);
       if (shutdown_)
-        return make_ready_future(FutureType(
+        return make_ready_future(StatusOr<TimePoint>(
             Status(StatusCode::kCancelled, "Queue already shutdown.")));
-      (void)timers_.emplace(tp, std::move(p));
-      timer_scheduled_ = true;
+      auto iter = timers_.emplace(tp, std::move(p));
+      should_notify = iter == timers_.begin();
     }
-    cv_.notify_all();
+    if (should_notify) cv_.notify_all();
     return f;
   }
 
@@ -58,20 +58,22 @@ class TimerQueue {
     cv_.notify_all();
   }
 
-  // The fn parameter is executed after the timer has been expired and is
-  // intended for testing purposes only.
-  void Service(std::function<void(void)> const& fn = {}) {
+  // Threads monitoring the queue and expiring timers should call Service.
+  void Service() {
     while (true) {
       std::unique_lock<std::mutex> lk(mu_);
-      if (NextExpiration() > Clock::now()) {
-        cv_.wait_until(lk, NextExpiration(), [this] {
-          return shutdown_ || timer_scheduled_ ||
-                 NextExpiration() <= Clock::now();
-        });
-        timer_scheduled_ = false;
-      }
+      auto const until = NextExpiration(lk);
+      auto predicate = [this, until, &lk] {
+        auto const ne = NextExpiration(lk);
+        return shutdown_ || ne < until || HasExpiredTimers(lk, Clock::now());
+      };
+      // TODO(#10512): Only wake one thread to service the next timer instead
+      // of all of them.
+      cv_.wait_until(lk, until, predicate);
       if (shutdown_) return NotifyShutdown(std::move(lk));
-      ExpireTimers(std::move(lk), Clock::now(), fn);
+      if (HasExpiredTimers(lk, Clock::now())) {
+        ExpireTimers(std::move(lk), Clock::now());
+      }
     }
   }
 
@@ -79,22 +81,19 @@ class TimerQueue {
     std::multimap<TimePoint, PromiseType> timers;
     {
       std::lock_guard<std::mutex> lk(mu_);
-      timers = std::move(timers_);
+      std::swap(timers, timers_);
     }
-    for (auto& kv : timers) {
-      kv.second.set_value(Status(StatusCode::kCancelled, "Timer cancelled."));
+    for (auto& timer : timers) {
+      timer.second.set_value(
+          Status(StatusCode::kCancelled, "Timer cancelled."));
     }
-  }
-
-  // For testing purposes.
-  std::int64_t expired_timer_counter() const {
-    return expired_timer_counter_.load();
   }
 
  private:
-  void ExpireTimers(std::unique_lock<std::mutex> lk, TimePoint tp,
-                    std::function<void(void)> const& fn = {}) {
-    if (timers_.empty() || timers_.begin()->first > tp) return;
+  bool HasExpiredTimers(std::unique_lock<std::mutex> const&, TimePoint tp) {
+    return !timers_.empty() && timers_.begin()->first < tp;
+  }
+  void ExpireTimers(std::unique_lock<std::mutex> lk, TimePoint tp) {
     auto p = std::move(timers_.begin()->second);
     timers_.erase(timers_.begin());
     auto const has_expired_timers =
@@ -105,8 +104,6 @@ class TimerQueue {
     // expiring timers.
     if (has_expired_timers) cv_.notify_one();
     p.set_value(tp);
-    if (fn) fn();
-    ++expired_timer_counter_;
   }
 
   void NotifyShutdown(std::unique_lock<std::mutex> lk) {
@@ -118,7 +115,7 @@ class TimerQueue {
     }
   }
 
-  TimePoint NextExpiration() {
+  TimePoint NextExpiration(std::unique_lock<std::mutex> const&) {
     auto constexpr kMaxSleep = std::chrono::seconds(3600);
     if (timers_.empty()) return Clock::now() + kMaxSleep;
     return timers_.begin()->first;
@@ -127,10 +124,8 @@ class TimerQueue {
   using PromiseType = promise<StatusOr<TimePoint>>;
   std::mutex mu_;
   std::condition_variable cv_;
-  std::atomic<std::int64_t> expired_timer_counter_{0};
   std::multimap<TimePoint, PromiseType> timers_;
   bool shutdown_ = false;
-  bool timer_scheduled_ = false;
 };
 
 }  // namespace internal
--- a/google/cloud/internal/timer_queue_test.cc
+++ b/google/cloud/internal/timer_queue_test.cc
@@ -1,0 +1,134 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/internal/timer_queue.h"
+#include "google/cloud/testing_util/status_matchers.h"
+#include <gmock/gmock.h>
+#include <chrono>
+#include <thread>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+namespace {
+
+using ::google::cloud::testing_util::IsOk;
+using ::google::cloud::testing_util::StatusIs;
+using testing::Each;
+using ::testing::Eq;
+
+TEST(TimerQueueTest, ScheduleSingleRunner) {
+  TimerQueue<std::chrono::system_clock> tq;
+  std::thread t([&tq] { tq.Service(); });
+  auto const duration = std::chrono::milliseconds(1);
+  auto now = std::chrono::system_clock::now();
+  auto f = tq.Schedule(now + duration);
+  auto expire_time = f.get();
+  tq.Shutdown();
+  t.join();
+
+  EXPECT_THAT(expire_time, IsOk());
+  EXPECT_GE(*expire_time - now, duration);
+  EXPECT_THAT(tq.expired_timer_counter(), Eq(1));
+}
+
+TEST(TimerQueueTest, ScheduleAndCancelAll) {
+  TimerQueue<std::chrono::system_clock> tq;
+  std::thread t([&tq] { tq.Service(); });
+  auto const duration = std::chrono::seconds(60);
+  auto now = std::chrono::system_clock::now();
+  auto f = tq.Schedule(now + duration);
+  auto f2 = tq.Schedule(now + duration);
+  tq.CancelAll();
+  tq.Shutdown();
+  t.join();
+
+  auto expire_time = f.get();
+  EXPECT_THAT(expire_time, StatusIs(StatusCode::kCancelled));
+  auto expire_time2 = f2.get();
+  EXPECT_THAT(expire_time2, StatusIs(StatusCode::kCancelled));
+  EXPECT_THAT(tq.expired_timer_counter(), Eq(0));
+}
+
+TEST(TimerQueueTest, ScheduleMultipleRunners) {
+  TimerQueue<std::chrono::system_clock> tq;
+  auto constexpr kRunners = 8;
+  std::vector<std::thread> runners;
+  // When a runner expires a timer, we want to add it to this set.
+  std::mutex mu;
+  std::set<std::thread::id> expiring_runner_ids;  // GUARDED_BY mu
+  for (auto i = 0; i != kRunners; ++i) {
+    runners.emplace_back([&] {
+      tq.Service([&] {
+        std::lock_guard<std::mutex> lk(mu);
+        expiring_runner_ids.insert(std::this_thread::get_id());
+      });
+    });
+  }
+
+  auto const duration = std::chrono::seconds(1);
+  auto now = std::chrono::system_clock::now();
+  std::vector<future<StatusOr<std::chrono::system_clock::time_point>>> futures;
+  futures.reserve(100);
+  for (int i = 0; i != 100; ++i) {
+    futures.push_back(tq.Schedule(now + duration));
+  }
+  std::vector<StatusOr<std::chrono::system_clock::time_point>> timepoints;
+  timepoints.reserve(100);
+  for (auto& f : futures) {
+    timepoints.push_back(f.get());
+  }
+  tq.Shutdown();
+  for (auto& t : runners) t.join();
+
+  EXPECT_THAT(timepoints, Each(IsOk()));
+  // Verify expiring timers are not handled serially by a single runner.
+  EXPECT_GT(expiring_runner_ids.size(), 1);
+  EXPECT_THAT(tq.expired_timer_counter(), Eq(100));
+}
+
+TEST(TimerQueueTest, ScheduleAndCancelAllMultipleRunners) {
+  TimerQueue<std::chrono::system_clock> tq;
+  auto constexpr kRunners = 8;
+  std::vector<std::thread> runners;
+  for (auto i = 0; i != kRunners; ++i) {
+    runners.emplace_back([&] { tq.Service(); });
+  }
+
+  auto const duration = std::chrono::seconds(1);
+  auto now = std::chrono::system_clock::now();
+  std::vector<future<StatusOr<std::chrono::system_clock::time_point>>> futures;
+  futures.reserve(100);
+  for (int i = 0; i != 100; ++i) {
+    futures.push_back(tq.Schedule(now + duration));
+  }
+  tq.CancelAll();
+  std::vector<StatusOr<std::chrono::system_clock::time_point>> timepoints;
+  timepoints.reserve(100);
+  for (auto& f : futures) {
+    timepoints.push_back(f.get());
+  }
+  tq.Shutdown();
+  for (auto& t : runners) t.join();
+
+  EXPECT_THAT(timepoints, Each(StatusIs(StatusCode::kCancelled)));
+  EXPECT_THAT(tq.expired_timer_counter(), Eq(0));
+}
+
+}  // namespace
+}  // namespace internal
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
+}  // namespace cloud
+}  // namespace google
--- a/google/cloud/internal/timer_queue_test.cc
+++ b/google/cloud/internal/timer_queue_test.cc
@@ -29,7 +29,7 @@ using ::google::cloud::testing_util::StatusIs;
 using testing::Each;
 
 TEST(TimerQueueTest, ScheduleSingleRunner) {
-  TimerQueue<std::chrono::system_clock> tq;
+  TimerQueue tq;
   std::thread t([&tq] { tq.Service(); });
   auto const duration = std::chrono::milliseconds(1);
   auto now = std::chrono::system_clock::now();
@@ -38,12 +38,12 @@ TEST(TimerQueueTest, ScheduleSingleRunner) {
   tq.Shutdown();
   t.join();
 
-  EXPECT_THAT(expire_time, IsOk());
+  ASSERT_THAT(expire_time, IsOk());
   EXPECT_GE(*expire_time - now, duration);
 }
 
 TEST(TimerQueueTest, ScheduleAndCancelAllSingleRunner) {
-  TimerQueue<std::chrono::system_clock> tq;
+  TimerQueue tq;
   auto const duration = std::chrono::seconds(60);
   auto now = std::chrono::system_clock::now();
   auto f = tq.Schedule(now + duration);
@@ -60,7 +60,7 @@ TEST(TimerQueueTest, ScheduleAndCancelAllSingleRunner) {
 }
 
 TEST(TimerQueueTest, ScheduleEarlierTimerSingleRunner) {
-  TimerQueue<std::chrono::system_clock> tq;
+  TimerQueue tq;
   std::thread t([&tq] { tq.Service(); });
   auto const duration = std::chrono::milliseconds(50);
   auto now = std::chrono::system_clock::now();
@@ -82,7 +82,7 @@ TEST(TimerQueueTest, ScheduleEarlierTimerSingleRunner) {
 }
 
 TEST(TimerQueueTest, ScheduleMultipleRunners) {
-  TimerQueue<std::chrono::system_clock> tq;
+  TimerQueue tq;
   // Schedule the timers first, so they are all ready to expire
   auto constexpr kTimers = 100;
   std::vector<future<std::thread::id>> futures(kTimers);
@@ -104,19 +104,16 @@ TEST(TimerQueueTest, ScheduleMultipleRunners) {
                  [&](auto& f) { return f.get(); });
 
   tq.Shutdown();
-  for (auto& t : runners)
-    t.join();  // BTW, there is a joinable thread class in
-               // testing_util/scoped_thread.h
+  for (auto& t : runners) t.join();
 
   EXPECT_GT(ids.size(), 1);
 }
 
 TEST(TimerQueueTest, ScheduleAndCancelAllMultipleRunners) {
-  TimerQueue<std::chrono::system_clock> tq;
+  TimerQueue tq;
   auto const duration = std::chrono::seconds(1);
   auto now = std::chrono::system_clock::now();
   std::vector<future<StatusOr<std::chrono::system_clock::time_point>>> futures;
-  futures.reserve(100);
   for (int i = 0; i != 100; ++i) {
     futures.push_back(tq.Schedule(now + duration));
   }
@@ -138,7 +135,7 @@ TEST(TimerQueueTest, ScheduleAndCancelAllMultipleRunners) {
 }
 
 TEST(TimerQueueTest, ScheduleEarlierTimerMultipleRunner) {
-  TimerQueue<std::chrono::system_clock> tq;
+  TimerQueue tq;
   auto constexpr kRunners = 8;
   std::vector<std::thread> runners;
   for (auto i = 0; i != kRunners; ++i) {
--- a/google/cloud/internal/timer_queue_test.cc
+++ b/google/cloud/internal/timer_queue_test.cc
@@ -27,7 +27,6 @@ namespace {
 using ::google::cloud::testing_util::IsOk;
 using ::google::cloud::testing_util::StatusIs;
 using testing::Each;
-using ::testing::Eq;
 
 TEST(TimerQueueTest, ScheduleSingleRunner) {
   TimerQueue<std::chrono::system_clock> tq;
@@ -41,17 +40,16 @@ TEST(TimerQueueTest, ScheduleSingleRunner) {
 
   EXPECT_THAT(expire_time, IsOk());
   EXPECT_GE(*expire_time - now, duration);
-  EXPECT_THAT(tq.expired_timer_counter(), Eq(1));
 }
 
-TEST(TimerQueueTest, ScheduleAndCancelAll) {
+TEST(TimerQueueTest, ScheduleAndCancelAllSingleRunner) {
   TimerQueue<std::chrono::system_clock> tq;
-  std::thread t([&tq] { tq.Service(); });
   auto const duration = std::chrono::seconds(60);
   auto now = std::chrono::system_clock::now();
   auto f = tq.Schedule(now + duration);
   auto f2 = tq.Schedule(now + duration);
   tq.CancelAll();
+  std::thread t([&tq] { tq.Service(); });
   tq.Shutdown();
   t.join();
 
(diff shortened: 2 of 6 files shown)


google/cloud/internal/timer_queue_test.cc
```diff
@@ -0,0 +1,134 @@
...
+namespace internal {
+namespace {
+
+using ::google::cloud::testing_util::IsOk;
+using ::google::cloud::testing_util::StatusIs;
+using testing::Each;
+using ::testing::Eq;
+
+TEST(TimerQueueTest, ScheduleSingleRunner) {
+  TimerQueue<std::chrono::system_clock> tq;
```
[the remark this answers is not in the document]
coryan: See my other comments though, it still might be useful.

google/cloud/internal/timer_queue.h
```diff
@@ -0,0 +1,141 @@
...
+    shutdown_ = true;
+    cv_.notify_all();
+  }
+
+  // The fn parameter is executed after the timer has been expired and is
+  // intended for testing purposes only.
+  void Service(std::function<void(void)> const& fn = {}) {
+    while (true) {
+      std::unique_lock<std::mutex> lk(mu_);
+      if (NextExpiration() > Clock::now()) {
```
coryan, line 66: You should not need the `if()` here.  It it is true, the condition variable will not block.

google/cloud/internal/timer_queue.h
```diff
@@ -0,0 +1,141 @@
...
+  void NotifyShutdown(std::unique_lock<std::mutex> lk) {
+    auto timers = std::move(timers_);
+    lk.unlock();
+    for (auto& kv : timers) {
+      kv.second.set_value(
+          Status(StatusCode::kCancelled, "Timer queue shutdown."));
+    }
+  }
+
+  TimePoint NextExpiration() {
```
coryan, line 121: Change the signature to:

```
TimePoint NextExpiration(std::unique_lock<std::mutex> const& lk)
```

google/cloud/internal/timer_queue.h
```diff
@@ -0,0 +1,141 @@
...
+    cv_.notify_all();
+  }
+
+  // The fn parameter is executed after the timer has been expired and is
+  // intended for testing purposes only.
+  void Service(std::function<void(void)> const& fn = {}) {
+    while (true) {
+      std::unique_lock<std::mutex> lk(mu_);
+      if (NextExpiration() > Clock::now()) {
+        cv_.wait_until(lk, NextExpiration(), [this] {
```
coryan, line 67: I think using the `timer_scheduled_` flag seems finicky.  Consider this instead:

```cc
void Service() {
  while (true) {
    auto const until = NextExpiration(lk); // NOTE: consistently pass `lk` parameter to functions that require a lock
    auto predicate = [this, until, &lk]() {
       auto const ne = NextExpiration(lk);
       return shutdown_ || ne < until /* need to adjust sleep time */ || HasExpiredTimers(lk);
    };
    cv_.wait_until(lk, until, predicate);
    if (shutdown_) return NotifyShutdown(std::move(lk));
    if (HasExpiredTimers(lk)) ExpireTimers(std::move(lk), Clock::now());
  }
}
```

google/cloud/internal/timer_queue.h
```diff
@@ -0,0 +1,141 @@
...
+
+  // For testing purposes.
+  std::int64_t expired_timer_counter() const {
+    return expired_timer_counter_.load();
+  }
+
+ private:
+  void ExpireTimers(std::unique_lock<std::mutex> lk, TimePoint tp,
+                    std::function<void(void)> const& fn = {}) {
+    if (timers_.empty() || timers_.begin()->first > tp) return;
```
coryan, line 97: With the changes suggested above, you can remove this check.

google/cloud/internal/timer_queue.h
```diff
@@ -0,0 +1,141 @@
...
+    auto f = p.get_future();
+    {
+      std::lock_guard<std::mutex> lk(mu_);
+      if (shutdown_)
+        return make_ready_future(FutureType(
+            Status(StatusCode::kCancelled, "Queue already shutdown.")));
+      (void)timers_.emplace(tp, std::move(p));
+      timer_scheduled_ = true;
+    }
+    cv_.notify_all();
```
devbww, line 51: It seems we don't need to notify anyone unless `tp` is a new minimum.

google/cloud/internal/timer_queue.h
devbww, line 67: It looks like this produces a thundering herd.  All the service threads will awaken when the earliest timer is ready, but only one will expire it.  Then they'll all block again, waiting for the next timer.  Only one thread should be awaiting the earliest timer (or a new, earlier one) ... the rest, if any, should be biding their time to take that role.

google/cloud/internal/timer_queue.h
```diff
@@ -0,0 +1,141 @@
...
+    auto timers = std::move(timers_);
+    lk.unlock();
+    for (auto& kv : timers) {
+      kv.second.set_value(
+          Status(StatusCode::kCancelled, "Timer queue shutdown."));
+    }
+  }
+
+  TimePoint NextExpiration() {
+    auto constexpr kMaxSleep = std::chrono::seconds(3600);
```
devbww, line 122: The arbitrary 1h is an indication that things might be better.  For example, consider calling `cv_.wait()` instead when `timers_` is empty.

google/cloud/internal/timer_queue.h
```diff
@@ -0,0 +1,141 @@
...
+    }
+    for (auto& kv : timers) {
+      kv.second.set_value(Status(StatusCode::kCancelled, "Timer cancelled."));
+    }
+  }
+
+  // For testing purposes.
+  std::int64_t expired_timer_counter() const {
+    return expired_timer_counter_.load();
+  }
```
devbww, line 92: It seems like this was redundant in all the tests, so probably better avoided.

devbww: The value returned is somewhat meaningless until `Shutdown()` is called, so an alternative (if you think you still need it) would be to return the count from `Shutdown()`.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499)* status: 0 of 5 files reviewed, 19 unresolved discussions (waiting on @coryan and @devbww)

___
*[`google/cloud/internal/timer_queue.h` line 34 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-34:-NL8Pjm5ChyBgttQ3IbA:b-84v791) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L34)):*

Q. Might things be clearer without this alias?
</blockquote></details>

Alias removed
___
*[`google/cloud/internal/timer_queue.h` line 51 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-51:-NL8Po8g5PAR_-64UjzM:bpghxmz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L51)):*

It seems we don't need to notify anyone unless `tp` is a new minimum.
</blockquote></details>

Added check before calling notify_all
___
*[`google/cloud/internal/timer_queue.h` line 61 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-61:-NL8PvoR1zp2ZLcqV_fp:bss4a2b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L61)):*

"the timer" doesn't have any meaning in this context.
</blockquote></details>

Removed.
___
*[`google/cloud/internal/timer_queue.h` line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-63:-NL8Pwys4T_22SdSz1wt:b-bizqh8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L63)):*

It is probably worth explaining who should call `Service()`.
</blockquote></details>

Added comment.
___
*[`google/cloud/internal/timer_queue.h` line 66 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-66:-NL8PzW4AOt2NROoX5ne:bo8w8qr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L66)):*

You should not need the `if()` here.  It it is true, the condition variable will not block.
</blockquote></details>

Removed if()
___
*[`google/cloud/internal/timer_queue.h` line 67 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-67:-NL8QxaC04ayLLHE5YFN:b-m9kz0x) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L67)):*

It looks like this produces a thundering herd.  All the service threads will awaken when the earliest timer is ready, but only one will expire it.  Then they'll all block again, waiting for the next timer.  Only one thread should be awaiting the earliest timer (or a new, earlier one) ... the rest, if any, should be biding their time to take that role.
</blockquote></details>

Removed `timer_scheduled_` and implemented suggestion. Created issue to track herd issue.
___
*[`google/cloud/internal/timer_queue.h` line 82 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-82:-NL8R8tp-tmKRIy_0wAd:bimuk21) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L82)):*

I don't believe there is any guarantee about the moved-from state of a `multimap`, so you're better off with a `swap` here.
</blockquote></details>

Swapped to `swap`
___
*[`google/cloud/internal/timer_queue.h` line 84 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-84:-NL8RBL7-IeEQ7x5D3jM:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L84)):*

s/kv/timer/ perhaps?
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 92 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-92:-NL8RDJ84QBeOAf9BFwb:b-2pgfhl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L92)):*

The value returned is somewhat meaningless until `Shutdown()` is called, so an alternative (if you think you still need it) would be to return the count from `Shutdown()`.
</blockquote></details>

Removed counter.
___
*[`google/cloud/internal/timer_queue.h` line 97 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-97:-NL8RF3X-6qZKDKWDPNM:b-7ct19x) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L97)):*

With the changes suggested above, you can remove this check.
</blockquote></details>

Check removed.
___
*[`google/cloud/internal/timer_queue.h` line 121 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-121:-NL8RGlO4n77Fu76HBSh:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L121)):*

Change the signature to:

```
TimePoint NextExpiration(std::unique_lock<std::mutex> const& lk)
```

</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 122 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-122:-NL8RKuH7Y5_vA4HMM-6:b-ihcefx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L122)):*

The arbitrary 1h is an indication that things might be better.  For example, consider calling `cv_.wait()` instead when `timers_` is empty.
</blockquote></details>

Not sure what value kMaxSleep should have or if we should refactor to use `wait`.
___
*[`google/cloud/internal/timer_queue_test.cc` line 33 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-33:-NL8RctDDBpDza0kFi1o:b-9ptgwe) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L33)):*

See my other comments though, it still might be useful.
</blockquote></details>

Left the template parameter for now in case it's useful in the future.
___
*[`google/cloud/internal/timer_queue_test.cc` line 43 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-43:-NL8RsJG2xt6yzdqFT-Q:b9cnp1d) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L43)):*

Shouldn't the yielded time point be exactly the scheduled time point?
</blockquote></details>

Avoiding being too specific to avoid flaky tests.
___
*[`google/cloud/internal/timer_queue_test.cc` line 46 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-46:-NL8S-Dw4-Z3BmOu0phQ:bgw04ot) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L46)):*


```
// Verify new timers with earlier expiration are handled correctly:
using std::chrono_literals; // for exposition only
TimerQueue<...> tq;

auto t100 = tq.Schedule(now + 100ms);
auto t50 = tq.Schedule(now + 50ms);

// Wait around 50ms
EXPECT_TRUE(t50.is_ready());
EXPECT_FALSE(t100.is_ready());

// Wait for 50ms more
EXPECT_TRUE(t100.is_ready());
```

This is the kind of thing where having a mockable clock is useful, but be aware that `std::condition_variable` probably uses low-level POSIX threads APIs where the mocking will have no effect (i.e. it might call [pthread_cond_timedwait()](https://pubs.opengroup.org/onlinepubs/009604599/functions/pthread_cond_timedwait.html))

</blockquote></details>

Added test using real clock.
___
*[`google/cloud/internal/timer_queue_test.cc` line 59 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-59:-NL8SPmO13nbPC6HVsin:b6ve07u) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L59)):*

OK (60s is a long time), but this seems subject to scheduling.  How can we avoid that?
</blockquote></details>

Scheduled and cancelled before starting runners.
___
*[`google/cloud/internal/timer_queue_test.cc` line 76 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-76:-NL8SThx14IMlF4-aXDh:b802lsb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L76)):*

You could avoid adding a test-only parameter to `Service()` by writing the test using the following outline:

```
TEST(...) {
  TimerQueue<std::chrono::system_clock> tq;
  // Schedule the timers first, so they are all ready to expire
  auto constexpr kTimers = 100;
  std::vector<future<std::thread::id>> futures(kTimers);
  std::generate(futures.begin(), futures.end(), [&tq]() { return tq.MakeRelativeTimer(2ms).then([](auto) { return std::this_thread::get_id(); });

  auto constexpr kRunners = 8;
  std::vector<std::thread> runners;
  for (auto i = 0; i != kRunners; ++i) runners.emplace_black([&]() { qt.Service(); }); // maybe use std::generate()

  std::vector<std::thread::id>> ids(
  for (auto& t : runners) t 
  std::generate(runners.begin(), runners.end(), [&tq]() { return std::thread([]
  
  std::set<std::thread> ids;
  std::transform(futures.begin(), futures.end(), std::inserter(ids, ids.begin()), [](auto& f) { return f.get(); });

  tq.Shutdown();
  for (auto& t : runners) t.join(); // BTW, there is a joinable thread class in testing_util/scoped_thread.h

  EXPECT_GT(ids.size(), 1);
}
```

Basically you use the `.then()` callback to find out what thread was used to expire each timer.

</blockquote></details>

Refactored using `.then()` and removed `std::function` parameter.
___
*[`google/cloud/internal/timer_queue_test.cc` line 97 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-97:-NL8S_qo8aq_JUECUMEb:b-o48zlt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L97)):*

How do you know they won't?
</blockquote></details>

By collecting the thread ids of the runners that expire the timers.
___
*[`google/cloud/internal/timer_queue_test.cc` line 127 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-127:-NL8SgN86TsB6PjpQi0G:b-ozrbs9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L127)):*

How do you know some don't expire before they're cancelled?
</blockquote></details>

Counter removed.

devbww (commented):
> Reviewed 3 of 5 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499)* status: all files reviewed, 29 unresolved discussions (waiting on @coryan and @scotthart)

___
*[`google/cloud/internal/timer_queue.h` line 51 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-51:-NL9mQcc8ArVu5nnVa-k:b-1dppox) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L51)):*

Added check before calling notify_all
</blockquote></details>

Note that we're lucky here that `multimap` guarantees (since C++11) that equal elements will be in insertion order, so a new equal-to-minimum timer won't cause a spurious wakeup.
___
*[`google/cloud/internal/timer_queue.h` line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-63:-NL9sUPcCMRoGGKPsk9B:b-jlyzmt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L63)):*

Added comment.
</blockquote></details>

OK, although I was thinking of something a little more prescriptive, like, "If you create a `TimerQueue` you must call `Service()` from at least one thread. Those calls will return when the queue is `Shutdown()`."
___
*[`google/cloud/internal/timer_queue.h` line 121 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-121:-NLA3CE02GS3VXI-HFIN:b-qj3hwr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L121)):*

Done.
</blockquote></details>

Is the intent here and with `HasExpiredTimer()` (to add an unused `unique_lock&`) just to document that the function is called with the lock held?  If so, is that clear?  is it worth it?
___
*[`google/cloud/internal/timer_queue.h` line 122 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-122:-NLA4hnE0ZsO1Eldb8Ro:b-fmkr30) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L122)):*

Not sure what value kMaxSleep should have or if we should refactor to use `wait`.
</blockquote></details>

I'm certainly voting for the latter.  There should be no "max sleep".
___
*[`google/cloud/internal/timer_queue.h` line 42 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9lmHu8bPk491T5uWc:-NL9lmHu8bPk491T5uWd:bkbe1nz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L42)):*
> ```objc
>     auto p = PromiseType();
>     auto f = p.get_future();
>     bool should_notify;
> ```

Despite the fact that it is unconditionally reset below, I would suggest also initializing (to `false`) here.
___
*[`google/cloud/internal/timer_queue.h` line 47 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA55AB48pEECEfZ3Ic:-NLA55AB48pEECEfZ3Id:b-yirkz8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L47)):*
> ```objc
>       if (shutdown_)
>         return make_ready_future(StatusOr<TimePoint>(
>             Status(StatusCode::kCancelled, "Queue already shutdown.")));
> ```

s/Queue/TimerQueue/
s/.//
___
*[`google/cloud/internal/timer_queue.h` line 75 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA4FOJ-JL2m_GGElqS:-NLA4FOJ-JL2m_GGElqT:b-hstuxy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L75)):*
> ```objc
>       if (shutdown_) return NotifyShutdown(std::move(lk));
>       if (HasExpiredTimers(lk, Clock::now())) {
>         ExpireTimers(std::move(lk), Clock::now());
> ```

I'd only call `Clock::now()` once for such a pair of calls.



_Code quote:_
```objc
      if (HasExpiredTimers(lk, Clock::now())) {
        ExpireTimers(std::move(lk), Clock::now());
```
___
*[`google/cloud/internal/timer_queue.h` line 88 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA5IDLCdm_rv8vA7yL:-NLA5IDLCdm_rv8vA7yM:b-7kibjn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L88)):*
> ```objc
>     for (auto& timer : timers) {
>       timer.second.set_value(
>           Status(StatusCode::kCancelled, "Timer cancelled."));
> ```

s/.//
___
*[`google/cloud/internal/timer_queue.h` line 93 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9vwWl11rPSZ1aSoWH:-NL9vwWl11rPSZ1aSoWI:b-ag04ex) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L93)):*
> ```objc
> 
>  private:
>   bool HasExpiredTimers(std::unique_lock<std::mutex> const&, TimePoint tp) {
> ```

I'd call this `HasExpiredTimer()` (i.e., singular).  It only looks for one, and the distinction between one and more than one is inconsequential.
___
*[`google/cloud/internal/timer_queue.h` line 94 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9wglP8H1v2a9iluKN:-NL9wglP8H1v2a9iluKO:bcgwr71) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L94)):*
> ```objc
>  private:
>   bool HasExpiredTimers(std::unique_lock<std::mutex> const&, TimePoint tp) {
>     return !timers_.empty() && timers_.begin()->first < tp;
> ```

`timers_.begin()->first <= tp`, yes?

(Aside: I would also use `timers_.front().first` instead, after establishing `!timers_.empty()`.)
___
*[`google/cloud/internal/timer_queue.h` line 96 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9wP4OD-RJxR9Hp_YM:-NL9wP4OD-RJxR9Hp_YN:b-5ss6je) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L96)):*
> ```objc
>     return !timers_.empty() && timers_.begin()->first < tp;
>   }
>   void ExpireTimers(std::unique_lock<std::mutex> lk, TimePoint tp) {
> ```

Similarly, I'd call it `ExpireTimer()` as it only expires one.
___
*[`google/cloud/internal/timer_queue.h` line 97 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA60hI-8L48yPCX0Dr:-NLA60hI-8L48yPCX0Ds:b-7pquw6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L97)):*
> ```objc
>   }
>   void ExpireTimers(std::unique_lock<std::mutex> lk, TimePoint tp) {
>     auto p = std::move(timers_.begin()->second);
> ```

Similarly, I'd use `timers_.front()` when I've already established `!timers_.empty()`.
___
*[`google/cloud/internal/timer_queue.h` line 100 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9vfmGCh8CsP_RhxZO:-NL9vfmGCh8CsP_RhxZP:b9vryhb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L100)):*
> ```objc
>     timers_.erase(timers_.begin());
>     auto const has_expired_timers =
>         !timers_.empty() && timers_.begin()->first < tp;
> ```

Call `HasExpiredTimer()` instead of repeating it's body.



_Code quote:_
```objc
    auto const has_expired_timers =
        !timers_.empty() && timers_.begin()->first < tp;
```
___
*[`google/cloud/internal/timer_queue.h` line 105 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA1DFs6BycFD9w7hiH:-NLA1DFs6BycFD9w7hiI:blprf66) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L105)):*
> ```objc
>     // serialize the expiration.  Instead, we pick another thread to continue
>     // expiring timers.
>     if (has_expired_timers) cv_.notify_one();
> ```

Note that this is redundant as all the other threads will already have been made runnable.  While we may want something similar to this in the long run (#10512), at the moment it is either confusing to the reader, or makes it look like we don't understand the control flow.
___
*[`google/cloud/internal/timer_queue.h` line 112 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9vQL-4UDsBpSN3_gM:-NL9vQL-4UDsBpSN3_gN:b-ct1lw3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L112)):*
> ```objc
>     auto timers = std::move(timers_);
>     lk.unlock();
>     for (auto& kv : timers) {
> ```

s/kv/timer/
___
*[`google/cloud/internal/timer_queue.h` line 114 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA5gfC6pGcb8jFb4Z3:-NLA5gfC6pGcb8jFb4Z4:bgyb2lt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L114)):*
> ```objc
>     for (auto& kv : timers) {
>       kv.second.set_value(
>           Status(StatusCode::kCancelled, "Timer queue shutdown."));
> ```

s/Timer queue/TimerQueue/
s/.//
___
*[`google/cloud/internal/timer_queue_test.cc` line 43 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-43:-NLA70skDaJBE5XT2tjH:bctoun5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L43)):*

Avoiding being too specific to avoid flaky tests.
</blockquote></details>

I don't think this concerns flakiness, but rather the specification of what `t2 = *tq.Schedule(t1).get()` means.
___
*[`google/cloud/internal/timer_queue_test.cc` line 97 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-97:-NLADOvM3yABDbPxn5H_:b-9xv29p) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L97)):*

By collecting the thread ids of the runners that expire the timers.
</blockquote></details>

This was previously attached to a "Verify expiring timers are not handled serially by a single runner" expectation, which has been refactored.  I've moved the discussion to the new location.
___
*[`google/cloud/internal/timer_queue_test.cc` line 31 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLAAS3f8UO_2n2usaBX:-NLAAS3f8UO_2n2usaBY:b6ufu8j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L31)):*
> ```cpp
> using testing::Each;
> 
> TEST(TimerQueueTest, ScheduleSingleRunner) {
> ```

If we want to stick with the "runner" term, perhaps we should change `Service()` to `Run()`.  That would also align it with the similar concept in a completion queue.
___
*[`google/cloud/internal/timer_queue_test.cc` line 41 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA8Yow39CzO6OnsAGa:-NLA8Yow39CzO6OnsAGb:bn3oq1n) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L41)):*
> ```cpp
>   t.join();
> 
>   EXPECT_THAT(expire_time, IsOk());
> ```

ASSERT_THAT
___
*[`google/cloud/internal/timer_queue_test.cc` line 81 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA9qaHCpDEZvDb5x59:-NLA9qaI7rrVZpaPCSUg:b2gvfsv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L81)):*
> ```cpp
>   ASSERT_THAT(earlier_expire_time, IsOk());
>   ASSERT_THAT(later_expire_time, IsOk());
>   EXPECT_GT(*later_expire_time, *earlier_expire_time);
> ```

As above, I reckon we should have tight expectations on these values.
___
*[`google/cloud/internal/timer_queue_test.cc` line 108 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLACCbs7pWouLHSWxx1:-NLACCbs7pWouLHSWxx2:b-gop38v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L108)):*
> ```cpp
>   tq.Shutdown();
>   for (auto& t : runners)
>     t.join();  // BTW, there is a joinable thread class in
> ```

Either use the class, or drop the comment?



_Code quote:_
```cpp
    t.join();  // BTW, there is a joinable thread class in
               // testing_util/scoped_thread.h
```
___
*[`google/cloud/internal/timer_queue_test.cc` line 111 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLABvoDE3dddtYl82Vw:-NLABvoE4lUpaGNEOmbc:bjqh19v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L111)):*
> ```cpp
>                // testing_util/scoped_thread.h
> 
>   EXPECT_GT(ids.size(), 1);
> ```

Previously I asked how you know all the timers weren't expired by a single thread, but I didn't grok the "By collecting the thread ids of the runners that expire the timers" answer.  It seems to me that there is no such guarantee, despite there being multiple runners.
___
*[`google/cloud/internal/timer_queue_test.cc` line 119 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLAFWkZ1BwgZMHCOYuA:-NLAFWkZ1BwgZMHCOYuB:bhn7lno) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L119)):*
> ```cpp
>   auto now = std::chrono::system_clock::now();
>   std::vector<future<StatusOr<std::chrono::system_clock::time_point>>> futures;
>   futures.reserve(100);
> ```

I suggest dropping `reserve()` calls from non-performance-critical code.
___
*[`google/cloud/internal/timer_queue_test.cc` line 140 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLAGFk_1pau_36eVC3P:-NLAGCAeDf8GGj03_Vz2:b-ons8h7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L140)):*
> ```cpp
> }
> 
> TEST(TimerQueueTest, ScheduleEarlierTimerMultipleRunner) {
> ```

The similarity between this and `ScheduleEarlierTimerSingleRunner` emphasizes that the observable behavior of a `TimerQueue` should not depend on the number of runners (modulo a `set_value()` not returning).  Therefore, I reckon that just the "single" test cases are sufficient.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499)* status: all files reviewed, 30 unresolved discussions (waiting on @coryan and @scotthart)

___
*[`google/cloud/internal/timer_queue.h` line 80 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLCl7nT79CiA9N-Zb8r:-NLCl7nT79CiA9N-Zb8s:b-fbgpvp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L80)):*
> ```objc
>   }
> 
>   void CancelAll() {
> ```

What's the use case for `CancelAll()`?
___
*[`google/cloud/internal/timer_queue_test.cc` line 33 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-33:-NLCjUg0BPuLfUCNRJcJ:b-c5wbdg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L33)):*

Left the template parameter for now in case it's useful in the future.
</blockquote></details>

I would argue that the `Clock` parameter is useless (and an attractive nuisance) when you can't affect the clock used by `condition_variable::wait_until()`, as I think Carlos mentioned elsewhere.  CppReference says, "existing implementations convert `timeout_time` from `Clock` to `std::chrono::system_clock` and delegate to POSIX `pthread_cond_timedwait` so that the wait honors adjustments to the system clock, but not to the user-provided `Clock`."  So, best if we also just used `std::chrono::system_clock`, methinks.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499)* status: all files reviewed, 31 unresolved discussions (waiting on @coryan and @scotthart)

___
*[`google/cloud/internal/timer_queue.h` line 68 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLFEP6B3bip8T5A7Jay:-NLFEP6B3bip8T5A7Jaz:b-xome7c) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L68)):*
> ```objc
>       auto predicate = [this, until, &lk] {
>         auto const ne = NextExpiration(lk);
>         return shutdown_ || ne < until || HasExpiredTimers(lk, Clock::now());
> ```

It strikes me that the `ne < until` condition should be narrower.  If `ne < until` then an earlier timer has been installed, and we should indeed terminate the `wait_until()`, and start a new one with a new deadline.

But if `ne > until` then the timer we were awaiting has either been expired by another thread, or it has been cancelled.  In either case we should also immediately terminate the `wait_until()`, and start a new one with a new deadline.

So, use `ne != until` instead.

And indeed, from there, we don't even need `HasExpiredTimers()`.  We have an expired timer when the wait terminates iff `NextExpiration() == until`.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499)* status: all files reviewed, 31 unresolved discussions (waiting on @coryan and @scotthart)

___
*[`google/cloud/internal/timer_queue.h` line 68 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLFEP6B3bip8T5A7Jay:-NLHOBMY1VhcnC6f1ef2:b-i2wnjh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L68)):*

It strikes me that the `ne < until` condition should be narrower.  If `ne < until` then an earlier timer has been installed, and we should indeed terminate the `wait_until()`, and start a new one with a new deadline.

But if `ne > until` then the timer we were awaiting has either been expired by another thread, or it has been cancelled.  In either case we should also immediately terminate the `wait_until()`, and start a new one with a new deadline.

So, use `ne != until` instead.

And indeed, from there, we don't even need `HasExpiredTimers()`.  We have an expired timer when the wait terminates iff `NextExpiration() == until`.
</blockquote></details>

There is something wrong with that last bit.  The possibility of a spurious wakeup means we'd also need `until <= now()` to have an expired timer.  But perhaps the return value from `wait_until()` could still be used to avoid the need for `Clock`.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499)* status: all files reviewed, 30 unresolved discussions (waiting on @coryan and @scotthart)

___
*[`google/cloud/internal/timer_queue.h` line 94 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9wglP8H1v2a9iluKN:-NLIKXRK9ja5V_J6B7X3:b-qu74t1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L94)):*

`timers_.begin()->first <= tp`, yes?

(Aside: I would also use `timers_.front().first` instead, after establishing `!timers_.empty()`.)
</blockquote></details>

Pardon me ... `multimap`, of course, doesn't have a `front()`.

coryan (approved):
I think @devbww is right.  I volunteer to address those comments in future PRs.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-:-NLLud9p6235KRsqDk9x:bo52pof)* status: all files reviewed, 27 unresolved discussions (waiting on @scotthart)

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499)* status: all files reviewed, 18 unresolved discussions (waiting on @coryan and @devbww)

___
*[`google/cloud/internal/timer_queue.h` line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-63:-NLMZwYBCyJ-CH9EqiqY:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L63)):*

OK, although I was thinking of something a little more prescriptive, like, "If you create a `TimerQueue` you must call `Service()` from at least one thread. Those calls will return when the queue is `Shutdown()`."
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 121 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-121:-NLMblrk0mreJsEyosRy:b-kaini) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L121)):*

Is the intent here and with `HasExpiredTimer()` (to add an unused `unique_lock&`) just to document that the function is called with the lock held?  If so, is that clear?  is it worth it?
</blockquote></details>

Added function comment.
___
*[`google/cloud/internal/timer_queue.h` line 122 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-122:-NLMcIX45sXM8_JhGwx5:b-pi4klg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L122)):*

I'm certainly voting for the latter.  There should be no "max sleep".
</blockquote></details>

created issue
___
*[`google/cloud/internal/timer_queue.h` line 42 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9lmHu8bPk491T5uWc:-NLMZrpC3n4gonXtHP8J:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L42)):*

Despite the fact that it is unconditionally reset below, I would suggest also initializing (to `false`) here.
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 47 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA55AB48pEECEfZ3Ic:-NLMZszM4Zi9PcOdMF6s:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L47)):*

s/Queue/TimerQueue/
s/.//
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 68 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLFEP6B3bip8T5A7Jay:-NLMaMh-0t4yhsDjjUA0:bq11zjt) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L68)):*
> It strikes me that the ne < until condition should be narrower.  If ne < until then an earlier timer has been installed, and we should indeed terminate the wait_until(), and start a new one with a new deadline.
> But if ne > until then the timer we were awaiting has either been expired by another thread, or it has been cancelled.  In either case we should also immediately terminate the wait_until(), and start a new one with a new deadline.
> So, use ne != until instead.
> And indeed, from there, we don't even need HasExpiredTimers().  We have an expired timer when the wait terminates iff NextExpiration() == until.
> There is something wrong with that last bit.  The possibility of a spurious wakeup means we'd also need until <= now() to have an expired timer.  But perhaps the return value from wait_until() could still be used to avoid the need for Clock.

created issue #10521
___
*[`google/cloud/internal/timer_queue.h` line 75 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA4FOJ-JL2m_GGElqS:-NLMaY9-9DnuODkftzcu:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L75)):*

I'd only call `Clock::now()` once for such a pair of calls.
</blockquote></details>

Done
___
*[`google/cloud/internal/timer_queue.h` line 80 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLCl7nT79CiA9N-Zb8r:-NLMbL864N770iC3X9tk:bamq1kr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L80)):*

What's the use case for `CancelAll()`?
</blockquote></details>

Provides a mechanism for a more graceful shutdown of the queue without requiring it.
___
*[`google/cloud/internal/timer_queue.h` line 88 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA5IDLCdm_rv8vA7yL:-NLMb_cj7HRZJF9DRJQu:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L88)):*

s/.//
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 93 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9vwWl11rPSZ1aSoWH:-NLMbbIf78z3EmTp4-r8:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L93)):*

I'd call this `HasExpiredTimer()` (i.e., singular).  It only looks for one, and the distinction between one and more than one is inconsequential.
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 96 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9wP4OD-RJxR9Hp_YM:-NLMbeFn1wK2hAdYXrE9:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L96)):*

Similarly, I'd call it `ExpireTimer()` as it only expires one.
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 100 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9vfmGCh8CsP_RhxZO:-NLMbg8r0DC_A0eLVbpT:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L100)):*

Call `HasExpiredTimer()` instead of repeating it's body.
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 112 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9vQL-4UDsBpSN3_gM:-NLMbiyCAyHanAyD46Pr:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L112)):*

s/kv/timer/
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue.h` line 114 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA5gfC6pGcb8jFb4Z3:-NLMbjnE1_3xPMII42dD:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L114)):*

s/Timer queue/TimerQueue/
s/.//
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue_test.cc` line 33 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8J-r1-33:-NLMcUh68b36btl4PDAs:b-5qwndm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue_test.cc#L33)):*

I would argue that the `Clock` parameter is useless (and an attractive nuisance) when you can't affect the clock used by `condition_variable::wait_until()`, as I think Carlos mentioned elsewhere.  CppReference says, "existing implementations convert `timeout_time` from `Clock` to `std::chrono::system_clock` and delegate to POSIX `pthread_cond_timedwait` so that the wait honors adjustments to the system clock, but not to the user-provided `Clock`."  So, best if we also just used `std::chrono::system_clock`, methinks.
</blockquote></details>

template parameter removed.
___
*[`google/cloud/internal/timer_queue_test.cc` line 31 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLAAS3f8UO_2n2usaBX:-NLMcMDUDra-pNI2HzH9:b-4yxxtu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L31)):*

If we want to stick with the "runner" term, perhaps we should change `Service()` to `Run()`.  That would also align it with the similar concept in a completion queue.
</blockquote></details>

I have no bike in this bike shed.
___
*[`google/cloud/internal/timer_queue_test.cc` line 41 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA8Yow39CzO6OnsAGa:-NLMcWQb2omxIHonrB6H:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L41)):*

ASSERT_THAT
</blockquote></details>

Done.
___
*[`google/cloud/internal/timer_queue_test.cc` line 81 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA9qaHCpDEZvDb5x59:-NLMdZboEFMoRSA7ocBI:bbeaku0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L81)):*

As above, I reckon we should have tight expectations on these values.
</blockquote></details>

I'm unsure what, if any, the tolerances should be.
___
*[`google/cloud/internal/timer_queue_test.cc` line 108 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLACCbs7pWouLHSWxx1:-NLMciqN5NIHPeHxbIak:b-y4p0xy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L108)):*

Either use the class, or drop the comment?
</blockquote></details>

removed comment
___
*[`google/cloud/internal/timer_queue_test.cc` line 111 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLABvoDE3dddtYl82Vw:-NLMdIRJ4I2bb-yyPiaH:b-njdbxs) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L111)):*

Previously I asked how you know all the timers weren't expired by a single thread, but I didn't grok the "By collecting the thread ids of the runners that expire the timers" answer.  It seems to me that there is no such guarantee, despite there being multiple runners.
</blockquote></details>

Perhaps we can't guarantee it with the current implementation, but as long as avoiding serial expiration of timers is a goal, I think we should test for it.
___
*[`google/cloud/internal/timer_queue_test.cc` line 119 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLAFWkZ1BwgZMHCOYuA:-NLMcpAKAoea-K4iIRB_:bkg663w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L119)):*

I suggest dropping `reserve()` calls from non-performance-critical code.
</blockquote></details>

I had to keep one of them to satisfy clang-tidy
___
*[`google/cloud/internal/timer_queue_test.cc` line 140 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLAGFk_1pau_36eVC3P:-NLMcunQ84hItYQoRKdY:boacer6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L140)):*

The similarity between this and `ScheduleEarlierTimerSingleRunner` emphasizes that the observable behavior of a `TimerQueue` should not depend on the number of runners (modulo a `set_value()` not returning).  Therefore, I reckon that just the "single" test cases are sufficient.
</blockquote></details>

Possibly, but I found some bugs only after I added the multiple thread tests, so I'd prefer to keep them.

devbww (commented):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499)* status: all files reviewed, 14 unresolved discussions

___
*[`google/cloud/internal/timer_queue.h` line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL3MOo2L7GiIO-ZVf8I-r1-63:-NLN08kHDi7wnRMNPtTg:bzfl3y9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/765a5e18fc7fb21dd2b6546157ff0954573c1280/google/cloud/internal/timer_queue.h#L63)):*

Done.
</blockquote></details>

In hindsight, it would probably be good to also mention that you must wait for those `Service()` calls to return before destroying the `TimerQueue`.
___
*[`google/cloud/internal/timer_queue.h` line 80 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLCl7nT79CiA9N-Zb8r:-NLN0hkh4G_-xp3DX-M5:b-mvmaov) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L80)):*

Provides a mechanism for a more graceful shutdown of the queue without requiring it.
</blockquote></details>

I'm not sure what "a more graceful shutdown" means, but I was more looking for the circumstances in which you intend to call `CancelAll()`.

If the canceler is (synchronized with) the scheduler then it seems like you could achieve the same thing with recreation.  If they're not synchronized then it seems like a race condition.
___
*[`google/cloud/internal/timer_queue.h` line 94 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NL9wglP8H1v2a9iluKN:-NLN218L5av4BDHSJuje:b-8m1xnd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue.h#L94)):*

Pardon me ... `multimap`, of course, doesn't have a `front()`.
</blockquote></details>

I would still consider "first == tp" as expired.
___
*[`google/cloud/internal/timer_queue_test.cc` line 81 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLA9qaHCpDEZvDb5x59:-NLN2ruDBzkr5PpBfT1Z:bhruot3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L81)):*

I'm unsure what, if any, the tolerances should be.
</blockquote></details>

As mentioned in another thread, I don't think this concerns tolerances, but rather the specification of what `t2 = *tq.Schedule(t1).get()` means.  I'm working under that expectation that `t2 == t1`.  That is, a timer yields it scheduled time.
___
*[`google/cloud/internal/timer_queue_test.cc` line 111 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLABvoDE3dddtYl82Vw:-NLN3T0jEesz6t-FnVKq:bebd32j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L111)):*

Perhaps we can't guarantee it with the current implementation, but as long as avoiding serial expiration of timers is a goal, I think we should test for it.
</blockquote></details>

If we don't guarantee it, it doesn't sound like something we should test.  It's just flaky.
___
*[`google/cloud/internal/timer_queue_test.cc` line 140 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10499#-NLAGFk_1pau_36eVC3P:-NLN3qqmCGFMOVqWpGlA:byqrnca) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8007f9ddba19d8f4bc1c012899953f48a3d8991c/google/cloud/internal/timer_queue_test.cc#L140)):*

Possibly, but I found some bugs only after I added the multiple thread tests, so I'd prefer to keep them.
</blockquote></details>

Then perhaps we should only keep the multi-thread tests?

Issue #10487, opened 2023-01-04
Synchronous OpenTelemetry tracing in common


dbolduc:
Some tasks from [Milestone 1](https://docs.google.com/document/d/1JTOWvZD59AGxojTV93tKZJ9b_q2w8YUMMyuxCu5_ZC4/edit?resourcekey=0-0XjX6fh2xNKFZN20cu2otg#bookmark=id.i1nfe2c9gho2) of an internal design doc. With these tasks completed, we can support tracing a synchronous unary client call. (the whole call, not the subcomponents of that call).

- [x] Add experimental tracing option; `TracingEnabled()`
- [x] Add `GetTracer()`, `MakeSpan()`
- [x] Add `CaptureReturn()` for status types
- [x] Add `CallContext` (current span + current options)
- [ ] Add traced `StreamRange<>` wrapper

PR #10501, opened 2023-01-06
impl(otel): add `MakeSpan()`

dbolduc:
Part of the work for #10487 

Add `GetTracer()` and `MakeSpan()` methods. `GetTracer` uses the global tracer provider for now.

OpenTelemetry offers an in-memory exporter that gives us access to the spans. The set up requires a few steps, so capture that in a dedicated test fixture. The test fixture and some span matchers are factored out in `testing_util`, so that they can be reused by the golden unit tests for the tracing decorators.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10501)


google/cloud/testing_util/opentelemetry.h
```diff
@@ -0,0 +1,58 @@
...
+::testing::Matcher<SpanDataPtr> SpanKindIsClient();
+
+::testing::Matcher<SpanDataPtr> SpanNamed(std::string const& name);
+
+// A test fixture that provides access to created spans.
+class OpenTelemetryTest : public ::testing::Test {
+ protected:
+  OpenTelemetryTest();
+
+  std::shared_ptr<opentelemetry::exporter::memory::InMemorySpanData> span_data_;
```
coryan, line 49: I think a standalone function (returning `std::shared_ptr<...::InMemorySpanData>`) is a cleaner start.

Creating a test fixture, just to initialize one member variable, yields less readable tests over time.

google/cloud/testing_util/opentelemetry.h
```diff
@@ -0,0 +1,58 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_TESTING_UTIL_OPENTELEMETRY_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_TESTING_UTIL_OPENTELEMETRY_H
+
+#ifdef GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
+#include "google/cloud/version.h"
+#include <gmock/gmock.h>
+#include <opentelemetry/exporters/memory/in_memory_span_exporter.h>
```
coryan, line 21: Are you certain you need all these `#includes` in the header?

google/cloud/testing_util/CMakeLists.txt
```diff
@@ -37,6 +37,8 @@ add_library(
     mock_backoff_policy.h
     mock_fake_clock.cc
     mock_fake_clock.h
+    opentelemetry.cc
```
coryan, line 40: maybe `opentelemetry_matchers.{h,cc}` that name is bound to conflict with something else?

coryan (changes requested):

google/cloud/testing_util/CMakeLists.txt
dbolduc: Updated.

google/cloud/testing_util/opentelemetry.h
dbolduc: I am certain I don't. Fixed.

google/cloud/testing_util/opentelemetry.h
dbolduc: Fixed.

google/cloud/internal/opentelemetry.h
```diff
@@ -26,6 +30,12 @@ namespace internal {
 
 bool TracingEnabled(Options const& options);
 
+opentelemetry::nostd::shared_ptr<opentelemetry::trace::Tracer> GetTracer(
+    Options const& options);
+
+opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> MakeSpan(
```
devbww, line 36: Forgive my continuing naivety, but is a span something that should always (usually?) be stack based?

dbolduc: The API to make a span returns it in a smart pointer: https://github.com/open-telemetry/opentelemetry-cpp/blob/8a9d085f05b04a2585390c1398f76c388cbcff0e/api/include/opentelemetry/trace/tracer.h#L40

So I think we should just pass that pointer around.

dbolduc:
PTAL

coryan (approved):

google/cloud/testing_util/opentelemetry_matchers.cc
```diff
@@ -0,0 +1,73 @@
...
+namespace testing_util {
+
+using ::testing::AllOf;
+using ::testing::Matcher;
+using ::testing::Pointee;
+using ::testing::Property;
+
+Matcher<SpanDataPtr> SpanHasInstrumentationScope() {
+  return Pointee(
+      Property(&opentelemetry::sdk::trace::SpanData::GetInstrumentationScope,
```
devbww, line 36: Can we use a lambda instead of `Property` and so avoid taking the address of a foreign symbol?

dbolduc: http://google.github.io/googletest/gmock_cook_book.html#validating-a-member-of-an-object

The gMock Cookbook agrees: "Don’t use `Property()` against member functions that you do not own, because taking addresses of functions is fragile and generally not part of the contract of the function."

I will work on it....

devbww: One suggestion I saw was ...

> if you currently have `Property(&Foo::bar, SubMatcher())`, you can replace it with `BarIs(SubMatcher())` where BarIs is defined by a matcher macro like
> ```
> MATCHER_P(BarIs, m) { return testing::ExplainMatchResult(m, arg.bar(), result_listener); }
> ```

dbolduc: Thanks, I was kind of stuck, and your suggestion gave me some direction.

I decided not to use `::testing::ExplainMatchResult` because it would just dump the `arg` object, not `arg.bar()`. So I manually added in the relevant information. This is what an error looks like: (scroll to the end)

```
google/cloud/internal/opentelemetry_test.cc:88: Failure
Value of: spans[0]
Expected: span named (name: "darren")
  Actual: (ptr = 0x556157162fc0, value = 304-byte object <90-CB D5-55 61-55 00-00 7B-FE C7-3F 52-00 E8-C7 3F-9D 8B-71 D4-B9 09-9C A6-13 00-2F 6D-A0 76-E5 01-00 00-00 00-00 00-00 98-CF D5-55 61-55 00-00 00-1E 16-57 61-55 00-00 C0-1D 16-57 61-55 00-00 ... 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 02-00 00-00 00-00 00-00 60-35 16-57 61-55 00-00 00-27 16-57 61-55 00-00>), span name: span1)
```

google/cloud/testing_util/opentelemetry_matchers.h
```diff
@@ -0,0 +1,104 @@
...
+    switch (k) {
+      case opentelemetry::trace::SpanKind::kInternal:
+        return "INTERNAL";
+      case opentelemetry::trace::SpanKind::kServer:
+        return "SERVER";
+      case opentelemetry::trace::SpanKind::kClient:
+        return "CLIENT";
+      case opentelemetry::trace::SpanKind::kProducer:
+        return "PRODUCER";
+      case opentelemetry::trace::SpanKind::kConsumer:
```
devbww, line 63: Why doesn't `kConsumer` map to, say, "CONSUMER"?

google/cloud/testing_util/opentelemetry_matchers.cc
devbww: Is there any opportunity to unify the shapes of `span named (name: "darren")` and `span name: span1`?

PR #10506, opened 2023-01-06
doc(pubsub): use real Doxygen groups

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10506)

### DIFF
--- a/google/cloud/pubsub/schema.h
+++ b/google/cloud/pubsub/schema.h
@@ -41,12 +41,12 @@ class Schema {
       : project_id_(std::move(project_id)), schema_id_(std::move(schema_id)) {}
 
   /// @name Copy and move
-  //@{
+  ///@{
   Schema(Schema const&) = default;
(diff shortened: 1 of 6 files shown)


google/cloud/pubsub/schema.h
```diff
@@ -41,12 +41,12 @@ class Schema {
       : project_id_(std::move(project_id)), schema_id_(std::move(schema_id)) {}
 
   /// @name Copy and move
```
devbww, line 43: Q. Is there any difference between the `@name` being inside or outside the group?

devbww (approved):

google/cloud/pubsub/schema.h
coryan: I believe not.  I think we need to pick one style to simplify global changes, probably "inside the group", but not in this series of PRs.

PR #10514, opened 2023-01-06
chore: ignore more directories

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10514)

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -4,3 +4,4 @@ cmake-out/
 cmake-build-debug/
 cmake-build-coverage/
 cmake-build-release/
+.build/


.bazelignore
```diff
@@ -4,3 +4,4 @@ cmake-out/
 cmake-build-debug/
 cmake-build-coverage/
 cmake-build-release/
```
dbolduc, line 6: I think these can be deleted now. (and from `.dockerignore` and `.gitignore`)

```sh
$ grep "cmake-build" $(git ls-files)
.bazelignore:cmake-build-debug/
.bazelignore:cmake-build-coverage/
.bazelignore:cmake-build-release/
.dockerignore:cmake-build-*/
.gitignore:cmake-build-*/
```

dbolduc (approved):

.bazelignore
coryan: No.  CLion creates those, and some of us use CLion.

dbolduc: Ack. TIL.

Issue #10491, opened 2023-01-04
OpenTelemetry tracing decorators in generator


dbolduc:
This issue tracks initial support only, corresponding to [Milestone 1](https://docs.google.com/document/d/1JTOWvZD59AGxojTV93tKZJ9b_q2w8YUMMyuxCu5_ZC4/edit?resourcekey=0-0XjX6fh2xNKFZN20cu2otg#bookmark=id.i1nfe2c9gho2) of an internal design doc. I will open separate issue(s) for tracing other types of client calls.

- [x] Add tracing connection to the generator
  - [x] initially, this will just pass through to the child
  - [x] implement tracing for sync unary methods
  - [x] include tracing connection decorator in `MakeConnection` factory function
- [ ] Add tracing stub to the generator
  - [x] initially, this will just pass through to the child
  - [x] implement tracing for sync unary methods
  - [ ] include tracing stub decorator in stub factory

PR #10526, opened 2023-01-09
impl(generator): add (unimplemented) tracing connection

dbolduc:
Part of the work for #10491

Add a tracing connection that just passes through to its child connection. I don't return `kUnimplemented`, because I want a partially implemented tracing connection decorator to still be functional.

I will later introduce a `MakeFooTracingConnnection(...)` factory function, which will be declared outside of the `#ifdef` guards.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10526)


generator/generator_test.cc
```diff
@@ -141,7 +141,7 @@ TEST_F(GeneratorTest, BadCommandLineArgs) {
 }
 
 TEST_F(GeneratorTest, GenerateServicesSuccess) {
-  int constexpr kNumMockOutputStreams = 23;
+  int constexpr kNumMockOutputStreams = 25;
```
devbww, line 144: That looks unfortunate.

generator/internal/tracing_connection_generator.h
```diff
@@ -0,0 +1,63 @@
...
+  Status GenerateCc() override;
+
+  static std::string MethodDeclaration(
+      google::protobuf::MethodDescriptor const& method);
+  static std::string AsyncMethodDeclaration(
+      google::protobuf::MethodDescriptor const& method);
+  static std::string MethodDefinition(
+      google::protobuf::MethodDescriptor const& method);
+  static std::string AsyncMethodDefinition(
+      google::protobuf::MethodDescriptor const& method);
```
devbww, line 56: Do these need to be members?

generator/integration_tests/golden/v1/internal/golden_kitchen_sink_tracing_connection.cc
```diff
@@ -0,0 +1,90 @@
+// Copyright 2022 Google LLC
```
devbww, line 1: Where did these years originate?

generator/internal/dern.cc
```diff
@@ -0,0 +1,5 @@
+/// Returns `child` if tracing is not enabled.
```
devbww, line 1: So what?

generator/generator_test.cc
dbolduc: Eh, it's not so bad. I'd rather have the test than not.

generator/internal/tracing_connection_generator.h
dbolduc: No. But it is consistent with what we do for the `ConnectionImpl` https://github.com/googleapis/google-cloud-cpp/blob/9196bdcd6281afb62456b648ca436cdfddaf6241/generator/internal/connection_impl_generator.h#L49-L56

(It's probably the laziness talking, but) I'd rather change neither than both.

generator/integration_tests/golden/v1/internal/golden_kitchen_sink_tracing_connection.cc
dbolduc: https://github.com/googleapis/google-cloud-cpp/blob/9196bdcd6281afb62456b648ca436cdfddaf6241/generator/integration_tests/golden_config.textproto#L21

generator/internal/dern.cc
dbolduc: Gah, thanks for catching this. `dern.*` is a pejorative version of my name that I use for temporary files. I did not mean to commit it. Removed.

generator/integration_tests/golden/v1/internal/golden_kitchen_sink_tracing_connection.cc
devbww: Hmm ... we add new output files to the generator and they appear from the past.  Spooky.

devbww (approved):

PR #10527, opened 2023-01-09
refactor: use CMake functions for Doxygen

coryan:
I think this will make additional changes to support XML output more readable. I am also planning to reduce the number of arguments passed via global variables.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10527)

### DIFF
--- a/cmake/GoogleCloudCppDoxygen.cmake
+++ b/cmake/GoogleCloudCppDoxygen.cmake
@@ -159,8 +159,10 @@ function (google_cloud_cpp_doxygen_targets library)
     )
 endfunction ()
 
-# TODO(#10519) - have each subproject call this directly Find out the name of
-# the subproject.
+# TODO(#10519) - have each subproject call this directly.
+# ~~~
+# Find out the name of the subproject.
+# ~~~
 get_filename_component(GOOGLE_CLOUD_CPP_SUBPROJECT
                        "${CMAKE_CURRENT_SOURCE_DIR}" NAME)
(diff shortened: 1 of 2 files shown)


cmake/GoogleCloudCppDoxygen.cmake
```diff
@@ -97,49 +113,69 @@ elseif (GOOGLE_CLOUD_CPP_GENERATE_DOXYGEN)
...
+        "${CMAKE_CURRENT_SOURCE_DIR}"
+        DEPENDS
+        "cloud-docs"
+        ${GOOGLE_CLOUD_CPP_DOXYGEN_DEPS}
+        TAGFILES
+        "${GOOGLE_CLOUD_CPP_COMMON_TAG}=https://googleapis.dev/cpp/google-cloud-common/${VERSION}/"
+    )
+endfunction ()
+
+# TODO(#10519) - have each subproject call this directly Find out the name of
```
devbww, line 162: s/Find/to find/?

devbww (approved):
> I think this will make additional changes to support XML output more readable.

And "Who am I to disagree".

cmake/GoogleCloudCppDoxygen.cmake
coryan: Argh... cmake-format joined two unrelated comments.  I am going to beat it with the `# ~~~` hammer.  That will teach it.

PR #10528, opened 2023-01-09
impl(common): use leader/followers in `TimerQueue`

### LINKED ISSUE #10512: TimerQueue should only wake one thread instead of all when a timer expires.
`TimerQueue::Service` currently has all threads blocked on on its condition variable using `wait_until`. This causes all threads to attempt to service a single expiring timer. It would be better to only wake one thread.

### LINKED ISSUE #10521: Refactor TimerQueue wait_until predicate
Per review comment:

It strikes me that the ne < until condition should be narrower. If ne < until then an earlier timer has been installed, and we should indeed terminate the wait_until(), and start a new one with a new deadline.

But if ne > until then the timer we were awaiting has either been expired by another thread, or it has been cancelled. In either case we should also immediately terminate the wait_until(), and start a new one with a new deadline.

So, use ne != until instead.

And indeed, from there, we don't even need HasExpiredTimers(). We have an expired timer when the wait terminates iff NextExpiration() == until.

There is something wrong with that last bit. The possibility of a spurious wakeup means we'd also need until <= now() to have an expired timer. But perhaps the return value from wait_until() could still be used to avoid the need for Clock.

### LINKED ISSUE #10522: Refactor TimerQueue kMaxSleep
Determine if there is a better value for kMaxSleep or if we need to remove the mechanism entirely in favor of some sort of wait call to the condition variable.

coryan:
Fixes #10522. Fixes #10512. Fixes #10521.  Probably introduces new bugs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10528)

### DIFF
--- a/google/cloud/internal/timer_queue.cc
+++ b/google/cloud/internal/timer_queue.cc
@@ -13,13 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/internal/timer_queue.h"
-#include "google/cloud/future.h"
 #include "google/cloud/internal/make_status.h"
-#include "google/cloud/status_or.h"
-#include "google/cloud/version.h"
-#include <chrono>
-#include <map>
-#include <mutex>
 
 namespace google {
 namespace cloud {
@@ -43,7 +37,7 @@ future<StatusOr<std::chrono::system_clock::time_point>> TimerQueue::Schedule(
     std::lock_guard<std::mutex> lk(mu_);
     if (shutdown_) {
       return make_ready_future(
-          MakeCancelled("TimerQueue already shutdown", GCP_ERROR_INFO()));
+          MakeCancelled("TimerQueue shutdown", GCP_ERROR_INFO()));
     }
     auto iter = timers_.emplace(tp, std::move(p));
     should_notify = iter == timers_.begin();
@@ -59,31 +53,27 @@ void TimerQueue::Shutdown() {
   cv_follower_.notify_all();
 }
 
+// The threads play two different roles:
+// - A single thread at a time plays the leader role.
+// - All other threads are followers and block on cv_follower_.
+//
+// Once a timer expires, the leader thread relinquishes its role and wakes up
+// one follower thread to become the new leader. Only after doing so it
+// runs the code to expire the timer.
+//
+// This is all complicated by shutdown. Basically all thread needs to wake
+// up when the TQ is shutdown and one of them will expire all the timers.
 void TimerQueue::Service() {
-  // It is easier to understand this if you first ignore shutdown and empty
-  // timer queues.
-  //
-  // The threads play two different roles:
-  // - A single thread at a time plays the leader role.
-  // - All other threads are followers and block on cv_follower_.
-  //
-  // Once a timer expires the leader thread relinquishes its role and wakes up
-  // one follower thread to become the new leader. Only after doing so it
-  // runs the code to expire the timer.
-  //
-  // This is all complicated by shutdown. Basically all thread needs to wake
-  // up when the TQ is shutdown and one of them will expire all the timers.
-  //
-  std::unique_lock<std::mutex> lk(mu_);
+  // The is_leader flag allows us to restart this loop without worrying about
+  // electing new leaders.
   auto is_leader = false;
+  std::unique_lock<std::mutex> lk(mu_);
   while (!shutdown_) {
     if (!is_leader && has_leader_) {
-      // The current thread becomes a follower until no leader is available.
+      // The current thread becomes a follower while there is a leader.
       cv_follower_.wait(lk, [this] { return shutdown_ || !has_leader_; });
       continue;
     }
-    // The is_leader flag allows us to restart this loop without worrying about
-    // electing new leaders.
     is_leader = true;
     has_leader_ = true;
     if (timers_.empty()) {
@@ -92,29 +82,29 @@ void TimerQueue::Service() {
     }
     // Should a new timer appear that changes the "first" timer, we need to
     // wake up and recompute the sleep time. But note that the leader thread
-    // does not need to relinquish that role to do so.
+    // does not need to relinquish its role to do so.
     auto until = timers_.begin()->first;
     auto predicate = [this, until] {
       return shutdown_ || timers_.empty() || timers_.begin()->first != until;
     };
-    if (cv_.wait_until(lk, until, predicate)) {
-      // Timers can be expired only if the CV exits due to a timeout.
+    if (cv_.wait_until(lk, until, std::move(predicate))) {
+      // Timers can be expired only if wait_util() exits due to a timeout.
       continue;
     }
-    // if we get here we know that predicate() is false. Which implies that
-    // timers_ is not empty.
+    // if we get here we know that predicate() is false, which implies that
+    // `timers_` is not empty and the first timer's key is `until`.
     auto p = std::move(timers_.begin()->second);
     timers_.erase(timers_.begin());
     // Relinquish the leader role, release the mutex and then signal a follower.
     has_leader_ = false;
-    is_leader = false;
     lk.unlock();
+    is_leader = false;
     // Elect a new leader (if available) to continue expiring timers.
     cv_follower_.notify_one();
     // This may run user code (in the continuations for the future). That may
-    // do all kinds of things, including calling back on this class to create
+    // do all kinds of things, including calling back into this class to create
     // new timers. We cannot hold the mutex while it is running.
-    p.set_value(std::chrono::system_clock::now());
+    p.set_value(until);
     lk.lock();
   }
   CancelAll(std::move(lk), "TimerQueue shutdown");
@@ -125,11 +115,12 @@ void TimerQueue::CancelAll() {
 }
 
 void TimerQueue::CancelAll(std::unique_lock<std::mutex> lk, char const* msg) {
-  decltype(timers_) timers;
-  std::swap(timers, timers_);
-  lk.unlock();
-  for (auto& timer : timers) {
-    timer.second.set_value(MakeCancelled(msg, GCP_ERROR_INFO()));
+  while (!timers_.empty()) {
+    auto p = std::move(timers_.begin()->second);
+    timers_.erase(timers_.begin());
+    lk.unlock();
+    p.set_value(MakeCancelled(msg, GCP_ERROR_INFO()));
+    lk.lock();
   }
 }
 
--- a/google/cloud/internal/timer_queue.cc
+++ b/google/cloud/internal/timer_queue.cc
@@ -88,10 +88,10 @@ void TimerQueue::Service() {
       return shutdown_ || timers_.empty() || timers_.begin()->first != until;
     };
     if (cv_.wait_until(lk, until, std::move(predicate))) {
-      // Timers can be expired only if wait_util() exits due to a timeout.
+      // Timers can be expired only if wait_util() returns due to a timeout.
       continue;
     }
-    // if we get here we know that predicate() is false, which implies that
+    // If we get here we know that predicate() is false, which implies that
     // `timers_` is not empty and the first timer's key is `until`.
     auto p = std::move(timers_.begin()->second);
     timers_.erase(timers_.begin());
@@ -119,6 +119,7 @@ void TimerQueue::CancelAll(std::unique_lock<std::mutex> lk, char const* msg) {
     auto p = std::move(timers_.begin()->second);
     timers_.erase(timers_.begin());
     lk.unlock();
+    cv_.notify_one();
     p.set_value(MakeCancelled(msg, GCP_ERROR_INFO()));
     lk.lock();
   }
--- a/google/cloud/internal/timer_queue.cc
+++ b/google/cloud/internal/timer_queue.cc
@@ -14,6 +14,7 @@
 
 #include "google/cloud/internal/timer_queue.h"
 #include "google/cloud/future.h"
+#include "google/cloud/internal/make_status.h"
 #include "google/cloud/status_or.h"
 #include "google/cloud/version.h"
 #include <chrono>
@@ -24,6 +25,14 @@ namespace google {
 namespace cloud {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
+namespace {
+
+StatusOr<std::chrono::system_clock::time_point> MakeCancelled(
+    std::string msg, ErrorInfoBuilder b) {
+  return CancelledError(std::move(msg), std::move(b));
+}
+
+}  // namespace
 
 future<StatusOr<std::chrono::system_clock::time_point>> TimerQueue::Schedule(
     std::chrono::system_clock::time_point tp) {
@@ -32,9 +41,10 @@ future<StatusOr<std::chrono::system_clock::time_point>> TimerQueue::Schedule(
   bool should_notify = false;
   {
     std::lock_guard<std::mutex> lk(mu_);
-    if (shutdown_)
-      return make_ready_future(StatusOr<std::chrono::system_clock::time_point>(
-          Status(StatusCode::kCancelled, "TimerQueue already shutdown")));
+    if (shutdown_) {
+      return make_ready_future(
+          MakeCancelled("TimerQueue already shutdown", GCP_ERROR_INFO()));
+    }
     auto iter = timers_.emplace(tp, std::move(p));
     should_notify = iter == timers_.begin();
   }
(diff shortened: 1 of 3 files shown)


google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
-      return shutdown_ || ne < until ||
-             HasExpiredTimer(lk, std::chrono::system_clock::now());
+  // It is easier to understand this if you first ignore shutdown and empty
+  // timer queues.
+  //
+  // The threads play two different roles:
+  // - A single thread at a time plays the leader role.
+  // - All other threads are followers and block on cv_follower_.
+  //
+  // Once a timer expires the leader thread relinquishes its role and wakes up
```
devbww, line 70: s/expires/expires, /

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
-  while (true) {
-    std::unique_lock<std::mutex> lk(mu_);
-    auto const until = NextExpiration(lk);
-    // TODO(#10521): Predicate condition ne < until may need to be reworked.
-    auto predicate = [this, until, &lk] {
-      auto const ne = NextExpiration(lk);
-      return shutdown_ || ne < until ||
-             HasExpiredTimer(lk, std::chrono::system_clock::now());
+  // It is easier to understand this if you first ignore shutdown and empty
+  // timer queues.
```
devbww, line 64: I would suggest dropping this preamble.  We address shutdown below, and never mention empty queues again.

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+  // - A single thread at a time plays the leader role.
+  // - All other threads are followers and block on cv_follower_.
+  //
+  // Once a timer expires the leader thread relinquishes its role and wakes up
+  // one follower thread to become the new leader. Only after doing so it
+  // runs the code to expire the timer.
+  //
+  // This is all complicated by shutdown. Basically all thread needs to wake
+  // up when the TQ is shutdown and one of them will expire all the timers.
+  //
```
devbww, line 76: Delete extra comment line?

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+  // The threads play two different roles:
+  // - A single thread at a time plays the leader role.
+  // - All other threads are followers and block on cv_follower_.
+  //
+  // Once a timer expires the leader thread relinquishes its role and wakes up
+  // one follower thread to become the new leader. Only after doing so it
+  // runs the code to expire the timer.
+  //
+  // This is all complicated by shutdown. Basically all thread needs to wake
+  // up when the TQ is shutdown and one of them will expire all the timers.
```
devbww, line 75: My version used all the threads to cancel the outstanding timers, which I thought was a feature.  You might consider it.

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+  //
+  std::unique_lock<std::mutex> lk(mu_);
+  auto is_leader = false;
```
devbww, line 78: Initialize outside the lock?

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+  while (!shutdown_) {
+    if (!is_leader && has_leader_) {
+      // The current thread becomes a follower until no leader is available.
```
devbww, line 81: s/until no leader is available/while there is a leader/ perhaps?

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+      cv_follower_.wait(lk, [this] { return shutdown_ || !has_leader_; });
+      continue;
+    }
+    // The is_leader flag allows us to restart this loop without worrying about
+    // electing new leaders.
```
devbww, line 86: I would move this comment to the `is_leader` declaration.

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+      return shutdown_ || timers_.empty() || timers_.begin()->first != until;
     };
-    // TODO(#10512): Only wake one thread to service the next timer instead
-    // of all of them.
-    cv_.wait_until(lk, until, predicate);
-    if (shutdown_) return NotifyShutdown(std::move(lk));
-    auto now = std::chrono::system_clock::now();
-    if (HasExpiredTimer(lk, now)) {
-      ExpireTimer(std::move(lk), now);
+    if (cv_.wait_until(lk, until, predicate)) {
```
devbww, line 100: `std::move(predicate)`

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+      // Timers can be expired only if the CV exits due to a timeout.
```
devbww, line 101: s/CV/wait_until()/?

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+    auto p = std::move(timers_.begin()->second);
+    timers_.erase(timers_.begin());
+    // Relinquish the leader role, release the mutex and then signal a follower.
+    has_leader_ = false;
+    is_leader = false;
+    lk.unlock();
+    // Elect a new leader (if available) to continue expiring timers.
+    cv_follower_.notify_one();
+    // This may run user code (in the continuations for the future). That may
+    // do all kinds of things, including calling back on this class to create
```
devbww, line 115: s/on/into/?

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+    // new timers. We cannot hold the mutex while it is running.
+    p.set_value(std::chrono::system_clock::now());
```
devbww, line 117: I still believe we should use `until` to satisfy the future.  It (1) bears some relationship to the expiring timer, rather than being some arbitrary point between scheduling and witnessing, and (2) removes the last call to `now()`.  It would also allow the tests to be more exacting and less flaky.

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
-  // serialize the expiration.  Instead, we pick another thread to continue
-  // expiring timers.
-  if (has_expired_timers) cv_.notify_one();
-  p.set_value(tp);
-}
-
-void TimerQueue::NotifyShutdown(std::unique_lock<std::mutex> lk) {
-  auto timers = std::move(timers_);
+void TimerQueue::CancelAll(std::unique_lock<std::mutex> lk, char const* msg) {
+  decltype(timers_) timers;
```
devbww, line 128: Nit: Declaring this outside the lock was previously a feature, but I guess it doesn't really matter.

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
-    cv_.wait_until(lk, until, predicate);
-    if (shutdown_) return NotifyShutdown(std::move(lk));
-    auto now = std::chrono::system_clock::now();
-    if (HasExpiredTimer(lk, now)) {
-      ExpireTimer(std::move(lk), now);
+    if (cv_.wait_until(lk, until, predicate)) {
+      // Timers can be expired only if the CV exits due to a timeout.
+      continue;
     }
+    // if we get here we know that predicate() is false. Which implies that
```
devbww, line 104: s/. Which/, which/

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+    // timers_ is not empty.
```
devbww, line 105: ... and that the first timer's key is `until`.

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+    auto p = std::move(timers_.begin()->second);
+    timers_.erase(timers_.begin());
+    // Relinquish the leader role, release the mutex and then signal a follower.
+    has_leader_ = false;
+    is_leader = false;
```
devbww, line 110: Move outside the lock.

google/cloud/internal/timer_queue.cc
```diff
@@ -32,9 +41,10 @@ future<StatusOr<std::chrono::system_clock::time_point>> TimerQueue::Schedule(
   bool should_notify = false;
   {
     std::lock_guard<std::mutex> lk(mu_);
-    if (shutdown_)
-      return make_ready_future(StatusOr<std::chrono::system_clock::time_point>(
-          Status(StatusCode::kCancelled, "TimerQueue already shutdown")));
+    if (shutdown_) {
+      return make_ready_future(
+          MakeCancelled("TimerQueue already shutdown", GCP_ERROR_INFO()));
```
devbww, line 46: Q. Is it worth distinguishing (in the status message) between a pre- and post-Shutdown() cancellation?  That is, s/ already// perhaps?  If `Schedule()` calls aren't already synchronized with `Shutdown()`, then it is just a race anyway.

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +56,83 @@ void TimerQueue::Shutdown() {
...
+    // electing new leaders.
+    is_leader = true;
+    has_leader_ = true;
+    if (timers_.empty()) {
+      cv_.wait(lk, [this] { return shutdown_ || !timers_.empty(); });
+      continue;
+    }
+    // Should a new timer appear that changes the "first" timer, we need to
+    // wake up and recompute the sleep time. But note that the leader thread
+    // does not need to relinquish that role to do so.
```
devbww, line 95: s/that/its/ perhaps.

google/cloud/internal/timer_queue_test.cc
```diff
@@ -83,31 +83,41 @@ TEST(TimerQueueTest, ScheduleEarlierTimerSingleRunner) {
...
+  auto insert_thread_id = [&](auto) {
+    std::unique_lock<std::mutex> lk(mu);
+    ids.insert(std::this_thread::get_id());
+    if (ids.size() == kRunners) {
+      lk.unlock();
+      cv.notify_all();
+      return;
+    }
+    // Block after inserting each thread id, to force all threads to
+    // participate.
```
devbww, line 101: Much nicer.  Thanks.

google/cloud/internal/timer_queue.cc
```diff
@@ -14,6 +14,7 @@
 
 #include "google/cloud/internal/timer_queue.h"
 #include "google/cloud/future.h"
+#include "google/cloud/internal/make_status.h"
```
devbww, line 17: The other includes, except the class one, can go away because they are necessary parts of the interface brought in by the class header.

google/cloud/internal/timer_queue.h
```diff
@@ -50,20 +53,15 @@ class TimerQueue {
...
-  void NotifyShutdown(std::unique_lock<std::mutex> lk);
-  // Requires a lock to be held before calling.
-  std::chrono::system_clock::time_point NextExpiration(
-      std::unique_lock<std::mutex> const&);
+  void CancelAll(std::unique_lock<std::mutex> lk, char const* msg);
 
   using PromiseType = promise<StatusOr<std::chrono::system_clock::time_point>>;
   std::mutex mu_;
   std::condition_variable cv_;
+  std::condition_variable cv_follower_;
```
devbww, line 61: Add `<condition_variable>` include, which was missing before too.

coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: Thanks, removed.

google/cloud/internal/timer_queue.cc
coryan: Done

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: N/A

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: Done

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: FWIW, your version was holding the lock through the `while(!timers_.empty())` loop, so effectively only one thread was doing the work.  But I should have noticed the intent, and I like this idea.  Done and added a test.

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.cc
coryan: Done.

google/cloud/internal/timer_queue.h
```diff
@@ -27,6 +27,9 @@ namespace cloud {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
 
+/**
+ * Implement a
```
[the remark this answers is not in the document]
coryan: Fixed.

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +50,80 @@ void TimerQueue::Shutdown() {
...
     };
-    // TODO(#10512): Only wake one thread to service the next timer instead
-    // of all of them.
-    cv_.wait_until(lk, until, predicate);
-    if (shutdown_) return NotifyShutdown(std::move(lk));
-    auto now = std::chrono::system_clock::now();
-    if (HasExpiredTimer(lk, now)) {
-      ExpireTimer(std::move(lk), now);
+    if (cv_.wait_until(lk, until, std::move(predicate))) {
+      // Timers can be expired only if wait_util() exits due to a timeout.
```
devbww, line 91: s/exits/returns/

google/cloud/internal/timer_queue.h
```diff
@@ -27,6 +28,40 @@ namespace cloud {
...
+ * tq.Schedule(now + 200ms).then([](auto f) { std::out << "timer 2\n"; });
+ * tq.Schedule(now + 200ms).then([](auto f) { std::out << "timer 3\n"; });
+ * @code
+ *
+ * To shutdown the timer queue you need to call `Shutdown()`:
+ * @code
+ * tq.Shutdown();
+ * @endcode
+ *
+ * Don't forget to join your thread pool.  Remember that this threads will not
```
devbww, line 59: s/this/these/

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +50,80 @@ void TimerQueue::Shutdown() {
...
+      continue;
     }
+    // if we get here we know that predicate() is false, which implies that
```
devbww, line 94: s/if/If/

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +50,80 @@ void TimerQueue::Shutdown() {
...
-  for (auto& timer : timers) {
-    timer.second.set_value(Status(StatusCode::kCancelled, "Timer cancelled"));
-  }
-}
-
-// Requires a lock to be held before calling.
-bool TimerQueue::HasExpiredTimer(std::unique_lock<std::mutex> const&,
-                                 std::chrono::system_clock::time_point tp) {
-  return !timers_.empty() && timers_.begin()->first < tp;
+  CancelAll(std::unique_lock<std::mutex>(mu_), "TimerQueue cancel all");
```
devbww, line 114: It looks like this guy should then also `cv_.notify_all()` now that `timers_` has changed.

google/cloud/internal/timer_queue.cc
devbww: > FWIW, your version was holding the lock through the while(!timers_.empty()) loop

And FWIW, I'm not sure which version you might have been looking at, but it dropped/reobtained the lock during the `Expire()` helper.

devbww (commented):
An extra notify methinks.

google/cloud/internal/timer_queue.cc
coryan: I think there is at most one thread blocked on `cv_`.  `cv_.notify_one()` seems like enough, but probably won't make a difference.

In any case, I am planning to remove `CancelAll()` in favor of per-timer cancels.

google/cloud/internal/timer_queue.cc
coryan: Fixed

google/cloud/internal/timer_queue.cc
coryan: Fixed

google/cloud/internal/timer_queue.cc
devbww: > I think there is at most one thread blocked on `cv_`. `cv_.notify_one()` seems like enough

I agree ... just the leader.  And therefore, that `Shutdown()` now need only `cv_.notify_one()` too.

> In any case, I am planning to remove `CancelAll()` in favor of per-timer cancels.

+1 ... It seemed difficult to use at best.

google/cloud/internal/timer_queue.cc
```diff
@@ -46,80 +50,81 @@ void TimerQueue::Shutdown() {
   std::unique_lock<std::mutex> const lk(mu_);
   shutdown_ = true;
   cv_.notify_all();
```
devbww, line 52: See the `CancelAll()` comment.

coryan:
PTAL

google/cloud/internal/timer_queue.cc
```diff
@@ -45,81 +49,82 @@ future<StatusOr<std::chrono::system_clock::time_point>> TimerQueue::Schedule(
...
-  lk.unlock();
-  for (auto& timer : timers) {
-    timer.second.set_value(
-        Status(StatusCode::kCancelled, "TimerQueue shutdown"));
+void TimerQueue::CancelAll(std::unique_lock<std::mutex> lk, char const* msg) {
+  while (!timers_.empty()) {
+    auto p = std::move(timers_.begin()->second);
+    timers_.erase(timers_.begin());
+    lk.unlock();
+    cv_.notify_one();
```
devbww, line 122: On second thought, this really only needs to happen once, when `timers_.empty()` ... it's a race whether a newly-runnable leader might expire a timer before we cancel it.

But, given your intention to do away with `CancelAll()`, it matters not.

devbww (approved):
Thanks.

PR #10539, opened 2023-01-10
impl(rest): add RestCompletionQueueImpl

scotthart:
This PR provides an implementation for use with `CompletionQueue` that does not rely on `grpc::CompletionQueue`. This implementation of `CompletionQueue` will be used to make LRO and Async calls for services that use REST transport.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539)

### DIFF
--- a/google/cloud/internal/rest_completion_queue_impl.cc
+++ b/google/cloud/internal/rest_completion_queue_impl.cc
@@ -21,27 +21,18 @@ namespace rest_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 RestCompletionQueueImpl::RestCompletionQueueImpl()
-    : shutdown_guard_(
-          // Capturing `this` here is safe because the lifetime of copies of
-          // this member do not outlive `StartOperation`.
-          std::shared_ptr<void>(reinterpret_cast<void*>(this),
-                                [this](void*) { tq_.Shutdown(); })) {}
+    : tq_(std::make_shared<internal::TimerQueue>()) {}
 
-void RestCompletionQueueImpl::Run() { tq_.Service(); }
+void RestCompletionQueueImpl::Run() { tq_->Service(); }
 
-void RestCompletionQueueImpl::Shutdown() {
-  std::lock_guard<std::mutex> lk(mu_);
-  tq_.Shutdown();
-  shutdown_ = true;
-  shutdown_guard_.reset();
-}
+void RestCompletionQueueImpl::Shutdown() { tq_->Shutdown(); }
 
-void RestCompletionQueueImpl::CancelAll() { tq_.CancelAll(); }
+void RestCompletionQueueImpl::CancelAll() { tq_->CancelAll(); }
 
 future<StatusOr<std::chrono::system_clock::time_point>>
 RestCompletionQueueImpl::MakeDeadlineTimer(
     std::chrono::system_clock::time_point deadline) {
-  return tq_.Schedule(deadline);
+  return tq_->Schedule(deadline);
 }
 
 future<StatusOr<std::chrono::system_clock::time_point>>
@@ -63,6 +54,12 @@ void RestCompletionQueueImpl::RunAsync(
   // TODO(#xxxxx): Refactor this mechanism to a more deterministic solution.
   MakeRelativeTimer(std::chrono::milliseconds(500))
       .then([f = std::move(function)](auto) { f->exec(); });
+}
+
+void RestCompletionQueueImpl::StartOperation(
+    std::shared_ptr<internal::AsyncGrpcOperation>,
+    absl::FunctionRef<void(void*)>) {
+  GCP_LOG(FATAL) << " function not supported.\n";
 }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
--- a/google/cloud/internal/rest_completion_queue_impl.cc
+++ b/google/cloud/internal/rest_completion_queue_impl.cc
@@ -51,7 +51,7 @@ void RestCompletionQueueImpl::RunAsync(
   ++run_async_counter_;
   // Some delay is necessary to reduce the likelihood that the timer expires
   // before .then() is called.
-  // TODO(#xxxxx): Refactor this mechanism to a more deterministic solution.
+  // TODO(#10553): Refactor this mechanism to a more deterministic solution.
   MakeRelativeTimer(std::chrono::milliseconds(500))
       .then([f = std::move(function)](auto) { f->exec(); });
 }
--- a/google/cloud/internal/rest_completion_queue_impl.cc
+++ b/google/cloud/internal/rest_completion_queue_impl.cc
@@ -1,0 +1,71 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/internal/rest_completion_queue_impl.h"
+#include "google/cloud/internal/timer_queue.h"
+
+namespace google {
+namespace cloud {
+namespace rest_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+RestCompletionQueueImpl::RestCompletionQueueImpl()
+    : shutdown_guard_(
+          // Capturing `this` here is safe because the lifetime of copies of
+          // this member do not outlive `StartOperation`.
+          std::shared_ptr<void>(reinterpret_cast<void*>(this),
+                                [this](void*) { tq_.Shutdown(); })) {}
+
+void RestCompletionQueueImpl::Run() { tq_.Service(); }
+
+void RestCompletionQueueImpl::Shutdown() {
+  std::lock_guard<std::mutex> lk(mu_);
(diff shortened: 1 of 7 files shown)


scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/internal/rest_completion_queue_impl.h` line 72 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539#-NLSbbEOsXtyVRlBIUpR-r1-72:-NLWymjV22-vOwQHjjIe:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/64a39f979e67862b257080231c4c80dbb62a6d5c/google/cloud/internal/rest_completion_queue_impl.h#L72)):*

Seems like this should be in the `.cc` file?
</blockquote></details>

Done.
___
*[`google/cloud/internal/rest_completion_queue_impl.h` line 83 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539#-NLSbbEOsXtyVRlBIUpR-r1-83:-NLWyp7AB36sbgygjB8h:b-ezgxjx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/64a39f979e67862b257080231c4c80dbb62a6d5c/google/cloud/internal/rest_completion_queue_impl.h#L83)):*

This is about to become `std::shared_ptr<internal::TimerQueue>`.
</blockquote></details>

Changed to std::shared_ptr
___
*[`google/cloud/internal/rest_completion_queue_impl.h` line 85 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539#-NLSbbEOsXtyVRlBIUpR-r1-85:-NLWys2c18h2sptrrovO:b-64p4cf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/64a39f979e67862b257080231c4c80dbb62a6d5c/google/cloud/internal/rest_completion_queue_impl.h#L85)):*

The `shutdown_guard_` in the gRPC-based completion queues was to avoid returning from `Run()` before all operations had completed.  `internal::TimerQueue` guarantees that all timers are cancelled before `Service()` returns.  I am not sure you need this flags anymore.
</blockquote></details>

Removed the shutdown apparatus.

scotthart:
Also removed tests that were duplicates from completion_queue_test.cc

coryan (changes requested):
I think we can remove the regression test, it does not apply to this implementation.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/internal/rest_completion_queue_impl.h` line 36 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539#-NLSbbEOsXtyVRlBIUpR-r2-36:-NLXaDCQ1vHJjxWL7xFV:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1d81f17b768a28baac9c0516d5ef0a0338eb97f2/google/cloud/internal/rest_completion_queue_impl.h#L36)):*

The format for Doxygen comments is

```
/**
  * brief description of the thing
  *   <<---- this blank line matters, it separates the brief description
  * long description of the thing
  */
```

</blockquote></details>

Done.
___
*[`google/cloud/internal/rest_completion_queue_impl.cc` line 24 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539#-NLSbbEOsXtyVRlBIUpQ-r2-24:-NLXaFU76TkhYl2un_Dg:bb6sh2i) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1d81f17b768a28baac9c0516d5ef0a0338eb97f2/google/cloud/internal/rest_completion_queue_impl.cc#L24)):*

My PR makes the constructor private.  You should merge first and I will fix this stuff.
</blockquote></details>

Ack.
___
*[`google/cloud/internal/rest_completion_queue_impl.cc` line 54 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539#-NLSbbEOsXtyVRlBIUpQ-r2-54:-NLXaGjZDATL5bMKI7t_:bqmjwcz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1d81f17b768a28baac9c0516d5ef0a0338eb97f2/google/cloud/internal/rest_completion_queue_impl.cc#L54)):*

Please add the actual bug number. 
</blockquote></details>

Added.
___
*[`google/cloud/internal/rest_completion_queue_impl_test.cc` line 34 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539#-NLSbbEOsXtyVRlBIUpS-r2-34:-NLXaI3e0xU2rBIRuUy3:bh6st6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1d81f17b768a28baac9c0516d5ef0a0338eb97f2/google/cloud/internal/rest_completion_queue_impl_test.cc#L34)):*

I am not sure why we need a test for a regression of that bug.  As you can see from the comments it was specific to the gRPC implementation:

https://github.com/googleapis/google-cloud-cpp/pull/10539/files#diff-308f858d99fedd31716c9d0a2f362d9c67cd11e693a01e0e5fad56ca309bddc6R34


</blockquote></details>

Overzealous test removed.

google/cloud/internal/rest_completion_queue_impl.cc
```diff
@@ -0,0 +1,68 @@
...
+future<StatusOr<std::chrono::system_clock::time_point>>
+RestCompletionQueueImpl::MakeDeadlineTimer(
+    std::chrono::system_clock::time_point deadline) {
+  return tq_->Schedule(deadline);
+}
+
+future<StatusOr<std::chrono::system_clock::time_point>>
+RestCompletionQueueImpl::MakeRelativeTimer(std::chrono::nanoseconds duration) {
+  using std::chrono::system_clock;
+  auto const d = std::chrono::duration_cast<system_clock::duration>(duration);
```
devbww, line 41: This should probably be `std::chrono::ceil<system_clock::duration>(duration)`, so timers don't fire early, but that is C++17 talk.

In the meantime, something like ...
```
  if (d < duration) d += system_clock::duration{1};
```
should suffice.

google/cloud/internal/rest_completion_queue_impl.cc
```diff
@@ -0,0 +1,68 @@
...
+
+// Use an "immediately" expiring timer in order to get the thread(s) servicing
+// the TimerQueue to execute the function. However, if the timer
+// expires before .then() is invoked, the lambda will be immediately called and
+// the passing of execution to the queue servicing thread will not occur.
+void RestCompletionQueueImpl::RunAsync(
+    std::unique_ptr<internal::RunAsyncBase> function) {
+  ++run_async_counter_;
+  // Some delay is necessary to reduce the likelihood that the timer expires
+  // before .then() is called.
```
devbww, line 53: Aside:  I'm afraid I don't grok that.  It is the timer expiring that invokes the `.then()`, no?

coryan: Consider:

```
promise<int> p;
auto f = p.get_future();
```

In this code the assertion will succeed, because the future was already satisfied when you called `.then()` and therefore the lambda runs immediately upon calling `.then()`:

```
p.set_value(42);
auto g = f.then([](auto) { return std::this_thread::get_id(); });
assert(g.get() == std::this_thread::get_id());
```

In this code the assertion will fail, because the future is not satisfied when `.then()` is called and therefore the lambda runs in the thread that satisfies the future:

```
auto g = f.then([](auto) { return std::this_thread::get_id(); });
auto t = std::thread([](promise<int> p) { p.set_value(42); }, std::move(p); );
assert(g.get() == std::this_thread::get_id());
```

In this code it is unknown if the assertion will succeed or fail because the future may or may not be called by the time `.then()` is called.

```
auto t = std::thread([](promise<int> p) { p.set_value(42); }, std::move(p); );
auto g = f.then([](auto) { return std::this_thread::get_id(); });
assert(g.get() == std::this_thread::get_id());
t.join();
```

We need some changes in TimerQueue (#10553) to make sure `RunAsync()` behaves like the second case (except when timer queues are shutdown, but that is another story).

coryan:
Disable auto merge so @devbww and @scotthart can talk.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539)* status: all files reviewed, 2 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/internal/rest_completion_queue_impl.cc` line 41 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10539#-NLSbbEOsXtyVRlBIUpQ-r4-41:-NLY-gJCD8HcykQlzVeC:bh6xm8m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a6c43334871daa8ed22eb61e146e1798e44515a2/google/cloud/internal/rest_completion_queue_impl.cc#L41)):*

This should probably be `std::chrono::ceil<system_clock::duration>(duration)`, so timers don't fire early, but that is C++17 talk.

In the meantime, something like ...
```
  if (d < duration) d += system_clock::duration{1};
```
should suffice.
</blockquote></details>

I'll fix this in a follow up PR.

google/cloud/internal/rest_completion_queue_impl.cc
devbww: >> Aside: I'm afraid I don't grok that. It is the timer expiring that invokes the .then(), no?
> 
> the future may or may not be [satisfied] by the time .then() is called.

OK.  Thanks.  Indeed, no.

I guess, then, the problem might be more clearly described as being about ensuring `exec()` is called from a service thread.

> We need some changes in TimerQueue (https://github.com/googleapis/google-cloud-cpp/issues/10553)

Ack.  The simplest approach (methinks) is unrelated to timers, so the young class's name has already started to lose some meaning.

PR #10561, opened 2023-01-12
doc: reorder top-level README

coryan:
Also use a `<detail>` to avoid a giant wall of text with all the libraries.

Maybe we should highlight the most commonly used libraries outside the zippy (is that the name?), and just hide the other 90+ libraries.  Thoughts?

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10561)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,78 @@ This repository contains idiomatic C++ client libraries for the following
 > Please check the [CHANGELOG] for important announcements and upcoming
 > changes.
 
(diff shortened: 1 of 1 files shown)


devbww (approved):
> Maybe we should highlight the most commonly used libraries outside the zippy (is that the name?), and just hide the other 90+ libraries. Thoughts?

Yes, maybe.  :-)

PR #10564, opened 2023-01-13
impl(storage): add a function for building XML for MPU

tmatsuo:
A function for building an XML request for `Completing multipart upload` API:

https://cloud.google.com/storage/docs/xml-api/post-object-complete.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10564)


google/cloud/storage/internal/xml_builders.h
```diff
@@ -0,0 +1,69 @@
...
+
+constexpr unsigned int kMaxPartNumber = 10000;
+
+/**
+ * A builder for an XML request for `Completing multipart upload` API
+ * described at
+ * https://cloud.google.com/storage/docs/xml-api/post-object-complete.
+ *
+ * @par Thread-safety
+ * This class is meant to be shared among multiple threads.
```
coryan, line 39: Can you say more?  Why do we need to have multiple threads adding to this?

google/cloud/storage/internal/xml_builders.h
```diff
@@ -0,0 +1,69 @@
...
+ * Each thread can add the part information to the builder.
+ */
+class CompleteMultipartUploadXmlBuilder {
+ public:
+  static std::shared_ptr<CompleteMultipartUploadXmlBuilder> Create() {
+    return std::shared_ptr<CompleteMultipartUploadXmlBuilder>{
+        new CompleteMultipartUploadXmlBuilder()};
```
coryan, line 46: s/Builder())/Builder)/.

google/cloud/storage/internal/xml_builders.h
coryan: In general, the bias in C++ is towards putting the locks outside the class, because often you need to lock more than one thing, for example, I expect you need to track how many parts are completed so you know when to call `Build()`?

google/cloud/storage/internal/xml_builders.h
```diff
@@ -0,0 +1,69 @@
...
+ * described at
+ * https://cloud.google.com/storage/docs/xml-api/post-object-complete.
+ *
+ * @par Thread-safety
+ * This class is meant to be shared among multiple threads.
+ * Each thread can add the part information to the builder.
+ */
+class CompleteMultipartUploadXmlBuilder {
+ public:
+  static std::shared_ptr<CompleteMultipartUploadXmlBuilder> Create() {
```
coryan, line 44: Why does this need to be a `std:shared_ptr<>`, can you add a comment?

google/cloud/storage/internal/xml_builders.h
```diff
@@ -0,0 +1,69 @@
...
+#include <memory>
+#include <mutex>
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+constexpr unsigned int kMaxPartNumber = 10000;
```
coryan, line 31: I would write `auto constexpr kMaxPartNumber = 10000U;` and I would probably put that in the `.cc` file.

google/cloud/storage/internal/xml_builders.h
```diff
@@ -0,0 +1,69 @@
...
+  static std::shared_ptr<CompleteMultipartUploadXmlBuilder> Create() {
+    return std::shared_ptr<CompleteMultipartUploadXmlBuilder>{
+        new CompleteMultipartUploadXmlBuilder()};
+  }
+
+  ~CompleteMultipartUploadXmlBuilder() = default;
+
+  /// Add a part (pair of int and string).
+  Status AddPart(unsigned int part_number, std::string etag);
+  /// Build an Xml tree and return it.
```
coryan, line 53: nit: blank line

google/cloud/storage/internal/xml_node.h
```diff
@@ -25,6 +25,10 @@ namespace cloud {
 namespace storage_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+std::string EscapeXmlTag(std::string const& val);
```
coryan, line 28: These probably belong in a different header, maybe `xml_escape.h`?

google/cloud/storage/internal/xml_builders.cc
```diff
@@ -0,0 +1,56 @@
...
+    return internal::InvalidArgumentError(
+        absl::StrCat("part_number can not be more than ", kMaxPartNumber));
+  }
+  std::lock_guard<std::mutex> lk{part_map_mu_};
+  part_map_[part_number] = std::move(etag);
+  return Status();
+}
+
+std::shared_ptr<XmlNode const> CompleteMultipartUploadXmlBuilder::Build() {
+  std::lock_guard<std::mutex> lk{part_map_mu_};
```
coryan, line 39: Hmm... Can I call `Build()` more than once?  If not, why does it need a lock?  If so, what are the semantics?

I suspect `Build()` can be called only once, in which case, the member function should be a `&&` overload (or should be a friend consuming rvalue references.

coryan (changes requested):

google/cloud/storage/internal/xml_builders.h
tmatsuo: I agree. I'll change it accordingly.

google/cloud/storage/internal/xml_builders.h
tmatsuo: ack

google/cloud/storage/internal/xml_builders.h
tmatsuo: I made it a standalone function

google/cloud/storage/internal/xml_builders.h
tmatsuo: Done

google/cloud/storage/internal/xml_builders.h
tmatsuo: Ack

google/cloud/storage/internal/xml_node.h
tmatsuo: Done

google/cloud/storage/internal/xml_builders.cc
tmatsuo: I made it a standalone function. WDYT?

google/cloud/storage/internal/xml_builders.cc
```diff
@@ -0,0 +1,57 @@
...
+  auto target_node = root->AppendTagNode("CompleteMultipartUpload");
+
+  for (auto const& p : parts) {
+    if (p.first == 0) {
+      return internal::InvalidArgumentError("part number cannot be zero",
+                                            GCP_ERROR_INFO());
+    }
+    if (p.first > kMaxPartNumber) {
+      return internal::InvalidArgumentError(
+          absl::StrCat("part number cannot be more than ", kMaxPartNumber),
```
devbww, line 42: It is probably a good idea to include the offending value, `p.first`, in the error message.

google/cloud/storage/internal/xml_builders.cc
```diff
@@ -0,0 +1,57 @@
...
+          GCP_ERROR_INFO());
+    }
+    auto part_tag = target_node->AppendTagNode("Part");
+    part_tag->AppendTagNode("PartNumber")
+        ->AppendTextNode(std::to_string(p.first));
+    part_tag->AppendTagNode("ETag")->AppendTextNode(
+        EscapeXmlContent(std::move(p.second)));
```
devbww, line 49: `p` is const, so the move will have no effect.  Not that it matters for `EscapeXmlContent()`, whose parameter is a `const&`.  So, just remove the `std::move`, I guess.

google/cloud/storage/internal/xml_builders.cc
```diff
@@ -0,0 +1,57 @@
...
+#include "google/cloud/internal/make_status.h"
+
+namespace google {
+namespace cloud {
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+namespace {
+
+auto constexpr kMaxPartNumber = 10000U;
```
devbww, line 26: `constexpr` globals already have internal linkage, so the anonymous namespace is redundant in practice.  That said, perhaps this value should somehow be part of the interface.  It sounds like something the caller would want to know.

That said, this doesn't look like something that should be enforced on the client side.

google/cloud/storage/internal/xml_builders.h
```diff
@@ -0,0 +1,45 @@
...
+#include <mutex>
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * A function for creating an XML request for `Completing multipart upload` API
```
devbww, line 33: s/A function for creating/Creates/

devbww, line 33: We usually limit `backticks` to code text.

google/cloud/storage/internal/xml_builders.h
```diff
@@ -0,0 +1,45 @@
...
+ * described at
+ * https://cloud.google.com/storage/docs/xml-api/post-object-complete.
+ */
+StatusOr<std::shared_ptr<XmlNode>> BuildCompleteMultipartUploadXml(
```
devbww, line 37: The "Build" verb makes it looks like "Complete" is an adjective describing the XML, whereas the documentation uses it as a verb ... _"Completes a multipart upload by concatenating the parts into a single object."_  So, perhaps this is just `CompleteMultipartUpload()`?  (Is the "Xml" understood?)

That said, "completes" seems to imply there should be some input to complete.

google/cloud/storage/internal/xml_escape.h
```diff
@@ -0,0 +1,34 @@
...
+#include "google/cloud/storage/version.h"
+
+namespace google {
+namespace cloud {
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+std::string EscapeXmlTag(std::string const& val);
+
+std::string EscapeXmlContent(std::string const& val);
```
devbww, line 27: Now that they have a header, do they deserve any documentation?

devbww, line 27: Consider using `absl::string_view` for the parameters.

google/cloud/storage/internal/xml_builders.cc
coryan: Funnily, I asked to move this to the `.cc` file.  I agree this is not something the library should enforce:

https://github.com/googleapis/google-cloud-cpp/blob/main/doc/adr/2018-06-19-do-not-duplicate-server-side-validation.md

Unless you need this for the library to work correct, just remove it.

google/cloud/storage/internal/xml_builders.cc
coryan: *if* we we were going to enforce the value, then we should print it, but I think it is better to not enforce it (see the ADR quoted above).

coryan (approved):
I think I have caused enough damage with my review.

google/cloud/storage/internal/xml_builders.cc
tmatsuo: Removed

google/cloud/storage/internal/xml_escape.h
tmatsuo: Done

tmatsuo (commented):
@coryan @devbww Thanks, PTAL

google/cloud/storage/internal/xml_builders_test.cc
```diff
@@ -0,0 +1,59 @@
...
+  </Part>
+</CompleteMultipartUpload>
+)xml";
+
+TEST(BuildCompleteMultipartUploadXmlTest, Build) {
+  std::map<unsigned int, std::string> parts{
+      {5U, "\"aaaa18db4cc2f85cedef654fccc4a4x8\""},
+      {2U, "\"7778aef83f66abc1fa1e8477f296d394\""}};
+  auto xml = CompleteMultipartUpload(parts);
+  ASSERT_TRUE(xml);
```
devbww, line 51: I think we prefer `ASSERT_NE(xml, nullptr)` despite the existence of `std::shared_ptr<T>::operator bool`.

google/cloud/storage/internal/xml_escape.h
```diff
@@ -0,0 +1,37 @@
...
+namespace google {
+namespace cloud {
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/// Escapes a string for using it as a tag name in XML.
+std::string EscapeXmlTag(absl::string_view const& val);
+
+/// Escapes a string for using it as a text in XML.
+std::string EscapeXmlContent(absl::string_view const& val);
```
devbww, line 30: `absl::string_view`s should be passed by value.

I also might call the parameters `tag` and `content` respectively.

google/cloud/storage/internal/xml_builders_test.cc
tmatsuo: I agree, done!

google/cloud/storage/internal/xml_escape.h
tmatsuo: Done

tmatsuo (commented):
@devbww Thanks! PTAL

devbww (approved):

Issue #10489, opened 2023-01-04
Synchronous OpenTelemetry tracing in grpc_utils


dbolduc:
Some tasks from [Milestone 1](https://docs.google.com/document/d/1JTOWvZD59AGxojTV93tKZJ9b_q2w8YUMMyuxCu5_ZC4/edit?resourcekey=0-0XjX6fh2xNKFZN20cu2otg#bookmark=id.i1nfe2c9gho2) of an internal design doc. With these tasks completed, we can support tracing a synchronous unary client call, including its subcomponents (network call, backoffs).

- [x] Add `MakeSpanGrpc()`
- [x] `EndSpan` from `grpc::ClientContext`
- [x] Add sync traced sleeper; Update `RetryLoop` to use it
- [x] Add `InjectTraceContext()`

coryan:
dbolduc is still working on this.

dbolduc:
The piece that is missing is parsing the gRPC peer string into attributes that follow OpenTelemetry's semantic conventions:

https://github.com/googleapis/google-cloud-cpp/blob/1145a8ccff7f8ce7fcebc0256efe6b3093e8e21a/google/cloud/internal/grpc_opentelemetry.cc#L95-L100

dbolduc:
Closing this because...
- Parsing the peer URI is really annoying
- The information is all there. It is just that we are not perfectly following OpenTelemetry semantic conventions.

PR #10567, opened 2023-01-13
impl(otel): make span for gRPC

dbolduc:
Part of the work for #10489 

Introduce a helper function to make spans following the [gRPC semantic conventions](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/rpc/). `MakeSpanGrpc()` doesn't use any gRPC symbols, but logically it belongs in `grpc_utils`.

I decided to preemptively include finer-grained `#ifdef` guards. (`opentelemetry.{h,cc}` will soon look like this too). My convention is to put the guards around the opentelemetry includes only, even if we don't use the other files.

Also I am anticipating that it will be more convenient to use `IsSupersetOf` instead of `UnorderedElementsAre` for span attributes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10567)


google/cloud/internal/grpc_opentelemetry.cc
```diff
@@ -0,0 +1,52 @@
...
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+#ifdef GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
+
+opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> MakeSpanGrpc(
+    opentelemetry::nostd::string_view service,
+    opentelemetry::nostd::string_view method) {
+  namespace SC = opentelemetry::trace::SemanticConventions;
```
coryan, line 35: I do not love the all caps namespace.  Matter of time before we get a `SC` macro over some weird `#include`.

google/cloud/internal/grpc_opentelemetry.h
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Make a span, setting attributes related to gRPC.
+ *
+ * @see
+ * https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/rpc/
+ * for the semantic conventions used for span names and attributes.
+ */
+opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> MakeSpanGrpc(
+    opentelemetry::nostd::string_view service,
+    opentelemetry::nostd::string_view method);
```
coryan, line 40: Here is a scary though: should we capture the target resources: all that stuff the metadata decorator puts in headers.

google/cloud/internal/grpc_opentelemetry.cc
```diff
@@ -0,0 +1,52 @@
...
+  opentelemetry::trace::StartSpanOptions options;
+  options.kind = opentelemetry::trace::SpanKind::kClient;
+  return GetTracer(internal::CurrentOptions())
+      ->StartSpan(absl::StrCat(service.data(), "/", method.data()),
+                  {{SC::kRpcSystem, SC::RpcSystemValues::kGrpc},
+                   {SC::kRpcService, service},
+                   {SC::kRpcMethod, method},
+                   {SC::kNetTransport, SC::NetTransportValues::kIpTcp}},
+                  options);
```
coryan, line 44: Is there a place to capture the `google-cloud-cpp` version and the `grpc` library versions?

coryan: Even if there is no standard place: should we capture them to help our troubleshooting efforts?

coryan (approved):

google/cloud/internal/grpc_opentelemetry.cc
dbolduc: > capture the google-cloud-cpp version?

The version of our library is captured in the instrumentation scope: https://github.com/googleapis/google-cloud-cpp/blob/70543f8de9edaea21e98370527f280f2ac8ee95f/google/cloud/internal/opentelemetry.cc#L34

The awkward thing is that the Cloud Trace `Span` message [does not have a dedicated field for this info](https://github.com/googleapis/googleapis/blob/4eb9a0e60227e1bbe8b80eb9553f463e43ba1817/google/devtools/cloudtrace/v2/trace.proto#L219-L277). But that is a problem for the exporter, which will likely set the instrumentation scope as attributes.

> capture the grpc library version?

I will address this in another commit...

google/cloud/internal/grpc_opentelemetry.h
dbolduc: Probably.

It is scarier than I originally thought. My vague plan was to capture this information outside of the tracing decorators. (the `grpc::ClientContext` does not give us access to the metadata we have set, so we could not do it at the end of the call). I would use something like: `AddSpanAttribute(...)` in the MD decorator: https://github.com/dbolduc/google-cloud-cpp/blob/9e9691476a167626486dba50efab0a7eafef2334/google/cloud/internal/opentelemetry.cc#L98-L106

But it is much better if the resource is known **when we create the span**. That way, the customers can set up a [sampler](https://opentelemetry.io/docs/reference/specification/trace/sdk/#sampler) to filter on a resource name.  (sampling happens when spans are created).

<hr />

Barely-related Aside: The [spec](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/rpc/#grpc-request-and-response-metadata) says: "Instrumentations SHOULD require an explicit configuration of which metadata values are to be captured. Including all request/response metadata values can be a security risk - explicit configuration helps avoid leaking sensitive information".

google/cloud/internal/grpc_opentelemetry.cc
coryan: SG

google/cloud/internal/grpc_opentelemetry.h
coryan: Hmm... Ack.  I think those should be used by the gRPC team.  Maybe we should have one in a different namespace, maybe: `rpc.gcloud-cpp.resource`

In any case, something to add later. Feel free to commit as-is.

google/cloud/internal/grpc_opentelemetry.cc
dbolduc: Fixed.

google/cloud/internal/grpc_opentelemetry.cc
dbolduc: Added `"grpc.version"` attribute.

google/cloud/internal/grpc_opentelemetry.h
devbww: > `grpc::ClientContext` does not give us access to the metadata we have set

```
namespace grpc {
namespace testing {

class ClientContextTestPeer {
 public:
  explicit ClientContextTestPeer(ClientContext* const ctx) : ctx_(ctx) {}

  std::multimap<std::string, std::string> GetSendInitialMetadata() const {
    return ctx_->send_initial_metadata_;
  } 
  
 private:
  ClientContext* const ctx_;  
};

}  // namespace testing
}  // namespace grpc
```
:-)

PR #10568, opened 2023-01-13
impl(storage): add the skeleton of the XML parse function

tmatsuo:
This function will be used for GCS MPU:
https://cloud.google.com/storage/docs/multipart-uploads

The actual parsing part is unimplemented, but it's too big for a single PR, so I chose to split the PRs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10568)


google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Note: This is not a general purpose XML parser. It is only intended to
+ * parse XML responses from the [GCS MPU]. It does not support many XML
+ * features.
+ */
+class XmlParser {
+ public:
+  /// Factory method
+  static std::shared_ptr<XmlParser> Create() {
+    return std::shared_ptr<XmlParser>{new XmlParser()};
```
coryan, line 46: s/new XmlParser()/new XmlParser/

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,76 @@
...
+      size_t max_xml_byte_size = kDefaultMaxXmlByteSize,
+      size_t max_xml_node_num = kDefaultMaxXmlNodeNum,
+      size_t max_xml_node_depth = kDefaultMaxXmlNodeDepth);
+
+ private:
+  XmlParser()
+      : xml_decl_re_{R"(^<\?xml[^>]*\?>)"},
+        xml_doctype_re_{R"(<!DOCTYPE[^>[]*(\[[^\]]*])?>)"},
+        xml_cdata_re_{R"(<!\[CDATA\[[^>]*]]>)"},
+        xml_comment_re_{R"(<!--[^>]*-->)"} {};
```
coryan, line 63: s/{};/{}/

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,76 @@
...
+#include "google/cloud/internal/make_status.h"
+#include "google/cloud/status_or.h"
+#include <regex>
+
+namespace google {
+namespace cloud {
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+constexpr size_t kDefaultMaxXmlByteSize = 1024 * 1024 * 1024;  // 1GiB
```
coryan, line 29: I think these should go to the `.cc` file.  If you need them for default values, consider using `google::cloud::Options`.

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,76 @@
...
+  /// Factory method
+  static std::shared_ptr<XmlParser> Create() {
+    return std::shared_ptr<XmlParser>{new XmlParser()};
+  }
+
+  /// Clean up the given XML string.
+  std::string CleanUpXml(std::string const& content);
+  /// Parse the given string and return an XML tree.
+  StatusOr<std::shared_ptr<XmlNode>> ParseXml(
+      std::string const& content,
```
coryan, line 53: Consider instead `StatusOr<std::shared_ptr<XmlNode>> ParseXml(std::string const& content, Options = {});`

You can create new options to pass the different configuration parameters.

google/cloud/storage/internal/xml_parser.h
coryan: Also, we spell that as `std::size_t`.

google/cloud/storage/internal/xml_parser.cc
```diff
@@ -0,0 +1,47 @@
...
+  ret = std::regex_replace(ret, xml_comment_re_, "");
+  return ret;
+}
+
+StatusOr<std::shared_ptr<XmlNode>> XmlParser::ParseXml(
+    std::string const& content, size_t max_xml_byte_size, size_t, size_t) {
+  // Check size first
+  if (content.length() > max_xml_byte_size) {
+    return internal::InvalidArgumentError(
+        absl::StrCat("XML exceeds the max byte size of ", max_xml_byte_size));
```
coryan, line 36: Please add the `GCP_ERROR_INFO()` call, as in: `return internal::InvalidArgumentError(..., GCP_ERROR_INFO())`.

It is relatively new, so you will not find a lot of usage. It annotates the error with line numbers and function information.

google/cloud/storage/internal/xml_parser.cc
```diff
@@ -0,0 +1,47 @@
...
+  ret = std::regex_replace(ret, xml_doctype_re_, "");
+  ret = std::regex_replace(ret, xml_cdata_re_, "");
+  ret = std::regex_replace(ret, xml_comment_re_, "");
+  return ret;
+}
+
+StatusOr<std::shared_ptr<XmlNode>> XmlParser::ParseXml(
+    std::string const& content, size_t max_xml_byte_size, size_t, size_t) {
+  // Check size first
+  if (content.length() > max_xml_byte_size) {
```
coryan, line 34: I am not sure this is a limit you need to enforce? Or rather, it is too late at this point, you already used all the memory.

coryan (commented):
Drive by

google/cloud/storage/internal/xml_parser.h
tmatsuo: Moved to `google::cloud::options`

google/cloud/storage/internal/xml_parser.h
tmatsuo: Done

google/cloud/storage/internal/xml_parser.h
tmatsuo: Done

google/cloud/storage/internal/xml_parser.h
tmatsuo: Done

tmatsuo (commented):
@coryan Thanks, PTAL

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,70 @@
...
+  /// Parse the given string and return an XML tree.
+  StatusOr<std::shared_ptr<XmlNode>> ParseXml(std::string const& content,
+                                              Options = {});
+
+ private:
+  XmlParser()
+      : xml_decl_re_{R"(^<\?xml[^>]*\?>)"},
+        xml_doctype_re_{R"(<!DOCTYPE[^>[]*(\[[^\]]*\])?>)"},
+        xml_cdata_re_{R"(<!\[CDATA\[[^>]*\]\]>)"},
+        xml_comment_re_{R"(<!--[^>]*-->)"} {}
```
devbww, line 57: Consider moving this to the cc file.  Or, use member initialization (and default the ctor).

google/cloud/storage/internal/xml_parser.cc
```diff
@@ -0,0 +1,52 @@
...
+
+  // Check size first
+  if (content.length() > options.get<XmlParserSourceMaxBytes>()) {
+    return internal::InvalidArgumentError(
+        absl::StrCat("XML exceeds the max byte size of ",
+                     options.get<XmlParserSourceMaxBytes>()),
+        GCP_ERROR_INFO());
+  }
+  // Remove unnecessary bits
+  auto trimmed = CleanUpXml(content);
```
devbww, line 44: Can't we just skip over them during parsing rather than having to edit them out up front?

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,70 @@
...
+ public:
+  /// Factory method
+  static std::shared_ptr<XmlParser> Create() {
+    return std::shared_ptr<XmlParser>{new XmlParser};
+  }
+
+  /// Clean up the given XML string.
+  std::string CleanUpXml(std::string const& content);
+  /// Parse the given string and return an XML tree.
+  StatusOr<std::shared_ptr<XmlNode>> ParseXml(std::string const& content,
```
devbww, line 49: Perhaps `XmlParser::ParseXml()` should just be `XmlParser::Parse()`?  Or, perhaps `Parse()` is a factory function in `XmlNode`?

google/cloud/storage/internal/xml_parser_options.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace google {
+namespace cloud {
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * Use with `google::cloud::Options` to configure the XML parser.
+ *
+ * The maximum size of the source string in bytes.
```
devbww, line 28: Do we need to advertise the defaults?

google/cloud/storage/internal/xml_parser_options.h
```diff
@@ -0,0 +1,63 @@
...
+ */
+struct XmlParserSourceMaxBytes {
```
devbww, line 30: "Bytes" probably isn't worth introducing.  How about something like `XmlParserMaxSourceSize`?

google/cloud/storage/internal/xml_parser_options.h
```diff
@@ -0,0 +1,63 @@
...
+  using Type = std::size_t;
+};
+
+/**
+ * Use with `google::cloud::Options` to configure the XML parser.
+ *
+ * The maximum number of total Nodes in the XML tree.
+ */
+struct XmlParserMaxNodeNum {
```
devbww, line 39: s/MaxNodeNum/MaxNodeCount/?

google/cloud/storage/internal/xml_parser.cc
tmatsuo: Technically possible, but removing makes the parsing part much simpler.

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,70 @@
...
+ */
+class XmlParser {
+ public:
+  /// Factory method
+  static std::shared_ptr<XmlParser> Create() {
+    return std::shared_ptr<XmlParser>{new XmlParser};
+  }
+
+  /// Clean up the given XML string.
+  std::string CleanUpXml(std::string const& content);
```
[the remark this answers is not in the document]
tmatsuo: Made it private

google/cloud/storage/internal/xml_parser.h
tmatsuo: Changed it to `XmlParser::Parse()`

google/cloud/storage/internal/xml_parser.h
tmatsuo: Changed to the default ctor.

google/cloud/storage/internal/xml_parser_options.h
tmatsuo: I added to the parser documentation. WDYT?

google/cloud/storage/internal/xml_parser_options.h
tmatsuo: Done

google/cloud/storage/internal/xml_parser_options.h
```diff
@@ -0,0 +1,63 @@
...
+ * The maximum size of the source string in bytes.
+ */
+struct XmlParserSourceMaxBytes {
+  using Type = std::size_t;
+};
+
+/**
+ * Use with `google::cloud::Options` to configure the XML parser.
+ *
+ * The maximum number of total Nodes in the XML tree.
```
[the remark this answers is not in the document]
tmatsuo: Done

google/cloud/storage/internal/xml_parser_options.h
tmatsuo: Done

google/cloud/storage/internal/xml_parser_options.h
```diff
@@ -0,0 +1,63 @@
...
+ */
+struct XmlParserMaxNodeNum {
+  using Type = std::size_t;
+};
+
+/**
+ * Use with `google::cloud::Options` to configure the XML parser.
+ *
+ * The maximum depth of the node in the XML tree.
```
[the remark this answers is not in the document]
tmatsuo: Done

tmatsuo (commented):
@devbww Thanks, PTAL

google/cloud/storage/internal/xml_parser.cc
```diff
@@ -0,0 +1,54 @@
...
+
+StatusOr<std::shared_ptr<XmlNode>> XmlParser::Parse(std::string const& content,
+                                                    Options options) {
+  internal::CheckExpectedOptions<XmlParserOptionsList>(options, __func__);
+  options = XmlParserDefaultOptions(std::move(options));
+
+  // Check size first
+  if (content.length() > options.get<XmlParserMaxSourceSize>()) {
+    return internal::InvalidArgumentError(
+        absl::StrCat("XML exceeds the max byte size of ",
```
devbww, line 33: I'd suggest including `content.length()` in the error message.

I think we also prefer `std::string::size()` over `length()`.

google/cloud/storage/internal/xml_parser.cc
devbww: It is difficult to say without seeing the implementation of the parser.  At the moment all I can say is that copying the string four times sounds like a significant pessimization.  Might we at least create a single, disjunction RE that can do all four removals in one copy?

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,73 @@
...
+ * [GCS MPU]: https://cloud.google.com/storage/docs/multipart-uploads
+ *
+ * Note: This is not a general purpose XML parser. It is only intended to
+ * parse XML responses from the [GCS MPU]. It does not support many XML
+ * features.
+ *
+ * As a defence to DOS type attacks, the parser has several limits.The following
+ * default values for the limits are large enough for parse XML responses from
+ * the API:
+ *   XmlParserMaxSourceSize>: 1GiB
```
devbww, line 41: s/>//

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,73 @@
...
+
+  /// Parse the given string and return an XML tree.
+  StatusOr<std::shared_ptr<XmlNode>> Parse(std::string const& content,
+                                           Options = {});
+
+ private:
+  XmlParser() = default;
+
+  /// Clean up the given XML string.
+  std::string CleanUpXml(std::string const& content);
```
devbww, line 60: "CleanUp" could probably be much more descriptive.  Perhaps something like `RemoveUnneededElements()`?

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /// Parse the given string and return an XML tree.
+  StatusOr<std::shared_ptr<XmlNode>> Parse(std::string const& content,
+                                           Options = {});
+
+ private:
+  XmlParser() = default;
+
+  /// Clean up the given XML string.
```
devbww, line 59: This doesn't add much, particularly for a private function.  I'd suggest removing it.

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,73 @@
...
+ */
+class XmlParser {
+ public:
+  /// Factory method
+  static std::shared_ptr<XmlParser> Create() {
+    return std::shared_ptr<XmlParser>{new XmlParser};
+  }
+
+  /// Parse the given string and return an XML tree.
+  StatusOr<std::shared_ptr<XmlNode>> Parse(std::string const& content,
```
devbww, line 53: I'd suggest preferring `absl::string_view` for all `string const&` parameters.

google/cloud/storage/internal/xml_parser_options.h
devbww: I think the parser documentation should say which options it uses, but the default values for those options should be given here.

google/cloud/storage/internal/xml_parser.cc
devbww, line 33: s/ byte//

google/cloud/storage/internal/xml_parser_options.h
devbww, line 28: s/ in bytes//

google/cloud/storage/internal/xml_parser.h
tmatsuo: Unfortunately in this case, `absl::string_view` doesn't work well with `std::regex_replace`, for now I'm keeping `std::string const&`.

google/cloud/storage/internal/xml_parser.h
tmatsuo: Now the function is removed

google/cloud/storage/internal/xml_parser.h
tmatsuo: Since I combined the regex, it's removed now

google/cloud/storage/internal/xml_parser_options.h
tmatsuo: Done

tmatsuo: Done

tmatsuo (commented):
@devbww Thanks as always! PTAL

google/cloud/storage/internal/xml_parser.h
devbww: Yes, the APIs of lower layers can be an issue when using `absl::string_view` as a parameter.

In this case, though, I had checked for that, and it seems all would be well if you used the iterator-based-input overloads for `std::regex_replace()`.  Did I miss something?

google/cloud/storage/internal/xml_parser.h
```diff
@@ -0,0 +1,78 @@
...
+  std::regex unneeded_re_{
+      absl::StrCat("(",
+                   R"(^<\?xml[^>]*\?>)",  // XML declaration
+                   "|",
+                   R"(<!DOCTYPE[^>[]*(\[[^\]]*\])?>)",  // DTD(DOCTYPE)
+                   "|",
+                   R"(<!\[CDATA\[[^>]*\]\]>)",  // CDATA
+                   "|",
+                   R"(<!--[^>]*-->)",  // XML comments
+                   ")")};
```
devbww, line 70: I don't know how many `XmlParser`s you expect to create, but it seems a shame to compile this RE multiple times.  I would suggest the "function static" method Darren referred to for only doing this once.

devbww, line 70: Once this is removed, the class doesn't have any state, and so should be converted to a simple function.  (It didn't have any **mutable** state even before, so the same observation really applied then too.)

And once `Parse()` is just a function, I'll re-suggest that it seems to make most sense as a factory in `XmlNode`.

google/cloud/storage/internal/xml_parser_options.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace google {
+namespace cloud {
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * Use with `google::cloud::Options` to configure the XML parser.
+ *
+ * The maximum size of the source string, the default value is 1GiB.
```
devbww, line 28: `s/, the/. The/` here are in the other options.

google/cloud/storage/internal/xml_parser.cc
devbww: It looks like it wouldn't be too complicated to skip during parsing, but OK.

google/cloud/storage/internal/xml_parser.h
tmatsuo: I think I overlooked the overload, will try

google/cloud/storage/internal/xml_parser.h
tmatsuo: I'll change this to a static factory of `XmlNode`.

google/cloud/storage/internal/xml_parser_options.h
tmatsuo: Done

google/cloud/storage/internal/xml_parser.h
tmatsuo: Done!

google/cloud/storage/internal/xml_parser.h
tmatsuo: Yes, it worked with one of the overloads

tmatsuo (commented):
@devbww Thanks!
I changed it to a factory method of XmlNode. PTAL

google/cloud/storage/internal/xml_node.cc
```diff
@@ -14,14 +14,50 @@
...
+                   R"(<!DOCTYPE[^>[]*(\[[^\]]*\])?>)",  // DTD(DOCTYPE)
+                   "|",
+                   R"(<!\[CDATA\[[^>]*\]\]>)",  // CDATA
+                   "|",
+                   R"(<!--[^>]*-->)",  // XML comments
+                   ")")};
+
+  std::string trimmed;
+  // Remove unnecessary bits
+  std::regex_replace(std::back_inserter(trimmed), content.begin(),
```
devbww, line 55: `#include <iterator>`

google/cloud/storage/internal/xml_node.cc
```diff
@@ -14,14 +14,50 @@
...
+
+  static auto* unnecessary_re = new std::regex{
+      absl::StrCat("(",
+                   R"(^<\?xml[^>]*\?>)",  // XML declaration
+                   "|",
+                   R"(<!DOCTYPE[^>[]*(\[[^\]]*\])?>)",  // DTD(DOCTYPE)
+                   "|",
+                   R"(<!\[CDATA\[[^>]*\]\]>)",  // CDATA
+                   "|",
+                   R"(<!--[^>]*-->)",  // XML comments
```
devbww, line 50: I'm not sure this is correct, or the other `*` quantifications.  For example, it would fail to match `<!-- 2 > 1 -->`.

In general, rather than using expressions like `<[^>]*>`, you'll find it easier (and more obviously correct) to use non-greedy matching, like `<.*?>`.

If the element can include newlines, you'll need to use an idiom like `[\s\S]` instead of `.`.

So, a comment would be `<!--[\s\S]*?-->`.

devbww (commented):
PS: You'll probably want to do a once-over on the PR description prior to submission, updating the things that have changed.

google/cloud/storage/internal/xml_node.cc
```diff
@@ -14,14 +14,50 @@
...
+                     " exceeds the max size of ",
+                     options.get<XmlParserMaxSourceSize>()),
+        GCP_ERROR_INFO());
+  }
+
+  static auto* unnecessary_re = new std::regex{
+      absl::StrCat("(",
+                   R"(^<\?xml[^>]*\?>)",  // XML declaration
+                   "|",
+                   R"(<!DOCTYPE[^>[]*(\[[^\]]*\])?>)",  // DTD(DOCTYPE)
```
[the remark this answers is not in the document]
tmatsuo: Good question. It turned out they are case insensitive. But the xml declaration is not.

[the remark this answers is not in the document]
tmatsuo: I'll skip XML declaration in the parser implementation and keep other regex with `icase` option.

google/cloud/storage/internal/xml_node.cc
```diff
@@ -14,14 +14,50 @@
...
+  // Check size first
+  if (content.size() > options.get<XmlParserMaxSourceSize>()) {
+    return internal::InvalidArgumentError(
+        absl::StrCat("The source size ", content.size(),
+                     " exceeds the max size of ",
+                     options.get<XmlParserMaxSourceSize>()),
+        GCP_ERROR_INFO());
+  }
+
+  static auto* unnecessary_re = new std::regex{
```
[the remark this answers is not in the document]
tmatsuo: Done

google/cloud/storage/internal/xml_node.cc
```diff
@@ -14,14 +14,50 @@
...
+      absl::StrCat("(",
```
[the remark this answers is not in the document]
tmatsuo: Done

google/cloud/storage/internal/xml_node.cc
tmatsuo: Done for CDATA and comment (I think the regex for DTD is ok).

google/cloud/storage/internal/xml_node.cc
tmatsuo: Done

tmatsuo (commented):
@devbww Thanks!

I decided to remove the xml declaration from the regex and added `icase` option, and updated the PR description.

PTAL

devbww (approved):
Thanks.

PR #10570, opened 2023-01-13
impl(rest): add AsyncRestRetryLoop

scotthart:
Adds REST version of existing AsyncRetryLoop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10570)


coryan (approved):
This is longer than I remembered.  I think you should commit this as-is, and file a bug to refactor the code (don't assign that to yourself, I may do it if time allows).

PR #10571, opened 2023-01-13
fix(common): use atomic scheduling in RestCompletionQueue

### LINKED ISSUE #10553: Refactor RestCompletionQueueImpl::RunAsync make it deterministic
Currently, RunAsync relies on an "immediately" expiring timer to cause the `.then()` to be executed on a thread servicing the TimerQueue. However, if the timer expires before .then is called the execution does not switch as intended. The current implementation adds a delay to the timer to help ensure this happens, but overhauling this mechanism to be something deterministic would be preferable.

coryan:
Fixes #10553

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10571)

### DIFF
--- a/google/cloud/internal/rest_completion_queue_impl.cc
+++ b/google/cloud/internal/rest_completion_queue_impl.cc
@@ -50,11 +50,8 @@ RestCompletionQueueImpl::MakeRelativeTimer(std::chrono::nanoseconds duration) {
 void RestCompletionQueueImpl::RunAsync(
     std::unique_ptr<internal::RunAsyncBase> function) {
   ++run_async_counter_;
-  // Some delay is necessary to reduce the likelihood that the timer expires
-  // before .then() is called.
-  // TODO(#10553): Refactor this mechanism to a more deterministic solution.
-  MakeRelativeTimer(std::chrono::milliseconds(500))
-      .then([f = std::move(function)](auto) { f->exec(); });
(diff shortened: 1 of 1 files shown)


scotthart (approved):

google/cloud/internal/rest_completion_queue_impl.cc
```diff
@@ -50,11 +50,8 @@ RestCompletionQueueImpl::MakeRelativeTimer(std::chrono::nanoseconds duration) {
 void RestCompletionQueueImpl::RunAsync(
     std::unique_ptr<internal::RunAsyncBase> function) {
   ++run_async_counter_;
-  // Some delay is necessary to reduce the likelihood that the timer expires
-  // before .then() is called.
-  // TODO(#10553): Refactor this mechanism to a more deterministic solution.
-  MakeRelativeTimer(std::chrono::milliseconds(500))
-      .then([f = std::move(function)](auto) { f->exec(); });
+  tq_->Schedule(std::chrono::system_clock::now(),
+                [f = std::move(function)](auto) { f->exec(); });
```
devbww, line 54: Perhaps we should have a `TimerQueue::Schedule(functor)` overload, which would do the same thing but without requiring the caller to utter `now()`.

[Aside: Even if such an overload was just a forwarder, it could avoid `now()` by using a time point like `std::chrono::system_clock::time_point::min()`, or something otherwise known to be in the past.]

Issue #8486, opened 2022-03-02
Use `GOOGLE_CLOUD_CPP_DEPRECATED` for deprecated RPCs


coryan:
RPCs can be annotated with `option deprecated = true`, e.g.:

https://github.com/googleapis/googleapis/blob/25d691b074e0b932e1c3f8a690ced8ddec8ae50e/google/container/v1/cluster_service.proto#L178-L194

For new libraries, we should not generate any RPCs that are deprecated, if necessary by adding them to the `omit_rpcs` list.  For existing GA libraries we should generate a `@deprecated` Doxygen comment, and (most likely) add the `GOOGLE_CLOUD_CPP_DEPRECATED` annotation, though this has problems with builds using `-Werror`.

devbww:
> For new libraries ...  For existing GA libraries ...

Right ... which in turn requires some way to distinguish those cases.  My only thought at the time was that any `option deprecated = true` RPC should be **required** to be listed in one of `omit_rpcs` (for new libraries) or a new "`emit_rpcs`" (for newly deprecated RPCs in existing libraries).

coryan:
We still want this, but do not have time.

coryan:
No time for this, to the backlog.

devbww:
@dbolduc points out in #10573 that we also need to be wary of deprecated message fields that appear in `google.api.method_signature` options, and hence in RPC overloads.

If the field is newly deprecated, we've probably previously released its overload, and so need to maintain that for backwards compatibility.  And in order to know that, we need long-term memory like the `emitted_rpcs` and `omitted_rpc` config elements.

And still, if we decide to generate code for a deprecated RPC, or an overload using a deprecated field, we want to add an `@deprecated` Doxygen comment and the `GOOGLE_CLOUD_CPP_DEPRECATED` annotation.

devbww:
Aside: Note that the proto compiler already adds `PROTOBUF_DEPRECATED` (`__attribute__((deprecated))`) to the accessors for deprecated message fields.

dbolduc:
Removing from backlog, because I duped the issue.

PR #10573, opened 2023-01-13
feat(generator): add some support for deprecated RPCs

devbww:
Do not generate APIs for deprecated RPCs unless their names appear in
an `emitted_rpcs` list, the reverse of the existing `omitted_rpcs`.

Furthermore, require that deprecated RPCs are listed in either of
`emitted_rpcs` or `omitted_rpcs`.  Use `emitted_rpcs` when the RPC
code has previously been generated (probably from before it was
deprecated), and so is needed for backwards compatibility.  Otherwise
use `omitted_rpcs`, typically for a new-generated service with
existing deprecated RPCs (which should be rare).

This will allow us to reliably omit generating APIs for deprecated
RPCs that we have never supported.

Part of #8486.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10573)

### DIFF
--- a/generator/generator_config.proto
+++ b/generator/generator_config.proto
@@ -20,15 +20,19 @@ message ServiceConfiguration {
   string service_proto_path = 1;
   string product_path = 2;
   string initial_copyright_year = 3;
+
+  // Omit the generation of the (sync) APIs for these RPCs. Include any
+  // deprecated RPCs here that we don't need for backwards compatibility.
   repeated string omitted_rpcs = 4;
+
   string service_endpoint_env_var = 5;
   string emulator_endpoint_env_var = 6;
 
-  // Generating Async APIs is typically unnecessary for RPCs that are
+  // Generating async APIs is typically unnecessary for RPCs that are
   // non-streaming, non-paginated, and not longrunning operations. If we need
-  // to generate an Async API for such an RPC, we can list it here.
+  // to generate an async API for such an RPC, we can list it here.
   //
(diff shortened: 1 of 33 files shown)


coryan (approved):
Very cool idea, I like it.

dbolduc (commented):
Thanks, I did not even realize we were generating deprecated methods as is. Something to think about, that shouldn't block this PR....

There is another piece to the puzzle, which is deprecated fields in a method signature: https://github.com/googleapis/google-cloud-cpp/blob/677c800829c1b39efe6a018c0b61e898c9960185/generator/internal/descriptor_utils.cc#L209-L213

I think we are still liable to lose a public client method if a field is newly deprecated, but the overall RPC is not. Can this be addressed with `emitted_rpcs`? I think at worst, we could do:

```
# generate the method signature for `GetFoo` at index 5.
emitted_rpcs: ["GetFoo.5"]
```

devbww:
> Something to think about, that shouldn't block this PR....
> 
> I think we are still liable to lose a public client method if a field is newly deprecated, but the overall RPC is not.

Yes, I agree that's an issue.  Perhaps we should expand #8486 to cover signature-method field deprecations in addition to RPC deprecations.  If you agree, I'll happily do that (if you don't first).

dbolduc:
> If you agree

I agree.

> (if you don't first)

I won't.

devbww:
> > (if you don't first)
> 
> I won't.

Done.

PR #10580, opened 2023-01-14
impl(otel): add CallContext

dbolduc:
Part of the work for #10489 

Add `CallContext`/`ScopedCallContext`, a generalization of per-operation configuration options. Now there could be a per-operation span as well.

The `google::cloud::Span` and `google::cloud::ScopedSpan` are defined outside of `internal/opentelemetry.h` as a way to avoid a future include cycle. (`StreamRange<>` will [need to include](https://github.com/googleapis/google-cloud-cpp/blob/1c6d329b013cab4587fb571fa081313f45a3f0c7/google/cloud/stream_range.h#L164) `CallContext()`, and I was planning on including `stream_range.h` in `internal/opentelemetry.h`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10580)


google/cloud/internal/call_context.h
```diff
@@ -0,0 +1,58 @@
...
+ * This class contains the configuration options set by the user. If tracing is
+ * enabled, it also contains the parent span that encompasses the client call.
+ */
+struct CallContext {
+  Options options = CurrentOptions();
+  Span span = CurrentSpan();
+};
+
+// For propagating context, typically across threads in async operations.
+class ScopedCallContext {
```
devbww, line 41: Should we give this some "should not/only be ..." attributes, like `OptionsSpan`?

google/cloud/internal/opentelemetry_aliases.h
```diff
@@ -0,0 +1,47 @@
...
+#include <opentelemetry/trace/scope.h>
+#include <opentelemetry/trace/span.h>
+#endif  // GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+// These aliases and functions are an organizational convenience while the
```
devbww, line 29: The function (`CurrentSpan()`) would stay either way, no?

google/cloud/internal/call_context.h
dbolduc: Done.

devbww (approved):

PR #10583, opened 2023-01-17
feat(generator): more support for deprecated RPC overloads

devbww:
Require that all "method_signature" overloads that use deprecated
fields appear in either the `emitted_rpcs` or `omitted_rpcs` lists.
This allows us to distinguish between RPC overloads using a newly-
deprecated field, which we want to keep generating for backwards
compatibility, and RPC overloads using deprecated fields in newly-
generated services, which we want to ignore.

Add `omitted_rpcs` entries for RPC overloads we currently ignore.

A subtle change is that we now compute the full "signature uid" for
all overloads, not just the un-deprecated ones.  This is a feature.
Without it, we ran the risk of choosing a new "first match" from a
clashing set should a field be newly deprecated, and therefore
silently changing the meaning of existing programs.

Add some simple support for allowing commas in the elements of
`emitted_rpcs` and `omitted_rpcs`.

Part of https://github.com/googleapis/google-cloud-cpp/issues/8486.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10583)

### DIFF
--- a/generator/generator_config.textproto
+++ b/generator/generator_config.textproto
@@ -259,7 +259,13 @@ service {
   service_proto_path: "google/bigtable/admin/v2/bigtable_table_admin.proto"
   product_path: "google/cloud/bigtable/admin"
   initial_copyright_year: "2021"
-  omitted_rpcs: ["CreateTableFromSnapshot", "SnapshotTable", "GetSnapshot", "ListSnapshots", "DeleteSnapshot"]
+  omitted_rpcs: [
+    "CreateTableFromSnapshot",
+    "SnapshotTable",
+    "GetSnapshot",
+    "ListSnapshots",
+    "DeleteSnapshot"
+  ]
   emulator_endpoint_env_var: "BIGTABLE_EMULATOR_HOST"
   gen_async_rpcs: ["CheckConsistency"]
   retryable_status_codes: ["kUnavailable", "kAborted"]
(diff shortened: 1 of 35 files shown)


scotthart (approved):
> Reviewed 35 of 35 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10583)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @devbww)

generator/generator_config.textproto
```diff
@@ -382,6 +388,31 @@ service {
   initial_copyright_year: "2022"
   retryable_status_codes: ["kUnavailable"]
   emitted_rpcs: ["SetLocations"]
+  omitted_rpcs: [
+    "CreateCluster(std::string const&, std::string const&, google::container::v1::Cluster const&)",
+    "UpdateCluster(std::string const&, std::string const&, std::string const&, google::container::v1::ClusterUpdate const&)",
+    "GetCluster(std::string const&, std::string const&, std::string const&)",
+    "ListClusters(std::string const&, std::string const&)",
```
dbolduc, line 395: Let's assume we have never generated the method because it has always contained a deprecated field.

The service team could introduce a `(std::string const&, std::string const&)` method signature that uses a different, non-deprecated field. We would want to generate that new method. But we would omit it, because the method signature is listed here.

I think we have to use the method signature index in some way to differentiate between these overload sets.

coryan: Hmmm... you mean `FooMethod(std::string const& field1)` and `FooMethod(std::string const& field2)`?  That is ambiguous and not implementable anyway, right?

dbolduc: If `field1` is deprecated, but `field2` is not, we should generate a `FooMethod(std::string const& field2)`. (currently, we would not generate either).

> That is ambiguous and not implementable anyway, right?

I think it can be implemented. The method signatures are ordered. The breaking change detector ensures that order is fixed (thanks to @tmatsuo).

It wouldn't work if a method signature can be removed from the original proto though (e.g. because the field is deprecated).

dbolduc: This is all small potatoes though. Customers can always call `FooMethod(FooRequest const&)` in these cases.

I am fine not doing anything about this edge case.

devbww: I agree that if we've never generated the "first match" overload (and therefore that it is a member of `omitted_rpcs`), that we could generate the second-match overload.

However, I would argue that we should still not generate the second-match overload, as you'd need to look at history (via the contents of the `emitted_rpcs`/`omitted_rpcs` lists) to be able to tell that this second match is really considered the first match.

That is, I'm claiming that we're all better off (customer included) if the "first match" is apparent directly from the proto.  Whether we generate that "real" first overload or not, then, is purely a question of backwards compatibility.

So, I vote for leaving things as they stand, no matter the "could"-ness.  Does that ring true to you?

dbolduc: > I vote for leaving things as they stand.

Me too.

<hr />

(For the record. And I am probably mischaracterizing your argument)

I don't agree with the statement: "we should not generate [because we] need to look at ... `emitted_rpcs`/`omitted_rpcs`".

I do agree with the statement: "there is not enough value to justify doing the work.", which I think you are implying.

devbww: > I don't agree with the statement: "we should not generate [because we] need to look at ... `emitted_rpcs`/`omitted_rpcs`".

Yes, I think that slightly mischaracterizes things, but we can chat later.

> I do agree with the statement: "there is not enough value to justify doing the work.", which I think you are implying.

No, it's not that I think there isn't enough value to do the work (the work would be easy, methinks), but rather that I think it would be a misstep.

So, I'll merge as-is.  I reckon that won't close any paths to "don't consider omitted signatures for first match", should we decide to go there in the future.

PR #10622, opened 2023-01-19
cleanup(otel): appease clang

dbolduc:
This was probably the right way to do the `ScopedSpan`. Otherwise clang complains about `ScopedCallContext::span_` being an unused member variable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10622)

### DIFF
--- a/google/cloud/internal/opentelemetry_aliases.h
+++ b/google/cloud/internal/opentelemetry_aliases.h
@@ -34,7 +34,9 @@ using Span = opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span>;
 using ScopedSpan = opentelemetry::trace::Scope;
 #else
 using Span = std::nullptr_t;
-using ScopedSpan = std::nullptr_t;
+struct ScopedSpan {
+  explicit ScopedSpan(Span) {}
+};
 #endif  // GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
 
 Span CurrentSpan();
--- a/google/cloud/internal/opentelemetry_aliases_test.cc
+++ b/google/cloud/internal/opentelemetry_aliases_test.cc
@@ -47,7 +47,6 @@ TEST(CurrentSpan, WithoutOpenTelemetry) {
   Span span = CurrentSpan();
   ScopedSpan scope(span);
   EXPECT_EQ(span, nullptr);
-  EXPECT_EQ(scope, nullptr);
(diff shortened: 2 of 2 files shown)


google/cloud/internal/opentelemetry_aliases.h
```diff
@@ -34,7 +34,9 @@ using Span = opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span>;
 using ScopedSpan = opentelemetry::trace::Scope;
 #else
 using Span = std::nullptr_t;
-using ScopedSpan = std::nullptr_t;
+struct ScopedSpan {
```
devbww, line 37: s/struct/class/?

google/cloud/internal/opentelemetry_aliases.h
```diff
@@ -34,7 +34,9 @@ using Span = opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span>;
 using ScopedSpan = opentelemetry::trace::Scope;
 #else
 using Span = std::nullptr_t;
```
devbww, line 36: Should we give `Span` the same treatment?  Rather than seeing this as just to "appease clang", think of it more as the right thing.  We wouldn't something like `span = scoped_span` to go undetected in some builds.

scotthart (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10622)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @dbolduc and @devbww)

___
*[`google/cloud/internal/opentelemetry_aliases_test.cc` line 48 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10622#-NMB7jgw9qxdRKXAY9fq:-NMB7jgw9qxdRKXAY9fr:b-2ki4ai) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/cfdac9b6ecaed99e40e40b0631841efa9cb69c0d/google/cloud/internal/opentelemetry_aliases_test.cc#L48)):*
> ```cpp
> TEST(CurrentSpan, WithoutOpenTelemetry) {
>   Span span = CurrentSpan();
>   ScopedSpan scope(span);
> ```

Remove this line too as it's now unused.

dbolduc:
I think #10626 is a better approach.

PR #10626, opened 2023-01-20
refactor: remove otel aliases

dbolduc:
I have decided (in my infinite wisdom) that this is the best way to organize/format the `CallContext` classes.

Originally, I did not want to use `#ifdef`s in a class that wasn't dedicated to `opentelemetry`. But after living with the "aliased types" implementation for a day, I think the "conditional member variables" implementation will be less burdensome.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10626)


devbww (approved):
> I have decided (in my infinite wisdom) that this is the best way to organize/format the `CallContext` classes.

OK.

Issue #10620, opened 2023-01-19
OpenTelemetry full tracing support for gRPC in generator


dbolduc:
This issue tracks full tracing support for our generated gRPC API surface. (Whereas #10491 is more about introducing the decorators). This means tracing the client operation **and** any subcomponents of that operation (e.g. each network call).

Subtasks not covered by other GH issues are listed under the API flavor.

Calls:
- [x] Sync Unary (non-paginated)
- [x] Paginated
- [x] Async Unary
  - [x] Update stubs to use `std::shared_ptr<grpc::ClientContext>`, instead of `unique_ptr`.
- [x] LRO
- [ ] #12361
- [x] Streaming Read
- [x] Streaming Write
- [x] Async Streaming Read/Write
- [x] Async Streaming Read
- [x] Async Streaming Write

dbolduc:
Closing, and we will track the one outstanding client call in a dedicated issue.

PR #10642, opened 2023-01-23
impl(otel): traced paginated connection

dbolduc:
Part of the work for #10620 

Add tracing for paginated calls at the connection level

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10642)


coryan (approved):
It would be nice to see how these look in Cloud Trace.

dbolduc:
> It would be nice to see how these look in Cloud Trace.

Ack, I should probably return to the exporter, so we can have integration tests.

At the moment they are not interesting. There is just a single span per call.

PR #10646, opened 2023-01-23
impl(otel): inject trace context for gRPC

dbolduc:
Part of the work for #10489 

Add `InjectTraceContext(...)`, which is used for context propagation. For now we only use the global propagator (exactly how we only use the global tracer provider). In the future it may become `Option` dependent.

If you are wondering when I will settle on the customer-facing options for tracing, tomorrow I will pick a date by which I must have made the decision.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10646)

### DIFF
--- a/google/cloud/internal/grpc_opentelemetry.cc
+++ b/google/cloud/internal/grpc_opentelemetry.cc
@@ -18,6 +18,8 @@
 #include "google/cloud/options.h"
 #include <grpcpp/grpcpp.h>
 #ifdef GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
+#include <opentelemetry/context/propagation/global_propagator.h>
+#include <opentelemetry/context/propagation/text_map_propagator.h>
 #include <opentelemetry/trace/semantic_conventions.h>
 #include <opentelemetry/trace/span_metadata.h>
 #include <opentelemetry/trace/span_startoptions.h>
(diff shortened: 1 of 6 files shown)


google/cloud/internal/grpc_opentelemetry.cc
```diff
@@ -30,6 +32,41 @@ namespace internal {
...
+class GrpcClientCarrier
+    : public opentelemetry::context::propagation::TextMapCarrier {
+ public:
+  explicit GrpcClientCarrier(grpc::ClientContext& context)
+      : context_(context) {}
+
+  // Unneeded by clients.
+  opentelemetry::nostd::string_view Get(
+      opentelemetry::nostd::string_view) const noexcept override {
+    return "";
```
coryan, line 56: Even though our code is client-only, somebody may use it in a server.  Does that matter?  What if they are using the functions framework and get a HTTP request with all the right tracing headers?

coryan (approved):

google/cloud/internal/grpc_opentelemetry.cc
dbolduc: > Does that matter?

I don't think so. This `Get()` would be used to continue some span started in a Google server, that was passed through the client context. That could be a thing, but I don't think it is at the moment.

> What if they are using the functions framework and get a HTTP request with all the right tracing headers?

Customers would:
1. extract the span context from the incoming HTTP request
2. start a span using that context (i.e. the new span would be a child of the remote span)
3. set that span as active (so the `google-cloud-cpp` spans will be a child of that span)

(It is the "Extract headers to context" part of this: https://opentelemetry.io/docs/instrumentation/cpp/manual/#context-propagation)

I am sure an example of how to do this would be helpful to the customers. But that is several milestones down the line.

coryan: SGTM

google/cloud/internal/grpc_opentelemetry.cc
```diff
@@ -30,6 +32,41 @@ namespace internal {
...
+  }
+
+  void Set(opentelemetry::nostd::string_view key,
+           opentelemetry::nostd::string_view value) noexcept override {
+    context_.AddMetadata(key.data(), value.data());
+  }
```
devbww, line 62: Using `sv.data()` without `sv.size()` is a red flag.

dbolduc: Ah, TIL. Thanks for checking and good catch.

I think that explains why I lost hours of time to this piece of code: https://github.com/dbolduc/google-cloud-cpp/blob/9e9691476a167626486dba50efab0a7eafef2334/google/cloud/trace/propagator/cloud_trace_context.cc#L62

PR #10651, opened 2023-01-24
impl(generator): add LRO endpoint option to rest stub factory

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10651)


coryan (commented):
Are you sure we need this, it seems that services using LROs just serve them out of their standard endpoint:

https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.backups.operations/list

coryan (approved):

scotthart (commented):
This is for the GetOperation and CancelOperation rpcs defined in `google/longrunning/operations.proto`. Spanner has a separate `google::longrunning::Operations::StubInterface` that gRPC generates. My understanding is that REST will need an analog for this.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10651#-:-NMZsQ960cGIN51gyZLA:b-f50u2w)* status: 0 of 3 files reviewed, all discussions resolved

PR #10663, opened 2023-01-25
cleanup(storage): move escaping functions to XmlNode impl

devbww:
Make the XML tag/text escaping functions an implementation detail of `XmlNode` rather than a separate library.  This continues the consolidation of the XML stuff, while its future is determined.

Fix a bug exposed by testing how `UnescapeXmlString()` is used, instead of only testing its result.

Also, use Abseil to handle all the "is space" decisions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10663)


google/cloud/storage/internal/xml_node.cc
```diff
@@ -208,12 +225,12 @@ class XmlParser {
     }
     if (c == '>') {
       auto close_tag = Rtrim(UnescapeXmlString(close_tag_));
-      if (current_parent_->GetTagName() != close_tag_) {
+      if (current_parent_->GetTagName() != close_tag) {
```
dbolduc, line 228: good catch

dbolduc (approved):

PR #10672, opened 2023-01-27
cleanup(storage): a handful of concluding cleanups for XmlParser

devbww:
[Reviewers: Yes, this is somewhat of a catchall, but it is convenient to gather all these together to finalize this work for now.]

Reformulate the `XmlParser` as a true state machine, where there are no branches on the next input char that are independent of the current state.  This means the current state can be represented simply by the handler function, without maintaining a parallel `enum`.

Correct a couple of transitions to the "base" state instead of the "end tag" state.  This should not affect a successful parse, but will improve the message given for errors after the transition.

There is no need to store `root_` and `current_parent_` separately when they can just be the bottom and top of `node_stack_`.

Accumulate the start/end tag names and the text content as views into the XML text, to eliminate copying into separate strings.

Check, when the input has been consumed, that we're not in the middle of an element.

Count the root node towards the `XmlParserMaxNodeDepth` limit.

Use `ASSERT_STATUS_OK(s)` rather than `ASSERT_TRUE(s.ok())` so that we get informative failures.

Use `ASSERT_STATUS_OK(s_or)` instead of `EXPECT_STATUS_OK(s_or)` when the next thing we would do is look at the `StatusOr`s value.

Rename all the `XmlNodeTest.Case` test cases to just `XmlNode.Case`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10672)

### DIFF
--- a/google/cloud/storage/internal/xml_node.cc
+++ b/google/cloud/storage/internal/xml_node.cc
@@ -37,6 +37,23 @@ std::string StripTrailingSpaces(absl::string_view str) {
   return std::string(absl::StripTrailingAsciiWhitespace(str));
 }
 
+// Should skip these during the parse instead, to avoid copying the document.
+std::string StripNonessential(absl::string_view document) {
+  static auto* nonessential_re = new std::regex{
+      "("
+      R"(<!DOCTYPE[^>[]*(\[[^\]]*\])?>)"  // DTD(DOCTYPE)
+      "|"
+      R"(<!\[CDATA\[[\s\S]*?\]\]>)"  // CDATA
+      "|"
+      R"(<!--[\s\S]*?-->)"  // XML comments
+      ")",
+      std::regex::icase | std::regex::nosubs | std::regex::optimize};
+  std::string essential;
+  std::regex_replace(std::back_inserter(essential), document.begin(),
+                     document.end(), *nonessential_re, "");
+  return essential;
+}
+
 std::string EscapeTagName(absl::string_view tag_name) {
   return absl::StrReplaceAll(tag_name, {{"&", "&amp;"},
                                         {"<", "&lt;"},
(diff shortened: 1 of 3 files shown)


google/cloud/storage/internal/xml_node.cc
```diff
@@ -63,232 +80,235 @@ std::string UnescapeTextContent(absl::string_view text_content) {
...
-    // parent if the limits allow.
-    auto text_node =
-        AppendTextNode(UnescapeTextContent(StripTrailingSpaces(text_content_)));
-    if (!text_node) return std::move(text_node).status();
+Status XmlParser::HandleBase() {
+  auto c = unparsed_.front();
+  if (!IsSpace(c)) {
+    if (c != '<') {
+      return internal::InvalidArgumentError(
+          absl::StrCat("Expected tag but found '", unparsed_.substr(0, 4), "'"),
```
dbolduc, line 179: I am guessing we include the next 4 characters because our test string `"text"` is 4 characters? :/

I am not sure what to return though. Maybe we also include the size of `unparsed_`, which would give an exact position of the error?

dbolduc (approved):

google/cloud/storage/internal/xml_node.cc
devbww: > I am guessing we include the next 4 characters because our test string `"text"` is 4 characters? :/

No ... I used to have it as only 1, but decided, from the look of the error message, that that didn't give enough context.  4 was just my next best guess.

> Maybe we also include the size of `unparsed_`, which would give an exact position of the error?

I did have `(unparsed_.data() - document_.data())` giving a position for a while, but soon realized this was mostly useless because of the RE stuff we do up front.  That is, the position we would give is probably unrelated to the position in the original string.

PR #10677, opened 2023-01-27
impl(otel): make tracing connection

dbolduc:
Part of the work for #10491 

Layer on the tracing connection, if the proper options are set. Note that the `Option` to enable tracing is still `internal`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10677)


generator/internal/connection_generator.cc
```diff
@@ -513,8 +513,9 @@ std::shared_ptr<$connection_class_name$> Make$connection_class_name$(
   auto background = internal::MakeBackgroundThreadsFactory(options)();
   auto stub = $product_internal_namespace$::CreateDefault$stub_class_name$(
     background->cq(), options);
-  return std::make_shared<$product_internal_namespace$::$connection_class_name$Impl>(
+  auto conn = std::make_shared<$product_internal_namespace$::$connection_class_name$Impl>(
       std::move(background), std::move(stub), std::move(options));
+  return $product_internal_namespace$::Make$tracing_connection_class_name$(std::move(conn));
```
coryan, line 518: You could say:

```cc
  return $product_internal_namespace$::Make$tracing_connection_class_name$(
      std::make_shared<$product_internal_namespace$::$connection_class_name$Impl>(
      std::move(background), std::move(stub), std::move(options)));
```

Assuming I counted the `)` correctly.

coryan (approved):

generator/internal/connection_generator.cc
dbolduc: Sure.

PR #10683, opened 2023-01-28
ci: no absl inline namespace (for check-api)

dbolduc:
Tell Abseil not to use inline namespaces in our cmake builds on fedora 36. One of which is `check-api`, which breaks when the Abseil LTS changes. We care that our API does not change, not our ABI.

Then update the ABI dumps. (the dumps were computed at HEAD).

I tested, and this will unblock #10680

AFAICT, this is harmless to do in our CI. That is, we do not need to split out `check-api` into its own Dockerfile. I will defer to the builds...

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10683)


ci/cloudbuild/dockerfiles/fedora-36-cmake.Dockerfile
```diff
@@ -61,10 +61,13 @@ RUN curl -sSL https://pkgconfig.freedesktop.org/releases/pkg-config-0.29.2.tar.g
     ldconfig
 ENV PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig:/usr/local/lib/pkgconfig:/usr/lib64/pkgconfig
 
+# We expose `absl::optional<>` in our public API. An Abseil LTS update will
+# break our `check-api` build, unless we disable the inline namespace.
 WORKDIR /var/tmp/build
 RUN curl -sSL https://github.com/abseil/abseil-cpp/archive/20220623.1.tar.gz | \
     tar -xzf - --strip-components=1 && \
     sed -i 's/^#define ABSL_OPTION_USE_\(.*\) 2/#define ABSL_OPTION_USE_\1 0/' "absl/base/options.h" && \
```
coryan, line 69: Unrelated, starting with this release of Abseil we should not need this.  Abseil install will do that themselves:

https://github.com/abseil/abseil-cpp/blob/75d2525117c8da93840ab256f07b191086fd7cbb/CMakeLists.txt#L229-L243

ci/cloudbuild/dockerfiles/fedora-36-cmake.Dockerfile
```diff
@@ -61,10 +61,13 @@ RUN curl -sSL https://pkgconfig.freedesktop.org/releases/pkg-config-0.29.2.tar.g
...
+    sed -i 's/^#define ABSL_OPTION_USE_INLINE_NAMESPACE 1$/#define ABSL_OPTION_USE_INLINE_NAMESPACE 0/' "absl/base/options.h" && \
```
coryan, line 70: This gave me a lot of pause, but I think you are right.  It is fine for CI builds, we should not show this in the `demo-*` Dockerfiles.

coryan (approved):

ci/cloudbuild/dockerfiles/fedora-36-cmake.Dockerfile
devbww: It gave me pause too, but the only problem I came up with was if, for some reason, we needed to use two Abseil releases at once.

I'll also assume that the distributions are different enough already that this does not affect caching for us.

PR #10688, opened 2023-01-28
cleanup: add some file-ending newlines

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10688)


coryan (approved):
Thanks, for what it is worth, I think I finally tamed my editor to insert these.

Issue #10254, opened 2022-11-15
Update CI builds to Fedora:37 before 2023-10


coryan:
We don't need to do all the builds at the same time.  Probably the first thing would be to update this description with the list of builds.  See #8922 for how we did this the last time.

PR #10701, opened 2023-01-31
cleanup(common): fix clang-tidy-15 warnings

coryan:
Part of the work for #10254

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10701)


devbww (approved):

google/cloud/internal/async_polling_loop.cc
```diff
@@ -47,7 +47,7 @@ class AsyncPollingLoopImpl
     auto const& options = CurrentOptions();
     promise_ = promise<StatusOr<Operation>>([w, options]() mutable {
       if (auto self = w.lock()) {
-        OptionsSpan span(std::move(options));
+        OptionsSpan const span(std::move(options));
```
dbolduc, line 50: I feel like there are many more instances with the same `OptionSpan` usage. e.g. https://github.com/googleapis/google-cloud-cpp/blob/c9b068b22cbe8805a7b05f6f758e740116d5282d/google/cloud/internal/async_rest_polling_loop.cc#L51

Do you know why it warns here and not elsewhere? Is it just an inconsistent warning?

coryan: Yes, there are many many more.  I am fixing some targets first (`google_cloud_cpp_{common,grpc_utils,testing_utils,testing_grpc,rest_internal}`.

There will be many more PRs like this one.

devbww: > There will be many more PRs like this one.

Oh.

What's the name of the tidy making the suggestion?  If its rationale is not particularly strong, I think I would vote for disabling it instead.

coryan: https://releases.llvm.org/15.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc/const-correctness.html

devbww: > This check implements detection of local variables which could be declared as const, but are not. Declaring variables as const is required or recommended by many coding guidelines

That indeed sounds like overkill to me.  What do you think?

Until `abseil-*` or `google-*` uses it, I'd vote for `-misc-const-correctness`.

coryan: Cannot say that I care enough to keep the warnings or silence them. Closing this and will send smaller PRs.

PR #10728, opened 2023-02-01
impl(spanner): add REST transport version of instance admin tests

scotthart:
In order to test gRPC/HTTP transcoding, we need to generate some REST versions of a service and run their integration tests. For this, we are starting with the `InstanceAdmin` service for Spanner.

The integration tests are the same ones used with the gRPC transport version of the generated service, but put into a new file as we needed some special handling to use the http version of the spanner emulator and we want a separate test target.

The REST versions of the `InstanceAdmin` and `DatabaseAdmin` libraries are built conditionally in cmake using a new `http-transcoding` feature label. For bazel, the new libraries are only built for integration test based builds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)


dbolduc (commented):
Drive by. I didn't look into the spanner-specific details. I left some questions about a gRPC vs. REST vs. gRPC + REST library.

(I am assuming that a generated REST only library would use the `Make*ConnectionRest()` function. If not, some of my questions are invalid.)

google/cloud/spanner/testing/cleanup_stale_instances.h
```diff
@@ -30,12 +32,20 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
  */
 Status CleanupStaleInstances(Project const& project);
 
+Status CleanupStaleInstances(
+    Project const& project, spanner_admin::InstanceAdminClient& client,
+    spanner_admin::DatabaseAdminClient& database_admin_client);
```
devbww, line 37: Pass clients by value.

google/cloud/spanner/testing/cleanup_stale_instances.cc
```diff
@@ -41,6 +41,19 @@ std::string CutoffDate() {
 }  // namespace
 
 Status CleanupStaleInstances(Project const& project) {
+  spanner_admin::InstanceAdminClient instance_admin_client(
+      spanner_admin::MakeInstanceAdminConnection());
+  spanner_admin::DatabaseAdminClient database_admin_client(
+      spanner_admin::MakeDatabaseAdminConnection());
+  return CleanupStaleInstances(project, instance_admin_client,
+                               database_admin_client);
```
devbww, line 49: `std::move` the clients, and then accept them by value (et seq).

google/cloud/spanner/testing/cleanup_stale_instances.cc
```diff
@@ -41,6 +41,19 @@ std::string CutoffDate() {
...
+}
+
+Status CleanupStaleInstances(
+    Project const& project,
+    spanner_admin::InstanceAdminClient& instance_admin_client,
+    spanner_admin::DatabaseAdminClient& database_admin_client) {
+  std::cout << __func__ << std::endl;
```
devbww, line 56: Debugging?

google/cloud/spanner/testing/cleanup_stale_instances.cc
```diff
@@ -58,14 +71,16 @@ Status CleanupStaleInstances(Project const& project) {
...
 
-  spanner_admin::InstanceAdminClient instance_admin_client(
-      spanner_admin::MakeInstanceAdminConnection());
   auto cutoff_date = CutoffDate();
   std::vector<std::string> instances;
   for (auto const& instance :
        instance_admin_client.ListInstances(project.FullName())) {
-    if (!instance) break;
+    if (!instance) {
+      std::cout << __func__ << " no instance" << std::endl;
```
devbww, line 79: Debugging?

google/cloud/spanner/testing/cleanup_stale_instances.cc
```diff
@@ -58,14 +71,16 @@ Status CleanupStaleInstances(Project const& project) {
...
+      break;
+    }
     auto name = instance->name();
+    std::cout << __func__ << " listed instance=" << name << std::endl;
```
devbww, line 83: Debugging?

google/cloud/spanner/testing/cleanup_stale_instances.cc
```diff
@@ -88,6 +101,13 @@ Status CleanupStaleInstances(Project const& project) {
 }
 
 Status CleanupStaleInstanceConfigs(Project const& project) {
+  spanner_admin::InstanceAdminClient instance_admin_client(
+      spanner_admin::MakeInstanceAdminConnection());
+  return CleanupStaleInstanceConfigs(project, instance_admin_client);
+}
+
+Status CleanupStaleInstanceConfigs(Project const& project,
+                                   spanner_admin::InstanceAdminClient& client) {
```
devbww, line 110: We might as well call this `instance_admin_client` to match.

google/cloud/spanner/testing/cleanup_stale_instances.cc
```diff
@@ -88,6 +101,13 @@ Status CleanupStaleInstances(Project const& project) {
 }
 
 Status CleanupStaleInstanceConfigs(Project const& project) {
+  spanner_admin::InstanceAdminClient instance_admin_client(
+      spanner_admin::MakeInstanceAdminConnection());
+  return CleanupStaleInstanceConfigs(project, instance_admin_client);
+}
```
devbww, line 107: I'd vote for not having these compatibility wrappers, and instead update the old calls to match the new style.

google/cloud/spanner/admin/integration_tests/instance_admin_rest_integration_test.cc
```diff
@@ -0,0 +1,430 @@
...
+using ::testing::AnyOf;
+using ::testing::EndsWith;
+using ::testing::Eq;
+using ::testing::HasSubstr;
+using ::testing::IsEmpty;
+using ::testing::Not;
+using ::testing::UnorderedElementsAre;
+
+std::string const& ProjectId() {
+  static std::string project_id =
```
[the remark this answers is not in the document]
devbww: FWIW, I'm not sure that tests need to play by the same rules.  That is, they should be multi-threaded at exit time.

google/cloud/spanner/admin/integration_tests/instance_admin_rest_integration_test.cc
```diff
@@ -0,0 +1,430 @@
...
+  return emulator;
+}
+
+// We need to override stub creation to use an alternate emulator endpoint for
+// the InstanceAdmin.
+std::shared_ptr<google::cloud::spanner_admin::InstanceAdminConnection>
+MakeInstanceAdminConnectionRestEmulator() {
+  Options options;
+  options =
+      spanner_admin_internal::InstanceAdminDefaultOptions(std::move(options));
```
devbww, line 88: Might as well just ...
```
  auto options = spanner_admin_internal::InstanceAdminDefaultOptions(Options{});
```

google/cloud/spanner/admin/integration_tests/instance_admin_rest_integration_test.cc
```diff
@@ -0,0 +1,430 @@
...
+bool Emulator() {
+  static bool emulator =
+      internal::GetEnv("SPANNER_EMULATOR_REST_HOST").has_value();
+  return emulator;
+}
+
+// We need to override stub creation to use an alternate emulator endpoint for
+// the InstanceAdmin.
+std::shared_ptr<google::cloud::spanner_admin::InstanceAdminConnection>
+MakeInstanceAdminConnectionRestEmulator() {
```
devbww, line 85: Why shouldn't `MakeInstanceAdminConnectionRest()` deal with all this?  (And similarly for `DatabaseAdmin`.)

google/cloud/spanner/admin/integration_tests/instance_admin_rest_integration_test.cc
```diff
@@ -0,0 +1,430 @@
+// Copyright 2023 Google LLC
```
devbww, line 1: Are we concerned about all the duplication between REST and non-REST hand-written code?

google/cloud/spanner/admin/integration_tests/instance_admin_rest_integration_test.cc
dbolduc: I think it is the `emulator_endpoint_end_var`.

https://github.com/googleapis/google-cloud-cpp/blob/10cfef588c819438b85c5a086fd9f44d8f087288/generator/generator_config.textproto#L1452

We could add a generator flag that supersedes that env var for REST... or split up the REST and gRPC into separate `service`s in the `generator_config.textproto`.

google/cloud/spanner/CMakeLists.txt
```diff
@@ -234,6 +234,54 @@ add_library(google-cloud-cpp::spanner ALIAS google_cloud_cpp_spanner)
 include(CreateBazelConfig)
 create_bazel_config(google_cloud_cpp_spanner YEAR "2019")
 
+if (GOOGLE_CLOUD_CPP_ENABLE_HTTP_TRANSCODING)
+    add_library(
+        google_cloud_cpp_spanner_rest # cmake-format: sort
+        admin/database_admin_connection.h
```
coryan, line 240: The `admin/database_admin_connection.h` and `admin/instance_admin_connection.h` headers also appear in the gRPC-based library.  That seems problematic: "I included a header, but the functions are in another library..."?

Maybe split the factory function?

cmake/GoogleCloudCppFeatures.cmake
```diff
@@ -225,6 +225,11 @@ function (google_cloud_cpp_enable_features)
             if (NOT ("storage" IN_LIST GOOGLE_CLOUD_CPP_ENABLE))
                 add_subdirectory(google/cloud/storage)
             endif ()
+        elseif ("${feature}" STREQUAL "http-transcoding")
+            option(
+                GOOGLE_CLOUD_CPP_ENABLE_HTTP_TRANSCODING
+                "Enable building gRPC/HTTP transcoding libraries and executing accompanying tests."
```
coryan, line 231: "executing accompanying tests"? I would just say "and their tests".

cmake/GoogleCloudCppFeatures.cmake
```diff
@@ -225,6 +225,11 @@ function (google_cloud_cpp_enable_features)
             if (NOT ("storage" IN_LIST GOOGLE_CLOUD_CPP_ENABLE))
                 add_subdirectory(google/cloud/storage)
             endif ()
+        elseif ("${feature}" STREQUAL "http-transcoding")
```
coryan, line 228: In this the final form of this feature?  If we have any doubts (e.g. how to support this in other services beyond spanner, where we need globs to work) then I would call it "experimental-http-transcoding".

google/cloud/spanner/CMakeLists.txt
```diff
@@ -234,6 +234,54 @@ add_library(google-cloud-cpp::spanner ALIAS google_cloud_cpp_spanner)
...
+                   VERSION "${PROJECT_VERSION}"
+                   SOVERSION "${PROJECT_VERSION_MAJOR}")
+    target_compile_options(google_cloud_cpp_spanner_rest
+                           PUBLIC ${GOOGLE_CLOUD_CPP_EXCEPTIONS_FLAG})
+    google_cloud_cpp_add_common_options(google_cloud_cpp_spanner_rest)
+    add_library(google-cloud-cpp::spanner_rest ALIAS
+                google_cloud_cpp_spanner_rest)
+    create_bazel_config(google_cloud_cpp_spanner_rest YEAR "2023")
+endif (GOOGLE_CLOUD_CPP_ENABLE_HTTP_TRANSCODING)
+
```
coryan, line 284: I would have expected changes below to get this library installed.  Or maybe we should reorganize the CMake code so it is easier to add everything about a library in a single place.

google/cloud/spanner/admin/instance_admin_connection.h
```diff
@@ -150,6 +150,33 @@ class InstanceAdminConnection {
...
+ * The optional @p options argument may be used to configure aspects of the
+ * returned `InstanceAdminConnection`. Expected options are any of the types in
+ * the following option lists:
+ *
+ * - `google::cloud::CommonOptionList`
+ * - `google::cloud::RestOptionList`
+ * - `google::cloud::UnifiedCredentialsOptionList`
+ * - `google::cloud::spanner_admin::InstanceAdminPolicyOptionList`
+ *
+ * @note Unexpected options will be ignored. To log unexpected options instead,
```
coryan, line 171: FYI: s/will be ignored/are ignored/ seems simpler.  Though maybe we should make a different PR to fix all of those.

google/cloud/spanner/admin/integration_tests/instance_admin_rest_integration_test.cc
coryan: Yes, I think we should change the existing integration tests if at all possible.

coryan (commented):
I think we should try to avoid the code duplication in the integration tests.

If you intend this new library to be used, we need to install it and we need a Bazel target with the right visibility at the top-level.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: 0 of 34 files reviewed, 6 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[`cmake/GoogleCloudCppFeatures.cmake` line 228 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9m-r1-228:-NNJjftQ1gTlsNpzFOU_:bfbr24z) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/cmake/GoogleCloudCppFeatures.cmake#L228)):*

In this the final form of this feature?  If we have any doubts (e.g. how to support this in other services beyond spanner, where we need globs to work) then I would call it "experimental-http-transcoding".
</blockquote></details>

added experimental to the label
___
*[`cmake/GoogleCloudCppFeatures.cmake` line 231 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9m-r1-231:-NNJji7jB7MX3pBPLhJz:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/cmake/GoogleCloudCppFeatures.cmake#L231)):*

"executing accompanying tests"? I would just say "and their tests".
</blockquote></details>

Done.
___
*[`google/cloud/spanner/CMakeLists.txt` line 240 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9p-r1-240:-NNJjjoh8M0vg7R3zLRH:b-ocmp0b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/google/cloud/spanner/CMakeLists.txt#L240)):*

The `admin/database_admin_connection.h` and `admin/instance_admin_connection.h` headers also appear in the gRPC-based library.  That seems problematic: "I included a header, but the functions are in another library..."?

Maybe split the factory function?

</blockquote></details>

Removed the connection headers. They did not need to be present in this library.
___
*[`google/cloud/spanner/CMakeLists.txt` line 284 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9p-r1-284:-NNJjslE877M1Rb9Oylj:bspmi6j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/google/cloud/spanner/CMakeLists.txt#L284)):*

I would have expected changes below to get this library installed.  Or maybe we should reorganize the CMake code so it is easier to add everything about a library in a single place.
</blockquote></details>

At this time, I don't want to install this library. It is for development testing purposes only.
___
*[`google/cloud/spanner/admin/database_admin_connection.h` line 164 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9q-r1-164:-NNJk0MUDT1E7fW7jeZp:b-hus8a8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/google/cloud/spanner/admin/database_admin_connection.h#L164)):*

what do we do about this in a REST only library? The `*ConnectionImpl` is unusable, right?

I am guessing we don't want to generate it.
</blockquote></details>

The generator will be adjusted accordingly when we get to that work effort.
___
*[`google/cloud/spanner/admin/instance_admin_connection.h` line 178 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9r-r1-178:-NNJkGvP4_omaqe0GXkU:b-e6jzqk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/google/cloud/spanner/admin/instance_admin_connection.h#L178)):*

What happens if I call this without enabling `http-transcoding`? Is it included in the doxygen?

Should we declare `Make*ConnectionRest()` in a separate generated header (e.g. `foo_rest_connection.h`)? Is there a benefit to declaring it in the same connection file?
</blockquote></details>

This is only generated for services that support rest transport as indicated in the generator_config.textproto.

generator/generator_config.textproto
```diff
@@ -1452,6 +1452,7 @@ service {
   emulator_endpoint_env_var: "SPANNER_EMULATOR_HOST"
   backwards_compatibility_namespace_alias: true
   retryable_status_codes: ["kUnavailable"]
+  generate_rest_transport: true
```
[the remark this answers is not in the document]
scotthart: As we add rest only services this field will likely change from a bool to some enum.

google/cloud/spanner/testing/cleanup_stale_instances.h
scotthart: Done.

google/cloud/spanner/testing/cleanup_stale_instances.cc
scotthart: Done.

google/cloud/spanner/testing/cleanup_stale_instances.cc
scotthart: Removed.

google/cloud/spanner/testing/cleanup_stale_instances.cc
scotthart: Removed.

google/cloud/spanner/testing/cleanup_stale_instances.cc
scotthart: Removed.

google/cloud/spanner/testing/cleanup_stale_instances.cc
scotthart: Done.

google/cloud/spanner/admin/integration_tests/instance_admin_rest_integration_test.cc
scotthart: Done.

google/cloud/spanner/admin/integration_tests/instance_admin_rest_integration_test.cc
scotthart: As Darren suggests, the issue with with needing an additional env var for the rest emulator port. As only spanner currently needs this workaround, I'm not planning on changing the generator at this time.

google/cloud/spanner/admin/integration_tests/instance_admin_rest_integration_test.cc
scotthart: I have reduced the amount of duplication such that there are tests sufficiency to verify that REST transport works a) with the emulator and b) with production, and we test LRO rpcs, paginated rpcs, and vanilla rpcs.

cmake/GoogleCloudCppFeatures.cmake
scotthart: Done.

scotthart:
I added a bazel flag to the rest integration test target. This flag is used to filter when bazel attempts to build/execute the tests. A new env var controls which bazel builds execute the rest integration test. I'm happy to adjust which builds enable it or even create a new build where rest tests would only be executed.

coryan (changes requested):
> Reviewed 27 of 34 files at r1, 13 of 14 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: all files reviewed, 4 unresolved discussions (waiting on @dbolduc and @scotthart)

___
*[`google/cloud/spanner/CMakeLists.txt` line 284 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9p-r1-284:-NNJloWS-dING92jKJ5e:b1e8oex) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/google/cloud/spanner/CMakeLists.txt#L284)):*

At this time, I don't want to install this library. It is for development testing purposes only.
</blockquote></details>

Fair enough.  But you are installing `admin/instance_admin_connection.h` and `admin/database_admin_connection.h` and both will declare a function that does not exist.  That seems bad for our customers.

I think we need to move the factory functions for the REST connections to a separate header file.
___
*[`google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJoErP08sk1uZ-yLvg:-NNJoErP08sk1uZ-yLvh:b1jvqmi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh#L42)):*
> ```Shell
> trap spanner_emulator::kill EXIT
> 
> test_tag_filters="integration-test,-http-transcoding-test"
> ```

How does this work?  I do not see any tests that has the `http-transcoding-test` tag.  And you also need to change the `run_integration_tests_emulator_cmake.sh`.

And for Bazel builds, is there a reason to just enable this for asan?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: all files reviewed, 4 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[`google/cloud/spanner/CMakeLists.txt` line 284 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9p-r1-284:-NNNObCM4mnJHe8VlAjy:b-1n5dbp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/google/cloud/spanner/CMakeLists.txt#L284)):*

Fair enough.  But you are installing `admin/instance_admin_connection.h` and `admin/database_admin_connection.h` and both will declare a function that does not exist.  That seems bad for our customers.

I think we need to move the factory functions for the REST connections to a separate header file.
</blockquote></details>

I was hoping to avoid adding any new "public" files for transcoding support. Would expanding the comments for this function to indicate it requires linking the rest library be sufficient?
If not, I propose creating an issue to track this change and I'll get to that next after this PR.
___
*[`google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJoErP08sk1uZ-yLvg:-NNNNyH6BrIp05We45Np:b-3k125p) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh#L42)):*

How does this work?  I do not see any tests that has the `http-transcoding-test` tag.  And you also need to change the `run_integration_tests_emulator_cmake.sh`.

And for Bazel builds, is there a reason to just enable this for asan?
</blockquote></details>

In `google/cloud/spanner/admin/integration_tests/BUILD.bazel` all tests in `spanner_admin_integration_tests_rest` have the `http-transocding-test` tag added.

Re: asan, I figured we want at least one build to execute these integration tests with the emulator. I'm happy to select a different build that accomplishes the same goal.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: all files reviewed, 4 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[`google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJoErP08sk1uZ-yLvg:-NNNPvJN7zp3eqUlgpix:b-1f3r9v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh#L42)):*

In `google/cloud/spanner/admin/integration_tests/BUILD.bazel` all tests in `spanner_admin_integration_tests_rest` have the `http-transocding-test` tag added.

Re: asan, I figured we want at least one build to execute these integration tests with the emulator. I'm happy to select a different build that accomplishes the same goal.
</blockquote></details>


As for cmake builds, it gets enabled by the `experimental-http-transcoding` feature that I added to the clang-tidy-build. It doesn't need an additional change locally.

coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: all files reviewed, 6 unresolved discussions (waiting on @dbolduc and @scotthart)

___
*[`ci/cloudbuild/build.sh` line 352 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNNXV2p5shiJ3AYX2pJ:-NNNXV2p5shiJ3AYX2pK:b4h4fgc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/ci/cloudbuild/build.sh#L352)):*
> ```Shell
>     # generator dir.
>     "--env=GENERATE_GOLDEN_ONLY=${GENERATE_GOLDEN_ONLY-}"
>     "--env=HTTP_TRANSCODING_TEST=${HTTP_TRANSCODING_TEST-}"
> ```

Why do we need to pass this environment variable?
___
*[`google/cloud/spanner/CMakeLists.txt` line 284 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9p-r1-284:-NNNQdZp7floBeGeopRS:bsd92zh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/google/cloud/spanner/CMakeLists.txt#L284)):*
> I was hoping to avoid adding any new "public" files for transcoding support. 

Hmmm... Okay, I see why that would seem like a good thing.

> Would expanding the comments for this function to indicate it requires linking the rest library be sufficient?

We would be telling people: "Link this library that we do not install and we are not planning to install either" (in the case of Bazel: "We are not planning to grant public visibility either").   That would make me hopping mad.

> If not, I propose creating an issue to track this change and I'll get to that next after this PR.

Maybe.... but it would be a release-blocking bug.  We rarely carry those (cannot recall ever doing so).  Seems better to fix the generator first and then rebase this PR on top of the newly generated code.
___
*[`google/cloud/spanner/admin/integration_tests/CMakeLists.txt` line 102 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNNTR28EaPB5w7jMLOq:-NNNTR28EaPB5w7jMLOr:b-pz762g) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/admin/integration_tests/CMakeLists.txt#L102)):*
> ```CMake
>         set_tests_properties(
>             ${target} PROPERTIES LABELS
>                                  "integration-test;integration-test-emulator")
> ```

Let's avoid more inconsistencies between CMake test labels and Bazel test  tags.  Please add the `http-transcoding*` tag here.
___
*[`google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJoErP08sk1uZ-yLvg:-NNNTi3c3iIx5O5OAQxr:b-tgojv3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh#L42)):*

As for cmake builds, it gets enabled by the `experimental-http-transcoding` feature that I added to the clang-tidy-build. It doesn't need an additional change locally.
</blockquote></details>

I see the new tags now, sorry I missed them.

FWIW, the clang-tidy build runs tests against the emulator too:

https://github.com/googleapis/google-cloud-cpp/blob/3757a3647ade6a491803c493fe6484285de73526/ci/cloudbuild/builds/clang-tidy.sh#L43

and then:

https://github.com/googleapis/google-cloud-cpp/blob/3757a3647ade6a491803c493fe6484285de73526/ci/cloudbuild/builds/lib/integration.sh#L274-L276

I am not sure if your intent was to run them in one CMake and one Bazel build.  I could be convinced that is needed, my prior is that once is enough for this.  If you just wanted a single build, you already achieved that without this change.

If you think we need to enable this for Bazel, I would like to avoid using an environment variable to control what gets to run (even if we do elsewhere, we should be trying to reduce the number of such problems).  It seems the `--test_tag_filters` tag here is redundant. Maybe setting it here would work better:

https://github.com/googleapis/google-cloud-cpp/blob/3757a3647ade6a491803c493fe6484285de73526/ci/cloudbuild/builds/lib/integration.sh#L165-L168

As we already pass them down:

https://github.com/googleapis/google-cloud-cpp/blob/3757a3647ade6a491803c493fe6484285de73526/ci/cloudbuild/builds/lib/integration.sh#L182-L184

dbolduc:
FTR, I think you can just remove the `Make*ConnectionRest` function.

https://github.com/googleapis/google-cloud-cpp/blob/e09eb4a1ca9dbfd8119406759730316d595d6320/generator/internal/connection_generator.cc#L269-L335

AFAICT it is not called in any golden unit tests, or used in the new integration tests.

google/cloud/spanner/admin/integration_tests/instance_admin_integration_test.cc
```diff
@@ -73,16 +73,23 @@ bool Emulator() {
 class CleanupStaleInstances : public ::testing::Environment {
  public:
   void SetUp() override {
-    EXPECT_STATUS_OK(
-        spanner_testing::CleanupStaleInstances(Project(ProjectId())));
+    spanner_admin::InstanceAdminClient instance_client(
```
devbww, line 76: We named these `instance_admin_client`, etc., in instance_admin_rest_integration_test.cc.  The fewer differences in the duplication the better I reckon.  That is, I'd like to be able to diff them and just see `s/Rest//`.

google/cloud/spanner/integration_tests/instance_admin_integration_test.cc
```diff
@@ -74,8 +74,13 @@ bool Emulator() {
 class CleanupStaleInstances : public ::testing::Environment {
  public:
   void SetUp() override {
-    EXPECT_STATUS_OK(
-        spanner_testing::CleanupStaleInstances(Project(ProjectId())));
+    spanner_admin::InstanceAdminClient instance_client(
```
devbww, line 77: Might as well rename these too.

google/cloud/spanner/testing/cleanup_stale_instances.cc
```diff
@@ -58,13 +61,13 @@ Status CleanupStaleInstances(Project const& project) {
...
-  spanner_admin::InstanceAdminClient instance_admin_client(
-      spanner_admin::MakeInstanceAdminConnection());
   auto cutoff_date = CutoffDate();
   std::vector<std::string> instances;
   for (auto const& instance :
        instance_admin_client.ListInstances(project.FullName())) {
-    if (!instance) break;
+    if (!instance) {
+      break;
+    }
```
devbww, line 70: Restore to just `if (!instance) break;` ... the fewer the diffs the better.

google/cloud/spanner/testing/cleanup_stale_instances.cc
```diff
@@ -87,7 +88,8 @@ Status CleanupStaleInstances(Project const& project) {
   return Status();
 }
 
-Status CleanupStaleInstanceConfigs(Project const& project) {
+Status CleanupStaleInstanceConfigs(Project const& project,
+                                   spanner_admin::InstanceAdminClient client) {
```
devbww, line 92: `s/client/instance_admin_client/`, which will eliminate some more edits below.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: all files reviewed, 11 unresolved discussions (waiting on @coryan, @dbolduc, and @devbww)

___
*[`ci/cloudbuild/build.sh` line 352 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNNXV2p5shiJ3AYX2pJ:-NNcbB6mBDKdkDvaerld:byowshk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/ci/cloudbuild/build.sh#L352)):*

Why do we need to pass this environment variable?
</blockquote></details>

It's the only way I know of to plumb the variable through to to docker.
___
*[`google/cloud/spanner/CMakeLists.txt` line 284 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9p-r1-284:-NNc_dfC-WXG5HqdsUCT:be55fq0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a519c7f15471c9fd40e5cec73f4bc5f203a58af7/google/cloud/spanner/CMakeLists.txt#L284)):*

> I was hoping to avoid adding any new "public" files for transcoding support. 

Hmmm... Okay, I see why that would seem like a good thing.

> Would expanding the comments for this function to indicate it requires linking the rest library be sufficient?

We would be telling people: "Link this library that we do not install and we are not planning to install either" (in the case of Bazel: "We are not planning to grant public visibility either").   That would make me hopping mad.

> If not, I propose creating an issue to track this change and I'll get to that next after this PR.

Maybe.... but it would be a release-blocking bug.  We rarely carry those (cannot recall ever doing so).  Seems better to fix the generator first and then rebase this PR on top of the newly generated code.

</blockquote></details>

Refactored into new generated files containing rest connection factory function.
___
*[`google/cloud/spanner/admin/integration_tests/CMakeLists.txt` line 102 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNNTR28EaPB5w7jMLOq:-NNc_lus5sIDatRSwStq:b-8b11j9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/admin/integration_tests/CMakeLists.txt#L102)):*

Let's avoid more inconsistencies between CMake test labels and Bazel test  tags.  Please add the `http-transcoding*` tag here.
</blockquote></details>

Added
___
*[`google/cloud/spanner/admin/integration_tests/instance_admin_integration_test.cc` line 76 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJl2ribumASbWU51LV-r3-76:-NNc_nG35Z3z0Pe2Rboe:boj5wv5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/admin/integration_tests/instance_admin_integration_test.cc#L76)):*

We named these `instance_admin_client`, etc., in instance_admin_rest_integration_test.cc.  The fewer differences in the duplication the better I reckon.  That is, I'd like to be able to diff them and just see `s/Rest//`.
</blockquote></details>

Edited for consistent naming.
___
*[`google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJoErP08sk1uZ-yLvg:-NNca8wy1QLPIiVbkUPA:boay50e) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh#L42)):*

I see the new tags now, sorry I missed them.

FWIW, the clang-tidy build runs tests against the emulator too:

https://github.com/googleapis/google-cloud-cpp/blob/3757a3647ade6a491803c493fe6484285de73526/ci/cloudbuild/builds/clang-tidy.sh#L43

and then:

https://github.com/googleapis/google-cloud-cpp/blob/3757a3647ade6a491803c493fe6484285de73526/ci/cloudbuild/builds/lib/integration.sh#L274-L276

I am not sure if your intent was to run them in one CMake and one Bazel build.  I could be convinced that is needed, my prior is that once is enough for this.  If you just wanted a single build, you already achieved that without this change.

If you think we need to enable this for Bazel, I would like to avoid using an environment variable to control what gets to run (even if we do elsewhere, we should be trying to reduce the number of such problems).  It seems the `--test_tag_filters` tag here is redundant. Maybe setting it here would work better:

https://github.com/googleapis/google-cloud-cpp/blob/3757a3647ade6a491803c493fe6484285de73526/ci/cloudbuild/builds/lib/integration.sh#L165-L168

As we already pass them down:

https://github.com/googleapis/google-cloud-cpp/blob/3757a3647ade6a491803c493fe6484285de73526/ci/cloudbuild/builds/lib/integration.sh#L182-L184

</blockquote></details>

I wanted the clang-tidy-build to make sure the code is tidy. Additionally, I wanted the rest code to run against production and the emulator, but only as many times as needed, so I enabled the test in the asan and integration-production builds.
I'm unclear what's being suggested regarding tag_filters as they are not appended to the args. Inside the spanner specific bazel ci script, tag_filters were already being set, I just added to them.
Lastly, I'm not sure how to limit the builds as I previously described without using some sort of variable that only those specific scripts set.
___
*[`google/cloud/spanner/integration_tests/instance_admin_integration_test.cc` line 77 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJl2ribumASbWU51LX-r3-77:-NNc_s_00-32nEW6M-To:b-q7hj08) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/integration_tests/instance_admin_integration_test.cc#L77)):*

Might as well rename these too.
</blockquote></details>

More naming consistency.
___
*[`google/cloud/spanner/testing/cleanup_stale_instances.h` line 34 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9v-r3-34:-NNc_ux8EWt9NcX5aCAo:bzgyihr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/testing/cleanup_stale_instances.h#L34)):*

`s/client/instance_admin_client/`
</blockquote></details>

Consistently being consistent.
___
*[`google/cloud/spanner/testing/cleanup_stale_instances.cc` line 70 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9u-r3-70:-NNc_yeQDj1b-CYRJHy8:b-bvwhm5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/testing/cleanup_stale_instances.cc#L70)):*

Restore to just `if (!instance) break;` ... the fewer the diffs the better.
</blockquote></details>

Restored.
___
*[`google/cloud/spanner/testing/cleanup_stale_instances.cc` line 92 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNEH6YafSwKxpesHE9u-r3-92:-NNca1GC9wTxdtoPMUPO:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/testing/cleanup_stale_instances.cc#L92)):*

`s/client/instance_admin_client/`, which will eliminate some more edits below.
</blockquote></details>

Done.

google/cloud/spanner/admin/instance_admin_rest_connection.h
```diff
@@ -0,0 +1,65 @@
...
+ *     set `GOOGLE_CLOUD_CPP_ENABLE_CLOG=yes` in the environment.
+ *
+ * @param options (optional) Configure the `InstanceAdminConnection` created by
+ * this function.
+ */
+std::shared_ptr<InstanceAdminConnection> MakeInstanceAdminConnectionRest(
+    Options options = {});
+
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
+namespace gcpcxxV1 = GOOGLE_CLOUD_CPP_NS;  // NOLINT(misc-unused-alias-decls)
```
devbww, line 60: This is a bit of a funny one, where we're generating backwards compatibility namespaces for a brand new interface.

google/cloud/spanner/admin/instance_admin_rest_connection.cc
```diff
@@ -0,0 +1,55 @@
...
+      rest_internal::AutomaticallyCreatedRestBackgroundThreads>();
+  auto stub =
+      spanner_admin_internal::CreateDefaultInstanceAdminRestStub(options);
+  return std::make_shared<
+      spanner_admin_internal::InstanceAdminRestConnectionImpl>(
+      std::move(background), std::move(stub), std::move(options));
+}
+
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
+namespace gcpcxxV1 = GOOGLE_CLOUD_CPP_NS;  // NOLINT(misc-unused-alias-decls)
```
devbww, line 52: These seem useless in the cc files (and not just for new APIs).

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: 29 of 45 files reviewed, 13 unresolved discussions (waiting on @coryan, @dbolduc, and @devbww)

___
*[`google/cloud/spanner/admin/instance_admin_rest_connection.h` line 60 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNcbnZrZ70wf7Cjh6CT-r4-60:-NNcqBydARDnn8FXEZss:b6ocqjj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c5fa464398c266559a53f78446c69433575cdb4/google/cloud/spanner/admin/instance_admin_rest_connection.h#L60)):*

This is a bit of a funny one, where we're generating backwards compatibility namespaces for a brand new interface.
</blockquote></details>

This is controlled by a service level setting, thus any files emitted from the spanner instance admin service get the backwards compatibility treatment.
___
*[`google/cloud/spanner/admin/instance_admin_rest_connection.cc` line 52 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNcbnZrZ70wf7Cjh6CS-r4-52:-NNcqYp-7LSDdnsU8Xbf:bhuuvtv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c5fa464398c266559a53f78446c69433575cdb4/google/cloud/spanner/admin/instance_admin_rest_connection.cc#L52)):*

These seem useless in the cc files (and not just for new APIs).
</blockquote></details>

Do we want to open an issue to clean this up?

coryan (changes requested):
> Reviewed 16 of 16 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: all files reviewed, 11 unresolved discussions (waiting on @dbolduc, @devbww, and @scotthart)

___
*[`google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJoErP08sk1uZ-yLvg:-NNcvzEW2r1xSm8Jftlr:bbok0zl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh#L42)):*
> I wanted the clang-tidy-build to make sure the code is tidy. 

Fair enough.

> Additionally, I wanted the rest code to run against production and the emulator, but only as many times as needed, so I enabled the test in the asan and integration-production builds.

Yes, but I think `clang-tidy` **already** runs against the emulator.  So you are running it twice against the emulator.

> I'm unclear what's being suggested regarding tag_filters as they are not appended to the args. Inside the spanner specific bazel ci script, tag_filters were already being set, I just added to them.

Yes, and this is confusing enough already.   The `integration.sh` script already passes some arguments down to `spanner/ci/run_integration_tests_emulator_bazel.sh`.  Then that script changes some stuff when the test is `msan`.  We should not add to that complexity, we should reduce it.

> Lastly, I'm not sure how to limit the builds as I previously described without using some sort of variable that only those specific scripts set.

The scripts could pass the right `--test_tag_filters` to begin with.  For example, you could change `asan.sh` to do this instead:

```sh
integration::bazel_with_emulators test "${args[@]}" "${integration_args[@]}" --test_tag_filters=integration-test,http-transcoding-test
```

and we should change `msan.sh` to do this instead:

```
integration::bazel_with_emulators test "${args[@]}" "${integration_args[@]}" --test_tag_filters=integration-test,-http-transconding,-no-msan
```

That may require changing `integration.sh` to remove the code greping for `--config=msan` and its use, e.g. from:

```sh
  tag_filters="integration-test"
  if echo "${args[@]}" | grep -w -q -- "--config=msan"; then
    tag_filters="integration-test,-no-msan"
  fi

  io::log_h2 "Running integration tests that require production access"
  bazel "${verb}" "${args[@]}" --test_tag_filters="${tag_filters}" \
    "${production_integration_tests[@]}"
```

To simply:

```sh
  io::log_h2 "Running integration tests that require production access"
  bazel "${verb}" --test_tag_filters=integration-test "${args[@]}"  \
    "${production_integration_tests[@]}"

...
..
```

 This approach depends on `--test_tag_filters` not accumulating (the last value wins).  Which seems to be what happens (see bazelbuild/bazel#7322).

In other words: if a build wants to apply tag filters it needs to supply them in the arguments it passes to `integration::bazel_with_emulators`.  Removing magic based on `grep` and avoiding environment variables seems like a good thing.

coryan:
*[`ci/cloudbuild/build.sh` line 352 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNNXV2p5shiJ3AYX2pJ:-NNdZ3KF2YhL9KdB107p:bys4j7k) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/ci/cloudbuild/build.sh#L352)):*

It's the only way I know of to plumb the variable through to to docker.
</blockquote></details>

Yes, but why?  `asan.sh` runs *inside* docker.  You only need to use an environment variable if you are planning to set this when running builds manually.  Is that the intent?  Why do we need to do that?

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: all files reviewed, 7 unresolved discussions (waiting on @dbolduc and @scotthart)

___
*[`google/cloud/spanner/admin/instance_admin_rest_connection.h` line 60 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNcbnZrZ70wf7Cjh6CT-r4-60:-NNevw5uB4jDjn5NBIKv:b-jbbnl0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c5fa464398c266559a53f78446c69433575cdb4/google/cloud/spanner/admin/instance_admin_rest_connection.h#L60)):*

This is controlled by a service level setting, thus any files emitted from the spanner instance admin service get the backwards compatibility treatment.
</blockquote></details>

https://github.com/googleapis/google-cloud-cpp/pull/10764 out for review, which would remove these from REST interfaces.
___
*[`google/cloud/spanner/admin/instance_admin_rest_connection.cc` line 52 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNcbnZrZ70wf7Cjh6CS-r4-52:-NNew7z6EdPYWNRt_WRj:bp8p9kz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c5fa464398c266559a53f78446c69433575cdb4/google/cloud/spanner/admin/instance_admin_rest_connection.cc#L52)):*

Do we want to open an issue to clean this up?
</blockquote></details>

https://github.com/googleapis/google-cloud-cpp/pull/10764 out for review, which would remove these from cc files.
___
*[`google/cloud/spanner/integration_tests/instance_admin_integration_test.cc` line 77 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJl2ribumASbWU51LX-r3-77:-NNewRJp5ug-KFs7yOQX:bh38cvm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/integration_tests/instance_admin_integration_test.cc#L77)):*

More naming consistency.
</blockquote></details>

`<xxx>_admin_client`?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728)* status: 16 of 46 files reviewed, 7 unresolved discussions (waiting on @coryan, @dbolduc, and @devbww)

___
*[`google/cloud/spanner/admin/instance_admin_rest_connection.h` line 60 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNcbnZrZ70wf7Cjh6CT-r4-60:-NNh7PjI02nqPFBLwqOy:b-mcn9ex) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c5fa464398c266559a53f78446c69433575cdb4/google/cloud/spanner/admin/instance_admin_rest_connection.h#L60)):*

https://github.com/googleapis/google-cloud-cpp/pull/10764 out for review, which would remove these from REST interfaces.
</blockquote></details>

Regenerated.
___
*[`google/cloud/spanner/admin/instance_admin_rest_connection.cc` line 52 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNcbnZrZ70wf7Cjh6CS-r4-52:-NNh7RMnCdY22EG91KIt:b-mcn9ex) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8c5fa464398c266559a53f78446c69433575cdb4/google/cloud/spanner/admin/instance_admin_rest_connection.cc#L52)):*

https://github.com/googleapis/google-cloud-cpp/pull/10764 out for review, which would remove these from cc files.
</blockquote></details>

Regenerated.
___
*[`google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJoErP08sk1uZ-yLvg:-NNh7TM13D_wjH0-sEnO:byty7qb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/ci/run_integration_tests_emulator_bazel.sh#L42)):*

> I wanted the clang-tidy-build to make sure the code is tidy. 

Fair enough.

> Additionally, I wanted the rest code to run against production and the emulator, but only as many times as needed, so I enabled the test in the asan and integration-production builds.

Yes, but I think `clang-tidy` **already** runs against the emulator.  So you are running it twice against the emulator.

> I'm unclear what's being suggested regarding tag_filters as they are not appended to the args. Inside the spanner specific bazel ci script, tag_filters were already being set, I just added to them.

Yes, and this is confusing enough already.   The `integration.sh` script already passes some arguments down to `spanner/ci/run_integration_tests_emulator_bazel.sh`.  Then that script changes some stuff when the test is `msan`.  We should not add to that complexity, we should reduce it.

> Lastly, I'm not sure how to limit the builds as I previously described without using some sort of variable that only those specific scripts set.

The scripts could pass the right `--test_tag_filters` to begin with.  For example, you could change `asan.sh` to do this instead:

```sh
integration::bazel_with_emulators test "${args[@]}" "${integration_args[@]}" --test_tag_filters=integration-test,http-transcoding-test
```

and we should change `msan.sh` to do this instead:

```
integration::bazel_with_emulators test "${args[@]}" "${integration_args[@]}" --test_tag_filters=integration-test,-http-transconding,-no-msan
```

That may require changing `integration.sh` to remove the code greping for `--config=msan` and its use, e.g. from:

```sh
  tag_filters="integration-test"
  if echo "${args[@]}" | grep -w -q -- "--config=msan"; then
    tag_filters="integration-test,-no-msan"
  fi

  io::log_h2 "Running integration tests that require production access"
  bazel "${verb}" "${args[@]}" --test_tag_filters="${tag_filters}" \
    "${production_integration_tests[@]}"
```

To simply:

```sh
  io::log_h2 "Running integration tests that require production access"
  bazel "${verb}" --test_tag_filters=integration-test "${args[@]}"  \
    "${production_integration_tests[@]}"

...
..
```

 This approach depends on `--test_tag_filters` not accumulating (the last value wins).  Which seems to be what happens (see bazelbuild/bazel#7322).

In other words: if a build wants to apply tag filters it needs to supply them in the arguments it passes to `integration::bazel_with_emulators`.  Removing magic based on `grep` and avoiding environment variables seems like a good thing.

</blockquote></details>

Removed env var from  docker. Only `clang-tidy` and `integration-production` build and test rest transcoding. Moved spanner `test_tag_filter` settings to `integration.sh`, left the rest for a separate PR. 
___
*[`google/cloud/spanner/integration_tests/instance_admin_integration_test.cc` line 77 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNJl2ribumASbWU51LX-r3-77:-NNh7vf2-gTYRq1drTHh:b-sc6d5x) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/google/cloud/spanner/integration_tests/instance_admin_integration_test.cc#L77)):*

`<xxx>_admin_client`?
</blockquote></details>

Fixed the other, other, `instance_admin_integration_test.cc` file.
___
*[`ci/cloudbuild/build.sh` line 352 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-NNNXV2p5shiJ3AYX2pJ:-NNh81KlDTMgWT2m7WwT:bss4a2b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a67e1145e3493bbb0f7e63edf30a7a610cbfd08a/ci/cloudbuild/build.sh#L352)):*

Yes, but why?  `asan.sh` runs *inside* docker.  You only need to use an environment variable if you are planning to set this when running builds manually.  Is that the intent?  Why do we need to do that?
</blockquote></details>

Removed.

coryan (approved):
:lgtm:

> Reviewed 30 of 30 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10728#-:-NNhAk9f7eMUe0XiRSAQ:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on @dbolduc and @devbww)

devbww (approved):

PR #10731, opened 2023-02-02
test: AsyncRetryLoop cancel options

dbolduc:
Verify the behavior of this line of code, with a test that would fail if it were removed: https://github.com/googleapis/google-cloud-cpp/blob/10cfef588c819438b85c5a086fd9f44d8f087288/google/cloud/internal/async_retry_loop.h#L194

Also, rearrange the order of the `ConfigureContext` test, so that the `TEST()`s are not broken up by `TEST_F()`s.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10731)

### DIFF
--- a/google/cloud/internal/async_retry_loop_test.cc
+++ b/google/cloud/internal/async_retry_loop_test.cc
@@ -391,6 +391,47 @@ TEST(AsyncRetryLoopTest, SetsTimeout) {
   ASSERT_THAT(actual.status(), StatusIs(StatusCode::kUnavailable));
 }
 
+TEST(AsyncRetryLoopTest, ConfigureContext) {
+  AsyncSequencer<StatusOr<int>> sequencer;
+
+  // The original options should be used in the first attempt and in the retry
+  // attempt.
+  MockFunction<void(grpc::ClientContext&)> setup;
+  EXPECT_CALL(setup, Call).Times(2);
+  OptionsSpan span(Options{}.set<GrpcSetupOption>(setup.AsStdFunction()));
+
+  AutomaticallyCreatedBackgroundThreads background;
+  future<StatusOr<int>> actual = AsyncRetryLoop(
+      TestRetryPolicy(), TestBackoffPolicy(), Idempotency::kIdempotent,
+      background.cq(),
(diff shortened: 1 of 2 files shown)


google/cloud/internal/async_retry_loop_test.cc
```diff
@@ -391,6 +391,47 @@ TEST(AsyncRetryLoopTest, SetsTimeout) {
...
+      "error message");
+
+  // Clear the current options before retrying.
+  OptionsSpan clear(Options{});
+  sequencer.PopFront().set_value(Status(StatusCode::kUnavailable, "try again"));
+  sequencer.PopFront().set_value(0);
+  (void)actual.get();
+}
+
+TEST(AsyncRetryLoopTest, CallOptionsDuringCancel) {
```
devbww, line 417: Why are `AsyncRetryLoopTest.CallOptionsDuringCancel` and `AsyncRestRetryLoopTest.CallOptionsDuringCancel` subtly different in ways that seem unrelated to REST?  We should strive to make the duplication as obvious as possible.

Indeed perhaps we'd be better off interleaving the `Foo.Bar` and `FooRest.Bar` cases, so that the duplication is localized.

dbolduc: > subtly different

Fixed. The differences are just the `*Rest*` invocations / `rest_internal` namespace.

> perhaps we'd be better off interleaving...

Maybe. It is not something I plan to do.

devbww: > Fixed. The differences are just the `*Rest*` invocations / `rest_internal` namespace.

It is a shame about the namespaces.  In the bigger picture, we should probably avoid `namespace [foo_]internal { }` wrapping in tests, I reckon.

devbww (approved):

google/cloud/internal/async_retry_loop_test.cc
dbolduc: Interesting, but I reckon that using the `*_internal` namespace produces more readable tests in many cases.

PR #10735, opened 2023-02-02
impl(otel): process gRPC Client Context

dbolduc:
Part of the work for #10489 

Introduce a method to extract relevant tracing info from a `grpc::ClientContext`.

The RPC Semantic Conventions would have us split up the peer URI into a port, IP address, IP version. Take on some technical debt by not initially doing this work.

gRPC has helpers to parse the URI, but they are an internal implementation. So leave a note for ourselves on where we can find a way to parse the URI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10735)

### DIFF
--- a/google/cloud/internal/grpc_opentelemetry.h
+++ b/google/cloud/internal/grpc_opentelemetry.h
@@ -65,8 +65,7 @@ void InjectTraceContext(grpc::ClientContext& context, Options const& options);
  * The span is ended. The original value is returned, for the sake of
  * composition.
  *
- * Note that this function should be called after the server has filled in the
- * relevant information, such as the peer.
+ * Note that this function should be called after the RPC has finished.
  */
 template <typename T>
 T EndSpan(grpc::ClientContext& context, opentelemetry::trace::Span& span,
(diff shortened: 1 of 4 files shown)


google/cloud/testing_util/opentelemetry_matchers.h
```diff
@@ -108,6 +108,13 @@ ::testing::Matcher<SpanDataPtr> SpanHasAttributes(Args const&... matchers) {
       ::testing::IsSupersetOf({matchers...}));
 }
 
+template <typename T>
+::testing::Matcher<
+    std::pair<std::string, opentelemetry::sdk::common::OwnedAttributeValue>>
+SpanAttribute(std::string const& key) {
```
devbww, line 114: Would it be clearer to require the user to utter the `_`?

dbolduc: Yes. I wanted to do this, but gave up too early on making it work. Fixed.

google/cloud/internal/grpc_opentelemetry.h
```diff
@@ -58,6 +59,24 @@ opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> MakeSpanGrpc(
  */
 void InjectTraceContext(grpc::ClientContext& context, Options const& options);
 
+/**
+ * Extracts information from the `grpc::ClientContext`, and adds it to a span.
+ *
+ * The span is ended. The original value is returned, for the sake of
+ * composition.
+ *
+ * Note that this function should be called after the server has filled in the
```
devbww, line 68: Offline we discussed something like "after the RPC has finished" instead of "after the server has filled ...", seeing as the server isn't doing anything on the client side.

devbww (approved):

PR #10740, opened 2023-02-03
ci: switch integration-daily build to Fedora:37

coryan:
Part of the work for #10254

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10740)

### DIFF
--- a/ci/cloudbuild/builds/integration-daily.sh
+++ b/ci/cloudbuild/builds/integration-daily.sh
@@ -23,7 +23,7 @@
 # `ci/cloudbuild/triggers` directory. "Manual" triggers exist only within the
 # GCB UI at the time of this writing. Users with the appropriate access can run
 # this build by hand with:
-#   `ci/cloudbuild/build.sh --distro fedora-36-bazel --build integration-daily --cloud cloud-cpp-testing-resources`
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/builds/integration-daily.sh
```diff
@@ -23,7 +23,7 @@
 # `ci/cloudbuild/triggers` directory. "Manual" triggers exist only within the
 # GCB UI at the time of this writing. Users with the appropriate access can run
 # this build by hand with:
-#   `ci/cloudbuild/build.sh --distro fedora-36-bazel --build integration-daily --cloud cloud-cpp-testing-resources`
+#   `ci/cloudbuild/build.sh --distro fedora-37-bazel --build integration-daily --cloud cloud-cpp-testing-resources`
```
devbww, line 26: I don't see the trigger change, but I presume you're doing that after this is in.

While you're there, it still has a "fedora-35" tag.

devbww (approved):

ci/cloudbuild/builds/integration-daily.sh
coryan: Or nearly at the same time.  Fixed.

devbww: It looks like the trigger went back to 35 instead of forward to 37.

PR #10761, opened 2023-02-06
impl(generator): disable deprecation warnings in generated Client code

devbww:
When a message field used in a `google.api.method_signature` is newly
deprecated, requiring an addition to `emitted_rpcs` for backwards
compatibility, the generated code will elicit a “'field' is deprecated”
warning.  So, disable those warnings (which become errors) in Client
implementations in anticipation.

Aside: Perhaps we should just do this for all generated code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10761)

### DIFF
--- a/generator/internal/client_generator.cc
+++ b/generator/internal/client_generator.cc
@@ -340,6 +340,9 @@ Status ClientGenerator::GenerateCc() {
 
   // includes
   CcPrint("\n");
+  // We will access a deprecated field when a google::api::method_signature
+  // overload appears in `emitted_rpcs`, so ...
+  CcLocalIncludes({"google/cloud/internal/disable_deprecation_warnings.inc"});
   CcLocalIncludes({vars("client_header_path")});
   CcSystemIncludes({"memory"});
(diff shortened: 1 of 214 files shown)


generator/internal/client_generator.cc
```diff
@@ -340,6 +340,9 @@ Status ClientGenerator::GenerateCc() {
 
   // includes
   CcPrint("\n");
+  // We will access a deprecated field when a google::api::method_signature
```
coryan, line 343: This suggests that you know when to add this `#include` and when to omit it.  Your call.

coryan (approved):

generator/internal/client_generator.cc
devbww: Done.  Which means this now only affects the `GoldenKitchenSinkClient`, where we have manufactured the circumstance.  No real services currently have emitted overloads using deprecated fields.  PTAL.

coryan (approved):

PR #10763, opened 2023-02-07
impl(docfx): scaffolding for conversion functions

coryan:
This PR introduces the first function to convert simple paragraphs in Doxygen XML to MarkDown strings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10763)

### DIFF
--- a/docfx/doxygen2markdown.cc
+++ b/docfx/doxygen2markdown.cc
@@ -22,20 +22,15 @@ namespace {
 [[noreturn]] void UnknownChildType(std::string_view where,
                                    pugi::xml_node const& child) {
   std::ostringstream os;
-  os << "Unknown child in " << where << "(): name=" << child.name();
-  os << ", attributes=[";
-  for (auto const& a : child.attributes()) {
-    os << ", " << a.name() << "=" << a.as_string();
-  }
-  os << "]";
-  child.print(os);
+  os << "Unknown child in " << where << "(): node=";
+  child.print(os, /*indent=*/"", /*flags=*/pugi::format_raw);
   throw std::runtime_error(std::move(os).str());
 }
 
 }  // namespace
 
 bool AppendIfPlainText(std::ostream& os, pugi::xml_node const& node) {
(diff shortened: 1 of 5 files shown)


docfx/doxygen2markdown.cc
```diff
@@ -0,0 +1,60 @@
...
+
+namespace {
+
+[[noreturn]] void UnknownChildType(std::string_view where,
+                                   pugi::xml_node const& child) {
+  std::ostringstream os;
+  os << "Unknown child in " << where << "(): name=" << child.name();
+  os << ", attributes=[";
+  for (auto const& a : child.attributes()) {
+    os << ", " << a.name() << "=" << a.as_string();
```
devbww, line 28: Avoid separator on first attribute?

docfx/doxygen2markdown.cc
```diff
@@ -0,0 +1,60 @@
...
+bool AppendIfPlainText(std::ostream& os, pugi::xml_node const& node) {
+  if (node.name() != std::string_view{} || !node.attributes().empty()) {
+    return false;
+  }
+  os << node.value();
+  return true;
+}
+
+bool AppendIfComputerOutput(std::ostream& os, pugi::xml_node const& node) {
+  if (node.name() != std::string_view{"computeroutput"}) return false;
```
devbww, line 46: If you haven't already considered it, I'm wondering whether `std::string_view(node.name()) != "literal"` might be clearer than `node.name() != std::string_view("literal")`.  I'm not sure, so your choice.

devbww (approved):

docfx/doxygen2markdown.cc
coryan: Good idea.  I am just going to rely on `child.print(os);` to do all the work (which it does).

docfx/doxygen2markdown.cc
coryan: Good idea, done.

PR #10765, opened 2023-02-07
cleanup: simplify Bazel dependency loads

coryan:
Take advantage of the `maybe()` utility to load dependencies only if they are defined.  I appeased `buildifier` by adding an (unused) `name` argument to the `google_cloud_cpp_deps()` macro, and documenting why it is there.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10765)

### DIFF
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -14,9 +14,10 @@
 
 """Load dependencies needed to compile and test the google-cloud-cpp library."""
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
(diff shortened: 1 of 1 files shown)


devbww (approved):

dbolduc:
Hmm... I think this may be throwing renovate-bot

coryan:
> Hmm... I think this may be throwing renovate-bot

Say more words... Why?  What do you think is broken?  I am guessing you are referring to the latest gRPC update, which is missing the changes to gRPC?  If so, maybe (he he) I figured it out.  Renovate seems to want the rules to be written without `repo_rule=`:

https://github.com/renovatebot/renovate/blob/e41681db3d2b4fc2ee47eaa2c6ad5b6a5d82b189/lib/modules/manager/bazel/parser.ts#L264-L266

I will send a CL so we can try that, and then we can rollback if it does not work.

dbolduc:
Ack. Thanks for fixing.

PR #10777, opened 2023-02-08
feat(tpu): generate library

### LINKED ISSUE #10759: Generate library for tpu/v2 API
https://github.com/googleapis/googleapis/tree/master/google/cloud/tpu/v2

It is unclear what documentation links to use, all the guides are about running models on the TPUs, not about how to configure them.  This is probably the closest:

https://cloud.google.com/tpu/docs/reference

devbww:
Generate library for tpu/v2 API.

Also move tpu/v1 into its own "product_path" subdirectory, with backwards-compatibility
headers specified via "forwarding_product_path".  This required a bunch of hand-edits which
we should look to automating in the future.

Fixes #10759.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10777)


coryan (approved):
Thanks

PR #10792, opened 2023-02-09
cleanup(GCS+gRPC): use `retention_duration` field

coryan:
The `retention_period` field is deprecated and will be retired "soon".

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10792)

### DIFF
--- a/google/cloud/storage/internal/grpc_bucket_metadata_parser.cc
+++ b/google/cloud/storage/internal/grpc_bucket_metadata_parser.cc
@@ -469,7 +469,8 @@ google::storage::v2::Bucket::RetentionPolicy ToProto(
   *result.mutable_effective_time() =
       google::cloud::internal::ToProtoTimestamp(rhs.effective_time);
   result.set_is_locked(rhs.is_locked);
-  result.set_retention_period(rhs.retention_period.count());
+  *result.mutable_retention_duration() =
+      google::cloud::internal::ToDurationProto(rhs.retention_period);
   return result;
 }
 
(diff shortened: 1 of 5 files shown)


google/cloud/storage/internal/grpc_bucket_metadata_parser.cc
```diff
@@ -479,7 +480,9 @@ storage::BucketRetentionPolicy FromProto(
   result.effective_time =
       google::cloud::internal::ToChronoTimePoint(rhs.effective_time());
   result.is_locked = rhs.is_locked();
-  result.retention_period = std::chrono::seconds(rhs.retention_period());
+  result.retention_period = std::chrono::duration_cast<std::chrono::seconds>(
+      std::chrono::seconds(rhs.retention_duration().seconds()) +
+      std::chrono::nanoseconds(rhs.retention_duration().nanos()));
```
devbww, line 485: Is this any different from just ...
```
  result.retention_period = std::chrono::seconds(rhs.retention_duration().seconds());
```
given that `rhs.retention_duration` must be positive.

I guess `rhs.retention_duration` could be a badly-encoded `google.protobuf.Duration`, with a >=1s `nanos` field.

In the end though, I seem to be thinking just to leave it as-is, despite all of that.  Perhaps we should have a "protobuf::Duration to chrono::duration" helper in `time_utils`.

devbww (approved):

google/cloud/storage/internal/grpc_bucket_metadata_parser.cc
coryan: More or less my train of thought too.  Frankly I think the use of a `google.protobuf.Duration` in the service seems like false precision.

Issue #10775, opened 2023-02-08
Create a CI build for 32-bits on Linux


coryan:
Motivated by #10772.  I think there are multiple little errors we will need to fix.  It may take a while to clean up these errors so I propose we do the following:

- [x] Introduce the CI build, but ignore build and test failures in the driver script (e.g., `cmake ... || true`).
- [x] Fix `google/cloud/storage/tests/object_read_large_integration_test.cc:43`
- [x] Fix `examples/grpc_credential_types.cc:106`
- [x] Fix `examples/grpc_credential_types.cc:135`
- [x] Fix `google/cloud/storage/benchmarks/aggregate_upload_throughput_benchmark.cc:323`
- [x] Fix `google/cloud/spanner/value_test.cc:189`
- [x] Fix `google/cloud/spanner/value_test.cc:763`
- [x] Fix the exceptions thrown in `pubsub_internal_subscription_message_queue_test`
- [x] Fix the spanner integration test freezes (or at least very long delays)
- [x] Fix the Pub/Sub integration tests (`pubsub_endurance`, `pubsub_samples`)
- [x] Change the script to fail when something breaks

coryan:
```
FAILED: google/cloud/storage/tests/CMakeFiles/storage_object_read_large_integration_test.dir/object_read_large_integration_test.cc.o 
/usr/bin/ccache /usr/bin/g++ -DGOOGLE_CLOUD_CPP_HAVE_GETRUSAGE=1 -DGOOGLE_CLOUD_CPP_HAVE_RUSAGE_THREAD=1 -DGTEST_LINKED_AS_SHARED_LIBRARY=1 -I/workspace -isystem /workspace/cmake-out -m32 -Wall -Wextra -Wconversion -Wno-sign-conversion -Werror -MD -MT google/cloud/storage/tests/CMakeFiles/storage_object_read_large_integration_test.dir/object_read_large_integration_test.cc.o -MF google/cloud/storage/tests/CMakeFiles/storage_object_read_large_integration_test.dir/object_read_large_integration_test.cc.o.d -o google/cloud/storage/tests/CMakeFiles/storage_object_read_large_integration_test.dir/object_read_large_integration_test.cc.o -c /workspace/google/cloud/storage/tests/object_read_large_integration_test.cc
/workspace/google/cloud/storage/tests/object_read_large_integration_test.cc: In function 'std::size_t google::cloud::storage::v2_8_0::{anonymous}::CurrentRss()':
[...]

PR #10793, opened 2023-02-09
fix(pubsub): fewer default threads for 32-bit builds

coryan:
When the code is compiled for 32-bit platforms using 4 * thread-count background threads can quickly exhaust the address space. Reduce the default number of background threads to just 4.

Part of the work for #10775

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10793)


devbww (approved):
s/thread-count/hardware-concurrency/ in the commit message.

PR #10798, opened 2023-02-10
cleanup(spanner): reorder class function declarations for clarity

### LINKED ISSUE #10755: spanner::Client documentation hides the non-deprecated functions
The deprecated functions are in groups, while this is probably a good idea, it effectively hides the non-deprecated functions in between those groups:

image

Maybe we should have a single group with all the deprecated functions, and another group with the non-deprecated ones?  Or maybe just not use groups?

devbww:
Move all the backwards-compatibility overloads in `spanner::Client` to the end of the public-functions section so that the recommended interfaces are more obvious, particularly in the doxygen-generated documentation.

Also remove some grouping around `ExecuteQuery()` and `ProfileQuery()` overloads.

Fixes #10755.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10798)

### DIFF
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -328,31 +328,6 @@ class Client {
   RowStream ExecuteQuery(QueryPartition const& partition, Options opts = {});
 
   /**
-   * Creates a set of partitions that can be used to execute a query
-   * operation in parallel.  Each of the returned partitions can be passed
-   * to `ExecuteQuery` to specify a subset of the query result to read.
-   *
-   * Partitions become invalid when the session used to create them is deleted,
-   * is idle for too long, begins a new transaction, or becomes too old. When
-   * any of these happen, it is not possible to resume the query, and the whole
-   * operation must be restarted from the beginning.
-   *
-   * @par Example
-   * @snippet samples.cc partition-query
-   *
-   * @param transaction The transaction to execute the operation in.
-   *     **Must** be a read-only snapshot transaction.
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/client.h
```diff
@@ -429,79 +326,32 @@ class Client {
...
+   *
+   * @return A `StatusOr` containing a vector of `QueryPartition`s or error
+   *     status on failure.
    */
-  RowStream ExecuteQuery(QueryPartition const& partition,
-                         std::initializer_list<internal::NonConstructible>) {
-    return ExecuteQuery(partition);
-  }
-  ///@}
+  StatusOr<std::vector<QueryPartition>> PartitionQuery(
```
dbolduc, line 351: (FYI, you might have already noticed that) the ordering used to be:
1. `ExecuteQuery`
2. `ProfileQuery`
3. `PartitionQuery`

and the .cc file [uses that ordering](https://github.com/googleapis/google-cloud-cpp/blob/3ae5c1ce53466090327348f849d197aff838e851/google/cloud/spanner/client.cc#L131-L173)

I think the diff will be cleaner if you restore that order.

devbww: > I think the diff will be cleaner if you restore that order.

Done.  I thought I might be improving things, but that's questionable, and it did certainly make it more confusing.  Thanks.

dbolduc (approved):

coryan:
Nice!

PR #10805, opened 2023-02-10
test(spanner): reset state in the DmlReturning test mutators

devbww:
A `Commit()` mutator might be re-run, so it should ensure its own idempotency.  Fix the `*DataTypeIntegrationTest.DmlReturning` tests accordingly.

Also raise visibility on the issue by adding a note in the samples about non-transactional state changes in mutators.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10805)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -2876,6 +2876,9 @@ void DmlWriteThenRead(google::cloud::spanner::Client client) {
             "SELECT FirstName, LastName FROM Singers where SingerId = 11");
         using RowType = std::tuple<std::string, std::string>;
         auto rows = client.ExecuteQuery(std::move(txn), std::move(select));
+        // Note: This mutator might be re-run, or its effects discarded, so
+        // changing non-transactional state (e.g., by producing output) is,
+        // in general, not something to be imitated.
         for (auto const& row : spanner::StreamOf<RowType>(rows)) {
           if (!row) return std::move(row).status();
           std::cout << "FirstName: " << std::get<0>(*row) << "\t";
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -2876,6 +2876,9 @@ void DmlWriteThenRead(google::cloud::spanner::Client client) {
             "SELECT FirstName, LastName FROM Singers where SingerId = 11");
         using RowType = std::tuple<std::string, std::string>;
         auto rows = client.ExecuteQuery(std::move(txn), std::move(select));
+        // Note: This mutator might be re-run, or its effects discarded, so
```
coryan, line 2879: I think we should try to show what to do instead of explaining what *not* to do.  I realize that might be difficult here.   Maybe something like:

```cc
int transaction_attempts = 0;
auto commit_result = client.Commit([&client, &transaction_attempts](...) {
 
  auto rows = client.ExecuteQuery(...);
   if (++transaction_attempts > 1) {
     std::cout << "*** Transaction mutator restarted, ignore any rows printed in earlier runs\n";
   }
   for (auto const& row : spanner::StreamOf<RowType>(rows)) { ...
}
```

devbww: > I think we should try to show what to do instead of explaining what not to do. I realize that might be difficult here. Maybe something like:

Maybe, yes.  A niggling issue is that the samples have all been spec'd out so that they look (mostly) the same in all languages.  I raised the question on the doc that specified the DML-RETURNING samples, so its probably best to wait and see what they have to say.  WDYT?

coryan: SGTM, we can always change them again.

coryan (approved):

PR #10807, opened 2023-02-10
cleanup(otel): StreamRange span inactive unless reading

dbolduc:
As discussed in our team meeting on 2023-02-01, a traced stream range parent span should only be active when making reads. It should not be active when the caller iterates over the range.

We can simplify the wrapper now that the base `StreamRange<>` does the work to re-instantiate the call span. (#10730)

<hr />

Note that the removed `SpanActiveDuringSubOperations` test is still covered. It is just not the responsibility of `MakeTracedStreamRange`.

We ensure that the underlying `StreamRange` captures the active span in this test:

https://github.com/googleapis/google-cloud-cpp/blob/13602cf4f4830c158181e37fd93b6b5fc144b4c9/google/cloud/stream_range_test.cc#L271

and we verify that the tracing connection starts an active span in these golden tests:

https://github.com/googleapis/google-cloud-cpp/blob/13602cf4f4830c158181e37fd93b6b5fc144b4c9/generator/integration_tests/tests/golden_kitchen_sink_tracing_connection_test.cc#L150

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10807)

### DIFF
--- a/generator/internal/tracing_connection_generator.cc
+++ b/generator/internal/tracing_connection_generator.cc
@@ -264,10 +264,10 @@ $tracing_connection_class_name$::$method_name$($request_type$ const& request) {
 StreamRange<$range_output_type$>
 $tracing_connection_class_name$::$method_name$($request_type$ request) {
   auto span = internal::MakeSpan("$product_namespace$::$connection_class_name$::$method_name$");
-  auto scope = absl::make_unique<opentelemetry::trace::Scope>(span);
+  auto scope = opentelemetry::trace::Scope(span);
   auto sr = child_->$method_name$(std::move(request));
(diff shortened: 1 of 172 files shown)


generator/internal/tracing_connection_generator.cc
```diff
@@ -267,7 +267,7 @@ StreamRange<$range_output_type$>
   auto scope = absl::make_unique<opentelemetry::trace::Scope>(span);
```
dbolduc, line 267: oops, this needn't be a `unique_ptr`

coryan (approved):

PR #10811, opened 2023-02-11
fix(common): fewer spurious warnings in the log

coryan:
When a streaming RPC is cancelled we should expect the final status of the RPC to be either kOk, or kCancelled. Both are normal outcomes in this case and do not deserve a warning in the log.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10811)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10811)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #10821, opened 2023-02-14
impl(GCS+gRPC): helpers to split upload data

coryan:
When uploading data we need to send `Write()` messages no larger than 2MiB (there is a constant, but that is the value).  Our input data comes in 2 forms, a single `absl::string_view` or a vector of `absl::Span<>` objects.  We also need to place the data into a `std::string` or `absl::Cord` depending on how the protobuf `bytes` field is represented in C++ (this changes depending on the environment).  These new helpers handle all these cases efficiently. I am planning to use them in the implementation of `GrpcClient`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10821)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10821)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #10828, opened 2023-02-14
impl(GCS+gRPC): better support for `absl::Cord`

coryan:
In some environments the storage protos use `absl::Cord`.  We believe we can minimize data copying in gRPC if the protos are initialized with an `absl::Cord` in these environments.  Even if the performance benefits do not materialize, this change makes the code easier to follow.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10828)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10828)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #10837, opened 2023-02-15
impl(bigquery): Interface headers with request/response and stubbed out implementation and tests for GetJob api.

jsrinnn:
Most of the implementation and tests are stubbed out.

I was trying to be mindful of the PR size as requested and also the request for the implmentation and testing files to be submitted in a single PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/job.h
+++ b/google/cloud/bigquery/v2/minimal/internal/job.h
@@ -12,36 +12,50 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "google/cloud/bigquery/v2/minimal/internal/job_rest_stub.h"
-#include "google/cloud/log.h"
-#include "google/cloud/status_or.h"
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_H
+
+#include "google/cloud/version.h"
 
 namespace google {
 namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-BigQueryJobStub::~BigQueryJobStub() = default;
+// Custom class for V2 Job proto fields.
 
-StatusOr<GetJobResponse> DefaultBigQueryJobStub::GetJob(
-    GetJobRequest const& request) {
-  GetJobResponse response;
-  if (request.project_id().empty()) {
-    GCP_LOG(DEBUG) << "Invalid request: " << request;
-    return Status(StatusCode::kInvalidArgument,
-                  "Invalid GetJobRequest: Project Id is empty");
-  }
-  if (request.job_id().empty()) {
-    GCP_LOG(DEBUG) << "Invalid request: " << request;
-    return Status(StatusCode::kInvalidArgument,
-                  "Invalid GetJobRequest: Job Id is empty");
-  }
-  // Not Implemented Yet: Call the rest client to get job details from the
-  // server.
-  return response;
-}
+struct JobConfiguration {
+  // Not Implemented Yet.
+};
+
+struct JobReference {
+  // Not Implemented Yet.
+};
+
+struct JobStatistics {
+  // Not Implemented Yet.
+};
+
+struct JobStatus {
+  // Not Implemented Yet.
+};
+
+struct Job {
+  std::string kind;
+  std::string etag;
+  std::string id;
+  std::string self_link;
+  std::string user_email;
+
+  JobStatus status;
+  JobConfiguration configuration;
+  JobStatistics stats;
+  JobReference reference;
+};
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace bigquery_v2_minimal_internal
 }  // namespace cloud
 }  // namespace google
+
+#endif  // GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_H
--- a/google/cloud/bigquery/v2/minimal/internal/job.h
+++ b/google/cloud/bigquery/v2/minimal/internal/job.h
@@ -16,6 +16,7 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_H
 
 #include "google/cloud/version.h"
+#include <string>
 
 namespace google {
 namespace cloud {
--- a/google/cloud/bigquery/v2/minimal/internal/job.h
+++ b/google/cloud/bigquery/v2/minimal/internal/job.h
@@ -22,8 +22,6 @@ namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-// Custom class for V2 Job proto fields.
-
 struct JobConfiguration {
   // Not Implemented Yet.
 };
@@ -40,6 +38,7 @@ struct JobStatus {
   // Not Implemented Yet.
 };
 
+// Custom object for V2 Job proto fields.
 struct Job {
   std::string kind;
   std::string etag;
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -12,33 +12,45 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "google/cloud/bigquery/v2/minimal/internal/job_rest_stub.h"
+#include "google/cloud/bigquery/v2/minimal/internal/job_request.h"
+#include "google/cloud/internal/absl_str_cat_quiet.h"
 #include "google/cloud/log.h"
-#include "google/cloud/status_or.h"
+#include "google/cloud/status.h"
 
 namespace google {
 namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-BigQueryJobStub::~BigQueryJobStub() = default;
+std::ostream& operator<<(std::ostream& os, GetJobRequest const& r) {
+  os << "GetJobRequest={project_id=" << r.project_id()
+     << ", job_id=" << r.job_id() << ", location=" << r.location();
+  return os << "}";
+}
 
-StatusOr<GetJobResponse> DefaultBigQueryJobStub::GetJob(
-    GetJobRequest const& request) {
-  GetJobResponse response;
-  if (request.project_id().empty()) {
-    GCP_LOG(DEBUG) << "Invalid request: " << request;
+StatusOr<rest_internal::RestRequest> GetJobRequest::BuildRestRequest(
+    GetJobRequest& r) {
+  rest_internal::RestRequest request;
+  if (r.project_id().empty()) {
+    GCP_LOG(DEBUG) << "Invalid request: " << r;
     return Status(StatusCode::kInvalidArgument,
                   "Invalid GetJobRequest: Project Id is empty");
   }
-  if (request.job_id().empty()) {
-    GCP_LOG(DEBUG) << "Invalid request: " << request;
+  if (r.job_id().empty()) {
+    GCP_LOG(DEBUG) << "Invalid request: " << r;
     return Status(StatusCode::kInvalidArgument,
                   "Invalid GetJobRequest: Job Id is empty");
   }
-  // Not Implemented Yet: Call the rest client to get job details from the
-  // server.
-  return response;
+  // Set path.
+  std::string path =
+      absl::StrCat("https://bigquery.googleapis.com/bigquery/v2/projects/",
+                   r.project_id(), "/jobs/", r.job_id());
+  request.SetPath(path);
+  // Add query params.
+  if (!r.location().empty()) {
+    request.AddQueryParameter("location", r.location());
+  }
+  return request;
 }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -18,15 +18,14 @@
 #include "google/cloud/internal/make_status.h"
 #include "google/cloud/status.h"
 #include "absl/strings/match.h"
-#include <ostream>
 
 namespace google {
 namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest& r,
-                                                      Options opts) {
+StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest const& r,
+                                                      Options& opts) {
   rest_internal::RestRequest request;
   if (r.project_id().empty()) {
     return internal::InvalidArgumentError(
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -13,39 +13,50 @@
 // limitations under the License.
 
 #include "google/cloud/bigquery/v2/minimal/internal/job_request.h"
+#include "google/cloud/common_options.h"
 #include "google/cloud/internal/absl_str_cat_quiet.h"
-#include "google/cloud/log.h"
+#include "google/cloud/internal/make_status.h"
 #include "google/cloud/status.h"
+#include "absl/strings/match.h"
+#include <ostream>
 
 namespace google {
 namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-std::ostream& operator<<(std::ostream& os, GetJobRequest const& r) {
-  os << "GetJobRequest={project_id=" << r.project_id()
-     << ", job_id=" << r.job_id() << ", location=" << r.location();
-  return os << "}";
-}
-
 StatusOr<rest_internal::RestRequest> GetJobRequest::BuildRestRequest(
-    GetJobRequest& r) {
+    Options opts, GetJobRequest& r) {
   rest_internal::RestRequest request;
   if (r.project_id().empty()) {
-    GCP_LOG(DEBUG) << "Invalid request: " << r;
-    return Status(StatusCode::kInvalidArgument,
-                  "Invalid GetJobRequest: Project Id is empty");
+    return internal::InvalidArgumentError(
+        "Invalid GetJobRequest: Project Id is empty",
+        GCP_ERROR_INFO().Build(StatusCode::kInvalidArgument));
   }
   if (r.job_id().empty()) {
-    GCP_LOG(DEBUG) << "Invalid request: " << r;
-    return Status(StatusCode::kInvalidArgument,
-                  "Invalid GetJobRequest: Job Id is empty");
+    return internal::InvalidArgumentError(
+        "Invalid GetJobRequest: Job Id is empty",
+        GCP_ERROR_INFO().Build(StatusCode::kInvalidArgument));
   }
-  // Set path.
+  if (!opts.has<EndpointOption>()) {
+    opts.set<EndpointOption>("https://bigquery.googleapis.com/");
+  }
+  // Fix the endpoints prefix and suffixes.
+  auto endpoint = opts.get<EndpointOption>();
+  if (!(absl::StartsWith(endpoint, "https://")) &&
+      !(absl::StartsWith(endpoint, "http://"))) {
+    endpoint = absl::StrCat("https://", endpoint);
+  }
+  if (absl::EndsWith(endpoint, "/")) {
+    absl::StrAppend(&endpoint, "bigquery/v2/projects/");
+  } else {
+    absl::StrAppend(&endpoint, "/bigquery/v2/projects/");
+  }
+
   std::string path =
-      absl::StrCat("https://bigquery.googleapis.com/bigquery/v2/projects/",
-                   r.project_id(), "/jobs/", r.job_id());
-  request.SetPath(path);
+      absl::StrCat(endpoint, r.project_id(), "/jobs/", r.job_id());
+  request.SetPath(std::move(path));
+
   // Add query params.
   if (!r.location().empty()) {
     request.AddQueryParameter("location", r.location());
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -25,7 +25,7 @@ namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest const& r,
-                                                      Options& opts) {
+                                                      Options const& opts) {
   rest_internal::RestRequest request;
   if (r.project_id().empty()) {
     return internal::InvalidArgumentError(
@@ -36,8 +36,10 @@ StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest const& r,
         "Invalid GetJobRequest: Job Id is empty", GCP_ERROR_INFO());
   }
   // Fix the endpoints prefix and suffixes.
-  auto endpoint =
-      opts.lookup<EndpointOption>("https://bigquery.googleapis.com/");
+  std::string endpoint = "https://bigquery.googleapis.com/";
+  if (opts.has<EndpointOption>()) {
+    endpoint = opts.get<EndpointOption>();
+  }
   if (!absl::StartsWith(endpoint, "https://") &&
       !absl::StartsWith(endpoint, "http://")) {
     endpoint = absl::StrCat("https://", endpoint);
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -25,33 +25,26 @@ namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-StatusOr<rest_internal::RestRequest> GetJobRequest::BuildRestRequest(
-    Options opts, GetJobRequest& r) {
+StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest& r,
+                                                      Options opts) {
   rest_internal::RestRequest request;
   if (r.project_id().empty()) {
     return internal::InvalidArgumentError(
-        "Invalid GetJobRequest: Project Id is empty",
-        GCP_ERROR_INFO().Build(StatusCode::kInvalidArgument));
+        "Invalid GetJobRequest: Project Id is empty", GCP_ERROR_INFO());
   }
   if (r.job_id().empty()) {
     return internal::InvalidArgumentError(
-        "Invalid GetJobRequest: Job Id is empty",
-        GCP_ERROR_INFO().Build(StatusCode::kInvalidArgument));
-  }
-  if (!opts.has<EndpointOption>()) {
-    opts.set<EndpointOption>("https://bigquery.googleapis.com/");
+        "Invalid GetJobRequest: Job Id is empty", GCP_ERROR_INFO());
   }
   // Fix the endpoints prefix and suffixes.
-  auto endpoint = opts.get<EndpointOption>();
-  if (!(absl::StartsWith(endpoint, "https://")) &&
-      !(absl::StartsWith(endpoint, "http://"))) {
+  auto endpoint =
+      opts.lookup<EndpointOption>("https://bigquery.googleapis.com/");
+  if (!absl::StartsWith(endpoint, "https://") &&
+      !absl::StartsWith(endpoint, "http://")) {
     endpoint = absl::StrCat("https://", endpoint);
   }
-  if (absl::EndsWith(endpoint, "/")) {
-    absl::StrAppend(&endpoint, "bigquery/v2/projects/");
-  } else {
-    absl::StrAppend(&endpoint, "/bigquery/v2/projects/");
-  }
+  if (!absl::EndsWith(endpoint, "/")) absl::StrAppend(&endpoint, "/");
+  absl::StrAppend(&endpoint, "bigquery/v2/projects/");
 
   std::string path =
       absl::StrCat(endpoint, r.project_id(), "/jobs/", r.job_id());
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.h
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.h
@@ -62,15 +62,15 @@ class GetJobRequest {
     return std::move(set_location(std::move(location)));
   }
 
-  // Builds RestRequest from GetJobRequest.
-  static StatusOr<rest_internal::RestRequest> BuildRestRequest(
-      Options opts, GetJobRequest& r);
-
  private:
   std::string project_id_;
   std::string job_id_;
   std::string location_;
 };
+
+// Builds RestRequest from GetJobRequest.
+StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest& r,
+                                                      Options opts);
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace bigquery_v2_minimal_internal
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.h
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.h
@@ -69,7 +69,7 @@ class GetJobRequest {
 
 // Builds RestRequest from GetJobRequest.
 StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest const& r,
-                                                      Options& opts);
+                                                      Options const& opts);
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace bigquery_v2_minimal_internal
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.h
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.h
@@ -16,29 +16,28 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_REQUEST_H
 
 #include "google/cloud/internal/rest_request.h"
+#include "google/cloud/options.h"
 #include "google/cloud/status_or.h"
 #include "google/cloud/version.h"
-#include <ostream>
+#include <iosfwd>
+#include <string>
 
 namespace google {
 namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-// Holds request parameters necessary to make the GetJob rest call.
+// Holds request parameters necessary to make the GetJob call.
 class GetJobRequest {
  public:
-  // Constructors
   GetJobRequest() = default;
   explicit GetJobRequest(std::string project_id, std::string job_id)
       : project_id_(std::move(project_id)), job_id_(std::move(job_id)) {}
 
-  // Getters
   std::string const& project_id() const { return project_id_; }
   std::string const& job_id() const { return job_id_; }
   std::string const& location() const { return location_; }
 
-  // Setters
   GetJobRequest& set_project_id(std::string project_id) & {
     project_id_ = std::move(project_id);
     return *this;
@@ -65,17 +64,13 @@ class GetJobRequest {
 
   // Builds RestRequest from GetJobRequest.
   static StatusOr<rest_internal::RestRequest> BuildRestRequest(
-      GetJobRequest& r);
+      Options opts, GetJobRequest& r);
 
-  // Members
  private:
   std::string project_id_;
   std::string job_id_;
   std::string location_;
 };
-
-// Logs GetJobRequest data.
-std::ostream& operator<<(std::ostream& os, GetJobRequest const& r);
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace bigquery_v2_minimal_internal
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.h
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.h
@@ -19,7 +19,6 @@
 #include "google/cloud/options.h"
 #include "google/cloud/status_or.h"
 #include "google/cloud/version.h"
-#include <iosfwd>
 #include <string>
 
 namespace google {
@@ -69,8 +68,8 @@ class GetJobRequest {
 };
 
 // Builds RestRequest from GetJobRequest.
-StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest& r,
-                                                      Options opts);
+StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest const& r,
+                                                      Options& opts);
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace bigquery_v2_minimal_internal
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.h
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.h
@@ -1,0 +1,85 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_REQUEST_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_REQUEST_H
+
+#include "google/cloud/internal/rest_request.h"
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include <ostream>
+
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// Holds request parameters necessary to make the GetJob rest call.
+class GetJobRequest {
+ public:
+  // Constructors
+  GetJobRequest() = default;
+  explicit GetJobRequest(std::string project_id, std::string job_id)
+      : project_id_(std::move(project_id)), job_id_(std::move(job_id)) {}
+
+  // Getters
+  std::string const& project_id() const { return project_id_; }
+  std::string const& job_id() const { return job_id_; }
+  std::string const& location() const { return location_; }
+
+  // Setters
+  GetJobRequest& set_project_id(std::string project_id) & {
+    project_id_ = std::move(project_id);
+    return *this;
+  }
+  GetJobRequest&& set_project_id(std::string project_id) && {
+    return std::move(set_project_id(std::move(project_id)));
+  }
+
+  GetJobRequest& set_job_id(std::string job_id) & {
+    job_id_ = std::move(job_id);
+    return *this;
+  }
+  GetJobRequest&& set_job_id(std::string job_id) && {
+    return std::move(set_job_id(std::move(job_id)));
+  }
+
+  GetJobRequest& set_location(std::string location) & {
+    location_ = std::move(location);
+    return *this;
+  }
+  GetJobRequest&& set_location(std::string location) && {
+    return std::move(set_location(std::move(location)));
+  }
+
+  // Builds RestRequest from GetJobRequest.
+  static StatusOr<rest_internal::RestRequest> BuildRestRequest(
+      GetJobRequest& r);
+
+  // Members
+ private:
+  std::string project_id_;
+  std::string job_id_;
+  std::string location_;
+};
+
+// Logs GetJobRequest data.
+std::ostream& operator<<(std::ostream& os, GetJobRequest const& r);
+
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
+}  // namespace bigquery_v2_minimal_internal
+}  // namespace cloud
+}  // namespace google
+
+#endif  // GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_REQUEST_H
--- a/google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 #include "google/cloud/bigquery/v2/minimal/internal/job_request.h"
+#include "google/cloud/common_options.h"
+#include "google/cloud/testing_util/status_matchers.h"
 #include <gmock/gmock.h>
 
 namespace google {
@@ -20,20 +22,77 @@ namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-TEST(GetJobRequestTest, BuildRestRequestSuccessWithLocation) {
-  // Not Implemented Yet.
+using ::google::cloud::testing_util::StatusIs;
+using ::testing::HasSubstr;
+
+TEST(JobRequestTest, SuccessWithLocation) {
+  GetJobRequest request("1", "2");
+  request.set_location("useast");
+  Options opts;
+  auto actual = GetJobRequest::BuildRestRequest(opts, request);
+  ASSERT_STATUS_OK(actual);
+
+  rest_internal::RestRequest expected;
+  expected.SetPath(
+      "https://bigquery.googleapis.com/bigquery/v2/projects/1/jobs/2");
+  expected.AddQueryParameter("location", "useast");
+  EXPECT_EQ(expected, actual.value());
 }
 
-TEST(GetJobRequestTest, BuildRestRequestSuccessWithoutLocation) {
-  // Not Implemented Yet.
+TEST(JobRequestTest, SuccessWithoutLocation) {
+  GetJobRequest request("1", "2");
+  Options opts;
+  auto actual = GetJobRequest::BuildRestRequest(opts, request);
+  ASSERT_STATUS_OK(actual);
+
+  rest_internal::RestRequest expected;
+  expected.SetPath(
+      "https://bigquery.googleapis.com/bigquery/v2/projects/1/jobs/2");
+  EXPECT_EQ(expected, actual.value());
 }
 
-TEST(GetJobRequestTest, BuildRestRequestFailureEmptyProjectId) {
-  // Not Implemented Yet.
+TEST(JobRequestTest, SuccessWithEndpoint) {
+  GetJobRequest request("1", "2");
+  Options opts;
+
+  struct EndpointTest {
+    std::string endpoint;
+    std::string expected;
+  } cases[] = {
+      {"https://myendpoint.google.com",
+       "https://myendpoint.google.com/bigquery/v2/projects/1/jobs/2"},
+      {"http://myendpoint.google.com",
+       "http://myendpoint.google.com/bigquery/v2/projects/1/jobs/2"},
+      {"myendpoint.google.com",
+       "https://myendpoint.google.com/bigquery/v2/projects/1/jobs/2"},
+      {"https://myendpoint.google.com/",
+       "https://myendpoint.google.com/bigquery/v2/projects/1/jobs/2"},
+  };
+
+  for (auto const& test : cases) {
+    SCOPED_TRACE("Testing for endpoint: " + test.endpoint +
+                 ", expected: " + test.expected);
+    opts.set<EndpointOption>(test.endpoint);
+    auto actual = GetJobRequest::BuildRestRequest(opts, request);
+    ASSERT_STATUS_OK(actual);
+    EXPECT_EQ(test.expected, actual.value().path());
+  }
 }
 
-TEST(GetJobRequestTest, BuildRestRequestFailureEmptyJobId) {
-  // Not Implemented Yet.
+TEST(JobRequestTest, EmptyProjectId) {
+  GetJobRequest request("", "job_id");
+  Options opts;
+  auto rest_request = GetJobRequest::BuildRestRequest(opts, request);
+  EXPECT_THAT(rest_request, StatusIs(StatusCode::kInvalidArgument,
+                                     HasSubstr("Project Id is empty")));
+}
+
+TEST(GetJobRequest, EmptyJobId) {
+  GetJobRequest request("project_id", "");
+  Options opts;
+  auto rest_request = GetJobRequest::BuildRestRequest(opts, request);
+  EXPECT_THAT(rest_request, StatusIs(StatusCode::kInvalidArgument,
+                                     HasSubstr("Job Id is empty")));
 }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
--- a/google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
@@ -29,26 +29,26 @@ TEST(JobRequestTest, SuccessWithLocation) {
   GetJobRequest request("1", "2");
   request.set_location("useast");
   Options opts;
-  auto actual = GetJobRequest::BuildRestRequest(opts, request);
+  auto actual = BuildRestRequest(request, opts);
   ASSERT_STATUS_OK(actual);
 
   rest_internal::RestRequest expected;
   expected.SetPath(
       "https://bigquery.googleapis.com/bigquery/v2/projects/1/jobs/2");
   expected.AddQueryParameter("location", "useast");
-  EXPECT_EQ(expected, actual.value());
+  EXPECT_EQ(expected, *actual);
 }
 
 TEST(JobRequestTest, SuccessWithoutLocation) {
   GetJobRequest request("1", "2");
   Options opts;
-  auto actual = GetJobRequest::BuildRestRequest(opts, request);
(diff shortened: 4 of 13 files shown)


google/cloud/bigquery/v2/minimal/internal/job.h
```diff
@@ -0,0 +1,61 @@
...
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_H
+
+#include "google/cloud/version.h"
+
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// Custom class for V2 Job proto fields.
```
coryan, line 25: What is this comment referring to?  Can you put it next to the class?

google/cloud/bigquery/v2/minimal/internal/job_request.h
```diff
@@ -0,0 +1,85 @@
...
+
+// Logs GetJobRequest data.
+std::ostream& operator<<(std::ostream& os, GetJobRequest const& r);
+
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
+}  // namespace bigquery_v2_minimal_internal
+}  // namespace cloud
+}  // namespace google
+
+#endif  // GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_REQUEST_H
```
coryan, line 85: nit: missing newline.

google/cloud/bigquery/v2/minimal/internal/job_request.h
```diff
@@ -0,0 +1,85 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_REQUEST_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_REQUEST_H
+
+#include "google/cloud/internal/rest_request.h"
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include <ostream>
```
coryan, line 21: s/<ostream>/<iosfwd>/ and then add the right `#include` in the `.cc` file.

google/cloud/bigquery/v2/minimal/internal/job_request.h
```diff
@@ -0,0 +1,85 @@
...
+  }
+  GetJobRequest&& set_location(std::string location) && {
+    return std::move(set_location(std::move(location)));
+  }
+
+  // Builds RestRequest from GetJobRequest.
+  static StatusOr<rest_internal::RestRequest> BuildRestRequest(
+      GetJobRequest& r);
+
+  // Members
```
coryan, line 70: We generally omit comments like "setters" "getters" and "members". Unless we intend the comments to be part of the public reference documentation, in which case they get far more elaborate. Please remove this and the others.

google/cloud/bigquery/v2/minimal/internal/job_request.h
```diff
@@ -0,0 +1,85 @@
...
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include <ostream>
+
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// Holds request parameters necessary to make the GetJob rest call.
```
coryan, line 28: s/rest/REST/ though maybe I would omit that completely.

google/cloud/bigquery/v2/minimal/internal/job_request.h
coryan, line 21: add `#include <string>`

google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+#include "google/cloud/bigquery/v2/minimal/internal/job_request.h"
+#include <gmock/gmock.h>
+
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
```
coryan, line 22: I think you have enough code to test `ostream& operator<<(...)`?

google/cloud/bigquery/v2/minimal/internal/job_response.cc
```diff
@@ -0,0 +1,35 @@
...
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+StatusOr<GetJobResponse> GetJobResponse::BuildFromHttpResponse(
+    BigQueryHttpResponse const& http_response) {
+  GetJobResponse response;
+  if (http_response.payload.empty()) {
+    return Status(StatusCode::kUnknown, "Empty payload in HTTP response.");
```
coryan, line 26: please use one of the helper functions in `google/cloud/internal/make_status.h`.  I am not sure `kUnknown` is the right thing.  Probably `kInternal` or maybe `kInvalidArgument`?

google/cloud/bigquery/v2/minimal/internal/job_response.h
```diff
@@ -0,0 +1,44 @@
...
+// limitations under the License.
+
+// Internal interface for Bigquery V2 Job resource.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_RESPONSE_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_RESPONSE_H
+
+#include "google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.h"
+#include "google/cloud/bigquery/v2/minimal/internal/job.h"
+#include "google/cloud/version.h"
```
coryan, line 22: missing `#include "google/cloud/status_or.h"

google/cloud/bigquery/v2/minimal/internal/job_response_test.cc
```diff
@@ -0,0 +1,34 @@
...
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+TEST(GetJobResponseTest, BuildFromHttpResponseSuccess) {
+  // Not Implemented Yet.
+}
+
+TEST(GetJobResponseTest, BuildFromHttpResponseFailure) {
+  // Not Implemented Yet.
```
coryan, line 28: I think you can add a test for the empty payload case?

google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc
```diff
@@ -12,15 +12,35 @@
...
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+BigQueryJobStub::~BigQueryJobStub() = default;
+
+StatusOr<GetJobResponse> DefaultBigQueryJobStub::GetJob(
+    GetJobRequest const& request) {
+  GetJobResponse response;
+  if (request.project_id().empty()) {
+    GCP_LOG(DEBUG) << "Invalid request: " << request;
```
coryan, line 30: There is no need to log these errors, the logging decorator will do that.

google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc
```diff
@@ -12,15 +12,35 @@
...
+    return Status(StatusCode::kInvalidArgument,
```
coryan, line 31: Please use the (relatively new) factory functions in `google/cloud/internal/make_status.h`. There is also a companion macro:

```cc
  return InvalidArgumentError("Invalid GetJobRequest: project id is empty", GCP_ERROR_INFO());
```

google/cloud/bigquery/v2/minimal/internal/job_rest_stub.h
```diff
@@ -12,18 +12,42 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Internal interface for Bigquery V2 Job resource.
+// Internal Stub interface for Bigquery V2 Job apis.
```
coryan, line 15: Consider removing these file comments.

google/cloud/bigquery/v2/minimal/internal/job_response.h
```diff
@@ -0,0 +1,44 @@
...
+  Job job;
+  BigQueryHttpResponse http_response;
+};
+
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
+}  // namespace bigquery_v2_minimal_internal
+}  // namespace cloud
+}  // namespace google
+
+#endif  // GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_RESPONSE_H
```
coryan, line 44: nit: missing newline.  Take some time to search for any others.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -0,0 +1,59 @@
...
+  os << "GetJobRequest={project_id=" << r.project_id()
+     << ", job_id=" << r.job_id() << ", location=" << r.location();
+  return os << "}";
+}
+
+StatusOr<rest_internal::RestRequest> GetJobRequest::BuildRestRequest(
+    GetJobRequest& r) {
+  rest_internal::RestRequest request;
+  if (r.project_id().empty()) {
+    GCP_LOG(DEBUG) << "Invalid request: " << r;
```
coryan, line 35: See below, please just let the logging decorator log stuff.  And use the factory functions in `google/cloud/make_status.h`.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -0,0 +1,59 @@
...
+  if (r.job_id().empty()) {
+    GCP_LOG(DEBUG) << "Invalid request: " << r;
+    return Status(StatusCode::kInvalidArgument,
+                  "Invalid GetJobRequest: Job Id is empty");
+  }
+  // Set path.
+  std::string path =
+      absl::StrCat("https://bigquery.googleapis.com/bigquery/v2/projects/",
+                   r.project_id(), "/jobs/", r.job_id());
+  request.SetPath(path);
```
coryan, line 48: I would have done `std::move(path)` since there are no plans to use `path` again.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -0,0 +1,59 @@
...
+    GCP_LOG(DEBUG) << "Invalid request: " << r;
+    return Status(StatusCode::kInvalidArgument,
+                  "Invalid GetJobRequest: Project Id is empty");
+  }
+  if (r.job_id().empty()) {
+    GCP_LOG(DEBUG) << "Invalid request: " << r;
+    return Status(StatusCode::kInvalidArgument,
+                  "Invalid GetJobRequest: Job Id is empty");
+  }
+  // Set path.
```
coryan, line 44: You can save yourself this comment, we know what is going on.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -0,0 +1,59 @@
...
+  std::string path =
+      absl::StrCat("https://bigquery.googleapis.com/bigquery/v2/projects/",
```
coryan, line 46: Seems like this URL should come from some `EndpointOption`.  Otherwise customers won't be able to override the endpoint and therefore won't be able to use this with VPC-SC and the other mysterious network configurations that require endpoint overrides.

google/cloud/bigquery/v2/minimal/internal/job_response.cc
```diff
@@ -0,0 +1,35 @@
...
+  }
+  // Not Implemented Yet: Parse HttpResponse and build GetJobResponse object.
```
coryan, line 28: Clever way to avoid `TODO` and creating bugs 😀

coryan (changes requested):

google/cloud/bigquery/v2/minimal/internal/job.h
jsrinnn: Done.

Its referring to the custom Job definition. I move it above the struct that defined Job. 

(Still working on the code review feedback. You'll see the changes once I push them. Commenting now so its easier for me to see which comments have been addressed)

google/cloud/bigquery/v2/minimal/internal/job_request.h
jsrinnn: Done. 

Fixed here and elsewhere. I would have thought checkers-pr would catch it. Maybe not.

google/cloud/bigquery/v2/minimal/internal/job_request.h
jsrinnn: Done

google/cloud/bigquery/v2/minimal/internal/job_request.h
jsrinnn: Done

google/cloud/bigquery/v2/minimal/internal/job_request.h
jsrinnn: Done

jsrinnn:
Addressed code review feedback and added unit tests. PTAL.

google/cloud/bigquery/v2/minimal/internal/job.h
```diff
@@ -0,0 +1,60 @@
...
+struct JobStatistics {
+  // Not Implemented Yet.
+};
+
+struct JobStatus {
+  // Not Implemented Yet.
+};
+
+// Custom object for V2 Job proto fields.
+struct Job {
```
devbww, line 42: It is a little unexpected to have so many plain structs.

google/cloud/bigquery/v2/minimal/internal/job.h
```diff
@@ -0,0 +1,60 @@
...
+  std::string kind;
```
devbww, line 43: `#include <string>`

google/cloud/bigquery/v2/minimal/internal/job.h
```diff
@@ -0,0 +1,60 @@
...
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_H
+
+#include "google/cloud/version.h"
+
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
```
devbww, line 22: Given that the feature now seems to be branded as "bigquery_rest", should we be updating path names and namespaces?

google/cloud/bigquery/v2/minimal/internal/job_request.h
```diff
@@ -0,0 +1,80 @@
...
+#include <iosfwd>
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// Holds request parameters necessary to make the GetJob call.
+class GetJobRequest {
```
devbww, line 31: It is a little unexpected for a file that declares `class GetJobRequest` to just be called `job_request.h`.

google/cloud/bigquery/v2/minimal/internal/job_request.h
```diff
@@ -0,0 +1,80 @@
...
+    location_ = std::move(location);
+    return *this;
+  }
+  GetJobRequest&& set_location(std::string location) && {
+    return std::move(set_location(std::move(location)));
+  }
+
+  // Builds RestRequest from GetJobRequest.
+  static StatusOr<rest_internal::RestRequest> BuildRestRequest(
+      Options opts, GetJobRequest& r);
```
devbww, line 67: Why have a `GetJobRequest` static member function that takes a `GetJobRequest`, rather than a plain member function?

google/cloud/bigquery/v2/minimal/internal/job_request.h
```diff
@@ -0,0 +1,80 @@
...
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_REQUEST_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_JOB_REQUEST_H
+
+#include "google/cloud/internal/rest_request.h"
+#include "google/cloud/options.h"
+#include "google/cloud/status_or.h"
+#include "google/cloud/version.h"
+#include <iosfwd>
```
devbww, line 22: Unused?

google/cloud/bigquery/v2/minimal/internal/job_request.h
```diff
@@ -0,0 +1,80 @@
...
+  std::string const& job_id() const { return job_id_; }
+  std::string const& location() const { return location_; }
+
+  GetJobRequest& set_project_id(std::string project_id) & {
+    project_id_ = std::move(project_id);
+    return *this;
+  }
+  GetJobRequest&& set_project_id(std::string project_id) && {
+    return std::move(set_project_id(std::move(project_id)));
+  }
```
devbww, line 47: Do we really need qualified overloads for these?

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -0,0 +1,70 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/bigquery/v2/minimal/internal/job_request.h"
+#include "google/cloud/common_options.h"
+#include "google/cloud/internal/absl_str_cat_quiet.h"
+#include "google/cloud/internal/make_status.h"
+#include "google/cloud/status.h"
+#include "absl/strings/match.h"
+#include <ostream>
```
devbww, line 21: Unused?

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -0,0 +1,70 @@
...
+    return internal::InvalidArgumentError(
+        "Invalid GetJobRequest: Job Id is empty",
+        GCP_ERROR_INFO().Build(StatusCode::kInvalidArgument));
+  }
+  if (!opts.has<EndpointOption>()) {
+    opts.set<EndpointOption>("https://bigquery.googleapis.com/");
+  }
+  // Fix the endpoints prefix and suffixes.
+  auto endpoint = opts.get<EndpointOption>();
+  if (!(absl::StartsWith(endpoint, "https://")) &&
```
devbww, line 46: We usually eschew redundant parentheses.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -0,0 +1,70 @@
...
+  if (r.job_id().empty()) {
+    return internal::InvalidArgumentError(
+        "Invalid GetJobRequest: Job Id is empty",
+        GCP_ERROR_INFO().Build(StatusCode::kInvalidArgument));
+  }
+  if (!opts.has<EndpointOption>()) {
+    opts.set<EndpointOption>("https://bigquery.googleapis.com/");
+  }
+  // Fix the endpoints prefix and suffixes.
+  auto endpoint = opts.get<EndpointOption>();
```
devbww, line 45: If you really want to `opts.set<>()` locally, consider `opts.lookup<EndpointOption>("https://bigquery.googleapis.com/")`, but it is probably better to only use `opts.get<>()` here.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -0,0 +1,70 @@
...
+  if (!(absl::StartsWith(endpoint, "https://")) &&
+      !(absl::StartsWith(endpoint, "http://"))) {
+    endpoint = absl::StrCat("https://", endpoint);
+  }
+  if (absl::EndsWith(endpoint, "/")) {
+    absl::StrAppend(&endpoint, "bigquery/v2/projects/");
+  } else {
+    absl::StrAppend(&endpoint, "/bigquery/v2/projects/");
+  }
```
devbww, line 54: I typically recommend making the non-conditional parts non-conditional, so ...
```
  if (!absl::EndsWith(endpoint, "/")) absl::StrAppend(&endpoint, "/");
  absl::StrAppend(&endpoint, "bigquery/v2/projects/");
```

google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
```diff
@@ -0,0 +1,101 @@
...
+  request.set_location("useast");
+  Options opts;
+  auto actual = GetJobRequest::BuildRestRequest(opts, request);
+  ASSERT_STATUS_OK(actual);
+
+  rest_internal::RestRequest expected;
+  expected.SetPath(
+      "https://bigquery.googleapis.com/bigquery/v2/projects/1/jobs/2");
+  expected.AddQueryParameter("location", "useast");
+  EXPECT_EQ(expected, actual.value());
```
devbww, line 39: We prefer `operator*()` over `value()` (after asserting `ok()`).

google/cloud/bigquery/v2/minimal/internal/job_response.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// Parses the BigQueryHttpResponse and builds a GetJobResponse.
+struct GetJobResponse {
+  // Builds GetJobResponse from HttpResponse.
+  static StatusOr<GetJobResponse> BuildFromHttpResponse(
```
devbww, line 31: We discourage structs with behavior.

google/cloud/bigquery/v2/minimal/internal/job_rest_stub.h
```diff
@@ -12,18 +12,40 @@
...
+#include "google/cloud/status_or.h"
 #include "google/cloud/version.h"
+#include <memory>
 
 namespace google {
 namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+class BigQueryJobStub {
```
devbww, line 30: If "rest" isn't part of the object name, need it be part of the file name?

devbww (commented):
Random comments.  Feel free to ignore them.

coryan (changes requested):
> Reviewed 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)* status: 10 of 13 files reviewed, 18 unresolved discussions (waiting on @devbww and @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 85 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r1-85:-NORB-VsAroS2WxenC9A:b-mq6o13) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/947c56865ec73ea82994067beedd3b36c659f721/google/cloud/bigquery/v2/minimal/internal/job_request.h#L85)):*

Done. 

Fixed here and elsewhere. I would have thought checkers-pr would catch it. Maybe not.
</blockquote></details>

You are absolutely right about `checkers`.  @devbww fixed it.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 47 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-47:-NORDsZeDYfMALIwSbGr:bla8g3v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L47)):*

Do we really need qualified overloads for these?
</blockquote></details>

They are relevant if we use these as temporary objects:

```cc
  Foo(GetJobRequest().set_project_id(p).set_job_id(j));
```
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 67 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOR9FtW6hjkV-XhgVwo:-NOR9FtW6hjkV-XhgVwp:b1ma593) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L67)):*
> ```objc
>   // Builds RestRequest from GetJobRequest.
>   static StatusOr<rest_internal::RestRequest> BuildRestRequest(
>       Options opts, GetJobRequest& r);
> ```

Usually our `options` parameter is the last one.  We also do not change the endpoint on each request (tempting as that is).  We usually have a "client" that caches the connections. The request can just have a "path" and then we prepend the Client's

That seems like work for a future PR though.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 34 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NORBKWZ7fcFuflb50zM:-NORBKWZ7fcFuflb50zN:b-p69brg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L34)):*
> ```cpp
>     return internal::InvalidArgumentError(
>         "Invalid GetJobRequest: Project Id is empty",
>         GCP_ERROR_INFO().Build(StatusCode::kInvalidArgument));
> ```

You should not need to call `.Build`, did it break if you did?  The intent is to just capture some location info, as in:

```cc
  return internal::InvalidArgumentError("Invalid GetJobRequest: project id is empty", GCP_ERROR_INFO());
```
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 50 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NORC4Xj6sWqMiC-nV1Y:-NORC4Xj6sWqMiC-nV1Z:b-eifsf7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L50)):*
> ```cpp
>     endpoint = absl::StrCat("https://", endpoint);
>   }
>   if (absl::EndsWith(endpoint, "/")) {
> ```

I think `RestClient` already does some of this, check with @scotthart 
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 37 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NORChQV5UahvdCBcpS3:-NORChQV5UahvdCBcpS4:b-apu1al) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L37)):*
> ```cpp
>     return internal::InvalidArgumentError(
>         "Invalid GetJobRequest: Job Id is empty",
>         GCP_ERROR_INFO().Build(StatusCode::kInvalidArgument));
> ```

Please remove all these `.Build()` calls.

google/cloud/bigquery/v2/minimal/internal/job.h
dbolduc: I don't think so. In my mind `bigquery_rest` is all encompassing and could include some `google::cloud::bigquery_v2_foo_internal` at a later date.

My understanding of this is limited though. (Which implies I probably did not have enough foresight to justify changing the name of the feature. Oh well.)

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)* status: 10 of 13 files reviewed, 18 unresolved discussions (waiting on @coryan, @dbolduc, and @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 47 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-47:-NORGo5c4_zhFpKCW_JC:bxiiv44) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L47)):*

They are relevant if we use these as temporary objects:

```cc
  Foo(GetJobRequest().set_project_id(p).set_job_id(j));
```

</blockquote></details>

https://godbolt.org/z/oohr3revo

google/cloud/bigquery/v2/minimal/internal/job.h
devbww: OK.

coryan:
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 47 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-47:-NORLeanAjEql2dLDK4w:b-gmaqm4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L47)):*

https://godbolt.org/z/oohr3revo
</blockquote></details>

https://godbolt.org/z/hs4ze8YbW

G() gets to use the move constructor, H() needs to use the copy constructor.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)* status: 10 of 13 files reviewed, 17 unresolved discussions (waiting on @coryan, @dbolduc, and @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 47 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-47:-NORQ0HHA5rMQqtrRWsq:b-vxakw6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L47)):*

https://godbolt.org/z/hs4ze8YbW

G() gets to use the move constructor, H() needs to use the copy constructor.

</blockquote></details>

I just wonder whether that optimization will be worth the extra clutter in this case.

If so, then it would probably also be valuable for `Foo(..., Options().set<FooOpt>(foo))`.

jsrinnn (commented):
Thank you for reviewing! I think I have addressed all of them. Do let me know if I have missed out any.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-:-NORWydFEzyLpNC4SU0P:bxnibqx)* status: 1 of 13 files reviewed, 17 unresolved discussions (waiting on @coryan, @dbolduc, @devbww, and @scotthart)

___
*[`google/cloud/bigquery/v2/minimal/internal/job.h` line 42 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdN-r2-42:-NOREKT49nuK30RQl75l:b-4bei9v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job.h#L42)):*

It is a little unexpected to have so many plain structs.
</blockquote></details>

These are nested objects that I'll implement and are part of the final Job struct. Class here seemed like an overkill and Struct seemed more appropriate as I am considering these as dat objects within the response. I also saw in other libraries that response object were mainly structs.
___
*[`google/cloud/bigquery/v2/minimal/internal/job.h` line 43 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdN-r2-43:-NORLig-5myIrNWWC9Bb:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job.h#L43)):*

`#include <string>`
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 22 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-22:-NORF3fL9IOn8bAYoMI9:b62ns83) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L22)):*

Unused?
</blockquote></details>

This was added based on previous code review feedback by carlos. Instead of including <iostream> directly here, <iosfwd> is included with <iostream> in .cc file
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 31 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-31:-NORG9ABBbBCrWaJiDVc:b-6jaynd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L31)):*

It is a little unexpected for a file that declares `class GetJobRequest` to just be called `job_request.h`.
</blockquote></details>

The intent was for this file to hold other job api's requests definition as well and not have one separate file for each api.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 47 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-47:-NORGWoFCcHdXnm0gCvq:b-wmniz6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L47)):*

I just wonder whether that optimization will be worth the extra clutter in this case.

If so, then it would probably also be valuable for `Foo(..., Options().set<FooOpt>(foo))`.
</blockquote></details>

Yes. This was based on Carlos feedback for fluent setters, in the design doc. Please let me know if further changes are required here.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 67 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-67:-NORHAkwAgWwauBKSwGY:b-s9es7d) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L67)):*

Why have a `GetJobRequest` static member function that takes a `GetJobRequest`, rather than a plain member function?
</blockquote></details>

Ah...yeah, you're correct. This shouldn't be a member function. Thanks for catching that! Fixed it. PTAL.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 67 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOR9FtW6hjkV-XhgVwo:-NORHX7dEcPDHXw1YyK7:b77mn8f) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L67)):*

Usually our `options` parameter is the last one.  We also do not change the endpoint on each request (tempting as that is).  We usually have a "client" that caches the connections. The request can just have a "path" and then we prepend the Client's

That seems like work for a future PR though.

</blockquote></details>

Ack. I have modified the Options parameter to the be the last.  

Regaring the endpoint...

I wanted to make sure if none is supplied then we fallback on the default public endpoint. Client can override it by supplying a custom one. I am opting this route instead of failing if no endpoint is specified. Although it is debatable if none is specified if it is better to error or fallback to default.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 21 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-21:-NORJuR5Af12zlvMrb9A:b-4wytfn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L21)):*

Unused?
</blockquote></details>

Please see my previous comment. This is needed because of <iosfwd> include in the header.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 34 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NORBKWZ7fcFuflb50zM:-NOReWW_12Edc5QTeDPc:b-gav89o) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L34)):*

You should not need to call `.Build`, did it break if you did?  The intent is to just capture some location info, as in:

```cc
  return internal::InvalidArgumentError("Invalid GetJobRequest: project id is empty", GCP_ERROR_INFO());
```

</blockquote></details>

Done.

Didn't realize InvalidArgumentError implementation returns the status code. Thanks for catching that.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 45 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-45:-NORNwdn5CRm3-4zsjnG:bmwmt29) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L45)):*

If you really want to `opts.set<>()` locally, consider `opts.lookup<EndpointOption>("https://bigquery.googleapis.com/")`, but it is probably better to only use `opts.get<>()` here.
</blockquote></details>

Done. 

I like the opts.lookup() suggestion as I want to ensure we fallback to a default endpoint if none is supplied.  Please also see my previous response to Carlos on this topic.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 46 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-46:-NORPuN48lOrOJxJ4xIf:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L46)):*

We usually eschew redundant parentheses.
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 50 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NORC4Xj6sWqMiC-nV1Y:-NORQ2niECLsuvqhdufu:b-f1l3o3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L50)):*

I think `RestClient` already does some of this, check with @scotthart 
</blockquote></details>

Ack. I'll leave it in this PR and cleanup in subsequent PRs if this is already being done in RestClient.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 54 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-54:-NORQYb0D4JLO0mfRMaQ:b7vbxps) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L54)):*

I typically recommend making the non-conditional parts non-conditional, so ...
```
  if (!absl::EndsWith(endpoint, "/")) absl::StrAppend(&endpoint, "/");
  absl::StrAppend(&endpoint, "bigquery/v2/projects/");
```
</blockquote></details>

Done. 

That is much cleaner. Thank You.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request_test.cc` line 39 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdQ-r2-39:-NORRTxK6y-Ln2rmponZ:bnsjbg9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request_test.cc#L39)):*

We prefer `operator*()` over `value()` (after asserting `ok()`).
</blockquote></details>

Done.

Ah nice. Fixed it.  It didn't apply in the case `actual.value().path()` so I kept that the same as before.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response.h` line 31 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdS-r2-31:-NORTdb34TgwjQ6Ik6fC:bmc6mpe) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_response.h#L31)):*

We discourage structs with behavior.
</blockquote></details>

Done.

I noticed this pattern was used a lot in google/cloud/storage so I assumed structs were preferred for response.

I have modified this to a class . 

But  I think I'll need to use the struct for response objects that are used as rvalues e.g BigQueryHttpResponse otherwise that was giving me a compile error. (Maybe it needed an explicit std::move).
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.h` line 30 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdM-r2-30:-NORUaCZ2KbomFqQ-eNa:b-3qw97a) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.h#L30)):*

If "rest" isn't part of the object name, need it be part of the file name?
</blockquote></details>

Done.

Made it part of the Class i.e object name as well.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 37 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NORChQV5UahvdCBcpS3:-NOReOv74isL5WeBoMdv:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L37)):*

Please remove all these `.Build()` calls.
</blockquote></details>

Done.

jsrinnn:
Addressed remaining code review feedback.

devbww (commented):
> Reviewed 3 of 13 files at r1, 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)* status: all files reviewed, 12 unresolved discussions (waiting on @coryan, @jsrinnn, and @scotthart)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 22 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-22:-NOSXMGN6n1y8w1hasTT:b-j09z02) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L22)):*

This was added based on previous code review feedback by carlos. Instead of including <iostream> directly here, <iosfwd> is included with <iostream> in .cc file
</blockquote></details>

See my followup in the cc file.  It still seems unused.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 72 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOSXsmkAS-N1nm9T3TK:-NOSXsmkAS-N1nm9T3TL:b-os8zhr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8044d3d64b71c2fffb3095973e0a6a7dd17a1cc2/google/cloud/bigquery/v2/minimal/internal/job_request.h#L72)):*
> ```objc
> 
> // Builds RestRequest from GetJobRequest.
> StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest& r,
> ```

`BuildRestRequest()` does appear to need to mutate the `GetJobRequest`, so `s/&/ const&/`.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 21 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-21:-NOSX-dIEF2frnYfKqFC:bzie6at) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L21)):*

Please see my previous comment. This is needed because of <iosfwd> include in the header.
</blockquote></details>

I don't think I understand.  There doesn't appear to be anything here, or in the header, that uses these includes.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 45 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-45:-NOSYiC58bQvNycfsclE:b-qwtnbd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L45)):*

Done. 

I like the opts.lookup() suggestion as I want to ensure we fallback to a default endpoint if none is supplied.  Please also see my previous response to Carlos on this topic.
</blockquote></details>

I don't think that ensuring we fallback to a default endpoint implies that we need to mutate the local `opts`.  Indeed, this function should probably take an `Options const&` to avoid an unnecessary copy.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request_test.cc` line 39 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdQ-r2-39:-NOSZd2_3Ass9pp0lNqK:b-4bdsqu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request_test.cc#L39)):*

Done.

Ah nice. Fixed it.  It didn't apply in the case `actual.value().path()` so I kept that the same as before.
</blockquote></details>

`s/actual.value().path()/actual->path()/`
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request_test.cc` line 75 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOSZvuC4kPBDg_5FgQd:-NOSZvuC4kPBDg_5FgQe:b-ol3zsz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8044d3d64b71c2fffb3095973e0a6a7dd17a1cc2/google/cloud/bigquery/v2/minimal/internal/job_request_test.cc#L75)):*
> ```cpp
>     SCOPED_TRACE("Testing for endpoint: " + test.endpoint +
>                  ", expected: " + test.expected);
>     opts.set<EndpointOption>(test.endpoint);
> ```

`opts` should probably be local to the loop.

coryan (approved):
The things I were blockers for me are fixed.  There is some additional work, but that is too much to ask for a single PR. So :lgtm: 

Please work with the other reviewers to make sure you address their comments too.

> Reviewed 2 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-:-NOpId-eBG6TcQcb1Sxa:b-dr5ieb)* status: all files reviewed, 8 unresolved discussions (waiting on @devbww and @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 67 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOR9FtW6hjkV-XhgVwo:-NOpI8KS2sO81F1AwhKB:b-fw5j1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L67)):*

Ack. I have modified the Options parameter to the be the last.  

Regaring the endpoint...

I wanted to make sure if none is supplied then we fallback on the default public endpoint. Client can override it by supplying a custom one. I am opting this route instead of failing if no endpoint is specified. Although it is debatable if none is specified if it is better to error or fallback to default.
</blockquote></details>

Ack.  The thing is: a `*Client` (or really a `*Connection`) object can cache the low-level sockets, because it always makes requests to the same point.  I think that is property we want to work towards (as I said, it does not need to happen in this PR).  And typically we initialize the `*Connection` with default options, including the endpoint.

jsrinnn (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)* status: all files reviewed, 7 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 22 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-22:-NOod_s_8VKaDxIAmc19:b-ugv4m9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L22)):*

See my followup in the cc file.  It still seems unused.
</blockquote></details>

Done.

Ah..thats right. I forgot I removed the operator overload for ostream since it wasn't being used. Forgot to remove the header files. Thanks for checking that.

This is another example I feel can be addressed by checkers-pr i.e unused or redundant header files. 
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 67 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOR9FtW6hjkV-XhgVwo:-NOpJ_qFAwVWhRchkiJ6:bb9j0vb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L67)):*

Ack.  The thing is: a `*Client` (or really a `*Connection`) object can cache the low-level sockets, because it always makes requests to the same point.  I think that is property we want to work towards (as I said, it does not need to happen in this PR).  And typically we initialize the `*Connection` with default options, including the endpoint.
</blockquote></details>

Ack. Sounds good. Once I work on the connection classes maybe we can revisit this then.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 72 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOSXsmkAS-N1nm9T3TK:-NOom2xc38BHRsbTA2OR:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c2d309c7e4f2bee84bf990822aa6991c13aaf60/google/cloud/bigquery/v2/minimal/internal/job_request.h#L72)):*

`BuildRestRequest()` does appear to need to mutate the `GetJobRequest`, so `s/&/ const&/`.
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 21 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-21:-NOom6tf56gZPXMyYlot:b-wpp4sp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L21)):*

I don't think I understand.  There doesn't appear to be anything here, or in the header, that uses these includes.
</blockquote></details>

Done.

Sorry I forgot that ostream overload was removed. Please see my previous comment, this is now addressed.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 45 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-45:-NOonA1yDHh1vIU9SFlo:b-w1a0ku) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L45)):*

I don't think that ensuring we fallback to a default endpoint implies that we need to mutate the local `opts`.  Indeed, this function should probably take an `Options const&` to avoid an unnecessary copy.
</blockquote></details>

Done.

I modified it to `Options&` to avoid the copy. Since I am calling lookup, doing `Options const&` gives a compile error.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request_test.cc` line 39 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdQ-r2-39:-NOonbxuEB-F2pffrXSV:b-6lwqu9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request_test.cc#L39)):*

`s/actual.value().path()/actual->path()/`
</blockquote></details>

Done. 

Thanks!
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request_test.cc` line 75 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOSZvuC4kPBDg_5FgQd:-NOov2zL2HRSpiK7Se5I:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c2d309c7e4f2bee84bf990822aa6991c13aaf60/google/cloud/bigquery/v2/minimal/internal/job_request_test.cc#L75)):*

`opts` should probably be local to the loop.
</blockquote></details>

Done.

jsrinnn:
Addressed remaining code review feedback. PTAL.

Thanks!

devbww (commented):
> Reviewed 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)* status: 12 of 13 files reviewed, 3 unresolved discussions (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 22 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdP-r2-22:-NOpieqo9mbDAWdi5DHa:b-s34ohp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.h#L22)):*
> This is another example I feel can be addressed by checkers-pr i.e unused or redundant header files.

Yeah ... but that's hard.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 45 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-45:-NOphBPF2utCIS3g-ifZ:b51c6bo) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L45)):*
`Options&` wouldn't be recommended ... it won't bind to an rvalue, and it changes someone else's value.  Unless this really should be considered an in/out, passing by value or by const-ref should be preferred.

> Since I am calling lookup, doing `Options const&` gives a compile error.

I think that mostly indicates that `lookup()` is the wrong tool here.  Ensuring we fallback to a default endpoint shouldn't imply that we need to mutate the local `opts`.

jsrinnn (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)* status: 12 of 13 files reviewed, 3 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 45 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-45:-NOpmVCj2skaW827ibD1:b-q26esn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L45)):*

`Options&` wouldn't be recommended ... it won't bind to an rvalue, and it changes someone else's value.  Unless this really should be considered an in/out, passing by value or by const-ref should be preferred.

> Since I am calling lookup, doing `Options const&` gives a compile error.

I think that mostly indicates that `lookup()` is the wrong tool here.  Ensuring we fallback to a default endpoint shouldn't imply that we need to mutate the local `opts`.
</blockquote></details>

The goal is to set a default endpoint if none is set in the opts. For that I was using set<> before which was changed to lookup<> based on the feedback. Now I am confused if lookup<> is the wrong tool.

I am not sure I follow what is the recommended feedback here and what needs to change?

devbww (commented):
> Reviewed 1 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)* status: all files reviewed, 3 unresolved discussions (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 45 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-NOLcBDlg2WkFan1lFdO-r2-45:-NOpnrJP1pNv8iJcChVL:bxwswmv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7d06d70fcff4d65bf090a926a12c065bae6eed4c/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L45)):*

The goal is to set a default endpoint if none is set in the opts. For that I was using set<> before which was changed to lookup<> based on the feedback. Now I am confused if lookup<> is the wrong tool.

I am not sure I follow what is the recommended feedback here and what needs to change?
</blockquote></details>

Sorry things were unclear.  The idea is: don't change `opts`, just use `opts`.  So, something along the lines of ...
```
  std::string endpoint = "https://bigquery.googleapis.com/";
  if (opts.has<EndpointOption>()) {
    endpoint = opts.get<EndpointOptions>();
  }
```
[Aside: All that said, it isn't clear why `EndpointOption` hasn't been given a default value by the time we get here, but perhaps that is related to another thread.]

jsrinnn (commented):
Addressed the remaining feedback. PTAL

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-:-NOpsIbKElMl0cBQYEfj:b-wsmmhb)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @devbww)

devbww (approved):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837)* status: all files reviewed, 1 unresolved discussion (waiting on @jsrinnn)

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10837#-:-NOq7OOH2fcNCLQdxle1:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @devbww)

PR #10840, opened 2023-02-16
chore(deps): add GCS alternatives for remaining bazel dependencies

devbww:
Add GCS alternative URLs to download archives for all the external Bazel dependencies that didn't already have one.  Hopefully this will help reduce download flakiness.

Note that Bazel says, "URLs are tried in order until one succeeds, so you should list local mirrors first."  We can do some re-ordering later, if desired, to list GCS URLs first.

Also add a script to verify and (optionally) populate the GCS cache. For now I only propose that we run it manually when dependencies are updated, but we might consider running it from, say, the `checkers` build in the future.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10840)

### DIFF
--- a/bazel/deps-cache.sh
+++ b/bazel/deps-cache.sh
@@ -1,0 +1,203 @@
+#!/bin/bash
+#
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Usage:
+#   $ deps-cache.sh [-c] [<google_cloud_cpp_deps.bzl>]
+#   $ deps-cache.sh -p [<google_cloud_cpp_deps.bzl>]
+#
+#   Args:
+#     google_cloud_cpp_deps.bzl    Path to the .bzl file that specifies the
+#                                  HTTP archives that we cache in GCS. By
+#                                  default uses "google_cloud_cpp_deps.bzl"
+#                                  in the same directory as this script.
+#
+#   Options:
+#     -c     Check that the HTTP archives are correctly cached (the default)
+#     -p     Populate the cache with current copies of the HTTP archives
+#     -h     Print help message
+#
+# Note:
+#   This does not edit the .bzl file to add cache files as URLs. It only
+#   checks/populates the cached files themselves, so they're ready for use.
+
+set -euo pipefail
+
+# Extracts all the documentation at the top of this file as the usage text.
+USAGE="$(sed -n '/^# Usage:/,/^$/s/^# \?//p' "$0")"
+readonly USAGE
+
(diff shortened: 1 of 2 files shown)


bazel/google_cloud_cpp_deps.bzl
```diff
@@ -174,6 +181,7 @@ def google_cloud_cpp_deps(name = None):
         name = "com_github_nlohmann_json",
         urls = [
             "https://github.com/nlohmann/json/releases/download/v3.11.2/include.zip",
+            "https://storage.googleapis.com/cloud-cpp-community-archive/com_github_nlohmann_json/include.zip",
```
coryan, line 184: Hmmm... this is unfortunate, without the version number in the name it will be unusable for older branches, or by any branch / PR that is out of date right after we move to v3.12.x.  Maybe rename this to `gs://.../v3.11.2.zip` and some ad-hoc code to handle this case?  I am not sure if this would work, but there are different tarball/zip files, e.g., https://github.com/nlohmann/json/archive/refs/tags/v3.11.2.tar.gz

bazel/deps-cache.sh
```diff
@@ -0,0 +1,203 @@
...
+esac
+readonly BZL_FILE
+
+PROGRAM="$(basename "$0" .sh)"
+TMPDIR=$(mktemp -d --tmpdir "${PROGRAM}.XXXXXXXXXX")
+readonly PROGRAM TMPDIR
+trap 'rm -fr "${TMPDIR}"' EXIT
+
+# Extract the http_archive information from ${BZL_FILE}.
+cat >"${TMPDIR}/script" <<'EOT'
```
coryan, line 86: This is pretty impressive, and a little over my head.  If it helps (and only if it does), you can use something like this:

https://github.com/googleapis/google-cloud-cpp/compare/main...coryan:google-cloud-cpp:ci-automate-fallbacks-for-bazel-deps?expand=1

to extract the information.  Basically exploits the fact that a `.bzl` file has Python syntax.  If you stub out the functions you don't care about, and replace the functions you care about with things you want, you can get the output in an easier to parse.

bazel/google_cloud_cpp_deps.bzl
devbww: How about we simply download the full https://github.com/nlohmann/json/archive/v3.11.2.tar.gz archive, even though our `bazel/nlohmann_json.BUILD` file only uses the include files, and it is 7.8M instead of 288K? 
 Then there is no special case?

bazel/deps-cache.sh
devbww: I had thought about something like that, but stuck with the devil I knew (and also given our penchant for `sed`ing).  But I'll re-examine given your example, and see what I can come up with.  Thanks.  Although, if we go that route, we might as well just do the whole thing in python.

coryan: SGTM

bazel/google_cloud_cpp_deps.bzl
coryan: The full download SGTM too.

coryan (approved):
I am Okay with doing it he nlohmann fixes separately too.

bazel/deps-cache.sh
devbww: OK.  I'll merge as-is, and we can deal with any Python migration at a later date.

bazel/google_cloud_cpp_deps.bzl
devbww: Done.

PR #10851, opened 2023-02-16
chore(deps): prefer the GCS URL for downloading Bazel dependencies

devbww:
"Closer" mirrors should be listed first in `http_archive.urls`. In an unscientific test of sequential downloads of all the archives to my workstation, GCS wins by a 10x factor over GitHub for elapsed time (~1s -v- ~10s).  Nothing huge, but every second counts.

Also canonicalize the ordering of urls/sha256/strip_prefix fields.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10851)

### DIFF
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -36,9 +36,9 @@ def google_cloud_cpp_deps(name = None):
         http_archive,
         name = "platforms",
         urls = [
+            "https://storage.googleapis.com/cloud-cpp-community-archive/platforms/platforms-0.0.6.tar.gz",
             "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/0.0.6/platforms-0.0.6.tar.gz",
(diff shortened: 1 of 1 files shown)


bazel/google_cloud_cpp_deps.bzl
```diff
@@ -130,24 +130,24 @@ def google_cloud_cpp_deps(name = None):
         name = "boringssl",
         # Use github mirror instead of https://boringssl.googlesource.com/boringssl
```
dbolduc, line 131: maybe s/github/a/ ?

dbolduc (approved):

bazel/google_cloud_cpp_deps.bzl
devbww: I don't know ... "github mirror" seems strictly more informative than "a mirror" when referring to https://github.com/google/boringssl.

devbww: Move comment location to make it clearer.  Thanks.

PR #10853, opened 2023-02-17
impl(docfx): support for itemized lists

coryan:
This introduces support for itemized lists in Doxygen sections. I think this is the simplest thing that justifies adding the context and indentation.  The basic ideas in this change are:

- New paragraphs need a empty line just before them **unless** the paragraph is the start of a list item, where you just need a line starting with "- " (maybe with some indentation.
- Sub itemized lists and list items with multiple paragraphs need to be indented to preserve the structure in Markdown.

  For example: this is new paragraph, as it is precedeed by an empty line. It is also part of the same list item, by virtue of its indentation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10853)

### DIFF
--- a/docfx/doxygen2markdown.h
+++ b/docfx/doxygen2markdown.h
@@ -22,9 +22,7 @@
  * Keeps the state for markdown generation.
  *
  * As we recurse through the XML tree, we need to keep some information to
- * generate valid Markdown text. At the moment, this is just the indentation
- * level for new paragraphs, but we use a `struct` in anticipation of more
- * state.
+ * generate valid Markdown text.
  */
 struct MarkdownContext {
   std::string paragraph_start = "\n\n";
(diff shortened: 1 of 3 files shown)


docfx/doxygen2markdown.h
```diff
@@ -18,8 +18,54 @@
...
-bool AppendIfPlainText(std::ostream& os, pugi::xml_node const& node);
-bool AppendIfComputerOutput(std::ostream& os, pugi::xml_node const& node);
-bool AppendIfParagraph(std::ostream& os, pugi::xml_node const& node);
+/**
+ * Keeps the state for markdown generation.
+ *
+ * As we recurse through the XML tree, we need to keep some information to
+ * generate valid Markdown text. At the moment, this is just the indentation
+ * level for new paragraphs, but we use a `struct` in anticipation of more
+ * state.
```
devbww, line 27: Opt: Delete the "At the moment..." bit?

devbww (approved):

docfx/doxygen2markdown.h
coryan: Done.

PR #10857, opened 2023-02-19
refactor: versioned clients for `apigateway`, `apigeeconnect`

dbolduc:
Part of the work for #10170 

Do >1 libraries at a time. If this goes well, I will start packing something like 5-10 libraries / PR. All future PRs will have this structure.

I will post the script in the original issue after sending this for review.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10857)


coryan (approved):

coryan:
Probably better if we do just one library per PR.

dbolduc:
> Probably better if we do just one library per PR.

Ack, although I worry about build quotas

PR #10868, opened 2023-02-21
impl(spanner): push TransactionContext further down in ConnectionImpl

devbww:
This will avoid the need for new parameters when TransactionContext is expanded, which will be happening shortly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10868)


dbolduc (approved):
nit: I would have called this a `refactor`

PR #10918, opened 2023-02-22
impl(spanner): support route_to_leader in query/read partitions

devbww:
Add "route to leader" support to `QueryPartition` and `ReadPartition` so that RPCs in the eventual partitioned `ExecuteQuery()` and `Read()` calls will be routed according to the transaction used in the original `PartitionQuery()` and `PartitionRead()` calls (respectively).

Note that we support (de)serialization of the `QueryPartition` and `ReadPartition` values so that they might be transferred to another address space, and that this is done using `ExecuteSqlRequest` and `ReadRequest` protobuf messages, which do not include route-to-leader fields.  The best (only?) way to handle this while maintaining backwards compatibility is to add the route-to-leader value to those messages using a protobuf unknown field.

### DIFF
--- a/google/cloud/spanner/query_partition.cc
+++ b/google/cloud/spanner/query_partition.cc
@@ -20,12 +20,16 @@ namespace cloud {
 namespace spanner {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-QueryPartition::QueryPartition(std::string transaction_id,
+// Local extension to google::spanner::v1::ExecuteSqlRequest.
+constexpr int kRouteToLeaderFieldNumber = 511037314;
+
+QueryPartition::QueryPartition(std::string transaction_id, bool route_to_leader,
                                std::string transaction_tag,
                                std::string session_id,
                                std::string partition_token,
                                SqlStatement sql_statement)
     : transaction_id_(std::move(transaction_id)),
+      route_to_leader_(route_to_leader),
       transaction_tag_(std::move(transaction_tag)),
       session_id_(std::move(session_id)),
       partition_token_(std::move(partition_token)),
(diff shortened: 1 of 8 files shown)


devbww:
Regarding the `check-api` failure ... I assert that the parameter lists that have changed are private (the field-wise ctors).  If you concur, I can update the ABI dump, and we can look into the false negative later.

google/cloud/spanner/query_partition.cc
```diff
@@ -20,19 +20,24 @@ namespace cloud {
 namespace spanner {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-QueryPartition::QueryPartition(std::string transaction_id,
+// Local extension to google::spanner::v1::ExecuteSqlRequest.
+constexpr int kRouteToLeaderFieldNumber = 511037314;
```
coryan, line 24: Quite the magical number, where does it come from?

coryan (approved):
I agree with your assessment on the `check-api` build failures.

I get the use of the proto extension. The source of the magical number is a mystery to me.  Maybe clarify that?

google/cloud/spanner/query_partition.cc
devbww: My understanding of the convention to allocate extension field numbers in Google-owned protos is to use the CL number of the change that introduces them.  So, I consumed a couple of CLs to use here.  [We could have just used one, seeing as they are different message types, but separate numbers seemed clearer.]

I didn't want to commit a description like that to the source code, but would be happy to add anything you think might be appropriate.

coryan: Optional: maybe just say "reserved using Google's conventions" ?

devbww: SGTM.  Done.  Thanks.

PR #10944, opened 2023-02-23
refactor(iam): versioned v1 clients

dbolduc:
Part of the work for #10170 

This one was a mixed library, that also has a handwritten backwards compatibility `retry_traits.h`.

I am not going to update the samples to use versioned clients yet. I think cloud docs use region tags at HEAD, so it would be wise for us to wait for a release with the versioned clients before updating the samples.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10944)


dbolduc:
Oops, I messed up some of the handwritten samples. This PR needs some more work before review.

devbww (approved):

PR #10946, opened 2023-02-23
impl: fix updating of strip_prefix in renovate.sh

devbww:
`renovate.sh` knew that the editable `bazel/google_cloud_cpp_deps.bzl` elements were terminated by the "sha256" field ... until they weren't. (See "canonicalize the ordering of urls/sha256/strip_prefix fields" in #10851.)

The new sed script is just as fragile, but hopefully now a little more stable.

Also update the order of the substitution commands to match the file, for just a little more clarity.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10946)


dbolduc (approved):
Ah, thanks.

Issue #10413, opened 2022-12-09
Make integration tests more tolerant of transient failures


coryan:
Our integration tests often perform "non-idempotent" operations, but these can fail even if our code is working fine.  A typical test may have this structure:

```cc
  auto create = client.CreateFoo(....);
  ASSERT_STATUS_OK(create);

  auto update = client.UpdateFoo(...);
  ASSERT_STATUS_OK(update);

  auto del = client.DeleteFoo(...);
  ASSERT_STATUS_OK(del);
```

None of those operations are retried by default, as the operations are non-idempotent.  I think we could rewirte the tests as follows:

```cc
  auto create = client.CreateFoo(... Options{}.set<IdempotencyPolicy>(RetryEverything()));
  ASSERT_THAT(create, StatusIs(AnyOf(StatusCode::kOk, StatusCode::kAlreadyExists)));
```

Or if we don't want to retry the operation, we can write it as:

```cc
  auto create = client.CreateFoo(...);
  if (IsTransient(...)) GTEST_SKIP() << create.status();
```

This probably requires a mini-design doc, and will require breaking down the work per test.

For now I am going to consolidate all existing "transient error on non-idempotent operation" flakes to this one.

coryan:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/58c92e1b-75be-4b56-ba5c-9ee196d61462;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/f601caa1d9fe320d16f7cbf9fb0ce039aad95d24/fedora-36-bazel-integration-daily/log-58c92e1b-75be-4b56-ba5c-9ee196d61462.txt

coryan:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/7739ac75-4eff-489a-90cd-a43fbd790fb5;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/main/e69ab2388d536ae42163086417c2bafeed0f1156/fedora-36-bazel-integration-daily/log-7739ac75-4eff-489a-90cd-a43fbd790fb5.txt

dbolduc:
[...]

PR #10957, opened 2023-02-26
fix: retries for GetIamPolicy, TestIamPermissions

dbolduc:
Part of the work for #10413 

It seemed better to add a special case to the generator's default idempotency method than to maintain the configuration in `generator_config.textproto` for `N` services. And the `N+1`th service will have the proper defaults for the IAM methods without manual intervention.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10957)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -430,6 +430,10 @@ void SetHttpGetQueryParameters(
 
 std::string DefaultIdempotencyFromHttpOperation(
     google::protobuf::MethodDescriptor const& method) {
+  if (method.name() == "GetIamPolicy" ||
+      method.name() == "TestIamPermissions") {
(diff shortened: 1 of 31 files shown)


devbww (approved):

generator/internal/descriptor_utils.cc
```diff
@@ -430,6 +430,10 @@ void SetHttpGetQueryParameters(
 
 std::string DefaultIdempotencyFromHttpOperation(
     google::protobuf::MethodDescriptor const& method) {
+  if (method.name() == "GetIamPolicy" ||
```
coryan, line 433: nit: there is prior code to detect `SetIamPolicy` methods, maybe you should have followed the same structure as those?

Issue #10980, opened 2023-03-01
Generate protobuf files from services defined in Discovery Docs


scotthart:
In order to leverage the existing generator to produce code for GCP services that only support REST, we first need to emit proto files from the data contained in the [Discovery Doc](https://developers.google.com/discovery/v1/reference/apis) for that service.

PR #10982, opened 2023-03-01
impl(generator): add discovery support to generator config

scotthart:
First of many PRs for the work on #10980 

- Adds config proto support for discovery document defined services.
- Adds discovery to proto control flow in main.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10982)

### DIFF
--- a/generator/standalone_main.cc
+++ b/generator/standalone_main.cc
@@ -174,8 +174,8 @@ int WriteFeatureList(
   return 0;
 }
 
-google::protobuf::RepeatedPtrField<
-    google::cloud::cpp::generator::ServiceConfiguration>
+google::cloud::StatusOr<google::protobuf::RepeatedPtrField<
+    google::cloud::cpp::generator::ServiceConfiguration>>
 GenerateProtosForRestProducts(
     CommandLineArgs const& generator_args,
     google::protobuf::RepeatedPtrField<
@@ -190,9 +190,8 @@ GenerateProtosForRestProducts(
         google::cloud::generator_internal::GenerateProtosFromDiscoveryDoc(
             p.discovery_document_url(), generator_args.protobuf_proto_path,
             generator_args.googleapis_proto_path, generator_args.output_path);
-    if (status.ok()) {
-      services.Add(p.rest_service().begin(), p.rest_service().end());
-    }
+    if (!status.ok()) return status;
+    services.Add(p.rest_service().begin(), p.rest_service().end());
   }
 
   return services;
@@ -331,6 +330,7 @@ std::vector<std::future<google::cloud::Status>> GenerateCodeFromProtos(
  *  --output_path=<path> OPTIONAL defaults to current directory.
  */
 int main(int argc, char** argv) {
+  int rc = 0;
   absl::ParseCommandLine(argc, argv);
   google::cloud::LogSink::EnableStdClog(
       google::cloud::Severity::GCP_LS_WARNING);
@@ -352,7 +352,7 @@ int main(int argc, char** argv) {
 
   auto config = GetConfig(args.config_file);
   if (!config.ok()) {
-    GCP_LOG(ERROR) << "Failed to parse config file: " << args.config_file
+    GCP_LOG(FATAL) << "Failed to parse config file: " << args.config_file
                    << "\n";
   }
 
@@ -370,9 +370,10 @@ int main(int argc, char** argv) {
   // For REST only services, generate protos from Discovery Docs as needed.
   auto rest_services =
       GenerateProtosForRestProducts(args, config->discovery_products());
+  if (!rest_services.ok()) GCP_LOG(FATAL) << rest_services.status();
   // After all proto generation is complete, generate C++ code from those
   // protos.
-  auto rest_tasks = GenerateCodeFromProtos(args, rest_services);
+  auto rest_tasks = GenerateCodeFromProtos(args, *rest_services);
 
   std::string error_message;
   tasks.insert(tasks.end(), std::make_move_iterator(rest_tasks.begin()),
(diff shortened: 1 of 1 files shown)


generator/standalone_main.cc
```diff
@@ -314,8 +313,70 @@ int main(int argc, char** argv) {
...
+  GCP_LOG(INFO) << "proto_path = " << args.protobuf_proto_path << "\n";
+  GCP_LOG(INFO) << "googleapis_path = " << args.googleapis_proto_path << "\n";
+  GCP_LOG(INFO) << "config_file = " << args.config_file << "\n";
+  GCP_LOG(INFO) << "output_path = " << args.output_path << "\n";
+
+  auto config = GetConfig(args.config_file);
+  if (!config.ok()) {
+    GCP_LOG(ERROR) << "Failed to parse config file: " << args.config_file
+                   << "\n";
+  }
```
devbww, line 357: It looks like there should be something like `return 1` here, otherwise we go on to use `*config`.

generator/standalone_main.cc
```diff
@@ -161,66 +174,52 @@ int WriteFeatureList(
...
-
-  auto config = GetConfig(config_file);
-  if (!config.ok()) {
-    GCP_LOG(ERROR) << "Failed to parse config file: " << config_file << "\n";
+  for (auto const& p : rest_products) {
+    auto status =
+        google::cloud::generator_internal::GenerateProtosFromDiscoveryDoc(
+            p.discovery_document_url(), generator_args.protobuf_proto_path,
+            generator_args.googleapis_proto_path, generator_args.output_path);
+    if (status.ok()) {
```
devbww, line 193: Are we presuming that `GenerateProtosFromDiscoveryDoc()` has already emitted an error message if it returns a bad `Status`?  I think I'd make this be the error-message emitter instead.

generator/standalone_main.cc
```diff
@@ -314,8 +313,70 @@ int main(int argc, char** argv) {
...
+  // After all proto generation is complete, generate C++ code from those
+  // protos.
+  auto rest_tasks = GenerateCodeFromProtos(args, rest_services);
 
   std::string error_message;
+  tasks.insert(tasks.end(), std::make_move_iterator(rest_tasks.begin()),
+               std::make_move_iterator(rest_tasks.end()));
   for (auto& t : tasks) {
     auto result = t.get();
     if (!result.ok()) {
```
devbww, line 382: Should we just `GCP_LOG(ERROR) << result` right here (and remember enough to return a non-zero code)?  In general we probably shouldn't `GCP_LOG` anything with embedded newlines.

generator/standalone_main.cc
devbww, line 193: How does an error here result in the whole generator invocation failing?

generator/internal/discovery_to_proto.cc
```diff
@@ -0,0 +1,31 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "generator/internal/discovery_to_proto.h"
+#include "google/cloud/internal/make_status.h"
+#include <string>
```
devbww, line 17: Opt Nit: Unnecessary unless we think the header file could have forward-declared `std::string`.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10982)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @devbww)

___
*[`generator/standalone_main.cc` line 193 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10982#-NPTzy_d-yT_QVRx7Xq_-r1-193:-NPYpT8M4mYH8oIbMMu2:blzezzg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/045c6b7c63eef6126ede959d7e972adc8d386f21/generator/standalone_main.cc#L193)):*

How does an error here result in the whole generator invocation failing?
</blockquote></details>

Added error handling.
___
*[`generator/standalone_main.cc` line 357 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10982#-NPTzy_d-yT_QVRx7Xq_-r1-357:-NPYcNPZ311uscpMVzfN:b-ngavvm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/045c6b7c63eef6126ede959d7e972adc8d386f21/generator/standalone_main.cc#L357)):*

It looks like there should be something like `return 1` here, otherwise we go on to use `*config`.
</blockquote></details>

Changed to FATAL.
___
*[`generator/standalone_main.cc` line 382 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10982#-NPTzy_d-yT_QVRx7Xq_-r1-382:-NPYgAEeBtwq80muqbcr:byg433c) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/045c6b7c63eef6126ede959d7e972adc8d386f21/generator/standalone_main.cc#L382)):*

Should we just `GCP_LOG(ERROR) << result` right here (and remember enough to return a non-zero code)?  In general we probably shouldn't `GCP_LOG` anything with embedded newlines.
</blockquote></details>

Refactored.
___
*[`generator/internal/discovery_to_proto.cc` line 17 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10982#-NPTzy_d-yT_QVRx7Xqa-r1-17:-NPYgKMQ5O5n_XVKTzga:bss4a2b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/045c6b7c63eef6126ede959d7e972adc8d386f21/generator/internal/discovery_to_proto.cc#L17)):*

Opt Nit: Unnecessary unless we think the header file could have forward-declared `std::string`.
</blockquote></details>

Removed.

devbww (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10982)* status: all files reviewed, 1 unresolved discussion (waiting on @scotthart)

PR #10995, opened 2023-03-03
refactor(composer): versioned clients

dbolduc:
Part of the work for #10170 

We all agreed that this one should be versioned too.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10995)


coryan (approved):

generator/generator_config.textproto
```diff
@@ -412,14 +412,16 @@ service {
 service {
   service_proto_path: "google/cloud/orchestration/airflow/service/v1/environments.proto"
   additional_proto_files: ["google/cloud/orchestration/airflow/service/v1/operations.proto"]
-  product_path: "google/cloud/composer"
+  product_path: "google/cloud/composer/v1"
```
devbww, line 415: Can we put the new stuff in "google/cloud/orchestration" and still forward from "google/cloud/composer"?

dbolduc: I think the build files would get very messy.

It would probably be less effort to:
- check whether `composer` is in use
- if not, claim no break
- move the library, ignoring backwards compatibility

devbww: Ack.

devbww (approved):

Issue #6268, opened 2021-04-15
Rationalize GCS+gRPC integration tests


coryan:
# Objective

We want our customers to have a good experience with the GCS+gRPC plugin when it becomes available, that requires extensive integration tests.  We also want to detect defects as early as possible during development, that argues for early integration tests.

# Background

Unlike other services the GCS+gRPC plugin is under development more or less at the same time that the service is.  We have often developed *features* at the same time as the service is under development, we just implement the features in the emulator and test them only against the emulator until the service is ready.  This is an extreme case of that approach.

To support simultaneous development the GCS emulator supports both REST and gRPC APIs. However, only the data operations (`InsertObject()`, `WriteObject`, `ReadObject()`) are implemented in the service.

To reuse existing integration tests and benchmarks we implemented a "hybrid" plugin. This routes requests to the gRPC plugin when possible (i.e. when implemented in production), and routes other requests to the existing REST-based implementation (i.e., storage::internal::CurlClient).

The plugin itself is under development, some features are not implemented in the GCS+gRPC plugin.

The storage integration tests use `GOOGLE_CLOUD_CPP_STORAGE_GRPC_CONFIG` to select which plugin they are testing. The integration tests use the following values:

* If set to `media`, use the hybrid plugin, only data operations are sent to production or the emulator.
* If set to `metadata`, use the gRPC plugin, both data and metadata operations are sent to production or the emulator, as only the emulator supports this, it should not be enabled in builds using production.
* any other value, or if unset, use the REST-based plugin.

# Goals

We want to:

* **(A)** Test the GCS+gRPC plugin as early as possible, using the emulator to test as many features as possible.
* **(B)** Test the GCS+gRPC plugin against production for those operations that are supported in production.
* **(C)** Test that the code can be compiled with the GCS+gRPC plugin turned on or off, so applications can omit building the plugin.

# Current Status

The `gcs-grpc` build is specifically designed to satisfy **(B)**.  It runs **all** the storage integration tests against production, with the hybrid plugin to route operations through the right protocol. Note that as of 2021-04-15 this is not true, because the service has been down since 2021-03-18 (see #6062), but normally it would be (see #6063).

The builds running against the emulator test **(A)** because there is one integration test (`grpc_integration_test.cc`) that always tests with both `GOOGLE_CLOUD_CPP_STORAGE_GRPC_CONFIG=media` and `GOOGLE_CLOUD_CPP_STORAGE_GRPC_CONFIG=metadata`.

[...]

PR #10996, opened 2023-03-03
test(storage): simplify testing transport variants

coryan:
The storage library integration tests can run using (1) the normal HTTP transport, (2) the legacy HTTP implementation, (3) use GCS+gRPC for only media operations, or (4) use GCS+gRPC for media and metadata operations. We were testing all these variations by calling `bazel` with different environment variables, that made our test scripts more complicated, it wastes parallelism opportunities, and it pays the bazel startup costs multiple times. It also seems that running the same test with different environment variables disables test result caching.

This PR moves the complexity to the Bazel file: it creates multiple "variants" of each integration test, with different environment variables. The program drivers can exclude some variants using test tags.

Part of the work for #6268

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10996)

### DIFF
--- a/google/cloud/storage/tests/service_account_integration_test.cc
+++ b/google/cloud/storage/tests/service_account_integration_test.cc
@@ -63,6 +63,10 @@ TEST_F(ServiceAccountIntegrationTest, Get) {
 }
 
 TEST_F(ServiceAccountIntegrationTest, CreateHmacKeyForProject) {
+  // HMAC keys are a scarse resource. Testing in production would require
+  // redesigning the tests to use a random service account (or creating one) on
+  // dynamically.  For now, simply skip these tests.
+  if (!UsingEmulator()) GTEST_SKIP();
   Client client(Options{}.set<ProjectIdOption>(project_id_));
 
   StatusOr<std::pair<HmacKeyMetadata, std::string>> key = client.CreateHmacKey(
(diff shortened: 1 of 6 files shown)


google/cloud/storage/tests/service_account_integration_test.cc
```diff
@@ -63,6 +63,10 @@ TEST_F(ServiceAccountIntegrationTest, Get) {
 }
 
 TEST_F(ServiceAccountIntegrationTest, CreateHmacKeyForProject) {
+  // HMAC keys are a scarse resource. Testing in production would require
```
dbolduc, line 66: nit: s/scarse/scarce/

google/cloud/storage/tests/service_account_integration_test.cc
```diff
@@ -63,6 +63,10 @@ TEST_F(ServiceAccountIntegrationTest, Get) {
...
+  // redesigning the tests to use a random service account (or creating one) on
```
dbolduc, line 67: s/one) on/one)/

google/cloud/storage/tests/service_account_integration_test.cc
```diff
@@ -81,6 +85,10 @@ TEST_F(ServiceAccountIntegrationTest, CreateHmacKeyForProject) {
 }
 
 TEST_F(ServiceAccountIntegrationTest, HmacKeyCRUD) {
+  // HMAC keys are a scarse resource. Testing in production would require
+  // redesigning the tests to use a random service account (or creating one) on
```
dbolduc, line 89: ditto

dbolduc (approved):

google/cloud/storage/tests/service_account_integration_test.cc
coryan: Fixed

google/cloud/storage/tests/service_account_integration_test.cc
coryan: Fixed

google/cloud/storage/tests/service_account_integration_test.cc
coryan: Fixed

PR #10998, opened 2023-03-03
feat(spanner): support "data boost" on partitioned queries and reads

devbww:
Setting the `spanner::PartitionDataBoostOption` on a `PartitionRead()` or `PartitionQuery()` will result in the respective partitioned `Read()` and `ExecuteQuery()` requests being executed via Spanner independent compute resources.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/10998)


google/cloud/spanner/partition_options.h
```diff
@@ -54,11 +54,20 @@ struct PartitionOptions {
    * returned may be smaller or larger than this maximum count request.
    */
   absl::optional<std::int64_t> max_partitions;
+
+  /**
```
coryan, line 58: Are we planning to deprecate these `*Options` classes?  I don't mean remove, I mean just stop recommending them.  If so, we should stop adding features to them.  Probably a topic for future discussion.

coryan (approved):

google/cloud/spanner/partition_options.h
devbww: We have kind-of deprecated them in the `Client` API, in as much as those interfaces are marked as only for backwards compatibility, and we encouraged users to migrate away in a changelog note.

However, they are still used in the `Connection` API, so they're not really deprecated individually, and we still depend on them being able to carry new data.  I don't believe we have any plans to change that.

PR #11000, opened 2023-03-03
cleanup: use versioned clients in quickstart

dbolduc:
Most of the remaining work for #10170 

(there is still updating IAM, BQ samples. and dealing with `composer`)

Any remotely interesting decisions were separated into different commits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11000)


dbolduc:
I think I need to wait until the next vcpkg release. Closing for now...

Issue #10618, opened 2023-01-19
Async OpenTelemetry tracing in common/grpc_utils


dbolduc:
Some tasks from [Milestone 2](https://docs.google.com/document/d/1JTOWvZD59AGxojTV93tKZJ9b_q2w8YUMMyuxCu5_ZC4/edit?resourcekey=0-0XjX6fh2xNKFZN20cu2otg#bookmark=id.6hmdie574rpy) of an internal design doc. With these tasks completed, we can support tracing an asynchronous gRPC client call.

- [ ] Implement `EndSpan(...)` for `future<>` types.
- [ ] Use `CallContext` instead of `OptionsSpan`. These are mainly in `AsyncGrpcOperation`s.
- [ ] Add traced async backoff; update `(Rest)Async{Retry,Polling}Loop` to use it
- [ ] Add `SetAttribute`; Update `AsyncPollingLoop` to set LRO name.

PR #11002, opened 2023-03-06
refactor: prepare for `shared_ptr<grpc::ClientContext>` in tests

dbolduc:
Part of the work for #10618 

This goofy looking changes just uses `auto` instead of `unique_ptr<grpc::ClientContext>` in unit tests.

We are going to change the `unique_ptr` to `shared_ptr`. Unless the `context` is moved, clang-tidy will yell at us for having a performance unnecessary value parameter. clang-tidy is apparently not smart enough to warn when we use `auto` instead. 

This also lets me break a chunk off of the larger refactor.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11002)

### DIFF
--- a/google/cloud/internal/async_polling_loop_test.cc
+++ b/google/cloud/internal/async_polling_loop_test.cc
@@ -66,20 +66,18 @@ class MockPollingPolicy : public PollingPolicy {
 };
 
 AsyncPollLongRunningOperation MakePoll(std::shared_ptr<MockStub> const& mock) {
-  return
-      [mock](CompletionQueue& cq, std::unique_ptr<grpc::ClientContext> context,
-             google::longrunning::GetOperationRequest const& request) {
-        return mock->AsyncGetOperation(cq, std::move(context), request);
-      };
+  return [mock](CompletionQueue& cq, auto context,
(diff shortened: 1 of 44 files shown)


coryan (approved):

google/cloud/internal/async_polling_loop_test.cc
```diff
@@ -66,20 +66,18 @@ class MockPollingPolicy : public PollingPolicy {
 };
 
 AsyncPollLongRunningOperation MakePoll(std::shared_ptr<MockStub> const& mock) {
-  return
-      [mock](CompletionQueue& cq, std::unique_ptr<grpc::ClientContext> context,
-             google::longrunning::GetOperationRequest const& request) {
-        return mock->AsyncGetOperation(cq, std::move(context), request);
-      };
+  return [mock](CompletionQueue& cq, auto context,
```
devbww, line 69: Do we need similar changes for cases like lines 51 and 56 in this file?

dbolduc: Those come later.

In one big PR I will update the stub signatures in the generator / handwritten code, internal classes, and the signatures for these test mocks.

PR #11004, opened 2023-03-06
refactor: MakeUnaryRpc that is not a CQ member

dbolduc:
Part of the work for #10618 

Introduce `MakeUnaryRpcImpl(...)` which exists outside of the CQ. Note that it accepts a `std::shared_ptr<grpc::ClientContext>`. Update `AsyncUnaryRpcFuture` to interact with a `std::shared_ptr<grpc::ClientContext>`.

I could use double checking on the templates when split into definition vs. declaration. Also on moving the context into the lambda. I am fairly sure I want the lambda to own the context.

Note that the new test is just a copy of the existing `TEST(CompletionQueueTest, MakeUnaryRpc)`. I considered updating the test to use a MockCQ, but it was more trouble than it's worth. (Just mocking `StartOperation` then calling `op->Notify(true);` will not actually test the stuff in `AsyncUnaryRpcFuture::Start()`.)

We should deprecate `CQ::MakeUnaryRpc` and `CQ::MakeStreamingReadRpc` later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11004)

### DIFF
--- a/google/cloud/completion_queue.h
+++ b/google/cloud/completion_queue.h
@@ -31,8 +31,19 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 class CompletionQueue;
 
 namespace internal {
+
 std::shared_ptr<CompletionQueueImpl> GetCompletionQueueImpl(
     CompletionQueue const& cq);
+
+template <
+    typename AsyncCallType, typename Request,
+    typename Sig = internal::AsyncCallResponseType<AsyncCallType, Request>,
+    typename Response = typename Sig::type,
+    typename std::enable_if<Sig::value, int>::type = 0>
+future<StatusOr<Response>> MakeUnaryRpcImpl(
+    CompletionQueue& cq, AsyncCallType async_call, Request const& request,
+    std::shared_ptr<grpc::ClientContext> context);
+
 }  // namespace internal
 
 /**
@@ -125,13 +136,9 @@ class CompletionQueue {
       typename std::enable_if<Sig::value, int>::type = 0>
   future<StatusOr<Response>> MakeUnaryRpc(
       AsyncCallType async_call, Request const& request,
-      std::unique_ptr<grpc::ClientContext> context) {
-    auto op =
-        std::make_shared<internal::AsyncUnaryRpcFuture<Request, Response>>();
-    impl_->StartOperation(op, [&](void* tag) {
-      op->Start(async_call, std::move(context), request, impl_->cq(), tag);
-    });
-    return op->GetFuture();
+      std::shared_ptr<grpc::ClientContext> context) {
+    return internal::MakeUnaryRpcImpl(*this, std::move(async_call), request,
+                                      std::move(context));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


google/cloud/completion_queue.h
```diff
@@ -125,13 +136,9 @@ class CompletionQueue {
...
   future<StatusOr<Response>> MakeUnaryRpc(
       AsyncCallType async_call, Request const& request,
-      std::unique_ptr<grpc::ClientContext> context) {
-    auto op =
-        std::make_shared<internal::AsyncUnaryRpcFuture<Request, Response>>();
-    impl_->StartOperation(op, [&](void* tag) {
-      op->Start(async_call, std::move(context), request, impl_->cq(), tag);
-    });
-    return op->GetFuture();
+      std::shared_ptr<grpc::ClientContext> context) {
```
coryan, line 139: Did you mean to make this a `std::shared_ptr<>`?  Seems like an unrelated change.

google/cloud/completion_queue.h
```diff
@@ -31,8 +31,19 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 class CompletionQueue;
 
 namespace internal {
+
 std::shared_ptr<CompletionQueueImpl> GetCompletionQueueImpl(
     CompletionQueue const& cq);
+
+template <
+    typename AsyncCallType, typename Request,
```
coryan, line 39: There may be an opportunity to simplify this.

https://godbolt.org/z/dhGqTWsbv

Where "simplify" means "use less hand-crafted type deduction and let the compiler do it".

google/cloud/completion_queue.h
dbolduc: No, good catch. That was an unintentional change.

(It only now dawns on me that changing this method to accept a `std::shared_ptr<>` is not a breaking change. So I was not blocked on it. Oh well, we want this change anyway.)

google/cloud/completion_queue.h
dbolduc: TIL: absl includes in godbolt.

A consequence of this seems to be that we have to explicitly name the `Request` and `Response` types when calling the method. Does that seem right?

```
<lambda(grpc::ClientContext*, const google::bigtable::admin::v2::GetTableRequest&, grpc::CompletionQueue*)>' 
is not derived from
'absl::lts_20230125::FunctionRef<std::unique_ptr<grpc::ClientAsyncResponseReaderInterface<Response> >(grpc::ClientContext*, const Request&, grpc::CompletionQueue*)>'
```

coryan: > 
> A consequence of this seems to be that we have to explicitly name the `Request` and `Response` types when calling the method. Does that seem right?

That seems unfortunate, but for an internal / implementation function it may be a plus.  Your call, I am not convinced my idea improved things.  But maybe it did.

coryan (approved):

google/cloud/completion_queue.h
dbolduc: Let's go with it. If you ask me, it is easier to understand.

PR #11005, opened 2023-03-06
impl(otel): handle future<T>

dbolduc:
Part of the work for #10618

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11005)


google/cloud/internal/opentelemetry.h
```diff
@@ -108,12 +109,28 @@ Status EndSpan(opentelemetry::trace::Span& span, Status const& status);
  * composition.
  */
 template <typename T>
```
devbww, line 111: Why declare the template immediately before defining it?

dbolduc: Because I don't know how to C++. Removed.

PR #11014, opened 2023-03-08
feat(storage): reduce copies in `InsertObject()`

coryan:
We can avoid a data copy if we use `absl::string_view` to contain the object payload.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11014)

### DIFF
--- a/google/cloud/storage/hashing_options.h
+++ b/google/cloud/storage/hashing_options.h
@@ -17,6 +17,7 @@
 
 #include "google/cloud/storage/internal/complex_option.h"
 #include "google/cloud/storage/version.h"
+#include "absl/strings/string_view.h"
 #include <string>
 
 namespace google {
@@ -45,9 +46,19 @@ struct MD5HashValue
 };
 
 /**
- * Compute the MD5 Hash of a string in the format preferred by GCS.
+ * Compute the MD5 Hash of a buffer in the format preferred by GCS.
  */
-std::string ComputeMD5Hash(std::string const& payload);
+std::string ComputeMD5Hash(absl::string_view payload);
+
+/// @overload ComputeMD5Hash(absl::string_view)
+inline std::string ComputeMD5Hash(std::string const& payload) {
+  return ComputeMD5Hash(absl::string_view(payload));
+}
+
+/// @overload ComputeMD5Hash(absl::string_view)
+inline std::string ComputeMD5Hash(char const* payload) {
+  return ComputeMD5Hash(absl::string_view(payload));
+}
 
 /**
  * Disable or enable MD5 Hashing computations.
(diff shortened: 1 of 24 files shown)


google/cloud/storage/internal/curl_request.cc
```diff
@@ -68,11 +68,11 @@ CurlRequest::~CurlRequest() {
   if (factory_) CurlHandle::ReturnToPool(*factory_, std::move(handle_));
 }
 
-StatusOr<HttpResponse> CurlRequest::MakeRequest(std::string const& payload) && {
+StatusOr<HttpResponse> CurlRequest::MakeRequest(absl::string_view payload) && {
   handle_.SetOption(CURLOPT_UPLOAD, 0L);
   if (!payload.empty()) {
     handle_.SetOption(CURLOPT_POSTFIELDSIZE, payload.length());
-    handle_.SetOption(CURLOPT_POSTFIELDS, payload.c_str());
+    handle_.SetOption(CURLOPT_POSTFIELDS, payload.data());
```
devbww, line 75: `string_view::data()` without `string_view::size()` is a red flag.

google/cloud/storage/hashing_options.h
```diff
@@ -45,10 +46,20 @@ struct MD5HashValue
...
+std::string ComputeMD5Hash(absl::string_view payload);
+
+/// @copydoc ComputeMD5Hash(absl::string_view)
 std::string ComputeMD5Hash(std::string const& payload);
 
+/// @copydoc ComputeMD5Hash(absl::string_view)
+inline std::string ComputeMD5Hash(char const* payload) {
+  auto p =
+      payload == nullptr ? absl::string_view{} : absl::string_view{payload};
+  return ComputeMD5Hash(std::move(p));
```
devbww, line 60: Moving a `string_view` is an anti-pattern, I believe (et seq).

google/cloud/storage/internal/openssl_util.h
```diff
@@ -80,7 +81,16 @@ inline std::string UrlsafeBase64Encode(Collection const& bytes) {
 StatusOr<std::vector<std::uint8_t>> UrlsafeBase64Decode(std::string const& str);
 
 /// Compute the MD5 hash of @p payload
-std::vector<std::uint8_t> MD5Hash(std::string const& payload);
+std::vector<std::uint8_t> MD5Hash(absl::Span<char const> payload);
```
devbww, line 84: Why did we need to introduce a `Span` overload?

coryan: Removed.  I am thinking that these should be receiving "bytes" as input, but we should do that in a separate PR.

google/cloud/storage/hashing_options.h
coryan: Fixed (everywhere I hope).

google/cloud/storage/internal/curl_request.cc
coryan: There was a `.length()` already, changed to `.size()`.

devbww: Ahh ... sorry I missed that.

devbww (approved):

PR #11016, opened 2023-03-09
impl(bigquery): Remaining implementation files for GetJob api.

jsrinnn:
All the files modified are connected to each other and are referenced which is why I didn't want to split them up.

Connection tests are currently stubbed out. Connection tests implementation with detailed mocks will be added in the subsequent PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016)

### DIFF
--- a/google/cloud/bigquery/BUILD.bazel
+++ b/google/cloud/bigquery/BUILD.bazel
@@ -109,7 +109,6 @@ cc_library(
     visibility = ["//:__pkg__"],
     deps = [
         "//:common",
-        "//:grpc_utils",
         "//google/cloud:google_cloud_cpp_rest_internal",
     ],
 )
--- a/google/cloud/bigquery/BUILD.bazel
+++ b/google/cloud/bigquery/BUILD.bazel
@@ -111,7 +111,6 @@ cc_library(
         "//:common",
         "//:grpc_utils",
         "//google/cloud:google_cloud_cpp_rest_internal",
-        "//google/cloud:google_cloud_cpp_rest_protobuf_internal",
     ],
 )
 
--- a/google/cloud/bigquery/BUILD.bazel
+++ b/google/cloud/bigquery/BUILD.bazel
@@ -109,7 +109,9 @@ cc_library(
     visibility = ["//:__pkg__"],
     deps = [
         "//:common",
+        "//:grpc_utils",
         "//google/cloud:google_cloud_cpp_rest_internal",
+        "//google/cloud:google_cloud_cpp_rest_protobuf_internal",
     ],
 )
 
--- a/google/cloud/bigquery/bigquery_rest.cmake
+++ b/google/cloud/bigquery/bigquery_rest.cmake
@@ -22,6 +22,8 @@ add_library(
     v2/minimal/internal/common_v2_resources.h
     v2/minimal/internal/job.h
     v2/minimal/internal/job_configuration.h
+    v2/minimal/internal/job_connection.cc
+    v2/minimal/internal/job_connection.h
     v2/minimal/internal/job_idempotency_policy.cc
     v2/minimal/internal/job_idempotency_policy.h
     v2/minimal/internal/job_logging.cc
@@ -47,8 +49,9 @@ target_include_directories(
            $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>
            $<INSTALL_INTERFACE:include>)
 target_link_libraries(
-    google_cloud_cpp_bigquery_rest PUBLIC google-cloud-cpp::rest_internal
-                                          google-cloud-cpp::common)
+    google_cloud_cpp_bigquery_rest
+    PUBLIC google-cloud-cpp::rest_internal google-cloud-cpp::grpc_utils
+           google-cloud-cpp::common)
 google_cloud_cpp_add_common_options(google_cloud_cpp_bigquery_rest)
 set_target_properties(
     google_cloud_cpp_bigquery_rest
@@ -78,6 +81,9 @@ function (bigquery_rest_define_tests)
         # cmake-format: sort
         v2/minimal/internal/bigquery_http_response_test.cc
         v2/minimal/internal/common_v2_resources_test.cc
+        v2/minimal/internal/job_connection_test.cc
+        v2/minimal/internal/job_idempotency_policy_test.cc
+        v2/minimal/internal/job_options_test.cc
         v2/minimal/internal/job_request_test.cc
         v2/minimal/internal/job_response_test.cc
         v2/minimal/internal/job_rest_stub_test.cc)
--- a/google/cloud/bigquery/bigquery_rest.cmake
+++ b/google/cloud/bigquery/bigquery_rest.cmake
@@ -50,7 +50,8 @@ target_include_directories(
            $<INSTALL_INTERFACE:include>)
 target_link_libraries(
     google_cloud_cpp_bigquery_rest
-    PUBLIC google-cloud-cpp::rest_internal google-cloud-cpp::grpc_utils
+    PUBLIC google-cloud-cpp::rest_internal
+           google_cloud_cpp_rest_protobuf_internal google-cloud-cpp::grpc_utils
            google-cloud-cpp::common)
 google_cloud_cpp_add_common_options(google_cloud_cpp_bigquery_rest)
 set_target_properties(
--- a/google/cloud/bigquery/bigquery_rest.cmake
+++ b/google/cloud/bigquery/bigquery_rest.cmake
@@ -50,8 +50,7 @@ target_include_directories(
            $<INSTALL_INTERFACE:include>)
 target_link_libraries(
     google_cloud_cpp_bigquery_rest
-    PUBLIC google-cloud-cpp::rest_internal
-           google_cloud_cpp_rest_protobuf_internal google-cloud-cpp::grpc_utils
+    PUBLIC google-cloud-cpp::rest_internal google-cloud-cpp::grpc_utils
            google-cloud-cpp::common)
 google_cloud_cpp_add_common_options(google_cloud_cpp_bigquery_rest)
 set_target_properties(
--- a/google/cloud/bigquery/bigquery_rest.cmake
+++ b/google/cloud/bigquery/bigquery_rest.cmake
@@ -49,9 +49,8 @@ target_include_directories(
            $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>
            $<INSTALL_INTERFACE:include>)
 target_link_libraries(
-    google_cloud_cpp_bigquery_rest
-    PUBLIC google-cloud-cpp::rest_internal google-cloud-cpp::grpc_utils
-           google-cloud-cpp::common)
+    google_cloud_cpp_bigquery_rest PUBLIC google-cloud-cpp::rest_internal
+                                          google-cloud-cpp::common)
 google_cloud_cpp_add_common_options(google_cloud_cpp_bigquery_rest)
 set_target_properties(
     google_cloud_cpp_bigquery_rest
--- a/google/cloud/bigquery/bigquery_rest_unit_tests.bzl
+++ b/google/cloud/bigquery/bigquery_rest_unit_tests.bzl
@@ -19,6 +19,9 @@
 bigquery_rest_unit_tests = [
     "v2/minimal/internal/bigquery_http_response_test.cc",
     "v2/minimal/internal/common_v2_resources_test.cc",
+    "v2/minimal/internal/job_connection_test.cc",
+    "v2/minimal/internal/job_idempotency_policy_test.cc",
+    "v2/minimal/internal/job_options_test.cc",
     "v2/minimal/internal/job_request_test.cc",
     "v2/minimal/internal/job_response_test.cc",
     "v2/minimal/internal/job_rest_stub_test.cc",
--- a/google/cloud/bigquery/google_cloud_cpp_bigquery_rest.bzl
+++ b/google/cloud/bigquery/google_cloud_cpp_bigquery_rest.bzl
@@ -21,6 +21,7 @@ google_cloud_cpp_bigquery_rest_hdrs = [
     "v2/minimal/internal/common_v2_resources.h",
     "v2/minimal/internal/job.h",
     "v2/minimal/internal/job_configuration.h",
+    "v2/minimal/internal/job_connection.h",
     "v2/minimal/internal/job_idempotency_policy.h",
     "v2/minimal/internal/job_logging.h",
     "v2/minimal/internal/job_metadata.h",
@@ -36,6 +37,7 @@ google_cloud_cpp_bigquery_rest_hdrs = [
 google_cloud_cpp_bigquery_rest_srcs = [
     "v2/minimal/internal/bigquery_http_response.cc",
     "v2/minimal/internal/common_v2_resources.cc",
+    "v2/minimal/internal/job_connection.cc",
     "v2/minimal/internal/job_idempotency_policy.cc",
     "v2/minimal/internal/job_logging.cc",
     "v2/minimal/internal/job_metadata.cc",
--- a/google/cloud/bigquery/v2/minimal/internal/job_connection.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_connection.cc
@@ -15,10 +15,8 @@
 #include "google/cloud/bigquery/v2/minimal/internal/job_connection.h"
 #include "google/cloud/bigquery/v2/minimal/internal/job_rest_connection_impl.h"
 #include "google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.h"
-#include "google/cloud/background_threads.h"
 #include "google/cloud/common_options.h"
 #include "google/cloud/credentials.h"
-#include "google/cloud/internal/rest_background_threads_impl.h"
 #include <memory>
 
 namespace google {
@@ -38,13 +36,11 @@ std::shared_ptr<BigQueryJobConnection> MakeBigQueryJobConnection(
                                  BigQueryJobPolicyOptionList>(options,
                                                               __func__);
   options = BigQueryJobDefaultOptions(std::move(options));
-  auto background = absl::make_unique<
-      rest_internal::AutomaticallyCreatedRestBackgroundThreads>();
 
   auto job_rest_stub = CreateDefaultBigQueryJobRestStub(options);
 
   return std::make_shared<BigQueryJobRestConnectionImpl>(
-      std::move(background), std::move(job_rest_stub), std::move(options));
+      std::move(job_rest_stub), std::move(options));
 }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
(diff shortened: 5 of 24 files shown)


jsrinnn (commented):
Sorry about that. I included all the files that demonstrated how the different classes were setup/wired.  I included implementation for classes that were boiler plate and fairly straight forward. Any class that required verbose or complex logic I saved for the next PR (example tests with detailed mocks etc) 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-:-NQ66OBf3MdZouvUB6Uj:b-dd5ke)* status: 22 of 25 files reviewed, 7 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/bigquery/bigquery_rest.cmake` line 53 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5ajQBCsGMrNde3Q5r:-NQ5tRL1Etn3PQUI26CB:b-x8wciv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/bigquery_rest.cmake#L53)):*

The list of libraries here does not match the libraries in `BUILD.bazel`.  I am not sure which one is right, but at least one of them is wrong.
</blockquote></details>

Done.

I have matched them both now.  Thank you for pointing this out. 

Cmake  build and tests seemed to  work fine without the google_cloud_cpp_rest_protobuf_internal but bazel build complained hence the initial discrepancy. But now I realize some of the pr checks like `shared-pr` were failing with "undefined reference" error. The specific error is now fixed with this change. 

I still can't entirely run the `shared-pr` locally as I don't have required test permissions. But I think this pr check should pass now with the above fix.

Similar to the above, I have also verified `clang-60-pr` is now passing . The other pr checks that were originally failing (quickstart-production-pr, gcc-7-3-pr) should pass now as well. I can test those locally due to permissions issue.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_connection.cc` line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5ioJd4DGjFYXCpoeN:-NQ5xGqbEF_a7VIzGHVS:b-mwdrxn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_connection.cc#L44)):*

Note that you always call `CreateDefaultBigQueryJobRestStub` with an `options` that has all the defaults set.
</blockquote></details>

The options passed in is merged with any missing default options inside the BigQueryJobDefaultOptions method.

Just wanted to check if there is an action item to follow up for this feedback?
___
*[`google/cloud/bigquery/v2/minimal/internal/job_connection_test.cc` line 25 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5j3vi2qTlkFh4hWYn:-NQ5ybE95OtTxN-LnDaB:b-9o4zfh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_connection_test.cc#L25)):*

grumble grumble ... I would rather you did not introduce the `JobConnection` class in this PR and have a separate PR with it and its tests.
</blockquote></details>

Apologies again. The connection class implementation itself was pretty straight forward and boiler plate which is why I included it. The test classes would have been a bigger change with the detailed mocks and all which I why I kept that for a separate PR. The other reason for including it was also to show case how the impl classes would be called and used.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_metadata.cc` line 39 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5hPUx7gl3f7y3NpF1:-NQ60rIU55CTJd9iJ1Xo:bpxflga) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_metadata.cc#L39)):*

Hmmm.. it seems odd to have no separator between `request.project_id()` and "job_id=".  How is the server going to know when one ends and the other starts?
</blockquote></details>

Done.

Ah..good catch. I put "," (i.e comma) as a separator but let me know if there is any other specific separator I need to put here.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 40 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jifL4jXs8voa23Lj:-NQ642zaE3er99bRoYcV:b8gd3y2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L40)):*

As I recall, your `BuildRestRequest()` sets the endpoint, but now you are using a stub with a fixed endpoint.
</blockquote></details>

It sets a default endpoint if none is provided. In this flow the endpoint is provided by the stub. At the time of writing the BuildRestRequest() method, I didn't want to make assumptions regarding how the method would be called and added the validation check and setting of the default endpoint. I can remove that if it seems redundant now. 

Just to note that options will still need to be passed to the BuildRestRequest() method so it can retrieve the endpoint to build the request url.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 35 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jt-6DseVY3a-j0e5:-NQ67baF4y_MqRJkNbtw:bwd4stk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L35)):*

I think you can assume the endpoint is always set.
</blockquote></details>

The only reason I would have it here is to not depend on the caller setting it similar to the reason why it is being set in BuildRestRequest() method as well.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 40 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jysU6cQwlJ09LbYC:-NQ69G4M5-jaLKJ5yA8y:b-q1lrur) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L40)):*

Because you are creating a pooled rest client you *must* ensure it always issues requests to the same endpoint.  That means the `BuildRestRequest()` function may need to change.
</blockquote></details>

I changed it to use the MakeDefaultRestClient() but let me know what is recommended/best practices here and what may need to change specifically with BuildRestRequest incase we use the pooled client.

coryan (changes requested):
> Reviewed 3 of 3 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016)* status: all files reviewed, 1 unresolved discussion (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_connection.cc` line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5ioJd4DGjFYXCpoeN:-NQAk3m07zhGOJfVgqDM:b-9e3uc9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_connection.cc#L44)):*

The options passed in is merged with any missing default options inside the BigQueryJobDefaultOptions method.

Just wanted to check if there is an action item to follow up for this feedback?
</blockquote></details>

The action item is to have a single spot where all the defaults are applied.  That can happen in a future PR.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_connection_test.cc` line 25 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5j3vi2qTlkFh4hWYn:-NQAkLkI8rYMP6-Gjssn:b8fmp0z) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_connection_test.cc#L25)):*

Apologies again. The connection class implementation itself was pretty straight forward and boiler plate which is why I included it. The test classes would have been a bigger change with the detailed mocks and all which I why I kept that for a separate PR. The other reason for including it was also to show case how the impl classes would be called and used.
</blockquote></details>

I often write code and then break it down into PRs that are easier to understand but "complete" (in that the class / function / type being introduced has interface, implementation and tests).  I am not asking you to do anything in this PR.  But take advantage of git's branches and rebase to break the work into easier-to-follow PRs.  You can always send more than one PR at the same time too, if you find something small that is independent of other changes.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_metadata.cc` line 39 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5hPUx7gl3f7y3NpF1:-NQ7dekUBDZVWmBj88Jy:b-h52zi9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_metadata.cc#L39)):*

Done.

Ah..good catch. I put "," (i.e comma) as a separator but let me know if there is any other specific separator I need to put here.
</blockquote></details>

I do not know.  This is being sent to the service.  What does the service *expects* to receive?
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 40 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jifL4jXs8voa23Lj:-NQ7eWg79IIjgBgJ7AGY:b-pnhr4h) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L40)):*

It sets a default endpoint if none is provided. In this flow the endpoint is provided by the stub. At the time of writing the BuildRestRequest() method, I didn't want to make assumptions regarding how the method would be called and added the validation check and setting of the default endpoint. I can remove that if it seems redundant now. 

Just to note that options will still need to be passed to the BuildRestRequest() method so it can retrieve the endpoint to build the request url.
</blockquote></details>

See below.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 35 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jt-6DseVY3a-j0e5:-NQ7duW1AiHx9fr74poW:b-51owi6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L35)):*

The only reason I would have it here is to not depend on the caller setting it similar to the reason why it is being set in BuildRestRequest() method as well.
</blockquote></details>

I would argue that you want to set the default in a single place, and then just assume that it was set.  We can discuss separately and fix that (if needed) in a future PR, but please open a issue or bug or make a note somewhere.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 40 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jysU6cQwlJ09LbYC:-NQ7eAuGCUNd9dbv-uS0:bquheeu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L40)):*

I changed it to use the MakeDefaultRestClient() but let me know what is recommended/best practices here and what may need to change specifically with BuildRestRequest incase we use the pooled client.
</blockquote></details>

Okay, but now you are opening a connection for each RPC (whereas the PooledRestClient would reuse them).  We can optimize this later, but please start tracking these items somewhere.

jsrinnn (commented):
Addressed remaining feedback. PTAL.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-:-NQCL6ri4Y8oQxPntScS:b5ooej0)* status: 23 of 25 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_connection.cc` line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5ioJd4DGjFYXCpoeN:-NQC7fIPA3xkDtubgf6H:b-3u3b4r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_connection.cc#L44)):*

The action item is to have a single spot where all the defaults are applied.  That can happen in a future PR.
</blockquote></details>

Ack.  Then lets keep BigQueryJobDefaultOptions() as the single spot where the options are applied. 

I'll remove any default setting of options from other places. Created a cleanup issue to track this: https://github.com/googleapis/google-cloud-cpp/issues/11026
___
*[`google/cloud/bigquery/v2/minimal/internal/job_metadata.cc` line 39 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5hPUx7gl3f7y3NpF1:-NQCAOLa2aL8WAlJzv8E:barw3yy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_metadata.cc#L39)):*

I do not know.  This is being sent to the service.  What does the service *expects* to receive?
</blockquote></details>

Done.

Actually I fixed this now. `params` is a vector of strings which gets joined by `&` in the SetMetadata() method. 

Based on "x-goog-request-params" documentation the parameter name/value pair, gets joined by "&": 

https://cloud.google.com/apis/docs/system-parameters
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 40 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jifL4jXs8voa23Lj:-NQCCXxR6CXtN7iX3GdK:bj1oi9p) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L40)):*

See below.
</blockquote></details>

Ack. I have responded in the comment below. Just to clarify, in this specific class no changes need to be made. BuildRestRequest and the file mentioned below would be changed. I have added a TODO there.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 35 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jt-6DseVY3a-j0e5:-NQCCbuC7FOhZiu5KjVv:b1d4q1c) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L35)):*

I would argue that you want to set the default in a single place, and then just assume that it was set.  We can discuss separately and fix that (if needed) in a future PR, but please open a issue or bug or make a note somewhere.
</blockquote></details>

Ack. SGTM. Added a TODO here with the bug number.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 40 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jysU6cQwlJ09LbYC:-NQCILk4-Ox8YtTyraqx:b-z4n68r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/63e6dc4c4fbfdd0e0711f3a61c53fffdf2f4b55f/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L40)):*

Okay, but now you are opening a connection for each RPC (whereas the PooledRestClient would reuse them).  We can optimize this later, but please start tracking these items somewhere.
</blockquote></details>

Ack. Added a TODO with a bug number.

coryan (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jsrinnn)

google/cloud/bigquery/BUILD.bazel
```diff
@@ -109,6 +109,7 @@ cc_library(
     visibility = ["//:__pkg__"],
     deps = [
         "//:common",
+        "//:grpc_utils",
```
dbolduc, line 112: This dep shouldn't be necessary anymore, right?

google/cloud/bigquery/bigquery_rest.cmake
```diff
@@ -47,8 +49,9 @@ target_include_directories(
            $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>
            $<INSTALL_INTERFACE:include>)
 target_link_libraries(
-    google_cloud_cpp_bigquery_rest PUBLIC google-cloud-cpp::rest_internal
-                                          google-cloud-cpp::common)
+    google_cloud_cpp_bigquery_rest
+    PUBLIC google-cloud-cpp::rest_internal google-cloud-cpp::grpc_utils
```
dbolduc, line 53: ditto about `grpc_utils`

dbolduc (commented):
Feel free to remove these in a follow up PR,.

jsrinnn (commented):
I have addressed the changes.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-:-NQDI0s8Cxs7U94W_6Uw:bx5rekj)* status: 23 of 25 files reviewed, 2 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[`google/cloud/bigquery/bigquery_rest.cmake` line 53 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ2yAQMqfZVWbZG31Uv-r4-53:-NQDHxZwBYteoy1-k5KJ:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0f0321aef6439a39a78d6c5802c2306b4c363922/google/cloud/bigquery/bigquery_rest.cmake#L53)):*

ditto about `grpc_utils`
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/BUILD.bazel` line 112 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ2yAQMqfZVWbZG31Uu-r4-112:-NQDHuAg2Dm8Uh0qaQWb:b-4e9yav) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0f0321aef6439a39a78d6c5802c2306b4c363922/google/cloud/bigquery/BUILD.bazel#L112)):*

This dep shouldn't be necessary anymore, right?
</blockquote></details>

Done.

Yup not needed.

PR #11021, opened 2023-03-09
cleanup(generator): remove obsolete parameter comment substitutions

devbww:
A couple of the ad hoc substitutions we apply to the comments for method_signature parameters are no longer needed (the ones from google/pubsub/v1/schema.proto), so remove them.

How do we know they're obsolete?  By making it a fatal error for a substitution to fail to match.  This will make it easy for us to remove dead cruft.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11021)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -469,38 +469,42 @@ std::string EscapePrinterDelimiter(std::string const& text) {
   return absl::StrReplaceAll(text, {{"$", "$$"}});
 }
 
+// Apply substitutions to the comments snarfed from the proto file for
+// method_signature parameters. This is mostly for the benefit of Doxygen,
+// but is also to fix mismatched quotes, etc.
+struct ParameterCommentSubtitution {
+  absl::string_view before;
+  absl::string_view after;
+  std::size_t uses = 0;
+};
+
 auto constexpr kDialogflowCXEnvironmentIdProto1 = R"""(
  list all environments for. Format: `projects/<Project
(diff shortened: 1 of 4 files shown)


generator/internal/descriptor_utils.cc
```diff
@@ -629,53 +658,16 @@ std::string FormatApiMethodSignatureParameters(
...
-            {kDialogflowESContextIdProto, kDialogflowESContextIdCpp},
-            {kDialogflowESSessionEntityTypeDisplayNameProto,
-             kDialogflowESSessionEntityTypeDisplayNameCpp},
-            {kArtifactRegistryRepositoryNameProto,
-             kArtifactRegistryRepositoryNameCpp},
-        });
-    comment = absl::StrReplaceAll(
-        EscapePrinterDelimiter(ChompByValue(comment)),
+    auto comment = EscapePrinterDelimiter(ChompByValue(loc.leading_comments));
+    for (auto& sub : substitutions) {
```
dbolduc, line 662: Is there a performance cost to splitting this into N calls?

I think there would be N memory accesses per byte of the comment. But I dont think a comment length will exceed the L1 cache. So performance should be unaffected. Does that sound right?

dbolduc (approved):

generator/internal/descriptor_utils.cc
devbww: I haven't checked the `StrReplaceAll()` implementation, but I'm sure there is some extra cost, yes.  But then again, we probably gain more back by doing the substitutions in place, rather than returning copies.  In the end, though, I'm reckon none of that matters in the big picture.

PR #11023, opened 2023-03-10
refactor: use shared_ptr<grpc::ClientContext>

dbolduc:
Part of the work for #10618 

Use `shared_ptr<grpc::ClientContext>` in the stubs, and internal auth classes. It was simplest to do this in one big change.

Handwritten changes are broken up by library.

Luckily `unique_ptr` and `shared_ptr` are the same number of characters, so the change does not change any line numbers. The one extra addition is a `NOLINT` added in `pubsub/testing/fake_streaming_pull.cc`.

In a follow up, I will replace the `absl::make_unique<grpc::ClientContext>`s with `std::make_shared<>`. That should complete the change.

NOTE: The ABI update was necessary, because templates like `StatusOr<unique_ptr<grpc::ClientContext>>` are no longer defined. See [this failing build](https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/11023/c0cf7c68634ab81d013fb3e92619e3c9b6b1ab80/fedora-37-cmake-check-api/log-033535d0-50d9-4e5a-8211-41208758b85c.txt), excerpted below:

```
                 Severity Count
 Added Symbols   -        40
Removed Symbols  High     40
 Problems with   High     0
   Data Types    Medium   0
                 Low      0
 Problems with   High     0
    Symbols      Medium   0
                 Low      0
 Problems with   Low      0
   Constants

...

Removed Symbols  40 

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

Step #3: status_or.h
Step #3: namespace google::cloud::v2_9_0
Step #3: StatusOr<std::unique_ptr<grpc::ClientContext, std::default_delete
Step #3: <grpc::ClientContext> > >::ok ( ) const
Step #3: _ZNK6google5cloud6v2_9_08StatusOrISt10unique_ptrIN4grpc13ClientContextESt14default_deleteIS5_EEE2okEv
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11023)


coryan (approved):
> NOTE: The ABI update was necessary, because templates like `StatusOr<unique_ptr<grpc::ClientContext>>` are no longer defined.

Do we know why that is the case?  If they were part of the public API that is a problem. If they were used in the implementation, that is a false positive in the API checker reports.

dbolduc:
> Do we know why that is the case? If they were part of the public API that is a problem. If they were used in the implementation, that is a false positive in the API checker reports.

It is a false positive. The "removed types" were used internally, but they get defined in the public namespace because that's where the templates are. The ABI checker does not know to filter them out because the names do not contain `internal`.

Nothing would stop the customers from using `StatusOr<std::unique_ptr<grpc::ClientContext>>` themselves. It just doesn't come with `grpc_utils` anymore.

We can check that only internal stuff has changed. (Legacy spanner admin defines the ConnectionImpl in those files).
```
$ git diff --name-only upstream/main | grep -Ev "(tests|testing|internal)"
ci/abi-dumps/google_cloud_cpp_grpc_utils.expected.abi.dump.gz
google/cloud/spanner/database_admin_connection.cc
google/cloud/spanner/instance_admin_connection.cc
```

PR #11036, opened 2023-03-14
doc: do not mention Conan

coryan:
After trying to get Conan working for 3 months, I think we should stop mentioning it as a viable package manager.  I will continue the work to get it working, but it seems too brittle and hard to use.  Mentioning seems to be interpreted as an implicit recommendation, and I cannot say that would recommend it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11036)

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,9 +401,8 @@ run the examples included with the libraries.
   fixed directory (such as `/usr/local` or `/opt`) should consult the
   [packaging guide](/doc/packaging.md).
 - Developers who prefer using a package manager such as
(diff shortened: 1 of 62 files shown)


scotthart (approved):
> Reviewed 4 of 62 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11036)* status: 4 of 62 files reviewed, all discussions resolved

dbolduc:
Just pointing out that this change affected ~50 libraries. So we have another ~50 libraries with non-standard quickstart README's.

Issue #11027, opened 2023-03-10
BigQuery V2 Job API: Optimize to use the Pooled Rest Client


jsrinnn:
Based on feedback here:
https://reviewable.io/reviews/googleapis/google-cloud-cpp/11016#-NQ5jifL4jXs8voa23Lj

Currently the CreateDefaultBigQueryJobRestStub() implementation creates a default rest client. Optimize to create a pooled rest client.

PR #11043, opened 2023-03-15
impl(bigquery): Fixes googleapis/google-cloud-cpp#11027

jsrinnn:
This replaces default rest client with pooled rest client. I looked at a few other examples in the repo for pooled rest clients especially 

https://github.com/googleapis/google-cloud-cpp/blob/ecf4413da87637e893601ac00f406cfc66b92660/google/cloud/spanner/admin/internal/database_admin_rest_stub.cc 

and

https://github.com/googleapis/google-cloud-cpp/blob/ecf4413da87637e893601ac00f406cfc66b92660/google/cloud/spanner/admin/internal/instance_admin_rest_stub.cc

And it doesn't seem like I need to change rest request path for GetJob  but let me know if that is not the case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11043)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/job_options.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_options.cc
@@ -18,6 +18,7 @@
 #include "google/cloud/bigquery/v2/minimal/internal/job_retry_policy.h"
 #include "google/cloud/internal/populate_common_options.h"
 #include <memory>
+#include <thread>
 
 namespace google {
 namespace cloud {
@@ -26,7 +27,23 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 namespace {
 auto constexpr kBackoffScaling = 2.0;
+std::size_t constexpr kJobDefaultConnectionPoolSize = 4;
+std::size_t constexpr kJobDefaultConnectionPoolSizeMax = 64;
 }  // namespace
+
+std::size_t DefaultConnectionPoolSize() {
+  // For better resource utilization and greater throughput, it is recommended
+  // to calculate the default pool size based on cores(CPU) available. However,
+  // as per C++11 documentation `std::thread::hardware_concurrency()` cannot be
+  // fully relied upon. It is only a hint and the value can be 0 if it is not
+  // well defined or not computable. Apart from CPU count, multiple channels
+  // can be opened for each CPU to increase throughput. The pool size is also
+  // capped so that servers with many cores do not create too many channels.
+  int cpu_count = std::thread::hardware_concurrency();
+  if (cpu_count == 0) return kJobDefaultConnectionPoolSize;
+  return (std::min)(kJobDefaultConnectionPoolSizeMax,
+                    cpu_count * kJobDefaultConnectionPoolSize);
+}
 
 Options BigQueryJobDefaultOptions(Options options) {
   options = google::cloud::internal::PopulateCommonOptions(
@@ -46,6 +63,10 @@ Options BigQueryJobDefaultOptions(Options options) {
   if (!options.has<BigQueryJobIdempotencyPolicyOption>()) {
     options.set<BigQueryJobIdempotencyPolicyOption>(
         MakeDefaultBigQueryJobIdempotencyPolicy());
+  }
+  if (!options.has<BigQueryJobConnectionPoolSizeOption>()) {
+    options.set<BigQueryJobConnectionPoolSizeOption>(
+        DefaultConnectionPoolSize());
   }
 
   return options;
--- a/google/cloud/bigquery/v2/minimal/internal/job_options.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_options.cc
@@ -27,9 +27,8 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 namespace {
 auto constexpr kBackoffScaling = 2.0;
-std::size_t constexpr kJobDefaultConnectionPoolSize = 4;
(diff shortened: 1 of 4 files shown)


google/cloud/bigquery/v2/minimal/internal/job_options_test.cc
```diff
@@ -48,6 +48,9 @@ TEST(JobOptionstTest, DefaultOptions) {
             expected_retry->IsExhausted());
 
   EXPECT_TRUE(actual.has<BigQueryJobBackoffPolicyOption>());
+
+  EXPECT_TRUE(actual.has<BigQueryJobConnectionPoolSizeOption>());
+  EXPECT_TRUE(actual.get<BigQueryJobConnectionPoolSizeOption>() > 0);
```
devbww, line 53: `EXPECT_GT`?

google/cloud/bigquery/v2/minimal/internal/job_options.cc
```diff
@@ -26,8 +27,24 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 namespace {
 auto constexpr kBackoffScaling = 2.0;
+std::size_t constexpr kJobDefaultConnectionPoolSize = 4;
+std::size_t constexpr kJobDefaultConnectionPoolSizeMax = 64;
 }  // namespace
 
+std::size_t DefaultConnectionPoolSize() {
```
devbww, line 34: It looks like this helper should be within the anonymous namespace.

devbww, line 34: It is a little weird to have a `kFoo` and then a function `Foo()` that doesn't just return `kFoo`.

That is, if the function is `DefaultConnectionPoolSize()`, then perhaps its default value should be `kDefaultDefaultConnectionPoolSize`.

And from there you might choose to remove one `Default` from each name.

devbww, line 34: If the function isn't named `JobDefaultConnectionPoolSize()`, I'd probably remove the "Job" from the constant names too.

jsrinnn (commented):
Addressed all feedback. PTAL.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11043#-:-NQb03Xm5TiEQREwCa6e:b-ik5xe1)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_options.cc` line 34 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11043#-NQam4NNHbykHMU4UgKy-r1-34:-NQb-_sg1xaFIQNyoQYR:bryosuu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c26c9863215815c8d67c43488da720d379d94f28/google/cloud/bigquery/v2/minimal/internal/job_options.cc#L34)):*

If the function isn't named `JobDefaultConnectionPoolSize()`, I'd probably remove the "Job" from the constant names too.
</blockquote></details>

Done.

I think I have modified to address all that is mentioned above, but pleast let me know if I have missed or misunderstood any feedback.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_options_test.cc` line 53 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11043#-NQam4NNHbykHMU4UgL--r1-53:-NQb00Tl1sdAA2_-mqy6:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c26c9863215815c8d67c43488da720d379d94f28/google/cloud/bigquery/v2/minimal/internal/job_options_test.cc#L53)):*

`EXPECT_GT`?
</blockquote></details>

Done.

google/cloud/bigquery/v2/minimal/internal/job_options_test.cc
```diff
@@ -48,6 +48,9 @@ TEST(JobOptionstTest, DefaultOptions) {
             expected_retry->IsExhausted());
 
   EXPECT_TRUE(actual.has<BigQueryJobBackoffPolicyOption>());
+
+  EXPECT_TRUE(actual.has<BigQueryJobConnectionPoolSizeOption>());
+  EXPECT_GT(actual.get<BigQueryJobConnectionPoolSizeOption>(), 0);
```
devbww, line 53: Aside: Note that if the first expectation fails, the second will too (because `std::size_t()` is 0).  We usually try to avoid such cascading errors, but not a big deal.

devbww (approved):

jsrinnn (commented):
Addressed remaining feedback.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11043#-:-NQeo9Yb92LxfozLW1bY:b-ppgltf)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_options_test.cc` line 53 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11043#-NQam4NNHbykHMU4UgL--r2-53:-NQecY7D1r0pEeiZa-ym:b3eecy7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/83b7a2aaa63d191eba9a2474a41766d7de3a7a77/google/cloud/bigquery/v2/minimal/internal/job_options_test.cc#L53)):*

Aside: Note that if the first expectation fails, the second will too (because `std::size_t()` is 0).  We usually try to avoid such cascading errors, but not a big deal.
</blockquote></details>

Done.

I removed the first one as it seemed redundant in light of the second expectation.

PR #11049, opened 2023-03-15
impl(otel): LRO tracing in stub

dbolduc:
Part of the work for #10620 

Almost identical to #11044. This time we are doing the async stub calls involved in LROs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11049)

### DIFF
--- a/generator/internal/tracing_stub_generator.cc
+++ b/generator/internal/tracing_stub_generator.cc
@@ -164,7 +164,13 @@ $tracing_stub_class_name$::Async$method_name$(
       google::cloud::CompletionQueue& cq,
       std::shared_ptr<grpc::ClientContext> context,
       $request_type$ const& request) {
-  return child_->Async$method_name$(cq, std::move(context), request);
+  auto span = internal::MakeSpanGrpc("$grpc_service$", "$method_name$");
+  {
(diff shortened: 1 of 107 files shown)


generator/internal/tracing_stub_generator.cc
```diff
@@ -241,14 +247,28 @@ future<StatusOr<google::longrunning::Operation>>
     google::cloud::CompletionQueue& cq,
     std::shared_ptr<grpc::ClientContext> context,
     google::longrunning::GetOperationRequest const& request) {
-  return child_->AsyncGetOperation(cq, std::move(context), request);
+  auto span =
+      internal::MakeSpanGrpc("google.longrunning.Operations", "GetOperation");
```
devbww, line 251: Opt: This and the next `auto span = ...` assignments are reflowed appropriately when formatting the generated libraries, but would it be nice to make them similarly wrapped within these templates?

devbww (approved):

generator/internal/tracing_stub_generator.cc
dbolduc: I did notice this. I think I'd rather get out of the business of thinking about how to format the code, even when it's in these raw strings.

PR #11050, opened 2023-03-15
feat(mocks): provide access to call options in client tests

### LINKED ISSUE #10030: Customers can verify call options
Our APIs look like:
```cc
T Client::Foo(FooRequest const& request, Options options);
T Connection::Foo(FooReqeust const& request);
```

Customers can verify they are giving the right `request` to our library. They cannot verify that they are giving the right `options` to our library.

```cc
void MakeClientCall() {
  // This function can be unit tested by customers
  auto options = MakeCallOptions();

  // But they can never catch the difference between:
  client.Foo(req, options);
  // and...
  client.Foo(req); // oops I forgot to include the options.
}
```

I think the solution is to add:

```cc
namespace google::cloud::mocks {
/**
 * Verify per-call options
 * 
 * @code
 * TEST(Foo, CallOptions) {
 *   auto mock = std::make_shared<MockConnection>();
 *   EXPECT_CALL(*mock, Foo).WillOnce([] {
 *         auto options = google::cloud::mocks::GetCallOptions();
 *         EXPECT_THAT(options, ...);
 *       });
 *   auto client = Client(mock);
 *   DoThing(client);
 * }
 * @endcode
 */
Options GetCallOptions() {
  return internal::CurrentOptions();
}
}
```

dbolduc:
Fixes #10030 

If we think this should be more publicized, say the words and I will add it to our [mocking exemplar](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/bigquery/samples/mock_bigquery_read.cc).

Also, while we are here, put the other mock test in an anonymous namespace.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11050)

### DIFF
--- a/google/cloud/mocks/current_options.h
+++ b/google/cloud/mocks/current_options.h
@@ -41,7 +41,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
  * }
  * @endcode
  */
-Options CurrentOptions() { return internal::CurrentOptions(); }
+inline Options const& CurrentOptions() { return internal::CurrentOptions(); }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace mocks
--- a/google/cloud/mocks/current_options.h
+++ b/google/cloud/mocks/current_options.h
@@ -1,0 +1,51 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_MOCKS_CURRENT_OPTIONS_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_MOCKS_CURRENT_OPTIONS_H
+
+#include "google/cloud/options.h"
+
+namespace google {
+namespace cloud {
+namespace mocks {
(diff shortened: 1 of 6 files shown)


google/cloud/mocks/current_options.h
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * This would be used to verify configuration options from within a
+ * MockConnection. It provides a way for applications to test the difference
+ * between `client.Foo(request, options)` and `client.Foo(request)`.
+ *
+ * @code
+ * TEST(Foo, CallOptions) {
+ *   auto mock = std::make_shared<MockConnection>();
+ *   EXPECT_CALL(*mock, Foo).WillOnce([] {
+ *         auto const& options = google::cloud::mocks::CurrentOptions();
```
devbww, line 36: If `CurrentOptions()` is going to return by value, then I wouldn't suggest capturing that value by `const&`.

dbolduc: changed to `Options const& mocks::CurrentOptions()`

google/cloud/mocks/current_options.h
```diff
@@ -0,0 +1,51 @@
...
+ *         EXPECT_THAT(options, ...);
+ *       });
+ *   auto client = Client(mock);
+ *   DoThing(client);
```
devbww, line 40: Opt: Perhaps `client->Foo()` would be clearer?

devbww (approved):

google/cloud/mocks/current_options.h
dbolduc: It's not quite the same thing. If a customer's test called Foo, they would just verify the options outside of the mock.

```cc
client.Foo(options);
EXPECT_THAT(options, ...);
```

We are helping them unit test their `DoThing` method which could look like:

```cc
void DoThing(Client c) {
  auto options = ...;
  c.Foo(options); // `c.Foo();` would be wrong, but how can I detect the difference?
}
```

That being said, I agree that the link between `DoThing` and `Foo` is unclear / imaginary. I will `s/DoThing/CallFoo/` to make the link clearer.

dbolduc: decided on: `s/DoThing/MyFunctionThatCallsFoo/`

devbww: SGTM.  Thanks.

PR #11066, opened 2023-03-20
impl(otel): async unary + LRO connection tracing

dbolduc:
Part of the work for #10620

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11066)

### DIFF
--- a/generator/internal/tracing_connection_generator.cc
+++ b/generator/internal/tracing_connection_generator.cc
@@ -284,7 +284,10 @@ future<Status>)"""
 future<StatusOr<$longrunning_deduced_response_type$>>)""",
         R"""(
 $tracing_connection_class_name$::$method_name$($request_type$ const& request) {
-  return child_->$method_name$(request);
+  auto span = internal::MakeSpan(
+      "$product_namespace$::$connection_class_name$::$method_name$");
+  auto scope = opentelemetry::trace::Scope(span);
+  return internal::EndSpan(std::move(span), child_->$method_name$(request));
 }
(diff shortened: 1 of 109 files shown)


generator/internal/tracing_connection_generator.cc
```diff
@@ -284,7 +284,10 @@ future<Status>)"""
 future<StatusOr<$longrunning_deduced_response_type$>>)""",
         R"""(
 $tracing_connection_class_name$::$method_name$($request_type$ const& request) {
-  return child_->$method_name$(request);
+  auto span = internal::MakeSpan(
+      "$product_namespace$::$connection_class_name$::$method_name$");
```
coryan, line 288: Should these be named after the C++ function or the proto method name?

coryan (approved):

generator/internal/tracing_connection_generator.cc
dbolduc: That is a good question, but I am going to merge this PR as-is, because naming is something we can change later. Arguably, I should be using the client name instead of the connection name too.

Note that the underlying gRPC calls use the proto method name.

My thinking at the moment is that it will be easier for customers to look up spans if it matches the C++ methods they are calling. (Which is more relevant for handwritten clients like `bigtable::Table::ReadRows`).

It does mean that the naming convention for these client level spans would vary across languages. But our client names already vary, so I am not sure it is a new problem.

I do not have strongly held opinions here.

PR #11068, opened 2023-03-20
impl(bigquery): ListJobs api - Response object parsing, implementation and unit tests

jsrinnn:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/job.h
+++ b/google/cloud/bigquery/v2/minimal/internal/job.h
@@ -42,7 +42,6 @@ struct JobReference {
   std::string location;
 };
 
-// Custom object for V2 Job proto fields.
 struct Job {
   std::string kind;
   std::string etag;
@@ -55,6 +54,20 @@ struct Job {
   JobConfiguration configuration;
 };
 
+struct ListFormatJob {
+  std::string id;
+  std::string kind;
+  std::string user_email;
+  std::string state;
+  std::string principal_subject;
+
+  JobReference reference;
+  JobConfiguration configuration;
+  JobStatus status;
+
+  ErrorProto error_result;
+};
+
 NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(JobStatus, error_result, errors,
                                                 state);
 NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(JobReference, project_id,
@@ -62,6 +75,11 @@ NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(JobReference, project_id,
 NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(Job, kind, etag, id, self_link,
                                                 user_email, status, reference,
                                                 configuration);
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(ListFormatJob, id, kind,
+                                                user_email, state,
+                                                principal_subject, reference,
+                                                configuration, status,
+                                                error_result);
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace bigquery_v2_minimal_internal
--- a/google/cloud/bigquery/v2/minimal/internal/job_response.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_response.cc
@@ -26,26 +26,74 @@ bool valid_job(nlohmann::json const& j) {
           j.contains("configuration"));
 }
 
-StatusOr<GetJobResponse> GetJobResponse::BuildFromHttpResponse(
-    BigQueryHttpResponse const& http_response) {
-  if (http_response.payload.empty()) {
+bool valid_list_format_job(nlohmann::json const& j) {
+  return (j.contains("kind") && j.contains("state") && j.contains("id") &&
+          j.contains("reference") && j.contains("configuration"));
+}
+
+bool valid_jobs_list(nlohmann::json const& j) {
+  return (j.contains("kind") && j.contains("etag") &&
+          j.contains("next_page_token") && j.contains("jobs"));
+}
+
+StatusOr<nlohmann::json> parse_json(std::string const& payload) {
+  if (payload.empty()) {
     return internal::InternalError("Empty payload in HTTP response",
                                    GCP_ERROR_INFO());
   }
   // Build the job response object from Http response.
-  auto json_obj = nlohmann::json::parse(http_response.payload, nullptr, false);
+  auto json_obj = nlohmann::json::parse(payload, nullptr, false);
   if (!json_obj.is_object()) {
     return internal::InternalError("Error parsing Json from response payload",
                                    GCP_ERROR_INFO());
   }
-  if (!valid_job(json_obj)) {
+
+  return json_obj;
+}
+
+StatusOr<GetJobResponse> GetJobResponse::BuildFromHttpResponse(
+    BigQueryHttpResponse const& http_response) {
+  auto json = parse_json(http_response.payload);
+  if (!json) return std::move(json).status();
+
+  if (!valid_job(*json)) {
     return internal::InternalError("Not a valid Json Job object",
                                    GCP_ERROR_INFO());
   }
 
   GetJobResponse result;
-  result.job = json_obj.get<Job>();
+  result.job = json->get<Job>();
   result.http_response = http_response;
+
+  return result;
+}
+
+StatusOr<ListJobsResponse> ListJobsResponse::BuildFromHttpResponse(
+    BigQueryHttpResponse const& http_response) {
+  auto json = parse_json(http_response.payload);
+  if (!json) return std::move(json).status();
+
+  if (!valid_jobs_list(*json)) {
+    return internal::InternalError("Not a valid Json JobList object",
+                                   GCP_ERROR_INFO());
+  }
+
+  ListJobsResponse result;
+  result.http_response = http_response;
+
+  result.kind = json->value("kind", "");
+  result.etag = json->value("etag", "");
+  result.next_page_token = json->value("next_page_token", "");
+
+  for (auto const& kv : json->at("jobs").items()) {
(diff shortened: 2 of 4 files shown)


coryan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068)* status: all files reviewed, 4 unresolved discussions (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_response.cc` line 30 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3BuVk5PcklhF4mUir:-NR3BuVk5PcklhF4mUis:b-tunna0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response.cc#L30)):*
> ```cpp
> 
> bool valid_list_format_job(nlohmann::json const& j) {
>   return (j.contains("kind") && j.contains("state") && j.contains("id") &&
> ```

FYI: there are many more ways for a JSON object to be invalid.  For example, "kind" cloud be an array of numbers instead of a string, or "state" could be a boolean.  The mere presence of a field does not guarantee that it is correct.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response.cc` line 51 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3CGg-3OqsyVhXmCGl:-NR3CGg-3OqsyVhXmCGm:byb77kb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response.cc#L51)):*
> ```cpp
>   }
> 
>   return json_obj;
> ```

nit: `json_obj` reads as "JavaScript Object Notation Obj"... Myabe just `json`?
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response_test.cc` line 40 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3ChLY0WeWhnAn0zfL:-NR3ChLY0WeWhnAn0zfM:b-nyy68q) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response_test.cc#L40)):*
> ```cpp
>           "reference": {},
>           "configuration": {}})";
>   auto const& job_response =
> ```

Why introduce a `const&` here?  I think the previous definition was quite good.  If you need a rule of thumb:

- Most of the time, `auto const foo = Function()` is the right call.
  - If the function returns a value, the compiler will optimize the assignment / copy away: https://en.cppreference.com/w/cpp/language/copy_elision
  - If the function returns a reference, then the copy has a longer lifetime and requires less reasoning about the lifetime of the source of the reference
- Some exceptions include:
  - If the function returns a pointer (not a reference) then `auto const* foo` may be more obvious to the reader
  - If the function is called in a loop, **and** it returns a reference, then maybe `auto const& foo` is more efficient and may be justified.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response_test.cc` line 149 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3EA2A9VzwiyOkI5Qk:-NR3EA2BBum3Z7FcxOk2:b1s70ta) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response_test.cc#L149)):*
> ```cpp
> 
>   auto const& jobs = list_jobs_response->jobs;
>   EXPECT_EQ(jobs.size(), 1);
> ```

FYI: If `ListFormatJob` had an `operator==` and a `operator<<` you could write:

```cc
EXPECT_THAT(jobs, ElementsAre(expected_job));
```

or maybe you can write some matchers and then write:

```cc
EXPECT_THAT(jobs, ElementsAre(MyMatcher(expected_job)));
```

No changes required in this PR, just start thinking about building infrastructure to make your tests easier, because (I am guessing) these types will get more and more elaborate.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response_test.cc` line 149 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3EN6A87o2q4TSPEp7:-NR3EN6A87o2q4TSPEp8:b-i52war) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response_test.cc#L149)):*
> ```cpp
> 
>   auto const& jobs = list_jobs_response->jobs;
>   EXPECT_EQ(jobs.size(), 1);
> ```

This should be `ASSERT_EQ(jobs.size(), 1)` because your test will crash if `.size()` is 0.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response_test.cc` line 199 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3F0D5-xcZ3MvqzexJ:-NR3F0D5-xcZ3MvqzexK:b-9etmez) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response_test.cc#L199)):*
> ```cpp
>               }
>   ]})";
>   auto const& response = ListJobsResponse::BuildFromHttpResponse(http_response);
> ```

Another unnecessary `const&`.

jsrinnn (commented):
Addressed code review feedback. Thanks!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-:-NR3rL6d5crdKzbrvYwH:b-8e44jq)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_response.cc` line 30 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3BuVk5PcklhF4mUir:-NR3n5sm7nWL6HRXoHjy:bukc616) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response.cc#L30)):*

FYI: there are many more ways for a JSON object to be invalid.  For example, "kind" cloud be an array of numbers instead of a string, or "state" could be a boolean.  The mere presence of a field does not guarantee that it is correct.
</blockquote></details>

Ack. 

1) This function is mainly doing basic validation that the expected required fields for the ListFormatJobs are present.

 2) As observed from tests, any json invalid syntax is caught by the parse() method().

3) Any field type errors gets caught when I do json.get<ListFormatJob> if the json does not confirm to the field types.

I do see your point and that additional validation can be added.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response.cc` line 51 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3CGg-3OqsyVhXmCGl:-NR3oT5U7iHky-mddeba:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response.cc#L51)):*

nit: `json_obj` reads as "JavaScript Object Notation Obj"... Myabe just `json`?
</blockquote></details>

Done
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response_test.cc` line 40 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3ChLY0WeWhnAn0zfL:-NR3ogCO3N6DxeyMQOGF:b-gepic) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response_test.cc#L40)):*

Why introduce a `const&` here?  I think the previous definition was quite good.  If you need a rule of thumb:

- Most of the time, `auto const foo = Function()` is the right call.
  - If the function returns a value, the compiler will optimize the assignment / copy away: https://en.cppreference.com/w/cpp/language/copy_elision
  - If the function returns a reference, then the copy has a longer lifetime and requires less reasoning about the lifetime of the source of the reference
- Some exceptions include:
  - If the function returns a pointer (not a reference) then `auto const* foo` may be more obvious to the reader
  - If the function is called in a loop, **and** it returns a reference, then maybe `auto const& foo` is more efficient and may be justified.


</blockquote></details>

Done.

I got the initial impression that the `auto const&` was the optimized way to go for using auto. I have modified it to `auto const foo` based on above.

Thank You for the detailed explanation.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response_test.cc` line 149 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3EN6A87o2q4TSPEp7:-NR3rCm567Q4XhuN2KD_:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response_test.cc#L149)):*

This should be `ASSERT_EQ(jobs.size(), 1)` because your test will crash if `.size()` is 0.
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_response_test.cc` line 199 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11068#-NR3F0D5-xcZ3MvqzexJ:-NR3rEbw5DTHwHegGySA:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4e79e14718d4062dcc65d3f5fe7476f1ec7c043e/google/cloud/bigquery/v2/minimal/internal/job_response_test.cc#L199)):*

Another unnecessary `const&`.
</blockquote></details>

Done.

PR #11070, opened 2023-03-20
feat(GCS+gRPC): more efficient CRC32C computations

### LINKED ISSUE #11060: Optimize CRC32C computation in GCS+gRPC
## Objective

Improve the customer experience using GCS+gRPC by reducing CPU overhead

## Background

GCS uses CRC32C to provide data integrity in object uploads.  The upload can include the CRC32C checksums and GCS will reject the upload if the data does not match the checksum.  In GCS+gRPC data is uploaded in `Write()` messages, each of at most 2MiB. GCS+gRPC strengthens the end-to-end data integrity by adding CRC32C checksums for each `Write()` message.

The current implementation of the GCS+gRPC computes the checksums for the messages and separately computes the checksums for the full object. It is possible to optimize this computation.  One can concatenate the CRC32C checksums of each `Write()` message to compute the end-to-end checksum of the full object.  This can save about 1/2 of the CPU time spent on checksums.

The end-to-end CRC32C computation is (currently) performed in the `ObjectWriteStream` and `ObjectWriteStreambuf` objects. The lifetime of these objects is at least as long as the lifetime of the upload (ignoring resumed uploads for a minute).  These objects can also process the data **before** any retry loop. We do not want to checksum bytes that are sent only because an upload was needed. 

Unfortunately these objects do not have access to the `Write()` messages in gRPC: they work for both REST and gRPC.

## Overview

We will move the computation of the end-to-end checksums from the `ObjectWriteStreambuf` objects to the `UploadChunk()` function in the `RawClient` layer.  In the REST case only the end-to-end checksum is updated as the data is sent.  In the gRPC case the per-`Write()` message CRC32C is computed and this checksum is efficiently concatenated to the end-to-end checksum.

## Details

We need to change the classes to compute checksums in several ways:

- They may be called multiple times for the same bytes when retrying an upload. We will introduce a "current offset" and ignore data for older offsets.
- They need to be able to receive a pre-computed CRC32C checksum with a new buffer, and use the more efficient algorithm in this case.
- They need to be able to handle data as `absl::string_view`, and as `absl::Cord const&` as the Protobufs may use either representation for the `Write()` message.
- The final hashes may be called multiple times because the "final" `Write()` message may need to be resent.

## Alternatives

I considered returning the `Write()` CRC32C checksum as part of the `UploadChunk()` response.  This response would be empty (maybe an `absl::optional<>`) for REST and if present would be used to save some computation.  This does not work with retries and partial retries, where the computation may only include

coryan:
Move the CRC32C and MD5 computations to the `GrpcClient` and `RestClient`. In the case of gRPC this allows us to compute the CRC32C checksum for each `Write()` message and then concatenate this checksum with the CRC32C checksum computed for the full object.  That saves one of the most expensive computations in the gRPC uploads.

Fixes #11060

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11070)

### DIFF
--- a/google/cloud/storage/internal/object_write_streambuf.cc
+++ b/google/cloud/storage/internal/object_write_streambuf.cc
@@ -142,25 +142,26 @@ void ObjectWriteStreambuf::FlushFinal() {
 
   // Calculate the portion of the buffer that needs to be uploaded, if any.
   auto const actual_size = put_area_size();
-  hash_function_->Update(absl::string_view{pbase(), actual_size});
 
   // After this point the session will be closed, and no more calls to the hash
   // function are possible.
-  auto function = std::move(hash_function_);
-  hash_values_ = std::move(*function).Finish();
   auto upload_request = UploadChunkRequest(upload_id_, committed_size_,
                                            {ConstBuffer(pbase(), actual_size)},
-                                           Merge(known_hashes_, hash_values_));
+                                           hash_function_, known_hashes_);
   request_.ForEachOption(internal::CopyCommonOptions(upload_request));
   OptionsSpan const span(span_options_);
   auto response = client_->UploadChunk(upload_request);
   if (!response) {
     last_status_ = std::move(response).status();
-  } else {
-    committed_size_ = response->committed_size.value_or(0);
-    metadata_ = std::move(response->payload);
-    headers_ = std::move(response->request_metadata);
+    return;
   }
+
+  auto function = std::move(hash_function_);
+  hash_values_ = std::move(*function).Finish();
+
+  committed_size_ = response->committed_size.value_or(0);
+  metadata_ = std::move(response->payload);
+  headers_ = std::move(response->request_metadata);
 
   // Reset the iostream put area with valid pointers, but empty.
   current_ios_buffer_.resize(1);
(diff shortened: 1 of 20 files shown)


google/cloud/storage/internal/object_write_streambuf.cc
```diff
@@ -142,26 +142,27 @@ void ObjectWriteStreambuf::FlushFinal() {
...
   request_.ForEachOption(internal::CopyCommonOptions(upload_request));
   OptionsSpan const span(span_options_);
   auto response = client_->UploadChunk(upload_request);
   if (!response) {
     last_status_ = std::move(response).status();
-  } else {
-    committed_size_ = response->committed_size.value_or(0);
-    metadata_ = std::move(response->payload);
-    headers_ = std::move(response->request_metadata);
+    return;
```
devbww, line 156: Nit: Change `if (!B) S1; else return; S2;` to `if (B) return; S1; S2;`.

google/cloud/storage/internal/object_write_streambuf.cc
```diff
@@ -212,34 +209,35 @@ void ObjectWriteStreambuf::FlushRoundChunk(ConstBufferSequence buffers) {
...
-    // Reset the internal buffer and copy any trailing bytes from `buffers` to
-    // it.
-    auto* pbeg = current_ios_buffer_.data();
-    setp(pbeg, pbeg + current_ios_buffer_.size());
-    PopFrontBytes(buffers, rounded_size);
-    for (auto const& b : buffers) {
-      std::copy(b.begin(), b.end(), pptr());
-      pbump(static_cast<int>(b.size()));
-    }
+    return;
```
devbww, line 212: Ditto.

devbww (approved):

google/cloud/storage/internal/object_write_streambuf.cc
coryan: I am not sure I follow, the new code reads:

```cc
if (!B) {
  S1;
  return;
}

S2;
S3;
S4
```

Something like `if (B) return;` is not semantically equivalent.  I would need to write:

```cc
if (B) {
  S2;
  S3;
  S4;
  return;
}

S1;
```

which seems less readable to me.

google/cloud/storage/internal/object_write_streambuf.cc
coryan: Ditto++.  I am going to merge as-is.  If I misread your comments we can fix the code in a followup PR.

google/cloud/storage/internal/object_write_streambuf.cc
devbww: Sorry ... it seems I can't read a diff correctly.

PR #11081, opened 2023-03-22
impl(bigquery): ListJobs api - Added ListJobs() support to BigQueryRestStub

jsrinnn:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
@@ -32,21 +32,16 @@ template <typename Functor, typename Request,
 Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
                      Request const& request, char const* where,
                      TracingOptions const& /*options*/) {
-  GCP_LOG(DEBUG) << where << "() << " << request;
-
-  if (!context.headers().empty()) {
-    GCP_LOG(DEBUG) << where << "() << RestContext{";
-    for (auto const& h : context.headers()) {
-      GCP_LOG(DEBUG) << "[header_name=" << h.first << ", header_value={"
-                     << absl::StrJoin(h.second, "&") << "}],";
-    }
-    GCP_LOG(DEBUG) << "}";
-  }
+  auto formatter = [](std::string* out, auto const& header) {
+    absl::StrAppend(out, "name=", header.first, ", value={",
+                    absl::StrJoin(header.second, "&"), "}");
+  };
+  GCP_LOG(DEBUG) << where << "() " << request << ", Context={"
+                 << absl::StrJoin(context.headers(), ", ", formatter) << "}";
 
   auto response = functor(context, request);
-  if (!response) {
-    GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
-  }
+  GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
+
   return response;
 }
 
--- a/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
@@ -57,6 +57,33 @@ StatusOr<GetJobResponse> BigQueryJobLogging::GetJob(
   return response;
 }
 
+StatusOr<ListJobsResponse> BigQueryJobLogging::ListJobs(
+    rest_internal::RestContext& rest_context, ListJobsRequest const& request) {
+  char const* context = __func__;
+  GCP_LOG(DEBUG) << context
+                 << "() << ListJobsRequest{project_id=" << request.project_id()
+                 << ", all_users=" << request.all_users()
+                 << ", max_results=" << request.max_results()
+                 << ", page_token=" << request.page_token()
+                 << ", projection=" << request.projection().value
+                 << ", state_filter=" << request.state_filter().value
+                 << ", parent_job_id=" << request.parent_job_id() << "}";
+
+  if (!rest_context.headers().empty()) {
+    GCP_LOG(DEBUG) << context << "() << RestContext{";
+    for (auto const& h : rest_context.headers()) {
+      GCP_LOG(DEBUG) << "[header_name=" << h.first << ", header_value={"
+                     << absl::StrJoin(h.second, "&") << "}],";
+    }
+    GCP_LOG(DEBUG) << "}";
+  }
+  auto response = child_->ListJobs(rest_context, request);
+  if (!response.ok()) {
+    GCP_LOG(DEBUG) << context << "() >> status={" << response.status() << "}";
+  }
+  return response;
+}
+
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace bigquery_v2_minimal_internal
 }  // namespace cloud
--- a/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
@@ -16,6 +16,7 @@
 
 #include "google/cloud/bigquery/v2/minimal/internal/job_logging.h"
 #include "google/cloud/internal/absl_str_join_quiet.h"
+#include "google/cloud/internal/invoke_result.h"
 #include "google/cloud/internal/rest_context.h"
 #include "google/cloud/log.h"
 #include "google/cloud/status_or.h"
@@ -24,6 +25,30 @@ namespace google {
 namespace cloud {
 namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+template <typename Functor, typename Request,
+          typename Result = google::cloud::internal::invoke_result_t<
+              Functor, rest_internal::RestContext&, Request const&>>
+Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
+                     Request const& request, char const* where,
+                     TracingOptions const& /*options*/) {
+  GCP_LOG(DEBUG) << where << "() << " << request;
+
+  if (!context.headers().empty()) {
+    GCP_LOG(DEBUG) << where << "() << RestContext{";
+    for (auto const& h : context.headers()) {
+      GCP_LOG(DEBUG) << "[header_name=" << h.first << ", header_value={"
+                     << absl::StrJoin(h.second, "&") << "}],";
+    }
+    GCP_LOG(DEBUG) << "}";
+  }
+
+  auto response = functor(context, request);
+  if (!response) {
+    GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
+  }
+  return response;
+}
 
 BigQueryJobLogging::BigQueryJobLogging(
     std::shared_ptr<BigQueryJobRestStub> child, TracingOptions tracing_options,
@@ -36,52 +61,22 @@ BigQueryJobLogging::BigQueryJobLogging(
 // not a protobuf message.
 StatusOr<GetJobResponse> BigQueryJobLogging::GetJob(
     rest_internal::RestContext& rest_context, GetJobRequest const& request) {
-  char const* context = __func__;
-  GCP_LOG(DEBUG) << context
-                 << "() << GetJobRequest{project_id=" << request.project_id()
-                 << ",job_id=" << request.job_id()
-                 << ",location=" << request.location() << "}";
-
-  if (!rest_context.headers().empty()) {
-    GCP_LOG(DEBUG) << context << "() << RestContext{";
-    for (auto const& h : rest_context.headers()) {
-      GCP_LOG(DEBUG) << "[header_name=" << h.first << ", header_value={"
-                     << absl::StrJoin(h.second, "&") << "}],";
-    }
-    GCP_LOG(DEBUG) << "}";
-  }
-  auto response = child_->GetJob(rest_context, request);
-  if (!response.ok()) {
-    GCP_LOG(DEBUG) << context << "() >> status={" << response.status() << "}";
-  }
-  return response;
+  return JobLogWrapper(
+      [this](rest_internal::RestContext& rest_context,
+             GetJobRequest const& request) {
+        return child_->GetJob(rest_context, request);
+      },
+      rest_context, request, __func__, tracing_options_);
 }
 
 StatusOr<ListJobsResponse> BigQueryJobLogging::ListJobs(
     rest_internal::RestContext& rest_context, ListJobsRequest const& request) {
-  char const* context = __func__;
-  GCP_LOG(DEBUG) << context
-                 << "() << ListJobsRequest{project_id=" << request.project_id()
-                 << ", all_users=" << request.all_users()
-                 << ", max_results=" << request.max_results()
-                 << ", page_token=" << request.page_token()
-                 << ", projection=" << request.projection().value
-                 << ", state_filter=" << request.state_filter().value
-                 << ", parent_job_id=" << request.parent_job_id() << "}";
-
-  if (!rest_context.headers().empty()) {
-    GCP_LOG(DEBUG) << context << "() << RestContext{";
-    for (auto const& h : rest_context.headers()) {
-      GCP_LOG(DEBUG) << "[header_name=" << h.first << ", header_value={"
-                     << absl::StrJoin(h.second, "&") << "}],";
-    }
-    GCP_LOG(DEBUG) << "}";
-  }
-  auto response = child_->ListJobs(rest_context, request);
-  if (!response.ok()) {
-    GCP_LOG(DEBUG) << context << "() >> status={" << response.status() << "}";
-  }
-  return response;
+  return JobLogWrapper(
+      [this](rest_internal::RestContext& rest_context,
+             ListJobsRequest const& request) {
+        return child_->ListJobs(rest_context, request);
+      },
+      rest_context, request, __func__, tracing_options_);
 }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
--- a/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
@@ -16,6 +16,7 @@
 
 #include "google/cloud/bigquery/v2/minimal/internal/job_logging.h"
 #include "google/cloud/internal/absl_str_join_quiet.h"
+#include "google/cloud/internal/debug_string.h"
 #include "google/cloud/internal/invoke_result.h"
 #include "google/cloud/internal/rest_context.h"
 #include "google/cloud/log.h"
@@ -31,13 +32,16 @@ template <typename Functor, typename Request,
               Functor, rest_internal::RestContext&, Request const&>>
 Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
                      Request const& request, char const* where,
-                     TracingOptions const& /*options*/) {
+                     TracingOptions const& options) {
   auto formatter = [](std::string* out, auto const& header) {
     absl::StrAppend(out, "name=", header.first, ", value={",
                     absl::StrJoin(header.second, "&"), "}");
   };
   GCP_LOG(DEBUG) << where << "() " << request << ", Context={"
-                 << absl::StrJoin(context.headers(), ", ", formatter) << "}";
+                 << internal::DebugString(
+                        absl::StrJoin(context.headers(), ", ", formatter),
+                        options)
+                 << "}";
 
   auto response = functor(context, request);
   GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
--- a/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc
@@ -24,20 +24,23 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 BigQueryJobRestStub::~BigQueryJobRestStub() = default;
 
-StatusOr<GetJobResponse> DefaultBigQueryJobRestStub::GetJob(
-    rest_internal::RestContext& rest_context, GetJobRequest const& request) {
-  internal::OptionsSpan span(options_);
+template <typename ReturnType>
+StatusOr<ReturnType> ParseFromRestResponse(
+    StatusOr<std::unique_ptr<rest_internal::RestResponse>> rest_response) {
+  if (!rest_response) return std::move(rest_response).status();
 
-  if (request.project_id().empty()) {
-    return internal::InvalidArgumentError(
-        "Invalid GetJobRequest: Project Id is empty", GCP_ERROR_INFO());
-  }
-  if (request.job_id().empty()) {
-    return internal::InvalidArgumentError(
-        "Invalid GetJobRequest: Job Id is empty", GCP_ERROR_INFO());
-  }
-  // Prepare the RestRequest
-  auto rest_request = BuildRestRequest(request, options_);
+  auto http_response =
+      BigQueryHttpResponse::BuildFromRestResponse(std::move(*rest_response));
+  if (!http_response) return std::move(http_response).status();
+
+  return ReturnType::BuildFromHttpResponse(*http_response);
+}
+
+template <typename RequestType>
+StatusOr<rest_internal::RestRequest> PrepareRestRequest(
+    rest_internal::RestContext& rest_context, RequestType const& request,
+    Options options) {
+  auto rest_request = BuildRestRequest(request, options);
   if (!rest_request) return std::move(rest_request).status();
 
   // Copy over any headers if provided.
@@ -49,52 +52,31 @@ StatusOr<GetJobResponse> DefaultBigQueryJobRestStub::GetJob(
     }
   }
 
-  // Call the rest client to get job details from the server as a RestResponse.
-  auto rest_response = rest_stub_->Get(std::move(*rest_request));
-  if (!rest_response) return std::move(rest_response).status();
+  return rest_request;
+}
 
-  // Convert RestResponse to HttpResponse.
-  auto http_response =
-      BigQueryHttpResponse::BuildFromRestResponse(std::move(*rest_response));
-  if (!http_response) return std::move(http_response).status();
+StatusOr<GetJobResponse> DefaultBigQueryJobRestStub::GetJob(
+    rest_internal::RestContext& rest_context, GetJobRequest const& request) {
+  internal::OptionsSpan span(options_);
+  // Prepare the RestRequest from GetJobRequest.
+  auto rest_request =
+      PrepareRestRequest<GetJobRequest>(rest_context, request, options_);
 
-  // Convert HttpResponse to GetJobResponse.
-  return GetJobResponse::BuildFromHttpResponse(*http_response);
+  // Call the rest stub and parse the RestResponse.
+  return ParseFromRestResponse<GetJobResponse>(
+      rest_stub_->Get(std::move(*rest_request)));
 }
 
 StatusOr<ListJobsResponse> DefaultBigQueryJobRestStub::ListJobs(
     rest_internal::RestContext& rest_context, ListJobsRequest const& request) {
   internal::OptionsSpan span(options_);
-
-  if (request.project_id().empty()) {
-    return internal::InvalidArgumentError(
-        "Invalid ListJobsRequest: Project Id is empty", GCP_ERROR_INFO());
-  }
-
   // Prepare the RestRequest from ListJobsRequest.
-  auto rest_request = BuildRestRequest(request, options_);
-  if (!rest_request) return std::move(rest_request).status();
+  auto rest_request =
+      PrepareRestRequest<ListJobsRequest>(rest_context, request, options_);
 
-  // Copy over any headers if provided.
-  if (!rest_context.headers().empty()) {
-    for (auto const& h : rest_context.headers()) {
-      if (!h.second.empty()) {
-        rest_request->AddHeader(h.first, absl::StrJoin(h.second, "&"));
-      }
-    }
-  }
-
-  // Call the rest client to get job details from the server as a RestResponse.
-  auto rest_response = rest_stub_->Get(std::move(*rest_request));
-  if (!rest_response) return std::move(rest_response).status();
-
-  // Convert RestResponse to HttpResponse.
-  auto http_response =
-      BigQueryHttpResponse::BuildFromRestResponse(std::move(*rest_response));
-  if (!http_response) return std::move(http_response).status();
-
-  // Convert HttpResponse to ListJobsResponse.
-  return ListJobsResponse::BuildFromHttpResponse(*http_response);
+  // Call the rest stub and parse the RestResponse.
+  return ParseFromRestResponse<ListJobsResponse>(
+      rest_stub_->Get(std::move(*rest_request)));
 }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
--- a/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc
@@ -62,6 +62,41 @@ StatusOr<GetJobResponse> DefaultBigQueryJobRestStub::GetJob(
   return GetJobResponse::BuildFromHttpResponse(*http_response);
 }
 
+StatusOr<ListJobsResponse> DefaultBigQueryJobRestStub::ListJobs(
+    rest_internal::RestContext& rest_context, ListJobsRequest const& request) {
+  internal::OptionsSpan span(options_);
+
+  if (request.project_id().empty()) {
+    return internal::InvalidArgumentError(
+        "Invalid ListJobsRequest: Project Id is empty", GCP_ERROR_INFO());
+  }
+
+  // Prepare the RestRequest from ListJobsRequest.
+  auto rest_request = BuildRestRequest(request, options_);
+  if (!rest_request) return std::move(rest_request).status();
+
+  // Copy over any headers if provided.
+  if (!rest_context.headers().empty()) {
+    for (auto const& h : rest_context.headers()) {
+      if (!h.second.empty()) {
+        rest_request->AddHeader(h.first, absl::StrJoin(h.second, "&"));
+      }
+    }
+  }
+
+  // Call the rest client to get job details from the server as a RestResponse.
+  auto rest_response = rest_stub_->Get(std::move(*rest_request));
+  if (!rest_response) return std::move(rest_response).status();
+
+  // Convert RestResponse to HttpResponse.
+  auto http_response =
+      BigQueryHttpResponse::BuildFromRestResponse(std::move(*rest_response));
+  if (!http_response) return std::move(http_response).status();
+
+  // Convert HttpResponse to ListJobsResponse.
+  return ListJobsResponse::BuildFromHttpResponse(*http_response);
+}
+
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
 }  // namespace bigquery_v2_minimal_internal
 }  // namespace cloud
--- a/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc
@@ -34,9 +34,11 @@ using ::google::cloud::testing_util::MakeMockHttpPayloadSuccess;
 using ::google::cloud::testing_util::MockHttpPayload;
 using ::google::cloud::testing_util::MockRestClient;
 using ::google::cloud::testing_util::MockRestResponse;
+using ::google::cloud::testing_util::StatusIs;
 using ::testing::An;
 using ::testing::ByMove;
 using ::testing::Eq;
+using ::testing::HasSubstr;
 using ::testing::Return;
 
 ListJobsRequest GetListJobsRequest() {
@@ -107,9 +109,8 @@ TEST(BigQueryJobStubTest, GetJobRestClientError) {
   opts.set<EndpointOption>("bigquery.googleapis.com");
   rest_internal::RestContext context;
   DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
-  auto result = rest_stub.GetJob(context, std::move(job_request));
-  EXPECT_FALSE(result.ok());
-  EXPECT_THAT(result.status().code(), Eq(StatusCode::kUnavailable));
+  auto response = rest_stub.GetJob(context, std::move(job_request));
+  EXPECT_THAT(response, StatusIs(StatusCode::kUnavailable, HasSubstr("")));
 }
 
 TEST(BigQueryJobStubTest, GetJobRestResponseError) {
@@ -132,9 +133,8 @@ TEST(BigQueryJobStubTest, GetJobRestResponseError) {
   opts.set<EndpointOption>("bigquery.googleapis.com");
   rest_internal::RestContext context;
   DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
-  auto result = rest_stub.GetJob(context, std::move(job_request));
-  EXPECT_FALSE(result.ok());
-  EXPECT_THAT(result.status().code(), Eq(StatusCode::kInvalidArgument));
+  auto response = rest_stub.GetJob(context, std::move(job_request));
+  EXPECT_THAT(response, StatusIs(StatusCode::kInvalidArgument, HasSubstr("")));
 }
 
 TEST(BigQueryJobStubTest, ListJobsSuccess) {
@@ -194,9 +194,8 @@ TEST(BigQueryJobStubTest, ListJobsRestClientError) {
   opts.set<EndpointOption>("bigquery.googleapis.com");
   rest_internal::RestContext context;
   DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
-  auto result = rest_stub.ListJobs(context, std::move(list_jobs_request));
-  EXPECT_FALSE(result.ok());
-  EXPECT_THAT(result.status().code(), Eq(StatusCode::kUnavailable));
+  auto response = rest_stub.ListJobs(context, std::move(list_jobs_request));
+  EXPECT_THAT(response, StatusIs(StatusCode::kUnavailable, HasSubstr("")));
 }
 
 TEST(BigQueryJobStubTest, ListJobsRestResponseError) {
@@ -217,9 +216,10 @@ TEST(BigQueryJobStubTest, ListJobsRestResponseError) {
   opts.set<EndpointOption>("bigquery.googleapis.com");
   rest_internal::RestContext context;
   DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
-  auto result = rest_stub.ListJobs(context, std::move(list_jobs_request));
-  EXPECT_FALSE(result.ok());
-  EXPECT_THAT(result.status().code(), Eq(StatusCode::kInvalidArgument));
+
+  auto response = rest_stub.ListJobs(context, std::move(list_jobs_request));
+
+  EXPECT_THAT(response, StatusIs(StatusCode::kInvalidArgument, HasSubstr("")));
 }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
--- a/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc
@@ -82,8 +82,8 @@ TEST(BigQueryJobStubTest, GetJobSuccess) {
   EXPECT_THAT(result->job.status.state, Eq("DONE"));
 }
 
-TEST(BigQueryJobStubTest, ProjectIdEmpty) {
-  auto mock_rest_client = std::make_unique<MockRestClient>();
+TEST(BigQueryJobStubTest, GetJobProjectIdEmpty) {
+  auto mock_rest_client = absl::make_unique<MockRestClient>();
   GetJobRequest job_request("", "j123");
   Options opts;
   opts.set<EndpointOption>("bigquery.googleapis.com");
@@ -109,7 +109,7 @@ TEST(BigQueryJobStubTest, JobIdEmpty) {
                        HasSubstr("Invalid GetJobRequest: Job Id is empty")));
 }
 
-TEST(BigQueryJobStubTest, RestClientError) {
+TEST(BigQueryJobStubTest, GetJobRestClientError) {
   // Get() fails.
   auto mock_rest_client = std::make_unique<MockRestClient>();
   EXPECT_CALL(*mock_rest_client, Get(An<rest::RestRequest const&>()))
@@ -126,7 +126,7 @@ TEST(BigQueryJobStubTest, RestClientError) {
   EXPECT_THAT(result.status().code(), Eq(StatusCode::kUnavailable));
 }
 
-TEST(BigQueryJobStubTest, BuildRestResponseError) {
+TEST(BigQueryJobStubTest, GetJobRestResponseError) {
   // Invalid Rest response.
   auto mock_payload = std::make_unique<MockHttpPayload>();
   auto mock_response = std::make_unique<MockRestResponse>();
@@ -147,6 +147,104 @@ TEST(BigQueryJobStubTest, BuildRestResponseError) {
   rest_internal::RestContext context;
   DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
   auto result = rest_stub.GetJob(context, std::move(job_request));
+  EXPECT_FALSE(result.ok());
+  EXPECT_THAT(result.status().code(), Eq(StatusCode::kInvalidArgument));
+}
+
+TEST(BigQueryJobStubTest, ListJobsSuccess) {
+  std::string job_response_payload = R"({"etag": "tag-1",
+          "kind": "kind-1",
+          "next_page_token": "npt-123",
+          "jobs": [
+              {
+                "id": "1",
+                "kind": "kind-2",
+                "reference": {"project_id": "p123", "job_id": "j123"},
+                "state": "DONE",
+                "configuration": {
+                   "job_type": "QUERY",
+                   "query_config": {"query": "select 1;"}
+                },
+                "status": {"state": "DONE"},
+                "user_email": "user-email",
+                "principal_subject": "principal-subj"
+              }
+  ]})";
+  auto mock_response = absl::make_unique<MockRestResponse>();
+
+  EXPECT_CALL(*mock_response, StatusCode)
+      .WillRepeatedly(Return(HttpStatusCode::kOk));
+  EXPECT_CALL(*mock_response, Headers)
+      .WillRepeatedly(Return(std::multimap<std::string, std::string>()));
+  EXPECT_CALL(std::move(*mock_response), ExtractPayload)
+      .WillOnce(
+          Return(ByMove(MakeMockHttpPayloadSuccess(job_response_payload))));
+
+  auto mock_rest_client = absl::make_unique<MockRestClient>();
+  EXPECT_CALL(*mock_rest_client, Get(An<rest::RestRequest const&>()))
+      .WillOnce(Return(ByMove(
+          std::unique_ptr<rest::RestResponse>(std::move(mock_response)))));
+
+  ListJobsRequest list_jobs_request("p123");
+  Options opts;
+  opts.set<EndpointOption>("bigquery.googleapis.com");
+  rest_internal::RestContext context;
+  DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
+
+  auto result = rest_stub.ListJobs(context, std::move(list_jobs_request));
+  ASSERT_STATUS_OK(result);
+  EXPECT_THAT(result->http_response.http_status_code, Eq(HttpStatusCode::kOk));
+}
+
+TEST(BigQueryJobStubTest, ListJobsProjectIdEmpty) {
+  auto mock_rest_client = absl::make_unique<MockRestClient>();
+  ListJobsRequest list_jobs_request("");
+  Options opts;
+  opts.set<EndpointOption>("bigquery.googleapis.com");
+  rest_internal::RestContext context;
+  DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
+  auto result = rest_stub.ListJobs(context, std::move(list_jobs_request));
+  EXPECT_THAT(
+      result,
+      StatusIs(StatusCode::kInvalidArgument,
+               HasSubstr("Invalid ListJobsRequest: Project Id is empty")));
+}
+
+TEST(BigQueryJobStubTest, ListJobsRestClientError) {
+  auto mock_rest_client = absl::make_unique<MockRestClient>();
+  EXPECT_CALL(*mock_rest_client, Get(An<rest::RestRequest const&>()))
+      .WillOnce(
+          Return(rest::AsStatus(HttpStatusCode::kInternalServerError, "")));
+
+  ListJobsRequest list_jobs_request("p123");
+  Options opts;
+  opts.set<EndpointOption>("bigquery.googleapis.com");
+  rest_internal::RestContext context;
+  DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
+  auto result = rest_stub.ListJobs(context, std::move(list_jobs_request));
+  EXPECT_FALSE(result.ok());
+  EXPECT_THAT(result.status().code(), Eq(StatusCode::kUnavailable));
+}
+
+TEST(BigQueryJobStubTest, ListJobsRestResponseError) {
+  auto mock_payload = absl::make_unique<MockHttpPayload>();
+  auto mock_response = absl::make_unique<MockRestResponse>();
+  EXPECT_CALL(*mock_response, StatusCode)
+      .WillRepeatedly(Return(HttpStatusCode::kBadRequest));
+  EXPECT_CALL(std::move(*mock_response), ExtractPayload)
+      .WillOnce(Return(std::move(mock_payload)));
+
+  auto mock_rest_client = absl::make_unique<MockRestClient>();
+  EXPECT_CALL(*mock_rest_client, Get(An<rest::RestRequest const&>()))
+      .WillOnce(Return(ByMove(
+          std::unique_ptr<rest::RestResponse>(std::move(mock_response)))));
+
+  ListJobsRequest list_jobs_request("p123");
+  Options opts;
+  opts.set<EndpointOption>("bigquery.googleapis.com");
+  rest_internal::RestContext context;
+  DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
+  auto result = rest_stub.ListJobs(context, std::move(list_jobs_request));
   EXPECT_FALSE(result.ok());
   EXPECT_THAT(result.status().code(), Eq(StatusCode::kInvalidArgument));
 }
--- a/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc
@@ -34,11 +34,9 @@ using ::google::cloud::testing_util::MakeMockHttpPayloadSuccess;
 using ::google::cloud::testing_util::MockHttpPayload;
 using ::google::cloud::testing_util::MockRestClient;
 using ::google::cloud::testing_util::MockRestResponse;
-using ::google::cloud::testing_util::StatusIs;
 using ::testing::An;
 using ::testing::ByMove;
 using ::testing::Eq;
-using ::testing::HasSubstr;
 using ::testing::Return;
 
 TEST(BigQueryJobStubTest, GetJobSuccess) {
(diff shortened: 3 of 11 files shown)


google/cloud/bigquery/v2/minimal/internal/job_logging.cc
```diff
@@ -57,6 +57,33 @@ StatusOr<GetJobResponse> BigQueryJobLogging::GetJob(
   return response;
 }
 
+StatusOr<ListJobsResponse> BigQueryJobLogging::ListJobs(
+    rest_internal::RestContext& rest_context, ListJobsRequest const& request) {
+  char const* context = __func__;
+  GCP_LOG(DEBUG) << context
```
coryan, line 63: if you provide `operator<<(std::ostream&, ListJobsRequest const&)` then this can be simplified using  one of the [log wrappers](https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/internal/log_wrapper.h).  The code would become:

```cc
StatusOr<ListJobsResponse> BigQueryJobLogging::ListJobs(
    rest_internal::RestContext& rest_context, ListJobsRequest const& request) {
  return google::cloud::internal::LogWrapper(
      [this](auto& context, auto const& request) { return child_->ListJobs(context, request); },
      rest_context, request, __func__, tracing_options_);
}
```

google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc
```diff
@@ -62,6 +62,41 @@ StatusOr<GetJobResponse> DefaultBigQueryJobRestStub::GetJob(
...
+      }
+    }
+  }
+
+  // Call the rest client to get job details from the server as a RestResponse.
+  auto rest_response = rest_stub_->Get(std::move(*rest_request));
+  if (!rest_response) return std::move(rest_response).status();
+
+  // Convert RestResponse to HttpResponse.
+  auto http_response =
```
coryan, line 92: This code will get repetitive very quickly, consider refactoring some of this to a helper (template) function.  Maybe something like:

https://github.com/googleapis/google-cloud-cpp/blob/63fd5d00d4305413f9abdb153c1c70b6399a906a/google/cloud/storage/internal/rest_client.cc#LL60-L76C2

But without the `failure_predicate`?

coryan (commented):
I am missing tests for the logging and metadata changes.

google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc
```diff
@@ -62,6 +62,41 @@ StatusOr<GetJobResponse> DefaultBigQueryJobRestStub::GetJob(
...
 }
 
+StatusOr<ListJobsResponse> DefaultBigQueryJobRestStub::ListJobs(
+    rest_internal::RestContext& rest_context, ListJobsRequest const& request) {
+  internal::OptionsSpan span(options_);
+
+  if (request.project_id().empty()) {
+    return internal::InvalidArgumentError(
+        "Invalid ListJobsRequest: Project Id is empty", GCP_ERROR_INFO());
+  }
```
devbww, line 72: This seems to relate to something I mentioned previously:

>> I have addeed input validation checks for the presence of all required parameters where we know the api will fail without them.
> 
> For what it's worth, this is the opposite of what I would suggest. That is, I would remove all input validation checks where we know the API will fail without them. It duplicates tests, possibly inventing new error values, and is fragile in the case of changing requirements in the backend.

So, unless we have some client-side code that depends on `request.project_id` being non-empty, we should avoid such checks.  Instead, just let the chips fall where they may.

coryan: https://github.com/googleapis/google-cloud-cpp/blob/main/doc/adr/2018-06-19-do-not-duplicate-server-side-validation.md

jsrinnn (commented):
Sorry missed replying to the above. I don't see tests in other projects as well for logging and metadata decorators. Is there an example that I can follow?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-:-NR9zncnDbLRC8cHKxtA:baflpev)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @coryan and @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 92 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZD-r1-92:-NRA-9fU411qTAXPv-Ha:buozgvc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce8ff008d21672384551859d672dd3d526657f70/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L92)):*

Done.

Ah...very nice. Thank You.
</blockquote></details>

The PrepareRestRequest template function seems to be causing an msan error related to MemorySanitizer: use-of-uninitialized-value (https://screenshot.googleplex.com/AzL7ANdpewswoQH). I am looking into it but if you see something obvioud please let me know.

coryan (changes requested):
We do not write unit tests for generated code (because we already tested the generator). But we do write tests for hand-crafted code, e.g.:

https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/storage/internal/logging_client_test.cc

> Reviewed 6 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-:-NRA3cRrB9Ma5xLk5Lqb:b-k6fvu)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @dbolduc, @devbww, and @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r1-63:-NRA0ziX9Q73wmtFHlcO:b-geqwv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce8ff008d21672384551859d672dd3d526657f70/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L63)):*

Ah...Actually I tried that when I was doing GetJob(). It fails in DebugString() method (inside LogWrapper) mainly because the request is not a protobuf message.
</blockquote></details>

I see. You can use something like this instead:

```cc
template <
    typename Functor, typename Request,
    typename Result = google::cloud::internal::invoke_result_t<
        Functor, rest_internal::RestContext&, Request const&>>
Result MyLogWrapper(Functor&& functor, rest_internal::RestContext& context,
                  Request const& request,
                  char const* where, TracingOptions const& /*options*/) {
  GCP_LOG(DEBUG) << where << "() << " << request;
  auto response = functor(context, request);
  if (!response) {
    GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
  } else {
    GCP_LOG(DEBUG) << where << "() >> status=" << *response;
  }
  return response;
}
```

The point is: you can refactor the repetitive code and get yourself better error messages in the tests. Feel free to push this to the next PR.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_metadata.cc` line 47 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZB-r1-47:-NRA3K9X2EY7bXQPqJUq:b-ph8tb7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce8ff008d21672384551859d672dd3d526657f70/google/cloud/bigquery/v2/minimal/internal/job_metadata.cc#L47)):*

To be honest I just tried to add as much metadata info as possible here to debug listJobs api as necessary. I couldn't find any docs that said if there was any limitations on the server side just that the parameter name value need to be joined by `&' (https://cloud.google.com/apis/docs/system-parameters). Since these parameters are all part of the ListJobsRequest my assumption was that , it should be recognized.
</blockquote></details>

@dbolduc you understand this stuff better than I do, any advice?

jsrinnn (commented):
Ack. Got it. Thanks. I'll do this in a follow-up PR.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-:-NRA9tlW0ysFKNGi8F9f:b-9h66i4)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @coryan, @dbolduc, and @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 63 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r1-63:-NRAFPovC24gbv06y7UD:bdk2f0e) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce8ff008d21672384551859d672dd3d526657f70/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L63)):*

I see. You can use something like this instead:

```cc
template <
    typename Functor, typename Request,
    typename Result = google::cloud::internal::invoke_result_t<
        Functor, rest_internal::RestContext&, Request const&>>
Result MyLogWrapper(Functor&& functor, rest_internal::RestContext& context,
                  Request const& request,
                  char const* where, TracingOptions const& /*options*/) {
  GCP_LOG(DEBUG) << where << "() << " << request;
  auto response = functor(context, request);
  if (!response) {
    GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
  } else {
    GCP_LOG(DEBUG) << where << "() >> status=" << *response;
  }
  return response;
}
```

The point is: you can refactor the repetitive code and get yourself better error messages in the tests. Feel free to push this to the next PR.
</blockquote></details>

Done.

Thank you. I have made the code changes in this PR. I'll add the unit tests for the decorators in the separate PR based on your earlier feedback.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_metadata.cc` line 47 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZB-r1-47:-NRAFeJl4YK6Wepgua9h:bx6e8ph) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce8ff008d21672384551859d672dd3d526657f70/google/cloud/bigquery/v2/minimal/internal/job_metadata.cc#L47)):*

@dbolduc you understand this stuff better than I do, any advice?
</blockquote></details>

I can address this in a followup PR based on Darren's feedback.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 92 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZD-r1-92:-NRDhmA5-j2JvQ-0Gmap:bu99145) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce8ff008d21672384551859d672dd3d526657f70/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L92)):*

The PrepareRestRequest template function seems to be causing an msan error related to MemorySanitizer: use-of-uninitialized-value (https://screenshot.googleplex.com/AzL7ANdpewswoQH). I am looking into it but if you see something obvioud please let me know.
</blockquote></details>

Done

This should now be fixed with the latest changes. Unfortunately I didn't have enough resources to run` msan-pr` locally so we will have to see how the check runs once I push the changes.

coryan (commented):
Looks like you need to rebase and replace `absl::make_unique<>` with `std::make_unique<>`.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-:-NRE2hwJApkNOCBf98-_:bn18sj2)* status: all files reviewed, 4 unresolved discussions (waiting on @devbww and @jsrinnn)

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081)* status: all files reviewed, 4 unresolved discussions (waiting on @coryan and @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 116 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDu5g298kRyCT6Rv9B:-NREag3a-r3doI_vGa-J:b-3vibub) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L116)):*

Maybe:

```cc
static auto const kEpoch = std::chrono::system_clock::from_time_t(0);
if (r.min_creation_time() > kEpoch) { ...
...
```

</blockquote></details>

Why is the UNIX epoch special?

jsrinnn:
FYI: I am still working on the code review feedback. The last push was just rebase from upstream main. Some of the checks will still fail since some of my project files are using absl::make_unique

dbolduc (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081)* status: all files reviewed, 4 unresolved discussions (waiting on @coryan, @devbww, and @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_metadata.cc` line 47 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZB-r1-47:-NRFTg0L3ghd7Rbtfu5I:b6mthnw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce8ff008d21672384551859d672dd3d526657f70/google/cloud/bigquery/v2/minimal/internal/job_metadata.cc#L47)):*

I can address this in a followup PR based on Darren's feedback.
</blockquote></details>

Sorry, I am busy. I will look at this next Monday.

jsrinnn (commented):
Done

Addressed code review feedback and rebased with upstream main. 

I have locally run the basic pr checks (checkers and clang-tidy) and the bazel build but not the previously failing checks as they were taking a really long time. I'll observe for any failures after I push the changes.

PTAL.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-:-NRE74ddBl_JUygFR7w0:bsvnkp)* status: 7 of 11 files reviewed, 4 unresolved discussions (waiting on @coryan and @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 35 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDrmsA-pcdfgNE_AS4:-NREDLie27978Ziv8Pbg:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L35)):*

Each `GCP_LOG()` goes into a separate line in the log, which makes this output hard to read / parse. Furthermore, while a single `GCP_LOG()` is thread safe, other threads may split your output.

Consider something like:

```
auto formatter = [](std::string* out, auto const& header) {
    absl::StrAppend(out, "name=", header.first, ", value={", absl::StrJoin(h.second, "&"), "}");
};
GCP_LOG(DEBUG) << where << "() " << request << ", Context={" << absl::StrJoin(context.headers(), ", ", formatter) << "}";
```

I am going from memory, check my math..

</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 49 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDtbMk9V5OQJDc1Gt8:-NREDZIm3qV3yduy_hxH:b-z2b0f3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L49)):*

At the very least you should log the success case.  It would be good to actually log the response. Otherwise folks will wonder what happened to their response.
</blockquote></details>

I modified to log the status for either case. I am hesitant logging the response because of the size as well as there maybe sensistive data.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 116 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDu5g298kRyCT6Rv9B:-NREEByI1x5IlclnvK9k:b-glmig0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L116)):*

Why is the UNIX epoch special?
</blockquote></details>

Done.

Nice!

Not sure I follow the gist of the question but with regards to the context here, the creation_time is a time_point object, so this comparison is just a zero equivalence check to see whether the value is initialized.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 126 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDuXHS0LT3QU8-voWd:-NREGZN88rabVz7zsHK5:b17znem) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L126)):*

FYI, you could do something like:

```cc
auto add_if_not_empty = [&](char const* key, auto const& v) { if (v.empty()) return; request.AddQueryParameter(key, v); };
add_if_not_empty("pageToken", r.page_token());
add_if_not_empty("projection", r.projection().value());
```

Your call.
</blockquote></details>

Done.

Thank You.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 148 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDv4NlDewDVdbB0WEV:-NREPUVr09DmGjYzWir0:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L148)):*

Good news, these functions are testable now.
</blockquote></details>

Done

coryan (approved):
I trust you will fix the epoch thing.

> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-:-NRFxBSW86HALDFrEZfj:bhoq98a)* status: all files reviewed, 3 unresolved discussions (waiting on @devbww and @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 49 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDtbMk9V5OQJDc1Gt8:-NRFvrBWC8eePuRCRgO3:bh8y6h1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L49)):*

I modified to log the status for either case. I am hesitant logging the response because of the size as well as there maybe sensistive data.
</blockquote></details>

SGTM.  For future consideration: there are some helpers to truncate large messages, but we can do that later.  As to the sensitive data: this is optional and lives in the customer side, they can chose to not log if they are concerned. 
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 116 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDu5g298kRyCT6Rv9B:-NRFwJKTBmcmp4yDFup5:bbkhk2q) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L116)):*

Done.

Nice!

Not sure I follow the gist of the question but with regards to the context here, the creation_time is a time_point object, so this comparison is just a zero equivalence check to see whether the value is initialized.
</blockquote></details>

Oh, if you want that, then say that:

```cc
if (r.min_creation_time() != std::chrono::system_clock::time_point{})
```

Though I will note that the epoch is not "special", maybe the application *wanted* to set the `min_creation_time()` to the epoch...

google/cloud/bigquery/v2/minimal/internal/job_logging.cc
```diff
@@ -25,6 +26,25 @@ namespace cloud {
...
+          typename Result = google::cloud::internal::invoke_result_t<
+              Functor, rest_internal::RestContext&, Request const&>>
+Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
+                     Request const& request, char const* where,
+                     TracingOptions const& /*options*/) {
+  auto formatter = [](std::string* out, auto const& header) {
+    absl::StrAppend(out, "name=", header.first, ", value={",
+                    absl::StrJoin(header.second, "&"), "}");
+  };
+  GCP_LOG(DEBUG) << where << "() " << request << ", Context={"
```
devbww, line 39: Note that the request/response streaming here does not heed the `TracingOptions`.  It probably should.

google/cloud/bigquery/v2/minimal/internal/job_logging.cc
```diff
@@ -25,6 +26,25 @@ namespace cloud {
...
+                 << absl::StrJoin(context.headers(), ", ", formatter) << "}";
+
+  auto response = functor(context, request);
+  GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
```
devbww, line 43: We should log the response, when available ... that's half the purpose of the logging stub.

google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc
```diff
@@ -49,17 +52,31 @@ StatusOr<GetJobResponse> DefaultBigQueryJobRestStub::GetJob(
...
+  return rest_request;
+}
 
-  // Convert RestResponse to HttpResponse.
-  auto http_response =
-      BigQueryHttpResponse::BuildFromRestResponse(std::move(*rest_response));
-  if (!http_response) return std::move(http_response).status();
+StatusOr<GetJobResponse> DefaultBigQueryJobRestStub::GetJob(
+    rest_internal::RestContext& rest_context, GetJobRequest const& request) {
+  internal::OptionsSpan span(options_);
```
devbww, line 60: These should be instantiated long before the stub layer is reached.  If that is part of the bottom-up implementation strategy, then I'm sorry to (re)raise the issue.  As long as we're remembering to remove these as we build upwards, I'm happy.

google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc
```diff
@@ -82,71 +95,129 @@ TEST(BigQueryJobStubTest, GetJobSuccess) {
...
   opts.set<EndpointOption>("bigquery.googleapis.com");
   rest_internal::RestContext context;
   DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
   auto result = rest_stub.GetJob(context, std::move(job_request));
-  EXPECT_THAT(
-      result,
-      StatusIs(StatusCode::kInvalidArgument,
-               HasSubstr("Invalid GetJobRequest: Project Id is empty")));
+  EXPECT_FALSE(result.ok());
+  EXPECT_THAT(result.status().code(), Eq(StatusCode::kUnavailable));
```
devbww, line 112: Why did we revert from `EXPECT_THAT(result, StatusIs(...))`?

google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc
```diff
@@ -82,71 +95,129 @@ TEST(BigQueryJobStubTest, GetJobSuccess) {
...
+      .WillRepeatedly(Return(HttpStatusCode::kBadRequest));
+  EXPECT_CALL(std::move(*mock_response), ExtractPayload)
+      .WillOnce(Return(std::move(mock_payload)));
+
+  // Get() is successful.
   auto mock_rest_client = std::make_unique<MockRestClient>();
-  GetJobRequest job_request("p123", "");
+  EXPECT_CALL(*mock_rest_client, Get(An<rest::RestRequest const&>()))
+      .WillOnce(Return(ByMove(
+          std::unique_ptr<rest::RestResponse>(std::move(mock_response)))));
```
devbww, line 128: What value are we expecting for `*mock_response` here, given that we moved it above?

google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc
```diff
@@ -82,71 +95,129 @@ TEST(BigQueryJobStubTest, GetJobSuccess) {
...
   Options opts;
   opts.set<EndpointOption>("bigquery.googleapis.com");
   rest_internal::RestContext context;
   DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
   auto result = rest_stub.GetJob(context, std::move(job_request));
-  EXPECT_THAT(result,
-              StatusIs(StatusCode::kInvalidArgument,
-                       HasSubstr("Invalid GetJobRequest: Job Id is empty")));
+  EXPECT_FALSE(result.ok());
+  EXPECT_THAT(result.status().code(), Eq(StatusCode::kInvalidArgument));
```
devbww, line 137: Ditto on `StatusIs`.

jsrinnn (commented):
Yes. I think the latest changes should address it but let me know if there is any additional changes needed and I can address it in a followup PR. Thanks.

Addressed all the latest feedback. PTAL.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-:-NRIuNIf5dC_NCQsiJAL:b2fqrk3)* status: 8 of 11 files reviewed, 10 unresolved discussions (waiting on @coryan and @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 49 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDtbMk9V5OQJDc1Gt8:-NRIvmrRAATmmAfDBrmN:b-dfp9nc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L49)):*

SGTM.  For future consideration: there are some helpers to truncate large messages, but we can do that later.  As to the sensitive data: this is optional and lives in the customer side, they can chose to not log if they are concerned. 
</blockquote></details>

Ack. SGTM. For now I'll leave this as is.  My main concern was sensitive data and size of the message.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 39 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-39:-NRIurJAB_QgTFt4_eBf:b-eug5ij) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L39)):*

Note that the request/response streaming here does not heed the `TracingOptions`.  It probably should.
</blockquote></details>

I am not sure how the tracing options gets logged for rest and custom classes. The references I have seen in the repo seem to use it in the context of protobuf classes.

If there is an example for non-protobuf please provide a link and I can address it in a follow up PR when I am adding unit tests for the logging and metadata decorators.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 43 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-43:-NRIv6gy-NwZM1OwePxr:b-vl8lef) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L43)):*

We should log the response, when available ... that's half the purpose of the logging stub.
</blockquote></details>

Please see below for response.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 116 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDu5g298kRyCT6Rv9B:-NRIz3ISDqVBOt4XgjL5:b-qob21w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L116)):*

Oh, if you want that, then say that:

```cc
if (r.min_creation_time() != std::chrono::system_clock::time_point{})
```

Though I will note that the epoch is not "special", maybe the application *wanted* to set the `min_creation_time()` to the epoch... 
</blockquote></details>

Done.

Based on the previous feedback, I have added default values for all members in the constructor now and check against these default values. I am using `std::chrono::system_clock::time_point{}` as default for max and min creation times as they are timepoint types.

Please LMK if there is any additional changes required here.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 28 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDph_FLEFarouxgRTB-r4-28:-NRIylGH-7E7zQM5Afmb:bczbq2m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L28)):*

I'm still not sure why the UNIX epoch is special, other than it happens to be the (unspecified until C++20) value of a default-constructed `time_point`.  We should probably be more explicit about how an "unset" `{min,max}_creation_time` is represented.

And related to that, it still doesn't look like the `ListJobRequest` constructors initialize members like `all_users_` and `max_results_`.
</blockquote></details>

Done.

Added default values for all members in the constructor and modified the request to check for these default values.

I did the same for GetJobRequest as well for the `location` member.

Thank you for pointing this out.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 60 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZD-r4-60:-NRIo8gm4bTWUkL5Uh2F:b4v65le) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L60)):*

These should be instantiated long before the stub layer is reached.  If that is part of the bottom-up implementation strategy, then I'm sorry to (re)raise the issue.  As long as we're remembering to remove these as we build upwards, I'm happy.
</blockquote></details>

Options is needed here to retrieve the Endpoint to build the request URL from the values in the request, and based on previous feedback, we had agreed upon using OptionsSpan as an alternative solution. I didn't get the impression that we would be removing this or atleast not anytime soon.

My apologies, but Its a bit frustrating for me to keep going back and forth on this when we had agreed on a solution before. It halts my progress for other apis as well.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc` line 112 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZF-r4-112:-NRIoilHBMxtrbySN42c:b-rnrkzg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc#L112)):*

Why did we revert from `EXPECT_THAT(result, StatusIs(...))`?
</blockquote></details>

I don't think it was ever reverted. Its always been like this. It may have been a different file which had StatusIs.

I have modified here and elsehere in this file to use StatusIs for error response, based on above feedback. I'll also keep in mind for future reference to always use StatusIs for checking error cases. Thank you.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc` line 128 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZF-r4-128:-NRItoON96yIITo-xFpc:bn5yjsq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc#L128)):*

What value are we expecting for `*mock_response` here, given that we moved it above?
</blockquote></details>

mockresponse here will give a bad request error . 
This test shows a Get() being successful but return a response with error. 

The previous test showed a Get() failure.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc` line 137 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZF-r4-137:-NRIuJXR5VE4EGHEb7Wx:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc#L137)):*

Ditto on `StatusIs`.
</blockquote></details>

Done.

coryan (approved):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081)* status: all files reviewed, 8 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 39 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-39:-NRK-Lj8AnAzkHBamQ5i:b-rp6j6d) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L39)):*

I am not sure how the tracing options gets logged for rest and custom classes. The references I have seen in the repo seem to use it in the context of protobuf classes.

If there is an example for non-protobuf please provide a link and I can address it in a follow up PR when I am adding unit tests for the logging and metadata decorators.
</blockquote></details>

FYI: `TracingOptions` controls the format of the output, including the maximum size of a string dumped in the log.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc` line 113 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK-1iGDJYfQp7E7P4q:-NRK-1iHCwx0dxMkAUbj:b56wxiu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc#L113)):*
> ```cpp
>   DefaultBigQueryJobRestStub rest_stub(std::move(mock_rest_client), opts);
>   auto response = rest_stub.GetJob(context, std::move(job_request));
>   EXPECT_THAT(response, StatusIs(StatusCode::kUnavailable, HasSubstr("")));
> ```

FYI, the `HasSubstr()` is optional

devbww (commented):
> Reviewed 1 of 2 files at r2, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081)* status: all files reviewed, 7 unresolved discussions (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 39 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-39:-NRKB4nq7QOelurtfGD0:b3j2e67) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L39)):*

FYI: `TracingOptions` controls the format of the output, including the maximum size of a string dumped in the log.
</blockquote></details>

We should use the `TracingOptions` for formatting all the trace-log objects, be they protobufs or not.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 43 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-43:-NRKBSXB5pchJv9B6Ojg:b-e6vzro) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L43)):*

Please see below for response.
</blockquote></details>

I'm not sure what this is referring to.  Sorry.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 200 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK-Q6m4tG4y97ror1R:-NRK-Q6m4tG4y97ror1S:bed1lhx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.h#L200)):*
> ```objc
>                                                       Options const& opts);
> 
> std::ostream& operator<<(std::ostream& os, GetJobRequest const& request);
> ```

`#include <ostream>`
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 116 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDu5g298kRyCT6Rv9B:-NRK19V58xletq4rzGig:bwqpyb6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L116)):*

Done.

Based on the previous feedback, I have added default values for all members in the constructor now and check against these default values. I am using `std::chrono::system_clock::time_point{}` as default for max and min creation times as they are timepoint types.

Please LMK if there is any additional changes required here.
</blockquote></details>

I'm all for initializing the members with appropriate values.

The issue is that the value is then being used as a sentinel for whether we should call `AddQueryParameter()`.  If we called `set_min_creation_time(kDefaultTimepoint)` we wouldn't add the parameter, but if it was 1ns later we would.  That feels wrong.

That is, `time_point` doesn't have a suitable out-of-band value.  Perhaps you're reaching for something like `optional<time_point>`.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 75 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK-p0dComfFNnjK3li:-NRK-p0dComfFNnjK3lj:b2c3h88) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L75)):*
> ```cpp
> GetJobRequest::GetJobRequest(std::string project_id, std::string job_id)
>     : project_id_(std::move(project_id)), job_id_(std::move(job_id)) {
>   location_ = "";
> ```

This is redundant ... `location_` will have already been default constructed (to the empty string).
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 83 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK3rTC89j2bnsxvkNV:-NRK3rTC89j2bnsxvkNW:b-2wtrza) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L83)):*
> ```cpp
>   max_results_ = 0;
>   min_creation_time_ = kDefaultTimepoint;
>   max_creation_time_ = kDefaultTimepoint;
> ```

These should be in the initializer list.



_Code quote:_
```cpp
  all_users_ = false;
  max_results_ = 0;
  min_creation_time_ = kDefaultTimepoint;
  max_creation_time_ = kDefaultTimepoint;
```
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 87 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK0goL2vyCdcOi-_q4:-NRK0goL2vyCdcOi-_q5:b-v44jq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L87)):*
> ```cpp
>   projection_.value = "";
>   state_filter_.value = "";
>   parent_job_id_ = "";
> ```

These re-initializations are similarly redundant.



_Code quote:_
```cpp
  page_token_ = "";
  projection_.value = "";
  state_filter_.value = "";
  parent_job_id_ = "";
```
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 147 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK3Gnw4EqjcWPZVDZU:-NRK3Gnw4EqjcWPZVDZV:b-nlj1hz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L147)):*
> ```cpp
>     if (v.empty()) return;
>     request.AddQueryParameter(key, v);
>   };
> ```

An "add if not empty" would probably be clearer as `if (not empty) add;`.



_Code quote:_
```cpp
  auto add_if_not_empty = [&](char const* key, auto const& v) {
    if (v.empty()) return;
    request.AddQueryParameter(key, v);
  };
```
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 72 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZD-r1-72:-NRK6uE87RTG8dVZNF2j:b-144wdv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ce8ff008d21672384551859d672dd3d526657f70/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L72)):*

Done.

ok. I have removed it from here and GetJob as well.

My thinking was: Project_id is a required parameter which will cause a server error so we could save a server rountrip by doing the pre-check.
</blockquote></details>

Hopefully the document Carlos cited describes why that is not a saving in practice.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 60 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZD-r4-60:-NRK4vt993pTs1l9w8M5:b-hvi8ts) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L60)):*

Options is needed here to retrieve the Endpoint to build the request URL from the values in the request, and based on previous feedback, we had agreed upon using OptionsSpan as an alternative solution. I didn't get the impression that we would be removing this or atleast not anytime soon.

My apologies, but Its a bit frustrating for me to keep going back and forth on this when we had agreed on a solution before. It halts my progress for other apis as well.
</blockquote></details>

I'm sorry if there has been back-and-forth.

All I'm saying is that `OptionsSpan` should really only be instantiated in calls directly from the user (and during callbacks on behalf of the user).  This typically excludes the stub layer.

If your stub layer is currently called directly from the user, then I guess you have to temporarily do something to populate the current `Options`, but that should not be the long-term state.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc` line 112 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZF-r4-112:-NRK8-Lz6O6b4030B2sr:bamko8w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc#L112)):*
> I don't think it was ever reverted. Its always been like this.

I think I was only responding to a diff GitHub was displaying.  Perhaps it was some intermediate snapshot.  But, as long as we're consistent on `StatusIs()`, all is well.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc` line 128 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZF-r4-128:-NRK98W71t3KqDop3GA4:b5ugvvc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc#L128)):*

mockresponse here will give a bad request error . 
This test shows a Get() being successful but return a response with error. 

The previous test showed a Get() failure.
</blockquote></details>

I'm not sure that gets to my concern.  But that said, the `MockRestReponse` is inherently stateless, so its moved-from state is probably immaterial.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc` line 137 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZF-r4-137:-NRKAe5_4-_ZQ3HSzvGH:b-vzedxr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc#L137)):*

Done.
</blockquote></details>

`HasSubstr("")` is a tautology, so you might as well remove it.

jsrinnn (commented):
Addressed remaining feedback. 

I am enabling auto merge. Please feel free to disable it, if you feel there is are any additional feedback that needs to be addressed in this PR.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-:-NRK5QAb0EqIWjMO6YUU:bntc7n3)* status: 7 of 11 files reviewed, 3 unresolved discussions (waiting on @coryan and @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 39 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-39:-NRK58ye45ql2Kg1Pixe:bqiyimi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L39)):*

We should use the `TracingOptions` for formatting all the trace-log objects, be they protobufs or not.
</blockquote></details>

Done.

I used it to log the context headers. When we add the response later, we can use it there as well.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 43 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-43:-NRKCFOU9ijQwjxJCqib:b-pdvv8o) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L43)):*

I'm not sure what this is referring to.  Sorry.
</blockquote></details>

Basically acking with Carlos that response can be logged later as needed. 
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 200 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK-Q6m4tG4y97ror1R:-NRKCqxB-QgXKHV9RvQZ:bze00zq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.h#L200)):*

`#include <ostream>`
</blockquote></details>

Done

Its actually not needed.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 116 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDu5g298kRyCT6Rv9B:-NRKGOCi8JlYAv62GWg3:b-tlvjho) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L116)):*

I'm all for initializing the members with appropriate values.

The issue is that the value is then being used as a sentinel for whether we should call `AddQueryParameter()`.  If we called `set_min_creation_time(kDefaultTimepoint)` we wouldn't add the parameter, but if it was 1ns later we would.  That feels wrong.

That is, `time_point` doesn't have a suitable out-of-band value.  Perhaps you're reaching for something like `optional<time_point>`.
</blockquote></details>

Since we agreed on a different feedback that we are not doing input validation here, I would argue that any value that is set by the caller (that is not a default) should be set. If it is bad value, then the server will indicate that with an error. 

However I do see your point regarding a suitable out-of-band value. I do see a lot of `time_point{}` usage in the repo as an empty value check or as setter. Maybe the optional suggestion can be addressed in a separate PR.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 75 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK-p0dComfFNnjK3li:-NRKF3gDDaD8a4J6gLqQ:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L75)):*

This is redundant ... `location_` will have already been default constructed (to the empty string).
</blockquote></details>

Done
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 83 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK3rTC89j2bnsxvkNV:-NRKFq8w7nBaZtDzJZRX:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L83)):*

These should be in the initializer list.
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 87 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK0goL2vyCdcOi-_q4:-NRKF4GrCekcpeGB8tUS:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L87)):*

These re-initializations are similarly redundant.
</blockquote></details>

Done
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 147 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK3Gnw4EqjcWPZVDZU:-NRKGfxZAIVo-g-CzxbE:b-896fjx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L147)):*

An "add if not empty" would probably be clearer as `if (not empty) add;`.
</blockquote></details>

Done
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 60 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZD-r4-60:-NRKHBPj39UcD3wvzCiM:b-cv7gx8) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L60)):*

I'm sorry if there has been back-and-forth.

All I'm saying is that `OptionsSpan` should really only be instantiated in calls directly from the user (and during callbacks on behalf of the user).  This typically excludes the stub layer.

If your stub layer is currently called directly from the user, then I guess you have to temporarily do something to populate the current `Options`, but that should not be the long-term state.
</blockquote></details>

Ack. From my perspective,  I am not sure I follow what the long term alternative is being suggested here. Based on previous discussions, OptionsSpan was the solution we agreed on.

Options is needed here to retrieve the endpoint and construct the request url.

We can definetely revisit this for long term.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc` line 137 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZF-r4-137:-NRKIkfT106I8FIPTOP9:b-nreho3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc#L137)):*

`HasSubstr("")` is a tautology, so you might as well remove it.
</blockquote></details>

Done.

Yup Removed
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc` line 113 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK-1iGDJYfQp7E7P4q:-NRK3R_f-NjDewr-g8_S:b-6x377) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_test.cc#L113)):*

FYI, the `HasSubstr()` is optional
</blockquote></details>

Done

Removed HasSubstr() as it wasn't needed.

devbww (commented):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081)* status: all files reviewed, 2 unresolved discussions

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 39 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-39:-NRKnuAQ-oXTRxfcG3mK:bpyfk6y) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L39)):*

Done.

I used it to log the context headers. When we add the response later, we can use it there as well.
</blockquote></details>

That's good in that it heeds `truncate_string_field_longer_than` for the `context`, but there are other tracing options.  And they're all ignored for `request`.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 43 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-43:-NRKow72-b7yJ9JNkpgo:b-7dylgm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L43)):*

Basically acking with Carlos that response can be logged later as needed. 
</blockquote></details>

I'm not sure what "as needed" means.  Logging the request and response is the single responsibility of this layer, so I would claim that it is needed.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 200 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK-Q6m4tG4y97ror1R:-NRKpbkB-Mfa2eE0MIOj:buykr4f) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.h#L200)):*

Done

Its actually not needed.
</blockquote></details>

I'm not sure why you say that.  We use `std::ostream` on this very line.

Don't equate "builds without it" with "not needed" ... you should always include what you use.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 116 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDu5g298kRyCT6Rv9B:-NRKqOk5DCbZxUJo5BCv:b-icrfms) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L116)):*

Since we agreed on a different feedback that we are not doing input validation here, I would argue that any value that is set by the caller (that is not a default) should be set. If it is bad value, then the server will indicate that with an error. 

However I do see your point regarding a suitable out-of-band value. I do see a lot of `time_point{}` usage in the repo as an empty value check or as setter. Maybe the optional suggestion can be addressed in a separate PR.
</blockquote></details>

This is not related to a "bad value".

If there are other uses of `time_point{}` as an "empty value", we should fix them too.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 147 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK3Gnw4EqjcWPZVDZU:-NRKrUiqAqfUc2uNZZ2L:bx72lf7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L147)):*

Done
</blockquote></details>

Sorry I wasn't clearer.  I wasn't talking about word order in the name, but rather having the implementation be a direct translation of the name.

Concretely, and using the new name, why wouldn't this be clearer as ...
```
  auto if_not_empty_add = [&request](char const* key, auto const& v) {
    if (!v.empty()) request.AddQueryParameter(key, v);
  };
```

`if` `!` `empty` `Add`

But, just a suggestion to make things clearer.  Not a big deal.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081)* status: all files reviewed, 2 unresolved discussions

___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 60 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZD-r4-60:-NRKuVbwADwhsyH6Vc4s:baeh4z0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L60)):*

Ack. From my perspective,  I am not sure I follow what the long term alternative is being suggested here. Based on previous discussions, OptionsSpan was the solution we agreed on.

Options is needed here to retrieve the endpoint and construct the request url.

We can definetely revisit this for long term.

</blockquote></details>

Sorry for not being clearer.  I'll try to remember to bring it up again, but the only `OptionsSpan`s you should need are in the per-operation calls in the client.  Everything below that should be accessing the current options via `CurrentOptions()`.

[You could choose to plumb all needed parameters explicitly up the call stack, but we didn't do that with other services for historical reasons, and there doesn't seem to be a reason to change that approach here.]

jsrinnn (commented):
Thanks for the thourough review Bradley. I have replied to your latest comments.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-:-NRL7OWSAqkMjcK20SXq:b1zrawx)* status: all files reviewed, 2 unresolved discussions

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 39 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-39:-NRL2kEIEpEFv2szPh9a:b-7hn2fg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L39)):*

That's good in that it heeds `truncate_string_field_longer_than` for the `context`, but there are other tracing options.  And they're all ignored for `request`.
</blockquote></details>

Just to clarify, are you saying do the same with the request as well i.e have my own version of DebugString that does the same for request? I can't use the one in log_wrapper.h due to the dependency on grpc

If  my assumption above is correct then I can try to address that in the follow up PR for logging. 
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 43 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZ9-r4-43:-NRL3Fb8-v-tk2OpasM0:b-og0hwr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L43)):*

I'm not sure what "as needed" means.  Logging the request and response is the single responsibility of this layer, so I would claim that it is needed.
</blockquote></details>

Sorry for not being clear, the needed part is the for the fields in the response.. The fields that are not user sensistive can be logged. I'll try to address this in a follow up PR for logging where I'll be adding some unit tests for the decorator. I can also address you earlier feedback regarding tracing options .
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 200 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRK-Q6m4tG4y97ror1R:-NRL3j9uBlvB3ND7SnQq:b3c71wx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/76fce02195a2b65edc268569d7dd56f2ab0722a8/google/cloud/bigquery/v2/minimal/internal/job_request.h#L200)):*

I'm not sure why you say that.  We use `std::ostream` on this very line.

Don't equate "builds without it" with "not needed" ... you should always include what you use.
</blockquote></details>

Its being included in another header so from that perspective inlcuding it again is redundant i.e not needed but I do see your point about being explicit which is why I added it.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 116 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NRDu5g298kRyCT6Rv9B:-NRL2BWOBZhXOrv56QR-:bwxc7v6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/91658563bdb52594d0a069f8f05b008acc55f220/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L116)):*

This is not related to a "bad value".

If there are other uses of `time_point{}` as an "empty value", we should fix them too.
</blockquote></details>

The other usages are not in my project so I'll leave that up to the respective project owners. if you just search for timepoint{} you can see it being used in setters and as comparison checks in several places in the repo. 

For this particular file, If you feel strongly about me changing this to Optional<> then please a log a bug for me and assign it to me and I can address it in a follow up PR otherwise I may lose track of it as there are bunch of other things I am addressing already. This will be lower in priority for me.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc` line 60 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081#-NR8sFXH8NY0pwPuobZD-r4-60:-NRL2V2BC_kJ5JpxpjEH:byyizkz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3b9b76a75c245dd83b038fd0f6565b057e3786a9/google/cloud/bigquery/v2/minimal/internal/job_rest_stub.cc#L60)):*

Sorry for not being clearer.  I'll try to remember to bring it up again, but the only `OptionsSpan`s you should need are in the per-operation calls in the client.  Everything below that should be accessing the current options via `CurrentOptions()`.

[You could choose to plumb all needed parameters explicitly up the call stack, but we didn't do that with other services for historical reasons, and there doesn't seem to be a reason to change that approach here.]
</blockquote></details>

Ah...I see. I din't know about the CurrentOptions(). If I can retrieve the endpoint option that I set here as default option: https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/bigquery/v2/minimal/internal/job_options.cc#L47-#L72

If the above assumption is correct then I can change this to use CurrentOptions() but I'll do this at a later point once I am done with other work for the job get and list apis..

jsrinnn:
I added a tracking (Cleanup) bug (https://github.com/googleapis/google-cloud-cpp/issues/11105) to address the remaining code review feedback that we agreed to do in a separate or followup PRs. If I have missed out anything please feel free to update the issue.

dbolduc:
> Sorry, I am busy. I will look at this next Monday.

I didn't realize #11105 was a thing, so I opened a new issue #11120 for feedback on the metadata business.

PR #11102, opened 2023-03-24
feat: consume less entropy for PRNG

coryan:
We initialized our PRNG with a lot of entropy. This was overkill, and the entropy sources are not always an infinite resource.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11102)

### DIFF
--- a/google/cloud/internal/random.h
+++ b/google/cloud/internal/random.h
@@ -36,27 +36,20 @@ std::vector<unsigned int> FetchEntropy(std::size_t desired_bits);
 // in particular.
 using DefaultPRNG = std::mt19937_64;
 
-/**
- * Initialize any of the C++11 PRNGs in `<random>` using std::random_device.
- */
-template <typename Generator>
-Generator MakePRNG() {
-  // We need to get enough entropy to fully initialize the PRNG, that depends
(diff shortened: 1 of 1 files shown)


google/cloud/internal/random.h
```diff
@@ -36,28 +36,21 @@ std::vector<unsigned int> FetchEntropy(std::size_t desired_bits);
...
-  // We need to get enough entropy to fully initialize the PRNG, that depends
-  // on the size of its state.  The size in bits is `Generator::state_size`.
-  // We convert that to the number of `unsigned int` values; as this is what
-  // std::random_device returns.
-  auto const desired_bits = Generator::state_size * Generator::word_size;
-
+/// Create a new PRNG.
+inline DefaultPRNG MakeDefaultPRNG() {
+  // Fetch a few (typically 64) bits of entropy to initialize the PRNG. We used
+  // to fetch a lot more entropy, but that is overkill for our purposes.
```
devbww, line 42: I'd reword to avoid talking about what we used to do, and rather speak to what we now do.

Perhaps just _"Fetch a few bits ..., which is sufficient for our purposes"_?

devbww (approved):

PR #11106, opened 2023-03-25
impl(otel): tracing for streaming reads in generated connections

dbolduc:
Part of the work for #10620 

It is basically the same code as for Paginated APIs. Only the response type differs.

Note that this change only impacts generated connections. Handwritten connections like `bigtable::DataConnection` will be instrumented later. (This explains why less files are changed in this PR than the tracing stub PR (#11104))

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11106)

### DIFF
--- a/generator/internal/tracing_connection_generator.cc
+++ b/generator/internal/tracing_connection_generator.cc
@@ -118,7 +118,9 @@ Status TracingConnectionGenerator::GenerateCc() {
   CcLocalIncludes({
       vars("tracing_connection_header_path"),
       "google/cloud/internal/opentelemetry.h",
-      HasPaginatedMethod() ? "google/cloud/internal/traced_stream_range.h" : "",
+      HasPaginatedMethod() || HasStreamingReadMethod()
+          ? "google/cloud/internal/traced_stream_range.h"
+          : "",
   });
   CcSystemIncludes({"memory"});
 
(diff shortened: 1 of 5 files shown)


generator/internal/tracing_connection_generator.cc
```diff
@@ -249,6 +251,11 @@ std::unique_ptr<::google::cloud::AsyncStreamingReadWriteRpc<
     return R"""(
 StreamRange<$response_type$>
 $tracing_connection_class_name$::$method_name$($request_type$ const& request) {
+  auto span = internal::MakeSpan("$product_namespace$::$connection_class_name$::$method_name$");
+  auto scope = opentelemetry::trace::Scope(span);
+  auto sr = child_->$method_name$(std::move(request));
+  return internal::MakeTracedStreamRange<$response_type$>(
+        std::move(span), std::move(sr));
   return child_->$method_name$(request);
```
devbww, line 259: ???

dbolduc: Redundancy is a fundamental concept in cloud computing.

.... Fixed. Also realized that this method accepts the request by `const&`, whereas the paginated API accepts the request by value. So I stopped moving it.

devbww (approved):

Issue #11111, opened 2023-03-27
change spanner::RouteToLeaderOption to be on by default


devbww:
We have received the go-ahead to release the `spanner::RouteToLeaderOption` feature, but as "off by default".

Users can override this programmatically with `opts.set<spanner::RouteToLeaderOption>(true)`, or environmentally with `GOOGLE_CLOUD_CPP_SPANNER_ROUTE_TO_LEADER=1`.

This issue is for changing that back to "on by default" when the time is right.

coryan:
Still waiting for the service team to say it is Okay to do this.

PR #11112, opened 2023-03-27
feat(spanner): add the final pieces for the RouteToLeaderOption

devbww:
The final glue to add a routing header to Spanner RPCs that should be served in the leader region.

For the time being, the boolean option defaults to off (set as false), but that default can be changed with a suitably-positive value for `${GOOGLE_CLOUD_CPP_SPANNER_ROUTE_TO_LEADER}`.  The conditional code can be removed to restore the defaults-to-on behavior.  See #11111.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11112)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/spanner/internal/logging_result_set_reader.h"
 #include "google/cloud/spanner/internal/partial_result_set_resume.h"
 #include "google/cloud/spanner/internal/partial_result_set_source.h"
+#include "google/cloud/spanner/internal/route_to_leader.h"
 #include "google/cloud/spanner/internal/status_utils.h"
 #include "google/cloud/spanner/options.h"
 #include "google/cloud/spanner/query_partition.h"
@@ -387,8 +388,10 @@ StatusOr<google::spanner::v1::Transaction> ConnectionImpl::BeginTransaction(
   auto response = RetryLoop(
       RetryPolicyPrototype()->clone(), BackoffPolicyPrototype()->clone(),
       Idempotency::kIdempotent,
-      [&stub](grpc::ClientContext& context,
-              google::spanner::v1::BeginTransactionRequest const& request) {
+      [&stub, route_to_leader = ctx.route_to_leader](
+          grpc::ClientContext& context,
+          google::spanner::v1::BeginTransactionRequest const& request) {
+        if (route_to_leader) RouteToLeader(context);
         return stub->BeginTransaction(context, request);
       },
       begin, func);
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/internal/defaults.cc
```diff
@@ -105,6 +105,20 @@ Options DefaultOptions(Options opts) {
       (std::min)(min_sessions, max_sessions_per_channel * num_channels);
 
   if (!opts.has<spanner::RouteToLeaderOption>()) {
+#if 1
```
coryan, line 108: Do we want to leave the `'#if` here or just let version control store the original version?

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -830,6 +841,7 @@ ConnectionImpl::PartitionQueryImpl(
         Idempotency::kIdempotent,
         [&stub](grpc::ClientContext& context,
                 google::spanner::v1::PartitionQueryRequest const& request) {
+          RouteToLeader(context);
```
coryan, line 844: Why no `if (route_to_leader)` here?

google/cloud/spanner/internal/defaults.cc
```diff
@@ -105,6 +105,20 @@ Options DefaultOptions(Options opts) {
...
+    // TODO(#11111): Enable on-by-default behavior.
+    opts.set<spanner::RouteToLeaderOption>(false);
+    if (auto e = internal::GetEnv("GOOGLE_CLOUD_CPP_SPANNER_ROUTE_TO_LEADER")) {
```
coryan, line 111: You have previously argued that environment variables should override the values set in the code.  This does not seem to do that, thoughts?

google/cloud/spanner/internal/defaults_test.cc
```diff
@@ -85,7 +85,13 @@ TEST(Options, Defaults) {
   EXPECT_TRUE(opts.has<SpannerBackoffPolicyOption>());
   EXPECT_TRUE(opts.has<spanner_internal::SessionPoolClockOption>());
 
+#if 1
```
coryan, line 88: Ditto regarding `#if` vs. version control.

google/cloud/spanner/internal/connection_impl.cc
devbww: Because `PartitionQuery()` calls should be unconditionally routed to the leader.

google/cloud/spanner/internal/defaults.cc
devbww: I thought this was a little easier/clearer, but I'm happy to go the other way.  Done.

google/cloud/spanner/internal/defaults_test.cc
devbww: Ditto response.  Done.

google/cloud/spanner/internal/defaults.cc
devbww: You're correct that this environment variable is different ... it is for overriding the default value rather than a value set in code.  Let me think about that and get back to you on whether that difference is a feature or a bug.  Thanks.

google/cloud/spanner/internal/connection_impl.cc
coryan: Consider a comment (here and the other places where it is not conditional).  You probably won't forget that, I will.

coryan (approved):

google/cloud/spanner/internal/defaults.cc
coryan: Ac. SGTM.  We may want to consider some updates to our ADR regarding this, but that can happen elsewhere.

google/cloud/spanner/internal/connection_impl.cc
devbww: Done.

Issue #10895, opened 2023-02-21
Complete the doxygen2docfx generator


coryan:
I have gained enough experience with the `doxygen2docfx` prototype to start listing what we need to do.  I expect these will become separate feature requests.  I also expect the list will grow.

- [x] Handle `<variablelist>` elements
- [x] #11244
- [x] #11243
- [x] Add support for `<sectiondef>` this is needed for some groups.
- [x] Add support for `<orderedlist>` this is needed for some groups.
- [x] Handle groups, these require a number of features.
- [x] Support `enum` and `enum` value definitions
- [x] Support `typedef` definitions
- [x] Support function definitions
- [x] Support member function definitions
- [x] Support structs
- [x] Support classes
- [x] Support namespaces
- [x] Add the `syntax` group to `enum` and `enumvalue`
- [x] Add the `syntax` group to `typedef`
- [x] Add the `syntax` group to functions
- [x] Add the `syntax` group to member functions
- [x] Add the `syntax` group to structs
- [x] Add the `syntax` group to classes
- [x] Add the `syntax` group to namespaces
- [x] Change `main()` to generate a YAML file
- [x] Consider grouping all `<simplesect kind="see">` into a single H6 section?

coryan:
I am closing this bug (it has served its purpose) and opening more detailed bugs to track each problem.

PR #11127, opened 2023-03-28
impl(docfx): add uid metadata to pages

coryan:
The YAML files generated from Doxygen pages need some metadata to be linked from other YAML files.

Part of the work for #10895

### DIFF
--- a/docfx/doxygen_pages.cc
+++ b/docfx/doxygen_pages.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "docfx/doxygen_pages.h"
+#include "docfx/config.h"
 #include "docfx/doxygen2markdown.h"
 #include "docfx/doxygen_errors.h"
 #include <algorithm>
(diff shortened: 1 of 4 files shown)


docfx/doxygen_pages.cc
```diff
@@ -78,7 +79,8 @@ namespace docfx {
 //   <xsd:attribute name="abstract" type="DoxBool" use="optional"/>
 // </xsd:complexType>
 // clang-format on
-std::string Page2Markdown(pugi::xml_node const& node) {
+std::string Page2Markdown(Config const& /*config*/,
```
devbww, line 82: I presume this will get used at some point.

devbww (approved):

docfx/doxygen_pages.cc
coryan: When I wrote it: I thought so.  Now that you ask: I am not sure.  Removed.

PR #11134, opened 2023-03-28
ci: split notifications for 'Friends of Cloud C++'

### LINKED ISSUE #10323: Create a build against gRPC's HEAD
We want one build compiling against gRPC's HEAD.

- Should be nightly
- Should be non-blocking
- Can send notification to existing chat (optionally maybe send to another chat room)
- We can use Bazel or CMake (probably does not matter), Bazel is probably "easier".

coryan:
The main motivation is the gRPC @ HEAD build. We want to notify the gRPC team when something breaks. I created a chat room for that purpose, any build labeled "friends" will be sent to that room if there is a failure.

Took the opportunity to modernize this notifier.

Fixes #10323

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11134)


ci/cloudbuild/notifiers/alerts/function/function.cc
```diff
@@ -95,4 +102,14 @@ void SendBuildAlerts(google::cloud::functions::CloudEvent event) {
   auto const chat = MakeChatPayload(bs);
   std::cout << nlohmann::json{{"severity", "INFO"}, {"chat", chat}} << "\n";
   HttpPost(webhook, chat.dump());
+  std::cout << "bs = " << bs.build.dump() << std::endl;
```
dbolduc, line 105: Where do these `cout`s show up?

dbolduc (approved):
I have a feeling that "friends" is convenient for us, but not convenient for our friends.

e.g. if I am Abseil, I probably do not want to get notified every time a Cloud C++ `grpc-at-head` build fails.

Should we s/friends/grpc/ ?

I'm also fine to wait until we have a second friend.

ci/cloudbuild/notifiers/alerts/function/function.cc
coryan: In the log for Cloud Run, where these are deployed.  I think I should remove this one though

coryan:
> I have a feeling that "friends" is convenient for us, but not convenient for our friends.

If they feel strongly: they can create their own builds, notifiers and chat rooms.

> I'm also fine to wait until we have a second friend.

I think we should do that and even then push back.

PR #11131, opened 2023-03-28
impl(bigquery): Logging decorator cleanup - Added DebugString member functions to …

jsrinnn:
…GetJobResponse and ListJobResponse so it can be used in the logging decorator.

Subsequent PR will contain the logging decorator changes that will use the logging from this PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11131)

PR #11141, opened 2023-03-29
impl(bigquery): Logging decorator cleanup (contd) - Adds DebugString() member functions to GetJobRequest and ListJobsRequest

jsrinnn:
As mentioned in the comment for PR-11131 (https://github.com/googleapis/google-cloud-cpp/pull/11131#issuecomment-1489134813),
This PR adds the  DebugString() member functions to GetJobRequest and ListJobsRequest objects.

The logging decorator changes that calls the DebugString for request and response logging will be in a separate PR. I'll send that PR once the PRs for request and responses are merged.

Thank you.


google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
```diff
@@ -177,6 +177,32 @@ TEST(ListJobsRequestTest, OutputStream) {
...
 
+TEST(GetJobRequest, DebugString) {
+  GetJobRequest request("test-project-id", "test-job-id");
+  request.set_location("test-location");
+  std::string expected = absl::StrCat(
+      "GetJobRequest{project_id=", request.project_id(),
+      ", job_id=", request.job_id(), ", location=", request.location(), "}");
+
+  EXPECT_EQ(expected, request.DebugString(TracingOptions{}.SetOptions(
+                          "truncate_string_field_longer_than=1024")));
```
dbolduc, line 188: nit: We should just use the default constructed `TracingOptions{}` unless we actively need the customization.

```cc
EXPECT_EQ(expected, request.DebugString(TracingOptions{}));
```

google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
```diff
@@ -177,6 +177,32 @@ TEST(ListJobsRequestTest, OutputStream) {
   EXPECT_EQ(expected, os.str());
 }
 
+TEST(GetJobRequest, DebugString) {
+  GetJobRequest request("test-project-id", "test-job-id");
+  request.set_location("test-location");
+  std::string expected = absl::StrCat(
+      "GetJobRequest{project_id=", request.project_id(),
+      ", job_id=", request.job_id(), ", location=", request.location(), "}");
```
dbolduc, line 185: optional: there is something strange about writing the same code again in a test.

Consider writing the expected string in full:

```cc
std::string expected =
     "GetJobRequest{project_id=test-project-id, job_id=test-job-id, "
     "location=test-location}";
```

google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
```diff
@@ -177,6 +177,32 @@ TEST(ListJobsRequestTest, OutputStream) {
...
+
+  EXPECT_EQ(expected, request.DebugString(TracingOptions{}.SetOptions(
+                          "truncate_string_field_longer_than=1024")));
+}
+
+TEST(ListJobsRequestTest, DebugString) {
+  auto const request = GetListJobsRequest();
+  std::string all_users = request.all_users() ? "true" : "false";
+  std::string expected = absl::StrCat(
```
dbolduc, line 194: same optional comment about writing out the expected string instead of computing it.

google/cloud/bigquery/v2/minimal/internal/job_request_test.cc
```diff
@@ -177,6 +177,32 @@ TEST(ListJobsRequestTest, OutputStream) {
...
+      "ListJobsRequest{project_id=", request.project_id(),
+      ", all_users=", all_users, ", max_results=", request.max_results(),
+      ", page_token=", request.page_token(),
+      ", projection=", request.projection().value,
+      ", state_filter=", request.state_filter().value,
+      ", parent_job_id=", request.parent_job_id(), "}");
+
+  EXPECT_EQ(expected, request.DebugString(TracingOptions{}.SetOptions(
+                          "truncate_string_field_longer_than=1024")));
```
dbolduc, line 203: same nit about just using `TracingOptions{}`

dbolduc (approved):

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -70,6 +71,25 @@ StateFilter StateFilter::Done() {
   return state_filter;
 }
 
+std::string GetJobRequest::DebugString(TracingOptions const& options) const {
+  std::string out;
+  absl::StrAppend(&out, "GetJobRequest{project_id=", project_id_,
+                  ", job_id=", job_id_, ", location=", location_, "}");
+  return internal::DebugString(out, options);
```
dbolduc, line 78: We want to actually use the TracingOptions. There is one about breaking a message onto multiple lines and another about truncating string fields.

This is an example that I know of. It seems to reveal that our tracing options were "inspired" by protobuf's print settings: https://github.com/googleapis/google-cloud-cpp/blob/952a50ddadb2dfc62536ccc02afb690afaa67ee6/google/cloud/internal/debug_string_protobuf.cc#L67-L85

We want to truncate **individual string fields** in the message and split the fields of the message by newline vs. space.

I think we want something like:
```cc
auto delim = options.single_line_mode() ? " " : "\n";
return absl::StrAppend(&out, "GetJobRequest{", delim, "project_id=", DebugString(project_id_, options), delim, "job_id=", DebugString(job_id_, options), delim, ...., delim, "}");
```

(I did not try to compile this thing.)

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -70,6 +71,25 @@ StateFilter StateFilter::Done() {
...
+std::string ListJobsRequest::DebugString(TracingOptions const& options) const {
+  std::string out;
+  std::string all_users = all_users_ ? "true" : "false";
+  absl::StrAppend(&out, "ListJobsRequest{project_id=", project_id_,
+                  ", all_users=", all_users, ", max_results=", max_results_,
+                  ", page_token=", page_token_,
+                  ", projection=", projection_.value,
+                  ", state_filter=", state_filter_.value,
+                  ", parent_job_id=", parent_job_id_, "}");
+  return internal::DebugString(out, options);
```
dbolduc, line 90: same goes for this one. we need
- a custom delimeter that depends on `options`
- individual string fields to use `DebugString(..., options)`

dbolduc (changes requested):
Sorry for the false approval. I forgot how formatting works. My bad.

devbww:
I'm OOO, so if these comments are inscrutable or insufficient, it will have to wait until next week.

- The `truncate_string_field_longer_than` option should (only) be applied to individual to string fields.
- The `single_line_mode` option should be heeded.

It doesn't look like the `use_short_repeated_primitives` option applies in your case, so it can be ignored.

jsrinnn (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11141)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @dbolduc)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 78 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11141#-NRjLNVn6GpX5or6nP_z-r1-78:-NRoB-y_3qIVS44sh4hD:baygtxk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/99ce0d07412931784292ef9384fda84e4aab92ad/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L78)):*

Done.

Thank you, Darren. The reference was super helpful! 

The code changes I did there is one difference from what you have stated which is that I don't use the member variables directly since internal::DebugString(..) modifies the string passed in.
</blockquote></details>

Actuall my bad. Its a pass by value so it shouldn't be modified. I don't think we need the extra local copies but let me know if you feel otherwise.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -73,35 +73,27 @@ StateFilter StateFilter::Done() {
...
   absl::StrAppend(
-      &out,
-      "ListJobsRequest{project_id=", internal::DebugString(project_id, options),
-      delim, ", all_users=", all_users, delim, ", max_results=", max_results_,
-      delim, ", page_token=", internal::DebugString(page_token, options), delim,
-      ", projection=", internal::DebugString(projection, options), delim,
-      ", state_filter=", internal::DebugString(state_filter, options), delim,
-      ", parent_job_id=", internal::DebugString(parent_job_id, options), delim,
-      "}");
+      &out, "ListJobsRequest{project_id=",
```
dbolduc, line 89: nit: I think we want a `delim` after the curly brace.

dbolduc (approved):

PR #11142, opened 2023-03-29
impl(docfx): a helper to filter the public API

coryan:
Some things are not to be documented, e.g., the private section of a class, or the references to the C++ `std::` namespace that are created by Doxygen.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11142)

### DIFF
--- a/docfx/doxygen2yaml.cc
+++ b/docfx/doxygen2yaml.cc
@@ -42,6 +42,24 @@ std::string Summary(pugi::xml_node const& node) {
 
 }  // namespace
 
+bool IncludeInPublicDocuments(pugi::xml_node const& node) {
+  // We do not generate documents for files and directories.
+  if (kind(node) == "file" || kind(node) == "dir") return false;
+  // Doxygen groups private attributes / functions in <sectiondef> elements of
+  // these kinds:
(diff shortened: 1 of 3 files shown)


docfx/doxygen2yaml.cc
```diff
@@ -42,6 +42,24 @@ std::string Summary(pugi::xml_node const& node) {
...
+  if (kind(node) == "file" || kind(node) == "dir") return false;
+  // Doxygen groups private attributes / functions in <sectiondef> elements of
+  // these kinds:
+  if (kind(node) == "private-attrib" || kind(node) == "private-func") {
+    return false;
+  }
+  // We do not generate documents for types in the C++ `std::` namespace:
+  auto const id = std::string_view{node.attribute("id").as_string()};
+  for (std::string_view prefix : {"namespacestd", "classstd", "structstd"}) {
+    if (id.substr(0, prefix.size()) == prefix) return false;
```
dbolduc, line 56: why not `absl::StartsWith(id, prefix)` ?

coryan: Because I am not using absl:: in this.  Dependencies are liabilities.

PR #11144, opened 2023-03-30
cleanup(bigtable)!: remove DeleteAppProfile overload that can only fail

dbolduc:
This API can only fail because the `ignore_warnings` field cannot be set to true. In my opinion, it is better to avoid it altogether.

Modify the generator so we can omit overloads for non-deprecated functions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11144)


coryan (approved):
FWIW, there are no calls to DeleteAppProfile() from external customers in the last 180 days.

PR #11170, opened 2023-04-04
doc(bigquery): typo in the quickstart path

ziming494:
Should be google rather than google

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11170)

PR #11171, opened 2023-04-04
doc: typo in quickstart paths

dbolduc:
Inspired by #11170 

I left `google/cloud/bigquery/...` alone.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11171)

### DIFF
--- a/generator/internal/scaffold_generator.cc
+++ b/generator/internal/scaffold_generator.cc
@@ -803,7 +803,7 @@ https://cloud.google.com/docs/authentication/production
    the dependencies:
 
    ```bash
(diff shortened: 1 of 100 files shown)


devbww (commented):
> I left google/cloud/bigquery/... alone.

My version includes it (but otherwise I have the same file count).  I would suggest editing it here too, so we're right independent of #11170.

dbolduc:
> My version includes it

Hah, too slow.

> I would suggest editing it here too

Ack, but I'd rather not step on any toes.

devbww (approved):

PR #11172, opened 2023-04-04
impl(bigquery): unify bigquery request/response logging format

devbww:
It is a feature if request/response logging uses a common format no matter whether the message types are protoc generated or not.  So, update the `DebugString()` functions for `GetJobRequest` and `ListJobsRequest` to produce that common format, as if they had been defined as messages like ...
```
message GetJobRequest {
  string project_id = 1;
  string job_id = 2;
  string location = 3;
}

message Projection {
  string value = 1;
}

message StateFilter {
  string value = 1;
}

message ListJobsRequest {
  string project_id = 1;
  bool all_users = 2;
  int32 max_results = 3;
  google.protobuf.Timestamp min_creation_time = 4;
  google.protobuf.Timestamp max_creation_time = 5;
  string page_token = 6;
  Projection projection = 7;
  StateFilter state_filter = 8;
  string parent_job_id = 9;
}
```

In particular, `ListJobsRequest::min_creation_time` and `max_creation_time` are now logged.

Also expand the tests to cover the `TracingOptions`, move some file locals to the anonymous namespace, and remove some unused variables.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11172)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -88,9 +88,11 @@ std::string GetJobRequest::DebugString(TracingOptions const& options) const {
     return absl::StrCat("\"", internal::DebugString(s, options), "\"");
   };
   return absl::StrCat(
-      "google::cloud::bigquery_v2_minimal_internal::GetJobRequest {", sep(1),
-      "project_id: ", quoted(project_id_), sep(1), "job_id: ", quoted(job_id_),
-      sep(1), "location: ", quoted(location_), sep(0), "}");
+      "google::cloud::bigquery_v2_minimal_internal::GetJobRequest {",  //
+      sep(1), "project_id: ", quoted(project_id_),                     //
+      sep(1), "job_id: ", quoted(job_id_),                             //
+      sep(1), "location: ", quoted(location_),                         //
+      sep(0), "}");
 }
 
 std::string ListJobsRequest::DebugString(TracingOptions const& options) const {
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -72,29 +81,40 @@ StateFilter StateFilter::Done() {
...
-      delim, ", parent_job_id=", internal::DebugString(parent_job_id_, options),
-      delim, "}");
-  return out;
+  auto sep = [single_line_mode = options.single_line_mode()](int indent) {
+    return DebugStringSeparator(single_line_mode, indent);
+  };
+  auto quoted = [&options](std::string const& s) {
+    return absl::StrCat("\"", internal::DebugString(s, options), "\"");
+  };
+  return absl::StrCat(
```
coryan, line 103: Consider forcing `sep(N)` to be the beginning of each line:

```cc
  return absl::StrCat(
      "google::cloud::bigquery_v2_minimal_internal::ListJobsRequest {", //
       sep(1), "project_id: ", quoted(project_id_), //
       sep(1), "all_users: ", all_users_ ? "true" : "false", // 
       sep(1), "max_results: ", max_results_, //
       sep(1), "min_creation_time {", //
       sep(2), "\"", internal::FormatRfc3339(min_creation_time_), "\"", //
       sep(1), "}", //
       sep(1), "max_creation_time {", //
       sep(2), "\"", internal::FormatRfc3339(max_creation_time_), "\"", //
       sep(1), "}", //
       sep(1), "page_token: ", quoted(page_token_), //
      sep(1),  "projection {", //
      sep(2), "value: ", quoted(projection_.value), //
      sep(1), "}", //
      sep(1), "state_filter {", //
      sep(2),  "value: ", quoted(state_filter_.value), //
      sep(1), "}", //
      sep(1), "parent_job_id: ", quoted(parent_job_id_), //
      sep(0), "}");
```

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -72,29 +81,40 @@ StateFilter StateFilter::Done() {
...
+      "google::cloud::bigquery_v2_minimal_internal::ListJobsRequest {", sep(1),
+      "project_id: ", quoted(project_id_), sep(1),
+      "all_users: ", all_users_ ? "true" : "false", sep(1),
+      "max_results: ", max_results_, sep(1),  //
+      "min_creation_time {", sep(2), "\"",
+      internal::FormatRfc3339(min_creation_time_), "\"", sep(1), "}", sep(1),
```
coryan, line 109: Does `StrCat()` support `'"'` instead of `"\""`

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -72,29 +81,40 @@ StateFilter StateFilter::Done() {
...
 
 std::string GetJobRequest::DebugString(TracingOptions const& options) const {
-  std::string out;
-  auto const* delim = options.single_line_mode() ? " " : "\n";
-  absl::StrAppend(&out, "GetJobRequest{", delim,
-                  "project_id=", internal::DebugString(project_id_, options),
-                  delim, "job_id=", internal::DebugString(job_id_, options),
-                  delim, internal::DebugString(location_, options), delim, "}");
-  return out;
+  auto sep = [single_line_mode = options.single_line_mode()](int indent) {
```
coryan, line 84: I think it is a certainty that `sep()` and `quoted` will appear in many more functions, any way to refactor them?

google/cloud/bigquery/v2/minimal/internal/job_request.cc
coryan: At the risk of overcomplicating things, we could do something like:

```cc
namespace {
class DebugFormatter {
   DebugFormatter(TracingOptions const& options, absl::string_view message_name)
    : options_(options), buffer_("google::cloud::bigquery_v2_minimal_internal::") {
    Nest(message_name);
   }

  std::string Done() {
    Unnest();
   return std::move(buffer_);
  }

   template<typename T>
   DebugFormatter& Field(absl::string_view name, T value) {
     absl::StrAppend(&buffer_,  Sep(), name, ": ", value);
   return *this;
  }

   DebugFormatter& Quoted(absl::string_view name, std::string const& value) {
     absl::StrAppend(&buffer_,  Sep(), name, ": ", "\"", internal::DebugString(value, options_), "\"");
     return *this;
  }

   DebugFormatter& Quoted(absl::string_value value) {
     absl::StrAppend(&buffer_,  Sep(), name, ": ", "\"", value, "\"");
     return *this;
  }

  DebugFormatter& Nest(absl::string_view name) {
    absl::StrAppend(&buffer, name, " {");
    ++indent_;
    return *this;
  }
  DebugFormatter& Unnest() {
    --indent_;
    absl::StrAppend(&buffer_, Sep(), "}");
    return *this;
  }

private:
  std::string Sep() const {
    return DebugStringSeparator(options.single_line_mode(), indent_);
  }

  TracingOptions options_;
  std::string buffer_;
  int indent_ = 0;
};

} // namespace

...

std::string ListJobsRequest::DebugString(TracingOptions const& options) const {
  return DebugFormatter(options, "ListJobsRequest") //
    .Quoted("project_id", project_id_) //
    .Field("all_users", all_users_ ? "true" : "false") //
    ....
    .Nest("min_creation_time").Quoted(internal::FormatRfc3339(min_creation_time)).Unnest() //
    .Done();
```

coryan (approved):
I think moving `sep(N)` to the beginning of each line makes the code more readable.

The `DebugFormatter` idea is maybe not well thought out, consider it optional, or may a suggestion for future changes.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
devbww: No, I'm afraid not.  https://github.com/abseil/abseil-cpp/blob/master/absl/strings/str_cat.h#L367-L368

google/cloud/bigquery/v2/minimal/internal/job_request.cc
devbww: > Consider forcing sep(N) to be the beginning of each line

Yes, that's much nicer.

> At the risk of overcomplicating things, we could do something like ...
> consider it optional, or may a suggestion for future changes

Yes, given that it appears we should probably apply this in other places, I agree.  I'll look into it as operator.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
devbww: I like your "DebugFormatter" proposal.

PR #11175, opened 2023-04-04
impl(GCS+gRPC): workarounds related to ctype=CORD

coryan:
With Protobuf < v23.x the effect of adding a `[ctype = CORD]` annotation is to make the accessors and modifiers for the field private. We expect that most, but not *all*, customers will use GCS+gRPC with Protobuf >= v23.x. With that version of Protobuf the field accessors and modifiers are public, and they return/consume `absl::Cord`.  To support customers using older Protobuf versions we implement a workaround to gain access to these private accessors and modifiers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11175)


dbolduc (approved):
FTR, the `BypassPrivateControl` business is way over my head.

PR #11182, opened 2023-04-05
impl(common): add formatter for request/response logging

devbww:
Follow up on a @coryan idea from #11172 for a `DebugFormatter` to produce request/response-logging strings for non-protobuf message types.

Use `DebugFormatter` to reimplement the message formatters from #11172.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11182)

### DIFF
--- a/google/cloud/internal/debug_string.h
+++ b/google/cloud/internal/debug_string.h
@@ -15,8 +15,10 @@
 #ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_DEBUG_STRING_H
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_DEBUG_STRING_H
 
+#include "google/cloud/internal/absl_str_cat_quiet.h"
 #include "google/cloud/tracing_options.h"
 #include "google/cloud/version.h"
+#include "absl/strings/string_view.h"
 #include <string>
 
 namespace google {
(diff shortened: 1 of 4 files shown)


google/cloud/internal/debug_string.h
```diff
@@ -15,15 +15,65 @@
...
+    return *this;
+  }
+
+  template <typename T>
+  DebugFormatter& QuotedField(T const& value) {
+    absl::StrAppend(&str_, Sep(), "\"", value, "\"");
+    return *this;
+  }
+
+  DebugFormatter& StringField(absl::string_view field_name, std::string value);
```
coryan, line 63: This member function name is slightly odd, but I am not sure I have better ideas.

coryan (approved):

google/cloud/internal/debug_string.h
devbww: I don't have any better ideas at the moment either.  Given that it is supposed to differentiate fields for which `truncate_string_field_longer_than` should apply, this seemed like a good first approximation.

google/cloud/internal/debug_string.h
```diff
@@ -15,15 +15,65 @@
...
 namespace internal {
 
+/**
+ * Build strings for use in request/response logging.
+ *
+ * The intent is to produce strings with a format similar to those returned
+ * by `DebugString(google::protobuf::Message const&, TracingOptions const&)`
+ * from `google/cloud/internal/debug_string_protobuf.h` for a proto message.
+ */
+class DebugFormatter {
```
devbww, line 36: I think we might want some easy way to compose `DebugFormatter`s for sub-messages, but that can wait for another day.

PR #11183, opened 2023-04-05
feat(pubsublite): publish idempotence

tmdiep:
Implements publish idempotence (default enabled), where the server will ensure that unique messages within a single publisher session are only stored once.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11183)


google/cloud/pubsublite/internal/partition_publisher.cc
```diff
@@ -176,7 +176,7 @@ void PartitionPublisher::OnRead(absl::optional<PublishResponse> response) {
               return a.start_index() < b.start_index();
             });
   int range_idx = 0;
-  for (int msg_idx = 0; msg_idx < batch.size(); ++msg_idx) {
+  for (int msg_idx = 0; msg_idx < static_cast<int>(batch.size()); ++msg_idx) {
```
coryan, line 179: Please do the opposite:

```cc
for (std::size_t msg_idx = 0; msg_idx != batch.size(); ++msg_idx) {
```

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -161,7 +170,7 @@ StatusOr<std::unique_ptr<PublisherConnection>> MakePublisherConnection(
               [] { return std::make_unique<StreamRetryPolicy>(); },
               backoff_policy, sleeper, stream_factory, std::move(initializer));
         },
-        batching_options, std::move(request), alarm_registry);
+        batching_options, std::move(request), 0, alarm_registry);
```
coryan, line 173: `s;0, ;/*initial_sequence=*/0,;`

google/cloud/pubsublite/internal/partition_publisher.h
```diff
@@ -46,7 +48,7 @@ class PartitionPublisher
           StreamInitializer<google::cloud::pubsublite::v1::PublishRequest,
                             google::cloud::pubsublite::v1::PublishResponse>)>,
       BatchingOptions, google::cloud::pubsublite::v1::InitialPublishRequest,
-      AlarmRegistry&);
+      PublishSequenceNumber, AlarmRegistry&);
```
coryan, line 51: it seems that only tests initialize `PulishSequenceNumber` with something other than 0.  Consider making this the last argument and maybe passing 0 by default.

google/cloud/pubsublite/internal/partition_publisher.cc
```diff
@@ -70,7 +73,8 @@ future<StatusOr<Cursor>> PartitionPublisher::Publish(PubSubMessage m) {
   if (!status.ok()) {
     return make_ready_future(StatusOr<Cursor>(std::move(status)));
   }
-  MessageWithPromise unbatched{std::move(m), promise<StatusOr<Cursor>>{}};
+  MessageWithPromise unbatched{std::move(m), next_sequence_++,
```
coryan, line 76: Is there a reason to use post-fix ++?  If so, please add a comment, we rarely use them.

google/cloud/pubsublite/internal/partition_publisher.cc
```diff
@@ -159,11 +168,30 @@ void PartitionPublisher::OnRead(absl::optional<PublishResponse> response) {
...
-  std::int64_t offset = response->message_response().start_cursor().offset();
-  for (auto& message : batch) {
+
+  // ensure cursor ranges are sorted by increasing message batch index
+  auto* ranges = response->mutable_message_response()->mutable_cursor_ranges();
+  std::sort(ranges->begin(), ranges->end(),
+            [](CursorRange const& a, CursorRange const& b) {
+              return a.start_index() < b.start_index();
+            });
+  int range_idx = 0;
```
coryan, line 178: It would help if you could explain what these loops are trying to do.  I am sure it is obvious to somebody familiar with Pub/Sub Lite, but it is hard to glean what is the intent, and therefore hard to offer useful comments.

google/cloud/pubsublite/internal/partition_publisher.cc
tmdiep: I did that initially and there were more errors because `ranges->Get(range_idx)` accepts an int. This requires the least number of static_casts.

coryan: Ack.

google/cloud/pubsublite/publisher_connection.cc
coryan: Or better yet, make it so you don't need to pass that argument.

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -147,10 +148,18 @@ StatusOr<std::unique_ptr<PublisherConnection>> MakePublisherConnection(
   auto publisher_service_stub = CreateDefaultPublisherServiceStub(cq, opts);
   auto batching_options = MakeBatchingOptions(opts);
 
+  bool enable_idempotence = true;
```
coryan, line 151: In all other libraries we do something like (paraphrasing):

```
std::unique_ptr<PublisherConnection> MakePublisherConnection(
    Topic topic, Options opts) {
  opts = ApplyDefaults(std::move(opts));
```

for example:

https://github.com/googleapis/google-cloud-cpp/blob/e11ad2c3a59eeff9ddb9f435fe71b4f89672a9a4/google/cloud/pubsublite/admin_connection.cc#L167-L173

and then we do not need to worry about what options are set.  Please consider *setting* this option in `opts` so a future refactoring is easier:

google/cloud/pubsublite/publisher_connection.cc
```diff
@@ -147,10 +148,18 @@ StatusOr<std::unique_ptr<PublisherConnection>> MakePublisherConnection(
...
+    enable_idempotence = opts.get<EnablePublishIdempotenceOption>();
+  }
+  std::string client_id;
+  if (enable_idempotence) client_id = pubsublite_internal::GenerateClientId();
+
   auto partition_publisher_factory = [=](std::uint32_t partition) {
     InitialPublishRequest request;
     request.set_topic(topic.FullName());
     request.set_partition(partition);
+    request.set_client_id(client_id);
```
coryan, line 162: nit: `requet.set_client_id(std::move(client_id));`

google/cloud/pubsublite/options.h
```diff
@@ -63,6 +63,15 @@ struct PublishFlushAlarmPeriodOption {
   using Type = std::chrono::milliseconds;
 };
 
+/**
+ * Whether idempotence is enabled, where the server will ensure that unique
```
coryan, line 67: The first line in these comments is the "brief" description and ideally should fit in a single line.  Consider rephrasing.  You may also want to reference the cloud site docs, e.g.:

```cc
  * 
  * @see https://cloud.google.com/pubsub/lite/docs/publishing#publish_idempotency
  */
```

coryan (commented):
A few nits here and there.  The business with grouping the batches needs some explanation, I do not know what the intent is.

google/cloud/pubsublite/google_cloud_cpp_pubsublite.bzl
```diff
@@ -35,6 +35,7 @@ google_cloud_cpp_pubsublite_hdrs = [
     "internal/alarm_registry.h",
     "internal/alarm_registry_impl.h",
     "internal/batching_options.h",
+    "internal/client_id_generator.h",
```
coryan, line 38: The source of truth for these `.bzl` files is CMake.  You need to add these here:

https://github.com/googleapis/google-cloud-cpp/blob/e11ad2c3a59eeff9ddb9f435fe71b4f89672a9a4/google/cloud/pubsublite/CMakeLists.txt#L47

You may want to read

https://github.com/googleapis/google-cloud-cpp/blob/main/doc/contributor/working-with-bazel-and-cmake.md

coryan:
FWIW, I am waiting for your changes before running the tests again.

tmdiep:
@coryan Ack, thanks for all the comments. I've just been distracted by other issues, so haven't had time to address them.

coryan:
I think we should close this PR.  You can reopen it when you have more time.

tmdiep:
Agreed. Putting this PR on hold for now.

Issue #11156, opened 2023-04-02
Add Opentelemetry Cloud Trace Exporter


dbolduc:
To summarize an internal document:

- dir: `google/cloud/opentelemetry/...`
- class: `google::cloud::otel::TraceExporter`
  - use namespace `otel` just to avoid conflicts with `::opentelemetry`.
- feature: `experimental-opentelemetry_sdk`
- ctor: `TraceExporter(Options options)`

Prototype is here: https://github.com/dbolduc/google-cloud-cpp/tree/ot-tracing-design-doc/google/cloud/trace/exporter

Of note, we need to translate the name of some span attributes into names that appease Cloud Trace. There is an internal doc which has the mappings. (If you can sift through the TODO's, unanswered comments, and scary red highlighting).

Also, several methods are unimplemented.

Implementation steps:
- [ ] Set up directory, feature, basic building
  - [ ] Install library. Might benefit from the remainder of #10795 
  - [ ] Include in docs
- [ ] Add Recordable class
- [ ] Add TraceExporter class
- [ ] Add TracerProvider factory function
- [ ] Write Integration test
  - [ ] Generate Cloud Trace v1. [We need it to list traces](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v1#listtracesrequest). Aside: We might want to check other vN > 1 services to see if they are replacements or additions to the v1 service....
- [ ] Add sample

dbolduc:
We decided not to filter out any options supplied to the exporter, so this is just done.

PR #11192, opened 2023-04-06
impl(otel): add exporter directory, feature

dbolduc:
Part of the work for #11156 

Introduce a new library that is built with `GOOGLE_CLOUD_CPP_ENABLE=experimental-opentelemetry_sdk`. This will house our OpenTelemetry exporters. It depends on the OpenTelemetry SDK, and the necessary google-cloud-cpp libraries.

The one difference from the bikeshed spec is that I will use the namespace `google::cloud::otel` instead of `google::cloud::opentelemetry`. I just want to avoid all the `::opentelemetry` confusion.

Add the shell of the `Recordable` class, just to verify that our dependencies work.

Installing and Doxygen will come next.

We edit the Cloud Trace quickstart target so that it will not build in the `noex-pr` build. Maybe this is lazy. In the long run, we probably want all of our "mostly generated" `CMakeLists.txt` to look the same.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11192)


dbolduc:
Notes for my future self because I am not thinking super clearly:

---

`noex-pr` fails because we build the `trace-quickstart` which can have exceptions.

https://github.com/googleapis/google-cloud-cpp/blob/fa11ef09ec85c7f9b9722cce19e2dcba3d41a943/google/cloud/trace/CMakeLists.txt#L119-L120

The difference is that `trace` uses `if (BUILD_TESTING)` and things like bigtable/storage use `if (GOOGLE_CLOUD_CPP_ENABLE_EXAMPLES)` to turn on the quickstart.

https://github.com/googleapis/google-cloud-cpp/blob/fa11ef09ec85c7f9b9722cce19e2dcba3d41a943/google/cloud/storage/CMakeLists.txt#L46-L47

The quickest fix is to just use `GOOGLE_CLOUD_CPP_ENABLE_EXAMPLES` in `trace`.

---

`m32-pr` fails because opentelemetry-cpp is not installed. If I copy the other installs, it will go into `/usr/local/lib64/`. If I add the toolchain defines, it starts to build.

https://github.com/googleapis/google-cloud-cpp/blob/fa11ef09ec85c7f9b9722cce19e2dcba3d41a943/ci/etc/m32-toolchain.cmake#L19-L30

Then I get a new breed of otel + `absl::variant` issues. Yay?

```
/usr/bin/ld: /usr/local/lib/libopentelemetry_exporter_in_memory.so: undefined reference to `absl::lts_20220623::variant_internal::ThrowBadVariantAccess()'
```

ci/cloudbuild/builds/lib/features.sh
```diff
@@ -28,6 +28,7 @@ function features::always_build() {
     # These have hand-crafted code, therefore we always want to build them.
     bigtable
     experimental-bigquery_rest
+    experimental-opentelemetry_sdk
```
coryan, line 31: I think I like `opentelemetry_exporters` better than `*_sdk`.

coryan (approved):
We can always rename this library later, but `SDK` suggest "a library to develop stuff for OpenTelemetry", not "a library containing OpenTelemetry exporters"

ci/cloudbuild/builds/lib/features.sh
dbolduc: The only thing stopping me is that the resource detector is not an exporter. What you suggest may still be better, though.

`opentelemetry_sdk` is named after "OpenTelemetry SDK". We would be perpetuating their fraught naming scheme.

dbolduc: I'll take your suggestion and go with `opentelemetry_exporters` for now.

dbolduc: Nevermind. I am merging and adding a discussion item to next meeting to sort out the names.

In the long run, we might just want to call the feature `opentelemetry`. The problem is that I am currently using `experimental-opentelemetry` to mean: compile the libraries with OpenTelemetry.

PR #11195, opened 2023-04-06
impl(generator): add DiscoveryFile

scotthart:
part of the work for #10980 

`DiscoveryFile::WriteFile` will be tested later in an integration test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195)


generator/internal/discovery_file.cc
```diff
@@ -0,0 +1,116 @@
...
+#else
+#include <sys/stat.h>
+#endif  // _WIN32
+
+namespace google {
+namespace cloud {
+namespace generator_internal {
+namespace {
+
+void MakeDirectory(std::string const& path) {
```
coryan, line 33: We also have one of these in:

https://github.com/googleapis/google-cloud-cpp/blob/0633e0deda8d91eda85ee12a597e08f3d01efd24/generator/internal/scaffold_generator.cc#L144-L150

Maybe it is time to refactor them to a [`create_directory()`](https://en.cppreference.com/w/cpp/filesystem/create_directory) in 

https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/internal/filesystem.h
https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/internal/filesystem.cc

coryan (commented):
A drive-by on the MakeDirectory() thing, I have not been following these changes in enough detail to comment on the rest.

generator/internal/discovery_file.cc
devbww: Whichever way you go, 777 is usually the right choice for the mode (unless there are special security concerns) so that the umask has a chance to work.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @coryan and @devbww)

___
*[`generator/internal/discovery_file.cc` line 33 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#gh-1160315555:-NSRJHRcAIX9lZMS0acd:b-1tl1oe) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/066ea20d2cf24280d192aa4071221a03049af9ef/generator/internal/discovery_file.cc#L33)):*

Whichever way you go, 777 is usually the right choice for the mode (unless there are special security concerns) so that the umask has a chance to work.
</blockquote></details>

My thought process when copying the function from scaffold_generator.cc was that my version introduces a 2nd copy and that if we end up with a 3rd copy somewhere, I would refactor it.

I will update the permissions.

coryan (commented):
Delaying the refactoring SGTM.  I still don't feel like I can comment on the rest.

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-:-NSX18Rt7ugasSyQPLYf:bnyycgk)* status: 1 of 8 files reviewed, all discussions resolved (waiting on @devbww)

dbolduc (commented):
Mostly due-diligence questions that aren't blocking or anything.

And a suggestion for how to manage the newlines between sections of the proto.

> Reviewed 6 of 8 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-:-NSgl6-X3bB68-GqRPeT:bvhbh6g)* status: all files reviewed, 7 unresolved discussions (waiting on @scotthart)

___
*[`generator/internal/discovery_file.h` line 21 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-NSgSoAG-KVcdhD1ZHLh:-NSgSoAG-KVcdhD1ZHLi:bdmu0lj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c98a7ad92025622d09c4368739fc8eac7f0024a/generator/internal/discovery_file.h#L21)):*
> ```objc
> #include "generator/internal/discovery_type_vertex.h"
> #include "google/cloud/status.h"
> #include <nlohmann/json.hpp>
> ```

unneeded by the header?
___
*[`generator/internal/discovery_file.h` line 34 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-NSgbJct2PZrqPu2pknV:-NSgbJct2PZrqPu2pknW:bv10odr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c98a7ad92025622d09c4368739fc8eac7f0024a/generator/internal/discovery_file.h#L34)):*
> ```objc
>  public:
>   DiscoveryFile() = default;
>   DiscoveryFile(DiscoveryResource const* resource, std::string file_path,
> ```

Can we document what `resource == nullptr` means? Something like:

"// Set `resources == nullptr` to define message types without a service"
___
*[`generator/internal/discovery_file.cc` line 33 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#gh-1160315555:-NSgU2NjCzE7nMphXW8v:bd8q1dj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/066ea20d2cf24280d192aa4071221a03049af9ef/generator/internal/discovery_file.cc#L33)):*

My thought process when copying the function from scaffold_generator.cc was that my version introduces a 2nd copy and that if we end up with a 3rd copy somewhere, I would refactor it.

I will update the permissions.
</blockquote></details>

In general, I would vote for refactoring when N == 2, not when N > 2.

.... but that's just like, my opinion, man.
___
*[`generator/internal/discovery_file.cc` line 62 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-NSg_2FPBiZwTV9Cp1PS:-NSg_2FPBiZwTV9Cp1PT:bcr52p2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c98a7ad92025622d09c4368739fc8eac7f0024a/generator/internal/discovery_file.cc#L62)):*
> ```cpp
>       {"resource_name", (resource_ ? resource_->name() : "")}};
>   google::protobuf::io::OstreamOutputStream output(&output_stream);
>   google::protobuf::io::Printer printer(&output, '$');
> ```

So why don't we want to use our `generator/internal/printer.h`? Is it that the `GeneratorContext` is not a thing for this work?

Similarly, is there a difference between `OstreamOutputStream` and `ZeroCopyOutputStream`? Is it just a minor performance difference that we don't care about?



_Code quote:_
```cpp
  google::protobuf::io::OstreamOutputStream output(&output_stream);
  google::protobuf::io::Printer printer(&output, '$');
```
___
*[`generator/internal/discovery_file.cc` line 68 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-NSgfeXO1-4mOJnXXwAB:-NSgfeXO1-4mOJnXXwAC:b-v39b1s) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c98a7ad92025622d09c4368739fc8eac7f0024a/generator/internal/discovery_file.cc#L68)):*
> ```cpp
> 
> package $package_name$;
> ```

If we remove this newline, we could print a leading newline for:

- the import section (if there is one)
- the service definition (if there is one)
- each message

If we do that we can just print the messages as we get them, instead of collecting them in `formatted_types`.
___
*[`generator/internal/discovery_file.cc` line 72 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-NSgVjXRAfyixBBQq0QM:-NSgVjXRAfyixBBQq0QN:b6myhlj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c98a7ad92025622d09c4368739fc8eac7f0024a/generator/internal/discovery_file.cc#L72)):*
> ```cpp
> 
>   if (!import_paths_.empty()) {
>     printer.Print(vars, absl::StrJoin(import_paths_, "\n",
> ```

I would suggest writing the loop instead of using `absl::StrJoin` to iterate:

```cc
for (auto const& path : import_paths_) {
  printer.Print(vars, absl::StrFormat("import \"%s\";\n", path).c_str());
}
printer.Print("\n"); // or move this in front of the for loop if you take my suggestion to remove the newline after "package ..."
```
___
*[`generator/internal/discovery_file.cc` line 93 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-NSgdGB20SJWsV2Yn41P:-NSgdGB20SJWsV2Yn41Q:b-i9i8za) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c98a7ad92025622d09c4368739fc8eac7f0024a/generator/internal/discovery_file.cc#L93)):*
> ```cpp
>   for (auto const& t : types_) {
>     auto message = t->JsonToProtobufMessage();
>     if (!message) return message.status();
> ```

nit: std::move(message).status();
___
*[`generator/internal/discovery_resource.h` line 37 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-NSgjOVJEpHYWvObgCzT:-NSgjOVK1iGQ3Bni_6Lh:b-q5s1o9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c98a7ad92025622d09c4368739fc8eac7f0024a/generator/internal/discovery_resource.h#L37)):*
> ```objc
>   void AddRequestType(std::string name, DiscoveryTypeVertex const* type);
> 
>   std::vector<DiscoveryTypeVertex const*> GetRequestTypesList() const;
> ```

Can these Request types repeat across `DiscoveryResource`s ?

I am guessing they can, and we deal with it by having `types_` as a `set`. If so, can we add a test case for a repeated `DiscoveryTypeVertex` ?

dbolduc (approved):
> Reviewed 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @coryan and @scotthart)

___
*[`generator/internal/discovery_file.cc` line 33 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#gh-1160315555:-NShIyQ49MXDBMNf-uNi:br1tv4z) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/066ea20d2cf24280d192aa4071221a03049af9ef/generator/internal/discovery_file.cc#L33)):*

Not my circus, not my monkeys: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)
</blockquote></details>

Two can play at that game: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself
___
*[`generator/internal/discovery_file.cc` line 72 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-NSgVjXRAfyixBBQq0QM:-NShJXoL2Ilkr_5m1Fg9:bw50fdn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c98a7ad92025622d09c4368739fc8eac7f0024a/generator/internal/discovery_file.cc#L72)):*

I prefer using `StrJoin(text, "\n")` instead of using raw loops, much in the same fashion as using `std::findif` or `std::transform` instead of a raw loop. I know I'll always get back a block of text with newlines between the items and no newline after the last item. It helps prevent off by one errors where I end up with too many newlines or not enough. Considering this is all in preparation to perform some expensive file i/o, I am unconcerned about any performance losses from the associated memcopies being performed. Furthermore, as this is not user code and is only executed infrequently by us, I don't feel that investing in optimizing this code is worth our time.
</blockquote></details>

I do not care about optimizing for performance; I care about optimizing for readability. And I think what I suggested is easier to read.

I approved the PR. You can make the changes, or not.

Aside: I think my suggestions are being dismissed a little too easily. I may start making demands instead. 👻
___
*[`generator/internal/discovery_resource.h` line 37 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11195#-NSgjOVJEpHYWvObgCzT:-NShO6nREcphFA4n1k40:be17u8d) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4c98a7ad92025622d09c4368739fc8eac7f0024a/generator/internal/discovery_resource.h#L37)):*

The request types are constructed and assigned to `DiscoveryResource`s in such a fashion that there are not any duplicates.
</blockquote></details>

Ack, thanks.

PR #11196, opened 2023-04-06
ci: build otel with m32

dbolduc:
Motivated by #11156 

opentelemetry-cpp does not have a binary package we can just install. So we have to build the code using our m32 toolchain.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11196)

### DIFF
--- a/ci/cloudbuild/dockerfiles/fedora-m32.Dockerfile
+++ b/ci/cloudbuild/dockerfiles/fedora-m32.Dockerfile
@@ -100,11 +100,21 @@ RUN curl -sSL https://github.com/nlohmann/json/archive/v3.11.2.tar.gz | \
     cmake --build cmake-out --target install -- -j ${NCPU:-4} && \
     ldconfig
 
+# There is no Fedora package for opentelemetry-cpp, so we must build it
+# from source. We use C++17 to match the language standard that Abseil is
+# compiled with. And we use the compiler flags from ci/etc/m32-toolchain.cmake
+# to force a 32-bit install.
 WORKDIR /var/tmp/build/opentelemetry
 RUN curl -sSL https://github.com/open-telemetry/opentelemetry-cpp/archive/v1.8.3.tar.gz | \
     tar -xzf - --strip-components=1 && \
     cmake \
-        -DCMAKE_CXX_STANDARD=20 \
+        -DCMAKE_CXX_STANDARD=17 \
+        -DCMAKE_CXX_COMPILER=g++ \
+        -DCMAKE_CXX_FLAGS=-m32 \
+        -DCMAKE_FIND_ROOT_PATH=/usr/ \
+        -DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER \
+        -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY \
+        -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY \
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/dockerfiles/fedora-m32.Dockerfile
```diff
@@ -100,6 +100,28 @@ RUN curl -sSL https://github.com/nlohmann/json/archive/v3.11.2.tar.gz | \
...
+        -DWITH_ABSEIL=ON \
+        -DBUILD_TESTING=OFF \
+        -DOPENTELEMETRY_INSTALL=ON \
+        -DCMAKE_CXX_COMPILER=g++ \
+        -DCMAKE_CXX_FLAGS=-m32 \
+        -DCMAKE_FIND_ROOT_PATH=/usr/ \
+        -DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER \
+        -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY \
+        -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY \
+        -H. -Bcmake-out && \
```
coryan, line 121: I think we started using `-S . -B cmake-out` a while ago?

ci/cloudbuild/dockerfiles/fedora-m32.Dockerfile
```diff
@@ -100,6 +100,28 @@ RUN curl -sSL https://github.com/nlohmann/json/archive/v3.11.2.tar.gz | \
     cmake --build cmake-out --target install -- -j ${NCPU:-4} && \
     ldconfig
 
+WORKDIR /var/tmp/build/opentelemetry
+RUN curl -sSL https://github.com/open-telemetry/opentelemetry-cpp/archive/v1.8.3.tar.gz | \
+    tar -xzf - --strip-components=1 && \
+    cmake \
+        -DCMAKE_CXX_STANDARD=20 \
```
coryan, line 107: Why is this compiled with C++20 but the other things in this file are not?

dbolduc: I need to match the standard that absl is compiled with. I do not know what it is, but I know that building opentelemetry-cpp with C++14 gave me this error:

```
/usr/bin/ld: /usr/local/lib/libopentelemetry_exporter_in_memory.so: undefined reference to `absl::lts_20220623::variant_internal::ThrowBadVariantAccess()'
```

I tried C++20, and it worked, so I moved on and didn't look back.

coryan: Ah.... Abseil is built with the default C++ version on that platform (C++17).  OpenTelemetry is probably forcing a -DCMAKE_CXX_STANDARD=... on us with no need to do so.

coryan (approved):

ci/cloudbuild/dockerfiles/fedora-m32.Dockerfile
dbolduc: Fixed.

ci/cloudbuild/dockerfiles/fedora-m32.Dockerfile
dbolduc: Thanks, I switched the standard to C++17.

I also added some comments so we don't have to dig up this PR to know why certain choices were made.

PR #11197, opened 2023-04-07
impl(common): enhance DebugFormatter support

devbww:
- composable support for sub-messages: `SubMessage()` can now use the `DebugString()` of the sub field to build the upper `DebugString()`. This requires `DebugString()` to now take a name, which should either be the top-most class name or an inner field name.

- direct support for `time_point` fields: Match the auto formatting of `Timestamp` proto messages as human-readable strings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11197)

### DIFF
--- a/google/cloud/internal/debug_string_test.cc
+++ b/google/cloud/internal/debug_string_test.cc
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 #include "google/cloud/internal/debug_string.h"
+#include "absl/strings/string_view.h"
 #include <gmock/gmock.h>
+#include <string>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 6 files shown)


coryan (approved):

google/cloud/internal/debug_string_test.cc
```diff
@@ -13,73 +13,79 @@
...
                 .Build(),
             R"(message_name {)"
             R"( field1: 42)"
             R"( sub_message {)"
-            R"( field2: "3.14159")"
+            R"( value: 3.14159)"
             R"( })"
             R"( field2: "foobar")"
             R"( field3: true)"
             R"( })");
```
dbolduc, line 51: I would expect this line to not have a leading space.

google/cloud/internal/debug_string_test.cc
```diff
@@ -13,73 +13,79 @@
...
-                .EndMessage()
+                .SubMessage("sub_message", SubMessage{3.14159})
                 .StringField("field2", "foobar")
                 .Field("field3", true)
                 .Build(),
             R"(message_name {)"
             R"( field1: 42)"
             R"( sub_message {)"
-            R"( field2: "3.14159")"
+            R"( value: 3.14159)"
```
dbolduc, line 47: I would expect one extra space in the indent. (But I also don't know what protobuf does, and if we are just copying their format)

dbolduc (commented):
The output isn't *exactly* what I would expect. I say nothing about how to make the changes, though. You can decide if (1) it is a problem (2) it is worth fixing.

google/cloud/internal/debug_string_test.cc
devbww: There is always a space after the `{` in single-line mode.

google/cloud/internal/debug_string_test.cc
devbww: I'm guessing you're expecting _"single_line_mode=F"_ output, but this is not it.

dbolduc: Yep.... My mistake.

PR #11199, opened 2023-04-07
feat(workstations): generate library

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11199)


google/cloud/workstations/v1/workstations_connection_idempotency_policy.cc
```diff
@@ -0,0 +1,148 @@
...
+}
+
+Idempotency WorkstationsConnectionIdempotencyPolicy::StopWorkstation(
+    google::cloud::workstations::v1::StopWorkstationRequest const&) {
+  return Idempotency::kNonIdempotent;
+}
+
+Idempotency WorkstationsConnectionIdempotencyPolicy::GenerateAccessToken(
+    google::cloud::workstations::v1::GenerateAccessTokenRequest const&) {
+  return Idempotency::kNonIdempotent;
```
dbolduc, line 137: this one could be idempotent. https://github.com/googleapis/googleapis/blob/f0cec0b201737db7e99ff1bc6dac09e16ed2ce65/google/cloud/workstations/v1/workstations_grpc_service_config.json#L13

"I don't want to micromanage these things" is also a reasonable response.

dbolduc (approved):

google/cloud/workstations/v1/workstations_connection_idempotency_policy.cc
devbww: I don't think we should even try to special-case things, except in egregious (and probably customer-raised) instances.  So, I'll leave it alone for now.  Thanks.

PR #11200, opened 2023-04-07
feat(common): a "lastN in thread" logger

coryan:
This is a variant of the "lastN" logger. It only dumps the lastN log lines *in the thread* when the library or application creates a log line over a minimum severity threshold. This is an improvement over the "lastN" logger, which dumps the last N logs from all the threads in similar circumstances.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11200)

### DIFF
--- a/google/cloud/internal/log_impl_test.cc
+++ b/google/cloud/internal/log_impl_test.cc
@@ -83,8 +83,8 @@ TEST(PerThreadCircularBufferBackend, Basic) {
 
   auto noise_generator = [&] {
     for (int i = 0; i != 10; ++i) {
-      buffer.ProcessWithOwnership(test_log_record(
-          Severity::GCP_LS_INFO, "noise " + std::to_string(i)));
+      buffer.ProcessWithOwnership(
+          test_log_record(Severity::GCP_LS_INFO, "noise " + std::to_string(i)));
     }
   };
 
--- a/google/cloud/internal/log_impl_test.cc
+++ b/google/cloud/internal/log_impl_test.cc
@@ -47,7 +47,7 @@ TEST(CircularBufferBackend, Basic) {
   buffer.ProcessWithOwnership(
       test_log_record(Severity::GCP_LS_WARNING, "msg 4"));
   buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_INFO, "msg 5"));
-  // The output should be empty, as no message with high enough severity have
+  // The output should be empty as no message with high enough severity has
   // been generated.
   EXPECT_THAT(be->ExtractLines(), IsEmpty());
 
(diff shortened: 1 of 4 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11200)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

google/cloud/internal/log_impl_test.cc
```diff
@@ -61,6 +61,35 @@ TEST(CircularBufferBackend, Basic) {
   EXPECT_THAT(be->ExtractLines(), ElementsAre("msg 9", "msg 10"));
 }
 
+TEST(PerThreadCircularBufferBackend, Basic) {
+  auto be = std::make_shared<ScopedLog::Backend>();
+  PerThreadCircularBufferBackend buffer(3, Severity::GCP_LS_ERROR, be);
```
devbww, line 66: Opt:  It took me a while to grok the significance of the `3` in the expectations (here, and in the `CircularBufferBackend` version).  Perhaps using a suitably-named variable could alleviate some of that?

coryan: Done. Also add some comments and make the per-thread test more interesting.

google/cloud/internal/log_impl_test.cc
```diff
@@ -46,21 +47,84 @@ TEST(CircularBufferBackend, Basic) {
   buffer.ProcessWithOwnership(
       test_log_record(Severity::GCP_LS_WARNING, "msg 4"));
   buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_INFO, "msg 5"));
+  // The output should be empty, as no message with high enough severity have
```
devbww, line 50: s/empty, as/empty as/
s/have/has/

google/cloud/internal/log_impl_test.cc
```diff
@@ -46,21 +47,84 @@ TEST(CircularBufferBackend, Basic) {
...
+  // been generated.
   EXPECT_THAT(be->ExtractLines(), IsEmpty());
 
+  // This should flush the messages, and only the last kBufferSize messages
+  // should be in the output.
   buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_ERROR, "msg 6"));
   EXPECT_THAT(be->ExtractLines(), ElementsAre("msg 4", "msg 5", "msg 6"));
 
+  // In this case, the buffer is not full. But a message with high enough
```
devbww, line 59: s/case, the/case the/
s/full. But/full, but/

google/cloud/internal/log_impl_test.cc
```diff
@@ -46,21 +47,84 @@ TEST(CircularBufferBackend, Basic) {
...
+  // severity forces a flush.
   buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_INFO, "msg 7"));
   buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_ERROR, "msg 8"));
   EXPECT_THAT(be->ExtractLines(), ElementsAre("msg 7", "msg 8"));
 
+  // In this case, the buffer is not full. An explicit flush should create
```
devbww, line 65: s/case, the/case the/
s/full. An/full, but an/

google/cloud/internal/log_impl_test.cc
```diff
@@ -46,21 +47,84 @@ TEST(CircularBufferBackend, Basic) {
...
+  };
+
+  auto noise = std::thread{noise_generator};
+  buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_INFO, "msg 1"));
+  buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_DEBUG, "msg 2"));
+  buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_TRACE, "msg 3"));
+  buffer.ProcessWithOwnership(
+      test_log_record(Severity::GCP_LS_WARNING, "msg 4"));
+  buffer.ProcessWithOwnership(test_log_record(Severity::GCP_LS_INFO, "msg 5"));
+  // The output should be empty, as no message with high enough severity have
```
devbww, line 98: Dittos.

devbww (approved):

devbww:
For my own clarification, I presume ...

1. that this isn't really applicable to us as library purveyors, but rather to users, who know about their threads and where they may need to flush, and
2. it is less useful for async/callback-based operations, where some of the sub-actions will be occurring in different threads.

Perhaps one way to address both concerns in the future might be the ability to apply a backend as a call `Option`.  That would be independent of threads, and so be applicable at user-defined granularity.  It would also collect all the callback-based logging through the existing "current options" support.

Or, perhaps that is all just re-inventing telemetry stuff.

PR #11202, opened 2023-04-07
fix(rest): remove newlines from 'status' header

coryan:
The HTTP status line is returned as one of the headers. We were not removing the trailing newline from it, resulting in log split across multiple lines.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11202)

### DIFF
--- a/google/cloud/internal/curl_wrappers.cc
+++ b/google/cloud/internal/curl_wrappers.cc
@@ -14,10 +14,12 @@
 
 #include "google/cloud/internal/curl_wrappers.h"
 #include "google/cloud/internal/absl_str_cat_quiet.h"
+#include "google/cloud/internal/absl_str_replace_quiet.h"
 #include "google/cloud/internal/curl_options.h"
 #include "google/cloud/internal/throw_delegate.h"
 #include "google/cloud/log.h"
 #include "absl/strings/match.h"
+#include "absl/strings/str_split.h"
 #include <openssl/crypto.h>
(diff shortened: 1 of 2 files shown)


google/cloud/internal/curl_wrappers.cc
```diff
@@ -242,16 +244,14 @@ std::size_t CurlAppendHeaderData(CurlReceivedHeaders& received_headers,
...
-    header_value = std::string(separator + 2, data + size - 2);
-  }
-  std::transform(header_name.begin(), header_name.end(), header_name.begin(),
-                 [](unsigned char x) { return std::tolower(x); });
-  received_headers.emplace(std::move(header_name), std::move(header_value));
+  // Split on the ':' (if any), ignore the trailing '\r\n'. Then cleanup the
+  // header name and value.
+  std::vector<absl::string_view> s = absl::StrSplit(
+      absl::string_view{data, size - 2}, absl::MaxSplits(':', 1));
+  auto value = s.size() == 2 ? absl::string_view(s[1]) : absl::string_view{};
```
devbww, line 251: `s[1]` is already a `string_view`, yes?

devbww (approved):

google/cloud/internal/curl_wrappers.cc
coryan: Thanks, fixed.

PR #11207, opened 2023-04-07
impl(otel): docs for opentelemetry_sdk

dbolduc:
Part of the work for #11156 

Generate documentation for `google/cloud/opentelemetry`. This is what the docs would look like if we had a public class: [otel_sdk_example.pdf](https://github.com/googleapis/google-cloud-cpp/files/11180893/otel_sdk_example.pdf). (Note that the interface will change).

Abstract Doxygen's CLANG_OPTIONS ever so slightly.

Without the `trace_exporter.h` file, Doxygen would not have any input. (It skips `internal/` directories). And if Doxygen does not have input, it throws an error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11207)

### DIFF
--- a/google/cloud/opentelemetry/trace_exporter.h
+++ b/google/cloud/opentelemetry/trace_exporter.h
@@ -1,0 +1,32 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


google/cloud/opentelemetry/trace_exporter.h
```diff
@@ -0,0 +1,32 @@
+// Copyright 2023 Google LLC
```
coryan, line 1: This file seems unrelated to the docs, but meh.

coryan (approved):

google/cloud/opentelemetry/trace_exporter.h
dbolduc: In case you missed the PR description:

> Without the trace_exporter.h file, Doxygen would not have any input. (It skips internal/ directories). And if Doxygen does not have input, it throws an error.

I'd rather set up the doxygen up front and add a temporarily dumb file than have to remember to add doxygen later. :shrug:

PR #11219, opened 2023-04-09
impl(docfx): generate files for every compounddef

coryan:
My plan to use the ToC entries in driving the file generation does not work. We need all compounddef generated so the cross-references work.

Part of the work for #10895

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11219)


dbolduc (approved):
:shrug:

PR #11221, opened 2023-04-10
impl(otel): add TruncatableString helper

dbolduc:
Part of the work for #11156 

The `google::devtools::v2::Span` proto has several such `TruncatableString` fields. Add a helper for setting them.

I will remove the `TEST(Recordable, Compiles)` later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11221)

### DIFF
--- a/google/cloud/opentelemetry/internal/recordable.cc
+++ b/google/cloud/opentelemetry/internal/recordable.cc
@@ -27,7 +27,7 @@ bool IsTrailByte(char x) { return static_cast<signed char>(x) < -0x40; }
 void SetTruncatableString(
     google::devtools::cloudtrace::v2::TruncatableString& proto,
     opentelemetry::nostd::string_view value, std::size_t limit) {
-  if (value.size() < limit) {
+  if (value.size() <= limit) {
     proto.set_value(value.data(), value.size());
     proto.set_truncated_byte_count(0);
     return;
--- a/google/cloud/opentelemetry/internal/recordable.cc
+++ b/google/cloud/opentelemetry/internal/recordable.cc
@@ -18,6 +18,32 @@ namespace google {
 namespace cloud {
 namespace otel_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace {
+// Taken from:
+// http://35.193.25.4/TensorFlow/models/research/syntaxnet/util/utf8/unilib_utf8_utils.h
+bool IsTrailByte(char x) { return static_cast<signed char>(x) < -0x40; }
+}  // namespace
+
+void SetTruncatableString(
+    google::devtools::cloudtrace::v2::TruncatableString& proto,
+    opentelemetry::nostd::string_view value, std::size_t limit) {
+  if (value.size() < limit) {
+    proto.set_value(value.data(), value.size());
+    proto.set_truncated_byte_count(0);
+    return;
+  }
+
+  // If limit points to the beginning of a utf8 character, truncate at the
+  // limit. Otherwise, backtrack to the beginning of utf8 character.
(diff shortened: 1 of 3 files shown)


google/cloud/opentelemetry/internal/recordable_test.cc
```diff
@@ -24,6 +24,35 @@ namespace {
...
+  EXPECT_EQ(proto.value(), "abc");
+  EXPECT_EQ(proto.truncated_byte_count(), 2);
+}
+
+TEST(SetTruncatableString, RespectsUnicodeSymbolBoundaries) {
+  google::devtools::cloudtrace::v2::TruncatableString proto;
+  // The unicode symbol "д" is 2 bytes wide. So the string "дд" is 4 bytes long.
+  // Truncation should respect the symbol boundaries. i.e. We should not cut the
+  // symbol in half.
+  SetTruncatableString(proto, "дд", 3);
```
coryan, line 46: I think this is assuming something about the encoding of the file...  For all we know it might be `UTF-32` encoded.  You may want to use the hex representation, e.g.: "\xb4\xd0\xb4\xd0" (or maybe I have the bytes backwards).

google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -18,6 +18,32 @@ namespace google {
...
+  if (value.size() < limit) {
+    proto.set_value(value.data(), value.size());
+    proto.set_truncated_byte_count(0);
+    return;
+  }
+
+  // If limit points to the beginning of a utf8 character, truncate at the
+  // limit. Otherwise, backtrack to the beginning of utf8 character.
+  auto truncation_pos = limit;
+  while (truncation_pos > 0 && IsTrailByte(value[truncation_pos])) {
```
coryan, line 39: `truncation_pos > 0` suggests it may be useful to have a test that hits the case where all the bytes are "trailing bytes".

coryan (changes requested):

google/cloud/opentelemetry/internal/recordable.cc
devbww, line 39: We need `truncation_pos < value.size()` here, so best all around to make the short-circuit-return condition `if (value.size() <= limit)`.

google/cloud/opentelemetry/internal/recordable_test.cc
```diff
@@ -24,6 +24,35 @@ namespace {
...
+  EXPECT_EQ(proto.value(), "д");
+  EXPECT_EQ(proto.truncated_byte_count(), 2);
+
+  // "断" is 3 bytes wide.
+  SetTruncatableString(proto, "断断", 5);
```
devbww, line 51: Consider adding a test case for `SetTruncatableString(proto, "断断", 2)`.

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: Good catch, fixed.

dbolduc: Added.

coryan (approved):
I suspect "a unicode" is technically incorrect, and certainly irrelevant.  Most likely they are is a "UTF-8 encoded character" which happen to be encoded in 2-bytes or 3-bytes.  What they represent in Unicode is not of any consequence for your code.

But then again, I know almost nothing about Unicode, UTF-8 or character sets in general.

dbolduc:
> "a unicode"

Yeah, that was a gaffe. I thought I had written "a unicode symbol".

> "UTF-8 encoded character"

Thanks, let's go with that.

devbww (approved):

PR #11223, opened 2023-04-10
impl(bigquery): Logging Decorator Cleanup: Added request and response logging to logging decorator

### LINKED ISSUE #11122: Cleanup bigquery REST logging decorator for the V2 job api
Based on code review feedback from https://reviewable.io/reviews/googleapis/google-cloud-cpp/11081

Modify logging decorator as follows

a) Add response logging 
b) Use tracing options for request and response logging 
c) Add unit tests

jsrinnn:
This is the final PR for the logging decorator cleanup and fixes #11122

This PR integrates request and response logging with tracing options in the logging decorator and adds unit tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.h"
 #include "google/cloud/internal/debug_string.h"
 #include "google/cloud/internal/make_status.h"
+#include "google/cloud/internal/rest_response.h"
 
 namespace google {
 namespace cloud {
@@ -22,6 +23,9 @@ namespace bigquery_v2_minimal_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 namespace rest = ::google::cloud::rest_internal;
+
+BigQueryHttpResponse::BigQueryHttpResponse()
+    : http_status_code(static_cast<rest_internal::HttpStatusCode>(0)) {}
 
 StatusOr<BigQueryHttpResponse> BigQueryHttpResponse::BuildFromRestResponse(
     std::unique_ptr<rest::RestResponse> rest_response) {
@@ -48,6 +52,7 @@ std::string BigQueryHttpResponse::DebugString(absl::string_view name,
                                               int indent) const {
   // Payload is not logged as it might contain user sensitive data like
   // ldap/emails.
+
   return internal::DebugFormatter(name, options, indent)
       .Field("status_code", http_status_code)
       .Field("http_headers", http_headers)
--- a/google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc
@@ -24,9 +24,6 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 namespace rest = ::google::cloud::rest_internal;
 
-BigQueryHttpResponse::BigQueryHttpResponse()
-    : http_status_code(static_cast<rest_internal::HttpStatusCode>(0)) {}
-
 StatusOr<BigQueryHttpResponse> BigQueryHttpResponse::BuildFromRestResponse(
     std::unique_ptr<rest::RestResponse> rest_response) {
   BigQueryHttpResponse response;
@@ -52,7 +49,6 @@ std::string BigQueryHttpResponse::DebugString(absl::string_view name,
                                               int indent) const {
   // Payload is not logged as it might contain user sensitive data like
   // ldap/emails.
-
   return internal::DebugFormatter(name, options, indent)
       .Field("status_code", http_status_code)
       .Field("http_headers", http_headers)
--- a/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
@@ -21,6 +21,7 @@
 #include "google/cloud/internal/rest_context.h"
 #include "google/cloud/log.h"
 #include "google/cloud/status_or.h"
+#include "absl/strings/string_view.h"
 
 namespace google {
 namespace cloud {
@@ -32,13 +33,15 @@ template <typename Functor, typename Request,
               Functor, rest_internal::RestContext&, Request const&>>
 Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
                      Request const& request, char const* where,
+                     absl::string_view request_name,
+                     absl::string_view response_name,
                      TracingOptions const& options) {
   auto formatter = [](std::string* out, auto const& header) {
     absl::StrAppend(out, "name=", header.first, ", value={",
                     absl::StrJoin(header.second, "&"), "}");
   };
-  GCP_LOG(DEBUG) << where << "() << " << request.DebugString(options)
-                 << ", Context={"
+  GCP_LOG(DEBUG) << where << "() << "
+                 << request.DebugString(request_name, options) << ", Context={"
                  << internal::DebugString(
                         absl::StrJoin(context.headers(), ", ", formatter),
                         options)
@@ -46,6 +49,10 @@ Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
 
   auto response = functor(context, request);
   GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
+
+  if (response.ok()) {
+    GCP_LOG(DEBUG) << ", " << response->DebugString(response_name, options);
+  }
 
   return response;
 }
@@ -57,7 +64,7 @@ BigQueryJobLogging::BigQueryJobLogging(
       tracing_options_(std::move(tracing_options)),
       components_(std::move(components)) {}
 
-// google::cloud::internal::LogWrapper cannot be used here since the request is
+// Customized LogWrapper is used here since GetJobRequest is
 // not a protobuf message.
 StatusOr<GetJobResponse> BigQueryJobLogging::GetJob(
     rest_internal::RestContext& rest_context, GetJobRequest const& request) {
@@ -66,9 +73,12 @@ StatusOr<GetJobResponse> BigQueryJobLogging::GetJob(
              GetJobRequest const& request) {
         return child_->GetJob(rest_context, request);
       },
-      rest_context, request, __func__, tracing_options_);
+      rest_context, request, __func__, "GetJobRequest", "GetJobResponse",
+      tracing_options_);
 }
 
+// Customized LogWrapper is used here since ListJobsRequest is
+// not a protobuf message.
 StatusOr<ListJobsResponse> BigQueryJobLogging::ListJobs(
     rest_internal::RestContext& rest_context, ListJobsRequest const& request) {
   return JobLogWrapper(
@@ -76,7 +86,8 @@ StatusOr<ListJobsResponse> BigQueryJobLogging::ListJobs(
              ListJobsRequest const& request) {
         return child_->ListJobs(rest_context, request);
       },
-      rest_context, request, __func__, tracing_options_);
+      rest_context, request, __func__, "ListJobsRequest", "ListJobsResponse",
+      tracing_options_);
 }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
--- a/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
@@ -37,7 +37,8 @@ Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
     absl::StrAppend(out, "name=", header.first, ", value={",
                     absl::StrJoin(header.second, "&"), "}");
   };
-  GCP_LOG(DEBUG) << where << "() " << request << ", Context={"
+  GCP_LOG(DEBUG) << where << "() << " << request.DebugString(options)
+                 << ", Context={"
                  << internal::DebugString(
                         absl::StrJoin(context.headers(), ", ", formatter),
                         options)
--- a/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
@@ -38,14 +38,15 @@ Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
                      absl::string_view response_name,
                      TracingOptions const& options) {
   auto formatter = [options](std::string* out, auto const& header) {
+    const auto* delim = options.single_line_mode() ? "&" : "\n";
     absl::StrAppend(
         out, " { name: \"", header.first, "\" value: \"",
-        internal::DebugString(absl::StrJoin(header.second, "&"), options),
-        "\" } ");
+        internal::DebugString(absl::StrJoin(header.second, delim), options),
+        "\" }");
   };
   GCP_LOG(DEBUG) << where << "() << "
                  << request.DebugString(request_name, options) << ", Context {"
-                 << absl::StrJoin(context.headers(), ", ", formatter) << " }";
+                 << absl::StrJoin(context.headers(), "", formatter) << " }";
 
   auto response = functor(context, request);
   if (!response) {
--- a/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_logging.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/bigquery/v2/minimal/internal/job_logging.h"
 #include "google/cloud/internal/absl_str_join_quiet.h"
 #include "google/cloud/internal/debug_string.h"
+#include "google/cloud/internal/debug_string_status.h"
 #include "google/cloud/internal/invoke_result.h"
 #include "google/cloud/internal/rest_context.h"
 #include "google/cloud/log.h"
@@ -36,22 +37,23 @@ Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
                      absl::string_view request_name,
                      absl::string_view response_name,
                      TracingOptions const& options) {
-  auto formatter = [](std::string* out, auto const& header) {
-    absl::StrAppend(out, "name=", header.first, ", value={",
-                    absl::StrJoin(header.second, "&"), "}");
+  auto formatter = [options](std::string* out, auto const& header) {
+    absl::StrAppend(
+        out, " { name: \"", header.first, "\" value: \"",
+        internal::DebugString(absl::StrJoin(header.second, "&"), options),
+        "\" } ");
   };
   GCP_LOG(DEBUG) << where << "() << "
-                 << request.DebugString(request_name, options) << ", Context={"
-                 << internal::DebugString(
-                        absl::StrJoin(context.headers(), ", ", formatter),
-                        options)
-                 << "}";
+                 << request.DebugString(request_name, options) << ", Context {"
+                 << absl::StrJoin(context.headers(), ", ", formatter) << " }";
 
   auto response = functor(context, request);
-  GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
-
-  if (response.ok()) {
-    GCP_LOG(DEBUG) << ", " << response->DebugString(response_name, options);
+  if (!response) {
+    GCP_LOG(DEBUG) << where << "() >> status="
+                   << internal::DebugString(response.status(), options);
+  } else {
+    GCP_LOG(DEBUG) << where << "() >> response="
+                   << response->DebugString(response_name, options);
   }
 
   return response;
@@ -73,7 +75,9 @@ StatusOr<GetJobResponse> BigQueryJobLogging::GetJob(
              GetJobRequest const& request) {
         return child_->GetJob(rest_context, request);
       },
-      rest_context, request, __func__, "GetJobRequest", "GetJobResponse",
+      rest_context, request, __func__,
+      "google.cloud.bigquery.v2.minimal.internal.GetJobRequest",
+      "google.cloud.bigquery.v2.minimal.internal.GetJobResponse",
       tracing_options_);
 }
 
@@ -86,7 +90,9 @@ StatusOr<ListJobsResponse> BigQueryJobLogging::ListJobs(
              ListJobsRequest const& request) {
         return child_->ListJobs(rest_context, request);
       },
-      rest_context, request, __func__, "ListJobsRequest", "ListJobsResponse",
+      rest_context, request, __func__,
+      "google.cloud.bigquery.v2.minimal.internal.ListJobsRequest",
+      "google.cloud.bigquery.v2.minimal.internal.ListJobsResponse",
       tracing_options_);
 }
 
--- a/google/cloud/bigquery/v2/minimal/internal/job_logging_test.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_logging_test.cc
@@ -80,19 +80,23 @@ TEST_F(JobLoggingClientTest, GetJob) {
         return GetJobResponse::BuildFromHttpResponse(std::move(http_response));
       });
 
+  // Not checking all fields exhaustively. Just that request and response are
+  // being logged.
   EXPECT_CALL(*log_backend_, ProcessWithOwnership)
       .WillOnce([](LogRecord const& lr) {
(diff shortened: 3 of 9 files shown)


jsrinnn:
Latest push fixes the msan-pr test failure in bigquery_http_response object.

Thanks!

google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc
```diff
@@ -24,6 +25,9 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 namespace rest = ::google::cloud::rest_internal;
 
+BigQueryHttpResponse::BigQueryHttpResponse()
+    : http_status_code(static_cast<rest_internal::HttpStatusCode>(0)) {}
```
devbww, line 29: Why not use an actual `HttpStatusCode` value, like `HttpStatusCode::kOk`, instead of casting arbitrary integers?

google/cloud/bigquery/v2/minimal/internal/job_logging.cc
```diff
@@ -32,12 +33,15 @@ template <typename Functor, typename Request,
...
+                     absl::string_view response_name,
                      TracingOptions const& options) {
   auto formatter = [](std::string* out, auto const& header) {
     absl::StrAppend(out, "name=", header.first, ", value={",
                     absl::StrJoin(header.second, "&"), "}");
   };
-  GCP_LOG(DEBUG) << where << "() " << request << ", Context={"
+  GCP_LOG(DEBUG) << where << "() << "
+                 << request.DebugString(request_name, options) << ", Context={"
                  << internal::DebugString(
```
devbww, line 45: Typically we want to apply the truncation to string values rather than formatted elements, so you'd get output like ...
```
  Context={name="name", value={"foo&ba...<truncated>..."}}
```
instead of ...
```
  Context={name="...<truncated>...
```

devbww, line 45: And to the end, you may like to think about how things are bracketed/quoted.

I would vote for something like ...
```
  Context { { name: "name1" value: "value1" } { name: "name2" value: "value2" } }
```
which would more closely align with the request/response logging.

google/cloud/bigquery/v2/minimal/internal/job_logging.cc
```diff
@@ -46,6 +50,10 @@ Result JobLogWrapper(Functor&& functor, rest_internal::RestContext& context,
   auto response = functor(context, request);
   GCP_LOG(DEBUG) << where << "() >> status=" << response.status();
 
+  if (response.ok()) {
+    GCP_LOG(DEBUG) << ", " << response->DebugString(response_name, options);
+  }
```
devbww, line 55: `LogWrapper` renders such cases as ...
```
  if (!response) {
    GCP_LOG(DEBUG) << where << "() >> status="
                   << DebugString(response.status(), options);
  } else {
    GCP_LOG(DEBUG) << where
                   << "() >> response=" << DebugString(*response, options);
  }
```
`JobLogWrapper` should follow the same pattern.

google/cloud/bigquery/v2/minimal/internal/job_logging.cc
```diff
@@ -65,17 +73,21 @@ StatusOr<GetJobResponse> BigQueryJobLogging::GetJob(
              GetJobRequest const& request) {
         return child_->GetJob(rest_context, request);
       },
-      rest_context, request, __func__, tracing_options_);
+      rest_context, request, __func__, "GetJobRequest", "GetJobResponse",
```
devbww, line 76: The protobuf formatter uses `m.GetTypeName()`, which renders the fully-qualified type name.  This should follow the same pattern.

google/cloud/bigquery/v2/minimal/internal/job_logging_test.cc
```diff
@@ -0,0 +1,170 @@
...
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+using ::google::cloud::bigquery_v2_minimal_testing::MockBigQueryJobRestStub;
+using ::testing::HasSubstr;
+using ::testing::IsEmpty;
+
+std::shared_ptr<BigQueryJobLogging> CreateMockJobLogging(
+    std::shared_ptr<BigQueryJobRestStub> mock) {
+  return std::make_shared<BigQueryJobLogging>(std::move(mock), TracingOptions{},
+                                              std::set<std::string>{});
```
devbww, line 36: These `components` appear to be unused.  Should we get rid of them until they're needed?

google/cloud/bigquery/v2/minimal/internal/job_logging_test.cc
```diff
@@ -0,0 +1,170 @@
...
+        EXPECT_THAT(lr.message, HasSubstr("job_id: \"j123\""));
+        EXPECT_THAT(lr.message, HasSubstr("project_id: \"p123\""));
+      })
+      .WillOnce([](LogRecord const& lr) {
+        EXPECT_THAT(lr.message, HasSubstr(" >> "));
+        EXPECT_THAT(lr.message, HasSubstr("status=OK"));
+      })
+      .WillOnce([](LogRecord const& lr) {
+        EXPECT_THAT(lr.message, HasSubstr("GetJobResponse"));
+        EXPECT_THAT(lr.message, HasSubstr("status: \"DONE\""));
```
devbww, line 98: Opt: Use raw string literals to avoid all the escaping.

jsrinnn (commented):
Addressed all feedback. Thank you, Bradley!

I would really appreciate if I can get an approval for this PR as the logging formatting changes have been going on for over a week now, and I would like to make progress on the core work for other apis. We should be able to address any remaining formatting changes as cleanup later on. 

Thanks.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#-:-NSw71sL58u6Qah2CqfU:brapxuz)* status: 0 of 9 files reviewed, 8 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc` line 29 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163045269:-NSvzR8H01gHMtBEPj95:b-fo37b4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc#L29)):*

Why not use an actual `HttpStatusCode` value, like `HttpStatusCode::kOk`, instead of casting arbitrary integers?
</blockquote></details>

Done.

I have modified this to use `HttpStatusCode::kOk` now. Please note the change is in the header file in accordance with clang-tidy recommendation.

Few reasons why I chose 0 initially. 

1) Since HttpStatusCode is an int32_t, I thought 0 was a good initializer for that type.

2) I didn't want to assume a specific success or failure `HttpStatusCode` as initial value, when the http operation has not been performed yet. I thought something like kUnknown or kUnspecified would be more appropriate as initial value but these were not defined.

3) The documentation in the HttpStatusCode header file mentioned any code < 100 would be treated as kUnknown so I thought 0 would be okay based on that.
___
*[`google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc` line 56 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163046152:-NSw0TKVDx3ChYAB7EgY:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc#L56)):*

Remove added blank line.
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 45 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163059280:-NSw1IRr9XjAeaMwvODj:b1hlgod) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L45)):*

Typically we want to apply the truncation to string values rather than formatted elements, so you'd get output like ...
```
  Context={name="name", value={"foo&ba...<truncated>..."}}
```
instead of ...
```
  Context={name="...<truncated>...
```
</blockquote></details>

Done.

Got it. Thanks.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 45 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163062929:-NSw2GOW0BAuC6dNIU5-:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L45)):*

And to the end, you may like to think about how things are bracketed/quoted.

I would vote for something like ...
```
  Context { { name: "name1" value: "value1" } { name: "name2" value: "value2" } }
```
which would more closely align with the request/response logging.
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 55 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163067051:-NSw2h_3-Z5UX5YUwD1g:b-2k96av) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L55)):*

`LogWrapper` renders such cases as ...
```
  if (!response) {
    GCP_LOG(DEBUG) << where << "() >> status="
                   << DebugString(response.status(), options);
  } else {
    GCP_LOG(DEBUG) << where
                   << "() >> response=" << DebugString(*response, options);
  }
```
`JobLogWrapper` should follow the same pattern.
</blockquote></details>

Done.

I have applied the same `if..else` format for status and response logging as above, with the following caveat

1) Our signatures for request and response logging are different from the protobuf versions, and are based on `DebugString()` member functions defined here: 

https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/bigquery/v2/minimal/internal/job_response.h#L38-#L40

2) DebugString() from `internal/debug_string_status.h` requires an include of `google_cloud_cpp_grpc_utils` in `bigquery_rest.cmake` file which didn't seem correct to me. But I have made the changes to be compliant with the feedback above. I'll let the cloud-cpp team decide how they want to refactor that header file.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 76 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163070163:-NSw3MWq5aitbyzI00Dq:b-3rrc16) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L76)):*

The protobuf formatter uses `m.GetTypeName()`, which renders the fully-qualified type name.  This should follow the same pattern.
</blockquote></details>

Done.

Made the change here and elsewhere in the file.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging_test.cc` line 36 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163075980:-NSw4Ci0EVXUAnnrtPdi:bkgb274) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/job_logging_test.cc#L36)):*

These `components` appear to be unused.  Should we get rid of them until they're needed?
</blockquote></details>

I am assuming you mean the `components` parameter? I followed the same pattern as other logging decorators for  bigquery and other projects 

https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/bigquery/storage/v1/internal/bigquery_read_logging_decorator.h#L37-#L39

so I'll keep the same api here for consistency. 
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging_test.cc` line 98 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163076745:-NSw70FA-G4sZECPJ2aa:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/job_logging_test.cc#L98)):*

Opt: Use raw string literals to avoid all the escaping.
</blockquote></details>

Done.

devbww (approved):
> Reviewed 3 of 6 files at r1, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223)* status: all files reviewed, 4 unresolved discussions (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc` line 29 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163045269:-NSy-d2n7XWsDsMQmFtV:b-7hzcs7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc#L29)):*

Done.

I have modified this to use `HttpStatusCode::kOk` now. Please note the change is in the header file in accordance with clang-tidy recommendation.

Few reasons why I chose 0 initially. 

1) Since HttpStatusCode is an int32_t, I thought 0 was a good initializer for that type.

2) I didn't want to assume a specific success or failure `HttpStatusCode` as initial value, when the http operation has not been performed yet. I thought something like kUnknown or kUnspecified would be more appropriate as initial value but these were not defined.

3) The documentation in the HttpStatusCode header file mentioned any code < 100 would be treated as kUnknown so I thought 0 would be okay based on that.

</blockquote></details>

It seems to me, then, that you should not be able to create a `BigQueryHttpResponse` until you have an actual `HttpStatusCode`.  That is, the single `BigQueryHttpResponse` constructor should take parameters that initialize all its fields.  And then a similar convention applies to `GetJobResponse` and `ListJobsResponse`.

Objects like you have, with public data members and no invariants, generally lead to fragile systems where you need to invent things like HTTP code 0.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 55 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163067051:-NSyAASm3ovkTzblh2Fh:btmgjtk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L55)):*
> DebugString() from `internal/debug_string_status.h` requires an include of `google_cloud_cpp_grpc_utils`

Right, because rendering a `google::cloud::Status` uses RPC-defined payloads.  In this case it seems those payloads will never exist though, so we could have another "debug_string_status" that doesn't use them and so can live in a friendlier library.

I'll raise the issue with the team.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#-NSyJ52K7sXPIvnRCCv5:-NSyJ52L6qOSCcQjEbAx:bpbfzrn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/731200c2d06a6f1cb5450d4457da8d9680f26d48/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L42)):*
> ```cpp
>   auto formatter = [options](std::string* out, auto const& header) {
>     absl::StrAppend(
>         out, " { name: \"", header.first, "\" value: \"",
> ```

If you're keeping a TODO list, note that this does not heed `single_line_mode=off`.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 44 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#-NSy94U7-s32AyF-L4LJ:-NSy94U7-s32AyF-L4LK:bdk3qy7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/731200c2d06a6f1cb5450d4457da8d9680f26d48/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L44)):*
> ```cpp
>         out, " { name: \"", header.first, "\" value: \"",
>         internal::DebugString(absl::StrJoin(header.second, "&"), options),
>         "\" } ");
> ```

It looks like we probably don't want the space after the `}`.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 48 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#-NSy9UEo6GKWEy_56Win:-NSy9UEo6GKWEy_56Wio:b-egojc3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/731200c2d06a6f1cb5450d4457da8d9680f26d48/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L48)):*
> ```cpp
>   GCP_LOG(DEBUG) << where << "() << "
>                  << request.DebugString(request_name, options) << ", Context {"
>                  << absl::StrJoin(context.headers(), ", ", formatter) << " }";
> ```

It looks like we don't want the `", "` separator (i.e., it should be `""`).

jsrinnn (commented):
Addressed all feedback. Thank you, Bradley!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#-:-NT-QD7QDLJwX8UTlCeG:b-7jlxxz)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc` line 29 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163045269:-NT-NTB9DPOyHBFkz9Tx:b-hpzbrw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc#L29)):*

It seems to me, then, that you should not be able to create a `BigQueryHttpResponse` until you have an actual `HttpStatusCode`.  That is, the single `BigQueryHttpResponse` constructor should take parameters that initialize all its fields.  And then a similar convention applies to `GetJobResponse` and `ListJobsResponse`.

Objects like you have, with public data members and no invariants, generally lead to fragile systems where you need to invent things like HTTP code 0.
</blockquote></details>

Ack. I agree regarding not using http code 0 . It wasn't very obvious what other code to use as I explained before. For the case in question only the httpstatuscode was problematic as it was being used in DebugString uninitialized. For the rest of the members default initializer were applied.

Resolving this since I have already addressed the prior feedback.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 55 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163067051:-NT-Q9tDAIrNHVhuQ2KL:bu7as39) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L55)):*

> DebugString() from `internal/debug_string_status.h` requires an include of `google_cloud_cpp_grpc_utils`

Right, because rendering a `google::cloud::Status` uses RPC-defined payloads.  In this case it seems those payloads will never exist though, so we could have another "debug_string_status" that doesn't use them and so can live in a friendlier library.

I'll raise the issue with the team.
</blockquote></details>

Ack. Thanks Bradley!
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#-NSyJ52K7sXPIvnRCCv5:-NT-OMAxEbw9Web4wWDp:b-o1gq54) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/731200c2d06a6f1cb5450d4457da8d9680f26d48/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L42)):*

If you're keeping a TODO list, note that this does not heed `single_line_mode=off`.
</blockquote></details>

Done

I have been keeping track of pending items via github issues. So far I have addressed all the requested feedback including the one above. 
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 44 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#-NSy94U7-s32AyF-L4LJ:-NT-PPgI6gHSSIvPWhRH:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/731200c2d06a6f1cb5450d4457da8d9680f26d48/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L44)):*

It looks like we probably don't want the space after the `}`.
</blockquote></details>

Done
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 48 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#-NSy9UEo6GKWEy_56Win:-NT-PvXI-pEd971eOXGD:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/731200c2d06a6f1cb5450d4457da8d9680f26d48/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L48)):*

It looks like we don't want the `", "` separator (i.e., it should be `""`).
</blockquote></details>

Done

devbww (commented):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc` line 29 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163045269:-NT-lqqkAs95gYg2rYYm:b-4mjyhg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/bigquery_http_response.cc#L29)):*

Ack. I agree regarding not using http code 0 . It wasn't very obvious what other code to use as I explained before. For the case in question only the httpstatuscode was problematic as it was being used in DebugString uninitialized. For the rest of the members default initializer were applied.

Resolving this since I have already addressed the prior feedback.
</blockquote></details>

Sorry, I think that is missing the point that you should not need to create a `BigQueryHttpResponse` until you have valid data for it.  Just because the header and payload types have default constructors does not mean they are not similarly problematic.  Either that, or the class needs some first-class "uninitialized" state.  But, so be it.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 42 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#-NSyJ52K7sXPIvnRCCv5:-NT-jear3qZT8VzCsg2C:bcs6kq5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/731200c2d06a6f1cb5450d4457da8d9680f26d48/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L42)):*

Done

I have been keeping track of pending items via github issues. So far I have addressed all the requested feedback including the one above. 
</blockquote></details>

That doesn't look right.  In `!single_line_mode` there should be newlines (with suitable indentation) between the header elements, not within the value string.  That is, ...
```
  Context {
    header {
      name: "name1"
      value: "value1A&value1B&val...<truncated>..."
    }
    header {
      name: "name2"
      value: "value2A&value2B"
    }
  }
```
But, so be it.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`google/cloud/bigquery/v2/minimal/internal/job_logging.cc` line 55 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11223#gh-1163067051:-NTF_ZO-5LLY1iRn-eVW:b1wvxae) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bda8fb861732a7e108198b89c708f8ef0a917507/google/cloud/bigquery/v2/minimal/internal/job_logging.cc#L55)):*

Ack. Thanks Bradley!
</blockquote></details>

So, the team consensus appears to be that they weren't concerned with the extra "grpc_utils" dependency.  But it seems to me that you could avoid it if you wanted (with no actual output change) by just streaming the `Status` value directly.  That is, ...
```
  GCP_LOG(DEBUG) << .... << response.status();
```
`internal::DebugString()` isn't buying you anything in these cases.

jsrinnn:
Ack. I agree regarding the redundancy in this case

I can remove this call, when I am addressing the logging decorator for the dataset resource.

Thanks for following up on this!

PR #11224, opened 2023-04-10
impl(bigquery): Addresses remaining cleanup items for job apis

jsrinnn:
Addressing some of the remaining cleanup items for the Job apis

This PR fixes Github issue#11123 and addresses the following changes with regards to Job Requests and stub

1) Removed passing of options to BuildRestRequest and Default job stub, to be consistent with other apis.

2) Made min_creation_time and max_creation_time optional so they can be checked and set easiy in request parameters.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -88,12 +88,19 @@ std::string GetJobRequest::DebugString(absl::string_view name,
 std::string ListJobsRequest::DebugString(absl::string_view name,
                                          TracingOptions const& options,
                                          int indent) const {
+  auto min_creation_time = min_creation_time_.has_value()
+                               ? min_creation_time_.value()
+                               : kDefaultTimepoint;
+  auto max_creation_time = max_creation_time_.has_value()
+                               ? max_creation_time_.value()
+                               : kDefaultTimepoint;
+
   return internal::DebugFormatter(name, options, indent)
       .StringField("project_id", project_id_)
       .Field("all_users", all_users_)
       .Field("max_results", max_results_)
-      .Field("min_creation_time", min_creation_time_.value())
-      .Field("max_creation_time", max_creation_time_.value())
+      .Field("min_creation_time", min_creation_time)
+      .Field("max_creation_time", max_creation_time)
       .StringField("page_token", page_token_)
       .SubMessage("projection", projection_)
       .SubMessage("state_filter", state_filter_)
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -92,8 +92,8 @@ std::string ListJobsRequest::DebugString(absl::string_view name,
       .StringField("project_id", project_id_)
       .Field("all_users", all_users_)
       .Field("max_results", max_results_)
-      .Field("min_creation_time", min_creation_time_)
-      .Field("max_creation_time", max_creation_time_)
+      .Field("min_creation_time", min_creation_time_.value())
+      .Field("max_creation_time", max_creation_time_.value())
       .StringField("page_token", page_token_)
       .SubMessage("projection", projection_)
       .SubMessage("state_filter", state_filter_)
@@ -172,13 +172,15 @@ StatusOr<rest_internal::RestRequest> BuildRestRequest(
   if (r.max_results() > 0) {
     request.AddQueryParameter("maxResults", std::to_string(r.max_results()));
   }
-  if (r.min_creation_time() != kDefaultTimepoint) {
-    request.AddQueryParameter("minCreationTime",
-                              internal::FormatRfc3339(r.min_creation_time()));
+  if (r.min_creation_time().has_value()) {
+    request.AddQueryParameter(
+        "minCreationTime",
+        internal::FormatRfc3339(r.min_creation_time().value()));
   }
-  if (r.max_creation_time() != kDefaultTimepoint) {
-    request.AddQueryParameter("maxCreationTime",
-                              internal::FormatRfc3339(r.max_creation_time()));
+  if (r.max_creation_time().has_value()) {
+    request.AddQueryParameter(
+        "maxCreationTime",
+        internal::FormatRfc3339(r.max_creation_time().value()));
   }
 
   auto if_not_empty_add = [&](char const* key, auto const& v) {
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -88,19 +88,12 @@ std::string GetJobRequest::DebugString(absl::string_view name,
 std::string ListJobsRequest::DebugString(absl::string_view name,
                                          TracingOptions const& options,
                                          int indent) const {
-  auto min_creation_time = min_creation_time_.has_value()
-                               ? min_creation_time_.value()
-                               : kDefaultTimepoint;
-  auto max_creation_time = max_creation_time_.has_value()
-                               ? max_creation_time_.value()
-                               : kDefaultTimepoint;
-
   return internal::DebugFormatter(name, options, indent)
       .StringField("project_id", project_id_)
       .Field("all_users", all_users_)
       .Field("max_results", max_results_)
-      .Field("min_creation_time", min_creation_time)
-      .Field("max_creation_time", max_creation_time)
+      .Field("min_creation_time", *min_creation_time_)
+      .Field("max_creation_time", *max_creation_time_)
       .StringField("page_token", page_token_)
       .SubMessage("projection", projection_)
       .SubMessage("state_filter", state_filter_)
@@ -179,15 +172,13 @@ StatusOr<rest_internal::RestRequest> BuildRestRequest(
   if (r.max_results() > 0) {
     request.AddQueryParameter("maxResults", std::to_string(r.max_results()));
   }
-  if (r.min_creation_time().has_value()) {
-    request.AddQueryParameter(
-        "minCreationTime",
-        internal::FormatRfc3339(r.min_creation_time().value()));
+  if (r.min_creation_time()) {
+    request.AddQueryParameter("minCreationTime",
+                              internal::FormatRfc3339(*r.min_creation_time()));
   }
-  if (r.max_creation_time().has_value()) {
-    request.AddQueryParameter(
-        "maxCreationTime",
-        internal::FormatRfc3339(r.max_creation_time().value()));
+  if (r.max_creation_time()) {
+    request.AddQueryParameter("maxCreationTime",
+                              internal::FormatRfc3339(*r.max_creation_time()));
   }
 
   auto if_not_empty_add = [&](char const* key, auto const& v) {
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -124,8 +124,9 @@ ListJobsRequest::ListJobsRequest(std::string project_id)
       min_creation_time_(kDefaultTimepoint),
       max_creation_time_(kDefaultTimepoint) {}
 
-StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest const& r,
-                                                      Options const& opts) {
+StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest const& r) {
+  auto const& opts = internal::CurrentOptions();
+
   rest_internal::RestRequest request;
   if (r.project_id().empty()) {
     return internal::InvalidArgumentError(
@@ -149,9 +150,11 @@ StatusOr<rest_internal::RestRequest> BuildRestRequest(GetJobRequest const& r,
   return request;
 }
 
-StatusOr<rest_internal::RestRequest> BuildRestRequest(ListJobsRequest const& r,
-                                                      Options const& opts) {
+StatusOr<rest_internal::RestRequest> BuildRestRequest(
+    ListJobsRequest const& r) {
   rest_internal::RestRequest request;
+  auto const& opts = internal::CurrentOptions();
+
   if (r.project_id().empty()) {
     return internal::InvalidArgumentError(
         "Invalid ListJobsRequest: Project Id is empty", GCP_ERROR_INFO());
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -28,8 +28,6 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 namespace {
 
-auto const kDefaultTimepoint = std::chrono::system_clock::time_point{};
-
 std::string GetJobsEndpoint(Options const& opts) {
   std::string endpoint = opts.get<EndpointOption>();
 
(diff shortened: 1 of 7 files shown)


google/cloud/bigquery/v2/minimal/internal/job_request.h
```diff
@@ -106,10 +107,12 @@ class ListJobsRequest {
   std::string const& project_id() const { return project_id_; }
   bool const& all_users() const { return all_users_; }
   std::int32_t const& max_results() const { return max_results_; }
-  std::chrono::system_clock::time_point const& min_creation_time() const {
+  optional<std::chrono::system_clock::time_point> const& min_creation_time()
```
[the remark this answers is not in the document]
coryan: The only reason we still have `google::cloud::optional<>` is for backwards compatibility.

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -92,8 +92,8 @@ std::string ListJobsRequest::DebugString(absl::string_view name,
       .StringField("project_id", project_id_)
       .Field("all_users", all_users_)
       .Field("max_results", max_results_)
-      .Field("min_creation_time", min_creation_time_)
-      .Field("max_creation_time", max_creation_time_)
+      .Field("min_creation_time", min_creation_time_.value())
```
coryan, line 95: Calling `.value()` if the optional is not set will result in a crash.  @devbww is working on making this easier (see #11228).

coryan (commented):
I am not the right person to approve these, as I am not familiar with the original feedback.  But I can identify some nits that you may want to fix ahead of a "for realsies" review.

jsrinnn (commented):
Thanks Carlos! I have addressed the nits. For context, the specific PR addresses the following original feedback

1) Making min and max creation time as Optional so it can be checked easily and set as request parameters when the building the request

2) Removing passing of options in request building and using currentoptions() instead to retrieve the endpoint.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#-:-NSkbWOG-Ein68UnFwX3:b-h7twus)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @coryan and @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.h` line 110 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1162848308:-NSkbMJ065ixQI0bec_I:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3cd8e70fecab54c251b3c92c1a83788cbe822399/google/cloud/bigquery/v2/minimal/internal/job_request.h#L110)):*

The only reason we still have `google::cloud::optional<>` is for backwards compatibility.
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 95 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1162851008:-NSkcK9L6rX7HvVUCPI4:b-58z3t2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3cd8e70fecab54c251b3c92c1a83788cbe822399/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L95)):*

Calling `.value()` if the optional is not set will result in a crash.  @devbww is working on making this easier (see #11228).
</blockquote></details>

Done.

Added checking for value via has_value(). These members are also set with default values in the constructor.

google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc
```diff
@@ -32,12 +32,14 @@ std::shared_ptr<BigQueryJobRestStub> CreateDefaultBigQueryJobRestStub(
   if (!local_opts.has<UnifiedCredentialsOption>()) {
     local_opts.set<UnifiedCredentialsOption>(MakeGoogleDefaultCredentials());
   }
+
+  internal::OptionsSpan span(local_opts);
```
devbww, line 36: It doesn't look like we need to instantiate an `OptionsSpan` here.

devbww (commented):
> Reviewed 5 of 7 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224)* status: all files reviewed, 4 unresolved discussions (waiting on @coryan and @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 95 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1162851008:-NSwV-it30o6Dv-sL2tc:bkwk9jn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3cd8e70fecab54c251b3c92c1a83788cbe822399/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L95)):*
Pass the `optional<time_point>` itself, not the (optionally) contained `time_point`.  That is, remove the `*` here and the on the next line.  Then the `DebugFormatter` can decide whether the optional is engaged, and so know if it should include the field in the log.

> These members are also set with default values in the constructor.

We should avoid that ... their default values should be "disengaged".  See below.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 96 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1163083292:-NSwUHgyEp9uixLY_6oN:bba7z6s) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/51502c6e02779cfc12f61d7804e6b19ed0da4554/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L96)):*

Done.

Can you PTAL and let me know if the modified  usage is correct? i.e Do I need to check if the optional has a value, before I do a `*min_creation_time_`? 

I am assuming the current usage won't cause a crash if the optional doesn't have a value but let me know if it is otherwise.
</blockquote></details>

Yes, you need to check that an optional has a value before "dereferencing" (`operator*()`) it.  In this particular case that is moot, however, as we don't want to dereference.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 125 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#-NSwapb7AQVm5W9pgrPK:-NSwapb7AQVm5W9pgrPL:bc30yzz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0730711f57605103daa864f0fdd35a04c73c78e8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L125)):*
> ```cpp
>       max_results_(0),
>       min_creation_time_(kDefaultTimepoint),
>       max_creation_time_(kDefaultTimepoint) {}
> ```

We should leave these default constructed (i.e., disengaged).  And at that point, `kDefaultTimepoint` can go away, which was the goal.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 36 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1163103902:-NSwZrosEt7TWvW_B_4Z:b-ojx88m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/51502c6e02779cfc12f61d7804e6b19ed0da4554/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L36)):*

The reason I added it was we do modify the local_opts value above. I wanted to make sure the modified value is updated in the options in the thread local storage otherwise `CurrentOptions()` may not reflect the updated value.

Let me know if the above understanding is incorrect i.e if a subsequent call to `CurrentOptions()` will reflect the latest updated value then I can remove this. 
</blockquote></details>

The observation is that nothing here does (or should) call `CurrentOptions()`.  All the options are being passed explicitly, so the span is redundant.

Compare this with a generated `CreateDefault*Stub()` function, like, for example, google/cloud/workstations/v1/internal/workstations_stub_factory.cc.

jsrinnn (commented):
Addressed all feedback. Thank you Bradley!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#-:-NT-GDW56ooA99d4a6C7:b-frg0nh)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @coryan and @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 95 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1162851008:-NT-3ud990Ak64rVpjsq:btg1txj) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3cd8e70fecab54c251b3c92c1a83788cbe822399/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L95)):*

Pass the `optional<time_point>` itself, not the (optionally) contained `time_point`.  That is, remove the `*` here and the on the next line.  Then the `DebugFormatter` can decide whether the optional is engaged, and so know if it should include the field in the log.

> These members are also set with default values in the constructor.

We should avoid that ... their default values should be "disengaged".  See below.
</blockquote></details>

Please see my above comment. Passing the optional directly causes a compiler error as stated in my earlier reply. Below is the compile error for reference

https://screenshot.googleplex.com/76uRjUckubxNV3j
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 96 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1163083292:-NT-5cwf8npNVMgaybxc:b-qlrmkk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/51502c6e02779cfc12f61d7804e6b19ed0da4554/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L96)):*

Yes, you need to check that an optional has a value before "dereferencing" (`operator*()`) it.  In this particular case that is moot, however, as we don't want to dereference.
</blockquote></details>

Ah....that what I originally tried i.e passing it without the `*` dereference but that doesn't work in this case as it causes a compile error. I think this is because `std::chrono Timepoint` does not have a DebugString() member function.  

Below is the compile error I get:

https://screenshot.googleplex.com/76uRjUckubxNV3j

I have modified to add the sanity checks before derefencing based on your above feedback.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 125 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#-NSwapb7AQVm5W9pgrPK:-NT-EYWCEDhNsfWMJZe9:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0730711f57605103daa864f0fdd35a04c73c78e8/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L125)):*

We should leave these default constructed (i.e., disengaged).  And at that point, `kDefaultTimepoint` can go away, which was the goal.
</blockquote></details>

Done
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 36 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1163103902:-NT-EgIW2Ruu_HbSxl_g:baqhrvu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/51502c6e02779cfc12f61d7804e6b19ed0da4554/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L36)):*

The observation is that nothing here does (or should) call `CurrentOptions()`.  All the options are being passed explicitly, so the span is redundant.

Compare this with a generated `CreateDefault*Stub()` function, like, for example, google/cloud/workstations/v1/internal/workstations_stub_factory.cc.
</blockquote></details>

Done.

I have removed the OptionsSpan.

But my original question was if CurrentOptions() is called later downstream (like in BuildRestRequest()), will it have the modified localopts that is set here. From your feedback it looks like it will which is why I have removed OptionsSpan from here. Thanks

devbww (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 95 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1162851008:-NT-WxoC68_paZMi_Lxs:b-tp6vjy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3cd8e70fecab54c251b3c92c1a83788cbe822399/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L95)):*

Please see my above comment. Passing the optional directly causes a compiler error as stated in my earlier reply. Below is the compile error for reference

https://screenshot.googleplex.com/76uRjUckubxNV3j
</blockquote></details>

OK, thanks.  Leave things as you have them, and I'll fix `DebugFormatter` and circle back to tidy this up.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 96 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1163083292:-NT-WIY6Egt-07CbPl2-:bjfle7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/51502c6e02779cfc12f61d7804e6b19ed0da4554/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L96)):*

Ah....that what I originally tried i.e passing it without the `*` dereference but that doesn't work in this case as it causes a compile error. I think this is because `std::chrono Timepoint` does not have a DebugString() member function.  

Below is the compile error I get:

https://screenshot.googleplex.com/76uRjUckubxNV3j

I have modified to add the sanity checks before derefencing based on your above feedback.


</blockquote></details>

Sorry about that.  This is a flaw in `DebugFormatter`.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 36 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1163103902:-NT-TYMh8NNMyJQh8z8Y:bdh5aws) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/51502c6e02779cfc12f61d7804e6b19ed0da4554/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L36)):*

Done.

I have removed the OptionsSpan.

But my original question was if CurrentOptions() is called later downstream (like in BuildRestRequest()), will it have the modified localopts that is set here. From your feedback it looks like it will which is why I have removed OptionsSpan from here. Thanks
</blockquote></details>

No, calls outside of those made here (where we happen to pass the options explicitly) will not see `local_opts`.  But instantiating an `OptionsSpan` here won't change that either.

If other call stacks need to see the default `UnifiedCredentialsOption` set here, then that default would need to be set at a higher level.  But it doesn't look like other call stacks need to see that option ... it is only used by `MakePooledRestClient()`.

jsrinnn (commented):
Addressed remaining feedback. Thanks Bradley

I'll go ahead and merge these changes once the logging decorator changes are merged.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#-:-NT-_f-UC-gFjrzu6DYx:bqkcmzy)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 95 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1162851008:-NT-_6CuERhtZQA9AAUO:b-i7wwnl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/3cd8e70fecab54c251b3c92c1a83788cbe822399/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L95)):*

OK, thanks.  Leave things as you have them, and I'll fix `DebugFormatter` and circle back to tidy this up.
</blockquote></details>

Ack. Thanks.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 96 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1163083292:-NT-ZydK8aFyVcyM6upI:b-7yl6ap) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/51502c6e02779cfc12f61d7804e6b19ed0da4554/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L96)):*

Sorry about that.  This is a flaw in `DebugFormatter`.
</blockquote></details>

No worries. I agree with the feedback that ideal change would be to not have dereferencing.
___
*[`google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc` line 36 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11224#gh-1163103902:-NT-_HYT-s5W39KFuvAz:b-598nl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/51502c6e02779cfc12f61d7804e6b19ed0da4554/google/cloud/bigquery/v2/minimal/internal/job_rest_stub_factory.cc#L36)):*

No, calls outside of those made here (where we happen to pass the options explicitly) will not see `local_opts`.  But instantiating an `OptionsSpan` here won't change that either.

If other call stacks need to see the default `UnifiedCredentialsOption` set here, then that default would need to be set at a higher level.  But it doesn't look like other call stacks need to see that option ... it is only used by `MakePooledRestClient()`.
</blockquote></details>

Ah..got it. Yeah...its not needed in this case.

PR #11229, opened 2023-04-11
impl(otel): Recordable set display name

dbolduc:
Part of the work for #11156

Implement `Recordable::SetName()`. Also add a method to extract the underlying `Span` proto from the `Recordable`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11229)

### DIFF
--- a/google/cloud/opentelemetry/internal/recordable.h
+++ b/google/cloud/opentelemetry/internal/recordable.h
@@ -27,6 +27,9 @@ namespace cloud {
 namespace otel_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+// https://github.com/googleapis/googleapis/blob/a5e03049afc43ccb1217d5dd3be53006fc278643/google/devtools/cloudtrace/v2/trace.proto#L233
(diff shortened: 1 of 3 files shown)


google/cloud/opentelemetry/internal/recordable.h
```diff
@@ -27,6 +27,9 @@ namespace cloud {
 namespace otel_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
+// https://github.com/googleapis/googleapis/blob/a5e03049afc43ccb1217d5dd3be53006fc278643/google/devtools/cloudtrace/v2/trace.proto#L233
+std::size_t constexpr kDisplayNameStringLimit = 128;
```
coryan, line 31: It sure would be nice if this limit was in a constant (an enum?) in the proto file instead of just a comment.

coryan (approved):

google/cloud/opentelemetry/internal/recordable.h
dbolduc: No opinion.

PR #11240, opened 2023-04-11
impl(GCS+gRPC): reduce allocations in absl::Cord

coryan:
The `absl::Cord(absl::string_view)` constructor splits the input data into small chunks of 4KiB each. That puts pressure on the allocator, as we often use this with 32MiB buffers. Esun had two really good suggestions:

- Use `absl::MakeCordFromExternal()` to avoid the pressure on the allocator.
- Only create a Cord when needed in `Next()`, to reduce the overall memory usage.

That made `Next()` the only function that needed specialization, so do that.

Thanks to @veblush for the original patch and the effort to test this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11240)

### DIFF
--- a/google/cloud/storage/internal/grpc_split_write_object_data.cc
+++ b/google/cloud/storage/internal/grpc_split_write_object_data.cc
@@ -21,20 +21,7 @@ namespace cloud {
 namespace storage_internal {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
-SplitObjectWriteData<std::string>::SplitObjectWriteData(
-    absl::string_view buffer)
-    : buffers_({buffer}),
-      total_size_(storage::internal::TotalBytes(buffers_)) {}
-
-SplitObjectWriteData<std::string>::SplitObjectWriteData(
-    google::cloud::storage::internal::ConstBufferSequence buffers)
-    : buffers_(std::move(buffers)),
-      total_size_(storage::internal::TotalBytes(buffers_)) {}
-
-bool SplitObjectWriteData<std::string>::Done() const {
-  return offset_ >= total_size_;
-}
-
+template <>
 std::string SplitObjectWriteData<std::string>::Next() {
   auto constexpr kMax = static_cast<std::size_t>(
       google::storage::v2::ServiceConstants::MAX_WRITE_CHUNK_BYTES);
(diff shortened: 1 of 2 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11240)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

devbww (approved):

google/cloud/storage/internal/grpc_split_write_object_data.cc
```diff
@@ -54,26 +41,35 @@ std::string SplitObjectWriteData<std::string>::Next() {
...
+      auto buffer = std::unique_ptr<char[]>(new char[n]);
+      std::copy_n(b.begin() + start, n, buffer.get());
+      // Get `buffer.get()` before we std::move-from.
+      auto contents = absl::string_view{buffer.get(), n};
+      result.Append(absl::MakeCordFromExternal(
+          contents, [b = std::move(buffer)]() mutable {}));
+    }
+    offset_ += n;
+    start = 0;
+    if (result.size() >= kMax) return result;
```
devbww, line 71: Nit: I think I would make both occurrences of this say `break` instead of `return`.

devbww (approved):

google/cloud/storage/internal/grpc_split_write_object_data.cc
```diff
@@ -21,20 +21,7 @@ namespace cloud {
...
-      total_size_(storage::internal::TotalBytes(buffers_)) {}
-
-bool SplitObjectWriteData<std::string>::Done() const {
-  return offset_ >= total_size_;
-}
-
+template <>
 std::string SplitObjectWriteData<std::string>::Next() {
   auto constexpr kMax = static_cast<std::size_t>(
       google::storage::v2::ServiceConstants::MAX_WRITE_CHUNK_BYTES);
```
devbww, line 27: PS: In the `std::string` case, it _might_ behoove you to precompute the final size and reserve space up front (assuming that copying dominates).

devbww, line 27: Aside: It looks like we can return more that `MAX_WRITE_CHUNK_BYTES`, which sounds odd.

google/cloud/storage/internal/grpc_split_write_object_data.cc
coryan: Ack.  I can clean this up tomorrow.

google/cloud/storage/internal/grpc_split_write_object_data.cc
coryan: Good idea, will do.

coryan: How?  `result` is always grown by at most `kMax - result.size()` bytes:

```
auto n = (std::min)(kMax - result.size(), b.size() - start);
```

and as soon as it reaches `kMax` the function returns.  I guess I should write `if (result.size() == kMax)`?

devbww: My bad.  Sorry.  But, yes, `if (result.size() == kMax)` would make that clearer.

devbww: Given your clarification regarding `kMax`, perhaps a simple `result.reserve(kMax)` would be best.

Issue #11236, opened 2023-04-11
Create matchers for `otel::Span` links


coryan:
This suggests to me that we need a matcher for span links. I think we would just check the span context. I don't think there is a simple way to match on names. (We would have to modify/wrap the in-memory exporter to maintain a span map, keyed by name, or something.)

I think we would want to write expectations like:

```cc
SpanDataPtr consumer_span;
opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> producer_span_1, producer_span_2;
EXPECT_THAT(consumer_span, SpanHasLinks(producer_span_1, producer_span_2));
```

Sorry that I am leaving this as an exercise for you.... Feel free to open an issue to do this work and assign it to me :)

Also parent ID, span links might be something to include in the `operator<<(, SpanData)` you added....

_Originally posted by @dbolduc in https://github.com/googleapis/google-cloud-cpp/pull/11227#discussion_r1162927969_

dbolduc:
We decided not to use span links. Closing.

PR #11246, opened 2023-04-12
impl(otel): wrap `rest_internal::RestResponse`

coryan:
We need to wrap the `RestResponse` object to create `TracingRestHttpPayload`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11246)

### DIFF
--- a/google/cloud/internal/tracing_rest_response_test.cc
+++ b/google/cloud/internal/tracing_rest_response_test.cc
@@ -1,0 +1,91 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


google/cloud/internal/tracing_rest_response_test.cc
```diff
@@ -0,0 +1,91 @@
...
+  ASSERT_STATUS_OK(status);
+  EXPECT_THAT(*status, Eq(MockContents().size()));
+  status = payload->Read(absl::Span<char>(buffer.data(), buffer.size()));
+  ASSERT_STATUS_OK(status);
+  EXPECT_THAT(*status, Eq(0));
+
+  auto spans = span_catcher->GetSpans();
+  EXPECT_THAT(
+      spans,
+      Contains(AllOf(SpanHasInstrumentationScope(), SpanKindIsConsumer(),
```
dbolduc, line 78: It seems like the span being passed to the class should show up in an expectation. Consider:

```cc
UnorderedElementsAre(SpanNamed("HTTP/GET"), AllOf(...))
```

No action required. We could just call it part of #11236

dbolduc (approved):
I noticed two opportunities to use `EXPECT_EQ`. I do not comment on such trivialities anymore.

google/cloud/internal/tracing_rest_response_test.cc
coryan: Done.

PR #11252, opened 2023-04-12
fix(bigtable): treat server-side DeadlineExceeded as transient failure...

liujiongxin:
... We do need to respect the client-side deadline, but what if the server-side deadline is much smaller than client-side deadline. The C++ client should ensure that client-side deadline is respected while retrying on server-side deadline error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11252)

### DIFF
--- a/google/cloud/bigtable/rpc_retry_policy.h
+++ b/google/cloud/bigtable/rpc_retry_policy.h
@@ -42,6 +42,7 @@ struct SafeGrpcRetry {
   static inline bool IsTransientFailure(Status const& status) {
     auto const code = status.code();
     return code == StatusCode::kAborted || code == StatusCode::kUnavailable ||
+           code == StatusCode::kDeadlineExceeded ||
            google::cloud::internal::IsTransientInternalError(status);
   }
   static inline bool IsPermanentFailure(Status const& status) {


coryan:
This is contrary to aip/194:

Non-retryable codes
"DEADLINE_EXCEEDED: An application can set a deadline, which must be honored."

https://google.aip.dev/194

> ... (Java Client does so)

That is not an argument in favor of this PR... or against it.  The Java Client folks are free to do whatever they think it is right.  I will do what I think it is right for the C++ client.

Unless I hear of a good reason to accept this change, I am going to close the PR with prejudice.

coryan (changes requested):
This change is not well justified, please fix the description.

liujiongxin:
I updated the description, hopefully it justifies the PR.

coryan:
> ... We do need to respect the client-side deadline, but what if the server-side deadline is much smaller than client-side deadline.

That sounds like a server-side bug.  gRPC propagates the client-side deadline, the server should respect it.

liujiongxin:
I am ok with closing the PR then. I built a hermetic (synthetic) test that doesn't make the server respect client-side deadline. 

I hope the client Devs can discuss this issue, as inconsistency in the client libraries is not good.

coryan:
> I hope the client Devs can discuss this issue, as inconsistency in the client libraries is not good.

Agree.  When this comes up I generally prefer consistency across C++ libraries vs. consistency across libraries for one service.

PR #11261, opened 2023-04-13
impl(bigquery): Dataset api - Custom classes and unit tests for V2 Dataset resource

jsrinnn:
This PR contains the custom classes and json parsing for  the Dataset resource.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/dataset.h
+++ b/google/cloud/bigquery/v2/minimal/internal/dataset.h
@@ -158,14 +158,14 @@ struct Dataset {
   bool published = false;
   bool is_case_insensitive = false;
 
-  std::chrono::milliseconds default_table_expiration_ms =
+  std::chrono::milliseconds default_table_expiration =
       std::chrono::milliseconds(0);
-  std::chrono::milliseconds default_partition_expiration_ms =
+  std::chrono::milliseconds default_partition_expiration =
       std::chrono::milliseconds(0);
   std::chrono::system_clock::time_point creation_time;
   std::chrono::system_clock::time_point last_modified_time;
   ;
-  std::chrono::hours max_time_travel_hours = std::chrono::hours(0);
+  std::chrono::hours max_time_travel = std::chrono::hours(0);
 
   std::map<std::string, std::string> labels;
   std::vector<Access> access;
--- a/google/cloud/bigquery/v2/minimal/internal/dataset.h
+++ b/google/cloud/bigquery/v2/minimal/internal/dataset.h
@@ -1,0 +1,195 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_DATASET_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_BIGQUERY_V2_MINIMAL_INTERNAL_DATASET_H
+
+#include "google/cloud/version.h"
+#include <nlohmann/json.hpp>
+#include <chrono>
+#include <memory>
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// Disabling clang-tidy here as the namespace is needed for using the
+// NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT.
+using namespace nlohmann::literals;  // NOLINT
+
+auto const kDefaultTimepoint =
+    std::chrono::system_clock::time_point{std::chrono::milliseconds(0)};
+
+struct StorageBillingModel {
+  static StorageBillingModel UnSpecified();
+  static StorageBillingModel Logical();
+  static StorageBillingModel Physical();
+
+  std::string value;
+};
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(StorageBillingModel, value);
+
+struct TargetType {
+  static TargetType UnSpecified();
+  static TargetType Views();
+  static TargetType Routines();
+
+  std::string value;
+};
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(TargetType, value);
+
+struct TableFieldSchemaRoundingMode {
+  static TableFieldSchemaRoundingMode UnSpecified();
+  static TableFieldSchemaRoundingMode RoundHalfAwayFromZero();
+  static TableFieldSchemaRoundingMode RoundHalfEven();
+
+  std::string value;
+};
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(TableFieldSchemaRoundingMode,
+                                                value);
+
+struct DatasetReference {
+  std::string dataset_id;
+  std::string project_id;
+};
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(DatasetReference, dataset_id,
+                                                project_id);
+
+struct LinkedDatasetSource {
+  DatasetReference source_dataset;
+};
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(LinkedDatasetSource,
+                                                source_dataset);
+
+struct TableReference {
+  std::string project_id;
+  std::string dataset_id;
+  std::string table_id;
+};
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(TableReference, dataset_id,
+                                                project_id, table_id);
+
+struct RoutineReference {
+  std::string project_id;
+  std::string dataset_id;
+  std::string routine_id;
+};
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(RoutineReference, dataset_id,
+                                                project_id, routine_id);
+
+struct DatasetAccessEntry {
+  DatasetReference dataset;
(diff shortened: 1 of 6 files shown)


google/cloud/bigquery/v2/minimal/internal/dataset.h
```diff
@@ -0,0 +1,195 @@
...
+
+struct DatasetAccessEntry {
+  DatasetReference dataset;
+  std::vector<TargetType> target_types;
+};
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(DatasetAccessEntry, dataset,
+                                                target_types);
+
+struct Access {
+  bool operator==(Access const& other);
```
coryan, line 101: If you are going to define `operator==()` then you should define `operator!=` (it is easy: `inline bool operator!=(Access const& a, Access const& b) { return !(a == b); }`)

If this is going to be a member function (as you have it here) it should be `const`.

I think some folks recommend using a standalone function outside the class because conversion rules for the first argument work better with them.

coryan: I do not see an implementation for this function.  Maybe just remove it?

google/cloud/bigquery/v2/minimal/internal/dataset.h
```diff
@@ -0,0 +1,195 @@
...
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+// Disabling clang-tidy here as the namespace is needed for using the
+// NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT.
+using namespace nlohmann::literals;  // NOLINT
+
+auto const kDefaultTimepoint =
+    std::chrono::system_clock::time_point{std::chrono::milliseconds(0)};
```
coryan, line 34: This is odd... 

https://en.cppreference.com/w/cpp/chrono/time_point/time_point

The default constructor of `std::chrono::system_clock::time_point` does exactly what you are doing here.  Why do we need a new symbol for it?

google/cloud/bigquery/v2/minimal/internal/dataset.h
```diff
@@ -0,0 +1,195 @@
...
+  bool published = false;
+  bool is_case_insensitive = false;
+
+  std::chrono::system_clock::time_point default_table_expiration_ms =
+      kDefaultTimepoint;
+  std::chrono::system_clock::time_point default_partition_expiration_ms =
+      kDefaultTimepoint;
+  std::chrono::system_clock::time_point creation_time = kDefaultTimepoint;
+  std::chrono::system_clock::time_point last_modified_time = kDefaultTimepoint;
+  std::chrono::system_clock::time_point max_time_travel_hours =
```
coryan, line 172: If this field is really "time in hours" then you should represent it with `std::chrono::hours`.  That is the C++ type to represent durations in hour units.

google/cloud/bigquery/v2/minimal/internal/dataset.h
```diff
@@ -0,0 +1,195 @@
...
+  std::string description;
+  std::string type;
+  std::string location;
+  std::string default_collation;
+
+  bool published = false;
+  bool is_case_insensitive = false;
+
+  std::chrono::system_clock::time_point default_table_expiration_ms =
+      kDefaultTimepoint;
```
coryan, line 167: This is probably wrong.  If the time is `...::timepoint` then the units should not be in the name (`..._ms`).  And you are just writing a complicated expression to calll the default constructor. This would be equivalent:

```cc
  std::chrono::system_clock::time_point default_partition_expiration;
```

coryan (commented):
Generally good, but I think we may have the wrong representation for a few fields.  It seems some contain durations (lengths of time) but are being stored as time points (instants in time).

jsrinnn (commented):
Ack.

Addressed all feedback. PTAL and let me know if you think the fields are still misprepresented. Thanks for the review!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#-:-NTEjyt4B09Ht8MbJOW-:b-aj69a5)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.h` line 34 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#gh-1167560701:-NTEeON76HxcXKMWLgyz:b-n1zqj7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed59f8105542a87dd52e53368ec15cff23965c5d/google/cloud/bigquery/v2/minimal/internal/dataset.h#L34)):*

This is odd... 

https://en.cppreference.com/w/cpp/chrono/time_point/time_point

The default constructor of `std::chrono::system_clock::time_point` does exactly what you are doing here.  Why do we need a new symbol for it?
</blockquote></details>

Done.

Yup. Not needed.
___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.h` line 101 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#gh-1167559806:-NTEeVBw3cbvtBLQn96N:bro8mif) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed59f8105542a87dd52e53368ec15cff23965c5d/google/cloud/bigquery/v2/minimal/internal/dataset.h#L101)):*

I do not see an implementation for this function.  Maybe just remove it?
</blockquote></details>

Done.

Yup, Sorry forgot to cleanup. I was trying out different things at one point. Currently this is not needed. 

Thanks for catching.
___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.h` line 167 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#gh-1167562438:-NTEfJIBB86khnNa2l0_:b-60rgys) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed59f8105542a87dd52e53368ec15cff23965c5d/google/cloud/bigquery/v2/minimal/internal/dataset.h#L167)):*

This is probably wrong.  If the time is `...::timepoint` then the units should not be in the name (`..._ms`).  And you are just writing a complicated expression to calll the default constructor. This would be equivalent:

```cc
  std::chrono::system_clock::time_point default_partition_expiration;
```

</blockquote></details>

Done.

Ah...Got it. The dataset proto defines this field as: `The default lifetime of all tables in the dataset, in milliseconds`. Similarly for `default_partition_expiration_ms`

 I was initially thinking of both of the above as a Timepoint but giving it more thought, they maybe better represented as `std::chrono::milliseconds`? 

I have modified both of them to be as mentioned above but let me know if you think otherwise.
___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.h` line 172 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#gh-1167562282:-NTEja8t3lGGGCNl1CWY:b24h9pk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed59f8105542a87dd52e53368ec15cff23965c5d/google/cloud/bigquery/v2/minimal/internal/dataset.h#L172)):*

If this field is really "time in hours" then you should represent it with `std::chrono::hours`.  That is the C++ type to represent durations in hour units.
</blockquote></details>

Done.

According to the proto docs, this fields is described as `Defines the time travel window in hours` so yeah std::chrono::hours would be better suited here. Thanks!

coryan (approved):
I think the field names should be changed.  The units are part of the type, no need to repeat them in the field name.  I realize that makes the C++ field name different from the proto/JSON field.  Please fix and then merge, we can chat if you disagree.

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#-:-NTFMxl5AXh-4HOF80AE:bxekwrm)* status: 2 of 6 files reviewed, 2 unresolved discussions (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.h` line 167 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#gh-1167562438:-NTFMKFoCVhsKQbEXHIg:b4mne00) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed59f8105542a87dd52e53368ec15cff23965c5d/google/cloud/bigquery/v2/minimal/internal/dataset.h#L167)):*

Done.

Ah...Got it. The dataset proto defines this field as: `The default lifetime of all tables in the dataset, in milliseconds`. Similarly for `default_partition_expiration_ms`

 I was initially thinking of both of the above as a Timepoint but giving it more thought, they maybe better represented as `std::chrono::milliseconds`? 

I have modified both of them to be as mentioned above but let me know if you think otherwise.
</blockquote></details>

The name should simply be `default_table_expiration`.  The units are part of the type.
___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.h` line 172 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#gh-1167562282:-NTFMiGn66U7_qvY-fbB:b-r1bi4d) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed59f8105542a87dd52e53368ec15cff23965c5d/google/cloud/bigquery/v2/minimal/internal/dataset.h#L172)):*

Done.

According to the proto docs, this fields is described as `Defines the time travel window in hours` so yeah std::chrono::hours would be better suited here. Thanks!
</blockquote></details>

Likewise, the name should just be `max_time_travel`.

jsrinnn (commented):
Ack.

Addressed remaining feedback and modified the field names, in the latest push. Thanks!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#-:-NTFO3nNENEANxUeRpwn:b-3fzohi)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.h` line 167 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#gh-1167562438:-NTFNwmZ8I7bGVgl9KHL:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed59f8105542a87dd52e53368ec15cff23965c5d/google/cloud/bigquery/v2/minimal/internal/dataset.h#L167)):*

The name should simply be `default_table_expiration`.  The units are part of the type.
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.h` line 172 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11261#gh-1167562282:-NTFO18_8k__tbYSjW57:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/ed59f8105542a87dd52e53368ec15cff23965c5d/google/cloud/bigquery/v2/minimal/internal/dataset.h#L172)):*

Likewise, the name should just be `max_time_travel`.
</blockquote></details>

Done.

Issue #11253, opened 2023-04-12
Generalize explicit routing parameter pattern matching


dbolduc:
The substitution `"*"` -> `"[^/]+"` assumes that the character after an asterisk is always a slash.

https://github.com/googleapis/google-cloud-cpp/blob/9406467ef5504687a39443b168ef58a5ce09964f/generator/internal/descriptor_utils.cc#L817

To be more general, we should map `"*" + c` -> `"[^" + c + "]+"`.

Off the top of my head, I do not know if there is a slick way to do this with `absl::StrReplaceAll()`. But note that efficiency is not a concern here. This transformation happens at code generation time, not at customer compile or run time.

PR #11272, opened 2023-04-13
impl(GCS+gRPC): alternative routing format for upload id

coryan:
This is the analog of #11253 for one RPC that needs hand-crafted metadata decoration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11272)

### DIFF
--- a/google/cloud/storage/internal/grpc_configure_client_context.cc
+++ b/google/cloud/storage/internal/grpc_configure_client_context.cc
@@ -29,10 +29,13 @@ void ApplyRoutingHeaders(
 
 void ApplyRoutingHeaders(grpc::ClientContext& context,
                          storage::internal::UploadChunkRequest const& request) {
-  static auto* bucket_regex =
+  static auto* slash_format =
       new std::regex{"(projects/[^/]+/buckets/[^/]+)/.*", std::regex::optimize};
-  std::smatch match;
-  if (std::regex_match(request.upload_session_url(), match, *bucket_regex)) {
+  static auto* colon_format =
+      new std::regex{"(projects/[^/]+/buckets/[^:]+):.*", std::regex::optimize};
+  for (auto const* re : {slash_format, colon_format}) {
+    std::smatch match;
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/grpc_configure_client_context.cc
```diff
@@ -29,10 +29,13 @@ void ApplyRoutingHeaders(
...
 void ApplyRoutingHeaders(grpc::ClientContext& context,
                          storage::internal::UploadChunkRequest const& request) {
-  static auto* bucket_regex =
+  static auto* slash_format =
       new std::regex{"(projects/[^/]+/buckets/[^/]+)/.*", std::regex::optimize};
-  std::smatch match;
-  if (std::regex_match(request.upload_session_url(), match, *bucket_regex)) {
+  static auto* colon_format =
+      new std::regex{"(projects/[^/]+/buckets/[^:]+):.*", std::regex::optimize};
+  for (auto const* re : {slash_format, colon_format}) {
```
dbolduc, line 36: (due dilligence) why do we need two formats?

I only see `UploadChunk` calling `rpc WriteObject`, which doesn't / can't have routing annotations. I don't remember the source of this regex.

coryan: The service is going to change from format 1 to format 2.  I don't want to have to perfectly coordinate which format works.

dbolduc (approved):
https://github.com/googleapis/google-cloud-cpp/blob/48449bf1061fb037deb245a2e6febccb25469fb4/google/cloud/storage/internal/grpc_configure_client_context_test.cc#L136

google/cloud/storage/internal/grpc_configure_client_context.cc
coryan: Also `WriteObject()` lacks the annotations because it is a streaming RPC. That is why we have to manually insert them too.

coryan:
Added a test, thanks.

PR #11274, opened 2023-04-13
impl(otel): single span for each HTTP request

coryan:
Using a separate span to handle the HTTP response / payload is not a good representation of the flow. A single span that lasts (but it is not necessarily active) until the full payload is received seems like a better model of the actual code flow.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11274)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11274)* status: 0 of 8 files reviewed, all discussions resolved

dbolduc (approved):

PR #11281, opened 2023-04-13
doc: announce servicecontrol/v2 support in the CHANGELOG

devbww:
Should have been a part of #11180.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11281)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,10 @@ the APIs in these libraries are stable, and are ready for production use.
 - [ConfidentialComputing API](/google/cloud/confidentialcomputing/README.md)
 - [Workstations API](/google/cloud/workstations/README.md)
 
+### [Service Control](/google/cloud/servicecontrol/README.md)
+
+The library has been expanded to include the v2 service.
+
 ### Other Changes
 
 **Bazel Testing**: starting with v2.10.0 we only test with Bazel >= 5.4.0. We


dbolduc (approved):

CHANGELOG.md
```diff
@@ -129,6 +129,10 @@ the APIs in these libraries are stable, and are ready for production use.
 - [ConfidentialComputing API](/google/cloud/confidentialcomputing/README.md)
 - [Workstations API](/google/cloud/workstations/README.md)
 
+### [Service Control](/google/cloud/servicecontrol/README.md)
+
+The library has been expanded to include the v2 service.
```
devbww, line 134: > Should have been a part of #11180.

On third thoughts, I may have been right the first time.  This will already show up in the generated CHANGELOG (under _"Common Libraries"_) as ...

- feat(servicecontrol): add v2 service ([#11180](https://github.com/googleapis/google-cloud-cpp/pull/11180))

But then again we add special announcements for "New Libraries" even though they also show up as, for example, ...

- feat(workstations): generate library ([#11199](https://github.com/googleapis/google-cloud-cpp/pull/11199))

So, my question is: Do/should we treat _"add vX service"_ differently from _"generate library"_ in CHANGELOG entries?  Your opinion sought.

coryan: I think major features deserve some razzmatazz.  I still think of new libraries and libraries supporting new versions of a service as a major feature (though the effort is hopefully shrinking).  My preference would be to keep doing these special notes, but I could be convinced otherwise (I ran out of Octomore, just saying).

devbww: I'm sold.  Razzamatazz for new versions it is.  Thanks.

dbolduc: > should we treat "add vX service" differently from "generate library" 

They aren't exactly the same. Maybe we should do a list of "newly generated" and a list of "newly expanded" at the top of the CHANGELOG. I don't feel strongly, though.

> This will already show up (under "Common Libraries")

I don't think `feat(servicecontrol): ...` belongs in "Common Libraries".

I like to announce new libraries / new services within libraries in the generation PR, but mostly out of habit.

PR #11282, opened 2023-04-13
feat(generator): add ProcessMethodsRequestsAndResponses

scotthart:
part of the work for #10980

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11282)

### DIFF
--- a/generator/internal/discovery_to_proto.cc
+++ b/generator/internal/discovery_to_proto.cc
@@ -218,6 +218,46 @@ StatusOr<DiscoveryTypeVertex> SynthesizeRequestType(
   return DiscoveryTypeVertex(id, synthesized_request);
 }
 
+Status ProcessMethodRequestsAndResponses(
+    std::map<std::string, DiscoveryResource>& resources,
+    std::map<std::string, DiscoveryTypeVertex>& types) {
+  for (auto& resource : resources) {
+    std::string resource_name = resource.first;
+    auto const& methods = resource.second.json().find("methods");
+    for (auto m = methods->begin(); m != methods->end(); ++m) {
+      auto response_type_name =
+          DetermineAndVerifyResponseTypeName(*m, resource.second, types);
+      if (!response_type_name) {
+        return Status(response_type_name.status().code(),
+                      response_type_name.status().message(),
+                      GCP_ERROR_INFO()
+                          .WithMetadata("resource", resource_name)
+                          .WithMetadata("method", m.key())
+                          .Build(response_type_name.status().code()));
+      }
+
+      if (m->contains("parameters")) {
+        auto request_type = SynthesizeRequestType(*m, resource.second,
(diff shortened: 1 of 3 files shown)


dbolduc (approved):
Hilariously, I only have style nits for you :)

generator/internal/discovery_to_proto.cc
```diff
@@ -218,6 +218,46 @@ StatusOr<DiscoveryTypeVertex> SynthesizeRequestType(
   return DiscoveryTypeVertex(id, synthesized_request);
 }
 
+Status ProcessMethodRequestsAndResponses(
+    std::map<std::string, DiscoveryResource>& resources,
+    std::map<std::string, DiscoveryTypeVertex>& types) {
+  for (auto& resource : resources) {
+    std::string resource_name = resource.first;
```
dbolduc, line 225: **optional:** Should we just capitalize the resource name here? (Instead of doing it for each `method`). The only difference is the error info.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11282)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @dbolduc)

___
*[`generator/internal/discovery_to_proto.cc` line 225 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11282#gh-1166262044:-NT05UFH3PivJ3qPQMmg:b-4x7cjs) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bc398a41d2c9edf91d1ad80197af5c651c163512/generator/internal/discovery_to_proto.cc#L225)):*

**optional:** Should we just capitalize the resource name here? (Instead of doing it for each `method`). The only difference is the error info.
</blockquote></details>

Moved the capitalization outside the loop, and kept the lowercase version of the string in the error message.
___
*[`generator/internal/discovery_to_proto.cc` line 239 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11282#gh-1166246762:-NT05bbR1NyGda02KFm5:b-w4664f) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bc398a41d2c9edf91d1ad80197af5c651c163512/generator/internal/discovery_to_proto.cc#L239)):*

**optional:** style nit: avoid unnecessary nesting

```
if (!m->contains("parameters")) continue;
// etc.
```
</blockquote></details>

I worry that a single continue line could get lost, where an indented block is more visually obvious. I looked at our style guide and didn't find a recommendation one way or the other, but I did find some guidance from the c++ core guidelines that prefers keeping the nested if block:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es77-minimize-the-use-of-break-and-continue-in-loops

PR #11285, opened 2023-04-14
feat(generator): allow for grpc transport to be disabled for a service

scotthart:
This PR is organized into multiple commits.

This change is necessary to support the use case where the service is defined by protobufs, supports a REST endpoint, but does NOT support a gRPC endpoint (e.q. sql aka sqladmin).

It was necessary to juggle the includes in the generated connection.h to no longer include the stub.h file, but instead direftly include the files it was getting transitively through the stub.h.

As the changes were all either include oriented or using a generated REST type in lieu of a generated gRPC type, the testing consists of being able to compile the generated golden code. Adding the sql suite of services will provide integration testing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11285)


generator/internal/connection_generator.cc
```diff
@@ -64,7 +65,8 @@ Status ConnectionGenerator::GenerateHeader() {
        IsExperimental() ? "google/cloud/experimental_tag.h" : "",
        "google/cloud/version.h"});
   HeaderSystemIncludes(
-      {HasLongrunningMethod() ? "google/longrunning/operations.grpc.pb.h" : "",
+      {vars("proto_grpc_header_path"), vars("additional_pb_header_paths"),
```
coryan, line 68: Consider `vars("proto_header_path")` instead of `vars("proto_grpc_header_path")`.  It seems we do not need the gRPC code, just the proto messages.

generator/integration_tests/test2.proto
```diff
@@ -0,0 +1,35 @@
...
+syntax = "proto3";
+
+package google.test.rest.only.v1;
+
+import "google/api/annotations.proto";
+import "google/api/client.proto";
+import "google/protobuf/empty.proto";
+
+// Service to test services that do not support gRPC transport.
+service GoldenRestOnly {
```
coryan, line 24: Thank you for creating a minimal proto file to test this.  Much easier to grok.

generator/internal/connection_generator.cc
```diff
@@ -384,29 +390,30 @@ future<StatusOr<$response_type$>>
         __FILE__, __LINE__);
   }
 
-  auto endpoint_location_style = EndpointLocationStyle();
+  if (HasGenerateGrpcTransport()) {
```
coryan, line 393: Consider moving these large `if()` blocks to a function.

coryan (approved):

generator/integration_tests/test2.proto
dbolduc: +1

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11285)* status: 0 of 260 files reviewed, 3 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[`generator/integration_tests/test2.proto` line 24 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11285#gh-1166770340:-NT-j90nE5zsKWFpPm3I:b1lau23) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8fd0deb713ea7be201aa5fbf9fe8973a40d1b406/generator/integration_tests/test2.proto#L24)):*

+1
</blockquote></details>

It's easier now that the golden file generation got moved into the generate-libraries build.
___
*[`generator/internal/connection_generator.cc` line 68 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11285#gh-1166760840:-NT-j6VGEG7iubqTJfND:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8fd0deb713ea7be201aa5fbf9fe8973a40d1b406/generator/internal/connection_generator.cc#L68)):*

Consider `vars("proto_header_path")` instead of `vars("proto_grpc_header_path")`.  It seems we do not need the gRPC code, just the proto messages.
</blockquote></details>

Done.
___
*[`generator/internal/connection_generator.cc` line 393 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11285#gh-1166778616:-NT-j77v4L1oNOdvtgsu:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8fd0deb713ea7be201aa5fbf9fe8973a40d1b406/generator/internal/connection_generator.cc#L393)):*

Consider moving these large `if()` blocks to a function.
</blockquote></details>

Done.

PR #11286, opened 2023-04-14
impl(otel): set Recordable identity and timestamps

dbolduc:
Part of the work for #11156 

Implement the identity and timestamp setters for a `Recordable`. This change should make the class usable with Cloud Trace.

### DIFF
--- a/google/cloud/opentelemetry/internal/recordable.cc
+++ b/google/cloud/opentelemetry/internal/recordable.cc
@@ -14,6 +14,7 @@
 
 #include "google/cloud/opentelemetry/internal/recordable.h"
 #include "google/cloud/internal/absl_str_cat_quiet.h"
+#include "google/cloud/internal/time_utils.h"
 
 namespace google {
 namespace cloud {
@@ -148,8 +149,28 @@ google::devtools::cloudtrace::v2::Span&& Recordable::as_proto() && {
 }
 
 void Recordable::SetIdentity(
-    opentelemetry::trace::SpanContext const& /*span_context*/,
-    opentelemetry::trace::SpanId /*parent_span_id*/) noexcept {}
+    opentelemetry::trace::SpanContext const& span_context,
+    opentelemetry::trace::SpanId parent_span_id) noexcept {
+  std::array<char, 2 * opentelemetry::trace::TraceId::kSize> hex_trace_buf;
+  span_context.trace_id().ToLowerBase16(hex_trace_buf);
+  std::string const hex_trace(hex_trace_buf.data(),
+                              2 * opentelemetry::trace::TraceId::kSize);
+
+  std::array<char, 2 * opentelemetry::trace::SpanId::kSize> hex_span_buf;
+  span_context.span_id().ToLowerBase16(hex_span_buf);
+  std::string const hex_span(hex_span_buf.data(),
+                             2 * opentelemetry::trace::SpanId::kSize);
+
+  std::array<char, 2 * opentelemetry::trace::SpanId::kSize> hex_parent_span_buf;
+  parent_span_id.ToLowerBase16(hex_parent_span_buf);
+  std::string const hex_parent_span(hex_parent_span_buf.data(),
+                                    2 * opentelemetry::trace::SpanId::kSize);
+
+  span_.set_name(project_.FullName() + "/traces/" + hex_trace + "/spans/" +
+                 hex_span);
+  span_.set_span_id(hex_span);
+  span_.set_parent_span_id(hex_parent_span);
+}
 
 void Recordable::SetAttribute(
     opentelemetry::nostd::string_view key,
@@ -184,9 +205,15 @@ void Recordable::SetResource(
     opentelemetry::sdk::resource::Resource const& /*resource*/) noexcept {}
 
 void Recordable::SetStartTime(
-    opentelemetry::common::SystemTimestamp /*start_time*/) noexcept {}
+    opentelemetry::common::SystemTimestamp start_time) noexcept {
+  *span_.mutable_start_time() = internal::ToProtoTimestamp(start_time);
+}
 
-void Recordable::SetDuration(std::chrono::nanoseconds /*duration*/) noexcept {}
+void Recordable::SetDuration(std::chrono::nanoseconds duration) noexcept {
+  auto const end_time =
+      internal::ToChronoTimePoint(span_.start_time()) + duration;
+  *span_.mutable_end_time() = internal::ToProtoTimestamp(end_time);
+}
 
 void Recordable::SetInstrumentationScope(
     opentelemetry::sdk::instrumentationscope::InstrumentationScope const&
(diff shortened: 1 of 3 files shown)


google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -148,8 +150,28 @@ google::devtools::cloudtrace::v2::Span&& Recordable::as_proto() && {
 }
 
 void Recordable::SetIdentity(
-    opentelemetry::trace::SpanContext const& /*span_context*/,
-    opentelemetry::trace::SpanId /*parent_span_id*/) noexcept {}
+    opentelemetry::trace::SpanContext const& span_context,
+    opentelemetry::trace::SpanId parent_span_id) noexcept {
```
devbww, line 154: I'm not sure how you can assert `noexcept` when you do things like construct strings.

google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -148,8 +150,28 @@ google::devtools::cloudtrace::v2::Span&& Recordable::as_proto() && {
...
+  std::array<char, 2 * opentelemetry::trace::TraceId::kSize> hex_trace_buf;
+  span_context.trace_id().ToLowerBase16(hex_trace_buf);
+  std::string const hex_trace(hex_trace_buf.data(),
+                              2 * opentelemetry::trace::TraceId::kSize);
```
devbww, line 158: `hex_trace_buf.size()` perhaps, to avoid repeating the template parameter, and to perpetuate the "data/size" pairing convention.

google/cloud/opentelemetry/internal/recordable_test.cc
```diff
@@ -242,6 +270,30 @@ TEST(Recordable, SetAttributeRespectsLimit) {
   EXPECT_EQ(attributes.dropped_attributes_count(), 1);
 }
 
+TEST(Recordable, SetStartTime) {
+  auto start = std::chrono::system_clock::now();
+  auto expected = absl::FromChrono(start);
+
+  auto rec = Recordable(Project(kProjectId));
+  rec.SetStartTime(start);
```
devbww, line 278: For what it's worth, it looks like `opentelemetry::common::SystemTimestamp` has nanosecond resolution, as does `absl::Time` and `google::protobuf::Timestamp`, but `std::chrono::system_clock::time_point` may not on some platforms (notably Windows and macOS), so you should be wary of conversions through `time_point`.

google/cloud/opentelemetry/internal/recordable_test.cc
```diff
@@ -242,6 +270,30 @@ TEST(Recordable, SetAttributeRespectsLimit) {
...
+  auto proto = std::move(rec).as_proto();
+  auto actual = internal::ToAbslTime(proto.start_time());
+  EXPECT_EQ(actual, expected);
+}
+
+TEST(Recordable, SetDuration) {
+  auto start = std::chrono::system_clock::now();
+  auto duration = std::chrono::nanoseconds(12345);
+  auto expected = absl::FromChrono(start) + absl::FromChrono(duration);
```
devbww, line 287: Here, for example, and in `SetDuration()`, it is important to do the addition in the `absl::Time` domain.

devbww (approved):

google/cloud/opentelemetry/internal/recordable.cc
coryan: You can't.  These are virtual functions and we need to respect the specification.

I [tried (and failed)](https://github.com/open-telemetry/opentelemetry-cpp/issues/4#issuecomment-544949758) to convince these folks to drop the `noexcept`.

It seems that we will need to capture all exceptions.  Maybe something like this (


```cc
// in some header...
bool NoExceptAction(absl::FunctionRef<void()> action) noexcept;

// in some cc
#if GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
bool NoExceptAction(absl::FunctionRef<void()> action) noexcept try {
  action();
  return true;
} catch(...) {
  return false;
}
#else
bool NoExceptAction(absl::FunctionRef<void()> action) noexcept {
  action(); // Will crash, would have crashed anyway.
  return true;
}
#endif

// In Recordable.cc
void Recordable::SetIdentity(...) {
  usable_ = usable_ & NoExceptAction(SetIdentityImpl(...));
}
```

Then we will drop (not send to Cloud Trace) any `Recordable`s where `usable_` is `false`.

This is all for a TODO bug and future PRs.

dbolduc: Opened #11288

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: Good idea. Fixed.

google/cloud/opentelemetry/internal/recordable_test.cc
dbolduc: Ah.. thanks for investigating. TIL about `system_clock::time_point` not having ns resolution.

I was converting through `chrono`, using this: https://github.com/open-telemetry/opentelemetry-cpp/blob/7cb7654552d68936d70986bc2ee67f3cc3e0b469/api/include/opentelemetry/common/timestamp.h#L53

I updated the `SetStartTime` implementation to convert through `absl::Time`. Do you mind taking another look?

google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -184,9 +204,19 @@ void Recordable::SetResource(
     opentelemetry::sdk::resource::Resource const& /*resource*/) noexcept {}
 
 void Recordable::SetStartTime(
-    opentelemetry::common::SystemTimestamp /*start_time*/) noexcept {}
+    opentelemetry::common::SystemTimestamp start_time) noexcept {
+  // std::chrono::system_clock may not have nanosecond resolution on some
+  // platforms, so we avoid using it for conversions between OpenTelemetry
+  // time and Protobuf time.
```
devbww, line 210: Looks good.  It seems like `opentelemetry::common::SystemTimestamp`'s implicit conversion to `std::chrono::system_clock::time_point` is a rounding accident waiting to happen.

google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -184,9 +204,19 @@ void Recordable::SetResource(
...
+  auto t = absl::FromUnixNanos(start_time.time_since_epoch().count());
+  *span_.mutable_start_time() = internal::ToProtoTimestamp(std::move(t));
```
devbww, line 212: `absl::Time` is one of those always-pass-by-value exceptions, so drop the `move()`.

devbww (approved):

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: Yeah, thanks for the warning.

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: Done.

Issue #11201, opened 2023-04-07
Some spanner instances do not support DataBoost?


coryan:
I just got this in a PR.  Since we pick the spanner instances at random it may not affect all builds:

```
Step #3: 2023-04-07T15:02:53.975881932Z [DEBUG] <139635551693376> Finish() >> PERMISSION_DENIED: This instance (projects/cloud-cpp-testing-resources/instances/test-instance-15) is not allowed to use DataBoost at this time. Please contact cloud-spanner-sa-preview@google.com for more information. (google/cloud/spanner/internal/logging_result_set_reader.cc:53)
```

* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/1f352100-6781-4147-bf9d-97b97d389703;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/11200/7db9ef34c2689c9e65f3ab392fbe2468b23c2bdc/fedora-37-cmake-integration-production/log-1f352100-6781-4147-bf9d-97b97d389703.txt

devbww:
It turns out that currently:
1. Spanner instances have to be allow-listed to use DataBoost, and
2. not all regions have "DataBoost OA capacity".

All our _"test-instance-**N**"_ instances have been allow listed, but the _"labels.samples:yes"_ subset are all in us-east1, which does not have capacity.

Given that we can't even discern a difference between DataBoost on and off, it doesn't seem worth it to temporarily shift instances/labels around.  Instead, I'll re-enable the samples but with an allowance for kPermissionDenied ... then they will start working whenever DataBoost is generally available.  Finally, we can remove the allowance any time after that.

PR #11293, opened 2023-04-15
ci(spanner): re-enable DataBoost samples but with workaround

devbww:
The problem with the DataBoost samples was that there is no capacity in the region used by our Spanner instances labelled as "samples:yes".

So, lets re-enable them, but with a specific allowance for seeing a `kPermissionDenied` status code.  Then, they will automatically start working when capacity is allocated, and after which we can remove the workaround.

Use `//!` to keep the workaround out of the published snippets.

Related to #11201.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11293)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -2500,6 +2500,10 @@ void UsePartitionQuery(google::cloud::spanner::Client client) {
   for (auto const& partition : *partitions) {
     auto rows = client.ExecuteQuery(partition);
     for (auto& row : spanner::StreamOf<RowType>(rows)) {
+      if (row.status().code() ==                           //! TODO(#11201)
+          google::cloud::StatusCode::kPermissionDenied) {  //! TODO(#11201)
+        continue;                                          //! TODO(#11201)
+      }                                                    //! TODO(#11201)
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -2500,6 +2500,10 @@ void UsePartitionQuery(google::cloud::spanner::Client client) {
   for (auto const& partition : *partitions) {
     auto rows = client.ExecuteQuery(partition);
     for (auto& row : spanner::StreamOf<RowType>(rows)) {
+      if (row.status().code() ==                           //! TODO(#11201)
```
coryan, line 2503: The `//!` comment is a sigil for Doxygen.  I am not sure what effect, if any, would this have in how the samples are rendered.

coryan (approved):

google/cloud/spanner/samples/samples.cc
devbww: Followed up offline to highlight that it should have an effect.  And while it will in this particular case, it also looks like the required support has not been applied to all tags.  I'll chase that down internally.  Thanks.

PR #11295, opened 2023-04-17
cleanup(otel): prefer `ResultOf` over `IsProtoEqual`

dbolduc:
Nicer, if you ask me. I think we would get better error messages, too.

Matching on `v2::Span::Attributes` may look like overkill, but note that the proto also has repeated `links` and `events` fields, which can have `attributes`. So matching on `v2::Span::Attributes` will be necessary when we implement events and links.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11295)


dbolduc:
I might make another push to this. No need to review just yet.

dbolduc:
Ok, the PR is ready now.

PR #11297, opened 2023-04-17
feat(generator): add AssignResourcesAndTypesToFiles

scotthart:
part of the work for #10980


generator/internal/discovery_to_proto.cc
```diff
@@ -274,6 +274,36 @@ std::vector<DiscoveryFile> CreateFilesFromResources(
...
+    std::map<std::string, DiscoveryResource> const& resources,
+    std::map<std::string, DiscoveryTypeVertex> const& types,
+    std::string const& product_name, std::string const& version,
+    std::string const& output_path) {
+  auto files =
+      CreateFilesFromResources(resources, product_name, version, output_path);
+
+  // TODO(#11190): For the first phase of implementation, we create one proto
+  // per resource and its request types. For the remainder of the types, we
+  // dump them all into one internal/common.proto file. This should be reworked
```
dbolduc, line 287: So is the `internal/` to make sure people don't `#include "google/cloud/foo/v1/internal/common.pb.h"` directly?

(I see that this will be changing, so no big deal anyway. I am just curious.)

dbolduc (approved):

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11297)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @dbolduc)

___
*[`generator/internal/discovery_to_proto.cc` line 287 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11297#gh-1169308682:-NTGhSETAE3MMOiodsdQ:bdz1xip) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e09445c9b42545a3e7a6d71c240fc51d86d68d01/generator/internal/discovery_to_proto.cc#L287)):*

So is the `internal/` to make sure people don't `#include "google/cloud/foo/v1/internal/common.pb.h"` directly?

(I see that this will be changing, so no big deal anyway. I am just curious.)
</blockquote></details>

Correct. Because as the service changes the contents of any particular .proto file in `internal` could change. By only including the `service.pb.h` (which will have the correct imports), the user can be assured to always get the necessary definitions transitively.
___
*[`generator/internal/discovery_to_proto_test.cc` line 910 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11297#gh-1169304314:-NTGhtVj2alysBnEJYqA:bi4knti) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/e09445c9b42545a3e7a6d71c240fc51d86d68d01/generator/internal/discovery_to_proto_test.cc#L910)):*

**optional**: Use container matchers for better error messages.

```cc
auto file_matcher = []{ ... };
EXPECT_THAT(result, ElementsAre(file_matcher(...), file_matcher(...)));
```

e.g.

https://github.com/googleapis/google-cloud-cpp/blob/13cf66d98a76a9b3aa16eb50743fd6d8a18b8bbc/google/cloud/storage/internal/grpc_hmac_key_request_parser_test.cc#L168-L192

FWIW, I am not convinced that the value added by using container matchers is worth the effort it takes to write them.
</blockquote></details>

Maybe I'll try some custom container matchers the next time...

PR #11302, opened 2023-04-17
impl(otel): Recordable add links

dbolduc:
Part of the work for #11156 

The `opentelemetry::common::KeyValueIterable` interface is... something...

https://github.com/open-telemetry/opentelemetry-cpp/blob/7cb7654552d68936d70986bc2ee67f3cc3e0b469/api/include/opentelemetry/common/key_value_iterable.h#L13-L34

I could not find any helpers that they offer to construct one easily, so I had to extend the class myself in the test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11302)

### DIFF
--- a/google/cloud/opentelemetry/internal/recordable.cc
+++ b/google/cloud/opentelemetry/internal/recordable.cc
@@ -222,7 +222,7 @@ void Recordable::AddLink(
   auto& proto = *link.mutable_attributes();
   attributes.ForEachKeyValue(
       [&proto](opentelemetry::nostd::string_view key,
-               opentelemetry::common::AttributeValue value) {
+               opentelemetry::common::AttributeValue const& value) {
         AddAttribute(proto, key, value, kSpanLinkAttributeLimit);
         return proto.attribute_map().size() != kSpanLinkAttributeLimit;
       });
--- a/google/cloud/opentelemetry/internal/recordable.cc
+++ b/google/cloud/opentelemetry/internal/recordable.cc
@@ -202,8 +202,35 @@ void Recordable::AddEvent(
     opentelemetry::common::KeyValueIterable const& /*attributes*/) noexcept {}
 
 void Recordable::AddLink(
-    opentelemetry::trace::SpanContext const& /*span_context*/,
-    opentelemetry::common::KeyValueIterable const& /*attributes*/) noexcept {}
+    opentelemetry::trace::SpanContext const& span_context,
+    opentelemetry::common::KeyValueIterable const& attributes) noexcept {
+  // Accept the first N links. Drop the rest.
+  auto& links = *span_.mutable_links();
+  if (links.link().size() == kSpanLinkLimit) {
+    links.set_dropped_links_count(links.dropped_links_count() + 1);
+    return;
+  }
+
+  std::array<char, 2 * opentelemetry::trace::TraceId::kSize> hex_trace_buf;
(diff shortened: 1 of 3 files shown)


google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -202,8 +202,35 @@ void Recordable::AddEvent(
...
+
+  auto& link = *links.add_link();
+  link.set_span_id(hex_span);
+  link.set_trace_id(hex_trace);
+
+  auto& proto = *link.mutable_attributes();
+  attributes.ForEachKeyValue(
+      [&proto](opentelemetry::nostd::string_view key,
+               opentelemetry::common::AttributeValue value) {
+        AddAttribute(proto, key, value, kSpanLinkAttributeLimit);
```
coryan, line 230: Shouldn't you return `false` once there are too many attributes in `&proto`?

coryan (approved):

google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -202,8 +202,35 @@ void Recordable::AddEvent(
...
+  std::array<char, 2 * opentelemetry::trace::TraceId::kSize> hex_trace_buf;
+  span_context.trace_id().ToLowerBase16(hex_trace_buf);
+  std::string const hex_trace(hex_trace_buf.data(), hex_trace_buf.size());
+
+  std::array<char, 2 * opentelemetry::trace::SpanId::kSize> hex_span_buf;
+  span_context.span_id().ToLowerBase16(hex_span_buf);
+  std::string const hex_span(hex_span_buf.data(), hex_span_buf.size());
+
+  auto& link = *links.add_link();
+  link.set_span_id(hex_span);
```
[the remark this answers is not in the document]
dbolduc: Updated.

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: Ugh, so that is clearly the intention of the API. I convinced myself I did not need to because we *could* have attributes with repeated keys in the `KeyValueIterable`. But there is really no reason (in this case) why last should win over first. Also, it is just better to stop when the map is full in case there are 10 gajillion attributes.

Stopped iteration early. Added a test to verify it stops early.

coryan: Seems like you could change `AddAttribute()` to simplify this further:

```cc
  attributes.ForEachKeyValue(
      [&proto](opentelemetry::nostd::string_view key, opentelemetry::common::AttributeValue value) {
        return AddAttribute(proto, key, std::move(value), kSpanLinkAttributeLimit);
     });
```

nit: do you need to copy `value`?  Can you consume `AttributeValue const&`?

dbolduc: > Seems like you could change `AddAttribute()` to simplify this further:

I am going to pass for now. I could have it return whether the map is full. That would save me one line of code in a future PR?

`AddAttribute()` cannot just return whether the attribute was added or not. e.g. We may drop an attribute because a key is too long. We want to keep iterating in that case.

> do you need to copy `value`? Can you consume `AttributeValue const&`?

I think we can consume `AttributeValue const&` given that `ForEachKeyValue(...)` is `const`. Thanks.

PR #11315, opened 2023-04-18
ci: publish-docs always run DocFX generation

coryan:
Running the `doxygen2docfx` tool in each PR will ensure we do not regress w.r.t. supporting all the libraries in the tool. We also need to run the tool to (eventually) upload the documents to start the publication pipeline, so this is a good place to run them.

I had to change how the tool is integrated into CMake. It cannot be a "feature" (an element in the `GOOGLE_CLOUD_CPP_ENABLE` list) because CMake needs to configure the `docfx/` directory before the `google/cloud/` directory and the features always go afted `google/cloud/`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11315)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,10 @@ endif ()
 option(GOOGLE_CLOUD_CPP_GENERATE_DOXYGEN "Generate Doxygen documentation" OFF)
 mark_as_advanced(GOOGLE_CLOUD_CPP_GENERATE_DOXYGEN)
 
+# Compile the DocFX tool.
+option(GOOGLE_CLOUD_CPP_INTERNAL_DOCFX "Compile the docfx tool. Only google-cloud-cpp developers are expected to use this flag." OFF)
+mark_as_advanced(GOOGLE_CLOUD_CPP_INTERNAL_DOCFX)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -99,6 +99,10 @@ endif ()
 option(GOOGLE_CLOUD_CPP_GENERATE_DOXYGEN "Generate Doxygen documentation" OFF)
 mark_as_advanced(GOOGLE_CLOUD_CPP_GENERATE_DOXYGEN)
 
+# Compile the DocFX tool.
+option(GOOGLE_CLOUD_CPP_INTERNAL_DOCFX "Compile the docfx tool. Only google-cloud-cpp developers are expected to use this flag." OFF)
```
devbww, line 103: Nit: I'm surprised `cmake-format` was happy with such a long line.

devbww (approved):

CMakeLists.txt
coryan: Me too.  I think we are not formatting the top-level CMake file at all, I will send a separate PR to fix that.

PR #11318, opened 2023-04-18
impl(generator): add alternate pagination support

scotthart:
PR is split into multiple commits for readability.

Modified the generator to check for the alternate pagination mechanism utilized by sql and compute in the case where AIP-4233 pagination is not detected. This cascaded into refactoring the sql quickstart which uses the List method (which is now paginated).

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11318)

### DIFF
--- a/generator/internal/pagination.cc
+++ b/generator/internal/pagination.cc
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 #include "generator/internal/pagination.h"
+#include "generator/internal/codegen_utils.h"
+#include "generator/internal/descriptor_utils.h"
 #include "google/cloud/log.h"
 #include "google/cloud/optional.h"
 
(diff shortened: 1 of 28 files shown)


google/cloud/sql/v1/sql_operations_connection.h
```diff
@@ -67,8 +68,8 @@ class SqlOperationsServiceConnection {
   virtual StatusOr<google::cloud::sql::v1::Operation> Get(
       google::cloud::sql::v1::SqlOperationsGetRequest const& request);
 
-  virtual StatusOr<google::cloud::sql::v1::OperationsListResponse> List(
-      google::cloud::sql::v1::SqlOperationsListRequest const& request);
+  virtual StreamRange<google::cloud::sql::v1::Operation> List(
+      google::cloud::sql::v1::SqlOperationsListRequest request);
```
coryan, line 72: Aside: the change from `T const&` to `T` is fairly expensive.  I assume we cannot change it due to backwards compatibility reasons.  Do we have a plan to fix this?

generator/internal/pagination_test.cc
```diff
@@ -290,6 +290,287 @@ TEST(PredicateUtilsTest, PaginationExactlyOneRepatedStringResponse) {
   EXPECT_EQ(result->second, nullptr);
 }
 
+TEST(PredicateUtilsTest, PaginationRestSuccess) {
```
coryan, line 293: s/PredicateUtilstest/PaginationTest/ maybe?

generator/internal/pagination.cc
```diff
@@ -84,10 +89,63 @@ DeterminePagination(MethodDescriptor const& method) {
                         std::get<1>(repeated_message_fields[0]));
 }
 
+// For both the sqladmin and compute proto definitions, the paging conventions
+// do not adhere to aip-4233, but the intent is there. If we can make it work,
+// add pagination for any such methods.
+google::cloud::optional<std::pair<std::string, Descriptor const*>>
+DetermineRestPagination(MethodDescriptor const& method) {
```
coryan, line 96: `DetermineRestPagination` seems too optimistic.  Maybe `DetermineAlternativePagination`?  No biggie, as this is in the anonymous namespace anyway.

coryan (approved):
Please take the opportunity to clean up the test names, sorry I missed that in the refactoring PR.

PR #11319, opened 2023-04-18
impl(docfx): iterate on TOC

coryan:
Start the table of contents (TOC) with the `indexpage` and use
`index.md` as the output file for that page. Doxygen uses `indexpage`
for the landing page, the cloud site uses `index.md`.

Move the namespaces up in the TOC, and use the C++ name (instead of the
Doxygen uid) as the name of the TOC entry.

Move the groups down in the TOC, and use the group title as the name of
the TOC entry.

Keep the TOC entries sorted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11319)

### DIFF
--- a/docfx/doxygen_pages.cc
+++ b/docfx/doxygen_pages.cc
@@ -129,9 +129,10 @@ std::vector<TocEntry> PagesToc(pugi::xml_document const& doc) {
     return TocEntry{std::string{id}, title.str(), std::move(filename)};
   });
   std::sort(result.begin(), result.end(), [](auto const& a, auto const& b) {
-    // Make `index.md` (if present) the first entry.
-    if (a.uid == "indexpage") return true;
+    // If there is an `indexpage` element (aka `index.md`) it should be the
+    // first entry.
     if (b.uid == "indexpage") return false;
+    if (a.uid == "indexpage") return true;
     return a.uid < b.uid;
   });
   return result;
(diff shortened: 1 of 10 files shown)


docfx/doxygen_pages.cc
```diff
@@ -122,8 +122,17 @@ std::vector<TocEntry> PagesToc(pugi::xml_document const& doc) {
...
+    std::ostringstream title;
+    AppendTitle(title, MarkdownContext{}, page);
+    auto filename =
+        std::string(id == "indexpage" ? std::string_view{"index"} : id) + ".md";
+    return TocEntry{std::string{id}, title.str(), std::move(filename)};
+  });
+  std::sort(result.begin(), result.end(), [](auto const& a, auto const& b) {
+    // Make `index.md` (if present) the first entry.
+    if (a.uid == "indexpage") return true;
+    if (b.uid == "indexpage") return false;
```
devbww, line 134: We should reverse these checks (i.e., have the `b.uid` one come first) so that the comparator is a true "strict weak ordering", where `cmp({.uid = "indexpage"}, {.uid = "indexpage"})` is false.

coryan: Good catch, fixed.

devbww (approved):

Issue #11264, opened 2023-04-13
Implement OTel Context propagation for HTTP Requests


coryan:
FYI: We still need to do context propagation for HTTP. So in a future PR, we should be setting the span as active, and injecting the trace context as a header.

Basically this class, but with a RestRequest? (Or would it be something with a RestContext?) I am not so sure.

https://github.com/googleapis/google-cloud-cpp/blob/f5f98d27ec1ebb23edc5249e67421fa5fb49159c/google/cloud/internal/grpc_opentelemetry.cc#L36-L66

https://github.com/googleapis/google-cloud-cpp/blob/f5f98d27ec1ebb23edc5249e67421fa5fb49159c/generator/integration_tests/golden/v1/internal/golden_kitchen_sink_tracing_stub.cc#L49

_Originally posted by @dbolduc in https://github.com/googleapis/google-cloud-cpp/pull/11249#discussion_r1164958136_

PR #11332, opened 2023-04-20
impl(otel): set the tracing scope on HTTP requests

coryan:
Part of the work for #11264

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11332)

### DIFF
--- a/google/cloud/internal/tracing_rest_client_test.cc
+++ b/google/cloud/internal/tracing_rest_client_test.cc
@@ -107,6 +107,64 @@ TEST(TracingRestClient, Delete) {
           SpanNamed("Read"), SpanNamed("Read")));
 }
 
+TEST(TracingRestClient, HasScope) {
+  namespace sc = ::opentelemetry::trace::SemanticConventions;
+  auto span_catcher = InstallSpanCatcher();
+
+  auto impl = std::make_unique<MockRestClient>();
+  EXPECT_CALL(*impl, Get).WillOnce([](RestContext&, RestRequest const&) {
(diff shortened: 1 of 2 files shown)


google/cloud/internal/tracing_rest_client_test.cc
```diff
@@ -107,6 +107,64 @@ TEST(TracingRestClient, Delete) {
           SpanNamed("Read"), SpanNamed("Read")));
 }
 
+TEST(TracingRestClient, HasScope) {
```
dbolduc, line 110: So I personally prefer to have more independent tests. As it stands, `Delete ⊂ HasScope`. I would suggest either

A. Consolidating the `Delete` and `HasScope` test into one.

B. Isolating only the things that have to do with the request span being active. e.g. removing fluff like 
```cc
  request.AddHeader("x-test-header-3", "value3");
```

or even validating the response and its contents.

and the final expectation would just look like:
```cc
  EXPECT_THAT(spans, UnorderedElementsAre(
                         // Request span
                         AllOf(SpanNamed("HTTP/GET"),
                               SpanHasAttributes(SpanAttribute<std::string>(
                                   "test.attribute", "test.value"))),
                         // Read span on the HttpPayload
                         SpanNamed("Read"), SpanNamed("Read")));
```

dbolduc (approved):

google/cloud/internal/tracing_rest_client_test.cc
coryan: Done.

PR #11336, opened 2023-04-20
impl(bigquery): Added DebugString() member functions for response logging

jsrinnn:
This PR adds DebugString() member functions to Dataset response and dataset objects so they can use used  in the dataset logging decorator which would be sent in a future PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336)


google/cloud/bigquery/v2/minimal/internal/dataset.cc
```diff
@@ -174,6 +175,172 @@ void from_json(nlohmann::json const& j, Dataset& d) {
...
+      .StringField("id", id)
+      .StringField("self_link", self_link)
+      .StringField("friendly_name", friendly_name)
+      .StringField("description", description)
+      .StringField("type", type)
+      .StringField("location", location)
+      .StringField("default_collation", default_collation)
+      .Field("published", published)
+      .Field("is_case_insensitive", is_case_insensitive)
+      .Field("default_table_expiration", default_table_expiration.count())
```
devbww, line 313: It looks like we need some formatter support for `std::chrono::duration`.  I'll do something for that, and then loop back here.

google/cloud/bigquery/v2/minimal/internal/dataset.h
```diff
@@ -167,7 +221,7 @@ struct Dataset {
   ;
   std::chrono::hours max_time_travel = std::chrono::hours(0);
 
-  std::map<std::string, std::string> labels;
+  std::multimap<std::string, std::string> labels;
```
devbww, line 224: If this isn't a change you would prefer to make, then I'm guessing it was because `DebugFormatter` doesn't support `std::map`.  We can address that if you'd like.  Let me know.

google/cloud/bigquery/v2/minimal/internal/dataset_response_test.cc
```diff
@@ -166,6 +167,293 @@ TEST(ListDatasetsResponseTest, InvalidListFormatDataset) {
...
+            R"( self_link: "d-selfLink")"
+            R"( friendly_name: "d-friendly-name")"
+            R"( description: "")"
+            R"( type: "")"
+            R"( location: "")"
+            R"( default_collation: "")"
+            R"( published: false)"
+            R"( is_case_insensitive: false)"
+            R"( default_table_expiration: 0 )"
+            R"(default_partition_expiration: 0)"
```
devbww, line 200: It would be nice to continue the convention of placing the separator at the beginning of this line rather than the end of the previous one (here and in a handful of following cases).

google/cloud/bigquery/v2/minimal/internal/dataset_test.cc
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      https://www.apache.org/licenses/LICENSE-2.0
+//      https://www.apache.org/licenses/LICENSE2.0
```
devbww, line 7: It looks like this should be reverted.

devbww (approved):

jsrinnn (commented):
Addressed all feedback. I had a question regarding the last one. Can you PTAL? 

Thanks!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336#-:-NTVdTkR3b9yp8jayxRT:bkj7fiu)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.h` line 224 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336#gh-1173098840:-NTVZUt02sQ9NS8jGdy9:b-qlb7bz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d964c98a79d39bb7ddf5d21ccf7799268d1fda9e/google/cloud/bigquery/v2/minimal/internal/dataset.h#L224)):*

If this isn't a change you would prefer to make, then I'm guessing it was because `DebugFormatter` doesn't support `std::map`.  We can address that if you'd like.  Let me know.
</blockquote></details>

Yeah...I mainly made the change for compatibility with DebugFormatter and at this point there is no strong dependency for the response object on std::map. But if DebugFromatter can support std::map, that would be ideal.

Thanks!
___
*[`google/cloud/bigquery/v2/minimal/internal/dataset.cc` line 313 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336#gh-1173096384:-NTVZr80BOR-mZGG1hkH:b-4104q2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d964c98a79d39bb7ddf5d21ccf7799268d1fda9e/google/cloud/bigquery/v2/minimal/internal/dataset.cc#L313)):*

It looks like we need some formatter support for `std::chrono::duration`.  I'll do something for that, and then loop back here.
</blockquote></details>

Awesome. Thank you, Bradley! Much appreciated.
___
*[`google/cloud/bigquery/v2/minimal/internal/dataset_response_test.cc` line 200 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336#gh-1173105876:-NTV_nqMDrMpGv59qKBV:b-418lx4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d964c98a79d39bb7ddf5d21ccf7799268d1fda9e/google/cloud/bigquery/v2/minimal/internal/dataset_response_test.cc#L200)):*

It would be nice to continue the convention of placing the separator at the beginning of this line rather than the end of the previous one (here and in a handful of following cases).
</blockquote></details>

Done.

Ah...sorry that unintentional. Fixed it here and elsehwere.
___
*[`google/cloud/bigquery/v2/minimal/internal/dataset_test.cc` line 7 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336#gh-1173106362:-NTV_mFrEo1RUq_lq7hp:b-3fx327) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d964c98a79d39bb7ddf5d21ccf7799268d1fda9e/google/cloud/bigquery/v2/minimal/internal/dataset_test.cc#L7)):*

It looks like this should be reverted.
</blockquote></details>

Not sure I follow. Why should this file be reverted? If not the file can you be more specific what needs to be reverted here? Thanks.

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336)* status: 0 of 6 files reviewed, 3 unresolved discussions

___
*[`google/cloud/bigquery/v2/minimal/internal/dataset_test.cc` line 7 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336#gh-1173106362:-NTVerujEeBkpwnne66z:bmmxqrr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d964c98a79d39bb7ddf5d21ccf7799268d1fda9e/google/cloud/bigquery/v2/minimal/internal/dataset_test.cc#L7)):*

Not sure I follow. Why should this file be reverted? If not the file can you be more specific what needs to be reverted here? Thanks.
</blockquote></details>

Sorry ... not the file ... just the edit on this line that removed the `-` from the URL.

jsrinnn:
Bradley, I'll go ahead and merge this PR when checks are done. If there are additional changes that needs to be done , please comment on this PR and I can follow that up when I do the logging decorator changes. Thanks!

jsrinnn (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/dataset_test.cc` line 7 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11336#gh-1173106362:-NTVgeKn9zcyz624I30c:bsdqc4r) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d964c98a79d39bb7ddf5d21ccf7799268d1fda9e/google/cloud/bigquery/v2/minimal/internal/dataset_test.cc#L7)):*

Sorry ... not the file ... just the edit on this line that removed the `-` from the URL.
</blockquote></details>

Ah...got it. Looks like Carlos already pushed a fix. Thank you, Carlos!

PR #11344, opened 2023-04-21
fix(spanner): propagate Options through SessionPool callbacks

devbww:
`BatchCreateSessions()` calls made during the `SessionPool` callback chain depend on the prevailing options to determine whether the RPC should be routed to a region leader.  So, we need to ensure that the construction-time options are propagated through those callbacks by instantiating `OptionsSpan` objects when appropriate.

Add similar `OptionsSpan` objects to `session_pool_integration_test.cc`, which uses "FriendForTest" access to reach into `SessionPool`, bypassing the normal span instantiation.

Add unit tests to verify that the "x-goog-spanner-route-to-leader" header is now set appropriately when the `RouteToLeaderOption` is both enabled and disabled.

Add `RouteToLeaderOption` to the `SessionPoolOptionList`, marking it as a valid option to pass to `spanner::MakeConnection()`.

Also add a slight optimization to `SessionPool::Grow()`, which skips even trying to create sessions when `ComputeCreateCounts()` returns an empty vector.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11344)

### DIFF
--- a/google/cloud/spanner/options.h
+++ b/google/cloud/spanner/options.h
@@ -91,9 +91,14 @@ using SpannerPolicyOptionList =
                SpannerPollingPolicyOption>;
 
 /**
- * Option for `google::cloud::Options` to, when present and false, suppress
- * adding headers to distinguish requests served by the leader v/s non-leader
- * region.
+ * Control "route to leader region" headers.
+ *
+ * Unless this option is present and `false` the client library will send
+ * headers that route the request to the leader region.
+ *
+ * @see https://cloud.google.com/spanner/docs/instance-configurations
+ * for more information on multi-regional spanner instances and the role of
+ * leader regions.
  *
  * @ingroup spanner-options
  */
--- a/google/cloud/spanner/options.h
+++ b/google/cloud/spanner/options.h
@@ -91,6 +91,17 @@ using SpannerPolicyOptionList =
                SpannerPollingPolicyOption>;
 
 /**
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/options.h
```diff
@@ -90,6 +90,17 @@ using SpannerPolicyOptionList =
     OptionList<spanner::SpannerRetryPolicyOption, SpannerBackoffPolicyOption,
                SpannerPollingPolicyOption>;
 
+/**
+ * Option for `google::cloud::Options` to, when present and false, suppress
+ * adding headers to distinguish requests served by the leader v/s non-leader
+ * region.
+ *
+ * @ingroup spanner-options
+ */
```
coryan, line 99: This assumes the reader understands leader regions already, and it is not a good "brief" comment.

```suggestion
/**
 * Control route-to-leader-region headers.
 *
 * Unless this option is present and `false` the client library will send 
 * headers that route the request to the leader region.
 *
 * @see https://cloud.google.com/spanner/docs/instance-configurations
 * for more information on multi-regional spanner instances and the role of
 * leader regions.
 * 
 * @ingroup spanner-options
 */
```

coryan (approved):

google/cloud/spanner/options.h
devbww: Excellent.  Done.  Thanks.

PR #11347, opened 2023-04-21
impl(otel): generate children and references

coryan:
List all the children of each namespace, class, struct, and enum. That makes it possible to display them in the generated HTML. Also generate the `references` section in each `.yml` file. That is where cross-references are discovered.


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11347)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

PR #11348, opened 2023-04-21
impl(bigquery): Metadata decorator and tests for Dataset apis

jsrinnn:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11348)


jsrinnn:
friendly ping. 
If some one can review this when they get a chance, that would be awesome. Thanks much in advance!

google/cloud/bigquery/v2/minimal/internal/dataset_metadata.cc
```diff
@@ -0,0 +1,74 @@
...
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+DatasetMetadata::DatasetMetadata(std::shared_ptr<DatasetRestStub> child)
+    : child_(std::move(child)),
+      api_client_header_(
+          google::cloud::internal::ApiClientHeader("bigquery_v2_dataset")) {}
+
+StatusOr<GetDatasetResponse> DatasetMetadata::GetDataset(
+    rest_internal::RestContext& context, GetDatasetRequest const& request) {
+  SetMetadata(context);
```
coryan, line 37: I am still surprised there is no specific set of request parameters associated with these requests, but we can always fix that later and let you make progress.

coryan (approved):

jsrinnn (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11348)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/bigquery/v2/minimal/internal/dataset_metadata.cc` line 37 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11348#gh-1175378153:-NTnxIOv8EsLa7Oo5xpc:b-m6v55h) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/7070809bedf7d70218c4d24fc32f41996eaef57b/google/cloud/bigquery/v2/minimal/internal/dataset_metadata.cc#L37)):*

I am still surprised there is no specific set of request parameters associated with these requests, but we can always fix that later and let you make progress.
</blockquote></details>

I had added some for job but Darren had pointed out that none was needed based on the proto, hence I had removed it.

Yeah...we can always add it later as needed.

PR #11350, opened 2023-04-21
impl(bigquery): refactored JobLogWrapper template function

jsrinnn:
Refactored JobLogWrapper template function to a separate header so it can be used from multiple places e.g. dataset logging decorator

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11350)


coryan (approved):
Much easier to grok.  Thanks!

PR #11351, opened 2023-04-21
impl(GCS+gRPC): larger `UploadBufferSizeOption`

coryan:
GCS+gRPC performs much better if the upload buffers are at least 32MiB.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11351)

### DIFF
--- a/google/cloud/storage/internal/grpc_client.cc
+++ b/google/cloud/storage/internal/grpc_client.cc
@@ -206,7 +206,12 @@ int DefaultGrpcNumChannels(std::string const& endpoint) {
 
 Options DefaultOptionsGrpc(Options options) {
   using ::google::cloud::internal::GetEnv;
-
+  // Experiments show that gRPC gets better upload throughput when the upload
+  // buffer is larger than 32MiB.
+  auto constexpr kDefaultGrpcUploadBufferSize = 32 * 1024 * 1024L;
+  options = google::cloud::internal::MergeOptions(
+      std::move(options), Options{}.set<storage::UploadBufferSizeOption>(
+                              kDefaultGrpcUploadBufferSize));
   options =
       storage::internal::DefaultOptionsWithCredentials(std::move(options));
   if (!options.has<UnifiedCredentialsOption>() &&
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/grpc_client_test.cc
```diff
@@ -121,6 +122,12 @@ TEST_F(GrpcClientTest, DefaultOptionsGrpcEndpointNoEnv) {
   }
 }
 
+TEST_F(GrpcClientTest, DefaultOptionsUploadBuffer) {
```
dbolduc, line 125: optional: `TEST(DefaultOptionsGrpc, UploadBuffer)`

and above with `EndpointNoEnv`, `ChannelCount`

dbolduc (approved):

google/cloud/storage/internal/grpc_client_test.cc
```diff
@@ -121,6 +122,12 @@ TEST_F(GrpcClientTest, DefaultOptionsGrpcEndpointNoEnv) {
...
+  auto const actual =
+      DefaultOptionsGrpc(Options{}).get<storage::UploadBufferSizeOption>();
+  EXPECT_GE(actual, 32 * 1024 * 1024L);
```
dbolduc, line 128: you might also want to check that a user-supplied `UploadBufferSizeOption` is used instead of the default value.

google/cloud/storage/internal/grpc_client.cc
```diff
@@ -206,7 +206,12 @@ int DefaultGrpcNumChannels(std::string const& endpoint) {
 
 Options DefaultOptionsGrpc(Options options) {
   using ::google::cloud::internal::GetEnv;
-
+  // Experiments show that gRPC gets better upload throughput when the upload
+  // buffer is larger than 32MiB.
+  auto constexpr kDefaultGrpcUploadBufferSize = 32 * 1024 * 1024L;
```
devbww, line 211: Nit: I presume you meant "at least 32MiB", otherwise we're setting the default to something we know could be better?

PR #11352, opened 2023-04-21
impl(otel): `TracingRestClient` tracing propagation

### LINKED ISSUE #11264: Implement OTel Context propagation for HTTP Requests

coryan:
With this change the `TracingRestClient` always includes the `x-cloud-trace-context` header and any other configured tracing propagators.

Fixes #11264

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11352)

### DIFF
--- a/google/cloud/internal/tracing_rest_client.cc
+++ b/google/cloud/internal/tracing_rest_client.cc
@@ -31,9 +31,9 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 namespace {
 
-// We always inject the CloudTraceContext, these requests are going to GCP where
-// the X-Cloud-Trace-Context header can connect the request with the internal
-// tracing service.
+// We always inject the CloudTraceContext. Typically, these requests are going
+// to GCP where the X-Cloud-Trace-Context header can connect the request with
+// the internal tracing service.
 void InjectCloudTraceContext(RestContext& ctx,
                              opentelemetry::trace::Span const& span) {
   auto context = span.GetContext();
--- a/google/cloud/internal/tracing_rest_client.cc
+++ b/google/cloud/internal/tracing_rest_client.cc
@@ -13,10 +13,14 @@
 // limitations under the License.
 
 #include "google/cloud/internal/tracing_rest_client.h"
+#include "google/cloud/internal/absl_str_cat_quiet.h"
 #include "google/cloud/internal/opentelemetry.h"
 #include "google/cloud/internal/rest_opentelemetry.h"
 #include "google/cloud/internal/tracing_http_payload.h"
 #include "google/cloud/internal/tracing_rest_response.h"
+#include "absl/strings/numbers.h"
+#include <array>
+#include <cstdint>
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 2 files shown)


google/cloud/internal/tracing_rest_client.cc
```diff
@@ -25,6 +29,35 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 #ifdef GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
 
+namespace {
+
+// We always inject the CloudTraceContext, these requests are going to GCP where
+// the X-Cloud-Trace-Context header can connect the request with the internal
+// tracing service.
```
dbolduc, line 36: s/, these/. These/

---

I think this is like 99% true? I am thinking of WIF. Don't we use the `RestClient` to send requests to AWS, for example, in that case?

We can deal with that later, but in the long run, I suspect we want to have a propagator determined by Options. (If the Options know whether we are also communicating with AWS, Ideally we could use their X-Ray context too).

With that being said, adding this header in a request to some other cloud is fine. It will just get ignored.

scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11352)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

google/cloud/internal/tracing_rest_client.cc
```diff
@@ -25,6 +29,35 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
...
+  if (!context.IsValid()) return;
+
+  using opentelemetry::trace::SpanId;
+  using opentelemetry::trace::TraceId;
+  std::array<char, 2 * TraceId::kSize> trace_id;
+  context.trace_id().ToLowerBase16(trace_id);
+  std::array<char, 2 * SpanId::kSize> span_id;
+  context.span_id().ToLowerBase16(span_id);
+  std::uint64_t value;
+  if (!absl::SimpleHexAtoi(absl::string_view{span_id.data(), span_id.size()},
```
dbolduc, line 49: TIL: `absl::SimpleHexAtoi`

google/cloud/internal/tracing_rest_client.cc
```diff
@@ -59,21 +92,27 @@ StatusOr<std::unique_ptr<RestResponse>> TracingRestClient::Delete(
...
+  InjectTraceContext(context, internal::CurrentOptions());
+  InjectCloudTraceContext(context, *span);
   return EndResponseSpan(std::move(span), impl_->Get(context, request));
 }
 
 StatusOr<std::unique_ptr<RestResponse>> TracingRestClient::Patch(
     RestContext& context, RestRequest const& request,
     std::vector<absl::Span<char const>> const& payload) {
-  auto span = MakeSpanHttp(request, "POST");
+  auto span = MakeSpanHttp(request, "PATCH");
```
dbolduc, line 112: :)

google/cloud/internal/tracing_rest_client_test.cc
```diff
@@ -150,6 +155,107 @@ TEST(TracingRestClient, HasScope) {
...
+    return ResultOf(
+        "headers have x-cloud-trace-context",
+        [](RestContext const& c) { return c.headers(); },
+        Contains(Pair("x-cloud-trace-context", _)));
+  };
+  EXPECT_CALL(*impl, Patch(expected_headers_matcher(), _, _))
+      .WillOnce([](auto&, auto const&, auto const&) {
+        // Inject an attribute to the current span, which should be the request
+        // span.
+        auto span = opentelemetry::trace::Tracer::GetCurrentSpan();
```
dbolduc, line 172: Remove, unused.

google/cloud/internal/tracing_rest_client_test.cc
```diff
@@ -150,6 +155,107 @@ TEST(TracingRestClient, HasScope) {
...
+        AllOf(Contains(Pair("x-cloud-trace-context", _)),
+              Contains(Pair("traceparent", _))));
+  };
+  auto disambiguate_overload =
+      ::testing::An<std::vector<absl::Span<char const>> const&>();
+  EXPECT_CALL(*impl, Post(expected_headers_matcher(), _, disambiguate_overload))
+      .WillOnce([](auto&, auto const&, auto const&) {
+        // Inject an attribute to the current span, which should be the request
+        // span.
+        auto span = opentelemetry::trace::Tracer::GetCurrentSpan();
```
dbolduc, line 228: ditto

dbolduc (commented):
This is progress, but I suspect we would rather use OpenTelemetry's propagator API.

We would define the cloud trace context propagator, and ensure it is used by `InjectTraceContext`.

Maybe just open an issue like `Consider using OpenTelemetry's Propagator API for "X-Cloud-Trace-Context" injection`.

google/cloud/internal/tracing_rest_client_test.cc
coryan: Done

google/cloud/internal/tracing_rest_client_test.cc
coryan: Done

google/cloud/internal/tracing_rest_client.cc
coryan: Yeah...

google/cloud/internal/tracing_rest_client.cc
coryan: You are right, changed the comment and I opened the bug you wanted.

Issue #11353, opened 2023-04-21
Make generator package aware when emitting protos.


scotthart:
Depending on the package defined in a proto file, some imported types may need to have their names package qualified.

This will involve changes to
- discovery_type_vertex
- discovery_resource
- discovery_to_proto
- discovery_file

PR #11357, opened 2023-04-21
impl(generator): add package name support for DiscoveryResource

scotthart:
part of the work for #11353

### DIFF
--- a/generator/internal/discovery_resource.cc
+++ b/generator/internal/discovery_resource.cc
@@ -29,9 +29,11 @@ namespace generator_internal {
 
 DiscoveryResource::DiscoveryResource() : json_("") {}
 
-DiscoveryResource::DiscoveryResource(std::string name, std::string default_host,
+DiscoveryResource::DiscoveryResource(std::string name, std::string package_name,
+                                     std::string default_host,
                                      std::string base_path, nlohmann::json json)
     : name_(std::move(name)),
+      package_name_(std::move(package_name)),
       default_host_(std::move(default_host)),
       base_path_(std::move(base_path)),
       json_(std::move(json)) {}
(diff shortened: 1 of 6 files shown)


generator/internal/discovery_resource.cc
```diff
@@ -29,9 +29,11 @@ namespace generator_internal {
 
 DiscoveryResource::DiscoveryResource() : json_("") {}
 
-DiscoveryResource::DiscoveryResource(std::string name, std::string default_host,
+DiscoveryResource::DiscoveryResource(std::string name, std::string package_name,
```
coryan, line 32: This many string parameters makes it really easy to swap them. Consider a struct or a small class.

coryan (approved):

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11357)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`generator/internal/discovery_resource.cc` line 32 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11357#gh-1175902527:-NTt74x7A3fPjkKQ5K3z:b-wjljq3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/feeca8adb794d9810b505ae943d324f4c3d243c0/generator/internal/discovery_resource.cc#L32)):*

This many string parameters makes it really easy to swap them. Consider a struct or a small class.
</blockquote></details>

Added an issue to fix this. My plan is to remove default_host and base_path as member variables and instead pass them (as part of a struct) to where they are used, `DiscoveryResource::JsonToProtobufService`.

PR #11359, opened 2023-04-22
impl(otel): a simplified tracing initializer

coryan:
For many applications we can provide a one-liner that initializes OpenTelemetry. Some applications will need more elaborate initialization, but making things easy for the common case is valuable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11359)

### DIFF
--- a/ci/cloudbuild/builds/lib/integration.sh
+++ b/ci/cloudbuild/builds/lib/integration.sh
@@ -163,6 +163,8 @@ function integration::bazel_with_emulators() {
     "google/cloud/pubsublite/..."
     # Unified Rest Credentials test
     "google/cloud:internal_unified_rest_credentials_integration_test"
+    # The OpenTelemetry integration tests are limited by quota
+    "google/cloud/opentelemetry/integration_tests/..."
   )
 
   production_tests_tag_filters="integration-test"
--- a/google/cloud/opentelemetry/configure_basic_tracing.cc
+++ b/google/cloud/opentelemetry/configure_basic_tracing.cc
@@ -29,21 +29,23 @@ class BasicTracingConfigurationImpl : public BasicTracingConfiguration {
  public:
   explicit BasicTracingConfigurationImpl(
       std::shared_ptr<opentelemetry::sdk::trace::TracerProvider> provider)
-      : provider_(std::move(provider)) {
+      : provider_(std::move(provider)),
+        previous_(opentelemetry::trace::Provider::GetTracerProvider()) {
     opentelemetry::trace::Provider::SetTracerProvider(
         opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider>(
             provider_));
   }
+
   ~BasicTracingConfigurationImpl() override {
     if (provider_) provider_->ForceFlush(std::chrono::microseconds(1000));
-    // Clear the global tracer provider.
-    opentelemetry::trace::Provider::SetTracerProvider(
-        opentelemetry::nostd::shared_ptr<
-            opentelemetry::trace::TracerProvider>());
+    // Reset the global tracer provider.
+    opentelemetry::trace::Provider::SetTracerProvider(std::move(previous_));
   }
 
  private:
   std::shared_ptr<opentelemetry::sdk::trace::TracerProvider> provider_;
+  opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider>
+      previous_;
 };
 
 }  // namespace
--- a/google/cloud/opentelemetry/configure_basic_tracing.cc
+++ b/google/cloud/opentelemetry/configure_basic_tracing.cc
@@ -1,0 +1,73 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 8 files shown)


google/cloud/opentelemetry/configure_basic_tracing.cc
```diff
@@ -0,0 +1,75 @@
...
+      previous_;
+};
+
+}  // namespace
+
+std::unique_ptr<BasicTracingConfiguration> ConfigureBasicTracing(
+    Project project, Options options) {
+  // Just return a nullptr if the project is not configured. This is intended
+  // as a function to make things easy, no reason to return complicated errors.
+  if (project.project_id().empty()) return {};
```
dbolduc, line 57: I would remove this check. I do not think it makes it easier for customers to debug their configuration when they accidentally provide an empty project ID. And we don't want them to start relying on the `if (project.project_id().empty())` behavior.

If we care enough, we could `GCP_LOG(INFO) << "Empty project ID provided to tracing configuration.";`

google/cloud/opentelemetry/configure_basic_tracing.h
```diff
@@ -0,0 +1,161 @@
...
+ *     enabled by default in `google-cloud-cpp`. Please consult the build
+ *     documentation to enable the additional libraries.
+ *
+ * @par Usage Example
+ * @parblock
+ * Change your build scripts to also the library that provides this function.
+ *
+ * Change your application to call this function once, for example in `main()`
+ * as follows:
+ *
```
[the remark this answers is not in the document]
coryan: > So basically, we _might_ need one more line of instruction. e.g. `Compile your application with the environment variable: 

s/Compile/Run/ ?

google/cloud/opentelemetry/configure_basic_tracing.cc
coryan: I think the check is a feature.  It let's me disable things easily.  OTOH, I don't care that much.  If we remove it, we need to return something sensible, either a `StatusOr<std::unique_ptr<...>>` or we add a `enabled()` accessor to the `BasicTracingConfiguration` class.  Your call.

google/cloud/opentelemetry/configure_basic_tracing.cc
```diff
@@ -0,0 +1,75 @@
...
+        previous_(opentelemetry::trace::Provider::GetTracerProvider()) {
+    opentelemetry::trace::Provider::SetTracerProvider(
+        opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider>(
+            provider_));
+  }
+
+  ~BasicTracingConfigurationImpl() override {
+    if (provider_) provider_->ForceFlush(std::chrono::microseconds(1000));
+    // Reset the global tracer provider.
+    opentelemetry::trace::Provider::SetTracerProvider(std::move(previous_));
```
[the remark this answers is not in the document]
coryan: I just added an integration test.  We really want to test that this thing publishes spans to Cloud Trace.

google/cloud/opentelemetry/configure_basic_tracing.h
```diff
@@ -0,0 +1,161 @@
...
+ * this function if your application needs fine control over OpenTelemetry
+ * settings.
+ *
+ * @note If you are using CMake as your build system, OpenTelemetry is not
+ *     enabled by default in `google-cloud-cpp`. Please consult the build
+ *     documentation to enable the additional libraries.
+ *
+ * @par Usage Example
+ * @parblock
+ * Change your build scripts to also the library that provides this function.
```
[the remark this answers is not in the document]
coryan: Done. This is why we need a quickstart, to link it from here.

google/cloud/opentelemetry/configure_basic_tracing.h
```diff
@@ -0,0 +1,161 @@
...
+  virtual ~BasicTracingConfiguration() = default;
+};
+
+// TODO(#11360) - reference the quickstart when discussing program linking.
+/**
+ * Configure the application for basic request tracing.
+ *
+ * This function configures basic request tracing to [Cloud Trace]. The
+ * `google-cloud-cpp` libraries use [OpenTelemetry] to provide telemetry into
+ * their operation at runtime.
```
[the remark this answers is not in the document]
coryan: Done

google/cloud/opentelemetry/configure_basic_tracing.cc
dbolduc: > I think the check is a feature

My intuition was that the check belongs in the application.

I think your intuition is that most applications will want to have the same code for tracing / non-tracing. And they would switch between the two with an env var. And it is more convenient if we write the `if` statement than them. Something like:

```cc
auto project_id = getenv("MY_PROJECT_ID");
ConfigureBasicTracing(gc::Project(project_id));
```

vs.

```cc
auto project_id = getenv("MY_PROJECT_ID");
if (project_id) {
  ConfigureBasicTracing(gc::Project(project_id));
}
```

I am happy to leave the code as is and go with your intuition.

> If we remove it, we need to return something sensible

(For the sake of arguing), a configuration with a useless exporter is a sensible return value.

GCP project IDs must be between 4 and 30 characters. But we do not also check that `project.project_id().size() < 4 || project.project_id().size() > 30`.

ci/cloudbuild/builds/lib/integration.sh
```diff
@@ -163,6 +163,8 @@ function integration::bazel_with_emulators() {
     "google/cloud/pubsublite/..."
     # Unified Rest Credentials test
     "google/cloud:internal_unified_rest_credentials_integration_test"
+    # The OpenTelemetry integration tests are limited by quota
+    "google/cloud/opentelemetry/integration_tests/..."
```
dbolduc, line 167: I think this is the opposite of what you want. Note that they are already in this list on L161.

I think we want to remove them from this list so they don't run.

dbolduc (approved):

ci/cloudbuild/builds/lib/integration.sh
dbolduc: We will address in a follow up.

PR #11361, opened 2023-04-23
test(otel): add Trace Exporter integration test

dbolduc:
Part of the work for #11156 

Add a simple integration test for the Cloud Trace exporter. We can build off this to test that an instrumented client with tracing enabled creates and sends spans.

FYI: I still have one annoying name thing to resolve for bazel builds. I still don't have a good answer. Maybe:
- `//:experimental-opentelemetry`: the `cc_library` for the exporters and such that live in `google/cloud/opentelemetry/...`
- `//:experimental-opentelemetry_instrumentation`: the `bool_flag` that builds common w/ OpenTelemetry

We cannot consolidate the targets like we did with cmake.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11361)


google/cloud/opentelemetry/integration_tests/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
+# ~~~
```
dbolduc, line 1: Need to `add_subdirectory()`

google/cloud/opentelemetry/trace_exporter.h
```diff
@@ -29,7 +29,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 
 // Make an OpenTelemetry Trace Exporter for Google Cloud Trace.
```
coryan, line 30: For a future PR: fix the documentation.  This is supposed to be a public function.

coryan (approved):

google/cloud/opentelemetry/trace_exporter.h
dbolduc: Ack.

PR #11365, opened 2023-04-24
impl(otel): add span while HTTP sends a request

coryan:
Sending a HTTP request can take a while and create confusing spans in the overall request. I also think using events to represent the connection setup steps (name lookup, tcp connection, and SSL handshakes) is a cleaner representation than custom attributes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11365)

### DIFF
--- a/google/cloud/internal/tracing_rest_client.cc
+++ b/google/cloud/internal/tracing_rest_client.cc
@@ -19,6 +19,7 @@
 #include "google/cloud/internal/rest_opentelemetry.h"
 #include "google/cloud/internal/tracing_http_payload.h"
 #include "google/cloud/internal/tracing_rest_response.h"
+#include "absl/functional/function_ref.h"
 #include "absl/strings/match.h"
 #include "absl/strings/numbers.h"
 #include <opentelemetry/trace/semantic_conventions.h>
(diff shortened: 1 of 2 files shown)


google/cloud/internal/tracing_rest_client.cc
```diff
@@ -57,151 +58,160 @@ void InjectCloudTraceContext(RestContext& ctx,
...
+    std::chrono::system_clock::time_point start, Options const& options) {
+  opentelemetry::trace::StartSpanOptions span_options;
+  span_options.kind = opentelemetry::trace::SpanKind::kClient;
+  span_options.start_system_time = start;
+  return internal::GetTracer(options)->StartSpan("SendRequest", span_options);
+}
+
+StatusOr<std::unique_ptr<RestResponse>> EndStartSpan(
+    opentelemetry::trace::Span& span,
+    std::chrono::system_clock::time_point start, RestContext& context,
```
dbolduc, line 126: Drive by: We want to do the time math using `absl::Time` to avoid rounding.

As @devbww noticed, `absl::Time` and `opentelemetry::common::SystemTimestamp` have nanosecond resolution, but `std::chrono::system_clock` may not have nanosecond resolution on all platforms.

google/cloud/internal/tracing_rest_client_test.cc
```diff
@@ -191,9 +193,9 @@ TEST(TracingRestClient, HasCloudTraceContext) {
   EXPECT_THAT(contents, IsOkAndHolds(MockContents()));
 
   auto spans = span_catcher->GetSpans();
-  EXPECT_THAT(spans,
-              UnorderedElementsAre(SpanNamed("HTTP/PATCH"), SpanNamed("Read"),
-                                   SpanNamed("Read")));
+  EXPECT_THAT(
+      spans, UnorderedElementsAre(SpanNamed("HTTP/PATCH"), SpanNamed("Read"),
+                                  SpanNamed("SendRequest"), SpanNamed("Read")));
```
dbolduc, line 198: nit: I know the expectation is unordered, but I would put the `SendRequest` before the `Read`s.

dbolduc (commented):
(Not a full review. I can look more closely later.)

google/cloud/internal/tracing_rest_client_test.cc
coryan: Fixed

google/cloud/internal/tracing_rest_client.cc
coryan: I don't think it can possibly make a difference:

- I am adding microseconds, in all platforms we support `std::chrono::system_clock` has higher resolution than one microsecond.
- `opentelemetry::common::SystemTimestamp` is initialized from a `std::chrono::system_clock::now()`.  Whatever rounding happens, it happens regardless of whether I do the math or the opentelemetry library does the math.

Unless I hear of an specific case in a real computer where this makes a difference, the code as-is is cleaner than going via `absl::Time`.

google/cloud/internal/tracing_rest_client_test.cc
```diff
@@ -298,13 +300,76 @@ TEST(TracingRestClient, WithRestContextDetails) {
...
+          AllOf(SpanNamed("HTTP/PUT"),
+                SpanHasAttributes(
+                    SpanAttribute<std::string>(sc::kNetTransport,
+                                               sc::NetTransportValues::kIpTcp),
+                    SpanAttribute<std::string>(sc::kHttpMethod, "PUT"),
+                    SpanAttribute<std::string>(sc::kHttpUrl, kUrl),
+                    SpanAttribute<std::string>(sc::kNetPeerName, "192.168.1.1"),
+                    SpanAttribute<std::int32_t>(sc::kNetPeerPort, 443),
+                    SpanAttribute<std::string>(sc::kNetHostName, "127.0.0.1"),
+                    SpanAttribute<std::int32_t>(sc::kNetHostPort, 32000))),
```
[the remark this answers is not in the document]
coryan: Done.

google/cloud/internal/tracing_rest_client_test.cc
```diff
@@ -298,13 +300,76 @@ TEST(TracingRestClient, WithRestContextDetails) {
...
+  auto impl = std::make_unique<MockRestClient>();
+  EXPECT_CALL(*impl, Put)
+      .WillOnce([](RestContext& context, auto const&, auto const&) {
+        context.set_namelookup_time(std::chrono::microseconds(50));
+        context.set_connect_time(std::chrono::microseconds(0));
+        context.set_appconnect_time(std::chrono::microseconds(0));
+        context.set_local_ip_address("127.0.0.1");
+        context.set_local_port(32000);
+        context.set_primary_ip_address("192.168.1.1");
+        context.set_primary_port(443);
```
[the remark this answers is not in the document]
coryan: Done, though that assumes more about the implementation than it should.  I was feeding likely values for a cached connections.  The values I am feeding now will never happen in production.

PR #11375, opened 2023-04-25
impl(docfx): handle mocked functions

### LINKED ISSUE #11233: Improve docfx handling of duplicate member functions in Mocks
Doxygen generates duplicate `<memberdef>` for member functions in mock classes. This may be a bug in Doxygen, or just a side-effect of the `MOCK_METHOD()` macros. In any case, the duplicate definitions cause problems for the DocFX YAML tool. I am about to send a PR that removes these duplicates by skipping recursion in mock classes (based on the name, gulp). We may want a cleaner solution.

coryan:
In mock classes Doxygen generates an entry for the `MOCK_METHOD()` invocation and for the derived function being mocked. We want to collapse both entries in the class definition, and only on entry in the `children` element and `references` section.

Fixes #11233


docfx/yaml_context.cc
```diff
@@ -0,0 +1,93 @@
...
+#include "docfx/public_docs.h"
+#include <iostream>
+#include <vector>
+
+namespace docfx {
+namespace {
+
+// We need to build a little data structure to deal with mocks. The member
+// functions of a mock appear twice. First Doxygen thinks there is a
+// MOCK_METHOD()
```
devbww, line 25: ", then ..."?

docfx/yaml_context.h
```diff
@@ -16,16 +16,29 @@
...
 
 namespace docfx {
 
 struct YamlContext {
   std::string library_root = "google/cloud/";
   std::string parent_id;
   Config config;
+  // Mocked function names, indexed by function name.
+  std::unordered_map<std::string, std::string> mocked_functions;
+  // Mocked functions, indexed by id.
```
devbww, line 32: Is this comment an un-updated cut-n-paste?

docfx/yaml_context.cc
coryan: Fixed.

docfx/yaml_context.h
```diff
@@ -27,12 +27,12 @@ struct YamlContext {
...
   std::string parent_id;
   Config config;
-  // Mocked function names, indexed by function name.
-  std::unordered_map<std::string, std::string> mocked_functions;
-  // Mocked functions, indexed by id.
+  // Mocking functions (the MOCK_METHOD() elements), indexed by function name.
+  std::unordered_map<std::string, std::string> mocking_functions;
+  // Mocking functions (the MOCK_METHOD() elements), indexed by their id.
+  std::unordered_map<std::string, std::string> mocking_functions_by_id;
+  // Mocked functions, indexed by id of the inherited function being mocked.
```
devbww, line 34: Nit: Given that this is just a set, I wouldn't think of it as "something indexed by something".

coryan: Fixed

PR #11387, opened 2023-04-26
doc: regenerate new environment-variables.dox files

devbww:
See #11381 for the new `environment-variables.dox` files, but
now restore the blank line from the scaffolding I erroneously
suggested we should remove.

The regeneration will also:
- add the `google:cloud:EndpointOption` reference to the
  "\<service>-env-endpoint" section, and
- correct the name of the "\<service>-env-logging" section.

And while we're there, use the `mdformat`-preferred syntax for
top-level unordered lists in the quickstart README.  There is no
point in generating something that `mdformat` will overwrite.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11387)


coryan (approved):
Thanks for all the cleanups.

Issue #7170, opened 2021-08-13
Observability for the storage library


coryan:
We have narrowed this to integration with OpenTelemetry Tracing.  At a high-level:

- We want all RPCs to be reported to OpenTelemetry Tracing.
- We want enough details in the RPCs to troubleshoot networking problems: we need to report IP addresses.
- We want enough details in the RPCs to troubleshoot GCS problems: we need to report the `x-guploader-uploadid` header.
- We want enough details to explain what the library was doing: we need to report the backoffs, any auth-related HTTP requests, the presence of the retry loop, any time blocked waiting for data to arrive or be sent.

A rough breakdown includes:

- [x] Instrument all HTTP requests
- [x] Instrument all gRPC requests
- [x] Instrument the logical function for a HTTP request (e.g. Client::GetObjectMetadata vs Client::ReadObject)
- [x] Report backoff sleeps as spans
- [x] Automatically inject the `TracingClient` to the stack of clients
- [ ] Create long-lived spans for `Client::ReadObject`
- [ ] Create long-lived spans for `Client::WriteObject`
- [x] Fix gRPC stub instrumentation to cover streaming writes (`WriteObject()`)
- [ ] Add thread annotations for long-lived requests
- [ ] Add progress annotations for long uploads
- [ ] Add progress annotations for long downloads

---
This is a broad feature request, with many details to be filled out later.  A number of stakeholders need the library to be more "observable", by which we mean:

- Integrate with tracing systems like [Zipkin](https://zipkin.io/)
- Have counters for interesting events (what is "interesting" is TBD)
- Expose these counters via something like [OpenTelemetry for C++](https://github.com/open-telemetry/opentelemetry-cpp)
- Export such counters to Stackdriver or other monitoring systems

coryan:
We still want this, no progress.

coryan:
We have larger effort around tracing for 2022-Q4

coryan:
Blocked until 2023-Q2

coryan:
We  are tracking the remaining problems in separate bugs.

PR #11407, opened 2023-04-27
impl(storage): add backoff spans

coryan:
Part of the work for #7170

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11407)

### DIFF
--- a/google/cloud/storage/internal/retry_client_test.cc
+++ b/google/cloud/storage/internal/retry_client_test.cc
@@ -17,6 +17,11 @@
 #include "google/cloud/storage/testing/mock_client.h"
 #include "google/cloud/testing_util/chrono_literals.h"
 #include "google/cloud/testing_util/status_matchers.h"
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/retry_client_test.cc
```diff
@@ -17,6 +17,11 @@
 #include "google/cloud/storage/testing/mock_client.h"
 #include "google/cloud/testing_util/chrono_literals.h"
 #include "google/cloud/testing_util/status_matchers.h"
+#ifdef GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
```
dbolduc, line 20: nit: we should only need the `#ifdef` around includes of `<opentelemetry/...>`. Our headers should have this conditional logic built in.

dbolduc (approved):

google/cloud/storage/internal/retry_client_test.cc
coryan: Fixed.

PR #11409, opened 2023-04-27
chore: update the Google Cloud CLI

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11409)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -2105,17 +2105,18 @@ void AutoRun(std::vector<std::string> const& argv) {
 
   auto const snapshot_id = RandomSnapshotId(generator);
 
-  auto ignore_emulator_failures = [](auto lambda) {
-    using ::google::cloud::StatusCode;
-    try {
-      lambda();
-    } catch (google::cloud::Status const& s) {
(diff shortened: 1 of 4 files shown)


google/cloud/pubsub/samples/samples.cc
```diff
@@ -2135,9 +2136,11 @@ void AutoRun(std::vector<std::string> const& argv) {
...
 
   std::cout << "\nRunning UpdateTopic() sample" << std::endl;
-  ignore_emulator_failures([&] {
-    UpdateTopic(topic_admin_client, {project_id, topic_id});
-  });
+  ignore_emulator_failures(
+      [&] {
+        UpdateTopic(topic_admin_client, {project_id, topic_id});
+      },
+      StatusCode::kInvalidArgument);
```
dbolduc, line 2143: So the new emulator implementation has a bug dealing with the `schema_settings` field mask? It is not our code that is at fault?

dbolduc (approved):

google/cloud/pubsub/samples/samples.cc
coryan: yes, b/135688497 It used for fail with `kUnimplemented`, now it fails with `kInvalidArgument` when the field is "labels".  🤷

PR #11410, opened 2023-04-27
test(bigtable): tolerate unimportant zone failures

coryan:
Part of the work for #10413

### DIFF
--- a/google/cloud/bigtable/admin/integration_tests/instance_admin_integration_test.cc
+++ b/google/cloud/bigtable/admin/integration_tests/instance_admin_integration_test.cc
@@ -43,6 +43,7 @@ namespace {
 using ::google::cloud::bigtable::testing::RandomInstanceId;
 using ::google::cloud::internal::GetEnv;
 using ::google::cloud::testing_util::ContainsOnce;
+using ::testing::AnyOf;
 using ::testing::Contains;
 using ::testing::HasSubstr;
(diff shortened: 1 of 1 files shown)


dbolduc (approved):
I duplicated these integration tests. So there is another test fixture where this change could apply.

https://github.com/googleapis/google-cloud-cpp/blob/ce9f5eb89b156f70671ef2d489d1f1f7d039ec99/google/cloud/bigtable/tests/instance_admin_integration_test.cc#L246

https://github.com/googleapis/google-cloud-cpp/blob/ce9f5eb89b156f70671ef2d489d1f1f7d039ec99/google/cloud/bigtable/tests/instance_admin_integration_test.cc#L273

https://github.com/googleapis/google-cloud-cpp/blob/ce9f5eb89b156f70671ef2d489d1f1f7d039ec99/google/cloud/bigtable/tests/instance_admin_integration_test.cc#L388

https://github.com/googleapis/google-cloud-cpp/blob/ce9f5eb89b156f70671ef2d489d1f1f7d039ec99/google/cloud/bigtable/tests/instance_admin_integration_test.cc#L415

coryan:
> I duplicated these integration tests. So there is another test fixture where this change could apply.

Thanks. I was looking for `failed_locations()`. FWIW, I think we should decide if Bigtable or Spanner is where we test the generated admin APIs. There is no need to test in both.

Issue #11309, opened 2023-04-18
docfx: punchlist to publish at cloud site


coryan:
This is a collection of problems noticed on the output of the C++ -> Doxygen -> DocFX -> HTML+Markdown -> Markdown pipeline.

- [x] When there are multiple `*Client` classes the back-to-back links to the examples do not render well.
        ![image](https://user-images.githubusercontent.com/6241635/232796613-654972f8-7caa-4014-b50b-691bc2421218.png)
- [x] Documentation for template member functions (e.g. the `google::cloud::Options` class) does not look right
        ![image](https://user-images.githubusercontent.com/6241635/233231224-be564556-d5dc-4e3e-a71f-f874163fd7db.png)
- [x] We should skip the snippet pages from the table of contents
- [x] The links to external entries are all wrong.  They link to the `cloud.tag` file, they should link to the uid of the external thing.
- [x] We need to escape any `:cloud:` text (e.g. `google::cloud::foo`), or it may render as google::cloud::foo
        ![image](https://user-images.githubusercontent.com/6241635/232872789-ddee7625-1f52-458f-a409-87f76055b293.png)
- [x] We should skip `namespace google` and `namespace google::cloud` from the ToC.
- [x] The `indexpage.md` should be called `index.md` to ensure the document gets in the right place.
- [x] The table of contents should be ordered differently. The landing page first, then namespaces, then any groups and subpages.
- [x] The table of contents should use the C++ names for namespaces
        ![image](https://user-images.githubusercontent.com/6241635/232793454-2be4c095-9807-45ed-95a3-23a95456ebb5.png)
- [x] The table of contents should use the title for pages and groups

PR #11415, opened 2023-04-27
impl(docfx): skip generated pages

coryan:
We generate small doxygen "pages" to host the examples showing how to override the default endpoint and/or authentication mechanism. Listing these pages in the TOC just clutters things. I am adopting a convention that all (and only) generated pages will use `::` as part of their uid.

Part of the work for #11309

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11415)

### DIFF
--- a/docfx/doxygen_pages.cc
+++ b/docfx/doxygen_pages.cc
@@ -130,7 +130,7 @@ std::vector<TocEntry> PagesToc(Config const& cfg,
     // usually examples showing how to use a specific class. We link such
     // examples from the landing page, and they otherwise clutter the navigation
     // page.
-    if (id.find("::") != std::string_view::npos) continue;  // NOLINT
+    if (id.find("::") != std::string_view::npos) continue;
     std::ostringstream title;
     AppendTitle(title, MarkdownContext{}, page);
     auto filename =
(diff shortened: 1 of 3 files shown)


docfx/doxygen_pages.cc
```diff
@@ -126,6 +126,11 @@ std::vector<TocEntry> PagesToc(Config const& cfg,
     auto const& page = i.node();
     if (!IncludeInPublicDocuments(cfg, page)) continue;
     auto const id = std::string_view{page.attribute("id").as_string()};
+    // We will skip groups with :: in their id. These are all generated pages,
+    // usually examples showing how to use a specific class. We link such
+    // examples from the landing page, and they otherwise clutter the navigation
+    // page.
+    if (id.find("::") != std::string_view::npos) continue;  // NOLINT
```
devbww, line 133: Consider adding the `abseil-string-find-str-contains` qualifier (assuming that is it, which I guess bolsters the suggestion).

devbww (approved):

docfx/doxygen_pages.cc
coryan: I just disabled the check in `docfx/.clang-tidy`, seems cleaner.

PR #11421, opened 2023-04-28
chore(deps): update grpc and protobuf

coryan:
We need to update both of these at the same time.  I merged the "Renovate Bot" PRs manually.

- update grpc to v1.55.0-pre1
- update protobuf to v23

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11421)


bazel/google_cloud_cpp_deps.bzl
```diff
@@ -116,11 +116,11 @@ def google_cloud_cpp_deps(name = None):
         http_archive,
         name = "com_google_protobuf",
         urls = [
-            "https://storage.googleapis.com/cloud-cpp-community-archive/com_google_protobuf/v21.12.tar.gz",
-            "https://github.com/protocolbuffers/protobuf/archive/v21.12.tar.gz",
+            "https://storage.googleapis.com/cloud-cpp-community-archive/com_google_protobuf/v23.1.tar.gz",
+            "https://github.com/protocolbuffers/protobuf/archive/refs/tags/v23.1.tar.gz",
```
devbww, line 120: Should we drop the "refs/tags/" (et seq)?

generator/integration_tests/golden/v1/internal/golden_kitchen_sink_option_defaults.cc
```diff
@@ -36,7 +36,7 @@ Options GoldenKitchenSinkDefaultOptions(Options options) {
   options = google::cloud::internal::PopulateCommonOptions(
       std::move(options), "GOLDEN_KITCHEN_SINK_ENDPOINT",
       "GOLDEN_KITCHEN_SINK_EMULATOR_HOST", "GOLDEN_KITCHEN_SINK_AUTHORITY",
-      "goldenkitchensink.googleapis.com");
+"goldenkitchensink.googleapis.com");
```
devbww, line 39: Q. Did the generator change in the meantime (x3)?

coryan: Not the generator: the Protobuf helpers that the generator depends on emits different output in this case.

bazel/google_cloud_cpp_deps.bzl
coryan: Done.  FWIW, I think it makes no difference.  Both redirect to the same URL.

generator/integration_tests/golden/v1/internal/golden_kitchen_sink_option_defaults.cc
devbww: OK, although I'm not sure I understand that given ... https://github.com/googleapis/google-cloud-cpp/blob/721a63a82e8ffff7aad408dde557737bb9fbdfe5/generator/internal/option_defaults_generator.cc#L128-L144

But I'll look later.

coryan: I am not sure I understand either.  I just don't think it matters.  These golden files will always be subject to changes in the Protobuf `Printer` class behavior.  Fortunately we `clang-format` the generated libraries.

FWIW, the class in question when through a number of changes:

https://github.com/protocolbuffers/protobuf/blob/21.x/src/google/protobuf/io/printer.cc

https://github.com/protocolbuffers/protobuf/blob/23.x/src/google/protobuf/io/printer.cc

and continues to change at head:

https://github.com/protocolbuffers/protobuf/blame/main/src/google/protobuf/io/printer.cc

PR #11423, opened 2023-04-28
doc: list main classes in landing page 

### LINKED ISSUE #11309: docfx: punchlist to publish at cloud site

coryan:
We want to make it easy to find these classes. This adds to the clutter,
but I will remove the samples to override endpoints and default
credentials to a separate page in a future pass.

Fixes #11309 

🆘 Please review the text before the list of classes, there are two versions and I know it could use some copy-editing.

Thanks.


ci/generate-markdown/update-library-landing-dox.sh
```diff
@@ -68,6 +68,43 @@ EOT
...
+  sed '/<!-- inject-client-list-start -->/q' "${MAIN_DOX}"
+  if [[ ${#samples_cc[@]} -eq 0 ]]; then
+    true
+  elif [[ ${#samples_cc[@]} -eq 1 ]]; then
+    sample_cc="${samples_cc[0]}"
+    client_name="$(sed -n '/main-dox-marker: / s;// main-dox-marker: \(.*\);\1;p' "${sample_cc}")"
+    cat <<_EOF_
+The main class in this library is
+[\`${client_name}\`](@ref google::cloud::${client_name}),
+all RPCs are exposed as member functions in this class. Other classes provide
```
devbww, line 81: `s/, all/. All/`?

ci/generate-markdown/update-library-landing-dox.sh
```diff
@@ -68,6 +68,43 @@ EOT
...
+[\`${client_name}\`](@ref google::cloud::${client_name}) when testing your
+application.
+_EOF_
+  else
+    cat <<'_EOF_'
+This library offers multiple `*Client` classes, which are listed below. Each
+one of these classes exposes all the RPCs for a gRPC `service` as member
+functions of the class. This library groups multiple gRPC services because they
+are part of the same product or are often used together. A typical example may
+be the administrative and data place operations for a single product.
```
devbww, line 92: `s/place/plane/`?

ci/generate-markdown/update-library-landing-dox.sh
devbww, line 81: Just below we say, "as member functions _of_ this class".

devbww (approved):

ci/generate-markdown/update-library-landing-dox.sh
coryan: Fixed

ci/generate-markdown/update-library-landing-dox.sh
coryan: Fixed

ci/generate-markdown/update-library-landing-dox.sh
coryan: Used "*of* this class" in both places

PR #11424, opened 2023-04-28
cleanup(otel): instrumentation scope

dbolduc:
Mini cleanup. The version is an optional argument to create a tracer. So there's no point in adding an empty attribute for it when it is not set.

Also rename the keys. The names are shorter. The Go exporter uses these names at HEAD. I think the exporter spec is just stale.

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/327f5866b4fc60227d44c666c9ccb160a8ef9bf6/exporter/trace/trace_proto.go#L68-L69

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11424)

### DIFF
--- a/google/cloud/opentelemetry/internal/recordable.cc
+++ b/google/cloud/opentelemetry/internal/recordable.cc
@@ -302,8 +302,9 @@ void Recordable::SetInstrumentationScope(
     opentelemetry::sdk::instrumentationscope::InstrumentationScope const&
         instrumentation_scope) noexcept {
   SetAttribute("otel.scope.name", instrumentation_scope.GetName());
-  if (instrumentation_scope.GetVersion().empty()) return;
-  SetAttribute("otel.scope.version", instrumentation_scope.GetVersion());
+  if (!instrumentation_scope.GetVersion().empty()) {
+    SetAttribute("otel.scope.version", instrumentation_scope.GetVersion());
+  }
 }
 
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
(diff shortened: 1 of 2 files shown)


google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -301,10 +301,9 @@ void Recordable::SetDuration(std::chrono::nanoseconds duration) noexcept {
 void Recordable::SetInstrumentationScope(
     opentelemetry::sdk::instrumentationscope::InstrumentationScope const&
         instrumentation_scope) noexcept {
-  SetAttribute("otel.instrumentation_library.name",
-               instrumentation_scope.GetName());
-  SetAttribute("otel.instrumentation_library.version",
-               instrumentation_scope.GetVersion());
+  SetAttribute("otel.scope.name", instrumentation_scope.GetName());
+  if (instrumentation_scope.GetVersion().empty()) return;
+  SetAttribute("otel.scope.version", instrumentation_scope.GetVersion());
```
devbww, line 306: I'd suggest `if (!x.empty()) SetAttribute(a, x)` instead.  It's not like the emptiness of the version makes the rest of the function improper, even though the effect is currently the same because the rest of the function is currently empty.

devbww (approved):

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: Done.

PR #11425, opened 2023-04-28
impl(otel): map attribute keys

dbolduc:
Part of the work for #11156 

Translate between OpenTelemetry semantic conventions for attribute names and Cloud Trace semantics for labels.

This is probably a "big whoop". I suspect it will matter more when we get into monitored resources. As far as I can tell this is the only difference:

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11425)

### DIFF
--- a/google/cloud/opentelemetry/internal/recordable.cc
+++ b/google/cloud/opentelemetry/internal/recordable.cc
@@ -32,9 +32,9 @@ bool IsTrailByte(char x) { return static_cast<signed char>(x) < -0x40; }
 // Trace's semantics for label keys. So translate from one to the other.
 //
 // See: https://cloud.google.com/trace/docs/trace-labels#canonical_labels
-opentelemetry::nostd::string_view MapKey(
-    opentelemetry::nostd::string_view key) {
-  static auto const* m = new std::map<std::string, std::string>{
+void MapKey(opentelemetry::nostd::string_view& key) {
+  static auto const* m = new std::map<opentelemetry::nostd::string_view,
+                                      opentelemetry::nostd::string_view>{
       {"http.host", "/http/host"},
       {"http.method", "/http/method"},
       {"http.target", "/http/path"},
@@ -46,9 +46,8 @@ opentelemetry::nostd::string_view MapKey(
       {"http.scheme", "/http/client_protocol"},
       {"http.route", "/http/route"},
   };
-  auto it = m->find({key.data(), key.size()});
-  if (it == m->end()) return key;
-  return it->second;
+  auto it = m->find(key);
+  if (it != m->end()) key = it->second;
 }
 
 class AttributeVisitor {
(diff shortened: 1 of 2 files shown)


google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -28,6 +28,29 @@ namespace {
 // http://35.193.25.4/TensorFlow/models/research/syntaxnet/util/utf8/unilib_utf8_utils.h
 bool IsTrailByte(char x) { return static_cast<signed char>(x) < -0x40; }
 
+// OpenTelemetry's semantic conventions for attribute keys differ from Cloud
+// Trace's semantics for label keys. So translate from one to the other.
+//
+// See: https://cloud.google.com/trace/docs/trace-labels#canonical_labels
+opentelemetry::nostd::string_view MapKey(
+    opentelemetry::nostd::string_view key) {
+  static auto const* m = new std::map<std::string, std::string>{
```
devbww, line 37: It seems like you should be able to make this `string_view -> string_view` and avoid the conversion to `string` on lookup.

google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -28,6 +28,29 @@ namespace {
...
+      {"http.status_code", "/http/status_code"},
+      {"http.url", "/http/url"},
+      {"http.user_agent", "/http/user_agent"},
+      {"http.request_content_length", "/http/request/size"},
+      {"http.response_content_length", "/http/response/size"},
+      {"http.scheme", "/http/client_protocol"},
+      {"http.route", "/http/route"},
+  };
+  auto it = m->find({key.data(), key.size()});
+  if (it == m->end()) return key;
```
devbww, line 50: Observation:  It doesn't matter in the one call to this function (because the result it assigned to the argument), but a `string_view` returning function usually has to make some statement about the lifetime of its result.

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: Fixed, thanks.

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: Changed the function to just modify a `string_view&`.

I assume the data in that `static auto const* m` thing lasts forever.

devbww: Yes ... the `"..."` string literals are valid for the entire program.

I guess the new signature is a little clearer in that "no change" is now really no change (instead of self assignment).  The function should probably still say that it would only extend the lifetime, but perhaps that is now clear without further statement (and it is only file local).

devbww (approved):

google/cloud/opentelemetry/internal/recordable.cc
coryan: I liked the original signature better 🤷

PR #11427, opened 2023-04-28
ci: initial docfx publication pipeline

coryan:
Change the build to start the publication of DocFX documents for `common`, `kms`, and `secretmanager`. Once these work we will add additional libraries in a series of PRs. . The reference docs are not public yet, but fine tuning the output is easier if they are published internally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11427)

### DIFF
--- a/ci/cloudbuild/builds/publish-docs.sh
+++ b/ci/cloudbuild/builds/publish-docs.sh
@@ -56,8 +56,10 @@ fi
 # For PR and manual builds, publish to the staging site. For CI builds (release
 # and otherwise), publish to the public URL.
 bucket="test-docs-staging"
+docfx_bucket="docs-staging-v2-dev"
 if [[ "${TRIGGER_TYPE}" == "ci" ]]; then
   bucket="docs-staging"
+  docfx_bucket="docs-staging-v2"
 fi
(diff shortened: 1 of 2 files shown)


coryan (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11427)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @scotthart)

___
*[`ci/cloudbuild/builds/publish-docs.sh` line 161 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11427#-NU7MjS8CV0ifhiRDtMR:-NU7oq-s8NXipgrjEGSO:bvgsmxx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/58393090450b431a762a0f2d68f38295cacaa3a7/ci/cloudbuild/builds/publish-docs.sh#L161)):*

just curious, what does this comparison achieve?
</blockquote></details>

Explained in the bug, but the plan is to slowly (maybe not that slowly) change that `"z"` to lower and lower letters in the alphabet.

PR #11433, opened 2023-04-28
impl(generator): support generating, building, and installing local protos

scotthart:
part of the work for #10980 

Because code generated from protos captures proto file line numbers, we need to:

1. emit protos
2. format protos
3. emit C++
4. format C++

Also add new googleapis proto dep and update automated tools to handle local protos.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11433)

### DIFF
--- a/cmake/CompileProtos.cmake
+++ b/cmake/CompileProtos.cmake
@@ -231,9 +231,21 @@ endfunction ()
 #
 # ${EXTERNAL_GOOGLEAPIS_SOURCE}/google/bigtable/v2/bigtable.proto
 #
-# ARGV2 can be specified as an optional path to substitute in lieu of
-# ${EXTERNAL_GOOGLEAPIS_SOURCE}
 function (google_cloud_cpp_load_protolist var file)
+    cmake_parse_arguments(
+        _opt
+        # No boolean flags
+        ""
+        # If present PROTO_DIR overrides the default
+        # "${EXTERNAL_GOOGLEAPIS_SOURCE}"
+        "PROTO_DIR"
+        # No multi-argument flags
+        ""
+        ${ARGN})
+    if (NOT DEFINED _opt_PROTO_DIR)
+        set(_opt_PROTO_DIR "${EXTERNAL_GOOGLEAPIS_SOURCE}")
+    endif ()
+
     file(READ "${file}" contents)
     string(REGEX REPLACE "\n" ";" contents "${contents}")
     set(protos)
@@ -243,11 +255,7 @@ function (google_cloud_cpp_load_protolist var file)
         if ("${line}" STREQUAL "")
             continue()
         endif ()
-        if (DEFINED ARGV2)
-            list(APPEND protos "${ARGV2}/${line}")
-        else ()
-            list(APPEND protos "${EXTERNAL_GOOGLEAPIS_SOURCE}/${line}")
-        endif ()
+        list(APPEND protos "${_opt_PROTO_DIR}/${line}")
     endforeach ()
     set(${var}
         "${protos}"
--- a/cmake/CompileProtos.cmake
+++ b/cmake/CompileProtos.cmake
@@ -231,6 +231,8 @@ endfunction ()
 #
 # ${EXTERNAL_GOOGLEAPIS_SOURCE}/google/bigtable/v2/bigtable.proto
 #
+# ARGV2 can be specified as an optional path to substitute in lieu of
+# ${EXTERNAL_GOOGLEAPIS_SOURCE}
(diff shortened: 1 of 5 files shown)


cmake/CompileProtos.cmake
```diff
@@ -231,6 +231,8 @@ endfunction ()
 #
 # ${EXTERNAL_GOOGLEAPIS_SOURCE}/google/bigtable/v2/bigtable.proto
 #
+# ARGV2 can be specified as an optional path to substitute in lieu of
```
coryan, line 234: The convention in CMake is to use `cmake_parse_arguments()`:

https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html

You can define a single-argument optional, e.g.

```cmake
function (google_cloud_cpp_load_protolist var file)
  cmake_parse_arguments(_opt
   # No boolean flags
   "" 
   # If present PROTO_DIR overrides the default "${EXTERNAL_GOOGLEAPIS_SOURCE}"
   "PROTO_DIR"
   # No multi-argument flags
    ""
  ${ARGN})
 if (NOT DEFINED _opt_PROTO_DIR)
   set(_opt_PROTO_DIR "${EXTERNAL_GOOGLEAPIS_SOURCE}")
 endif ()
...

```

coryan (changes requested):

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11433)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`cmake/CompileProtos.cmake` line 234 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11433#gh-1181082057:-NUNC9rQCCk6RiT2uPKj:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f921574d2e5c4a5402c088e578b2ff5e46b203b9/cmake/CompileProtos.cmake#L234)):*

The convention in CMake is to use `cmake_parse_arguments()`:

https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html

You can define a single-argument optional, e.g.

```cmake
function (google_cloud_cpp_load_protolist var file)
  cmake_parse_arguments(_opt
   # No boolean flags
   "" 
   # If present PROTO_DIR overrides the default "${EXTERNAL_GOOGLEAPIS_SOURCE}"
   "PROTO_DIR"
   # No multi-argument flags
    ""
  ${ARGN})
 if (NOT DEFINED _opt_PROTO_DIR)
   set(_opt_PROTO_DIR "${EXTERNAL_GOOGLEAPIS_SOURCE}")
 endif ()
...

```
</blockquote></details>

Done.

coryan (approved):
> Reviewed 3 of 5 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11433)* status: all files reviewed, 1 unresolved discussion (waiting on @scotthart)

___
*[`generator/standalone_main.cc` line 123 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11433#-NUX92vw1-Kro4R4J_dR:-NUX92vw1-Kro4R4J_dS:b-tgo7zl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/134f377f0e0f058ae4df11ec590d9465df5d569e/generator/standalone_main.cc#L123)):*
> ```cpp
>     std::string const& output_path) {
>   std::vector<std::string> install_directories{".", "./lib64", "./lib64/cmake"};
>   google::protobuf::RepeatedPtrField<
> ```

nit: I would think `auto` is cleaner here.  And I am not just saying that as a AAA devotee.

```
auto services = config.service();
```

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11433)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`generator/standalone_main.cc` line 123 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11433#-NUX92vw1-Kro4R4J_dR:-NUXTqry4oxS7t57N3CK:b-pbdqz3) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/134f377f0e0f058ae4df11ec590d9465df5d569e/generator/standalone_main.cc#L123)):*

nit: I would think `auto` is cleaner here.  And I am not just saying that as a AAA devotee.

```
auto services = config.service();
```


</blockquote></details>

Agreed.

PR #11444, opened 2023-04-29
impl(otel): traced streaming writes in stubs

dbolduc:
Part of the work for #10620 

Note that there is no associated Connection level change, because all streaming writes are handwritten at that layer.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11444)


coryan (approved):

coryan:
Also helps with #7170

PR #11478, opened 2023-05-02
docs: update set up workstation how-to

alevenberg:
* Update doc to handle x86_64 (uname -p produces unknown for x86_64 on my VM)
* Add sudo before running the install google sdk script so the mkdir commands can create directories
* Add a note about installing clangtidy in newer versions of ubuntu

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11478)


ci/install-cloud-sdk.sh
```diff
@@ -26,7 +26,11 @@ ARCH="$(uname -p)"
 if [[ "${ARCH}" == "aarch64" ]]; then
   # The tarball uses this name
   ARCH="arm"
-fi
+else 
+  # For x86_64, uname -p might give unknown 
+  ARCH="$(uname -m)"
```
coryan, line 31: Thanks.  Sounds like we should simply change line 25?

doc/contributor/howto-guide-setup-development-workstation.md
```diff
@@ -61,6 +61,11 @@ sudo update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang
 sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-10 100
 ```
 
+Note: newer versions of Ubuntu might require 
+```console
+sudo apt install -y clang libc++-dev libc++abi-dev cmake ninja-build
```
coryan, line 66: We are about to drop Ubuntu 18.  Maybe just update the instructions to Debian / Ubuntu 22?  Or just remove the obsolete instructions!

coryan (changes requested):
Seems like we should be removing some existing code at the same time?

coryan:
FYI, we use conventional commits: https://www.conventionalcommits.org/en/v1.0.0/ The title for this PR should be something like:

```
ci: update script to install gcloud CLI
```

or

```
docs: update set up workstation how-to
```

The fact that both are possible suggests they could be two separate PRs.  But for small changes like this mixing two things is not the end of the world.

ci/install-cloud-sdk.sh
alevenberg: How would I verify this commands works as intended on the ARM architecture?

coryan: We don't have a standard answer for this, or even good non-standard answers (this is a gap in our automated testing). One approach is to spin up a T2A VM:

https://cloud.google.com/compute/docs/machine-resource#arm

Another approach would be to run a Docker container with qemu.  I have read that is doable, don't know how to make it work.

doc/contributor/howto-guide-setup-development-workstation.md
alevenberg: Is clang 1:1 with clang-10, clang-tidy, and clang-formatter? I'll remove the old instructions if it is

coryan: I /think/ (emphasis on "I might be wrong") `clang-10` installs an specific version of Clang, while `clang` installs the default version.  The same applies to the other Clang-related packages.  With older versions of Ubuntu we probably wanted to name an specific version because the default was "too old".  With something like Ubuntu 22 (or Debian 11) we are happy with the default.

I am not even sure I even landed in the ballpark of an answer to your question.

ci/install-cloud-sdk.sh
alevenberg: Ok I'll look into both

doc/contributor/howto-guide-setup-development-workstation.md
alevenberg: That explanation makes sense to me. I'll remove the old instructions.

alevenberg: Discussed outside of PR. We will be keeping both since other users might have varying Ubuntu versions.

ci/install-cloud-sdk.sh
```diff
@@ -22,11 +22,12 @@ declare -A -r GOOGLE_CLOUD_CPP_SDK_SHA256=(
   ["arm"]="d3d7d8bdde1abc8e8279b813b3341d20fa3af1928268d97c61b0553a8e590124"
 )
 
-ARCH="$(uname -p)"
+ARCH="$(uname -m)"
 if [[ "${ARCH}" == "aarch64" ]]; then
   # The tarball uses this name
   ARCH="arm"
 fi
+
```
coryan, line 30: nit: remove the additional blank line.

coryan (approved):

coryan (approved):
:lgtm:

> Reviewed 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11478#-:-NUmKeNzA24m-YIMIMe6:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alevenberg)

PR #11482, opened 2023-05-03
cleanup(docfx): use `pugi::xml_node` as pointers

coryan:
A `pugi::xml_node` is akin to an iterator or a pointer. They should be passed by value and not by `const&`. It is possible some additional cleanups will be needed, but this should have caught most cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11482)

### DIFF
--- a/docfx/doxygen2markdown.cc
+++ b/docfx/doxygen2markdown.cc
@@ -49,13 +49,13 @@ std::string EscapeCodeLink(std::string link) {
 //    </xsd:complexType>
 // clang-format on
 bool AppendIfSect4(std::ostream& os, MarkdownContext const& ctx,
-                   pugi::xml_node const& node) {
+                   pugi::xml_node node) {
   if (std::string_view{node.name()} != "sect4") return false;
   // A single '#' title is reserved for the document title. The sect4 title uses
   // '#####':
   os << "\n\n##### ";
   AppendTitle(os, ctx, node);
-  for (auto const& child : node) {
+  for (auto const child : node) {
     // Unexpected: internal -> we do not use this.
     if (std::string_view(child.name()) == "title") continue;  // already handled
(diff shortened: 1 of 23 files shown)


docfx/doxygen2markdown.cc
```diff
@@ -986,18 +986,18 @@ bool AppendIfSimpleSect(std::ostream& os, MarkdownContext const& ctx,
...
   // Markdown.
   return std::string_view{node.name()} == "anchor";
 }
 
 void AppendTitle(std::ostream& os, MarkdownContext const& ctx,
-                 pugi::xml_node const& node) {
+                 pugi::xml_node node) {
   // The XML schema says there is only one of these, but it is easier to write
   // the loop.
   for (auto const& title : node.children("title")) {
```
dbolduc, line 999: s/const&/const/

docfx/doxygen2syntax.cc
```diff
@@ -155,7 +153,7 @@ std::string ReturnDescription(YamlContext const& /*ctx*/,
 //   </xsd:complexType>
 // clang-format on
 bool ParameterItemMatchesName(std::string_view parameter_name,
-                              pugi::xml_node const& item) {
+                              pugi::xml_node item) {
   for (auto const& list : item.children("parameternamelist")) {
     for (auto const& name : list.children("parametername")) {
```
dbolduc, line 158: s/const&/const/

dbolduc (approved):

docfx/doxygen2syntax.cc
coryan: Fixed and a couple more.

docfx/doxygen2markdown.cc
coryan: Fixed.

PR #11483, opened 2023-05-03
fix(docfx): skip nested namespaces in children list

### LINKED ISSUE #11452: Nested namespaces are listed but there is no title.
image

coryan:
Listing the nested namespaces as children is (arguably) technically correct, but it does not render right, and does not aid in navigation.

Fixes #11452

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11483)

### DIFF
--- a/docfx/doxygen2references.cc
+++ b/docfx/doxygen2references.cc
@@ -50,6 +50,7 @@ std::list<Reference> ExtractReferences(YamlContext const& ctx,
   auto const name = std::string_view{node.name()};
   // Skip <innernamespace> elements. They are listed in ToC (the left-side
   // navigation).
+  if (name == "innernamespace") return {};
   if (name == "innerclass") {
     auto const uid = std::string{node.attribute("refid").as_string()};
     return {Reference(uid, node.child_value())};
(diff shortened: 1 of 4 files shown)


dbolduc (approved):

docfx/doxygen2references.cc
```diff
@@ -48,10 +48,8 @@ std::list<Reference> ExtractReferences(YamlContext const& ctx,
   if (!IncludeInPublicDocuments(ctx.config, node)) return {};
 
   auto const name = std::string_view{node.name()};
-  if (name == "innernamespace") {
-    auto const uid = std::string{node.attribute("refid").as_string()};
-    return {Reference(uid, node.child_value())};
-  }
+  // Skip <innernamespace> elements. They are listed in ToC (the left-side
+  // navigation).
```
dbolduc, line 52: Consider adding the following to be explicit
```cc
if (name == "innernamespace") return {};
```

dbolduc: Well, you know better than I whether we should `RecurseReferences` for this case or not.

coryan: Done.

PR #11493, opened 2023-05-03
impl(docfx): generate full table of contents

coryan:
Change the table of contents generation to include namespaces, classes, structs, functions, enums, typedefs, etc. Group elements by "type", so structs appear together. The grouping matches the groups in the right-side navigation for a namespace or class. Compound elements (namespaces, classes, structs) have an 'Overview' link to point to the element description. There is no way to have a grouping that is also a link, that is, a group for "all the stuff in class Foo that also links to Foo" is not a thing.

Links to enums and typedefs are not pointing anywhere at the moment, that will be fixed in separate PRs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11493)

### DIFF
--- a/docfx/toc_entry.cc
+++ b/docfx/toc_entry.cc
@@ -13,12 +13,33 @@
 // limitations under the License.
 
 #include "docfx/toc_entry.h"
+#include <algorithm>
 #include <iostream>
 
 namespace docfx {
 
+bool operator==(TocEntry const& lhs, TocEntry const& rhs) {
+  if (lhs.name != rhs.name || lhs.attr != rhs.attr) return false;
+  return std::equal(lhs.items.begin(), lhs.items.end(), rhs.items.begin(),
+                    rhs.items.end(),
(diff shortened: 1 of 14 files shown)


docfx/toc_entry.cc
```diff
@@ -13,12 +13,33 @@
 // limitations under the License.
 
 #include "docfx/toc_entry.h"
+#include <algorithm>
 #include <iostream>
 
 namespace docfx {
 
+bool operator==(TocEntry const& lhs, TocEntry const& rhs) {
+  if (lhs.name != rhs.name || lhs.attr != rhs.attr) return false;
```
devbww, line 22: Possible nit:
```
  return lhs.name == rhs.name && lhs.attr == rhs.attr && std::equal(...);
```

devbww (approved):

docfx/toc_entry.cc
coryan: Done.

Issue #11455, opened 2023-05-01
The `enum` values are not generated in any HTML document.


coryan:

coryan:
Looks like we need to create a separate `.yml` file for the enum definition.

PR #11497, opened 2023-05-04
impl(docfx): generate separate file for enums

coryan:
The tooling at google expects enums to be generated as separate files. This adds a small loop in `main()` to generate the files. We need to remove the enum definitions from the previous yml files, or we get duplicate definitions. And to make the enumvalues visible we need to list them as children of the enum.

Part of the work for #11455

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11497)

### DIFF
--- a/docfx/doxygen2yaml.cc
+++ b/docfx/doxygen2yaml.cc
@@ -68,8 +68,7 @@ void CompoundRecurse(YAML::Emitter& yaml, YamlContext const& ctx,
   for (auto const child : node) {
     if (!IncludeInPublicDocuments(ctx.config, child)) continue;
     if (IgnoreForRecurse(child)) continue;
-    // Enums need to get their own files, never recurse in them.
-    // if (AppendIfEnum(yaml, ctx, child)) continue;
+    // Enums need to get their own files, so never recurse in them:
     if (kind(child) == "enum") continue;
     if (AppendIfSectionDef(yaml, ctx, child)) continue;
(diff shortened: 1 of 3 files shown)


docfx/doxygen2yaml.cc
```diff
@@ -68,12 +68,14 @@ void CompoundRecurse(YAML::Emitter& yaml, YamlContext const& ctx,
   for (auto const child : node) {
     if (!IncludeInPublicDocuments(ctx.config, child)) continue;
     if (IgnoreForRecurse(child)) continue;
+    // Enums need to get their own files, never recurse in them.
```
devbww, line 71: s/,/, so/

docfx/doxygen2yaml.cc
```diff
@@ -68,12 +68,14 @@ void CompoundRecurse(YAML::Emitter& yaml, YamlContext const& ctx,
...
+    // if (AppendIfEnum(yaml, ctx, child)) continue;
```
devbww, line 72: Remnant?

devbww (approved):

docfx/doxygen2yaml.cc
coryan: Fixed.

docfx/doxygen2yaml.cc
coryan: Kind of, I thought people would serve as "here is the code that you think may be needed". But it was a bad idea. Removed.

PR #11499, opened 2023-05-04
fix: patch releases do not change the ABI

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11499)


google/cloud/version.h
```diff
@@ -20,7 +20,7 @@
 #include "google/cloud/internal/version_info.h"
 #include <string>
 
-#define GOOGLE_CLOUD_CPP_VCONCAT(Ma, Mi, Pa) v##Ma##_##Mi##_##Pa
+#define GOOGLE_CLOUD_CPP_VCONCAT(Ma, Mi, Pa) v##Ma##_##Mi
 #define GOOGLE_CLOUD_CPP_VEVAL(Ma, Mi, Pa) GOOGLE_CLOUD_CPP_VCONCAT(Ma, Mi, Pa)
 #define GOOGLE_CLOUD_CPP_NS                              \
   GOOGLE_CLOUD_CPP_VEVAL(GOOGLE_CLOUD_CPP_VERSION_MAJOR, \
```
devbww, line 26: Perhaps we should also not pass `GOOGLE_CLOUD_CPP_VERSION_PATCH` to `GOOGLE_CLOUD_CPP_VEVAL` (and `GOOGLE_CLOUD_CPP_VCONCAT`)?

coryan: Perhaps. `<lawyerly>`In an abundance of caution I decided to keep the macros backwards compatible, because it is not clear if they are part of the public API or not.`</lawerly>`.  Thoughts?

devbww: Ack.

But while it might be a little hazy whether `GOOGLE_CLOUD_CPP_NS` itself is public, hopefully `VEVAL` and `VCONCAT` are more clearly implementation details.  We should have put "INTERNAL" in their names I guess.

In the end though, I'm happy with your abundance of caution if you're unconvinced.

devbww (approved):

PR #11511, opened 2023-05-05
impl(sql): enable integration tests

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11511)

### DIFF
--- a/google/cloud/sql/integration_tests/CMakeLists.txt
+++ b/google/cloud/sql/integration_tests/CMakeLists.txt
@@ -53,3 +53,5 @@ function (google_cloud_cpp_sql_define_integration_tests)
         return()
     endif ()
 endfunction ()
+
+google_cloud_cpp_sql_define_integration_tests()
--- a/google/cloud/sql/integration_tests/CMakeLists.txt
+++ b/google/cloud/sql/integration_tests/CMakeLists.txt
@@ -14,44 +14,36 @@
 # limitations under the License.
 # ~~~
 
-function (google_cloud_cpp_sql_define_integration_tests)
-    # The tests require googletest to be installed. Force CMake to use the
-    # config file for googletest (that is, the CMake file installed by
-    # googletest itself), because the generic `FindGTest` module does not define
-    # the GTest::gmock target, and the target names are also weird.
-    find_package(GTest CONFIG REQUIRED)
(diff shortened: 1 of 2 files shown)


google/cloud/sql/integration_tests/CMakeLists.txt
```diff
@@ -53,3 +53,5 @@ function (google_cloud_cpp_sql_define_integration_tests)
         return()
     endif ()
 endfunction ()
+
+google_cloud_cpp_sql_define_integration_tests()
```
devbww, line 57: Does this need to be guarded by an `if (BUILD_TESTING)` conditional?  (All similar ones seem to be.)

dbolduc: > Does this need to be guarded

it is: https://github.com/googleapis/google-cloud-cpp/blob/8b4fc2f96604b704b2629dc7a5e454d34af4e533/google/cloud/sql/CMakeLists.txt#L126

---

We don't need to define a `function`, then call it. We can just run the things inside.

dbolduc (approved):

google/cloud/sql/integration_tests/CMakeLists.txt
devbww: OK, thanks.  I'll just note that sometimes we have `add_subdirectory(integration_tests)` being conditional, and sometimes not.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11511)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @dbolduc and @devbww)

___
*[`google/cloud/sql/integration_tests/CMakeLists.txt` line 57 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11511#gh-1186348372:-NUh2qXI4UHl2W7C6Xpc:b8af9ec) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/4f58e017dd0d415ebf623408108cf35d44744d3f/google/cloud/sql/integration_tests/CMakeLists.txt#L57)):*

OK, thanks.  I'll just note that sometimes we have `add_subdirectory(integration_tests)` being conditional, and sometimes not.
</blockquote></details>

Removed function definition plus call.

PR #11515, opened 2023-05-06
cleanup(ci): disentangle cmake::common_args from GOOGLE_CLOUD_CPP_ENABLE

devbww:
The enabled features aren't really a "common argument" as more often than not we override them with an extra `-DGOOGLE_CLOUD_CPP_ENABLE` setting. So, stop adding a setting in `cmake::common_args`, and instead have each build script specify its own (typically starting from a `features::*` call).  This makes what we're building much clearer in the scripts, and when looking at an executed CMake command.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11515)


coryan (approved):
Thanks for cleaning this up.

PR #11518, opened 2023-05-07
impl: prepare for next Protobuf release

coryan:
On Windows, `GetCurrentTime()` is a macro. The current and previous releases of Protobuf undefine this macro in one of its headers.  That was probably a bug in Protobuf.  We use `TimeUtil::GetCurrentTime()`. With the next release of Protobuf that will expand the macro. We either need to prevent the macro expansion, using something like `(TimeUtil::GetCurrentTime)()`, or we need to avoid the function altogether.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11518)

### DIFF
--- a/google/cloud/bigtable/tests/admin_backup_integration_test.cc
+++ b/google/cloud/bigtable/tests/admin_backup_integration_test.cc
@@ -96,7 +96,7 @@ TEST_F(AdminBackupIntegrationTest, CreateListGetUpdateRestoreDeleteBackup) {
   EXPECT_EQ(backup->name(), backup_name);
 
   // Update backup
-  expire_time = expire_time + std::chrono::hours(12);
+  expire_time += std::chrono::hours(12);
   backup = table_admin_->UpdateBackup({cluster_id, backup_id, expire_time});
   ASSERT_STATUS_OK(backup);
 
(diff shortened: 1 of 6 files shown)


google/cloud/bigtable/tests/admin_backup_integration_test.cc
```diff
@@ -97,17 +96,16 @@ TEST_F(AdminBackupIntegrationTest, CreateListGetUpdateRestoreDeleteBackup) {
   EXPECT_EQ(backup->name(), backup_name);
 
   // Update backup
-  expire_time = expire_time + TimeUtil::HoursToDuration(12);
-  backup = table_admin_->UpdateBackup(
-      {cluster_id, backup_id,
-       google::cloud::internal::ToChronoTimePoint(expire_time)});
+  expire_time = expire_time + std::chrono::hours(12);
```
devbww, line 99: Nit: `+=`.

devbww (approved):

google/cloud/bigtable/tests/admin_backup_integration_test.cc
coryan: Done.

Issue #8755, opened 2022-04-14
Consider offering a full jitter backoff policy


dbolduc:
Here is a write up on [full jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/). Apparently the java clients [do this](https://github.com/googleapis/gax-java/blob/main/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java#L253-L256).

With pure exponential backoff policy we might have backoff ranges like:
req1: (.5, 1)
req2: (1, 2)
req3: (2, 4)
....

With full jitter, these backoff ranges would look like:
req1: (0, 1)
req2: (0, 2)
req3: (0, 4)
....

(make up your own units)

coryan:
For some services (storage comes to mind) a minimum back off is recommended.

coryan:
@dbolduc is going to research some more and make a recommendation.

dbolduc:
This is the code to run the simulation in the article: https://github.com/aws-samples/aws-arch-backoff-simulator/blob/master/src/backoff_simulator.py

I added our backoff algorithm:
```diff
34a35,39
> class ExpoBackoffCloudCpp(Backoff):
>     def backoff(self, n):
>         v = self.expo(n)
>         return random.uniform(v/2, v)
> 
```

And a MinJitter (i.e. full jitter, but it doesn't start at 0)
```diff
39a45,49
> class ExpoBackoffMinJitter(Backoff):
>     def backoff(self, n):
>         v = self.expo(n)
>         return random.uniform(self.base, v)
> 
```

Results:
![image](https://user-images.githubusercontent.com/23088558/189767201-5b1d98fb-b4d1-47ee-b3fa-ce8d183c68a1.png)

![image](https://user-images.githubusercontent.com/23088558/189767255-5261e03c-dc60-4336-b0d1-42d2f0e9b38b.png)

To summarize (with these exact settings, in this exact model), min jitter is indistinguishable from full jitter. Our strategy is indistinguishable from the jitter strategies in terms of total calls, but takes more time to complete the work.

I will phone a friend before making a recommendation...

dbolduc:
Seems like other client library languages do full jitter only. Given the supposed, slight performance benefit, I think we should implement min-jitter (which is like full jitter, but slightly more general).

[...]

PR #11529, opened 2023-05-08
fix: correct exponential backoff ranges

### LINKED ISSUE #11107: Exponential Backoff ranges are incorrect
We have a `scaling` factor, but we multiply the initial range by 2, instead.

https://github.com/googleapis/google-cloud-cpp/blob/9737c4bfc27eea6c593dcbe2d59af41d266acc0c/google/cloud/internal/backoff_policy.h#L131

Moreover, we do not check if we have hit the maximum until after generating the first backoff 

https://github.com/googleapis/google-cloud-cpp/blob/9737c4bfc27eea6c593dcbe2d59af41d266acc0c/google/cloud/internal/backoff_policy.cc#L45-L55

This leads to kooky behavior. Let's say I have an `ExponentialBackoffPolicy(10s, 11s, 1.1)`.

The first range will be: [10s, 20s], which does not respect the maximum.

All subsequent ranges will be: [5.5s, 11s], which does not respect the minimum.

Lol.

alevenberg:
Fixes #11107

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)


google/cloud/internal/backoff_policy.cc
```diff
@@ -42,7 +42,7 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
     generator_ = google::cloud::internal::MakeDefaultPRNG();
   }
   std::uniform_int_distribution<microseconds::rep> rng_distribution(
-      current_delay_range_.count() / 2, current_delay_range_.count());
+      std::max(initial_delay_.count(), current_delay_range_.count() / 2), current_delay_range_.count());
```
coryan, line 45: nit: on some platforms[^1] `max()` is a macro. We protect against macro expansion using `(std::max)(...)`.


[^1]: Windows really, depending on what set of `#include` directives you have, including indirect ones

google/cloud/internal/backoff_policy.h
```diff
@@ -132,6 +132,9 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
         maximum_delay_(std::chrono::duration_cast<std::chrono::microseconds>(
             maximum_delay)),
         scaling_(scaling) {
+    if (current_delay_range_ > maximum_delay_) {
+      current_delay_range_ = maximum_delay_;
+    }
```
devbww, line 137: Could we move this into the initialization (and move "current" down in the member list) with:
```
  current_delay_range_((std::min)(2 * initial_delay_, maximum_delay_))
```

And then (unrelated to this PR, but perhaps addressable here), it seems like just `current_delay_` would be preferable to `current_delay_range_`.

google/cloud/internal/backoff_policy.cc
```diff
@@ -42,7 +42,7 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
     generator_ = google::cloud::internal::MakeDefaultPRNG();
   }
   std::uniform_int_distribution<microseconds::rep> rng_distribution(
-      current_delay_range_.count() / 2, current_delay_range_.count());
+      (std::max)(initial_delay_.count(), current_delay_range_.count() / 2), current_delay_range_.count());
```
devbww, line 45: Could we instead also take care of this during initialization by ensuring "current / 2 >= initial"?

devbww (approved):

dbolduc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @alevenberg and @devbww)

___
*[`google/cloud/internal/backoff_policy.h` line 131 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#-NUwxnSj5_QRoWJ5Z3rx:-NUwxnSj5_QRoWJ5Z3ry:b-855sld) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1e77ece4a9d4916b04de4398aaa8284e5aefd294/google/cloud/internal/backoff_policy.h#L131)):*
> ```objc
>       : initial_delay_(std::chrono::duration_cast<std::chrono::microseconds>(
>             initial_delay)),
>         current_delay_range_(2 * initial_delay_),
> ```

Part of the problem is that we should not be using `2` here. We should be using `scaling`. (We should also use `scaling` in the range calculation).

I should be able to write the following test:

```
TEST(ExponentialBackoffPolicy, RangeDeterminedByScalingFactor) {
  ExponentialBackoffPolicy tested(ms(10), ms(20), 1.1);

  auto delay = tested.OnCompletion();
  EXPECT_LE(ms(10), delay);
  EXPECT_GE(ms(11), delay);
}
```

google/cloud/internal/backoff_policy.cc
alevenberg: Done

google/cloud/internal/backoff_policy.cc
```diff
@@ -42,14 +42,18 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
...
   // Randomized sleep period because it is possible that after some time all
   // client have same sleep period if we use only exponential backoff policy.
   auto delay = microseconds(rng_distribution(*generator_));
-  current_delay_range_ = microseconds(static_cast<microseconds::rep>(
-      static_cast<double>(current_delay_range_.count()) * scaling_));
-  if (current_delay_range_ >= maximum_delay_) {
-    current_delay_range_ = maximum_delay_;
+
+  current_delay_start_ = current_delay_start_ == current_delay_end_
+                             ? initial_delay_
```
devbww, line 51: Q.  Is the specification really to start again at the initial delay?  I don't know, but it seems strange.

google/cloud/internal/backoff_policy.cc
```diff
@@ -42,14 +42,18 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
...
+                             : current_delay_end_;
+  current_delay_end_ = microseconds(static_cast<microseconds::rep>(
+      static_cast<double>(current_delay_end_.count()) * scaling_));
+  if (current_delay_end_ >= maximum_delay_) {
+    current_delay_end_ = maximum_delay_;
```
devbww, line 56: Opt Nit: Perhaps use `std::min()` here?
```
current_delay_end_ = std::min(f(current_delay_end_), maximum_delay_);
```

ci/install-cloud-sdk.sh
```diff
@@ -26,7 +26,11 @@ ARCH="$(uname -m)"
 if [[ "${ARCH}" == "aarch64" ]]; then
   # The tarball uses this name
   ARCH="arm"
+else
+  # For x86_64, uname -p might give unknown
+  ARCH="$(uname -m)"
```
devbww, line 31: Did this sneak back in again?

google/cloud/internal/backoff_policy.h
```diff
@@ -128,9 +128,13 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
                            double scaling)
       : initial_delay_(std::chrono::duration_cast<std::chrono::microseconds>(
             initial_delay)),
-        current_delay_range_(2 * initial_delay_),
         maximum_delay_(std::chrono::duration_cast<std::chrono::microseconds>(
             maximum_delay)),
+        current_delay_start_(initial_delay_),
+        current_delay_end_(
+            (std::min)(std::chrono::duration_cast<std::chrono::microseconds>(
+                           scaling * initial_delay_),
```
devbww, line 136: Nit: I'd suggest moving `scaling_` up in the initialization order, so that we can use it here (instead of the parameter that we're otherwise about to copy).

google/cloud/internal/backoff_policy.cc
devbww: Nit: Notice also how the initialization of `current_delay_end_` (from `initial_delay_`) is different from its re-initialization (from itself).  The latter is careful to use floating-point multiplication, while the former is not.

It **may** make things easier if `current_delay_start_` and `current_delay_end_` were of type `std::chrono::duration<std::micro, double>` instead.  Then we could just say `current_delay_end_ * scaling_`.

google/cloud/internal/backoff_policy.h
alevenberg: Done

google/cloud/internal/backoff_policy.cc
alevenberg: Updated the specification. It isn't to start again at the current delay, but if we have an edge case like

```
  ExponentialBackoffPolicy tested(ms(10), ms(12), 2.0);

  auto delay = tested.OnCompletion();
  EXPECT_LE(ms(10), delay);
  EXPECT_GE(ms(12), delay);
  delay = tested.OnCompletion();
  EXPECT_LE(ms(10), delay);
  EXPECT_GE(ms(12), delay);
```

We don't want it to the delay range to keep growing, we want to the policy to maintain the max and min values

ci/install-cloud-sdk.sh
alevenberg: Yes :(

I think I figured out how git works and have successfully removed the old commit after rebasing 20 times

google/cloud/internal/backoff_policy.cc
alevenberg: I also considered that, but with all of the casting I felt the original intention was obscured away

Alternatively changing the types requires casting when using  casting elsewhere since it is not compatible with `initial_delay_`

alevenberg:
Discussed with @dbolduc 

Will be rewriting to catch some edge cases

alevenberg (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @alevenberg, @dbolduc, and @devbww)

___
*[`google/cloud/internal/backoff_policy.h` line 137 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#gh-1187826967:-NV1OsuE81cYjb-dIS9P:b-ytdmrl) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/995065ed41db9730d3ad9dec1b6e17b07183ec37/google/cloud/internal/backoff_policy.h#L137)):*

Could we move this into the initialization (and move "current" down in the member list) with:
```
  current_delay_range_((std::min)(2 * initial_delay_, maximum_delay_))
```

And then (unrelated to this PR, but perhaps addressable here), it seems like just `current_delay_` would be preferable to `current_delay_range_`.
</blockquote></details>

ack, addressed with the changes
___
*[`google/cloud/internal/backoff_policy.h` line 131 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#-NUwxnSj5_QRoWJ5Z3rx:-NV1EwnPEpLsOrx0ZhfG:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/1e77ece4a9d4916b04de4398aaa8284e5aefd294/google/cloud/internal/backoff_policy.h#L131)):*

Part of the problem is that we should not be using `2` here. We should be using `scaling`. (We should also use `scaling` in the range calculation).

I should be able to write the following test:

```
TEST(ExponentialBackoffPolicy, RangeDeterminedByScalingFactor) {
  ExponentialBackoffPolicy tested(ms(10), ms(20), 1.1);

  auto delay = tested.OnCompletion();
  EXPECT_LE(ms(10), delay);
  EXPECT_GE(ms(11), delay);
}
```

</blockquote></details>

Done
___
*[`google/cloud/internal/backoff_policy.cc` line 45 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#gh-1187814197:-NV1OxDE0fUf-kn50Bt5:bb7f2mf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/995065ed41db9730d3ad9dec1b6e17b07183ec37/google/cloud/internal/backoff_policy.cc#L45)):*

nit: on some platforms[^1] `max()` is a macro. We protect against macro expansion using `(std::max)(...)`.


[^1]: Windows really, depending on what set of `#include` directives you have, including indirect ones
</blockquote></details>

done

alevenberg:
Changes are made. Ready for review.

google/cloud/internal/backoff_policy.cc
devbww: Consider these diffs.  I believe that clarify the intention, and reduce casting.
```
diff --git a/google/cloud/internal/backoff_policy.cc b/google/cloud/internal/backoff_policy.cc
index 61bbe4df56..d4c21acd00 100644
--- a/google/cloud/internal/backoff_policy.cc
+++ b/google/cloud/internal/backoff_policy.cc
@@ -25,9 +25,6 @@ std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
 }
 
 std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
-  using std::chrono::duration_cast;
-  using std::chrono::microseconds;
-  using std::chrono::milliseconds;
   // We do not want to copy the seed in `clone()` because then all operations
   // will have the same sequence of backoffs. Nor do we want to use a shared
   // PRNG because that would require locking and some more complicated lifecycle
@@ -41,21 +38,17 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
   if (!generator_) {
     generator_ = google::cloud::internal::MakeDefaultPRNG();
   }
-  std::uniform_int_distribution<microseconds::rep> rng_distribution(
+  std::uniform_real_distribution<DoubleMicroseconds::rep> rng_distribution(
       current_delay_start_.count(), current_delay_end_.count());
+
   // Randomized sleep period because it is possible that after some time all
   // client have same sleep period if we use only exponential backoff policy.
-  auto delay = microseconds(rng_distribution(*generator_));
+  auto delay = DoubleMicroseconds(rng_distribution(*generator_));
 
-  current_delay_start_ = current_delay_start_ == current_delay_end_
-                             ? initial_delay_
-                             : current_delay_end_;
-  current_delay_end_ = microseconds(static_cast<microseconds::rep>(
-      static_cast<double>(current_delay_end_.count()) * scaling_));
-  if (current_delay_end_ >= maximum_delay_) {
-    current_delay_end_ = maximum_delay_;
-  }
-  return duration_cast<milliseconds>(delay);
+  current_delay_start_ = current_delay_end_;
+  current_delay_end_ = CurrentDelayEnd();
+
+  return std::chrono::duration_cast<std::chrono::milliseconds>(delay);
 }
 
 }  // namespace internal
diff --git a/google/cloud/internal/backoff_policy.h b/google/cloud/internal/backoff_policy.h
index 19b4ddebde..b5721ec7c7 100644
--- a/google/cloud/internal/backoff_policy.h
+++ b/google/cloud/internal/backoff_policy.h
@@ -132,10 +132,7 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
             maximum_delay)),
         scaling_(scaling),
         current_delay_start_(initial_delay_),
-        current_delay_end_(
-            (std::min)(std::chrono::duration_cast<std::chrono::microseconds>(
-                           scaling_ * initial_delay_),
-                       maximum_delay_)) {
+        current_delay_end_(CurrentDelayEnd()) {
     if (scaling_ <= 1.0) {
       google::cloud::internal::ThrowInvalidArgument(
           "scaling factor must be > 1.0");
@@ -157,12 +154,22 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
   std::chrono::milliseconds OnCompletion() override;
 
  private:
+  using DoubleMicroseconds =
+      std::chrono::duration<double, std::chrono::milliseconds::period>;
+
+  DoubleMicroseconds CurrentDelayEnd() {
+    return std::min<DoubleMicroseconds>(current_delay_start_ * scaling_,
+                                        maximum_delay_);
+  }
+
   std::chrono::microseconds initial_delay_;
   std::chrono::microseconds maximum_delay_;
   double scaling_;
+
   // Stores both ends of the current delay range.
-  std::chrono::microseconds current_delay_start_;
-  std::chrono::microseconds current_delay_end_;
+  DoubleMicroseconds current_delay_start_;
+  DoubleMicroseconds current_delay_end_;
+
   absl::optional<DefaultPRNG> generator_;
 };
 
```

google/cloud/internal/backoff_policy.cc
devbww: I'm all for keeping values with the min:max range, but starting back at the initial delay after the start hits the end seems contrary to "backoff".

google/cloud/internal/backoff_policy_test.cc
```diff
@@ -41,6 +41,27 @@ TEST(ExponentialBackoffPolicy, Simple) {
   EXPECT_GE(ms(100), delay);
 }
 
+/// @test Verify the initial and maximum delay are respected.
+TEST(ExponentialBackoffPolicy, RespectMinimumAndMaximumDelay) {
+  ExponentialBackoffPolicy tested(ms(10), ms(12), 2.0);
+
+  auto delay = tested.OnCompletion();
+  EXPECT_LE(ms(10), delay);
```
devbww, line 49: We probably want to `#include "google/cloud/testing_util/chrono_output.h"` here so that test failures are intelligible.

coryan (commented):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r7, 2 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)* status: 4 of 5 files reviewed, 7 unresolved discussions (waiting on @alevenberg, @dbolduc, and @devbww)

___
*[`google/cloud/internal/backoff_policy.cc` line 29 at r9](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#-NV1UbfnA5_MFsxcKgJ1:-NV1UbfnA5_MFsxcKgJ2:b527vf2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/429c41221f6bdd273c5720fd3059b293ef3fdf1a/google/cloud/internal/backoff_policy.cc#L29)):*
> ```cpp
> std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
>   using std::chrono::duration_cast;
>   using std::chrono::microseconds;
> ```

I think you want something like `std::duration<double, std::micro>` instead?
___
*[`google/cloud/internal/backoff_policy.cc` line 50 at r9](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#-NV1UJCI884RAgHU7zeW:-NV1UJCI884RAgHU7zeX:b-dpvnpq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/429c41221f6bdd273c5720fd3059b293ef3fdf1a/google/cloud/internal/backoff_policy.cc#L50)):*
> ```cpp
>   }
> 
>   std::uniform_int_distribution<microseconds::rep> rng_distribution(
> ```

Since you changed `current_delay_start` to be `std::duration<double, ..>` you want this to be `std::uniform_real_distribution<...>`.

alevenberg (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)* status: 4 of 5 files reviewed, 7 unresolved discussions (waiting on @alevenberg, @dbolduc, and @devbww)

___
*[`google/cloud/internal/backoff_policy.cc` line 51 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#gh-1188846183:-NV4y9VZ4QTPV9Zh0SYt:bz11kp9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c7e9e7906c1e301ebf7d6d3e46bd30ffb7141f73/google/cloud/internal/backoff_policy.cc#L51)):*

I'm all for keeping values with the min:max range, but starting back at the initial delay after the start hits the end seems contrary to "backoff".
</blockquote></details>

Ack. Modified the code so it will either return to the initial delay or the max divided by the scaling factor, whichever is higher.

Do you think it is worth throwing an error when we can't "backoff"?
___
*[`google/cloud/internal/backoff_policy.cc` line 56 at r6](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#gh-1188849510:-NV4wsvA1zMy8tpOHxCO:blov4u) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c7e9e7906c1e301ebf7d6d3e46bd30ffb7141f73/google/cloud/internal/backoff_policy.cc#L56)):*

Consider these diffs.  I believe that clarify the intention, and reduce casting.
```
diff --git a/google/cloud/internal/backoff_policy.cc b/google/cloud/internal/backoff_policy.cc
index 61bbe4df56..d4c21acd00 100644
--- a/google/cloud/internal/backoff_policy.cc
+++ b/google/cloud/internal/backoff_policy.cc
@@ -25,9 +25,6 @@ std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
 }
 
 std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
-  using std::chrono::duration_cast;
-  using std::chrono::microseconds;
-  using std::chrono::milliseconds;
   // We do not want to copy the seed in `clone()` because then all operations
   // will have the same sequence of backoffs. Nor do we want to use a shared
   // PRNG because that would require locking and some more complicated lifecycle
@@ -41,21 +38,17 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
   if (!generator_) {
     generator_ = google::cloud::internal::MakeDefaultPRNG();
   }
-  std::uniform_int_distribution<microseconds::rep> rng_distribution(
+  std::uniform_real_distribution<DoubleMicroseconds::rep> rng_distribution(
       current_delay_start_.count(), current_delay_end_.count());
+
   // Randomized sleep period because it is possible that after some time all
   // client have same sleep period if we use only exponential backoff policy.
-  auto delay = microseconds(rng_distribution(*generator_));
+  auto delay = DoubleMicroseconds(rng_distribution(*generator_));
 
-  current_delay_start_ = current_delay_start_ == current_delay_end_
-                             ? initial_delay_
-                             : current_delay_end_;
-  current_delay_end_ = microseconds(static_cast<microseconds::rep>(
-      static_cast<double>(current_delay_end_.count()) * scaling_));
-  if (current_delay_end_ >= maximum_delay_) {
-    current_delay_end_ = maximum_delay_;
-  }
-  return duration_cast<milliseconds>(delay);
+  current_delay_start_ = current_delay_end_;
+  current_delay_end_ = CurrentDelayEnd();
+
+  return std::chrono::duration_cast<std::chrono::milliseconds>(delay);
 }
 
 }  // namespace internal
diff --git a/google/cloud/internal/backoff_policy.h b/google/cloud/internal/backoff_policy.h
index 19b4ddebde..b5721ec7c7 100644
--- a/google/cloud/internal/backoff_policy.h
+++ b/google/cloud/internal/backoff_policy.h
@@ -132,10 +132,7 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
             maximum_delay)),
         scaling_(scaling),
         current_delay_start_(initial_delay_),
-        current_delay_end_(
-            (std::min)(std::chrono::duration_cast<std::chrono::microseconds>(
-                           scaling_ * initial_delay_),
-                       maximum_delay_)) {
+        current_delay_end_(CurrentDelayEnd()) {
     if (scaling_ <= 1.0) {
       google::cloud::internal::ThrowInvalidArgument(
           "scaling factor must be > 1.0");
@@ -157,12 +154,22 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
   std::chrono::milliseconds OnCompletion() override;
 
  private:
+  using DoubleMicroseconds =
+      std::chrono::duration<double, std::chrono::milliseconds::period>;
+
+  DoubleMicroseconds CurrentDelayEnd() {
+    return std::min<DoubleMicroseconds>(current_delay_start_ * scaling_,
+                                        maximum_delay_);
+  }
+
   std::chrono::microseconds initial_delay_;
   std::chrono::microseconds maximum_delay_;
   double scaling_;
+
   // Stores both ends of the current delay range.
-  std::chrono::microseconds current_delay_start_;
-  std::chrono::microseconds current_delay_end_;
+  DoubleMicroseconds current_delay_start_;
+  DoubleMicroseconds current_delay_end_;
+
   absl::optional<DefaultPRNG> generator_;
 };
 
```
</blockquote></details>

Yes agreed the intention is much clearer.
___
*[`google/cloud/internal/backoff_policy.cc` line 29 at r9](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#-NV1UbfnA5_MFsxcKgJ1:-NV4w7d23me3KpYVSEqP:boj9pue) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/429c41221f6bdd273c5720fd3059b293ef3fdf1a/google/cloud/internal/backoff_policy.cc#L29)):*

I think you want something like `std::duration<double, std::micro>` instead?
</blockquote></details>

What is the purpose of the using decl inside the function body?
___
*[`google/cloud/internal/backoff_policy.cc` line 50 at r9](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#-NV1UJCI884RAgHU7zeW:-NV4w8uZ8Sb-vConjty7:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/429c41221f6bdd273c5720fd3059b293ef3fdf1a/google/cloud/internal/backoff_policy.cc#L50)):*

Since you changed `current_delay_start` to be `std::duration<double, ..>` you want this to be `std::uniform_real_distribution<...>`.
</blockquote></details>

Done
___
*[`google/cloud/internal/backoff_policy_test.cc` line 49 at r9](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#gh-1189155713:-NV4wU8hA1ii5MyzEfgc:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/429c41221f6bdd273c5720fd3059b293ef3fdf1a/google/cloud/internal/backoff_policy_test.cc#L49)):*

We probably want to `#include "google/cloud/testing_util/chrono_output.h"` here so that test failures are intelligible.
</blockquote></details>

Done.

coryan (commented):
> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)* status: all files reviewed, 7 unresolved discussions (waiting on @dbolduc and @devbww)

___
*[`google/cloud/internal/backoff_policy.cc` line 29 at r9](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#-NV1UbfnA5_MFsxcKgJ1:-NV52_pKAtXTI4TIUXWH:b28bfdy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/429c41221f6bdd273c5720fd3059b293ef3fdf1a/google/cloud/internal/backoff_policy.cc#L29)):*

What is the purpose of the using decl inside the function body?
</blockquote></details>

Probably just to save typing / make the code easier to grok.  You can reasonably argue that it serves no purpose anymore.  Or even that it never did.  I won't feel bad if you remove it.

google/cloud/internal/backoff_policy.cc
```diff
@@ -41,16 +41,21 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
...
   // client have same sleep period if we use only exponential backoff policy.
   auto delay = microseconds(rng_distribution(*generator_));
-  current_delay_range_ = microseconds(static_cast<microseconds::rep>(
-      static_cast<double>(current_delay_range_.count()) * scaling_));
-  if (current_delay_range_ >= maximum_delay_) {
-    current_delay_range_ = maximum_delay_;
-  }
+
+  current_delay_start_ = current_delay_end_;
+  current_delay_end_ = current_delay_end_ * scaling_;
```
dbolduc, line 57: nit: I might have written

```cc
  current_delay_start_ *= scaling_;
  current_delay_end_ *= scaling_;
```

google/cloud/internal/backoff_policy.h
```diff
@@ -126,12 +126,19 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
...
+        maximum_delay_(
+            std::chrono::duration_cast<
+                std::chrono::duration<double, std::micro>>(maximum_delay)),
+        scaling_(scaling),
+        current_delay_start_(initial_delay_),
+        current_delay_end_((std::min)(
+            std::chrono::duration_cast<
+                std::chrono::duration<double, std::micro>>(scaling_ *
+                                                           initial_delay_),
+            maximum_delay_)) {
```
dbolduc, line 141: nit: now that we do the bounds checking before generating the random number, we don't need to do a min here. We can just say it = `initial_delay_ * scaling_`

google/cloud/internal/backoff_policy.h
```diff
@@ -101,7 +101,7 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
    * @endcode
    *
    * @param initial_delay how long to wait after the first (unsuccessful)
-   *     operation.
+   *     operation and the minimum value for the delay between operations.
```
dbolduc, line 104: optional: rename `initial_delay` to `minimum_delay`.

(It is fine to say: "I will do this in a later PR". FWIW, I think we will want it for #8755)

dbolduc (approved):
> Do you think it is worth throwing an error when we can't "backoff"?

I think the case you are talking about is when we only use the min/max delays and ignore the scaling factor. That is fine, and we should not throw an error which could unnecessarily break existing code.

alevenberg (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)* status: all files reviewed, 9 unresolved discussions (waiting on @dbolduc and @devbww)

___
*[`google/cloud/internal/backoff_policy.h` line 104 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#gh-1190036166:-NV5Hrrk5lFYVF9Hkkx2:box2rhm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/afe1723a61d69187050e0a393f3384bf106e4fc0/google/cloud/internal/backoff_policy.h#L104)):*

optional: rename `initial_delay` to `minimum_delay`.

(It is fine to say: "I will do this in a later PR". FWIW, I think we will want it for #8755)
</blockquote></details>

Ack. Let's wait.
___
*[`google/cloud/internal/backoff_policy.h` line 141 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#gh-1190019048:-NV5I3Vu409nGyLGnqKA:b-k2rwtz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/afe1723a61d69187050e0a393f3384bf106e4fc0/google/cloud/internal/backoff_policy.h#L141)):*

nit: now that we do the bounds checking before generating the random number, we don't need to do a min here. We can just say it = `initial_delay_ * scaling_`
</blockquote></details>

Good catch. Fixed.
___
*[`google/cloud/internal/backoff_policy.cc` line 29 at r9](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#-NV1UbfnA5_MFsxcKgJ1:-NV5I8N2C02MQMY7LqXD:b-kll1oz) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/429c41221f6bdd273c5720fd3059b293ef3fdf1a/google/cloud/internal/backoff_policy.cc#L29)):*

Probably just to save typing / make the code easier to grok.  You can reasonably argue that it serves no purpose anymore.  Or even that it never did.  I won't feel bad if you remove it.
</blockquote></details>

I had never seen a using inside a function body before, I think it is fine to keep. 
___
*[`google/cloud/internal/backoff_policy.cc` line 57 at r10](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#gh-1190015099:-NV5IOHd5T6b3w_1Fuh1:bwrh462) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/afe1723a61d69187050e0a393f3384bf106e4fc0/google/cloud/internal/backoff_policy.cc#L57)):*

nit: I might have written

```cc
  current_delay_start_ *= scaling_;
  current_delay_end_ *= scaling_;
```
</blockquote></details>

Done for the second.

I think since the eventual implementation is going to be setting the start to the initial I'll leave as is for now.

google/cloud/internal/backoff_policy.h
```diff
@@ -134,11 +134,9 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
                 std::chrono::duration<double, std::micro>>(maximum_delay)),
         scaling_(scaling),
         current_delay_start_(initial_delay_),
-        current_delay_end_((std::min)(
-            std::chrono::duration_cast<
-                std::chrono::duration<double, std::micro>>(scaling_ *
-                                                           initial_delay_),
-            maximum_delay_)) {
+        current_delay_end_(std::chrono::duration_cast<
```
dbolduc, line 137: we don't need the `duration_cast`, right?

alevenberg: Good catch. Done

devbww: I don't think you need `duration_cast` on any of those initializations now.

google/cloud/internal/backoff_policy.cc
```diff
@@ -26,7 +26,7 @@ std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
 
 std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
   using std::chrono::duration_cast;
-  using std::chrono::microseconds;
+  using microseconds = std::chrono::duration<double, std::micro>;
```
devbww, line 29: I would proffer that this is more confusing than it's worth ... having `microseconds` and `milliseconds` be different types of things is just confusing.

[In my diff I added a `DoubleMicroseconds` private alias for the former.]

Nit: After that, I'd probably just expand the few uses of the other names, getting rid of the `using`s altogether.

google/cloud/internal/backoff_policy.h
alevenberg: Done

google/cloud/internal/backoff_policy.cc
alevenberg: Changed the using to DoubleMicroseconds and removed the other two

google/cloud/internal/backoff_policy.cc
```diff
@@ -25,9 +25,7 @@ std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
 }
 
 std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
-  using std::chrono::duration_cast;
-  using std::chrono::microseconds;
-  using std::chrono::milliseconds;
+  using DoubleMicroseconds = std::chrono::duration<double, std::micro>;
```
devbww, line 28: Opt: If you make it a class-private alias instead, you can use it for the declarations of the data members too.

alevenberg: Done

devbww (approved):

alevenberg (commented):
> Dismissed @dbolduc and @devbww from 12 discussions.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)* status: 3 of 5 files reviewed, all discussions resolved (waiting on @coryan)

alevenberg (commented):
Ack

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529#-:-NV9iDC6AtJ2BI9uLXGv:b2nhb2c)* status: 3 of 5 files reviewed, all discussions resolved (waiting on @coryan)

alevenberg (commented):
> Reviewed 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alevenberg)

coryan (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11529)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alevenberg)

Issue #11519, opened 2023-05-07
Include `.repo-metadata.json` files in each directory


coryan:
To produce a correct top-level index in the reference documentation we need to include a `.repo-metadata.json` file in each generated directory.

go/yoshi-metadata
go/library-data-integrity

We also need to backfill this data across all current libraries.

This page may be useful when reviewing changes related to this issue: https://cloud.google.com/support/docs/issue-trackers

PR #11555, opened 2023-05-10
feat(generator): generate repository metadata files

coryan:
The documentation index is based on the `.repo-metadata.json` files
committed to our GitHub repository. With this change, the generator
will automatically create and update these files.

We need to get some information from the service config YAML file, so
this change introduces changes to find and parse the file. There are
other cool bits of information on that file, which we can use later.

The generator only creates a `.repo-metadata.json` file if all the
necessary information is found. This can be useful to manually create
the files when the information is missing, as the generator will not
overwrite them.

Part of the work for #11519

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11555)


generator/internal/scaffold_generator.cc
```diff
@@ -138,6 +147,62 @@ std::map<std::string, std::string> ScaffoldVars(
...
+  // convert that to a link to list issues, which is what we want to generate.
+  auto l = vars.find("new_issue_uri");
+  if (l != vars.end()) {
+    auto issue_tracker = l->second;
+    auto const re = std::regex(
+        R"re((https://issuetracker.google.com/issues).*[^a-z]component=([0-9]*).*)re");
+    std::smatch match;
+    if (std::regex_match(issue_tracker, match, re)) {
+      issue_tracker = absl::StrCat(
+          match[1].str(), "?q=componentid:", match[2].str(), " status=open");
```
dbolduc, line 201: `' '` in the link seems weird, but :shrug:

generator/internal/scaffold_generator.cc
```diff
@@ -149,11 +214,77 @@ void MakeDirectory(std::string const& path) {
...
+    google::cloud::cpp::generator::ServiceConfiguration const& service,
+    bool allow_placeholders) {
+  auto const destination = [&]() {
+    MakeDirectory(output_path);
+    auto d = output_path + "/" + LibraryPath(service.product_path());
+    MakeDirectory(d);
+    auto l = vars.find("service_subdirectory");
+    if (l == vars.end()) return d;
+    if (l->second.empty()) return d;
+    d += l->second;
```
dbolduc, line 229: will the subsequent `MakeDirectory(d)` call work if `l->second` is like `foo/v1` ?

generator/internal/scaffold_generator.cc
```diff
@@ -149,11 +214,77 @@ void MakeDirectory(std::string const& path) {
...
+    d += "/";
```
dbolduc, line 230: I don't think this is necessary. `service_subdirectory` should always end in a slash.

https://github.com/googleapis/google-cloud-cpp/blob/c435bc9ac0cc6dc26833164a3f28e82c111eced4/generator/internal/scaffold_generator.cc#L82

dbolduc (approved):

dbolduc (approved):

generator/internal/scaffold_generator.cc
coryan: Removed.

generator/internal/scaffold_generator.cc
coryan: Don't know, but that is what the rest of the scaffold does.

generator/internal/scaffold_generator.cc
coryan: Fixed

generator/generator_config.proto
```diff
@@ -144,6 +144,10 @@ message ServiceConfiguration {
   // This need not be explicitly set to true as the generator will check if the
   // field has been set, and assign a value of true if not set.
   optional bool generate_grpc_transport = 22;
+
+  // If set to `true` the generator will *NOT* create a .repo-metadata.json
+  // file. This is useful when multiple services target the same directory.
+  bool omit_metadata = 23;
```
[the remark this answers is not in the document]
coryan: Fixed

PR #11560, opened 2023-05-10
impl(bigquery): Table connection and client classes

jsrinnn:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11560)


jsrinnn (commented):
Thanks Carlos!

The client classes are helkpful as I'll need them later when I am using the api in the external interface. I'll probably be using the same code from here. Same for Mock classes.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11560#-:-NV5vh3Z-PVyoEWn7C1E:b-4qwppl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jsrinnn)

PR #11576, opened 2023-05-12
impl(common): support extracting a typed option from an Options

devbww:
```
absl::optional<typename T::Type> internal::ExtractOption<T>(Options&);
```
Extracts and returns the value for `T` from the options, or nullopt if `T` was not set.

This is intended for code paths that want to consume an option, for example in a client call that passes the option's value via conventional mechanisms, rather than through an `OptionsSpan`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11576)

### DIFF
--- a/google/cloud/options.h
+++ b/google/cloud/options.h
@@ -18,6 +18,7 @@
 #include "google/cloud/internal/type_list.h"
 #include "google/cloud/version.h"
 #include "absl/base/attributes.h"
+#include "absl/types/optional.h"
 #include <memory>
 #include <set>
 #include <typeindex>
@@ -35,6 +36,8 @@ void CheckExpectedOptionsImpl(std::set<std::type_index> const&, Options const&,
                               char const*);
 // TODO(#8800) - Remove when bigtable::Table no longer uses bigtable::DataClient
 bool IsEmpty(Options const&);
+template <typename T>
+absl::optional<typename T::Type> ExtractOption(Options&);
 template <typename T>
 inline T const& DefaultValue() {
   static auto const* const kDefaultValue = new T{};
@@ -214,6 +217,8 @@ class Options {
   friend void internal::CheckExpectedOptionsImpl(
       std::set<std::type_index> const&, Options const&, char const*);
   friend bool internal::IsEmpty(Options const&);
+  template <typename T>
+  friend absl::optional<typename T::Type> internal::ExtractOption(Options&);
 
   // The type-erased data holder of all the option values.
   class DataHolder {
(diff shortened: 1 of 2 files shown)


google/cloud/options_test.cc
```diff
@@ -270,6 +270,30 @@ TEST(MergeOptions, Basics) {
...
+  EXPECT_TRUE(opts.has<IntOption>());
+  EXPECT_TRUE(opts.has<StringOption>());
+  EXPECT_FALSE(b.has_value());
+
+  auto i = internal::ExtractOption<IntOption>(opts);
+  EXPECT_FALSE(opts.has<BoolOption>());
+  EXPECT_FALSE(opts.has<IntOption>());
+  EXPECT_TRUE(opts.has<StringOption>());
+  ASSERT_TRUE(i.has_value());
+  EXPECT_EQ(*i, 42);
```
dbolduc, line 287: nit: `EXPECT_THAT(i, Optional(42));`

google/cloud/options_test.cc
```diff
@@ -270,6 +270,30 @@ TEST(MergeOptions, Basics) {
...
+
+  auto s = internal::ExtractOption<StringOption>(opts);
+  EXPECT_FALSE(opts.has<BoolOption>());
+  EXPECT_FALSE(opts.has<IntOption>());
+  EXPECT_FALSE(opts.has<StringOption>());
+  ASSERT_TRUE(s.has_value());
+  EXPECT_EQ(*s, "foo");
```
dbolduc, line 294: nit: `EXPECT_THAT(i, Optional("foo"));`

google/cloud/options.h
```diff
@@ -320,6 +325,22 @@ void CheckExpectedOptions(Options const& opts, char const* caller) {
  */
 Options MergeOptions(Options preferred, Options alternatives);
 
+/**
+ * Extracts and returns the value for `T` from @p opts, or nullopt if `T` was
+ * not set.  This is intended for code paths that want to consume an option,
+ * for example, a client call that passes the option's value via conventional
+ * mechanisms, rather than though an OptionsSpan.
```
dbolduc, line 332: s/though/through/

dbolduc (approved):

google/cloud/options_test.cc
devbww: Done.

google/cloud/options_test.cc
devbww: Done.

google/cloud/options.h
devbww: Done.

PR #11595, opened 2023-05-13
chore(b*): backfill `.repo-metadata.json` files

coryan:
Part of the work for #11519

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11595)

### DIFF
--- a/google/cloud/batch/v1/.repo-metadata.json
+++ b/google/cloud/batch/v1/.repo-metadata.json
@@ -1,0 +1,14 @@
+{
+    "api_id": "batch.googleapis.com",
+    "api_shortname": "batch",
+    "client_documentation": "https://cloud.google.com/cpp/reference/batch/latest",
+    "distribution_name": "google-cloud-cpp",
+    "issue_tracker": "",
+    "language": "cpp",
+    "library_type": "GAPIC_AUTO",
+    "name_pretty": "Batch API",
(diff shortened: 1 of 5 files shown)


google/cloud/batch/v1/.repo-metadata.json
```diff
@@ -0,0 +1,14 @@
+{
+    "api_id": "batch.googleapis.com",
+    "api_shortname": "batch",
+    "client_documentation": "https://cloud.google.com/cpp/reference/batch/latest",
+    "distribution_name": "google-cloud-cpp",
+    "issue_tracker": "",
```
dbolduc, line 6: remove?

dbolduc (approved):
are there specific fields we should be looking at in the review? e.g. which ones are done manually?

google/cloud/batch/v1/.repo-metadata.json
coryan: I was able to find it by browsing the components (the documentation has nothing about support, sigh)

coryan:
> are there specific fields we should be looking at in the review? e.g. which ones are done manually?

The issue tracker was all manual.  The "name_pretty" was based on our READMEs and may be wrong.  Ditto for the "product_documentation" field.

Issue #10031, opened 2022-10-11
Downloading GCS cache stalls


coryan:
Seems like we need to configure a timeout for these downloads.  It should complete in around 60 seconds when things are working well.  A timeout of 5x would not be unreasonable.

```
2022-10-11T14:13:39Z (+0s): Running:
env -C '/h' /workspace/ci/cloudbuild/cache.sh restore --bucket_url=gs://cloud-cpp-testing-resources_cloudbuild/build-cache/google-cloud-cpp --key=main/fedora-36-quickstart-production/h --fallback_key=main/fedora-36-quickstart-production/h
2022-10-11T14:13:39Z (+0s): ====> cache.sh: restore
gs://cloud-cpp-testing-resources_cloudbuild/build-cache/google-cloud-cpp/main/fedora-36-quickstart-production/h/cache.tar.gz:
    Creation time:          Tue, 11 Oct 2022 13:27:54 GMT
    Update time:            Tue, 11 Oct 2022 13:27:54 GMT
    Storage class:          STANDARD
    Content-Length:         4585769035
    Content-Type:           application/x-tar
    Component-Count:        32
    Hash (crc32c):          aYEg+g==
    ETag:                   CJDK4PKj2PoCEAE=
    Generation:             1665494874072336
    Metageneration:         1
2022-10-11T14:13:41Z (+2s): Fetching cache url gs://cloud-cpp-testing-resources_cloudbuild/build-cache/google-cloud-cpp/main/fedora-36-quickstart-production/h/cache.tar.gz
Copying gs://cloud-cpp-testing-resources_cloudbuild/build-cache/google-cloud-cpp/main/fedora-36-quickstart-production/h/cache.tar.gz to file:///tmp/tmp.S2dl2lyXhQ/cache.tar.gz
```

This failed before we created public logs:

https://pantheon.corp.google.com/cloud-build/builds;region=us-east1/a148579a-ec23-4055-89cd-0eed715a67eb;step=2?mods=logs_tg_prod&project=cloud-cpp-testing-resources

dbolduc:
https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/10319/85926ad424b01ce5c82363a575e479fa6afa37ef/demo-fedora-demo-install/log-d7f6e486-1332-4bba-8a1b-7bd51109d1da.txt

coryan:
[...]

PR #11604, opened 2023-05-14
ci: timeout Linux GCS cache download

dbolduc:
Motivated by #10031

I think this command takes ~4s normally. So 2m should be generous enough. Anything seems better than this command struggling for 3.5 hours.

The comments say this is not necessary that downloading the cache is not necessary for running the build, so we just continue on timeout.

We could implement a retry loop (with timeouts) instead.... thoughts?

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11604)

### DIFF
--- a/ci/cloudbuild/cache.sh
+++ b/ci/cloudbuild/cache.sh
@@ -138,7 +138,7 @@ function restore_cache() {
   for url in "${urls[@]}"; do
     if gsutil stat "${url}"; then
       io::log "Fetching cache url ${url}"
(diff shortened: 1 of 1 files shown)


coryan (approved):

coryan:
> We could implement a retry loop (with timeouts) instead.... thoughts?

I believe it is a good idea, but I would not be surprised that in practice such retries did not work so well.  The failures must be pretty bad for `gcloud storage` to freeze.

dbolduc:
I will just merge this, because I forgot what I was doing.

PR #11633, opened 2023-05-16
feat(storage): lazy allocation for upload buffer

coryan:
Allocating large upload buffers upfront can be slow if the application is going to send a small amount of data.  Better grow the buffer as needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11633)

### DIFF
--- a/google/cloud/storage/internal/object_write_streambuf.cc
+++ b/google/cloud/storage/internal/object_write_streambuf.cc
@@ -105,7 +105,7 @@ std::streamsize ObjectWriteStreambuf::xsputn(char const* s,
   auto const actual_size = put_area_size();
   // One of the invariants in this class is that actual_size is always less than
   // max_buffer_size_. Using max_buffer_size_ - actual size avoids overflow.
-  if (count < max_buffer_size_ - actual_size) {
+  if (static_cast<std::size_t>(count) < max_buffer_size_ - actual_size) {
     current_ios_buffer_.insert(current_ios_buffer_.end(), s, s + count);
     UpdatePutArea();
     return count;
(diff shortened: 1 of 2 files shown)


google/cloud/storage/internal/object_write_streambuf.cc
```diff
@@ -107,20 +103,20 @@ std::streamsize ObjectWriteStreambuf::xsputn(char const* s,
...
-    if (actual_size == 0) {
-      FlushRoundChunk({ConstBuffer(s, static_cast<std::size_t>(count))});
-    } else {
-      FlushRoundChunk({
-          ConstBuffer(pbase(), actual_size),
-          ConstBuffer(s, static_cast<std::size_t>(count)),
-      });
-    }
-    if (!last_status_.ok()) return traits_type::eof();
+  if (count + actual_size < max_buffer_size_) {
```
devbww, line 106: Q. Is there any reason to rewrite the test to avoid any possibility of overflow?

For example, if we know `actual_size <= max_buffer_size_`, then we might instead say...
```
  if (count < max_buffer_size_ - actual_size) {
```
Aside: Should the test be `<=`?

devbww (approved):

google/cloud/storage/internal/object_write_streambuf.cc
coryan: > Q. Is there any reason to rewrite the test to avoid any possibility of overflow?
> 
> For example, if we know `actual_size <= max_buffer_size_`, then we might instead say...
> 
> ```
>   if (count < max_buffer_size_ - actual_size) {
> ```

Thanks, done.

> 
> Aside: Should the test be `<=`?

I think it should be `<`: we want to flush the data when `count == max_buffer_size - actual_size` because at that point the buffer is large enough to justify the cost of said flush.

devbww: I see.  It's not "is there space", but rather "is there space plus some".  Thanks.

PR #11641, opened 2023-05-17
docs(spanner): some documentation tweaks

coryan:
Fixed some typos in the landing page: `THis` vs. `This` and `C++11` when we only support C++14.  With DocFX we don't have deep linking to sections of pages working, the page is small, so just link the page for now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11641)

### DIFF
--- a/google/cloud/spanner/doc/spanner-main.dox
+++ b/google/cloud/spanner/doc/spanner-main.dox
@@ -3,7 +3,7 @@
 @mainpage Cloud Spanner C++ Client Library
 
 The Cloud Spanner C++ Client library offers types and functions to use Cloud
-Spanner from C++11 applications.
+Spanner from C++ applications.
 
 @tableofcontents{HTML:2}
 
@@ -25,9 +25,9 @@ into your project.
 - [Client::Read()](@ref google::cloud::spanner::Client::Read()) - to read the
   rows in a table.
 - @ref spanner-error-handling to learn how the library reports run-time errors.
-- @ref spanner-env-logging enable logging. THis can be an effective approach to
-  diagnose many problems.
 - @ref spanner-env for environment variables affecting the library.
+- @ref spanner-env includes environment variables to enable logging. This can be
+  an effective approach to diagnose many problems.
 - @ref spanner-endpoint-example
 - @ref spanner-auth-example
 - @ref spanner-retry-example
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/doc/spanner-main.dox
```diff
@@ -25,9 +25,9 @@ into your project.
 - [Client::Read()](@ref google::cloud::spanner::Client::Read()) - to read the
```
devbww, line 25: `s/ -//` to match the form of the previous two entries.

google/cloud/spanner/doc/spanner-main.dox
```diff
@@ -25,9 +25,9 @@ into your project.
...
   rows in a table.
 - @ref spanner-error-handling to learn how the library reports run-time errors.
-- @ref spanner-env-logging enable logging. THis can be an effective approach to
-  diagnose many problems.
 - @ref spanner-env for environment variables affecting the library.
```
devbww, line 28: This link leads to a page where a `:cloud:` becomes an image.

google/cloud/spanner/doc/spanner-main.dox
```diff
@@ -25,9 +25,9 @@ into your project.
...
+- @ref spanner-env includes environment variables to enable logging. This can be
```
devbww, line 29: s/spanner-env/spanner-env-logging/?  Although that seemed to lead nowhere.

Perhaps it is just that the two `spanner-env` entries should be combined?

google/cloud/spanner/doc/spanner-main.dox
```diff
@@ -25,9 +25,9 @@ into your project.
...
+  an effective approach to diagnose many problems.
 - @ref spanner-endpoint-example
 - @ref spanner-auth-example
```
devbww, line 32: This leads to a page where two links are not separated by any space.

devbww (approved):

google/cloud/spanner/doc/spanner-main.dox
coryan: Should be fixed in #11642

google/cloud/spanner/doc/spanner-main.dox
coryan: Fixed

google/cloud/spanner/doc/spanner-main.dox
coryan: `spanner-env-logging` is not working with the DocFX-based docs, but works with Doxygen.  Merged the two entries.

google/cloud/spanner/doc/spanner-main.dox
coryan: Fixed.  I used the same structure as the generated documents, if you want we can make it nicer as this is a hand-crafted library.

PR #11648, opened 2023-05-18
impl(docfx): support odd mocks in pubsub library

coryan:
The Pub/Sub library has a few mocks that don't mock anything. We kept the `MOCK_METHOD(..., ack_id...)` definitions for backwards compatibility, but the base class has no `ack_id()` to mock. Skip these in the documentation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11648)

### DIFF
--- a/docfx/yaml_context.cc
+++ b/docfx/yaml_context.cc
@@ -97,4 +97,21 @@ YamlContext NestedYamlContext(YamlContext const& ctx, pugi::xml_node node) {
   return nested;
 }
 
+bool IsSkippedChild(YamlContext const& ctx, pugi::xml_node node) {
+  // Mocked functions are not children.
+  auto id = std::string{node.attribute("id").as_string()};
+  if (ctx.mocked_ids.count(id) != 0) return true;
+
+  // Things that are not MOCK_METHOD() are always present
+  auto qname = std::string_view{node.child("qualifiedname").child_value()};
+  auto const p = qname.find("::MOCK_METHOD");
+  if (p == std::string_view::npos) return false;
+
+  // In a few places we kept a MOCK_METHOD() definition for a function that
+  // does not exist in the base class.  These only exist for backwards
+  // compatibility. Skip them is no need to document those.
+  auto const m = ctx.mocking_functions_by_id.find(id);
+  return m == ctx.mocking_functions_by_id.end();
+}
+
 }  // namespace docfx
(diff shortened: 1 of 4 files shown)


docfx/yaml_context.cc
```diff
@@ -97,4 +97,21 @@ YamlContext NestedYamlContext(YamlContext const& ctx, pugi::xml_node node) {
...
+  if (ctx.mocked_ids.count(id) != 0) return true;
+
+  // Things that are not MOCK_METHOD() are always present
+  auto qname = std::string_view{node.child("qualifiedname").child_value()};
+  auto const p = qname.find("::MOCK_METHOD");
+  if (p == std::string_view::npos) return false;
+
+  // In a few places we kept a MOCK_METHOD() definition for a function that
+  // does not exist in the base class.  These only exist for backwards
+  // compatibility. Skip them is no need to document those.
```
dbolduc, line 112: Skip them as there is no need

dbolduc (approved):

alevenberg (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11648)* status: all files reviewed, 3 unresolved discussions

___
*[`docfx/yaml_context.cc` line 105 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11648#-NVj90yg75YZmvwosD0n:-NVj90yg75YZmvwosD0o:bx22tqx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/cf107ee678463583af5e56b5b255e91e4b3ef3ce/docfx/yaml_context.cc#L105)):*
> ```cpp
>   if (ctx.mocked_ids.count(id) != 0) return true;
> 
>   // Things that are not MOCK_METHOD() are always present
> ```

nit: add a period
___
*[`docfx/yaml_context.cc` line 111 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11648#-NVj8lmgEaZFngGA9Lr4:-NVj8lmh3IXaAkwWowq-:b-rhuq1w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/cf107ee678463583af5e56b5b255e91e4b3ef3ce/docfx/yaml_context.cc#L111)):*
> ```cpp
> 
>   // In a few places we kept a MOCK_METHOD() definition for a function that
>   // does not exist in the base class.  These only exist for backwards
> ```

nit: extra space

PR #11649, opened 2023-05-18
fix: workaround curl_multi_poll returning an error on EINTR

### LINKED ISSUE #11647: curl_multi_wait might erroneously return CURLM_UNRECOVERABLE_POLL
In https://github.com/curl/curl/issues/11135 it was determined that `curl_multi_wait` could sometimes return `CURLM_UNRECOVERABLE_POLL` on EINTR, while in most cases it turns EINTR into a success.

It is a plausible cause for a sporadic error that a user [reported](https://github.com/apache/arrow/issues/35318) when using GCS from Arrow C++:
`google::cloud::Status(UNKNOWN: Permanent error ReadObjectNotWrapped: WaitForHandles(): unexpected error code in curl_multi_*, [12]=Unrecoverable error in select/poll)`

pitrou:
Fixes #11647.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11649)

### DIFF
--- a/google/cloud/internal/curl_impl.cc
+++ b/google/cloud/internal/curl_impl.cc
@@ -705,15 +705,28 @@ Status CurlImpl::PerformWorkUntil(absl::FunctionRef<bool()> predicate) {
 Status CurlImpl::WaitForHandles(int& repeats) {
   int const timeout_ms = 1000;
   int numfds = 0;
+  CURLMcode result;
 #if CURL_AT_LEAST_VERSION(7, 66, 0)
-  CURLMcode result =
-      curl_multi_poll(multi_.get(), nullptr, 0, timeout_ms, &numfds);
+  // Use curl_multi_poll()
+#if CURL_AT_LEAST_VERSION(7, 84, 0)
+  // Work around https://github.com/curl/curl/issues/11135 by retrying
+  // when EINTR fails poll(). We want to retry so that post-poll()
+  // curl_multi bookkeeping code runs.
+  do {
+    errno = 0;
+    result =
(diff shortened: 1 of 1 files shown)


pitrou:
For now this implements a workaround in `CurlImpl::WaitForHandles`. Once we agree on that we have to port the workaround to `CurlDownloadRequest::WaitForHandles` (or factor out the logic into a shared helper routine).

@coryan

pitrou:
I'll note that I can't test whether this actually fixes the underlying issue. At least it doesn't seem to break the test suite on my machine.

pitrou:
By the way, this fix assumes we want to retry on EINTR, which seems to make sense but is not the case currently.
In any case, it seems `PerformWorkUntil` loops if `WaitForHandles` returns with success. Thoughts?

coryan:
Thanks very much for the PR. 

> For now this implements a workaround in `CurlImpl::WaitForHandles`. Once we agree on that, I'll have to port the workaround to `CurlDownloadRequest::WaitForHandles` (or factor out the logic into a shared helper routine).

 In all likelihood we will remove that code in a couple of months.  It is there just as an emergency hatch in case the new code does not work for some people, and so far it looks like it is not needed.

The code looks good to me, though it is a bit sad to have so many `#if` branches.


Let's run the build so see if the robots have something to say.

coryan:
The failure with `checkers-pr` is just formatting, you can see what the robots want at the end of:

https://github.com/googleapis/google-cloud-cpp/pull/11649/checks?check_run_id=13593517862

coryan (approved):

PR #11650, opened 2023-05-18
feat: add new constructor for exponential backoff policy 

alevenberg:
Add a new constructor to enable a full jitter policy implementation by providing a minimum delay and initial delay range (part of work for #8755)

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650)

### DIFF
--- a/google/cloud/internal/backoff_policy.cc
+++ b/google/cloud/internal/backoff_policy.cc
@@ -40,14 +40,11 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
     generator_ = google::cloud::internal::MakeDefaultPRNG();
   }
 
-  if (current_delay_end_ >= maximum_delay_) {
-    // If the scaling lower bound does not grow, do not increase it.
-    if (scaling_lower_bound_ != 1) {
-      current_delay_start_ =
-          (std::max)(minimum_delay_, current_delay_end_ / scaling_upper_bound_);
-    }
-    current_delay_end_ = (std::min)(current_delay_end_, maximum_delay_);
+  if (current_delay_start_ >= (maximum_delay_ / scaling_upper_bound_)) {
+    current_delay_start_ =
+        (std::max)(minimum_delay_, maximum_delay_ / scaling_upper_bound_);
   }
+  current_delay_end_ = (std::min)(current_delay_end_, maximum_delay_);
 
   std::uniform_real_distribution<DoubleMicroseconds::rep> rng_distribution(
       current_delay_start_.count(), current_delay_end_.count());
--- a/google/cloud/internal/backoff_policy.cc
+++ b/google/cloud/internal/backoff_policy.cc
@@ -21,7 +21,7 @@ namespace internal {
 
 std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
   return std::make_unique<ExponentialBackoffPolicy>(
-      initial_delay_upper_bound_, minimum_delay_, maximum_delay_, scaling_);
+      minimum_delay_, initial_delay_upper_bound_, maximum_delay_, scaling_);
 }
 
 std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
--- a/google/cloud/internal/backoff_policy.cc
+++ b/google/cloud/internal/backoff_policy.cc
@@ -21,7 +21,8 @@ namespace internal {
 
 std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
   return std::make_unique<ExponentialBackoffPolicy>(
-      minimum_delay_, initial_delay_upper_bound_, maximum_delay_, scaling_);
+      minimum_delay_, initial_delay_upper_bound_, maximum_delay_,
+      scaling_lower_bound_, scaling_upper_bound_);
 }
 
 std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
@@ -39,15 +40,22 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
     generator_ = google::cloud::internal::MakeDefaultPRNG();
   }
 
-  current_delay_end_ = (std::min)(current_delay_end_, maximum_delay_);
+  if (current_delay_end_ >= maximum_delay_) {
+    if (scaling_lower_bound_ != 1) {
+      current_delay_start_ =
+          (std::max)(minimum_delay_, current_delay_end_ / scaling_lower_bound_);
+    }
+    current_delay_end_ = (std::min)(current_delay_end_, maximum_delay_);
+  }
 
   std::uniform_real_distribution<DoubleMicroseconds::rep> rng_distribution(
-      minimum_delay_.count(), current_delay_end_.count());
+      current_delay_start_.count(), current_delay_end_.count());
   // Randomized sleep period because it is possible that after some time all
   // client have same sleep period if we use only exponential backoff policy.
   auto delay = DoubleMicroseconds(rng_distribution(*generator_));
 
-  current_delay_end_ *= scaling_;
+  current_delay_start_ *= scaling_lower_bound_;
+  current_delay_end_ *= scaling_upper_bound_;
 
   return std::chrono::duration_cast<std::chrono::milliseconds>(delay);
 }
--- a/google/cloud/internal/backoff_policy.cc
+++ b/google/cloud/internal/backoff_policy.cc
@@ -21,7 +21,7 @@ namespace internal {
 
 std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
   return std::make_unique<ExponentialBackoffPolicy>(
-      initial_delay_, minimum_delay_, maximum_delay_, scaling_);
+      initial_delay_upper_bound_, minimum_delay_, maximum_delay_, scaling_);
 }
 
 std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
@@ -39,17 +39,15 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
     generator_ = google::cloud::internal::MakeDefaultPRNG();
   }
 
+  current_delay_end_ = (std::min)(current_delay_end_, maximum_delay_);
+
   std::uniform_real_distribution<DoubleMicroseconds::rep> rng_distribution(
-      current_delay_start_.count(), current_delay_end_.count());
+      minimum_delay_.count(), current_delay_end_.count());
   // Randomized sleep period because it is possible that after some time all
   // client have same sleep period if we use only exponential backoff policy.
   auto delay = DoubleMicroseconds(rng_distribution(*generator_));
 
-  if (current_delay_end_ < maximum_delay_) {
-    current_delay_range_ *= scaling_;
-    current_delay_end_ =
-        (std::min)(minimum_delay_ + current_delay_range_, maximum_delay_);
-  }
+  current_delay_end_ *= scaling_;
 
   return std::chrono::duration_cast<std::chrono::milliseconds>(delay);
 }
--- a/google/cloud/internal/backoff_policy.cc
+++ b/google/cloud/internal/backoff_policy.cc
@@ -41,9 +41,10 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
   }
 
   if (current_delay_end_ >= maximum_delay_) {
+    // If the scaling lower bound does not grow, do not increase it.
     if (scaling_lower_bound_ != 1) {
       current_delay_start_ =
-          (std::max)(minimum_delay_, current_delay_end_ / scaling_lower_bound_);
+          (std::max)(minimum_delay_, current_delay_end_ / scaling_upper_bound_);
     }
     current_delay_end_ = (std::min)(current_delay_end_, maximum_delay_);
   }
--- a/google/cloud/internal/backoff_policy.cc
+++ b/google/cloud/internal/backoff_policy.cc
@@ -20,8 +20,8 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
 
 std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
-  return std::make_unique<ExponentialBackoffPolicy>(initial_delay_,
-                                                    maximum_delay_, scaling_);
+  return std::make_unique<ExponentialBackoffPolicy>(
+      initial_delay_range_, minimum_delay_, maximum_delay_, scaling_);
 }
 
 std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
@@ -39,19 +39,17 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
     generator_ = google::cloud::internal::MakeDefaultPRNG();
   }
 
-  if (current_delay_end_ >= maximum_delay_) {
-    current_delay_start_ = std::max(maximum_delay_ / scaling_, initial_delay_);
-    current_delay_end_ = maximum_delay_;
-  }
-
   std::uniform_real_distribution<DoubleMicroseconds::rep> rng_distribution(
       current_delay_start_.count(), current_delay_end_.count());
   // Randomized sleep period because it is possible that after some time all
   // client have same sleep period if we use only exponential backoff policy.
   auto delay = DoubleMicroseconds(rng_distribution(*generator_));
 
-  current_delay_start_ = current_delay_end_;
-  current_delay_end_ *= scaling_;
+  if (current_delay_end_ < maximum_delay_) {
+    current_delay_range_ *= scaling_;
+    current_delay_end_ =
+        (std::min)(minimum_delay_ + current_delay_range_, maximum_delay_);
+  }
 
   return std::chrono::duration_cast<std::chrono::milliseconds>(delay);
 }
--- a/google/cloud/internal/backoff_policy.cc
+++ b/google/cloud/internal/backoff_policy.cc
@@ -21,7 +21,7 @@ namespace internal {
 
 std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
   return std::make_unique<ExponentialBackoffPolicy>(
-      initial_delay_range_, minimum_delay_, maximum_delay_, scaling_);
+      initial_delay_, minimum_delay_, maximum_delay_, scaling_);
 }
 
 std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
--- a/google/cloud/internal/backoff_policy.h
+++ b/google/cloud/internal/backoff_policy.h
@@ -181,8 +181,8 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
   template <typename Rep1, typename Period1, typename Rep2, typename Period2,
             typename Rep3, typename Period3>
   ExponentialBackoffPolicy(
-      std::chrono::duration<Rep1, Period1> initial_delay_range,
-      std::chrono::duration<Rep2, Period2> minimum_delay,
+      std::chrono::duration<Rep1, Period1> minimum_delay,
+      std::chrono::duration<Rep2, Period2> initial_delay_upper_bound,
       std::chrono::duration<Rep3, Period3> maximum_delay, double scaling)
       : minimum_delay_(minimum_delay),
         initial_delay_upper_bound_(initial_delay_upper_bound),
--- a/google/cloud/internal/backoff_policy.h
+++ b/google/cloud/internal/backoff_policy.h
@@ -148,7 +148,7 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
    * auto r2 = ExponentialBackoffPolicy<S,T>(0min, 10min, 10min + 2s, 1.002);
    * @endcode
    *
-   * @param initial_delay_range how long to wait after the first (unsuccessful)
+   * @param initial_delay how long to wait after the first (unsuccessful)
    *     operation.
    * @param minimum_delay the minimum value for the delay between operations.
    * @param maximum_delay the maximum value for the delay between operations.
@@ -175,15 +175,15 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
    */
   template <typename Rep1, typename Period1, typename Rep2, typename Period2,
             typename Rep3, typename Period3>
-  ExponentialBackoffPolicy(
-      std::chrono::duration<Rep1, Period1> initial_delay_range,
+  ExponentialBackoffPolicy(std::chrono::duration<Rep1, Period1> initial_delay,
                            std::chrono::duration<Rep2, Period2> minimum_delay,
-      std::chrono::duration<Rep3, Period3> maximum_delay, double scaling)
-      : initial_delay_range_(initial_delay_range),
+                           std::chrono::duration<Rep3, Period3> maximum_delay,
+                           double scaling)
+      : initial_delay_(initial_delay),
         minimum_delay_(minimum_delay),
         maximum_delay_(maximum_delay),
         scaling_(scaling),
-        current_delay_range_(initial_delay_range_),
+        current_delay_range_(initial_delay_),
         current_delay_start_(minimum_delay_),
         current_delay_end_(
             (std::min)(minimum_delay_ + current_delay_range_, maximum_delay_)) {
@@ -198,7 +198,7 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
   //    know specifically which one is at fault)
   //  - We want uncorrelated data streams for each copy anyway.
   ExponentialBackoffPolicy(ExponentialBackoffPolicy const& rhs) noexcept
-      : initial_delay_range_(rhs.initial_delay_range_),
+      : initial_delay_(rhs.initial_delay_),
         minimum_delay_(rhs.minimum_delay_),
         maximum_delay_(rhs.maximum_delay_),
         scaling_(rhs.scaling_),
@@ -211,7 +211,7 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
 
  private:
   using DoubleMicroseconds = std::chrono::duration<double, std::micro>;
-  DoubleMicroseconds initial_delay_range_;
+  DoubleMicroseconds initial_delay_;
   DoubleMicroseconds minimum_delay_;
   DoubleMicroseconds maximum_delay_;
   double scaling_;
--- a/google/cloud/internal/backoff_policy.h
+++ b/google/cloud/internal/backoff_policy.h
@@ -103,7 +103,7 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
    * auto r2 = ExponentialBackoffPolicy<S,T>(10min, 10min + 2s, 1.002);
    * @endcode
    *
-   * @param initial_delay how long to wait after the first (unsuccessful)
+   * @param minimum_delay how long to wait after the first (unsuccessful)
    *     operation and the minimum value for the delay between operations.
    * @param maximum_delay the maximum value for the delay between operations.
    * @param scaling how fast does the delay increase between iterations.
@@ -130,8 +130,9 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
                            std::chrono::duration<Rep2, Period2> maximum_delay,
                            double scaling)
       : ExponentialBackoffPolicy(
+            minimum_delay,
             /*initial_delay_upper_bound=*/minimum_delay * scaling,
-            minimum_delay, maximum_delay, scaling) {}
+            maximum_delay, scaling) {}
 
   /**
    * Constructor for an exponential backoff policy with an initial delay upper
@@ -148,24 +149,27 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
    * auto r1 = ExponentialBackoffPolicy<S,T>(0ms, 10ms, 500ms, 1.618);
    * @endcode
    *
-   * @param initial_delay_upper_bound how long to wait after the first
-   *     (unsuccessful) operation.
+   * @param minimum_delay the minimum value for the delay between operations.
+   * @param initial_delay_upper_bound the longest possible delay to wait after
+   *     the first (unsuccessful) operation.
    * @param minimum_delay the minimum value for the delay between operations.
    * @param maximum_delay the maximum value for the delay between operations.
    * @param scaling how fast does the delay increase between iterations.
    *
    * @tparam Rep1 a placeholder to match the Rep tparam for
-   *     @p initial_delay_upper_bound's type. The semantics of this template
+   *     @p minimum_delay's type. The semantics of this template
    *     parameter are documented in `std::chrono::duration<>` (in brief, the
    *     underlying arithmetic type used to store the number of ticks). For our
    *     purposes, it is simply a formal parameter.
    * @tparam Period1 a placeholder to match the Period tparam for
-   *     @p initial_delay_upper_bound's type. The semantics of this template
+   *     @p minimum_delay's type. The semantics of this template
    *     parameter are documented in `std::chrono::duration<>` (in brief, the
    *     underlying arithmetic type used to store the number of ticks). For our
    *     purposes, it is simply a formal parameter.
-   * @tparam Rep2 similar formal parameter for the type of @p minimum_delay.
-   * @tparam Period2 similar formal parameter for the type of @p minimum_delay.
+   * @tparam Rep2 similar formal parameter for the type of
+   *    @p initial_delay_upper_bound.
+   * @tparam Period2 similar formal parameter for the type of
+   *    @p initial_delay_upper_bound.
    * @tparam Rep3 similar formal parameter for the type of @p maximum_delay.
    * @tparam Period3 similar formal parameter for the type of @p maximum_delay.
    *
@@ -176,21 +180,17 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
   template <typename Rep1, typename Period1, typename Rep2, typename Period2,
             typename Rep3, typename Period3>
   ExponentialBackoffPolicy(
-      std::chrono::duration<Rep1, Period1> initial_delay_upper_bound,
-      std::chrono::duration<Rep2, Period2> minimum_delay,
+      std::chrono::duration<Rep1, Period1> minimum_delay,
+      std::chrono::duration<Rep2, Period2> initial_delay_upper_bound,
       std::chrono::duration<Rep3, Period3> maximum_delay, double scaling)
-      : initial_delay_upper_bound_(initial_delay_upper_bound),
-        minimum_delay_(minimum_delay),
+      : minimum_delay_(minimum_delay),
+        initial_delay_upper_bound_(initial_delay_upper_bound),
         maximum_delay_(maximum_delay),
         scaling_(scaling),
         current_delay_end_(initial_delay_upper_bound_) {
     if (initial_delay_upper_bound_ < minimum_delay_) {
       google::cloud::internal::ThrowInvalidArgument(
           "initial delay upper bound must be >= minimum delay");
-    }
-    if (initial_delay_upper_bound_.count() == 0) {
-      google::cloud::internal::ThrowInvalidArgument(
-          "initial delay upper bound must be non zero");
     }
     if (scaling_ <= 1.0) {
       google::cloud::internal::ThrowInvalidArgument(
@@ -203,8 +203,8 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
   //    know specifically which one is at fault)
   //  - We want uncorrelated data streams for each copy anyway.
   ExponentialBackoffPolicy(ExponentialBackoffPolicy const& rhs) noexcept
-      : initial_delay_upper_bound_(rhs.initial_delay_upper_bound_),
-        minimum_delay_(rhs.minimum_delay_),
+      : minimum_delay_(rhs.minimum_delay_),
+        initial_delay_upper_bound_(rhs.initial_delay_upper_bound_),
         maximum_delay_(rhs.maximum_delay_),
         scaling_(rhs.scaling_),
         current_delay_end_(rhs.current_delay_end_) {}
@@ -214,8 +214,8 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
 
  private:
   using DoubleMicroseconds = std::chrono::duration<double, std::micro>;
-  DoubleMicroseconds initial_delay_upper_bound_;
   DoubleMicroseconds minimum_delay_;
+  DoubleMicroseconds initial_delay_upper_bound_;
   DoubleMicroseconds maximum_delay_;
   double scaling_;
   DoubleMicroseconds current_delay_end_;
--- a/google/cloud/internal/backoff_policy.h
+++ b/google/cloud/internal/backoff_policy.h
@@ -90,7 +90,7 @@ class BackoffPolicy {
 class ExponentialBackoffPolicy : public BackoffPolicy {
  public:
   /**
-   * Default constructor for an exponential backoff policy.
+   * Constructor for an exponential backoff policy.
    *
    * Define the initial delay, maximum delay, and scaling factor for an instance
    * of the policy. While the constructor accepts `std::chrono::duration`
@@ -181,8 +181,8 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
   template <typename Rep1, typename Period1, typename Rep2, typename Period2,
             typename Rep3, typename Period3>
   ExponentialBackoffPolicy(
-      std::chrono::duration<Rep1, Period1> minimum_delay,
-      std::chrono::duration<Rep2, Period2> initial_delay_upper_bound,
+      std::chrono::duration<Rep1, Period1> initial_delay_range,
+      std::chrono::duration<Rep2, Period2> minimum_delay,
       std::chrono::duration<Rep3, Period3> maximum_delay, double scaling)
       : minimum_delay_(minimum_delay),
         initial_delay_upper_bound_(initial_delay_upper_bound),
--- a/google/cloud/internal/backoff_policy.h
+++ b/google/cloud/internal/backoff_policy.h
@@ -73,21 +73,24 @@ class BackoffPolicy {
 };
 
 /**
- * Implements a truncated exponential backoff with randomization policy.
+ * Implements a truncated exponential backoff with randomization policy and a
+ * minimum delay.
  *
  * This policy implements the truncated exponential backoff policy for
  * retrying operations. After a request fails, and subject to a separate
- * retry policy, the client library will wait for an initial delay before
- * trying again. If the second attempt fails the delay time is increased,
- * using an scaling factor. The delay time growth stops at a maximum delay
- * wait time. The policy also randomizes the delay each time, to avoid
- * [thundering herd
+ * retry policy, the client library will wait for an initial delay after
+ * the specified minimum delay before trying again. If the second attempt fails
+ * the delay time is increased, using a scaling factor.The delay time begins at
+ * the minimum delay, if no minimum is specified the default constructor will
+ * begin use the initial delay as the minimum.  The delay time growth stops at a
+ * maximum delay wait time. The policy also randomizes the delay each time, to
+ * avoid [thundering herd
  * problem](https://en.wikipedia.org/wiki/Thundering_herd_problem).
  */
 class ExponentialBackoffPolicy : public BackoffPolicy {
  public:
   /**
-   * Constructor for an exponential backoff policy.
+   * Default constructor for an exponential backoff policy.
    *
    * Define the initial delay, maximum delay, and scaling factor for an instance
    * of the policy. While the constructor accepts `std::chrono::duration`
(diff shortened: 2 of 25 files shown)


google/cloud/internal/backoff_policy.h
```diff
@@ -126,11 +129,64 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
...
+   *     operation.
+   * @param minimum_delay the minimum value for the delay between operations.
+   * @param maximum_delay the maximum value for the delay between operations.
+   * @param scaling how fast does the delay increase between iterations.
+   *
+   * @tparam Rep1 a placeholder to match the Rep tparam for @p
+   * initial_delay_range's type, the semantics of this template parameter are
+   * documented in `std::chrono::duration<>` (in brief, the underlying
+   * arithmetic type used to store the number of ticks), for our purposes it is
+   * simply a formal parameter.
```
devbww: Nit: I know this is a copy of previous text, but it might be nice to eliminate the comma splices in both places while you're here.

google/cloud/internal/backoff_policy.h
```diff
@@ -126,11 +129,64 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
...
+   * @tparam Period1 a placeholder to match the Period tparam for
+   *     @p initial_delay_range's type, the semantics of this template
+   * parameter are documented in `std::chrono::duration<>` (in brief, the length
+   * of the tick in seconds, expressed as a `std::ratio<>`), for our purposes it
+   *     is simply a formal parameter.
```
devbww: Should the `@tparam` body be indented like it is in the other constructor?

google/cloud/internal/backoff_policy.h
```diff
@@ -126,11 +129,64 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
...
+   * @endcode
+   *
+   * @param initial_delay_range how long to wait after the first (unsuccessful)
+   *     operation.
+   * @param minimum_delay the minimum value for the delay between operations.
+   * @param maximum_delay the maximum value for the delay between operations.
+   * @param scaling how fast does the delay increase between iterations.
+   *
+   * @tparam Rep1 a placeholder to match the Rep tparam for @p
+   * initial_delay_range's type, the semantics of this template parameter are
```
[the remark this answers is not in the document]
alevenberg: My formatter won't auto allign the doc. I manually moved it for this commit. Do you know which formatter I can use to be consistent (in VSCode)?

google/cloud/internal/backoff_policy.h
```diff
@@ -73,21 +73,24 @@ class BackoffPolicy {
...
+ * begin use the initial delay as the minimum.  The delay time growth stops at a
+ * maximum delay wait time. The policy also randomizes the delay each time, to
+ * avoid [thundering herd
  * problem](https://en.wikipedia.org/wiki/Thundering_herd_problem).
  */
 class ExponentialBackoffPolicy : public BackoffPolicy {
  public:
   /**
-   * Constructor for an exponential backoff policy.
+   * Default constructor for an exponential backoff policy.
```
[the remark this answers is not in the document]
alevenberg: Ack. That makes sense. Removed.

google/cloud/internal/backoff_policy.h
alevenberg: Yes. Done.

google/cloud/internal/backoff_policy.h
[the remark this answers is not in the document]
coryan: I just use the space bar.  If you find this annoying we can add a tool, or change the "standard".  I am not particularly obsessed with consistency, more with "I can read it".  Without indentation it is hard to tell when a new parameter description starts.

google/cloud/internal/backoff_policy.h
alevenberg: Done.

google/cloud/internal/backoff_policy.h
```diff
@@ -126,11 +129,64 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
...
+   * internally in microseconds. Sub-microsecond delays seem unnecessarily
+   * precise for this application.
+   *
+   * @code
+   * using namespace std::chrono_literals; // C++14
+   * auto r1 = ExponentialBackoffPolicy<S,T>(0ms, 10ms, 500ms, 1.618);
+   * auto r2 = ExponentialBackoffPolicy<S,T>(0min, 10min, 10min + 2s, 1.002);
+   * @endcode
+   *
+   * @param initial_delay_range how long to wait after the first (unsuccessful)
```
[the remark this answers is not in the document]
alevenberg: I agree it is not ideal.

An initial change is /s/initial_delay_range/initial_delay

google/cloud/internal/backoff_policy.h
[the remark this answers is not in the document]
alevenberg: Ok manually changing is fine. Just wanted to make sure there wasn't something I was missing.

dbolduc (changes requested):
The main thing is that we should not think in terms of ranges. Let's think in terms of the endpoints of the range. We will fix the minimum delay. Then grow the upper end of the range by the scaling factor, until it hits the overall maximum delay.

The only modifications to existing tests should be the lower bounds changing to equal the minimum/initial delay. (We should still add new tests for the new feature)

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-:-NVoheBzBxJ-3YYH_tYL:bedw6qc)* status: 0 of 4 files reviewed, 26 unresolved discussions (waiting on @alevenberg, @coryan, and @devbww)

___
*[`-- commits` line 2 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoiV949tiQxL_EwEfS):*
nit: I would call this a `feat: ...`. You are adding something new to the library that it could not do before. And the interface is user-facing.
___
*[`google/cloud/internal/backoff_policy.h` line 129 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVogLAHA7ldalrxHChF:-NVogLAHA7ldalrxHChG:b-1kn3de) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L129)):*
> ```objc
>    */
>   template <typename Rep1, typename Period1, typename Rep2, typename Period2>
>   ExponentialBackoffPolicy(std::chrono::duration<Rep1, Period1> initial_delay,
> ```

I would change the name from `initial_delay` to `minimum_delay` now.
___
*[`google/cloud/internal/backoff_policy.h` line 148 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVojSj8AAF4M3DInL0Y:-NVojSj9AOcDO4pqkGoR:b1xp1zm) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L148)):*
> ```objc
>    * using namespace std::chrono_literals; // C++14
>    * auto r1 = ExponentialBackoffPolicy<S,T>(0ms, 10ms, 500ms, 1.618);
>    * auto r2 = ExponentialBackoffPolicy<S,T>(0min, 10min, 10min + 2s, 1.002);
> ```

I am not sure what this example is demonstrating. I would just remove it.
___
*[`google/cloud/internal/backoff_policy.h` line 178 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoZVTwExGYY3arNxmL:-NVoZVTwExGYY3arNxmM:bzgsqn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L178)):*
> ```objc
>   template <typename Rep1, typename Period1, typename Rep2, typename Period2,
>             typename Rep3, typename Period3>
>   ExponentialBackoffPolicy(std::chrono::duration<Rep1, Period1> initial_delay,
> ```

I am very confused about the difference between the `initial_delay` and `minimum_delay`. And the examples do not help me.

I think we can make the API clearer:

```cc
ExponentialBackoffPolicy(minimum_delay, initial_delay_upper_bound, maximum_delay, scaling);
```

I think this is more intuitive because `minimum_delay <= initial_delay_upper_bound <= maximum_delay`.
___
*[`google/cloud/internal/backoff_policy.h` line 186 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVodvPECjvvqTTlLllC:-NVodvPECjvvqTTlLllD:bdjy74e) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L186)):*
> ```objc
>         maximum_delay_(maximum_delay),
>         scaling_(scaling),
>         current_delay_range_(initial_delay_),
> ```

We shouldn't think in terms of the range. We should delete this.
___
*[`google/cloud/internal/backoff_policy.h` line 189 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVob1EZ7ePthlTwVaYw:-NVob1EZ7ePthlTwVaYx:b4se1ki) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L189)):*
> ```objc
>         current_delay_start_(minimum_delay_),
>         current_delay_end_(
>             (std::min)(minimum_delay_ + current_delay_range_, maximum_delay_)) {
> ```

We don't need to do this arithmetic if we ask for an `initial_delay_upper_bound` instead of an `initial_delay`.

I do not understand the arithmetic either. Doubling the initial/minimum delay is why tests like`DetermineRangeUsingScalingFactor` started failing.
___
*[`google/cloud/internal/backoff_policy.h` line 191 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVog_YH91b4B8cTEKof:-NVog_YH91b4B8cTEKog:b-t5wrc5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L191)):*
> ```objc
>             (std::min)(minimum_delay_ + current_delay_range_, maximum_delay_)) {
>     if (scaling_ <= 1.0) {
>       google::cloud::internal::ThrowInvalidArgument(
> ```

We might want to add argument checking for the durations. Especially if our code makes assumptions about them. Something like:

```cc
if (minimum_delay_ > initial_delay_upper_bound_) {
  ThrowInvalidArgument("Initial delay upper bound must be greater than or equal to the minimum delay");
}
```

Note that no existing code would be broken by adding this check.
___
*[`google/cloud/internal/backoff_policy.h` line 219 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVo_exO9Ok5nivVa0Bk:-NVo_exO9Ok5nivVa0Bl:bc8a2t7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L219)):*
> ```objc
>   double scaling_;
>   DoubleMicroseconds current_delay_range_;
>   DoubleMicroseconds current_delay_start_;
> ```

`current_delay_start_` should always be `minimum_delay_`. So we don't need it.
___
*[`google/cloud/internal/backoff_policy.cc` line 24 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoaTvREjJu_6XIg-qq:-NVoaTvSDwvHdAf_vHa6:bxgilz7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.cc#L24)):*
> ```cpp
> std::unique_ptr<BackoffPolicy> ExponentialBackoffPolicy::clone() const {
>   return std::make_unique<ExponentialBackoffPolicy>(
>       initial_delay_, minimum_delay_, maximum_delay_, scaling_);
> ```

Lol, we flipped the order of the parameters.

Now you need to add a test where `initial_delay_ != minimum_delay_` to catch it.
___
*[`google/cloud/internal/backoff_policy.cc` line 42 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVofAv2DyvuPlBL9UpR:-NVofAv2DyvuPlBL9UpS:b-seqomi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d84ef9bee5a980047b3bc606bb56acadfd1983d4/google/cloud/internal/backoff_policy.cc#L42)):*
> ```cpp
>   }
> 
>   if (current_delay_end_ >= maximum_delay_) {
> ```

I liked doing this check before generating the random number. We could also write it as:

```cc
current_delay_end_ = (std::min)(current_delay_end_, maximum_delay_);
```
___
*[`google/cloud/internal/backoff_policy.cc` line 48 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoflZsE0K9QaGsoFgQ:-NVoflZt5OMOteN9R2P9:b2kw8gi) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.cc#L48)):*
> ```cpp
>   auto delay = DoubleMicroseconds(rng_distribution(*generator_));
> 
>   if (current_delay_end_ < maximum_delay_) {
> ```

If we do bound checking before random generation, this can just be:

```cc
current_delay_end_ *= scaling_;
```
___
*[`google/cloud/internal/backoff_policy_test.cc` line 36 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoWvzt-2zhHuQr_E3-:-NVoWvzt-2zhHuQr_E30:b1oidsh) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy_test.cc#L36)):*
> ```cpp
>   delay = tested.OnCompletion();
>   EXPECT_LE(ms(10), delay);
>   EXPECT_GE(ms(30), delay);
> ```

why is this changing? Only the lower bounds should change.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 63 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoXCJpEIcHiXNgptxY:-NVoXCJpEIcHiXNgptxZ:b-p16uk9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d84ef9bee5a980047b3bc606bb56acadfd1983d4/google/cloud/internal/backoff_policy_test.cc#L63)):*
> ```cpp
>   auto delay = tested.OnCompletion();
>   EXPECT_LE(ms(1000), delay);
>   EXPECT_GE(ms(1001), delay);
> ```

This test was fine. Why did it change?
___
*[`google/cloud/internal/backoff_policy_test.cc` line 111 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoZG7K27-iirCIXTUm:-NVoZG7K27-iirCIXTUn:b64dleu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy_test.cc#L111)):*
> ```cpp
>   auto delay = tested.OnCompletion();
>   EXPECT_LE(ms(100), delay) << "delay=" << delay.count() << "ms";
>   EXPECT_GE(ms(200), delay) << "delay=" << delay.count() << "ms";
> ```

I do not think upper bound expectations should change.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 130 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoZNq1Cx-HpkZnRJbB:-NVoZNq1Cx-HpkZnRJbC:b-7jl271) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy_test.cc#L130)):*
> ```cpp
>   delay = tested->OnCompletion();
>   EXPECT_LE(ms(10), delay);
>   EXPECT_GE(ms(30), delay);
> ```

same.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 175 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVojkihDHuqkppvO8br:-NVojkihDHuqkppvO8bs:b2nhjic) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy_test.cc#L175)):*
> ```cpp
>   // whatever the C++ library uses for entropy (typically /dev/random and/or the
>   // RND instruction) manage to produce the same 20 numbers. If that happens,
>   // my apologies.... and remember to buy yourself a lottery ticket today.
> ```

Lol
___
*[`google/cloud/storage/examples/storage_object_samples.cc` line 583 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoYf8v36CeSLJKUyu7:-NVoYf8v36CeSLJKUyu8:b-4lrknw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/storage/examples/storage_object_samples.cc#L583)):*
> ```cpp
>   options.set<gcs::IdempotencyPolicyOption>(
>       gcs::StrictIdempotencyPolicy().clone());
>   // On error, it backs off for 2 seconds, then 5 seconds, then 14 seconds, etc.
> ```

"On error, it backs off for 1-3 seconds, then 1-9 seconds, then 1-27 seconds, etc." ?

Or just revert the change. I think the reader will get the point.
___
*[`google/cloud/storage/examples/storage_object_samples.cc` line 589 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoXjpF4s3GxBb_ND9f:-NVoXjpF4s3GxBb_ND9g:b-czmqax) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/storage/examples/storage_object_samples.cc#L589)):*
> ```cpp
>       gcs::ExponentialBackoffPolicy(
>           /*initial_delay=*/std::chrono::seconds(1),
>           /*minimum_delay=*/std::chrono::seconds(1),
> ```

I would revert this. It is equivalent to the existing code. So why change it?

The point of adding the other argument to the constructor is to support ranges like (0, N). Storage suggests using a minimum backoff of 1 second, so the new constructor is less applicable.

alevenberg (commented):
Ack, made changes

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-:-NVpf7ah-HfJd9GZX9I-:by7s0hc)* status: 0 of 4 files reviewed, 25 unresolved discussions (waiting on @coryan, @dbolduc, and @devbww)

___
*[`google/cloud/internal/backoff_policy.h` line 83 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1198292498:-NVpJZiy97AMRVRlcip6:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f15b8b0823e3031b0a5f05ba6e63243eab5e2e7c/google/cloud/internal/backoff_policy.h#L83)):*

`s/factor.The/factor. The/`
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 84 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1198293235:-NVpJZ2SCt2ByQYsowNB:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f15b8b0823e3031b0a5f05ba6e63243eab5e2e7c/google/cloud/internal/backoff_policy.h#L84)):*

`s/delay, if/delay. If/`
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 84 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1198295210:-NVpJZIv0a94wo7FpKz3:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f15b8b0823e3031b0a5f05ba6e63243eab5e2e7c/google/cloud/internal/backoff_policy.h#L84)):*

Re _"default contructor"_: See below.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 85 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1198293414:-NVpJYlzBkZonJBgW7Cy:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f15b8b0823e3031b0a5f05ba6e63243eab5e2e7c/google/cloud/internal/backoff_policy.h#L85)):*

`s/begin //`
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 86 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1198293634:-NVpJYZA8XYh0f2jIWQl:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f15b8b0823e3031b0a5f05ba6e63243eab5e2e7c/google/cloud/internal/backoff_policy.h#L86)):*

`s/wait //`
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 87 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1198294234:-NVpJYB_58IHmahpmyfk:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f15b8b0823e3031b0a5f05ba6e63243eab5e2e7c/google/cloud/internal/backoff_policy.h#L87)):*

Inherited problem, but `s/avoid/avoid the/`
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 129 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVogLAHA7ldalrxHChF:-NVpK9EB1CjXMfSh2wte:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L129)):*

I would change the name from `initial_delay` to `minimum_delay` now.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 148 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVojSj8AAF4M3DInL0Y:-NVpKDvDDrB0Pdshu15S:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L148)):*

I am not sure what this example is demonstrating. I would just remove it.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 178 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoZVTwExGYY3arNxmL:-NVpKIVh3NKgl0Ls3TeC:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L178)):*

I am very confused about the difference between the `initial_delay` and `minimum_delay`. And the examples do not help me.

I think we can make the API clearer:

```cc
ExponentialBackoffPolicy(minimum_delay, initial_delay_upper_bound, maximum_delay, scaling);
```

I think this is more intuitive because `minimum_delay <= initial_delay_upper_bound <= maximum_delay`.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 186 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVodvPECjvvqTTlLllC:-NVpY6qX97TnWnvZu_lz:b-ho5wpv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L186)):*

We shouldn't think in terms of the range. We should delete this.
</blockquote></details>

Agreed. Done.
___
*[`google/cloud/internal/backoff_policy.h` line 189 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVob1EZ7ePthlTwVaYw:-NVpZHbx7WFPv3khViL5:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L189)):*

We don't need to do this arithmetic if we ask for an `initial_delay_upper_bound` instead of an `initial_delay`.

I do not understand the arithmetic either. Doubling the initial/minimum delay is why tests like`DetermineRangeUsingScalingFactor` started failing.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 191 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVog_YH91b4B8cTEKof:-NVpZIffCzW_Km0LiGW-:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L191)):*

We might want to add argument checking for the durations. Especially if our code makes assumptions about them. Something like:

```cc
if (minimum_delay_ > initial_delay_upper_bound_) {
  ThrowInvalidArgument("Initial delay upper bound must be greater than or equal to the minimum delay");
}
```

Note that no existing code would be broken by adding this check.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 219 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVo_exO9Ok5nivVa0Bk:-NVpL5939TFH9NPEwiQh:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L219)):*

`current_delay_start_` should always be `minimum_delay_`. So we don't need it.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.cc` line 24 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoaTvREjJu_6XIg-qq:-NVpagUiAqIkkGb1dYZb:b1ix52j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.cc#L24)):*

Lol, we flipped the order of the parameters.

Now you need to add a test where `initial_delay_ != minimum_delay_` to catch it.
</blockquote></details>

I think the order of the params is correct and the check initial_delay_upper_bound_ < minimum_delay_ should catch the issue
___
*[`google/cloud/internal/backoff_policy.cc` line 42 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVofAv2DyvuPlBL9UpR:-NVpOZKj1zvj4Zx48_Je:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d84ef9bee5a980047b3bc606bb56acadfd1983d4/google/cloud/internal/backoff_policy.cc#L42)):*

I liked doing this check before generating the random number. We could also write it as:

```cc
current_delay_end_ = (std::min)(current_delay_end_, maximum_delay_);
```
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.cc` line 48 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoflZsE0K9QaGsoFgQ:-NVpOiOmBjj2DYJ54O_6:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.cc#L48)):*

If we do bound checking before random generation, this can just be:

```cc
current_delay_end_ *= scaling_;
```
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 36 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoWvzt-2zhHuQr_E3-:-NVpU5rW4M9GQkxDcVkf:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy_test.cc#L36)):*

why is this changing? Only the lower bounds should change.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 63 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoXCJpEIcHiXNgptxY:-NVpU6Vb8ZEKHKf0hCMR:b-o5sd4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d84ef9bee5a980047b3bc606bb56acadfd1983d4/google/cloud/internal/backoff_policy_test.cc#L63)):*

This test was fine. Why did it change?
</blockquote></details>

discussed offline. fixed tests.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 111 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoZG7K27-iirCIXTUm:-NVpUBer8O1SRL1VhEPA:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy_test.cc#L111)):*

I do not think upper bound expectations should change.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 130 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoZNq1Cx-HpkZnRJbB:-NVpUCTZ4ZyslwskQO3y:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy_test.cc#L130)):*

same.
</blockquote></details>

Done.
___
*[`google/cloud/storage/examples/storage_object_samples.cc` line 583 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoYf8v36CeSLJKUyu7:-NVpa6U16ORRXUpZIp7T:b9fe11w) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/storage/examples/storage_object_samples.cc#L583)):*

"On error, it backs off for 1-3 seconds, then 1-9 seconds, then 1-27 seconds, etc." ?

Or just revert the change. I think the reader will get the point.
</blockquote></details>

I think it's not obvious what the behavior is with this description. Updated it to include the intervals.
___
*[`google/cloud/storage/examples/storage_object_samples.cc` line 589 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoXjpF4s3GxBb_ND9f:-NVpZfHl1oFA4ZAJx-po:bd4c0uu) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/storage/examples/storage_object_samples.cc#L589)):*

I would revert this. It is equivalent to the existing code. So why change it?

The point of adding the other argument to the constructor is to support ranges like (0, N). Storage suggests using a minimum backoff of 1 second, so the new constructor is less applicable.
</blockquote></details>

Reverted this change. Leaving the change to the comment to clarify the behavior.

google/cloud/storage/examples/storage_object_samples.cc
```diff
@@ -580,12 +580,12 @@ google::cloud::storage::Client StorageRetries(std::vector<std::string> const&) {
   // Retries only idempotent operations.
   options.set<gcs::IdempotencyPolicyOption>(
       gcs::StrictIdempotencyPolicy().clone());
-  // On error, it backs off for 2 seconds, then 5 seconds, then 14 seconds, etc.
-  // The backoff time never grows larger than 1 minute. The strategy introduces
-  // jitter around the backoff delay.
+  // On error, it backs off for a random delay between the following intervals
```
dbolduc, line 583: s/the following intervals//

google/cloud/storage/examples/storage_object_samples.cc
```diff
@@ -580,12 +580,12 @@ google::cloud::storage::Client StorageRetries(std::vector<std::string> const&) {
...
+  // [1, 3] seconds, then [1, 9] seconds, then [1, 27] seconds, etc. The backoff
+  // time never grows larger than 1 minute. The strategy introduces jitter
+  // around the backoff delay.
```
dbolduc, line 586: We can remove this sentence because "introduces jitter" means the same thing as "random delay"

google/cloud/internal/backoff_policy.h
```diff
@@ -175,18 +175,23 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
...
   template <typename Rep1, typename Period1, typename Rep2, typename Period2,
             typename Rep3, typename Period3>
-  ExponentialBackoffPolicy(std::chrono::duration<Rep1, Period1> initial_delay,
-                           std::chrono::duration<Rep2, Period2> minimum_delay,
-                           std::chrono::duration<Rep3, Period3> maximum_delay,
-                           double scaling)
-      : initial_delay_(initial_delay),
+  ExponentialBackoffPolicy(
+      std::chrono::duration<Rep1, Period1> initial_delay_upper_bound,
+      std::chrono::duration<Rep2, Period2> minimum_delay,
```
dbolduc, line 180: I think it is way more intuitive to switch the order of these two parameters.

google/cloud/internal/backoff_policy.h
```diff
@@ -126,44 +126,44 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
...
    *
    * @code
    * using namespace std::chrono_literals; // C++14
    * auto r1 = ExponentialBackoffPolicy<S,T>(0ms, 10ms, 500ms, 1.618);
-   * auto r2 = ExponentialBackoffPolicy<S,T>(0min, 10min, 10min + 2s, 1.002);
    * @endcode
    *
-   * @param initial_delay how long to wait after the first (unsuccessful)
-   *     operation.
+   * @param initial_delay_upper_bound how long to wait after the first
```
dbolduc, line 151: s/how long to wait/the longest possible delay/

google/cloud/internal/backoff_policy.h
```diff
@@ -126,44 +126,44 @@ class ExponentialBackoffPolicy : public BackoffPolicy {
    *     for more details.
    */
   template <typename Rep1, typename Period1, typename Rep2, typename Period2>
-  ExponentialBackoffPolicy(std::chrono::duration<Rep1, Period1> initial_delay,
+  ExponentialBackoffPolicy(std::chrono::duration<Rep1, Period1> minimum_delay,
```
dbolduc, line 129: Thanks for updating the name. Can we also update the `@p initial_delay`s in the documentation?

alevenberg (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650)* status: 0 of 25 files reviewed, 30 unresolved discussions (waiting on @coryan, @dbolduc, and @devbww)

___
*[`google/cloud/internal/backoff_policy.h` line 129 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1199356061:-NVprxW17yr8m9xiPMRt:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5e29e159867dbaa33a89af59f907d1b507db13a0/google/cloud/internal/backoff_policy.h#L129)):*

Thanks for updating the name. Can we also update the `@p initial_delay`s in the documentation?
</blockquote></details>

Done
___
*[`google/cloud/internal/backoff_policy.h` line 151 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1199346229:-NVprRSk9OCA1iVx9jcB:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5e29e159867dbaa33a89af59f907d1b507db13a0/google/cloud/internal/backoff_policy.h#L151)):*

s/how long to wait/the longest possible delay/
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.h` line 180 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1199345195:-NVprSZu-BBLhJ6Gi2Rj:b-ho5wpv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5e29e159867dbaa33a89af59f907d1b507db13a0/google/cloud/internal/backoff_policy.h#L180)):*

I think it is way more intuitive to switch the order of these two parameters.
</blockquote></details>

Agreed. Done.
___
*[`google/cloud/storage/examples/storage_object_samples.cc` line 583 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1199343661:-NVpr5PWE6Bv_Jj6XbiI:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5e29e159867dbaa33a89af59f907d1b507db13a0/google/cloud/storage/examples/storage_object_samples.cc#L583)):*

s/the following intervals//
</blockquote></details>

Done.
___
*[`google/cloud/storage/examples/storage_object_samples.cc` line 586 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#gh-1199344062:-NVpr5di9MDb6qEDO8G1:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/5e29e159867dbaa33a89af59f907d1b507db13a0/google/cloud/storage/examples/storage_object_samples.cc#L586)):*

We can remove this sentence because "introduces jitter" means the same thing as "random delay"
</blockquote></details>

Done.

dbolduc (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650)* status: 0 of 25 files reviewed, 34 unresolved discussions (waiting on @alevenberg, @coryan, and @devbww)

___
*[`google/cloud/internal/backoff_policy.h` line 129 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVogLAHA7ldalrxHChF:-NW8U06e1jqpsPe0qmQb:b-nyne4e) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L129)):*

Done.
</blockquote></details>

I hate to say it, but `initial_delay` is a better name again. Sorry!!!
___
*[`google/cloud/internal/backoff_policy.cc` line 43 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NW8VcJs2ns4YOF-0Q3M:-NW8VcJs2ns4YOF-0Q3N:bjv9vd0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6fb98eade8e66a6b007ba872329c78a51ea8e217/google/cloud/internal/backoff_policy.cc#L43)):*
> ```cpp
> 
>   if (current_delay_end_ >= maximum_delay_) {
>     current_delay_start_ = std::max(maximum_delay_ / scaling_, initial_delay_);
> ```

I think this would become:

```
current_delay_start_ = (std::min)(current_delay_start_, (std::max)(maximum_delay_ / scaling_, initial_delay_));
```

to support this case: https://github.com/googleapis/google-cloud-cpp/blob/f8db4553470ce283d4935f763c4800b34fbdf993/google/cloud/internal/backoff_policy_test.cc#L55-L57
___
*[`google/cloud/internal/backoff_policy.cc` line 53 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NW8UfZtBr-sklNGoFZe:-NW8UfZtBr-sklNGoFZf:b-qd7qou) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6fb98eade8e66a6b007ba872329c78a51ea8e217/google/cloud/internal/backoff_policy.cc#L53)):*
> ```cpp
>   auto delay = DoubleMicroseconds(rng_distribution(*generator_));
> 
>   current_delay_start_ = current_delay_end_;
> ```

I think we should restore the `current_delay_start_` logic, except changing this line to be `current_delay_start_ *= scaling_;`
___
*[`google/cloud/internal/backoff_policy_test.cc` line 35 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NW8UzoS7pAGw9EgwfK4:-NW8UzoS7pAGw9EgwfK5:bfwimh7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f8db4553470ce283d4935f763c4800b34fbdf993/google/cloud/internal/backoff_policy_test.cc#L35)):*
> ```cpp
>   EXPECT_GE(ms(20), delay);
>   delay = tested.OnCompletion();
>   EXPECT_LE(ms(10), delay);
> ```

Here, and elsewhere, we should restore the lower bounds in tests of the old constructor.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 43 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NW8WjW697e9FuFMbzIC:-NW8WjW697e9FuFMbzID:b-9rlmvf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f8db4553470ce283d4935f763c4800b34fbdf993/google/cloud/internal/backoff_policy_test.cc#L43)):*
> ```cpp
> 
> /// @test Verify a full jitter policy, where the minimum delay is set to 0.
> TEST(ExponentialBackoffPolicy, VerifyFullJitterPolicy) {
> ```

comment: This remains a good test 👍
___
*[`google/cloud/storage/examples/storage_object_samples.cc` line 583 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoYf8v36CeSLJKUyu7:-NW8XgwG8fzK3GMtEUWx:b7pclh4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/storage/examples/storage_object_samples.cc#L583)):*

I think it's not obvious what the behavior is with this description. Updated it to include the intervals.
</blockquote></details>

Argh, this is wrong again. Now its [1, 3], [3, 9], [9, 27]

alevenberg (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650)* status: 0 of 25 files reviewed, 34 unresolved discussions (waiting on @coryan, @dbolduc, and @devbww)

___
*[`google/cloud/internal/backoff_policy.h` line 129 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVogLAHA7ldalrxHChF:-NW8tOJS-5RJCJ5N-mnK:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/internal/backoff_policy.h#L129)):*

I hate to say it, but `initial_delay` is a better name again. Sorry!!!
</blockquote></details>

Done.

alevenberg (commented):
Modified so that there is an extra parameter (scaling_lower_bound) that can move if set. The constructor can support all 3 behaviors discussed (FullJitter, MinJitter, and the initial ExponentialBackoff)

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-:-NWE2t_n0kxKKs0Y5h4I:blx0tsw)* status: 0 of 25 files reviewed, 34 unresolved discussions (waiting on @coryan, @dbolduc, and @devbww)

___
*[`google/cloud/internal/backoff_policy.cc` line 43 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NW8VcJs2ns4YOF-0Q3M:-NWE2WJH7x_MgQxewhsr:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f589c230d628fc37055be50794c60b529ca82c86/google/cloud/internal/backoff_policy.cc#L43)):*

I think this would become:

```
current_delay_start_ = (std::min)(current_delay_start_, (std::max)(maximum_delay_ / scaling_, initial_delay_));
```

to support this case: https://github.com/googleapis/google-cloud-cpp/blob/f8db4553470ce283d4935f763c4800b34fbdf993/google/cloud/internal/backoff_policy_test.cc#L55-L57
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy.cc` line 53 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NW8UfZtBr-sklNGoFZe:-NWE2WgI3ciDnHWw-4z1:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f589c230d628fc37055be50794c60b529ca82c86/google/cloud/internal/backoff_policy.cc#L53)):*

I think we should restore the `current_delay_start_` logic, except changing this line to be `current_delay_start_ *= scaling_;`
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 35 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NW8UzoS7pAGw9EgwfK4:-NWE2XnJBHSwq1ax_Gzi:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d0f192c30c4ba75779202900bfb3d2b26fd1a52f/google/cloud/internal/backoff_policy_test.cc#L35)):*

Here, and elsewhere, we should restore the lower bounds in tests of the old constructor.
</blockquote></details>

Done.
___
*[`google/cloud/internal/backoff_policy_test.cc` line 43 at r7](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NW8WjW697e9FuFMbzIC:-NWE2YIU80-XvaAtGxMB:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d0f192c30c4ba75779202900bfb3d2b26fd1a52f/google/cloud/internal/backoff_policy_test.cc#L43)):*

comment: This remains a good test 👍
</blockquote></details>

Done.
___
*[`google/cloud/storage/examples/storage_object_samples.cc` line 583 at r4](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-NVoYf8v36CeSLJKUyu7:-NWE2Yw4EqPQNV-TuuAN:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f3fa1db1e62b4142ec47c7d511fed7a61fa1bd83/google/cloud/storage/examples/storage_object_samples.cc#L583)):*

Argh, this is wrong again. Now its [1, 3], [3, 9], [9, 27]
</blockquote></details>

Done.

dbolduc:
> Modified ...

I do not see any new changes.

dbolduc (commented):
Very close. I have mostly documentation nits. I don't quite agree with the lower bound logic.

google/cloud/internal/backoff_policy.cc
```diff
@@ -40,8 +41,11 @@ std::chrono::milliseconds ExponentialBackoffPolicy::OnCompletion() {
   }
 
   if (current_delay_end_ >= maximum_delay_) {
-    current_delay_start_ = std::max(maximum_delay_ / scaling_, initial_delay_);
-    current_delay_end_ = maximum_delay_;
+    if (scaling_lower_bound_ != 1) {
+      current_delay_start_ =
+          (std::max)(minimum_delay_, current_delay_end_ / scaling_lower_bound_);
+    }
```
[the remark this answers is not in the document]
dbolduc: > consider: ExponentialBackoffPolicy(ms(1), ms(10), ms(10), 2.0, 2.0)
>
> The range will be:
> [1, 10], [1, 10], [1, 10], etc.

argh, sorry. The range would be:
[5, 10], [5, 10], [5, 10], etc.

> `scaling_lower_bound == 1 addresses the minjitter case.`

Ok, but if `scaling_lower_bound == 1.0000000000000000000001`, the behavior will be very different. That's weird.

> Added the test case you mentioned

The test is passing because the actual ranges are subsets of the expected/desired ranges.

Consider:

```cc
TEST(ExponentialBackoffPolicy, DumbTestButYouGetTheIdea) {
  for (auto i = 0; i != 20; ++i) {
    ExponentialBackoffPolicy tested(ms(0), ms(1001), ms(1001), 1.001, 1.001);
    auto delay = tested.OnCompletion();
    if (delay < ms(1000)) GTEST_SUCCEED();
  }
  GTEST_FAIL();
}
```

With the current code, the test would fail every run.

If the ranges are correct, the test would fail every 1001^20 runs.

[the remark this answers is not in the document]
alevenberg: Addressed the test cases and got rid of the scaling check.

I *think* this addresses the edge cases.

dbolduc (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11650#-:-NWOGT1hEw7h3uFr5szO:bnfp4nl)* status: 0 of 25 files reviewed, 41 unresolved discussions (waiting on @alevenberg, @coryan, and @devbww)

Issue #11531, opened 2023-05-08
Handle LRO operations that don't use `google::longrunning::operation`


scotthart:
Some REST only services support LRO style operations but have their own bespoke `Operation` type and polling mechanisms. 
Modify the existing async infrastructure to handle non-standard LRO mechanisms and enhance the generator to emit code invoking the async infrastructure correctly.

PR #11653, opened 2023-05-18
impl(generator): add longrunning var support for compute operation types

scotthart:
part of the work for #11531 

We're currently hardcoding compute LRO values for many of these variables as that's all we need right now for non `google::longrunning::Operation` types. If in the future, we need support additional LRO mechanisms, we can figure out if we want to abstract the values out and inject them via the config textproto, or some additional branching in this code to handle additional LRO schemes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11653)


coryan (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11653)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

Issue #11428, opened 2023-04-28
Punchlist to make new reference docs public


coryan:
Soon (the pipelines are running) the reference docs will be available internally at Google. There are a number of fixes we need to make afterwards.

- [x] The ToC links the `MOCK_METHOD()` entries for mock classes.
- [x] #11496
- [x] Missing `virtual` vs. `static` vs. `non-virtual` attributes.
- [x] #11455
- [x] #11456
- [x] #11451
- [x] #11519 
- [x] #11583
- [x] The table of contents should not use an "Overview" for functions
- [x] #11457
- [x] #11458
        **Note:** this is because the classes are not in the generated `toc.yml` file. We need to add all classes, structs, typedefs, enums, functions to this `toc.yml`. Fortunately it [can be a tree](https://dotnet.github.io/docfx/docs/table-of-contents.html) and the tree nodes collapse.
- [x] #11453
- [x] #11449
- [x] #11452
- [x] #11588
- [x] #11586
- [x] #11587
- [x] #11614
- [x] A number of input proto files have zero-width joiner characters in their input.  If these appear in a code span (something in backticks), Doxygen converts them to  `&zwj;`.
- [x] Consider splitting namespace-level functions, enums, etc. to their own `.yml` files.  They make the namespace page very long. **DONE:** I considered it, and will keep the functions as they are.  The enums already got their own namespace.
- [x] #11674

----

These are nice to haves (I think):

- [x] Bad example text in [storage::Client::Patch...](https://cloud.google.com/cpp/docs/reference/storage/latest/classgoogle_1_1cloud_1_1storage_1_1Client#classgoogle_1_1cloud_1_1storage_1_1Client_1a5553a035a482ab8a411a6bfd2d8c8dec)
- [x] The typedefs should use `<code>` style for `AliasOf`.
- [x] `simplesect` elements such as `@warning` should use the standard rendering for the cloud site, e.g.:
  ```.html
          <aside class="warning">
            <b>Warning:</b> Stronger than caution; it means "Don't do this."
          </aside>
  ```
- [x] Consider removing destructors from the documentation. They rarely describe anything interesting.

PR #11665, opened 2023-05-19
impl(docfx): annotate virtual and static functions

coryan:
Add a little annotation in the function name when the function is static or virtual.

Part of the work for #11428

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11665)

### DIFF
--- a/docfx/node_name.cc
+++ b/docfx/node_name.cc
@@ -71,6 +71,19 @@ std::string FunctionName(pugi::xml_node node) {
   auto const refqual = std::string_view{node.attribute("refqual").as_string()};
   if (refqual == "rvalue") os << " &&";
   if (refqual == "lvalue") os << " &";
+  sep = "  [";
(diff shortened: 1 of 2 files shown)


docfx/node_name.cc
```diff
@@ -71,6 +71,19 @@ std::string FunctionName(pugi::xml_node node) {
   auto const refqual = std::string_view{node.attribute("refqual").as_string()};
   if (refqual == "rvalue") os << " &&";
   if (refqual == "lvalue") os << " &";
+  sep = "  [";
```
devbww, line 74: Nit Q: Is two spaces a feature?

devbww (approved):

docfx/node_name.cc
coryan: I thought so, but you changed my mind.

PR #11697, opened 2023-05-22
feat(storage): offer mock client without decorators

dbolduc:
Motivated by #11613 

This allows customers to mock a single response from a `RawClient` (instead of N responses in the case of transient errors).

Prefer the `UndecoratedClientFromMock` > `ClientFromMock` in the public samples. I did not update our unit tests.

I did not think about whether extra effort is needed to make the install work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11697)

### DIFF
--- a/google/cloud/storage/testing/mock_client.h
+++ b/google/cloud/storage/testing/mock_client.h
@@ -185,13 +185,25 @@ class MockStreambuf : public internal::ObjectWriteStreambuf {
   MOCK_METHOD(std::uint64_t, next_expected_byte, (), (const, override));
 };
 
-/// Create a client configured to use the given mock.
+/**
+ * Create a client configured to use the given mock.
+ *
+ * Unless you specifically need to mock the behavior of retries, prefer
+ * #google::cloud::storage::testing::UndecoratedClientFromMock
+ */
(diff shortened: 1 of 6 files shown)


google/cloud/storage/testing/mock_client.h
```diff
@@ -185,14 +185,26 @@ class MockStreambuf : public internal::ObjectWriteStreambuf {
   MOCK_METHOD(std::uint64_t, next_expected_byte, (), (const, override));
 };
 
-/// Create a client configured to use the given mock.
+/**
+ * Create a client configured to use the given mock.
+ *
+ * Unless you specifically need to mock the behavior of retries, prefer
+ * #google::cloud::storage::testing::UndecoratedClientFromMock
```
coryan, line 192: Typically if you say `UndecoratedClientFromMock()` (note the parenthesis) Doxygen will automatically link the function.

coryan (approved):

google/cloud/storage/testing/mock_client.h
dbolduc: Fixed, thanks.

PR #11708, opened 2023-05-23
cleanup(storage): inline UndecoratedFromMock()

dbolduc:
Based on feedback from https://github.com/googleapis/google-cloud-cpp/discussions/11613#discussioncomment-5980069

Inline the function while it is only one line, and we have no other need for a source file.

This makes it easier for customers who `#include "google/cloud/storage/testing/mock_client.h"` without linking to `google_cloud_cpp_storage_testing` to adopt the new function.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11708)


scotthart (approved):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11708)* status: 1 of 4 files reviewed, all discussions resolved

PR #11713, opened 2023-05-24
chore(generator): add workaround for protoc_lib defined main

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11713)

### DIFF
--- a/generator/BUILD.bazel
+++ b/generator/BUILD.bazel
@@ -52,7 +52,6 @@ cc_library(
         "//google/cloud/testing_util:google_cloud_cpp_testing_private",
         "@com_google_googleapis//google/longrunning:longrunning_cc_proto",
         "@com_google_googletest//:gtest_main",
-        "@com_google_protobuf//:protoc_lib",
     ],
 )
 
--- a/generator/BUILD.bazel
+++ b/generator/BUILD.bazel
@@ -51,7 +51,7 @@ cc_library(
         "//:common",
         "//google/cloud/testing_util:google_cloud_cpp_testing_private",
         "@com_google_googleapis//google/longrunning:longrunning_cc_proto",
-        "@com_google_googletest//:gtest_main",
+        "@com_google_googletest//:gtest",
     ],
 )
 
@@ -63,7 +63,7 @@ load(":google_cloud_cpp_generator_unit_tests.bzl", "google_cloud_cpp_generator_u
     deps = [
         ":google_cloud_cpp_generator",
         ":google_cloud_cpp_generator_testing",
-        "@com_google_googletest//:gtest_main",
+        "@com_google_googletest//:gtest",
     ],
 ) for test in google_cloud_cpp_generator_unit_tests]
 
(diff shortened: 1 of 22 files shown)


generator/BUILD.bazel
```diff
@@ -52,7 +52,6 @@ cc_library(
         "//google/cloud/testing_util:google_cloud_cpp_testing_private",
         "@com_google_googleapis//google/longrunning:longrunning_cc_proto",
         "@com_google_googletest//:gtest_main",
-        "@com_google_protobuf//:protoc_lib",
```
devbww, line 55: If we can get by without this, doesn't that solve the problem by itself?

generator/generator_test.cc
```diff
@@ -144,3 +144,8 @@ TEST_F(GeneratorTest, GenerateServicesSuccess) {
 }  // namespace generator
 }  // namespace cloud
 }  // namespace google
+
+int main(int argc, char** argv) {
```
devbww, line 148: If we add a `main()`, shouldn't we also be changing a `gtest_main` to a `gtest`?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11713)* status: 0 of 22 files reviewed, 2 unresolved discussions (waiting on @devbww)

___
*[`generator/generator_test.cc` line 148 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11713#gh-1203300801:-NWAVyGv5F4Y1YL3QYJU:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/62d473921d33438d339c5e3a6e85391b65cafcca/generator/generator_test.cc#L148)):*

If we add a `main()`, shouldn't we also be changing a `gtest_main` to a `gtest`?
</blockquote></details>

Done.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11713)* status: 0 of 22 files reviewed, 2 unresolved discussions (waiting on @devbww)

___
*[`generator/BUILD.bazel` line 0 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11713#gh-1203299958:-NWAW3jI0Kxb0TJ4Zu09:b-5zu6rr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/62d473921d33438d339c5e3a6e85391b65cafcca/generator/BUILD.bazel#L0)):*

If we can get by without this, doesn't that solve the problem by itself?
</blockquote></details>

It was a redundant dependency that we already previously depended on.

devbww (approved):
> Reviewed 20 of 21 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11713)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

PR #11723, opened 2023-05-24
impl(generator): emit LRO code for REST rpc with bespoke Operation type

scotthart:
part of the work for #11531 

Adds an additional case in the generator for LRO methods that do not use the `google::longrunning::Operation` type and its associated methods. Existing tests verify that nothing has changed for the `google::longrunning::Operation` paths. Tests for the custom operation type path will be introduced later and will take the form of the generated compute services and integration tests for those services. This is a consequence of needing to hard code some of the substitution variables to compute specific types and services.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723)

### DIFF
--- a/generator/internal/logging_decorator_rest_generator.cc
+++ b/generator/internal/logging_decorator_rest_generator.cc
@@ -51,10 +51,11 @@ Status LoggingDecoratorRestGenerator::GenerateHeader() {
       {vars("stub_rest_header_path"), "google/cloud/internal/rest_context.h",
        "google/cloud/future.h", "google/cloud/tracing_options.h",
        "google/cloud/version.h"});
-  HeaderSystemIncludes(
-      {vars("proto_header_path"),
-       HasLongrunningMethod() ? "google/longrunning/operations.pb.h" : "",
-       "memory", "set", "string"});
+  HeaderSystemIncludes({vars("proto_header_path"),
+                        HasLongrunningMethod()
+                            ? vars("longrunning_operation_include_header")
+                            : "",
+                        "memory", "set", "string"});
 
   auto result = HeaderOpenNamespaces(NamespaceType::kInternal);
   if (!result.ok()) return result;
@@ -74,7 +75,7 @@ class $logging_rest_class_name$ : public $stub_rest_class_name$ {
     if (!HasHttpAnnotation(method)) continue;
     if (IsLongrunningOperation(method)) {
       HeaderPrintMethod(method, __FILE__, __LINE__, R"""(
-  future<StatusOr<google::longrunning::Operation>> Async$method_name$(
+  future<StatusOr<$response_type$>> Async$method_name$(
       google::cloud::CompletionQueue& cq,
       std::unique_ptr<google::cloud::rest_internal::RestContext> rest_context,
       $request_type$ const& request) override;
@@ -118,8 +119,9 @@ class $logging_rest_class_name$ : public $stub_rest_class_name$ {
 
   if (HasLongrunningMethod()) {
     // long running operation support methods
-    HeaderPrint(
-        R"""(
+    if (HasGRPCLongrunningOperation()) {
+      HeaderPrint(
+          R"""(
   future<StatusOr<google::longrunning::Operation>> AsyncGetOperation(
       google::cloud::CompletionQueue& cq,
       std::unique_ptr<google::cloud::rest_internal::RestContext> rest_context,
@@ -130,6 +132,20 @@ class $logging_rest_class_name$ : public $stub_rest_class_name$ {
       std::unique_ptr<google::cloud::rest_internal::RestContext> rest_context,
       google::longrunning::CancelOperationRequest const& request) override;
 )""");
+    } else {
+      HeaderPrint(
+          R"""(
+  future<StatusOr<$longrunning_response_type$>> AsyncGetOperation(
+      google::cloud::CompletionQueue& cq,
+      std::unique_ptr<google::cloud::rest_internal::RestContext> rest_context,
+      $longrunning_get_operation_request_type$ const& request) override;
+
+  future<Status> AsyncCancelOperation(
+      google::cloud::CompletionQueue& cq,
+      std::unique_ptr<google::cloud::rest_internal::RestContext> rest_context,
+      $longrunning_cancel_operation_request_type$ const& request) override;
+)""");
+    }
   }
 
   HeaderPrint(R"""(
@@ -179,7 +195,7 @@ $logging_rest_class_name$::$logging_rest_class_name$(
     if (!HasHttpAnnotation(method)) continue;
     if (IsLongrunningOperation(method)) {
       CcPrintMethod(method, __FILE__, __LINE__, R"""(
-future<StatusOr<google::longrunning::Operation>>
+future<StatusOr<$response_type$>>
 $logging_rest_class_name$::Async$method_name$(
       CompletionQueue& cq,
       std::unique_ptr<rest_internal::RestContext> rest_context,
@@ -266,7 +282,9 @@ $logging_rest_class_name$::Async$method_name$(
 
   if (HasLongrunningMethod()) {
     // long running operation support methods
-    CcPrint(R"""(
+    if (HasGRPCLongrunningOperation()) {
+      CcPrint(
+          R"""(
 future<StatusOr<google::longrunning::Operation>>
 $logging_rest_class_name$::AsyncGetOperation(
     google::cloud::CompletionQueue& cq,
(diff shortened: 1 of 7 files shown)


generator/internal/logging_decorator_rest_generator.cc
```diff
@@ -118,8 +119,9 @@ class $logging_rest_class_name$ : public $stub_rest_class_name$ {
 
   if (HasLongrunningMethod()) {
     // long running operation support methods
-    HeaderPrint(
-        R"""(
+    if (HasGRPCLongrunningOperation()) {
```
dbolduc, line 122: I think this is too much duplication for me.

Does `$longrunning_response_type$ == "google::longrunning::Operation"` for the gRPC ones?

If not, can we just add some new substitution that is:

```cc
$longruning_get_operation_response_type$ = HasGRPCLongrunningOperation() ?
    "google::longrunning::Operation" : $longrunning_response_type$
```

Then we should only need one version of the `AsyncGetOperation` and `AsyncCancelOperation`.

---

Also we are assuming that the service only has one type of LRO. That is, it cannot return `google::longrunning::Operation` in one API and `my::bespoke::Operation` in another. Is that ok?

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @dbolduc)

___
*[`generator/internal/connection_impl_rest_generator.cc` line 300 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723#gh-1205712129:-NWJlsxz0RfFSmBIM31o:b-qj92pf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/69ed72a01ba71d9faedb7b5c3503abd320c231a2/generator/internal/connection_impl_rest_generator.cc#L300)):*

nit: "result. Some operations use the metadata, some use the data. We need to ..."
</blockquote></details>

Reworded.
___
*[`generator/internal/connection_impl_rest_generator.cc` line 314 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723#gh-1205728242:-NWJlvb81csq165b1Z9E:b31h2ts) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/69ed72a01ba71d9faedb7b5c3503abd320c231a2/generator/internal/connection_impl_rest_generator.cc#L314)):*

As it stands today, we could/should just `return op;`. Seems like you have plans to change it though.

If you don't want to change it, we should `std::move(op).status()` below.
</blockquote></details>

Changed to `return op;` until we need it to do something different.
___
*[`generator/internal/connection_impl_rest_generator.cc` line 316 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723#gh-1205713557:-NWJm6kM0NaMAGwLLcc-:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/69ed72a01ba71d9faedb7b5c3503abd320c231a2/generator/internal/connection_impl_rest_generator.cc#L316)):*

I think we should remove these. Leaving the TODO in the generator seems sufficient.
</blockquote></details>

Done.
___
*[`generator/internal/connection_impl_rest_generator.cc` line 331 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723#gh-1205730662:-NWJm7fx-HazytenFGsw:bt1b19p) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/69ed72a01ba71d9faedb7b5c3503abd320c231a2/generator/internal/connection_impl_rest_generator.cc#L331)):*

nit: I would drop the `_name` suffix here and in `cancel_request_set_operation_name`.
</blockquote></details>

Renamed.
___
*[`generator/internal/logging_decorator_rest_generator.cc` line 122 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723#gh-1205969769:-NWJm8qU6tJJf6gUkmkX:b-ckoimy) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/69ed72a01ba71d9faedb7b5c3503abd320c231a2/generator/internal/logging_decorator_rest_generator.cc#L122)):*

I think this is too much duplication for me.

Does `$longrunning_response_type$ == "google::longrunning::Operation"` for the gRPC ones?

If not, can we just add some new substitution that is:

```cc
$longruning_get_operation_response_type$ = HasGRPCLongrunningOperation() ?
    "google::longrunning::Operation" : $longrunning_response_type$
```

Then we should only need one version of the `AsyncGetOperation` and `AsyncCancelOperation`.

---

Also we are assuming that the service only has one type of LRO. That is, it cannot return `google::longrunning::Operation` in one API and `my::bespoke::Operation` in another. Is that ok?
</blockquote></details>

At this point of the code generation pipeline, the substitution variable dictionaries are no longer mutable. But, turns out we can remove some of the duplication because we have the correct values already set for grpc vs non-grpc longrunning operations.

Mixing LRO mechanisms would be weird, and we currently have no evidence that it occurs.

generator/internal/stub_rest_generator.cc
```diff
@@ -115,10 +116,11 @@ class $stub_rest_class_name$ {
...
 
-  // long running operation support methods
   if (HasLongrunningMethod()) {
-    HeaderPrint(
-        R"""(
+    // long running operation support methods
+    if (HasGRPCLongrunningOperation()) {
+      HeaderPrint(
+          R"""(
   virtual future<StatusOr<google::longrunning::Operation>> AsyncGetOperation(
```
dbolduc, line 124: `$longrunning_response_type$` here ?

generator/internal/stub_rest_generator.cc
```diff
@@ -255,14 +270,26 @@ Status StubRestGenerator::GenerateCc() {
...
 
   if (HasLongrunningMethod()) {
-    CcPrint(R"""(
+    if (HasGRPCLongrunningOperation()) {
+      CcPrint(R"""(
 Default$stub_rest_class_name$::Default$stub_rest_class_name$(Options options)
     : service_(rest_internal::MakePooledRestClient(
           options.get<EndpointOption>(), options)),
       operations_(rest_internal::MakePooledRestClient(
           options.get<rest_internal::LongrunningEndpointOption>(), options)),
```
dbolduc, line 279: comment: argh, that is annoying.

generator/internal/stub_rest_generator.cc
```diff
@@ -390,8 +417,9 @@ Default$stub_rest_class_name$::Async$method_name$(
 
   if (HasLongrunningMethod()) {
     // long running operation support methods
-    CcPrint(
-        R"""(
+    if (HasGRPCLongrunningOperation()) {
```
dbolduc, line 420: I see that these are very close. Standardizing it could be a clean up for now / the future.

"google::longrunning::GetOperationRequest" vs. "$longrunning_get_operation_request_type$"
"absl::StrCat("/v1/", request.name())" vs. "$longrunning_get_operation_path$"

dbolduc (approved):

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @dbolduc)

___
*[`generator/internal/stub_rest_generator.cc` line 124 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723#gh-1206018337:-NWJw8oO3zbMjIOhdb16:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c8d61bf29474b972c00ee4d71ca580a80073e189/generator/internal/stub_rest_generator.cc#L124)):*

`$longrunning_response_type$` here ?
</blockquote></details>

Done.
___
*[`generator/internal/stub_rest_generator.cc` line 420 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723#gh-1206029482:-NWJwBco9CEWZdKgcUFQ:bor3bu2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c8d61bf29474b972c00ee4d71ca580a80073e189/generator/internal/stub_rest_generator.cc#L420)):*

I see that these are very close. Standardizing it could be a clean up for now / the future.

"google::longrunning::GetOperationRequest" vs. "$longrunning_get_operation_request_type$"
"absl::StrCat("/v1/", request.name())" vs. "$longrunning_get_operation_path$"
</blockquote></details>

Added changes to longrunning.cc to set these vars for GRPC LRO. Removed the now redundant branch.

dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @scotthart)

___
*[`generator/internal/stub_rest_generator.cc` line 420 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11723#gh-1206029482:-NWJyo2y-QJjf9clIT0W:b-83sjo4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/c8d61bf29474b972c00ee4d71ca580a80073e189/generator/internal/stub_rest_generator.cc#L420)):*

Added changes to longrunning.cc to set these vars for GRPC LRO. Removed the now redundant branch.
</blockquote></details>

Nice.

PR #11725, opened 2023-05-24
refactor(bigtable): future-proof ReadRows mocking

dbolduc:
Part of the work for \<mystery bigtable feature\>

Googlers can see: go/cloud-cxx:how-to-grow-bigtable-read-rows

The new Connection API allows us to add additional `ReadRowsRequest` configuration parameters.

I think we should include `app_profile_id` in the params struct. It is easier to check in the params struct than in `google::cloud::mocks::CurrentOptions()`.

Note that we cannot `internal::ExtractOption<AppProfileIdOption>` in the connection layer, because the current options are ` const&`. I think it is safer to just use `.get<>` instead of changing the method to `Options& internal::CurrentOptions()`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11725)

### DIFF
--- a/google/cloud/bigtable/data_connection.cc
+++ b/google/cloud/bigtable/data_connection.cc
@@ -74,9 +74,17 @@ future<std::vector<FailedMutation>> DataConnection::AsyncBulkApply(
       Status(StatusCode::kUnimplemented, "not-implemented"), mut.size()));
 }
 
-RowReader DataConnection::ReadRows(
-    // NOLINTNEXTLINE(performance-unnecessary-value-param)
-    std::string const&, RowSet, std::int64_t, Filter) {
+RowReader DataConnection::ReadRows(std::string const& table_name,
+                                   RowSet row_set, std::int64_t rows_limit,
+                                   Filter filter) {
+  return ReadRowsFull(ReadRowsParams{
+      std::move(table_name),
+      google::cloud::internal::CurrentOptions().get<AppProfileIdOption>(),
+      std::move(row_set), rows_limit, std::move(filter)});
+}
+
+// NOLINTNEXTLINE(performance-unnecessary-value-param)
+RowReader DataConnection::ReadRowsFull(ReadRowsParams) {
   return MakeRowReader(std::make_shared<bigtable_internal::StatusOnlyRowReader>(
       Status(StatusCode::kUnimplemented, "not implemented")));
 }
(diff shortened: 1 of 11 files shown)


devbww:
> Note that we cannot `internal::ExtractOption<AppProfileIdOption>` in the connection layer, because the current options are  `const&`.

Can you stop the `AppProfileIdOption` from getting into an `OptionsSpan` in the first place?

In some stuff I'm working on with a new "params-destined option", the client calls now look like ...
```
  opts = internal::MergeOptions(std::move(opts), options_);
  auto foo_option = internal::ExtractOption<FooOption>(opts);
  internal::OptionsSpan span(std::move(opts));
  // use foo_option somehow
```

Or, might an existing `ReadRows()` mock depend on seeing the `AppProfileIdOption` in `CurrentOptions()`?  (In my case, `FooOption` is brand new.)

dbolduc:
> Can you stop the `AppProfileIdOption` from getting into an `OptionsSpan` in the first place?

No, I don't think so. I have to keep calling this connection method from the client (which doesn't accept an app profile id):

https://github.com/googleapis/google-cloud-cpp/blob/12600479a816a1a74757ff94f2343774f4e44c4f/google/cloud/bigtable/table.cc#L222-L223

I could reverse the flow of the Connection methods. But then there would be no point to extracting the option. I would have to add it back in.

```cc
Table::ReadRows(..., opts) {
  opts = internal::MergeOptions(std::move(opts), options_);
  auto app_profile_id = internal::ExtractOption<AppProfileIdOption>(opts);
  OptionsSpan o(std::move(opts));
  auto params = ReadRowsParams{...};
  connection_->ReadRowsFull(params);
}

DataConnection::ReadRowsFull(ReadRowsParams params) {
  auto options = internal::CurrentOptions();
  options.set<AppProfileIdOption>(params.app_profile_id);
  OptionsSpan o(std::move(options));
  return ReadRows(params.foo, params.bar, params.baz, params.etc);
}
```

> Might an existing `ReadRows()` mock depend on seeing the `AppProfileIdOption` in `CurrentOptions()`?

Yeah, that can also be a problem.

> In some stuff I'm working on with a new "params-destined option", the client calls now look like ...

Yeah, that looks right to me. I just can't use it in `bigtable` for backwards compatibility reasons. :(

devbww (approved):

PR #11741, opened 2023-05-27
feat(storage): better error messages in signed URLs

### LINKED ISSUE #11740: Bad error message when Signing URLs without a service account email
The storage client can created signed URLs.  These are URLs that can access GCS resources via a *very* specific request without the need for additional authentication.  This involves creating a the request in a canonical form, hashing it, and signing the request and the hash.

When the credentials are obtained from a service account key file, the client library can sign the canonical request plus hash (it is just treated as a blob at the signing stage) locally, using the private key from the key file.

When the credentials are not a service account key file, the client library can use [IAM signBlob](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signBlob) to sign the blob.

In the latter case it needs the service account email of the request to sign. For some credential types there is a default email associated with the credentials.  But sometimes there is none, or the value cannot be retrieved, for example, the metadata service may be unresponsive.

Without a default value for the service account email, the client library sends a request to `IAM signBlob` with an empty email address.  This results in a 404 error without additional details.

This is a case where the client library should not rely on the service to return an error.  We should make an exception to our [ADR](https://github.com/googleapis/google-cloud-cpp/blob/main/doc/adr/2018-06-19-do-not-duplicate-server-side-validation.md) and return a more meaningful error.  An empty string here will never succeed, it is a common error, and the returned error lacks enough information to troubleshoot effectively.

coryan:
The error messages for signed URL functions (`CreateV2SignedUrl()`, `CreateV4SignedUrl()` and `CreatedSignedPolicyDocument()`) were not helpful when the signing account is missing.

Fixes #11740

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11741)

### DIFF
--- a/google/cloud/storage/client.cc
+++ b/google/cloud/storage/client.cc
@@ -358,7 +358,6 @@ StatusOr<Client::SignBlobResponseRaw> Client::SignBlobImpl(
     SigningAccount const& signing_account, std::string const& string_to_sign) {
   auto credentials = raw_client_->client_options().credentials();
 
-  std::string signing_account_email = SigningEmail(signing_account);
   // First try to sign locally.
   auto signed_blob = credentials->SignBlob(signing_account, string_to_sign);
   if (signed_blob) {
@@ -370,7 +369,8 @@ StatusOr<Client::SignBlobResponseRaw> Client::SignBlobImpl(
   // credentials account. In either case, try to sign using the API.
   // In this case, however, we want to validate the signing account, because
   // otherwise the errors are almost impossible to troubleshoot.
-  if (signing_account_email.empty()) {
+  auto signing_email = SigningEmail(signing_account);
+  if (signing_email.empty()) {
     return google::cloud::internal::InvalidArgumentError(
         "signing account cannot be empty."
         " The client library was unable to fetch a valid signing email from"
@@ -379,7 +379,7 @@ StatusOr<Client::SignBlobResponseRaw> Client::SignBlobImpl(
         GCP_ERROR_INFO());
   }
   internal::SignBlobRequest sign_request(
-      signing_account_email, internal::Base64Encode(string_to_sign), {});
+      std::move(signing_email), internal::Base64Encode(string_to_sign), {});
   auto response = raw_client_->SignBlob(sign_request);
   if (!response) return response.status();
   auto decoded = internal::Base64Decode(response->signed_blob);
--- a/google/cloud/storage/client.cc
+++ b/google/cloud/storage/client.cc
@@ -22,6 +22,7 @@
 #include "google/cloud/internal/algorithm.h"
 #include "google/cloud/internal/curl_options.h"
 #include "google/cloud/internal/filesystem.h"
+#include "google/cloud/internal/make_status.h"
 #include "google/cloud/internal/opentelemetry.h"
 #include "google/cloud/log.h"
 #include <fstream>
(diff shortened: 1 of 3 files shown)


dbolduc (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11741)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/storage/client.cc` line 361 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11741#-NWU0x3-DDVauxZrqzqQ:-NWU0x3-DDVauxZrqzqR:binq7ho) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d41f218d535f5cbc82cb99d74ea1498a5f73848e/google/cloud/storage/client.cc#L361)):*
> ```cpp
>   std::string signing_account_email = SigningEmail(signing_account);
> ```

Looks like we should move this after the "sign locally" attempt.

```cc
  // First try to sign locally.
  auto signed_blob = credentials->SignBlob(signing_account, string_to_sign);
  if (signed_blob) {
    return SignBlobResponseRaw{credentials->KeyId(), *std::move(signed_blob)};
  }

  // If signing locally fails that may be because the credentials do not
  // support signing, or because the signing account is different than the
  // credentials account. In either case, try to sign using the API.
  std::string signing_account_email = SigningEmail(signing_account);
  if (signing_account_email.empty()) {
    return google::cloud::internal::InvalidArgumentError(
        "signing account cannot be empty."
        " The client library was unable to fetch a valid signing email from"
        " the configured credentials, and the application did not provide"
        " a value in the `google::cloud::storage::SigningAccount` option.",
        GCP_ERROR_INFO());
  }

  internal::SignBlobRequest sign_request(
      signing_account_email, internal::Base64Encode(string_to_sign), {});
```

coryan (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11741)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @dbolduc)

___
*[`google/cloud/storage/client.cc` line 361 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11741#-NWU0x3-DDVauxZrqzqQ:-NWUHEqF4IMlpurSHvxJ:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/d41f218d535f5cbc82cb99d74ea1498a5f73848e/google/cloud/storage/client.cc#L361)):*

Looks like we should move this after the "sign locally" attempt.

```
  // First try to sign locally.
  auto signed_blob = credentials->SignBlob(signing_account, string_to_sign);
  if (signed_blob) {
    return SignBlobResponseRaw{credentials->KeyId(), *std::move(signed_blob)};
  }

  // If signing locally fails that may be because the credentials do not
  // support signing, or because the signing account is different than the
  // credentials account. In either case, try to sign using the API.
  std::string signing_account_email = SigningEmail(signing_account);
  if (signing_account_email.empty()) {
    return google::cloud::internal::InvalidArgumentError(
        "signing account cannot be empty."
        " The client library was unable to fetch a valid signing email from"
        " the configured credentials, and the application did not provide"
        " a value in the `google::cloud::storage::SigningAccount` option.",
        GCP_ERROR_INFO());
  }

  internal::SignBlobRequest sign_request(
      signing_account_email, internal::Base64Encode(string_to_sign), {});
```
</blockquote></details>

Done.

PR #11753, opened 2023-05-31
impl(bigquery): InsertJob api prep work - Part 1

jsrinnn:
InsertJob api requires additional fields to be returned in the `Job` response. 

This PR and subsequent PRs will add the missing fields and info to Custom `Job` classes, as needed by the InsertJob api. 
These were not added in the initial PRs for the `Job` resource, as they were not needed for the read-only apis.

I have divided the above changes into 4 PRs. The first two will add the missing fields and sub classes. The remaining two will mainly add the DebugString() member methods and unit tests, for the logging decorator .

This PR includes the following changes:
 
 1) Adds new `JobConfigurationQuery` class with following sub classes:

- SystemVariables
- StandardSqlField
- StandardSqlDataType
- StandardSqlStructType
- Struct
- Value

  Fields are recursive and self-refrential. Protobuf `oneof` functionality is simulated by `absl::variant`.
  
2) Modifies `JobConfiguration` class with the query class abopve 
  
3) Json Parsers and Unit Tests for all.


coryan (changes requested):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753)* status: all files reviewed, 10 unresolved discussions (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h` line 136 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWjZ7Rn8WdGiqcVcoPF:-NWjZ7Rn8WdGiqcVcoPG:b-r7fs90) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h#L136)):*
> ```objc
> NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(TypeKind, value);
> 
> // Self referential and circular structures below is based on the bigquery v2
> ```

These types got to the level of complexity where they need longer comments explaining how they work.  Try to answer questions like:

- Why do we need these in the first place?  Why not use nlohmann::json directly?
- Link the definition of the JSON objects they represent, e.g. https://cloud.google.com/bigquery/docs/reference/rest/v2/StandardSqlDataType
- A brief explanation about why the `TypeKind` thing is needed, i.e., how different `string` values on the wire represent different types.
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h` line 163 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWj_6TwCqWcFOjchBXU:-NWj_6TwCqWcFOjchBXV:b-uwkaps) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h#L163)):*
> ```objc
>                 StandardSqlStructType>
>       sub_type;
> };
> ```

These variants could use a name (I am terrible at naming, so don't feel tied to `ValueType`):

```cc
struct StandardSqlDataType {
  using ValueType = absl::variant<absl::monostate, std::shared_ptr<StandardSqlDataType>,
                StandardSqlStructType>;

  TypeKind type_kind;
   ValueType sub_type;
};
```



_Code quote:_
```objc
struct StandardSqlDataType {
  TypeKind type_kind;

  absl::variant<absl::monostate, std::shared_ptr<StandardSqlDataType>,
                StandardSqlStructType>
      sub_type;
};
```
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h` line 171 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWj_W8HCL4GHu9O1gpu:-NWj_W8HCL4GHu9O1gpv:b-lq0ipp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h#L171)):*
> ```objc
> 
> struct Value {
>   absl::variant<absl::monostate, double, std::string, bool,
> ```

These variant instantiation also could use a name.
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h` line 173 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWjSISQ4U3q5FsCNWD7:-NWjSISQ4U3q5FsCNWD8:b-k43nxb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h#L173)):*
> ```objc
>   absl::variant<absl::monostate, double, std::string, bool,
>                 std::shared_ptr<Struct>, std::vector<Value>>
>       kind;
> ```

nit: is `kind` a good name?  Normally that is reserved for the name of the JSON "type"?
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h` line 185 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWjZ0O92akWpJV1qxhu:-NWjZ0O92akWpJV1qxhv:bc76hp2) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h#L185)):*
> ```objc
> void from_json(nlohmann::json const& j, Struct& s);
> 
> struct SystemVariables {
> ```

What are `SystemVariables`? 
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 134 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWjV7wM-dHvb4uFz3sA:-NWjV7wM-dHvb4uFz3sB:b-eb3afw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L134)):*
> ```cpp
>       }
>     }
>   }
> ```

FYI: I think a visitor would be more idiomatic here (and in other places). Basically you want to decompose the variant using functions instead of `std:;get<>`:

```cc
void to_json(nlohmann::json& j, StandardSqlDataType const& t) {
  if (t.sub_type.valueless_by_exception()) {
    t = nlohmann::json{{"type_kind", t.type_kind}};
    return;
  }

  struct Visitor {
    std::string type_kind;
    nlohmann::json& j;

    void operator()(absl::monostate const&) { j = nlohman::json{{"type_kind", std::move(type_kind)}}; }
    void operator()(std::shared_ptr<StandardSqlDataType> const& type) {
      j = nlohmann::json{{"type_kind", std::move(type_kind)}, {"sub_type", *type}, {"sub_type_index", 1}};
    }
    void operator()(StandardSqlStructType const& type) {
      j = nlohmann::json{{"type_kind", std::move(type_kind)}, {"sub_type", type}, {"sub_type_index", 2}};
    }
  };
  absl::visit(t.sub_type, Visitor{t.type_kind, j});
}
```



_Code quote:_
```cpp
  if (t.sub_type.valueless_by_exception()) {
    j = nlohmann::json{{"type_kind", t.type_kind}};
  } else {
    auto index = t.sub_type.index();
    switch (index) {
      case 1: {
        auto type = absl::get<1>(t.sub_type);
        if (type != nullptr) {
          j = nlohmann::json{{"type_kind", t.type_kind},
                             {"sub_type", *type},
                             {"sub_type_index", index}};
        }
        break;
      }
      case 2: {
        auto type = absl::get<2>(t.sub_type);
        j = nlohmann::json{{"type_kind", t.type_kind},
                           {"sub_type", type},
                           {"sub_type_index", index}};
        break;
      }
      default: {
        j = nlohmann::json{{"type_kind", t.type_kind}};
        break;
      }
    }
  }
```
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 155 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWjXowL-opb7jGYQ0em:-NWjXowL-opb7jGYQ0en:bqta2cp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L155)):*
> ```cpp
> 
> void to_json(nlohmann::json& j, Value const& v) {
>   if (!v.kind.valueless_by_exception()) {
> ```

Likewise, a visitor may be more idiomatic here.
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 254 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWjY9nnETiw1Ermq5IA:-NWjY9nnETiw1Ermq5IB:be3eu5m) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L254)):*
> ```cpp
>       auto lhs_ptr = absl::get<1>(lhs.sub_type);
>       auto rhs_ptr = absl::get<1>(rhs.sub_type);
>       if ((lhs_ptr != nullptr && rhs_ptr == nullptr) ||
> ```

I think the `operator==()` function could be simplified, here and below (more details below).  Thoughts?
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 292 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWjU7kVDSvAsFjwdxRk:-NWjU7kVDSvAsFjwdxRl:b-hbg35o) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L292)):*
> ```cpp
>     return false;
>   }
>   auto index = lhs.kind.index();
> ```

The use of `kind.index()` instead of `absl::visit` is not very idiomatic, but maybe simpler than the more idiomatic alternatives.
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 299 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWjSx2KEGT2_PvvzcUn:-NWjSx2L3nR_u4m8nq0S:bhrdtkv) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L299)):*
> ```cpp
>       if (lhs_val != rhs_val) {
>         return false;
>       }
> ```

I would write simply:

```cc
return absl::get<1>(lhs.kind) == absl::get<1>(rhs.kind);
```



_Code quote:_
```cpp
      auto lhs_val = absl::get<1>(lhs.kind);
      auto rhs_val = absl::get<1>(rhs.kind);
      if (lhs_val != rhs_val) {
        return false;
      }
```
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 321 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWjTHLBE-XjMOHoJlJy:-NWjTHLBE-XjMOHoJlJz:b-rch3aq) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/33e81bd6158ac2d9761480f260fa54386334b6e3/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L321)):*
> ```cpp
>       auto lhs_val = absl::get<4>(lhs.kind);
>       auto rhs_val = absl::get<4>(rhs.kind);
>       if ((lhs_val != nullptr && rhs_val == nullptr) ||
> ```

Consider:

```cc
if (lhs_val == nullptr && rhs_val == nullptr) return true;
if ((lhs_val == nullptr) != (rhs_val == nullptr)) return false;
return std::equal(lhs_val->fields.begin(), lhs_val->fields.end(), rhs_val->fields.begin());
```

it can be simplified further I think.

jsrinnn (commented):
Addressed feedback with some follow-up questions/comments.

Thanks!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-:-NWrKYM0-T5IqU584PgH:b-drejvg)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 208 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWnouAd3EpWfQeZLlz5:-NWrF__U95q0SvNs_m3b:b-nfk9mg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/fedfae7fc1f42c0e3bb9b3b7876a7fa37187efb0/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L208)):*

FYI: it is a bit odd that no errors are created in this case.
</blockquote></details>

Hmm...I got the impression we need to be exception less i.e. cannot throw exceptions. Given the function is void I let it be handled as a noop.

I have added a log statement here but let me know if a throw is more appropriate and I can add that.
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 245 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWnpYZ6D5jYuuVVkoT1:-NWrHGfS5sqGv3YCPNP0:b-f0f2sb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/fedfae7fc1f42c0e3bb9b3b7876a7fa37187efb0/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L245)):*

nit: is this correct?  Does it make sense to you?

```cc
bool operator==(StandardSqlField const& lhs, StandardSqlField const& rhs) {
  auto const eq_name = (lhs.name == rhs.name);
  if (lhs.type != nullptr && rhs.type != nullptr) {
    return eq_name && (*lhs.type == *rhs.type);
  }
  return eq_name && lhs.type == nullptr && rhs.type == nullptr;
}
```

</blockquote></details>

So the alternative suggested is doing the same thing as the original code. 

Is the suggestion mainly for readability? The semantics are the same for both.

FYI: I have incorporated the suggestion.

coryan (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-:-NWrVOalApDHtbTpwF34:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 208 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWnouAd3EpWfQeZLlz5:-NWrTUr8Byjt2RMQWVV4:bf92536) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/fedfae7fc1f42c0e3bb9b3b7876a7fa37187efb0/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L208)):*

Hmm...I got the impression we need to be exception less i.e. cannot throw exceptions. Given the function is void I let it be handled as a noop.

I have added a log statement here but let me know if a throw is more appropriate and I can add that.
</blockquote></details>

The change is fine.  I was happy with the code as it was too.  My comment was just a discussion starter.

One of the "problems" of the `nlohmann::json` library is that it assumes standard C++.  That means with exceptions, there is no "C++ without exceptions" standard.  OTOH, we need to support at least one large customer that turns off exceptions (their name starts with "Goo", and ends with "gle").

Logging might be fine.  Using `GCP_LOG(FATAL)` (that is analogous to a `CHECK()`) may work too.  Maybe some kind of validation after the parsing exits the nlohmann::json functions?  Maybe declare that the ODBC driver will require exceptions and then use something like https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/internal/throw_delegate.h) to throw or crash. I am not sure what is the right answer, just thought you may want to start thinking about it.
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 245 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWnpYZ6D5jYuuVVkoT1:-NWrVCjeEh706lHlcAco:bv7s7qn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/fedfae7fc1f42c0e3bb9b3b7876a7fa37187efb0/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L245)):*

So the alternative suggested is doing the same thing as the original code. 

Is the suggestion mainly for readability? The semantics are the same for both.

FYI: I have incorporated the suggestion.
</blockquote></details>

Yeah, it was just readability. It is important that it is readable for you too, thus my question.  Thanks for making the change.
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 209 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWrTiAH-mJrwb7M_RRW:-NWrTiAH-mJrwb7M_RRX:by7wyfe) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/6fbf0cd3cdf03bfcd9396c0fd1101988cf3956db/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L209)):*
> ```cpp
>         break;
>       default:
>         GCP_LOG(ERROR) << "Invalid kind_idex for Value: " << index;
> ```

nit: typo in `idex` vs. `index`

jsrinnn (commented):
Addressed latest feedback.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-:-NWrXApN1EbSswBfFVrM:b-8c7hvi)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 208 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWnouAd3EpWfQeZLlz5:-NWrWoeg09xd0E1h7-wU:bwse7l1) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/fedfae7fc1f42c0e3bb9b3b7876a7fa37187efb0/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L208)):*

The change is fine.  I was happy with the code as it was too.  My comment was just a discussion starter.

One of the "problems" of the `nlohmann::json` library is that it assumes standard C++.  That means with exceptions, there is no "C++ without exceptions" standard.  OTOH, we need to support at least one large customer that turns off exceptions (their name starts with "Goo", and ends with "gle").

Logging might be fine.  Using `GCP_LOG(FATAL)` (that is analogous to a `CHECK()`) may work too.  Maybe some kind of validation after the parsing exits the nlohmann::json functions?  Maybe declare that the ODBC driver will require exceptions and then use something like https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/internal/throw_delegate.h) to throw or crash. I am not sure what is the right answer, just thought you may want to start thinking about it.

</blockquote></details>

Ack. Thanks. Yup, the caller would need some parsing validation after calling this.

Modified it to a GCP_LOG(FATAL)
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 209 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11753#-NWrTiAH-mJrwb7M_RRW:-NWrX7n98815K3vDt-hC:b2cz0ny) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/83ed7f028423f651b478e2220b510579b534be6a/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L209)):*

nit: typo in `idex` vs. `index`
</blockquote></details>

Done.

Thanks.

Issue #11759, opened 2023-06-01
Read in schema file for pubsub samples


alevenberg:
Pubsub currently uses hardcoded strings for the .avsc and .proto files for schema creation. The other libraries read in files (https://cloud.google.com/pubsub/docs/schemas#c++). 

This will allow for more flexibility for someone getting started (can just modify the existing file to create the schema they want) and add consistency between the other client libraries.

The schemas can be created using files of either json (.avsc) or proto (.pb).

PR #11770, opened 2023-06-02
refactor(pubsub): add file parameters for the create/validate schema commands

alevenberg:
Issue #11759

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770)


google/cloud/pubsub/samples/pubsub_samples_common.cc
```diff
@@ -159,6 +160,19 @@ std::string RandomSchemaId(google::cloud::internal::DefaultPRNG& generator) {
...
+    return google::cloud::Status{google::cloud::StatusCode::kNotFound,
+                                 "can't find file: " + path};
+  }
+
+  std::stringstream buffer;
+  buffer << ifs.rdbuf();
+  ifs.close();
+  return buffer.str();
+}
+
```
coryan, line 175: We use exceptions in samples, you can "simplify" this code to something like:

```suggestion
std::string ReadFile(std::string const& path) {
  std::ifstream ifs(path);
  if (!ifs.is_open()) throw std::runtime_error("Cannot open file: " + path);
  ifs.exceptions(std::ios::badbit);
  return std::string{std::istreambuf_iterator<char>{ifs.rdbuf()}, {}};
}
```

That will make the samples themselves easier to grok, because they will need less error-handling code.

google/cloud/pubsub/samples/samples.cc
```diff
@@ -780,30 +785,15 @@ void CreateAvroSchema(google::cloud::pubsub::SchemaServiceClient client,
...
-        },
-        {
-          "name": "post_abbr",
-          "type": "string",
-          "doc": "The postal code abbreviation of the state."
-        }
-      ]
-    })js";
+     std::string const& schema_id, std::string const& schema_definition_file) {
+    auto definition = ReadFile(schema_definition_file);
```
coryan, line 789: FYI (no action required): in general we avoid subroutines in sample code.  Somebody reading the code needs to figure out what the subroutine does and that distracts from the sample. These are pretty small, and it is easy to figure out what they do (almost).  But keep this in mind in future samples.

This is actionable: the one thing that is not obvious in this case is the return type.  A customer reading *just* the code snippet will need to wonder what is the return type for `ReadFile()`, and why it needs a `if (!definition)`.

Consider the suggestion to use exceptions and maybe change the code here to:

```cc
std::string const definition = ReadFile(schema_definition_file);
```

google/cloud/pubsub/samples/samples.cc
```diff
@@ -780,30 +785,15 @@ void CreateAvroSchema(google::cloud::pubsub::SchemaServiceClient client,
...
+    if (!definition) throw std::move(definition).status();
+
     google::pubsub::v1::CreateSchemaRequest request;
     request.set_parent(google::cloud::Project(project_id).FullName());
     request.set_schema_id(schema_id);
     request.mutable_schema()->set_type(google::pubsub::v1::Schema::AVRO);
-    request.mutable_schema()->set_definition(kDefinition);
+    request.mutable_schema()->set_definition(definition.value());
```
coryan, line 796: Once we check a value for a `StatusOr<>` we prefer (and recommend) `*definition`.

google/cloud/pubsub/samples/samples.cc
```diff
@@ -912,137 +896,105 @@ void ValidateAvroSchema(google::cloud::pubsub::SchemaServiceClient client,
...
-  [](pubsub::SchemaServiceClient client, std::string const& project_id) {
+  [](pubsub::SchemaServiceClient client, std::string const& project_id,
+     std::string const& schema_definition_file,
+     std::string const& message_file) {
+    auto definition = ReadFile(schema_definition_file);
+    if (!definition) throw std::move(definition).status();
+
+    auto message_file_contents = ReadFile(message_file);
+    if (!message_file_contents) throw std::move(message_file_contents).status();
     google::cloud::pubsub::samples::State data;
```
coryan, line 977: So this example loads the `.proto` and `.textproto` files, but can only work if both use a `google::cloud::pubsub::samples::State` message?

Seems like we either allow for any message type, or we don't bother with the additional flexibility?

coryan (changes requested):
Some suggestions to (I believe) make the samples more readable.

There is something odd in the use of `google::cloud::pubsub::samples::State`

alevenberg (commented):
Not sure what you mean by "something odd"

Addressed the other comments

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#-:-NXQ__1zD219EOVU1neF:b-1z6tvj)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/pubsub/samples/pubsub_samples_common.cc` line 175 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#gh-1223065575:-NXQ_6cB1x0XN5F0X9ce:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2cb9660afa5a6d7ebdd0105f6d9cdd8bc6e9cb00/google/cloud/pubsub/samples/pubsub_samples_common.cc#L175)):*

We use exceptions in samples, you can "simplify" this code to something like:

```suggestion
std::string ReadFile(std::string const& path) {
  std::ifstream ifs(path);
  if (!ifs.is_open()) throw std::runtime_error("Cannot open file: " + path);
  ifs.exceptions(std::ios::badbit);
  return std::string{std::istreambuf_iterator<char>{ifs.rdbuf()}, {}};
}
```

That will make the samples themselves easier to grok, because they will need less error-handling code.

</blockquote></details>

Done
___
*[`google/cloud/pubsub/samples/samples.cc` line 789 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#gh-1223072402:-NXQZziU9cP4IJYXVlbP:b-cszsw0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2cb9660afa5a6d7ebdd0105f6d9cdd8bc6e9cb00/google/cloud/pubsub/samples/samples.cc#L789)):*

FYI (no action required): in general we avoid subroutines in sample code.  Somebody reading the code needs to figure out what the subroutine does and that distracts from the sample. These are pretty small, and it is easy to figure out what they do (almost).  But keep this in mind in future samples.

This is actionable: the one thing that is not obvious in this case is the return type.  A customer reading *just* the code snippet will need to wonder what is the return type for `ReadFile()`, and why it needs a `if (!definition)`.

Consider the suggestion to use exceptions and maybe change the code here to:

```cc
std::string const definition = ReadFile(schema_definition_file);
```


</blockquote></details>

Modified.
___
*[`google/cloud/pubsub/samples/samples.cc` line 796 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#gh-1223073137:-NXQ_-Z3E5F7SEDc5wGD:b-q2x4ch) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2cb9660afa5a6d7ebdd0105f6d9cdd8bc6e9cb00/google/cloud/pubsub/samples/samples.cc#L796)):*

Once we check a value for a `StatusOr<>` we prefer (and recommend) `*definition`.
</blockquote></details>

Removed `StatusOr`, but ack for future reference
___
*[`google/cloud/pubsub/samples/samples.cc` line 977 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#gh-1223077221:-NXQW8UE4SYd2ILVmLgz:bmlgmzx) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2cb9660afa5a6d7ebdd0105f6d9cdd8bc6e9cb00/google/cloud/pubsub/samples/samples.cc#L977)):*

So this example loads the `.proto` and `.textproto` files, but can only work if both use a `google::cloud::pubsub::samples::State` message?

Seems like we either allow for any message type, or we don't bother with the additional flexibility?
</blockquote></details>

It looks like you need to have the proto as a dependency at compile time. You can't load an arbitrary textproto into an Any message. 

Removing the proto reading for the commands that will only work with the state.proto: ValidateMessageProtobuf and ValidateMessageNamedSchema
___
*[`google/cloud/pubsub/samples/samples.cc` line 1009 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#gh-1223078074:-NXQZyrY7LErYE13ESf9:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2cb9660afa5a6d7ebdd0105f6d9cdd8bc6e9cb00/google/cloud/pubsub/samples/samples.cc#L1009)):*

Same thing as above.
</blockquote></details>

Done.

coryan (changes requested):
> Not sure what you mean by "something odd"

The fact that a couple of samples only work with `google::cloud::pubsub::samples::State`, but give the impression that they would work with arbitrary files.

> Reviewed 5 of 9 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#-:-NXQcK8J41mpF9RHf-6l:bsrc6q8)* status: all files reviewed, 2 unresolved discussions (waiting on @alevenberg)

___
*[`google/cloud/pubsub/samples/samples.cc` line 977 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#gh-1223077221:-NXQehLa1jQOq6i3Wa_x:b-5aby3b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2cb9660afa5a6d7ebdd0105f6d9cdd8bc6e9cb00/google/cloud/pubsub/samples/samples.cc#L977)):*

It looks like you need to have the proto as a dependency at compile time. You can't load an arbitrary textproto into an Any message. 

Removing the proto reading for the commands that will only work with the state.proto: ValidateMessageProtobuf and ValidateMessageNamedSchema
</blockquote></details>

Maybe use a file for the textproto and use something like:

```cc
auto const definition = data.Descriptor()->FileDescriptor()->DebugString();
```

for the proto definition?  Does that work? 

Alternatively, we can just load the *bytes* from a binary proto file and **not** load them into an object of type `google::cloud::pubsub::samples::State`.  We don't even need to have a valid binary file (though that would be nice), we are trying to show how to use the API.

Or maybe we can just use a JSON-encoded protobuf file and use that, no need to go through a `google::cloud::pubsub::samples::State` object in that case either.


Check my math, I am just cut&pasting from https://stackoverflow.com/questions/19418655/restoring-proto-file-from-descriptor-string-possible
___
*[`google/cloud/pubsub/samples/testdata/valid_message.avsc` line 2 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#-NXQhfX-4BblncCT5PCR:-NXQhfX-4BblncCT5PCS:b948bgf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/94718af7e3c244bc58901d973a723805abb8a876/google/cloud/pubsub/samples/testdata/valid_message.avsc#L2)):*
> ```JSON
> {
>    "name": "New York",
> ```

nit: formatting

alevenberg (commented):
There are still 2 that have the State proto hardcoded in: SubscribeProtobufRecords, PublishProtobufRecords, but I'm making that out of scope for this PR

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#-:-NXQqa8k2cwhjh6G_hwv:b-bzu37i)* status: 6 of 10 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/pubsub/samples/samples.cc` line 977 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#gh-1223077221:-NXRC7bh7AQOORZP9ZMm:bx71tmg) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/2cb9660afa5a6d7ebdd0105f6d9cdd8bc6e9cb00/google/cloud/pubsub/samples/samples.cc#L977)):*

Maybe use a file for the textproto and use something like:

```cc
auto const definition = data.Descriptor()->FileDescriptor()->DebugString();
```

for the proto definition?  Does that work? 

Alternatively, we can just load the *bytes* from a binary proto file and **not** load them into an object of type `google::cloud::pubsub::samples::State`.  We don't even need to have a valid binary file (though that would be nice), we are trying to show how to use the API.

Or maybe we can just use a JSON-encoded protobuf file and use that, no need to go through a `google::cloud::pubsub::samples::State` object in that case either.


Check my math, I am just cut&pasting from https://stackoverflow.com/questions/19418655/restoring-proto-file-from-descriptor-string-possible
</blockquote></details>

The definition file works since it's just reading the file into a text string.

For the message file, I generated the binary version using protoc
`cat google/cloud/pubsub/samples/testdata/valid_message.textproto | protoc --encode="google.cloud.pubsub.samples.State" google/cloud/pubsub/samples/testdata/schema.proto > google/cloud/pubsub/samples/testdata/valid_message.pb`

And ran to test
`bazel run //google/cloud/pubsub/samples:samples validate-message-protobuf alevenb-test google/cloud/pubsub/samples/testdata/schema.proto google/cloud/pubsub/samples/testdata/valid_message.pb`

`bazel run //google/cloud/pubsub/samples:samples validate-message-named-schema alevenb-test state google/cloud/pubsub/samples/testdata/valid_message.pb`

coryan (approved):
:lgtm:

> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#-:-NXRH5Id756VOcHhoT55:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alevenberg)

alevenberg (commented):
Yes that is correct the `.pb` is a binary file -- thanks for offering to make the change to checkers.sh! I reverted the new line in the pb file. 

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770#-:-NXRPlg82VUh20_KJs22:b-ygfudy)* status: 9 of 10 files reviewed, all discussions resolved (waiting on @coryan)

coryan (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11770)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alevenberg)

devbww:
> Yes that is correct the `.pb` is a binary file -- thanks for offering to make the change to checkers.sh! I reverted the new line in the pb file.

The `checkers.sh` change is in, so hopefully now you should be able to rebase and have it pass.

Issue #11775, opened 2023-06-03
Support OpenTelemetry resource detection


dbolduc:
We need to:

- [x] Add a GCP resource detector
- [ ] Include the resource detector in the basic tracing configuration
- [x] Implement `SetResource` for the TraceExporter
  - [x] Copy resource attributes -> span attributes
  - [x] Add monitored resource attributes

PR #11776, opened 2023-06-03
feat(otel): add resource detector

dbolduc:
Part of the work for #11775 

If this is too big, I can try to split it up.

Use a rest client to query the metadata server. The contents of the response + the environment variables determine which resource attributes we set.

The logic is copied from the go implementation:
* https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/detectors/gcp/detector.go
* https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/gcp/detector.go

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776)


google/cloud/opentelemetry/resource_detector.h
```diff
@@ -0,0 +1,49 @@
...
+/// Make an OpenTelemetry Resource Detector for Google Cloud Platform.
+std::unique_ptr<opentelemetry::sdk::resource::ResourceDetector>
+MakeResourceDetector(Options options = {});
+
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_END
+}  // namespace otel
+namespace otel_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+using HttpClientFactory =
```
coryan, line 38: nit-ish-maybe-it-matters: opening the `*_internal` namespaces in a public header is a smell.  This probably belongs in `google/cloud/opentelemetry/internal/make_resource_detector.h`

google/cloud/opentelemetry/resource_detector.cc
```diff
@@ -0,0 +1,241 @@
...
+namespace sc = opentelemetry::sdk::resource::SemanticConventions;
+
+// The metadata server returns fully qualified names. (e.g. a zone may be
+// "projects/p/zones/us-central1-a"). Return the IDs only.
+std::string Tail(std::string const& value) {
+  auto const pos = value.rfind('/');
+  return value.substr(pos + 1);
+}
+
+std::string MD(nlohmann::json const& json,
```
coryan, line 40: The function name is not great.  `FindRecursive()`?

google/cloud/opentelemetry/resource_detector.cc
```diff
@@ -0,0 +1,241 @@
...
+               std::vector<std::string> const& keys) {
+  auto const* j = &json;
+  for (auto const& k : keys) {
+    if (!j->contains(k)) return {};
```
coryan, line 44: I think here you want something like:

```cc
auto const f = j->find(k);
if (f == j->end()) return {};
j = &(*f);
```

But even that is a bit weird and obscures what you are doing.  Maybe:

```cc
std::string FindRecursive(nlohmann::json const& json, std::deque<std::string> keys) {
  if (!keys.empty()) {
    auto const l = json.find(keys.front());
    if (l ==json.end()) return {};
    keys.pop_front();
    return FindRecursive(*l, std::move(keys));
  }
  if (json.is_string()) return json.get<std::string>();
  if (json.is_number()) return std::to_string(json.get<std::int64_t>());
  return {};
}
```

google/cloud/opentelemetry/resource_detector.cc
```diff
@@ -0,0 +1,241 @@
...
+
+    auto log_error = [](Status const& status) {
+      GCP_LOG(INFO) << "Could not query the metadata server. status=" << status;
+      return opentelemetry::sdk::resource::Resource::GetEmpty();
+    };
+
+    if (!response) {
+      auto status = std::move(response).status();
+      if (status.code() == StatusCode::kUnavailable) {
+        // This is most likely a "curl: (6) Could not resolve host" error, which
```
coryan, line 166: If only... GKE is notorious for returning kUnavailable.  You need a retry loop.

google/cloud/opentelemetry/resource_detector.cc
```diff
@@ -0,0 +1,241 @@
...
+    if (!response) return log_error(std::move(response).status());
+    if (rest_internal::IsHttpError(**response)) {
+      auto status = rest_internal::AsStatus(std::move(**response));
+      return log_error(status);
+    }
+
+    auto payload =
+        rest_internal::ReadAll(std::move(**response).ExtractPayload());
+    if (!payload.ok()) return log_error(payload.status());
+    auto json = nlohmann::json::parse(*payload, nullptr, false);
```
coryan, line 181: You need more validation.  First, we should check that the headers returned matched our expectations.  Second, some people have ISPs that return HTML (full of adds) when you send a HTTP request to an unknown hostname.

google/cloud/opentelemetry/resource_detector.cc
coryan: I think you can assume that a successful parse is "good forever".  A failed attempt should be retried after a few minutes.

google/cloud/opentelemetry/resource_detector.cc
```diff
@@ -0,0 +1,241 @@
...
+
+    opentelemetry::sdk::resource::ResourceAttributes attributes;
+    attributes.SetAttribute(sc::kCloudProvider, "gcp");
+    SetAttribute(attributes, sc::kCloudAccountId,
+                 MD(json, {"project", "projectId"}));
+
+    switch (DetectCloudPlatform(json)) {
```
coryan, line 188: This screams for typecode -> class refactoring again.

coryan (changes requested):

google/cloud/opentelemetry/resource_detector.cc
dbolduc: > You need a retry loop.

Good call. The retry/backoff settings are not obvious to me.

> A failed attempt should be retried after a few minutes.

Hmmm... thinking out loud.... `Detect()` should be called once from within `ConfigureBasicTracing()` when we initialize the `TracerProvider`.

https://github.com/googleapis/google-cloud-cpp/blob/0c7f077b96f802c0e10f500a3e601b2663bcf353/google/cloud/opentelemetry/configure_basic_tracing.cc#L66

It seems wrong to block an application not running on GCP for a few minutes. And I think you are saying we can't distinguish between a transient error and the case where there is no MD server.

We could introduce some `GcpResourceDetectorOption` that determines whether we include the resource detector or not.

Another (bad) workaround would be to build the resource detector implementation into the trace exporter.

coryan: When is `Detect()` called?  Or more important: how often?    If it is called only once, we need to retry right there on the spot.  If it is called each time something gets exported we can try to fetch the values at that time, with caching to avoid expensive operations.

dbolduc (commented):
I heavily refactored the code to include
- validation of headers / payload
- retry loop for querying the metadata server
- attribute caching

Hopefully it is more readable, too.

I considered factoring out the logic into a testable interface, but decided against it because the tests (basically) look the same either way. It might have been a good idea to start to keep the PR size down... but it's too late. Sorry.

I also considered factoring out the client/connection that queries the metadata server. I decided to just accept a retry/backoff policy in the `otel_internal::MakeResourceDetector()` instead.

---

I see the windows lambda capture failure... ugh, I will fix.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#-:-NXHMPAo7sDh1FUbwFEB:b3bxsq2)* status: 0 of 11 files reviewed, 6 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/opentelemetry/resource_detector.h` line 38 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#gh-1215556523:-NXHLpCn5moaQ9_AypQV:b-w3vgv9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b5a101fae53b65641eb041ca5b1c1d7210814993/google/cloud/opentelemetry/resource_detector.h#L38)):*

nit-ish-maybe-it-matters: opening the `*_internal` namespaces in a public header is a smell.  This probably belongs in `google/cloud/opentelemetry/internal/make_resource_detector.h`
</blockquote></details>

Ack. I moved this out.

I am sure there are more such instances of this pattern: https://github.com/googleapis/google-cloud-cpp/blob/3eab36478055874b3acb30a844494e95f99e49dc/google/cloud/pubsub/subscriber_connection.h#L182-L190
___
*[`google/cloud/opentelemetry/resource_detector.cc` line 40 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#gh-1215559913:-NXHLcds-my0dB1r6S_N:bfhqot7) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b5a101fae53b65641eb041ca5b1c1d7210814993/google/cloud/opentelemetry/resource_detector.cc#L40)):*

The function name is not great.  `FindRecursive()`?
</blockquote></details>

I kept a member function named `MD()` that returns metadata, instead of looking up a key in any given `nlohmann::json`.
___
*[`google/cloud/opentelemetry/resource_detector.cc` line 44 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#gh-1215562487:-NXHLWNW66sK2FjcuVq6:b-ge115q) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b5a101fae53b65641eb041ca5b1c1d7210814993/google/cloud/opentelemetry/resource_detector.cc#L44)):*

I think here you want something like:

```cc
auto const f = j->find(k);
if (f == j->end()) return {};
j = &(*f);
```

But even that is a bit weird and obscures what you are doing.  Maybe:

```cc
std::string FindRecursive(nlohmann::json const& json, std::deque<std::string> keys) {
  if (!keys.empty()) {
    auto const l = json.find(keys.front());
    if (l ==json.end()) return {};
    keys.pop_front();
    return FindRecursive(*l, std::move(keys));
  }
  if (json.is_string()) return json.get<std::string>();
  if (json.is_number()) return std::to_string(json.get<std::int64_t>());
  return {};
}
```

</blockquote></details>

Accepted.
___
*[`google/cloud/opentelemetry/resource_detector.cc` line 166 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#gh-1215565028:-NXHLNWs2J3lXvLnEnxX:bdlmofb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b5a101fae53b65641eb041ca5b1c1d7210814993/google/cloud/opentelemetry/resource_detector.cc#L166)):*

When is `Detect()` called?  Or more important: how often?    If it is called only once, we need to retry right there on the spot.  If it is called each time something gets exported we can try to fetch the values at that time, with caching to avoid expensive operations.
</blockquote></details>

`Detect()` should be called once, when initializing the `TracerProvider`.
___
*[`google/cloud/opentelemetry/resource_detector.cc` line 181 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#gh-1215575475:-NXHL8SX1RCeqvaX1vk7:bo8h9v0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b5a101fae53b65641eb041ca5b1c1d7210814993/google/cloud/opentelemetry/resource_detector.cc#L181)):*

You need more validation.  First, we should check that the headers returned matched our expectations.  Second, some people have ISPs that return HTML (full of ads) when you send a HTTP request to an unknown hostname.
</blockquote></details>

Added `ValidateJson()` and `ValidateHeaders()`
___
*[`google/cloud/opentelemetry/resource_detector.cc` line 188 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#gh-1215576706:-NXHLEq-CK0psUzEdEnK:bmz1il6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b5a101fae53b65641eb041ca5b1c1d7210814993/google/cloud/opentelemetry/resource_detector.cc#L188)):*

This screams for typecode -> class refactoring again.
</blockquote></details>

Ack. I threw this into `class Parser`.

dbolduc (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776)* status: 9 of 11 files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/opentelemetry/internal/resource_detector_impl.cc` line 61 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#-NXHTVW471J6GnKQu5iw:-NXLM9Wr2I08FC7vfMMH:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b2b4c86dffb33ff2133e21009030e10816a5fdab/google/cloud/opentelemetry/internal/resource_detector_impl.cc#L61)):*

Content-Type can be `application/json; charset=utf-8`. Maybe "StartsWith` or something.
</blockquote></details>

Done.
___
*[`google/cloud/opentelemetry/internal/resource_detector_impl.cc` line 89 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#-NXHU9G00fVM05mQRA-Y:-NXLMAJA8v_J2Z8ZfRR7:bbwjn8v) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b2b4c86dffb33ff2133e21009030e10816a5fdab/google/cloud/opentelemetry/internal/resource_detector_impl.cc#L89)):*

This is fine as-is, I don't want to make your life more difficult.  I expected something more like this:

```cc
class ResourceAttributeProvider {
 public:
   virtual ~ResourceAttributeProvider() = default;
   virtual opentelemetry::sdk::resource::ResourceAttributes Attributes() const  = 0;
};

std::unique<ResourceAttributeProvider> MakeProvider(... stuff ...) {
   if (internal::GetEnv("KUBERNETES_SERVICE_HOST")) return std::make_unique<GkeProvider>( ... );
   if (!MD({"instance", "machineType"}).empty()) return std::make_unique<GceProvider>( ... );
   return absl::make_unique<NoopProvider>();
}
```

That makes it "easier" to add new providers in the future (e.g., maybe a new version of Cloud Run, or something for AWS or Azure, or whatevs).

</blockquote></details>

Ah, I misinterpreted your earlier comment. Thanks for spelling it out.

I left the code as-is, but I can see how your proposal is more modular.
___
*[`google/cloud/opentelemetry/internal/resource_detector_impl.cc` line 152 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#-NXHTiql1TnODRxApmK3:-NXLOIns75bl6uK0EEDw:b-738y9s) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/b2b4c86dffb33ff2133e21009030e10816a5fdab/google/cloud/opentelemetry/internal/resource_detector_impl.cc#L152)):*

One of the problems with `MD` is that it also matches the naming conventions for C and C++ macros.  Think it over, feel free to keep it if you are comfortable with that name (it is only in a `.cc` file, we could just `#undef` the macro if it ever becomes a problem).
</blockquote></details>

Thanks for the reminder. s/MD/Metadata/

coryan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11776#-:-NXLPVvy5wPig89E9rLV:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dbolduc)

PR #11782, opened 2023-06-05
fix: better support for cross-compilation

### LINKED ISSUE #11779: Support cross-compilation: accept a different protoc binary
For CMake-based builds, when cross-compiling (think "macOS on x64 compiling for ARM", or "Android builds").

The `protobuf::protoc` program found by `find_package(Protobuf ...)` is the program for the *target* environment. This may not be runnable in the *host* environment (where the build is running).  However our build scripts use `$<TARGET_FILE:protobuf::protoc>`:

https://github.com/googleapis/google-cloud-cpp/blob/0c7f077b96f802c0e10f500a3e601b2663bcf353/cmake/CompileProtos.cmake#L105

https://github.com/googleapis/google-cloud-cpp/blob/0c7f077b96f802c0e10f500a3e601b2663bcf353/cmake/CompileProtos.cmake#L116

https://github.com/googleapis/google-cloud-cpp/blob/0c7f077b96f802c0e10f500a3e601b2663bcf353/cmake/CompileProtos.cmake#L203-L204

All of these need to be some  variable that can be configured from the command-line.  It seems that Protobuf >= 23.1 uses `Protobuf_PROTOC_EXECUTABLE`.  Older versions of `find_package(Protobuf)` also use that variable. Seems like a good option.

coryan:
Allow overrides for the `protobuf::protoc` and `gRPC::grpc_cpp_plugin` targets. On cross-compilation builds the targets point to the binaries for the *target*, but we need binaries that can run on the *host* build.

Fixes #11779

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11782)

### DIFF
--- a/cmake/CompileProtos.cmake
+++ b/cmake/CompileProtos.cmake
@@ -203,15 +203,11 @@ function (google_cloud_cpp_generate_grpcpp SRCS)
         if (NOT Protobuf_PROTOC_EXECUTABLE)
             set(Protobuf_PROTOC_EXECUTABLE $<TARGET_FILE:protobuf::protoc>)
         endif ()
-        if (NOT _gRPC_CPP_PLUGIN_EXECUTABLE)
-            set(_gRPC_CPP_PLUGIN_EXECUTABLE
-                $<TARGET_FILE:gRPC::grpc_cpp_plugin>)
-        endif ()
         add_custom_command(
             OUTPUT "${grpc_pb_cc}" "${grpc_pb_h}"
             COMMAND
                 ${Protobuf_PROTOC_EXECUTABLE} ARGS
-                --plugin=protoc-gen-grpc=${_gRPC_CPP_PLUGIN_EXECUTABLE}
+                --plugin=protoc-gen-grpc=${GOOGLE_CLOUD_CPP_GRPC_PLUGIN_EXECUTABLE}
                 "--grpc_out=${CMAKE_CURRENT_BINARY_DIR}"
                 "--cpp_out=${CMAKE_CURRENT_BINARY_DIR}" ${protobuf_include_path}
                 "${file_path}"
--- a/cmake/CompileProtos.cmake
+++ b/cmake/CompileProtos.cmake
@@ -205,7 +205,7 @@ function (google_cloud_cpp_generate_grpcpp SRCS)
         endif ()
         if (NOT _gRPC_CPP_PLUGIN_EXECUTABLE)
             set(_gRPC_CPP_PLUGIN_EXECUTABLE
-                $<TARGET_FILE:grpc::grpc_cpp_plugin>)
+                $<TARGET_FILE:gRPC::grpc_cpp_plugin>)
         endif ()
         add_custom_command(
             OUTPUT "${grpc_pb_cc}" "${grpc_pb_h}"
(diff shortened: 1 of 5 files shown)


cmake/CompileProtos.cmake
```diff
@@ -197,15 +200,22 @@ function (google_cloud_cpp_generate_grpcpp SRCS)
             "${CMAKE_CURRENT_BINARY_DIR}/${D}/${file_stem}.grpc.pb.cc")
         set(grpc_pb_h "${CMAKE_CURRENT_BINARY_DIR}/${D}/${file_stem}.grpc.pb.h")
         list(APPEND ${SRCS} "${grpc_pb_cc}" "${grpc_pb_h}")
+        if (NOT Protobuf_PROTOC_EXECUTABLE)
+            set(Protobuf_PROTOC_EXECUTABLE $<TARGET_FILE:protobuf::protoc>)
+        endif ()
+        if (NOT DEFINED _gRPC_CPP_PLUGIN_EXECUTABLE)
+            set(_gRPC_CPP_PLUGIN_EXECUTABLE
+                $<TARGET_FILE:grpc::grpc_cpp_plugin>)
+        endif ()
```
h-vetinari, line 209: It should be safe even for cross-compilation to locate grpc-cpp wherever `Protobuf_PROTOC_EXECUTABLE` is. This is what we ended up [doing](https://github.com/conda-forge/cpp-opentelemetry-sdk-feedstock/blob/190fa3fc4e661099703e519a7591477406a377c2/recipe/patches/0001-cross_compile_grpc_cpp_plugin_path.patch#L4) for open-telemetry, for example.

That would save you a magical, private variable, but to be clear: that's a perfectly fine solution as well.

coryan: That seems intriguing.  If I am following the suggestion one would install `protoc` and `grpc_cpp_plugin` in `$PREFIX/bin` or something similar?  Unfortunately vcpkg installs `protoc` in `$PREFIX/tools/protobuf/protoc` and the gRPC plugins in `$PREFIX/tools/grpc/*`.

I can give that a try once I get something working across all our builds.

h-vetinari: > If I am following the suggestion one would install `protoc` and `grpc_cpp_plugin` in `$PREFIX/bin` or something similar?

To be honest I don't know how exactly CMake does it, but the idea is to search `grpc_cpp_plugin` _relative_ to `protoc`. For cross-compilation, the latter will be in `$BUILD_PREFIX/bin`, which will similarly be correct for grpc, while for native compilation, protoc will be in `$PREFIX/bin` and again that suits grpc too.

h-vetinari: > Unfortunately vcpkg installs `protoc` in `$PREFIX/tools/protobuf/protoc` and the gRPC plugins in `$PREFIX/tools/grpc/*`.

Yeah, that might mean it's not a directly suitable idea for you here. 😑

coryan: > 
> I don't know how the path traversal in `find_program(<something> HINTS ${PATH} ...)` would look like. Perhaps it's still salvageable, but we'd have to try to find out...

Seems like we are getting to diminishing returns pretty quickly.  The default works for most builds, and a CMake flag should work in all (I hope) other cases.  Seems like each package manager would require a different `HINT` entry and at the point we might as well keep the setting in the package manager recipe.

dbolduc (approved):

cmake/CompileProtos.cmake
h-vetinari: My guess was that CMake will exhaustively search the entire folder hierarchy under `HINT`, and then the hint of "start with protoc" would also have been correct for vcpkg. But I don't care about this very much, it was just a thought to save you some complexity

PR #11783, opened 2023-06-05
Add option to set AuthorityOption

mfschwartz:


google/cloud/storage/benchmarks/throughput_options.cc
```diff
@@ -352,6 +353,11 @@ google::cloud::StatusOr<ThroughputOptions> ParseThroughputOptions(
        [&options](std::string const& val) {
          options.direct_path_options.set<EndpointOption>(val);
        }},
+      {"--gprc-authority-hostname",
```
coryan, line 356: Maybe `--direct-path-authority-hostname`?  You are storing this in the `options.direct_path_options` variable.

Do we also need to do something similar for `options.grpc_options`?

coryan (changes requested):
See below.

coryan (approved):

dbolduc (approved):

PR #11789, opened 2023-06-06
test(spanner): integration tests for fine-grained access control

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11789)


coryan (approved):
This is dangerously close to a test for the service.  But I guess we need to verify setting the `SessionCreatorRoleOption` has the desired effect.  Sold.

Issue #11786, opened 2023-06-06
Bigquery Job Custom Class Cleanup


jsrinnn:
Based on [code review feedback](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11784#gh-1219585795) for PR-11784, this issue is tracks the refactor of `to_json` and `from_json` conversion of `std::chrono::milliseconds` fields

Copied the code review feedback below:

////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
I think you should consider a clean up PR where you refactor these conversions to and from milliseconds to something more like:

```c
  s.start_time = FromMillisecondField(j, "start_time");
```

or even

```c
  FromMillisecondField(s.start_time, j, "start_time");
```

where the helper would be something like:

```c
void FromMillisecondField(std::chrono::milliseconds& field, nlohman::json const& j, char const* name) {
  auto const l = j.find(name);
  if (l == j.end()) return;
  std::int64_t m;
  l->get_to(m);
  field = std::chrono::milliseconds(m);
}
```

jsrinnn:
This has been fixed. I will close this issue.

PR #11790, opened 2023-06-06
impl(bigquery): Refactor json parsing

jsrinnn:
This PR refactors to and from json parsing for populating chrono fields specifically:

`std::chrono:: milliseconds`, 
`std::chrono::timepoint` 
`std::chrono::hours`

 Addresses and fixes Issue #11786

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11790)


google/cloud/bigquery/v2/minimal/internal/json_utils.h
```diff
@@ -0,0 +1,50 @@
...
+#include "google/cloud/version.h"
+#include <nlohmann/json.hpp>
+#include <chrono>
+
+namespace google {
+namespace cloud {
+namespace bigquery_v2_minimal_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+void FromJsonMilliseconds(std::chrono::milliseconds& field,
```
coryan, line 27: nit: I think the units are implicit in the parameters of these functions.  Don't feel like you must change these.  Only if it makes more sense to you.

coryan (approved):
Very cool cleanup, thanks.

google/cloud/bigquery/v2/minimal/internal/json_utils.cc
```diff
@@ -0,0 +1,71 @@
...
+}
+
+void FromJsonTimepoint(std::chrono::system_clock::time_point& field,
+                       nlohmann::json const& j, char const* name) {
+  auto const l = j.find(name);
+  if (l == j.end()) return;
+  std::int64_t m;
+  l->get_to(m);
+  field = std::chrono::time_point<std::chrono::system_clock>(
+      std::chrono::milliseconds(m));
```
devbww, line 58: I assume that the value `m` has some well-defined meaning as a time instant.  Unfortunately, the `time_point(duration)` constructor, and `time_point::time_since_epoch()` used below, are not so well defined (until C++20).  So, I expect you are going to have to do something involving `from_time_t()`/`to_time_t()` to make the  `time_point` conversions correct.

coryan (approved):
:lgtm:

> Reviewed 4 of 11 files at r1, 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11790#-:-NXLR_1mBVJ7jofzzVdQ:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/json_utils.cc` line 58 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11790#gh-1220254904:-NXLPvnV7MUfXL7qIp0w:b-m1pxlf) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/68e9c88fa29b518f4c39f080922aace814bbb73e/google/cloud/bigquery/v2/minimal/internal/json_utils.cc#L58)):*

Sorry but I see ``time_point::time_since_epoch()` and `duration_cast` being used everywhere in the repo for various different projects so I don't specifically follow the objection here

https://github.com/search?q=repo%3Agoogleapis%2Fgoogle-cloud-cpp+time_since_epoch%28%29&type=code

Additionally `to_time_`t is truncating the value such that it is returning 0 for small timepoint values which is causing a lot of test failures.

If you have specific examples on how to use the `to_time_t` and `from_time_t` to convert to and from time_point that would be helpful and I can address that in a separate PR. All the examples that I have seen use the time_point::time_since_epoch()` and `duration_cast`

</blockquote></details>

@devbww is pointing out that it is not *guaranteed* that `std::chrono::time_point<std::chrono::system_clock>` has the same epoch as unix time (1970-01-01T00:00:00Z), not until C++20.  So its constructor from milliseconds (or any "duration") may yield a different time point than you would expect.  He is right (as usual, some day I will catch him in a mistake), though in practice all the platforms where we support use unix time as their epoch.  OTOH, writing "guaranteed correct" code is not hard:

```cc
field = std::chrono::system_clock::from_time_t(0) + std::chrono::milliseconds(m);
```

Should do the trick.


Incidentally, you can say `std::chrono::system_clock::timepoint` instead of the (slightly longer)  `std::chrono::time_point<std::chrono::system_clock>`

jsrinnn (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11790)* status: all files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/json_utils.cc` line 58 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11790#gh-1220254904:-NXLU_vdB-YdgohRsCHN:b-flh33t) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/68e9c88fa29b518f4c39f080922aace814bbb73e/google/cloud/bigquery/v2/minimal/internal/json_utils.cc#L58)):*

@devbww is pointing out that it is not *guaranteed* that `std::chrono::time_point<std::chrono::system_clock>` has the same epoch as unix time (1970-01-01T00:00:00Z), not until C++20.  So its constructor from milliseconds (or any "duration") may yield a different time point than you would expect.  He is right (as usual, some day I will catch him in a mistake), though in practice all the platforms where we support use unix time as their epoch.  OTOH, writing "guaranteed correct" code is not hard:

```cc
field = std::chrono::system_clock::from_time_t(0) + std::chrono::milliseconds(m);
```

Should do the trick.


Incidentally, you can say `std::chrono::system_clock::timepoint` instead of the (slightly longer)  `std::chrono::time_point<std::chrono::system_clock>`
</blockquote></details>

Thanks.  I have made the above change in `FromJson()`. 

 Just to clarify, for `ToJson()` I am using `duration_cast` with `Timepoint::time_since_epoch` shown below: Does that need to change ? Thats the impression I got from the original comment i.e that it needed to change

```
void ToJson(std::chrono::system_clock::time_point const& field,
            nlohmann::json& j, char const* name) {
  j[name] = std::chrono::duration_cast<std::chrono::milliseconds>(
                field.time_since_epoch())
                .count();
}
```

Basically, Does `to_time_t` needs to be used for the above instead of `time_since_epoch`?  When I use `to_time_t(field)` where field is `Timepoint`, it returns a 0 when field was given a value of `std::chrono::system_clock::time_point(10)`. Maybe more manipulation is needed here with `system_clock::now()`?. Thats where I was asking for examples which I couldn't find in the current repo. External examples all use `time_since_epoch()` same as I have done above.  I am not very familiar with these timepoint structures so feedback is appreciated.

Thanks.

coryan:
*[`google/cloud/bigquery/v2/minimal/internal/json_utils.cc` line 58 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11790#gh-1220254904:-NXLhC6B2rSOlPUCckgC:bpnw4aw) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/68e9c88fa29b518f4c39f080922aace814bbb73e/google/cloud/bigquery/v2/minimal/internal/json_utils.cc#L58)):*

Thanks.  I have made the above change in `FromJson()`. 

 Just to clarify, for `ToJson()` I am using `duration_cast` with `Timepoint::time_since_epoch` shown below: Does that need to change ? Thats the impression I got from the original comment i.e that it needed to change

```
void ToJson(std::chrono::system_clock::time_point const& field,
            nlohmann::json& j, char const* name) {
  j[name] = std::chrono::duration_cast<std::chrono::milliseconds>(
                field.time_since_epoch())
                .count();
}
```

Basically, Does `to_time_t` needs to be used for the above instead of `time_since_epoch`?  When I use `to_time_t(field)` where field is `Timepoint`, it returns a 0 when field was given a value of `std::chrono::system_clock::time_point(10)`. Maybe more manipulation is needed here with `system_clock::now()`?. Thats where I was asking for examples which I couldn't find in the current repo. External examples all use `time_since_epoch()` same as I have done above.  I am not very familiar with these timepoint structures so feedback is appreciated.

Thanks.
</blockquote></details>

I think you would want something like this in the opposite direction:

```cc
j[name] = std::chrono::duration_cast<std::chrono::milliseconds>(field - std::chrono::system_clock::from_time_t(0)).count();
```

jsrinnn (commented):
Addressed latest feedback. Thanks Carlos!

Bradley, 

I believe the latest changes addresses your original feedback. I'll go ahead and enable auto-merge on this PR for forward progress. 

If you think additional changes are needed please feel free to provide feedback here and I can address them separately. 

Thanks

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11790#-:-NXLkmCxEyr7BY4EK2Wn:bol5ttc)* status: all files reviewed, 1 unresolved discussion (waiting on @devbww)

___
*[`google/cloud/bigquery/v2/minimal/internal/json_utils.cc` line 58 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11790#gh-1220254904:-NXLjqN13l1WwQIQA3iF:bvq5p4j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/68e9c88fa29b518f4c39f080922aace814bbb73e/google/cloud/bigquery/v2/minimal/internal/json_utils.cc#L58)):*

I think you would want something like this in the opposite direction:

```cc
j[name] = std::chrono::duration_cast<std::chrono::milliseconds>(field - std::chrono::system_clock::from_time_t(0)).count();
```



</blockquote></details>

Done.

Thanks Carlos. I have made the above change.

Just wanted to point out that if the objection is to the usage of `time_since_epoch` for guaranteed correctness, then what I said in my original reply still holds i.e there are other places in the repo which are using `time_since_epoch` with `duration_cast` like I was doing before. I'll leave it upto the cpp team to change those as necessary.

devbww:
> Bradley, the change requested by you is out of scope for this CL

Correctness is always in scope.

> there are other places in the repo which are using time_since_epoch with duration_cast like I was doing before.

Yes, we certainly have latent bugs.  Fortunately, in this case, we created `google/cloud/internal/parse_rfc3339_test.cc:ParseRfc3339Test.ParseEpoch` to help determine when they might bite.  But that doesn't mean we shouldn't do the right thing in new code.

jsrinnn:
> Correctness is always in scope.
Yes, in that it needs to be addressed but necessarily in the same PR.

Thanks

Issue #10619, opened 2023-01-19
OpenTelemetry tracing for streams


dbolduc:
Some tasks from [Milestone 3](https://docs.google.com/document/d/1JTOWvZD59AGxojTV93tKZJ9b_q2w8YUMMyuxCu5_ZC4/edit?resourcekey=0-0XjX6fh2xNKFZN20cu2otg#bookmark=id.a4oyji96qrlm) of an internal design doc. With these tasks completed, our internals can support tracing all streaming APIs.

Implement tracing wrapper for:
- [x] `StreamingReadRpc`
- [x] `AsyncStreamingReadRpc`
- [x] `AsyncStreamingReadWriteRpc`
- [x] `StreamingWriteRpc`
- [ ] `AsyncStreamingWriteRpc`

PR #11812, opened 2023-06-08
impl(otel): `AsyncStreamingReadWriteRpcTracing`

dbolduc:
Part of the work for #10619

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11812)

### DIFF
--- a/google/cloud/internal/async_read_write_stream_tracing.h
+++ b/google/cloud/internal/async_read_write_stream_tracing.h
@@ -1,0 +1,110 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


google/cloud/internal/async_read_write_stream_tracing.h
```diff
@@ -0,0 +1,110 @@
...
+                                      {"message.id", ++write_count_},
+                                      {"message.is_last", is_last},
+                                      {"message.success", success}});
+          return success;
+        });
+  }
+
+  future<bool> WritesDone() override {
+    return impl_->WritesDone().then([this](future<bool> f) {
+      span_->AddEvent("gcloud.writes_done");
```
coryan, line 81: FYI: when you implement some kind of reservoir sampler for events you will want to preserve `start` and `writes_done` and maybe `cancel`.

coryan (approved):

google/cloud/internal/async_read_write_stream_tracing.h
dbolduc: Ack, although I don't know how or when I would implement that.

(But I take your general point that all events are not created equally, and we should play favorites when events are limited)

PR #11828, opened 2023-06-08
impl(generator): metadata decorator fixed headers

### LINKED ISSUE #11823: Support custom metadata in stubs (known at stub creation time)
# Background

Bigtable needs to send custom metadata to tell the server which FeatureFlags it supports.

Spanner needs to send a header set to the `spanner::Database` name. https://github.com/googleapis/google-cloud-cpp/blob/a21084878d15216fbdd6a8e66160d799fc22e9ea/google/cloud/spanner/internal/metadata_spanner_stub.cc#L156

With this feature, we could generate the spanner stub layers. (The stub factory would stay handwritten, as is typical for the more complicated services).

# Implementation

We can add a `std::unordered_map<std::string, std::string> fixed_headers` to the MD decorator constructors. And loop over it in `*Metadata::SetMetadata()`.

dbolduc:
Fixes #11823 

- I used a defaulted parameter (against @devbww's recommendation) to minimize the changes in this PR.
- I did not touch the `RestMetadataGenerator`, given that only services with handwritten stub factories can use the fixed metadata feature.
  - (currently, only `bigtable` and `spanner` care)

### DIFF
--- a/generator/integration_tests/tests/golden_kitchen_sink_metadata_decorator_test.cc
+++ b/generator/integration_tests/tests/golden_kitchen_sink_metadata_decorator_test.cc
@@ -39,6 +39,7 @@ using ::google::cloud::testing_util::ValidateMetadataFixture;
 using ::google::test::admin::database::v1::Request;
 using ::google::test::admin::database::v1::Response;
 using ::testing::_;
+using ::testing::AllOf;
 using ::testing::AnyOf;
 using ::testing::Contains;
 using ::testing::Not;
@@ -464,6 +465,28 @@ TEST_F(MetadataDecoratorTest, ExplicitRoutingNestedField) {
   // extract its metadata).
   (void)stub.ExplicitRouting2(context1, request);
   (void)stub.ExplicitRouting2(context2, request);
+}
+
+TEST_F(MetadataDecoratorTest, FixedMetadata) {
+  // We do this for a single RPC, we are using some knowledge of the
+  // implementation to assert that this is enough.
+  EXPECT_CALL(*mock_, GenerateAccessToken)
+      .WillOnce([this](grpc::ClientContext& context,
+                       google::test::admin::database::v1::
+                           GenerateAccessTokenRequest const&) {
+        auto metadata = GetMetadata(context);
+        EXPECT_THAT(metadata,
+                    AllOf(Contains(Pair("test-key-1", "test-value-1")),
+                          Contains(Pair("test-key-2", "test-value-2"))));
+        return TransientError();
+      });
+
+  GoldenKitchenSinkMetadata stub(
+      mock_, {{"test-key-1", "test-value-1"}, {"test-key-2", "test-value-2"}});
+  grpc::ClientContext context;
+  google::test::admin::database::v1::GenerateAccessTokenRequest request;
+  auto status = stub.GenerateAccessToken(context, request);
+  EXPECT_EQ(TransientError(), status.status());
 }
 
 }  // namespace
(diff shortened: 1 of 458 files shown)


generator/integration_tests/tests/golden_kitchen_sink_metadata_decorator_test.cc
```diff
@@ -466,6 +467,28 @@ TEST_F(MetadataDecoratorTest, ExplicitRoutingNestedField) {
   (void)stub.ExplicitRouting2(context2, request);
 }
 
+TEST_F(MetadataDecoratorTest, FixedMetadata) {
+  // We do this for a single RPC, we are using some knowledge of the
+  // implementation to assert that this is enough.
```
devbww, line 472: "We use knowledge of the implementation to assert that testing a single RPC is sufficient."

generator/internal/metadata_decorator_generator.cc
```diff
@@ -126,19 +126,20 @@ Status MetadataDecoratorGenerator::GenerateHeader() {
...
-    "  ~$metadata_class_name$() override = default;\n"
-    "  explicit $metadata_class_name$(std::shared_ptr<$stub_class_name$> child);\n");
-  // clang-format on
+  HeaderPrint(R"""(
+class $metadata_class_name$ : public $stub_class_name$ {
+ public:
+  ~$metadata_class_name$() override = default;
+  explicit $metadata_class_name$(
+      std::shared_ptr<$stub_class_name$> child,
+      std::unordered_map<std::string, std::string> fixed_metadata = {});
```
devbww, line 141: The `grpc::ClientContext` map is actually a `std::multimap<std::string, std::string>`.  We should probably match that.

generator/internal/metadata_decorator_generator.cc
```diff
@@ -365,6 +368,9 @@ void $metadata_class_name$::SetMetadata(grpc::ClientContext& context,
 
 void $metadata_class_name$::SetMetadata(grpc::ClientContext& context) {
   context.AddMetadata("x-goog-api-client", api_client_header_);
+  for (auto const& kv : fixed_metadata_) {
+    context.AddMetadata(kv.first, kv.second);
+  }
```
devbww, line 373: Do we have any idea about which value (first or last) would prevail in the service should the same key be repeated?

Given such knowledge, and the knowledge that iteration order of a `multimap` is insertion order for equal keys, we should be deliberate in the order of  `AddMetadata()` calls between `fixed_metadata_` and the two literal keys.

devbww:
> I used a defaulted parameter (against @devbww's recommendation) to minimize the changes in this PR.

Does that imply we'll un-default it in a followup?

generator/integration_tests/tests/golden_kitchen_sink_metadata_decorator_test.cc
dbolduc: Fixed here and elsewhere.

generator/internal/metadata_decorator_generator.cc
dbolduc: Good catch. Fixed.

generator/internal/metadata_decorator_generator.cc
dbolduc: > we should be deliberate

I think we can be blasé about this. As far as we know, the only fixed keys are "bigtable_features" and "google-cloud-resource-prefix". I don't think we will accidentally cause conflicts with the "x-goog-*" keys.

dbolduc:
> Does that imply we'll un-default it in a followup?

I will not. I cannot speak to whether we will un-default it.

generator/internal/metadata_decorator_generator.cc
devbww: OK.  I think the new ordering is at least more locally consistent.  Thanks.

devbww:
> > Does that imply we'll un-default it in a followup?
> 
> I will not. I cannot speak to whether we will un-default it.

OK ... I'll look into it after this lands.

devbww (approved):

Issue #11720, opened 2023-05-24
Implement pubsub samples


alevenberg:
There are 10 pubsub samples that are missing for C++

 https://cloud.google.com/pubsub/docs/samples

This issue will track the implementation

alevenberg:
Two more PRS
4.

pubsub_subscribe_avro_records_with_revisions
https://cloud.google.com/pubsub/docs/samples/pubsub-subscribe-avro-records-with-revisions

5.
pubsub_create_cloud_storage_subscription

PR #11840, opened 2023-06-09
fix(pubsub): add commit schema and list schema revision samples

alevenberg:
Issue #11720

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11840)

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -2002,9 +2002,6 @@ void AutoRunAvro(
   std::cout << "\nRunning ListSchemaRevisions() sample" << std::endl;
   ListSchemaRevisions(schema_admin, {project_id, avro_schema_id});
 
-  std::cout << "\nRunning ValidateAvroMessage() sample" << std::endl;
-  ValidateMessageAvro(schema_admin, {project_id});
-
   std::cout << "\nRunning CreateTopicWithSchema() sample [avro]" << std::endl;
   auto const avro_topic_id = RandomTopicId(generator);
   CreateTopicWithSchema(topic_admin_client,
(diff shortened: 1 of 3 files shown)


coryan (approved):

google/cloud/pubsub/samples/samples.cc
```diff
@@ -834,6 +835,65 @@ void CreateProtobufSchema(google::cloud::pubsub::SchemaServiceClient client,
   (std::move(client), argv.at(0), argv.at(1), argv.at(2));
 }
 
+void CommitAvroSchema(google::cloud::pubsub::SchemaServiceClient client,
+                      std::vector<std::string> const& argv) {
+  //! [START pubsub_commit_avro_schema] [commit-avro-schema]
```
dbolduc, line 840: FYI: The second tag is unnecessary and can be removed.  (unless you plan to write a doc that links it). Here's why...

---

There are two types of tags for samples.

The first tag `[START pubsub_commit_avro_schema]` is used by the DevRel teams. These tags denote samples that show up on https://cloud.google.com/pubsub/docs/samples (for example).

The second tag, `[commit-avro-schema]`, is something our team uses to include samples in our documentation.

For example, the `google::cloud::pubsub::Publisher`:

https://github.com/googleapis/google-cloud-cpp/blob/facd5b44ec871cb4a40985615315c024e88d6835/google/cloud/pubsub/publisher.h#L75

.... links to this sample:

https://github.com/googleapis/google-cloud-cpp/blob/facd5b44ec871cb4a40985615315c024e88d6835/google/cloud/pubsub/samples/samples.cc#L1400-L1439

... which results in the following documentation:

https://cloud.google.com/cpp/docs/reference/pubsub/latest/classgoogle_1_1cloud_1_1pubsub_1_1Publisher#example-using-a-custom-thread-pool

dbolduc (commented):
nit: This type of PR should be a `doc(pubsub)` instead of a `fix(pubsub)`.

(Oops, I am too late on the review. It's not a big deal. Just use `doc` for the next one).

google/cloud/pubsub/samples/samples.cc
alevenberg: Ah i see will send a PR to fix

google/cloud/pubsub/samples/samples.cc
```diff
@@ -1917,6 +1999,9 @@ void AutoRunAvro(
   std::cout << "\nRunning ListSchemas() sample" << std::endl;
   ListSchemas(schema_admin, {project_id});
 
+  std::cout << "\nRunning ListSchemaRevisions() sample" << std::endl;
```
0000854453, line 2002: @rebase

0000854453 (commented):
[]()

0000854453:
> []()

PR #11844, opened 2023-06-09
doc(pubsub): remove unnecessary documentation tags

alevenberg:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11844)


dbolduc:
we could remove our style of region tags from any samples in `samples.cc` using a
- `TopicAdminClient`
- `SchemaServiceClient`
- `SubscriptionAdminClient`

(We used to handwrite these classes. In the past, we linked the samples from the client's documentation. But now that the clients are fully generated, they do not contain the links).

---

Alternatively, it may be wiser to leave all these tags, and add a document (like https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/pubsub/doc/publisher-mock.dox) that includes the samples in one place.

dbolduc (approved):
Your call

alevenberg:
Gonna use this PR to just clean up #11840

Issue #11854, opened 2023-06-10
Add Spanner tracing stub


dbolduc:
- [x] factor out spanner stub factory
- [x] use generator for spanner stubs
- [x] include tracing stub in stub factory

While I am surely stepping on toes, I could not help myself. This seemed sorta fun. I have done the first three tasks here:

https://github.com/googleapis/google-cloud-cpp/compare/main...dbolduc:spanner-use-generator-full?expand=1

devbww:
For what it's worth, when you generate the Spanner stubs, you can ...
```
  omitted_rpcs: [
    "GetSession",
    "ListSessions",
    "Read"
  ]
```

dbolduc:
This is done. There is an optional refactor tracked in #11888

PR #11855, opened 2023-06-10
refactor(spanner): dedicated stub factory

dbolduc:
Part of the work for #11854

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11855)


devbww (approved):
LGTM, but just checking on who is following through on the generated-Spanner-stubs effort.

devbww:
> LGTM, but just checking on who is following through on the generated-Spanner-stubs effort.

Well, I see #11854 now, so I guess its you.  That's fine.

dbolduc:
> Well, I see https://github.com/googleapis/google-cloud-cpp/issues/11854 now, so I guess its you. That's fine.

My angle is to do the bare minimum to include the generated tracing stub.

I *think* the resulting code will call for an additional refactor of `PartialResultSetReader` (to drop the `Finish()`). and skip things like this: https://github.com/googleapis/google-cloud-cpp/blob/a91d7d4f160194c64c4bb02acd54b9f3a28a7cbc/google/cloud/spanner/internal/partial_result_set_source.cc#L97-L107

I briefly considered doing that refactor, but decided it was too complicated for me.

---

Aside: I will wait for #11853 before doing any (re)generation

PR #11857, opened 2023-06-10
refactor(spanner): use generated stubs

dbolduc:
Part of the work for #11854 

The most interesting changes are in `connection_impl.cc`. The re-implementation of `DefaultPartialResultSetReader` could use scrutiny. The refactor could also go further.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11857)


google/cloud/spanner/internal/spanner_stub_factory.cc
```diff
@@ -47,11 +47,14 @@ std::shared_ptr<SpannerStub> CreateDefaultSpannerStub(
   if (auth->RequiresConfigureContext()) {
     stub = std::make_shared<SpannerAuth>(std::move(auth), std::move(stub));
   }
-  stub = std::make_shared<MetadataSpannerStub>(std::move(stub), db.FullName());
+  stub = std::make_shared<SpannerMetadata>(
```
coryan, line 50: Shouldn't this use the new `SpannerTracingStub` too?

coryan (commented):
I will let @devbww have the last word on this one.  Maybe you intended to turn on OTel integration in a future PR, but if not, I think you missed it?

google/cloud/spanner/internal/spanner_stub_factory.cc
dbolduc: My intention was to have this PR be a pure refactor. The next PR will include the tracing stub:

https://github.com/googleapis/google-cloud-cpp/commit/742b02d40b119a9ea7db0542d87cd07a4a7ae412

generator/generator_config.textproto
```diff
@@ -2563,6 +2563,26 @@ service {
 }
 
 # Spanner
+service {
+  service_proto_path: "google/spanner/v1/spanner.proto"
+  product_path: "google/cloud/spanner"
```
devbww, line 2568: Is there any reason to consider putting the generated files in a version-specific directory?

devbww (approved):

generator/generator_config.textproto
dbolduc: I briefly considered it. I decided it doesn't matter because only internal files are generated in this configuration.

PR #11858, opened 2023-06-10
feat(generator): handle comments with 3 newlines

coryan:
We need to preserve newlines in comments as `@n` commands. We do not need (or want?) to preserve each newline because in markdown one or more act as paragraph separators.

There are no comments with 4 newlines in a row (yet).

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11858)

### DIFF
--- a/generator/internal/descriptor_utils.cc
+++ b/generator/internal/descriptor_utils.cc
@@ -396,6 +396,17 @@ ParameterCommentSubstitution substitutions[] = {
     {R"""(`{cluster} = '-'`)""", R"""(``{cluster} = '-'``)"""},
     {R"""(`projects/<Project ID or '-'>`)""",
      R"""(``projects/<Project ID or '-'>``)"""},
+
+    // Some comments include multiple newlines in a row. We need to preserve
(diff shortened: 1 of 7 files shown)


generator/internal/descriptor_utils.cc
```diff
@@ -396,6 +396,17 @@ ParameterCommentSubstitution substitutions[] = {
     {R"""(`{cluster} = '-'`)""", R"""(``{cluster} = '-'``)"""},
     {R"""(`projects/<Project ID or '-'>`)""",
      R"""(``projects/<Project ID or '-'>``)"""},
+
+    // Some comments include multiple newlines in a row. We need to preserve
+    // these because they are paragraph separators. When used in `@param`
+    // commands we need to represent them as `@n` or they do more than separate
+    // the paragraph. The would terminate the `@param` comment.
```
dbolduc, line 403: "we need to represent them as `@n` or they would terminate the `@param` comment."

dbolduc (approved):

PR #11872, opened 2023-06-12
doc(pubsub): add update topic schema and create topic with schema revisions to samples

alevenberg:
Fixes part of Issue #11720

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11872)

### DIFF
--- a/google/cloud/pubsub/samples/pubsub_samples_common.cc
+++ b/google/cloud/pubsub/samples/pubsub_samples_common.cc
@@ -186,7 +186,7 @@ std::pair<std::string, std::string> CommitSchemaWithRevisionsForTesting(
   create_request.mutable_schema()->set_definition(initial_definition);
   auto schema = client.CreateSchema(create_request);
   if (!schema) throw std::move(schema).status();
-  std::string const first_revision_id = schema.value().revision_id();
+  auto first_revision_id = schema->revision_id();
 
   google::pubsub::v1::CommitSchemaRequest commit_request;
   std::string const name =
@@ -197,9 +197,9 @@ std::pair<std::string, std::string> CommitSchemaWithRevisionsForTesting(
   commit_request.mutable_schema()->set_definition(revised_definition);
   schema = client.CommitSchema(commit_request);
   if (!schema) throw std::move(schema).status();
-  std::string const last_revision_id = schema.value().revision_id();
+  auto last_revision_id = schema->revision_id();
 
-  return {first_revision_id, last_revision_id};
+  return {std::move(first_revision_id), std::move(last_revision_id)};
 }
 
 }  // namespace examples
--- a/google/cloud/pubsub/samples/pubsub_samples_common.cc
+++ b/google/cloud/pubsub/samples/pubsub_samples_common.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/pubsub/samples/pubsub_samples_common.h"
 #include "google/cloud/pubsub/testing/random_names.h"
 #include "google/cloud/internal/getenv.h"
+#include "google/cloud/project.h"
 #include <fstream>
 #include <sstream>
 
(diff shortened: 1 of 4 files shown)


google/cloud/pubsub/samples/pubsub_samples_common.cc
```diff
@@ -167,25 +168,38 @@ std::string ReadFile(std::string const& path) {
...
+                               : google::pubsub::v1::Schema::PROTOCOL_BUFFER;
+
+  google::pubsub::v1::CreateSchemaRequest create_request;
+  create_request.set_parent(google::cloud::Project(project_id).FullName());
+  create_request.set_schema_id(schema_id);
+  create_request.mutable_schema()->set_type(schema_type);
+  create_request.mutable_schema()->set_definition(initial_definition);
+  auto schema = client.CreateSchema(create_request);
+  if (!schema) throw std::move(schema).status();
+  std::string const first_revision_id = schema.value().revision_id();
```
coryan, line 189: We spell these `schema->revision_id()`.  Also, making this a `const` prevents move-from.  Does not matter in this test, but generally it would be more efficient to write:

```cc
... ..
  auto first_revision_id = schema->value();
...
  return {std::move(first_revision_id), std::move(last_revision_id)}
```

If you want to save even more copying you would write:

```cc
  auto first_revision_id = std::move(*schema->mutable_revision_id());
 ..
...
  return {std::move(first_revision_id), std::move(*schema->mutable_revision_id())};
```

google/cloud/pubsub/samples/pubsub_samples_common.cc
```diff
@@ -167,25 +168,38 @@ std::string ReadFile(std::string const& path) {
...
-  std::string first_revision_id = schema.value().revision_id();
-  schema = schema_admin.CommitSchema(request);
+  commit_request.set_name(name);
+  commit_request.mutable_schema()->set_name(name);
+  commit_request.mutable_schema()->set_type(schema_type);
+  commit_request.mutable_schema()->set_definition(revised_definition);
+  schema = client.CommitSchema(commit_request);
   if (!schema) throw std::move(schema).status();
-  return first_revision_id;
+  std::string const last_revision_id = schema.value().revision_id();
```
coryan, line 200: See above

google/cloud/pubsub/samples/pubsub_samples_common.h
```diff
@@ -78,11 +78,12 @@ std::string RandomSchemaId(google::cloud::internal::DefaultPRNG& generator);
 
 std::string ReadFile(std::string const& path);
 
-// Commit two new schema revision and return the first revision id.
-std::string CommitSchemaRevisionsForRollbackSchemaTesting(
-    google::cloud::pubsub::SchemaServiceClient& schema_admin,
+// Commit a schema with a revision and return the first and last revision id.
```
coryan, line 81: s/revision id/revision ids/ ?

google/cloud/pubsub/samples/samples.cc
```diff
@@ -1174,6 +1173,74 @@ void CreateTopicWithSchema(google::cloud::pubsub::TopicAdminClient client,
...
+                                             : google::pubsub::v1::BINARY)
+            .set_first_revision_id(first_revision_id)
+            .set_last_revision_id(last_revision_id));
+
+    // Note that kAlreadyExists is a possible error when the
+    // library retries.
+    if (topic.status().code() == google::cloud::StatusCode::kAlreadyExists) {
+      std::cout << "The topic already exists\n";
+      return;
+    } else if (topic.status().code() ==
```
coryan, line 1200: We typically write

```suggestion
      return;
    }
    if (topic.status().code() == google::cloud::StatusCode::kInvalidArgument) {
```

google/cloud/pubsub/samples/samples.cc
```diff
@@ -1174,6 +1173,74 @@ void CreateTopicWithSchema(google::cloud::pubsub::TopicAdminClient client,
...
+  [](pubsub::TopicAdminClient client, std::string project_id,
+     std::string topic_id, std::string const& first_revision_id,
+     std::string const& last_revision_id) {
+    auto topic = client.UpdateTopic(
+        pubsub::TopicBuilder(
+            pubsub::Topic(std::move(project_id), std::move(topic_id)))
+            .set_first_revision_id(first_revision_id)
+            .set_last_revision_id(last_revision_id));
+
+    if (topic.status().code() == google::cloud::StatusCode::kInvalidArgument) {
```
coryan, line 1229: There are a multitude of reasons why we may get this error, just skip this `if()` I think.

google/cloud/pubsub/samples/samples.cc
```diff
@@ -1174,6 +1173,74 @@ void CreateTopicWithSchema(google::cloud::pubsub::TopicAdminClient client,
...
+            .set_first_revision_id(first_revision_id)
+            .set_last_revision_id(last_revision_id));
+
+    // Note that kAlreadyExists is a possible error when the
+    // library retries.
+    if (topic.status().code() == google::cloud::StatusCode::kAlreadyExists) {
+      std::cout << "The topic already exists\n";
+      return;
+    } else if (topic.status().code() ==
+               google::cloud::StatusCode::kInvalidArgument) {
```
coryan, line 1201: I am not sure the message for `kInvalidArgument` is accurate, or more informative than the message in `topic.status()`?

coryan (changes requested):

google/cloud/pubsub/samples/pubsub_samples_common.cc
alevenberg: Ack. For StatusOr's, use indirection instead of .value().

Changed so that we are using std::move in the return. I think the second version is overkill for the samples.

google/cloud/pubsub/samples/pubsub_samples_common.h
alevenberg: Done

google/cloud/pubsub/samples/samples.cc
alevenberg: Ack. Removed since the code is thrown says the same thing.

google/cloud/pubsub/samples/samples.cc
alevenberg: Removed the if

google/cloud/pubsub/samples/samples.cc
alevenberg: The messages are the same if the revision ids are incorrect, but it could be something else that triggers the error.

Instead of trying to catch the error, changed to rely on throwing the error at the end

PR #11894, opened 2023-06-14
doc(pubsub): add CloudStorageConfigBuilder for samples

alevenberg:
Part of Issue #11720

Adding a sample for pubsub_create_cloud_storage_subscription

It currently is returning an error, so this PR is only builder code for the cloud storage configuration

### DIFF
--- a/google/cloud/pubsub/subscription_builder.cc
+++ b/google/cloud/pubsub/subscription_builder.cc
@@ -77,6 +77,20 @@ SubscriptionBuilder& SubscriptionBuilder::set_bigquery_config(
   return *this;
 }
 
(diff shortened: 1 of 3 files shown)


coryan (approved):
LGTM.  

Something to ponder: how do we avoid this type of work in the future?  Can we change the code so we do not need to expand the builder for new proto fields?  Can we retire the builder (e.g. if we know nobody is using it)?

coryan:
> It currently is returning an error, so this PR is only builder code for the cloud storage configuration

Sometimes we put a `if (UsingEmulator())` around the samples, with a TODO to clean up once the feature is GA.

PR #11897, opened 2023-06-14
feat(otel): control tracing with environment variable

dbolduc:
Part of the work for #12316 

Add a `GOOGLE_CLOUD_CPP_OPENTELEMETRY_TRACING` environment variable. I do not think we need `EXPERIMENTAL` in the name when it touches an `experimental::` option.

~Note that this option violates the "Environment variables take precedence" ADR. We have a reason to. I think it is fine as long as we communicate the expected behavior.~

Touch up the documentation for the tracing option.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11897)


coryan:
Should we keep this PR open?

google/cloud/internal/populate_common_options.cc
```diff
@@ -58,8 +59,16 @@ Options PopulateCommonOptions(Options opts, std::string const& endpoint_env_var,
     opts.set<UserProjectOption>(*std::move(e));
   }
 
+  if (!opts.has<experimental::OpenTelemetryTracingOption>()) {
+    // The Option takes precedence over the environment variable, so we can
+    // explicitly turn off tracing in the Cloud Trace client that implements our
```
coryan, line 64: Is that a configuration that customers will want to use?  Meaning: "I have turned on tracing via an environment variable, but I also explicitly turned off exporting via code"?  Or is this just something we need for our testing?

Do we need two separate options, one to turn off exporting and one to turn off tracing?

dbolduc: > Is that a configuration that customers will want to use? Meaning: "I have turned on tracing via an environment variable, but I also explicitly turned off exporting via code"? Or is this just something we need for our testing?

It is really for us. I would not say it is for our testing. It is not really intended as a customer feature, although the crafty ones may use it.

It is working around the fact that our Cloud Trace client is no different from a Cloud Foo client.

Alternatively, we could add something like an `internal::OverrideOpenTelemetryTracingOption` that is used only by the exporter to accomplish the same thing. Something about an `internal` only options seems fishy though. Maybe it is justified in this case.

I do not think there is a good alternative that adds knobs to the generator knobs.

> Do we need two separate options, one to turn off exporting and one to turn off tracing?

No. They are separate. Exporting is configured through `google_cloud_cpp_opentelemetry` / `opentelemetry-cpp`. It is not done via `google_cloud_cpp_common`.

coryan: > It is working around the fact that our Cloud Trace client is no different from a Cloud Foo client.

Ah, so we don't want to trace the trace exports?  Maybe we can convince ourselves that it is Okay to do so:

- Inside Google we trace the exports, and things do not blow up.
- It helps to get a small sample of "the traces are exporting" traces.
- Typically traces are sampled, so the trace exports are rare.
- Typically traces are batched, so there is no explosion due to "tracing the exports creates even more traces!".

dbolduc: FTR, I did experiment with tracing the exporter, and it only creates 1 extra span.

I wasn't sure if other certain settings might lead to the recursive explosion, though.

coryan: I think we should keep this simple and trace the exporter.  If it becomes a problem we can create some ad-hoc thing. Thoughts?

dbolduc: That works for me. I'll simplify this PR.

coryan (approved):

google/cloud/internal/populate_common_options.cc
```diff
@@ -58,8 +59,11 @@ Options PopulateCommonOptions(Options opts, std::string const& endpoint_env_var,
     opts.set<UserProjectOption>(*std::move(e));
   }
 
+  if (GetEnv("GOOGLE_CLOUD_CPP_OPENTELEMETRY_TRACING")) {
```
devbww, line 62: Consider also checking that the result is not empty.  That tends to be the general tactic.

devbww: > Do you think we should be using `ParseBoolean(...)` instead?

Possibly.  (I do have half a PR to add some parsing to getenv.h.)

But mostly I'm thinking about distinguishing between being set with an empty -v- non-empty value.  For example, just a couple of lines up we do that.  (Maybe I'm just too stuck in the past where it was difficult to actually remove something from the environment, so an empty value was usually taken to mean the same thing as unset.)

Anyway, I'll leave it to you.

devbww (approved):

google/cloud/internal/populate_common_options.cc
dbolduc: I'm going to treat the presence of the env var as a boolean and ignore its value.

I added a test case for the empty env var to fully enshrine my decision.

dbolduc: Nevermind.

PR #11898, opened 2023-06-15
impl(common): enhance DebugFormatter support

devbww:
- std::map<std::string, T>

Also make the tests clearer by renaming the `SubMessage` field.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11898)

### DIFF
--- a/google/cloud/internal/debug_string.h
+++ b/google/cloud/internal/debug_string.h
@@ -99,6 +99,20 @@ class DebugFormatter {
     return *this;
   }
 
+  template <typename T>
+  DebugFormatter& Field(absl::string_view field_name,
+                        std::map<std::string, T> const& value) {
+    for (auto const& e : value) {
+      absl::StrAppend(&str_, Sep(), field_name, " {");
+      ++indent_;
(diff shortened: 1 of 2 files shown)


google/cloud/internal/debug_string.h
```diff
@@ -99,6 +99,20 @@ class DebugFormatter {
...
   }
 
+  template <typename T>
+  DebugFormatter& Field(absl::string_view field_name,
+                        std::map<std::string, T> const& value) {
+    for (auto const& e : value) {
+      absl::StrAppend(&str_, Sep(), field_name, " {");
+      ++indent_;
+      absl::StrAppend(&str_, Sep(), "key: ", "\"", e.first, "\"");
+      absl::StrAppend(&str_, e.second.DebugString("value", options_, indent_));
```
coryan, line 109: Food for thought: this only works if `T` has a `DebugString()` member function.  You already have a special case for `std::map<std::string, std::string>`, but what about `std::map<std::string, int>`?

Maybe there is a way to implement this with some ADL, such that we can add types over time.

coryan (approved):

PR #11899, opened 2023-06-15
feat(spanner): implement at-least-once Commit

### LINKED ISSUE #4890: Implement non-idempotent Commit (write at least once).
For earlier context see https://github.com/googleapis/google-cloud-cpp-spanner/issues/689

The summary is that we decided that our default `Commit` implementation should explicitly call `BeginTransaction` if no transaction was started, to avoid potentially surprising behavior that occurs if we inline transaction creation with `Commit`.  However, there may be cases where the user wants that behavior (since it avoids an extra RPC), so we should expose a method to do that.

devbww:
Add a new operation, `spanner::Client::CommitAtLeastOnce()`, with all its caveats, that exposes the single-RPC, temporary-transaction commit support.

Fixes #4890.

### DIFF
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -184,7 +184,7 @@ class Client {
    * @copydoc Read
    *
    * @param transaction_options Execute this read in a single-use transaction
-   * with these options.
+   *     with these options.
    */
   RowStream Read(Transaction::SingleUseOptions transaction_options,
                  std::string table, KeySet keys,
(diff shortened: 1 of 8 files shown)


google/cloud/spanner/client.h
```diff
@@ -603,6 +603,34 @@ class Client {
...
+   * without replay protection.  That is, it is possible that the mutations
+   * will be applied more than once. If the mutations are not idempotent, this
+   * may lead to a failure (for example, an insert may fail with "already
+   * exists" even though the row did not exist before the call was made).
+   * Accordingly, this call may only be appropriate for idempotent, latency-
+   * sensitive and/or high-throughput blind writing.
+   *
+   * @note Prefer the `Commit` overloads if you want exactly-once semantics
+   *     or want to reapply mutations after a `kAborted` error.
+   *
```
coryan, line 619: It would be nice to have a really simple example using this function.

coryan (approved):

google/cloud/spanner/client.h
devbww: Good idea.  I'll add one in a follow-up.

PR #11916, opened 2023-06-19
feat(common): Added proxy options for curl

HoraceBianchon:
Added proxy options for curl.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11916)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ cmake-build-*/
 # Ignore Visual Studio Code files
 .vsbuild/
 .vscode/
+/out/build/x64-Debug
+/.vs
--- a/google/cloud/google_cloud_cpp_common.bzl
+++ b/google/cloud/google_cloud_cpp_common.bzl
@@ -96,6 +96,7 @@ google_cloud_cpp_common_hdrs = [
     "polling_policy.h",
     "project.h",
     "retry_policy.h",
+    "proxy_options.h",
     "status.h",
     "status_or.h",
     "stream_range.h",
--- a/google/cloud/google_cloud_cpp_common.cmake
+++ b/google/cloud/google_cloud_cpp_common.cmake
@@ -146,6 +146,7 @@ add_library(
     project.cc
     project.h
     retry_policy.h
+    proxy_options.h
     status.cc
     status.h
     status_or.h
@@ -376,6 +377,7 @@ if (BUILD_TESTING)
         options_test.cc
         polling_policy_test.cc
         project_test.cc
+		proxy_options_test.cc
         status_or_test.cc
         status_test.cc
         stream_range_test.cc
--- a/google/cloud/google_cloud_cpp_common_unit_tests.bzl
+++ b/google/cloud/google_cloud_cpp_common_unit_tests.bzl
@@ -70,6 +70,7 @@ google_cloud_cpp_common_unit_tests = [
     "options_test.cc",
     "polling_policy_test.cc",
     "project_test.cc",
+    "proxy_options_test.cc",
     "status_or_test.cc",
     "status_test.cc",
     "stream_range_test.cc",
--- a/google/cloud/internal/curl_handle_factory.cc
+++ b/google/cloud/internal/curl_handle_factory.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/internal/curl_handle_factory.h"
 #include "google/cloud/credentials.h"
 #include "google/cloud/internal/curl_options.h"
+#include "absl/strings/str_format.h"
 #include <algorithm>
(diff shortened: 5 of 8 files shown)


coryan:
First, thanks for the PR.  Some high-level observations before we jump into details of a review:

> Added proxy options for curl.

In general, we recommend that people use the `http_proxy` environment variable.  I believe these changes are needed because no environment variable supports proxy authentication.  Did I get that right?

If I did, then this change or something like it looks like a missing feature.  However for this PR:

- I think these options should live in `google/cloud/common_options.h`, if they can be made to work with gRPC via `GRPC_ARG_HTTP_PROXY` (I think they can), and we have the (experimental) gRPC plugin for GCS to consider (we would like most options to work with either transport).
- We need to figure out how to support `https://` proxies. This only supports `http://`.
- curl supports more than just basic authentication (see [CURLOPT_PROXYAUTH] and [CURLOPT_HTTPAUTH]). I am not sure if we should support those (because gRPC does not seem to support them).

With that in mind:

1. If you would rather file a feature request, we can take it from there, don't feel obliged to work in this PR as it will get more complicated.
2. If you do decide to work on this PR:
    - We need to figure out if we use a separate option for the `ProxyServerSchemeOption` (with values `https` or `http` (the default)) or we just use a `ProxyServerUriOption` (e.g. `http://my-proxy.example.com:8080`).  I am inclined for the second, as it seems easier to use in most cases.
    - If we use a separate option for the scheme, and keep the `ProxyServerHostPortOption` we need to make this work for empty port values.
    - For gRPC, we would need to parse `ProxyServerUriOption` to inject the username and password in exactly the right place.
    - Alternatively, we could just consume a URI (well, everything before the [`path` element][1]) in `ProxyServerUriOption` and pass that directly to gRPC.  For CURL we would parse it, extract the username and password (if present) and pass that to the corresponding options in the proxy.
 
Thoughts?

[CURLOPT_PROXYAUTH]: https://curl.se/libcurl/c/CURLOPT_PROXYAUTH.html
[CURLOPT_HTTPAUTH]: https://curl.se/libcurl/c/CURLOPT_HTTPAUTH.html
[1]: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

HoraceBianchon:
> * I think these options should live in `google/cloud/common_options.h`, if they can be made to work with gRPC via `GRPC_ARG_HTTP_PROXY` (I think they can), and we have the (experimental) gRPC plugin for GCS to consider (we would like most options to work with either transport).

I have never tried gRPC in GCS, so I need some research on how it is constructed. If you can give me some docs, it would be better.


> * We need to figure out how to support `https://` proxies. This only supports `http://`.

Agree with you.

> * curl supports more than just basic authentication (see [CURLOPT_PROXYAUTH](https://curl.se/libcurl/c/CURLOPT_PROXYAUTH.html) and [CURLOPT_HTTPAUTH](https://curl.se/libcurl/c/CURLOPT_HTTPAUTH.html)). I am not sure if we should support those (because gRPC does not seem to support them).

I don't think supporting other authentications is so necessary. If sometimes some feature requests would be for other types of authentication, then can discuss them at that time.

> 1. If you would rather file a feature request, we can take it from there, don't feel obliged to work in this PR as it will get more complicated.

I need GCS with proxies in one of my current projects(proxies are obligated), so I need to finish the PR.

>    * We need to figure out if we use a separate option for the `ProxyServerSchemeOption` (with values `https` or `http` (the default)) or we just use a `ProxyServerUriOption` (e.g. `http://my-proxy.example.com:8080`).  I am inclined for the second, as it seems easier to use in most cases.

I agree with you. I am going to use your mentioned ProxyServerUriOption. It is easier than my version.

>    * For gRPC, we would need to parse `ProxyServerUriOption` to inject the username and password in exactly the right place.

Do you mean to use `ProxyServerUriOption` for setting also username/password? If so, I am not sure it is a good idea.


I am going to do some research on gRPC side.

coryan:
@HoraceBianchon I am sure you saw the errors.  I have been trying to get to this and make more helpful suggestions, but just scanning the logs it looks like some `BUILD.bazel` file needs an additional dependency.  Bazel is very strict about hiding `#include` files that are not explicitly part of a dependency.

I still think we need to think about how this will work with gRPC-based services, including the GCS+gRPC plugin.

coryan:
I think #12766 may be a better solution for this.  It works with gRPC and REST-based client libraries.  It also side-steps the issue of URL-encoding usernames and/or passwords with `:` in them.

coryan:
Closing now that #12766 is merged.

PR #11937, opened 2023-06-21
fix(common): pagination must support empty pages

### LINKED ISSUE #11936: Pagination APIs can stop too early
The wrappers that convert pagination APIs into ranges (and iterators) may stop too early if the service returns a page with no empty and a non-empty pagination token.

Normally I do not create bugs for things fixed with a simple PR.  In this case I think it would be useful to have something to link to.

coryan:
Some services return empty pages with a non-empty pagination token. Our code would stop such iterations immediately. The correct behavior is to retrieve the next page until some non-empty range is returned.

Fixes #11936

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11937)

### DIFF
--- a/google/cloud/internal/pagination_range.h
+++ b/google/cloud/internal/pagination_range.h
@@ -93,8 +93,7 @@ class PagedStreamReader {
    *   successful end of stream.
    */
   typename StreamReader<T>::result_type GetNext() {
-    if (current_ == page_.end()) {
-      if (last_page_) return Status{};
+    while (current_ == page_.end() && !last_page_) {
       request_.set_page_token(std::move(token_));
       auto response = loader_(request_);
       if (!response.ok()) return std::move(response).status();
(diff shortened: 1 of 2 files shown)


devbww (approved):

google/cloud/internal/pagination_range.h
```diff
@@ -93,18 +93,17 @@ class PagedStreamReader {
...
       if (!response.ok()) return std::move(response).status();
       token_ = ExtractPageToken(*response);
       if (token_.empty()) last_page_ = true;
       page_ = extractor_(*std::move(response));
       current_ = page_.begin();
-      if (current_ == page_.end()) return Status{};
     }
-    return std::move(*current_++);
+    if (current_ != page_.end()) return std::move(*current_++);
+    return Status{};
```
devbww, line 106: [Oops ... a comment seems to have been lost ... undoubtedly my mistake.]

Opt nit:  I would reverse the logic here.
```
    if (current_ == page_.end()) return Status{};
    return std::move(*current_++);
```

devbww (approved):

google/cloud/internal/pagination_range.h
coryan: Done.

PR #11939, opened 2023-06-21
feat: Add max_batching_delay option to CommitRequest.

nginsberg-google:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11939)


google/cloud/spanner/commit_options.h
```diff
@@ -70,6 +71,16 @@ class CommitOptions {
     return request_priority_;
   }
 
+  CommitOptions& set_max_batching_delay(
+      absl::optional<absl::Duration> max_batching_delay) {
```
coryan, line 75: We should not use `absl::Duration` in a public API.

google/cloud/spanner/samples/samples.cc
```diff
@@ -3802,6 +3802,16 @@ void QueryInformationSchemaDatabaseOptions(
 }
 // [END spanner_query_information_schema_database_options]
 
+// [START spanner_set_batching_delay]
+void CommitWithBatchingDelay(google::cloud::spanner::Client client,
```
coryan, line 3806: This example does not show me anything that I did not know from reading the prototype.  Do we really need it?

Can we change it to do something less obvious?

google/cloud/spanner/commit_options.h
```diff
@@ -19,6 +19,7 @@
 #include "google/cloud/spanner/version.h"
 #include "google/cloud/options.h"
 #include "absl/types/optional.h"
+#include "absl/time/time.h"
```
devbww, line 22: See the `checkers` build failure for problems with include ordering and indentation.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -1017,6 +1017,12 @@ StatusOr<spanner::CommitResult> ConnectionImpl::CommitImpl(
   request.mutable_request_options()->set_priority(
       ProtoRequestPriority(params.options.request_priority()));
 
+  if (params.options.max_batching_delay().has_value()) {
+    *request.mutable_max_batching_delay() =
```
devbww, line 1021: There is no `max_batching_delay` field in the published `CommitRequest`.

google/cloud/spanner/samples/samples.cc
devbww: In particular, `batching_delay` is unused.

google/cloud/spanner/samples/samples.cc
```diff
@@ -3802,6 +3802,16 @@ void QueryInformationSchemaDatabaseOptions(
...
+			     google::cloud::spanner::Mutations mutations,
+			     absl::Duration batching_delay) {
+  google::cloud::spanner::CommitOptions options;
```
devbww, line 3809: `CommitOptions` is a deprecated part of the public API.  Whatever is needed here should be expressed in terms of `google::cloud::Options`, and then by calling another `Commit()` overload.

google/cloud/spanner/samples/samples.cc
```diff
@@ -3802,6 +3802,16 @@ void QueryInformationSchemaDatabaseOptions(
 }
 // [END spanner_query_information_schema_database_options]
 
+// [START spanner_set_batching_delay]
```
devbww, line 3805: This tag does not appear in the published list.  If we're making up a new one, it should at least include "commit".

google/cloud/spanner/samples/samples.cc
devbww, line 3806: The samples should be executed, but no one calls this.

coryan:
@nginsberg-google do you think you will have time to work on this PR?  If not, I think we will just close it.

nginsberg-google:
Yes, I will still work on this PR. Sorry for the delay. Please don't close it :)

google/cloud/spanner/commit_options.h
nginsberg-google: When I look at that build failure it appears empty. When I click "see more information" it looks like I don't have permission. Am I missing something?

google/cloud/spanner/commit_options.h
nginsberg-google: Done, changed to type int and renamed to `max_batching_delay_ms`.

google/cloud/spanner/internal/connection_impl.cc
nginsberg-google: This field is currently hidden by a TRUSTED_TESTER label. My impression was the order in which I was supposed to do things was:

1. Get this PR (and an equivalent one for other client libraries) approved
2. Remove TT label
3. Merge these PRs.

I have changed my version of this codebase to include the newly generated code when the TT label is removed, but do not have permission to add it to a branch on this codebase. Please let me know if I'm doing something wrong.

google/cloud/spanner/samples/samples.cc
nginsberg-google: Updated.

google/cloud/spanner/samples/samples.cc
nginsberg-google: I've changed the tag. This is new - is there somewhere I need to add it that I've missed?

google/cloud/spanner/samples/samples.cc
nginsberg-google: Done.

google/cloud/spanner/samples/samples.cc
nginsberg-google: I **think** I've done this. This field is much like RequestPriorityOption, and I've pretty much copied it. I don't think we need to overload Commit() again, as the current versions all take an Options field. Please take a look and let me know if I've messed this up.

nginsberg-google (commented):
Appreciate the detailed review. This is now ready for rereview - PTAL. Thanks!

google/cloud/spanner/internal/connection_impl.cc
coryan: That sequence will not work for C++.  For complicated reasons[^1], the C++ clients can only compile against public protos.  In this case the public version of [CommitRequest](https://github.com/googleapis/googleapis/blob/e8cca30589a3fadf8e93954bd30197ef12a39ff8/google/spanner/v1/spanner.proto#L990) does not have the field you want to use.


[^1]: The C++ client libraries cannot enforce what version of Protobuf our customers will use, and the generated `*.pb.h` and `*.pb.cc` files can only work with the version of Protobuf that generated them. Therefore, these files are not included in the Cloud C++ source, the customer generates these files, and the customer only has access to the public proto files.

google/cloud/spanner/samples/samples.cc
```diff
@@ -3951,6 +3951,22 @@ void QueryInformationSchemaDatabaseOptions(
...
+void UpdateDataWithBatchingDelay(google::cloud::spanner::Client client) {
+  namespace spanner = ::google::cloud::spanner;
+  Options ops;
+  ops.set<MaxBatchingDelayMsOption>(100);
+  auto commit_result = client.Commit(spanner::Mutations{
+      spanner::UpdateMutationBuilder("Albums",
+                                     {"SingerId", "AlbumId", "MarketingBudget"})
+          .EmplaceRow(1, 1, 100000)
+          .EmplaceRow(2, 2, 500000)
+          .Build()}, ops);
```
coryan, line 3964:
```suggestion
  auto commit_result = client.Commit(spanner::Mutations{
      spanner::UpdateMutationBuilder("Albums",
                                     {"SingerId", "AlbumId", "MarketingBudget"})
          .EmplaceRow(1, 1, 100000)
          .EmplaceRow(2, 2, 500000)
          .Build()},
          google::cloud::Options{}.set<spanner::MaxBatchDelayOption>(std::chrono::milliseconds(100));
```

google/cloud/spanner/options.h
```diff
@@ -232,6 +232,15 @@ struct RequestPriorityOption {
...
 
+/**
+ * Option for `google::cloud::Options` to set a maximum batching delay.
+ *
+ * @ingroup spanner-options
+ */
+struct MaxBatchingDelayMsOption {
+  using Type = int;
+};
+
```
coryan, line 243: I do not think we would want to use `int` to represent time intervals:

```suggestion
struct MaxBatchingDelayOption {
  using Type = std::chrono::milliseconds;
};

```

google/cloud/spanner/commit_options.h
```diff
@@ -70,6 +70,16 @@ class CommitOptions {
     return request_priority_;
   }
 
+  CommitOptions& set_max_batching_delay_ms(
```
coryan, line 73: @devbww do we want to keep adding methods to this class?  Can we support this new feature *only* via `google::cloud::Options{}`?

coryan (changes requested):

google/cloud/spanner/internal/connection_impl.cc
nginsberg-google: Hm, maybe I'm missing something. Step (2) in the sequence I listed will make the CommitRequest proto contain this field. Once we have a PR for each client library that the client library team is happy with, I will remove the visibility label internally and then the public version of CommitRequest will contain the max_batching_delay field. At that point we will be able to merge this PR. Is there a different sequence of steps the client library team usually takes?

coryan: You need at least:

2.5) Update the googleapis SHA in this repository to use a newer version.
2.6) Get the code to actually compile and pass under CI.

>  Is there a different sequence of steps the client library team usually takes?

I usually don't start writing code in C++ until the protos are published.  I think @devbww has created patches for the proto files and then tweaks the build scripts to apply these patches (there are placeholders in [CMake](https://github.com/googleapis/google-cloud-cpp/blob/611d8445058473b6776abd94d6a8b501243fc36a/external/googleapis/CMakeLists.txt#L139-L149) and [Bazel](https://github.com/googleapis/google-cloud-cpp/blob/611d8445058473b6776abd94d6a8b501243fc36a/bazel/google_cloud_cpp_deps.bzl#L151-L158) to support this).  For obvious reasons that work happens in branches that do not get published to PRs or any public location.

devbww: As @coryan points out, we really need the new protos published publicly, and google-cloud-cpp reconfigured to use the new protos, before worrying too much about actually using those new protos.

That said, we're happy to comment on the state of the PR, but having it in a non-compilable state make things more difficult.

coryan:
FWIW, most builds are useless in this case.  Except for this:

https://github.com/googleapis/google-cloud-cpp/pull/11939/checks?check_run_id=15378864615

google/cloud/spanner/samples/samples.cc
devbww: Firstly, it looks like you only updated the START tag, and not the END one.

But to the bigger issue, the point of these tags is to delimit regions that will be scraped into the Cloud Spanner documentation, so introducing a tag requires some cross-team coordination.  For example, here are all the defined Spanner tags containing "commit": http://devrel/snippets?pageSize=100&sort=region_tag&direction=asc&product_id=9&region_tag=%25commit%25

Normally the tech writers would introduce the tag when they document the feature, otherwise it has no purpose.

google/cloud/spanner/samples/samples.cc
devbww: All the current overloads take an `Options` argument, not a `CommitOptions`.  The latter is deprecated, and we shouldn't be changing it.  Whatever is needed here should **only** be expressed in terms of `google::cloud::Options`.

google/cloud/spanner/commit_options.h
devbww: Sorry for the confusion.  As @coryan pointed out, "should not use `absl::Duration` in a public API" means to use `std::chrono::duration` instead.

google/cloud/spanner/commit_options.h
devbww: There are other `checkers` failures too.  Let me know if you still can't see them and we can try to fix that and/or quote them for you.

google/cloud/spanner/commit_options.h
devbww: No to `spanner::CommitOptions`.  Yes, to `google::cloud::Options`.

google/cloud/spanner/samples/samples.cc
devbww: I'm sorry ... my "only" was too strict.  We still need your modifications to `CommitOptions` as they are used internally (and for mocking).  But we don't want to use `CommitOptions` as part of any sample.  Instead, the sample should use `MaxBatchingDelayOption` directly.

coryan:
Closing for now. It can be reopened if needed or the same branch can be used to push a new PR.

PR #11940, opened 2023-06-22
cleanup(pubsub): use double-quote include syntax

devbww:
Use `#include "..."` for in-project headers, the goal being to eliminate a special-case transformation used during the Google import, that recently displayed its fragility.

### DIFF
--- a/google/cloud/pubsub/samples/samples.cc
+++ b/google/cloud/pubsub/samples/samples.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/pubsub/samples/pubsub_samples_common.h"
+#include "google/cloud/pubsub/samples/testdata/schema.pb.h"
 #include "google/cloud/pubsub/schema.h"
 #include "google/cloud/pubsub/schema_client.h"
 #include "google/cloud/pubsub/subscriber.h"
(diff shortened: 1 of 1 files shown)


google/cloud/pubsub/samples/samples.cc
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/pubsub/samples/pubsub_samples_common.h"
+#include "google/cloud/pubsub/samples/testdata/schema.pb.h"
```
coryan, line 16: With MSVC we use angle brackets to say "this is an external (or externally generated) header, reduce the warning level":

https://github.com/googleapis/google-cloud-cpp/blob/afa1e7635abf21d2308689abe007ead22cc20d2c/ci/kokoro/windows/lib/bazel.ps1#L44-L48

https://github.com/googleapis/google-cloud-cpp/blob/afa1e7635abf21d2308689abe007ead22cc20d2c/cmake/EnableWerror.cmake#L32-L38

coryan: ^ That is just an FYI.  Maybe using quotes is the best solution here, and we need to handle the warnings in a different way.

devbww: Looking at the windows/bazel-debug-2022 build log, I don't see any related warnings, but I probably don't know what I'm looking for or where to enable/find it.

For the record, it looks like I made the same change in #10135, but reverted it in #10139 because it "did not address the real [#10135] issue."

coryan: Sold.

coryan (approved):

PR #11952, opened 2023-06-22
impl(otel): monitored resource mapping

dbolduc:
Part of the work for #11775 

Add support for mapping OTel attributes to GCP [Monitored Resources](https://cloud.google.com/monitoring/api/resources).

This is a whole lot of if statements, copied from: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/02fd6f23e8557907cda61ef01c94198dec4ccd71/internal/resourcemapping/resourcemapping.go#L165

(If you are wondering how we keep the logic up to date, there is an end-to-end test suite written by google's OpenTelemetry team. It does things like test the resource detector on GCE/GKE/Cloud Run/etc. It verifies mapping logic like in this PR. It seems like a lot of set up work. I will consider whether it should be required for C++)

---

I considered code like:

```cc
bool PlatformIsGce(std::string const& platform) {
  return platform == "gcp_compute_engine";
}
//etc.

MonitoredResourceProvider MakeProvider(
    opentelemetry::sdk::resource::ResourceAttributes const& attributes) {
  std::string platform;
  auto p = attributes.find(sc::kCloudPlatform);
  if (p != attributes.end()) platform = AsString(p->second);

  if (PlatformIsGce(platform)) return GceInstance();
  if (PlatformIsGke(platform)) {
    if (GkeIsContainer(attributes)) return GkeContainer();
    if (GkeIsPod(attributes)) return GkePod();
    if (GkeIsNode(attributes)) return GkeNode();
    return GkeCluster();
  }
  if (PlatformIsAwsEc2(platform)) return AwsEc2Instance();
  if (IsGenericTask(attributes)) return GenericTask();
  return GenericNode();
```

but I preferred to see the conditionals inline.

---

Also there are some:
```cc
if (foo) {
  WouldHaveFitOnOneLine();
}
```

I thought they looked worse on one line. :shrug:

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11952)

### DIFF
--- a/google/cloud/opentelemetry/internal/monitored_resource.cc
+++ b/google/cloud/opentelemetry/internal/monitored_resource.cc
@@ -1,0 +1,222 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/opentelemetry/internal/monitored_resource.h"
+#include "google/cloud/version.h"
+#include "absl/types/optional.h"
+#include "absl/types/variant.h"
+#include <opentelemetry/common/attribute_value.h>
+#include <opentelemetry/sdk/resource/resource.h>
+#include <opentelemetry/sdk/resource/semantic_conventions.h>
+#include <unordered_map>
+
+namespace google {
+namespace cloud {
+namespace otel_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace {
+
+namespace sc = opentelemetry::sdk::resource::SemanticConventions;
+
+std::string AsString(
+    opentelemetry::sdk::common::OwnedAttributeValue attribute) {
+  if (absl::holds_alternative<std::string>(attribute)) {
+    return absl::get<std::string>(attribute);
+  }
+  // TODO(#11775) - convert other attribute types to strings
+  // Note that our resource detector only uses string attributes.
+  return {};
+}
+
+struct OTelKeyMatch {
+  std::vector<std::string> otel_keys;
+  absl::optional<std::string> fallback = absl::nullopt;
+};
+
+class MonitoredResourceProvider {
+ public:
+  MonitoredResourceProvider(
+      std::string type, std::unordered_map<std::string, OTelKeyMatch> lookup)
+      : type_(std::move(type)), lookup_(std::move(lookup)) {}
+
+  MonitoredResource Process(
+      opentelemetry::sdk::resource::ResourceAttributes const& attributes) {
+    MonitoredResource mr;
+    mr.type = type_;
+    for (auto const& kv : lookup_) {
+      auto found = false;
+      for (auto const& otel_key : kv.second.otel_keys) {
+        auto p = attributes.find(otel_key);
+        if (p != attributes.end()) {
+          found = true;
+          mr.labels[kv.first] = AsString(p->second);
+          break;
+        }
+      }
+      if (!found && kv.second.fallback)
+        mr.labels[kv.first] = *kv.second.fallback;
+    }
+    return mr;
+  }
+
+ private:
+  std::string type_;
+  std::unordered_map<std::string, OTelKeyMatch> lookup_;
+};
+
+MonitoredResourceProvider GceInstance() {
+  return MonitoredResourceProvider("gce_instance",
+                                   {
+                                       {"zone", {{sc::kCloudAvailabilityZone}}},
+                                       {"instance_id", {{sc::kHostId}}},
+                                   });
+}
+
+MonitoredResourceProvider K8sContainer() {
+  return MonitoredResourceProvider(
+      "k8s_container",
+      {
+          {"location", {{sc::kCloudAvailabilityZone, sc::kCloudRegion}}},
+          {"cluster_name", {{sc::kK8sClusterName}}},
+          {"namespace_name", {{sc::kK8sNamespaceName}}},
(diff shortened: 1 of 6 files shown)


google/cloud/opentelemetry/internal/monitored_resource.cc
```diff
@@ -0,0 +1,222 @@
...
+      "gae_instance",
+      {
+          {"location", {{sc::kCloudAvailabilityZone, sc::kCloudRegion}}},
+          {"module_id", {{sc::kFaasName}}},
+          {"version_id", {{sc::kFaasVersion}}},
+          {"instance_id", {{sc::kFaasInstance}}},
+      });
+}
+
+MonitoredResourceProvider AwsEc2Instance() {
```
coryan, line 139: I may have friends at Azure... Why no support for them?

google/cloud/opentelemetry/internal/monitored_resource_test.cc
```diff
@@ -0,0 +1,321 @@
...
+  auto tests = std::vector<TestCase>{
+      {"us-central1-a", "us-central1", "us-central1-a"},
+      {"us-central1-a", absl::nullopt, "us-central1-a"},
+      {absl::nullopt, "us-central1", "us-central1"},
+  };
+  for (auto const& test : tests) {
+    auto attributes = opentelemetry::sdk::resource::ResourceAttributes{
+        {sc::kCloudPlatform, "gcp_kubernetes_engine"},
+        {sc::kK8sClusterName, "test-cluster"},
+        //{sc::kK8sNodeName, "test-node"},
```
coryan, line 155: Why are these commented out?

google/cloud/opentelemetry/internal/monitored_resource.cc
```diff
@@ -0,0 +1,222 @@
...
+
+  MonitoredResource Process(
+      opentelemetry::sdk::resource::ResourceAttributes const& attributes) {
+    MonitoredResource mr;
+    mr.type = type_;
+    for (auto const& kv : lookup_) {
+      auto found = false;
+      for (auto const& otel_key : kv.second.otel_keys) {
+        auto p = attributes.find(otel_key);
+        if (p != attributes.end()) {
```
coryan, line 61: I would have said 

```cc
  if (p == attributes.end()) continue;
```

google/cloud/opentelemetry/internal/monitored_resource.cc
```diff
@@ -0,0 +1,222 @@
...
+          found = true;
+          mr.labels[kv.first] = AsString(p->second);
+          break;
+        }
+      }
+      if (!found && kv.second.fallback)
```
coryan, line 67: nit: need braces since it is too long for a one liner.

google/cloud/opentelemetry/internal/monitored_resource.cc
```diff
@@ -0,0 +1,222 @@
...
+        mr.labels[kv.first] = *kv.second.fallback;
+    }
```
coryan, line 69: I think this should work:

https://en.cppreference.com/w/cpp/algorithm/find_first_of



```suggestion
auto const& k = kv.second.otel_keys;
auto found = std::find_first_of(attributes.begin(), attributes.end(), k.begin(), k.end(),
  [](auto const& attr, auto const& key) { return attr.first == key; });
  mr.labels[kv.first] = found != attributes.end() ? AsString(found->second) :  *kv.second.fallback;
```

dbolduc: That is almost it. We need to find the first `otel_key` that is found in `attributes`. Not the first `attribute` (in an unordered_map) that is found in `otel_keys`.

I used `std::find_first_of()` but I am agnostic as to whether it is any nicer.

As for the fallback logic, I want to skip empty values. I don't think the go implementation does this. I think they make labels like: `{ "g.co/r/generic_task/task_id", "" }`. That just seems unhelpful to me.

google/cloud/opentelemetry/internal/monitored_resource.cc
dbolduc: N/A

google/cloud/opentelemetry/internal/monitored_resource.cc
dbolduc: N/A

google/cloud/opentelemetry/internal/monitored_resource_test.cc
dbolduc: I was being cheeky. I thought it would better convey the conditions of these if statement: https://github.com/googleapis/google-cloud-cpp/blob/6b87fd6484997467a219be2a30d5a41519493433/google/cloud/opentelemetry/internal/monitored_resource.cc#L185-L194

I was inspired by this: https://github.com/googleapis/google-cloud-cpp/blob/0ca17bf611207ab8e2db84d4ceb539e0acdfafa9/google/cloud/internal/oauth2_external_account_credentials_test.cc#L404-L419

Anyways, I removed them.

google/cloud/opentelemetry/internal/monitored_resource.cc
dbolduc: Because I am only capable of copying the work that was done in golang, which does not include Azure.

coryan: Please file a feature request (in google-cloud-cpp).  It is possible we won't get to it, but we should have a bug to point to.

coryan (approved):

google/cloud/opentelemetry/internal/monitored_resource.cc
coryan: Okay.  One of the reasons to use OTel is that our customers may be sending the tracing / monitoring data to something other than Cloud Monitoring.

dbolduc: Yeah, this PR is to support a special mapping from OTel conventions to Cloud Monitoring conventions.

The spec calls for doing this translation, and also dumping the resource attributes as span attributes anyways.

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/182fb273d196283030190c6f8d62cca466edd48a/exporter/trace/trace_proto.go#L100-L128

coryan: Doh

PR #11966, opened 2023-06-26
impl(otel): copy resource attributes

dbolduc:
Part of the work for #11775 

Add resource attributes as span attributes. We do this because the `google::devtools::trace::v2::Span` proto does not have a concept of resources.

The work is annoying because the types are different:
- `Span`'s [AttributeValue](https://github.com/open-telemetry/opentelemetry-cpp/blob/cfcda5728e75e51215d118eccd639e21cf4fd74d/api/include/opentelemetry/common/attribute_value.h#L37-L59)
- `Resource`'s [OwnedAttributeValue](https://github.com/open-telemetry/opentelemetry-cpp/blob/cfcda5728e75e51215d118eccd639e21cf4fd74d/sdk/include/opentelemetry/sdk/common/attribute_utils.h#L30-L44)

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11966)

### DIFF
--- a/google/cloud/opentelemetry/internal/recordable.cc
+++ b/google/cloud/opentelemetry/internal/recordable.cc
@@ -160,6 +160,52 @@ google::devtools::cloudtrace::v2::Span::SpanKind MapSpanKind(
   }
 }
 
+template <typename T>
+opentelemetry::common::AttributeValue convertVector(std::vector<T> const& v) {
+  return opentelemetry::nostd::span<T const>(v.data(), v.size());
+}
+template <>
+opentelemetry::common::AttributeValue convertVector(std::vector<bool> const&) {
+  // We discard the data in vector<bool> attributes because:
+  // - We cannot convert a vector<bool> to a span<bool> without allocating
+  //   memory.
+  // - The exporter drops span<bool> attributes anyways.
+  // - vector<bool> attributes are probably rare.
+  return opentelemetry::nostd::span<bool>{};
+}
+template <>
+opentelemetry::common::AttributeValue convertVector(
+    std::vector<std::string> const& v) {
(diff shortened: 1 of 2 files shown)


google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -160,6 +160,52 @@ google::devtools::cloudtrace::v2::Span::SpanKind MapSpanKind(
...
+template <typename T>
+opentelemetry::common::AttributeValue convertVector(std::vector<T> const& v) {
+  return opentelemetry::nostd::span<T const>(v.data(), v.size());
+}
+template <>
+opentelemetry::common::AttributeValue convertVector(std::vector<bool> const&) {
+  // We discard the data in vector<bool> attributes because:
+  // - We cannot convert a vector<bool> to a span<bool> without allocating
+  //   memory.
+  // - The exporter drops span<bool> attributes anyways.
```
coryan, line 172: I don't buy the first reason, particularly in view of the third reason.

I buy the second reason.  But it does create a trap if we ever fix the exporter 🤷

coryan (approved):

dbolduc:
I am just gonna close this. It was probably dumb for me to try to reuse the span attribute logic by mapping the resource attributes to them.

What I can do instead is map both the span attribute variant and the resource attribute variant to a proto variant: https://github.com/dbolduc/google-cloud-cpp/blob/ef28275b17acc9d94003be38f58fa470ba3b0cc7/google/cloud/opentelemetry/internal/recordable.cc#L77-L132

I'll just convert `vector<T>`, `span<T>` to strings so we can stop dropping these attributes.

Issue #11025, opened 2023-03-10
pubsub BlockingPull integration test receives INTERNAL error against prod


dbolduc:
* GCB: https://console.cloud.google.com/cloud-build/builds;region=us-east1/152d8a74-f984-4105-942b-b6c9deaa2f50;tab=detail?project=cloud-cpp-testing-resources
* Raw: https://storage.googleapis.com/cloud-cpp-community-publiclogs/logs/google-cloud-cpp/11023/c0cf7c68634ab81d013fb3e92619e3c9b6b1ab80/fedora-37-cmake-integration-production/log-152d8a74-f984-4105-942b-b6c9deaa2f50.txt

```
[ RUN      ] SubscriberIntegrationTest.BlockingPull
google/cloud/pubsub/integration_tests/subscriber_integration_test.cc:546: Failure
Value of: response
Expected: code is equal to OK and message is anything
  Actual: 200-byte object <10-98 02-D4 FF-7E 00-00 00-18 45-59 BA-9F 1A-23 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 ... 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00>, whose status is INTERNAL: invalid response, mismatched ID count error_info={reason=INTERNAL, domain=gcloud-cpp, metadata={gcloud-cpp.source.function=Pull, gcloud-cpp.source.line=69, gcloud-cpp.source.filename=google/cloud/pubsub/internal/subscriber_connection_impl.cc, gcloud-cpp.version=v2.9.0-rc}}, with a code that isn't equal to OK, but a message that is anything
```

coryan:
This looks like a real bug.  The response included 0 messages:

```
Step #3: 2023-03-10T01:03:30.679721853Z [DEBUG] <139638922328512> Pull() << google.pubsub.v1.PullRequest {
Step #3:   subscription: "projects/cloud-cpp-testing-resources/subscriptions/cloud-cpp-jdivzmxushyidwhrxiqunpwkrngexndb"
[...]

PR #11983, opened 2023-06-28
fix: pubsub blocking pull integration test (#11025)

alevenberg:
Fixes issue #11025

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11983)

### DIFF
--- a/google/cloud/pubsub/internal/subscriber_connection_impl.cc
+++ b/google/cloud/pubsub/internal/subscriber_connection_impl.cc
@@ -92,11 +92,10 @@ StatusOr<pubsub::PullResponse> SubscriberConnectionImpl::Pull() {
   }
   if (!last_status.ok()) {
     return last_status;
-  } else {
+  } else if (last_status.ok()) {
     return google::cloud::internal::UnavailableError("no messages returned",
                                                      GCP_ERROR_INFO());
-  };
-  if (!retry_policy->IsExhausted()) {
+  } else if (!retry_policy->IsExhausted()) {
     return google::cloud::internal::RetryLoopError("Permanent error in",
                                                    __func__, last_status);
   }
--- a/google/cloud/pubsub/internal/subscriber_connection_impl.cc
+++ b/google/cloud/pubsub/internal/subscriber_connection_impl.cc
@@ -90,16 +90,16 @@ StatusOr<pubsub::PullResponse> SubscriberConnectionImpl::Pull() {
     };
     std::this_thread::sleep_for(backoff_policy->OnCompletion());
   }
-  if (last_status.ok())
+  if (last_status.ok()) {
     return google::cloud::internal::UnavailableError("no messages returned",
                                                      GCP_ERROR_INFO());
-
-  if (!retry_policy->IsExhausted())
-    return google::cloud::internal::RetryLoopError("Permanent error in",
+  }
+  if (retry_policy->IsExhausted()) {
+    return google::cloud::internal::RetryLoopError("Retry policy exhausted in",
                                                    __func__, last_status);
-
-  return google::cloud::internal::RetryLoopError("Retry policy exhausted in",
-                                                 __func__, last_status);
+  }
+  return google::cloud::internal::RetryLoopError("Permanent error in", __func__,
+                                                 last_status);
 }
 
 Options SubscriberConnectionImpl::options() { return opts_; }
--- a/google/cloud/pubsub/internal/subscriber_connection_impl.cc
+++ b/google/cloud/pubsub/internal/subscriber_connection_impl.cc
@@ -60,11 +60,17 @@ StatusOr<pubsub::PullResponse> SubscriberConnectionImpl::Pull() {
       current.get<pubsub::BackoffPolicyOption>()->clone(),
       google::cloud::Idempotency::kIdempotent,
       [stub = stub_](auto& context, auto const& request) {
-        return stub->Pull(context, request);
+        auto r = stub->Pull(context, request);
+
+        if (!r || !r->received_messages().empty()) return r;
+        // No messages received, return a retryable error to try again. This
+        // respects the retry policy.
+        return google::cloud::internal::AbortedError("no messages received");
       },
       request, __func__);
+
   if (!response) return std::move(response).status();
-  if (response->received_messages_size() != 1) {
+  if (response->received_messages_size() > 1) {
     return internal::InternalError("invalid response, mismatched ID count",
                                    GCP_ERROR_INFO());
   }
--- a/google/cloud/pubsub/internal/subscriber_connection_impl.cc
+++ b/google/cloud/pubsub/internal/subscriber_connection_impl.cc
@@ -92,12 +92,12 @@ StatusOr<pubsub::PullResponse> SubscriberConnectionImpl::Pull() {
   }
   if (!last_status.ok()) {
     return last_status;
-  } 
+  }
   if (last_status.ok()) {
     return google::cloud::internal::UnavailableError("no messages returned",
                                                      GCP_ERROR_INFO());
   }
-   if (!retry_policy->IsExhausted()) {
+  if (!retry_policy->IsExhausted()) {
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/internal/subscriber_connection_impl.cc
```diff
@@ -55,16 +55,22 @@ StatusOr<pubsub::PullResponse> SubscriberConnectionImpl::Pull() {
...
+        current.get<pubsub::BackoffPolicyOption>()->clone(),
+        google::cloud::Idempotency::kIdempotent,
+        [stub = stub_](auto& context, auto const& request) {
+          return stub->Pull(context, request);
+        },
+        request, __func__);
+
+    if (!response) return std::move(response).status();
+    messages_received = response->received_messages_size();
+  }
```
coryan, line 72: This implies we have a retry loop inside a retry loop.  That seems odd.  If I (or an application developer writes):

```cc
  auto m = client.Pull(..., Options{}.set<pubsub::RetryPolicyOption>(LimitedTimeRetryPolicy(60s)));
```

I expect the client library to block for at most 60s, not for "60s unless no message arrive in which case block for 60s more"?

Maybe something like this instead?

```suggestion
    auto response = google::cloud::internal::RetryLoop(
        current.get<pubsub::RetryPolicyOption>()->clone(),
        current.get<pubsub::BackoffPolicyOption>()->clone(),
        google::cloud::Idempotency::kIdempotent,
        [stub = stub_](auto& context, auto const& request) {
          auto r = stub->Pull(context, request);
          if (!r || !r->received_messages().empty()) return r;
          // No messages received, return a retryable error to
          // try again. This respects the retry policy.
          return google::cloud::internal::AbortedError("no messages received", GCP_ERROR_INFO());
        },
        request, __func__);
```

coryan: I am not sure `AbortedError` is the right way to represent this condition.  We may need a different error code, or a completely different retry loop.  But the point about the double loop stands I think.

google/cloud/pubsub/internal/subscriber_connection_impl_test.cc
```diff
@@ -409,6 +409,52 @@ TEST(SubscriberConnectionTest, Pull) {
   t.join();
 }
 
+TEST(SubscriberConnectionTest, PullReturnsNoMessage) {
```
coryan, line 412: Consider a test that:

1) Sets the retry policy to tolerate at most N failures.
2) Returns empty messages N times.
3) The response should be an error of some kind.

alevenberg: Added test

alevenberg:
Accidentally reset the initial commit (my bad)

Added a new retry loop and test that you suggested. Also, fixed the off by 1 error in the limited count retry policy.

Tried to run on the CI locally, but my resources aren't set up right. Looking into now.

google/cloud/pubsub/internal/subscriber_connection_impl.cc
alevenberg: Added a custom retry loop.

alevenberg:
I can't tell if it's an off by 1 error or not. A lot of tests break when I changed the IsExhausted() condition on the limited error count retry policy.  

It looks like there are some other places where we set a error count, but then make one more RPC (like in DefaultRowReaderTest, RetryPolicyExhausted [here](https://github.com/alevenberg/google-cloud-cpp/blob/284eadb43d46927fe81655c986467bc56262b7cb/google/cloud/bigtable/internal/default_row_reader_test.cc#L243))

google/cloud/pubsub/internal/subscriber_connection_impl.cc
```diff
@@ -53,31 +55,54 @@ StatusOr<pubsub::PullResponse> SubscriberConnectionImpl::Pull() {
...
-  auto impl = std::make_unique<pubsub_internal::DefaultPullAckHandler>(
-      background_->cq(), stub_, current, std::move(subscription),
-      std::move(*received_message.mutable_ack_id()),
-      received_message.delivery_attempt());
-  auto message = pubsub_internal::FromProto(
-      std::move(*received_message.mutable_message()));
-  return pubsub::PullResponse{pubsub::PullAckHandler(std::move(impl)),
-                              std::move(message)};
+  return google::cloud::internal::RetryLoopError("Retry policy exhausted in",
+                                                 __func__, last_status);
```
devbww, line 105: It's a little hard to tell in the diff, but this looks strange on the face of things.
```
  if (!last_status.ok()) return A;
  if (last_status.ok()) return B;
  if (!retry_policy->IsExhausted()) return C;
  return D;
```
Isn't everything after the second statement unreachable?

alevenberg: That's correct. Removed the first if.

Now, it does the following 1) if the last status was ok, it means the rpc returned 0 messages and will give the unavailable error, 2) if the policy is not exhausted, it was a permanent error, 3) return the policy was exhausted.

I tried removing the braces {}, but the formatting is still a little weird. Any suggestions?

devbww: We generally don't remove braces when the `if` covers multiple lines, so I'd only suggest restoring them.

And one more optional nit ... I'd suggest inverting the final `IsExhausted()` check so that your flow is _"2) if the policy is exhausted, return the policy is exhausted, 3) return permanent error"_, just so the exhausted conditional is more directly related to the exhausted error (and eliminating negations is usually a good thing).

alevenberg: Re added the brackets and agreed, swapped the order of the ifs/return

google/cloud/pubsub/internal/subscriber_connection_impl.cc
```diff
@@ -53,31 +55,51 @@ StatusOr<pubsub::PullResponse> SubscriberConnectionImpl::Pull() {
 
   google::pubsub::v1::PullRequest request;
   request.set_subscription(subscription.FullName());
+  // Pubsub returns at most 1 message.
```
coryan, line 58: nit: maybe "Ask Pub/Sub to return at most 1 message".

google/cloud/pubsub/internal/subscriber_connection_impl.cc
```diff
@@ -53,31 +55,51 @@ StatusOr<pubsub::PullResponse> SubscriberConnectionImpl::Pull() {
...
-      },
-      request, __func__);
-  if (!response) return std::move(response).status();
-  if (response->received_messages_size() != 1) {
-    return internal::InternalError("invalid response, mismatched ID count",
-                                   GCP_ERROR_INFO());
+  auto retry_policy = current.get<pubsub::RetryPolicyOption>()->clone();
+  auto backoff_policy = current.get<pubsub::BackoffPolicyOption>()->clone();
+  google::cloud::Status last_status;
+  google::cloud::StatusOr<google::pubsub::v1::PullResponse> response;
```
coryan, line 64: `response` is only used inside the loop, so maybe move it there and you can use `auto response = ...`.

google/cloud/pubsub/internal/subscriber_connection_impl.cc
```diff
@@ -53,31 +55,51 @@ StatusOr<pubsub::PullResponse> SubscriberConnectionImpl::Pull() {
...
+          received_message.delivery_attempt());
+      auto message = pubsub_internal::FromProto(
+          std::move(*received_message.mutable_message()));
+      return pubsub::PullResponse{pubsub::PullAckHandler(std::move(impl)),
+                                  std::move(message)};
+    }
+    last_status = std::move(response).status();
+    if (!retry_policy->OnFailure(last_status)) {
+      break;
+    };
```
coryan, line 90: nit: extra semi-colon

coryan (approved):
I think this is almost ready.  Just a few nits.

PR #11992, opened 2023-06-29
ci: retry integration tests on failure

coryan:
This is part of the work for #10413

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/11992)

### DIFF
--- a/ci/cloudbuild/builds/lib/integration.sh
+++ b/ci/cloudbuild/builds/lib/integration.sh
@@ -148,6 +148,9 @@ function integration::bazel_with_emulators() {
 
   local verb="$1"
   local args=("${@:2}")
+  # Integration tests are inherently flaky. Make up to three attempts to get the
+  # test passing.
+  args+=(--flaky_test_attempts=3)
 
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/builds/lib/integration.sh
```diff
@@ -271,6 +274,9 @@ function integration::ctest_with_emulators() {
 
   local cmake_out="$1"
   mapfile -t ctest_args < <(ctest::common_args)
+  # The integration tests are inherently flaky.  Try them 3 times before
+  # declaring a failure.
```
devbww, line 278: Nit:  Make the two comments the same so that it is clearer they have the same purpose?

devbww (approved):

ci/cloudbuild/builds/lib/integration.sh
coryan: Fixed and move the setting to a more general place.

dbolduc (approved):

PR #12000, opened 2023-06-30
cleanup: speed up library generation

### LINKED ISSUE #11999: Speed up `generate-libraries.sh`
The slowest part of `generate-libraries.sh` is this:

https://github.com/googleapis/google-cloud-cpp/blob/bdf7cc27c12c338aaa4885a0086fdc253aa5a683/ci/cloudbuild/builds/generate-libraries.sh#L88-L93

This script only really needs to run when the googleapis SHA changes (or when new libraries are generated).  It seems like we could remove it and make this code part of the googleapis SHA steps.

It also seems like a parallelizable step.

coryan:
`generate-libraries.sh` is used to:

- drive a CI build to verify the generated libraries and support files are up to date
- update the libraries after the googleapis SHA changes
- update the libraries after (and during) changes to the generator

The first two cases need to run `update_libraries.sh`. The last case is probably the most common case, and the slowest part was running the `update_libraries.sh`.

I took the opportunity to make the output more readable when making large updates.

Fixes #11999

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12000)

### DIFF
--- a/ci/cloudbuild/builds/generate-libraries.sh
+++ b/ci/cloudbuild/builds/generate-libraries.sh
@@ -85,20 +85,25 @@ else
     xargs -r -P "$(nproc)" -n 50 -0 sed -i 's/[[:blank:]]\+$//'
 fi
 
-if [ -z "${GENERATE_GOLDEN_ONLY}" ]; then
+if [[ "${TRIGGER_TYPE:-}" != "manual" ]]; then
   io::log_h2 "Updating protobuf lists/deps"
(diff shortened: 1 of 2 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12000)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @coryan)

Issue #12010, opened 2023-07-04
Support reverse scans in Bigtable


dbolduc:
The [reverse scans](https://cloud.google.com/bigtable/docs/reads#reverse-scan) feature is now in "experimental public preview" or something.

The googleapis SHA is up to date. And we already have a `bigtable::ReadRowsParams`. So the remaining work is to:

- [x] Include `bigtable-features: ...` header.
- [x] Add `reverse` flag to lower layers of API (I think everything below `ReadRowsParams`)
- [ ] Release feature (1 PR)
  - [ ] Add `bigtable::ReverseScanOption` (and use it)
  - [ ] Add integration test
  - [ ] Add `bigtable_reverse_scan` sample

This is more or less where we are going: https://github.com/dbolduc/google-cloud-cpp/commit/fb6fba3f29bab33805ccd467774f6bc501ff7bbe

Also, the emulator does not currently support this flag. I sent https://github.com/googleapis/google-cloud-go/pull/8198, we will see how it is received. Even if that PR is merged rolling out the emulator can take weeks.

PR #12013, opened 2023-07-04
impl(bigtable): reverse scan internals

dbolduc:
Part of the work for #12010 

- Have sync/async row readers set the `reversed` flag
- Teach sync/async row readers how to resume a reverse scan
- Teach the row parser about reverse row orders

The diff might have been prettier had I defaulted the `bool reverse` used to initialize a `ReadRowsParser`. I know at least one person on the team who does not like to use defaulted arguments for internal-only APIs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12013)

### DIFF
--- a/google/cloud/bigtable/internal/async_row_reader.cc
+++ b/google/cloud/bigtable/internal/async_row_reader.cc
@@ -30,6 +30,7 @@ void AsyncRowReader::MakeRequest() {
 
   request.set_app_profile_id(app_profile_id_);
   request.set_table_name(table_name_);
+  request.set_reversed(reverse_);
   auto row_set_proto = row_set_.as_proto();
   request.mutable_rows()->Swap(&row_set_proto);
 
@@ -39,7 +40,7 @@ void AsyncRowReader::MakeRequest() {
   if (rows_limit_ != NO_ROWS_LIMIT) {
     request.set_rows_limit(rows_limit_ - rows_count_);
   }
-  parser_ = bigtable::internal::ReadRowsParserFactory().Create();
+  parser_ = bigtable::internal::ReadRowsParserFactory().Create(reverse_);
 
   internal::ScopedCallContext scope(call_context_);
   auto context = std::make_shared<grpc::ClientContext>();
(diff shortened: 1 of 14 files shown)


google/cloud/bigtable/internal/default_row_reader.cc
```diff
@@ -107,8 +109,13 @@ absl::variant<Status, bigtable::Row> DefaultRowReader::Advance() {
     if (!last_read_row_key_.empty()) {
       // We've returned some rows and need to make sure we don't
       // request them again.
-      row_set_ =
-          row_set_.Intersect(bigtable::RowRange::Open(last_read_row_key_, ""));
+      if (reverse_) {
+        row_set_ = row_set_.Intersect(
+            bigtable::RowRange::Open("", last_read_row_key_));
```
coryan, line 114: Same comment as above.  For all I know this is the "Right Thing". If it is, it deserves a comment explaining why.

coryan (changes requested):

google/cloud/bigtable/internal/async_row_reader.cc
```diff
@@ -183,8 +184,13 @@ void AsyncRowReader::OnStreamFinished(Status status) {
   if (!last_read_row_key_.empty()) {
     // We've returned some rows and need to make sure we don't
     // request them again.
-    row_set_ =
-        row_set_.Intersect(bigtable::RowRange::Open(last_read_row_key_, ""));
+    if (reverse_) {
+      row_set_ =
+          row_set_.Intersect(bigtable::RowRange::Open("", last_read_row_key_));
```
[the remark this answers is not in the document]
dbolduc: > That is the empty set if `last_read_row_key_` is non-empty

No. `"" < k` for all non-empty `k` .

---

The thing that is confusing is that an end key of `""` means end of table. Setting the start key to `""` is equivalent to not setting the start key at all.

So let's just do that and hope the code speaks better for itself.

google/cloud/bigtable/internal/default_row_reader.cc
dbolduc: Ditto

coryan (approved):

PR #12032, opened 2023-07-05
cleanup: reduce tarball size

coryan:
Exclude the ci/abi-dumps directory from the tarball, saving about 78MiB from the tarball, or about 30% of the size.


devbww (approved):

dbolduc:
🧠

PR #12038, opened 2023-07-06
impl(otel): trace exporter handles composite attributes

dbolduc:
Part of the work for #11775 

The cloud trace proto only supports attributes of type: `bool`, `int`, or `string`. So convert composite attributes into a string.

I picked "|" as a separator. :shrug:

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12038)

### DIFF
--- a/google/cloud/opentelemetry/internal/recordable.cc
+++ b/google/cloud/opentelemetry/internal/recordable.cc
@@ -55,19 +55,25 @@ void MapKey(opentelemetry::nostd::string_view& key) {
 
 template <typename T>
 std::string ToString(opentelemetry::nostd::span<T const> values) {
-  return absl::StrJoin(values, "|");
+  return absl::StrCat("[", absl::StrJoin(values, ", "), "]");
 }
 template <>
 std::string ToString(
     opentelemetry::nostd::span<opentelemetry::nostd::string_view const>
         values) {
-  return absl::StrJoin(values, "|", absl::StreamFormatter());
+  return absl::StrCat(
+      R"""([")""",
+      absl::StrJoin(values, R"""(", ")""", absl::StreamFormatter()),
+      R"""("])""");
 }
 template <>
 std::string ToString(opentelemetry::nostd::span<bool const> values) {
-  return absl::StrJoin(values, "|", [](std::string* out, bool v) {
-    out->append(v ? "true" : "false");
-  });
+  return absl::StrCat("[",
+                      absl::StrJoin(values, ", ",
+                                    [](std::string* out, bool v) {
+                                      out->append(v ? "true" : "false");
+                                    }),
+                      "]");
 }
 
 class AttributeVisitor {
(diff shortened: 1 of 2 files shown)


google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -107,8 +125,11 @@ class AttributeVisitor {
   // There is no mapping from a `span<T>` to the Cloud Trace proto. We just drop
   // these attributes.
   template <typename T>
-  void operator()(opentelemetry::nostd::span<T>) {
-    Drop();
+  void operator()(opentelemetry::nostd::span<T> value) {
```
coryan, line 128: Is the comment still appropriate?

google/cloud/opentelemetry/internal/recordable.cc
```diff
@@ -52,6 +53,23 @@ void MapKey(opentelemetry::nostd::string_view& key) {
   if (it != m->end()) key = it->second;
 }
 
+template <typename T>
+std::string ToString(opentelemetry::nostd::span<T const> values) {
+  return absl::StrJoin(values, "|");
```
coryan, line 58: Whether `|` is a good separator depends on the context.  Maybe formatting as an array is better:

```cc
return absl::StrCat("[", absl::StrJoin(values, ", "), "]")
```

In some contexts `\003` (ETX) may be appropriate:

https://en.wikipedia.org/wiki/End-of-Text_character

or one of the "separators" (`\x1C' to `\x1F`):

https://en.wikipedia.org/wiki/C0_and_C1_control_codes

Though that would make the string non-printable.

google/cloud/opentelemetry/internal/recordable_test.cc
```diff
@@ -67,9 +67,11 @@ class KVIterable : public opentelemetry::common::KeyValueIterable {
 };
 
 template <typename T>
-opentelemetry::nostd::span<T> MakeCompositeAttribute() {
-  T v[]{T{}, T{}};
-  return opentelemetry::nostd::span<T>(v);
+opentelemetry::nostd::span<T const> MakeCompositeAttribute(T t1, T t2) {
+  // We use static storage, so make sure that if this method is called multiple
+  // times for a given T, it is called with the same t1 and t2.
+  static auto const* const kData = new T[2]{t1, t2};
```
coryan, line 73: This is a bit scary, but in a test is Okay I guess.

coryan (approved):

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: Nope. Fixed.

google/cloud/opentelemetry/internal/recordable.cc
dbolduc: I don't think there is a perfect answer. We will go with your array suggestion.

PR #12057, opened 2023-07-10
cleanup(bigtable): websafe encoding of feature flags

dbolduc:
cc: @igorbernstein2 who pointed out that the feature flags metadata should be web safe escaped.

Aside: there are probably several uses of `internal::Base64Encoder` that could be replaced with the abseil implementation. I do not plan to refactor them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12057)

### DIFF
--- a/google/cloud/bigtable/internal/bigtable_stub_factory_test.cc
+++ b/google/cloud/bigtable/internal/bigtable_stub_factory_test.cc
@@ -256,14 +256,14 @@ TEST_F(BigtableStubFactory, AsyncMutateRow) {
 }
 
 TEST_F(BigtableStubFactory, FeaturesFlags) {
+  auto constexpr kWebSafeBase64Regex = "[A-Z0-9_-]+";
   MockFactory factory;
   EXPECT_CALL(factory, Call)
-      .WillOnce([](std::shared_ptr<grpc::Channel> const&) {
+      .WillOnce([=](std::shared_ptr<grpc::Channel> const&) {
         auto mock = std::make_shared<MockBigtableStub>();
         EXPECT_CALL(*mock, MutateRow)
-            .WillOnce([](grpc::ClientContext& context,
-                         google::bigtable::v2::MutateRowRequest const&) {
-              auto constexpr kWebSafeBase64Regex = "[A-Z0-9_-]+";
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/internal/bigtable_stub_factory_test.cc
```diff
@@ -259,15 +264,15 @@ TEST_F(BigtableStubFactory, AsyncMutateRow) {
 TEST_F(BigtableStubFactory, FeaturesFlags) {
   MockFactory factory;
   EXPECT_CALL(factory, Call)
-      .WillOnce([](std::shared_ptr<grpc::Channel> const&) {
+      .WillOnce([=](std::shared_ptr<grpc::Channel> const&) {
```
devbww, line 267: What local objects are we wanting to capture?

devbww (commented):
Is the PR description mention of abseil now obsolete?

google/cloud/bigtable/internal/bigtable_stub_factory_test.cc
dbolduc: D'oh those were artifacts from a previous commit. Removed.

google/cloud/bigtable/internal/bigtable_stub_factory_test.cc
```diff
@@ -41,11 +42,15 @@ using ::google::cloud::testing_util::StatusIs;
...
 using ::testing::Contains;
 using ::testing::HasSubstr;
-using ::testing::IsEmpty;
-using ::testing::Not;
 using ::testing::NotNull;
+using ::testing::Pair;
 using ::testing::Return;
 
+MATCHER(IsWebSafeBase64, "") {
+  std::regex regex(R"re([A-Z0-9_-]+)re");
```
devbww, line 50: Is the empty string WebSafeBase64?  If so, you may prefer to keep `IsWebSafeBase64()` and `Not(IsEmpty())` orthogonal.

dbolduc: Done.

devbww (approved):

PR #12064, opened 2023-07-11
feat(oauth2): new library to create oauth2 tokens

### LINKED ISSUE #11920: Create a public API to get access token
The APIs to get (and cache) access tokens are all in `*internal*` namespaces.  We need a simple API to create access tokens.

Maybe something like this:

```cc
namespace google::cloud {
struct AccessToken { std::string token; std::chrono::time_point expiration; };

class AccessTokenGenerator {
 public:
   virtual ~AccessTokenGenerator() = default;
   virtual StatusOr<AccessToken> GetToken() = 0;
};

std::shared_ptr<AccessTokenGenerator> MakeAccessTokenGenerator(std::shared_ptr<Credentials>);
```

That is a fairly minimal API, whatever our implementation details we will always have enough to support it. 

We need to think about the name of the libraries too.  All of this lives in `rest_internal` which suggests it is not a public library.  We may need to create `google_cloud_cpp_auth` ?

coryan:
This is basically a wrapper around some classes in the `rest_internal` library.  We need a new library because applications should not link `rest_internal`: all libraries with `internal` in their name are not part of the public API.

Fixes #11920

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12064)


scotthart (commented):
If we anticipate more public OAuth2 files/classes, should we consider adding a google/cloud/oauth2 directory for these to live in, as opposed to google/cloud?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12064#-:-N_4uGexBZVm1eoZ3Yvb:bcob5ws)* status: 0 of 14 files reviewed, all discussions resolved

ci/cloudbuild/builds/check-api.sh
```diff
@@ -39,14 +39,14 @@ io::run cmake "${cmake_args[@]}" \
   -DGOOGLE_CLOUD_CPP_ENABLE="${ENABLED_FEATURES}" \
   -DCMAKE_CXX_FLAGS="-Og -Wno-maybe-uninitialized"
 io::run cmake --build cmake-out
-io::run cmake --install cmake-out >/dev/null
+io::run cmake --install cmake-out #>/dev/null
```
devbww, line 42: Remnant or feature?

google/cloud/access_token_generator.cc
```diff
@@ -0,0 +1,48 @@
...
+      std::shared_ptr<google::cloud::oauth2_internal::Credentials> impl)
+      : impl_(std::move(impl)) {}
+  ~AccessTokenGeneratorImpl() override = default;
+
+  StatusOr<AccessToken> GetToken() override {
+    return impl_->GetToken(std::chrono::system_clock::now());
+  }
+
+ private:
+  std::shared_ptr<google::cloud::oauth2_internal::Credentials> impl_;
```
devbww, line 35: Is `impl_` the right name, versus, say `creds_`?

google/cloud/access_token_generator.cc
```diff
@@ -0,0 +1,48 @@
...
+};
+
+}  // namespace
+
+std::shared_ptr<AccessTokenGenerator> MakeAccessTokenGenerator(
+    std::shared_ptr<Credentials> credentials) {  // NOLINT
```
devbww, line 41: Aside: What lint is that?

ci/cloudbuild/builds/check-api.sh
coryan: Remnant, fixed.

google/cloud/access_token_generator.cc
coryan: I can argue both sides of that.  And sound like I believe both of them too!  Changed to `creds_`.

google/cloud/access_token_generator.cc
coryan: Passing by value and only using as a `const&`.  I think we should consume `Credentials const&`, but that requires further refactoring.  I will do that and then rebase this PR.

coryan (commented):
I am thinking about it.  Sounds like a good idea. Let me do some other refactoring first and maybe discuss in the meeting tomorrow (won't merge until after that discussion).

> Reviewed 11 of 13 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12064#-:-N_6BE809yuvuO-Mzk8w:b-1bhq4u)* status: all files reviewed, 3 unresolved discussions (waiting on @devbww)

devbww (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12064)* status: all files reviewed, 1 unresolved discussion (waiting on @coryan)

devbww (approved):

PR #12088, opened 2023-07-13
docs(pubsub): add cloud storage subscription sample

alevenberg:
Added the sample, but didn't include it in our CI test since it requires creating a storage bucket. CreateBigQuerySubscription is also not in our CI testing.

Verified using: `bazel run //google/cloud/pubsub/samples:samples create-cloud-storage-subscription alevenb-test test-gcs-ordered-topic notification alevenb-test-bucket-staging`


coryan (commented):
Shouldn't we run this as part of the `AutoRun()` function?

alevenberg:
Added the sample to AutoRun

coryan (approved):

google/cloud/pubsub/samples/samples.cc
```diff
@@ -400,6 +400,32 @@ void CreateBigQuerySubscription(
...
+    auto sub = client.CreateSubscription(
+        pubsub::Topic(project_id, std::move(topic_id)),
+        pubsub::Subscription(project_id, std::move(subscription_id)),
+        pubsub::SubscriptionBuilder{}.set_cloud_storage_config(
+            pubsub::CloudStorageConfigBuilder{}.set_bucket(bucket)));
+    if (sub.status().code() == google::cloud::StatusCode::kAlreadyExists) {
+      std::cout << "The subscription already exists\n";
+      return;
+    }
+    if (!sub) throw std::move(sub).status();
```
devbww, line 420: Nit: While `StatusOr<T>::status()` doesn't document having a `!ok()` pre-condition (nor does the implementation require it), I always suggest assuming that it does.  That is, check that something actually failed before checking the reason it failed.  It reads a little nicer, I think.

alevenberg: Added a check for the status, and then inside checked the error code.

google/cloud/pubsub/samples/samples.cc
```diff
@@ -400,6 +400,34 @@ void CreateBigQuerySubscription(
...
+  namespace pubsub = ::google::cloud::pubsub;
+  [](pubsub::SubscriptionAdminClient client, std::string const& project_id,
+     std::string const& topic_id, std::string const& subscription_id,
+     std::string const& bucket) {
+    auto sub = client.CreateSubscription(
+        pubsub::Topic(project_id, std::move(topic_id)),
+        pubsub::Subscription(project_id, std::move(subscription_id)),
+        pubsub::SubscriptionBuilder{}.set_cloud_storage_config(
+            pubsub::CloudStorageConfigBuilder{}.set_bucket(bucket)));
+    if (!sub.ok()) {
```
devbww, line 416: Looks good.  Thanks.  Although we would still typically write just `if (!sub)`.

devbww (approved):

google/cloud/pubsub/samples/samples.cc
alevenberg: Done

PR #12097, opened 2023-07-14
cleanup(otel): tests use scoped span catcher

dbolduc:
Use a scoped span catcher object, that restores the `NoopTracerProvider` after a test runs.

If we forget to `InstallSpanCatcher()` in a test that creates spans, we can see a different valid test fail, because it picks up the extra spans. I don't quite understand why that behavior happens, but I think this PR will eliminate it.

Keep the interface the same so I don't have to update N tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12097)

### DIFF
--- a/google/cloud/testing_util/opentelemetry_matchers.cc
+++ b/google/cloud/testing_util/opentelemetry_matchers.cc
@@ -144,11 +144,11 @@ bool ThereIsAnActiveSpan() {
   return opentelemetry::trace::Tracer::GetCurrentSpan()->GetContext().IsValid();
 }
 
-std::shared_ptr<opentelemetry::exporter::memory::InMemorySpanData>
-InstallSpanCatcher() {
+SpanCatcher::SpanCatcher()
+    : previous_(opentelemetry::trace::Provider::GetTracerProvider()) {
   auto exporter =
       std::make_unique<opentelemetry::exporter::memory::InMemorySpanExporter>();
-  auto span_data = exporter->GetData();
+  span_data_ = exporter->GetData();
   auto processor =
       std::make_unique<opentelemetry::sdk::trace::SimpleSpanProcessor>(
           std::move(exporter));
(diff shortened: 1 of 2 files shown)


google/cloud/testing_util/opentelemetry_matchers.cc
```diff
@@ -144,19 +144,31 @@ bool ThereIsAnActiveSpan() {
...
       std::make_unique<opentelemetry::exporter::memory::InMemorySpanExporter>();
-  auto span_data = exporter->GetData();
+  span_data_ = exporter->GetData();
   auto processor =
       std::make_unique<opentelemetry::sdk::trace::SimpleSpanProcessor>(
           std::move(exporter));
   std::shared_ptr<opentelemetry::trace::TracerProvider> provider =
       opentelemetry::sdk::trace::TracerProviderFactory::Create(
           std::move(processor));
   opentelemetry::trace::Provider::SetTracerProvider(provider);
```
devbww, line 158: `std::move(provider)`?

coryan (approved):

google/cloud/testing_util/opentelemetry_matchers.h
```diff
@@ -198,8 +212,7 @@ ::testing::Matcher<SpanDataPtr> SpanEventsAre(Args const&... matchers) {
  * 1. a new exporter is installed for each test
  * 2. the tests within a fixture do not execute in parallel
  */
-std::shared_ptr<opentelemetry::exporter::memory::InMemorySpanData>
-InstallSpanCatcher();
+std::shared_ptr<SpanCatcher> InstallSpanCatcher();
```
devbww, line 215: Do the comments on this function need to be revisited?

google/cloud/testing_util/opentelemetry_matchers.cc
dbolduc: Done.

google/cloud/testing_util/opentelemetry_matchers.h
dbolduc: Done.

devbww: There is also mention of `InMemorySpanData::GetSpans()`, which doesn't appear to apply any longer.

dbolduc: Argh, missed that. Thanks.

devbww (approved):

PR #12122, opened 2023-07-17
feat(spanner): tests and samples for foreign-key delete cascade

devbww:
As is often the case, the integration tests don't exercise anything that is particularly specific to the client implementation, but we adhere to the cross-language test specification none the less.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12122)

### DIFF
--- a/google/cloud/spanner/integration_tests/client_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/client_integration_test.cc
@@ -30,7 +30,9 @@ namespace {
 
 using ::google::cloud::testing_util::IsOk;
 using ::google::cloud::testing_util::StatusIs;
+using ::testing::AllOf;
 using ::testing::AnyOf;
+using ::testing::HasSubstr;
 using ::testing::NotNull;
 using ::testing::UnorderedElementsAre;
 using ::testing::UnorderedElementsAreArray;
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -3445,6 +3445,90 @@ void DeleteUsingDmlReturning(google::cloud::spanner::Client client) {
...
+          CONSTRAINT FKShoppingCartsCustomerId
+              FOREIGN KEY (CustomerId)
+              REFERENCES Customers (CustomerId)
+              ON DELETE CASCADE
+      ) PRIMARY KEY (CartId))""");
+  auto metadata =
+      client.UpdateDatabaseDdl(database.FullName(), std::move(statements))
+          .get();
+  google::cloud::spanner_testing::LogUpdateDatabaseDdl(  //! TODO(#4758)
+      client, database, metadata.status());              //! TODO(#4758)
```
dbolduc, line 3475: So I see that `//! TODO(#4758)` is used in other samples, too. I forget, are these lines excluded from the public samples?

google/cloud/spanner/integration_tests/client_integration_test.cc
```diff
@@ -1076,6 +1078,284 @@ TEST_F(PgClientIntegrationTest, FineGrainedAccessControl) {
...
+  auto delete_commit = client_->Commit(
+      Mutations{MakeDeleteMutation("Customers", KeySet().AddKey(MakeKey(1)))});
+  ASSERT_STATUS_OK(delete_commit);
+  auto carts =
+      client_->ExecuteQuery(SqlStatement("SELECT CartId FROM ShoppingCarts"));
+  for (auto& cart : StreamOf<std::tuple<std::int64_t>>(carts)) {
+    EXPECT_THAT(cart, IsOk());
+    if (!cart) break;
+    EXPECT_FALSE(true) << "Unexpected cart " << std::get<0>(*cart);
+  }
```
dbolduc, line 1157: comment: This took me a second to understand.

Something like:

```cc
EXPECT_THAT(StreamOf<std::tuple<std::int64_t>>(carts), IsEmpty());
```

... would set a clearer expectation, but the errors in the failure modes would be less useful. Up to you.

dbolduc (approved):

google/cloud/spanner/integration_tests/client_integration_test.cc
devbww: Unfortunately, it would be more complicated than that, as `StreamOf` returns a "range", which would then have to be used to produce a container that could be matched against `IsEmpty`.

For now then, I choose to leave it as-is, if only to align with the other, similar `StreamOf` expectations.  Perhaps we could later introduce a test helper that converts a stream to a `StatusOr<vector<RowType>>`.

Issue #11792, opened 2023-06-06
Parallelize doxygen generation for compute


scotthart:
Currently all .h files are processed serially, taking way too long. Create multiple doxygen targets for compute (with correct dependencies) in order to speed this up.

PR #12124, opened 2023-07-17
ci: include discovery services in feature list

dbolduc:
`checkers.sh` now knows about `google/cloud/compute`.

The one place where we use `list_full()` and `list_full_cmake()` differently is in our `publish-docs` build. We iterate over the `features::list_full()` when uploading:

https://github.com/googleapis/google-cloud-cpp/blob/de9cf7ece900ebe8f758091b2baaae8e3a932d8f/ci/cloudbuild/builds/publish-docs.sh#L24

https://github.com/googleapis/google-cloud-cpp/blob/de9cf7ece900ebe8f758091b2baaae8e3a932d8f/ci/cloudbuild/builds/publish-docs.sh#L148

https://github.com/googleapis/google-cloud-cpp/blob/de9cf7ece900ebe8f758091b2baaae8e3a932d8f/ci/cloudbuild/builds/publish-docs.sh#L166

This is also a necessary step towards documenting `compute`, but there is harder work to be done to unblock it. See #11792.

https://github.com/googleapis/google-cloud-cpp/blob/de9cf7ece900ebe8f758091b2baaae8e3a932d8f/google/cloud/compute/CMakeLists.txt#L132-L134

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12124)


coryan:
Do we need `ci/etc/full_feature_list`?  We can do something like `cmake -DCMAKE_MODULE_PATH=$PWD/cmake -P cmake/print-ga-libraries.cmake`...

google/cloud/compute/doc/main.dox
```diff
@@ -24,16 +24,106 @@ which should give you a taste of the Cloud Compute Engine API C++ client library
 
 <!-- inject-client-list-start -->
 This library offers multiple `*Client` classes, which are listed below. Each
-one of these classes exposes all the RPCs for a REST `resource` as member
-functions of the class. This library groups multiple REST services because they
+one of these classes exposes all the RPCs for a gRPC `service` as member
+functions of the class. This library groups multiple gRPC services because they
 are part of the same product or are often used together. A typical example may
```
dbolduc, line 29: Hm, maybe we don't want this change.

dbolduc: I propose that we merge this PR, then in a follow up, change the checkers script to `s/gRPC //` in these two places.

dbolduc (commented):
@scotthart - I will let you decide what to do with this PR. If you do not want it, I will just close the thing. We can always find it later.

scotthart (changes requested):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12124)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @dbolduc)

___
*[`google/cloud/compute/doc/environment-variables.dox` line 13 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12124#-NbHEctzCog_MO9XlPOF:-NbHEctzCog_MO9XlPOG:bf3a1gd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bdad028382726eb4aa49fad0a2e0c7011cb5a459/google/cloud/compute/doc/environment-variables.dox#L13)):*
> ```
> 
> 
> - `GOOGLE_CLOUD_CPP_ACCELERATOR_TYPES_ENDPOINT=...` overrides the
> ```

would we want this to to be `GOOGLE_CLOUD_CPP_COMPUTE_ACCELERATOR_TYPES_ENDPOINT` to properly scope the var name? 
Looking forward, I can't guarantee that compute won't add a resource name that conflicts with another existing service.
___
*[`google/cloud/compute/doc/main.dox` line 29 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12124#gh-1265980811:-NbHFEAwAluXtl_gBSp9:bb7ohn6) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/681391dcbf1268f2a5a7ddda1d2bdacd82d7ab68/google/cloud/compute/doc/main.dox#L29)):*

I propose that we merge this PR, then in a follow up, change the checkers script to `s/gRPC //` in these two places.
</blockquote></details>

sgtm

dbolduc (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12124)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @devbww and @scotthart)

___
*[`google/cloud/compute/doc/environment-variables.dox` line 13 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12124#-NbHEctzCog_MO9XlPOF:-NbKkVbJ6Q7cfCI40eIV:b-dw2p89) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/bdad028382726eb4aa49fad0a2e0c7011cb5a459/google/cloud/compute/doc/environment-variables.dox#L13)):*

would we want this to to be `GOOGLE_CLOUD_CPP_COMPUTE_ACCELERATOR_TYPES_ENDPOINT` to properly scope the var name? 
Looking forward, I can't guarantee that compute won't add a resource name that conflicts with another existing service.
</blockquote></details>

I think that is a good question, but not relevant to this PR. (Aside: it might have been wise to do this for the gRPC based services).

`checkers` is just grabbing the variable name from the code.

https://github.com/googleapis/google-cloud-cpp/blob/10f89ff813e56e1ca531d4515b3c705932e57bfa/google/cloud/compute/accelerator_types/v1/internal/accelerator_types_option_defaults.cc#L37

 So if we change the variable name the docs should update accordingly.

Also, I think @devbww made `checkers` smart enough to not repeat env var names within a library. So if we go with one `GOOGLE_CLOUD_CPP_COMPUTE_ENDPOINT` for all compute services, the docs should still look fine.

PTAL

scotthart (approved):
> Reviewed 1 of 8 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12124)* status: all files reviewed, 1 unresolved discussion (waiting on @dbolduc)

Issue #12112, opened 2023-07-17
Cloud SQL Admin API: `CreateEphemeral` sends malformed API request


kypp:
**Does this issue affect the google-cloud-cpp project?**
Yes

**What component of `google-cloud-cpp` is this related to?**
`google/cloud/sql`

**Describe the bug**
The `google::cloud::sql::SqlInstancesServiceClient::CreateEphemeral()` method doesn't work as expected - complaining about missing `public_key` when one is provided.

It seems that the request being made to the REST endpoint embeds the public key in `body` in JSON while the endpoint expects it to be a top-level field (see logs at the end). Apart from that, `ProtoRequestToJsonPayload` converts `public_key` to `publicKey` due to `preserve_proto_field_names` being `false` by default.

**To Reproduce**

```cpp
namespace sql = ::google::cloud::sql_v1;
auto client = sql::SqlInstancesServiceClient(
    google::cloud::ExperimentalTag{}, sql::MakeSqlInstancesServiceConnectionRest(google::cloud::ExperimentalTag{}));

google::cloud::sql::v1::SqlInstancesCreateEphemeralCertRequest create_req;
create_req.set_project("xxx");
create_req.set_instance("xxx");
create_req.mutable_body()->set_public_key("THE_PUBLIC_KEY");

auto eph_res = client.CreateEphemeral(create_req);

if (eph_res) {
  std::cout << eph_res->DebugString();
} else {
  std::cout << "Error: " << eph_res.status() << "\n";
}
```

**Expected behavior**
An ephemeral certificate is returned (as in [the api definition](https://cloud.google.com/sql/docs/postgres/admin-api/rest/v1/connect/generateEphemeralCert))

**Operating system:**
Ubuntu 22.04.2 LTS

**What compiler and version are you using?**
clang 16.0.5

**What version of `google-cloud-cpp` are you using?**
v2.13.0

**Additional context**
Selected curl logs from running with `GOOGLE_CLOUD_CPP_ENABLE_CLOG=DEBUG GOOGLE_CLOUD_CPP_ENABLE_TRACING=http`:

```log
(...)
>> curl(Send Header): POST /v1/projects/{project}/instances/{instance}/createEphemeral HTTP/2
Host: sqladmin.googleapis.com
[...]

PR #12145, opened 2023-07-18
fix(rest): preserve proto field snake case naming style in json

scotthart:
part of the work for for #12112

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12145)

### DIFF
--- a/google/cloud/internal/rest_stub_helpers.cc
+++ b/google/cloud/internal/rest_stub_helpers.cc
@@ -47,8 +47,10 @@ Status RestResponseToProto(google::protobuf::Message& destination,
 
 Status ProtoRequestToJsonPayload(google::protobuf::Message const& request,
                                  std::string& json_payload) {
-  auto proto_to_json_status =
-      google::protobuf::util::MessageToJsonString(request, &json_payload);
+  google::protobuf::util::JsonPrintOptions print_options;
+  print_options.preserve_proto_field_names = true;
+  auto proto_to_json_status = google::protobuf::util::MessageToJsonString(
+      request, &json_payload, print_options);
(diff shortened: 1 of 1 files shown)


dbolduc (approved):
can we verify this behavior in a unit test?

scotthart (commented):
We don't typically write unit tests for functions that we don't maintain and `google::protobuf::util::MessageToJsonString` is part of protobuf and should be tested there. Our existing integration tests should verify that this change does not break any existing behavior. Future improvements that leverage this change will be accompanied by tests that require this `JsonPrintOption` to behave as advertised.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12145#-:-N_ejHidEqnAE7mo_sjm:b-qxni6o)* status: 0 of 1 files reviewed, all discussions resolved

dbolduc:
I think having a test that we pass the correct options to the protobuf function is better than no test. Can't we verify the contents of `json_payload` after the call?

scotthart (commented):
This seems like a slippery slope. Do we really want to test that every function we use from our dependencies does what it says it does?

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12145#-:-N_enpZj0cjRyvNT6-dX:blq6ixl)* status: 0 of 1 files reviewed, all discussions resolved

dbolduc:
> This seems like a slippery slope. Do we really want to test that every function we use from our dependencies does what it says it does?

We would not be testing protobuf's function. We would be testing that our function calls their function with the right options.

If we removed the code from this PR, it would be nice if a unit test broke accordingly. :shrug:

PR #12158, opened 2023-07-19
feat(channel): add Reporting API

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12158)


CHANGELOG.md
```diff
@@ -110,6 +110,10 @@ info, see [#7463] and [#5976].
 
 ## v2.14.0 - TBD
 
+### [Channel Services](/google/cloud/channel/README.md)
+
+The library has been expanded to include the Reporting API.
```
devbww, line 115: I presume we have previously decided not to worry about quickstarts for these companion APIs, and I have just forgotten that?

devbww (approved):

CHANGELOG.md
dbolduc: Currently we do 1 quickstart per **library**, with rare exceptions. (pubsub now has two. storage has `quickstart.cc` and `quickstart_grpc.cc`).

I think in a team meeting we decided it was ideal to have one quickstart per **service**. But we have not taken any actions to realize that goal.

dbolduc: I think you are making a broader point about: how do we know the new API works?

Did I test the thing: No..... I will make sure that I can call an API before merging. And in future PR's I will try to indicate to the reviewer that I have performed some sort of smoke test.

devbww: There's a new sample file though.  That seems like it ticks the testing box.  And I'm certainly happy with one quickstart per library (but if we've decided otherwise, so be it).

dbolduc: > There's a new sample file though.

The generated samples do not make RPCs. So they tell us almost nothing.

dbolduc: Turns out we have never run the `channel` quickstart. See: #12163

So I am going to say this PR makes things no worse, and merge it.

PR #12162, opened 2023-07-19
impl(generator): handle paths with multiple variables

scotthart:
part of the work for #12112 

The generator has been correctly handling url paths specified in the google.api.http option that only had one variable, e.g.:
```
    option (google.api.http) = {
      get: "/v1/{name=projects/*/serviceAccounts/*}"
    };
```
but not handling url paths that contain multiple variables, e.g.:
```
    option (google.api.http) = {
      post: "/v1/projects/{project}/instances/{instance}/addServerCa"
    };
```
(Currently, all google.api.http paths are formatted as one of these two examples.)

This PR adds support for the second case which affects both metadata contents for gRPC, as well as correctly identifying ALL path fields for REST that need not be added as query parameters.

The PR consists of one commit for the generator changes and a second commit for all the regenerated service files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12162)


generator/internal/http_option_utils.cc
```diff
@@ -242,21 +252,37 @@ ParseHttpExtension(google::protobuf::MethodDescriptor const& method) {
                      << ": google::api::HttpRule not handled";
   }
 
-  static std::regex const kUrlPatternRegex(R"((.*)\{(.*)=(.*)\}(.*))");
-  std::smatch match;
-  if (!std::regex_match(url_pattern, match, kUrlPatternRegex)) {
+  static std::regex const kImplicitNamedFieldUrlPatternRegex(
+      R"((.*)\{(.*)\}(.*))");
```
coryan, line 256: Can you have something like: `"/v1/{parent=projects/*/instances/*}/databases/{bar}/baz"`?

Is there a full grammar for these expressions we are trying to match?

generator/internal/http_option_utils_test.cc
```diff
@@ -201,6 +208,23 @@ char const* const kServiceProto =
...
+    "       post: "
+    "\"/v1/projects/{project}/databases\"\n"
+    "       body: \"body\"\n"
+    "    };\n"
+    "  }\n"
+    "  // Leading comments about rpc Method7.\n"
+    "  rpc Method7(Implicit) returns (Empty) {\n"
+    "    option (google.api.http) = {\n"
+    "       post: "
+    "\"/v1/projects/{project}/instances/{instance}/databases\"\n"
```
coryan, line 223: Do we need a test using both the `{parent=...}` and `foo/{bar}/baz` syntax?

generator/internal/http_option_utils.cc
```diff
@@ -242,21 +252,37 @@ ParseHttpExtension(google::protobuf::MethodDescriptor const& method) {
...
+  std::smatch implicit_match;
+
+  // The implicit named field regex matches both the implicit and explicit
```
coryan, line 259: Note to self: this comment was useful, don't miss it next time.

generator/internal/http_option_utils.cc
```diff
@@ -242,21 +252,37 @@ ParseHttpExtension(google::protobuf::MethodDescriptor const& method) {
...
+    info.body = http_rule.body();
+    info.path_suffix = implicit_match[3];
+  }
 
   std::size_t current = 0;
-  for (auto open = url_pattern.find('{'); open != std::string::npos;
-       open = url_pattern.find('{', current)) {
+  auto open = url_pattern.find('{');
+  info.path_prefix = url_pattern.substr(current, open - current);
+  for (; open != std::string::npos; open = url_pattern.find('{', current)) {
```
coryan, line 285: I think there is a code smell here.  We are mixing regexp and explicit parsing.  That suggests we are doing something odd, like the grammar we want to parse is too complex for regex, but we don't want to write a full parser.

🤷

coryan (approved):
Assuming both implicit and explicit annotations can never appear in the same RPC, this LGTM. Otherwise maybe we should IM.

generator/internal/http_option_utils.cc
coryan: I found it:

https://github.com/googleapis/googleapis/blob/02aeb715005c633115bc77d563d6920731c98189/google/api/http.proto#L227-L234

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12162)* status: 0 of 360 files reviewed, 4 unresolved discussions (waiting on @coryan)

*[a discussion](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12162#gh-1268908342:-N_t-sOi6o_0hSaJiCqa:bb5okle) (no related file):*

I found it:

https://github.com/googleapis/googleapis/blob/02aeb715005c633115bc77d563d6920731c98189/google/api/http.proto#L227-L234

</blockquote></details>

Technically the grammar allows for the path you described. However, in practice that does not occur. I wrote a quick function to audit all the existing paths and none of them mix "Variable" formats (per their grammar). I also created an issue to add a check for unsupported grammars to this function and error if we encounter one https://github.com/googleapis/google-cloud-cpp/issues/12197

___
*[a discussion](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12162#gh-1268915962:-N_t0XNzEhIWPol9lLg_:b-vpxesn) (no related file):*

I think there is a code smell here.  We are mixing regexp and explicit parsing.  That suggests we are doing something odd, like the grammar we want to parse is too complex for regex, but we don't want to write a full parser.

🤷 
</blockquote></details>

Agreed. We could write a full parser, but there's no current usage of the grammar that exceeds our simple implementation of it. I will create an issue to consider adding a full parser in case it's something we want to tackle in the future https://github.com/googleapis/google-cloud-cpp/issues/12196

___

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12162)* status: 0 of 360 files reviewed, 4 unresolved discussions (waiting on @coryan)

*[a discussion](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12162#gh-1268910845:-N_t2JIo4hb1f0Wli8Uf:b-uyrlow) (no related file):*

Do we need a test using both the `{parent=...}` and `foo/{bar}/baz` syntax?
</blockquote></details>

It should be added as part of the solution for #12197 

___

coryan (commented):
:lgtm:

> Reviewed 1 of 360 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12162#-:-N_t2ZXZ0jOj864CdgyQ:bnfp4nl)* status: 1 of 360 files reviewed, all discussions resolved

PR #12175, opened 2023-07-20
cleanup(retail): update #12170 after #12172

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12175)


dbolduc (approved):
Thanks

PR #12189, opened 2023-07-20
ci: parallelize uploads in `publish-docs`

coryan:
Use the same ideas as `check-api`. Move the retry loop outside the function so we can see the output when a retry is needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12189)

### DIFF
--- a/ci/cloudbuild/builds/publish-docs.sh
+++ b/ci/cloudbuild/builds/publish-docs.sh
@@ -132,12 +132,18 @@ io::log_h2 "Publishing DocFX"
 io::log "branch:  ${BRANCH_NAME}"
 io::log "bucket:  gs://${docfx_bucket}"
 
-echo common "${FEATURE_LIST[@]}" | xargs -P "$(nproc)" -n 1 \
+# Upload the documents for all features, including common. We need to
+# rename `experimental-opentelemetry`. Some features do not have documentation,
+# such as `grafeas` and `experimental-storage-grpc`. These are harmless, as the
+# `stage_docfx()`` function skips missing directories without an error.
+uploaded=(common)
+uploaded+=("${FEATURE_LIST[@]/experimental-opentelemetry/opentelemetry}")
+echo "${uploaded[@]}" | xargs -P "$(nproc)" -n 1 \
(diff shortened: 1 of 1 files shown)


coryan:
Does not make a huge difference, but a few seconds saved are nothing to sneeze at.

Most of the time is spent in Doxygen, parsing the code.  All CPUs are busy, there is no more parallelism to squeeze here, the next improvement will come from sharding.

devbww (approved):

ci/cloudbuild/builds/publish-docs.sh
```diff
@@ -103,40 +103,49 @@ export LANG=C.UTF-8
...
 
+echo common "${FEATURE_LIST[@]}" | xargs -P "$(nproc)" -n 1 \
+  bash -c "TIMEFORMAT=\"\${0} completed in %0lR\";
+           time ci/retry-command.sh 3 120 stage_docfx \"\${0}\" \"${docfx_bucket}\" cmake-out \"cmake-out/\${0}.docfx.log\""
+
+errors=0
 for feature in common "${FEATURE_LIST[@]}"; do
-  if [[ "${feature}" == "experimental-storage-grpc" ]]; then continue; fi
-  if [[ "${feature}" == "grafeas" ]]; then continue; fi
-  if [[ "${feature}" == "experimental-opentelemetry" ]]; then feature="opentelemetry"; fi
```
dbolduc, line 131: Don't we still want these?

dbolduc (changes requested):

ci/cloudbuild/builds/publish-docs.sh
dbolduc: I think every time we loop over `FEATURE_LIST` we want to
- skip `experimental-storage-grpc`
- skip `grafeas`
- transform `experimental-opentelemetry` -> `opentelemetry`

coryan: I fixed `experimental-opentelemetry` -> `opentelemetry`. The other changes are not needed for the reasons explained in the new comments.

PR #12208, opened 2023-07-25
impl(bigquery): Applies SafeGetTo function to common resources

jsrinnn:
1) This PR applies the SafeGetTo() from [PR-12199](https://github.com/googleapis/google-cloud-cpp/pull/12199) to the resources defined in `common_v2_resources.h` that all apis use.

2) it also adds a clang-tidy warning suppression related to recursive fields that are used in the common resources.

FYI: I will be sending around 4 to 5 PRs after this one, that applies the SafeGetTo() 
to the different V2 apis i.e Job, Table, Dataset and Projects.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/bigquery/v2/minimal/internal/common_v2_resources.h"
+#include "google/cloud/bigquery/v2/minimal/internal/json_utils.h"
 #include "google/cloud/internal/debug_string.h"
 #include "google/cloud/internal/format_time_point.h"
 #include "google/cloud/log.h"
@@ -37,14 +38,14 @@ void to_json(nlohmann::json& j, QueryParameterStructType const& q) {
 }
 
 void from_json(nlohmann::json const& j, QueryParameterStructType& q) {
-  if (j.contains("name")) j.at("name").get_to(q.name);
+  SafeGetTo(q.name, j, "name");
   if (j.contains("type")) {
     if (q.type == nullptr) {
       q.type = std::make_shared<QueryParameterType>();
     }
-    j.at("type").get_to(*q.type);
+    SafeGetTo(*q.type, j, "type");
   }
-  if (j.contains("description")) j.at("description").get_to(q.description);
+  SafeGetTo(q.description, j, "description");
 }
 
 void to_json(nlohmann::json& j, QueryParameterType const& q) {
@@ -58,14 +59,14 @@ void to_json(nlohmann::json& j, QueryParameterType const& q) {
 }
 
 void from_json(nlohmann::json const& j, QueryParameterType& q) {
-  if (j.contains("type")) j.at("type").get_to(q.type);
+  SafeGetTo(q.type, j, "type");
   if (j.contains("arrayType")) {
     if (q.array_type == nullptr) {
       q.array_type = std::make_shared<QueryParameterType>();
     }
-    j.at("arrayType").get_to(*q.array_type);
+    SafeGetTo(*q.array_type, j, "arrayType");
   }
-  if (j.contains("structTypes")) j.at("structTypes").get_to(q.struct_types);
+  SafeGetTo(q.struct_types, j, "structTypes");
 }
 
 void to_json(nlohmann::json& j, QueryParameterValue const& q) {
@@ -75,9 +76,9 @@ void to_json(nlohmann::json& j, QueryParameterValue const& q) {
 }
 
 void from_json(nlohmann::json const& j, QueryParameterValue& q) {
-  if (j.contains("value")) j.at("value").get_to(q.value);
-  if (j.contains("arrayValues")) j.at("arrayValues").get_to(q.array_values);
-  if (j.contains("structValues")) j.at("structValues").get_to(q.struct_values);
+  SafeGetTo(q.value, j, "value");
+  SafeGetTo(q.array_values, j, "arrayValues");
+  SafeGetTo(q.struct_values, j, "structValues");
 }
 
 void to_json(nlohmann::json& j, StandardSqlField const& f) {
@@ -89,11 +90,11 @@ void to_json(nlohmann::json& j, StandardSqlField const& f) {
 }
 
 void from_json(nlohmann::json const& j, StandardSqlField& f) {
-  if (j.contains("name")) j.at("name").get_to(f.name);
+  SafeGetTo(f.name, j, "name");
   if (f.type == nullptr) {
     f.type = std::make_shared<StandardSqlDataType>();
   }
-  if (j.contains("type")) j.at("type").get_to(*f.type);
+  SafeGetTo(*f.type, j, "type");
 }
 
 void to_json(nlohmann::json& j, StandardSqlStructType const& t) {
@@ -101,7 +102,7 @@ void to_json(nlohmann::json& j, StandardSqlStructType const& t) {
 }
 
 void from_json(nlohmann::json const& j, StandardSqlStructType& t) {
-  if (j.contains("fields")) j.at("fields").get_to(t.fields);
+  SafeGetTo(t.fields, j, "fields");
 }
 
 void to_json(nlohmann::json& j, StandardSqlDataType const& t) {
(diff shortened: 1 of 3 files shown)


google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc
```diff
@@ -194,22 +195,36 @@ void from_json(nlohmann::json const& j, Value& v) {
       case 0:
         // Do not set any value
         break;
-      case 1:
-        v.value_kind = j.at("valueKind").get<double>();
+      case 1: {
+        double val;
+        SafeGetTo(val, j, "valueKind");
+        v.value_kind = val;
```
dbolduc, line 201: you might want a `SafeValueOr` here and below.

https://github.com/googleapis/google-cloud-cpp/pull/12209#discussion_r1273010101

google/cloud/bigquery/v2/minimal/internal/json_utils.h
```diff
@@ -42,12 +42,16 @@ void FromJson(std::chrono::hours& field, nlohmann::json const& j,
 void ToJson(std::chrono::hours const& field, nlohmann::json& j,
             char const* name);
 
+// Suppress recursive clang-tidy warnings
+//
+// NOLINTBEGIN(misc-no-recursion)
```
dbolduc, line 47: why are these necessary? `SafeGetTo` does not call itself.

jsrinnn (commented):
Addressed feedback. Thank You!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208#-:-NaL7jRJERcSabRpktdK:bh8mug1)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @dbolduc)

___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 47 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208#gh-1273021774:-NaKqIdw4eWchziVWC6q:bs7wumb) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/46c61de16f55390357d1870b55f3a9e39af82583/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L47)):*

If we call `j.contains(...)` and `SafeGetTo(..., j, ...)`, we are still doing two look ups.

Consider:

```cc
template <typename T>
SafeGetTo(std::shared_ptr<T> value, nlohmann::json const& j, std::string const& key) {
  auto i = j.find(key);
  if (i == j.end()) return;
  if (value == nullptr) value = std::make_shared<T>();
  i->get_to(*value);
}
```

(here and elsewhere)
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 201 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208#gh-1273040771:-NaL2L5M8heCj3moGXw-:b-3w4m81) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/46c61de16f55390357d1870b55f3a9e39af82583/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L201)):*

you might want a `SafeValueOr` here and below.

https://github.com/googleapis/google-cloud-cpp/pull/12209#discussion_r1273010101
</blockquote></details>

Based on the latest feedback on #12209 (comment), we only want the value to be set if `j.contains()` is true. 
The latest modifications in that PR addresses the above for all the setter member functions.

Since we don't have setters here, I have addressed the above a little differently. 
PTAL and let me know if you think of a better alternative.
___
*[`google/cloud/bigquery/v2/minimal/internal/json_utils.h` line 47 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208#gh-1273041388:-NaKvLvj0dBcZ5_n4IlM:bo34b9j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/46c61de16f55390357d1870b55f3a9e39af82583/google/cloud/bigquery/v2/minimal/internal/json_utils.h#L47)):*

why are these necessary? `SafeGetTo` does not call itself.
</blockquote></details>

This gives a clang-tidy error because of their call/usage in `common_v2_resource.cc` with fields that are self-referential.

google/cloud/bigquery/v2/minimal/internal/json_utils.h
```diff
@@ -51,6 +51,28 @@ void SafeGetTo(ResponseType& value, nlohmann::json const& j,
...
+  if (i == j.end()) return value;
+  if (value == nullptr) {
+    value = std::make_shared<T>();
+  }
+  i->get_to(*value);
+  return value;
+}
+
+template <typename T>
+bool SafeValueOr(T& value, nlohmann::json const& j, std::string const& key) {
```
dbolduc, line 68: We don't need `void SafeGetTo(...)` and `bool SafeValueOr(...)`. We should just modify `SafeGetTo` to return a `bool` like this function does.

https://github.com/googleapis/google-cloud-cpp/blob/916b808c9e918807dd3abd09049dc72e2e1bfd0c/google/cloud/bigquery/v2/minimal/internal/json_utils.h#L48-L53

google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc
```diff
@@ -139,9 +129,8 @@ void from_json(nlohmann::json const& j, StandardSqlDataType& t) {
     auto const index = j.at("sub_type_index").get<int>();
```
dbolduc, line 129: (Don't fix in this PR, just an FYI):

This is another double look up. `j.contains(...)` followed by `j.at(...)`.

google/cloud/bigquery/v2/minimal/internal/json_utils.h
```diff
@@ -51,6 +51,28 @@ void SafeGetTo(ResponseType& value, nlohmann::json const& j,
   auto i = j.find(key);
   if (i != j.end()) i->get_to(value);
 }
+
+template <typename T>
+std::shared_ptr<T> SafeGetTo(std::shared_ptr<T>& value, nlohmann::json const& j,
```
dbolduc, line 56: I think we should return a `bool` to keep the semantics the same.

jsrinnn (commented):
Addressed all feedback. Thank You!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208#-:-NaP5p56-FRcJsjRsk_6:b5jke68)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @dbolduc)

___
*[`google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc` line 129 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208#gh-1276313183:-NaP-hda7K0XSkOD_N44:b-k513va) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/916b808c9e918807dd3abd09049dc72e2e1bfd0c/google/cloud/bigquery/v2/minimal/internal/common_v2_resources.cc#L129)):*

(Don't fix in this PR, just an FYI):

This is another double look up. `j.contains(...)` followed by `j.at(...)`.


</blockquote></details>

Done.

Thanks for catching that. Fixed it here and elsewhere in the file. There should be no more `contains`.
___
*[`google/cloud/bigquery/v2/minimal/internal/json_utils.h` line 56 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208#gh-1276325312:-NaOy3ew9S09g4nppxb7:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/916b808c9e918807dd3abd09049dc72e2e1bfd0c/google/cloud/bigquery/v2/minimal/internal/json_utils.h#L56)):*

I think we should return a `bool` to keep the semantics the same.
</blockquote></details>

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/json_utils.h` line 68 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208#gh-1276301412:-NaOy3P02Wefp9sRXT9j:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/916b808c9e918807dd3abd09049dc72e2e1bfd0c/google/cloud/bigquery/v2/minimal/internal/json_utils.h#L68)):*

We don't need `void SafeGetTo(...)` and `bool SafeValueOr(...)`. We should just modify `SafeGetTo` to return a `bool` like this function does.

https://github.com/googleapis/google-cloud-cpp/blob/916b808c9e918807dd3abd09049dc72e2e1bfd0c/google/cloud/bigquery/v2/minimal/internal/json_utils.h#L48-L53
</blockquote></details>

Done.

jsrinnn:
Friendly Ping.

@dbolduc , can you PTAL and approve when you get a chance if there are no more changes needed? I think I have addressed all the feedback. Thank You!

dbolduc (approved):

jsrinnn:
Resolved conflicts and rebased from main

PR #12209, opened 2023-07-25
impl(bigquery): Applies SafeGetTo function Job response parsing

jsrinnn:
This PR applies the SafeGetTo() from [PR-12199](https://github.com/googleapis/google-cloud-cpp/pull/12199) to `Job` resource response parsing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12209)

### DIFF
--- a/google/cloud/bigquery/v2/minimal/internal/job_request.cc
+++ b/google/cloud/bigquery/v2/minimal/internal/job_request.cc
@@ -300,62 +300,83 @@ void to_json(nlohmann::json& j, QueryRequest const& q) {
 }
 
 void from_json(nlohmann::json const& j, PostQueryRequest& q) {
-  if (j.contains("projectId")) {
-    q.set_project_id(j.at("projectId").value("projectId", ""));
-  }
-  if (j.contains("queryRequest")) {
-    q.set_query_request(j.at("queryRequest").get<QueryRequest>());
-  }
+  std::string project_id;
+  SafeGetTo(project_id, j, "projectId");
+  q.set_project_id(project_id);
+
+  QueryRequest query_request;
+  SafeGetTo(query_request, j, "queryRequest");
+  q.set_query_request(query_request);
 }
 
 void from_json(nlohmann::json const& j, QueryRequest& q) {
-  if (j.contains("query")) q.set_query(j.at("query").get<std::string>());
-  if (j.contains("kind")) q.set_kind(j.at("kind").get<std::string>());
-  if (j.contains("parameterMode")) {
-    q.set_parameter_mode(j.at("parameterMode").get<std::string>());
-  }
-  if (j.contains("location")) {
-    q.set_location(j.at("location").get<std::string>());
-  }
-  if (j.contains("requestId")) {
-    q.set_request_id(j.at("requestId").get<std::string>());
-  }
-  if (j.contains("dryRun")) q.set_dry_run(j.at("dryRun").get<bool>());
-  if (j.contains("preserveNulls")) {
-    q.set_preserve_nulls(j.at("preserveNulls").get<bool>());
-  }
-  if (j.contains("useQueryCache")) {
-    q.set_use_query_cache(j.at("useQueryCache").get<bool>());
-  }
-  if (j.contains("useLegacySql")) {
-    q.set_use_legacy_sql(j.at("useLegacySql").get<bool>());
-  }
-  if (j.contains("createSession")) {
-    q.set_create_session(j.at("createSession").get<bool>());
-  }
-  if (j.contains("maxResults")) {
-    q.set_max_results(j.at("maxResults").get<std::uint32_t>());
-  }
-  if (j.contains("maximumBytesBilled")) {
-    q.set_maximum_bytes_billed(j.at("maximumBytesBilled").get<std::int64_t>());
-  }
-  if (j.contains("connectionProperties")) {
-    q.set_connection_properties(
-        j.at("connectionProperties").get<std::vector<ConnectionProperty>>());
-  }
-  if (j.contains("queryParameters")) {
-    q.set_query_parameters(
-        j.at("queryParameters").get<std::vector<QueryParameter>>());
-  }
-  if (j.contains("defaultDataset")) {
-    q.set_default_dataset(j.at("defaultDataset").get<DatasetReference>());
-  }
-  if (j.contains("formatOptions")) {
-    q.set_format_options(j.at("formatOptions").get<DataFormatOptions>());
-  }
-  if (j.contains("labels")) {
-    q.set_labels(j.at("labels").get<std::map<std::string, std::string>>());
-  }
+  std::string query;
+  SafeGetTo(query, j, "query");
+  q.set_query(query);
+
+  std::string kind;
+  SafeGetTo(kind, j, "kind");
+  q.set_kind(kind);
+
+  std::string parameter_mode;
+  SafeGetTo(parameter_mode, j, "parameterMode");
+  q.set_parameter_mode(parameter_mode);
+
+  std::string location;
+  SafeGetTo(location, j, "location");
+  q.set_location(location);
+
+  std::string request_id;
+  SafeGetTo(request_id, j, "requestId");
+  q.set_request_id(request_id);
+
+  bool dry_run;
+  SafeGetTo(dry_run, j, "dryRun");
+  q.set_dry_run(dry_run);
+
+  bool preserve_nulls;
+  SafeGetTo(preserve_nulls, j, "preserveNulls");
+  q.set_preserve_nulls(preserve_nulls);
+
+  bool use_query_cache;
+  SafeGetTo(use_query_cache, j, "useQueryCache");
+  q.set_use_query_cache(use_query_cache);
+
+  bool use_legacy_sql;
+  SafeGetTo(use_legacy_sql, j, "useLegacySql");
+  q.set_use_legacy_sql(use_legacy_sql);
+
+  bool create_session;
+  SafeGetTo(create_session, j, "createSession");
+  q.set_create_session(create_session);
+
+  std::int32_t max_results;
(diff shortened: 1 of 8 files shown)


dbolduc (approved):

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -300,62 +300,83 @@ void to_json(nlohmann::json& j, QueryRequest const& q) {
...
 void from_json(nlohmann::json const& j, PostQueryRequest& q) {
-  if (j.contains("projectId")) {
-    q.set_project_id(j.at("projectId").value("projectId", ""));
-  }
-  if (j.contains("queryRequest")) {
-    q.set_query_request(j.at("queryRequest").get<QueryRequest>());
-  }
+  std::string project_id;
+  SafeGetTo(project_id, j, "projectId");
+  q.set_project_id(project_id);
```
dbolduc, line 305: optionally, consider:

```cc
template <typename T>
T SafeValueOr(nlohmann::json const& j, std::string const& field, T default_value) {
  // left as an exercise...
}
```
```cc
q.set_project_id(SafeValueOr(j, "projectId", ""));
```

google/cloud/bigquery/v2/minimal/internal/job_request.cc
```diff
@@ -300,62 +300,83 @@ void to_json(nlohmann::json& j, QueryRequest const& q) {
...
+
+  std::string location;
+  SafeGetTo(location, j, "location");
+  q.set_location(location);
+
+  std::string request_id;
+  SafeGetTo(request_id, j, "requestId");
+  q.set_request_id(request_id);
+
+  bool dry_run;
```
dbolduc, line 333: Oops I am just noticing that you need to initialize these variables.

`SafeGetTo` does not always set their value.

dbolduc (changes requested):

google/cloud/bigquery/v2/minimal/internal/job_request.cc
coryan: I am not sure that is the "Right Thing":tm: Previously the code looked like this:

```cc
if (j.contains("dryRun")) q.set_dry_run(j.at("dryRun").get<bool>());
```

If we unroll the functions, now it looks more like this:

```cc
bool dry_run;
if (j.contains("dryRun")) dry_run = j.at("dryRun").get<bool>();
q.set_dry_run(dry_run);
```

I think we want something like this:

```cc
auto if_present = [&j](char const* key, auto const& apply) {
  if (auto i = j.at(key); i != j.end()) apply(*i); // TODO: map to C++14, this requires C++17
};

if_present("dryRun", [&q](auto const& j) { q.set_dry_run(j.get<bool>()); });
if_present("preserveNulls", [&q](auto const& j) { q.set_preserve_nulls(j.get<bool>()); }); 
... ...
```

https://godbolt.org/z/dEqqxrzh6

jsrinnn (commented):
Addressed feedback. Thank you!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12209#-:-NaKepqK08Pdp1K8YNSO:bh97a0x)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 305 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12209#gh-1273010101:-NaAfdyC6PTLrc4F0PcG:b-hk59sk) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/215c5133da1fa87a6cfe272600701fb468f9217e/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L305)):*

optionally, consider:

```cc
template <typename T>
T SafeValueOr(nlohmann::json const& j, std::string const& field, T default_value) {
  // left as an exercise...
}
```
```cc
q.set_project_id(SafeValueOr(j, "projectId", ""));
```
</blockquote></details>

Done

I have tried to address both yours and Carlos's feedback below so the signature of the template and the usage is different from what is mentioned above. 
___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 333 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12209#gh-1273011617:-NaJR-Cu0sYWgsW3yo6Q:bgiqyv9) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/215c5133da1fa87a6cfe272600701fb468f9217e/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L333)):*

I am not sure that is the "Right Thing":tm: Previously the code looked like this:

```cc
if (j.contains("dryRun")) q.set_dry_run(j.at("dryRun").get<bool>());
```

If we unroll the functions, now it looks more like this:

```cc
bool dry_run;
if (j.contains("dryRun")) dry_run = j.at("dryRun").get<bool>();
q.set_dry_run(dry_run);
```

That always calls `q.set_dry_run()`, sometimes with an uninitialized value.  I think what @dbolduc suggests will always call `q.set_dry_run()`, always with an initialized value.  That is better, but I think we want to call `q.set_dry_run()` only if `j` contains `dryRun`, and then call it with the value in `j`.

Maybe something like this:

```cc
auto if_present = [&j](char const* key, auto const& apply) {
  if (auto i = j.at(key); i != j.end()) apply(*i); // TODO: map to C++14, this requires C++17
};

if_present("dryRun", [&q](auto const& j) { q.set_dry_run(j.get<bool>()); });
if_present("preserveNulls", [&q](auto const& j) { q.set_preserve_nulls(j.get<bool>()); }); 
... ...
```

https://godbolt.org/z/zfsK77rEG

</blockquote></details>

Done

I am not very familiar with the lambda construct above so I have used a template with a function parameter that achieves the same and in my opinion relatively simpler to read. This also addresses both concerns.

PTAL and please let me know if further changes are needed.

jsrinnn:
FYI: There is one additional commit after I sent it for review. In the latest commit I moved the `SafeValueOr` method to json_utils.h as it will be used by other files as well. 
See feedback: https://reviewable.io/reviews/googleapis/google-cloud-cpp/12208#gh-1273040771

coryan (approved):
> Reviewed 6 of 7 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12209)* status: all files reviewed, 2 unresolved discussions (waiting on @dbolduc)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 333 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12209#gh-1273011617:-NaMLG7B2V4TAAcvFmXj:b5jsbjr) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/215c5133da1fa87a6cfe272600701fb468f9217e/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L333)):*

Done

I am not very familiar with the lambda construct above so I have used a template with a function parameter that achieves the same and in my opinion relatively simpler to read. This also addresses both concerns.

PTAL and please let me know if further changes are needed.
</blockquote></details>

LGTM.  

I see why the lambda may not work, you need the type `T`.... 

Just FYI, pointer to member functions are ... a bit old fashioned, and brittle when you don't control the class you are taking a pointer to member from.  With C++20 (or Abseil >= "something too new for us") you could use `std::bind_front()` / `absl::bind_front()`.

google/cloud/bigquery/v2/minimal/internal/json_utils.h
```diff
@@ -49,6 +49,15 @@ void SafeGetTo(ResponseType& value, nlohmann::json const& j,
   if (i != j.end()) i->get_to(value);
 }
 
+template <typename C, typename T, typename R>
+void SafeValueOr(nlohmann::json const& j, std::string const& key,
```
dbolduc, line 53: nit: I don't think the name reflects what the function does. (It does not set a **value or** a default).

I think you can just call it `SafeGetTo`.

dbolduc (approved):

jsrinnn (commented):
Addressed remaining feedback and enabled automerge.

Thanks!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12209#-:-NaOshZq60XSlBN_zZpF:b-7kk7xw)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @coryan and @dbolduc)

___
*[`google/cloud/bigquery/v2/minimal/internal/job_request.cc` line 333 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12209#gh-1273011617:-NaOmhsR5ok3qP-NutiW:b8rstva) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/215c5133da1fa87a6cfe272600701fb468f9217e/google/cloud/bigquery/v2/minimal/internal/job_request.cc#L333)):*

LGTM.  

I see why the lambda may not work, you need the type `T`.... 

Just FYI, pointer to member functions are ... a bit old fashioned, and brittle when you don't control the class you are taking a pointer to member from.  With C++20 (or Abseil >= "something too new for us") you could use `std::bind_front()` / `absl::bind_front()`.


</blockquote></details>

Ah...Thanks!
___
*[`google/cloud/bigquery/v2/minimal/internal/json_utils.h` line 53 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12209#gh-1276293612:-NaOmtzT9-IcfHbRCaWm:bb6un1j) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/a34496d2535745f59e8f8a6c11eafd0f541f220d/google/cloud/bigquery/v2/minimal/internal/json_utils.h#L53)):*

nit: I don't think the name reflects what the function does. (It does not set a **value or** a default).

I think you can just call it `SafeGetTo`.
</blockquote></details>

Done

PR #12229, opened 2023-07-26
fix(otel): remove the codes dependency on `rpc::Code`

alevenberg:
- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12229)


dbolduc (approved):
Can you also drop this bazel dep:

https://github.com/googleapis/google-cloud-cpp/blob/6d8b9eb86853ca4a07377dd7dbd1083fc827fd8b/google/cloud/opentelemetry/BUILD.bazel#L29

Issue #11095, opened 2023-03-23
Add advanced field_number determination to Discovery Doc to protobuf


scotthart:
The generator assign protobuf fields numbers in alphabetical order for the first generation of protos, starting with field number = 1. For subsequent generations of protobuf files, this process needs to consider the existing name, type, and number from the previous version and reuse the existing number. At the same time, if a field's name or type has changed from the previous generation, the generator needs to raise an error to indicate a breaking change has been introduced in the discovery document. Lastly, field numbers of deleted fields need to be reserved to ensure they are not used again in the future. New fields should always be added with field number equal to the max(reserved, currently_in_use) + 1.

PR #12230, opened 2023-07-26
impl(generator): create descriptor pool and plumb through

scotthart:
part of the work for #11095 

ThisPR creates the descriptor pool for the current generation of compute protos and plumbs the pointer down to DiscoveryTypeVertex where it will be used later (in a follow up PR) to determine field numbers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12230)

### DIFF
--- a/generator/internal/discovery_type_vertex.h
+++ b/generator/internal/discovery_type_vertex.h
@@ -34,9 +34,11 @@ namespace generator_internal {
 // of edges are unidirectional, and both of the resulting graphs are acyclic.
 class DiscoveryTypeVertex {
  public:
-  DiscoveryTypeVertex();
+  // descriptor_pool should never be NULL, and the constructor will assert if it
+  // is.
   DiscoveryTypeVertex(std::string name, std::string package_name,
-                      nlohmann::json json);
+                      nlohmann::json json,
+                      google::protobuf::DescriptorPool const* descriptor_pool);
 
   std::string const& name() const { return name_; }
   std::string const& package_name() const { return package_name_; }
--- a/generator/internal/discovery_file.cc
+++ b/generator/internal/discovery_file.cc
@@ -51,10 +51,13 @@ std::string GeneratedProtoPreamble() {
 }  // namespace
 
 DiscoveryFile::DiscoveryFile(DiscoveryResource const* resource,
-                             std::string file_path, std::string package_name,
+                             std::string file_path,
+                             std::string relative_proto_path,
+                             std::string package_name,
                              std::vector<DiscoveryTypeVertex const*> types)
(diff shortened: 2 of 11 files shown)


coryan (approved):

generator/internal/discovery_type_vertex.h
```diff
@@ -34,9 +34,11 @@ namespace generator_internal {
 // of edges are unidirectional, and both of the resulting graphs are acyclic.
 class DiscoveryTypeVertex {
  public:
-  DiscoveryTypeVertex();
+  // descriptor_pool should never be NULL, and the constructor will assert if it
+  // is.
```
devbww, line 38: It doesn't look like the constructor does this.  Indeed, it ignores the `DescriptorPool` argument.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12230)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @coryan and @devbww)

___
*[`generator/internal/discovery_to_proto.cc` line 439 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12230#gh-1275234461:-NaI7ifXAKJ8DU2EgED4:b9umy7i) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/734295f61bde71d207bf32d07c51fb6e99d4b4ae/generator/internal/discovery_to_proto.cc#L439)):*

Is there a bootstrap problem if new proto files get added in the future?
</blockquote></details>

With discovery doc to proto generation, a new, non-internal service proto file will be added that defines its own proto subpackage along with service, rpcs, and messages, thus no advanced field numbering is required for this new service/resource.
For generated internal proto files, which all belong to the same proto package, they are all loaded into the pool indirectly because the service proto files import them. We search the pool by the fully qualified name of the type, so regardless of what current file a type exists in, it will be found, and in the case of internal proto files possibly written to a new internal proto file.
___
*[`generator/internal/discovery_type_vertex.h` line 38 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12230#gh-1275242112:-NaI99Y22xIricxEl5LO:bumv7qn) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/734295f61bde71d207bf32d07c51fb6e99d4b4ae/generator/internal/discovery_type_vertex.h#L38)):*

It doesn't look like the constructor does this.  Indeed, it ignores the `DescriptorPool` argument.
</blockquote></details>

It will be in the next PR. The compiler warnings would not allow me to add the member variable and assert without the new member variable being "used" somewhere in the code.

PR #12272, opened 2023-07-31
cleanup: always include `/bigobj` for MSVC

dbolduc:
Always compile with `/bigobj` for MSVC.

Adding this flag on a case-by-case basis is untenable. So we just add it to all targets.

---

Motivation: Some of our libraries will start to require this flag.

https://github.com/microsoft/vcpkg/pull/32779#issuecomment-1657697502 

---

Why we might not want to do this is discussed here:

https://stackoverflow.com/questions/15110580/penalty-of-the-msvs-compiler-flag-bigobj

I don't think we need to worry about the Visual C++ 2005 problem.

The `.obj` files being 2% larger comment matters. I still think it is worth reducing the headache on our end, to apply the `/bigobj` flag indiscriminately.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12272)


coryan (approved):

dbolduc:
> It may be a good idea to:
> 
> * set the `gha:full-build` label
> * re-run the GHA builds

Done.

Note to self: I wanted to apply the `gha:full-build` label before creating the PR.

[This comment](https://github.com/googleapis/google-cloud-cpp/blob/7b7361e1dbb9226bc69ca8b44979c7db9f7416ce/.github/workflows/test-runner.yml#L65-L67) helped me realize I needed a new commit to rerun the GHA builds.

Issue #12294, opened 2023-08-02
Support idempotency token for Storage client


coryan:
The GCS API supports (or soon will support) an idempotency token header.  The header allows GCS to detect if a request is a duplicate, and returns the same value for duplicate requests.

To implement this we need to:

- Create the unique id in the `RetryClient`
- The decorators below `RetryClient` need to accept a new "context" field on each function
- The gRPC and REST-based implementations need to consume this field and send it to the service as a header.

It is time to split the `RawClient` stack in two, like we do for all the other libraries.  The `*Connection` stack will be used for mocks and to implement the retry loop.  The typical function will look like so:

```cc
  virtual StatusOr<BucketMetadata> GetBucketMetadata(
      GetBucketMetadataRequest const& request) = 0;
```

The `*Stub` will implement logging, tracing, and actually sending the data to the service, the typical function will look like so:

```cc
  virtual StatusOr<BucketMetadata> GetBucketMetadata(
      rest_internal::RestContext& context,
      Options const& options,
      GetBucketMetadataRequest const& request) = 0;
```

Note how the options are passed explicitly and not tunneled via `CurrentOptions`. Also note that the "context" object is REST-based.  We can change both over time, as the classes in the `*Stub` hierarchy are not needed for mocking.

----

Java implemented this in:

https://github.com/googleapis/java-storage/pull/2027

Internally, the design doc is [go/gcs-client-idem-token](http://goto.google.com/gcs-client-idem-token)

coryan:
I also need to stop using the idempotency policy, as these RPCs are now always idempotent.

coryan:
This is blocked on service-side work.

coryan:
Still blocked on service-side work.

scotthart:
Still blocked on service-side work.

PR #12301, opened 2023-08-02
fix(common): handle expired policies in `*RetryLoop`

coryan:
The retry policy may be expired when it enters the retry loop. For example, time-based policies may suffer from poor scheduling luck.

Part of the work for #12294

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12301)

### DIFF
--- a/google/cloud/internal/rest_retry_loop.h
+++ b/google/cloud/internal/rest_retry_loop.h
@@ -87,9 +87,8 @@ auto RestRetryLoopImpl(RetryPolicy& retry_policy, BackoffPolicy& backoff_policy,
   // The last error cannot be retried, but it is not because the retry
   // policy is exhausted. We call these "permanent errors", and they
   // get a special message.
-  char const* prefix = !retry_policy.IsExhausted()
-                           ? "Permanent error in"
-                           : "Retry policy exhausted in";
+  auto const* prefix = retry_policy.IsExhausted() ? "Retry policy exhausted in"
+                                                  : "Permanent error in";
   return internal::RetryLoopError(prefix, location, last_status);
 }
 
--- a/google/cloud/internal/rest_retry_loop.h
+++ b/google/cloud/internal/rest_retry_loop.h
@@ -18,6 +18,7 @@
 #include "google/cloud/backoff_policy.h"
 #include "google/cloud/idempotency.h"
 #include "google/cloud/internal/invoke_result.h"
+#include "google/cloud/internal/make_status.h"
 #include "google/cloud/internal/opentelemetry.h"
 #include "google/cloud/internal/rest_request.h"
 #include "google/cloud/internal/retry_loop_helpers.h"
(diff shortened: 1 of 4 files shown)


google/cloud/internal/rest_retry_loop.h
```diff
@@ -61,41 +62,35 @@ template <
     typename std::enable_if<google::cloud::internal::is_invocable<
                                 Functor, RestContext&, Request const&>::value,
                             int>::type = 0>
-auto RestRetryLoopImpl(std::unique_ptr<RetryPolicy> retry_policy,
-                       std::unique_ptr<BackoffPolicy> backoff_policy,
+auto RestRetryLoopImpl(RetryPolicy& retry_policy, BackoffPolicy& backoff_policy,
```
dbolduc, line 65: If we are going to make this change, we should do it in `RetryLoopImpl` too.

google/cloud/internal/rest_retry_loop.h
```diff
@@ -61,41 +62,35 @@ template <
...
+    sleeper(backoff_policy.OnCompletion());
   }
-  return internal::RetryLoopError("Retry policy exhausted in", location,
-                                  last_status);
+  // The last error cannot be retried, but it is not because the retry
+  // policy is exhausted. We call these "permanent errors", and they
+  // get a special message.
+  char const* prefix = !retry_policy.IsExhausted()
+                           ? "Permanent error in"
+                           : "Retry policy exhausted in";
```
dbolduc, line 92:
```suggestion
  // Distinguish between a permanent error and too many transient errors.
  char const* prefix = retry_policy.IsExhausted()
                           ? "Retry policy exhausted in"
                           : "Permanent error in";
```

google/cloud/internal/rest_retry_loop.h
```diff
@@ -61,41 +62,35 @@ template <
...
+  return internal::RetryLoopError(prefix, location, last_status);
```
dbolduc, line 93: observation: in the (rare) exhausted before start case, the error message is a bit redundant:

```
Retry policy exhausted in <location>:  Retry policy exhausted before first request attempt
```

I am fine with this.

google/cloud/internal/rest_retry_loop.h
dbolduc: nit: I'd drop the negation.

dbolduc (approved):

google/cloud/internal/rest_retry_loop.h
devbww: Nit nit:  It looks like `RetryLoopError()` could well handle the " in " part itself.

google/cloud/internal/rest_retry_loop.h
coryan: Done.

google/cloud/internal/rest_retry_loop.h
coryan: Remove the negation(s).  I am going to leave the `in` for a future PR.

google/cloud/internal/rest_retry_loop.h
coryan: I think I will open a separate thread to ~shave this baby yak~ cleanup the error messages from retry loops.

PR #12306, opened 2023-08-03
fix(pubsub): increase backoff from 1.3 to 4.0

### LINKED ISSUE #12328: Default retry settings result in too-frequent retries for Resource Exhausted errors
With the given initial retry of 0.1 sec, multiplier of 1.3, it will take over 18 attempts before the retry interval is > 10s.

The backoff should be more aggressive for RPCs that could retry ResourceExhausted. In the C++ library, that is all RPCs: https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/pubsub/retry_policy.h#L29-L37

alevenberg:
Fixes #12328

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12306)

### DIFF
--- a/google/cloud/pubsub/internal/defaults.cc
+++ b/google/cloud/pubsub/internal/defaults.cc
@@ -59,7 +59,7 @@ Options DefaultCommonOptions(Options opts) {
   if (!opts.has<pubsub::BackoffPolicyOption>()) {
     opts.set<pubsub::BackoffPolicyOption>(
         pubsub::ExponentialBackoffPolicy(std::chrono::milliseconds(100),
-                                         std::chrono::seconds(60), 1.3)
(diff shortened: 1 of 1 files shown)


coryan (approved):

dbolduc (approved):
This is not a `feat`. Maybe it is a `fix`? Although I have no idea what we are fixing...

Maybe put the motivation for this change in the PR description?

alevenberg:
Discussed with Darren.

Closing since the motivating change is due to how the GAPIC libraries implement exponential backoff.

C++ does not use that implementation.

alevenberg:
Re-opening, the pub/sub team will create an issue motivating the change

PR #12309, opened 2023-08-03
ci: faster ABI dump for new libraries

dbolduc:
In #11719, we added a way to only generate ABI dumps for a subset of the libraries. But we still built all of the GA libraries in `google-cloud-cpp`, first.

This PR makes it so that we only build the libraries we need.

---

Reminder that `GOOGLE_CLOUD_CPP_CHECK_API` is used (for developer convenience) when we add a new library or expand an existing library.

Given that our operator cycles are spaced out by a month, and our releases happen monthly, we often have a cold cache for this local build.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12309)


devbww (approved):
OK.

PR #12312, opened 2023-08-04
feat(compute): generate interconnect_remote_locations code

scotthart:
The most recent update of the compute discovery document added a new resource, and in turn yielded a new proto. This was overlooked and not added to the generator_config.textproto at that time.

Depending on how often new resources get added, we may want to consider attempting to automate this, but it would not be trivial.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12312)


coryan (approved):
FWIW, we have the same problem for gRPC-based services.  We miss new services in new `.proto` files.

> Reviewed 28 of 28 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12312#-:-NayLRD2-2asP03UQaf9:b-5d9alv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

PR #12319, opened 2023-08-04
feat(spanner): enable leader aware routing by default

### LINKED ISSUE #11111: change spanner::RouteToLeaderOption to be on by default

devbww:
I've been told to include this text in the PR description:

"This update contains performance optimisations that will reduce the latency of read/write transactions that originate from a region other than the default leader region."

See #11112.  Fixes #11111.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12319)


google/cloud/spanner/internal/defaults.cc
```diff
@@ -104,22 +104,17 @@ Options DefaultOptions(Options opts) {
...
-      }
-    }
-    for (auto const* enable : {"Y", "y", "T", "t", "1", "on"}) {
-      if (*e == enable) {
-        // Route to leader for RW/PartitionedDml transactions.
-        opts.unset<spanner::RouteToLeaderOption>();
+    // The option defaults to on (unset), but the default can be changed with a
+    // suitably-negative value in `${GOOGLE_CLOUD_CPP_SPANNER_ROUTE_TO_LEADER}`.
+    if (auto e = internal::GetEnv("GOOGLE_CLOUD_CPP_SPANNER_ROUTE_TO_LEADER")) {
+      for (auto const* disable : {"N", "n", "F", "f", "0", "off"}) {
```
coryan, line 111: Is it me, or we have used this sequence of values elsewhere?  Should we have a function?  And why not `OFF`, since we have `N` and `n`?
Neither of those questions is blockin.

coryan (approved):

google/cloud/spanner/internal/defaults.cc
devbww: Yes, we have `ParseBoolean()` in google/cloud/tracing_options.cc, which we could probably try to common-ize now.

As for the values, I just stuck with the `ParseBoolean()` choices, so that refactoring was a possibility.

Issue #11190, opened 2023-04-05
Emit multiple protobuf files for common schema types


scotthart:
Currently, all schema defined types from the discovery doc are defined in one common.proto. As part of the third phase of implementation, we need to split these types into multiple protobuf files and inject the appropriate import statements where they are needed. While having each type in its own file would be one solution, it would be best to group schema that are always used in conjunction with each other into the same file in order to minimize the number of different protobuf files.

PR #12320, opened 2023-08-04
impl(generator): set dependencies between discovery types

scotthart:
part of the work for #11190 

Add dependency edges between the type vertices for later use.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12320)

### DIFF
--- a/generator/internal/discovery_to_proto.cc
+++ b/generator/internal/discovery_to_proto.cc
@@ -297,8 +297,6 @@ std::set<std::string> FindAllRefValues(nlohmann::json const& json) {
     fields = json["properties"];
   } else if (json.contains("additionalProperties") || json.contains("items")) {
     fields = json;
-  } else {
-    return ref_values;
   }
 
   for (auto const& f : fields) {
@@ -322,8 +320,8 @@ Status EstablishTypeDependencies(
     for (auto const& ref : ref_values) {
       auto ref_iter = types.find(ref);
       if (ref_iter == types.end()) {
-        return internal::InternalError(
-            "Unknown depended upon type",
+        return internal::InvalidArgumentError(
+            absl::StrCat("Unknown depended upon type: ", ref),
             GCP_ERROR_INFO()
                 .WithMetadata("dependent type", type.first)
                 .WithMetadata("depended upon type", ref));
--- a/generator/internal/discovery_to_proto.cc
+++ b/generator/internal/discovery_to_proto.cc
@@ -298,7 +298,7 @@ std::set<std::string> FindAllRefValues(nlohmann::json const& json) {
   } else if (json.contains("additionalProperties") || json.contains("items")) {
     fields = json;
   } else {
-    return {};
+    return ref_values;
   }
 
   for (auto const& f : fields) {
@@ -312,6 +312,28 @@ std::set<std::string> FindAllRefValues(nlohmann::json const& json) {
   }
 
   return ref_values;
+}
+
+Status EstablishTypeDependencies(
+    std::map<std::string, DiscoveryTypeVertex>& types) {
+  for (auto& type : types) {
+    auto const& json = type.second.json();
+    auto ref_values = FindAllRefValues(json);
+    for (auto const& ref : ref_values) {
+      auto ref_iter = types.find(ref);
+      if (ref_iter == types.end()) {
+        return internal::InternalError(
+            "Unknown depended upon type",
+            GCP_ERROR_INFO()
+                .WithMetadata("dependent type", type.first)
+                .WithMetadata("depended upon type", ref));
+      }
+      type.second.AddNeedsType(&ref_iter->second);
+      ref_iter->second.AddNeededByType(&type.second);
+    }
+  }
+
+  return {};
 }
 
 std::vector<DiscoveryFile> CreateFilesFromResources(
--- a/generator/internal/discovery_type_vertex.cc
+++ b/generator/internal/discovery_type_vertex.cc
@@ -88,12 +88,11 @@ bool DiscoveryTypeVertex::IsSynthesizedRequestType() const {
   return json_.value("synthesized_request", false);
 }
 
-void DiscoveryTypeVertex::AddNeedsTypeName(std::string type_name) {
-  needs_.insert(std::move(type_name));
+void DiscoveryTypeVertex::AddNeedsType(DiscoveryTypeVertex* type) {
+  needs_type_.insert(type);
 }
-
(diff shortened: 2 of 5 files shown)


generator/internal/discovery_to_proto.cc
```diff
@@ -298,7 +298,7 @@ std::set<std::string> FindAllRefValues(nlohmann::json const& json) {
   } else if (json.contains("additionalProperties") || json.contains("items")) {
     fields = json;
   } else {
-    return {};
+    return ref_values;
```
devbww, line 301: I would suggest dropping the `else` altogether, and allowing an empty `fields` to return an empty `ref_values` without a special case.

generator/internal/discovery_to_proto.cc
```diff
@@ -314,6 +314,28 @@ std::set<std::string> FindAllRefValues(nlohmann::json const& json) {
...
 
+Status EstablishTypeDependencies(
+    std::map<std::string, DiscoveryTypeVertex>& types) {
+  for (auto& type : types) {
+    auto const& json = type.second.json();
+    auto ref_values = FindAllRefValues(json);
+    for (auto const& ref : ref_values) {
+      auto ref_iter = types.find(ref);
+      if (ref_iter == types.end()) {
+        return internal::InternalError(
```
devbww, line 325: Is this really an internal error, or a problem with the input data?

generator/internal/discovery_to_proto.cc
```diff
@@ -314,6 +314,28 @@ std::set<std::string> FindAllRefValues(nlohmann::json const& json) {
...
+            "Unknown depended upon type",
+            GCP_ERROR_INFO()
+                .WithMetadata("dependent type", type.first)
+                .WithMetadata("depended upon type", ref));
```
devbww, line 329: I would suggest that the name of the unknown, depended-upon type is more than metadata, and should be included in the message as well (instead?).

generator/internal/discovery_type_vertex.cc
```diff
@@ -465,9 +464,13 @@ StatusOr<std::string> DiscoveryTypeVertex::JsonToProtobufMessage(
...
-  return absl::StrFormat("name: %s; needs: %s; needed_by: %s", name_,
-                         absl::StrJoin(needs_, ","),
-                         absl::StrJoin(needed_by_, ","));
+  auto formatter = [](std::string* out, DiscoveryTypeVertex* t) {
+    out->append(t->name());
+  };
+  return absl::StrFormat(
+      "name: %s; needs_type_name: %s; needed_by_type_name: %s", name_,
+      absl::StrJoin(needs_type_, ",", formatter),
+      absl::StrJoin(needed_by_type_, ",", formatter));
```
devbww, line 473: That looks like a situation better suited to `StrCat()` than `StrFormat()`.

generator/internal/discovery_type_vertex.h
```diff
@@ -44,16 +44,16 @@ class DiscoveryTypeVertex {
   std::string const& name() const { return name_; }
   std::string const& package_name() const { return package_name_; }
   nlohmann::json const& json() const { return json_; }
+  std::set<DiscoveryTypeVertex*> needs_type() { return needs_type_; }
+  std::set<DiscoveryTypeVertex*> needed_by_type() { return needed_by_type_; }
```
devbww, line 48: Given the look of the other accessors, it seems like these should probably return by reference too, and particularly so given that there is an ownership question about the set members.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12320)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @devbww)

___
*[`generator/internal/discovery_to_proto.cc` line 301 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12320#gh-1284830444:-NbFjC5LE7qm-TLM2H2K:b-6tmwt0) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/341f6ea50a70453d88cd5e5a345009b13519c45a/generator/internal/discovery_to_proto.cc#L301)):*

I would suggest dropping the `else` altogether, and allowing an empty `fields` to return an empty `ref_values` without a special case.
</blockquote></details>

`else` removed
___
*[`generator/internal/discovery_to_proto.cc` line 325 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12320#gh-1284836303:-NbFjD_d6tEtG4C3EwW6:bhrsi03) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/341f6ea50a70453d88cd5e5a345009b13519c45a/generator/internal/discovery_to_proto.cc#L325)):*

Is this really an internal error, or a problem with the input data?
</blockquote></details>

Changed error type
___
*[`generator/internal/discovery_to_proto.cc` line 329 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12320#gh-1284836930:-NbFjQJi6ahRe8dFLIEC:bsyccss) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/341f6ea50a70453d88cd5e5a345009b13519c45a/generator/internal/discovery_to_proto.cc#L329)):*

I would suggest that the name of the unknown, depended-upon type is more than metadata, and should be included in the message as well (instead?).
</blockquote></details>

error message enhanced
___
*[`generator/internal/discovery_type_vertex.h` line 48 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12320#gh-1284839717:-NbFjUOO-XHjVB6bveQX:b-6bogez) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/341f6ea50a70453d88cd5e5a345009b13519c45a/generator/internal/discovery_type_vertex.h#L48)):*

Given the look of the other accessors, it seems like these should probably return by reference too, and particularly so given that there is an ownership question about the set members.
</blockquote></details>

Changed to return a reference to the `std::set`
___
*[`generator/internal/discovery_type_vertex.cc` line 473 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12320#gh-1284837973:-NbFjiM86SEHl8EPoohB:b-7206pd) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/341f6ea50a70453d88cd5e5a345009b13519c45a/generator/internal/discovery_type_vertex.cc#L473)):*

That looks like a situation better suited to `StrCat()` than `StrFormat()`.
</blockquote></details>

changed to `StrCat`

devbww (approved):

PR #12327, opened 2023-08-07
impl(storage): populate idempotency token header

coryan:
With this change the `RetryClient` loops populate the idempotency token header. More PRs are required to propagate this value to the stubs.

Part of the work for #12294

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12327)

### DIFF
--- a/google/cloud/storage/testing/retry_tests.cc
+++ b/google/cloud/storage/testing/retry_tests.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/storage/retry_policy.h"
 #include "google/cloud/storage/testing/canonical_errors.h"
 #include "google/cloud/testing_util/status_matchers.h"
+#include <algorithm>
 #include <chrono>
 
 namespace google {
@@ -29,10 +30,13 @@ using ::google::cloud::storage::testing::canonical_errors::TransientError;
 using ::google::cloud::testing_util::StatusIs;
 using ::testing::Contains;
 using ::testing::HasSubstr;
+using ::testing::IsEmpty;
 using ::testing::Pair;
 using ::testing::ResultOf;
+using ::testing::Truly;
 
 auto constexpr kTooManyFailuresCount = 2;
+auto constexpr kIdempotencyTokenHeader = "x-goog-gcs-idempotency-token";
 
 Options RetryClientTestOptions() {
   return Options{}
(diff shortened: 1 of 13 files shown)


google/cloud/storage/testing/retry_tests.h
```diff
@@ -39,6 +43,35 @@ ::testing::Matcher<Status> StoppedOnTooManyTransients(char const* api_name);
...
+  template <typename Request>
+  Status operator()(rest_internal::RestContext& context, Options const&,
+                    Request const&) {
+    CaptureIdempotencyToken(*tokens_, context);
+    return status_;
+  }
+
+ private:
+  std::shared_ptr<std::vector<std::string>> tokens_ =
+      std::make_shared<std::vector<std::string>>();
```
devbww, line 71: The `shared_ptr`-ness doesn't seem to escape this class, so why not a value?

google/cloud/storage/testing/retry_tests.cc
```diff
@@ -81,6 +85,36 @@ ::testing::Matcher<Status> StoppedOnPermanentError(char const* api_name) {
...
+          [](std::vector<std::string> const& v) { return v; },
+          Truly([](std::vector<std::string> const& v) {
+            auto const s = v.empty() ? std::string{} : v.front();
+            return std::all_of(v.begin(), v.end(),
+                               [s](auto const& x) { return x == s; });
+          })));
+}
+
+void CaptureIdempotencyToken(std::vector<std::string>& tokens,
+                             rest_internal::RestContext const& context) {
```
devbww, line 108: Q.  Do we have (or inherit) any preference for inputs/outputs ordering?

google/cloud/storage/testing/retry_tests.h
coryan: Because (AFAICT) these are copied by `WillRepeatedly()` and `WillOnce()`.  We want them to share the captured tokens to examine them after the mocking / test took place.

google/cloud/storage/testing/retry_tests.cc
coryan: Not really.  I am happy to swap them if it makes the code more readable, or more consistent with something...

google/cloud/storage/testing/retry_tests.h
devbww: Hmm.  Then perhaps some form of "must be shared between copied instances" comment would be appropriate.

google/cloud/storage/testing/retry_tests.cc
devbww: Nah, I'm happy without such a convention.

devbww (approved):

google/cloud/storage/testing/retry_tests.h
coryan: Done.

PR #12333, opened 2023-08-07
impl(compute): remove obsolete internal compute protos

scotthart:
part of the work for #11190 

As the generator dynamically creates the internal proto files for compute, there exists the case where we end up with fewer proto files than the previous generation. These unnecessary proto files need to be removed before we generate C++ code from the current generation of proto files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12333)

### DIFF
--- a/ci/cloudbuild/builds/generate-libraries.sh
+++ b/ci/cloudbuild/builds/generate-libraries.sh
@@ -33,6 +33,8 @@ bazel run --action_env=GOOGLE_CLOUD_CPP_ENABLE_CLOG=yes \
   --config_file="${PWD}/generator/integration_tests/golden_config.textproto"
 
 if [ -z "${GENERATE_GOLDEN_ONLY}" ]; then
+  newer_tmp_file="$(mktemp)"
+
   io::log_h2 "Running the generator to emit protos from discovery docs"
   bazel run --action_env=GOOGLE_CLOUD_CPP_ENABLE_CLOG=yes \
     //generator:google-cloud-cpp-codegen -- \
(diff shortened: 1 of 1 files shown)


ci/cloudbuild/builds/generate-libraries.sh
```diff
@@ -44,6 +46,10 @@ if [ -z "${GENERATE_GOLDEN_ONLY}" ]; then
     --generate_discovery_protos=true \
     --config_file="${PROJECT_ROOT}/generator/generator_config.textproto"
 
+  io::log_h2 "Removing obsolete google/cloud/compute/v1/internal/*.proto"
+  find "${PROJECT_ROOT}/google/cloud/compute/v1/internal/" -name '*.proto' -a ! -newer "${newer_tmp_file}" -exec git rm {} \;
```
coryan, line 50: The common idiom in `git` is to `git rm` *everything* before generating the files, then generate the files and then `git add` what is generated.  I am guessing the idiom won't work because here we both parse and regenerate the `*.proto` files to preserve the field numbers?  If I guessed wrong, consider changing the code.

coryan (approved):

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12333)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`ci/cloudbuild/builds/generate-libraries.sh` line 50 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12333#gh-1286451510:-NbKa7Pl7e4c8VxigLii:b-6ro5kp) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/8e0332779002a5e05ebe99b02fe6f365a3445c9e/ci/cloudbuild/builds/generate-libraries.sh#L50)):*

The common idiom in `git` is to `git rm` *everything* before generating the files, then generate the files and then `git add` what is generated.  I am guessing the idiom won't work because here we both parse and regenerate the `*.proto` files to preserve the field numbers?  If I guessed wrong, consider changing the code.
</blockquote></details>

Correct. We snapshot the old ones at the beginning of program execution and overwrite them during execution.

PR #12342, opened 2023-08-08
impl(generator): split common proto and import resulting files

### LINKED ISSUE #11190: Emit multiple protobuf files for common schema types

scotthart:
fixes #11190 

This PR is split into multiple commits to separate out the generated files.

This PR adds the final changes to finish splitting the monolithic common.proto into a minimal set of common_file_xxx.proto that are subsequently imported only where they are needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12342)


generator/internal/discovery_to_proto.cc
```diff
@@ -358,65 +369,106 @@ void ApplyResourceLabelsToTypes(
...
+  for (auto& kv : types) {
+    if (!kv.second.IsSynthesizedRequestType()) {
+      auto* type = &kv.second;
+      std::string resource_key = absl::StrJoin(type->needed_by_resource(), "");
+      auto iter = common_files_by_resource.find(resource_key);
+      if (iter != common_files_by_resource.end()) {
+        iter->second.AddType(type);
+      } else {
+        auto relative_proto_path = absl::StrFormat(
+            "google/cloud/%s/%s/internal/common_file_%03d.proto",
```
devbww, line 424: Consider `s/_file_/_/`.

scotthart (commented):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12342)* status: 0 of 328 files reviewed, 3 unresolved discussions (waiting on @devbww)

___
*[`generator/internal/discovery_file.h` line 77 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12342#gh-1288656458:-NbR7NAa6cuf5mnl7Wjr:bss4a2b) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0e20f20609b499695a5b90dc7d6d883102aa88ef/generator/internal/discovery_file.h#L77)):*

Unused?
</blockquote></details>

Removed.
___
*[`generator/internal/discovery_to_proto.cc` line 91 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12342#gh-1288668202:-NbR7P-T1ZfeACJkWId1:b1af3fs) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0e20f20609b499695a5b90dc7d6d883102aa88ef/generator/internal/discovery_to_proto.cc#L91)):*

Should we be separating these with some OOB char?
</blockquote></details>

Added and refactored to a helper function to make sure we use the same character at all times.
___
*[`generator/internal/discovery_to_proto.cc` line 424 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12342#gh-1288682052:-NbR7Y_d5vH_z7wKR8ge:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/0e20f20609b499695a5b90dc7d6d883102aa88ef/generator/internal/discovery_to_proto.cc#L424)):*

Consider `s/_file_/_/`.
</blockquote></details>

Done.

devbww (approved):

PR #12346, opened 2023-08-09
impl(storage): `CurlClient` is a `GenericStub`

### LINKED ISSUE #12294: Support idempotency token for Storage client

coryan:
This is the last of the `RawClient` to `GenericStub` conversions, and fixes #12294.  There is a lot of cleanup work which I have separated to a new different bug.

### DIFF
--- a/google/cloud/storage/internal/curl/client.h
+++ b/google/cloud/storage/internal/curl/client.h
@@ -16,7 +16,7 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_STORAGE_INTERNAL_CURL_CLIENT_H
 
 #include "google/cloud/storage/internal/curl/handle_factory.h"
-#include "google/cloud/storage/internal/raw_client.h"
+#include "google/cloud/storage/internal/generic_stub.h"
 #include "google/cloud/storage/oauth2/credentials.h"
 #include "google/cloud/storage/version.h"
 #include "google/cloud/internal/random.h"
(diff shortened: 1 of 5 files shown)


google/cloud/storage/internal/curl/client.h
```diff
@@ -47,163 +47,215 @@ std::string HostHeader(Options const& options, char const* service);
...
  public:
-  static std::shared_ptr<CurlClient> Create(Options options) {
-    // Cannot use std::make_shared because the constructor is private.
-    return std::shared_ptr<CurlClient>(new CurlClient(std::move(options)));
-  }
-  static std::shared_ptr<CurlClient> Create(ClientOptions options) {
-    return Create(MakeOptions(std::move(options)));
-  }
+  // The constructor is private because the class must always be created
+  // as a shared_ptr<>.
```
devbww, line 53: Looks public to me.

devbww (approved):

PR #12347, opened 2023-08-09
feat(storage): all operations are idempotent

### LINKED ISSUE #12294: Support idempotency token for Storage client

coryan:
The introduction of the idempotency token header makes all operations idempotent.  Some applications may want to override this, though in most cases that may not be a great idea.

We can simplify the implementation of the idempotency policies, and the documentation for them too.

Fixes #12294 (for realsies this time).

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12347)


google/cloud/storage/client.h
```diff
@@ -228,8 +228,12 @@ struct ClientImplDetails;
...
  * The default policies are to continue retrying for up to 15 minutes, and to
  * use truncated (at 5 minutes) exponential backoff, doubling the maximum
- * backoff period between retries. Likewise, the idempotency policy is
- * configured to retry all operations.
+ * backoff period between retries.
+ *
+ * By default, all operations are treated as idempotent. The client library
+ * sends the `x-goog-gcs-idempotency-token` header with a unique value for
+ * each group of retried RPCs. The service will not execute a request with the
+ * same value for this header and send the previous response.
```
devbww, line 236: That isn't as clear as it could be, methinks.  How about something like, "If the server receives a request with a token that it has previously seen, it will simply resend the matching response."

devbww (approved):

google/cloud/storage/client.h
coryan: Thanks, s/server/service/ but otherwise stole your text.

Issue #12340, opened 2023-08-08
Cleanups after idempotency token


coryan:
- [x] Rename all the `*Client` classes derived from `GenericStub` to `*Stub`:
- [x] GrpcClient
- [x] RestClient
- [x] HybridClient
- [x] CurlClient
- [x] LoggingClient
- [x] Rename `RetryClient`  `DefaultConnection` or `ConnectionImpl` or whatever name we use in all the other libraries.
- [x] Rename `TracingClient` to `TracingConnection` or whatever name we use in all other libraries.
- [x] Rename `RawClient` to `StorageConnection`
- [x] Update ARCHITECTURE.md
- ~The `GenericStub` class (and derived classes) should not need the `options()` member function~. We need this to have different default options for gRPC vs. REST.
- ~It is not clear if we need `rest_internal::RestContext&` *and* `Optionst const&` parameters, the context also holds options.~ We will keep both to maintain a similar API to the generated `*Stub` classes (soon).

PR #12362, opened 2023-08-10
cleanup(storage): rename `internal::RetryClient`

coryan:
This class plays the same role as the `*ConnectionImpl` classes in other services, so name it accordingly.

Part of the work #12340

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12362)


coryan:
This is growing into a cleanup of the factory functions. Closing until that is resolved.

PR #12368, opened 2023-08-10
feat(common): retry loops preserve original message

coryan:
Some applications want to have access to the original message, even after a retry failure.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12368)


dbolduc (approved):
Ok.

This makes me wonder if we even want to manipulate the status message. We could leave the retry information as error info metadata only.

I am sure there are customers who `<< status.message()` instead of `<< status`, though.

Issue #12359, opened 2023-08-10
Explicit options in the `*Stub` layer


coryan:
We pass `Options` parameters from the `*Client` to the `*Connection` layer via a thread local variable[^1]. We should not use this after the `*Connection` layer.  All the `*Stub` decorators should receive an explicit `Options const&` parameter.

Roughly speaking we need to do this:

- Change the `MakeTracedSleeper()` to consume an explicit `Options const&` parameter.
- Add  `RetryLoop` overloads that consume an explicit `Options const&` parameter.
- Add  `RestRetryLoop` overloads that consume an explicit `Options const&` parameter.
- Add  `AsyncRetryLoop` overloads that consume an explicit `Options const&` parameter.
- Add  `AsyncRestRetryLoop` overloads that consume an explicit `Options const&` parameter.
- Add `LogWrapper` overloads that consume an explicit `Options const&` parameter.

These overloads could be based on the type of the functor. We can use `std::enable_if` and `std::is_invocable` to remove the variants that cannot use explicit parameters.

- Then we can change the generator to emit all the gRPC-based stubs with an extra `Options const&` parameter.
- We will need to change all the unit tests for golden files at the same time.

The REST-based stubs should not need changes because the `Options` are already provided in the `RestContext&` parameter. Though we need to think about the non-constness of this parameter, i.e., we do want it to remain non-const, but maybe we cannot trust its `options()` function to carry the original options all the way through.

[^1]: We do this because changing the API to `*Connection` would be breaking, and we don't want a massive breaking change like modifying all these APIs.

PR #12369, opened 2023-08-11
cleanup(common): `internal::CallContext` supports explicit options

coryan:
Part of the work for #12359

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12369)

### DIFF
--- a/google/cloud/internal/call_context.h
+++ b/google/cloud/internal/call_context.h
@@ -37,11 +37,16 @@ namespace internal {
  * enabled, it also contains the parent span that encompasses the client call.
  */
 struct CallContext {
+  explicit CallContext(Options o) : options(std::move(o)) {}
+  // TODO(#12359) - maybe this can be removed once the explicit options cleanup
+  //     is done.
+  CallContext() : CallContext(CurrentOptions()) {}
+
+  Options options;
 #ifdef GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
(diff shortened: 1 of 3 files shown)


google/cloud/internal/call_context.h
```diff
@@ -70,12 +77,10 @@ class ScopedCallContext {
...
 
  private:
+  OptionsSpan options_span_;
 #ifdef GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
-  using ScopedSpan = opentelemetry::trace::Scope;
-  ScopedSpan span_;
+  opentelemetry::trace::Scope otel_scope_;
 #endif  // GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
-  using ScopedOptions = OptionsSpan;
-  ScopedOptions options_;
```
dbolduc, line 78: I prefer the old names. I think the `using` directives helped disambiguate which "span" we are talking about.

dbolduc (approved):
At the moment, this does not feel like a cleanup. :shrug:

google/cloud/internal/call_context.h
coryan: The old names and aliases confused at least one person (me).

I wanted to remove the default constructor, but it is used in too many places to do in one pass and the new constructor unblocks cleaning up AsyncRetryLoop and AsyncRestRetryLoop.

PR #12370, opened 2023-08-11
cleanup(storage): split factory functions

coryan:
Move the factory functions for `*Stub` and `*Connection` objects to their own files. That makes it easier to understand what these do. They are more complex than our usual factory functions in other libraries because Storage has more implementations of the lowest-level `*Stub` object, and because some tests create more complex mocks than needed.

Note that I am calling the factory functions for `RawClient` the `*Connection` factory functions. I expect a future PR will rename the class to `StorageConnection` (with an alias for backwards compatibility).

Part of the work for #12340 

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12370)


google/cloud/storage/internal/connection_factory.h
```diff
@@ -0,0 +1,60 @@
...
+    std::shared_ptr<storage::internal::RawClient> connection);
+
+/**
+ * Creates a storage connection object.
+ *
+ * In most libraries we have no need for such a factory function. Storage has
+ * the GCS+gRPC plugin. This plugin needs to create its own stubs and then
+ * wrap them with all the decorators, including the `*Connection`. We also want
+ * to make the plugin compile-time and link-time optional, so we cannot simply
+ * initialize this stus in the usual `MakeStorageConnection(Options)` function.
```
dbolduc, line 46: s/stus/stub/

google/cloud/storage/client.h
```diff
@@ -3291,16 +3291,16 @@ class Client {
       " Please file a bug at https://github.com/googleapis/google-cloud-cpp"
       " if you have a use-case not covered by these.")
 #endif  // _MSC_VER
-  // NOLINTNEXTLINE(performance-unnecessary-value-param)
-  explicit Client(std::shared_ptr<internal::RawClient> client,
+  explicit Client(std::shared_ptr<internal::RawClient> const& client,
```
devbww, line 3294: `shared_ptr<T> const& param` smells funny.

google/cloud/storage/grpc_plugin.cc
```diff
@@ -29,13 +31,15 @@ google::cloud::storage::Client DefaultGrpcClient(Options opts) {
...
-      opts, std::make_unique<storage_internal::HybridStub>(opts));
+  auto stub = [&]() -> std::unique_ptr<storage_internal::GenericStub> {
+    if (config == "metadata" || config.empty()) {
+      opts = google::cloud::storage_internal::DefaultOptionsGrpc(opts);
+      return std::make_unique<storage_internal::GrpcStub>(opts);
+    }
+    return std::make_unique<storage_internal::HybridStub>(opts);
+  }();
+  return storage::internal::ClientImplDetails::CreateWithoutDecorations(
+      MakeStorageConnection(opts, std::move(stub)));
```
devbww, line 42: `std::move(opts)`?

google/cloud/storage/internal/connection_factory.h
devbww, line 46: "stus"?

google/cloud/storage/internal/generic_stub_factory.cc
```diff
@@ -0,0 +1,63 @@
...
+}
+
+}  // namespace
+
+std::unique_ptr<GenericStub> DecorateStub(Options const& opts,
+                                          std::unique_ptr<GenericStub> stub) {
+  return DecorateStub(RequiresLogging(opts), std::move(stub));
+}
+
+std::unique_ptr<GenericStub> MakeDefaultStorageStub(Options opts) {
```
dbolduc, line 50: This function is being fed `Options const&`: https://github.com/googleapis/google-cloud-cpp/blob/441c3b620b4a2c046cd006141f2968cbbcd7d806/google/cloud/storage/internal/connection_factory.cc#L44-L47

I think it would simplify the code a smidge if we accept `Options const&` here. WDYT?

```cc
std::unique_ptr<GenericStub> MakeDefaultStorageStub(Options const& opts) {
  auto stub = [&opts] -> std::unique_ptr<GenericStub> {
    if (opts.get<storage::internal::UseRestClientOption>()) {
      return std::make_unique<storage::internal::RestStub>(opts);
    }
    return std::make_unique<storage::internal::RestStub>(opts);    
  }();
  return DecorateStub(RequiresLogging(opts), std::move(stub));
}
```

google/cloud/storage/internal/connection_factory.cc
```diff
@@ -0,0 +1,52 @@
...
+
+namespace google {
+namespace cloud {
+namespace storage_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+std::shared_ptr<storage::internal::RawClient> DecorateConnection(
+    Options const& opts,
+    std::shared_ptr<storage::internal::RawClient> connection) {
+  auto stub = MakeGenericStubAdapter(std::move(connection));
```
dbolduc, line 30: `stub = ... connection` 🫨

dbolduc (commented):
I didn't get to the `class Client` changes, but it looks like @devbww did.

google/cloud/storage/client.h
coryan: It does, added a comment explaining why.

google/cloud/storage/grpc_plugin.cc
coryan: Done.

google/cloud/storage/internal/connection_factory.h
coryan: Fixed.

google/cloud/storage/internal/generic_stub_factory.cc
coryan: Not sure: (1) the root factory function for `*Stub` in other libraries consumes a `Options opts`. (2) the copy of the options has to happen somewhere.  As-is, it happens at the `MakeDefaultStorageStub()` call site. As proposed it would happen as part of the `std::make_unique<>` calls. (3) I can read the code in the PR and in your changes just fine.

I am leaving it as-is. easy to change if we think it was the wrong call.

google/cloud/storage/internal/connection_factory.cc
coryan: I think we want to get rid of that non-sense, but not in this PR.

devbww (approved):

PR #12376, opened 2023-08-14
feat(language): add v2 service

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12376)


google/cloud/language/CMakeLists.txt
```diff
@@ -112,7 +112,8 @@ target_include_directories(
 target_compile_options(google_cloud_cpp_language_mocks
                        INTERFACE ${GOOGLE_CLOUD_CPP_EXCEPTIONS_FLAG})
 
-if (BUILD_TESTING AND GOOGLE_CLOUD_CPP_ENABLE_CXX_EXCEPTIONS)
+include(CTest)
+if (BUILD_TESTING)
```
coryan, line 116: The quickstart has a `catch(Status const&)`.  Now I wonder if the `noex` build is compiling the quickstarts.

coryan: FWIW, I think we should leave the `AND GOOGLE_CLOUD_CPP_ENABLE_CXX_EXCEPTIONS` and fix the build or all the CMake files in a separate PR.

coryan (approved):
A small nit in the CMake file, otherwise LGTM.

google/cloud/language/CMakeLists.txt
devbww: I messed up the re-basing, so I'm going to merge this and address your comment in a follow-up PR.

PR #12378, opened 2023-08-14
Reduce TimerCancel gRPC test load to 1000 iterations

drfloob:
This is a gRPC test that relied on the high performance of an uncommon (synthetic) gRPC Alarm usage pattern. Per the test comments, this change will reduce the likelihood of seeing a regression in gRPC from 99% of runs, to 10% of runs. I think this should still be sufficient to identify problems. Running this test with the next generation of gRPC Alarm implementations, in 100 runs, this `TimerCancel` test never exceeded 10 seconds in the `asan-pr` docker environment (most commonly: 5 to 7s).

### DIFF
--- a/google/cloud/completion_queue_test.cc
+++ b/google/cloud/completion_queue_test.cc
@@ -87,7 +87,7 @@ TEST(CompletionQueueTest, TimerCancel) {
 
   using TimerFuture = future<StatusOr<std::chrono::system_clock::time_point>>;
   auto worker = [&](CompletionQueue cq) {
-    for (int i = 0; i != 10000; ++i) {
+    for (int i = 0; i != 1000; ++i) {
       std::vector<TimerFuture> timers;
       for (int j = 0; j != 10; ++j) {
         timers.push_back(cq.MakeRelativeTimer(std::chrono::microseconds(10)));


coryan:
SGTM. Let's see what the builds say.

google/cloud/completion_queue_test.cc
```diff
@@ -87,7 +87,7 @@ TEST(CompletionQueueTest, TimerCancel) {
 
   using TimerFuture = future<StatusOr<std::chrono::system_clock::time_point>>;
   auto worker = [&](CompletionQueue cq) {
-    for (int i = 0; i != 10000; ++i) {
+    for (int i = 0; i != 1000; ++i) {
```
coryan, line 90: I think we should change the comments on line 81 to reflect the new expected values.  Something like:

```cc
// The magic numbers in this test (number of iterations, thread, etc.) were originally tuned
// to repro #5141 in 99 out of 100 runs using the workstations available at the time.  Since
// then, the number of iterations was reduced. The higher number was making the test too
// long with a the new, more deterministic albeit slower implementation of gRPC alarms.
// The new implementation expires the alarm in a I/O thread vs. sometimes using the thread
// scheduling the alarm.  We expect a repro in 10 out of 100 runs.
```

drfloob: Done, PTAL.

coryan (approved):

PR #12394, opened 2023-08-16
impl: create string vector from most things

scotthart:
In the process of refactoring the generator's parsing of google.api.http paths, I find myself needing this functionality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12394)

### DIFF
--- a/google/cloud/internal/make_string_vector.h
+++ b/google/cloud/internal/make_string_vector.h
@@ -1,0 +1,51 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_MAKE_STRING_VECTOR_H
+#define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_INTERNAL_MAKE_STRING_VECTOR_H
+
+#include "google/cloud/version.h"
+#include "absl/strings/string_view.h"
+#include <string>
+#include <vector>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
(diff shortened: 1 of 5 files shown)


google/cloud/internal/make_string_vector.h
```diff
@@ -0,0 +1,51 @@
...
+  return std::to_string(s);
+}
+
+inline std::string Stringify(std::string s) { return s; }
+
+inline std::string Stringify(char const* s) { return std::string(s); }
+
+inline std::string Stringify(absl::string_view s) {
+  return std::string{s.data(), s.size()};
+}
```
devbww, line 39: Semi-serious question: Given that this is "make_string_vector.h", perhaps all of these helpers should be in an "internal_internal" namespace?

google/cloud/internal/make_string_vector.h
```diff
@@ -0,0 +1,51 @@
...
+#include <vector>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+template <typename T>
+std::string Stringify(T s) {
+  return std::to_string(s);
```
devbww, line 30: This mapping is why I preferred `ToString()` over `Stringify()`, but your choice.

google/cloud/internal/make_string_vector.h
```diff
@@ -0,0 +1,51 @@
...
+#include <string>
+#include <vector>
+
+namespace google {
+namespace cloud {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+namespace internal {
+
+template <typename T>
+std::string Stringify(T s) {
```
devbww, line 29: I'd choose `t` or `v` over `s` (the point being that the parameter is not `s`tring-like).

devbww (approved):

google/cloud/internal/make_string_vector.h
```diff
@@ -0,0 +1,51 @@
...
+
+inline std::string Stringify(char const* s) { return std::string(s); }
+
+inline std::string Stringify(absl::string_view s) {
+  return std::string{s.data(), s.size()};
+}
+
+template <typename... Ts>
+std::vector<std::string> MakeStringVector(Ts&&... ts) {
+  return {Stringify(std::forward<Ts>(ts))...};
```
coryan, line 43: This is a parameter pack expansion, which will expand to a `std::initializer_list<std::string>`. Initializer lists are passed by value:

https://en.cppreference.com/w/cpp/container/vector/vector

So this is copying all the strings.

Depending on the context that may not matter.  If this is going to be used once when creating a `*Client` then I cannot see how it makes any difference.  If it is used once per RPC it probably won't make any difference either, but it might.  You have more context than I do, I think you know what to do.

coryan (approved):

PR #12406, opened 2023-08-17
fix(common): gRPC alarms require more locking

coryan:
Turns out `grpc::Alarm` is not thread-safe. The documentation is about to say so, and a future implementation will trigger more faults in our code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12406)

### DIFF
--- a/google/cloud/internal/default_completion_queue_impl.cc
+++ b/google/cloud/internal/default_completion_queue_impl.cc
@@ -64,12 +64,25 @@ class AsyncTimerFuture : public AsyncGrpcOperation {
 
   void Set(grpc::CompletionQueue& cq,
            std::chrono::system_clock::time_point deadline, void* tag) {
+    std::unique_lock<std::mutex> lk(mu_);
+    if (state_ != kIdle) return;
     deadline_ = deadline;
+    state_ = kSet;
     alarm_.Set(&cq, deadline, tag);
   }
 
   void Cancel() override {
     ScopedCallContext scope(call_context_);
+    std::unique_lock<std::mutex> lk(mu_);
+    if (state_ == kCancelled) return;
+    if (state_ == kIdle) {
+      state_ = kCancelled;
+      lk.unlock();
+      // Release the lock before (potentially) calling application code.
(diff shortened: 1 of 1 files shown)


google/cloud/internal/default_completion_queue_impl.cc
```diff
@@ -64,12 +64,25 @@ class AsyncTimerFuture : public AsyncGrpcOperation {
...
     deadline_ = deadline;
+    state_ = kSet;
     alarm_.Set(&cq, deadline, tag);
   }
 
   void Cancel() override {
     ScopedCallContext scope(call_context_);
+    std::unique_lock<std::mutex> lk(mu_);
+    if (state_ == kCancelled) return;
+    if (state_ == kIdle) {
```
drfloob, line 78: Note that since gRPC's Alarm cancellation is idempotent, this could be simplified a bit:

```
    if (std::exchange(state_, kCancelled) == kIdle) {
      lk.unlock();
      // Release the lock before (potentially) calling application code.
      promise_.set_value(Cancelled());
      return;
    }
    alarm_.Cancel();
```

google/cloud/internal/default_completion_queue_impl.cc
```diff
@@ -64,12 +64,25 @@ class AsyncTimerFuture : public AsyncGrpcOperation {
 
   void Set(grpc::CompletionQueue& cq,
            std::chrono::system_clock::time_point deadline, void* tag) {
+    std::unique_lock<std::mutex> lk(mu_);
+    if (state_ != kIdle) return;
```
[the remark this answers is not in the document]
drfloob: My mistake, that will not work (as you pointed out), since you need to hold a lock while calling `Set` to avoid the `Cancel` race.

google/cloud/internal/default_completion_queue_impl.cc
drfloob: This may be incorrect as well, actually. Please disregard.

dbolduc (approved):

PR #12412, opened 2023-08-18
feat(config): generate library

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12412)


google/cloud/config/quickstart/quickstart.cc
```diff
@@ -0,0 +1,42 @@
...
+    std::cerr << "Usage: " << argv[0] << " project-id location-id\n";
+    return 1;
+  }
+
+  namespace config = ::google::cloud::config_v1;
+  auto client = config::ConfigClient(config::MakeConfigConnection());
+
+  std::string const project = argv[1];
+  std::string const location = argv[2];
+  for (auto r : client.ListDeployments("/projects/" + project + "/locations/" +
```
coryan, line 31: The leading slash is not right here.  Resource names start without a slash.  It may be time for us to create a `Location` resource name helper.

PR #12415, opened 2023-08-18
impl(generator): explicit options for policy helpers 

coryan:
Part of the work for #12359 and for #9168 

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12415)


generator/internal/connection_impl_generator.cc
```diff
@@ -119,39 +119,25 @@ class $connection_class_name$Impl
   // So, we have to fallback to `options_`.
   HeaderPrint(R"""(
  private:
-  std::unique_ptr<$product_namespace$::$retry_policy_name$> retry_policy() {
-    auto const& options = internal::CurrentOptions();
-    if (options.has<$product_namespace$::$retry_policy_name$Option>()) {
-      return options.get<$product_namespace$::$retry_policy_name$Option>()->clone();
-    }
-    return options_.get<$product_namespace$::$retry_policy_name$Option>()->clone();
+  static std::unique_ptr<$product_namespace$::$retry_policy_name$>
```
dbolduc, line 122: optional nit: do we prefer to make this a standalone function in the anonymous namespace?

generator/internal/connection_impl_generator.cc
```diff
@@ -313,21 +299,17 @@ std::string ConnectionImplGenerator::MethodDefinition(
     return R"""(
 StreamRange<$response_type$>
 $connection_class_name$Impl::$method_name$($request_type$ const& request) {
-  auto& stub = stub_;
-  auto retry = std::shared_ptr<$product_namespace$::$retry_policy_name$ const>(retry_policy());
-  auto backoff = std::shared_ptr<BackoffPolicy const>(backoff_policy());
-
-  auto factory = [stub]($request_type$ const& request) {
+  auto current = google::cloud::internal::SaveCurrentOptions();
```
dbolduc, line 302: optional nit: you can drop the `google::cloud::` if you want. here and below.

dbolduc (approved):
FYI, you did a lot of the remaining work for #9168

`pubsub` / `pubsublite` might remain. I remember them being weird.

generator/internal/connection_impl_generator.cc
```diff
@@ -119,39 +119,25 @@ class $connection_class_name$Impl
...
-  std::unique_ptr<$product_namespace$::$retry_policy_name$> retry_policy() {
-    auto const& options = internal::CurrentOptions();
-    if (options.has<$product_namespace$::$retry_policy_name$Option>()) {
-      return options.get<$product_namespace$::$retry_policy_name$Option>()->clone();
-    }
-    return options_.get<$product_namespace$::$retry_policy_name$Option>()->clone();
+  static std::unique_ptr<$product_namespace$::$retry_policy_name$>
+  retry_policy(Options const& options) {
+    return options.get<$product_namespace$::$retry_policy_name$Option>()->clone();
   }
```
devbww, line 125: I see two checks, one on the thread-local options and one on the member options, being replaced by a single check, just on the thread-local options.  But I don't think I see anything (at least here) that compensates for that reduction.

devbww: I guess that just echoes the "We may need to update some unit tests" statement in #9168.

generator/internal/connection_impl_generator.cc
coryan: Done.

generator/internal/connection_impl_generator.cc
coryan: Leaving as-is

generator/internal/connection_impl_generator.cc
coryan: > But I don't think I see anything (at least here) that compensates for that reduction.

Other than in unit tests, the thread-local options is a superset of `options_`. As you probably noted, I changed the unit tests when needed.

dbolduc (approved):

dbolduc:
Argh, I remember the `pubsublite` weirdness. Its implementation uses a raw connection from `pubsub`, which won't create an `OptionsSpan`: https://github.com/googleapis/google-cloud-cpp/blob/522d1322931d03c9290a1052186dae0425beadd4/google/cloud/pubsublite/internal/multipartition_publisher.cc#L65

coryan:
> Argh, I remember the `pubsublite` weirdness. Its implementation uses a raw connection from `pubsub`, which won't create an `OptionsSpan`:


That would be easy to fix. Except that the admin API for Pub/Sub Lite requires regional endpoints, and the `AuthorityOption` must have a matching value:


https://github.com/googleapis/google-cloud-cpp/blob/522d1322931d03c9290a1052186dae0425beadd4/google/cloud/pubsublite/integration_tests/publisher_integration_test.cc#L98-L100


This suggests that the `Publisher` is initializing the admin connection incorrectly (it uses its own options).  I find the Pub/Sub Lite API (why does one need the admin API to publish messages?) and the implementation (why magically initialize a connection instead of receiving this as an argument?) disgusting.

dbolduc (approved):

PR #12416, opened 2023-08-19
impl(common): add internal::RemoveOptions<T...>(Options)

devbww:
Add a helper function to remove all options in an `OptionList`.  This is intended as a companion to `internal::ExtractOption<T>(Options)`, for use in services where the `Options` passed to a client operation are split into two subsets: (1) those options passed explicitly to the connection layer and beyond, and (2) the remainder that are passed implicitly via thread-local state.

For example, in a `Client::Foo(..., Options opts)`, ...
```
  auto foo1_option = internal::ExtractOption<Foo1Option>(opts);
  auto foo2_option = internal::ExtractOption<Foo2Option>(opts);
  internal::OptionsSpan span(
      internal::RemoveOptions<ExplicitOptions>(std::move(opts)));
  return conn_->Foo(..., {foo1_option, foo2_option});
```
where `ExplicitOptions` contains all the option types in category (1).

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12416)

### DIFF
--- a/google/cloud/options.cc
+++ b/google/cloud/options.cc
@@ -25,6 +25,14 @@ namespace cloud {
 GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 namespace internal {
 
+Options RemoveOptionsImpl(std::set<std::type_index> const& remove,
+                          Options opts) {
+  for (auto ti : remove) {
+    opts.m_.erase(ti);
+  }
+  return opts;
+}
+
 void CheckExpectedOptionsImpl(std::set<std::type_index> const& expected,
                               Options const& opts, char const* const caller) {
   for (auto const& p : opts.m_) {
--- a/google/cloud/options.h
+++ b/google/cloud/options.h
@@ -32,6 +32,7 @@ GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
 class Options;
 namespace internal {
 Options MergeOptions(Options, Options);
+Options RemoveOptionsImpl(std::set<std::type_index> const&, Options);
 void CheckExpectedOptionsImpl(std::set<std::type_index> const&, Options const&,
                               char const*);
 // TODO(#8800) - Remove when bigtable::Table no longer uses bigtable::DataClient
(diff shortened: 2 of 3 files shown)


coryan:
> 
> For example, in a `Client::Foo(..., Options opts)`, ...
> 
> ```
>   auto foo1_option = internal::ExtractOption<Foo1Option>(opts);
>   auto foo2_option = internal::ExtractOption<Foo2Option>(opts);
>   internal::OptionsSpan span(
>       internal::RemoveOptions<ExplicitOptions>(std::move(opts)));
>   return conn_->Foo(..., {foo1_option, foo2_option});
> ```

I think this is fine, assuming `opts` has not been merged with the `*Client` options yet.  That is, the code would be:

```cc
... Client::Foo(..., Options opts) {
  auto foo1_option = internal::ExtractOption<Foo1Option>(opts);
  auto foo2_option = internal::ExtractOption<Foo2Option>(opts);
  opts = internal::RemoveOptions<ExplicitOptions>(std::move(opts));
  internal::OptionsSpan span(internal::MergeOptions(std::move(opts), options());
  return conn_->Foo(..., {foo1_option, foo2_option});
```

You probably just forgot about the merge.  No big deal.  Eventually I want to change the prototype from merge options to something like:

```cc
ImmutableOptions MergeOptions(Options preferred, ImmutableOptions fallback);
```

At that time any changes to the result of `MergeOptions` will not be possible (not without an expensive copy).

coryan (approved):

coryan:
> > I think this is fine, assuming opts has not been merged with the *Client options yet.
> 
> I did indeed misplace the merge in the example. It should instead read ...
> 
> ```
>   opts = internal::MergeOptions(std::move(opts), opts_);
>   // Update `opts` via extraction and removal.
>   internal::OptionsSpan span(std::move(opts));
> ```
> 
> so that the "foo options" could come from the client or the operation.
> 
> Given that, I'll hold off on this until the dust settles. Perhaps trying to remove the "explicit" options from the span isn't worth it. 

> But it does mean that I might have to copy an operation's explicit options from (at least) the connection options.

Currently you would be making one copy (in the `Merge()` call), and avoid the second copy via `ExtractOption<>`.  Once `ImmutableOptions` is in place, you would be making one copy (in a `get<>()` call), since `ExtractOption<>` would not be applicable anymore [^1].

That is, in the fullness of time, both approaches yield the same number of copies for these explicit options.

> 
> Maybe every non-scalar option type should be a `shared_ptr<T const>` instead?

Possibly. If we only access these via `get<>()` then that may not optimize anything.  It depends on how the option value is used after accessing it.

PR #12420, opened 2023-08-19
cleanup(bigtable): explicit options for policy helpers

coryan:
Part of the work for #12359 

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12420)

### DIFF
--- a/google/cloud/bigtable/internal/data_connection_impl.cc
+++ b/google/cloud/bigtable/internal/data_connection_impl.cc
@@ -36,23 +36,18 @@ inline std::string const& app_profile_id() {
   return internal::CurrentOptions().get<bigtable::AppProfileIdOption>();
 }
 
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/internal/data_connection_impl.cc
```diff
@@ -36,23 +36,18 @@ inline std::string const& app_profile_id() {
   return internal::CurrentOptions().get<bigtable::AppProfileIdOption>();
```
dbolduc, line 36: this too

dbolduc: please :)

dbolduc (approved):
Thanks.

archaeology nit: I would not say this is part of the work for #9168 when the policy getters were already simplified and removed from the ConnectionImpl class.

PR #12422, opened 2023-08-19
feat(pubsublite): caller provides admin connection

coryan:
The Pub/Sub Lite publisher needs an admin connection to query attributes of the topic at run-time. Initializing an admin connection is not trivial, and the code was doing it subtly wrong. It is better to receive the `*Connection` from the caller, so they can provide the right `Options` in its initialization.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12422)


ci/abi-dumps/google_cloud_cpp_pubsublite.expected.abi.dump.gz
dbolduc, line 1: I thought we only did ABI dumps for GA libraries:

https://github.com/googleapis/google-cloud-cpp/blob/be01dd16b2bc5facb9fdf7c6b439e51be2e90b04/ci/cloudbuild/builds/check-api.sh#L33

Did you update this manually?

google/cloud/pubsublite/internal/multipartition_publisher.cc
```diff
@@ -62,6 +62,7 @@ future<Status> MultipartitionPublisher::Start() {
 }
 
 future<StatusOr<std::uint32_t>> MultipartitionPublisher::GetNumPartitions() {
+  google::cloud::internal::OptionsSpan span(admin_connection_->options());
```
dbolduc, line 65: nit: (I am short on time, and not looking through the code) do we need to support any call options here? Like if I want to set the pubsub topic admin retry policy in a call?

ci/abi-dumps/google_cloud_cpp_pubsublite.expected.abi.dump.gz
coryan: Yes. The baseline was already in place.

google/cloud/pubsublite/internal/multipartition_publisher.cc
coryan: Cannot say that I care, at least not for this change which is just trying to unblock #12415.  But I looked, this runs in a background thread. There are no call options that are relevant.

ci/abi-dumps/google_cloud_cpp_pubsublite.expected.abi.dump.gz
coryan: I am removing it. As you pointed out it is not used.

dbolduc (approved):

Issue #12428, opened 2023-08-20
Factor out cmake for gRPC-based client libraries


dbolduc:
# Background

- Our libraries' `CMakeLists.txt` are generated once by the scaffold generator, then maintained by hand.
- The `CMakeLists.txt` are mostly the same, up to the library name and the quickstart runner.
- Making sweeping changes across the cmake build files can be cumbersome for those who are IDE-challenged (me).
  - Issues like #5782 become easier

# Audit Methodology

Look at `eye_test.sh` from this commit: https://github.com/dbolduc/google-cloud-cpp/commit/ef20659c443ba29fb4a94549d02942abd9556e41

# Audit Results

96 libraries are just good. I made a note for 20 of the libraries. They are grouped into categories below.

## Actually fine
- channel
- containeranalysis 
- dlp

In these libraries, the changes are outside the common library functions. These
are things such as DOXYGEN_EXTRA_INCLUDES, or disabling a library for a given
compiler.

## Handwritten
- bigquery (samples. Also, it is in a separate .cmake file)
- bigtable
- iam (samples)
- pubsub
- spanner
- storage

These are too complicated. We will leave these alone for now.

## No versioned clients
- dialogflow_cx
- dialogflow_es

Some service dirs contain "", referencing the forwarding headers. We skip these
from the doxygen example path by checking: `if (NOT dir STREQUAL "") ...`. This
would incorrectly skip a directory that has samples in Dialogflow.

We can update the logic to check if the samples directory exists, instead.

## Extra PROTO_LIST
- asset

This has been handled in the PoC with an optional variable. https://github.com/googleapis/google-cloud-cpp/commit/b1c76b3eca4fbf90845ed06ef08aff97952fad95

## Extra proto library dependencies
- binaryauthorization
- contentwarehouse

These have cross-library proto dependencies.

We can add a variable for additional protos and add it to:
- google_cloud_cpp_doxygen_targets(...)
- google_cloud_cpp_add_pkgconfig(...)

## Extra argument to external_googleapis_install_pc(...)
- optimization

This is unnecessary. We can remove it. I sent #12425

## Protos compiled in external/googleapis/CMakeLists.txt
- iam
- logging
- monitoring
- speech
- texttospeech
- trace

See #8022 for more details. I will ask Carlos what's up. I would say it is a bug
that we are building these extra proto libraries (the service ones). I do not
think customers will be using these protos without using the corresponding
client libraries.

[...]

PR #12429, opened 2023-08-20
refactor(cmake): add gRPC based library helper function

dbolduc:
Part of the work for #12428 

Factors out the cookie cutter gRPC based client library cmake code.

I am setting list variables in the parent scope and using them in the function. I am more comfortable doing this than using `cmake_parse_arguments(...)` but that is an alternative approach we could use.


cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -0,0 +1,197 @@
...
+# * library:      the short name of the library, e.g. `kms`.
+# * display_name: the display name of the library, e.g. "Cloud Key Management
+#   Service (KMS) API"
+# * GOOGLE_CLOUD_CPP_SERVICE_DIRS: a list of service directories within the
+#   library.
+#
+function (google_cloud_cpp_add_ga_grpc_library library display_name)
+    set(library_target "google_cloud_cpp_${library}")
+    set(mocks_target "google_cloud_cpp_${library}_mocks")
+    set(protos_target "google_cloud_cpp_${library}_protos")
```
coryan, line 28:
```suggestion
function (google_cloud_cpp_add_ga_grpc_library library display_name)
    cmake_parse_arguments(_opt "" "" "SERVICE_DIRS" ${ARGN})
    set(library_target "google_cloud_cpp_${library}")
    set(mocks_target "google_cloud_cpp_${library}_mocks")
    set(protos_target "google_cloud_cpp_${library}_protos")
```

cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -0,0 +1,197 @@
...
+    set(DOXYGEN_PROJECT_NAME "${display_name} C++ Client")
+    set(DOXYGEN_PROJECT_BRIEF "A C++ Client Library for the ${display_name}")
+    set(DOXYGEN_PROJECT_NUMBER "${PROJECT_VERSION}")
+    set(DOXYGEN_EXCLUDE_SYMBOLS "internal")
+    set(DOXYGEN_EXAMPLE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/quickstart")
+
+    unset(mocks_globs)
+    unset(source_globs)
+    foreach (dir IN LISTS GOOGLE_CLOUD_CPP_SERVICE_DIRS)
+        string(REPLACE "/" "_" ns "${dir}")
```
coryan, line 40:
```suggestion
    foreach (dir IN LISTS _opt_SERVICE_DIRS)
        string(REPLACE "/" "_" ns "${dir}")
```

cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -0,0 +1,197 @@
...
+
+# A function to add targets for GA libraries that use gRPC for transport.
+#
+# * library:      the short name of the library, e.g. `kms`.
+# * display_name: the display name of the library, e.g. "Cloud Key Management
+#   Service (KMS) API"
+# * GOOGLE_CLOUD_CPP_SERVICE_DIRS: a list of service directories within the
+#   library.
+#
+function (google_cloud_cpp_add_ga_grpc_library library display_name)
```
coryan, line 25: This may be a case where a macro works better than a function.  Macros do not define a new scope.

cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -0,0 +1,197 @@
...
+                COMPONENT google_cloud_cpp_development)
+    # With CMake-3.12 and higher we could avoid this separate command (and the
+    # duplication).
+    install(
+        TARGETS ${library_target} ${protos_target}
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+                COMPONENT google_cloud_cpp_development
+                NAMELINK_ONLY
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+                COMPONENT google_cloud_cpp_development)
```
coryan, line 155:
```suggestion
```

cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -0,0 +1,197 @@
...
+    # Install the libraries and headers in the locations determined by
+    # GNUInstallDirs
+    install(
+        TARGETS ${library_target} ${protos_target}
+        EXPORT ${library_target}-targets
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+                COMPONENT google_cloud_cpp_runtime
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+                COMPONENT google_cloud_cpp_runtime
+                NAMELINK_SKIP
```
coryan, line 144: Consider (though maybe it is better to post-pone the change to a future PR)

```suggestion
                NAMELINK_COMPONENT google_cloud_cpp_development
```

coryan (commented):
I think using list parameters is not that hard, see below.

cmake/GoogleCloudCppLibrary.cmake
dbolduc: Hm, so I tried this but `cmake_parse_arguments(...)` ignores the empty string as an argument. So it would not work for us.

```diff
diff --git a/cmake/GoogleCloudCppLibrary.cmake b/cmake/GoogleCloudCppLibrary.cmake
 function (google_cloud_cpp_add_ga_grpc_library library display_name)
+    cmake_parse_arguments(_opt "" "" "SERVICE_DIRS" ${ARGN})
     set(library_target "google_cloud_cpp_${library}")
     set(mocks_target "google_cloud_cpp_${library}_mocks")
     set(protos_target "google_cloud_cpp_${library}_protos")
@@ -36,7 +36,8 @@ function (google_cloud_cpp_add_ga_grpc_library library display_name)

     unset(mocks_globs)
     unset(source_globs)
-    foreach (dir IN LISTS GOOGLE_CLOUD_CPP_SERVICE_DIRS)
+    foreach (dir IN LISTS _opt_SERVICE_DIRS)
+        message(STATUS "DARREN: dir=${dir}")
         string(REPLACE "/" "_" ns "${dir}")
         list(APPEND source_globs "${dir}*.h" "${dir}*.cc" "${dir}internal/*")
         list(APPEND mocks_globs "${dir}mocks/*.h")
@@ -46,6 +47,7 @@ function (google_cloud_cpp_add_ga_grpc_library library display_name)
                  "${CMAKE_CURRENT_SOURCE_DIR}/${dir}samples")
         endif ()
     endforeach ()
+    message(FATAL_ERROR "breaking...")

diff --git a/google/cloud/kms/CMakeLists.txt b/google/cloud/kms/CMakeLists.txt
@@ -16,10 +16,9 @@

 include(GoogleCloudCppLibrary)

-set(GOOGLE_CLOUD_CPP_SERVICE_DIRS "" "inventory/v1/" "v1/")
-
-google_cloud_cpp_add_ga_grpc_library(kms
-                                     "Cloud Key Management Service (KMS) API")
+google_cloud_cpp_add_ga_grpc_library(
+    kms "Cloud Key Management Service (KMS) API" SERVICE_DIRS ""
+    "inventory/v1/" "v1/")

 if (BUILD_TESTING AND GOOGLE_CLOUD_CPP_ENABLE_CXX_EXCEPTIONS)
     add_executable(kms_quickstart "quickstart/quickstart.cc")
```

Output:
```
-- Found RE2 via CMake.
-- CURL linkage will be non-static
-- CURL linkage will be non-static
-- DARREN: dir=inventory/v1/
-- DARREN: dir=v1/
CMake Error at cmake/GoogleCloudCppLibrary.cmake:50 (message):
  breaking...
```

cmake/GoogleCloudCppLibrary.cmake
dbolduc: I don't think we need any variables defined in the inner scope to exist outside. So I think a `function` is appropriate.

cmake/GoogleCloudCppLibrary.cmake
dbolduc: Yeah, let's wait on this and the above (and any other cleanups) to keep this refactor a little more pure.

dbolduc:
PTAL

coryan (approved):

Issue #8022, opened 2022-01-19
Proto libraries in `external/googleapis/` are always compiled


coryan:
The proto libraries in `external/googleapis/` are always compiled (unless gRPC is completely disabled, but that is Okay).

Most of them are small libraries with only one proto file in them, and many of them (things like "api.proto") are used by almost all libraries.  Not particularly onerous, but there are a few (e.g. `google_cloud_cpp_cloud_texttospeech`, `google_cloud_cpp_monitoring`) that are larger.  We could move them to `google/cloud/${service}` and compile them only when needed.

We need to have a plan on how to deal with these larger libraries.

We should also consider merging all the smaller libraries into a single `.a` file.  "One library per proto" makes no sense for the CMake build + install model, creates friction for customers, and (if we solve the many other problems) is very inconvenient for DLLs exports.  We can be backwards compatible with a series of interface-only targets.

dbolduc:
A customer is asking for this.  See the discussion in: #10174

coryan:
Realistically we will not have time to work on this. Closing for now.

dbolduc:
We discussed this, and decided that building extra protos is in fact a bug. It is an unnecessary burden on customers who build with cmake and do not want to compile these extra protos.

PR #12431, opened 2023-08-21
cleanup(cmake): cleanups around library helper function

dbolduc:
Part of the work for #12428 

- Check whether the sample directory exists instead of skipping `""` outright. (This unblocks one of the dialogflows)
- Uses `NAMELINK_COMPONENT`, as suggested by @coryan in review comments for #12429 
- No need for `google_cloud_cpp_common` in the `.pc` file when `google_cloud_cpp_grpc_utils` depends on `google_cloud_cpp_common`

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12431)

### DIFF
--- a/cmake/GoogleCloudCppLibrary.cmake
+++ b/cmake/GoogleCloudCppLibrary.cmake
@@ -41,7 +41,7 @@ function (google_cloud_cpp_add_ga_grpc_library library display_name)
         list(APPEND source_globs "${dir}*.h" "${dir}*.cc" "${dir}internal/*")
         list(APPEND mocks_globs "${dir}mocks/*.h")
         list(APPEND DOXYGEN_EXCLUDE_SYMBOLS "${library}_${ns}internal")
-        if (NOT dir STREQUAL "")
+        if (IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/${dir}samples")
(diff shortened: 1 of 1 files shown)


cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -41,7 +41,7 @@ function (google_cloud_cpp_add_ga_grpc_library library display_name)
         list(APPEND source_globs "${dir}*.h" "${dir}*.cc" "${dir}internal/*")
         list(APPEND mocks_globs "${dir}mocks/*.h")
         list(APPEND DOXYGEN_EXCLUDE_SYMBOLS "${library}_${ns}internal")
-        if (NOT dir STREQUAL "")
+        if (IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/${dir}samples")
```
coryan, line 44: FWIW, `bigquery`, `iam` and `pubsublite` also have this directory.

coryan (approved):

cmake/GoogleCloudCppLibrary.cmake
dbolduc: Ack. Those are outside of the scope of #8022.

PR #12433, opened 2023-08-21
refactor(cmake): asset uses library helper

dbolduc:
Part of the work for #12428

`asset` is a weird one. We always compile that `orgpolicy` proto. Note that it is just messages. So it is not a huge deal, and I don't expect us to treat it any differently in light of #8022

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12433)

### DIFF
--- a/cmake/GoogleCloudCppLibrary.cmake
+++ b/cmake/GoogleCloudCppLibrary.cmake
@@ -19,15 +19,21 @@
 # * library:      the short name of the library, e.g. `kms`.
 # * display_name: the display name of the library, e.g. "Cloud Key Management
 #   Service (KMS) API"
+#
+# Additionally, we must set the following **variable** in the parent scope. We
+# cannot use a `cmake_parse_arguments()` keyword because it will skip the empty
+# string when provided in a list. We often need to use the empty string.
+#
 # * GOOGLE_CLOUD_CPP_SERVICE_DIRS: a list of service directories within the
 #   library.
 #
-# The following conditional arguments can be supplied to handle edge cases:
+# The following **keywords** can be optionally supplied to handle edge cases:
 #
-# * GOOGLE_CLOUD_CPP_ADDITIONAL_PROTO_LISTS: a list of proto files that may be
-#   used indirectly. `asset` sets this.
+# * ADDITIONAL_PROTO_LISTS: a list of proto files that may be used indirectly.
+#   `asset` sets this.
 #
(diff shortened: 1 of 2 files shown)


cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -22,6 +22,11 @@
 # * GOOGLE_CLOUD_CPP_SERVICE_DIRS: a list of service directories within the
 #   library.
 #
+# The following conditional arguments can be supplied to handle edge cases:
```
coryan, line 25: Can this be an argument via `cmake_parse_arguments()`?

coryan (changes requested):

cmake/GoogleCloudCppLibrary.cmake
dbolduc: Yes. Done.

This was a bit tricky to document. Hopefully it is clear.

Note that I am still setting `DOXYGEN_EXTRA_INCLUDES` in the parent scope. I do not intend to add that as an argument or document it in the helper function.

cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -19,10 +19,21 @@
 # * library:      the short name of the library, e.g. `kms`.
 # * display_name: the display name of the library, e.g. "Cloud Key Management
 #   Service (KMS) API"
+#
+# Additionally, we must set the following **variable** in the parent scope. We
```
coryan, line 23: For a separate PR, maybe we should have a `SERVICE_DIRS` list parameter and a boolean parameter that adds `""` to it, something like `WITH_LOCAL_SERVICE_DIR` or `WITH_FORWARDING_HEADERS`

coryan (approved):

cmake/GoogleCloudCppLibrary.cmake
dbolduc: Eh. I think that would add unnecessary complexity. What we have works and I am not trying to get cmake readability.

coryan: Passing values through global variables in CMake is bad for the same reason that passing values in global variables is bad in all languages.  This is not a good habit to get into.

dbolduc: Ack.

PR #12436, opened 2023-08-21
refactor(cmake): use GA gRPC library helper in `[ab].*`

dbolduc:
Part of the work for #12428 

These may seem scary to review. I am going to keep them to <20 libraries per PR.

I am working off the libraries from `good.txt` in the audit[^1]. These files are without nuance. You could run the commands in `eye-test.sh` to confirm this for yourself[^2]. https://github.com/dbolduc/google-cloud-cpp/commit/ef20659c443ba29fb4a94549d02942abd9556e41

Also, I used `vim`-fu to make these edits. So I have slightly more confidence that the library short name, display name, and service directories are correct than if I had made the edits by hand.

### Alternate PR strategy (if y'all don't trust me)

I could break down the common library helper into different parts. (I do not know exactly what the splits would be, but something like: (1) defining the library, (2) the mocks library (3) the install (4) the doxygen (5) the samples, etc.

Then we could refactor these parts one at a time across the libraries.

I think that would be easier to review visually because fewer different lines of code are changing at the same time.

The downside is that it would take more work for me to generate these PRs.

[^1]: Plus `apikeys`. That one was missing from the audit because it was still listed as a transition library at the time.

[^2]: Actually, the exact commands wouldn't work, because I used `kms` as the exemplar, which has already been refactored. You could use `tpu` instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12436)


google/cloud/alloydb/CMakeLists.txt
```diff
@@ -14,100 +14,11 @@
...
-              GTest::gtest)
-set_target_properties(google_cloud_cpp_alloydb_mocks
-                      PROPERTIES EXPORT_NAME google-cloud-cpp::alloydb_mocks)
-target_include_directories(
-    google_cloud_cpp_alloydb_mocks
-    INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
-              $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>
-              $<INSTALL_INTERFACE:include>)
-target_compile_options(google_cloud_cpp_alloydb_mocks
-                       INTERFACE ${GOOGLE_CLOUD_CPP_EXCEPTIONS_FLAG})
```
coryan, line 110: It may be easier to review all these PRs if you first moved the quickstart section to the bottom of each file.

coryan (approved):

google/cloud/alloydb/CMakeLists.txt
dbolduc: Discussed elsewhere, and the diff is gonna be ugly either way.

PR #12437, opened 2023-08-21
refactor(cmake): use GA gRPC library helper in `[cd].*`

dbolduc:
Part of the work for #12428 

I am working off the libraries from `good.txt` in the audit[^1]. These files are without nuance. You could run the commands in `eye-test.sh` to confirm this for yourself[^2]. https://github.com/dbolduc/google-cloud-cpp/commit/ef20659c443ba29fb4a94549d02942abd9556e41

Also, I used `vim`-fu to make these edits. So I have slightly more confidence that the library short name, display name, and service directories are correct than if I had made the edits by hand.

[^1]: Plus `apikeys`. That one was missing from the audit because it was still listed as a transition library at the time.

[^2]: Actually, the exact commands wouldn't work, because I used `kms` as the exemplar, which has already been refactored. You could use `tpu` instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12437)


dbolduc:
ping

devbww (approved):

Issue #12232, opened 2023-07-26
URL encode key value pairs in x-goog-request-params


scotthart:
Per https://google.aip.dev/client-libraries/4222

This mostly affects the metadata decorators.

---

@dbolduc 's understanding is that we encode the keys and values and leave the `&` and `=` joiners as they are.

We need to:

- [x] URL encode the keys for implicit and explicit routing at code generation time
- [x] URL encode the values for...
  - [x] explicit routing in GAPICs + legacy bigtable + `testing_util::ValidateMetadataFixture`
  - [x] explicit routing -> simple case (that doesn't need a `RoutingMatcher`)
  - [x] implicit routing in GAPICs + `testing_util::ValidateMetadataFixture` ?
  - [x] storage's client side streaming APIs
  - [x] handwritten pubsub client side streaming APIs
  - [x] handwritten minimal IAM stub
  - [x] ~maybe handwritten bigquery v2~, although no requests seem to require routing. :shrug: (none of the requests set the params)
  - [x] ~legacy spanner admin, if we care to~ (We don't care to keep spanner legacy admin classes up to date, since it is legacy and now generated)

dbolduc:
> Do we need to do anything for REST-based clients?

Yeah. Hopefully the work is identical to the gRPC GAPICs

alevenberg:
Looked for `x-goog-request-params` in the handwritten bigquery: `google/cloud/bigquery/v2/minimal`

And none of the 4 decorators (dataset, job, project, or table) set any headers on the context.

PR #12440, opened 2023-08-22
impl: url encode routing keys

dbolduc:
Part of the work for #12232 

I am not adding any tests. I am almost certain that these keys cannot contain characters that need to be escaped. Note that there are no corresponding changes to any generated code...

But it is simple enough to encode them and the encoding is done at code generation time.... So whatever.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12440)

### DIFF
--- a/generator/internal/http_option_utils.cc
+++ b/generator/internal/http_option_utils.cc
@@ -17,6 +17,7 @@
 #include "generator/internal/printer.h"
 #include "google/cloud/internal/absl_str_join_quiet.h"
 #include "google/cloud/internal/algorithm.h"
+#include "google/cloud/internal/url_encode.h"
 #include "google/cloud/log.h"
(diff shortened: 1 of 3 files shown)


coryan (approved):
I think I remember reading that while we *currently* don't have cases that need this encoding, in the future we might.

This seems like a unit testable thing, if nothing else as a regression test.

dbolduc:
> This seems like a unit testable thing, if nothing else as a regression test.

I was overcome by laziness. Sorry. Still I think I am just going to merge this thing because it seems likely to conflict with other incoming changes.

Most of our unit tests involve parsing protos. I doubt field names can contain characters that would be url escaped... So I doubt that the protos could be parsed.... But admittedly, I didn't try.

Issue #9131, opened 2022-06-01
Implement AsyncClient::InsertObject()


coryan:

hitesh-kumain:
Hi @coryan this seems to be a good initiation of my journey towards contribution to FOSS projects.
can you please assign this to me ??

coryan:
I do not believe this is a good bug to get started. It would require using a lot of ad-hoc tooling in the `google-cloud-cpp` project.  I will ping you in a separate bug that may be a better start.

coryan:
We will try to schedule this for 2023/H2

coryan:
This is now scheduled for 2023/H2

PR #12458, opened 2023-08-24
impl(GCS+gRPC): a class to represent write payloads

coryan:
We need a public API for object payloads. Applications may need to examine the contents of the upload in their mocks, to ensure their code is sending the right data.

We want the implementation of this type to be an `absl::Cord`, so we can pass the data to gRPC and Protobuf without copying the contents.

Part of the work for #9131

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12458)

### DIFF
--- a/google/cloud/storage/internal/async/write_payload_impl.cc
+++ b/google/cloud/storage/internal/async/write_payload_impl.cc
@@ -1,0 +1,59 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/storage/internal/async/write_payload_impl.h"
+#include <numeric>
+
+namespace google {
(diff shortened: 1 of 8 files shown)


google/cloud/storage/internal/async/write_payload_impl.cc
```diff
@@ -0,0 +1,59 @@
...
+  return WritePayloadImpl::Make(MakeCord(std::move(p)));
+}
+
+storage_experimental::WritePayload MakeWritePayload(
+    std::vector<std::string> p) {
+  std::vector<absl::Cord> cords(p.size());
+  std::transform(std::make_move_iterator(p.begin()),
+                 std::make_move_iterator(p.end()), cords.begin(),
+                 [](std::string v) { return MakeCord(std::move(v)); });
+  return MakeWritePayload(std::move(cords));
```
dbolduc, line 53: Hm, I do not see extra copies, but we are looping over `p` twice. If you care, consider:

```cc
storage_experimental::WritePayload MakeWritePayload(
    std::vector<std::string> p) {
  auto full = std::accumulate(std::make_move_iterator(p.begin()),
                              std::make_move_iterator(p.end()), absl::Cord(),
                              [](absl::Cord a, std::string b) {
                                a.Append(MakeCord(std::move(b)));
                                return a;
                              });
  return WritePayloadImpl::Make(std::move(full));
}
```

dbolduc (approved):
LGTM, but I am not the strongest with this stuff.

google/cloud/storage/internal/async/write_payload_impl.cc
coryan: Done, here and in the header file for `std::vector<T>`.  Also removed some functions that won't be needed and their tests, and rebased.

PR #12461, opened 2023-08-24
ci: verify old proto libraries exist

dbolduc:
Part of the work for #8022 

Let's get these tests in place first before we start pulling the targets out of `external/googleapis/CMakeLists.txt` and compiling them as part of the client library targets.

Preemptively `find_package(...)` for the libraries that these targets will go to (that aren't already loaded earlier in the CMakeLists.txt). Alternatively, we could just add them to the GA libraries list above, but I'd rather keep it with the backwards compatibility tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12461)


scotthart (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12461)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dbolduc)

PR #12466, opened 2023-08-25
cleanup: remove spanner dep from grpc_utils unit tests

dbolduc:
Motivated by #8022 

Use IAM protos instead of Spanner protos in `grpc_utils` unit test, so we can drop the dep on `spanner_protos`.

### DIFF
--- a/google/cloud/internal/debug_string_protobuf_test.cc
+++ b/google/cloud/internal/debug_string_protobuf_test.cc
@@ -15,10 +15,10 @@
 #include "google/cloud/internal/debug_string_protobuf.h"
 #include "google/cloud/testing_util/scoped_log.h"
 #include "google/cloud/tracing_options.h"
+#include <google/iam/v1/policy.pb.h>
 #include <google/protobuf/duration.pb.h>
 #include <google/protobuf/text_format.h>
 #include <google/protobuf/timestamp.pb.h>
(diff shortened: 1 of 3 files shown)


scotthart (approved):
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12466)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dbolduc)

google/cloud/internal/debug_string_protobuf_test.cc
```diff
@@ -15,10 +15,10 @@
 #include "google/cloud/internal/debug_string_protobuf.h"
 #include "google/cloud/testing_util/scoped_log.h"
 #include "google/cloud/tracing_options.h"
+#include <google/iam/v1/policy.pb.h>
```
coryan, line 18: Maybe using `Duration` and `Timestamp` would have been enough and we can break the dependency on IAM too?

dbolduc: Well, we are never getting rid of `MinimalIamCredentialsStub`.

(But that uses `google::iam::credentials::v1`... so maybe it was foolish of me to use something from `google::iam::v1` :shrug:)

(and we have other tests that use `google::iam::admin::v1`, but those could probably be changed to `bigtable` too).

PR #12482, opened 2023-08-28
DO NOT REVIEW

dbolduc:
I just want to see if the builds pass, and I am too lazy to run them by hand.

I would break this PR down into a few different pieces.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12482)


dbolduc:
FWIW the `publish-docs` build fails because the grafeas protos were being generated in `google/cloud/binaryauthorization/grafeas/v1/*.pb.h`, whereas they used to be generated in `google/cloud/grafeas/`.

I think we would need to generalize the `--cpp_out` directory used by protoc.

Q: So the proto headers from #8022 used to be generated in `${build_out}/external/googleapis/google/bigtable/v2` (for example). Now they are in `${build_out}/google/cloud/bigtable/google/bigtable/v2`.....

@coryan - Is that a problem?

coryan:
> FWIW the `publish-docs` build fails because the grafeas protos were being generated in `google/cloud/binaryauthorization/grafeas/v1/*.pb.h`, whereas they used to be generated in `google/cloud/grafeas/`.
> 

That seems odd.

> I think we would need to generalize the `--cpp_out` directory used by protoc.

Maybe.  Or maybe we should use the functions Protobuf provides to generate code (assuming they exist in all versions of Protobuf we use and support).

> 
> Q: So the proto headers from #8022 used to be generated in `${build_out}/external/googleapis/google/bigtable/v2` (for example). Now they are in `${build_out}/google/cloud/bigtable/google/bigtable/v2`.....
> 
> @coryan - Is that a problem?

As long as the `-I` flags are right, it should not matter where the `*.pb.h` files are generated.  Protobuf has some unfortunate rules about import directories for protos, but these are not affected by the output directory.

dbolduc:
> That seems odd.

Yeah, that was plain wrong.

> As long as the -I flags are right, it should not matter where the *.pb.h files are generated.

Ack, that's what I thought, but I wasn't completely sure. Thanks.

Issue #12483, opened 2023-08-28
Consider using a template for GAPIC's `config.cmake.in`


dbolduc:
A follow up to #12428  (our cmake yak has more hair)

These `config.cmake.in` files are basically all the same. The only difference is the name of the library, and some of them have extra dependencies. (e.g. `asset` must `find_dependency(osconfig)` and others).

#12482 has an example showing what this work would probably look like:

https://github.com/googleapis/google-cloud-cpp/blob/d5382a01e480c6dd58b31e670472381425c5dcb4/cmake/GoogleCloudCppLibrary.cmake#L96-L98

https://github.com/googleapis/google-cloud-cpp/blob/d5382a01e480c6dd58b31e670472381425c5dcb4/cmake/templates/config.cmake.in

---

The gain is one less file per library created by the scaffolding and manually maintained by hand. (not that these change often)

---

### Cross library dependencies (in GA libraries):
- `asset` -> `accesscontextmanager`, `osconfig`
- `binaryauthorization`, `containeranalysis` -> `grafeas`
- `contentwarehouse` -> `documentai`

### Non-standard `*-targets.cmake` file name.
- `pubsub`
- `spanner`

### REST-based (and weird, anyways):
- `storage`

Issue #12467, opened 2023-08-25
Consider using `cmake_parse_arguments()` in cmake library helper


dbolduc:
A follow up issue to #12428 

We pass the service directories as a variable, not as a keyword, because `cmake_parse_arguments()` cannot handle empty strings in a list.

But even that fails when we try to pass a list with one element - the empty string (which we need to do for `dialogflow_*` at the moment).

So if we add support for handling the list, `""`, we might as well use `cmake_parse_arguments()`.

We should also consider adding a keyword for `GOOGLE_CLOUD_CPP_EXTRA_DOXYGEN_INCLUDES`, instead of passing that one by variable.

PR #12487, opened 2023-08-29
Refactor config cmake in

dbolduc:
Part of the work for #12483. Also related to #12428 and #12467.

Add explicit support to the client library helper to handle cross library deps. This also cleans up the passing of `GOOGLE_CLOUD_CPP_DOXYGEN_EXTRA_INCLUDES` from the `${library}/CMakeLists.txt` as a global. (Of course we later pass it as a global to our doxygen helper functions, but :shushing_face: )

This also introduces a `config.cmake.in` template and lets us remove 100+ identical `${library}/config.cmake.in` files.

I plan to update the scaffolding in a follow up. And I might update handwritten libraries like `bigtable`, `pubsub`, `spanner` which have identical `config.cmake.in`'s to the template.


cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -57,6 +59,11 @@ function (google_cloud_cpp_add_ga_grpc_library library display_name)
     endif ()
     set(DOXYGEN_EXCLUDE_SYMBOLS "internal")
     set(DOXYGEN_EXAMPLE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/quickstart")
+    unset(GOOGLE_CLOUD_CPP_DOXYGEN_EXTRA_INCLUDES)
```
coryan, line 62: Starting with CMake 3.12 we can use `list(TRANSFORM ...)`:

https://cmake.org/cmake/help/latest/command/list.html#transform

```CMake
set(GOOGLE_CLOUD_CPP_DOXYGEN_EXTRA_INCLUDES "${opt_CROSS_LIB_DEPS}")
list(TRANSFORM
      GOOGLE_CLOUD_CPP_DOXYGEN_EXTRA_INCLUDES
      PREPEND "${PROJECT_BINARY_DIR}/google/cloud/")
```

cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -201,7 +208,16 @@ function (google_cloud_cpp_add_ga_grpc_library library display_name)
 
     # Create and install the CMake configuration files.
     include(CMakePackageConfigHelpers)
-    configure_file("config.cmake.in" "${library_target}-config.cmake" @ONLY)
+    set(GOOGLE_CLOUD_CPP_CONFIG_LIBRARY "${library_target}")
+    unset(find_dependencies)
```
coryan, line 212: as above.

dbolduc: Done, but I am skeptical as to whether this one was an improvement. I briefly explored `list(TRANSFORM ... REPLACE ...)`, but gave up.

PR #12489, opened 2023-08-29
ci: scaffolding stops generating config.cmake.in

dbolduc:
Part of the work for #12483

I have not thought through what to do for transition libraries. We don't really use these any more. So deferring the problem isn't horrible.

https://github.com/googleapis/google-cloud-cpp/blob/main/doc/contributor/howto-guide-declare-a-library-ga.md#googlecloudlibraryconfigcmakein

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12489)

### DIFF
--- a/generator/internal/scaffold_generator.cc
+++ b/generator/internal/scaffold_generator.cc
@@ -292,7 +292,6 @@ void GenerateScaffold(
     std::string name;
     Generator generator;
   } files[] = {
-      {"config.cmake.in", GenerateCmakeConfigIn},
       {"README.md", GenerateReadme},
       {"BUILD.bazel", GenerateBuild},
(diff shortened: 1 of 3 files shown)


generator/internal/scaffold_generator.cc
```diff
@@ -326,37 +325,6 @@ void GenerateScaffold(
   GenerateQuickstartWorkspace(os, vars, contents);
 }
 
-void GenerateCmakeConfigIn(
```
devbww, line 329: There is a declaration for this in the header.

dbolduc: D'oh. Removed.

devbww (approved):

PR #12493, opened 2023-08-29
fix: URL encode explicit routing values

alevenberg:
Part of #12232

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12493)

### DIFF
--- a/generator/integration_tests/golden/v1/internal/golden_kitchen_sink_metadata_decorator.cc
+++ b/generator/integration_tests/golden/v1/internal/golden_kitchen_sink_metadata_decorator.cc
@@ -178,15 +178,15 @@ GoldenKitchenSinkMetadata::ExplicitRouting2(
   params.reserve(2);
 
   if (!request.app_profile_id().empty()) {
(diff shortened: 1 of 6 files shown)


alevenberg:
I ran `i/cloudbuild/build.sh -t asan-pr `

I'm not sure what unit tests I should change for this change

dbolduc:
* `GENERATE_GOLDEN_ONLY` is good for developing, but we should remove it and rerun `generate-libraries-pr` before sending the PR.
* I would add a new unit test in [golden_kitchen_sink_metadata_decorator_test.cc](https://github.com/googleapis/google-cloud-cpp/blob/main/generator/integration_tests/tests/golden_kitchen_sink_metadata_decorator_test.cc). Look at the `ExplicitRouting*` tests for inspiration.
* We should probably make identical changes to the REST metadata decorator

alevenberg:
I did the first two.

Looking into the rest metadata decorator.

dbolduc (approved):
> Maybe in a 2nd PR?

SGTM

PR #12502, opened 2023-08-30
refactor(cmake)!: only compile protos if asked - logging

dbolduc:
Part of the work for #8022 and #12428

Only `iam` remains for both...

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12502)

### DIFF
--- a/google/cloud/appengine/CMakeLists.txt
+++ b/google/cloud/appengine/CMakeLists.txt
@@ -18,7 +18,8 @@ include(GoogleCloudCppLibrary)
 
 set(GOOGLE_CLOUD_CPP_SERVICE_DIRS "" "v1/")
 
-google_cloud_cpp_add_ga_grpc_library(appengine "App Engine Admin API")
+google_cloud_cpp_add_ga_grpc_library(appengine "App Engine Admin API"
+                                     SHARED_PROTO_DEPS "logging_type")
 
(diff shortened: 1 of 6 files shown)


google/cloud/appengine/CMakeLists.txt
```diff
@@ -18,7 +18,8 @@ include(GoogleCloudCppLibrary)
 
 set(GOOGLE_CLOUD_CPP_SERVICE_DIRS "" "v1/")
 
-google_cloud_cpp_add_ga_grpc_library(appengine "App Engine Admin API")
+google_cloud_cpp_add_ga_grpc_library(appengine "App Engine Admin API"
+                                     SHARED_PROTO_DEPS "logging_type")
```
coryan, line 22: I see, `logging_type` is generated here.

coryan (approved):

google/cloud/appengine/CMakeLists.txt
dbolduc: yeah, it will be generated by whatever comes first in the alphabet, pretty much.

If you `-DGOOGLE_CLOUD_CPP_ENABLE=logging`, then the `google/cloud/logging` subdirectory will be responsible for generating `logging_type`.

PR #12507, opened 2023-08-31
doc: CHANGELOG entry for #8022

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12507)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,7 +171,7 @@ If you are dependent on any of these proto libraries **and** you are not
 compiling the corresponding client library, you will need to update your build
 scripts.
 
-For example, if you depend on `google_cloud_cpp_speech_protos`, (e.g., if you
+For example, if you depend on `google_cloud_cpp_speech_protos` (e.g., if you
 have been using this library to make calls to Cloud Speech using raw gRPC), add
 `-DGOOGLE_CLOUD_CPP_ENABLE=...,speech,...` to your CMake configure command. If
 you build with `vcpkg`, include `speech` in your install command.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,11 +140,19 @@ The library has been expanded to include the v2 service.
 consider this a breaking change, as Bazel 5.x has been in maintenance mode for
 more than 6 months.
 
-**CMake Proto Libraries**: We no longer compile certain protos, unless the
+**CMake Proto Libraries**: We only compile service-specific protos if the
 corresponding client library is enabled, via `-DGOOGLE_CLOUD_CPP_ENABLE=...`.
 
-This change reduces build times with `cmake` for customers who build with
-`cmake` and who are not using **all** of the client libraries listed below.
+This change reduces build times for customers who use CMake but who are not
+using **all** of the client libraries listed below.
+
+We considered it [a bug][#8022] that customers building with CMake were forced
+to compile large proto libraries that they did not need. It was certainly
+[confusing][#10174].
+
+Any change in behavior, including fixing bugs, can be considered "breaking". By
+policy we don't consider bug fixes to be breaking changes. We applied that
+policy in this case.
 
 The impacted libraries are:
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -140,6 +140,48 @@ The library has been expanded to include the v2 service.
 consider this a breaking change, as Bazel 5.x has been in maintenance mode for
 more than 6 months.
 
+**CMake Proto Libraries**: We no longer compile certain protos, unless the
```
coryan, line 143: Maybe "we only compile the service-specific protos if the corresponding client library is enabled via `-DGOOGLE_CLOUD_CPP_ENABLE" ?

CHANGELOG.md
```diff
@@ -140,6 +140,48 @@ The library has been expanded to include the v2 service.
...
+corresponding client library is enabled, via `-DGOOGLE_CLOUD_CPP_ENABLE=...`.
+
+This change reduces build times with `cmake` for customers who build with
```
coryan, line 146: I would use CMake (without code style) in this context.  I think we should use `cmake` (with code style) when referring to the command itself, not the system.

CHANGELOG.md
```diff
@@ -140,6 +140,48 @@ The library has been expanded to include the v2 service.
...
+`experimental-storage-grpc` feature, not the `storage` feature.
+
+We do not make changes that can break our customers lightly. We considered it
+[a bug][#8022] that the majority of customers building with `cmake` were forced
+to compile large proto libraries that they did not need. It was needlessly
+[confusing][#10174]. And we expect the number of customers who will benefit from
+this change eclipse the number of customers who are affected negatively by it.
+Hence, we think this change is necessary.
+
+We offered many of these proto libraries before we could write and maintain the
```
coryan, line 181: This information probably belongs in the comments on #8022, but maybe not here.

CHANGELOG.md
```diff
@@ -140,6 +140,48 @@ The library has been expanded to include the v2 service.
...
+
+For example, if you depend on `google_cloud_cpp_speech_protos`, (e.g. if you
+have been using this library to make calls to Cloud Speech using raw gRPC), add
+`-DGOOGLE_CLOUD_CPP_ENABLE=...,speech,...` to your CMake configure command. If
+you build with `vcpkg`, include `speech` in your install command.
+
+Note that `google_cloud_cpp_storage_protos` are associated with the
+`experimental-storage-grpc` feature, not the `storage` feature.
+
+We do not make changes that can break our customers lightly. We considered it
```
coryan, line 174: Maybe move the paragraph earlier, maybe before "The impacted libraries".  And maybe consider something like:

> Compiling these protos when not strictly needed was a bug for some customers, and confusing to other customers.
> Any change in behavior, including fixing bugs, can be considered "breaking". By policy we don't consider bug fixes to be breaking changes. And we applied that policy in this case.

Skip the comment about the number of customers affected.

coryan (changes requested):

CHANGELOG.md
devbww: And remove the *"with CMake"* duplication.  So, *"This change reduces build times for customers who use CMake but who are not using ...."*

CHANGELOG.md
dbolduc: Done.

CHANGELOG.md
dbolduc: Done and done.

CHANGELOG.md
dbolduc: > Maybe move the paragraph earlier, maybe before "The impacted libraries".

Done.

>  maybe consider something like:

I used your second paragraph, but I preferred my first paragraph.

> Skip the comment about the number of customers affected.

Done.

CHANGELOG.md
```diff
@@ -140,6 +140,45 @@ The library has been expanded to include the v2 service.
...
+- `speech`
+- `storage`
+- `texttospeech`
+- `trace`
+
+If you are dependent on any of these proto libraries **and** you are not
+compiling the corresponding client library, you will need to update your build
+scripts.
+
+For example, if you depend on `google_cloud_cpp_speech_protos`, (e.g. if you
```
devbww, line 174: `s/, (e.g. if/ (e.g., if/`

devbww (approved):

CHANGELOG.md
dbolduc: took two commits, but done.

PR #12517, opened 2023-08-31
fix: remove testing_util::chrono_literals from async test

alevenberg:
Due to clang tool failing to build in google3 import

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12517)

### DIFF
--- a/google/cloud/internal/async_retry_unary_rpc_test.cc
+++ b/google/cloud/internal/async_retry_unary_rpc_test.cc
@@ -38,6 +38,7 @@ using ::testing::HasSubstr;
 using ::testing::Return;
 using Request = google::protobuf::Timestamp;
 using Response = google::protobuf::Duration;
+using namespace std::chrono_literals;
 
 /**
  * A class to test the retry loop.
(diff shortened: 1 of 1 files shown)


google/cloud/internal/async_retry_unary_rpc_test.cc
```diff
@@ -38,6 +38,7 @@ using ::testing::HasSubstr;
 using ::testing::Return;
 using Request = google::protobuf::Timestamp;
 using Response = google::protobuf::Duration;
+using namespace std::chrono_literals;
```
devbww, line 41: I'm not sure this is right, because the `std` chrono literals are like `10us`, not the `10_us` we get from google/cloud/testing_util/chrono_literals.h.  Did this actually fix things in the import?

alevenberg: It's still building... 

I'll update this if (or when) it fails :(

alevenberg: Didn't like that even when I changed then to 10us

Trying the 10_us -> us(10)

alevenberg: Built in google3!

devbww (approved):
You'll have to update the PR title during the merge.

alevenberg:
Updated PR title

PR #12518, opened 2023-08-31
fix(cmake): shared proto deps in pc files

dbolduc:
Part of the work for #8022, should have been done in #12514. Sorry for the churn.

We also need to add the shared proto deps to the package config file.

CMake readability note: I opted to link the private targets (`google_cloud_cpp_foo`) instead of the public ones (`google-cloud-cpp::foo`) because it saved me a variable and some list transforms. :shrug:

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12518)

### DIFF
--- a/cmake/GoogleCloudCppLibrary.cmake
+++ b/cmake/GoogleCloudCppLibrary.cmake
@@ -210,7 +210,7 @@ function (google_cloud_cpp_add_ga_grpc_library library display_name)
                                         ${_opt_ADDITIONAL_PROTO_LISTS})
 
     set(shared_proto_dep_targets "${_opt_SHARED_PROTO_DEPS}")
-    list(TRANSFORM shared_proto_dep_targets PREPEND "google-cloud-cpp::")
+    list(TRANSFORM shared_proto_dep_targets PREPEND "google_cloud_cpp_")
     list(TRANSFORM shared_proto_dep_targets APPEND "_protos")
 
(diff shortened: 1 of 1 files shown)


cmake/GoogleCloudCppLibrary.cmake
```diff
@@ -307,9 +307,12 @@ function (google_cloud_cpp_add_ga_grpc_library library display_name)
...
 
     google_cloud_cpp_add_pkgconfig(
-        ${library} "The ${display_name} C++ Client Library"
+        ${library}
+        "The ${display_name} C++ Client Library"
         "Provides C++ APIs to use the ${display_name}"
-        "google_cloud_cpp_grpc_utils" "${protos_target}")
+        "google_cloud_cpp_grpc_utils"
+        "${protos_target}"
+        ${shared_proto_dep_targets})
```
coryan, line 315: nit: in these context these are pkg-config "modules" (the name of a `.pc` file).  We happen to use the same name for the module and the target and the library.

coryan (approved):

cmake/GoogleCloudCppLibrary.cmake
dbolduc: Ack, thanks. That will help me talk about these things more intelligently.

I don't think I am going to change the variable names though :/

PR #12522, opened 2023-09-01
fix(GCS+gRPC): conflict with Win32 global function

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12522)

### DIFF
--- a/google/cloud/storage/examples/storage_async_samples.cc
+++ b/google/cloud/storage/examples/storage_async_samples.cc
@@ -140,7 +140,7 @@ void InsertObjectVectorVectors(
   (client, argv.at(0), argv.at(1));
 }
 
-// The name is redundant, but worksaround a conflict with the global
+// The name is redundant, but works around a conflict with the global
 // `DeleteObject()` function on Windows.
 void AsyncDeleteObject(google::cloud::storage_experimental::AsyncClient& client,
                        std::vector<std::string> const& argv) {
--- a/google/cloud/storage/examples/storage_async_samples.cc
+++ b/google/cloud/storage/examples/storage_async_samples.cc
@@ -140,8 +140,10 @@ void InsertObjectVectorVectors(
   (client, argv.at(0), argv.at(1));
 }
 
-void DeleteObject(google::cloud::storage_experimental::AsyncClient& client,
-                  std::vector<std::string> const& argv) {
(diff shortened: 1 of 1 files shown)


google/cloud/storage/examples/storage_async_samples.cc
```diff
@@ -140,8 +140,10 @@ void InsertObjectVectorVectors(
   (client, argv.at(0), argv.at(1));
 }
 
-void DeleteObject(google::cloud::storage_experimental::AsyncClient& client,
-                  std::vector<std::string> const& argv) {
+// The name is redundant, but worksaround a conflict with the global
```
dbolduc, line 143: works around

dbolduc (approved):

google/cloud/storage/examples/storage_async_samples.cc
coryan: Fixed and merged

google/cloud/storage/examples/storage_async_samples.cc
```diff
@@ -140,8 +140,10 @@ void InsertObjectVectorVectors(
   (client, argv.at(0), argv.at(1));
 }
 
-void DeleteObject(google::cloud::storage_experimental::AsyncClient& client,
-                  std::vector<std::string> const& argv) {
+// The name is redundant, but works around a conflict with the global
```
devbww, line 143: "The name is arbitrary" perhaps?

coryan: The `Async*` part of the name is redundant in a file called `storage_async_samples.cc`. The file name is also repetitive: google/cloud/**storage**/examples/**storage**_async_samples.cc, but I digress).

But the intent is obviously not clear, maybe "Calling this function `Async*` seems repetitive given the file name"?

devbww: Don't listen to me too much, but "the name is redundant" implies that the name could be removed.  So "the `Async` part of the name is redundant" would be better, yes.

But it sounds like you're really saying is, "We'd like to call this just `DeleteObject()`, but that conflicts with the name of a Windows function."  (Presumably it is actually a macro given that this is in an anonymous namespace.)

Anyway ... not a big deal of course.  Forget about it if the argument isn't compelling enough to warrant a change.

PR #12524, opened 2023-09-01
fix: remove chrono_literals namespace and add namespace alias

alevenberg:
The google3 import broke when using the chrono literal. 

I think this is a good fix. I did not test it in google3 yet.

### DIFF
--- a/google/cloud/storage/internal/async/connection_impl_test.cc
+++ b/google/cloud/storage/internal/async/connection_impl_test.cc
@@ -45,6 +45,7 @@ using ::testing::_;
 using ::testing::HasSubstr;
 using ::testing::Pair;
 using ::testing::UnorderedElementsAre;
+using ms=std::chrono::milliseconds;
 
 using AsyncWriteObjectStream =
     ::google::cloud::internal::AsyncStreamingWriteRpc<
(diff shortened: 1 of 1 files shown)


coryan (approved):

google/cloud/storage/internal/async/connection_impl_test.cc
```diff
@@ -45,6 +45,7 @@ using ::testing::_;
 using ::testing::HasSubstr;
 using ::testing::Pair;
 using ::testing::UnorderedElementsAre;
+using ms=std::chrono::milliseconds;
```
coryan, line 48: Consider moving this to the one function where it is used.

alevenberg:
> FWIW, there are dozens of other places in the code where we use `_ms` and other literals. I think the import failures are fairly specific to one file.

I'm guessing because we changed the BUILD rule for that file in the import so whatever checker ran over it and did not like it. Probably best practice to start using ms() or us() instead.

google/cloud/storage/internal/async/connection_impl_test.cc
alevenberg: Done

PR #12541, opened 2023-09-05
chore: update googleapis SHA circa 2023-09-05

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12541)


alevenberg (approved):

coryan:
@alevenberg I had to make one more change.  You may want to take another look.

generator/internal/descriptor_utils.cc
```diff
@@ -439,6 +439,11 @@ ParameterCommentSubstitution substitutions[] = {
     {"`filter='strategy=TrafficPercentStrategy'`",
      "``filter='strategy=TrafficPercentStrategy'``"},
 
+    {"conversionWorkspace/123/mappingRules/"
```
alevenberg, line 442: I can see that both of these strings are in the doxygen comments for the data migration client, but I don't know why there is a mapping from "text" -> "`text`"

In the client, it looks like we are using both in line 1963 and line 1968.

coryan: Because `@foo` (or more precisely `@c7cfa2a8c7cfa2a8c7cfa2a8c7cfa2a8`) is treated as a Doxygen command unless is enclosed in backticks.  Doxygen breaks because that is not a valid command.

PR #12558, opened 2023-09-06
fix(internal): add url encoding for minimal iam creds

alevenberg:
Part of https://github.com/googleapis/google-cloud-cpp/issues/12232

I added no tests since DISABLED_AsyncGenerateAccessTokenMetadata should cover the case with the changes to the ValidateMetadataFixture in https://github.com/googleapis/google-cloud-cpp/pull/12544

I tried undisabling the test and it passed using bazel, so I think the issue is something with the Cmake configuration for the iam protos, but nothing stood out to me.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12558)

### DIFF
--- a/google/cloud/internal/minimal_iam_credentials_stub.cc
+++ b/google/cloud/internal/minimal_iam_credentials_stub.cc
@@ -21,6 +21,7 @@
 #include "google/cloud/internal/log_wrapper.h"
 #include "google/cloud/internal/opentelemetry.h"
 #include "google/cloud/internal/unified_grpc_credentials.h"
+#include "google/cloud/internal/url_encode.h"
 #include <google/iam/credentials/v1/iamcredentials.grpc.pb.h>
 
 namespace google {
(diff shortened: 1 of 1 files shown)


coryan:
Can you send a separate PR to enable the test and see what fails?

coryan (approved):

alevenberg:
> Can you send a separate PR to enable the test and see what fails?

I ran it locally and got:

```
/usr/local/google/home/alevenb/google/dev/google-cloud-cpp-issues/google/cloud/testing_util/validate_metadata.cc:324: Failure
Value of: method
Expected: isn't NULL
  Actual: NULL (of type google::protobuf::MethodDescriptor const*)
Method google.iam.credentials.v1.IAMCredentials.GenerateAccessToken is unknown.
```
There was a comment above the test: `// For some reasons the method name is not found in IsContextMDValid()`, which should have been an indicator to me that someone else already tried to resolve the issue.

Specifically failing at https://github.com/alevenberg/google-cloud-cpp/blob/6e9d901672d54851ed718b5723936a0849ba362a/google/cloud/testing_util/validate_metadata.cc#L322

I couldn't find a method for the DescriptorPool to "print all methods" and the FindMethodByName returns null if nothing is found.

Since it was working for bazel, I'm guessing this has to do with how we are building in cmake

Here is the PR: https://github.com/googleapis/google-cloud-cpp/pull/12560

PR #12572, opened 2023-09-07
cleanup: remove inline namespace usage from tests

devbww:
Tests, like any other user of the library, should not care about versioning in the single release they link against.  Nor do they need to version their own symbols.

So, remove the inline namespaces from tests.  This then removes the need to include `google/cloud/version.h`, eliminating warnings that we failed to do so.

One slight complication is test-only friends, which should now be forward-declared outside the inline namespace.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12572)


alevenberg:
What script did you use to generate the PR?

devbww:
> What script did you use to generate the PR?

I found the tests that used the inline namespace with something like ...
```
git grep -l GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_ | grep '_test\.cc$'
```

And then a `sed -i` script over those files to delete those macros and any `google/cloud/version.h` include.

PR #12573, opened 2023-09-07
chore(deps): tweak definition of libpfm dependency

devbww:
Followup to #12568.

Remove the `str()` wrapper around `Label()`, which confuses the deps cacher script `bazel/deps-cache.py`.

Also remove the `/download` suffix from the URL, that similarly confused the deps cacher, and which appears to be superfluous.  This also removes the need for the `type` specifier.

Finally, reorder the `maybe()` arguments to match the other invocations. Provide an alternative URL to

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12573)

### DIFF
--- a/bazel/google_cloud_cpp_deps.bzl
+++ b/bazel/google_cloud_cpp_deps.bzl
@@ -44,14 +44,13 @@ def google_cloud_cpp_development_deps(name = None):
     maybe(
         http_archive,
         name = "libpfm",
-        build_file = str(Label("//bazel:libpfm.BUILD")),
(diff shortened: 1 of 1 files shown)


coryan (approved):
I wonder if we should tweak this to define an empty project instead.  We don't really use it, and it is optional in google/benchmark:

https://github.com/google/benchmark/blob/344117638c8ff7e239044fd0fa7085839fc03021/BUILD.bazel#L64-L72

devbww:
> I wonder if we should tweak this to define an empty project instead. We don't really use it, and it is optional in google/benchmark

Understanding the benefits of "perfcounters" is another whole thing, so I'll leave this as-is for now.  Thanks.

PR #12577, opened 2023-09-08
docs(storage): more on InsertObject vs. WriteObject

coryan:
Show that both `InsertObject()` and `WriteObject()` can be used to upload data from memory. At least one customer missed the existence of `InsertObject()` because they thought the example showed **all** the ways.

Fixed the links showing the tradeoffs between single-shot and resumable-uploads. Be more explicit about what upload type does `InsertObject()` use.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12577)

### DIFF
--- a/google/cloud/storage/client.h
+++ b/google/cloud/storage/client.h
@@ -854,6 +854,12 @@ class Client {
   /**
    * Creates an object given its name and contents.
    *
+   * If you need to perform larger uploads or uploads where the data is not
+   * contiguous in memory use `WriteObject()`. This function always performs a
+   * single-shot upload, while `WriteObject()` always uses resumable uploads.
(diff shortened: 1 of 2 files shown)


google/cloud/storage/client.h
```diff
@@ -854,6 +854,12 @@ class Client {
   /**
    * Creates an object given its name and contents.
    *
+   * If you need to perform larger uploads or uploads where the data is not
+   * contiguous in memory use `WriteObject()`. This function always performs a
```
devbww, line 858: s/memory use/memory, use/

devbww (approved):

google/cloud/storage/client.h
coryan: Thanks

Issue #12562, opened 2023-09-06
Generated client libraries send wrong library type in x-goog-api-client header


dangazineu:
The issue seems to be "gccl/" hardcoded here:
https://github.com/googleapis/google-cloud-cpp/blob/5c780b27aa2bd4a6df14760b749dd84fdef123ad/google/cloud/internal/api_client_header.cc#L37

For 100% generated libraries, the value should be "gapic/"

coryan:
What should the value be for partially generated libraries?

coryan:
> What should the value be for partially generated libraries?

Doh! The description says "100% generated".  That answer my question.  I think with the last PR this is fixed.  I would like to visually inspect some requests and see how the integration tests show up in the request dashboards.

coryan:
I verified the `x-goog-api-client-header` is correct using the debugger:

```
Thread 1 "translate_quick" hit Breakpoint 1, google::cloud::translate_v3_internal::v2_16::DefaultTranslationServiceStub::TranslateText (this=0x555556b80af0, client_context=..., request=...) at /usr/local/google/home/coryan/cpp-generate/google/cloud/translate/v3/internal/translation_stub.cc:37
37        google::cloud::translation::v3::TranslateTextResponse response;

(gdb) print client_context.send_initial_metadata_
$3 = std::multimap with 3 elements = {["x-goog-api-client"] = "gl-cpp/GNU-12.2.0-ex-2017 gapic/v2.16.0-rc+generated", ["x-goog-request-params"] = "parent=projects%2Fcloud-cpp-testing-resources", ["x-goog-user-project"] = "cloud-cpp-testing-resources"}

```

PR #12582, opened 2023-09-08
impl(storage): identify generated libraries

coryan:
Use a different x-goog-api-client header for the RPCs using fully generated code vs. the RPCs using at least some hand-crafted code.

Part of the work for #12562

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12582)

### DIFF
--- a/google/cloud/storage/internal/storage_stub_factory_test.cc
+++ b/google/cloud/storage/internal/storage_stub_factory_test.cc
@@ -54,8 +54,7 @@ class StorageStubFactory : public ::testing::Test {
   void IsContextMDValid(grpc::ClientContext& context, std::string const& method,
                         google::protobuf::Message const& request) {
     return validate_metadata_fixture_.IsContextMDValid(
-        context, method, request,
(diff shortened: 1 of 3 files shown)


google/cloud/storage/internal/storage_stub_factory_test.cc
```diff
@@ -55,7 +55,7 @@ class StorageStubFactory : public ::testing::Test {
                         google::protobuf::Message const& request) {
     return validate_metadata_fixture_.IsContextMDValid(
         context, method, request,
-        google::cloud::internal::ApiClientHeader("generator"));
+        google::cloud::internal::HandCraftedLibClientHeader());
```
devbww, line 58: Should this be calling `storage::x_goog_api_client()`?

devbww (approved):

google/cloud/storage/internal/storage_stub_factory_test.cc
coryan: Done.

PR #12605, opened 2023-09-09
impl(generator): use singular form of discovery resource names as appropriate

### LINKED ISSUE #11138: Use correct plural/singular form for resource methods
When we format method names for primitive operation methods, we need to change the typically plural resource name from the discovery doc to the singular form for all primitives except for `aggregatedList` and `list`.

scotthart:
fixes #11138 

I attempted a more generic solution leveraging grammar rules, but while going from singular to plural is somewhat doable (minus the exceptions), going from plural to singular is much less deterministic and more trial an error.


coryan:
I wonder if we will regret fixing the plural vs. singular names of these RPCs in the future.  Eventually we want to update the discovery doc and the generated files as part of the weekly "update the SHA" task for the operator, and we may want to automate that away.

Having a custom list of transformations will require human judgement and intervention when new resources are added to compute.  Is that something we want to take on?

What would we do if a proto for any other random API had the wrong number in one RPC, would we try to fix that?

Not that this is binding, but do other languages do this for DIREGAPIC services?

> I attempted a more generic solution leveraging grammar rules, but while going from singular to plural is somewhat doable (minus the exceptions), going from plural to singular is much less deterministic and more trial and error.

Is it possible to use the name of the `GET` response type to determine the singular name of the resource?  Or some other field that maybe requires less human intervention or manual updates?

google/cloud/compute/projects/v1/projects.proto
```diff
@@ -102,7 +102,7 @@ service Projects {
   // need. For example, to only include the `id` and `selfLink` fields, add the
   // query parameter `?fields=id,selfLink` to your request.
   // https://cloud.google.com/compute/docs/reference/rest/v1/projects/get
-  rpc GetProjects(GetProjectsRequest)
```
coryan, line 105: There is a `GetXpnResources()` rpc defined a few lines below.  Should that change?

scotthart (commented):
Using the `GET` response schema name works well and removes the burden of adding new exceptions to the code. It does slightly alter some of the method names by removing qualifiers that could be deemed redundant as such qualifiers also appear in the service name and package name. I think it's the best path forward.

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12605#-:-Ne5-LFqAtTFwNFEjJED:bbmsepi)* status: 0 of 1653 files reviewed, 1 unresolved discussion (waiting on @coryan)

___
*[`google/cloud/compute/projects/v1/projects.proto` line 0 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12605#gh-1320667138:-Ne500nN88LP97ZCrQ3c:btae0p5) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/f13d525f7afb86ce454cd82868cb4c29d759febe/google/cloud/compute/projects/v1/projects.proto#L0)):*

There is a `GetXpnResources()` rpc defined a few lines below.  Should that change?

</blockquote></details>

We're only attempting to use the correct singular noun for the generic http crud methods (delete, get, insert, list, update, etc.) that do not include the name (resource name) of what they are operating upon. Method names outside of these we copy verbatim from the discovery document.

coryan (commented):
Is this something we **need** to do?  As opposed to something that is nice of us to do?

If we had received a number of proto files with the wrong number for some operations, for example, if Bigtable called the message `InsertTablesRequest` and the RPC was `InsertTable()` would we change it?

Not that this is binding, but it may provide context: do the other languages make such changes for DIREGAPIC libraries?

> Reviewed 753 of 1652 files at r1, 900 of 900 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12605#-:-NdwBCJb04AEr4R3uN1D:b-jd3m6b)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

coryan (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12605#-:-Ne5F6yW-RNIt4-jdMhh:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @scotthart)

PR #12615, opened 2023-09-11
impl(pubsub): add data size accessor for publish span

alevenberg:
Part of [12525](https://github.com/googleapis/google-cloud-cpp/issues/12525)

### DIFF
--- a/google/cloud/pubsub/message.h
+++ b/google/cloud/pubsub/message.h
@@ -83,6 +83,7 @@ class Message {
   PubsubMessageDataType&& data() && {
     return std::move(*proto_.mutable_data());
   }
+  size_t data_size() const { return proto_.data().size(); }
   std::string const& message_id() const { return proto_.message_id(); }
   std::string const& ordering_key() const { return proto_.ordering_key(); }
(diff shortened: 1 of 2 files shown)


google/cloud/pubsub/message.h
```diff
@@ -83,6 +83,7 @@ class Message {
   PubsubMessageDataType&& data() && {
     return std::move(*proto_.mutable_data());
   }
+  size_t data_size() const { return proto_.data().size(); }
```
coryan, line 86: We should avoid adding public APIs unless they are needed by our customers.  In this case, there is already a `pubsub_internal::MessageSize(pubsub::Message const&)` function that you can use.

alevenberg: That includes the size of the attributes, but this attribute is supposed to only be the data size. I can use the .data accessor and calculate the size directly if we don't want to add this to the api.

coryan: Sure.  Though I suspect the data size is not the important information to save in traces.  The CPU costs, network transfer time, and dollar costs are all controlled by the total size of the message, including attributes.

PR #12617, opened 2023-09-11
impl(pubsub): add start and end span for publisher tracing connection

alevenberg:
Part of #[12525](https://github.com/googleapis/google-cloud-cpp/issues/12525)

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12617)

### DIFF
--- a/google/cloud/pubsub/internal/publisher_tracing_connection.cc
+++ b/google/cloud/pubsub/internal/publisher_tracing_connection.cc
@@ -15,7 +15,7 @@
 #include "google/cloud/pubsub/internal/publisher_tracing_connection.h"
 #include "google/cloud/pubsub/message.h"
 #include "google/cloud/pubsub/publisher_connection.h"
-#include "google/cloud/future_generic.h"
+#include "google/cloud/future.h"
 #include "google/cloud/status.h"
 #include "google/cloud/version.h"
 #include <memory>
@@ -91,28 +91,43 @@ future<StatusOr<std::string>> EndPublishSpan(
   });
 }
 
+/**
+ * A decorator that adds tracing for the PublisherConnection.
+ */
+class PublisherTracingConnection : public pubsub::PublisherConnection {
+ public:
+  explicit PublisherTracingConnection(
+      pubsub::Topic topic, std::shared_ptr<pubsub::PublisherConnection> child)
+      : topic_(std::move(topic)), child_(std::move(child)) {}
+
+  ~PublisherTracingConnection() override = default;
+
+  future<StatusOr<std::string>> Publish(PublishParams p) override {
+    auto span = StartPublishSpan(topic_.FullName(), p.message);
+    auto scope = opentelemetry::trace::Scope(span);
+    return EndPublishSpan(std::move(span), child_->Publish(std::move(p)));
+  };
+
+  void Flush(FlushParams p) override {
+    auto span = internal::MakeSpan("pubsub::Publisher::Flush");
+    auto scope = opentelemetry::trace::Scope(span);
+    child_->Flush(std::move(p));
+    internal::EndSpan(*span);
+  }
+
+  void ResumePublish(ResumePublishParams p) override {
+    auto span = internal::MakeSpan("pubsub::Publisher::ResumePublish");
+    auto scope = opentelemetry::trace::Scope(span);
+    child_->ResumePublish(std::move(p));
+    internal::EndSpan(*span);
+  };
+
+ private:
+  pubsub::Topic const topic_;
+  std::shared_ptr<pubsub::PublisherConnection> child_;
+};
+
 }  // namespace
-
-future<StatusOr<std::string>> PublisherTracingConnection::Publish(
-    PublishParams p) {
-  auto span = StartPublishSpan(topic_.FullName(), p.message);
-  auto scope = opentelemetry::trace::Scope(span);
-  return EndPublishSpan(std::move(span), child_->Publish(std::move(p)));
-}
-
-void PublisherTracingConnection::Flush(FlushParams p) {
-  auto span = internal::MakeSpan("pubsub::Publisher::Flush");
-  auto scope = opentelemetry::trace::Scope(span);
-  child_->Flush(std::move(p));
-  internal::EndSpan(*span);
-}
-
-void PublisherTracingConnection::ResumePublish(ResumePublishParams p) {
-  auto span = internal::MakeSpan("pubsub::Publisher::ResumePublish");
-  auto scope = opentelemetry::trace::Scope(span);
-  child_->ResumePublish(std::move(p));
-  internal::EndSpan(*span);
-}
 
 std::shared_ptr<pubsub::PublisherConnection> MakePublisherTracingConnection(
     pubsub::Topic topic,
--- a/google/cloud/pubsub/internal/publisher_tracing_connection.cc
+++ b/google/cloud/pubsub/internal/publisher_tracing_connection.cc
@@ -126,6 +126,8 @@ std::shared_ptr<pubsub::PublisherConnection> MakePublisherTracingConnection(
 std::shared_ptr<pubsub::PublisherConnection> MakePublisherTracingConnection(
     pubsub::Topic topic,
     std::shared_ptr<pubsub::PublisherConnection> connection) {
+  // Silence the unused parameter warnings.
+  (void)topic;
   return connection;
 }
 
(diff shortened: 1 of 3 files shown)


google/cloud/pubsub/internal/publisher_tracing_connection.cc
```diff
@@ -52,14 +115,19 @@ void PublisherTracingConnection::ResumePublish(ResumePublishParams p) {
...
+                                                      std::move(connection));
 }
 
 #else  // GOOGLE_CLOUD_CPP_HAVE_OPENTELEMETRY
 
 std::shared_ptr<pubsub::PublisherConnection> MakePublisherTracingConnection(
+    pubsub::Topic topic,
     std::shared_ptr<pubsub::PublisherConnection> connection) {
+  // Silence the unused parameter warnings.
+  (void)topic;
```
devbww, line 130: I believe you can just not give the parameter a name to achieve the same effect.

google/cloud/pubsub/internal/publisher_tracing_connection_test.cc
```diff
@@ -15,35 +15,134 @@
...
+}
+
+TEST(PublisherTracingConnectionTest, PublishSpanOnError) {
+  namespace sc = ::opentelemetry::trace::SemanticConventions;
+  auto span_catcher = InstallSpanCatcher();
+  auto mock = std::make_shared<MockPublisherConnection>();
+  EXPECT_CALL(*mock, Publish)
+      .WillOnce([&](pubsub::PublisherConnection::PublishParams const&) {
+        EXPECT_TRUE(ThereIsAnActiveSpan());
+        return google::cloud::make_ready_future(
```
coryan, line 105: We rarely fully qualify types in `google::cloud`.  Your call.

coryan (approved):
A few comments for you to consider, but I would be happy if the code got it as-is.

google/cloud/pubsub/internal/publisher_tracing_connection.cc
```diff
@@ -13,15 +13,22 @@
 // limitations under the License.
 
 #include "google/cloud/pubsub/internal/publisher_tracing_connection.h"
+#include "google/cloud/pubsub/message.h"
 #include "google/cloud/pubsub/publisher_connection.h"
+#include "google/cloud/future_generic.h"
```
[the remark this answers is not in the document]
alevenberg: Whoops. I think this was clang-tidy. Changed.

google/cloud/pubsub/internal/publisher_tracing_connection_test.cc
alevenberg: Changed.

google/cloud/pubsub/internal/publisher_tracing_connection.cc
```diff
@@ -30,11 +37,67 @@ namespace pubsub_internal {
...
+  options.kind = opentelemetry::trace::SpanKind::kProducer;
+  auto span =
+      internal::GetTracer(internal::CurrentOptions())
+          ->StartSpan(
+              topic + " send",
+              {
+                  {sc::kMessagingSystem, "pubsub"},
+                  {sc::kMessagingDestinationName, topic},
+                  {sc::kMessagingDestinationTemplate, "topic"},
+                  {"messaging.pubsub.ordering_key", m.ordering_key()},
```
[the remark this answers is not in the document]
alevenberg: That's fair. I think that means using a different OTel api though, so I will make those changes in a separate PR.

#12620

google/cloud/pubsub/internal/publisher_tracing_connection.cc
```diff
@@ -30,11 +37,67 @@ namespace pubsub_internal {
...
+future<StatusOr<std::string>> EndPublishSpan(
+    opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> span,
+    future<StatusOr<std::string>> f) {
+  return f.then([span = std::move(span)](auto fut) {
+    auto message_id = fut.get();
+    Status const& status = message_id.status();
+    if (status.ok()) {
+      span->SetStatus(opentelemetry::trace::StatusCode::kOk);
+      span->SetAttribute("gcloud.status_code", 0);
+      span->SetAttribute("messaging.message_id", message_id.value());
```
dbolduc, line 71: Consider just:

```cc
  return f.then([span = std::move(span)](auto fut) {
    auto message_id = fut.get();
    if (message_id) span->SetAttribute("messaging.message_id", *message_id);
    return EndSpan(span, message_id);
  }
```

google/cloud/pubsub/internal/publisher_tracing_connection_test.cc
```diff
@@ -15,35 +15,132 @@
...
-#include "google/cloud/internal/opentelemetry.h"
-#include "google/cloud/opentelemetry_options.h"
+#include "google/cloud/pubsub/topic.h"
+#include "google/cloud/future_generic.h"
+#include "google/cloud/internal/make_status.h"
+#include "google/cloud/status.h"
+#include "google/cloud/status_or.h"
 #include "google/cloud/testing_util/opentelemetry_matchers.h"
+#include "google/cloud/testing_util/status_matchers.h"
+#include "gtest/gtest.h"
```
dbolduc, line 28: we should probably remove this include. Or use `<gtest/gtest.h>` if it is actually necessary.

google/cloud/pubsub/internal/publisher_tracing_connection_test.cc
```diff
@@ -15,35 +15,132 @@
...
+      Topic("test-project", "test-topic"), std::move(mock));
+
+  auto response = connection
+                      ->Publish({pubsub::MessageBuilder{}
+                                     .SetData("test-data-0")
+                                     .SetOrderingKey("ordering-key-0")
+                                     .Build()})
+                      .get();
+
+  EXPECT_THAT(response, StatusIs(StatusCode::kOk));
```
dbolduc, line 75: nit: I think we tend to say `EXPECT_STATUS_OK(response)`

google/cloud/pubsub/internal/publisher_tracing_connection_test.cc
```diff
@@ -15,35 +15,132 @@
...
+              OTelAttribute<std::string>(sc::kMessagingSystem, "pubsub"),
+              OTelAttribute<std::string>(
+                  sc::kMessagingDestinationName,
+                  "projects/test-project/topics/test-topic"),
+              OTelAttribute<std::string>(sc::kMessagingDestinationTemplate,
+                                         "topic"),
+              OTelAttribute<std::string>("messaging.pubsub.ordering_key",
+                                         "ordering-key-0"),
+              OTelAttribute<int>("gcloud.status_code", 0),
+              OTelAttribute<size_t>("messaging.message.total_size_bytes", 45),
```
dbolduc, line 92: nit: I don't think `std::size_t` is a consistently sized attribute type: https://github.com/open-telemetry/opentelemetry-cpp/blob/cdbc90abf3b7170f898439fd34126469c42686c6/sdk/include/opentelemetry/sdk/common/attribute_utils.h#L30-L44

Maybe use `std::uint64_t`?

dbolduc (approved):
Just nits

google/cloud/pubsub/internal/publisher_tracing_connection.cc
alevenberg: Nice. Much cleaner. Done.

google/cloud/pubsub/internal/publisher_tracing_connection_test.cc
alevenberg: Yeah.... I shouldn't accept clang-tidys help.

Whoops.

Removed.

google/cloud/pubsub/internal/publisher_tracing_connection_test.cc
alevenberg: Done

google/cloud/pubsub/internal/publisher_tracing_connection_test.cc
alevenberg: It isn't. I tried `std::uint64_t` and it failed the m32 build. 

I'll try`std::uint32_t`.

alevenberg: From one of the failing builds
```
Step #3:   Actual: { [messaging.message_id=std::string:test-id-0, messaging.message.total_size_bytes=std::uint64_t:45 ]} 

Step #3: matcher #5: has a first field that is equal to "messaging.message.total_size_bytes", and has a second field that is a variant<> with value of type 'unsigned int' and the value is equal to 45
```
Gonna go back to `size_t`.

Let me know if there's a better way to do this. Maybe we could check the architecture size then set the type at run time?

Issue #12451, opened 2023-08-23
Use `HttpTraceContext` and `CloudTraceContext` propagators by default


dbolduc:
@alevenberg reminds me that the plan was to use both of these propagators, instead of making the customer set a global propagator.

These propagators are right in 99% of cases. Only in obscure WIF auth cases do we send a request somewhere other than Google, which might have its own encoding scheme for the trace context.

---

I think we still want to keep the signature taking `Options const&` even if we continue to ignore those options.

---

Relevant code:

https://github.com/googleapis/google-cloud-cpp/blob/a97d711971791e1466481606fea63abf58d3e582/google/cloud/internal/opentelemetry.h#L64-L66

https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/propagation/http_trace_context.h

Issue #11355, opened 2023-04-21
Consider using propagator API for `X-Cloud-Trace-Context`


coryan:
This is progress, but I suspect we would rather use OpenTelemetry's propagator API.

We would define the cloud trace context propagator, and ensure it is used by `InjectTraceContext`.

Maybe just open an issue like `Consider using OpenTelemetry's Propagator API for "X-Cloud-Trace-Context" injection`.

_Originally posted by @dbolduc in https://github.com/googleapis/google-cloud-cpp/pull/11352#pullrequestreview-1396348901_

dbolduc:
If we want to be a goody-two-shoes and use OpenTelemetry's Propagator API, we would have to make a CloudTraceContext textmap propagator. It would look a lot like:

https://github.com/dbolduc/google-cloud-cpp/blob/b65a2fc5ab85850319d0d768a475b13e24326b4e/google/cloud/internal/open_telemetry.cc#L59-L103

Some notes:

- Terminating the `string_view` with a `\0` was not necessary. I was creating a `std::string{sv}` instead of a `std::string{sv.data(), sv.size()}`, later in the code. :monkey: 

- Internally, Google does not want to keep using the `X-Cloud-Trace-Context` header. I have been told that the w3 trace context format is at parity with `X-Cloud-Trace-Context`.... but I think they were lying. All this to say, we still need to use this header, but we should not make a Cloud Trace Context propagator public.

alevenberg:
I think we can break this down into 

- [ ] Create a TextMapCarrier that holds the headers for the `RestContext`
- [ ] Create a custom `opentelemetry::context::propagation::GlobalTextMapPropagator` called `CloudTraceContext` that encodes the field
- [ ] Use the OTel API instead of setting the headers [here](https://github.com/coryan/google-cloud-cpp/blob/81874e0c57e663234d3d8ea67804c451d331757f/google/cloud/internal/tracing_rest_client.cc#L53-L56)

PR #12623, opened 2023-09-12
impl(otel): add cloud trace context propagator

dbolduc:
Part of the work for #12451 and #11355 

- Add an "X-Cloud-Trace-Context" propagator.
- Add a composite propagator that sets the Cloud Trace Context, and the W3 traceparent headers.

The two header formats are not at parity internally in Google. So we should just set both. This is what our OpenTelemetry team recommends.

Note that this PR just adds the propagator. It does not yet use it in requests.

Note also that we add the Cloud Trace Context propagator as an internal only API. We do not want to release this publicly.

---

This implementation code is relevant:

https://github.com/googleapis/google-cloud-cpp/blob/fbea7f0d72fdb5527362c59d79f06bab21dc6947/google/cloud/internal/tracing_rest_client.cc#L41-L58


google/cloud/internal/trace_propagator.h
```diff
@@ -0,0 +1,54 @@
...
+ * servers look for when they receive a request, and we almost exclusively send
+ * requests to Google.
+ *
+ * @see https://opentelemetry.io/docs/instrumentation/cpp/manual/#context-propagation
+ *
+ * [propagator]:
+ * https://opentelemetry.io/docs/reference/specification/context/api-propagators/#textmap-propagator
+ */
+std::unique_ptr<opentelemetry::context::propagation::TextMapPropagator>
+MakePropagator(Options const&);
```
coryan, line 45: For an internal API, I would think it is better to have no arguments vs. consume an `Options const&` that goes unused.  Requiring an unused parameter is viral.

coryan (changes requested):

google/cloud/internal/trace_propagator.h
dbolduc: You are probably right. Fixed.

dbolduc:
PTAL

coryan (approved):

PR #12560, opened 2023-09-06
do not merge or review

alevenberg:
Related to https://github.com/googleapis/google-cloud-cpp/issues/12232

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12560)

PR #12624, opened 2023-09-12
test(common): allow for FindMethodByName() failure

devbww:
`MinimalIamCredentialsStubTest.AsyncGenerateAccessTokenMetadata`: rather than disabling the test altogether, allow for failures of `FindMethodByName()` in `IsContextMDValid()`.  The code for the method may have been dropped by the linker, so it never registers itself.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12624)

### DIFF
--- a/google/cloud/testing_util/validate_metadata.cc
+++ b/google/cloud/testing_util/validate_metadata.cc
@@ -48,7 +48,6 @@ using ::testing::Contains;
 using ::testing::IsEmpty;
 using ::testing::Matcher;
 using ::testing::Not;
-using ::testing::NotNull;
 using ::testing::Pair;
 using ::testing::UnorderedElementsAreArray;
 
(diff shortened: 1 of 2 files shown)


alevenberg:
For context, linking with https://github.com/googleapis/google-cloud-cpp/pull/12560

alevenberg (approved):

google/cloud/testing_util/validate_metadata.cc
```diff
@@ -321,7 +320,11 @@ void ValidateMetadataFixture::IsContextMDValid(
 
   auto const* method =
       DescriptorPool::generated_pool()->FindMethodByName(method_name);
-  ASSERT_THAT(method, NotNull()) << "Method " + method_name + " is unknown.";
+  if (method == nullptr) {
+    GCP_LOG(INFO) << "`x-goog-request-params` header not verified for "
```
coryan, line 324: When running this from the command-line, logging is likely disabled.  I wonder if there is a way to log from googletest.  `SUCCEED() << "msg"` ?  Or `SCOPED_TRACE`?

coryan (approved):

google/cloud/testing_util/validate_metadata.cc
dbolduc: Should we worry that misspelled method names will not get caught?

We can rewrite the IAM credentials stub test to use `ValidateMetadataFixture::GetMetadata` instead of `ValidateMetadataFixture::IsContextMDValid` to not lose any test coverage. (In fact, we could gain some).

devbww: > Should we worry that misspelled method names will not get caught?

Perhaps?  But the response may well be to obtain the name via reflection rather than typing it out.

> We can rewrite the IAM credentials stub test to use `ValidateMetadataFixture::GetMetadata` instead of `ValidateMetadataFixture::IsContextMDValid` to not lose any test coverage. (In fact, we could gain some).

In this PR we're strictly gaining some test coverage, so I'll leave any question of how/what we should be testing to some future PR.

PR #12633, opened 2023-09-14
fix(pubsub): change size_t -> int for otel attribute

alevenberg:
Fix for https://github.com/googleapis/google-cloud-cpp/actions/runs/6168147169/job/16740231408

Caused by https://github.com/googleapis/google-cloud-cpp/pull/12617

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12633)

### DIFF
--- a/google/cloud/pubsub/internal/publisher_tracing_connection.cc
+++ b/google/cloud/pubsub/internal/publisher_tracing_connection.cc
@@ -53,7 +53,7 @@ opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> StartPublishSpan(
                           {sc::kMessagingDestinationTemplate, "topic"},
                           {"messaging.pubsub.ordering_key", m.ordering_key()},
                           {"messaging.message.total_size_bytes",
-                           static_cast<int>(MessageSize(m))},
+                           static_cast<std::int64_t>(MessageSize(m))},
                       },
                       options);
   return span;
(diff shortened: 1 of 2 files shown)


google/cloud/pubsub/internal/publisher_tracing_connection.cc
```diff
@@ -46,16 +46,16 @@ opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> StartPublishSpan(
...
-              },
-              options);
+          ->StartSpan(topic + " send",
+                      {
+                          {sc::kMessagingSystem, "pubsub"},
+                          {sc::kMessagingDestinationName, topic},
+                          {sc::kMessagingDestinationTemplate, "topic"},
+                          {"messaging.pubsub.ordering_key", m.ordering_key()},
+                          {"messaging.message.total_size_bytes",
+                           static_cast<int>(MessageSize(m))},
```
dbolduc, line 56: Googling says pubsub message size is 10MB, which would be greater than the max int32

I think we should use `std::int64_t`

dbolduc (changes requested):

google/cloud/pubsub/internal/publisher_tracing_connection.cc
alevenberg: Done

dbolduc (approved):

PR #12635, opened 2023-09-14
impl(otel): use composite propagator over gRPC

dbolduc:
Part of the work for #12451 

Send both `x-cloud-trace-context` and `traceparent` headers over gRPC when tracing with OTel.

Note that we allocate a few unique_ptrs per request now. In a follow up PR I will cache the propagator in the gRPC stubs, so we only do this once at stub creation time.

Bland test updates are relegated to a second commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12635)

### DIFF
--- a/google/cloud/bigtable/internal/bigtable_stub_factory_test.cc
+++ b/google/cloud/bigtable/internal/bigtable_stub_factory_test.cc
@@ -25,6 +25,7 @@
 #include "google/cloud/testing_util/scoped_log.h"
 #include "google/cloud/testing_util/status_matchers.h"
 #include "google/cloud/testing_util/validate_metadata.h"
+#include "google/cloud/testing_util/validate_propagator.h"
 #include <gmock/gmock.h>
 #include <chrono>
 #include <regex>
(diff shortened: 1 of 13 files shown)


google/cloud/bigtable/internal/bigtable_stub_factory_test.cc
```diff
@@ -303,15 +306,15 @@ using ::testing::IsEmpty;
...
-  EXPECT_CALL(*propagator, Inject);
 
   MockFactory factory;
   EXPECT_CALL(factory, Call)
       .WillOnce([](std::shared_ptr<grpc::Channel> const&) {
         auto mock = std::make_shared<MockBigtableStub>();
-        EXPECT_CALL(*mock, MutateRow)
-            .WillOnce(Return(internal::AbortedError("fail")));
+        EXPECT_CALL(*mock, MutateRow).WillOnce([](auto& context, auto const&) {
+          ValidatePropagator(context);
```
coryan, line 315: Do you need a similar test for `storage_stub_factory.cc`?  It is also hand-crafted.

coryan (approved):

google/cloud/bigtable/internal/bigtable_stub_factory_test.cc
dbolduc: Yeah, we should have a test. I blame you: #11369

dbolduc: I'll send a PR when I get around to it.

Issue #12620, opened 2023-09-11
Only add ordering key attribute on span when it is set


alevenberg:
Add any context about the cleanup here.

PR #12636, opened 2023-09-14
cleanup(pubsub): set attribute only when present

alevenberg:
Closes #[12620](https://github.com/googleapis/google-cloud-cpp/issues/12620)

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12636)

### DIFF
--- a/google/cloud/pubsub/internal/publisher_tracing_connection.cc
+++ b/google/cloud/pubsub/internal/publisher_tracing_connection.cc
@@ -44,19 +44,18 @@ opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> StartPublishSpan(
   namespace sc = opentelemetry::trace::SemanticConventions;
   opentelemetry::trace::StartSpanOptions options;
   options.kind = opentelemetry::trace::SpanKind::kProducer;
-  auto span =
-      internal::GetTracer(internal::CurrentOptions())
-          ->StartSpan(topic + " send",
-                      {
-                          {sc::kMessagingSystem, "pubsub"},
-                          {sc::kMessagingDestinationName, topic},
-                          {sc::kMessagingDestinationTemplate, "topic"},
-                          {"messaging.message.total_size_bytes",
-                           static_cast<std::int64_t>(MessageSize(m))},
-                      },
-                      options);
-    
-      if (!m.ordering_key().empty()) {
+  auto span = internal::GetTracer(internal::CurrentOptions())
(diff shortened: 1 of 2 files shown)


coryan (approved):

google/cloud/pubsub/internal/publisher_tracing_connection_test.cc
```diff
@@ -135,6 +136,35 @@ TEST(PublisherTracingConnectionTest, PublishSpanOnError) {
...
+                      .get();
+
+  EXPECT_STATUS_OK(response);
+  EXPECT_THAT(span_catcher->GetSpans(),
+              ElementsAre(AllOf(
+                  SpanHasInstrumentationScope(), SpanKindIsProducer(),
+                  SpanNamed("projects/test-project/topics/test-topic send"),
+                  SpanWithStatus(opentelemetry::trace::StatusCode::kOk),
+                  Not(SpanHasAttributes(OTelAttribute<std::string>(
+                      "messaging.pubsub.ordering_key", ""))))));
```
dbolduc, line 165: To be more general, I would `s/""/_/`

(and maybe you need a `using ::testing::_;` somewhere.

alevenberg: Done

google/cloud/pubsub/internal/publisher_tracing_connection.cc
```diff
@@ -44,18 +44,20 @@ opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> StartPublishSpan(
...
+                                  {sc::kMessagingSystem, "pubsub"},
+                                  {sc::kMessagingDestinationName, topic},
+                                  {sc::kMessagingDestinationTemplate, "topic"},
+                                  {"messaging.message.total_size_bytes",
+                                   static_cast<std::int64_t>(MessageSize(m))},
+                              },
+                              options);
+
+  if (!m.ordering_key().empty()) {
+    span->SetAttribute("messaging.pubsub.ordering_key", m.ordering_key());
```
dbolduc, line 59: FYI: If we set the attribute after the span is created, our customers can no longer use the ordering_key attribute to make sampling decisions. Is that ok?

alevenberg: Ack. I think it should be ok, but it's a good call out.

PR #12646, opened 2023-09-15
cleanup: deal with some include-what-you-use issues

devbww:
Correct a mostly-random set of include-what-you-use problems by directly including the header previously found transitively.

Also tidy up some `future` comments while we're in the file.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12646)


google/cloud/future_generic.h
```diff
@@ -26,10 +27,12 @@
 #include "google/cloud/internal/future_then_meta.h"
 #include "google/cloud/version.h"
 #include "absl/meta/type_traits.h"
+#include <future>
```
dbolduc, line 30: Ah, I see that we use `std::future_errc` below

dbolduc (approved):

PR #12650, opened 2023-09-15
test(storage): verify tracing stub included

dbolduc:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12650)


coryan (approved):
FWIW, I think conventional commits recommends `type(scope): description` where:

> A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis

dbolduc:
> FWIW, I think conventional commits recommends `type(scope): description` where:
> 
> > A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis

Ok. In the case of overlapping components, I will just pick the one that feels more important. In this case, `storage`.

PR #12657, opened 2023-09-15
cleanup: deal with a few more random include-what-you-use issues

devbww:
Includes an update to #12646, where the include was moved to a header.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12657)

### DIFF
--- a/google/cloud/completion_queue_test.cc
+++ b/google/cloud/completion_queue_test.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/completion_queue.h"
 #include "google/cloud/future.h"
 #include "google/cloud/internal/default_completion_queue_impl.h"
+#include "google/cloud/status.h"
 #include "google/cloud/testing_util/async_sequencer.h"
 #include "google/cloud/testing_util/fake_completion_queue_impl.h"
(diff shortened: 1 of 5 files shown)


google/cloud/completion_queue_test.cc
```diff
@@ -15,17 +15,26 @@
...
 #include "google/cloud/testing_util/async_sequencer.h"
 #include "google/cloud/testing_util/fake_completion_queue_impl.h"
 #include "google/cloud/testing_util/mock_async_response_reader.h"
 #include "google/cloud/testing_util/status_matchers.h"
 #include <google/protobuf/duration.pb.h>
 #include <google/protobuf/timestamp.pb.h>
 #include <gmock/gmock.h>
+#include <grpcpp/completion_queue.h>
+#include <grpcpp/support/async_unary_call.h>
+#include <grpcpp/support/status.h>
```
coryan, line 28: Just FYI: I do not know whether these are public headers or not.  I have not found anything in gRPC documenting which headers are intended as part of the public API.

coryan (approved):

Issue #12631, opened 2023-09-14
Change default quickstart to use project + location


coryan:
The scaffold generator assumes most services have a `List*` operation that just consumes a project.  This is less and less true in GCP, most services need a project and location (typically a region).  We should change this:

https://github.com/googleapis/google-cloud-cpp/blob/43cd3d9a65df32310bca8d8878ee322d73493a03/generator/internal/scaffold_generator.cc#L943-L952

To assume that is the case.  Maybe add a helper `google::cloud::Location` class.

Likewise, the scaffold generator can assume this is the case in:

https://github.com/googleapis/google-cloud-cpp/blob/43cd3d9a65df32310bca8d8878ee322d73493a03/generator/internal/scaffold_generator.cc#L474-L477

And remove the `EDIT HERE` comment from the CMakeLists.txt file.

devbww:
Some remaining automation work here is to try to guess an appropriate `ListFoo()` call, and eliminate the remaining _"EDIT HERE"_ markers.

coryan:
> Some remaining automation work here is to try to guess an appropriate `ListFoo()` call, and eliminate the remaining _"EDIT HERE"_ markers.

That is the work in #12632 I think @devbww  fixed this bug.

PR #12659, opened 2023-09-16
impl(common): introduce an abstraction for a Cloud location

devbww:
Introduce `google::cloud::Location` to model a location resource like `"projects/project-id/locations/location-id"`, and use this in places where we previously worked directly with strings.

After this change has made it into a release, we can update the quickstart programs (and the scaffolding code that generates them) to use `Location`.

Part of #12631.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12659)


google/cloud/internal/oauth2_external_account_credentials_test.cc
```diff
@@ -150,13 +152,14 @@ TEST(ExternalAccount, ParseAwsSuccess) {
       "https://sts.{region}.aws.example.com"
       "?Action=GetCallerIdentity&Version=2011-06-15";
   auto constexpr kTestTokenType = "urn:ietf:params:aws:token-type:aws4_request";
-  auto constexpr kTestAudience =
-      "//iam.googleapis.com/projects/$PROJECT_NUMBER/locations/global/"
-      "workloadIdentityPools/$POOL_ID/providers/$PROVIDER_ID";
+  auto const k_test_audience =
+      std::string{"//iam.googleapis.com/" +
+                  Location(Project("$PROJECT_NUMBER"), "global").FullName() +
```
coryan, line 157: I am not sure this is a good application of the new class.  The URL path and resource names match in this case, but they don't have to match (e.g. storage uses`/b/$BUCKET/o/$OBJECT` for the URL path and `projects/_/buckets/$BUCKET/objects/$OBJECT` for the resource name).

Also, the leading `k_` is no longer applicable.

examples/batch_logging.cc
```diff
@@ -45,16 +46,16 @@ void JobLogs(std::vector<std::string> const& argv) {
...
+    auto const name = location.FullName() + "/jobs/" + job_id;
     auto batch = google::cloud::batch_v1::BatchServiceClient(
         google::cloud::batch_v1::MakeBatchServiceConnection());
     auto job = batch.GetJob(name);
     if (!job) throw std::move(job).status();
 
     auto logging = google::cloud::logging_v2::LoggingServiceV2Client(
         google::cloud::logging_v2::MakeLoggingServiceV2Connection());
-    auto const project = google::cloud::Project(project_id);
     auto const log_name = "projects/" + project_id + "/logs/batch_task_logs";
```
coryan, line 59: Should we use the `project.FullName()` here too?

coryan (changes requested):

google/cloud/internal/oauth2_external_account_credentials_test.cc
devbww: Reverted.

examples/batch_logging.cc
devbww: Done.

coryan (approved):
I fixed a typo in the PR title.  You may need to do the same in the commit message.

PR #12660, opened 2023-09-16
feat: support builds for only common libraries

coryan:
In some package ecosystems it may be valuable to compile only the common libraries. Previously it was impossible to compile `grpc_utils` or `rest` unless some feature was also enabled that required these common components.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12660)

### DIFF
--- a/ci/cloudbuild/builds/cmake-split-install.sh
+++ b/ci/cloudbuild/builds/cmake-split-install.sh
@@ -48,13 +48,13 @@ io::run cmake --build cmake-out/common-libraries
 io::run cmake --install cmake-out/common-libraries --prefix "${INSTALL_PREFIX}"
 
 io::log_h2 "Building and installing popular libraries"
-mapfile -t core_cmake_args < <(cmake::common_args cmake-out/core-libraries)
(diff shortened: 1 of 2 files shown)


ci/cloudbuild/builds/cmake-split-install.sh
```diff
@@ -38,24 +38,24 @@ install_args=(
...
+
+io::log_h2 "Building and installing common libraries"
 # We need a custom build directory
+mapfile -t core_cmake_args < <(cmake::common_args cmake-out/common-libraries)
+io::run cmake "${core_cmake_args[@]}" "${install_args[@]}" \
+  -DGOOGLE_CLOUD_CPP_ENABLE="__common__"
+io::run cmake --build cmake-out/common-libraries
+io::run cmake --install cmake-out/common-libraries --prefix "${INSTALL_PREFIX}"
+
+io::log_h2 "Building and installing popular libraries"
```
devbww, line 50: Given the names of the "args" array, and the cmake-out directory, perhaps we should `s/popular/core/` for consistency.

devbww (approved):

PR #12666, opened 2023-09-18
impl(bigquery): Table benchmark program

jsrinnn:
This PR includes the following:

1) A table benchmark program that runs the GetTable and ListTables apis end to end based on input parameters 
and prints latency and throughput results . See the Table benchmark program output here: https://paste.googleplex.com/4980482805596160

2) Additionally fixed bugs in Table apis exposed by the benchmark program, related to response parsing. 
(e.g. some (expected) numbers fields are returned as strings)

3) Updated unit tests for bug fixes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12666)


coryan (changes requested):
FWIW, the code for the benchmarks is getting substantial, and it looks less and less like something we need to host in `google-cloud-cpp`.  I don't want to disrupt your work, so maybe you can finish them here and then we move them to a more appropriate repository?

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12666#-:-Nee2cmJEu8VHf_2268K:b-2wd4i8)* status: all files reviewed, 1 unresolved discussion (waiting on @jsrinnn)

___
*[`google/cloud/bigquery/v2/minimal/internal/table.h` line 58 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12666#-Nee0WWS7pUpB2zaSL1s:-Nee0WWS7pUpB2zaSL1t:bb7gkfc) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/819e29f851bc671e63fc2d90eb6665086a0b7b90/google/cloud/bigquery/v2/minimal/internal/table.h#L58)):*
> ```objc
>   std::string default_collation;
>   std::string max_staleness;
>   std::string num_bytes;
> ```

It is normal for JSON to return 64-bit numbers as strings.  As you probably know, Javascript does not support 64-bit integers.  We should be converting these from strings to integers in our code, not leaving them as strings.

coryan:
*[`google/cloud/bigquery/v2/minimal/benchmarks/table_benchmark_programs.cc` line 184 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12666#-Nee31dR6GIKQHFrW7ZG:-Nee31dS3Q79LcC7rkmC:undefined) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/819e29f851bc671e63fc2d90eb6665086a0b7b90/google/cloud/bigquery/v2/minimal/benchmarks/table_benchmark_programs.cc#L184)):*
> ```cpp
>       // If the user requests only one thread, use the current thread.
>       launch_policy = std::launch::deferred;
>     }
> ```





_Suggestion:_
```cpp
    // If the user requests only one thread, use the current thread.
    auto launch_policy = config.thread_count == 1 ? std::launch::deferred : std::launch::async;
```

jsrinnn (commented):
Yup. That is the plan. I am doing some very basic benchmarking here to make sure the apis are behaving as expected and we are in the ball-park when compared to the current driver performance. This is also helping me achieve some basic end to end testing for the v2/minimal apis with the server.  There would be comphrensive driver performance testing later on in a separate repo.

Addressedall all feedback. Benchmark Output is same as before. Please take a look when you get a chance!

Thank you!

> *[Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12666#-:-Nee3bPf1j9mMPr7jzzN:b-u94zkp)* status: all files reviewed, 2 unresolved discussions (waiting on @coryan)

___
*[`google/cloud/bigquery/v2/minimal/benchmarks/table_benchmark_programs.cc` line 184 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12666#-Nee31dR6GIKQHFrW7ZG:-NeeCfwGAoFwoeqqL_o4:b-896fix) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/819e29f851bc671e63fc2d90eb6665086a0b7b90/google/cloud/bigquery/v2/minimal/benchmarks/table_benchmark_programs.cc#L184)):*
> ```cpp
>       // If the user requests only one thread, use the current thread.
>       launch_policy = std::launch::deferred;
>     }
> ```

Done.
___
*[`google/cloud/bigquery/v2/minimal/internal/table.h` line 58 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12666#-Nee0WWS7pUpB2zaSL1s:-Nee4jKt9N2cWNLecOEA:be6pop4) ([raw file](https://github.com/googleapis/google-cloud-cpp/blob/819e29f851bc671e63fc2d90eb6665086a0b7b90/google/cloud/bigquery/v2/minimal/internal/table.h#L58)):*

It is normal for JSON to return 64-bit numbers as strings.  As you probably know, Javascript does not support 64-bit integers.  We should be converting these from strings to integers in our code, not leaving them as strings.
</blockquote></details>

Done.

You make a good point. I initially thought all the conversion can be done on the calling side in the driver repo, and this layer would just return the data as is. But it makes sense to do it here.

I have made the changes accordingly.

coryan (approved):

Issue #12332, opened 2023-08-07
Tracing for REST / oauth2 components


dbolduc:
## Problem

If the customer asks for it, we want to trace requests that any `RestClient`s make to the metadata server. This means we must pass along configuration options to the creation of the RestClient. We do not always remember to do that. Sometimes we just supply an empty set of `Options`:

Aside: We used to trace these requests to the metadata server. However, since #12098, we have had a gap in our trace coverage. I noticed this gap in the traces from our OpenTelemetry quickstart.

For example these `options_`: https://github.com/googleapis/google-cloud-cpp/blob/0679ceb1cf844c5151572e0333947dcdc7626bfa/google/cloud/internal/oauth2_authorized_user_credentials.cc#L101

originate from:

https://github.com/googleapis/google-cloud-cpp/blob/0679ceb1cf844c5151572e0333947dcdc7626bfa/google/cloud/storage/oauth2/authorized_user_credentials.cc#L83

---

## Fix

A lazy person would say:
```cc
  auto client = client_factory_(
      internal::MergeOptions(internal::CurrentOptions(), options_));
```

... but that would be in conflict with the goals of #12330 

We probably need to plumb the `Options` somehow. I am not sure how this will work for the legacy storage/oauth2 credentials.

---

There is also:

https://github.com/googleapis/google-cloud-cpp/blob/0679ceb1cf844c5151572e0333947dcdc7626bfa/google/cloud/storage/oauth2/compute_engine_credentials.cc#L75

https://github.com/googleapis/google-cloud-cpp/blob/0679ceb1cf844c5151572e0333947dcdc7626bfa/google/cloud/internal/oauth2_service_account_credentials.cc#L205

I do not know how many more such occurrences there are. Typically we should look for the factories to create `RestClient`s.

PR #12671, opened 2023-09-19
impl(rest): auth tracing with assumed credentials

dbolduc:
Part of the work for #12332 

If tracing is enabled in the connection, and no credentials are supplied via `Options`, assume that the customer wants to trace the auth components.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12671)


coryan (approved):
s/assumed credentials/default credentials/ ?

I guess in the case where the application provides the credentials it is under their control to enable (or not) tracing for them?

dbolduc:
"Default credentials" is overloaded. It could mean either `Options{}` or `Options{}.set<UnifiedCredentialsOption>(MakeGoogleDefaultCredentials())`. I am trying to refer to the former.

dbolduc:
> ... it is under [the application's] control to enable (or not) tracing for [credentials the application creates]?

Yeah, and I don't think there is a way around this.

PR #12674, opened 2023-09-19
cmake: distinguish between compute feature and libraries

scotthart:
With the addition of compute, there's no longer a 1:1 relationship between features and libraries. This PR is a start at distinguishing between them. Additional changes will be needed over time as we encounter all the areas where we conflate the two terms.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12674)


coryan (approved):
It is a bit wonky, but trying to fix it will probably snowball rather quickly.

PR #12676, opened 2023-09-19
impl(pubsub): add default message batch impl

alevenberg:
Part of [12528](https://github.com/googleapis/google-cloud-cpp/issues/12528)

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12676)


google/cloud/pubsub/internal/default_message_batch.h
```diff
@@ -0,0 +1,48 @@
...
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * A default no-op implementation of `MessageBatch`
+ */
+class DefaultMessageBatch : public MessageBatch {
+ public:
+  static std::shared_ptr<DefaultMessageBatch> Create() {
+    return std::shared_ptr<DefaultMessageBatch>(new DefaultMessageBatch());
+  }
```
devbww, line 32: This makes it seem like the default constructor should be private.

google/cloud/pubsub/internal/default_message_batch.h
```diff
@@ -0,0 +1,51 @@
...
+  ~DefaultMessageBatch() override = default;
+
+  void SaveMessage(pubsub::Message m){};
+
+  void Flush(){};
+
+  void FlushCallback(){};
+
+ private:
+  DefaultMessageBatch();
```
devbww, line 43: `= default` I presume?

alevenberg:
Rebased onto main since the ci was failing

google/cloud/pubsub/internal/default_message_batch.h
```diff
@@ -0,0 +1,51 @@
...
+
+namespace google {
+namespace cloud {
+namespace pubsub_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * A default no-op implementation of `MessageBatch`
+ */
+class DefaultMessageBatch : public MessageBatch {
```
devbww, line 28: Q.  In what way is this "default"?  Might "NoOpMessageBatch" be better?

google/cloud/pubsub/internal/noop_message_batch.h
```diff
@@ -0,0 +1,51 @@
...
+namespace cloud {
+namespace pubsub_internal {
+GOOGLE_CLOUD_CPP_INLINE_NAMESPACE_BEGIN
+
+/**
+ * A no-op implementation of `MessageBatch`
+ */
+class NoOpMessageBatch : public MessageBatch {
+ public:
+  static std::shared_ptr<NoOpMessageBatch> Create() {
```
devbww, line 30: A couple of last questions I'm probably just lacking context for.

Why have a factory at all, rather than just a constructor?  Normally we would resort to the former when there has to be something done outside the constructor to deliver a fully-formed object.  But that is not the case here, and seemingly could never be.

Why `shared_ptr` instead of `unique_ptr`?  Is the class overspecializing itself?

devbww (approved):

google/cloud/pubsub/internal/noop_message_batch.h
alevenberg: I was using factory method because it's what I saw elsewhere.

I don't think we need it. 

Removed the `Create` and made the constructor public.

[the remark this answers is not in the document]
alevenberg: No longer relevant

devbww: I think that's a good choice, at least until an external requirement emerges.

google/cloud/pubsub/internal/noop_message_batch.h
```diff
@@ -0,0 +1,46 @@
...
+ public:
+  NoOpMessageBatch() = default;
+
+  ~NoOpMessageBatch() override = default;
+
+  void SaveMessage(pubsub::Message m){};
+
+  void Flush(){};
+
+  void FlushCallback(){};
```
devbww, line 38: Should these be `override` too?

alevenberg: Done

devbww (approved):

PR #12680, opened 2023-09-20
docs(spanner): set max_delay to 16s

olavloite:
Update the custom retry sample to use a max retry delay of 16 seconds, as that is smaller than the total timeout of 60 seconds. The current values are slightly confusing for users, as it could give the impression that retry delays are not counted towards the total timeout.

- - -
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp/12680)


coryan:
Thanks for the PR.  It all looks good to me, but I will let @devbww decide as he owns this code.

devbww:
> Thanks for the PR. It all looks good to me, but I will let @devbww decide as he owns this code.

It's fine with me.  I'm only waiting to see if it exits draft mode.

devbww (approved):

